PR #99, opened 2017-09-19
Excluding nodes from restoration

### LINKED ISSUE #71: Exclude nodes from restores by default
It doesn't really make sense to restore a node, but there may be some value in including the node in the backup (especially once we get #16). We should make it so nodes are excluded from restores by default, **AND** we should make it "hard" for the user to say they want to include nodes as part of a restore.

xref #23

jrnt30:
Fixes #71

- Introduced a blacklist of resources that are non-restorable.  The
goal being that the backup can still include these resources for
logging/auditing purposes but they are explicitly added to
ExcludedResources in the RestorController's "defaulting" logic
to ensure that if someone were to explicitly ask for Node that 
they would be expressly denied.

Signed-off-by: Justin Nauman <justin.r.nauman@gmail.com>

### DIFF
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -29,6 +29,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
@@ -43,6 +44,10 @@ import (
 	"github.com/heptio/ark/pkg/util/collections"
 	kubeutil "github.com/heptio/ark/pkg/util/kube"
 )
+
+// nonRestorableResources is a blacklist for the restoration process. Any resources
+// included here are explicitly excluded from the restoration process.
+var nonRestorableResources = []string{"nodes"}
 
 type restoreController struct {
 	restoreClient       arkv1client.RestoresGetter
@@ -234,6 +239,13 @@ func (controller *restoreController) processRestore(key string) error {
 		restore.Spec.IncludedResources = []string{"*"}
 	}
 
+	excludedResources := sets.NewString(restore.Spec.ExcludedResources...)
+	for _, nonrestorable := range nonRestorableResources {
+		if !excludedResources.Has(nonrestorable) {
+			restore.Spec.ExcludedResources = append(restore.Spec.ExcludedResources, nonrestorable)
+		}
+	}
+
 	// validation
 	if restore.Status.ValidationErrors = controller.getValidationErrors(restore); len(restore.Status.ValidationErrors) > 0 {
 		restore.Status.Phase = api.RestorePhaseFailedValidation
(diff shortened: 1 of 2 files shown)


ncdc:
@jrnt30 we need to be able to restore some endpoints: those that are managed by a user. Those are services that don't have a selector specified.

ncdc:
Also FYI the GitHub "fixes" notation requires each issue to have "fixes" in front of it.

jrnt30:
Thanks for the clarification.  I'll throw this in WIP and see if I can use what I have going or if it needs to be significantly refactored for endpoints.  If so I will probably just remove the endpoint support for now.

pkg/controller/restore_controller.go
```diff
@@ -229,6 +234,13 @@ func (controller *restoreController) processRestore(key string) error {
 		restore.Spec.IncludedResources = []string{"*"}
 	}
 
+	curIncludes := sets.NewString(restore.Spec.ExcludedResources...)
```
ncdc: Maybe call this excludedResources or something other than *Includes? 😄

ncdc:
Generally LGTM. Just the 1 comment on variable naming. @skriss any other comments?

pkg/controller/restore_controller.go
```diff
@@ -284,6 +296,10 @@ func (controller *restoreController) getValidationErrors(itm *api.Restore) []str
 		validationErrors = append(validationErrors, "BackupName must be non-empty and correspond to the name of a backup in object storage.")
 	}
 
+	if errs := collections.ValidateIncludesExcludes(itm.Spec.IncludedResources, nonRestorableResources); len(errs) > 0 {
```
skriss: This could be misleading if the error returned from ValidateIncludesExcludes is due to something other than nodes being explicitly included. Maybe rather than calling ValidateIncludesExcludes for this, just check if anything in the nonRestorable list is in IncludedResources?

jrnt30: Good point, thank you.

pkg/controller/restore_controller.go
```diff
@@ -41,8 +41,13 @@ import (
 	listers "github.com/heptio/ark/pkg/generated/listers/ark/v1"
 	"github.com/heptio/ark/pkg/restore"
 	"github.com/heptio/ark/pkg/util/collections"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+// NonRestorableResources is a blacklist for the restoration process. Any resources
```
ncdc: Leading lowercase n

ncdc: @jrnt30 bump

jrnt30: I'll try and take a look on the train on that work.

jrnt30: Looking at this closer again, I suppose I related the blacklist of resources similarly to the setting of a default as we do with the `IncludedResources` and `IncludedNamespaces`.  

If we push this down to the `pkg/restore` I don't think as cleanly segregates the validation logic being exposed to the client either.  

Were you thinking that we put this in the `kubernetesRestorer.Restore` directly or simply a utility function checking these things?  I'm afraid I may be missing part of your thought process here (it may be simply that you are proposing a larger refactor than I am considering looking at the type signature here).

ncdc: Let's say that someone decides to pull in `github.com/heptio/ark/pkg/...` to use Ark as a library (we already know of at least one project interested in doing so). If the logic to exclude nodes is only in the controller, then anyone using the library directly and not using the controller would end up restoring nodes unless they also write logic to exclude them.

The more I think about this, given that the other defaulting logic is in the controllers, I think it's fine to include this here for now in the controller. We may decide to move more logic "down" in the future, but I don't think that possibility needs to hold up this PR.

jrnt30: Understood, thanks.  I do think having that more centralized may be beneficial in those cases. 

Out of curiosity, is the project you are referring to public?

ncdc: https://github.com/kris-nova/kubicorn/issues/195

ncdc:
@jrnt30 squash please & I'll merge

ncdc:
@jrnt30 sorry, decided to get logrus in first. Mind rebasing?

jrnt30:
No problem, that was a big one for sure. Should be good now.

PR #102, opened 2017-09-25
Preserve AWS & GCP volume AZ when snapshotting and restoring PVs.

ashish-amarnath:
Update VolumeInfo struct to store the AZ where the vol was provisioned. Update GetVolumeInfo method to get AZ information for the volume Lookup volume's AZ from VolumeInfo to use while restoring volume from it's snapshot.

Basic validation with change on heptio/ark:
- Compiles successfully
- 'go test./...' all tests pass
- 'make all' is successful


ashish-amarnath:
Looking for some initial feedback and suggestions on validating this change, esp on the GCP and Azure interface implementations.

skriss:
Thanks for the PR @ashish-amarnath! We'll take a look shortly and provide some feedback.

ashish-amarnath:
It would be ideal to remove the availability zone in the AWSConfig altogether. As having a comma-separated list of strings may break assumptions that other areas of code may be making when using the `AvailabilityZone` from the AWSConfig.

skriss:
A few initial thoughts on approach:

-I think it'll be best to pull the AvailabilityZone for a PV from the Kubernetes API rather than from the cloud provider API. We can do this in [volume_snapshot_action.go](https://github.com/heptio/ark/blob/master/pkg/backup/volume_snapshot_action.go) by looking up the appropriate label in the unstructured volume map. The reason for this is that for GCE, knowing the zone is required to make API calls (for example, to describe the volume).
-storing the AZ in the `VolumeBackupInfo` struct (as you have it) makes sense.
-we should be able to remove `AvailabilityZone` and `Zone` from AWS and GCP configs, respectively, since they're only used for snapshot creation/restores. The validation step in the constructor for the BlockStorageAdapters is no longer necessary.
-we'll have to pass the AZ as an arg to a few additional SnapshotService methods to support GCP - looks like `CreateVolumeFromSnapshot`, `GetVolumeInfo`, `IsVolumeReady`, and `CreateSnapshot` at a glance
-Azure has [just announced AZs](https://azure.microsoft.com/en-us/blog/introducing-azure-availability-zones-for-resiliency-and-high-availability/) in preview, so for now we don't need to add support

Let me know if this approach seems reasonable. Happy to do more detailed review/guidance as well.

cc @jrnt30 @ncdc - feel free to comment if you have other thoughts or disagree with any of my comments!

jrnt30:
@skriss I need to take a closer look at your first suggestion to understand the distinction there, but will let you know if I have any concerns.  Otherwise the approach is inline with what I was thinking should be possible, thanks for the validation and the heads up on the GCP nuances. 
 
@ashish-amarnath If you get stuck or want to pair, I would be happy to spend some time on Zoom or Screen Hero chatting about this depending on schedule alignment.

ashish-amarnath:
@skriss: I understand your suggestions and you make valid points. Let me take another stab at it and update the PR. I will not be able to work on this for the next day or so.
@jrnt30: I am not stuck but It would be great to pair. I will ping you on slack in the next day or so.

ashish-amarnath:
@skriss I've updated the PR to read the availability-zone info from the PV itself, instead of getting it from the cloud provider. This way we'll take a dependency on kubernetes' labelling of AZs instead of the cloud-provider APIs.

jrnt30:
Thanks @ashish-amarnath One thing we need to have you do is `sign-off` on the commits please.  Instructions can be found at [Contributing Docs](https://github.com/heptio/ark/blob/master/CONTRIBUTING.md)

pkg/backup/volume_snapshot_action_test.go
```diff
@@ -77,49 +77,49 @@ func TestVolumeSnapshotAction(t *testing.T) {
...
 			volumeInfo: map[string]v1.VolumeBackupInfo{
-				"vol-abc123": v1.VolumeBackupInfo{Type: "io1", Iops: &iops, SnapshotID: "snap-1"},
+				"vol-abc123": v1.VolumeBackupInfo{Type: "io1", Iops: &iops, SnapshotID: "snap-1", AvailabilityZone: "us-east-1c"},
 			},
 		},
 		{
 			name:                   "aws - dynamically provisioned volume id",
 			snapshotEnabled:        true,
-			pv:                     `{"apiVersion": "v1", "kind": "PersistentVolume", "metadata": {"name": "mypv"}, "spec": {"awsElasticBlockStore": {"volumeID": "aws://us-west-2a/vol-abc123"}}}`,
+			pv:                     `{"apiVersion": "v1", "kind": "PersistentVolume", "metadata": {"name": "mypv", "labels": {"failure-domain.beta.kubernetes.io/zone": "us-east-1c"}}, "spec": {"awsElasticBlockStore": {"volumeID": "aws://us-west-2a/vol-abc123"}}}`,
```
jrnt30: It's a nit, but can we have the `failureDomain` and the nested "AZ" from the `volumeID` attributes match up to be more logically accurate please?

pkg/cloudprovider/snapshot_service.go
```diff
@@ -67,8 +67,8 @@ func NewSnapshotService(blockStorage BlockStorageAdapter) SnapshotService {
 	}
 }
 
-func (sr *snapshotService) CreateVolumeFromSnapshot(snapshotID string, volumeType string, iops *int64) (string, error) {
-	volumeID, err := sr.blockStorage.CreateVolumeFromSnapshot(snapshotID, volumeType, iops)
+func (sr *snapshotService) CreateVolumeFromSnapshot(snapshotID string, volumeType string, volAZ string, iops *int64) (string, error) {
```
jrnt30: It's a nit... but you can have `snapshotID string, volumeType, volAZ string, iops *int64` here

pkg/cloudprovider/aws/block_storage_adapter.go
```diff
@@ -88,10 +73,10 @@ func NewBlockStorageAdapter(region, availabilityZone string) (cloudprovider.Bloc
 // from snapshot.
 var iopsVolumeTypes = sets.NewString("io1")
 
-func (op *blockStorageAdapter) CreateVolumeFromSnapshot(snapshotID, volumeType string, iops *int64) (volumeID string, err error) {
+func (op *blockStorageAdapter) CreateVolumeFromSnapshot(snapshotID, volumeType, volAZ string, iops *int64) (volumeID string, err error) {
```
jrnt30: Perhaps we just fully spell this out to `volumeAZ` similar to `volumeType`

pkg/backup/volume_snapshot_action_test.go
```diff
@@ -155,6 +155,30 @@ func TestVolumeSnapshotAction(t *testing.T) {
 			pv:              `{"apiVersion": "v1", "kind": "PersistentVolume", "metadata": {"name": "mypv"}, "spec": {"gcePersistentDisk": {"pdName": "pd-abc123"}}}`,
 			expectError:     true,
 		},
+		{
```
ncdc: Why do we need these 2 additional test cases - how are they different from what's above?

pkg/cloudprovider/aws/block_storage_adapter.go
```diff
@@ -64,22 +60,11 @@ func NewBlockStorageAdapter(region, availabilityZone string) (cloudprovider.Bloc
 		return nil, err
 	}
 
-	// validate the availabilityZone
 	var (
```
ncdc: No need for a `var` block any more - please move this down below into the `blockStorageAdapter` assignment directly.

pkg/apis/ark/v1/backup.go
```diff
@@ -109,6 +109,10 @@ type VolumeBackupInfo struct {
 	// API.
 	Type string `json:"type"`
 
+	// AvailabilityZone is the where the volume is provisioned
+	// in the cloud provider.
+	AvailabilityZone string `json:"avaialabiltyZone"`
```
skriss: small typo in the json tag -- should be `json:"avaialabiltyZone"`

pkg/backup/volume_snapshot_action.go
```diff
@@ -59,6 +59,14 @@ func (a *volumeSnapshotAction) Execute(ctx ActionContext, volume map[string]inte
 
 	metadata := volume["metadata"].(map[string]interface{})
 	name := metadata["name"].(string)
+	labels := metadata["labels"]
```
skriss: There are some helpers for navigating these maps in pkg/util/collections. You can get the labels map with `labelsMap, err := collections.GetMap(metadata, "labels")`. Unfortunately, you can't go straight to the zone right now because our helpers take dot-separated paths, which won't work given this label, but this should at least simplify getting the labels map.

pkg/backup/volume_snapshot_action.go
```diff
@@ -59,6 +59,14 @@ func (a *volumeSnapshotAction) Execute(ctx ActionContext, volume map[string]inte
...
+	var pvfailureDomainZone string
+	if labels != nil {
+		labelsMap := labels.(map[string]interface{})
+		pvfailureDomainZone = labelsMap["failure-domain.beta.kubernetes.io/zone"].(string)
+	} else {
+		ctx.log("Backup %q: PersistentVolume %q has no 'failure-domain.beta.kubernetes.io/zone' label", backupName, name)
```
skriss: I think we probably need to return an error here, since we can't proceed if we don't know the zone.

pkg/apis/ark/v1/backup.go
ashish-amarnath: Thanks for catching that!

pkg/backup/volume_snapshot_action.go
ashish-amarnath: Will do. That should simplify things a little bit.

pkg/backup/volume_snapshot_action_test.go
ashish-amarnath: Good catch! Will do.

pkg/backup/volume_snapshot_action_test.go
ashish-amarnath: These are not different from the other test cases and are repetitive tests. Will remove them.

pkg/cloudprovider/aws/block_storage_adapter.go
ashish-amarnath: Makes sense. Will do.

pkg/cloudprovider/aws/block_storage_adapter.go
ashish-amarnath: Good idea.

pkg/cloudprovider/snapshot_service.go
ashish-amarnath: Agree. Didn't want to change the style of the original code. But, will change.

ashish-amarnath:
I am not sure what I am doing wrong for the sign-off not to be picked up. Any suggestions on how to resolve the sign-off?

jrnt30:
I think the issue you have is there a mixture of signed/unsigned commits and a few that are repeated.  The good news is that it's fixable, the bad news it that it's a bit of a pain :) 

This assumes that you have two remotes for your repo `origin` that points to the Heptio Ark repo upstream and (asish-amarnath) current branch points to your own fork at `ashish-amarnath`. 

```
git fetch --all 			# ensure origin/master is updated
git rebase -i origin/master # rewrite your local history to fix a few issues
```

This should be the pick-list you need to collapse the incorrect commits to remove the duplicate sets of commits.
```
# pick 48459dc WIP: Preserve AWS EBS volume AZ when snapshotting and restoring PVs.
# drop d9cbaf8 WIP: Preserve AWS EBS volume AZ when snapshotting and restoring PVs.
# pick 7d91f34 WIP: Read availability zone on PVs from PV fault-domain label
# pick b286455 WIP: Remove AvailabilityZone config from AWS and GCP configuration
# drop 805b315 WIP: Read availability zone on PVs from PV fault-domain label
# drop 82744ef WIP: Remove AvailabilityZone config from AWS and GCP configuration
# pick f5e8565 Address code review comments:
```

If you do a `git log` after that, you should see a bit cleaner git history in relation to your branch and `origin/master`.  

Test your stuff and if that looks good `git push --force ashish-amarnath` and lets see what the sign-off check does then.

NOTE: If you run into a problem, you should be able to `git pull` to reset and try again

skriss:
I'm happy with the code. We do need to update the docs and sample files to remove AZ references from config. The main places are:

examples/aws/00-ark-config.yaml
examples/gcp/00-ark-config.yaml
docs/config-definition.md

But I would also look around the docs/ folder a bit more to see if there are any other references.

ncdc:
@ashish-amarnath when doing the interactive rebase, you could also choose to squash everything down to 1 commit if that makes sense. Let us know if you need any more guidance here. Thanks!

ashish-amarnath:
@jrnt30 Thanks for the suggestions. I will follow it and reach out if I run into issues.
@skriss Will look at the docs to make changes.
@ncdc Squashing commits make sense.

ncdc:
@ashish-amarnath when you have time, please rebase, squash down to 1 commit, and update docs as @skriss suggested. Thanks!

ashish-amarnath:
@skriss, @ncdc, @jrnt30 This PR should be good to go.

skriss:
@ashish-amarnath which platforms (AWS, GCP, Azure) have you been able to test this on? I can help out with some testing on any of those.

pkg/apis/ark/v1/backup.go
```diff
@@ -109,6 +109,10 @@ type VolumeBackupInfo struct {
 	// API.
 	Type string `json:"type"`
 
+	// AvailabilityZone is the where the volume is provisioned
+	// in the cloud provider.
+	AvailabilityZone string `json:"availabilityZone"`
```
skriss: let's an an `omitempty` to the JSON tag (same as for Iops just below).

pkg/backup/volume_snapshot_action.go
```diff
@@ -60,6 +61,13 @@ func (a *volumeSnapshotAction) Execute(ctx ActionContext, volume map[string]inte
 
 	metadata := volume["metadata"].(map[string]interface{})
 	name := metadata["name"].(string)
+	var pvfailureDomainZone string
+	labelsMap, err := collections.GetMap(metadata, "labels")
+	if err == nil {
+		pvfailureDomainZone = labelsMap["failure-domain.beta.kubernetes.io/zone"].(string)
```
skriss: need to do a nil-check here when getting `failure-domain.beta.kubernetes.io/zone` out of the map to handle the case where this label doesn't exist (currently it panics when it tries to convert to string). Can you also make sure each situation has an appropriate log statement (no labels collection, no zone key, or zone found)?

skriss:
@ashish-amarnath I was able to test this on all three platforms; two minor issues which I added comments for but otherwise it's looking good!

pkg/backup/volume_snapshot_action.go
```diff
@@ -60,6 +61,17 @@ func (a *volumeSnapshotAction) Execute(ctx ActionContext, volume map[string]inte
 
 	metadata := volume["metadata"].(map[string]interface{})
 	name := metadata["name"].(string)
+	var pvfailureDomainZone string
+	labelsMap, err := collections.GetMap(metadata, "labels")
+	if err == nil {
+		if labelsMap["failure-domain.beta.kubernetes.io/zone"] != nil {
+			pvfailureDomainZone = labelsMap["failure-domain.beta.kubernetes.io/zone"].(string)
+		} else {
+			ctx.log("error getting 'failure-domain.beta.kubernetes.io/zone' label on PersistentVolume %q for backup %q.\n", name, backupName)
```
ncdc: No `\n` needed

pkg/backup/volume_snapshot_action.go
```diff
@@ -60,6 +61,17 @@ func (a *volumeSnapshotAction) Execute(ctx ActionContext, volume map[string]inte
...
+		}
+	} else {
+		ctx.log("error getting labels on PersistentVolume %q for backup %q. ", name, backupName)
```
ncdc: Please remove the extra space at the end of the string

ncdc: Please include the error via `%v`

pkg/backup/volume_snapshot_action.go
ncdc: Please update the log message to indicate that this label isn't present

pkg/backup/volume_snapshot_action.go
```diff
@@ -60,6 +61,17 @@ func (a *volumeSnapshotAction) Execute(ctx ActionContext, volume map[string]inte
 
 	metadata := volume["metadata"].(map[string]interface{})
 	name := metadata["name"].(string)
+	var pvfailureDomainZone string
```
ncdc: Capital F for Failure

skriss:
re-tested on Azure & GCP; looks good

skriss:
good to merge. we'll tweak the log messages in a follow-up. @ashish-amarnath thanks again for the contribution!!

ashish-amarnath:
@skriss Thanks for the merge!
@ncdc If the logging changes that you requested didn't make it in, will follow this up with another PR just for the logging changes.

skriss:
@ashish-amarnath I addressed the requested changes from @ncdc so we're good here. Thanks!

PR #113, opened 2017-10-04
Add a FAQ

julia-stripe:
I had some questions about Ark that the documentation didn't answer. @ncdc very helpfully answered my questions in Slack so I am contributing them as documentation!

Feel totally free to reject/change this PR if you don't think it's helpful to other people, this is partly just an attempt on my part to make sure I've understood Ark properly.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -2,7 +2,7 @@
 
 ## When is it appropriate to use Ark instead of etcd's built in backup/restore?
 
-If you only want to backup/restore a single etcd cluster, you may be better of using etcd's backup
+If you only want to backup/restore a single etcd cluster, you may be better off using etcd's backup
 and restore tooling. However, doing so restricts you to reconstructing your Kubernetes cluster data
 exactly as it was preserved. Etcd's restore tooling is good if what you want is to recover from
 data loss in a single etcd cluster, but does not support more sophisticated restores such as cluster
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,0 +1,20 @@
+# FAQ
+
+## When is it appropriate to use Ark instead of etcd's built in backup/restore?
+
+If you only want to backup/restore a single etcd cluster, you're probably better off using etcd's
+built in backup and restore tooling.
+
+Ark is useful for:
+
+* backing up etcd resources and persistent volume state
+* cluster migrations
+* backing up a subset of your Kubernetes resources
+* backing up Kubernetes resources that are stored across multiple etcd clusters (for example if you
+  run a custom apiserver)
+
+## Will Ark restore my Kubernetes resources exactly the way they were before?
+
+Yes, with some exceptions. For example, when Ark restores pods it deletes the `nodeName` from the
+pod so that they can be scheduled onto a new node. You can see some more examples of the differences
+in [pod_restorer.go](https://github.com/heptio/ark/blob/master/pkg/restore/restorers/pod_restorer.go)
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -2,12 +2,15 @@
 
 ## When is it appropriate to use Ark instead of etcd's built in backup/restore?
 
-If you only want to backup/restore a single etcd cluster, you're probably better off using etcd's
-built in backup and restore tooling.
+If you only want to backup/restore a single etcd cluster, you may be better of using etcd's backup
+and restore tooling. However, doing so restricts you to reconstructing your Kubernetes cluster data
+exactly as it was preserved. Etcd's restore tooling is good if what you want is to recover from
+data loss in a single etcd cluster, but does not support more sophisticated restores such as cluster
+migrations and restoring Kubernetes state stored across multiple etcd clusters.
 
 Ark is useful for:
 
-* backing up etcd resources and persistent volume state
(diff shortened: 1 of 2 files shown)


julia-stripe:
Made some updates!

docs/faq.md
```diff
@@ -0,0 +1,23 @@
+# FAQ
+
+## When is it appropriate to use Ark instead of etcd's built in backup/restore?
+
+If you only want to backup/restore a single etcd cluster, you may be better of using etcd's backup
```
abiogenesis-now: small typo: s/of/off

skriss: nit: "...better of using..." -> "...better off using..."

docs/faq.md
```diff
@@ -0,0 +1,23 @@
...
+and restore tooling. However, doing so restricts you to reconstructing your Kubernetes cluster data
+exactly as it was preserved. Etcd's restore tooling is good if what you want is to recover from
```
abiogenesis-now: There's a little redundancy here, so I'd suggest rephrasing as:

> Etcd's restore tooling is good for recovering from data loss in a single etcd cluster, but does not support more sophisticated restores. 
>
> Ark is useful in these scenarios, which include:

docs/faq.md
```diff
@@ -0,0 +1,23 @@
...
+
+* backing up both Kubernetes resources and persistent volume state
+* cluster migrations
+* backing up a subset of your Kubernetes resources
+* backing up Kubernetes resources that are stored across multiple etcd clusters (for example if you
+  run a custom apiserver)
+
+## Will Ark restore my Kubernetes resources exactly the way they were before?
+
+Yes, with some exceptions. For example, when Ark restores pods it deletes the `nodeName` from the
```
abiogenesis-now: @ncdc should we also mention the volume mount removal for the default service account token?

e.g. if we're mentioning "exceptions" we should be able to say something along the lines of:

> The chief differences that Ark makes during a restore are:
> * ...
> * ...

So that readers can feel like they got most of what they needed to know without resorting to skimming the source code. (If we mention exceptions and only describe one, they may feel less comfortable).

abiogenesis-now: Also, I'm wondering if this particular answer should actually be documented in https://github.com/julia-stripe/ark/blob/bbef7a3e33f0d8d9ad5e948c09f557ec616bb11a/docs/concepts.md#3-restores, or if that is too deep in the docs that no one will see it.

docs/faq.md
abiogenesis-now: bump

docs/faq.md
julia-stripe: i'm going to leave this one to you all to resolve because I don't actually know what the chief differences with Ark are.

ncdc: I think it is a good idea to document the behavior of each custom backup action and restorer. I don't think it needs to be in this PR. For now it's ok to link to the pod restorer source code. Longer term, that link should be replaced to point at the docs on the actions and restorers.

abiogenesis-now: SGTM!

docs/faq.md
```diff
@@ -0,0 +1,21 @@
+# FAQ
+
+## When is it appropriate to use Ark instead of etcd's built in backup/restore?
+
+Etcd's restore tooling is good for recovering from data loss in a single etcd cluster, but does not
+support more sophisticated restores. If you only want to recover from data loss in a single etcd
+cluster, you're likely better off using etcd's backup/restore tooling.
```
ncdc: I still prefer avoiding saying "likely" here. I don't have the best way to phrase this in my head yet, but basically if your goal is to backup and restore your cluster _exactly as it was_, then you're really treating your cluster as a pet, which can make recovery more difficult (the node names have to be the same because they're recorded in pod.spec.nodeName, for example). We feel that Ark is a better approach because it gives you the ability to throw away an unstable cluster and restore your Kube resources & data into a new cluster.

ncdc:
ok to test

docs/faq.md
jvns: feel free to change it however you think is best! =)

jvns: it might be good to say

> We feel that Ark is a better approach because it gives you the ability to throw away an unstable cluster and restore your Kube resources & data into a new cluster.

explicitly, that's an interesting perspective that I hadn't considered.

docs/faq.md
```diff
@@ -0,0 +1,21 @@
...
+
+Examples of cases where Ark is useful:
+
+* backing up both Kubernetes resources and persistent volume state
+* cluster migrations
+* backing up a subset of your Kubernetes resources
+* backing up Kubernetes resources that are stored across multiple etcd clusters (for example if you
+  run a custom apiserver)
```
timoreimann: I wonder if the following might be worthwhile adding to the list too: Ark is also a great option if you don't have access to etcd at all, such as with managed Kubernetes offerings. One example provider where that's the case is Google Container Engine.

(We actually run on GKE and picked Ark for this particular reason (and a few more)).

docs/faq.md
ncdc: I think it'd be a good idea to back up etcd prior to upgrading etcd itself, for example.

I'm going to merge this and do a follow-up.

PR #122, opened 2017-10-05
logging improvements

skriss:
Fixes #116 
Fixes #118 
Fixes #120

PR #123, opened 2017-10-05
switch per-backup/restore logs to use logrus

### LINKED ISSUE #114: Convert backup logs to use logrus

### LINKED ISSUE #115: Convert restore logs to use logrus

skriss:
Fixes #114 
Fixes #115 

(to be merged after #122)

### DIFF
--- a/pkg/backup/backup.go
+++ b/pkg/backup/backup.go
@@ -20,12 +20,12 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"encoding/json"
-	"fmt"
 	"io"
 	"strings"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


pkg/backup/backup.go
```diff
@@ -133,13 +133,11 @@ func getNamespaceIncludesExcludes(backup *api.Backup) *collections.IncludesExclu
 }
 
 type logger struct {
```
davecheney: Does this provide much value? Why not just use the logrus logger directly?

pkg/restore/restore.go
```diff
@@ -223,13 +225,11 @@ func (kr *kubernetesRestorer) Restore(restore *api.Restore, backup *api.Backup,
...
 type logger struct {
-	w io.Writer
+	impl *logrus.Logger
 }
 
 func (l *logger) log(msg string, args ...interface{}) {
-	// TODO use a real logger that supports writing to files
-	now := time.Now().Format(time.RFC3339)
-	fmt.Fprintf(l.w, now+" "+msg+"\n", args...)
+	l.impl.Infof(msg, args...)
```
davecheney: same question as above; is there value in adding an abstraction above logrus?

pkg/backup/backup.go
skriss: @davecheney no, was just taking the smallest possible step towards logrus here. Plan is to get rid of this intermediate layer.

davecheney:
SGTM. Make it so.

PR #126, opened 2017-10-10
tweak volume snapshot action logging messages

skriss:
Per https://github.com/heptio/ark/pull/102#pullrequestreview-68318327


skriss:
@ncdc PTAL.

PR #132, opened 2017-10-12
Restructure backup format for resource prioritization

dgoodwin:
Restructure backups for resource prioritization.
    
    Previously the directory structure separated resources depending on
    whether or not they were cluster or namespace scoped. All cluster
    resources were restored first, then all namespace resources. Priority
    did not apply across both and you could not order any namespace
    resources before any cluster resources.
    
    This restructure sorts first on resource type.
    
    resources/serviceaccounts/namespaces/ns1.json
    resources/nodes/cluster/node1.json
    
    This will break old backups as the format is no longer consistent as
    announced on the Google group.

### DIFF
--- a/pkg/backup/backup.go
+++ b/pkg/backup/backup.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -473,9 +474,9 @@ func (ib *realItemBackupper) backupItem(ctx *backupContext, item map[string]inte
 
 	var filePath string
 	if namespace != "" {
-		filePath = strings.Join([]string{api.NamespaceScopedDir, namespace, groupResource.String(), name + ".json"}, "/")
+		filePath = filepath.Join(api.ResourcesDir, groupResource.String(), api.NamespaceScopedDir, namespace, name+".json")
 	} else {
-		filePath = strings.Join([]string{api.ClusterScopedDir, groupResource.String(), name + ".json"}, "/")
+		filePath = filepath.Join(api.ResourcesDir, groupResource.String(), api.ClusterScopedDir, name+".json")
 	}
 
 	itemBytes, err := json.Marshal(item)
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -20,10 +20,10 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
-	"path"
 	"path/filepath"
 	"sort"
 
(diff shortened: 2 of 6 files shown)


dgoodwin:
@ncdc I think this is ready to be looked at now.

ncdc:
@dgoodwin thanks, I'll start reviewing shortly. When we're ready to merge, you'll need to make sure your commit(s) have signoff (`git commit --signoff`) as per https://github.com/heptio/ark/blob/master/CONTRIBUTING.md.

ok test test

ncdc:
ok test test

ncdc:
ok to test

pkg/backup/backup.go
```diff
@@ -473,9 +473,9 @@ func (ib *realItemBackupper) backupItem(ctx *backupContext, item map[string]inte
 
 	var filePath string
 	if namespace != "" {
-		filePath = strings.Join([]string{api.NamespaceScopedDir, namespace, groupResource.String(), name + ".json"}, "/")
+		filePath = strings.Join([]string{api.ResourcesDir, groupResource.String(), api.NamespaceScopedDir, namespace, name + ".json"}, "/")
```
ncdc: `filepath.Join`

pkg/backup/backup.go
```diff
@@ -473,9 +473,9 @@ func (ib *realItemBackupper) backupItem(ctx *backupContext, item map[string]inte
...
 	} else {
-		filePath = strings.Join([]string{api.ClusterScopedDir, groupResource.String(), name + ".json"}, "/")
+		filePath = strings.Join([]string{api.ResourcesDir, groupResource.String(), api.ClusterScopedDir, name + ".json"}, "/")
```
ncdc: `filepath.Join`

pkg/restore/restore.go
```diff
@@ -259,50 +259,98 @@ func (ctx *context) execute() (api.RestoreResult, api.RestoreResult) {
...
 	if err != nil {
 		addArkError(&errors, err)
 		return warnings, errors
 	}
-	if !exists {
-		return warnings, errors
-	}
 
-	nses, err := ctx.fileSystem.ReadDir(namespacesPath)
+	resourceDirs, err := ctx.fileSystem.ReadDir(resourcesDir)
```
ncdc: Do you think there's any performance advantage one way or another between reading the resources dir up front like this, and just calling `DirExists` on each `resource` from `ctx.prioritizedResources`?

pkg/restore/restore.go
```diff
@@ -259,50 +259,98 @@ func (ctx *context) execute() (api.RestoreResult, api.RestoreResult) {
...
+		rscDir := resourceDirsMap[resource.String()]
+		if rscDir == nil {
 			continue
 		}
-		nsPath := path.Join(namespacesPath, ns.Name())
 
-		if !namespaceFilter.ShouldInclude(ns.Name()) {
-			ctx.infof("Skipping namespace %s", ns.Name())
-			continue
+		resourcePath := path.Join(resourcesDir, rscDir.Name())
```
ncdc: `filepath.Join`

pkg/restore/restore.go
```diff
@@ -259,50 +259,98 @@ func (ctx *context) execute() (api.RestoreResult, api.RestoreResult) {
...
+
+		clusterSubDir := path.Join(resourcePath, api.ClusterScopedDir)
```
ncdc: `filepath.Join`

pkg/restore/restore.go
```diff
@@ -259,50 +259,98 @@ func (ctx *context) execute() (api.RestoreResult, api.RestoreResult) {
...
+		clusterSubDirExists, err := ctx.fileSystem.DirExists(clusterSubDir)
+		if err != nil {
+			addArkError(&errors, err)
+			return warnings, errors
+		}
+		if clusterSubDirExists {
+			w, e := ctx.restoreResource(resource.String(), "", clusterSubDir)
+			merge(&warnings, &w)
+			merge(&errors, &e)
```
ncdc: `continue`?

pkg/restore/restore.go
```diff
@@ -259,50 +259,98 @@ func (ctx *context) execute() (api.RestoreResult, api.RestoreResult) {
...
+			if err != nil {
+				addArkError(&errors, err)
+				return warnings, errors
+			}
+
+			for _, nsDir := range nsDirs {
+				if !nsDir.IsDir() {
+					continue
+				}
+				nsPath := path.Join(nsSubDir, nsDir.Name())
```
ncdc: `filepath.Join`

pkg/restore/restore.go
```diff
@@ -259,50 +259,98 @@ func (ctx *context) execute() (api.RestoreResult, api.RestoreResult) {
...
+			addArkError(&errors, err)
+			return warnings, errors
+		}
+		if clusterSubDirExists {
+			w, e := ctx.restoreResource(resource.String(), "", clusterSubDir)
+			merge(&warnings, &w)
+			merge(&errors, &e)
+		}
+
+		nsSubDir := path.Join(resourcePath, api.NamespaceScopedDir)
```
ncdc: `filepath.Join`

pkg/restore/restore.go
```diff
@@ -259,50 +259,98 @@ func (ctx *context) execute() (api.RestoreResult, api.RestoreResult) {
...
+					continue
+				}
 
-		w, e := ctx.restoreNamespace(ns.Name(), nsPath)
-		merge(&warnings, &w)
-		merge(&errors, &e)
+				nsName := nsDir.Name()
+
+				// fetch mapped NS name
+				// TODO: what is a mapped ns name? for renames?
```
ncdc: This is when you do `ark restore create --namespace-mappings foo:bar,hello:goodbye,...`. Any resources in `foo` will be restored to `bar` instead, etc.

pkg/restore/restore.go
```diff
@@ -259,50 +259,98 @@ func (ctx *context) execute() (api.RestoreResult, api.RestoreResult) {
...
+
+				// ensure namespace exists
+				ns := &v1.Namespace{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: mappedNsName,
+					},
+				}
+				if _, err := kube.EnsureNamespaceExists(ns, ctx.namespaceClient); err != nil {
+					addArkError(&errors, err)
+					return warnings, errors
```
ncdc: Shouldn't this continue the `nsDirs` loop on line 318 to try additional namespaces?

pkg/restore/restore.go
dgoodwin: I can't come up with a theory one way or the other, would have to test but in context of restore my guess would be not significant at this point. (backup might be another story as I assume people might be running that a lot more)

If worthwhile let me know and I'll test next week.

pkg/restore/restore.go
dgoodwin: Must be no such thing as a resource that could be cluster or namespace scoped. Will add that in.

pkg/restore/restore.go
dgoodwin: Thanks, I think we might need to make use of this at some point.

pkg/restore/restore.go
dgoodwin: Whoah, yeah that's a bad artifact from the refactor, good catch. Changing to continue.

dgoodwin:
Updated and added signoff.

dgoodwin:
One potential point of consideration, should I try to detect and warn about old backup tar.gzs? 

Probably a case for backup.tar.gz versioning as well here.

ncdc:
Let's do that in a follow-up, at least the bit about writing some metadata to the backup tarball (and let's make sure we add an issue for that). I'm less concerned right now about detecting the old backup format and bailing out early, although that's probably easy to do, once we add in the metadata.

I am in the process of restructuring pkg/backup so it'll be easier if we do this after my exec hooks land.

dgoodwin:
Works for me.

pkg/restore/restore.go
```diff
@@ -340,73 +387,17 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
 	}
 }
 
-// restoreNamespace restores the resources from a specified namespace directory in the backup,
-// or from the cluster-scoped directory if no namespace is specified.
-func (ctx *context) restoreNamespace(nsName, nsPath string) (api.RestoreResult, api.RestoreResult) {
+// restoreResource restores the specified cluster or namespace scoped resource. If namespace is
+// empty we are restoring a cluster level resource, otherwise into the specified namespace.
+func (ctx *context) restoreResource(resource string, namespace string, resourcePath string) (api.RestoreResult, api.RestoreResult) {
```
skriss: `restoreResource(resource, namespace, resourcePath string)`

skriss:
[This doc](https://github.com/heptio/ark/blob/master/docs/output-file-format.md) also needs to be updated

pkg/restore/restore.go
```diff
@@ -259,50 +258,98 @@ func (ctx *context) execute() (api.RestoreResult, api.RestoreResult) {
...
-		merge(&errors, &e)
-	}
+	namespaceFilter := collections.NewIncludesExcludes().Includes(ctx.restore.Spec.IncludedNamespaces...).Excludes(ctx.restore.Spec.ExcludedNamespaces...)
 
-	// namespace-scoped
-	namespacesPath := path.Join(dir, api.NamespaceScopedDir)
-	exists, err = ctx.fileSystem.DirExists(namespacesPath)
+	// Make sure the top level "resources" dir exists:
+	resourcesDir := filepath.Join(dir, api.ResourcesDir)
+	_, err := ctx.fileSystem.DirExists(resourcesDir)
```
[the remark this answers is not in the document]
dgoodwin: Nice catch.

dgoodwin:
Comments addressed I think!

ncdc:
Thanks again @dgoodwin!!!

PR #139, opened 2017-10-19
Update config-definition.md

lli-hiya:
"kmsKeyID" should be "kmsKeyId"


skriss:
Thanks for the PR @lli-hiya! In order to be able to merge, we need you to add signoff to your commit. You can do this with `git commit --amend --signoff` and then re-push.

PR #145, opened 2017-10-20
give ark cluster-admin privileges

### LINKED ISSUE #73: Consider broadening ark service account permissions to cluster admin
The service account we use is currently allowed to list/watch/create all resources in all namespaces. While this is sufficient for all backup operations, it is only partially so for restore operations.

While most restore operations will succeed, attempting to restore roles and/or clusterroles that have greater privileges than the ark service account fail.

We might want to broaden the privileges to cluster-admin. WDYT @mattmoyer @jbeda @skriss?

skriss:
Fixes #73 

### DIFF
--- a/examples/common/00-prereqs.yaml
+++ b/examples/common/00-prereqs.yaml
@@ -104,30 +104,6 @@ metadata:
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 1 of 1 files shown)


ncdc:
Tested locally & it got rid of the escalation errors.

PR #147, opened 2017-10-20
Add --include-cluster-resources flag for restores

### LINKED ISSUE #136: implement --include-cluster-resources for restores
#128 implemented this flag for backups; this is useful for restores also. Implement.

skriss:
Fixes #136

### DIFF
--- a/pkg/restore/restore_test.go
+++ b/pkg/restore/restore_test.go
@@ -159,10 +159,15 @@ func TestRestoreNamespaceFiltering(t *testing.T) {
 			},
 		},
 		{
-			name:             "namespacesToRestore properly filters with inclusion & exclusion filters",
-			fileSystem:       newFakeFileSystem().WithDirectories("bak/resources/nodes/cluster", "bak/resources/secrets/namespaces/a", "bak/resources/secrets/namespaces/b", "bak/resources/secrets/namespaces/c"),
-			baseDir:          "bak",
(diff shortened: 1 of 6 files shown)


pkg/restore/restore_test.go
```diff
@@ -394,6 +401,24 @@ func TestRestoreResourceForNamespace(t *testing.T) {
 			restorers:     map[schema.GroupResource]restorers.ResourceRestorer{schema.GroupResource{Resource: "foo-resource"}: newFakeCustomRestorer()},
 			expectedObjs:  toUnstructured(newTestConfigMap().WithArkLabel("my-restore").ConfigMap),
 		},
+		// NEW
```
ncdc: Remove

ncdc:
@skriss I was just thinking that if you're creating a `Restore` by hand in yaml/json and you omit `includeClusterResources`, it will default to false and they won't be included. Do you think it would make more sense to change this to `excludeClusterResources` so the default behavior is to include them?

ncdc:
Or make it an optional bool (pointer)

skriss:
Yeah, I think defaulting to include makes sense. I think I like using an optional bool that defaults to true if not specified better; it's more consistent with the corresponding flag on backups.

ncdc:
Yeah, let's go with the optional bool that defaults to true.

PR #152, opened 2017-10-25
Fix #137. Remove clock from volume_snapshot_action

### LINKED ISSUE #137: remove clock from volume_snapshot_action
I'm 99% sure https://github.com/heptio/ark/blob/master/pkg/backup/volume_snapshot_action.go#L38 is obsolete and can be removed. Double-check that this is right, and fully remove it.

athampy:


ncdc:
@athampy thanks for helping out! Could you please do a couple of things?

1. Sign your commit (see https://github.com/heptio/ark/blob/master/CONTRIBUTING.md)
2. Remove the merge commit so that this pull request only has a single commit it in

Thanks!

PR #153, opened 2017-10-26
Add 'make release'

ncdc:
Add 'make release' that automatically creates tarballs for all CLI
platforms in _output/release. It also creates a CHECKSUM file using
sha256 sums. Finally, it displays the contents of the CHECKSUM file as
well as the sha256sum of the CHECKSUM file.


skriss:
LGTM

PR #157, opened 2017-10-26
Upgrade to client-go v5.0.1, kubernetes v1.8.2

ncdc:
Also switch to using k8s.io/code-generator


pkg/cmd/util/output/output.go
```diff
@@ -168,7 +168,7 @@ func NewPrinter(cmd *cobra.Command) (*printers.HumanReadablePrinter, error) {
 
 	printer := printers.NewHumanReadablePrinter(
 		encoder,
-		scheme.Codecs.UniversalDecoder(api.SchemeGroupVersion),
+		scheme.Codecs.UniversalDecoder(v1.SchemeGroupVersion),
```
skriss: was this another instance where we were using the wrong SchemeGroupVersion? Since this is switching from core/v1 to ark/v1

ncdc: Yeah, I'm wondering if it matters...

ncdc: Ok I just read the `HumanReadablePrinter` code. If we specify encoder & decoder, and if the object is `runtime.Unstructured` or `*runtime.Unknown`, it attempts to encode to bytes and then immediately decode it using the decoder to an actual type. Given that we are printing our types using our real client, I think we can pass nil for encoder & decoder. WDYT?

ncdc: Pushed a new commit removing them

skriss:
@ncdc this looks good to me - ready to merge?

ncdc:
Yes

PR #164, opened 2017-10-30
Add OpenStack support

jamiehannaford:
This PR adds the OpenStack cloud provider, using Swift for object storage and Cinder for block storage. I've tested this locally using minikube and things seem to work.

I added the `gophercloud/gophercloud` imports first then ran `dep ensure -add github.com/gophercloud/gophercloud@master` and it imported like ~900 files. When I look at the AWS and Azure deps, however, only the files that are used seem to be there. Did I run the incorrect command or is that the way dep works? I'd like to keep this PR as lean as possible.


ncdc:
@jamiehannaford thanks for the PR! We have a script, `hack/dep-save.sh`, that automatically calls ensure and then does some pruning. So you'd want to do `hack/dep-save.sh -add github.com/gophercloud/gophercloud@master` and it should take care of things. But now that you've vendored already, it should be sufficient to just run `hack/dep-save.sh` with no arguments to get it to prune everything for you. Does that make sense?

jamiehannaford:
@ncdc Yep! Great script btw. That's reduced it down to only 70 files :metal:

ncdc:
Kicking sign-off checker & travis

ncdc:
@jamiehannaford we are currently working on a design for moving cloud provider integrations to plugins. I think it will be easier if we do the plugin conversion first, and then look at an OpenStack plugin. Hope you don't mind the delay! Also, if you've got time & interest, we'll be sharing the plugin design doc soon and would love a review + comments.

jamiehannaford:
Sure, SGTM. Feel free to ping when you have a design doc :)

ncdc:
@jamiehannaford https://docs.google.com/document/d/112y6Z4qqeMgVguPtfBYva_3qhh0oBfkLkwkQLerOcbI/edit#

jamiehannaford:
After discussion with @ncdc I'm closing this and will target a new PR in another location

nrb:
@jamiehannaford Wanted to let you know that [v0.6.0](https://github.com/heptio/ark/tree/v0.6.0) was released today, and we also put up an [example repo](https://github.com/heptio/ark-plugin-example/)

PR #173, opened 2017-11-01
Move restore status warnings/errors to object storage; add restore describe command

### LINKED ISSUE #108: Restore can complete but fail writing to etcd with request too large
I created a backup for a specific namespace, and restored it into a rebuilt cluster. My particular cluster may contain a larger than normal amount of cluster objects. (cluster role bindings, etc) I saw quite a bit of throttling request messages in the log. 
[ark-restore.log](https://github.com/heptio/ark/files/1344456/ark-restore.log)

The restore completed, my namespace is restored (though failed with service account problems probably due to the new cluster, I will dig into that next), but fails to write status because etcd finds the request too large. My restore then is hung in progress.

```
I0929 12:56:11.805636       1 restore_controller.go:255] restore heptio-ark/test1-11-20170929125444 completed      
I0929 12:56:11.805661       1 restore_controller.go:258] updating restore heptio-ark/test1-11-20170929125444 final status                                                                                                              
I0929 12:56:11.991224       1 restore_controller.go:260] error updating restore heptio-ark/test1-11-20170929125444 final status: etcdserver: request is too large    
```

```
$ ark restore get
NAME                      BACKUP     STATUS       WARNINGS   ERRORS    CREATED                         SELECTOR
test1-11-20170929125444   test1-11   InProgress   0          0         2017-09-29 12:54:45 +0000 UTC   <none>
```

Environment was OpenShift 3.7 based on Kubernetes 1.7. Etcd 3.2.5.

ncdc:
etcd has a hard limit for a value (either 1 or 1.5MB, can't remember which). If there are a lot of warnings and/or errors during the restore, it's possible that the entire `Restore` object can exceed that maximum.

This PR changes `restore.status.errors` and `restore.status.warnings` to `int`s (just a counter) and moves the actual details into a separate file in object storage. There is a new `ark restore describe` command that retrieves the restore and then the file from object storage, merges the information together, and prints it in a human-readable format.

Fixes #108 

### DIFF
--- a/pkg/cloudprovider/backup_service.go
+++ b/pkg/cloudprovider/backup_service.go
@@ -59,6 +59,9 @@ type BackupService interface {
 
 	// UploadRestoreLog uploads the restore's log file to object storage.
 	UploadRestoreLog(bucket, backup, restore string, log io.Reader) error
+
+	// UploadRestoreResults uploads the restore's results file to object storage.
+	UploadRestoreResults(bucket, backup, restore string, results io.Reader) error
 }
 
 // BackupGetter knows how to list backups in object storage.
@@ -68,10 +71,11 @@ type BackupGetter interface {
 }
 
 const (
-	metadataFileFormatString   = "%s/ark-backup.json"
-	backupFileFormatString     = "%s/%s.tar.gz"
-	backupLogFileFormatString  = "%s/%s-logs.gz"
-	restoreLogFileFormatString = "%s/restore-%s-logs.gz"
+	metadataFileFormatString       = "%s/ark-backup.json"
+	backupFileFormatString         = "%s/%s.tar.gz"
+	backupLogFileFormatString      = "%s/%s-logs.gz"
+	restoreLogFileFormatString     = "%s/restore-%s-logs.gz"
+	restoreResultsFileFormatString = "%s/restore-%s-results.gz"
 )
 
 func getMetadataKey(backup string) string {
@@ -88,6 +92,10 @@ func getBackupLogKey(backup string) string {
 
 func getRestoreLogKey(backup, restore string) string {
 	return fmt.Sprintf(restoreLogFileFormatString, backup, restore)
+}
+
+func getRestoreResultsKey(backup, restore string) string {
+	return fmt.Sprintf(restoreResultsFileFormatString, backup, restore)
 }
 
 type backupService struct {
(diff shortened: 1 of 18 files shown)


ncdc:
Note, I haven't fixed the broken tests yet. Looking for a 👍  on the impl before I do that. @skriss @nrb @dgoodwin @jrnt30

nrb:
This seems reasonable to me. I like keeping the detailed info around and having a structured way of viewing it. Also +1 on using `describe` semantics.

pkg/cmd/cli/restore/describe.go
```diff
@@ -17,18 +17,181 @@ limitations under the License.
...
+	if len(restore.Status.ValidationErrors) == 0 {
+		fmt.Fprintf(out, "\t<none>\n")
+	} else {
+		for _, ve := range restore.Status.ValidationErrors {
+			fmt.Fprintf(out, "\t%s\n", ve)
+		}
+	}
+
+	fmt.Fprintln(out)
+	describeRestoreResults(out, arkClient, restore.Name)
```
skriss: maybe optimize the case where there are 0 errors/warnings?

skriss:
Impl LGTM. I just submitted #174 which switches PutObject to an io.Reader, so one of us will need to rebase.

pkg/cmd/cli/restore/describe.go
ncdc: Good idea, I'll do that tomorrow.

ncdc:
Tests should be good to go now.

ncdc:
Rebased on top of your readseeker change

ncdc:
@skriss ready for another review

skriss:
Looking now

pkg/cloudprovider/backup_service.go
```diff
@@ -219,23 +227,35 @@ func (br *backupService) CreateSignedURL(target api.DownloadTarget, bucket strin
...
 	case api.DownloadTargetKindRestoreLog:
-		// restore name is formatted as <backup name>-<timestamp>
-		i := strings.LastIndex(target.Name, "-")
-		if i < 0 {
-			i = len(target.Name)
-		}
-		backup := target.Name[0:i]
+		backup := extractBackupName(target.Name)
 		return br.objectStorage.CreateSignedURL(bucket, getRestoreLogKey(backup, target.Name), ttl)
+	case api.DownloadTargetKindRestoreResults:
```
skriss: want to add a unit case?

ncdc: Knew I had forgotten something

ncdc: Added

pkg/cmd/cli/restore/describe.go
```diff
@@ -17,18 +17,186 @@ limitations under the License.
...
+	fmt.Fprintln(out)
+	s = "<none>"
+	if restore.Spec.LabelSelector != nil {
+		s = metav1.FormatLabelSelector(restore.Spec.LabelSelector)
+	}
+	fmt.Fprintf(out, "Label selector:\t%s\n", s)
+
+	fmt.Fprintln(out)
+	b := true
+	if restore.Spec.RestorePVs != nil {
```
skriss: probably best to do auto/yes/no printing here like with IncludeClusterResources

skriss:
one last comment, otherwise LGTM

pkg/cmd/cli/restore/describe.go
ncdc: fixing

skriss:
LGTM.

PR #174, opened 2017-11-01
Switch PutObject to take an io.Reader rather than an io.ReadSeeker

skriss:
This simplifies the interface requirement. For AWS, the S3 Uploader is used to enable this; for Azure, remove the code that was setting ContentLength (this is now done within the Azure client library version that we're on).


skriss:
@ncdc @nrb PTAL when you get a chance. This is pre-work for plugins.

ncdc:
LGTM. @nrb ?

skriss:
@nrb this is partly tech debt (we were never happy about needing a ReadSeeker, because it's more restrictive than just a Reader, but the cloud APIs essentially required it), and partly prep for plugins (it's much simpler to deal with just the io.Reader interface over gRPC).

PR #175, opened 2017-11-02
WIP: Fix missing namespace resource when backing up namespaces with auto cluster resources

dgoodwin:

### DIFF
--- a/pkg/backup/resource_backupper.go
+++ b/pkg/backup/resource_backupper.go
@@ -129,7 +129,9 @@ func (rb *defaultResourceBackupper) backupResource(
 		// (all namespaces) backup. Note that in the case of a subset of
 		// namespaces being backed up, some related cluster-scoped resources
 		// may still be backed up if triggered by a custom action (e.g. PVC->PV).
-		if !resource.Namespaced && !rb.namespaces.IncludeEverything() {
+		// If we're processing namespaces themselves, we will not skip here, they may be
+		// filtered out later.
+		if !resource.Namespaced && resource.Kind != "Namespace" && !rb.namespaces.IncludeEverything() {
 			log.Info("Skipping resource because it's cluster-scoped and only specific namespaces are included in the backup")
 			return nil
 		}
@@ -173,6 +175,8 @@ func (rb *defaultResourceBackupper) backupResource(
 		rb.discoveryHelper,
 	)
 
+	// TODO: when processing namespaces, and only including certain namespaces, we still list
+	// them all here. Could optimize to get specifics, but watch out for label selector.
 	var namespacesToList []string
 	if resource.Namespaced {
 		namespacesToList = getNamespacesToList(rb.namespaces)
--- a/pkg/backup/backup_test.go
+++ b/pkg/backup/backup_test.go
@@ -230,7 +230,7 @@ func TestGetNamespaceIncludesExcludes(t *testing.T) {
 var (
 	v1Group = &metav1.APIResourceList{
 		GroupVersion: "v1",
-		APIResources: []metav1.APIResource{configMapsResource, podsResource},
+		APIResources: []metav1.APIResource{configMapsResource, podsResource, namespacesResource},
 	}
 
 	configMapsResource = metav1.APIResource{
@@ -263,6 +263,14 @@ var (
 		SingularName: "role",
 		Namespaced:   true,
 		Kind:         "Role",
+		Verbs:        metav1.Verbs([]string{"create", "update", "get", "list", "watch", "delete"}),
+	}
+
+	namespacesResource = metav1.APIResource{
+		Name:         "namespaces",
+		SingularName: "namespace",
+		Namespaced:   false,
+		Kind:         "Namespace",
 		Verbs:        metav1.Verbs([]string{"create", "update", "get", "list", "watch", "delete"}),
 	}
 
--- a/pkg/backup/resource_backupper_test.go
+++ b/pkg/backup/resource_backupper_test.go
@@ -195,6 +195,24 @@ func TestBackupResource(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:                     "should include specified namespaces if backing up subset of namespaces and --include-cluster-resources=nil",
+			namespaces:               collections.NewIncludesExcludes().Includes("ns-1"),
+			resources:                collections.NewIncludesExcludes(),
+			includeClusterResources:  nil,
+			expectedListedNamespaces: []string{"ns-1"},
+			apiGroup:                 v1Group,
+			apiResource:              namespacesResource,
+			groupVersion:             schema.GroupVersion{Group: "", Version: "v1"},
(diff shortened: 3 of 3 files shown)


ncdc:
Not quite @nrb. @dgoodwin is working through one issue.

pkg/backup/resource_backupper.go
```diff
@@ -174,7 +176,7 @@ func (rb *defaultResourceBackupper) backupResource(
 	)
 
 	var namespacesToList []string
-	if resource.Namespaced {
+	if resource.Namespaced || resource.Kind == "Namespace" {
```
ncdc: As we discussed on Slack, this doesn't do what you think it might. If you call `rb.dynamicFactory.ClientForGroupVersionResource(gv, resource, namespace)` where `resource` is the namespaces resource, and `namespace` is a specific name of a namespace, the call that immediately follows:

```go
unstructuredList, err := resourceClient.List(metav1.ListOptions{LabelSelector: rb.labelSelector})
```

will still return a full listing of all of the namespaces.

dgoodwin:
Ok updated with passing unit tests, and manual testing, this does fix the issue I hit.

pkg/backup/resource_backupper.go
```diff
@@ -174,37 +176,57 @@ func (rb *defaultResourceBackupper) backupResource(
...
-	if resource.Namespaced {
-		namespacesToList = getNamespacesToList(rb.namespaces)
-	} else {
-		namespacesToList = []string{""}
-	}
-	for _, namespace := range namespacesToList {
-		resourceClient, err := rb.dynamicFactory.ClientForGroupVersionResource(gv, resource, namespace)
+	// If processing the Namespace resource specifically and you've specified to include only some,
+	// fetch them individually instead of a full list of all in the cluster:
+	if resource.Kind == "Namespace" && !rb.namespaces.IncludeEverything() {
```
ncdc: The `!rb.namespaces.IncludeEverything()` check isn't going to work here. This expression will evaluate to true if you have any excludes, even if you don't have any specific includes, for example. You're better off checking if length of `GetIncludes()` >= 1 and it doesn't contain `*`. Or we could create another helper `HasSpecificIncludes` or something like that.

dgoodwin: Updated with another pass at that, I see validation that if * is in the list it must be the only element in the list, so I hope I got it right this time.

pkg/backup/resource_backupper.go
```diff
@@ -174,37 +176,59 @@ func (rb *defaultResourceBackupper) backupResource(
...
-	if resource.Namespaced {
-		namespacesToList = getNamespacesToList(rb.namespaces)
-	} else {
-		namespacesToList = []string{""}
-	}
-	for _, namespace := range namespacesToList {
-		resourceClient, err := rb.dynamicFactory.ClientForGroupVersionResource(gv, resource, namespace)
+	// If processing the Namespace resource specifically and you've specified to include only some,
+	// fetch them individually instead of a full list of all in the cluster:
+	if resource.Kind == "Namespace" &&
```
ncdc: Probably should also make sure that gv.Group == ""

pkg/backup/resource_backupper.go
```diff
@@ -174,37 +176,59 @@ func (rb *defaultResourceBackupper) backupResource(
...
+		!(len(rb.namespaces.GetIncludes()) == 0 || rb.namespaces.GetIncludes()[0] == "*") {
```
ncdc: This could be written in the positive as `(len(rb.namespaces.GetIncludes()) >= 1 && rb.namespaces.GetIncludes()[0] != "*")`. Maybe a bit easier to read?

pkg/backup/resource_backupper.go
```diff
@@ -174,37 +176,59 @@ func (rb *defaultResourceBackupper) backupResource(
...
+
+		resourceClient, err := rb.dynamicFactory.ClientForGroupVersionResource(gv, resource, "")
 		if err != nil {
 			return err
 		}
+		for _, namespace := range rb.namespaces.GetIncludes() {
+			// TODO: labelSelector relevant here?
```
ncdc: It is relevant, but GetOptions doesn't support it. You'd have to get a selector object and call Matches on it by hand.

Maybe we should just ignore this optimization for the time being?

dgoodwin:
Updated to skip the optimization for now as it looked more complicated than expected, left a todo. Looking ok so far in my testing.

ncdc:
LGTM. WDYT @skriss? I'd probably eventually like a unit test that doesn't mock out the itemBackupper completely so we can make sure that when you are only including certain namespaces, the itemBackupper filters out a namespace that doesn't match.

dgoodwin:
You may already have such a test: https://github.com/heptio/ark/blob/master/pkg/backup/item_backupper_test.go#L51

ncdc:
Yes, but I want one that ensures that when the resourceBackupper is listing namespaces and sending them one by one to the itemBackupper, the itemBackupper filters out ones that are in the includes list.

skriss:
I'm probably missing something, but isn't this going to end up backing up all namespaces? since namespaces themselves are cluster-scoped, itemBackupper isn't going to filter them out based on namespace includes/excludes..

pkg/backup/resource_backupper.go
```diff
@@ -129,7 +129,9 @@ func (rb *defaultResourceBackupper) backupResource(
 		// (all namespaces) backup. Note that in the case of a subset of
 		// namespaces being backed up, some related cluster-scoped resources
 		// may still be backed up if triggered by a custom action (e.g. PVC->PV).
-		if !resource.Namespaced && !rb.namespaces.IncludeEverything() {
+		// If we're processing namespaces themselves, we will not skip here, they may be
+		// filtered out later.
+		if !resource.Namespaced && resource.Kind != "Namespace" && !rb.namespaces.IncludeEverything() {
```
nrb: This will skip if the resource is a namespace, but doesn't appear to filter based on included/excluded namespaces. Is that intentional?

ncdc:
@skriss you're right. Back to the drawing board.

skriss:
@ncdc see my updated comment ^^^ - does that seem reasonable?

TLDR namespace API objects are kind of a pain.

pkg/backup/resource_backupper.go
ncdc: It was, until we realized that the itemBackupper's filtering of namespaces based on includes/excludes only applies to namespace-scoped resources, but it won't work for namespaces (they're cluster-scoped).

ncdc:
Could work

pkg/backup/resource_backupper.go
nrb: Right, I saw that code earlier. Is there some sort of cluster-scoped filtering that could be introduced?

ncdc: This is only messy because it's the namespace object itself

ncdc:
Replaced by #182

Issue #176, opened 2017-11-02
Namespace mappings is broken


ncdc:
We're passing the wrong namespace name [here](https://github.com/heptio/ark/blob/c49d11f17a66c12117c2a621b9896c076e682e24/pkg/restore/restore.go#L354).

PR #179, opened 2017-11-02
fix restore namespace remapping bug

skriss:

Fixes: #176


pkg/restore/restore_test.go
```diff
@@ -292,6 +292,53 @@ func TestRestorePriority(t *testing.T) {
...
+	for i := range expectedObjs {
+		resourceClient.On("Create", &expectedObjs[i]).Return(&expectedObjs[i], nil)
+	}
+
+	dynamicFactory := &FakeDynamicFactory{}
+	resource := metav1.APIResource{Name: "configmaps", Namespaced: true}
+	gv := schema.GroupVersion{Group: "", Version: "v1"}
+	dynamicFactory.On("ClientForGroupVersionResource", gv, resource, expectedNS).Return(resourceClient, nil)
+
+	log, _ := testlogger.NewNullLogger()
```
ncdc: `arktest.NewLogger()`, move it down to line 326 directly

skriss: context.logger is a *logrus.Logger, not a *logrus.LogEntry -- you want me to change it to a FieldLogger as part of this PR?

ncdc:
Oh, never mind for now.

skriss:
OK - good change to make in general, that interface covers both the Logger and the LogEntry, but we can circle back.

PR #182, opened 2017-11-03
Fix missing namespace resource when backing up namespaces with auto cluster resources

ncdc:
This fixes a few issues:
- Make sure we include namespaces in the backup, regardless of the value of backup.spec.includeClusterResources
- Make sure we exclude the actual namespace objects specified in backup.spec.excludedNamespaces from the backup

Replaces #175


skriss:
LGTM..

Issue #167, opened 2017-10-31
[Feature] Support storing Ark backups within a sub-dir of object storage buckets


skriss:
Right now Ark only supports storing backups within the root of an object storage bucket. It would be nice to support storing them within sub-dirs as well.

ncdc:
Moving out of v0.7.0 as we'll probably do this as part of the backup target redesign.

rosskukulinski:
I recommend default `subPath` value is `/` which will use the top-level directory of the storage bucket.

We should not assume `subpaths` exist, so we should create the complete path, `mkdir -p`, if it doesn't exist.

rosskukulinski:
@ncdc works for me.  My `mkdir` comment was simply trying to clarify my point about creating subdirectories (if necessary)

PR #185, opened 2017-11-06
Support storing Ark backups within a sub-dir of object storage bucket

mstump:
Fixes issue #167
All tests pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ BIN := ark
 PKG := github.com/heptio/ark
 
 # Where to push the docker image.
-REGISTRY ?= gcr.io/heptio-images
+REGISTRY ?= vorstella
 
 # Which architecture to build - see $(ALL_ARCH) for options.
 ARCH ?= linux-amd64
--- a/pkg/cloudprovider/backup_cache_test.go
+++ b/pkg/cloudprovider/backup_cache_test.go
@@ -45,32 +45,32 @@ func TestNewBackupCache(t *testing.T) {
 		test.NewTestBackup().WithName("backup1").Backup,
 		test.NewTestBackup().WithName("backup2").Backup,
 	}
-	delegate.On("GetAllBackups", "bucket1").Return(bucket1, nil).Once()
+	delegate.On("GetAllBackups", "bucket1", "path1").Return(bucket1, nil).Once()
 
 	// should be updated via refresh
 	updatedBucket1 := []*v1.Backup{
 		test.NewTestBackup().WithName("backup2").Backup,
 	}
-	delegate.On("GetAllBackups", "bucket1").Return(updatedBucket1, nil)
+	delegate.On("GetAllBackups", "bucket1", "path1").Return(updatedBucket1, nil)
 
 	// nothing in cache, live lookup
 	bucket2 := []*v1.Backup{
 		test.NewTestBackup().WithName("backup5").Backup,
 		test.NewTestBackup().WithName("backup6").Backup,
 	}
-	delegate.On("GetAllBackups", "bucket2").Return(bucket2, nil).Once()
+	delegate.On("GetAllBackups", "bucket2", "path2").Return(bucket2, nil).Once()
 
 	// should be updated via refresh
 	updatedBucket2 := []*v1.Backup{
 		test.NewTestBackup().WithName("backup7").Backup,
 	}
-	delegate.On("GetAllBackups", "bucket2").Return(updatedBucket2, nil)
+	delegate.On("GetAllBackups", "bucket2", "path2").Return(updatedBucket2, nil)
 
-	backups, err := c.GetAllBackups("bucket1")
+	backups, err := c.GetAllBackups("bucket1", "path1")
 	assert.Equal(t, bucket1, backups)
 	assert.NoError(t, err)
 
-	backups, err = c.GetAllBackups("bucket2")
+	backups, err = c.GetAllBackups("bucket2", "path2")
 	assert.Equal(t, bucket2, backups)
 	assert.NoError(t, err)
 
@@ -85,14 +85,14 @@ func TestNewBackupCache(t *testing.T) {
 			}
 		}
 
-		backups, err = c.GetAllBackups("bucket1")
+		backups, err = c.GetAllBackups("bucket1", "path1")
 		if len(backups) == 1 {
 			if assert.Equal(t, updatedBucket1[0], backups[0]) {
 				done1 = true
 			}
 		}
 
-		backups, err = c.GetAllBackups("bucket2")
+		backups, err = c.GetAllBackups("bucket2", "path2")
 		if len(backups) == 1 {
 			if assert.Equal(t, updatedBucket2[0], backups[0]) {
 				done2 = true
@@ -121,9 +121,9 @@ func TestBackupCacheRefresh(t *testing.T) {
 		test.NewTestBackup().WithName("backup1").Backup,
 		test.NewTestBackup().WithName("backup2").Backup,
 	}
-	delegate.On("GetAllBackups", "bucket1").Return(bucket1, nil)
+	delegate.On("GetAllBackups", "bucket1", "").Return(bucket1, nil)
 
-	delegate.On("GetAllBackups", "bucket2").Return(nil, errors.New("bad"))
+	delegate.On("GetAllBackups", "bucket2", "").Return(nil, errors.New("bad"))
 
 	c.refresh()
 
@@ -159,13 +159,13 @@ func TestBackupCacheGetAllBackupsUsesCacheIfPresent(t *testing.T) {
 		test.NewTestBackup().WithName("backup4").Backup,
 	}
 
-	delegate.On("GetAllBackups", "bucket2").Return(bucket2, nil)
+	delegate.On("GetAllBackups", "bucket2", "path2").Return(bucket2, nil)
 
-	backups, err := c.GetAllBackups("bucket1")
+	backups, err := c.GetAllBackups("bucket1", "path1")
 	assert.Equal(t, bucket1, backups)
 	assert.NoError(t, err)
 
-	backups, err = c.GetAllBackups("bucket2")
+	backups, err = c.GetAllBackups("bucket2", "path2")
 	assert.Equal(t, bucket2, backups)
 	assert.NoError(t, err)
 }
--- a/pkg/cloudprovider/backup_service.go
+++ b/pkg/cloudprovider/backup_service.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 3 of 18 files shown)


Makefile
```diff
@@ -21,7 +21,7 @@ BIN := ark
 PKG := github.com/heptio/ark
 
 # Where to push the docker image.
-REGISTRY ?= gcr.io/heptio-images
+REGISTRY ?= vorstella
```
nrb: I believe this change will have to be reverted; others don't have access to the Vorstella registry :)

pkg/cloudprovider/backup_cache_test.go
```diff
@@ -45,32 +45,32 @@ func TestNewBackupCache(t *testing.T) {
...
-	delegate.On("GetAllBackups", "bucket1").Return(updatedBucket1, nil)
+	delegate.On("GetAllBackups", "bucket1", "path1").Return(updatedBucket1, nil)
 
 	// nothing in cache, live lookup
 	bucket2 := []*v1.Backup{
 		test.NewTestBackup().WithName("backup5").Backup,
 		test.NewTestBackup().WithName("backup6").Backup,
 	}
-	delegate.On("GetAllBackups", "bucket2").Return(bucket2, nil).Once()
+	delegate.On("GetAllBackups", "bucket2", "path2").Return(bucket2, nil).Once()
```
ncdc: All of the paths should be identical for a given test or set of tests. Right now Ark only supports a single configuration, so it doesn't make sense to vary the path.

pkg/cloudprovider/backup_service.go
```diff
@@ -40,54 +41,59 @@ type BackupService interface {
...
-	GetAllBackups(bucket string) ([]*api.Backup, error)
+	GetAllBackups(bucket, bucketPath string) ([]*api.Backup, error)
 }
 
 const (
 	metadataFileFormatString   = "%s/ark-backup.json"
 	backupFileFormatString     = "%s/%s.tar.gz"
 	backupLogFileFormatString  = "%s/%s-logs.gz"
 	restoreLogFileFormatString = "%s/restore-%s-logs.gz"
+	pathFormatString           = "%s/%s"
```
ncdc: Unused?

pkg/cloudprovider/backup_service.go
```diff
@@ -137,12 +143,12 @@ func (br *backupService) UploadBackup(bucket, backupName string, metadata, backu
...
-func (br *backupService) DownloadBackup(bucket, backupName string) (io.ReadCloser, error) {
-	return br.objectStorage.GetObject(bucket, getBackupContentsKey(backupName))
+func (br *backupService) DownloadBackup(bucket, bucketPath, backupName string) (io.ReadCloser, error) {
+	return br.objectStorage.GetObject(bucket, getBackupContentsKey(bucketPath, backupName))
 }
 
-func (br *backupService) GetAllBackups(bucket string) ([]*api.Backup, error) {
-	prefixes, err := br.objectStorage.ListCommonPrefixes(bucket, "/")
+func (br *backupService) GetAllBackups(bucket string, bucketPath string) ([]*api.Backup, error) {
+	prefixes, err := br.objectStorage.ListCommonPrefixes(bucket, joinPathPrefix(bucketPath, "/"))
```
ncdc: If `bucketPath` is not blank, this ends up listing e.g. `"foo/bar"`. Note there is no trailing `/`. Is that ok? I'm thinking we need the trailer? cc @skriss

Makefile
mstump: oops, will correct.

pkg/cloudprovider/backup_service.go
```diff
@@ -40,54 +41,59 @@ type BackupService interface {
...
 	metadataFileFormatString   = "%s/ark-backup.json"
 	backupFileFormatString     = "%s/%s.tar.gz"
 	backupLogFileFormatString  = "%s/%s-logs.gz"
 	restoreLogFileFormatString = "%s/restore-%s-logs.gz"
+	pathFormatString           = "%s/%s"
 )
 
-func getMetadataKey(backup string) string {
-	return fmt.Sprintf(metadataFileFormatString, backup)
+func joinPathPrefix(prefix string, suffix string) string {
```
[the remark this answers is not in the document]
mstump: it is, but it does make it clear what we're doing and makes it easy to make modifications... fine either way.

Makefile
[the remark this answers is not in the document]
mstump: Fixed

pkg/cloudprovider/backup_service.go
mstump: removed

pkg/cloudprovider/backup_service.go
mstump: fixed

mstump:
All requested changes with the exception of joinPathPrefix path have been incorporated. Want that done too?

pkg/cloudprovider/aws/object_storage_adapter.go
```diff
@@ -35,9 +36,10 @@ type objectStorageAdapter struct {
 	s3         *s3.S3
 	s3Uploader *s3manager.Uploader
 	kmsKeyID   string
+	logger     *logrus.Logger
 }
 
-func NewObjectStorageAdapter(region, s3URL, kmsKeyID string, s3ForcePathStyle bool) (cloudprovider.ObjectStorageAdapter, error) {
+func NewObjectStorageAdapter(region, s3URL, kmsKeyID string, s3ForcePathStyle bool, logger *logrus.Logger) (cloudprovider.ObjectStorageAdapter, error) {
```
ncdc: Just a note: we'll have a new logging mechanism once this becomes a plugin

pkg/cloudprovider/aws/object_storage_adapter.go
```diff
@@ -69,10 +71,12 @@ func NewObjectStorageAdapter(region, s3URL, kmsKeyID string, s3ForcePathStyle bo
 		s3:         s3.New(sess),
 		s3Uploader: s3manager.NewUploader(sess),
 		kmsKeyID:   kmsKeyID,
+		logger:     logger,
 	}, nil
 }
 
 func (op *objectStorageAdapter) PutObject(bucket string, key string, body io.Reader) error {
+	op.logger.Infof("put s3 object s3://%s/%s", bucket, key)
```
ncdc: WDYT about doing bucket and key as structured fields?

pkg/cloudprovider/aws/object_storage_adapter.go
```diff
@@ -104,12 +109,17 @@ func (op *objectStorageAdapter) GetObject(bucket string, key string) (io.ReadClo
...
 
-func (op *objectStorageAdapter) ListCommonPrefixes(bucket string, delimiter string) ([]string, error) {
+func (op *objectStorageAdapter) ListCommonPrefixes(bucket string, delimiter string, prefix string) ([]string, error) {
+	op.logger.Infof("list s3 common prefix s3://%s/%s", bucket, prefix)
 	req := &s3.ListObjectsV2Input{
 		Bucket:    &bucket,
 		Delimiter: &delimiter,
 	}
 
+	if prefix != "" {
```
ncdc: What happens if we always specify `&prefix` regardless of its value? Does it still work?

pkg/cloudprovider/aws/object_storage_adapter.go
```diff
@@ -158,6 +170,7 @@ func (op *objectStorageAdapter) DeleteObject(bucket string, key string) error {
 }
 
 func (op *objectStorageAdapter) CreateSignedURL(bucket, key string, ttl time.Duration) (string, error) {
+	op.logger.Debug("create signed s3 url s3://%s/%s %s ttl", bucket, key, ttl)
```
ncdc: I'm curious what made you decide to make this debug vs info like the others?

pkg/cloudprovider/azure/object_storage_adapter.go
```diff
@@ -93,6 +93,10 @@ func (op *objectStorageAdapter) ListCommonPrefixes(bucket string, delimiter stri
 		Delimiter: delimiter,
 	}
 
+	if prefix != "" {
```
ncdc: What happens if you pass this even when it's `""`?

pkg/cloudprovider/backup_service.go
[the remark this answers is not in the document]
ncdc: I'd probably take it out, but I won't block the PR if you don't 😄

pkg/cloudprovider/gcp/object_storage_adapter.go
```diff
@@ -76,8 +76,14 @@ func (op *objectStorageAdapter) GetObject(bucket string, key string) (io.ReadClo
 	return res.Body, nil
 }
 
-func (op *objectStorageAdapter) ListCommonPrefixes(bucket string, delimiter string) ([]string, error) {
-	res, err := op.gcs.Objects.List(bucket).Delimiter(delimiter).Do()
+func (op *objectStorageAdapter) ListCommonPrefixes(bucket string, delimiter string, prefix string) ([]string, error) {
+	req := op.gcs.Objects.List(bucket).Delimiter(delimiter)
+
+	if prefix != "" {
```
ncdc: Same question as above - can we call this all the time?

pkg/controller/backup_sync_controller.go
```diff
@@ -69,8 +72,8 @@ func (c *backupSyncController) Run(ctx context.Context, workers int) error {
 }
 
 func (c *backupSyncController) run() {
-	c.logger.Info("Syncing backups from object storage")
-	backups, err := c.backupService.GetAllBackups(c.bucket)
+	c.logger.Infof("Syncing backups from object storage %s/%s", c.bucket, c.path)
```
ncdc: I'd probably do bucket and path as fields

pkg/cloudprovider/backup_service.go
```diff
@@ -161,7 +166,7 @@ func (br *backupService) GetAllBackups(bucket string) ([]*api.Backup, error) {
 	output := make([]*api.Backup, 0, len(prefixes))
 
 	for _, backupDir := range prefixes {
-		backup, err := br.GetBackup(bucket, backupDir)
+		backup, err := br.GetBackup(bucket, bucketPath, backupDir)
```
skriss: the prefixes that come back from ListCommonPrefixes include the path. so when we call GetBackup here we end up double-prefixing with the path. Probably makes sense to strip the path off of `backupDir` before passing it to GetBackup

pkg/cloudprovider/backup_service.go
```diff
@@ -145,12 +150,12 @@ func (br *backupService) UploadBackup(bucket, backupName string, metadata, backu
...
-func (br *backupService) DownloadBackup(bucket, backupName string) (io.ReadCloser, error) {
-	return br.objectStorage.GetObject(bucket, getBackupContentsKey(backupName))
+func (br *backupService) DownloadBackup(bucket, bucketPath, backupName string) (io.ReadCloser, error) {
+	return br.objectStorage.GetObject(bucket, getBackupContentsKey(bucketPath, backupName))
 }
 
-func (br *backupService) GetAllBackups(bucket string) ([]*api.Backup, error) {
-	prefixes, err := br.objectStorage.ListCommonPrefixes(bucket, "/")
+func (br *backupService) GetAllBackups(bucket string, bucketPath string) ([]*api.Backup, error) {
+	prefixes, err := br.objectStorage.ListCommonPrefixes(bucket, "/", bucketPath)
```
skriss: `bucketPath` needs to have a trailing `/`here if it's non-empty, otherwise ListCommonPrefixes ends up returning the `bucketPath` value rather than the backups contained within it.

pkg/cloudprovider/aws/object_storage_adapter.go
skriss: I think these make sense to all be at debug level -WDYT?

ncdc:
@mstump just a friendly checkin - will you have time to update this in the next few days? Thanks!

pkg/cloudprovider/aws/object_storage_adapter.go
mstump: yeah, as long it's a valid path value. If it's not specified ("") then you get the old behavior anything else it will try to join the path and will use it.

pkg/cloudprovider/azure/object_storage_adapter.go
mstump: same as above

pkg/cloudprovider/gcp/object_storage_adapter.go
mstump: Don't really understand.

nrb:
Thanks for the updates @mstump! Plugin support has since merged into master, so it looks like this will need a rebase.

mstump:
ok, feedback incorporated and rebased with master

pkg/cloudprovider/backup_service.go
```diff
@@ -145,12 +150,13 @@ func (br *backupService) UploadBackup(bucket, backupName string, metadata, backu
...
-func (br *backupService) DownloadBackup(bucket, backupName string) (io.ReadCloser, error) {
-	return br.objectStore.GetObject(bucket, getBackupContentsKey(backupName))
+func (br *backupService) DownloadBackup(bucket, bucketPath, backupName string) (io.ReadCloser, error) {
+	return br.objectStore.GetObject(bucket, getBackupContentsKey(bucketPath, backupName))
 }
 
-func (br *backupService) GetAllBackups(bucket string) ([]*api.Backup, error) {
-	prefixes, err := br.objectStore.ListCommonPrefixes(bucket, "/")
+func (br *backupService) GetAllBackups(bucket string, bucketPath string) ([]*api.Backup, error) {
+	prefixes, err := br.objectStore.ListCommonPrefixes(bucket, "/", path.Clean(bucketPath)+"/")
```
skriss: I had to modify this to the following to make it work when not specifying a path:

```
var prefix string
	if bucketPath != "" {
		prefix = path.Clean(bucketPath) + "/"
	}

	prefixes, err := br.objectStore.ListCommonPrefixes(bucket, "/", prefix)
```

pkg/cloudprovider/aws/object_store.go
```diff
@@ -128,12 +128,16 @@ func (o *objectStore) GetObject(bucket string, key string) (io.ReadCloser, error
...
 		Delimiter: &delimiter,
 	}
 
+	if prefix != "" {
+		req.Prefix = &prefix
+	}
+
 	var ret []string
 	err := o.s3.ListObjectsV2Pages(req, func(page *s3.ListObjectsV2Output, lastPage bool) bool {
 		for _, prefix := range page.CommonPrefixes {
```
skriss: I also had to modify the line below:

```
val := *prefix.Prefix
ret = append(ret, val[0:strings.LastIndex(val, delimiter)])
```

It appears that AWS is forgiving in some cases if you have a `//` (i.e. our current scheme) and not forgiving in others (i.e. with your change).

skriss:
hi @mstump - friendly reminder that I added some comments to this PR last week.

ncdc:
Removing from v0.7.0 for now as this will likely be supported as part of the backup targets design

PR #186, opened 2017-11-06
Include git describe info in binary names.

ncdc:
New format now looks like this:
ark-v0.5.0-44-g72b0bdd-dirty-linux-amd64.tar.gz


skriss:
LGTM.

Issue #135, opened 2017-10-17
Convert to plugins


ncdc:
We should convert the following interfaces to plugins:
- backup actions
- restore actions (restorers)
- object storage provider
- persistent volume provider

This will allow integrators to provide custom logic without having to recompile Ark.

We need to make it easy for the Ark pod to be able to locate "sidecar" plugins, e.g. maybe something like:
1. Define a volume + volumeMount
1. Define an init container using your image that contains your plugins
1. Init container copies plugins from its image/filesystem to the volumeMount from above
1. Ark container has the same volumeMount and is configured to look for plugins there

There's hopefully an easier way, but this is one I could come up with quickly.

ncdc:
Design doc: https://docs.google.com/document/d/112y6Z4qqeMgVguPtfBYva_3qhh0oBfkLkwkQLerOcbI/edit#

ncdc:
This is mostly done. #214 should be the last piece.

PR #188, opened 2017-11-07
Convert existing cloudproviders to plugins

skriss:
This is a breaking change and you will have to edit your `config/default` in order for Ark to continue working. Please see the changes to the config file format in this PR for a description of what changed.

Part of the fix for #135, but we still need to convert backup and restore actions.


pkg/plugin/logrus_logger.go
```diff
@@ -0,0 +1,130 @@
...
+type logrusLogger struct {
+	impl  logrus.FieldLogger
+	level logrus.Level
+	name  string
+}
+
+// args are alternating key, value pairs, where the keys
+// are expected to be strings, and values can be any type.
+func argsToFields(args ...interface{}) logrus.Fields {
+	var (
```
ncdc: I implemented something yesterday like this:
```go
l := someLogrusLogger
for i := 0; i < len(fields); i += 2 {
  l = l.WithField(fmt.Sprintf("%v", fields[i]), fields[i+1])
}
```

That obviously isn't super efficient for large varargs. We could instead do something like
```go
fields := make(logrus.Fields, len(args)/2)
for i := 0; i < len(args); i += 2 {
  fields[args[i]] = args[i+1]
}
```

pkg/plugin/logrus_logger.go
```diff
@@ -0,0 +1,130 @@
...
+func (l *logrusLogger) With(args ...interface{}) hclog.Logger {
+	return &logrusLogger{
+		impl:  l.impl.WithFields(argsToFields(args...)),
+		level: l.level,
+	}
+}
+
+// Named creates a logger that will prepend the name string on the front of all messages.
+// If the logger already has a name, the new value will be appended to the current
+// name. That way, a major subsystem can use this to decorate all it's own logs
```
ncdc: its

pkg/plugin/manager.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/heptio/ark/pkg/cloudprovider"
+)
+
+type pluginInfo struct {
+	dir    string
+	config plugin.ClientConfig
+}
+
+var pluginBinaries = map[string]pluginInfo{
```
ncdc: I'd put this inside the `manager` instead of making it a package var

pkg/plugin/manager.go
```diff
@@ -0,0 +1,110 @@
...
+	plugin "github.com/hashicorp/go-plugin"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	"github.com/heptio/ark/pkg/cloudprovider"
+)
+
+type pluginInfo struct {
+	dir    string
+	config plugin.ClientConfig
```
ncdc: I wouldn't anticipate this changing per entry, would you?

pkg/plugin/manager.go
```diff
@@ -0,0 +1,110 @@
...
+	logger hclog.Logger
+}
+
+func NewManager(logger logrus.FieldLogger, level logrus.Level) *manager {
+	return &manager{
+		logger: &logrusLogger{impl: logger, level: level},
+	}
+}
+
+func pluginDir() string {
```
ncdc: Do you intend to use this?

pkg/plugin/object_store.go
```diff
@@ -0,0 +1,254 @@
...
+
+// GetObject retrieves the object with the given key from the specified
+// bucket in object storage.
+func (c *ObjectStoreGRPCClient) GetObject(bucket, key string) (io.ReadCloser, error) {
+	stream, err := c.grpcClient.GetObject(context.Background(), &proto.GetObjectRequest{Bucket: bucket, Key: key})
+	if err != nil {
+		return nil, err
+	}
+
+	receive := func() ([]byte, error) {
```
ncdc: I'm guessing if we have other uses for StreamReadCloser, they're probably going to look identical to this? Consider having StreamReadCloser take in an interface?
```go
type RecvCloser interface {
  Recv() ([]byte, error)
  CloseSend() error
}
```

If we need an adapter to turn `*proto.Bytes` into `[]byte`, we could make one instead of doing the inline function. This may be premature optimization, though.

pkg/plugin/stream_reader.go
```diff
@@ -0,0 +1,55 @@
...
+type StreamReadCloser struct {
+	buf     *bytes.Buffer
+	receive ReceiveFunc
+	close   CloseSendFunc
+}
+
+func (s *StreamReadCloser) Read(p []byte) (n int, err error) {
+	// if buf exists and holds more than we're trying to read,
+	// read from the buffer
+	if s.buf != nil && s.buf.Len() >= len(p) {
```
ncdc: Should this be checking `cap` instead?

pkg/plugin/stream_reader.go
```diff
@@ -0,0 +1,55 @@
...
+
+type CloseSendFunc func() error
+
+type StreamReadCloser struct {
+	buf     *bytes.Buffer
+	receive ReceiveFunc
+	close   CloseSendFunc
+}
+
+func (s *StreamReadCloser) Read(p []byte) (n int, err error) {
```
ncdc: We should get @davecheney to look at this too

pkg/plugin/stream_reader.go
```diff
@@ -0,0 +1,55 @@
...
+	// try another read.
+	data, err := s.receive()
+	if err == io.EOF {
+		return s.buf.Read(p)
+	}
+	if err != nil {
+		return 0, err
+	}
+
+	if _, err := s.buf.Write(data); err != nil {
```
ncdc: I wonder if we should copy what we can to `p` (keeping track of `n`), put the leftovers in `s.buf`, and return `n, nil`?

pkg/plugin/logrus_logger.go
skriss: yeah, i was getting tired when i wrote this code..

pkg/plugin/manager.go
```diff
@@ -56,24 +52,72 @@ func pluginDir() string {
...
+	if !found {
+		cfg := &plugin.ClientConfig{
+			HandshakeConfig:  Handshake,
+			AllowedProtocols: []plugin.Protocol{plugin.ProtocolGRPC},
+			Plugins:          make(map[string]plugin.Plugin),
+			Logger:           logger,
+		}
+
+		if (descriptor.kind == PluginKindBlockStore || descriptor.kind == PluginKindObjectStore) &&
+			(descriptor.name == "aws" || descriptor.name == "gcp" || descriptor.name == "azure") {
```
ncdc: Packer lets you override the built-ins. WDYT about supporting that?

pkg/cmd/server/plugin/plugin.go
```diff
@@ -30,65 +30,54 @@ import (
...
-				case "aws":
-					objectStore = aws.NewObjectStore()
-				case "gcp":
-					objectStore = gcp.NewObjectStore()
-				case "azure":
-					objectStore = azure.NewObjectStore()
-				default:
-					log.Fatalf("Unrecognized plugin name %q", pluginName)
+			switch kind {
+			case "cloudprovider":
```
ncdc: This makes it harder to support overriding (see comment below).

pkg/plugin/manager.go
skriss: so look for an external for the kind/name, if it exists use that, else use the built-in? i like that

pkg/cmd/server/plugin/plugin.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+func NewCommand() *cobra.Command {
+	c := &cobra.Command{
+		Use:    "plugin [TYPE] [NAME]",
+		Hidden: true,
+		Short:  "INTERNAL COMMAND ONLY - not intended to be run directly by users",
+		Run: func(c *cobra.Command, args []string) {
+			if len(args) != 2 {
+				log.Fatalf("You must specify exactly two arguments, the plugin type and the plugin name")
```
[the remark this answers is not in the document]
skriss: replaced with a built-for-plugins logrus logger

pkg/cmd/server/plugin/plugin.go
```diff
@@ -0,0 +1,97 @@
...
+			pluginName := args[1]
+
+			log.Printf("Running plugin command for type=%s, name=%s", pluginType, pluginName)
+
+			switch pluginType {
+			case "objectstore":
+				var objectStore cloudprovider.ObjectStore
+
+				switch pluginName {
+				case "aws":
```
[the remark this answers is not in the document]
skriss: done (I think this is what you meant)

pkg/cmd/server/plugin/plugin.go
```diff
@@ -0,0 +1,97 @@
...
+					blockStore = aws.NewBlockStore()
+				case "gcp":
+					blockStore = gcp.NewBlockStore()
+				case "azure":
+					blockStore = azure.NewBlockStore()
+				default:
+					log.Fatalf("Unrecognized plugin name %q", pluginName)
+				}
+
+				plugin.Serve(&plugin.ServeConfig{
```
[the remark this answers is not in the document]
skriss: good point, done

pkg/plugin/block_store.go
```diff
@@ -0,0 +1,211 @@
...
+// CreateVolumeFromSnapshot creates a new block volume, initialized from the provided snapshot,
+// and with the specified type and IOPS (if using provisioned IOPS).
+func (c *BlockStoreGRPCClient) CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ string, iops *int64) (volumeID string, err error) {
+	req := &proto.CreateVolumeRequest{
+		SnapshotID: snapshotID,
+		VolumeType: volumeType,
+		VolumeAZ:   volumeAZ,
+	}
+
+	if iops == nil {
```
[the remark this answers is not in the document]
skriss: eh, this is the only instance of this I can think of.

pkg/plugin/logrus_logger.go
skriss: updated this func.

pkg/plugin/logrus_logger.go
skriss: yep (copied from hclog, i would never make that mistake :)

pkg/plugin/stream_reader.go
skriss: @davecheney could you take a look at this type if you get a sec? I can provide context if helpful

pkg/plugin/stream_reader.go
skriss: I don't think so...

pkg/cmd/server/plugin/plugin.go
skriss: I changed how discovery works in the manager, but I'm not sure I understand this comment - could you clarify?

pkg/cmd/server/plugin/plugin.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	blockStores := map[string]cloudprovider.BlockStore{
+		"aws":   aws.NewBlockStore(),
+		"gcp":   gcp.NewBlockStore(),
+		"azure": azure.NewBlockStore(),
+	}
+
+	c := &cobra.Command{
+		Use:    "plugin [TYPE] [NAME]",
```
ncdc: TYPE -> KIND

pkg/cmd/server/plugin/plugin.go
```diff
@@ -0,0 +1,86 @@
...
+		Short:  "INTERNAL COMMAND ONLY - not intended to be run directly by users",
+		Run: func(c *cobra.Command, args []string) {
+			if len(args) != 2 {
+				logger.Fatal("You must specify exactly two arguments, the plugin kind and the plugin name")
+			}
+
+			kind := args[0]
+			name := args[1]
+
+			logger.Debugf("Running plugin command for kind=%s, name=%s", kind, name)
```
ncdc: Worth moving kind and name to fields?

pkg/cmd/server/server.go
```diff
@@ -341,112 +338,46 @@ func (s *server) initSnapshotService(config *api.Config) error {
...
+	s.snapshotService = cloudprovider.NewSnapshotService(blockStore)
 	return nil
 }
 
-func hasOneCloudProvider(cloudConfig api.CloudProviderConfig) bool {
-	found := false
-
-	if cloudConfig.AWS != nil {
-		found = true
+func getObjectStore(cloudConfig api.CloudProviderConfig, field string, manager plugin.Manager) (cloudprovider.ObjectStore, error) {
```
ncdc: `field` is unused?

pkg/cmd/server/server.go
```diff
@@ -341,112 +338,46 @@ func (s *server) initSnapshotService(config *api.Config) error {
...
-			email = jwtConfig.Email
-			privateKey = jwtConfig.PrivateKey
-		} else {
-			logger.Warning("GOOGLE_APPLICATION_CREDENTIALS is undefined; some features such as downloading log files will not work")
-		}
-
-		objectStorage, err = gcp.NewObjectStorageAdapter(email, privateKey)
-	case cloudConfig.Azure != nil:
-		objectStorage, err = azure.NewObjectStorageAdapter()
+func getBlockStore(cloudConfig api.CloudProviderConfig, field string, manager plugin.Manager) (cloudprovider.BlockStore, error) {
```
ncdc: `field` is unused?

pkg/plugin/block_store.go
```diff
@@ -0,0 +1,227 @@
...
+import (
+	"github.com/hashicorp/go-plugin"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+
+	"github.com/heptio/ark/pkg/cloudprovider"
+	proto "github.com/heptio/ark/pkg/plugin/generated"
+)
+
+type BlockStorePlugin struct {
```
ncdc: Godoc

pkg/plugin/block_store.go
```diff
@@ -0,0 +1,227 @@
...
+	plugin.NetRPCUnsupportedPlugin
+
+	impl cloudprovider.BlockStore
+}
+
+func NewBlockStorePlugin(blockStore cloudprovider.BlockStore) *BlockStorePlugin {
```
ncdc: Godoc

pkg/plugin/block_store.go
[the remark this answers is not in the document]
ncdc: k

pkg/plugin/block_store.go
```diff
@@ -0,0 +1,227 @@
...
+// cannot be initialized from the provided config.
+func (c *BlockStoreGRPCClient) Init(config map[string]string) error {
+	_, err := c.grpcClient.Init(context.Background(), &proto.InitRequest{Config: config})
+
+	return err
+}
+
+// CreateVolumeFromSnapshot creates a new block volume, initialized from the provided snapshot,
+// and with the specified type and IOPS (if using provisioned IOPS).
+func (c *BlockStoreGRPCClient) CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ string, iops *int64) (volumeID string, err error) {
```
ncdc: Can we avoid named return args?

pkg/plugin/block_store.go
```diff
@@ -0,0 +1,227 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return res.VolumeID, nil
+}
+
+// GetVolumeInfo returns the type and IOPS (if using provisioned IOPS) for a specified block
+// volume.
+func (c *BlockStoreGRPCClient) GetVolumeInfo(volumeID, volumeAZ string) (volumeType string, iops *int64, err error) {
```
ncdc: Can we avoid named return args?

pkg/plugin/block_store.go
```diff
@@ -0,0 +1,227 @@
...
+	volumeType = res.VolumeType
+	if res.Iops != 0 {
+		iops = &res.Iops
+	}
+
+	return
+}
+
+// IsVolumeReady returns whether the specified volume is ready to be used.
+func (c *BlockStoreGRPCClient) IsVolumeReady(volumeID, volumeAZ string) (ready bool, err error) {
```
ncdc: Can we avoid named return args?

pkg/plugin/block_store.go
```diff
@@ -0,0 +1,227 @@
...
+	res, err := c.grpcClient.IsVolumeReady(context.Background(), &proto.IsVolumeReadyRequest{VolumeID: volumeID, VolumeAZ: volumeAZ})
+	if err != nil {
+		return false, err
+	}
+
+	return res.Ready, nil
+}
+
+// ListSnapshots returns a list of all snapshots matching the specified set of tag key/values.
+func (c *BlockStoreGRPCClient) ListSnapshots(tagFilters map[string]string) (snapshotIDs []string, err error) {
```
ncdc: Can we avoid named return args?

pkg/plugin/block_store.go
```diff
@@ -0,0 +1,227 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return res.SnapshotIDs, nil
+}
+
+// CreateSnapshot creates a snapshot of the specified block volume, and applies the provided
+// set of tags to the snapshot.
+func (c *BlockStoreGRPCClient) CreateSnapshot(volumeID, volumeAZ string, tags map[string]string) (snapshotID string, err error) {
```
ncdc: Can we avoid named return args?

pkg/util/logging/log_location_hook.go
```diff
@@ -19,23 +19,39 @@ package logging
...
+// value stored in the entry. The default priority value is 0. If
+// removePriorityField is true, the log priority value will be removed
+// from the entry.
+type logLocationHook struct {
+	priority            int
+	removePriorityField bool
 }
 
-func (h *LogLocationHook) Levels() []logrus.Level {
+func NewLogLocationHook(priority int, removePriorityField bool) *logLocationHook {
```
ncdc: Moving forward, let's try to avoid args where it's difficult to discern what they mean (`0, true`, `1, false`). Instead, we can do `NewLogLocationHook().WithPriority(1)` and let it to default to 0 otherwise.

Do we need the concept of removing the priority field? Do you ever envision there being more than "this log came from the plugin" and "this log came from ark"? Could we just do that with a simple marker field that the plugin logger/hook sets, and the ark logger sees it and removes it?

pkg/plugin/manager.go
```diff
@@ -0,0 +1,175 @@
...
+		if _, err := os.Stat(externalPath); err == nil {
+			config := addPlugins(baseConfig, descriptor.kind)
+			config.Cmd = exec.Command(externalPath)
+
+			client = plugin.NewClient(&config)
+
+			m.clients[descriptor] = client
+		} else if _, found := m.internalPlugins[descriptor]; found {
+			config := addPlugins(baseConfig, PluginKindObjectStore, PluginKindBlockStore)
+			config.Cmd = exec.Command("/ark", "plugin", "cloudprovider", descriptor.name)
```
nrb: So even if a plugin is internal, it's still invoked as a subprocess, rather than same-process?

pkg/plugin/object_store.go
```diff
@@ -0,0 +1,270 @@
...
+	res, err := c.grpcClient.ListCommonPrefixes(context.Background(), &proto.ListCommonPrefixesRequest{Bucket: bucket, Delimiter: delimiter})
+	if err != nil {
+		return nil, err
+	}
+
+	return res.Prefixes, nil
+}
+
+// ListObjects gets a list of all objects in bucket that have the same prefix.
+func (c *ObjectStoreGRPCClient) ListObjects(bucket, prefix string) (keys []string, err error) {
```
ncdc: Avoid named return args

pkg/plugin/object_store.go
```diff
@@ -0,0 +1,270 @@
...
+	return stream.SendAndClose(&proto.Empty{})
+}
+
+func (s *ObjectStoreGRPCServer) GetObject(req *proto.GetObjectRequest, stream proto.ObjectStore_GetObjectServer) error {
+	rdr, err := s.impl.GetObject(req.Bucket, req.Key)
+	if err != nil {
+		return err
+	}
+
+	chunk := make([]byte, 1024)
```
ncdc: I wonder if we'd be better off with something between 4 and 32kb?

pkg/plugin/stream_reader.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"bytes"
+	"io"
+)
+
+// ReceiveFunc is a function that either returns a slice
+// of an arbitrary number of bytes OR an error. Returning
+// an io.EOF means there is no more data to be read; any
+// other error is considered an actual error.
+type ReceiveFunc func() ([]byte, error)
```
ncdc: Any thoughts on my previous comment about using an interface instead of a function pointer?

pkg/plugin/stream_reader.go
```diff
@@ -0,0 +1,75 @@
...
+	buf     *bytes.Buffer
+	receive ReceiveFunc
+	close   CloseFunc
+}
+
+func (s *StreamReadCloser) Read(p []byte) (n int, err error) {
+	for {
+		// if buf exists and holds at least as much as we're trying to read,
+		// read from the buffer
+		if s.buf != nil && s.buf.Len() >= len(p) {
```
ncdc: I still think you want to check `cap`, not `len`. If `p` is a slice that has len=0, cap=100, and `s.buf.Len()` is 50, this doesn't match your comment. We're trying to read up to 100 (cap), and you only want to go down this path if `s.buf.Len()` is >= 100, right?

ncdc: Also, if we have data in the buffer, does it make sense to flush it to `p` asap, or always to try to fill the buffer up to `cap(p)` (assuming you agree with ^^^^) before flushing it?

pkg/plugin/manager.go
ncdc: Yes, there's no way around that with the go-plugin package.

pkg/plugin/stream_reader.go
skriss: So the godoc for io.Reader says:

```
Read reads up to len(p) bytes into p. It returns the number of bytes read (0 <= n <= len(p))...
```

ncdc: Ok :-)

hack/_update-proto.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -e
```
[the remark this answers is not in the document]
skriss: was disabling it for now. we need the proto compiler in our build container for this to work -- where would you add it?

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -18,46 +18,70 @@ package gcp
...
+			return errors.WithStack(err)
+		}
+		jwtConfig, err := google.JWTConfigFromJSON(creds)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		email = jwtConfig.Email
+		privateKey = jwtConfig.PrivateKey
+	} else {
+		op.log.Warning("GOOGLE_APPLICATION_CREDENTIALS is undefined; some features such as downloading log files will not work")
```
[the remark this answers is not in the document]
skriss: yeah I think this makes sense.

pkg/cloudprovider/aws/object_store.go
```diff
@@ -18,28 +18,44 @@ package aws
...
 	s3Uploader *s3manager.Uploader
 	kmsKeyID   string
 }
 
-func NewObjectStorageAdapter(region, s3URL, kmsKeyID string, s3ForcePathStyle bool) (cloudprovider.ObjectStorageAdapter, error) {
+func NewObjectStore() *objectStore {
+	return &objectStore{}
+}
+
+func (op *objectStore) Init(config map[string]string) error {
```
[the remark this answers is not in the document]
skriss: I think this is from way back when, when I had these named operators. I updated them all.

pkg/cmd/server/plugin/plugin.go
skriss: you mean creating an Options struct with these as fields?

pkg/plugin/manager.go
```diff
@@ -0,0 +1,175 @@
...
+
+type manager struct {
+	logger          hclog.Logger
+	clients         map[pluginInfo]*plugin.Client
+	internalPlugins map[pluginInfo]interface{}
+}
+
+func NewManager(logger logrus.FieldLogger, level logrus.Level) *manager {
+	return &manager{
+		logger:  (&logrusAdapter{impl: logger, level: level}).Named("plugin"),
```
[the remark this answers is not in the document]
skriss: yeah, I did this just to mirror the hclog implementation. I'd say let's change the key to pluginName, and drop "plugin" as the default name. that way we'll have a structured field telling us which plugin logs are coming from.

pkg/plugin/manager.go
```diff
@@ -0,0 +1,175 @@
...
+			{kind: PluginKindBlockStore, name: "gcp"}:  struct{}{},
+
+			{kind: PluginKindObjectStore, name: "azure"}: struct{}{},
+			{kind: PluginKindBlockStore, name: "azure"}:  struct{}{},
+		},
+	}
+}
+
+func pluginDir() string {
+	pluginDir := os.Getenv("ARK_PLUGIN_DIR")
```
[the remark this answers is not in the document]
skriss: probably not, I didn't have any particular use in mind for this. I'll remove.

pkg/plugin/manager.go
```diff
@@ -0,0 +1,175 @@
...
+	if pluginDir == "" {
+		pluginDir = "/plugins"
+	}
+
+	return pluginDir
+}
+
+func addPlugins(config plugin.ClientConfig, kinds ...PluginKind) plugin.ClientConfig {
+	for _, kind := range kinds {
+		if kind == PluginKindObjectStore {
```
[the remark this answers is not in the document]
skriss: yeah

pkg/plugin/manager.go
```diff
@@ -0,0 +1,175 @@
...
+func pluginDir() string {
+	pluginDir := os.Getenv("ARK_PLUGIN_DIR")
+	if pluginDir == "" {
+		pluginDir = "/plugins"
+	}
+
+	return pluginDir
+}
+
+func addPlugins(config plugin.ClientConfig, kinds ...PluginKind) plugin.ClientConfig {
```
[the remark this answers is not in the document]
skriss: yeah, I was trying to do something clever here that ended up not making sense. switching.

pkg/plugin/stream_reader.go
skriss: I'm fine with that -- I originally went with funcs because I didn't want the interface def to be tied just to a grpc Bytes stream client, and I didn't want to define another adapter type, but that's all it's used for. Are you thinking:

```
type ByteReceiver interface {
    Recv() (*generated.Bytes, error)
    CloseSend() error
}
``` 
that the Bytes stream client already satisfies? or a more generic
```
type ByteReceiver interface {
    Receive() ([]byte, error)
    Close() error
}
```
along with an adapter?

pkg/util/logging/log_location_hook.go
skriss: I originally had something simpler like that, then thought I'd generalize to what I have now. Somehow I feel like I always end up on the other side of these kinds of decisions as you :) I suppose if someone implements a plugin framework for their plugin, my current impl might be useful..

pkg/plugin/stream_reader.go
skriss: Ah, I see the former is exactly what you commented before. I'm happy with that.

skriss:
@ncdc addressed most feedback, finishing up the LogLocation change now.

pkg/cmd/server/plugin/plugin.go
ncdc: No, I meant structured logrus fields.

pkg/plugin/block_store.go
```diff
@@ -0,0 +1,247 @@
...
+	return nil
+}
+
+// GRPCClient returns a BlockStore gRPC client.
+func (p *BlockStorePlugin) GRPCClient(c *grpc.ClientConn) (interface{}, error) {
+	return &BlockStoreGRPCClient{grpcClient: proto.NewBlockStoreClient(c)}, nil
+}
+
+// BlockStoreGRPCClient implements the cloudprovider.BlockStore interface and forwards
+// calls to a gRPC client.
```
ncdc: It forwards to a gRPC server, right?

pkg/plugin/object_store.go
```diff
@@ -0,0 +1,289 @@
...
+// object storage bucket with the given key.
+func (c *ObjectStoreGRPCClient) PutObject(bucket, key string, body io.Reader) error {
+	stream, err := c.grpcClient.PutObject(context.Background())
+	if err != nil {
+		return err
+	}
+
+	// read from the provider io.Reader into chunks, and send each one over
+	// the gRPC stream
+	chunk := make([]byte, 1024)
```
ncdc: Same q from before - do you think we could get better performance with a larger chunk size?

skriss: yes, probably (I imagine the actual wall-time impact is pretty minimal), although I don't know what value makes the most sense.

pkg/plugin/block_store.go
skriss: this struct makes calls on the gRPC client, which get routed to the gRPC server (plugin). Maybe "forward" is a bad word? Could say "...uses a gRPC client to make calls"?

pkg/plugin/object_store.go
ncdc: I would do somewhere between 4 and 32kb. @davecheney any suggestions?

skriss:
@ncdc are you okay with me doing some squashing now? If it'll make your review job harder I'll hold off but would like to do some consolidation.

ncdc:
Squash away

skriss:
@ncdc do you want me to move the generated proto/grpc code into the proto dir with package name `proto`?

ncdc:
No need right now

docs/config-definition.md
```diff
@@ -44,9 +46,13 @@ The configurable parameters are as follows:
...
-| `backupStorageProvider`/(inline) | CloudProviderConfig<br><br>(Supported key values are `aws`, `gcp`, and `azure`, but only one can be present. See the corresponding [AWS][0], [GCP][1], and [Azure][2]-specific configs.) | Required Field | The specification for whichever cloud provider will be used to actually store the backups. |
+| `persistentVolumeProvider` | CloudProviderConfig | None (Optional) | The specification for whichever cloud provider the cluster is using for persistent volumes (to be snapshotted), if any.<br><br>If not specified, Backups and Restores requesting PV snapshots & restores, respectively, are considered invalid. <br><br> *NOTE*: For Azure, your Kubernetes cluster needs to be version 1.7.2+ in order to support PV snapshotting of its managed disks. |
+| `persistentVolumeProvider/name` | String<br><br>(Ark natively supports `aws`, `gcp`, and `azure`. Other providers may be available via external plugin.) | None (Optional) | The name of the cloud provider the cluster is using for persistent volumes, if any. |
```
ncdc: "via external plugin**s**."

docs/config-definition.md
```diff
@@ -44,9 +46,13 @@ The configurable parameters are as follows:
...
+| `persistentVolumeProvider/config` | map[string]string<br><br>(See the corresponding [AWS][0], [GCP][1], and [Azure][2]-specific configs or your provider's documentation.) | None (Optional) | Configuration keys/values to be passed to the cloud provider BlockStore.  |
```
ncdc: Can't say BlockStore

docs/config-definition.md
```diff
@@ -44,9 +46,13 @@ The configurable parameters are as follows:
...
+| `backupStorageProvider` | CloudProviderConfig | Required Field | The specification for whichever cloud provider will be used to actually store the backups. |
+| `backupStorageProvider/name` | String<br><br>(Ark natively supports `aws`, `gcp`, and `azure`. Other providers may be available via external plugin.) | Required Field | The name of the cloud provider that will be used to actually store the backups. |
```
ncdc: plugins

docs/config-definition.md
```diff
@@ -44,9 +46,13 @@ The configurable parameters are as follows:
...
 | `backupStorageProvider/bucket` | String | Required Field | The storage bucket where backups are to be uploaded. |
+| `backupStorageProvider/config` | map[string]string<br><br>(See the corresponding [AWS][0], [GCP][1], and [Azure][2]-specific configs or your provider's documentation.) | None (Optional) | Configuration keys/values to be passed to the cloud provider ObjectStore. |
```
ncdc: Can't say ObjectStore

ncdc:
This is a big change. We would really appreciate it if people could test out the master branch (once this is merged), especially making sure that the following operations function properly:

- Basic backups - the backup tar.gz file is uploaded to object storage correctly
- Backup PV snapshotting - if you have a PV, it gets snapshotted
- Backup logs - `ark backup logs` works correctly
- Backup downloads - `ark backup download` works correctly, and you receive a valid, full tar.gz file
- Basic restores - work as expected
- Restore logs - `ark restore logs` works correctly
- Restore details - `ark restore describe` works correctly (you should see information about warnings and errors, if you have any)
- Restore PVs from snapshots - if your backup had a PV, your restore works correctly to create a new PV from the snapshot. NOTE, we don't support restoring into the same cluster when the PV still exists, so you either need to delete the PV and then restore, or restore into a different cluster.

To test:

- `git clone https://github.com/heptio/ark`
- cd into the ark directory
- Install ark following the instructions for your cloud provider of choice
- Run `kubectl -n heptio-ark set image deploy/ark ark=gcr.io/heptio-images/ark:master`
- Wait for the deployment to finish
- Take it for a spin
- You can see logs from the ark server via `kubectl -n heptio-ark logs deploy/ark`
- File any issues you encounter at https://github.com/heptio/ark/issues/new

Thanks!

PR #199, opened 2017-11-14
Log enhancements

ncdc:
- Log when starting a backup in the ark server log (we were previously only logging completion)
- Log when starting a restore in the ark server log
- Log when a restore has completed in the ark server log
- Add error and log location hooks to the backup and restore loggers
- Make logSource more concise by stripping off the leading .../github.com/heptio/ark/


skriss:
LGTM, merge on travis

PR #211, opened 2017-11-22
AWS KMS encrypted restore volumes

mstump:
Builds on #185 to add support for restoring to encrypted volumes on AWS


skriss:
@ncdc I tested again, and volumes created from encrypted snapshots are automatically encrypted (and vice versa). OK to close this PR?

ncdc:
Yes. If we need to do more work in the future, we can revisit.

PR #213, opened 2017-11-27
Convert restorers to plugins

skriss:

@ncdc I'd like to do some re-testing of this before merge since I've done a bit of refactoring since my last full round, but the code is ready for review. I should be able to address comments tomorrow night on the flight home


pkg/restore/restore.go
```diff
@@ -533,6 +677,107 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+		}
+	}
+
+	var warning error
+
+	if ctx.snapshotService == nil && len(backup.Status.VolumeBackups) > 0 {
+		warning = errors.New("unable to restore PV snapshots: Ark server is not configured with a PersistentVolumeProvider")
+	}
+
+	return obj, warning, nil
```
nrb: Is having 2 `error` fields on return with one for warnings a Go idiom? I think I understand that it's meant to protect against callers being overzealous with `if err != nil`, but it looks like there's only one case here where we raise a warning in this function

ncdc: This is because we (ark) support both warnings and errors when restoring. This allows individual restore actions to distinguish as well.

pkg/restore/restore.go
```diff
@@ -363,6 +419,41 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+		return &v1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: remappedName,
+			},
+		}
+	}
+
+	var backupNS v1.Namespace
+	if err := json.Unmarshal(nsBytes, &backupNS); err != nil {
+		return &v1.Namespace{
```
ncdc: Should we log here? Or return an error? Presumably it means the file is corrupt?

pkg/restore/restore.go
```diff
@@ -363,6 +419,41 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: remappedName,
+			},
+		}
+	}
+
+	return &v1.Namespace{
```
ncdc: Would it make sense to start with `backupNS` and mutate fields as needed? That would ensure we get everything from the backed up NS, instead of "opting in" to just labels & annotations.

pkg/restore/restore.go
```diff
@@ -455,72 +565,106 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 				addArkError(&errs, fmt.Errorf("error getting resource client for namespace %q, resource %q: %v", namespace, &groupResource, err))
 				return warnings, errs
 			}
+		}
+
+		if groupResource.Group == "" && groupResource.Resource == "persistentvolumes" {
+			// we need to remove annotations from PVs since they potentially contain
```
ncdc: Should we make this a func?

pkg/restore/restore.go
skriss: I think logging a warning is reasonable -- I don't really see a reason to error out though. Do you?

pkg/restore/restore.go
skriss: hmm. not sure. with other objects we're restoring, in the `resetMetadataAndStatus` func we are taking the opt-in approach. If we take the alternate approach there's a whole bunch of stuff that we have to explicitly clear out of ObjectMeta (`selfLink`, `uid`, `creationTimestamp`, etc.)

pkg/restore/restore.go
skriss: sure.

pkg/restore/restore.go
ncdc: It just means that you have to go digging for why your labels/annotations weren't restored properly. I think logging is a good enough place to start.

PR #215, opened 2017-11-27
Support custom volume snapshots & restores

### LINKED ISSUE #214: BlockStore: add way for plugins to tell Ark what a PV's volume ID is
pkg/util/kube/utils.go currently hard-codes support for extracting the volume ID from a PV if and only if the volume type is for aws, azure, or gce. We need to add a new rpc method to the block store proto so plugins can tell Ark what the volume id is for a given PV.

ncdc:
The main Ark code was hard-coding specific support for AWS, GCE, and Azure volume snapshots and restores, and anything else was considered unsupported. This PR adds GetVolumeID and SetVolumeID to the BlockStore interface, to allow block store plugins to handle volume snapshots and restores.

Fixes #214 

cc @stevesloka


skriss:
The plugin change definitely makes sense - thanks for catching that. 

I think the PV restorer change is OK, although maybe we still want to return a warning if there is a VolumeBackups map but the current PV isn't in it? It's certainly not an expected case -- likely either the snapshot failed during backup, or someone manually deleted it - so probably worth explicitly recording in the log.

pkg/cloudprovider/aws/block_store.go
```diff
@@ -205,3 +209,29 @@ func (b *blockStore) DeleteSnapshot(snapshotID string) error {
...
+
+func (b *blockStore) SetVolumeID(pv runtime.Unstructured, volumeID string) (runtime.Unstructured, error) {
+	aws, err := collections.GetMap(pv.UnstructuredContent(), "spec.awsElasticBlockStore")
+	if err != nil {
+		return nil, err
+	}
+
+	aws["volumeID"] = volumeID
+
+	return pv, nil
```
stevesloka: `pv` was never changed here and `aws` is not returned?

ncdc: We have to return the full object. `aws` comes from `pv` (`collections.GetMap(pv.UnstructuredContent(), "spec.awsElasticBlockStore")`). When we mutate `aws`, it's mutating `pv`.

ncdc:
@skriss 
> I think the PV restorer change is OK, although maybe we still want to return a warning if there is a VolumeBackups map but the current PV isn't in it? It's certainly not an expected case -- likely either the snapshot failed during backup, or someone manually deleted it - so probably worth explicitly recording in the log.

I'll continue to have it return an error.

pkg/cloudprovider/snapshot_service.go
```diff
@@ -46,6 +47,9 @@ type SnapshotService interface {
 
 	// GetVolumeInfo gets the type and IOPS (if applicable) from the cloud API.
 	GetVolumeInfo(volumeID, volumeAZ string) (string, *int64, error)
+
+	GetVolumeID(pv runtime.Unstructured) (string, error)
```
skriss: godoc

pkg/cloudprovider/storage_interfaces.go
```diff
@@ -65,6 +67,11 @@ type BlockStore interface {
 	// and with the specified type and IOPS (if using provisioned IOPS).
 	CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ string, iops *int64) (volumeID string, err error)
 
+	// GetVolumeID returns the cloud provider specific identifier for the PersistentVolume.
+	GetVolumeID(pv runtime.Unstructured) (string, error)
+
+	SetVolumeID(pv runtime.Unstructured, volumeID string) (runtime.Unstructured, error)
```
skriss: godoc

pkg/plugin/block_store.go
```diff
@@ -150,6 +154,49 @@ func (c *BlockStoreGRPCClient) DeleteSnapshot(snapshotID string) error {
...
+		return "", err
+	}
+
+	req := &proto.GetVolumeIDRequest{
+		PersistentVolume: encodedPV,
+	}
+
+	resp, err := c.grpcClient.GetVolumeID(context.Background(), req)
+	if err != nil {
+		return "", nil
```
skriss: shouldn't this be returning `"", err`?

pkg/plugin/block_store.go
```diff
@@ -150,6 +154,49 @@ func (c *BlockStoreGRPCClient) DeleteSnapshot(snapshotID string) error {
...
+	}
+
+	req := &proto.SetVolumeIDRequest{
+		PersistentVolume: encodedPV,
+		VolumeID:         volumeID,
+	}
+
+	resp, err := c.grpcClient.SetVolumeID(context.Background(), req)
+	if err != nil {
+		return nil, nil
```
skriss: shouldn't this be returning `nil, err`?

pkg/restore/restorers/pv_restorer.go
```diff
@@ -19,12 +19,13 @@ package restorers
 import (
 	"github.com/pkg/errors"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	api "github.com/heptio/ark/pkg/apis/ark/v1"
 	"github.com/heptio/ark/pkg/cloudprovider"
+	"github.com/heptio/ark/pkg/util/boolptr"
```
skriss: this pkg is missing from the commit

pkg/restore/restorers/pv_restorer.go
```diff
@@ -56,66 +57,46 @@ func (sr *persistentVolumeRestorer) Prepare(obj runtime.Unstructured, restore *a
...
-		}
-
-		// if there are snapshots, and this is a supported PV type, but there's no
-		// snapshot for this PV, it's an error
-		if backup.Status.VolumeBackups[pvName] == nil {
-			return nil, nil, errors.Errorf("no snapshot found to restore volume %s from", pvName)
-		}
-
-		restoreFromSnapshot = true
+	if sr.snapshotService == nil {
```
skriss: so this is going to return an error if SnapshotVolumes is auto for both the backup and the restore, and there was no snapshotService for the backup (so no backups).  I think this is definitely not desired behavior. I think we want to exit prior to this (without error) if `backup.status.VolumeBackups` is nil/empty, right? (we may have discussed this already this AM).

pkg/restore/restorers/pv_restorer.go
```diff
@@ -56,66 +57,46 @@ func (sr *persistentVolumeRestorer) Prepare(obj runtime.Unstructured, restore *a
...
 
-	if sr.snapshotService == nil && len(backup.Status.VolumeBackups) > 0 {
-		warning = errors.New("unable to restore PV snapshots: Ark server is not configured with a PersistentVolumeProvider")
+	updatedObj, err := sr.snapshotService.SetVolumeID(obj, volumeID)
+	if err != nil {
+		return nil, nil, err
 	}
 
-	return obj, warning, nil
+	return updatedObj, nil, nil
```
skriss: Looks like we can get rid of the `warning` return arg since it's always nil

pkg/plugin/block_store.go
ncdc: good catch, thx

pkg/restore/restorers/pv_restorer.go
ncdc: oops :)

pkg/restore/restorers/pv_restorer.go
ncdc: I'll do that when I rebase

pkg/restore/restorers/pv_restorer.go
ncdc: Blech. That wasn't a case I was thinking about. I'll add that check.

pkg/backup/item_backupper_test.go
```diff
@@ -455,7 +458,7 @@ func TestTakePVSnapshot(t *testing.T) {
 			name:            "unsupported PV source type",
 			snapshotEnabled: true,
 			pv:              `{"apiVersion": "v1", "kind": "PersistentVolume", "metadata": {"name": "mypv"}, "spec": {"unsupportedPVSource": {}}}`,
-			expectError:     false,
+			expectError:     true,
```
skriss: Why did this change from false to true? Shouldn't it still not return an error since it doesn't contain a supported PV source?

skriss:
Just the one unit test Q, otherwise LGTM

pkg/backup/item_backupper_test.go
ncdc: You're right, thanks.

skriss:
LGTM

PR #217, opened 2017-11-29
add plugin add/remove commands

### LINKED ISSUE #210: Make it easy to add/remove plugins
This is what I've been testing for a plugin installation mechanism:

1. Create a docker image (from alpine) that contains the plugin binaries in `/plugins` inside the image
1. Set the image's entrypoint to `["sh", "-c", "cp -a /plugins/* /target/."]` (this copies everything from `/plugins` to `/target`, where `/target` is actually a volume mount for `/plugins` in the `ark` container)
1. In the `ark` deployment, add an init container like this:
```yaml
- name: objectstore-andy
  image: gcr.io/heptio-images/ark-plugin-example:latest
  imagePullPolicy: IfNotPresent
  volumeMounts:
    - name: plugins
      mountPath: /target
```
4. In the `ark` deployment, add a volume like this:
```yaml
- name: plugins
  emptyDir: {}
```
5. In the `ark` deployment, add a volume mount for the `ark` container like this:
```yaml
- name: plugins
  mountPath: /plugins
```

I think this could be simplified if
1. By default, the `ark` deployment has the `plugins` volume & volume mount
1. We add `ark plugin add` and `ark plugin remove` to manipulate the init containers

skriss:

Fixes #210

### DIFF
--- a/pkg/cmd/cli/plugin/add.go
+++ b/pkg/cmd/cli/plugin/add.go
@@ -1,0 +1,162 @@
+/*
+Copyright 2017 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package plugin
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"k8s.io/api/apps/v1beta1"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
+
+	ark "github.com/heptio/ark/pkg/apis/ark/v1"
+	"github.com/heptio/ark/pkg/client"
+	"github.com/heptio/ark/pkg/cmd"
+	"github.com/heptio/ark/pkg/cmd/util/flag"
+)
+
+const (
+	pluginsVolumeName = "plugins"
+	arkDeployment     = "ark"
+	arkContainer      = "ark"
+)
+
+func NewAddCommand(f client.Factory) *cobra.Command {
+	var (
+		imagePullPolicies   = []string{string(v1.PullAlways), string(v1.PullIfNotPresent), string(v1.PullNever)}
+		imagePullPolicyFlag = flag.NewEnum(string(v1.PullIfNotPresent), imagePullPolicies...)
+	)
+
+	c := &cobra.Command{
+		Use:   "add IMAGE",
+		Short: "Add a plugin",
+		Run: func(c *cobra.Command, args []string) {
+			if len(args) != 1 {
+				cmd.CheckError(errors.New("you must specify only one argument, the plugin container image"))
+			}
+
+			kubeClient, err := f.KubeClient()
+			if err != nil {
+				cmd.CheckError(err)
+			}
+
+			arkDeploy, err := kubeClient.AppsV1beta1().Deployments(ark.DefaultNamespace).Get(arkDeployment, metav1.GetOptions{})
+			if err != nil {
+				cmd.CheckError(err)
+			}
+
+			original, err := json.Marshal(arkDeploy)
+			cmd.CheckError(err)
+
+			// ensure the plugins volume & mount exist
+			volumeExists := false
+			for _, volume := range arkDeploy.Spec.Template.Spec.Volumes {
+				if volume.Name == pluginsVolumeName {
+					volumeExists = true
+					break
+				}
+			}
+
+			if !volumeExists {
+				volume := v1.Volume{
+					Name: pluginsVolumeName,
+					VolumeSource: v1.VolumeSource{
+						EmptyDir: &v1.EmptyDirVolumeSource{},
+					},
+				}
+
+				volumeMount := v1.VolumeMount{
+					Name:      pluginsVolumeName,
+					MountPath: "/plugins",
+				}
+
+				arkDeploy.Spec.Template.Spec.Volumes = append(arkDeploy.Spec.Template.Spec.Volumes, volume)
+
+				containers := arkDeploy.Spec.Template.Spec.Containers
+				containerIndex := -1
+				for x, container := range containers {
+					if container.Name == arkContainer {
+						containerIndex = x
+						break
+					}
+				}
+
+				if containerIndex < 0 {
+					cmd.CheckError(errors.New("ark container not found in ark deployment"))
+				}
+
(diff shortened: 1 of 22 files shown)


ncdc:
SHould use patch, not update

skriss:
Any good example of how to do that correctly? Is `	"k8s.io/apimachinery/pkg/util/strategicpatch.CreateTwoWayMergePatch()` the way to go?

ncdc:
Yes, that looks like the right one. Just grep for CreateTwoWayMergePatch in the kube code base for examples.

pkg/cmd/cli/plugin/add.go
```diff
@@ -0,0 +1,83 @@
...
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	ark "github.com/heptio/ark/pkg/apis/ark/v1"
+	"github.com/heptio/ark/pkg/client"
+	"github.com/heptio/ark/pkg/cmd"
+)
+
+func NewAddCommand() *cobra.Command {
```
ncdc: Have this take in a client.Factory

pkg/cmd/cli/plugin/add.go
```diff
@@ -0,0 +1,83 @@
...
+
+	c := &cobra.Command{
+		Use:   "add NAME IMAGE",
+		Short: "Add a plugin",
+		Run: func(c *cobra.Command, args []string) {
+			if len(args) != 2 {
+				cmd.CheckError(errors.New("you must specify exactly two arguments, the plugin container's name and image"))
+			}
+
+			clientConfig, err := client.Config(kubeconfig, fmt.Sprintf("%s-%s", c.Parent().Name(), c.Name()))
```
ncdc: Use the passed in client.Factory to get your client

pkg/cmd/cli/plugin/add.go
```diff
@@ -0,0 +1,83 @@
...
+			if err != nil {
+				cmd.CheckError(err)
+			}
+
+			kubeClient, err := kubernetes.NewForConfig(clientConfig)
+			if err != nil {
+				cmd.CheckError(err)
+			}
+
+			arkDeploy, err := kubeClient.Apps().Deployments(ark.DefaultNamespace).Get("ark", metav1.GetOptions{})
```
ncdc: Don't use `.Apps()` - use a specific version instead (e.g. `AppsV1beta1()`).

pkg/cmd/cli/plugin/add.go
```diff
@@ -0,0 +1,83 @@
...
+			if err != nil {
+				cmd.CheckError(err)
+			}
+
+			plugin := v1.Container{
+				Name:            args[0],
```
ncdc: Maybe we could derive `Name` from the image, so the command is just `ark plugin add IMAGE`? Or if deriving the name is annoying, we could just call it plugin-0001, plugin-0002, etc.

pkg/cmd/cli/plugin/add.go
```diff
@@ -0,0 +1,83 @@
...
+				VolumeMounts: []v1.VolumeMount{
+					{
+						Name:      "plugins",
+						MountPath: "/target",
+					},
+				},
+			}
+
+			arkDeploy.Spec.Template.Spec.InitContainers = append(arkDeploy.Spec.Template.Spec.InitContainers, plugin)
+
```
ncdc: It might be nice to create the "plugins" `Volume` if it doesn't already exist.

pkg/cmd/cli/plugin/add.go
```diff
@@ -0,0 +1,83 @@
...
+			if _, err := kubeClient.Apps().Deployments(ark.DefaultNamespace).Update(arkDeploy); err != nil {
+				cmd.CheckError(err)
+			}
+		},
+	}
+
+	c.Flags().StringVar(&kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use to talk to the Kubernetes apiserver. If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration")
```
ncdc: You can remove this once you switch to using the factory

pkg/cmd/cli/plugin/plugin.go
```diff
@@ -0,0 +1,36 @@
...
+limitations under the License.
+*/
+
+package plugin
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func NewCommand() *cobra.Command {
```
ncdc: factory

pkg/cmd/cli/plugin/plugin.go
```diff
@@ -0,0 +1,36 @@
...
+	c := &cobra.Command{
+		Use:   "plugin",
+		Short: "Work with plugins",
+		Long:  "Work with plugins",
+	}
+
+	c.AddCommand(
+		NewAddCommand(),
```
ncdc: factory

pkg/cmd/cli/plugin/plugin.go
```diff
@@ -0,0 +1,36 @@
...
+		NewRemoveCommand(),
```
ncdc: factory

pkg/cmd/cli/plugin/remove.go
```diff
@@ -0,0 +1,86 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	ark "github.com/heptio/ark/pkg/apis/ark/v1"
+	"github.com/heptio/ark/pkg/client"
+	"github.com/heptio/ark/pkg/cmd"
+)
+
+func NewRemoveCommand() *cobra.Command {
```
ncdc: factory

pkg/cmd/cli/plugin/remove.go
```diff
@@ -0,0 +1,86 @@
...
+
+	c := &cobra.Command{
+		Use:   "remove [NAME | IMAGE]",
+		Short: "Remove a plugin",
+		Run: func(c *cobra.Command, args []string) {
+			if len(args) != 1 {
+				cmd.CheckError(errors.New("you must specify only one argument, the plugin container's name or image"))
+			}
+
+			clientConfig, err := client.Config(kubeconfig, fmt.Sprintf("%s-%s", c.Parent().Name(), c.Name()))
```
ncdc: factory

pkg/cmd/cli/plugin/remove.go
```diff
@@ -0,0 +1,86 @@
...
+
+			arkDeploy.Spec.Template.Spec.InitContainers = append(initContainers[0:index], initContainers[index+1:]...)
+
+			if _, err := kubeClient.Apps().Deployments(ark.DefaultNamespace).Update(arkDeploy); err != nil {
+				cmd.CheckError(err)
+			}
+		},
+	}
+
+	c.Flags().StringVar(&kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use to talk to the Kubernetes apiserver. If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration")
```
ncdc: Remove

pkg/cmd/cli/plugin/add.go
skriss: client.Factory currently only returns an Ark clientset...I suppose I should move the logic for creating a standard kube clientset in there?

ncdc: Yeah I think that make sense. Just add a KubeClient() func maybe?

skriss: sgtm

PR #222, opened 2017-11-29
Fix log location hook prefix stripping

ncdc:
The log location hook was matching github.com/heptio/ark and stripping
off that + 1 more char. This meant that
github.com/heptio/ark-plugin-example/foo.go was being listed as
plugin-example/foo.go instead of
github.com/heptio/ark-plugin-example/foo.go.


skriss:
LGTM.

PR #227, opened 2017-11-29
Bug fixes: obj/block store plugin logging and remapped namespaces issue

skriss:

-Each plugin needs its own logrusAdapter so that when item actions are redirected to log in the backup/restore logs, the obj/block store plugins continue logging to the server log.
-When restoring, don't restore Namespace API objects independently. They'll be restored when objects get restored into them. The bug seen was that when remapping namespaces, an empty namespace with the old name got created (the objects were correctly restored into the renamed NS).

### DIFF
--- a/pkg/plugin/manager.go
+++ b/pkg/plugin/manager.go
@@ -128,7 +128,8 @@ type Manager interface {
 }
 
 type manager struct {
-	logger         *logrusAdapter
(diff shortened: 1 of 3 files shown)


ncdc:
Would like to see a unit test for the namespace fix, if possible

ncdc:
Tested locally, both fixes appear to have worked.

PR #237, opened 2017-12-06
Document GKE cluster-admin step

nrb:
When using GKE, there's a missing step, so document it for users on that platform.

### DIFF
--- a/docs/cloud-provider-specifics.md
+++ b/docs/cloud-provider-specifics.md
@@ -157,10 +157,10 @@ To integrate Heptio Ark with GCP, you should follow the instructions below to cr
 
 #### Credentials and configuration
 
-When using Google Kubernetes Engine (GKE), be sure to make your current IAM user a cluster-admin. See [their docs][22] for more information.
+When using Google Kubernetes Engine (GKE), be sure to make your current IAM user a cluster-admin, since creating RBAC objects requires it.
+See [their docs][22] for more information.
 
 In the Ark root directory, run the following to first set up namespaces, RBAC, and other scaffolding:
-
 
 ```bash
 kubectl apply -f examples/common/00-prereqs.yaml
--- a/docs/cloud-provider-specifics.md
+++ b/docs/cloud-provider-specifics.md
@@ -105,6 +105,10 @@ Now that you have your IAM user credentials stored in a Secret, you need to repl
 
 ### GCP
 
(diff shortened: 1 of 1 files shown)


docs/cloud-provider-specifics.md
```diff
@@ -155,6 +155,8 @@ To integrate Heptio Ark with GCP, you should follow the instructions below to cr
 
 In the Ark root directory, run the following to first set up namespaces, RBAC, and other scaffolding:
 
+On GKE, be sure to make your current IAM user a cluster-admin. See [their docs][22] for more information.
```
Bradamant3: s/GKE/GCP yes? (it's what we refer to ^^, and what the link points to)

nrb: This particular sentence is referring to their Kubernetes Engine product, which is part of GCP's offering. I added this sentence to clarify for GKE users because Ark's prereq file actually errors when applied to GKE clusters.

I was considering making an entirely separate section, but this appears to be the only difference between running one's own cluster on top of GCP virtual machines versus using one generated by their hosted offering so far. It could probably use a little more clarification, though.

Bradamant3: got it. Maybe leave as is for now and review for clarification later? Up to you ...

nrb:
@Bradamant3 I've updated the added sentence a bit - do you think it could use any more context or detail?

skriss:
@Bradamant3 @nrb would be great to get this merged soon.

nrb:
@skriss I'm fine with it as is for now. IMO we could merge it, but don't want to self-merge things.

docs/cloud-provider-specifics.md
```diff
@@ -155,6 +155,8 @@ To integrate Heptio Ark with GCP, you should follow the instructions below to cr
 
 In the Ark root directory, run the following to first set up namespaces, RBAC, and other scaffolding:
 
+When using Google Kubernetes Engine (GKE), be sure to make your current IAM user a cluster-admin. See [their docs][22] for more information.
```
skriss: Personally I would probably move this above the prior line, and add a brief explanation of why this is necessary (i.e. on GKE you are not cluster-admin by default and this prevents you from creating the RBAC resources in 00-prereqs.yaml

nrb: Thanks for that feedback - does the new commit help?

skriss: WFM. Would appreciate a last glance from @Bradamant3

PR #241, opened 2017-12-11
switch from Update() to Patch()

### LINKED ISSUE #133: Switch to an interface for the logger
Not that we want to switch impls again, but it's annoying to pass around `*logrus.Logger` vs `*logrus.Entry`. Let's move to an interface and call it a day.

cc @davecheney b/c I know you'll love this 😄

### LINKED ISSUE #181: Replace testlogger.NewNullLogger with arktest.NewLogger

### LINKED ISSUE #221: Replace Update() calls with Patch()
Patch is safer and less likely to blow away changes unintentionally.

- [ ] schedule controller
- [ ] backup controller
- [ ] restore controller
- [ ] download request controller

skriss:

Fixes #221 
Fixes #133 
Fixes #181


pkg/cmd/cli/plugin/add.go
```diff
@@ -131,7 +131,7 @@ func NewAddCommand(f client.Factory) *cobra.Command {
 			patchBytes, err := strategicpatch.CreateTwoWayMergePatch(original, updated, v1beta1.Deployment{})
 			cmd.CheckError(err)
 
-			_, err = kubeClient.AppsV1beta1().Deployments(ark.DefaultNamespace).Patch(arkDeploy.Name, types.StrategicMergePatchType, patchBytes)
+			_, err = kubeClient.AppsV1beta1().Deployments(ark.DefaultNamespace).Patch(arkDeploy.Name, types.MergePatchType, patchBytes)
```
ncdc: Why the change?

pkg/controller/backup_controller.go
```diff
@@ -224,6 +227,7 @@ func (controller *backupController) processBackup(key string) error {
 
 	logContext.Debug("Cloning backup")
 	// don't modify items in the cache
+	original := *backup
```
ncdc: Why'd you deref the pointer?

ncdc: This also makes it look like this is a safe way to avoid modifying items in the cache

pkg/cmd/cli/plugin/add.go
skriss: I guess it's not actually necessary; saw that strategic merge patch doesn't work OOTB for CRD's so switched those to `MergePatchType` and changed this one to be consistent. It works either way given what we're doing, but I can leave it as `StrategicMergePatchType`.

pkg/controller/backup_controller_test.go
```diff
@@ -201,15 +201,30 @@ func TestProcessBackup(t *testing.T) {
 			}
 
 			// this is necessary so the Update() call returns the appropriate object
```
ncdc: Update->Patch

skriss:
@ncdc just pushed some refactoring to the unit tests to validate patches appropriately. Would appreciate a review on those before I do the fixes for the other 2 controllers' unit tests,

skriss:
would still like to do some manual testing but this is now passing all unit tests. I ended up also switching the rest of the codebase over to `logrus.FieldLogger` and `arktest.NewLogger()` because I had already done most of it anyway.

ncdc:
Run it through the e2e?

PR #248, opened 2017-12-13
fix examples/common yaml indent/whitespace

dogopupper:
Was failing multiple linting tests.


ncdc:
Thank you for your PR. However, we prefer the format we currently have, as we believe it is easier to read when arrays are indented. Both ways are valid yaml and can be parsed and processed successfully.

PR #249, opened 2017-12-13
update GCP client libraries

skriss:


skriss:
We can go ahead with this PR. I've tested everything on a GKE cluster -- @nrb would be nice to also run through e2e if possible.

nrb:
I can double check. If you're looking for PV testing, though, that likely isn't ready. Last  I checked, the Kubicorn GCP profile didn't have the service account integration necessary to manage cloud provider objects.

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -67,78 +67,88 @@ func (o *objectStore) Init(config map[string]string) error {
...
+		if err == iterator.Done {
+			return res, nil
+		}
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
 
-	return ret, nil
+		if obj.Prefix != "" {
+			res = append(res, obj.Prefix[0:strings.LastIndex(obj.Prefix, delimiter)])
```
ncdc: I'm not sure we still need to take the substring? According to https://godoc.org/cloud.google.com/go/storage#Query:
> // ... Objects whose names,
>    // aside from the prefix, contain delimiter will have their name,
>    // truncated after the delimiter, returned in prefixes.

skriss: the `Prefix` values always contain a trailing `delimiter` (just re-confirmed this), so we need to remove it to get backup names. The language in that godoc comment is super-confusing, but I'm fairly certain that how this works is:

- if there are any objects in the bucket that don't contain `delimiter` in their name, they'll be returned, with the `Name` field set and no `Prefix`. (note that we don't have any of these in Ark because all of our objects are prefixed with `<backup-name>/`).
-for any objects that *do* contain `delimiter`, a result will be returned with the `Name` field empty and `Prefix` set to the substring of their name up to and including `delimiter`.
-results containing only `Prefix` are de-duped.

ncdc: Ah, truncated *after* the delimiter. OK.

ncdc:
Can you split out the dep/vendor changes into a separate commit?

skriss:
split.

PR #252, opened 2017-12-15
add delete backup cmd using finalizer and simplify GC process

### LINKED ISSUE #77: GC should evaluate all backup dirs (common prefixes) in object storage to handle partial/orphaned files

### LINKED ISSUE #212: How to delete / replace backups
Hi,

I have a feeling that this is a really stupid question, but I haven't managed to find a conclusive answer in the docs:

How does one go about deleting or overwriting a backup object?
Can it just be deleted from S3 (in our case)?

An `ark backup delete` subcommand would be useful.

Thanks,

Phil

skriss:

Fixes #212 
Fixes #77 

@ncdc PTAL and provide feedback before I fix unit tests. I've manually tested successfully.

### DIFF
--- a/pkg/cmd/cli/backup/delete.go
+++ b/pkg/cmd/cli/backup/delete.go
@@ -20,11 +20,14 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	api "github.com/heptio/ark/pkg/apis/ark/v1"
 	"github.com/heptio/ark/pkg/client"
 	"github.com/heptio/ark/pkg/cmd"
+	"github.com/heptio/ark/pkg/controller"
+	kubeutil "github.com/heptio/ark/pkg/util/kube"
 )
 
 func NewDeleteCommand(f client.Factory, use string) *cobra.Command {
@@ -35,6 +38,16 @@ func NewDeleteCommand(f client.Factory, use string) *cobra.Command {
 			if len(args) != 1 {
 				c.Usage()
 				os.Exit(1)
+			}
+
+			kubeClient, err := f.KubeClient()
+			cmd.CheckError(err)
+
+			serverVersion, err := kubeutil.ServerVersion(kubeClient.Discovery())
+			cmd.CheckError(err)
+
+			if !serverVersion.AtLeast(controller.MinVersionForDelete) {
+				cmd.CheckError(errors.Errorf("this command requires the Kubernetes server version to be at least %s", controller.MinVersionForDelete))
 			}
 
 			arkClient, err := f.Client()
--- a/pkg/cmd/cli/backup/delete.go
+++ b/pkg/cmd/cli/backup/delete.go
@@ -27,9 +27,9 @@ import (
 	"github.com/heptio/ark/pkg/cmd"
 )
 
-func NewDeleteCommand(f client.Factory) *cobra.Command {
+func NewDeleteCommand(f client.Factory, use string) *cobra.Command {
 	c := &cobra.Command{
-		Use:   "delete NAME",
+		Use:   fmt.Sprintf("%s NAME", use),
 		Short: "Delete a backup",
 		Run: func(c *cobra.Command, args []string) {
 			if len(args) != 1 {
--- a/pkg/controller/gc_controller.go
+++ b/pkg/controller/gc_controller.go
@@ -31,6 +31,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/kubernetes/pkg/util/version"
 
 	api "github.com/heptio/ark/pkg/apis/ark/v1"
 	"github.com/heptio/ark/pkg/cloudprovider"
@@ -41,6 +42,13 @@ import (
 )
 
 const gcFinalizer = "gc.ark.heptio.com"
+
+// MinVersionForDelete is the minimum Kubernetes server version that Ark
+// requires in order to be able to properly delete backups (including
+// the associated snapshots and object storage files). This is because
+// Ark uses finalizers on the backup CRD to implement garbage-collection
+// and deletion.
+var MinVersionForDelete = version.MustParseSemantic("1.7.5")
 
 // gcController removes expired backup content from object storage.
 type gcController struct {
--- a/pkg/controller/gc_controller.go
+++ b/pkg/controller/gc_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"encoding/json"
 	"time"
 
 	"github.com/pkg/errors"
@@ -25,7 +26,9 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/clock"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 
@@ -36,6 +39,8 @@ import (
 	listers "github.com/heptio/ark/pkg/generated/listers/ark/v1"
 	"github.com/heptio/ark/pkg/util/kube"
 )
+
+const gcFinalizer = "gc.ark.heptio.com"
 
 // gcController removes expired backup content from object storage.
 type gcController struct {
@@ -70,7 +75,7 @@ func NewGCController(
 		syncPeriod = time.Minute
 	}
 
-	return &gcController{
+	c := &gcController{
 		backupService:       backupService,
 		snapshotService:     snapshotService,
 		bucket:              bucket,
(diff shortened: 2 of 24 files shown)


pkg/controller/gc_controller.go
```diff
@@ -84,9 +86,63 @@ func NewGCController(
 		restoreClient:       restoreClient,
 		logger:              logger,
 	}
-}
 
-var _ Interface = &gcController{}
+	backupInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(old interface{}, new interface{}) {
```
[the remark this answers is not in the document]
ncdc: Since `new` is a reserved keyword, it might be better not to use it. Looking at the kube source, I've seen
- `func(oldObj, newObj interface{})`
- `func(old, cur interface{})`

I think I prefer the first form.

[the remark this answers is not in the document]
ncdc: Also since you're not using `old`, you can just do `func(_, newObj interface{})`

pkg/controller/gc_controller.go
```diff
@@ -84,9 +86,63 @@ func NewGCController(
...
+				)
+
+				// we're only interested in backups that have a deletionTimestamp and at
+				// least one finalizer.
+				if backup.DeletionTimestamp == nil || len(backup.Finalizers) == 0 {
+					return
+				}
+				log.Debugf("Backup has finalizers %s", backup.Finalizers)
+
+				for x, finalizer := range backup.Finalizers {
```
ncdc: I'd extract this to a function that just returns a bool. No need for idx.

pkg/controller/gc_controller.go
```diff
@@ -84,9 +86,63 @@ func NewGCController(
...
+
+				backup.Finalizers = append(backup.Finalizers[0:idx], backup.Finalizers[idx+1:]...)
+
+				upd, err := backupClient.Backups(api.DefaultNamespace).Update(backup)
+				if err != nil {
+					log.WithError(errors.WithStack(err)).Error("Error updating backup")
+					return
+				}
+
+				// no more finalizers: we don't need to issue another delete
```
ncdc: Shouldn't this be "There are other finalizers, so we won't issue a delete as other controllers still need to process" along with a len >= 1 check?

pkg/controller/gc_controller.go
ncdc: And then maybe make another helper that returns a new slice w/the finalizer removed. See e.g. https://github.com/kubernetes/kubernetes/blob/722cb7a758d4678e7ed5c4f09ded99d6bc5b0b4c/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/helpers.go#L109-L118

pkg/controller/gc_controller.go
skriss: Not 100% sure on how this is supposed to work. I'm going off of https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#finalizers which says:

```
...Each controller then removes its finalizer from the list and issues the delete request again. This request only deletes the object if the list of finalizers is now empty, meaning all finalizers are done.
```

However, the behavior I was seeing was that if I issued a delete after removing the last/only finalizer, it would fail with object not found; the object appeared to be deleted once the last finalizer had been removed.

pkg/controller/gc_controller.go
```diff
@@ -84,9 +88,94 @@ func NewGCController(
...
+			"finalizers": except(backup.Finalizers, gcFinalizer),
+		},
+	}
+
+	patchBytes, err := json.Marshal(patchMap)
+	if err != nil {
+		log.WithError(err).Error("Error marshaling finalizers patch")
+	}
+
+	upd, err := c.backupClient.Backups(api.DefaultNamespace).Patch(backup.Name, types.MergePatchType, patchBytes)
```
ncdc: s/upd/updatedBackup/ or something like that

pkg/controller/gc_controller.go
```diff
@@ -84,9 +88,94 @@ func NewGCController(
...
+
+	log.Infof("Garbage-collecting backup")
+	if err := c.garbageCollect(backup, log); err != nil {
+		log.WithError(err).Error("Error deleting backup's related objects")
+		return
+	}
+
+	patchMap := map[string]interface{}{
+		"metadata": map[string]interface{}{
+			"finalizers": except(backup.Finalizers, gcFinalizer),
```
ncdc: SIG api machinery suggested watching out for some edge cases:
- accidentally removing a finalizer that someone else just added (it's in etcd but our copy of the backup from the shared informer hasn't caught up yet)
- accidentally re-adding a finalizer that someone else just removed (same reason as above)

They suggested adding the backup's current `resourceVersion` to the patchMap.

pkg/controller/gc_controller.go
```diff
@@ -84,9 +88,94 @@ func NewGCController(
...
+	if backup.DeletionTimestamp == nil || len(backup.Finalizers) == 0 {
+		return
+	}
+	log.Debugf("Backup has finalizers %s", backup.Finalizers)
+
+	// GC finalizer not found
+	if !has(backup.Finalizers, gcFinalizer) {
+		return
+	}
+
```
ncdc: SIG api machinery also suggests that we only process our finalizer when it's the first element in the Finalizers slice

pkg/controller/gc_controller.go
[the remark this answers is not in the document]
ncdc: We might want to pull in client-go's `retry` package and use `retry.RetryOnConflict`. Or we could just stop here on conflict and let the next run of the loop try again.

pkg/controller/gc_controller.go
```diff
@@ -84,9 +88,94 @@ func NewGCController(
...
+		log.WithError(err).Error("Error marshaling finalizers patch")
+	}
+
+	upd, err := c.backupClient.Backups(api.DefaultNamespace).Patch(backup.Name, types.MergePatchType, patchBytes)
+	if err != nil {
+		log.WithError(errors.WithStack(err)).Error("Error patching backup")
+		return
+	}
+
+	// no more finalizers: we don't need to issue another delete
```
ncdc: Talked to api machinery: no need to issue additional deletes. The only action we need to perform is the patch to remove our finalizer.

pkg/controller/gc_controller.go
```diff
@@ -84,9 +88,94 @@ func NewGCController(
...
+	}
+	log.Debugf("Backup has finalizers %s", backup.Finalizers)
+
+	// GC finalizer not found
+	if !has(backup.Finalizers, gcFinalizer) {
+		return
+	}
+
+	log.Infof("Garbage-collecting backup")
+	if err := c.garbageCollect(backup, log); err != nil {
```
ncdc: If this call errors once, it's entirely possible (likely even?) that it will continue to error the next time we try it for the same backup. It might be worth considering switching to a rate limited queue (like our other controllers), or keeping track of how many times we've tried to delete (via annotations) and stopping after some fixed limit.

pkg/controller/gc_controller_test.go
```diff
@@ -191,38 +132,50 @@ func TestGarbageCollectBackup(t *testing.T) {
...
+		},
+		{
+			name:                           "nil snapshot service when backup doesn't have snapshots does not error",
+			backup:                         arktest.NewTestBackup().WithName("backup-1").Backup,
+			nilSnapshotService:             true,
+			expectedErr:                    false,
+			expectedObjectStorageDeletions: sets.NewString("backup-1"),
+		},
+		{
+			name: "if snapshot deletion fails, don't delete kube backup and return error",
```
ncdc: We aren't deleting the kube backup from `garbageCollect` any more

pkg/controller/gc_controller_test.go
```diff
@@ -191,38 +132,50 @@ func TestGarbageCollectBackup(t *testing.T) {
...
 		backup                         *api.Backup
-		deleteBackupFile               bool
 		snapshots                      sets.String
 		backupFiles                    sets.String
 		backupMetadataFiles            sets.String
 		restores                       []*api.Restore
+		nilSnapshotService             bool
+		expectedErr                    bool
 		expectedRestoreDeletes         []string
 		expectedBackupDelete           string
```
ncdc: not used any more by this test

pkg/controller/gc_controller_test.go
```diff
@@ -191,38 +132,50 @@ func TestGarbageCollectBackup(t *testing.T) {
...
 		expectedSnapshots              sets.String
 		expectedObjectStorageDeletions sets.String
 	}{
 		{
```
ncdc: We may just want to convert this to a few test cases:
1. nil snapshot server, backup has snapshots, return error (ensure that no DeleteSnapshot, DeleteBackupDir, restore Delete calls are issued (handled by mock.AssertExpectations etc))
1. happy path
1. do as much as you can (some DeleteSnapshot errors & successes, DeleteBackupDir error & success, some restore Delete errors & successes) and return error(s) encountered

ncdc:
Need unit test coverage on handleFinalizer

skriss:
addressed everything except the rate-limited queue comment.

ncdc:
This also fixes #77, doesn't it?

pkg/controller/gc_controller.go
```diff
@@ -84,9 +89,77 @@ func NewGCController(
...
+	)
+
+	// we're only interested in backups that have a deletionTimestamp and at
+	// least one finalizer.
+	if backup.DeletionTimestamp == nil || len(backup.Finalizers) == 0 {
+		return
+	}
+	log.Debugf("Backup has finalizers %s", backup.Finalizers)
+
+	// only process the GC finalizer when it's the first item
```
ncdc: I went through the kube code and can't find any place where it only acts on a finalizer if it's the first element in the slice. Let's go back to proceeding if the slice contains our finalizer, regardless of position. Sorry for the flip-flop.

pkg/controller/gc_controller.go
```diff
@@ -103,104 +176,74 @@ func (c *gcController) Run(ctx context.Context, workers int) error {
...
-// garbageCollectBackups checks backups for expiration and triggers garbage-collection for the expired
-// ones.
-func (c *gcController) garbageCollectBackups(backups []*api.Backup, expiration time.Time, deleteBackupFiles bool) {
-	for _, backup := range backups {
-		if backup.Status.Expiration.Time.After(expiration) {
-			c.logger.WithField("backup", kube.NamespaceAndName(backup)).Info("Backup has not expired yet, skipping")
-			continue
+			restoreLog.Info("Deleting restore referencing backup")
+			if err := c.restoreClient.Restores(restore.Namespace).Delete(restore.Name, &metav1.DeleteOptions{}); err != nil {
+				restoreLog.WithError(errors.WithStack(err)).Error("Error deleting restore")
```
ncdc: Do you think this should just be an error so we consider the GC of the backup a failure?

pkg/controller/gc_controller.go
skriss: np, easy to switch back.

pkg/controller/gc_controller.go
skriss: eh, in my opinion no -- I'm not too worried about orphaned restores (since they don't cost any cloud resources), plus that would have the result of leaving the backup API object around, despite its tarball/etc. potentially having been deleted, which could make a user believe it's still restorable.

ncdc: wfm

skriss:
Re: #77 , it doesn't address it directly but does mean that Ark won't orphan files in obj storage. However, if a user manually deletes a backup API obj, or manually adds files to obj storage, Ark won't be able to clean them up.

pkg/controller/gc_controller.go
skriss: Do you think this is PR-blocking?

ncdc: No

ncdc:
> So the only case we don't handle is files manually added to the bucket outside of Ark

I'm pretty sure BackupService.DeleteBackupDir deletes everything

skriss:
everything within a backup dir, yes. I think we can consider #77 addressed and if some other related issue pops up we can create a new one.

nrb:
Probably worth making a release note for the version, though I don't think we currently have a process for doing so til release time.

pkg/cmd/cli/backup/delete.go
```diff
@@ -37,6 +39,16 @@ func NewDeleteCommand(f client.Factory, use string) *cobra.Command {
...
 			}
 
+			kubeClient, err := f.KubeClient()
+			cmd.CheckError(err)
+
+			serverVersion, err := kubeutil.ServerVersion(kubeClient.Discovery())
+			cmd.CheckError(err)
+
+			if !serverVersion.AtLeast(controller.MinVersionForDelete) {
+				cmd.CheckError(fmt.Errorf("this command requires the Kubernetes server version to be at least %s", controller.MinVersionForDelete))
```
ncdc: Do you want to do `errors.Errorf` for consistency with the rest of the code base? (Apparently we're still using it in pkg/restore/restore.go when adding warnings & errors to the restore result.)

ncdc:
1 minor q

skriss:
sure, updated.

skriss:
@ncdc did you have anything further here? this is ready to merge from my perspective.

ncdc:
Nope, merging

PR #256, opened 2017-12-19
don't error if PVC is not bound to a PV

### LINKED ISSUE #254: Unbound PVC causes a backup to fail
If a PVC has no `spec.volumeName`, it marks the entire backup as failed. A PVC might not have `spec.volumeName` if there is no default `storageclass` and there isn't one specified on the PVC itself, or maybe the desired dynamic provisioner isn't configure yet, to name a couple of reasons. We probably should not treat this as an error condition. WDYT @skriss?

skriss:

Fixes #254

### DIFF
--- a/pkg/backup/backup_pv_action.go
+++ b/pkg/backup/backup_pv_action.go
@@ -17,7 +17,6 @@ limitations under the License.
 package backup
 
 import (
-	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


pkg/backup/backup_pv_action.go
```diff
@@ -55,8 +54,12 @@ func (a *backupPVAction) Execute(item runtime.Unstructured, backup *v1.Backup) (
 	pvc := item.UnstructuredContent()
 
 	volumeName, err := collections.GetString(pvc, "spec.volumeName")
```
ncdc: Consider using `collections.Exists()`?

skriss: I didn't change it because we still use the volumeName if it's there, and under the hood GetString and Exists go through the same codepath anyway.

ncdc: Ok. Hmm looks like as of kube 1.9 they've finally exported the unstructured helper funcs, so we can maybe get rid of our collections package for the most part.

Issue #208, opened 2017-11-20
AWS Quickstart Docs


stevesloka:
Would be nice to have the quick start docs for AWS (https://github.com/heptio/ark/blob/master/docs/cloud-provider-specifics.md#aws) also provide steps on how to create the S3 bucket from command line to follow with the command line steps to create the IAM user / policy. 

Also, would be a good practice to set the IAM policy to only allow access to that bucket and not all of S3.

Bradamant3:
Is this issue closed with PR #259 ?

ncdc:
I think so. @stevesloka could you please look at the latest docs on master for AWS and see if your requests were resolved?

jrnt30:
The second piece of this, limiting the scope of the permissions grants is still outstanding.

ncdc:
Last bit fixed by #363

PR #259, opened 2017-12-21
add create bucket step to cloud provider docs, simplify example with PVs

### LINKED ISSUE #170: Simplify example that uses PVs
In https://github.com/heptio/ark/blob/master/docs/cloud-provider-specifics.md#snapshot-example-with-pvs, we have the user manually label their PV so that it matches the label-selector used in the Backup spec. This is no longer necessary as a result of #65, and we should remove it because it's slightly confusing and is a distraction from the demonstration of Ark functionality. Additionally, we *could* change this to select on the namespace rather than the label-selctor.

### LINKED ISSUE #219: Add docs on how to create a bucket for GCP

### LINKED ISSUE #220: Add docs on how to create a bucket for AWS

skriss:

Fixes #220 
Fixes #219 
Fixes #170

### DIFF
--- a/docs/cloud-provider-specifics.md
+++ b/docs/cloud-provider-specifics.md
@@ -393,7 +393,7 @@ kubectl apply -f examples/nginx-app/base.yaml
 Now create a backup:
 
 ```bash
-ark backup create nginx-backup --selector app=nginx
+ark backup create nginx-backup --include-namespaces nginx-example
 ```
 
 Simulate a disaster:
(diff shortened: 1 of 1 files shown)


skriss:
I haven't modified the permissions we're granting (per #208). We could either merge this first, or wait until I can figure out the appropriate policies for each plat and add.

skriss:
thanks @Bradamant3, rebased.

PR #266, opened 2017-12-27
Adjustment to environment variable for GCP

robbyt:
By setting an environment variable called `CLOUDSDK_CORE_PROJECT` for the GCP project ID, the service account creation will work when you do not have a default project set.

### DIFF
--- a/docs/cloud-provider-specifics.md
+++ b/docs/cloud-provider-specifics.md
@@ -139,7 +139,7 @@ To integrate Heptio Ark with GCP, you should follow the instructions below to cr
     gcloud config list
     ```
 
-    Store the `project` value from the results in the environment variable `$PROJECT_ID`.
+    Store the `project` value from the results in the environment variable `$CLOUDSDK_CORE_PROJECT`.
 
 2. Create a service account:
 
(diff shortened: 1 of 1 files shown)


docs/cloud-provider-specifics.md
```diff
@@ -139,7 +139,7 @@ To integrate Heptio Ark with GCP, you should follow the instructions below to cr
     gcloud config list
     ```
 
-    Store the `project` value from the results in the environment variable `$PROJECT_ID`.
+    Store the `project` value from the results in the environment variable `$CLOUDSDK_CORE_PROJECT`.
```
ncdc: @Bradamant3 fyi, it looks like GCP's project precedence is 1) `--project` flag, 2) `$CLOUDSDK_CORE_PROJECT` env var, 3) preconfigured default project (`gcloud config set core/project PROJECT`).

Do you think we should just use this env var? It has the effect of overriding the preconfigured default project, which may be undesirable or unexpected (based on our docs)(?). WDYT?

skriss:
I'd vote to leave the docs as-is rather than setting a global gcloud env var. @Bradamant3 any objections? If not, we'll close this out.

@robbyt appreciate the input regardless!

Bradamant3:
Sorry to have missed this one for so long. Agree with @ncdc and @skriss -- we document one flow, and it should be the closest thing we can determine to The Right Way. This also seems like a workaround to catch the case where you don't do what you should probably be doing in GCP (which isn't on Ark to take care of). I understand the desire to catch a potential GCP issue, though.

PR #268, opened 2018-01-02
update copyright header to 2018 and make language/casing consistent

skriss:


ncdc:
For non-generated files, I wouldn't change the year.

skriss:
OK. I'll just update the casing for those.

Issue #20, opened 2017-08-09
Backup progress


kstewart:
Provide a way for users to see the progress of an in-flight backup. Some thoughts:

- if backing up all namespaces, first get a list of all namespaces so we know how many there are to process
- try to find out how many different types of GroupResources are being backed up, so we can record progress per namespace
- periodically record status somewhere in backup.status

ncdc:
Idea: store per-backup log file to object storage. Add `ark backup log` command to retrieve it.

ncdc:
Note, backup logs are separate from progress. We'll be coming up with ways to track real-time progress as described above to close out this issue.

skriss:
For backups, we process first by resource then by namespace.
For restores, we process first by namespace then by resource.

I think for progress reporting, we should not be tightly coupled to the current mode/order of processing since that could change. We can count the number of resource-namespace combinations and then report as each pair gets completed.

My initial thought is something like:
```
// OperationProgress describes the overall progress for a backup
// or restore operation.
//
// BackupStatus and RestoreStatus will each have a field of this type.
type OperationProgress struct {
	// PercentComplete is a weighted sum of individual items' progress.
	// Each item has a weight of 1/len(Items), and a value equal to the
	// item's PercentComplete. This is provided as a useful summary
	// of the more detailed item progress information.
	PercentComplete int

	// Items is the individual units of work that make up the
	// operation (currently defined as namespace-resource pairs).
	Items []ItemProgress
}

// ItemProgress describes the current backup/restore progress for
// a namespace-resource pair.
type ItemProgress struct {
	Namespace string
	Resource  string

	// PercentComplete is computed simply as (# items processed)
	// divided by (total # items for namespace-resource pair)
	PercentComplete int
}
```

skriss:
@ncdc let me know if you have any thoughts on this

ncdc:
@skriss it would be nice if operationProgress.percentComplete were more accurate, based on the percentages of the individual ItemProgresses.

When we do a backup, we know up front how many different types of resources we have, and how many items we have per resource type. If we store that information in backup.status, we can use it when restoring. Doing it that way, we could simply use % complete per resource type, and not worry about namespace.

ncdc:
We could store a map of GroupResource string to

```go
type ResourceStatus struct {
  Processed int
  Total int
}
```

It could look like this:

```yaml
resourceProgress:
  pods:
    processed: 15
    total: 100
  storageclasses.storage.k8s.io:
[...]

PR #270, opened 2018-01-04
Clarify backup and restore creation messages

### LINKED ISSUE #232: Clarify text printed after you create a backup or restore
After you run `ark backup create` or `ark backup restore`, the responses are `Backup x created successfully.` and `Restore x created successfully.`. These are misleading, as it might sound like the backup/restore is done, when in fact it might still be in progress.

Maybe `Backup/Restore x submitted successfully.` would be better, possibly along with some text indicating that they can run `ark backup/restore get` or `ark backup/restore describe` for more details.

nrb:
When running `ark <resource> create`, a request is sent to the server,
but the status is not immediately known. Inform the user that a request
was sent and provide a way to get more information on it.

I'm open to changing the message text if others think of something better.

Fixes #232

### DIFF
--- a/pkg/cmd/cli/backup/create.go
+++ b/pkg/cmd/cli/backup/create.go
@@ -143,6 +143,7 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
 		return err
 	}
 
-	fmt.Printf("Backup %q created successfully.\n", backup.Name)
+	fmt.Printf("Backup request %q submitted successfully.\n", backup.Name)
+	fmt.Printf("Run `ark backup describe %s` for more details.\n", backup.Name)
 	return nil
 }
--- a/pkg/cmd/cli/restore/create.go
+++ b/pkg/cmd/cli/restore/create.go
@@ -144,6 +144,7 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
 		return err
 	}
 
-	fmt.Printf("Restore %q created successfully.\n", restore.Name)
+	fmt.Printf("Restore request %q submitted successfully.\n", restore.Name)
+	fmt.Printf("Run `ark restore describe %s` for more details.\n", restore.Name)
 	return nil
 }


ncdc:
cc @Bradamant3

ncdc:
Jennifer plans to review this tomorrow

pkg/cmd/cli/backup/create.go
```diff
@@ -143,6 +143,7 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
 		return err
 	}
 
-	fmt.Printf("Backup %q created successfully.\n", backup.Name)
+	fmt.Printf("Backup request %q submitted successfully.\n", backup.Name)
+	fmt.Printf("Use `ark backup describe %s` for more details.\n", backup.Name)
```
Bradamant3: Nit: s/Use/Run
(same for restore)
reason: "use" is vague, "run" is more precise.

Bradamant3: Also, maybe this is my ignorance showing, but does `ark backup describe ...` provide additional information about the status of the backup request? If so, maybe let's ponder further ...

ncdc: What we have now is `ark backup get` and `ark backup describe`. `get` shows you the basic details in tabular format (the essential info being the phase - new/in progress/failed/completed). `describe` shows you slightly more information, but it's mostly the `spec` - what namespaces/resources/etc are included in the backup. Until we have #20 implemented, there's not really any info you can get other than seeing phase transitions.

Bradamant3: OK, good as is then.

nrb:
@Bradamant3 Here's what `backup describe`'s output looks like currently.

```
x1c in /home/nrb/go/src/github.com/heptio/ark (git) fix-246 U
% _output/bin/linux/amd64/ark backup create backup-1
Backup "backup-1" created successfully.

x1c in /home/nrb/go/src/github.com/heptio/ark (git) fix-246 U
% _output/bin/linux/amd64/ark backup describe backup-1
Name:         backup-1
Namespace:    heptio-ark
Labels:       <none>
Annotations:  <none>

Namespaces:
  Included:  *
  Excluded:  <none>

Resources:
  Included:        *
  Excluded:        <none>
  Cluster-scoped:  auto

Label selector:  <none>

Snapshot PVs:  auto

TTL:  720h0m0s

Hooks:  <none>

Phase:  Completed

Backup Format Version:  1

Expiration:  2018-02-04 12:39:13 -0500 EST

Validation errors:  <none>

Persistent Volumes: <none included>
```

nrb:
@Bradamant3 Updated the messages

pkg/cmd/cli/restore/create.go
```diff
@@ -144,6 +144,7 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
 		return err
 	}
 
-	fmt.Printf("Restore %q created successfully.\n", restore.Name)
+	fmt.Printf("Restore request for %q submitted successfully.\n", restore.Name)
```
skriss: I feel like this should either say "Restore request <restore.Name>", or "Restore request for <o.BackupName>" - WDYT?

ncdc: I think if/until we change to `ark restore create <name> --backup <backup name>`, the output here must include at least the restore's generated name. Since that includes the backup's name, I think the former (just removing "for") is probably better.

nrb: Ok, can do that. I used “for” as I was viewing it was the request for the restore object to be created. However, the request isn’t really an object in Ark so I suppose it makes sense to drop the “for”.

nrb: Message updated.

ncdc:
@Bradamant3 please take a final look - thanks!

PR #272, opened 2018-01-05
Support running in any namespace

### LINKED ISSUE #106: Allow Ark server to run in arbitrary namespaces
Currently the Ark server component assumes that it's configuration is stored in api.DefaultNamespace (heptio-ark) with no way to override. As cluster admins we put these kind of tools into a 'utils' namespace and would like to do the same for Ark.

The server component should probably assume it's configuration is stored in the namespace it's deployed in. The client would need to support a namespace flag or configuration or both.

ncdc:
Support running in any namespace

Add the ability for the Ark server to run in any namespace.

Add `ark client config get/set` for manipulating the new client
configuration file in $HOME/.config/ark/config.json. This holds client
defaults, such as the Ark server's namespace (to avoid having to specify
the --namespace flag all the time).

Add a --namespace flag to all client commands.

This also upgrades us to client-go v6.0.0.

Fixes #106


pkg/client/config.go
```diff
@@ -0,0 +1,75 @@
...
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	ConfigKeyNamespace = "namespace"
+)
+
+// LoadConfig loads the Ark client configuration file and returns it as a map[string]string. If the
```
ncdc: Should we consider using something like Viper instead?

pkg/cmd/cli/client/client.go
```diff
@@ -14,16 +14,22 @@ See the License for the specific language governing permissions and
...
-
-	"github.com/heptio/ark/pkg/client"
 )
 
-func NewGetCommand(f client.Factory) *cobra.Command {
-	c := &cobra.Command{}
+func NewCommand() *cobra.Command {
+	c := &cobra.Command{
+		Use:   "client",
+		Short: "Ark client related commands",
```
ncdc: @Bradamant3 could use some word-smithing here.

pkg/cmd/cli/client/config/config.go
```diff
@@ -18,20 +18,17 @@ package config
...
-	"github.com/heptio/ark/pkg/client"
 )
 
-func NewCommand(f client.Factory) *cobra.Command {
+func NewCommand() *cobra.Command {
 	c := &cobra.Command{
 		Use:   "config",
-		Short: "Work with config",
-		Long:  "Work with config",
+		Short: "Get and set client configuration file values",
```
ncdc: @Bradamant3 could use some word-smithing here. Is it worthwhile to have `Long` and `Example` as well?

pkg/cmd/cli/client/config/get.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/heptio/ark/pkg/client"
+	"github.com/heptio/ark/pkg/cmd"
+	"github.com/spf13/cobra"
+)
+
+func NewGetCommand() *cobra.Command {
+	c := &cobra.Command{
+		Use:   "get [KEY 1] [KEY 2] [...]",
+		Short: "Get client configuration file values",
```
ncdc: @Bradamant3 could use some word-smithing here. Is it worthwhile to have `Long` and `Example` as well?

pkg/cmd/cli/client/config/set.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/heptio/ark/pkg/client"
+	"github.com/heptio/ark/pkg/cmd"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func NewSetCommand() *cobra.Command {
+	c := &cobra.Command{
+		Use:   "set KEY=VALUE [KEY=VALUE]...",
+		Short: "Set client configuration file values",
```
ncdc: @Bradamant3 could use some word-smithing here. Is it worthwhile to have `Long` and `Example` as well?

pkg/cmd/server/server.go
```diff
@@ -86,7 +87,21 @@ func NewCommand() *cobra.Command {
 			logger := newLogger(logLevel, &logging.ErrorLocationHook{}, &logging.LogLocationHook{})
 			logger.Infof("Starting Ark server %s", buildinfo.FormattedGitSHA())
 
-			s, err := newServer(kubeconfig, fmt.Sprintf("%s-%s", c.Parent().Name(), c.Name()), logger)
+			// NOTE: the namespace flag is bound to ark's persistent flags when the root ark command
```
ncdc: @skriss @nrb let me know if you think we should make this use Factory

pkg/client/factory.go
```diff
@@ -49,7 +55,19 @@ func NewFactory(baseName string) Factory {
 		flags:    pflag.NewFlagSet("", pflag.ContinueOnError),
 		baseName: baseName,
 	}
+
+	if config, err := LoadConfig(); err == nil {
+		f.namespace = config[ConfigKeyNamespace]
+	} else {
+		fmt.Fprintf(os.Stderr, "WARNING: error retrieving namespace from config file: %v\n", err)
```
skriss: this will now warn every time if a user hasn't set up a config file, right? i'd say this is probably undesirable since the majority of folks will just continue running with the default.

pkg/client/factory.go
```diff
@@ -49,7 +55,19 @@ func NewFactory(baseName string) Factory {
...
+	}
+
+	if f.namespace == "" {
+		f.namespace = v1.DefaultNamespace
+	}
+
 	f.flags.StringVar(&f.kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use to talk to the Kubernetes apiserver. If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration")
+	f.flags.StringVarP(&f.namespace, "namespace", "n", f.namespace, "TODO")
```
skriss: TODO :)

pkg/cmd/cli/client/config/get.go
```diff
@@ -0,0 +1,57 @@
+/*
+Copyright 2017 the Heptio Ark contributors.
```
skriss: 2018

pkg/cmd/cli/client/config/get.go
```diff
@@ -0,0 +1,57 @@
...
+		Use:   "get [KEY 1] [KEY 2] [...]",
+		Short: "Get client configuration file values",
+		Run: func(c *cobra.Command, args []string) {
+			config, err := client.LoadConfig()
+			cmd.CheckError(err)
+
+			if len(args) == 0 {
+				keys := make([]string, 0, len(config))
+				for key := range config {
+					keys = append(keys, key)
```
skriss: why not just Printf in this loop and get rid of the next one?

pkg/cmd/cli/client/config/set.go
```diff
@@ -0,0 +1,62 @@
+/*
+Copyright 2017 the Heptio Ark contributors.
```
skriss: 2018

skriss:
also, all the references to v1.DefaultNamespace are accounted for, right?

pkg/cmd/util/downloadrequest/downloadrequest.go
```diff
@@ -47,18 +47,18 @@ func Stream(client arkclientv1.DownloadRequestsGetter, name string, kind v1.Down
...
-	req, err := client.DownloadRequests(v1.DefaultNamespace).Create(req)
+	req, err := client.DownloadRequests(namespace).Create(req)
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	defer client.DownloadRequests(v1.DefaultNamespace).Delete(req.Name, nil)
+	defer client.DownloadRequests(namespace).Delete(req.Name, nil)
 
 	listOptions := metav1.ListOptions{
 		//TODO: once kube-apiserver http://issue.k8s.io/51046 is fixed, uncomment
```
skriss: feel like fixing this while you're in here?

pkg/client/factory.go
ncdc: No, see the godoc/impl for LoadConfig

pkg/cmd/cli/client/config/get.go
ncdc: This was actually a move from a previous file, but I can adjust it.

pkg/cmd/cli/client/config/get.go
ncdc: Hah because I forgot to put the sort in

pkg/cmd/util/downloadrequest/downloadrequest.go
ncdc: I can update the comment, but we can't use a field selector until Kubernetes v1.9.0 is the minimum supported version.

ncdc:
> also, all the references to v1.DefaultNamespace are accounted for, right?

Yes. It's only used in tests and in the couple of places where we fall back to it as a default if it's not set elsewhere.

pkg/client/factory.go
skriss: OK, so this would only print if the file exists but does not contain namespace. Could be an issue in the future if we add additional fields to client config (so a user has another field set but not namespace), but fine for now.

pkg/cmd/util/downloadrequest/downloadrequest.go
skriss: good point :)

skriss:
I guess we should probably add a doc explaining that it's possible to run Ark outside the `heptio-ark` namespace and how to do it (adjust NS in all of the YAMLs, deploy, set CLI flag or config)

pkg/client/factory.go
ncdc: No, it will only print if there is an error opening the file or decoding its contents as json to a map[string]string.

ncdc:
Do we want to tackle `ark install` now? I'm still tempted to say no, because a good amount will change with backup targets. But it would handle the variable namespace issue.

skriss:
I think it makes sense to wait until we're a little more stable there.

pkg/client/config.go
skriss: I'd say keep it in mind for the future but not needed now.

pkg/client/factory.go
skriss: d'oh. brain's still on vacation!

skriss:
Can go ahead and squash here, looks like you need to re-gen docs too.

ncdc:
Done

skriss:
LGTM. merging.

PR #273, opened 2018-01-05
remove outdated line about backups removing PV annotations

skriss:

As of #263 we no longer remove PV annotations. Remove statement in doc that says we do.

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -5,7 +5,7 @@
     * [1. Backups][2]
     * [2. Schedules][3]
     * [3. Restores][4]
-* [API types][9]
+* [API types][8]
 * [Expired backup deletion][5]
 * [Cloud storage sync][6]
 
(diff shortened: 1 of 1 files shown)


docs/concepts.md
```diff
@@ -20,7 +20,7 @@ Ark is ideal for the disaster recovery use case, as well as for snapshotting you
 This section gives a quick overview of the Ark operation types.
 
 ### 1. Backups
-The *backup* operation (1) uploads a tarball of copied Kubernetes resources into cloud object storage and (2) uses the cloud provider API to make disk snapshots of persistent volumes, if specified. [Annotations][8] are cleared for PVs but kept for all other object types.
```
ncdc: It doesn't look like we reference url [8] anywhere any more. Remove it?

skriss: done.

ncdc: Oh, I didn't mean you needed to renumber, but ok

PR #285, opened 2018-01-16
Update s3 bucket creation docs for us-east-1

### LINKED ISSUE #284: Location constraint fails if `us-east-1` is envoked
### Issue: 
When following the [AWS specific markdown](https://github.com/heptio/ark/blob/master/docs/cloud-provider-specifics.md#s3-bucket-creation) instructions to create a bucket, the command fails if the `us-east-1` region is called.

```
$ aws s3api create-bucket --bucket foo --region us-east-1 --create-bucket-configuration LocationConstraint=us-east-1

An error occurred (InvalidLocationConstraint) when calling the CreateBucket operation: The specified location-constraint is not valid
```

### Expected result: 
Bucket creation.

### Actual result:
Bucket creation fails with `invalidLocationConstraint`

### Steps to remedy: 
Remove the `create-bucket-configuration` flag when `us-east-1` is the desired region.  I realize this error is upstream in the aws-cli codebase, but instructions to circumvent this error may be helpful for others.

I'm submitting a small pr to notate this in the markdown.  Thanks!

lypht:
Fixes #284. (This actually only addresses the issue within documentation, but the included command allows a user with us-east-1 to resume.

### DIFF
--- a/docs/cloud-provider-specifics.md
+++ b/docs/cloud-provider-specifics.md
@@ -31,6 +31,14 @@ aws s3api create-bucket \
     --region <YOUR_REGION> \
     --create-bucket-configuration LocationConstraint=<YOUR_REGION>
 ```
+Note: us-east-1 does not support a `LocationConstraint`.  If your region is `us-east-1`, omit the bucket configuration:
+
+```bash
+
+aws s3api create-bucket \
+    --bucket <YOUR_BUCKET> \
+    --region us-east-1
+```
 
 #### IAM user creation
 
@@ -462,4 +470,3 @@ ark restore create nginx-backup
 [20]: /CHANGELOG.md
 [21]: /docs/build-from-scratch.md
 [22]: https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#prerequisites_for_using_role-based_access_control
-


ncdc:
@lypht thanks for your contribution! Would you please sign your commit (https://github.com/heptio/ark/blob/master/CONTRIBUTING.md / `git commit --amend --no-edit --signoff && git push --force` should do it)?

lypht:
Thanks for improving the title of the PR.  Definitely in the weeds this AM 👍

docs/cloud-provider-specifics.md
```diff
@@ -31,6 +31,14 @@ aws s3api create-bucket \
     --region <YOUR_REGION> \
     --create-bucket-configuration LocationConstraint=<YOUR_REGION>
 ```
+Note: us-east-1 does not require a `location-constraint`.  If your region is `us-east-1`, omit the bucket configuration:
```
nrb: This phrasing reads as if `location-constraint` is optional, but I believe the issue is that using it causes creation to fail.

Perhaps something like "us-east-1 does not support a `location-constraint`".  Any thoughts @Bradamant3?

lypht: Location-constraint is optional in the sense that the command with the flag removed successfully creates the bucket; however, not having the constraint may break a downstream dependency.

```
$ aws s3api create-bucket --bucket bar --region us-east-1
{
    "Location": "/bar"
}
```

lypht: Ah, @nrb, I now savvy your semantic distinction.  I'll edit my pr.

docs/cloud-provider-specifics.md
```diff
@@ -31,6 +31,14 @@ aws s3api create-bucket \
     --region <YOUR_REGION> \
     --create-bucket-configuration LocationConstraint=<YOUR_REGION>
 ```
+Note: us-east-1 does not support a `location-constraint`.  If your region is `us-east-1`, omit the bucket configuration:
```
ncdc: Should location-constraint be LocationConstraint?

docs/cloud-provider-specifics.md
```diff
@@ -31,6 +31,14 @@ aws s3api create-bucket \
     --region <YOUR_REGION> \
     --create-bucket-configuration LocationConstraint=<YOUR_REGION>
 ```
+Note: us-east-1 does not support a `LocationConstraint`.  If your region is `us-east-1`, omit the bucket configuration:
+
+```bash
+
+aws s3api create-bucket \
+    --bucket <YOUR_BUCKET> \
+    --region <YOUR_REGION>
```
ncdc: Sorry for another comment here... but do you think that we should have `us-east-1` here instead of `<YOUR_REGION>`?

ncdc:
LGTM. Please squash down to 1 commit - thanks!

PR #294, opened 2018-02-02
Create the config directory in case it's missing

ncdc:
Always try to create the config directory when saving the client config
in case it doesn't exist.


skriss:
LGTM.

PR #296, opened 2018-02-03
Adds a kubecontext global flag

### LINKED ISSUE #207: Support specifying kube context
Other tools like helm / stern allow specifying which context to use from `~/.kube/config`, this is especially helpful if you have many clusters defined in a single config.

Example of desired usage:
`ark --kube-context staging create backup foo-01`

blakebarnett:
Fixes #207

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -30,10 +30,12 @@ import (
 
 // Config returns a *rest.Config, using either the kubeconfig (if specified) or an in-cluster
 // configuration.
-func Config(kubeconfig, baseName string) (*rest.Config, error) {
-	loader := clientcmd.NewDefaultClientConfigLoadingRules()
-	loader.ExplicitPath = kubeconfig
-	clientConfig, err := clientcmd.BuildConfigFromKubeconfigGetter("", loader.Load)
+func Config(kubeconfig, kubecontext, baseName string) (*rest.Config, error) {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	loadingRules.ExplicitPath = kubeconfig
+	configOverrides := &clientcmd.ConfigOverrides{CurrentContext: kubecontext}
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+	clientConfig, err := kubeConfig.ClientConfig()
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
--- a/pkg/client/factory.go
+++ b/pkg/client/factory.go
@@ -43,10 +43,11 @@ type Factory interface {
 }
 
 type factory struct {
-	flags      *pflag.FlagSet
-	kubeconfig string
-	baseName   string
-	namespace  string
+	flags       *pflag.FlagSet
+	kubeconfig  string
+	kubecontext string
+	baseName    string
+	namespace   string
 }
 
 // NewFactory returns a Factory.
@@ -68,6 +69,7 @@ func NewFactory(baseName string) Factory {
 
 	f.flags.StringVar(&f.kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use to talk to the Kubernetes apiserver. If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration")
 	f.flags.StringVarP(&f.namespace, "namespace", "n", f.namespace, "The namespace in which Ark should operate")
+	f.flags.StringVar(&f.kubecontext, "kubecontext", "", "The context to use to talk to the Kubernetes apiserver. If unset defaults to whatever your current-context is (kubectl config current-context)")
 
 	return f
 }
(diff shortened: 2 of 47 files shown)


pkg/client/factory.go
```diff
@@ -66,8 +67,9 @@ func NewFactory(baseName string) Factory {
 		f.namespace = v1.DefaultNamespace
 	}
 
-	f.flags.StringVar(&f.kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use to talk to the Kubernetes apiserver. If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration")
 	f.flags.StringVarP(&f.namespace, "namespace", "n", f.namespace, "The namespace in which Ark should operate")
+	f.flags.StringVar(&f.kubecontext, "kubecontext", "", "The context to use to talk to the Kubernetes apiserver. If unset defaults to whatever your current-context is (kubectl config current-context)")
```
ncdc: @Bradamant3 pls review the text here

pkg/client/factory.go
```diff
@@ -66,8 +67,9 @@ func NewFactory(baseName string) Factory {
...
+	f.flags.StringVar(&f.kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use to talk to the Kubernetes apiserver. If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration")
```
ncdc: @blakebarnett any particular reason you moved this down?

pkg/client/client.go
```diff
@@ -30,10 +30,13 @@ import (
...
 // Config returns a *rest.Config, using either the kubeconfig (if specified) or an in-cluster
 // configuration.
-func Config(kubeconfig, baseName string) (*rest.Config, error) {
-	loader := clientcmd.NewDefaultClientConfigLoadingRules()
-	loader.ExplicitPath = kubeconfig
-	clientConfig, err := clientcmd.BuildConfigFromKubeconfigGetter("", loader.Load)
+func Config(kubeconfig, kubecontext, baseName string) (*rest.Config, error) {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	loadingRules.ExplicitPath = kubeconfig
+	configOverrides := &clientcmd.ConfigOverrides{}
```
ncdc: Since you're only setting 1 field in here, how about this?
```go
configOverrides := &clientcmd.ConfigOverrides{CurrentContext: kubecontext}
```

ncdc:
@blakebarnett thanks for the PR! Would you please sign your commit, and also run `gofmt` on the files you changed? Thanks!

pkg/client/client.go
blakebarnett: 👍

pkg/client/factory.go
blakebarnett: nope, I must have done that when I rebased, will fix.

blakebarnett:
Ah I see, I need to actually update the docs to include, I'll take care of that in the next commit.

ncdc:
@blakebarnett looking pretty good! Just run `make update` to update everything (docs).

pkg/client/factory.go
```diff
@@ -68,6 +69,7 @@ func NewFactory(baseName string) Factory {
 
 	f.flags.StringVar(&f.kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use to talk to the Kubernetes apiserver. If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration")
```
Bradamant3: This part of the message isn't clear:
"If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration"
Are these meant to be alternatives to setting the path? Do you need both? Is either of them the default if not set? Or would you have to try one or the other for the command to succeed? In other words, does not specifying result in an error condition?
Or does the command look for a KUBECONFIG environment variable, and if not found then look for the path in the cluster config?
If it's the latter, the fix is fairly simple:
"If unset, tries the environment variable KUBECONFIG, and if not present, tries in-cluster configuration"
If not the latter, questions ^^ might help with clearer text.

ncdc: @Bradamant3 we can address the `kubeconfig` text in another issue. This PR is about `kubecontext` 😄

ncdc:
@blakebarnett FYI, we'll pick this up as soon as master is open for v0.8.0 later this week. Thanks for your patience!

ncdc:
@blakebarnett we'd like to get this in sometime soon. Could you please squash & sign your commit?

blakebarnett:
I signed it, I thought GH would auto-squash on merge?

ncdc:
You currently have 4 commits, and each one must have the DCO stanza in it. I don't see it in any of them?

blakebarnett:
sorry, never used GH's signing stuff

ncdc:
DCO is good, thanks!

nrb:
Tested this, can confirm that it works as described.

bryanlarsen:
Ugh, helm uses `--kube-context`, kubectl uses `--context` and now ark uses `--kubecontext`.  I'm probably going to get it wrong every single time.

ncdc:
We could probably support both `--kubecontext` and `--kube-context`. I think `--context` is potentially too generic for Ark, though.

blakebarnett:
Yeah, I did it this way because it already had `--kubeconfig` but I'd prefer if it matched helm also.

PR #301, opened 2018-02-06
document ability to change namespace

Bradamant3:
Fixes issue #290 
with apologies for messy commit history, but my local log isn't giving me enough info to squash or remove commits with confidence. Can clean up before merge with some coaching. Figure it's better to get review going first.
Do we want to provide this information elsewhere in the docs? It'll go in the release notes (CHANGELOG.md), but if y'all have suggestions for other locations I'm open to considering them. I pondered the README, but there's currently no obvious place for it there.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,0 +1,1 @@
+remote_theme: tomjoht/documentation-theme-jekyll
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,1 +1,2 @@
-remote_theme: tomjoht/documentation-theme-jekyll
+theme: jekyll-theme-minimal
+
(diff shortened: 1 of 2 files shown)


ncdc:
Other items that need documenting:

1. We've added `ark client config set namespace=<your namespace>` so that when you're running the `ark` cli, it can remember your namespace instead of you having to do `ark [command] -n <your namespace>` all the time
1. The example yaml files are all hard-coded to `heptio-ark`. The forthcoming `ark install` command will make that easier, but until we have it, we need some documentation that tells people how to install ark into a different namespace

Bradamant3:
Thanks, @ncdc -- this is the direction that I needed.

Bradamant3:
Closing this one unmerged bc need to restage multiple other commits for doc changes. Copying comments out for future reference.

PR #302, opened 2018-02-07
Error if backup missing when creating a restore

### LINKED ISSUE #297: User should be informed when restore is invalid
Presently a user could run the following command

`ark restore create threefiddy` which is a backup that doesn't exist.

Immediately the user receives the following feedback

```
Restore request "fiddy-20180203165643" submitted successfully.
Run `ark restore describe fiddy-20180203165643` for more details.
```

Which is true in the technical sense (i.e. the request was successfully completed), but deceiving in that the process will 100% fail because it's a non-existent backup. 

For the sake of UX, the command should at the very least check the backup exists, as a circuit breaker, before a success message is displayed on screen.

ncdc:
Fixes #297


ncdc:
@nrb please review & test

Issue #264, opened 2017-12-22
Document problem with restoring LoadBalancer Services


ncdc:
The way that Kubernetes currently (<=v1.9.x) handles Services with type=LoadBalancer is it generates the name of the associate cloud resource (e.g. ELB) using the following formula:

`a<service UID, dashes removed, up to 31 characters>`, e.g. `a4149f70ee9dc11e68c3bacbc32c1ca8`

`UID` is a field set by the Kubernetes apiserver. It is not possible for a user to specify it.

When Ark restores a LoadBalancer Service, the Service's UID will change from its original value, meaning the identifier for the cloud load balancer resource changes too. If you have a DNS CNAME for yourapp.com that points at the DNS name of the cloud load balancer, it will likely become invalid post-restore, requiring you to update yourapp.com's CNAME pointer.

Need to document this.

TODO: link to Kubernetes issue (if it exists)

ncdc:
Some cloud providers support reattaching to an existing IP via `spec.loadBalancerIP`, but not all of them do (e.g. AWS).

ncdc:
xref:

https://github.com/kubernetes/kubernetes/pull/56832
https://github.com/kubernetes/kubernetes/pull/49903

Bradamant3:
yeah, this is one I'm newly aware of for upstream also. Changing milestone to address sooner. Is anyone aware of newer work upstream to address? As of 1.12 I believe was still an issue (at least, is still documented as such)

ncdc:
It is still an issue. There is https://github.com/kubernetes/kubernetes/pull/66589 which moves the implementation to be cloud provider specific, but it's more of a code reorganization to allow for future changes.

https://github.com/kubernetes/kubernetes/pull/64358 was an attempt to allow users to control the prefix used when naming LBs (in response to https://github.com/kubernetes/kubernetes/issues/62640), but it was closed with a suggestion to propose a KEP first.

skriss:
FIxed by #1019

Issue #245, opened 2017-12-12
Add fsfreeze example to documentation


ncdc:
Note that fsfreeze requires a privileged container.

Suggest that 1 option is to run a sidecar pause container that contains fsfreeze and has the appropriate volumeMount(s).

Show using pre/post hooks once #244 is resolved.

robbyt:
It would be cool if this was written as an operator service, and run as a daemonSet. The service could watch for annotation events to freeze/unfreeze the volumes.

ncdc:
A daemonset might be challenging (it would need access to all the volumes for all the pods it wants to freeze, and it would be difficult to signal easily), but we are thinking that it would be nice to use an annotation+initializer+controller to make this happen automagically. For example:

1. Set an annotation on your pod, e.g. `volume.backup.ark.heptio.com/freeze=true`
1. Our initializer controller would see that on the pod and then:
    1. Set up pre and post hooks to freeze & unfreeze
    1. Add a slim container that has fsfreeze that just runs pause

What do you think about something like that?

robbyt:
That seems simpler from an implementation- but if I'm understanding it all correctly, that might be a bit fragile and prone to race conditions.

Here's why: if you patch the pod resource directly, what's keeping the replica set from editing the pod while the fs is frozen? If you patch the RS, what happens if a deployment is updated during the fs freeze? Also, if you patch the deployment, it causes noise in my audit stream.

I think the even simpler & less magical way is to require users to manually add a sidecar controller to pods that need this feature.

ncdc:
There's not a race that I can foresee. When a Kubernetes object has an initializer, it means that it's not visible to most of the system. Only components that ask for uninitialized objects (such as initializer controllers) will see them. So when the RS controller submits a pod for creation, our initializer controller would intercept it, mutate it to add the annotations & sidecar container, and then once there were no more pending initializers, the pod would be "created" and visible to everyone.

Does that make more sense?

robbyt:
Yes it does make sense- I didn't realize that was possible. Is this k8s feature by design? Could this behavior change in the future?

My recommendation from an end-user's perspective is for this to be designed as two features:
1. easily freeze the file system during snapshots
2. minimize end-user effort required to install ark's necessary components

ncdc:
Initializers is an alpha feature in Kubernetes. The behavior could change in the future given that it's alpha.

[...]

Issue #275, opened 2018-01-05
Document restore flow


ncdc:
Include information about downloading tarball, iterating through api groups, resources, items. Cluster-scoped vs namespace-scoped. Custom actions. Conflicts. Errors/warnings.

eleanor-millman:
Closing until we hear more demand.

Issue #274, opened 2018-01-05
Document backup flow


ncdc:
Include info about discovery, api groups, resources, items, pre/post hooks, custom actions, additional items, etc.

ddebrunner:
Is there any initial documentation about the flow currently, or where to look to try and figure out the flow (e.g. when are pre/post hooks executed when multiple pods exist)?

a-mccarthy:
@nrb and I were just talking about making improvements to the backup docs, https://velero.io/docs/v1.5/how-velero-works/#backup-workflow, to make them more clear. We should make sure that the velero docs cover backups in enough detail

eleanor-millman:
Closing until we hear more demand.

PR #303, opened 2018-02-07
[WIP] address a few doc issues

Bradamant3:
This PR addresses #301 and provides one approach to #264.
It also makes a few small copyedits to `hooks.md`, which needs more attention for #245

### DIFF
--- a/docs/config-definition.md
+++ b/docs/config-definition.md
@@ -2,7 +2,8 @@
 
 * [Overview][8]
 * [Example][9]
-* [Parameter Reference][6]
+* [Install in any namespace][13]
+* [Parameter reference][6]
   * [Main config][7]
   * [AWS][0]
   * [GCP][1]
@@ -12,7 +13,9 @@
 
 Heptio Ark defines its own Config object (a custom resource) for specifying Ark backup and cloud provider settings. When the Ark server is first deployed, it waits until you create a Config--specifically one named `default`--in the `heptio-ark` namespace.
 
-> *NOTE*: There is an underlying assumption that you're running the Ark server as a Kubernetes deployment. If the `default` Config is modified, the server shuts down gracefully. Once the kubelet restarts the Ark server pod, the server then uses the updated Config values.
+In version 0.7.0 and later, you can run Ark in any namespace. In earlier versions, the `heptio-ark` namespace was required.
+
+> **NOTE**: This example assumes that the Ark server is running as a Kubernetes deployment. If you modify the `default` Config, the server shuts down gracefully. The kubelet then restarts the Ark server pod with the updated Config values.
 
 ## Example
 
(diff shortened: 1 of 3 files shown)


docs/config-definition.md
```diff
@@ -38,7 +41,17 @@ scheduleSyncPeriod: 1m
 restoreOnlyMode: false
 ```
 
-## Parameter Reference
+## Install in any namespace
+
+To install Ark in another namespace, you can specify the namespace in your `Config` YAML. That is, you can replace the value `heptio-ark` in the example with the appropriate namespace value.
```
ncdc: It's not just the Config yaml. It's all of the required resources that are created by the example yaml files:
- service account
- cluster-admin cluster role binding for service account
- ark role
- role binding for service account
- ark config
- ark deployment

docs/config-definition.md
```diff
@@ -38,7 +41,17 @@ scheduleSyncPeriod: 1m
...
+
+Or you can run:
```
ncdc: This is **only** for the CLI. This is complementary. If you install the ark server into the `foo` namespace, you would likely want to run `ark client config set namespace=foo` so that any time you run `ark backup create` or any of the other CLI commands, you avoid needing to specify `--namespace=foo`.

docs/debugging-restores.md
```diff
@@ -99,5 +100,15 @@ Both errors and warnings are structured in the same way:
 
 * `Namespaces`: A map of namespaces to the list of issues related to the restore of their respective resources.
 
+## Restoring a load balancer Service
```
ncdc: I would prefer to see this in a "Known Issues" section/page - wdyt?

Bradamant3: yes. that's my preference too, but I'm also used to putting known issues into a changelog/release notes. Is this sufficient?

ncdc: This type of information isn't really appropriate for a changelog or release notes. It doesn't affect a single version.

docs/config-definition.md
Bradamant3: Yeah, this is all also not unrelated to issues #274 and #275, which will take more time. (but which I should be able to get into the release ... just not today). Right now there's no obvious place for any of this stuff, bc the example YAML files are just that, YAML only. Discuss in today's squad sync?

docs/config-definition.md
Bradamant3: At a minimum, will clarify. See also previous comment.

docs/debugging-restores.md
Bradamant3: got it. separate section/page it is

Bradamant3:
Closing because most of these changes were made as part of other pull requests. Some issues remain outstanding, will be addressed in subsequent PR.

PR #304, opened 2018-02-07
Add --wait support to ark backup create

### LINKED ISSUE #233: Support watching/waiting for a backup/restore to finish after creating it
Something like `ark backup/restore create foo --wait`. Could combine with #20 #21 to show progress updates as they occur.

ncdc:
Fixes #233 

### DIFF
--- a/pkg/cmd/cli/backup/create.go
+++ b/pkg/cmd/cli/backup/create.go
@@ -20,10 +20,12 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/heptio/ark/pkg/generated/informers/externalversions/ark/v1"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
 
 	api "github.com/heptio/ark/pkg/apis/ark/v1"
 	"github.com/heptio/ark/pkg/client"
@@ -47,6 +49,7 @@ func NewCreateCommand(f client.Factory, use string) *cobra.Command {
 	}
 
 	o.BindFlags(c.Flags())
+	o.BindWait(c.Flags())
 	output.BindFlags(c.Flags())
 	output.ClearOutputFlagDefault(c)
 
@@ -64,6 +67,7 @@ type CreateOptions struct {
 	Labels                  flag.Map
 	Selector                flag.LabelSelector
 	IncludeClusterResources flag.OptionalBool
+	Wait                    bool
 }
 
 func NewCreateOptions() *CreateOptions {
@@ -91,6 +95,12 @@ func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 
 	f = flags.VarPF(&o.IncludeClusterResources, "include-cluster-resources", "", "include cluster-scoped resources in the backup")
 	f.NoOptDefVal = "true"
+}
+
+// BindWait binds the wait flag separately so it is not called by other create
+// commands that reuse CreateOptions's BindFlags method.
+func (o *CreateOptions) BindWait(flags *pflag.FlagSet) {
+	flags.BoolVarP(&o.Wait, "wait", "w", o.Wait, "wait for the operation to complete")
 }
 
 func (o *CreateOptions) Validate(c *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


pkg/cmd/cli/backup/create.go
```diff
@@ -138,12 +143,77 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
...
 
+	var sharedInformerFactory informers.SharedInformerFactory
+	var updates chan *api.Backup
+	if o.Wait {
+		stop := make(chan struct{})
+		defer close(stop)
+
+		updates = make(chan *api.Backup)
+
+		sharedInformerFactory = informers.NewFilteredSharedInformerFactory(arkClient, 0, f.Namespace(), nil)
```
skriss: I don't think we really need an informer factory, since we know we only need a single informer for one client - https://github.com/heptio/ark/blob/c8989231eb468f1ba9bdc531668fca8e2580c0ba/pkg/generated/informers/externalversions/ark/v1/backup.go#L54 might require fewer resources.

pkg/cmd/cli/backup/create.go
```diff
@@ -138,12 +143,77 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
...
+			case <-ticker.C:
+				fmt.Print(".")
+			case backup, ok := <-updates:
+				if !ok {
+					fmt.Println("\nError waiting: unable to watch backups.")
+					return nil
+				}
+
+				if backup.Status.Phase != api.BackupPhaseNew && backup.Status.Phase != api.BackupPhaseInProgress {
+					fmt.Printf("\nBackup completed with status: %s\n", backup.Status.Phase)
```
skriss: would be nice to also print something about using `ark backup describe` and `ark backup logs` for more information.

pkg/cmd/cli/backup/create.go
```diff
@@ -138,12 +143,77 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
...
+				},
+				Handler: cache.ResourceEventHandlerFuncs{
+					UpdateFunc: func(_, obj interface{}) {
+						backup, ok := obj.(*api.Backup)
+						if !ok {
+							return
+						}
+						updates <- backup
+					},
+					DeleteFunc: func(obj interface{}) {
```
skriss: Not sure that we really need the delete handler, but I guess it doesn't hurt

carlisia:
Copy that, thank you.

carlisia:
Last night I was testing canceling the backup with ctrl-c and the background process disappeared. I expected that it would keep alive til the backup finished. 

Another thing is that the backup took forever to the point that I never let it finish.

Want to look at these again today.

ncdc:
The expected outcome of `ark backup create --wait` + `ctrl-c` is that the backup creation continues in the background. The only thing `ctrl-c` should do is cancel the waiting.

carlisia:
image


Tested creating backups using the `wait` flag with and without canceling it with `ctrl-c`, all works.

docs/cli-reference/ark_create_schedule.md
```diff
@@ -42,6 +42,7 @@ ark create schedule NAME --schedule="0 */6 * * *"
       --show-labels                                     show labels in the last column
       --snapshot-volumes optionalBool[=true]            take snapshots of PersistentVolumes as part of the backup
       --ttl duration                                    how long before the backup can be garbage collected (default 720h0m0s)
+  -w, --wait                                            wait for the operation to complete
```
skriss: we don't want this flag on `ark schedule create`. It looks like it's getting inherited because the code assumes anything that's an option for creating a backup, should be an option for creating a schedule too (https://github.com/heptio/ark/blob/master/pkg/cmd/cli/schedule/create.go#L76 and https://github.com/heptio/ark/blob/master/pkg/cmd/cli/schedule/create.go#L81 do it).  Need to find a way to separate that out so it applies for backup creates but not schedule creates.

pkg/cmd/cli/backup/create.go
```diff
@@ -91,6 +94,8 @@ func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 
 	f = flags.VarPF(&o.IncludeClusterResources, "include-cluster-resources", "", "include cluster-scoped resources in the backup")
 	f.NoOptDefVal = "true"
+
+	flags.BoolVarP(&o.Wait, "wait", "w", o.Wait, "wait for the operation to complete")
```
skriss: related to the prev comment, we could just move the binding of the `--wait` flag into a different method that only gets called by this command.

skriss:
LGTM. Could you just add a comment explaining why `BindWait` is a separate func, and squash this to one commit? Then good to merge.

PR #312, opened 2018-02-13
adds helm chart to ark

domcar:
## Premise
Helm cannot handle properly CRD becauses it has a validation mechanism that checks the installation before the CRD are actually created,
hence each resource that uses a CRD cannot be validated because the CRD doesn't exist yet!

## Solution
The solution here is to create CRD via helm chart, and only after (using a `post-install`) to install the resources with a container.
The container has the only job to execute a `kubectl create -f filename` and create the resources.

At the same time the resources created with the hook are completely transparent to Helm, that is, when you delete the
chart those resources remain there. Hence we need a sencond hook for deleting them (see delete.yaml)

### DIFF
--- a/contribution/ark-server/.helmignore
+++ b/contribution/ark-server/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
--- a/contribution/ark-server/Chart.yaml
+++ b/contribution/ark-server/Chart.yaml
@@ -4,5 +4,5 @@ description: A Helm chart for ark-server
 name: ark-server
 version: 0.6.0
 maintainers:
-- name: Domenico Caruso
+- name: domcar
   email: domenico.caruso@de.clara.net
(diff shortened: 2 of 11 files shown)


unguiculus:
I'd be good if you could follow best practices and guidelines from the [charts](https://github.com/kubernetes/charts) repo. Or maybe even post the chart there?

Here's a few pointers:
* https://github.com/kubernetes/charts/blob/master/REVIEW_GUIDELINES.md
* https://github.com/kubernetes/helm/tree/master/docs/chart_best_practices
* https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/labels.md
* https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/rbac.md

ncdc:
Hi @domcar, thanks for the PR. I do have a question - if we had an `ark server install` command that did all of this for you, do you think that would be sufficient? Or would you prefer a Helm chart?

jrnt30:
I would 👍 a chart as well.  We have a lot of existing operational expertise and work around deploying Helm charts via Helmfile.  

I think adding it to the Kubernetes repo directly would be a bit more discoverable similar to @unguiculus point

domcar:
@ncdc I read about the `ark server install` in a Issue but I still would prefer Helm. 
The reason is mainly that when deploying to K8s I would like to have a single approach to install all applications rather than having different ones for each tool.

@unguiculus thanks for the links

ncdc:
cc @bryanl @abiogenesis-now

ncdc:
Does it make more sense for the chart to live here in the Ark repo, or in the charts repo?

abiogenesis-now:
I think charts repo (but clearly linked from the Ark README/install guide) makes the most sense?

ncdc:
Hi @domcar, I just noticed you submitted https://github.com/kubernetes/charts/pull/3795 to the charts repo. Would it be ok to close this?

Assuming your charts PR is merged, would you be willing to do another PR to Ark describing how to install using the chart?

domcar:
@ncdc I'll close this now.

No problem, I'll make another PR as soon as the PR gets merged.

PR #318, opened 2018-02-16
[docs] Minor fixes for subsection links

abiogenesis-now:
@Bradamant3 PTAL thanks!

There might be other broken links that need to be corrected---the results of the link checker test were a bit noisy (since I was testing the raw output of the site gen tool). Figured that I might as well fix these ones though!

### DIFF
--- a/docs/build-from-scratch.md
+++ b/docs/build-from-scratch.md
@@ -55,11 +55,11 @@ the files:
 
 If you make the following change, you must run [generate-proto.sh][13] to regenerate files:
 
-* Add/edit/remove protobuf message or service definitions. These changes require the [proto compiler][14]. 
+* Add/edit/remove protobuf message or service definitions. These changes require the [proto compiler][14].
 
 ### Cross compiling
 
-By default, `make` builds an `ark` binary that runs on your host operating system and architecture. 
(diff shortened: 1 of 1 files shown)


Bradamant3:
/lgtm and thanks!
(also weird, bc I could have sworn I had these same changes in a commit somewhere. Evidently did not make it in ;) )

Bradamant3:
Also @ncdc @skriss I'll get the site rebuilt with these changes sometime today. It's a little fiddly to fix up the 0.7.0 branch, but it's also a good opportunity to make sure my assumptions about what to do make sense going forward.

ncdc:
@Bradamant3 I'm not sure how the tooling works, but we could maybe have it check to see if there's a `release-MAJOR-MINOR` branch to match a tag's major & minor versions, and if there is, use that instead of the tag?

For example, if you need to adjust what's published for 0.7.x docs, and it doesn't require any ark code changes in git, then we could create a `release-0.7` branch, you could make your copy edits there, and the tooling could see there's a `v0.7.0` tag, see there's a `release-0.7` branch, and build off the branch. Do you think that would be a useful feature?

Bradamant3:
Closing because these changes made it in in other PRs. Tracking suggestion from @ncdc in issue in docs tooling repo.

PR #319, opened 2018-02-16
fixes broken cloud provider doc links

chasestarr:
noticed some links in the docs pointed nowhere - thanks for putting this tool together

### DIFF
--- a/docs/cloud-common.md
+++ b/docs/cloud-common.md
@@ -70,8 +70,8 @@ After you set up the Ark server, try these examples:
     ark restore create nginx-backup
     ```
 
-[0]: /aws-config.md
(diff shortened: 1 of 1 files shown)


Bradamant3:
/lgtm
And thank you @chasestarr!
(Will fix more like this in another PR)

Bradamant3:
n.b. if anyone else trips over the issue before I get to it (and also for myself): the docs tool supports three different syntaxes for the link paths. GitHub supports two. They aren't all normalized atm; I'll get them fixed up first thing next week.
no need to rework tool: fixing path syntax is cleanest approach but will also document it for future reference.

PR #322, opened 2018-02-20
Move ark server & minio to heptio-ark-server ns

ncdc:
Move ark server deployment & minio deployment to a separate namespace
from the backups/schedules/restores/config because backups now have a
finalizer. If everything lives in one namespace, you have to delete all
the backups and wait for the GC controller to process them and remove the
finalizer from each before deleting the namespace.

By moving the server into a separate namespace, users can now delete the
heptio-ark namespace the normal way (kubectl delete), and once that
namespace is fully removed, they can delete the heptio-ark-server
namespace.

### DIFF
--- a/docs/namespace.md
+++ b/docs/namespace.md
@@ -1,15 +1,30 @@
 # Run in custom namespace
 
-In Ark version 0.7.0 and later, you can run Ark in any namespace. To do so, you specify the namespace in the YAML files that configure the Ark server. You then also specify the namespace when you run Ark client commands.
+In Ark version 0.7.0 and later, you can run Ark in any namespace. To do so, you specify the
+namespace in the YAML files that configure the Ark server. You then also specify the namespace when
+you run Ark client commands.
 
 ## Edit the example files
 
-The Ark repository includes [a set of examples][0] that you can use to set up your Ark server. The examples specify only the default `heptio-ark` namespace. To run in another namespace, you edit the relevant files to specify your custom namespace.
+The Ark repository includes [a set of examples][0] that you can use to set up your Ark server. The
+examples place the server in the `heptio-ark-server` namespace, and backup/schedule/restore/config
+data in the `heptio-ark` namespace.
+
+To run the server in another namespace, you edit the relevant files, changing `heptio-ark-server` to
+your desired namespace.
+
+To store your backups, schedules, restores, and config in another namespace, you edit the relevant
+files, changing `heptio-ark` to your desired namespace.
+
+WARNING: It is recommended to run the Ark server in one namespace, and place your backups, schedules,
+restores, and config in a different namespace. You might encounter issues with deleting a single Ark
+namespace that contains everything.
 
 For all cloud providers, edit `https://github.com/heptio/ark/blob/master/examples/common/00-prereqs.yaml`. This file defines:
 
 * CustomResourceDefinitions for the Ark objects (backups, schedules, restores, configs, downloadrequests)
-* The Ark namespace
+* The namespace where the Ark server runs
+* The namespace where backups, schedules, restores, and the config are stored
 * The Ark service account
 * The RBAC rules to grant permissions to the Ark service account
 
--- a/examples/common/00-prereqs.yaml
+++ b/examples/common/00-prereqs.yaml
@@ -95,10 +95,16 @@ metadata:
 
 ---
 apiVersion: v1
+kind: Namespace
+metadata:
+  name: heptio-ark-server
+
+---
+apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: ark
-  namespace: heptio-ark
+  namespace: heptio-ark-server
   labels:
     component: ark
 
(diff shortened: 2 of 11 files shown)


ncdc:
Still need to update the docs, as they say you can just `kubectl delete -f examples/common`, which is still going to be a problem for the moment.

ncdc:
Docs updated

ncdc:
Prior -> Before

nrb:
Haven't run this yet, but does this somewhat nullify running Ark in any namespace? It's more like the backups are now managed in any namespace and the server will always run in `heptio-ark-server`, isn't it?

These questions are more clarifying docs and terminology than the code.

ncdc:
@nrb not exactly. You can run the server in whatever namespace you want (assuming you adjust the yaml appropriately). And you can point the server at whatever namespace you want for the backups/schedules/restores/config. I debated renaming the `--namespace` flag for `ark server` to something like `--backup-namespace`, but I don't think that name is entirely clear about its purpose, plus that only mentions backups and not the rest of the custom resources.

This again highlights the need for a simple way to put the server in 1 ns and the backups etc in another one. Something like `ark server install --server-namespace foo --data-namespace bar` could do it.

ncdc:
@Bradamant3 looks like I probably should update `docs/namespace.yaml` too, since we now have 2 namespaces to possibly change instead of 1.

nrb:
Credentials namespace in https://github.com/heptio/ark/blob/master/docs/aws-config.md would need to be changed.

e.g.
```
kubectl create secret generic cloud-credentials \
    --namespace <ARK_NAMESPACE> \
    --from-file cloud=credentials-ark
```

The secret needs to be in `heptio-ark-server` or wherever the server is.

ncdc:
@nrb correct, thanks for pointing that out

docs/namespace.md
```diff
@@ -1,15 +1,30 @@
...
+data in the `heptio-ark` namespace.
+
+To run the server in another namespace, you edit the relevant files, changing `heptio-ark-server` to
+your desired namespace.
+
+To store your backups, schedules, restores, and config in another namespace, you edit the relevant
+files, changing `heptio-ark` to your desired namespace.
+
+NOTE: it is strongly advised that you run the Ark server in its own namespace, separate from the
+namespace where your backups, schedules, restores, and config live. If you don't, you may not be
```
Bradamant3: edit:
WARNING: It is recommended to run the Ark server in one namespace, and your backups, schedules, restores, and config in a different namespace. You might encounter issues with deleting a single Ark namespace that runs everything.
(edits for clarity only)

Bradamant3:
one edit for clarity, otherwise /lgtm

ncdc:
Thanks, I've updated (with a minor edit to your suggestion).

examples/common/00-prereqs.yaml
```diff
@@ -111,42 +117,9 @@ metadata:
...
-    namespace: heptio-ark
+    namespace: heptio-ark-server
     name: ark
 roleRef:
   kind: ClusterRole
   name: cluster-admin
   apiGroup: rbac.authorization.k8s.io
-
----
-apiVersion: rbac.authorization.k8s.io/v1beta1
```
skriss: why's the RBAC stuff going away?

ncdc: Because our service account has cluster-admin (granted above)

docs/namespace.md
```diff
@@ -1,15 +1,30 @@
...
+examples place the server in the `heptio-ark-server` namespace, and backup/schedule/restore/config
+data in the `heptio-ark` namespace.
+
+To run the server in another namespace, you edit the relevant files, changing `heptio-ark-server` to
+your desired namespace.
+
+To store your backups, schedules, restores, and config in another namespace, you edit the relevant
+files, changing `heptio-ark` to your desired namespace.
+
+WARNING: It is recommended to run the Ark server in one namespace, and place your backups, schedules,
```
skriss: Given that someone could still end up "stuck" if they remove Ark in the wrong way (i.e. delete the server NS first), it'd probably be helpful to add an entry to the `Troubleshooting` section in the docs about what to do if that happens (including your cmd @ncdc ?) 

Also, @Bradamant3 can we add a link to the GH pages site near the top of the README? How do folks find it otherwise?

skriss:
Did some testing on GKE and code/config changes LGTM. One other observation is that using the jq snippet from @ncdc to remove finalizers doesn't seem to work if there's only a single backup.

ncdc:
Oh that's probably because it returns a single one instead of a list

ncdc:
We can use `kubectl -n heptio-ark get backup` instead of `ark backup get` and it will always return a list

skriss:
yep, confirmed that works

Bradamant3:
LGTM

ncdc:
Going to merge as-is and @Bradamant3 will submit a follow-up for any additional docs edits.

PR #324, opened 2018-02-21
Fix link to vendoring deps

stevesloka:
Link was broken to vendoring deps.

Signed-off-by: Steve Sloka <steves@heptio.com>

### DIFF
--- a/docs/build-from-scratch.md
+++ b/docs/build-from-scratch.md
@@ -114,7 +114,7 @@ If you need to add or update the vendored dependencies, see [Vendoring dependenc
 [8]: /config-definition.md
 [9]: /cloud-common.md
 [10]: #vendoring-dependencies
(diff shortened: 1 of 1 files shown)


Bradamant3:
Other links in same file are broken. See PR #319. Having issues atm getting the syntax right for both gh-pages and links in repository files. Tracking down, please hold.

Bradamant3:
OK, cleanest approach (not the fix in #319), is different from the one in this PR. And in fact, this PR would break the gh-pages docs site next time we generate it. I have a reliable fix for this and some other broken links in another commit (needed to test on site and GH), so if it's OK with y'all I'll close this one, add some more commits to fix more links in my branch, and then submit another PR.

stevesloka:
I don't want to break other items like docs, so we can close, no problem. =)

PR #334, opened 2018-02-27
Run ark server locally

ncdc:
If you want to test changes to the ark server without having to rebuild
and redeploy the ark container, this change allows you to do something
like this (assuming you've created your cloud credentials file):

AWS_SHARED_CREDENTIALS_FILE=credentials-minio ark server -n heptio-ark

Also added a `--plugin-dir` flag so you can test plugins locally too.


skriss:
LGTM.

PR #341, opened 2018-02-28
Add tags to snapshots, restored volumes

### LINKED ISSUE #278: Volume snapshots don't include informative metadata
Currently when a Snapshot is created in AWS due to a Backup, the snapshot has no name nor description. It has only a label pointing to Ark: "tag-key=ark-snapshot"

Having lots of snapshots will be difficult to see why some snapshot is for.

The idea is to add a name, and a description adding more data, for example the Backup name and/or the POD's name owning the PVC. Additional tags can also be added.

### LINKED ISSUE #307: [Feature Request] PV identifiers as tags on snapshots
It would be helpful if specific identifying information of the PV were included in a tag on the snapshot. For the sake of snapshot auditing and cost attribution.

### LINKED ISSUE #369: kops master is unauthorized to attach ark-restored volumes
Hi, I'm trying to solve the nginx use-case but using S3 for object storage. 
I'm trying to restore a backup which is created by running this command `ark backup create nginx-backup --selector app=nginx --snapshot-volumes`. The command used for restoring is `ark restore create nginx-backup --restore-volumes`.

The backup however is being created successfully and the backup files are uploaded to the object storage and the snapshot is getting created. The issue that I'm facing is that it is pointing to the same PV while restoring in a different k8s cluster. And the pod that is supposed to restore is struck in STATUS `ContainerCreating`. Is there anyway that I can get it to create a new PV while restoring in a different cluster? 

Output of `ark backup describe nginx-backup`
```
Name:         nginx-backup
Namespace:    heptio-ark
Labels:       <none>
Annotations:  <none>

Namespaces:
  Included:  *
  Excluded:  <none>

Resources:
  Included:        *
  Excluded:        <none>
  Cluster-scoped:  auto

Label selector:  app=nginx

Snapshot PVs:  true

TTL:  720h0m0s

Hooks:  <none>

Phase:  Completed

Backup Format Version:  1

Expiration:  2018-04-07 19:10:11 +0000 UTC

Validation errors:  <none>

Persistent Volumes:
  pvc-2db45bb0-22f8-11e8-82f2-0e21f011a24c:
    Snapshot ID:        snap-0c9aad251b280516d
    Type:               gp2
    Availability Zone:  us-east-1a
    IOPS:               <N/A>
```

Output of `ark restore describe nginx-backup-20180308194129`
```
Name:         nginx-backup-20180308194129
Namespace:    heptio-ark
Labels:       <none>
Annotations:  <none>

Backup:  nginx-backup

Namespaces:
  Included:  *
  Excluded:  <none>

Resources:
  Included:        *
  Excluded:        nodes
  Cluster-scoped:  auto

Namespace mappings:  <none>

Label selector:  <none>

Restore PVs:  true

Phase:  Completed

Validation errors:  <none>

Warnings:  <none>
Errors:    <none>
```
yaml file used to create nginx-example
```
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: nginx-deployment
  namespace: nginx-example
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: nginx
    spec:
      volumes:
        - name: nginx-logs
          persistentVolumeClaim:
           claimName: nginx-logs
      containers:
      - image: nginx:1.7.9
        name: nginx
        ports:
        - containerPort: 80
        volumeMounts:
          - mountPath: "/var/log/nginx"
            name: nginx-logs
            readOnly: false

---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: nginx
  name: my-nginx
  namespace: nginx-example
spec:
  ports:
  - port: 80
    targetPort: 80
  selector:
    app: nginx
  type: LoadBalancer
```
NOTE: I'm using the k8s nodes iam instance profile to provide EC2 and S3 access to the ark server instead of a secret as mentioned in the example use-case.

skriss:
Fixes #307 
Fixes #278 
Fixes #369 

Initial implementation for AWS. These changes:
- copy tags from the source EBS volume to the snapshot when taking a backup
- copy tags from the snapshot to the new EBS volume when doing a restore
- add ark-specific tags to the snapshot with the backup name and pv name (may be redundant - is this true in all cases?)

This is intentionally the MVP for this feature request. I didn't modify plugin interfaces or do significant refactoring to be able to tag with things like pod name. The dynamic provisioner already tags volumes with pv, pvc, namespace name and that seems sufficient.

Looking for feedback before proceeding to other cloud providers or fixing unit tests.

cc @totemcaf @activeshadow @rbowlby @kaspernissen @ncdc @nrb


rbowlby:
This is great! Thank you. 👍 

In our case we have multiple k8s clusters in the same region so there's always a chance that the snapshots could originate from either. A tag specifying the k8s cluster name could prove useful, if not merely for our use case.

nrb:
I like this, even in the most basic case of retaining tags for restore. I think I agree with @rbowlby that having some sort of tag for the cluster could prove useful.

[Amazon's docs](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/allocation-tag-restrictions.html) point out that there's a limit of 50 user-defined tags, though I doubt we'd really get close to that.

skriss:
cluster name is added by the AWS dynamic provisioner, so covered there

skriss:
@ncdc are you happy with this approach?

pkg/cloudprovider/aws/block_store.go
```diff
@@ -84,6 +84,31 @@ func (b *blockStore) CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ s
...
+	snapRes, err := b.ec2.DescribeSnapshots(snapReq)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	if count := len(snapRes.Snapshots); count != 1 {
+		return "", errors.Errorf("expected 1 snapshot from DescribeSnapshots for %s, got %v", snapshotID, count)
+	}
+
+	if tags := snapRes.Snapshots[0].Tags; len(tags) > 0 {
```
ncdc: Can you just do `tagSpec.Tags = snapRes.Snapshots[0].Tags`?

pkg/cloudprovider/aws/block_store.go
```diff
@@ -154,20 +178,24 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
...
-
-	ec2Tags := make([]*ec2.Tag, 0, len(tags))
+	tagsReq := &ec2.CreateTagsInput{
+		Resources: []*string{res.SnapshotId},
+	}
 
 	for k, v := range tags {
-		key := k
-		val := v
+		key, val := k, v
```
ncdc: I think you should be able to remove the `key, val` copy and change the contents of this loop to this, if you want:
```go
tagsReq.Tags = append(tagsReq.Tags, (&ec2.Tag{}).SetKey(k).SetValue(v))
```

pkg/cloudprovider/aws/block_store.go
```diff
@@ -154,20 +178,24 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
...
-		ec2Tags = append(ec2Tags, tag)
+	volumeInfo, err := b.describeVolume(volumeID)
+	if err != nil {
+		return "", err
 	}
 
-	tagsReq.SetTags(ec2Tags)
+	for _, tag := range volumeInfo.Tags {
+		key, val := *tag.Key, *tag.Value
+		tagsReq.Tags = append(tagsReq.Tags, &ec2.Tag{Key: &key, Value: &val})
```
ncdc: Same here re SetKey/SetValue

ncdc:
@skriss approach seems reasonable

skriss:
addressed feedback and added implementations for Azure, GCP. haven't tested those or re-tested AWS yet.

skriss:
This approach isn't right for GCP; will need to look at that more.

skriss:
OK, I've updated the implementations for Azure and GCP to address the nuances there, and have tested both successfully.

ncdc:
@skriss we should test the ark server without and then with this change to see if it fixes a problem where the kops master isn't allowed to attach ebs volumes that ark creates from snapshots.

ncdc:
xref #369

skriss:
okay, I can try that out - never used kops before but it'll be a good chance to get some familiarity.

skriss:
btw, @ncdc this should be ready for another review, as I've tested the core functionality across clouds. whenever you get a chance.

ncdc:
I used kops for the first time this morning. Installed using `brew` and it was pretty simple.

pkg/cloudprovider/azure/block_store.go
```diff
@@ -190,6 +196,13 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
 		snapshotName = volumeID[0:80-len(suffix)] + suffix
 	}
 
+	snapshotTags := diskInfo.Tags
+	for k, v := range tags {
+		key := strings.Replace(k, "/", "-", -1)
```
ncdc: Comment: azure doesn't support / or whatever

pkg/cloudprovider/azure/block_store.go
```diff
@@ -190,6 +196,13 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
...
+		val := v
+		(*snapshotTags)[key] = &val
```
ncdc: Do you need to deref snapshotTags?

pkg/cloudprovider/azure/block_store.go
```diff
@@ -190,6 +196,13 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
 		snapshotName = volumeID[0:80-len(suffix)] + suffix
 	}
 
+	snapshotTags := diskInfo.Tags
```
ncdc: Comment: combine tags from disk with ark specified tags etc

pkg/cloudprovider/gcp/block_store.go
```diff
@@ -104,10 +104,16 @@ func (b *blockStore) CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ s
 		return "", errors.WithStack(err)
 	}
 
+	// upstream uses the description field of GCP disks to store a JSON doc containing
```
ncdc: s/upstream/Kubernetes/

pkg/cloudprovider/gcp/block_store.go
```diff
@@ -148,12 +154,36 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
 		snapshotName = volumeID[0:63-len(suffix)] + suffix
 	}
 
+	disk, err := b.gce.Disks.Get(b.project, volumeAZ, volumeID).Do()
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	// upstream uses the description field of GCP disks to store a JSON doc containing
```
ncdc: s/upstream/Kubernetes/

pkg/cloudprovider/gcp/block_store.go
```diff
@@ -148,12 +154,36 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
...
+	// tags.
+	//
+	// use the tags in the disk's description (if a valid JSON doc) plus the tags arg
+	// to set the snapshot's description.
+	var snapshotTags map[string]string
+	if err := json.Unmarshal([]byte(disk.Description), &snapshotTags); err != nil {
+		snapshotTags = tags
```
ncdc: Comment: error decoding the description, so just use our tags.

Might want to log an error?

pkg/cloudprovider/gcp/block_store.go
```diff
@@ -148,12 +154,36 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
...
+	} else {
+		for k, v := range tags {
```
ncdc: Comment: merging ark assigned tags with the ones from the disk

pkg/cloudprovider/gcp/block_store.go
```diff
@@ -148,12 +154,36 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
...
+			snapshotTags[k] = v
+		}
+	}
+
+	tagsJSON, err := json.Marshal(snapshotTags)
+	if err != nil {
+		return "", errors.WithStack(err)
```
ncdc: If we can't assign tags, we shouldn't create a snapshot?

pkg/cloudprovider/snapshot_service.go
```diff
@@ -94,11 +94,7 @@ func (sr *snapshotService) CreateVolumeFromSnapshot(snapshotID string, volumeTyp
 	}
 }
 
-func (sr *snapshotService) CreateSnapshot(volumeID, volumeAZ string) (string, error) {
-	tags := map[string]string{
-		snapshotTagKey: snapshotTagVal,
```
ncdc: Can you remove these constants from wherever they're defined?

skriss:
OK did kops testing - first tried a backup/restore with ark:master; the restore failed due to unable to attach the volume. Then did the same with this PR; the restore succeeded. (separately, I need to add one more item to the minimal IAM permissions in our docs since this PR adds a DescribeSnapshots call).

ncdc:
Yay!

skriss:
addressed all feedback and fixed 2 other minor issues

ncdc:
Generally LGTM. Would be nice to have some unit tests for the tagging (copy from volume to snapshot, copy from snapshot to volume, ark assigned tags, new ark assigned tags overwrite previously set ark assigned tags).

skriss:
Can definitely add units, although I'll have to do some refactoring first to separate the tag logic into helpers that I can test without actually making any API calls. Want me to do that before merge?

ncdc:
Yes please. LMK if you want to discuss the splits.

skriss:
@ncdc please review https://github.com/heptio/ark/pull/341/commits/22d4ce27db95e6197b67234f68ef9404c2d99fb5 before I proceed to the other providers

ncdc:
@skriss that looks good, but do you need to do any sorting in the unit test since `getCreateTagsInput` uses map-order sorting (i.e. not guaranteed) to iterate through `arkTags`?

skriss:
good idea.

nrb:
So far this looks good to me, I'd echo some sorting in the units.

skriss:
okey doke, unit-tested all the tags. i'll do some squashing before merge.

skriss:
squashing done - @ncdc this is ready for a final review.

pkg/cloudprovider/azure/block_store.go
```diff
@@ -221,6 +217,28 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
 	return snapshotName, nil
 }
 
+func getSnapshotTags(arkTags map[string]string, diskTags *map[string]*string) *map[string]*string {
+	// copy tags from disk to snapshot
```
ncdc: This technically isn't a copy, since it assigns snapshotTags to be the same as diskTags if diskTags isn't nil. Do you want to be mutating diskTags?

skriss: it doesn't really matter here since we're not updating the disk, but easy enough to create a new map. I'll do that for clarity.

skriss:
went through a backup & restore on AWS and everything looks good.

ncdc:
LGTM. You've tested GCP and Azure too?

skriss:
I tested GCP and Azure manually before adding the unit testing for tags - think I need to go through that again given the little bit of refactoring?

ncdc:
Can't hurt :)

skriss:
GKE looks good. Having a hard time getting a working azure env atm.

skriss:
Ok, was able to successfully test on azure too.

skriss:
@ncdc this is ready to merge when you are.

puneethpk:
how to add tags to the snapshot when taking the backup?

PR #342, opened 2018-03-01
add --from-backup flag to ark restore create & allow restore naming

### LINKED ISSUE #60: Consideration for refactoring restore command
When creating a restore, I find it a bit confusing that the `backup`'s name is in the form of an argument, as we aren't actually creating a `backup`, it is just a piece of a given restore.  

One option would be: 
- Have the backup be an explicitly named, required flag.  
- Optionally support an argument which would then be the restore's name itself, not that of the backup

Current: `ark restore create <backup name>`
Proposed: `ark restore create --backup <backup name>`

A broader question would be, has there been consideration to flipping the verb and resource to more closely match how Kubectl works?  `ark create restore ...`

skriss:

Fixes #60

### DIFF
--- a/pkg/cmd/cli/restore/create.go
+++ b/pkg/cmd/cli/restore/create.go
@@ -38,8 +38,13 @@ func NewCreateCommand(f client.Factory, use string) *cobra.Command {
 	o := NewCreateOptions()
 
 	c := &cobra.Command{
-		Use:   use + " BACKUP",
+		Use:   use + " [RESTORE_NAME] --from-backup BACKUP_NAME",
 		Short: "Create a restore",
+		Example: `  # create a restore named "restore-1" from backup "backup-1"
+  ark restore create restore-1 --from-backup backup-1
+
+  # create a restore with a default name ("backup-1-<timestamp>") from backup "backup-1"
+  ark restore create --from-backup backup-1`,
 		Run: func(c *cobra.Command, args []string) {
 			cmd.CheckError(o.Validate(c, args, f))
 			cmd.CheckError(o.Complete(args))
@@ -56,6 +61,7 @@ func NewCreateCommand(f client.Factory, use string) *cobra.Command {
 
 type CreateOptions struct {
 	BackupName              string
+	RestoreName             string
 	RestoreVolumes          flag.OptionalBool
 	Labels                  flag.Map
 	IncludeNamespaces       flag.StringArray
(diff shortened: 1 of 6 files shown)


pkg/cmd/cli/restore/create.go
```diff
@@ -38,7 +38,7 @@ func NewCreateCommand(f client.Factory, use string) *cobra.Command {
 	o := NewCreateOptions()
 
 	c := &cobra.Command{
-		Use:   use + " BACKUP",
+		Use:   use + " [NAME] --from-backup <backup-name>",
```
skriss: Not sure this is entirely clear, open to input on how best to document that restore name is optional, --from-backup is required

nrb:
Nitpick - could you fix up the commit message so it's not wrapping?

skriss:
done

pkg/cmd/cli/restore/create.go
jrnt30: For the `Use` I think this is the form I would expect.  If you wanted to elaborate on it, I think the `Short`, `Long` or `Examples` would be the best place to add that additional context

jrnt30:
Haven't looked into the new PR process with docs, but should there be an associated doc update for the examples in the PR that introduces the changes?

pkg/cmd/cli/restore/create.go
ncdc: I would put `--from-backup` first, e.g. `--from-backup BACKUP_NAME [RESTORE_NAME]`. Putting something in `[braces]` typically denotes that it's optional. And +1 to Justin's suggestion of using short/long/example to show what happens both with and without the restore name.

pkg/cmd/cli/restore/create.go
```diff
@@ -97,8 +99,12 @@ func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 }
 
 func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Factory) error {
-	if len(args) != 1 {
-		return errors.New("you must specify only one argument, the backup's name")
+	if len(o.BackupName) == 0 {
+		return errors.New("you must specify a backup to restore from, via --from-backup")
```
ncdc: Could make this shorter: `--from-backup is required`

skriss:
addressed all feedback, PTAL.

lypht:
I think this a great addition; my full support.

ncdc:
@skriss let's proceed. Please squash into either 2 or 1 commit.

skriss:
squashed.

skriss:
switched order of restore name and backup name in docs/examples.

ncdc:
LGTM

PR #343, opened 2018-03-01
add an AddFunc to handle finalizers in GC for initial lists/resyncs

### LINKED ISSUE #338: GC controller doesn't process backups deleted while ark server was offline
We need to have an AddFunc too

https://github.com/heptio/ark/blob/52641e8e29eb48f0516c34dee315d1d1001d7b7e/pkg/controller/gc_controller.go#L102

AddFunc will be called on the initial list and on any resyncs.

skriss:

Fixes #338

### DIFF
--- a/pkg/controller/gc_controller.go
+++ b/pkg/controller/gc_controller.go
@@ -99,7 +99,10 @@ func NewGCController(
 
 	backupInformer.Informer().AddEventHandler(
 		cache.ResourceEventHandlerFuncs{
-			UpdateFunc: c.handleFinalizer,
+			AddFunc: c.handleFinalizer,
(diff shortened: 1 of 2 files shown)


ncdc:
Might be worth making sure we have a unit test that fails without this fix

ncdc:
@skriss can you do a unit test?

skriss:
How do I write a unit test for this? By accessing the informer's store and `.Add()`ing an item to it?

ncdc:
Let me look into it - it's been a while since I've done one of these

ncdc:
Ok the easiest way to unit test this is to convert the entire GC controller to use a queue and allow the syncHandler to be overridden in the tests. I'll file an issue so we can take care of that later.

PR #344, opened 2018-03-01
Remove Azure location requirement

ncdc:
Instead of requiring the Ark admin to specify a "location" in the azure
persistentVolumeProvider config (meaning only a single location is
supported), get info about the disk (for its location) when creating a
snapshot, and get info about the snapshot (for its location) when
creating a disk from a snapshot.

### DIFF
--- a/docs/config-definition.md
+++ b/docs/config-definition.md
@@ -98,17 +98,14 @@ No parameters required.
 
 | Key | Type | Default | Meaning |
 | --- | --- | --- | --- |
-| `location` | string | Required Field | *Example*: "Canada East"<br><br>See [the list of available locations][5] (note that this particular page refers to them as "Regions"). |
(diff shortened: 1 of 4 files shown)


nrb:
This appears to have worked for me, but I'd like a 2nd look as it appears I also had some Azure permissions issues. I don't believe they're related, but want to be sure.

docs/config-definition.md
```diff
@@ -100,7 +100,6 @@ No parameters required.
 
 | Key | Type | Default | Meaning |
 | --- | --- | --- | --- |
-| `location` | string | Required Field | *Example*: "Canada East"<br><br>See [the list of available locations][5] (note that this particular page refers to them as "Regions"). |
```
skriss: can remove link [5] below.

ncdc:
@skriss fixed and rebased on top of latest master

skriss:
LGTM

PR #345, opened 2018-03-02
extract GCP project ID from creds file, remove from config

skriss:

As discussed this AM - this pulls the project ID from the (required) credentials file rather than requiring it to be provided in the config. I used this code in diluvian.

### DIFF
--- a/pkg/cloudprovider/gcp/block_store.go
+++ b/pkg/cloudprovider/gcp/block_store.go
@@ -17,6 +17,10 @@ limitations under the License.
 package gcp
 
 import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+
 	"github.com/pkg/errors"
 	uuid "github.com/satori/go.uuid"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/gcp/block_store.go
```diff
@@ -73,6 +76,30 @@ func (b *blockStore) Init(config map[string]string) error {
 	return nil
 }
 
+func extractProjectFromCreds() (string, error) {
+	credsBytes, err := ioutil.ReadFile(os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"))
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	var data map[string]interface{}
```
ncdc: You can do this with an inline (or top level) struct:
```go
type credentials struct {
  ProjectID string `json:"project_id"`
}

var c credentials
if err := json.Unmarshal(credsBytes, &c); err != nil {
  // ...
}

if c.ProjectID == "" {
 // error
}

return c.ProjectID

skriss: oh, yeah, nice - i'll do this

skriss:
Gotta actually test this on GCP before merge.

skriss:
OK, tested and works for me!

ncdc:
LGTM

PR #346, opened 2018-03-02
document use cases in more detail

Bradamant3:

Per discussion in #ark-dr on K8s Slack. Plus a couple of links to encourage attention/reading. n.b. this PR is for `master` -- will update `gh-pages` and site separately, probably next week. Process is still a bit manual and fiddly.

@ncdc PTAL

### DIFF
--- a/docs/scenarios.md
+++ b/docs/scenarios.md
@@ -1,0 +1,53 @@
+# Use Cases
+
+This doc provides sample Ark commands for the following common scenarios:
+* [Disaster recovery][0]
+* [Cluster migration][1]
+
+## Disaster recovery
+
+*Using Schedules and Restore-Only Mode*
+
+If you periodically back up your cluster's resources, you are able to return to a previous state in case of some unexpected mishap, such as a service outage. Doing so with Heptio Ark looks like the following:
+
+1. After you first run the Ark server on your cluster, set up a daily backup (replacing `<SCHEDULE NAME>` in the command as desired):
+
+    ```
+    ark schedule create <SCHEDULE NAME> --schedule "0 7 * * *"
+    ```
+    This creates a Backup object with the name `<SCHEDULE NAME>-<TIMESTAMP>`.
+
+2. A disaster happens and you need to recreate your resources.
+
+3. Update the [Ark server Config][3], setting `restoreOnlyMode` to `true`. This prevents Backup objects from being created or deleted during your Restore process.
+
+4. Create a restore with your most recent Ark Backup:
+    ```
+    ark restore create <SCHEDULE NAME>-<TIMESTAMP>
+    ```
+
+## Cluster migration
+
+*Using Backups and Restores*
+
+Heptio Ark can help you port your resources from one cluster to another, as long as you point each Ark Config to the same cloud object storage. In this scenario, we are also assuming that your clusters are hosted by the same cloud provider. **Note that Heptio Ark does not support the migration of persistent volumes across cloud providers.**
+
+1. *(Cluster 1)* Assuming you haven't already been checkpointing your data with the Ark `schedule` operation, you need to first back up your entire cluster (replacing `<BACKUP-NAME>` as desired):
+
+   ```
+   ark backup create <BACKUP-NAME>
+   ```
+   The default TTL is 30 days (720 hours); you can use the `--ttl` flag to change this as necessary.
+
+2. *(Cluster 2)* Make sure that the `persistentVolumeProvider` and `backupStorageProvider` fields in the Ark Config match the ones from *Cluster 1*, so that your new Ark server instance is pointing to the same bucket.
+
+3. *(Cluster 2)* Make sure that the Ark Backup object has been created. Ark resources are synced with the backup files available in cloud storage.
+
+4. *(Cluster 2)* Once you have confirmed that the right Backup (`<BACKUP-NAME>`) is now present, you can restore everything with:
+```
+ark restore create <BACKUP-NAME>
+```
+
+[0]: #disaster-recovery
+[1]: #cluster-migration
+[3]: config-definition.md#main-config-parameters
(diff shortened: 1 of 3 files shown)


docs/scenarios.md
```diff
@@ -0,0 +1,53 @@
...
+    ```
+    This creates a Backup object with the name `<SCHEDULE NAME>-<TIMESTAMP>`.
+
+2. A disaster happens and you need to recreate your resources.
+
+3. Update the [Ark server Config][3], setting `restoreOnlyMode` to `true`. This prevents Backup objects from being created or deleted during your Restore process.
+
+4. Create a restore with your most recent Ark Backup:
+    ```
+    ark restore create <SCHEDULE NAME>-<TIMESTAMP>
```
ncdc: @skriss fyi this will need to be changed after #342 goes in

docs/use-cases.md
```diff
@@ -1,53 +1,17 @@
...
+* Disaster recovery for your persistent volumes
 
-*Using Schedules and Restore-Only Mode*
+* If you don't start with your Kubernetes manifests in source control, you can run Ark to download the entire cluster and then put the manifests in source control.
 
-If you periodically back up your cluster's resources, you are able to return to a previous state in case of some unexpected mishap, such as a service outage. Doing so with Heptio Ark looks like the following:
+* Copy your environment by cloning a cluster, or a portion of a cluster, to another cluster or to the same cluster. You can use the copy for scale testing, reproducing bugs, and more.
 
-1. After you first run the Ark server on your cluster, set up a daily backup (replacing `<SCHEDULE NAME>` in the command as desired):
+  * If you specify CPU and memory requests and limits for a pod, you can test these settings with a clone.
```
ncdc: "For example" ... ?

docs/use-cases.md
```diff
@@ -1,53 +1,17 @@
...
 
-    ```
-    ark schedule create <SCHEDULE NAME> --schedule "0 7 * * *"
-    ```
-    This creates a Backup object with the name `<SCHEDULE NAME>-<TIMESTAMP>`.
+* Support for backing up any directory in any container's file system is planned for a future release. This feature will let you migrate your data across cloud providers more easily.
```
ncdc: Should we include this here, now?

Bradamant3: Yeah, wasn't sure. Tossed it in for discussion. Default Official Docs position: no, we shouldn't.

ncdc: Ok yeah I'd remove

docs/scenarios.md
ncdc: @Bradamant3 #342 has merged, so this needs to be `ark restore create --from-backup <SCHEDULE NAME>-<TIMESTAMP>`

docs/scenarios.md
```diff
@@ -0,0 +1,53 @@
+# Use Cases
+
+This doc provides sample Ark commands for the following common scenarios:
+* [Disaster recovery][0]
+* [Cluster migration][1]
+
+## Disaster recovery
+
+*Using Schedules and Restore-Only Mode*
```
ncdc: I know this was copy/paste, but I think we should make this section about DR & restore-only mode. I don't think mixing in schedules makes sense, especially since restore-only mode applies to all backups, not just those created by a schedule.

Having a use case that is "here's how you configure a daily backup" makes more sense as a standalone topic.

docs/scenarios.md
```diff
@@ -0,0 +1,53 @@
...
+    ark restore create <SCHEDULE NAME>-<TIMESTAMP>
+    ```
+
+## Cluster migration
+
+*Using Backups and Restores*
+
+Heptio Ark can help you port your resources from one cluster to another, as long as you point each Ark Config to the same cloud object storage. In this scenario, we are also assuming that your clusters are hosted by the same cloud provider. **Note that Heptio Ark does not support the migration of persistent volumes across cloud providers.**
+
+1. *(Cluster 1)* Assuming you haven't already been checkpointing your data with the Ark `schedule` operation, you need to first back up your entire cluster (replacing `<BACKUP-NAME>` as desired):
```
ncdc: Same comment as above - I'd remove the mention about scheduling. Assume there is already a backup, be it from a schedule or a one-off.

ncdc: We can move the text below about creating a backup and its TTL up to the to-be sections on creating one-off/ad hoc backups and schedules.

docs/scenarios.md
```diff
@@ -0,0 +1,53 @@
...
+   ```
+   The default TTL is 30 days (720 hours); you can use the `--ttl` flag to change this as necessary.
+
+2. *(Cluster 2)* Make sure that the `persistentVolumeProvider` and `backupStorageProvider` fields in the Ark Config match the ones from *Cluster 1*, so that your new Ark server instance is pointing to the same bucket.
+
+3. *(Cluster 2)* Make sure that the Ark Backup object has been created. Ark resources are synced with the backup files available in cloud storage.
+
+4. *(Cluster 2)* Once you have confirmed that the right Backup (`<BACKUP-NAME>`) is now present, you can restore everything with:
+```
+ark restore create <BACKUP-NAME>
```
ncdc: create --from-backup

docs/use-cases.md
```diff
@@ -1,53 +1,15 @@
-# Use Cases
+# Why Ark?
 
-This doc provides sample Ark commands for the following common scenarios:
-* [Disaster recovery][0]
-* [Cluster migration][1]
+Kubernetes workloads are ephemeral. You describe your cluster's state, and Kubernetes makes it so. So why do you still need to worry about backups and restores? Ark addresses these use cases:
```
ncdc: I wouldn't characterize Kubernetes workloads as ephemeral. I would highlight that Kubernetes lets you define your workloads in a declaratively, which means you describe your cluster's state....

Bradamant3:
Closing in favor of an issue so I can revisit and rework more thoughtfully.

PR #356, opened 2018-03-06
Azure: store snapshot URI to support cross-resource group restores

### LINKED ISSUE #352: Cross-resource group volume restores don't work for Azure
In Azure, all resources are contained within a resource group (RG). If a user takes a backup, including PV snapshots, in Cluster A, which is contained in RG A, then tries to restore it into Cluster B, which is contained in RG B, this will fail. The issue is that we take a resource group as part of the cloud-credentials secret for Azure, and this value is used for both fetching snapshots and creating volumes.

We need to keep track of the RG where the snapshot exists so that we can correctly locate it. We could do this by capturing the full resourceID of the snapshot in VolumeBackupInfo and using that when restoring.

skriss:

Fixes #352

### DIFF
--- a/pkg/cloudprovider/azure/block_store.go
+++ b/pkg/cloudprovider/azure/block_store.go
@@ -20,7 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strings"
+	"regexp"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/arm/disk"
@@ -29,6 +29,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/pkg/errors"
 	"github.com/satori/uuid"
+
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/heptio/ark/pkg/cloudprovider"
@@ -36,15 +37,16 @@ import (
 )
 
 const (
-	azureClientIDKey         string = "AZURE_CLIENT_ID"
-	azureClientSecretKey     string = "AZURE_CLIENT_SECRET"
-	azureSubscriptionIDKey   string = "AZURE_SUBSCRIPTION_ID"
-	azureTenantIDKey         string = "AZURE_TENANT_ID"
-	azureStorageAccountIDKey string = "AZURE_STORAGE_ACCOUNT_ID"
-	azureStorageKeyKey       string = "AZURE_STORAGE_KEY"
-	azureResourceGroupKey    string = "AZURE_RESOURCE_GROUP"
-
-	apiTimeoutKey = "apiTimeout"
+	azureClientIDKey         = "AZURE_CLIENT_ID"
+	azureClientSecretKey     = "AZURE_CLIENT_SECRET"
+	azureSubscriptionIDKey   = "AZURE_SUBSCRIPTION_ID"
+	azureTenantIDKey         = "AZURE_TENANT_ID"
+	azureStorageAccountIDKey = "AZURE_STORAGE_ACCOUNT_ID"
+	azureStorageKeyKey       = "AZURE_STORAGE_KEY"
+	azureResourceGroupKey    = "AZURE_RESOURCE_GROUP"
+	apiTimeoutKey            = "apiTimeout"
+	snapshotsResource        = "snapshots"
+	disksResource            = "disks"
 )
 
 type blockStore struct {
@@ -53,6 +55,12 @@ type blockStore struct {
 	subscription  string
 	resourceGroup string
 	apiTimeout    time.Duration
+}
+
+type snapshotIdentifier struct {
+	subscription  string
+	resourceGroup string
+	name          string
 }
 
 func getConfig() map[string]string {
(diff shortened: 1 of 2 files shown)


ncdc:
Please reword the commit message to indicate it's for Azure (could do `Azure: ...`)

pkg/cloudprovider/azure/block_store.go
```diff
@@ -233,11 +250,19 @@ func (b *blockStore) DeleteSnapshot(snapshotID string) error {
...
+// parseFullSnapshotName takes a snapshot URI and returns the subscription ID,
+// resource group, and snapshot name, or an error if the URI does not match
+// the regexp.
+func parseFullSnapshotName(name string) (string, string, string, error) {
+	submatches := snapshotURIRegexp.FindStringSubmatch(name)
+	if len(submatches) != len(snapshotURIRegexp.SubexpNames()) {
+		return "", "", "", errors.New("snapshot URI could not be parsed")
+	}
+
+	return submatches[1], submatches[2], submatches[3], nil
```
ncdc: Might be overkill, but if you're doing named capture groups, it might be worth looking up the matches by name (https://stackoverflow.com/questions/20750843/using-named-matches-from-go-regex).

pkg/cloudprovider/azure/block_store.go
```diff
@@ -233,11 +250,19 @@ func (b *blockStore) DeleteSnapshot(snapshotID string) error {
...
-	return fmt.Sprintf("/subscriptions/%v/resourceGroups/%v/providers/Microsoft.Compute/disks/%v", subscription, resourceGroup, diskName)
+	return fmt.Sprintf(diskURIFormatString, subscription, resourceGroup, diskName)
 }
 
-func getFullSnapshotName(subscription string, resourceGroup string, snapshotName string) string {
-	return fmt.Sprintf("/subscriptions/%v/resourceGroups/%v/providers/Microsoft.Compute/snapshots/%v", subscription, resourceGroup, snapshotName)
+// parseFullSnapshotName takes a snapshot URI and returns the subscription ID,
+// resource group, and snapshot name, or an error if the URI does not match
+// the regexp.
+func parseFullSnapshotName(name string) (string, string, string, error) {
```
ncdc: How would you feel about returning this? I find anything returning more than 2 non-error values is usually confusing down the road...

```go
type SnapshotIdentifier struct {
  Subscription string
  ResourceGroup string
  Name string
}
```

pkg/cloudprovider/azure/block_store.go
```diff
@@ -29,24 +29,28 @@ import (
...
+	azureClientIDKey         = "AZURE_CLIENT_ID"
+	azureClientSecretKey     = "AZURE_CLIENT_SECRET"
+	azureSubscriptionIDKey   = "AZURE_SUBSCRIPTION_ID"
+	azureTenantIDKey         = "AZURE_TENANT_ID"
+	azureStorageAccountIDKey = "AZURE_STORAGE_ACCOUNT_ID"
+	azureStorageKeyKey       = "AZURE_STORAGE_KEY"
+	azureResourceGroupKey    = "AZURE_RESOURCE_GROUP"
+
+	apiTimeoutKey       = "apiTimeout"
+	diskURIFormatString = "/subscriptions/%v/resourceGroups/%v/providers/Microsoft.Compute/disks/%v"
```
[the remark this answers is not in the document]
skriss: OK; when I first moved it I was thinking (incorrectly) that the regex was also for disks so I wanted to put them near each other. I'll move it back

skriss:
updated. FYI I've tested this by taking a backup from one AKS cluster (in RG A) and restoring it into another (RG B)

ncdc:
cc @paulbouwer @jimzim @noelbundick

skriss:
@paulbouwer @jimzim @noelbundick checking in to see if you have any feedback on this PR - input would be appreciated. Thanks!

noelbundick:
This makes sense & looks good to me!

nrb:
I was able to successfully restore across 2 AKS clusters in different namespaces with this PR.

pkg/cloudprovider/azure/block_store.go
```diff
@@ -211,14 +222,13 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
...
 	_, errChan := b.snaps.CreateOrUpdate(b.resourceGroup, *snap.Name, snap, ctx.Done())
-
 	err = <-errChan
 
 	if err != nil {
 		return "", errors.WithStack(err)
 	}
 
-	return snapshotName, nil
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/snapshots/%s", b.subscription, b.resourceGroup, snapshotName), nil
```
ncdc: @skriss do you think it would be worth creating `getComputeResourceName(subscription, resourceGroup, resource, name string) string`? Then we could call it here (with resource=snapshots) and wherever we call `getFullDiskName` (with resource=disks).

skriss: good idea. the fewer hardcoded URI templates the better.

skriss:
extracted helper & added unit test. merge if ready!

ncdc:
LGTM

PR #360, opened 2018-03-07
Remove duplicate words and minor phrasing fix

GuessWhoSamFoo:
Minor doc fixes :pencil2: 

Signed-off-by: sfoo <sfoohei@gmail.com>


ncdc:
@GuessWhoSamFoo thanks for your PR. However, these are all files that come from other repositories and we don't directly control or modify them. We instead get them via `dep`.

If you'd like to see the wording corrected, I'd recommend that you create PRs in aws-sdk-go and go-plugin directly, and we can pull in your changes some time after they're merged in those repositories.

Does this make sense?

GuessWhoSamFoo:
Will do!

PR #363, opened 2018-03-07
add policy document for AWS with only necessary permissions

hopkinsth:
This PR addresses the remaining request in #208, which is to limit the permissions the AWS Quickstart Docs give to the IAM user they would have you create. I wrote the policy document in this PR after looking at the source for `pkg/cloudprovider/aws` at 178a007 and the AWS API calls that package makes. After some testing with a full cluster backup with `ark`, the permissions I defined here seem to work well.

I'm not sure if `ark` would ever upload anything large enough to cause `aws-sdk-go` to start a multipart upload with S3, but I included the permissions it would need in that case.

### DIFF
--- a/docs/aws-config.md
+++ b/docs/aws-config.md
@@ -40,12 +40,52 @@ For more information, see [the AWS documentation on IAM users][14].
 2. Attach policies to give `heptio-ark` the necessary permissions:
 
     ```bash
-    aws iam attach-user-policy \
-        --policy-arn arn:aws:iam::aws:policy/AmazonS3FullAccess \
-        --user-name heptio-ark
-    aws iam attach-user-policy \
-        --policy-arn arn:aws:iam::aws:policy/AmazonEC2FullAccess \
-        --user-name heptio-ark
+    BUCKET=<YOUR_BUCKET>
+    cat > heptio-ark-policy.json <<EOF
+    {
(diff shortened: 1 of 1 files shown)


ncdc:
@mattmoyer ptal!

docs/aws-config.md
```diff
@@ -40,12 +40,52 @@ For more information, see [the AWS documentation on IAM users][14].
...
+                ]
+            }
+        ]
+    }
+    EOF
+    
+    aws iam put-user-policy \
+      --user-name heptio-ark \
+      --policy-name heptio-ark \
+      --policy-document heptio-ark-policy.json
```
ncdc: I tested this and I had to do `file://heptio-ark-policy.json`, otherwise I got an error (I think it was trying to read the string `heptio-ark-policy.json` as the policy, instead of reading from the file).

hopkinsth:
Sorry about that! I think I changed that argument when I tested this but never updated the doc. Will push that up in a moment.

docs/aws-config.md
hopkinsth: Sorry about that! I think I changed that argument when I tested this but never updated the doc. Will push that up in a moment.

ncdc:
LGTM. Please squash & we'll merge. Thanks!

PR #371, opened 2018-03-09
gh-pages: Master/latest updates

ncdc:
- Redirect to latest stable by default
- Display warning text if you're looking at master

### DIFF
--- a/_includes/version-warning.html
+++ b/_includes/version-warning.html
@@ -1,10 +1,19 @@
-{% if page.version != 'master' and page.version != site.latest %}
+{% if page.version == 'master' %}
+  <div class="version-warning">
+    <p>
+    <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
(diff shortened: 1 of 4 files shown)


_includes/version-warning.html
```diff
@@ -1,10 +1,18 @@
-{% if page.version != 'master' and page.version != site.latest %}
+{% if page.version == 'master' %}
+  <div class="version-warning">
+    <p>
+    <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
+    This is documentation for the latest development version of Ark. Things may be broken.
+    For documentation on the latest stable version, see the <a href="{{latest_url}}">latest version</a>.
```
Bradamant3: This is the documentation for the latest development version of Ark. Both code and docs may be unstable, and these docs are not guaranteed to be up to date or correct. See <a href="{{latest_url}}">the latest stable version</a>.

ncdc:
@Bradamant3 updated, ptal

PR #383, opened 2018-03-15
Switch from finalizer to DeleteBackupRequest for deleting backups

### LINKED ISSUE #358: Convert GC controller to use a work queue
Also make sure we have a unit test that verifies that we handle finalizers for both the AddFunc (backup was deleted while GC controller was offline) and the UpdateFunc (backup was deleted while the GC controller was running).

### LINKED ISSUE #375: Avoid deleting backup files from backup storage
I'm playing around with ark to protect our cluster in case of a disaster. I followed the examples for setting up in the google cloud and created a Schedule to backup hourly. With that in place I wanted to verify what happens when an operator deletes the cluster or necessary resources by accident manually. I tested such a scenario by deleting the `heptio-ark` namespace containing the CRD objects. It took me quite by surprise that deleting the namespace also emptied the google buckets containing the actual backup files.

I guess this would also happen when someone deletes the cluster? Is there a supported way in ark to avoid deleting any files from the backup storages or do you suggest another approach to protect against operator errors?

I'm using ark 0.7.1.

### LINKED ISSUE #376: Error removing backups.ark.heptio.com objects
I deleted the `heptio-ark` namespace in order to test what happens in case of manual operator errors. This deleted all objects from the namespace properly, except the `backups.ark.heptio.com` objects. It did empty the GCP bucket, which is actually something I'd like to avoid as described in #375 but not the objects itself. When viewing the logs from the server it's full of this messages:

```
time="2018-03-13T07:42:14Z" level=info msg="Garbage-collecting backup" backup=heptio-ark/hcg-dev-dust-k8s-backup-20180310220008 logSource="pkg/controller/gc_controller.go:129"
time="2018-03-13T07:42:14Z" level=info msg="Removing backup from object storage" backup=heptio-ark/hcg-dev-dust-k8s-backup-20180310220008 logSource="pkg/controller/gc_controller.go:244"
time="2018-03-13T07:42:14Z" level=error msg="Error patching backup" backup=heptio-ark/hcg-dev-dust-k8s-backup-20180310220008 error="unable to find api field in struct Unstructured for the json field \"metadata\"" error.file="gc_controller.go:152" error.function="(*gcController).handleFinalizer" logSource="pkg/controller/gc_controller.go:152"
```

Also the heptio-ark namespace is still in state "Terminating".

I'm using ark version 0.7.1 and followed the GCP setup instructions. This is my config:

```
apiVersion: ark.heptio.com/v1
kind: Config
metadata:
  namespace: heptio-ark
  name: default
backupStorageProvider:
  name: gcp
  bucket: my-bucket
backupSyncPeriod: 60m
gcSyncPeriod: 60m
scheduleSyncPeriod: 1m
restoreOnlyMode: false
---
apiVersion: ark.heptio.com/v1
kind: Schedule
metadata:
  name: my-backup
  namespace: heptio-ark
spec:
  schedule: 0 * * * *
  template:
    includedNamespaces:
    - development
    - sandbox
    - heptio-ark
    includeClusterResources: false
    ttl: 336h0m0s # 2 weeks
```

And here the Kubernetes versions:
```
Client Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.6", GitCommit:"6260bb08c46c31eea6cb538b34a9ceb3e406689c", GitTreeState:"clean", BuildDate:"2017-12-21T06:34:11Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"8+", GitVersion:"v1.8.7-gke.1", GitCommit:"192ccad06d24af9828cbf42330e1d915cb586406", GitTreeState:"clean", BuildDate:"2018-01-31T21:39:04Z", GoVersion:"go1.8.3b4", Compiler:"gc", Platform:"linux/amd64"}
```

ncdc:
We ran into a lot of problems using a finalizer on the backup to allow
the Ark server to clean up all associated backup data when deleting a
backup.

Users also found it less than desirable that deleting the heptio-ark
namespace resulted in all the backup data being deleted.

This removes the finalizer and replaces it with an explicit
DeleteBackupRequest that is created as a means of requesting the
deletion of a backup and all its associated data. This is what `ark
backup delete` does.

If you use kubectl to delete a backup or to delete the heptio-ark
namespace, this no longer deletes associated backups. Additionally, as
long as the heptio-ark namespace still exists, the Ark server's
BackupSyncController will continually sync backups into the heptio-ark
namespace from object storage.

TODO:
- [x] Modify `ark backup describe` to show `DeleteBackupRequest` status if relevant
- [x] Write one-shot code to run at server startup to remove our old backup finalizer if it's there
- [x] Update unit tests

Fixes #375 
Fixes #376
Fixes #358


ncdc:
@skriss @nrb I haven't updated tests yet - wanted to publish this now so you can start reviewing the logic in the delete and gc controllers.

ncdc:
I also think it would be worth it to add some code that runs once each time the server starts, removing the ark GC finalizer from any backups that have it.

pkg/backup/delete_helpers.go
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+package backup
+
+import (
+	"github.com/heptio/ark/pkg/apis/ark/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func NewDeleteBackupRequest(name string) *v1.DeleteBackupRequest {
```
ncdc: TODO: godoc

pkg/controller/backup_deletion_controller.go
```diff
@@ -0,0 +1,290 @@
...
+		restoreLister:             restoreInformer.Lister(),
+		restoreClient:             restoreClient,
+	}
+
+	c.syncHandler = c.processQueueItem
+	c.cacheSyncWaiters = append(c.cacheSyncWaiters, deleteBackupRequestInformer.Informer().HasSynced, restoreInformer.Informer().HasSynced)
+
+	deleteBackupRequestInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: c.enqueue,
```
ncdc: TODO: need an UpdateFunc (e.g. request is created and modified before the controller sees it, status is still ""/New)

pkg/controller/gc_controller.go
```diff
@@ -17,227 +17,108 @@ limitations under the License.
...
+	backupInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.ResourceEventHandlerFuncs{
-			AddFunc: c.handleFinalizer,
-			UpdateFunc: func(_, upd interface{}) {
-				c.handleFinalizer(upd)
-			},
+			AddFunc:    c.enqueue,
+			UpdateFunc: func(oldObj, newObj interface{}) { c.enqueue(newObj) },
 		},
+		syncPeriod,
```
ncdc: TODO: switch to resyncFunc in genericController

pkg/controller/generic_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+)
+
+type genericController struct {
+	name             string
+	queue            workqueue.RateLimitingInterface
+	logger           logrus.FieldLogger
+	syncHandler      func(key string) error
```
ncdc: TODO: add resyncFunc

pkg/controller/backup_deletion_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	// Try to delete backup from object storage
+	log.Info("Removing backup from object storage")
+	if err := c.backupService.DeleteBackupDir(c.bucket, backup.Name); err != nil {
+		errs = append(errs, errors.Wrap(err, "error deleting backup from object storage").Error())
+	}
+
+	// Try to delete restores
+	log.Info("Removing restores")
```
nrb: IMO we should remove restores when explicitly deleting a backup, but I have seen others mention that retaining restores as a matter of record would be beneficial. The restores would then be cleaned up via the GC controller. I think that's also a reasonable approach, wdyt?

I don't think this should be an option, though, we should only do one or the other.

nrb: I'm seeing now that the GC controller uses the `BackupDeleteRequest`s to do a delete, so I suppose the 2nd approach would require shuffling some logic around. I'm fine with it as is.

pkg/controller/gc_controller.go
```diff
@@ -17,227 +17,108 @@ limitations under the License.
...
+	c.cacheSyncWaiters = append(c.cacheSyncWaiters, backupInformer.Informer().HasSynced)
+
+	backupInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.ResourceEventHandlerFuncs{
-			AddFunc: c.handleFinalizer,
-			UpdateFunc: func(_, upd interface{}) {
-				c.handleFinalizer(upd)
-			},
+			AddFunc:    c.enqueue,
+			UpdateFunc: func(oldObj, newObj interface{}) { c.enqueue(newObj) },
```
skriss: `func(_, newObj interface{}) ...`

pkg/controller/generic_controller.go
```diff
@@ -0,0 +1,138 @@
...
+		// We have to wait here in the deferred function instead of at the bottom of the function body
+		// because we have to shut down the queue in order for the workers to shut down gracefully, and
+		// we want to shut down the queue via defer and not at the end of the body.
+		wg.Wait()
+
+		c.logger.Info("All workers have finished")
+
+	}()
+
+	c.logger.Info("Starting %s", c.name)
```
skriss: would probably be useful to add the controller name as a field to the c.logger

README.md
```diff
@@ -135,28 +135,6 @@ ark restore describe <RESTORE_NAME>
 
 For more information, see [the debugging information][18].
 
-### Clean up
```
skriss: maybe leave this section, and specify that you can just delete the NS, and note that backups/snapshots will **not** be deleted

docs/azure-config.md
```diff
@@ -115,7 +115,7 @@ Now you need to create a Secret that contains all the seven environment variable
 
 ```bash
 kubectl create secret generic cloud-credentials \
-    --namespace <ARK_SERVER_NAMESPACE> \
+    --namespace <ARK_SERVER_ARK_NAMESPACE> \
```
skriss: `ARK_NAMESPACE`

docs/ibm-config.md
```diff
@@ -47,7 +47,7 @@ Create a Secret. In the directory of the credentials file you just created, run:
 
 ```bash
 kubectl create secret generic cloud-credentials \
-    --namespace <ARK_SERVER_NAMESPACE> \
+    --namespace <ARK_SERVER_ARK_NAMESPACE> \
```
skriss: `ARK_NAMESPACE`

docs/namespace.md
```diff
@@ -7,10 +7,9 @@ you run Ark client commands.
 ## Edit the example files
 
 The Ark repository includes [a set of examples][0] that you can use to set up your Ark server. The
-examples place the server in the `heptio-ark-server` namespace, and backup/schedule/restore/config
-data in the `heptio-ark` namespace.
+examples place the server and backup/schedule/restore/config data in the `heptio-ark` namespace.
 
-To run the server in another namespace, you edit the relevant files, changing `heptio-ark-server` to
+To run the server in another namespace, you edit the relevant files, changing `heptio-ark` to
```
skriss: One thing not explicitly called out here is that the cloud-credentials secret also needs to be created in the alternate namespace. It's templatized in each cloud-provider's walkthrough, but maybe worth just referencing it here too.

pkg/controller/backup_deletion_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	// If the backup includes snapshots but we don't currently have a PVProvider, we don't
+	// want to orphan the snapshots so skip deletion.
+	if c.snapshotService == nil && len(backup.Status.VolumeBackups) > 0 {
+		req, err = c.patchDeleteBackupRequest(req, func(r *v1.DeleteBackupRequest) {
+			r.Status.Phase = v1.DeleteBackupRequestPhaseProcessed
+			r.Status.Errors = []string{"unable to delete backup because it includes PV snapshots and Ark is not configured with a PersistentVolumeProvider"}
+		})
+		if err != nil {
```
skriss: `return err`

pkg/apis/ark/v1/delete_backup_request.go
```diff
@@ -0,0 +1,67 @@
...
+	metav1.ObjectMeta `json:"metadata"`
+
+	Spec   DeleteBackupRequestSpec   `json:"spec"`
+	Status DeleteBackupRequestStatus `json:"status,omitempty"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// DeleteBackupRequestList is a list of DeleteBackupRequests.
+type DeleteBackupRequestList struct {
```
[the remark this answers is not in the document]
ncdc: Probably not really worth the effort

pkg/backup/delete_helpers.go
```diff
@@ -0,0 +1,36 @@
...
+	return &v1.DeleteBackupRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: name + "-",
+			Labels: map[string]string{
```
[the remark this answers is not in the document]
ncdc: Yes, so we can delete all requests for backup x

pkg/cmd/cli/backup/delete.go
```diff
@@ -59,37 +52,21 @@ func NewDeleteCommand(f client.Factory, use string) *cobra.Command {
...
-	flags.BoolVar(&o.Confirm, "confirm", o.Confirm, "Confirm forceful deletion")
+	flags.BoolVar(&o.Confirm, "confirm", o.Confirm, "Confirm deletion")
 }
 
 func (o *DeleteOptions) Validate(c *cobra.Command, args []string, f client.Factory) error {
 	if len(args) != 1 {
 		return errors.New("you must specify only one argument, the backup's name")
 	}
 
-	kubeClient, err := f.KubeClient()
```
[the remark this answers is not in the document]
ncdc: Will do

pkg/controller/backup_deletion_controller.go
```diff
@@ -0,0 +1,290 @@
...
+		return c.processRequest(reqCopy)
+	}
+
+	return nil
+}
+
+func (c *backupDeletionController) processRequest(req *v1.DeleteBackupRequest) error {
+	log := c.logger.WithFields(logrus.Fields{
+		"namespace": req.Namespace,
+		"name":      req.Spec.BackupName,
```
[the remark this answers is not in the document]
ncdc: Will do

pkg/controller/backup_deletion_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error getting DeleteBackupRequest")
+	}
+
+	switch req.Status.Phase {
+	case "", v1.DeleteBackupRequestPhaseNew:
+		// Don't mutate the shared cache
+		reqCopy := req.DeepCopy()
+		return c.processRequest(reqCopy)
+	}
```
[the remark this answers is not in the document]
ncdc: Agreed

pkg/controller/gc_controller.go
ncdc: Will do

README.md
ncdc: Ok. I'll make it clearer that this is optional and only if you want to remove stuff.

docs/azure-config.md
ncdc: Thanks

docs/ibm-config.md
ncdc: Thanks

docs/namespace.md
ncdc: Will do

skriss:
Some observations from testing:

- all of the normal-case functionality looks good (removal of finalizer from old backups, `ark backup delete`, garbage-collection)
- the warning message when running `ark backup delete foo` without the `--confirm` flag is misleading (references force-deletion, orphaning cloud resources)
- in Azure, if I manually delete a snapshot before deleting a backup that references it, Ark doesn't error and deletes the backup (I think the outcome is OK, but I would've expected an error. It may be that the Azure API doesn't return an error if you try to delete a non-existent snapshot?)
- when running `ark backup delete` for a non-existent backup, the Ark server log has this error:
    ```
    time="2018-03-16T16:57:58Z" level=error msg="Error in syncHandler, re-adding item to queue" controller=backup-deletion error="error patching DeleteBackupRequest: Object 'Kind' is missing in '{\"object\":{\"apiVersion\":\"ark.heptio.com/v1\",\"kind\":\"DeleteBackupRequest\",\"metadata\":{\"clusterName\":\"\",\"creationTimestamp\":\"2018-03-16T16:57:58Z\",\"deletionGracePeriodSeconds\":null,\"deletionTimestamp\":null,\"generateName\":\"nginx-\",\"generation\":0,\"labels\":{\"ark.heptio.com/backup-name\":\"nginx\"},\"name\":\"nginx-3hp7g\",\"namespace\":\"heptio-ark\",\"resourceVersion\":\"1049234\",\"selfLink\":\"\",\"uid\":\"2ddbb8b4-293b-11e8-bf7d-0a58ac1f100c\"},\"spec\":{\"backupName\":\"nginx\"},\"status\":{\"errors\":[\"backup not found\"],\"phase\":\"Processed\"}}}'" error.file="backup_deletion_controller.go:259" error.function="(*backupDeletionController).patchDeleteBackupRequest" key=heptio-ark/nginx-3hp7g logSource="pkg/controller/generic_controller.go:136"
    ```
- if I take a backup with a snapshot, then remove the `persistentVolumeProvider` from config, then try to delete the backup, the failed delete request doesn't show up in `ark backup describe` -- looks like this is because we only show the associated requests if the backup is in phase `Deleting`, which the backup is not in this case
- if I try to delete a non-existent backup `foo`, then later I actually create a backup `foo`, when I go to delete backup `foo` I now see the failed backup requests from prior to the backup existing in `ark backup describe`. Maybe they should be linked by UID rather than name?

ncdc:
> the warning message when running ark backup delete foo without the --confirm flag is misleading (references force-deletion, orphaning cloud resources)

Fixed

> in Azure, if I manually delete a snapshot before deleting a backup that references it, Ark doesn't error and deletes the backup (I think the outcome is OK, but I would've expected an error. It may be that the Azure API doesn't return an error if you try to delete a non-existent snapshot?)

I'm not sure an error is appropriate here. Your intent is to delete a backup and all its associated data. If we can't find some of the associated data, that's not really an issue, is it? I would expect it to be a problem if you wanted to **restore**, of course.

Maybe we need something as part of the backup sync controller that periodically verifies the integrity of a backup's associated data?

> you were going to add a check for non-existent backups in the command code, looks like that's not there yet

That's pushed up now.

> when running ark backup delete for a non-existent backup, the Ark server log has this error:

I manually created a `DeleteBackupRequest` with what should be the same as what `ark backup delete` does (since I now have the check for nonexistent backups), and I can't reproduce the patch error. Would you have some time to try to reproduce & debug?

> if I take a backup with a snapshot, then remove the persistentVolumeProvider from config, then try to delete the backup, the failed delete request doesn't show up in ark backup describe -- looks like this is because we only show the associated requests if the backup is in phase Deleting, which the backup is not in this case

So maybe always show if at least 1 has errors?

> if I try to delete a non-existent backup foo, then later I actually create a backup foo, when I go to delete backup foo I now see the failed backup requests from prior to the backup existing in ark backup describe. Maybe they should be linked by UID rather than name?

I'll link on both name and uid.


This brings up something I haven't addressed yet - there's no expiration on `DeleteBackupRequests` nor any good way to clean them up (other than successfully deleting a backup).

skriss:
> in Azure, if I manually delete a snapshot before deleting a backup that references it, Ark doesn't error and deletes the backup (I think the outcome is OK, but I would've expected an error. It may be that the Azure API doesn't return an error if you try to delete a non-existent snapshot?)

> I'm not sure an error is appropriate here. Your intent is to delete a backup and all its associated data. If we can't find some of the associated data, that's not really an issue, is it? I would expect it to be a problem if you wanted to **restore**, of course.

Yeah, I think it's fine as-is.

> Maybe we need something as part of the backup sync controller that periodically verifies the integrity of a backup's associated data?

Would be a nice feature to add, but not needed as part of this PR.

> when running ark backup delete for a non-existent backup, the Ark server log has this error:

> I manually created a `DeleteBackupRequest` with what should be the same as what `ark backup delete` does (since I now have the check for nonexistent backups), and I can't reproduce the patch error. Would you have some time to try to reproduce & debug?

Yep, I'll look into it

> if I take a backup with a snapshot, then remove the persistentVolumeProvider from config, then try to delete the backup, the failed delete request doesn't show up in ark backup describe -- looks like this is because we only show the associated requests if the backup is in phase Deleting, which the backup is not in this case

> So maybe always show if at least 1 has errors?

I *think* that makes sense. Makes more sense to me than putting the backup in `Deleting`.

> This brings up something I haven't addressed yet - there's no expiration on `DeleteBackupRequests` nor any good way to clean them up (other than successfully deleting a backup).

Yeah - probably not urgent but would be a good idea to GC old ones, esp. those that aren't associated with any existing backup.

skriss:
I can no longer produce the patch error - attempted deletion of a nonexistent backup is blocked by the CLI, and applying the resource directly works fine.

ncdc:
@skriss added deletion of processed delete backup requests > 24h

skriss:
Latest commit LGTM

skriss:
2 things found testing:

- deleting a backup while it's in progress is allowed. this could lead to snapshots getting orphaned if it's deleted after the snapshot is triggered but before it's recorded in the backup obj (I was able to make this happen). It's certainly an edge case, and blocking deletion of In Progress backups may not be desirable (e.g. if they get stuck there for some reason), so not sure if it makes sense to attempt to prevent or not.
- I tested on GKE this time, and if I manually deleted a snapshot prior to `ark backup delete`ing the backup linked to the snapshot, I now get an error with the deletion because the GCP API returns a not found error from the DeleteSnapshot call. Per prior comments, this goes through without error on Azure (I think the Azure API doesn't return an error in this scenario). Haven't tried on AWS yet. It would be nice to have consistent behavior across clouds if possible.

skriss:
As a followup to the previous comment, the scenario in the 2nd bullet also errors on AWS.

ncdc:
> deleting a backup while it's in progress is allowed. this could lead to snapshots getting orphaned if it's deleted after the snapshot is triggered but before it's recorded in the backup obj (I was able to make this happen). It's certainly an edge case, and blocking deletion of In Progress backups may not be desirable (e.g. if they get stuck there for some reason), so not sure if it makes sense to attempt to prevent or not.

The problem is knowing if an InProgress backup is truly in progress, or stuck. Maybe it's running, or maybe the Ark server was restarted in the middle of the backup.

We could add an in-memory list/map/set of in progress backups and disallow deleting those.

> I tested on GKE this time, and if I manually deleted a snapshot prior to ark backup deleteing the backup linked to the snapshot, I now get an error with the deletion because the GCP API returns a not found error from the DeleteSnapshot call. Per prior comments, this goes through without error on Azure (I think the Azure API doesn't return an error in this scenario). Haven't tried on AWS yet. It would be nice to have consistent behavior across clouds if possible.

This requires that we change the protobuf messages to return better errors, so we can identify "not found" properly. I'd like to defer fixing this until we make the plugin/protobuf changes. That ok with you?

ncdc:
Latest commit keeps track of in-progress backups in memory and doesn't allow deletion of them.

PR #394, opened 2018-03-20
Fix typo in code comment

containscafeine:


ncdc:
@containscafeine this isn't a typo. The comment is correct. We are checking for resource == namespaces.

containscafeine:
whoops, interpreted it incorrectly then, sorry about that

PR #402, opened 2018-03-23
Documentation on using ark with kube2iam

domderen:
This PR updates the documentation & example deployment template to show how `ark` can be ran utilizing [https://github.com/jtblin/kube2iam](Kube2iam) for AWS IAM permissions, rather than using access key & secret key.


examples/common/10-deployment-kube2iam.yaml
```diff
@@ -0,0 +1,46 @@
...
+      labels:
+        component: ark
+      annotations:
+        iam.amazonaws.com/role: arn:aws:iam::<AWS_ACCOUNT_ID>:role/<HEPTIO_ARK_ROLE_NAME>
+    spec:
+      restartPolicy: Always
+      serviceAccountName: ark
+      containers:
+        - name: ark
+          image: gcr.io/heptio-images/ark:latest
```
jrnt30: Based on the file name, are you intending to deploy the Kube2iam daemonset here?

domderen: No, this file does not attempt to install kube2iam, and I don’t think it should. It can be used by so many different apps, that it should be installed manually up front I think.

One potential option would be to make kube2iam an optional dependency of Ark in the Ark helm package, when this will be available. Then Helm could install kube2iam if user decides that they want to use it, and it is not yet available in the cluster.

docs/aws-config.md
```diff
@@ -149,6 +149,117 @@ Specify the following values in the example files:
 
     * Replace `<YOUR_STORAGE_CLASS_NAME>` with `gp2`. This is AWS's default `StorageClass` name.
 
+## ALTERNATIVE: Setup permissions using kube2iam
+
+[https://github.com/jtblin/kube2iam](Kube2iam) Kubernetes app allows managing AWS IAM permissions for pod via annotations rather than operating on api keys.
```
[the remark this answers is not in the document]
domderen: Changed

docs/aws-config.md
```diff
@@ -149,6 +149,117 @@ Specify the following values in the example files:
 
     * Replace `<YOUR_STORAGE_CLASS_NAME>` with `gp2`. This is AWS's default `StorageClass` name.
 
+## ALTERNATIVE: Setup permissions using kube2iam
```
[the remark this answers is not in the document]
domderen: Added info about `kube2iam` installation.

docs/aws-config.md
```diff
@@ -149,6 +149,119 @@ Specify the following values in the example files:
...
+
+> This path assumes you have `kube2iam` already running in your Kubernetes cluster. If that is not the case, please install it first, following the docs here: https://github.com/jtblin/kube2iam
+
+[https://github.com/jtblin/kube2iam](Kube2iam) Kubernetes app allows managing AWS IAM permissions for pod via annotations rather than operating on API keys.
```
ncdc: The markdown order is incorrect for link text vs url - it needs to be `[kube2iam](https://github.com/jtblin/kube2iam)`. I'm also not sure if you're trying to say something like "kube2iam _is a_ Kubernetes app _that_ allows managing..."?

docs/aws-config.md
```diff
@@ -149,6 +149,119 @@ Specify the following values in the example files:
...
+
+It can be set up for ark by creating a role that will have required permissions, and later by adding the permissions annotation on the ark deployment to define which role it should use internally.
```
ncdc: Please capitalize Ark

docs/aws-config.md
```diff
@@ -149,6 +149,119 @@ Specify the following values in the example files:
...
+
+1. Create Trust Policy document to allow the role being used for EC2 management & assume kube2iam role:
```
ncdc: Create **a** Trust Policy document

docs/aws-config.md
domderen: Updated according to your comments.

docs/aws-config.md
domderen: Updated

docs/aws-config.md
```diff
@@ -149,6 +149,119 @@ Specify the following values in the example files:
...
+        ]
+    }
+    EOF
+    
+    aws iam put-role-policy \
+      --role-name heptio-ark-role \
+      --policy-name heptio-ark-role-policy \
+      --policy-document file://heptio-ark-role-policy.json
+    ```
+4. Update AWS_ACCOUNT_ID & HEPTIO_ARK_ROLE_NAME in the file `examples/common/10-deployment-kube2iam.yaml`:
```
ncdc: Rather than have a separate example file, would it be sufficient to have these instructions say that you need to add the appropriate annotation to `spec.template.metadata.annotations`? I'd like to avoid another variation of the deployment yaml if possible. WDYT?

docs/aws-config.md
```diff
@@ -149,6 +149,119 @@ Specify the following values in the example files:
...
+
+    ```
+    ---
+    apiVersion: apps/v1beta1
+    kind: Deployment
+    metadata:
+    namespace: heptio-ark-server
```
ncdc: The indentation is off here and below

docs/aws-config.md
```diff
@@ -149,6 +149,119 @@ Specify the following values in the example files:
...
+            }
+        ]
+    }
+    EOF
+    ```
+
+2. Create the IAM role:
+
+    ```bash
+    aws iam create-role --role-name heptio-ark-role --assume-role-policy-document file://./heptio-ark-role-trust-policy.json
```
ncdc: Would you be ok calling the role just `heptio-ark` instead of `heptio-ark-role`?

ncdc: Here you have `file://./....` whereas below you have `file://....` - would it make sense to be consistent?

docs/aws-config.md
domderen: Hmm it is not just adding of the annotation, but also removal of the `AWS_SHARED_CREDENTIALS_FILE` env from the container, and removal of the `cloud-credentials` volume. I'm afraid that trying to just describe how those changes should be done on the existing `10-deployment.yaml` file is quite cumbersome.

docs/aws-config.md
domderen: Fixed

docs/aws-config.md
domderen: Fixed

domderen: Fixed

docs/aws-config.md
ncdc: @domderen good point. Maybe it would be better to have `examples/common` for all the non-cloud specific bits, followed by `examples/aws`, `examples/azure`, `examples/gcp`, and `examples/minio` for the cloud-specific ones. And inside the aws directory, we could have 1 file using shared credentials, and another one for kube2iam. WDYT?

cc @skriss

nrb: That separation makes sense to me. afaik kub2iam is only relevant to AWS, though.

domderen: Makes sense, though it suddenly generates a lot more manifests, and I thought you wanted to avoid maintaining a lot of manifests? ;)

ncdc: Yeah.... but the more I thought about it, the more it makes sense to have multiple variations, so we don't have to tell people stuff like "take this AWS-based one and rename an env var so it'll work for Google"...

domderen: Pushed a change where each cloud provider directory has it's own deployment file, and updated the docs where applicable.

docs/aws-config.md
```diff
@@ -155,9 +151,122 @@ In the root of your Ark directory, run:
...
+
+    ```bash
+    aws iam create-role --role-name heptio-ark --assume-role-policy-document file://./heptio-ark-trust-policy.json
+    ```
+
+3. Attach policies to give `heptio-ark` the necessary permissions:
+
+    ```bash
+    BUCKET=<YOUR_BUCKET>
+    cat > heptio-ark-policy.json <<EOF
```
ncdc: Is this policy identical to the one listed above?

examples/aws/10-deployment-kube2iam.yaml
```diff
@@ -0,0 +1,46 @@
+# Copyright 2017 the Heptio Ark contributors.
```
ncdc: Probably should do 2018

examples/aws/10-deployment-kube2iam.yaml
```diff
@@ -0,0 +1,46 @@
...
+      restartPolicy: Always
+      serviceAccountName: ark
+      containers:
+        - name: ark
+          image: gcr.io/heptio-images/ark:latest
+          command:
+            - /ark
+          args:
+            - server
+            - --namespace
```
ncdc: `--namespace` and `heptio-ark` are no longer required as of what's in master now.

examples/gcp/10-deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+      restartPolicy: Always
+      serviceAccountName: ark
+      containers:
+        - name: ark
+          image: gcr.io/heptio-images/ark:latest
+          command:
+            - /ark
+          args:
+            - server
+            - --namespace
```
ncdc: `--namespace` and `heptio-ark` are no longer required as of what's in master now.

examples/gcp/10-deployment.yaml
```diff
@@ -0,0 +1,52 @@
+# Copyright 2017 the Heptio Ark contributors.
```
ncdc: Let's do 2018

examples/ibm/10-deployment.yaml
```diff
@@ -0,0 +1,52 @@
+# Copyright 2017 the Heptio Ark contributors.
```
ncdc: Let's do 2018

examples/ibm/10-deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+      restartPolicy: Always
+      serviceAccountName: ark
+      containers:
+        - name: ark
+          image: gcr.io/heptio-images/ark:latest
+          command:
+            - /ark
+          args:
+            - server
+            - --namespace
```
ncdc: `--namespace` and `heptio-ark` are no longer required as of what's in master now.

ncdc:
@domderen re the copyright dates - sorry, I was only asking you to change from 2017 to 2018 on files you _added_ as part of this PR. Would you mind undoing the other changes to preexisting files?

ncdc:
@nrb would you mind taking another look?

domderen:
Should be better now ;)

docs/aws-config.md
nrb: Based on L168, yes.

nrb:
Also, we'll need signed-off-by on all the commits. @domderen You can sign existing commits with `git commit -s --amend`. I use an alias of `git ci` which is mapped to `git commit -vs` to sign all commits.

docs/aws-config.md
ncdc: It would be ideal if we can find a way to only list the policy a single time. Any thoughts?

examples/aws/10-deployment.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017 the Heptio Ark contributors.
+# Copyright 2018 the Heptio Ark contributors.
```
ncdc: Since this is just a straight move, I don't think there's a need to change the date.

examples/ibm/00-ark-config.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 the Heptio Ark contributors.
+# Copyright 2017 the Heptio Ark contributors.
```
ncdc: I don't think we need this change?

docs/aws-config.md
```diff
@@ -155,9 +151,122 @@ In the root of your Ark directory, run:
...
+                ],
+                "Resource": [
+                    "arn:aws:s3:::${BUCKET}/*"
+                ]
+            },
+            {
+                "Effect": "Allow",
+                "Action": [
+                    "s3:ListBucket",
+                    "s3:AbortMultipartUpload",
```
ncdc: If we do keep a copy of the policy here, it needs to match the recent changes that went into master to fix the multipart upload being in the wrong place - #419

Bradamant3:
From a docs POV, LGTM. I'd rather see the content merged now. Unlikely to get to this level of reorg for the 0.8 release, and would be good to include this content. Can copyedit (lightly) later with reorg.

ncdc:
@domderen ok I think we're ready for squashing down to 1 commit + adding the commit sign-off, then we can merge. Let us know if you need any help with git...

Issue #354, opened 2018-03-06
Should service action (restore) remove all node ports from a service?


ncdc:
We are currently removing all nodePorts from all services when we restore them:

https://github.com/heptio/ark/blob/973f630cc7b2bdd464cc4e78d776ab4099a43381/pkg/restore/service_action.go#L60-L63

This doesn't really make sense to me. @skriss do you remember what the original motivation for this was?

skriss:
IIRC it was because keeping them makes the "clone namespace" use case fail due to port collision.

ncdc:
Ok, that makes sense. I don't think they should generally be dropped by default, though. We should think through ways to allow the user to express the desired behavior in a clean manner.

ncdc:
cc @jbeda - another UX question. How do we support both 1) restore into new cluster and 2) restore into new cluster into a different namespace?

jbeda:
My answer would be similar to https://github.com/heptio/ark/issues/192#issuecomment-371653916.  By all means let's come up with granular flags for stuff like this. Or have a config file for specifying what gets remapped how.  But then package up the common use cases under a single flag.  Document it so that users can tweak it if they want.

ncdc:
Makes sense. Maybe we allow users to remap node ports as an option?
```
ark restore create --from-backup foo --namespace-mappings bob1:bob2 --node-port-mappings 40001:40002,41234:42345
```

jbeda:
That sounds great.  But from a user point of view you'd want to be able to ask "what node ports does this backup/namespace use?" and "Will those conflict with the cluster I'm trying to apply to?"  Answering those questions doesn't fit cleanly into this model unfortunately.

In addition, you may want to let the system pick the node port as you remap.  Unfortunately, it is difficult to know if the node port in the backup was assigned by the system during admission control or if it was assigned explicitly by the user.  Once Daniel's big multi-merge thing goes in we may have more data.  To really tell what the intent was we'd have to start cracking into the annotation added by `kubectl apply`.

Perhaps we need some sort of a restore dry run that helps to suss out these issues and can give you a summary of the resources to be created along with perhaps testing those against the target cluster?  That is a pretty big new feature though.

ncdc:
[...]

PR #405, opened 2018-03-29
Compare for duplicates before logging object exists errors

nrb:
When restoring resources that raise an already exists error, check their
equality before logging a message on the restore. If they're the same
except for some metadata, don't generate a message.

The restore process was modified so that if an object had an empty
namespace string, no namespace key is created on the object. This was to
avoid manipulating the copy of the current cluster's object by adding
the target namespace.

There are some cases right now that are known to not be equal via this
method:

- The `default` ServiceAccount in a namespace will not match, primarily
because of differing default tokens. These will be handled in their own
patch
- IP addresses for Services are recorded in the backup object, but are
either not present on the cluster object, or different. An issue for
this already exists at #354
- Endpoints have differing values for `renewTime`. This may be
insubstantial, but isn't currently handled by the resetMetadataAndStatus
function.
- PersistentVolume objects in the backup have a `claimRef` section,
while those in cluster do not.

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -31,6 +31,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -634,20 +635,37 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 		}
 
 		// necessary because we may have remapped the namespace
-		obj.SetNamespace(namespace)
+		// if the namespace is blank, don't create the key
+		if namespace != "" {
+			obj.SetNamespace(namespace)
+		}
 
 		// add an ark-restore label to each resource for easy ID
 		addLabel(obj, api.RestoreLabelKey, ctx.restore.Name)
 
 		ctx.infof("Restoring %s: %v", obj.GroupVersionKind().Kind, obj.GetName())
-		_, err = resourceClient.Create(obj)
-		if apierrors.IsAlreadyExists(err) {
-			addToResult(&warnings, namespace, err)
+		_, restoreErr := resourceClient.Create(obj)
+		if apierrors.IsAlreadyExists(restoreErr) {
+			equal := false
+			if fromCluster, err := resourceClient.Get(obj.GetName(), metav1.GetOptions{}); err == nil {
+				equal, err = objectsAreEqual(fromCluster, obj)
+				// Log any errors trying to check equality
+				if err != nil {
+					ctx.infof("error checking %s against cluster: %v", obj.GetName(), err)
+				}
+			} else {
+				ctx.infof("Error retrieving cluster version of %s: %v", obj.GetName(), err)
+			}
+			if !equal {
+				e := errors.Errorf("not restored: %s and is different from backed up version.", restoreErr)
+				addToResult(&warnings, namespace, e)
+			}
 			continue
 		}
-		if err != nil {
+		// Error was something other than an AlreadyExists
+		if restoreErr != nil {
 			ctx.infof("error restoring %s: %v", obj.GetName(), err)
-			addToResult(&errs, namespace, fmt.Errorf("error restoring %s: %v", fullPath, err))
+			addToResult(&errs, namespace, fmt.Errorf("error restoring %s: %v", fullPath, restoreErr))
 			continue
 		}
 
@@ -720,6 +738,25 @@ func (ctx *context) executePVAction(obj *unstructured.Unstructured) (*unstructur
 	}
 
 	return updated2, nil
+}
+
+// objectsAreEqual takes two unstructured objects and checks for equality.
+// The fromCluster object is mutated to remove any insubstantial runtime
+// information that won't match
+func objectsAreEqual(fromCluster, fromBackup *unstructured.Unstructured) (bool, error) {
+	// Remove insubstantial metadata
+	fromCluster, err := resetMetadataAndStatus(fromCluster)
+	if err != nil {
+		return false, err
+	}
+
+	// We know the cluster won't have the restore name label, so
+	// copy it over from the backup
+	restoreName := fromBackup.GetLabels()[api.RestoreLabelKey]
+	addLabel(fromCluster, api.RestoreLabelKey, restoreName)
+
+	// If there are no specific actions needed based on the type, simply check for equality.
+	return equality.Semantic.DeepEqual(fromBackup, fromCluster), nil
 }
 
 func isPVReady(obj runtime.Unstructured) bool {
--- a/pkg/restore/restore_test.go
+++ b/pkg/restore/restore_test.go
@@ -676,6 +676,61 @@ func TestResetMetadataAndStatus(t *testing.T) {
 	}
 }
 
+func TestObjectsAreEqual(t *testing.T) {
+	tests := []struct {
+		name        string
+		backupObj   *unstructured.Unstructured
+		clusterObj  *unstructured.Unstructured
+		expectedErr bool
+		expectedRes bool
+	}{
+		{
+			name:        "objects are already equal",
+			backupObj:   NewTestUnstructured().WithName("obj").WithArkLabel("test").Unstructured,
+			clusterObj:  NewTestUnstructured().WithName("obj").Unstructured,
+			expectedErr: false,
+			expectedRes: true,
+		},
+		{
+			name:        "objects reset correctly",
+			backupObj:   NewTestUnstructured().WithName("obj").WithArkLabel("test").Unstructured,
+			clusterObj:  NewTestUnstructured().WithMetadata("blah", "foo").WithName("obj").Unstructured,
+			expectedErr: false,
+			expectedRes: true,
+		},
+		{
+			name:        "cluster object has no metadata to reset",
+			backupObj:   NewTestUnstructured().WithName("obj").WithArkLabel("test").Unstructured,
+			clusterObj:  NewTestUnstructured().Unstructured,
+			expectedErr: true,
+			expectedRes: false,
+		},
+		{
+			name:        "Test JSON objects",
+			backupObj:   unstructuredOrDie(`{"apiVersion":"v1","kind":"ServiceAccount","metadata":{"name":"default","namespace":"nginx-example", "labels": {"ark-restore": "test"}},"secrets":[{"name":"default-token-xhjjc"}]}`),
+			clusterObj:  unstructuredOrDie(`{"apiVersion":"v1","kind":"ServiceAccount","metadata":{"creationTimestamp":"2018-04-05T20:12:21Z","name":"default","namespace":"nginx-example","resourceVersion":"650","selfLink":"/api/v1/namespaces/nginx-example/serviceaccounts/default","uid":"a5a3d2a2-390d-11e8-9644-42010a960002"},"secrets":[{"name":"default-token-xhjjc"}]}`),
+			expectedErr: false,
+			expectedRes: true,
+		},
+		{
+			name:        "Test ServiceAccount secrets mismatch",
+			backupObj:   unstructuredOrDie(`{"apiVersion":"v1","kind":"ServiceAccount","metadata":{"name":"default","namespace":"nginx-example", "labels": {"ark-restore": "test"}},"secrets":[{"name":"default-token-abcde"}]}`),
+			clusterObj:  unstructuredOrDie(`{"apiVersion":"v1","kind":"ServiceAccount","metadata":{"creationTimestamp":"2018-04-05T20:12:21Z","name":"default","namespace":"nginx-example","resourceVersion":"650","selfLink":"/api/v1/namespaces/nginx-example/serviceaccounts/default","uid":"a5a3d2a2-390d-11e8-9644-42010a960002"},"secrets":[{"name":"default-token-xhjjc"}]}`),
+			expectedErr: false,
+			expectedRes: false,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			res, err := objectsAreEqual(test.clusterObj, test.backupObj)
+			if assert.Equal(t, test.expectedErr, err != nil) {
+				assert.Equal(t, test.expectedRes, res)
+			}
+		})
+	}
+}
+
 func TestExecutePVAction(t *testing.T) {
 	iops := int64(1000)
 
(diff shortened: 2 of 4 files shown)


pkg/restore/restore.go
```diff
@@ -634,16 +635,29 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
 
 		// add an ark-restore label to each resource for easy ID
 		addLabel(obj, api.RestoreLabelKey, ctx.restore.Name)
 
 		ctx.infof("Restoring %s: %v", obj.GroupVersionKind().Kind, obj.GetName())
 		_, err = resourceClient.Create(obj)
 		if apierrors.IsAlreadyExists(err) {
-			addToResult(&warnings, namespace, err)
-			continue
+			fromCluster, _ := resourceClient.Get(obj.GetName(), metav1.GetOptions{})
```
skriss: need to handle error

pkg/restore/restore.go
```diff
@@ -634,16 +635,29 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+			// Remove insubstantial metadata
+			fromCluster, _ = resetMetadataAndStatus(fromCluster)
```
skriss: need to handle error

pkg/restore/restore.go
```diff
@@ -634,16 +635,29 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+			// Add the restore's label to the in-cluster object since we know that's different
+			addLabel(fromCluster, api.RestoreLabelKey, ctx.restore.Name)
+
+			if equality.Semantic.DeepEqual(obj, fromCluster) {
+				continue
```
skriss: probably makes sense to do a `ctx.infof(...)` here stating that the object already exists with matching state in-cluster

pkg/restore/restore.go
```diff
@@ -634,16 +635,29 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+			} else {
+				addToResult(&warnings, namespace, err)
```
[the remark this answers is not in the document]
skriss: yeah probably makes sense to wrap

pkg/restore/restore.go
```diff
@@ -427,6 +428,41 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+	}
+
+	// Add the restore's label to the in-cluster fromBackup since we know that's different
+	addLabel(fromCluster, api.RestoreLabelKey, ctx.restore.Name)
+
+	// If there are no specific actions needed based on the type, simply check for equality.
+	if !equality.Semantic.DeepEqual(fromBackup, fromCluster) {
+		es := fmt.Sprintf("%s and backup does not match.", restoreErr.Error())
+		err := errors.New(es)
+		return err
```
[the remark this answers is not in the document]
ncdc: Yes

pkg/restore/restore.go
```diff
@@ -427,6 +428,41 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
 	return warnings, errs
 }
 
+// resolveConflict tries to resolve conflicts between a backed up object
+// and it's representation in the cluster, deciding on what action to take
+// based on type. Errors are wrapped with some additional data and returned to
+// the caller to handle. If the items are equivalent, nil is returned
+func (ctx *context) resolveConflict(fromBackup *unstructured.Unstructured, resourceClient client.Dynamic, restoreErr error) error {
```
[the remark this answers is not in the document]
nrb: That makes sense. Perhaps `resolveConflict` would be the top level that gets `fromCluster`, cleans it up, and passes it down?

pkg/restore/restore.go
```diff
@@ -634,20 +647,33 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
 		ctx.infof("Restoring %s: %v", obj.GroupVersionKind().Kind, obj.GetName())
-		_, err = resourceClient.Create(obj)
-		if apierrors.IsAlreadyExists(err) {
-			addToResult(&warnings, namespace, err)
+		_, restoreErr := resourceClient.Create(obj)
+		if apierrors.IsAlreadyExists(restoreErr) {
+			equal, err := ctx.objectsAreEqual(obj, resourceClient, restoreErr)
+			// Log any errors trying to resolve the conflict; nil means no conflic
+			if err != nil {
+				addToResult(&warnings, namespace, err)
```
ncdc: This needs to be `ctx.infof` so it goes into the restore log file.

pkg/restore/restore.go
```diff
@@ -634,20 +647,33 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+			}
+			if equal {
+				ctx.infof("Backup for %s: %v matches cluster version, skipping restore.", obj.GroupVersionKind(), obj.GetName())
```
ncdc: Do you think we need this log line?

pkg/restore/restore.go
```diff
@@ -634,20 +647,33 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+			} else {
+				e := fmt.Errorf("%s and backup does not match.", restoreErr)
```
ncdc: What does this error text end up looking like - do you have an example?

pkg/restore/restore.go
```diff
@@ -634,20 +647,33 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+				addToResult(&warnings, namespace, e)
+			}
 			continue
 		}
-		if err != nil {
+		if restoreErr != nil {
```
ncdc: Please add a comment indicating that if we get here, we got some error other than already-exists

pkg/restore/restore.go
```diff
@@ -722,6 +748,28 @@ func (ctx *context) executePVAction(obj *unstructured.Unstructured) (*unstructur
...
+		return false, err
+	}
+
+	// We know the cluster won't have the restore name label, so
+	// copy it over from the backup
+	restoreName := fromBackup.GetLabels()[api.RestoreLabelKey]
+	addLabel(fromCluster, api.RestoreLabelKey, restoreName)
+
+	// If there are no specific actions needed based on the type, simply check for equality.
+	if !equality.Semantic.DeepEqual(fromBackup, fromCluster) {
```
ncdc: `return equality.Semantic.DeepEqual(fromBackup, fromCluster), nil`

pkg/restore/restore.go
```diff
@@ -427,6 +428,18 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
 	return warnings, errs
 }
 
+// objectsAreEqual indentifies whether a backed up object
+// and its representation in the cluster are equal
+func (ctx *context) objectsAreEqual(fromBackup *unstructured.Unstructured, resourceClient client.Dynamic, restoreErr error) (bool, error) {
```
ncdc: I don't think we need to pass in `restoreErr` any more?

ncdc: Sorry for going back and forth on this, but I'm wondering if it would just be clearer to move the `Get()` into the `IsAlreadyExists` block and get rid of this helper. WDYT?

pkg/restore/restore.go
nrb: Here's a snippet from a restore:

```
Warnings:
  Ark:        <none>
  Cluster:  persistentvolumes "pvc-c5c8bb8c-3758-11e8-a959-42010a960060" already exists and backup does not match.
            persistentvolumes "pvc-f07ce647-336d-11e8-9d79-42010a96006f" already exists and backup does not match.
  Namespaces:
    default:        services "kubernetes" already exists and backup does not match.
    kube-system:    endpoints "kube-controller-manager" already exists and backup does not match.
                    endpoints "kube-scheduler" already exists and backup does not match.
                    services "default-http-backend" already exists and backup does not match.
                    services "heapster" already exists and backup does not match.
                    services "kube-dns" already exists and backup does not match.
                    services "kubernetes-dashboard" already exists and backup does not match.
    nginx-example:  services "my-nginx" already exists and backup does not match.
```

pkg/restore/restore.go
nrb: My very first pass did that. The reason I named this function `resolveConflict` was because I envisioned adding more complex cases like service accounts there later. That way, the resolution was mostly out of the `IsAlreadyExists` block.

Doing the `Get()` in the block seems reasonable to me, and some function that contains different actions or policy checks added when it's needed.

pkg/restore/restore.go
nrb: I'm not particularly attached to it. Fine with deleting it.

pkg/restore/restore.go
ncdc: Thanks

pkg/restore/restore.go
```diff
@@ -634,20 +635,36 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
 		addLabel(obj, api.RestoreLabelKey, ctx.restore.Name)
 
 		ctx.infof("Restoring %s: %v", obj.GroupVersionKind().Kind, obj.GetName())
-		_, err = resourceClient.Create(obj)
-		if apierrors.IsAlreadyExists(err) {
-			addToResult(&warnings, namespace, err)
+		_, restoreErr := resourceClient.Create(obj)
+		if apierrors.IsAlreadyExists(restoreErr) {
+			fromCluster, err := resourceClient.Get(obj.GetName(), metav1.GetOptions{})
+			if err != nil {
```
ncdc: I'm thinking of something like this:
```go
equal := false

if fromCluster, err := resourceClient.Get(obj.GetName(), metav1.GetOptions{}); err == nil {
  equal, err = objectsAreEqual(fromCluster, obj)
  if err != nil {
    // Log any errors trying to check equality
    ctx.infof(err.Error())
  }
} else {
  ctx.infof("error getting from-cluster...")
}

if !equal {
  e := fmt.Errorf("%s and backup does not match.", restoreErr)
  addToResult(&warnings, namespace, e)
}
```

This way, we only call `objectsAreEqual` if we successfully retrieve `fromCluster`. No matter what, if we get an error anywhere in here, it gets logged, and we add the warning in all cases unless the objects are equal. WDYT?

nrb: Yeah, that makes sense.

pkg/restore/restore_test.go
```diff
@@ -676,6 +676,45 @@ func TestResetMetadataAndStatus(t *testing.T) {
...
+			name:        "cluster object has no metadata to reset",
+			backupObj:   NewTestUnstructured().WithName("obj").WithArkLabel("test").Unstructured,
+			clusterObj:  NewTestUnstructured().Unstructured,
+			expectedErr: true,
+			expectedRes: false,
+		},
+	}
+
+	for _, test := range tests {
+		res, err := objectsAreEqual(test.clusterObj, test.backupObj)
```
ncdc: Please run in a subtest: `t.Run(test.name, func(t *testing.T) { ... })`

pkg/restore/restore_test.go
```diff
@@ -897,6 +936,15 @@ func (obj *testUnstructured) WithName(name string) *testUnstructured {
 	return obj.WithMetadataField("name", name)
 }
 
+func (obj *testUnstructured) WithArkLabel(restoreName string) *testUnstructured {
+	ls := make(map[string]string)
```
ncdc: I would check if the `obj` already has labels and merge if it does.

pkg/restore/restore_test.go
```diff
@@ -676,6 +676,45 @@ func TestResetMetadataAndStatus(t *testing.T) {
...
+			expectedErr: false,
+			expectedRes: true,
+		},
+		{
+			name:        "cluster object has no metadata to reset",
+			backupObj:   NewTestUnstructured().WithName("obj").WithArkLabel("test").Unstructured,
+			clusterObj:  NewTestUnstructured().Unstructured,
+			expectedErr: true,
+			expectedRes: false,
+		},
```
ncdc: It might be worth getting the json output from a real cluster of something like a secret or service or pod and putting that in here as a test case. We have a test helper to convert json to Unstructured.

pkg/restore/restore.go
```diff
@@ -722,6 +740,23 @@ func (ctx *context) executePVAction(obj *unstructured.Unstructured) (*unstructur
 	return updated2, nil
 }
 
+// objectsAreEqual takes two unstructured objects and checks for equality.
```
ncdc: I'd add a comment that `fromCluster` is mutated

pkg/restore/restore.go
```diff
@@ -634,20 +635,37 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+				equal, err = objectsAreEqual(fromCluster, obj)
+				// Log any errors trying to check equality
+				if err != nil {
+					ctx.infof(err.Error())
+				}
+			} else {
+				ctx.infof("Error retrieving cluster version of %s: %v", obj.GetName(), err)
+			}
+			if !equal {
+				e := fmt.Errorf("%s and backup does not match.", restoreErr)
```
ncdc: We're trying to use `errors.Errorf` (from `github.com/pkg/errors`) wherever possible.

nrb:
@ncdc @skriss  Latest comments addressed, PTAL when you can.

pkg/restore/restore.go
```diff
@@ -634,20 +635,37 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
-		if apierrors.IsAlreadyExists(err) {
-			addToResult(&warnings, namespace, err)
+		_, restoreErr := resourceClient.Create(obj)
+		if apierrors.IsAlreadyExists(restoreErr) {
+			equal := false
+			if fromCluster, err := resourceClient.Get(obj.GetName(), metav1.GetOptions{}); err == nil {
+				equal, err = objectsAreEqual(fromCluster, obj)
+				// Log any errors trying to check equality
+				if err != nil {
+					ctx.infof(err.Error())
```
skriss: I'd probably make this something like `ctx.infof("error checking %s for equality: %v", obj.getName(), err)`

pkg/restore/restore.go
```diff
@@ -634,20 +635,37 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+				}
+			} else {
+				ctx.infof("Error retrieving cluster version of %s: %v", obj.GetName(), err)
+			}
+			if !equal {
+				e := errors.Errorf("%s and backup does not match.", restoreErr)
```
skriss: wordsmithing here, but since this is directly part of the UX it's important to make this super clear - maybe something like "<type> <name> not restored: <restoreErr>, and is different than backed-up version."  I think it's important to make clear that it wasn't restored, because it already exists, and there's a diff (so the user might want to manually remediate).

pkg/restore/restore.go
```diff
@@ -722,6 +740,25 @@ func (ctx *context) executePVAction(obj *unstructured.Unstructured) (*unstructur
 	return updated2, nil
 }
 
+// objectsAreEqual takes two unstructured objects and checks for equality.
+// The fromCluster object is copy that is mutated to remove any insubstantial runtime
```
ncdc: I would remove `is copy that`

skriss:
LGTM!

PR #409, opened 2018-04-02
get backup name from restore spec when handling restore DownloadRequests

### LINKED ISSUE #398: Incorrect pre-signed urls generated for named restores
If you do `ark restore create some-name --from-backup foo`, we end up generating the wrong pre-signed urls for `ark restore describe` and `ark restore logs`. We need to generate a url for (bucket, backup "directory", file), and the code assumes that all restores are named like `<backup name>-<timestamp>`. But for user-named restores, that assumption isn't necessarily valid. If you have a restore named `r1` for a backup `b1`, the url needs to be (bucket, b1, r1-logs.gz), for example, but it ends up being (bucket, r1, r1-logs.gz).

skriss:

Fixes #398

### DIFF
--- a/pkg/controller/download_request_controller.go
+++ b/pkg/controller/download_request_controller.go
@@ -47,6 +47,8 @@ type downloadRequestController struct {
 	downloadRequestClient       arkv1client.DownloadRequestsGetter
 	downloadRequestLister       listers.DownloadRequestLister
 	downloadRequestListerSynced cache.InformerSynced
+	restoreLister               listers.RestoreLister
+	restoreListerSynced         cache.InformerSynced
 	backupService               cloudprovider.BackupService
 	bucket                      string
 	syncHandler                 func(key string) error
@@ -59,6 +61,7 @@ type downloadRequestController struct {
 func NewDownloadRequestController(
 	downloadRequestClient arkv1client.DownloadRequestsGetter,
 	downloadRequestInformer informers.DownloadRequestInformer,
+	restoreInformer informers.RestoreInformer,
 	backupService cloudprovider.BackupService,
 	bucket string,
 	logger logrus.FieldLogger,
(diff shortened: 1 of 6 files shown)


ncdc:
@skriss this works, or we could modify the `DownloadTarget` struct to have a field for the name of the backup. Modifying the struct seems like a lot less code, but it does mean that we'd have to resolve `Name` and `BackupName` - would we keep `Name` or change it to `RestoreName`? WDYT re approach?

skriss:
Yeah, I thought about going that way as well - I couldn't think of any significant pros/cons between them so just went with the one that didn't change the API. If we go that approach, we'd presumably still want to fetch the restore's backup and automatically fill that in, so that code would just move from the controller into the `downloadrequest.Stream` func.

ncdc:
I'd put it into the `ark restore <cmd>` code, and just fail the download request in the controller if the required field(s) are missing.

ncdc:
If we go that route

pkg/controller/download_request_controller.go
```diff
@@ -214,8 +219,24 @@ const signedURLTTL = 10 * time.Minute
...
-	var err error
-	update.Status.DownloadURL, err = c.backupService.CreateSignedURL(downloadRequest.Spec.Target, c.bucket, signedURLTTL)
+	var (
+		directory string
+		err       error
+	)
+
+	switch downloadRequest.Spec.Target.Kind {
+	case v1.DownloadTargetKindRestoreLog, v1.DownloadTargetKindRestoreResults:
+		restore, err := c.restoreLister.Restores(v1.DefaultNamespace).Get(downloadRequest.Spec.Target.Name)
```
ncdc: s/v1.DefaultNamespace/downloadRequest.Namespace/

pkg/controller/download_request_controller.go
```diff
@@ -214,8 +219,24 @@ const signedURLTTL = 10 * time.Minute
...
+		if err != nil {
+			return errors.Wrap(err, "error getting Restore")
+		}
+
+		directory = restore.Spec.BackupName
+	default:
+		directory = downloadRequest.Spec.Target.Name
```
ncdc: This flow technically isn't used anywhere - do you think it's worth it?

ncdc:
@skriss let's just go with this approach. I'd rather not have the messy `name` vs `restoreName` api change.

pkg/controller/download_request_controller.go
skriss: The `default` case? That should catch anything backup-related (logs, tarball)

ncdc:
It’s not used, fwiw

skriss:
Ah, I see what you mean. I removed that case and added a comment to godoc about if/how `directory` is used within `BackupService`.

ncdc:
LGTM

PR #412, opened 2018-04-03
use cobra's arg-count validation & call Complete() before Validate()

### LINKED ISSUE #410: Use cobra.ExactArgs where appropriate
If we know a command requires a specific number of arguments, we can use `cobra.ExactArgs()` instead of manually checking the count.

### LINKED ISSUE #411: Update cmd flow
Should be
1. Complete()
1. Validate()
1. Run()

skriss:

Fixes #410 
Fixes #411

### DIFF
--- a/pkg/cmd/cli/backup/create.go
+++ b/pkg/cmd/cli/backup/create.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/pkg/errors"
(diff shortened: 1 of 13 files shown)


skriss:
yeah, complete probably makes more sense.

skriss:
updated - looked like only the one.

ncdc:
LGTM

PR #414, opened 2018-04-06
add helper function to compare slices of actions reliably and fix test flake

skriss:

Fixes a unit test flake.

### DIFF
--- a/pkg/util/test/comparisons.go
+++ b/pkg/util/test/comparisons.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2018 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"reflect"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


ncdc:
@skriss what if we just extracted this as a test helper method, and used it here too? https://github.com/heptio/ark/blob/25d46a709b5ef00072d26bb9075ff76600d34dcd/pkg/controller/backup_deletion_controller_test.go#L435-L459

skriss:
Ah, nice. done.

pkg/util/test/comparisons.go
```diff
@@ -0,0 +1,45 @@
+package test
```
ncdc: Please add boilerplate header

pkg/util/test/comparisons.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/stretchr/testify/assert"
+
+	core "k8s.io/client-go/testing"
+)
+
+// CompareActions checks slices of actual and expected Actions
+// for equality. It checks that the lengths of the slices are
+// the same, that each actual Action has a corresponding expected
+// Action, and that each expected Action has a corresponding
+// actual Action.
```
ncdc: Should we add a sentence saying that it doesn't validate the order?

pkg/util/test/comparisons.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, e := range expected {
+		found := false
+		for _, a := range actual {
+			if reflect.DeepEqual(e, a) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			t.Errorf("missing expected action %#v", e)
```
ncdc: One thing that's annoying about `%#v` for `PatchAction`s is that it prints out the patch bytes in hex. We don't need to do this now, but it might be nice to have a custom printer for `PatchAction`s.

ncdc:
LGTM

PR #415, opened 2018-04-09
add nodeSelector so that ark server doesn't end up on a windows node …

ffd2subroutine:
Add nodeSelector so that ark server doesn't end up on a windows node when running a hybrid cluster on Azure.

Signed-off-by: ffd2subroutine <ffd2subroutine@users.noreply.github.com>


ncdc:
Hi @ffd2subroutine, thanks for your PR!

ncdc:
Restarting travis - it hit a flakey test. Will merge when it's green.

PR #413, opened 2018-04-06
[WIP] replace ark's map_utils with kubernetes unstructured helpers

skriss:
Signed-off-by: Steve Kriss <steve@heptio.com>

Replace Ark's `map_utils.go` with Kubernetes' unstructured helpers.

Unit tests are passing but I need to re-review and do some manual testing.

PR #418, opened 2018-04-10
use typed structs for decoding patch JSON in unit tests

skriss:

As I was looking at moving the `hasKeyAndVal` func around as part of #413 I realized that the patch unit tests (which are the only thing that use this func) can be simplified by using typed structs for decoding the patches. Updated all of our tests. I think it's cleaner.

### DIFF
--- a/pkg/controller/schedule_controller_test.go
+++ b/pkg/controller/schedule_controller_test.go
@@ -18,7 +18,6 @@ package controller
 
 import (
 	"encoding/json"
-	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


ncdc:
```
pkg/util/test/comparisons.go:75:9: decoder.DisallowUnknownFields undefined (type *json.Decoder has no field or method DisallowUnknownFields)
```

Is this a go1.10 addition?

skriss:
oh, yeah, looks like it.

skriss:
@ncdc we should be able to upgrade to go 1.10 after v0.8.0 gets out the door, right?

ncdc:
Definitely!

skriss:
rebased for go 1.10, should be good to merge if CI is green.

pkg/controller/schedule_controller_test.go
```diff
@@ -239,27 +238,13 @@ func TestProcessSchedule(t *testing.T) {
...
-				res, _ := collections.GetMap(patch, "status")
-				assert.Equal(t, 1, len(res), "patch's status has the wrong number of keys")
+				expected := Patch{
+					Status: PatchStatus{
+						LastBackup: parseTime(test.expectedLastBackup),
+					},
+				}
 
-				index++
+				arktest.ValidatePatch(t, actions[index], expected, decode)
```
ncdc: We still need to `index++`, right?

skriss: We don't need this one since this is the last action to be validated (this isn't in a loop, there are just 3 static action checks). Was probably there before just from copy-pasting each block.

ncdc:
@skriss why'd you close?

ncdc:
LGTM

PR #425, opened 2018-04-13
Reverting namespace to single namespace deployment

lypht:
Signed-off-by: Josh Hull <josh.hull@lypht.com>


ncdc:
There's 3 other occurrences in the repo of `heptio-ark-server` - mind fixing all of them? Thanks!

PR #427, opened 2018-04-15
Add examples/common/10-deployment.yaml back

containscafeine:
The changes in https://github.com/heptio/ark/commit/1db966bf3c128fb5e5482d99d2504a7fa5d4d0d8#diff-c31a634b9c201723f2478087620b10c5 removed examples/common/10-deployment.yaml, which is listed in
README.md to start the server  at -
https://github.com/heptio/ark#set-up-server

This commit adds that file back so that things still work.

Signed-off-by: Shubham <shubham@linux.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,6 @@ NOTE: Make sure to check out the appropriate version. We recommend that you chec
     ```bash
     kubectl apply -f examples/common/00-prereqs.yaml
     kubectl apply -f examples/minio/
-    kubectl apply -f examples/common/10-deployment.yaml
     ```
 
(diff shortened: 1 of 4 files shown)


ncdc:
Hi @containscafeine, instead of restoring the file in `examples/common`, would you please put it as `examples/minio/20-ark-deployment.yaml` and update `README.md` as needed?

skriss:
Note: `examples/common/10-deployment.yaml` is also referenced [here](https://github.com/heptio/ark/blob/master/examples/common/README.md)

containscafeine:
@ncdc @skriss made the changes, PTAL

ncdc:
Thanks!

PR #429, opened 2018-04-16
GCP: create/use IAM role with minimal set of permissions

### LINKED ISSUE #393: GCP Bucket Permissions Documentation
Just as with [issue #49](https://github.com/heptio/ark/issues/49), there should be a documentation for the precise gcp bucket permissions ark requires to function.

Here are some good documentation entry points:
[GCS Access Control](https://cloud.google.com/storage/docs/access-control/)
[GCS Using IAM Permissions](https://cloud.google.com/storage/docs/access-control/using-iam-permissions)

skriss:

Fixes #393 

I've tested this and it looks good, but would like to do a full teardown/re-test before merging.

### DIFF
--- a/docs/gcp-config.md
+++ b/docs/gcp-config.md
@@ -44,12 +44,29 @@ To integrate Heptio Ark with GCP, create an Ark-specific [Service Account][15]:
 3. Attach policies to give `heptio-ark` the necessary permissions to function:
 
     ```bash
+    BUCKET=<YOUR_BUCKET>
+    
+    ROLE_PERMISSIONS=(
+        compute.disks.get
+        compute.disks.create
+        compute.disks.createSnapshot
+        compute.snapshots.get
+        compute.snapshots.create
+        compute.snapshots.useReadOnly
+        compute.snapshots.delete
+        compute.projects.get
+    )
+
+    gcloud iam roles create heptio_ark.server \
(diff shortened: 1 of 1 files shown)


skriss:
Checked this again and LGTM. It's only a docs change so can go in whenever.

docs/gcp-config.md
```diff
@@ -42,14 +42,31 @@ To integrate Heptio Ark with GCP, create an Ark-specific [Service Account][15]:
     Set the `$SERVICE_ACCOUNT_EMAIL` variable to match its `email` value.
 
 3. Attach policies to give `heptio-ark` the necessary permissions to function:
-
+
```
ncdc: Don't need the indent here?

ncdc:
@nrb could you please test this?

nrb:
This appears to work correctly on backups/restores created after the IAM policy was applied.

However, I had 2 backups in the bucket prior to applying the policy, and they're both stuck in the `Deleting` phase now. GCP snapshots and the objects in the bucket are cleaned up, so it looks like only the backup CRD remains. The `DeleteBackupRequest` objects are also gone by the time I look.

I'm not positive that the IAM policy change do this, but I also haven't observed this in v0.8.0-rc.1 anywhere else, either.

skriss:
@nrb what's in the server logs?

PR #431, opened 2018-04-18
Remove finalizers from backups before syncing

nrb:

### DIFF
--- a/pkg/controller/backup_sync_controller_test.go
+++ b/pkg/controller/backup_sync_controller_test.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/heptio/ark/pkg/apis/ark/v1"
 	"github.com/heptio/ark/pkg/generated/clientset/versioned/fake"
+	"github.com/heptio/ark/pkg/util/stringslice"
 	arktest "github.com/heptio/ark/pkg/util/test"
 )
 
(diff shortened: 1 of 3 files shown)


skriss:
Code lgtm. Please add unit test(s).

skriss:
(lmk if you need any direction on the units!)

pkg/controller/backup_sync_controller_test.go
```diff
@@ -77,6 +90,7 @@ func TestBackupSyncControllerRun(t *testing.T) {
 
 			// we only expect creates for items within the target bucket
 			for _, cloudBackup := range test.cloudBackups {
+				cloudBackup.Finalizers = stringslice.Except(cloudBackup.Finalizers, gcFinalizer)
```
skriss: I'd add a comment here just stating that we expect the controller to remove the gc finalizer.

It feels slightly weird that the code here to prep the expectation is identical to the code in the test, but I think given that we're separately testing `stringslice.Except`, what we're verifying here is just that it's actually called, so I think this is fine as-is.

pkg/util/test/test_backup.go
```diff
@@ -124,3 +124,9 @@ func (b *TestBackup) WithResourceVersion(version string) *TestBackup {
 	b.ResourceVersion = version
 	return b
 }
+
+func (b *TestBackup) WithFinalizer(finalizers ...string) *TestBackup {
```
skriss: nit: s/WithFinalizer/WithFinalizers/

pkg/controller/backup_sync_controller_test.go
nrb: > It feels slightly weird that the code here to prep the expectation is identical to the code in the test, but I think given that we're separately testing stringslice.Except, what we're verifying here is just that it's actually called, so I think this is fine as-is.

Right, I agree. An alternative may be to call `!stringslice.Has(cloudBackup.Finalizers, gcFinalizer)` on each backup, in addition to verifying the backup passed in to the create action. WDYT?

skriss: I think it's ok as-is. This test fails if the wrong finalizer is removed, or if no finalizer is removed, which is what we want to see.

nrb: Actually, looking at this more, we're doing the exact same thing the code inside the `c.run()` call is doing, right? We're not even verifying it, really. So I think doing the `Has` check makes more sense.

PR #435, opened 2018-04-19
docker tag and push :latest when building a git tag

skriss:

Since we're no longer using the Jenkins docker build/publish plugin, we need to tag and push `:latest` in our Makefile.

Ref. https://github.com/heptio/sonobuoy/blob/master/Makefile#L84-L92


ncdc:
LGTM

PR #437, opened 2018-04-19
Add library code to install required server components

marpaia:
This PR includes Go library code to assist with the installation of the following required server-side components:

- [x] Namespace
- [x] Custom Resource Definitions
- [x] Service Account
- [x] Cluster Role Binding
- [x] Config
- [x] Deployment


ncdc:
@marpaia thanks for this!

We support running the Ark server in any namespace you want - would you please make anything in here that hard-codes `heptio-ark` as the namespace take in a `namespace` parameter?

marpaia:
@ncdc I added a namespace parameter where needed. Can you also take a look at the deployment spec? It's a little GCP centric right now with the google credentials file, but I'm wondering if you have any insight into what the best way to make the logic of that function more generic given how Ark usually gets deployed.

marpaia:
Reading through the example configurations for other cloud providers, maybe it makes sense to have a set of `GCPDeployment`, `AzureDeployment`, etc functions? It's a little random from provider to provider which needs which environment variables, secrets, node annotations, etc.

ncdc:
@marpaia we've been thinking about adjusting how we deal with credentials for providers. We currently support 1 provider at a time, either with the `/credentials/cloud` volume mount / file, or with something else such as an EC2 instance profile. Where I think I'd like to end up is something like this:

1. Stop having the user manually set the cloud specific environment variable in the Ark deployment
    1. But still allow the user to set it manually if desired
1. If the env var isn't manually set:
    1. Make each cloud provider plugin for Ark expect to find a credentials file in a well-known location, such as `/credentials/aws`, `/credentials/gcp`, etc.
    1. Have the plugin set the env var to point at this well-known location

The long-term plan is to remove the `backupStorageProvider` and `persistentVolumeProvider` sections from the Ark config, splitting them out into their own custom resources (`BackupTarget` and `VolumeTarget`).

ncdc:
cc @nrb

nrb:
Thanks @ncdc. Double checking here - in this proposed set up, `BackupTarget`s and `VolumeTarget`s would share the same credential, correct? We'll start with assuming that there's only 1 set of creds for a provider and iterate if it's necessary later.

ncdc:
Correct

marpaia:
@ncdc is there anything concrete that you need from me on this PR in order for it to be merged?

ncdc:
@marpaia I have to apologize but I've been laser focused on prepping for my KubeCon talk+demo next week. Assigning to @skriss to see this through. Thanks for your patience!

marpaia:
No worries, good luck!

marpaia:
@ncdc hey, just wanted to follow up post-kubecon on next steps here to get this upstreamed. We have some cluster bootstrap tools at work that depend on my fork directly for these functions and I would love to make any modifications necessary to upstream these.

ncdc:
@marpaia I'll take another look in a few - thanks for the prompt

pkg/resources/config.go
```diff
@@ -0,0 +1,44 @@
...
+	"time"
+
+	arkv1 "github.com/heptio/ark/pkg/apis/ark/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Config(
+	namespace string,
+	cloudProviderName string,
```
ncdc: I would have 1 for backup storage & 1 for pv storage - the configs are different

pkg/resources/access.go
```diff
@@ -0,0 +1,42 @@
+package resources
```
ncdc: I'm a bit hesistant about `pkg/resources`. All of these things are installation specific. Should we consider `pkg/install` maybe? cc @skriss

pkg/resources/crd.go
```diff
@@ -0,0 +1,124 @@
...
+	"fmt"
+
+	arkv1 "github.com/heptio/ark/pkg/apis/ark/v1"
+
+	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ListCRDs returns a list of the CRD types for all of the required Ark CRDs
+func ListCRDs() []*apiextv1beta1.CustomResourceDefinition {
```
ncdc: Should we just call this `CRDs()`?

pkg/resources/crd.go
```diff
@@ -0,0 +1,124 @@
...
+		scheduleCRD(),
+		restoreCRD(),
+		configCRD(),
+		downloadRequestCRD(),
+		deleteBackupRequestsCRD(),
+	}
+}
+
+func backupCRD() *apiextv1beta1.CustomResourceDefinition {
+	return &apiextv1beta1.CustomResourceDefinition{
```
ncdc: Would you want to create a helper function that takes in a `resource string` and then returns a CRD? Everything is the same except for objectMeta.name and spec.names.plural & .kind. We could uppercase the first letter to get kind. WDYT?

pkg/resources/deployment.go
```diff
@@ -0,0 +1,74 @@
...
+	appsv1beta1 "k8s.io/api/apps/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Deployment(namespace string) *appsv1beta1.Deployment {
+	return &appsv1beta1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "ark",
+			Namespace: namespace,
```
ncdc: We should probably label the deployment itself with component=ark too. Not sure offhand if we do that in our yaml...

pkg/resources/deployment.go
```diff
@@ -0,0 +1,74 @@
...
+						"component": "ark",
+					},
+				},
+				Spec: corev1.PodSpec{
+					RestartPolicy:      corev1.RestartPolicyAlways,
+					ServiceAccountName: "ark",
+					Containers: []corev1.Container{
+						{
+							Name:            "ark",
+							Image:           "gcr.io/heptio-images/ark:latest",
```
ncdc: We should parameterize either this entire value or at least the tag.

pkg/resources/deployment.go
```diff
@@ -0,0 +1,74 @@
...
+							ImagePullPolicy: corev1.PullAlways,
```
ncdc: I usually prefer PullIfNotPresent. WDYT?

pkg/resources/deployment.go
```diff
@@ -0,0 +1,74 @@
...
+									MountPath: "/credentials",
+								},
+								{
+									Name:      "plugins",
+									MountPath: "/plugins",
+								},
+							},
+							Env: []corev1.EnvVar{
+								{
+									Name:  "GOOGLE_APPLICATION_CREDENTIALS",
```
ncdc: We could add the AWS env var here too - no harm in having both. Eventually we'll remove them and make the go code set them if needed.

pkg/resources/access.go
nrb: I think `pkg/install` is more descriptive.

pkg/resources/crd.go
marpaia: Yup, good idea.

pkg/resources/deployment.go
marpaia: I always pull if not present, but the yaml hard-codes the latest tag and we can't pull if not present with latest. How about we parameterize the tag and `if tag == "latest"`, we set the `ImagePullPolicy` to pull always and use if not present otherwise?

pkg/resources/deployment.go
marpaia: Ok, is that the only thing that is different between the GCP and AWS configurations?

pkg/resources/deployment.go
ncdc: SGTM

pkg/resources/deployment.go
ncdc: Yes

marpaia:
Thanks for the reviews @ncdc, @nrb. `package install` sounds good to me, but I'm going to wait to do that for a bit, because if I do, GitHub will resolve all of your code review comments prematurely and there's a few I had follow-up questions on.

pkg/resources/deployment.go
marpaia: I think tag would be sufficient and allows us to ensure some naive degree of provenance for API users. Happy to parameterize both path and tag or just the whole image though, it's your API surface at the end of the day.

ncdc: I know when we've done testing in the past, it's been nice to use a different image. Before we make a final decision, let's let @skriss weigh in too.

marpaia: What do you all think of potentially using [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) here. I am a big fan of this pattern for variadic API configuration.

Perhaps something like:

```go
// whole image
install.Deployment("heptio-ark", install.WithImage("gcr.io/heptio-images/ark:latest"))

// just the tag
install.Deployment("heptio-ark", install.WithTag("1.0.0"))
```

marpaia:
Hey @ncdc, I have addressed the review comments over the weekend, I'd love your thoughts on the revisions whenever you get a chance.

pkg/install/resources.go
```diff
@@ -0,0 +1,161 @@
...
+type config struct {
+	container string
+	tag       string
+}
+
+func (c *config) image() string {
+	return fmt.Sprintf("%s:%s", c.container, c.tag)
+}
+
+func WithContainer(container string) Option {
```
ncdc: Let's call this `WithImage`

pkg/install/resources.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"fmt"
+
+	appsv1beta1 "k8s.io/api/apps/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1beta1 "k8s.io/api/rbac/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type Option func(*config)
```
ncdc: I don't think this needs to be exported?

Can you move all the option-related bits closer to where they're used (`Deployment()`). Also maybe we should call it `deploymentOption` instead of just `option`?

pkg/install/resources.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+func labels() map[string]string {
+	return map[string]string{
+		"component": "ark",
+	}
+}
+
+func objectMeta(name, namespace string) metav1.ObjectMeta {
```
ncdc: I typically think in terms of `namespace, name` - would you consider swapping the order?

pkg/install/resources.go
```diff
@@ -0,0 +1,161 @@
...
+								{
+									Name:  "AWS_SHARED_CREDENTIALS_FILE",
+									Value: "/credentials/cloud",
+								},
+							},
+						},
+					},
+					Volumes: []corev1.Volume{
+						{
+							Name: "cloud-credentials",
```
ncdc: Some deployments don't require this volume (e.g. AWS using the node's IAM instance profile credentials). Maybe we should make it a `deploymentOption`?

pkg/install/resources.go
marpaia: "image" usually implies a container and tag concatenated with a `:`. If I add `WithImage`, I would assume it should have different semantics (ie: specifying the whole image and not just the container)

ncdc: A container is the running entity that used the image for its root filesystem.

An image can either contain a tag or not. If the tag is omitted, it defaults to latest.

marpaia: ok, that makes sense. i decided to leave `WithImage` as the only functional option. The deployment will default to a pull-always pull policy and if a non-latest tag is specified, the deployment will be created with a pull-if-not-available pull policy.

pkg/install/resources.go
marpaia: is there an alternative env that has to be set if this is the case?

ncdc: For that particular case, no.

marpaia:
I added a `deploymentOption` called `install.WithoutCredentialsVolume`. If this option is included, the secret will not be mounted as a volume. Now, you can do something like:

```
deployment, err := install.Deployment(
  "heptio-ark",
  install.WithImage("gcr.io/heptio-images/ark:latest"),
  install.WithoutCredentialsVolume(),
)
```

ncdc:
cc @bryanl - prepping for `ark install` command

pkg/install/config.go
```diff
@@ -0,0 +1,46 @@
...
+func Config(
+	namespace string,
+	pvCloudProviderName string,
+	pvCloudProviderConfig map[string]string,
+	backupCloudProviderName string,
+	backupCloudProviderConfig map[string]string,
+	bucket string,
+) *arkv1.Config {
+	return &arkv1.Config{
+		ObjectMeta: metav1.ObjectMeta{
```
skriss: I'd use the `objectMeta()` func here

pkg/install/config.go
```diff
@@ -0,0 +1,46 @@
...
+		},
+		BackupStorageProvider: arkv1.ObjectStorageProviderConfig{
+			CloudProviderConfig: arkv1.CloudProviderConfig{
+				Name:   backupCloudProviderName,
+				Config: backupCloudProviderConfig,
+			},
+			Bucket: bucket,
+		},
+		BackupSyncPeriod: metav1.Duration{
+			Duration: 30 * time.Minute,
```
skriss: It would be nice to have the sync periods, and restoreonlymode, be configurable. Use the variadic config functions pattern here?

pkg/install/resources.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+func ServiceAccount(namespace string) *corev1.ServiceAccount {
+	return &corev1.ServiceAccount{
+		ObjectMeta: objectMeta(namespace, "ark"),
+	}
+}
+
+func ClusterRoleBinding(namespace string) *rbacv1beta1.ClusterRoleBinding {
```
skriss: this should be cluster-scoped, so no `namespace` arg

pkg/install/config.go
marpaia: I would be happy to contribute this, but such an API change would be completely additive. Thus, would it be possible to deliver this in another follow-up PR?

pkg/install/resources.go
marpaia: Cluster Roles are cluster-scoped, but bindings are namespace scoped afaik.

ncdc: @marpaia is correct - it's the namespace for the ark service account. We could make the service account's name configurable too but I don't think we need that right now.

pkg/install/config.go
ncdc: Follow-up is fine with me.

pkg/install/resources.go
ncdc: Yes they are cluster-scoped but you need the namespace for the subject. We do _not_ need it for the object meta

marpaia: Got it, that's updated now.

ncdc:
LGTM and thanks for your patience

PR #443, opened 2018-04-20
Add --from-schedule flag to `ark restore create`

### LINKED ISSUE #440: Cannot restore to scheduled backup by schedule name
I have one cluster in my environment scheduled to run a backup every 5 minutes. I did this via:

```
$ ark schedule create control-plane --schedule=" 0/5 * * * *" --include-namespaces=global,edge,operations,shared
```

This works, I'm getting backups that I can see from other clusters in this environment:

```
$ ark get backups
NAME                           STATUS      CREATED                         EXPIRES   SELECTOR
control-plane-20180420014525   Completed   2018-04-19 19:49:47 -0600 MDT   29d       <none>
control-plane-20180420014025   Completed   2018-04-19 19:49:47 -0600 MDT   29d       <none>
$
```

Now the help says that I can omit the timestamp:

```
$ ark create restore -h
Create a restore

Usage:
  ark create restore [RESTORE_NAME] --from-backup BACKUP_NAME [flags]

Examples:
  # create a restore named "restore-1" from backup "backup-1"
  ark restore create restore-1 --from-backup backup-1

  # create a restore with a default name ("backup-1-<timestamp>") from backup "backup-1"
  ark restore create --from-backup backup-1
```

When when I omit the timestamp when trying to restore from another cluster, I get an error that the backup isn't found:

```
$ ark create restore control-plane --from-backup control-plane
An error occurred: backups.ark.heptio.com "control-plane" not found
```

skriss:
Enables restoring from a scheduled backup. By default, the most recent backup from the series is used; a git-like offset syntax (`~2`) is also possible for specifying the nth-most recent backup.

Fixes #440

### DIFF
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"sort"
 	"sync"
 	"time"
 
@@ -32,6 +33,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -45,6 +47,7 @@ import (
 	listers "github.com/heptio/ark/pkg/generated/listers/ark/v1"
 	"github.com/heptio/ark/pkg/plugin"
 	"github.com/heptio/ark/pkg/restore"
+	"github.com/heptio/ark/pkg/util/boolptr"
 	"github.com/heptio/ark/pkg/util/collections"
 	kubeutil "github.com/heptio/ark/pkg/util/kube"
 )
@@ -252,15 +255,8 @@ func (controller *restoreController) processRestore(key string) error {
 	// don't modify items in the cache
 	restore = restore.DeepCopy()
 
-	excludedResources := sets.NewString(restore.Spec.ExcludedResources...)
-	for _, nonrestorable := range nonRestorableResources {
-		if !excludedResources.Has(nonrestorable) {
-			restore.Spec.ExcludedResources = append(restore.Spec.ExcludedResources, nonrestorable)
-		}
-	}
-
-	// validation
-	if restore.Status.ValidationErrors = controller.getValidationErrors(restore); len(restore.Status.ValidationErrors) > 0 {
+	// complete & validate restore
+	if restore.Status.ValidationErrors = controller.completeAndValidate(restore); len(restore.Status.ValidationErrors) > 0 {
 		restore.Status.Phase = api.RestorePhaseFailedValidation
 	} else {
 		restore.Status.Phase = api.RestorePhaseInProgress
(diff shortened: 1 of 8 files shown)


marpaia:
@skriss this is great, but it seems like the functionality is only available via the CLI. Consider the following "restore" as implemented in my program:

```go
func (a *Ark) RestoreControlPlane() error {
	backupList, err := a.arkClient.Backups("heptio-ark").List(metav1.ListOptions{
		LabelSelector: "ark-schedule=control-plane",
	})
	if err != nil {
		return errors.Wrap(err, "error listing backups in heptio-ark")
	}

	if len(backupList.Items) < 1 {
		return errors.New("no backups found")
	}

	backups := make([]string, len(backupList.Items))
	for idx, backup := range backupList.Items {
		backups[idx] = backup.Name
	}

	sort.Strings(backups)
	latest := backups[len(backups)-1]

	level.Debug(a.logger).Log(
		"msg", "restoring control plane",
		"backup", latest,
	)

	restore := &arkv1.Restore{
			Name:      "control-plane",
			Namespace: "heptio-ark",
		},
		Spec: arkv1.RestoreSpec{
			BackupName: latest,
		},
	}
	if _, err := a.arkClient.Restores("heptio-ark").Create(restore); err != nil {
		return errors.Wrap(err, "error restoring control plane")
	}

	return nil
}
```

Would it be possible to maybe add a field to `arkv1.Restore.Spec` called `ScheduleName` which would provide this functionality more succinctly to API users?

skriss:
Good point, I think we can rejigger that, stay tuned :)

skriss:
The API change for this feature could take a couple of forms.

**Option 1:**
```
type RestoreSpec struct {
    // controller validates that exactly one
    // of BackupName and RestoreName are
    // non-empty
    BackupName string
    
    RestoreName string
    ....
}
```

**Option 2:**
```
type RestoreSpec struct {
    // Need to retain this for compatibility but deprecate 
    // for v1.0. If specified, controller should use this
    // to set RestoreSource's Backup field. If specified
    // and conflicts with a field in RestoreSource, object
    // should fail validation.
    BackupName string
    
    RestoreSource
}

// Exactly one field should be set
type RestoreSource struct {
    Backup *BackupSource

    Schedule *ScheduleSource
}

type BackupSource struct {
    // alternately could be an object ref
    BackupName string
}

type ScheduleSource struct {
    // alternately could be an object ref
    ScheduleName string
}
```

@ncdc @nrb @carlisia thoughts?

ncdc:
```
type RestoreSpec struct {
    // controller validates that exactly one
    // of BackupName and RestoreName are
    // non-empty
    BackupName string
    
    RestoreName string
    ....
}
```

Did you mean `ScheduleName` instead of `RestoreName`?

skriss:
Whoops, yes.

ncdc:
I think I'm leaning towards option 1, with the expectation that both BackupName and ScheduleName ultimately end up having values in them. If BackupName is set, we can use that. If ScheduleName is also set, we could either verify that BackupName belongs to the specified schedule, or not. If BackupName is unset and ScheduleName is set, look up the latest backup and set BackupName.

carlisia:
I'd like to throw another option for consideration, but I'm not clear on the tradeoffs since I don't know the code base too well just yet:

What if we only have `BackupName string`, and add a flag with a `backupType` boolean that would default to `backup`?

The cli cmd would look like this:

```
ark create restore [RESTORE_NAME] --from-backup [BACKUP_NAME] [flags] --backupType [BACKUP | SCHEDULED]
```

This could be the least disruptive option as far as backwards compatibility. It seems to me it could also be more self-documenting up front.

If this were to be a good option, and we could make breaking changes, I would also `s/--from-backup/--name`.

ncdc:
Because we aren't at 1.0 yet, we are allowing breaking changes, although if there's a way to do it without breakage, that's certainly better.

I prefer explicit `--from-backup` and `--from-schedule` over `--from-backup foo --backup-type schedule`. I think it's a cleaner and clearer UX.... my 2c

nrb:
I think I prefer `--from-backup` and `--from-schedule` as well. Using the boolean would be overloading the `--from-backup` flag to match a specific backup, or the latest of a schedule, which could be a little confusing, IMO.

skriss:
@ncdc @nrb I've revised this PR to add `ScheduleName` as an API field and to handle getting the right backup in the restore controller. Unit & manual tests look good; PTAL.

pkg/controller/restore_controller.go
```diff
@@ -297,46 +293,123 @@ func (controller *restoreController) processRestore(key string) error {
...
+// completed from a list of backups. Since the backups are expected
+// to be from a single schedule, "most recent" is defined as first
+// when sorted in reverse alphabetical order by name.
+func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
+	sort.Slice(backups, func(i, j int) bool {
+		// Use '>' because we want descending sort.
+		// Using Name rather than CreationTimestamp because in the case of
+		// backups synced into a new cluster, the CreationTimestamp value is
+		// time of creation in the new cluster rather than time of backup.
+		// TODO would be useful to have a new API field in backup.status
```
nrb: Would https://github.com/heptio/ark/blob/master/pkg/apis/ark/v1/backup.go#L178 be useful for that?

skriss: Ah, yes, thanks for pointing that out. That didn't exist when I first did this a couple months ago :)

pkg/controller/restore_controller_test.go
```diff
@@ -450,6 +485,171 @@ func TestProcessRestore(t *testing.T) {
...
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "e",
+			},
+			Status: api.BackupStatus{
+				Phase: api.BackupPhaseFailed,
+			},
+		},
+	}
```
nrb: Maybe add some cases where there are numerals in the names, too. Or switch to using the Backup.Status.StartTimestamp

skriss:
Updated to use  `.status.startTimestamp`.  Will squash before merge.

pkg/controller/restore_controller.go
```diff
@@ -297,46 +293,117 @@ func (controller *restoreController) processRestore(key string) error {
...
+		return false
+	}
+
+	return true
+}
+
+// mostRecentCompletedBackup returns the most recent backup that's
+// completed from a list of backups. Since the backups are expected
+// to be from a single schedule, "most recent" is defined as first
+// when sorted in reverse alphabetical order by name.
```
nrb: Comment still reflects the use of names

skriss: fixed.

PR #449, opened 2018-04-23
azure: pass full snapshot name when restoring disk

skriss:

Pass the fully-qualified snapshot name to the Azure API when restoring a disk from a snapshot, regardless of which snapshot name format (legacy or fully-qualified) is stored in the Ark backup

### DIFF
--- a/pkg/cloudprovider/azure/block_store.go
+++ b/pkg/cloudprovider/azure/block_store.go
@@ -64,6 +64,10 @@ type snapshotIdentifier struct {
 	name          string
 }
 
+func (si *snapshotIdentifier) String() string {
+	return getComputeResourceName(si.subscription, si.resourceGroup, snapshotsResource, si.name)
+}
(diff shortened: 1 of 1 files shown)


skriss:
successfully tested this on AKS.

pkg/cloudprovider/azure/block_store.go
```diff
@@ -64,6 +64,10 @@ type snapshotIdentifier struct {
 	name          string
 }
 
+func (si *snapshotIdentifier) fullName() string {
```
ncdc: Would it make more sense for this to fulfill the `Stringer` interface - change `fullName` to `String`?

skriss: i guess it wouldn't hurt

skriss:
upd.

ncdc:
LGTM

PR #452, opened 2018-04-24
Switch to upstream Go build images.

mattmoyer:
These internal gcr.io/heptio-images/golang images are deprecated. It looks like `git` and `bash` are the only things the Ark build needed that aren't in the upstream `golang:1.9-alpine3.6` image.


ncdc:
Thanks @mattmoyer!

Issue #193, opened 2017-11-09
Create GenericAWS base struct


tomplus:
I use local Ceph installation as S3 compatible storage for backups and Ark works very well. However I have to define `region` which in this case is not necessary. I see that Ark needs it to create `s3url` if someone uses AWS. Or maybe you have a plan to use it somewhere in the future (support multiple regions or something) ... ?

ncdc:
Hi @tomplus, thanks for your question. For AWS specifically, the region is required for S3 (ObjectStorageAdapter) and EBS (BlockStorageAdapter) operations. The `s3URL` is only used if you want to override aws-sdk-go's default endpoints for S3, which is necessary for something like Minio. I'm guessing you also need it for Ceph?

What happens if you specify some bogus value for region with your Ceph installation?

I can think of a couple of options:
1. Remove the requirement on `region`. I'd be concerned that people who do use AWS might accidentally omit it and it could make debugging what's wrong more difficult.
1. Add something like a `GenericAWS` ObjectStorageAdapater/BlockStorageAdapter that would then have concrete implementations for AWS, Minio, Ceph, etc. The GenericAWS base would allow the downstreams to indicate if region is required, etc.

cc @skriss @nrb @jrnt30 @dgoodwin for additional inputs.

nrb:
I like option 2. I don't know if that could be handled via the plugin work that @skriss is currently doing, either now or in future iterations.

ncdc:
@nrb it certainly could. For an external plugin, you would vendor in the `GenericAWS` code into your repo.

I wouldn't want to try to do this as part of the conversion to plugins, though.

nrb:
That makes sense. Definitely easier to do once plugins are merged.

tomplus:
@ncdc Yes, I set some bogus value and backups are stored correctly. My config:

```yaml
apiVersion: ark.heptio.com/v1
kind: Config
metadata:
  namespace: heptio-ark
  name: default
backupStorageProvider:
  bucket: backups
  aws:
    region: undef
    s3Url: http://rgw.ceph
backupSyncPeriod: 1m
gcSyncPeriod: 1m
scheduleSyncPeriod: 1m
restoreOnlyMode: false
```

I like idea of `GenericAWS` too - It would be clear and easy to use.

ncdc:
@tomplus ok great! We can consider the generic approach after #188 merges

ncdc:
[...]

PR #455, opened 2018-04-24
AWS: remove region requirement from config for backupStorageProvider

skriss:

I learned yesterday that we can do this! See https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#GetBucketRegion

I'm going to suggest leaving the examples/docs as-is for now so we don't get the usual slew of questions due to folks using master docs but an older Ark version.

### DIFF
--- a/pkg/cloudprovider/aws/object_store.go
+++ b/pkg/cloudprovider/aws/object_store.go
@@ -17,12 +17,14 @@ limitations under the License.
 package aws
 
 import (
+	"context"
 	"io"
 	"strconv"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/pkg/errors"
@@ -34,6 +36,7 @@ const (
 	s3URLKey            = "s3Url"
 	kmsKeyIDKey         = "kmsKeyId"
 	s3ForcePathStyleKey = "s3ForcePathStyle"
+	bucketKey           = "bucket"
 )
 
 type objectStore struct {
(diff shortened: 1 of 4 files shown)


skriss:
Need to do some more testing, but have done a basic level of validation already.

skriss:
I guess this would need to be tweaked to not break non-AWS S3-compatible APIs (e.g. minio)

skriss:
ref. #193

ncdc:
Yeah, we'd need to do the `GenericAWS` struct for that

skriss:
I think we could do it without the `GenericAWS` struct:
- if `s3Url` is empty, that means it's really AWS. Use this code to get the bucket's region
- if `s3Url` is not empty, it's some other AWS-compatible API. The user can specify `region` in their config if it's needed (e.g. IBM) and it'll flow through; if it's not needed, it doesn't have to be specified in config.

ncdc:
True, that could work

skriss:
@mattmoyer could you take a look at this PR -- specifically https://github.com/heptio/ark/pull/455/files#diff-b59d2b0c4d4762fbd19de3684f507a19R52 - and let us know if this looks reasonable? It'll loop through each AWS partition (public, govcloud, CN cloud) and grab a region from each to use as a "region hint" for querying the bucket's region.

skriss:
@mattmoyer / @asenchi when you get a chance could you take a quick look at this PR, given your AWS experience? thanks :)

skriss:
Region can still be specified to avoid any auto-discovery if it's causing problems for anyone.

skriss:
@ncdc this one's ready for you to review.

pkg/cmd/server/server.go
```diff
@@ -329,6 +329,11 @@ func applyConfigDefaults(c *api.Config, logger logrus.FieldLogger) {
 	} else {
 		logger.WithField("priorities", c.ResourcePriorities).Info("Using resource priorities from config")
 	}
+
+	if c.BackupStorageProvider.Config == nil {
+		c.BackupStorageProvider.Config = make(map[string]string)
+	}
+	c.BackupStorageProvider.Config["bucket"] = c.BackupStorageProvider.Bucket
```
ncdc: Please add a comment why we're doing this

pkg/cloudprovider/aws/object_store.go
```diff
@@ -46,26 +49,58 @@ func NewObjectStore() cloudprovider.ObjectStore {
...
+	return "", errors.New("unable to determine bucket's region")
+}
+
 func (o *objectStore) Init(config map[string]string) error {
 	var (
 		region              = config[regionKey]
 		s3URL               = config[s3URLKey]
 		kmsKeyID            = config[kmsKeyIDKey]
 		s3ForcePathStyleVal = config[s3ForcePathStyleKey]
+		bucket              = config[bucketKey]
```
ncdc: It's probably worth commenting that the server code copies this from backupStorageProvider.bucket and that the user doesn't have to specify it.

skriss:
I was planning to hold off on docs updates until the end of the release cycle so we don't get folks using a mismatched build vs. docs

ncdc:
I'd like to do them with the PR. Current stable docs are always at heptio.github.com/ark/

ncdc:
LGTM

PR #456, opened 2018-04-24
upgrade to go 1.10 / alpine 3.7

skriss:
Also make use of the go 1.10 build & test cache (https://golang.org/cmd/go/#hdr-Build_and_test_caching)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.9.x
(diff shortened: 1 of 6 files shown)


skriss:
Not 100% sure the changes for the build cache are correct (I think the additions are fine, just not sure about changing/removing older optimizations)

ncdc:
LGTM, thanks!

PR #460, opened 2018-04-25
add installation troubleshooting doc with common issues

skriss:

### DIFF
--- a/docs/debugging-install.md
+++ b/docs/debugging-install.md
@@ -1,0 +1,59 @@
+# Debugging Installation Issues
+
+## General
+
+### `invalid configuration: no configuration has been provided`
+This typically means that no `kubeconfig` file can be found for the Ark client to use. Ark looks for a kubeconfig in the 
+following locations:
+* the path specified by the `--kubeconfig` flag, if any
+* the path specified by the `$KUBECONFIG` environment variable, if any
+* `~/.kube/config`
+
+### Backups or restores stuck in `New` phase
+This means that the Ark controllers are not processing the backups/restores, which usually happens because the Ark server is not running. Check the pod description and logs for errors:
+```
+kubectl -n heptio-ark describe pods
+kubectl -n heptio-ark logs deployment/ark
+```
+
+
+## AWS
+
+### `NoCredentialProviders: no valid providers in chain`
+This means that the secret containing the AWS IAM user credentials for Ark has not been created/mounted properly 
+into the Ark server pod. Ensure the following:
+* The `cloud-credentials` secret exists in the Ark server's namespace
+* The `cloud-credentials` secret has a single key, `cloud`, whose value is the contents of the `credentials-ark` file
+* The `credentials-ark` file is formatted properly and has the correct values:
+    
+    ```
+    [default]
+    aws_access_key_id=<your AWS access key ID>
+    aws_secret_access_key=<your AWS secret access key>
+    ```
+* The `cloud-credentials` secret is defined as a volume for the Ark deployment
+* The `cloud-credentials` secret is being mounted into the Ark server pod at `/credentials`
+
+
+## Azure
+
+### `Failed to refresh the Token` or `adal: Refresh request failed`
+This means that the secrets containing the Azure service principal credentials for Ark has not been created/mounted 
+properly into the Ark server pod. Ensure the following:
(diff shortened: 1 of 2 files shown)


docs/debugging-install.md
```diff
@@ -0,0 +1,59 @@
+# Debugging Installation Issues
+
+## General
+
+#### `invalid configuration: no configuration has been provided`
```
Bradamant3: Bump all these up one level (3 #s, not 4) (wouldn't nit like this but I've got a couple of more substantial comments)

docs/debugging-install.md
```diff
@@ -0,0 +1,59 @@
...
+This typically means that no `kubeconfig` file can be found for the Ark client to use. Ark looks for a kubeconfig in the 
+following locations:
+* the path specified by the `--kubeconfig` flag, if any
+* the path specified by the `$KUBECONFIG` environment variable, if any
+* `~/.kube/config`
+
+#### Backups or restores stuck in `New` phase
+This implies that the Ark server is not up and running. Check the pod description and logs for errors:
```
Bradamant3: s/implies/means ? "implies" is a little vague, suggests that other things might cause the issue. IF this is indeed the case (other things), we might want to mention them. If it's not, then "means" is the right word.

docs/debugging-install.md
```diff
@@ -0,0 +1,59 @@
...
+```
+kubectl -n heptio-ark describe pods
+kubectl -n heptio-ark logs deployment/ark
+```
+
+## AWS
+#### `NoCredentialProviders: no valid providers in chain`
```
Bradamant3: Heading plus code formatting renders a bit unexpectedly. Can we describe more generically for the heading string and then put the actual error string in an intro para?  (Same for similar below)

docs/debugging-install.md
```diff
@@ -0,0 +1,59 @@
...
+
+#### `Failed to refresh the Token` or `adal: Refresh request failed`
+This means that the secrets containing the Azure service principal credentials for Ark has not been created/mounted 
+properly into the Ark server pod. Ensure the following:
+* The `cloud-credentials` secret exists in the Ark server's namespace
+* The `cloud-credentials` secret has seven keys and each one has the correct value (see [setup instructions](0))
+* The `cloud-credentials` secret is defined as a volume for the Ark deployment
+* The `cloud-credentials` secret is being mounted into the Ark server pod at `/credentials`
+
+This usually means that one of the values in the `cloud-credentials` secret is incorrect. You should double-check all of
```
Bradamant3: What does this para refer to? Is there another error? Does it apply to all the errors? It seems like another kind of error, but I can't figure out which one it might map to.

docs/debugging-install.md
skriss: It's *possible* that there could be other causes for this, but what's described here is definitely the most common and the only one I care to capture. Should I just use "means"? "almost always means"?

docs/debugging-install.md
skriss: I wanted to make it symptom-driven (if I was a user with an issue, I'd search by error). Maybe I structure each one with a generic header, then have a "symptoms" and "resolution" sub-sectio for each?

docs/debugging-install.md
skriss: Was supposed to be replaced with the content above it. Will remove :)

docs/debugging-install.md
skriss: Eh, I tried that and it adds a lot of noise. Personally I prefer it as-is - do you feel strongly about this?

Bradamant3: nope, if you tried other things and this seems like best UX, it's all good.

docs/debugging-install.md
Bradamant3: 😀

docs/debugging-install.md
Bradamant3: "means" is good. We can change it if an edge case shows up and we should doc it ;)

Bradamant3:
I'm thinking I'll pull these changes into the docs site when y'all have had a chance to look and merge, too. More food for automation thought, @nrb -- we ought to figure out a continuous improvement/publication workflow too. I have an idea or two. Down the road ...

Issue #454, opened 2018-04-24
Don't restore completed pods, jobs


ncdc:
We need to exclude completed (succeeded or failed) pods and jobs when restoring. This also means we won't restore any associated items (such as PVCs).

I think this is potentially what is happening in #450.

Plan:
- Phase 1: hard-code logic specifically for excluding completed pods and jobs in the item backupper
- Phase 2: make this pluggable, either by adding to the `ItemAction` interface, or by creating a new plugin interface

smm190:
@ncdc What is the typical release timeline for something like this?  In the meantime, do you have any ideas for a workaround we could use to backup and restore our cluster?

ncdc:
@smm190 we just released v0.8.x last week. Our typical release cycle is every 2 months, so v0.9.0 should be out in June. However, this is a fairly significant oversight, so we can hopefully get this coded and tested and released soon. No promises, but hopefully by next week if not sooner 😀.

PR #463, opened 2018-04-25
Don't restore completed pods or jobs

nrb:
Phase 1 of #454.

Backs up jobs and pods for auditing purposes, but skips them on restoring.

### DIFF
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -162,9 +162,6 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 
 	log.Info("Backing up resource")
 
-	// Never save status
-	delete(obj.UnstructuredContent(), "status")
-
 	log.Debug("Executing pre hooks")
 	if err := ib.itemHookHandler.handleHooks(log, groupResource, obj, ib.resourceHooks, hookPhasePre); err != nil {
 		return err
--- a/pkg/backup/item_backupper_test.go
+++ b/pkg/backup/item_backupper_test.go
@@ -164,13 +164,6 @@ func TestBackupItemNoSkips(t *testing.T) {
 			expectedTarHeaderName: "resources/resource.group/cluster/bar.json",
 		},
 		{
-			name:                  "make sure status is deleted",
-			item:                  `{"metadata":{"name":"bar"},"spec":{"color":"green"},"status":{"foo":"bar"}}`,
-			expectError:           false,
-			expectExcluded:        false,
-			expectedTarHeaderName: "resources/resource.group/cluster/bar.json",
-		},
-		{
 			name:                "tar header write error",
 			item:                `{"metadata":{"name":"bar"},"spec":{"color":"green"},"status":{"foo":"bar"}}`,
 			expectError:         true,
@@ -376,17 +369,14 @@ func TestBackupItemNoSkips(t *testing.T) {
 				return
 			}
 
-			// we have to delete status as that's what backupItem does,
-			// and this ensures that we're verifying the right data
-			delete(item, "status")
-			itemWithoutStatus, err := json.Marshal(&item)
+			// Convert to JSON for comparing number of bytes to the tar header
+			itemJSON, err := json.Marshal(&item)
 			if err != nil {
 				t.Fatal(err)
 			}
-
 			require.Equal(t, 1, len(w.headers), "headers")
 			assert.Equal(t, test.expectedTarHeaderName, w.headers[0].Name, "header.name")
-			assert.Equal(t, int64(len(itemWithoutStatus)), w.headers[0].Size, "header.size")
+			assert.Equal(t, int64(len(itemJSON)), w.headers[0].Size, "header.size")
 			assert.Equal(t, byte(tar.TypeReg), w.headers[0].Typeflag, "header.typeflag")
 			assert.Equal(t, int64(0755), w.headers[0].Mode, "header.mode")
 			assert.False(t, w.headers[0].ModTime.IsZero(), "header.modTime set")
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -559,6 +559,16 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 			continue
 		}
 
+		complete, err := isCompleted(obj, groupResource)
+		if err != nil {
+			addToResult(&errs, namespace, fmt.Errorf("error checking completion %q: %v", fullPath, err))
+			continue
+		}
+		if complete {
+			ctx.infof("%s is complete - skipping", kube.NamespaceAndName(obj))
+			continue
+		}
+
 		if resourceClient == nil {
 			// initialize client for this Resource. we need
 			// metadata from an object to do this.
@@ -782,8 +792,7 @@ func resetMetadataAndStatus(obj *unstructured.Unstructured) (*unstructured.Unstr
 		}
 	}
 
-	// this should never be backed up anyway, but remove it just
-	// in case.
+	// Never restore status
 	delete(obj.UnstructuredContent(), "status")
 
 	return obj, nil
@@ -812,6 +821,37 @@ func hasControllerOwner(refs []metav1.OwnerReference) bool {
 		}
 	}
 	return false
+}
+
+// TODO(0.9): These are copied from backup/item_backupper. Definititions should be moved
+// to a shared package and imported here and in the backup package.
+var podsGroupResource = schema.GroupResource{Group: "", Resource: "pods"}
+var jobsGroupResource = schema.GroupResource{Group: "batch", Resource: "jobs"}
+
+// isCompleted returns whether or not an object is considered completed.
+// Used to identify whether or not an object should be restored. Only Jobs or Pods are considered
+func isCompleted(obj *unstructured.Unstructured, groupResource schema.GroupResource) (bool, error) {
+	switch groupResource {
+	case podsGroupResource:
+		phase, _, err := unstructured.NestedString(obj.UnstructuredContent(), "status", "phase")
+		if err != nil {
+			return false, errors.WithStack(err)
+		}
+		if phase == string(v1.PodFailed) || phase == string(v1.PodSucceeded) {
+			return true, nil
+		}
+
+	case jobsGroupResource:
+		ct, found, err := unstructured.NestedString(obj.UnstructuredContent(), "status", "completionTime")
+		if err != nil {
+			return false, errors.WithStack(err)
+		}
+		if found && ct != "" {
+			return true, nil
+		}
+	}
+	// Assume any other resource isn't complete and can be restored
+	return false, nil
 }
 
 // unmarshal reads the specified file, unmarshals the JSON contained within it
--- a/pkg/restore/restore_test.go
+++ b/pkg/restore/restore_test.go
@@ -18,6 +18,7 @@ package restore
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"os"
 	"testing"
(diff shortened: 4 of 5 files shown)


pkg/backup/item_backupper_test.go
```diff
@@ -106,6 +106,35 @@ func TestBackupItemSkips(t *testing.T) {
...
+			name:  "Failed pods aren't backed up",
+			phase: "Failed",
+		},
+		{
+			name:  "Succeeded pods aren't backed up",
+			phase: "Succeeded",
+		},
+	}
+
+	for _, test := range tests {
```
nrb: @skriss what would you recommend for test logic here? I was thinking about asserting that no actions were called, since there's a few cases where `backupItem` could return nil, but I don't see a mock in the code base that would do that already.

pkg/backup/item_backupper.go
```diff
@@ -160,6 +162,29 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 	}
 	ib.backedUpItems[key] = struct{}{}
 
+	// We don't want to back up completed jobs or pods
+	if groupResource == podsGroupResource {
+		phase, _, err := unstructured.NestedString(obj.UnstructuredContent(), "status", "phase")
+		if err != nil {
+			return err
+		}
+		if phase == "Failed" || phase == "Succeeded" {
```
skriss: "k8s.io/api/core/v1".PodFailed/PodSucceeded

pkg/backup/item_backupper.go
```diff
@@ -160,6 +162,29 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 	}
 	ib.backedUpItems[key] = struct{}{}
 
+	// We don't want to back up completed jobs or pods
+	if groupResource == podsGroupResource {
+		phase, _, err := unstructured.NestedString(obj.UnstructuredContent(), "status", "phase")
+		if err != nil {
+			return err
```
skriss: errors.WithStack(err)

pkg/backup/item_backupper.go
```diff
@@ -160,6 +162,29 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
...
+		if phase == "Failed" || phase == "Succeeded" {
+			log.Info("Skipping completed pod and associated volumes")
+			return nil
+		}
+	}
+
+	if groupResource == jobsGroupResource {
+		_, found, err := unstructured.NestedString(obj.UnstructuredContent(), "status", "completionTime")
+		if err != nil {
+			return err
```
skriss: errors.WithStack(err)

pkg/backup/item_backupper_test.go
skriss: I'm thinking you probably want to add these as cases to `TestBackupItemSkips` which already has the test logic set up. I'd just confirm that adding the test cases fail before you make the code change, and pass after.

pkg/backup/item_backupper.go
ncdc: FYI it's worth checking to see if we're already importing an external package if all we need is a constant or two. You never know how much of an impact (size wise) importing a new package may be.

In this case, the pod command executor already imports this package. If it didn't, I'd recommend creating our own constants instead of importing.

pkg/backup/item_backupper.go
```diff
@@ -112,6 +114,33 @@ type defaultItemBackupper struct {
 
 var podsGroupResource = schema.GroupResource{Group: "", Resource: "pods"}
 var namespacesGroupResource = schema.GroupResource{Group: "", Resource: "namespaces"}
+var jobsGroupResource = schema.GroupResource{Group: "batch", Resource: "jobs"}
+
+// shouldBackup checks completed status on a given object and determines if the object should be backed up
+func shouldBackup(obj runtime.Unstructured, groupResource schema.GroupResource) (bool, string, error) {
```
skriss: `shouldBackup` is a little ambiguous since there are lots of reasons why a resource should/shouldn't be backed up. Maybe `isCompletedResource`?

pkg/backup/item_backupper.go
```diff
@@ -112,6 +114,33 @@ type defaultItemBackupper struct {
...
+	switch groupResource {
+
```
skriss: remove blank line

pkg/backup/item_backupper.go
```diff
@@ -160,6 +189,18 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 	}
 	ib.backedUpItems[key] = struct{}{}
 
+	backup, msg, err := shouldBackup(obj, groupResource)
```
skriss: i'd probably do away with the `msg` return value, and just log a generic "skipping completed <resourceType>" message based on the value of `backup`

pkg/backup/item_backupper.go
```diff
@@ -160,6 +189,18 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
...
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	if !backup {
+		log.Info(msg)
+		return nil
+	}
+
+	if groupResource == jobsGroupResource {
```
skriss: remove

pkg/backup/item_backupper_test.go
```diff
@@ -106,6 +106,63 @@ func TestBackupItemSkips(t *testing.T) {
...
+			groupResource: schema.GroupResource{Group: "batch", Resource: "jobs"},
+		},
+		{
+			name:          "Jobs with a completion time aren't backed up",
+			expected:      false,
+			content:       `{"apiVersion":"v1","kind":"Pod","metadata":{"namespace":"ns","name":"pod1"}, "status": {"completionTime": "bar"}}`,
+			groupResource: schema.GroupResource{Group: "batch", Resource: "jobs"},
+		},
+		{
+			name:          "Something not a pod or a job should be backed up.",
```
skriss: you could add a `phase` = Completed and non-empty `completionTime` to the status here just to validate that non-pods/jobs with those fields are backed up.

ncdc:
The more I think about it, the more I wonder if it would be better to back up completed pods/jobs, but just don't restore them. WDYT?

skriss:
I'm on board. This is consistent with our principle of backing up more information and pruning as necessary during restore.

pkg/backup/item_backupper_test.go
```diff
@@ -376,17 +368,8 @@ func TestBackupItemNoSkips(t *testing.T) {
...
-			// and this ensures that we're verifying the right data
-			delete(item, "status")
-			itemWithoutStatus, err := json.Marshal(&item)
-			if err != nil {
-				t.Fatal(err)
-			}
-
 			require.Equal(t, 1, len(w.headers), "headers")
 			assert.Equal(t, test.expectedTarHeaderName, w.headers[0].Name, "header.name")
-			assert.Equal(t, int64(len(itemWithoutStatus)), w.headers[0].Size, "header.size")
```
skriss: I'd keep this line and switch `itemWithoutStatus` to `item`

pkg/restore/restore.go
```diff
@@ -814,6 +823,35 @@ func hasControllerOwner(refs []metav1.OwnerReference) bool {
 	return false
 }
 
+// isComplete returns whether or not an object is considered completed
+// Used to identify whether or not an object should be restored. Only Jobs or Pods are considered
+func isCompleted(obj *unstructured.Unstructured, groupResource schema.GroupResource) (bool, error) {
+	podsGroupResource := schema.GroupResource{Group: "", Resource: "pods"}
+	jobsGroupResource := schema.GroupResource{Group: "batch", Resource: "jobs"}
+	switch groupResource {
+
```
skriss: remove blank line

pkg/restore/restore.go
```diff
@@ -814,6 +823,35 @@ func hasControllerOwner(refs []metav1.OwnerReference) bool {
 	return false
 }
 
+// isComplete returns whether or not an object is considered completed
+// Used to identify whether or not an object should be restored. Only Jobs or Pods are considered
+func isCompleted(obj *unstructured.Unstructured, groupResource schema.GroupResource) (bool, error) {
+	podsGroupResource := schema.GroupResource{Group: "", Resource: "pods"}
```
skriss: we have some similar ones defined in `pkg/backup/item_backupper.go`, for different resources. WDYT about moving these into a shared pkg and exporting them?

pkg/restore/restore.go
```diff
@@ -814,6 +823,35 @@ func hasControllerOwner(refs []metav1.OwnerReference) bool {
...
+	case jobsGroupResource:
+		_, found, err := unstructured.NestedString(obj.UnstructuredContent(), "status", "completionTime")
+		if err != nil {
+			return false, errors.WithStack(err)
+		}
+		if found {
+			return true, nil
+		}
+	}
+	// Assume any other resource isn't complete and can be backed up
```
skriss: s/backed up/restored

pkg/restore/restore.go
```diff
@@ -814,6 +823,35 @@ func hasControllerOwner(refs []metav1.OwnerReference) bool {
...
+		if phase == string(v1.PodFailed) || phase == string(v1.PodSucceeded) {
+			return true, nil
+		}
+
+	case jobsGroupResource:
+		_, found, err := unstructured.NestedString(obj.UnstructuredContent(), "status", "completionTime")
+		if err != nil {
+			return false, errors.WithStack(err)
+		}
+		if found {
```
skriss: is it possible a job would have an empty `completionTime`? In which case the key might be found, but the job isn't actually complete (i'd assume)? Maybe just check for non-empty value here.

pkg/restore/restore.go
```diff
@@ -814,6 +823,35 @@ func hasControllerOwner(refs []metav1.OwnerReference) bool {
 	return false
 }
 
+// isComplete returns whether or not an object is considered completed
```
skriss: s/isComplete/isCompleted

pkg/restore/restore.go
nrb: Taking a look, I'm not seeing independent helper funcs in `item_backupper.go`. They're all methods.

I see 4 in `backup.go` - were those perhaps the ones you were referring to?

skriss: sorry, I meant the schema.GroupResource definitions

nrb: That makes more sense :) I considered doing this, but wanted to keep the change here relatively small. By updating all other uses of it, the change is touching several backup package files that aren't otherwise relevant to this change.

Where would you put them when they are moved? I was thinking of a new file in the `pkg/util/kube` package, but I'm not sure that's the best place.

skriss: Yeah, that's probably where I'd put them. Can just put a TODO in the code and circle back.

nrb: Comment added and the other changes have been addressed.

pkg/restore/restore.go
```diff
@@ -814,6 +823,36 @@ func hasControllerOwner(refs []metav1.OwnerReference) bool {
 	return false
 }
 
+// isCompleted returns whether or not an object is considered completed
+// Used to identify whether or not an object should be restored. Only Jobs or Pods are considered
+func isCompleted(obj *unstructured.Unstructured, groupResource schema.GroupResource) (bool, error) {
+	// TODO(0.9): These are copied from backup/item_backupper. Definititions should be moved
+	// to a shared package and imported here and in the backup package.
+	podsGroupResource := schema.GroupResource{Group: "", Resource: "pods"}
```
ncdc: I'd make these package-level `var`s (just above this function) so they aren't allocated over and over again

pkg/restore/restore.go
```diff
@@ -559,6 +559,16 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 			continue
 		}
 
+		complete, err := isCompleted(obj, groupResource)
+		if err != nil {
+			addToResult(&errs, namespace, fmt.Errorf("error checking completion %q: %v", fullPath, err))
+			continue
+		}
+		if complete {
+			ctx.infof("%s/%s is complete - skipping", obj.GetNamespace(), obj.GetName())
```
ncdc: We have a `NamespaceAndName()` helper that you can use here

pkg/restore/restore.go
```diff
@@ -782,8 +792,7 @@ func resetMetadataAndStatus(obj *unstructured.Unstructured) (*unstructured.Unstr
 		}
 	}
 
-	// this should never be backed up anyway, but remove it just
-	// in case.
+	// Clear status that may have been used to determine completed work
```
ncdc: I would just say "Never restore status"

pkg/restore/restore.go
```diff
@@ -814,6 +823,36 @@ func hasControllerOwner(refs []metav1.OwnerReference) bool {
 	return false
 }
 
+// isCompleted returns whether or not an object is considered completed
```
ncdc: Godoc doesn't distinguish based on line endings - you'll want to add a period at the end of this sentence.

pkg/restore/restore_test.go
```diff
@@ -676,6 +677,69 @@ func TestResetMetadataAndStatus(t *testing.T) {
...
+			expected:      true,
+			content:       `{"apiVersion":"v1","kind":"Pod","metadata":{"namespace":"ns","name":"pod1"}, "status": {"phase": "Succeeded"}}`,
+			groupResource: schema.GroupResource{Group: "", Resource: "pods"},
+		},
+		{
+			name:          "Pending pods aren't complete",
+			expected:      false,
+			content:       `{"apiVersion":"v1","kind":"Pod","metadata":{"namespace":"ns","name":"pod1"}, "status": {"phase": "Pending"}}`,
+			groupResource: schema.GroupResource{Group: "", Resource: "pods"},
+		},
```
ncdc: I'd add a "running pods" case too

pkg/restore/restore_test.go
```diff
@@ -676,6 +677,75 @@ func TestResetMetadataAndStatus(t *testing.T) {
...
+		{
+			name:          "Something not a pod or a job may actually be complete, but we're not concerned with that",
+			expected:      false,
+			content:       `{"apiVersion": "v1", "kind": "Namespace", "metadata": {"name": "ns"}, "status": {"completionTime": "bar", "phase":"Completed"}}`,
+			groupResource: schema.GroupResource{Group: "", Resource: "namespaces"},
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			u := unstructuredOrDie(fmt.Sprintf(test.content))
```
ncdc: No need to `fmt.Sprintf` here. Will fix in a follow-up

pkg/util/kube/groupresource.go
```diff
@@ -0,0 +1,25 @@
+/*
```
ncdc: This PR isn't using this file, which is ok for now. I think I'd like to move it to a standalone package that doesn't import anything other than `schema`.

ncdc:
LGTM with a couple of post-merge follow-ups

PR #465, opened 2018-04-26
Add a direct link to the slack channel

castrojo:
This is to make it trivial for a user to just click the link and go directly to the channel like we do for other Heptio OSS projects. 

Signed-off-by: Jorge O. Castro <jorge.castro@gmail.com>


ncdc:
Thanks @castrojo !

PR #470, opened 2018-05-02
backups: include clusterroles/bindings that reference serviceaccounts

### LINKED ISSUE #339: Gap: backing up/restoring with --include-namespaces doesn't include associated cluster-scoped resources such as clusterroles
If you do `ark backup create foo --include-namespaces myapp`, and your app requires a cluster-scoped resource such as `clusterroles`, `clusterrolebindings`, etc. to function, this backup won't include those cluster-scoped resources because they aren't in the `myapp` namespace. We should try to fix this (e.g. identify service accounts in the namespace, search for cluster-scoped rbac resources involving the service accounts, and so on).

skriss:

Fixes #339 

Add a backup item action for serviceaccounts that looks for clusterrolebindings that have the SA as a subject, and includes the clusterrolebinding & associated clusterrole in the list of additional items to backup.

Need to add units.


pkg/backup/service_account_action.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+// Execute checks for any ClusterRoleBindings that have this service account as a subject, and
+// adds the ClusterRoleBinding and associated ClusterRole to the list of additional items to
+// be backed up.
+func (a *serviceAccountAction) Execute(item runtime.Unstructured, backup *v1.Backup) (runtime.Unstructured, []ResourceIdentifier, error) {
+	a.log.Info("Running serviceAccountAction")
+	defer a.log.Info("Done running serviceAccountAction")
+
+	crbList, err := a.client.List(metav1.ListOptions{})
```
ncdc: I don't think we can use a real client here. Doing so would mean that we list all ClusterRoleBindings in the entire cluster every time we see a ServiceAccount.

We should either look into using a Lister from a SharedInformer, or caching the list per Backup UID (with something like a 15 minute expiration).

skriss: ah, good point. I'll look into one of those.

skriss: We could just fetch the ClusterRoleBindings in the action's constructor and store them for the life of the action since actions are created anew for each backup - WDYT?

ncdc: SGTM

ncdc:
LGTM, although we should probably add some logging indicating what's going on. We could either do it per plugin (when adding additional items), or globally in the item backupper.

skriss:
Added logging in the action (some of the other actions already do this, plus can include more context there on why the additional items are being included).

pkg/backup/service_account_action.go
```diff
@@ -0,0 +1,113 @@
...
+		namespace = objectMeta.GetNamespace()
+		name      = objectMeta.GetName()
+		bindings  = sets.NewString()
+		roles     = sets.NewString()
+	)
+
+	for _, roleBinding := range a.roleBindings {
+		for _, subj := range roleBinding.Subjects {
+			if subj.Kind == rbac.ServiceAccountKind && subj.Namespace == namespace && subj.Name == name {
+				a.log.Infof("Adding clusterrole %s and clusterrolebinding %s to additionalItems since serviceaccount %s/%s is a subject",
```
ncdc: I would move the logging down to the bindings and roles loops, separately. WDYT?

pkg/backup/service_account_action.go
```diff
@@ -0,0 +1,113 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	rbacclient "k8s.io/client-go/kubernetes/typed/rbac/v1"
+
+	"github.com/heptio/ark/pkg/apis/ark/v1"
+)
+
+// serviceAccountAction implements ItemAction.
+type serviceAccountAction struct {
+	log          logrus.FieldLogger
+	roleBindings []rbac.ClusterRoleBinding
```
ncdc: Sorry for the late comment, but could you call this `clusterRoleBindings`? Every time I see `roleBindings`, I think of the non-cluster scoped version.

skriss: done

pkg/backup/service_account_action.go
skriss: Oops, missed this comment. I had it there first then moved it :) I thought it was more informative to have the CR and CRB in the same log statement since they're linked.

ncdc: My thinking was that you might run into the same `clusterrrole` for multiple `clusterrolebindings` for a single `serviceaccount`, but this is fine as it is.

ncdc:
LGTM

PR #476, opened 2018-05-07
Dep cleanup

skriss:
Added some organization to our Gopkg.toml file, and removed a duplicate import of satori/uuid (see comment in Gopkg.toml).


ncdc:
LGTM

PR #478, opened 2018-05-07
when building on macOS, bind-mount volumes with delegated config

skriss:

Add `:delegated` configuration to volume bind-mounts when building, to improve performance when running on macOS. This option is ignored on non-macOS platforms.

Ref. https://docs.docker.com/storage/bind-mounts/#configure-mount-consistency-for-macos


ncdc:
Nice find!

PR #480, opened 2018-05-09
Azure: fix issues with setup instructions on macOS

### LINKED ISSUE #473: some of the Azure setup instructions don't work on macOS bash
As reported by a user, they work on linux but not OS X:

```
AZURE_STORAGE_ACCOUNT_ID="ark`cat /proc/sys/kernel/random/uuid | cut -d '-' -f5`"
```
and
```
--query [0].value
```

skriss:

Fixes #473

### DIFF
--- a/docs/azure-config.md
+++ b/docs/azure-config.md
@@ -28,10 +28,11 @@ The storage account can be created in the same Resource Group as your Kubernetes
 separated into its own Resource Group. The example below shows the storage account created in a
 separate `Ark_Backups` Resource Group.
 
-The storage account needs to be created with a globally unique id since this is used for dns. The
-random function ensures you don't have to come up with a unique name. The storage account is
-created with encryption at rest capabilities (Microsoft managed keys) and is configured to only
-allow access via https.
+The storage account needs to be created with a globally unique id since this is used for dns. In
+the sample script below, we're generating a random name using `uuidgen`, but you can come up with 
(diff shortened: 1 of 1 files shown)


docs/azure-config.md
```diff
@@ -39,7 +39,7 @@ AZURE_BACKUP_RESOURCE_GROUP=Ark_Backups
 az group create -n $AZURE_BACKUP_RESOURCE_GROUP --location WestUS
 
 # Create the storage account
-AZURE_STORAGE_ACCOUNT_ID="ark`cat /proc/sys/kernel/random/uuid | cut -d '-' -f5`"
+AZURE_STORAGE_ACCOUNT_ID="ark$(uuidgen | cut -d '-' -f5 | tr '[A-Z]' '[a-z]')"
```
ncdc: I checked various docker images to see if they have uuidgen:

- `fedora:27`: ✅ 
- `centos:7`: ✅
- `ubuntu:16.04`: ❌ but if you install `uuid-runtime` then ✅ 
- `alpine:3.7`: ❌ but if you install `util-linux` then ✅ 

We probably should say that we're just trying to come up with a globally unique identifier, 3-24 lowercase alphanumeric characters in length (https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions). If the `uuidgen` command isn't working for them, they can manually generate one too.

skriss: Ah, thanks for checking that. Pushed an update to the instructions.

PR #482, opened 2018-05-10
add replicasets and daemonsets to cohabitating resources

skriss:

ReplicaSets and DaemonSets also moved from extensions to apps. Add them to cohabitating resources so they don't get backed up twice.


ncdc:
LGTM

PR #483, opened 2018-05-10
shutdown gracefully on SIGINT/SIGTERM

### LINKED ISSUE #366: Clean up plugin processes on SIGINT when running locally
PR #334 made it possible to run the Ark server locally. However, when interrupting the server with ctrl-c, child processes such as the GCP object store plugin are not always cleaned up.

We should add a handler to kill any running plugin processes before exiting.

skriss:

Fixes #366 

Currently we don't handle these signals. Per https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods, SIGTERM is sent when terminating a pod, and SIGINT is sent if running the server locally (i.e. for development) and ctrl+c is pressed.

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -21,10 +21,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"os/signal"
 	"reflect"
 	"sort"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/heptio/ark/pkg/buildinfo"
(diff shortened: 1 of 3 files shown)


ncdc:
If we do something like this it terminates plugins on exit: https://gist.github.com/ncdc/c502eae499d63b0b1e445e8545d9b9f8

skriss:
Nice - wasn't thinking about that. Added.

ncdc:
This will fix #366

nrb:
This makes sense to me. I'll give it a try later today.

skriss:
@nrb did you get a chance to try this? Also @ncdc just a reminder of this given your plugin refactoring.

ncdc:
It'll be fine to merge. I'm actually working on top of this.

nrb:
@skriss Not yet - will fire it up against GKE now.

ncdc:
LGTM

PR #484, opened 2018-05-11
Consolidate commonly used GroupResource objects

nrb:
Cleans up some tech debt from #463


pkg/kuberesource/kuberesource.go
```diff
@@ -14,12 +14,16 @@ See the License for the specific language governing permissions and
...
+package kuberesource
 
 import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-var PodsGroupResource = schema.GroupResource{Group: "", Resource: "pods"}
-var JobsGroupResource = schema.GroupResource{Group: "batch", Resource: "jobs"}
-var NamespacesGroupResource = schema.GroupResource{Group: "", Resource: "namespaces"}
+var Pod = schema.GroupResource{Group: "", Resource: "pods"}
```
ncdc: I think these should all be plural. WDYT @skriss ?

nrb: I actually started it that way, then got skeptical of myself.

I think it makes sense to be plural and match the actual `Resource` field.

skriss: plural sounds good to me.

pkg/kuberesource/kuberesource.go
```diff
@@ -14,12 +14,16 @@ See the License for the specific language governing permissions and
...
+package kuberesource
 
 import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-var PodsGroupResource = schema.GroupResource{Group: "", Resource: "pods"}
-var JobsGroupResource = schema.GroupResource{Group: "batch", Resource: "jobs"}
-var NamespacesGroupResource = schema.GroupResource{Group: "", Resource: "namespaces"}
+var Pods = schema.GroupResource{Group: "", Resource: "pods"}
```
skriss: formatting nit, but I'd (a) sort by group and (b) declare like:
```
var (
    Pods = ...
    Namespaces = ...
)
```
which gofmt will auto-align nicely.

pkg/restore/restore.go
```diff
@@ -825,14 +826,12 @@ func hasControllerOwner(refs []metav1.OwnerReference) bool {
 
 // TODO(0.9): These are copied from backup/item_backupper. Definititions should be moved
```
skriss: Pls remove this comment :)

nrb: Nice catch, done.

skriss:
LGTM.

PR #485, opened 2018-05-11
use new cohabitatingResources map for each backup

skriss:

Fixes bug I introduced in #482 where moving cohabitatingResources to a package-level var caused state about "seen" to be tracked across backups.


ncdc:
How hard would it be to unit test that multiple backups each get their own copy?

skriss:
I think I found a way to do it without requiring any refactoring. Stay tuned..

ncdc:
LGTM

PR #490, opened 2018-05-14
Use JSON merge patches everywhere

skriss:

Unit tests pass but haven't tested manually yet.


ncdc:
-4k lines of code, I like it!

skriss:
Exercised all of the modified code paths and everything LGTM.

skriss:
@ncdc this is ready for review/merge when you get a chance.

ncdc:
If we want, we can keep the strategic merge patch for deployments (plugin add/remove).

skriss:
Could, although then we don't get to delete ~4k lines of code :( Any particular reason not to switch those too?

ncdc:
We'd need to look at what bits we're patching and see if the enhancements in strategic merge patch would be beneficial (see https://github.com/kubernetes/community/blob/master/contributors/devel/strategic-merge-patch.md).

skriss:
I double-checked this and it looks like the json-patch library correctly assembles a JSON merge patch to handle list changes. Tested by repeatedly adding plugins, and the list correctly grows rather than being replaced each time. Same for removing plugins.

ncdc:
Thanks for researching! LGTM.

PR #495, opened 2018-05-16
Refactor plugin management

### LINKED ISSUE #481: Ark server needs to handle crashed/stopped plugin processes
I've had a question that I don't know how to debug -- I noticed that Ark can suddenly fail to make its backup (after it made several successful backups) and the error logs looks like 

> 2018-05-10T00:04:13.972+00:00 10.1.69.73 dockerd[4085]: time="2018-05-10T00:04:13Z" level=error msg="Error uploading log file" bucket=\<some-bucket\> error="rpc error: code = Internal desc = connection error: desc = \"transport: error while dialing: dial unix /tmp/plugin763922082: connect: connection refused\"" key=\<some-key\>-logs.gz logSource="pkg/cloudprovider/backup_service.go:148".

This error was first noticed when I was using ark v0.6, and now I upgrade to ark v0.8.1 and the error still persists. 

Some symptoms that I noticed when the error happens

- Ark client can still list backups (`ark backup get` works), but it cannot show backups logs (`ark backup logs <backup>` fails)
- Kubectl can still show pod logs (`kubectl logs -f <pod>` works)
- If I don't take any actions, all future backups will fail and won't be uploaded to S3.
- If I bounce the pod (`kubectl delete pod`), everything works again and further backups can succeed maybe for a couple of days and then this error will happen again

My interpretation

- This error is caused by the failure to connect to S3
- It is not an authentication/authorization error, because it DOES work for a while after I bounce the pod

What I am asking here is

1.  Has anyone seen this problem, or this kind of error logs before?
2.  What might cause Ark to suddenly lose its ability to connect to S3?

Thanks for all the attentions.

ncdc:
Refactor plugin management:
- support multiple plugins per executable
- support restarting a plugin process in the event it terminates
- simplify plugin lifecycle management by using separate managers for
  each scope (server vs backup vs restore)

To-dos:
- [ ] Add max # of restarts and/or restart backoff

Fixes #481 

Things to verify:

- [ ] all plugins work as expected
- [ ] when a plugin process terminates, all Init()able plugins are reinitialized
- [ ] we can run a backup & a restore at the same time and the plugins work ok (no races)
- [ ] log messages from plugins are delivered to the server log appropriately
- [ ] per-backup/restore logs continue to work


ncdc:
One thing I want to do is some benchmarking of running plugins such as backup item actions. It seemed "slow" to me watching the backup in real time, but maybe that's just because I never paid attention before.

examples/gcp/00-ark-config.yaml
```diff
@@ -22,8 +22,8 @@ persistentVolumeProvider:
   name: gcp
 backupStorageProvider:
   name: gcp
-  bucket: <YOUR_BUCKET>
-backupSyncPeriod: 30m
+  bucket: andy-heptio-ark-data
```
skriss: revert changes to this file

pkg/cloudprovider/backup_service.go
```diff
@@ -134,53 +84,56 @@ func (br *backupService) seekAndPutObject(bucket, key string, file io.Reader) er
...
 	}
 
-	return br.objectStore.PutObject(bucket, key, file)
+	return objectStore.PutObject(bucket, key, file)
 }
 
-func (br *backupService) UploadBackup(bucket, backupName string, metadata, backup, log io.Reader) error {
-	// Uploading the log file is best-effort; if it fails, we log the error but it doesn't impact the
-	// backup's status.
+func UploadBackupLog(objectStore ObjectStore, bucket, backupName string, log io.Reader) error {
```
skriss: godoc

pkg/cloudprovider/backup_service.go
```diff
@@ -230,71 +185,76 @@ func (br *backupService) GetBackup(bucket, backupName string) (*api.Backup, erro
...
 
-func (br *backupService) UploadRestoreResults(bucket, backup, restore string, results io.Reader) error {
+// UploadRestoreResults uploads the restore's results file to object storage.
+func UploadRestoreResults(objectStore ObjectStore, bucket, backup, restore string, results io.Reader) error {
 	key := getRestoreResultsKey(backup, restore)
-	return br.objectStore.PutObject(bucket, key, results)
+	return objectStore.PutObject(bucket, key, results)
 }
 
 // cachedBackupService wraps a real backup service with a cache for getting cloud backups.
```
skriss: is this going to stick around?

pkg/cmd/server/server.go
```diff
@@ -243,9 +248,9 @@ func (s *server) run() error {
 
 	s.watchConfig(originalConfig)
 
-	if err := s.initBackupService(config); err != nil {
-		return err
-	}
+	// if err := s.initBackupService(config); err != nil {
```
skriss: delete?

pkg/cmd/server/server.go
```diff
@@ -381,16 +386,16 @@ func (s *server) handleShutdownSignals() {
 	}()
 }
 
-func (s *server) initBackupService(config *api.Config) error {
-	s.logger.Info("Configuring cloud provider for backup service")
-	objectStore, err := getObjectStore(config.BackupStorageProvider.CloudProviderConfig, s.pluginManager)
-	if err != nil {
-		return err
-	}
+// func (s *server) initBackupService(config *api.Config) error {
```
skriss: delete?

pkg/controller/backup_controller.go
```diff
@@ -377,6 +384,48 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
 	return kerrors.NewAggregate(errs)
 }
 
+func (controller *backupController) UploadBackup(logger logrus.FieldLogger, objectStore cloudprovider.ObjectStore, bucket, backupName string, metadata, backup, log io.Reader) error {
```
skriss: I wonder if it would make sense to extract this to `pkg/backup` so that it's part of the reusable lib?

pkg/restore/restore.go
```diff
@@ -64,9 +64,9 @@ type kindString string
 
 // kubernetesRestorer implements Restorer for restoring into a Kubernetes cluster.
 type kubernetesRestorer struct {
-	discoveryHelper    discovery.Helper
-	dynamicFactory     client.DynamicFactory
-	backupService      cloudprovider.BackupService
+	discoveryHelper discovery.Helper
+	dynamicFactory  client.DynamicFactory
+	// backupService      cloudprovider.BackupService
```
skriss: rm line

pkg/restore/restore.go
```diff
@@ -137,7 +137,7 @@ func prioritizeResources(helper discovery.Helper, priorities []string, includedR
 func NewKubernetesRestorer(
 	discoveryHelper discovery.Helper,
 	dynamicFactory client.DynamicFactory,
-	backupService cloudprovider.BackupService,
+	// backupService cloudprovider.BackupService,
```
skriss: rm line

pkg/plugin/backup_item_action.go
```diff
@@ -38,41 +39,43 @@ import (
...
 
-// GRPCServer registers a BackupItemAction gRPC server.
-func (p *BackupItemActionPlugin) GRPCServer(s *grpc.Server) error {
-	proto.RegisterBackupItemActionServer(s, &BackupItemActionGRPCServer{impl: p.impl})
-	return nil
-}
-
-// GRPCClient returns a BackupItemAction gRPC client.
+// GRPCClient returns a clientMux for BackupItemAction gRPC clients.
 func (p *BackupItemActionPlugin) GRPCClient(c *grpc.ClientConn) (interface{}, error) {
```
skriss: The naming that we're locked into here (`GRPCClient(...)`) due to go-plugin is unfortunate, since what we're returning is not actually a gRPC client but something that makes clients. Might be helpful to expand the godoc to explain that this is called by go-plugin, and can return any type but it should be a client or something that can be used to create a client

pkg/plugin/manager.go
```diff
@@ -17,405 +17,189 @@ limitations under the License.
...
 
-	return plugin, nil
+	m.lock.Unlock()
 }
 
-func (m *manager) registerPlugins() error {
-	arkCommand := os.Args[0]
+// getWrapper returns a wrapper for a plugin identified by kind and name, creating a wrapper if it is the first time it
+// has been requested.
+func (m *manager) getWrapper(kind PluginKind, name string) (*restartablePluginProcess, error) {
```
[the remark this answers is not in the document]
ncdc: This missed getting renamed when I changed wrapper to restartablePluginProcess. It should be called getRestartablePluginProcess.

pkg/plugin/restartable_plugin_process.go
```diff
@@ -0,0 +1,177 @@
...
+// instance for key.name.
+func (p *restartablePluginProcess) dispenseLH(key kindAndName) (interface{}, error) {
+	// This calls GRPCClient(clientConn) on the plugin instance registered for key.name.
+	dispensed, err := p.protocolClient.Dispense(key.kind.String())
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	// Currently all plugins except for PluginLister dispense clientMux instances.
+	if mux, ok := dispensed.(*clientMux); ok {
```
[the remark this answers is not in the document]
ncdc: Is it the "restartable" part of the name that's tripping you up? Other than the registry that calls `Dispense()` to list plugins, this is the only other place in the code where we call `Dispense()`. We could call this something other than `restartablePluginProcess`.

Or if it's less the name and more that restarting & dispensing seem like 2 different things, maybe we could find something. We need the `protocolClient` to be able to dispense, so we could introduce some other struct that solely focuses on dispensing using that. And then have the `restartablePluginProcess` hold on to those.

> wonder if there's a generic interface that all GRPCClient funcs could return that enables callers to get an actual client/proxy from the result

Different from the type assertion to `*clientMux`?

pkg/plugin/server_mux.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// ServerInitializer is a function that initializes and returns a new instance of one of Ark's plugin interfaces
+// (ObjectStore, BlockStore, BackupItemAction, RestoreItemAction).
+type ServerInitializer func() (interface{}, error)
+
+// serverMux manages multiple implementations of a single plugin kind, such as pod and pvc BackupItemActions.
+type serverMux struct {
```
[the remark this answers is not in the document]
ncdc: SGTM

pkg/plugin/restartable_plugin_process.go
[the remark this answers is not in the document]
ncdc: I have some local changes where I extracted the dispensing into a `pluginDispenser` struct. It's a small change but it separates that logic, at least.

pkg/plugin/client_builder.go
```diff
@@ -1,43 +1,60 @@
...
 func (b *clientBuilder) withLogger(logger hclog.Logger) *clientBuilder {
-	b.config.Logger = logger
+	b.logger = logger
 
 	return b
 }
 
-func (b *clientBuilder) withCommand(name string, args ...string) *clientBuilder {
-	b.config.Cmd = exec.Command(name, args...)
+// withCommand sets the clientBuilder's commandName to name and commandArgs to args.
```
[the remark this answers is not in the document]
ncdc: We don't need it as an arg. Plugin authors must expect their plugin executable file to be invoked directly, with no arguments. We do still need it in the struct, as it's used for the special `ark run-plugin` case.

pkg/plugin/restartable_plugin_process.go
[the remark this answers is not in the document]
ncdc: Just pushed up the changes. I quickly replaced `pluginDispenser` with `process`, responsible for managing a single execution of a plugin process.

skriss:
The updates LGTM, I like the additional types and revised names!

pkg/controller/backup_deletion_controller.go
```diff
@@ -46,16 +46,36 @@ type backupDeletionController struct {
...
 	podvolumeBackupLister     listers.PodVolumeBackupLister
 
-	processRequestFunc func(*v1.DeleteBackupRequest) error
-	clock              clock.Clock
+	backupStorageDeleter backupStorageDeleter
+	processRequestFunc   func(*v1.DeleteBackupRequest) error
+	clock                clock.Clock
+}
+
+type backupStorageDeleter interface {
```
skriss: the need to do this kind of thing is an annoying downside to having package-level funcs vs. methods on a struct :/

pkg/controller/restore_controller.go
```diff
@@ -211,7 +278,9 @@ func (controller *restoreController) processRestore(key string) error {
 	logContext.Debug("Running processRestore")
 	ns, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
-		return errors.Wrap(err, "error splitting queue key")
+		logContext.WithError(err).Error("unable to process restore: error splitting queue key")
```
skriss: I think we have this returning an error in most/all of our existing controllers, we probably want to add an issue to go change that across the board

pkg/plugin/backup_item_action.go
```diff
@@ -38,41 +38,43 @@ import (
...
-}
+//////////////////////////////////////////////////////////////////////////////
+// client code
+//////////////////////////////////////////////////////////////////////////////
 
-// GRPCClient returns a BackupItemAction gRPC client.
+// GRPCClient returns a clientMux for BackupItemAction gRPC clients.
 func (p *BackupItemActionPlugin) GRPCClient(c *grpc.ClientConn) (interface{}, error) {
-	return &BackupItemActionGRPCClient{grpcClient: proto.NewBackupItemActionClient(c), log: p.log}, nil
+	return newClientMux(c, newBackupItemActionGRPCClient), nil
```
skriss: what did you think about renaming `clientMux` -> `clientDispenser` (and this func as well)?

pkg/plugin/client_mux.go
```diff
@@ -0,0 +1,71 @@
...
+	c.plugin = name
+}
+
+func (c *clientBase) setLog(log *logrusAdapter) {
+	c.log = log
+}
+
+// clientMux supports the initialization and retrieval of multiple implementations for a single plugin kind, such as
+// "aws" and "azure" implementations of the object store plugin.
+type clientMux struct {
```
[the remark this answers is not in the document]
skriss: Per previous - what about `clientDispenser`?

pkg/plugin/server_mux.go
```diff
@@ -0,0 +1,65 @@
...
+
+// ServerInitializer is a function that initializes and returns a new instance of one of Ark's plugin interfaces
+// (ObjectStore, BlockStore, BackupItemAction, RestoreItemAction).
+type ServerInitializer func(logger logrus.FieldLogger) (interface{}, error)
+
+// serverMux manages multiple implementations of a single plugin kind, such as pod and pvc BackupItemActions.
+type serverMux struct {
+	kind         PluginKind
+	initializers map[string]ServerInitializer
+	instances    map[string]interface{}
```
skriss: WDYT about using `handler` in place of `instance`?

Makefile
```diff
@@ -24,7 +24,7 @@ PKG := github.com/heptio/ark
 REGISTRY ?= gcr.io/heptio-images
 
 # Which architecture to build - see $(ALL_ARCH) for options.
-ARCH ?= linux-amd64
+ARCH ?= $(shell go env GOOS)-$(shell go env GOARCH)
```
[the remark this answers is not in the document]
ncdc: I realized that this will hurt `make container` if you're running on macOS. I'll see what I can do. We don't have to include the Makefile changes in this PR if it's too much of a pain.

pkg/cloudprovider/backup_service.go
```diff
@@ -26,49 +25,23 @@ import (
...
-	UploadRestoreResults(bucket, backup, restore string, results io.Reader) error
-}
-
 // BackupGetter knows how to list backups in object storage.
 type BackupGetter interface {
 	// GetAllBackups lists all the api.Backups in object storage for the given bucket.
 	GetAllBackups(bucket string) ([]*api.Backup, error)
 }
 
+type XXXBackupGetter interface {
```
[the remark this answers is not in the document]
ncdc: You read my mind :-) haven't done the rename yet.

pkg/controller/backup_deletion_controller.go
ncdc: I'll see if I can find something simpler

pkg/controller/restore_controller.go
ncdc: +1

pkg/plugin/backup_item_action.go
ncdc: Yes, we talked about that, but I overlooked it. I'll make the change.

pkg/plugin/server_mux.go
ncdc: +1

pkg/plugin/restartable_process.go
```diff
@@ -80,8 +81,13 @@ func (p *restartableProcess) reset() error {
 //
 // Callers of resetLH *must* acquire the lock before calling it.
 func (p *restartableProcess) resetLH() error {
+	if p.resetFailures > 10 {
```
skriss: are you thinking about including any sort of backoff for this?

skriss:
haven't gone through the tests themselves yet, but went through the rest of the rename & refactor stuff and LGTM.

pkg/plugin/restartable_process.go
ncdc: I haven't fully fleshed this out yet, but we can/should discuss. Ideally we can handle the following scenarios:
1. process dies for whatever reason (e.g. oom kill), and restarting succeeds the first try
1. process dies for whatever reason, attempt(s) to restart fail early and quickly -> we surface the issue somehow (ark server shuts itself down, or we create events, or ?)

I would like to try to avoid the situation where the process dies, and it takes minutes or more to finish all the attempts to restart it.

ncdc:
Replaced by #710

PR #496, opened 2018-05-16
Check restore Phase before downloading logs

### LINKED ISSUE #477: Check restore phase when getting logs
When using `ark restore logs my-restore` before a restore is completed, the user will see a not found error from the object storage. This happens because the log file isn't uploaded until the restore is complete.

Before checking object storage, the `restore logs` subcommand should check the `Phase` of a restore and only get the file if `Phase` is `Completed`.

nrb:
Fixes #477 


nrb:
Is this a reasonable fix? I added the logic to the `Validate` function based on my interpretation of the godoc.

pkg/cmd/cli/restore/logs.go
```diff
@@ -82,6 +83,12 @@ func (l *LogsOptions) Validate(f client.Factory) error {
 	}
 	l.client = c
 
-	_, err = l.client.ArkV1().Restores(f.Namespace()).Get(l.RestoreName, metav1.GetOptions{})
+	r, err := l.client.ArkV1().Restores(f.Namespace()).Get(l.RestoreName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if r.Status.Phase != v1.RestorePhaseCompleted {
+		err = errors.Errorf("Restore %s is not yet complete.")
```
ncdc: I would say "unable to retrieve logs because restore is not yet complete." I don't think we need to include the name of the restore.

ncdc:
@nrb yes, this is reasonable, thanks.

pkg/cmd/cli/restore/logs.go
```diff
@@ -82,6 +83,12 @@ func (l *LogsOptions) Validate(f client.Factory) error {
 	}
 	l.client = c
 
-	_, err = l.client.ArkV1().Restores(f.Namespace()).Get(l.RestoreName, metav1.GetOptions{})
+	r, err := l.client.ArkV1().Restores(f.Namespace()).Get(l.RestoreName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if r.Status.Phase != v1.RestorePhaseCompleted {
+		err = errors.Errorf("unable to retrieve logs because restore is not complete")
```
skriss: `errors.New(...)` since it's not a format string, and I'd probably just return the error here and change L93 to a return nil.

ncdc: +1 to returning here

I'm almost always in favor of Errorf 100% of the time even if there aren't format strings - saves you the time of having to replace New with Errorf if you do decide you need formatted vars

Issue #404, opened 2018-03-28
Backup everything except for a few resources


rhefner1:
I want to do an entire cluster backup but I want to exclude a few persistent volumes from being snapshotted (since $0.10/GB/mo can add up fast). I know that there is the `--selector` flag but is it possible to do the opposite - only exclude items that match the selector?

If this isn't possible, does it seem like a simple code change or would it be more involved?

skriss:
This isn't something we support currently, but the use case makes sense and it shouldn't be too complicated to implement. Thanks for the feedback!

dhananjaysathe:
+1 for this feature , i may purposely want to leave out certain parts
At this point a workaround is to use the negation of a selector , building on the example on the default docs it would look something like this 
`ark backup create not-nginx-backup --selector 'app notin (nginx)'`
 cc @rhefner1 
I tried it out and can confirm it works. You could simple add a lable to all those disks and use a negation

rosskukulinski:
Can you combine a negative selector with a positive selector in the same string?

If you can, then I don't see a need for a new CLI flag now that docs were updated in #501 and #518. If not, then yes, we should add a new flag.

We'll need to make sure the interaction between --selector and --not-selector work in an expected way with #445.

jbeda:
Sorry for not responding here earlier.  Happy to hop on a call at some point and brainstorm ideas.

I think that it might be worthwhile to have an ordered list of things that are included or not.  This mirrors some of the config/options for file system systems -- I'd look at rsync or find for ideas.

ncdc:
> an ordered list of things that are included or not

@jbeda could you expand a bit - do you have any examples?

@rosskukulinski this looks similar to #317

ncdc:
It sounds like the `notin` selector condition satisfies the request. If not, please let us know and we can revisit.

PR #501, opened 2018-05-17
Add documentation on how to use a selector negation

dhananjaysathe:
This will help users use the `--selector` flag to selectively exclude objects from being backed up by ark

workaround for #404 until dedicated flags are implemented

Signed-off-by: Dhananjay Sathe <dhanajaysathe@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,12 @@ Make sure that you install somewhere in your `$PATH`.
     ark backup create nginx-backup --selector app=nginx
     ```
 
+   Alternatively if you want to backup all objects *except* those matching the label `backup=ignore` :
+
+   ```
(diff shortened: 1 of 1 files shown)


dhananjaysathe:
@ncdc reopened it , had my ide and env messing around with the previous one

skriss:
@dhananjaysathe thanks for the PR! We'll see what @ncdc thinks, but I think this might be best to put in the command text itself -- perhaps in the Examples (e.g. https://github.com/heptio/ark/blob/master/pkg/cmd/cli/restore/create.go#L43)

ncdc:
cc @Bradamant3

PR #506, opened 2018-05-18
Add functional options for the Ark config install library

marpaia:
In #437, we decided we wanted to use functional options to manage Ark config details (instead of defaulting to static times). This PR adds such a capability.

### DIFF
--- a/pkg/install/config.go
+++ b/pkg/install/config.go
@@ -8,6 +8,32 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type arkConfigOption func(*arkConfig)
+
+type arkConfig struct {
+	backupSyncPeriod time.Duration
+	gcSyncPeriod     time.Duration
+	restoreOnly      bool
+}
+
+func WithBackupSyncPeriod(t time.Duration) arkConfigOption {
+	return func(c *arkConfig) {
+		c.backupSyncPeriod = t
+	}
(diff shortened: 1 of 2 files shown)


pkg/install/config.go
```diff
@@ -8,14 +8,58 @@ import (
...
+	}
+}
+
+func WithGCSyncPeriod(t time.Duration) arkConfigOption {
+	return func(c *arkConfig) {
+		c.gcSyncPeriod = t
+	}
+}
+
+func WithScheduleSyncPeriod(t time.Duration) arkConfigOption {
```
ncdc: We're planning on removing this configuration option and instead hardcoding it to 1 minute, so we may want to omit this.

marpaia: OK, i'll remove the option for now and leave the resource specification hardcoded to 1 minute.

pkg/install/config.go
```diff
@@ -8,14 +8,51 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type arkConfigOption func(*arkConfig)
+
+type arkConfig struct {
+	backupSyncPeriod   time.Duration
+	gcSyncPeriod       time.Duration
+	scheduleSyncPeriod time.Duration
```
ncdc: You can remove this too

ncdc:
LGTM

PR #508, opened 2018-05-18
[WIP] restic integration with Ark

skriss:
Reviewers: I wouldn't bother looking through individual commits, just the overall diff. I'll probably squash them all.

TODO:
- [ ] update Makefile / associated build artifacts
- [ ] finalize maintenance schedule
- [ ] run backups/restores of multiple volumes within a pod concurrently


cmd/restic-init-container/main.go
```diff
@@ -0,0 +1,61 @@
...
+
+		if _, err := os.Stat(doneFile); os.IsNotExist(err) {
+			fmt.Printf("Not found: %s\n", doneFile)
+			return false
+		} else if err != nil {
+			fmt.Fprintf(os.Stderr, "ERROR looking for %s: %s\n", doneFile, err)
+			return false
+		}
+
+		fmt.Printf("Found %s", doneFile)
```
ncdc: Do you need to `break` here?

cmd/restic-wrapper/main.go
```diff
@@ -0,0 +1,97 @@
...
+
+	passwordFile, err := createPasswordFile(secret, repo)
+	checkError(err, "ERROR creating password temp file")
+	defer os.Remove(passwordFile)
+
+	// set the env vars that restic uses for azure credentials
+	os.Setenv("AZURE_ACCOUNT_NAME", os.Getenv("AZURE_STORAGE_ACCOUNT_ID"))
+	os.Setenv("AZURE_ACCOUNT_KEY", os.Getenv("AZURE_STORAGE_KEY"))
+
+	resticArgs := append([]string{"-p=" + passwordFile}, os.Args[1:]...)
```
ncdc: If there are longhand forms for the flags, it might be clearer to use those.

cmd/restic-wrapper/main.go
```diff
@@ -0,0 +1,97 @@
...
+	resticCmd := exec.Command("/restic", resticArgs...)
+	resticCmd.Stdout = os.Stdout
+	resticCmd.Stderr = os.Stderr
+
+	checkError(resticCmd.Run(), "ERROR running restic command")
+}
+
+func repoName(args []string) string {
```
ncdc: doc/example pls

cmd/restic-wrapper/main.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return client.Secrets(namespace), nil
+}
+
+// checkError prints an error message and the error's text to stderr and
+// exits with a status code of 1 if the error is not nil, or is a no-op
+// otherwise.
+func checkError(err error, msg string) {
+	if err != nil {
+		fmt.Fprintf(os.Stderr, msg+": %s\n", err)
```
ncdc: We usually `%v` errors

cmd/restic-wrapper/main.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, err
+	}
+
+	return client.Secrets(namespace), nil
+}
+
+// checkError prints an error message and the error's text to stderr and
+// exits with a status code of 1 if the error is not nil, or is a no-op
+// otherwise.
+func checkError(err error, msg string) {
```
ncdc: Since you're printing the msg and then the error, I think it would make sense to reverse the order of the arguments.

cmd/restic-wrapper/main.go
```diff
@@ -0,0 +1,97 @@
...
+		os.Exit(1)
+	}
+}
+
+func createPasswordFile(secret *corev1api.Secret, repo string) (string, error) {
+	file, err := ioutil.TempFile("", fmt.Sprintf("restic-credentials-%s", repo))
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
```
ncdc: As we learned the hard way today, need to handle any error this might return. See https://mijailovic.net/2017/05/09/error-handling-patterns-in-go/ and https://joeshaw.org/dont-defer-close-on-writable-files/ for some ideas.

examples/aws/20-restic-daemonset.yaml
```diff
@@ -0,0 +1,54 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata: 
+  name: restic-daemon
```
ncdc: Please don't include the kind/resource in the name (remove `-daemon`)

pkg/backup/item_backupper.go
```diff
@@ -228,6 +235,25 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
...
+		pod := new(corev1api.Pod)
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), pod); err != nil {
+			return errors.WithStack(err)
+		}
+
+		if err := ib.resticBackupper.BackupPodVolumes(ib.backup, pod, log); err != nil {
+			return err
+		}
+
+		// pod was permuted by BackupPodVolumes if there were any volumes backed up, so
```
ncdc: permuted -> mutated?

pkg/restic/backupper_restorer.go
```diff
@@ -0,0 +1,286 @@
...
+
+	// for each volume to backup:
+	for _, volumeName := range volumesToBackup {
+		wg.Add(1)
+		go func(volumeName string) {
+			defer wg.Done()
+
+			volume := getVolume(pod, volumeName)
+			if volume == nil {
+				lock.Lock()
```
ncdc: For coordinating goroutines & their returned errors, there's at least `golang.org/x/sync/errgroup` and @davecheney may have a better suggestion.

pkg/restic/backupper_restorer.go
```diff
@@ -0,0 +1,286 @@
...
+	var (
+		errs []error
+		lock sync.Mutex
+		wg   sync.WaitGroup
+	)
+
+	// for each volume to backup:
+	for _, volumeName := range volumesToBackup {
+		wg.Add(1)
+		go func(volumeName string) {
```
ncdc: Could this be declared standalone instead of as an anonymous function?

pkg/restic/backupper_restorer.go
```diff
@@ -0,0 +1,286 @@
...
+				"pod":        pod.Name,
+				"pod-uid":    string(pod.UID),
+				"ns":         pod.Namespace,
+				"volume":     volumeName,
+			}
+
+			cmd := backupCommand(br.metadataManager.RepoPrefix(), pod.Namespace, string(pod.UID), volumeDir, snapshotTags)
+
+			// exec it in the daemonset pod
+			if err := br.daemonSetExecutor.Exec(pod.Spec.NodeName, cmd, 10*time.Minute, log); err != nil {
```
ncdc: 10 minutes?

pkg/restic/backupper_restorer.go
```diff
@@ -0,0 +1,286 @@
...
+				lock.Lock()
+				errs = append(errs, err)
+				lock.Unlock()
+				return
+			}
+
+			// save the snapshot's ID in the pod annotation
+			lock.Lock()
+			SetPodSnapshotAnnotation(pod, volumeName, snapshotID)
+			backupSnapshots = append(backupSnapshots, fmt.Sprintf("%s/%s", pod.Namespace, snapshotID))
```
ncdc: Instead of using a lock, could we somehow use channel(s) to coordinate retrieving the snapshot id?

pkg/restic/daemon_set_executor.go
```diff
@@ -0,0 +1,93 @@
...
+		log,
+		unstructuredPod,
+		dsPod.Namespace,
+		dsPod.Name,
+		"restic-backup",
+		dsCmd)
+}
+
+func (dse *defaultDaemonSetExecutor) getDaemonSetPod(node string) (*corev1api.Pod, error) {
+	// TODO we may want to cache this list
```
ncdc: We can start the k8s shared informer factory beforehand and use the pod informer

pkg/restic/repository_manager.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	var repos []string
+	for _, prefix := range prefixes {
+		if len(prefix) <= 1 {
+			continue
+		}
+
+		// strip the trailing '/' if it exists
+		if prefix[len(prefix)-1] == '/' {
```
ncdc: `strings.TrimSuffix`

cmd/restic-init-container/main.go
skriss: no, since we need to make it through this loop for *every* child that's a directory. the `Printf` is for information only.

cmd/restic-wrapper/main.go
skriss: good idea.

cmd/restic-wrapper/main.go
```diff
@@ -0,0 +1,97 @@
...
+	"os/exec"
+	"strings"
+
+	corev1api "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+)
+
+func main() {
```
skriss: @ncdc I was considering trying to get rid of this binary. Since it's used in the daemonset, that would require the Ark server to create a temp creds file in the appropriate daemonset pod. Do you think that's a good idea?

pkg/restic/backupper_restorer.go
skriss: we probably need to make this tuneable.

pkg/restic/backupper_restorer.go
ncdc: Something like this would test your change: https://gist.github.com/ncdc/12c3b6b6f4db0e5c60f4ecf77e8d85c2

skriss:
@ncdc any issue with me squashing most/all of the commits down to 1?

pkg/restic/command.go
```diff
@@ -0,0 +1,73 @@
...
+	return strings.Join(c.StringSlice(), " ")
+}
+
+func (c *command) Command() *exec.Cmd {
+	parts := c.StringSlice()
+	return exec.Command(parts[0], parts[1:]...)
+}
+
+func repoFlag(prefix, repo string) string {
+	return fmt.Sprintf("-r=%s/%s", prefix, repo)
```
ncdc: full flag name

pkg/restic/command.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func passwordFlag(file string) string {
+	return fmt.Sprintf("-p=%s", file)
```
ncdc: full flag name

pkg/restic/command.go
```diff
@@ -0,0 +1,73 @@
...
+func backupTagFlags(vals map[string]string) []string {
+	var flags []string
+	for k, v := range vals {
+		flags = append(flags, fmt.Sprintf("--tag=%s=%s", k, v))
+	}
+	return flags
+}
+
+func restoreTargetFlag(podUID string) string {
+	return fmt.Sprintf("-t=/restores/%s", podUID)
```
ncdc: full flag name

pkg/restic/common.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil
+	}
+
+	backupsValue := annotations[volumesToBackupAnnotation]
+	if backupsValue == "" {
+		return nil
+	}
+
+	var backups []string
+	// check for json array
```
ncdc: Would it make more sense to do a comma-separated string instead of a json-encoded array?

pkg/restic/repository_manager.go
```diff
@@ -0,0 +1,404 @@
...
+		command:    "init",
+		repoPrefix: rm.repoPrefix,
+		repo:       name,
+	}
+
+	return errorOnly(rm.exec(cmd))
+}
+
+func (rm *repositoryManager) getAllRepos() ([]string, error) {
+	prefixes, err := rm.objectStore.ListCommonPrefixes(rm.bucket, "/")
```
ncdc: If the resticLocation that came from the config was bucket/dir, that goes into rm.bucket - will that work here having both the bucket and the directory in a single string? Or do we need to call ListCommonPrefixes(bucketName, dir+"/")?

pkg/backup/item_backupper.go
```diff
@@ -228,6 +235,25 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 		}
 	}
 
+	if groupResource == kuberesource.Pods && len(restic.GetVolumesToBackup(metadata)) > 0 {
+		pod := new(corev1api.Pod)
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), pod); err != nil {
+			return errors.WithStack(err)
+		}
+
+		if err := ib.resticBackupper.BackupPodVolumes(ib.backup, pod, log); err != nil {
```
ncdc: Note, this invocation ends up listing all the common prefixes (i.e. repos) in resticLocation in object storage (ensuring the restic repo exists for the namespace) for each pod. Are you concerned about how frequently we'll be running that same query for a given namespace? I know the cost is fairly minimal, but there is still a cost associated with listing from object storage. We may want something more efficient and/or a cache of the data.

pkg/restic/repository_manager.go
```diff
@@ -0,0 +1,404 @@
...
+		if repo == name {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func (rm *repositoryManager) InitRepo(name string) error {
+	resticCreds, err := rm.secretsClient.Get(credsSecret, metav1.GetOptions{})
```
ncdc: The plan was to require an admin to precreate manually per namespace, wasn't it?

ncdc: We should discuss the ui/ux. We could have something like
```
# generates an encryption key, creates a secret for it, maybe creates a new ResticRepo, ark server sees it's New and runs restic init?
ark restic init --namespace foo

# same as above, but uses a precreated encryption key instead of generating
ark restic init --namespace foo --encryption-key /path/to/file
```

ncdc: If we were to add a ResticRepo CRD, we could use that to manage its status (needs init, already done, etc). Not sure if we need this or how it would impact backup replication. Would probably have to be tied to a BackupTarget when we get there.

pkg/restic/repository_manager.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	// write it to a temp file
+	file, err := ioutil.TempFile("", fmt.Sprintf("restic-credentials-%s", cmd.repo))
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	defer func() {
+		file.Close()
```
ncdc: Need better error handling around close/remove errors.

pkg/restore/restic_restore_action.go
```diff
@@ -0,0 +1,111 @@
...
+
+	for volumeName := range volumeSnapshots {
+		mount := corev1.VolumeMount{
+			Name:      volumeName,
+			MountPath: "/restores/" + volumeName,
+		}
+		initContainer.VolumeMounts = append(initContainer.VolumeMounts, mount)
+	}
+
+	if len(pod.Spec.InitContainers) == 0 || pod.Spec.InitContainers[0].Name != "restic-wait" {
```
ncdc: If there are no init containers, list=initContainer

If the first init container isn't restic-wait, set list=initContainer, rest...

If the first init container is restic-wait, overwrite it with initContainer

?

skriss:
@ncdc switched restores to using a CRD, and did some cleanup with the backup CRD/controller.

Still need to change backupper_restorer to not be global (and should be easy to split in half now).
Still need to support manual repo key creation.
Some lower-level cleanup to do as well.

pkg/restic/repository_manager.go
```diff
@@ -0,0 +1,574 @@
...
+	}
+
+	// ensure a repo exists for the pod's namespace
+	if err := rm.ensureRepo(pod.Namespace); err != nil {
+		return err
+	}
+
+	resultChan := make(chan *arkv1api.PodVolumeBackup)
+
+	informer := arkv1informers.NewFilteredPodVolumeBackupInformer(
```
ncdc: We will probably want to have 1 informer running per backup, and this does 1 informer per backup+pod. We could add a new struct that would be per-backup, and create a new one at the beginning of a backup.

pkg/restic/repository_manager.go
```diff
@@ -0,0 +1,574 @@
...
+ForLoop:
+	for i := 0; i < len(volumesToBackup); i++ {
+		select {
+		case <-ctx.Done():
+			errs = append(errs, errors.New("timed out waiting for all PodVolumeBackups to complete"))
+			break ForLoop
+		case res := <-resultChan:
+			switch res.Status.Phase {
+			case arkv1api.PodVolumeBackupPhaseCompleted:
+				SetPodSnapshotAnnotation(pod, res.Spec.Volume, res.Status.SnapshotID)
```
ncdc: I still think I'd like to see this function return the pod/snapshot associations, and have another function modify that pod. It hopefully will make testing easier.

pkg/restic/repository_manager.go
```diff
@@ -0,0 +1,574 @@
...
+					"pod":        pod.Name,
+					"pod-uid":    string(pod.UID),
+					"ns":         pod.Namespace,
+					"volume":     volumeName,
+				},
+				RepoPrefix: rm.repoPrefix,
+			},
+		}
+
+		// TODO should we return here, or continue with what we can?
```
ncdc: record the error & continue

pkg/restic/repository_manager.go
```diff
@@ -0,0 +1,574 @@
...
+		func(opts *metav1.ListOptions) {
+			opts.LabelSelector = fmt.Sprintf("%s=%s", arkv1api.BackupNameLabel, backup.Name)
+		},
+	)
+	informer.AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(_, obj interface{}) {
+				pvb := obj.(*arkv1api.PodVolumeBackup)
+
+				if pvb.Status.Phase == arkv1api.PodVolumeBackupPhaseCompleted || pvb.Status.Phase == arkv1api.PodVolumeBackupPhaseFailed {
```
ncdc: Can this function be run in parallel for 2 pods for the same backup? We'll likely need to match PVBs for this backup's name/UID.

pkg/restic/repository_manager.go
```diff
@@ -0,0 +1,574 @@
...
+		func(opts *metav1.ListOptions) {
+			opts.LabelSelector = fmt.Sprintf("%s=%s", arkv1api.RestoreNameLabel, restore.Name)
+		},
+	)
+	informer.AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(_, obj interface{}) {
+				pvr := obj.(*arkv1api.PodVolumeRestore)
+
+				if pvr.Status.Phase == arkv1api.PodVolumeRestorePhaseCompleted || pvr.Status.Phase == arkv1api.PodVolumeRestorePhaseFailed {
```
ncdc: Can this function be run in parallel for 2 pods for the same restore? We'll likely need to match PVRs for this restore's name/UID.

pkg/restic/repository_manager.go
```diff
@@ -0,0 +1,574 @@
...
+func (rm *repositoryManager) RestorePodVolumes(ctx context.Context, restore *arkv1api.Restore, pod *corev1api.Pod, log logrus.FieldLogger) error {
+	// get volumes to restore from pod's annotations
+	volumesToRestore := GetPodSnapshotAnnotations(pod)
+	if len(volumesToRestore) == 0 {
+		return nil
+	}
+
+	resultChan := make(chan *arkv1api.PodVolumeRestore)
+
+	informer := arkv1informers.NewFilteredPodVolumeRestoreInformer(
```
ncdc: Same comment as above re 1 informer per restore instead of 1 per restore+pod.

restic-container/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Copyright 2018 the Heptio Ark contributors.
```
ncdc: This file is no longer used?

ncdc:
Also need to remove the `restic-init-container/restic-init-container` binary executable from the branch

pkg/backup/item_backupper.go
```diff
@@ -97,19 +104,19 @@ type ItemBackupper interface {
...
+	namespaces              *collections.IncludesExcludes
+	resources               *collections.IncludesExcludes
+	backedUpItems           map[itemKey]struct{}
+	actions                 []resolvedAction
+	tarWriter               tarWriter
+	resourceHooks           []resourceHook
+	dynamicFactory          client.DynamicFactory
+	discoveryHelper         discovery.Helper
+	snapshotService         cloudprovider.SnapshotService
 	itemHookHandler         itemHookHandler
```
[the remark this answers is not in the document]
skriss: Do you mean fields directly passed into the constructor? bc in this example, `itemHookHandler` and `additionalItemBackupper` are both set in the constructor, just derived rather than directly passed in.

ncdc:
Yes, fields passed in vs derived.

pkg/restic/backupper.go
```diff
@@ -0,0 +1,147 @@
...
+	b.results[resultsKey(pod.Namespace, pod.Name)] = make(chan *arkv1api.PodVolumeBackup)
+	b.resultsLock.Unlock()
+
+	var (
+		errs            []error
+		volumeSnapshots = make(map[string]string)
+	)
+
+	for _, volumeName := range volumesToBackup {
+		b.repoManager.repoLocker.Lock(pod.Namespace, false)
```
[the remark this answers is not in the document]
skriss: how come?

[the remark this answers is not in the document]
ncdc: If we lock here, and we're backing up a namespace that has 20, 50, 100+ pods, we are repeatedly locking and unlocking it. If repo maintenance is about to trigger, I'd rather wait until the entire backup is done instead of interleaving it into backing up pods. WDYT?

skriss:
@ncdc I believe I've addressed all your material comments, with a handful of minor things to do prior to 0.9. I also pulled your CLI commit into this PR. When you've had a chance to go through the latest code review commits, I'm going to squash this whole thing down to just 2 commits, one for vendor/ updates and one for the rest.

skriss:
We'll need to update CI to use the new `make all-containers` target.

pkg/server/generic_server.go
```diff
@@ -0,0 +1,107 @@
...
+	logger               logrus.FieldLogger
+	controllers          []arkcontroller.Interface
+	preStartControllers  func(ctx context.Context, cancelFunc context.CancelFunc) error
+	postStartControllers func(ctx context.Context, cancelFunc context.CancelFunc) error
+	cleanup              func()
+	ctx                  context.Context
+	cancelFunc           context.CancelFunc
+}
+
+func NewGenericServer(
```
ncdc: I am not sure this is helpful. There is not enough commonality between the daemonset command and the server command.

DS:
- ✅ sets `controllers`
- ❌ doesn't set `preStartControllers`
- ✅ sets `postStartControllers`
- ❌ doesn't set `cleanup`

Server:
- ❌ doesn't set `controllers`
- ✅ sets `preStartControllers`
- ✅ sets `postStartControllers`
- ✅ sets `cleanup`

If we decide to keep this, it would look better using functional options, so the constructor would be reduced to
```go
func NewGenericServer(logger logrus.FieldLogger, options ...Option)
```

The common functionality used by both DS and Server is essentially:
- handling shutdown signals
- running controllers (although the registration is handled differently between DS and Server)

Pre/PostStartControllers and Cleanup are not used by both.

I don't think it's a burden to have 2 independent `Run()` implementations each invoke `handleShutdownSignals`.

We could consider creating a "controller executor" like this:
```go
type ControllerRegistration struct {
  controller controller.Interface
  workers int
}

type ControllerExecutor struct {
  controllers []ControllerRegistration
}

func (ce *ControllerExecutor) Execute(ctx context.Context) {
        var wg sync.WaitGroup
	for _, c := range ce.controllers {
		wg.Add(1)
		go func(c ControllerRegistration) {
			c.controller.Run(ctx, c.workers)
                        wg.Done()
		}(c)
	}
}
```

Both the DS and Server structs could use this. I think I'd prefer an approach in this vein as opposed to the genericServer.

skriss: OK. I'll back this out.

skriss:
Updated to just create common funcs for getting loggers, handling signals. Can do any other refactoring later.

ncdc:
Thanks. Squash away.

skriss:
squashed and green!

PR #510, opened 2018-05-21
Fail backup if object store upload fails

### LINKED ISSUE #491: Backups that fail to upload to object storage are marked as 'Completed'
A user reported that a backup had a phase of `Completed` but the target object storage bucket was empty. Upon investigation, Ark was getting access denied errors when connecting to the bucket, so it couldn't upload. We set the backup phase to `Completed` before uploading (see https://github.com/heptio/ark/blob/master/pkg/controller/backup_controller.go#L359). We should probably check for errors uploading, and change it to `Failed` if there are any. 

Note that the phase should be `Completed` before json-encoding it for upload, so that the JSON in object storage (if it uploads successfully) has the right phase.

nrb:
Fixes #491 


pkg/cloudprovider/gcp/object_store.go
```diff
@@ -76,16 +91,23 @@ func (o *objectStore) Init(config map[string]string) error {
...
+	// The writer returned by NewWriter is asynchronous, so errors aren't guaranteed
+	// until Close() is called
 	_, err := io.Copy(w, body)
+	if err != nil {
+		return err
+	}
 
-	return errors.WithStack(err)
+	err = w.Close()
+	return err
```
nrb: I went with this approach as it's equivalent to "Update 2" on https://joeshaw.org/dont-defer-close-on-writable-files/, but without the extraneous `err == nil` check.

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -35,10 +35,25 @@ import (
 
 const credentialsEnvVar = "GOOGLE_APPLICATION_CREDENTIALS"
 
+// Interface retriever wraps the GCP SDK functions for accessing object store so they can be faked for testing.
```
ncdc: Godoc comments need to start with what they're describing (`retriever` in this case)

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -35,10 +35,25 @@ import (
...
+type retriever interface {
+	// function getWriteCloser returns a struct confirming to the io.WriteCloser interface
```
ncdc: getWriteCloser returns an io.WriteCloser that can be used to upload data to the specified bucket for the specified key.

pkg/cloudprovider/gcp/object_store.go
ncdc: I would probably call this bucketWriter for now, and eventually rename it to bucketAccessor once we migrate the rest of this file to the interface. WDYT?

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -76,16 +91,23 @@ func (o *objectStore) Init(config map[string]string) error {
...
 func (o *objectStore) PutObject(bucket string, key string, body io.Reader) error {
-	w := o.client.Bucket(bucket).Object(key).NewWriter(context.Background())
-	defer w.Close()
+	w := o.retriever.getWriteCloser(bucket, key)
 
+	// The writer returned by NewWriter is asynchronous, so errors aren't guaranteed
+	// until Close() is called
 	_, err := io.Copy(w, body)
+	if err != nil {
+		return err
```
ncdc: I don't think we want to return early here. We still should try to close the writer.

pkg/cloudprovider/gcp/object_store.go
nrb: bucketWriter works, especially given it's current contents. I'll update.

pkg/cloudprovider/gcp/object_store.go
nrb: Yeah, good catch. I'll update it.

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -76,16 +91,24 @@ func (o *objectStore) Init(config map[string]string) error {
...
-	defer w.Close()
+	w := o.bucketWriter.getWriteCloser(bucket, key)
 
+	// The writer returned by NewWriter is asynchronous, so errors aren't guaranteed
+	// until Close() is called
 	_, err := io.Copy(w, body)
 
-	return errors.WithStack(err)
+	// Ensure we close w and report errors properly
+	if closeErr := w.Close(); err == nil {
```
ncdc: This is confusing to me - normally you check the LHS error as the RHS condition. Would you mind splitting this up so it's more like this
```go
closeErr := w.Close()
if err != nil { return err }
return closeErr
```

probably should rename `err` to `copyErr`

ncdc:
LGTM (will have a couple of follow-ups that I'll do after I merge it, as Nolan's out at the moment)

Issue #511, opened 2018-05-22
Post hooks should always fire regardless of item action failures


ncdc:
Scenario:
1. pre-hook calls fsfreeze
1. some item action fails
1. return error immediately

This leaves the pod in an unusable state, since the file system is frozen.

ncdc:
WDYT @skriss?

PR #512, opened 2018-05-22
Handle errors in additionalItemBackupper

### LINKED ISSUE #509: Backups with errors aren't marked as partially complete
I had a permission issue with my cloud credentials which resulted in Ark not being able to snapshot volumes. However, without going through the logs, it seems like there is no indication of this failure: `ark backup get` lists it as ´Completed`, and `ark backup describe my-backup` only has a field for `Validation errors`, which is empty (reasonably enough).

Shouldn't there be some more indication that errors happened at least? Since I didn't see any errors, I restored the backup, and get to a semi-broken state since the pv/pvc objects aren't created.

carlpett:
Fixes #509.

Currently the corresponding test fails, which I _think_ is because of #511.


pkg/backup/item_backupper_test.go
```diff
@@ -317,11 +329,16 @@ func TestBackupItemNoSkips(t *testing.T) {
 
 			var snapshotService *arktest.FakeSnapshotService
 			if test.snapshottableVolumes != nil {
-				snapshotService = &arktest.FakeSnapshotService{
-					SnapshottableVolumes: test.snapshottableVolumes,
-					VolumeID:             "vol-abc123",
+				if test.snapshotError != nil {
```
carlpett: This is a bit of a mess - would appreciate some ideas on how to make this clearer

pkg/util/test/fake_snapshot_service.go
```diff
@@ -89,3 +89,31 @@ func (s *FakeSnapshotService) SetVolumeID(pv runtime.Unstructured, volumeID stri
 	s.VolumeIDSet = volumeID
 	return pv, nil
 }
+
```
carlpett: Not sure if this is the right place?

pkg/backup/item_backupper.go
```diff
@@ -206,7 +206,10 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 					return err
 				}
 
-				ib.additionalItemBackupper.backupItem(log, additionalItem, gvr.GroupResource())
+				err = ib.additionalItemBackupper.backupItem(log, additionalItem, gvr.GroupResource())
```
ncdc: When we are only checking the returned error and aren't doing anything else with it, we usually use the syntax `if err := foo(); err != nil {`

pkg/util/test/fake_snapshot_service.go
ncdc: Instead of creating a new type, please add `err error` to `FakeSnapshotService` and return `s.err` wherever we need to return an error.

pkg/backup/item_backupper_test.go
ncdc: See my comment below

pkg/backup/item_backupper.go
carlpett: Right. I'll do the same for the three if:s above at the same time, then?

carlpett: Ah, sorry. That doesn't work though, since they set other vars. Never mind, then, just fixing this one :)

pkg/backup/item_backupper_test.go
```diff
@@ -242,6 +243,17 @@ func TestBackupItemNoSkips(t *testing.T) {
 				"vol-abc123": {SnapshotID: "snapshot-1", AvailabilityZone: "us-east-1c"},
 			},
 		},
+		{
+			name: "backup fails when takePVSnapshot fails",
```
ncdc: To test your change to item_backupper.go, we need a test case that has this mocked behavior return a non-nil error: `additionalItemBackupper.On("backupItem", mock.AnythingOfType("*logrus.Entry"), test.customActionAdditionalItems[i], item.GroupResource).Return(nil)`.

The test case you've added here is valuable too, so we should keep it, but it doesn't cover your fix: namely, that when we call `ib.additionalItemBackupper.backupItem(log, additionalItem, gvr.GroupResource())`, it checks the returned error and returns it if it's non-nil.

ncdc: Something like this would cover your change: https://gist.github.com/ncdc/12c3b6b6f4db0e5c60f4ecf77e8d85c2

carlpett: Thanks! Updated with this, will ask some questions on Slack too :)

ncdc:
LGTM. @skriss ptal

PR #514, opened 2018-05-23
Run post-hooks even if backup actions fail

### LINKED ISSUE #511: Post hooks should always fire regardless of item action failures

carlpett:
Fixes #511 

This pulls all the action execution out into a new function (should there be matching changes in unit tests for this?), then ensures post-action hooks are run regardless of errors in the actions, and then returns any errors (including any in the post hook)

### DIFF
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	kuberrs "k8s.io/apimachinery/pkg/util/errors"
 
 	api "github.com/heptio/ark/pkg/apis/ark/v1"
 	"github.com/heptio/ark/pkg/client"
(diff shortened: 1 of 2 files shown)


ncdc:
Thanks @carlpett! For the time being, I think it's ok to leave the unit tests as they're currently structured. We may split `backupItem()` into multiple helper functions and then unit test them individually, as I wrote about [here](https://medium.com/@andy.goldstein/a-crazy-approach-to-unit-testing-in-go-3b7160bf9199), but we haven't decided on that just yet.

pkg/backup/item_backupper.go
```diff
@@ -165,6 +166,64 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 		return err
 	}
 
+	backupErrs := make([]error, 0)
+	err = ib.executeActions(log, obj, groupResource, name, namespace, metadata)
+	if err != nil {
+		log.WithError(err).Error("Error executing backup actions")
```
ncdc: I'd change "backup" to "item"

carlpett:
>This was mostly a cut & paste, right?
Yes.

pkg/backup/item_backupper.go
carlpett: Fixed!

ncdc:
LGTM

PR #515, opened 2018-05-24
WIP Support multiple backup storage backends

ncdc:


ncdc:
Not actively working on this. Might resume it later.

PR #517, opened 2018-05-26
Correct default restorePriorities docs

c-knowles:
Changed in https://github.com/heptio/ark/pull/258


ncdc:
@c-knowles thanks for the PR! Since you're updating this, there's another entry that's missing. Here is the full current list: https://github.com/heptio/ark/blob/6dbde599bfac0ee0f38f0880b51664bd8b576c67/pkg/cmd/server/server.go#L309-L317. Would you mind including limitranges too?

ncdc:
LGTM, thanks!

PR #530, opened 2018-06-06
restic quickstart doc for alpha testing

skriss:

### DIFF
--- a/docs/restic.md
+++ b/docs/restic.md
@@ -1,0 +1,92 @@
+# Restic Integration
+
+Ark now has support for backups/restores of pod volume data using [restic][1], an open-source tool for doing
+backups and restores of filesystem data. This enables you to take backups of additional Kubernetes volume 
+types beyond those with snapshot APIs that are integrated with Ark (i.e. AWS EBS, GCP PD, Azure PD). It also lays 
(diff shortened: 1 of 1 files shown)


docs/restic.md
```diff
@@ -0,0 +1,91 @@
+# Restic Integration
+
+Ark has alpha support for backups/restores of pod volume data using [restic][1], an open-source tool for doing
```
ncdc: I'd remove
- alpha
- linux

docs/restic.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+```bash
+# have Ark generate a random encryption key
+ark restic init-repository --namespace YOUR_NAMESPACE --key-size ENCRYPTION_KEY_SIZE
+```
+
+**IMPORTANT**: store this key safely and securely. All restic backup data is encrypted and cannot be accessed
+without this key.
+
```
ncdc: Add a TODO that we'll be supporting key rotation soon

docs/restic.md
ncdc: And also mention that names/commands/etc may change as we receive feedback and refine things

skriss:
Addressed comments

ncdc:
LGTM and we can iterate as needed post-merge.

Issue #286, opened 2018-01-16
Consider consolidating per-restore logs into a single file? (Similarly for backups)


skriss:
Right now, for restores, we log:
- validation errors into the restore API object (.status.validationErrors)
- info-level logs to <restore-name>-logs.gz in object storage, using logrus
- warnings/errors to <restore-name>-results.gz in object storage, as JSON

We follow a similar pattern for backups, except we don't have the third item (yet). 

I wonder whether it would make sense to consolidate this down into just a single per-backup/per-restore log file (e.g. <restore-name>-logs.gz). Since we're now using logrus, we have the ability to use differing log levels as well as other structured logging fields to differentiate between different types of log output. The main benefit that I see of consolidating is that a user now only needs to look in a single place for the full set of results from a backup or restore.  So, for a restore:

We could retain the view that we currently provide in `ark restore describe` by adding logic to that command to parse the log file and separate out the log statements into different sections within the command output.

Interested to see if other folks think this makes sense/is useful.

skriss:
See #225 for some related discussion re: backup warnings/errors.

nrb:
I think this makes sense. Consolidating everything in one place makes for a better user experience, and I don't think altering the commands to parse the logs should be too onerous.

ncdc:
Would you move validation errors from the restore resource to the log file?

skriss:
That is what I was thinking. We could add a unique log field to identify them as such if we want to retain the distinction between them and errors during/after execution.

I'm not 100% convinced this is the right idea yet, so certainly interested in alternate opinions.

ncdc:
The "nice" (maybe?) thing about validation errors is you don't have to go to object storage to see them, as they're directly on the restore in kube.

Maybe once we have #155, we could drop validation errors entirely. We'd still need some code to check for things like invalid overlapping includes and excludes, but those could just be "errors".

skriss:
^^ yeah, I was thinking along the same lines re: schema validation vs. logical validation.

ncdc:
And we could have our phases be New, InProgress, Failed (we couldn't start at all), Processed (maybe has errors and/or warnings, but we did some work and we've done as much as we could with it).

ncdc:
@jbeda this is an item we'd like to discuss soon

rosskukulinski:
My read on this would be that this is an internal or developer focused improvement.  Would this add any value for end-users?

rosskukulinski:
[...]

Issue #84, opened 2017-09-13
Add and expose metrics


ncdc:
Add various metrics using Prometheus.

timoreimann:
As announced on Slack, we (i.e., me and/or some of my team colleagues) would be keen on adding some initial support for Prometheus-based metrics collection.

Is there anything in particular that should be included from the project maintainer's / Ark community's view? The baseline for us would probably be somewhere along the lines of collecting info about

- when a backup / restore took place
- how many objects were included
- how many errors occurred

That would allow us to set up alerts like "notify me when the last successful backup is older than x time units" and "notify me when less than x objects were considered".

Implementation-wise, we usually just include the official Go Prometheus client library. So far, it has served us well.

Happy to discuss design considerations and implementation details up front.

ncdc:
@timoreimann I think all of those are good ideas. I'm also interested in things like
- time to list all pods, services, etc
- time to write the backup tarball
- time to upload a file to object storage
- size of tarball, log file, etc
- time to run custom backup actions and restorers

+1 to the official client lib. We're using `dep` now so you can run `hack/dep-save.sh`, where any arguments you pass are handed directly to `dep ensure`.

I think it would be a good idea to see a rough sketch of how you envision integrating metrics gathering into the code base.

Thanks!

timoreimann:
Appreciate the prompt feedback, @ncdc. I'll try to dedicate some time next week to dig into the relevant code sections and come up with a metrics master plan to post here. :-)

ncdc:
@timoreimann this makes sense, thanks!

FYI I did stumble across this branch where someone was already doing a POC: https://github.com/heptio/ark/compare/master...chadcatlett:stab_at_metrics. Probably worth getting @chadcatlett involved in the discussion 😄 

If we do our own registry, would it be a singleton, or would we ever want or need to have multiple?

Another thing to consider is that we're using gRPC based plugins. We'll need to figure out how to get any metrics generated by plugin instances - either by somehow getting their data into the ark server process, or by sending them to a metrics gateway or to a prometheus server directly. Something to think about...

timoreimann:
@ncdc I don't think we'll need more one than registry, so it should be a singleton basically. The design options I outlined above should only affect how (easily) we can test things IMHO.

[...]

PR #531, opened 2018-06-06
Add metrics package to publish metrics

ashish-amarnath:
Add a metrics package to handle metric registration and publishing
Add a metricsAddress field to the server struct
Start a metrics endpoint as part of starting the controllers
Report backup status on backup completion


ashish-amarnath:
Starting a WorkInProgress PR to get some feedback on the approach here to add metrics. Related to #84  Adding some code to the discussion on the issue.

ashish-amarnath:
I will sign my next commit :)

pkg/cmd/server/server.go
```diff
@@ -501,6 +509,13 @@ func (s *server) runControllers(config *api.Config) error {
 		s.logger,
 	)
 
+	go func() {
+		http.Handle("/metrics", prometheus.Handler())
```
ncdc: I would prefer not to use package-level http handler registration. Would you mind creating an `http.ServeMux`, registering the handler with that, and then passing the `ServeMux` as the 2nd parameter to `http.ListenAndServe`?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	/****Metrics related to Ark server backup***********/
+	lastBackupStatus = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: "heptio-ark", // TODO: don't hard-code this value
```
ncdc: Given that the namespace needs to be flexible, I think it might make more sense to have a struct that holds on to all the metrics, e.g.
```go
type Metrics struct {
  lastBackupStatus prometheus.Guage
}

func NewMetrics(namespace string) *Metrics {
  return &Metrics{
    lastBackupStatus: prometheus.NewGuage(
      prometheus.GaugeOpts{
        Namespace: namespace,
        // ...
      },
    ),
  }
}

func (m *Metrics) UpdateLastBackupResult(success bool) {
  if success {
    m.lastBackupStatus.Set(1)
  } else {
    m.lastBackupStatus.Set(0)
  }
}
```

WDYT?

ncdc:
Just did a very quick first pass

pkg/metrics/metrics.go
ashish-amarnath: Yes, that makes sense. I would have gone down that route to get rid of that `TODO` that I had.

ashish-amarnath:
@ncdc Thanks for the quick review! I've made changes to address your comments. I will now start adding some metrics

pkg/metrics/metrics.go
```diff
@@ -1,30 +1,49 @@
...
-			Namespace: "heptio-ark", // TODO: don't hard-code this value
-			Name:      "last_backup_status",
-			Help:      "Status of the most recent backup. 1 if it was successful, 0 otherwise",
-		},
-	)
-)
+	lastBackupStatus prometheus.Gauge
+}
+
+var metrics *serverMetrics
```
ncdc: Sorry, 1 more change here, please 😄. I'd prefer not to have any package level variables.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+				Help:      "Status of the most recent backup. 1 if it was successful, 0 otherwise",
+			},
+		),
+	}
+}
+
+// GetMetricsMux returns a mux with the metrics handle
+func GetMetricsMux() *http.ServeMux {
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", prometheus.Handler())
```
ncdc: It looks like the new-fangled way to do this is to use `promhttp.Handler()`.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type serverMetrics struct {
+	/****Metrics related to Ark server backup***********/
+	lastBackupStatus prometheus.Gauge
+}
+
+var metrics *serverMetrics
```
ncdc: I'd like to avoid the package-level variable. Would you mind looking at https://github.com/heptio/gimbal/blob/master/discovery/pkg/metrics/metrics.go and basing this file on that one?

ashish-amarnath: that's helpful!

pkg/cmd/server/server.go
```diff
@@ -96,7 +98,14 @@ func NewCommand() *cobra.Command {
 			}
 			namespace := getServerNamespace(namespaceFlag)
 
-			s, err := newServer(namespace, fmt.Sprintf("%s-%s", c.Parent().Name(), c.Name()), pluginDir, logger)
+			metricAddressFlag := c.Flag("metric-address")
```
ncdc: Please add a variable at the top of this function (where `logLevelFlag` and `pluginDir` are) and set it to the default address, and then please bind it similar to how we bind `log-level` and `plugin-dir`). Also please call the flag `metrics-address`.

pkg/cmd/server/server.go
```diff
@@ -123,8 +132,17 @@ func getServerNamespace(namespaceFlag *pflag.Flag) string {
 	return api.DefaultNamespace
 }
 
+func getMetricAddress(metricAddressFlag *pflag.Flag) string {
```
ncdc: You shouldn't need this function any more once you do what I commented above.

pkg/apis/ark/v1/constants.go
```diff
@@ -37,4 +37,8 @@ const (
 	// NamespaceScopedDir is the name of the directory containing namespace-scoped
 	// resource within an Ark backup.
 	NamespaceScopedDir = "namespaces"
+
+	// DefaultMetricAddress is the port at which the metric server will listen on to make the
```
ncdc: Let's call this DefaultMetricsAddress please. I'd also move the constant to `pkg/cmd/server/server.go`.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,68 @@
...
+package metrics
+
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// ServerMetrics Prometheus metrics for the server
```
ncdc: ServerMetrics contains Prometheus metrics for the Ark server.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,68 @@
...
+type ServerMetrics struct {
+	/****Metrics related to Ark server backup***********/
```
ncdc: Not sure the comment is needed?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,68 @@
...
+			prometheus.GaugeOpts{
+				Namespace: namespace,
+				Name:      lastBackupStatus,
+				Help:      "Status of the most recent backup. 1 if it was successful, 0 otherwise",
+			},
+		),
+	}
+}
+
+// GetMetricsMux returns a mux with the metrics handle
```
ncdc: Let's get rid of this function and set up the mux where you spawn the goroutine in server.go so this package doesn't need to know anything about http.

pkg/cmd/server/server.go
```diff
@@ -501,6 +514,15 @@ func (s *server) runControllers(config *api.Config) error {
 		s.logger,
 	)
 
+	go func() {
+		metricsMux := http.NewServeMux()
+		metricsMux.Handle("/metrics", promhttp.Handler())
+		err := http.ListenAndServe(s.metricsAddress, metricsMux)
+		s.logger.Fatalf("Failed to start metrics: %v", err)
```
ncdc: Let's only make the fatal call if err != nil

ncdc: Can you please also verify that if you run `ark server` by hand and you do `ctrl-c` that it shuts down cleanly and this doesn't keep it running?

ashish-amarnath: oops! Thanks for catching that.

pkg/controller/backup_controller.go
```diff
@@ -269,16 +273,22 @@ func (controller *backupController) processBackup(key string) error {
...
 	logContext.Debug("Running backup")
 	// execution & upload of backup
+	backupName := kubeutil.NamespaceAndName(backup)
+	backupScheduleName := getBackupScheduleName(backup)
+	controller.metrics.RegisterBackupAttempt(backupName, backupScheduleName)
+
 	if err := controller.runBackup(backup, controller.bucket); err != nil {
 		logContext.WithError(err).Error("backup failed")
 		backup.Status.Phase = api.BackupPhaseFailed
+		controller.metrics.RegisterBackupFailed(backupName, backupScheduleName)
```
nrb: I think the `RegisterBackupSuccess` case may need to be moved into an `else` clause here, such that a backup doesn't register as failed and succeeded.

I tried to run a backup using a GCP serviceaccount that didn't have access to the bucket selected, and my metrics stated that the backup failed and succeeded.

```
ERRO[0436] Error uploading log file                      bucket=<BUCKET_REDACTED> error="rpc error: code = Unknown desc = googleapi: Error 403: <ACCOUNT_REDACTED>.iam.gserviceaccount.com does not have storage.objects.create access to <BUCKET_REDACTED>/default-backup/default-backup-logs.gz., forbidden" key=default-backup/default-backup-logs.gz logSource="pkg/cloudprovider/backup_service.go:148"
```

`ark backup get` shows the backup as failed.

```
% ./_output/bin/linux/amd64/ark backup get
NAME             STATUS    CREATED                         EXPIRES   SELECTOR
default-backup   Failed    2018-06-12 09:52:38 -0400 EDT   29d       <none>
```

These are the metrics:

```
ark_server_backup_attempt_total{backupName="heptio-ark/default-backup",schedule="UNSCHEDULED"} 1
ark_server_backup_failed_total{backupName="heptio-ark/default-backup",schedule="UNSCHEDULED"} 1
ark_server_backup_size_bytes{backupName="heptio-ark/default-backup",schedule="UNSCHEDULED"} 4865
ark_server_backup_success_total{backupName="heptio-ark/default-backup",schedule="UNSCHEDULED"} 1
```

pkg/controller/backup_controller.go
```diff
@@ -385,3 +406,11 @@ func closeAndRemoveFile(file *os.File, log logrus.FieldLogger) {
 		log.WithError(err).WithField("file", file.Name()).Error("error removing file")
 	}
 }
+
+func getBackupScheduleName(backup *api.Backup) string {
+	scheduleName := kubeutil.LabelValue(backup, "ark-schedule")
+	if scheduleName == "" {
+		scheduleName = "UNSCHEDULED"
```
rosskukulinski: If a Backup doesn't have a Schedule, we should simply not add the `schedule` label to the metrics rather than adding an arbitrary value..  This will simplify the prometheus queries later on.

docs/cli-reference/ark_server.md
```diff
@@ -14,9 +14,10 @@ ark server [flags]
 ### Options
 
 ```
-  -h, --help                help for server
-      --log-level           the level at which to log. Valid values are debug, info, warning, error, fatal, panic. (default info)
-      --plugin-dir string   directory containing Ark plugins (default "/plugins")
+  -h, --help                     help for server
+      --log-level                the level at which to log. Valid values are debug, info, warning, error, fatal, panic. (default info)
+      --metrics-address string   the address to expose prometheus metrics (default ":8085")
```
rosskukulinski: Should this be simplified to just port? Or does this allow us to specify a listen address & port (e.g. localhost:8085 or $public_ip:8085)?

pkg/controller/backup_controller.go
ashish-amarnath: @nrb thanks for running the test for me and catching this bug! Pushed a change with the fix.

docs/cli-reference/ark_server.md
ashish-amarnath: This does allow you to specify an address. In this case, it will always be `localhost:8085` or just `:8085`.

pkg/controller/backup_controller.go
ashish-amarnath: I think this allows users to collect metrics for all unscheduled backups easily. I don't have a strong preference on this though.

pkg/cmd/server/server.go
```diff
@@ -501,6 +514,16 @@ func (s *server) runControllers(config *api.Config) error {
 		s.logger,
 	)
 
+	go func() {
+		metricsMux := http.NewServeMux()
+		metricsMux.Handle("/metrics", promhttp.Handler())
+		if err := http.ListenAndServe(s.metricsAddress, metricsMux); err != nil {
```
stevesloka: Small nit: It's nice to log the address/port that the metrics are using. It's helpful in troubleshooting them not working.

pkg/controller/backup_controller.go
rosskukulinski: How I've seen this work with Prometheus/Grafana is to use label matching operators:

`ark_server_backup_attempt_total{ark-schedule=".+"}` --> Matches any backup with a schedule specified
`ark_server_backup_attempt_total{ark-schedule=~"${Schedule}"` --> matches any backup with schedule using a variable, defaults to ".+" for "All", ignores any backups without a schedule
`ark_server_backup_attempt_total{ark-schedule!=".+"}` --> Matches all backups without a schedule

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,109 @@
...
+
+// ServerMetrics contains Prometheus metrics for the Ark server.
+type ServerMetrics struct {
+	metrics map[string]prometheus.Collector
+}
+
+const (
+	metricNamespace      = "ark_server"
+	backupSizeBytesGauge = "backup_size_bytes"
+	backupAttemptsCount  = "backup_attempts_total"
```
ncdc: Is it typical to append something like `total` here - it seems weird to me.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,109 @@
...
+	backupSuccessCount   = "backup_success_total"
```
ncdc: If we keep total, this name is fine as-is. Otherwise I'd maybe call it `backup_successes`

Also note that we may move away from success/failure for backups. For restores, the valid phases are new, in progress, failed validation, and completed. In each restore's status, we record the number of warnings and errors that occurred, but as long as we're able to restore something, the status is always completed. We are considering doing something similar for backups. (#286 is related)

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,109 @@
...
+	backupFailedCount    = "backup_failed_total"
```
ncdc: If we keep total, I'd call this `backup_failure_total`; otherwise, `backup_failures`.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// ServerMetrics contains Prometheus metrics for the Ark server.
+type ServerMetrics struct {
+	metrics map[string]prometheus.Collector
+}
+
+const (
+	metricNamespace      = "ark_server"
+	backupSizeBytesGauge = "backup_size_bytes"
```
ncdc: @rosskukulinski please compare with your naming suggestions [here](https://github.com/heptio/ark/issues/84#issuecomment-393983911)

pkg/util/kube/utils.go
```diff
@@ -28,6 +28,14 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 )
 
+// LabelValue looks up and returns the value for a label on the object
+func LabelValue(objMeta metav1.Object, labelName string) string {
+	if value, exists := objMeta.GetLabels()[labelName]; exists {
```
ncdc: Given that labels are `map[string]string` and the zero value of any undefined key is the empty string, do we need this function?

pkg/controller/backup_controller.go
```diff
@@ -269,9 +273,16 @@ func (controller *backupController) processBackup(key string) error {
 
 	logContext.Debug("Running backup")
 	// execution & upload of backup
+	backupName := kubeutil.NamespaceAndName(backup)
+	backupScheduleName := kubeutil.LabelValue(backup, "ark-schedule")
+	controller.metrics.RegisterBackupAttempt(backupName, backupScheduleName)
```
ncdc: If there's no schedule, won't this generate a new metric per backup, e.g.
```
ark_server_backup_attempts_total{backupName="foo1", schedule=""} 1
ark_server_backup_attempts_total{backupName="foo2", schedule=""} 1
ark_server_backup_attempts_total{backupName="foo3, schedule=""} 1
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,109 @@
...
+func NewServerMetrics() *ServerMetrics {
+	return &ServerMetrics{
+		metrics: map[string]prometheus.Collector{
+			backupSizeBytesGauge: prometheus.NewGaugeVec(
+				prometheus.GaugeOpts{
+					Namespace: metricNamespace,
+					Name:      backupSizeBytesGauge,
+					Help:      "Size, in bytes, of a backup",
+				},
+				[]string{"schedule", "backupName"},
```
ncdc: Let's use constants for these. Also please rename `backupName` to `backup`.

pkg/controller/backup_controller.go
ashish-amarnath: Yes, it will. I thought that is the behaviour @rosskukulinski was preferring, no?

ncdc: I don't think we want that

pkg/metrics/metrics.go
rosskukulinski: @ncdc it's a normal thing - it can help with understanding the units of metrics.  See:  https://prometheus.io/docs/practices/naming/

ncdc: Thanks, prom newbie here :)

ashish-amarnath: That's the convention I've seen for counters. The idea is we are counting 'events' and the value is the total number of events that have occurred, hence the 'total'.

pkg/metrics/metrics.go
rosskukulinski: Is the `backup_size_bytes` the size of the resulting tarbal?  If so, I suggest `ark_backup_tarball_size_bytes` because that's different than the PV-related sizes.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// ServerMetrics contains Prometheus metrics for the Ark server.
+type ServerMetrics struct {
+	metrics map[string]prometheus.Collector
+}
+
+const (
+	metricNamespace      = "ark_server"
```
ncdc: Do we want the `_server` part of this? I'm leaning toward just `ark`.

pkg/metrics/metrics.go
ashish-amarnath: When we make that change, we can call this metric `backup_completed_total`. Or even have a metric per phase of the backup. WDYT?

ncdc: Sure, and we can discuss it if/when we make that change.

pkg/metrics/metrics.go
ashish-amarnath: done

pkg/util/kube/utils.go
ashish-amarnath: removed

pkg/controller/backup_controller.go
ashish-amarnath: I take back my initial answer. I am not sure if it will create a metric per backup. trying to get the correct answer

ashish-amarnath: OK.. after consulting with a collegue who knows more prom than me! we might just drop the backup name altogether and just have the schedule name.

pkg/metrics/metrics.go
ashish-amarnath: renamed.

pkg/controller/backup_controller.go
```diff
@@ -269,9 +273,15 @@ func (controller *backupController) processBackup(key string) error {
 
 	logContext.Debug("Running backup")
 	// execution & upload of backup
+	backupScheduleName, _ := backup.GetLabels()["ark-schedule"]
```
ncdc: You don't need the `, _`

pkg/controller/backup_controller.go
```diff
@@ -359,19 +369,29 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
...
+		errs = append(errs, errors.Wrap(err, "error getting file info"))
+	} else {
+		backupSizeBytes = backupFileStat.Size()
+	}
+
+	if err := controller.backupService.UploadBackup(bucket, backup.Name, backupJSONToUpload, backupFileToUpload, logFile); err != nil {
 		errs = append(errs, err)
 	}
 
+	backupScheduleName, _ := backup.GetLabels()["ark-schedule"]
```
ncdc: ditto

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,111 @@
...
+
+// ServerMetrics contains Prometheus metrics for the Ark server.
+type ServerMetrics struct {
+	metrics map[string]prometheus.Collector
+}
+
+const (
+	metricNamespace             = "ark_server"
+	backupTarballSizeBytesGauge = "backup_tarball_size_bytes"
+	backupAttemptsCount         = "backup_attempts_total"
```
ncdc: I would like to be consistent with singular vs plural and our nomenclature. Should this be attempt or attempts?

success or successes?

failure or failures?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// ServerMetrics contains Prometheus metrics for the Ark server.
+type ServerMetrics struct {
+	metrics map[string]prometheus.Collector
+}
+
+const (
+	metricNamespace             = "ark_server"
```
ncdc: We're already in the `metrics` package - can we call this `namespace` instead?

skriss:
I don't have any additional comments. Excited for this!

PR #534, opened 2018-06-07
replace complete-restore.sh with go code

skriss:


skriss:
Updated with lots of comments

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -250,40 +251,94 @@ func (c *podVolumeRestoreController) processRestore(req *arkv1api.PodVolumeResto
...
+	if err != nil {
+		return errors.Wrap(err, "error identifying path of volume")
+	}
+
+	// Move the contents of the staging directory into the new volume directory to finalize the restore. This
+	// is being executed through a shell because attempting to do the same thing in go (via os.Rename()) is
+	// giving errors about renames not being allowed across filesystem layers in a container. This is occurring
+	// whether /restores is part of the writeable container layer, or is an emptyDir volume mount. This may
+	// be solvable but using the shell works so not investigating further.
+	if _, stderr, err := runCommand(exec.Command("/bin/sh", "-c", fmt.Sprintf("mv %s/* %s/", restorePath, volumePath))); err != nil {
```
ncdc: Does this handle dotfiles?

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -292,18 +347,16 @@ func (c *podVolumeRestoreController) processRestore(req *arkv1api.PodVolumeResto
...
 	}
 
-	cmd := exec.Command("/bin/sh", "-c", strings.Join([]string{"/complete-restore.sh", string(req.Spec.Pod.UID), volumeDir, string(restoreUID)}, " "))
-
-	if stdout, stderr, err = runCommand(cmd); err != nil {
-		log.WithError(errors.WithStack(err)).Errorf("Error running command=%s, stdout=%s, stderr=%s", resticCmd.String(), stdout, stderr)
-		return c.fail(req, fmt.Sprintf("error running restic restore, stderr=%s: %s", stderr, err.Error()), log)
+	// Create the .ark directory within the volume dir so we can write a done file
+	// for this restore.
+	if err := os.MkdirAll(filepath.Join(volumePath, ".ark"), 0755); err != nil {
```
ncdc: Would it be possible to have the done file go somewhere other than inside the actual volume?

skriss: it would need to be somewhere that the init container could see it - could add an emptyDir volume to the pod specifically for done files

ncdc: Oh right this is the communications "protocol" between the DS pod and the actual pod.

What happens if you do
1. backup
1. restore (which creates the .ark done file)
1. backup
1. restore

Does the 2nd backup include the done file, and does it trick the 2nd restore into thinking it's finished before it actually is?

skriss: It doesn't trick the restore - there's a .ark/ dir, but the file is named with the restore's UID, and the init container is given this UID and looks specifically for file with that name. I can add code to not copy over prior done files when doing a restore (had it before but it got dropped somewhere along the way)

ncdc: Ok cool, glad it's UID based. If you want to add the exclusion, that's fine, or you can just put in a TODO for now.

pkg/controller/pod_volume_restore_controller.go
skriss: checking

skriss: Good call, did not, fixed in new commit

pkg/controller/pod_volume_restore_controller.go
skriss: added a TODO for now

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -292,18 +347,16 @@ func (c *podVolumeRestoreController) processRestore(req *arkv1api.PodVolumeResto
...
 	}
-	log.Debugf("Ran command=%s, stdout=%s, stderr=%s", resticCmd.String(), stdout, stderr)
 
-	// update status to Completed
-	if _, err = c.patchPodVolumeRestore(req, updatePodVolumeRestorePhaseFunc(arkv1api.PodVolumeRestorePhaseCompleted)); err != nil {
-		log.WithError(err).Error("Error setting phase to Completed")
-		return err
+	// Write a done file with name=<restore-uid> into the just-created .ark dir
+	// within the volume.
+	if err := ioutil.WriteFile(filepath.Join(volumePath, ".ark", string(restoreUID)), nil, 0755); err != nil {
```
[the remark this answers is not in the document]
skriss: fixed

skriss:
@ncdc I'm happy with the matching/moving approach & code as implemented now. let me know if you feel otherwise.

ncdc:
Ok I’ll take a closer look over the weekend or Monday

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -250,40 +252,99 @@ func (c *podVolumeRestoreController) processRestore(req *arkv1api.PodVolumeResto
...
+	// 	.[^.]* : a dot followed by a non-dot and anything else (dotfiles starting with a single dot, excluding '.')
+	// 	..?*   : two dots followed by any single character and anything else (dotfiles starting with two dots, excluding '..')
+	if err := moveMatchingFiles(restorePath, volumePath, "[^.]*", ".[^.]*", "..?*"); err != nil {
+		return errors.Wrapf(err, "error moving files from restore staging directory into volume")
+	}
+
+	// The staging directory should be empty at this point since we moved everything, but
+	// make sure.
+	if err := os.RemoveAll(restorePath); err != nil {
+		return errors.Wrap(err, "error removing restore staging directory and its contents")
```
ncdc: Are there any conceivable situations where we successfully moved the matching files, but we get an error on the RemoveAll call? Do we need to be more detailed here in reporting the error - list all the files/dirs in `restorePath` or something?

ncdc: If we return here, we never write the done file, so the pod's init container will be waiting forever. What do you think we can do to shed more light for these waiting init containers?

ncdc:
We should add unit tests 😄

pkg/controller/pod_volume_restore_controller.go
skriss: I suppose it is possible there's an error. Now that you raise it, I'm thinking this shouldn't be an error for the restore - since this is just (probably redundant) cleanup in the daemonset pod.

skriss: I think it makes sense just to add a warning to the server log here - since this should just be removing the empty staging directory. Do you think that's reasonable?

ncdc:
Sgtm

pkg/controller/pod_volume_restore_controller.go
skriss: Updated.

skriss:
Do you want to block on units? I need to do that across all of the new code, so would prefer to get this merged for the alpha and then circle back holistically on units.

ncdc:
No need to block

ncdc:
LGTM

PR #537, opened 2018-06-08
Dockerfile: move restic fetch before ark add so it can be cached

skriss:

Current ordering means restic gets re-fetched every time the ark binary changes.


ncdc:
LGTM

PR #541, opened 2018-06-11
rename restic-init-container to ark-restic-restore-helper

skriss:


pkg/restore/restic_restore_action.go
```diff
@@ -66,7 +66,7 @@ func (a *resticRestoreAction) Execute(obj runtime.Unstructured, restore *api.Res
 
 	initContainer := corev1.Container{
 		Name:  restic.InitContainer,
-		Image: "gcr.io/heptio-images/restic-init-container:latest",
+		Image: "gcr.io/heptio-images/ark-restic-restore-helper:latest",
```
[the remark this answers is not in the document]
skriss: TODO'ed

ncdc:
Actually since we aren't going to be setting `latest` for `v0.9.0.alpha.1`, we may need to handle that TODO now. WDYT?

skriss:
can we just grab buildinfo.Version? That's an easy change

ncdc:
That helps with the tag but not the rest of the image URL. But it's a good start, so let's do it. And have the cli flag before we cut 0.9 final.

pkg/restore/restic_restore_action.go
```diff
@@ -25,20 +27,32 @@ import (
...
 	}
 }
 
+func initContainerImage() string {
+	tag := buildinfo.Version
+	if tag == "" {
+		tag = "latest"
+	}
+
+	return fmt.Sprintf("gcr.io/heptio-images/ark-restic-restore-helper:%s", tag)
```
ncdc: You lost the TODO

skriss: urgh. cut and failed to paste :/

skriss: upd.

ncdc:
LGTM

PR #542, opened 2018-06-11
Add TAG_LATEST support to Makefile

ncdc:
Only tag an image as "latest" if TAG_LATEST is set to "true".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,8 @@ ARCH ?= linux-amd64
 
 VERSION ?= master
 
+TAG_LATEST ?= false
+
 ###
 ### These variables should not need tweaking.
 ###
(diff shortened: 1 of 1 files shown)


ncdc:
@asenchi @skriss if I redirect the output from `git describe` to `/dev/null`, the overall `if` condition is always false. If you have suggestions for how to do this better, please share!

asenchi:
> if I redirect the output from git describe to /dev/null, the overall if condition is always false. If you have suggestions for how to do this better, please share!

I think this is because you are using double brackets which considers an empty string "false". I'm not sure you even need the `&& ...` here as that a check to ensure we tagged latest on a release. The envvar accomplishes that now.

ncdc:
@asenchi so you think the env var is entirely sufficient? WDYT @skriss @nrb ?

asenchi:
At least as far as I understand it, I do think it is sufficient. Perhaps there is a situation I am missing.

ncdc:
Since a human should be the one manually setting it to true, I'm ok removing the tag check.

ncdc:
Updated

PR #545, opened 2018-06-12
update restic setup instructions

skriss:

Update to point to v0.9.0-alpha.1 artifacts


ncdc:
LGTM

PR #547, opened 2018-06-13
use rsync to copy restored data from staging dir to volume dir

skriss:


ncdc:
LGTM

PR #532, opened 2018-06-06
add ResticRepository CRD and move repo-level actions to a controller

skriss:
Signed-off-by: Steve Kriss <steve@heptio.com>

@ncdc still a WIP, but take a look at what I've done so far if you have time. Known TODOs are captured inline in comments.

PR #550, opened 2018-06-13
[WIP] Add pod volume backups/restores to ark backup/restore describe

skriss:
Builds on #532 

Added a section to the `ark backup describe` output to show associated pod volume backups.  It looks like:

```
Pod Volume Backups:
  Completed:
    workload/restic-test-workload-fd54b5b48-59qdx: volume-1, volume-2
```

The first grouping is by phase (Completed, Failed, In Progress), and only those groups that have PVBs are shown. Within each phase, there's a line per pod: shown as `<namespace>/<pod>: volume-1, volume-2, ...`

There will potentially be a lot of info to show for larger backups, so I tried to keep it compact while still enabling seeing what succeeded/failed.

Feedback on format and code requested before I move onto `ark restore describe`.


rosskukulinski:
If a Pod has two volumes, one that is Completed, one that is Failed -- will the Pod get listed twice (once for reach section)?

At first glance, I have any nits, but I agree -- for even a medium sized cluster that section is going to be HUGE.

skriss:
@rosskukulinski yeah, that's how I have it currently. The thinking there was I wanted to make it really easy to see at a glance if/what failed, rather than having to scan through a bunch of different sections like you would if it was grouped at the top level by namespace/pod

rosskukulinski:
@skriss We should put the `Failed` section first for easier scanning.

Which brings up a good question: Does a user need to see the list of all Pods & Volumes that were successful?  Could we get away with listing Failed & In Progress, and have a count for Succesfull?  

`Completed: 55 of 60 volumes across 35 Pods`

skriss:
That's definitely an option (I considered it as well). I do think there's sometimes value in seeing the specifics, so maybe we collapse to counts by default, and have a flag that lists everything out

rosskukulinski:
Is there prior art for the `describe` command in k8s to accept output configuration?

Another option would be to list it all out up to a certain length, then collapse (especially the `Completed section).

rosskukulinski:
in #579 I proposed adding another section, `Skipped` to help users identify volumes that weren't backed up.

skriss:
Could definitely use some UX help to sort out what this should look like.

skriss:
@rosskukulinski @ncdc I think we should get this in for the official v0.9.0 release, since otherwise the only way to see details on restic backups is to look at YAML. We just need to decide on a reasonable format to go out with. WFYT?

ncdc:
👍 let me do some thinking about UX ideas

ncdc:
@rosskukulinski 
> Is there prior art for the describe command in k8s to accept output configuration?

Minimally. The only control you have for `kubectl describe` is to turn off showing events related to the item you're describing (it defaults to true).

However, I think it would be fine for `ark backup describe` to show the restic backup summary, and to use a flag (e.g. `--volume-details`) to show the full details about the restic snapshots.

@skriss 
I'd like to suggest we be explicit about restic - instead of saying `Pod Volume Backups`, let's call them something like `Restic Backups`. Or were you thinking we could stay generic and eventually move the `BlockStorage` plugin snapshot metadata there too (when we move it out of `backup.status.volumeBackups`)?

skriss:
Updated so the default format looks like:
```
Restic Backups (specify --volume-details for more information):
  Completed:    1
  In Progress:  1
```

And specifying `--volume-details` gives:
```
Restic Backups:
  Completed:
    default/restic-test-workload-5b4c865bcb-ncvsh: volume-1
  In Progress:
    default/restic-test-workload-5b4c865bcb-ncvsh: volume-2
```

Where restic backups are grouped by status (Completed, Failed, In Progress, New), and then sorted by namespace+pod

pkg/cmd/util/output/backup_describer.go
```diff
@@ -241,3 +247,98 @@ func failedDeletionCount(requests []v1.DeleteBackupRequest) int {
 	}
 	return count
 }
+
+// DescribePodVolumeBackups describes pod volume backups in human-readable format.
+func DescribePodVolumeBackups(d *Describer, backups []v1.PodVolumeBackup, details bool) {
+	if len(backups) == 0 {
+		d.Printf("Restic Backups: <none>\n")
```
ncdc: This is never printed based on the check on line 51?

pkg/cmd/util/output/backup_describer.go
```diff
@@ -241,3 +247,98 @@ func failedDeletionCount(requests []v1.DeleteBackupRequest) int {
...
+	} else {
+		if details {
+			d.Printf("Restic Backups:\n")
+		} else {
+			d.Printf("Restic Backups (specify --volume-details for more information):\n")
+		}
+
+		// separate backups by phase (combining <none>, New, InProgress into a single group)
```
ncdc: InProgress has its own group?

pkg/cmd/util/output/backup_describer.go
```diff
@@ -241,3 +247,98 @@ func failedDeletionCount(requests []v1.DeleteBackupRequest) int {
...
+						volumes:   []string{backup.Spec.Volume},
+					}
+				} else {
+					group.volumes = append(group.volumes, backup.Spec.Volume)
+				}
+			}
+
+			// move all the groupings (grouped by pod) into a slice so we can sort it
+			var backupsByPodSlice []*podVolumeBackupGroup
+			for _, backups := range backupsByPod {
```
ncdc: Instead of looping here, could we add to the slice right after you create a new podVolumeBackupGroup and add it to the map in the loop above?

skriss: Yes. Ended up extracting the map+slice into a separate struct to make the main logic simpler.

pkg/cmd/util/output/backup_describer.go
```diff
@@ -241,3 +247,111 @@ func failedDeletionCount(requests []v1.DeleteBackupRequest) int {
...
+// label is "namespace/name".
+type backupsByPod struct {
+	backupsByPodMap   map[string]*podVolumeBackupGroup
+	backupsByPodSlice []*podVolumeBackupGroup
+}
+
+// Add adds a pod volume backup with the specified pod namespace, name
+// and volume to the appropriate group.
+func (b *backupsByPod) Add(namespace, name, volume string) {
+	if b.backupsByPodMap == nil {
```
ncdc: If you wanted to, you could init the map above where you're calling `backupsByPod := new(backupsByPod)`, to avoid this check here. But not necessary.

rosskukulinski:
Looking good, @skriss.  Earlier I made a [comment](https://github.com/heptio/ark/pull/550#issuecomment-399979288) about having a `Skipped` section.  I think we're not currently at a place to plumb that through at this point?

skriss:
@rosskukulinski yeah, that would be difficult right now, the only way to access that information would be to open the backup tarball and read through all the manifests to get the list of pod volumes that weren't backed up, which I don't think we want to do.  This could be easier if we add a manifest to backups that includes info about pod volumes.

rosskukulinski:
LGTM then

ncdc:
LGTM

ncdc:
Is this ready?

skriss:
squashed; tested it and looked good to me

PR #551, opened 2018-06-14
Expose metrics server port from container

ashish-amarnath:
Expose metrics server port from the container in example deployment and pkg/install tooling

Signed-off-by: Ashish Amarnath <ashish.amarnath@gmail.com>


ncdc:
LGTM, thanks!

PR #553, opened 2018-06-14
Log to stdout instead of stderr

### LINKED ISSUE #507: All logs from Ark server containers on k8s/GKE show up in Stackdriver as ERROR logs
screen shot 2018-05-18 at 5 00 55 pm

Example of a specific log message:

redacted

(Notice the `ERROR` severity.)

We're using the default example Ark server Deployment available at https://github.com/heptio/ark/blob/master/examples/gcp/10-deployment.yaml

Does the Ark server log to stderr by default by any chance? We've seen this kind of behavior in the past with services that deploy to stderr (anything logged to stderr by a container in a k8s cluster will show up as ERROR logs in Stackdriver).

ncdc:
Fixes #507

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 	"reflect"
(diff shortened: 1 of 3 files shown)


ncdc:
Better test - `ark server >ark.stdout 2>ark.stderr` - make sure that everything goes to `ark.stdout` and that per-backup and per-restore logs continue to work.

nrb:
Double checked with the separate redirects - backup and restore logs were properly populated. Nothing went to stderr during standard operation.

A message did come across stderr if I provided no credentials or incorrect ones:

```
x1c in /home/nrb/go/src/github.com/heptio/ark (git) master
% ./_output/bin/linux/amd64/ark server  >ark.stdout 2>ark.stderr

x1c in /home/nrb/go/src/github.com/heptio/ark (git) master
% tail ark.stderr
An error occurred: rpc error: code = Unknown desc = GOOGLE_APPLICATION_CREDENTIALS is undefined
```

This seems correct to me, as it's an error [generated by the plugin](https://github.com/heptio/ark/blob/master/pkg/cloudprovider/gcp/object_store.go#L66), and prevents Ark from actually launching. It would still show up in GKE's logs, though.

ncdc:
Yes I think that's fine. It's an error that occurs during server startup.

skriss:
LGTM.

PR #555, opened 2018-06-14
get rid of restore staging dir by backing up/restoring within volume dir

skriss:

By running `restic backup` from within the volume's root directory, we can then `restic restore` directly into the volume root dir.

Also fixed a bug in the PVR controller, in the code that was checking if the restic initcontainer was running.


ncdc:
LGTM

PR #557, opened 2018-06-15
Use ResticRepository's resticIdentifier field as the source of truth for repo IDs

skriss:
Builds on #532 , only the last commit is new.

This PR:

- replaces the `RepoPrefix` field on `PodVolumeBackup` and `PodVolumeRestore` with a `RepoIdentifier` field, which contains the full restic repo identifier (populated by the Ark server by pulling from the relevant `ResticRepository` CR)
- removes the global `repoPrefix` field from the `pkg/restic/repositoryManager` and instead sources repos' identifiers from the `ResticRepository` CRs (paving the way for supporting heterogeneous restic backends - no need for them all to be the same, or to be AWS/Azure/GCP anymore)
- checks for a ready repo before trying to create/run a `PodVolumeRestore`, similarly to for backups

### DIFF
--- a/pkg/restic/backupper.go
+++ b/pkg/restic/backupper.go
@@ -84,6 +84,31 @@ func resultsKey(ns, name string) string {
 	return fmt.Sprintf("%s/%s", ns, name)
 }
 
+func getRepo(repoLister arkv1listers.ResticRepositoryLister, ns, name string) (*arkv1api.ResticRepository, error) {
+	repo, err := repoLister.ResticRepositories(ns).Get(name)
+	if apierrors.IsNotFound(err) {
+		return nil, errors.Wrapf(err, "restic repository not found")
+	}
+	if err != nil {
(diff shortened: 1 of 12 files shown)


ncdc:
last commit lgtm

skriss:
rebased, ready for merge.

pkg/restic/backupper.go
```diff
@@ -84,22 +84,41 @@ func resultsKey(ns, name string) string {
 	return fmt.Sprintf("%s/%s", ns, name)
 }
 
+func getRepo(repoLister arkv1listers.ResticRepositoryLister, ns, name string) (*arkv1api.ResticRepository, error) {
+	repo, err := repoLister.ResticRepositories(ns).Get(name)
+	if apierrors.IsNotFound(err) {
```
ncdc: Do we get any benefit by distinguishing the kind of error here?

PR #561, opened 2018-06-19
move restic binary into /usr/bin so it's in PATH

skriss:

This keeps local running of the Ark server easy - just requires restic to be in your local path.

### DIFF
--- a/pkg/restic/command.go
+++ b/pkg/restic/command.go
@@ -35,7 +35,7 @@ type Command struct {
 
 // StringSlice returns the command as a slice of strings.
 func (c *Command) StringSlice() []string {
-	res := []string{"/restic"}
+	res := []string{"restic"}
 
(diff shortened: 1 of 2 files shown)


pkg/restic/command.go
```diff
@@ -35,7 +35,7 @@ type Command struct {
 
 // StringSlice returns the command as a slice of strings.
 func (c *Command) StringSlice() []string {
-	res := []string{"/restic"}
+	res := []string{"restic"}
```
ncdc: Should we do `exec.LookPath` somewhere?

skriss: why would we need that? that would get us the full path, but if it's in $PATH then we don't need it, right?

ncdc: I couldn't remember if it would find it or not. If it works as-is, no need.

PR #563, opened 2018-06-20
add a /scratch emptyDir to ark pods to use for restic cache

skriss:

### DIFF
--- a/examples/aws/10-deployment.yaml
+++ b/examples/aws/10-deployment.yaml
@@ -43,12 +43,18 @@ spec:
               mountPath: /credentials
             - name: plugins
               mountPath: /plugins
+            - name: scratch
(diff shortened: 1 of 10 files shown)


ncdc:
What is the default cache dir if we don't specify it?

skriss:
See https://github.com/restic/restic/blob/master/internal/cache/dir.go#L53

ncdc:
Ok and we want this so the pod can use a volume instead of the container's filesystem

skriss:
Yep, and also, in the current state, the ark server pod can't use the container FS because it doesn't have permissions

skriss:
(tested on GCP, looks good)

ncdc:
LGTM

PR #564, opened 2018-06-20
Record backup start and completion times, add timing metrics

nrb:

### DIFF
--- a/pkg/cmd/util/output/backup_describer.go
+++ b/pkg/cmd/util/output/backup_describer.go
@@ -162,9 +162,22 @@ func DescribeBackupStatus(d *Describer, status v1.BackupStatus) {
 	d.Printf("Backup Format Version:\t%d\n", status.Version)
 
 	d.Println()
-	d.Printf("Expiration:\t%s\n", status.Expiration.Time)
+	// "<n/a>" output should only be applicable for backups that failed validation
+	if status.StartTimestamp.Time.IsZero() {
+		d.Printf("Started:\t%s\n", "<n/a>")
+	} else {
+		d.Printf("Started:\t%s\n", status.StartTimestamp.Time)
+	}
+	if status.CompletionTimestamp.Time.IsZero() {
+		d.Printf("Completed:\t%s\n", "<n/a>")
+	} else {
+		d.Printf("Completed:\t%s\n", status.CompletionTimestamp.Time)
+	}
 
 	d.Println()
+	d.Printf("Expiration:\t%s\n", status.Expiration.Time)
+	d.Println()
+
 	d.Printf("Validation errors:")
 	if len(status.ValidationErrors) == 0 {
 		d.Printf("\t<none>\n")
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -337,6 +337,7 @@ func (controller *backupController) getValidationErrors(itm *api.Backup) []strin
 func (controller *backupController) runBackup(backup *api.Backup, bucket string) error {
 	log := controller.logger.WithField("backup", kubeutil.NamespaceAndName(backup))
 	log.Info("Starting backup")
+	backup.Status.StartTimestamp.Time = controller.clock.Now()
 
 	logFile, err := ioutil.TempFile("", "")
 	if err != nil {
@@ -391,6 +392,11 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
 
 	backupScheduleName := backup.GetLabels()["ark-schedule"]
 	controller.metrics.SetBackupTarballSizeBytesGauge(backupScheduleName, backupSizeBytes)
+
+	backup.Status.CompletionTimestamp.Time = controller.clock.Now()
+	backupDuration := backup.Status.CompletionTimestamp.Time.Sub(backup.Status.StartTimestamp.Time)
+	backupDurationSeconds := float64(backupDuration / time.Second)
+	controller.metrics.RegisterBackupDuration(backupScheduleName, backupDurationSeconds)
 
 	log.Info("Backup completed")
 
--- a/pkg/controller/backup_controller_test.go
+++ b/pkg/controller/backup_controller_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/clock"
 	core "k8s.io/client-go/testing"
@@ -175,12 +176,14 @@ func TestProcessBackup(t *testing.T) {
 
 			c.clock = clock.NewFakeClock(clockTime)
 
-			var expiration time.Time
+			var expiration, startTime time.Time
 
 			if test.backup != nil {
 				// add directly to the informer's store so the lister can function and so we don't have to
 				// start the shared informers.
 				sharedInformers.Ark().V1().Backups().Informer().GetStore().Add(test.backup.Backup)
+
+				startTime = c.clock.Now()
 
 				if test.backup.Spec.TTL.Duration > 0 {
 					expiration = c.clock.Now().Add(test.backup.Spec.TTL.Duration)
@@ -194,6 +197,7 @@ func TestProcessBackup(t *testing.T) {
 				backup.Spec.SnapshotVolumes = test.backup.Spec.SnapshotVolumes
 				backup.Status.Phase = v1.BackupPhaseInProgress
 				backup.Status.Expiration.Time = expiration
+				backup.Status.StartTimestamp.Time = startTime
 				backup.Status.Version = 1
 				backupper.On("Backup", backup, mock.Anything, mock.Anything, mock.Anything).Return(nil)
 
@@ -231,6 +235,21 @@ func TestProcessBackup(t *testing.T) {
 				res.Status.Expiration.Time = expiration
 				res.Status.Phase = v1.BackupPhase(phase)
 
+				// If there's an error, it's mostly likely that the key wasn't found
+				// which is fine since not all patches will have them.
+				completionString, err := collections.GetString(patchMap, "status.completionTimestamp")
+				if err == nil {
+					completionTime, err := time.Parse(time.RFC3339Nano, completionString)
+					require.NoError(t, err, "unexpected completionTimestamp parsing error %v", err)
+					res.Status.CompletionTimestamp.Time = completionTime
+				}
+				startString, err := collections.GetString(patchMap, "status.startTimestamp")
+				if err == nil {
+					startTime, err := time.Parse(time.RFC3339Nano, startString)
+					require.NoError(t, err, "unexpected startTimestamp parsing error %v", err)
+					res.Status.StartTimestamp.Time = startTime
+				}
+
 				return true, res, nil
 			})
 
@@ -254,9 +273,11 @@ func TestProcessBackup(t *testing.T) {
 
 			// structs and func for decoding patch content
 			type StatusPatch struct {
-				Expiration time.Time      `json:"expiration"`
-				Version    int            `json:"version"`
-				Phase      v1.BackupPhase `json:"phase"`
+				Expiration          time.Time      `json:"expiration"`
+				Version             int            `json:"version"`
+				Phase               v1.BackupPhase `json:"phase"`
+				StartTimestamp      metav1.Time    `json:"startTimestamp"`
+				CompletionTimestamp metav1.Time    `json:"completionTimestamp"`
 			}
 
 			type Patch struct {
@@ -281,13 +302,14 @@ func TestProcessBackup(t *testing.T) {
 
 			arktest.ValidatePatch(t, actions[0], expected, decode)
 
-			// validate Patch call 2 (setting phase)
+			// validate Patch call 2 (setting phase, startTimestamp, completionTimestamp)
 			expected = Patch{
 				Status: StatusPatch{
-					Phase: v1.BackupPhaseCompleted,
+					Phase:               v1.BackupPhaseCompleted,
+					StartTimestamp:      metav1.Time{Time: c.clock.Now()},
+					CompletionTimestamp: metav1.Time{Time: c.clock.Now()},
 				},
 			}
-
 			arktest.ValidatePatch(t, actions[1], expected, decode)
 		})
 	}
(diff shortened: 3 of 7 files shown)


nrb:
CI's currently failing due to [a known issue](https://github.com/stretchr/testify/issues/502) in Testify, stemming from [a time package behavior](https://github.com/golang/go/issues/19502). In summary, marshalling and unmarshalling times in Go drop the monotonic clock, and alter the `time.Location` property. This affects Ark because the `startTimestamp` and `completionTimestamp` values will be updated via HTTP patch calls with JSON payloads.

We can fix this with the [`time.Equal` function](https://golang.org/pkg/time/#Time.Equal) (which the Go docs recommend), but will need to do a bit of extra conversion from the `interface{}` that `ValidatePatch` currently compares, and probably write a helper to hook into the `*testing.T` facilities. We could also use [apimachinery's DeepEqual](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/equality/semantic.go#L40) function, though it's using the advised-against `==` comparison method.

@ncdc @skriss Any preferences on approach here? Is this something we need to be aware of in general for restores, given that all times going through Ark will encounter this issue?

nrb:
@ashish-amarnath Since you've been working on metrics code, I wanted to let you know I'm grabbing the backup time/duration ones in this PR.

ashish-amarnath:
@nrb I see you are only capturing the backup duration for now. Feel free to get a hold of me either here or on the #ark-dr slack channel if you need to consult on the metrics package.

ncdc:
> We could also use apimachinery's DeepEqual function, though it's using the advised-against == comparison method.

`UTC()` actually ends up stripping the monotonic clock field, so `==` works in this case. I vote we go this route for now.

pkg/metrics/metrics.go
```diff
@@ -31,10 +31,17 @@ const (
 	backupAttemptCount          = "backup_attempt_total"
 	backupSuccessCount          = "backup_success_total"
 	backupFailureCount          = "backup_failure_total"
+	backupSecondsCount          = "backup_seconds_total"
```
ashish-amarnath: Suggest re-naming to this `backupDurationSeconds = "backup_duration_seconds"`.
The convention that I've seen is I count 'events' and then report them as the total number of events, seen in the look back time window.
In this case, what we are capturing is how long it took to run the backup which is not an event. It is kinda similar to the backup size.

pkg/metrics/metrics.go
```diff
@@ -31,10 +31,17 @@ const (
...
 
 	scheduleLabel = "schedule"
 )
 
+// SecondsInMinute returns the number of seconds in a minute.
+// This is mostly a helper to create Prometheus histogram buckets cleanly.
+func SecondsInMinute(minutes uint64) float64 {
```
ashish-amarnath: Not sure you need this method. Suggest creating a const like `secondsInMinute = 60` and 
```
Buckets: []float64{
	SecondsInMinute(1 * secondsInMinute),
	SecondsInMinute(2 * secondsInMinute),
	SecondsInMinute(3 * secondsInMinute),
	SecondsInMinute(4 * secondsInMinute),
	SecondsInMinute(5 * secondsInMinute),
	SecondsInMinute(6 * secondsInMinute),
	SecondsInMinute(7 * secondsInMinute),
	SecondsInMinute(8 * secondsInMinute),
	SecondsInMinute(9 * secondsInMinute),
	SecondsInMinute(10 * secondsInMinute),
},
```
WDYT?

pkg/metrics/metrics.go
```diff
@@ -71,41 +78,69 @@ func NewServerMetrics() *ServerMetrics {
 				},
 				[]string{scheduleLabel},
 			),
+			backupSecondsCount: prometheus.NewHistogramVec(
+				prometheus.HistogramOpts{
+					Namespace: metricNamespace,
+					Name:      backupSecondsCount,
+					Help:      "Total seconds taken for backups",
```
ashish-amarnath: Suggest changing this to 
```
Name: backupDurationSeconds,
Help: "Time taken to complete backup, in seconds",
```

pkg/controller/backup_controller_test.go
```diff
@@ -231,6 +235,19 @@ func TestProcessBackup(t *testing.T) {
 				res.Status.Expiration.Time = expiration
 				res.Status.Phase = v1.BackupPhase(phase)
 
+				// We don't care about the value of the timestamps here, just whether or
```
ncdc: Is it not possible to use the value from the patch?

pkg/metrics/metrics.go
ncdc: +1

pkg/controller/backup_controller_test.go
nrb: It is - I can extract it and parse it to make the test more accurate. I was taking a bit of a shortcut here to avoid that work, but I'll update it.

pkg/metrics/metrics.go
nrb: Good suggestion, I'll go with a `float64` constant and use it directly, without a function call.

pkg/metrics/metrics.go
```diff
@@ -140,7 +136,7 @@ func (m *ServerMetrics) RegisterBackupFailed(backupSchedule string) {
 
 // RegisterBackupSeconds records the number of seconds a backup took.
 func (m *ServerMetrics) RegisterBackupSeconds(backupSchedule string, seconds float64) {
```
ncdc: Call this RegisterBackupDuration?

pkg/controller/backup_controller_test.go
```diff
@@ -231,6 +235,23 @@ func TestProcessBackup(t *testing.T) {
...
+				// not they're present in the patchMap.
+				// If there's an error, it's mostly likely that the key wasn't found
```
ashish-amarnath: nit: `// If there's an error, it's most likely that the key wasn't found`

pkg/metrics/metrics.go
```diff
@@ -71,41 +74,69 @@ func NewServerMetrics() *ServerMetrics {
...
-// RegisterBackupFailed records a failed backup
+// RegisterBackupFailed records a failed backup.
 func (m *ServerMetrics) RegisterBackupFailed(backupSchedule string) {
 	if c, ok := m.metrics[backupFailureCount].(*prometheus.CounterVec); ok {
 		c.WithLabelValues(backupSchedule).Inc()
 	}
 }
+
+// RegisterBackupSeconds records the number of seconds a backup took.
+func (m *ServerMetrics) RegisterBackupSeconds(backupSchedule string, seconds float64) {
```
ashish-amarnath: suggest naming this method `RegisterBackupDuration`.

pkg/controller/backup_controller_test.go
nrb: Fixed

pkg/metrics/metrics.go
nrb: Done

pkg/cmd/util/output/backup_describer.go
```diff
@@ -163,6 +163,18 @@ func DescribeBackupStatus(d *Describer, status v1.BackupStatus) {
 
 	d.Println()
 	d.Printf("Expiration:\t%s\n", status.Expiration.Time)
+	d.Println()
+	// "<n/a>" output should only be applicable for backups that failed validation
+	if status.StartTimestamp.Time.IsZero() {
+		d.Printf("Start Time:\t%s\n", "<n/a>")
```
ncdc: Let's call this `Started`?

pkg/cmd/util/output/backup_describer.go
```diff
@@ -163,6 +163,18 @@ func DescribeBackupStatus(d *Describer, status v1.BackupStatus) {
...
+	} else {
+		d.Printf("Start Time:\t%s\n", status.StartTimestamp.Time)
+	}
+	if status.CompletionTimestamp.Time.IsZero() {
+		d.Printf("Completion Time:\t%s\n", "<n/a>")
```
ncdc: Let's call this `Completed`?

pkg/cmd/util/output/backup_describer.go
```diff
@@ -163,6 +163,18 @@ func DescribeBackupStatus(d *Describer, status v1.BackupStatus) {
 
 	d.Println()
 	d.Printf("Expiration:\t%s\n", status.Expiration.Time)
+	d.Println()
+	// "<n/a>" output should only be applicable for backups that failed validation
```
ncdc: I would put started & completed above expiration

ncdc:
After reading up a bit more on Prometheus histograms, this PR divides the backup duration into 10 buckets, from 1 to 10 seconds, plus an additional `+Inf` bucket for anything over 10 seconds. The histogram is essentially a group of counters, and it's cumulative, so if we execute 1 backup and it takes 8 seconds to run, the buckets for 8, 9, 10, and +Inf would all show `1` (remember, it's a count, not an actual duration).

Unfortunately, the buckets have to be hard-coded up front when we create the histogram. I'm wondering if 1-10 seconds is reasonable. I'm thinking we might need a greater range, like 5s, 30s, 1m, 5m, 15m, 30m, etc.

I'm definitely looking for guidance from people more experienced with metrics gathering than I am 😄

nrb:
> this PR divides the backup duration into 10 buckets, from 1 to 10 seconds, plus an additional +Inf bucket for anything over 10 seconds. 

The buckets are definitely 1-10 minutes. https://github.com/prometheus/client_golang/blob/master/prometheus/histogram.go#L54 shows the default values, and states that the values are based on seconds.

That said I'm not confident that this is a good distribution. It may take some testing to tune these.

ncdc:
My mistake - 1-10 minutes is right. Thanks for the correction!

pkg/metrics/metrics.go
```diff
@@ -71,41 +74,69 @@ func NewServerMetrics() *ServerMetrics {
...
+						1 * secondsInMinute,
+						2 * secondsInMinute,
+						3 * secondsInMinute,
+						4 * secondsInMinute,
+						5 * secondsInMinute,
+						6 * secondsInMinute,
+						7 * secondsInMinute,
+						8 * secondsInMinute,
+						9 * secondsInMinute,
+						10 * secondsInMinute,
```
mattmoyer: This will mean any backup longer than 10 minutes is counted as "infinite". Is that reasonable?

It might be useful to use [exponential buckets](https://godoc.org/github.com/prometheus/client_golang/prometheus#ExponentialBuckets) to get more resolution at the higher end of this range.

stevesloka:
The buckets can be tricky to define since they are typically specific to an environment. I'd recommend sane defaults to start, then allow users to define those buckets if they need to outside the defaults.

nrb:
@stevesloka How would you recommend letting users define them? For what it's worth, we're considering getting rid of the Ark `Config` CRD.

ncdc:
Anything in config that doesn't move to a backup target will most likely become a flag to `ark server`

stevesloka:
I'm not sure specifically in Ark, but should be something on the server since you can only configure it once.

pkg/controller/backup_controller.go
```diff
@@ -392,6 +393,11 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
 	backupScheduleName := backup.GetLabels()["ark-schedule"]
 	controller.metrics.SetBackupTarballSizeBytesGauge(backupScheduleName, backupSizeBytes)
 
+	backup.Status.CompletionTimestamp.Time = controller.clock.Now()
+	backupDuration := backup.Status.CompletionTimestamp.Time.Sub(backup.Status.StartTimestamp.Time)
+	backupDurationSeconds := float64(backupDuration / time.Second)
```
skriss: why's the ` / time.Second` needed? (Not saying it's wrong, just wondering)

nrb: Good catch - that should be multiplied to get the seconds, since Duration is nanoseconds.

nrb: Actually I'm confusing myself here - this is correct. We need to divide to get the seconds out of a Duration.

Multiplication would be seconds to nanoseconds.

nrb: To elaborate a bit more: the histogram we're building uses seconds as it's unit, which is why we're going from ns to seconds.

pkg/metrics/metrics.go
```diff
@@ -71,41 +72,58 @@ func NewServerMetrics() *ServerMetrics {
 				},
 				[]string{scheduleLabel},
 			),
+			backupDurationSeconds: prometheus.NewHistogramVec(
+				prometheus.HistogramOpts{
+					Namespace: metricNamespace,
+					Name:      backupDurationSeconds,
+					Help:      "Time taken to complete backup, in seconds",
+					// Use 10 exponential buckets starting at 30s
+					Buckets: prometheus.ExponentialBuckets(30.0, 3, 10),
```
ncdc: This would create the following buckets, correct?

1. 30s
1. 1m30s
1. 4m30s
1. 13m30s
1. 40m30s
1. 2h1m30s
1. 6h4m30s
1. 18h13m30s
1. 54h40m30s
1. 164h1m30s

pkg/util/test/comparisons.go
```diff
@@ -77,5 +82,30 @@ func ValidatePatch(t *testing.T, action core.Action, expected interface{}, decod
...
+
+// AssertDeepEqual asserts the semantic equality of objects.
+// This function exists in order to make sure time.Time and metav1.Time objects
+// can be compared correctly. See https://github.com/stretchr/testify/issues/502.
+func AssertDeepEqual(t *testing.T, expected, actual interface{}) bool {
+	// By default, the equality.Semantic object doesn't have a function for comparing time.Times
+	err := equality.Semantic.AddFunc(TimesAreEqual)
+	if err != nil {
+		// Programmer error, the service should die.
+		panic(errors.Wrap(err, "Could not register equality function"))
```
ncdc: I think `t.Fatal*` would be more appropriate?

pkg/metrics/metrics.go
nrb: Yeah - I think 9 & 10 might be too high, perhaps even 8.

I think Linear buckets of 1 minute may leave out the high end if we only have 10, but I don't know if there's any sort of penalty for making many buckets that go unused.

pkg/controller/backup_controller.go
skriss: got it, ok. duration math is weird :)

ncdc:
LGTM

PR #566, opened 2018-06-21
Minio should support read-only file systems

runyontr:
When running in OpenShift or with read-only file systems, the minio deployment does not have access to `/` to write the config object.

This creates a volume for read/writing of the config object, and leverages that config in CLI calls.


ncdc:
Thanks for the PR. We'll try to test it tomorrow.

ncdc:
LGTM. Tested locally & it works. Thanks!

PR #570, opened 2018-06-21
Restic controller improvements

skriss:
- various logging improvements - helpful fields, adjusting levels, removing noise
- for the PVB/PVR controllers, filter out items in the informer event handlers *before* adding them to the queue, to reduce noise
- refactor/simplify the code in the PVR controller that determines which PVRs are ready to process (based on the status of them and their associated pod)

### DIFF
--- a/pkg/controller/pod_volume_backup_controller.go
+++ b/pkg/controller/pod_volume_backup_controller.go
@@ -86,17 +86,36 @@ func NewPodVolumeBackupController(
 
 	podVolumeBackupInformer.Informer().AddEventHandler(
 		cache.ResourceEventHandlerFuncs{
-			AddFunc:    c.enqueue,
-			UpdateFunc: func(_, obj interface{}) { c.enqueue(obj) },
+			AddFunc:    c.pvbHandler,
+			UpdateFunc: func(_, obj interface{}) { c.pvbHandler(obj) },
 		},
 	)
 
 	return c
 }
 
+func (c *podVolumeBackupController) pvbHandler(obj interface{}) {
+	req := obj.(*arkv1api.PodVolumeBackup)
+
+	// only enqueue items for this node
+	if req.Spec.Node != c.nodeName {
+		return
+	}
+
+	log := loggerForPodVolumeBackup(c.logger, req)
+
+	if req.Status.Phase != "" && req.Status.Phase != arkv1api.PodVolumeBackupPhaseNew {
+		log.Debug("Backup is not new, not enqueuing")
+		return
+	}
+
+	log.Debug("Enqueueing")
+	c.enqueue(obj)
+}
+
 func (c *podVolumeBackupController) processQueueItem(key string) error {
 	log := c.logger.WithField("key", key)
-	log.Debug("Running processItem")
+	log.Debug("Running processQueueItem")
 
 	ns, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controller/pod_volume_backup_controller.go
```diff
@@ -120,22 +139,29 @@ func (c *podVolumeBackupController) processQueueItem(key string) error {
...
-func (c *podVolumeBackupController) processBackup(req *arkv1api.PodVolumeBackup) error {
-	log := c.logger.WithFields(logrus.Fields{
+func loggerForPodVolumeBackup(baseLogger logrus.FieldLogger, req *arkv1api.PodVolumeBackup) logrus.FieldLogger {
+	log := baseLogger.WithFields(logrus.Fields{
 		"namespace": req.Namespace,
 		"name":      req.Name,
 	})
 
+	if len(req.OwnerReferences) == 1 {
+		log = log.WithField("arkBackup", req.OwnerReferences[0].Name)
```
ncdc: We've used `backup` as the key with `namespace/name` as the value before - let's be consistent?

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -111,98 +111,84 @@ func NewPodVolumeRestoreController(
...
-}
-
-func (c *podVolumeRestoreController) podHandler(obj interface{}) {
-	pod := obj.(*corev1api.Pod)
-	log := c.logger.WithField("key", kube.NamespaceAndName(pod))
-
-	for _, pvr := range pvrsToEnqueueForPod(pod, c.podVolumeRestoreLister, c.nodeName, log) {
-		c.enqueue(pvr)
+	pod, err := c.podLister.Pods(pvr.Spec.Pod.Namespace).Get(pvr.Spec.Pod.Name)
+	if err != nil {
```
ncdc: I think we should log not-found errors at debug level, and all other errors at error level. WDYT?

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -111,98 +111,84 @@ func NewPodVolumeRestoreController(
...
+func (c *podVolumeRestoreController) podHandler(obj interface{}) {
+	pod := obj.(*corev1api.Pod)
+	log := c.logger.WithField("key", kube.NamespaceAndName(pod))
 
-func pvrsToEnqueueForPod(pod *corev1api.Pod, pvrLister listers.PodVolumeRestoreLister, nodeName string, log logrus.FieldLogger) []*arkv1api.PodVolumeRestore {
-	if !shouldProcessPod(pod, nodeName, log) {
-		return nil
+	// the pod should always be for this node since the podInformer is filtered
+	// based on node, so this is just a failsafe.
+	if pod.Spec.NodeName != c.nodeName {
```
ncdc: Use `isPodOnNode`?

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -111,98 +111,84 @@ func NewPodVolumeRestoreController(
...
+		return
 	}
 
-	return res
-}
-
-func shouldEnqueuePVR(pvr *arkv1api.PodVolumeRestore, podLister corev1listers.PodLister, nodeName string, log logrus.FieldLogger) bool {
-	if !shouldProcessPVR(pvr, log) {
-		return false
+	if !isResticInitContainerRunning(pod) {
```
ncdc: Move this check up before the lister call (it's less expensive)?

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -241,12 +227,24 @@ func (c *podVolumeRestoreController) processQueueItem(key string) error {
...
-func (c *podVolumeRestoreController) processRestore(req *arkv1api.PodVolumeRestore) error {
-	log := c.logger.WithFields(logrus.Fields{
+func loggerForPodVolumeRestore(baseLogger logrus.FieldLogger, req *arkv1api.PodVolumeRestore) logrus.FieldLogger {
+	log := baseLogger.WithFields(logrus.Fields{
 		"namespace": req.Namespace,
 		"name":      req.Name,
 	})
 
+	if len(req.OwnerReferences) == 1 {
+		log = log.WithField("arkRestore", req.OwnerReferences[0].Name)
```
ncdc: Use `restore=namespace/name`?

skriss:
addressed comments & added a buncha unit tests.

skriss:
Will squash once you get a chance to look at updates.

pkg/controller/pod_volume_restore_controller_test.go
```diff
@@ -0,0 +1,572 @@
...
+			shouldEnqueue: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			var (
+				client      = kubefake.NewSimpleClientset()
+				informers   = kubeinformers.NewSharedInformerFactory(client, 0)
+				podInformer = informers.Core().V1().Pods()
```
ncdc: You could construct the podInformer by hand instead of having the factory do it. Then you don't need the fake kube clientset.

pkg/controller/pod_volume_restore_controller_test.go
```diff
@@ -0,0 +1,572 @@
...
+						{
+							State: corev1api.ContainerState{
+								Running: &corev1api.ContainerStateRunning{StartedAt: metav1.Time{Time: time.Now()}},
+							},
+						},
+					},
+				},
+			},
+			expected: true,
+		},
```
ncdc: Need a test case for first initContainer is our restic, but container state != running?

ncdc:
lgtm

skriss:
squashed.

PR #571, opened 2018-06-22
restic: if S3, get bucket's region up-front

### LINKED ISSUE #567: Restic integration errors out
I'm trying to get restic up and running but it keeps failing. My ark config looks like this:
```
backupStorageProvider:
  name: aws
  bucket: stuff-dev-heptio-ark
  config:
    region: eu-west-1
  resticLocation: stuff-dev-heptio-ark-restic
```
(btw _please_ provide yaml files instead of fancy kubectl/jq commands to configure the required settings. While those are super-impressive, yaml is easier to read and understand. I have no idea what most of the commands in the restic documentation do, which makes it very hard to follow.

Here's the error I'm getting from the ark server:
```
time="2018-06-21T21:02:16Z" level=error msg="backup failed" error="error running command, stderr=Fatal: create repository at s3:s3.amazonaws.com/stuff-dev-heptio-ark-restic/nginx-example failed: client.BucketExists: The specified bucket does not exist.\n\n: exit status 1" key=heptio-ark/nginx-example logSource="pkg/controller/backup_controller.go:280"
```
Running ark without restic integration works, and the iam permissions are correct. The bucket exists, but maybe the restic integration expects a certain bucket location? AFAIK there's no `resticLocationRegion` setting or similar.

skriss:

Fixes #567 

If using restic with S3, fetch the bucket's region up-front and put it in the endpoint URL to use for restic. This makes all restic requests faster and also doesn't require any additional permission to determine the right S3 endpoint (which the default restic method does).

### DIFF
--- a/pkg/cloudprovider/aws/helpers.go
+++ b/pkg/cloudprovider/aws/helpers.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2018 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package aws
+
+import (
+	"context"
+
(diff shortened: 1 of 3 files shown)


skriss:
Got this tested on a new S3 bucket in eu-central-1 and worked for me.

nrb:
I assume this code needs to be in the restic daemonset as well as the ark server pod to get the full functionality? Were you just building your own images for that?

skriss:
this is only needed in the ark server pod - the region is determined during repo initialization, and stored in the restic repo CR's spec. the daemonset pods just pull it from there, whether it has the region as part of it or not

PR #606, opened 2018-06-25
Automatically manage restic repos

### LINKED ISSUE #577: Streamline Ark Restic setup
Following up on a [discussion](https://groups.google.com/forum/#!topic/heptio-ark/6JljWGzTc0c) this week regarding the security and user experience of the new Restic backup process, we are planning the following changes before releasing 0.9:

* Ark will automatically manage Restic repositories without requiring any user involvement (other than configuring the bucket for Restic in the Ark config)
* You won't have to run ark restic repo init
* The Restic encryption key secrets will move from one per namespace and instead all be contained in the heptio-ark namespace (or a custom one if you've changed it)
* When creating a backup, Ark will include the relevant Restic encryption keys in their own separate location in the tarball (distinct from the Kubernetes resources)
* When restoring, Ark will use the encryption keys present in the Ark backup tarball
* We will not support changing the Restic encryption keys; instead, when we introduce full backup encryption, we'll use a master key to encrypt the backup tarball (more details on this to come as we spec out encryption)

skriss:
Fixes #577 

Builds on #570, ignore first commit.

Replaces `ark restic repo init` with automatic repository initialization, using a single static non-secret restic key.


pkg/restic/repository_ensurer.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (r *repositoryEnsurer) EnsureRepo(ctx context.Context, namespace, name string) (*arkv1api.ResticRepository, error) {
+	if repo, err := r.repoLister.ResticRepositories(namespace).Get(name); err != nil && !apierrors.IsNotFound(err) {
+		return nil, errors.WithStack(err)
+	} else if err == nil {
+		if repo.Status.Phase != arkv1api.ResticRepositoryPhaseReady {
+			return nil, errors.New("restic repository is not ready")
+		}
+		return repo, nil
+	}
```
nrb: Might add a comment here that the rest of the function is for the `apierrors.IsNotFound(err) == true` case.

skriss: added.

pkg/restic/restorer.go
```diff
@@ -85,7 +84,7 @@ func (r *restorer) RestorePodVolumes(restore *arkv1api.Restore, pod *corev1api.P
 		return nil
 	}
 
-	repo, err := getReadyRepo(r.repoLister, restore.Namespace, pod.Namespace)
+	repo, err := r.repoEnsurer.EnsureRepo(r.ctx, restore.Namespace, pod.Namespace)
```
nrb: Would we ever end up creating a repo here? I'm thinking that even if we do, the data and buckets are in an invalid state already.

pkg/restic/backupper.go
```diff
@@ -84,39 +82,14 @@ func resultsKey(ns, name string) string {
...
 	volumesToBackup := GetVolumesToBackup(pod)
 	if len(volumesToBackup) == 0 {
 		return nil, nil
 	}
 
-	repo, err := getReadyRepo(b.repoLister, backup.Namespace, pod.Namespace)
+	repo, err := b.repoEnsurer.EnsureRepo(b.ctx, backup.Namespace, pod.Namespace)
 	if err != nil {
 		return nil, []error{err}
 	}
```
nrb: This function is executed in the daemonset, correct?

I ran into an issue yesterday that was partially my fault, but probably warrants a log message. If the daemonset exists, but no pods in it are ready, the restic backup process will still proceed and then hang waiting for the daemonset pod to finish working. In my testing, it waited for over 5 minutes and couldn't be stopped with ctrl-c, only a kill -9.

I can probably replicate if needed to walk through this.

pkg/restic/repository_manager.go
```diff
@@ -141,72 +142,55 @@ func (rm *repositoryManager) NewRestorer(ctx context.Context, restore *arkv1api.
...
 }
 
-func (rm *repositoryManager) Forget(snapshot SnapshotIdentifier) error {
-	repo, err := getReadyRepo(rm.repoLister, rm.namespace, snapshot.Repo)
+func (rm *repositoryManager) Forget(ctx context.Context, snapshot SnapshotIdentifier) error {
+	if !cache.WaitForCacheSync(ctx.Done(), rm.repoInformerSynced) {
+		return errors.New("timed out waiting for cache to sync")
+	}
+
+	repo, err := rm.repoEnsurer.EnsureRepo(ctx, rm.namespace, snapshot.Repo)
```
ncdc: This has the potential to create the restic repo. We shouldn't be doing that as part of deletions/GC.

pkg/restic/restorer.go
ncdc: This has the potential to create the restic repo. We shouldn't be doing that as part of a restore.

pkg/restic/repository_ensurer.go
```diff
@@ -0,0 +1,117 @@
...
+// repositoryEnsurer ensures that Ark restic repositories are created and ready.
+type repositoryEnsurer struct {
+	repoLister arkv1listers.ResticRepositoryLister
+	repoClient arkv1client.ResticRepositoriesGetter
+
+	readyChansLock sync.Mutex
+	readyChans     map[string]chan *arkv1api.ResticRepository
+}
+
+func newRepositoryEnsurer(repoInformer arkv1informers.ResticRepositoryInformer, repoClient arkv1client.ResticRepositoriesGetter) *repositoryEnsurer {
```
ncdc: I would recommend changing the logic so we create a new one of these per `*restic.backupper`. Then we don't need a lock or multiple channels and can instead just have a single channel.

I'm even tempted to say we can do away with the informer+channel. Move the `initializeRepo` functionality from `*resticRepositoryController` directly into the `*restic.Backupper` or this `repositoryEnsurer`. Then it's a lot simpler to implement `EnsureRepo`:
1. Try to get the repo.
1. If it exists & is ready, done.
1. If it exists & isn't ready, wait (?) or call `Check()`. Return result of wait/check
1. If it doesn't exist, try to create it
1. If create succeeds, run the logic in `initializeRepo`

pkg/restic/restorer.go
skriss: How else would you handle the case where a user installs Ark in a fresh cluster and immediately wants to restore an existing backup with restic snapshots?

pkg/restic/repository_manager.go
skriss: same question re: spinning up a new cluster, then having GC process backups that were synced in from an existing bucket.

pkg/restic/repository_ensurer.go
skriss: If you want to avoid having multiple channels, we'd need to have one of these per call to `BackupPodVolumes` in the backupper, since a backup can have >1 namespace/repo. Not sure exactly what that would look like, since I don't think we'd want to be creating a new informer each time, but we can't be adding/removing event handlers while an existing one is running.

I'm not sure I like the idea of moving the repo init logic into the backupper or ensurer, but I think we need to talk it through in person.

Why don't you like what I did?

pkg/restic/restorer.go
ncdc: My bad, I wasn't thinking about the restore half of the equation (that's what happens when I try to wrap up a code review at the end of the day after a long meeting!).

pkg/restic/repository_manager.go
ncdc: Yup, my bad

pkg/restic/repository_ensurer.go
ncdc: Let me take a look at it again with fresh eyes. I had to draw out the flow and the relationship between the components (ensurer vs repo controller). I was trying to find a way to have a single component responsible for the creation and initialization of the repo, instead of having that split across 2 (the ensurer that creates the ResticRepository, and the controller that initializes it in object storage and sets the phase to Ready).

ncdc:
@skriss does this look accurate for the EnsureRepo flow?

EnsureRepo flow

skriss:
Looks right to me

ncdc:
Backup flow: backup flow

ncdc:
Delete snapshot flow (from backup deletion controller): delete snapshot flow

ncdc:
Daemonset pod backup flow: daemonset pod backup flow

ncdc:
Restore overview: restore overview

ncdc:
DaemonSet pod restore flow: daemonset pod restore flow

skriss:
Addressed items discussed; will do squashing post-review.

ncdc:
LGTM

skriss:
squashed.

PR #607, opened 2018-06-25
Restore metrics

nrb:
This builds on the work in #564 with the latest commit.


pkg/cmd/util/output/backup_describer.go
```diff
@@ -163,6 +164,18 @@ func DescribeBackupStatus(d *Describer, status v1.BackupStatus) {
 
 	d.Println()
 	d.Printf("Expiration:\t%s\n", status.Expiration.Time)
+	d.Println()
+	// "<never>" output should only be applicable for backups that failed validation
+	if status.StartTimestamp.Time.Equal(time.Time{}) {
```
skriss: `status.StartTimestamp.Time.IsZero()`

pkg/cmd/util/output/backup_describer.go
```diff
@@ -163,6 +164,18 @@ func DescribeBackupStatus(d *Describer, status v1.BackupStatus) {
...
+		d.Printf("Start Time:\t%s\n", "<never>")
```
skriss: maybe `<n/a>` instead of `<never>`?

pkg/cmd/util/output/backup_describer.go
nrb: Thanks. btw only the latest commit in this PR is new - the first one is from #564.

pkg/metrics/metrics.go
```diff
@@ -32,8 +32,12 @@ const (
 	backupSuccessCount          = "backup_success_total"
 	backupFailureCount          = "backup_failure_total"
 	backupDurationSeconds       = "backup_duration_seconds"
+	restoreAttemptCount         = "restore_attempt_total"
+	restoreSuccessCount         = "restore_sucess_total"
+	restoreFailureCount         = "restore_failure_count"
 
-	scheduleLabel = "schedule"
+	scheduleLabel   = "schedule"
+	backupNameLabel = "backupName"
```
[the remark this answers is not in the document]
ncdc: Although I don't think we want to group restore metrics by backup. That means we'll get unique restore attempt/success/failure metrics per backup. It might be better to group by schedule, like we do for backups.

pkg/controller/restore_controller.go
```diff
@@ -280,6 +284,11 @@ func (controller *restoreController) processRestore(key string) error {
 	for _, e := range restoreErrors.Namespaces {
 		restore.Status.Errors += len(e)
 	}
+	if restore.Status.Errors > 0 {
```
ncdc: xref https://github.com/heptio/ark/pull/531#discussion_r195196289 cc @ashish-amarnath @rosskukulinski - this is 1 place where we currently have phases `New`, `InProgress`, `FailedValidation`, and `Completed`. Once the restore is `Completed`, it has 0+ warnings and 0+ errors. How do you think we should record this information as metrics?

pkg/metrics/metrics.go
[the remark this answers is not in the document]
nrb: I was wondering about that - do we have the schedule information available on restore?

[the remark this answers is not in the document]
ncdc: It's available as a label in the backup (which we have).

pkg/metrics/metrics.go
```diff
@@ -32,8 +32,12 @@ const (
 	backupSuccessCount          = "backup_success_total"
 	backupFailureCount          = "backup_failure_total"
 	backupDurationSeconds       = "backup_duration_seconds"
+	restoreAttemptCount         = "restore_attempt_total"
```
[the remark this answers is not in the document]
nrb: Agreed; I was maintaining parity with what was already present.

pkg/controller/restore_controller.go
nrb: On the backup side, when they fail validation I don't think they even get counted as starting if they failed validation. The attempts are logged at [L277](https://github.com/heptio/ark/blob/master/pkg/controller/backup_controller.go#L277), but the validation failure returns at [L267](https://github.com/heptio/ark/blob/master/pkg/controller/backup_controller.go#L267).

Do we want to register attempts for restores/backups before doing the validation?

ncdc: > Do we want to register attempts for restores/backups before doing the validation?

+1. We may want to add a metric for `backup/restore_validation_failure_total`?

rosskukulinski: We probably should have gauges for each of phases New, InProgress with the current number of backups (and also restores) in those states.  That will help alert users if they have backups that are stuck for some reason.

As for 0+ warnings and 0+ errors, we should increment a counter each (per-schedule, or empty schedule) (plus the counter somewhere for backups completed).  Users won't be able to know _which_ backup/restore had Errors or Warnings, but they will know that it happened and can then go look at Ark server logs (or describe backup/restore).

pkg/controller/restore_controller.go
```diff
@@ -247,6 +251,14 @@ func (controller *restoreController) processRestore(key string) error {
 		}
 	}
 
+	// TODO: What do we do with the error here? It'll get handled as a validation error later,
+	// but if we need a schedule name for the restore before that, we're throwing our metrics off.
+	backup, _ := controller.fetchBackup(controller.bucket, restore.Spec.BackupName)
```
nrb: The error is ignored for now as I want to get some input on this.

We're getting the backup itself in 2 functions below here - `getValidationErrors` and `runRestore`. Neither of those returns the backup fetched up to this `processRestore` function.

By calling `fetchBackup` here, we're duplicating one of the validation checks (if the backup even exists). Should we do that here? If the backup doesn't exist, we can't get a schedule name, which could skew the number of attempts towards the `schedule=""` label;  I'm not sure we can do much else in this case.

ncdc: This code is in need of a cleanup, but that'll be a broader effort to move as close to 100% of the defaulting, validation, and processing logic from the controller to the restorer (and the same for backups). In the meantime:

I would recommend that we call `controller.fetchBackup` one time: in `processRestore`, immediately before we call `controller.getValidationErrors`. Let's store the fetched backup in a variable, and pass it - `getValidationErrors(restore, backup)`. Let's also pass it to `runRestore`.

As for the metrics, and potentially skewing the `schedule=""` label if we can't find the backup, that's probably ok, at least for now, and hopefully for the long term too. We _could_ try to parse the name of the backup to see if it looks like it has a timestamp suffix, but let's not worry about that for now.

nrb:
I'm going to leave this PR with the 4 metrics included for now, and we can revisit the gauges and error/warning tracking in another.

pkg/metrics/metrics.go
```diff
@@ -95,6 +101,38 @@ func NewServerMetrics() *ServerMetrics {
...
+					Name:      restoreFailureTotal,
+					Help:      "Total number of failed restores",
+				},
+				[]string{scheduleLabel},
+			),
+			restoreValidationFailedTotal: prometheus.NewCounterVec(
+				prometheus.CounterOpts{
+					Namespace: metricNamespace,
+					Name:      restoreValidationFailedTotal,
+					Help:      "Total number of failed restore validations",
```
ncdc: Total number of restores failing validation?

pkg/controller/restore_controller.go
ncdc: > Maybe we can say backup completed with %d errors and %d warnings and track the error and warning counts inside the backup object?

We do plan to do this, similar to how we record this information for restores, but we haven't done it yet.

> Validation to me reads like there is something wrong with the way the backup object was created.

Restore validation errors include:
- restore spec is missing backup name
- ark couldn't find the backup you want to restore
- you asked to include a non-restorable resource in the restore (nodes, events, backups, restores)
- the requested include/exclude namespaces overlap
- the requested include/exclude resources overlap
- you asked to restore PVs but ark isn't configured with a PV provider

pkg/controller/restore_controller.go
```diff
@@ -280,6 +292,11 @@ func (controller *restoreController) processRestore(key string) error {
 	for _, e := range restoreErrors.Namespaces {
 		restore.Status.Errors += len(e)
 	}
+	if restore.Status.Errors > 0 {
+		controller.metrics.RegisterRestoreFailed(backupScheduleName)
```
ncdc: I'm a bit on the fence about recording the restore as a "failure" if we have at least 1 error. It's possible that 99% of the restore completed ok, and some small portion had an error. What do you all think we should do here?

ncdc: @rosskukulinski @ashish-amarnath

rosskukulinski: Would it maybe be more accurate to say that the restore was "incomplete"?

ncdc: +1, incomplete >>> failure imho

rosskukulinski: What other reasons might a restore completely fail?

ncdc: - couldn't find the requested backup
- error downloading the backup tarball
- error creating temp file for the restore log
- error creating temp file for the restore warnings/errors
- error initializing restore item action plugins
- etc (there are more if you want me to keep listing)

nrb: Looking at the possible phases for restores, we only have 4 - new, failed validation, in progress, or completed. So we don't have a specific phase that indicates complete failure. That may be useful to add.

For the metric, Incomplete makes sense.

ncdc: Yes, we should have Failed for complete failures, and we need to record the reason for the failure in restore.status.

rosskukulinski: I like where this is going.  It's important to separate failures from incomplete.

nrb: Metric name updated. The `restoreFailedTotal` metric can come back when the Failed phase is implemented.

nrb: The more I think about it, the less useful I think this Incomplete metric is. It doesn't match an existing status, so the metric and the CRD will send a slightly different message. The information might be better captured by the error/warnings gauges, without a counter. Thoughts @ncdc, @rossfenrick @ashish-amarnath ?

ncdc: Want to add the `Failed` phase as part of this?

rosskukulinski: @nrb: How/where are these gauges?  I don't think the fundamental behavior of gauges would work well with this type of data.

nrb: @rosskukulinski They're hypothetical at the moment. I'm going to take a stab at implementing a Failed phase and using that for the Failed metric and defer reporting the number of errors/warnings to a later PR.

pkg/controller/restore_controller.go
```diff
@@ -255,8 +259,31 @@ func (c *restoreController) processRestore(key string) error {
...
+	// attempt to get a schedule name whether provided directly or through a backup
+	backupScheduleName := restore.Spec.ScheduleName
+	if backupScheduleName == "" {
+		backup, fetchErr = c.fetchBackup(c.bucket, restore.Spec.BackupName)
+		if backup != nil {
+			backupScheduleName = backup.GetLabels()["ark-schedule"]
+		}
+	}
+
+	// Register attempts before we do validation so we can get better tracking
```
skriss: What does "better tracking" actually mean here?  IOW, what would be the problem if we move the call to `RegisterRestoreAttempt` to below `completeAndValidate`?

pkg/controller/restore_controller.go
```diff
@@ -255,8 +259,31 @@ func (c *restoreController) processRestore(key string) error {
 	// don't modify items in the cache
 	restore = restore.DeepCopy()
 
-	// complete & validate restore
-	if restore.Status.ValidationErrors = c.completeAndValidate(restore); len(restore.Status.ValidationErrors) > 0 {
+	excludedResources := sets.NewString(restore.Spec.ExcludedResources...)
```
skriss: L262-267 were moved into `completeAndValidate` so should be removed from here.

pkg/metrics/metrics.go
```diff
@@ -158,3 +208,31 @@ func (m *ServerMetrics) RegisterBackupDuration(backupSchedule string, seconds fl
...
+
+// RegisterRestoreAttempt records an attempt to restore a backup.
+func (m *ServerMetrics) RegisterRestoreAttempt(backupSchedule string) {
+	if c, ok := m.metrics[restoreAttemptTotal].(*prometheus.CounterVec); ok {
+		c.WithLabelValues(backupSchedule).Inc()
+	}
+}
+
+// RegisterRestoreSuccess records a successful (maybe partial) completion of a restore.
+func (m *ServerMetrics) RegisterRestoreSuccess(backupSchedule string) {
```
skriss: Would `RegisterRestoreCompleted` be a better name here?

pkg/controller/restore_controller.go
```diff
@@ -289,8 +316,19 @@ func (c *restoreController) processRestore(key string) error {
...
 	}
 
+	if restoreFailure != nil {
+		logContext.Debug("restore failed")
+		restore.Status.Phase = api.RestorePhaseFailed
+		restore.Status.FailureReason = restoreFailure.Error()
+		c.metrics.RegisterRestoreFailed(backupScheduleName)
+	}
+
 	logContext.Debug("restore completed")
```
skriss: I think L326-331 would make more sense in an `else`, since we don't want to log "restore completed" right after logging "restore failed" for failed restores, and also then you can remove L328.

pkg/controller/restore_controller.go
```diff
@@ -361,9 +398,10 @@ func (c *restoreController) completeAndValidate(restore *api.Restore) []string {
 		}
 	}
 
-	// validate that we can fetch the source backup
-	if _, err := c.fetchBackup(c.bucket, restore.Spec.BackupName); err != nil {
-		return append(validationErrors, fmt.Sprintf("Error retrieving backup: %v", err))
+	// re-use the validation error we got when doing an earlier lookup to check for
+	// a given backup's schedule name for metrics.
+	if fetchErr != nil {
```
skriss: In the current state of the code, `fetchErr` will **always** be nil if the restore was created with a non-nil `.spec.scheduleName`, essentially bypassing this validation.  I have some potential thoughts on how to simplify the whole flow around fetch/fetchErr but need to understand the answer to my q around "better tracking" first.

pkg/controller/restore_controller.go
```diff
@@ -433,7 +471,7 @@ func (c *restoreController) fetchBackup(bucket, name string) (*api.Backup, error
 	return backup, nil
 }
 
-func (c *restoreController) runRestore(restore *api.Restore, bucket string) (restoreWarnings, restoreErrors api.RestoreResult) {
+func (c *restoreController) runRestore(restore *api.Restore, bucket string, backup *api.Backup) (restoreWarnings, restoreErrors api.RestoreResult, restoreFailure error) {
```
skriss: I don't think the new `backup` param is actually used anywhere

pkg/apis/ark/v1/restore.go
```diff
@@ -103,6 +107,9 @@ type RestoreStatus struct {
 	// Errors is a count of all error messages that were generated during
 	// execution of the restore. The actual errors are stored in object storage.
 	Errors int `json:"errors"`
+
+	// FailureReason is an error that caused the entire restore to fail.
+	FailureReason string `json:"failureReason"`
```
skriss: I understand the motivation for this, but I do worry about the fact that we're adding yet another place to look for restore errors (ark server logs, restore logs, restore warnings/errors file, and now this field).   Undecided right now on whether we need to do something about it for this PR, or if it can hold off 'til we address that holistically.

pkg/controller/restore_controller.go
```diff
@@ -492,6 +535,8 @@ func (c *restoreController) runRestore(restore *api.Restore, bucket string) (res
 	}
 	defer c.pluginManager.CloseRestoreItemActions(restore.Name)
 
+	// Any return statement above this line means a total restore failure
+	// Some failures after this line *may* be a total restore failure
 	logContext.Info("starting restore")
```
skriss: not related directly to this PR but our logging is very redundant, since we're also logging start/complete of restores in the calling function

pkg/controller/restore_controller.go
nrb: Good question - essentially, I wanted to increment attempts before incrementing any other type of restore metric. I believe that's doable after calling into `completeAndValidate`, though, since it would just register an attempt then register a failed validation.

pkg/metrics/metrics.go
nrb: Maybe...it currently mirrors the backup success name, though.

IMO, if we change one we should change both. There's a TODO on L33 to update the strings to match the names, too.

pkg/apis/ark/v1/restore.go
nrb: This is a good point. Per Slack convo, most of the failure cases should be in the restore log today; they'd just be marked as `Completed`. The only one where it would fail and it's not logged is if the upload of the log itself fails.

skriss: Eh, looking at this again I think I'm fine with it.  Right now, this will only be populated if the restore doesn't even start, in which case it won't have a log or results in object storage.

pkg/controller/restore_controller.go
nrb: I think I see the mismatch - I kept L481 in rebasing. I was trying to keep it around and save a fetch. If the backup's not returned by `completeAndValidate`, we'll be doing a second fetch in this case.

I'm fine with leaving it as is for now and not passing a backup in, though it does seem a little inefficient and a duplication of work.

pkg/controller/restore_controller.go
```diff
@@ -289,8 +297,19 @@ func (c *restoreController) processRestore(key string) error {
...
-	restore.Status.Phase = api.RestorePhaseCompleted
+	if restoreFailure != nil {
+		logContext.Debug("restore failed")
+		restore.Status.Phase = api.RestorePhaseFailed
+		restore.Status.FailureReason = restoreFailure.Error()
+		c.metrics.RegisterRestoreFailed(backupScheduleName)
+	} else {
+		logContext.Debug("restore completed")
+		// We got through the restore process without failing validation or restore execution
+		if restore.Status.Phase == api.RestorePhaseInProgress {
```
skriss: i don't think this `if` is necessary

skriss:
otherwise LGTM.

pkg/controller/restore_controller.go
nrb: Done

Issue #586, opened 2018-06-25
Add prioritized resource list for backups


ncdc:
Similar to restores, backups should be able to process resources in a well-defined order.

dymurray:
Is there some documentation around how to prioritize resources on restore?

dymurray:
I see the default list, but It's not obviout to me how to extend this to support custom resources.

https://github.com/heptio/velero/blob/4583aa707821f5860b09b67cc091b70d504e1385/pkg/cmd/server/server.go#L440

ncdc:
It's a flag to `velero server`: https://github.com/heptio/velero/blob/73514a003b4cdaf1feed4e0976cff5a3400e4405/pkg/cmd/server/server.go#L167

dymurray:
Awesome, thank you @ncdc!

skriss:
Thanks for this request. We are not currently prioritizing this item within our roadmap, and as such will be closing out the issue. Please feel free to reach out as needed in this issue or a new one.

PR #608, opened 2018-06-25
don't snapshot volumes that have been backed up with restic

### LINKED ISSUE #604: Make volumes annotated for restic backups take priority over PV snapshots
If a volume is annotated to be backed up by restic, make sure we don't also take a PV snapshot for it. The restic annotation needs to take priority.

skriss:

Fixes #604


pkg/backup/volume_snapshot_tracker.go
```diff
@@ -0,0 +1,46 @@
...
+
+package backup
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// volumeSnapshotTracker keeps track of volumes that have been snapshotted.
```
nrb: I'd recommend stating that it tracks restic volumes that have been snapshotted.

pkg/backup/item_backupper.go
```diff
@@ -178,30 +182,36 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
...
+		backupErrs       []error
+		pod              *corev1api.Pod
+		podVolumeBackups []string
+	)
+
+	if groupResource == kuberesource.Pods {
+		// get the volumes to backup using restic, and add any of them that are PVCs to the pvc snapshot
+		// tracker, so that when we backup PVCs/PVs via an item action in the next step, we don't snapshot
+		// PVs that will have their data backed up with restic.
+		if pod, podVolumeBackups, err = getPodVolumesToBackup(obj, ib.pvcSnapshotTracker); err != nil {
```
ncdc: Having `getPodVolumesToBackup` also add to the tracker seems like it's doing more work than it should. What if we did this?
```go
pod, err = podFromUnstructured(obj)
// check err

resticVolumesToBackup := restic.GetVolumesToBackup(pod)
ib.pvcSnapshotTracker.Track(pod, resticVolumesToBackup) // this has the looping logic
```

(I think `resticVolumesToBackup` is more descriptive than `podVolumeBackups`.)

skriss: Good idea. I think I was tired when I wrote this. updated.

pkg/backup/volume_snapshot_tracker.go
skriss: done.

pkg/backup/item_backupper.go
```diff
@@ -248,18 +266,18 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 	return nil
 }
 
-func backupPodVolumes(log logrus.FieldLogger, backup *api.Backup, obj runtime.Unstructured, backupper restic.Backupper) (runtime.Unstructured, []error) {
-	if backupper == nil {
-		log.Warn("No restic backupper, not backing up pod's volumes")
+func (ib *defaultItemBackupper) backupPodVolumes(log logrus.FieldLogger, obj runtime.Unstructured, pod *corev1api.Pod, volumes []string) (runtime.Unstructured, []error) {
```
ncdc: Do you think this function would be cleaner if it didn't require both `obj` (an unstructured pod) and `pod`? We could have it return `volumeSnapshots` and call `restic.SetPodSnapshotAnnotation` outside of here.

skriss: Yes. updated and this is now much cleaner :)

pkg/backup/pvc_snapshot_tracker.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"fmt"
+
+	corev1api "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// pvcSnapshotTracker keeps track of persistent volume claims that have been snapshotted
+// with restic.
+type pvcSnapshotTracker struct {
```
nrb: Thanks for the comment on this. On a second pass, I'm wondering if having `restic` in the name might be helpful.

For example, in [backup.go, L289](https://github.com/heptio/ark/pull/608/files#diff-ca7997ad3f41210fc3bfc1fb6f537110R289), it's near `snapshotService`, which is for cloud volumes. Looking at the two names could get confusing, given they're for different types of snapshots.

I don't want to encode too much information in names, but the closeness here seems a bit off.

skriss: I renamed the variable to `resticSnapshotTracker` and left the type name as `pvcSnapshotTracker`. I think that makes it pretty clear. let me know what you think.

ncdc:
I've been thinking a bit about how we're encoding more and more custom per-resource logic into the backupper/restorer instead of using item actions.

One way we could achieve this with an item action would be to create one that applies to pods and pvs. We would also need a prioritized order for resources when backing up (#586), where we make sure we do pods before pvs. We'd maybe want to consider creating a special-case interface specifically for snapshotting, that would run after the normal item actions. In this interface we'd need a way for 1 plugin to signal to ark either a) that it's handling the snapshot, or b) that no other plugins should execute a snapshot. If we have all that, we could have a theoretical resticSnapshotAction:

1. The plugin would get its own copy of a restic repo manager (maybe a bad thing, but let's say it's ok)
1. Upon seeing a pod, the plugin would create PVBs for all appropriate volumes and add the ns/pvcName to its tracking map
1. Upon seeing a pv, if the pv's claimRef is in the tracking map, tell ark it's "handling" the snapshot / no other plugins should execute a snapshot

We should not do this now - this PR is fine to go in. But we should start to think about ways that we can make bits and pieces of logic like this a bit more self contained as plugins, instead of having to modify the core backup/restore functions.

pkg/backup/item_backupper.go
```diff
@@ -355,6 +359,23 @@ func (ib *defaultItemBackupper) takePVSnapshot(pv runtime.Unstructured, backup *
 		return nil
 	}
 
+	// If this PV is claimed, see if we've already taken a (restic) snapshot of the contents
+	// of this PV. If so, don't take a snapshot.
+	if collections.Exists(pv.UnstructuredContent(), "spec.claimRef") {
```
ncdc: Do you want to convert to an actual PV here?

ncdc:
Any plans to unit test the tracking in `backupItem`?

skriss:
Yeah, I'll add some tests.

pkg/backup/item_backupper.go
skriss: sure, will make this simpler

skriss:
OK got tests updated

ncdc:
LGTM

PR #610, opened 2018-06-25
log and skip backups for non-existent & hostPath volumes

skriss:

### DIFF
--- a/pkg/restic/backupper.go
+++ b/pkg/restic/backupper.go
@@ -115,9 +115,27 @@ func (b *backupper) BackupPodVolumes(backup *arkv1api.Backup, pod *corev1api.Pod
 	var (
 		errs            []error
 		volumeSnapshots = make(map[string]string)
+		podVolumes      = make(map[string]corev1api.Volume)
 	)
 
+	// put the pod's volumes in a map for efficient lookup below
+	for _, podVolume := range pod.Spec.Volumes {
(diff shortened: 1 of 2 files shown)


pkg/restic/backupper.go
```diff
@@ -170,6 +182,26 @@ ForEachVolume:
 	return volumeSnapshots, errs
 }
 
+func volumeExists(pod *corev1api.Pod, volumeName string) bool {
```
ncdc: Is it worth optimizing this into a single function so we don't have to range over pod.Spec.Volumes twice?

ncdc:
```go
func resticCanBackup(log logrus.FieldLogger, pod *corev1api.Pod, volumeName string) bool {
  for _, volume := range pod.Spec.Volumes {
    if volume.Name == volumeName {
      if volume.HostPath != nil {
        // log that it's hostPath
        return false
      }
      return true
    }
  }
  // log that we couldn't find it
  return false
}

ncdc: If it's easier to unit test them separately, it's fine to leave as-is. We know how to optimize and can do it later if needed.

skriss:
switched to the map-based approach so we loop through all pods in the volume only once total, rather than once or twice per volume to backup.

ncdc:
LGTM

PR #624, opened 2018-06-28
Fix k8s slack link in main navigation

carlisia:
Signed-off-by: Carlisia Pinto <carlisia@grokkingtech.io>

This fixes the slack link in the main navigation:

image

Currently it points to a non existing channel `ark`, but it needs to point to `ark-dr`.


ncdc:
LGTM!

PR #625, opened 2018-06-28
add FAQ about using a bucket per cluster

skriss:


ncdc:
LGTM. @Bradamant3 I'd like to feature this prominently in the docs reorg - something like a "recommended configuration settings" section.

skriss:
Yeah - adding a "best practices" section would be a good idea i think

PR #626, opened 2018-06-28
Don't restore backups or restores

### LINKED ISSUE #622: Properly handle restoring backups and restores from the backup tarball
If a backup tarball contains ark `backups` and `restores`, when we restore those items, we'll reset their metadata and status (just like we do with all other items). Unfortunately this means that the ark backup and restore controllers end up seeing them as `New`. Ark then runs backups/restores for all of them.

We should add `backups.ark.heptio.com` to the list of `nonRestorableResources`. There is no reason to restore them, as they'll come in via the backup sync controller.

If we do decide to restore `restores`, we need to special-case them and preserve their status, at least the phase.

I do think it's valid to restore `schedules`.

ncdc:
Add backups and restores the list of non restorable resources. Backups,
if applicable, are synced from object storage by the backup sync
controller. Restores are specific to a cluster and don't have value
moving across clusters.

Fixes #622

### DIFF
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -51,7 +51,19 @@ import (
 
 // nonRestorableResources is a blacklist for the restoration process. Any resources
 // included here are explicitly excluded from the restoration process.
-var nonRestorableResources = []string{"nodes", "events", "events.events.k8s.io"}
+var nonRestorableResources = []string{
+	"nodes",
+	"events",
+	"events.events.k8s.io",
+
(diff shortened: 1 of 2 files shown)


skriss:
LGTM.

skriss:
Tested by creating a backup that included backups & restores (manually verified they were in the tarball), deleted the backups/restores that were backed up, then did a restore and confirmed that the backups/restores were not re-created, and the logs properly indicated they were skipped.

ncdc:
Merging based on LGTM

PR #628, opened 2018-06-29
pkg/restic unit tests

skriss:

### DIFF
--- a/pkg/restic/command_factory_test.go
+++ b/pkg/restic/command_factory_test.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2018 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package restic
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBackupCommand(t *testing.T) {
+	c := BackupCommand("repo-id", "password-file", "path", map[string]string{"foo": "bar", "c": "d"})
+
+	assert.Equal(t, "backup", c.Command)
+	assert.Equal(t, "repo-id", c.RepoIdentifier)
+	assert.Equal(t, "password-file", c.PasswordFile)
+	assert.Equal(t, "path", c.Dir)
+	assert.Equal(t, []string{"."}, c.Args)
+
+	expected := []string{"--tag=foo=bar", "--tag=c=d", "--hostname=ark"}
+	sort.Strings(expected)
+	sort.Strings(c.ExtraFlags)
+	assert.Equal(t, expected, c.ExtraFlags)
+}
+
+func TestRestoreCommand(t *testing.T) {
+	c := RestoreCommand("repo-id", "password-file", "snapshot-id", "target")
+
+	assert.Equal(t, "restore", c.Command)
+	assert.Equal(t, "repo-id", c.RepoIdentifier)
+	assert.Equal(t, "password-file", c.PasswordFile)
+	assert.Equal(t, "target", c.Dir)
+	assert.Equal(t, []string{"snapshot-id"}, c.Args)
+	assert.Equal(t, []string{"--target=."}, c.ExtraFlags)
+}
+
+func TestGetSnapshotCommand(t *testing.T) {
+	c := GetSnapshotCommand("repo-id", "password-file", map[string]string{"foo": "bar", "c": "d"})
+
+	assert.Equal(t, "snapshots", c.Command)
+	assert.Equal(t, "repo-id", c.RepoIdentifier)
+	assert.Equal(t, "password-file", c.PasswordFile)
+
+	expected := []string{"--json", "--last", "--tag=foo=bar,c=d"}
+	sort.Strings(expected)
+	sort.Strings(c.ExtraFlags)
+	assert.Equal(t, expected, c.ExtraFlags)
+}
+
+func TestInitCommand(t *testing.T) {
+	c := InitCommand("repo-id")
+
+	assert.Equal(t, "init", c.Command)
+	assert.Equal(t, "repo-id", c.RepoIdentifier)
+}
+
+func TestCheckCommand(t *testing.T) {
+	c := CheckCommand("repo-id")
+
+	assert.Equal(t, "check", c.Command)
+	assert.Equal(t, "repo-id", c.RepoIdentifier)
+}
+
+func TestPruneCommand(t *testing.T) {
+	c := PruneCommand("repo-id")
+
+	assert.Equal(t, "prune", c.Command)
+	assert.Equal(t, "repo-id", c.RepoIdentifier)
+}
+
+func TestForgetCommand(t *testing.T) {
+	c := ForgetCommand("repo-id", "snapshot-id")
+
+	assert.Equal(t, "forget", c.Command)
+	assert.Equal(t, "repo-id", c.RepoIdentifier)
+	assert.Equal(t, []string{"snapshot-id"}, c.Args)
+}
--- a/pkg/restic/command_test.go
+++ b/pkg/restic/command_test.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2018 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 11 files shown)


pkg/restic/config_test.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	arkv1api "github.com/heptio/ark/pkg/apis/ark/v1"
+)
+
+func TestGetRepoIdentifier(t *testing.T) {
+	// TODO decide how to mock out aws.GetBucketRegion
```
skriss: interested in input here -- this is a good example of the general question "how to unit-test package-level functions that have dependencies?" I could create an interface/struct, `BucketRegionGetter`, and require that as an arg to the function - but that obviously makes the signatures more complicated and requires more object construction. Any other good ideas?

ncdc: @heptio/bluesteelsquad for input. I can't think of anything other than putting it behind an interface/struct offhand.

skriss: See latest commit for what it looks like

pkg/restic/command_factory_test.go
```diff
@@ -0,0 +1,82 @@
...
+	assert.Equal(t, "restore", c.Command)
+	assert.Equal(t, "repo-id", c.RepoIdentifier)
+	assert.Equal(t, "password-file", c.PasswordFile)
+	assert.Equal(t, "target", c.Dir)
+	assert.Equal(t, []string{"snapshot-id"}, c.Args)
+	assert.Equal(t, []string{"--target=."}, c.ExtraFlags)
+}
+
+func TestGetSnapshotCommand(t *testing.T) {
+	c := GetSnapshotCommand("repo-id", "password-file", map[string]string{"foo": "bar"})
```
ncdc: Test multiple tags?

pkg/restic/command_test.go
```diff
@@ -0,0 +1,106 @@
...
+	c := &Command{
+		Command:        "cmd",
+		RepoIdentifier: "repo-id",
+		PasswordFile:   "/path/to/password-file",
+		Dir:            "/some/pwd",
+		Args:           []string{"arg-1", "arg-2"},
+		ExtraFlags:     []string{"--foo=bar"},
+	}
+
+	require.NoError(t, os.Unsetenv("ARK_SCRATCH_DIR"))
```
ncdc: Should we consider renaming this to `ARK_RESTIC_CACHE_DIR` and removing the `.cache/restic` subdirectory creation?

pkg/restic/common_test.go
```diff
@@ -0,0 +1,284 @@
...
+			expected:    map[string]string{"": "bar"},
+		},
+		{
+			name:        "has snapshot annotation only, with suffix",
+			annotations: map[string]string{podAnnotationPrefix + "foo": "bar"},
+			expected:    map[string]string{"foo": "bar"},
+		},
+		{
+			name:        "has snapshot annotation, with suffix",
+			annotations: map[string]string{"foo": "bar", podAnnotationPrefix + "foo": "bar", podAnnotationPrefix + "abc": "123"},
```
ncdc: Let's use something besides foo/bar for the non-prefixed annotation, just to be safe.

pkg/restic/common_test.go
```diff
@@ -0,0 +1,284 @@
...
+			expected:    map[string]string{"foo": "bar", "abc": "123"},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			pod := &corev1api.Pod{}
+			pod.Annotations = test.annotations
+			assert.EqualValues(t, test.expected, GetPodSnapshotAnnotations(pod))
```
ncdc: Does `assert.Equal` fail here?

pkg/restic/common_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+			assert.EqualValues(t, test.expected, GetVolumesToBackup(pod))
+		})
+	}
+}
+
+func TestGetSnapshotsInBackup(t *testing.T) {
+	tests := []struct {
+		name             string
+		arkBackupName    string
```
ncdc: Since this doesn't appear to vary, should we remove it as a test case variable?

pkg/restic/common_test.go
```diff
@@ -0,0 +1,284 @@
...
+		{
+			name:          "some pod volume backups with matching label",
+			arkBackupName: "backup-1",
+			podVolumeBackups: []arkv1api.PodVolumeBackup{
+				{ObjectMeta: metav1.ObjectMeta{Name: "foo", Labels: map[string]string{arkv1api.BackupNameLabel: "non-matching-backup-1"}}},
+				{ObjectMeta: metav1.ObjectMeta{Name: "bar", Labels: map[string]string{arkv1api.BackupNameLabel: "non-matching-backup-2"}}},
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "pvb-1", Labels: map[string]string{arkv1api.BackupNameLabel: "backup-1"}},
+					Spec: arkv1api.PodVolumeBackupSpec{
+						Pod: corev1api.ObjectReference{Namespace: "ns-1"},
```
ncdc: Let's add the pod's name as it's invalid without it.

pkg/restic/common_test.go
```diff
@@ -0,0 +1,284 @@
...
+					},
+					Status: arkv1api.PodVolumeBackupStatus{SnapshotID: "snap-1"},
+				},
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "pvb-2", Labels: map[string]string{arkv1api.BackupNameLabel: "backup-1"}},
+					Spec: arkv1api.PodVolumeBackupSpec{
+						Pod: corev1api.ObjectReference{Namespace: "ns-2"},
```
ncdc: Let's add the pod's name as it's invalid without it.

pkg/restic/common_test.go
```diff
@@ -0,0 +1,284 @@
...
+					},
+					Status: arkv1api.PodVolumeBackupStatus{SnapshotID: ""},
```
ncdc: Is the snapshot ID blank to simulate an in-progress or failed PVB?

pkg/restic/common_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+			res, err := GetSnapshotsInBackup(arkBackup, pvbInformer.Lister())
+			assert.NoError(t, err)
+			assert.EqualValues(t, test.expected, res)
+		})
+	}
+}
+
+func TestTempCredentialsFile(t *testing.T) {
+	// TODO
```
ncdc: 😄

pkg/restic/config_test.go
```diff
@@ -0,0 +1,76 @@
...
+	getAWSBucketRegion = func(string) (string, error) {
+		return "", errors.New("no region found")
+	}
+	config = arkv1api.ObjectStorageProviderConfig{
+		CloudProviderConfig: arkv1api.CloudProviderConfig{Name: "aws"},
+		ResticLocation:      "bucket/prefix",
+	}
+	assert.Equal(t, "s3:s3.amazonaws.com/bucket/prefix/repo-1", GetRepoIdentifier(config, "repo-1"))
+
+	config = arkv1api.ObjectStorageProviderConfig{
```
ncdc: `getAWSBucketRegion` is still set to return an error - do we want to change it back?

pkg/restic/command_factory_test.go
```diff
@@ -0,0 +1,82 @@
...
+package restic
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBackupCommand(t *testing.T) {
+	c := BackupCommand("repo-id", "password-file", "path", map[string]string{"foo": "bar"})
```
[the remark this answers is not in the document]
skriss: done.

pkg/restic/command_factory_test.go
skriss: done.

pkg/restic/common_test.go
skriss: done.

pkg/restic/common_test.go
skriss: no, misunderstood the diff between the two, switched backed to `assert.Equal`

pkg/restic/common_test.go
skriss: done.

pkg/restic/common_test.go
skriss: done.

pkg/restic/common_test.go
skriss: done.

pkg/restic/common_test.go
skriss: yes; enhanced this test case with some more data.

pkg/restic/config_test.go
skriss: rearranged the test case so the error condition is tested first.

pkg/restic/command_test.go
skriss: I think it's fine as-is for now

pkg/restic/common_test.go
skriss: added.

skriss:
Squashed.

PR #629, opened 2018-06-29
[DO NOT MERGE] Fail backup if it already exists in object storage

### LINKED ISSUE #623: Disallow overwriting objects in object storage
We must not overwrite anything in object storage (backup json, logs, etc). If something needs to be replaced and retain the same name, the user must delete and then recreate (e.g. delete backup `foo`, then create a new backup `foo`).

ncdc:
Fixes #623


nrb:
I can build this successfully, but can't run it against GCP:

```
x1c in /home/nrb/go/src/github.com/heptio/ark (git) ncdc-check-for-backup
% GOOGLE_APPLICATION_CREDENTIALS=credentials-ark-gke ./_output/bin/linux/amd64/ark server --log-level debug
INFO[0000] setting log-level to DEBUG
INFO[0000] Starting Ark server v0.9.0-alpha.3-2-ge248811-dirty  logSource="pkg/cmd/server/server.go:100"
INFO[0000] Ensuring namespace exists for backups         logSource="pkg/cmd/server/server.go:250" namespace=heptio-ark
INFO[0000] Namespace already exists                      logSource="pkg/cmd/server/server.go:262" namespace=heptio-ark
INFO[0000] Retrieving Ark configuration                  logSource="pkg/cmd/server/server.go:267"
INFO[0000] Successfully retrieved Ark configuration      logSource="pkg/cmd/server/server.go:285"
INFO[0000] Using default resource priorities             logSource="pkg/cmd/server/server.go:335" priorities="[namespaces persistentvolumes persistentvolumeclaims secrets configmaps serviceaccounts limitranges pods]"
INFO[0000] Configuring cloud provider for backup service  logSource="pkg/cmd/server/server.go:384"
DEBU[0000] starting plugin                               args="[./_output/bin/linux/amd64/ark run-plugin cloudprovider gcp]" logSource="pkg/plugin/logrus_adapter.go:74" path=./_output/bin/linux/amd64/ark
DEBU[0000] waiting for RPC address                       logSource="pkg/plugin/logrus_adapter.go:74" path=./_output/bin/linux/amd64/ark
DEBU[0000] plugin address                                address=/tmp/plugin804103274 logSource="pkg/plugin/logrus_adapter.go:74" network=unix pluginName=ark
2018/06/29 16:04:07 [DEBUG] plugin: waiting for all plugin processes to complete...
DEBU[0000] plugin process exited                         logSource="pkg/plugin/logrus_adapter.go:74" path=./_output/bin/linux/amd64/ark
An error occurred: could not convert gRPC client to cloudprovider.ObjectStore
```

I notice that `pkg/plugin/proto/ObjectStore.proto` doesn't have this method added. Should it be there? Likewise, there's no implementation in `pkg/plugin/object_store.go`

nrb:
Error came from https://github.com/heptio/ark/blob/master/pkg/plugin/manager.go#L245-L247, which looks like object doesn't conform to the interface to me.

ncdc:
I’ll fix it when I get back to my desk. Thanks for testing!

ncdc:
I did some testing and came to the realization that if we have this sequence of events:

1. Backup `my-backup` already exists in object storage but not in Kubernetes
1. Start ark server (backup sync process hasn't run yet)
1. Immediately run `ark backup create my-backup`

then this PR results in a backup in Kubernetes called `my-backup` that has a phase of `FailedValidation`. When the backup sync controller runs and tries to sync the backup from object storage to Kubernetes, that will end up being a no-op because there's already a backup called `my-backup` in Kubernetes, even though it's invalid.

If we decide to continue down this path, I can think of a few options, none of which is fantastic:

1. When the backup sync controller runs, if it sees a backup in Kubernetes whose phase is `FailedValidation`, delete it, then try to create the backup using the data from object storage.
1. If the ark server encounters a backup with the same name in object storage, immediately delete the backup from Kubernetes and then create a `FailedBackup` custom resource that holds the information on the failure, and incorporate this new CRD when listing/getting/describing backups.
1. Do nothing. Require the user to delete the `FailedValidation` backup so the sync process can succeed for the backup in question.

@rosskukulinski @heptio/bluesteelsquad

ncdc:
@rosskukulinski I'll list them but I'm not sure they're relevant (will explain after the list):

- overlapping included/excluded namespaces (e.g. include foo, exclude foo)
- overlapping included/excluded resources (include pods, exclude pods)
- backup.spec.snapshotVolumes=true and ark server isn't configured with a pvProvider for snapshots

If there is a backup `good` in object storage, it was successfully processed and uploaded. If there is a backup `bad` in Kubernetes that failed validation, it can't be the same backup that's in object storage. Given that object storage is the source of truth, and there's a valid backup there, any invalid backup with the same name in Kubernetes is something else entirely.

rosskukulinski:
Ok.  I'm leaning towards (3) but I'm wondering how common is this situation?  I guess it happens because you found it while testing.  Does the backup-sync happen when Ark server starts up?

ncdc:
@skriss PTAL when you have a moment. If you think the Azure object store changes are too risky, I can back them out.

carlisia:
Closing this in favor of https://github.com/heptio/velero/pull/1390.

Issue #635, opened 2018-07-02
Consider tweaking restic metadata to allow better parent snapshot finding


skriss:
When taking a backup, restic first looks for a "parent" snapshot, i.e. **a previous snapshot from the same host of the same directory**, from which it can perform a differential backup.  This is key for making incremental backups efficient.

With Ark, the host name corresponds to the name of the daemonset pod from which the restic cmd is run. This means that if the pod is recreated, it will have a different name, and the next set of backups for pods on that node will be full rather than incremental.

Additionally, the path we're backing up includes a pod UID element. This means that if the workload's pod is recreated (i.e. after a backup and restore, or as part of a reschedule), the next backup will be full rather than incremental.

The hostname can be overriden as a flag when running `restic backup`. I think we may want to set this to a generic value, i.e. "ark", to prevent breaks in the incremental snapshot chain.

I think workload pod UIDs in the path are less of an issue (it doesn't seem unreasonable to take a new full backup after a restore), but we should think about this.

nrb:
What kind of severity/priority would you give this? Sounds to m like it should be in v0.9.0 in order to make incremental backups work as expected.

skriss:
If we agree that we should use a generic hostname, then yes I'll do it for 0.9.0.  I don't think the volume path thing is essential for 0.9.0, even if we decide we want to change something.

nrb:
Does the current behavior break restores?

skriss:
No - nothing's "broken" in the current state, it's just that we'll end up taking full rather than incremental backups in more situations than is strictly necessary, so some backups will be slower.

ncdc:
We can close this?

skriss:
Fixed with #636.

PR #636, opened 2018-07-02
set --hostname flag for restic backups

skriss:

Addresses part of #635 

Have tested on an AWS cluster by taking a backup, deleting the daemonset pods, taking a new backup, and manually verifying that the new restic backups have the parent snapshot ID set.


skriss:
Re: restic's description for the `--hostname` flag ("set the hostname for the snapshot manually. To prevent an expensive rescan use the "parent" flag") -- pretty sure it's alluding to the fact that if hostname is different than the previous backup's, it won't find it as a parent.  The code path for finding a parent is the same whether `--hostname` is specified or not (in which case it uses `os.Hostname()`).

ncdc:
Tested on minikube, lgtm!

PR #637, opened 2018-07-02
[WIP] convert from unstructured to structured objs when we know the type

skriss:
DO NOT MERGE.

Here's a sample of a couple of our item actions rewritten using the unstructured converter to convert to API types.  WDYT?

### DIFF
--- a/pkg/backup/backup_pv_action.go
+++ b/pkg/backup/backup_pv_action.go
@@ -17,13 +17,14 @@ limitations under the License.
 package backup
 
 import (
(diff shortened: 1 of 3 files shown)


ncdc:
Seems ok to me

nrb:
:+1: 

One thing we may need to watch out for is that the `DefaultUnstructuturedConverter` will insert timestamps if they're missing.  This was mostly an issue with testing expectations in the Service Account comparisons I did.

skriss:
replaced with #724

PR #639, opened 2018-07-05
Docs reorg

Bradamant3:
Will squash later -- assume we might want some further discussion. Change summary:

- edit root-level README further. N.B. This content is still intended to be the landing page for the docs site because it seems like the right high-level intro regardless of where the user sees it. (Open for discussion.)
- Move getting started content into its own file (quickstart.md)
- Make new "how it works" file (mostly out of old content, somewhat edited but could use more attention.)
- Edited restic doc a bit for conciseness. TODO (nice to have for this release but I'll try to get to it soon): break out the "how it works" part into its own page/file.

### DIFF
--- a/docs/about.md
+++ b/docs/about.md
@@ -1,22 +1,12 @@
-# About Heptio Ark
+# How Ark Works
 
-Heptio Ark provides customizable degrees of recovery for all Kubernetes objects (Pods, Deployments, Jobs, Custom Resource Definitions, etc.), as well as for persistent volumes. This recovery can be cluster-wide, or fine-tuned according to object type, namespace, or labels.
+Each Ark operation -- on-demand backup, scheduled backup, restore -- is a custom resource, defined with a Kubernetes [Custom Resource Definition (CRD)][20] and stored in [etcd][22]. The config custom resource specifies core information and options such as cloud provider settings. Ark also includes controllers that process the custom resources to perform backups, restores, and all related operations.
+
+You can back up or restore all objects in your cluster, or you can filter objects by type, namespace, and/or label.
 
 Ark is ideal for the disaster recovery use case, as well as for snapshotting your application state, prior to performing system operations on your cluster (e.g. upgrades).
 
-## Features
-
-Ark provides the following operations:
-
-* On-demand backups
-* Scheduled backups
-* Restores
-
-Each operation is a custom resource, defined with a Kubernetes [Custom Resource Definition (CRD)][20] and stored in [etcd][22]. An additional custom resource, Config, specifies required information and customized options, such as cloud provider settings. These resources are handled by [custom controllers][21] when their corresponding requests are submitted to the Kubernetes API server.
-
-Each controller watches its custom resource for API requests (Ark operations), performs validations, and handles the logic for interacting with the cloud provider API -- for example, managing object storage and persistent volumes.
-
-### On-demand backups
+## On-demand backups
 
 The **backup** operation:
 
@@ -29,25 +19,23 @@ need to tell a database to flush its in-memory buffers to disk before taking a s
 
 Note that cluster backups are not strictly atomic. If Kubernetes objects are being created or edited at the time of backup, they might not be included in the backup. The odds of capturing inconsistent information are low, but it is possible.
 
-### Scheduled backups
+## Scheduled backups
 
 The **schedule** operation allows you to back up your data at recurring intervals. The first backup is performed when the schedule is first created, and subsequent backups happen at the schedule's specified interval. These intervals are specified by a Cron expression.
 
-A Schedule acts as a wrapper for Backups; when triggered, it creates them behind the scenes.
-
 Scheduled backups are saved with the name `<SCHEDULE NAME>-<TIMESTAMP>`, where `<TIMESTAMP>` is formatted as *YYYYMMDDhhmmss*.
 
-### Restores
+## Restores
 
-The **restore** operation allows you to restore all of the objects and persistent volumes from a previously created Backup. Heptio Ark supports multiple namespace remapping--for example, in a single restore, objects in namespace "abc" can be recreated under namespace "def", and the ones in "123" under "456".
+The **restore** operation allows you to restore all of the objects and persistent volumes from a previously created backup. You can also restore only a filtered subset of objects and persistent volumes. Ark supports multiple namespace remapping--for example, in a single restore, objects in namespace "abc" can be recreated under namespace "def", and the objects in namespace "123" under "456".
 
-Kubernetes objects that have been restored can be identified with a label that looks like `ark-restore=<BACKUP NAME>-<TIMESTAMP>`, where `<TIMESTAMP>` is formatted as *YYYYMMDDhhmmss*.
+The default name of a restore is `<BACKUP NAME>-<TIMESTAMP>`, where `<TIMESTAMP>` is formatted as *YYYYMMDDhhmmss*. You can also specify a custom name. A restored object also includes a label with key `ark-restore` and value `<RESTORE NAME>`.
 
 You can also run the Ark server in restore-only mode, which disables backup, schedule, and garbage collection functionality during disaster recovery.
 
 ## Backup workflow
 
-Here's what happens when you run `ark backup create test-backup`:
+When you run `ark backup create test-backup`:
 
 1. The Ark client makes a call to the Kubernetes API server to create a `Backup` object.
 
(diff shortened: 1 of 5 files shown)


docs/about.md
```diff
@@ -1,22 +1,14 @@
-# About Heptio Ark
+# How Ark Works
 
-Heptio Ark provides customizable degrees of recovery for all Kubernetes objects (Pods, Deployments, Jobs, Custom Resource Definitions, etc.), as well as for persistent volumes. This recovery can be cluster-wide, or fine-tuned according to object type, namespace, or labels.
+Each Ark operation -- on-demand backup, scheduled back, restore -- is a custom resource, defined with a Kubernetes [Custom Resource Definition (CRD)][20] and stored in [etcd][22]. The config custom resource specifies core information and customized options such as cloud provider settings. These resources are handled by [custom controllers][21] on the Kubernetes API server.
```
ncdc: typo: scheduled back -> backup

ncdc: I know this is was there before, but would "customizable options" make more sense than "customized options"?

ncdc: "These resources are handled by custom controllers on the Kubernetes API server" doesn't sound right to me. What if we removed this sentence, and started the following paragraph with "Ark contains controllers that process its custom resources to execute backups, restores, etc. Each controller watches ..."? Or something like this?

docs/about.md
```diff
@@ -1,22 +1,14 @@
...
-
-Ark provides the following operations:
-
-* On-demand backups
-* Scheduled backups
-* Restores
+Each controller watches its custom resource for API requests (Ark operations), performs validations, and handles the logic for interacting with the cloud provider API -- for example, managing object storage and persistent volumes.
 
-Each operation is a custom resource, defined with a Kubernetes [Custom Resource Definition (CRD)][20] and stored in [etcd][22]. An additional custom resource, Config, specifies required information and customized options, such as cloud provider settings. These resources are handled by [custom controllers][21] when their corresponding requests are submitted to the Kubernetes API server.
+You can back up or restore all objects in your cluster, or you can specify operations accoording to object type, namespace, or label.
```
ncdc: and/or label?

docs/restic.md
```diff
@@ -47,15 +49,16 @@ If you're upgrading from a previous pre-release version of Ark with restic, befo
     - GCP: `kubectl apply -f examples/gcp/20-restic-daemonset.yaml`
     - Minio: `kubectl apply -f examples/minio/30-restic-daemonset.yaml`
 
-1. Update the image tag on the Ark daemonset and deployment to match the release version you used in Step 1 (e.g. `v0.9.0-alpha.3`):
+1. Update the image tag on the Ark daemonset and deployment to match the release version you're working with -- for example, `v0.9.0-alpha.3`:
```
ncdc: Let's remove `-alpha.3` 😄

docs/restic.md
```diff
@@ -108,21 +112,22 @@ You're now ready to use Ark with restic.
...
+
     ```bash
     ark backup create NAME OPTIONS...
     ```
 
-1. When the backup has completed, view information about your pod volume backups:
+1. When the backup completes, view information about the backups:
+
     ```bash
     kubectl -n heptio-ark get podvolumebackups -l ark.heptio.com/backup-name=YOUR_BACKUP_NAME -o yaml
```
ncdc: @skriss this needs to be updated now that #550 is in

docs/about.md
```diff
@@ -29,25 +21,23 @@ need to tell a database to flush its in-memory buffers to disk before taking a s
...
 Scheduled backups are saved with the name `<SCHEDULE NAME>-<TIMESTAMP>`, where `<TIMESTAMP>` is formatted as *YYYYMMDDhhmmss*.
 
-### Restores
+## Restores
 
-The **restore** operation allows you to restore all of the objects and persistent volumes from a previously created Backup. Heptio Ark supports multiple namespace remapping--for example, in a single restore, objects in namespace "abc" can be recreated under namespace "def", and the ones in "123" under "456".
+The **restore** operation allows you to restore all of the objects and persistent volumes from a previously created backup. Ark supports multiple namespace remapping--for example, in a single restore, objects in namespace "abc" can be recreated under namespace "def", and the objects in namespace "123" under "456".
 
-Kubernetes objects that have been restored can be identified with a label that looks like `ark-restore=<BACKUP NAME>-<TIMESTAMP>`, where `<TIMESTAMP>` is formatted as *YYYYMMDDhhmmss*.
+Restored objects are labeled with the name `ark-restore=<BACKUP NAME>-<TIMESTAMP>`, where `<TIMESTAMP>` is formatted as *YYYYMMDDhhmmss*.
```
nrb: This sentence reads strangely to me. The label's name is `ark-restore`, and the values are `<BACKUP NAME>-<TIMESTAMP>` in the default case, or the restore's name if a custom restore name was provided.

docs/restic.md
```diff
@@ -108,21 +112,22 @@ You're now ready to use Ark with restic.
     kubectl -n foo annotate pod/sample backup.ark.heptio.com/backup-volumes=pvc-volume,emptydir-volume
     ```
 
-
-    This annotation can also be provided in the pod template spec if using a deployment, daemonset, etc.
+    This annotation can also be provided in the pod template spec if you're using a deployment or daemonset
```
nrb: I'd update this to be "a deployment, daemonset, or replicaset"

ncdc: Also works for replicationcontrollers, jobs, statefulsets... hence the original "etc"

nrb: Good point. Perhaps, "This annotation can also be provider in a pod template spec."

Since the list is pretty exhaustive we don't want to list them all, but I also want to make sure we're not giving the false impression that it only supports the kinds explicitly listed.

docs/about.md
skriss: link [21] is missing

docs/about.md
```diff
@@ -1,22 +1,14 @@
...
-Ark is ideal for the disaster recovery use case, as well as for snapshotting your application state, prior to performing system operations on your cluster (e.g. upgrades).
-
-## Features
-
-Ark provides the following operations:
-
-* On-demand backups
-* Scheduled backups
-* Restores
+Each controller watches its custom resource for API requests (Ark operations), performs validations, and handles the logic for interacting with the cloud provider API -- for example, managing object storage and persistent volumes.
```
skriss: "handles the logic for interacting with the cloud provider API..." sounds a little weird to me, as if that's the only thing each controller does - that's just one of many things. I'd probably say something more general like "executes backups, restores, etc."

docs/about.md
skriss: acoording -> according

skriss: maybe "filter objects" rather than "specify operations"?

docs/about.md
```diff
@@ -29,25 +21,23 @@ need to tell a database to flush its in-memory buffers to disk before taking a s
...
 
-A Schedule acts as a wrapper for Backups; when triggered, it creates them behind the scenes.
-
 Scheduled backups are saved with the name `<SCHEDULE NAME>-<TIMESTAMP>`, where `<TIMESTAMP>` is formatted as *YYYYMMDDhhmmss*.
 
-### Restores
+## Restores
 
-The **restore** operation allows you to restore all of the objects and persistent volumes from a previously created Backup. Heptio Ark supports multiple namespace remapping--for example, in a single restore, objects in namespace "abc" can be recreated under namespace "def", and the ones in "123" under "456".
+The **restore** operation allows you to restore all of the objects and persistent volumes from a previously created backup. Ark supports multiple namespace remapping--for example, in a single restore, objects in namespace "abc" can be recreated under namespace "def", and the objects in namespace "123" under "456".
```
skriss: Maybe worthwhile stating that you can restore all, **or a filtered subset of**, objects/PVs from a previously created backup?

docs/about.md
```diff
@@ -57,24 +47,26 @@ Here's what happens when you run `ark backup create test-backup`:
 
 1. The `BackupController` makes a call to the object storage service -- for example, AWS S3 -- to upload the backup file.
 
-By default `ark backup create` makes disk snapshots of any persistent volumes. You can adjust the snapshots by specifying additional flags. See [the CLI help][30] for more information. Snapshots can be disabled with the option `--snapshot-volumes=false`.
+By default, `ark backup create` makes disk snapshots of any persistent volumes. You can adjust the snapshots by specifying additional flags. See [the CLI help][30] for more information. Snapshots can be disabled with the option `--snapshot-volumes=false`.
 
 ![19]
```
skriss: What's this for?

docs/restic.md
skriss: I'd either leave as-is, i.e. "deployment, daemonset, etc.", or say something like "in the pod template spec if you're using a controller to manage your pods"

docs/about.md
Bradamant3: :-) Backup flow diagram
You can see it live here: https://bradamant3.github.io/ark/master/index.html

skriss: aha!

skriss:
Could we add restic to the side navbar too? Not sure which section would make the most sense, since it's kind of a cross-cutting doc (setup, run, troubleshoot)

docs/about.md
Bradamant3: It occurs to me that this sentence boils down to "The controller does what a controller does." Let's try ... getting rid of it altogether :-)

Bradamant3:
@skriss plan is to add restic to nav. I don't do that bit until I build the docs from the release tag. It'll be its own top-level nav item; when we break it up it'll look better. (Next release!)

docs/about.md
Bradamant3: >1 comment on this one, so went with fewer fewer words. Happy to revisit if cut to excess.

docs/restic.md
Bradamant3: Steve tells me it's in #643

docs/about.md
```diff
@@ -29,9 +27,9 @@ Scheduled backups are saved with the name `<SCHEDULE NAME>-<TIMESTAMP>`, where `
 
 ## Restores
 
-The **restore** operation allows you to restore all of the objects and persistent volumes from a previously created backup. Ark supports multiple namespace remapping--for example, in a single restore, objects in namespace "abc" can be recreated under namespace "def", and the objects in namespace "123" under "456".
+The **restore** operation allows you to restore all of the objects and persistent volumes from a previously created backup. You can also restore only a filtered subset of objects and persistent volumes. Ark supports multiple namespace remapping--for example, in a single restore, objects in namespace "abc" can be recreated under namespace "def", and the objects in namespace "123" under "456".
 
-Restored objects are labeled with the name `ark-restore=<BACKUP NAME>-<TIMESTAMP>`, where `<TIMESTAMP>` is formatted as *YYYYMMDDhhmmss*.
+Restored objects have the default name `ark-restore=<BACKUP NAME>-<TIMESTAMP>`, where `<TIMESTAMP>` is formatted as *YYYYMMDDhhmmss*. If you provided a custom name, it is used instead.
```
skriss: This still isn't quite right. Each object that's restored has a Kubernetes label attached to it, whose key is `ark-restore`, and whose value is the name of the Ark restore (by default `BACKUP_NAME-TIMESTAMP`).  E.g. (see last line):

```
Name:           restic-test-workload-5b4c865bcb-ncvsh
Namespace:      default
Start Time:     Thu, 05 Jul 2018 09:10:08 -0700
Labels:         app=restic-test-workload
                ark-restore=a-20180705091011
```

ncdc:
LGTM, thanks @Bradamant3! Please let us know if you need help squashing (or if you'd prefer to leave as-is).

Bradamant3:
Squashed. Do y'all still YOLO? (It's been too long for me ...)

ncdc:
I prefer someone other than the author merge the PR

PR #646, opened 2018-07-05
replace pkg/restore's osFileSystem with pkg/util/filesystem's

skriss:

Somewhere along the line I extracted some code to `pkg/util/filesystem` but failed to replace its usage within `pkg/restore`. This PR does that.


ncdc:
Hold until after we cut 0.9

PR #649, opened 2018-07-06
use posix-compliant conditional for checking TAG_LATEST

skriss:


skriss:
@nrb ptal

PR #653, opened 2018-07-06
Remove outdated namespace deletion information

### LINKED ISSUE #652: Remove debugging-deletes.md
`debugging-deletes.md` shouldn't be relevant anymore, as we're now two releases past the version that had problems, and we've added code to remove the finalizer that caused the problem.  Remove and re-gen the docs without it.

nrb:
Fixes #652 


v0.9.0/troubleshooting.md
```diff
@@ -8,7 +8,6 @@ These tips can help you troubleshoot known issues. If they don't help, you can [
 
 * [Debug restores][1]
 
-[0]: debugging-deletes.md
```
skriss: this is still referenced above, need to remove the ref

Bradamant3:
Also need to update the TOC

nrb:
@skriss Removed from master and v0.9.0 docs

skriss:
Fixes #652

skriss:
LGTM

PR #655, opened 2018-07-06
Remove outdated namespace deletion content

nrb:
Longer term fix for #652 


skriss:
LGTM

PR #667, opened 2018-07-10
Record backup completion time before uploading

nrb:

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -370,6 +370,10 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
 		backup.Status.Phase = api.BackupPhaseCompleted
 	}
 
+	// Mark completion timestamp before serializing and uploading.
+	// Otherwise, the JSON file in object storage has a CompletionTimestamp of 'null'.
+	backup.Status.CompletionTimestamp.Time = controller.clock.Now()
+
 	backupJSON := new(bytes.Buffer)
 	if err := encode.EncodeTo(backup, "json", backupJSON); err != nil {
 		errs = append(errs, errors.Wrap(err, "error encoding backup"))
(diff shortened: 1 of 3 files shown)


carlisia:
@nrb lgtm. The comment is helpful.

I missed seeing a change to test file too but I see this controller fell a bit behind on tests  :)

nrb:
Good point - I'll update the tests to make sure we catch this change.

nrb:
I'm looking at using https://godoc.org/github.com/stretchr/testify/mock#MatchedBy, though I'm having a hard time implementing it.

ncdc:
You could try `mock.AnythingOfType`. Not sure if that matches nils though.

pkg/controller/backup_controller.go
```diff
@@ -385,19 +385,21 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
...
 	}
-
-	if err := controller.backupService.UploadBackup(bucket, backup.Name, backupJSONToUpload, backupFileToUpload, logFile); err != nil {
-		errs = append(errs, err)
-	}
-
 	backupScheduleName := backup.GetLabels()["ark-schedule"]
-	controller.metrics.SetBackupTarballSizeBytesGauge(backupScheduleName, backupSizeBytes)
 
+	// Mark completion timestamp before uploading. Otherwise, the JSON file in object storage
```
skriss: Can we organize lines 388-401 so that we (1) set the completion timestamp on the backup, (2) upload it, (3) record metrics (including creation of any vars needed to do so)? It's all a bit mixed together right now.

skriss:
@nrb can we use a JSON-serialized version of `backup` in L204 as the 3rd arg? You'd have to add something after L201 to set the expected completed time.

skriss:
Also needs a rebase.

nrb:
@skriss Comments addressed.

My tests were failing because I wasn't adding the timestamp before serialization in the original change...so the test was actually working, but the output was confusing to me.

pkg/controller/backup_controller.go
```diff
@@ -370,6 +370,11 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
 		backup.Status.Phase = api.BackupPhaseCompleted
 	}
 
+	// Mark completion timestamp before serializing and uploading.
+	// Otherwise, the JSON file in object storage has a CompletionTimestamp of 'null'.
+	backupScheduleName := backup.GetLabels()["ark-schedule"]
```
skriss: this line should move down to the metrics section.

pkg/controller/backup_controller_test.go
```diff
@@ -201,7 +203,14 @@ func TestProcessBackup(t *testing.T) {
 				backup.Status.Version = 1
 				backupper.On("Backup", backup, mock.Anything, mock.Anything, mock.Anything).Return(nil)
 
-				cloudBackups.On("UploadBackup", "bucket", backup.Name, mock.Anything, mock.Anything, mock.Anything).Return(nil)
+				// Ensure we don't have a nil CompletionTimestamp when uploading.
+				completionTimestampIsPresent := func(buf *bytes.Buffer) bool {
```
skriss: we should be able to test for the specific value, right? Rather than just non-null?

PR #676, opened 2018-07-13
WIP require namespace for backups/etc. to exist at server startup

### LINKED ISSUE #673: ark sa always needs namespace create perms
We want to run ark without any restore permissions at this moment in time.
We are running openshift 3.9
I created a copy of cluster-reader clusterrole and added:
```
- apiGroups:
  - ark.heptio.com
  resources:
  - '*'
  verbs:
  - '*'
```
when starting ark:
An error occurred: error creating namespace config-hist: namespaces is forbidden: User "system:serviceaccount:config-hist:ark" cannot create namespaces at the cluster scope: User "system:serviceaccount:config-hist:ark" cannot create namespaces at the cluster scope

running ark, able to make backups

**Environment:**

- Ark version (use `ark version`):

skriss:

Fixes #673 

Still needs to be tested.

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -34,7 +34,6 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
-	"k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


skriss:
testing looks OK to me.

nrb:
Should probably include updated quickstart docs with this.

ncdc:
@nrb 
> Should probably include updated quickstart docs with this.

In what way?

skriss:
@nrb not sure which docs would need updated here?

nrb:
I was thinking listing it here https://github.com/heptio/ark/blob/master/docs/quickstart.md#prerequisites

skriss:
I don't think it's necessary there - this is really only relevant if (a) you're trying to set up a read-only Ark role, per OP, or (b) you're storing backups/restores/schedules/etc. in a different namespace than you're running the Ark server.  In the quickstart, everything is in the same namespace, and applying the prereqs file creates it before deploying anything.

PR #680, opened 2018-07-17
Delete backups from etcd if they're not in storage

### LINKED ISSUE #309: Consider auto-deleting backups from etcd if they're not in object storage
Modify the backup sync controller to delete backups from kubernetes if they don't exist in object storage. WDYT @skriss?

carlisia:
This fixes #309.

Working on manual and automated testing of this.

### DIFF
--- a/pkg/controller/backup_sync_controller.go
+++ b/pkg/controller/backup_sync_controller.go
@@ -25,21 +25,28 @@ import (
 
 	kuberrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/heptio/ark/pkg/cloudprovider"
 	arkv1client "github.com/heptio/ark/pkg/generated/clientset/versioned/typed/ark/v1"
+	informers "github.com/heptio/ark/pkg/generated/informers/externalversions/ark/v1"
+	listers "github.com/heptio/ark/pkg/generated/listers/ark/v1"
 	"github.com/heptio/ark/pkg/util/kube"
 	"github.com/heptio/ark/pkg/util/stringslice"
 )
 
 type backupSyncController struct {
-	client        arkv1client.BackupsGetter
-	backupService cloudprovider.BackupService
-	bucket        string
-	syncPeriod    time.Duration
-	namespace     string
-	logger        logrus.FieldLogger
+	client               arkv1client.BackupsGetter
+	backupService        cloudprovider.BackupService
+	bucket               string
+	syncPeriod           time.Duration
+	namespace            string
+	backupLister         listers.BackupLister
+	backupInformerSynced cache.InformerSynced
+	logger               logrus.FieldLogger
 }
 
 func NewBackupSyncController(
@@ -48,6 +55,7 @@ func NewBackupSyncController(
 	bucket string,
 	syncPeriod time.Duration,
 	namespace string,
+	backupInformer informers.BackupInformer,
 	logger logrus.FieldLogger,
 ) Interface {
 	if syncPeriod < time.Minute {
@@ -55,12 +63,14 @@ func NewBackupSyncController(
 		syncPeriod = time.Minute
 	}
 	return &backupSyncController{
-		client:        client,
-		backupService: backupService,
-		bucket:        bucket,
-		syncPeriod:    syncPeriod,
-		namespace:     namespace,
-		logger:        logger,
+		client:               client,
+		backupService:        backupService,
+		bucket:               bucket,
+		syncPeriod:           syncPeriod,
+		namespace:            namespace,
+		backupLister:         backupInformer.Lister(),
+		backupInformerSynced: backupInformer.Informer().HasSynced,
+		logger:               logger,
 	}
 }
 
@@ -69,6 +79,11 @@ func NewBackupSyncController(
 // receives on the ctx.Done() channel.
 func (c *backupSyncController) Run(ctx context.Context, workers int) error {
 	c.logger.Info("Running backup sync controller")
+	c.logger.Info("Waiting for caches to sync")
+	if !cache.WaitForCacheSync(ctx.Done(), c.backupInformerSynced) {
+		return errors.New("timed out waiting for caches to sync")
+	}
+	c.logger.Info("Caches are synced")
 	wait.Until(c.run, c.syncPeriod, ctx.Done())
 	return nil
 }
@@ -84,9 +99,12 @@ func (c *backupSyncController) run() {
 	}
 	c.logger.WithField("backupCount", len(backups)).Info("Got backups from object storage")
 
+	cloudBackupNames := sets.NewString()
 	for _, cloudBackup := range backups {
 		logContext := c.logger.WithField("backup", kube.NamespaceAndName(cloudBackup))
 		logContext.Info("Syncing backup")
+
+		cloudBackupNames.Insert(cloudBackup.Name)
 
 		// If we're syncing backups made by pre-0.8.0 versions, the server removes all finalizers
 		// faster than the sync finishes. Just process them as we find them.
@@ -107,4 +125,32 @@ func (c *backupSyncController) run() {
 			}
 		}
 	}
+
+	c.deleteUnused(cloudBackupNames)
+	return
+}
+
+// deleteUnused deletes backup objects from Kubernetes if there is no corresponding backup in the object storage.
+func (c *backupSyncController) deleteUnused(cloudBackupNames sets.String) {
+	// Backups objects in Kubernetes
+	backups, err := c.backupLister.Backups(c.namespace).List(labels.Everything())
+	if err != nil {
+		c.logger.WithError(errors.WithStack(err)).Error("Error listing backup from Kubernetes")
+	}
+	if len(backups) == 0 {
+		return
+	}
+
+	// For each backup object in Kubernetes, verify if has a corresponding backup in the object storage. If not, delete it.
+	for _, backup := range backups {
+		if !cloudBackupNames.Has(backup.Name) {
+			if err := c.client.Backups(backup.Namespace).Delete(backup.Name, nil); err != nil {
+				c.logger.WithError(errors.WithStack(err)).Error("Error deleting unused backup from Kubernetes")
+			} else {
+				c.logger.Debugf("Deleted backup: %s", backup.Name)
+			}
+		}
+	}
+
+	return
 }
--- a/pkg/controller/backup_sync_controller_test.go
+++ b/pkg/controller/backup_sync_controller_test.go
@@ -21,12 +21,14 @@ import (
 	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	core "k8s.io/client-go/testing"
 
 	"github.com/heptio/ark/pkg/apis/ark/v1"
 	"github.com/heptio/ark/pkg/generated/clientset/versioned/fake"
+	informers "github.com/heptio/ark/pkg/generated/informers/externalversions"
 	"github.com/heptio/ark/pkg/util/stringslice"
 	arktest "github.com/heptio/ark/pkg/util/test"
 	"github.com/pkg/errors"
@@ -94,9 +96,10 @@ func TestBackupSyncControllerRun(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			var (
-				bs     = &arktest.BackupService{}
-				client = fake.NewSimpleClientset()
-				logger = arktest.NewLogger()
+				bs              = &arktest.BackupService{}
+				client          = fake.NewSimpleClientset()
+				sharedInformers = informers.NewSharedInformerFactory(client, 0)
+				logger          = arktest.NewLogger()
 			)
 
 			c := NewBackupSyncController(
@@ -105,6 +108,7 @@ func TestBackupSyncControllerRun(t *testing.T) {
 				"bucket",
 				time.Duration(0),
 				test.namespace,
+				sharedInformers.Ark().V1().Backups(),
 				logger,
 			).(*backupSyncController)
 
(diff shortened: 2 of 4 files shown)


pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +111,30 @@ func (c *backupSyncController) run() {
...
+	err = deleteUnused(c, cloudBackups)
+	if err != nil {
+		c.logger.WithError(errors.WithStack(err)).Error("Error deleting unused backup from Kubernetes")
+	}
+}
+
+// deleteUnused deletes a backup from Kubernetes if it does not exist in object storage
+func deleteUnused(c *backupSyncController, cloudBackups map[string]*api.Backup) error {
+	// Backups from Kubernetes
+	bl, err := c.client.Backups(c.namespace).List(metav1.ListOptions{})
```
skriss: We'll want to switch to using a backup lister here, rather than a client.  Listers/informers are shared caches that can be used by many different controllers, that reflect the contents of the Kubernetes API, and are used to minimize load on the API server.  Take a look at the restore_controller, for a good example of using a `backupLister`.

carlisia: @skriss should I include the `backupInformer.Lister()` as a a param for the initialization of the sync_controller? Else I'm trying to see how to access that element since it resides in a different controller.

pkg/controller/backup_sync_controller.go
```diff
@@ -25,21 +25,28 @@ import (
...
-	syncPeriod    time.Duration
-	namespace     string
-	logger        logrus.FieldLogger
+	client             arkv1client.BackupsGetter
+	backupService      cloudprovider.BackupService
+	bucket             string
+	syncPeriod         time.Duration
+	namespace          string
+	backupLister       listers.BackupLister
+	backupListerSynced cache.InformerSynced
```
skriss: nit: we usually call these e.g. `backupInformerSynced`, so I'd switch to that for consistency

pkg/controller/backup_sync_controller.go
```diff
@@ -84,10 +94,13 @@ func (c *backupSyncController) run() {
 	}
 	c.logger.WithField("backupCount", len(backups)).Info("Got backups from object storage")
 
+	cloudBackups := map[string]*api.Backup{}
```
skriss: you could use a string set here since you only need the names

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +120,32 @@ func (c *backupSyncController) run() {
...
+		c.logger.WithError(errors.WithStack(err)).Error("Error deleting unused backup from Kubernetes")
+	}
+}
+
+// deleteUnused deletes a backup from Kubernetes if it does not exist in object storage
+func deleteUnused(c *backupSyncController, cloudBackups map[string]*api.Backup) error {
+	// Backups from Kubernetes
+	backups, err := c.backupLister.Backups(c.namespace).List(labels.Everything())
+	if err != nil {
+		return err
```
skriss: use `errors.WithStack(err)` - anywhere we're returning an error from non-Ark code, we wrap it using `errors.WithStack()` or `errors.Wrap()` which gives us a stack trace when it gets logged.

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +120,32 @@ func (c *backupSyncController) run() {
...
+	if len(backups) == 0 {
+		return nil
+	}
+
+	// For each backup item in Kubernetes, verify if has a corresponding backup in the cloud object storage. If not, delete it.
+	for _, backup := range backups {
+		if _, ok := cloudBackups[backup.Name]; !ok {
+			c.client.Backups(backup.Namespace).Delete(backup.Name, &metav1.DeleteOptions{})
+			if err != nil {
+				return err
```
skriss: do you want to break on the first error here? or should you delete as many as you can, and return a slice of errors encountered at the end?

carlisia: I wasn't clear if there could be multiples but this answers it! Also, thanks for the other comments with corresponding explanations.

pkg/controller/backup_sync_controller.go
```diff
@@ -84,10 +99,13 @@ func (c *backupSyncController) run() {
 	}
 	c.logger.WithField("backupCount", len(backups)).Info("Got backups from object storage")
 
+	cloudBackups := make(map[string]string)
```
skriss: I was thinking you could use a `sets.String` (from `k8s.io/apimachinery/pkg/util/sets`). We use this throughout the codebase for situations like this.

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +125,38 @@ func (c *backupSyncController) run() {
...
 	}
+
+	err = deleteUnused(c, cloudBackups)
+	if err != nil {
+		c.logger.WithError(errors.WithStack(err)).Error("Error deleting unused backup from Kubernetes")
+	}
+}
+
+// deleteUnused deletes backups from Kubernetes if they do not exist in object storage
+func deleteUnused(c *backupSyncController, cloudBackups map[string]string) error {
```
skriss: Thinking a little more about this, I'd probably change this func to take a `logrus.FieldLogger` (pass `c.logger`), not return an error, and just log immediately if any errors are encountered.  `run()` is just going to log them anyway.

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +125,38 @@ func (c *backupSyncController) run() {
...
+		return nil
+	}
+
+	// For each backup item in Kubernetes, verify if has a corresponding backup in the cloud object storage. If not, delete it.
+	var deleteErrors error
+	for _, backup := range backups {
+		if _, ok := cloudBackups[backup.Name]; !ok {
+			c.client.Backups(backup.Namespace).Delete(backup.Name, &metav1.DeleteOptions{})
+			if err != nil {
+				deleteErrors = multierror.Append(deleteErrors, errors.WithStack(err))
```
skriss: This is just an FYI given my previous comment recommending we not return errors:

When we're dealing with multiple errors, we usually either return a slice of errors, or use `k8s.io/apimachinery/pkg/util/errors` and the `NewAggregate()` func, which will aggregate them into a single error.  We could discuss whether using the hashicorp library would be a better solution in general, but otherwise, I think we'd prefer not to pull in an additional library.

carlisia:
@skriss how can I fake add items to Kubernetes so when this gets called: https://github.com/heptio/ark/pull/680/files#diff-fa7d01c5de5ce8bba301a9734797a863R138

...it will have items that don't exist in the client list of backups?

nrb:
So far I don't have much to add in addition to what Steve's already said. +1 to using `k8s.io/apimachinery/pkg/util/sets.NewString()` and relying on the API machinery `NewAggregate` func for now.

Gopkg.toml
```diff
@@ -119,4 +119,4 @@
 
 [[constraint]]
   name = "github.com/hashicorp/go-plugin"
-  branch = "master"
+  branch = "master"
```
ncdc: Please revert

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +126,33 @@ func (c *backupSyncController) run() {
 			}
 		}
 	}
+
+	deleteUnused(c, cloudBackupNames)
+	return
+}
+
+// deleteUnused deletes backups from Kubernetes if they do not exist in object storage
+func deleteUnused(c *backupSyncController, cloudBackupNames sets.String) {
```
ncdc: I would make this a method on `backupSyncController` instead of passing it in as a parameter

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +126,33 @@ func (c *backupSyncController) run() {
...
+	if err != nil {
+		c.logger.WithError(errors.WithStack(err)).Error("Error listing backup from Kubernetes")
+	}
+	if len(backups) == 0 {
+		return
+	}
+
+	// For each backup item in Kubernetes, verify if has a corresponding backup in the cloud object storage. If not, delete it.
+	for _, backup := range backups {
+		if ok := cloudBackupNames.Has(backup.Name); !ok {
```
ncdc: You could simplify this to `if !cloudBackupNames.Has(backup.Name)`

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +126,33 @@ func (c *backupSyncController) run() {
...
+			c.client.Backups(backup.Namespace).Delete(backup.Name, &metav1.DeleteOptions{})
```
ncdc: You need to assign the returned error from this call to `err`

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +126,33 @@ func (c *backupSyncController) run() {
...
+			if err != nil {
+				c.logger.WithError(errors.WithStack(err)).Error("Error deleting unused backup from Kubernetes")
+			} else {
+				fmt.Println("deleted ", backup.Name)
```
ncdc: You could log this at the debug or info level

pkg/controller/backup_sync_controller_test.go
```diff
@@ -154,3 +159,107 @@ func TestBackupSyncControllerRun(t *testing.T) {
...
+					}
+				}
+
+				// assert newly inserted backups exist in Kubernetes
+				assert.True(t, backupsExist(t, c, test.k8sBackups))
+
+				deleteUnused(c, cloudBackupNames)
+
+				// assert new backups have been deleted from Kubernetes
+				assert.False(t, backupsExist(t, c, test.k8sBackups))
```
ncdc: You'll want to validate the the "delete" calls are being invoked against the fake clientset. That's a better approach then checking in the lister to see if the backups have been deleted.

ncdc: You can use `pkg/util/test.CompareActions` - see `backup_deletion_controller_test.go`

pkg/controller/backup_sync_controller_test.go
```diff
@@ -154,3 +159,107 @@ func TestBackupSyncControllerRun(t *testing.T) {
...
+				if test.k8sBackups != nil {
+					err := sharedInformers.Ark().V1().Backups().Informer().GetStore().Add(backup.Backup)
+					if err != nil {
+						fmt.Println("Error adding k8s backup: ", err)
+						return
+					}
+				}
+
+				// assert newly inserted backups exist in Kubernetes
+				assert.True(t, backupsExist(t, c, test.k8sBackups))
```
ncdc: You don't need to do this - you added them to the informer, and that's sufficient.

carlisia:
I am suspicious of my test. This is what is happening:

When I call the test helper function `backupsExist`, I get:
`existing backup:  backupA` - this is a backup returned from fetching a list of k8s backups, as expected.

`>>>test backup:  backupA` - just to make sure we are iterating thru the expect list of test cases (of which there is only one at this point)

Then from `deleteUnused` I get:

```
----> trying to delete:  backupA
if I am here, delete failed because:  backups.ark.heptio.com "backupA" not found
```

And then, as expected because my test exclusively asserts the expected calls were made:

```
    --- PASS: TestDeleteUnused/normal_case (0.00s)
PASS
ok  	github.com/heptio/ark/pkg/controller	0.030s
Success: Tests passed.
```

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +126,36 @@ func (c *backupSyncController) run() {
...
+	return
+}
+
+// deleteUnused deletes backups from Kubernetes if they do not exist in object storage
+func (c *backupSyncController) deleteUnused(cloudBackupNames sets.String) {
+	// Backups from Kubernetes
+	backups, err := c.backupLister.Backups(c.namespace).List(labels.Everything())
+	if err != nil {
+		c.logger.WithError(errors.WithStack(err)).Error("Error listing backup from Kubernetes")
+	}
```
skriss: probably should explicitly return here

skriss:
@carlisia I think the issue is that when you're setting up the test, you're adding `backupA` to the informer, but not to the fake `clientset`.  After https://github.com/heptio/ark/pull/680/files#diff-0d56a17487f5913ed801d2528ceb5eaaR224, try creating the backup in the client with something like `client.Ark().V1().Backups(backup.Namespace).Create(backup)`.

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +126,36 @@ func (c *backupSyncController) run() {
...
+	if len(backups) == 0 {
+		return
+	}
+
+	// For each backup item in Kubernetes, verify if has a corresponding backup in the cloud object storage. If not, delete it.
+	for _, backup := range backups {
+		fmt.Println("----> trying to delete: ", backup.Name)
+		if !cloudBackupNames.Has(backup.Name) {
+			err := c.client.Backups(backup.Namespace).Delete(backup.Name, &metav1.DeleteOptions{})
```
skriss: you can pass nil for the second parameter. you can also collapse this and the next line (`if err := ...; err != nil {...`)

skriss:
@carlisia need anything from me to get this over the finish line?

carlisia:
@skriss if this test doesn't make sense I'll delete it: https://github.com/heptio/ark/pull/680/files#diff-0d56a17487f5913ed801d2528ceb5eaaR243

PS: I still have print statements to be deleted.

nrb:
@carlisia That test makes sense to me. I can take a look to see if I can find out why it's failing.

pkg/controller/backup_sync_controller.go
```diff
@@ -107,4 +126,35 @@ func (c *backupSyncController) run() {
 			}
 		}
 	}
+
+	c.deleteUnused(cloudBackupNames)
+	return
+}
+
+// deleteUnused deletes backups from Kubernetes if there is no corresponding backup in the client
```
nrb: Do you mean `if there is no corresponding backup in the object storage`?

carlisia: Yes. I'll correct that.

pkg/controller/backup_sync_controller_test.go
```diff
@@ -154,3 +158,133 @@ func TestBackupSyncControllerRun(t *testing.T) {
...
+				arktest.NewTestBackup().WithNamespace("ns-1").WithName("backup-1").Backup,
+				arktest.NewTestBackup().WithNamespace("ns-1").WithName("backup-2").Backup,
+				arktest.NewTestBackup().WithNamespace("ns-1").WithName("backup-3").Backup,
+			},
+			k8sBackups: []*arktest.TestBackup{
+				arktest.NewTestBackup().WithNamespace("ns-1").WithName("backupA").WithPhase(v1.BackupPhaseCompleted),
+				arktest.NewTestBackup().WithNamespace("ns-1").WithName("backupB").WithPhase(v1.BackupPhaseCompleted),
+				arktest.NewTestBackup().WithNamespace("ns-1").WithName("backupC").WithPhase(v1.BackupPhaseCompleted),
+			},
+			numMacthingBackups: 0,
```
nrb: I think this should be 3.

None of the 3 `k8sBackups` entries are present in `cloudBackups`. Therefore, they should be deleted.

pkg/controller/backup_sync_controller_test.go
```diff
@@ -154,3 +158,133 @@ func TestBackupSyncControllerRun(t *testing.T) {
...
 }
+
+func TestDeleteUnused(t *testing.T) {
+	tests := []struct {
+		name               string
+		getAllBackupsError error
+		cloudBackups       []*v1.Backup
+		k8sBackups         []*arktest.TestBackup
+		namespace          string
+		numMacthingBackups int
```
nrb: Typo: numMatchingBackups

pkg/controller/backup_sync_controller_test.go
nrb: Hm, no, this is correct. But something may not be mocked correctly since the `k8sBackups` still show up in the `c.backupLister.Backups(c.namespace).List` call.

pkg/controller/backup_sync_controller_test.go
```diff
@@ -154,3 +158,133 @@ func TestBackupSyncControllerRun(t *testing.T) {
...
+
+			// get names of client backups
+			testBackupNames := sets.NewString()
+			for _, cloudBackup := range test.cloudBackups {
+				testBackupNames.Insert(cloudBackup.Name)
+			}
+
+			c.deleteUnused(testBackupNames)
+
+			numBackups, err := numBackups(t, c) // this is returning 3, but all backups should've been deleted since the test case doesn't have overlapping backups
```
skriss: The issue here is that, because the **informer** isn't actually *running* in this test, it's not being updated when the sync controller issues a Delete call to the **client**.  The easiest thing to do would be to change `numBackups` to use the **client** to list backups, e.g. `client.Ark().Backups(test.Namespace).List(...)`

skriss: Or maybe not? Catching up on slack real quick, h/o before processing this comment :)

pkg/controller/backup_sync_controller_test.go
```diff
@@ -154,3 +158,133 @@ func TestBackupSyncControllerRun(t *testing.T) {
...
+			assert.NoError(t, err)
+			assert.Equal(t, test.numMacthingBackups, numBackups)
```
skriss: so this is really verifying "how many backups do I expect to exist in etcd after running the sync controller", which needs to account for both creates from the sync controller as well as deletes. 

I would change your test cases so that there are no cloud backups that aren't already k8s backups, so that you're never triggering creates, and you just need to count deletes.

pkg/controller/backup_sync_controller_test.go
skriss: OK I think both this and the comment below are at play here

pkg/controller/backup_sync_controller_test.go
```diff
@@ -154,3 +158,133 @@ func TestBackupSyncControllerRun(t *testing.T) {
...
+			// setup: insert backups into Kubernetes
+			for _, backup := range test.k8sBackups {
+				actionCreate := core.NewCreateAction(
+					v1.SchemeGroupVersion.WithResource("backups"),
+					test.namespace,
+					backup.Backup,
+				)
+				expectedActions = append(expectedActions, actionCreate)
+
+				actionDelete := core.NewDeleteAction(
```
skriss: you're not going to expect a delete for **every** k8s backup in each test case -- this is only expected for k8s backups that don't have a corresponding cloud backup.  Need to conditionally setup this expectation

carlisia: Ha. This is the worse offender! Thanks for catching this.

carlisia:
@skriss @nrb please take another look.

pkg/controller/backup_sync_controller_test.go
```diff
@@ -154,3 +158,139 @@ func TestBackupSyncControllerRun(t *testing.T) {
 		})
 	}
 }
+
+func TestDeleteUnused(t *testing.T) {
+	tests := []struct {
+		name               string
+		getAllBackupsError error
```
skriss: can remove `getAllBackupsError` as it's not used

pkg/controller/backup_sync_controller_test.go
```diff
@@ -154,3 +158,139 @@ func TestBackupSyncControllerRun(t *testing.T) {
...
+				client.ArkV1(),
+				bs,
+				"bucket",
+				time.Duration(0),
+				test.namespace,
+				sharedInformers.Ark().V1().Backups(),
+				logger,
+			).(*backupSyncController)
+
+			expectedActions := make([]core.Action, 0)
```
skriss: maybe rename to `expectedDeletes` or `expectedDeleteActions` just to clarify

skriss:
LGTM

carlisia:
@nrb good to ship?

PR #681, opened 2018-07-17
[WIP] backup replication API types

skriss:
I've been iterating on some go structs for the API types we've been discussing.  This is very much a work in progress and may not be complete or accurate, but wanted to share and get input!


skriss:
@ncdc when you get a chance, I'd like your input on https://github.com/heptio/ark/pull/681/files#diff-0f1f79eb1e02c642504e6637a78d7547R1 - in particular, I've been thinking of having one CR per *provider*, with a list of *locations* as part of the spec, rather than having each provider + location combo be an independent CR.  One reason I like this is that the type enforces that there can only be one default location per provider.  There could also be some generic provider-level configuration that we could put in the provider spec, that would have to be duplicated for each target if they were all independent CRs.

Note that this PR is just to sketch out design possibilities, don't take it as complete/final by any means.

PR #682, opened 2018-07-18
Support for rbac v1beta1 on kubernetes v1.7.5+

### LINKED ISSUE #660: Ark v0.9.0 not compatible with pre-1.8 clusters
A user reported that Ark was crashing on a 1.7.x cluster.  It looks like the proximate cause for that is that the new backup item action that backs up cluster roles and role bindings related to a service account is using the `rbac.authorization.k8s.io/v1` API group/version, which doesn't exist pre-1.8 when RBAC went GA.

It's possible there are other issues too; we haven't tested with 1.7 clusters.

Need to decide what our level of support should be for older versions of Kube going forward, and decide if we want to try to fix this specific issue.

cc @heptio/bluesteelsquad @rosskukulinski

nrb:
Fixes #660 

### DIFF
--- a/pkg/backup/rbac.go
+++ b/pkg/backup/rbac.go
@@ -1,0 +1,140 @@
+/*
+Copyright 2018 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package backup
+
+import (
+	"github.com/pkg/errors"
+	rbac "k8s.io/api/rbac/v1"
+	rbacbeta "k8s.io/api/rbac/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	rbacclient "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	rbacbetaclient "k8s.io/client-go/kubernetes/typed/rbac/v1beta1"
+)
+
+// ClusterRoleBindingLister allows for listing ClusterRoleBindings in a version-independent way.
+type ClusterRoleBindingLister interface {
+	// List returns a slice of ClusterRoleBindings which can represent either v1 or v1beta1 ClusterRoleBindings.
+	List() ([]ClusterRoleBinding, error)
+}
+
+// noopClusterRoleBindingLister exists to handle clusters where RBAC is disabled.
+type noopClusterRoleBindingLister struct {
+}
+
+func (noop noopClusterRoleBindingLister) List() ([]ClusterRoleBinding, error) {
+	return []ClusterRoleBinding{}, nil
+}
+
+type v1ClusterRoleBindingLister struct {
+	client rbacclient.ClusterRoleBindingInterface
+}
+
+func (v1 v1ClusterRoleBindingLister) List() ([]ClusterRoleBinding, error) {
+	crbList, err := v1.client.List(metav1.ListOptions{})
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	var crbs []ClusterRoleBinding
+	for _, crb := range crbList.Items {
+		crbs = append(crbs, v1ClusterRoleBinding{crb: crb})
+	}
+
+	return crbs, nil
+}
+
+type v1beta1ClusterRoleBindingLister struct {
+	client rbacbetaclient.ClusterRoleBindingInterface
+}
+
+func (v1beta1 v1beta1ClusterRoleBindingLister) List() ([]ClusterRoleBinding, error) {
+	crbList, err := v1beta1.client.List(metav1.ListOptions{})
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	var crbs []ClusterRoleBinding
+	for _, crb := range crbList.Items {
+		crbs = append(crbs, v1beta1ClusterRoleBinding{crb: crb})
+	}
+
+	return crbs, nil
+}
+
+// NewClusterRoleBindingListerMap creates a map of RBAC version strings to their associated
+// ClusterRoleBindingLister structs.
+// Necessary so that callers to the ClusterRoleBindingLister interfaces don't need the kubernetes.Interface.
+func NewClusterRoleBindingListerMap(clientset kubernetes.Interface) map[string]ClusterRoleBindingLister {
+	return map[string]ClusterRoleBindingLister{
+		rbac.SchemeGroupVersion.Version:     v1ClusterRoleBindingLister{client: clientset.RbacV1().ClusterRoleBindings()},
+		rbacbeta.SchemeGroupVersion.Version: v1beta1ClusterRoleBindingLister{client: clientset.RbacV1beta1().ClusterRoleBindings()},
+		"": noopClusterRoleBindingLister{},
+	}
+}
+
+// ClusterRoleBinding abstracts access to ClusterRoleBindings whether they're v1 or v1beta1.
+type ClusterRoleBinding interface {
+	// Name returns the name of a ClusterRoleBinding.
+	Name() string
+	// ServiceAccountSubjects returns the names of subjects that are service accounts in the given namespace.
+	ServiceAccountSubjects(namespace string) []string
+	// RoleRefName returns the name of a ClusterRoleBinding's RoleRef.
+	RoleRefName() string
+}
+
+type v1ClusterRoleBinding struct {
+	crb rbac.ClusterRoleBinding
+}
+
+func (c v1ClusterRoleBinding) Name() string {
+	return c.crb.Name
+}
+
+func (c v1ClusterRoleBinding) RoleRefName() string {
+	return c.crb.RoleRef.Name
+}
+
+func (c v1ClusterRoleBinding) ServiceAccountSubjects(namespace string) []string {
+	var saSubjects []string
+	for _, s := range c.crb.Subjects {
+		if s.Kind == rbac.ServiceAccountKind && s.Namespace == namespace {
+			saSubjects = append(saSubjects, s.Name)
+		}
+	}
+	return saSubjects
+}
+
+type v1beta1ClusterRoleBinding struct {
+	crb rbacbeta.ClusterRoleBinding
+}
+
+func (c v1beta1ClusterRoleBinding) Name() string {
+	return c.crb.Name
+}
+
+func (c v1beta1ClusterRoleBinding) RoleRefName() string {
+	return c.crb.RoleRef.Name
+}
+
+func (c v1beta1ClusterRoleBinding) ServiceAccountSubjects(namespace string) []string {
+	var saSubjects []string
+	for _, s := range c.crb.Subjects {
+		if s.Kind == rbac.ServiceAccountKind && s.Namespace == namespace {
+			saSubjects = append(saSubjects, s.Name)
+		}
+	}
+	return saSubjects
+}
--- a/pkg/backup/service_account_action.go
+++ b/pkg/backup/service_account_action.go
@@ -25,28 +25,41 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
-	rbacclient "k8s.io/client-go/kubernetes/typed/rbac/v1"
 
 	"github.com/heptio/ark/pkg/apis/ark/v1"
+	arkdiscovery "github.com/heptio/ark/pkg/discovery"
 	"github.com/heptio/ark/pkg/kuberesource"
 )
 
 // serviceAccountAction implements ItemAction.
 type serviceAccountAction struct {
 	log                 logrus.FieldLogger
-	clusterRoleBindings []rbac.ClusterRoleBinding
+	clusterRoleBindings []ClusterRoleBinding
 }
 
 // NewServiceAccountAction creates a new ItemAction for service accounts.
-func NewServiceAccountAction(log logrus.FieldLogger, client rbacclient.ClusterRoleBindingInterface) (ItemAction, error) {
-	clusterRoleBindings, err := client.List(metav1.ListOptions{})
+func NewServiceAccountAction(log logrus.FieldLogger, clusterRoleBindingListers map[string]ClusterRoleBindingLister, discoveryHelper arkdiscovery.Helper) (ItemAction, error) {
+	// Look up the supported RBAC version
+	var supportedAPI metav1.GroupVersionForDiscovery
+	for _, ag := range discoveryHelper.APIGroups() {
+		if ag.Name == rbac.GroupName {
+			supportedAPI = ag.PreferredVersion
+			break
+		}
+	}
+
+	crbLister := clusterRoleBindingListers[supportedAPI.Version]
+
+	// This should be safe because the List call will return a 0-item slice
+	// if there's no matching API version.
+	crbs, err := crbLister.List()
 	if err != nil {
-		return nil, errors.WithStack(err)
+		return nil, err
 	}
 
 	return &serviceAccountAction{
 		log:                 log,
-		clusterRoleBindings: clusterRoleBindings.Items,
+		clusterRoleBindings: crbs,
 	}, nil
 }
 
@@ -76,14 +89,14 @@ func (a *serviceAccountAction) Execute(item runtime.Unstructured, backup *v1.Bac
 		roles     = sets.NewString()
 	)
 
-	for _, clusterRoleBinding := range a.clusterRoleBindings {
-		for _, subj := range clusterRoleBinding.Subjects {
-			if subj.Kind == rbac.ServiceAccountKind && subj.Namespace == namespace && subj.Name == name {
+	for _, crb := range a.clusterRoleBindings {
+		for _, s := range crb.ServiceAccountSubjects(namespace) {
+			if s == name {
 				a.log.Infof("Adding clusterrole %s and clusterrolebinding %s to additionalItems since serviceaccount %s/%s is a subject",
-					clusterRoleBinding.RoleRef.Name, clusterRoleBinding.Name, namespace, name)
+					crb.RoleRefName(), crb, namespace, name)
 
-				bindings.Insert(clusterRoleBinding.Name)
-				roles.Insert(clusterRoleBinding.RoleRef.Name)
+				bindings.Insert(crb.Name())
+				roles.Insert(crb.RoleRefName())
 				break
 			}
 		}
--- a/pkg/backup/service_account_action_test.go
+++ b/pkg/backup/service_account_action_test.go
@@ -24,29 +24,61 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"k8s.io/api/rbac/v1"
+	rbac "k8s.io/api/rbac/v1"
+	rbacbeta "k8s.io/api/rbac/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	rbacclient "k8s.io/client-go/kubernetes/typed/rbac/v1"
 
 	"github.com/heptio/ark/pkg/kuberesource"
 	arktest "github.com/heptio/ark/pkg/util/test"
 )
 
-type fakeClusterRoleBindingClient struct {
-	clusterRoleBindings []v1.ClusterRoleBinding
+func newV1ClusterRoleBindingList(rbacCRBList []rbac.ClusterRoleBinding) []ClusterRoleBinding {
+	var crbs []ClusterRoleBinding
+	for _, c := range rbacCRBList {
+		crbs = append(crbs, v1ClusterRoleBinding{crb: c})
+	}
 
-	rbacclient.ClusterRoleBindingInterface
+	return crbs
 }
 
-func (c *fakeClusterRoleBindingClient) List(opts metav1.ListOptions) (*v1.ClusterRoleBindingList, error) {
-	return &v1.ClusterRoleBindingList{
-		Items: c.clusterRoleBindings,
-	}, nil
+func newV1beta1ClusterRoleBindingList(rbacCRBList []rbacbeta.ClusterRoleBinding) []ClusterRoleBinding {
+	var crbs []ClusterRoleBinding
+	for _, c := range rbacCRBList {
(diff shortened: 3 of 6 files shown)


pkg/backup/service_account_action.go
```diff
@@ -21,32 +21,71 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/sets"
-	rbacclient "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/heptio/ark/pkg/apis/ark/v1"
+	arkdiscovery "github.com/heptio/ark/pkg/discovery"
 	"github.com/heptio/ark/pkg/kuberesource"
 )
 
+type clusterRoleBindingVersionWrapper struct {
```
nrb: Would like some guidance here - I don't think this struct is the way to go, but I was struggling coming up with an interface, given the ClusterRoleBinding is just a struct w/o methods.

pkg/backup/rbac.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return lister, nil
+}
+
+// ClusterRoleBinding abstracts access to ClusterRoleBindings whether they're v1 or v1beta1.
+type ClusterRoleBinding interface {
+	// Name returns the name of a ClusterRoleBinding.
+	Name() string
+	// Subjects translates a ClusterRoleBindings Subjects slice into a slice of subject wrapper objects.
+	Subjects() []subject
```
skriss: One thing you could do is instead of exposing `Subjects()` and needing the wrapper type for that, just add a method to the `ClusterRoleBinding` interface called `HasSubject` that takes a subject kind, namespace, and name, and returns true or false -- since that's all we really need. That would minimize the amount of this wrapper code that we need

skriss: Actually may need to be `HasServiceAccountSubject` so it doesn't have to take the kind (since that's part of the API group/version), but same idea.

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/sirupsen/logrus"
+	rbac "k8s.io/api/rbac/v1"
+	rbacbeta "k8s.io/api/rbac/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	rbacclient "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	rbacclientbeta "k8s.io/client-go/kubernetes/typed/rbac/v1beta1"
+)
+
+type CRBLister interface {
```
skriss: I think I'd prefer to spell out `ClusterRoleBinding` in the type names (interface + 4 structs). I'm fine using abbreviations for local variable names though

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+	// List returns a slice of ClusterRoleBindings which can represent either v1 or v1beta1 ClusterRoleBindings.
+	List() ([]ClusterRoleBinding, error)
+}
+
+type V1CRBLister struct {
```
skriss: this and the `V1beta1CRBLister1` don't need to be exported, just the interface

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	for _, crb := range crbList.Items {
+		crbs = append(crbs, v1beta1crb{crb: crb})
+	}
+
+	return crbs, nil
+}
+
+// NewCRBLister wraps the v1beta1 and v1 RBAC client's List with a common interface.
+func NewCRBLister(log logrus.FieldLogger, clientset kubernetes.Interface) (CRBLister, error) {
```
skriss: spell out `ClusterRoleBinding` in the func name

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+	var lister CRBLister
```
skriss: usually declare local vars closer to their usage (unlike a bunch of other languages)

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+	discoveryHelper, err := arkdiscovery.NewHelper(clientset.Discovery(), log)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	var supportedAPI metav1.GroupVersionForDiscovery
+	apigroups := discoveryHelper.APIGroups()
```
skriss: no real need to declare `apigroups` - can just range over `discoveryHelper.APIGroups()` in the next line

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+	for _, ag := range apigroups {
+		if ag.Name == "rbac.authorization.k8s.io" {
```
skriss: `rbac.GroupName`

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+			supportedAPI = ag.PreferredVersion
+			break
+		}
+	}
+
+	switch supportedAPI.Version {
+	case "v1":
+		lister = V1CRBLister{client: clientset.RbacV1().ClusterRoleBindings()}
```
skriss: I'd just directly return the lister, nil

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+	case "v1beta1":
+		lister = V1beta1CRBLister{client: clientset.RbacV1beta1().ClusterRoleBindings()}
```
skriss: I'd just directly return the lister, nil

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+	// Return an empty client if we don't have a known version.
+	default:
```
skriss: I'd just directly return `nil, nil` for the default (or get rid of default and return it outside the switch)

pkg/backup/service_account_action_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 the Heptio Ark contributors.
+
```
skriss: undo deletion of copyright

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+	apigroups := discoveryHelper.APIGroups()
+	for _, ag := range apigroups {
+		if ag.Name == "rbac.authorization.k8s.io" {
+			supportedAPI = ag.PreferredVersion
+			break
+		}
+	}
+
+	switch supportedAPI.Version {
+	case "v1":
```
skriss: could use `rbac.SchemeGroupVersion.Version` rather than "v1" here, and `rbacbeta.`... for the v1beta1  case

pkg/backup/rbac.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return crbs, nil
+}
+
+type V1beta1CRBLister struct {
+	client rbacclientbeta.ClusterRoleBindingInterface
+}
+
+func (v1beta1 V1beta1CRBLister) List() ([]ClusterRoleBinding, error) {
+	var crbs []ClusterRoleBinding
```
carlisia: Wanted to extend on what @skriss said to make sure this and I think one additional case of this is also changed: I'd move this `crbs` var to right above the `for` loop where it's used. It's a common thing in Go so it's expected but also avoids scrolling all over for where it's used if we expect its first use to happen right after where it's declared.

nrb:
> There are some exported methods that don't have a description

Which ones in particular? Happy to add more comments, though I'm not sure what's missing. The ones exposed via interfaces should be commented on the interfaces themselves.

carlisia:
> The ones exposed via interfaces should be commented on the interfaces themselves.

My bad, @nrb, those were precisely the already commented methods of interfaces :) 👍

nrb:
Tested this with Kubernetes v1.7.16. The server started (where it would previously fail), and the cluster role bindings we create with the Ark installation were backed up.

```
% tar xzvf arktest-data.tar.gz
resources/namespaces/cluster/heptio-ark.json
resources/secrets/namespaces/heptio-ark/ark-token-rd7w2.json
resources/secrets/namespaces/heptio-ark/cloud-credentials.json
resources/secrets/namespaces/heptio-ark/default-token-lr9x1.json
resources/clusterrolebindings.rbac.authorization.k8s.io/cluster/ark.json
resources/clusterroles.rbac.authorization.k8s.io/cluster/cluster-admin.json
resources/serviceaccounts/namespaces/heptio-ark/ark.json
resources/serviceaccounts/namespaces/heptio-ark/default.json
resources/configs.ark.heptio.com/namespaces/heptio-ark/default.json
resources/backups.ark.heptio.com/namespaces/heptio-ark/arktest.json
resources/backups.ark.heptio.com/namespaces/heptio-ark/nginx.json
```

```
% cat resources/clusterrolebindings.rbac.authorization.k8s.io/cluster/ark.json
{"apiVersion":"rbac.authorization.k8s.io/v1beta1","kind":"ClusterRoleBinding","metadata":{"annotations":{"kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"rbac.authorization.k8s.io/v1beta1\",\"kind\":\"ClusterRoleBinding\",\"metadata\":{\"annotations\":{},\"labels\":{\"component\":\"ark\"},\"name\":\"ark\",\"namespace\":\"\"},\"roleRef\":{\"apiGroup\":\"rbac.authorization.k8s.io\",\"kind\":\"ClusterRole\",\"name\":\"cluster-admin\"},\"subjects\":[{\"kind\":\"ServiceAccount\",\"name\":\"ark\",\"namespace\":\"heptio-ark\"}]}\n"},"creationTimestamp":"2018-07-20T16:30:02Z","labels":{"component":"ark"},"name":"ark","resourceVersion":"1348","selfLink":"/apis/rbac.authorization.k8s.io/v1beta1/clusterrolebindings/ark","uid":"2721ffd3-8c3a-11e8-9722-0a1f5e4b1da4"},"roleRef":{"apiGroup":"rbac.authorization.k8s.io","kind":"ClusterRole","name":"cluster-admin"},"subjects":[{"kind":"ServiceAccount","name":"ark","namespace":"heptio-ark"}]}% 
```

pkg/backup/rbac.go
```diff
@@ -0,0 +1,152 @@
...
+
+	return crbs, nil
+}
+
+type v1beta1ClusterRoleBindingLister struct {
+	client rbacclientbeta.ClusterRoleBindingInterface
+}
+
+func (v1beta1 v1beta1ClusterRoleBindingLister) List() ([]ClusterRoleBinding, error) {
+
```
skriss: rm blank line

pkg/backup/rbac.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	return crbs, nil
+}
+
+// NewClusterRoleBindingLister wraps the v1beta1 and v1 RBAC client's List with a common interface.
+func NewClusterRoleBindingLister(log logrus.FieldLogger, clientset kubernetes.Interface) (ClusterRoleBindingLister, error) {
+	discoveryHelper, err := arkdiscovery.NewHelper(clientset.Discovery(), log)
+	if err != nil {
+		return nil, errors.WithStack(err)
```
skriss: don't need a `errors.WithStack` here since the error is coming from ark code so should already have a stack

pkg/backup/rbac.go
```diff
@@ -0,0 +1,152 @@
...
+	// Name returns the name of a ClusterRoleBinding.
+	Name() string
+	// ServiceAccountSubjects returns the names of subjects that are service accounts in the given namespace.
+	ServiceAccountSubjects(namespace string) []string
+	// RoleRefName returns the name of a ClusterRoleBinding's RoleRef.
+	RoleRefName() string
+}
+
+// v1 implementation
+type v1crb struct {
```
skriss: i'd probably spell out ClusterRoleBinding in the type name here

pkg/backup/rbac.go
```diff
@@ -0,0 +1,152 @@
...
+		if s.Kind == rbac.ServiceAccountKind && s.Namespace == namespace {
+			saSubjects = append(saSubjects, s.Name)
+		}
+	}
+	return saSubjects
+}
+
+// v1beta1 implementation
+
+type v1beta1crb struct {
```
skriss: i'd probably spell out ClusterRoleBinding in the type name here

pkg/backup/rbac.go
```diff
@@ -0,0 +1,152 @@
...
+	var saSubjects []string
+	for _, s := range c.crb.Subjects {
+		if s.Kind == rbac.ServiceAccountKind && s.Namespace == namespace {
+			saSubjects = append(saSubjects, s.Name)
+		}
+	}
+	return saSubjects
+}
+
+// v1beta1 implementation
```
skriss: I'd remove this, put a godoc comment on the struct if you want but not necessary

pkg/backup/service_account_action.go
```diff
@@ -76,15 +73,14 @@ func (a *serviceAccountAction) Execute(item runtime.Unstructured, backup *v1.Bac
...
+			if s == name {
 				a.log.Infof("Adding clusterrole %s and clusterrolebinding %s to additionalItems since serviceaccount %s/%s is a subject",
-					clusterRoleBinding.RoleRef.Name, clusterRoleBinding.Name, namespace, name)
+					crb.RoleRefName(), crb, namespace, name)
 
-				bindings.Insert(clusterRoleBinding.Name)
-				roles.Insert(clusterRoleBinding.RoleRef.Name)
-				break
+				bindings.Insert(crb.Name())
+				roles.Insert(crb.RoleRefName())
```
skriss: you can still break once you've found a matching subject

pkg/discovery/helper.go
```diff
@@ -45,6 +45,8 @@ type Helper interface {
 	// Refresh pulls an updated set of Ark-backuppable resources from the
 	// discovery API.
 	Refresh() error
+
+	APIGroups() []metav1.APIGroup
```
skriss: add godoc

pkg/backup/service_account_action.go
nrb: Is it not valid to have multiple service account subjects? Wouldn't we want to back them all up?

nrb:
Double-checking with a 1.9 cluster on GKE, ClusterRoleBindings still back up as expected with this change.

pkg/backup/service_account_action.go
nrb: Added back

pkg/cmd/server/plugin/plugin.go
```diff
@@ -87,8 +87,13 @@ func NewCommand(f client.Factory) *cobra.Command {
 					clientset, err := f.KubeClient()
 					cmd.CheckError(err)
 
-					action, err = backup.NewServiceAccountAction(logger, clientset.RbacV1().ClusterRoleBindings())
+					crbLister, err := backup.NewClusterRoleBindingLister(logger, clientset)
 					cmd.CheckError(err)
+
+					if crbLister != nil {
```
skriss: What happens if crbLister **is** null (i.e. RBAC is not enabled on this cluster)?  We still go through the code paths of setting up an action item plugin and serving it - seems like there may be a problem there.

skriss:
Apart from the question re: plugin serving, the rest LGTM.

pkg/cmd/server/plugin/plugin.go
nrb: Good catch - I had intended to address this and forgot.

Would it make sense to check if the action is nil before adding to the plugin configuration? Or maybe move the NewClusterRoleBindingLister logic into the NewServiceAccount action function?

Also, do we support clusters without RBAC? I didn't think we did, but even if that's the case this behavior needs to be fixed.

skriss: I don't know if we've ever really tried on clusters without RBAC.  Hopefully shouldn't be much of a use case going fwd, but who knows.

I'd say it makes sense to always have a non-nil action.  We can move the `NewClusterRoleBindingLister` into the `NewServiceAccount` func.  If no API group is found, just leave the item action's `clusterRoleBindings` field as nil, and return the action/no error.  Then `Execute()` becomes a no-op (more or less) in that scenario.

skriss: Sorry I didn't notice this earlier

nrb: np, better now than after tagging!

pkg/backup/rbac.go
```diff
@@ -0,0 +1,143 @@
...
+	var crbs []ClusterRoleBinding
+	for _, crb := range crbList.Items {
+		crbs = append(crbs, v1beta1clusterrolebinding{crb: crb})
+	}
+
+	return crbs, nil
+}
+
+// NewClusterRoleBindingLister wraps the v1beta1 and v1 RBAC client's List with a common interface.
+func NewClusterRoleBindingLister(log logrus.FieldLogger, clientset kubernetes.Interface, discoveryHelper arkdiscovery.Helper) (ClusterRoleBindingLister, error) {
```
skriss: We could change the signature here so instead of taking a `clientset`, it takes a `map[string]ListerConstructor`, where the keys are RBAC API versions, and the values are functions that return a `ClusterRoleBindingLister`. Then, instead of having a switch statement on concrete versions, we just do a lookup into the map for the server's preferred version.

This logic would be easily testable without actually requiring a `kubernetes.Interface`, and the correctness of the map (which would be defined in `plugin.go` could be determined by inspection.

Food for thought

nrb: I like that idea - are you thinking of pairing that with a no-op lister then?

skriss: Yeah, I think that'd be the way to go

nrb: Sketching this out, it might look something like

```
// In rbac.go
type ListerConstructor func() ClusterRoleBindingLister
NewV1clusterrolebindinglister()
NewV1beta1clusterrolebindinglister()
NewRBACDisabledLister

// In plugins.go
listerFuncs := map[string]ListerConstructor {
rbac.SchemeGroupVersion.Version: NewV1clusterrolebindinglister,
rbacbeta.SchemeGroupVersion.Version: NewV1betaclusterrolebindinglister,
"": NewRBACDisabledLister // returns an empty list
}

action, err := NewServiceAccountAction(log, discoveryHelper, listerFuncs)

// In service_item_action.go/NewServiceAccountAction
// apigroup logic moved from NewClusterRoleBindingLister

crbLister := listerFuncs[apigroup.Version]()
crbs, err := crbLister.List()
if err != nil {
 return nil, err
}
return &serviceAccountAction{
log: log,
clusterRoleBindings: crbs,
```

I think it works; the listerFuncs map may also be specified in rbac.go and exported, looking at it here. WDYT?

skriss: yep, looks good!

pkg/backup/service_account_action.go
```diff
@@ -25,28 +25,41 @@ import (
...
 type serviceAccountAction struct {
 	log                 logrus.FieldLogger
-	clusterRoleBindings []rbac.ClusterRoleBinding
+	clusterRoleBindings []ClusterRoleBinding
 }
 
 // NewServiceAccountAction creates a new ItemAction for service accounts.
-func NewServiceAccountAction(log logrus.FieldLogger, client rbacclient.ClusterRoleBindingInterface) (ItemAction, error) {
-	clusterRoleBindings, err := client.List(metav1.ListOptions{})
+func NewServiceAccountAction(log logrus.FieldLogger, clusterRoleBindingListers map[string]ClusterRoleBindingLister, discoveryHelper arkdiscovery.Helper) (ItemAction, error) {
```
skriss: It would be nice to have a unit test on this now that it doesn't depend a full clientset.  You could fake/mock a couple of `ClusterRoleBindingListers` to return a specific list of CRBs each, put them in the `clusterRoleBindingListers` arg, then set up a fake `discoveryHelper` with the relevant API groups that correspond to what's in the map (easiest way to set up the fake `discoveryHelper` is probably by not using the constructor function, but just instantiating directly and setting the `apiGroups` to what you need).  Then you can verify that the service account action's `clusterRoleBindings` match what the expected lister had.

pkg/backup/rbac.go
```diff
@@ -0,0 +1,153 @@
...
+		crbs = append(crbs, v1beta1clusterrolebinding{crb: crb})
+	}
+
+	return crbs, nil
+}
+
+// NewClusterRoleBindingListerMap creates a map of RBAC version strings to their associated
+// ClusterRoleBindingLister structs.
+// Necessary so that callers to the ClusterRoleBindingLister interfaces don't need the kubernetes.Interface.
+func NewClusterRoleBindingListerMap(clientset kubernetes.Interface) map[string]ClusterRoleBindingLister {
```
skriss: I think it would be slightly better if this returned a `map[string]ListerConstructor`, so that you don't have to construct all of them upfront. This way, only the one that's actually selected is instantiated.  I would just inline the constructors as anonymous functions/closures of the `clientset` argument.

pkg/backup/rbac.go
```diff
@@ -0,0 +1,153 @@
...
+	var crbs []ClusterRoleBinding
+	for _, crb := range crbList.Items {
+		crbs = append(crbs, v1clusterrolebinding{crb: crb})
+	}
+
+	return crbs, nil
+}
+
+// NewV1beta1ClusterRoleBindingLister returns a ClusterRoleBindingLister for RBAC v1beta1.
+func NewV1beta1ClusterRoleBindingLister() ClusterRoleBindingLister {
```
skriss: this should be removed

pkg/backup/rbac.go
```diff
@@ -0,0 +1,153 @@
...
+// noopClusterRoleBindingLister exists to handle clusters where RBAC is disabled.
+type noopClusterRoleBindingLister struct {
+}
+
+func (noop noopClusterRoleBindingLister) List() ([]ClusterRoleBinding, error) {
+	return []ClusterRoleBinding{}, nil
+}
+
+// NewV1ClusterRoleBindingLister returns a ClusterRoleBindingLister for RBAC v1.
+func NewV1ClusterRoleBindingLister() ClusterRoleBindingLister {
```
skriss: this should be removed

pkg/backup/rbac.go
nrb: Is there a problem or cost with instantiating them all, given we have 3? To me, returning them directly simplifies reading the code and reduces the number of types we have to carry around.

skriss: ok, not a big deal.

pkg/backup/rbac.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+// ClusterRoleBindingLister allows for listing ClusterRoleBindings in a version-independent way.
+type ClusterRoleBindingLister interface {
+	// List returns a slice of ClusterRoleBindings which can represent either v1 or v1beta1 ClusterRoleBindings.
+	List() ([]ClusterRoleBinding, error)
+}
+
+// ListerConstructor specifies a function which can return a ClusterRoleBindingLister.
+type ListerConstructor func() ClusterRoleBindingLister
```
skriss: can remove this since it's not being used

nrb: Done

skriss:
@nrb all set to merge?

nrb:
Yep, ready to merge from my POV. @carlisia do you want to do another pass?

skriss:
I'm gonna go ahead and merge so we can get the patch release out. Can always follow up with additional PRs if necessary,

PR #683, opened 2018-07-18
exit server if not all Ark CRDs exist at startup

skriss:

### DIFF
--- a/pkg/apis/ark/v1/register.go
+++ b/pkg/apis/ark/v1/register.go
@@ -41,27 +41,41 @@ func Resource(resource string) schema.GroupResource {
 	return SchemeGroupVersion.WithResource(resource).GroupResource()
 }
 
+type typeInfo struct {
+	PluralName   string
+	ItemType     runtime.Object
+	ItemListType runtime.Object
+}
+
+func newTypeInfo(pluralName string, itemType, itemListType runtime.Object) typeInfo {
+	return typeInfo{
+		PluralName:   pluralName,
+		ItemType:     itemType,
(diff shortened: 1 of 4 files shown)


pkg/apis/ark/v1/register.go
```diff
@@ -41,27 +41,32 @@ func Resource(resource string) schema.GroupResource {
...
 }
 
+type typeInfo struct {
+	ItemType     runtime.Object
+	ItemListType runtime.Object
+}
+
+// CustomResources returns a map of all custom resources within the Ark
+// API group, keyed on Kind.
+func CustomResources() map[string]typeInfo {
```
nrb: Would it make sense to re-use https://github.com/heptio/ark/blob/master/pkg/install/crd.go here? That way we'd have them defined in one place.

skriss: I'd rather go the other way around (have `pkg/install` use the `pkg/apis/ark/v1` func), but yeah good point, I'll update the install package.

nrb: Yeah, that direction does make more sense. If we use one common source then we're less likely to miss requiring any new CRDs getting added.

skriss: updated.

skriss:
Working on adding a unit test for the server func.

skriss:
added unit.

skriss:
Manual testing looks good on this - if any CRD is missing, the server exits.  If all are present, the server proceeds.

carlisia:
I 👀 👍

PR #687, opened 2018-07-19
fix delete issues

### LINKED ISSUE #685: Ark making a lot of delete backup requests
We're seeing an issue where a snapshot may be deleted by hand or by ark itself and the backup just stays in the deleting status. This causes ark to make a delete backup request then fail to remove it because the corresponding snapshot doesn't exist anymore. This continues to happen until there are thousands of deletebackuprequests. Is there a way to force delete backups? Or a code change to continue deleting the backup even if the snapshot doesn't exist?

`Deletion Attempts (50 failed):
  2018-07-18 14:28:24 -0400 EDT: Processed
  Errors:
    error deleting snapshot snap-088dc54a0627c1c41: rpc error: code = Unknown desc = InvalidSnapshot.NotFound: The snapshot 'snap-088dc54a0627c1c41' does not exist.
  status code: 400, request id: e2328806-9d25-494d-b043-9b57b91921e5
    error deleting snapshot snap-0bbb439b4aeb62e3f: rpc error: code = Unknown desc = InvalidSnapshot.NotFound: The snapshot 'snap-0bbb439b4aeb62e3f' does not exist.
  status code: 400, request id: cd4b57d5-8a79-461c-a502-79bcc3423705`

skriss:
Fixes #685 

Ready for review, but needs some more testing.

### DIFF
--- a/pkg/cloudprovider/aws/block_store.go
+++ b/pkg/cloudprovider/aws/block_store.go
@@ -20,6 +20,7 @@ import (
 	"regexp"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
@@ -219,7 +220,17 @@ func (b *blockStore) DeleteSnapshot(snapshotID string) error {
 
 	_, err := b.ec2.DeleteSnapshot(req)
 
-	return errors.WithStack(err)
+	// if it's a NotFound error, we don't need to return an error
+	// since the snapshot is not there.
+	// see https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html
+	if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "InvalidSnapshot.NotFound" {
+		return nil
+	}
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	return nil
 }
 
 var ebsVolumeIDRegex = regexp.MustCompile("vol-.*")
--- a/pkg/cloudprovider/azure/block_store.go
+++ b/pkg/cloudprovider/azure/block_store.go
@@ -19,6 +19,7 @@ package azure
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"os"
 	"regexp"
 	"strings"
@@ -280,7 +281,16 @@ func (b *blockStore) DeleteSnapshot(snapshotID string) error {
 
 	err = <-errChan
 
-	return errors.WithStack(err)
+	// if it's a 404 (not found) error, we don't need to return an error
+	// since the snapshot is not there.
+	if azureErr, ok := err.(autorest.DetailedError); ok && azureErr.StatusCode == http.StatusNotFound {
+		return nil
+	}
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	return nil
 }
 
 func getComputeResourceName(subscription, resourceGroup, resource, name string) string {
--- a/pkg/cloudprovider/gcp/block_store.go
+++ b/pkg/cloudprovider/gcp/block_store.go
@@ -19,6 +19,7 @@ package gcp
 import (
 	"encoding/json"
 	"io/ioutil"
+	"net/http"
 	"os"
 
 	"github.com/pkg/errors"
@@ -27,6 +28,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
 
 	"k8s.io/apimachinery/pkg/runtime"
 
@@ -212,7 +214,16 @@ func getSnapshotTags(arkTags map[string]string, diskDescription string, log logr
 func (b *blockStore) DeleteSnapshot(snapshotID string) error {
 	_, err := b.gce.Snapshots.Delete(b.project, snapshotID).Do()
 
-	return errors.WithStack(err)
+	// if it's a 404 (not found) error, we don't need to return an error
+	// since the snapshot is not there.
+	if gcpErr, ok := err.(*googleapi.Error); ok && gcpErr.Code == http.StatusNotFound {
+		return nil
+	}
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	return nil
 }
 
 func (b *blockStore) GetVolumeID(pv runtime.Unstructured) (string, error) {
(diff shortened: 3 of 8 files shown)


pkg/controller/gc_controller.go
```diff
@@ -130,12 +137,43 @@ func (c *gcController) processQueueItem(key string) error {
...
+		arkv1api.BackupUIDLabel:  string(backup.UID),
+	}))
 
-	_, err = c.deleteBackupRequestClient.DeleteBackupRequests(ns).Create(req)
+	dbrs, err := c.deleteBackupRequestLister.DeleteBackupRequests(ns).List(selector)
 	if err != nil {
+		return errors.Wrap(err, "error listing existing DeleteBackupRequests for backup")
+	}
+
+	// if there's an existing unprocessed deletion request for this backup, don't create
```
nrb: Do we want to keep the current one, or delete it and create a new one?

nrb: I'm not sure if the cases where we've been seeing a flood of DBRs have actually been processed; they're getting generated and not processed, which I think is the problem.

I requested a copy of a DBR from one of the failing clusters to double check this.

nrb: Looks like I'm incorrect; the duplicates being created do end up being processed.

skriss: The case where they weren't processed was because the controller hadn't started because of the missing CRDs, which has been addressed. Additionally, this change won't create any additional DBRs if an unprocessed one exists.

pkg/controller/gc_controller.go
```diff
@@ -130,12 +137,43 @@ func (c *gcController) processQueueItem(key string) error {
...
+	for _, dbr := range dbrs {
+		switch dbr.Status.Phase {
+		case "", arkv1api.DeleteBackupRequestPhaseNew, arkv1api.DeleteBackupRequestPhaseInProgress:
+			log.Info("Backup already has a pending deletion request")
+			return nil
+		}
+	}
+
+	// otherwise, any deletion requests that exist have already been processed (and must have
+	// failed), so delete them all
```
ncdc: I have a feeling it might be better to handle this in the backup deletion controller - we need to cap the manual creation of DBRs at 1 also. WDYT?

skriss: "this" being just the part that deletes existing processed backups?  Yeah, that seems reasonable to move over.  I wasn't too worried about manual DBRs but may as well handle them the same.

I'd leave the section that checks for an existing unprocessed request, and doesn't create a new one if there's already one, in the GC controller - that make sense?

ncdc: > "this" being just the part that deletes existing processed backups? Yeah, that seems reasonable to move over. I wasn't too worried about manual DBRs but may as well handle them the same.

Correct. I'm thinking in the backup deletion controller, when we process a DBR, we list all the DBRs for the associated backup, and delete all except the one being processed.

> I'd leave the section that checks for an existing unprocessed request, and doesn't create a new one if there's already one, in the GC controller - that make sense? Obv putting it as-is in the DBR controller wouldn't work.

Agreed, leave this in GC

skriss: I guess the other question is, if deleting existing DBRs in the DBR controller, what should we do with new/in-progress DBRs for the same backup?  I suppose we could just delete them - given that we're currently only running one worker, anything with a phase of In Progress that's not the current DBR isn't *actually* being processed

ncdc: Let's imagine we run multiple workers. Workers 1 and 2 each get a DBR for the same backup (2 different DBRs). There are a couple of ways this could play out:

1. Worker 1 gets DBR 1
1. Worker 1 lists all DBRs for the backup and sees 0 InProgress
1. Worker 1 sets status to InProgress
1. Worker 2 gets DBR 2
1. Worker 2 lists all DBRs for the backup and sees 1 InProgress
1. Worker 2 deletes DBR 2 because there's another DBR In Progress

or

1. Worker 1 gets DBR 1
1. Worker 2 gets DBR 2
1. Worker 1 lists all DBRs and sees none InProgress so it sets DBR 1 to InProgress
1. Worker 2 lists all DBRs and sees none InProgress (informer cache hasn't seen DBR 1 InProgress update yet) so it sets DBR 2 to InProgress
1. Now we have 2 workers trying to delete the same backup :-(

In both scenarios, we could instead use a shared "in progress DBRs" data structure to ensure there's never more than 1 InProgress at a time.

skriss:
Added a new commit, which I'll squash, to move the deletion of existing DBRs into the DBR controller.  Units and basic manual validation pass.

pkg/cloudprovider/aws/block_store.go
```diff
@@ -218,6 +219,18 @@ func (b *blockStore) DeleteSnapshot(snapshotID string) error {
...
+	}
+
+	if awsErr, ok := err.(awserr.Error); ok {
+		// if it's a NotFound error, we don't need to return an error
+		// since the snapshot is not there.
+		// see https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html
+		if awsErr.Code() == "InvalidSnapshot.NotFound" {
+			return nil
+		}
+	}
```
carlisia: A bit of a nit comment incoming :)

I would nest the code in line 226 inside a `if err != nil`  and after that, still inside that if, return `errors.WithStack(err)`, so that the very last line of this func would be `return nil`.  This mainly because it causes least surprise but the reasoning I think is that it's more important to read the error conditions so if all of that come first/at the top it is easier to read/find.

pkg/cloudprovider/azure/block_store.go
```diff
@@ -279,6 +280,17 @@ func (b *blockStore) DeleteSnapshot(snapshotID string) error {
 	_, errChan := b.snaps.Delete(snapshotInfo.resourceGroup, snapshotInfo.name, ctx.Done())
 
 	err = <-errChan
+	if err == nil {
```
carlisia: I would do the same nesting here and return `nil` at the very end.

pkg/cloudprovider/gcp/block_store.go
```diff
@@ -211,6 +213,17 @@ func getSnapshotTags(arkTags map[string]string, diskDescription string, log logr
 
 func (b *blockStore) DeleteSnapshot(snapshotID string) error {
 	_, err := b.gce.Snapshots.Delete(b.project, snapshotID).Do()
+	if err == nil {
```
carlisia: Same here.

skriss:
Alright, I've been able to manually test this on all three public clouds and looking good so far.

skriss:
@carlisia thx for the review, addressed in a new commit which i'll squash before merge.

carlisia:
@skriss gladly! Just of note: did you mean to remove the `if it's a NotFound error, we don't need to return an error`?

carlisia:
Argh! You are right! Btw, reads a ton better this way. 👍

skriss:
I've tested and squashed, so ready for final review, merge when looks good.

carlisia:
👍

PR #692, opened 2018-07-20
Clarify that object storage should per-cluster

metadave:
After reading through the AWS install guide, the first thing I searched the docs for was whether a bucket should be shared between clusters. I found the answer in the FAQ, but thought it might be good to have right up front in the install docs.

### DIFF
--- a/docs/azure-config.md
+++ b/docs/azure-config.md
@@ -49,9 +49,12 @@ az storage account create \
     --https-only true \
     --kind BlobStorage \
     --access-tier Hot
+```
 
-# Create the blob container named "ark". Feel free to use a different name; you'll need to
-# adjust the `bucket` field under `backupStorageProvider` in the Ark Config accordingly if you do.
+Create the blob container named `ark`. Feel free to use a different name, preferrably unique to a single Kubernetes cluster. See the [FAQ][20] for more details. You'll need to
+adjust the `bucket` field under `backupStorageProvider` in the Ark Config accordingly if you do.
+
+```bash
 az storage container create -n ark --public-access off --account-name $AZURE_STORAGE_ACCOUNT_ID
 
 # Obtain the storage access key for the storage account just created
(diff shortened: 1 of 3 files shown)


skriss:
Thanks for this @metadave !

docs/azure-config.md
```diff
@@ -50,7 +50,8 @@ az storage account create \
     --kind BlobStorage \
     --access-tier Hot
 
-# Create the blob container named "ark". Feel free to use a different name; you'll need to
+# Create the blob container named "ark". Feel free to use a different name, preferrably unique to a 
+# single Kubernetes cluster. See the FAQ for more details. You'll need to
```
skriss: could you add a link to the FAQ here like on the other pages?

skriss:
It'd be great if you could also add the same note about using a unique name for the service account/principal to the Azure page, around this section:

https://github.com/heptio/ark/pull/692/files#diff-24de5554c14da06519d0c8d39ff4abe5R94

docs/azure-config.md
metadave: I split the section up so Markdown would render the FAQ link, not sure if that's desirable on your end though.

metadave:
fixes pushed, thanks!

docs/gcp-config.md
```diff
@@ -34,6 +34,8 @@ To integrate Heptio Ark with GCP, create an Ark-specific [Service Account][15]:
         --display-name "Heptio Ark service account"
     ```
 
+> If you'll be using Ark to backup multiple clusters with multiple GCS buckets, it may be desirable to create a unique username per cluster rather than the default `heptio-ark`.
+
     Then list all accounts and find the `heptio-ark` account you just created:
```
skriss: The formatting starting here is messed up (it's not immediately apparent to me why) -- see https://github.com/metadave/ark/blob/c2d65ae0bd35154f5f7bb6daf493824393917e97/docs/gcp-config.md#create-service-account -- "Then list ... its `email` value" is all within a code block.

docs/azure-config.md
```diff
@@ -92,6 +95,8 @@ To integrate Ark with Azure, you must create an Ark-specific [service principal]
 
 1. Create a service principal with `Contributor` role. This will have subscription-wide access, so protect this credential. You can specify a password or let the `az ad sp create-for-rbac` command create one for you.
 
+> If you'll be using Ark to backup multiple clusters with multiple blob containers, it may be desirable to create a unique username per cluster rather than the default `heptio-ark`.
+
     ```bash
```
skriss: This section's formatting is also off - a literal "```bash" string is showing up inside the code block

docs/gcp-config.md
metadave: I needed to tab the `>` over.

skriss:
LGTM, thanks!

PR #690, opened 2018-07-20
fix tagging latest by using make's ifeq

skriss:
Signed-off-by: Steve Kriss <steve@heptio.com>

Ref. https://github.com/heptio/ksonnet-website/pull/26/files

PR #695, opened 2018-07-23
cherry-pick commits for v0.9.1

skriss:
##### Bug Fixes:
  * Require namespace for Ark's CRDs to already exist at server startup (#676, @skriss)
  * Require all Ark CRDs to exist at server startup (#683, @skriss)
  * Fix `latest` tagging in Makefile (#690, @skriss)
  * Make Ark compatible with clusters that don't have the `rbac.authorization.k8s.io/v1` API group (#682, @nrb)
  * Don't consider missing snapshots an error during backup deletion, limit backup deletion requests per backup to 1 (#687, @skriss)

### DIFF
--- a/pkg/cloudprovider/gcp/block_store.go
+++ b/pkg/cloudprovider/gcp/block_store.go
@@ -19,6 +19,7 @@ package gcp
 import (
 	"encoding/json"
 	"io/ioutil"
+	"net/http"
 	"os"
 
 	"github.com/pkg/errors"
@@ -27,6 +28,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
 
 	"k8s.io/apimachinery/pkg/runtime"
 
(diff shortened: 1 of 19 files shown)


pkg/cloudprovider/gcp/block_store.go
```diff
@@ -212,7 +214,16 @@ func getSnapshotTags(arkTags map[string]string, diskDescription string, log logr
 func (b *blockStore) DeleteSnapshot(snapshotID string) error {
 	_, err := b.gce.Snapshots.Delete(b.project, snapshotID).Do()
 
-	return errors.WithStack(err)
+	// if it's a 404 (not found) error, we don't need to return an error
+	// since the snapshot is not there.
+	if gcpErr, ok := err.(*googleapi.Error); ok && gcpErr.Code == http.StatusNotFound {
+		return nil
+	}
+	if err != nil {
```
carlisia: This is super nit, might want to ignore: it'd read better if the flow of handling error cases here was err, err nil, as opposed to nil, err, nil.

skriss: OK, not going to address on this PR since this is a backport of existing master commits, but we can discuss separately :)

carlisia: Fair enough!

PR #697, opened 2018-07-23
check s3URL scheme upon AWS ObjectStore Init()

metadave:
fixes https://github.com/heptio/ark/issues/641

### DIFF
--- a/pkg/cloudprovider/aws/helpers.go
+++ b/pkg/cloudprovider/aws/helpers.go
@@ -18,6 +18,7 @@ package aws
 
 import (
 	"context"
+	"net/url"
 
 	"github.com/pkg/errors"
 
@@ -50,4 +51,17 @@ func GetBucketRegion(bucket string) (string, error) {
 	}
 
 	return "", errors.New("unable to determine bucket's region")
+}
+
+// IsValidS3URLScheme returns true if the scheme is http:// or https://
+// and the url parses correctly, otherwise, return false
+func IsValidS3URLScheme(s3URL string) bool {
+	u, err := url.Parse(s3URL)
+	if err != nil {
+		return false
+	}
+	if u.Scheme != "http" && u.Scheme != "https" {
+		return false
+	}
+	return true
 }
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/aws/object_store.go
```diff
@@ -85,6 +85,10 @@ func (o *objectStore) Init(config map[string]string) error {
 		WithS3ForcePathStyle(s3ForcePathStyle)
 
 	if s3URL != "" {
+		if !IsValidS3URLScheme(s3URL) {
```
skriss: I think we could make this a little bit more robust by using `net/url`'s `Parse` function, which will return an error if it's not a valid URL, and then we can check the returned struct's `Scheme` property to explicitly ensure it's http/https.  WDYT?

metadave: I agree! I'll push a change shortly.

skriss:
LGTM - please squash before we merge.  Thanks for the contribution!  

cc @nrb @carlisia for review/merge.

pkg/cloudprovider/aws/helpers.go
```diff
@@ -51,3 +52,17 @@ func GetBucketRegion(bucket string) (string, error) {
...
+// IsValidS3URLScheme returns true if the scheme is http:// or https://
+// and the url parses correctly, otherwise, return false
+func IsValidS3URLScheme(s3URL string) bool {
+	u, err := url.Parse(s3URL)
+	if err != nil {
+		return false
+	}
+	if u.Scheme == "http" || u.Scheme == "https" {
+		return true
+	} else {
```
carlisia: A couple comments minor in consequence but with the potential to improve readability:
1. The `else` here is unnecessary.
2. I also propose to switch and say `if u.Scheme != "http" && u.Scheme != "https" ` then return `false`, lastly return `true`.

metadave: that's a good suggestion, thanks!

skriss:
Thanks again @metadave!

PR #704, opened 2018-07-26
fix bug preventing backup item action item updates from saving

### LINKED ISSUE #702: backup item action changes to objects aren't being persisted in the backup tarball

1. build sample plugin container image
2. `ark plugin add <my-repo>/ark-plugin-example:latest`
3. `ark backup create foo`

The backed-up items should be annotated, per https://github.com/heptio/ark-plugin-example/blob/master/ark-backupitemaction-log-and-annotate/myplugin.go#L53. They are not.

It looks like https://github.com/heptio/ark/blob/82f1cd87dcc70965ef093872daaca92685dda90c/pkg/backup/item_backupper.go#L312 is to blame, as a pointer is being updated to point to a new location within a helper func, but this only lasts for the scope of the function, so the parent function which actually writes the item to the backup still has a pointer to the original obj.

skriss:

Fixes #702

### DIFF
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -204,10 +204,20 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 		}
 	}
 
-	if err := ib.executeActions(log, obj, groupResource, name, namespace, metadata); err != nil {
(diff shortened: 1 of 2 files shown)


skriss:
Confirmed manually that this fixes the observed issue.

PR #708, opened 2018-07-26
Only delete unused backup if they are complete

### LINKED ISSUE #705: sync controller: don't delete API objects that are new/in-progress/etc.
backups with `.spec.phase` of "", New, InProgress should be excluded from the `deleteUnused` func.

backups with `.spec.phase` of FailedValidation -- I think it makes sense to exclude them as well.  This ensure we don't delete the API obj before a user can see why it failed validation. It will still be GC'ed according to its TTL.  We could possibly look at GC'ing backups that failed validation more quickly, but I don't think it's important for now.

backups with `.spec.phase` of Failed - these **should** be in object storage (just metadata + logs) -- so the existing code won't delete them, unless the upload of metadata to object storage fails. I'm tempted to exclude these from `deleteUnused` as well, as it could cause confusion for the user if a backup fails, uploading metadata/logs fail, and then the backup API obj just disappears.

So I think the net result of this is `deleteUnused` should *only* process `Completed` backups.

@carlisia WDYT?

carlisia:
Fixes #705

### DIFF
--- a/pkg/controller/backup_sync_controller.go
+++ b/pkg/controller/backup_sync_controller.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 
+	api "github.com/heptio/ark/pkg/apis/ark/v1"
 	"github.com/heptio/ark/pkg/cloudprovider"
 	arkv1client "github.com/heptio/ark/pkg/generated/clientset/versioned/typed/ark/v1"
 	informers "github.com/heptio/ark/pkg/generated/informers/externalversions/ark/v1"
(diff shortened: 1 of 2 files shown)


pkg/controller/backup_sync_controller.go
```diff
@@ -141,9 +143,10 @@ func (c *backupSyncController) deleteUnused(cloudBackupNames sets.String) {
 		return
 	}
 
-	// For each backup object in Kubernetes, verify if has a corresponding backup in the object storage. If not, delete it.
+	// For each backup object in Kubernetes, verify if it is complete and
```
skriss: this comment is slightly confusing/misleading. sounds like all non-completed backups get deleted.  I'd say `for each completed backup object in Kubernetes, delete it if it does not have a corresponding backup in object storage`

skriss:
otherwise lgtm.

pkg/controller/backup_sync_controller.go
carlisia: Fixing..

PR #709, opened 2018-07-26
update testify to latest released version

skriss:


skriss:
Can someone merge this if you don't mind? Doing some heavy-duty rebasing & cleanup with the plugins PR and it would be helpful to have this in. thanks :)

PR #711, opened 2018-07-27
Bring back 'make local'

skriss:

I pulled this out of #495, creating a PR so we don't lose track of it.

I'd like to see the goarch detection move to only happen in make local, bc this is going to break `make container` when running on macOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,9 @@ PKG := github.com/heptio/ark
 REGISTRY ?= gcr.io/heptio-images
 
 # Which architecture to build - see $(ALL_ARCH) for options.
+# if the 'local' rule is being run, detect the ARCH from 'go env'
+# if it wasn't specified by the caller.
+local : ARCH ?= $(shell go env GOOS)-$(shell go env GOARCH)
 ARCH ?= linux-amd64
 
 VERSION ?= master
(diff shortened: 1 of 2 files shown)


ncdc:
Thanks for splitting it out, and I agree

nrb:
LGTM.

May want to update https://github.com/heptio/ark/blob/master/docs/build-from-scratch.md#cross-compiling to reference this target.

skriss:
not ready to merge

skriss:
@nrb which part are you saying should be updated in the cross-compiling section?

skriss:
@ncdc PTAL if you get a chance, I think this addresses the comment re: only detecting local arch when running `make local`

Makefile
```diff
@@ -79,6 +82,15 @@ all-build: $(addprefix build-, $(CLI_PLATFORMS))
 
 #all-push: $(addprefix push-, $(CONTAINER_PLATFORMS))
 
+local: build-dirs
+	GOOS=$(GOOS) \
```
ncdc: I _think_ we can omit specifying all the variables that are unmodified and `make` should pass them through when executing `build.sh`. That would leave `OUTPUT_DIR` as the only one we need. Try that and see if it works?

skriss: nope :(

nrb:
@skriss I was thinking this text: `By default, make builds an ark binary that runs on your host operating system and architecture.`

My Makefile knowledge isn't great, but isn't the point of this PR that the default target `all` doesn't actually do that? `local` will do the detection, and `all` will build `linux-amd64` all the time.

Am I getting that right?

Makefile
ncdc: Aha, we could `export` whatever variables we want to be passed down to *all* shell invocations from all targets, if desired.

skriss: Eh, maybe in a separate PR :)

skriss:
Ah, I see what you mean. Yeah, that sentence isn't really true in the current state. Let me see if I can do some rewording there.

skriss:
OK after re-reading that section + the Makefile, I made an edit to reflect existing reality better, but I'm not mentioning `make local` there because I think `make build` is the appropriate target to use.

skriss:
@carlisia PTAL, merge if +1.

Issue #578, opened 2018-06-25
RFE: ark bug command


rosskukulinski:
Inspired by the [traefik bug](https://docs.traefik.io/basics/#command-bug) command, it would be great to have an `ark bug` command that would open a GitHub issue with pre-populated environment details.

For reference, I recommend watching [this youtube video](https://www.youtube.com/watch?v=Lyz62L8m93I) of their command in action.

![screen shot 2018-06-24 at 11 06 31 pm](https://user-images.githubusercontent.com/2746479/41828453-46260eca-7803-11e8-9314-379c32e07993.png)

PR #713, opened 2018-07-27
add an ark bug command

metadave:
Closes https://github.com/heptio/ark/issues/578

### DIFF
--- a/pkg/cmd/cli/bug/bug.go
+++ b/pkg/cmd/cli/bug/bug.go
@@ -1,0 +1,144 @@
+/*
+Copyright 2018 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bug
+
+import (
+	"bytes"
+	"fmt"
+	"net/url"
+	"os/exec"
+	"runtime"
+	"text/template"
+
+	"github.com/heptio/ark/pkg/buildinfo"
+	"github.com/heptio/ark/pkg/cmd"
+	"github.com/spf13/cobra"
+)
+
+const (
+	linuxBrowserLaunchCmd = "xdg-open"
+	issueURL              = "https://github.com/heptio/ark/issues/new"
+	issueTemplate         = `
+**What steps did you take and what happened:**
+[A clear and concise description of what the bug is, and what commands you ran.)
+
+
+**What did you expect to happen:**
+
+
+**The output of the following commands will help us better understand what's going on**:
+(Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)
+
+* ` + "`kubectl logs deployment/ark -n heptio-ark`" + `
+* ` + "`ark backup describe <backupname>` or `kubectl get backup/<backupname> -n heptio-ark -o yaml`" + `
+* ` + "`ark backup logs <backupname>`" + `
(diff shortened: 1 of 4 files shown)


ncdc:
FYI you have to write `Fixes #578` or `Closes #578` to have the issue automatically closed when this is merged.

metadave:
Looks like I missed some docs, I'll update shortly.

pkg/cmd/cli/bug/bug.go
```diff
@@ -0,0 +1,144 @@
...
+**Anything else you would like to add:**
+[Miscellaneous information that will assist in solving the issue.]
+
+
+**Environment:**
+
+- Ark version (use ` + "`ark version`" + `): {{.ArkVersion}}
+    - GitCommit: {{.GitCommit}}
+    - GitTreeState: {{.GitTreeState}}
+- Kubernetes version (use ` + "`kubectl version`" + `):
```
nrb: Any thoughts about running `kubectl version` and including it automatically here? I don't think it'd required, but I think it would make filing even easier.

metadave: I think it makes sense to include, I wasn't sure if it was easier to use `kubectl` or the api.

nrb: That's a good point. Ideally we'd have both the `kubectl` version and the server version, I think, so calling the executable makes sense to me. Open to other ideas here, though.

pkg/cmd/cli/bug/bug.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"github.com/heptio/ark/pkg/buildinfo"
+	"github.com/heptio/ark/pkg/cmd"
+	"github.com/spf13/cobra"
+)
+
+const (
+	linuxBrowserLaunchCmd = "xdg-open"
+	issueURL              = "https://github.com/heptio/ark/issues/new"
+	issueTemplate         = `
```
ncdc: Is there any way to avoid having the template hard-coded here? We already have it in `.github/ISSUE_TEMPLATE/bug_report.md`, but I realize that's not (at the moment) necessarily easy to fill in programatically.

metadave: there's also https://github.com/GeertJohan/go.rice, but it seems a bit weird to me.

pkg/cmd/cli/bug/bug.go
metadave: `kubectl version` output added

pkg/cmd/cli/bug/bug.go
metadave: Embedding the template in go does make a bit of copy-pasta. Could `.github/ISSUE_TEMPLATE/bug_report.md` be generated from a go template with empty variables somewhere in the build process, with the same template being loaded/downloaded + populated in `ark bug`?

nrb: I'm not sure how well that would work - as it stands, I think GitHub uses that file directly, without any pre-processors. I think we'd like to keep that integration working.

I do suppose it's possible to use a command that updates it and we commit the result, similar to how we do the cli-reference docs.

@jhamilton1 Would having a command that dumps contents of a Go variable into `.github/ISSUE_TEMPLATES/bug_report.md` be workable for your usage? It would mean updates to the text would happen inside a Go file, but the format should still be Markdown.

metadave: > I do suppose it's possible to use a command that updates it and we commit the result, similar to how we do the cli-reference docs.

That's what I failed to articulate :-)

metadave: I think this would need:

- hack/update-generated-issue-template.sh
- hack/verify-generated-issue-template.sh
- a raw issue template
  - the issue "template template"
- a tiny program to render this template file (similar to https://github.com/heptio/ark/blob/master/docs/generate/ark.go)
- a generated issue template (`.github/ISSUE_TEMPLATE/bug_report.md`) made with the program above
- the `ark bug` command would need to download the raw template and populate it's fields to open a new issue on Github
- maybe more?

metadave:
I'm going to close this for now and rework a few things.

nrb:
@metadave Sure thing. Feel free to shout if you need anything from us.

PR #717, opened 2018-08-01
Extend doc about synchronization (gh-pages)

carlisia:
This is related to the feature improvement done here:

https://github.com/heptio/ark/pull/680

### DIFF
--- a/master/about.md
+++ b/master/about.md
@@ -64,6 +64,8 @@ Heptio Ark treats object storage as the source of truth. It continuously checks 
 
 This allows restore functionality to work in a cluster migration scenario, where the original backup objects do not exist in the new cluster.
 
+Likewise, if a backup object exists in Kubernetes but not in object storage, it will be deleted from Kubernetes since the backup tarball no longer exists.
+
 [10]: hooks.md
 [19]: /img/backup-process.png
 [20]: https://kubernetes.io/docs/concepts/api-extension/custom-resources/#customresourcedefinitions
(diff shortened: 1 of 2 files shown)


master/about.md
```diff
@@ -64,6 +64,8 @@ Heptio Ark treats object storage as the source of truth. It continuously checks
 
 This allows restore functionality to work in a cluster migration scenario, where the original backup objects do not exist in the new cluster.
 
+Likewise, if a backup object exists in Kubernetes but not in the object storage, it will be deleted from Kubernetes since this means it is no longer in use.
```
nrb: I would probably word the first part `Likewise, if a backup object exists in Kubernetes but not in object storage, ...`.

nrb:
One fairly minor comment, otherwise looks fine.

We'll also need a PR for master that adds this to the `docs/about.md` file in the `master` branch, so that docs for new versions include it.

master/about.md
carlisia: Ok cool. I like the attention to detail!

master/about.md
```diff
@@ -64,6 +64,8 @@ Heptio Ark treats object storage as the source of truth. It continuously checks
 
 This allows restore functionality to work in a cluster migration scenario, where the original backup objects do not exist in the new cluster.
 
+Likewise, if a backup object exists in Kubernetes but not in object storage, it will be deleted from Kubernetes since this means it is no longer in use.
```
skriss: instead of "...this means it is no longer in use", I'd say something like "...the backup tarball no longer exists."

PR #719, opened 2018-08-02
Document CRD not found errors

nrb:
Will create a PR for the gh-pages branch once this content gets merged.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -35,4 +35,9 @@ Related to this, if you need to restore a backup from cluster A into cluster B, 
 mode in cluster B's Ark instance while it's configured to use cluster A's bucket. This will ensure no 
 new backups are created, and no existing backups are deleted or overwritten.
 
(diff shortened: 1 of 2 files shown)


ncdc:
Can we also change the error in the log to make it clearer that Ark's CRDs need to be installed?

carlisia:
LGTM + Andy's comment. 👍

nrb:
I like updating the error message, too.

Right now, it's an aggregated error reporting all CRDs that don't exist. I could either add to the aggregated error list, or just return it. Any preferences?

carlisia:
@nrb I looked at the code and imo adding to the existing error list would be consistent with what's there already, and also useful to inform about all the missing CRDs at once, so this gets my vote.

nrb:
Returned errors updated.

carlisia:
Yes, I think this works well. 👍

skriss:
LGTM

PR #723, opened 2018-08-02
Document supported object and volume providers

nrb:

### DIFF
--- a/docs/support-matrix.md
+++ b/docs/support-matrix.md
@@ -1,0 +1,50 @@
+# Compatible Storage Providers
+
+Ark supports a variety of storage providers for different backup and snapshot operations. As of version 0.6.0, a plugin system allows anyone to add compatibility for additional backup and volume storage platforms without modifying the Ark codebase.
+
+## Backup Storage Providers
+
+| Provider                  | Owner    | Contact                         |
+|---------------------------|----------|---------------------------------|
+| [AWS S3][2]               | Ark Team | [Slack][10], [GitHub Issue][11] |
+| [Azure Blob Storage][3]   | Ark Team | [Slack][10], [GitHub Issue][11] |
+| [Google Cloud Storage][4] | Ark Team | [Slack][10], [GitHub Issue][11] |
+
+## S3-Compatible Backup Storage Providers
+
+Ark uses [Amazon's Go SDK][12] to connect to the S3 API. Some third-party storage providers also support the S3 API, and users have reported the following providers work with Ark:
+
+_Note that these providers are not regularly tested by the Ark team._
+
+ * [IBM Cloud][5]
+ * [Minio][9]
+ * Ceph RADOS v12.2.7
+
+## Volume Snapshot Providers
+| Provider                         | Owner           | Contact                         |
+|----------------------------------|-----------------|---------------------------------|
+| [AWS EBS][2]                     | Ark Team        | [Slack][10], [GitHub Issue][11] |
+| [Azure Managed Disks][3]         | Ark Team        | [Slack][10], [GitHub Issue][11] |
+| [Google Compute Engine Disks][4] | Ark Team        | [Slack][10], [GitHub Issue][11] |
+| [Restic][1]                      | Ark Team        | [Slack][10], [GitHub Issue][11] |
+| [Portworx][6]                    | PortWorx        |                                 |
+| [DigitalOcean][7]                | StackPointCloud |                                 |
+
+### Adding a new plugin
+
+To write a plugin for a new backup or volume storage system, take a look at the [example repo][8].
+
+After you publish your plugin, open a PR that adds your plugin to the appropriate list.
+
+[1]: restic.md
+[2]: aws-config.md
+[3]: azure-config.md
+[4]: gcp-config.md
+[5]: ibm-config.md
+[6]: https://docs.portworx.com/scheduler/kubernetes/ark.html
+[7]: https://github.com/StackPointCloud/ark-plugin-digitalocean
+[8]: https://github.com/heptio/ark-plugin-example/
+[9]: quickstart.md
+[10]: https://kubernetes.slack.com/messages/ark-dr
+[11]: https://github.com/heptio/ark/issues
+[12]: https://github.com/aws/aws-sdk-go/aws


nrb:
I haven't yet added this into the table of contents; @Bradamant3 do you have any guidance on where you think it might best fit?

Overall, looking for feedback on format and messaging.

docs/support-matrix.md
```diff
@@ -0,0 +1,43 @@
+# Supported Storage Providers
+
+Ark supports sending backup data to a variety of object storage providers, as well as volume snapshots using either a plugin for the volume provider, or [restic][4] for filesystem level backups.
+
+## In-tree providers
```
ncdc: I'm thinking a single matrix that includes in-tree and plugins would be better. Something that looks like https://github.com/kubernetes/client-go/#compatibility-matrix. We may need 2 matrices: 1 for object storage, and 1 for volumes.

docs/support-matrix.md
```diff
@@ -0,0 +1,43 @@
...
+| Portworx      | https://docs.portworx.com/scheduler/kubernetes/ark.html    | no                          | yes                     | v0.8.0 to v0.9.0       |
+| Digital Ocean | https://github.com/StackPointCloud/ark-plugin-digitalocean | yes (via S3-compatible API) | yes                     | v0.8.0 to v0.9.0       |
+
+### Adding a new plugin
+
+If you would like to write a plugin for another object or volume storage system, take a look at the [example repo][1].
+
+Then, open a PR adding the relevant information to the plugins table above.
+
+## Object Storage Supported via S3 API
```
ncdc: I would include this information in the object storage table

rosskukulinski:
Does it make sense for us to have a maturity indicator for plugins?  We should probably have something for the over-all project, but I wonder about plugins.  @carlisia suggested this which I quite like:

image 2

docs/support-matrix.md
```diff
@@ -0,0 +1,43 @@
...
+| Google Compute Engine | Google Cloud Storage | Google Compute block devices |
+| Azure                 | Azure Blob Storage   | Azure Compute block devices  |
+
+## Plugins
+
+Plugins provide Ark a mechanism to use object and volume storage from providers not already supported in tree.
+
+| Provider Name | Plugin Site                                                | Object Storage Provided     | Volume Storage Provided | Ark versions supported |
+|---------------|------------------------------------------------------------|-----------------------------|-------------------------|------------------------|
+| Portworx      | https://docs.portworx.com/scheduler/kubernetes/ark.html    | no                          | yes                     | v0.8.0 to v0.9.0       |
```
carlisia: I wonder if it wouldn't be best to have it as a guide to use the code repo for all plugins.

nrb: What if the plugin's code isn't public?

ncdc: I would link to the documentation. We could optionally link to the code.

carlisia: Ah I didn't think the plugin could be private. Good point.

docs/support-matrix.md
nrb: I want to make sure we're clear on what the Ark team will take responsibility for. Encoding too much information into columns can make the tabled width get out of hand, which is what I'd like to avoid.

I'll give it a shot and see how it looks, though.

nrb:
> Does it make sense for us to have a maturity indicator for plugins? 

That probably does make sense. I think it would be up to external developers to set their maturity level, though. Additionally, I'm not sure if we'd have separate indicators for our in-tree plugins until we had drift or we moved them out.

Happy to have more discussion on that, though.

docs/support-matrix.md
```diff
@@ -0,0 +1,43 @@
...
+| Digital Ocean | https://github.com/StackPointCloud/ark-plugin-digitalocean | yes (via S3-compatible API) | yes                     | v0.8.0 to v0.9.0       |
+
+### Adding a new plugin
+
+If you would like to write a plugin for another object or volume storage system, take a look at the [example repo][1].
+
+Then, open a PR adding the relevant information to the plugins table above.
```
skriss: Maybe instead of "Then," say "Once you've published your plugin,"

docs/support-matrix.md
```diff
@@ -0,0 +1,43 @@
+# Supported Storage Providers
+
+Ark supports sending backup data to a variety of object storage providers, as well as volume snapshots using either a plugin for the volume provider, or [restic][4] for filesystem level backups.
+
+## In-tree providers
+
+Ark supports the follow object and volume storage providers in-tree.
```
skriss: follow -> following

docs/support-matrix.md
```diff
@@ -0,0 +1,43 @@
...
+
+| Provider              | Object Storage       | Block Storage                |
+|-----------------------|----------------------|------------------------------|
+| Amazon Web Services   | S3                   | Elastic Block Store          |
+| Google Compute Engine | Google Cloud Storage | Google Compute block devices |
+| Azure                 | Azure Blob Storage   | Azure Compute block devices  |
+
+## Plugins
+
+Plugins provide Ark a mechanism to use object and volume storage from providers not already supported in tree.
```
skriss: I'd rather say something like "Ark's plugin architecture enables anyone to add support for additional object and block storage platforms without modifying the core Ark codebase.  The following third-party plugins are available for use:"

skriss:
added some comments. this is going to be really great to have. thanks for pulling it together @nrb !

docs/support-matrix.md
nrb: @ncdc Clarifying your suggestion - are you thinking of column headers being the versions of Ark?

ncdc: Brainstorming (data below not necessarily accurate):

Provider|Backup Storage|Volume Snapshots|Docs|Owner|Compatibility
-|-|-|-|-|-
AWS|✅|✅|[Setup Guide](todo)|Ark Team|All Ark versions
Azure|✅|✅ <sup>1</sup>|[Setup Guide](todo)|Ark Team|All Ark versions
Portworx|❌|✅ |[Setup Guide](https://docs.portworx.com/scheduler/kubernetes/ark.html)|PortWorx|Ark 0.8+

<sup>1</sup> Only Azure managed disks are supported

nrb:
I've updated this to be a single table that I think works fairly well.

@carlisia @rosskukulinski I've not added the maturity information because I have a question for the group - given that Ark is still not 1.0, are we classifying our own plugins as alpha/beta?

skriss:
Looks good to me! I'll review in a little more detail, but i like it.

carlisia:
@nrb I wonder if it would be cost effective to have some explicit guidelines. We could talk about it. For now, leaving this as it might be helpful: https://en.wikipedia.org/wiki/Software_release_life_cycle#Stages_of_development

skriss:
I wonder if we can just tackle the maturity thing once we hit 1.0.  Until then, everything's alpha/beta/whatever you want to call it

skriss:
I'd say Im happy with it as-is. One question - where did the compatibility versions come from re: portworx and digitalocean? are they not compatible with any post-plugins Ark release (v0.6.0)?

skriss:
Oh, one other thought. For external plugins we should add a "support contact" and make that a requirement for listing. Could be an email, a slack, a link to a GH issue, whatever. Anything so it doesn't get directed to core.

ncdc:
Should we get rid of the Documentation column and move those links to the text in the first column (provider name)?

nrb:
> where did the compatibility versions come from re: portworx and digitalocean? are they not compatible with any post-plugins Ark release (v0.6.0)?

I'll update that, though I haven't tested them that far back. We do know, however, that v0.10.0 will break compat and they'll need to update. I'm leaving it off the table since it's not released.

nrb:
I've moved the documentation links onto the provider names and included links to the ark-dr channel and GH issues page for contact. Using the Owner field may be awkward if there's more than one method, like we have. Or it could just point to GH issues, I'm not really picky here.

skriss:
What you have now LGTM

nrb:
@Bradamant3 Would you mind taking a look at this when you have a chance?

rosskukulinski:
@nrb maybe we for Ark team, we can just link to #troubleshooting in our main Readme?

Instead of "Contact", it could be "Support" or "Troubleshooting".

carlisia:
This table looks so awesome!

One comment: When I read `Not supported`, it made me think that it will not be supported. I know it doesn't mean anyone will think the same, but wanted to ask, should we say something like `Not supported yet` or `Not officially supported`?

rosskukulinski:
@carlisia Well in the case of Minio, it's _not_ a block-storage system at all, so it will _never_ support snapshots.

carlisia:
Hm. I guess you articulated for me what my reaction was about: should we distinguish between what's in that list that will never be supported vs not supported at the moment?

carlisia:
Oops. I was not reading that table correctly. NVM my comment!

nrb:
@carlisia That's partially why I started with separate tables - for example, OpenStack Swift is an object store. It does nothing with volumes/block devices and never will. However, that doesn't mean Ark will never support writing back ups to it - someone may write a plugin.

I don't want to encode too much information in symbols, but would it be clearer if we marked it as `n/a` for providers that only do one?

ncdc:
I'd be ok with 2 similar-looking tables, or using N/A. N/A is probably >>> the red checkmark.

skriss:
N/A SGTM

nrb:
I'm thinking of removing `Ark Team` from the owner columns of IBM Cloud, Minio, and Ceph. My reasoning is that support for these is incidental, and while we'll make an effort to help fix them, they're not explicitly supported.

Are others ok with that stance?  I mainly want to be sure we're clear and up front about the level of effort we'll put in to fixing potential issues.

ncdc:
Hmm that's a tricky one. @rosskukulinski wdyt?

skriss:
could we say "community-supported" or something like that?

docs/support-matrix.md
```diff
@@ -0,0 +1,51 @@
+# Compatible Storage Providers
+
+Ark is compatible with a variety of storage providers for backup data, as well as for volume snapshots using either a plugin for the volume provider, or [restic][4] for filesystem level backups.
```
Bradamant3: Let's try introducing the topic a bit differently. This sentence/paragraph tries to pack too much information into too few words, but unpacking it doesn't add much value. Can we simplify down to something like:

"Ark supports a variety of storage providers for different backup and snapshot operations. As of version 0.6.0, a plugin system allows anyone ..." (and continue with rest of following paragraph)

docs/support-matrix.md
```diff
@@ -0,0 +1,51 @@
...
+
+| Provider                  | Owner    | Contact                         |
+|---------------------------|----------|---------------------------------|
+| [AWS S3][2]               | Ark Team | [Slack][10], [GitHub Issue][11] |
+| [Azure Blob Storage][3]   | Ark Team |                                 |
+| [Google Cloud Storage][4] | Ark Team |                                 |
+
+## S3-Compatible Backup Storage Providers
+
+Ark interfaces with the S3 API via [Amazon's Go SDK][12]. Some third party object storage providers use the S3 API, and users have reported the following will work using Ark's S3 support.
```
Bradamant3: "Ark implements [Amazon's Go SDK][12] to connect to the S3 API. Some third-party storage providers also support the S3 API, and users have reported the following providers work with Ark:"

docs/support-matrix.md
```diff
@@ -0,0 +1,51 @@
...
+|----------------------------------|-----------------|---------------------------------|
+| [AWS EBS][2]                     | Ark Team        | [Slack][10], [GitHub Issue][11] |
+| [Azure Managed Disks][3]         | Ark Team        |                                 |
+| [Google Compute Engine Disks][4] | Ark Team        |                                 |
+| [Portworx][6]                    | PortWorx        |                                 |
+| [DigitalOcean][7]                | StackPointCloud |                                 |
+
+### Adding a new plugin
+
+If you would like to write a plugin for a new backup or volume storage system, take a look at the [example repo][8].
```
Bradamant3: S/If you would like to/To

docs/support-matrix.md
```diff
@@ -0,0 +1,51 @@
...
+
+Once you've published your plugin, open a PR adding the relevant information to the tables above.
```
Bradamant3: s/Once you've published/After you publish
s/adding the relevant information to the tables above./that adds your plugin to the appropriate list.

carlisia:
I was gonna merge but it still says `WIP` in the title so gonna wait.

nrb:
Title updated

docs/support-matrix.md
```diff
@@ -0,0 +1,50 @@
+# Compatible Storage Providers
+
+Ark supports a variety of storage providers for different backup and snapshot operations. As of version 0.6.0, a plugin system allows anyone to add compatibility for additional backup and volume storage platforms without modifying the Ark codebase.
+
+## Backup Storage Providers
+
+| Provider                  | Owner    | Contact                         |
+|---------------------------|----------|---------------------------------|
+| [AWS S3][2]               | Ark Team | [Slack][10], [GitHub Issue][11] |
```
skriss: I'd put the contact stuff in each row or find a way to merge the cells

docs/support-matrix.md
```diff
@@ -0,0 +1,50 @@
...
+_Note that these providers are not regularly tested by the Ark team._
+
+ * [IBM Cloud][5]
+ * [Minio][9]
+ * Ceph RADIOS v12.2.7
+
+## Volume Snapshot Providers
+| Provider                         | Owner           | Contact                         |
+|----------------------------------|-----------------|---------------------------------|
+| [AWS EBS][2]                     | Ark Team        | [Slack][10], [GitHub Issue][11] |
```
skriss: I'd put the contact stuff in each Ark Team row, or find a way to merge the cells

docs/support-matrix.md
```diff
@@ -0,0 +1,50 @@
...
+
+| Provider                  | Owner    | Contact                         |
+|---------------------------|----------|---------------------------------|
+| [AWS S3][2]               | Ark Team | [Slack][10], [GitHub Issue][11] |
+| [Azure Blob Storage][3]   | Ark Team |                                 |
+| [Google Cloud Storage][4] | Ark Team |                                 |
+
+## S3-Compatible Backup Storage Providers
+
+Ark implements [Amazon's Go SDK][12] to connect to the S3 API. Some third-party storage providers also support the S3 API, and users have reported the following providers work with Ark:
```
skriss: s/implements/uses

docs/support-matrix.md
```diff
@@ -0,0 +1,50 @@
...
+
+_Note that these providers are not regularly tested by the Ark team._
+
+ * [IBM Cloud][5]
+ * [Minio][9]
+ * Ceph RADIOS v12.2.7
```
skriss: should this be `RADOS`?

skriss:
couple last-min things but otherwise LGTM.  We should reach out to portworx and stackpointcloud to ask them to file PRs with their support contact info. cc @rosskukulinski

docs/support-matrix.md
nrb: Doh, yep.

docs/support-matrix.md
nrb: Looks like GH markdown doesn't support merging cells, and I don't want to mix HTML tables in, so I'll just duplicate it.

skriss:
@nrb signoff checker is unhappy, you may need to force-push to get it to try again

PR #725, opened 2018-08-03
Document CRD prerequisites errors

nrb:
This change has the contents from #719, but for the live site docs.


skriss:
LGTM

PR #744, opened 2018-08-07
Add BackupStorageLocation API type

### LINKED ISSUE #735: [Feature] Support backing up/restoring multiple PV types per backup

### LINKED ISSUE #736: Add BackupStorageLocation CRD and reference it in the Backup CRD type

nrb:
This commit only provides the data model for further work. It does not
implement any logic around locations, nor does it remove anything from
the Config API type.

Closes #736


pkg/apis/ark/v1/backup_locations.go
```diff
@@ -0,0 +1,89 @@
+/*
```
skriss: I'd call the file `backup_storage_location.go`

pkg/apis/ark/v1/backup_locations.go
```diff
@@ -0,0 +1,89 @@
...
+	// BackupStorageLocationAccessModeReadOnly represents read-only access to a BackupStorageLocation.
+	BackupStorageLocationAccessModeReadOnly BackupStorageLocationAccessMode = "ReadOnly"
+
+	// BackupStorageLocationAccessModeReadWrite represents read and write access to a BackupStorageLocation.
+	BackupStorageLocationAccessModeReadWrite BackupStorageLocationAccessMode = "ReadWrite"
+)
+
+// BackupStorageLocationStatus describes the current status of an Ark BackupStorageLocation.
+type BackupStorageLocationStatus struct {
+	Phase      BackupStorageLocationPhase      `json:"phase"`
```
skriss: Let's add an `omitempty` tag to each of these status fields. I'm not sure we'll actually get around to actually implementing anything that uses them until later, so will be nice not to actually show them in json/yaml.

pkg/apis/ark/v1/backup_locations.go
```diff
@@ -0,0 +1,89 @@
...
+type BackupStorageLocationList struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Items             []BackupStorageLocation `json:"items"`
+}
+
+// ObjectStorageLocation specifies the settings necessary to connect to a provider's object storage.
+type ObjectStorageLocation struct {
+	// Provider is the cloud provider where object storage resides.
+	Provider string `json:"provider"`
```
skriss: Do you think `Provider` should be a field directly on `spec`, or under `ObjectStorageLocation`? I'm not sure, but thinking about scenarios where we add additional location types (e.g. git).  Same question for `Config`

pkg/apis/ark/v1/backup_locations.go
```diff
@@ -0,0 +1,89 @@
...
+	// Bucket is the bucket to use for object storage.
+	Bucket string `json:"bucket"`
+
+	// Prefix is the path inside a bucket to use for Ark storage. Optional.
+	Prefix string `json:"prefix"`
+}
+
+// BackupStorageLocationSpec defines the specification for an Ark BackupStorageLocation
+type BackupStorageLocationSpec struct {
+	ObjectStorage ObjectStorageLocation `json:"objectStorage"`
```
skriss: I think this should be a pointer, with an `omitempty` tag - over time we may want to support additional types of backup storage locations, so the idea would be that exactly one would be non-nil.

We might want to do this the same way `PersistentVolumeSource` is implemented - ref. https://github.com/kubernetes/api/blob/master/core/v1/types.go#L298 and https://github.com/kubernetes/api/blob/master/core/v1/types.go#L170

pkg/apis/ark/v1/backup_locations.go
nrb: I think that makes sense. I'll update it.

pkg/apis/ark/v1/backup_locations.go
nrb: Are you thinking of maybe a `StorageType` struct that held pointers to `ObjectStorageLocation` and (for example) `GitStorageLocation`?

skriss: yeah, exactly

skriss: And looks like for PVs, `PersistentVolumeSpec` embeds the `PersistentVolumeSource`, so could do the same here

skriss:
this is looking good to me, probably just need to re-gen deep-copy stuff

pkg/apis/ark/v1/backup.go
```diff
@@ -56,6 +59,9 @@ type BackupSpec struct {
 
 	// Hooks represent custom behaviors that should be executed at different phases of the backup.
 	Hooks BackupHooks `json:"hooks"`
+
+	// StorageLocation is a reference to a BackupStorageLocation where the backup should be stored.
+	StorageLocation corev1api.ObjectReference `json:"storageLocation"`
```
ncdc: Could this just be a `string` for the name of the `BackupStorageLocation`?

pkg/apis/ark/v1/backup_storage_locations.go
```diff
@@ -0,0 +1,95 @@
...
+	// Bucket is the bucket to use for object storage.
+	Bucket string `json:"bucket"`
+
+	// Prefix is the path inside a bucket to use for Ark storage. Optional.
+	Prefix string `json:"prefix"`
+}
+
+// BackupStorageLocationSpec defines the specification for an Ark BackupStorageLocation
+type BackupStorageLocationSpec struct {
+	// Provider is the cloud provider where object storage resides.
```
ncdc: Maybe try to avoid calling it a "cloud" provider since it could be on-prem?

pkg/apis/ark/v1/backup_storage_locations.go
```diff
@@ -0,0 +1,95 @@
...
+	Provider string `json:"provider"`
+
+	// Config is for provider-specific configuration fields.
+	Config map[string]string `json:"config"`
+
+	StorageType `json:",inline"`
```
ncdc: What is the motivation for splitting this out into a separate struct and then inlining it?

skriss: I proposed this, following the pattern of `PersistentVolume` and `PersistentVolumeSource`.  We could just eliminate the `StorageType` struct but it does provide some clarity around "the set of pointers of which exactly one should be non-nil"

ncdc: It's certainly not harmful by any means. Ok to keep it

pkg/apis/ark/v1/backup.go
skriss: @ncdc is there a best practice or pattern around when to use a string/strings vs. `LocalObjectReference` vs. `ObjectReference`?

ncdc: `ObjectReference` is if you need to go across namespaces. It supports `APIVersion` and `Kind` for specifying different resources.

`LocalObjectReference` currently only has a `Name` field, so you can't specify `APIVersion` and `Kind` in the event you needed to be able to reference different resources.

As to `string` vs `LocalObjectReference`, I don't know that there's any good guidance. If `LocalObjectReference` supported `APIVersion` and `Kind`, I'd say use `string` if you don't need to vary the resource, and `LocalObjectReference` if you do. Given that this isn't a supported use case, and we only care about a single kind of resource (`BackupStorageLocation`), my recommendation would be to use a `string` here. It also makes the json/yaml easier to write.

WDYT - do you agree?

nrb: That makes sense - a `string` seems fine to me.

skriss: yeah, that works for me. I was thinking this was a `LocalObjectRef` when i reviewed it, and i'm fine just using a string in place.

pkg/apis/ark/v1/backup_storage_locations.go
```diff
@@ -0,0 +1,95 @@
+/*
```
skriss: file name: `backup_storage_locations.go` -> `backup_storage_location.go`

skriss:
Also should add it to `examples/common/00-prereqs.yaml`

skriss:
@carlisia PTAL

skriss:
LGTM once you update `00-prereqs.yaml`.  And we can always tweak if we need to as we go through the rest of the work.

pkg/apis/ark/v1/backup.go
```diff
@@ -56,6 +58,9 @@ type BackupSpec struct {
 
 	// Hooks represent custom behaviors that should be executed at different phases of the backup.
 	Hooks BackupHooks `json:"hooks"`
+
+	// StorageLocation is a string containing the name of a BackupStorageLocation where the backup should be stored.
```
carlisia: Super nit-picky comment incoming: it's not necessary to describe the type of the variable on the comment, because it's redundant. The comment should be limited to what the variable does or is for. There are existing variables that are described this way as well.

pkg/apis/ark/v1/backup_storage_location.go
```diff
@@ -0,0 +1,95 @@
...
+
+// BackupStorageLocationList is a list of BackupStorageLocations.
+type BackupStorageLocationList struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Items             []BackupStorageLocation `json:"items"`
+}
+
+// StorageType represents the type of storage that a backup location uses.
+// Exactly one must be set.
```
carlisia: Making this comment with trepidation due to low confidence I'm getting this right: is it possible to set more than one `StorageType`? If not, this comment would be confusing.

skriss: currently, yes. it's set up to be extensible over time for other types of backends besides object storage to be added. fine to make the comment explicit about that, i.e. "ObjectStorage must be non-nil as it's currently the only supported StorageType"

carlisia: I like that ⬆️ and modifying it to "Exactly one of its members must be set." when another member is added.

pkg/apis/ark/v1/backup.go
nrb: I agree it's redundant. In many cases this is fitting in with the existing format (see https://github.com/heptio/ark/blob/master/pkg/apis/ark/v1/backup.go#L223, for example).

I'd like to remove the redundancy in the comments across the code base, but I'd like to be consistent.

Issue #750, opened 2018-08-08
Add a --default-backup-location flag to `ark server`, whose value defaults to "default"


skriss:

skriss:
This will be used to specify which `BackupStorageLocation` is the default one. By default, the one named "default" will be the default, but the user can change this by explicitly specifying this flag.

PR #753, opened 2018-08-08
Add --storage-location argument to create commands

### LINKED ISSUE #738: Modify `ark backup create` and `ark schedule create` CLI commands to accept a backup location as an optional flag

nrb:
Closes #738


pkg/cmd/cli/backup/create.go
```diff
@@ -87,6 +88,7 @@ func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.Var(&o.IncludeResources, "include-resources", "resources to include in the backup, formatted as resource.group, such as storageclasses.storage.k8s.io (use '*' for all resources)")
 	flags.Var(&o.ExcludeResources, "exclude-resources", "resources to exclude from the backup, formatted as resource.group, such as storageclasses.storage.k8s.io")
 	flags.Var(&o.Labels, "labels", "labels to apply to the backup")
+	flags.StringVar(&o.StorageLocation, "storage-location", "default", "location in which to store the backup")
```
skriss: I'd say let's leave off the default value of "default" - we'll do defaulting of this field in the controller, based on the server flag that'll be implemented in #750

skriss:
I'm thinking we should add code to validate that the location exists here - that'd be consistent with e.g. `ark restore create` which validates that the backup/schedule exists in the CLI command.

pkg/cmd/cli/backup/create.go
nrb: Done

nrb:
@skriss Comments should be addressed

skriss:
@nrb this LGTM. Ideally I'd split this into two commits, one for all the generated stuff and then one for adding the flag - but if that's going to be too much of a pain to do now, we can skip. LMK.

nrb:
Yeah, I'll split them and fix the CI failure.

Issue #747, opened 2018-08-08
Create a BackupStorageLocation from the existing Config CRD at server startup


skriss:

skriss:
This is to make migration from pre-v0.10 versions easy.

carlisia:
I think we'll better serve ark's users by explicitly moving them on to the new concept of BSL and also remove the possibility of using a CRD for configuration sooner rather than later.

ncdc:
I think it will be easier to fully delete the `Config`. If we do go this route, we should start socializing the change now - Twitter, Google Group, Slack, etc.

rosskukulinski:
At this point in the project lifecycle, I'm on-board withfully deleting `Config` for 0.10 and forcing users to upgrade.

That said, we should plan to have solid docs on _how_ to upgrade 0.9.x to 0.10.0 and test them thoroughly.

nrb:
I'm for deleting the `Config` CRD earlier so long as we have documentation and clear errors when they're found.

skriss:
We decided not to do this and to require users to add their locations a/o v0.10.

PR #757, opened 2018-08-09
Switch Config CRD elements to server flags

### LINKED ISSUE #538: Replace Config CRD
I'm trying to help get the PR for a Ark Helm chart merged.

https://github.com/kubernetes/charts/pull/3795

Over at the chats repo we have a CI that lints and install charts. I figured this is not possible with Ark. I'll have to wrap the deployment into an if statement that makes sure it is not created by the CI job because Ark won't come up if it doesn't have a cloud provider and credentials set.

Now, since Ark has its config separate in the `Config` CRD, I find that kind of strange. Actually, this mean the config is spread over two places. I understand that this separation is probably necessary because the secret needs to be mounted. But then why separate it at all instead of having a bunch of CLI args and everything in the deployment?

On the other hand, I like the separate Config resource. I wonder if there is a way to also reference secrets in there. I guess the initial Ark deployment would then have to be turned into a little daemon whose sole purpose is to boostrap real Ark server deployments based on the complete config mounting the configured secret. In theory this would allow boostrapping multiple Ark servers with different configs. I think, this is when a separate config really would make sense. Has such a setup been discussed? Would that make sense?

carlisia:
Fixes #538.

### DIFF
--- a/docs/cli-reference/ark_server.md
+++ b/docs/cli-reference/ark_server.md
@@ -14,10 +14,14 @@ ark server [flags]
 ### Options
 
 ```
-  -h, --help                     help for server
-      --log-level                the level at which to log. Valid values are debug, info, warning, error, fatal, panic. (default info)
-      --metrics-address string   the address to expose prometheus metrics (default ":8085")
-      --plugin-dir string        directory containing Ark plugins (default "/plugins")
+      --backup-sync-period duration               how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster (default 1h0m0s)
+  -h, --help                                      help for server
+      --log-level                                 the level at which to log. Valid values are debug, info, warning, error, fatal, panic. (default info)
+      --metrics-address string                    the address to expose prometheus metrics (default ":8085")
+      --plugin-dir string                         directory containing Ark plugins (default "/plugins")
+      --restic-timeout duration                   how long backups/restores of pod volumes should be allowed to run before timing out (default 1h0m0s)
+      --restore-only                              run in a mode where only restores are allowed; backups, schedules, and garbage-collection are all disabled
+      --restore-resource-priorities stringSlice   desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources (default [namespaces,persistentvolumes,persistentvolumeclaims,secrets,configmaps,serviceaccounts,limitranges,pods])
 ```
 
 ### Options inherited from parent commands
--- a/docs/config-definition.md
+++ b/docs/config-definition.md
@@ -1,23 +1,27 @@
-# Ark Config definition
+# Ark Config definition and Ark server deployment
 
-* [Overview][8]
-* [Example][9]
-* [Parameter Reference][6]
-  * [Main config][7]
-  * [AWS][0]
-  * [GCP][1]
-  * [Azure][2]
+* [Config][8]
+  * [Example][9]
+  * [Parameter Reference][6]
+    * [Main config][7]
+    * [AWS][0]
+    * [GCP][1]
+    * [Azure][2]
+* [Deployment][11]
+  * [Sample Deployment][13]
+  * [Parameter Options][14]
 
-## Overview
+## Config
 
-Heptio Ark defines its own Config object (a custom resource) for specifying Ark backup and cloud provider settings. When the Ark server is first deployed, it waits until you create a Config--specifically one named `default`--in the `heptio-ark` namespace.
+Heptio Ark defines its own Config object (a custom resource) for specifying Ark backup and cloud provider settings. When the Ark server is first deployed, it waits until you create a Config --specifically one named `default`-- in the `heptio-ark` namespace.
 
 > *NOTE*: There is an underlying assumption that you're running the Ark server as a Kubernetes deployment. If the `default` Config is modified, the server shuts down gracefully. Once the kubelet restarts the Ark server pod, the server then uses the updated Config values.
 
-## Example
+### Example
 
 A sample YAML `Config` looks like the following:
-```
+
+```yaml
 apiVersion: ark.heptio.com/v1
 kind: Config
 metadata:
@@ -32,17 +36,13 @@ backupStorageProvider:
   bucket: ark
   config:
     region: us-west-2
-backupSyncPeriod: 60m
-gcSyncPeriod: 60m
-scheduleSyncPeriod: 1m
-restoreOnlyMode: false
 ```
 
-## Parameter Reference
+### Parameter Reference
 
 The configurable parameters are as follows:
 
-### Main config parameters
+#### Main config parameters
 
 | Key | Type | Default | Meaning |
 | --- | --- | --- | --- |
@@ -53,17 +53,12 @@ The configurable parameters are as follows:
 | `backupStorageProvider/name` | String<br><br>(Ark natively supports `aws`, `gcp`, and `azure`. Other providers may be available via external plugins.) | Required Field | The name of the cloud provider that will be used to actually store the backups. |
 | `backupStorageProvider/bucket` | String | Required Field | The storage bucket where backups are to be uploaded. |
 | `backupStorageProvider/config` | map[string]string<br><br>(See the corresponding [AWS][0], [GCP][1], and [Azure][2]-specific configs or your provider's documentation.) | None (Optional) | Configuration keys/values to be passed to the cloud provider for backup storage. |
-| `backupSyncPeriod` | metav1.Duration | 60m0s | How frequently Ark queries the object storage to make sure that the appropriate Backup resources have been created for existing backup files. |
-| `gcSyncPeriod` | metav1.Duration | 60m0s | How frequently Ark queries the object storage to delete backup files that have passed their TTL. |
-| `scheduleSyncPeriod` | metav1.Duration | 1m0s | How frequently Ark checks its Schedule resource objects to see if a backup needs to be initiated. |
-| `resourcePriorities` | []string | `[namespaces, persistentvolumes, persistentvolumeclaims, secrets, configmaps, serviceaccounts, limitranges]` | An ordered list that describes the order in which Kubernetes resource objects should be restored (also specified with the `<RESOURCE>.<GROUP>` format.<br><br>If a resource is not in this list, it is restored after all other prioritized resources. |
-| `restoreOnlyMode` | bool | `false` | When RestoreOnly mode is on, functionality for backups, schedules, and expired backup deletion is *turned off*. Restores are made from existing backup files in object storage. |
 
-### AWS
+#### AWS
 
 **(Or other S3-compatible storage)**
 
-#### backupStorageProvider/config
+##### backupStorageProvider/config
 
 | Key | Type | Default | Meaning |
 | --- | --- | --- | --- |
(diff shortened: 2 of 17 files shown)


carlisia:
Note: I'm fixing the tests.

pkg/cmd/server/server.go
```diff
@@ -75,11 +75,18 @@ const (
 	defaultMetricsAddress = ":8085"
 )
 
+type apiConfig struct {
```
skriss: I'd probably call this `serverConfig`

pkg/cmd/server/server.go
```diff
@@ -75,11 +75,18 @@ const (
...
+	restoreResourcePriorities                   []string
+	restoreOnly                                 bool
+}
+
 func NewCommand() *cobra.Command {
 	var (
 		logLevelFlag   = logging.LogLevelFlag(logrus.InfoLevel)
 		pluginDir      = "/plugins"
 		metricsAddress = defaultMetricsAddress
+		config         = apiConfig{}
```
skriss: I like using a struct here - I think I'd move the above three fields onto it as well

pkg/cmd/server/server.go
```diff
@@ -124,6 +131,10 @@ func NewCommand() *cobra.Command {
 	command.Flags().Var(logLevelFlag, "log-level", fmt.Sprintf("the level at which to log. Valid values are %s.", strings.Join(logLevelFlag.AllowedValues(), ", ")))
 	command.Flags().StringVar(&pluginDir, "plugin-dir", pluginDir, "directory containing Ark plugins")
 	command.Flags().StringVar(&metricsAddress, "metrics-address", metricsAddress, "the address to expose prometheus metrics")
+	command.Flags().Duration("backup-sync-period", config.backupSyncPeriod, "how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster")
+	command.Flags().Duration("pod-vol-op-timeout", config.podVolumeOperationTimeout, "how long backups/restores of pod volumes should be allowed to run before timing out")
+	command.Flags().BoolVar(&config.restoreOnly, "restore-only", config.restoreOnly, "how long backups/restores of pod volumes should be allowed to run before timing out")
```
skriss: description is wrong

pkg/cmd/server/server.go
```diff
@@ -124,6 +131,10 @@ func NewCommand() *cobra.Command {
 	command.Flags().Var(logLevelFlag, "log-level", fmt.Sprintf("the level at which to log. Valid values are %s.", strings.Join(logLevelFlag.AllowedValues(), ", ")))
 	command.Flags().StringVar(&pluginDir, "plugin-dir", pluginDir, "directory containing Ark plugins")
 	command.Flags().StringVar(&metricsAddress, "metrics-address", metricsAddress, "the address to expose prometheus metrics")
+	command.Flags().Duration("backup-sync-period", config.backupSyncPeriod, "how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster")
+	command.Flags().Duration("pod-vol-op-timeout", config.podVolumeOperationTimeout, "how long backups/restores of pod volumes should be allowed to run before timing out")
```
skriss: don't love the flag name but I get that it's super-long if you spell it all out.  I wonder if we should just call this `restic-timeout`.  Probably way clearer.

pkg/cmd/server/server.go
```diff
@@ -232,7 +245,7 @@ func (s *server) run() error {
 		return err
 	}
 
-	// check to ensure all Ark CRDs exist
+	// check to ensure all Ark CRDs exis
```
skriss: pls revert this line

pkg/cmd/server/server.go
```diff
@@ -246,6 +259,7 @@ func (s *server) run() error {
 	// separate object, and instead apply defaults to a clone.
 	config := originalConfig.DeepCopy()
 	applyConfigDefaults(config, s.logger)
+	s.applyConfigDefaults()
```
skriss: I don't like having this func be separate from the previous function, esp. with the same names.  I think it makes sense to keep the new method you created with the server receiver, to pass it the *api.Config object, and have it do what it's already doing plus whatever's left in the old `applyConfigDefaults`, then we can get rid of the old func.

pkg/cmd/server/server.go
```diff
@@ -379,10 +410,10 @@ func (s *server) loadConfig() (*api.Config, error) {
 }
 
 const (
-	defaultGCSyncPeriod              = 60 * time.Minute
 	defaultBackupSyncPeriod          = 60 * time.Minute
-	defaultScheduleSyncPeriod        = time.Minute
 	defaultPodVolumeOperationTimeout = 60 * time.Minute
+	gcSyncPeriod                     = 60 * time.Minute
```
skriss: since `gcSyncPeriod` and `scheduleSyncPeriod` aren't configurable, let's move them into `pkg/controller` (first in `gc_controller.go` and second in `schedule_controller.go`) and remove them from the constructors for those controllers.

pkg/cmd/server/server.go
```diff
@@ -643,7 +651,7 @@ func (s *server) runControllers(config *api.Config) error {
 			s.arkClient.ArkV1(),
 			s.arkClient.ArkV1(),
 			s.sharedInformerFactory.Ark().V1().Schedules(),
-			config.ScheduleSyncPeriod.Duration,
+			gcSyncPeriod,
```
skriss: this shoulda been `scheduleSyncPeriod` but per previous comment, let's just remove this parameter from the constructor.

pkg/apis/ark/v1/config.go
```diff
@@ -45,31 +45,6 @@ type Config struct {
 	// Ark backups are stored in object storage. This may be a different cloud than
 	// where the cluster is running.
 	BackupStorageProvider ObjectStorageProviderConfig `json:"backupStorageProvider"`
-
```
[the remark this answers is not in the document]
carlisia: @skriss given the feedback on #747, we are clearly do away with these two fields. Should I do anything about them in this PR?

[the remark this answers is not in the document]
skriss: nope, can disregard this comment.

pkg/cmd/server/server.go
```diff
@@ -124,6 +132,10 @@ func NewCommand() *cobra.Command {
 	command.Flags().Var(logLevelFlag, "log-level", fmt.Sprintf("the level at which to log. Valid values are %s.", strings.Join(logLevelFlag.AllowedValues(), ", ")))
```
skriss: Still need to switch these 2 flags (`plugin-dir`, `metrics-address`) to use the fields on the `serverConfig` struct. Then remove the unused local vars for these.

pkg/cmd/server/server.go
```diff
@@ -124,6 +132,10 @@ func NewCommand() *cobra.Command {
...
 	command.Flags().StringVar(&pluginDir, "plugin-dir", pluginDir, "directory containing Ark plugins")
 	command.Flags().StringVar(&metricsAddress, "metrics-address", metricsAddress, "the address to expose prometheus metrics")
+	command.Flags().Duration("backup-sync-period", config.backupSyncPeriod, "how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster")
```
skriss: Need to use `DurationVar` for these two and give it the address of the config field

pkg/cmd/server/server.go
```diff
@@ -124,6 +132,10 @@ func NewCommand() *cobra.Command {
...
+	command.Flags().Duration("restic-timeout", config.podVolumeOperationTimeout, "how long backups/restores of pod volumes should be allowed to run before timing out")
+	command.Flags().BoolVar(&config.restoreOnly, "restore-only", config.restoreOnly, "run in a mode where only restores are allowed; backups, schedules, and garbage-collection are all disabled")
+	command.Flags().StringSlice("restore-resource-priorities", config.restoreResourcePriorities, "desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources")
```
skriss: `StringSliceVar`

pkg/cmd/server/server.go
```diff
@@ -75,11 +75,19 @@ const (
...
+	restoreResourcePriorities                   []string
+	restoreOnly                                 bool
+}
+
 func NewCommand() *cobra.Command {
 	var (
 		logLevelFlag   = logging.LogLevelFlag(logrus.InfoLevel)
 		pluginDir      = "/plugins"
 		metricsAddress = defaultMetricsAddress
+		config         = serverConfig{}
```
skriss: let's set the fields that have defaults to those values here, that way the default values will show up in the CLI help.  Still keep the `applyConfigDefaults` func since we don't want to allow zero values for the timeouts, etc.

docs/cli-reference/ark_server.md
```diff
@@ -14,10 +14,14 @@ ark server [flags]
...
-      --log-level                the level at which to log. Valid values are debug, info, warning, error, fatal, panic. (default info)
-      --metrics-address string   the address to expose prometheus metrics (default ":8085")
-      --plugin-dir string        directory containing Ark plugins (default "/plugins")
+      --backup-sync-period duration               how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster
+  -h, --help                                      help for server
+      --log-level                                 the level at which to log. Valid values are debug, info, warning, error, fatal, panic. (default info)
+      --metrics-address string                    the address to expose prometheus metrics (default ":8085")
+      --plugin-dir string                         directory containing Ark plugins (default "/plugins")
+      --restic-timeout duration                   how long backups/restores of pod volumes should be allowed to run before timing out
+      --restore-only                              run in a mode where only restores are allowed; backups, schedules, and garbage-collection are all disabled
```
nrb: https://github.com/heptio/ark/blob/master/docs/use-cases.md#L22 will need to be updated to reflect this change.

nrb:
Some files that will need to be updated to reflect these changes:

* https://github.com/heptio/ark/blob/master/docs/config-definition.md should be removed
* Any `config.yaml` file in `examples/<provider>` should be removed
* Deployments in `examples/<provider>` may need to be updated (I haven't looked closely here though)
* Docs should be added to https://github.com/heptio/ark/blob/master/docs/use-cases.md for modifying the Ark deployment instead of the Config for restore only mode. These may be temporary til we get access mode logic in, but I think keeping the docs consistent is important.

skriss:
Just FYI this PR won't fully remove the CRD yet - we still have PVProvider in there (and we haven't removed BackupStorageProvider quite yet). I'd remove references to the fields that are being removed from the docs @nrb pointed out above, but hold off on deleting things wholesale until we get thru the PVProvider/SnapshotLocation work

carlisia:
Ready for another review.

carlisia:
image

I redid the apply config and deploy. Running a manual test and `ark create` works but trying to get the log fails. I would expect `ark create` not to work either. I think I did the yaml deployment file wrong. Trying to fix it.

carlisia:
Ok this test is not gonna work because the image in the deploy is set to `latest`. Need to learn how to push my own image. 

Also, `ark create` never finishes, I feel better about this.

carlisia:
I applied the new config and applied the deployment with the new `backup-sync-period` flag, and was able to create a new backup running against an image for this branch.

Ready for review.

docs/config-definition.md
```diff
@@ -53,11 +49,6 @@ The configurable parameters are as follows:
 | `backupStorageProvider/name` | String<br><br>(Ark natively supports `aws`, `gcp`, and `azure`. Other providers may be available via external plugins.) | Required Field | The name of the cloud provider that will be used to actually store the backups. |
 | `backupStorageProvider/bucket` | String | Required Field | The storage bucket where backups are to be uploaded. |
 | `backupStorageProvider/config` | map[string]string<br><br>(See the corresponding [AWS][0], [GCP][1], and [Azure][2]-specific configs or your provider's documentation.) | None (Optional) | Configuration keys/values to be passed to the cloud provider for backup storage. |
-| `backupSyncPeriod` | metav1.Duration | 60m0s | How frequently Ark queries the object storage to make sure that the appropriate Backup resources have been created for existing backup files. |
```
skriss: I'm thinking that we should add a section to this page that states that `ark server` has a number of config flags.  We can include a brief snippet of YAML that shows how to specify them, and then put a link to the `ark server` CLI reference docs for the details of the flags

examples/aws/10-deployment-kube2iam.yaml
```diff
@@ -42,6 +42,8 @@ spec:
             - /ark
           args:
             - server
+            - --backup-sync-period
```
skriss: I think we can probably skip explicitly specifying this flag in all of the example YAMLs.  Not sure why we had the value explicitly set to something other than the default value - likely not intentional. @ncdc any reason you can recall? (the default value for backup sync is `60m` in code).

carlisia: I propose we leave it for the purpose of an example how to correct pass in flags for the server. I think this has value even tho changing this value doesn't seem useful.

docs/config-definition.md
carlisia: Ah I read the comment below first. I think this is a great idea. In this case I'm good with removing the flag in the deployment files.

carlisia:
@skriss PTAL

pkg/cmd/server/server.go
```diff
@@ -114,16 +126,20 @@ func NewCommand() *cobra.Command {
...
 			cmd.CheckError(s.run())
 		},
 	}
 
 	command.Flags().Var(logLevelFlag, "log-level", fmt.Sprintf("the level at which to log. Valid values are %s.", strings.Join(logLevelFlag.AllowedValues(), ", ")))
-	command.Flags().StringVar(&pluginDir, "plugin-dir", pluginDir, "directory containing Ark plugins")
-	command.Flags().StringVar(&metricsAddress, "metrics-address", metricsAddress, "the address to expose prometheus metrics")
+	command.Flags().StringVar(&config.pluginDir, "plugin-dir", config.pluginDir, "directory containing Ark plugins")
+	command.Flags().StringVar(&config.metricsAddress, "metrics-address", config.metricsAddress, "the address to expose prometheus metrics")
+	command.Flags().DurationVar(&config.backupSyncPeriod, "backup-sync-period", config.backupSyncPeriod, "how often, in minutes, to ensure all Ark backups in object storage exist as Backup API objects in the cluster")
```
skriss: this doesn't have to be "in minutes" -- any valid duration string will do.  I'd just remove ", in minutes,"

pkg/cmd/server/server.go
```diff
@@ -114,16 +126,20 @@ func NewCommand() *cobra.Command {
...
+	command.Flags().DurationVar(&config.podVolumeOperationTimeout, "restic-timeout", config.podVolumeOperationTimeout, "how long, in minutes, backups/restores of pod volumes should be allowed to run before timing out")
```
skriss: ditto

PR #766, opened 2018-08-14
Use backup storage location during restore

### LINKED ISSUE #740: Update restore controller to use the backup's specified location

carlisia:
Closes #740

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -712,13 +712,13 @@ func (s *server) runControllers(config *api.Config) error {
 		s.arkClient.ArkV1(),
 		s.arkClient.ArkV1(),
 		restorer,
-		config.BackupStorageProvider.CloudProviderConfig,
-		config.BackupStorageProvider.Bucket,
 		s.sharedInformerFactory.Ark().V1().Backups(),
+		s.sharedInformerFactory.Ark().V1().BackupStorageLocations(),
 		s.blockStore != nil,
 		s.logger,
 		s.logLevel,
 		s.pluginRegistry,
+		s.defaultBackupLocation,
 		s.metrics,
 	)
 
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -71,23 +71,24 @@ var nonRestorableResources = []string{
 }
 
 type restoreController struct {
-	namespace           string
-	restoreClient       arkv1client.RestoresGetter
-	backupClient        arkv1client.BackupsGetter
-	restorer            restore.Restorer
-	objectStoreConfig   api.CloudProviderConfig
-	bucket              string
-	pvProviderExists    bool
-	backupLister        listers.BackupLister
-	backupListerSynced  cache.InformerSynced
-	restoreLister       listers.RestoreLister
-	restoreListerSynced cache.InformerSynced
-	syncHandler         func(restoreName string) error
-	queue               workqueue.RateLimitingInterface
-	logger              logrus.FieldLogger
-	logLevel            logrus.Level
-	pluginRegistry      plugin.Registry
-	metrics             *metrics.ServerMetrics
+	namespace                  string
+	restoreClient              arkv1client.RestoresGetter
+	backupClient               arkv1client.BackupsGetter
+	restorer                   restore.Restorer
+	pvProviderExists           bool
+	backupLister               listers.BackupLister
+	backupListerSynced         cache.InformerSynced
+	restoreLister              listers.RestoreLister
+	restoreListerSynced        cache.InformerSynced
+	backupLocationLister       listers.BackupStorageLocationLister
+	backupLocationListerSynced cache.InformerSynced
+	syncHandler                func(restoreName string) error
+	queue                      workqueue.RateLimitingInterface
+	logger                     logrus.FieldLogger
+	logLevel                   logrus.Level
+	pluginRegistry             plugin.Registry
+	defaultBackupLocation      string
+	metrics                    *metrics.ServerMetrics
 
 	getBackup            cloudprovider.GetBackupFunc
 	downloadBackup       cloudprovider.DownloadBackupFunc
@@ -102,33 +103,34 @@ func NewRestoreController(
 	restoreClient arkv1client.RestoresGetter,
 	backupClient arkv1client.BackupsGetter,
 	restorer restore.Restorer,
-	objectStoreConfig api.CloudProviderConfig,
-	bucket string,
 	backupInformer informers.BackupInformer,
+	backupLocationInformer informers.BackupStorageLocationInformer,
 	pvProviderExists bool,
 	logger logrus.FieldLogger,
 	logLevel logrus.Level,
 	pluginRegistry plugin.Registry,
+	defaultBackupLocation string,
 	metrics *metrics.ServerMetrics,
 
 ) Interface {
 	c := &restoreController{
-		namespace:           namespace,
-		restoreClient:       restoreClient,
-		backupClient:        backupClient,
-		restorer:            restorer,
-		objectStoreConfig:   objectStoreConfig,
-		bucket:              bucket,
-		pvProviderExists:    pvProviderExists,
-		backupLister:        backupInformer.Lister(),
-		backupListerSynced:  backupInformer.Informer().HasSynced,
-		restoreLister:       restoreInformer.Lister(),
-		restoreListerSynced: restoreInformer.Informer().HasSynced,
-		queue:               workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "restore"),
-		logger:              logger,
-		logLevel:            logLevel,
-		pluginRegistry:      pluginRegistry,
-		metrics:             metrics,
+		namespace:                  namespace,
+		restoreClient:              restoreClient,
+		backupClient:               backupClient,
+		restorer:                   restorer,
+		pvProviderExists:           pvProviderExists,
+		backupLister:               backupInformer.Lister(),
+		backupListerSynced:         backupInformer.Informer().HasSynced,
+		restoreLister:              restoreInformer.Lister(),
+		restoreListerSynced:        restoreInformer.Informer().HasSynced,
+		backupLocationLister:       backupLocationInformer.Lister(),
+		backupLocationListerSynced: backupLocationInformer.Informer().HasSynced,
+		queue:                 workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "restore"),
+		logger:                logger,
+		logLevel:              logLevel,
+		pluginRegistry:        pluginRegistry,
+		defaultBackupLocation: defaultBackupLocation,
+		metrics:               metrics,
 
 		getBackup:            cloudprovider.GetBackup,
 		downloadBackup:       cloudprovider.DownloadBackup,
@@ -193,7 +195,7 @@ func (c *restoreController) Run(ctx context.Context, numWorkers int) error {
 	defer c.logger.Info("Shutting down RestoreController")
 
 	c.logger.Info("Waiting for caches to sync")
-	if !cache.WaitForCacheSync(ctx.Done(), c.backupListerSynced, c.restoreListerSynced) {
+	if !cache.WaitForCacheSync(ctx.Done(), c.backupListerSynced, c.restoreListerSynced, c.backupLocationListerSynced) {
 		return errors.New("timed out waiting for caches to sync")
 	}
 	c.logger.Info("Caches are synced")
@@ -283,28 +285,25 @@ func (c *restoreController) processRestore(key string) error {
 	pluginManager := c.newPluginManager(logContext, logContext.Level, c.pluginRegistry)
 	defer pluginManager.CleanupClients()
 
-	objectStore, err := getObjectStore(c.objectStoreConfig, pluginManager)
-	if err != nil {
-		return errors.Wrap(err, "error initializing object store")
-	}
-
 	actions, err := pluginManager.GetRestoreItemActions()
 	if err != nil {
 		return errors.Wrap(err, "error initializing restore item actions")
 	}
 
-	// complete & validate restore
-	if restore.Status.ValidationErrors = c.completeAndValidate(objectStore, restore); len(restore.Status.ValidationErrors) > 0 {
-		restore.Status.Phase = api.RestorePhaseFailedValidation
-	} else {
-		restore.Status.Phase = api.RestorePhaseInProgress
-	}
-
+	// validate the restore and fetch the backup
+	info := c.validateAndComplete(restore, pluginManager)
 	backupScheduleName := restore.Spec.ScheduleName
 	// Register attempts after validation so we don't have to fetch the backup multiple times
 	c.metrics.RegisterRestoreAttempt(backupScheduleName)
 
-	// update status
+	if len(restore.Status.ValidationErrors) > 0 {
+		restore.Status.Phase = api.RestorePhaseFailedValidation
+		c.metrics.RegisterRestoreValidationFailed(backupScheduleName)
+	} else {
+		restore.Status.Phase = api.RestorePhaseInProgress
+	}
+
+	// patch to update status and persist to API
 	updatedRestore, err := patchRestore(original, restore, c.restoreClient)
 	if err != nil {
 		return errors.Wrapf(err, "error updating Restore phase to %s", restore.Status.Phase)
@@ -314,15 +313,16 @@ func (c *restoreController) processRestore(key string) error {
 	restore = updatedRestore.DeepCopy()
 
 	if restore.Status.Phase == api.RestorePhaseFailedValidation {
-		c.metrics.RegisterRestoreValidationFailed(backupScheduleName)
 		return nil
 	}
+
 	logContext.Debug("Running restore")
+
 	// execution & upload of restore
 	restoreWarnings, restoreErrors, restoreFailure := c.runRestore(
 		restore,
 		actions,
-		objectStore,
+		info,
 	)
 
 	restore.Status.Warnings = len(restoreWarnings.Ark) + len(restoreWarnings.Cluster)
@@ -355,7 +355,13 @@ func (c *restoreController) processRestore(key string) error {
 	return nil
 }
 
-func (c *restoreController) completeAndValidate(objectStore cloudprovider.ObjectStore, restore *api.Restore) []string {
+type backupInfo struct {
+	bucketName  string
+	backup      *api.Backup
+	objectStore cloudprovider.ObjectStore
+}
+
+func (c *restoreController) validateAndComplete(restore *api.Restore, pluginManager plugin.Manager) backupInfo {
 	// add non-restorable resources to restore's excluded resources
 	excludedResources := sets.NewString(restore.Spec.ExcludedResources...)
 	for _, nonrestorable := range nonRestorableResources {
@@ -363,34 +369,34 @@ func (c *restoreController) completeAndValidate(objectStore cloudprovider.Object
 			restore.Spec.ExcludedResources = append(restore.Spec.ExcludedResources, nonrestorable)
 		}
 	}
-	var validationErrors []string
 
 	// validate that included resources don't contain any non-restorable resources
 	includedResources := sets.NewString(restore.Spec.IncludedResources...)
 	for _, nonRestorableResource := range nonRestorableResources {
 		if includedResources.Has(nonRestorableResource) {
-			validationErrors = append(validationErrors, fmt.Sprintf("%v are non-restorable resources", nonRestorableResource))
+			restore.Status.ValidationErrors = append(restore.Status.ValidationErrors, fmt.Sprintf("%v are non-restorable resources", nonRestorableResource))
 		}
 	}
 
 	// validate included/excluded resources
 	for _, err := range collections.ValidateIncludesExcludes(restore.Spec.IncludedResources, restore.Spec.ExcludedResources) {
-		validationErrors = append(validationErrors, fmt.Sprintf("Invalid included/excluded resource lists: %v", err))
+		restore.Status.ValidationErrors = append(restore.Status.ValidationErrors, fmt.Sprintf("Invalid included/excluded resource lists: %v", err))
 	}
 
 	// validate included/excluded namespaces
 	for _, err := range collections.ValidateIncludesExcludes(restore.Spec.IncludedNamespaces, restore.Spec.ExcludedNamespaces) {
-		validationErrors = append(validationErrors, fmt.Sprintf("Invalid included/excluded namespace lists: %v", err))
+		restore.Status.ValidationErrors = append(restore.Status.ValidationErrors, fmt.Sprintf("Invalid included/excluded namespace lists: %v", err))
 	}
 
 	// validate that PV provider exists if we're restoring PVs
 	if boolptr.IsSetToTrue(restore.Spec.RestorePVs) && !c.pvProviderExists {
-		validationErrors = append(validationErrors, "Server is not configured for PV snapshot restores")
+		restore.Status.ValidationErrors = append(restore.Status.ValidationErrors, "Server is not configured for PV snapshot restores")
 	}
 
 	// validate that exactly one of BackupName and ScheduleName have been specified
 	if !backupXorScheduleProvided(restore) {
-		return append(validationErrors, "Either a backup or schedule must be specified as a source for the restore, but not both")
+		restore.Status.ValidationErrors = append(restore.Status.ValidationErrors, "Either a backup or schedule must be specified as a source for the restore, but not both")
+		return backupInfo{}
 	}
 
 	// if ScheduleName is specified, fill in BackupName with the most recent successful backup from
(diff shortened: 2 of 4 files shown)


pkg/controller/restore_controller.go
```diff
@@ -71,23 +71,25 @@ var nonRestorableResources = []string{
...
-	syncHandler         func(restoreName string) error
-	queue               workqueue.RateLimitingInterface
-	logger              logrus.FieldLogger
-	logLevel            logrus.Level
-	pluginRegistry      plugin.Registry
-	metrics             *metrics.ServerMetrics
+	namespace                   string
+	restoreClient               arkv1client.RestoresGetter
+	backupClient                arkv1client.BackupsGetter
+	backupStorageLocation       arkv1client.BackupStorageLocationsGetter
```
carlisia: I'm thinking now that I don't have to add this or the lister in the constructor.

pkg/cmd/server/server.go
```diff
@@ -2,7 +2,7 @@
 Copyright 2017 the Heptio Ark contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
+you may not use this file except in compliance695gg with the License.
```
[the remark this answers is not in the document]
carlisia: Can't you tell? I was trying to go to ln 695. lol

pkg/controller/restore_controller.go
carlisia: At this point I was thinking the backup here was being handled as one of the CRDs, and I needed to handle the location as a CRD too (since it is). I figured we would need to access its elements somewhere inside the restore function. But when I looked closer at the restore function it doesn't seem like that. Then I also realized the schedule isn not being handled this way.

carlisia: Yes. As I was answering your question I was realizing this is the reasonable approach. Thanks for confirming. I started already on that route and will have questions soon.

pkg/apis/ark/v1/restore.go
```diff
@@ -29,6 +29,10 @@ type RestoreSpec struct {
 	// from the most recent successful backup created from this schedule.
 	ScheduleName string `json:"scheduleName,omitempty"`
 
+	// BackupStorageLocationName is the unique name of the
```
skriss: this field is not needed and should be removed.

pkg/controller/restore_controller.go
```diff
@@ -113,22 +112,22 @@ func NewRestoreController(
...
-		queue:               workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "restore"),
-		logger:              logger,
-		logLevel:            logLevel,
-		pluginRegistry:      pluginRegistry,
-		metrics:             metrics,
+		namespace:                  namespace,
+		restoreClient:              restoreClient,
+		backupClient:               backupClient,
+		restorer:                   restorer,
+		backupLocationLister:       backupLocationInformer.Lister(),
```
skriss: easier if we keep these in the same order as the fields on the struct, so pls move `backupLocationLister` and `backupLocationListerSynced` below `restoreListerSynced`

pkg/controller/restore_controller.go
```diff
@@ -301,8 +312,10 @@ func (c *restoreController) processRestore(key string) error {
 	}
 
 	backupScheduleName := restore.Spec.ScheduleName
+	backupStorageLocationName := restore.Spec.BackupStorageLocationName
 	// Register attempts after validation so we don't have to fetch the backup multiple times
 	c.metrics.RegisterRestoreAttempt(backupScheduleName)
+	c.metrics.RegisterRestoreAttempt(backupStorageLocationName)
```
skriss: I don't think we want this

pkg/controller/restore_controller.go
```diff
@@ -464,32 +475,8 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
 
 func (c *restoreController) fetchBackup(objectStore cloudprovider.ObjectStore, name string) (*api.Backup, error) {
```
skriss: if we decide to remove the "fallback to obj storage" functionality, then we can get rid of this function entirely

ncdc: I'd recommend reviewing https://github.com/heptio/ark/issues/42

skriss: That's a helpful refresh.  My preference would be to make the sync process more intelligent -- off the top of my head, we could run it much more frequently, but have some kind of checksum or index file within the backup locations so the sync can easily detect whether or not things have changed

carlisia: I read the referred issue with a fresh mind today. At the cost of speaking the obvious, seems if we just remove the fallback then we'd be regressing and running into that undesirable use case.

I'd think doing a diff on a checksum even if very frequently should be comfortably low cost. My question at the moment is, would adding the operation to do a checksum in a backup add any significant processing cost to a single operation, and would the operations happen too frequently to make this an issue? Or we need a different (daemon) process to run thru the backups and continuously spit out the checksum somewhere? Things that I'm thinking about now.

skriss: Mulled this over the weekend, and I'm thinking we should just update this code with the naive implementation for now (if backup isn't found via Kube API, then check *each* backup location, starting with the default), and look to do an improvement on the backup sync controller as part of the "backup sharing" epic

pkg/controller/restore_controller.go
```diff
@@ -141,7 +142,7 @@ func NewRestoreController(
 
 	c.syncHandler = c.processRestore
 
-	restoreInformer.Informer().AddEventHandler(
+	backupInformer.Informer().AddEventHandler(
```
skriss: FYI this shouldn't have changed -- you should revert.  May cause challenges when testing as-is :)

carlisia: Yes, it was a find/replace snaffu. Thanks for catching! 🤦‍♀️

pkg/controller/restore_controller.go
```diff
@@ -462,43 +461,97 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
 	return nil
 }
 
-func (c *restoreController) fetchBackup(objectStore cloudprovider.ObjectStore, name string) (*api.Backup, error) {
-	backup, err := c.backupLister.Backups(c.namespace).Get(name)
-	if err == nil {
-		return backup, nil
+// fetchBackupInfo checks the object storage for a backup that matches the given name. If it doesn't
```
skriss: fetchBackupInfo checks the backup lister for a backup that matches...

pkg/controller/restore_controller.go
```diff
@@ -462,43 +461,97 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+// find it, it tries to retrieve it from one of the backup storage locations.
+func (c *restoreController) fetchBackupInfo(backupName string, pluginManager plugin.Manager) (backupInfo, error) {
+	var info backupInfo
```
skriss: inline the declarations

pkg/controller/restore_controller.go
```diff
@@ -462,43 +461,97 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+	var err error
+	info.backup, err = c.backupLister.Backups(c.namespace).Get(backupName)
+	if err != nil {
```
skriss: you removed the code that checks to see if the returned error is a 404 or not.  I'd say that if we get a non-nil, non-404 error, we should return it, rather than processing to the "fallback" code

How about organizing the logic in this order:

1. non-nil, non-404 error (i.e. `err != nil && !apierrors.IsNotFound(err)`) -- return the error
2. no error - get the BackupStorageLocation from the lister, get an object store for the location, return a `*backupInfo`
3. if we get here it means we got a 404 error - so call and return the results of `fetchFallBack`

pkg/controller/restore_controller.go
```diff
@@ -462,43 +461,97 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+		return c.fetchFallBack(backupName, pluginManager)
 	}
 
-	if !apierrors.IsNotFound(err) {
-		return nil, errors.WithStack(err)
+	if info.backup.Spec.StorageLocation == "" {
```
skriss: this block should be removed -- backups' storage locations are defaulted when the backup is created

pkg/controller/restore_controller.go
```diff
@@ -462,43 +461,97 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+}
 
-	created, createErr := c.backupClient.Backups(c.namespace).Create(backup)
-	if createErr != nil {
-		logContext.WithError(errors.WithStack(createErr)).Error("Unable to create API object for Backup")
-	} else {
-		backup = created
+// fetchFallBack checks each backup storage location, starting with the default,
+// looking for a backup that matches the given backup name.
+func (c *restoreController) fetchFallBack(backupName string, pluginManager plugin.Manager) (backupInfo, error) {
```
skriss: I'd call this `fetchFromBackupStorage`

pkg/controller/restore_controller.go
```diff
@@ -462,43 +461,97 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+	var info backupInfo
+	var err error
+
+	logContext := c.logger.WithField("backupName", backupName)
+	logContext.Debug("Backup not found in backupLister, checking each backup location directly, starting with default...")
```
skriss: I'd probably log this in the calling function rather than here

pkg/controller/restore_controller.go
```diff
@@ -462,43 +461,97 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+
+	location, err := c.backupLocationLister.BackupStorageLocations(c.namespace).Get(c.defaultBackupLocation)
```
skriss: Instead of needing a separate largely copy-pasted block of code to check the default location first, how about we get the full list of all BackupStorageLocations (per https://github.com/heptio/ark/pull/766/files#diff-7d210f9ced78ff8be1ba497bf49e1a40R512), then construct an ordered slice of locations to iterate through, where the first one is the default location, and the rest are the remainder, in any order

pkg/controller/restore_controller.go
```diff
@@ -462,43 +461,97 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+	locations, err := c.backupLocationLister.BackupStorageLocations(c.namespace).List(labels.Everything())
+	if err != nil {
+		return backupInfo{}, err
 	}
 
-	return backup, nil
+	for _, location := range locations {
+		info.objectStore, err = getObjectStoreForLocation(location, pluginManager)
+		if err != nil {
+			continue
```
skriss: we should log the error here

pkg/controller/restore_controller.go
```diff
@@ -462,43 +461,97 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+		}
+
+		info.backup, err = c.getBackup(info.objectStore, location.Spec.ObjectStorage.Bucket, backupName)
+		if err != nil {
```
skriss: we should log the error

pkg/controller/restore_controller.go
```diff
@@ -462,43 +461,97 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+		var createErr error
+		info.backup, createErr = c.backupClient.Backups(c.namespace).Create(info.backup)
+		if createErr != nil {
+			logContext.WithError(errors.WithStack(createErr)).Error("Unable to create API object for Backup")
+		} else {
+			return info, nil
+		}
+	}
+
+	return info, nil
```
skriss: you need to set `info.bucketName` before you return it from here

carlisia:
Ready for another review.

pkg/controller/restore_controller.go
```diff
@@ -462,43 +472,95 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+
+		info.backup, err = c.getBackup(info.objectStore, location.Spec.ObjectStorage.Bucket, backupName)
+		if err != nil {
+			logContext.WithError(errors.WithStack(err)).Error("Unable to find a backup in the location ", location.Name)
+			continue
+		}
+
+		// ResourceVersion needs to be cleared in order to create the object in the API
+		info.backup.ResourceVersion = ""
+		// Clear out the namespace too, just in case
```
nrb: Just in case the backup was made in a different cluster, with a different namespace? That might be worth saying explicitly

carlisia:
To note: most of the inlining suggestions were not possible. 

Also, I didn't know how to make this test pass: https://github.com/heptio/ark/pull/766/files#diff-23ca448ecec41f6d5ef89325e8a41eafR538

pkg/controller/restore_controller.go
carlisia: Will do!

carlisia:
I'll squash after another look!

pkg/controller/restore_controller.go
```diff
@@ -462,43 +472,95 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+	return info, nil
+}
+
+// fetchFromBackupStorage checks each backup storage location, starting with the default,
+// looking for a backup that matches the given backup name.
+func (c *restoreController) fetchFromBackupStorage(backupName string, pluginManager plugin.Manager) (backupInfo, error) {
+	locations, err := c.backupLocationLister.BackupStorageLocations(c.namespace).List(labels.Everything())
 	if err != nil {
-		return nil, err
+		return backupInfo{}, err
```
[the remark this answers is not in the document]
skriss: this still applies

pkg/controller/restore_controller.go
```diff
@@ -462,43 +468,92 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+	}
+
+	idx := 1
+	orderedLocations := make([]*api.BackupStorageLocation, len(locations))
+	for i := 0; i < len(locations); i++ {
```
skriss: One issue with this implementation is that if for some reason, there is no backup storage location with the default name, then `orderedLocations[0]` will never get filled in and we'll eventually get a panic (I think).  One alternative way of writing this for your consideration:
```
func orderedBackupLocations(locations []*api.BackupStorageLocation, defaultLocationName string) []*api.BackupStorageLocation {
	var result []*api.BackupStorageLocation

	for i := range locations {
		if locations[i].Name == defaultLocationName {
			// put the default location first
			result = append(result, locations[i])
			// append everything before the default
			result = append(result, locations[:i]...)
			// append everything after the default
			result = append(result, locations[i+1:]...)

			return result
		}
	}

	return locations
}
```

pkg/controller/restore_controller.go
```diff
@@ -462,43 +468,92 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
 	}
 
-	// ResourceVersion needs to be cleared in order to create the object in the API
-	backup.ResourceVersion = ""
-	// Clear out the namespace too, just in case
-	backup.Namespace = ""
+	logContext := c.logger.WithField("backupName", backupName)
+	var info backupInfo
+	for _, location := range orderedLocations {
+		info.objectStore, err = getObjectStoreForLocation(location, pluginManager)
```
skriss: within the `orderedLocations` loop, I'd start with `log := logContext.WithField("locationName", location.Name)` -- then you don't need to append it in each of the subsequent log statements within the loop (and it's a structured field)

pkg/controller/restore_controller.go
```diff
@@ -462,43 +468,92 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+		if err != nil {
+			logContext.WithError(errors.WithStack(err)).Error("Unable to get an object store in the location ", location.Name)
```
skriss: "...get an object store *for* location"

pkg/controller/restore_controller.go
```diff
@@ -462,43 +468,92 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+		}
+
+		// ResourceVersion needs to be cleared in order to create the object in the API
+		info.backup.ResourceVersion = ""
+		// Clear out the namespace, in case the backup was made in a different cluster, with a different namespace
+		info.backup.Namespace = ""
+
+		info.backup, err = c.backupClient.Backups(c.namespace).Create(info.backup)
+		if err != nil {
+			logContext.WithError(errors.WithStack(err)).Error("Unable to create API object for Backup")
```
skriss: I think we should just return the error here -- we shouldn't be continuing the loop in this scenario

pkg/controller/restore_controller.go
```diff
@@ -462,43 +468,92 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+		} else {
```
skriss: can get rid of the `else {`

pkg/controller/restore_controller.go
```diff
@@ -462,43 +468,92 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+			// backup found in a location's object store and successfully created in the API
+			info.bucketName = location.Spec.ObjectStorage.Bucket
+			return info, nil
+		}
 	}
 
-	return backup, nil
+	return backupInfo{}, errors.New("not able to fetch from backup storage")
```
skriss: ....`errors.New("backup not found in any backup storage location")`

pkg/controller/restore_controller.go
carlisia: Don't we want to move on to the next location?

pkg/controller/restore_controller.go
carlisia: Ah. I am totally confused why this is `for` (saw it in Nola's code) and not `from`. And, apparently, why not `in`. lol

pkg/controller/restore_controller.go
skriss: Don't think so - we've already found the backup we're looking for.  Continuing to look elsewhere wouldn't help

pkg/controller/restore_controller.go
carlisia: Yep. Very good point!

pkg/controller/restore_controller.go
carlisia: Ok I see. I didn't make my reasoning at all clear. Here it is: I thought a valid backup to use was one that was found + created. So this is not the case? If we create it or not doesn't matter?

skriss: So in general there *should* only be a single backup with a given name.  In which case, once we've found it, continuing to search elsewhere if we can't create it is pointless, because we've already found the one we're looking for.

In the current state, there **could** be >1 backup with a given name if the user is importing backups from other clusters.  I think you're arguing that in this case, if we find one, but the API call to sync it into the cluster fails, we should just try another one?  This is already kind of an unsupported scenario so it's a little hard to reason about, but I'd say that if creating the first one failed, then attempting to create any other one with the same name will probably fail too (maybe the API's down, maybe something's wrong with RBAC, maybe a backup with the same name got created in between when we checked for existence and when we went to sync one in, ...).  And, soon, the possibility of this scenario existing will go away.

carlisia: Ah. I really appreciate the explanation! Thank you.

pkg/controller/restore_controller.go
carlisia: This is a lot more elegant 👍

pkg/controller/restore_controller.go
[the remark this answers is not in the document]
carlisia: 😅 ok let's see if there's any other comment and maybe I can manage to get this one addressed.

pkg/controller/restore_controller.go
```diff
@@ -506,48 +506,74 @@ func (c *restoreController) fetchFromBackupStorage(backupName string, pluginMana
...
+
+			return result
 		}
 	}
 
-	return backupInfo{}, errors.New("not able to fetch from backup storage")
+	return locations
+}
+
+func (c *restoreController) backupInfoForLocation(location *api.BackupStorageLocation, backupName string, pluginManager plugin.Manager) (backupInfo, error) {
```
carlisia: I feel much better about splitting this code into a separate function. I wasn't sure if I should have a receiver or just pass the stuff I need as an argument. Any preference?

pkg/controller/restore_controller.go
```diff
@@ -506,48 +506,74 @@ func (c *restoreController) fetchFromBackupStorage(backupName string, pluginMana
...
 	for _, location := range orderedLocations {
-		info.objectStore, err = getObjectStoreForLocation(location, pluginManager)
-		if err != nil {
-			logContext.WithError(errors.WithStack(err)).Error("Unable to get an object store in the location ", location.Name)
-			continue
+		info, err = c.backupInfoForLocation(location, backupName, pluginManager)
+		if err == nil {
+			break
 		}
+		log := logContext.WithField("locationName", location.Name)
```
carlisia: @skriss you specifically said to put this line at the beginning of the loop. Does it make a difference? I put it here because it's close to where it's being used.

pkg/controller/restore_controller.go
```diff
@@ -506,48 +506,74 @@ func (c *restoreController) fetchFromBackupStorage(backupName string, pluginMana
...
+	// Clear out the namespace, in case the backup was made in a different cluster, with a different namespace
+	backup.Namespace = ""
+
+	backupCreated, _ := c.backupClient.Backups(c.namespace).Create(backup)
+	if err == nil {
+		backup = backupCreated
+	}
+	bucketName := location.Spec.ObjectStorage.Bucket
+
+	info := backupInfo{
```
carlisia: I think putting together this struct in parts all over the inside of this block was a bad idea. I like it better using local variables to fetch the values and then doing a single construction here. It feels a lot more explicit and reads better to me. And even so the bucket name is "right there", I think being consistent with all the variables/elements trumps aiming to have less code (which I also like).

pkg/controller/restore_controller.go
```diff
@@ -462,43 +468,118 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
-		return nil, err
+		return backupInfo{}, err
+	}
+
+	orderedLocations := orderedBackupLocations(locations, c.defaultBackupLocation)
+
+	logContext := c.logger.WithField("backupName", backupName)
+	var info backupInfo
+	for _, location := range orderedLocations {
+		info, err = c.backupInfoForLocation(location, backupName, pluginManager)
```
skriss: how about:
```
info, err := c.backupInfoForLocation(location, backupName, pluginManager)
if err != nil {
    logContext.WithField("locationName", location.Name).WithError(err).Error("Unable to fetch backup from object storage location")
    continue
}
return info, nil
```

Then below the loop, you can remove everything but L523 since we'll only get there if we didn't find the backup anywhere

carlisia:
@skriss I pointed somethings out in my code, please review.

pkg/controller/restore_controller.go
skriss: I think I generally agree. Though I do prefer that if we're doing something like the following:

```
bucketName := location.Spec.ObjectStorage.Bucket

info := backupInfo{
    bucketName: bucketName,
    ....
}
```

that we skip the declaration of `bucketName` and just do:
```
info := backupInfo{
    bucketName: location.Spec.ObjectStorage.Bucket,
    ....
}
```

skriss: One less variable to keep track of :)

pkg/controller/restore_controller.go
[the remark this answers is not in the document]
skriss: Missing GitHub feature: mark a code review comment as "addressed", and see all unaddressed ones.

pkg/controller/restore_controller.go
```diff
@@ -462,43 +468,118 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+	// Clear out the namespace, in case the backup was made in a different cluster, with a different namespace
 	backup.Namespace = ""
 
-	created, createErr := c.backupClient.Backups(c.namespace).Create(backup)
-	if createErr != nil {
-		logContext.WithError(errors.WithStack(createErr)).Error("Unable to create API object for Backup")
-	} else {
-		backup = created
+	backupCreated, _ := c.backupClient.Backups(c.namespace).Create(backup)
+	if err == nil {
```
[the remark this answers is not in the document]
carlisia: I really like the code like that. But question about logic, as per our other thread: if the create fails, should we not return the backup we had fetched anyway? If the answer is yes, then the above would fail this logic because in the case of a failure on `create` `backupCreated` would not be a clean backup. Right?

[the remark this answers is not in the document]
skriss: If the create fails, we should return the `backupInfo` zero value + the error.

pkg/controller/restore_controller.go
skriss: But given that we're now returning from within the loop if we successfully get a result from `backupInfoForLocation`, then if we exit the loop, we know that we failed.  So we should just be able to `return backupInfo{}, errors.New("Unable to get backup from any backup storage location")`, right?

pkg/controller/restore_controller.go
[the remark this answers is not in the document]
carlisia: Re-read your comments about this. I took `we've already found the backup we're looking for. ` to mean, "we're good with the found backup", when in reality it means we found it but it didn't work, so just ditch it. Copy that.

pkg/controller/restore_controller.go
carlisia: Ah yes, I missunderstood your suggestion.  I also personally will do anything to avoid the last return being a negative path statement but yes, this works indeed!

carlisia:
@skriss!

pkg/controller/restore_controller.go
```diff
@@ -462,43 +468,111 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
-	// Clear out the namespace too, just in case
+	// Clear out the namespace, in case the backup was made in a different cluster, with a different namespace
 	backup.Namespace = ""
 
-	created, createErr := c.backupClient.Backups(c.namespace).Create(backup)
-	if createErr != nil {
-		logContext.WithError(errors.WithStack(createErr)).Error("Unable to create API object for Backup")
-	} else {
-		backup = created
+	backupCreated, _ := c.backupClient.Backups(c.namespace).Create(backup)
```
skriss: don't ignore the returned error :)

pkg/controller/restore_controller.go
```diff
@@ -462,43 +468,111 @@ func mostRecentCompletedBackup(backups []*api.Backup) *api.Backup {
...
+	locations, err := c.backupLocationLister.BackupStorageLocations(c.namespace).List(labels.Everything())
 	if err != nil {
-		return nil, err
+		return backupInfo{}, errors.WithStack(err)
+	}
+
+	orderedLocations := orderedBackupLocations(locations, c.defaultBackupLocation)
+
+	logContext := c.logger.WithField("backupName", backupName)
+	// var info backupInfo
```
skriss: rm

pkg/controller/restore_controller.go
carlisia: pssstt don't ever tell Dave I did dis

skriss:
otherwise LGTM

skriss:
LGTM

PR #768, opened 2018-08-15
[WIP] prototype of new storage interface

skriss:
I started looking at adding prefix support and that triggered some thoughts for refactoring around our ObjectStore and related code.

The third commit here is the interesting one (the first two are just moving existing code around).  The idea is that we add what amounts to a factory function - `BackupAccessor(...)`, which takes a `BackupStorageLocation` and other things, and returns an `Accessor`, which is an interface that can get/save/delete backups & related data from a storage backend.  This interface is implemented for an object storage backend, and reuses the code that was formerly known as `BackupService`.  If/when we add other backend types (git, filesystem, etc), we can add additional implementations easily, since the controllers are no longer tightly coupled to object storage as a backend.

Would appreciate thoughts on this.  @carlisia @nrb @ncdc 

This code is by no means complete, just implemented the interfaces, the object store implementation, and one example of how a controller/etc. would use this.

I also recognize that this will not by itself solve something like git - don't want to tackle all of that now.

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -383,11 +383,6 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
 		return err
 	}
 
-	objectStore, err := getObjectStore(controller.objectStoreConfig, pluginManager)
-	if err != nil {
-		return err
-	}
-
 	var errs []error
 
 	var backupJSONToUpload, backupFileToUpload io.Reader
@@ -425,7 +420,14 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
 		controller.logger.WithError(err).Error("error closing gzippedLogFile")
 	}
 
-	if err := storage.UploadBackup(log, objectStore, bucket, backup.Name, backupJSONToUpload, backupFileToUpload, logFile); err != nil {
+	// TODO get location with name = backup.StorageLocation from lister
+	var location *api.BackupStorageLocation
+	accessor, err := storage.BackupAccessor(backup.Name, location, pluginManager, controller.logger)
+	if err != nil {
+		return err
+	}
+
+	if err := accessor.SaveBackup(backupJSONToUpload, backupFileToUpload, logFile); err != nil {
 		errs = append(errs, err)
 	}
 
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -48,6 +48,7 @@ import (
 	listers "github.com/heptio/ark/pkg/generated/listers/ark/v1"
 	"github.com/heptio/ark/pkg/metrics"
 	"github.com/heptio/ark/pkg/plugin"
+	"github.com/heptio/ark/pkg/storage"
 	"github.com/heptio/ark/pkg/util/collections"
 	"github.com/heptio/ark/pkg/util/encode"
 	kubeutil "github.com/heptio/ark/pkg/util/kube"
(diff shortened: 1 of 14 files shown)


pkg/controller/backup_controller.go
```diff
@@ -425,7 +420,14 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
...
 
-	if err := storage.UploadBackup(log, objectStore, bucket, backup.Name, backupJSONToUpload, backupFileToUpload, logFile); err != nil {
+	// TODO get location with name = backup.StorageLocation from lister
+	var location *api.BackupStorageLocation
+	accessor, err := storage.BackupAccessor(backup.Name, location, pluginManager, controller.logger)
+	if err != nil {
+		return err
+	}
+
+	if err := accessor.SaveBackup(backupJSONToUpload, backupFileToUpload, logFile); err != nil {
```
ncdc: This doesn't easily allow something like a git backend (git wouldn't use/have `backupFileToUpload`). Take a look at https://github.com/heptio/ark/commit/c07bec22ec59ccbe68ec56e2d249072b6e5d5521 for comparison. I don't think this commit is entirely correct, but maybe we can find a way to merge our two approaches into something awsome 😄

carlisia:
I looked intently at these changes and I really like the direction. When I read Andy's comment I thought of having an interface type for the backup, and it seems that's what he has done in his code. So maybe having a writer for a git backend and another for the object storage could work. I'm not completely clear if my thinking applies though.

skriss:
Thought about this some more and since I don't really want to tackle designing a generic interface that works for git right now, I'm going to not do some of the changes here. I would still like to introduce a struct that encapsulates objectStore + bucket + prefix so that we don't have to add prefix to add of the `backup_service.go` funcs - any thoughts?

nrb:
> Thought about this some more and since I don't really want to tackle designing a generic interface that works for git right now

+1, I think this should be tackled on it's own.

> I would still like to introduce a struct that encapsulates objectStore + bucket + prefix so that we don't have to add prefix to all of the backup_service.go funcs - any thoughts?

I think that makes sense.

ncdc:
> Thought about this some more and since I don't really want to tackle designing a generic interface that works for git right now, I'm going to not do some of the changes here. I would still like to introduce a struct that encapsulates objectStore + bucket + prefix so that we don't have to add prefix to all of the backup_service.go funcs - any thoughts?

This sounds kind of like what the `BackupService` was before, but maybe with some better naming and planning we could end up liking it this time around :-)

```go
type BackupStore interface {
  ListBackups() ([]*v1.Backup, error)

  PutBackupLog(backupName string, file io.Reader) error
  GetBackupLog(backupName string) (io.ReadCloser, error) // maybe not needed

  PutBackupMetadata(backup *v1.Backup) error
  GetBackupMetadata(backupName string) (*v1.Backup, error)

  PutBackupData(backupName string, file io.Reader) error
  GetBackupData(backupName string) (io.ReadCloser, error)
}

type ObjectStorageBackupStore struct {
  objectStore cloudprovider.ObjectStore
  bucket string
  prefix string
}
```

This might even be on the right path to "generic enough" for git too?

skriss:
Yeah, what you have is pretty much what I was thinking.  Wanted your input since you got rid of the `BackupService` - but I think this revision is reasonable -- not a singleton, instantiated per request/location, and keeps the method signatures much simpler

skriss:
thanks for the input, closing this PR

PR #769, opened 2018-08-16
Update backup-deletion and download-request controllers for backup locations

### LINKED ISSUE #741: Update backup deletion controller to use the backup's specified location

### LINKED ISSUE #743: Update download request controller to use the backup's specified location

skriss:
Fixes #743 
Fixes #741 

@nrb @carlisia we'll need to do a little bit of coordination between all of our controller work so let's not merge this until we're all a little further and can discuss.

### DIFF
--- a/pkg/controller/backup_deletion_controller.go
+++ b/pkg/controller/backup_deletion_controller.go
@@ -28,6 +28,7 @@ import (
 	arkv1client "github.com/heptio/ark/pkg/generated/clientset/versioned/typed/ark/v1"
 	informers "github.com/heptio/ark/pkg/generated/informers/externalversions/ark/v1"
 	listers "github.com/heptio/ark/pkg/generated/listers/ark/v1"
+	"github.com/heptio/ark/pkg/plugin"
 	"github.com/heptio/ark/pkg/restic"
 	"github.com/heptio/ark/pkg/util/kube"
 	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


skriss:
I also spent some time cleaning up the test for the download request controller, using some ideas I had for how to simplify our tests and make them less brittle.  I'd like to discuss this w/everyone and see if the patterns makes sense and should be extended to other tests.

carlisia:
I looked some more at the `TestProcessDownloadRequest` and have a couple questions. The `expectedErr` variable is `nil` for all test cases, not sure if this is intentional. I would expect that there'd be test cases that would have errors. But I understand this is a wip.

Another question is: there is this block:
```
			// verify results
			if tc.expectedErr != nil {
				require.EqualError(t, err, tc.expectedErr.Error())
			} else {
				assert.Nil(t, err)
			}
```

and I would expect that the code that comes after would go inside one or the two paths, no?

pkg/controller/backup_deletion_controller.go
```diff
@@ -312,6 +319,27 @@ func (c *backupDeletionController) processRequest(req *v1.DeleteBackupRequest) e
...
+	defer pluginManager.CleanupClients()
+
+	backupLocation, err := c.backupLocationLister.BackupStorageLocations(backup.Namespace).Get(backup.Spec.StorageLocation)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	objectStore, err := getObjectStoreForLocation(backupLocation, pluginManager)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
skriss: The pattern we follow is that when Ark code is creating an error, we use `github.com/pkg/errors` to create it (which automatically includes a stack), and if we're handling an error returned from external code, we wrap it. In this case, since we're handling an error from an Ark function, it should already be wrapped/with a stack so we don't need to do it again.

skriss:
yep, forgot to add error cases.  added now

skriss:
Fixed an issue with setting log levels, and this is now ready for review.  I'd like to squash before merging so don't merge until I do so!!

pkg/controller/backup_deletion_controller.go
```diff
@@ -449,3 +478,20 @@ func (c *backupDeletionController) patchBackup(backup *v1.Backup, mutate func(*v
 
 	return backup, nil
 }
+
+func getObjectStoreForLocation(location *v1.BackupStorageLocation, manager plugin.Manager) (cloudprovider.ObjectStore, error) {
```
carlisia: Not sure if it needs to be addressed and if so if it should be in this PR: we use the alias `api` in other places, and `v1` here. Example: https://github.com/heptio/ark/blob/60c6624d60bb23c494aeca8d16c25073d4a7717a/pkg/controller/backup_controller.go#L475

Do we want to make it all the same?

carlisia: I am surprised GH is not saying there's a conflict since Nola's PR has been merged with this same function..

nrb: > I am surprised GH is not saying there's a conflict since Nola's PR has been merged with this same function..

Probably because it's a different file. I expect the conflict will be caught at build time.

skriss:
rebased, did some squashing, and address comment on comments.

skriss:
awesome, tested this rebased on @nrb's PR and backup, download requests (logs), and deletions all seem to work correctly for multiple locations!

skriss:
@nrb @carlisia I believe I addressed all comments, plus did some testing, so this should be ready to merge when LGTY.

PR #771, opened 2018-08-16
Use backup location in the backup controller

nrb:

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -57,21 +57,22 @@ import (
 const backupVersion = 1
 
 type backupController struct {
-	backupper         backup.Backupper
-	objectStoreConfig api.CloudProviderConfig
-	bucket            string
-	pvProviderExists  bool
-	lister            listers.BackupLister
-	listerSynced      cache.InformerSynced
-	client            arkv1client.BackupsGetter
-	syncHandler       func(backupName string) error
-	queue             workqueue.RateLimitingInterface
-	clock             clock.Clock
-	logger            logrus.FieldLogger
-	logLevel          logrus.Level
-	pluginRegistry    plugin.Registry
-	backupTracker     BackupTracker
-	metrics           *metrics.ServerMetrics
+	backupper                  backup.Backupper
+	pvProviderExists           bool
+	lister                     listers.BackupLister
+	listerSynced               cache.InformerSynced
+	client                     arkv1client.BackupsGetter
+	syncHandler                func(backupName string) error
+	queue                      workqueue.RateLimitingInterface
+	clock                      clock.Clock
+	logger                     logrus.FieldLogger
+	logLevel                   logrus.Level
+	pluginRegistry             plugin.Registry
+	backupTracker              BackupTracker
+	backupLocationLister       listers.BackupStorageLocationLister
+	backupLocationListerSynced cache.InformerSynced
+	defaultBackupLocation      string
+	metrics                    *metrics.ServerMetrics
 
 	newPluginManager func(logger logrus.FieldLogger, logLevel logrus.Level, pluginRegistry plugin.Registry) plugin.Manager
 }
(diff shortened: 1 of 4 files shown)


pkg/controller/backup_controller.go
```diff
@@ -80,30 +79,28 @@ func NewBackupController(
...
+		lister:               backupInformer.Lister(),
+		listerSynced:         backupInformer.Informer().HasSynced,
+		client:               client,
+		queue:                workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "backup"),
+		clock:                &clock.RealClock{},
+		logger:               logger,
+		logLevel:             logLevel,
+		pluginRegistry:       pluginRegistry,
+		backupTracker:        backupTracker,
+		backupLocationLister: backupLocationInformer.Lister(),
```
skriss: you'll need to wait for the `backupLocationInformer` to sync in the `Run()` func like is being done for the `backupInformer`

pkg/controller/backup_controller.go
```diff
@@ -80,30 +79,28 @@ func NewBackupController(
...
-		listerSynced:      backupInformer.Informer().HasSynced,
-		client:            client,
-		queue:             workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "backup"),
-		clock:             &clock.RealClock{},
-		logger:            logger,
-		logLevel:          logLevel,
-		pluginRegistry:    pluginRegistry,
-		backupTracker:     backupTracker,
-		metrics:           metrics,
+		backupper:            backupper,
```
skriss: it would be nice to convert the controllers to use the `genericController` type while we're in there modifying them.  This requires:

- embedding a `*genericController` in the `backupController`
- initializing it within the `NewBackupController` func using the `newGenericController` func
- removing the `syncHandler` field from the `backupController` (`genericController` has its own) 
- setting the `cacheSyncWaiters` field in the constructor (removes the need to store each of them as their own fields)
- removing the `queue` field
- removing the `Run`, `runWorker`, and `processNextWorkItem` funcs

I did the same for the download request controller -- see https://github.com/heptio/ark/pull/769/files#diff-cdd2d4325fc0100b6d5a4039fc99792e -- although I think I missed at least one step now that I'm testing.

pkg/controller/backup_controller.go
```diff
@@ -382,7 +384,12 @@ func (controller *backupController) runBackup(backup *api.Backup, bucket string)
 		return err
 	}
 
-	objectStore, err := getObjectStore(controller.objectStoreConfig, pluginManager)
+	backupLocation, err := controller.backupLocationLister.BackupStorageLocations(backup.Namespace).Get(backup.Spec.StorageLocation)
```
skriss: I would just pass `backupLocation` into this func in place of `bucket`, then you won't have to fetch it again

pkg/controller/backup_controller.go
```diff
@@ -458,6 +465,24 @@ func getObjectStore(cloudConfig api.CloudProviderConfig, manager plugin.Manager)
 	return objectStore, nil
 }
 
+// TODO(nrb): Consolidate with other implementations
```
skriss: @nrb @carlisia let's say that whichever PR of our 3 merges first will include a function with a new name, like @nrb has here, for getting an object store from a location.  The other 2 of us will then rebase and use that function.  The last person to merge can drop the old function, and rename the new one if they want (or not).  Work for everyone?

nrb: That makes sense to me.

carlisia: Works for me!

pkg/controller/backup_controller.go
nrb: Is this just using a `Synced` field and updating the condition on L165 to look for it? Or do I need a separate condition?

skriss: yep add a `backupLocationListerSynced` field to the struct, set it in the constructor, then add it as an arg to the `WaitForCacheSync(...)` call (it's variadic)

pkg/controller/backup_controller.go
skriss: If you want to defer this to a separate PR I'm ok with that, but let's get it done for 0.10

nrb: I hadn't looked at it yet; want me to push the other fixes then see where we're at?

skriss: sure

skriss:
Just noticed that you'll need to add defaulting as well - if no location is specified on the API obj, it should default to the server's `defaultBackupLocation` value

nrb:
> Just noticed that you'll need to add defaulting as well

Good catch, will add it + a test case.

nrb:
Got default location setting working, but I think I'll need to improve error handling when the location isn't found. Probably make the backup failed.

> ERRO[0095] Error in syncHandler, re-adding item to queue  error="backupstoragelocation.ark.heptio.com \"default\" not found" error.file="backup_controller.go:297" error.function="(*backupController).processBackup" key=heptio-ark/b1 logSource="pkg/controller/backup_controller.go:212"

skriss:
Nice - yeah, would be ideal if that could be checked during validation and it's moved to `FailedValidation` if the location doesn't exist

nrb:
I've updated to set the default location if the field is empty on the backup, and added some tests around missing locations.

I'm not super happy with how the tests are implemented, so open to suggestions there.

pkg/controller/backup_controller_test.go
```diff
@@ -224,6 +244,43 @@ func TestProcessBackup(t *testing.T) {
 					mock.Anything, // actions
 				).Return(nil)
 
+				var location *v1.BackupStorageLocation
```
skriss: What I would do here is always add two locations to the informer - `default` and `loc1`.  Then, keep the 3 test cases you have, but change the third to use location `loc2` (which won't exist).  Your verification can stay as-is but the setup will be more straightforward

skriss:
The rest LGTM - will take a final look Monday but should be good to merge

nrb:
Done. I think it will be even cleaner with the `newBackupLocation` function, but no real need to hold this up just for that, either.

Issue #770, opened 2018-08-16
Ability to retrieve Ark server version


nrb:
Right now, it's not immediately apparent that the output of `ark version` is the client only, and getting the version of the server is done via `kubectl`.

We should consider adding an endpoint so the client can report the version, much in the same way `kubectl version` works.

nrb:
@mattmoyer Hey Matt, do you see this kind of feature being a security concern? One point that was brought up was that if there's an endpoint for the version, a client could do version sniffing and then use known vulnerabilities.

I'm wondering if it's really that much different than `kubectl version` and our current way of looking up the server version - `kubectl logs deploy/ark -n heptio-ark`, which has the version in the first few lines of output.

ncdc:
Thinking about this more, it's probably ok (especially since `kubectl version` reports similar information about the server).

Getting the request from the client to the ark server pod could be challenging, depending on the networking setup. We should assume that most invocations of `ark version` happen outside the Kubernetes cluster (e.g., on your laptop). This means there needs to be some sort of ingress into the cluster - a regular `Service` won't be sufficient. We _could_ consider attempting to proxy through the apiserver, but many/most clusters may have that locked down and unavailable.

We may want to consider adding new configuration options to `ark client config set`, such as (names TBD):
- serverUrl
- useKubernetesProxy

We also will want to be careful not to expose the metrics endpoint - presumably that should remain private/protected (although as I understand it, if you run a pod on the cluster and there aren't any network policy rules, the pod can access the metrics endpoint, and we don't have any authn/authz for it...).

cc @rosskukulinski

ncdc:
Decision: add a `Status` CRD that for now just contains the server version. Update the `ark version` command to use that to get the server's version. We'll create a new issue for an `ark status` command with additional details.

PR #774, opened 2018-08-17
add an ark bug command

### LINKED ISSUE #578: RFE: ark bug command

metadave:
This PR adds the `ark bug` subcommand, which gathers a few bits of information and opens a new browser window with the Github issue template partially populated. 

- some of this is based on the existing `./hack/update-*` and `./hack/verify-*` scripts, so hopefully it's not too off base.
- The Github issue golang template is stored as a constant in `pkg/cmd/cli/bug/bug.go`. 
    - I experimented with using a standalone template file, but ran into a backtick escaping nightmare and ETOOCOMPLEX. 
- The `hack/update-generated-issue-template.sh` script renders the template (without populating any template vars) to `.github/ISSUE_TEMPLATE/bug_report.md`, and the `hack/verify-generated-issue-template.sh` will ensure that the `bug_report.md` file wasn't updated by hand. This will happen automatically during testing by way of `./hack/verify-all.sh`.
    - This means that changes to the [issue template](https://github.com/heptio/ark/blob/master/.github/ISSUE_TEMPLATE/bug_report.md) must be made in `pkg/cmd/cli/bug/bug.go`. 

- I realize that outside of the `./hack/verify-generated-issue-template.sh` script, there isn't any testing. I'm open to suggestions on ways to test this without jumping through too many hoops :-)

- I experimented with downloading the template, however I feel as though it's better to embed directly into the binary as the struct or the template itself could change and cause a mismatch between `ark` client versions. One way around this is to link a particular hosted version, but it seems like overkill. Also, it's hard to submit a PR that depends on a hosted file that's in the same PR :-)

- Eventually, this could include an ark server version via #770

see also https://github.com/heptio/ark/pull/713

Closes #578

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -27,7 +27,7 @@ about: Tell us about a problem you are experiencing
 
 **Environment:**
 
-- Ark version (use `ark version`):
+- Ark version (use `ark version`):  
 - Kubernetes version (use `kubectl version`):
 - Kubernetes installer & version:
 - Cloud provider or hardware configuration:
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -2,6 +2,8 @@
 
 These tips can help you troubleshoot known issues. If they don't help, you can [file an issue][4], or talk to us on the [#ark-dr channel][25] on the Kubernetes Slack server. 
 
+In `ark` version >= `0.1.0`, you can use the `ark bug` command to open a [Github issue][4] by launching a browser window with some prepopulated values. Values included are OS, CPU architecture, `kubectl` client and server versions (if available) and the `ark` client version. This information isn't submitted to Github until you click the `Submit new issue` button in the Github UI, so feel free to add, remove or update whatever information you like.
+
 Some general commands for troubleshooting that may be helpful:
 
 * `ark backup describe <backupName>` - describe the details of a backup
--- a/pkg/cmd/cli/bug/bug.go
+++ b/pkg/cmd/cli/bug/bug.go
@@ -1,0 +1,203 @@
+/*
+Copyright 2018 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bug
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"net/url"
+	"os"
+	"os/exec"
+	"runtime"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/heptio/ark/pkg/buildinfo"
+	"github.com/heptio/ark/pkg/cmd"
+	"github.com/spf13/cobra"
+)
+
+const (
+	// kubectlTimeout is how long we wait in seconds for `kubectl version`
+	// before killing the process
+	kubectlTimeout = 5 * time.Second
+	issueURL       = "https://github.com/heptio/ark/issues/new"
+	// IssueTemplate is used to generate .github/ISSUE_TEMPLATE/bug_report.md
+	// as well as the initial text that's place in a new Github issue as
+	// the result of running `ark bug`.
+	IssueTemplate = `---
+name: Bug report
+about: Tell us about a problem you are experiencing
+
+---
+
+**What steps did you take and what happened:**
+[A clear and concise description of what the bug is, and what commands you ran.)
+
+
+**What did you expect to happen:**
+
+
+**The output of the following commands will help us better understand what's going on**:
+(Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)
+
+* ` + "`kubectl logs deployment/ark -n heptio-ark`" + `
+* ` + "`ark backup describe <backupname>` or `kubectl get backup/<backupname> -n heptio-ark -o yaml`" + `
+* ` + "`ark backup logs <backupname>`" + `
+* ` + "`ark restore describe <restorename>` or `kubectl get restore/<restorename> -n heptio-ark -o yaml`" + `
+* ` + "`ark restore logs <restorename>`" + `
+
+
+**Anything else you would like to add:**
+[Miscellaneous information that will assist in solving the issue.]
+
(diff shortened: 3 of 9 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -27,7 +27,7 @@ about: Tell us about a problem you are experiencing
 
 **Environment:**
 
-- Ark version (use `ark version`):
+- Ark version (use `ark version`):
```
metadave: whitespace was tricky

pkg/cmd/cli/bug/bug.go
```diff
@@ -0,0 +1,198 @@
...
+			err = exec.Command(linuxBrowserLaunchCmd, url).Start()
+		} else {
+			err = fmt.Errorf("Ark can't open a browser window using the command '%s'", linuxBrowserLaunchCmd)
+		}
+	case "windows":
+		err = exec.Command("rundll32", "url.dll,FileProtocolHandler", url).Start()
+	default:
+		err = fmt.Errorf("Ark can't open a browser window on platform %s", os)
+	}
+	return err
```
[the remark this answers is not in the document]
metadave: Your proposed changes are indeed easier to read :-) Thank you for the feedback! 

I generally introduce a constant if I have to type a string value more than once or twice, as I've been burned by typos in the past. However, I agree with your comment above and will apply that change.

pkg/cmd/cli/bug/bug.go
```diff
@@ -0,0 +1,194 @@
...
+	if _, err := exec.LookPath(name); err != nil {
+		return false
+	}
+	return true
+}
+
+// Make a best-effort to run `kubectl version` and return it's output.
+// This func will Timeout and return an empty string after
+// kubectlTimeout if we're not connected to a cluster.
+func getKubectlVersion() (kubectlVersion string) {
```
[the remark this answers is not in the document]
carlisia: To be consistent with the rest of the code base, please remove the naked return from this method's signature. See the other comment elsewhere, so this should end up being: `getKubectlVersion() (string, error)`

pkg/cmd/cli/bug/bug.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	done := make(chan error, 1)
+	go func() {
+		done <- kubectlCmd.Wait()
+	}()
+	select {
+	case <-time.After(kubectlTimeout):
+		// we don't care about the possible error returned from Kill() here,
+		// just return an empty string
```
carlisia: Thanks for adding a clarifying comment.

pkg/cmd/cli/bug/bug.go
```diff
@@ -0,0 +1,194 @@
...
+{{if .RuntimeArch}}	- RuntimeArch: {{.RuntimeArch}}{{end -}}
+`
+)
+
+func NewCommand() *cobra.Command {
+	c := &cobra.Command{
+		Use:   "bug",
+		Short: "Report an Ark bug",
+		Long:  "Open a browser window to report an Ark bug",
+		Run: func(c *cobra.Command, args []string) {
```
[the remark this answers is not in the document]
metadave: I cleaned up the way `getKubectlVersion` returns errors, but I'd prefer not to use `cmd.CheckError(err)` on the result, as not being connected to a cluster would prevent a user from opening a new bug.

For example, I have access to several clusters, but most of the time I can't access those via the `kubectl` unless I unencrypt the kubeconfig and/or open the firewall.

metadave:
@carlisia thank you for the insightful review (and on a weekend no less!). Changes pushed.

pkg/cmd/cli/bug/bug.go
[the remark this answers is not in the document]
carlisia: Ah. I like this approach. We are not halting the program at all. Good alternative in my view.

pkg/cmd/cli/bug/bug.go
```diff
@@ -0,0 +1,204 @@
...
+			return "", errors.New("error waiting for kubectl process")
+		}
+	}
+	versionOut := outbuf.String()
+	kubectlVersion := strings.TrimSpace(string(versionOut))
+	return kubectlVersion, nil
+}
+
+func newBugInfo(kubectlVersion string) *ArkBugInfo {
+	bugInfo := ArkBugInfo{
```
carlisia: Shoot, I forgot to add this comment: I'm pretty sure this variable here is redundant, you may `s/bugInfo :=/return`.

metadave: good catch, fixed!

carlisia:
Awesome, thank you  🎉. Let's get one more 👍 from the team.

pkg/cmd/cli/bug/bug.go
```diff
@@ -0,0 +1,203 @@
...
+
+const (
+	// kubectlTimeout is how long we wait in seconds for `kubectl version`
+	// before killing the process
+	kubectlTimeout = 5 * time.Second
+	issueURL       = "https://github.com/heptio/ark/issues/new"
+	// IssueTemplate is used to generate .github/ISSUE_TEMPLATE/bug_report.md
+	// as well as the initial text that's place in a new Github issue as
+	// the result of running `ark bug`.
+	IssueTemplate = `---
```
nrb: @jhamilton1 This will be the location for the bug template, if we merge this. What it means is that the Go file will have to be updated, then we run the `update-generated-issue-template.sh` script in this repo to get the file that GitHub uses. That enables us to not repeat the template in 2 files.

Is this approach ok with you?

jhamilton1: Works for us @nrb

nrb: Awesome, thanks!

metadave:
`troubleshooting.md` updated. Thanks for the review @nrb!

docs/troubleshooting.md
```diff
@@ -2,6 +2,8 @@
 
 These tips can help you troubleshoot known issues. If they don't help, you can [file an issue][4], or talk to us on the [#ark-dr channel][25] on the Kubernetes Slack server. 
 
+In `ark` version >= `0.1.0`, you can use the `ark bug` command to open a [Github issue][4] by launching a browser window with some prepopulated values. Values included are OS, CPU architecture, `kubectl` client and server versions (if available) and the `ark` version. This information isn't submitted to Github until you click the `Submit new issue` button in the Github UI, so feel free to add, remove or update whatever information you like.
```
nrb: Should probably state that it's the ark client version, at least for now.

metadave: fixed

PR #776, opened 2018-08-20
use the default backup storage location for restic

skriss:

Replace the `resticLocation` field on BackupStorageProvider with an opaque parameter in the `config` map on `BackupStorageLocation`.

For the time being, all restic backups should go into the default location.  We'll need to consider if/how to support restic backups going to multiple locations during the volume snapshot epic.

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -253,6 +253,11 @@ func (s *server) run() error {
 
 	s.watchConfig(originalConfig)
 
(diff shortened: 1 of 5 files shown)


nrb:
Should this piggyback on a BackupStorageLocation, or does it justify using a VolumeLocation?

Restic might be tricky since it straddles the two.

skriss:
We're gonna need to discuss more since we forgot to talk much about it during the design sessions. It's going to be tricky when it gets to replication, bc it's not really possible to replicate an individual restic snaphot

skriss:
grr, forgot to do a few other things here, hold for a bit.

skriss:
Alright, this should be ready for review now.

carlisia:
lgtm 👍

PR #782, opened 2018-08-21
update sync controller for backup locations

### LINKED ISSUE #742: Update backup sync controller to sync from all locations

skriss:

Fixes #742 

Still need to update tests, and do some manual testing, but feel free to take a look if you want.


pkg/controller/backup_controller.go
```diff
@@ -349,6 +349,13 @@ func (controller *backupController) getLocationAndValidate(itm *api.Backup, defa
 		itm.Spec.StorageLocation = defaultBackupLocation
 	}
 
+	// add the storage location as a label for easy filtering later.
```
nrb: Should we also add this when creating the backup?

skriss: in the CLI code you mean? Wouldn't hurt, though everything should end up going thru this code path anyway

nrb: I actually meant the controller. It seems a little disjointed to set information for a new backup in both the backup controller and sync controller, to me. I do think this should stay for backups imported from a newly added location, though.

skriss: This is the backup controller

skriss: Maybe I'm confused about your comment.

nrb: :man_facepalming: nevermind then :)

skriss: Ah, you mean additionally do this in the backup sync controller when importing? Yeah, that's probably necessary.  I'm thinking we are also going to need to rewrite the `StorageLocation` field since the name of it in the importing cluster may be different than the name in the source cluster.

skriss:
Some open questions:
- should we rewrite the backup's .spec.storageLocation to match the name on the importing cluster, when a backup is imported? I think yes, because if the storage location had a different name in cluster A than it does in cluster B, we want the backup to have cluster B's name for it when it's imported
- do we need to add/update the storage location label when importing a backup? I think yes

skriss:
Tests are now updated

pkg/controller/backup_controller.go
nrb: > I'm thinking we are also going to need to rewrite the StorageLocation field since the name of it in the importing cluster may be different than the name in the source cluster.

Right, that's what I thought this was doing, but thought I was reading the sync controller.

nrb:
> - should we rewrite the backup's .spec.storageLocation to match the name on the importing cluster, when a backup is imported? I think yes, because if the storage location had a different name in cluster A than it does in cluster B, we want the backup to have cluster B's name for it when it's imported

Yes, I agree. The cluster's view should be consistent.

> - do we need to add/update the storage location label when importing a backup? I think yes

I agree.

skriss:
Okey doke, implemented those things, added a test case, and fixed a dumb bug in my backup_controller code and updated the test for it.

skriss:
Will squash before merge!!

carlisia:
Nice tests! 👍 
Merging..

PR #783, opened 2018-08-22
Add DigitalOcean to S3-compatible backup providers

nrb:

### DIFF
--- a/docs/support-matrix.md
+++ b/docs/support-matrix.md
@@ -19,6 +19,7 @@ _Note that these providers are not regularly tested by the Ark team._
  * [IBM Cloud][5]
  * [Minio][9]
  * Ceph RADOS v12.2.7
+ * [DigitalOcean][7]
 
 ## Volume Snapshot Providers
 | Provider                         | Owner           | Contact                         |


docs/support-matrix.md
```diff
@@ -47,4 +48,5 @@ After you publish your plugin, open a PR that adds your plugin to the appropriat
 [9]: quickstart.md
 [10]: https://kubernetes.slack.com/messages/ark-dr
 [11]: https://github.com/heptio/ark/issues
-[12]: https://github.com/aws/aws-sdk-go/aws
\ No newline at end of file
+[12]: https://github.com/aws/aws-sdk-go/aws
+[13]: https://github.com/StackPointCloud/ark-plugin-digitalocean/blob/master/examples/10-ark-config.yaml#L15
```
rosskukulinski: it seems brittle to me to link directly to source code here.  I think we should re-use the existing link (number 7)

skriss:
LGTM.  It would be nice if we could get a more specific docs page/section to link to that specifically walks through how to configure ark to use DO Spaces as an object store, since the README in the repo mostly focuses on the block store, which might be confusing to folks.  But fine merging for now and can update the link later if our plugin contributors want to do that.

rosskukulinski:
@skriss are you thinking about having additional docs for Spaces in Ark docs (like we have for `Run on AWS`, `IBM Cloud Object Store`, etc)?

skriss:
@rosskukulinski I'd rather have it in the SPC repo - but a focused page or section about configuring Ark for using DO spaces.  Their README does cover it but it's mixed in with a lot of content about how to use the block store plugin, which may not be relevant if the user doesn't want the block store plugin.

PR #790, opened 2018-08-23
Azure changes to allow resource group/storage account to vary per backup location

skriss:
In the current state, Ark gets an Azure storage account name and access key via the `cloud-credentials` secret, which the admin sets up at install time.

This is insufficient now that we have backup storage locations, because the user may want to use blob containers that are in different storage accounts (which includes wanting to use two different resource groups, since a storage account is contained within a resource group).

This PR: 
- moves the storage account name out of the `cloud-credentials` secret and into the provider config map on `BackupStorageLocation`.
- adds `resourceGroup` as an expected value in the provider config map on `BackupStorageLocation`, since it (a) is not required to be the same as the resource group in the `cloud-credentials` secret (which specifies the RG where the cluster/persistent volumes are); and (b) should not be required to be the same across all `BackupStorageLocations`.
- removes the requirement for the user to specify a storage account access key.  Instead, an access key is fetched from the Azure API at runtime, using the service principal credentials that continue to be provided by the admin via the `cloud-credentials` secret.  This is inline with [Azure's recommendations](https://docs.microsoft.com/en-us/azure/storage/common/storage-security-guide#granting-access).
- bumps the Azure Go SDK version by a minor version, to get access to the storage `AccountsClient` that enables fetching storage account access keys.


skriss:
Docs and examples will need to be updated as well, but wanted to get a review on the approach first.

skriss:
@ncdc if you've got time would appreciate a quick review on this

ncdc:
Looking now

pkg/cloudprovider/azure/object_store.go
```diff
@@ -37,10 +41,57 @@ func NewObjectStore(logger logrus.FieldLogger) cloudprovider.ObjectStore {
 	return &objectStore{log: logger}
 }
 
+func getStorageAccountsClient(envVars map[string]string) (*storagemgmt.AccountsClient, error) {
+	spt, err := helpers.NewServicePrincipalTokenFromCredentials(envVars, azure.PublicCloud.ResourceManagerEndpoint)
```
ncdc: We'll need to copy the implementation of `helpers.NewServicePrincipalTokenFromCredentials` and adjust it slightly. It currently panics if it gets an error creating the oauth config.

I would also recommend that we check `envVars` and make sure that `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, and `AZURE_CLIENT_SECRET` are set before proceeding, or return an error if they're not. Or, perhaps better yet, make `getStorageAccountsClient` take in the 4 required parameters (the 3 previously mentioned, plus the subscription id key) or a struct containing them. Or maybe have it take in `(spt *adal.ServicePrincipalToken, subscriptionIDKey string)` where you get the `spt` before calling it.

pkg/cloudprovider/azure/object_store.go
```diff
@@ -37,10 +41,57 @@ func NewObjectStore(logger logrus.FieldLogger) cloudprovider.ObjectStore {
...
+
+	if storageKey == "" {
+		return "", errors.New("No storage key with Full permissions found")
+	}
+
+	return storageKey, nil
+}
+
 func (o *objectStore) Init(config map[string]string) error {
-	cfg := getConfig()
```
ncdc: Should we first validate that `config` has the required keys - `resourceGroup` and `storageAccount`?

pkg/cloudprovider/azure/object_store.go
```diff
@@ -37,10 +41,57 @@ func NewObjectStore(logger logrus.FieldLogger) cloudprovider.ObjectStore {
...
+		return "", errors.WithStack(err)
+	}
+	if res.Keys == nil || len(*res.Keys) == 0 {
+		return "", errors.New("No storage keys found")
+	}
+
+	var storageKey string
+
+	for _, key := range *res.Keys {
+		// uppercase both strings for comparison because the ListKeys call returns e.g. "FULL" but
```
ncdc: Do we need to update the docs to make it 100% clear that we require a storage account key with "full" permissions? Could we get by with fewer?

ncdc: Hmm I see that now, too. I was looking at your link on Azure key management practices, and it mentioned different storage account RBAC permissions, including owner, contributor, reader, etc.

pkg/cloudprovider/azure/object_store.go
skriss: Makes sense -- we're already using that `helpers` func in the `block_store` so will need to change it there too.

skriss:
PS I still need to test with the revisions so don't merge yet!

skriss:
This is looking good - tested across multiple storage accounts, regions, and resource groups.

skriss:
@ncdc will you be able to re-review this PR, or should we proceed without?

ncdc:
@skriss I can review it later today or tomorrow, if you can wait

skriss:
yep, that works. thanks!

skriss:
@nrb FYI I added a documentation commit to this PR now that your docs PR is in

ncdc:
Generally LGTM. I'm almost wondering if `getRequiredValues` should instead be an `ensure` style function, where instead of constructing a new `map`, all it does is check if the desired keys exist. Either way is probably fine. If it's an `ensure` function, you have to do a little more work to call `os.Getenv` multiple times to use the individual values, but you also would avoid the unnecessary creation of the ignored map when you ensure the config values exist for initializing the object store.

PR #796, opened 2018-08-27
Locations cleanup

skriss:
This PR does some minor cleanup on the backup-locations work.  The main change is to standardize/simplify how controllers construct a plugin manager.  They all now take a `newPluginManager` function pointer as an argument, that only requires a logger to be passed at execution time.

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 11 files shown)


nrb:
Is this to aid in testing the controllers? That seems to be the biggest change that I see looking over it.

skriss:
Well, passing in a `newPluginManager` function pointer is to make testing easier, so that function can be replaced with a test fake - but we were already doing that.  In the current state, though, we're passing a `pluginRegistry` and `logLevel` to most/all of the controllers too, but they're usually only being used as arguments to pass to the `newPluginManager` func.  This change constructs a closure, in `server.go`, that encloses a `pluginRegistry` and `logLevel` -- so now the controllers don't need to take those things as arguments, and the `newPluginManager` signature is simpler.  

tldr -- simplification and standardization.

skriss:
To clarify, the controllers already had a `newPluginManager` function pointer field, which we were setting in the constructors, reason being for testing.  This PR passes it in, as a closure.

PR #799, opened 2018-08-28
Replace config.backupStorageProvider with backup storage locations

skriss:
This PR merges the `backup-locations` feature branch into `master`.  All of the commits contained here have already been reviewed as they've merged into the feature branch.

Question for everyone - would you prefer to see this squashed, or left as-is?  I'm inclined to leave it as-is, as I think we squashed to reasonable commits when merging into the feature branch.


nrb:
I would prefer the commits as is.

Issue #451, opened 2018-04-23
Adjust AWS cluster tags when restoring volumes


ncdc:
Resources for Kubernetes clusters running in AWS have AWS tags identifying the cluster to which they belong. Some clusters may run with an IAM policy that limits the control plane so it's only allowed to attach volumes tagged with the appropriate cluster name (such as kops).

When Ark creates a volume from a snapshot, it needs to adjust the new volume's tag for the cluster name to be the name of the target cluster.

I'm not yet sure the best way to represent the cluster name's tag value when doing a restore. AFAIK that value is not guaranteed to be available with `kubectl get`. We could potentially try to query the ec2 metadata for the node where the Ark pod is running, but that's not a great idea because it could (and probably should) be blocked. We could require the user set an environment variable on the Ark deployment/pod that contains the value. That might be our best option?

skriss:
We could put it in the Ark config (could go within `.persistentVolumeProvider.config`) - although I know long-term we're trying to gut that

ncdc:
Yeah, I don't want to touch `persistentVolumeProvider`.

I don't think it would make sense to put it in the to-be `VolumeTarget` since that's meant to deal with replication.

I think it needs to go on something specific to the `heptio-ark` namespace in the target cluster. Could be an annotation, label, env var on the Ark pod. Could be a ConfigMap or some other resource in the ns...

mauilion:
+1 on the environment option. Until something like cluster-api or cloud-controller-manager provide this info consistently I think we'd have to ask.

ncdc:
Agreed - I think an env var is an "easy win" for starters.

ncdc:
I think this is the logic we want:

1. Ark admin sets `AWS_CLUSTER_NAME` environment variable in the Ark deployment (e.g. `ABC`)
1. When restoring, Ark's AWS `blockStore` code, in `CreateVolumeFromSnapshot`:
    1. Removes all `kubernetes.io/cluster/*` tags
    1. Sets `KubernetesCluster` tag to cluster name from env var (`ABC`) (this is the "legacy" tag name)
    1. Sets `kubernetes.io/cluster/ABC` tag to `owned`

ncdc:
Closed by #801

skriss:
@mqsoh do you have `AWS_CLUSTER_NAME` set in your deployment?

mqsoh:
[...]

PR #801, opened 2018-08-29
Claim ownership of new AWS volumes by Kubernetes cluster restoring the backup

ljakimczuk:
This PR adds very simple support for tagging the cloud snapshots when taking a backup. These tags are further inherited by the volumes created from snapshots when restoring the backup. This may be helpful when migrating resources between two different k8s clusters, where the 2nd one expects certain tagging scheme in order to utilize volumes as PVs, especially in case of backing up large releases/number of resources. An example of such use-case may be the Kops clusters on AWS that creates and expects _KubernetesCluster=<name>_ Tag on volumes. Although it does not seem to be a very common case, having support for tagging in place may prove to be useful time-saver when restoring a backup.

What do you think of this? Does it make any sense? If it does but I'm getting ahead of myself by submitting this PR without creating an issue I will change it to the appropriate issue/request.

### DIFF
--- a/docs/aws-config.md
+++ b/docs/aws-config.md
@@ -151,6 +151,31 @@ Specify the following values in the example files:
 
     * Replace `<YOUR_STORAGE_CLASS_NAME>` with `gp2`. This is AWS's default `StorageClass` name.
 
+* (Optional) If you have multiple clusters and you want to support migration of resources between them, in file `examples/aws/10-deployment.yaml`:
+
+    * Uncomment the environment variable `AWS_CLUSTER_NAME` and replace `<YOUR_CLUSTER_NAME>` with the current cluster's name. When restoring backup, it will make Ark (and cluster it's running on) claim ownership of AWS volumes created from snapshots taken on different cluster.
+    The best way to get the current cluster's name is to either check it with used deployment tool or to read it directly from the EC2 instances tags. 
+    
+      The following listing shows how to get the cluster's nodes EC2 Tags. First, get the nodes external IDs (EC2 IDs):
+
+        ```bash
+        kubectl get nodes -o jsonpath='{.items[*].spec.externalID}'
+        ```
+    
+      Copy one of the returned IDs `<ID>` and use it with the `aws` CLI tool to search for one of the following:
+
+      * The `kubernetes.io/cluster/<AWS_CLUSTER_NAME>` tag of the value `owned`. The `<AWS_CLUSTER_NAME>` is then your cluster's name:
+
+          ```bash
+          aws ec2 describe-tags --filters "Name=resource-id,Values=<ID>" "Name=value,Values=owned"
+          ```
+    
+      * If the first output returns nothing, then check for the legacy Tag `KubernetesCluster` of the value `<AWS_CLUSTER_NAME>`:
+
+          ```bash
+          aws ec2 describe-tags --filters "Name=resource-id,Values=<ID>" "Name=key,Values=KubernetesCluster"
+        ```
+
 ## Start the server
 
 In the root of your Ark directory, run:
--- a/examples/aws/10-deployment.yaml
+++ b/examples/aws/10-deployment.yaml
@@ -50,6 +50,8 @@ spec:
               value: /credentials/cloud
             - name: ARK_SCRATCH_DIR
               value: /scratch
+            #- name: AWS_CLUSTER_NAME
+            #  value: <YOUR_CLUSTER_NAME>
       volumes:
         - name: cloud-credentials
           secret:
--- a/pkg/cloudprovider/aws/block_store.go
+++ b/pkg/cloudprovider/aws/block_store.go
@@ -17,7 +17,9 @@ limitations under the License.
 package aws
 
 import (
+	"os"
 	"regexp"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -95,6 +97,8 @@ func (b *blockStore) CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ s
 		return "", errors.Errorf("expected 1 snapshot from DescribeSnapshots for %s, got %v", snapshotID, count)
 	}
 
+	// filter tags through getTagsForCluster() function in order to apply
+	// proper ownership tags to restored volumes
 	req := &ec2.CreateVolumeInput{
 		SnapshotId:       &snapshotID,
 		AvailabilityZone: &volumeAZ,
@@ -102,7 +106,7 @@ func (b *blockStore) CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ s
 		TagSpecifications: []*ec2.TagSpecification{
 			{
 				ResourceType: aws.String(ec2.ResourceTypeVolume),
-				Tags:         snapRes.Snapshots[0].Tags,
+				Tags:         getTagsForCluster(snapRes.Snapshots[0].Tags),
 			},
 		},
 	}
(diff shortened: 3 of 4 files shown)


ncdc:
Hi @ljakimczuk, thanks for your PR. We'll review it soon. We do have an open issue for adjusting the tags to support kops cluster migrations (#451). We have a proposed different solution in there. Please take a look and let us know your thoughts on it - thanks!

ljakimczuk:
Hi @ncdc, sorry for not noticing this issue earlier, the solution proposed there seems much better approach. In case of moving resources between clusters it does not require taking new backup with new tags for the new clusters, it's cleaner.

pkg/cloudprovider/aws/block_store.go
```diff
@@ -180,6 +190,29 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
...
+	// set current cluster ownership tags
+	for k, v := range clusterTags {
+		result = append(result, ec2Tag(k, v))
+	}
+
+	for _, tag := range snapshotTags {
+		if clusterNameTag, _ := regexp.MatchString("kubernetes.io/cluster/*", *tag.Key); clusterNameTag {
+			// if the AWS_CLUSTER_NAME variable is found we want current cluster
+			// to overwrite the old ownership on volumes
+			if _, found := os.LookupEnv("AWS_CLUSTER_NAME"); found {
```
ncdc: Do this check once, outside the loop

ncdc:
@ljakimczuk thanks for changing the approach! Please update the PR title to reflect what it's now doing. @skriss @nrb @carlisia @wwitzel3 PTAL

pkg/cloudprovider/aws/block_store.go
ljakimczuk: Hi, thank you for the review and your suggestion under #451! By doing this check once you mean this:
`	if _, found := os.LookupEnv("AWS_CLUSTER_NAME"); found {
		for _, tag := range snapshotTags {
			if clusterNameTag, _ := regexp.MatchString("kubernetes.io/cluster/*", *tag.Key); clusterNameTag {
				// if the AWS_CLUSTER_NAME variable is found we want current cluster
				// to overwrite the old ownership on volumes
				continue
			}

			result = append(result, ec2Tag(*tag.Key, *tag.Value))
		}
	} else {
		for _, tag := range snapshotTags {
			result = append(result, ec2Tag(*tag.Key, *tag.Value))
		}
	}`

ncdc: More like this:

```go
        haveAWSClusterNameEnvVar := os.GetEnv("AWS_CLUSTER_NAME") != ""

        for _, tag := range snapshotTags {
 		if haveAWSClusterNameEnvVar && strings.HasPrefix(*tag.Key, "kubernetes.io/cluster/") {
 			// if the AWS_CLUSTER_NAME variable is found we want current cluster
 			// to overwrite the old ownership on volumes
 			continue
 		}

 		result = append(result, ec2Tag(*tag.Key, *tag.Value))
 	}
```

ljakimczuk: Shame on me for not figuring this out, thank you very much! :)

docs/aws-config.md
```diff
@@ -151,6 +151,10 @@ Specify the following values in the example files:
 
     * Replace `<YOUR_STORAGE_CLASS_NAME>` with `gp2`. This is AWS's default `StorageClass` name.
 
+* (Optional) If you have multiple clusters and you want to support migration of resources between them, in file `examples/aws/10-deployment.yaml`:
+
+    * Uncomment the environment variable `AWS_CLUSTER_NAME` and replace `<YOUR_CLUSTER_NAME>` with the current cluster's name. When restoring backup, it will make Ark (and cluster it's running on) claim ownership of AWS volumes created from snapshots taken on different cluster.
```
skriss: could you add another sentence and possibly a sample `kubectl` command explaining how to get the current cluster's name?

examples/aws/10-deployment.yaml
```diff
@@ -50,11 +50,13 @@ spec:
               value: /credentials/cloud
             - name: ARK_SCRATCH_DIR
               value: /scratch
+            #-name: AWS_CLUSTER_NAME
```
skriss: add a space between `-` and `name`

pkg/cloudprovider/aws/block_store.go
```diff
@@ -95,14 +97,23 @@ func (b *blockStore) CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ s
 		return "", errors.Errorf("expected 1 snapshot from DescribeSnapshots for %s, got %v", snapshotID, count)
 	}
 
+	tags := make(map[string]string)
```
skriss: any reason to have this section of code outside of the `getTagsForCluster` func? I'd probably opt to move this inside

pkg/cloudprovider/aws/block_store.go
```diff
@@ -180,6 +191,29 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
...
+
+	// set current cluster ownership tags
+	for k, v := range clusterTags {
+		result = append(result, ec2Tag(k, v))
+	}
+
+	_, haveAWSClusterNameEnvVar := os.LookupEnv("AWS_CLUSTER_NAME")
+
+	for _, tag := range snapshotTags {
+		if haveAWSClusterNameEnvVar && strings.HasPrefix(*tag.Key, "kubernetes.io/cluster/") {
```
skriss: if `haveAWSClusterNameEnvVar` is true, we should also skip the `KubernetesCluster` tag

ljakimczuk:
Hi @skriss, thank you for the review! Let me know if the answers satisfies you, if yes then tomorrow/today's evening I will push the changes.

docs/aws-config.md
ncdc: I'm pretty sure the only way to get this information is to go to EC2 and look at the tags applied to the cluster's instances. The kubectl doesn't is not guaranteed to be the right name.

pkg/cloudprovider/aws/block_store.go
```diff
@@ -180,6 +184,29 @@ func (b *blockStore) CreateSnapshot(volumeID, volumeAZ string, tags map[string]s
...
 
+func getTagsForCluster(snapshotTags []*ec2.Tag) []*ec2.Tag {
+	var result []*ec2.Tag
+
+	if clusterName, found := os.LookupEnv("AWS_CLUSTER_NAME"); found {
+		result = append(result, ec2Tag("kubernetes.io/cluster/"+clusterName, "owned"))
+		result = append(result, ec2Tag("KubernetesCluster", clusterName))
+	}
+
+	_, haveAWSClusterNameEnvVar := os.LookupEnv("AWS_CLUSTER_NAME")
```
ncdc: I wouldn't look this up a second time, since you added it above. I'd rename `found` above to `haveAWSClusterNameEnvVar` and remove this line.

docs/aws-config.md
skriss: Ah, ok - it would be really useful to either describe that or include an `aws` CLI cmd to fetch it

skriss:
The approach/code LGTM pending addressing @ncdc's latest comment. Could you also add a unit test for the `getTagsForCluster` function?  Then just adding a description of how to get the cluster name in the docs, and I think we're good.  Thanks again!

docs/aws-config.md
ljakimczuk: I agree that the `kubectl` context is unreliable, even though the chances of getting right name from it are still high I guess if it was configured according to deployment tool instructions. I changed it to short listing of how to get the name-related Tags from the instances using aws CLI

pkg/cloudprovider/aws/block_store.go
ljakimczuk: I moved the lookup above the first if statement and do it only once, then only use the `haveAWSClusterNameEnvVar`, is that ok?

skriss:
Thanks for the updates @ljakimczuk.  This LGTM

skriss:
Let's squash this down to a single commit before merge.

ljakimczuk:
Thank you @skriss, it's squashed!

carlisia:
👍

nicktate:
Hey @skriss , was there a reason this wasn't picked into the `1.9-release` branch?

skriss:
Nope, we probably just missed it - I'll label it as a backport candidate and we should be able to get it into v0.9.8 (release date 10/18) assuming it cherry-picks cleanly.

containershipsupport:
Awesome, thanks for the quick response.

nicktate:
Awesome, thanks for the quick response.

PR #820, opened 2018-09-05
set azure restic env vars based on default backup location's config

skriss:

The location-related changes to Azure to support multiple storage accounts broke restic support on Azure, because storage accounts/keys are no longer static config values.  This PR fixes the restic integration.

Note that this is likely a short-term fix, since we'll be adding multi-location support for restic as well. But in the interim, this keeps it working.


skriss:
@carlisia PTAL when you get a chance. waiting to tear down my AKS cluster until this goes in :) thanks!

PR #821, opened 2018-09-05
Add restic docs to ToC

nrb:
This is currently live at http://nrb.github.io/ark if you'd like to see how it's rendered.

### DIFF
--- a/_data/v9-toc.yml
+++ b/_data/v9-toc.yml
@@ -21,6 +21,8 @@ toc:
         url: /ibm-config
       - page: Configuration reference
         url: /config-definition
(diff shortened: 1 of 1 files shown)


_data/v9-toc.yml
```diff
@@ -21,6 +21,8 @@ toc:
         url: /ibm-config
       - page: Configuration reference
         url: /config-definition
+      - page: Restic Install
+        url: /restic#setup
```
carlisia: Could this both be Install/#install or Setup/#setup?

nrb: Updated to match

PR #824, opened 2018-09-06
convert all controllers to use genericController, logContext -> log

skriss:

Still need to do a little more manual testing

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -19,13 +19,11 @@ package controller
 import (
 	"bytes"
 	"compress/gzip"
-	"context"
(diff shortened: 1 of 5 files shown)


skriss:
alright, ran through tests of all these controllers and everything seems to be working

ncdc:
> I think it would be helpful to have a codified list of possible values for the first arg of newGenericController

I'd rather keep this open. That's in line w/how you create work queues (arbitrary names).

carlisia:
Ah got it.

skriss:
OK, good to go.

nrb:
Will take a pass on this after lunch

PR #825, opened 2018-09-06
sync controller improvements

### LINKED ISSUE #785: Make sync controller more intelligent
Right now, the backup sync controller runs every hour (by default), and each time it runs, it:
- lists all backups in each backup storage location (1 cloud API call per backup location)
- for each backup, gets its metadata (1 cloud API call **per backup**)
- checks the Kube API to see if a backup with that name already exists -- if so, do nothing; if not, create one based on the metadata from the backup location

This works OK but is not the most efficient implementation.  I have a couple of proposals for how to optimize this:
- start putting a `lastmodified` file in the root of each backup location.  Any time a new backup is saved to object storage, this file's modified timestamp will be updated
- when the sync controller runs, for each backup location first check the modified timestamp on the `lastmodified` file to see if it's more recent than the last time that location was synced (we'll store that value on the BackupStorageLocation status).  This takes a single cloud API call.  If it's not, do nothing.
- if the `lastmodified` has a more recent timestamp than our last update, then get a list of all of the **names** of the backups in this backup location (this takes only a single cloud API call, rather than one per backup)
- check each name against the cluster, and only if a backup with that name does not exist in the cluster do we fetch the metadata from the cloud and sync it into the cluster (one cloud API call per backup that needs to be synced, which should be << the total # of backups in the location)

The main benefit of this optimization is that we can afford to run the sync controller much more frequently, since we won't be making a bunch of extraneous cloud API calls each time it runs.  This will help keep the cluster more in sync with the contents of the backup storage locations, and it allows us to remove code in the restore controller that works around the fact that the sync currently runs infrequently.

Would appreciate questions/comments/etc. on this!  I might try to squeeze this into v0.10 if we agree it makes sense, because it will require an object store plugin API change, which we're already making for 0.10.  But if not, it would fit into the "backup ownership and sharing" epic nicely too.

cc @nrb @carlisia @rosskukulinski @wwitzel3

skriss:
Fixes #785 

Here's a first draft of the sync controller improvements, would appreciate input if anyone has time.  Unit tests are broken for now.

Note: the current version uses a UUID as a backup-location "revision".  I think this is the best option for now -- avoids the timestamp problems, but also avoids some complexity with using an integer counter.  With a counter, you always need to *read* the file before you can write an updated value.  This requires more API calls and introduces potential synchronization issues, and also complexities around what to do if there's an error reading (does the file just not exist yet? some other problem?)  With a UUID, we can just always write a new value.  Open to input on this, though.


skriss:
Oh, ignore the first commit, it's from a different PR.

pkg/persistence/object_store.go
```diff
@@ -276,6 +281,30 @@ func (s *objectBackupStore) GetDownloadURL(backup string, target arkv1api.Downlo
 	}
 }
 
+func (s *objectBackupStore) GetRevision() (string, error) {
+	rdr, err := s.objectStore.GetObject(s.bucket, s.prefix+"revision")
```
nrb: Would it make sense to move `s.prefix+"revision"` into a `getRevisionKey` function?

pkg/cmd/server/server.go
```diff
@@ -565,6 +565,7 @@ func (s *server) runControllers(config *api.Config, defaultBackupLocation *api.B
 	}
 
 	backupSyncController := controller.NewBackupSyncController(
+		s.arkClient.ArkV1(),
```
nrb: Is there a reason we shouldn't use the same object in memory for both of these?

skriss: Eh, this is how we do it everywhere, though I find this pattern annoying.  I'd rather switch controllers to either (a) just take an `ArkV1Interface` and let them access whichever type-specific getters they need, or (b) take namespace-specific clients (e.g. `s.arkClient.ArkV1().Backups(s.namespace)`)

pkg/persistence/object_store.go
skriss: yeah, makes sense

pkg/cmd/server/server.go
nrb: Yeah, I think if we're dealing with multiple Ark resources in a single place, these would make sense. My initial suggestion was going to be (a).

skriss: Pushed a commit showing what (b) would look like

skriss: @nrb if it's OK with you I'd rather tackle this problem as part of a different PR, and use the existing pattern here for now.

pkg/controller/backup_sync_controller.go
```diff
@@ -34,24 +36,24 @@ import (
...
 	*genericController
 
-	client                      arkv1client.BackupsGetter
-	backupLister                listers.BackupLister
-	backupStorageLocationLister listers.BackupStorageLocationLister
-	namespace                   string
+	backupClient                arkv1client.BackupInterface
+	backupLocationClient        arkv1client.BackupStorageLocationInterface
+	backupLister                listers.BackupNamespaceLister
+	backupStorageLocationLister listers.BackupStorageLocationNamespaceLister
```
carlisia: Would it still make sense if we removed `backup` from the name of these vars?

pkg/controller/backup_sync_controller.go
```diff
@@ -103,35 +132,36 @@ func (c *backupSyncController) run() {
...
 
 		backupStore, err := c.newBackupStore(location, pluginManager, log)
 		if err != nil {
 			log.WithError(err).Error("Error getting backup store for location")
 			continue
 		}
 
-		backupsInBackupStore, err := backupStore.ListBackups()
+		syncTime := time.Now().UTC()
+		ok, revision := shouldSync(location, syncTime, backupStore, log)
```
carlisia: Could we not just use `time.Now().UTC()` both inside the `shouldSync` and down below?

pkg/controller/backup_sync_controller.go
```diff
@@ -103,35 +132,36 @@ func (c *backupSyncController) run() {
...
 
-		log.WithField("backupCount", len(backupsInBackupStore)).Info("Got backups from backup store")
-
-		cloudBackupNames := sets.NewString()
-		for _, cloudBackup := range backupsInBackupStore {
-			log = log.WithField("backup", kube.NamespaceAndName(cloudBackup))
-			log.Debug("Checking cloud backup to see if it needs to be synced into the cluster")
+		for backupName := range backupStoreBackups {
+			log = log.WithField("backup", backupName)
+			log.Debug("Checking backup store backup to see if it needs to be synced into the cluster")
```
carlisia: Is this debug msg correct?

pkg/cmd/server/server.go
nrb: Yep, that's fine.

pkg/controller/backup_sync_controller.go
skriss: given that we have clients and listers for multiple API objects here, I'd prefer to explicitly name them

pkg/controller/backup_sync_controller.go
skriss: I think so? which part were you questioning

skriss:
@nrb @carlisia @wwitzel3 this is rebased and ready for a full review

skriss:
rebased. I will squash this before merge.

wwitzel3:
I guess it is because we are using a UUID now, so we don't need the full backup instance just the name?

skriss:
The sync controller is the only code that uses the `ListBackups` function, and it only needs the full `*arkv1api.Backup` if it's actually syncing the backup in.  So returning just the names from `ListBackups` reduces that function to a single cloud-provider API call, and then the sync controller will get the full obj only if it needs to actually sync.

Basically trying to minimize the # of cloud API calls so that it can be run much more frequently without blowing up users' API limits

skriss:
I squashed and added a unit test for the `shouldSync` func

skriss:
@carlisia you are right I mis-titled one of my test cases :) updated.

skriss:
ready for merge if you're happy with it.

carlisia:
Will merge after CI comes back.

PR #826, opened 2018-09-07
Wait for namespace to terminate before restoring

### LINKED ISSUE #691: Wait for namespace to finish terminating on restore.
If the user needs to restore an entire namespace, they will first need to delete the namespace. It would be helpful if ark would wait for the namespace to finish terminating and then execute the restore.  It would also be nice if ark would inform the user that it is waiting for the termination to complete so that the user has some idea about the time that elapses before the restore is complete.  

Does there need to be a hard cap on how long ark will wait before dropping out and notify the user that there is some issue?

Depending on the user-defined settings for a particular pod, deleting a namespace can take a considerable amount of time. Parameters such as "terminationGracePeriodSeconds: <value>" can have a significant impact on the termination time. Should we remind the user that certain conditions will prolong the restore time? 

**Environment:**

- Ark version (use `ark version`): v0.9.0
- Kubernetes version (use `kubectl version`):v1.10.3
- Kubernetes installer & version: heptio quickstart

nrb:
Fixes #691


pkg/restore/restore.go
```diff
@@ -459,11 +464,34 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+						addArkError(&errs, err)
+						continue
+					}
+					// TODO: maybe make these their own things vs using the resources
+					ctx.resourceWatches = append(ctx.resourceWatches, namespaceWatch)
+					ctx.resourceWaitGroup.Add(1)
+					go func() {
+						defer ctx.resourceWaitGroup.Done()
+
+						if _, err := waitForNamespaceDelete(namespaceWatch.ResultChan(), mappedNsName, ctx.namespaceTimeout, ctx.logger); err != nil {
```
nrb: Is this pattern what I want to be using here? I see there's also the informer event handlers, as seen at https://github.com/heptio/ark/blob/master/pkg/restic/restorer.go#L64. Is that only for the additions, updates, and deletions from the cache?

ncdc: @nrb it's better to use the informer. If you create your own watch, it can terminate for a variety of reasons, and you might miss an event. While you're not guaranteed to see 100% of events with the informers, they do handle watch terminations and automatically reestablish new watches for you, and you'll get the latest state of the object (or a "deleted final state unknown" meaning it was deleted but it missed the deletion event).

nrb: Ok, thanks - I'll rework this to use an informer DeleteFunc.

pkg/cmd/server/server.go
```diff
@@ -669,12 +674,22 @@ func (s *server) runControllers(config *api.Config, defaultBackupLocation *api.B
...
 		s.logger,
 	)
 	cmd.CheckError(err)
 
+	// Wait for cache sync here so that the event handlers added in NewKubernetesRestorer are ready by the time
+	// the restore controller is started
+	// TODO(nrb): We never hit this debug statement if cache.WaitForCacheSync is uncommented.
+	// Removing that call allows the server to start.
+	s.logger.Debugln("Waiting for kubesharedinformer to sync")
+	cache.WaitForCacheSync(ctx.Done(), s.kubeSharedInformerFactory.Core().V1().Namespaces().Informer().HasSynced)
```
nrb: See the comment here - this appears to deadlock the server at startup.

pkg/restore/restore.go
```diff
@@ -696,6 +797,8 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 				go func() {
 					defer ctx.resourceWaitGroup.Done()
 
+					// TODO(nrb): This is currently conflicting with the namespace termination waits. If the namespace becomes
+					// ready to restore, everything in the NS seems to be restored but the PV never gets created and the PVC is "lost"
```
nrb: Thoughts on handling this? I'm going to keep testing and see if a new PV is made then deleted during the termination wait, but that may be a dead end.

skriss: I don't really see why this would affect PVs on first glance - let me see if I can test it and figure out what's going on

pkg/cmd/server/server.go
nrb: Thanks - I realized this later :)

pkg/restore/restore.go
nrb: I believe it's cause of the PV timeout being 1 minute, where I was using 2-5 minute namespace timeouts. I believe the PV restore attempt would timeout before the namespace termination did, and thus abandon doing a restore. I did try using the same `ctx.namespaceTimeout` value for both, but I didn't see any improvement.

I *think* upping the timeout on PV restoration would cause an issue with mapping the PVC, but I'm not positive.

skriss: OK, here's what I think is happening:

1. namespace `foo` has a PVC claiming PV `bar`
2. namespace `foo` is deleted and moves to `Terminating` state
3. user executes `ark restore create` which starts waiting on `foo` to be deleted
4. `foo` finishes deleting, at which point the PVC claiming PV `bar` is gone, and PV `bar` is released
5. Ark immediately restores resources in NS `foo`, then moves on to trying to restore the PV
6. The dynamic provisioner that deletes released PVs hasn't yet deleted the existing PV
7. Ark sees that it's trying to restore a resource that already exists, and can't, so it moves on
8. The dynamic provisioner deletes the PV
9. End result: Ark restores the PVC in namespace `foo`, but the PV never gets restored

skriss: We'd need to also wait on PVs in the `Released` state with a reclaim policy of `Delete` to disappear before trying to restore them if we want to handle this correctly

nrb: So you're saying this would be a change to the `waitForReady` function? Such as how we determine the value of `isPVReady`?

skriss: I'm saying we'd need a `pvWaiter` that does basically the same thing as the `namespaceWaiter`, but for pvs

nrb: Thanks!

nrb:
In manual testing, this appears to work in that it waits for both persistent volumes and namespaces to be fully deleted. 

To simulate deletion and timeouts during testing, I added a fake finalizer to pods (`ark.heptio.com/fake`), deleted the namespace, started a restore, then removed the finalizer via `kubectl edit` or kept it there to see if the timeout was reached.

I'm going to continue working on the tests since a few things did change that need to be added.

nrb:
Tests are updated. I'm not explicitly testing the `DeleteFunc` handlers on the informers. Do we want to do that? It looks like it'll be a lot more scaffolding to get them set up, somewhat similar to the `backup_controller_test.go` tests that interact directly with the Informer store.

nrb:
@wwitzel3 @skriss I think this is ready for a more thorough review.

pkg/cmd/server/server.go
```diff
@@ -706,12 +711,24 @@ func (s *server) runControllers(config *api.Config, defaultBackupLocation *api.B
...
+		s.kubeClient.CoreV1().PersistentVolumes(),
+		s.kubeSharedInformerFactory.Core().V1().PersistentVolumes(),
 		s.resticManager,
 		s.config.podVolumeOperationTimeout,
+		s.config.namespaceTimeout,
 		s.logger,
 	)
 	cmd.CheckError(err)
 
+	// Start the informat and wait for cache sync here so that the event handlers added in NewKubernetesRestorer are ready by the time
```
ncdc: typo: informat

pkg/restore/restore.go
```diff
@@ -149,23 +155,80 @@ func NewKubernetesRestorer(
...
+		nsWaiter:              &namespaceWaiter{chans: make(map[string]chan *v1.Namespace)},
+		pvWaiter:              &pvWaiter{chans: make(map[string]chan *v1.PersistentVolume)},
 		resourcePriorities:    resourcePriorities,
 		logger:                logger,
 
 		fileSystem: filesystem.NewFileSystem(),
-	}, nil
+	}
+
+	namespaceInformer.Informer().AddEventHandler(
```
nrb: Upon some reflection, I'd like to have test coverage for these two event handlers, but the two solutions I came up with to do so seemed kind of heavy weight considering what they do.

* Mock the namespace and PV informers, such that the tests will add/remove/change objects so the handlers will be triggered. This is more like what our code currently does, but is likely much more scaffolding.
* Instead of doing anonymous functions with closures over `kr`, make a factory function that will wrap `kr` and return a handler function. This means all we'd need to create to get handlers to test would be a `KubernetesRestorer`, and the functions can be called directly with dummy objects. This is a little further removed from the actual code, though.

skriss:
@nrb sorry, I didn't make any progress on this while you were out.

nrb:
np - I'll get the tests more fleshed out this week. Would appreciate a review, too.

pkg/cmd/server/server.go
```diff
@@ -78,10 +79,10 @@ const (
 )
 
 type serverConfig struct {
-	pluginDir, metricsAddress, defaultBackupLocation string
-	backupSyncPeriod, podVolumeOperationTimeout      time.Duration
-	restoreResourcePriorities                        []string
-	restoreOnly                                      bool
+	pluginDir, metricsAddress, defaultBackupLocation              string
+	backupSyncPeriod, podVolumeOperationTimeout, namespaceTimeout time.Duration
```
skriss: since we're using this value for PV waiting as well, maybe we want to call it something more generic, like `terminatingResourceTimeout` or sthg along those lines

pkg/cmd/server/server.go
```diff
@@ -144,6 +146,7 @@ func NewCommand() *cobra.Command {
 	command.Flags().BoolVar(&config.restoreOnly, "restore-only", config.restoreOnly, "run in a mode where only restores are allowed; backups, schedules, and garbage-collection are all disabled")
 	command.Flags().StringSliceVar(&config.restoreResourcePriorities, "restore-resource-priorities", config.restoreResourcePriorities, "desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources")
 	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "name of the default backup storage location")
+	command.Flags().DurationVar(&config.namespaceTimeout, "namespace-timeout", config.namespaceTimeout, "duration to wait on namespace termination before failing a restore")
```
skriss: genericize flag name and text since being used for PVs as well

pkg/cmd/server/server.go
```diff
@@ -706,12 +711,24 @@ func (s *server) runControllers(config *api.Config, defaultBackupLocation *api.B
...
 		s.logger,
 	)
 	cmd.CheckError(err)
 
+	// Start the informer and wait for cache sync here so that the event handlers added in NewKubernetesRestorer are ready by the time
+	// the restore controller is started
+	s.logger.Debugln("Starting kube shared informer factory")
+	go s.kubeSharedInformerFactory.Start(ctx.Done())
+	s.logger.Debugln("Waiting for kubesharedinformer to sync")
+	cache.WaitForCacheSync(ctx.Done(), s.kubeSharedInformerFactory.Core().V1().Namespaces().Informer().HasSynced)
```
skriss: check return val here

pkg/restore/restore.go
```diff
@@ -149,23 +155,80 @@ func NewKubernetesRestorer(
...
+				if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+					obj = tombstone.Obj
+				}
+
+				pv, ok := obj.(*v1.PersistentVolume)
+				if !ok {
+					return
+				}
+
+				updateChan, ok := kr.pvWaiter.getDeleteChan(pv.Name)
```
skriss: s/updateChan/deleteChan

pkg/restore/restore.go
```diff
@@ -308,6 +375,97 @@ func resolveActions(actions []ItemAction, helper discovery.Helper) ([]resolvedAc
...
+	case <-time.After(timeout):
+		return nil, errors.Errorf("Namespace %s did not terminate before the alloted timeout.", ns)
+	case res := <-deleteChan:
+		return res, nil
+	}
+}
+
+// pvWaiter encapsulates waiting for a PV with a reclaim policy of 'Delete' to exit the 'Released' state.
+// This is to avoid the condition where a PV associated with a PV is released but not yet deleted not getting restored.
+type pvWaiter struct {
```
skriss: I think we could probably eliminate the need for a second PV-specific waiter type.  Within the waiter struct, we really only need the ObjectMeta; it doesn't really matter what specific type it is.  I'd probably use the `metav1.Object` interface - so if you change the channel type to use that, then I think you can consolidate to a single `resourceWaiter` struct.

pkg/restore/restore.go
skriss: Can remove this comment now

pkg/restore/restore.go
```diff
@@ -455,11 +619,36 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+					if err != nil && apierrors.IsNotFound(err) {
+						addArkError(&errs, err)
+						continue
+					}
+
+					if isNSTerminating(tempNs) {
+						logger.Infof("Namespace still terminating, waiting")
+
+						ctx.nsWaiter.addDeleteChan(mappedNsName)
+						_, err = ctx.nsWaiter.waitForDelete(mappedNsName, ctx.namespaceTimeout)
```
skriss: I think we probably need to check again here to see if it's been terminated, since it's possible it terminated between our first check and adding the delete channel for it.  Only if it still hasn't been terminated to we then need to wait for delete

pkg/restore/restore.go
```diff
@@ -664,6 +853,22 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+			// See if the PV already exists and we need to wait for it delete
+			fromCluster, err := ctx.pvClient.Get(name, metav1.GetOptions{})
+			if err != nil && apierrors.IsNotFound(err) {
+				addArkError(&errs, fmt.Errorf("error getting copy of persistent volume %q from cluster: %v", name, err))
+			}
+
+			if isPVReleasedOrBound(fromCluster) {
+				ctx.log.Infof("Persistent volume %q still in use, waiting for deletion", name)
+				ctx.pvWaiter.addDeleteChan(name)
+				_, err := ctx.pvWaiter.waitForDelete(name, ctx.namespaceTimeout)
```
skriss: probably need to do a similar double-check on whether the PV's been deleted here, since it could've been deleted between the first check and adding the channel

pkg/restore/restore.go
nrb: Ok, that makes sense.

pkg/cmd/server/server.go
nrb: I notice we don't check this elsewhere in the server - should I add that? Is it a total failure mode if it returns `false`?

pkg/restore/restore.go
```diff
@@ -455,11 +619,36 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
 			if !existingNamespaces.Has(mappedNsName) {
 				logger := ctx.log.WithField("namespace", nsName)
 				ns := getNamespace(logger, filepath.Join(dir, api.ResourcesDir, "namespaces", api.ClusterScopedDir, nsName+".json"), mappedNsName)
-				if _, err := kube.EnsureNamespaceExists(ns, ctx.namespaceClient); err != nil {
+				nsIsReady, err := kube.EnsureNamespaceExistsAndIsReady(ns, ctx.namespaceClient)
```
ncdc: I'm wondering if the logic here (pseudocode) should be:

1. Try to get the namespace
1. If it exists and is not terminating, add to existingNamespaces and move on to the next step
1. Otherwise
    1. If it doesn't exist, create it
    1. If it does exist and is terminating, wait for it to no longer exist, up to a timeout. The easiest initial implementation would be to poll with a real client once per second. Given that we aren't doing restores in goroutines and instead are waiting synchronously, I think this logic would be an acceptable tradeoff vs the shared informer/events approach, which makes things much more confusing. WDYT?

nrb: I'll give that a try - I assume we'd have to do the same thing for PVs, then? I ran in to a race condition when I only looked at the NS. (See https://github.com/heptio/ark/pull/826#discussion_r220338830)

ncdc: Yes, we'll have to wait for anything that's in a terminal state to exit that state (which presumably means "be deleted") before we can proceed restoring that item.

nrb: @ncdc I pushed an update to this using clients and polling. Mind looking over it? I think I've covered the big use cases, though I'm not sure if we should fail the restore if we get some error besides not found when doing our checks each tick.

ncdc: 👀

pkg/cmd/server/server.go
```diff
@@ -152,6 +154,7 @@ func NewCommand() *cobra.Command {
 	command.Flags().StringSliceVar(&config.restoreResourcePriorities, "restore-resource-priorities", config.restoreResourcePriorities, "desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources")
 	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "name of the default backup storage location")
 	command.Flags().Var(&volumeSnapshotLocations, "default-volume-snapshot-locations", "list of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...)")
+	command.Flags().DurationVar(&config.resourceTerminatingTimeout, "resource-timeout", config.resourceTerminatingTimeout, "duration to wait on resource termination before failing a restore")
```
ncdc: I think `--resource-timeout` isn't 100% clear what it's for - we should adjust

pkg/restore/restore.go
ncdc: @nrb I would still like to do what I suggested above in steps 1-3. In other words, let's not use `kube.EnsureNamespaceExistsAndIsReady` or let's modify it to work with these steps. WDYT?

pkg/restore/restore.go
```diff
@@ -474,11 +485,41 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+				if err != nil {
 					addArkError(&errs, err)
 					continue
 				}
 
+				if !nsIsReady {
+					timeout := time.After(ctx.resourceTerminatingTimeout)
+					tick := time.Tick(1 * time.Second)
```
ncdc: Need a deferred stop of `tick` I believe

pkg/restore/restore.go
```diff
@@ -474,11 +485,41 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+					var proceed bool
+					for proceed != true {
+						select {
+						case <-timeout:
+							err := errors.Errorf("Timed out waiting on NS %s to be ready", ns.Name)
```
ncdc: NS->namespace

pkg/restore/restore.go
```diff
@@ -474,11 +485,41 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+							addArkError(&errs, err)
+							return warnings, errs
+						case <-tick:
+							nsIsReady, err = kube.EnsureNamespaceExistsAndIsReady(ns, ctx.namespaceClient)
```
ncdc: I don't want to re-call `kube.EnsureNamespaceExistsAndIsReady` here. Let's implement my suggestion above unless you have objections?

pkg/restore/restore.go
nrb: So I understood EnsureNamespaceExistsAndIsReady was doing 1-2 and 3i; I see I missed putting it in to existingNamespaces. I'll move that logic out and do it all here so it's all present.

ncdc: Nothing is doing 1 yet

pkg/restore/restore.go
nrb: Looks like `tick` is just the channel, not a `Ticker`. I'll redo it so we have a `Ticker` that we can actually stop.

pkg/restore/restore.go
```diff
@@ -696,7 +748,40 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+				addToResult(&errs, namespace, fmt.Errorf("error checking existence for PV %s: %v", name, err))
+			}
+
+			if isPVReleasedOrBound(pv) {
+				ctx.log.Infof("Persistent volume %s still in use, waiting for deletion", name)
+				timeout := time.After(ctx.resourceTerminatingTimeout)
+				ticker := time.NewTicker(time.Second)
+				defer ticker.Stop()
+			loop:
+				for {
```
nrb: @skriss PTAL. I'm not feeling super great about the for/select combo, but maybe it's my unfamiliarity with the Go channels mechanisms.

2 things I'm noticing with this and the namespace loop (L513):

1. Need some way to honor ctrl-c or other graceful shutdown. Should we hook it in to [`signals.CancelOnShutdown`](https://github.com/heptio/ark/blob/master/pkg/cmd/server/server.go#L269) somehow?
2. Do we need to find some way to mark the restore as `Failed` for the NS/PV timeout cases? Or is completed with errors ok? For reference, this is what a restore that hits the timeout looks like currently:

```
x1c in /home/nrb/go/src/github.com/heptio/ark (git) fix-691 U
% ark restore describe voltest-20181109132929
Name:         voltest-20181109132929
Namespace:    heptio-ark
Labels:       <none>
Annotations:  <none>

Backup:  voltest

Namespaces:
  Included:  *
  Excluded:  <none>

Resources:
  Included:        *
  Excluded:        nodes, events, events.events.k8s.io, backups.ark.heptio.com, restores.ark.heptio.com
  Cluster-scoped:  auto

Namespace mappings:  <none>

Label selector:  <none>

Restore PVs:  auto

Phase:  Completed

Validation errors:  <none>

Warnings:
  Ark:        <none>
  Cluster:    <none>
  Namespaces:
    nginx-example:  not restored: persistentvolumeclaims "nginx-logs" already exists and is different from backed up version.
                    not restored: pods "nginx-deployment-f9d69f8fd-c6k6g" already exists and is different from backed up version.
                    not restored: deployments.apps "nginx-deployment" already exists and is different from backed up version.
                    not restored: endpoints "my-nginx" already exists and is different from backed up version.
                    not restored: replicasets.apps "nginx-deployment-f9d69f8fd" already exists and is different from backed up version.
                    not restored: services "my-nginx" already exists and is different from backed up version.

Errors:
  Ark:      Timed out waiting on PV pvc-98f7668e-e390-11e8-9753-42010a96003b to be deleted
  Cluster:    <none>
  Namespaces: <none>
```

ncdc:
As long as we can do some work, I think we said the status is completed (or
processed). Individual item warning or errors cover specific issues. Right?

pkg/restore/restore.go
```diff
@@ -474,14 +485,55 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+						addArkError(&errs, err)
+						continue
+					}
+				}
+
+				// Namespace exists and is active
+				if clusterNS.Status.Phase == v1.NamespaceActive {
+					// keep track of namespaces that we know exist so we don't
+					// have to try to create them multiple times
+					existingNamespaces.Insert(mappedNsName)
```
skriss: i would move the insertion into `existingNamespaces` down to below your `for` loop; it won't need to be within an `if` statement there - we'll always execute it if we get there.  See next comment as well.

pkg/restore/restore.go
```diff
@@ -474,14 +485,55 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+				}
+
+				// Namespace exists but is terminating
+				if clusterNS.Status.Phase == v1.NamespaceTerminating {
```
skriss: this needs to be `if clusterNS != nil && ...` since we'll get here if the cluster didn't have the namespace, and we just created it.

pkg/restore/restore.go
```diff
@@ -474,14 +485,55 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+					timeout := time.After(ctx.resourceTerminatingTimeout)
```
skriss: i'd use a `time.NewTimer` here (for the timeout) and `defer timeout.Stop()` just after

pkg/restore/restore.go
```diff
@@ -474,14 +485,55 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+					ticker := time.NewTicker(time.Second)
+					defer ticker.Stop()
+				loop:
```
skriss: I'd probably move L510-534 into a helper, something like `func createNamespaceAfterTerminate(...) error`, and call it here. If it returns an error, `addArkError` and continue.

pkg/restore/restore.go
```diff
@@ -474,14 +485,55 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+					for {
+						select {
+						// TODO: Need a way to break on process exit here; ctrl-c isn't working when we're in this loop
+						case <-timeout:
```
skriss: `case <-timeout.C:`

pkg/restore/restore.go
```diff
@@ -474,14 +485,55 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+							err := errors.Errorf("Timed out waiting on namespace %s to be ready", mappedNsName)
```
skriss: if you move this block to a helper, then you can just return this error

pkg/restore/restore.go
```diff
@@ -474,14 +485,55 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+							addArkError(&errs, err)
+							return warnings, errs
+						case <-ticker.C:
+							c, err := ctx.namespaceClient.Create(ns)
+							if apierrors.IsAlreadyExists(err) && c.Status.Phase == v1.NamespaceTerminating {
+								logger.Debugf("Namespace %s still terminating, waiting", mappedNsName)
+							} else if err != nil && !apierrors.IsAlreadyExists(err) {
+								addArkError(&warnings, err)
```
skriss: if you move this block to a helper, then you can just return this error

pkg/restore/restore.go
```diff
@@ -474,14 +485,55 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+								break loop
+							} else {
+								logger.Infof("Termination successful, namespace %s restored", mappedNsName)
+								break loop
```
skriss: if you move this block to a helper, then you can return nil here

pkg/restore/restore.go
skriss: For #2, completed with errors is what we'd expect.

Looking at #1.

skriss: For #1, I'm not sure that we actually want to do this - would like to think some more and maybe @ncdc will find time to provide input.  We don't currently have any code that will stop a backup or restore in the middle if a shutdown signal is received - we try to finish the backup/restore, then the controllers/server exit.  We'd want to be able to properly mark the restore's status as failed/interrupted, which would take more code that we don't currently have.

If we were going to implement it, looks like the most straight-forward way to do this is in `server.go`, around L590, pass `ctx` as a new arg to the `restore.NewKubernetesRestorer` func.  Have that func set it as a field on the `kubernetesRestorer`.  In the `Restore` method on the `kubernetesRestorer`, pass that `ctx` down to the `restoreCtx` that's created (probably as a field).  Then, in your `select` statements that are doing the polling, add an additional case that receives on `ctx.Done()` and exits -- though we'd probably want to end up marking restores that were exited early like this as Failed.

nrb: Thanks. That makes sense. My use case was lowering the timeout for testing (default of 10, but wanted to drop to 1 or 2), so doing a `kill -9` on it I guess isn't too terrible.

I can leave that part as is and address the remaining comments and update tests.

pkg/restore/restore.go
nrb: It can be returned, but it's going to warnings; would you suggest all the errors returned by the helper go into `&errs`?

skriss: hmm, i think it makes sense to record this as an error, so yeah

nrb:
@skriss @ncdc Updated with helpers to contain the wait loops. Would it make more sense to use the `resourceClient` and just have 1 helper?

Also, ideas on the name for the flag are welcome. I don't want to spell out `namespace-and-persistent-volume-timeout`, but maybe that's the best plan.

pkg/restore/restore.go
```diff
@@ -696,7 +772,21 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 
 			// Check if the PV exists in the cluster before attempting to create
 			// a volume from the snapshot, in order to avoid orphaned volumes (GH #609)
-			_, err := resourceClient.Get(name, metav1.GetOptions{})
+			pv, err := ctx.pvClient.Get(name, metav1.GetOptions{})
+			if err != nil && !apierrors.IsNotFound(err) {
+				addToResult(&errs, namespace, fmt.Errorf("error checking existence for PV %s: %v", name, err))
+			}
+
+			if isPVReleasedOrBound(pv) {
```
skriss: I don't think `isPVReleasedOrBound` is quite right.  I think we want to wait if the PV has a phase of Released (or Failed, since sometimes they go to this phase and are subsequently deleted) *and* has a reclaim policy of Delete.  Does that sound right?  Is there something I'm overlooking?

pkg/restore/restore.go
```diff
@@ -510,6 +540,52 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+	for {
+		select {
+		case <-timeout.C:
+			return errors.Errorf("Timed out waiting on PV %s to be deleted", pv.Name)
+		case <-ticker.C:
+			pv, err := ctx.pvClient.Get(pv.Name, metav1.GetOptions{})
+			if err != nil && !apierrors.IsNotFound(err) {
+				return err
+			}
+			if isPVReleasedOrBound(pv) {
```
skriss: I think this should be `if err == nil`, right? we should continue waiting in all cases where there's no error getting the PV, i.e. it still exists?

pkg/restore/restore.go
```diff
@@ -510,6 +540,52 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+				ctx.log.Debugf("Persistent volume %s still exists, waiting", pv.Name)
+			} else {
```
skriss: related to prev comment, I don't think we need this `else`

skriss:
Maybe `--terminating-resource-wait-timeout`? It's a little long but not something that will need to be typed regularly.

Also, I'm feeling like maybe we should defer this to v0.10.1.  We're getting down to the wire and I want to make sure we have time to test adequately since this is in the core restore path.

nrb:
> Also, I'm feeling like maybe we should defer this to v0.10.1.

I'm ok with that; I'd like to wrap this up as it's fairly old, but also it's taken a while to get right, and yeah, it's in the critical path.

pkg/restore/restore.go
```diff
@@ -474,14 +485,33 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
 			if !existingNamespaces.Has(mappedNsName) {
 				logger := ctx.log.WithField("namespace", nsName)
 				ns := getNamespace(logger, filepath.Join(dir, api.ResourcesDir, "namespaces", api.ClusterScopedDir, nsName+".json"), mappedNsName)
-				if _, err := kube.EnsureNamespaceExists(ns, ctx.namespaceClient); err != nil {
+
+				clusterNS, err := ctx.namespaceClient.Get(mappedNsName, metav1.GetOptions{})
+				if err != nil && !apierrors.IsNotFound(err) {
```
ncdc: I think this reads cleaner as
```go
  // ...
} else if err != nil {
  // ...
}
```

pkg/restore/restore.go
```diff
@@ -474,14 +485,33 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+				} else if apierrors.IsNotFound(err) {
+					_, err := ctx.namespaceClient.Create(ns)
+					if err != nil {
+						addArkError(&errs, err)
+						continue
+					}
 				}
 
+				// Namespace exists but is terminating
+				if clusterNS != nil && clusterNS.Status.Phase == v1.NamespaceTerminating {
```
ncdc: You shouldn't need to check for `clusterNS != nil` - if you make it here, it means `ctx.namespaceClient.Get()` returned no error.

pkg/restore/restore.go
```diff
@@ -510,6 +540,52 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
 	return warnings, errs
 }
 
+func (ctx *context) createNamespaceAfterTerminate(ns *v1.Namespace, logger *logrus.Entry) error {
+	timeout := time.NewTimer(ctx.resourceTerminatingTimeout)
```
ncdc: You can replace the timer & ticker with apimachinery's `wait.PollImmediate` or `wait.PollImmediateUntil` (if you need to plumb through a stop channel).

pkg/restore/restore.go
nrb: I think that makes sense. Would it make more sense to check if the PV has a reclaim policy of Delete and a phase that's not Terminating?

I included Bound because, depending on when the delete is issued, the PV may not have entered the Released state yet.

pkg/restore/restore.go
```diff
@@ -474,14 +485,33 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
 			if !existingNamespaces.Has(mappedNsName) {
 				logger := ctx.log.WithField("namespace", nsName)
 				ns := getNamespace(logger, filepath.Join(dir, api.ResourcesDir, "namespaces", api.ClusterScopedDir, nsName+".json"), mappedNsName)
-				if _, err := kube.EnsureNamespaceExists(ns, ctx.namespaceClient); err != nil {
+
+				clusterNS, err := ctx.namespaceClient.Get(mappedNsName, metav1.GetOptions{})
```
ncdc: I think this part and the `ctx.createNamespaceAfterTerminate` part can be unified instead of being split:

```go
wait.PollImmediate(time.Second, ctx.resourceTerminatingTimeout, func() (bool, error) {
  clusterNS, err := ctx.namespaceClient.Get(mappedNsName, metav1.GetOptions{})
    if _, err := ctx.namespaceClient.Create(ns); err != nil {
      addArkError(&errs, err)
      return false, err
    }

    // ns created successfully; exit the loop
    return true, nil
  } else if err != nil {
    addArkError(&errs, err)
    return false, err
  }

  if clusterNS.Status.Phase == v1.NamespaceTerminating {
    // keep looping
    return false, nil
  }

  // namespace exists and is not terminating; exit the loop
  return true, nil
})

File: pkg/restore/restore.go [hunk 34]
```diff
@@ -696,7 +772,21 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 
 			// Check if the PV exists in the cluster before attempting to create
 			// a volume from the snapshot, in order to avoid orphaned volumes (GH #609)
-			_, err := resourceClient.Get(name, metav1.GetOptions{})
+			pv, err := ctx.pvClient.Get(name, metav1.GetOptions{})
```
ncdc: Similar to my ns comment, can we unify this? pseudo:

```go
wait.PollImmediate(time.Second, ctx.resourceTerminatingTimeout, func() (bool, error) {
  // try to get PV
  // if not found, executePVAction, return "true, nil" to exit loop
  // else if err != nil, addArkError, return "false, err" to exit loop
  // else
    // i'm not sure we really need to check phase and/or reclaim policy - the goal is for the PV to not exist so we can run executePVAction above ... right?
})
```

nrb:
The current code is resulting in errors like this on a happy path (namespace/PV not found) run:

```
Errors:
  Ark:        <none>
  Cluster:  error restoring /tmp/079419147/resources/persistentvolumes/cluster/pvc-abbf2618-f35b-11e8-825d-42010a96001e.json: persistentvolumes "pvc-abbf2618-f35b-11e8-825d-42010a96001e" is forbidden: error querying GCE PD volume gke-cluster-1-81372a07-pvc-abbf2618-f35b-11e8-825d-42010a96001e: GCE persistent disk not found: diskName="gke-cluster-1-81372a07-pvc-abbf2618-f35b-11e8-825d-42010a96001e" zone="us-east4-b"
  Namespaces: <none>
```

In my Ark logs, I see:

``time="2018-11-28T17:53:00-05:00" level=info msg="starting restore" backup=voltest2 logSource="pkg/controller/restore_controller.go:515" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:22-05:00" level=info msg="Not including resource" backup=voltest2 groupResource=nodes logSource="pkg/restore/restore.go:134" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:22-05:00" level=info msg="Not including resource" backup=voltest2 groupResource=events logSource="pkg/restore/restore.go:134" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:22-05:00" level=info msg="Not including resource" backup=voltest2 groupResource=restores.ark.heptio.com logSource="pkg/restore/restore.go:134" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:22-05:00" level=info msg="Not including resource" backup=voltest2 groupResource=backups.ark.heptio.com logSource="pkg/restore/restore.go:134" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:28-05:00" level=info msg="Starting restore of backup heptio-ark/voltest2" backup=voltest2 logSource="pkg/restore/restore.go:356" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:28-05:00" level=info msg="Restoring cluster level resource 'persistentvolumes' from: /tmp/079419147/resources/persistentvolumes/cluster" backup=voltest2 logSource="pkg/restore/restore.go:678" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:28-05:00" level=info msg="Getting client for /v1, Kind=PersistentVolume" backup=voltest2 logSource="pkg/restore/restore.go:735" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:28-05:00" level=info msg="Restoring  PersistentVolume pvc-abbf2618-f35b-11e8-825d-42010a96001e" backup=voltest2 logSource="pkg/restore/restore.go:558" restore=heptio-ark/voltest2-20181128175259
DEBU[0099] looking for plugin in registry                controller=restore key=heptio-ark/voltest2-20181128175259 kind=ObjectStore logSource="pkg/plugin/manager.go:99" name=gcp
DEBU[0099] found preexisting restartable plugin process  command=./_output/bin/linux/amd64/ark controller=restore key=heptio-ark/voltest2-20181128175259 kind=ObjectStore logSource="pkg/plugin/manager.go:110" name=gcp
time="2018-11-28T17:53:29-05:00" level=info msg="successfully restored persistent volume from snapshot" backup=voltest2 logSource="pkg/restore/restore.go:1114" persistentVolume=pvc-abbf2618-f35b-11e8-825d-42010a96001e providerSnapshotID=gke-cluster-1-81372a07-pvc-98ae112e-8b69-4f59-8ad0-3359afbd34d0 restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:29-05:00" level=info msg="Restoring PersistentVolume: pvc-abbf2618-f35b-11e8-825d-42010a96001e" backup=voltest2 logSource="pkg/restore/restore.go:847" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:29-05:00" level=info msg="error restoring pvc-abbf2618-f35b-11e8-825d-42010a96001e: <nil>" backup=voltest2 logSource="pkg/restore/restore.go:906" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:30-05:00" level=info msg="Termination successful, namespace nginx-example restored" backup=voltest2 logSource="pkg/restore/restore.go:527" namespace=nginx-example restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:30-05:00" level=info msg="Restoring resource 'persistentvolumeclaims' into namespace 'nginx-example' from: /tmp/079419147/resources/persistentvolumeclaims/namespaces/nginx-example" backup=voltest2 logSource="pkg/restore/restore.go:676" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:30-05:00" level=info msg="Getting client for /v1, Kind=PersistentVolumeClaim" backup=voltest2 logSource="pkg/restore/restore.go:735" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:30-05:00" level=info msg="Restoring PersistentVolumeClaim: nginx-logs" backup=voltest2 logSource="pkg/restore/restore.go:847" restore=heptio-ark/voltest2-20181128175259
time="2018-11-28T17:53:30-05:00" level=info msg="Restoring resource 'secrets' into namespace 'nginx-example' from: /tmp/079419147/resources/secrets/namespaces/nginx-example" backup=voltest2 logSource="pkg/restore/restore.go:676" restore=heptio-ark/voltest2-20181128175259
```

I tried adding back the resource watch code to enforce waiting on the PV being ready, but that didn't seem to help. @ncdc and @skriss, any ideas on what to investigate here?

**Comment by nrb** [comment]
The source disk mentioned is indeed gone from my GCP console, but the snapshot (`gke-cluster-1-81372a07-pvc-98ae112e-8b69-4f59-8ad0-3359afbd34d0`) does.

**Comment by nrb** [comment]
Alright, I've tested the current code in the following cases against an actual cluster, using the `nginx-example` namespace as my target. My expectations were met in each case, but it'd be a good idea for someone else to go through this, too.

1. Namespace & associated PV completely missing. *Expected*: PV, namespace, & namespace contents restored.
1. Namespace & associated PV already present. **Expected**: Restore results note already-present objects, and the wait loop is bypassed.
1. With the namespace & associated PV already present, issue `kubectl delete namespace nginx-example --wait=false && ark restore create --from-backup <backupname>`. **Expected**: Since the delete and restore requests are so close together, the PV itself will not be marked for deletion, but will wait for the namespace to delete and the currently in-cluster PV to disappear, then restore all resources.
1. With the namespace & associated PV already present, edit the pod to add the `ark.heptio.com/bogus` finalizer, then issue `kubectl delete namespace nginx-example --wait=false && ark restore create --from-backup <backupname>`. **Expected**: The restore for the PV and namespace should reach the timeout, and skip them since they still exist, due to the finalizer keeping the pod and it's PVC around.
1. With the namespace & associated PV already present, edit the pod to add the `ark.heptio.com/bogus` finalizer, issue `kubectl delete namespace nginx-example --wait=false && ark restore create --from-backup <backupname>`, then edit the pod to remove the finalizer. **Expected**: The restore should wait until shortly after the pod is saved, then proceed successfully, recreating all objects in the namespace, in addition to the PV.

File: pkg/restore/restore.go [hunk 35]
```diff
@@ -510,6 +501,116 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+			updatedObj.DeepCopyInto(obj)
+
+			return true, nil
+		} else if err != nil {
+			addToResult(errs, "", fmt.Errorf("error checking existence for PV %s: %v", name, err))
+			return false, err
+		}
+
+		if clusterPV != nil && clusterPV.GetDeletionTimestamp() == nil {
+			namespace, err := collections.GetString(clusterPV.UnstructuredContent(), "spec.claimRef.namespace")
```
ncdc: after getting the namespace, adjust based on remappings

pkg/restore/restore.go
```diff
@@ -510,6 +501,116 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+	}
+}
+
+// ensurePVReadyToCreate creates a wait.ConditionFunc for controlling a wait.PollImmediate loop.
+// The returned function will break the loop if:
+//		- the target persistent volume exists and does not have a DeletionTimestamp set.
+//		- the target persistent volume is not found
+//		- the target persistent volume has a PVC that's associated with a namespace, and looking up the namespace fails
+// If the target persistent volume is not found, a new persistent disk is created and the object-to-restore updated to reference the new disk. Otherwise, a new persistent disk is not made.
+func (ctx *context) ensurePVReadyToCreate(obj *unstructured.Unstructured, errs *api.RestoreResult, warnings *api.RestoreResult, resourceClient client.Dynamic, fullPath string) wait.ConditionFunc {
```
ncdc: Remove `errs` (check return value of `wait.PollImmediate` and handle it there)

pkg/restore/restore.go
```diff
@@ -510,6 +501,116 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+		clusterPV, err := resourceClient.Get(name, metav1.GetOptions{})
+		// Check if the PV exists in the cluster before attempting to create
+		// a volume from the snapshot, in order to avoid orphaned volumes (GH #609)
+		// PV's existence will be recorded later. Just skip the volume restore logic.
+		if apierrors.IsNotFound(err) {
+			ctx.log.Infof("Restoring  PersistentVolume %s", name)
+			// restore the PV from snapshot (if applicable)
+			updatedObj, err := ctx.pvRestorer.executePVAction(obj)
+			if err != nil {
+				addToResult(errs, "", fmt.Errorf("error executing PVAction for %s: %v", fullPath, err))
```
ncdc: Remove

pkg/restore/restore.go
```diff
@@ -510,6 +501,116 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+				return false, err
```
ncdc: `errors.WithMessage` if you want to preserve the message from the line above

pkg/restore/restore.go
```diff
@@ -510,6 +501,116 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+			}
+
+			// Copy the new disks's information into the PV we're trying to restore
+			updatedObj.DeepCopyInto(obj)
+
+			return true, nil
+		} else if err != nil {
+			addToResult(errs, "", fmt.Errorf("error checking existence for PV %s: %v", name, err))
```
ncdc: Remove

pkg/restore/restore.go
```diff
@@ -510,6 +501,116 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+			return false, err
```
ncdc: `errors.WithMessage` if you want to preserve the message from the line above

pkg/restore/restore.go
```diff
@@ -510,6 +501,116 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+		}
+
+		if clusterPV != nil && clusterPV.GetDeletionTimestamp() == nil {
+			namespace, err := collections.GetString(clusterPV.UnstructuredContent(), "spec.claimRef.namespace")
+
+			// The PV is likely missing a claimRef, Assume that this means we're waiting for a namespace to delete.
+			if err != nil {
+				clusterPV.DeepCopyInto(obj)
+				newErr := errors.Wrap(err, "Error looking up PV claimRef")
+				addToResult(errs, "", newErr)
```
ncdc: Remove

pkg/restore/restore.go
```diff
@@ -510,6 +501,116 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+		} else if err != nil {
+			addToResult(errs, "", fmt.Errorf("error checking existence for PV %s: %v", name, err))
+			return false, err
+		}
+
+		if clusterPV != nil && clusterPV.GetDeletionTimestamp() == nil {
+			namespace, err := collections.GetString(clusterPV.UnstructuredContent(), "spec.claimRef.namespace")
+
+			// The PV is likely missing a claimRef, Assume that this means we're waiting for a namespace to delete.
+			if err != nil {
```
ncdc: I'd rather be certain as to why we're getting an error - I'd suggest either using `collections.Exists` first, or converting to a `v1.PersistentVolume`

pkg/restore/restore.go
```diff
@@ -510,6 +501,116 @@ func (ctx *context) restoreFromDir(dir string) (api.RestoreResult, api.RestoreRe
...
+				clusterPV.DeepCopyInto(obj)
+				newErr := errors.Wrap(err, "Error looking up PV claimRef")
+				addToResult(errs, "", newErr)
+				ctx.log.Debugf("%v", newErr)
+				return false, nil
+			}
+
+			ns, err := ctx.namespaceClient.Get(namespace, metav1.GetOptions{})
```
ncdc: As per Zoom chat, replace this part about the namespace with logic that checks the PVC (if 404, proceed; if terminating, wait; otherwise, I guess return an error that we can't restore the PV because both it and its bound PVC still exist (?))

pkg/restore/restore.go
ncdc: Naming suggestion: `pvReadyFunc`

ncdc:
Summarizing what we talked about:

Resources may be in various states at various times (present, terminating, absent), and it's impossible for Ark to successfully cover all scenarios.

The "easiest" case is a restore where none of the objects Ark is restoring exist in the cluster. We don't need any fancy logic here.

Another case is where the user has issued `kubectl delete` for some amount of resources - this could be an entire namespace and everything in it, a group of resources (either by type+name or by label selector), or an individual resource; and then after issuing the delete, they start an Ark restore. With recent changes to kubectl, it now waits until the items are deleted prior to returning. But with previous versions, or if using the api directly, it's hypothetically possible for resources to be terminating while Ark is restoring. In this instance, we'll code Ark to do the best it can. We'll adjust the logic a bit so that it always tries to restore namespaces first (including handling remappings), as long as `namespaces` is an included/not excluded resource. When we see a PV, we'll check its status (404 vs terminating vs present), then check its associated PVC if its state is present. If the PVC is present and not being deleted, we won't restore the PV.

If the user issues a delete after starting an Ark restore, I think all bets are off and the behavior is unspecified.

nrb:
Pairing with @ncdc today, we moved the polling deletion check to be more generic and apply to all resources. We started looking at namespaces, and we've decided to try to treat namespaces more generically and restore them through the same code path we do for everything else.

This is likely going to result in splitting up `restoreResource` (which restores all items of a given type, like pods or configmaps) into 2 functions: one that does all the items of that type, and then one that does the work for a single item. I'll investigate this path more and push what I come up with.

nrb:
Pushed up my current work - I have not tested this yet, and I know namespace remappings aren't handled, but I wanted to see if I could get something compiling for handling namespaces up front as well as "just in time" with the `restoreItem` function.

pkg/restore/restore.go
```diff
@@ -578,6 +590,102 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+			// namespace is in the process of deleting, keep looping
+			return false, nil
+		}
+
+		// None of the PV, PVC, or NS are marked for deletion, break the loop.
+		pvLogger.Debug("PV, associated PVC and namespace are not marked for deletion")
+		return true, nil
+	})
+
+	end := time.Now()
```
ncdc: Check the `err` that `wait.PollImmediate` returns instead of comparing start/end times.

pkg/restore/restore.go
```diff
@@ -578,6 +590,102 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+
+	if !end.Before(timeout) {
```
ncdc: remove

pkg/util/kube/utils.go
```diff
@@ -35,18 +38,51 @@ func NamespaceAndName(objMeta metav1.Object) string {
...
-func EnsureNamespaceExists(namespace *corev1api.Namespace, client corev1client.NamespaceInterface) (bool, error) {
-	if _, err := client.Create(namespace); err == nil {
+// namespace already exists and is not ready, this function will return (false, nil).
+func EnsureNamespaceExistsAndIsReady(namespace *corev1api.Namespace, client corev1client.NamespaceInterface, timeout time.Duration) (bool, error) {
+	var shouldCreate bool
+	err := wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		clusterNS, err := client.Get(namespace.Name, metav1.GetOptions{})
+
+		if apierrors.IsNotFound(err) {
+			// Namespace isn't in cluster, we're good to restore.
```
ncdc:
```suggestion
			// Namespace isn't in cluster, we're good to create.
```

pkg/util/kube/utils.go
```diff
@@ -35,18 +38,51 @@ func NamespaceAndName(objMeta metav1.Object) string {
...
+			return false, nil
+		}
+
+		// clusterNS found, is not nil, and not marked for deletion, therefore we shouldn't create it.
 		return true, nil
-	} else if apierrors.IsAlreadyExists(err) {
+	})
+
+	if err != nil {
+		return false, errors.Wrapf(err, "err getting namespace %s", namespace.Name)
```
ncdc:
```suggestion
		return false, errors.Wrapf(err, "error getting namespace %s", namespace.Name)
```

pkg/util/kube/utils.go
```diff
@@ -35,18 +38,51 @@ func NamespaceAndName(objMeta metav1.Object) string {
...
+	}
+
+	// shouldCreate is true *only* if we got a 404 during polling, and we should only be here if we didn't get an error, which means the timeout was hit
```
ncdc: This is confusing - could you please reword?

pkg/util/kube/utils.go
```diff
@@ -35,18 +38,51 @@ func NamespaceAndName(objMeta metav1.Object) string {
...
 		return false, nil
-	} else {
-		return false, errors.Wrapf(err, "error creating namespace %s", namespace.Name)
 	}
+
+	if _, err := client.Create(namespace); err == nil {
+		return true, nil
+	}
+
+	// if we get an AlreadyExists here, something else created the NS after all of our polling...
```
ncdc:
```suggestion
	// if we get any error here, something else created the NS after all of our polling...
```

pkg/restore/restore.go
nrb: I think I can just return the `err` that `wait.PollImmediate` uses, unless I want to add a debug line for timeout. It will be added to the result at L810 in this version of the PR.

pkg/util/kube/utils.go
```diff
@@ -35,18 +38,52 @@ func NamespaceAndName(objMeta metav1.Object) string {
...
+	if err != nil {
+		return false, errors.Wrapf(err, "error getting namespace %s", namespace.Name)
+	}
+
+	// In the case the namespace already exists and isn't marked for deletion, assume it's ready for use.
+	if ready {
+		return true, nil
+	}
+
 	if _, err := client.Create(namespace); err == nil {
```
skriss: let's switch this to handle the non-nil error case first

pkg/restore/restore.go
```diff
@@ -578,6 +590,97 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+
+		if apierrors.IsNotFound(err) {
+			pvLogger.Debug("PV not found, safe to restore")
+			// PV not found, can safely exit loop and proceed with restore.
+			shouldRestore = true
+			return true, nil
+		}
+
+		if err != nil {
+			return true, errors.Wrapf(err, "could not retrieve in-cluster copy of PV %s", name)
```
skriss: doesn't really matter functionally, since the `PollImmediate` will break if it gets either a `true` or an error, but I'd probably return `false, errors.Wrapf(...)` here and in the subsequent error cases

pkg/restore/restore.go
```diff
@@ -578,6 +590,97 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+
+		}
+
+		var namespace, pvcName string
+		if clusterPV != nil {
```
skriss: `clusterPV` should always be non-nil at this point, right? since we already returned if we got a not-found error. so can probably remove this `if`.  if you want to be safe, you could check that it's not nil and just return an error if it is, since it should never be the case.

pkg/restore/restore.go
nrb: What's the reasoning for using `false`? My reasoning for returning `true` was to be consistent in the cases where the loop was being broken.

skriss: Semantically, the first return argument indicates if the condition that's being checked for has been met, where the condition here is something like "we've determined whether the PV should be restored". IMHO, in an error case, we're breaking not because we've determined that, but because we've hit an error.

`false` makes more sense to me but since functionally it's the same either way, if you prefer it as-is then I'm fine leaving it.

nrb: Ok, that explanation makes sense to me. I'll switch it.

pkg/restore/restore.go
nrb: Removed this in the latest commit.

pkg/util/kube/utils.go
```diff
@@ -35,18 +38,53 @@ func NamespaceAndName(objMeta metav1.Object) string {
...
+		return false, errors.Wrapf(err, "error getting namespace %s", namespace.Name)
+	}
+
+	// In the case the namespace already exists and isn't marked for deletion, assume it's ready for use.
+	if ready {
+		return true, nil
+	}
+
+	if _, err := client.Create(namespace); err != nil {
+		// if we get any error here, something else created the NS after all of our polling
```
ncdc: Do you want to check for an already-exists error?

pkg/restore/restore.go
```diff
@@ -578,6 +590,94 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+
+		if err != nil {
+			return false, errors.Wrapf(err, "could not retrieve in-cluster copy of PV %s", name)
+
+		}
+
+		phase, err := collections.GetString(clusterPV.UnstructuredContent(), "status.phase")
+		if err != nil {
+			// Break the loop since we couldn't read the phase
+			return true, errors.Wrapf(err, "error getting phase for in-cluster PV %s", name)
```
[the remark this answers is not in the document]
nrb: To my mind, `true` means we're breaking the loop. I realize it's not being used, but it provides consistency in the cases where we break.

That said, I'm not tied to it, so if it's clearer for others to use `false`, I'm fine with it.

pkg/restore/restore.go
```diff
@@ -30,9 +30,11 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
nrb: This appears to be coming from https://github.com/golang/go/issues/28428. Openshift-Azure ran into it too, at https://github.com/openshift/openshift-azure/pull/862, but resolved by pinning the commit of goimports at https://github.com/openshift/release/pull/2335

pkg/restore/restore.go
```diff
@@ -578,6 +590,94 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+		if err != nil {
+			return false, errors.Wrapf(err, "error looking up namespace name for in-cluster PV %s", name)
+		}
+		pvcName, err := collections.GetString(clusterPV.UnstructuredContent(), "spec.claimRef.name")
+		if err != nil {
+			return false, errors.Wrapf(err, "error looking up persistentvolumeclaim for in-cluster PV %s", name)
+		}
+
+		// Have to create the PVC client here because we don't know what namespace we're using til we get to this point.
+		pvcClient := ctx.coreClient.PersistentVolumeClaims(namespace)
```
nrb: Testing this right now would require either generating my own fake CoreV1 and PersistentVolumeClaims clients, or bringing in all of client-go's fakes. I don't really want to do either of those, though writing our own seems lighter weight.

Would using `ctx.DynamicFactory` work in it's place? This was originally changed to a typed client since we know exactly what we want, but maybe a dynamic client would be more practical from the testing perspective.

nrb: Correct, just the deletion timestamp there. Deletion timestamp and phase for namespaces.

ncdc: I think the dynamic client is probably fine for those.

pkg/util/kube/utils.go
nrb: So I think we can, but does that mean we should check the phase and deletion timestamp again? Do we care if something else made the namespace before we got to?

ncdc: No, I don't think so. The reason I suggested it was to handle
1. we decide we need to create the namespace (it doesn't exist, or it did and we waited for it to be deleted)
1. something else creates the namespace
1. we try to create the namespace, and it fails because it already exists

If this is the sequence, I don't think we should return an error.

If this is the sequence, *and* the namespace somehow is terminating, I think it's ok to return an error. I don't expect this to happen, and I don't want to get into an infinite loop 😄

pkg/util/kube/utils_test.go
```diff
@@ -18,12 +18,105 @@ package kube
...
 	"testing"
+	"time"
+
+	"k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	arktest "github.com/heptio/ark/pkg/util/test"
+	"github.com/stretchr/testify/assert"
```
skriss: looks like this line needs to move up to the prev section

pkg/restore/restore.go
```diff
@@ -578,6 +584,99 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
 	}
 }
 
+func (ctx *context) waitForPVDeletion(name string, pvClient client.Dynamic) (bool, error) {
+	pvLogger := ctx.log.WithField("pvName", name)
+
+	var shouldRestore bool
+	err := wait.PollImmediate(time.Second, ctx.resourceTerminatingTimeout, func() (bool, error) {
+		clusterPV, err := pvClient.Get(name, metav1.GetOptions{})
+
```
skriss: style nit: rm blank line

pkg/restore/restore.go
```diff
@@ -578,6 +584,99 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+		if apierrors.IsNotFound(err) {
+			pvLogger.Debug("PV not found, safe to restore")
+			// PV not found, can safely exit loop and proceed with restore.
+			shouldRestore = true
+			return true, nil
+		}
+
```
skriss: style nit: rm blank line

pkg/restore/restore.go
```diff
@@ -578,6 +584,99 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+		if err != nil {
+			return false, errors.Wrapf(err, "could not retrieve in-cluster copy of PV %s", name)
+
```
skriss: style nit: rm blank line

pkg/restore/restore.go
```diff
@@ -578,6 +584,99 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+			return false, errors.Wrapf(err, "error getting phase for in-cluster PV %s", name)
+		}
+
+		if phase == string(v1.VolumeReleased) || clusterPV.GetDeletionTimestamp() != nil {
+			// PV was found and marked for deletion, or it was released; wait for it to go away.
+			pvLogger.Debugf("PV found, but marked for deletion, waiting")
+			return false, nil
+		}
+
+		namespace, err := collections.GetString(clusterPV.UnstructuredContent(), "spec.claimRef.namespace")
```
skriss: pls add a comment explaining why we're getting the NS and querying the PVC at this point

pkg/restore/restore.go
```diff
@@ -578,6 +584,99 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+			return false, errors.Wrapf(err, "error getting claim %s for persistent volume", pvcName)
+		}
+
+		if pvc != nil && pvc.GetDeletionTimestamp() != nil {
+			pvLogger.Debugf("PVC for PV marked for deletion, waiting")
+			// PVC is still deleting, continue to wait.
+			return false, nil
+		}
+
+		ns, err := ctx.namespaceClient.Get(namespace, metav1.GetOptions{})
```
skriss: probably worth another comment here explaining why we're now querying the NS

pkg/restore/restore.go
```diff
@@ -578,6 +584,99 @@ func addToResult(r *api.RestoreResult, ns string, e error) {
...
+			return false, errors.Wrapf(err, "error getting pvc client")
+		}
+
+		pvc, err := pvcClient.Get(pvcName, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			pvLogger.Debugf("PVC %s for PV not found, waiting", pvcName)
+			// PVC wasn't found, but the PV still exists, so continue to wait.
+			return false, nil
+		}
+
```
skriss: style nit: rm blank line

pkg/util/kube/utils.go
```diff
@@ -35,18 +37,58 @@ func NamespaceAndName(objMeta metav1.Object) string {
...
-// namespace already exists, this function will return (false, nil).
-func EnsureNamespaceExists(namespace *corev1api.Namespace, client corev1client.NamespaceInterface) (bool, error) {
-	if _, err := client.Create(namespace); err == nil {
+// namespace already exists and is not ready, this function will return (false, nil).
+// If the namespace exists and is marked for deletion, this function will wait up to the timeout for it to fully delete.
+func EnsureNamespaceExistsAndIsReady(namespace *corev1api.Namespace, client corev1client.NamespaceInterface, timeout time.Duration) (bool, error) {
+	var ready bool
+	err := wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		clusterNS, err := client.Get(namespace.Name, metav1.GetOptions{})
+
```
skriss: style nit: rm blank line

pkg/util/kube/utils.go
```diff
@@ -35,18 +37,58 @@ func NamespaceAndName(objMeta metav1.Object) string {
...
+		if apierrors.IsNotFound(err) {
+			// Namespace isn't in cluster, we're good to create.
+			return true, nil
+		}
+
```
skriss: style nit: rm blank line

pkg/restore/restore.go
```diff
@@ -696,10 +798,15 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 
 			// Check if the PV exists in the cluster before attempting to create
 			// a volume from the snapshot, in order to avoid orphaned volumes (GH #609)
-			_, err := resourceClient.Get(name, metav1.GetOptions{})
+			shouldRestoreSnapshot, err := ctx.waitForPVDeletion(name, resourceClient)
```
carlisia: I think it would make the logic more clear if method and variable were both named the same, whichever makes the most sense. Either `waitForPVDeletion` or `shouldRestoreSnapshot`.

nrb: I'm not sure about naming the method and variable the same name; @skriss or @ncdc thoughts?

nrb:
@rbankston confirmed this; are there any other changes that need to be made?

pkg/restore/restore.go
nrb: I'll change the names to match and push another revision up.

nrb:
Function name change made, and rebased on the Velero rename commits. PTAL and if it's good, I'll squash.

nrb:
@skriss Are you ok with me squashing this for a merge?

Issue #576, opened 2018-06-24
Enable sharing bucket between Ark and Restic


rosskukulinski:
Right now Restic requires a separate bucket to be created with identical credentials as the main Ark server bucket.

For many users, having a separate bucket only makes the configuration & installation more complex.  Ark should allow sharing a bucket with Restic.

This would require creating root/parent-level directories to distinguish between backups and restic backup.  This would be a breaking change.

Perhaps we can link this with #103 which will also be having an impact on CRD compatibility.

ncdc:
I do think that how we configure backup storage and restic storage will potentially change slightly as we do replication.

We also have #167 which is related

skriss:
I'm leaning towards removing the ability for the location of the restic data to vary independently from the rest of the backup data, and just always making it a sub-dir of a backup storage location, e.g.:

```
<bucket> ark-backups
    cluster-1/
        _restic/
            ...
        backup-1/
            ...
        backup-2/
            ...
```
What do you all think about this? @rosskukulinski @ncdc @nrb @carlisia 

(by the way, using an `_` before `restic` would be an easy way for us to add structure to backup storage locations while still being able to support the legacy layout where backups are directly under the location, because backup names cannot start with an `_`.  So we'd know anything starting with an `_` is a system directory rather than a backup)

ncdc:
+1 to them sharing the same bucket in the BSL.

We alternatively could do
```
<bucket>
  <cluster>
    <backups>
      backup1
    <restic>
      ns1
```

skriss:
+1 to moving backups under a sub-dir too.  Are you saying use angle brackets to denote system dirs?  That works for me

ncdc:
No angle brackets. Those are variables.

```
(bucket=foo)
  cluster=abc
    backups
      my-backup1
      my-backup2
    restic
      my-namespace
```

We could even make a `clusters` dir as the root in the bucket/prefix.

nrb:
I like the proposed root/cluster/(backups,restic) approach.

Would `root/clusters/my-cluster` only be owned by `my-cluster`? We might need to brainstorm a bit on how that would work on migration restores - not that it's necessarily difficult, I just don't think we'd discussed that layout yet.

ncdc:
[...]

PR #830, opened 2018-09-07
[do not merge] proposal for new backup storage location layout

skriss:

We've discussed this across a few different issues now, and I wanted to put a formal proposal out for us all to review.  I think I'd like to do this for v0.10, as it will address #576 and will also be used by #825.  Please provide your input on both the content of the proposal and the timing. Thanks!

### DIFF
--- a/docs/PROPOSAL-backup-location-layout.md
+++ b/docs/PROPOSAL-backup-location-layout.md
@@ -1,0 +1,116 @@
+# Overview
+
+This document is a proposal for a revised layout for backup storage locations.
+
+# Current Layout
+
+Backup storage locations are structured as follows:
+```
+<bucket>/
+  <prefix>/ (optional)
+    backup-1/
+      ark-backup.json
+      backup-1.tar.gz
+      backup-1-logs.tar.gz
+      restore-of-backup-1-logs.gz
+      restore-of-backup-1-results.gz
+    backup-2/
+      ark-backup.json
+      backup-2.tar.gz
+      backup-2-logs.tar.gz
+      restore-of-backup-2-logs.gz
+      restore-of-backup-2-results.gz
+    ...
+  <prefix>/
+    ...
+```
+
+Ark's restic integration requires a separate "root" location, i.e. a different bucket + prefix.  Restic locations are structured as follows:
+```
+<bucket>/
+  <prefix>/ (optional)
+    namespace-1/
+      config
+      data/
+      index/
+      keys/
+      locks/
+      snapshots/
+    namespace-2/
+      config
+      data/
+      index/
+      keys/
+      locks/
+      snapshots/
+    ...
+```
+
+# Revised Layout
+
+The revised layout adds top-level "directories" within the root of the backup storage location to organize the contents.
+The initial set of top-level directories is:
+- **metadata** - holds metadata files describing the backup storage location, including `owner`, `revision`, etc.
+- **backups** - holds all backup-related data, similar to what's stored directly in the root of the backup storage location today.
+- **restores** - holds restore logs and results, with a subdirectory per restore. 
+- **restic** - holds the data currently stored in the restic location, i.e. a subdirectory per namespace, each containing a restic repository.
+
+Here's what it looks like:
+```
+<bucket>/
+  <prefix>/ (optional)
+    metadata/
+      revision
+      owner
+      ...
+    backups/
+      backup-1/
+        ark-backup.json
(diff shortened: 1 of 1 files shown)


docs/PROPOSAL-backup-location-layout.md
```diff
@@ -0,0 +1,116 @@
...
+      keys/
+      locks/
+      snapshots/
+    ...
+```
+
+# Revised Layout
+
+The revised layout adds top-level "directories" within the root of the backup storage location to organize the contents.
+The initial set of top-level directories is:
```
nrb: I would add how the concept of ownership and read/write access factors into each directory here.

e.g. I expect only the owner to write to `backups` and `restic`. All other clusters could read all data, and potentially write to `restores` and possibly `metadata`, such as when claiming ownership after a migration.

docs/PROPOSAL-backup-location-layout.md
```diff
@@ -0,0 +1,116 @@
...
+There are two main approaches I see for migrating to the new layout:
+- **Manual Migration** - the Ark codebase gets modified to support only the new layout. As part of the release containing this change,
+we provide a set of script templates, one per provider, to rearrange existing backup storage locations to conform to the new layout.
+These scripts would execute a series of commands (e.g. `aws s3 cp` for S3) to move existing directories/files around.
+- **Backwards Compatibility** - the Ark codebase is modified such that new backups/restores/etc. are stored within the new layout structure,
+but existing backup data stored directly in the root of the backup storage location is still recognized and supported as the basis for restores,
+delete operations, etc. Over time, all backups directly in the root should eventually be garbage-collected and only the new layout will remain.
+Additionally, we would probably remove all traces of the backwards-compatibility code as of Ark v1.0.
+
+My preferred approach is **manual migration**. We're still pre-1.0 so we don't make any guarantees about backwards compatibility. Any code that
```
nrb: I think I'd prefer manual migration, too, with the caveat that I think we will see quite a few questions from people shortly after whichever release contains this change happens. I think people trying to restore from backups made prior to this layout will be most affected, especially if they do a cluster migration and happen to be using the `latest` image tag. But, as is stated, we're currently not guaranteeing backwards compat.

We should probably have the instructions for migrating in the docs *and* front and center on our announcements.

docs/PROPOSAL-backup-location-layout.md
skriss: Will do. I think it probably makes the most sense for restore logs/results of backups not owned by the target cluster to just go into the target cluster's default location under restores/ - WDYT?

skriss: Added some info on this.  Gonna punt on deciding where restore logs go 'til next release.

docs/PROPOSAL-backup-location-layout.md
skriss: We could probably add some code that detects if a location is using the old layout and if so, log something pointing the user to a migration script and terminate.

nrb: +1 to that.

Honestly I don't think we can overcommunicate when changes like this happen - there'll always be someone who doesn't read the messages or errors, but plenty of pointers to the process should help.

skriss:
@carlisia @nrb @wwitzel3 @ncdc I'd like to decide soon if this layout makes sense, and if we want to do this as part of v0.10.  I'm in favor of both.  In particular, implementing this layout is going to make restic work related to locations much easier (i.e. sharing a bucket/prefix with the main Ark store, allowing multiple locations) and makes the sync controller improvements easier since we can put the `lastmodified` file into `metadata/`. 

Re: migration, I was thinking that the migration scripts can be really simple if we just move everything that's currently in the root of an Ark bucket into `backups/` in the new layout.  This means that any existing restore files won't go into `restores/`, so won't be accessible via CLI, but are still around.  I think this is fine since these are kind of ephemeral anyway and really only provide value immediately after a restore.  We could do the restore file migration too but the scripts will just get more complicated.  WDYT?

ncdc:
@skriss I'm probably going to be 100% heads down on other things this week. If I have time to come up for air, I'll review & respond. Otherwise, it'll be next week from me.

rosskukulinski:
I'm on board with manual migration, but we should definitely be clear in release notes, email to group, and I like the idea of exiting Ark Server if the storage layout doesn't match what it expects.

skriss:
I'm closing this PR - any future discussion can move to #853

PR #835, opened 2018-09-10
Make fsfreeze image building consistent

### LINKED ISSUE #833: image builds of master are failing
The changes from #795 are causing image builds of `master` to fail.  I haven't had a chance to look at why, but I think it's probably an easy fix.  I'll get to it later today or tomorrow if someone else doesn't take it first.

cc @carlisia @nrb @wwitzel3

nrb:
Due to the version of docker used to build images, the Dockerfile used
with -f must be in the same directory that's used for the context. Copy
the Dockerfile into the _output directory and make the custom targets
more closely match the standard ones.

Fixes #833

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ endif
 #		DOCKERFILE ?= Dockerfile.arm64 #aarch64/busybox
 #endif
 
-IMAGE := $(REGISTRY)/$(BIN)
+IMAGE = $(REGISTRY)/$(BIN)
 
 # If you want to build all binaries, see the 'all-build' rule.
 # If you want to build all containers, see the 'all-container' rule.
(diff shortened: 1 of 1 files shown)


nrb:
I have not yet tested this on Jenkins, however I've documented what I found in the commit. We may want to create a branch to build with, but then we'll end up with extra tags on GCR that will need to clean up.

skriss:
I'm testing locally by running each of the following commands:
```bash
REGISTRY=myregistry VERSION=master make all-containers
REGISTRY=myregistry VERSION=master make all-push
REGISTRY=myregistry VERSION=master make build-fsfreeze
REGISTRY=myregistry VERSION=master make push-fsfreeze
```

skriss:
I had to change https://github.com/heptio/ark/pull/835/files#diff-b67911656ef5d18c4ae36cb6741b7965R61 to `IMAGE = $(REGISTRY)/$(BIN)` which I think allows it to be re-evaluated. otherwise the fsfreeze-pause image was getting tagged as `.../ark:master`

skriss:
I'm fine merging this to master if it looks good locally. we can confirm it works there and fix if anything doesn't work.

Makefile
```diff
@@ -21,7 +21,7 @@ BIN ?= ark
 PKG := github.com/heptio/ark
 
 # Where to push the docker image.
-REGISTRY ?= gcr.io/heptio-images
```
skriss: undo :)

nrb: Fixed

Issue #840, opened 2018-09-11
POD + SA restores fail due to invalid token


james-powis:
**What steps did you take and what happened:**

Prerequisits: 
- Persistent volumes - I use https://github.com/vmware/vsphere-storage-for-kubernetes
- Restic configured ark
- Helm
- Ingress Controller (I use haproxy)

Steps to reproduce working example:
1. Create values-works.yaml:
```yaml
Master:
  ServiceType: ClusterIP
  HostName: jenkins.example.com
  PodAnnotations:
    backup.ark.heptio.com/backup-volumes: jenkins-home
Persistence:
  Enabled: true
  AccessMode: ReadWriteOnce
  Size: 8Gi
```
2. Deploy helm chart and create backup.
```
helm upgrade --install test-jenkins stable/jenkins -f values-works.yaml
# Monitor for everything being complete
ark create backup test-jenkins-works
# Monitor for backup completion.
```
3. Delete the helm chart and restore from backup
```
helm delete --purge test-jenkins
# Monitor for all objects being removed
ark create restore test-jenkins-works --from-backup test-jenkins-works
# Monitor the restore, it will complete successfully.
# Cleanup for next example:
helm delete --purge test-jenkins
```

Steps to create the issue which breaks the restore (all we are adding is the service account):
1. Create values-broken.yaml
```yaml
Master:
  ServiceType: ClusterIP
  HostName: jenkins.example.com
  PodAnnotations:
    backup.ark.heptio.com/backup-volumes: jenkins-home
Persistence:
  Enabled: true
  AccessMode: ReadWriteOnce
  Size: 8Gi
rbac:
  install: true
  serviceAccountName: jenkins
  apiVersion: v1
  roleRef: cluster-admin
  roleBindingKind: ClusterRoleBinding
```
2. Deploy helm chart and create backup.
```
helm upgrade --install test-jenkins stable/jenkins -f values-broken.yaml
# Monitor for everything being complete
ark create backup test-jenkins-broken
# Monitor for backup completion.
```
3. Delete the helm chart and restore from backup
```
helm delete --purge test-jenkins
# Monitor for all objects being removed
ark create restore test-jenkins-broken --from-backup test-jenkins-broken
# Monitor the restore, it will not complete
```

Cleanup... (note: this might not be the best way to clean everything up, but it worked for me.)

1. Delete the helm chart `helm delete --purge test-jenkins`
2. Remove the restore which is hung in progress: `ark delete restore test-jenkins`
[...]

PR #843, opened 2018-09-12
Discard all SA Token volumes from pod restore

james-powis:
Addresses #840 

SA Tokens are dynamically defined and auto attributed when ServiceAccountName is attributed to the pod. As such these should be safe to remove from the restore. To be doubly sure we are only deleting the secret from the restore if it matches ${serviceAccountName}-token-

### DIFF
--- a/pkg/restore/merge_service_account.go
+++ b/pkg/restore/merge_service_account.go
@@ -47,8 +47,8 @@ func mergeServiceAccounts(fromCluster, fromBackup *unstructured.Unstructured) (*
 
 	for i := len(backupSA.Secrets) - 1; i >= 0; i-- {
 		secret := &backupSA.Secrets[i]
-		if strings.HasPrefix(secret.Name, "default-token-") {
-			// Copy all secrets *except* default-token
+		if strings.HasPrefix(secret.Name, backupSA.Name+"-token-") {
+			// Copy all secrets *except* -token-
 			backupSA.Secrets = append(backupSA.Secrets[:i], backupSA.Secrets[i+1:]...)
 			break
 		}
--- a/pkg/restore/pod_action.go
+++ b/pkg/restore/pod_action.go
@@ -17,7 +17,7 @@ limitations under the License.
 package restore
 
 import (
-	"regexp"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 
@@ -41,10 +41,6 @@ func (a *podAction) AppliesTo() (ResourceSelector, error) {
 	}, nil
 }
 
-var (
-	defaultTokenRegex = regexp.MustCompile("default-token-.*")
-)
-
 func (a *podAction) Execute(obj runtime.Unstructured, restore *api.Restore) (runtime.Unstructured, error, error) {
 	a.logger.Debug("getting spec")
 	spec, err := collections.GetMap(obj.UnstructuredContent(), "spec")
@@ -55,6 +51,11 @@ func (a *podAction) Execute(obj runtime.Unstructured, restore *api.Restore) (run
 	a.logger.Debug("deleting spec.NodeName")
 	delete(spec, "nodeName")
 
+	serviceAccountName, err := collections.GetString(spec, "serviceAccountName")
+	if err != nil {
+		return nil, nil, err
+	}
+
 	newVolumes := make([]interface{}, 0)
 	a.logger.Debug("iterating over volumes")
 	err = collections.ForEach(spec, "volumes", func(volume map[string]interface{}) error {
@@ -64,11 +65,11 @@ func (a *podAction) Execute(obj runtime.Unstructured, restore *api.Restore) (run
 		}
 
 		a.logger.WithField("volumeName", name).Debug("Checking volume")
-		if !defaultTokenRegex.MatchString(name) {
+		if strings.HasPrefix(name, serviceAccountName+"-token-") {
+			a.logger.WithField("volumeName", name).Debug("Excluding volume")
+		} else {
 			a.logger.WithField("volumeName", name).Debug("Preserving volume")
 			newVolumes = append(newVolumes, volume)
-		} else {
-			a.logger.WithField("volumeName", name).Debug("Excluding volume")
 		}
 
 		return nil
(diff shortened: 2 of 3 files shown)


james-powis:
OK This is ready for review when someone has a chance.

jrnt30:
Nice job tracking this all down!  I think your long-term solution of associating the pod's SA name as an explicit filter would be a nice change here.  

I think the context should be available: 
- in merge_service_account via `backupSA.ObjectMeta.Name`
- in pod_action the unstructured content of the pod spec as `serviceAccountName`

ncdc:
@james-powis what if instead, we make sure that we restore all the secrets before we restore any serviceaccounts? I think this will help with every serviceaccount except for `default` (which is why we have the logic for ignoring `default-token-*`).

james-powis:
@ncdc Unfortunately that is not possible... While kubectl reports that the secret was created, it is discarded...

```
root@k8sm01:~/helm/jenkins/ark-testing# cat example.yaml 
apiVersion: v1
kind: ServiceAccount
metadata:
  name: my-sa
  namespace: default
---
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
  - name: nginx
    image: nginx:1.7.9
    ports:
    - containerPort: 80
  serviceAccountName: my-sa

root@k8sm01:~/helm/jenkins/ark-testing# kubectl apply -f example.yaml 
serviceaccount/my-sa created
pod/nginx created

root@k8sm01:~/helm/jenkins/ark-testing# kubectl get secret | grep my-sa
my-sa-token-pfx5k          kubernetes.io/service-account-token   3         27s

# backup each object in the order we want to restore in
root@k8sm01:~/helm/jenkins/ark-testing# kubectl get secret my-sa-token-pfx5k -o yaml > backup1.yaml 
root@k8sm01:~/helm/jenkins/ark-testing# kubectl get sa my-sa -o yaml > backup2.yaml 
root@k8sm01:~/helm/jenkins/ark-testing# kubectl get pod nginx -o yaml > backup3.yaml 

# delete so we can test restore of sa token
root@k8sm01:~/helm/jenkins/ark-testing# kubectl delete -f example.yaml 
serviceaccount "my-sa" deleted
pod "nginx" deleted

# restore the SA token
root@k8sm01:~/helm/jenkins/ark-testing# kubectl apply -f backup1.yaml 
secret/my-sa-token-pfx5k created
root@k8sm01:~/helm/jenkins/ark-testing# kubectl get secret | grep my-sa
root@k8sm01:~/helm/jenkins/ark-testing#
```

It didn't work... Looking at the dump of the sa-token easily shows us why...
```
apiVersion: v1
data:
  ca.crt: ...omitted for readability... 
  namespace: ZGVmYXVsdA==
  token: ...omitted for readability... 
  kind: Secret
metadata:
  annotations:
    kubernetes.io/service-account.name: my-sa <------- DOES NOT EXIST YET
    kubernetes.io/service-account.uid: bac6bc66-b765-11e8-a6ca-005056aea577 <------ WILL NEVER EXIST
  creationTimestamp: 2018-09-13T15:00:18Z
  name: my-sa-token-pfx5k
  namespace: default
  resourceVersion: "2889954"
  selfLink: /api/v1/namespaces/default/secrets/my-sa-token-pfx5k
  uid: bac88754-b765-11e8-a6ca-005056aea577
type: kubernetes.io/service-account-token
```

The SA tokens are internal K8S magic to which there is not really a reason to mess with it... There is no user data, it is created internally, and pods only need to reference the SA name to which k8s internals mount the sa-token volumes automagically.

jrnt30:
@ncdc I was making an assumption that the `default-token-*` was simply an oversight to the fact that the token pattern associated to SA accounts (which as James mentioned are typically automated and essentially opaque to users as I understand it).  

Is that an inaccurate assumption?  Any concern with explicitly detecting the SA's token and not restoring those (the same way that we do this for the default SA in a namespace)?

ncdc:
No, it’s not an inaccurate assumption. I was trying to look for an easier
fix. But I do agree looking at the secrets to identify them as tokens and
not restoring them is probably our only option.

pkg/restore/pod_action.go
```diff
@@ -64,7 +60,7 @@ func (a *podAction) Execute(obj runtime.Unstructured, restore *api.Restore) (run
 		}
 
 		a.logger.WithField("volumeName", name).Debug("Checking volume")
-		if !tokenRegex.MatchString(name) {
+		if !regexp.MustCompile(serviceAccountName + "-token-.*").MatchString(name) {
```
jrnt30: Somewhat nit picky, but is there any value in changing the way we compare these?  The regex vs. `strings.HasPrefix`?

james-powis: It would allow us to drop regexp entirely... Not a bad idea I will fold that in

james-powis:
All right gents, I am sticking a fork in this bad boy... We are now using the spec.serviceAccountName to determine what the token name is to drop. Also switched from using regexp to strings.HasPrefix to better align with the rest of the code base, and fixed the associated negative if statements to better conform to avoiding anti-patterns best practices... Let me know if anything else needs to be shored up.

pkg/restore/pod_action_test.go
```diff
@@ -40,31 +40,36 @@ func TestPodActionExecute(t *testing.T) {
...
 			expectedErr: false,
 			expectedRes: NewTestUnstructured().WithName("pod-1").WithSpec("foo").
+				WithSpec("serviceAccountName", "foo").
 				WithSpecField("volumes", []interface{}{}).
 				WithSpecField("containers", []interface{}{}).
 				Unstructured,
 		},
 		{
-			name: "volumes matching default-token regex should be deleted",
+			name: "volumes matching ServiceAccount-token- regex should be deleted",
```
nrb: Would you mind updating `regex` in these test names to be something more like `pattern`, since we're not using a regex anymore?

pkg/restore/merge_service_account.go
```diff
@@ -47,8 +47,8 @@ func mergeServiceAccounts(fromCluster, fromBackup *unstructured.Unstructured) (*
 
 	for i := len(backupSA.Secrets) - 1; i >= 0; i-- {
 		secret := &backupSA.Secrets[i]
-		if strings.HasPrefix(secret.Name, "default-token-") {
-			// Copy all secrets *except* default-token
+		if strings.HasPrefix(secret.Name, backupSA.ObjectMeta.Name+"-token-") {
```
skriss: nit, but you can just say `backupSA.Name` since `ObjectMeta` is embedded.

pkg/restore/pod_action_test.go
james-powis: It is done, Thanks!

pkg/restore/merge_service_account.go
james-powis: The nit has been picked.

PR #852, opened 2018-09-17
Bump alpine image for security fix

nrb:
This change includes the fix at https://github.com/docker-library/official-images/pull/4834

### DIFF
--- a/Dockerfile-ark-restic-restore-helper.alpine
+++ b/Dockerfile-ark-restic-restore-helper.alpine
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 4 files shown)


skriss:
@nrb do you want to include this in v0.9.6? As long as you've smoke-tested the new images I'm fine with it

nrb:
@skriss Probably doesn't hurt. I confirmed the images build. Let me test the result in a cluster before we decide.

nrb:
Tested, this appears to backup/restore correctly using this. There's a separate issue with the fsfreeze image, which I'll open a new PR for.

skriss:
cool, thanks for testing!

PR #853, opened 2018-09-18
Reorganize how data is stored in a backup storage location

skriss:
This PR implements the proposal in #830.

Unit tests are updated & pass and I've manually tested the code as well.

Tested the S3 migration script but I still need to test on GCS and Azure.

### DIFF
--- a/docs/storage-layout-reorg-v0.10.md
+++ b/docs/storage-layout-reorg-v0.10.md
@@ -1,0 +1,112 @@
+# Object Storage Layout Changes in v0.10
+
+## Overview
+
+Ark v0.10 includes breaking changes to where data is stored in your object storage bucket. You'll need to run a [one-time migration procedure](#upgrading-to-v0.10)
+if you're upgrading from prior versions of Ark.
+
+## Details
+
+Prior to v0.10, Ark stored data in an object storage bucket using the following structure:
+
+```
+<your-bucket>/
+    backup-1/
+        ark-backup.json
+        backup-1.tar.gz
+        backup-1-logs.gz
+        restore-of-backup-1-logs.gz
+        restore-of-backup-1-results.gz
+    backup-2/
+        ark-backup.json
+        backup-2.tar.gz
+        backup-2-logs.gz
+        restore-of-backup-2-logs.gz
+        restore-of-backup-2-results.gz
+    ...
+```
+
+As of v0.10, we've reorganized this layout to provide a cleaner and more extensible directory structure. The new layout looks like:
+
+```
+<your-bucket>[/<your-prefix>]/
+    backups/
+        backup-1/
+            ark-backup.json
+            backup-1.tar.gz
+            backup-1-logs.gz
+        backup-2/
+            ark-backup.json
+            backup-2.tar.gz
+            backup-2-logs.gz
+        ...
+    restores/
+        restore-of-backup-1/
+            restore-of-backup-1-logs.gz
+            restore-of-backup-1-results.gz
+        restore-of-backup-2/
+            restore-of-backup-2-logs.gz
+            restore-of-backup-2-results.gz
+        ...
+    ...
+```
+
+## Upgrading to v0.10
+
+Before upgrading to v0.10, you'll need to run a one-time upgrade script to rearrange the contents of your existing Ark bucket(s) to be compatible with
+the new layout.
+
+Please note that the following scripts **will not** migrate existing restore logs/results into the new `restores/` subdirectory. This means that they
+will not be accessible using `ark restore describe` or `ark restore logs`. They *will* remain in the relevant backup's subdirectory so they are manually
+accessible, and will eventually be garbage-collected along with the backup. We've taken this approach in order to keep the migration scripts simple 
+and less error-prone.
+
+### rclone-Based Script
+
+This script uses [rclone][1], which you can download and install following the instructions [here][2]. 
+Please read through the script carefully before starting and execute it step-by-step.
+
+```bash
+ARK_BUCKET=<your-ark-bucket>
+ARK_TEMP_MIGRATION_BUCKET=<a-temp-bucket-for-migration>
+
+# 1. This is an interactive step that configures rclone to be 
+#    able to access your storage provider. Follow the instructions,
+#    and keep track of the "remote name" for the next step:
+rclone config
+
+# 2. Store the name of the rclone remote that you just set up
+#    in Step #1:
+RCLONE_REMOTE_NAME=<your-remote-name>
+
+# 3. Create a temporary bucket to be used as a backup of your 
+#    current Ark bucket's contents:
+rclone mkdir ${RCLONE_REMOTE_NAME}:${ARK_TEMP_MIGRATION_BUCKET}
+
+# 4. Do a full copy of the contents of your Ark bucket into the 
+#    temporary bucket:
+rclone copy ${RCLONE_REMOTE_NAME}:${ARK_BUCKET} ${RCLONE_REMOTE_NAME}:${ARK_TEMP_MIGRATION_BUCKET}
+
+# 5. Verify that the temporary bucket contains an exact copy of 
+#    your Ark bucket's contents. You should see a short block
+#    of output stating "0 differences found":
+rclone check ${RCLONE_REMOTE_NAME}:${ARK_BUCKET} ${RCLONE_REMOTE_NAME}:${ARK_TEMP_MIGRATION_BUCKET}
+
+# 6. Delete your Ark bucket's contents (this command does not 
+#    delete the bucket itself, only the contents):
+rclone delete ${RCLONE_REMOTE_NAME}:${ARK_BUCKET}
+
+# 7. Copy the contents of the temporary bucket into your Ark bucket, 
+#    under the 'backups/' directory/prefix:
+rclone copy ${RCLONE_REMOTE_NAME}:${ARK_TEMP_MIGRATION_BUCKET} ${RCLONE_REMOTE_NAME}:${ARK_BUCKET}/backups
+
+# 8. Verify that the 'backups/' directory in your Ark bucket now
+#    contains an exact copy of the temporary bucket's contents:
+rclone check ${RCLONE_REMOTE_NAME}:${ARK_BUCKET}/backups ${RCLONE_REMOTE_NAME}:${ARK_TEMP_MIGRATION_BUCKET}
+
+# 9. Once you've confirmed that Ark v0.10 works with your revised Ark
+#    bucket, you can delete the temporary migration bucket.
+```
+
+[1]: https://rclone.org/
+[2]: https://rclone.org/downloads/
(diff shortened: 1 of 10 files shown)


skriss:
Ran into some issues with the GCS/Azure implementations of the script where it's not as easy as with S3 to move bucket contents into a sub-directory, so I decided to use `rclone` and was able to consolidate to a single script that works for all providers.  I've confirmed it does what I expect it to on Azure, but still need to do a full end-to-end upgrade test.

skriss:
@nrb @carlisia @wwitzel3 this is ready for a full review.

docs/storage-layout-reorg-v0.10.md
```diff
@@ -0,0 +1,100 @@
...
+    ...
+```
+
+## Upgrading to v0.10
+
+Before upgrading to v0.10, you'll need to run a one-time upgrade script to rearrange the contents of your existing Ark bucket(s) to be compatible with
+the new layout.
+
+Please note that the following scripts **will not** migrate existing restore logs/results into the new `restores/` subdirectory. This means that they
+will not be accessible using `ark restore describe` or `ark restore logs`. They *will* remain in the relevant backup's subdirectory so are manually
```
carlisia: s/so are/so they are

docs/storage-layout-reorg-v0.10.md
```diff
@@ -0,0 +1,100 @@
...
+accessible, and will eventually be garbage-collected along with the backup. We've taken this approach this in order to keep the migration scripts simple
```
carlisia: s/taken this approach this in order/taken this approach in order

docs/storage-layout-reorg-v0.10.md
```diff
@@ -0,0 +1,100 @@
...
+
+Please note that the following scripts **will not** migrate existing restore logs/results into the new `restores/` subdirectory. This means that they
+will not be accessible using `ark restore describe` or `ark restore logs`. They *will* remain in the relevant backup's subdirectory so are manually
+accessible, and will eventually be garbage-collected along with the backup. We've taken this approach this in order to keep the migration scripts simple 
+and less error-prone.
+
+### rclone-Based Script
+
+This script uses [rclone][1]. Please read through the script carefully before starting and execute it step-by-step.
+
```
carlisia: It's obvious but maybe add an instruction to install `rclone` + link (https://rclone.org/downloads/)?

docs/storage-layout-reorg-v0.10.md
```diff
@@ -0,0 +1,100 @@
...
+```bash
+PROVIDER=<azure|gcs|s3>
+ARK_BUCKET=<your-ark-bucket>
+ARK_TEMP_MIGRATION_BUCKET=<a-temp-bucket-for-migration>
+
+# 1. This is an interactive step that configures rclone to be 
+#    able to access your storage provider. Follow the instructions.
```
carlisia: Instead of ending these comments w/ `.`, I'd end them with a `:`.

pkg/cmd/server/server.go
```diff
@@ -257,6 +258,12 @@ func (s *server) run() error {
 		return err
 	}
 
+	// check to ensure backup storage locations exist and have a
```
carlisia: This comment should go on top of where the function is defined, imo.

skriss:
Addressed feedback.  I also did an end-to-end test of this on GKE by using ark:master to create some backups and restores, then ran the migration, then upgraded Ark to this PR, then confirmed backups  were there, restorable, and delete-able.

pkg/persistence/object_store_layout.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"fmt"
+	"path"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+var validRootDirs = sets.NewString("backups", "restores")
```
nrb: I assume you're leaving metadata out until it's actually implemented?

pkg/persistence/object_store_test.go
```diff
@@ -62,6 +62,99 @@ func newObjectBackupStoreTestHarness(bucket, prefix string) *objectBackupStoreTe
...
+			name:   "backup store with a prefix and only unsupported directories is invalid",
+			prefix: "backups",
+			storageData: map[string][]byte{
+				"backups/backup-1/ark-backup.json": {},
+				"backups/backup-2/ark-backup.json": {},
+			},
+			expectErr: true,
+		},
+		{
+			name: "backup store with no prefix and both supported and unsupported directories is invalid",
```
nrb: So the logic here is to make sure Ark still owns the root of the path it's given, whether that's the bucket root or a prefix inside it.

Is it this just to avoid future name collisions? Additionally, will the directory specified for restic be considered invalid in the current code, since it's only 2 hard-coded values?

docs/storage-layout-reorg-v0.10.md
```diff
@@ -0,0 +1,102 @@
+# Object Storage Layout Changes in v0.10
```
nrb: Thanks for adding this!

I think it would make sense to update our docs to have an upgrades landing page that links to each target version so this stuff is easily findable in the future, but we should do that in another PR.

pkg/persistence/object_store_layout.go
skriss: correct

pkg/persistence/object_store_test.go
skriss: Yeah, this is to ensure Ark owns the root location, and also that the migration was done properly.

It will, but I'm working on another PR for restic to share the bucket, at which point it'll be a valid dir.

docs/storage-layout-reorg-v0.10.md
skriss: Agree

docs/storage-layout-reorg-v0.10.md
```diff
@@ -0,0 +1,102 @@
...
+
+# 2. Create a temporary bucket to be used as a backup of your 
+#    current Ark bucket's contents:
+rclone mkdir ${PROVIDER}:${ARK_TEMP_MIGRATION_BUCKET}
+
+# 3. Do a full copy of the contents of your Ark bucket into the 
+#    temporary bucket:
+rclone copy ${PROVIDER}:${ARK_BUCKET} ${PROVIDER}:${ARK_TEMP_MIGRATION_BUCKET}
+
+# 4. STOP and manually verify that the temporary bucket contains
```
nrb: Past experience has led me to be a little wary of steps like this, as they'll often be skipped or skimmed since they're not executable lines.

Without adding too much logic, maybe we could diff the output of `rclone ls` on each bucket?

skriss: Let me see what I can come up with

skriss:
@nrb @carlisia @wwitzel3 any other comments on the approach?  I would like to get this merged pretty soon as I have a couple other PRs that are going to come on the back of it.

skriss:
@ncdc if you're up for air and have a chance to look at this, I'd appreciate it!

nrb:
@skriss I wanna run through the migration steps this afternoon, just so I have the experience if people need help. Should be done within an hour or so of this posting.

ncdc:
@skriss I skimmed it and it looks good. Do you plan to add additional directories in the bucket for things like restic snapshots in the future?

nrb:
@ncdc Yeah, per the convo at https://github.com/heptio/ark/pull/853/files/d70a6632fe245733a6920d18fab10ab3a756c31c#diff-1600269d68d610ae9c62198a23b72bb5R99 it's being left out until the logic to actually handle it is there.

docs/storage-layout-reorg-v0.10.md
```diff
@@ -0,0 +1,102 @@
...
+
+This script uses [rclone][1], which you can download and install following the instructions [here][2]. 
+Please read through the script carefully before starting and execute it step-by-step.
+
+```bash
+PROVIDER=<azure|gcs|s3>
+ARK_BUCKET=<your-ark-bucket>
+ARK_TEMP_MIGRATION_BUCKET=<a-temp-bucket-for-migration>
+
+# 1. This is an interactive step that configures rclone to be
```
nrb: Should probably mention in the comments that the later commands expect the remote you configure to be the value of `${PROVIDER}`. This tripped me up when following them exactly.

docs/storage-layout-reorg-v0.10.md
nrb: Might have to play with the options a little bit, but this is kind of what I had in mind.

```
% diff <(rclone lsf -R ${PROVIDER}:${ARK_TEMP_MIGRATION_BUCKET}) <(rclone lsf -R ${PROVIDER}:${ARK_BUCKET})

13a14,16
> default/ark-backup.json
> default/default-logs.gz
> default/default.tar.gz
19,21d21
< default/ark-backup.json
< default/default-logs.gz
< default/default.tar.gz
```

Used https://rclone.org/commands/rclone_ls/ for reference

nrb: Also, here's what it looked like after I did the `rclone delete`.

```
x1c in /home/nrb
% rclone delete ${PROVIDER}:${ARK_BUCKET}

x1c in /home/nrb
% diff <(rclone lsf -R ${PROVIDER}:${ARK_TEMP_MIGRATION_BUCKET}) <(rclone lsf -R ${PROVIDER}:${ARK_BUCKET})

1,24d0
< all/
< b1/
< def/
< default/
< nginx/
< nginx/ark-backup.json
< nginx/nginx-logs.gz
< nginx/nginx.tar.gz
< nginx/restore-nginx-20180921115829-logs.gz
< nginx/restore-nginx-20180921115829-results.gz
< def/ark-backup.json
< def/def-logs.gz
< def/def.tar.gz
< b1/ark-backup.json
< b1/b1-logs.gz
< b1/b1.tar.gz
< b1/restore-b1-20180907170648-logs.gz
< b1/restore-b1-20180907170648-results.gz
< default/ark-backup.json
< default/default-logs.gz
< default/default.tar.gz
< all/all-logs.gz
< all/all.tar.gz
< all/ark-backup.json
```

docs/storage-layout-reorg-v0.10.md
skriss: Ah, good call, this is an oversight. I'm actually going to insert a step right after `rclone config` to assign the name of the rclone remote to an env var and use that instead of `$PROVIDER` which is misleading.

docs/storage-layout-reorg-v0.10.md
skriss: Ooh, found an even better option - https://rclone.org/commands/rclone_check/

skriss:
@nrb if you get a chance, take a look at the updated migration script using `rclone check`

nrb:
@skriss I had read it this morning but hadn't gotten to double checking it.

Having just run `rclone check` post migration, it gives me output that seems pretty helpful, showing exactly what's missing. I haven't set it up for a full "happy path" migration yet, but I think the `check` command does what I was hoping for.

skriss:
@nrb I just added one additional `rclone check` step after copying data from the temp bucket back into the Ark bucket, under `backups/`.  Ran through the script start to finish and all LGTM

nrb:
@skriss Go ahead a squash

skriss:
squished!

PR #869, opened 2018-09-24
Fix error when restoring ExternalName services

### LINKED ISSUE #786: ExternalName service restores log errors saying "key ports not found"

Backup and restore an ExternalName service (https://kubernetes.io/docs/concepts/services-networking/service/#externalname)

The services restore properly, but the restore results show an error saying `key ports not found`

There shouldn't be any errors in the restore results.

This is caused by the restore service action which clears out auto-assigned NodePort values from service ports.  It returns an error if `.spec.ports` isn't found.  Should be updated to not error here, at least for `ExternalName` services.

shubheksha:
Fixes #786

### DIFF
--- a/pkg/restore/service_action.go
+++ b/pkg/restore/service_action.go
@@ -57,29 +57,10 @@ func (a *serviceAction) Execute(obj runtime.Unstructured, restore *api.Restore) 
 		delete(spec, "clusterIP")
 	}
 
-	preservedPorts, err := getPreservedPorts(obj)
+	err = deleteNodePorts(obj, &spec)
 	if err != nil {
 		return nil, nil, err
 	}
-
-	ports, err := collections.GetSlice(obj.UnstructuredContent(), "spec.ports")
-	serviceType, _ := collections.GetString(spec, "type")
-	if err != nil && serviceType != "ExternalName" {
-		return nil, nil, err
-	}
-
-	for _, port := range ports {
-		p := port.(map[string]interface{})
-		var name string
-		if nameVal, ok := p["name"]; ok {
-			name = nameVal.(string)
-		}
-		if preservedPorts[name] {
-			continue
-		}
-		delete(p, "nodePort")
-	}
-
 	return obj, nil, nil
 }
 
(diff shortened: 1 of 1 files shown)


shubheksha:
Made changes based on the discussion with @carlisia on Slack.
Not sure if this is the correct approach - how do we identify if a service is of type ExternalName?
I'd appreciate some direction here :)

pkg/restore/service_action.go
```diff
@@ -57,6 +57,11 @@ func (a *serviceAction) Execute(obj runtime.Unstructured, restore *api.Restore)
 		delete(spec, "clusterIP")
 	}
 
+	// Since spec.ports is an optional key, we can ignore 'not found' errors. Also assuming it was a string already.
+	if val, _ := collections.GetString(spec, "spec.ports"); val != "None" {
+		delete(spec, "spec.ports")
```
ncdc: I don't think we want to delete `spec.ports`; instead, down below where we access `spec.ports`, we want to be ok with the field not being there for `spec.type == ExternalName` services.

shubheksha: Ah, ok! Makes sense 👍

ncdc:
I think it might make the code more readable if we had a separate function for dealing with nodePorts (essentially all of https://github.com/heptio/ark/blob/a28327b47e80dcd4f393c0eba942c741280925db/pkg/restore/service_action.go#L60-L103). We could then either have it no-op if type=ExternalName, or do the type check first, and only call the separate function if type != ExternalName.

@carlisia how's that sound to you?

carlisia:
@ncdc I really like this. This piece of code is obscure enough that branching out and not having to look at it unless you need to in this case makes the code as a whole more readable. I also particularly like the no-op alternative.

nrb:
Thanks for the PR @shubheksha!

pkg/restore/service_action.go
```diff
@@ -57,28 +57,10 @@ func (a *serviceAction) Execute(obj runtime.Unstructured, restore *api.Restore)
 		delete(spec, "clusterIP")
 	}
 
-	preservedPorts, err := getPreservedPorts(obj)
-	if err != nil {
-		return nil, nil, err
-	}
-
-	ports, err := collections.GetSlice(obj.UnstructuredContent(), "spec.ports")
+	err = deleteNodePorts(obj, &spec)
```
ncdc: I know we just merged this, but:
`if err := ...; err != nil {`

pkg/restore/service_action.go
```diff
@@ -101,3 +83,29 @@ func getPreservedPorts(obj runtime.Unstructured) (map[string]bool, error) {
...
+
+func deleteNodePorts(obj runtime.Unstructured, spec *map[string]interface{}) error {
+	preservedPorts, err := getPreservedPorts(obj)
+	if err != nil {
+		return err
+	}
+
+	ports, err := collections.GetSlice(obj.UnstructuredContent(), "spec.ports")
+	serviceType, _ := collections.GetString(*spec, "type")
+	if err != nil && serviceType != "ExternalName" {
```
ncdc: I think it would be clearer if we instead checked if the type is ExternalName up front, either at the top of this function (with an immediate `return nil`) or in the calling function (don't call this function if ExternalName).

carlisia: I'm the one who opted for the alternative of a no-op but when I read it again I had the same thought. Apologies for misleading but this is a good minor change. I'll open a new tix.

PR #868, opened 2018-09-24
Remove cli reference docs generation

shubheksha:
This PR removes:
- cli reference docs folder
- scripts for updating and verifying docs

Fixes #798

PR #870, opened 2018-09-24
add create CLI command for snapshot locations

### LINKED ISSUE #802: Add `ark snapshot-location create` CLI command

wwitzel3:

Fixes #802 

I did not add the generated docs from `make update` because #868 as recently merged and those are going away anyway.

### DIFF
--- a/pkg/cmd/cli/snapshotlocation/create.go
+++ b/pkg/cmd/cli/snapshotlocation/create.go
@@ -1,0 +1,120 @@
+/*
+Copyright 2018 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


wwitzel3:
CI still looks for those auto generated CLI files so I added them in.

pkg/cmd/cli/snapshotlocation/create.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Factory) error {
+	if err := output.ValidateFlags(c); err != nil {
+		return err
+	}
+
+	if o.Provider == "" {
+		return errors.New("--provider is required")
+	}
```
carlisia: Do we need to check if the provider already exists? Or if it exists it should be overwritten? I'm not sure this question applies.

skriss: We want to be able to have >1 VolumeSnapshotLocation per provider.  No validation needed

skriss:
Ah, it's still expecting generated docs because this is on a feature branch and the removal of the docs is only in `master`.

carlisia:
Can we cherry-pick that into this branch?

skriss:
I think it'll be easiest to deal with the docs when we go to merge into `master`.  Shouldn't be too much more in the way of CLI work on this branch anyway.

Issue #864, opened 2018-09-21
Add --default-volume-snapshot-locations flag to `ark server` command


skriss:
example --default-volume-snapshot-locations aws:aws-default,rook:rook-default,...

- add the flag to the cobra command
- add a field to the server struct to store the value
- flag validation: ensure there's exactly one default location for each volume provider
    - ensure each provider exists, and each location exists
    - list all VolumeSnapshotLocations, get a unique list of providers, ensure there's one default per provider
- pass the list of default VSLs into each controller that needs it

skriss:
@wwitzel3 you doing OK here? Haven't heard from you in awhile :)

wwitzel3:
@skriss Yep, I've been trying to workout a good way to write tests here and haven't come up with any. I will squash things down and get a PR up shortly so it can start getting reviewed.

wwitzel3:
I was reviewing #873 and it looks like that is doing some of the things I assumed this was doing. Specifically the `pass the list of default VSLs into each controller that needs it`.

It seems this is doing that for the `BackupController`. Should I rebase of that PR? @skriss is this right?

carlisia:
Wait... I was doing this as part of my PR. Oh gosh. @wwitzel3 if you have started this, I have code that could be useful, particularly https://github.com/spf13/pflag/issues/182.

wwitzel3:
@carlisia No, you're good. I will get my PR up that binds and validates the default volume snapshot locations and then follow up with PRs for the other controllers. This works out because I just accidentally torched my controller changes in commit clean up attempt anyway.

skriss:
lol. @wwitzel3 I'd say don't bother with PRs for the other controllers, it can just be done inline with #807 #805 #808

PR #873, opened 2018-09-25
[WIP] Default and validate VolumeSnapshotLocations

carlisia:
Working on test, I'd appreciate a check on the logic.


pkg/controller/backup_controller.go
```diff
@@ -55,18 +57,22 @@ const backupVersion = 1
...
+	newPluginManager             func(logrus.FieldLogger) plugin.Manager
+	backupTracker                BackupTracker
+	backupLocationLister         listers.BackupStorageLocationLister
+	defaultBackupLocation        string
+	volumeSnapshotLocationLister listers.VolumeSnapshotLocationLister
+	volumeSnapshotLocations      []string
+	metrics                      *metrics.ServerMetrics
+	newBackupStore               func(*api.BackupStorageLocation, persistence.ObjectStoreGetter, logrus.FieldLogger) (persistence.BackupStore, error)
+	// TODO: figure out how to create a new volume
+	newVolume func(*api.VolumeSnapshotLocation, logrus.FieldLogger) (persistence.BackupStore, error)
```
skriss: I *think* we'll be instantiating block stores within `pkg/backup` (which I'm guessing is what this addition was eventually intended to do) - so I don't think this is needed here

pkg/controller/backup_controller.go
```diff
@@ -55,18 +57,22 @@ const backupVersion = 1
...
+	lister                       listers.BackupLister
+	client                       arkv1client.BackupsGetter
+	clock                        clock.Clock
+	backupLogLevel               logrus.Level
+	newPluginManager             func(logrus.FieldLogger) plugin.Manager
+	backupTracker                BackupTracker
+	backupLocationLister         listers.BackupStorageLocationLister
+	defaultBackupLocation        string
+	volumeSnapshotLocationLister listers.VolumeSnapshotLocationLister
+	volumeSnapshotLocations      []string
```
skriss: Maybe `defaultSnapshotLocations`?  Also, I think it might be easier if this was a `map[string]string`, with provider names as keys, and VSL names as values? And the `NewBackupController` func can just require it to be passed in as a map (which #864 can do)

pkg/controller/backup_controller.go
```diff
@@ -178,9 +188,14 @@ func (c *backupController) processBackup(key string) error {
 		backup.Status.Expiration = metav1.NewTime(c.clock.Now().Add(backup.Spec.TTL.Duration))
 	}
 
-	var backupLocation *api.BackupStorageLocation
-	// validation
-	if backupLocation, backup.Status.ValidationErrors = c.getLocationAndValidate(backup, c.defaultBackupLocation); len(backup.Status.ValidationErrors) > 0 {
+	backupLocation, statusBackup := c.getLocationAndValidate(backup, c.defaultBackupLocation)
+	backup.Status.ValidationErrors = append(backup.Status.ValidationErrors, statusBackup...)
+
+	// TODO: add VSLs to the actual backup
```
skriss: I'm thinking we don't need to store the `VolumeSnapshotLocation` objects anywhere within the controller.  `pkg/backup` will handle instantiating a BlockStore based on the appropriate location as necessary.  This should make the signature of this function simpler, and it can probably be called from within `getLocationAndValidate`

pkg/controller/backup_controller.go
```diff
@@ -207,6 +222,7 @@ func (c *backupController) processBackup(key string) error {
 	backupScheduleName := backup.GetLabels()["ark-schedule"]
 	c.metrics.RegisterBackupAttempt(backupScheduleName)
 
+	// TODO: add VSLs to the actual backup
```
skriss: per previous comment, I don't think we actually need to do this in the controller

pkg/controller/backup_controller.go
```diff
@@ -281,6 +293,45 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
 	return backupLocation, validationErrors
 }
 
+// getVolumeLocationAndValidate validates that, for each default storage provider, if
+// backup.spec.volumeSnapshotLocations doesn't already already have a location for
+// a provider, we then add the default VolumeSnapshotLocation to that list of providers
+func (c *backupController) getVolumeLocationAndValidate(itm *api.Backup, defaultVSLNames []string) ([]*api.VolumeSnapshotLocation, []string) {
```
skriss: let's pair on this function - whenever you're ready

skriss:
@carlisia need any more input from me here atm?

carlisia:
No thank you. I did some variable renaming to make it more clear (to me!). I finished implementing your suggestions, changing the test.

pkg/controller/backup_controller.go
```diff
@@ -178,9 +187,13 @@ func (c *backupController) processBackup(key string) error {
 		backup.Status.Expiration = metav1.NewTime(c.clock.Now().Add(backup.Spec.TTL.Duration))
 	}
 
-	var backupLocation *api.BackupStorageLocation
-	// validation
-	if backupLocation, backup.Status.ValidationErrors = c.getLocationAndValidate(backup, c.defaultBackupLocation); len(backup.Status.ValidationErrors) > 0 {
+	backupLocation, statusBackup := c.getLocationAndValidate(backup, c.defaultBackupLocation)
```
skriss: `statusBackup` -> `errs` or `validationErrs` or something similar

pkg/controller/backup_controller.go
```diff
@@ -178,9 +187,13 @@ func (c *backupController) processBackup(key string) error {
...
+	backup.Status.ValidationErrors = append(backup.Status.ValidationErrors, statusBackup...)
+
+	statusVolume := c.getVolumeLocationAndValidate(backup, c.defaultSnapshotLocations)
```
skriss: can just reuse the var from the previous block (`errs` or `validationErrs`)

pkg/controller/backup_controller.go
```diff
@@ -281,6 +290,47 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
 	return backupLocation, validationErrors
 }
 
+// getVolumeLocationAndValidate ensures:
+// - each VSL name in Spec VolumeSnapshotLocation exists as a VSL
+// - exactly 1 VSL per existing or default provider
+// - a given default provider's VSL name is added to the Spec VolumeSnapshotLocation if it does not exist as a VSL
+func (c *backupController) getVolumeLocationAndValidate(itm *api.Backup, defaultVSLs map[provider]string) []string {
```
skriss: maybe `defaultAndValidateSnapshotLocations`? It's a long one I know but i think that's a better description of what it does

skriss: Personally speaking I find acronyms in the middle of variable names more difficult to read.  WDYT about just using `location` in place of `VSL` - since this function is all about snapshot locations, there's not really any ambiguity about what `location` means?

pkg/controller/backup_controller.go
```diff
@@ -281,6 +290,47 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
...
+	var errors []string
+	perProviderVSLName := make(map[provider]string)
+	var finalVSLNamesList []string
+	for _, specVSLName := range itm.Spec.VolumeSnapshotLocations {
+		// validate each specVSLName exists as a VolumeSnapshotLocation
+		vsl, err := c.snapshotLocationLister.VolumeSnapshotLocations(itm.Namespace).Get(specVSLName)
+		if err != nil {
+			errors = append(errors, fmt.Sprintf("Error getting volume snapshot location for %s: %v", specVSLName, err))
```
skriss: maybe "error getting volume snapshot location **named** %s: ..."?

pkg/controller/backup_controller.go
```diff
@@ -281,6 +290,47 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
...
+			continue
+		}
+
+		// ensure we end up with exactly 1 specVSLName *per provider*
+		providerVSLName := perProviderVSLName[provider(vsl.Spec.Provider)]
+		if providerVSLName != "" {
+			// if > 1 VSL name per provider as in ["aws-us-east-1" | "aws-us-west-1"] (same provider, different names)
+			if providerVSLName != specVSLName {
+				errors = append(errors, fmt.Sprintf("more than one VolumeSnapshotLocation found for provider %s: %s", vsl.Spec.Provider, specVSLName))
```
skriss: s/found/specified

pkg/controller/backup_controller.go
```diff
@@ -281,6 +290,47 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
...
+				continue
+			}
+
+			perProviderVSLName[provider(vsl.Spec.Provider)] = specVSLName
```
skriss: I think you want this line after L320 - I don't think it will ever be hit as-is because this is only executing if there's already an entry in the map for this provider, but this is the only place we'd insert into the map

pkg/controller/backup_controller.go
```diff
@@ -281,6 +290,47 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
...
+			finalVSLNamesList = append(finalVSLNamesList, specVSLName)
```
skriss: I think you want this in an `else` block following L320 - you want to add `specVSLName` to the final list only if it's not a dupe, meaning there wasn't already an entry in `perProviderVSLName` for its provider

pkg/controller/backup_controller.go
```diff
@@ -281,6 +290,47 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
...
+	}
+
+	for provider, defaultVSLName := range defaultVSLs {
+		// if the VSL name for a given default provider does not exist, add the default VSL name for that provider
+		if _, ok := perProviderVSLName[provider]; !ok {
+			finalVSLNamesList = append(finalVSLNamesList, defaultVSLName)
+		}
+	}
+
+	itm.Spec.VolumeSnapshotLocations = finalVSLNamesList
```
skriss: I think we should probably only update `itm.Spec.VolumeSnapshotLocations` if there were no validation errs - otherwise it's going to be very confusing for the user trying to figure out what they did wrong

pkg/controller/backup_controller.go
```diff
@@ -281,6 +288,57 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
 	return backupLocation, validationErrors
 }
 
+// defaultAndValidateSnapshotLocations ensures:
+// - each location name in Spec VolumeSnapshotLocation exists as a location
+// - exactly 1 location per existing or default provider
+// - a given default provider's location name is added to the Spec VolumeSnapshotLocation if it does not exist as a VSL
+func (c *backupController) defaultAndValidateSnapshotLocations(itm *api.Backup, defaultLocations map[provider]string) []string {
+	var errors []string
+	perProviderLocationName := make(map[provider]string)
```
skriss: not sure how much value the `provider` type alias is adding at this point - i'd probably just drop it. the var name makes it pretty clear what the key is.

pkg/controller/backup_controller.go
```diff
@@ -281,6 +288,57 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
...
+				fmt.Println("!!! error!!!")
+				errors = append(errors, fmt.Sprintf("more than one VolumeSnapshotLocation specified for provider %s: %s", location.Spec.Provider, locationName))
+				continue
+			}
+		} else {
+			// no dup exists: add locationName to the final list
+			finalLocationNameList = append(finalLocationNameList, locationName)
+		}
+		// keep track of all valid existing locations, per provider
+		perProviderLocationName[provider(location.Spec.Provider)] = locationName
```
skriss: I guess I'd probably put this in the `else` block since we only need to to it the first time we encounter a VSL for a provider. functionally the same but i think slightly clearer that way.

pkg/controller/backup_controller.go
```diff
@@ -281,6 +288,57 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
...
+	}
+
+	if len(errors) > 0 {
+		return errors
+	}
+
+	for provider, defaultLocationName := range defaultLocations {
+		// if a default location name for a given provider does not already exist, add it for that provider
```
skriss: "if a location name for a given provider does not already exist, add the provider's default"

carlisia:
I squashed all the things and cleaned tests + added the remainder.

pkg/controller/backup_controller.go
```diff
@@ -281,6 +286,53 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
 	return backupLocation, validationErrors
 }
 
+// defaultAndValidateSnapshotLocations ensures:
+// - each location name in Spec VolumeSnapshotLocation exists as a location
+// - exactly 1 location per existing or default provider
+// - a given default provider's location name is added to the Spec VolumeSnapshotLocation if it does not exist as a VSL
+func (c *backupController) defaultAndValidateSnapshotLocations(itm *api.Backup, defaultLocations map[string]string) []string {
```
carlisia: Should `defaultLocations map[string]string` be `defaultLocations []map[string]string` instead?

carlisia: Need to know this ^ before I finish fixing the server cmd flag bug.

skriss: no - should be a map of provider->VSL name

skriss:
@nrb @wwitzel3 pls review when you get a chance

pkg/util/test/test_volume_snapshot_location.go
```diff
@@ -0,0 +1,72 @@
+/*
+Copyright 2017 the Heptio Ark contributors.
```
wwitzel3: 2018 :)

pkg/controller/backup_controller_test.go
```diff
@@ -422,3 +426,150 @@ func TestProcessBackup(t *testing.T) {
...
+				sharedInformers = informers.NewSharedInformerFactory(client, 0)
+				logger          = logging.DefaultLogger(logrus.DebugLevel)
+				pluginManager   = &pluginmocks.Manager{}
+				backupStore     = &persistencemocks.BackupStore{}
+			)
+			defer backupper.AssertExpectations(t)
+			defer pluginManager.AssertExpectations(t)
+			defer backupStore.AssertExpectations(t)
+
+			c := NewBackupController(
```
skriss: it looks like the only field of `*backupController` that `defaultAndValidateSnapshotLocations` uses is the `snapshotLocationLister` - so I'd switch this to just say: 
```
c := &backupController{
  snapshotLocationLister: sharedInformers.Ark().V1().VolumeSnapshotLocations().Lister(),
}
```
and then get rid of all the extraneous dependencies that aren't actually used

pkg/controller/backup_controller_test.go
```diff
@@ -422,3 +426,150 @@ func TestProcessBackup(t *testing.T) {
...
+				sharedInformers.Ark().V1().VolumeSnapshotLocations(),
+				test.defaultLocations,
+				metrics.NewServerMetrics(),
+			).(*backupController)
+
+			c.newBackupStore = func(*v1.BackupStorageLocation, persistence.ObjectStoreGetter, logrus.FieldLogger) (persistence.BackupStore, error) {
+				return backupStore, nil
+			}
+
+			// set up a Backup object to represent what we expect to be passed to backupper.Backup()
```
skriss: L541-561 aren't really necessary - all you need to do is call `sharedInformers.Ark().V1().VolumeSnapshotLocations().Informer().GetStore().Add(...)` for each item in `test.locations`

pkg/controller/backup_controller_test.go
```diff
@@ -422,3 +426,150 @@ func TestProcessBackup(t *testing.T) {
...
+					},
+					Spec: v1.VolumeSnapshotLocationSpec{
+						Provider: location.Spec.Provider,
+						Config:   location.Spec.Config,
+					},
+				}
+				require.NoError(t, sharedInformers.Ark().V1().VolumeSnapshotLocations().Informer().GetStore().Add(vsl))
+			}
+
+			errs := c.defaultAndValidateSnapshotLocations(backup, test.defaultLocations)
```
skriss: It would be nice to also have the test validate that `test.backup.Spec.VolumeSnapshotLocations` and `test.backup.Status.ValidationErrors` contain what you expect.

pkg/controller/backup_controller_test.go
carlisia: It seems I need to create a `v1.VolumeSnapshotLocatio` out of the test one. This is what I got if I used the test:

```
TestDefaultAndValidateSnapshotLocations/duplicate_locationName_per_provider:_should_filter_out_dups
panic: interface conversion: interface {} is *test.TestVolumeSnapshotLocation, not *v1.VolumeSnapshotLocation [recovered]
	panic: interface conversion: interface {} is *test.TestVolumeSnapshotLocation, not *v1.VolumeSnapshotLocation

```

pkg/controller/backup_controller_test.go
carlisia: Good point, added.

pkg/controller/backup_controller_test.go
skriss: Ah. for each item in `test.location`, add `location.VolumeSnapshotLocation`.

pkg/controller/backup_controller.go
```diff
@@ -281,6 +286,53 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
...
+			errors = append(errors, fmt.Sprintf("error getting volume snapshot location named %s: %v", locationName, err))
+			continue
+		}
+
+		// ensure we end up with exactly 1 locationName *per provider*
+		providerLocationName := perProviderLocationName[location.Spec.Provider]
+		if providerLocationName != "" {
+			// if > 1 location name per provider as in ["aws-us-east-1" | "aws-us-west-1"] (same provider, multiple names)
+			if providerLocationName != locationName {
+				errors = append(errors, fmt.Sprintf("more than one VolumeSnapshotLocation specified for provider %s: %s", location.Spec.Provider, locationName))
```
skriss: We should probably also include `providerLocationName` in this error msg to be clear about what the dupe for the provider is.  Picked up on this from looking at the unit test msgs.

skriss:
2 more minor things then LGTM.  If this is ready to merge before #892 then you'll want to fix the compile error by passing a `nil` or empty map to the `NewBackupController` func in `server.go`.

@nrb PTAL.

pkg/controller/backup_controller_test.go
carlisia: Ohhhh

pkg/cmd/server/server.go
```diff
@@ -606,6 +606,8 @@ func (s *server) runControllers(config *api.Config, defaultBackupLocation *api.B
 			backupTracker,
 			s.sharedInformerFactory.Ark().V1().BackupStorageLocations(),
 			s.config.defaultBackupLocation,
+			nil,
```
skriss: you can pass pass the VSL informer here (`s.sharedInformerFactory.Ark().V1().VolumeSnapshotLocations()`), just leave the line below as `nil` 'til wayne's PR goes in

PR #874, opened 2018-09-25
Add support for bulk deletion to `ark schedule delete`

### LINKED ISSUE #748: Add bulk deletion support to ark schedule delete
Modify ark schedule delete to use a --pattern flag, so that a user could do bulk delete with a PCRE pattern as follows ark schedule delete --pattern nginx*

This is related to this issue: https://github.com/heptio/ark/issues/295

shubheksha:
Signed-off-by: Shubheksha Jalan <jshubheksha@gmail.com>
Fixes #748 
/cc @carlisia @skriss


pkg/cmd/cli/schedule/delete.go
```diff
@@ -19,29 +19,149 @@ package schedule
...
 
-			name := args[0]
+	o.BindFlags(c.Flags())
+	return c
+}
 
-			err = arkClient.ArkV1().Schedules(f.Namespace()).Delete(name, nil)
-			cmd.CheckError(err)
+// DeleteOptions specifies params used for deleting a schedule.
+type DeleteOptions struct {
```
shubheksha: This struct is common across all three commands

pkg/cmd/cli/schedule/delete.go
```diff
@@ -19,29 +19,149 @@ package schedule
...
 	}
+	o.client = client
+	o.Names = args
 
-	return c
+	return nil
+}
+
+// Validate validatest the params for deletion of schedules.
+func (o *DeleteOptions) Validate(c *cobra.Command, f client.Factory, args []string) error {
```
shubheksha: This method too

pkg/cmd/cli/schedule/delete.go
```diff
@@ -19,29 +19,149 @@ package schedule
...
+	namespace string
+	All       bool
+	Selector  flag.LabelSelector
+	Confirm   bool
+}
 
-			fmt.Printf("Schedule %q deleted\n", name)
-		},
+// Complete fills in missing options required for deletion of schedules to work.
+func (o *DeleteOptions) Complete(f client.Factory, args []string) error {
```
shubheksha: And this as well

pkg/cmd/cli/schedule/delete.go
shubheksha: Would it make sense to extract these in a common place?

nrb: @shubheksha Thanks, I was just looking and saw the repetition.

I'm not opposed to putting the `DeleteOptions` type and its methods into a central location. At first I thought `pkg/cmd/cli/delete` might be a good place for it, but I think that would create circular dependencies, as it imports these files to provide the same functions, just with reversed names.

Perhaps `pkg/cmd/cli/types`?

shubheksha: SGTM!
What about the `Complete()` and `Validate()` methods? AFAICT, they are identical. Would it make sense to move those to a common place too?

nrb: @shubheksha Since the methods are on that type, moving them to the same file makes sense to me, but I'm not sure if that's a go antipattern. @carlisia what do you think?

shubheksha: Ping @carlisia 
I wasn't sure if it's an anti-pattern either. How does one decide if it's one or not? 😄

carlisia: Hi, I missed the ping, sorry! Not an anti-pattern, it's exactly what I would do (keep the types + methods in the same file). How about we add them to this file: https://github.com/heptio/ark/blob/master/pkg/cmd/cli/common.go

pkg/cmd/cli/common.go
```diff
@@ -18,11 +18,65 @@ package cli
...
+	clientset "github.com/heptio/ark/pkg/generated/clientset/versioned"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
+// DeleteOptions contains parameters used for deleting a restore.
+type DeleteOptions struct {
+	Names     []string
+	All       bool
```
carlisia: `All` and any other element that is not used outside this package can be unexported.

carlisia:
@shubheksha you may squash everything and be ready for one more review!

pkg/cmd/cli/common.go
```diff
@@ -18,11 +18,65 @@ package cli
 
 import (
```
skriss: I'd probably rename this file to `delete_options.go` since that's now what it contains.

pkg/cmd/cli/common.go
```diff
@@ -18,11 +18,65 @@ package cli
...
+	if o.Client == nil {
+		return errors.New("Ark client is not set; unable to proceed")
+	}
+	var (
+		hasNames    = len(o.Names) > 0
+		hasAll      = o.all
+		hasSelector = o.Selector.LabelSelector != nil
+	)
+	if !xor(hasNames, hasAll, hasSelector) {
+		return errors.New("you must specify exactly one of: specific restore name(s), the --all flag, or the --selector flag")
```
skriss: This error currently hardcodes the string "restore" which isn't what we want since we're using this struct in multiple places.  I'd add a `NewDeleteOptions(singularTypeName string)` constructor function that takes the type this is being used for as an arg (`backup`, `restore`, `schedule`), then use that in this error message and in the `BindFlags` func.

pkg/cmd/cli/common.go
```diff
@@ -18,11 +18,65 @@ package cli
...
+	}
+
+	return nil
+}
+
+// BindFlags binds options for this command to flags.
+func (o *DeleteOptions) BindFlags(flags *pflag.FlagSet) {
+	flags.BoolVar(&o.Confirm, "confirm", o.Confirm, "Confirm deletion")
+	flags.BoolVar(&o.all, "all", o.all, "Delete all backups")
```
skriss: replace hardcoded "backups" with the type name passed to the constructor plus a trailing `s`.

pkg/cmd/cli/common.go
```diff
@@ -18,11 +18,65 @@ package cli
...
+	flags.VarP(&o.Selector, "selector", "l", "Delete all backups matching this label selector")
```
skriss: replace hardcoded "backups" with the type name passed to the constructor, plus a trailing `s`.

shubheksha:
@skriss, addressed your comments, PTAL

pkg/cmd/cli/delete_options.go
```diff
@@ -0,0 +1,123 @@
...
+	Names            []string
+	all              bool
+	Selector         flag.LabelSelector
+	Confirm          bool
+	Client           clientset.Interface
+	Namespace        string
+	singularTypeName string
+}
+
+func (o *DeleteOptions) NewDeleteOptions(singularTypeName string) {
```
skriss: Let's remove the receiver from this function, and have it return a *DeleteOptions, so:
```
func NewDeleteOptions(singularTypeName string) *DeleteOptions {
    // instantiate and return a *DeleteOptions
}
```

shubheksha:
@skriss, done!

pkg/cmd/cli/delete_options.go
```diff
@@ -0,0 +1,125 @@
...
+	if o.Client == nil {
+		return errors.New("Ark client is not set; unable to proceed")
+	}
+	var (
+		hasNames    = len(o.Names) > 0
+		hasAll      = o.all
+		hasSelector = o.Selector.LabelSelector != nil
+	)
+	if !xor(hasNames, hasAll, hasSelector) {
+		return errors.New("you must specify exactly one of: specific restore name(s), the --all flag, or the --selector flag")
```
skriss: sorry, just spotted one more place with `restore` hardcoded that needs to be replaced with `o.singularTypeName`.  You can just amend the existing commit, no need to push a separate one.

PR #876, opened 2018-09-25
Move restic repos into the main Ark bucket, under restic/

### LINKED ISSUE #576: Enable sharing bucket between Ark and Restic

skriss:
Fixes #576 

Note that this PR **requires** that restic share the main Ark bucket, not just *enables* it

### DIFF
--- a/docs/restic.md
+++ b/docs/restic.md
@@ -23,14 +23,8 @@ cross-volume-type data migrations. Stay tuned as this evolves!
 
 ### Prerequisites
 
-- A working install of Ark version 0.9.0 or later. See [Set up Ark][2]
+- A working install of Ark version 0.10.0 or later. See [Set up Ark][2]
 - A local clone of [the latest release tag of the Ark repository][3]
-
-#### Additional steps if upgrading from version 0.9 alpha
-
-- Manually delete all of the repositories/data from your existing restic bucket
-- Delete all Ark backups from your cluster using `ark backup delete`
-- Delete all secrets named `ark-restic-credentials` across all namespaces in your cluster
 
 ### Instructions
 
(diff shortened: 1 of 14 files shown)


nrb:
Haven't read in depth yet, but thanks for providing docs with this! Does this change still send restic directly and only to the `default` backup storage location, or can it be directed to a different one?

skriss:
This PR still only sends them to the default location.  I've got another branch on-deck to support multiple locations.

nrb:
Walking through the migration steps again, somewhat unrelated comment:  L100 of `storage-layout-reorg-v0.10.md` says to execute the script step by step - if people are copying their commands into the terminal one at a time, the environment variables need to be `export`ed. I think that's going to be the primary method by which people interact with it, versus copying it into a file.

skriss:
@carlisia @wwitzel3 would appreciate comments or a merge when you get a chance, thanks!

wwitzel3:
Want to make sure I am following this correctly. We are removing the restic location and replacing it with a default under the new directory structure of `/restic`.

The rclone migration will move existing restic data to that location in our new expected format.

We no longer need to pass the restic location around because it is not determined by us, so if you enable restic, you can expect to see the `/restic` folder turn up.

skriss:
@wwitzel3 correct on all counts

PR #888, opened 2018-09-28
support multiple backup locations for restic

skriss:
This adds proper multi-location support for restic backups.  I've tested the core functionality but haven't yet tested the env-var stuff on Azure yet; will do so shortly.

I'm happy to walk folks through this if it helps, since I know nobody's touched the restic code much apart from me.

### DIFF
--- a/pkg/apis/ark/v1/labels_annotations.go
+++ b/pkg/apis/ark/v1/labels_annotations.go
@@ -40,4 +40,8 @@ const (
 	// StorageLocationLabel is the label key used to identify the storage
 	// location of a backup.
 	StorageLocationLabel = "ark.heptio.com/storage-location"
+
+	// ResticVolumeNamespaceLabel is the label key used to identify which
+	// namespace a restic repository stores pod volume backups for.
+	ResticVolumeNamespaceLabel = "ark.heptio.com/volume-namespace"
 )
--- a/pkg/apis/ark/v1/restic_repository.go
+++ b/pkg/apis/ark/v1/restic_repository.go
@@ -22,12 +22,20 @@ import (
 
 // ResticRepositorySpec is the specification for a ResticRepository.
 type ResticRepositorySpec struct {
-	// MaintenanceFrequency is how often maintenance should be run.
-	MaintenanceFrequency metav1.Duration `json:"maintenanceFrequency"`
+	// VolumeNamespace is the namespace this restic repository contains
+	// pod volume backups for.
+	VolumeNamespace string `json:"volumeNamespace"`
+
+	// BackupStorageLocation is the name of the BackupStorageLocation
+	// that should contain this repository.
+	BackupStorageLocation string `json:"backupStorageLocation"`
 
 	// ResticIdentifier is the full restic-compatible string for identifying
 	// this repository.
 	ResticIdentifier string `json:"resticIdentifier"`
+
+	// MaintenanceFrequency is how often maintenance should be run.
+	MaintenanceFrequency metav1.Duration `json:"maintenanceFrequency"`
 }
 
 // ResticRepositoryPhase represents the lifecycle phase of a ResticRepository.
--- a/pkg/controller/pod_volume_backup_controller.go
+++ b/pkg/controller/pod_volume_backup_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/pkg/errors"
@@ -50,6 +51,7 @@ type podVolumeBackupController struct {
 	secretLister          corev1listers.SecretLister
 	podLister             corev1listers.PodLister
 	pvcLister             corev1listers.PersistentVolumeClaimLister
+	backupLocationLister  listers.BackupStorageLocationLister
 	nodeName              string
 
 	processBackupFunc func(*arkv1api.PodVolumeBackup) error
@@ -64,6 +66,7 @@ func NewPodVolumeBackupController(
 	podInformer cache.SharedIndexInformer,
 	secretInformer cache.SharedIndexInformer,
 	pvcInformer corev1informers.PersistentVolumeClaimInformer,
+	backupLocationInformer informers.BackupStorageLocationInformer,
 	nodeName string,
 ) Interface {
 	c := &podVolumeBackupController{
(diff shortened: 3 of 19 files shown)


skriss:
Note: Backport candidate label applies to commit https://github.com/heptio/ark/pull/888/commits/d0411f12eaa720cdd421f1c4f331bf6ffe7876ee only.

skriss:
FYI working on testing this on azure now.

skriss:
okay, tested on azure, fixed one small bug but otherwise this is looking good to me.

pkg/apis/ark/v1/labels_annotations.go
```diff
@@ -40,4 +40,8 @@ const (
 	// StorageLocationLabel is the label key used to identify the storage
 	// location of a backup.
 	StorageLocationLabel = "ark.heptio.com/storage-location"
+
+	// ForNamespaceLabel is the label key used to identify which namespace
+	// a restic repository stores pod volume backups for.
+	ForNamespaceLabel = "ark.heptio.com/for-namespace"
```
ncdc: This feels a bit too generic - can we include something that indicates this is for restic? Also, can we remove "for"?

pkg/apis/ark/v1/restic_repository.go
```diff
@@ -22,12 +22,20 @@ import (
 
 // ResticRepositorySpec is the specification for a ResticRepository.
 type ResticRepositorySpec struct {
-	// MaintenanceFrequency is how often maintenance should be run.
-	MaintenanceFrequency metav1.Duration `json:"maintenanceFrequency"`
+	// ForNamespace is the namespace this restic repository contains
+	// pod volume backups for.
+	ForNamespace string `json:"forNamespace"`
```
ncdc: Can we call this Namespace instead?

skriss: I don't like `ForNamespace` or `Namespace` :) Maybe `PodNamespace`, `BackupNamespace`, something like that?

ncdc: What are some of the reasons you dislike Namespace?

skriss: it's just very generic.  namespace of what?

skriss: `TargetNamespace`? idk. naming is hard.

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -213,6 +219,28 @@ func (c *podVolumeBackupController) processBackup(req *arkv1api.PodVolumeBackup)
 		req.Spec.Tags,
 	)
 
+	// if this is azure, set resticCmd.Cmd().Env to be os.Environ() plus azure.GetResticEnvVars
```
wwitzel3: How would you feel about extracting this to it's own method? I think this method is already just a procedural list of things to do and check for errors, and having the specific azure logic here makes the natural flow or reading this feel interrupted.

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -213,6 +219,28 @@ func (c *podVolumeBackupController) processBackup(req *arkv1api.PodVolumeBackup)
...
+		if err != nil {
+			return c.fail(req, errors.Wrap(err, "error getting azure restic env vars").Error(), log)
+		}
+
+		env = os.Environ()
+		for k, v := range azureVars {
+			env = append(env, fmt.Sprintf("%s=%s", k, v))
+		}
+
+		resticCmd.Env = env
```
wwitzel3: I notice you set Env here, but also pass env in to GetSnapshotID which, sets Env. I feel like we can just remove this and let the call to GetSnapshotID do the work.

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -312,6 +318,27 @@ func restorePodVolume(req *arkv1api.PodVolumeRestore, credsFile, volumeDir strin
...
+		for k, v := range azureVars {
+			env = append(env, fmt.Sprintf("%s=%s", k, v))
+		}
+
+		resticCmd.Env = env
+	}
+
 	var stdout, stderr string
 
 	if stdout, stderr, err = arkexec.RunCommand(resticCmd.Cmd()); err != nil {
```
wwitzel3: do we want to make `restic.Cmd` take an env like `GetSnapshotID` ?

pkg/restic/repository_ensurer.go
```diff
@@ -75,30 +77,47 @@ func newRepositoryEnsurer(repoInformer arkv1informers.ResticRepositoryInformer,
...
+		if repos[0].Status.Phase != arkv1api.ResticRepositoryPhaseReady {
 			return nil, errors.New("restic repository is not ready")
 		}
-		return repo, nil
+		return repos[0], nil
 	}
 
-	// if we're here, it means we got an IsNotFound error, meaning we need to create a new
-	// repo and wait for it to be ready
+	// if we're here, it means we didn't find a repo, meaning we need to create a new
```
wwitzel3: Bikeshed: No repo found, create one and wait for it to become ready.

pkg/restic/repository_manager.go
```diff
@@ -210,6 +219,27 @@ func (rm *repositoryManager) exec(cmd *Command) error {
 
 	cmd.PasswordFile = file
 
+	if strings.HasPrefix(cmd.RepoIdentifier, "azure") {
```
wwitzel3: This code seems to be in enough places to warrant extracting it out? Maybe to a `provider_environ.go` that will hold Azure and any other providers that end up needing special cases for restic? Maybe not *shrug* I go either way.

pkg/controller/pod_volume_restore_controller.go
wwitzel3: I guess that's silly since restic.Env is what you are setting in this case.

pkg/controller/pod_volume_backup_controller.go
wwitzel3: I see now, this is needed for both the `GetSnapshotID` AND the later call to Cmd.

pkg/restic/repository_manager.go
skriss: good idea - extracting what I can.  I *think* over time this code will go away, because Azure is adding support for RBAC-based access control to storage (more inline with the other public cloud providers).

skriss:
@wwitzel3 PTAL at the latest commit

skriss:
I'm gonna squash before merge.

skriss:
@carlisia agree on unit test coverage -- it's quite low for all things restic.  I'm willing to put time in to improve that, but I'd probably do it as a separate PR since I'd like to tackle restic holistically, not just this particular tweak to it - WDYT?

carlisia:
@skriss this is a perfectly good, actually, even better plan!

carlisia:
Is this ready to merge or needs more squashing?

skriss:
good as is - ready to merge!

PR #891, opened 2018-09-30
remove restore log helper for accurate line #'s

### LINKED ISSUE #588: logSource field in restore logs is always pkg/restore/restore.go:292
Because we use a helper function (`ctx.infof()`) for restore logs, the `logSource` field has the same line number for almost every log line in a restore log. Once we do #158, this should no longer be an issue.

skriss:

Fixes #588 

Still need to manual-test.


skriss:
test looks good - actual line #'s coming thru!

PR #892, opened 2018-10-01
add default-volume-snapshot-locations to server cmd

### LINKED ISSUE #864: Add --default-volume-snapshot-locations flag to `ark server` command

wwitzel3:
Closes #864 

### DIFF
--- a/docs/cli-reference/ark_server.md
+++ b/docs/cli-reference/ark_server.md
@@ -13,15 +13,16 @@ ark server [flags]
 ### Options
 
 ```
-      --backup-sync-period duration              how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster (default 1h0m0s)
-      --default-backup-storage-location string   name of the default backup storage location (default "default")
-  -h, --help                                     help for server
-      --log-level                                the level at which to log. Valid values are debug, info, warning, error, fatal, panic. (default info)
-      --metrics-address string                   the address to expose prometheus metrics (default ":8085")
-      --plugin-dir string                        directory containing Ark plugins (default "/plugins")
-      --restic-timeout duration                  how long backups/restores of pod volumes should be allowed to run before timing out (default 1h0m0s)
-      --restore-only                             run in a mode where only restores are allowed; backups, schedules, and garbage-collection are all disabled
-      --restore-resource-priorities strings      desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources (default [namespaces,persistentvolumes,persistentvolumeclaims,secrets,configmaps,serviceaccounts,limitranges,pods])
+      --backup-sync-period duration                         how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster (default 1h0m0s)
+      --default-backup-storage-location string              name of the default backup storage location (default "default")
+      --default-volume-snapshot-locations mapStringString   list of unique volume providers and default volume snapshot location (provider1:location-01, provider2:location-02, ...)
+  -h, --help                                                help for server
+      --log-level                                           the level at which to log. Valid values are debug, info, warning, error, fatal, panic. (default info)
+      --metrics-address string                              the address to expose prometheus metrics (default ":8085")
+      --plugin-dir string                                   directory containing Ark plugins (default "/plugins")
+      --restic-timeout duration                             how long backups/restores of pod volumes should be allowed to run before timing out (default 1h0m0s)
+      --restore-only                                        run in a mode where only restores are allowed; backups, schedules, and garbage-collection are all disabled
+      --restore-resource-priorities strings                 desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources (default [namespaces,persistentvolumes,persistentvolumeclaims,secrets,configmaps,serviceaccounts,limitranges,pods])
 ```
 
 ### Options inherited from parent commands
--- a/docs/cli-reference/ark_server.md
+++ b/docs/cli-reference/ark_server.md
@@ -15,7 +15,7 @@ ark server [flags]
 ```
       --backup-sync-period duration                         how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster (default 1h0m0s)
       --default-backup-storage-location string              name of the default backup storage location (default "default")
-      --default-volume-snapshot-locations mapStringString   list of unique volume providers and default volume snapshot location (provider1:location-01, provider2:location-02, ...)
+      --default-volume-snapshot-locations mapStringString   list of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...)
   -h, --help                                                help for server
       --log-level                                           the level at which to log. Valid values are debug, info, warning, error, fatal, panic. (default info)
       --metrics-address string                              the address to expose prometheus metrics (default ":8085")
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -87,8 +87,9 @@ type serverConfig struct {
 
 func NewCommand() *cobra.Command {
 	var (
-		logLevelFlag = logging.LogLevelFlag(logrus.InfoLevel)
-		config       = serverConfig{
+		volumeSnapshotLocations = flag.NewMap().WithKeyValueDelimiter(":")
+		logLevelFlag            = logging.LogLevelFlag(logrus.InfoLevel)
+		config                  = serverConfig{
 			pluginDir:                      "/plugins",
 			metricsAddress:                 defaultMetricsAddress,
 			defaultBackupLocation:          "default",
@@ -131,6 +132,10 @@ func NewCommand() *cobra.Command {
 			}
 			namespace := getServerNamespace(namespaceFlag)
 
+			if volumeSnapshotLocations.Data() != nil {
+				config.defaultVolumeSnapshotLocations = volumeSnapshotLocations.Data()
+			}
+
 			s, err := newServer(namespace, fmt.Sprintf("%s-%s", c.Parent().Name(), c.Name()), config, logger)
 			cmd.CheckError(err)
 
@@ -146,12 +151,7 @@ func NewCommand() *cobra.Command {
 	command.Flags().BoolVar(&config.restoreOnly, "restore-only", config.restoreOnly, "run in a mode where only restores are allowed; backups, schedules, and garbage-collection are all disabled")
 	command.Flags().StringSliceVar(&config.restoreResourcePriorities, "restore-resource-priorities", config.restoreResourcePriorities, "desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources")
 	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "name of the default backup storage location")
-
-	volumeSnapshotLocations := flag.NewMap().WithKeyValueDelimiter(":")
-	command.Flags().Var(&volumeSnapshotLocations, "default-volume-snapshot-locations", "list of unique volume providers and default volume snapshot location (provider1:location-01, provider2:location-02, ...)")
-	if volumeSnapshotLocations.Data() != nil {
-		config.defaultVolumeSnapshotLocations = volumeSnapshotLocations.Data()
-	}
+	command.Flags().Var(&volumeSnapshotLocations, "default-volume-snapshot-locations", "list of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...)")
 
 	return command
 }
@@ -283,7 +283,7 @@ func (s *server) run() error {
 		return errors.WithStack(err)
 	}
 
-	defaultVolumeSnapshotLocations, err := s.getDefaultVolumeSnapshotLocations()
+	defaultVolumeSnapshotLocations, err := getDefaultVolumeSnapshotLocations(s.arkClient, s.namespace, s.config.defaultVolumeSnapshotLocations)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


pkg/cmd/server/server.go
```diff
@@ -543,7 +558,12 @@ func (s *server) initRestic(location *api.BackupStorageLocation) error {
 	return nil
 }
 
-func (s *server) runControllers(config *api.Config, defaultBackupLocation *api.BackupStorageLocation) error {
+func (s *server) runControllers(
+	config *api.Config,
+	defaultBackupLocation *api.BackupStorageLocation,
+	defaultVolumeSnapshotLocations map[string]string,
```
wwitzel3: I noticed that in another PR we have the controller expecting a `map[string]string` but since we are doing CLI validation anyway, at this point, I can pass it a `[]*api.VolumeSnapshotLocation`

pkg/cmd/server/server.go
```diff
@@ -606,6 +626,7 @@ func (s *server) runControllers(config *api.Config, defaultBackupLocation *api.B
 			backupTracker,
 			s.sharedInformerFactory.Ark().V1().BackupStorageLocations(),
 			s.config.defaultBackupLocation,
+			//s.config.defaultVolumeSnapshotLocations,
```
wwitzel3: See #873

pkg/cmd/server/server.go
```diff
@@ -144,6 +147,11 @@ func NewCommand() *cobra.Command {
 	command.Flags().StringSliceVar(&config.restoreResourcePriorities, "restore-resource-priorities", config.restoreResourcePriorities, "desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources")
 	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "name of the default backup storage location")
 
+	volumeSnapshotLocations := flag.NewMap()
```
wwitzel3: This was the "best" way I could think to do this, since the default value here is really just `nil` or empty.

pkg/cmd/server/server.go
```diff
@@ -144,6 +147,14 @@ func NewCommand() *cobra.Command {
...
+	volumeSnapshotLocations = volumeSnapshotLocations.WithKeyValueDelimiter(":")
+	command.Flags().Var(&volumeSnapshotLocations, "default-volume-snapshot-locations", "provider map of the default volume snapshot locations")
```
skriss: probably worth adding an example of the format (`provider-1:loc-1,provider-2:loc-2,...`)

pkg/cmd/server/server.go
```diff
@@ -144,6 +147,14 @@ func NewCommand() *cobra.Command {
 	command.Flags().StringSliceVar(&config.restoreResourcePriorities, "restore-resource-priorities", config.restoreResourcePriorities, "desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources")
 	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "name of the default backup storage location")
 
+	volumeSnapshotLocations := flag.NewMap()
+	volumeSnapshotLocations = volumeSnapshotLocations.WithKeyValueDelimiter(":")
```
skriss: you can chain `.WithKeyValueDelimiter(..)` onto the previous line

pkg/cmd/server/server.go
skriss: "map of each volume provider's default volume snapshot location"

pkg/cmd/server/server.go
skriss: How about a `map[string]*api.VolumeSnapshotLocation`, where keys are provider names?  I think that's probably the most useful format to have it in.  cc @carlisia for input

pkg/cmd/server/server.go
```diff
@@ -274,6 +285,13 @@ func (s *server) run() error {
 		return errors.WithStack(err)
 	}
 
+	for _, vslName := range s.config.defaultVolumeSnapshotLocations {
```
skriss: We also want to validate that every provider has a default VSL specified.  So you probably want to `s.ArkClient.ArkV1().VolumeSnapshotLocations(s.namespace).List(...)`, iterate through it to get a unique list of providers (`volumeSnapshotLocation.Spec.Provider`), then make sure that each of them had a value provided in the map, and that the value corresponds to a VSL **for that provider**.

pkg/cmd/server/server.go
carlisia: Thinking...

carlisia: Having a `[]*api.VolumeSnapshotLocation` will work perfectly, so let's go with that.

pkg/cmd/server/server.go
carlisia: I always try to avoid using name of a data structure, even if it totally fits. Could we say instead: `list of unique volume providers and corresponding default volume snapshot location`?

pkg/cmd/server/server.go
```diff
@@ -291,13 +305,50 @@ func (s *server) run() error {
...
+
+		for _, location := range locations {
+			if location.ObjectMeta.Name == defaultLocation {
+				providerDefaults[provider] = location
+				break
+			}
+		}
+
+		if _, ok := providerDefaults[provider]; !ok {
+			return providerDefaults, errors.Errorf("missing provider %s. You must provide a default for every provider.", provider)
```
skriss: this error should say that the default VSL specified for the provider isn't valid for that provider, and include both the provider & VSL name

pkg/cmd/server/server.go
```diff
@@ -291,13 +305,50 @@ func (s *server) run() error {
...
 
-	if err := s.runControllers(config, backupStorageLocation); err != nil {
+	if err := s.runControllers(config, backupStorageLocation, defaultVolumeSnapshotLocations); err != nil {
 		return err
 	}
 
 	return nil
 }
 
+func (s *server) getDefaultVolumeSnapshotLocations() (map[string]*api.VolumeSnapshotLocation, error) {
```
skriss: you should be able to add a unit test for this - create a server struct with a fake ark clientset, use the clientset to create some VSLs, set s.config.defaultVolumeSnapshotLocations, then run this func

pkg/cmd/server/server.go
```diff
@@ -80,19 +81,21 @@ type serverConfig struct {
 	pluginDir, metricsAddress, defaultBackupLocation string
 	backupSyncPeriod, podVolumeOperationTimeout      time.Duration
 	restoreResourcePriorities                        []string
+	defaultVolumeSnapshotLocations                   map[string]string
```
carlisia: It'd be better if this were to be a slice. I'll change the controller.

pkg/cmd/server/server.go
```diff
@@ -291,13 +305,50 @@ func (s *server) run() error {
...
+
+	providerLocations := make(map[string][]*api.VolumeSnapshotLocation)
+	for _, vsl := range volumeSnapshotLocations.Items {
+		providerLocations[vsl.Spec.Provider] = append(providerLocations[vsl.Spec.Provider], &vsl)
+	}
+
+	for provider, locations := range providerLocations {
+		defaultLocation, ok := s.config.defaultVolumeSnapshotLocations[provider]
+		if !ok {
+			return providerDefaults, errors.Errorf("missing provider %s. You must provide a default for every provider.", provider)
```
carlisia: I think we want to say something like `for every existing provider.`? Not sure this is the proper language but to indicate that it is not *all* providers, only those that have been specified.

docs/cli-reference/ark_server.md
```diff
@@ -13,15 +13,16 @@ ark server [flags]
...
-      --plugin-dir string                        directory containing Ark plugins (default "/plugins")
-      --restic-timeout duration                  how long backups/restores of pod volumes should be allowed to run before timing out (default 1h0m0s)
-      --restore-only                             run in a mode where only restores are allowed; backups, schedules, and garbage-collection are all disabled
-      --restore-resource-priorities strings      desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources (default [namespaces,persistentvolumes,persistentvolumeclaims,secrets,configmaps,serviceaccounts,limitranges,pods])
+      --backup-sync-period duration                         how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster (default 1h0m0s)
+      --default-backup-storage-location string              name of the default backup storage location (default "default")
+      --default-volume-snapshot-locations mapStringString   list of unique volume providers and default volume snapshot location (provider1:location-01, provider2:location-02, ...)
```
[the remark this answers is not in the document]
wwitzel3: I think taking advantage of the helper we have makes this process way easier. We control what that value represents in `flag/map.go:88` maybe we can agree to change it there at some point.

pkg/cmd/server/server.go
wwitzel3: I figured we were passing `map[string]*api.VolumeSnapshotLocation` to the controller (see: https://github.com/heptio/ark/pull/892/files#diff-cd1a009fa409db8496ea35b53f9776d2R600)

pkg/cmd/server/server.go
wwitzel3: oops, good catch, copy pasta I forgot to update

pkg/cmd/server/server.go
carlisia: No, if it was a slice, it should be a slice of vsl objects, which could be built at the time you check if they exist. But very good point tho, I wasn't thinking of the cli validation. I was mostly thinking of the variable that gets sent to the controller. There I only need a list of provider/locationName. This could be a map or a slice of vsl objects w/ unique providers. Pls nvm what I said and reuse this maps if it's simpler.

carlisia: Now that I think about it, a map with provider/name would be way better than passing the objects. Yep, nvm what I said!

wwitzel3: I see, yeah, if we can to just send in a `[]*api.VolumeSnapshotLocation` for the controllers, I think that would be fine, since you end up unpacking the VSL and checking the `Spec.Provider` anyway.

wwitzel3: I'm happy to go either direction as far as that goes. 👍

carlisia: I'm thinking building a list of entire objects just to pluck a couple fields out of it is not worth it, especially if you already have the map. On the other hand, the vsl is a small obj and the []vsl is much more explicit than the map. I can go either way too 👍

pkg/cmd/server/server.go
```diff
@@ -144,6 +147,12 @@ func NewCommand() *cobra.Command {
 	command.Flags().StringSliceVar(&config.restoreResourcePriorities, "restore-resource-priorities", config.restoreResourcePriorities, "desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources")
 	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "name of the default backup storage location")
 
+	volumeSnapshotLocations := flag.NewMap().WithKeyValueDelimiter(":")
+	command.Flags().Var(&volumeSnapshotLocations, "default-volume-snapshot-locations", "list of unique volume providers and default volume snapshot location (provider1:location-01, provider2:location-02, ...)")
+	if volumeSnapshotLocations.Data() != nil {
```
skriss: you'll need to move this up to ~L130 since this section of code here is just setting up the flags.

pkg/cmd/server/server_test.go
```diff
@@ -97,3 +98,39 @@ func TestArkResourcesExist(t *testing.T) {
...
+		Spec:       v1.VolumeSnapshotLocationSpec{Provider: "provider1"},
+	}
+	arkClient.ArkV1().VolumeSnapshotLocations(namespace).Create(location)
+
+	defaultVolumeSnapshotLocations := make(map[string]string)
+
+	// No defaults
+	volumeSnapshotLocations, err := getDefaultVolumeSnapshotLocations(arkClient, namespace, defaultVolumeSnapshotLocations)
+	assert.Equal(t, len(volumeSnapshotLocations), 0)
+	assert.NoError(t, err)
```
skriss: shouldn't this case error per L334?

pkg/cmd/server/server_test.go
```diff
@@ -97,3 +98,39 @@ func TestArkResourcesExist(t *testing.T) {
 	arkAPIResourceList.APIResources = arkAPIResourceList.APIResources[:3]
 	assert.Error(t, server.arkResourcesExist())
 }
+
+func TestDefaultVolumeSnapshotLocations(t *testing.T) {
+	namespace := "heptio-ark"
+	arkClient := fakeclientset.NewSimpleClientset()
+	location := &v1.VolumeSnapshotLocation{
```
skriss: would be nice to add a couple more cases to cover multiple providers/multiple VSLs per provider :)

pkg/cmd/server/server.go
```diff
@@ -144,6 +147,12 @@ func NewCommand() *cobra.Command {
 	command.Flags().StringSliceVar(&config.restoreResourcePriorities, "restore-resource-priorities", config.restoreResourcePriorities, "desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources")
 	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "name of the default backup storage location")
 
+	volumeSnapshotLocations := flag.NewMap().WithKeyValueDelimiter(":")
+	command.Flags().Var(&volumeSnapshotLocations, "default-volume-snapshot-locations", "list of unique volume providers and default volume snapshot location (provider1:location-01, provider2:location-02, ...)")
```
skriss: let's remove the spaces in the example just after the commas (so `provider1:location-01, provider2:location-02,...`)

pkg/cmd/server/server_test.go
wwitzel3: No, line 317 (https://github.com/heptio/ark/pull/892/files#diff-cd1a009fa409db8496ea35b53f9776d2R317) makes this a NOOP. The flag parsing library itself enforces that an argument is actually passed in:

```An error occurred: flag needs an argument: --default-volume-snapshot-locations```

So for the sake of unit testing, this is the behavior I would expect.

skriss: Ah, makes sense.

skriss:
LGTM - squash away

skriss:
ah, needs a `make update`.  Once I rebase this branch we'll never have to do that again

PR #895, opened 2018-10-02
Update backups for volume snapshot locations

skriss:
This PR is completely still a work in progress, but wanted to get folks' input, particularly on the second commit..

I introduced a new struct, `backup.Request`, that embeds an `*arkv1api.Backup` and also has fields for holding an `*arkv1api.BackupStorageLocation`, `[]*arkv1api.VolumeSnapshotLocation`, and other objects related to a backup that are needed to actually execute it.  The backup controller instantiates it, adds some of the fields, then the Backupper in `pkg/backup` accepts it and populates some additional fields before passing it down to the other structs involved in backups.  The main goal with doing this was to avoid continued expansion of the parameter list across all the types in `pkg/backup` - I think keeping these things in a single struct is more reasonable.  What do you all think?  Good path to continue on?

### DIFF
--- a/pkg/backup/group_backupper.go
+++ b/pkg/backup/group_backupper.go
@@ -27,31 +27,25 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	kuberrs "k8s.io/apimachinery/pkg/util/errors"
 
-	"github.com/heptio/ark/pkg/apis/ark/v1"
 	"github.com/heptio/ark/pkg/client"
-	"github.com/heptio/ark/pkg/cloudprovider"
 	"github.com/heptio/ark/pkg/discovery"
 	"github.com/heptio/ark/pkg/podexec"
 	"github.com/heptio/ark/pkg/restic"
-	"github.com/heptio/ark/pkg/util/collections"
 )
 
 type groupBackupperFactory interface {
 	newGroupBackupper(
 		log logrus.FieldLogger,
-		backup *v1.Backup,
-		namespaces, resources *collections.IncludesExcludes,
+		backupRequest *Request,
 		dynamicFactory client.DynamicFactory,
 		discoveryHelper discovery.Helper,
 		backedUpItems map[itemKey]struct{},
 		cohabitatingResources map[string]*cohabitatingResource,
-		actions []resolvedAction,
 		podCommandExecutor podexec.PodCommandExecutor,
 		tarWriter tarWriter,
-		resourceHooks []resourceHook,
-		blockStore cloudprovider.BlockStore,
 		resticBackupper restic.Backupper,
 		resticSnapshotTracker *pvcSnapshotTracker,
+		blockStoreGetter BlockStoreGetter,
 	) groupBackupper
 }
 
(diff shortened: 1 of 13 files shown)


pkg/backup/item_backupper.go
```diff
@@ -358,17 +343,41 @@ func (ib *defaultItemBackupper) executeActions(
 	return obj, nil
 }
 
+// getBlockStore gets and initializes a BlockStore given a VolumeSnapshotLocation, returning an
+// existing one if one's already been initialized for the location.
+func (ib *defaultItemBackupper) getBlockStore(snapshotLocation *api.VolumeSnapshotLocation) (cloudprovider.BlockStore, error) {
```
carlisia: Could this be named `blockStore` instead of `getBlockStore`?

skriss:
Alright, this PR is ready for a full review from @carlisia and @wwitzel3.

skriss:
will squash before merge

nrb:
I haven't given this a detailed review, but I like the approach of having most of the necessary data in a single struct since we're passing around so much.

skriss:
@wwitzel3 you get a chance to look at this?

pkg/controller/backup_controller.go
```diff
@@ -256,7 +251,7 @@ func patchBackup(original, updated *api.Backup, client arkv1client.BackupsGetter
 	return res, nil
 }
 
-func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackupLocation string) (*api.BackupStorageLocation, []string) {
+func (c *backupController) getLocationAndValidate(itm *pkgbackup.Request, defaultBackupLocation string) []string {
```
wwitzel3: I think these methods might need to be renamed for the updated behavior. By its name, I'm not expecting the side-effects to be applied to the Request.

pkg/controller/backup_controller.go
```diff
@@ -277,23 +272,23 @@ func (c *backupController) getLocationAndValidate(itm *api.Backup, defaultBackup
...
 
 // defaultAndValidateSnapshotLocations ensures:
 // - each location name in Spec VolumeSnapshotLocation exists as a location
 // - exactly 1 location per existing or default provider
 // - a given default provider's location name is added to the Spec VolumeSnapshotLocation if it does not exist as a VSL
-func (c *backupController) defaultAndValidateSnapshotLocations(itm *api.Backup, defaultLocations map[string]string) []string {
-	var errors []string
-	perProviderLocationName := make(map[string]string)
-	var finalLocationNameList []string
+func (c *backupController) defaultAndValidateSnapshotLocations(itm *pkgbackup.Request, defaultLocations map[string]*api.VolumeSnapshotLocation) []string {
```
wwitzel3: Maybe rename so that it is clear we are calling this for the side effects that get applied to itm?

pkg/backup/item_backupper.go
```diff
@@ -105,21 +95,18 @@ type ItemBackupper interface {
 }
 
 type defaultItemBackupper struct {
-	backup                *api.Backup
-	namespaces            *collections.IncludesExcludes
-	resources             *collections.IncludesExcludes
+	backup                *Request
```
wwitzel3: If we are creating a new struct to hold a backup + the context for creation, I think it would be good to distinguish that here. This is a bit overloaded and I think it would be better as backupRequest or request.

pkg/backup/group_backupper.go
```diff
@@ -99,20 +87,17 @@ type groupBackupper interface {
 
 type defaultGroupBackupper struct {
 	log                      logrus.FieldLogger
-	backup                   *v1.Backup
-	namespaces, resources    *collections.IncludesExcludes
+	backup                   *Request
```
wwitzel3: Same as itemBackupper, maybe just request or backupRequest?

skriss:
@wwitzel3 check out the latest 3 commits, I think they address your comments

skriss:
gonna do some squashing post-LGTY

wwitzel3:
LGTM, happy to merge when ready.

skriss:
alright, whenever it's green

PR #899, opened 2018-10-03
 Update default-volume-snapshot-locations flag to use new pflag primitive

carlisia:
There was a new release of `spf13/pflag` that contains a primitive to handle a flag with map-like input. I thought it'd be good to upgrade to that.


skriss:
If we go this route I would want to switch all of our uses of `"github.com/heptio/ark/pkg/cmd/util/flag".Map`.  One thing we lose if we switch is the ability to customize the delimiters -- pflag requires `key1=val1,key2=val2`, whereas we can support e.g. `key1:val1;key2:val2`.  We did that intentionally, particularly for restore namespace remappings where we felt the `=` was a weird UX.

What do others think?  One option would be to contribute a PR to pflag that allows delimiter customization, then switch to it once it's in.

nrb:
Apparently my comment on this last week was lost.

I like the idea of using the same thing for all of our code, and keeping the ability to use our custom delimiter, so I support the idea of contributing it to pflag if they want it.

carlisia:
I second what @nrb said. I'll open an issue and ask them if this is possible.

carlisia:
FYI: https://github.com/spf13/pflag/issues/183

skriss:
Moving this to icebox for now until we hear back from the pflag folks

PR #902, opened 2018-10-04
Fix broken storage providers link

scottslowe:
This PR fixes a broken link in `README.md` for the list of compatible storage providers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,4 +71,4 @@ See [the list of releases][6] to find out about feature changes.
 [29]: https://heptio.github.io/ark/
 [30]: /docs/troubleshooting.md
 
-[99]: support-matrix.md
+[99]: /docs/support-matrix.md


skriss:
@scottslowe thanks for the PR.

@nrb @Bradamant3 what is the preferred way to refer to other pages within the ark repo? Looking at `README.md`, we have some absolute links (using `/blob/master`) and some relative links like @scottslowe has updated here.  IIRC, how we do this has an impact on the GH pages site generator?

nrb:
Yeah, I think I agree with Wayne. Also, I think the README can be treated a little differently than the docs site. The site should be self-contained within the docs dir, whereas the README is likely hit from GH or by cloning the repo.

Bradamant3:
Scott's approach works -- the docs-site-gen tool strips the /docs/ prefix for the docs site. But things will change a bit when we move to a full-on SSG build (which I'll try to get to within the next week or so)

Issue #909, opened 2018-10-04
More Sam Account Token issues when sam account and pod are collectively deleted and restored


james-powis:
Service Account Tokens are ephemeral, and stateless, created when a service account is created. Mounting of the service-account-token into volumes as part of the init stage of restic restores results in the init pod failing to start, thus failing the restic portion of the restore (though the pod is restored none the less to an empty functional status.

```
   default:  error restoring /tmp/293264120/resources/pods/namespaces/default/test-jenkins-99fbbdb95-8rms7.json: Pod "test-jenkins-99fbbdb95-8rms7" is invalid: spec.initContainers[1].volumeMounts[4].name: Not found: "test-jenkins-token-mbgcr"
```

I will get a PR to address this issue.

#840 had similar symptoms

nrb:
PR's merged but apparently GitHub didn't pick it up. Closing the issue.

PR #910, opened 2018-10-04
Drop volumes matching name ServiceAccountName-token-

james-powis:
Continuation of #843 and #840 

Addresses #909 

There does not appear to be a test file for the restic_restore_action.go file so this is one of those "this should not break anything, I hope..." I just don't know enough about this project (or go testing in general) to create a test file from scratch. 

To that end it compiles, and in go that is a good sign.

### DIFF
--- a/pkg/restore/pod_action.go
+++ b/pkg/restore/pod_action.go
@@ -112,5 +112,36 @@ func (a *podAction) Execute(obj runtime.Unstructured, restore *api.Restore) (run
 		return nil, nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


james-powis:
@skriss You are right about that... And so it is, ready for your review, should be pretty easy it is nearly an identical duplication of the container volumeMount function and test.

james-powis:
Also can I request a release after this is merged, even if it is a pre/beta release, I have a restore I need to perform....

PR #922, opened 2018-10-10
[WIP] switch to using VolumeSnapshots for backup, sync

### LINKED ISSUE #863: Switch to storing information about PV snapshots as VolumeSnapshot API objects
- during backup, instead of storing snapshot info in `backup.status.volumeBackupInfo`, create a new VolumeSnapshot API object per snapshot
- upload a file containing all VolumeSnapshots' JSON for a backup into the backup's directory in object storage
- think about how this scales with large clusters/large # PVs - will we have too many API objects? should these only exist as files in S3?

skriss:

Fixes #863 

@wwitzel3 @carlisia before I continue / fix units, would appreciate an initial review on this PR and whether the approach I took seems reasonable.


pkg/backup/item_backupper.go
```diff
@@ -454,30 +456,62 @@ func (ib *defaultItemBackupper) takePVSnapshot(obj runtime.Unstructured, log log
...
-		Iops:             iops,
-		AvailabilityZone: pvFailureDomainZone,
+	createdSnap, err := ib.volumeSnapshotClient.VolumeSnapshots(snapshot.Namespace).Create(snapshot)
+	if err == nil {
+		ib.backupRequest.VolumeSnapshots = append(ib.backupRequest.VolumeSnapshots, createdSnap)
 	}
 
-	return nil
+	// nil errors are automatically removed
+	return kubeerrs.NewAggregate([]error{
```
carlisia: I would do this a bit differently. On ln 479, `if err != nil`, then return this aggregate error. Return `nil` at the end.

pkg/backup/item_backupper.go
```diff
@@ -454,30 +456,62 @@ func (ib *defaultItemBackupper) takePVSnapshot(obj runtime.Unstructured, log log
...
 		log.WithError(err).Error("error getting volume info")
 		return errors.WithMessage(err, "error getting volume info")
 	}
 
-	if ib.backupRequest.Status.VolumeBackups == nil {
-		ib.backupRequest.Status.VolumeBackups = make(map[string]*api.VolumeBackupInfo)
+	log.Info("Snapshotting PersistentVolume")
+	snapshot := newVolumeSnapshot(ib.backupRequest.Backup, volumeType, pvFailureDomainZone, location, iops)
+
+	snapshotID, createSnapErr := blockStore.CreateSnapshot(volumeID, pvFailureDomainZone, tags)
```
carlisia: Maybe it's a naming issue but what is the relation between creating a snapshot here, and the snapshot on ln 467?

carlisia: I looked further down and I see now what the code is doing, makes sense. Because the `snapshot` on ln 467 is not a "newly created" snapshot, is just an object that was setup, I'd rename that function `newVolumeSnapshot` to `volumeSnapshot` if it doesn't create a conflict.

I know `CreateSnapshot` is an existing method, but it seems confusing. It seems to create a snapshot, but there's also a create snapshot call on ln 478. What is throwing me off here as well is that nothing from the `snapshot` object from ln 467 is being used to make the call on ln 469.

pkg/controller/backup_controller.go
```diff
@@ -385,51 +385,75 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
...
+		volumeSnapshotList.Items = append(volumeSnapshotList.Items, *vs)
+	}
 
-	backupDuration := backup.Status.CompletionTimestamp.Time.Sub(backup.Status.StartTimestamp.Time)
-	backupDurationSeconds := float64(backupDuration / time.Second)
-	c.metrics.RegisterBackupDuration(backupScheduleName, backupDurationSeconds)
+	volumeSnapshotJSON := new(bytes.Buffer)
+	if err := encode.EncodeTo(volumeSnapshotList, "json", volumeSnapshotJSON); err != nil {
+		errs = append(errs, errors.Wrap(err, "error encoding list of volume snapshots"))
+		// TODO what do we want to do if we fail to encode the VolumeSnapshots list to JSON?
```
carlisia: One answer is log the reason why for our internal consumption and attempt to prevent it in the future if there's a way.  We could also maybe create a json with an error message.

skriss:
Updated this PR to store volume snapshot info as JSON in a file in object storage only.  If we agree on this approach, I still need to:

- [ ] remove VolumeSnapshot CRD and all associated generated code
- [ ] update `ark backup describe` to fetch and display the volume snapshot data

skriss:
Generated a 10,000-snapshot JSON file using representative data from Azure (which tends to have long IDs) - ~5MB uncompressed, 500k gzipped.

skriss:
Actually I'm going to tackle the `ark backup describe` changes in a separate PR.  So this should be ready for a full review.  Should be helpful to get this in sooner for the 2 remaining controller issues.

As usual I'll squash before merge.

carlisia:
Looking..

PR #923, opened 2018-10-10
add --volume-snapshot-locations flag to ark backup create

skriss:

### DIFF
--- a/pkg/cmd/cli/backup/create.go
+++ b/pkg/cmd/cli/backup/create.go
@@ -70,6 +70,7 @@ type CreateOptions struct {
 	IncludeClusterResources flag.OptionalBool
 	Wait                    bool
 	StorageLocation         string
+	SnapshotLocations       []string
 
 	client arkclient.Interface
 }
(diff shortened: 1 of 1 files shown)


pkg/cmd/cli/backup/create.go
```diff
@@ -92,6 +93,7 @@ func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.Var(&o.ExcludeResources, "exclude-resources", "resources to exclude from the backup, formatted as resource.group, such as storageclasses.storage.k8s.io")
 	flags.Var(&o.Labels, "labels", "labels to apply to the backup")
 	flags.StringVar(&o.StorageLocation, "storage-location", "", "location in which to store the backup")
+	flags.StringSliceVar(&o.SnapshotLocations, "volume-snapshot-locations", o.SnapshotLocations, "list of locations (at most one per provider) where volume snapshots should be stored")
```
carlisia: Wonder if it'd sound better this way: `list of each location per provider where...`.

skriss: Personally I think what I have is clearer, let's see what @wwitzel3 thinks

wwitzel3: I prefer the current `list of locations ...`

PR #930, opened 2018-10-11
don't require a default backup storage location to exist

### LINKED ISSUE #898: Enable to start ark-server without defined default BackupStorageLocation

In next ark version (v0.10) the default BackupStorageLocation is required (without setting it I get `An error occurred: backupstoragelocations.ark.heptio.com "default" not found` and ark-server crashes). 
Is it possible to make it optional? 
My use-case is that during ark installation details about future backups (where to store them) may not be available.

Slack discussion:
Ref: https://kubernetes.slack.com/archives/C6VCGP4MT/p1538489856000100

**Environment:**

- Ark version (use `ark version`): master tag

skriss:

Fixes #898 

I think this is the most reasonable approach to addressing this.  The server flag retains a default value of "default", but the server won't terminate if it can't find a BackupStorageLocation custom resource named "default".  Backups will fail validation if they don't have an explicitly-specified BSL that actually exists.  Later, if a BSL named "default" (or whatever the value of the server flag is) is created, it will automatically start being used by default for backups.

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -272,9 +272,9 @@ func (s *server) run() error {
 
 	s.watchConfig(originalConfig)
 
-	backupStorageLocation, err := s.arkClient.ArkV1().BackupStorageLocations(s.namespace).Get(s.config.defaultBackupLocation, metav1.GetOptions{})
-	if err != nil {
-		return errors.WithStack(err)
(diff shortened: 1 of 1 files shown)


skriss:
@carlisia @wwitzel3 PTAL when you get a chance, thanks.

PR #937, opened 2018-10-15
Azure: allow snapshots to be stored in alternate resource group

skriss:
This PR allows Azure `VolumeSnapshotLocations` to specify a `resourceGroup` that's different than the one where the cluster is, and to have snapshots created in that resource group.


skriss:
@nrb adding you to review this now that you're back :)

skriss:
@nrb would be awesome if you could provide comments or merge soon, thanks!

PR #938, opened 2018-10-16
update restore to use correct volume snapshot locations.

### LINKED ISSUE #808: Update restore process to use the appropriate locations (snapshot, destination for volume create) when doing PV restores
Ark currently assumes there's only one location for snapshots and creating new disks. The restore process (the controller and `restore` package) will need to be updated to inspect the relevant `Backup` and use the right locations.

- pkg/restore will get a list of the VolumeSnapshot objects that are associated with the backup being restored
- restore will either use backup.spec.volumeSnapshotLocations or the list of VSLs from the previous step to identify the locations where snapshots exist
- restore will instantiate a BlockStore per VSL and use it to execute CreateVolumeFromSnapshot
- TODO: figure out UX around how we figure out where the nodes are/where the volume should be created. Currently, we always restore into the same AZ as the original backed-up volume, but that's not ideal. (also consider how this changes in v0.11 with replication)

wwitzel3:
Use correct snapshot locations when doing restores.

Closes #808

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -728,9 +728,7 @@ func (s *server) runControllers(
 	restorer, err := restore.NewKubernetesRestorer(
 		s.discoveryHelper,
 		client.NewDynamicFactory(s.dynamicClient),
-		nil,
 		s.config.restoreResourcePriorities,
-		s.arkClient.ArkV1(),
 		s.kubeClient.CoreV1().Namespaces(),
 		s.resticManager,
 		s.config.podVolumeOperationTimeout,
@@ -746,6 +744,7 @@ func (s *server) runControllers(
 		restorer,
 		s.sharedInformerFactory.Ark().V1().Backups(),
 		s.sharedInformerFactory.Ark().V1().BackupStorageLocations(),
+		s.sharedInformerFactory.Ark().V1().VolumeSnapshotLocations(),
 		false,
 		s.logger,
 		s.logLevel,
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -68,17 +68,18 @@ var nonRestorableResources = []string{
 type restoreController struct {
 	*genericController
 
-	namespace             string
-	restoreClient         arkv1client.RestoresGetter
-	backupClient          arkv1client.BackupsGetter
-	restorer              restore.Restorer
-	pvProviderExists      bool
-	backupLister          listers.BackupLister
-	restoreLister         listers.RestoreLister
-	backupLocationLister  listers.BackupStorageLocationLister
-	restoreLogLevel       logrus.Level
-	defaultBackupLocation string
-	metrics               *metrics.ServerMetrics
+	namespace              string
+	restoreClient          arkv1client.RestoresGetter
+	backupClient           arkv1client.BackupsGetter
+	restorer               restore.Restorer
+	pvProviderExists       bool
+	backupLister           listers.BackupLister
+	restoreLister          listers.RestoreLister
+	backupLocationLister   listers.BackupStorageLocationLister
+	snapshotLocationLister listers.VolumeSnapshotLocationLister
+	restoreLogLevel        logrus.Level
+	defaultBackupLocation  string
+	metrics                *metrics.ServerMetrics
 
 	newPluginManager func(logger logrus.FieldLogger) plugin.Manager
 	newBackupStore   func(*api.BackupStorageLocation, persistence.ObjectStoreGetter, logrus.FieldLogger) (persistence.BackupStore, error)
@@ -92,6 +93,7 @@ func NewRestoreController(
 	restorer restore.Restorer,
 	backupInformer informers.BackupInformer,
 	backupLocationInformer informers.BackupStorageLocationInformer,
+	snapshotLocationInformer informers.VolumeSnapshotLocationInformer,
 	pvProviderExists bool,
 	logger logrus.FieldLogger,
 	restoreLogLevel logrus.Level,
@@ -100,18 +102,19 @@ func NewRestoreController(
 	metrics *metrics.ServerMetrics,
 ) Interface {
 	c := &restoreController{
-		genericController:     newGenericController("restore", logger),
-		namespace:             namespace,
-		restoreClient:         restoreClient,
-		backupClient:          backupClient,
-		restorer:              restorer,
-		pvProviderExists:      pvProviderExists,
-		backupLister:          backupInformer.Lister(),
-		restoreLister:         restoreInformer.Lister(),
-		backupLocationLister:  backupLocationInformer.Lister(),
-		restoreLogLevel:       restoreLogLevel,
-		defaultBackupLocation: defaultBackupLocation,
-		metrics:               metrics,
+		genericController:      newGenericController("restore", logger),
+		namespace:              namespace,
+		restoreClient:          restoreClient,
+		backupClient:           backupClient,
+		restorer:               restorer,
+		pvProviderExists:       pvProviderExists,
+		backupLister:           backupInformer.Lister(),
+		restoreLister:          restoreInformer.Lister(),
+		backupLocationLister:   backupLocationInformer.Lister(),
+		snapshotLocationLister: snapshotLocationInformer.Lister(),
+		restoreLogLevel:        restoreLogLevel,
+		defaultBackupLocation:  defaultBackupLocation,
+		metrics:                metrics,
 
 		// use variables to refer to these functions so they can be
 		// replaced with fakes for testing.
@@ -124,6 +127,7 @@ func NewRestoreController(
 		backupInformer.Informer().HasSynced,
 		restoreInformer.Informer().HasSynced,
 		backupLocationInformer.Informer().HasSynced,
+		snapshotLocationInformer.Informer().HasSynced,
 	)
 
 	restoreInformer.Informer().AddEventHandler(
@@ -233,6 +237,7 @@ func (c *restoreController) processRestore(key string) error {
 		restore,
 		actions,
 		info,
+		pluginManager,
 	)
 
 	restore.Status.Warnings = len(restoreWarnings.Ark) + len(restoreWarnings.Cluster)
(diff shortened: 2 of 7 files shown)


pkg/controller/restore_controller.go
```diff
@@ -531,10 +531,13 @@ func (c *restoreController) runRestore(
...
 	defer closeAndRemoveFile(resultsFile, c.logger)
 
+	pluginManager := c.newPluginManager(log)
+	defer pluginManager.CleanupClients()
+
 	// Any return statement above this line means a total restore failure
 	// Some failures after this line *may* be a total restore failure
 	log.Info("starting restore")
-	restoreWarnings, restoreErrors = c.restorer.Restore(log, restore, info.backup, backupFile, actions)
+	restoreWarnings, restoreErrors = c.restorer.Restore(log, restore, info.backup, info.backupStore, backupFile, actions, pluginManager)
```
skriss: I'm thinking that instead of passing an `info.backupStore`, the controller should just download the volume snapshot data and pass that, since it's already doing that for the backup tarball itself -- unless there's a reason not to that I'm missing

pkg/restore/restore.go
```diff
@@ -917,44 +924,59 @@ func (r *pvRestorer) executePVAction(obj *unstructured.Unstructured) (*unstructu
...
 	if err != nil {
 		return nil, err
 	}
-	log.Info("successfully restored persistent volume from snapshot")
 
-	updated1, err := r.blockStore.SetVolumeID(obj, volumeID)
-	if err != nil {
-		return nil, err
+	for _, snapshot := range volumeSnapshots {
+		if snapshot.Spec.ProviderVolumeID == pvName {
```
skriss: these two things aren't expected to be equal (ProviderVolumeID is e.g. the AWS EBS volume ID in AWS, while pvName is the name of the PV object in Kubernetes), looks like we probably need to add `PersistentVolumeName` to the `VolumeSnapshot` struct

pkg/restore/restore.go
```diff
@@ -917,44 +930,60 @@ func (r *pvRestorer) executePVAction(obj *unstructured.Unstructured) (*unstructu
...
+				},
+			)
 
-	updated1, err := r.blockStore.SetVolumeID(obj, volumeID)
-	if err != nil {
-		return nil, err
-	}
+			log.Info("successfully restored persistent volume from snapshot")
+
+			// used to be update1 which is then cast to Unstructured and returned
```
wwitzel3: So do we return a list of all the ones we updated and then perform all the resource actions on a range?

pkg/cmd/server/server.go
```diff
@@ -728,9 +728,9 @@ func (s *server) runControllers(
 	restorer, err := restore.NewKubernetesRestorer(
 		s.discoveryHelper,
 		client.NewDynamicFactory(s.dynamicClient),
-		nil,
 		s.config.restoreResourcePriorities,
 		s.arkClient.ArkV1(),
+		s.sharedInformerFactory.Ark().V1().VolumeSnapshotLocations(),
```
skriss: we need to make sure that this has synced before using it.  This is always a little trickier when a non-controller takes a dependency on a lister/informer.  One option would be to have the restore controller wait for it to sync, then pass it as an argument to the `Restore` func.  Or, you could pass it to both the restore controller and the restorer, but have the restore controller wait for it to sync, and have the restorer's `Restore` func just return an error if it hasn't synced yet.  Since logically we'll only call `Restore` once all of the restore controller's informers have synced, we should be good.

pkg/restore/restore.go
```diff
@@ -886,11 +895,15 @@ type PVRestorer interface {
...
-	volumeBackups   map[string]*api.VolumeBackupInfo
-	blockStore      cloudprovider.BlockStore
+	logger                 logrus.FieldLogger
+	backupName             string
+	backupNamespace        string
+	snapshotVolumes        *bool
+	restorePVs             *bool
+	volumeSnapshots        []*volume.Snapshot
+	blockStoreGetter       BlockStoreGetter
+	backupClient           arkv1client.BackupsGetter
```
skriss: I don't think this used?

pkg/restore/restore.go
```diff
@@ -917,44 +930,60 @@ func (r *pvRestorer) executePVAction(obj *unstructured.Unstructured) (*unstructu
...
-	// If we can't find a snapshot record for this particular PV, it most likely wasn't a PV that Ark
-	// could snapshot, so return early instead of trying to restore from a snapshot.
-	backupInfo, found := r.volumeBackups[pvName]
-	if !found {
-		return obj, nil
-	}
+	// TODO: Determine if we check for existence of old backup format here
 
-	// Past this point, we expect to be doing a restore
+	for _, snapshot := range r.volumeSnapshots {
```
skriss: you should find the snapshot whose `PersistentVolumeName` matches `obj.GetName()` -- that's the one we should be restoring.

pkg/restore/restore.go
```diff
@@ -917,44 +930,60 @@ func (r *pvRestorer) executePVAction(obj *unstructured.Unstructured) (*unstructu
...
+			return nil, err
+		}
 
-	log.Info("restoring persistent volume from snapshot")
-	volumeID, err := r.blockStore.CreateVolumeFromSnapshot(backupInfo.SnapshotID, backupInfo.Type, backupInfo.AvailabilityZone, backupInfo.Iops)
-	if err != nil {
-		return nil, err
-	}
-	log.Info("successfully restored persistent volume from snapshot")
+		if _, err := blockStore.GetVolumeID(obj); err != nil {
```
skriss: we should still check this once we've gotten the snapshot & VSL to make sure it returns a value, but if it doesn't, then we should return an error rather than continue, since there's only one possible snapshot we should be restoring.

skriss:
@wwitzel3 would really like to get this done and merged tomorrow so we can get `snapshot-locations` merged, so let me know if you need any further input to get this across the finish line.

pkg/cmd/server/server.go
wwitzel3: Yeah, I think I'll just pass it to the controller, wait for the sync, and pass it in on the `Restore` call.

wwitzel3:
Right now I think the implementation is good, I am just futzing with some testing/mock/fake issues.

pkg/controller/restore_controller.go
```diff
@@ -531,10 +535,21 @@ func (c *restoreController) runRestore(
 	}
 	defer closeAndRemoveFile(resultsFile, c.logger)
 
+	pluginManager := c.newPluginManager(log)
```
skriss: There's already a plugin manager instantiated for the restore at https://github.com/heptio/ark/pull/938/files#diff-7d210f9ced78ff8be1ba497bf49e1a40R199 so I'd pass that into the `runRestore` func and use it here instead of instantiating a new one.

pkg/restore/restore.go
```diff
@@ -917,44 +939,57 @@ func (r *pvRestorer) executePVAction(obj *unstructured.Unstructured) (*unstructu
...
 
-	// If we can't find a snapshot record for this particular PV, it most likely wasn't a PV that Ark
-	// could snapshot, so return early instead of trying to restore from a snapshot.
-	backupInfo, found := r.volumeBackups[pvName]
-	if !found {
-		return obj, nil
-	}
-
-	// Past this point, we expect to be doing a restore
+	for _, snapshot := range r.volumeSnapshots {
```
skriss: let's invert this loop to remove nesting and handle the error case first - so just loop through `r.volumeSnapshots` looking for the snapshot with the right `PersistentVolumeName`.  If it's found, break.  After the loop, if we didn't find the snapshot, it shouldn't be an error (PVs aren't required to have snapshots -- see https://github.com/heptio/ark/pull/938/files#diff-7e2ddda55b692c82f9fa5f89a3d05e54L920), so just return.  Otherwise, if we did find a snapshot, then go through the rest of the logic that's currently inside the loop.

pkg/restore/restore.go
```diff
@@ -917,44 +939,57 @@ func (r *pvRestorer) executePVAction(obj *unstructured.Unstructured) (*unstructu
...
+		if snapshot.Spec.PersistentVolumeName == pvName {
+			location, err := r.snapshotLocationLister.VolumeSnapshotLocations(r.backupNamespace).Get(snapshot.Spec.Location)
+			if err != nil {
+				return nil, err
```
skriss: `errors.WithStack(err)`

pkg/restore/restore.go
```diff
@@ -939,57 +939,61 @@ func (r *pvRestorer) executePVAction(obj *unstructured.Unstructured) (*unstructu
...
-			if err != nil {
-				return nil, err
-			}
+			foundSnapshot = snapshot
+			break
+		}
+	}
+	if foundSnapshot == nil {
+		log.Info("skipping no snapshot found")
+		return nil, nil
```
skriss: `return obj, nil`

skriss:
The build's failing due to some formatting issues, but there's something weird going on with `make update` where it wants to rewrite all of our import blocks now.  Could you run `make update` locally but only commit the 2 files that the build is complaining about?

pkg/restore/restore.go
```diff
@@ -917,43 +939,59 @@ func (r *pvRestorer) executePVAction(obj *unstructured.Unstructured) (*unstructu
...
+	log = log.WithFields(logrus.Fields{"snapshot": foundSnapshot.Status.ProviderSnapshotID})
 	log.Info("successfully restored persistent volume from snapshot")
+	// used to be update1 which is then cast to Unstructured and returned
+	updated1, err := blockStore.SetVolumeID(obj, volumeID)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
 
-	updated1, err := r.blockStore.SetVolumeID(obj, volumeID)
+	verifyVolumeID, err := blockStore.GetVolumeID(updated1)
```
skriss: was there a specific reason to have this `GetVolumeID` call here?  I don't think it should be necessary, since the `SetVolumeID` call above will return an error if it can't set it properly, but I might be overlooking something

skriss:
One question, but otherwise ready to merge.  I may make some adjustments to the units while we're doing testing, but it's good enough for now.

pkg/restore/restore.go
wwitzel3: That is just oversight from when I was debugging some test failures.

PR #944, opened 2018-10-17
include snapshot locations in created schedule's backup spec

skriss:


skriss:
@carlisia @nrb PTAL.

PR #947, opened 2018-10-17
upgrade to restic v0.9.3

skriss:


nrb:
Change looks fine, though should we call out any bug fixes/features for the new version, even if just in the commit for a changelog entry?

nrb:
> I  just want to stay up-to-date so future upgrades span fewer versions.

Ok, that's fair. This also looks like it's not a point release, so the delta is hopefully small.

skriss:
Yeah, looked at the notes and I don't think there's anything worth calling out.

nrb:
@carlisia @wwitzel3 Would either of you mind reviewing and merging?

PR #949, opened 2018-10-17
remove Config CRD

### LINKED ISSUE #197: refresh the Ark config definition doc
The config-definition doc is (IMHO) becoming unwieldy, particularly with recent changes to the Ark config. The nesting structure of the file has changed, so we're now using `/` to indicate sub-properties more frequently within this doc, which is not entirely clear. I think it's ready for a refresh, but would like to get input from a writer. Maybe we should add commented examples to this page?

### LINKED ISSUE #809: Fully remove `Config` API type, CRD, generated code, documentation, etc.
With the move to backup storage locations, volume snapshot locations, and server command line flags, the `Config` API type is outdated.

Documentation for migrating to v0.10.x and moving relevant data to their new locations should also be provided.

skriss:
This PR removes the Config CRD and all references to it.

Closes #809 
Closes #197


docs/azure-config.md
```diff
@@ -160,7 +143,7 @@ In the root of your Ark directory, run:
   ```
```
carlisia: Just wondering why this `apply` is on the entire directory, and in the other examples we specify each file.

skriss: I don't even remember the specific reason, but there was some difference in either the contents of the azure dir vs. the others, or the order in which things needed to be applied.  I know that for AWS at least, we have 2 different versions of the Ark deployment in the examples/ dir, so we wouldn't want to apply the whole directory.

skriss: Probably could be made to be consistent with the others, though.

skriss: I'll add a docs issue for this.

skriss:
@wwitzel3 @nrb can one of you review/merge? thanks.

PR #960, opened 2018-10-18
#959 Update restoreController runRestore function to specify return arguments

### LINKED ISSUE #959: Remove naked returns
https://github.com/heptio/ark/blob/17b3a3b0730a23e1903663a21efb511ac4e1f95f/pkg/controller/restore_controller.go#L491

The return values for this function should be changed to not be naked returns.

Atul9:
closes #959

### DIFF
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -488,7 +488,9 @@ func (c *restoreController) runRestore(
 	actions []restore.ItemAction,
 	info backupInfo,
 	pluginManager plugin.Manager,
-) (restoreWarnings, restoreErrors api.RestoreResult, restoreFailure error) {
+) (api.RestoreResult, api.RestoreResult, error) {
+	var restoreWarnings, restoreErrors api.RestoreResult
+	var restoreFailure error
 	logFile, err := ioutil.TempFile("", "")
 	if err != nil {
 		c.logger.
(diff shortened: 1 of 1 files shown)


Atul9:
@carlisia I have updated the PR. Is `string` the datatype for `restoreWarnings`? https://github.com/heptio/ark/pull/960/files#diff-7d210f9ced78ff8be1ba497bf49e1a40R491

ncdc:
@Atul9 the original code was `(restoreWarnings, restoreErrors api.RestoreResult, restoreFailure error)`. Because there is no type immediately after `restoreWarnings`, you have to keep reading to the right until you get to a type. In other words, `restoreWarnings, restoreErrors api.RestoreResult` is shorthand for `restoreWarnings api.RestoreResult, restoreErrors api.RestoreResult`, so the type is not `string`; it's `api.RestoreResult`. Does this make sense?

Atul9:
@ncdc Yes, I got it. I am new to Go. Sorry, I should have asked this before opening the PR. Why do we want to remove naked returns? 
What should I update for the build to pass? I didn't get this - https://travis-ci.org/heptio/ark/builds/443754755#L549 (not enough arguments to return)

ncdc:
@Atul9 everywhere in this function that it says only `return`, you'll need to do `return restoreWarnings, restoreErrors, restoreFailure` instead.

Naked returns can lead to confusion - it's not immediately clear what specific values are being returned for each position.

pkg/controller/restore_controller.go
```diff
@@ -488,7 +488,9 @@ func (c *restoreController) runRestore(
 	actions []restore.ItemAction,
 	info backupInfo,
 	pluginManager plugin.Manager,
-) (restoreWarnings, restoreErrors api.RestoreResult, restoreFailure error) {
+) (api.RestoreResult, api.RestoreResult, error) {
```
ncdc: @skriss @carlisia we don't have to do this here, but I dislike having functions that return multiple things of the same type (warnings vs errors for `api.RestoreResult`, in this case). It can lead to confusion and programmer error if you accidentally mess up the order. We should explore alternatives.

carlisia: Agree. An alternative that I often use is to create a struct.

ncdc: exactly :-)

PR #962, opened 2018-10-19
Add CRDs to list of prioritized resources

### LINKED ISSUE #424: Add CRDs to list of prioritized resources
**User Story**

As a cluster operator, I would like to use Ark to backup and restore my Kubernetes clusters which leverage Custom Resource Definitions.

As a Gimbal cluster operator, I would like to be able to use Ark to backup and restore my Gimbal cluster

---
We need to restore CRDs before we restore their associated CRs, so we need to add CRDs to the list of default resource priorities.

It would also be nice if we could wait for a CRD to be "ready" (as indicated in its `status`) before proceeding; otherwise, trying to create CRs for it might fail.

domenicrosati:
Fixes #424

I was struggling with this earlier this week as we use a lot of CRDs in our setup.

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -446,6 +446,8 @@ func (s *server) validateBackupStorageLocations() error {
 // - Limit ranges go before pods or controllers so pods can use them.
 // - Pods go before controllers so they can be explicitly restored and potentially
 //	 have restic restores run before controllers adopt the pods.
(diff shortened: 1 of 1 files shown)


ncdc:
@domenicrosati thanks for your PR! I'm going to set this as fixing 424 and I'll open a new issue for waiting for the CRDs to be ready.

nrb:
@wwitzel3 This has the backport candidate label but was merged just after v0.9.9.

Issue #669, opened 2018-07-11
restic-wait  restore data not found


geofflamrock:
Hi,

I've been trying to get restic backups and restores to minio working, and I think the backup is working ok.

When I attempt to restore the data, the restore says it completed, however the restic-wait container on the pods repeatedly says "Not found: /restores/data/.ark/a4323ca2-84cd-11e8-bcec-00155da8f01e"

Is there some way I can work out why this isn't working?

Thanks.

ncdc:
Hi @geofflamrock, sorry you're having trouble. Could you please share the logs from the `restic-*` pod in the `heptio-ark` namespace?

geofflamrock:
Hi @ncdc , thanks for getting back to me, attached is a zip file with the logs. 

I truncated them at the certain point as they were getting relatively large in size and they just had the same lines over and over again about "Pod is not running restic-wait init container, not enqueuing restores for pod" and "Restore is not new, not enqueuing". If you want the full logs please let me know.

Thanks.
[restic-logs.zip](https://github.com/heptio/ark/files/2186413/restic-logs.zip)

skriss:
@geofflamrock could you also provide the output of `ark restore describe` and `ark restore logs` for your restore?

geofflamrock:
@ncdc attached, thanks.
[restore-describe.txt](https://github.com/heptio/ark/files/2186555/restore-describe.txt)
[restore-logs.txt](https://github.com/heptio/ark/files/2186556/restore-logs.txt)

skriss:
Hmm, it sure looks like 6 restic restores completed successfully.  Are the relevant pods still around (waiting for the init container to complete)? If so, could you provide the YAML for all of them?

geofflamrock:
Hi @ncdc, attached are the restic logs, restore logs and describe and the restored yaml.

The resources I'm specifically trying to backup and restore are kafka and zookeeper stateful sets with persistent volume claims, and I do wonder whether this is part of the problem, as I'm not sure I'm configuring the backup correctly. 

When the restore happens, the stateful set doesn't come back, just the pods. Also, I tried removing the restored pods, which leaves the pvc based volume behind and then re-deploy the stateful set, it picks up the correct volumes but there isn't anything in the data itself, like it never actually got backed up. 

All of this makes me think that my original backup is bad, but everything I could find in it said it completed successfully.

[...]

PR #963, opened 2018-10-19
update restic prerequisite with min k8s version

kmova:
Restic DaemonSet pods require shared mounts with mountPropagation feature enabled on the Kubernetes Cluster.  Refer #669 

Signed-off-by: kmova <kiran.mova@openebs.io>

### DIFF
--- a/docs/restic.md
+++ b/docs/restic.md
@@ -25,6 +25,8 @@ cross-volume-type data migrations. Stay tuned as this evolves!
 
 - A working install of Ark version 0.10.0 or later. See [Set up Ark][2]
 - A local clone of [the latest release tag of the Ark repository][3]
+- Ark's restic integration requires the Kubernetes [MountPropagation feature][6], which is enabled by default in Kubernetes v1.10.0 and later.
+
 
 ### Instructions
 
@@ -243,3 +245,4 @@ on to running other init containers/the main containers.
 [3]: https://github.com/heptio/ark/releases/
 [4]: https://kubernetes.io/docs/concepts/storage/volumes/#local
 [5]: http://restic.readthedocs.io/en/latest/100_references.html#terminology
+[6]: https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation


docs/restic.md
```diff
@@ -25,6 +25,7 @@ cross-volume-type data migrations. Stay tuned as this evolves!
 
 - A working install of Ark version 0.10.0 or later. See [Set up Ark][2]
 - A local clone of [the latest release tag of the Ark repository][3]
+- Enable Mount propagation feature on Kubernetes Cluster. Requires Kubernetes v1.10 or higher. See [Setup MountFlags][6]
```
ncdc: The `MountPropagation` feature is beta in Kubernetes 1.10 and enabled by default. Would it make more sense to say that Ark's restic integration requires Kubernetes 1.10 or higher (make sure you have not disabled the `MountPropagation` feature gate)? We could then also update our example yaml files for the restic DaemonSets to configure `mountPropagation`. I'm guessing we need `hostToContainer`?

kmova: Thanks @ncdc - Modified the text - let me know if this sounds better. 

I have currently tested this with gcp without making any change to the example yaml. Once I have tested on other platforms, I will raise a new PR with updated example yamls.

ncdc:
My gut feeling is that the presence of the `MountPropagation` feature is not what makes things work. The default value is `None` if you don't specify anything else, which would match the previous behavior.

I have seen some propagation issues on Centos with docker-ce 17.03.02, whereas on Ubuntu 16.04 with the same docker version, the problem isn't there. I wonder if something change in the kernel version between GKE 1.9.x and 1.10.x?

skriss:
@ncdc we're specifying `HostToContainer` in our restic daemonset (https://github.com/heptio/ark/blob/master/examples/aws/20-restic-daemonset.yaml#L54)

ncdc:
@skriss thanks, my bad (should have looked at the file!)

docs/restic.md
```diff
@@ -25,6 +25,8 @@ cross-volume-type data migrations. Stay tuned as this evolves!
 
 - A working install of Ark version 0.10.0 or later. See [Set up Ark][2]
 - A local clone of [the latest release tag of the Ark repository][3]
+- Ark's restic integration requires [Kubernetes MountPropagation feature][6]. MountPropgation feature moved to beta in Kubernetes v1.10 and is enabled by default. Make sure it is not disabled via feature-gate.
```
ncdc:
```suggestion
- Ark's restic integration requires the Kubernetes [MountPropagation feature][6], which is enabled by default in Kubernetes v1.10.0 and later.
```

kmova: Thanks @ncdc - updated!

ncdc:
LGTM, thanks @kmova!

PR #970, opened 2018-10-20
create a struct for multiple return of same type in restore_contoroll…

### LINKED ISSUE #967: Create a struct for multiple returns of the same type
To address this: https://github.com/heptio/ark/pull/960#discussion_r226715972

u5surf:
i resolved #967 to create struct which can accepted two api.RestoreResults.


pkg/controller/restore_controller.go
```diff
@@ -85,6 +83,11 @@ type restoreController struct {
 	newBackupStore   func(*api.BackupStorageLocation, persistence.ObjectStoreGetter, logrus.FieldLogger) (persistence.BackupStore, error)
 }
 
+type apiRestoreResult struct {
+	warnings api.RestoreResult
+	errors   api.RestoreResult
```
carlisia: Please move the `errors` var to the same line as `warnings`.

pkg/controller/restore_controller.go
```diff
@@ -233,20 +236,20 @@ func (c *restoreController) processRestore(key string) error {
 	log.Debug("Running restore")
 
 	// execution & upload of restore
-	restoreWarnings, restoreErrors, restoreFailure := c.runRestore(
+	apiRestoreResults, restoreFailure := c.runRestore(
```
carlisia: A bit of a nit pick :) but I'd rather `apiRestoreResults` be in the singular, so `apiRestoreResult`.

ncdc: I'd prefer not to have `api` as part of the name, if possible

pkg/controller/restore_controller.go
u5surf: which means to declear single field of apiRestoreResult to be mixing warnings with errors? or to be declaring as an array?

carlisia: Yes, the first, just move the variable so both are on the same line, like so: `warnings, errors api.RestoreResult`

pkg/controller/restore_controller.go
```diff
@@ -85,6 +83,10 @@ type restoreController struct {
 	newBackupStore   func(*api.BackupStorageLocation, persistence.ObjectStoreGetter, logrus.FieldLogger) (persistence.BackupStore, error)
 }
 
+type restoreResult struct {
+	warnings, errors   api.RestoreResult
```
carlisia: There are extra spaces here, it'd be great if you could remove it :)

I think if you run ` go fmt ./... ` at the root of the project it should tell you about any misformating in the code. Or you could configure your editor to use `gofmt` on save.

u5surf: I remove it and rebased the commit

ncdc:
Hold

PR #972, opened 2018-10-22
Support pre v0.10 snaps

### LINKED ISSUE #966: [v0.10] support restoring backups w/snapshots taken prior to v0.10
Prior to v0.10, Ark backups store metadata about PV snapshots as a map in their status (`.status.volumeBackups`).  A/o v0.10, this information is moving into a JSON file that gets stored in object storage (see #922).  As part of this, the JSON file will store the `VolumeSnapshotLocation` where the snapshot is stored.  This raises an issue around how v0.10 will restore backups taken prior to v0.10 that include snapshots.

My proposal is:
- retain the `.status.volumeBackups` field, but mark it deprecated/read-only
- during a restore, if a backup has *both* a `.status.volumeBackups` field and a volumesnapshots JSON file, immediately error and return (possibly fail validation?), as Ark should never put a backup into this state.
- during a restore, if a backup has the `.status.volumeBackups` field, restore the snapshots *only if* there is only a single VolumeSnapshotLocation configured, making it obvious where the snapshots are expected to be.

I think this is reasonable because (a) it keeps the amount of backwards-compatibility code very small, (b) I expect that the vast majority of users will only have a single VolumeSnapshotLocation configured, and (c) over time, these "legacy" backups will expire and be garbage-collected, making it easier for us to officially drop support for them in subsequent releases.

Appreciate input from everyone! cc @wwitzel3 @nrb @carlisia @ncdc

skriss:
Fixes #966

### DIFF
--- a/pkg/controller/backup_deletion_controller.go
+++ b/pkg/controller/backup_deletion_controller.go
@@ -245,25 +245,47 @@ func (c *backupDeletionController) processRequest(req *v1.DeleteBackupRequest) e
 
 	if backupStore != nil {
 		log.Info("Removing PV snapshots")
-		if snapshots, err := backupStore.GetBackupVolumeSnapshots(backup.Name); err != nil {
-			errs = append(errs, errors.Wrap(err, "error getting backup's volume snapshots").Error())
-		} else {
-			blockStores := make(map[string]cloudprovider.BlockStore)
-
-			for _, snapshot := range snapshots {
-				log.WithField("providerSnapshotID", snapshot.Status.ProviderSnapshotID).Info("Removing snapshot associated with backup")
-
-				blockStore, ok := blockStores[snapshot.Spec.Location]
-				if !ok {
-					if blockStore, err = blockStoreForSnapshotLocation(backup.Namespace, snapshot.Spec.Location, c.snapshotLocationLister, pluginManager); err != nil {
-						errs = append(errs, err.Error())
-						continue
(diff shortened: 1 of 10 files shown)


pkg/controller/backup_deletion_controller.go
```diff
@@ -245,25 +245,47 @@ func (c *backupDeletionController) processRequest(req *v1.DeleteBackupRequest) e
...
+			// v0.10+ backup
+			if snapshots, err := backupStore.GetBackupVolumeSnapshots(backup.Name); err != nil {
+				errs = append(errs, errors.Wrap(err, "error getting backup's volume snapshots").Error())
+			} else {
+				blockStores := make(map[string]cloudprovider.BlockStore)
+
+				for _, snapshot := range snapshots {
+					log.WithField("providerSnapshotID", snapshot.Status.ProviderSnapshotID).Info("Removing snapshot associated with backup")
+
+					blockStore, ok := blockStores[snapshot.Spec.Location]
```
nrb: I think I'm missing something here - wouldn't this always return `nil, false` since `blockStores` is created on L272?

nrb: Nevermind, I didn't read far enough.

nrb:
LGTM overall. Do you want to squash before merge?

skriss:
Made one change here to add the `omitempty` tag to `backup.status.volumeBackups`, so it stops showing up in the JSON/YAML for v0.10+ backups - otherwise should be good to merge

PR #984, opened 2018-10-23
instructions for upgrading to v0.10

skriss:

First pass at writing up end-to-end upgrade instructions.  Please provide feedback!  If anyone hasn't gone through their upgrading testing yet, would be awesome if you could use this as the basis for it.

I'm assuming that we'll link prominently to this page in the release notes, and we'll also need to find a place for it on the docs site.

### DIFF
--- a/docs/upgrading-to-v0.10.md
+++ b/docs/upgrading-to-v0.10.md
@@ -1,0 +1,82 @@
+# Upgrading to Ark v0.10
+
+## Overview
+
+Ark v0.10 includes a number of breaking changes. Below, we outline what those changes are, and what steps you should take to ensure
+a successful upgrade from prior versions of Ark.
+
+## Breaking Changes
+
+### Switch from Config to BackupStorageLocation and VolumeSnapshotLocation CRDs, and new server flags
+
+Prior to v0.10, Ark used a `Config` CRD to capture information about your backup storage and persistent volume providers, as well
+some miscellaneous Ark settings. In v0.10, we've eliminated this CRD and replaced it with:
+
+- A [BackupStorageLocation][1] CRD to capture information about where to store your backups
+- A [VolumeSnapshotLocation][2] CRD to capture information about where to store your persistent volume snapshots
+- Command-line flags for the `ark server` command (run by your Ark deployment) to capture miscellaneous Ark settings
+
+When upgrading to v0.10, you'll need to transfer the configuration information that you currently have in the `Config` CRD
+into the above. We'll cover exactly how to do this below.
+
+### Reorganization of data in object storage
+
+We've made [changes to the layout of data stored in object storage][3] for simplicity and extensibility. You'll need to
+rearrange any pre-v0.10 data as part of the upgrade. We've provided a script to help with this.
+
+## Step-by-Step Upgrade Instructions
+
+1. Scale down your existing Ark deployment:
+```bash
+kubectl scale -n heptio-ark deploy/ark --replicas 0
+```
+
+2. Re-apply the `00-prereqs.yaml` file to create new CRDs:
+```bash
+kubectl apply -f examples/common/00-prereqs.yaml
+```
+
+3. Create one or more [BackupStorageLocation][1] resources based on the examples provided in the `examples/` directory for your platform, using information from
+the existing `Config` resource as necessary.
+4. If you're using Ark to take PV snapshots, create one or more [VolumeSnapshotLocation][2] resources based on the examples provided in the `examples/` directory for your platform, using information from
+the existing `Config` resource as necessary.
+
+5. Perform the one-time object storage migration detailed [here][3].
+
+6. In your Ark deployment YAML (see the `examples/` directory for samples), specify flags to the `ark server` command under the container's `args`:
+
+    a. The names of the `BackupStorageLocation` and `VolumeSnapshotLocation(s)` that should be used by default for backups. If defaults are set here, 
+    users won't need to explicitly specify location names when creating backups (though they still can, if they want to store backups/snapshots in
+    alternate locations). If no value is specified for `--default-backup-storage-location`, the Ark server looks for a `BackupStorageLocation` 
+    named `default` to use.
+
+    Flag | Default Value | Description | Example
+    ---- | ------------- | ----------- | -------
+    `--default-backup-storage-location` | "default" | name of the backup storage location that should be used by default for backups | aws-us-east-1-bucket
+    `--default-volume-snapshot-locations` | [none] | name of the volume snapshot location(s) that should be used by default for PV snapshots, for each PV provider | aws:us-east-1,portworx:local
+
+    **NOTE:** the values of these flags should correspond to the names of a `BackupStorageLocation` and `VolumeSnapshotLocation(s)` custom resources
+    in the cluster.
+
+    b. Any non-default Ark server settings:
+
+    Flag | Default Value | Description
+    ---- | ------------- | -----------
+    `--backup-sync-period` | 1m | how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster
+    `--restic-timeout` | 1h | how long backups/restores of pod volumes should be allowed to run before timing out (previously `podVolumeOperationTimeout` in the `Config` resource in pre-v0.10 versions)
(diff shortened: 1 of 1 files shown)


docs/upgrading-to-v0.10.md
```diff
@@ -0,0 +1,75 @@
...
+
+6. In your Ark deployment YAML (see the `examples/` directory for samples), specify any non-default settings as args to the `ark server` command:
+
+| Flag | Default Value |
+| ---- | ------------- |
+| `--backup-sync-period` | 1m |
+| `--restic-timeout` | 1h |
+| `--restore-only` | false |
+
+7. In your Ark deployment YAML, define your default `BackupStorageLocation` and `VolumeSnapshotLocations`  as args to the `ark server` command:
```
carlisia:
```suggestion
7. Still in your Ark deployment YAML, define the following as args to the `ark server` command:
- change `BackupStorageLocation` from the preset `default` name to another name if you would like to; this is the name of the location your backup will automatically be stored at so you don't need to specify it when you create a backup. If you want to use this default feature, it needs to match the name of at least 1 `BackupStorageLocation` CRD.
- if you will be backing up PVs, add `VolumeSnapshotLocations` (1 per provider, ex: `provider1:location-01,provider2:location-02,...`)
```

docs/upgrading-to-v0.10.md
```diff
@@ -0,0 +1,75 @@
...
+
+| Flag | Default Value | Description |
+| ---- | ------------- | ----------- |
+| `--default-backup-storage-location` | "default" | name of the backup storage location that should be used by default for backups |
```
carlisia: Technically this is redundant except if they change the name to something other than `default`. If you think the suggestion I made above makes sense, maybe change the name here to something else, idk, maybe `new-location`.

carlisia:
```suggestion
| `--default-backup-storage-location` | "new-location" | overwrite the preset name of `default` for the backup storage location that will be used by default for backups. This means you don't have to specify it when you create backups to be stored in this location (but you still need to define a `BackupStorageLocation` CRD) |
```

docs/upgrading-to-v0.10.md
```diff
@@ -0,0 +1,75 @@
+# Upgrading to Ark v0.10
+
+## Overview
+
+Ark v0.10 includes a number of breaking changes. Below, we outline what those changes are, and what steps you should take to ensure
+a successful upgrade from prior versions of Ark.
+
+## Breaking Changes
+
+### Switch from Config CRD to BackupStorageLocation, VolumeSnapshotLocation, and server flags
```
carlisia:
```suggestion
### Switch from Config to BackupStorageLocation and VolumeSnapshotLocation CRDs, and new server flags
```

docs/upgrading-to-v0.10.md
carlisia:
```suggestion
7. In your Ark deployment YAML, define your default `BackupStorageLocation` and `VolumeSnapshotLocations` as args to the `ark server` command:
```

carlisia: Ah one thing I almost forgot to mention: in the deployment yaml under each provider's example folder, I think we should add at least an example for the BSL (w/ a name that is not `default` since that name is already the preset name in ark server), maybe even a couple for VSL. What do you think?

docs/upgrading-to-v0.10.md
carlisia: I was referring to the name `default`, since if you don't specify it it is the name ark server gives the default bsl. So naming it `default` is redundant.

docs/upgrading-to-v0.10.md
nrb: Torn on this - the examples show the structure of the CRD. For most people, they're going to apply the example YAML file for their provider and only have 1 backupstoragelocation, much like how things were prior to v0.10. If we provide other examples, I think it will lead to confusion as to what file(s) to apply.

docs/upgrading-to-v0.10.md
nrb: The server doesn't create or name a BSL. That's name of the BSL it looks for - that CR still has to be created.

carlisia: Oh, thanks for pointing that out @nrb. I missed that completely.

carlisia: This was very confusing to me. I'm just saying, *now* I grasp this, but it took a lot of effort in trying to wrap my head around it. I suspect the language we are using could lead to understanding that we are creating a bsl, even tho we know this is not so.

carlisia: `That's name of the BSL it looks for` -> I really like this language.

docs/upgrading-to-v0.10.md
carlisia: I `s/define/may define` to maybe reduce the potential for misunderstanding that you `must` define those args.

skriss:
thanks for the feedback @carlisia @nrb! I made some revisions based on your input - PTAL at the latest commit and LMK if it addresses your comments

docs/upgrading-to-v0.10.md
```diff
@@ -0,0 +1,82 @@
...
+        `--default-volume-snapshot-locations` | [none] | name of the volume snapshot location(s) that should be used by default for PV snapshots, for each PV provider | aws:us-east-1,portworx:local
+
+    **NOTE:** the values of these flags should correspond to the names of a `BackupStorageLocation` and `VolumeSnapshotLocation(s)` that you've created as
+    custom resources in the cluster, per steps (3) and (4).
+
+    b. Any non-default Ark server settings:
+
+        Flag | Default Value | Description
+        ---- | ------------- | -----------
+        `--backup-sync-period` | 1m | how often to ensure all Ark backups in object storage exist as Backup API objects in the cluster
```
nrb: Would it make sense to provide the names of the keys from `Config` struct here, too? `restic-timeout` was `podVolumeOperationTimeout` before, which isn't really obvious.

skriss: Maybe I'll just mention it in the description for that one?  Seems redundant for the rest.

docs/upgrading-to-v0.10.md
```diff
@@ -0,0 +1,82 @@
...
+    a. The names of the `BackupStorageLocation` and `VolumeSnapshotLocation(s)` that should be used by default for backups, if none are explicitly
+    specified by the user when creating a backup. Note that by default, the Ark server looks for a `BackupStorageLocation` named `default` to
+    use.
+
+        Flag | Default Value | Description | Example
+        ---- | ------------- | ----------- | -----------------------
+        `--default-backup-storage-location` | "default" | name of the backup storage location that should be used by default for backups | aws-us-east-1-bucket
+        `--default-volume-snapshot-locations` | [none] | name of the volume snapshot location(s) that should be used by default for PV snapshots, for each PV provider | aws:us-east-1,portworx:local
+
+    **NOTE:** the values of these flags should correspond to the names of a `BackupStorageLocation` and `VolumeSnapshotLocation(s)` that you've created as
```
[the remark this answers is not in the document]
skriss: Maybe I'll just remove this note? I think it's redundant with the main text of 6a - wdyt?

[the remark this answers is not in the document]
skriss: eh, nvm, wordsmithing.

docs/upgrading-to-v0.10.md
nrb: Yeah, that's fine. I think it was undocumented earlier, too, so I'm skeptical it was in widespread use.

skriss:
Another rev @nrb @carlisia

carlisia:
Perfect.

skriss:
squashed to one commit, but will make more changes as needed

skriss:
ugh, looks like i have formatting issues with my tables.  looking at it...

nrb:
fwiw https://marketplace.visualstudio.com/items?itemName=josa.markdown-table-formatter was useful for me.

skriss:
OK, this is better.

nrb:
Merging this, if there are further changes necessary we can iterate.

PR #985, opened 2018-10-23
Add prometheus metrics for successful and attempted volume snapshots per backup

### LINKED ISSUE #977: Add prometheus metrics for backup.status.volumeSnapshotsAttempted/volumeSnapshotsSucceeded
These two new fields capture the total number of volume snapshots tried, and succeeded, per backup.  We should add prometheus metrics to capture these (or the converse - failures).

shubheksha:
Fixes #977 
Signed-off-by: Shubheksha Jalan <jshubheksha@gmail.com>
/cc @skriss


pkg/metrics/metrics.go
```diff
@@ -236,3 +254,17 @@ func (m *ServerMetrics) RegisterRestoreValidationFailed(backupSchedule string) {
 		c.WithLabelValues(backupSchedule).Inc()
 	}
 }
+
+// RegisterVolumeSnapshotsAttemptedPerBackup records a restore that failed validation.
```
ncdc: Please update this comment 😄

pkg/metrics/metrics.go
```diff
@@ -236,3 +254,17 @@ func (m *ServerMetrics) RegisterRestoreValidationFailed(backupSchedule string) {
...
+func (m *ServerMetrics) RegisterVolumeSnapshotsAttemptedPerBackup(backupSchedule string) {
+	if c, ok := m.metrics[volumeSnapshotsAttemptedPerBackup].(*prometheus.CounterVec); ok {
+		c.WithLabelValues(backupSchedule).Inc()
+	}
+}
+
+// RegisterVolumeSnapshotsCompletedPerBackup records a restore that failed validation.
```
ncdc: Please update this comment 😄

pkg/controller/backup_controller.go
```diff
@@ -428,6 +428,8 @@ func recordBackupMetrics(backup *api.Backup, backupFile *os.File, serverMetrics
 	backupDuration := backup.Status.CompletionTimestamp.Time.Sub(backup.Status.StartTimestamp.Time)
 	backupDurationSeconds := float64(backupDuration / time.Second)
 	serverMetrics.RegisterBackupDuration(backupScheduleName, backupDurationSeconds)
+	serverMetrics.RegisterVolumeSnapshotsCompletedPerBackup(backupScheduleName)
```
ncdc: We need to include the number of completed snapshots, don't we?

pkg/controller/backup_controller.go
```diff
@@ -428,6 +428,8 @@ func recordBackupMetrics(backup *api.Backup, backupFile *os.File, serverMetrics
...
+	serverMetrics.RegisterVolumeSnapshotsAttemptedPerBackup(backupScheduleName)
```
ncdc: We need to include the number of attempted snapshots, don't we?

pkg/metrics/metrics.go
```diff
@@ -31,14 +31,16 @@ const (
...
-	restoreFailedTotal           = "restore_failed_total"
+	backupAttemptCount                = "backup_attempt_total"
+	backupSuccessCount                = "backup_success_total"
+	backupFailureCount                = "backup_failure_total"
+	backupDurationSeconds             = "backup_duration_seconds"
+	restoreAttemptTotal               = "restore_attempt_total"
+	restoreValidationFailedTotal      = "restore_validation_failed_total"
+	restoreSuccessTotal               = "restore_success_total"
+	restoreFailedTotal                = "restore_failed_total"
+	volumeSnapshotsAttemptedPerBackup = "volume_snapshot_attempted_per_backup"
```
ncdc: This is probably better named `volume_snapshot_attempt_total`

pkg/metrics/metrics.go
```diff
@@ -31,14 +31,16 @@ const (
...
+	volumeSnapshotsCompletedPerBackup = "volume_snapshot_completed_per_backup"
```
ncdc: This is probably better named `volume_snapshot_success_total`

pkg/metrics/metrics.go
shubheksha: Won't that be confusing - across backups vs per backup?

pkg/controller/backup_controller.go
shubheksha: I was following what's done for other counter variables: https://github.com/heptio/ark/blob/master/pkg/metrics/metrics.go#L186-L190

shubheksha: Do we need to explicitly pass it?

pkg/metrics/metrics.go
ncdc: Prometheus metrics are a bit "weird" compared to what you might expect. If we do this per backup, then the individual metrics will balloon - you'd have

```
volume_snapshot_attempted_per_backup[backup-1]=12
volume_snapshot_attempted_per_backup[some-other-backup]=3
...
```

@rosskukulinski may be able to provide some better guidance here

shubheksha: So the current implementation will not give us metrics per backup? Because that's what the issue description mentions. So I'm a little confused here 😅

pkg/controller/backup_controller.go
ncdc: Yes. Looking at what @skriss wrote in the issue:
> These two new fields capture the total number of volume snapshots tried, and succeeded, per backup. We should add prometheus metrics to capture these (or the converse - failures).

The way it's coded in this PR now, these 2 new counters are identical to and effectively duplicates of `backupAttemptCount`.

shubheksha: Ah, gotcha! Fixing it

pkg/metrics/metrics.go
ncdc: We generally record metrics per schedule. It's possible the issue was misleading. What I think we're really after is
- # of attempted volume snapshots, per schedule
- # of successful volume snapshots, per schedule
- # of failed volume snapshots, per schedule

pkg/controller/backup_controller.go
shubheksha: So here we essentially get that value and set the counter to it, right?

ncdc:
No, we’d still want to increment by the number

pkg/metrics/metrics.go
```diff
@@ -236,3 +253,17 @@ func (m *ServerMetrics) RegisterRestoreValidationFailed(backupSchedule string) {
 		c.WithLabelValues(backupSchedule).Inc()
 	}
 }
+
+// RegisterVolumeSnapshotsAttempt records an attempt to snapshot a volume.
+func (m *ServerMetrics) RegisterVolumeSnapshotsAttempt(backupSchedule string, volumeSnapshotsAttempted int) {
+	if c, ok := m.metrics[volumeSnapshotsAttemptTotal].(*prometheus.CounterVec); ok {
+		c.WithLabelValues(backupSchedule).Set(float64(volumeSnapshotsAttempted))
+	}
```
shubheksha: Will change this to `Add()` in the next pass along with other comments

pkg/metrics/metrics.go
```diff
@@ -236,3 +253,17 @@ func (m *ServerMetrics) RegisterRestoreValidationFailed(backupSchedule string) {
...
+}
+
+// RegisterVolumeSnapshotsCompleted records a completed volume snapshot.
+func (m *ServerMetrics) RegisterVolumeSnapshotsCompleted(backupSchedule string, volumeSnapshotsCompleted int) {
+	if c, ok := m.metrics[volumeSnapshotsCompletedTotal].(*prometheus.CounterVec); ok {
+		c.WithLabelValues(backupSchedule).Set(float64(volumeSnapshotsCompleted))
+	}
```
shubheksha: Will change this to `Add()` in the next pass along with other comments

pkg/metrics/metrics.go
```diff
@@ -28,17 +28,18 @@ type ServerMetrics struct {
...
+	backupAttemptTotal            = "backup_attempt_total"
+	backupSuccessTotal            = "backup_success_total"
+	backupFailureTotal            = "backup_failure_total"
+	backupDurationSeconds         = "backup_duration_seconds"
+	restoreAttemptTotal           = "restore_attempt_total"
+	restoreValidationFailedTotal  = "restore_validation_failed_total"
+	restoreSuccessTotal           = "restore_success_total"
+	restoreFailedTotal            = "restore_failed_total"
+	volumeSnapshotsAttemptTotal   = "volume_snapshot_attempt_total"
+	volumeSnapshotsCompletedTotal = "volume_snapshot_success_total"
```
ncdc: Let's call the variable `volumeSnapshotsSuccessTotal` to match the metric name. Please also add `volumeSnapshotsFailureTotal`.

pkg/metrics/metrics.go
```diff
@@ -133,6 +134,22 @@ func NewServerMetrics() *ServerMetrics {
 				},
 				[]string{scheduleLabel},
 			),
+			volumeSnapshotsAttemptTotal: prometheus.NewCounterVec(
+				prometheus.CounterOpts{
+					Namespace: metricNamespace,
+					Name:      volumeSnapshotsAttemptTotal,
+					Help:      "Total number of volume snapshots attempted per backup",
```
ncdc: Total number of attempted volume snapshots

pkg/metrics/metrics.go
```diff
@@ -133,6 +134,22 @@ func NewServerMetrics() *ServerMetrics {
...
+				},
+				[]string{scheduleLabel},
+			),
+			volumeSnapshotsCompletedTotal: prometheus.NewCounterVec(
+				prometheus.CounterOpts{
+					Namespace: metricNamespace,
+					Name:      volumeSnapshotsCompletedTotal,
+					Help:      "Total number of volume snapshots attempted per backup",
```
ncdc: Total number of successful volume snapshots

pkg/metrics/metrics.go
```diff
@@ -145,13 +162,13 @@ func (m *ServerMetrics) RegisterAllMetrics() {
...
 	}
-	if c, ok := m.metrics[backupSuccessCount].(*prometheus.CounterVec); ok {
+	if c, ok := m.metrics[backupSuccessTotal].(*prometheus.CounterVec); ok {
 		c.WithLabelValues(scheduleName).Set(0)
 	}
-	if c, ok := m.metrics[backupFailureCount].(*prometheus.CounterVec); ok {
+	if c, ok := m.metrics[backupFailureTotal].(*prometheus.CounterVec); ok {
 		c.WithLabelValues(scheduleName).Set(0)
 	}
 	if c, ok := m.metrics[restoreAttemptTotal].(*prometheus.CounterVec); ok {
```
ncdc: I think we also need to initialize the new snapshot metrics to 0 here too.

pkg/metrics/metrics.go
shubheksha: I believe a new field needs to be added to `BackupStatus` here for that and then we can check the status?
https://github.com/heptio/ark/blob/master/pkg/apis/ark/v1/backup.go#L163

ncdc: It's the difference between attempted & completed (right @skriss?)

pkg/metrics/metrics.go
```diff
@@ -28,17 +28,18 @@ type ServerMetrics struct {
...
+	backupTarballSizeBytesGauge  = "backup_tarball_size_bytes"
+	backupAttemptTotal           = "backup_attempt_total"
+	backupSuccessTotal           = "backup_success_total"
+	backupFailureTotal           = "backup_failure_total"
 	backupDurationSeconds        = "backup_duration_seconds"
 	restoreAttemptTotal          = "restore_attempt_total"
 	restoreValidationFailedTotal = "restore_validation_failed_total"
 	restoreSuccessTotal          = "restore_success_total"
 	restoreFailedTotal           = "restore_failed_total"
+	volumeSnapshotsAttemptTotal  = "volume_snapshot_attempt_total"
```
ncdc: 1 more edit - it should be volumeSnapshot... (singular) instead of volumeSnapshots... (plural). Same for the next line please. Thanks!

pkg/metrics/metrics.go
skriss: correct

pkg/metrics/metrics.go
shubheksha: Fixed it, PTAL :)

pkg/metrics/metrics.go
shubheksha: Oh, I added a new field for that and computed it based on `SnapshotPhaseFailed` - is that incorrect, @skriss?

pkg/apis/ark/v1/backup.go
```diff
@@ -203,6 +203,10 @@ type BackupStatus struct {
 	// VolumeSnapshotsCompleted is the total number of successfully
 	// completed volume snapshots for this backup.
 	VolumeSnapshotsCompleted int `json:"volumeSnapshotsCompleted"`
+
+	// VolumeSnapshotsFailed is the total number of fialed volume
+	// snapshots for this backup.
+	VolumeSnapshotsFailed int `json:"volumeSnapshotsFailed"`
```
ncdc: I don't think we necessarily need this recorded here - I assumed we would just calculate it for the metrics by doing the subtraction. @skriss wdyt?

pkg/apis/ark/v1/backup.go
```diff
@@ -203,6 +203,10 @@ type BackupStatus struct {
 	// VolumeSnapshotsCompleted is the total number of successfully
 	// completed volume snapshots for this backup.
 	VolumeSnapshotsCompleted int `json:"volumeSnapshotsCompleted"`
+
+	// VolumeSnapshotsFailed is the total number of fialed volume
```
ncdc: If we do decide to keep this, typo: fialed -> failed

pkg/apis/ark/v1/backup.go
skriss: Yeah I think we can just calculate it, no need for API field

shubheksha: Alright, fixing it :)

shubheksha:
@ncdc addressed all your comments, PTAL :)

skriss:
@shubheksha sorry for the delay, have been super-focused on getting the first v0.10 beta out :) will review and provide input tomorrow!

pkg/metrics/metrics.go
```diff
@@ -166,6 +193,12 @@ func (m *ServerMetrics) InitSchedule(scheduleName string) {
 	if c, ok := m.metrics[restoreValidationFailedTotal].(*prometheus.CounterVec); ok {
 		c.WithLabelValues(scheduleName).Set(0)
 	}
+	if c, ok := m.metrics[volumeSnapshotSuccessTotal].(*prometheus.CounterVec); ok {
+		c.WithLabelValues(scheduleName).Set(0)
+	}
+	if c, ok := m.metrics[volumeSnapshotAttemptTotal].(*prometheus.CounterVec); ok {
+		c.WithLabelValues(scheduleName).Set(0)
+	}
```
skriss: Looks like we need to initialize the "failed" metric here too

pkg/metrics/metrics.go
```diff
@@ -236,3 +269,24 @@ func (m *ServerMetrics) RegisterRestoreValidationFailed(backupSchedule string) {
 		c.WithLabelValues(backupSchedule).Inc()
 	}
 }
+
+// RegisterVolumeSnapshotsAttempt records an attempt to snapshot a volume.
```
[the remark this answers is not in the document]
skriss: I'd go with `RegisterVolumeSnapshotAttempts`

pkg/metrics/metrics.go
```diff
@@ -236,3 +269,24 @@ func (m *ServerMetrics) RegisterRestoreValidationFailed(backupSchedule string) {
...
+func (m *ServerMetrics) RegisterVolumeSnapshotsAttempt(backupSchedule string, volumeSnapshotsAttempted int) {
+	if c, ok := m.metrics[volumeSnapshotAttemptTotal].(*prometheus.CounterVec); ok {
+		c.WithLabelValues(backupSchedule).Add(float64(volumeSnapshotsAttempted))
+	}
+}
+
+// RegisterVolumeSnapshotsCompleted records a completed volume snapshot.
```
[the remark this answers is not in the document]
skriss: I'd go with `RegisterVolumeSnapshotSuccesses`

pkg/metrics/metrics.go
```diff
@@ -236,3 +269,24 @@ func (m *ServerMetrics) RegisterRestoreValidationFailed(backupSchedule string) {
...
+func (m *ServerMetrics) RegisterVolumeSnapshotsCompleted(backupSchedule string, volumeSnapshotsCompleted int) {
+	if c, ok := m.metrics[volumeSnapshotSuccessTotal].(*prometheus.CounterVec); ok {
+		c.WithLabelValues(backupSchedule).Add(float64(volumeSnapshotsCompleted))
+	}
+}
+
+// RegisterVolumeSnapshotsFailed records a failed volume snapshot.
+func (m *ServerMetrics) RegisterVolumeSnapshotsFailed(backupSchedule string, volumeSnapshotsFailed int) {
```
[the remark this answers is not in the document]
skriss: I'd go with `RegisterVolumeSnapshotFailures`

shubheksha:
@ncdc / @skriss - this is ready for a final review!

shubheksha:
@skriss done!

PR #988, opened 2018-10-24
Improve directions when location CRDs not deployed

carlisia:
Wonder if any or all of this would help nudge the user into understanding there's a `step 2` in using our default location/s, which is to deploy the corresponding CRD or any CRD.

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -29,6 +29,7 @@ import (
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
@@ -274,7 +275,11 @@ func (c *backupController) prepareBackupRequest(backup *api.Backup) *pkgbackup.R
 
 	// validate the storage location, and store the BackupStorageLocation API obj on the request
 	if storageLocation, err := c.backupLocationLister.BackupStorageLocations(request.Namespace).Get(request.Spec.StorageLocation); err != nil {
-		request.Status.ValidationErrors = append(request.Status.ValidationErrors, fmt.Sprintf("Error getting backup storage location: %v", err))
+		if apierrors.IsNotFound(err) {
+			request.Status.ValidationErrors = append(request.Status.ValidationErrors, fmt.Sprintf("a BackupStorageLocation CRD with the name specified in the backup spec needs to be created before this backup can be executed. Error: %v", err))
+		} else {
+			request.Status.ValidationErrors = append(request.Status.ValidationErrors, fmt.Sprintf("error getting backup storage location: %v", err))
+		}
 	} else {
 		request.StorageLocation = storageLocation
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/backup_controller.go
```diff
@@ -273,7 +274,11 @@ func (c *backupController) prepareBackupRequest(backup *api.Backup) *pkgbackup.R
 
 	// validate the storage location, and store the BackupStorageLocation API obj on the request
 	if storageLocation, err := c.backupLocationLister.BackupStorageLocations(request.Namespace).Get(request.Spec.StorageLocation); err != nil {
-		request.Status.ValidationErrors = append(request.Status.ValidationErrors, fmt.Sprintf("Error getting backup storage location: %v", err))
+		if strings.Contains(err.Error(), "not found") {
```
nrb: I think I would use `apierrors.IsNotFound(err)` here.

pkg/controller/backup_controller.go
```diff
@@ -273,7 +274,11 @@ func (c *backupController) prepareBackupRequest(backup *api.Backup) *pkgbackup.R
...
+			request.Status.ValidationErrors = append(request.Status.ValidationErrors, fmt.Sprintf("a BackupStorageLocation CRD needs to be deployed before this backup can be created. Error: %v", err))
```
nrb: +1 to this message.

carlisia:
I fixed the merge conflict.

pkg/controller/backup_controller.go
skriss: I'd tweak this to clarify that a BSL *with the name specified in the backup spec* needs to be created/deployed, not just any BSL (similarly to the VSL messages below).  And maybe "before this backup can be executed" rather than "created".

pkg/controller/backup_controller.go
```diff
@@ -309,7 +315,11 @@ func (c *backupController) validateAndGetSnapshotLocations(backup *api.Backup) (
 		// validate each locationName exists as a VolumeSnapshotLocation
 		location, err := c.snapshotLocationLister.VolumeSnapshotLocations(backup.Namespace).Get(locationName)
 		if err != nil {
-			errors = append(errors, fmt.Sprintf("error getting volume snapshot location named %s: %v", locationName, err))
+			if strings.Contains(err.Error(), "not found") {
```
skriss: use `apierrors.isNotFound(err)`

pkg/controller/backup_controller.go
```diff
@@ -309,7 +315,11 @@ func (c *backupController) validateAndGetSnapshotLocations(backup *api.Backup) (
...
+				errors = append(errors, fmt.Sprintf("a VolumeSnapshotLocation CRD for the location %s needs to be deployed before this snapshot can be created. Error: %v", locationName, err))
```
skriss: Maybe "executed" instead of "created"

pkg/controller/backup_sync_controller.go
```diff
@@ -122,7 +122,7 @@ func shouldSync(location *arkv1api.BackupStorageLocation, now time.Time, backupS
 }
 
 func (c *backupSyncController) run() {
-	c.logger.Info("Checking for backup storage locations to sync into cluster")
+	c.logger.Info("Waiting for existing backup storage locations to sync into cluster")
```
skriss: I changed this message to be "Checking for existing backup storage locations...." which I think is more appropriate in all scenarios, including the one where BSLs **do** exist (also changed it to debug, so it's not filling up the log).  So I'd revert this change

carlisia:
I'll make these changes and also rebase.

carlisia:
@skriss updated, ptal when you have a chance.

PR #997, opened 2018-10-25
Only require a default VSL if > 1 vsl/provider exists

### LINKED ISSUE #995: Don't require default volume snapshot location to be specified for a provider if only 1 exists
Right now, if you don't specify a default VSL for a provider and you don't specify the name of a VSL when creating a backup, you won't get snapshots.  This is overly burdensome for the typical case of "one provider, one snapshot location".  In the case where a provider has a single VSL, we should automatically use that.  Specifying a default should only be necessary in the case of >1.

I think the changes probably need to be made here: https://github.com/heptio/ark/blob/master/pkg/controller/backup_controller.go#L302

carlisia:
I made it so in the scenario (A) where:
- only 1 vsl exists
- existing vsl location is named `primary`
- admin later boots up ark server with default named `random`

We return the existing VSL and disregard the default. Because in the scenario (B) below:
- Admin boots up ark with vsl default set to `random`
- User later creates vsl named `primary`

Obviously in this later case the user will not have access to the default feature. I think the system should behave the same as it does in scenario (B) as it would if the order of operations had been reversed, ie, scenario (A).

Fixes #995.

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -273,7 +273,7 @@ func (s *server) run() error {
 			Warnf("Default backup storage location %q not found; backups must explicitly specify a location", s.config.defaultBackupLocation)
 	}
 
-	defaultVolumeSnapshotLocations, err := getDefaultVolumeSnapshotLocations(s.arkClient, s.namespace, s.config.defaultVolumeSnapshotLocations)
+	defaultVolumeSnapshotLocations, err := getVolumeSnapshotLocations(s.arkClient, s.namespace, s.config.defaultVolumeSnapshotLocations)
 	if err != nil {
 		return err
 	}
@@ -289,15 +289,10 @@ func (s *server) run() error {
 	return nil
 }
 
-func getDefaultVolumeSnapshotLocations(arkClient clientset.Interface, namespace string, defaultVolumeSnapshotLocations map[string]string) (map[string]*api.VolumeSnapshotLocation, error) {
-	providerDefaults := make(map[string]*api.VolumeSnapshotLocation)
-	if len(defaultVolumeSnapshotLocations) == 0 {
-		return providerDefaults, nil
-	}
-
+func getVolumeSnapshotLocations(arkClient clientset.Interface, namespace string, defaultVolumeSnapshotLocations map[string]string) (map[string]*api.VolumeSnapshotLocation, error) {
 	volumeSnapshotLocations, err := arkClient.ArkV1().VolumeSnapshotLocations(namespace).List(metav1.ListOptions{})
 	if err != nil {
-		return providerDefaults, errors.WithStack(err)
+		return nil, errors.WithStack(err)
 	}
 
 	providerLocations := make(map[string][]*api.VolumeSnapshotLocation)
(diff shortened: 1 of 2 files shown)


pkg/cmd/server/server.go
```diff
@@ -306,25 +301,34 @@ func getDefaultVolumeSnapshotLocations(arkClient clientset.Interface, namespace
...
+			for _, location := range locations {
+				// require a default VSL to be specified only if there’s more than one vsl for a given provider
+				if verifiedVolumeSnapshotLocations[provider] != nil {
+					return nil, errors.Errorf("a default volume snapshot location is required for %s since more than 1 location exists for provider %s", verifiedVolumeSnapshotLocations[provider].Name, provider)
+				}
+
+				verifiedVolumeSnapshotLocations[provider] = location
+			}
+		} else {
+			for _, location := range locations {
```
nrb: A small nitpick, L318 is extra whitespace.

pkg/cmd/server/server.go
```diff
@@ -289,15 +289,10 @@ func (s *server) run() error {
 	return nil
 }
 
-func getDefaultVolumeSnapshotLocations(arkClient clientset.Interface, namespace string, defaultVolumeSnapshotLocations map[string]string) (map[string]*api.VolumeSnapshotLocation, error) {
-	providerDefaults := make(map[string]*api.VolumeSnapshotLocation)
-	if len(defaultVolumeSnapshotLocations) == 0 {
-		return providerDefaults, nil
-	}
-
+func getVolumeSnapshotLocations(arkClient clientset.Interface, namespace string, defaultVolumeSnapshotLocations map[string]string) (map[string]*api.VolumeSnapshotLocation, error) {
```
nrb: Why the name change? It's still only getting the default, whether it's because it's the only one or one was defined.

pkg/cmd/server/server_test.go
```diff
@@ -80,37 +80,51 @@ func TestDefaultVolumeSnapshotLocations(t *testing.T) {
...
+	// No default fails when there’s more than one vsl for a that provider
+	defaultVolumeSnapshotLocations = make(map[string]string)
+	location3 := &v1.VolumeSnapshotLocation{ObjectMeta: metav1.ObjectMeta{Name: "location3"}, Spec: v1.VolumeSnapshotLocationSpec{Provider: "provider1"}}
+	arkClient.ArkV1().VolumeSnapshotLocations(namespace).Create(location3)
+	volumeSnapshotLocations, err = getVolumeSnapshotLocations(arkClient, namespace, defaultVolumeSnapshotLocations)
 	assert.Equal(t, 0, len(volumeSnapshotLocations))
 	assert.Error(t, err)
 
-	// Bad provider
+	// Non-existing provider given in a < 2 vsl scenario
```
nrb: Re-using the same `defaultVolumeSnapshotLocations` map as the above here was confusing to me. Could we clean up the state between our cases?

pkg/cmd/server/server_test.go
```diff
@@ -80,37 +80,51 @@ func TestDefaultVolumeSnapshotLocations(t *testing.T) {
...
+	defaultVolumeSnapshotLocations["provider1"] = "badlocation"
+	volumeSnapshotLocations, err = getVolumeSnapshotLocations(arkClient, namespace, defaultVolumeSnapshotLocations)
+	assert.Equal(t, 1, len(volumeSnapshotLocations))
```
nrb: A comment here as to what the 1 location return should be is helpful. I had to go to the beginning of the function to find that out.

pkg/cmd/server/server_test.go
```diff
@@ -80,37 +80,51 @@ func TestDefaultVolumeSnapshotLocations(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	// Non-existing provider given in a < 2 vsls scenario
```
nrb: Is this test verifying a provider with 0 valid VSLs does return 0? The comment here matches the case above.

skriss:
I think we should handle this a little bit differently:
- do basically no validation on `--default-volume-snapshot-locations` at server startup
- when a backup is created:
  - if VSLs are explicitly specified in the spec, use those (ensuring that they exist and there's at most one per provider specified)
  - for any provider without a VSL specified in the spec:
    - if there's only one VSL for that provider, use that
    - else, look for a default from the server flag (only now do we validate that it exists)
  - if we can't resolve to one VSL per provider here, have the backup fail validation

The nice things about this approach:
- it doesn't matter if VSLs exist at server startup time or not; they're validated just-in-time, when a backup is created
- the simple case is simple for the user: the "one provider, one VSL" scenario doesn't require setting a server flag or providing a VSL name in the backup spec
- an admin can delete a VSL and create a new one, and the server doesn't have to be restarted for this new one to become the default for its provider

WDYT?

ncdc:
@skriss I like these ideas

skriss:
Was thinking along these lines: https://github.com/heptio/ark/compare/master...skriss:vsl-defaulting?expand=1

@nrb @carlisia @wwitzel3 WDYT? If this approach makes sense, I can either polish this up or @carlisia you can take my code.

nrb:
@skriss That sounds good

pkg/cmd/server/server.go
carlisia: @nrb  VSLs don't have a preset default so if the user doesn't specify one, there will be none.

This makes it so in some cases what is returned is the default, and some cases not. This new name is more generic and less misleading, that was my thinking. Makes sense?

nrb: I don't think I follow - the only time it's used in code outside of tests, it's for default locations. Have you looked at Steve's alternative suggestion?

carlisia: To clarify: I don't think as the default as "the only one", I think of it as the feature that we currently allow to be set only via that `ark server` flag. This means the user doesn't have to specify the name of the VSL location for that provider. 

This is true for BSL. Is it incorrect for VSL?

carlisia:
Looking!

carlisia:
This sounds more user friendly and like it will allow for simpler documentation. I'm all for it. 👍

PR #998, opened 2018-10-25
Allow to use AWS Signature v1 for creating signed AWS urls

bashofmann:
Some s3 implementations, for example the quobyte object storage, do not
support the v4 signing algorithm, but only v1.
This makes it possible to configure the signatureVersion.

The algorithm implementation was ported from https://github.com/oNestLab/botocore/blob/d6c1be296e8cfe0706cb0c8bbcad9c095d0f4d09/botocore/auth.py#L860-L862
which is used by the aws CLI client.

This fixes https://github.com/heptio/ark/issues/811.

### DIFF
--- a/docs/support-matrix.md
+++ b/docs/support-matrix.md
@@ -20,6 +20,9 @@ _Note that these providers are not regularly tested by the Ark team._
  * [Minio][9]
  * Ceph RADOS v12.2.7
  * [DigitalOcean][7]
+ * Quobyte
+
+_Some storage providers, like Quobyte, may need a different [signature algorithm version][15]._
 
 ## Volume Snapshot Providers
 
@@ -52,3 +55,4 @@ After you publish your plugin, open a PR that adds your plugin to the appropriat
 [12]: https://github.com/aws/aws-sdk-go/aws
 [13]: https://portworx.slack.com/messages/px-k8s
 [14]: https://github.com/portworx/ark-plugin/issues
+[15]: api-types/backupstoragelocation.md#aws
--- a/pkg/cloudprovider/aws/object_store.go
+++ b/pkg/cloudprovider/aws/object_store.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/signer/v4"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/pkg/errors"
@@ -38,18 +39,28 @@ const (
 	kmsKeyIDKey         = "kmsKeyId"
 	s3ForcePathStyleKey = "s3ForcePathStyle"
 	bucketKey           = "bucket"
+	signatureVersionKey = "signatureVersion"
 )
 
 type objectStore struct {
-	log        logrus.FieldLogger
-	s3         *s3.S3
-	preSignS3  *s3.S3
-	s3Uploader *s3manager.Uploader
-	kmsKeyID   string
+	log              logrus.FieldLogger
+	s3               *s3.S3
+	preSignS3        *s3.S3
+	s3Uploader       *s3manager.Uploader
+	kmsKeyID         string
+	signatureVersion string
 }
 
 func NewObjectStore(logger logrus.FieldLogger) cloudprovider.ObjectStore {
 	return &objectStore{log: logger}
+}
+
+func isValidSignatureVersion(signatureVersion string) bool {
+	switch signatureVersion {
+	case "1", "4":
+		return true
+	}
+	return false
 }
 
 func (o *objectStore) Init(config map[string]string) error {
@@ -59,6 +70,7 @@ func (o *objectStore) Init(config map[string]string) error {
 		publicURL           = config[publicURLKey]
 		kmsKeyID            = config[kmsKeyIDKey]
 		s3ForcePathStyleVal = config[s3ForcePathStyleKey]
+		signatureVersion    = config[signatureVersionKey]
 
 		// note that bucket is automatically added to the config map
 		// by the server from the ObjectStorageProviderConfig so
@@ -99,6 +111,13 @@ func (o *objectStore) Init(config map[string]string) error {
 	o.s3 = s3.New(serverSession)
 	o.s3Uploader = s3manager.NewUploader(serverSession)
 	o.kmsKeyID = kmsKeyID
+
+	if signatureVersion != "" {
+		if !isValidSignatureVersion(signatureVersion) {
+			return errors.Errorf("invalid signature version: %s", signatureVersion)
+		}
+		o.signatureVersion = signatureVersion
+	}
 
 	if publicURL != "" {
 		publicConfig, err := newAWSConfig(publicURL, region, s3ForcePathStyle)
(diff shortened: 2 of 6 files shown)


bashofmann:
@nrb I just added documentation to `docs/api-types/backupstoragelocation.md`. Is this enough or is there also some other place where this should be mentioned?

@ncdc I could not find any v1 implementation in go, so I ported this myself from the python implementation mentioned above.

nrb:
>Is this enough or is there also some other place where this should be mentioned?

Thanks! It may also be worth calling out on, in case a provider doesn't support v4 signing. https://github.com/heptio/ark/blob/master/docs/support-matrix.md#s3-compatible-backup-storage-providers

bashofmann:
> Thanks! It may also be worth calling out on, in case a provider doesn't support v4 signing. https://github.com/heptio/ark/blob/master/docs/support-matrix.md#s3-compatible-backup-storage-providers

Done

pkg/cloudprovider/aws/object_store.go
```diff
@@ -207,5 +212,10 @@ func (o *objectStore) CreateSignedURL(bucket, key string, ttl time.Duration) (st
 		Key:    aws.String(key),
 	})
 
+	if o.signatureVersion == "1" {
```
ncdc: I think we may want to check for valid vs invalid values. If the user specifies anything other than 1, they get 4.

pkg/cloudprovider/aws/v1_sign_request_handler.go
```diff
@@ -0,0 +1,147 @@
+/*
+Copyright 2017 the Heptio Ark contributors.
```
ncdc: 2018

pkg/cloudprovider/aws/object_store.go
bashofmann: Like this https://github.com/heptio/ark/pull/998/commits/2cbe34f3af58e41be653662cfca55f038f148ced ?

pkg/cloudprovider/aws/object_store.go
```diff
@@ -36,25 +37,38 @@ const (
...
+	s3Uploader       *s3manager.Uploader
+	kmsKeyID         string
+	signatureVersion string
 }
 
 func NewObjectStore(logger logrus.FieldLogger) cloudprovider.ObjectStore {
 	return &objectStore{log: logger}
 }
 
+func IsValidSignatureVersion(signatureVersion string) bool {
```
ncdc: We probably don't need to make this an exported function - would you be ok making it private?

pkg/cloudprovider/aws/object_store.go
```diff
@@ -36,25 +37,38 @@ const (
...
+	switch signatureVersion {
+	case "1":
```
ncdc:
```suggestion
	case "1", "4":
```

pkg/cloudprovider/aws/v1_sign_request_handler.go
```diff
@@ -0,0 +1,147 @@
...
+	Logger      aws.Logger
+
+	Query        url.Values
+	stringToSign string
+	signature    string
+}
+
+// SignRequestHandler is a named request handler the SDK will use to sign
+// service client request with using the V4 signature.
+var V1SignRequestHandler = request.NamedHandler{
```
ncdc: Do we need to export this?

pkg/cloudprovider/aws/v1_sign_request_handler.go
```diff
@@ -0,0 +1,147 @@
...
+	Name: "v1.SignRequestHandler", Fn: SignSDKRequest,
+}
+
+func SignSDKRequest(req *request.Request) {
```
ncdc: Do we need to export this?

pkg/cloudprovider/aws/v1_sign_request_handler.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/aws/aws-sdk-go/aws/request"
+)
+
+var (
+	errInvalidMethod = errors.New("v1 signer only handles HTTP GET")
+)
+
+type signer struct {
+	// Values that must be populated from the request
+	Request     *http.Request
```
ncdc: Do these fields need to be exported?

pkg/cloudprovider/aws/v1_sign_request_handler.go
```diff
@@ -0,0 +1,147 @@
...
+	req.Error = v1.Sign()
+
+	if req.Error != nil {
+		return
+	}
+
+	req.HTTPRequest.URL.RawQuery = v1.Query.Encode()
+}
+
+func (v1 *signer) Sign() error {
```
ncdc: Does this need to be exported?

pkg/cloudprovider/aws/v1_sign_request_handler.go
```diff
@@ -0,0 +1,147 @@
...
+	credValue, err := v1.Credentials.Get()
+	if err != nil {
+		return err
```
ncdc:
```suggestion
		return errors.Wrap(err, "error getting credentials")
```

pkg/cloudprovider/aws/v1_sign_request_handler.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if req.Error != nil {
+		return
+	}
+
+	req.HTTPRequest.URL.RawQuery = v1.Query.Encode()
+}
+
+func (v1 *signer) Sign() error {
+	credValue, err := v1.Credentials.Get()
```
ncdc: Call this credentials?

pkg/cloudprovider/aws/v1_sign_request_handler.go
```diff
@@ -0,0 +1,147 @@
...
+	v1.Query.Set("AWSAccessKeyId", credValue.AccessKeyID)
+	if credValue.SessionToken != "" {
+		v1.Query.Set("SecurityToken", credValue.SessionToken)
+	}
+
+	// in case this is a retry, ensure no signature present
+	v1.Query.Del("Signature")
+
+	method := v1.Request.Method
+	//host := v1.Request.URL.Host
```
ncdc: Remove?

pkg/cloudprovider/aws/v1_sign_request_handler.go
```diff
@@ -0,0 +1,147 @@
...
+	path := v1.Request.URL.Path
+	if path == "" {
+		path = "/"
+	}
+
+	expires := strconv.FormatInt(v1.Time.Unix()+3600, 10)
```
ncdc: Use `ExpireTime` from the original request

pkg/cloudprovider/aws/v1_sign_request_handler.go
bashofmann: How about `credentialsValue`, `credentials` would clash with the imported package name

docs/support-matrix.md
```diff
@@ -14,12 +14,13 @@ Ark supports a variety of storage providers for different backup and snapshot op
 
 Ark uses [Amazon's Go SDK][12] to connect to the S3 API. Some third-party storage providers also support the S3 API, and users have reported the following providers work with Ark:
 
-_Note that these providers are not regularly tested by the Ark team._
+_Note that these providers are not regularly tested by the Ark team. Some storage providers, like Quobyte, may need a different [signature algorithm version][15]._
```
skriss: I think I'd move the note about Quobyte/v1 signature down to next to/under the Quobyte bullet.

pkg/cloudprovider/aws/object_store.go
```diff
@@ -114,6 +126,13 @@ func (o *objectStore) Init(config map[string]string) error {
 		o.preSignS3 = o.s3
 	}
 
+	if signatureVersion != "" {
+		if !isValidSignatureVersion(signatureVersion) {
+			return errors.Errorf("Invalid signature version: %s", signatureVersion)
```
skriss: Invalid -> invalid

ncdc:
@bashofmann LGTM here. You'll need to add an entry to `CHANGELOG.md` (new thing we're doing).

@skriss please do a final review.

Once Steve is 👍, please squash down to 1 commit and make sure it has the signoff. Thanks!

skriss:
LGTM pending `CHANGELOG.md entry`, signoff, and squash

PR #1002, opened 2018-10-25
Verify PV doesn't exist before creating new volume

### LINKED ISSUE #609: Restoring: better handling of snapshots/volumes if PV already exists
If a PV already exists in the cluster, trying to create a volume from the snapshot will succeed, but we won't be able to create the PV. We need better handling of this situation.

nrb:
Check to see if a PV for this volume exists before trying to create a new provider volume.

Fixes #609

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -694,30 +694,41 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 				continue
 			}
 
-			// restore the PV from snapshot (if applicable)
-			updatedObj, err := ctx.pvRestorer.executePVAction(obj)
-			if err != nil {
-				addToResult(&errs, namespace, fmt.Errorf("error executing PVAction for %s: %v", fullPath, err))
+			// Check if the PV exists in the cluster before attempting to create
+			// a volume from the snapshot, in order to avoid orphaned volumes (GH #609)
+			pv, err := resourceClient.Get(name, metav1.GetOptions{})
+			if err != nil && !apierrors.IsNotFound(err) {
+				addToResult(&errs, namespace, fmt.Errorf("error checking existence for PV %s: %v", name, err))
 				continue
 			}
-			obj = updatedObj
 
-			if resourceWatch == nil {
-				resourceWatch, err = resourceClient.Watch(metav1.ListOptions{})
+			// PV's existence will be recorded later. Just skip the volume restore logic.
+			if pv == nil {
+				// restore the PV from snapshot (if applicable)
+				updatedObj, err := ctx.pvRestorer.executePVAction(obj)
 				if err != nil {
-					addToResult(&errs, namespace, fmt.Errorf("error watching for namespace %q, resource %q: %v", namespace, &groupResource, err))
-					return warnings, errs
+					addToResult(&errs, namespace, fmt.Errorf("error executing PVAction for %s: %v", fullPath, err))
+					continue
 				}
-				ctx.resourceWatches = append(ctx.resourceWatches, resourceWatch)
-				ctx.resourceWaitGroup.Add(1)
-				go func() {
-					defer ctx.resourceWaitGroup.Done()
+				obj = updatedObj
 
-					if _, err := waitForReady(resourceWatch.ResultChan(), name, isPVReady, time.Minute, ctx.log); err != nil {
-						ctx.log.Warnf("Timeout reached waiting for persistent volume %s to become ready", name)
-						addArkError(&warnings, fmt.Errorf("timeout reached waiting for persistent volume %s to become ready", name))
+				if resourceWatch == nil {
+					resourceWatch, err = resourceClient.Watch(metav1.ListOptions{})
+					if err != nil {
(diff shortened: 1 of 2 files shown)


nrb:
@wwitzel3 fyi, https://github.com/heptio/ark/commit/195e6aaf000415583fa9aabac9bf529510279e39 introduced differences between `master` and `release-0.9`, but I think moving the logic over manually shouldn't be too hard. The test case updates will likely be the most work.

pkg/restore/restore.go
```diff
@@ -694,6 +694,18 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 				continue
 			}
 
+			// Check if the PV exists in the cluster before attempting to create
+			// a volume from the snapshot, in order to avoid orphaned volumes (GH #609)
+			pv, err := resourceClient.Get(name, metav1.GetOptions{})
+			if err != nil && !apierrors.IsNotFound(err) {
+				ctx.log.Errorf("Error checking for existence of PV %s: %s", name, err)
+				continue
```
skriss: Wondering if we should attempt the restore if we get an error here, rather than `continue`ing - WDYT?

pkg/restore/restore.go
```diff
@@ -694,6 +694,18 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+			}
+			if pv != nil {
+				ctx.log.Infof("PV %s already exists in cluster, not restoring from snapshot", name)
+				continue
```
skriss: Wondering if we should `continue` here, or *just* skip restoring the snapshot.  One downside of continuing is that the restore results won't show that the PV wasn't restored because it already existed.  WDYT?

pkg/restore/restore.go
nrb: Ah, good point - we really should continue with the restore. I'll update.

pkg/restore/restore.go
nrb: Agreed, should probably just skip the snapshot part and keep going with the restore.

pkg/restore/restore.go
ncdc: Please hold off for a minute

ncdc: If there's some other error besides not-found, it could be
- failure to reach the apiserver for some reason
- GET permission denied
- ???

The only one of these I'm concerned about, and hopefully this is extremely unlikely, is the ark serviceaccount is denied `GET persistentvolumes`. In this case, we probably shouldn't continue. WDYT?

pkg/restore/restore.go
wwitzel3: I'll update the backport to reflect this as well.

pkg/restore/restore.go
nrb: The permission denied would make sense. I assume that could be either of [`IsUnauthorized`](https://github.com/kubernetes/apimachinery/blob/release-1.11/pkg/api/errors/errors.go#L474) or [`IsForbidden`](https://github.com/kubernetes/apimachinery/blob/release-1.11/pkg/api/errors/errors.go#L480)? Is preferred over the other by the kube API server?

ncdc: I think I'm leaning more toward leaving the code as-is - if we can't successfully determine a PV's existence or non-existence, skip it.

pkg/restore/restore.go
nrb: If I'm understanding you correctly here @skriss, you're saying remove the `continue`, and make L709-732 an `else` branch of the `if pv != nil` condition? Looking at this again, wouldn't that be functionally equivalent, since everything in this block only applies to PersistentVolumes?

nrb: Actually, if we just call `addToResult` instead of `ctx.log.Infof`, would that solve the issue?

pkg/restore/restore.go
skriss: I'd like to have any errors show up the same way as for other resources (i.e. in the restore results).  I think as long as that's the case, either way is defensible.

pkg/restore/restore.go
skriss: if we `continue`, then L755+ don't get executed.  I don't necessarily see a reason *not* to execute them - yes, the PV will already exist, but I'd rather have that (the API object already existing) go through the same codepath as all the other resources, and keep this change *just* to the restoring of the snapshot.  Thoughts?

nrb: I believe the latest update addresses this. The error now shows up on the restore object, just like everything else, and we only execute snapshot restore logic if there's no pv already found.

pkg/restore/restore.go
```diff
@@ -30,6 +30,19 @@ import (
 	"sync"
 	"time"
 
+	api "github.com/heptio/ark/pkg/apis/ark/v1"
```
skriss: imports look like they're not using the changes andy made to `make update`

pkg/restore/restore_test.go
```diff
@@ -21,6 +21,16 @@ import (
 	"testing"
 	"time"
 
+	api "github.com/heptio/ark/pkg/apis/ark/v1"
```
skriss: ditto

nrb:
@skriss Fixed

Issue #978, opened 2018-10-22
Docs related issues from Ark v0.10 Testing


carlisia:
- [ ] This section has a reference to Config, which no longer exists https://github.com/heptio/ark/blob/master/docs/build-from-scratch.md#prerequisites-1
- [ ] This section instructs to setup multiple backets, but should we instruct to use the prefix instead? https://github.com/heptio/ark/blob/master/docs/aws-config.md#create-iam-user

Bradamant3:
Fixed in PR #1004

PR #1004, opened 2018-10-26
[WIP] address docs issue #978

Bradamant3:

starts to address issue #978. TODOs/Qs:

- reference to config object removed as requested, but we should now doc the new CRDs, yes? Will push another commit after we discuss

- there's also the IAM Q in the related issue. Not clear on the prefix option. Can/should we doc both? Need more info either way.


docs/build-from-scratch.md
```diff
@@ -93,11 +93,10 @@ When running Heptio Ark, you will need to account for the following (all of whic
 * Cloud provider credentials
   * Read/write access to volumes
   * Read/write access to object storage for backup data
-* A [Config object][8] definition for the Ark server
```
skriss: Let's change this line to say "A BackupStorageLocation object definition for the Ark server"
And add another line: "(Optional)" a VolumeSnapshotLocation object definition for the Ark server, if you want to be able to take PV snapshots"

skriss:
I think we can punt on the bucket/prefix/IAM question -- getting that exactly right would require some revisions to the IAM policies, and we'd want to potentially change the other providers too, and it's not wrong as-is.

skriss:
@Bradamant3 updates LGTM but this comments still needs to be addressed:

"L140 and L153 in build-from-scratch.md shoudl also update to point to the sample backupstoragelocation.yaml file, and below each we should also point to optionally setting up the volumesnapshotlocation.yaml file"

Bradamant3:
Addressed latest comments, but we need to come up with a cleaner version of the whole doc when we get a chance.

skriss:
LGTM - @Bradamant3 would it be possible to squash into a single commit before merge?

Bradamant3:
@skriss done with thanks for your old reminder to rebase i/o merge :D

PR #1013, opened 2018-10-29
Update LICENSE - Punctuation errors

midhunbiju:
Punctuation errors corrected


ncdc:
@midhunbiju thank you for your pull request, but this is the text of the Apache 2.0 license. We need it to match the original text.

PR #1014, opened 2018-10-29
add docs site for 0.10 beta release

Bradamant3:

For review: https://bradamant3.github.io/ark/v0.10.0-beta.1/

n.b. breaking changes warning in `index.html` is manually added; will add PR for master branch
please double-check TOC completeness


skriss:
Still poking around but this LGTM - @wwitzel3 @nrb @carlisia PTAL and see if all looks good (I'm viewing in Jennifer's fork per above link)

carlisia:
Looking..

_config.yml
```diff
@@ -1,8 +1,9 @@
 name: ark
 baseurl: /ark
 google_analytics: FILL_OUT
-latest: v0.9.0
```
nrb: Do we want latest to point to this still? Or only let people see it if they're specifically looking for it?

ncdc: We should continue to show 0.9 as the latest stable imho

Bradamant3: Only sorta done. It's taking me a little while to get the conditional logic right for a better version warning statement for 0.10.0-beta so I figure it's better to get at least a partial fix up first. Will continue to noodle (conditional logic brain evidently sleepy atm, also lazy)

Note that version warning logic doesn't work, but doesn't break.

Bradamant3:
See my comment in latest commit

_includes/version-warning.html
```diff
@@ -7,6 +7,15 @@
     <a href="{{latest_url}}">latest version</a>.
     </p>
   </div>
+{% elsif page.version == 'v.0.10.0-beta.1' %}
```
skriss: looks like you've got an extraneous `.` after `v`

Bradamant3: lolyep of course you know when I saw it ...

PR #1019, opened 2018-10-30
add troubleshooting for loadbalancer restores

Bradamant3:

See issue #264

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,6 +1,8 @@
 # Troubleshooting
 
-These tips can help you troubleshoot known issues. If they don't help, you can [file an issue][4], or talk to us on the [#ark-dr channel][25] on the Kubernetes Slack server. 
+These tips can help you troubleshoot known issues. If they don't help, you can [file an issue][4], or talk to us on the [#ark-dr channel][25] on the Kubernetes Slack server.
+
+## General troubleshooting information
 
 In `ark` version >= `0.1.0`, you can use the `ark bug` command to open a [Github issue][4] by launching a browser window with some prepopulated values. Values included are OS, CPU architecture, `kubectl` client and server versions (if available) and the `ark` client version. This information isn't submitted to Github until you click the `Submit new issue` button in the Github UI, so feel free to add, remove or update whatever information you like.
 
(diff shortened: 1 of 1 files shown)


Bradamant3:
Makes sense -- content moved.

docs/troubleshooting.md
```diff
@@ -32,6 +34,11 @@ kubectl edit deployment/ark -n heptio-ark
 ...
 ```
 
+## Known issue with restoring LoadBalancer Service
+
+Because of how Kubernetes handles Service objects of `type=LoadBalancer`, you might encounter an issue with changed values for Service UIDs. If the DNS CNAME for your application points to the DNS name of your cloud load balancer, you'll need to update the CNAME pointer when you perform an Ark restore.
```
skriss: "...objects of `type=LoadBalancer`, when restoring them, you might encounter an issue with changed DNS names for external load balancers.  Kubernetes automatically generates the name of the cloud resource based on the `Service` UID, which will be different when restored, resulting in a different name for the cloud load balancer.  If the DNS CNAME for..."

docs/troubleshooting.md
```diff
@@ -32,6 +34,11 @@ kubectl edit deployment/ark -n heptio-ark
...
+
+Alternatively, if your cloud provider supports an external load balancer, you might be able to use the Service's `spec.loadBalancerIP` field to keep connections valid. See [the Kubernetes documentation about Services of Type LoadBalancer](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer). For more information, see [the documentation about how to create an External Load Balancer while preserving the client source IP](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip).
```
skriss: "Alternatively, you might be able to use the Service's `spec.loadBalancerIP` field if your cloud provider supports it to keep connections valid."

Also, that last sentence/link is not relevant here so can be removed.

docs/troubleshooting.md
skriss: It changes when you run the restore -- when the `Service` obj is applied, it gets a new UID, and that's used to name the corresponding load balancer that gets created

skriss:
@Bradamant3 I'll merge once you squash

PR #1020, opened 2018-10-30
fix goreleaser bugs

skriss:

- use `.Tag` instead of `.Version` to include the leading `v` in both artifact names and ldflags.
- fix issue when `$SKIP_PUBLISH` is empty.


skriss:
@wwitzel3 PTAL when you get a chance, thanks

PR #1022, opened 2018-10-30
Remove code that checks backup storage directly from restore controller

### LINKED ISSUE #945: restore controller: remove code that checks backup storage directly for a backup
If the restore controller can't find the requested backup in the Kube API, it checks backup storage directly.  This was implemented primarily for the scenario where Cluster B was importing backups from Cluster A, and Cluster B hadn't run a sync loop since the desired backup was taken in Cluster A.

With the sync controller changes in #825, the sync process now runs every minute.  The issue described above should no longer be material.  I'm proposing removing the code in the restore controller that checks backup storage directly, for simplification.

shubheksha:
Fixes #945

### DIFF
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -28,7 +28,6 @@ import (
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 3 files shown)


nrb:
@shubheksha I believe L235 should stay; it's mean to replicate https://github.com/heptio/ark/blob/master/pkg/controller/restore_controller.go#L581 in tests, though I do notice the error texts don't match.

I think L236 should stay too, or at least not be removed in this PR. It does look like there's only 1 test case it affects currently, though.

shubheksha:
@nrb, sounds good! I'll not remove those. Does everything else look good? I wasn't able to figure out why exactly the error texts don't match though.

carlisia:
> I wasn't able to figure out why exactly the error texts don't match though.

@shubheksha this is why:
image

Both before and after your change, the error when you are at LN 404 `IsNotFound`, so it skips that check and moves to LN 408. 

Before your change, the error returned came from `fetchFromBackupStorage`, and the clue for the reason why for that error is in the test name: `restore with non-existent backup name fails` (which would've best been named `restore attempt with non-existent backup name fails`.

After your change, it was simply a `not found` error since we are no longer attempting to fetch the restore from the backup storage.

pkg/controller/restore_controller.go
```diff
@@ -426,29 +426,6 @@ func (c *restoreController) fetchBackupInfo(backupName string, pluginManager plu
...
-		}
-		return info, nil
-	}
-
-	return backupInfo{}, errors.New("not able to fetch from backup storage")
-}
-
 // orderedBackupLocations returns a new slice with the default backup location first (if it exists),
 // followed by the rest of the locations in no particular order.
 func orderedBackupLocations(locations []*api.BackupStorageLocation, defaultLocationName string) []*api.BackupStorageLocation {
```
skriss: this function is now only used in `backup_sync_controller.go` so I'd probably move it there

pkg/controller/restore_controller.go
```diff
@@ -406,8 +406,8 @@ func (c *restoreController) fetchBackupInfo(backupName string, pluginManager plu
 		}
 
 		log := c.logger.WithField("backupName", backupName)
-		log.Debug("Backup not found in backupLister, checking each backup location directly, starting with default...")
-		return c.fetchFromBackupStorage(backupName, pluginManager)
+		log.Debug("Backup not found in backupLister")
```
skriss: the godoc for this function should be updated (L399-400) -- can just say that it gets the backup API obj from the lister and a backup store for the backup's storage location

skriss: L404-409 can be removed - any error when trying to get the backup from the lister should now be returned

skriss:
@shubheksha you can squash it to 1 commit so it's ready to go :) thanks again!

nrb:
Seems like it's removing unused/unexecuted code, so I don't see a reason to keep it out of v0.10.

shubheksha:
@skriss, done!

carlisia:
This sounds safe enough to include in v0.10.

PR #1026, opened 2018-10-31
tweak version to accommodate any beta

Bradamant3:

Might want to move to case statement but first efforts were unsuccessful (not sure whether liquid, but only simplest conditions seem to be allowed)

Thanks to @MischaDeNola for both suggestions!


skriss:
@Bradamant3 were you able to test and confirm this works for our current beta?

Bradamant3:
@skriss yes indeed tested with current version

PR #999, opened 2018-10-25
AWS: Ensure that the order returned by ListObjects is consistent

bashofmann:
When a backup is deleted, the delete method uses ListObjects to get a list of
files it needs to delete in s3. Different s3 implementations may return
the object lists in different, even non-deterministic orders, which can
result in the deletion not working because ark tries to delete a non empty folder
before it tries to delete the files in the folder.

Signed-off-by: Bastian Hofmann <bashofmann@gmail.com>

PR #971, opened 2018-10-21
Restore storageclasses before pvs and pvcs

shubheksha:
Signed-off-by: Shubheksha Jalan <jshubheksha@gmail.com>
Fixes #594

Issue #594, opened 2018-06-25
Restore storageclasses before persistentvolumes and persistentvolumeclaims


ncdc:
This is necessary so the PVs and PVCs can be created correctly.

shubheksha:
I'd like to pick this up! @ncdc, can I get some more info to get started on it?

ncdc:
@shubheksha sure! We have default restore priorities here: https://github.com/heptio/ark/blob/d7dfffa37358584d66e2d433cd79c7b8f2f43f4e/pkg/cmd/server/server.go#L456-L475. Need to insert storageclasses in the right place and test it.

PR #1028, opened 2018-11-01
0.9 backports

wwitzel3:
backport of #962 (a fix for #424)
backport of #999 (fix for s3 compatible object list ordering)
backport of #971 (a fix for #594)


wwitzel3:
Updated to include #971 
@skriss and @nrb PTAL

nrb:
Will need the CHANGELOG entries here and on master, too.

Issue #965, opened 2018-10-19
Enable setting metadata.ownerReference in RestoreItemAction plugin


mwieczorek:
**Describe the solution you'd like**
I'd like to backup/restore resources related to [service catalog](https://github.com/kubernetes-incubator/service-catalog).
When you create ServiceBinding (CRD from service catalog) there's also created a secret with ownerReference set to that ServiceBinding.

When I restore the secret on another cluster the ownerReference is deleted ( [code](https://github.com/heptio/ark/blob/master/pkg/restore/restore.go#L1007) ).

I'd like to be able to restore the secret with ownerReference set.

I don't want to preserve the old ownerReference because the uid of restored SB is different than from the origin cluster.

I'd like to set the new one, f.e. by plugin. But looking through the code I think it's not possible now because the resetMetadata is after the plugin execution, am I right?

Is it possible to move plugin execution after the metadata reset?

**Anything else you would like to add:**
Ref: [slack thread](https://kubernetes.slack.com/archives/C6VCGP4MT/p1539952674000100)

One more 'nice to have': 
is it possible to pass to plugin not only object with deleted metadata but also original object read from the backup? 
If I get object with deleted metadata I have to rely on secret's name (which I think should be equal to SB name) but if I get object from the backup I can get old ownerReference.

**Environment:**

- Ark version (use `ark version`): master
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

ncdc:
> is it possible to pass to plugin not only object with deleted metadata but also original object read from the backup?

If we move the execution of the restore item actions first, before resetting metadata, that would do what you need, wouldn't it?

mwieczorek:
> > is it possible to pass to plugin not only object with deleted metadata but also original object read from the backup?
> 
> If we move the execution of the restore item actions first, before resetting metadata, that would do what you need, wouldn't it?

In such case I have access to the original object, but even if I change the uid in ownerReference than it will be deleted in resetMetadata step.

ncdc:
Right. It's times like this when having all the objects in memory at the same time, prior to restoration, would be useful. But Ark isn't currently set up that way, so the options I can think of are:

1. Run restore item actions, then reset metadata (don't strip ownerReferences)
1. Reset metadata, then run restore item actions (modify interface as you suggested, to include original object's state too)
[...]

PR #1032, opened 2018-11-02
Move clearing up of metadata before plugin's actions

mwieczorek:
As discussed in #965 and in this [slack thread](https://kubernetes.slack.com/archives/C6VCGP4MT/p1540932054141300):
I'd like to move clearing up of metadata before plugin execution. In such case I will be able to set ownerReference.

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -740,6 +740,12 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 			}
 		}
 
+		// clear out non-core metadata fields & status
+		if obj, err = resetMetadataAndStatus(obj); err != nil {
+			addToResult(&errs, namespace, err)
+			continue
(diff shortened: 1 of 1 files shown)


ncdc:
@mwieczorek does this give you what you need for your plugin to be able to function?

mwieczorek:
@ncdc For one of them (serviceBinding -> secret) yes. For others I may need original resource without deleted metadata as a func param. 
But I understand you’d like to think about change of plugin interface and not rush with the implementation.

ncdc:
No objections here

ncdc:
At some point we could/should unit test?

skriss:
@ncdc did you want units added to this PR?

ncdc:
I’m ok without them. I do still want to overhaul the entire package to make
it more maintainable and testable. Obviously not a part of this though.

PR #1038, opened 2018-11-07
add draft content for docs issue 819

### LINKED ISSUE #819: Ark Troubleshooting Two Clusters

Update troubleshooting documentation to verify two clusters are replicating as expected. If the back ups are configured to use the same bucket and the first cluster is showing successfull backups the next step would be to  get the output of `ark restore get` from the second cluster and then `ark restore describe <last restore>`

**Environment:**

- Ark version (use `ark version`): 0.9.3
- Kubernetes version (use `kubectl version`): 1.10
- Kubernetes installer & version: Kops
- OS (e.g. from `/etc/os-release`): Debian Jesse

Bradamant3:

Fixes #819 

@rbankston is this what you had in mind? File location is not as specified in original issue, but I think it's more effective here.

Note also that this PR pulls apart the old use case topic as a start at reorganizing use cases/tutorials. Not fully addressing any other docs reorg issue yet, though.


docs/migration-case.md
```diff
@@ -0,0 +1,44 @@
...
+Heptio Ark can help you port your resources from one cluster to another, as long as you point each Ark instance to the same cloud object storage location. In this scenario, we are also assuming that your clusters are hosted by the same cloud provider. **Note that Heptio Ark does not support the migration of persistent volumes across cloud providers.**
+
+1.  *(Cluster 1)* Assuming you haven't already been checkpointing your data with the Ark `schedule` operation, you need to first back up your entire cluster (replacing `<BACKUP-NAME>` as desired):
+
+    ```
+    ark backup create <BACKUP-NAME>
+    ```
+    The default TTL is 30 days (720 hours); you can use the `--ttl` flag to change this as necessary.
+
+1.  *(Cluster 2)* Make sure that the `BackupStorageLocation` and `VolumeSnapshotLocation` CRDs match the ones from *Cluster 1*, so that your new Ark server instance points to the same bucket.
```
skriss: we should also have them put cluster 2 in restore-only mode (using the server flag) for this

docs/migration-case.md
```diff
@@ -0,0 +1,44 @@
...
+
+1.  *(Cluster 2)* Make sure that the Ark Backup object is created. Ark resources are synchronized with the backup files in cloud storage.
```
skriss: we should note that the default sync interval is 1 minute (as of v0.10), so they can expect a delay of up to that in the backups being imported into cluster 2. And they can adjust that with a server flag if they like

docs/migration-case.md
ncdc: I had always assumed (and apparently never communicated) that "restore only mode" would move from the `Config` to each `BackupStorageLocation` and that you could enable/disable with updates to a BSL. How do you feel about that?

skriss: Yeah, we were planning to do that in v0.11 as part of the ownership/sharing epic

skriss: Since we're pausing on the replication train though, we'll need to decide if we want to do any parts of those epics for v1.0.  They weren't in our issue list since we hadn't broken them out yet.

ncdc: Makes sense, thanks

skriss:
@Bradamant3 this LGTM, can you squash down to 1 commit and I'll merge?

skriss:
@Bradamant3 reminder, just waiting on you to squash here.

PR #1040, opened 2018-11-08
Fix check for non-found PV

ncdc:
We were checking for nil, but were getting back an empty
*unstructured.Unstructured{} instead, along with a NotFound error.
Change the logic to check for the NotFound error instead of a nil
object.


ncdc:
Will need to backport to 0.9.x too

ncdc:
I have a backport ready and will open a PR after this is merged

skriss:
looks fine to merge - ready to go ahead?

ncdc:
Yes, with this fix, my restore actually restored the PV. Without the fix, it did not.

ncdc:
Feel free to test it too!

ncdc:
Backport: #1041

skriss:
yep, looks good.

PR #1041, opened 2018-11-08
[0.9] Fix check for non-found PV

ncdc:
We were checking for nil, but were getting back an empty
*unstructured.Unstructured{} instead, along with a NotFound error.
Change the logic to check for the NotFound error instead of a nil
object.


ncdc:
Test image up at `ncdc/ark:pv-fix` if you want to test it

PR #1023, opened 2018-10-30
add rbac content and a candidate for more versioning information

Bradamant3:
Signed-off-by: JENNIFER RONDEAU <jrondeau@heptio.com>

- fixes #350 
- fixes #931 
- fixes #387 
- fixes #839 

Also cleans up a fewe other small usability/style issues not reported separately

Issue #952, opened 2018-10-18
Update install docs to use `velero install` instead of in-tree YAML


skriss:
The install docs for our various supported platforms sometimes differ from each other for no apparent reason.  For example, for Azure we `kubectl apply` the entire example folder, whereas for GCP we don't.  

Additionally, we have a legacy numbering scheme on our YAML files that probably isn't necessary anymore and may just lead to confusion/annoyance.  Look at possibly removing that. 

cc @carlisia

carlisia:
Good point. This ticket could be the organization phase of moving towards an install.

nrb:
Looking at what I may do for #52, I'm thinking the following:

* YAML examples will be entirely from the repository. If YAML is needed, users can run `ark install --dry-run -o yaml` (exact args and flags not necessarily final)
* Each provider's install invocation will be with instructions for preparing the necessary provider-specific steps. I'm thinking that doing provider set up like creating buckets and IAM roles/policy is out of scope for `ark install` - it just focuses on creating the CRDs + necessary objects in the cluster, plus the deployment. I'm not sure if I care where that goes, in light of the larger docs refactor.

@Bradamant3 The two points above seem fairly minimal to me. I can discuss further, but I think those are my main concerns at the moment.

skriss:
Now that we have `velero install`, we'll standardize by updating the install docs to use this command, and largely doing away with static sample YAML.

skriss:
Can we close this issue now?

PR #1045, opened 2018-11-08
package version-specific example YAML with each release

skriss:

This updates the goreleaser process so it adjusts the image tag in each example YAML file to match the version of Ark being released.

Open to suggestions on the name of the directory that contains the examples within the archives - just can't be "examples" (unless we change the existing directory name in the repo) since goreleaser requires a clean git state before running.

I'll move on to updating our docs next.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 dist: _output
+before:
+  hooks:
+    - ./hack/set-example-tags.sh
 builds:
   - main: ./cmd/ark/main.go
     env:
@@ -41,7 +44,7 @@ archive:
   name_template: "{{ .ProjectName }}-{{ .Tag }}-{{ .Os }}-{{ .Arch }}"
   files:
     - LICENSE
-    - examples/**/*
+    - config/**/*
 checksum:
   name_template: 'CHECKSUM'
 snapshot:
--- a/hack/set-example-tags.sh
+++ b/hack/set-example-tags.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Copyright 2018 the Heptio Ark contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# this script copies all of the files under examples/ into a new directory,
+# config/ (which is gitignored so it doesn't result in the git
+# state being dirty, which would prevent goreleaser from running), and then
+# updates all of the image tags in those files to use $GIT_SHA (which will
+# be the release/tag name).
+
+rm -rf config/ && cp -r examples/ config/
+
+# the "-i'.bak'" flag to sed is necessary, with no space between the flag
+# and the value, for this to be compatible across BSD/OSX sed and GNU sed.
+# remove the ".bak" files afterwards (they're copies of the originals).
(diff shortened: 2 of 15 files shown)


hack/set-example-tags.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# this script copies all of the files under examples/ into a new directory,
+# release-examples/ (which is gitignored so it doesn't result in the git
+# state being dirty, which would prevent goreleaser from running), and then
+# updates all of the image tags in those files to use $GIT_SHA (which will
+# be the release/tag name).
+
+rm -rf release-examples/ && cp -r examples/ release-examples/
+
+find release-examples -type f -name "*.yaml" | xargs sed -i '' "s|gcr.io/heptio-images/ark:latest|gcr.io/heptio-images/ark:$GIT_SHA|g"
```
ncdc: There's also `gcr.io/heptio-images/fsfreeze-pause:latest` in the nginx with pv example. Maybe we have the regex do a capture group for what comes after `gcr.io/heptio-images/` but before `:latest`?

ncdc: Oh wait, nevermind... wasn't thinking that that's not per ark version.

ncdc: Does this `sed` work on linux and mac? I had a lot of trouble getting something that was cross platform.

ncdc:
Maybe `config` for the dir name?

hack/set-example-tags.sh
skriss: good q, will check on linux

skriss: Actually a/o v0.10 we will be tagging this image per release, so your comment is valid.

ncdc: Win! 😄

skriss:
latest commit addresses all review comments.  I opted to just add a second `sed` for the `fsfreeze-pause` image for clarity.

.goreleaser.yml
```diff
@@ -41,7 +44,7 @@ archive:
   name_template: "{{ .ProjectName }}-{{ .Tag }}-{{ .Os }}-{{ .Arch }}"
   files:
     - LICENSE
-    - examples/**/*
+    - release-examples/**/*
```
ncdc: config?

hack/set-example-tags.sh
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# this script copies all of the files under examples/ into a new directory,
+# release-examples/ (which is gitignored so it doesn't result in the git
```
ncdc: release-examples->config?

ncdc:
Is `$GIT_SHA` really the git tag?

skriss:
`export GIT_SHA=$(git describe --tags --always)` -- so yes, if the commit has been tagged. the whole goreleaser process fails on a non-tagged commit anyway. 

May simplify this per prior discussion on removing/changing some of the things we show in `ark version`.

skriss:
Fixed issues, added commit with docs changes

nrb:
I think it would be helpful to update https://github.com/heptio/ark/blob/master/examples/README.md as well, so that it mentions that the files will be put into the tarball, and the tarred files are preferred. Mostly as a pointer for people who may go straight to the GitHub code site for the YAML.

skriss:
@nrb thanks for the comment, updated.

skriss:
FIxed conflict. @nrb was this LGTY apart from the one comment?

skriss:
Let's hold off on merging this until #1023 goes in since there's likely to be a conflict there.

skriss:
Agree the numbering should be revisited, I think #952 was opened to address that.

nrb:
Yeah, I'd say we can revisit the numbering a bit later. If I remember correctly, it was to help get the order right when doing `kubectl apply -f <directory>`, since `kubectl` sorted in alpha-numeric order.

carlisia:
True! I forgot about that.

skriss:
@nrb or @carlisia I've rebased this now that #1023 has gone in (just a couple conflicts in docs), so please merge when green. thanks!

PR #1048, opened 2018-11-09
Remove default token from all service accounts

### LINKED ISSUE #1047: Non-existing ServiceAccounts need to have their default token removed
If we try to create a ServiceAccount and it already exists in the cluster, we remove the default token secret (it's autogenerated by Kubernetes), merge the secrets, image pull secrets, labels, and annotations from the item in the backup with the item in the cluster, and attempt to patch it.

If we try to create a ServiceAccount and it does **not** already exist in the cluster, we are not removing the default token secret, and the result is a ServiceAccount that references the old default token (from the backup) that doesn't exist in the new cluster, and the newly generated default token in the new cluster.

One possible fix is to move the logic that removes the default token from `mergeServiceAccounts` (which is only called in the already-exists flow) to a new restore item action that runs for all service accounts.

ncdc:
Instead of only removing the default token from a service account when
it already exists in the cluster, always remove it. If the service
account already exists, continue to do the merging logic.

Fixes #1047

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,9 @@
   * Delete spec.priority in pod restore action (#879, @mwieczorek)
   * Added brew reference (#1051, @omerlh)
   * Update to go 1.11 (#1069, @gliptak)
-  * Initialize empty schedule metrics on server init (#1054, @cbeneke) 
+  * Initialize empty schedule metrics on server init (#1054, @cbeneke)
   * Update CHANGELOGs (#1063, @wwitzel3)
+  * Remove default token from all service accounts (#1048, @ncdc)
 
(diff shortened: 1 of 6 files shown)


ncdc:
Will fix the imports issue - 1 sec

ncdc:
Hopefully travis will be happy now!

ncdc:
@skriss I don't remember if I tested this in minikube or not. I think I did but am not 100% sure.

skriss:
@ncdc now that @wwitzel3's changelog PR has merged you'll need to add one for this PR.

ncdc:
@skriss changelog updated. Tested just now on minikube.

PR #1049, opened 2018-11-09
[WIP] Plumb the log level through to plugin processes

ncdc:

### DIFF
--- a/pkg/plugin/logger.go
+++ b/pkg/plugin/logger.go
@@ -24,8 +24,9 @@ import (
 
 // NewLogger returns a logger that is suitable for use within an
 // Ark plugin.
-func NewLogger() logrus.FieldLogger {
+func NewLogger(level logrus.Level) logrus.FieldLogger {
 	logger := logrus.New()
+	logger.Level = level
 	/*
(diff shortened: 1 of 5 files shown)


ncdc:
This only does it currently for the Ark internal plugins. I think we should do it for external plugins too.

pkg/plugin/logger.go
```diff
@@ -24,8 +24,9 @@ import (
 
 // NewLogger returns a logger that is suitable for use within an
 // Ark plugin.
-func NewLogger() logrus.FieldLogger {
+func NewLogger(level logrus.Level) logrus.FieldLogger {
```
ncdc: This is a breaking change to plugins and will require an update in the ark-plugin-example repo

ncdc:
We potentially could update `arkplugin.NewServer` to make it support `--log-level`

nrb:
@disrani-px @edevenport This change will be coming in v0.10.1 and will require another update to plugins once we merge it.

king-jam:
This going to be merged in v0.11.1 or v0.12.1?

skriss:
@king-jam probably `v0.12.0`

PR #1050, opened 2018-11-09
add doc explaining locations

skriss:

Still a WIP but dumped as much as I could for today.  Is this a useful document?  What's missing or unclear?  

If we all like this then I'd like to link to it prominently, in the release notes, probably the blog post, etc., so that folks can read it and get the background before diving into upgrading.

cc @Bradamant3 @nrb @wwitzel3 @carlisia @jrnt30

### DIFF
--- a/docs/locations.md
+++ b/docs/locations.md
@@ -1,0 +1,168 @@
+# Backup Storage Locations and Volume Snapshot Locations
+
+Ark v0.10 introduces a new way of configuring where Ark backups and their associated persistent volume snapshots are stored.
+
+## Motivations
+
+In Ark versions prior to v0.10, the configuration for where to store backups & volume snapshots is specified in a `Config` custom resource. The `backupStorageProvider` section captures the place where all Ark backups should be stored. This is defined by a **provider** (e.g. `aws`, `azure`, `gcp`, `minio`, etc.), a **bucket**, and possibly some additional provider-specific settings (e.g. `region`). Similarly, the `persistentVolumeProvider` section captures the place where all persistent volume snapshots taken as part of Ark backups should be stored, and is defined by a **provider** and additional provider-specific settings (e.g. `region`).
+
+There are a number of use cases that this basic design does not support, such as:
+
+- Take snapshots of more than one kind of persistent volume in a single Ark backup (e.g. in a cluster with both EBS volumes and Portworx volumes)
+- Have some Ark backups go to a bucket in an eastern USA region, and others go to a bucket in a western USA region
+- For volume providers that support it (e.g. Portworx), have some snapshots be stored locally on the cluster and have others be stored in the cloud
+
+Additionally, as we look ahead to backup replication, a major feature on our roadmap, we know that we'll need Ark to be able to support multiple possible storage locations.
+
+## Overview
+
+In Ark v0.10 we got rid of the `Config` custom resource, and replaced it with two new custom resources, `BackupStorageLocation` and `VolumeSnapshotLocation`. The new resources directly replace the legacy `backupStorageProvider` and `persistentVolumeProvider` sections of the `Config` resource, respectively. 
+
+Now, the user can pre-define more than one possible `BackupStorageLocation` and more than one `VolumeSnapshotLocation`, and can select *at backup creation time* the location in which the backup and associated snapshots should be stored. 
+
+A `BackupStorageLocation` is defined as a bucket, a prefix within that bucket under which all Ark data should be stored, and a set of additional provider-specific fields (e.g. AWS region, Azure storage account, etc.) The [API documentation][1] captures the configurable parameters for each in-tree provider.
+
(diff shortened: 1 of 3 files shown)


skriss:
I rearranged based on @jrnt30's feedback, and added some more color to the overview. Still need to fill out some details in the additional use cases section.  PTAL if anyone has time!

docs/locations.md
```diff
@@ -0,0 +1,165 @@
+# Backup Storage Locations and Volume Snapshot Locations
+
+Ark v0.10 introduces a new way of configuring where Ark backups and their associated persistent volume snapshots are stored.
+
+## Motivations
+
+In Ark prior to v0.10, the configuration for where to store backups & volume snapshots was stored in a `Config` custom resource. The `backupStorageProvider` section captured the place where all Ark backups should be stored. This was defined by a **provider** (e.g. `aws`, `azure`, `gcp`, `minio`, etc.), a **bucket**, and possibly some additional provider-specific settings (e.g. `region`). Similarly, the `persistentVolumeProvider` section captured the place where all persistent volume snapshots taken as part of Ark backups should be stored, and was defined by a **provider** and additional provider-specific settings (e.g. `region`).
```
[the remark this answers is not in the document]
carlisia: I agree with @wwitzel3 on the tense. Because that version is not deprecated, it would sound more natural to talk about it in the present term because it is still in use.

docs/locations.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Limitations / Caveats
+
+- For volume snapshots, you are still limited by where your provider allows you to create snapshots. For example, AWS and Azure do not allow you to create a volume snapshot in a different region than where the volume is. If you try to take an Ark backup using a volume snapshot location with a different region than where your cluster's volumes are, the backup will fail.
+
+- Each Ark backup has one `BackupStorageLocation`, and one `VolumeSnapshotLocation` per volume provider. It is not possible (yet) to send a single Ark backup to multiple backup storage locations simultaneously, and similarly for volume snapshots. However, you can always set up multiple scheduled backups that differ only in the storage locations used if redundancy of backups across locations is important.
+
+- Cross-provider snapshots are not supported. If you have a cluster with more than one type of volume (e.g. EBS and Portworx), but you only have a `VolumeSnapshotLocation` configured for EBS, then Ark will **only** snapshot the EBS volumes.
```
jrnt30: Trying to clear up a bit to my current understanding (which may be off) which is that you can have a backup span multiple VSLs, they are just constrained by the volume's underlying provider.

```suggestion
- Cross-provider snapshots are not supported. If you have a cluster with more than one type of volume then Ark will **only** snapshot the volumes for their underlying provider. As an example, if we had a cluster with Portworx and EBS volumes, and a `VolumeStorageLocation` defined for each provider, the EBS volumes would only be snapshot by the EBS `VolumeStorageLocation` and the Portworx would only be snapshot by the Portworx `VolumeStorageLocation`.
```

docs/locations.md
```diff
@@ -0,0 +1,165 @@
...
+
+ark snapshot-location create portworx-cloud \
+    --provider portworx \
+    --config type=cloud
+```
+
+During backup creation:
+
+```shell
+# Note that since in this example we have two possible volume snapshot locations for the Portworx
```
jrnt30: Is it worth calling out anywhere that the server also has some additional flags now to indicate which of these  VSL (provider/location tuple) should be the default?

docs/locations.md
```diff
@@ -0,0 +1,168 @@
+# Backup Storage Locations and Volume Snapshot Locations
+
+Ark v0.10 introduces a new way of configuring where Ark backups and their associated persistent volume snapshots are stored.
+
+## Motivations
+
+In Ark versions prior to v0.10, the configuration for where to store backups & volume snapshots is stored in a `Config` custom resource. The `backupStorageProvider` section captures the place where all Ark backups should be stored. This is defined by a **provider** (e.g. `aws`, `azure`, `gcp`, `minio`, etc.), a **bucket**, and possibly some additional provider-specific settings (e.g. `region`). Similarly, the `persistentVolumeProvider` section captures the place where all persistent volume snapshots taken as part of Ark backups should be stored, and is defined by a **provider** and additional provider-specific settings (e.g. `region`).
```
carlisia:
```suggestion
In Ark versions prior to v0.10, the configuration for where to store backups & volume snapshots is specified in a `Config` custom resource. The `backupStorageProvider` section captures the place where all Ark backups should be stored. This is defined by a **provider** (e.g. `aws`, `azure`, `gcp`, `minio`, etc.), a **bucket**, and possibly some additional provider-specific settings (e.g. `region`). Similarly, the `persistentVolumeProvider` section captures the place where all persistent volume snapshots taken as part of Ark backups should be stored, and is defined by a **provider** and additional provider-specific settings (e.g. `region`).
```

carlisia: Super minor change just to avoid repeating the concept of "storage" and perhaps avoid confusion.

skriss: thanks, updated!

skriss:
@carlisia @wwitzel3 if y'all are happy with this, let's get it merged. we can refine going fwd as we field user questions post-v0.10.

PR #1054, opened 2018-11-14
Initialize empty schedule metrics on server init

cbeneke:
When backups are run manually (outside of a schedule) the metrics will be counted for `ark_*{schedule=""}`. To prevent partial `NaN` metrics they will be initialised on server init.

The current behaviour creates e.g. only `ark_backup_failure_total` or `ark_backup_success_total` but never both on the first manual run. Afaict the `backupName` shall be included in the metrics in the future, when this is implemented the initialisation must be moved to the point in time, where the `backupName` is already known (I guess the `processBackup()` function in `pkg/controller/backup_controller.go`), but so far the server init should be most sufficient for the metrics init.


nrb:
Thanks for this @cbeneke! I'll take a closer look this afternoon, but my initial thought is that this makes sense.

Issue #1051, opened 2018-11-12
Add brew cask to install cli?


omerlh:
[Brew](https://brew.sh/) is used to installing various utilities on macos, it will be nice to use it to install the CLI.

wiegandf:
https://formulae.brew.sh/formula/ark

ncdc:
Thanks @wiegandf for pointing out the existing formula.

@skriss we might want to submit a PR against the brew formula for ark to add the yaml.

omerlh:
Thanks! I can confirm it's working. Added a quick note to the docs - what do you think?

PR #1056, opened 2018-11-14
Added brew reference

omerlh:
Based on the answer to #1051, I thought it will be helpful for others to know about the formula :)

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -50,6 +50,13 @@ NOTE: Make sure to check out the appropriate version. We recommend that you chec
 
 Make sure that you install somewhere in your PATH.
 
+#### MacOS
+
+On Mac you can use HomeBrew to install ark CLI, by running the following command:
+```
+brew install ark
+```
+
 ### Back up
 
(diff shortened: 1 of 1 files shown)


nrb:
Thanks @omerlh! Would you mind amending the commit and adding a [sign off](https://git-scm.com/docs/git-commit#git-commit---signoff)?

omerlh:
Done

nrb:
LGTM, I'll let @Bradamant3 be the 2nd review.

Bradamant3:
The only potential issue I see here is that this PR is against the 0.9 branch. We should probably take it against master, yes? (I can then backport it to the other versions for the doc site if you think it would be useful.)

nrb:
Good catch, @Bradamant3. @omerlh mind doing this change against master and re-pushing? I changed in the GH UI, but unfortunately that included a ton of release-0.9 changes, so it'll need to be done against master manually.

omerlh:
I think I used this branch because it's the default, I'll fix it

skriss:
Closing this PR as it's been replaced by #1065

PR #1058, opened 2018-11-14
add docs site for v0.10.0

Bradamant3:

Here it is, folks. Links all checked, but double-checking is always an excellent thing (yes, we should automate this :D )

Built site to check: https://bradamant3.github.io/ark/


_config.yml
```diff
@@ -1,8 +1,9 @@
 name: ark
 baseurl: /ark
 google_analytics: FILL_OUT
-latest: v0.9.0
+latest: v0.10.0
 versions:
+- v0.10.0
 - v0.10.0-beta.1
```
skriss: I think we can drop all refs to the beta

_config.yml
```diff
@@ -30,6 +31,11 @@ defaults:
   values:
     version: master
     gh: https://github.com/heptio/ark/tree/master
+- scope:
+    path: v0.10.0
+  values:
+    version: v0.10.0
+    gh: https:/github.com/heptio/ark/tree/v0.10.0
 - scope:
```
skriss: I think we can drop all refs to the beta

PR #1062, opened 2018-11-16
Fix link to upgrade documentation in readme

danielcb:
The link to the upgrading to v0.10 docs was broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ You can run Ark in clusters on a cloud provider or on-premises. For detailed inf
 
 ## Breaking changes
 
-Ark version 0.10.0 introduces a number of breaking changes. Before you upgrade to version 0.10.0, make sure to read [the documentation on upgrading](upgrading-to-v0.10.md).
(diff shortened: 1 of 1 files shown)


nrb:
@skriss I think this is fine, since the README used by the site is in the `gh-pages` branch under the v0.10.0 directory.

skriss:
OK, looking at other links on this page, this looks fine, however @danielcb can you move the link to a numbered footnote like all the other links on this page (not sure why this one was different), and include a leading `/` like links `[30]` and `[99]`?  The number you use for this one doesn't really matter as long as it's unique.  Thanks!!

Bradamant3:
Yes, this change is fine. The doc-site-gen tooling we use strips the /docs directory prefix in the links, for the gh-pages site. Not sure why/how this one got stripped in the original, but many thanks to @danielcb for catching!

skriss:
Alright I'm gonna merge this as-is so we get the link fixed, we can come back and move to a numbered link later.

Issue #879, opened 2018-09-26
Error restoring pod with spec.priority set


cihangirbesiktas:
**What steps did you take and what happened:**
I created a Restic backup of a pod with an Azure persistent volume. Then, I tried to restore the backup on a different cluster and it failed.

Command outputs of the cluster that has been backed up.

```
$ kubectl get pod -n backup-example
NAME                                READY     STATUS    RESTARTS   AGE
nginx-deployment-6597d6d787-lthpx   1/1       Running   0          4h
$ kubectl get pod nginx-deployment-6597d6d787-lthpx -n backup-example -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    backup.ark.heptio.com/backup-volumes: nginx-logs
    cni.projectcalico.org/podIP: 100.96.0.88/32
  creationTimestamp: 2018-09-26T08:46:52Z
  generateName: nginx-deployment-6597d6d787-
  labels:
    app: nginx
    mybackup: ark
    pod-template-hash: "2153828343"
  name: nginx-deployment-6597d6d787-lthpx
  namespace: backup-example
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: ReplicaSet
    name: nginx-deployment-6597d6d787
    uid: b731a35e-c168-11e8-a1c8-0e68ddfac52d
  resourceVersion: "2828513"
  selfLink: /api/v1/namespaces/backup-example/pods/nginx-deployment-6597d6d787-lthpx
  uid: b73a3e40-c168-11e8-a1c8-0e68ddfac52d
spec:
  containers:
  - image: nginx:1.7.9
    imagePullPolicy: IfNotPresent
    name: nginx
    ports:
    - containerPort: 80
      protocol: TCP
    resources: {}
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /var/log/nginx
      name: nginx-logs
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: default-token-bxvgd
      readOnly: true
  dnsPolicy: ClusterFirst
  nodeName: shoot--msa-infra--pv-backup-cpu-worker-77dfd85c7d-6kw72
  priority: 0
  restartPolicy: Always
  schedulerName: default-scheduler
  securityContext: {}
  serviceAccount: default
  serviceAccountName: default
  terminationGracePeriodSeconds: 30
  tolerations:
[...]

PR #1074, opened 2018-11-21
Delete spec.priority in pod restore action

mwieczorek:
Fixes: #879 

Added deletion of field + test.
Also fixed another test (`nodeName (only) should be deleted from spec`) as it was 'green' even if I commented `//delete(spec, "nodeName")` in `pod_action.go`.

One remark:
I was testing it on cluster and without my fix I was able to add pod with spec.priority set if the priority on pod spec was equal to value defined in PriorityClass. I get an error only if I chenged value  in PriorityClass. 
But IMO it's better to delete this field.

### DIFF
--- a/pkg/restore/pod_action.go
+++ b/pkg/restore/pod_action.go
@@ -50,6 +50,9 @@ func (a *podAction) Execute(obj runtime.Unstructured, restore *api.Restore) (run
 	a.logger.Debug("deleting spec.NodeName")
 	delete(spec, "nodeName")
 
+	a.logger.Debug("deleting spec.priority")
+	delete(spec, "priority")
+
 	// if there are no volumes, then there can't be any volume mounts, so we're done.
 	if !collections.Exists(spec, "volumes") {
 		return obj, nil, nil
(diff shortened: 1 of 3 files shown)


ncdc:
@mwieczorek thanks for this! We could also do this to ensure we don't end up with a false positive test like the nodeName one that you identified:

```diff
diff --git a/pkg/restore/restore_test.go b/pkg/restore/restore_test.go
index 2e10fdc6..5d35dc2a 100644
--- a/pkg/restore/restore_test.go
+++ b/pkg/restore/restore_test.go
@@ -1605,6 +1605,9 @@ func (obj *testUnstructured) WithMetadata(fields ...string) *testUnstructured {
 }

 func (obj *testUnstructured) WithSpec(fields ...string) *testUnstructured {
+       if _, found := obj.Object["spec"]; found {
+               panic("spec already set - you probably didn't mean to do this twice!")
+       }
        return obj.withMap("spec", fields...)
 }
```

mwieczorek:
@ncdc thanks, I added your suggestion

ncdc:
lgtm too, thanks @mwieczorek

PR #1090, opened 2018-11-27
Added GOFLAGS support

### LINKED ISSUE #502: Make it possible to pass arguments to make test
It would nice to be able to do something like `GOFLAGS='-v -run TestFoo' make test`

dobozysaurus:
First time PR provider here :) Took a look at https://github.com/heptio/ark/issues/502. Not sure if you needed anything other than GOFLAGS, but if that's all that's needed, I think this PR ought to do it. Works for the -v flag, at least.

Fixes #502


ncdc:
Thanks @dobozysaurus! Nice to see you here! 😄 We'll test it out soon.

ncdc:
@dobozysaurus do you think you'll have time to pull in my `diff` and update the PR?

PR #1105, opened 2018-12-03
document Ark's behavior re: API versions

### LINKED ISSUE #592: Make sure we're backing up the "right" API versions to support upgrades
For example, if Kubernetes `vX` has API group/versions `abc/v1alpha1`, `abc/v1beta1`, and `abc/v1`, and the next Kubernetes version `vY` drops `abc/v1alpha1`, we shouldn’t be backing up that version.

We currently use the `ServerPreferredResources` from discovery, which returns the version of each API group preferred by the server.

skriss:

Fixes #592 

I'm assuming the location of this documentation may change going forward, but this seemed like a sensible place for it to land for now.

### DIFF
--- a/docs/about.md
+++ b/docs/about.md
@@ -49,6 +49,12 @@ By default, `ark backup create` makes disk snapshots of any persistent volumes. 
 
 ![19]
 
+## Backed-up API versions
+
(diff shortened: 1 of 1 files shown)


ncdc:
LGTM

PR #1108, opened 2018-12-03
Update current development changelog

wwitzel3:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 ## Development release:
 
-  * _Example: Add XYZ support (@user, #issue)_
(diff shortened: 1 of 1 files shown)


skriss:
idk if we need to capture a changelog for docs changes - i don't think we really have in the past - WDYT?

wwitzel3:
Yeah, I was kind of on auto pilot, just going through the merge commits since 0.10 release. I'll pull them out.

nrb:
I think I agree with leaving them out for the most part. A big docs change might be worth putting in, but fixing broken links doesn't seem like it quite meets the bar.

wwitzel3:
So with what @nrb said, should I leave in the update to our changelogs as an `Other Change`?

PR #1117, opened 2018-12-05
Add backup-version file in backup tarball

### LINKED ISSUE #134: Backup tarball should include backup version in a file
Let's put it in `backup-version`

wwitzel3:
Closes #134

### DIFF
--- a/pkg/backup/backup.go
+++ b/pkg/backup/backup.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"path/filepath"
 	"time"
 
 	"github.com/pkg/errors"
@@ -40,6 +41,9 @@ import (
 	"github.com/heptio/ark/pkg/util/collections"
 	kubeutil "github.com/heptio/ark/pkg/util/kube"
 )
+
+// BackupVersion is the current backup version for Ark.
+const BackupVersion = 1
 
 // Backupper performs backups.
 type Backupper interface {
@@ -221,6 +225,10 @@ func (kb *kubernetesBackupper) Backup(logger logrus.FieldLogger, backupRequest *
 	log := logger.WithField("backup", kubeutil.NamespaceAndName(backupRequest))
 	log.Info("Starting backup")
 
+	if err := kb.writeBackupVersion(tw); err != nil {
+		return errors.WithStack(err)
+	}
+
 	backupRequest.NamespaceIncludesExcludes = getNamespaceIncludesExcludes(backupRequest.Backup)
 	log.Infof("Including namespaces: %s", backupRequest.NamespaceIncludesExcludes.IncludesString())
 	log.Infof("Excluding namespaces: %s", backupRequest.NamespaceIncludesExcludes.ExcludesString())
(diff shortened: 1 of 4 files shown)


pkg/backup/backup.go
```diff
@@ -292,6 +297,23 @@ func (kb *kubernetesBackupper) Backup(logger logrus.FieldLogger, backupRequest *
 	return err
 }
 
+func (kb *kubernetesBackupper) writeBackupVersion(tw *tar.Writer) error {
+	hdr := &tar.Header{
+		Name:     "backup-version",
+		Size:     int64(len(buildinfo.Version)),
```
ncdc: This isn't the ark version. It needs to be a "backup version" that starts at something like 1 or 1.0.0 and then we'll manually increment as needed. And eventually have code that handles restoring different backup versions.

wwitzel3: Ahh, ok. So we'll just store that in a const then?

pkg/backup/backup.go
```diff
@@ -41,6 +41,9 @@ import (
 	kubeutil "github.com/heptio/ark/pkg/util/kube"
 )
 
+// BackupVersion is the current backup version for Ark.
+const BackupVersion = "1.0.0"
```
ncdc: Should probably be `1`

ncdc: https://github.com/heptio/ark/blob/c34e3b91276f7a2787264deb5f6be345990c7c8d/pkg/controller/backup_controller.go#L55

wwitzel3:
@ncdc PTAL

pkg/backup/backup.go
```diff
@@ -292,6 +300,25 @@ func (kb *kubernetesBackupper) Backup(logger logrus.FieldLogger, backupRequest *
...
+		Name:     "backup-version",
+		Size:     int64(len(versionString)),
+		Typeflag: tar.TypeReg,
+		Mode:     0644,
+		ModTime:  time.Now(),
+	}
+	if err := tw.WriteHeader(hdr); err != nil {
+		return err
+	}
+	if _, err := tw.Write([]byte(versionString)); err != nil {
```
ncdc: Do you want to include a newline? `fmt.Fprintln(tw, versionString)`, or even `fmt.Fprintf(tw, "%d\n", BackupVersion)`?

skriss:
Now that I'm thinking about this PR - we will now store some metadata as objects directly in object storage (e.g. `ark-backup.json`), and some (e.g. `backup-version`) as files within the tarball. Do we have a principled stance on what should go where? Would it make sense to duplicate any of it?

nrb:
@skriss That's a good question on the duplication. My current thought is that the backup version should be in the tarball, so that it could be processed if moved around. Information that's useful for periodic checks should be outside the tarball, though, since we don't want to download and extract them all on a regular basis.

wwitzel3:
I remember during the original discussion we wanted to intentionally keep `backup-version` in the tarball for the reason mentioned by @nrb

ncdc:
Nolan is correct - we want the `backup-version` somewhere in the tarball so that we can identify its format version even if we don't have the json to go with it. Maybe we should also include the json in it as well...

👍 to `metadata/backup-version` or even just `metadata/version`

skriss:
SGTM for now, though we probably need more thinking/work about how a tarball without an `ark-backup.json` file would be portable, if that's a goal.  Can address later.

ncdc:
I guess it probably can't be 100% portable, since we have other information in the bucket (snapshot details, logs)

skriss:
this lgtm pending moving the file into a `metadata/` dir

wwitzel3:
Ok, I think this is ready to go now. PTAL @skriss @ncdc

wwitzel3:
This is ready to merge if everyone is happy with it.

pkg/backup/backup.go
```diff
@@ -292,6 +300,26 @@ func (kb *kubernetesBackupper) Backup(logger logrus.FieldLogger, backupRequest *
...
+
+	hdr := &tar.Header{
+		Name:     versionFile,
+		Size:     int64(len(versionString)),
+		Typeflag: tar.TypeReg,
+		Mode:     0644,
+		ModTime:  time.Now(),
+	}
+	if err := tw.WriteHeader(hdr); err != nil {
+		return err
```
ncdc: I think we want `errors.Wrap` here

pkg/backup/backup.go
```diff
@@ -292,6 +300,26 @@ func (kb *kubernetesBackupper) Backup(logger logrus.FieldLogger, backupRequest *
...
+	}
+	if _, err := tw.Write([]byte(versionString)); err != nil {
+		return err
```
ncdc: errors.Wrap

skriss:
@ncdc I think we were just waiting for your approval here since you had a comment - merge if happy :)

ncdc:
Didn't realize y'all were waiting on me!

PR #1119, opened 2018-12-06
[WIP] plugin error handling

ncdc:


ncdc:
@carlisia here's a start so you can begin reviewing. There is some stuff that's commented out that I wasn't sure if I still needed it or not. I'll remove it as we finalize.

PR #1120, opened 2018-12-06
Fix output of Schedule Delete

jwhitcraft:
It should be `\n` and not `/n/` when doing the printf on a schedule delete.


ncdc:
@jwhitcraft thanks for being a first time contributor! Would you mind squashing this down to a single commit?

PR #1123, opened 2018-12-07
Add original item to restore plugin interface

### LINKED ISSUE #1092: Expand restore item action interfaces to include unmodified object from backup
Forked from https://github.com/heptio/ark/issues/965. 

> Is it possible to pass to plugin not only object with deleted metadata but also original object read from the backup?
If I get object with deleted metadata I have to rely on secret's name (which I think should be equal to SB name) but if I get object from the backup I can get old ownerReference.

mwieczorek:
Fixes #1092


pkg/plugin/proto/RestoreItemAction.proto
```diff
@@ -7,6 +7,7 @@ message RestoreExecuteRequest {
     string plugin = 1;
     bytes item = 2;
     bytes restore = 3;
+    bytes originalitem = 4;
```
ncdc: Please capitalize the `I` in `originalItem`

pkg/restore/restore.go
```diff
@@ -627,6 +627,7 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 	for _, file := range files {
 		fullPath := filepath.Join(resourcePath, file.Name())
 		obj, err := ctx.unmarshal(fullPath)
+		originalObj := obj.DeepCopy()
```
ncdc: Please move this down below the error check

pkg/plugin/proto/RestoreItemAction.proto
ncdc: @nrb @skriss in other bits of the restore code, we've called something like this `fromBackup` - how do you feel about that vs `originalItem`?

nrb: I think `fromBackup` is a little clearer here

mwieczorek: changed to `fromBackup`

pkg/restore/restore.go
mwieczorek: fixed

mwieczorek:
@ncdc Thanks for a quick response. 
I fixed the name of the parameter (`fromBackup`) and other small issues. Also, I added a file to the changelog folder.

BTW: I tested those changes with a simple plugin: https://github.com/mwieczorek/ark-plugins-test

changelogs/unreleased/1123-mwieczorek
```diff
@@ -0,0 +1 @@
+Add original resource (from backup) to restore plugin interface
```
ncdc:
```suggestion
Add original resource (from backup) to restore item action interface
```

pkg/restore/item_action.go
```diff
@@ -34,7 +34,7 @@ type ItemAction interface {
 	// should be returned, along with a warning (which will be logged but will not prevent
 	// the item from being restored) or error (which will be logged and will prevent the item
 	// from being restored) if applicable.
-	Execute(obj runtime.Unstructured, restore *api.Restore) (res runtime.Unstructured, warning error, err error)
+	Execute(obj runtime.Unstructured, fromBackup runtime.Unstructured, restore *api.Restore) (res runtime.Unstructured, warning error, err error)
```
ncdc: This has 3 inputs and 3 outputs. My PR #1125 adds a 4th output. Maybe we need to consider switching to a single input struct and an output struct + error. WDYT? cc @carlisia

pkg/restore/mocks/item_action.go
```diff
@@ -1,19 +1,5 @@
-/*
```
ncdc: Please don't remove the license header 😄

pkg/restore/restore.go
```diff
@@ -632,6 +632,8 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 			continue
 		}
 
+		fromBackup := obj.DeepCopy()
```
ncdc: Mind adding a comment above with a brief blurb about why we're doing the copy?

mwieczorek: comment added

pkg/restore/mocks/item_action.go
mwieczorek: sorry, it wasn't me - it was mockery ;)

changelogs/unreleased/1123-mwieczorek
mwieczorek: fixed

pkg/restore/item_action.go
ncdc: We should do this for backup item actions too, but we can do it in a separate PR.

The `warning` output variable makes things a bit messy here. We could do something like this:

```go
type RestoreItemActionExecuteOutput struct {
  UpdatedItem runtime.Unstructured
  Warning error
  Error error
}
```

but then the signature becomes `Execute(...) RestoreItemActionExecuteOutput`, with the `error` inside the struct instead of as the last output variable.

We could do this:

```go
type RestoreItemActionExecuteOutput struct {
  UpdatedItem runtime.Unstructured
  Warning error
}
```

along with `Execute(...) (RestoreItemActionExecuteOutput, error)`. Does that look weird or does it seem better?

mwieczorek: Indeed the `warning` makes it a little bit messy, but I can live with that. So I prefer option 2. (it will also remove warning about two errors in golint). Also, maybe others from the team would like to suggest something?
Let me know so I can make changes (but after holidays, if you prefer to do it in your PR - go ahead and I'll rebase on it)

mwieczorek: Added another commit with proposed changes. 
I had to tweak a little bit some tests (f.e. https://github.com/heptio/ark/pull/1123/files#diff-4cce89ebc50495fa33e9d2c373e87e6fR94) - and as I'm a golang newbie there may be a better way to do it - let me know

pkg/restore/item_action.go
```diff
@@ -34,7 +34,23 @@ type ItemAction interface {
 	// should be returned, along with a warning (which will be logged but will not prevent
 	// the item from being restored) or error (which will be logged and will prevent the item
 	// from being restored) if applicable.
-	Execute(obj runtime.Unstructured, restore *api.Restore) (res runtime.Unstructured, warning error, err error)
+	Execute(input *RestoreItemActionExecuteInput) (*RestoreItemActionExecuteOutput, error)
+}
+
+// RestoreItemActionExecuteInput is a struct passed to Execute function
```
ncdc: I would keep the type doc succinct and add per-field comments.
```go
// RestoreItemActionExecuteInput contains the input parameters for the ItemAction's Execute function.
```

pkg/restore/item_action.go
```diff
@@ -34,7 +34,23 @@ type ItemAction interface {
...
+// It contains item being processed by actions, unchanged item taken from backup
+// and restore spec
+type RestoreItemActionExecuteInput struct {
+	Item           runtime.Unstructured
```
ncdc:
```go
// Item is the item being restored. It is likely different from the pristine backed up version (metadata reset, changed by various restore item action plugins, etc.).
```

pkg/restore/item_action.go
```diff
@@ -34,7 +34,23 @@ type ItemAction interface {
...
+	ItemFromBackup runtime.Unstructured
+	Restore        *api.Restore
+}
+
+// RestoreItemActionExecuteOutput is a struct returned from Execute function
```
ncdc: `RestoreItemActionExecuteOutput contains the output variables for the ItemAction's Execution function.`

pkg/restore/item_action.go
```diff
@@ -34,7 +34,23 @@ type ItemAction interface {
...
-	Execute(obj runtime.Unstructured, restore *api.Restore) (res runtime.Unstructured, warning error, err error)
+	Execute(input *RestoreItemActionExecuteInput) (*RestoreItemActionExecuteOutput, error)
+}
+
+// RestoreItemActionExecuteInput is a struct passed to Execute function
+// It contains item being processed by actions, unchanged item taken from backup
+// and restore spec
+type RestoreItemActionExecuteInput struct {
+	Item           runtime.Unstructured
+	ItemFromBackup runtime.Unstructured
```
ncdc: Add field comment

pkg/restore/item_action.go
```diff
@@ -34,7 +34,23 @@ type ItemAction interface {
...
+	Restore        *api.Restore
```
ncdc: Add field comment

pkg/restore/item_action.go
```diff
@@ -34,7 +34,23 @@ type ItemAction interface {
...
+}
+
+// RestoreItemActionExecuteOutput is a struct returned from Execute function
+// It contains updated item and possibly a warning message
+type RestoreItemActionExecuteOutput struct {
+	UpdatedItem runtime.Unstructured
```
ncdc: Add field comment

pkg/restore/item_action.go
```diff
@@ -34,7 +34,23 @@ type ItemAction interface {
...
+	Warning     error
```
ncdc: Add field comment

pkg/plugin/proto/RestoreItemAction.proto
```diff
@@ -7,6 +7,7 @@ message RestoreExecuteRequest {
     string plugin = 1;
     bytes item = 2;
     bytes restore = 3;
+    bytes fromBackup = 4;
```
ncdc: I think `itemFromBackup` is probably a bit clearer - wdyt?

pkg/restore/job_action.go
```diff
@@ -38,21 +36,21 @@ func (a *jobAction) AppliesTo() (ResourceSelector, error) {
...
+		labels, err := collections.GetMap(input.Item.UnstructuredContent(), k)
 		if err != nil {
 			a.logger.WithError(err).Debugf("Unable to get %s", k)
 		} else {
 			delete(labels, v)
 		}
 	}
 
-	return obj, nil, nil
+	return &RestoreItemActionExecuteOutput{input.Item, nil}, nil
```
ncdc: WDYT about this (no warning)

```go
return NewRestoreItemActionExecuteOutput(input.Item), nil
```

and this (with warning)

```go
NewRestoreItemActionExecuteOutput(input.Item).WithWarning(someError)
```

pkg/restore/job_action_test.go
```diff
@@ -124,10 +124,10 @@ func TestJobActionExecute(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			action := NewJobAction(arktest.NewLogger())
 
-			res, _, err := action.Execute(test.obj, nil)
+			res, err := action.Execute(&RestoreItemActionExecuteInput{test.obj, test.obj, nil})
```
ncdc: Please use named fields

pkg/restore/pod_action_test.go
```diff
@@ -169,17 +169,18 @@ func TestPodActionExecute(t *testing.T) {
...
+			res, err := action.Execute(&RestoreItemActionExecuteInput{test.obj, test.obj, nil})
 
 			if test.expectedErr {
 				assert.NotNil(t, err, "expected an error")
 			} else {
 				assert.Nil(t, err, "expected no error, got %v", err)
 			}
 
-			assert.Equal(t, test.expectedRes, res)
+			if res != nil {
```
ncdc: I think it's stronger/safer to assert/require that res is non-nil when test.expectedErr is false?

pkg/restore/restore.go
```diff
@@ -632,6 +632,10 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 			continue
 		}
 
+		// make a copy of object retrieved from backup
+		// to make it available unchanged inside restore action
```
ncdc: action -> actions

mwieczorek:
@ncdc I think I addressed all above comments. PTAL

pkg/restore/item_action.go
carlisia: Sorry for the delay. Putting the multiple args in a struct is a great way to go. I too prefer the second option.

Unfortunately the code that this code calls has all been restructured (PR #1147), but we're going to have to go thru some merging pains, all good.

mwieczorek:
@ncdc can you take a look at my fixes? and let me know if I can do anything else. thanks

ncdc:
@mwieczorek sorry, it's been a crazy past couple of weeks. I'll get back to this now.

ncdc:
I just opened #1175 to deal with the year issue in generated code. Once it's merged, @mwieczorek would you mind rerunning `make update` and pushing the changes?

pkg/plugin/restartable_delegate_test.go
```diff
@@ -88,6 +88,14 @@ func runRestartableDelegateTests(
 						continue
 					}
 
+					// If function returns nil as struct return type, we cannot just
+					// compare the interface to nil as it's type will not be nil,
```
ncdc: it's -> its

pkg/restore/pod_action_test.go
```diff
@@ -169,17 +169,23 @@ func TestPodActionExecute(t *testing.T) {
...
 
 			if test.expectedErr {
 				assert.NotNil(t, err, "expected an error")
 			} else {
 				assert.Nil(t, err, "expected no error, got %v", err)
 			}
 
-			assert.Equal(t, test.expectedRes, res)
+			if !test.expectedErr {
+				assert.NotNil(t, res)
```
ncdc: I'd to `require` instead of `assert` here

pkg/restore/pod_action_test.go
```diff
@@ -169,17 +169,23 @@ func TestPodActionExecute(t *testing.T) {
...
-
-			assert.Nil(t, warning)
+			res, err := action.Execute(&RestoreItemActionExecuteInput{
+				Item:           test.obj,
+				ItemFromBackup: test.obj,
+				Restore:        nil,
+			})
 
 			if test.expectedErr {
 				assert.NotNil(t, err, "expected an error")
```
ncdc: We could add a `continue` here, and then we wouldn't need the `else` that follows (but we'd keep the assert-no-error check), and we also wouldn't need to check `if !test.expectedErr` again below.

mwieczorek: adding `continue` tells me `you're not in a loop` ;)
I added `return`

mwieczorek:
rebased and pushed, PTAL and let me know if I should squash the commits etc

ncdc:
Thanks for keeping on top of this. We'll working on the renaming effort, but we should have spare cycles again next week. Thanks for being patient!

mwieczorek:
@ncdc Hi, just checking out with this. 
Should I rebase and resolve all conflicts or there's still some work around plugins and renaming which I should wait for?

ncdc:
@carlisia @skriss would you mind working with @mwieczorek on this?

carlisia:
I'm taking a look.

carlisia:
@mwieczorek the `plugins` branch has been updated with the renaming work we did. You may squash and rebase things. Thank you.

mwieczorek:
@carlisia Squashed and rebased.

carlisia:
Let's wait for one more 👍

mwieczorek:
@skriss - could you take a look?

skriss:
Yes, I'll look at this tomorrow.

pkg/restore/pod_action_test.go
```diff
@@ -169,17 +170,23 @@ func TestPodActionExecute(t *testing.T) {
...
-			} else {
-				assert.Nil(t, err, "expected no error, got %v", err)
+				return
 			}
+			assert.Nil(t, err, "expected no error, got %v", err)
 
-			assert.Equal(t, test.expectedRes, res)
+			if !test.expectedErr {
```
skriss: this `if` can now be dropped since you're returning early if `test.expectedErr == true`

mwieczorek: `if` removed

mwieczorek:
> Also - it doesn't look like we're actually using the `Warning` field returned from a restore item action anywhere - are we? If not, I'd motion to remove it as part of this branch.

Do you want to remove it completely from the interface?

skriss:
@mwieczorek we can do that in a separate PR -- I meant the shared `plugins` branch, not your branch for this PR.

skriss:
I added #1230 to capture that. This PR LGTM1!

PR #1124, opened 2018-12-07
ark version changes

### LINKED ISSUE #1111: Simplify what's shown in `ark version`
`ark version` output is currently as follows:

For a release:
```
Version: v0.10.0
Git commit: v0.10.0
Git tree state: clean
```

For a development binary or container image binary built with `make build`, `make local`, etc:
```
Version: master
Git commit: v0.10.0-14-g4c5f352a
Git tree state: dirty
```

I'd like to make some changes to this:
- `Version` should remain as-is, which is either the git tag (if for a release built with goreleaser) or the value specified for `VERSION` (if built using `make build`)
- `Git commit` should *always* show the full git commit SHA that the build is for, rather than what it shows today
- `Git tree state`: I'm not sure on this one. Do folks think it's useful, or should we just drop it? Goreleaser doesn't actually let you build a release if the tree state isn't clean, so this is really only relevant for container image builds (either dev or release) and dev binary builds.  For release container image builds, since Jenkins is pulling the code directly from GH, logically it shouldn't be possible to have a non-clean status, unless something crazy is happening in Jenkins-land.  :shrug:

cc @ncdc @nrb @wwitzel3 @carlisia

skriss:
Fixes #1111 

- removes `Git tree state` line
- uses full git SHA in `Git commit` line, with a `-dirty` suffix if git tree state is dirty

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -39,7 +39,7 @@ builds:
       - goos: windows
         goarch: arm64
     ldflags:
-      - -X "github.com/heptio/ark/pkg/buildinfo.Version={{ .Tag }}" -X "github.com/heptio/ark/pkg/buildinfo.GitSHA={{ .Env.GIT_SHA }}" -X "github.com/heptio/ark/pkg/buildinfo.GitTreeState={{ .Env.GIT_TREE_STATE }}"
+      - -X "github.com/heptio/ark/pkg/buildinfo.Version={{ .Tag }}" -X "github.com/heptio/ark/pkg/buildinfo.GitSHA={{ .FullCommit }}" -X "github.com/heptio/ark/pkg/buildinfo.GitTreeState={{ .Env.GIT_TREE_STATE }}"
 archive:
   name_template: "{{ .ProjectName }}-{{ .Tag }}-{{ .Os }}-{{ .Arch }}"
   files:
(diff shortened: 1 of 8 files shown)


.goreleaser.yml
```diff
@@ -39,16 +39,14 @@ builds:
       - goos: windows
         goarch: arm64
     ldflags:
-      - -X "github.com/heptio/ark/pkg/buildinfo.Version={{ .Tag }}" -X "github.com/heptio/ark/pkg/buildinfo.GitSHA={{ .Env.GIT_SHA }}" -X "github.com/heptio/ark/pkg/buildinfo.GitTreeState={{ .Env.GIT_TREE_STATE }}"
+      - -X "github.com/heptio/ark/pkg/buildinfo.Version={{ .Tag }}" -X "github.com/heptio/ark/pkg/buildinfo.GitSHA={{ .FullCommit }}" -X "github.com/heptio/ark/pkg/buildinfo.GitTreeState={{ .Env.GIT_TREE_STATE }}"
```
ncdc: This line still has `GitTreeState`?

skriss: We're still showing it, just not as a separate line

ncdc: Nvm, I saw the removal of it below but didn't pay attention that it was from the `bug` command file.

ncdc:
LGTM

PR #1125, opened 2018-12-07
[WIP] Add additionalItems to restore item actions

### LINKED ISSUE #314: Restoration with label selector does not properly restore PVs
I recently took a full backup of our cluster `ark create backup complete`.  This backup includes several components and I wanted to test the restoration of just a subset of these initially.  

In an alternate cluster which was setup with to leverage the same S3 bucket for backup state I ran `ark create restore complete -l release=jenkins` as an attempt to only restore the relevant Jenkins resources.  

The restoration shows that it was successful and in the logs it does show that the PVC itself is being restored, however that PVC remains in a `Lost` state.  It seems that the logic to restore the underlying `PV` is filtering out the resource. 

NOTE: I also attempted to create a more granular backup, using the same selector `ark create backup jenkins -l release=jenkins`.  Restoring that application specific backup without an explicit label selector *does* work properly.  So it seems to be the combination of the `--selector` in the `restore` command that negatively impacts the resulting restore. 

Talking with Andy, he mentioned that the logic now does some additional `Pod -> PVC -> PV` logic that may need to be incorporated into the restoration process as well.  

Ark version: 
```
ark version
Version: v0.6.0
Git commit: v0.6.0
Git tree state: clean
```

K8 Version: 
```kubectl version
Client Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.1", GitCommit:"3a1c9449a956b6026f075fa3134ff92f7d55f812", GitTreeState:"clean", BuildDate:"2018-01-04T19:59:01Z", GoVersion:"go1.9.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.6", GitCommit:"6260bb08c46c31eea6cb538b34a9ceb3e406689c", GitTreeState:"clean", BuildDate:"2017-12-21T06:23:29Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}```

### PR DESCRIPTION
**Pull request by ncdc**
Add additionalItems to restore item actions, allowing use cases such as
pod -> pvc -> pv restores, even when using a label selector.

Note, this depends on #1123 because we need the PVC from the backup in addPVFromPVCAction.

Fixes #314

### CODE REVIEW

File: pkg/restore/add_pv_from_pvc_action.go [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+		return obj, nil, nil, nil
+	}
+
+	pv := ResourceIdentifier{
+		GroupResource: kuberesource.PersistentVolumes,
+		Name:          pvc.Spec.VolumeName,
+	}
+
+	a.logger.Infof("Adding PV %s as an additional item to restore", pvc.Spec.VolumeName)
+	return obj, []ResourceIdentifier{pv}, nil, nil
```
nrb: I know PVCs can use [selectors](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#selector); are they always a 1:1 mapping with PVs? Put another way, can 1 PVC select more than one PV, or is that an invalid state?

ncdc: Always 1:1

skriss:
Updated in #1304, closing.

PR #1131, opened 2018-12-18
Add note about kubeconfig in doc get-started.md

hex108:
Signed-off-by: Jun Gong <jungong@tencent.com>

I followed instructions in [get-started](https://heptio.github.io/ark/v0.10.0/get-started). When executing ark command, there is an error: `An error occurred: invalid configuration: no configuration has been provided`. In this PR, add a note for it.

### DIFF
--- a/changelogs/unreleased/1131-hex108
+++ b/changelogs/unreleased/1131-hex108
@@ -1,0 +1,1 @@
+Add debugging-install link in doc get-started.md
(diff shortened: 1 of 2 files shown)


ncdc:
@hex108 thanks for your contribution! We already have this information in `docs/debugging-install.md`. Let's work with @Bradamant3 to figure out the best place to capture this information.

Bradamant3:
@hex108 yes, this is a nice catch. But a more complete solution that also covers this case would be to add something like the following at l18 of `get-started.md`:
`If you encounter issues with installing or configuring, see [Debugging Installation Issues](debugging-install.md).`

hex108:
Thanks @ncdc @Bradamant3 ! Just updated it.

Bradamant3:
LGTM

PR #1147, opened 2019-01-08
Refactor plugin structure

### LINKED ISSUE #488: Reduce import surface of plugin interfaces
If you want to write a `backup.ItemAction` plugin, for example, you have to import all of `pkg/backup`, which is probably significantly more than necessary. Let's evaluate our plugin interfaces and make sure they're in 1 or more small packages that don't have large dependency trees.

carlisia:
This fixes #488 

*Update 1/10*

Done:
-  the plugin interface brakes because of moving the cloudprovider.ObjectStore and cloudprovider.BlockStore to their own object* and volume* packages
- moved ObjectStore, BlockStore and BackupStore interfaces each into a separate package to minimize the importable surface even further, for the cases where authors don't need to import all three.
-  moved code that was importing from `pkg/backup`, `pkg/restore` into an `action` package.
- created a package named `actioninterface` (`plugin/interface/actioninterface`) that holds the interfaces for backup and restore item actions. Now items corresponding to backup and restore are named `BackupItemAction` and `RestoreItemAction` throughout the code. This might be an improvement for clarity as a side effect. The biggest win was removing those interfaces (everything under `plugin/interface`) from the packages where they were being implemented. 


pkg/apis/ark/v1/zz_generated.deepcopy.go
```diff
@@ -1,7 +1,7 @@
 // +build !ignore_autogenerated
 
 /*
-Copyright 2018 the Heptio Ark contributors.
+Copyright 2019 the Heptio Ark contributors.
```
nrb: I think this file might be out of date - https://github.com/heptio/ark/commit/e38e2fc5a1e2dcd94ce135d925c5172b7b7b6b8b removed the year from the copyright lines.

nrb: Oh, nevermind - this is on the `plugins` branch, not `master`. My mistake.

pkg/plugin/action/backup/pod_action.go
```diff
@@ -22,32 +22,33 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 
-	"github.com/heptio/ark/pkg/apis/ark/v1"
+	v1 "github.com/heptio/ark/pkg/apis/ark/v1"
```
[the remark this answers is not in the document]
carlisia: Yes, I'm glad you noticed this, I agree with you. I looked thru the code and there's no reason for it to live under the plugin tree structure. 

I am moving it back to under `pkg/backup` but with one mod: I'm adding those files to a package named `backupaction`. It could alternatively be named `actionbackup`, or something else. Likewise for restore. I think separating those files keeps similar code together and could be beneficial when thinking about it. PTAL after I commit.

pkg/plugin/object_store.go
```diff
@@ -46,10 +28,6 @@ func NewObjectStorePlugin(options ...pluginOption) *ObjectStorePlugin {
 	}
 }
 
-//////////////////////////////////////////////////////////////////////////////
```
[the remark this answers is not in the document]
carlisia: Ah yes, I didn't tune into this but completely agree. Changing 👍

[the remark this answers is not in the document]
carlisia: I still have to do this.

pkg/backup/backupaction/rbac.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package backup
```
nrb: Should this file be moved to its own package? It doesn't seem like it fits in `backupaction` to me, even if some actions depend on it.

pkg/backup/mocks/item_action.go
```diff
@@ -16,25 +16,28 @@ limitations under the License.
 // Code generated by mockery v1.0.0. DO NOT EDIT.
 package mocks
 
-import backup "github.com/heptio/ark/pkg/backup"
```
nrb: Are these modifications by hand, or did mockery make them?

skriss:
@carlisia I'm not sure how much of a pain this is for you, but I think it'd be easier to work through this if we break it up into some smaller chunks, with separate PRs for each.  I'd propose starting with a PR that **just** extracts the 4 pluggable interfaces & related types out of their current packages and into independent packages.  Is that doable?

I'm also going back and forth between where it makes sense to move these interfaces + types *to*.  One option, similar to what you have now, is to move them all under `pkg/plugin`, either in a single subpackage, or in individual subpackages. What I don't like about this is that within the velero code base, other packages that require an `ObjectStore`, for example, now have to reference `pkg/plugin` or some subpackage of it, but really they don't need to know/care about the fact that this interface is pluggable.

An alternative would be to create a couple new packages, along the lines of:
```
pkg/storage
    object_store.go
pkg/volume (already exists)
    block_store.go
pkg/itemactions
    backup_item_action.go
    restore_item_action.go
    resource_identifier.go
    resource_selector.go
```

Another alternative would be a hybrid of the two, where we actually separate out the interfaces that *a plugin author implements* from the interfaces that *the velero code requires*.  The former would go into a subpackage of `pkg/plugin`, and the latter would go into some structure like the above.  The interface definitions would probably start as exact copies of each other, but could diverge over time.  The only hard requirements are that the GRPCServer implementations can translate gRPC requests into calls to the implemented interfaces, and the GRCPClient implementations can translate required interface calls to gRPC requests.

Let's discuss some more on this issue before proceeding further down a specific path.  @nrb please weigh in as well when you're back online!

carlisia:
I don't mind redoing this from scratch.

My preference is to go with the option to separate out the interfaces that _a plugin author implements_ from the interfaces that _the velero code requires_. This is the idea I had when I did the `plugin` directory vs the `pluginmanagement` directory. 

If we go this route, I think I would prefer the first PR to be either extract all of the non author code into a separate package, OR extract all of the author code into its own package. And then refine it from there. And hopefully that wouldn't cause a cyclical issue. But I could do it any other way.

There are many interfaces in these packages, what are you referring to as the "4 pluggable interfaces"?

skriss:
The 4 pluggable interfaces are:

- `pkg/backup.ItemAction`
- `pkg/restore.ItemAction`
- `pkg/cloudprovider.BlockStore`
- `pkg/cloudprovider.ObjectStore`

These are the ones that plugin authors can implement a plugin for.  The intent of the original issue, #488, was to not require a plugin author to import all of `pkg/backup` just to be able to implement a `pkg/backup.ItemAction` interface (for example).

Separately, there's **plugin framework** code that a plugin author needs to import in order to serve their plugin.  Today, that's all contained in `pkg/plugin`.  Specifically -- they need to use `pkg/plugin/server.go` and `pkg/plugin/logger.go` (e.g. https://github.com/heptio/velero-plugin-example/blob/master/velero-examples/main.go#L25).  As you point out, that package contains both framework code to be used by a plugin author, and internal code to be used only by velero, and ideally we can split that out as well.

I'd like to tackle these two issues in separate PRs, and I was proposing starting with the first since I think it's more straightforward, but if you want to start with the second, I'm open to it.

carlisia:
Ok I'm pretty sure we are saying the same thing for the most part.

skriss:
I agree - I just want to tackle them separately. Big PRs are too hard to transact on.

carlisia:
@skriss, @nrb how about this:

```
pkg/plugin/backup
    backup_item_action.go
pkg/plugin/restore
    restore_item_action.go
pkg/plugin/storage
    object_store.go
pkg/plugin/volume
    block_store.go
pkg/plugin/framework
    server.go
    logger.go

pkg/pluginmanagement/backup
    backup_item_action.go
pkg/pluginmanagement/restore
    restore_item_action.go
pkg/pluginmanagement/storage
    object_store.go
pkg/pluginmanagement/volume
    block_store.go
pkg/pluginmanagement/framework
    server.go
    logger.g
pkg/pluginmanagement
    TBD
```

skriss:
I like the direction. Couple thoughts:
- I'd be fine putting the 4 interfaces (`backup_item_action.go`, `restore_item_action.go`, `object_store.go`, `block_store.go`) in one package.  I think the import surface on the 4 combined interfaces is still quite small. But I'll defer to your judgment on this one.
- I like splitting out `pkg/plugin/framework`.  I'm guessing that when we start splitting it out, more items will have to go in there, but we can work thru that when you get to that PR.  I did take a look at this and I think I have a sense of where the cut line is.
- I think a flat `pkg/pluginmanagement` makes sense.  Some alternate name/nesting options -  `pkg/pluginmgmt`? `pkg/plugin/clientmgmt`? Food for thought.

As discussed, let's tackle moving the 4 interfaces into their own package(s) in one PR, and then splitting `pkg/plugin` out in a separate PR.  We can close this one and start fresh if you want.

skriss:
One other thing - maybe we should just do this development in `master` now that we're past v0.11.  Fewer merge conflicts down the road. Thoughts?

carlisia:
I like `pkg/plugin/clientmgmt`, let's go with that.

+1 for doing it against master. I'll open a PR to merge the commit that is ahead on the plugins branch.

carlisia:
To be sure, for the code to pass CI I will have to start moving the code into the `plugin/framework` package because of dependencies.

image

PR #1152, opened 2019-01-09
Restore Log Entry Clarification

### LINKED ISSUE #1142: Restore log needs more clarity when object already exists and isn't different
While debugging #1138, what I discovered is the following sequence of events:

1. Ark tries to restore a PV
1. It gets an already-exists error
1. There are no differences between `fromBackup` and `fromCluster`
1. Therefore it's a no-op

The only relevant/related log line for the above is

```
time="2019-01-03T14:36:29Z" level=info msg="Restoring PersistentVolume: pvc-18ceea3e-0a8d-11e9-aef1-02f9e6d3ca22" backup=pont logSource="pkg/restore/restore.go:796" restore=heptio-ark/pont-20190103153555
```

The restore has no warnings or errors.

I think we should consider changing the log line from `Restoring <type>: <name> ...` to `Attempting to restore <type>: ...`

I also think we should consider adding a log line to the effect `Did not restore <type>: <name> because it already exists and is identical to the backed up copy`

rdodev:
Fixes #1142 

Feel free to close if you want to wait for an outsider newbie who wants to take it.

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -793,7 +793,7 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 		// and which backup they came from
 		addRestoreLabels(obj, ctx.restore.Name, ctx.restore.Spec.BackupName)
 
-		ctx.log.Infof("Restoring %s: %v", obj.GroupVersionKind().Kind, name)
+		ctx.log.Infof("Attempting to restore %s: %v", obj.GroupVersionKind().Kind, name)
 		createdObj, restoreErr := resourceClient.Create(obj)
 		if apierrors.IsAlreadyExists(restoreErr) {
 			fromCluster, err := resourceClient.Get(name, metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


pkg/restore/restore.go
```diff
@@ -848,6 +848,7 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 					addToResult(&warnings, namespace, e)
 				}
 			}
+			ctx.log.Infof("Unable to restore %s: %v because it already exists and is identical to the backed up copy", obj.GroupVersionKind().Kind, name)
```
skriss: I think this needs to go into an `else` block, right? Otherwise, there are a couple of cases from the "not equal" if block above that will fall through and hit this, which will be incorrect.  If I'm reading it correctly the continue should remain outside the `else` since we want to continue for any case where we hit an AlreadyExists error.

rdodev: Ah, hmmm, good point.

rdodev: Which of 3 `if` statements would it be more pertinent to attach the `else` clause to? You think it might be worth refactoring by joining the 3 ifseses by `else if` and then finally `else`?

ncdc:
@rdodev since you're ok closing this, let's go with the new contributor PR #1153 😄 thanks!

PR #1153, opened 2019-01-09
Clarify conditional nature of restore in log

### LINKED ISSUE #1142: Restore log needs more clarity when object already exists and isn't different

daved:
Fixes #1142 

I took a shot at this, but am green enough to not have much surety about the change.

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -793,7 +793,7 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 		// and which backup they came from
 		addRestoreLabels(obj, ctx.restore.Name, ctx.restore.Spec.BackupName)
 
-		ctx.log.Infof("Restoring %s: %v", obj.GroupVersionKind().Kind, name)
+		ctx.log.Infof("Attempting to restore %s: %v", obj.GroupVersionKind().Kind, name)
 		createdObj, restoreErr := resourceClient.Create(obj)
 		if apierrors.IsAlreadyExists(restoreErr) {
 			fromCluster, err := resourceClient.Get(name, metav1.GetOptions{})
@@ -815,34 +815,36 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 			labels := obj.GetLabels()
 			addRestoreLabels(fromCluster, labels[api.RestoreNameLabel], labels[api.BackupNameLabel])
 
-			if !equality.Semantic.DeepEqual(fromCluster, obj) {
+			if equality.Semantic.DeepEqual(fromCluster, obj) {
+				ctx.log.Infof("Skipping restore of %s: %v because it already exists in the cluster and is unchanged from the backed up version", obj.GroupVersionKind().Kind, name)
+			} else {
 				switch groupResource {
 				case kuberesource.ServiceAccounts:
 					desired, err := mergeServiceAccounts(fromCluster, obj)
 					if err != nil {
 						ctx.log.Infof("error merging secrets for ServiceAccount %s: %v", kube.NamespaceAndName(obj), err)
 						addToResult(&warnings, namespace, err)
-						continue
+						break
 					}
 
 					patchBytes, err := generatePatch(fromCluster, desired)
 					if err != nil {
 						ctx.log.Infof("error generating patch for ServiceAccount %s: %v", kube.NamespaceAndName(obj), err)
 						addToResult(&warnings, namespace, err)
-						continue
+						break
 					}
 
 					if patchBytes == nil {
 						// In-cluster and desired state are the same, so move on to the next item
-						continue
+						break
 					}
 
 					_, err = resourceClient.Patch(name, patchBytes)
 					if err != nil {
 						addToResult(&warnings, namespace, err)
-					} else {
-						ctx.log.Infof("ServiceAccount %s successfully updated", kube.NamespaceAndName(obj))
+						break
 					}
+					ctx.log.Infof("ServiceAccount %s successfully updated", kube.NamespaceAndName(obj))
 				default:
 					e := errors.Errorf("not restored: %s and is different from backed up version.", restoreErr)
 					addToResult(&warnings, namespace, e)
--- a/changelogs/unreleased/1153-daved
+++ b/changelogs/unreleased/1153-daved
@@ -1,0 +1,1 @@
+Clarify restore log when object unchanged


daved:
I'm also not sure if the changelogs file is necessary for fixes like this one.

ncdc:
@daved thanks for the PR! Looks like @rdodev got in first by a couple of hours (#1152), although we haven't merged his yet, and he's fine deferring to you since you're a new community contributor. I'll leave it up to you - do you want your PR to go in, or should we proceed with @rdodev's?

daved:
Because the solutions take different approaches, I'm happy to defer to whichever approach is most appropriate. My instinct is that the verbiage and placement of the logging I chose fits a bit better, but, again, I'm not certain about it.

ncdc:
I think the verbiage that's in the other PR is better, but I'm happy to work with you here and proceed with this one. I'll add comments to the diffs.

pkg/restore/restore.go
```diff
@@ -793,7 +793,7 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 		// and which backup they came from
 		addRestoreLabels(obj, ctx.restore.Name, ctx.restore.Spec.BackupName)
 
-		ctx.log.Infof("Restoring %s: %v", obj.GroupVersionKind().Kind, name)
+		ctx.log.Infof("Verifying %s: %v", obj.GroupVersionKind().Kind, name)
```
ncdc: This isn't a verification. We need to tell the user were are going to attempt to restore the item. The call to `resourceClient.Create()` attempts to create the item in the Kubernetes apiserver. That is why I suggested changing this from "Restoring" to "Attempting to restore".

pkg/restore/restore.go
```diff
@@ -816,6 +816,7 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
 			addRestoreLabels(fromCluster, labels[api.RestoreNameLabel], labels[api.BackupNameLabel])
 
 			if !equality.Semantic.DeepEqual(fromCluster, obj) {
+				ctx.log.Infof("Restoring %s: %v", obj.GroupVersionKind().Kind, name)
```
ncdc: This isn't accurate. If we've made it here, it means that (1) we failed to restore the item because it already exists, and (2) that there is some difference between the backed-up copy of the item and the item that's currently in the cluster. But at this line, we don't yet know what's going to happen next. If the item is a ServiceAccount, then we'll attempt to patch it so we can merge some of the data from the backed-up copy with what's in the cluster. Otherwise, we're going to log that we couldn't restore it because it already exists and there are differences between the backed-up copy and the in-cluster version.

I think it might make more sense to do something like this:
```go
if equality.Semantic.DeepEqual(fromCluster, obj) {
  ctx.log.Infof("Skipping restore of xyz because it already exists in the cluster and is unchanged from the backed up version")
} else {
  // do the ServiceAccounts logic
}
continue
```

@skriss @rdodev wdyt about this reorg?

daved: Is this ok?

```go
if equality.Semantic.DeepEqual(fromCluster, obj) {
  ctx.log.Infof("Skipping restore of xyz because it already exists in the cluster and is unchanged from the backed up version")
  continue
}
switch groupResource {
case kuberesource.ServiceAccounts:
  // swap continues for breaks since a continue will occur at end of switch
default:
  // no change
}
continue
}
```

ncdc: Should be. I was trying to avoid multiple `continue` statements to try to keep it a bit less confusing. Also cc @carlisia for 👀

daved: A new commit is up, but I'm not marking as resolved to wait for the requested "eyes on" (or for someone else to mark).

ncdc:
I won't be able to finish reviewing this today/tomorrow, so I need to hand it off for now. @skriss can you take point?

pkg/restore/restore.go
```diff
@@ -815,38 +815,41 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
-			if !equality.Semantic.DeepEqual(fromCluster, obj) {
-				switch groupResource {
-				case kuberesource.ServiceAccounts:
-					desired, err := mergeServiceAccounts(fromCluster, obj)
-					if err != nil {
-						ctx.log.Infof("error merging secrets for ServiceAccount %s: %v", kube.NamespaceAndName(obj), err)
-						addToResult(&warnings, namespace, err)
-						continue
-					}
+			if equality.Semantic.DeepEqual(fromCluster, obj) {
```
carlisia: An `else` here should suffice.

pkg/restore/restore.go
```diff
@@ -815,38 +815,41 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+				ctx.log.Infof("Skipping restore of %v because it already exists in the cluster and is unchanged from the backed up version", name)
```
carlisia: Please add info for the type as well as the name. Ie: `obj.GroupVersionKind().Kind, name`

pkg/restore/restore.go
```diff
@@ -815,38 +815,41 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+				continue
```
carlisia: Instead of adding this `continue` statement, I suggest removing it and adding an `else` at the end of this statement. It's a small change that arguably makes it easier to read.

daved:
@carlisia 
I believe that this is ready to go now. Would you, please, have a look?

changelogs/unreleased/1142-daved
```diff
@@ -0,0 +1 @@
+Clarify restore log when object unchanged
```
skriss: can you rename this file to `1153-daved`? we're currently using the PR # rather than the issue #.

daved: No problem. Is it safe to assume that I should re-squash after?

skriss: yes please!

carlisia: Yes

skriss: @daved would you be able to get to this in the next day or two? thanks!

daved: @skriss Sure! Sorry for the delay, I had taken the weekend+Friday off from code. If I screwed anything up I'll be keeping an eye on this throughout the day.

skriss: awesome, thanks @daved!

PR #1166, opened 2019-01-23
velero server: log version and git SHA at startup

skriss:

We're currently only logging the Git SHA in the server log on startup. Log both the version and the Git SHA.

(Note: prior to #1124 this was actually displaying the version, because `FormattedGitSHA()` actually showed a version number if it was a tagged commit. A/o 1124 this changed to actually displaying the git SHA. In any case, I'd like to show both)


skriss:
@nrb this one could go to since it's trivial

PR #1167, opened 2019-01-24
Logging and misc cleanup

skriss:
Cleaned up a couple of logging issues that have come up over the last few days in Slack. Also noticed an unnecessary use of `meta.Accessor` while doing so, so removed it.


ncdc:
LGTM

nrb:
LGTM. Did you want to squash before merge, @skriss?

skriss:
nah

PR #1171, opened 2019-01-24
use restic stats instead of check to check repo existence

skriss:

The first time Velero is performing an operation against a restic repository, it first checks to see if the repo exists, and if not, it initializes it.  Today, the way it checks if the repo exists is by running the `restic check` command with the right flags and auth, and sees if that succeeds. 

Turns out that this command can be quite slow the first time that it's run on a host/container that doesn't already have a local restic cache set up (e.g. in Cluster B's Velero pod in a Cluster A -> Cluster B migration).  An alternate command, `restic stats`, is much faster and also achieves the desired outcome of ensuring the restic repo exists and can be authenticated to.

I've tested this manually and can replicate the slow behavior with the current code, and can confirm that the updated code is much faster.

### DIFF
--- a/pkg/controller/restic_repository_controller.go
+++ b/pkg/controller/restic_repository_controller.go
@@ -162,10 +162,10 @@ func (c *resticRepositoryController) initializeRepo(req *v1.ResticRepository, lo
 	})
 }
 
-// ensureRepo first checks the repo, and returns if check passes. If it fails,
-// attempts to init the repo, and returns the result.
+// ensureRepo first tries to connect to the repo, and returns if it succeeds. If it fails,
+// it attempts to init the repo, and returns the result.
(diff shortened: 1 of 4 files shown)


pkg/controller/restic_repository_controller.go
```diff
@@ -162,10 +162,10 @@ func (c *resticRepositoryController) initializeRepo(req *v1.ResticRepository, lo
 	})
 }
 
-// ensureRepo first checks the repo, and returns if check passes. If it fails,
+// ensureRepo first tries to connect to the repo, and returns if it succeeds. If it fails,
 // attempts to init the repo, and returns the result.
```
carlisia: Super nit pick: should be "it attempts".

skriss:
addressed.

PR #1186, opened 2019-02-05
Update templates to use current command

nrb:
Currently, the executable binary in the wild is `ark`, so leave it in
our github issue templates

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -27,7 +27,7 @@ about: Tell us about a problem you are experiencing
 
 **Environment:**
 
-- Velero version (use `velero version`): 
+- Velero version (use `ark version`): 
 - Kubernetes version (use `kubectl version`):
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -27,7 +27,7 @@ about: Tell us about a problem you are experiencing
 
 **Environment:**
 
-- Velero version (use `velero version`):
```
skriss: I think you switched the wrong one here (should be "Velero version (use `ark version`).:"

pkg/cmd/cli/bug/bug.go
```diff
@@ -71,7 +71,7 @@ about: Tell us about a problem you are experiencing
 
 **Environment:**
 
-- Velero version (use ` + "`velero version`" + `):{{.VeleroVersion}} {{.GitCommit}}
```
skriss: ditto

PR #1187, opened 2019-02-05
Use current info in bug template

nrb:
The binary, deployment, and namespace have yet to change for most users,
so make sure the bug template has the correct information when filing.


skriss:
I guess `bug.go` has to still refer to `ark` to pass `make verify`?

PR #1196, opened 2019-02-07
Document building from GitHub Release Source code archive

dvhart:
Update the documentation to include minimal instructions for building
the velero binary from the GitHub Release "Source code" archive.

Signed-off-by: Darren Hart <dvhart@vmware.com>

### DIFF
--- a/docs/build-from-scratch.md
+++ b/docs/build-from-scratch.md
@@ -1,7 +1,7 @@
 # Build from source
 
 * [Prerequisites][1]
-* [Download][2]
+* [Getting the source][2]
 * [Build][3]
 * [Test][12]
 * [Run][7]
@@ -16,6 +16,8 @@
 
 ## Getting the source
 
+### Option 1) Get latest (recommended)
+
 ```bash
 mkdir $HOME/go
 export GOPATH=$HOME/go
@@ -26,6 +28,8 @@ Where `go` is your [import path][4] for Go.
 
 For Go development, it is recommended to add the Go import path (`$HOME/go` in this example) to your path.
 
+### Option 2) Release archive
+Download the archive named `Source code` from the [release page][22] and extract it in your Go import path as `src/github.com/heptio/velero`.
 
 ## Build
 
@@ -40,6 +44,11 @@ make container
 ```
 
 To push your image to a registry, use `make push`.
+
+To build only the `velero` binary, run:
+```
+go build ./cmd/velero
+```
 
 ### Update generated files
 
@@ -204,7 +213,7 @@ If you need to add or update the vendored dependencies, see [Vendoring dependenc
 
 [0]: ../README.md
 [1]: #prerequisites
-[2]: #download
+[2]: #getting-the-source
 [3]: #build
 [4]: https://blog.golang.org/organizing-go-code
 [5]: https://golang.org/doc/install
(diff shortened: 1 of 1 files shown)


dvhart:
Dropped the changelog entry as I mispredicted the PR#, and it was deemed unnecessary via slack chat with Carlisia.

carlisia:
@dvhart while you're at it, mind doing a s/Download/Getting the source [here](https://github.com/heptio/velero/pull/1196/files#diff-dfb472ef2bde5d22f3cd94ed8845ad53R4)?

dvhart:
I noticed that broken link there as well - thought I should keep it separate. But sure, I'll add a commit to the PR with that.

docs/build-from-scratch.md
```diff
@@ -16,6 +16,8 @@
 
 ## Getting the source
 
+### Latest (Recommended)
```
carlisia: To make it super clear we are talking about two different options, I would make this line something like `Option 1) get latest (Recommended). Then on the next item `Option 2) Release Archive

docs/build-from-scratch.md
```diff
@@ -26,6 +28,9 @@ Where `go` is your [import path][4] for Go.
 
 For Go development, it is recommended to add the Go import path (`$HOME/go` in this example) to your path.
 
+### GitHub Release Source Code Archive
+Download the corresponding "Source code" archive from the release artifacts and
```
carlisia: No need to put source code in quotes. It'd be great to add a link to the release page under the word `release`.

docs/build-from-scratch.md
carlisia: Please note: I made only the first letter capitalized above.

dvhart: Can do.

docs/build-from-scratch.md
dvhart: I used quotes because "Source code" is the name github automatically gives the archive, and I was trying to call attention that specific term. Would you still prefer no quotes?

carlisia: Ah, gotcha. You may bold the words, **Source code**. Better yet: 

Download the archive named `Source code` from the release page.

dvhart: Done

dvhart:
Thanks for the review @carlisia , I believe the latest addresses all the feedback.

dvhart:
Squashed as requested.

docs/build-from-scratch.md
```diff
@@ -26,6 +28,8 @@ Where `go` is your [import path][4] for Go.
 
 For Go development, it is recommended to add the Go import path (`$HOME/go` in this example) to your path.
 
+### Option 2) Release archive
```
skriss: I think it's worth noting here that as we saw this afternoon, the standard `make` targets described in subsequent sections won't work properly if you use this option, but you can use `go build` to just build a binary.

dvhart: I should have added a bit more context in the PR description about how I ended up getting around to this implementation which wasn't quite what we discussed. Since ultimately I think we want the make targets to work as well, I wonder if rather than adding that caveat here, we shouldn't just update the Makefile to cope with building without the git repository. I'll have a look, but that will probably need some input from the team here regarding what to use in the absence of the git sha, etc. I'll have a closer look in the morning.

skriss: Got it - that works for me. We'll just want to have that makefile PR in before our next release.

dvhart: Upon closer inspection, it is not obvious how to best handle building from an archive - and there are too many questions regarding what happens if the buildinfo fields are empty or invalid. So I will update with a note about restricting archive builds to the go build commands and update the PR.

dvhart:
@skriss since this was merged already, I'll send a follow up PR re the go build caveat to source archives.

PR #1204, opened 2019-02-11
add deprecation notices to pkg/apis/ark/v1 types

skriss:

@nrb @carlisia I thought this was probably a good idea to add to all the types in `pkg/apis/ark/v1`.  Thoughts? Will add to the rest pending any feedback.


nrb:
I'm in favor of this, maybe adding a deadline version (v0.13? 1.0?) and an issue to actually delete the code during that cycle.

skriss:
Updated notice and added to all types.

skriss:
added #1208

@nrb please merge if you're happy with the update.

PR #1226, opened 2019-02-21
Add makefile to build site locally

carlisia:


Makefile
```diff
@@ -0,0 +1,22 @@
+PROJ_NAME={{.Name}}
+JEKYLL_VERSION=3.5
+IMAGE_NAME=doc-test
```
nrb: Is this image pushed to the repo? I'm not able to use `docker run gcr.io/heptio-images/doc-test:latest`

nrb:
Overall looks good, and super useful!

Do you intend to keep the `test` target, or can that be removed?

carlisia:
I should've added more context: I copied this makefile directly from the template here: https://github.com/heptio/doc-site-gen/blob/master/templates/Makefile.tmpl. I think we should keep it entirely as is. The `test` target is useful, see https://github.com/heptio/doc-site-gen#test-your-site.

nrb:
On my machine, it only appears to error.

```
% make test
docker run \
-v /home/nrb/go/src/github.com/heptio/velero:/project \
-e PROJ_NAME={{.Name}} \
gcr.io/heptio-images/doc-test:latest
Unable to find image 'gcr.io/heptio-images/doc-test:latest' locally
docker: Error response from daemon: manifest for gcr.io/heptio-images/doc-test:latest not found.
See 'docker run --help'.
make: *** [Makefile:8: test] Error 125
```

It also looks like `{{ .Name }}` needs to be replaced with `Velero`, since the source was a template file, meant to be parsed out.

Makefile
carlisia: I included the image making code.

carlisia:
Fixed all issues (I think). Note that running the test will generate errors, but fixing those errors should not be part of this PR.

Makefile
nrb: Awesome, thanks. I was able to get it to work, but had to check out the doc-site-gen repo in order to do so.

carlisia: Oh darn it. Good point.

nrb: fwiw, I think this approach is fine

README.md
```diff
@@ -4,9 +4,54 @@ redirect_to:
...
+  * If changes need to be made inside a subdirectory like `v0.3.0/` or `master/`, they should be made as PRs to that tagged branch.
+  * **Do not** make version-specific changes to the `gh-pages` branch, since they may be overridden by the automated logic of `doc-site-gen`.
+
+
+* **Project-specific changes**
+  * *Ex:* TOC changes
+  * These changes should be made as PRs to the project's `gh-pages`. branch, so that they apply across all versions.
+
+
+* **Global changes across *all* open source projects**
```
nrb: This section is for doc-site-gen itself. Style changes can probably go in all the repos at this point.

carlisia:
This ended up going in https://github.com/heptio/velero/pull/1228, leaving it there and closing this one.

PR #1231, opened 2019-02-22
update kubernetes and azure dependencies to 1.12

### LINKED ISSUE #1086: Upgrade Azure SDK to non-beta

skriss:
Fixes #1086 
 
This should not merge until *after* v0.11 is released.

Still needs to be manually tested on Azure.

Note: there is a breaking change in the 1.13 testing libraries that I've submitted an upstream patch for, but we'll probably want to skip over that version and wait for the 1.14 versions to come out before getting totally current.


skriss:
Did some testing on Azure and this seems to be working fine. I had to make one change to how we check for a 404 when deleting a snapshot.

carlisia:
How about these `ark` references, does it matter?: https://github.com/heptio/velero/pull/1231/files#diff-0a83572280909ad67a79837ac9eecfe4R22

skriss:
no, they're OK - we're keeping the old ark API group's generated code around until 1.0, so those can't be helped.

skriss:
alright, I tested this once more on Azure and everything looks good to me. let me know if you want me to squash anything or leave as-is (tried to do logical commits).

nrb:
High level pass on the commits looks good to me, probably doesn't need squashing.

carlisia:
Merging then...

PR #1232, opened 2019-02-22
add more documentation about using official releases

skriss:

On a fresh read through our docs I can see why there's confusion - I think these updates should help a lot. 🤞

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,19 @@ Velero consists of:
 
 You can run Velero in clusters on a cloud provider or on-premises. For detailed information, see [Compatible Storage Providers][99].
 
+## Installation
+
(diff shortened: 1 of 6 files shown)


ncdc:
Should we mention Ark/`ark` at all, since all the releases currently available are all `ark`?

skriss:
IDK - once we have 0.11 out (next week) I'd rather not, to keep things simpler for current versions.  I'm fine not merging this 'til ~Wednesday to avoid any additional confusion in the interim.

ncdc:
SGTM

skriss:
I'm dropped the word `velero` from just before "command-line client" on the README so it's applicable for both now and future. I think it's fine to merge now.

PR #1238, opened 2019-02-26
v0.10.2 cherrypicks

skriss:
@carlisia @nrb FYI this is what I'd like to release for v0.10.2 -- I think we can just do it at the same time as v0.11. I'll add release notes.


skriss:
updated with a proper changelog entry

skriss:
@nrb or @carlisia pls merge!

PR #1246, opened 2019-02-28
when restoring a PV, don't remove its spec.storageClassName

### LINKED ISSUE #1097: Document why Ark doesn't restore storageClassName on a PersistentVolume
We should document in both code comments and .md docs why Ark does not restore the storageClassName on a PersistentVolume. See 
https://github.com/heptio/ark/issues/791#issuecomment-416631566 for details.

skriss:

Fixes #1097 

I tested this on a couple public clouds and it worked fine.

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -1126,7 +1126,6 @@ func (r *pvRestorer) executePVAction(obj *unstructured.Unstructured) (*unstructu
 	}
 
 	delete(spec, "claimRef")
-	delete(spec, "storageClassName")
 
 	if boolptr.IsSetToFalse(r.snapshotVolumes) {
 		// The backup had snapshots disabled, so we can return early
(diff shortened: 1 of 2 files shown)


nrb:
Did you use this with a custom storage class across providers? For example, regional clusters on GCP require the user to create a custom storage class that represents the zones that the underlying volume is replicated against. This storage class isn't necessarily present in the destination cluster, so the PVs couldn't be scheduled.

skriss:
`StorageClasses` themselves are restorable, and are properly prioritized ahead of PVs, so I think in your scenario the user could always restore those StorageClasses first or along with the workload. If a user needs to change the storage class a PV is tagged as, or to remove it entirely, I'd think they would be expected to write a restore plugin to do that.

nrb:
@skriss Good points, thanks.

PR #1248, opened 2019-03-01
Improve `describe` output

### LINKED ISSUE #987: Improve `describe` output
I'd like to:
1) move the field `Validation errors` to right after the field phase (all operations).
2) make the value for the field `Validation errors` red for all values != `none`.

image

DheerajSShetty:
Improve `describe` output
   * Move Phase to right under Metadata(name/namespace/label/annotations)
   * Move Validation errors: section right after Phase: section and only
     show it if the item has a phase of FailedValidation
   * For restores move Warnings and Errors under Validation errors. Do not
      show Warnings or Errors if there are none.

Signed-off-by: DheerajSShetty <dheerajs@vmware.com>

Fixes #987

### DIFF
--- a/changelogs/unreleased/1248-DheerajSShetty
+++ b/changelogs/unreleased/1248-DheerajSShetty
@@ -1,0 +1,6 @@
+Improve `describe` output
+   * Move Phase to right under Metadata(name/namespace/label/annotations)
+   * Move Validation errors: section right after Phase: section and only
+     show it if the item has a phase of FailedValidation
+   * For restores move Warnings and Errors under Validation errors. Leave
+     their display as is.
--- a/pkg/cmd/util/output/backup_describer.go
+++ b/pkg/cmd/util/output/backup_describer.go
@@ -49,6 +49,15 @@ func DescribeBackup(
 		}
 		d.Printf("Phase:\t%s\n", phase)
 
+		status := backup.Status
+		if len(status.ValidationErrors) > 0 {
+			d.Println()
+			d.Printf("Validation errors:")
+			for _, ve := range status.ValidationErrors {
+				d.Printf("\t%s\n", ve)
+			}
+		}
+
 		d.Println()
 		DescribeBackupSpec(d, backup.Spec)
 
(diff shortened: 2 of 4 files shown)


DheerajSShetty:
Tested them by running the describe output command.
Before making the changes
before

After making the changes
after

DheerajSShetty:
These are the outputs before making changes and after the changes

Before
before

After
after

carlisia:
Love the output screenshots!

DheerajSShetty:
@skriss Can we merge this in?

skriss:
A couple comments:

- whatever changes we make for backups, we should make similar changes to the describers for other resources (restores, schedules).
- this change results in `status` information being split up (some's at the top, some's at the bottom). for example, with restores, validation errs would now be towards the top, but general warnings/errors are still towards the bottom. I think we might want to rearrange a few other things as well, to keep the output coherent.

Here's a refined proposal:
- for all the describers, move `Phase` to right under the top block of name/namespace/labels/annotations.
- for all the describers, move `Validation Errors` to right under Phase, **and only show it if the item has a phase of FailedValidation** (this is a change from current).
- for restores, move `Warnings` and `Errors` to right under `Validation Errors`. Leave their display as-is.

Let me know if you think this makes sense from a UX perspective. cc @carlisia @rosskukulinski @nrb

carlisia:
Yes, this flow makes total sense. I like it. 👍

DheerajSShetty:
restore
restore

schedule
schedule

backup
backup

carlisia:
image


I like the idea of not displaying the labels for errors if there isn't any. Should we not do the same for the restores?

DheerajSShetty:
@carlisia As per my discussion with @skriss, he asked me to keep Errors: and Warnings: the way it was. I am fine with anything.

carlisia:
@skriss keeping the errors for restores but not for the others make it appear inconsistent. PTAL.

DheerajSShetty:
@carlisia only `restore` has the section `Warnings:` and `Errors:` , `schedule` and `backup` does not have this section.

carlisia:
I understand. But to a user, an error is an error is an error. A validation error and an error will not be differentiated. I am suggesting we show it always or never in all cases.

skriss:
alright, let's hide a restore's warnings & errors if they're empty.

rosskukulinski:
@skriss All of this looks great to me.  My only comment might be to move the timestamp information higher up with the other "status" related things. +1 either way from me

DheerajSShetty:
Backup without Warning and Errors Section when there are none.
screen shot 2019-03-04 at 1 25 28 pm

skriss:
Looks good!

DheerajSShetty:
> Looks good!

Thank you!

PR #1250, opened 2019-03-01
move pvRestorer and tests to their own files

skriss:

I pulled the `pvRestorer` type and its associated tests out of `restore.go` and into their own files.  `restore.go` is a behemoth and I'm trying to do some cleanup to make it a little more manageable.


skriss:
argh, always `make update`!! updated :)

PR #1251, opened 2019-03-01
move backup extraction logic to its own type

skriss:

Another `restore.go` code move following #1250.


skriss:
@wwitzel3 when you're ready to look at some code

PR #1254, opened 2019-03-04
remove restore code that waits for a PV to become Available

skriss:

One more cleanup PR for `pkg/restore`.  This code ran during a restore, and waited (up to a minute) for all restored PVs to have a phase of `Available` before moving on to restoring other resources (e.g. PVCs).  The intent was to streamline creation of PVs, but we've looked at this previously and determined it's not necessary. Kubernetes will do the necessary retries, it immediately moves new PVs to `Available` phase if they're not already claimed, and PVCs will claim `Pending` PVs just fine anyway.

I did some testing on this, but if someone has time, would be great to get a second pair of eyes on it.

### DIFF
--- a/changelogs/unreleased/1254-skriss
+++ b/changelogs/unreleased/1254-skriss
@@ -1,0 +1,1 @@
+remove restore code that waits for a PV to become Available
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -25,7 +25,6 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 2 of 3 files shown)


nrb:
Code looks ok, though appears to have a merge conflict with previous PRs. I'd also like to try this against a cluster just to double check before approving/merging

skriss:
Ah yeah I thought that might happen, I'll fix the conflict soon. And yes another test would be great!

skriss:
fixed conflict & added changelog

PR #1256, opened 2019-03-06
Add original item to restore plugin interface

carlisia:
This had originally been merged into the `plugins` branch.

Signed-off-by: Michal Wieczorek <wieczorek-michal@wp.pl>


skriss:
@nrb PTAL (we're closing down the `plugins` branch and doing the work directly in `master` since we're past v0.11 now)

PR #1257, opened 2019-03-06
Refactor out plugin interfaces

carlisia:
This is a continuation of https://github.com/heptio/velero/pull/1147.

Highlights:
- This PR moves all the plugin interfaces and their dependencies into their own separate packages.
- All the code at this point has been duplicated; changing the code to consume these interfaces where they are now will be done in a separate PR or an amendment to this PR
- to avoid "stuttering", BackupItemAction and RestoreItemAction are not both named ItemAction; also, the code in those files are almost identical! I wonder if we could optimize this.
- the directory `pkg/plugin/framework` will have more files (like server.go and logger.go) after the code is changed to point to these new packages. I left anything that wasn't a dependency out for now.

### DIFF
--- a/pkg/plugin/backup/backup_item_action.go
+++ b/pkg/plugin/backup/backup_item_action.go
@@ -1,0 +1,235 @@
+/*
+Copyright 2017 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


skriss:
Is this PR 1 or 2 from https://github.com/heptio/velero/pull/1147#issuecomment-469909148?

skriss:
I think we're still missing each other on this - let's pair on it this afternoon.

PR #1262, opened 2019-03-06
AWS: Change format of volumeID to include availability zone

tsturzl:
This relates to the issue: https://github.com/heptio/velero/issues/1249

Highlights:
 * Changes to the SetVolumeID method for AWS block storage PVs
 * Format the volumeID to include the availability zone(eg. `aws://us-east-1a/volume-id`)


tsturzl:
Added the test case. Not exactly sure how that commit got into my working branch, but I went ahead and removed it.

skriss:
Thanks for the test case. Looks like there's still a stray commit in here that should be removed - `9458cdcd3f5ba533610913b326d8df934afe5f3e`.  Could you also please squash the remaining commits down to just 1?

@nrb / @carlisia PTAL.

changelogs/unreleased/1262-tsturzl
```diff
@@ -0,0 +1,3 @@
+Changes to AWS EBS
+  * modify the method SetVolumeID
+  * format PV volume IDs to variant which includes availability zone(eg. `aws://us-east-1a/volume-id`)
```
nrb: I think the changelog could just be this line.

PR #1270, opened 2019-03-11
add basic plugin panic handlers

### LINKED ISSUE #521: Add panic recovery handlers to all server-side plugin code
If a plugin encounters a panic during execution, right now we don't get any information about what caused the panic. Instead, all we see is something like this in the Ark server log:

```
ERRO[0004] backup failed                                 error="error executing custom action (groupResource=persistentvolumeclaims, namespace=brett, name=orientdb-storage-orientdb3.0-0): rpc error: code = Unavailable desc = transport is closing" key=heptio-ark/brett15 logSource="pkg/controller/backup_controller.go:273"
```

We should add panic handlers to all server-side plugin functions so we can capture the error (and maybe the full stack trace) and return gracefully instead of crashing the plugin process.

skriss:
Fixes #521

@carlisia @nrb this is a basic fix for #521.  It captures any info about the panic and returns it as a normal error.  Stack traces (per Andy's WIP #1119) to be added in subsequent PR(s).

If this looks reasonable for `block_store.go` then i'll add it to the other 3 plugins as well.

### DIFF
--- a/pkg/plugin/backup_item_action.go
+++ b/pkg/plugin/backup_item_action.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 the Heptio Ark contributors.
+Copyright 2017, 2019 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/plugin/backup_item_action.go
+++ b/pkg/plugin/backup_item_action.go
@@ -160,7 +160,13 @@ func (s *BackupItemActionGRPCServer) getImpl(name string) (velerobackup.ItemActi
 	return itemAction, nil
 }
 
-func (s *BackupItemActionGRPCServer) AppliesTo(ctx context.Context, req *proto.AppliesToRequest) (*proto.AppliesToResponse, error) {
+func (s *BackupItemActionGRPCServer) AppliesTo(ctx context.Context, req *proto.AppliesToRequest) (response *proto.AppliesToResponse, err error) {
+	defer func() {
+		if recoveredErr := handlePanic(recover()); recoveredErr != nil {
+			err = recoveredErr
+		}
+	}()
+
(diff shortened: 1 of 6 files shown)


nrb:
LGTM!

skriss:
👍 will proceed with the rest.

skriss:
Updated the rest. Will squash before merge, and would like to do a quick manual test.

skriss:
Squashed, test looks fine.

pkg/plugin/backup_item_action.go
```diff
@@ -160,7 +160,13 @@ func (s *BackupItemActionGRPCServer) getImpl(name string) (velerobackup.ItemActi
 	return itemAction, nil
```
nrb: Modify copyright header w/ year.

pkg/plugin/block_store.go
```diff
@@ -220,7 +220,13 @@ func (s *BlockStoreGRPCServer) getImpl(name string) (cloudprovider.BlockStore, e
 // Init prepares the BlockStore for usage using the provided map of
```
nrb: Modify copyright header w/ year.

pkg/plugin/object_store.go
```diff
@@ -216,7 +216,13 @@ func (s *ObjectStoreGRPCServer) getImpl(name string) (cloudprovider.ObjectStore,
 // Init prepares the ObjectStore for usage using the provided map of
```
nrb: Modify copyright header w/ year.

pkg/plugin/restore_item_action.go
```diff
@@ -159,7 +159,13 @@ func (s *RestoreItemActionGRPCServer) getImpl(name string) (restore.ItemAction,
 	return itemAction, nil
```
nrb: Modify copyright header w/ year.

skriss:
Updated all, try out squash & merge when ready

PR #1274, opened 2019-03-11
AWS: Inclue zone in volume ID

tsturzl:
Referencing this PR: https://github.com/heptio/velero/pull/1262

It became easier to just create a new branch rather than performing brain surgery(which I tired and resulted in the PR being closed). I just created a new branch off master and cherry picked my changes. This should include all previously requested changes.

Signed-off-by: Travis Sturzl<travis@metismachine.com>

### DIFF
--- a/pkg/cloudprovider/aws/block_store.go
+++ b/pkg/cloudprovider/aws/block_store.go
@@ -17,6 +17,7 @@ limitations under the License.
 package aws
 
 import (
+	"fmt"
 	"os"
 	"regexp"
 	"strings"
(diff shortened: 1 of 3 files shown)


tsturzl:
The details of the issue getting the previous branch to work, the commit got squashed which included removing a file from a commit that had been committed to that branch which shouldn't have been, then this commit that included said file was merged into master, and I had a commit later in the timeline(the squashed commit) which removes that file, which effectively breaks things. So rather than reworking the whole timeline I'm just opening a new PR with a new branch where I had just reapplied only the changes I had made. Not sure how these commits leaked in in the first place, possibly when I rebased to add a sign off on a commit. Sorry for any confusion.

skriss:
Looks good, just need to add back the changelog file :)

pkg/cloudprovider/aws/block_store.go
```diff
@@ -282,7 +283,13 @@ func (b *blockStore) SetVolumeID(unstructuredPV runtime.Unstructured, volumeID s
 		return nil, errors.New("spec.awsElasticBlockStore not found")
 	}
 
-	pv.Spec.AWSElasticBlockStore.VolumeID = volumeID
+	pvFailureDomainZone := pv.Labels["failure-domain.beta.kubernetes.io/zone"]
```
carlisia: I'm wondering if `failure-domain.beta.kubernetes.io/zone` should be a constant.

carlisia: @skriss I guess this is a question for you ^

tsturzl:
Change log as been added back.

skriss:
I think we're good to merge. @tsturzl - if possible, I'd prefer to squash this to a single commit.

tsturzl:
Done

pkg/cloudprovider/aws/block_store.go
tsturzl: I figured since it wasn't used more than once it wasn't buying much benefit having it as a separate variable. Finding it shouldn't too difficult, because changing it would likely mean you're directly concerned with this method. Though this string is used in other files, so maybe if this was a project or module wide it could be beneficial, but I didn't want to fit that all into a PR especially because I've not worked extensively with go.

PR #1277, opened 2019-03-12
Add location resources and tests

nrb:
Lays groundwork for #952

### DIFF
--- a/pkg/install/resources.go
+++ b/pkg/install/resources.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 the Heptio Ark contributors.
+Copyright 2018 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -20,6 +20,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1beta1 "k8s.io/api/rbac/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/heptio/velero/pkg/apis/velero/v1"
 )
 
 func labels() map[string]string {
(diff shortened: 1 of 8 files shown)


pkg/install/resources.go
```diff
@@ -56,13 +58,22 @@ func objectMeta(namespace, name string) metav1.ObjectMeta {
...
+			Kind:       "ServiceAccount",
+			APIVersion: corev1.SchemeGroupVersion.String(),
+		},
 	}
 }
 
 func ClusterRoleBinding(namespace string) *rbacv1beta1.ClusterRoleBinding {
-	return &rbacv1beta1.ClusterRoleBinding{
+	crb := &rbacv1beta1.ClusterRoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
```
skriss: you can use the existing `objectMeta` helper func here if you want

pkg/install/resources.go
```diff
@@ -77,12 +88,53 @@ func ClusterRoleBinding(namespace string) *rbacv1beta1.ClusterRoleBinding {
 			APIGroup: "rbac.authorization.k8s.io",
 		},
 	}
+
+	return crb
 }
 
 func Namespace(namespace string) *corev1.Namespace {
 	return &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
```
skriss: you can use the existing `objectMeta` helper func here if you want

pkg/install/resources.go
nrb: Does passing in `""` for a namespace make it cluster-scoped? I think that's why it had been left of previously.

skriss: yeah, it should

pkg/install/resources.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 the Heptio Ark contributors.
+Copyright 2018 the Velero contributors.
```
carlisia: I'm not on top of this discussion, should we be making this change now? If so, should the same change be made on the other files?

skriss: good q, we should try to be more on top of this. Any time we touch a file, we should update it, so this one should technically be:
```
Copyright 2018, 2019 the Velero contributors.
```

nrb: If it was simply `Heptio Ark -> Velero` I'd say we could do a single find/replace, but since we need to append rather than replace the year I'm not sure if it's worth modifying them in bulk. I'm guessing the regex wouldn't be too complicated, but also...regex.

carlisia: One pass: `s/Copyright 2018 the Heptio Ark/Copyright 2018, 2019 the Velero`, another pass: `s/Copyright 2017 the Heptio Ark/Copyright 2017, 2019 the Velero`. Should cover all?

PR #1278, opened 2019-03-12
pass --log-level to plugins

### LINKED ISSUE #1211: Plugin log levels not respected
Wrote a plugin and tried to use log levels other than Info.

When the ark `--log-level` option is set to `debug`, my plugin's `Debug/Debugf` logs should be printed.

N/A

Never seeing this line printed: https://github.com/heptio/velero/blob/1675943f44dfec16f369d631046e52f6f252283c/pkg/cmd/server/plugin/plugin.go#L41

which infers that: https://github.com/heptio/velero/blob/1675943f44dfec16f369d631046e52f6f252283c/pkg/cmd/server/plugin/plugin.go#L34 is not pulling in log levels from anywhere.

Inside: https://github.com/heptio/velero/blob/7e4fca428d73899cbbdb00da8d340d44984975d8/pkg/plugin/logger.go#L28, we can see that the log level isn't referenced so the default: `Info` is inferred.

**Environment:**

- Velero version (use `ark version`): v0.10.1
```
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.0", GitCommit:"ddf47ac13c1a9483ea035a79cd7c10005ff21a6d", GitTreeState:"clean", BuildDate:"2018-12-03T21:04:45Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.0", GitCommit:"ddf47ac13c1a9483ea035a79cd7c10005ff21a6d", GitTreeState:"clean", BuildDate:"2018-12-03T20:56:12Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}
```
- Kubernetes installer & version: Minikube v0.30.0
- Cloud provider or hardware configuration: vSphere VM
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

skriss:
Fixes #1211
Update of #1049

This PR passes the velero server's log level to all plugins (internal and external) via a `--log-level` flag. It adds helpers to `pkg/plugin.Server` so plugin authors can support this flag with a single line of code.

@nrb @carlisia I did this in a way that should work for plugins that use cobra (e.g. our internal `run-plugins` command) as well as those that don't (e.g. our plugin-example).  Please review this with an eye towards plugin developer experience.

Still doing some testing on this.

### DIFF
--- a/pkg/plugin/server.go
+++ b/pkg/plugin/server.go
@@ -17,10 +17,15 @@ limitations under the License.
 package plugin
 
 import (
+	"fmt"
 	"os"
+	"strings"
 
 	plugin "github.com/hashicorp/go-plugin"
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/pflag"
+
+	"github.com/heptio/velero/pkg/util/logging"
 )
 
 // Handshake is configuration information that allows go-plugin clients and servers to perform a handshake.
@@ -35,6 +40,14 @@ var Handshake = plugin.HandshakeConfig{
 
 // Server serves registered plugin implementations.
 type Server interface {
+	// BindFlags defines the plugin server's command-line flags
+	// on the provided FlagSet. If you're not sure what flag set
+	// to use, pflag.CommandLine is the default set of command-line
+	// flags.
+	//
+	// This method must be called prior to calling .Serve().
+	BindFlags(flags *pflag.FlagSet) Server
+
 	// RegisterBackupItemAction registers a backup item action.
 	RegisterBackupItemAction(name string, initializer HandlerInitializer) Server
 
@@ -65,6 +78,9 @@ type Server interface {
 
 // server implements Server.
 type server struct {
+	log               *logrus.Logger
+	logLevelFlag      *logging.LevelFlag
+	flagSet           *pflag.FlagSet
 	backupItemAction  *BackupItemActionPlugin
 	blockStore        *BlockStorePlugin
 	objectStore       *ObjectStorePlugin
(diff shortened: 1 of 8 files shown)


pkg/plugin/server.go
```diff
@@ -65,22 +78,36 @@ type Server interface {
...
 	restoreItemAction *RestoreItemActionPlugin
 }
 
 // NewServer returns a new Server
-func NewServer(log logrus.FieldLogger) Server {
+func NewServer() Server {
+	log := newLogger()
+
 	return &server{
+		log:               log,
```
carlisia: Could `newLogger()` be called on this line directly?

skriss: no, it needs to be declared ahead of time so it can be referenced in the subsequent lines of the struct def

skriss:
Fixed a possible NPE

skriss:
Addressed comments. We'll need to update the sample repo as well, but it'll be easier to do it once we get several of these plugin PRs merged.

skriss:
I've tested this with our internal plugins (using Cobra), and a toy external plugin both using Cobra and not.

nrb:
Agreed with holding example repo changes til most/all of the plugin changes land here.

nrb:
Oh - needs a changelog entry :)

skriss:
Doh. Added.

skriss:
@carlisia @nrb I think this is good to "Squash and merge" now if you're happy with it

Issue #1077, opened 2018-11-22
New gauge metrics for number of existing backups and restores


mwieczorek:
Ref: [thread on google groups](https://groups.google.com/forum/#!topic/heptio-ark/8mECN2CuQ14)

**Describe the solution you'd like**
I'd like to get new metrics from ark:
- [gauge] number of existing backups based on the number of CRDs (f.e. to track how many backups do we have and if they are deleted properly)
- [gauge] number of existing restores based on the number of CRDs (this one I don't have a use case, rather to make it consistent with backup metrics)

Another metric which can be valuable for me (but maybe questionable for you, I can move it to another issue for discussion or 'reject'):
- [gauge] size of all PV snapshots (labeled with a schedule, volumeSnapshotLocation) - to track how much space do I use.

skriss:
I split the snapshot size metric into its own issue, #1309, so let's scope this one just to the gauge metrics for total # of backups/restores that currently exist

PR #1280, opened 2019-03-13
add new counter metrics for backup deletion

### LINKED ISSUE #1059: New metrics for backup deletion attempt/failure/success
Ark provides metrics for f.e. failed backups and we can use it to display graphs in Grafana or make some alerts.
But in case of failed backup deletion, we don't have such metrics.

Also, Ark provides `ark_backup_tarball_size_bytes` but no corresponding metric for the PV snapshots size.

Would it be propose to add new metrics:

```
# TYPE ark_backup_delete_attempt_total counter
# TYPE ark_backup_delete_failure_total counter
# TYPE ark_backup_delete_success_total counter

# TYPE ark_backup_snapshots_size_bytes gauge
```

fabito:
Closes #1059 

This PR aims to solve https://github.com/heptio/velero/issues/1077
First I registered 2 new Gauges (number of backups and restores). I'm assuming we don't need GaugeVec here since we don't need grouping.

I could verify they were properly registered:

```bash
$ curl -s localhost:8585/metrics | grep -e "^velero_" 
velero_backup_attempt_total{schedule=""} 0
velero_backup_failure_total{schedule=""} 0
velero_backup_success_total{schedule=""} 0
velero_backup_total 1
velero_restore_attempt_total{schedule=""} 0
velero_restore_failed_total{schedule=""} 0
velero_restore_success_total{schedule=""} 0
velero_restore_total 25
velero_restore_validation_failed_total{schedule=""} 0
velero_volume_snapshot_attempt_total{schedule=""} 0
velero_volume_snapshot_failure_total{schedule=""} 0
velero_volume_snapshot_success_total{schedule=""} 0
```
Now I'm trying to figure out what's the best place to collect the actual metrics values. 
I'm thinking on collecting **number of backups** on `backup_controller` using something like:

```go
c.metrics.SetBackupCountGauge(len(backups));
```

Am I on the right direction ?

PS. Thanks for the patience, this is my first experience with Velero's codebase (and Go, to be honest)


fabito:
Sorry, I randomly picked a `Good first issue` from the issue list , I'm actually just trying to learn some Go.
[This is the thread](https://groups.google.com/d/msg/projectvelero/8mECN2CuQ14/j7F0tWJkCQAJ)  where the use case is described?

pkg/metrics/metrics.go
```diff
@@ -73,6 +75,13 @@ const (
 func NewServerMetrics() *ServerMetrics {
 	return &ServerMetrics{
 		metrics: map[string]prometheus.Collector{
+			backupCountGauge: prometheus.NewGauge(
```
nrb: I think this could be a `CounterVec` instead of a gauge. That way it can just be incremented with a call to `Inc()` inside the function.

pkg/metrics/metrics.go
```diff
@@ -372,6 +388,13 @@ func (m *ServerMetrics) InitSchedule(scheduleName string) {
 	// -------------------------------------------------------------------
 }
 
+// SetBackupCountGauge records the existing number of backups
+func (m *ServerMetrics) SetBackupCountGauge(size int) {
```
nrb: I think this function could just increment a counter, rather than take a size. That would be similar to `RegisterBackupAttempt`.

pkg/metrics/metrics.go
```diff
@@ -448,6 +471,13 @@ func toSeconds(d time.Duration) float64 {
 	return float64(d / time.Second)
 }
 
+// SetBackupCountGauge records the existing number of restores
+func (m *ServerMetrics) SetRestoreCountGauge(size int) {
```
nrb: I think this function could just increment a counter, rather than take a size. That would be similar to `RegisterRestoreAttempt`.

fabito:
@nrb , 

Regarding the  **number of backups deleted** metric, I think a good place to increment it is:
https://github.com/heptio/velero/blob/409116fce85d913490d1e1e02a73d375a04f9a9b/pkg/controller/backup_deletion_controller.go#L360
Agree ?

nrb:
Yes, that looks like the right place.

pkg/metrics/metrics.go
```diff
@@ -372,6 +408,13 @@ func (m *ServerMetrics) InitSchedule(scheduleName string) {
 	// -------------------------------------------------------------------
 }
 
+// RegisterBackup records the existing number of backups
+func (m *ServerMetrics) RegisterBackup(backupSchedule string) {
```
nrb: I'm wondering if we want these particular metric to be scoped to schedules. Since they're meant to represent all the backups (or restores) present, getting an accurate count would require subtracting the deletion metric from the number of backups per schedule, then summing all of those values.

This has the drawback of being more work on the scraping side, but the benefit of getting a fine-grained count of backups/restores per-schedule. What are your thoughts, @fabito?

pkg/metrics/metrics.go
```diff
@@ -124,6 +143,14 @@ func NewServerMetrics() *ServerMetrics {
...
 				[]string{scheduleLabel},
 			),
+			restoreTotal: prometheus.NewCounterVec(
+				prometheus.CounterOpts{
+					Namespace: metricNamespace,
+					Name:      restoreTotal,
+					Help:      "Number of existing restores",
+				},
+				[]string{scheduleLabel},
+			),
```
nrb: I think we need a corresponding `restoreDeletionTotal` with the appropriate register function, as well. Otherwise, the number of restores will always go up and we have no accounting for ones deleted.

There are 3 ways a restore can be deleted - via `velero restore delete`, via `velero backup delete` deleting all restores associated with the backup, and via garbage control, which also deletes all of a backup's existing restores.

L324 (on this branch) in backup_deletion_controller.go is where restores are deleted in the case of a backup being deleted, either explicitly or via garbage collection.

Deleting a restore via `velero restore delete` doesn't happen on a controller, though - it happens [here](https://github.com/heptio/velero/blob/master/pkg/cmd/cli/restore/delete.go#L108), on the client side. I'm honestly not sure how we communicate that to the velero server running in the cluster. We'll have to do some more research on this, I think.

carlisia:
The original issue was mainly requesting the count on backups, not restore. Must we add restore too for symmetry or can we skip it?

Things look fine so far pending the issues raised.

nrb:
> Must we add restore too for symmetry or can we skip it?

Worth considering leaving them out for now. My hunch is that someone will eventually be interested in that info for parity, but it also looks like properly counting restores is more complex, which I think justifies a separate PR.

@fabito does leaving the restore counts out of this PR entirely, then filing an issue for getting the restore parts added/fixed work for you?

fabito:
It makes sense @nrb, I'll remove `restore_total` from this PR and then we can open a separate issue to address restore metrics.

skriss:
Sorry I'm late to the party here, but finally caught up on this.  Here's what I'd propose doing:

- focus this PR solely on #1059, and specifically only the deletion attempt/failure/success counter metrics. It looks like there's a good start already for this.
- do a separate PR to address #1077.  I think for that one we want to add [gauge metrics](https://prometheus.io/docs/concepts/metric_types/#gauge) (which can go up or down - they can just be arbitrarily set to any value).  The values for these metrics would be based on the current count of backups and restores (either by counting custom resource instances, or by counting the contents of the object storage bucket - TBD).  We could set these metrics in one of a number of places:
    - in the backup sync controller (which periodically syncs backups in the obj storage bucket into the cluster as CRs)
    - in resync functions in the backup / restore controller
    - in a new metrics updater type, running in a goroutine and updating these things periodically

fabito:
@skriss ,

I'll create a separate PR for #1077.
Do the assumptions in the [initial comment](https://github.com/heptio/velero/pull/1280#issue-260881886) make sense ?

skriss:
@fabito let me think a little about where it makes sense to update the gauge metrics for #1077.  I may need a few days, super busy right now :/ I'll update further on issue #1077.

PR #1281, opened 2019-03-13
Handle redirects in Curl

jlcrow:


nrb:
Thanks for the PR! We require that all commits are signed off. You can add a sign-off with `git commit --amend --sign` and then `git push --force` to get the change into the PR.

skriss:
Thanks @jlcrow! Could you additionally make this change in the `master/migrating-to-velero.md` file in this same branch?

We should also update this file in the `master` branch (under `docs/`). That'll have to be a separate PR.  Yes, fixing docs bugs is not a great process right now.

skriss:
resolved in #1294, #1295, closing.

PR #1287, opened 2019-03-14
Add velero install command

### LINKED ISSUE #52: Add `velero install` command
It would be great to have E2E scripts for each cloud provider that go through a full install of Ark including pre-reqs and server installation. We can have the user fill in the necessary options at the beginning. The docs are great for understanding what's going on step-by-step but it takes a lot of copy-pasting to get a working install going.

### LINKED ISSUE #952: Update install docs to use `velero install` instead of in-tree YAML

nrb:
Fixes #52 
Will fix #952 

### DIFF
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -49,6 +49,7 @@ type InstallOptions struct {
 	DryRun               bool
 	BackupStorageConfig  flag.Map
 	VolumeSnapshotConfig flag.Map
+	UseRestic            bool
 }
 
 // BindFlags adds command line values to the options struct.
@@ -62,6 +63,7 @@ func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.Var(&o.VolumeSnapshotConfig, "snapshot-location-config", "configuration to use for the volume snapshot location. Format is key1=value1,key2=value2")
 	flags.BoolVar(&o.RestoreOnly, "restore-only", o.RestoreOnly, "run the server in restore-only mode. Optional.")
 	flags.BoolVar(&o.DryRun, "dry-run", o.DryRun, "generate resources, but don't send them to the cluster. Use with -o. Optional.")
+	flags.BoolVar(&o.UseRestic, "use-restic", o.UseRestic, "create restic deployment. Optional.")
 }
 
 // NewInstallOptions instantiates a new, default InstallOptions stuct.
@@ -92,6 +94,7 @@ func (o *InstallOptions) AsVeleroOptions() (*install.VeleroOptions, error) {
 		Prefix:       o.Prefix,
 		SecretData:   secretData,
 		RestoreOnly:  o.RestoreOnly,
+		UseRestic:    o.UseRestic,
 		BSLConfig:    o.BackupStorageConfig.Data(),
 		VSLConfig:    o.VolumeSnapshotConfig.Data(),
 	}, nil
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -57,9 +57,9 @@ func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Prefix, "prefix", o.Prefix, "prefix under which all Velero data should be stored within the bucket. Optional.")
 	flags.StringVar(&o.ProviderName, "provider", o.ProviderName, "provider name for backup and volume storage")
 	flags.StringVar(&o.Image, "image", o.Image, "image to use for the Velero and restic server deployment")
-	flags.StringVar(&o.Secret, "secret", o.Secret, "file containing credentials to backup and volume provider")
-	flags.Var(&o.BackupStorageConfig, "backup-location-config", "configuration to use for the backup storage location")
-	flags.Var(&o.VolumeSnapshotConfig, "snapshot-location-config", "configuration to use for the volume snapshot location")
+	flags.StringVar(&o.Secret, "secret", o.Secret, "file containing credentials for backup and volume provider")
+	flags.Var(&o.BackupStorageConfig, "backup-location-config", "configuration to use for the backup storage location. Format is key1=value1,key2=value2")
+	flags.Var(&o.VolumeSnapshotConfig, "snapshot-location-config", "configuration to use for the volume snapshot location. Format is key1=value1,key2=value2")
 	flags.BoolVar(&o.RestoreOnly, "restore-only", o.RestoreOnly, "run the server in restore-only mode")
 	flags.BoolVar(&o.DryRun, "dry-run", o.DryRun, "only print resources that would be installed, without sending them to the cluster")
 }
@@ -156,15 +156,15 @@ func (o *InstallOptions) Validate(c *cobra.Command, args []string, f client.Fact
 	}
 
 	if o.BucketName == "" {
-		return errors.Errorf("Bucket name must be provided")
+		return errors.New("--bucket is required")
 	}
 
 	if o.ProviderName == "" {
-		return errors.Errorf("Backup storage proivder name must be provided")
+		return errors.New("--provider is required")
 	}
 
 	if o.Secret == "" {
-		return errors.Errorf("No secret file provided")
+		return errors.New("--secret is required")
 	}
 
 	return nil
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -53,15 +53,15 @@ type InstallOptions struct {
 
 // BindFlags adds command line values to the options struct.
 func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
-	flags.StringVar(&o.BucketName, "bucket-name", o.BucketName, "name of the object storage bucket where backups should be stored")
-	flags.StringVar(&o.Prefix, "prefix", o.Prefix, "prefix under which all Velero data should be stored within the bucket. Optional.")
 	flags.StringVar(&o.ProviderName, "provider", o.ProviderName, "provider name for backup and volume storage")
-	flags.StringVar(&o.Image, "image", o.Image, "image to use for the Velero and restic server deployment")
+	flags.StringVar(&o.BucketName, "bucket", o.BucketName, "name of the object storage bucket where backups should be stored")
 	flags.StringVar(&o.Secret, "secret", o.Secret, "file containing credentials for backup and volume provider")
+	flags.StringVar(&o.Image, "image", o.Image, "image to use for the Velero and restic server pods. Optional.")
+	flags.StringVar(&o.Prefix, "prefix", o.Prefix, "prefix under which all Velero data should be stored within the bucket. Optional.")
 	flags.Var(&o.BackupStorageConfig, "backup-location-config", "configuration to use for the backup storage location. Format is key1=value1,key2=value2")
 	flags.Var(&o.VolumeSnapshotConfig, "snapshot-location-config", "configuration to use for the volume snapshot location. Format is key1=value1,key2=value2")
 	flags.BoolVar(&o.RestoreOnly, "restore-only", o.RestoreOnly, "run the server in restore-only mode")
-	flags.BoolVar(&o.DryRun, "dry-run", o.DryRun, "only print resources that would be installed, without sending them to the cluster")
+	flags.BoolVar(&o.DryRun, "dry-run", o.DryRun, "generate resources, but don't send them to the cluster. Use with -o")
 }
 
 // NewInstallOptions instantiates a new, default InstallOptions stuct.
@@ -80,7 +80,22 @@ func NewCommand(f client.Factory) *cobra.Command {
 	c := &cobra.Command{
 		Use:   "install",
 		Short: "Install Velero",
-		Long:  "Install Velero into the Kubernetes cluster using provided information",
+		Long: `
+Install Velero onto a Kubernetes cluster using the supplied provider information, such as
+the provider's name, a bucket name, and a file containing the credentials to access that bucket.
+A prefix within the bucket and configuration for the backup store location may also be supplied.
+Additionally, volume snapshot information for the same provider may be supplied.
+
+All required CustomResourceDefinitions will be installed to the server, as well as the
+Velero Deployment and associated Restic DaemonSet.
+
+The provided secret data will be created in a Secret named 'cloud-credentials'.
+
+All namespaced resources will be placed in the 'velero' namespace.
+
+Use '-o yaml' or '-o json'  with '--dry-run' to output all generated resources as text instead of sending the resources to the server.
+This is useful as a starting point for more customized installations.
+		`,
 		Run: func(c *cobra.Command, args []string) {
 			o.Namespace = c.Flag("namespace").Value.String()
 			cmd.CheckError(o.Validate(c, args, f))
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -169,6 +169,7 @@ func (o *InstallOptions) Run(c *cobra.Command) error {
 	if err != nil {
 		return err
 	}
+	fmt.Println("Velero is installed and ready! ⛵")
 	return nil
 }
 
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -74,6 +74,29 @@ func NewInstallOptions() *InstallOptions {
 	}
 }
 
+// AsVeleroOptions translates the values provided at the command line into values used to instantiate Kubernetes resources
+func (o *InstallOptions) AsVeleroOptions() (*install.VeleroOptions, error) {
+	realPath, err := filepath.Abs(o.Secret)
+	if err != nil {
+		return nil, err
+	}
+	secretData, err := ioutil.ReadFile(realPath)
+	if err != nil {
+		return nil, err
+	}
+	return &install.VeleroOptions{
+		Namespace:    o.Namespace,
+		Image:        o.Image,
+		ProviderName: o.ProviderName,
+		Bucket:       o.BucketName,
+		Prefix:       o.Prefix,
+		SecretData:   secretData,
+		RestoreOnly:  o.RestoreOnly,
+		BSLConfig:    o.BackupStorageConfig.Data(),
+		VSLConfig:    o.VolumeSnapshotConfig.Data(),
+	}, nil
+}
+
 // NewCommand creates a cobra command.
 func NewCommand(f client.Factory) *cobra.Command {
 	o := NewInstallOptions()
@@ -97,7 +120,6 @@ Use '-o yaml' or '-o json'  with '--dry-run' to output all generated resources a
 This is useful as a starting point for more customized installations.
 		`,
 		Run: func(c *cobra.Command, args []string) {
-			o.Namespace = c.Flag("namespace").Value.String()
 			cmd.CheckError(o.Validate(c, args, f))
 			cmd.CheckError(o.Complete(args, f))
 			cmd.CheckError(o.Run(c))
@@ -113,25 +135,12 @@ This is useful as a starting point for more customized installations.
 
 // Run executes a command in the context of the provided arguments.
 func (o *InstallOptions) Run(c *cobra.Command) error {
-	realPath, err := filepath.Abs(o.Secret)
+	vo, err := o.AsVeleroOptions()
 	if err != nil {
 		return err
 	}
-	secretData, err := ioutil.ReadFile(realPath)
-	if err != nil {
-		return err
-	}
-	resources, err := install.AllResources(
-		o.Namespace,
-		o.Image,
-		o.ProviderName,
-		o.BucketName,
-		o.Prefix,
-		o.BackupStorageConfig.Data(),
-		o.VolumeSnapshotConfig.Data(),
-		secretData,
-		o.RestoreOnly,
-	)
+
+	resources, err := install.AllResources(vo)
 	if err != nil {
 		return err
 	}
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -30,8 +30,6 @@ import (
 	"github.com/heptio/velero/pkg/install"
 )
 
-const defaultImage = "gcr.io/heptio/velero:latest"
-
 type InstallOptions struct {
 	Namespace                 string
 	DeploymentName            string
@@ -43,6 +41,7 @@ type InstallOptions struct {
 	LogLevel                  string
 	ResticTimeout             time.Duration
 	SecretName                string
+	DryRun                    bool
 }
 
 // Args:
@@ -54,16 +53,21 @@ type InstallOptions struct {
 
 func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
 	// TODO Send this string down into the deployment
-	flags.StringVar(&o.DeploymentName, "deploy-name", "", "name to apply to the Velero deployment")
-	flags.StringVar(&o.BucketName, "bucket-name", "", "bucket name in which to store backups")
-	flags.StringVar(&o.Prefix, "prefix", "", "prefix under the bucket in which to store backups")
-	flags.StringVar(&o.BackupStorageProviderName, "backup-provider", "", "provider name for backup storage")
-	flags.StringVar(&o.Image, "image", "", "image to use for the Velero server deployment")
-	flags.BoolVar(&o.RestoreOnly, "restore-only", false, "run the server in restore-only mode")
+	flags.StringVar(&o.DeploymentName, "deploy-name", o.DeploymentName, "name to apply to the Velero deployment")
+	flags.StringVar(&o.BucketName, "bucket-name", o.BucketName, "bucket name in which to store backups")
+	flags.StringVar(&o.Prefix, "prefix", o.Prefix, "prefix under the bucket in which to store backups")
+	flags.StringVar(&o.BackupStorageProviderName, "backup-provider", o.BackupStorageProviderName, "provider name for backup storage")
+	flags.StringVar(&o.Image, "image", o.Image, "image to use for the Velero server deployment")
+	flags.BoolVar(&o.RestoreOnly, "restore-only", o.RestoreOnly, "run the server in restore-only mode")
+	flags.BoolVar(&o.DryRun, "dry-run", o.DryRun, "don't create resources on the Kubernetes cluster")
 }
 
 func NewInstallOptions() *InstallOptions {
-	return &InstallOptions{}
+	return &InstallOptions{
+		Namespace:      api.DefaultNamespace,
+		DeploymentName: api.DefaultNamespace,
+		Image:          install.DefaultImage,
+	}
 }
 
 func NewCommand(f client.Factory) *cobra.Command {
@@ -71,7 +75,7 @@ func NewCommand(f client.Factory) *cobra.Command {
 	c := &cobra.Command{
 		Use:   "install",
 		Short: "Install Velero",
-		Long:  "Install Velero",
+		Long:  "Install Velero into the Kubernetes cluster using provided information",
 		Run: func(c *cobra.Command, args []string) {
 			o.Namespace = c.Flag("namespace").Value.String()
 			cmd.CheckError(o.Complete(args, f))
@@ -87,29 +91,23 @@ func NewCommand(f client.Factory) *cobra.Command {
 }
 
 func (o *InstallOptions) Run(c *cobra.Command) error {
-
+	//TODO pass backup and volume config down
 	resources, err := install.AllResources(o.Namespace, o.Image, o.BackupStorageProviderName, o.BucketName, o.Prefix)
 	if err != nil {
 		return err
 	}
 
-	// TODO: Wrap the resources in a list instead o just printing with separators
 	if _, err := output.PrintWithFormat(c, resources); err != nil {
 		return err
+	}
+
+	if o.DryRun {
+		return nil
 	}
 
 	return nil
 }
 
 func (o *InstallOptions) Complete(args []string, f client.Factory) error {
-	if o.Namespace == "" {
-		o.Namespace = api.DefaultNamespace
-	}
-	if o.DeploymentName == "" {
-		o.DeploymentName = api.DefaultNamespace
-	}
-	if o.Image == "" {
-		o.Image = defaultImage
-	}
 	return nil
 }
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -1,0 +1,142 @@
+/*
+Copyright 2019 the Heptio Ark contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package install
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/spf13/pflag"
+
+	"github.com/spf13/cobra"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
+	api "github.com/heptio/velero/pkg/apis/velero/v1"
+	"github.com/heptio/velero/pkg/client"
+	"github.com/heptio/velero/pkg/cmd"
+	"github.com/heptio/velero/pkg/cmd/util/output"
+	"github.com/heptio/velero/pkg/install"
+)
+
+const defaultImage = "gcr.io/heptio/velero:latest"
+
+type InstallOptions struct {
+	Namespace                 string
+	DeploymentName            string
+	Image                     string
+	BucketName                string
+	Prefix                    string
+	BackupStorageProviderName string
+	RestoreOnly               bool
+	LogLevel                  string
+	ResticTimeout             time.Duration
+	SecretName                string
+}
+
+// Args:
+//   secret
+//   plugin-dir
+//   basically all the server args?
+// Flags
+//   dry-run (yaml dump)
+
+func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
+	// TODO Send this string down into the deployment
+	flags.StringVar(&o.DeploymentName, "deploy-name", "", "name to apply to the Velero deployment")
+	flags.StringVar(&o.BucketName, "bucket-name", "", "bucket name in which to store backups")
+	flags.StringVar(&o.Prefix, "prefix", "", "prefix under the bucket in which to store backups")
+	flags.StringVar(&o.BackupStorageProviderName, "backup-provider", "", "provider name for backup storage")
+	flags.StringVar(&o.Image, "image", "", "image to use for the Velero server deployment")
+	flags.BoolVar(&o.RestoreOnly, "restore-only", false, "run the server in restore-only mode")
+}
+
+func NewInstallOptions() *InstallOptions {
+	return &InstallOptions{}
+}
+
+func NewCommand(f client.Factory) *cobra.Command {
+	o := NewInstallOptions()
+	c := &cobra.Command{
+		Use:   "install",
+		Short: "Install Velero",
+		Long:  "Install Velero",
+		Run: func(c *cobra.Command, args []string) {
+			o.Namespace = c.Flag("namespace").Value.String()
+			cmd.CheckError(o.Complete(args, f))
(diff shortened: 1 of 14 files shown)


pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return c
+}
+
+func (o *InstallOptions) Run(c *cobra.Command) error {
+
+	//TODO: Extract this logic into a function inside of the install package that anyone could use.
+	var resources []runtime.Object
+
+	crds := install.CRDs()
```
carlisia: Not suer I'm reading this right: `CRDs` is in this same package right, do we need the `install.`?

carlisia: Also, could we just do `range CRDs`?

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,142 @@
+/*
+Copyright 2019 the Heptio Ark contributors.
```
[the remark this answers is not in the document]
nrb: Yes, I need to update these.

pkg/cmd/cli/install/install.go
nrb: Probably could range.

`install` is `"github.com/heptio/velero/pkg/install"`, not this package. I'm not wild about that naming, but this package is following the same conventions as the rest of the client commands (package named after the command)

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,142 @@
...
+		fmt.Println("---")
+		if _, err := output.PrintWithFormat(c, r); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (o *InstallOptions) Complete(args []string, f client.Factory) error {
```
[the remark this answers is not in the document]
nrb: Probably doesn't have to yet, though the rest of the `Complete` funcs for CLI commands are...mostly to facilitate using both `<resource> <verb>` and `<verb> <resource>` ordering.

nrb:
This is a draft PR, so it doesn't actually install things yet. All it will do is print things. Use `-o yaml` or `-o json` to see the objects.

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,142 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+
+	api "github.com/heptio/velero/pkg/apis/velero/v1"
+	"github.com/heptio/velero/pkg/client"
+	"github.com/heptio/velero/pkg/cmd"
+	"github.com/heptio/velero/pkg/cmd/util/output"
+	"github.com/heptio/velero/pkg/install"
+)
+
+const defaultImage = "gcr.io/heptio/velero:latest"
```
skriss: s/heptio/heptio-images

skriss: rather than hardcoding `latest` it would be nice to get the version tag from `pkg/buildinfo`.

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,142 @@
...
+		if _, err := output.PrintWithFormat(c, r); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (o *InstallOptions) Complete(args []string, f client.Factory) error {
+	if o.Namespace == "" {
```
skriss: rather than defaulting here, I'd instantiate the options struct with default values for each relevant field. Then, when the flags are bound, you can use the field itself as the default value for the flag, and it'll be obvious to the user what the default values are.

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,142 @@
...
+	o.BindFlags(c.Flags())
+	output.BindFlags(c.Flags())
+	output.ClearOutputFlagDefault(c)
+
+	return c
+}
+
+func (o *InstallOptions) Run(c *cobra.Command) error {
+
+	//TODO: Extract this logic into a function inside of the install package that anyone could use.
```
skriss: 👍

nrb:
I wrapped all the resources into an UnstructuredList, still need to address the review comments though.

As it stands, I can do `velero install -o json | kubectl apply -f -` and it will create all the core Kubernetes objects, but fails to create the BSL instance. Running it again succeeds. So, I think having everything in 1 document like that will result in a race condition where the CRD's Kind isn't fully defined by the time we create a single instance of it.

nrb:
Current status:

* Creates all resources except secret on the cluster

TODO:
* Decide how to report progress (logs vs plain print statements)
* Create the secret
* Pass down configurations
* Write docs

skriss:
Sweet. Mind rebasing? Does it make sense for me to do a review of the current state?

nrb:
Yeah, a pass now wouldn't hurt. I'm especially curious about using a logrus logger to provide output, or just using fmt.Printf statements. I think the majority of output is there now.

nrb:
Oh, `-o table` fails currently, as well. I'll have to review the printing code to see if I can find a way to make that an invalid option for the install command, since the heterogeneous collection won't have common fields.

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/heptio/velero/pkg/cmd"
+	"github.com/heptio/velero/pkg/cmd/util/output"
+	"github.com/heptio/velero/pkg/install"
+	"github.com/heptio/velero/pkg/util/logging"
+)
+
+// InstallOptions collects all the options for installing Velero into a Kubernetes cluster.
+type InstallOptions struct {
+	Namespace                 string
+	DeploymentName            string
```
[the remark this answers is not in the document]
nrb: That's fine with me. I'll take another pass through the args and see if anything makes sense to eliminate, as I haven't really adjusted them since scoping the issue down.

nrb:
I was able to use this code to create everything on a KinD cluster and back up the default namespace to the BSL.

Need to write docs and test with each cloud provider now, but I think the bulk of the work is ready for review now.

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,187 @@
...
+
+The provided secret data will be created in a Secret named 'cloud-credentials'.
+
+All namespaced resources will be placed in the 'velero' namespace.
+
+Use '-o yaml' or '-o json'  with '--dry-run' to output all generated resources as text instead of sending the resources to the server.
+This is useful as a starting point for more customized installations.
+		`,
+		Run: func(c *cobra.Command, args []string) {
+			o.Namespace = c.Flag("namespace").Value.String()
```
skriss: Any reason why this isn't bound like all the other flags?

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	if o.ProviderName == "" {
+		return errors.New("--provider is required")
+	}
+
+	if o.Secret == "" {
+		return errors.New("--secret is required")
+	}
+
```
skriss: probably makes sense to add a check here that the value of the `-o` flag is not `table`.

pkg/install/install.go
```diff
@@ -0,0 +1,87 @@
...
+	"DaemonSet":                "daemonsets",
+	"Secret":                   "secrets",
+	"BackupStorageLocation":    "backupstoragelocations",
+	"VolumeSnapshotLocation":   "volumesnapshotlocations",
+}
+
+// Install creates resources on the Kubernetes cluster.
+// An unstructured list of resources is sent, one at a time, to the server. These are assumed to be in the preferred order already.
+// An io.Writer can be used to output to a log or the console.
+func Install(factory client.DynamicFactory, resources *unstructured.UnstructuredList, w io.Writer) error {
```
skriss: I'm thinking that when you add waiting for CRDs to be ready, you'll want this function to accept a slice of structs like:

```go
type ResourceGroup struct {
    Resources []*unstructured.Unstructured
    Ready func(dynamicFactory client.DynamicFactory) (bool, error)
}
```

So you can have one `ResourceGroup` for all the CRDs, and the Ready func uses the dynamic factory to check them for readiness. The `Install` func will loop through each `ResourceGroup`, and for each, create all of the `Resources`, then call the `Ready` func in a polling loop, up to a certain period. Not dissimilar from the "wait for terminated" stuff in restore.go.  This is my initial thought off the top of my head but obv can be refined/changed.

pkg/install/install.go
```diff
@@ -0,0 +1,87 @@
...
+			Name:       kindToResource[r.GetKind()],
+			Namespaced: (r.GetNamespace() != ""),
+		}
+
+		c, err := factory.ClientForGroupVersionResource(gvk.GroupVersion(), apiResource, r.GetNamespace())
+		if err != nil {
+			return errors.Wrapf(err, "Error creating client for resource %s", id)
+		}
+
+		_, err = c.Create(&r)
```
skriss: you can collapse this line into the next line (`if _, err = c.Create(&r); apierrors.Is...`)

pkg/install/install.go
```diff
@@ -0,0 +1,87 @@
...
+		if apierrors.IsAlreadyExists(err) {
```
skriss: Probably not needed for MVP, but it would be really nice if the `install` command could be more like `kubectl apply`, where any changes to the resources are applied as patches. That might solve the bulk of the `upgrade` case. And it would also make it really easy for a user to realize they mis-configured something, and just re-run `velero install` and have the change be applied.

pkg/install/install.go
```diff
@@ -0,0 +1,87 @@
...
+			log("already exists, proceeding")
+			continue
+		} else if err != nil {
+			return errors.Wrapf(err, "Error creating resource %s", id)
```
skriss: Do we want to create as much as we can, or immediately terminate on the first failure? I'm not sure, but consider it.

pkg/install/install.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		log("created")
+		time.Sleep(1 * time.Second)
```
skriss: Let's do proper CRD waiting instead of this.

pkg/install/resources.go
```diff
@@ -132,3 +139,81 @@ func VolumeSnapshotLocation(namespace, provider string, config map[string]string
...
+// Items are unstructured, since there are different data types returned.
+func AllResources(namespace, image, backupStorageProviderName, bucketName, prefix string, bslConfig, vslConfig map[string]string, secretData []byte, restoreOnly bool) (*unstructured.UnstructuredList, error) {
+	resources := new(unstructured.UnstructuredList)
+	// Set the GVK so that the serialization framework outputs the list properly
+	resources.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "List"})
+
+	for _, crd := range CRDs() {
+		unstructuredCRD, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&crd)
+		// Remove the status field, as it's added with null values by the code, but is rejected by server-side validation
+		delete(unstructuredCRD, "status")
```
skriss: we should probably do this for all of the resources -- I noticed that if you do `-o yaml`, then you get empty statuses for some of the other types. Would be best to not have that.

pkg/install/resources.go
```diff
@@ -132,3 +139,81 @@ func VolumeSnapshotLocation(namespace, provider string, config map[string]string
...
+	if err != nil {
+		return err
+	}
+	list.Items = append(list.Items, unstructured.Unstructured{Object: u})
+	return nil
+}
+
+// AllResources returns a list of all resources necessary to install Velero, in the appropriate order, into a Kubernetes cluster.
+// Items are unstructured, since there are different data types returned.
+func AllResources(namespace, image, backupStorageProviderName, bucketName, prefix string, bslConfig, vslConfig map[string]string, secretData []byte, restoreOnly bool) (*unstructured.UnstructuredList, error) {
```
skriss: I would probably create an `Options` struct with the params, since the list is pretty long.

pkg/install/resources.go
```diff
@@ -132,3 +139,81 @@ func VolumeSnapshotLocation(namespace, provider string, config map[string]string
...
+	)
+	if restoreOnly {
+		deploy = Deployment(namespace,
+			WithImage(image),
+			WithRestoreOnly(),
+		)
+	}
+	appendUnstructured(resources, deploy)
+
+	ds := DaemonSet(namespace,
```
skriss: I think we probably want to add a flag to optionally deploy the restic daemonset, since for folks not using restic it's completely unnecessary and a waste of resources.

pkg/util/encode/encode.go
```diff
@@ -42,17 +46,19 @@ func Encode(obj runtime.Object, format string) ([]byte, error) {
...
 	if err != nil {
 		return err
 	}
-
 	return errors.WithStack(encoder.Encode(obj, w))
 }
 
 // EncoderFor gets the appropriate encoder for the specified format.
-func EncoderFor(format string) (runtime.Encoder, error) {
+func EncoderFor(format string, gv schema.GroupVersion) (runtime.Encoder, error) {
```
skriss: I need to look at this in a little more detail.

pkg/install/install.go
```diff
@@ -0,0 +1,74 @@
...
+	"Deployment":               "deployments",
+	"DaemonSet":                "daemonsets",
+	"BackupStorageLocation":    "backupstoragelocations",
+	"VolumeSnapshotLocation":   "volumesnapshotlocations",
+}
+
+// Install creates resources on the Kubernetes cluster.
+func Install(factory client.DynamicFactory, resources *unstructured.UnstructuredList, logger *logrus.Logger) error {
+	for _, r := range resources.Items {
+		// TODO: do we want to use the logger, or just a print?
```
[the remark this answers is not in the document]
nrb: Switched this to an `io.Writer` so that `pkg/install` could be used as a library, too. I think this strikes a decent balance for our use case and what others might use it for.

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// BindFlags adds command line values to the options struct.
+func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
+	flags.StringVar(&o.ProviderName, "provider", o.ProviderName, "provider name for backup and volume storage")
+	flags.StringVar(&o.BucketName, "bucket", o.BucketName, "name of the object storage bucket where backups should be stored")
+	flags.StringVar(&o.Secret, "secret", o.Secret, "file containing credentials for backup and volume provider")
+	flags.StringVar(&o.Image, "image", o.Image, "image to use for the Velero and restic server pods. Optional.")
+	flags.StringVar(&o.Prefix, "prefix", o.Prefix, "prefix under which all Velero data should be stored within the bucket. Optional.")
+	flags.Var(&o.BackupStorageConfig, "backup-location-config", "configuration to use for the backup storage location. Format is key1=value1,key2=value2")
```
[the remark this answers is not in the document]
nrb: Fairly sure they're required for some providers/buckets (AWS needs it for S3), but not for others (Google buckets).

I'll see if I can come up with some concise wording to convey that, but it's going to largely depend on the provider and type of bucket used.

pkg/install/resources.go
nrb: I had this in the `appendUnstructured` function; I'll move it there again.

pkg/cmd/cli/install/install.go
nrb: I actually put this further down the call chain, since we use `output.ValidateFlags`. I'll see if I can move it into that function.

pkg/install/install.go
nrb: Hmm...so in this case, we'd need a group for each type? CRDs have the most, while the rest are 1 object. This also means creating different structs for printing and submitting to the API, right? Our current printer doesn't like using `[]*unstructured.Unstructured` directly, instead needing a list.

I do prefer the idea of doing a wait vs sleep, though.

pkg/install/install.go
nrb: Yeah, I was wondering about this too. It would indeed be nice.

It'd also be nice if, at the end of all this, we waited up to 1m or so to see if the deployment became `Ready`, so the user knew if things were ok or not.

pkg/install/install.go
nrb: If we terminate at first failure, should we walk back and remove all resources already created, so we leave the cluster in a consistent state?

pkg/install/install.go
skriss: You don't need a group per type - you need the CRDs in one group with an appropriate `Ready`, and then any CR instances in a second group.  All the rest of the stuff could go into the CRD group since they're not dependent on the CRDs being ready.

You can make Resources an unstructuredlist, that's not an issue

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,187 @@
...
+	DryRun               bool
+	BackupStorageConfig  flag.Map
+	VolumeSnapshotConfig flag.Map
+}
+
+// BindFlags adds command line values to the options struct.
+func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
+	flags.StringVar(&o.ProviderName, "provider", o.ProviderName, "provider name for backup and volume storage")
+	flags.StringVar(&o.BucketName, "bucket", o.BucketName, "name of the object storage bucket where backups should be stored")
+	flags.StringVar(&o.Secret, "secret", o.Secret, "file containing credentials for backup and volume provider")
```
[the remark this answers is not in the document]
skriss: One other thought here - for Azure we currently use env vars to populate the secret, not a file. Will need to make sure it works for that, somehow.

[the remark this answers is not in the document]
nrb: Yep, I need to look at that more closely still.

pkg/cmd/cli/install/install.go
nrb: Nope. I actually removed it, since we're not going to allow namespace customization.

pkg/install/install.go
```diff
@@ -46,42 +51,191 @@ var kindToResource = map[string]string{
...
+			}
 		}
+		return true, nil
+	})
+	return isReady, err
+}
+
+// GroupResources groups resources into ResourcesGroups based on whether the resources are CustomResourceDefinitions or other types of kubernetes objects
+// This is useful to wait for readiness before creating CRD objects
+func GroupResources(resources *unstructured.UnstructuredList) []*ResourceGroup {
```
nrb: @skriss The logic in this for loop appears to work with the print statements inside the loop, but outside the loop, the `crdObjs` slice is only comprised of the `Deployment/velero` object. That's also the last element seen from `resources.Items`.

My suspicion is that this is a variable shadowing issue, but I'm not seeing where that might be occurring.

skriss: Guessing it's from this line:
https://github.com/heptio/velero/pull/1287/commits/9ae653d712dfdb2974011811b4407d0960c4eba7#diff-90259a58ac8d65a2b486198e4eebb9c3R165

And this line:
https://github.com/heptio/velero/pull/1287/commits/9ae653d712dfdb2974011811b4407d0960c4eba7#diff-90259a58ac8d65a2b486198e4eebb9c3R174

Taking the address of the range variable (`r` in this case) is always a recipe for bugs like this. Easiest thing to do is switch your loop to `for i, r := resource.Items {`, and then in your appends, instead of `&r`, append `resources.Items[i]`. 

The problem is that the memory allocated to `r` is used over and over again; so you're adding pointers to the same memory address over and over, which at the end of the loop will only have the last thing that was in that memory.

nrb: Yep, that appears to have been it! Thanks.

nrb:
Still need to look at Azure, but I've adding resource waits instead of sleeps. Also switched to an options struct for the `AllResources` function, and the restic flag.

nrb:
I started looking at Azure today. I think I may want to defer Azure support to a separate PR.

Here's why:

We currently use the `--from-literal` method for creating the environment variable secrets. I don't want to do that for install if I can avoid it; I'd rather just point to a file. Luckily, there is a `--from-env-file` parameter for creating secrets in `kubectl`. This basically takes a [Docker .env](https://docs.docker.com/compose/env-file/) file and encodes the values. The downside is that all this logic is inside the `kubectl` [codebase](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/generate/versioned/secret.go#L235) - to replicate it, we either have to duplicate or vendor that code (unless I'm missing something).

I'd like to take our current secret-creating function to process the file based on a `fromEnv` parameter, but I also don't want to hold up this PR while we get that figured out.

What do you think @skriss and @carlisia?

skriss:
I like the idea of using an env file for the Azure variables. I'm fine tackling it in a separate PR if you prefer. I think it probably makes sense to just write our own env-file-parsing function, since the format is pretty simple.

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,204 @@
...
+
+// InstallOptions collects all the options for installing Velero into a Kubernetes cluster.
+type InstallOptions struct {
+	Namespace            string
+	Image                string
+	BucketName           string
+	Prefix               string
+	ProviderName         string
+	RestoreOnly          bool
+	Secret               string
```
skriss: rename to SecretFile

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,204 @@
...
+	if o.BucketName == "" {
+		return errors.New("--bucket is required")
+	}
+
+	if o.ProviderName == "" {
+		return errors.New("--provider is required")
+	}
+
+	if o.Secret == "" {
+		return errors.New("--secret is required")
```
skriss: --secret-file

pkg/install/install.go
```diff
@@ -0,0 +1,246 @@
...
+			return false, errors.Wrap(err, "error converting deployment from unstructured")
+		}
+
+		for _, cond := range deploy.Status.Conditions {
+			if isAvailable(cond) {
+				readyObservations++
+			}
+		}
+		// Make sure we query the deployment enough times to see the state change, provided there is one.
+		if readyObservations > 4 {
```
skriss: Just to make sure I'm following here - we won't get our first `readyObservation` until 10 seconds after the `Available` condition, then assuming all is well we'll get another one each second, so in the normal case, ~15 seconds after the deployment shows `Available`, we'll report it's good to go?

pkg/install/install.go
```diff
@@ -0,0 +1,246 @@
...
+			isReady = true
+		} else {
+			return false, nil
+		}
+		return true, nil
```
skriss: can this move up into just below L166?

pkg/install/install.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// GroupResources groups resources into ResourcesGroups based on whether the resources are CustomResourceDefinitions or other types of kubernetes objects
+// This is useful to wait for readiness before creating CRD objects
+func GroupResources(resources *unstructured.UnstructuredList) []*ResourceGroup {
+	crdObjs := new(ResourceGroup)
+	crdObjs.Ready = crdsAreReady
+	crdObjs.UnreadyMessage = "CRDs were not fully ready to create objects"
+
+	otherObjs := new(ResourceGroup)
```
skriss: It would be nice to wait for the restic daemonset to be "ready", too, if it's being installed - have to think about what that would look like.

pkg/cmd/cli/install/install.go
```diff
@@ -0,0 +1,204 @@
...
+	clientConfig, err := client.Config("", "", fmt.Sprintf("%s-%s", c.Parent().Name(), c.Name()))
+	if err != nil {
+		return err
+	}
+	dynamicClient, err := dynamic.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
+
+	err = install.Install(client.NewDynamicFactory(dynamicClient), resources, os.Stdout)
```
skriss: I'm wondering if it would make sense to expose a `--wait` flag, and only to wait for the Deployment (and possibly DaemonSet, see below) to become ready if that's true. WDYT?

pkg/install/install.go
```diff
@@ -0,0 +1,246 @@
...
+	gvk := schema.FromAPIVersionAndKind(apiextv1beta1.SchemeGroupVersion.String(), "CustomResourceDefinition")
+	apiResource := metav1.APIResource{
+		Name:       kindToResource["CustomResourceDefinition"],
+		Namespaced: false,
+	}
+	c, err := factory.ClientForGroupVersionResource(gvk.GroupVersion(), apiResource, "")
+	if err != nil {
+		return false, errors.Wrapf(err, "Error creating client for CustomResourceDefinition polling")
+	}
+	var isReady bool
```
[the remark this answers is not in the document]
nrb: L166?

pkg/install/install.go
nrb: Yes, that's accurate.

The goal is to make sure that the `Deployment` has been ready longer than the transient ready condition I observed when providing an invalid configuration. The `Deployment` will be ready pretty much immediately, when the pod is created, then the pod will get into a CrashLoopBackOff due to the `velero` process failing. Once that happens, the `Deployment` will no longer be ready.

pkg/cmd/cli/install/install.go
nrb: I'm thinking that might be a good idea, too. It simplifies the actual install logic. I still think we need the wait for the CRDs, though.

Thinking about usage, if the user doesn't provide `--wait` and there's an issue, how would they know? They'll have to do `kubectl logs deploy/velero -n velero` and/or `kubectl logs ds/restic -n velero` to check it. Maybe we output those commands on completion, kind of like we do with `backup create`.

pkg/install/install.go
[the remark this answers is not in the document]
nrb: Oh, nvm. Yes, this is a left over.

pkg/util/encode/encode.go
skriss: Alright, I wrapped my head around this - let me write up some notes.

skriss: So - I *think* our primary goal here is that, when we serialize objects to JSON/YAML and output them, via `velero backup get`, `velero install`, etc., we need to have the `Kind` and `APIVersion` fields (i.e. the `TypeMeta`) filled in.  Specifically for `velero install`, this enables the output to be `kubectl apply`-ed.

`encode.go` today only knows about the Velero types. So, if it's given a structured object that's a Velero type, it will encode it and include the `TypeMeta`. If it's given an unstructured object without its `TypeMeta` filled it, it will encode it, but won't add the `TypeMeta`. And if it's given a structured object that's not a Velero type (e.g. a `Deployment`), it won't encode it.

One option here is to enhance `encode.go` to know about other non-Velero types, e.g. the Kube types that we care about. We'd do this along the lines of what you have already in this PR, though there are a couple things that should change. If we do this, then we can encode all objects that we care about, without issue.  The drawback to this option is that, since you're outputting *unstructured* objects during `velero install`, the encoder can't help us by filling in the `TypeMeta` -- we always have to set it explicitly ourselves when creating the unstructured objects.

The other option is to switch to outputting structured objects. We'd have to modify `output.go` -- my initial thought is that we'd switch from:
```go
func PrintWithFormat(c *cobra.Command, obj runtime.Object) (bool, error)
```
to
```go
func PrintWithFormat(c *cobra.Command, objs ...runtime.Object) (bool, error)
```

That way, we could just pass a list of structured objects to it to output. It would have to be enhanced to add separators as appropriate so the overall output could still be `kubectl apply`-ed.  The benefit of this approach is that we don't have to explicitly set `TypeMeta` anywhere - if we use the structured representations, and we set up the `runtime.Scheme` properly in `encode.go`, then it'll automatically get filled in during encoding.

My gut says the second option is more ideal, though refining the first one is probably less work at the moment. WDYT?

FYI here's my hacked-up modification to `encode.go`, that supports both encoding unstructured objects with their `TypeMeta` filled in, **and** encoding Velero & Kube structured objects without `TypeMeta` filled in:
```go
// EncoderFor gets the appropriate encoder for the specified format.
func EncoderFor(format string, obj runtime.Object) (runtime.Encoder, error) {
	scheme := runtime.NewScheme()
	vscheme.AddToScheme(scheme) // vscheme "github.com/heptio/velero/pkg/generated/clientset/versioned/scheme"
	kscheme.AddToScheme(scheme) // kscheme "k8s.io/client-go/kubernetes/scheme"

	codecs := serializer.NewCodecFactory(scheme)

	var encoder runtime.Encoder
	desiredMediaType := fmt.Sprintf("application/%s", format)
	serializerInfo, found := runtime.SerializerInfoForMediaType(codecs.SupportedMediaTypes(), desiredMediaType)
	if !found {
		return nil, errors.Errorf("unable to locate an encoder for %q", desiredMediaType)
	}
	if serializerInfo.PrettySerializer != nil {
		encoder = serializerInfo.PrettySerializer
	} else {
		encoder = serializerInfo.Serializer
	}

	
	if gvks, _, _ := scheme.ObjectKinds(obj); len(gvks) > 0 {
		return codecs.EncoderForVersion(encoder, gvks[0].GroupVersion()), nil
	}

	return encoder, nil
}
```

pkg/cmd/cli/install/install.go
skriss: Yeah, agree we need to wait for the CRDs regardless, since not doing that could cause creation of BSL/VSL instances to fail.

Yeah -- I think outputting the relevant `kubectl get` commands and `kubectl logs` commands would make sense there, and we could output some text saying that the API objects have been created, but we haven't verified they're up and running.

pkg/util/encode/encode.go
nrb: I would prefer to use the structured types, if possible. What exists now was trying to find a way to pass a single `UnstructuredList` (since nothing else besides `[]runtime.Object` could hold everything` in and have the current printing machinery 'just work'.

I think the structured approach allows the whole of the install code base to use `[]runtime.Object` instead of the `UnstructuredList`, which is also a win. That means that I'll have to redo some of the work we have right now, but longer term it's more maintainable as we'll get more compiler help.

I'll double check if the TypeMeta fields for anything in the `pkg/install` package are used - I think it makes sense to leave those populated (for things like filtering/grouping), but not to add TypeMeta to all the other, existing create commands.

nrb: I've got `PrintWithFormat` modified to take the variadic list, but I'll need to think some on how to properly print out the installed resources as JSON. With the in-tree types it's fine, since they're in a list type. Printing things one at a time is fine in YAML, since they can be broken up with a single line. JSON needs to wrap the whole list as an array, and can't have a comma on the last item.

I'll do more research to see if we can manage to do this without making it too gross.

nrb: How important is omitting the empty `status` fields? It looks like by using the structured types we're going to have those output. This is true of our current create commands with our own data types.

skriss: Ah, yeah, having `status` be output is annoying.  Does it cause issues with `kubectl apply`-ing the resources? Or is it just a UX thing? Let me think about this some more..

nrb: Looking back at my notes, it does for CustomResourceDefinitions. One of the status fields comes in as null and causes an error. I believe those are the only resources that require removing it, the rest is cosmetic.

skriss: Hmm. I'm thinking maybe we shouldn't try to switch to non-unstructured objects right now - we can get this PR merged more or less as-is (I'll add a few specific comments to `encode.go`) - and then look at switching afterwards.

nrb: That seems reasonable to me. I think switching `EncoderFor` to use the logic you've provided here is probably a better fix for our other create commands than populating them all with TypeMeta manually.

skriss:
@nrb I was thinking about Azure some more -- right now the secret is created from env vars on the client machine, which then get loaded into the velero pod as env vars. However, the Azure SDK doesn't actually automatically consume these env vars - we have velero code that reads them and passes them as normal parameters to the Azure SDK. So - we could change the structure of the secret for Azure, so it just stores the contents of a .env file, and then we could mount it into the velero pod as a file, like with AWS and GCP. We'd then just have to update velero code to parse the file, and pass the appropriate values to the Azure SDK.

This would mean secret creation & usage would be no different than for AWS and GCP - we'd have the user create a .env file on their local machine with the appropriate values, and then the secret would be created from that file. Similarly, the secret would mount into the velero pod as a file, not as env vars.

WDYT? I think we could probably tackle this for 1.0 and it would help with both `velero install` and the Helm chart (I'm guessing).

nrb:
That makes sense to me. I think it definitely simplifies our deployments.

The Helm chart for v0.11 wouldn't change, I don't think, but we can with the 1.0 one.

nrb:
Pulling this out of draft. Can still fix anything outstanding, and need to get docs done/tested.

pkg/util/encode/encode.go
```diff
@@ -63,6 +65,6 @@ func EncoderFor(format string) (runtime.Encoder, error) {
 	} else {
 		encoder = serializerInfo.Serializer
 	}
-	encoder = scheme.Codecs.EncoderForVersion(encoder, v1.SchemeGroupVersion)
+	encoder = scheme.Codecs.EncoderForVersion(encoder, gv)
```
skriss: Here, I'd check if the `runtime.Object` argument has a non-empty GroupVersion. If so, I'd use this `EncoderForVersion` func to get the encoder. If not, I'd do something similar to my code sample (use `scheme.ObjectKinds(obj)`) to try to find the `GroupVersionKind` for the object (this would apply if it's a strongly-typed object, that doesn't have `TypeMeta` filled in).  Have to decide how to handle an `error` or <1 GVK being returned from that function -- probably just return the generic `serializerInfo.PrettySerializer` or `serializerInfo.Serializer`.

nrb: I tried out returning the generic encoder when no GVK was found by removing the velero scheme and running `velero backup create` - it still worked, even without the type meta attached.

I don't believe `scheme.ObjectKinds` can return a slice of len(0), due to the fact that all the versions have to be explicitly registered in a map, and the `Kind` key missing from that map generates a not found error.

nrb:
I think I've addressed all comments, and `make ci` passes locally.

Docs are still pending, but I think we can merge without them if we want this in 1.0-alpha1.

pkg/util/encode/encode.go
```diff
@@ -42,19 +43,24 @@ func Encode(obj runtime.Object, format string) ([]byte, error) {
...
 	if err != nil {
 		return err
 	}
-
 	return errors.WithStack(encoder.Encode(obj, w))
 }
 
 // EncoderFor gets the appropriate encoder for the specified format.
-func EncoderFor(format string) (runtime.Encoder, error) {
+func EncoderFor(format string, obj runtime.Object) (runtime.Encoder, error) {
```
skriss: I'm still not loving this function as-is -- if you pass it a structured object, with the `TypeMeta` filled in, and it's not one of the `vscheme` or `kscheme` types, the encoder won't produce any output. And, if it's a recognized type, without `TypeMeta`, the encoder will be for a random version for the API group (if there are multiple).

Stepping back, right now we're only ever passing this function (a) structured Velero objects, or (b) unstructured objects with the `TypeMeta` filled in.

Perhaps we should go back to the original version of this function, but with one addition - before L66 (`encoder = scheme.Codecs.EncoderForVersion(encoder, v1.SchemeGroupVersion)`), we add:

```go
if !obj.GetObjectKind().GroupVersionKind().Empty() {
    return encoder
}
```

That way, if this function is passed an object, structured or unstructured, with its `TypeMeta` filled in, it will be properly encoded, retaining its `TypeMeta`. Otherwise, it'll behave the same as it does today -- if passed a structured object that's in the velero scheme, it'll encode it; structured non-velero objects produce no output.

skriss:
Other than the one comment on `EncoderFor`, I'm ready to merge this for the alpha.  It looks like one other change will have to be made to `pkg/install` to get this to work for Azure (assuming #1364 merges), to set the AZURE_CREDENTIALS_FILE env var.

skriss:
ran a partial test of `velero install` on Azure, manually dealing with the known issues, on top of #1364 -- looks like it all worked!

pkg/util/encode/encode.go
nrb: By original version of the function, you mean original but with passing in a `runtime.Object`?

skriss: yep. does this seem reasonable to you?

nrb: Yeah, it does - I kind of like that it's requiring non-Velero objects to be explicitly set. It is, after all, a Velero encoding function.

Testing that change seemed to retain functionality on both the install command and existing commands, so I'm for using the original code and just leaving a comment documenting this behavior.

nrb:
> manually dealing with the known issues

Which known issues? Should probably have a known issues doc for the alpha.

skriss:
@nrb I think it was just setting the Azure env var in the deployment (I did the test before I updated #1364 to set it in `pkg/install`).  I'll retest on Azure once everything gets merged.

Issue #1282, opened 2019-03-13
Split pkg/plugin up so plugin authors don't have to import things they don't need


skriss:
ref. #1147

skriss:
@carlisia let's discuss the proposed split here when you're ready. cc @nrb

carlisia:
PR II: plugin client plumbing
```
pkg/plugin/clientmgmt
    backup_item_action
    block_store
    object_store
    restore_item_action
```

PR II: alternative to above, preferred
```
pkg/plugin/clientmgmt
    backup_item_action_client
    backup_item_action_server
    block_store_client
    block_store_server
    object_store_client
    object_store_server
    restore_item_action_client
    restore_item_action_server
```


PR III: plugin server and related framework
```
pkg/plugin/framework
    client_builder
    client_dispenser
    interface
    logger 
    logrus_adapter
    manager
    plugin_base
    plugin_kind
    plugin_lister
    process
    registry
    restartable_*
    server_mux
    server
    stream_reader
```

skriss:
PR II (alt) looks good to me. I'd actually call that package `pkg/plugin/framework`, though.  In my mind this is the common package that both plugin authors *and* Velero core need to import.  And I think you'll need to move a few more things in there - `server`, `server_mux`, `logger`, `client_dispenser`, maybe one or two other things. The goal for this package is to not have it import anything from PR 3's package. 

PR III -- I'd call that package `pkg/plugin/clientmgmt` -- since Velero is the "client" when it comes to plugins.  It's fine for this package to import from PR II's package, but not vice-versa.  I agree with most of the items in here, with the few exceptions noted above.  There may be some additional dependencies that need to be untangled as we go through it, though.

carlisia:
I appreciate the explanation! Copy that, will start on PR II.

PR #1288, opened 2019-03-15
Split plugin framework into its own package

carlisia:
This addresses the first part of #1282.


carlisia:
Note I: next commit will have the framework code split up into client/server files.

carlisia:
Note II:

I can't figure out just yet why the godoc for the new package `framework` is showing multiple instances of the copyright notice.

image

carlisia:
Fixed documentation. Ready for review now.

image

pkg/plugin/framework/backup_item_action_client.go
```diff
@@ -0,0 +1,126 @@
...
+
+	api "github.com/heptio/velero/pkg/apis/velero/v1"
+	proto "github.com/heptio/velero/pkg/plugin/generated"
+	"github.com/heptio/velero/pkg/plugin/velero"
+)
+
+// BackupItemActionPlugin is an implementation of go-plugin's Plugin
+// interface with support for gRPC for the backup/ItemAction
+// interface.
+type BackupItemActionPlugin struct {
```
skriss: this type is technically neither client nor server - it exposes the `GRPCClient` and `GRPCServer` methods. I'd probably opt to move it into its own file (and similarly for the 3 other plugin types).

pkg/plugin/framework/backup_item_action_server.go
```diff
@@ -14,126 +14,21 @@ See the License for the specific language governing permissions and
...
-
-	return &updatedItem, additionalItems, nil
-}
-
-//////////////////////////////////////////////////////////////////////////////
-// server code
-//////////////////////////////////////////////////////////////////////////////
-
 // GRPCServer registers a BackupItemAction gRPC server.
 func (p *BackupItemActionPlugin) GRPCServer(s *grpc.Server) error {
```
skriss: per previous comment, this method is part of the `BackupItemActionPlugin` type which I'd probably opt to move into its own file rather than having it exist across the client and server files.

pkg/plugin/framework/examples_test.go
```diff
@@ -0,0 +1,95 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import (
+	"github.com/aws/aws-sdk-go/service/ec2"
```
skriss: can we drop the `ec2` reference here since it's not actually used?

pkg/plugin/framework/server.go
```diff
@@ -14,7 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package plugin
+// Package framework is the common package that any plugin client
```
skriss: Could we move the package-level godoc into its own file (`doc.go`) and maybe include a small code sample of how to use the package (specifically calling the NewServer method, then registering a plugin or two, then serving)?

pkg/plugin/framework/backup_item_action_client.go
carlisia: By `into its own file` do you mean name the file `backup_item_action.go` instead of `backup_item_action_client.go`? And then when I split this into `client/server` packages I'll duplicate it?

skriss: I'd move `type BackupItemActionPlugin struct` into a file called `backup_item_action.go`, and leave `type BackupItemActionGRPCClient struct` in `backup_item_action_client.go`. And yeah, if/when we do the client/server split, then each of those packages would end up with a `backup_item_action.go` file/struct, with slightly different implementations.

skriss:
This LGTM - squash/signoff and we can merge

Issue #638, opened 2018-07-03
Allow admin to tweak restic restore helper image


ncdc:
**User Story**

As a user of Ark with Restic, I would like to customize the restic container image that is deployed as an initContainer to Pods during a restore.

---

https://github.com/heptio/ark/blob/32907931e13f38ef4e055652245eeb78d20ac76e/pkg/restore/restic_restore_action.go#L53

Make it a flag to `ark server`?

rosskukulinski:
As this isn't a breaking change, I don't think it must land in v1.0.0.  Community contributions welcome if someone wants to accelerate the development of this feature.

ljfranklin:
@ncdc @rosskukulinski I'd be willing to try a PR to add this flag. I'm trying to use Ark on a Raspberry Pi, the backup is working fine but the restore stumbles due to the restic helper image not supporting ARM. That said, I'm having trouble figuring out how to pass the image name from [server.go](https://github.com/heptio/ark/blob/2781e4e8de08bb33585eb945819ad989bf972b49/pkg/cmd/server/server.go#L166) down to [restic_restore_action.go](https://github.com/heptio/ark/blob/2781e4e8de08bb33585eb945819ad989bf972b49/pkg/restore/restic_restore_action.go#L53). It looks like this restic restore object is initialized in a separate [`run-plugins` command](https://github.com/heptio/ark/blob/2781e4e8de08bb33585eb945819ad989bf972b49/pkg/cmd/server/plugin/plugin.go#L55) and it seems awkward to pass the restic image name as a command line arg to `run-plugins`. Is there some config object that would make sense to hold a `resticRestoreHelperImage` property?

patrickstjohn:
+1 on this. This issue prevents restoration of restic backups on any internet segregated cluster.

encryptedfear:
Hi,

If I made canges to the `ark/pkg/restore/resti_restore_action.go`, will it take effect on my next backup? Or is there another step that I've to do? I can't pull the image from gcr.io/heptio-images/ark-restic-restore-helper:v0.10.0 due to internet restrictions in our cluster.

ncdc:
@encryptedfear did you build a new ark image with your changes?

Another short-term option would be to pre-pull our image into your environment, then pre-load it onto all your nodes. I think that might work - unless the `ImagePullPolicy` (which we aren't setting) gets set to `Always`.

encryptedfear:
Hi,

[...]

Issue #584, opened 2018-06-25
Define how to enable/disable and configure plugins


ncdc:
For example, some users might not want to enable the plugins that automatically walk from pod to pvc to pv.

ncdc:
Let's say we want to enable/disable and configure the "foo.com/pod" BackupItemAction. Possible options:

- Create a ConfigMap in the heptio-ark namespace, the name can be whatever:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo-com-pod-config
  labels:
    ark.heptio.com/plugin-config: foo.com/pod
data:
  enabled: false
  addAnnotations:
    color: green
    size: small
```

Let's say we just want to disable a plugin that has no configuration:

- add `--enable-plugins NAME(s)` and/or `--disable-plugins NAME(s)` flags to the `ark server`
- use ConfigMaps and Ark is responsible for checking for an enable/disable annotation:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo-com-pod-config
  labels:
    ark.heptio.com/plugin-config: foo.com/pod
    ark.heptio.com/plugin-enabled: false
```


Proposed approach:
1. Admin uses `--enable-plugins` and/or `--disable-plugins` flag for the `ark server` to control which plugins are enabled
1. Admin and/or users can create ConfigMaps to configure their plugins, assuming plugins support configuration

carlisia:
Un-assigned myself since I'm not currently working on this.

PR #1290, opened 2019-03-15
restic restore helper image can be configured with cli flag

ryane:
This is an attempt to address #638. I [also](https://github.com/heptio/velero/issues/638#issuecomment-437639874) did not find it straightforward to pass the image from a velero server argument to the plugin command. So, I *tried* to keep changes to a minimum and to just get something working but I'm sure this could be improved. I'd be happy to continue to refine this PR but would love to get some feedback before spending too much time on it. I'd really just like to help get a fix for this so that we can use an official velero build in our on-prem, air gapped environment.

### DIFF
--- a/pkg/restore/restic_restore_action.go
+++ b/pkg/restore/restic_restore_action.go
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+		http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


pkg/restore/restic_restore_action.go
```diff
@@ -36,10 +36,14 @@ type ResticRestoreAction struct {
 	initContainerImage string
 }
 
-func NewResticRestoreAction(logger logrus.FieldLogger) *ResticRestoreAction {
+func NewResticRestoreAction(logger logrus.FieldLogger, initContainerImageName string) *ResticRestoreAction {
+	if initContainerImageName == "" {
```
nrb: I'll have to think a bit more on how to accomplish this, but it'd be awesome if the `initContainerImage` function could just return the image from the args.

skriss:
Thanks for this @ryane. We should consider the approach here as part of #584 - do we want to use ConfigMaps instead of CLI flags for configuring plugins? We shouldn't have to change code every time a new flag/config option is added. cc @carlisia

carlisia:
This looks good @ryane. If it could be made to use ConfigMap for this plugin config it would be better.

skriss:
@ryane we decided to go with a ConfigMap-based approach since it will be more extensible for generic configuration of all plugins. See #1311 which recently merged. Appreciate the PR regardless!

ryane:
excellent, thx!

PR #1291, opened 2019-03-15
Split velero plugin client into its own package

### LINKED ISSUE #1282: Split pkg/plugin up so plugin authors don't have to import things they don't need

carlisia:
Closes #1282

This addresses the second part of #1282.


skriss:
This looks good. https://github.com/heptio/velero/pull/1288#pullrequestreview-215672247, if implemented, would have some impact on this -- and if we go that route, I wonder if we should just go with `pkg/plugin/server` and `pkg/plugin/client` as names?

carlisia:
Since we're not gonna do the client/server split now, then yes this closes #1282.

PR #1301, opened 2019-03-19
log error locations from plugin logger and don't overwrite in client

skriss:

Add the "error location" hook to the server-side plugin logger, so that if a plugin author logs an error from within their plugin, the location of that error is included in the log statement as fields. Also update the hook to not overwrite this data if it already exists as fields on the entry, so that log entries coming from a plugin server don't get overwritten on the client.

Pulled from #1119.

### DIFF
--- a/changelogs/unreleased/1301-skriss
+++ b/changelogs/unreleased/1301-skriss
@@ -1,0 +1,1 @@
+log error locations from plugin logger, and don't overwrite them in the client logger if they exist already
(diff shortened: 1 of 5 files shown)


skriss:
updated years in license headers.

carlisia:
Good catch 👍

skriss:
I've got a WIP script for automatically updating them through `make update`, will try to get it in soon.

skriss:
@nrb this one's ready to go in if you're happy with it

PR #1303, opened 2019-03-20
Update go-plugin

### LINKED ISSUE #1102: Upgrade go-plugin
Newer commits include protocol version negotiation, and fixes to make Kill() idempotent

carlisia:
This closes #1102.


skriss:
@carlisia what version of `dep` do you have installed locally? (`dep version`) Looks like maybe an older version - if so, can you please update to v0.5.0?

skriss:
Oh, maybe you upgraded partway through this PR? I was just looking at the first commit.

Gopkg.toml
```diff
@@ -116,3 +116,15 @@
 [[constraint]]
   name = "github.com/hashicorp/go-plugin"
   branch = "master"
```
skriss: let's pin this to the specific revision we're using now - so instead of `branch = "master"`, use `revision = "3f118e8ee104b6f22aeb12453fab56aed1356186"`

Gopkg.toml
```diff
@@ -116,3 +116,15 @@
...
+
+[[constraint]]
+  name = "github.com/golang/protobuf"
+  branch = "master"
```
skriss: let's pin this to a specific semver that's compatible with the go-plugin commit - so instead of `branch = "master"`, `version = "~v1.3.1"`

Gopkg.toml
```diff
@@ -116,3 +116,15 @@
...
+
+[[constraint]]
+  name = "google.golang.org/grpc"
+  branch = "master"
```
skriss: let's pin this to a specific semver that's compatible with the go-plugin commit - so instead of `branch = "master"`, `version = "~v1.19.0"`

carlisia:
I updated partway thru. Should I redo it?

carlisia:
And, how can you tell?

skriss:
Nope, it's fine, end result is good

skriss:
The newest version adds some new fields to the `Gopkg.lock` (digest, pruneopts).  In the first commit, those had all been removed. They were added back in the second.

PR #1304, opened 2019-03-20
add additionalItems to restore item actions

### LINKED ISSUE #314: Restoration with label selector does not properly restore PVs

skriss:
Co-authored-by: Andy Goldstein <goldsteina@vmware.com>

Fixes #314

Rebase/replacement of #1125


skriss:
Couple of quick tests on this look good, but I need to do some more testing and self-review.

pkg/restore/restore.go
```diff
@@ -708,251 +727,335 @@ func (ctx *context) restoreResource(resource, namespace, resourcePath string) (a
...
+			var itemPath string
+			if additionalItem.Namespace == "" {
+				itemPath = filepath.Join(ctx.restoreDir, api.ResourcesDir, additionalItem.GroupResource.String(), api.ClusterScopedDir, additionalItem.Name+".json")
+			} else {
+				itemPath = filepath.Join(ctx.restoreDir, api.ResourcesDir, additionalItem.GroupResource.String(), api.NamespaceScopedDir, additionalItem.Namespace, additionalItem.Name+".json")
 			}
 
-			ctx.log.Infof("Executing item action for %v", &groupResource)
+			if _, err := ctx.fileSystem.Stat(itemPath); err != nil {
+				ctx.log.WithError(err).WithFields(logrus.Fields{
```
skriss: TODO is logging this as a warning the right thing to do? might want to add it to the restore results?

skriss:
Alright, @carlisia @nrb this is ready for review. I'd still like to do a little more testing, but so far it looks good, and I've gone through the code in detail again myself.

pkg/restore/restore.go
carlisia: Since we already are returning warnings, then it makes more sense to return all warnings instead of returning some and login some.

pkg/restore/add_pvc_from_pod_action.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	var additionalItems []velero.ResourceIdentifier
+
+	for _, volume := range pod.Spec.Volumes {
+		if volume.PersistentVolumeClaim == nil {
+			continue
+		}
+
+		a.logger.Infof("Found a PVC we need to back up: %s/%s", pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
```
nrb: Small thing, but the language in the log message could be changed to be more consistent with the PV plugin. `Adding PVC %s/%s as an additional item to restore` seems like it would work.

skriss:
Thanks for doing testing. I'll work on some unit tests today.

skriss:
I added units

nrb:
Double checked the latest changes with restic, and the PV was provisioned as expected.

skriss:
I've continued with testing here and things look good to me -- related unlabeled PVs are restored when a label-selector is used; restic looks good; regression scenarios all look fine.

nrb:
Looks like this will need a rebase now that other changes have merged.

skriss:
rebased - will walk through this before merging

PR #1308, opened 2019-03-21
update plugins to work with updated go-plugin

skriss:

Turns out #1303 introduced some runtime breaking changes.  the `GRPCPlugin` interface exported by go-plugin changed -- added context and broker fields.  Unfortunately, it wasn't a compile-time error due to how the types work.  I updated our implementations and added type checks for the future.

cc @carlisia

### DIFF
--- a/pkg/plugin/framework/backup_item_action.go
+++ b/pkg/plugin/framework/backup_item_action.go
@@ -35,12 +35,12 @@ type BackupItemActionPlugin struct {
 var _ plugin.GRPCPlugin = &BackupItemActionPlugin{}
 
 // GRPCClient returns a clientDispenser for BackupItemAction gRPC clients.
-func (p *BackupItemActionPlugin) GRPCClient(ctx context.Context, broker *plugin.GRPCBroker, clientConn *grpc.ClientConn) (interface{}, error) {
+func (p *BackupItemActionPlugin) GRPCClient(_ context.Context, _ *plugin.GRPCBroker, clientConn *grpc.ClientConn) (interface{}, error) {
 	return newClientDispenser(p.clientLogger, clientConn, newBackupItemActionGRPCClient), nil
 }
 
 // GRPCServer registers a BackupItemAction gRPC server.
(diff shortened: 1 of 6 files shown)


pkg/plugin/framework/backup_item_action.go
```diff
@@ -31,13 +32,15 @@ type BackupItemActionPlugin struct {
 	*pluginBase
 }
 
+var _ plugin.GRPCPlugin = &BackupItemActionPlugin{}
+
 // GRPCClient returns a clientDispenser for BackupItemAction gRPC clients.
-func (p *BackupItemActionPlugin) GRPCClient(c *grpc.ClientConn) (interface{}, error) {
-	return newClientDispenser(p.clientLogger, c, newBackupItemActionGRPCClient), nil
+func (p *BackupItemActionPlugin) GRPCClient(ctx context.Context, broker *plugin.GRPCBroker, clientConn *grpc.ClientConn) (interface{}, error) {
```
carlisia: To give a clear indication that `context.Context` isn't being used, you can drop `ctx` and I forget, either `context.Context` or `_ context.Context`. Same with the others.

pkg/plugin/framework/backup_item_action.go
```diff
@@ -31,13 +32,15 @@ type BackupItemActionPlugin struct {
 	*pluginBase
 }
 
+var _ plugin.GRPCPlugin = &BackupItemActionPlugin{}
```
carlisia: 1- what does this do?
2- how did you figure out this was needed?

skriss: This ensures we get a compile-time error if `*BackupItemActionPlugin` doesn't implement the `plugin.GRPCPlugin` interface.  It's declaring a package-level variable, with no name (`var _`, of type `plugin.GRPCPlugin`, and assigning to it a `&BackupItemActionPlugin{}` instance.  If `&BackupItemActionPlugin{}` doesn't implement `plugin.GRPCPlugin`, then this line won't compile.

carlisia: Ah. Is there anyway we could do this in a test instead, to avoid having a pkg level variable?

skriss:
updated

PR #1311, opened 2019-03-22
allow restic restore helper image to be specified via ConfigMap

### LINKED ISSUE #638: Allow admin to tweak restic restore helper image

skriss:

Fixes #638 
xref #1290
xref #584 

Here's an alternate approach to allowing users to configure the restic restore helper image, via ConfigMap -- along the lines of what Andy proposed in #584.  I think I like this approach.  Here's what a ConfigMap would look like for this case:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    # the key is the unique name of the plugin, and the value is the plugin kind. 
    # Note that if we use namespaced names for plugins (i.e. with slashes/dots), 
    # they can't be label values, they have to be keys
    velero.io/restic: RestoreItemAction
  # the name doesn't matter, the code fetches based on the label. when we add 
  # CLI commands, we should use generated names
  name: restic-restore-action
  namespace: velero
data:
  # data can be arbitrary for each plugin. this particular one supports 
  # 'imageWithoutTag' or 'imageWithTag'. 
  imageWithoutTag: steveheptio/velero-restic-restore-helper
```

Eventually I'd want to add CLI commands along the lines of `velero plugin config set key=val --plugin-name NAME --plugin-kind KIND` that would actually create/update these ConfigMaps so users don't have to do so directly. That would be part of a full solution for #584, but not part of the MVP for addressing #638.

@carlisia @nrb @ryane thoughts?

### DIFF
--- a/pkg/restore/restic_restore_action.go
+++ b/pkg/restore/restic_restore_action.go
@@ -149,7 +149,7 @@ func getImage(log logrus.FieldLogger, config *corev1.ConfigMap) string {
 	switch {
 	case len(parts) == 1:
 		// tag-less image name: add tag
-		log.Debugf("Plugin config contains image name without tag")
+		log.Debugf("Plugin config contains image name without tag. Adding tag.")
 		return initContainerImage(image)
 	case len(parts) == 2:
 		// tagged image name
--- a/pkg/restore/restic_restore_action.go
+++ b/pkg/restore/restic_restore_action.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 the Velero contributors.
+Copyright 2018, 2019 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


nrb:
To make sure I understand your proposal, would the metadata for an out-of-tree plugin might look something like the below?

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    mysql.com/msqldump: BackupItemAction
  name: mysql-backup-action
  namespace: velero
```

It may also be useful to have a label on all of the plugin configmaps so we can query all of them at once, in case we start adding more configmaps into the namespace. Something like `velero.io/component=plugin`.

ncdc:
@skriss another option: have the plugin look for an environment variable that you'd specify in the pod definition. That works when there's a small number of configuration options, but would probably get messy when lots of plugins all have their own settings.

skriss:
@nrb yes, and agree, would be useful to have that common label. 

@ncdc yeah, makes sense - that's probably the easiest short-term solution.

skriss:
@nrb @carlisia please take another look at this - we'll most likely ship a v0.11.1 with this patch.

pkg/restore/restic_restore_action.go
```diff
@@ -123,3 +130,72 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
...
+		return initContainerImage(defaultImageBase)
+	}
+
+	log = log.WithField("image", image)
+
+	parts := strings.Split(image, ":")
+	switch {
+	case len(parts) == 1:
+		// tag-less image name: add tag
+		log.Debugf("Plugin config contains image name without tag")
```
carlisia: Would this be more useful: `s/Plugin config contains image name without tag/Plugin config contains image name without tag. Adding tag.`

skriss: wfm

bshas:
@skriss any update on patch v0.11.1 yet? if/when it will be shipped?

skriss:
@bshas this will definitely be included in our 1.0 release, which is targeted for around the end of May. We'll likely be shipping a v0.11.1 in the next few weeks, as well.

suji007:
Can someone tell me how to set limits for the restic init container pod via configmaps?

PR #1316, opened 2019-03-26
Explicitly set encrypt on EC2 CreateVolume

### LINKED ISSUE #1312: Velero omitting Encrypted param when calling ec2:CreateVolume results in ec2:Encrypted being evaluated as false
This is on a cluster where AWS permissions prevent the creation of a volume without encryption.

1. Provision velero `v0.11.0` on AWS, configure block storage provider with KMS key.
1. Create a backup containing EBS volumes encrypted to KMS key
1. Restore snapshot

Volumes should be restored

The AWS API responds with `403 UnauthorizedOperation` when executing `PVAction`. Upon decrypting the failure message the request contains the parameter `ec2:Ecnrypted` with the value set to false. I verified that the snapshot is indeed encrypted.

I originally raised this issue in [pr-211](https://github.com/heptio/velero/pull/211), and they've been on that fork ever since but they've been running into a controller bug that can cause old versions of ark to spam the EC2 API at a rate of thousands of requests per minute.

It's a secure environment, and they're super sketchy about publishing logs or request details in a public forum but I'm in your slack.

**Environment:**

- Velero version (use `velero version`): `v0.11.0`
- Kubernetes version (use `kubectl version`): `1.13.0`
- Kubernetes installer & version: `kops v1.11 heavily customized`
- Cloud provider or hardware configuration: `AWS`
- OS (e.g. from `/etc/os-release`): `RHEL 7.6`

mstump:
use describe to determine if snapshot is encrypted and pass that in EC2 call. Fixes #1312

### DIFF
--- a/pkg/cloudprovider/aws/block_store.go
+++ b/pkg/cloudprovider/aws/block_store.go
@@ -102,6 +102,7 @@ func (b *BlockStore) CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ s
 		SnapshotId:       &snapshotID,
 		AvailabilityZone: &volumeAZ,
 		VolumeType:       &volumeType,
+		Encrypted:        snapRes.Snapshots[0].Encrypted,
 		TagSpecifications: []*ec2.TagSpecification{
 			{
 				ResourceType: aws.String(ec2.ResourceTypeVolume),
--- a/changelogs/unreleased/1316-mstump
+++ b/changelogs/unreleased/1316-mstump
@@ -1,0 +1,1 @@
+Fix for #1312, use describe to determine if AWS EBS snapshot is encrypted and explicitly pass that value in EC2 CreateVolume call.


mstump:
Validated that the patch fixes the issue in the target environment. Ready for merge if you are.

pkg/cloudprovider/aws/block_store.go
```diff
@@ -81,7 +81,30 @@ func (b *BlockStore) Init(config map[string]string) error {
...
+		SnapshotIds: []*string{&snapshotID},
+	}
+
+	res, err := b.ec2.DescribeSnapshots(req)
+	if err != nil {
+		return false, errors.WithStack(err)
+	}
+
+	if len(res.Snapshots) != 1 {
+		return false, errors.Errorf("Expected one snapshot from DescribeSnaphots for snapshot ID %v, got %v", snapshotID, len(res.Snapshots))
```
skriss: - pls lowercase `Expected`
- s/DescribeSnaphots/DescribeSnapshots

pkg/cloudprovider/aws/block_store.go
```diff
@@ -81,7 +81,30 @@ func (b *BlockStore) Init(config map[string]string) error {
...
+	}
+
+	snapshot := res.Snapshots[0]
+	return *snapshot.Encrypted, nil
```
skriss: `return aws.BoolValue(snapshot.Encrypted), nil` so we avoid NPE

pkg/cloudprovider/aws/block_store.go
```diff
@@ -81,7 +81,30 @@ func (b *BlockStore) Init(config map[string]string) error {
...
+}
+
 func (b *BlockStore) CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ string, iops *int64) (volumeID string, err error) {
+	encrypted, err := b.isSnapshotEncrypted(snapshotID)
+	if err != nil {
+		return "", errors.WithStack(err)
```
skriss: you can just `return "", err` here since the error coming from `isSnapshotEncrypted` will already have a stack

mstump:
Review feedback incorporated

pkg/cloudprovider/aws/block_store.go
```diff
@@ -102,6 +125,7 @@ func (b *BlockStore) CreateVolumeFromSnapshot(snapshotID, volumeType, volumeAZ s
 		SnapshotId:       &snapshotID,
 		AvailabilityZone: &volumeAZ,
 		VolumeType:       &volumeType,
+		Encrypted:        &encrypted,
```
skriss: Sorry I didn't notice this on the first pass, but we don't even need the `isSnapshotEncrypted` helper anymore since we're already calling `DescribeSnapshots` just above to get the tags anyway.  This whole PR can be collapsed into editing this line to say:

```go
Encrypted: snapRes.Snapshots[0].Encrypted,
```

mstump: OK, updated to include new change.

skriss:
Fixes #1312

PR #1318, opened 2019-03-27
remove Warning from restore item action output

### LINKED ISSUE #1230: Drop `warning` from restore item action plugin interface
xref #1123 
xref #286 

Restore item actions can return a warning, in the form of an `error`. AFAIK this isn't actually used anywhere, and I think we may want to get away from it in general.  Investigate whether it's okay to remove, and if so, do it as part of the plugins work.  

@carlisia

skriss:

Fixes #1230


skriss:
fixed conflicts

Issue #857, opened 2018-09-20
Update go-plugin magic cookie key and/or protocol version


nrb:
v0.10.0 has changed the plugin API. In order to ensure a v0.9.x server doesn't try to run later plugins, we should update at least the [protocol version](https://github.com/heptio/ark/blob/master/pkg/plugin/server.go#L31) so that they will fail fast.

This may also require code changes to our examples and updating plugin docs, but I haven't dug in depth on that yet.

nrb:
We'll hold off on this until we have more substantive changes to the API (like adding new methods or changing any signatures)

carlisia:
Question for the @heptio/ark-team: if we bump the protocol version, none of the existing external plugins will work until they update to the new interface (and maybe something else to indicate a correspondence with the new version, but haven't looked at what that would be yet.)

The way to both bump the protocol version AND maintain backwards compatibility is to have two client implementations, one for each version.

I'm not sure how important backwards compatibility is for plugins, or how important it is to add this gatekeeping.

skriss:
We're not attempting to maintain backwards-compatibility for this set of changes. Once we're at 1.0, that'll change, but fine to break here.

The internal plugins will work fine with this change - since they're part of the same binary.  Any externally-developed plugins will need to be recompiled with the latest code to work.

carlisia:
If I change both the`ProtocolVersion` and the `MagicCookeyKey` (from `ARK_PLUGIN` to `VELERO_PLUGIN` we'll get this message:

![image](https://user-images.githubusercontent.com/16508/55101416-4cbd7a00-5081-11e9-9428-803bbad43447.png)

If I only change the `ProtocolVersion` we get this message:

![image](https://user-images.githubusercontent.com/16508/55101446-5e9f1d00-5081-11e9-9acc-ee4a3c7cac26.png)

Should I change both?

PR #1319, opened 2019-03-27
Bump plugin client version

carlisia:

Closes: #857

Please see this comment: https://github.com/heptio/velero/issues/857#issuecomment-477288063

### DIFF
--- a/pkg/plugin/framework/handshake.go
+++ b/pkg/plugin/framework/handshake.go
@@ -1,0 +1,32 @@
+/*
+Copyright 2017, 2019 the Velero contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


skriss:
I think changing both makes sense -- that error seems more descriptive and good to replace ARK anyway.

Mind addressing the TODO about turning it into a function while you're in there? I think it should be super-straightforward. If not, don't waste too much time on it.

pkg/plugin/framework/handshake.go
```diff
@@ -0,0 +1,32 @@
+/*
+Copyright 2017, 2019 the Velero contributors.
```
skriss: just 2019

pkg/plugin/framework/handshake.go
```diff
@@ -0,0 +1,32 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import plugin "github.com/hashicorp/go-plugin"
+
+// Handshake returns the configuration information that allows go-plugin clients and servers to perform a handshake.
+func Handshake() plugin.HandshakeConfig {
```
skriss: gotta update callers to use this as a function instead of a field

skriss:
LGTM - add a changelog and we're all set

PR #1322, opened 2019-03-27
don't wrap io.EOF errors during gRPC streaming

skriss:

Fixes the error we've been seeing.


carlisia:
I don't feel left out now, I got it too:

image

carlisia:
I tested this branch and was able to do a backup w/o the error.

PR #1325, opened 2019-03-28
turn down log levels in plugin server to DEBUG

skriss:

These should've been debug to begin with, they're pretty noisy.


skriss:
merge away

PR #1326, opened 2019-03-28
Added ability to dynamically disable controllers

amanw:
There are scenarios where we would like to use a different scheduler for dealing with Scheduling and GC of backups. Therefore we would need to disable the Schedule & GC Controllers for this capability.

This change is by default disabled.

Signed-off-by: James King <james.king@emc.com>

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -84,9 +84,34 @@ const (
 
 	defaultProfilerAddress = "localhost:6060"
 
+	// keys used to map out available controllers with disable-controllers flag
+	BackupControllerKey              = "backup"
+	BackupSyncControllerKey          = "backup-sync"
+	ScheduleControllerKey            = "schedule"
+	GcControllerKey                  = "gc"
+	BackupDeletionControllerKey      = "backup-deletion"
+	RestoreControllerKey             = "restore"
+	DownloadRequestControllerKey     = "download-request"
+	ResticRepoControllerKey          = "restic-repo"
+	ServerStatusRequestControllerKey = "server-status-request"
+
+	defaultControllerWorkers = 1
 	// the default TTL for a backup
 	defaultBackupTTL = 30 * 24 * time.Hour
 )
+
+// list of available controllers for input validation
+var disableControllerList = []string{
+	BackupControllerKey,
+	BackupSyncControllerKey,
+	ScheduleControllerKey,
+	GcControllerKey,
+	BackupDeletionControllerKey,
+	RestoreControllerKey,
+	DownloadRequestControllerKey,
+	ResticRepoControllerKey,
+	ServerStatusRequestControllerKey,
+}
 
 type serverConfig struct {
 	pluginDir, metricsAddress, defaultBackupLocation                        string
@@ -95,9 +120,15 @@ type serverConfig struct {
 	restoreResourcePriorities                                               []string
 	defaultVolumeSnapshotLocations                                          map[string]string
 	restoreOnly                                                             bool
+	disabledControllers                                                     []string
 	clientQPS                                                               float32
 	clientBurst                                                             int
 	profilerAddress                                                         string
+}
+
+type controllerRunInfo struct {
+	controller controller.Interface
+	numWorkers int
 }
 
 func NewCommand() *cobra.Command {
@@ -169,6 +200,7 @@ func NewCommand() *cobra.Command {
 	command.Flags().DurationVar(&config.backupSyncPeriod, "backup-sync-period", config.backupSyncPeriod, "how often to ensure all Velero backups in object storage exist as Backup API objects in the cluster")
 	command.Flags().DurationVar(&config.podVolumeOperationTimeout, "restic-timeout", config.podVolumeOperationTimeout, "how long backups/restores of pod volumes should be allowed to run before timing out")
 	command.Flags().BoolVar(&config.restoreOnly, "restore-only", config.restoreOnly, "run in a mode where only restores are allowed; backups, schedules, and garbage-collection are all disabled")
+	command.Flags().StringSliceVar(&config.disabledControllers, "disable-controllers", config.disabledControllers, fmt.Sprintf("list of controllers to disable on startup. Valid values are %s", strings.Join(disableControllerList, ",")))
 	command.Flags().StringSliceVar(&config.restoreResourcePriorities, "restore-resource-priorities", config.restoreResourcePriorities, "desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources")
 	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "name of the default backup storage location")
 	command.Flags().Var(&volumeSnapshotLocations, "default-volume-snapshot-locations", "list of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...)")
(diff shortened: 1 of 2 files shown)


nrb:
>There are scenarios where we would like to use a different scheduler for dealing with Scheduling and GC of backups.

Can you provide some more details here? What sorts of scenarios?

skriss:
I think I'd prefer an approach here where the flag is `--disable-controllers` and the value can be a list of controller names - that way, we don't have to keep creating new flags for every combo of controllers that folks want to disable. WDYT?

amanw:
@nrb mainly around where we want to integrate the Velero Kubernetes functionality into existing backup tech that manages our other stacks, then we can centralize the lifecycle of backups across disparate platforms (VMWare, Kubernetes, AWS, etc) but not have to rewrite an entire backup capability for Kubernetes.

@skriss I agree on that, we can definitely change it up and put the controller startup blocks into a map/slice that just deletes the desired references and then runs through their startup accordingly.

skriss:
@amanw yep - that sounds good to me!

pkg/cmd/server/server.go
```diff
@@ -83,14 +83,39 @@ const (
 	defaultClientBurst int     = 30
 
 	defaultProfilerAddress = "localhost:6060"
+
+	// keys used to map out available controllers with disable-controllers flag
+	BackupControllerKey              = "backupcontroller"
+	BackupSyncControlleKey           = "backupsynccontroller"
```
nrb: Missing an `r` here :)

pkg/cmd/server/server.go
```diff
@@ -622,80 +654,122 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
+			GcControllerKey,
+			BackupDeletionControllerKey,
+		)
+	}
+
+	// remove disabled controllers
+	for _, controllerName := range s.config.disabledControllers {
+		_, ok := enabledControllers[controllerName]
+		if ok {
+			delete(enabledControllers, controllerName)
```
nrb: I think an info log message is warranted as we disable controllers; it would ease troubleshooting in case someone accidentally disabled something.

pkg/cmd/server/server.go
```diff
@@ -622,80 +654,122 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
+		}
+	}
+
+	for _, startupFunction := range enabledControllers {
+		startupFunction()
```
skriss: it would be nice to extract the common stuff from each startup func:

```
wg.Add(1)
go func() {
    <controller>.Run(ctx, 1)
    wg.Done()
}()
```
You could have each "startup func" actually just instantiate & return the controller, so the type would be something like `type newController func() controller.Interface`, and then this loop would change to something like:

```go
for _, newController := range enabledControllers {
    controller := newController()
    wg.Add(1)
    go func() {
        controller.Run(ctx, 1)
        wg.Done()
}()
}
``` 

Think this is a worthwhile change?

pkg/cmd/server/server.go
amanw: Will fix. Good catch.

pkg/cmd/server/server.go
```diff
@@ -83,14 +83,39 @@ const (
 	defaultClientBurst int     = 30
 
 	defaultProfilerAddress = "localhost:6060"
+
+	// keys used to map out available controllers with disable-controllers flag
+	BackupControllerKey              = "backupcontroller"
```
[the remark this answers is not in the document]
amanw: Good point, we will remove the "controller" repetition.

pkg/cmd/server/server.go
```diff
@@ -622,80 +654,122 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
+			BackupControllerKey,
+			ScheduleControllerKey,
+			GcControllerKey,
+			BackupDeletionControllerKey,
+		)
+	}
+
+	// remove disabled controllers
+	for _, controllerName := range s.config.disabledControllers {
+		_, ok := enabledControllers[controllerName]
```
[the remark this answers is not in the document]
amanw: We will collapse this.

pkg/cmd/server/server.go
```diff
@@ -622,80 +654,122 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
+		if ok {
+			delete(enabledControllers, controllerName)
+		}
```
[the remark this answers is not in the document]
amanw: We will Fataf out on an invalid flag.

pkg/cmd/server/server.go
amanw: We talked about this type of change but ultimately didn't know how we would want to handle the variable worker allocation. That "1" (magic number) could be varied in each controller so it would be more along the lines of:

```golang
serverStatusRequestControllerGetter := func() (*statusRequestController, int) {
	serverStatusRequestController := controller.NewServerStatusRequestController(
		s.logger,
		s.veleroClient.VeleroV1(),
		s.sharedInformerFactory.Velero().V1().ServerStatusRequests(),
	)
	return serverStatusRequestController, 1 // 1 is for the numWorkers
}
```

``` golang
for _, newController := range enabledControllers {
	controller, numWorkers := newController()
	wg.Add(1)
	go func() {			
            controller.Run(ctx, numWorkers)
	    wg.Done()		
        }()
}
```

Thoughts? We can make the changes pretty trivially.

amanw: ...and we pushed a commit with a proposed example....

pkg/cmd/server/server.go
```diff
@@ -524,27 +550,28 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
-		newPluginManager,
-		s.logger,
-	)
-	wg.Add(1)
-	go func() {
-		backupSyncController.Run(ctx, 1)
-		wg.Done()
-	}()
+	backupSyncControllerGetter := func() (controller.Interface, int) {
+		numControllerWorkers := 1
```
skriss: I think this works. It might be a little bit clearer if we defined a simple struct along the lines of:
```go
type controllerRunInfo struct {
    controller controller.Interface
    numWorkers int
}
```
and return that from the anonymous functions - just slightly more self-documenting. I would also define a constant , `defaultControllerWorkers`, and just use that in all of these functions rather than declaring a new variable in each one. Can be overridden in individual functions later if needed, but right now it's not needed :)

pkg/cmd/server/server.go
```diff
@@ -615,87 +645,126 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
+		)
+	}
+
+	// remove disabled controllers
+	for _, controllerName := range s.config.disabledControllers {
+		if _, ok := enabledControllers[controllerName]; ok {
+			s.logger.Infof("Disabling controller: %s", controllerName)
+			delete(enabledControllers, controllerName)
+		} else {
+			s.logger.Fatalf("Invalid Controller flag provided: %s", controllerName)
```
skriss: "Invalid value for --disable-controllers flag provided: ..."

skriss:
Thanks for the update - this LGTM!

pkg/cmd/server/server.go
```diff
@@ -615,87 +654,134 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
+	}
+
+	// remove disabled controllers
+	for _, controllerName := range s.config.disabledControllers {
+		if _, ok := enabledControllers[controllerName]; ok {
+			s.logger.Infof("Disabling controller: %s", controllerName)
+			delete(enabledControllers, controllerName)
+		} else {
+			s.logger.Fatalf("Invalid value for --disable-controllers flag provided: %s", controllerName)
+		}
```
nrb: Could we enumerate the valid values in the error? I know it's in the help message, but this removes a little friction for users.

skriss:
Looks like there's a small merge conflict now as a result of #1352 going in yesterday, but if you can fix + address nrb's comment above, we should be good to merge today.

pkg/cmd/server/server.go
king-jam: Added!

skriss:
LGTM

nrb:
LGTM! @amanw @king-jam I don't want to lose your attributions on squashing, could one of you squash this into 1 commit and make sure the updated commit message is generic?

king-jam:
Finished the `git squash` games and added a `changelog` file in as well. Should be all set once Travis finishes.

PR #1329, opened 2019-03-28
Fixes logger configuration for restore plugins

### LINKED ISSUE #1328: Logs from restore plugins are getting lost
Running a restore with a custom plugin registered, log messages from the restore Execute action are not being logged to the restore log. For comparison, backup logs from plugin Execute actions are showing up in the velero logs as expected.

I expected that 'velero restore logs <my-restore>' would show the log messages from my plugin.

I tracked this down to controller/restore_controller.go. For the backup plugin, this all works properly. First the logger is re-initialized, setting logger.out to an io.MultiWriter: https://github.com/heptio/velero/blob/master/pkg/controller/backup_controller.go#L407-L409
*then* the pluginManager is initialized with newPluginManager, passing in the new logger: https://github.com/heptio/velero/blob/master/pkg/controller/backup_controller.go#L419
All of this happens within runBackup, which is called by processBackup.

For restore, however, things are more complicated, but the root of the problem is that the pluginManager is initialized *before* the logger is re-initialized, so anything written to the log configured for the pluginManager gets lost. For restore, newPluginManager is called in processRestore, and this manager is passed into the call to runRestore, which then re-initializes logs before running the restore.

PluginManager is initialized here: https://github.com/heptio/velero/blob/master/pkg/controller/restore_controller.go#L207
passed into runRestore here: https://github.com/heptio/velero/blob/master/pkg/controller/restore_controller.go#L244-L249
Logger is re-initialized here: https://github.com/heptio/velero/blob/master/pkg/controller/restore_controller.go#L464-L470

I have a fix in our velero fork which seems to work, but it's somewhat messy. I essentially had to merge the logic of processRestore and runRestore in order to get the pluginManager initialization to work properly without changing anything else, but it's at the expense of readability. I'll submit this as a PR here -- if this fix isn't appropriate for an upstream commit, it will at least point you in the right direction as to what's wrong and what needs to change.

**Environment:**

- Velero version (use `velero version`): master branch
- Cloud provider or hardware configuration: Running in an Openshift 4.0 environment in AWS

sseago:
Fixes #1328 

In restoreController.runRestore, the logger is replaced with a new
io.MultiWriter-based logger. Currently this is done after initializing
the pluginManager, which means the restore plugin logs are being lost.

For backup, newPluginManager is called after the MultiWriter setup,
so log messages from plugins are written to the log as expected.

Untangling this is complicated, because the pluginManager is used for
several actions in restoreController.processRestore before calling
runRestore.

This commit combines restoreController.processRestore and
restoreController.runRestore into one function, with logger reconfiguration
moved above the call to newPluginManager. This is further complicated
by the different return situations in the original two functions
and an attempt to make the result of calling processRestore as close
as possible to what was happening before. A more elegant solution
is probably possible with a more thorough redesign of these functions.

### DIFF
--- a/changelogs/unreleased/1329-sseago
+++ b/changelogs/unreleased/1329-sseago
@@ -1,0 +1,1 @@
+Fixes logger configuration for restore plugins
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -41,11 +41,11 @@ import (
 	"github.com/heptio/velero/pkg/metrics"
 	"github.com/heptio/velero/pkg/persistence"
 	"github.com/heptio/velero/pkg/plugin/clientmgmt"
-	"github.com/heptio/velero/pkg/plugin/velero"
 	"github.com/heptio/velero/pkg/restore"
 	"github.com/heptio/velero/pkg/util/collections"
 	kubeutil "github.com/heptio/velero/pkg/util/kube"
 	"github.com/heptio/velero/pkg/util/logging"
+	"github.com/heptio/velero/pkg/volume"
 )
 
 // nonRestorableResources is a blacklist for the restoration process. Any resources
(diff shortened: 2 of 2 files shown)


sseago:
Addresses https://github.com/heptio/velero/issues/1328

nrb:
Thanks @sseago! We'll give this a look over the next few days. The restore package is definitely in need of some refactoring attention, that's for sure.

skriss:
@sseago thanks for this - I've been looking at making some changes to this same section of code that may overlap with yours.  I'm going to keep this PR open for now as I work through that one, and can figure out what makes the most sense here.  Thanks for spotting this issue!

skriss:
we ended up merging #1358 to fix this issue; thanks again @sseago.

PR #1332, opened 2019-03-29
Export plugin functionality

carlisia:

Turns out plugin authors need these. Ex:

```
func newBackupPlugin(options ...veleroplugin.PluginOption) *veleroplugin.BackupItemActionPlugin {
	return &veleroplugin.BackupItemActionPlugin{
		PluginBase: veleroplugin.NewPluginBase(options...),
	}
}
```

### DIFF
--- a/pkg/plugin/framework/backup_item_action.go
+++ b/pkg/plugin/framework/backup_item_action.go
@@ -29,7 +29,7 @@ import (
 // interface.
 type BackupItemActionPlugin struct {
 	plugin.NetRPCUnsupportedPlugin
-	*pluginBase
+	*PluginBase
 }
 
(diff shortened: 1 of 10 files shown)


skriss:
What is this needed for? I'd expect all they need to serve a plugin is the following - https://github.com/heptio/velero-plugin-example/blob/master/velero-examples/main.go#L25-L30

carlisia:
I thought we wanted to change and pass a concrete `*veleroplugin.BackupItemActionPlugin` into `RegisterBackupItemAction` but I guess not :)

skriss:
the functions that are passed to `RegisterXXX` should return implementations of the `pkg/plugin/velero` interfaces.  Means the plugin author doesn't need to know much about the plugin framework apart from that one block of code I linked to.

PR #1334, opened 2019-03-29
Added example for NooBaa as S3 backend

cwilkers:
This PR adds example yaml for using an S3 volume as the Velero backupstoragelocation.


nrb:
Thanks for the PR!

The example directory is going to get removed soon as part of the `install` command work, so I don't think we'll want to merge this PR. However, since NooBaa appears to use the S3 interface, it could be added into the [support matrix](https://github.com/heptio/velero/blob/master/docs/support-matrix.md#s3-compatible-backup-storage-providers) file.

PR #1336, opened 2019-03-31
Adds support for allowing a RestoreItemAction to skip item restore

### LINKED ISSUE #1335: Add support for allowing a RestoreItemAction plugin to skip item restore
It would be useful if the Execute action for a RestoreItemAction plugin could signal to velero to *not* continue with the object restore. Doing so would cause velero to stop processing the current item and move onto the next, just as it does for an error condition, except without logging an error state.

**My current use case that calls for this**
I'm working on backup and restore plugins for OpenShift imagestream resources. For imagestreams that represent images in the internal docker registry, the backup plugin copies all images referenced in the 'status' field to a standalone backup/restore docker registry, in addition to velero creating a backup of the imagestream resource in the normal velero backup tarball. The restore plugin will then iterate over the images referenced in ItemFromBackup and copies the associated docker images from the backup/restore registry into the cluster's internal docker registry.

(here's where the RFE comes into play)
Since importing a docker image into the openshift internal registry automatically creates the appropriate imagestream and imagestreamtag resources, I don't want velero to run the `resourceClient.Create` call. The plugin's custom operations already do everything needed to restore the object, so the correct behavior at this point is to discard the Item and move on.

Another use case is where restore is conditionalized based on some data within the item -- i.e. for some reason you want to restore some, but not all, pods. For example, maybe you want to restore standalone pods but discard pods associated with deployments, since those will be created automatically in the restored cluster.

I have a working PR for this that I will submit shortly. 

My initial idea was to implement this by simply looking for a nil return value, but none of my attempts to do that were successful. I suspect it's a combination of something in the current codebase not handling nil values properly, and various places in go where we're getting a struct with zero value elements rather than a nil, especially going through the client/server path from plugin to server. Also, it's more confusing to rely on "magic" values like that rather than having an explicit field, which leads to my PR. I've added a `SkipRestore` bool to `RestoreItemActionExecuteOutput`, so that a plugin which wants to stop processing and move on can simply `return velero.NewRestoreItemActionExecuteOutput(input.Item).WithoutRestore(), nil`

Also, I may not be generating the autogenerated code properly. When I call `hack/generate-proto.sh` in my checkout, the resulting code seems to work fine, but it's making more changes than simply adding the new field to the various types and functions that need it. I'll isolate those changes in a separate commit within the PR so it will be easy enough to swap them out with something different if needed.

sseago:
Fixes #1335

Adds support for allowing a RestoreItemAction to skip item restore

This allows a RestoreItemAction plugin to signal to velero that
the returned item should be skipped rather than restored to the
cluster.

To support this, a boolean SkipRestore attribute is added to
RestoreItemActionExecuteOutput. If restore.restoreResource finds
this set to true, any remaining actions on this item are skipped,
and restore on this item is skipped. Execution continues with
the next item of this resource type.

To signal this for a particular item, the RestoreItemAction's
Execute method should call WithoutRestore() on the
RestoreItemActionExecuteOutput before returning it.

The autogenerated code may need to be redone. I don't know if I'm using the wrong protoc version or missing a flag of some sort, because it changed a lot more than just around the code changes for this PR.


sseago:
Addresses https://github.com/heptio/velero/issues/1335

sseago:
In our velero fork repo, instead of the full "Autogenerated code to support SkipRestore" commit, I applied a much smaller commit, basically just taking the subset of protoc-generated output that was relevant to my code changes.  Basically, instead of a bunch of changes across several files, it only modifies one file, adding SkipRestore to `type RestoreItemActionExecuteResponse`, and adding `func (m *RestoreItemActionExecuteResponse) GetSkipRestore() bool` with all the other changes from `hack/generate-proto.sh`. So far everything seems to work with this more limited commit, but I wasn't sure which approach was recommended

pkg/plugin/framework/restore_item_action_server.go
```diff
@@ -116,17 +116,20 @@ func (s *RestoreItemActionGRPCServer) Execute(ctx context.Context, req *proto.Re
 	// If the plugin implementation returned a nil updateItem (meaning no modifications), reset updatedItem to the
 	// original item.
 	var updatedItemJSON []byte
+	skipRestore := false
```
skriss: I think you don't need this line or L124, see next comment

pkg/plugin/framework/restore_item_action_server.go
```diff
@@ -116,17 +116,20 @@ func (s *RestoreItemActionGRPCServer) Execute(ctx context.Context, req *proto.Re
...
+		skipRestore = executeOutput.SkipRestore
 		if err != nil {
 			return nil, newGRPCError(errors.WithStack(err))
 		}
 	}
 
 	res := &proto.RestoreItemActionExecuteResponse{
-		Item: updatedItemJSON,
+		Item:        updatedItemJSON,
+		SkipRestore: skipRestore,
```
skriss: I think you want this to be `SkipRestore: executeOutput.SkipRestore`

pkg/plugin/velero/restore_item_action.go
```diff
@@ -57,6 +57,10 @@ type RestoreItemActionExecuteOutput struct {
 	// AdditionalItems is a list of additional related items that should
 	// be restored.
 	AdditionalItems []ResourceIdentifier
+
+	// SkipRestore tells velero to stop executing further actions
```
skriss: I would note in this comment that if this field's value is `true`, then `AdditionalItems` will be ignored.

skriss:
Output of `protoc --version` for me:
> `libprotoc 3.6.1`

skriss:
I did a quick proto update with the field you added - you can grab https://github.com/skriss/velero/commit/2293e0012e750201d13a806d29ad3f4004001f60 if that's easier

pkg/plugin/framework/restore_item_action_server.go
sseago: Good catch. I think at one point I was concerned about cases where executeOutput might be nil, but this code doesn't actually check for that anyway. The only case where executeOutput will be nil is when err != nil, which is already handled anyway. Also, the way this is written, if UpdatedItem is nil, then skipRestore won't be set correctly -- the right thing to do is always set it in L132 below.

pkg/plugin/framework/restore_item_action_server.go
sseago: Yes, this is right. I'll change this.

sseago:
@skriss Looks like grabbing protoc-gen-go v1.0.0 was all I needed. My proto output now matches the one you linked above.

sseago:
I've made the suggested changes and fixed the generated code output.

PR #1338, opened 2019-03-31
objectstores/volumesnapshotters: check for invalid keys in config

### LINKED ISSUE #489: Error if unknown keys are included in configs
If we have an Ark `Config` with invalid persistentVolumeProvider or backupStorageProvider `config` fields, we should error instead of silently allowing (and ignoring) them. We had a user mistype `s3Url` as `s3url` and it took a long time to determine the root cause of the issue.

skriss:

Fixes #489

### DIFF
--- a/pkg/cloudprovider/config.go
+++ b/pkg/cloudprovider/config.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2019 the Velero contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


carlisia:
Changelog needed?

skriss:
yep - added

pkg/cloudprovider/config.go
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+package cloudprovider
+
+import (
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+func ValidateConfigKeys(config map[string]string, validKeys ...string) error {
```
nrb: This function is fairly simple, but maybe it makes sense to expose in the plugin `framework` package, so 3rd party plugins can also do validation. What do you think?

skriss: Hmm. I think I'd rather have them vendor `pkg/cloudprovider` if they want to use it, since it's not related to the gRPC/go-plugin infrastructure for plugins.

nrb: Fair enough.

PR #1339, opened 2019-04-01
Add validation for plugin name format and dups

carlisia:
Fixes #https://github.com/heptio/velero/issues/583

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -1,14 +1,14 @@
 # Plugins
 
-Velero has a plugin architecture that allows users to add their own custom functionality to Velero backups & restores 
-without having to modify/recompile the core Velero binary. To add custom functionality, users simply create their own binary 
-containing implementations of Velero's plugin kinds (described below), plus a small amount of boilerplate code to 
-expose the plugin implementations to Velero. This binary is added to a container image that serves as an init container for 
-the Velero server pod and copies the binary into a shared emptyDir volume for the Velero server to access. 
+Velero has a plugin architecture that allows users to add their own custom functionality to Velero backups & restores without having to modify/recompile the core Velero binary. To add custom functionality, users simply create their own binary containing implementations of Velero's plugin kinds (described below), plus a small amount of boilerplate code to expose the plugin implementations to Velero. This binary is added to a container image that serves as an init container for the Velero server pod and copies the binary into a shared emptyDir volume for the Velero server to access.
 
 Multiple plugins, of any type,  can be implemented in this binary.
 
-A fully-functional [sample plugin repository][1] is provided to serve as a convenient starting point for plugin authors.
+A fully-functional [sample plugin repository][1] is provided to serve as a convenient starting point for plugin authors. Note that when naming your plugin, the name needs to conform to these rules:
+- have two parts separated by '/'
+- non of the above parts can be empty
+- the prefix is a valid DNS subdomain name
+- a plugin with the same name cannot not already exist
 
 ## Plugin Kinds
 
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -4,9 +4,13 @@ Velero has a plugin architecture that allows users to add their own custom funct
 
 Multiple plugins, of any type,  can be implemented in this binary.
 
-A fully-functional [sample plugin repository][1] is provided to serve as a convenient starting point for plugin authors. Note that when naming your plugin, the name needs to conform to these rules:
+A fully-functional [sample plugin repository][1] is provided to serve as a convenient starting point for plugin authors.
+
+## Plugin Naming
+
+When naming your plugin, keep in mind that the name needs to conform to these rules:
 - have two parts separated by '/'
-- non of the above parts can be empty
+- none of the above parts can be empty
 - the prefix is a valid DNS subdomain name
 - a plugin with the same name cannot not already exist
 
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -14,6 +14,15 @@ When naming your plugin, keep in mind that the name needs to conform to these ru
 - the prefix is a valid DNS subdomain name
 - a plugin with the same name cannot not already exist
 
+### Some examples:
+```
+- example.io/azure
+- 1.2.3.4/5678
+- example-with-dash.io/azure
+```
+
+You will need to give your plugin(s) a name when registering them by calling the appropriate `RegisterX` function: <https://github.com/heptio/velero/blob/0e0f357cef7cf15d4c1d291d3caafff2eeb69c1e/pkg/plugin/framework/server.go#L42-L60>
+
 ## Plugin Kinds
 
 Velero currently supports the following kinds of plugins:
--- a/pkg/plugin/framework/examples_test.go
+++ b/pkg/plugin/framework/examples_test.go
@@ -23,8 +23,8 @@ import (
 
 func ExampleNewServer_volumeSnapshotter() {
 	NewServer(). // call the server
-			RegisterVolumeSnapshotter("example/volumesnapshotter", newVolumeSnapshotter). // register the plugin
-			Serve()                                                                       // serve the plugin
+			RegisterVolumeSnapshotter("example.io/volumesnapshotter", newVolumeSnapshotter). // register the plugin with a valid name
+			Serve()                                                                          // serve the plugin
 }
 
 func newVolumeSnapshotter(logger logrus.FieldLogger) (interface{}, error) {
--- a/pkg/plugin/framework/examples_test.go
+++ b/pkg/plugin/framework/examples_test.go
@@ -23,7 +23,7 @@ import (
 
 func ExampleNewServer_volumeSnapshotter() {
 	NewServer(). // call the server
-			RegisterVolumeSnapshotter("example-volumesnapshotter", newVolumeSnapshotter). // register the plugin
+			RegisterVolumeSnapshotter("example/volumesnapshotter", newVolumeSnapshotter). // register the plugin
 			Serve()                                                                       // serve the plugin
 }
 
--- a/pkg/plugin/framework/server_mux.go
+++ b/pkg/plugin/framework/server_mux.go
@@ -70,7 +70,7 @@ func (m *serverMux) getHandler(name string) (interface{}, error) {
 
 	initializer, found := m.initializers[name]
 	if !found {
-		return nil, errors.Errorf("%v plugin: %s was not found or has an invalid format", m.kind, name)
+		return nil, errors.Errorf("%v plugin: %s was not found or has an invalid name format", m.kind, name)
 	}
 
 	instance, err := initializer(m.serverLog)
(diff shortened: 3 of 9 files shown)


pkg/plugin/framework/server_mux.go
```diff
@@ -43,9 +45,20 @@ func newServerMux(logger logrus.FieldLogger) *serverMux {
 	}
 }
 
-// register registers the initializer for name.
+// register registers the initializer for the given name. Proper format
+// for the name is <namespace>/<name> and there can be no duplicates.
 func (m *serverMux) register(name string, f HandlerInitializer) {
-	// TODO(ncdc): return an error on duplicate registrations for the same name.
+	name = strings.ToLower(name)
+	if !strings.Contains(name, "/") {
```
nrb: While this works, `/` is a valid name here. :)

pkg/plugin/framework/server_mux.go
```diff
@@ -63,7 +76,7 @@ func (m *serverMux) getHandler(name string) (interface{}, error) {
 
 	initializer, found := m.initializers[name]
 	if !found {
-		return nil, errors.Errorf("unknown %v plugin: %s", m.kind, name)
+		return nil, errors.Errorf("%v plugin: %s not found or had invalid format. Valid format: <namespace>/<name>", m.kind, name)
```
nrb: I don't think the first component is a namespace (`velero.io` isn't). I think it's mean to be a domain name, if you're following annotation/label key conventions.

pkg/plugin/framework/server_mux.go
carlisia: Argh true. Thanks.

pkg/plugin/framework/server_mux.go
carlisia: I'm trying to indicate that the second component might have an invalid format. Otherwise I am misinterpreting what you said, please say it in a different way?

carlisia: I cleaned up the message a bit.

nrb: I read `<namespace>/<name>` as `namespace = velero.io` and `name = job`. In `velero.io/job`, I wouldn't necessarily called `velero.io` a namespace, mostly because that's a specific thing in Kubernetes.

Was the intent to make the first part a vendor identifier?

carlisia: So `namespace` is a concept I just added. And yes, the idea is to be a vendor identifier. Maybe the issue will give better context: https://github.com/heptio/velero/issues/583

nrb: Yeah, the issues helps, thanks :)

There's a similar concept for annotations, labels, and finalizers. (like `velero.io/schedule-name`). As an example, for finalizers, they expect the names to be fully qualified domain names.

This is the error I get if I add one that's not an FQDN: ` metadata.finalizers[0]: Invalid value: "velero": name is neither a standard finalizer name nor is it fully qualified`. Unfortunately, [the docs](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers) don't specify that restriction :(

skriss: FYI - here's what I found re: labels -- https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

pkg/plugin/framework/server_mux.go
```diff
@@ -75,3 +80,18 @@ func (m *serverMux) getHandler(name string) (interface{}, error) {
 
 	return m.handlers[name], nil
 }
+
+func validPluginName(name string, existingNames []string) bool {
+	name = strings.ToLower(name)
+	tokens := strings.Split(name, "/")
+	if len(tokens) <= 1 || tokens[0] == "" || tokens[1] == "" {
```
nrb: I think `velero.io/my-plugin/third-field` would pass here. How strict do we want to be? Do we want to have it be exactly 1 `/` and exactly 2 strings?

carlisia: I think the requirement should be `at least one`. @skriss wdyt?

pkg/plugin/framework/server_mux.go
carlisia: I think then we should call `velero.io` a prefix and validate it to be a FQDN. We shouldn't call it a lable because `labels do not provide uniqueness` (as per that ^ doc). But we can say it's a key/value pair of prefix/name. Yeay/nay?

nrb: Yeah, it's not actually a label, but it's a part of a string in a very similar vein. I like "prefix" or "vendor prefix," to be a little more specific. Vendors are then expected to use their own for their plugins.

pkg/plugin/framework/server_mux.go
skriss: Yeah, I think we want to limit to exactly one slash. Ideally we validate that the first part is an FQDN as well (I *think* that spec is valid characters of a-z, 0-9, `.`, `-`, max total length 253, max length between `.`'s 63).

pkg/plugin/framework/server_mux.go
```diff
@@ -43,9 +45,12 @@ func newServerMux(logger logrus.FieldLogger) *serverMux {
 	}
 }
 
-// register registers the initializer for name.
+// register registers the initializer for the given name. Proper format
+// for the name is <namespace>/<name> and there can be no duplicates.
 func (m *serverMux) register(name string, f HandlerInitializer) {
-	// TODO(ncdc): return an error on duplicate registrations for the same name.
+	if !validPluginName(name, m.names()) {
```
skriss: I think it makes sense to have this check here, on the server side; additionally, I think we need a similar check in `clientmgmt/registry.go`, on the client side -- probably in this function https://github.com/heptio/velero/blob/master/pkg/plugin/clientmgmt/registry.go#L199

carlisia:
I wanted to circle back to this and make a proposal that is similar to what we've talked about. I started doing this from scratch then decided to look at the kubernetes code and see how they handle it. Here is the test case: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation_test.go#L247

It is very much what we talked about, except that it doesn't require a `/`. I think we could use (call) that function as is and so also not require a `/`. Users can "namespace" the plugin name by using `dots`.

That function return a slice of errors when it doesn't pass, and we could log those errors since the `register` function doesn't return anything.

nrb:
Could we use `IsQualifiedName` in conjuction with a `/`? My thought is we validate the first chunk with `IsQualifiedName`, having a single slash, and a name.

carlisia:
I thought about that too, doing a last pass to validate whatever exception to fit our needs. But, if we do that at least 1 error messages from within that function won't make sense. 

However, we can "re-use" the code and shape it to fit what we want. The tradeoff of doing this of course is that we are no longer completely compatible with the k8s format.

nrb:
> The tradeoff of doing this of course is that we are no longer completely compatible with the k8s format.

That is true, certainly.  We don't have to stay completely compatible, but it also means we have less to be responsible for.

skriss:
I took a quick look at this - I *believe* the `IsDNS1123Subdomain` function is exactly what we want for the prefix. So we can validate that there's exactly one slash, that the first part is valid according to `IsDNS1123Subdomain`, and that the second part is non-empty.  Sound right?

skriss:
Also, we're already vendoring this apimachinery package, so doesn't require adding a new dependency to use.

pkg/plugin/framework/examples_test.go
```diff
@@ -23,7 +23,7 @@ import (
 
 func ExampleNewServer_volumeSnapshotter() {
 	NewServer(). // call the server
-			RegisterVolumeSnapshotter("example-volumesnapshotter", newVolumeSnapshotter). // register the plugin
+			RegisterVolumeSnapshotter("example/volumesnapshotter", newVolumeSnapshotter). // register the plugin
```
skriss: let's make this `example.io/volumesnapshotter` for consistency.

pkg/plugin/framework/server_mux.go
```diff
@@ -63,7 +70,7 @@ func (m *serverMux) getHandler(name string) (interface{}, error) {
 
 	initializer, found := m.initializers[name]
 	if !found {
-		return nil, errors.Errorf("unknown %v plugin: %s", m.kind, name)
+		return nil, errors.Errorf("%v plugin: %s was not found or has an invalid format", m.kind, name)
```
skriss: "invalid name format"

pkg/plugin/framework/server_mux.go
```diff
@@ -75,3 +82,34 @@ func (m *serverMux) getHandler(name string) (interface{}, error) {
...
+// ValidatePluginName checks if the given name:
+// - the plugin name has two parts separated by '/'
+// - non of the above parts is empty
+// - the prefix is a valid DNS subdomain name,
+// - a plugin with the same name does not already exist (if list of existing names is passed in)
+func ValidatePluginName(name string, existingNames []string) error {
+	// validate there is one "/" and two parts
+	parts := strings.Split(name, "/")
+	if len(parts) != 2 {
+		return errors.Errorf("%s is not a DNS subdomain prefix followed by '/' and a non-empty name (e.g. 'example.com/name'", name)
```
skriss: missing closing parenthesis in the error msg - should be `return errors.Errorf("%s is not a DNS subdomain prefix followed by '/' and a non-empty name (e.g. 'example.com/name')", name)`

pkg/plugin/framework/server_mux.go
```diff
@@ -75,3 +82,34 @@ func (m *serverMux) getHandler(name string) (interface{}, error) {
 
 	return m.handlers[name], nil
 }
+
+// ValidatePluginName checks if the given name:
+// - the plugin name has two parts separated by '/'
+// - non of the above parts is empty
+// - the prefix is a valid DNS subdomain name,
+// - a plugin with the same name does not already exist (if list of existing names is passed in)
+func ValidatePluginName(name string, existingNames []string) error {
```
skriss: it'd be nice to have all of the validation errors that could be returned from this function have consistent phrasing about what a valid name is (`<DNS subdomain>/<non-empty name>`), and which part of that was invalid.

skriss: Maybe something like the following:

```
- plugin name must have exactly two parts separated by a `/`. Accepted format: <DNS subdomain>/<non-empty name>
- both parts of the plugin name must be non-empty. Accepted format: <DNS subdomain>/<non-empty name>
- first part of the plugin name must be a valid DNS subdomain. Accepted format: <DNS subdomain>/<non-empty name>
```

nrb:
LGTM - I think a short sentence on naming guidelines [in this file](https://github.com/heptio/velero/blob/master/docs/plugins.md) would be helpful, as well as in the example repo.

carlisia:
K, adding!

skriss:
Actually, looking at this one more time, I think we want to add comments to each of the "Register" functions in `pkg/plugin/framework/server.go` that specifies the valid format for plugin names. That's the "SDK" that plugin authors will actually use, so let's document it there.

docs/plugins.md
```diff
@@ -1,14 +1,14 @@
...
-the Velero server pod and copies the binary into a shared emptyDir volume for the Velero server to access. 
+Velero has a plugin architecture that allows users to add their own custom functionality to Velero backups & restores without having to modify/recompile the core Velero binary. To add custom functionality, users simply create their own binary containing implementations of Velero's plugin kinds (described below), plus a small amount of boilerplate code to expose the plugin implementations to Velero. This binary is added to a container image that serves as an init container for the Velero server pod and copies the binary into a shared emptyDir volume for the Velero server to access.
 
 Multiple plugins, of any type,  can be implemented in this binary.
 
-A fully-functional [sample plugin repository][1] is provided to serve as a convenient starting point for plugin authors.
+A fully-functional [sample plugin repository][1] is provided to serve as a convenient starting point for plugin authors. Note that when naming your plugin, the name needs to conform to these rules:
```
skriss: I would probably move the blob on naming to its own section (a la Plugin Logging) - seems a little random here.

docs/plugins.md
```diff
@@ -1,14 +1,14 @@
...
+- have two parts separated by '/'
+- non of the above parts can be empty
```
nrb: none*

Issue #1347, opened 2019-04-04
Restic backup error with RancherOS


ipochi:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is, and what commands you ran.)
In RancherOS the the default location of `/var/lib/kubelet/pods` is not correct.
Rather it is `/opt/rke/var/lib/kubelet/pods`

Hence the restic-daemonset.yaml needs a documentation so that before deployment the yaml is edited to reflect the above change

      ```- name: host-pods
          hostPath:
            path: /var/lib/kubelet/pods
      ```
to

      ```
        - name: host-pods
          hostPath:
            path: /opt/rke/var/lib/kubelet/pods
      ```

**What did you expect to happen:**

**The output of the following commands will help us better understand what's going on**:
(Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)

* `kubectl logs deployment/velero -n velero`
* `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
* `velero backup logs <backupname>`
* `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
* `velero restore logs <restorename>`

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Velero version (use `velero version`):  v0.11.0
- Kubernetes version (use `kubectl version`): v1.13.5
- Kubernetes installer & version: RKE
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`): RancherOS

skriss:
Closed by #1348

PR #1348, opened 2019-04-04
Update Restic documentation for RancherOS

ipochi:
For #1347 

Signed-off-by: Imran Pochi <pochiimran@yahoo.co.in>

### DIFF
--- a/docs/restic.md
+++ b/docs/restic.md
@@ -40,6 +40,21 @@ cross-volume-type data migrations. Stay tuned as this evolves!
 
 1. Run one of the following for your platform to create the daemonset:
 
+   Please note: In RancherOS , the path is not `/var/lib/kubelet/pods` , rather it is `/opt/rke/var/lib/kubelet/pods`
+   thereby requires modifying the restic-daemonset.yaml before applying.
+
+   ```
+   hostPath:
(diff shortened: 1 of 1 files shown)


docs/restic.md
```diff
@@ -40,6 +40,21 @@ cross-volume-type data migrations. Stay tuned as this evolves!
...
+   path: /var/lib/kubelet/pods
+   ```
+
+   to 
+
+   ```
+   hostPath:
+   path: /opt/rke/var/lib/kubelet/pods
+   ```
+
```
kmova: Just wondering if it would be a good idea to add a directory for rancher os, just like aws, azure etc.

nrb: > just like aws, azure etc.

I'm not sure which directories you're referring to for this. The examples directory?

kmova: yes @nrb, the examples directory that can includes restic ds YAML with the modified path for rancher os. However a generic instruction with Rancher as example also is fine.

ipochi:
@kmova @nrb , can we agree upon the docs  and close this ?

docs/restic.md
nrb: @kmova I think that would be fine, but I think this can be merged as is, too.

Issue #138, opened 2017-10-19
pkg/backup should default backup.spec.ttl if it isn't present


ncdc:
If you create a backup by hand and then use `kubectl create -f backup.yaml`, and the backup does not have any value for spec.ttl, it ends up being garbage collected the next time the gc controller runs, because it has no ttl and we end up setting the expiration to "now". We should probably default a backup's TTL in pkg/backup if it's not set.

rosskukulinski:
This is related to #251.  I'm leaning towards recommend defaulting to `0` which means that the backup is 'pinned', or to the default value that `ark backup create` uses.

ncdc:
@rosskukulinski the default `ark backup create` TTL is 30 days. Some options for handling an unset TTL:
1. always default to 30 days, hard-coded in the Ark server code
1. always default to 0 (never expire), hard-coded in the Ark server code
1. allow the admin to specify a default TTL using a flag to `ark server`, and if unspecified, default to 30 days

I think I favor option 3, as that allows the admin some flexibility - they could default to _n_ days or they could opt to default to never expire. And if they don't specify a value for the flag, it defaults to our selected default (which we'd have to decide if we want 0 or some # of days). WDYT?

rosskukulinski:
@ncdc I love the idea of server admins being able to customize the default behavior via CLI flag to `ark server`.

(3) Works for me.

Re: # of days vs 0, I'd be curious to hear more about the reasoning behind the current default of 30 days.  As a backup tool, I would think `never` would be the default (don't automatically delete people's data)

rosskukulinski:
Talked with @ncdc on slack a bit on this topic.

As a backup tool, Ark should be conservative and not delete data by default.  We should default TTL to 0 instead of 30 days.  That said:

"Pinning" a backup (#251) requires updating metadata of a backup object (backup.json) which would need to be synced to other backup targets.  Checksumming of the backup.json for replication may mean that we can't store pinning data in the backup.json . Andy suggested moving the "pinning" metadata to another file in object storage

skriss:
Here's what we need to do to close this issue:

- add a new optional flag to `velero server`, something like `--default-backup-ttl`
- the default value for this flag will be 30 days (easily changeable later if we want to move to 0/never)
- the value of this flag should be passed to the backup controller, and during defaulting, if a backup doesn't have a TTL, the default one should be used

skriss:
Assigning to myself to reserve - Rohan will pick this up.

[...]

PR #1352, opened 2019-04-04
Set default backup TTL

rohandvora:
Set default backup TTL to 30 days when TTL
is not provided in the backup yaml configuration.

Updates #138

Signed-off-by: Rohan Vora <vorar@vmware.com>

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -64,6 +64,7 @@ type backupController struct {
 	backupTracker            BackupTracker
 	backupLocationLister     listers.BackupStorageLocationLister
 	defaultBackupLocation    string
+	defaultBackupTTL         time.Duration
 	snapshotLocationLister   listers.VolumeSnapshotLocationLister
 	defaultSnapshotLocations map[string]string
 	metrics                  *metrics.ServerMetrics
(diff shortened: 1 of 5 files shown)


pkg/controller/backup_controller.go
```diff
@@ -250,6 +253,11 @@ func (c *backupController) prepareBackupRequest(backup *velerov1api.Backup) *pkg
 	// calculate expiration
 	if request.Spec.TTL.Duration > 0 {
 		request.Status.Expiration = metav1.NewTime(c.clock.Now().Add(request.Spec.TTL.Duration))
+	} else {
+		// set default backup TTL
```
skriss: we should also set `request.Spec.TTL` here so that the API object gets patched with the default TTL. So, you might want to reorg L254-261 to first check if there's a TTL, and if not, assign it the default. Then, calculate the `Expiration`, whether there was a TTL specified or not.

pkg/controller/backup_controller.go
```diff
@@ -250,6 +253,11 @@ func (c *backupController) prepareBackupRequest(backup *velerov1api.Backup) *pkg
...
+		if c.defaultBackupTTL > 0 {
+			request.Status.Expiration = metav1.NewTime(c.clock.Now().Add(c.defaultBackupTTL))
```
skriss: please add unit test(s) for this new case.

pkg/controller/backup_controller.go
rohandvora: Sure, will do that.

pkg/controller/backup_controller.go
rohandvora: Done

PR #1358, opened 2019-04-09
instantiate plugin manager with per-restore logger so plugin logs are captured

### LINKED ISSUE #1328: Logs from restore plugins are getting lost

skriss:
Fixes #1328
Replaces #1329 

I extracted the per-restore log into its own simple struct which makes fixing this a little bit easier, without having to combine the two methods mentioned in #1329. I think this is a little bit cleaner for now. Definitely want to further revisit this code (along with all of `pkg/restore`) post-v1.0.

cc @nrb @carlisia

### DIFF
--- a/changelogs/unreleased/1358-skriss
+++ b/changelogs/unreleased/1358-skriss
@@ -1,0 +1,1 @@
+instantiate the plugin manager with the per-restore logger so plugin logs are captured in the per-restore log
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -17,6 +17,7 @@ limitations under the License.
 package controller
 
 import (
+	"bytes"
 	"compress/gzip"
 	"encoding/json"
 	"fmt"
@@ -41,7 +42,6 @@ import (
 	"github.com/heptio/velero/pkg/metrics"
 	"github.com/heptio/velero/pkg/persistence"
 	"github.com/heptio/velero/pkg/plugin/clientmgmt"
-	"github.com/heptio/velero/pkg/plugin/velero"
 	"github.com/heptio/velero/pkg/restore"
 	"github.com/heptio/velero/pkg/util/collections"
 	kubeutil "github.com/heptio/velero/pkg/util/kube"
@@ -91,10 +91,6 @@ type restoreController struct {
 	newBackupStore   func(*api.BackupStorageLocation, persistence.ObjectStoreGetter, logrus.FieldLogger) (persistence.BackupStore, error)
 }
 
-type restoreResult struct {
-	warnings, errors api.RestoreResult
-}
-
 func NewRestoreController(
 	namespace string,
 	restoreInformer informers.RestoreInformer,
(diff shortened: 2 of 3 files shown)


skriss:
Ran some basic manual testing on this and confirmed that logs from plugins now show up in the per-restore log, whereas before they didn't.

skriss:
I can look at doing something like that, though to be clear the comments are not describing a new situation as a result of this PR - it was already true that the log object could be written to after being uploaded, and those log statements would not go into the uploaded log file.

nrb:
Yeah, that's true, but it still relies on programmers remembering that. Having an explicit close and throwing an error for write after close would, hopefully, catch any problems without finding out later that logging was being dropped unintentionally.

If it's not a huge value add, we can leave it out for now.

skriss:
Totally agree it's a good idea in general. Let me look at how much work it would be to accomplish in the current state. If we don't tackle it now, it would make sense to do when we do the rest of #286 to move all restore-related outputs into the log file. I have a feeling that there will be some more significant redesign as part of that, regardless.

skriss:
I did some refactoring in the latest commit to isolate usage of `restoreLog` to just a single function. I also renamed the restoreLog's `getReader(...)` function to `done(...)` and nil'ed out the `FieldLogger` when it's called, so any subsequent call to try to log with it will panic.  I also removed some useless code that was updating `restoreResult` objects in the case of restore failure, even though they'd never be uploaded.  I *hope* it's a little bit clearer now.  All of this code still needs a more thorough review/refactor post-1.0.

nrb:
I'll look over this again in the morning. Thanks for looking into closing out the log!

And agreed for a more holistic refactoring of the restore code in the future.

nrb:
LGTM - do you want to squash with an updated commit message?

skriss:
will do

skriss:
updated

PR #1362, opened 2019-04-12
Add example for restoring with --include-resources

nrb:
Created in response to some user questions on Slack.


pkg/cmd/cli/restore/create.go
```diff
@@ -49,6 +49,9 @@ func NewCreateCommand(f client.Factory, use string) *cobra.Command {
  
   # create a restore from the latest successful backup triggered by schedule "schedule-1"
   velero restore create --from-schedule schedule-1
+
+  # create a restore for only persistenvolumeclaims and persistentvolumes within a backup
```
skriss: s/persistenvolumeclaims/persistentvolumeclaims

PR #1364, opened 2019-04-12
Azure: add support for loading env vars from a file, $AZURE_CREDENTIALS_FILE

skriss:
xref #1287 
cc @nrb 

Turns out this was pretty easy to accomplish. I tested snapshots and restic, both with the new env var + file, and with the old env vars.

So - going forward, we'll have users create a secret file, something like:

```bash
cat > velero-credentials <<EOF
AZURE_TENANT_ID=${AZURE_TENANT_ID}
AZURE_SUBSCRIPTION_ID=${AZURE_SUBSCRIPTION_ID}
AZURE_CLIENT_ID=${AZURE_CLIENT_ID}
AZURE_CLIENT_SECRET=${AZURE_CLIENT_SECRET}
AZURE_RESOURCE_GROUP=${AZURE_RESOURCE_GROUP}
EOF
```

and then secret creation, as well as mounting of the secret into the velero pods, will look just like it does for AWS and GCP.


nrb:
Probably justifies a release note, since new installations going forward should use this.

skriss:
See latest commit, which addresses https://github.com/heptio/velero/pull/1287#issuecomment-482739444

PR #1367, opened 2019-04-12
Restic repository management fixes

skriss:
This does not need to make it into the first alpha, so look at other open PRs first.

Currently if, when creating a new restic repo as part of taking a backup, the repo creation fails, Velero waits for an hour before reporting the failure, because it may "eventually" succeed.  This is a bad UX, though, and if it failed to create the first time, it won't likely succeed the next time without user intervention. So, I changed it so the backup fails fast if the repo creation fails the first time.

I also lowered the resync interval for the restic repo controller, so repos that become "not ready" are retried more often - makes it more likely to get back to a good state after getting into a bad state.

One or two other small fixes in their own commits.

Testing looks good so far, but would like to do some more.

### DIFF
--- a/changelogs/unreleased/1367-skriss
+++ b/changelogs/unreleased/1367-skriss
@@ -1,0 +1,1 @@
+restic repo ensurer: return error if new repository does not become ready within a minute, and fix channel closing/deletion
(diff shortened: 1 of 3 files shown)


nrb:
> I also lowered the resync interval for the restic repo controller, so repos that become "not ready" are retried more often - makes it more likely to get back to a good state after getting into a bad state.

Is this a separate state from the one mentioned initially (that would take an hour to retry)?

skriss:
Yeah - existing restic repositories are periodically checked for integrity, and if they fail that for any reason, they'll become `NotReady`. That's a different case than a new repo that fails to initialize for some reason.

PR #1370, opened 2019-04-15
add some missing config to pkg/install daemonset, deployment

skriss:


skriss:
Once I added this, I was able to `velero install` into Azure, and do a restic backup and restore.

nrb:
Yeah, this was definitely an oversight from the original install structs.

PR #1375, opened 2019-04-16
GCP: Implement configurable snapshot location

### LINKED ISSUE #1361: GCP: Configurable Snapshot location
Currently the snapshots of compute disks on GCP created by Velero are always multi-regional.  This can be changed in the `Snapshot` [struct](https://godoc.org/google.golang.org/api/compute/v1#Snapshot) by supplying `StorageLocations`.

This has been added to the compute API very recently so I guess the vendored files would have to be updated. Also a new [VolumeSnapshotLocation config parameter](https://heptio.github.io/velero/v0.11.0/api-types/volumesnapshotlocation.html#gcp) would need to be added.

I might open a PR if you agree with the approach.

**Environment:**

- Velero version: v0.11.0
- Cloud provider or hardware configuration: GCP

ctrox:
This adds a new config parameter to the `VolumeSnapshotLocation` to set the location where GCP persistent volume snapshots are stored. I had to update the vendored google API as the field was only recently added. Also the field does not take multiple values as [thought before](https://github.com/heptio/velero/issues/1361#issuecomment-483121007). If you specify more than one the API responds with the following message:
```
googleapi: Error 413: Value for field 'snapshot.storageLocations' is too large: maximum size 1 element(s); actual size 2., fieldSizeTooLarge
```

The field just takes any location documented [here](https://cloud.google.com/storage/docs/locations#available_locations).

Fixes #1361


skriss:
Thanks @ctrox, this looks good on first glance? Were you able to successfully test both a backup and a restore?

Gopkg.lock
```diff
@@ -617,7 +617,7 @@
     "transport/http/internal/propagation",
   ]
   pruneopts = "NUT"
-  revision = "3f6e8463aa1d824abe11b439d178c02220079da5"
+  revision = "9e7f2a839bfefc34110ebbc321850d5a7ad1ebd4"
```
skriss: it would be nice to switch off `master` in `Gopkg.toml` for this reference, and onto a tagged release.  Does `v0.3.2` work here?  If so, let's update the `Gopkg.toml` like:
```
[[constraint]]
  name = "google.golang.org/api"
  version = "~v0.3.2"
```

ctrox: For sure, makes sense. `v0.3.2` is recent enough to have the `storageLocations` field.

ctrox:
> Thanks @ctrox, this looks good on first glance! Were you able to successfully test both a backup and a restore?

Yes, I have tested several scenarios. For example a backup of a disk in `europe-west6` to `us-central` and restore that again to a new disk in `europe-west6-c` works fine.

Just noticed that I missed adding the `storageLocation` for regional disks, that is now also done.

PR #1377, opened 2019-04-16
Fix restic repository identifier

jmontleon:


jmontleon:
I think this was missed in the commit below. If the provider is created with the velero.io/ prefix then formatting for the repository is never done and the backup gets stuck during the restic init.

  Restic Identifier:        :bucketname:/velero-backups/restic/nfs-test
vs
  Restic Identifier:        s3:s3-us-east-2.amazonaws.com/bucketname/velero-backups/restic/nfs-test


https://github.com/heptio/velero/commit/c326f59627b8d420bd536748e7a6868db640ee83#diff-1313d8196dcf9bad218ac1a277774978R129

nrb:
Thanks for catching that, @jmontleon!

jmontleon:
@nrb , you're welcome! Let me know if I missed anything. Let me know if it needs any improvements or corrections. I tried to used a similar approach as used in the commit I referenced.

PR #1378, opened 2019-04-16
move RestoreResult from api to pkg/restore, rename to Result

skriss:

This type is not actually an API type -- it's used to serialize data to JSON that gets uploaded to object storage as a file. This PR moves it out of `pkg/apis/velero/v1`.

rel. #286


skriss:
Yeah, the problem is we have variables named `restore` (that refer to Restore API objs) so one or the other has to change.

PR #1386, opened 2019-04-19
Add PartiallyFailed phase for backups, don't exit on first error

skriss:
rel #1371

This PR adds the PartiallyFailed phase for backups.  It modifies the backup execution logic so that rather than immediately exiting when the first error is encountered, the error is logged and execution continues. At the end, the number of errors logged is counted, and a >0 result moves the backup to PartiallyFailed.

Still a WIP, but the basic functionality is working. An easy way to see what it looks like is to create a cluster role binding for a cluster role that doesn't exist, to a service account in the namespace you're backing up -- this will create a "partial failure" error.

@nrb @carlisia appreciate any input at this point!

Issue #1371, opened 2019-04-15
standardize meaning of terminal phases for backups/restores


skriss:
Breaking this sub-issue out from #286

Today, backups and restores terminate with one of the following phases:
  - `FailedValidation` - the item's spec was invalid (restore of a nonexistent backup, etc.)
  - `Completed`
    - for backups, this means there were no errors
    - for restores, this means there were no fatal errors (e.g. failure to download the backup tarball), but there *may* have been 1+ errors restoring individual items
  - `Failed`
    - for backups, this means one or more errors were encountered. It's possible that there's still a restorable tarball in object storage (e.g. if the error was just a single item failing to back up).
    - for restores, this means there was a fatal error, e.g. unable to download the backup tarball

There are a couple of problems with the above:
  - the meaning of `Completed` vs. `Failed` is inconsistent between backups and restores
  - there's no distinct phase to report a partial failure (i.e. most items backed up/restored successfully, but 1+ individual items did not)

I'm proposing the following changes for v1.0:
  - `Completed` means "no errors" for both backups and restores
  - `Failed` means "fatal error encountered" for both backups and restores -- e.g. failure to upload/download tarball
  - `CompletedWithErrors` / `PartiallyFailed` (new phase) means "some errors encountered backing up/restoring individual items" for both backups and restores

Possible future work:
  - `FailedValidation` goes away, is combined with `Failed`
  - eliminate phases and replace with conditions?

skriss:
Closed by #1386 #1389

PR #1389, opened 2019-04-23
Add PartiallyFailed phase for restores

skriss:
rel. #1371 

If a restore has >=1 errors, set its phase to `PartiallyFailed`.  Update CLI output to accomodate this new phase.

xref #1386


skriss:
Did some basic validation on this

PR #1390, opened 2019-04-23
Fail backup if it already exists in object storage

### LINKED ISSUE #623: Disallow overwriting objects in object storage

carlisia:
In lieu of https://github.com/heptio/velero/pull/629

Fixes #623

Notes:

- The test below "as is" was left out of this version. The intention of testing for failures in the case of a backup already existing was incorporated in other existing (current) tests: https://github.com/heptio/velero/pull/629/files#diff-d0a3ff018929aec3abbaa4c8a2c299ccR148

- This commit was in the original PR, but I don't think we want it? https://github.com/heptio/velero/pull/629/commits/d1cd91d7466d81b5fd6cc6a6f3f7159d11e4dd4d

- WIP: working on testing these against the main SCPs.


pkg/controller/backup_controller.go
```diff
@@ -479,6 +498,18 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
 	return kerrors.NewAggregate(errs)
 }
 
+func validateUniqueness(backupStore persistence.BackupStore, bucket, name string) []error {
+	var errs []error
+	exists, err := backupStore.BackupExists(bucket, name)
+	if err != nil {
+		errs = append(errs, errors.Errorf("Error checking if backup already exists in object storage: %v", err))
+	}
+	if exists {
```
skriss: this will only ever be `true` if `err` is `nil`, so I don't think this function really needs to return a slice of errors, could just be a single one.  I would also consider just getting rid of this function and inlining it at L463, since it's just calling `backupStore.BackupExists` and checking the result.

pkg/controller/backup_controller_test.go
```diff
@@ -56,23 +56,24 @@ func (b *fakeBackupper) Backup(logger logrus.FieldLogger, backup *pkgbackup.Requ
 	return args.Error(0)
 }
 
-func TestProcessBackupNonProcessedItems(t *testing.T) {
+func TestProcessBackupProcessing(t *testing.T) {
```
skriss: the intent of this test was to cover all the cases where a backup would *not* be run, hence the "NonProcessedItems" part of the name - ok with a name change but I'd try to convey that intent.

pkg/controller/backup_controller_test.go
```diff
@@ -340,6 +344,83 @@ func TestProcessBackupCompletions(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:           "backup with existing backup will fail",
```
skriss: this test is not actually covering a scenario of an existing backup - see L349 and L364 - not sure if the name just needs to be updated to something like "backup without an existing backup in object storage will succeed", or if this is a dupe of an existing case?

pkg/controller/backup_controller_test.go
```diff
@@ -406,6 +488,8 @@ func TestProcessBackupCompletions(t *testing.T) {
 			res, err := clientset.VeleroV1().Backups(test.backup.Namespace).Get(test.backup.Name, metav1.GetOptions{})
 			require.NoError(t, err)
 
+			// failed tests for failed backup should have a phase of failed
```
skriss: rm

pkg/plugin/framework/object_store_server.go
```diff
@@ -128,6 +128,27 @@ func (s *ObjectStoreGRPCServer) PutObject(stream proto.ObjectStore_PutObjectServ
...
+	}()
+
+	impl, err := s.getImpl(req.Plugin)
+	if err != nil {
+		return nil, newGRPCError(err)
+	}
+
+	exists, err := impl.ObjectExists(req.Bucket, req.Key)
+	if err != nil {
+		return nil, err
```
skriss: `return nil, newGRPCError(err)`

pkg/controller/backup_controller_test.go
carlisia: Great catch. It was a dupe name, just need to change that.

pkg/controller/backup_controller_test.go
carlisia: Turns out the name should not change, it's all unprocessed cases. Changing.

examples/azure/05-backupstoragelocation.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
 spec:
```
skriss: undo changes to this file

examples/azure/06-volumesnapshotlocation.yaml
```diff
@@ -21,4 +21,4 @@ metadata:
 spec:
```
skriss: undo changes to this file

pkg/cloudprovider/aws/object_store.go
```diff
@@ -191,6 +201,50 @@ func (o *ObjectStore) PutObject(bucket, key string, body io.Reader) error {
...
+
+	req := &s3.HeadObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(key),
+	}
+
+	log.Debug("Checking if object exists")
+	_, err := o.s3.HeadObject(req)
+
+	if err != nil {
```
skriss: I would combine this with L221

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -117,6 +122,19 @@ func (o *ObjectStore) PutObject(bucket, key string, body io.Reader) error {
 	return closeErr
 }
 
+func (o *ObjectStore) ObjectExists(bucket, key string) (bool, error) {
+	_, err := o.bucketWriter.getAttrs(bucket, key)
+
+	if err != nil {
```
skriss: combine with L126

skriss:
code LGTM, just waiting on testing

pkg/plugin/framework/object_store_client.go
```diff
@@ -96,6 +96,21 @@ func (c *ObjectStoreGRPCClient) PutObject(bucket, key string, body io.Reader) er
 	}
 }
 
+// ObjectExists checks if there is an object with the given key in the object storage bucket.
+func (c *ObjectStoreGRPCClient) ObjectExists(bucket, key string) (bool, error) {
+	req := &proto.ObjectExistsRequest{
+		Bucket: bucket,
```
skriss: Need to add `Plugin: c.plugin,` between L101 and L102

pkg/controller/backup_controller.go
```diff
@@ -446,9 +452,18 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
 		return err
 	}
 
-	var errs []error
+	exists, err := backupStore.BackupExists(backup.StorageLocation.Spec.StorageType.ObjectStorage.Bucket, backup.Name)
+	if exists || err != nil {
+		backup.Status.Phase = velerov1api.BackupPhaseFailed
+		backup.Status.CompletionTimestamp.Time = c.clock.Now()
+		if err != nil {
+			return errors.Errorf("Error checking if backup already exists in object storage: %v", err)
```
skriss: `errors.Wrapf` is better than `errors.Errorf` here because it retains the inner error + stack trace

skriss:
After locally fixing the first comment from above, I was able to confirm that on Azure, backups correctly fail if there's already a `backups/<backup-name>/velero-backup.json` in object storage, and if that's not the case, they proceed.

carlisia:
Sweet!!! @ncdc did great work!

skriss:
An observation: in the case that this code is intended to catch, we end up with a `Failed` backup in-cluster.  So, now, if I `velero backup delete` this backup -- the files in object storage get deleted. However, this almost certainly wasn't my intent, since those files in object storage weren't created by this backup I just created.

I'm guessing that to fix this, we'll need to add velero-managed UIDs to backups, so that we have a unique ID for each backup across time and space that we can use to determine if two items are the same or not.  I don't think it's in scope for this PR, but we should follow up on it. It's somewhat related to https://github.com/heptio/velero/pull/629#issuecomment-402214368.

nrb:
The UID fix seems reasonable to me. I briefly wondered if it would then be better to use those identifiers on the object storage vs the user-provided name, as that would remove the need for this code. However, that's too invasive a change at this stage, so I'm for just attaching it and using it on sync comparison.

Issue #1387, opened 2019-04-19
Additions to the CONTRIBUTING doc


carlisia:
1) Imports

For Kube/Ark imports, since many of the packages have names like `v1`, we often give them explicit names like `corev1`, `arkv1`, etc. to disambiguate. The problem is that we have no consistency in how we do this, so every file is different.

I'm proposing a standard way of naming these across Ark. The convention would be:

    <group><version><api | client | informer | ...>

So, for example:
```
    import (
        corev1api "k8s.io/api/core/v1"
    	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
    	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
    	corev1listers "k8s.io/client-go/listers/core/v1"
        
        velerov1api ""github.com/heptio/velero/pkg/apis/velero/v1"
        velerov1client "github.com/heptio/velero/pkg/generated/clientset/versioned/typed/velero/v1"
    )
```
Ref: https://github.com/heptio/velero/issues/494

2) Code

- Our standard is to capitalize log messages, but keep actual error messages lower-cased

~3) Protobuf~

~To generate protobuf files, must have version 1.0.0 of protoc-gen-go. Follow instructions in the link below and set git tag to: `GIT_TAG="v1.0.0"`
https://github.com/golang/protobuf#installation~

Protobuf generation was addressed here: https://github.com/heptio/velero/pull/1887

4) ...

nrb:
For 1, we have https://github.com/heptio/velero/issues/494, which has kind of been an "as we go" thing. Maybe it's worth doing a pass just to clean these up?

carlisia:
This issue is about documenting that this is even a thing. We know that but new contributors won't.

skriss:
Let's definitely start to document this stuff. xref #867

skriss:
xref #1104

carlisia:
More things:

>  wrap/add a stack only to errors that are being directly returned from non-velero code (see: https://github.com/heptio/velero/pull/1616#discussion_r299526614)

> use the k8s.io/apimachinery/pkg/util/sets (see: https://github.com/heptio/velero/pull/1616#discussion_r299528781)

carlisia:
More:

Generating mocks for our interfaces.

We use a package to do that. 

Example: if you want to change this mock: https://github.com/heptio/velero/blob/master/pkg/restic/mocks/backupper.go

Run:

```
[...]

PR #1392, opened 2019-04-24
Validate restore name label length

canshul08:
Velero should handle cases when the label length exceeds 63 characters.

- if the length of the backup/restore name is <= 63 characters, use it as the value of the label
- if it's > 63 characters, take the SHA256 hash of the name. the value of
  the label will be the first 57 characters of the backup/restore name
  plus the first six characters of the SHA256 hash.

Fixes heptio#1021

Signed-off-by: Anshul Chandra <anshulc@vmware.com>

### DIFF
--- a/changelogs/unreleased/1392-anshulc
+++ b/changelogs/unreleased/1392-anshulc
@@ -1,0 +1,1 @@
+shorten label values when they're longer than 63 characters
--- a/pkg/label/label.go
+++ b/pkg/label/label.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2019 the Velero contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package label
+
+import (
+	"crypto/sha256"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+// GetValidName converts an input string to valid kubernetes label string in accordance to rfc1035 DNS Label spec
+// (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/identifiers.md)
+// Length of the label is adjusted basis the DNS1035LabelMaxLength (defined at k8s.io/apimachinery/pkg/util/validation)
+// If length exceeds, we trim the label name to contain only max allowed characters
+// Additionally, the last 6 characters of the label name are replaced by the first 6 characters of the sha256 of original label
+func GetValidName(label string) string {
+	if len(label) <= validation.DNS1035LabelMaxLength {
+		return label
+	}
+
+	sha := sha256.Sum256([]byte(label))
+	strSha := fmt.Sprintf("%x", sha)
+	charsFromLabel := validation.DNS1035LabelMaxLength - 6
+	if charsFromLabel < 0 {
+		// Derive the label name from sha hash in case the DNS1035LabelMaxLength is less than 6
+		return string(strSha[validation.DNS1035LabelMaxLength])
+	}
+
+	return label[:charsFromLabel] + strSha[:6]
+}
--- a/pkg/backup/delete_helpers.go
+++ b/pkg/backup/delete_helpers.go
@@ -21,7 +21,8 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	v1 "github.com/heptio/velero/pkg/apis/velero/v1"
+	"github.com/heptio/velero/pkg/apis/velero/v1"
+	"github.com/heptio/velero/pkg/label"
 )
 
 // NewDeleteBackupRequest creates a DeleteBackupRequest for the backup identified by name and uid.
@@ -30,7 +31,7 @@ func NewDeleteBackupRequest(name string, uid string) *v1.DeleteBackupRequest {
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: name + "-",
 			Labels: map[string]string{
-				v1.BackupNameLabel: name,
+				v1.BackupNameLabel: label.GetValidName(name),
 				v1.BackupUIDLabel:  uid,
 			},
 		},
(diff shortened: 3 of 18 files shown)


pkg/labels/labels.go
```diff
@@ -0,0 +1,56 @@
...
+	"crypto/sha256"
+	"fmt"
+)
+
+// maxLabelLength stores the max label length value supported by kubernetes.
+const maxLabelLength = 63
+
+// GetValidLabelName converts an input string to valid kubernetes label string in accordance to rfc1035 DNS Label spec
+// (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/identifiers.md)
+func GetValidLabelName(label string) string {
```
nrb: I would advocate for changing this function name to `GetValidName` and changing the import alias in other files to `labels`. In use, it would then read as `labels.GetValidName(input)`, which has less repetition.

canshul08: Yes, I think it a good idea. I can update the function name to GetValidName. As for the alias, I wanted to use "labels", however, it collides with "k8s.io/apimachinery/pkg/labels". Can we use "label" instead?

nrb: Ah, good point. We have a convention for these conflicts that's not yet made its way into the contributing docs (See #1387).

I think `velerolabels` would make sense with the convention.

pkg/labels/labels.go
```diff
@@ -0,0 +1,56 @@
...
+			// Derive the label name from sha hash in case the maxLabelLength is less than 6
+			return string(strSha[maxLabelLength])
+		}
+		label = label[:charsFromLabel] + strSha[:6]
+	}
+	return label
+}
+
+// GetMaxConfiguredLabelLength returns the max label length configured
+func GetMaxConfiguredLabelLength() int {
```
carlisia: You may remove this method in lieu of making `maxLabelLength` external, ie `MaxLabelLength`. In this case, I'd rename it to `MaxLength` so we'll have `labels.MaxLength` and not `labels.MaxLabelLength`.

On this note, please rename the package to `label` so it's singular. This is on par with our other package names. Then you can alias this package as `veleroLabel` or not alias it at all.

pkg/labels/labels.go
```diff
@@ -0,0 +1,56 @@
...
+	finalLabel := handleLabelLength(label)
+
+	return finalLabel
+}
+
+// handleLabelLength adjusts the label length when it exceeds the maxLabelLength (i.e. 63 characters)
+// if length exceeds, we trim the label name to contain only max allowed characters
+// additionally, the last 6 characters of the label name are replaced by the first 6 characters of the sha256 of original label
+func handleLabelLength(label string) string {
+	if len(label) > maxLabelLength {
```
carlisia: I would rather have this code folded under the `GetValidName` function. There is no value in having two functions. It's ok to add the comment from this one to the other one, even if it's gonna be long.

pkg/labels/labels.go
canshul08: Sure.

pkg/labels/labels.go
canshul08: Sure. I had created a function as earlier I had yet another function where I was validating the characters for the label (should be alphanumeric etc). Later, I removed that validation as Steve suggested it was not required. I'll move this to GetValidName function.

skriss:
@canshul08 this looks like it's in pretty good shape. Have you been able to do any manual testing at all?

canshul08:
> @canshul08 this looks like it's in pretty good shape. Have you been able to do any manual testing at all?
@skriss Yes, I tried testing the functionality by creating backups with name more than 63 characters and later restoring the same. However, I'm not very sure of Restic, I'll check with you on that.

skriss:
OK - I can help out with some testing there

pkg/label/label.go
```diff
@@ -0,0 +1,44 @@
...
+
+package label
+
+import (
+	"crypto/sha256"
+	"fmt"
+)
+
+// MaxLength stores the max label length value supported by kubernetes.
+const MaxLength = 63
```
carlisia: I wonder if we should delete this and use a call to here instead: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L156

canshul08: Sure, I will update.

pkg/label/label.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"crypto/sha256"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+// GetValidName converts an input string to valid kubernetes label string in accordance to rfc1035 DNS Label spec
+// (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/identifiers.md)
+// Length of the label is adjusted basis the MaxLength (i.e. 63 characters)
```
nrb: // Length of the label is adjusted based on the MaxLength (i.e. 63 characters)

pkg/label/label.go
```diff
@@ -0,0 +1,44 @@
...
+// If length exceeds, we trim the label name to contain only max allowed characters
+// Additionally, the last 6 characters of the label name are replaced by the first 6 characters of the sha256 of original label
+func GetValidName(label string) string {
+	maxLength := validation.DNS1035LabelMaxLength
```
nrb: This is a nitpick, but is the variable assignment necessary here?

pkg/label/label.go
nrb: `// Length of the label is adjusted based on the MaxLength (63 characters)`

pkg/label/label.go
canshul08: I'll remove the assignment.

pkg/label/label.go
canshul08: Will fix the comment.

changelogs/unreleased/1392-anshulc
```diff
@@ -0,0 +1 @@
+Validate restore name label length
```
skriss: Can you update this to something like "shorten label values when they're longer than 63 characters"?

pkg/label/label.go
```diff
@@ -0,0 +1,43 @@
...
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+// GetValidName converts an input string to valid kubernetes label string in accordance to rfc1035 DNS Label spec
+// (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/identifiers.md)
+// Length of the label is adjusted basis the DNS1035LabelMaxLength (defined at k8s.io/apimachinery/pkg/util/validation)
+// If length exceeds, we trim the label name to contain only max allowed characters
+// Additionally, the last 6 characters of the label name are replaced by the first 6 characters of the sha256 of original label
+func GetValidName(label string) string {
+	if len(label) > validation.DNS1035LabelMaxLength {
```
skriss: I would invert this `if`, so check if the length is less than the max, and if so return immediately. Then you can un-nest the rest of the logic.

changelogs/unreleased/1392-anshulc
canshul08: Done

pkg/label/label.go
canshul08: Done.

nrb:
Awesome, thanks for the PR and your patience, @canshul08!

PR #1376, opened 2019-04-16
Document install command

nrb:
Signed-off-by: Nolan Brubaker <brubakern@vmware.com>

PR #1393, opened 2019-04-24
update IBM install doc to use velero install, remove examples

skriss:

xref #1376


docs/ibm-config.md
```diff
@@ -51,40 +51,34 @@ Several comments:
...
-## Credentials and configuration
+## Install and start Velero
 
-In the Velero directory (i.e. where you extracted the release tarball), run the following to first set up namespaces, RBAC, and other scaffolding. To run in a custom namespace, make sure that you have edited the YAML files to specify the namespace. See [Run in custom namespace][0].
+Install Velero, including all prerequisites, into the cluster and start the deployment. This will create a namespace called `velero`, and place a deployment named `velero` in it.
 
 ```bash
-kubectl apply -f config/common/00-prereqs.yaml
+velero install \
+    --provider aws \
```
nrb: I wonder, would this make more sense to be a general AWS-compatible page? I kinda don't want to carry anything specific except minio and the big 3 cloud providers.

skriss: Yeah, I definitely would like to move the IBM stuff out of the core docs and add it to some kind of `contrib` section like we talked about, to clarify that we're not officially supporting/testing it.  I'd vote to add an issue for that change and do a follow up for it, since we should think a little about the structure before moving stuff. does that work for you?

PR #1405, opened 2019-04-25
GCP: add optional 'project' config to volume snapshot location

### LINKED ISSUE #1268: GCP: Ability to restore snapshots from different projects
I'm trying to migrate a cluster from one Google Cloud Project to another, but the restoration fails because it looks like velero derives the project ID for the disk snapshots from the generated service credentials, which references the project ID the cluster is in, as opposed to the source project.

Ideally, if there was a way to specify the source project ID in either the VolumeSnapshotLocation resource, or as a CLI flag, then this would be great.

**Environment:**

- Velero version (use `velero version`): 0.11.0
- Cloud provider or hardware configuration: GKE
- OS (e.g. from `/etc/os-release`): Google container optimised OS

skriss:

Fixes #1268 

Still needs to be tested.

### DIFF
--- a/changelogs/unreleased/1405-skriss
+++ b/changelogs/unreleased/1405-skriss
@@ -1,0 +1,1 @@
+GCP: add optional 'project' config to volume snapshot location for if snapshots are in a different project than the IAM account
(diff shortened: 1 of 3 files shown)


skriss:
This is incomplete, need to make a few more modifications.

skriss:
pushed an update.

skriss:
pls squash & merge

PR #1407, opened 2019-04-25
Throw an error on an invalid bucket name

nrb:
This PR is to help troubleshoot some issues arising from unset environment variables when running the `velero install` command. More of a UX/early warning change than major bug fix.

### DIFF
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/cmd/cli/install/install.go
```diff
@@ -208,6 +209,11 @@ func (o *InstallOptions) Validate(c *cobra.Command, args []string, f client.Fact
 
 	if o.BucketName == "" {
 		return errors.New("--bucket is required")
+	} else if strings.HasPrefix(o.BucketName, "-") {
```
carlisia: You don't need the `else` here :)

PR #1415, opened 2019-04-29
remove all remaining legacy Ark references

skriss:

I think this removes the last batch of "Ark" refs from the code, apart from docs that explicitly discuss the migration from Heptio Ark to Velero.

@nrb @carlisia

### DIFF
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -58,18 +58,15 @@ var nonRestorableResources = []string{
 
 	// Don't ever restore backups - if appropriate, they'll be synced in from object storage.
 	// https://github.com/heptio/velero/issues/622
-	"backups.ark.heptio.com",
 	"backups.velero.io",
 
 	// Restores are cluster-specific, and don't have value moving across clusters.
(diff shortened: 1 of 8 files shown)


pkg/controller/restore_controller.go
```diff
@@ -58,18 +58,15 @@ var nonRestorableResources = []string{
 
 	// Don't ever restore backups - if appropriate, they'll be synced in from object storage.
 	// https://github.com/heptio/velero/issues/622
-	"backups.ark.heptio.com",
```
skriss: @nrb @carlisia I think removing these are fine, since users should've migrated their CRDs, and we won't be processing the *.heptio.com API group anymore - make sense?

nrb: I agree users should have, though I'm unsure that everyone will.

That said, since our controllers won't actually process anything in the old API group, the restored Ark backups will be inert, so it's not going to trigger a cascade.

skriss: Yep, that was my thinking.

PR #1416, opened 2019-04-29
switch to snapshots cmd for checking repo exists

skriss:

#1171 switched from using `restic check` to `restic stats` for determining if a restic repository existed.  It turns out that `restic stats` is still not the optimal choice, as it can be slow on very large repositories.  

Per https://restic.readthedocs.io/en/stable/075_scripting.html#check-if-a-repository-is-already-initialized, the recommended command to use is `restic snapshots`. This PR makes that switch.  I'm also using the `--last` flag to minimize the amount of data returned from the command, but this is just a small optimization.

### DIFF
--- a/changelogs/unreleased/1416-skriss
+++ b/changelogs/unreleased/1416-skriss
@@ -1,0 +1,1 @@
+switch from `restic stats` to `restic snapshots` for checking restic repository existence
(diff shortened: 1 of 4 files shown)


skriss:
manual testing LGTM.

PR #1420, opened 2019-04-30
doc on upgrading to v1.0

skriss:

@nrb @carlisia 

This will replace the existing "Upgrade from Ark to Velero" doc - I need to decide exactly how to rejigger the TOC on the docs site - may get rid of the Upgrade section and just move this into Install or About.

### DIFF
--- a/docs/upgrade-to-1.0.md
+++ b/docs/upgrade-to-1.0.md
@@ -1,0 +1,85 @@
+# Upgrading to Velero 1.0
+
+## Prerequisites
+- Velero v0.11 installed. If you're not already on v0.11, see the [instructions for upgrading to v0.11][0]. **Upgrading directly from v0.10.x or earlier to v1.0 is not supported!**
+- (Optional, but strongly recommended) Create a full copy of the object storage bucket(s) Velero is using. Part 1 of the upgrade procedure will modify the contents of the bucket, so we recommend creating a backup copy of it prior to upgrading.
+
+## Instructions
+
+### Part 1 - Rewrite Legacy Metadata
+
+#### Overview
+
+You need to replace legacy metadata in object storage with updated versions **for any backups that were originally taken with a version prior to v0.11 (i.e. when the project was named Ark)**. While Velero v0.11 is backwards-compatible with these legacy files, Velero v1.0 is not.
+
+_If you're sure that you do not have any backups that were originally created prior to v0.11 (with Ark), you can proceed directly to Part 2._
+
+We've added a CLI command to [Velero v0.11.1][1], `velero migrate-backups`, to help you with this. This command will:
+
+- Replace `ark-backup.json` files in object storage with equivalent `velero-backup.json` files. 
+- Create `<backup-name>-volumesnapshots.json.gz` files in object storage if they don't already exist, containing snapshot metadata populated from the backups' `status.volumeBackups` field*. 
+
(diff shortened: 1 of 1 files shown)


carlisia:
Isn't this structure all going to change with the imminent website?

carlisia:
I have suggestions for a different break down of the content for your consideration.

- I would put the pre-requisite section at the very top.
- I would delete the first paragraph.
- I could cut the 2nd paragraph and bullets (maybe it can be reused below).

Then I would split the instruction section into:
Upgrading from v0.11.1
- Description of what will be done.
- Instructions

Upgrading to v0.11.1 (Alternatively, could be titled "Upgrading from versions earlier than v0.11.1" or something like that.
"Maybe" add something brief about the need to upgrade to v0.11.1 before upgrading to v1.0. Maybe the content below will be self-evident.
- Description of what will be done.
- Instructions

I *think* this works. The intent is to guide people more obviously to the use case they find themselves in.

skriss:
I pushed an update which I *think* mostly addresses your comments. I left the order of the two parts as-is since they need to be executed in that order.  Let me know if I didn't address anything specific that you still think needs to change!

carlisia:
If you are already on version 0.11.1, do you need to do part 1?

skriss:
It's not about what version of Velero you're currently running, it's about whether you have backups around that were originally created prior to v0.11.

carlisia:
Ah yeah, I hadn't totally grasped this fact. Love it that you added `While Velero v0.11 is backwards-compatible with these files, Velero v1.0 is not.`.

This structure looks good. Gonna add 1 tiny comment.

docs/upgrade-to-1.0.md
```diff
@@ -0,0 +1,85 @@
...
+- Velero v0.11 installed. If you're not already on v0.11, see the [instructions for upgrading to v0.11][0]. **Upgrading directly from v0.10.x or earlier to v1.0 is not supported!**
+- (Optional, but strongly recommended) Create a full copy of the object storage bucket(s) Velero is using. Part 1 of the upgrade procedure will modify the contents of the bucket, so we recommend creating a backup copy of it prior to upgrading.
+
+## Instructions
+
+### Part 1 - Rewrite Legacy Metadata
+
+#### Overview
+
+If you have any backups that were originally taken with a version prior to v0.11 (i.e. when the project was named Ark), you need to replace the legacy backup metadata in object storage with updated versions. While Velero v0.11 is backwards-compatible with these files, Velero v1.0 is not.
```
carlisia: `s/If you have any backups/In case` you have any backups.

For some reason this sounds more like a prompt to be cautious and run this part 1.

skriss: that edit sounds slightly strange to my ear, but i pushed an update that i think accomplishes the same thing.

carlisia:
Yes, looks great now.

docs/upgrade-to-1.0.md
carlisia: Yes, looks great now.

skriss:
squashed, still need to decide on TOC update.

carlisia:
I'd move it to the install page and on that page have a TOC 

- Install
- Update

then the sections.

skriss:
@nrb this is ready to merge if you're happy with it. I'll do updates on `gh-pages` next.

Issue #1424, opened 2019-05-01
build-from-scratch.md needs to be updated for `velero install`


skriss:
build-from-scratch.md has a bunch of references to the `examples/` directory which will no longer contain platform-specific YAML.  It needs to be updated for v1.0.

PR #1425, opened 2019-05-01
remove refs to pre-v1.0 velero versions from docs

skriss:

Here's a first batch of updates to remove references to pre-v1.0 versions of Velero from the `master` docs.

A few related tasks:
- I filed #1424 related to this  
- locations.md also needs some more significant rework - I'll work on this next
- migrating-to-velero.md should be dropped from master just before we cut 1.0 (it'll be retained in the v0.11-tagged docs).

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -26,12 +26,15 @@ in [pod_action.go](https://github.com/heptio/velero/blob/master/pkg/restore/pod_
 
 ## I'm using Velero in multiple clusters. Should I use the same bucket to store all of my backups?
 
-We **strongly** recommend that you use a separate bucket per cluster to store backups. Sharing a bucket
-across multiple Velero instances can lead to numerous problems - failed backups, overwritten backups,
-inadvertently deleted backups, etc., all of which can be avoided by using a separate bucket per Velero
-instance.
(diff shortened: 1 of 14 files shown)


docs/faq.md
```diff
@@ -26,12 +26,15 @@ in [pod_action.go](https://github.com/heptio/velero/blob/master/pkg/restore/pod_
 
 ## I'm using Velero in multiple clusters. Should I use the same bucket to store all of my backups?
 
-We **strongly** recommend that you use a separate bucket per cluster to store backups. Sharing a bucket
-across multiple Velero instances can lead to numerous problems - failed backups, overwritten backups,
-inadvertently deleted backups, etc., all of which can be avoided by using a separate bucket per Velero
-instance.
+We **strongly** recommend that each Velero instance use a distinct bucket/prefix combination to store backups.
```
nrb: I think this could be clarified that a bucket/prefix can have multiple readers (so read-only BSLs or a server in restore mode), but only 1 writer.

skriss: I did some reorganization of the response - this is covered a little more explicitly in the third paragraph now.

PR #1428, opened 2019-05-02
Add ability to use wildcard in includes/excludes

### LINKED ISSUE #373: Support wildcards for --include-resources and --exclude-resources

guilhem:
Use https://github.com/gobwas/glob as glob engine
Fix #373

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -145,6 +145,23 @@
   version = "v1.28.2"
 
 [[projects]]
+  digest = "1:6a7159c5f8f8826207545407a458b43ab6599c1c8a2271465c2e979ecea1dcd4"
+  name = "github.com/gobwas/glob"
+  packages = [
+    ".",
+    "compiler",
+    "match",
+    "syntax",
+    "syntax/ast",
+    "syntax/lexer",
+    "util/runes",
+    "util/strings",
+  ]
+  pruneopts = "NUT"
+  revision = "5ccd90ef52e1e632236f7326478d4faa74f99438"
+  version = "v0.2.3"
+
+[[projects]]
   digest = "1:a6afc27b2a73a5506832f3c5a1c19a30772cb69e7bd1ced4639eb36a55db224f"
   name = "github.com/gogo/protobuf"
   packages = [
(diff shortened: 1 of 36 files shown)


skriss:
thanks for this @guilhem! we'll check it out next week and see if we can get it into v1.0.

skriss:
@guilhem we're going to hold off on this until after 1.0.  One thing I noticed is that it looks like you may have an older version of `dep` based on some of the updates that were made to `Gopkg.lock` - would you mind upgrading to `v0.5.x` and re-running `dep ensure`? Thanks.

guilhem:
@skriss oops
I just run `dep ensure` against dep 0.5.2

Gopkg.lock
```diff
@@ -144,6 +144,23 @@
...
+    "match",
+    "syntax",
+    "syntax/ast",
+    "syntax/lexer",
+    "util/runes",
+    "util/strings",
+  ]
+  pruneopts = "NUT"
+  revision = "5ccd90ef52e1e632236f7326478d4faa74f99438"
+  version = "v0.2.3"
```
skriss: Please add an entry to `Gopkg.toml` that version-locks this dependency.  `~v0.2.3` should be fine.

Gopkg.lock
```diff
@@ -1069,6 +1087,7 @@
     "k8s.io/apimachinery/pkg/runtime/serializer",
     "k8s.io/apimachinery/pkg/types",
     "k8s.io/apimachinery/pkg/util/clock",
+    "k8s.io/apimachinery/pkg/util/diff",
```
skriss: any idea why this got added? i don't see any refs to it

pkg/util/collections/includes_excludes.go
```diff
@@ -19,24 +19,46 @@ package collections
...
+	"github.com/gobwas/glob"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+type globStringSet struct {
+	sets.String
+}
+
+func NewGlobStringSet() globStringSet {
```
skriss: let's keep this unexported since it's only used internally.

pkg/util/collections/includes_excludes.go
```diff
@@ -19,24 +19,46 @@ package collections
...
+	return globStringSet{sets.NewString()}
+}
+
+func (gss globStringSet) Match(match string) bool {
```
skriss: unexport

skriss:
@nrb @carlisia PTAL as well.

Gopkg.lock
guilhem: Absolutely not :)

skriss:
Thanks for the updates @guilhem - this looks good to go - can you squash down to one commit and make sure you've added DCO signoff (`git commit --amend --signoff` once you've squashed would do it)? Thanks!

PR #1432, opened 2019-05-02
more docs updates

skriss:
Continuing with some additional docs updates:

- I revised the `versions.md` page for 1.0. I'm not 100% sure we need this page in the future, but leaving some minimal content there for now.
- I moved information about supported Kubernetes versions into the `support-matrix.md` page, which I think is a more natural home for it. In the TOC this page is called "Supported platforms" so I think that makes sense.
- I took the content in `expose-minio.md` and appended it to the bottom of `get-started.md`, which is the only page it's referenced from and relevant for.  

I have some revisions to the table of contents to make after these changes -- I'll just open a PR since it'll be easiest to see the changes by looking at the diff.


docs/get-started.md
```diff
@@ -13,11 +13,11 @@ If you encounter issues with installing or configuring, see [Debugging Installat
 
 ### Prerequisites
 
-* Access to a Kubernetes cluster, version 1.7 or later. Version 1.7.5 or later is required to run `velero backup delete`.
+* Access to a Kubernetes cluster, version 1.7 or later.
```
skriss: FYI, I believe the 1.7.5 note was a leftover from when we used finalizers for doing backup deletion - no longer the case.

nrb: Have you or @carlisia tested with a 1.7 cluster? Are we sure that we still support that far back?

docs/get-started.md
```diff
@@ -191,8 +191,59 @@ kubectl delete crds -l component=velero
 kubectl delete -f examples/nginx-app/base.yaml
 ```
 
-[31]: expose-minio.md
+## Expose Minio outside your cluster
+
+When you run commands to get logs or describe a backup, the Velero server generates a pre-signed URL to download the requested items. To access these URLs from outside the cluster -- that is, from your Velero client -- you need to make Minio available outside the cluster. You can:
+
+- Change the Minio Service type from `ClusterIP` to `NodePort`.
```
nrb: fyi, these don't work with KinD; using KinD, one must use `kubectl port-foward` due to how docker networking is set up. https://github.com/kubernetes-sigs/kind/issues/99 for details.

docs/versions.md
```diff
@@ -1,6 +1,6 @@
 # Upgrading Velero versions
 
-Velero supports multiple concurrent versions. Whether you're setting up Velero for the first time or upgrading to a new version, you need to pay careful attention to versioning. This doc page is new as of version 0.10.0, and will be updated with information about subsequent releases.
+Velero supports multiple concurrent versions. Whether you're setting up Velero for the first time or upgrading to a new version, you need to pay careful attention to versioning.
```
nrb: Reading this paragraph, I'm not 100% sure what the first sentence means, especially as we move into 1.0. Does it mean that the Velero team supports multiple concurrent versions? Or that users can run multiple concurrent versions in the cluster?

docs/get-started.md
skriss: I have not, but am not aware of any evidence to the contrary

docs/get-started.md
skriss: OK - let's file a docs issue to add instructions for that case

docs/versions.md
skriss: It means the former. Should I just delete this page?

docs/get-started.md
nrb: Created #1436

docs/get-started.md
nrb: Ok. 

Could we include (or move) the note about 1.10/mount propagation for restic here? I could see someone reading only this page and getting confused why restic didn't work.

docs/versions.md
nrb: That works for me; I'm not seeing a whole lot of unique information here.

skriss: Yeah, I think the intention was to describe breaking changes across pre-1.0 versions, but that shouldn't really be a concern going fwd. And we've added other notes to tell people to use official released versions and version-tagged docs. I'll go ahead and remove.

docs/get-started.md
skriss: I'm fine copying it here

skriss: added

docs/versions.md
skriss: deleted

PR #1433, opened 2019-05-02
Update gh-pages 'master' docs with latest docs, and reorg TOC

skriss:
I'll need to revise the first commit to capture the docs changes from #1432 when they're merged, but you can look at the second commit which has the updates to the TOC.  Here's a summary of what I changed:

- Move “Getting started” from Introduction to Install, as “Quick start with in-cluster Mino”
- Move “About locations” from Upgrade to Introduction
- Move “Get the right Velero version” to the bottom of Introduction
- Rename “Set up Velero” to “Overview” as first page under Install
- Remove “Upgrade” section, put “Upgrade to 1.0” just under “Overview” in Install
- Alphabetize AWS/Azure/GCP install pages
- Remove “IBM Cloud Object Store” from Install section (the page is still linked from support-matrix.md)
- Rename Tutorials section to Use, move to just under Install section

I think these changes help a lot in terms of organization and flow, and putting information where it belongs.  We can continue to build/revise from here.


_data/master-toc.yml
```diff
@@ -7,37 +7,37 @@ toc:
...
-  - title: Upgrade
-    subfolderitems:
-      - page: Upgrade from Ark to Velero 
-        url: /migrating-to-velero
       - page: About locations
         url: /locations
+      - page: Supported platforms
+        url: /support-matrix
+      - page: Get the right Velero version
+        url: /versions
```
nrb: This link can go per review on #1432

_data/master-toc.yml
```diff
@@ -7,37 +7,37 @@ toc:
...
   - title: Install
     subfolderitems:
-      - page: Set up Velero
+      - page: Overview
         url: /install-overview
-      - page: Expose Minio
-        url: /expose-minio
+      - page: Upgrade to 1.0
+        url: /upgrade-to-1.0
+      - page: Quick Start with In-Cluster Minio
```
nrb: :+1: to putting Minion in the title here.

master/locations.md
```diff
@@ -1,30 +1,20 @@
...
-Additionally, since multiple `VolumeSnapshotLocations` can be created, the user can now configure locations for more than one volume provider, and if the cluster has volumes from multiple providers (e.g. AWS EBS and Portworx), all of them can be snapshotted in a single Velero backup.
+The user can pre-configure one or more possible `BackupStorageLocations` and one or more `VolumeSnapshotLocations`, and can select *at backup creation time* the location in which the backup and associated snapshots should be stored.
+
+This configuration design enables a number of different use cases, including:
+
+- Take snapshots of more than one kind of persistent volume in a single Velero backup (e.g. in a cluster with both EBS volumes and Portworx volumes)
+- Have some Velero backups go to a bucket in an eastern USA region, and others go to a bucket in a western USA region
+- For volume providers that support it (e.g. Portworx), have some snapshots be stored locally on the cluster and have others be stored in the cloud
 
 ## Limitations / Caveats
```
nrb: It's probably worth noting that Velero only supports 1 set of credentials per provider here.

master/upgrade-to-1.0.md
```diff
@@ -0,0 +1,85 @@
+# Upgrading to Velero 1.0
+
+## Prerequisites
+- Velero v0.11 installed. If you're not already on v0.11, see the [instructions for upgrading to v0.11][0]. **Upgrading directly from v0.10.x or earlier to v1.0 is not supported!**
```
nrb: Should we be even more explicit and required `v0.11.1`, so the `migrate-backups` command is present?

master/locations.md
skriss: Yeah, good idea, I'll add a PR to update that in master then include it over here.

master/upgrade-to-1.0.md
skriss: The nuance there is that they don't need to be running v0.11.1 on the server, they just need the v0.11.1 client.  That's why i left this as just v0.11, and included the v0.11.1 CLI download instructions below. WDYT?

nrb: That makes sense. In that case it can be left.

skriss:
Addressed open comments in latest commits

PR #1434, opened 2019-05-02
Separate upload part for the BlockStore

satyamz:
VolumeSnapshotter i.e BlockStore interface should have an extra method,
called as `UploadSnapshot()`. This method should make use of all the tags and
snapshot objects used for `CreateSnapshot()` method. Hence, add a way to store
these information to the `itemBackupper`.

Signed-off-by: Satyam Zode <satyamzode@gmail.com>


satyamz:
Please ignore this; This is just a POC.

PR #1438, opened 2019-05-03
add note about single credentials per provider to locations doc

skriss:

Addresses https://github.com/heptio/velero/pull/1433#discussion_r280838354

### DIFF
--- a/docs/locations.md
+++ b/docs/locations.md
@@ -18,6 +18,8 @@ This configuration design enables a number of different use cases, including:
 
 ## Limitations / Caveats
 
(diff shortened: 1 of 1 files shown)


docs/locations.md
```diff
@@ -18,6 +18,8 @@ This configuration design enables a number of different use cases, including:
 
 ## Limitations / Caveats
 
+- Velero only supports a single set of credentials *per provider*. It's not yet possible to use different credentials for different locations, if they're for the same provider.
```
carlisia: Wonder if it'd be a ton more clear to say `even if`. Otherwise it's fine as is.

skriss: that wouldn't be accurate though. you can use different credentials for different providers.

PR #1462, opened 2019-05-08
Add flag to disable creation of VSL on install

### LINKED ISSUE #1453: Consider adding a flag to `velero install` to disable volume snapshots

Add a flag, e.g. `--use-volume-snapshots`, that's `true` by default, that if false, doesn't create any `VolumeSnapshotLocations`.  This would be useful for our quick-start, where we make the user delete the VSL after setup, or for any other users who don't want to use snapshots.

cc @nrb

nrb:
Fixes #1453

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -61,16 +61,14 @@ These instructions start the Velero server and a Minio instance that is accessib
         --provider aws \
         --bucket velero \
         --secret-file ./credentials-velero \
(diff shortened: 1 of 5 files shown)


nrb:
@skriss renamed, and documentation updated

skriss:
Oh, actually, `ibm-config.md` could be updated to use this flag and remove the step to delete VSLs too.

Issue #1464, opened 2019-05-09
Velero could generate common crontab expression from CLI flag


alexbrand:
**Describe the solution you'd like**
As a user, I would like to leverage the `velero` CLI to figure out common crontab expressions for me.

For example:

Create a scheduled backup that runs every 5 minutes:
`./velero schedule create --every=5m`

Create a scheduled backup that runs every 1 hour:
`./velero schedule create --every=1h`

skriss:
@alexbrand you can actually use the "@every" syntax to create schedules -- e.g. `velero schedule create foo --schedule '@every 5m'`.  We're using the https://godoc.org/github.com/robfig/cron#ParseStandard method for parsing.

alexbrand:
Oh sweet! Is that documented anywhere? Might be worth documenting or adding an example in the CLI usage docs if not documented.

skriss:
it's probably not :) we can leave this issue open to track adding some docs/examples.

skriss:
I'm going to consider this issue resolved as @nrb added examples to the CLI help in #1473, but feel free to reopen as needed.

PR #1473, opened 2019-05-13
Add more examples to schedule create

nrb:
Partially addresses #1464


nrb:
I think a page in the docs for schedules would be good, but also this is (hopefully) easier to find. @alexbrand does this address your concerns?

PR #1487, opened 2019-05-15
Add documentation for fetching credentials

nrb:
Since backup migration accesses the buckets from the client computer and
doesn't interact with the server, the user needs to get the correct
secrets for their backup storage locations in order to actually access
the bucket.

### DIFF
--- a/docs/upgrade-to-1.0.md
+++ b/docs/upgrade-to-1.0.md
@@ -42,6 +42,19 @@ _*backups created prior to v0.10 stored snapshot metadata in the `status.volumeB
     velero backup-location get
     ```
 
+1. Fetch your bucket credentials and make them accessible to Velero
+
+    ```bash
+    # AWS
+    AWS_SHARED_CREDENTIALS_FILE=$(kubectl get secrets/<YOUR_SECRET_NAME> -n velero -o jsonpath="{.data.cloud}" | base64 -d)
+
+    #GCP
+    GOOGLE_APPLICATION_CREDENTIALS=$(kubectl get secrets/<YOUR_SECRET_NAME> -n velero -o jsonpath="{.data.cloud}" | base64 -d)
+
+    # Azure
+    source $(kubectl get secrets/<YOUR_SECRET_NAME> -n velero -o jsonpath="{.data.cloud}" | base64 -d)
+    ```
+
 1. For each backup storage location that you want to use with Velero 1.0, replace any legacy pre-v0.11 backup metadata with the equivalent current formats:
     ```bash
     # - BACKUP_LOCATION_NAME is the name of a backup location from the previous step, whose


docs/upgrade-to-1.0.md
```diff
@@ -42,6 +42,19 @@ _*backups created prior to v0.10 stored snapshot metadata in the `status.volumeB
     velero backup-location get
     ```
 
+1. Fetch your bucket credentials and make them accessible to Velero
+
+    ```bash
+    # AWS
+    AWS_SHARED_CREDENTIALS_FILE=$(kubectl get secrets/<YOUR_SECRET_NAME> -n velero -o jsonpath="{.data.cloud}" | base64 -d)
```
nrb: Looks like the `-d` flag is `-D` on macOS; dunno which one we want to default to

skriss: this is going to dump the content of the secret into the env var, right? I'm not sure that works - we may need to write the data into a local file, then set the env var to have the path of the file

docs/upgrade-to-1.0.md
```diff
@@ -42,6 +42,19 @@ _*backups created prior to v0.10 stored snapshot metadata in the `status.volumeB
...
+
+    #GCP
+    GOOGLE_APPLICATION_CREDENTIALS=$(kubectl get secrets/<YOUR_SECRET_NAME> -n velero -o jsonpath="{.data.cloud}" | base64 -d)
```
skriss: ditto here - I think the env var needs to have a file path to a local credentials file

docs/upgrade-to-1.0.md
```diff
@@ -42,6 +42,19 @@ _*backups created prior to v0.10 stored snapshot metadata in the `status.volumeB
...
+
+    # Azure
+    source $(kubectl get secrets/<YOUR_SECRET_NAME> -n velero -o jsonpath="{.data.cloud}" | base64 -d)
```
skriss: in the old (pre-1.0) way of providing Azure creds, the secret has a bunch of keys, each of which corresponds to an env var - so we'd need to read each of them and then set the appropriate local env var

nrb: I don't think we need to do them one at a time, but `source` is the incorrect command.

See this example where I cat a file that's just a list of env var key/value pairs (like the pre-1.0 Azure creds file that's given to `fromEnv`)

```bash
x1c in /home/nrb/go/src/github.com/heptio/velero (git) document-migration-secret
% cat test.env
ONE=one
TWO=two

x1c in /home/nrb/go/src/github.com/heptio/velero (git) document-migration-secret
% source $(cat test.env)
source: no such file or directory: ONE=one

x1c in /home/nrb/go/src/github.com/heptio/velero (git) document-migration-secret
% export $(cat test.env)

x1c in /home/nrb/go/src/github.com/heptio/velero (git) document-migration-secret
% echo $ONE
one

x1c in /home/nrb/go/src/github.com/heptio/velero (git) document-migration-secret
% echo $TWO
two
```

docs/upgrade-to-1.0.md
nrb: Yeah, you're right. For AWS/Google we'll need to make (and clean up) a temp file for the operation.

skriss:
I made a couple updates to this and merged in #1501, so closing.

PR #1506, opened 2019-05-20
    Only run CI when relevant files have changed

nrb:
* If nothing was modified under the site directory, don't run a netlify build
 * If nothing outside of the site directory was modified, don't run `make ci`.

### DIFF
--- a/hack/ci-check.sh
+++ b/hack/ci-check.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+# If we're doing push build, as opposed to a PR, always run make ci
+if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
+    make ci
+    # Exit script early, returning make ci's error
+    exit $?
+fi
+
+# Only run `make ci` if files outside of the site directory changed in the branch
+# In a PR build, $TRAVIS_BRANCH is the destination branch.
+if [[ $(git diff --name-only $TRAVIS_BRANCH | grep --invert-match site/) ]]; then
+  make ci
+else
+  echo "Skipping make ci since nothing outside of site directory changed."
+  exit 0
+fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,4 @@ sudo: required
 services:
   - docker
 
-script: make ci
+script: hack/ci-check.sh


nrb:
Netlify is failing because the path I'm providing in the `netlify.toml` file isn't correct for the environment. Working on correcting this.

skriss:
> 11:42:32 AM: Different base path detected, going to use the one specified in the toml file: 'site/' versus 'site' in the site

Looks like maybe the root is `site/`?

nrb:
> Looks like maybe the root is site/?

That was part of it. It's now failing because the built script returns a 0 exit code, but there's not actually anything in the publish directory to publish.

I think for now, it's probably better to keep the changes for `make ci` and leave Netlify untouched here.

hack/ci-check.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# Only run `make ci` if files outside of the site directory changed in the last commit
```
skriss: The only issue I see with this is that if you have a PR that has 2 commits, the first one containing code and the second one containing docs updates, then `make ci` won't end up getting run since the *last* commit doesn't have code changes.  Not sure off the top of my head if there's an easy way to deal with this.

nrb: Good point! I've updated it to `git diff` against the destination branch, which should mean it's applicable across all commits being added.

hack/ci-check.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+# Depending on whether or not the current build is based on a PR or not,
+# the $TRAVIS_BRANCH variable can be inaccurate.
+# For push builds, always diff against master
+BRANCH="master"
+if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
```
skriss: I'm thinking that if it's not a PR build, we should just always run `make ci` -  WDYT?  I'm just wondering about if we're building a tag, or one of the release branches for a patch, diffing against master may not make sense.  Either way, isn't this `if` backwards -- i.e. we want to diff against `$TRAVIS_BRANCH` if it *is* a pull request?

nrb: Yeah, I'd agree on those points. And you're right that it was backwards.

I'm gonna simplify this a little further - in this `if` check, do make ci, and exit early. Otherwise, do the below using `$TRAVIS_BRANCH` directly. Sound good?

skriss: Yeah, I think that makes sense

nrb:
Yep!

PR #1517, opened 2019-05-22
allow individual backup storage locations to be read-only

### LINKED ISSUE #1514: Replace --restore-only flag with read-only backup storage locations
Today, the velero server can be put in "restore-only" mode by using the `--restore-only` flag on the `velero server` command. This turns off backups & garbage collection for the entire server.  However, this pre-dates BackupStorageLocations.  It would be nice to be able to put individual backup storage locations into restore/read-only mode, so that, for example, Cluster B could restore a backup from Cluster A using a read-only BSL, while still creating backups into its own read/write BSL.

The implementation would probably involve something like the following:
- add a `Mode` field to the BSL's spec, which could be read/write, or read-only
- add CLI commands and flags to create a BSL as read-only, or to switch an existing one to read-only
- modify backup validation to disallow backups being created for read-only locations
- modify schedule validation to disallow schedules being created for read-only locations
- modify GC & backup deletion controllers to disallow deletion of backups in read-only locations

skriss:

Fixes #1514 

TODO:
- [x] test cases
- [x] documentation updates
- [x] add deprecation notice to `--restore-only` flag for `velero server` (should wait until 2.0 to drop, I think)

### DIFF
--- a/pkg/apis/velero/v1/backup_storage_location.go
+++ b/pkg/apis/velero/v1/backup_storage_location.go
@@ -66,6 +66,10 @@ type BackupStorageLocationSpec struct {
 	Config map[string]string `json:"config"`
 
 	StorageType `json:",inline"`
+
+	// ReadOnly is whether the backup storage location is in read-only mode. If true,
+	// backups cannot be created or deleted in this location.
+	ReadOnly bool `json:"readOnly"`
 }
 
 // BackupStorageLocationPhase is the lifecyle phase of a Velero BackupStorageLocation.
--- a/pkg/apis/velero/v1/backup_storage_location.go
+++ b/pkg/apis/velero/v1/backup_storage_location.go
@@ -67,9 +67,8 @@ type BackupStorageLocationSpec struct {
 
 	StorageType `json:",inline"`
 
-	// ReadOnly is whether the backup storage location is in read-only mode. If true,
-	// backups cannot be created or deleted in this location.
-	ReadOnly bool `json:"readOnly"`
+	// AccessMode defines the permissions for the backup storage location.
+	AccessMode BackupStorageLocationAccessMode `json:"accessMode,omitempty"`
 }
 
 // BackupStorageLocationPhase is the lifecyle phase of a Velero BackupStorageLocation.
@@ -94,10 +93,17 @@ const (
 	BackupStorageLocationAccessModeReadWrite BackupStorageLocationAccessMode = "ReadWrite"
 )
 
+// TODO(2.0): remove the AccessMode field from BackupStorageLocationStatus.
+
 // BackupStorageLocationStatus describes the current status of a Velero BackupStorageLocation.
 type BackupStorageLocationStatus struct {
-	Phase              BackupStorageLocationPhase      `json:"phase,omitempty"`
-	AccessMode         BackupStorageLocationAccessMode `json:"accessMode,omitempty"`
-	LastSyncedRevision types.UID                       `json:"lastSyncedRevision,omitempty"`
-	LastSyncedTime     metav1.Time                     `json:"lastSyncedTime,omitempty"`
+	Phase              BackupStorageLocationPhase `json:"phase,omitempty"`
+	LastSyncedRevision types.UID                  `json:"lastSyncedRevision,omitempty"`
+	LastSyncedTime     metav1.Time                `json:"lastSyncedTime,omitempty"`
+
+	// AccessMode is an unused field.
+	//
+	// Deprecated: there is now an AccessMode field on the Spec and this field
+	// will be removed entirely as of v2.0.
+	AccessMode BackupStorageLocationAccessMode `json:"accessMode,omitempty"`
 }
--- a/pkg/cmd/cli/backuplocation/create.go
+++ b/pkg/cmd/cli/backuplocation/create.go
@@ -59,6 +59,7 @@ type CreateOptions struct {
 	Prefix   string
 	Config   flag.Map
 	Labels   flag.Map
+	ReadOnly bool
 }
 
 func NewCreateOptions() *CreateOptions {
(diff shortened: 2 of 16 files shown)


pkg/apis/velero/v1/backup_storage_location.go
```diff
@@ -66,6 +66,10 @@ type BackupStorageLocationSpec struct {
 	Config map[string]string `json:"config"`
 
 	StorageType `json:",inline"`
+
+	// ReadOnly is whether the backup storage location is in read-only mode. If true,
+	// backups cannot be created or deleted in this location.
+	ReadOnly bool `json:"readOnly"`
```
skriss: Alternately, this could be a `Mode` field, with values of "ReadWrite", "ReadOnly", etc.

nrb: I think `Mode` is more flexible, in case we add any other states. I don't know what they would be at the moment, though. Maybe something for replication, such as `ReplicateOnly`?

skriss: Yeah, I agree on the flexibility but also couldn't think of possible other states, but probably makes sense to go that way regardless.

skriss: Oh, lol: https://github.com/heptio/velero/blob/master/pkg/apis/velero/v1/backup_storage_location.go#L83

Guess we defined this back when we added the BSL type, just never implemented it.

skriss: Hmm, this ended up as a `status` field, though, not sure that makes sense.

carlisia:
This looks sensible. 

I think disallowing schedule for read-onlys is also missing?

skriss:
> I think disallowing schedule for read-onlys is missing?

I looked at the code there, and the schedule controller itself doesn't do any validation of the backup spec that's provided to it -- it defers that to the backup controller when each individual backup is triggered. So to remain consistent with that, I decided not to add any specific validation. When the schedule controller creates a backup, the backup controller will run its validation and fail the backup.

skriss:
OK, I still need to do some more manual testing, but this is ready for a full review @carlisia @nrb.

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -53,17 +54,23 @@ func NewCreateCommand(f client.Factory, use string) *cobra.Command {
...
+	Labels     flag.Map
+	AccessMode *flag.Enum
 }
 
 func NewCreateOptions() *CreateOptions {
 	return &CreateOptions{
 		Config: flag.NewMap(),
+		AccessMode: flag.NewEnum(
+			string(velerov1api.BackupStorageLocationAccessModeReadWrite),
+			string(velerov1api.BackupStorageLocationAccessModeReadWrite),
```
carlisia: This is a dup of the prior one.

site/docs/master/about.md
```diff
@@ -31,7 +31,7 @@ The **restore** operation allows you to restore all of the objects and persisten
 
 The default name of a restore is `<BACKUP NAME>-<TIMESTAMP>`, where `<TIMESTAMP>` is formatted as *YYYYMMDDhhmmss*. You can also specify a custom name. A restored object also includes a label with key `velero.io/restore-name` and value `<RESTORE NAME>`.
 
-You can also run the Velero server in restore-only mode, which disables backup, schedule, and garbage collection functionality during disaster recovery.
+You can also configure backup storage locations in read-only mode, which disables backup creation and deletion for the storage location during disaster recovery.
```
carlisia: To be extra clear, I'd say: You can configure... in read-write and read-only modes. In the read-only mode...

I'd also `s/disaster recovery/restore` at the end of this sentence.

site/docs/master/disaster-case.md
```diff
@@ -13,7 +13,14 @@ If you periodically back up your cluster's resources, you are able to return to
 
 1.  A disaster happens and you need to recreate your resources.
 
-1.  Update the Velero server deployment, adding the argument for the `server` command flag `restore-only` set to `true`. This prevents Backup objects from being created or deleted during your Restore process.
+1.  Update your backup storage location to be in read-only mode. This prevents Backup objects from being created or deleted in the backup storage location during your Restore process:
```
carlisia: `s/prevents Backup/prevents backup`

carlisia: and `s/your Restore/your restore`

pkg/cmd/cli/backuplocation/create.go
skriss: the first arg is the default value, then you provide a list of all the possible options, so it needs to be specified twice.

site/docs/master/about.md
skriss: addressed.

site/docs/master/disaster-case.md
skriss: addressed.

skriss:
rebased, @nrb @carlisia ready for review (and if merging, pls squash).

site/docs/master/disaster-case.md
```diff
@@ -14,10 +14,26 @@ If you periodically back up your cluster's resources, you are able to return to
 
 1.  A disaster happens and you need to recreate your resources.
 
-1.  Update the Velero server deployment, adding the argument for the `server` command flag `restore-only` set to `true`. This prevents Backup objects from being created or deleted during your Restore process.
+1.  Update your backup storage location to read-only mode (this prevents backup objects from being created or deleted in the backup storage location during the restore process):
+
+    ```bash
```
nrb: Would it be useful to include our own command to do these operations later? Something like `velero backup-locations edit mylocation --read-only/--read-write`? I'd think we wouldn't need a controller for it, and it's a little less for users to type.

skriss: Yeah, agree it would be nice to have.  We don't really have any `edit` commands right now so I'd want to think a little about what the UX should be.

PR #1535, opened 2019-06-03
Add cmd to list plugins

carlisia:
Closes #1474.

Highlights:
- outputting to yaml and json also works
- list sorted alphabetically by plugin name
- made code in the now named `cmd/serverstatus` package reusable (exported)
- I deleted a test because it seems to be one of those that is testing the language. I can put it back if not.

Screenshots:
image

image

image

image


pkg/apis/velero/v1/server_status_request.go
```diff
@@ -56,6 +58,9 @@ type ServerStatusRequestStatus struct {
 
 	// ServerVersion is the Velero server version.
 	ServerVersion string `json:"serverVersion"`
+
+	// Plugins list information about the plugins running on the Velero server
+	Plugins [][]string `json:"plugins"`
```
skriss: Instead of having nested slices here, I would define a new struct to capture the plugin info, something along the lines of:

```go
type PluginInfo struct {
    Name string // TODO json tags
    Kind string // TODO json tags
}
```

And then it becomes:

```go
Plugins []PluginInfo
```

pkg/cmd/cli/plugin/get.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2017 the Velero contributors.
```
skriss: 2019

pkg/cmd/serverstatus/server_status.go
```diff
@@ -0,0 +1,78 @@
+package serverstatus
```
skriss: needs license header

skriss: I think this should probably go under `pkg/cmd/cli/serverstatus` - WDYT?

pkg/cmd/serverstatus/version.go
```diff
@@ -0,0 +1,80 @@
+/*
+Copyright 2017 the Velero contributors.
```
skriss: 2019

pkg/cmd/serverstatus/version.go
```diff
@@ -0,0 +1,80 @@
+/*
```
skriss: I would put this in `pkg/cmd/cli/version`.  I think it belongs under `cli` (I see that currently the version code is directly under `pkg/cmd`, but to me that doesn't make sense, as it's part of the CLI), and I would keep the package names aligned to command names.

pkg/cmd/serverstatus/version_test.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
skriss: I would put this under `pkg/cmd/cli/version` (see previous comment).

pkg/cmd/util/output/output.go
```diff
@@ -43,6 +43,11 @@ func BindFlags(flags *pflag.FlagSet) {
 	flags.Bool("show-labels", false, "show labels in the last column")
 }
 
+// BindFlagsSimple defines the output flag within the provided FlagSet.
```
skriss: Could you update the godoc to disambiguate this from `BindFlags`?

pkg/cmd/util/output/plugin_printer.go
```diff
@@ -0,0 +1,60 @@
+/*
+Copyright 2017, 2019 the Velero contributors.
```
skriss: Just 2019, right?

pkg/plugin/framework/plugin_kinds_test.go
```diff
@@ -1,34 +0,0 @@
-/*
```
skriss: Yeah, I don't think this was adding much value. 👍 to deleting.

pkg/serverstatusrequest/process_test.go
```diff
@@ -105,8 +107,12 @@ func TestProcess(t *testing.T) {
 	for _, tc := range tests {
```
skriss: let's add some test cases

skriss:
One other thing - I'm wondering if it would be clearer to show the output sorted first by plugin kind, then by name - WDYT?

pkg/cmd/serverstatus/version.go
carlisia: I agree with that too.

carlisia:
Argh, header fails!

+1 for comments, will address.

> One other thing - I'm wondering if it would be clearer to show the output sorted first by plugin kind, then by name - WDYT?

Yeah, I thought about this. I'm def not a heavy user but my rational was: If I'm using `aws`, I want to see what plugins types there are for that provider. Or if I added custom plugins, I don't want to scan the entire list to find all of them. I think this would be more useful then, "If I want an `objectstore` type of plugin, what are the ones available?" I can be persuaded to go with kind, just wanted to run this by you. Let me know.

carlisia:
I'll also add a changelog.

skriss:
> > One other thing - I'm wondering if it would be clearer to show the output sorted first by plugin kind, then by name - WDYT?

> Yeah, I thought about this. I'm def not a heavy user but my rational was: If I'm using aws, I want to see what plugins types there are for that provider. Or if I added custom plugins, I don't want to scan the entire list to find all of them. I think this would be more useful then, "If I want an objectstore type of plugin, what are the ones available?" I can be persuaded to go with kind, just wanted to run this by you. Let me know.

Let's see what @nrb thinks.  It would be nice if we had a `--sort-by` flag like `kubectl`, but we don't.

nrb:
>  It would be nice if we had a --sort-by flag like kubectl, but we don't.

Agreed here. Honestly, this seems like a feature worth adding itself, but not in this PR.

I can see the list of plugins getting pretty long if a lot of item action plugins are installed. I'm leaning towards sorting by kind then by name within each kind, as well as adding a line break between each grouping.

There are definitely going to be times when users are looking for different information and the different sorts would be useful, but I think this strikes a reasonable balance for a default. Plus, the user can pipe the list into `grep` if they're looking for something specific.

Longer term I'd be all for adding a `--sort-by` flag to help with this, though.

carlisia:
https://github.com/heptio/velero/pull/1535/files#diff-8067a89e94388cc2f5a9a40b5107e9eaR126

The test above is failing because I'm not sure how I would inject some fake plugins into the registry directory. Any hint? 

One way but the `fs` element is private.
registry.fs = test.NewFakeFileSystem().
				WithFileAndMode("/plugins/velero.io", []byte("ObjectStore"), 0755).
				WithFileAndMode("/plugins/custom.io", []byte("VolumeSnapshotter"), 0755)

pkg/serverstatusrequest/process.go
```diff
@@ -29,20 +29,23 @@ import (
...
+	"github.com/heptio/velero/pkg/plugin/clientmgmt"
+	"github.com/heptio/velero/pkg/plugin/framework"
 )
 
 const ttl = time.Minute
 
 // Process fills out new ServerStatusRequest objects and deletes processed ones
 // that have expired.
-func Process(req *velerov1api.ServerStatusRequest, client velerov1client.ServerStatusRequestsGetter, clock clock.Clock, log logrus.FieldLogger) error {
+func Process(req *velerov1api.ServerStatusRequest, client velerov1client.ServerStatusRequestsGetter, pluginRegistry clientmgmt.Registry, clock clock.Clock, log logrus.FieldLogger) error {
```
skriss: For testing purposes, I would probably do the following:

1. Define a new interface in this package, along the lines of:
```go
type PluginLister interface {
  List(...) // TODO fill in parameters and return types to match plugin.Registry
}
```

2. Switch the `pluginRegistry` parameter to this function to be `pluginLister PluginLister`

3. For testing, implement a simple `fakePluginLister` struct that implements the `PluginLister` interface and returns whatever you tell it to, and pass that as the argument to `Process(...)`.

I'm still thinking about some ideas for making it easier to unit test code that relies on plugins, but for now I think this is a reasonable approach.

carlisia: Great, thank you!

pkg/serverstatusrequest/process.go
```diff
@@ -29,23 +29,31 @@ import (
 	velerov1api "github.com/heptio/velero/pkg/apis/velero/v1"
 	"github.com/heptio/velero/pkg/buildinfo"
 	velerov1client "github.com/heptio/velero/pkg/generated/clientset/versioned/typed/velero/v1"
-	"github.com/heptio/velero/pkg/plugin/clientmgmt"
 	"github.com/heptio/velero/pkg/plugin/framework"
 )
 
 const ttl = time.Minute
 
+type PluginLister interface {
```
skriss: We should only need the `List` func here - any particular reason why you included `DiscoverPlugins` and `Get` also?

pkg/serverstatusrequest/process_test.go
```diff
@@ -142,3 +170,26 @@ func TestProcess(t *testing.T) {
 		})
 	}
 }
+
+type fakePluginLister struct {
+	plugins []framework.PluginIdentifier
+}
+
+func (l *fakePluginLister) DiscoverPlugins() error {
```
skriss: Per prior comment, this method isn't really needed.

pkg/serverstatusrequest/process_test.go
```diff
@@ -142,3 +170,26 @@ func TestProcess(t *testing.T) {
...
+	for _, plugin := range l.plugins {
+		if plugin.Kind == kind {
+			plugins = append(plugins, plugin)
+		}
+	}
+
+	return plugins
+}
+
+func (l *fakePluginLister) Get(kind framework.PluginKind, name string) (framework.PluginIdentifier, error) {
```
skriss: Per prior comment, this method isn't really needed.

pkg/serverstatusrequest/process.go
carlisia: Because this and `Get` are part of the `PluginLister` interface. The compiler fails w/o them.

carlisia: Ohhh NVM! Fixing.

skriss:
@nrb ready for squash/merge if you are.

Issue #1169, opened 2019-01-24
store PodVolumeBackups in obj storage & sync into cluster


skriss:
Currently, `PodVolumeBackups` (representing a restic backup) are not stored in object storage as JSON. This causes a couple of problems:
- if a backup is taken in Cluster A, and then Cluster B is set up to restore from Cluster A's bucket, running an `ark backup describe` on a backup in Cluster B won't show the restic backups, even if they exist
- there's a second source of truth for mapping pods & volumes to restic snapshots: an annotation on the pod within the backup tarball. This is what's actually used during the restore process to determine if there's a restic backup to restore

`PodVolumeBackups` should be serialized and stored in object storage, and synced into the cluster through the backup sync controller. They should also be used as the source of truth during the restore process. The pod annotation that's stored in the tarball should be deprecated and eventually removed.

carlisia:
### Design considerations

**Goals** 

- Deprecate the annotation on the pod within the backup tarball (eventually remove) 
- Serialize the PodVolumeBackups and store it in object storage like the regular backups 
- Sync to the cluster through the backup sync controller 
- Modify the code that relies on that annotation to map pods and volumes to restic snapshots/Modify restore to use this source, but maintain existing code so we can restore “old” backups
- Implementation needs to be backwards compatible with 1.0 ability for restoring backups  
- When storing in object storage, the `PodVolumeBackups` will go in the same JSON file as the regular backups

carlisia:
@skriss / @nrb I think this ^ covers all of the design considerations for this change. Please check. 

I don't see any major tradeoffs between creating a separate file for this or not, except that if there was a separate file, there would be that additional file to open when there are `PodVolumeBackups`. I think the expectation is that weather it is restic of a regular PV, it would all be found in the same place.

skriss:
This all looks reasonable.  Re: how to store the `PodVolumeBackups`:
- we almost definitely don't want to store them *inside* the backup tarball. If we were to do this,  it would mean the backup sync controller would have to extract each tarball to be able to sync backups from object storage into the cluster.
- we *could* modify the existing <backup-name>-volumesnapshots.json.gz file to store both regular PV snapshots *and* PodVolumeBackups, although I'm not necessarily sure it's worth overloading.
- we could add a new file, something like <backup-name>-podvolumebackups.json.gz, that specifically stores a serialized list of PodVolumeBackups.

carlisia:
Ah. Thanks for the explanation, it's super helpful.

[...]

PR #1538, opened 2019-06-04
[WIP] properly restore Retain PVs with restic backups

### LINKED ISSUE #1151: Improve handling of PersistentVolumes with Retain reclaim policy
PersistentVolumes can have a reclaim policy of either Delete or Retain.

When a PVC is deleted, if the PV's reclaim policy is Delete, the underlying storage disk is deleted and the PV is deleted.

When a PVC is deleted, if the PV's reclaim policy is Retain, the underlying storage disk is preserved, and the PV remains.

Ark works quite well with Delete:
- if we have a native snapshot, we create a new storage disk from the snapshot, then associate it with the PV
- if we have a restic snapshot, we don't restore the PV, and instead only restore the PVC, which results in a new PV being created, and then we restic restore into it

Ark doesn't work as well with Retain, depending on your restore scenario:
- if you've "lost" your cluster or a portion of it, but you still have the underlying storage disk, and you're restoring into the same project/region/zone (if in the cloud), then we restore the PV and keep its association to the original underlying storage disk, so you get back to where you were pre-disaster
- if you're trying to restore into a new cluster and you don't want to reuse the underlying storage disk, restoring the PV and preserving the association to the underlying storage disk is a problem

We should try to figure out how to handle the 80-90% case as well as possible. Maybe we need a new flag when creating the restore that indicates how to handle Retain PVs - to either keep the current behavior of preserving the association to the underlying disk, or to wipe the association and restore from a snapshot (if we have one) or create a new disk (assuming we have a restic snapshot).

Discuss!

skriss:

Fixes #1151 

@nrb @carlisia wanted to get your feedback on this approach.  To distill the issue - Velero today will not restore PVs that have a reclaim policy of Retain, and a restic backup.  It does properly restore Retain PVs from snapshots if they exist, so I consider this to just be a bug related to how the code is structured (there's some other special-case handling for Retain PVs that *don't* have snapshots, that pre-dates restic, and so inadvertently now catches the case where there's no snapshot, but there is a restic backup).

The main challenge in fixing this is that during restore time, we need to figure out if a given PV has a restic backup or not, but this information is stored as an annotation on the Pod that mounts the PVC that claims the PV, and this chain of relationships can't be traversed starting from the PV.

This PR puts an annotation on the relevant PVC during backup time, indicating that it has a restic backup, which is then looked for during a restore to determine whether or not there's a restic backup.  Since we *can* relatively easily go from a PV to the PVC that claims it, it's relatively easy to look this up during a restore.

This would all be easier if we did a multi-pass restore (we could collect this info in the "prepare" phase and then use it during the "restore" phase).

### DIFF
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -39,6 +39,7 @@ import (
 	"github.com/heptio/velero/pkg/plugin/velero"
 	"github.com/heptio/velero/pkg/podexec"
 	"github.com/heptio/velero/pkg/restic"
+	"github.com/heptio/velero/pkg/util/boolptr"
 	"github.com/heptio/velero/pkg/volume"
 )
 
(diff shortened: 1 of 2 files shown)


carlisia:
Will this have to change once we get https://github.com/heptio/velero/issues/1169 done?

skriss:
You're right, once we do #1169 we'd probably want to change our solution for this (we wouldn't strictly *have* to, but it would probably make sense to).  If we're going to tackle #1169 relatively soon then I'm fine holding off on fixing this one until after then.

skriss:
I'm closing this as we'll probably modify our approach post-#1169.

PR #1547, opened 2019-06-05
fixes inlining of command in GCP setup docs

prydonius:
Currently this command gets inlined in the docs site, when it should appear in its own block.

Screen Shot 2019-06-05 at 16 27 43

Let me know if this is not the right place to make this fix, or if I need to replicate the fix in the other version folders?


carlisia:
Good catch!

Yes, please do the same for the version in master, and scan the folders for the other versions as well.

skriss:
I'd say making the fix for `master/` and `v1.0.0/` is sufficient - going forward there's no real need to update pre-1.0 docs, we're just keeping them around for reference.

prydonius:
thanks @carlisia @skriss - updated to fix the master folder too!

Issue #1064, opened 2018-11-17
Support for multiple different provider credentials


jaxxstorm:
**Describe the solution you'd like**
In v0.10.0, the ability to specify multiple backup locations we created. This is an excellent step forward for redundancy, however in practice it's only possible to backup to multiple locations within the same provider.

If you want to backup to say, minio and AWS S3, you'll more than likely have different API keys for each bucket in the different provider. As far as I can tell (unless I'm missing something) only a single set of credentials can be provided to a deployment.

This can be solved by running multiple deployments in different namespaces, but as the number grows this might become unwieldy. It would be good if the keys for a specific provider could be moved into the BackupStorageProvider itself, rather than in the deployment

**Environment:**

- Ark version (use `ark version`): v0.10.0
- Kubernetes version (use `kubectl version`): v1.12.0
- Kubernetes installer & version: digitalocean

jenting:
@zubron I guess this issue could be closed?

zubron:
Ooh, I wasn't aware of this issue. Thanks, @jenting! Yes, this can be closed now as it was handled in #3304.

PR #1548, opened 2019-06-06
Support for multiple AWS profiles

pranavgaikwad:
Currently, velero does not support different credentials for different S3 buckets. The S3 connection established uses `default` profile. AWS SDK supports multiple profiles to be used. This change adds support for multiple profiles by introducing a new config variable in backup storage location definition. When a profile is not set explicitly by the user, AWS SDK automatically selects the `default` profile making this change not break existing configurations.

### DIFF
--- a/site/docs/master/api-types/backupstoragelocation.md
+++ b/site/docs/master/api-types/backupstoragelocation.md
@@ -20,6 +20,7 @@ spec:
     bucket: myBucket
   config:
     region: us-west-2
+    profile: "default"
 ```
 
 ### Parameter Reference
@@ -36,6 +37,7 @@ The configurable parameters are as follows:
 | `objectStorage/prefix` | String | Optional Field | The directory inside a storage bucket where backups are to be uploaded. |
 | `config` | map[string]string<br><br>(See the corresponding [AWS][0], [GCP][1], and [Azure][2]-specific configs or your provider's documentation.) | None (Optional) | Configuration keys/values to be passed to the cloud provider for backup storage. |
 
+
 #### AWS
 
 **(Or other S3-compatible storage)**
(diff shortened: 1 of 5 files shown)


dymurray:
Pranav found this solution in reference to the issue I created awhile back: https://github.com/heptio/velero/issues/1368. We are aware it does not completely solve the issue that mine was a duplicate from here: https://github.com/heptio/velero/issues/1064 but it does allow someone using S3 to specify different profiles in a BSL. Please let us know if you have any concerns or feel this should be broadened to solve #1064.

skriss:
This looks pretty reasonable to me. The one thing we do need to do is add code to use "default" as the profile if no profile is specified, for backwards compatibility. Thanks for the PR!

pranavgaikwad:
@skriss This change is already backward compatible. When the config variable is not present, the profile is simply passed as an empty string. AWS SDK then loads it from the credentials file. Here's relevant doc - https://docs.aws.amazon.com/sdk-for-go/v1/api/aws.session.Options.html

skriss:
Ah, thanks, didn't realize that.

site/docs/master/api-types/backupstoragelocation.md
```diff
@@ -50,6 +52,7 @@ The configurable parameters are as follows:
 | `publicUrl` | string | Empty | *Example*: https://minio.mycluster.com<br><br>If specified, use this instead of `s3Url` when generating download URLs (e.g., for logs). This field is primarily for local storage services like Minio.|
 | `kmsKeyId` | string | Empty | *Example*: "502b409c-4da1-419f-a16e-eif453b3i49f" or "alias/`<KMS-Key-Alias-Name>`"<br><br>Specify an [AWS KMS key][10] id or alias to enable encryption of the backups stored in S3. Only works with AWS S3 and may require explicitly granting key usage rights.|
 | `signatureVersion` | string | `"4"` | Version of the signature algorithm used to create signed URLs that are used by velero cli to download backups or fetch logs. Possible versions are "1" and "4". Usually the default version 4 is correct, but some S3-compatible providers like Quobyte only support version 1.|
+| `profile` | string | "default" | AWS profile within the credential file to use for given store |
```
skriss: Could you make a similar entry on the `volumesnaphotlocation.md` page, since it applies there as well?

skriss:
@nrb @carlisia @prydonius please take a look as well.

site/docs/master/api-types/backupstoragelocation.md
pranavgaikwad: Thank you for pointing that out. I just updated the doc.

Issue #1522, opened 2019-05-25
Blog author link is broken


carlisia:
![image](https://user-images.githubusercontent.com/16508/58372506-ffb82180-7ed2-11e9-9690-1d86861fabd3.png)

Clicking on the name of the author gives a 404.

coonsd:
I added the author as a tag on the blog post, which should fix the 404 error.  I forgot to do the sign-off on my first pull request, so I closed it and resubmitted it with it correctly included.

skriss:
@carlisia @coonsd is this issue now resolved? If so, please close this issue (and thanks @coonsd for the PR!)

carlisia:
Ah there wasn't a reference. Closing. Fixed by: https://github.com/heptio/velero/pull/1550

skriss:
OK, closing.

PR #1550, opened 2019-06-06
Fixed 404 Link in Blog - Issue #1522

coonsd:
#1522 : Added the author's name as a tag. This should fix the broken link.


carlisia:
I verified that this works well:
image

Thank you so much for this contribution!

PR #1559, opened 2019-06-10
remove glog

skriss:
Kubernetes no longer uses glog - instead it uses the klog fork (https://github.com/kubernetes/klog).  This removes all glog references from our codebase, updates with klog references as appropriate, and drops glog from our deps.


prydonius:
Just out of curiosity, when do we use glog/klog vs. logrus?

skriss:
Velero itself only uses logrus, not glog/klog, but we do expose the flags for controlling klog, so that if a users wants to get logs from client-go/apimachinery, it's possible.

PR #1561, opened 2019-06-10
hide server commands in velero CLI

### LINKED ISSUE #1527: Remove the server cmd from the velero/restic CLI
image

The `velero restic server` has no utility for the Velero CLI user. While it doesn't need to be removed from the list of commands, it should be removed from the help text so users don't think there is use for it.

Update: also remove `velero server`, for the same reason.

prydonius:
Hides `velero server` and `velero restic server` commands from the list of available commands as these are not intended for use by the velero CLI user.

fixes #1527 

### DIFF
--- a/changelogs/unreleased/1561-prydonius
+++ b/changelogs/unreleased/1561-prydonius
@@ -1,0 +1,1 @@
+Hides `velero server` and `velero restic server` commands from the list of available commands as these are not intended for use by the velero CLI user.
--- a/pkg/cmd/cli/restic/server.go
+++ b/pkg/cmd/cli/restic/server.go
@@ -46,10 +46,9 @@ func NewServerCommand(f client.Factory) *cobra.Command {
 	logLevelFlag := logging.LogLevelFlag(logrus.InfoLevel)
 
 	command := &cobra.Command{
-		Use:   "server",
-		Short: "Run the velero restic server",
-		Long:  "Run the velero restic server",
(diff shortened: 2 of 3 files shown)


prydonius:
@carlisia :+1:, removed the comments

not sure why this particular test is failing: https://travis-ci.org/heptio/velero/builds/543954241

carlisia:
Hm. I added that test, let me try it locally.

prydonius:
@carlisia looks like the problem is the order of the PluginInfo array 

It expected the custom plugin to be last:
```
Plugins:[]v1.PluginInfo{v1.PluginInfo{Name:"velero.io/aws", Kind:"ObjectStore"}, v1.PluginInfo{Name:"custom.io/myown", Kind:"VolumeSnapshotter"}}}}
```

but the actual order is switched:

```
Plugins:[]v1.PluginInfo{v1.PluginInfo{Name:"custom.io/myown", Kind:"VolumeSnapshotter"}, v1.PluginInfo{Name:"velero.io/aws", Kind:"ObjectStore"}}}}
```

skriss:
Yeah, sounds like we need to sort the slices before comparing.

carlisia:
Yes,  my bad I didn't remember that slices don't keep the order and that `ServerStatusRequest` contained that slice. `

carlisia:
Fixing.

skriss:
Merging this one since it's fine as-is and the unrelated test flake is being addressed in a separate PR.

PR #1562, opened 2019-06-10
Fix test comparison

carlisia:
I think this will fix it (see https://travis-ci.org/heptio/velero/builds/543954241). I cleared the test cache and ran this test a buncha times.

This is making me think though: what if...

1) I moved the `sortPluginsByKindAndName` function to the `process.go` file, used it to build the list of sorted plugins (by name and kind) here: https://github.com/heptio/velero/blob/e183c4b5978ce4f1d68e78acb0343e1eec9f6ee1/pkg/serverstatusrequest/process.go#L110

2) Removed this function (since the list is being built in order): https://github.com/heptio/velero/blob/e183c4b5978ce4f1d68e78acb0343e1eec9f6ee1/pkg/cmd/util/output/plugin_printer.go#L35

3) Used the `process.go/sortPluginsByKindAndName` to sort (only) the test slice.

I think this would work?

### DIFF
--- a/pkg/serverstatusrequest/process_test.go
+++ b/pkg/serverstatusrequest/process_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package serverstatusrequest
 
 import (
+	"sort"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


skriss:
Hmm. IMHO it's fine as-is -- I'd rather keep the sorting logic for display purposes in the CLI layer, and I'm not bothered by repeating a couple of lines of sort code.

One note - the reason the ordering is unstable is because `AllPluginKinds()` (https://github.com/heptio/velero/blob/master/pkg/plugin/framework/plugin_kinds.go#L47) returns a map, which is being iterated over here: https://github.com/heptio/velero/blob/master/pkg/serverstatusrequest/process.go#L101. Maps aren't ordered, so the kinds may come back in random order, which has the effect of ~randomizing the order of the `.Plugins` field on the API obj.

carlisia:
> Maps aren't ordered, so the kinds may come back in random order, which has the effect of ~randomizing the order of the .Plugins field on the API obj.

I really appreciate this correction! 🙏

carlisia:
I also don't mind at all repeating code, and keeping the formatting of the display code in the layer where it naturally belongs is a good call. So yes, let's keep it as is.

PR #1574, opened 2019-06-12
migrate backup PV snapshot tests

skriss:
In particular, please look at the `fakeVolumeSnapshotter` struct, which is important for these tests. What can I do to improve the clarity/usability of it?

### DIFF
--- a/pkg/backup/backup_new_test.go
+++ b/pkg/backup/backup_new_test.go
@@ -43,7 +43,6 @@ import (
 	dynamicfake "k8s.io/client-go/dynamic/fake"
 	kubefake "k8s.io/client-go/kubernetes/fake"
 
-	v1 "github.com/heptio/velero/pkg/apis/velero/v1"
 	velerov1 "github.com/heptio/velero/pkg/apis/velero/v1"
 	"github.com/heptio/velero/pkg/client"
 	"github.com/heptio/velero/pkg/discovery"
@@ -52,6 +51,7 @@ import (
 	"github.com/heptio/velero/pkg/plugin/velero"
 	"github.com/heptio/velero/pkg/test"
 	kubeutil "github.com/heptio/velero/pkg/util/kube"
+	"github.com/heptio/velero/pkg/volume"
 )
 
 // TestBackupResourceFiltering runs backups with different combinations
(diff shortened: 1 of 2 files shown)


pkg/backup/backup_new_test.go
```diff
@@ -1178,6 +1254,382 @@ func TestBackupActionAdditionalItems(t *testing.T) {
...
+		iops:        int64Ptr(iops),
+		snapshotErr: snapshotErr,
+	}
+
+	return vs
+}
+
+// Init is a no-op.
+func (vs *fakeVolumeSnapshotter) Init(config map[string]string) error {
+	return nil
```
carlisia: Whenever `vs` is not being used inside the function, it can be omitted. Makes it easier to know it's not used. The functions that call panic down below can also have this omitted. Not a big deal but I think we should aim to do this for the code base.

skriss: done

carlisia:
\o| great!

pkg/backup/backup_new_test.go
```diff
@@ -1254,6 +1254,382 @@ func TestBackupActionAdditionalItems(t *testing.T) {
...
+	if vs.PVVolumeNames == nil {
+		vs.PVVolumeNames = make(map[string]string)
+	}
+	vs.PVVolumeNames[pvName] = id
+
+	if vs.Volumes == nil {
+		vs.Volumes = make(map[volumeIdentifier]*volumeInfo)
+	}
+
+	identifier := volumeIdentifier{
```
prydonius: would it be simpler if volumeIdentifier was just a string concatenation of volumeID and volumeAZ? It doesn't look like we're accessing the volumeID or volumeAZ values from the volumeIdentifier struct anywhere (though I'm probably missing something)

skriss: That is an option, yeah. I often use structs instead of string concats in this kind of scenario to avoid the general issue where, for example, "a" + "bc" gives the same result as "ab" + "c". Shouldn't be a big issue here with a reasonable separator in the middle, e.g. `||`.

prydonius: Good point, yeah a separator would help there. Just a thought though, the struct is fine, just thought it might save a few lines over having to initialise the struct.

PR #1577, opened 2019-06-12
Store PodVolumeBackups in obj storage & use as source of truth

### LINKED ISSUE #1169: store PodVolumeBackups in obj storage & sync into cluster

carlisia:
- [x] Deprecate the annotation on the pod within the backup tarball (eventually remove)

- [x] Serialize the `PodVolumeBackups` as `<backup-name>-podvolumebackups.json.gz` that stores a serialized list of PodVolumeBackups and store it in object storage

- [x] Sync to the cluster through the backup sync controller

- [ ] Modify the code that relies on original annotation to map pods and volumes to restic snapshots/Modify restore to use this source, but maintain existing code so we can restore “old” backups

- [ ] Implementation needs to be backwards compatible with 1.0 ability for restoring backups

Fixes #1169.


pkg/restic/common.go
```diff
@@ -39,13 +39,18 @@ const (
 	InitContainer               = "restic-wait"
 	DefaultMaintenanceFrequency = 24 * time.Hour
 
-	podAnnotationPrefix       = "snapshot.velero.io/"
+	// Deprecated.
+	podAnnotationPrefix = "snapshot.velero.io/"
+
+	// Deprecated.
```
skriss: We're not deprecating this one, at least not as part of this PR.  This one indicates which volumes the user **wants** to back up.

site/docs/master/restic.md
```diff
@@ -84,47 +84,6 @@ You're now ready to use Velero with restic.
 
 ## Back up
 
-1. Run the following for each pod that contains a volume to back up:
```
skriss: This shouldn't be removed - this is still the way to indicate which volumes the user **wants** to back up.

site/docs/master/restic.md
```diff
@@ -259,21 +218,15 @@ and `restic prune`.
 
     You can see information about your Velero restic repositories by running `velero restic repo get`.
 
-- `PodVolumeBackup` - represents a restic backup of a volume in a pod. The main Velero backup process creates
-one or more of these when it finds an annotated pod. Each node in the cluster runs a controller for this
-resource (in a daemonset) that handles the `PodVolumeBackups` for pods on that node. The controller executes
-`restic backup` commands to backup pod volume data.
+- `PodVolumeBackup` - represents a restic backup of any volume in a pod. The main Velero backup process creates
```
skriss: I don't think this needs to be edited, given that we're still using pod annotations to indicate the desire to back up a specific pod volume.

site/docs/master/restic.md
```diff
@@ -259,21 +218,15 @@ and `restic prune`.
...
+one or more of these. Each node in the cluster runs a controller for this resource (in a daemonset) that handles the `PodVolumeBackups` for pods on that node. The controller executes `restic backup` commands to backup pod volume data.
 
-- `PodVolumeRestore` - represents a restic restore of a pod volume. The main Velero restore process creates one
-or more of these when it encounters a pod that has associated restic backups. Each node in the cluster runs a
-controller for this resource (in the same daemonset as above) that handles the `PodVolumeRestores` for pods
-on that node. The controller executes `restic restore` commands to restore pod volume data.
+- `PodVolumeRestore` - represents a restic restore of any pod volume. The main Velero restore process creates one
```
skriss: Don't think this needs to be edited either.

site/docs/master/restic.md
```diff
@@ -259,21 +218,15 @@ and `restic prune`.
...
+or more of these. Each node in the cluster runs a controller for this resource (in the same daemonset as above) that handles the `PodVolumeRestores` for pods on that node. The controller executes `restic restore` commands to restore pod volume data.
 
 ### Backup
 
-1. The main Velero backup process checks each pod that it's backing up for the annotation specifying a restic backup
-should be taken (`backup.velero.io/backup-volumes`)
-1. When found, Velero first ensures a restic repository exists for the pod's namespace, by:
+1. Velero first ensures a restic repository exists for the pod's namespace, by:
```
skriss: Shouldn't need to change

site/docs/master/restic.md
```diff
@@ -283,14 +236,12 @@ should be taken (`backup.velero.io/backup-volumes`)
     - finds the pod volume's subdirectory within the above volume
     - runs `restic backup`
     - updates the status of the custom resource to `Completed` or `Failed`
-1. As each `PodVolumeBackup` finishes, the main Velero process captures its restic snapshot ID and adds it as an annotation
-to the copy of the pod JSON that's stored in the Velero backup. This will be used for restores, as seen in the next section.
+1. As each `PodVolumeBackup` finishes, the main Velero process adds it to the Velero backup in a file named `<backup-name>-podvolumebackups.json.gz`. This will be used for restores, as seen in the next section.
```
skriss: 👍 

Might clarify that this new file gets uploaded to object storage alongside the backup tarball, but it's not part of the tarball itself.

site/docs/master/restic.md
```diff
@@ -283,14 +236,12 @@ should be taken (`backup.velero.io/backup-volumes`)
...
 
 ### Restore
 
-1. The main Velero restore process checks each pod that it's restoring for annotations specifying a restic backup
-exists for a volume in the pod (`snapshot.velero.io/<volume-name>`)
-1. When found, Velero first ensures a restic repository exists for the pod's namespace, by:
+1. The main Velero restore process checks each pod that it's restoring from a previously created restic backup by iterating through the content of the `<backup-name>-podvolumebackups.json.gz` file.
```
skriss: 👍

pkg/restic/common.go
```diff
@@ -39,25 +39,16 @@ const (
...
-			return true
-		}
-	}
-
-	return false
-}
-
 // GetPodSnapshotAnnotations returns a map, of volume name -> snapshot id,
 // of all restic snapshots for this pod.
+// Deprecated: we will stop using pod annotation to determine restic
```
skriss: This comment isn't accurate - should say that we're no longer using pod annotations to record restic snapshot IDs after they're taken.

pkg/restic/common.go
```diff
@@ -79,6 +70,8 @@ func GetPodSnapshotAnnotations(obj metav1.Object) map[string]string {
 
 // SetPodSnapshotAnnotation adds an annotation to a pod to indicate that
 // the specified volume has a restic snapshot with the provided id.
+// Deprecated: we will stop using pod annotation to determine restic
```
skriss: This comment isn't accurate - should say that we're no longer using pod annotations to record restic snapshot IDs after they're taken.

skriss:
:+1: everything so far looks fine now. 

I would definitely refer to the code for the `<backup-name>-volumesnapshots.json.gz` file as you go through the rest of this - while this won't be identical, it will be similar and tracing how that file gets created/saved/restored from should give you a pretty good idea of what code needs to change.

carlisia:
I super appreciate the links, thank you! 🙏

carlisia:
image

The content of the json file looks exactly like the podvolumebackups CR.

carlisia:
@skriss, @nrb question: do we want to add any podvolumebackup info to the `velero backup describe` output? If so, what? I'm thinking useful options could be number of `PodVolumeBackup` items, as in:

`Pod Volume Backups: 1`

skriss:
Per slack we're already showing it, just labeled as "restic backups"

pkg/controller/backup_controller.go
```diff
@@ -527,6 +531,34 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
 		backup.Status.Phase = velerov1api.BackupPhaseCompleted
 	}
 
+	// Find and add the PodVolumeBackup after the backup is processed
```
skriss: Instead of using the lister here to get the related pod volume backups, I think it would make more sense to modify https://github.com/heptio/velero/blob/master/pkg/restic/backupper.go#L39 to return a slice of PodVolumeBackups that were created, replacing the `map[string]string` that's currently being returned.  Then, the `item_backupper` can just grab them and assign them to `ib.backupRequest.PodVolumeBackups` here: https://github.com/heptio/velero/blob/master/pkg/backup/item_backupper.go#L222

pkg/backup/request.go
```diff
@@ -10,6 +10,7 @@ import (
 // materialized (e.g. backup/snapshot locations, includes/excludes, etc.)
 type Request struct {
 	*velerov1api.Backup
+	PodVolumeBackup *velerov1api.PodVolumeBackup
```
skriss: This needs to be a slice, as one `Backup` can have more than one `PodVolumeBackups`. I would also group the new field with the other non-embedded fields just below (i.e. leave L13 blank, move to L14).

pkg/controller/backup_controller.go
```diff
@@ -576,14 +608,25 @@ func persistBackup(backup *pkgbackup.Request, backupContents, backupLog *os.File
 		errs = append(errs, errors.Wrap(err, "error closing gzip writer"))
 	}
 
+	podVolumeBackup := new(bytes.Buffer)
```
skriss: `podVolumeBackups` (plural)

pkg/controller/backup_controller.go
```diff
@@ -576,14 +608,25 @@ func persistBackup(backup *pkgbackup.Request, backupContents, backupLog *os.File
...
+	gzw = gzip.NewWriter(podVolumeBackup)
+	defer gzw.Close()
+
+	if err := json.NewEncoder(gzw).Encode(backup.PodVolumeBackup); err != nil {
+		errs = append(errs, errors.Wrap(err, "error encoding pod volume backup"))
```
skriss: "error encoding pod volume backups" (plural)

pkg/persistence/object_store.go
```diff
@@ -43,7 +43,7 @@ type BackupStore interface {
 
 	ListBackups() ([]string, error)
 
-	PutBackup(name string, metadata, contents, log, volumeSnapshots io.Reader) error
+	PutBackup(name string, metadata, contents, log, podVolumeBackup, volumeSnapshots io.Reader) error
```
skriss: `podVolumeBackups` (plural)

pkg/persistence/object_store_layout.go
```diff
@@ -83,6 +83,10 @@ func (l *ObjectStoreLayout) getBackupLogKey(backup string) string {
 	return path.Join(l.subdirs["backups"], backup, fmt.Sprintf("%s-logs.gz", backup))
 }
 
+func (l *ObjectStoreLayout) getBackupPodVolumesKey(backup string) string {
```
skriss: `getPodVolumeBackupsKey`

changelogs/unreleased/1577-carlisia
```diff
@@ -0,0 +1 @@
+Store restic PodVolumeBackups in obj storage & use tht as source of truth like regular backups.
```
nrb: tht -> that

carlisia:
@skriss question wrt `Sync to the cluster through the backup sync controller`: I'm not clear on what is new that needs to be done with the sync that is already not being done.

skriss:
> question wrt Sync to the cluster through the backup sync controller: I'm not clear on what is new that needs to be done with the sync that is already not being done.

The backup sync controller will need to be updated so when it's syncing a backup from object storage into the cluster, it also reads the contents of the PodVolumeBackups file and ensures that the custom resources contained in that file exist in the cluster - if they don't exist, they'll need to be created.

site/docs/master/restic.md
```diff
@@ -307,14 +307,12 @@ should be taken (`backup.velero.io/backup-volumes`)
...
-1. As each `PodVolumeBackup` finishes, the main Velero process captures its restic snapshot ID and adds it as an annotation
-to the copy of the pod JSON that's stored in the Velero backup. This will be used for restores, as seen in the next section.
+1. As each `PodVolumeBackup` finishes, the main Velero process adds it to the Velero backup in a file named `<backup-name>-podvolumebackups.json.gz`. This file gets uploaded to object storage alongside the backup tarball. It will be used for restores, as seen in the next section.
 
 ### Restore
 
-1. The main Velero restore process checks each pod that it's restoring for annotations specifying a restic backup
-exists for a volume in the pod (`snapshot.velero.io/<volume-name>`)
-1. When found, Velero first ensures a restic repository exists for the pod's namespace, by:
+1. The main Velero restore process checks each pod that it's restoring from a previously created restic backup by iterating through the contents of the `<backup-name>-podvolumebackups.json.gz` file.
```
nrb: I think Steve already commented on this, but is this documentation accurate?

My expectation would be that the backup sync controller would take the contents of `<backup-name>-podvolumebackups.json.gz` and submit each item to the Kubernetes API server so that they exist in the cluster.

carlisia: I just finished doing the sync so yes, what you say is correct. But doesn't the restore need to do the same?

nrb: I don't believe so. Once the `PodVolumeBackup`s exist in the Kubernetes API server from the sync, the restore process could query the Kubernetes API server for them, no need to download the gzipped json file. At least, that's how I'm reading the description in #1169.

So https://github.com/heptio/velero/blob/master/pkg/restic/restorer.go#L89 would be replaced (eventually) by a query to the Kubernetes API server for all `PodVolumeBackup`s labeled with the `Backup` name (can be gotten from the `Restore`).

carlisia: Ah. I see. Thanks for clarifying that. I'll fix this doc too.

pkg/backup/builderBackup.go
```diff
@@ -31,12 +31,12 @@ type Builder struct {
```
skriss: let's call this file `backup_builder.go`

pkg/backup/builderPodVolumes.go
```diff
@@ -0,0 +1,92 @@
+/*
```
skriss: Let's call this file pod_volume_backup_builder.go

pkg/backup/item_backupper.go
```diff
@@ -217,13 +218,14 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
...
 
 	if groupResource == kuberesource.Pods && pod != nil {
-		// this function will return partial results, so process volumeSnapshots
+		// this function will return partial results, so process podVolumeBackups
 		// even if there are errors.
-		volumeSnapshots, errs := ib.backupPodVolumes(log, pod, resticVolumesToBackup)
+		podVolumeBackups, errs := ib.backupPodVolumes(log, pod, resticVolumesToBackup)
+		ib.backupRequest.PodVolumeBackups = podVolumeBackups
 
 		// annotate the pod with the successful volume snapshots
```
skriss: We don't want to set the annotation anymore

pkg/backup/item_backupper_test.go
```diff
@@ -309,9 +310,29 @@ func TestResticAnnotationsPersist(t *testing.T) {
...
+	var podVolumeBackup2 = &velerov1api.PodVolumeBackup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "volume-2",
+		},
+		Status: velerov1api.PodVolumeBackupStatus{
+			SnapshotID: "snapshot-2",
+		},
+	}
+
+	podVolumeBackups := []*velerov1api.PodVolumeBackup{podVolumeBackup1, podVolumeBackup2}
```
skriss: doesn't look like there's any need to give `podVolumeBackup1` and `podVolumeBackup2` their own variables - let's just define them inline in the slice definition

pkg/backup/request.go
```diff
@@ -11,6 +11,7 @@ import (
 type Request struct {
 	*velerov1api.Backup
 
+	PodVolumeBackups          []*velerov1api.PodVolumeBackup
```
skriss: Let's move this down to be grouped with `VolumeSnapshots` since the top group is all "spec"-related fields.

pkg/controller/backup_controller.go
```diff
@@ -58,6 +58,7 @@ type backupController struct {
 
 	backupper                pkgbackup.Backupper
 	lister                   listers.BackupLister
+	listerPodVolumes         listers.PodVolumeBackupLister
```
skriss: let's call this `podVolumeBackupLister` or `pvbLister`, and update `lister` to be `backupLister`

pkg/controller/backup_sync_controller.go
```diff
@@ -232,9 +238,38 @@ func (c *backupSyncController) run() {
 			default:
 				log.Debug("Synced backup into cluster")
 			}
+
+			// process the pod volume backups from object store, if any
+			podVolumeBackups, err := backupStore.GetBackupPodVolumes(backupName)
+			if !kuberrs.IsNotFound(err) {
```
skriss: I don't think you would ever get a kube "not found" error here, since this isn't calling the Kubernetes API. I think you just want to check for any error here, and log/continue if one is gotten

pkg/controller/backup_sync_controller.go
```diff
@@ -232,9 +238,38 @@ func (c *backupSyncController) run() {
...
+			}
+
+			if kuberrs.IsNotFound(err) {
+				log.WithError(errors.WithStack(err)).Error("Error getting pod volumes for this backup")
+				continue
+			}
+
+			for _, podvolumeBackup := range podVolumeBackups {
+				log = log.WithField("podVolumeBackup", podvolumeBackup.Name)
+				log.Debug("Checking this pod volume to see if it needs to be synced into the cluster")
```
skriss: "Checking this pod volume backup..."

pkg/controller/backup_sync_controller.go
```diff
@@ -232,9 +238,38 @@ func (c *backupSyncController) run() {
...
+
+				_, err = c.podVolumeBackupClient.PodVolumeBackups(backup.Namespace).Create(podvolumeBackup)
+				switch {
+				case err != nil && kuberrs.IsAlreadyExists(err):
+					log.Debug("Pod volume already exists in cluster")
```
skriss: "Pod volume backup..."

pkg/controller/backup_sync_controller.go
```diff
@@ -232,9 +238,38 @@ func (c *backupSyncController) run() {
...
+					continue
+				case err != nil && !kuberrs.IsAlreadyExists(err):
+					log.WithError(errors.WithStack(err)).Error("Error syncing pod volume into cluster")
```
skriss: "Error syncing pod volume backup..."

pkg/controller/backup_sync_controller.go
```diff
@@ -232,9 +238,38 @@ func (c *backupSyncController) run() {
...
+					continue
+				default:
+					log.Debug("Synced pod volume into cluster")
```
skriss: "Synced pod volume backup..."

pkg/controller/backup_sync_controller.go
```diff
@@ -232,9 +238,38 @@ func (c *backupSyncController) run() {
...
+				}
+			}
 		}
 
 		c.deleteOrphanedBackups(location.Name, backupStoreBackups, log)
+		c.deleteOrphanedPodVolumeBackups(location.Name, backupStoreBackups, log)
```
skriss: I'm not sure that this function is needed -- `PodVolumeBackups` are linked by owner ref to the `Backup` that triggered them, and when the `Backup` is deleted, the owned `PodVolumeBackups` are automatically deleted as well. Need to think about it a little more.

pkg/persistence/object_store.go
```diff
@@ -43,9 +43,10 @@ type BackupStore interface {
 
 	ListBackups() ([]string, error)
 
-	PutBackup(name string, metadata, contents, log, volumeSnapshots io.Reader) error
+	PutBackup(name string, metadata, contents, log, podVolumeBackups, volumeSnapshots io.Reader) error
```
skriss: Wondering if we should define a struct to encapsulate some/all of the arguments to this function, given that we're up to 6 - WDYT?

pkg/persistence/object_store.go
```diff
@@ -43,9 +43,10 @@ type BackupStore interface {
...
 	GetBackupMetadata(name string) (*velerov1api.Backup, error)
 	GetBackupVolumeSnapshots(name string) ([]*volume.Snapshot, error)
+	GetBackupPodVolumes(name string) ([]*velerov1api.PodVolumeBackup, error)
```
skriss: GetPodVolumeBackups

pkg/persistence/object_store.go
```diff
@@ -166,7 +167,7 @@ func (s *objectBackupStore) ListBackups() ([]string, error) {
 	return output, nil
 }
 
-func (s *objectBackupStore) PutBackup(name string, metadata, contents, log, volumeSnapshots io.Reader) error {
+func (s *objectBackupStore) PutBackup(name string, metadata, contents, log, podVolumeBackup, volumeSnapshots io.Reader) error {
```
skriss: podVolumeBackups

pkg/restic/backupper.go
```diff
@@ -184,7 +185,19 @@ ForEachVolume:
 	delete(b.results, resultsKey(pod.Namespace, pod.Name))
 	b.resultsLock.Unlock()
 
-	return volumeSnapshots, errs
+	for volumeName, snapshotID := range volumeSnapshots {
```
skriss: It doesn't look like this approach is going to work - we need to capture the full PodVolumeBackup objects as they exist in the API server.  I would suggest changing the `volumeSnapshots` field to be a  `map[string]*velerov1api.PodVolumeBackup` or something similar, and then around https://github.com/heptio/velero/pull/1577/files#diff-d86635f0e9c2aff5dfda0f868fd8062fR170 you'll need to capture the PodVolumeBackups as they're completed/failed. Let me know if you need more input here.

pkg/restic/common.go
```diff
@@ -39,25 +39,15 @@ const (
...
 	DefaultMaintenanceFrequency = 24 * time.Hour
 
-	podAnnotationPrefix       = "snapshot.velero.io/"
+	// Deprecated.
+	podAnnotationPrefix = "snapshot.velero.io/"
+
 	volumesToBackupAnnotation = "backup.velero.io/backup-volumes"
 )
 
-// PodHasSnapshotAnnotation returns true if the object has an annotation
```
skriss: I think you'll still need this function to maintain backwards-compatibility for restoring pre-v1.1 backups

skriss:
Let me know if you want to talk through/pair on any of the comments!

pkg/persistence/object_store.go
carlisia: Agree. More then 3 and it's something to think about, in this case it totally makes sense.

pkg/controller/backup_sync_controller.go
carlisia: Ok, let me know. I'm not familiar with how the owner ref, probably a good opportunity for me to look into it too.

pkg/persistence/object_store.go
nrb: :+1:. The CSI proposal adds yet another argument for VolumeSnapshots, so I'm for a struct.

pkg/controller/backup_sync_controller.go
nrb: The owner reference is set on the metadata, had to look this up myself. https://github.com/heptio/velero/blob/master/pkg/restic/backupper.go#L231

nrb: See also https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#ownerreference-v1-meta

pkg/restic/common.go
carlisia: I'm thinking we should add an annotation so this doesn't seem to be completely unused code. Not remembering we have a standard. Going to add something.

nrb: It's what we use now, so it has use for existing backups. The function should stick around for now to support them, as Steve says.

Are you thinking we continue to add annotations to PodVolumeBackups going forward, too?

carlisia: I don't think we use it now: https://github.com/heptio/velero/search?q=PodHasSnapshotAnnotation&unscoped_q=PodHasSnapshotAnnotation

carlisia: Sorry. I said annotation but I meant documentation to this function if it needs to stick around. I understand we use annotation still, but I'm saying this function in specific is not being used in the current version of the code.

carlisia: @skriss ^

skriss: Oh - guess this is already not used! Delete away. Looks like our restore code uses the `GetPodSnapshotAnnotations` function instead.

pkg/controller/backup_sync_controller.go
carlisia: @skriss  So.. `The owner reference is set on the metadata`, and the metadata has the backup info, deleting the backup would delete any `podvolumebackup` associated with it?

skriss: Yep - that's correct. So (a) we shouldn't have a scenario where a PVB exists that doesn't have a Backup as an owner, and (b) deleting a backup will automatically delete associated PVBs. So as far as I can tell, we shouldn't have a need for this function

carlisia: Thank you 👍

carlisia:
I tried to address the comments, PTAL.

I'm stuck on how to resolve this mess, any hint: https://github.com/heptio/velero/pull/1577/files#diff-d0a3ff018929aec3abbaa4c8a2c299ccR575

I'm suspecting the answer will be initialize the entire struct with mocked objects for each element?

pkg/controller/backup_sync_controller.go
```diff
@@ -232,6 +238,30 @@ func (c *backupSyncController) run() {
...
+			podVolumeBackups, err := backupStore.GetPodVolumeBackups(backupName)
+			if err != nil {
+				log.WithError(errors.WithStack(err)).Error("Error getting pod volumes for this backup from backup store")
+				continue
+			}
+
+			for _, podvolumeBackup := range podVolumeBackups {
+				log = log.WithField("podVolumeBackup", podvolumeBackup.Name)
+				log.Debug("Checking this pod volume backup to see if it needs to be synced into the cluster")
+
```
skriss: We'll need to update the PVB's owner ref's UID to match the backup from L230, since if we're syncing into a different cluster, it'll be different. It looks like we also set a label on the PVB with this UID, so we need to update that as well (see https://github.com/heptio/velero/blob/master/pkg/restic/backupper.go#L236 and https://github.com/heptio/velero/blob/master/pkg/restic/backupper.go#L242)

pkg/persistence/object_store.go
```diff
@@ -288,6 +306,40 @@ func (s *objectBackupStore) GetBackupVolumeSnapshots(name string) ([]*volume.Sna
 	return volumeSnapshots, nil
 }
 
+func (s *objectBackupStore) GetPodVolumeBackups(name string) ([]*velerov1api.PodVolumeBackup, error) {
+	key := s.layout.getPodVolumeBackupsKey(name)
```
skriss: It would be nice to extract 1+ helpers from this, since most of this new function is a copy-paste of `GetBackupVolumeSnapshots`. You could have just one that takes the backup name, key, and an interface{} to json-decode into and does the whole process, or you could have one for trying to get the object and another for gzip/json-decoding. If you want to file a tech debt issue to follow up on this, I'm fine with not doing it right now in the interest of getting this merged

pkg/restic/backupper.go
```diff
@@ -150,13 +151,11 @@ func (b *backupper) BackupPodVolumes(backup *velerov1api.Backup, pod *corev1api.
 		}
 
 		volumeBackup := newPodVolumeBackup(backup, pod, volumeName, repo.Spec.ResticIdentifier)
-
-		if err := errorOnly(b.repoManager.veleroClient.VeleroV1().PodVolumeBackups(volumeBackup.Namespace).Create(volumeBackup)); err != nil {
+		volumeSnapshots[volumeName] = volumeBackup
```
skriss: it doesn't look like we really need the `volumeSnapshots` map anymore - the only thing I see us using it for is to know the number of PVB's that we created and are waiting for. So I'd just keep an `int` count here instead, and use that on https://github.com/heptio/velero/pull/1577/files#diff-d86635f0e9c2aff5dfda0f868fd8062fR162

pkg/restic/common.go
```diff
@@ -79,6 +69,7 @@ func GetPodSnapshotAnnotations(obj metav1.Object) map[string]string {
 
 // SetPodSnapshotAnnotation adds an annotation to a pod to indicate that
 // the specified volume has a restic snapshot with the provided id.
+// Deprecated: we will stop using pod annotations to record restic snapshot IDs after they're taken.
 func SetPodSnapshotAnnotation(obj metav1.Object, volumeName, snapshotID string) {
```
skriss: we should be able to delete this function now, since we shouldn't be using it anywhere anymore.

pkg/controller/backup_controller_test.go
```diff
@@ -561,11 +559,20 @@ func TestProcessBackupCompletions(t *testing.T) {
...
+			// backupInfo := persistence.BackupInfo{
+			// 	Name: test.backup.Name,
+			// 	// Metadata:         mock.MatchedBy(completionTimestampIsPresent),
+			// 	// Contents: mock.Anything,
+			// 	// Log: mock.Anything,
+			// 	// PodVolumeBackups: mock.Anything,
+			// 	// VolumeSnapshots:  mock.Anything,
+			// }
+			// TODO: [carlisia] resolve this mock
+			backupStore.On("PutBackup", mock.Anything).Return(nil)
```
skriss: I **think** what you want to do here is define a function that can be used as an arg to `mock.MatchedBy` -- so something like the following:

```go
func hasNameAndCompletionTimestamp(info persistence.BackupInfo) bool {
    return info.Name == test.backup.Name &&
               strings.Contains(info.Metadata.String(), `"completionTimestamp": "2006-01-02T22:04:05Z"`)    
}
```

And then this line becomes:

```go
backupStore.On("PutBackup", mock.MatchedBy(hasNameAndCompletionTimestamp)).Return(nil)
```

pkg/controller/backup_sync_controller.go
```diff
@@ -179,26 +185,25 @@ func (c *backupSyncController) run() {
...
 
-				if backup.Spec.StorageLocation != "" {
-					continue
-				}
-
 				// pre-v0.10 backups won't initially have a .spec.storageLocation so fill it in
-				log.Debug("Patching backup's .spec.storageLocation because it's missing")
-				if err := patchStorageLocation(backup, c.backupClient.Backups(c.namespace), location.Name); err != nil {
-					log.WithError(err).Error("Error patching backup's .spec.storageLocation")
+				if backup.Spec.StorageLocation == "" {
```
nrb: Should we drop this code? 1.0.0 wasn't supposed to support pre-v0.10 backups.

pkg/persistence/object_store.go
nrb: +1 to merging with a todo/tech debt issue.

pkg/controller/backup_sync_controller.go
skriss: Yeah I think that's fine, we probably just missed it bc there was no // TODO(1.0) on it

pkg/controller/backup_controller_test.go
```diff
@@ -556,16 +556,18 @@ func TestProcessBackupCompletions(t *testing.T) {
...
+			backupStore.On("BackupExists", test.backupLocation.Spec.StorageType.ObjectStorage.Bucket, test.backup.Name).Return(test.backupExists, test.existenceCheckError)
 
 			// Ensure we have a CompletionTimestamp when uploading.
 			// Failures will display the bytes in buf.
-			completionTimestampIsPresent := func(buf *bytes.Buffer) bool {
-				return strings.Contains(buf.String(), `"completionTimestamp": "2006-01-02T22:04:05Z"`)
+			completionTimestampIsPresent := func(info persistence.BackupInfo) bool {
+				buf := new(bytes.Buffer)
+				buf.ReadFrom(info.Metadata)
+				return info.Name == test.backup.Name &&
```
carlisia: This I think is closer now to the original. But it's failing with a panic intermittently. I just realized I think I know the fix, hold for that, 1 min.

carlisia:
Thank you, making those changes!

carlisia:
I'm trying to make sense of this output that I'm getting intermittently: https://gist.github.com/carlisia/8eb210744758a70e41af47216a4107ae

It happens when I run `TestBackupItemNoSkips` on the `item_backupper_test`.

skriss:
I see:
```
expected: "resources/pods/cluster/.json"
actual  : "resources/pods/namespaces/foo/bar.json"
```

Which looks related to https://github.com/heptio/velero/pull/1577#discussion_r306853387 - you need to fully revert changes that test case.

I also see a typo error related to the backup controller tests -- "mock.Anythin" (should be "mock.Anything")

carlisia:
Ah yes, I had reverted that line but not `expectedTrackedPVCs:       sets.NewString(key("foo", "bar"), key("foo", "baz")),`. Added it, pushing it in a sec.

If you search my diff, there's no `resources/pods/cluster/.json` anywhere. Same when I do a global search in my branch. This is puzzling.

There is no `mock.Anythin` in my diff. I did have this typo yesterday, but corrected it. It's as if an old test result is being outputted.

skriss:
> There is no mock.Anythin in my diff. I did have this typo yesterday, but corrected it. It's as if an old test result is being outputted.

Often this implies the tests are failing to compile - I'll try your pushed code locally

carlisia:
Still getting the failure intermittently  (as I expected since I didn't change anything relevant).

I think it's about this:

```
mock: Unexpected Method Call
-----------------------------

PutBackup(persistence.BackupInfo)
		0: persistence.BackupInfo{Name:"backup-1", Metadata:(*bytes.Buffer)(0xc000600000), Contents:(*os.File)(0xc0000105a0), Log:(*os.File)(0xc0000d4610), PodVolumeBackups:(*bytes.Buffer)(0xc000374f90), VolumeSnapshots:(*bytes.Buffer)(0xc000374e10)}

The closest call I have is: 

PutBackup(persistence.BackupInfo)
		0: persistence.BackupInfo{Name:"backup-1", Metadata:io.Reader(nil), Contents:io.Reader(nil), Log:io.Reader(nil), PodVolumeBackups:io.Reader(nil), VolumeSnapshots:io.Reader(nil)}
```

Which I think it's complaining that the `BackupInfo` is not being built as expected when I'm calling that `PutBackup` mock.

skriss:
Everything runs fine for me - tried running things 100x to look for flakes. It's also passing CI

nrb:
What commands are you both running for these tests?

carlisia:
Yeah, good to go!

I think vscode was caching something. Closed it and opened it and I don't get it anymore. 

I ran it a few times on my terminal and ran fine.

Issue #1575, opened 2019-06-12
Full blog posts show excerpt separator


seemonicago:
When viewing a full blog post, the text '' is visible.

Example: https://velero.io/velero-is-an-open-source-tool-to-back-up-and-migrate-kubernetes-clusters/

The text '' should not be visible. This text is used to identify the end of a post excerpt shown in the blog-post-cards page. 

https://github.com/heptio/velero/blob/master/site/_includes/blog-post-card.html

velero version 1.0

carlisia:
![image](https://user-images.githubusercontent.com/16508/59372914-64f67a00-8cfd-11e9-9cdc-4c40993992b2.png)

skriss:
Closed by #1578

PR #1578, opened 2019-06-12
Discontinue use of excerpt separators in blog posts.

seemonicago:
Fixes #1575

Signed-off-by: Monica rodriguezmo@vmware.com


seemonicago:
After doing some research, it seems like this is an issue with the way that the redcarpet md rendered works. There were issues opened a few years back related to this, but it has since gone back and forth with reverts as well as alternative solutions. Alternative solutions include switching renderers or explicitly specifying the page excerpt in the post. I also looked at using different separators than '', but did not have any more success. I'm not super pleased with this solution as it duplicates text. Open to better suggestions.

jekyll/jekyll#1839
 vmg/redcarpet#268

skriss:
@seemonicago this seems reasonable to me - it's very clear how to specify the excerpt and we're not relying on any "magic" to have things be rendered properly.  Thanks for digging into this!

cc @nrb @carlisia @prydonius @jonasrosland

PR #1584, opened 2019-06-14
Add redirect for docs

### LINKED ISSUE #1563: velero.io/docs returns 404
Navigate to https://velero.io/docs/, get 404

To be redirected to the latest docs

jonasrosland:
This will add a redirect for https://velero.io/docs and https://velero.io/docs/latest.

Fixes #1563.

Signed-off-by: jonasrosland <jrosland@vmware.com>

### DIFF
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -157,7 +157,7 @@ plugins:
   - jekyll-optional-front-matter # Parse Markdown files that do not have front-matter callouts
   - jekyll-titles-from-headings # pull the page title from the first Markdown heading when none is specified.
   - jekyll-paginate # pagination object for collections (e.g. posts)
-
+  - jekyll-redirect-from
 
 # Include these subdirectories
 include:
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -164,7 +164,6 @@ plugins:
   - jekyll-titles-from-headings # pull the page title from the first Markdown heading when none is specified.
   - jekyll-paginate # pagination object for collections (e.g. posts)
   - jekyll-redirect-from
-#  - data_page_generator
 
 # Include these subdirectories
 include:
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -121,6 +121,12 @@ defaults:
       gh: https://github.com/heptio/velero/tree/v0.3.0
       layout: "docs"
 
+page_gen:
+  - data: shortlinks
+    template: redirect
+    name: key
+    dir: docs
+
 collections:
   - contributors
   - casestudies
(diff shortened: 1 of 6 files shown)


site/docs/index.md
```diff
@@ -0,0 +1,4 @@
+---
+redirect_from: "docs/latest"
+redirect_to: "docs/v1.0.0"
```
nrb: So we have a `latest` variable in the _config.yaml file. I'm wondering if we could include it here as `docs/{ site.latest }`, but I'm not sure. Being able to do so would mean we don't have to update the latest version in 2 spots.

jonasrosland: Yeah that doesn't work unfortunately. It's been asked for by the Jekyll community, but I tried it, and nope, it doesn't work.
Right now we'll have to do 1 update for the main site to point to the right docs, and then 1 minor update for this to work.

nrb: Ah, that's unfortunate.

Here's what we had before, and it worked fairly well. Looks like it also gets to use site.latest 

https://github.com/heptio/velero/blob/gh-pages/_layouts/redirect.html

jonasrosland: I'll see if I can make that work here.

jonasrosland: Had to change it a bit, got it working. Will not work with `docs/latest`, but `/docs` will always point to the `latest` configured in `_config.yml`.

skriss:
Awesome! Seems to work great. Can you also update https://github.com/heptio/velero/blame/master/README.md#L76 to just point to `velero.io/docs` now?

Unfortunately we still can't do e.g. `velero.io/docs/troubleshooting/` and get a specific page to redirect to the latest version - so a few of those other links on README.md need to remain hardcoded for now. Based on looking at this though, it looks like it might be possible with a little more jekyll hackery.

site/_config.yml
```diff
@@ -121,6 +121,12 @@ defaults:
       gh: https://github.com/heptio/velero/tree/v0.3.0
       layout: "docs"
 
+page_gen:
```
nrb: So I think I've pieced this together - this tells the pagegen plugin to use the data in the `shortlinks` file to generate a page per entry, using the `redirect.html` template.

Adding in new redirects should be as simple as adding the entry into `shortlinks.yml`, right?

jonasrosland: Yup, you got it!

site/_config.yml
```diff
@@ -157,7 +163,8 @@ plugins:
   - jekyll-optional-front-matter # Parse Markdown files that do not have front-matter callouts
   - jekyll-titles-from-headings # pull the page title from the first Markdown heading when none is specified.
   - jekyll-paginate # pagination object for collections (e.g. posts)
-
+  - jekyll-redirect-from
+#  - data_page_generator
```
skriss: curious why we have to include the code for this one vs. referencing it as a plugin? If that's necessary, let's delete this commented out . line

nrb: It doesn't appear to be published on Ruby gems, unlike the others.

jonasrosland: Yeah it's a manual add, I've now removed the commented out line.

Issue #1440, opened 2019-05-03
Document resource inclusion/exclusion logic


nrb:
Having a document that describes the logic for including/excluding resources with combination of backup specs would be a useful reference for users to understand what their options are, as well as a place to point people instead of re-explaining.

carlisia:
Things to consider adding:
- a bit of set theory that we use
- the logic
- examples
- list of things/examples not possible to select

prydonius:
This should also include the standard label used to exclude resources introduced in #1588

nrb:
I would propose putting this in it's own document under `Use`, because it's a key part of using the backup and restore commands.

skriss:
👍 agree - I'd like to expand the Use section to have more examples of ways you can use Velero, that cover all of these different flags/etc.

jaxxstorm:
question @carlisia and @nrb - why was a label chosen to exclude resources instead of an annotation? The accepted convention is generally to use annotations for this kind of exclusion decision

nrb:
@jaxxstorm Mostly for the query capabilities that labels provide. Annotations don't provide any way of querying via the client APIs.

Can you provide a link to the conventions you're describing?

jaxxstorm:
I am yet to find another kubernetes based project or operator that use labels in this way. Some examples:

https://github.com/hjacobs/kube-downscaler - uses annotations for exclusion logic
https://github.com/kubernetes-incubator/external-dns - uses annotations for operational logic

I guess if you want to make this queryable the using labels seems fine, but I haven't found any other project that follows this pattern

prydonius:
@jaxxstorm that's a good point. In fact with #1843 we no longer query using exclusion label, so we could consider changing it to an annotation. Of course, that would be a breaking change for anyone using the label right now, maybe we could support both? WDYT @skriss @nrb @carlisia?

jaxxstorm:
i think supporting both seems reasonable

nrb:
Yeah, I'd be +1 on doing both and dropping the label in v2.0. I consider the well-known labels/annotations in a project to be part of the public API, even if it makes the process of changing a little onerous.

For further context, the use of the label came as an extension of an example in our docs before it was actually codified.

carlisia:
More logic for backup/restore:

`IncludeClusterResources` is a nullable bool, so has 3 possible modes:

- `true`: all cluster-scoped resources that are selected by the resource/label selectors are restored
- `false`: no cluster-scoped resources are restored
[...]

PR #1588, opened 2019-06-17
allow exclusion of resources using standard label

### LINKED ISSUE #1373: Exclude resource from backup by a standard label
Provide a standard label that when applied to a resource will exclude that resource from being backed up.

_Example:_
```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: do-not-backup
  namespace: default
  labels:
    # Any resource with this a standard label/value will be excluded from all backups.
    velero.io/backup-excluded: "true"
data:
  foo: {{bar}}
```

Our specific use-case is for cluster migrations. We'd like to prevent a small subset of the resources that are created during cluster bootstrap (applying manifests via Ansible) being overwritten when we apply a restore. The restore from another cluster may include cluster-specific configuration that would be invalid in the new cluster, so there is no point in backing those resources up given that we will bootstrap them.

_Our Current solution:_
```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: cluster-specific-config-created-at-cluster-bootstrap
  namespace: default
  labels:
    backup: "false" # Include our own standard label.
data:
  clusterSpecificData: {{cluster_specific_data}}
```
```bash
# Use a label selector to filter where backup==false
ark backup create example -l 'backup notin (false)'
```

I understand there are alternatives but this seems like a simple way for us to cherry-pick out certain resources that should not be backed up. What do you think?

**Environment:**

- Velero version (use `velero version`): 0.10.0
- Kubernetes version (use `kubectl version`): 1.10
- Kubernetes installer & version: Terraform & Ansible

prydonius:
excludes any resources with the velero.io/exclude-from-backup=true label

fixes #1373 

### DIFF
--- a/pkg/backup/backup_new_test.go
+++ b/pkg/backup/backup_new_test.go
@@ -249,16 +249,16 @@ func TestBackupResourceFiltering(t *testing.T) {
 				Backup(),
 			apiResources: []*apiResource{
 				pods(
-					withLabel(withLabel(newPod("foo", "bar"), "velero.io/exclude-from-backup", "true"), "a", "b"),
+					withLabel(newPod("foo", "bar"), "velero.io/exclude-from-backup", "true", "a", "b"),
 					withLabel(newPod("zoo", "raz"), "a", "b"),
 				),
 				deployments(
 					newDeployment("foo", "bar"),
-					withLabel(withLabel(newDeployment("zoo", "raz"), "velero.io/exclude-from-backup", "true"), "a", "b"),
+					withLabel(newDeployment("zoo", "raz"), "velero.io/exclude-from-backup", "true", "a", "b"),
 				),
 				pvs(
 					withLabel(newPV("bar"), "a", "b"),
-					withLabel(withLabel(newPV("baz"), "a", "b"), "velero.io/exclude-from-backup", "true"),
+					withLabel(newPV("baz"), "a", "b", "velero.io/exclude-from-backup", "true"),
 				),
 			},
 			want: []string{
@@ -1870,12 +1870,14 @@ func newHarness(t *testing.T) *harness {
 	}
 }
 
-func withLabel(obj metav1.Object, key, val string) metav1.Object {
+func withLabel(obj metav1.Object, labelPairs ...string) metav1.Object {
 	labels := obj.GetLabels()
 	if labels == nil {
 		labels = make(map[string]string)
 	}
-	labels[key] = val
+	for i := 0; i < len(labelPairs); i += 2 {
+		labels[labelPairs[i]] = labelPairs[i+1]
+	}
 	obj.SetLabels(labels)
 
 	return obj
--- a/pkg/backup/backup_new_test.go
+++ b/pkg/backup/backup_new_test.go
@@ -219,6 +219,54 @@ func TestBackupResourceFiltering(t *testing.T) {
 			},
 		},
 		{
+			name: "resources with velero.io/exclude-from-backup=true label are not included",
+			backup: defaultBackup().
+				Backup(),
+			apiResources: []*apiResource{
+				pods(
+					withLabel(newPod("foo", "bar"), "velero.io/exclude-from-backup", "true"),
+					newPod("zoo", "raz"),
+				),
+				deployments(
+					newDeployment("foo", "bar"),
+					withLabel(newDeployment("zoo", "raz"), "velero.io/exclude-from-backup", "true"),
+				),
+				pvs(
+					withLabel(newPV("bar"), "a", "b"),
+					withLabel(newPV("baz"), "velero.io/exclude-from-backup", "true"),
+				),
+			},
+			want: []string{
+				"resources/pods/namespaces/zoo/raz.json",
+				"resources/deployments.apps/namespaces/foo/bar.json",
+				"resources/persistentvolumes/cluster/bar.json",
+			},
+		},
+		{
+			name: "resources with velero.io/exclude-from-backup=true label are not included even if matching label selector",
+			backup: defaultBackup().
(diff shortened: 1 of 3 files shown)


prydonius:
I guess we should also add some documentation about this new feature, where do we think that should go?

pkg/backup/backup_new_test.go
```diff
@@ -218,6 +218,54 @@ func TestBackupResourceFiltering(t *testing.T) {
...
+			},
+		},
+		{
+			name: "resources with velero.io/exclude-from-backup=true label are not included even if matching label selector",
+			backup: defaultBackup().
+				LabelSelector(&metav1.LabelSelector{MatchLabels: map[string]string{"a": "b"}}).
+				Backup(),
+			apiResources: []*apiResource{
+				pods(
+					withLabel(withLabel(newPod("foo", "bar"), "velero.io/exclude-from-backup", "true"), "a", "b"),
```
prydonius: this was a lazy way of adding multiple labels to the pod, I can create another helper that takes multiple labels that should be easier to read

carlisia: Not sure we have a place, but should go together with this: #1440

pkg/backup/resource_backupper.go
```diff
@@ -222,8 +222,10 @@ func (rb *defaultResourceBackupper) backupResource(group *metav1.APIResourceList
 		}
 
 		var labelSelector string
+		excludeLabelSelector := "velero.io/exclude-from-backup!=true"
+		labelSelector = excludeLabelSelector
 		if selector := rb.backupRequest.Spec.LabelSelector; selector != nil {
-			labelSelector = metav1.FormatLabelSelector(selector)
+			labelSelector = labelSelector + "," + metav1.FormatLabelSelector(selector)
```
carlisia: I would `s/labelSelector/"velero.io/exclude-from-backup!=true"` and delete lns 225,226. WDYT? I don't think this is so complex that needs to be expanded. Less lines is better if it's still clear.

pkg/backup/backup_new_test.go
skriss: Yeah, it would be nice to update the helper. One option that keeps it pretty concise would be to make it variadic, and just assume it goes `..., label key 1, label value 1, label key 2, label value 2, ...`. That way you don't have to use a map or anything else.

carlisia:
> Not sure we have a place, but should go together with this: #1440

This ^ was supposed to be in response to:

> I guess we should also add some documentation about this new feature, where do we think that should go?

pkg/backup/backup_new_test.go
```diff
@@ -218,6 +218,54 @@ func TestBackupResourceFiltering(t *testing.T) {
...
+				pvs(
+					withLabel(newPV("bar"), "a", "b"),
+					withLabel(withLabel(newPV("baz"), "a", "b"), "velero.io/exclude-from-backup", "true"),
+				),
+			},
+			want: []string{
+				"resources/pods/namespaces/zoo/raz.json",
+				"resources/persistentvolumes/cluster/bar.json",
+			},
+		},
```
skriss: Can you add a case that covers the new label key being specified, but the value not being true?

pkg/backup/resource_backupper.go
skriss: I'd recommend:

```go
labelSelector := "velero.io/exclude-from-backup!=true"
if selector := rb.backupRequest.Spec.LabelSelector; selector != nil {
    labelSelector = labelSelector + "," + metav1.FormatLabelSelector(selector)
...
```

skriss:
Yeah, I'm not sure there's a great place to put docs for it right now -- agree it would make sense to go with https://github.com/heptio/velero/issues/1440, probably on a new page under "Use" in the TOC.

pkg/backup/backup_new_test.go
prydonius: @skriss nice, I was thinking about something like that, will look into variadic variables

pkg/backup/resource_backupper.go
prydonius: Agree there's a few more lines than necessary here, I like @skriss's suggestion as it will make it easier to find just that labelSelector in the future

pkg/backup/backup_new_test.go
```diff
@@ -1795,12 +1870,14 @@ func newHarness(t *testing.T) *harness {
...
 
-func withLabel(obj metav1.Object, key, val string) metav1.Object {
+func withLabel(obj metav1.Object, labelPairs ...string) metav1.Object {
 	labels := obj.GetLabels()
 	if labels == nil {
 		labels = make(map[string]string)
 	}
-	labels[key] = val
+	for i := 0; i < len(labelPairs); i += 2 {
+		labels[labelPairs[i]] = labelPairs[i+1]
```
prydonius: @skriss this will panic if `labelPairs` is not even. I was thinking about catching that, but since the `testing.T` object isn't passed in here, we can't fail the test more gracefully. Not sure if we want to pass `t` into each `withLabel` call. We could also just panic with a more helpful error (e.g. "withLabel requires key-value pairs when specifying labels")

skriss: I think panicking with a helpful error if there's an odd number of values should be fine. It's really just to detect programmer error.

prydonius:
> Yeah, I'm not sure there's a great place to put docs for it right now -- agree it would make sense to go with #1440, probably on a new page under "Use" in the TOC.

added a comment in #1440 to include this

Also updated based on review, PTAL!

prydonius:
@skriss really good, the existing tests are really helpful and pretty easy to work off of!

PR #1591, opened 2019-06-19
migrate backup hooks tests

skriss:

@nrb @carlisia @prydonius in migrating the tests related to backup hooks, I didn't really see a better way to structure these than using a mock for the `PodCommandExecutor`. The interaction between `pkg/backup` and the `PodCommandExecutor` is really simple - just a single method call - and this test is just verifying that the method is called when a hook is configured. WDYT about this?


skriss:
added a couple more test cases

pkg/backup/backup_new_test.go
```diff
@@ -1629,6 +1631,280 @@ func TestBackupWithSnapshots(t *testing.T) {
 	}
 }
 
+// TestBackupWithInvalidHooks runs backups with invalid hook specifications and verifies
+// that an error is returned.
+func TestBackupWithInvalidHooks(t *testing.T) {
+	tests := []struct {
```
nrb: Is this a table based test just to make it easier to add cases in the future?

skriss: Yeah, less typing if someone wants to add more cases down the road

skriss:
If the mock usage seems reasonable, then this can be reviewed since I added a couple more test cases. There's probably an opportunity to add a few more test cases if someone works with the hooks code later, but TBH I think moving on to other things (e.g. `pkg/restore`) is a better use of time right now.

pkg/backup/backup_new_test.go
```diff
@@ -1629,6 +1631,280 @@ func TestBackupWithSnapshots(t *testing.T) {
...
+						"hook-1",      // hook name
+						&velerov1.ExecHook{
+							Command: []string{"ls", "/tmp"},
+							OnError: velerov1.HookErrorModeFail,
+						},
+					},
+					returnArguments: []interface{}{errors.New("exec hook error")},
+				},
+				{
+					arguments: []interface{}{
```
prydonius: Given these arguments are pretty similar across all these tests, couldn't we omit the first two args and last one and just build them in the test code? Or is it possible other tests could change these args?

skriss: Yeah, I think that's reasonable - I'll tweak next week.

skriss: I did a little refactoring here which I think addresses your comment, plus added one more test case for the pre+post hook scenario. If it looks good, please squash + merge!

skriss:
merge conflict, 1 sec

pkg/backup/backup_new_test.go
prydonius: Thanks, this looks a lot simpler!

PR #1598, opened 2019-06-21
remove unused hasControllerOwner func and tests

skriss:


skriss:
@nrb @prydonius PTAL.

PR #1601, opened 2019-06-24
ensure correct backup item actions run with namespace selector

prydonius:

### DIFF
--- a/pkg/backup/backup_new_test.go
+++ b/pkg/backup/backup_new_test.go
@@ -806,7 +806,7 @@ func TestBackupActionsRunForCorrectItems(t *testing.T) {
 			},
 		},
 		{
-			name: "single action with a namespace selector runs only for resources in that namespace and namespace resources",
+			name: "single action with a namespace selector runs only for resources in that namespace",
 			backup: defaultBackup().
 				Backup(),
 			apiResources: []*test.APIResource{
@@ -828,7 +828,7 @@ func TestBackupActionsRunForCorrectItems(t *testing.T) {
 				),
 			},
 			actions: map[*recordResourcesAction][]string{
-				new(recordResourcesAction).ForNamespace("ns-1"): {"ns-1/pod-1", "ns-1/pvc-1", "ns-1", "ns-2"},
+				new(recordResourcesAction).ForNamespace("ns-1"): {"ns-1/pod-1", "ns-1/pvc-1"},
 			},
 		},
 		{
--- a/pkg/backup/backup_new_test.go
+++ b/pkg/backup/backup_new_test.go
@@ -806,8 +806,7 @@ func TestBackupActionsRunForCorrectItems(t *testing.T) {
 			},
 		},
 		{
-			// TODO this seems like a bug
-			name: "single action with a namespace selector runs for resources in that namespace plus cluster-scoped resources",
+			name: "single action with a namespace selector runs only for resources in that namespace and namespace resources",
 			backup: defaultBackup().
 				Backup(),
 			apiResources: []*test.APIResource{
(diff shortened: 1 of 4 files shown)


pkg/backup/backup_new_test.go
```diff
@@ -827,9 +826,13 @@ func TestBackupActionsRunForCorrectItems(t *testing.T) {
...
 					newPV("pv-2"),
 				),
+				namespaces(
+					newNamespace("ns-1"),
+					newNamespace("ns-2"),
+				),
 			},
 			actions: map[*recordResourcesAction][]string{
-				new(recordResourcesAction).ForNamespace("ns-1"): {"ns-1/pod-1", "ns-1/pvc-1", "pv-1", "pv-2"},
+				new(recordResourcesAction).ForNamespace("ns-1"): {"ns-1/pod-1", "ns-1/pvc-1", "ns-1", "ns-2"},
```
prydonius: @skriss not sure if I got this right that all Namespace resources should have actions run, seems odd that actions should run for "ns-2"?

skriss: Yeah, I was just looking at that. TBH I think it's probably more logical to *not* run these actions for Namespace resources themselves, only for the namespaced resources in the specified namespace(s).

skriss: @nrb thoughts?

pkg/backup/item_backupper.go
```diff
@@ -302,6 +302,23 @@ func (ib *defaultItemBackupper) executeActions(
 			continue
 		}
 
+		// Determine whether to run actions for cluster-scoped resources based on IncludeClusterResources
+		// NOTE: we specifically allow actions to be run for namespaces even if IncludeClusterResources is
+		// false.
+		if namespace == "" && groupResource != kuberesource.Namespaces {
```
skriss: I think we can simplify this whole block (L308-321) a lot - some of the checks here aren't relevant. I *believe* what we want is:

```go

if namespace == "" && !action.namespaceIncludesExcludes.IncludeEverything() {
    log.Debug("Skipping action because resource is cluster-scoped and action only applies to specific namespaces")
    continue
}
```

pkg/backup/backup_new_test.go
prydonius: oops - just updated to not run these for namespace resources, but happy to change again based on @nrb's feedback

skriss: I think what you have now makes sense, but we'll get @nrb's opinion too.

pkg/backup/item_backupper.go
prydonius: Ah I see, so IncludeClusterResources is only relevant for backups and not for actions?

pkg/backup/backup_new_test.go
prydonius: I left the namespace resources in the test as well to make this more explicit, but can remove if we don't think it's really necessary

pkg/backup/item_backupper.go
skriss: Correct. It specifies whether to back up all cluster-scoped resources or not, but shouldn't have any impact on whether specific backup item actions are run for those resources or not.

pkg/backup/backup_new_test.go
skriss: I'd keep it, adds some clarity.

pkg/backup/item_backupper.go
prydonius: Thanks that makes sense!

prydonius:
@skriss @nrb rebased

skriss:
👍

PR #1606, opened 2019-06-26
Refactor pkg/restore tests (part 2)

skriss:
Again, probably easiest to review one commit at a time.

I found some bugs in the code that determines which restore item actions to run - I omitted the test cases that failed from this PR, but will be submitting another PR with the bug fix and the test cases that cover them.

@nrb @prydonius @carlisia


skriss:
found some more obsolete code, amended the last commit to remove it

pkg/restore/restore_new_test.go
```diff
@@ -955,6 +958,173 @@ func TestRestoreItems(t *testing.T) {
...
+			name:    "actions with selectors that don't match anything don't run for any resources",
+			restore: defaultRestore().Restore(),
+			backup:  defaultBackup().Backup(),
+			tarball: newTarWriter(t).
+				addItems("pods", test.NewPod("ns-1", "pod-1")).
+				addItems("persistentvolumeclaims", test.NewPVC("ns-2", "pvc-2")).
+				done(),
+			apiResources: []*test.APIResource{test.Pods(), test.PVCs(), test.PVs()},
+			actions: map[*recordResourcesAction][]string{
+				new(recordResourcesAction).ForNamespace("ns-1").ForResource("persistentvolumeclaims"): nil,
```
nrb: Should these values be `nil` or `""`?

pkg/restore/restore_new_test.go
```diff
@@ -721,6 +724,237 @@ func TestInvalidTarballContents(t *testing.T) {
...
+				test.Pods(),
+			},
+			want: []*test.APIResource{
+				test.Pods(
+					test.NewPod("ns-1", "pod-1", test.WithLabels("velero.io/backup-name", "backup-1", "velero.io/restore-name", "restore-1")),
+				),
+			},
+		},
+		{
+			name:    "object gets labeled with full backup and restore names when they're both longer than 63 characters",
```
nrb: Isn't this the shorter case, not the longer one?

skriss: yep, will update

pkg/restore/restore_new_test.go
skriss: they should be nil, since the action's `ids` field should never be initialized/should always stay nil

nrb: Ah, I see my confusion - it's a `nil` slice of strings, not a string.

skriss:
@prydonius would be great if you could review this today - I have a few other PRs backing up waiting for this to merge. Thanks!

PR #1607, opened 2019-06-26
Fix bugs in determining which restore item actions to run

skriss:
Note: only the last commit is unique to this PR, the previous ones are from #1606. I'll rebase this after #1606 gets merged.

There were a couple of bugs related to determining which restore item actions to run for a resource, all around the use of namespace selectors -- essentially they were being ignored.

@nrb @prydonius @carlisia


skriss:
This is waiting on #1606

skriss:
Alright, rebased and ready for merge on green

PR #1609, opened 2019-06-26
record PodVolumeBackup/Restore start and completion timestamps

### LINKED ISSUE #1518: add start/completion timestamps to PodVolumeBackups, PodVolumeRestores
Currently we don't have start/complete timestamps on the restic operation CRDs. This would be useful for visibility purposes.

prydonius:
adds startTimestamp and completionTimestamp fields to the
PodVolumeBackup and PodVolumeRestore status spec

closes #1518


pkg/controller/pod_volume_backup_controller.go
```diff
@@ -300,19 +307,14 @@ func (c *podVolumeBackupController) fail(req *velerov1api.PodVolumeBackup, msg s
...
 		r.Status.Message = msg
+		r.Status.CompletionTimestamp.Time = c.clock.Now()
 	}); err != nil {
 		log.WithError(err).Error("Error setting phase to Failed")
 		return err
 	}
 	return nil
 }
 
-func updatePhaseFunc(phase velerov1api.PodVolumeBackupPhase) func(r *velerov1api.PodVolumeBackup) {
```
prydonius: this was only being used in one place so I thought I'd remove it and just use an anonymous function to update both the phase and StartTimestamp field

skriss: 👍

skriss:
I'm comfortable not adding tests for this given that there's not a clear place to add them *to*.

Are you planning to cover `PodVolumeRestores` in this PR too, or separate?

prydonius:
@skriss yeah I was thinking of starting to add some tests for this controller as part of this, though let me know if you think that's not worth doing right now (or if there is already another issue for that).

As I was going to add tests, I thought I'd do PodVolumeRestores in a separate PR, but otherwise I can just do it in this PR.

skriss:
Gotcha - if you're up for adding some tests, go for it. Let me know if you need any input.

prydonius:
I'll give it a shot - will reach out if I need help with the fake clients, but hopefully can base it off of the other controller tests!

prydonius:
Hmm this controller will definitely be quite difficult to test, as it needs to look for volumes in `/host_pods` and runs restic commands. Do you think it makes sense to create a separate issue to look into testing this controller and leave this without tests for now?

prydonius:
Added the PodVolumeRestore implementation in this PR too

skriss:
> Do you think it makes sense to create a separate issue to look into testing this controller and leave this without tests for now?

That works for me - perhaps with some of the other restic work, we'll look at adding tests, but I don't think it needs to be coupled to this issue.

skriss:
@prydonius FYI - unfortunately, connecting a PR to an issue via ZenHub doesn't result in the issue getting closed when the PR is merged, so the best practice is to both (1) connect it via ZenHub for tracking purposes on the board, and (2) add the "Closes #<issue-num>" in the PR itself to get the auto-close.

skriss:
@nrb this one's ready for your review

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -173,9 +176,12 @@ func (c *podVolumeBackupController) processBackup(req *velerov1api.PodVolumeBack
...
 
 	// update status to InProgress
-	req, err = c.patchPodVolumeBackup(req, updatePhaseFunc(velerov1api.PodVolumeBackupPhaseInProgress))
+	req, err = c.patchPodVolumeBackup(req, func(r *velerov1api.PodVolumeBackup) {
+		r.Status.Phase = velerov1api.PodVolumeBackupPhaseInProgress
+		r.Status.StartTimestamp.Time = c.clock.Now()
+	})
 	if err != nil {
-		log.WithError(err).Error("Error setting phase to InProgress")
+		log.WithError(err).Error("Error setting backup StartTimestamp and phase to InProgress")
```
nrb: I'd update the message to say "podvolumebackup" for clarity.

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -258,9 +261,12 @@ func (c *podVolumeRestoreController) processRestore(req *velerov1api.PodVolumeRe
...
 
 	// update status to InProgress
-	req, err = c.patchPodVolumeRestore(req, updatePodVolumeRestorePhaseFunc(velerov1api.PodVolumeRestorePhaseInProgress))
+	req, err = c.patchPodVolumeRestore(req, func(r *velerov1api.PodVolumeRestore) {
+		r.Status.Phase = velerov1api.PodVolumeRestorePhaseInProgress
+		r.Status.StartTimestamp.Time = c.clock.Now()
+	})
 	if err != nil {
-		log.WithError(err).Error("Error setting phase to InProgress")
+		log.WithError(err).Error("Error setting startTimestamp and phase to InProgress")
```
nrb: See similar error in the backup file - I'd mention podvolumerestore in the error.

prydonius:
@nrb updated, ptal!

PR #1611, opened 2019-06-27
Refactor pkg/restore tests (part 3)

skriss:
Includes commits from #1606 -- will need a rebase after that merges.


skriss:
Rebased. I don't think this will conflict with #1607 but need to keep an eye on it.

skriss:
Note that the commented-out test cases in ff81e1f393246e7831ca658c10df6d81d6bb5f2d are currently failing due to bugs, which I'll fix and then uncomment the tests for in a subsequent PR.

PR #1612, opened 2019-06-28
properly filter additional items returned from restore item actions

skriss:

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -51,6 +51,7 @@ import (
 	"github.com/heptio/velero/pkg/label"
 	"github.com/heptio/velero/pkg/plugin/velero"
 	"github.com/heptio/velero/pkg/restic"
+	"github.com/heptio/velero/pkg/util/boolptr"
 	"github.com/heptio/velero/pkg/util/collections"
 	"github.com/heptio/velero/pkg/util/filesystem"
 	"github.com/heptio/velero/pkg/util/kube"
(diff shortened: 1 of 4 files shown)


pkg/restore/restore.go
```diff
@@ -787,6 +793,42 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
...
+			}).Info("Not restoring item because namespace is excluded")
+			return warnings, errs
+		}
+	} else {
+		if boolptr.IsSetToFalse(ctx.restore.Spec.IncludeClusterResources) {
+			ctx.log.WithFields(logrus.Fields{
+				"namespace":     obj.GetNamespace(),
+				"name":          obj.GetName(),
+				"groupResource": groupResource.String(),
+			}).Info("Not restoring item because it's cluster-scoped")
```
nrb: So, thinking about CSI implementation, on restore we'll absolutely want VolumeSnapshotContents, a cluster-scoped resource, to come back in, otherwise the VolumeSnapshot, a namespace-scoped resource, won't work.

This would seem counter to the PV/PVC behavior, which will create the PV even if the restore is created with `--include-namespaces`.

skriss: So this case is specifically around when the user specifies `--include-cluster-resources=false`, which is **not** the default (the default is `nil`, which roughly means "include any cluster resources that are related to the namespaced resources I'm restoring").

PR #1614, opened 2019-06-28
migrate restore PV tests

skriss:

Some of these test cases are a little verbose due to the sheer number of objects that need to be set up for each - open to any input there. I think they're clear enough.


pkg/restore/restore_new_test.go
```diff
@@ -1662,6 +1664,425 @@ func assertRestoredItems(t *testing.T, h *harness, want []*test.APIResource) {
...
+			}
+		}
+	}
+	withLabels := func(labels ...string) func(*corev1api.PersistentVolume) {
+		return func(pv *corev1api.PersistentVolume) {
+			test.WithLabels(labels...)(pv)
+		}
+	}
+
+	newPV := func(name string, opts ...func(*corev1api.PersistentVolume)) *corev1api.PersistentVolume {
```
prydonius: is there a difference between these helpers and the ones in pkg/test/resources?

skriss: The difference is that these take functional options that can specifically customize PV spec/status, rather than any generic API object's metadata. These should probably move into `pkg/test` and be combined with the other ones in some way, my brain's just too tired to figure out a reasonable way to do it at the moment :)

prydonius: Ah I see, yeah I can't see a better way to do this either right now. Maybe worth adding a comment to revisit this?

skriss: I'm looking at this again this morning. One option I'm considering is to turn these into functional options similar to what's already in `pkg/test/resources.go`, that take an arg of type `metav1.Object`, but to type-assert that object to a PV and to panic if it fails. That keeps the signature for using these simple but provides immediate feedback if they're used incorrectly. And since it's test code, I don't have any issue with just panicking. An example:

```go
func WithReclaimPolicy(policy corev1.PersistentVolumeReclaimPolicy) func(obj metav1.Object) {
	return func(obj metav1.Object) {
		pv, ok := obj.(*corev1.PersistentVolume)
		if !ok {
			panic("WithReclaimPolicy is only valid for persistent volumes")
		}

		pv.Spec.PersistentVolumeReclaimPolicy = policy
	}
}
```

Thoughts?

skriss: Went ahead and made this change, see latest commit.

prydonius: Nice, yeah I was wondering if it would be possible to use a type assertion here. This is much nicer now!

skriss:
@nrb @carlisia PTAL.

PR #1616, opened 2019-07-01
validate pod volumes hostpath mount on restic server startup

### LINKED ISSUE #1557: Validate for hostPath mount for restic
As can be seen here: https://github.com/heptio/velero/blob/4e2e4cd5c434791ab865abcb858e2a6c053355ad/site/docs/master/restic.md#instructions, there a hostPath will be different depending on the backend.

In addition to documenting it, we should add a validation check and return an error when the hostPath is misconfigured.

prydonius:
Adds validation to check if the pod volumes hostpath mount contains directories for the pods on this node.

fixes #1557 

### DIFF
--- a/pkg/cmd/cli/restic/server.go
+++ b/pkg/cmd/cli/restic/server.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 the Velero contributors.
+Copyright 2019 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -28,6 +28,7 @@ import (
 	"github.com/spf13/cobra"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	kubeinformers "k8s.io/client-go/informers"
 	corev1informers "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
@@ -94,9 +95,8 @@ func newResticServer(logger logrus.FieldLogger, baseName string) (*resticServer,
 		return nil, errors.WithStack(err)
 	}
 
-	err = validatePodVolumesHostPath(kubeClient, "/host_pods/")
-	if err != nil {
-		return nil, errors.WithStack(err)
+	if err := validatePodVolumesHostPath(kubeClient, "/host_pods/"); err != nil {
+		return nil, err
 	}
 
 	veleroClient, err := clientset.NewForConfig(clientConfig)
@@ -204,14 +204,14 @@ func (s *resticServer) run() {
 func validatePodVolumesHostPath(kubeClient kubernetes.Interface, podVolumesPath string) error {
 	files, err := ioutil.ReadDir(podVolumesPath)
 	if err != nil {
-		return errors.WithStack(err)
+		return errors.Wrap(err, "could not read pod volumes host path")
 	}
 
 	// create a map of directory names inside the podVolumesPath
-	dirs := map[string]bool{}
+	dirs := sets.NewString()
 	for _, f := range files {
 		if f.IsDir() {
-			dirs[f.Name()] = true
+			dirs.Insert(f.Name())
 		}
 	}
 
@@ -229,9 +229,10 @@ func validatePodVolumesHostPath(kubeClient kubernetes.Interface, podVolumesPath 
 			dirName = hash
 		}
 
-		if _, ok := dirs[dirName]; !ok {
+		if !dirs.Has(dirName) {
 			return errors.WithStack(fmt.Errorf("could not find volumes for pod %s/%s in hostpath, check if the pod volumes hostPath mount is correct", pod.GetNamespace(), pod.GetName()))
 		}
 	}
+
 	return nil
 }
--- a/pkg/cmd/cli/restic/server.go
+++ b/pkg/cmd/cli/restic/server.go
@@ -219,7 +219,7 @@ func (s *resticServer) validatePodVolumesHostPath() error {
 		}
 	}
 
-	pods, err := s.kubeClient.CoreV1().Pods("").List(metav1.ListOptions{FieldSelector: fmt.Sprintf("spec.nodeName=%s", os.Getenv("NODE_NAME"))})
+	pods, err := s.kubeClient.CoreV1().Pods("").List(metav1.ListOptions{FieldSelector: fmt.Sprintf("spec.nodeName=%s,status.phase=Running", os.Getenv("NODE_NAME"))})
 	if err != nil {
 		return errors.WithStack(err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/cmd/cli/restic/server.go
```diff
@@ -191,3 +198,40 @@ func (s *resticServer) run() {
...
+		dirName := string(pod.GetUID())
+
+		// if the pod is a mirror pod, the directory name is the hash value of the
+		// mirror pod annotation
+		if hash, ok := pod.GetAnnotations()[v1.MirrorPodAnnotationKey]; ok {
+			dirName = hash
+		}
+
+		if _, ok := dirs[dirName]; !ok {
+			return errors.WithStack(fmt.Errorf("could not find volumes for pod %s/%s in hostpath, check if the pod volumes hostPath mount is correct", pod.GetNamespace(), pod.GetName()))
```
skriss: I'm wondering if instead of returning immediately the first time we don't find a pod, we should collect all the pods that aren't found, and then return an error listing all of them? WDYT?

pkg/cmd/cli/restic/server_test.go
```diff
@@ -0,0 +1,98 @@
+/*
+Copyright 2018 the Velero contributors.
```
skriss: 2019

pkg/cmd/cli/restic/server_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes/fake"
+
+	"github.com/heptio/velero/pkg/test"
+)
+
+func Test_validatePodVolumesHostPath(t *testing.T) {
```
skriss: TestValidatePodVolumesHostPath

carlisia: So, when we use a generated test from VScode, that's how it generates a test name for private functions. I sorta like having the ability to distinguish tests for exported vs non-exported functions. Could we adopt this style?

skriss: I'm open to it if the team prefers it in general

pkg/cmd/cli/restic/server.go
```diff
@@ -92,6 +94,11 @@ func newResticServer(logger logrus.FieldLogger, baseName string) (*resticServer,
 		return nil, errors.WithStack(err)
 	}
 
+	err = validatePodVolumesHostPath(kubeClient, "/host_pods/")
```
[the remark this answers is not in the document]
prydonius: Ah I used a parameter here so that I could change the directory in the test. Another way to do it would be to use a var in the package and change it in the test. Is there any other way I can change the value in the test without using a var?

pkg/cmd/cli/restic/server.go
```diff
@@ -191,3 +198,40 @@ func (s *resticServer) run() {
...
+// validatePodVolumesHostPath validates that the podVolumesPath contains a
+// directory for each Pod running on this node
+func validatePodVolumesHostPath(kubeClient kubernetes.Interface, podVolumesPath string) error {
+	files, err := ioutil.ReadDir(podVolumesPath)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	// create a map of directory names inside the podVolumesPath
+	dirs := map[string]bool{}
```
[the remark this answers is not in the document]
prydonius: oh nice! will use that

pkg/cmd/cli/restic/server.go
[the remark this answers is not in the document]
carlisia: Hm, I see. One way to address this is to have a struct that would contain both elements. That would be preferable to passing a hard-coded string imo.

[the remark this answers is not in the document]
carlisia: But now that I said that it sounds weird... 🤷‍♀

pkg/cmd/cli/restic/server_test.go
nrb: I'm not opposed.

pkg/cmd/cli/restic/server.go
prydonius: So actually I was thinking that it might not even make sense to display the name of the Pod volume that's missing, as most likely the issue is that the host mount is wrong. It might be less confusing to just have the error here be "unexpected directory structure for pod volumes mount, check if the hostPath mount is correct".

After discussing at standup, @skriss suggested debug logging each missing pod volume too in case a user is trying to debug why the host mount is wrong.

pkg/cmd/cli/restic/server.go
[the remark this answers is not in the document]
skriss: another option: use a fake for the file system so you don't have to change the path -- see https://github.com/heptio/velero/blob/master/pkg/util/test/fake_file_system.go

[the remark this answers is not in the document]
prydonius: @skriss oh great, I was looking for something like that! That's a much better approach

prydonius:
@skriss @carlisia updated PTAL!

prydonius:
@carlisia good catch, I accidentally updated server.go instead of server_test.go, fixed!

carlisia:
👍

pkg/cmd/cli/restic/server.go
```diff
@@ -191,3 +202,50 @@ func (s *resticServer) run() {
...
+		}
+	}
+
+	pods, err := s.kubeClient.CoreV1().Pods("").List(metav1.ListOptions{FieldSelector: fmt.Sprintf("spec.nodeName=%s", os.Getenv("NODE_NAME"))})
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	valid := true
+	for _, pod := range pods.Items {
```
skriss: I'm thinking that we might want to ignore pods that don't currently have a phase of `Running`, since there could be pods that have been assigned a node name by the scheduler but haven't yet been initialized on the node, that would cause this check to fail.

pkg/cmd/cli/restic/server.go
```diff
@@ -191,3 +202,50 @@ func (s *resticServer) run() {
...
+			valid = false
+			s.logger.WithFields(logrus.Fields{
+				"pod":  fmt.Sprintf("%s/%s", pod.GetNamespace(), pod.GetName()),
+				"path": "/host_pods/" + dirName,
+			}).Debug("could not find volumes for pod in host path")
+		}
+	}
+
+	if !valid {
+		return errors.New("unexpected directory structure for pod volumes path, check if the pod volumes hostPath mount is correct")
```
skriss: This volume is named `host-pods` in the daemonset, so maybe something like "unexpected directory structure for host-pods volume, ensure that the host-pods volume corresponds to the pods subdirectory of the kubelet root directory" (feel free to wordsmith)

pkg/cmd/cli/restic/server.go
```diff
@@ -191,3 +202,50 @@ func (s *resticServer) run() {
...
+
+	// create a map of directory names inside the pod volumes path
+	dirs := sets.NewString()
+	for _, f := range files {
+		if f.IsDir() {
+			dirs.Insert(f.Name())
+		}
+	}
+
+	pods, err := s.kubeClient.CoreV1().Pods("").List(metav1.ListOptions{FieldSelector: fmt.Sprintf("spec.nodeName=%s,status.phase=Running", os.Getenv("NODE_NAME"))})
```
prydonius: unfortunately there isn't a great way to unit test the field selector (https://github.com/kubernetes/client-go/issues/326). Seems like it might be possible to mimic the functionality by creating a custom reactor, but then that won't really test that this field selector is correct.

prydonius:
rebased

PR #1621, opened 2019-07-02
add restore item action to change PVC/PV storage class name

### LINKED ISSUE #595: Decide if/how to allow users to change the storageclass of a PVC/PV when restoring
For example. if the original PV was hdd, and the user wants to switch to ssd. This might be part of the more generic #474.

skriss:

Fixes #595 

Since we've been hearing regular demand for this, I put together a plugin that can update a PV or PVC's storage class. It fetches old->new storage class mappings from a configmap, and it's a no-op if that configmap does not exist or is empty.  I thought about making this an out-of-tree plugin, but I didn't really see any reason not to keep it in-tree given that it's a no-op if the corresponding ConfigMap doesn't exist. Interested in your thoughts here.

Working on adding unit tests, but wanted to get input on approach first.

cc @nrb @carlisia @prydonius

### DIFF
--- a/pkg/cmd/server/plugin/plugin.go
+++ b/pkg/cmd/server/plugin/plugin.go
@@ -163,6 +163,10 @@ func newChangeStorageClassRestoreItemAction(f client.Factory) veleroplugin.Handl
 			return nil, err
 		}
 
-		return restore.NewChangeStorageClassAction(logger, client.CoreV1().ConfigMaps(f.Namespace())), nil
+		return restore.NewChangeStorageClassAction(
+			logger,
+			client.CoreV1().ConfigMaps(f.Namespace()),
+			client.StorageV1().StorageClasses(),
+		), nil
 	}
 }
--- a/pkg/cmd/server/plugin/plugin.go
+++ b/pkg/cmd/server/plugin/plugin.go
@@ -52,8 +52,8 @@ func NewCommand(f client.Factory) *cobra.Command {
 				RegisterRestoreItemAction("velero.io/restic", newResticRestoreItemAction(f)).
 				RegisterRestoreItemAction("velero.io/service", newServiceRestoreItemAction).
 				RegisterRestoreItemAction("velero.io/serviceaccount", newServiceAccountRestoreItemAction).
-				RegisterRestoreItemAction("velero.io/addPVCFromPod", newAddPVCFromPodRestoreItemAction).
-				RegisterRestoreItemAction("velero.io/addPVFromPVC", newAddPVFromPVCRestoreItemAction).
+				RegisterRestoreItemAction("velero.io/add-pvc-from-pod", newAddPVCFromPodRestoreItemAction).
+				RegisterRestoreItemAction("velero.io/add-pv-from-pvc", newAddPVFromPVCRestoreItemAction).
 				RegisterRestoreItemAction("velero.io/change-storageclass", newChangeStorageClassRestoreItemAction(f)).
 				Serve()
 		},
--- a/pkg/cmd/server/plugin/plugin.go
+++ b/pkg/cmd/server/plugin/plugin.go
@@ -54,6 +54,7 @@ func NewCommand(f client.Factory) *cobra.Command {
 				RegisterRestoreItemAction("velero.io/serviceaccount", newServiceAccountRestoreItemAction).
 				RegisterRestoreItemAction("velero.io/addPVCFromPod", newAddPVCFromPodRestoreItemAction).
 				RegisterRestoreItemAction("velero.io/addPVFromPVC", newAddPVFromPVCRestoreItemAction).
+				RegisterRestoreItemAction("velero.io/change-storageclass", newChangeStorageClassRestoreItemAction(f)).
 				Serve()
 		},
 	}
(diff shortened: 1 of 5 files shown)


pkg/restore/change_storageclass_action.go
```diff
@@ -0,0 +1,104 @@
...
+		}, nil
+	}
+
+	log.Infof("Updating item's storage class name to %s", newStorageClass)
+
+	if err := unstructured.SetNestedField(obj.UnstructuredContent(), newStorageClass, "spec", "storageClassName"); err != nil {
+		return nil, errors.Wrap(err, "unable to set item's spec.storageClassName")
+	}
+
+	return velero.NewRestoreItemActionExecuteOutput(obj), nil
```
prydonius: should the above returns also be using `NewRestoreItemActionExecuteOutput` instead of directly creating the struct?

skriss:
> lgtm. But wondering why change-storageclass is being used, instead of changeStorageclass, which seems more inline with the existing plugins (ex: actionNoun no dash).

Yeah - I assume you're referring to `velero.io/addPVCFromPod` and `velero.io/addPVFromPVC`? I was thinking about changing those to lowercase/hyphens -  the current style looks weird to me in that context since it's not a go identifier - WDYT?

nrb:
>One minor thought is that the ConfigMap can only store the mappings and won't be able to store any other config we might want to add in the future for this plugin

I view the ConfigMaps as being specific to plugins, and this plugin has a fairly narrow scope (changing one storage class to another). It's true that doing a ConfigMap per plugin would lead to a lot of ConfigMaps to manage, but for the time being I think it's best to do a 1:1 mapping between a plugin and ConfigMap.

pkg/restore/change_storageclass_action.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, errors.Wrap(err, "error getting item's spec.storageClassName")
+	}
+	if storageClass == "" {
+		log.Debug("Item has no storage class specified")
+		return &velero.RestoreItemActionExecuteOutput{
+			UpdatedItem: input.Item,
+		}, nil
+	}
+
+	newStorageClass, ok := config.Data[storageClass]
```
nrb: At the risk of complicating the plugin more, would it make sense to validate that the new storage class exists in the cluster?

nrb:
Approach looks good to me, only one question about the design.

carlisia:
> I was thinking about changing those to lowercase/hyphens - the current style looks weird to me in that context since it's not a go identifier - WDYT?

Yes, it didn't even occur to me, I was looking at that string "as if" it was a go identifier... just because it looked like one! I'm on board to changing all to not look like a go identifier. 👍

pkg/restore/change_storageclass_action.go
skriss: seems reasonable, will add validation

skriss:
Addressed PR comments and added unit tests - please take another look!

pkg/cmd/server/plugin/plugin.go
```diff
@@ -52,8 +52,8 @@ func NewCommand(f client.Factory) *cobra.Command {
 				RegisterRestoreItemAction("velero.io/restic", newResticRestoreItemAction(f)).
 				RegisterRestoreItemAction("velero.io/service", newServiceRestoreItemAction).
 				RegisterRestoreItemAction("velero.io/serviceaccount", newServiceAccountRestoreItemAction).
```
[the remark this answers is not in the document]
carlisia: I did that ^ wrong, please revert the string order :) 

Correct is: `s/serviceaccount/service-account`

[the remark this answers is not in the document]
skriss: Yeah, i considered that change - i'm fine making it

pkg/restore/change_storageclass_action_test.go
```diff
@@ -0,0 +1,394 @@
...
+				},
+			},
+			storageClass: &storagev1api.StorageClass{
+				ObjectMeta: objectMeta("", "storageclass-2"),
+			},
+			wantErr: errors.New("error getting storage class nonexistent-storage-class from API: storageclasses.storage.k8s.io \"nonexistent-storage-class\" not found"),
+		},
+		{
+			name: "when persistent volume claim's storage class is mapped to a nonexistent storage class, an error is returned",
+			pvOrPVC: &corev1api.PersistentVolumeClaim{
```
nrb: Is there a reason you're not using the `test` package's constructors for these?

skriss: No, good call - updated.

pkg/cmd/server/plugin/plugin.go
```diff
@@ -46,14 +46,15 @@ func NewCommand(f client.Factory) *cobra.Command {
 				RegisterVolumeSnapshotter("velero.io/gcp", newGcpVolumeSnapshotter).
 				RegisterBackupItemAction("velero.io/pv", newPVBackupItemAction).
 				RegisterBackupItemAction("velero.io/pod", newPodBackupItemAction).
-				RegisterBackupItemAction("velero.io/serviceaccount", newServiceAccountBackupItemAction(f)).
+				RegisterBackupItemAction("velero.io/service-account", newServiceAccountBackupItemAction(f)).
```
prydonius: will these renames be a breaking change for users already using these plugins?

skriss: No, they shouldn't be - the user doesn't currently interact with these names in any way.

skriss:
@carlisia @nrb @prydonius would appreciate another look here - all comments have been addressed.

PR #1626, opened 2019-07-03
Add pod-annotations CLI flag to the install command

### LINKED ISSUE #1581: CLI option for IAM Role in annotations
Following the [AWS Config](https://velero.io/docs/v1.0.0/aws-config/) instructions for using kube2iam enabled EKS clusters, we have to edit the Velero Deployment object after it is generated, either interactively through `kubectl edit deploy/velero` or through another means (creating and applying a patch for the deployment, using a mutating admission webhook to intercept the deployment); this is not great for deploying Velero in automation. 

It would be ideal if the `velero install ...` command provided an option to set the value of the `iam.amazonaws.com/role` annotation to a given value to generate the Deployment fully.

I would like to propose a workflow like this:

```shell
velero install --provider aws \
  --aws-iam-role velero-role \
  --secret-file ./credentials-velero \
...etc
```

that results in a Deployment containing the following: 

```yaml
---
apiVersion: apps/v1beta1
kind: Deployment
metadata:
    namespace: velero
    name: velero
spec:
    replicas: 1
    template:
        metadata:
            labels:
                component: velero
            annotations:
                iam.amazonaws.com/role: velero-role
```

If there is a better way to handle this situation, I would also be open to suggestions.

**Environment:**
AWS EKS with kube2iam

- Velero version (use `velero version`): 1.0.0
- Kubernetes version (use `kubectl version`): 1.10.13
- OS (e.g. from `/etc/os-release`): Amazon Linux 2 (CentOS)

tlkamp:
Adds a CLI flag, `--pod-annotations` to the `velero install` options to allow users to specify additional annotations for the Velero pod.

Fixes #1581


tlkamp:
@nrb sure thing. I'll push up the changes shortly.

tlkamp:
@nrb I've updated the DaemonSet to include the annotations. The annotations on the DaemonSet are only the ones specified by the user (the prometheus annotations are not included).

tlkamp:
@carlisia I pondered that myself! I think it's important to call out that annotations are applied to both Velero and Restic if specified using `--pod-annotations`. It could probably be worded better to get that point across. What do you think of:

"annotations to add to the Velero and Restic pods."

Or do you think that's still too specific?

tlkamp:
@skriss thanks! 

It should be all ready to go once the Travis build finishes. 😄

PR #1636, opened 2019-07-09
add design/ dir and doc template

skriss:

Adds a `design` directory and a doc template to be used going forward for design proposals for non-trivial items. The template is copied verbatim from Contour - it looks reasonable to me but we can tweak as needed.

cc @nrb @carlisia @prydonius

### DIFF
--- a/design/_template.md
+++ b/design/_template.md
@@ -1,0 +1,48 @@
+# Design proposal template (replace with your proposal's title)
+
+Status: {Draft,Accepted,Declined}
+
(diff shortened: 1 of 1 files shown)


skriss:
Open to suggestions - I'm not hugely concerned at the moment as we can always retroactively rename things if we need and I don't expect a huge volume.

carlisia:
Going to merge. If @prydonius has any add/change we can iterate.

PR #1649, opened 2019-07-11
fix link for hooks example

### LINKED ISSUE #1648: [Documentation] The link to hook example returns 404
Hi, it looks like the link to the nginx example in [this page](https://velero.io/docs/v1.0.0/hooks/) points to a nonexistent page. I think the correct URL should be [this](https://github.com/heptio/velero/blob/master/examples/nginx-app/with-pv.yaml). Thanks

skriss:

Fixes #1648

### DIFF
--- a/site/docs/v1.0.0/hooks.md
+++ b/site/docs/v1.0.0/hooks.md
@@ -77,5 +77,5 @@ velero backup logs nginx-hook-test | grep hookCommand
 
 
 [1]: api-types/backup.md
-[2]: examples/nginx-app/with-pv.yaml
+[2]: https://github.com/heptio/velero/blob/v1.0.0/examples/nginx-app/with-pv.yaml
 [3]: cloud-common.md
--- a/site/docs/master/hooks.md
+++ b/site/docs/master/hooks.md
@@ -77,5 +77,5 @@ velero backup logs nginx-hook-test | grep hookCommand
 
 
 [1]: api-types/backup.md
-[2]: examples/nginx-app/with-pv.yaml
+[2]: https://github.com/heptio/velero/blob/master/examples/nginx-app/with-pv.yaml
 [3]: cloud-common.md


site/docs/v1.0.0/hooks.md
```diff
@@ -77,5 +77,5 @@ velero backup logs nginx-hook-test | grep hookCommand
 
 
 [1]: api-types/backup.md
-[2]: examples/nginx-app/with-pv.yaml
+[2]: https://github.com/heptio/velero/blob/master/examples/nginx-app/with-pv.yaml
```
carlisia: Actually, should this not point to:?

https://github.com/heptio/velero/blob/v1.0.0/examples/nginx-app/with-pv.yaml

skriss: Yeah -- we can set this one to v1.0, but the problem is there's no way to automatically update these version-specific links as we cut new versions of the docs site - so, for example, when we cut v1.1, we'll copy the `master/` docs into v1.1, and without manually looking through all the docs links, we'll end up with the link pointing to `master` again. Open to ideas here.

skriss:
> I see version 0.11 and 0.10 have the same issue:

I don't see any reason to bother updating pre-1.0 docs going forward as a policy. Let me know if you disagree.

site/docs/v1.0.0/hooks.md
nrb: Can we not use the ``{{ site.latest }}` variable in the URLs?

So `https://github.com/heptio/velero/blob/{{ site.latest }}/examples/nginx-app/with-pv.yaml`.

nrb: That wouldn't actually solve the copy, problem, though...

jonasrosland: That does work @nrb, and could be used here.

carlisia: I would not be against having a dup `examples` folder inside `site`.

carlisia:
> I don't see any reason to bother updating pre-1.0 docs going forward as a policy. Let me know if you disagree.

Yes, that's fine. I forgot about this guideline.

site/docs/v1.0.0/hooks.md
skriss: Downside to using {{ site.latest }} is the link no longer works if the page is being viewed directly on GitHub.

We do also package the examples in the release tarball, so maybe we should just remove the link entirely and reference that, since it'll properly match whatever version the user is on.

carlisia: what of ppl who are using something like `brew` and not the tarball directly?

jonasrosland: Would it make sense to move the examples in under `site/docs`?

nrb: I'm fine with the `site` directory not working 100% under the GitHub UI, personally.

While we could refer to the examples in the tarball, that also requires the user to stop reading. go download it, extract it, and find the file.

Duplicating the examples under the `site` dir is likely to lead to drift unless we add automation around it. Moving it over entirely might be ok, but viewing the yaml file off of GitHub, as plaintext, isn't a great experience either.

Not really a fan of most of these options at the moment, but I also don't have anything better to offer.

skriss: If we can add a script to generate a new versioned set of docs from `master/` that handles updating GH links with appropriate versions then I think we'll be OK here. Let me look into that.

skriss: OK, see https://github.com/heptio/velero/pull/1662 to address https://github.com/heptio/velero/pull/1649#discussion_r302744191. If everyone's happy with that approach then I think this PR can get merged as-is now.

skriss:
Per https://github.com/heptio/velero/pull/1649#discussion_r303144248, this should be good to merge if everyone's OK with https://github.com/heptio/velero/pull/1662

skriss:
OK now that #1662 is done, we can merge this!

Issue #487, opened 2018-05-14
Properly handle multiple workers/controllers processing the same backup/restore


ncdc:
We need to account for a few things:

1. If 2 server pods run simultaneously, they both might get the same Backup or Restore in a New state. They would both attempt to change the state to InProgress, they would likely both succeed, resulting in undesirable behavior.
1. If a Backup or Restore is InProgress and the server crashes, we ideally need to be able to have the replacement server process pick up whatever was in progress instead of having it linger.

rosskukulinski:
From a use-ability perspective, I think this is a particularly important issue for us to tackle.  There are a number of reasons why two Server pods might be running simultaneously and we need to handle that gracefully.

In addition, there's lots of ways an InProgress backup get stuck because a server exits.  Again, we need to handle this gracefully.  Metrics in #84 may be able to help (gauge displaying the current number of backups InProgress or Failed) visualize these issues, but it won't fix them.

ncdc:
A backup/restore won't get stuck during normal shutdown operations because the Ark server waits for all in-progress work to complete before terminating. If that work takes too long and exceeds the pod's deletion grace period, then Kubernetes would forcefully kill the container, and that would interrupt the in-progress work before it had a chance to finish.

There are, however, plenty of situations where the Ark server could exit while doing work:
- exceeding the grace period on a normal shutdown
- OOM killed
- a bug of some sort that causes a crash

This is definitely something we need to handle

rosskukulinski:
This needs a quick test (from code) to trigger:

* get a backup thats new
* patch it to inprogress
* patch again to inprogress (this should fail, since already in progress)

ncdc:
I had a thought about how to implement this.

Each ark server process is assigned a unique identifier - the name of the pod (we can get the value using the downward API and pass it to `ark server` as a flag).

Each controller worker is also assigned a unique identifier.

When a new item (backup, restore) is processed by a controller, the first thing the controller attempts to do is set `status.arkServerID` and `status.workerID`. Assuming that succeeds without a conflict, the worker can proceed to do its work.

When a worker sees an InProgress item, it checks `status.arkServerID`
- If there are no running pods matching that name, the worker resets the status back to New for reprocessing.
- If there is a running pod matching that name, and it matches this ark server, reset the status to New if there are no active workers matching `status.workerID`

The controller would also need add event handlers for pods. Upon a change, we'd want to reevaluate all InProgress items to see if they need to be taken over.

[...]

PR #1653, opened 2019-07-11
worker pods design doc

### LINKED ISSUE #2601: design for running multiple velero backups/restores concurrently

skriss:

closes #2601

This is a draft design doc for #487. Please provide feedback!

@nrb @carlisia @prydonius

### DIFF
--- a/design/worker-pods.md
+++ b/design/worker-pods.md
@@ -1,0 +1,161 @@
+# Worker Pods for Backup, Restore
+
+Status: Draft
+
+This document proposes a new approach to executing backups and restores where each operation is run in its own "worker pod" rather than in the main Velero server pod. This approach has significant benefits for concurrency, scalability, and observability.
+
+## Goals
+
+- Enable multiple backups/restores to be run concurrently by running each operation in its own worker pod.
+- Improve Velero's scalability by distributing work across multiple pods.
+- Allow logs for in-progress backups/restores to be streamed by the user.
+
+## Non Goals
+
+- Creating CronJobs for scheduled backups.
+
+## Background
+
+Today, the Velero server runs as a Kubernetes deployment with a single replica. The deployment's one pod runs all of the Velero controllers, including the backup and restore controllers. The backup and restore controllers are responsible for doing all of the work associated with backups/restores, including interacting with the Kubernetes API server, writing to or reading from object storage, creating or restoring disk snapshots, etc. 
+
+Because each Velero controller is configured to run with a single worker, only one backup or one restore can be executed at a time. This can lead to performance issues if a large number of backups/restores are triggered within a short time period. This issue will only be exacerbated as Velero moves towards a multi-tenancy model and better supports on-premises snapshot + backup scenarios.
+
+## High-Level Design
+
+Velero controllers will no longer directly execute backup/restore logic themselves (note: the rest of this document will refer only to backups, for brevity, but it applies equally to restores). Instead, when the controller is informed of a new backup custom resource, it will immediately create a new worker pod which is responsible for end-to-end processing of the backup, including validating the spec, scraping the Kubernetes API server for resources, triggering persistent volume snapshots, writing data to object storage, and updating the custom resource's status as appropriate.
+
+A worker pod will be given a deterministic name based on the name of the backup it's executing. This will prevent Velero from inadvertently creating multiple worker pods for the same backup, since any subsequent attempts to create a pod with the specified name will fail.
+
+This design trivially enables running multiple backups concurrently, as each one runs in its own isolated pod and the Velero server's backup controller does not need to wait for the backup to complete before spawning a worker pod for the next one. Additionally, Velero becomes much more scalable, as the resource requirements for the Velero server itself are largely unaffacted by the number of backups. Instead, Velero's scalability becomes limited only by the total amount of resources available in the cluster to process worker pods.
+
+## Detailed Design
+
+### Commands for worker pods
+
+A new hidden command will be added to the velero binary, `velero backup run BACKUP_NAME`. This command will be used solely by the worker pod. The vast majority of the backup processing logic currently in the backup controller will be moved into this new command.
+
+`velero backup run` will accept the following flags, which will be passed down from the Velero server:
+
+```bash
+--client-burst
+--client-qps
+--default-backup-storage-location
+--default-backup-ttl
+--default-volume-snapshot-locations
+--log-level
+--restic-timeout
+```
+
+`velero restore run` will accept the following flags:
+
+```bash
+--client-burst
+--client-qps
+--log-level
+--restic-timeout
+--restore-resource-priorities
+--terminating-resource-timeout
+```
+
+### Controller changes
+
+Most of the logic currently in the backup controller will be moved into the `velero backup run` command. The controller will be updated so that when it's informed of a new backup custom resource, it attempts to create a new worker pod for the backup. If the create is successful or a worker pod already exists with the specified name, no further action is needed and the backup will not be re-added to the controller's work queue.
+
+Additionally, the backup controller will be given a resync function that periodically re-enqueues all backups, to ensure that no new backups are missed.
+
+The spec for the worker pod will look like the following:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    component: velero
+    velero.io/backup-name: <BACKUP_NAME>
+  name: backup-<BACKUP_NAME>
+  namespace: velero
+  ownerReferences:
+  - apiVersion: velero.io/v1
+    controller: true
+    kind: Backup
+    name: <BACKUP_NAME>
+    uid: <BACKUP_UID>
+spec:
+  containers:
+  - args:
+    - <BACKUP_NAME>
+    - --client-burst=<VAL>
+    - --client-qps=<VAL>
+    - --default-backup-storage-location=<VAL>
+    - --default-backup-ttl=<VAL>
+    - --default-volume-snapshot-locations=<VAL>
+    - --log-level=<VAL>
+    - --restic-timeout=<VAL>
+    command:
+    - /velero
+    - backup
+    - run
+    env:
+    - name: VELERO_SCRATCH_DIR
+      value: /scratch
+    - name: AWS_SHARED_CREDENTIALS_FILE
+      value: /credentials/cloud
+    image: <VELERO_SERVER_IMAGE>
+    imagePullPolicy: IfNotPresent
+    name: velero
+    volumeMounts:
+    - mountPath: /plugins
+      name: plugins
+    - mountPath: /scratch
+      name: scratch
+    - mountPath: /credentials
+      name: cloud-credentials
+    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+      name: velero-token-plwsw
+      readOnly: true
+  serviceAccount: velero
+  volumes:
+  - emptyDir: {}
+    name: plugins
+  - emptyDir: {}
+    name: scratch
+  - name: cloud-credentials
+    secret:
+      defaultMode: 420
+      secretName: cloud-credentials
+  - name: velero-token-plwsw
+    secret:
+      defaultMode: 420
+      secretName: velero-token-plwsw
+```
+
+### Updates to `velero backup logs`
+
+The `velero backup logs` command will be modified so that if a backup is in progress, the logs are gotten from the worker pod's stdout (essentially proxying to `kubectl -n velero logs POD_NAME`). A  `-f/--follow` flag will be added to allow streaming of logs for an in-progress backup.
+
+Once a backup is complete, the logs will continue to be uploaded to object storage, and `velero backup logs` will still fetch them from there.
+
+### Changes to backup deletion controller
+
+The Velero server keeps a record of current in-progress backups and disallows them from being deleted. This will need to be updated to account for backups running in worker pods. The most sensible approach is for the backup deletion controller to first check the backup's status via the API, and to allow deletion to proceed if it's not new/in-progress. If the backup is new or in-progress, the controller can check to see if a worker pod is running for the backup. If so, the deletion is disallowed because the backup is actively being processed. If there is no worker pod for the backup, the backup can be deleted as it's not actively being processed.
+
+### Open items
+
+- It probably makes sense to use Kubernetes Jobs to control worker pods, rather than directly creating "bare pods". The Job will ensure that a worker pod successfully runs to completion.
+- Currently, restic repository lock management is handled by an in-process lock manager in the Velero server. In order for backups/restores to safely run concurrently, the design for restic lock management needs to change. There is an [open issue](https://github.com/heptio/velero/issues/1540) for this which is currently scoped for the v1.1 release.
+- There are several prometheus metrics that are emitted as part of the backup process. Since backups will no longer be running in the Velero server, we need to find a way to expose those values. One option is to store any value that would feed a metric as a field on the backup's `status`, and to have the Velero server scrape values from there for completed backups.
+- Over time, many completed worker pods will exist in the `velero` namespace. We need to consider whether this poses any issue and whether we should garbage-collect them more aggressively.
+
+## Alternatives Considered
+
+### Enable >1 worker per controller
+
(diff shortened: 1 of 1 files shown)


skriss:
> This also has implications for developer workflow - in order to do basic backup/restore operations, it will no longer be sufficient to run only a local server. An image with the relevant changes must also be uploaded before they can be tested.

Yes, good point.

I definitely want to debate if this is the right model or not - it has several benefits but also potentially some drawbacks (this being one). I'm not yet at the point where I'm 100% on it.

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+      secretName: cloud-credentials
+  - name: velero-token-plwsw
+    secret:
+      defaultMode: 420
+      secretName: velero-token-plwsw
+```
+
+### Updates to `velero backup logs`
+
+The `velero backup logs` command will be modified so that if a backup is in progress, the logs are gotten from the worker pod's stdout (essentially proxying to `kubectl -n velero logs POD_NAME`). A  `-f/--follow` flag will be added to allow streaming of logs for an in-progress backup.
```
prydonius: Even after completion, the Pod will be kept around until it's deleted, so we can choose to always pull the logs from the Pod as long as it is still around.

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+
+Once a backup is complete, the logs will continue to be uploaded to object storage, and `velero backup logs` will still fetch them from there.
+
+### Changes to backup deletion controller
+
+The Velero server keeps a record of current in-progress backups and disallows them from being deleted. This will need to be updated to account for backups running in worker pods. The most sensible approach is for the backup deletion controller to first check the backup's status via the API, and to allow deletion to proceed if it's not new/in-progress. If the backup is new or in-progress, the controller can check to see if a worker pod is running for the backup. If so, the deletion is disallowed because the backup is actively being processed. If there is no worker pod for the backup, the backup can be deleted as it's not actively being processed.
```
prydonius: Does this mean that the deletion request is marked as disallowed and never reconsidered, or will the deletion request just be re-enqueued at a later time and eventually cause the backup to be deleted after it has been completed?

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Open items
+
+- It probably makes sense to use Kubernetes Jobs to control worker pods, rather than directly creating "bare pods". The Job will ensure that a worker pod successfully runs to completion.
```
prydonius: One of the benefits of Jobs over Pods is being able to configure the maximum number of retries, but if we don't need to retry failed backups then a Pod should suffice. Do we retry backups today?

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+- Currently, restic repository lock management is handled by an in-process lock manager in the Velero server. In order for backups/restores to safely run concurrently, the design for restic lock management needs to change. There is an [open issue](https://github.com/heptio/velero/issues/1540) for this which is currently scoped for the v1.1 release.
+- There are several prometheus metrics that are emitted as part of the backup process. Since backups will no longer be running in the Velero server, we need to find a way to expose those values. One option is to store any value that would feed a metric as a field on the backup's `status`, and to have the Velero server scrape values from there for completed backups.
+- Over time, many completed worker pods will exist in the `velero` namespace. We need to consider whether this poses any issue and whether we should garbage-collect them more aggressively.
```
prydonius: Just FYI there is an alpha feature to clean up Jobs/Pods after some amount of time from completion (https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#ttl-mechanism-for-finished-jobs). There's some work to be done to take this to beta, but no one's picked it up right now. So not clear when it will get promoted, but worth keeping in mind.

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Run multiple replicas of the Velero server
+
+Another option is to increase the number of replicas in the Velero deployment, so that multiple instances of the Velero server pod are running. This would require adding some sort of "leader-election" process for each backup to ensure that only one replica was processing it. It also would result in multiple replicas of the non-core controllers (backup deletion controller, GC controller, download request controller, etc.), which would require modifications to ensure there wasn't any duplicate processing. Additionally, this approach also doesn't easily enable log streaming for in-progress backups.
+
+## Security Considerations
+
+Each worker pod will be created in the `velero` namespace, will use the `velero` service account for identity, and will have the `cloud-credentials` secret mounted as a volume. As a result the worker pods will have the same level of privilege as the Velero server pod.
```
prydonius: Presumably we could lower the privileges of the Velero server if it now only needs to create these Jobs/Pods and no longer needs cluster-admin access for backing up all the resources.

design/worker-pods.md
nrb: Good to know about that alpha feature. I think, absent that feature, we'd be ok cleaning up pods when a `Backup`'s TTL is met.

design/worker-pods.md
nrb: The Velero server would be reduced, but Velero overall would still need the cluster-admin access.

So it does change, but I think in terms of the overall set of permissions, it doesn't change things.

That said, this design _may_ enable the server to instantiate a worker pod with the requesting user's same permissions, and the backup flow within that worker pod could differentiate between not found and permission denied errors to handle logging. This requires more detailed thought, though, as even then access to cluster resources is needed if we continue to do things like follow a Pod->PV->PVC.

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+- There are several prometheus metrics that are emitted as part of the backup process. Since backups will no longer be running in the Velero server, we need to find a way to expose those values. One option is to store any value that would feed a metric as a field on the backup's `status`, and to have the Velero server scrape values from there for completed backups.
+- Over time, many completed worker pods will exist in the `velero` namespace. We need to consider whether this poses any issue and whether we should garbage-collect them more aggressively.
+
+## Alternatives Considered
+
+### Enable >1 worker per controller
+
+One alternative option is to simply increase the number of workers being used per controller. This would, in theory, allow multiple backups/restores to be processed concurrently by the Velero server. The downsides to this option are that (1) this doesn't do anything to improve Velero's scalability, since all of the work is still being done within a single pod/process, and (2) it doesn't easily enable log streaming for in-progress backups.
+
+### Run multiple replicas of the Velero server
```
nrb: Have you given thought to leader election at server startup? That was my initial thought, not per-backup.

The problem with per-backup log streaming is still there, though.

design/worker-pods.md
nrb: As of today, we just fail backups and restores. Retries may be helpful; I'd appreciate community feedback on that.

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Open items
+
+- It probably makes sense to use Kubernetes Jobs to control worker pods, rather than directly creating "bare pods". The Job will ensure that a worker pod successfully runs to completion.
+- Currently, restic repository lock management is handled by an in-process lock manager in the Velero server. In order for backups/restores to safely run concurrently, the design for restic lock management needs to change. There is an [open issue](https://github.com/heptio/velero/issues/1540) for this which is currently scoped for the v1.1 release.
+- There are several prometheus metrics that are emitted as part of the backup process. Since backups will no longer be running in the Velero server, we need to find a way to expose those values. One option is to store any value that would feed a metric as a field on the backup's `status`, and to have the Velero server scrape values from there for completed backups.
+- Over time, many completed worker pods will exist in the `velero` namespace. We need to consider whether this poses any issue and whether we should garbage-collect them more aggressively.
+
+## Alternatives Considered
```
nrb: Another alternative may be a static set of workers that process some queue of items. However, at the moment I would prefer pod-per-item due to the simplicity in implementing the streaming logs.

design/worker-pods.md
prydonius: If the Backup is deleted when the TTL is met, an easy way to do this would be to just have the Pod/Job's ownerReference set to the Backup.

design/worker-pods.md
prydonius: Yeah you're right that the workers will still need cluster-admin, though this still feels like an improvement as the long-running server process no longer needs it.

> That said, this design may enable the server to instantiate a worker pod with the requesting user's same permissions

Interesting, how do you see this working? Would the user need to specify the ServiceAccount to use for the worker?

design/worker-pods.md
skriss: yeah, I was thinking we'd use owner refs here to get cascading deletion

design/worker-pods.md
nrb: That's a good question. It could be exposed for them to set, or maybe it's created on the fly to match the current user? That may be tricky to get right, so likely less desirable.

design/worker-pods.md
skriss: Yeah, it may be easiest to always do that if the pod is still around, otherwise fall back to the archived logs in object storage.

design/worker-pods.md
skriss: I believe that currently they're disallowed, marked as "processed", and never re-processed. We can certainly look at changing that.

design/worker-pods.md
skriss: In my mind there are two different scenarios to consider here:

1. The backup completes, with a final phase of "Failed" or "PartiallyFailed". I would not currently consider this grounds for a retry.
2. The worker pod gets evicted or otherwise forcibly terminated before the backup has had a chance to complete. In this case I *would* want to retry.

design/worker-pods.md
skriss: What would the non-leaders do in that scenario?

nrb: They would become workers, similar in function to what's proposed here. However, that means we can still only process `replica` number of backups at once.

skriss: Ah, so leader assigns a worker to a backup, and worker processes backups?

nrb: Yeah, that's what I was thinking.

design/worker-pods.md
nrb: Would it be possible that the logs grow too large for the cluster to keep around?

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Open items
+
+- It probably makes sense to use Kubernetes Jobs to control worker pods, rather than directly creating "bare pods". The Job will ensure that a worker pod successfully runs to completion.
+- Currently, restic repository lock management is handled by an in-process lock manager in the Velero server. In order for backups/restores to safely run concurrently, the design for restic lock management needs to change. There is an [open issue](https://github.com/heptio/velero/issues/1540) for this which is currently scoped for the v1.1 release.
+- There are several prometheus metrics that are emitted as part of the backup process. Since backups will no longer be running in the Velero server, we need to find a way to expose those values. One option is to store any value that would feed a metric as a field on the backup's `status`, and to have the Velero server scrape values from there for completed backups.
```
nrb: I think we should do some more research on how distributed metrics are tracked with prometheus. It may be wishful thinking, but we can't be the only ones to want to do this.

I also think the public API on our `metrics` package can be sufficient for masking most of this from the backup code, though the logic inside may change to route the metric correctly.

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Open items
+
+- It probably makes sense to use Kubernetes Jobs to control worker pods, rather than directly creating "bare pods". The Job will ensure that a worker pod successfully runs to completion.
+- Currently, restic repository lock management is handled by an in-process lock manager in the Velero server. In order for backups/restores to safely run concurrently, the design for restic lock management needs to change. There is an [open issue](https://github.com/heptio/velero/issues/1540) for this which is currently scoped for the v1.1 release.
```
carlisia: `v1.x` release now.

design/worker-pods.md
carlisia: Could this be an appropriate place to delete the associated pod?

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+
+## High-Level Design
+
+Velero controllers will no longer directly execute backup/restore logic themselves (note: the rest of this document will refer only to backups, for brevity, but it applies equally to restores). Instead, when the controller is informed of a new backup custom resource, it will immediately create a new worker pod which is responsible for end-to-end processing of the backup, including validating the spec, scraping the Kubernetes API server for resources, triggering persistent volume snapshots, writing data to object storage, and updating the custom resource's status as appropriate.
+
+A worker pod will be given a deterministic name based on the name of the backup it's executing. This will prevent Velero from inadvertently creating multiple worker pods for the same backup, since any subsequent attempts to create a pod with the specified name will fail.
```
carlisia: If the pod name will match the given backup name, I suppose before creating the pod we will first check if that backup already exists in storage? If *before*, then we would be able to make use of the worker pod itself for all validations except this one.

design/worker-pods.md
```diff
@@ -0,0 +1,161 @@
...
+
+This design trivially enables running multiple backups concurrently, as each one runs in its own isolated pod and the Velero server's backup controller does not need to wait for the backup to complete before spawning a worker pod for the next one. Additionally, Velero becomes much more scalable, as the resource requirements for the Velero server itself are largely unaffacted by the number of backups. Instead, Velero's scalability becomes limited only by the total amount of resources available in the cluster to process worker pods.
```
carlisia: Is remaining resource available something we would need to check for before initiating a backup? What happens if there's not enough resources to complete a backup? (I suppose the same questions would apply with todays design but the scenario is much less likely to happen).

design/worker-pods.md
nrb: Doing the check before making the pod makes sense to me. Fairly sure we already check to see if the backup exists before starting one now, so this would go after that check.

skriss:
crazy idea: we could do away with the restic daemonset, move the PVB/PVR controllers into the core velero server process, and have them dispatch worker pods to run restic backups/restores on the appropriate nodes. 

That would even enable us to back up hostPath volumes, since we would know at the time of the worker pod creation which volumes to back up.

nrb:
Doesn't sound that crazy to me. If I'm following, would this mean the `--use-restic` flag in `velero install` is essentially a no-op as the restic controllers would just always run?

skriss:
Yeah, probably. We *could* disable the PVB/PVR controllers by default if we wanted to, and only enable them if `--use-restic` is specified. Not sure yet.

nrb:
In the abstract, I like the idea of moving the restic work from the DaemonSet to, well, workers. It logically unifies the "work" vs "scheduling" concepts.

prydonius:
+1 to running restic operations in worker pods, that's a great idea!

> Yeah, probably. We _could_ disable the PVB/PVR controllers by default if we wanted to, and only enable them if `--use-restic` is specified. Not sure yet.

IMO we should definitely disable them by default until we consider restic GA, and even once restic is GA it might make sense to disable them especially if we don't see restic being used in conjunction with volume snapshots. (good usecase for feature flags)

skriss:
Another thing to think about: we probably need to synchronize execution of hooks. It's probably not ok to run `fsfreeze` twice for the same pod because two backups are occurring simultaneously.

design/worker-pods.md
skriss: potentially relevant: https://prometheus.io/docs/practices/pushing/

carlisia:
👀

design/worker-pods.md
alaypatel07: @carlisia I am still new to the project, but if the cluster did not have enough resources for the backup with this new design, I guess it will be reported in worker pod status? Assuming the worker pod would be unschedulable, instead of preemptively checking for cluster resources, reporting lack it from Pod API might be an easier alternative.

nrb:
@skriss I think this PR will be closed out and a new one opened; is that correct?

skriss:
Yeah, I think that would make sense. Feel free to use and change any of my content!

christian-vent:
@nrb @skriss did you create a new PR already? If yes could you please add it to this thread for reference?

nrb:
@christian-vent Not yet - I will add it when I have one.

thangamani-arun:
@nrb has this been closed without PR?

kaovilai:
I am looking to continue this work.

PR #1654, opened 2019-07-11
Add JSON as an option for log output

carthewd:
This enhancement is quite useful for my team so I took a crack at adding it. Adds an option to output logs as JSON via --log-json (e.g., velero server --log-json)

Fixes #[598](https://github.com/heptio/velero/issues/598)

Signed-off-by: Donovan Carthew <donovan.carthew@gmail.com>


pkg/util/logging/default_logger.go
```diff
@@ -33,9 +33,13 @@ func DefaultHooks() []logrus.Hook {
 
 // DefaultLogger returns a Logger with the default properties
 // and hooks.
-func DefaultLogger(level logrus.Level) *logrus.Logger {
+func DefaultLogger(level logrus.Level, j bool) *logrus.Logger {
```
nrb: I would probably rename this new variable `json`, and extend the godoc.

Just reading the function signature, it's not clear what `j` means.

carthewd: Thanks for the feedback and taking a look... makes sense, I've pushed some updates

skriss:
thanks for the PR @carthewd! I'm wondering if it would make more sense to add a flag like `--log-format`, and then allow `text` (the current) or `json` as options. It makes it easier to add a third option later - otherwise you end up with a flag per option where only one can be true. What do others think?

prydonius:
+1 looks like logrus only supports text and json outside of external formatters, but I agree that a `--log-format` option makes more sense and will make this easy to extend in the future.

pkg/util/logging/default_logger.go
```diff
@@ -32,10 +32,14 @@ func DefaultHooks() []logrus.Hook {
 }
 
 // DefaultLogger returns a Logger with the default properties
-// and hooks.
-func DefaultLogger(level logrus.Level) *logrus.Logger {
+// and hooks. The desired output format is passed as a string.
+func DefaultLogger(level logrus.Level, format string) *logrus.Logger {
```
skriss: I'd prefer to define an enum in this package for the valid formats and use that everywhere rather than a plain string.

pkg/cmd/cli/restic/server.go
```diff
@@ -68,6 +69,7 @@ func NewServerCommand(f client.Factory) *cobra.Command {
 	}
 
 	command.Flags().Var(logLevelFlag, "log-level", fmt.Sprintf("the level at which to log. Valid values are %s.", strings.Join(logLevelFlag.AllowedValues(), ", ")))
+	command.Flags().StringVar(&logFormat, "log-format", logFormat, fmt.Sprintf("the format for log output"))
```
skriss: Let's use the `Enum` flag type here (from `pkg/cmd/util/flag`) - better UX since there are a defined set of valid values. You can find an example of using it here: https://github.com/heptio/velero/blob/master/pkg/cmd/cli/backuplocation/create.go#L63

pkg/cmd/server/server.go
```diff
@@ -191,6 +196,7 @@ func NewCommand() *cobra.Command {
 	}
 
 	command.Flags().Var(logLevelFlag, "log-level", fmt.Sprintf("the level at which to log. Valid values are %s.", strings.Join(logLevelFlag.AllowedValues(), ", ")))
+	command.Flags().StringVar(&config.logFormat, "log-format", config.logFormat, fmt.Sprintf("the format for log output"))
```
skriss: Let's use the `Enum` flag type here.

skriss:
@carthewd thanks for the update! there are some more changes that I'd like to see made, but you're under no obligation to keep refining - I'm happy to make the changes myself (you'll still be tagged on the commit and changelog). Or, if you do have the appetite, I can provide another round of comments. Totally up to you - just let me know!

carthewd:
Thanks @skriss - I really appreciate the feedback! I’m quite happy to keep refining this and get it right if it’s on the right track so feel free to add more comments

changelogs/unreleased/598-carthewd
```diff
@@ -0,0 +1 @@
+enhancement: allow option to choose JSON log output
```
skriss: we usually name this file based on the PR #, so let's change it to `1654-carthewd`

pkg/util/logging/default_logger.go
```diff
@@ -19,9 +19,12 @@ package logging
 import (
 	"os"
 
+	"github.com/heptio/velero/pkg/cmd/util/flag"
 	"github.com/sirupsen/logrus"
 )
 
+const defaultLogFormat = "text"
```
skriss: let's define an enum here, like this:

```go
type Format string

const (
    FormatText Format = "text"
    FormatJSON Format = "json"
)
```

And then instead of passing strings around, we'll pass around a var of type `logging.Format`

pkg/util/logging/default_logger.go
```diff
@@ -31,11 +34,31 @@ func DefaultHooks() []logrus.Hook {
 	}
 }
 
+type LogFormatOptions struct {
```
skriss: Let's call this `FormatFlag` (so it'll be referenced from other packages as `logging.FormatFlag`)

pkg/util/logging/default_logger.go
```diff
@@ -31,11 +34,31 @@ func DefaultHooks() []logrus.Hook {
...
+	LogFormat    *flag.Enum
```
skriss: You can just embed the `*flag.Enum` (so remove `LogFormat` as the field name) -- similarly to https://github.com/heptio/velero/blob/master/pkg/util/logging/log_level_flag.go#L32

pkg/util/logging/default_logger.go
```diff
@@ -31,11 +34,31 @@ func DefaultHooks() []logrus.Hook {
...
+	defaultValue string
```
skriss: change this to `defaultValue Format` to use our new enum type

pkg/util/logging/default_logger.go
```diff
@@ -31,11 +34,31 @@ func DefaultHooks() []logrus.Hook {
...
+}
+
+func NewCreateLogFormatOptions() *LogFormatOptions {
```
skriss: rename to `NewFormatFlag`

pkg/util/logging/default_logger.go
```diff
@@ -31,11 +34,31 @@ func DefaultHooks() []logrus.Hook {
...
+	return &LogFormatOptions{
+		LogFormat: flag.NewEnum(
+			"text",
```
skriss: change this to use `string(FormatText)` and `string(FormatJSON)`

pkg/util/logging/default_logger.go
```diff
@@ -31,11 +34,31 @@ func DefaultHooks() []logrus.Hook {
...
+		),
+		defaultValue: defaultLogFormat,
+	}
+}
+
 // DefaultLogger returns a Logger with the default properties
-// and hooks.
-func DefaultLogger(level logrus.Level) *logrus.Logger {
+// and hooks. The desired output format is passed as a LogFormat Enum.
+func DefaultLogger(level logrus.Level, logFormat *LogFormatOptions) *logrus.Logger {
```
skriss: change this to take a `format Format`

pkg/util/logging/default_logger.go
```diff
@@ -19,9 +19,12 @@ package logging
 import (
```
skriss: Let's move all of the changes currently in this file that are related to creating the new flag to a new file, `format_flag.go`, in this same package.

skriss:
Once you've made the updates, make sure to run `make update` locally to properly format all of the go imports.

pkg/util/logging/default_logger.go
```diff
@@ -32,10 +32,14 @@ func DefaultHooks() []logrus.Hook {
 }
 
 // DefaultLogger returns a Logger with the default properties
-// and hooks.
-func DefaultLogger(level logrus.Level) *logrus.Logger {
+// and hooks. The desired output format is passed as a LogFormat Enum.
+func DefaultLogger(level logrus.Level, format Format) *logrus.Logger {
 	logger := logrus.New()
 
+	if (string(format)) == "json" {
```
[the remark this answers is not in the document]
carthewd: Thanks! Not sure how I missed this one :-/

PR #1661, opened 2019-07-12
CSI support proposal

nrb:
Remaining:

* [x] disabling CSI snapshotting, similar to disabling Velero-native snapshots
* [x] skipping CSI snapshotting in the plugins if the volume is covered by restic
* [x] mention feature flags/gates
* [x] use ownerRefs
* [x] deleting the underlying snapshot when the backup expires (using a `Retain` deletion policy means it will not be cleaned up by default)

### DIFF
--- a/design/csi-snapshots.md
+++ b/design/csi-snapshots.md
@@ -108,6 +108,9 @@ These plugins should be provided with Velero, as there will also be some changes
 
 ### Velero server changes
 
+Any non-plugin code changes must be behind a `EnableCSI` feature flag and the behavior will be opt-in until it's exited beta status.
+This will allow the development to continue on the feature while it's in pre-production state, while also reducing the need for long-lived feature branches.
+
 [`persistBackup`][8] will be extended to query for all `VolumeSnapshot`s associated with the backup, and persist the list to JSON.
 
 [`BackupStore.PutBackup`][9] will receive an additional argument, `volumeSnapshots io.Reader`, that contains the JSON representation of `VolumeSnapshots`.
@@ -134,10 +137,29 @@ var defaultRestorePriorities = []string{
     "replicaset",
 }
 ```
+### Restic and CSI interaction
 
+Volumes found in a `Pod`'s `backup.velero.io/backup-volumes` list will use Velero's current Restic code path.
+This also means Velero will continue to offer Restic as an option for CSI volumes.
 
+To ensure this is the case, server code must be extended so that PVCs are labeled in such a way that the CSI BackupItemAction plugin is not invoked for PVCs known to be covered by Velero.
+This can be accomplished by:
+    putting a label on the PVC, so that the AppliesTo function will only match PVCs labeled correctly.
+    using an annotation so the Execute function can inspect it and decide to exit early
+    querying for PodVolumeBackups and filtering on the pvc annotation
+    Adding a label for the associated PVC to PodVolumeBackups and query based on that selector
+
+### Garbage collection and deletion
+
+To ensure that all created resources are deleted when a backup expires or is deleted, `VolumeSnapshot`s will have an `ownerRef` defined pointing to the Velero backup that created them.
+
+In order to fully delete these objects, each `VolumeSnapshotContent`s object will need to be edited to ensure the associated provider snapshot is deleted.
+This will be done by editing the object and setting `VolumeSnapshotContent.Spec.DeletionPolicy` to `Delete`, regardless of whether or not the default policy for the class is `Retain`.
+See the Deletion Policies section below.
 
 ## Velero client changes
+
+To use CSI features, the Velero client must use the `EnableCSI` feature flag.
 
 [`DescribeBackupStatus`][13] will be extended to download the `csi-snapshots.json.gz` file for processing. GitHub Issue [1568][19] captures this work.
 
@@ -151,18 +173,6 @@ It was [introduced with dynamic provisioning support][15] in 2016, predating CSI
 
 In the `BackupItemAction` for PVCs, the associated PV will be queried and checked for the presence of `PersistentVolume.Spec.PersistentVolumeSource.CSI`.
 Volumes with any other `PersistentVolumeSource` set will use Velero's current VolumeSnapshotter plugin code path.
-
-Volumes found in a `Pod`'s `backup.velero.io/backup-volumes` list will use Velero's current Restic code path.
-This also means Velero will continue to offer Restic as an option for CSI volumes.
-
-
-### Garbage collection and deletion
-
-To ensure that all created resources are deleted when a backup expires or is deleted, `VolumeSnapshot`s will have an `ownerRef` defined pointing to the Velero backup that created them.
-
-In order to fully delete these objects, each `VolumeSnapshotContent`s object will need to be edited to ensure the associated provider snapshot is deleted.
-This will be done by editing the object and setting `VolumeSnapshotContent.Spec.DeletionPolicy` to `Delete`, regardless of whether or not the default policy for the class is `Retain`.
-See the Deletion Policies section below.
 
 
 ## Alternatives Considered
--- a/design/csi-snapshots.md
+++ b/design/csi-snapshots.md
@@ -50,11 +50,13 @@ The plugins will be as follows:
 
 This plugin will act directly on PVCs, since an implementation of Velero's VolumeSnapshotter does not have enough information about the StorageClass to properly create the `VolumeSnapshot` objects.
 
-Only `PersistentVolumeClaim` objects without the `velero.io/restic-backed` label will be processed, using the plugin's `AppliesTo` function.
-
 The associated PV will be queried and checked for the presence of `PersistentVolume.Spec.PersistentVolumeSource.CSI`. (See the "Snapshot Mechanism Selection" section below).
 If this field is `nil`, then the plugin will return early without taking action.
 If the `Backup.Spec.SnapshotVolumes` value is `false`, the plugin will return early without taking action.
+
+Additionally, to prevent creating CSI snapshots for volumes backed up by restic, the plugin will query for all pods in the `PersistentVolumeClaim`'s namespace.
+It will then filter out the pods that have the PVC mounted, and inspect the `backup.velero.io/backup-volumes` annotation for the associated volume's name.
+If the name is found in the list, then the plugin will return early without taking further action.
 
 Create a `VolumeSnapshot.snapshot.storage.k8s.io` object from the PVC.
 Label the `VolumeSnapshot` object with the [`velero.io/backup-name`][10] label for ease of lookup later.
@@ -144,10 +146,12 @@ var defaultRestorePriorities = []string{
 Volumes found in a `Pod`'s `backup.velero.io/backup-volumes` list will use Velero's current Restic code path.
 This also means Velero will continue to offer Restic as an option for CSI volumes.
 
-To ensure this is the case, server code must be extended so that PVCs are marked in such a way that the CSI BackupItemAction plugin is not invoked for PVCs known to be backed up by restic.
+The `velero.io/csi-pvc` BackupItemAction plugin will inspect pods in the namespace to ensure that it does not act on PVCs already being backed up by restic.
 
-The restic backupper's `BackupPodVolumes` method will be extended to get PersistentVolumeClaims where relevant and add a `velero.io/restic-backed: true` label.
-This label will be used by the `velero.io/csi-pvc` BackupItemAction plugin to ensure only PersistentVolumeClaims _without_ that label are processed.
+This is preferred to modifying the PVC due to the fact that Velero's current backup process backs up PVCs and PVs mounted to pods at the same time as the pod.
+
+A drawback to this approach is that we're querying all pods in the namespace per PVC, which could be a large number.
+In the future, the plugin interface could be improved to have some sort of context argument, so that additional data such as our existing `resticSnapshotTracker` could be passed to plugins and reduce work.
 
 ### Garbage collection and deletion
 
--- a/design/csi-snapshots.md
+++ b/design/csi-snapshots.md
@@ -142,7 +142,7 @@ var defaultRestorePriorities = []string{
 Volumes found in a `Pod`'s `backup.velero.io/backup-volumes` list will use Velero's current Restic code path.
 This also means Velero will continue to offer Restic as an option for CSI volumes.
 
-To ensure this is the case, server code must be extended so that PVCs are labeled in such a way that the CSI BackupItemAction plugin is not invoked for PVCs known to be covered by Velero.
+To ensure this is the case, server code must be extended so that PVCs are marked in such a way that the CSI BackupItemAction plugin is not invoked for PVCs known to be backed up by restic.
 This can be accomplished by:
     putting a label on the PVC, so that the AppliesTo function will only match PVCs labeled correctly.
     using an annotation so the Execute function can inspect it and decide to exit early
@@ -156,6 +156,9 @@ To ensure that all created resources are deleted when a backup expires or is del
 In order to fully delete these objects, each `VolumeSnapshotContent`s object will need to be edited to ensure the associated provider snapshot is deleted.
 This will be done by editing the object and setting `VolumeSnapshotContent.Spec.DeletionPolicy` to `Delete`, regardless of whether or not the default policy for the class is `Retain`.
 See the Deletion Policies section below.
+The edit will happen before making Kubernetes API deletion calls to ensure that the cascade works as expected.
+
+Deleting a Velero `Backup` or any associated CSI object via `kubectl` is unsupported; data will be lost or orphaned if this is done.
 
 ## Velero client changes
 
--- a/design/csi-snapshots.md
+++ b/design/csi-snapshots.md
@@ -3,7 +3,7 @@
 Status: Draft
 
 The Container Storage Interface (CSI) [introduced an alpha snapshot API in Kubernetes v1.12][1].
-It will reach beta support in Kubernetes v1.17.
+It will reach beta support in Kubernetes v1.17, scheduled for release in December 2019.
 This proposal documents an approach for integrating support for this snapshot API within Velero, augmenting its existing capabilities.
 
 ## Goals
@@ -110,7 +110,6 @@ Only `PersistentVolumeClaim` objects with the `velero.io/volume-snapshot-name` l
 Metadata (excluding labels) will be cleared, and the `velero.io/volume-snapshot-name` label will be used to find the relevant `VolumeSnapshot`.
 A reference to the `VolumeSnapshot` will be added to the `PersistentVolumeClaim.DataSource` field.
 
-
 No special logic is required to restore `VolumeSnapshotClass` objects.
 
 These plugins should be provided with Velero, as there will also be some changes to core Velero code to enable association of a `Backup` to the included `VolumeSnapshot`s.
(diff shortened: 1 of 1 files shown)


design/csi-snapshots.md
```diff
@@ -0,0 +1,193 @@
...
+One to two sentences that describes the goal of this proposal.
+The reader should be able to tell by the title, and the opening paragraph, if this document is relevant to them.
+
+The Container Storage Interface (CSI) has [introduced an alpha snapshot API in Kubernetes v1.12][1].
+Current plans indicate it will reach beta support in Kubernetes v1.16.
+This document suggests an approach for integrating support for this snapshot API within Velero, augmenting its existing capabilities.
+
+## Goals
+
+- Enable Velero to create CSI snapshot objects
```
skriss: I'd reword to something like "Enable Velero to back up and restore CSI persistent volumes using the CSI snapshot API" for clarity

design/csi-snapshots.md
```diff
@@ -0,0 +1,192 @@
...
+The CSI controllers will create a `VolumeSnapshotContent.snapshot.storage.k8s.io` object associated with the `VolumeSnapshot`.
+Associated `VolumeSnapshotContent` objects will be retrieved and updated with the [`velero.io/backup-name`][10] label for ease of lookup later.
+`velero.io/volume-snapshot-name` will be applied as a label to the PVC so that the `VolumeSnapshot` can be found easily for restore.
+`VolumeSnapshot`, `VolumeSnapshotContent`, and `VolumeSnapshotClass` objects would be returned as additional items to be backed up. GitHub issue [1566][18] represents this work.
+The `VolumeSnapshotContent.Spec.VolumeSnapshotSource.SnapshotHandle` field is the link to the underlying platform's snapshot, and must be preserved for restoration.
+
+The plugin will _not_ wait for the `VolumeSnapshot.Status.IsReady` field to be `true` before returning.
+This maintains current Velero behavior, though may not be desirable for all storage providers.
+
+* A `RestoreItemAction` for `VolumeSnapshotContent` objects, named `velero.io/csi-vsc`.
```
skriss: I'm wondering if some of this code belongs in the backup sync controller - so prior to even executing a restore, we'd create the CSI CR's in the cluster to represent the snapshots. This would be more consistent with how we're handing PodVolumeBackups, for example. Still thinking through it - interested in your thoughts.

design/csi-snapshots.md
```diff
@@ -0,0 +1,192 @@
...
+In the `BackupItemAction` for PVCs, the associated PV will be queried and checked for the presence of `PersistentVolume.Spec.PersistentVolumeSource.CSI`.
+Volumes with any other `PersistentVolumeSource` set will use Velero's current VolumeSnapshotter plugin code path.
+
+Volumes found in a `Pod`'s `backup.velero.io/backup-volumes` list will use Velero's current Restic code path.
+This also means Velero will continue to offer Restic as an option for CSI volumes.
+
+
+## Alternatives Considered
+
+* Implementing similar logic in a Velero VolumeSnapshotter plugin was considered.
```
skriss: How are we going to handle GC'ing CSI snapshots for expired backups?

design/csi-snapshots.md
```diff
@@ -0,0 +1,192 @@
...
+In fact, much of the logic for creating the CSI objects will be driven entirely inside of the plugin implementation.
+
+However, Velero currently has no way for plugins to communicate that some arbitrary data should be stored in or retrieved from object storage, such as list of all `VolumeSnapshot` objects associated with a given `Backup`.
+This is important, because to display snapshots included in a backup, whether as native snapshots or Restic backups, separate JSON-encoded lists are stored within the backup on object storage.
+Snapshots are not listed directly on the `Backup` to fit within the etcd size limitations.
+Additionally, there are no client-side Velero plugin mechanisms, which means that the `velero describe backup --details` command would have no way of displaying the objects to the user, even if they were stored.
+
+## Notes on Usage
+
+In order for underlying, provider-level snapshots to be retained similarly to Velero's current functionality, the `VolumeSnapshotContent.DeletionPolicy` field must be set to `Retain`.
```
skriss: Is this something the user would be required to set up themselves, ahead of time? IOW the VolumeSnapshotClass isn't something managed by Velero, correct?

nrb: No, it would not be handled by Velero. However, we should definitely document this behavior so users aren't surprised.

design/csi-snapshots.md
nrb: Hmm, maybe. This may have implications for how they're currently backed up, but I'm not sure. I'll have to think on that some more. 

As it stands, the JSON file with the VolumeSnapshots listed is just for getting backup details, but it may make sense to sync them in and just query the cluster. The controller would need to sync in the VolumeSnapshotClass, VolumeSnapshotContent, and VolumeSnapshot at sync time.

Also note that VolumeSnapshot objects live inside the same namespace as the PVC they captured from, so we'd also have to recreate the NS(es).

design/csi-snapshots.md
nrb: Great question, this is something I missed. I'll do some review.

Initial thought is that the deletion controller would do a label-selection query for any VolumeSnapshots labeled with the backup name and delete them. However, if the `Retain` deletion policy is specified then the volume provider snapshot will stick around, which isn't desirable once we hit TTL.

design/csi-snapshots.md
skriss: Good point, yeah I don't think we want to create the NSes as part of it. I suppose we could create the SnapshotClass + VSCs only, but even that might be weird.

Will ponder some more.  The inconsistencies between current snapshots, restic, and CSI snapshots bother me, but they're also hard to reconcile given the different models that all three of these have.

design/csi-snapshots.md
```diff
@@ -0,0 +1,192 @@
...
+A new `describeCSIVolumeSnapshots` function should be added to the [output][12] package that knows how to render the included `VolumeSnapshot` names referenced in the `csi-snapshots.json.gz` file.
+
+### Snapshot mechanism selection
+
+The most accurate, reliable way to detect if a PersistentVolume is a CSI volume is to check for a non-`nil` [`PersistentVolume.Spec.PersistentVolumeSource.CSI`][16] field.
+Using the [`volume.beta.kubernetes.io/storage-provisioner`][14] is not viable, since the usage is for any PVC that should be dynamically provisioned, and is _not_ limited to CSI implementations.
+It was [introduced with dynamic provisioning support][15] in 2016, predating CSI.
+
+In the `BackupItemAction` for PVCs, the associated PV will be queried and checked for the presence of `PersistentVolume.Spec.PersistentVolumeSource.CSI`.
+Volumes with any other `PersistentVolumeSource` set will use Velero's current VolumeSnapshotter plugin code path.
```
prydonius: Is there a case where you'd want to use Velero's current VolumeSnapshotter instead of CSI even if CSI was used to provision the volume? For example, if the CSI driver supported provisioning but not snapshotting?

nrb: Hm, that's a good possibility. I know there's gRPC methods for finding out if the driver supports snapshots, but I'd have to check to see if we can query the driver from k8s.

design/csi-snapshots.md
```diff
@@ -0,0 +1,192 @@
+# CSI Snapshot Support
+
+Status: Draft
+
+The Container Storage Interface (CSI) has [introduced an alpha snapshot API in Kubernetes v1.12][1].
+Current plans indicate it will reach beta support in Kubernetes v1.16.
+This document suggests an approach for integrating support for this snapshot API within Velero, augmenting its existing capabilities.
```
timh: s/This document suggests/This proposal documents/

design/csi-snapshots.md
```diff
@@ -0,0 +1,192 @@
...
+- Enable Velero to backup and restore CSI-backed volumes using the Kubernetes CSI CustomResourceDefinition API
+
+## Non Goals
+
+- Replacing Velero's existing [VolumeSnapshotter][7] API
+- Replacing Velero's Restic support
+
+## Background
+
+Velero has had support for performing persistent volume snapshots since it's inception.
```
timh: s/it's/its/

design/csi-snapshots.md
```diff
@@ -0,0 +1,192 @@
...
+However, support has been limited to a handful of providers.
+The plugin API introduced in Velero v0.7 enabled the community to expand the number of supported providers.
+In the meantime, the Kubernetes sig-storage advanced the CSI spec to allow for a generic storage interface, opening up the possibility of moving storage code out of the core Kubernetes code base.
+The CSI working group has also developed a generic snapshotting API that any CSI driver developer may implement, giving users the ability to snapshot volumes from a standard interface.
+
+By supporting the CSI snapshot API, Velero can extend it's support to any CSI driver, without requiring a Velero-specific plugin be written, easing the development burden on providers while also reaching more end users.
```
timh: s/it's/its/

design/csi-snapshots.md
```diff
@@ -0,0 +1,192 @@
...
+
+## High-Level Design
+
+In order to support CSI's snapshot API, Velero must interact with the [`VolumeSnapshot`][2] and [`VolumeSnapshotContent`][3] CRDs.
+These act as requests to the CSI driver to perform a snapshot on the underlying provider's volume.
+This can largely be accomplished with Velero `BackupItemAction` and `RestoreItemAction` plugins that operate on these CRDs.
+
+Additionally, changes to the Velero server and client code are necessary to track `VolumeSnapshots` that are associated with a given backup, similarly to how Velero tracks its own [`volume.Snapshot`][4] type.
```
timh: s/`VolumeSnapshots`/`VolumeSnapshot`s/

nrb:
Per feedback, we need a way of telling Velero not to create CSI snapshots. Maybe use the `SnapshotVolumes` field on the backup?

design/csi-snapshots.md
skriss: Our in-tree VolumeSnapshotters won't work as-is for CSI volumes, even if they're on the right platform, due to how the GetVolumeID/SetVolumeID functions work (and maybe other things too, not sure). We could look at changing that if we did want to be able to support using them to snapshot CSI volumes, but right now it won't work.

design/csi-snapshots.md
skriss: We could use owner refs here - make the relevant CSI objects owned by the Velero backup. Then we get cascading deletion.

design/csi-snapshots.md
skriss: Yeah, thinking about this more, I don't think we want to try to do anything with the sync controller.

design/csi-snapshots.md
```diff
@@ -0,0 +1,192 @@
...
+
+A set of [prototype][6] plugins was developed that informed this design.
+
+The plugins will be as follows:
+
+* A `BackupItemAction` for `PersistentVolumeClaim`s, named `velero.io/csi-pvc` 
+The associated PV will be queried and checked for the presence of `PersistentVolume.Spec.PersistentVolumeSource.CSI`. (See the "Snapshot Mechanism Selection" section below)
+If this field is `nil`, then the plugin will return early without taking action.
+Create a `VolumeSnapshot.snapshot.storage.k8s.io` object from the PVC.
+Label the `VolumeSnapshot` object with the [`velero.io/backup-name`][10] label for ease of lookup later.
```
skriss: I *think* it makes sense to use an owner ref here as well to track the dependency and get cascading deletion for GC purposes.

design/csi-snapshots.md
```diff
@@ -0,0 +1,192 @@
...
+
+* A `RestoreItemAction` for `VolumeSnapshotContent` objects, named `velero.io/csi-vsc`.
+Only `VolumeSnapshotContent` objects with the `velero.io/backup-name` label will be processed; if the label is missing, the plugin will return the unmodified object.
+The metadata (excluding labels), `PersistentVolumeClaim.UUID`, and `VolumeSnapshotRef.UUID` fields will be cleared.
+The reference fields are cleared because the associated objects will get new UUIDs in the cluster.
+This also maps to the "import" case of [the snapshot API][1].
+
+* A `RestoreItemAction` for `VolumeSnapshot` objects, named `velero.io/csi-vs`.
+Only `VolumeSnapshot` objects with the `velero.io/backup-name` label will be processed; if the label is missing, the plugin will return the unmodified object.
+Metadata (excluding labels) and `Source` fields on the object will be cleared.
```
skriss: If we go with the ownerRef approach for linking Velero backups to VolumeSnapshots, we'll need to update the ownerRef to have the right namespace & UID here.

skriss:
I know you brought up feature flags recently - I do think we should consider adding them, and adding one for the CSI work, as we're developing here. I imagine we're going to ship this in parts (e.g. first add support for *creating* CSI snapshots, then add support for restoring them, ...) and it would be nice to allow users to opt-in until they're ready.

skriss:
One other thing to think about - similar to existing code, we'll want to skip creating snapshots if restic is being used to back up a PVC - I'm not exactly sure how we'll do that given that the CSI code is in plugins. Need to think about it.

nrb:
I created #1759 for the feature flag work. It likely needs a bit more detail, but at least we have a place to discuss it now.

design/csi-snapshots.md
nrb: Update: currently there is no way to query what a driver supports from k8s. So we can't really know if CSI snapshotting is supported via a query to something like the storageclass at the moment.

I would say to simplify things, if CSI volumes are in use but the driver doesn't have snapshotting, restic would be the best option.

nrb:
@skriss @prydonius @carlisia I believe all the items in the checklist are now addressed. I've also done a bit of reorganization of the sections. Please take another look!

design/csi-snapshots.md
```diff
@@ -0,0 +1,241 @@
...
+}
+```
+### Restic and CSI interaction
+
+Volumes found in a `Pod`'s `backup.velero.io/backup-volumes` list will use Velero's current Restic code path.
+This also means Velero will continue to offer Restic as an option for CSI volumes.
+
+To ensure this is the case, server code must be extended so that PVCs are marked in such a way that the CSI BackupItemAction plugin is not invoked for PVCs known to be backed up by restic.
+
+The restic backupper's `BackupPodVolumes` method will be extended to get PersistentVolumeClaims where relevant and add a `velero.io/restic-backed: true` label.
```
skriss: I see a couple of challenges with this approach - (1) BackupPodVolumes will actually run *after* the PVC has been backed up, since the PVC is backed up as an additional item for the pod prior to the restic backups running; and (2) the label value would need to have the name or UID of the backup, so we know if the PVC is being backed up with restic for *this* backup, not just any previous backup.

One alternative approach for consideration - you could have the `csi-pvc` backup item action list all pods in the namespace that the PVC is in, find the pod that is mounting the PVC, and check if it's annotated for restic backup. If we do something like that, it means we don't need to add a new label/annotation which I think is desirable.

skriss: Longer-term, I wonder if we should pass a richer "context" object to our plugins that includes information like this. Right now one of the drawbacks of implementing logic in plugins is that you don't get access to some of those internal data structures used by `pkg/backup` or `pkg/restore` to track what's happening. If we could expose those to plugins, things would be  a lot easier.

nrb: Yeah, I really wanted to use the resticSnapshotTracker, but opted for the label since that could be used with the `AppliesTo` function, thereby not even invoking the `Execute` method for `csi-pvc` when a PVC should be skipped. Was hoping to avoid too many extra API server calls to determine a no-op, but it may be unavoidable for now. 

I'm not sure I want to completely redo the plugin interface in this proposal, as I'd like for CSI support to be a 1.x feature. However, I think providing some sort of context object would be very useful for plugins in the future.

Thanks for the call out on the PVCs being backed up during the Pod's process; I had forgotten that the PVCs were backed up immediately instead of being added to the PVC group's loop.

Your alternative sounds do-able, I think.

design/csi-snapshots.md
```diff
@@ -0,0 +1,245 @@
...
+Using the [`volume.beta.kubernetes.io/storage-provisioner`][14] is not viable, since the usage is for any PVC that should be dynamically provisioned, and is _not_ limited to CSI implementations.
+It was [introduced with dynamic provisioning support][15] in 2016, predating CSI.
+
+In the `BackupItemAction` for PVCs, the associated PV will be queried and checked for the presence of `PersistentVolume.Spec.PersistentVolumeSource.CSI`.
+Volumes with any other `PersistentVolumeSource` set will use Velero's current VolumeSnapshotter plugin code path.
+
+
+## Alternatives Considered
+
+* Implementing similar logic in a Velero VolumeSnapshotter plugin was considered.
```
nrb: @betta1 Antony, is this summary of your experiments with making a VolumeSnapshotter plugin accurate? Is there anything I missed here that you encountered?

betta1: @nrb yes this is accurate and is in line with our initial experiments with a CSI VolumeSnapshotter plugin -- it's been a while but i remember our biggest issue was that the VolumeSnapshotter interface didn't provide some of the metadata like PVC/NameSpace/StorageClass/ that are required  for CSI snapshots. We preferred the approach you had taken with your CSI prototype  utilizing Backup and restore Item actions since with item actions we can get access to these metadata. Btw I tested your prototype and  had filed an issue (https://github.com/vmware-tanzu/velero-plugin-for-csi/issues/1) -- backup worked ok, ran into some issue with restore.

nrb:
Possible issue to think through:

* Take a backup of a namespace w/ snapshot on Monday - the VolumeSnapshot CRD, which is namespaced should be included, along with its associated VolumeSnapshotContent CRD
* Take a backup of the same namespace w/ snapshot on Tuesday - the VolumeSnapshot and VolumeSnapshotContent CRDs for Tuesday _and_ Monday should be included.
* Delete the Monday backup - Velero should remove the VolumeSnapshot, VolumeSnapshotContent, and snapshot.
* Delete the namespace.
* Restore the Tuesday backup - this restores the Monday VolumeSnapshot & VolumeSnapshotContent objects, which are no longer valid as they point to a snapshot that was deleted.

Should the RestoreItemAction for VolumeSnapshot/VolumeSnapshotContent perhaps validate that the snapshot is there before adding the CRDs to the cluster? If so, how? I'm not sure if there's a generic way to query CSI for a snapshot's existence right now.

nrb:
> I'm not sure if there's a generic way to query CSI for a snapshot's existence right now.

There's likely a way to do this via the gRPC API, otherwise the CSI drivers couldn't work.

skriss:
This is an "import" case where the underlying snapshot does not exist, right? Will the snapshot controller observe this behavior and set some phase/condition on the VS/VSC object?

nrb:
Yeah, this would be a case where the snapshot doesn't exist. I'll review the snapshot controller code and see.

design/csi-snapshots.md
```diff
@@ -0,0 +1,245 @@
+# CSI Snapshot Support
+
+Status: Draft
+
+The Container Storage Interface (CSI) [introduced an alpha snapshot API in Kubernetes v1.12][1].
+Current plans indicate it will reach beta support in Kubernetes v1.16.
```
xing-yang: v1.17

nrb:
[Here is the recording of the meeting for our discussion of the Velero/CSI integration](https://vmware.zoom.us/recording/play/eoK-dh96BRkf1aIrbLsekeilCss6wEHiogShg8CHiR_HysaH2BgppLXYkL61g__g?continueMode=true).

The summary slides are at https://docs.google.com/presentation/d/1QG0gu5Y9xiF6rjkqFgsOMOWMjSkj6Il18dNWKWRjipo/edit?usp=sharing.

The plan currently is to get the proof of concept plugins up to date with the v1beta types, get the design document merged, then ensure that the Velero core code base is updated to reflect necessary changes behind the `EnableCSI` feature flag - these changes will be communicated in release notes and blog posts when it's released. We'll also be giving regular updates during our community meetings.

design/csi-snapshots.md
nrb: Updated!

skriss:
@carlisia @prydonius please try to review this again over the next few days, with an eye towards any questions/issues that need to be resolved before moving on to implementation. I'll do the same as well. I think we're just about there - but raise anything you see!

cc @ashish-amarnath

nrb:
@skriss I think the only thing I have left in my notes to put in here is to mention the VSL question you had. I'll get that in tomorrow morning.

nrb:
@skriss VSL and VolumeSnapshotClass comparison added, let me know if I should expand more.

skriss:
LGTM. @carlisia @prydonius @ashish-amarnath let's set a deadline of this Friday for getting this reviewed - if there are no further comments by then, we'll merge and get on with it :)

skriss:
merging!

PR #1662, opened 2019-07-12
add make gen-docs to generate a new versioned docs directory

skriss:

cc @carlisia @nrb @prydonius 

I added a helper script/make target to generate a new versioned docs directory, that both rewrites specific versioned links and does a little bit of git work to make it easy to view a diff of what's changed since the previous version's docs for review purposes.

You can try it now with `NEW_DOCS_VERSION=v1.1.0 make gen-docs` to see its behavior.


nrb:
Not yet reviewed this, but I propose that pending reviews, this officially replaces the doc-site-gen tool for Velero, since that's private and no longer actively maintained.

hack/gen-docs.sh
```diff
@@ -0,0 +1,73 @@
...
+# 'git add' the previous version's docs as-is so we get a useful diff when we copy the master docs in
+echo "Running 'git add' for previous version's doc contents to use as a base for diff"
+git add site/docs/${NEW_DOCS_VERSION}
+
+
+# now copy the contents of site/docs/master into the same directory so we can get a nice
+# git diff of what changed since previous version
+rm -rf site/docs/${NEW_DOCS_VERSION}/ && cp -r site/docs/master/ site/docs/${NEW_DOCS_VERSION}/
+
+# replace known version-specific links -- the sed syntax is slightly different in OS X and Linux,
```
nrb: Since this is presumably to generate the latest docs, would it be worth a `sed` to update `site/_config.yaml`'s `latest` value?

I will note that that also does not add the newest version in to the `versions` list, either, which is likely desirable but may be better suited for a follow-up PR.

skriss:
Ah - yeah, I bet the CSS is broken due to some missing metadata entries, e.g. in `site/_config.yaml`.

Let me look at updating those as well.

skriss:
Yeah, looks like to get everything updated we need to:
- insert into `defaults` and `versions` in `_config.yml`
- update `latest` in _config.yml`
- create a new TOC file (copied from master) in `_data/`
- update `_data/toc-mapping.yml`

At a quick glance everything looked good after making these updates as well. I'll see how annoying this is gonna be to scriptify.

skriss:
So, I looked at scripting the updates here and it's kind of a pain. WDYT about just documenting the additional manual steps for now?

nrb:
I'd be ok with documenting them and opening an issue to automate them furhter.

In my mind, the ideal would be to have the docs copied, fixed, and committed by some CI worker. But that doesn't have to happen today or in 1 PR, either.

Makefile
```diff
@@ -225,3 +225,21 @@ serve-docs:
...
+#	   'master' docs directory, updating any version-specific links (e.g. to a
+#	   specific branch of the GitHub repository) to use the new version.
+#
+# The unstaged changes in the working directory can now be diff'ed against the
+# staged changes using 'git diff' to review all docs changes made since the previous 
+# tagged version. Once the unstaged changes are ready, they can be added to the
+# staging area using 'git add' and then committed.
+#
+# To run gen-docs: "NEW_DOCS_VERSION=v1.1.0 make gen-docs"
+gen-docs:
```
nrb: Would it be too onerous to make this a dependency for `make release`, so it's all contained in one process?

skriss:
OK, updated with documentation on the other steps.

Makefile
skriss: I think long-term this is a good goal to shoot for, but I don't think it fits right now given the manual steps involved.

hack/gen-docs.sh
```diff
@@ -0,0 +1,76 @@
...
+    echo "[OS X] updating version-specific links"
+    find site/docs/${NEW_DOCS_VERSION} -type f -name "*.md" | xargs sed -i '' "s|https://velero.io/docs/master|https://velero.io/docs/$NEW_DOCS_VERSION|g"
+    find site/docs/${NEW_DOCS_VERSION} -type f -name "*.md" | xargs sed -i '' "s|https://github.com/heptio/velero/blob/master|https://github.com/heptio/velero/blob/$NEW_DOCS_VERSION|g"
+else
+    echo "[Linux] updating version-specific links"
+    find site/docs/${NEW_DOCS_VERSION} -type f -name "*.md" | xargs sed -i'' "s|https://velero.io/docs/master|https://velero.io/docs/$NEW_DOCS_VERSION|g"
+    find site/docs/${NEW_DOCS_VERSION} -type f -name "*.md" | xargs sed -i'' "s|https://github.com/heptio/velero/blob/master|https://github.com/heptio/velero/blob/$NEW_DOCS_VERSION|g"
+fi
+
+echo "Success! site/docs/$NEW_DOCS_VERSION has been created. You can now run 'git diff' to review all docs changes made since the previous tagged version."
```
nrb: May be worth referring to the manual steps in README-JEKYLL.md in this message so they're less likely to be missed.

nrb:
LGTM with one minor comment.

site/README-JEKYLL.md
```diff
@@ -31,3 +31,25 @@ This mirrors the plug-ins used by GitHub Pages on your local machine including J
...
+1. In `$VELEROROOT`, run:
+
+   ```bash
+   # set to the appropriate version number
+   NEW_DOCS_VERSION=vX.Y.Z make gen-docs
+   ```
+
+1. In `site/_config.yml`, under the `defaults` field, add an entry for the new version just under `master` by copying the most recent version's entry and updating the version numbers.
+
+1. Also in `site/_config.yml`, update the `latest` field to have the new version number as its value.
```
nrb: I think `sed "s/latest: ${PREVIOUS_DOCS_VERSION}/latest: ${NEW_DOCS_VERSION}/" site/_config.yml` (this is the Linux form) will do this.

site/README-JEKYLL.md
```diff
@@ -31,3 +31,25 @@ This mirrors the plug-ins used by GitHub Pages on your local machine including J
...
+
+1. Also in `site/_config.yml`, add the new version to the `versions` field, just under `- master`.
```
nrb: I think `sed "/- master/a - ${NEW_DOCS_VERSION}" site/_config.yml` will do this

The "a" directive in the middle is a sed append operation.

site/README-JEKYLL.md
```diff
@@ -31,3 +31,25 @@ This mirrors the plug-ins used by GitHub Pages on your local machine including J
...
+
+1. In `site/_data/`, create a new table of contents file by copying `master-toc.yml` to `vX-Y-Z.yml` and editing as needed.
+
+1. In `site/_data/toc-mapping.yml`, add an entry for the new version just under `master` by copying the most recent version's entry and updating the key and value.
```
nrb:
```
# Translate from vX.Y.Z to vX-Y-Z-toc
NEW_DOCS_TOC="$(echo ${NEW_DOCS_VERSION} | tr . -)-toc"
cp site/_data/master-toc.yml "${NEW_DOCS_TOC}.yml"
# The "a" directive in the middle is a sed append operation.
sed "/master: master-toc/a ${NEW_DOCS_VERSION}: ${NEW_DOCS_TOC}" site/_data/toc-mapping.yml
```

I believe these will do the last 2 steps.

site/README-JEKYLL.md
```diff
@@ -31,3 +31,25 @@ This mirrors the plug-ins used by GitHub Pages on your local machine including J
...
+   NEW_DOCS_VERSION=vX.Y.Z make gen-docs
+   ```
+
+1. In `site/_config.yml`, under the `defaults` field, add an entry for the new version just under `master` by copying the most recent version's entry and updating the version numbers.
+
+1. Also in `site/_config.yml`, update the `latest` field to have the new version number as its value.
+
+1. Also in `site/_config.yml`, add the new version to the `versions` field, just under `- master`.
+
+1. In `site/_data/`, create a new table of contents file by copying `master-toc.yml` to `vX-Y-Z.yml` and editing as needed.
```
prydonius: looks like so far we have a `-toc.yml` prefix even on versioned tocs, so should this be: 

```suggestion
1. In `site/_data/`, create a new table of contents file by copying `master-toc.yml` to `vX-Y-Z-toc.yml` and editing as needed.
```

?

nrb:
Gathered my suggestions for the manual steps into one commit. https://github.com/nrb/velero/commit/f86e039be284ad82c78e6fe3fab755e736ac6c28

skriss:
Alright, I addressed @nrb and @prydonius's comments. I tested this on both mac and ubuntu and it works as I'd expect. The only manual step remaining is updating `site/_config.yml`'s `defaults` list to add the new version.

I also tried to be as clear as I could about why I'm doing the `git` stuff - the whole idea is that if we're generating say `v2.0.0` docs, first copy the `v1.0.0` docs into the dir named `v2.0.0`, and `git add` it. Then, replace the contents of the `v2.0.0` directory with a copy of the `master` docs. Now, running `git diff` gives us a clean diff of what's changed since the previous tagged version. (a) does this make sense? (I've done this manually in the past to aid reviewing changes) (b) are my comments clear?

Makefile
```diff
@@ -225,3 +225,29 @@ serve-docs:
...
+
+# gen-docs generates a new versioned docs directory under site/docs. It follows
+# the following process:
+#	1. Copies the contents of the most recently tagged docs directory into the new
+#	   directory, to establish a useful baseline to diff against.
+#	2. Adds all copied content from step 1 to git's staging area via 'git add'.
+#	3. Replaces the contents of the new docs directory with the contents of the
+#	   'master' docs directory, updating any version-specific links (e.g. to a
+#	   specific branch of the GitHub repository) to use the new version
+#   4. Copies the previous version's ToC file and runs 'git add' to establish
```
prydonius: These last 3 points are indented to the same level as the previous 3, is that intentional?

Makefile
```diff
@@ -225,3 +225,29 @@ serve-docs:
 	-it -p 4000:4000 \
 	jekyll/jekyll \
 	jekyll serve --livereload
+
+# gen-docs generates a new versioned docs directory under site/docs. It follows
```
prydonius: I'm wondering if it's really worth duplicating this documentation both here and in the script, it'll be two places we'll need to update it if we change anything. That said, I'm also not sure where the best place to document this would be (probably here given it's unlikely that gen-docs.sh will be used directly?

hack/gen-docs.sh
```diff
@@ -0,0 +1,127 @@
...
+echo "Running 'git add' for previous version's doc contents to use as a base for diff"
+git add site/docs/${NEW_DOCS_VERSION}
+
+# now copy the contents of site/docs/master into the same directory so we can get a nice
+# git diff of what changed since previous version
+echo "Copying site/docs/master/ to site/docs/${NEW_DOCS_VERSION}/"
+rm -rf site/docs/${NEW_DOCS_VERSION}/ && cp -r site/docs/master/ site/docs/${NEW_DOCS_VERSION}/
+
+# make a copy of the previous versioned ToC
+NEW_DOCS_TOC="$(echo ${NEW_DOCS_VERSION} | tr . -)-toc"
```
nrb: Wouldn't we want to copy `master`'s TOC here, rather than the previous version?

I'm imagining that we'd update `master` with any new ToC layouts, and carry that forward. Are you thinking that `master` and the current release would be updated in sync?

Makefile
skriss: not intentional, will fix

nrb:
>(a) does this make sense? (I've done this manually in the past to aid reviewing changes)

Yeah, it makes sense to me.

>(b) are my comments clear?

I believe they are, yep.

Makefile
skriss: Yeah - I'm ok with just having it in the Makefile, I'll just mention in the script that it's used for a make target

hack/gen-docs.sh
skriss: This is to get a nice `git diff` also - so first copy over the previous version's ToC, `git add` it, then overwrite it with the `master` ToC so  you can now `git diff` to see what's changing since last version

nrb: Oh, ok, so the same logic as the directory. Got it!

skriss:
OK, fixed comments

PR #1673, opened 2019-07-17
Remove references to apps/v1beta1 API group

### LINKED ISSUE #1672: Update velero install package to use apps/v1 API Group
In Kubernetes 1.16, the [extensions/v1beta1 API group will no longer be served by default](https://groups.google.com/forum/#!topic/kubernetes-dev/je0rjyfTVyc/discussion).

`velero install` currently uses this group for its Deployment, so it will need to be updated. The Helm chart already uses this API group.

nrb:
In Kubernetes v1.16, the apps/v1 API group will be the default served
for relevant resources.

Update any references to apps/v1beta1 for fowards compatibility.

Fixes #1672

### DIFF
--- a/changelogs/unreleased/1673-nrb
+++ b/changelogs/unreleased/1673-nrb
@@ -1,1 +1,1 @@
-Update Velero Deployment to use apps/v1 API group
+Update Velero Deployment to use apps/v1 API group. `velero install` and `velero plugin add/remove` commands will now require Kubernetes 1.9+
--- a/changelogs/unreleased/1673-nrb
+++ b/changelogs/unreleased/1673-nrb
@@ -1,0 +1,1 @@
+Update Velero Deployment to use apps/v1 API group
--- a/examples/minio/00-minio-deployment.yaml
+++ b/examples/minio/00-minio-deployment.yaml
@@ -19,7 +19,7 @@ metadata:
   name: velero
 
 ---
-apiVersion: apps/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   namespace: velero
--- a/examples/nginx-app/base.yaml
+++ b/examples/nginx-app/base.yaml
@@ -21,7 +21,7 @@ metadata:
     app: nginx
 
 ---
-apiVersion: apps/v1beta1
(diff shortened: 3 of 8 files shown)


skriss:
This all looks good.  Looks like `pkg/cmd/cli/plugin` also uses the beta group so needs to update as well.

We're also going to want to be super-clear in the release note that this is effectively bumping our minimum supported Kube version to v1.9, at least for `velero install` and `velero plugin add/remove`.

nrb:
> We're also going to want to be super-clear in the release note that this is effectively bumping our minimum supported Kube version to v1.9

Thanks for verifying that version for me! I'll make sure that's mentioned in the release note, and should probably go in release announcements.

nrb:
Updated the plugin commands.

I think it may be worth providing instructions/a command to take the current, in-cluster deployment, and update it to an `app/v1` version. It'll have to be copied since the `apiVersion` field can't be modified after creation, but I don't feel great about breaking `velero plugin add/remove` here.

With the change to the client's API group as this PR stands, we're _only_modifying an app/v1 deployment, so it would no longer work against existing deployments afaik. So we could look for both or require the deployment API group change.

prydonius:
@nrb if I understand correctly, you're suggesting we need a way to convert a user's existing Deployments that might be in the old apps/v1beta1 group to apps/v1? That's actually not necessary as the Kubernetes API server handles this conversion internally (you can see this by curl'ing the API and results for `apis/apps/v1beta1/deployments`, `apis/extensions/v1beta1/deployments` and `apis/apps/v1/deployments` should all be the same).

When `extensions/v1beta1` and `apps/v1beta1` go away in the next release, Deployments created using that API endpoint will still be accessible using the `apps/v1` APIs.

nrb:
@prydonius I'm thinking about existing users who may update their Velero installation to v1.1.0+ in an existing, sub-1.16 cluster.

The Deployment would still be in an older API group; would our plugin's change in client still work?

prydonius:
It will, yes, as long as the `apps/v1` API is available (i.e. k8s 1.9+).

nrb:
I believe so!

PR #1674, opened 2019-07-17
Document --pod-annotations install argument

### LINKED ISSUE #1667: Document --pod-annotations install flag
PR #1626 added support for adding pod annotations to the Velero deployment and Restic DaemonSet.

This should include adding examples in the CLI and on the website, probably in the AWS section.

nrb:
Fixes #1667

### DIFF
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -143,7 +143,7 @@ This is useful as a starting point for more customized installations.
 
 	# velero install --bucket gcp-backups --provider gcp --secret-file ./gcp-creds.json --wait
 
-	# velero install --bucket backups --provider aws --backup-location-config region=us-west-2 --snapshot-location-config region=us-west-2 --pod-annotations iam.amazonaws.com/role=arn:aws:iam::<AWS_ACCOUNT_ID>:role/<VELERO_ROLE_NAME>
+	# velero install --bucket backups --provider aws --backup-location-config region=us-west-2 --secret-file ./an-empty-file --snapshot-location-config region=us-west-2 --pod-annotations iam.amazonaws.com/role=arn:aws:iam::<AWS_ACCOUNT_ID>:role/<VELERO_ROLE_NAME>
 
 		`,
 		Run: func(c *cobra.Command, args []string) {
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -143,6 +143,8 @@ This is useful as a starting point for more customized installations.
 
 	# velero install --bucket gcp-backups --provider gcp --secret-file ./gcp-creds.json --wait
 
+	# velero install --bucket backups --provider aws --backup-location-config region=us-west-2 --snapshot-location-config region=us-west-2 --pod-annotations iam.amazonaws.com/role=arn:aws:iam::<AWS_ACCOUNT_ID>:role/<VELERO_ROLE_NAME>
+
 		`,
 		Run: func(c *cobra.Command, args []string) {
 			cmd.CheckError(o.Validate(c, args, f))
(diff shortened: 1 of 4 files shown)


pkg/cmd/cli/install/install.go
```diff
@@ -143,6 +143,8 @@ This is useful as a starting point for more customized installations.
 
 	# velero install --bucket gcp-backups --provider gcp --secret-file ./gcp-creds.json --wait
 
+	# velero install --bucket backups --provider aws --backup-location-config region=us-west-2 --snapshot-location-config region=us-west-2 --pod-annotations iam.amazonaws.com/role=arn:aws:iam::<AWS_ACCOUNT_ID>:role/<VELERO_ROLE_NAME>
```
nrb: @tlkamp Question for your usage - do you specify a `--secret-file` argument? I notice that #1626 we missed that it's a required argument currently. We might have to do another change where we require either the secret file or annotations.

tlkamp: @nrb we do since it's required, but we are simply sending an empty file (it isn't a big deal for automation since there are no contents to keep in sync 😉 ). I wasn't sure how best to handle this situation in my PR since the presence of additional annotations doesn't necessarily mean another auth. mechanism is being used. 

I think making the `--secret-file` optional and detailing _when_ to provide it (and what to provide) is probably the best happy-medium that can be reached.

nrb: Awesome, thanks for that input. I agree with that assessment, so I’ll see about making that change.

nrb: Another idea I had while thinking about this is to introduce a 2nd flag that explicitly says not to create a secret. Something like `--no-secret` or `--other-auth`.

This way, we don't need to create an empty secret, and it's saying the auth will be handled in some other way without making an assumption about what's in annotations.

I would think this would mean we'd require one of `--secret-file` or this new argument, not both. I also don't think it's relevant to the Helm chart, as it has ways of controlling whether to create secrets or not already.

tlkamp: @nrb that's indeed an option! Thinking back to what @skriss said about wanting to keep flags limited, I think just making `--secret-file` optional achieves the same functionality in a single flag.

site/docs/master/aws-config.md
```diff
@@ -283,25 +283,11 @@ It can be set up for Velero by creating a role that will have required permissio
...
-        template:
-            metadata:
-                labels:
-                    component: velero
-                annotations:
-                    iam.amazonaws.com/role: arn:aws:iam::<AWS_ACCOUNT_ID>:role/<VELERO_ROLE_NAME>
-    ...
-    ```
+```
+velero install --pod-annotations iam.amazonaws.com/role=arn:aws:iam::<AWS_ACCOUNT_ID>:role/<VELERO_ROLE_NAME> [...]
```
skriss: Let's include the full set of `velero install` flags here (essentially copied from the non-kube2iam section)

site/docs/v1.0.0/aws-config.md
```diff
@@ -283,25 +283,11 @@ It can be set up for Velero by creating a role that will have required permissio
       --policy-document file://./velero-policy.json
     ```
 
-4. Update `AWS_ACCOUNT_ID` & `VELERO_ROLE_NAME` with `kubectl edit deploy/velero -n velero` and add the following annotation:
+4. Use the `--pod-annotations` argument on `velero install` to add the following annotation:
```
skriss: This is going to ship in v1.1 so we don't want to update the v1.0 docs.

pkg/cmd/cli/install/install.go
skriss: xref #1689

Issue #94, opened 2017-09-14
Spike - Add cpu, memory requests & limits to deployment


ncdc:
Measure ark's memory usage over time, set reasonable values

rosskukulinski:
We should do this pre-1.0 and have some metrics for cluster-size and impact on CPU/Memory.

ncdc:
In case anyone is following the question from @twforeman, we've moved the discussion to #780

skriss:
xref https://github.com/heptio/velero/issues/1452#issuecomment-491848487 (specifically around adding request/limit flags to `velero install`).

skriss:
As part of this issue, we should probably add request/limit flags to `velero install` to be able to specify these.

prydonius:
I've been monitoring Velero's CPU and memory usage and in my GKE cluster using Stackdriver I've found that the server pod uses 42M memory and 5m CPU at rest. I installed various applications in the cluster, including the nginx-example in this repo and 10 instances of the WordPress Helm chart to experiment with backup/restore operations. A full cluster backup created the largest spike of 53M memory and 16m CPU (see results of other operations below).

We could generously set requests to 128Mi memory and 100m CPU, and be even more generous with limits (e.g. 256Mi, 200m). What do you think?

| operation | memory (M) | cpu (m) |
|-|-|-|
|backup of example nginx|42|8.4|
|backup 1 instance of Helm wordpress|42|8.4|
|backup 10 instances of Helm wordpress|42|12|
|full cluster backup|53|16|
|delete backup|42|6.7|
|delete multiple backups|53|16|
|multiple backup requests (4)|42|14|
|restore nginx-example|53|8.6|

rosskukulinski:
cool!

I wonder if the number of backups/restores that Velero is tracking has an impact on cpu/memory?

What about Restic? I assume the daemonset consumes much more CPU/memory when doing filesystem backup & restore.

skriss:
@prydonius this is great. Let's continue to collect some more info.

> I wonder if the number of backups/restores that Velero is tracking has an impact on cpu/memory?

Yeah, I imagine this will have an impact - let's look at it.

Things will definitely get fuzzier with restic. The daemonset pods are one thing (that's where restic backups and  restores run), but  also the `restic prune` operation probably consumes a decent amount  of resources for larger repos, and that runs in the main velero pod.

prydonius:
> > I wonder if the number of backups/restores that Velero is tracking has an impact on cpu/memory?
> 
[...]

PR #1677, opened 2019-07-17
Add resource limits to restic init container

### LINKED ISSUE #1201: Restic initContainer has no limits, can't be restored into namespaces with resource quotas
The restic init container defined at https://github.com/heptio/velero/blob/master/pkg/restore/restic_restore_action.go#L81-L103 is hard code, and restoring pods into namespaces that have quotas enabled results in an error such as this:

```
    test-namespace:  error restoring /tmp/671807868/resources/pods/namespaces/test-namespace/solr-trigger-test-0.json: pods "solr-trigger-test-0" is forbidden: failed quota: compute-resources: must specify limits.cpu,limits.memory,requests.cpu,requests.memory
```

I can think of 2 possible solutions off the top of my head:

* Allow the administrator to specify spec details for restic's initContainer
* Set limits on the initContainer in the code

nrb:
Fixes #1201

* [ ] Add tests for the plugin's Execute method.


pkg/restore/restic_restore_action.go
```diff
@@ -107,6 +108,12 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
 				},
 			},
 		},
+		Resources: corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("100m"),
```
skriss: We should consider making these optionally configurable via the existing ConfigMap for this plugin.

nrb: I was thinking about that, but seeing how valid quantities in the wrong fields can break with #1678 gives me pause on the failure mode.

For example, if we allow the user to specify the limits and/or requests and they put `cpu: 128Mi`, it's a valid quantity string, but not for CPU. `MustParse` and `ParseQuantity` only seem to care about the string. So far I don't know that there's a way to find out about that until the pod fails to get scheduled, which I think would cause the Velero Restore to get stuck InProgress.

I'll do some more digging in the Kubernetes code to see if there's a helper function that we may have missed.

nrb:
Added 2 test cases and the necessary helpers for creating initContainers. Let me know if this approach makes sense.

skriss:
Approach LGTM.

Looking back at the original issue, looks like we need to add requests *and* limits (see the error message).

Is the `1Mi` mem limit sufficient? Let's just make sure it's big enough that we won't run into issues.

nrb:
> Is the 1Mi mem limit sufficient? Let's just make sure it's big enough that we won't run into issues.

I need to actually run this against a cluster to be sure, was mostly making sure I got compilation working first.

Do you have any more thoughts about the failure mode when allowing users to configure the values?

skriss:
> Do you have any more thoughts about the failure mode when allowing users to configure the values?

Let me think about this a little more.

skriss:
Per other PR, I think if we (a) validate that the request <= the limit for both CPU and mem, and (b) put some generous upper bounds on each (e.g. 1 CPU, 128Mi mem?), then we can allow user-configuration. If their values don't conform to these validations, we can log an error and proceed with our own defaults

nrb:
#1577 and #1678 should be merged before this.

carlisia:
lgtm so far 👍

prydonius:
@nrb should I wait for this before adding flags to install for restic requests/limits (#94), since I think you wanted to move around the `parseResourceRequests` helper? https://github.com/heptio/velero/blob/master/pkg/cmd/cli/install/install.go#L262

nrb:
@prydonius That'll work - I'll get this out of draft today.

pkg/restore/restic_restore_action.go
```diff
@@ -162,6 +159,28 @@ func getImage(log logrus.FieldLogger, config *corev1.ConfigMap) string {
...
+func getResourceRequests(log logrus.FieldLogger, config *corev1.ConfigMap) (string, string) {
+	if config == nil {
+		log.Debug("No config found for plugin")
+		return "", ""
+	}
+
+	return config.Data["cpuRequest"], config.Data["memRequest"]
+}
+
+// getResourceRequests extracts the CPU and memory limits from a ConfigMap.
```
skriss: getResourceLimits

nrb:
@skriss So I'm trying to test it now, but I think `master` might be in a broken state with restic right now.

We're not adding the snapshot annotations to the pod anymore, but the restore action code still aborts if it's not present. (https://github.com/heptio/velero/blob/master/pkg/restore/restic_restore_action.go#L71-L72). The pod comes back unaltered.

skriss:
Ah, you're right, we should've continued to write the annotation in the first PR. @carlisia how's progress on #1691? Can we do anything to help?

skriss:
We should probably document the additional ConfigMap fields [here](https://github.com/heptio/velero/blob/master/site/docs/master/restic.md#customize-restore-helper-image).

nrb:
I think this is currently still broken; the init container never finishes, as it doesn't find the Velero directories.

```
% k logs deployment/nginx-deployment -n nginx-example --container restic-wait
Not found: /restores/nginx-logs/.velero/1bdff6db-b553-11e9-9cae-42010a96001e
Not found: /restores/nginx-logs/.velero/1bdff6db-b553-11e9-9cae-42010a96001e
Not found: /restores/nginx-logs/.velero/1bdff6db-b553-11e9-9cae-42010a96001e
Not found: /restores/nginx-logs/.velero/1bdff6db-b553-11e9-9cae-42010a96001e
Not found: /restores/nginx-logs/.velero/1bdff6db-b553-11e9-9cae-42010a96001e
Not found: /restores/nginx-logs/.velero/1bdff6db-b553-11e9-9cae-42010a96001e
Not found: /restores/nginx-logs/.velero/1bdff6db-b553-11e9-9cae-42010a96001e
[....repeated...]
```

```
% k exec nginx-deployment-757ddf98bb-7xj4b  -n nginx-example --container restic-wait -it -- bash
nobody@nginx-deployment-757ddf98bb-7xj4b:/$ ls
bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  restores  root  run  sbin  srv  sys  tmp  usr  var  velero-restic-restore-helper
nobody@nginx-deployment-757ddf98bb-7xj4b:/$ cd restores
nobody@nginx-deployment-757ddf98bb-7xj4b:/restores$ ls
nginx-logs
nobody@nginx-deployment-757ddf98bb-7xj4b:/restores$ cd nginx-logs/
nobody@nginx-deployment-757ddf98bb-7xj4b:/restores/nginx-logs$ ls
lost+found
nobody@nginx-deployment-757ddf98bb-7xj4b:/restores/nginx-logs$ ls -al
total 24
drwxr-xr-x 3 root root  4096 Aug  2 18:27 .
drwxr-xr-x 3 root root  4096 Aug  2 18:27 ..
drwx------ 2 root root 16384 Aug  2 18:27 lost+found
```

I don't _believe_ my changes caused this, but perhaps the change to the builders may have left out some info.

nrb:
With the latest code, I'm able to restore into a namespace with a quota.

```
k describe quota -n nginx-example
Name:                       gke-resource-quotas
Namespace:                  nginx-example
Resource                    Used  Hard
--------                    ----  ----
count/ingresses.extensions  0     100
count/jobs.batch            0     5k
pods                        1     1500
services                    1     500


Name:          test
Namespace:     nginx-example
Resource       Used   Hard
--------       ----   ----
limits.cpu     100m   2
limits.memory  128Mi  2Gi
```

nrb:
@skriss Ok, all documentation should be updated now.

prydonius:
@skriss @nrb is this ok to merge for the beta?

skriss:
yep - merging

PR #1678, opened 2019-07-17
add configurable CPU/memory requests/limits for velero pod on install

prydonius:
Adds configurable CPU/memory requests and limits to the Velero Deployment generated by `velero install`.

By default, requests of 100m, 128Mi and limits of 200m, 256Mi are used.

ref #94


pkg/install/resources.go
```diff
@@ -37,7 +37,13 @@ func imageVersion() string {
 }
 
 // DefaultImage is the default image to use for the Velero deployment and restic daemonset containers.
-var DefaultImage = "gcr.io/heptio-images/velero:" + imageVersion()
+var (
+	DefaultImage               = "gcr.io/heptio-images/velero:" + imageVersion()
+	DefaultVeleroPodCPURequest = "100m"
```
prydonius: Is there a reason we want to define these defaults here instead of simply hardcoding them in the install command?

pkg/cmd/cli/install/install.go
```diff
@@ -237,3 +257,41 @@ func (o *InstallOptions) Validate(c *cobra.Command, args []string, f client.Fact
 
 	return nil
 }
+
+// parseResourceRequests takes a set of CPU and memory requests and limit string
+// values and returns a ResourceRequirements struct to be used in a Container.
+// An error is returned if we cannot parse the request/limit.
+func parseResourceRequests(cpuRequest, memRequest, cpuLimit, memLimit string) (corev1.ResourceRequirements, error) {
```
prydonius: it might make sense to move this to a helper package, though I'm not sure where it could go

pkg/cmd/cli/install/install.go
```diff
@@ -95,13 +109,19 @@ func (o *InstallOptions) AsVeleroOptions() (*install.VeleroOptions, error) {
 	if err != nil {
 		return nil, err
 	}
+	veleroPodResources, err := parseResourceRequests(o.VeleroPodCPURequest, o.VeleroPodMemRequest, o.VeleroPodCPULimit, o.VeleroPodMemLimit)
```
prydonius: parseResourceRequests takes parameters instead of simply getting them from the struct because it will need to be used for parsing restic pod resources too

pkg/cmd/cli/install/install.go
```diff
@@ -237,3 +257,41 @@ func (o *InstallOptions) Validate(c *cobra.Command, args []string, f client.Fact
...
 	return nil
 }
+
+// parseResourceRequests takes a set of CPU and memory requests and limit string
+// values and returns a ResourceRequirements struct to be used in a Container.
+// An error is returned if we cannot parse the request/limit.
+func parseResourceRequests(cpuRequest, memRequest, cpuLimit, memLimit string) (corev1.ResourceRequirements, error) {
+	var resources corev1.ResourceRequirements
+
+	parsedCPURequest, err := resource.ParseQuantity(cpuRequest)
```
prydonius: FYI it's possible to have an invalid CPU/memory request parsed successfully here that is later not accepted by the Kubernetes API. For example, a CPU value of "128Mi" would parse here (since it's a valid quantity for memory), but the API server will return the following error:

```
Error installing Velero. Use `kubectl logs deploy/velero -n velero` to check the deploy logs: Error creating resource Deployment/velero: Deployment.apps "velero" is invalid: spec.template.spec.containers[0].resources.requests: Invalid value: "5Mi": must be less than or equal to cpu limit
```

I think this is okay as the API server error is pretty clear, and there doesn't seem to be a way to parse CPU/memory values independently.

pkg/install/resources.go
nrb: I don't remember exactly why I put the DefaultImage here. I think it may have been either referenced elsewhere in the codebase and I consolidated it here, or I was thinking external consumers of the install package could use it without relying on the CLI.

pkg/cmd/cli/install/install.go
nrb: Looks like we can parse them separately, but there's no helpers to do the requests <= limits check.

While I agree the message is fairly clear, I think it results in an incomplete install state - e.g. some things are created on the API server and some aren't. It's possible that this is already the case with some other errors during installation and not worth addressing here, though.

pkg/install/resources.go
prydonius: Looks like it's only being used in the install command package right now. I guess it could be useful for external consumers, though.

pkg/cmd/cli/install/install.go
prydonius: @nrb hmm couldn't find how to parse them separately, can you point me to where you found that?

nrb: Maybe I'm misunderstanding what you mean by separately, but this code is using `ParseQuantity` for each of the CPU/memory requests/limits (so 4 total calls for 4 total values).

prydonius: Sorry, what I meant is that I cannot parse just CPU quantities or just Memory quantities to avoid the issue where I could set CPU to "128Mi" which would be invalid.

nrb: Ahh, I see. Got it.

prydonius:
@skriss @carlisia PTAL

pkg/install/resources.go
skriss: I think it makes sense to leave it where it is given the current package structure.

pkg/cmd/cli/install/install.go
```diff
@@ -62,6 +68,10 @@ func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Prefix, "prefix", o.Prefix, "prefix under which all Velero data should be stored within the bucket. Optional.")
 	flags.Var(&o.PodAnnotations, "pod-annotations", "annotations to add to the Velero and Restic pods. Optional. Format is key1=value1,key2=value2")
 	flags.StringVar(&o.Namespace, "namespace", o.Namespace, "namespace to install Velero and associated data into. Optional.")
+	flags.StringVar(&o.VeleroPodCPURequest, "velero-pod-cpu-request", o.VeleroPodCPURequest, "CPU request for Velero pod. Optional.")
```
skriss: Can a user explicitly specify "0" for all of these, and get a Velero deployment without reqs/limits?

pkg/cmd/cli/install/install.go
```diff
@@ -231,3 +251,41 @@ func (o *InstallOptions) Validate(c *cobra.Command, args []string, f client.Fact
...
+
+// parseResourceRequests takes a set of CPU and memory requests and limit string
+// values and returns a ResourceRequirements struct to be used in a Container.
+// An error is returned if we cannot parse the request/limit.
+func parseResourceRequests(cpuRequest, memRequest, cpuLimit, memLimit string) (corev1.ResourceRequirements, error) {
+	var resources corev1.ResourceRequirements
+
+	parsedCPURequest, err := resource.ParseQuantity(cpuRequest)
+	if err != nil {
+		return resources, errors.WithMessage(err, "couldn't parse CPU request")
```
skriss: let's switch these to use `errors.Wrap(...)` instead of `errrors.WithMessage(...)` so we get a stack trace as well.

pkg/cmd/cli/install/install.go
prydonius: Just double checked, yes this is possible.

prydonius:
> If I'm reading all the comments correctly, you're saying that these defaults are fine as a baseline even when using restic?

Yes, even with restic this appears to be a good baseline for the Velero Pod. I did see slightly larger CPU/memory spikes when prune occurs (https://github.com/heptio/velero/issues/94#issuecomment-512483346), but it is still way under the default requests.

> We'll also want to do some docs updates to cover this and explain that it may need to be tuned.

+1 I wasn't sure where this should go, a new section under the install overview or a mention under the "Cloud provider" section in the overview?

pkg/cmd/cli/install/install.go
prydonius: FYI, I changed this but I'm not seeing a stack trace printed with the error

pkg/cmd/cli/install/install.go
skriss: OK I see what's going on here.  Technically, "128Mi" isn't invalid for CPU, since there aren't actually units attached to these quantities - it just means a whole lot more CPUs than any machine actually has.

I think we should:
- *definitely* validate that the request <= the limit for both CPU and mem -- this will help avoid fat-finger errors
- *consider* adding reasonable upper bounds (e.g. CPU <= 8 (?), mem <= 64Gi (?)), again to catch blatant fat-finger errors (don't take my numbers as correct, just illustrative)

@nrb whatever we do here is probably relevant for #1677 as well, if we decide to allow user-configurable vals there

pkg/cmd/cli/install/install.go
skriss: I guess we're probably just outputting it to stderr rather than using logrus so we don't get location output, which is fine here. In any case, we pretty much exclusively use `WithStack` or `Wrap` across the code base so no reason to deviate.

skriss:
I think a new section in the Install Overview page makes sense. I think we probably also want to include links to that section from each of the provider-specific install pages, just under where the sample `velero install` commands are shown.

pkg/cmd/cli/install/install.go
nrb: Probably worth a function that we share to do the request <= limit comparison.

prydonius: Added a check for request <= limit, and added some tests around it.

> Probably worth a function that we share to do the request <= limit comparison.

 Currently just added it in the same parseResourceRequests function, do you think it makes sense to split it up?

prydonius: > consider adding reasonable upper bounds (e.g. CPU <= 8 (?), mem <= 64Gi (?)), again to catch blatant fat-finger errors (don't take my numbers as correct, just illustrative)

I guess this doesn't hurt, but I'm not sure how useful it would actually be. It's pretty clear when providing these flags what quantity (CPU or memory) is being set, and it might be a little arbitrary to define an upper bound. It might be better to check if the unit provided is acceptable for the quantity type (e.g. a simple check could just be to ensure the CPU quantity either has no unit, or the unit == "m")

nrb: >Could #1677 make use of that whole function?

Yeah, it could.

skriss: I think it depends what we're trying to protect against. To be clear -- "128Mi" is a valid request for CPU, as is "100m" for memory. The suffixes do not imply "cores" or "bytes", they're just shorthand for expressing unit-less numbers. However, a pod with a "128Mi" CPU request won't be scheduled unless you have a node with a hundred million+ CPUs :) The original error you posted wasn't complaining about using Mi for CPU, only about the fact that the request was bigger than the limit.

As long as we ensure request <= limit, then the pod will be accepted by the API server, though it may still not be scheduled if the requests are too high. This would be the case if the CPU request was, for example, "100", or "100Mi" (assuming no node with >=100 cores).

So if we just want to ensure the API server accepts the object, then I think we're good as-is. If we want to do some logical validation on the magnitude of the values, then we should decide what the boundaries of reasonable are there.

skriss: `kubectl apply` this pod, then describe it:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: resource-request
spec:
  containers:
  - image: gcr.io/heptio-images/fsfreeze-pause:latest
    imagePullPolicy: Always
    name: fsfreeze
    resources:
      requests:
        cpu: 1Mi
  restartPolicy: Always
  serviceAccount: default
  serviceAccountName: default
```

skriss: BTW, I think we're now fine as-is with the req<=lim check in place.

prydonius: > "128Mi" is a valid request for CPU, as is "100m" for memory. The suffixes do not imply "cores" or "bytes", they're just shorthand for expressing unit-less numbers.

Ah I see.

> So if we just want to ensure the API server accepts the object, then I think we're good as-is. If we want to do some logical validation on the magnitude of the values, then we should decide what the boundaries of reasonable are there.

Ultimately I think we want to result in a valid install, and ideally one that is not only accepted by the API server but is also configured correctly. That said, we can't guarantee that Velero will run with _any_ defined requests/limits (even the default) as we'd need to know what resources are available in the cluster, so I think there is a still an element of user intervention here to debug things if something went wrong. This is slightly different to a case where a user accidentally uses the wrong unit, but I think the end result is the same (pod can't be scheduled, user needs to debug at the Kubernetes level).

skriss: Yep - and I do agree that any boundary we put in place will be arbitrary and may be problematic for a certain user.  Given that a `kubectl describe` will be pretty clear about there being insufficient resources to schedule the pod if the requests are too high, I think this is an acceptable UX.

skriss:
This LGTM pending docs updates! We'll do some testing and make sure the defaults seem reasonable across a few different environments before we release.

prydonius:
@skriss @nrb  added documentation - PTAL!

prydonius:
@skriss @nrb updated the req/limit for CPU here based on https://github.com/heptio/velero/issues/94#issuecomment-514794326

prydonius:
@nrb added and rebased

Issue #396, opened 2018-03-20
Increase visibility of what's inside a backup


containscafeine:
I might have missed something, but from what I found out from the docs is that when I want to look inside a backup, I run `ark backup describe <name>`
I think that the information this gives me is pretty limited.

I'd expect to see to see the Kubernetes resources that have been backed up. Something like this -
```console
$ ark backup describe wp
Name:         wp
Namespace:    heptio-ark
Labels:       <none>
Annotations:  <none>

Namespaces:
  Included:  *
  Excluded:  <none>

Resources:
  Included:        *
  Excluded:        <none>
  Cluster-scoped:  auto

Label selector:  <none>

Snapshot PVs:  auto

TTL:  720h0m0s

Hooks:  <none>

Phase:  New

Backup Format Version:  0

Expiration:  0001-01-01 00:00:00 +0000 UTC

Validation errors:  <none>

Persistent Volumes: <none included>

Resources:
  Deployments:
  - database
  - wordpress
  Services:
  - database (3306/TCP)
  - wordpress (80:31229/TCP)
  Secrets:
  - database-root-password
  - database-user-password
  ConfigMaps:
  - database
```

This would really help me keep track of backups and what to restore at what time.

ncdc:
We'd need to figure out how to do this without blowing up etcd. Etcd values are capped at 1.5MB, so if you're backing up an entire cluster, I don't think we can store all of that information on the backup object itself. We could include a manifest in object storage, and have `describe` download the manifest and display it.

containscafeine:
@ncdc that makes sense. How do I start hacking on this? I've been trying to read through the code, but I'm not super familiar with how it is structured. Where should I start looking? Also, would love if there is an architecture doc that'd guide me through this.

ncdc:
If you just want to hack, you could either add a method to the BackupService interface:

https://github.com/heptio/ark/blob/master/pkg/cloudprovider/backup_service.go#L38-L65

or add yet another arg to the UploadBackup method.

This is where we call UploadBackup:

https://github.com/heptio/ark/blob/master/pkg/controller/backup_controller.go#L371

We do plan on creating some architecture docs. Just not done yet...

robbyt:
It would be great if there was more information about the backup in metadata stored in the bucket. (not in etcd) e.g., the contents of the k8s manifests, and even file data from volumes (file names, dates, and hashes(!))

ncdc:
[...]

PR #1682, opened 2019-07-19
proposal: expose list of backed up resources in backup details

prydonius:

ref #396

### DIFF
--- a/design/backup-resource-list.md
+++ b/design/backup-resource-list.md
@@ -1,6 +1,6 @@
 # Expose list of backed up resources in backup details
 
-Status: Draft
+Status: Accepted
 
 To increase the visibility of what a backup might contain, this document proposes storing metadata about backed up resources in object storage and adding a new section to the detailed backup description output to list them.
 
--- a/design/backup-resource-list.md
+++ b/design/backup-resource-list.md
@@ -59,8 +59,8 @@ This property will be moved to the [Backup request struct](https://github.com/he
 `backedUpItems` currently uses the `schema.GroupResource` as a key for the resource.
 In order to record the API group, version and kind for the resource, this key will be constructed from the object's `schema.GroupVersionKind` in the format `{group}/{version}/{kind}` (e.g. `apps/v1/Deployment`).
 
-The `backedUpItems` map is currently a flat structure and will need to be converted to the nested structure above, grouped by `schema.GroupVersionKind` string.
-
+The `backedUpItems` map is kept as a flat structure internally for quick lookup.
+When the backup is ready to upload, `backedUpItems` will be converted to a nested structure representing the metadata file above, grouped by `schema.GroupVersionKind`.
 After converting to the right format, it can be passed to the `persistBackup` function to persist the file in object storage.
 
 ### Changes to DownloadRequest CRD and processing
--- a/design/backup-resource-list.md
+++ b/design/backup-resource-list.md
@@ -33,18 +33,18 @@ A section will be added to the output of `velero backup describe <name> --detail
 This metadata will be in JSON (or YAML) format so that it can be easily inspected from the bucket outside of Velero tooling, and will contain the API resource and group, namespaces and names of the resources:
 
 ```
-deployments.apps:
+apps/v1/Deployment:
 - default/database
 - default/wordpress
-services:
+v1/Service:
 - default/database
 - default/wordpress
-secrets:
+v1/Secret:
 - default/database-root-password
 - default/database-user-password
-configmaps:
+v1/ConfigMap:
 - default/database
-persistentvolumes:
+v1/PersistentVolume:
 - my-pv
 ```
 
@@ -56,7 +56,10 @@ The top-level key is the string form of the `schema.GroupResource` type that we 
 The Backupper currently initialises a map to track the `backedUpItems` (https://github.com/heptio/velero/blob/1594bdc8d0132f548e18ffcc1db8c4cd2b042726/pkg/backup/backup.go#L269), this is passed down through GroupBackupper, ResourceBackupper and ItemBackupper where ItemBackupper records each backed up item.
 This property will be moved to the [Backup request struct](https://github.com/heptio/velero/blob/16910a6215cbd8f0bde385dba9879629ebcbcc28/pkg/backup/request.go#L11), allowing the BackupController to access it after a successful backup.
 
-The `backedUpItems` map is currently a flat structure and will need to be converted to the nested structure above, grouped by `schema.GroupResource`.
+`backedUpItems` currently uses the `schema.GroupResource` as a key for the resource.
+In order to record the API group, version and kind for the resource, this key will be constructed from the object's `schema.GroupVersionKind` in the format `{group}/{version}/{kind}` (e.g. `apps/v1/Deployment`).
+
+The `backedUpItems` map is currently a flat structure and will need to be converted to the nested structure above, grouped by `schema.GroupVersionKind` string.
 
 After converting to the right format, it can be passed to the `persistBackup` function to persist the file in object storage.
 
@@ -76,8 +79,6 @@ If we choose to store the metadata in YAML, it can likely be directly printed ou
 If the metadata file does not exist, this is an older backup and we cannot display the list of resources that were backed up.
 
 ## Open Questions
-
-- Do we want to show the group version in the metadata (i.e. do we want users to see `apps/v1/Deployment` instead of `deployments.apps`)? If so, we wouldn't be able to use the existing list of backedUpItems list, or we'd need to change it to track schema.GroupVersionKinds instead.
 
 ## Alternatives Considered
 
--- a/design/backup-resource-list.md
+++ b/design/backup-resource-list.md
@@ -91,6 +91,6 @@ Additionally, if we wanted to expose more information about the backed up resour
 
 The disadvantages are:
 - downloading the whole backup archive will be larger than just downloading a smaller file with metadata
-- doesn't solve the usecase of wanting to see the list of backed up resources in the bucket outside of Velero tooling
+- reduces the metadata available in the bucket that users might want to inspect outside of Velero tooling (though this is not an explicit requirement)
 
 ## Security Considerations
--- a/design/backup-resource-list.md
+++ b/design/backup-resource-list.md
@@ -1,0 +1,92 @@
+# Expose list of backed up resources in backup details
+
+Status: Draft
+
+To increase the visibility of what a backup might contain, this document proposes storing metadata about backed up resources in object storage and adding a new section to the detailed backup description output to list them.
+
+## Goals
+
+- Include a list of backed up resources as metadata in the bucket
+- Enable users to get a view of what resources are included in a backup using the Velero CLI
+
+## Non Goals
+
+- Expose the full manifests of the backed up resources
+
+## Background
+
+As reported in #396, the information reported in a `velero backup describe <name> --details` command is fairly limited, and does not easily describe what resources a backup contains.
+In order to see what a backup might contain, a user would have to download the backup tarball and extract it.
+This makes it difficult to keep track of different backups in a cluster.
+
+## High-Level Design
+
+After performing a backup, a new file will be created that contains the list of the resources that have been included in the backup.
+This file will be persisted in object storage alongside the backup contents and existing metadata.
+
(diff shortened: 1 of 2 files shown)


design/backup-resource-list.md
```diff
@@ -0,0 +1,92 @@
...
+configmaps:
+- default/database
+```
+
+The filename for this metadata will be `<backup name>-resource-list.json`.
+The top-level key is the string form of the `schema.GroupResource` type that we currently keep track of in the backup controller code path.
+
+### Changes in Backup controller
+
+The Backupper currently initialises a map to track the `backedUpItems` (https://github.com/heptio/velero/blob/1594bdc8d0132f548e18ffcc1db8c4cd2b042726/pkg/backup/backup.go#L269), this is passed down through GroupBackupper, ResourceBackupper and ItemBackupper. Moving the map initialisation to the BackupController will provide access to the resulting list there after a successful backup.
```
skriss: We're already a little bit inconsistent with how we return information from `pkg/backup` to the backup controller, but I think my preferred approach here would be to add a field to the `pkg/backup.Request` struct to store backed-up items. This can get populated by `pkg/backup` during execution of the `Backup(...)` method, and then the backup controller can read it when it goes to persist.  This is consistent with how volume snapshot metadata is passed currently.  

Alternately, we could define a new struct to capture returned info about the backup that would include volume snapshots, restic backup data (WIP by carlisia), and this info.

WDYT?

design/backup-resource-list.md
```diff
@@ -0,0 +1,92 @@
...
+This command will need to be updated to fetch the metadata from the bucket using the `Stream` method used in other commands.
+The file will be read in memory and displayed in the output of the command.
+Depending on the format the metadata is stored in, it may need processing to print in a more human-readable format.
+If we choose to store the metadata in YAML, it can likely be directly printed out.
+
+If the metadata file does not exist, this is an older backup and we cannot display the list of resources that were backed up.
+
+## Open Questions
+
+- Do we want to show the group version in the metadata (i.e. do we want users to see `apps/v1/Deployment` instead of `deployments.apps`)? If so, we wouldn't be able to use the existing list of backedUpItems list, or we'd need to change it to track schema.GroupVersionKinds instead.
```
skriss: I think this would be nice to have -- particularly for the use case where a user is restoring an older backup into a new cluster, and needs to ensure that the backed-up API versions exist in their new cluster before restoring. What do others think?

design/backup-resource-list.md
prydonius: Ah, I thought the backup.Request was the CR, but I see now that it's a different internal struct. I agree that it makes sense to use this or another struct to pass down the result.

Defining a new struct (i.e. backup.Response) will help make it more clear what it is used for, but using the Request struct isn't too terrible and won't require as many changes. What do others think?

design/backup-resource-list.md
prydonius: That's a very good point, do you think we can change backedUpItems to track with versions? I wasn't sure if it relies on unambiguous versions to make sure there are no duplicates?

skriss: Velero only backs up the server's "preferred version" of each API group, so in theory we should only have one version per resource. We could update the `itemKey` struct to use a GVR rather than just a GR, which should work OK (need to figure out details though).

design/backup-resource-list.md
skriss: 👍using the `backup.Request` WFM.

design/backup-resource-list.md
```diff
@@ -0,0 +1,92 @@
...
+- default/database
+- default/wordpress
+services:
+- default/database
+- default/wordpress
+secrets:
+- default/database-root-password
+- default/database-user-password
+configmaps:
+- default/database
```
nrb: An example of cluster-scoped resources would be helpful.

I assume it would look something like:

```yaml
persistentvolumes:
- my-pv
```

design/backup-resource-list.md
nrb: Agreed that writing the version of the backed up item would be nice, especially since we should only have one per item.

design/backup-resource-list.md
```diff
@@ -0,0 +1,92 @@
...
+### Fetch backup contents archive and walkthrough to list contents
+
+Instead of recording new metadata about what resources have been backed up, we could simply download the backup contents archive and walkthrough it to list the contents everytime `velero backup describe <name> --details` is run.
+
+The advantage of this approach is that we don't need to change any backup procedures as we already have this content, and we will also be able to list resources for older backups.
+Additionally, if we wanted to expose more information about the backed up resources, we can do so without having to update what we store in the metadata.
+
+The disadvantages are:
+- downloading the whole backup archive will be larger than just downloading a smaller file with metadata
+- doesn't solve the usecase of wanting to see the list of backed up resources in the bucket outside of Velero tooling
```
nrb: I'm not 100% sure this is a design goal for Velero as a whole, especially if/when we add encryption. That said, we've also never purposefully worked to obfuscate what we read from the bucket, so using tar listing is going against the status quo.

design/backup-resource-list.md
```diff
@@ -0,0 +1,92 @@
...
+- Include a list of backed up resources as metadata in the bucket
+- Enable users to get a view of what resources are included in a backup using the Velero CLI
+
+## Non Goals
+
+- Expose the full manifests of the backed up resources
+
+## Background
+
+As reported in #396, the information reported in a `velero backup describe <name> --details` command is fairly limited, and does not easily describe what resources a backup contains.
```
nrb: May be worth adding the actual link to the issue at `#396`.

prydonius: huh, I thought it would be added by GitHub directly, will do!

design/backup-resource-list.md
prydonius: I can't remember where I got this from, but I agree this is not exactly a usecase we're trying to solve for. Let me reword this to reduce the emphasis on this point.

prydonius:
Updated to mention changing backedUpItems resource key to using schema.GroupVersionKind instead.

prydonius:
@skriss @nrb PTAL!

design/backup-resource-list.md
```diff
@@ -0,0 +1,96 @@
...
+
+The filename for this metadata will be `<backup name>-resource-list.json`.
+The top-level key is the string form of the `schema.GroupResource` type that we currently keep track of in the backup controller code path.
+
+### Changes in Backup controller
+
+The Backupper currently initialises a map to track the `backedUpItems` (https://github.com/heptio/velero/blob/1594bdc8d0132f548e18ffcc1db8c4cd2b042726/pkg/backup/backup.go#L269), this is passed down through GroupBackupper, ResourceBackupper and ItemBackupper where ItemBackupper records each backed up item.
+This property will be moved to the [Backup request struct](https://github.com/heptio/velero/blob/16910a6215cbd8f0bde385dba9879629ebcbcc28/pkg/backup/request.go#L11), allowing the BackupController to access it after a successful backup.
+
+`backedUpItems` currently uses the `schema.GroupResource` as a key for the resource.
```
nrb: Is there a function that allows for easy parsing of this format back into a GVR?

prydonius: Do you mean converting from `schema.GroupResource` to `schema.GroupVersionKind` or the other way around? We can't get the GVK from the GroupResource, but we can get it from the `runtime.Unstructured` obj. We don't really need the GroupResource, AFAICT we're only using it as a key in backedUpItems and not using it for anything else after.

nrb: Sorry, should have commented lower down. I meant the `{group}/{version}/{kind}` string, and I meant GVK, not GVR. I suppose writing a deserialization function would be simple enough.

It may not actually be all that important, as we could just print the key out verbatim when displaying details.

prydonius: Ah I see what you mean now, I was just thinking of printing out the key as-is, but we wouldn't be able to directly parse that string (since the GVK.String() method has a slightly different format), but (without looking at the schema pkg) I believe we can parse the `schema.GroupVersion` using `{group}/{version}` and then convert that to GVK using the remaining part.

design/backup-resource-list.md
```diff
@@ -0,0 +1,96 @@
...
+In order to record the API group, version and kind for the resource, this key will be constructed from the object's `schema.GroupVersionKind` in the format `{group}/{version}/{kind}` (e.g. `apps/v1/Deployment`).
+
+The `backedUpItems` map is currently a flat structure and will need to be converted to the nested structure above, grouped by `schema.GroupVersionKind` string.
```
skriss: You may want to consider decoupling the way this data is stored internally in `pkg/backup` from how it gets written to file for display purposes. Having a flat map makes lookup very efficient during backup, whereas having something like a `map[schema.GroupVersionKind][]string` (which I think is what you're proposing), while closer to what you want for view purposes, would result in less efficient lookups during backup. Not a big deal for small clusters, but it could add some overhead in larger ones. I'm fine sorting this out as you move to implementation.

design/backup-resource-list.md
```diff
@@ -0,0 +1,96 @@
...
+The filename for this metadata will be `<backup name>-resource-list.json`.
+The top-level key is the string form of the `schema.GroupResource` type that we currently keep track of in the backup controller code path.
+
+### Changes in Backup controller
+
+The Backupper currently initialises a map to track the `backedUpItems` (https://github.com/heptio/velero/blob/1594bdc8d0132f548e18ffcc1db8c4cd2b042726/pkg/backup/backup.go#L269), this is passed down through GroupBackupper, ResourceBackupper and ItemBackupper where ItemBackupper records each backed up item.
+This property will be moved to the [Backup request struct](https://github.com/heptio/velero/blob/16910a6215cbd8f0bde385dba9879629ebcbcc28/pkg/backup/request.go#L11), allowing the BackupController to access it after a successful backup.
+
+`backedUpItems` currently uses the `schema.GroupResource` as a key for the resource.
+In order to record the API group, version and kind for the resource, this key will be constructed from the object's `schema.GroupVersionKind` in the format `{group}/{version}/{kind}` (e.g. `apps/v1/Deployment`).
```
skriss: Is there a specific reason to use Kind rather than Resource? I don't think it makes much of a difference from a UX perspective, but the directories in the tarball do use "resource.group" as the name format so it will look slightly different (capitalization and pluralization generally).

design/backup-resource-list.md
prydonius: You're right, I think I was envisioning that we would do this conversion when creating the metadata file (i.e. in persistBackup), so I think internally in pkg/backup it would keep the flat structure - I'll update this to make that more clear.

design/backup-resource-list.md
prydonius: I guess the difference will be "apps/v1/deployments" vs "apps/v1/Deployment"? No specific reason, but Kind might be easier to read especially for resources with multiple words ("persistentvolume" vs "PersistentVolume").

design/backup-resource-list.md
```diff
@@ -0,0 +1,96 @@
...
+v1/Secret:
+- default/database-root-password
+- default/database-user-password
+v1/ConfigMap:
+- default/database
+v1/PersistentVolume:
+- my-pv
+```
+
+The filename for this metadata will be `<backup name>-resource-list.json`.
```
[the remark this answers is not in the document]
nrb: Should we also store this in the `metadata` dir within the backup structure?

nrb:
I think I'm :+1: on this in general. Once this is in I can see a few more feature requests being generated, such as being able to filter the items returned by this, but that can be addressed later.

design/backup-resource-list.md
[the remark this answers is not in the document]
skriss: It should just go within the backup's directory (`backups/<backup-name>/`)

design/backup-resource-list.md
skriss: Ah, I see - SGTM

design/backup-resource-list.md
skriss: OK, seems fine

design/backup-resource-list.md
prydonius: @skriss reworded a bit to make this more clear

skriss:
LGTM - let's update to "approved" and merge!

PR #1690, opened 2019-07-23
Fixes documentation about aws region.

pierreozoux:
[Reading the code](https://github.com/heptio/velero/blob/master/pkg/cloudprovider/aws/volume_snapshotter.go#L75-L79), you understand that this sentence is misleading.

### DIFF
--- a/site/docs/master/api-types/volumesnapshotlocation.md
+++ b/site/docs/master/api-types/volumesnapshotlocation.md
@@ -40,7 +40,7 @@ The configurable parameters are as follows:
 
 | Key | Type | Default | Meaning |
 | --- | --- | --- | --- |
-| `region` | string | Empty | *Example*: "us-east-1"<br><br>See [AWS documentation][3] for the full list.<br><br>Queried from the AWS S3 API if not provided. |
+| `region` | string | Empty | *Example*: "us-east-1"<br><br>See [AWS documentation][3] for the full list.<br><br>Required. |
 | `profile` | string | "default" | AWS profile within the credential file to use for given store |
 
 #### Azure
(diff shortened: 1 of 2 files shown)


site/docs/master/api-types/volumesnapshotlocation.md
```diff
@@ -40,7 +40,7 @@ The configurable parameters are as follows:
 
 | Key | Type | Default | Meaning |
 | --- | --- | --- | --- |
-| `region` | string | Empty | *Example*: "us-east-1"<br><br>See [AWS documentation][3] for the full list.<br><br>Queried from the AWS S3 API if not provided. |
+| `region` | string | Empty | *Example*: "us-east-1"<br><br>See [AWS documentation][3] for the full list.<br><br>You have to specify it. It is not queried from the AWS S3 API if not provided. |
```
skriss: Let's remove the sentence about it not being queried from the S3 API, since S3 is not relevant here (this is for EBS snapshots). You could just say "Required field."

skriss:
Thanks for the PR @pierreozoux! One minor comment, and you'll also need to add DCO signoff to your commit (`git commit --amend --signoff` followed by `git push -f` would work).

pierreozoux:
done

skriss:
Awesome, thanks. Also just realized that this change should be duplicated in `site/docs/v1.0.0/api-types/volumesnapshotlocation.md` -- we have a separate docs directory per Velero version that we maintain and it would be good to fix in the v1.0.0-tagged docs as well. No need to edit the pre-1.0 docs, though, and since you made the change in master, it will automatically get carried forward to new version. Thanks!

carlisia:
@pierreozoux please add the same change to our `v1.0` docs (`site/docs/v1.0.0/api-types/volumesnapshotlocation.md`)

pierreozoux:
done, sorry for delays.

skriss:
no problem - thanks for the update!

PR #1699, opened 2019-07-25
Make secret file optional on install

### LINKED ISSUE #1689: --secret-file is optionnal for AWS deployment.
Currently when you issue the install command, it requires you to give a `--secret-file` argument.

If you deploy on AWS, the aws sdk go might be able to use the kube2iam or instance role if you don't pass this env var.

As it is a possibility, I think the `secret-file` parameter should be optional and the documentation updated to reflect that.

nrb:
Fixes #1689

### DIFF
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -55,6 +55,7 @@ type InstallOptions struct {
 	ResticPodMemLimit    string
 	RestoreOnly          bool
 	SecretFile           string
+	NoSecret             bool
 	DryRun               bool
 	BackupStorageConfig  flag.Map
 	VolumeSnapshotConfig flag.Map
@@ -67,7 +68,8 @@ type InstallOptions struct {
 func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.ProviderName, "provider", o.ProviderName, "provider name for backup and volume storage")
 	flags.StringVar(&o.BucketName, "bucket", o.BucketName, "name of the object storage bucket where backups should be stored")
-	flags.StringVar(&o.SecretFile, "secret-file", o.SecretFile, "file containing credentials for backup and volume provider. Optional.")
+	flags.StringVar(&o.SecretFile, "secret-file", o.SecretFile, "file containing credentials for backup and volume provider. Cannot be used with --no-secret. Optional.")
+	flags.BoolVar(&o.NoSecret, "no-secret", o.NoSecret, "flag indicating if a secret should be created. Cannot be used with --secret-file. Optional.")
 	flags.StringVar(&o.Image, "image", o.Image, "image to use for the Velero and restic server pods. Optional.")
 	flags.StringVar(&o.Prefix, "prefix", o.Prefix, "prefix under which all Velero data should be stored within the bucket. Optional.")
 	flags.Var(&o.PodAnnotations, "pod-annotations", "annotations to add to the Velero and restic pods. Optional. Format is key1=value1,key2=value2")
@@ -113,7 +115,7 @@ func NewInstallOptions() *InstallOptions {
 // AsVeleroOptions translates the values provided at the command line into values used to instantiate Kubernetes resources
 func (o *InstallOptions) AsVeleroOptions() (*install.VeleroOptions, error) {
 	var secretData []byte
-	if o.SecretFile != "" {
+	if o.SecretFile != "" && !o.NoSecret {
 		realPath, err := filepath.Abs(o.SecretFile)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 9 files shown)


pkg/install/resources.go
```diff
@@ -217,8 +217,10 @@ func AllResources(o *VeleroOptions) (*unstructured.UnstructuredList, error) {
 	sa := ServiceAccount(o.Namespace)
 	appendUnstructured(resources, sa)
 
-	sec := Secret(o.Namespace, o.SecretData)
-	appendUnstructured(resources, sec)
+	if o.SecretData != nil {
```
skriss: If we're not creating the secret, then we need to remove the secret mount from the Deployment/DaemonSet as well (see `WithoutCredentialsVolume`). It would also be nice to not set the credentials environment variables in this case

pkg/install/deployment.go
```diff
@@ -68,6 +69,14 @@ func WithEnvFromSecretKey(varName, secret, key string) podTemplateOption {
 	}
 }
 
+func WithSecretData(secretPresent bool) podTemplateOption {
+	return func(c *podTemplateConfig) {
+		c.withSecretData = secretPresent
```
skriss: Do we need `withSecretData` in addition to `withoutCredentialsVolume`? Seems like they're the same thing.

pkg/cmd/cli/install/install.go
```diff
@@ -86,13 +86,16 @@ func NewInstallOptions() *InstallOptions {
...
 func (o *InstallOptions) AsVeleroOptions() (*install.VeleroOptions, error) {
-	realPath, err := filepath.Abs(o.SecretFile)
-	if err != nil {
-		return nil, err
-	}
-	secretData, err := ioutil.ReadFile(realPath)
-	if err != nil {
-		return nil, err
+	var secretData []byte
+	if o.SecretFile != "" {
```
carlisia: I wonder if it would make sense to try and reduce the possibility of this value not being there by mistake by only making secret optional if the provider is `aws`. WDYT?

pkg/install/deployment.go
```diff
@@ -68,6 +69,14 @@ func WithEnvFromSecretKey(varName, secret, key string) podTemplateOption {
...
+		// Since there's no secret, don't mount the credentials volume in the pods
+		c.withoutCredentialsVolume = true
```
prydonius: shouldn't this only be set if `secretPresent` is `false`?

pkg/cmd/cli/install/install.go
nrb: Currently our only documented use case is AWS, so this is a good point. However, GCP and Azure both have something similar to instance roles/profiles which could theoretically work for Velero, they're just not documented right now.

I've gone back and forth on making a `--no-secret` flag and requiring either that or `--secret-file`, so that it's explicit that the user doesn't want to create a secret rather than simply forgetting the flag. WDYT about that option?

skriss: Another option would be an interactive confirmation prompt ("Are you sure you don't want to create a credentials secret?").

nrb: Also true. But I'd think we wouldn't want to completely close off scripting, so it seems to me like we'd have to do both in that scenario...or make them pipe something to stdin.

prydonius: I like the idea of an explicit `--no-secret` flag

skriss: Yeah..going with just the flag seems reasonable to me.

carlisia: I'd rather not have any prompt during the install. I would not at all mind the one flag OR the other as stated ☝️ and make sure the user knows this upfront - 👍

nrb: Added `--no-secret` in latest commit.

pkg/install/deployment.go
nrb: I've revised this so there's less duplication.

site/docs/master/aws-config.md
```diff
@@ -292,13 +292,10 @@ velero install \
     --pod-annotations iam.amazonaws.com/role=arn:aws:iam::<AWS_ACCOUNT_ID>:role/<VELERO_ROLE_NAME> \
     --provider aws \
     --bucket $BUCKET \
-    --secret-file ./credentials-velero \
     --backup-location-config region=$REGION \
     --snapshot-location-config region=$REGION
 ```
```
skriss: we'd need to specify `--no-secret` here, right?

site/docs/master/install-overview.md
```diff
@@ -23,6 +23,8 @@ velero install \
     [--pod-annotations] \
 ```
 
+The `--secret-file` argument is optional if using node-based IAM policies.
```
skriss: Probably worth noting here as well that `--no-secret` needs to be specified if not using `--secret-file`.

site/docs/master/aws-config.md
nrb: Yep, missed that.

PR #1702, opened 2019-07-25
add pkg/builder with BackupBuilder and ObjectMeta functional options

skriss:

@nrb @prydonius this is what I'm leaning towards based on our discussions yesterday:

- consolidated `pkg/builder` for all of these
- builder structs are named e.g. `BackupBuilder` 
- constructors are called e.g. `ForBackup` (so usage outside the package is `builder.ForBackup(...)`
- constructors take namespace (if namespaced) + name, since this is almost always used
- builders have an `ObjectMeta(...)` method that takes `metav1.Object` functional options
- we'd have builders for both Velero and core Kubernetes types here


skriss:
@nrb @carlisia @prydonius please let me know what you think about this approach for test object construction. I'd like to agree on something so I can make a bunch of updates!

skriss:
> would it make sense to move the example (which it's great having that, btw!) to an actual godoc testable example?

Yeah, works for me!

pkg/builder/backup_builder.go
```diff
@@ -0,0 +1,164 @@
...
+
+*/
+
+// BackupBuilder builds Backup objects.
+type BackupBuilder struct {
+	object *velerov1api.Backup
+}
+
+// ForBackup is the constructor for a BackupBuilder.
+func ForBackup(ns, name string) *BackupBuilder {
```
prydonius: just a thought: we could forgo the `For` prefix, `builder.Backup` still makes sense

skriss: Personally I slightly prefer it with the `For` prefix, but I'm 100% fine with whatever the majority prefers - @nrb @carlisia?

carlisia: In this case, I agree with @skriss and prefer `ForBackup`. If the function was named `Backup`, I'd be left wondering why I'm getting back an object that is not a `Backup`. Adding the `For` answers that question for me.

PR #1703, opened 2019-07-25
Migrate restic backup tests

skriss:
Migrate the remaining `pkg/backup` tests covering restic scenarios into the new structure.


skriss:
The "tracking" is just the internal mechanism by which Velero doesn't take regular snapshots of PVs that are being backed up with restic - so that's exactly what these test cases are testing, just at the "behavior" level rather than based on the internals.

Issue #192, opened 2017-11-08
Support restoring a PV into the same cluster


ncdc:
Forked from #189 

Scenario:
- PVC in namespace _N1_, bound to PV _P1_
- Create backup (snapshotting the PV)
- Restore backup, remapping namespace _N1_ to _N2_
- Have the PVC in _N2_ reference a **new** PV _P2_, created from _P1_'s snapshot

Currently, this is not possible, because Ark does not change the name of the PV when restoring it. The result is 2 PVCs both trying to claim the same (original) PV _P1_.

Questions:
- Should this be done automatically?
- Should this be optional?
- What sort of controls/flags should we expose to the user?
- How do we cover the 80-90% use case?

cc @evaldasou

ncdc:
@rosskukulinski would you find this useful?

dgoodwin:
Our usage is planned to always delete a namespace after backup. (as we're using it to archive it) As such I don't think we'd hit a scenario where we backed up, but still had a live namespace using the PV. 

However restoring with a remapped namespace and preserving the PV would definitely be useful, we don't plan to do this for a bit but it is on the radar. (I forget if Ark can handle remapped namespace with PV snapshots if nothing is still using the PV)

rosskukulinski:
@ncdc reading through #189, yes I think this would be a useful (and expected IMO) feature for CI/CD and testing use-cases.

As a more specific example, this feature request enables what you describe in the [ephemeral ci testing env w/ ark](https://blog.heptio.com/ephemeral-ci-testing-environments-with-heptio-ark-heptioprotip-f22171fa0bba) blog post but supports this behavior within a single cluster.

evaldasou:
wow, awesome! So as @rosskukulinski mentioned , it looks like in Your blog post @ncdc,  [ephemeral ci testing env w/ ark](https://blog.heptio.com/ephemeral-ci-testing-environments-with-heptio-ark-heptioprotip-f22171fa0bba) - it is supported to do this type of restore from one cluster to another? (without deleting original PV/PVC?). I am going to test this!

Regarding new feature, I think there should be optional flag , so we can select to re-create PVC/PV or not. Somethinkg like _--restore-volumes-recreate_ (to recreate PVC/PV) flag.

ncdc:
Current thinking is `ark restore create NAME --clone-volumes` would take each PV `p` in the backup, restore as PV `p'`, and ensure any PVC being restored that references `p` is updated to reference `p'`.

[...]

PR #1705, opened 2019-07-29
draft design doc for cloning PVs while remapping namespaces

skriss:

ref. #192 

cc @arianitu @agolomoodysaada @robertgates55

### DIFF
--- a/design/pv-cloning.md
+++ b/design/pv-cloning.md
@@ -14,7 +14,8 @@ This document proposes a solution that allows new copies of persistent volumes t
 
 ## Non Goals
 
-- Cloning of persistent volumes in any scenario other than when using `velero restore create --namespace-mappings ...` flag. 
+- Cloning of persistent volumes in any scenario other than when using `velero restore create --namespace-mappings ...` flag.
+- [CSI-based cloning](https://kubernetes.io/docs/concepts/storage/volume-pvc-datasource/).
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


skriss:
Adding the annotation sounds like a good idea

arianitu:
This looks good to me. I also notice that this handles actually cloning the PV so both namespaces will continue to work?

Excited for this to be implemented.

design/pv-cloning.md
```diff
@@ -0,0 +1,52 @@
...
+This limitation is documented in detail in [issue #192](https://github.com/heptio/velero/issues/192).
+This document proposes a solution that allows new copies of persistent volumes to be created during a namespace clone.
+
+## Goals
+
+- Enable persistent volumes to be cloned when using `velero restore create --namespace-mappings ...` to create a second copy of a namespace within a cluster.
+
+## Non Goals
+
+- Cloning of persistent volumes in any scenario other than when using `velero restore create --namespace-mappings ...` flag.
```
nrb: May want to add that CSI cloning support is out of scope here.

skriss:
Addressed comments. Is this ready for approval + merge?

design/pv-cloning.md
```diff
@@ -0,0 +1,54 @@
...
+- [CSI-based cloning](https://kubernetes.io/docs/concepts/storage/volume-pvc-datasource/).
+
+## Background
+
+(Omitted, see introduction)
+
+## High-Level Design
+
+During a restore, Velero will detect that it needs to assign a new name to a persistent volume being restored if and only if both of the following conditions are met:
```
nrb: To make sure I understand how a user would do the clone operation, the workflow would look something like this:

1. User does `velero backup create --include-namespaces ns1 src`
1. User then does `velero restore create --from-backup src --namespace-mappings ns1=ns2`

Is that right?

How will this interact with resource selection? If we just target PVs will it work?

skriss: Yeah, that's what the workflow would look like. If you just do `--include-resources pvs`, then the `--namespace-mappings` flag doesn't really make sense since you're not restoring any namespaced resources; in any case, the design as proposed would not create a second copy of the PV in that scenario.

nrb: Got it, that makes sense.

So if the NS already exists, it would behave like a normal Velero restore - anything already existing wouldn't be overridden. I know that's another issue, so I'm fine with this design not addressing it explicitly.

How about interaction with restic? Would this work to create a PVR and copy from the restic repository?

skriss: Correct, yeah this won't do a delete/recreate if other resources already exist.

With restic, we actually avoid this issue entirely, because we rely on a dynamic re-provisioning of an empty PV to be able to do a restic restore into, and the dynamic provisioner will come up with a new PV name. So that should already work today.

skriss:
Updated to "Approved" so ready to merge unless there are any other comments!

skriss:
Alright, I think we have lazy consensus - let's merge this

PR #1707, opened 2019-07-30
Update tests to use object builders in pkg/builder

skriss:
@carlisia @nrb @prydonius I'm not sure what the best approach is for reviewing this. Possibilities:

- I split it up into 10-20 smaller PRs, each one covering a separate object
- I leave it as-is and you sample/scan
- We take the fact that all unit tests are passing as an indication that the updates were correct, and merge as-is

Thoughts?


skriss:
I'd be comfortable with a light review/relying on the tests passing, as this was a lot of find/replace and some manual fixes.

carlisia:
I'm comfortable with a light review. Doing that now.

pkg/controller/schedule_controller_test.go
```diff
@@ -259,41 +262,41 @@ func parseTime(timeString string) time.Time {
 func TestGetNextRunTime(t *testing.T) {
```
nrb: It looks like all the remaining tests in this file use the struct literals.

nrb:
Everything looks good except for some literals in `schedule_controller_test.go`.

pkg/controller/schedule_controller_test.go
skriss: Got those updated

skriss:
OK, took it out of draft status, ready for merge!

PR #1708, opened 2019-07-30
periodically check for stale restic repo locks

### LINKED ISSUE #1511: add restic stale lock checking
Currently, for a variety of reasons it's possible for a restic repository to have a stale lock.  We don't have any way of detecting and handling this currently, so it can lead to things getting "stuck" from a velero perspective.  It can cause a ResticRepository to go into a `NotReady` phase which then blocks all operations on that repository from executing, and I don't think it will ever really self-heal from that situation.  We should add code to the restic repository controller (I think) that uses the `restic unlock` command to remove stale locks.

skriss:

Fixes #1511 

I think this is all we need to do, but I'm still working on testing it.

### DIFF
--- a/pkg/controller/restic_repository_controller.go
+++ b/pkg/controller/restic_repository_controller.go
@@ -120,9 +120,19 @@ func (c *resticRepositoryController) processQueueItem(key string) error {
 	// Don't mutate the shared cache
 	reqCopy := req.DeepCopy()
 
-	switch req.Status.Phase {
-	case "", v1.ResticRepositoryPhaseNew:
+	if req.Status.Phase == "" || req.Status.Phase == v1.ResticRepositoryPhaseNew {
 		return c.initializeRepo(reqCopy, log)
+	}
+
+	// Check both ready and not-ready repos for stale locks, but if this fails
+	// for any reason, it's non-critical so we still continue on to the rest of
+	// the "process" logic.
+	log.Debug("Checking repository for stale locks")
+	if err := c.repositoryManager.UnlockRepo(reqCopy); err != nil {
+		log.WithError(err).Error("Error checking repository for stale locks")
+	}
+
+	switch req.Status.Phase {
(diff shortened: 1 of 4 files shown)


pkg/controller/restic_repository_controller.go
```diff
@@ -120,9 +120,19 @@ func (c *resticRepositoryController) processQueueItem(key string) error {
 	// Don't mutate the shared cache
 	reqCopy := req.DeepCopy()
 
-	switch req.Status.Phase {
-	case "", v1.ResticRepositoryPhaseNew:
+	if req.Status.Phase == "" || req.Status.Phase == v1.ResticRepositoryPhaseNew {
 		return c.initializeRepo(reqCopy, log)
+	}
+
+	// Check both ready and not-ready repos for stale locks, but if this fails
```
carlisia: I would change `Check both ready and not-ready` to be `Check ready or not-ready`. Otherwise it sounds like you are checking multiple repos at a time but it's not the case.

skriss:
I've confirmed that when there are no stale locks, this is a no-op, and when there is a stale lock, this successfully clears it out.

I also tweaked the `ensureRepo` function in the restic repo controller to only try to initialize a repo if the error gotten from attempting to connect to the repo indicates that the repo doesn't exist. We were seeing issues where, if the repo had a stale lock, Velero would try to re-initialize it rather than reporting that it was locked, which just led to confusing error output for the user.

skriss:
One other note - restic considers locks "stale" if they've been around for >30min, even though in the normal case, they get refreshed every 5min. We might want to consider a PR to restic to make this "stale threshold" a CLI flag, so that we could say `restic unlock --older-than 10m`, for example.

skriss:
Hold on merging this, checking one thing.

skriss:
OK, looks fine, I just had some versioning issues.

PR #1709, opened 2019-07-30
store backup resource list metadata in object storage

prydonius:
Adds a new file (`<backup-name>-resource-list.json.gz`) to object storage that contains a list of the resources (API version, kind, namespace and name) that are contained in the backup. E.g.:

```
{
  "apps/v1/ReplicaSet": [
    "nginx-example/nginx-deployment-5b658f5d7f",
    "nginx-example/nginx-deployment-75675f5897"
  ],
  "v1/Endpoints": ["nginx-example/my-nginx"],
  "v1/Namespace": ["nginx-example"],
  "v1/PersistentVolume": ["pvc-dd06dd30-b259-11e9-a63b-025000000001"],
  "v1/PersistentVolumeClaim": ["nginx-example/nginx-logs"],
  "v1/Pod": ["nginx-example/nginx-deployment-5b658f5d7f-28qpb"],
  "v1/Service": ["nginx-example/my-nginx"]
}
```

Also:
- backedUpItems has been moved to the `pkg/backup.Request` struct so it can be accessed by the BackupController
- backedUpItems now forms the resource field in itemKey using the `schema.GroupVersionKind` of the object, so it can be used as keys in the resource list metadata

ref #396

### DIFF
--- a/pkg/backup/backup_test.go
+++ b/pkg/backup/backup_test.go
@@ -53,6 +53,54 @@ import (
 	"github.com/heptio/velero/pkg/volume"
 )
 
+func TestBackedUpItemsMatchesTarballContents(t *testing.T) {
+	// TODO: figure out if this can be replaced with the restmapper
+	// (https://github.com/kubernetes/apimachinery/blob/035e418f1ad9b6da47c4e01906a0cfe32f4ee2e7/pkg/api/meta/restmapper.go)
+	gvkToResource := map[string]string{
+		"v1/Pod":              "pods",
+		"apps/v1/Deployment":  "deployments.apps",
+		"v1/PersistentVolume": "persistentvolumes",
+	}
+
+	h := newHarness(t)
+	req := &Request{Backup: defaultBackup().Result()}
+	backupFile := bytes.NewBuffer([]byte{})
+
+	apiResources := []*test.APIResource{
+		test.Pods(
+			builder.ForPod("foo", "bar").Result(),
+			builder.ForPod("zoo", "raz").Result(),
+		),
+		test.Deployments(
+			builder.ForDeployment("foo", "bar").Result(),
+			builder.ForDeployment("zoo", "raz").Result(),
+		),
+		test.PVs(
+			builder.ForPersistentVolume("bar").Result(),
+			builder.ForPersistentVolume("baz").Result(),
+		),
+	}
+	for _, resource := range apiResources {
+		h.addItems(t, resource)
+	}
+
+	h.backupper.Backup(h.log, req, backupFile, nil, nil)
+
+	var expectedFiles []string
+	for item := range req.BackedUpItems {
+		file := "resources/" + gvkToResource[item.resource]
+		if item.namespace != "" {
+			file = file + "/namespaces/" + item.namespace
+		} else {
+			file = file + "/cluster"
+		}
+		file = file + "/" + item.name + ".json"
+		expectedFiles = append(expectedFiles, file)
+	}
+
+	assertTarballContents(t, backupFile, append(expectedFiles, "metadata/version")...)
+}
+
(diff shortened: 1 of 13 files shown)


pkg/backup/item_backupper.go
```diff
@@ -149,19 +146,18 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
...
 	}
 
-	if _, exists := ib.backedUpItems[key]; exists {
+	if _, exists := ib.backupRequest.BackedUpItems[key]; exists {
 		log.Info("Skipping item because it's already been backed up.")
 		return nil
 	}
-	ib.backedUpItems[key] = struct{}{}
-
-	log.Debug(obj.GetObjectKind().GroupVersionKind().GroupVersion().String() + "/" + obj.GetObjectKind().GroupVersionKind().Kind)
```
prydonius: oops - I managed to accidentally merge this debug line in #1682

pkg/backup/request_test.go
```diff
@@ -0,0 +1,42 @@
+package backup
```
skriss: need a copyright header

pkg/persistence/object_store.go
```diff
@@ -36,8 +36,8 @@ import (
 )
 
 type BackupInfo struct {
-	Name                                                       string
-	Metadata, Contents, Log, PodVolumeBackups, VolumeSnapshots io.Reader
+	Name                                                                           string
+	Metadata, Contents, Log, PodVolumeBackups, VolumeSnapshots, BackupResourceList io.Reader
```
skriss: probably time to split this into a line per field since it's so long

pkg/backup/backup_test.go
```diff
@@ -53,6 +53,54 @@ import (
 	"github.com/heptio/velero/pkg/volume"
 )
 
+func TestBackedUpItemsMatchesTarballContents(t *testing.T) {
+	// TODO: figure out if this can be replaced with the restmapper
+	// (https://github.com/kubernetes/apimachinery/blob/035e418f1ad9b6da47c4e01906a0cfe32f4ee2e7/pkg/api/meta/restmapper.go)
```
prydonius: Couldn't figure out how to get the k8s client to do this mapping for me, so just hardcoded for now. Looks like the only usable method that converts GVKs to GVRs is `UnsafeGuessKindToResource` (https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/restmapper.go#L123), which I am wary to use (although might be ok for this test, though so is hardcoding). Would appreciate any pointers if anyone knows how this works?

pkg/backup/backup_test.go
```diff
@@ -53,6 +53,54 @@ import (
...
+		),
+	}
+	for _, resource := range apiResources {
+		h.addItems(t, resource)
+	}
+
+	h.backupper.Backup(h.log, req, backupFile, nil, nil)
+
+	var expectedFiles []string
+	for item := range req.BackedUpItems {
```
prydonius: Another way would be to just hardcode the expected BackedUpItems and filepaths and assert that they match?

prydonius:
@skriss added a test, though not 100% sure this is what you were looking for

pkg/backup/backup_test.go
skriss: This works for me - let's just add a comment that explains the approach here (ensure BackedUpItems matches the tarball contents by looking at BackedUpItems after backup to assemble a list of files expected to be in the tarball, then comparing).

prydonius: :+1: added!

skriss:
I'm going to merge this in prep for the beta, we can follow up if there's any further feedback.

PR #1710, opened 2019-07-31
add velero install flags for configuring restic resource requirements

### LINKED ISSUE #94: Spike - Add cpu, memory requests & limits to deployment

prydonius:
closes #94


pkg/cmd/cli/install/install.go
```diff
@@ -112,6 +124,10 @@ func (o *InstallOptions) AsVeleroOptions() (*install.VeleroOptions, error) {
 	if err != nil {
 		return nil, err
 	}
+	resticPodResources, err := parseResourceRequests(o.ResticPodCPURequest, o.ResticPodMemRequest, o.ResticPodCPULimit, o.ResticPodMemLimit)
```
prydonius: should be changed to use helper from kubeutil package when #1677 is merged

pkg/cmd/cli/install/install.go
```diff
@@ -72,6 +76,10 @@ func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.VeleroPodMemRequest, "velero-pod-mem-request", o.VeleroPodMemRequest, "memory request for Velero pod. Optional.")
 	flags.StringVar(&o.VeleroPodCPULimit, "velero-pod-cpu-limit", o.VeleroPodCPULimit, "CPU limit for Velero pod. Optional.")
 	flags.StringVar(&o.VeleroPodMemLimit, "velero-pod-mem-limit", o.VeleroPodMemLimit, "memory limit for Velero pod. Optional.")
+	flags.StringVar(&o.ResticPodCPURequest, "restic-pod-cpu-request", o.ResticPodCPURequest, "CPU request for Restic pod. Optional.")
```
skriss: "restic pods" since it's for the daemonset (same for all 4 flags).

prydonius: @skriss are you just referring to the description here, or should the flag also be plural? I think it might be easier to keep the flag singular as that will be consistent with the velero ones?

skriss: Ah - yeah, I think just updating the description is fine.

prydonius:
@carlisia @skriss updated, PTAL!

Issue #1691, opened 2019-07-23
Change restore code to get restic snapshot info from PodVolumeBackups if they exist


skriss:
Splitting this out from #1169

skriss:
Closed by #1723

PR #1713, opened 2019-07-31
Properly restore PVs with a reclaim policy of Retain and restic backups

### LINKED ISSUE #1151: Improve handling of PersistentVolumes with Retain reclaim policy

skriss:
Fixes #1151 
Pending #1691 being addressed.

Currently, PVs with a reclaim policy of `Retain` and a restic backup don't get restored properly. This code fixes that by enabling them to be dynamically re-provisioned (empty), and then doing a restore into them.

An annotation is used on the PodVolumeBackup that indicates if the PVB is for a PVC. During restore, this annotation is checked for, and if it exists, the PV is dynamically re-provisioned.

Note this PR doesn't actually do anything as-is yet; waiting for #1691 to be implemented to make the PVB's available during restore.

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -874,41 +874,42 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
 	}
 
 	if groupResource == kuberesource.PersistentVolumes {
-		var hasSnapshot bool
-
-		for _, snapshot := range ctx.volumeSnapshots {
-			if snapshot.Spec.PersistentVolumeName == name {
-				hasSnapshot = true
-				break
+		switch {
+		case hasSnapshot(name, ctx.volumeSnapshots):
+			// Check if the PV exists in the cluster before attempting to create
+			// a volume from the snapshot, in order to avoid orphaned volumes (GH #609)
+			shouldRestoreSnapshot, err := ctx.shouldRestore(name, resourceClient)
+			if err != nil {
+				addToResult(&errs, namespace, errors.Wrapf(err, "error waiting on in-cluster persistentvolume %s", name))
+				return warnings, errs
 			}
-		}
 
-		if !hasSnapshot && hasDeleteReclaimPolicy(obj.Object) {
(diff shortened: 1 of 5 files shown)


pkg/restore/restore.go
```diff
@@ -358,6 +358,7 @@ type context struct {
 	pvsToProvision             sets.String
 	pvRestorer                 PVRestorer
 	volumeSnapshots            []*volume.Snapshot
+	podVolumeBackups           []*api.PodVolumeBackup
```
skriss: This field isn't currently being populated with data, but will be when #1691 merges.

skriss:
OK, rebased and manually tested, things look OK. Looks like we're missing coverage on restores + restic in general - I can look at adding some but probably in a separate PR.

PR #1714, opened 2019-07-31
print resource list metadata in velero backup describe --details

### LINKED ISSUE #396: Increase visibility of what's inside a backup

prydonius:
Download and print backup resource list metadata when running `velero backup describe --details <name>`.

e.g.
```
$ velero backup describe --details nginx-backup-2
Name:         nginx-backup-2
Namespace:    velero
Labels:       velero.io/storage-location=default
Annotations:  <none>

Phase:  PartiallyFailed (run `velero backup logs nginx-backup-2` for more information)

Errors:    2
Warnings:  0

Namespaces:
  Included:  *
  Excluded:  <none>

Resources:
  Included:        *
  Excluded:        <none>
  Cluster-scoped:  auto

Label selector:  app=nginx

Storage Location:  default

Snapshot PVs:  auto

TTL:  720h0m0s

Hooks:  <none>

Backup Format Version:  1

Started:    2019-07-31 13:55:29 -0700 PDT
Completed:  2019-07-31 13:55:46 -0700 PDT

Expiration:  2019-08-30 13:55:29 -0700 PDT

Resource List:
  apps/v1/ReplicaSet:
    - nginx-example/nginx-deployment-5b658f5d7f
    - nginx-example/nginx-deployment-75675f5897
  v1/Endpoints:
    - nginx-example/my-nginx
  v1/Namespace:
    - nginx-example
  v1/PersistentVolume:
    - pvc-dd06dd30-b259-11e9-a63b-025000000001
  v1/PersistentVolumeClaim:
    - nginx-example/nginx-logs
  v1/Pod:
    - nginx-example/nginx-deployment-5b658f5d7f-28qpb
  v1/Service:
    - nginx-example/my-nginx

Persistent Volumes: <none included>
```

depends on #1709 

fixes #396 

### DIFF
--- a/pkg/cmd/util/output/backup_describer.go
+++ b/pkg/cmd/util/output/backup_describer.go
@@ -234,7 +234,7 @@ func DescribeBackupStatus(d *Describer, backup *velerov1api.Backup, details bool
 	d.Println()
 
 	if details {
-		printBackupResourceList(d, backup, veleroClient)
+		describeBackupResourceList(d, backup, veleroClient)
 		d.Println()
 	}
 
@@ -258,12 +258,43 @@ func DescribeBackupStatus(d *Describer, backup *velerov1api.Backup, details bool
 
 		d.Printf("Persistent Volumes:\n")
 		for _, snap := range snapshots {
-			printSnapshot(d, snap.Spec.PersistentVolumeName, snap.Status.ProviderSnapshotID, snap.Spec.VolumeType, snap.Spec.VolumeAZ, snap.Spec.VolumeIOPS)
+			describeSnapshot(d, snap.Spec.PersistentVolumeName, snap.Status.ProviderSnapshotID, snap.Spec.VolumeType, snap.Spec.VolumeAZ, snap.Spec.VolumeIOPS)
 		}
 		return
 	}
 
 	d.Printf("Persistent Volumes: <none included>\n")
+}
+
+func describeBackupResourceList(d *Describer, backup *velerov1api.Backup, veleroClient clientset.Interface) {
+	buf := new(bytes.Buffer)
+	if err := downloadrequest.Stream(veleroClient.VeleroV1(), backup.Namespace, backup.Name, velerov1api.DownloadTargetKindBackupResourceList, buf, downloadRequestTimeout); err != nil {
+		d.Printf("Resource List:\t<error getting backup resource list: %v>\n", err)
+		return
+	}
+
+	var resourceList map[string][]string
+	if err := json.NewDecoder(buf).Decode(&resourceList); err != nil {
+		d.Printf("Resource List:\t<error reading backup resource list: %v>\n", err)
+		return
+	}
+
+	d.Println("Resource List:")
+	for gvk, items := range resourceList {
+		d.Printf("\t%s:\n\t\t- %s\n", gvk, strings.Join(items, "\n\t\t- "))
+	}
+}
+
+func describeSnapshot(d *Describer, pvName, snapshotID, volumeType, volumeAZ string, iops *int64) {
+	d.Printf("\t%s:\n", pvName)
+	d.Printf("\t\tSnapshot ID:\t%s\n", snapshotID)
+	d.Printf("\t\tType:\t%s\n", volumeType)
+	d.Printf("\t\tAvailability Zone:\t%s\n", volumeAZ)
+	iopsString := "<N/A>"
+	if iops != nil {
+		iopsString = fmt.Sprintf("%d", *iops)
+	}
+	d.Printf("\t\tIOPS:\t%s\n", iopsString)
 }
 
 // DescribeDeleteBackupRequests describes delete backup requests in human-readable format.
--- a/pkg/cmd/util/output/backup_describer.go
+++ b/pkg/cmd/util/output/backup_describer.go
@@ -269,7 +269,11 @@ func DescribeBackupStatus(d *Describer, backup *velerov1api.Backup, details bool
 func describeBackupResourceList(d *Describer, backup *velerov1api.Backup, veleroClient clientset.Interface) {
 	buf := new(bytes.Buffer)
 	if err := downloadrequest.Stream(veleroClient.VeleroV1(), backup.Namespace, backup.Name, velerov1api.DownloadTargetKindBackupResourceList, buf, downloadRequestTimeout); err != nil {
-		d.Printf("Resource List:\t<error getting backup resource list: %v>\n", err)
+		if err == downloadrequest.ErrNotFound {
+			d.Println("Resource List:\t<backup resource list not found, this could be because this backup was taken prior to Velero 1.1.0>")
+		} else {
+			d.Printf("Resource List:\t<error getting backup resource list: %v>\n", err)
+		}
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/persistence/object_store.go
```diff
@@ -433,6 +433,8 @@ func (s *objectBackupStore) GetDownloadURL(target velerov1api.DownloadTarget) (s
 		return s.objectStore.CreateSignedURL(s.bucket, s.layout.getBackupLogKey(target.Name), DownloadURLTTL)
 	case velerov1api.DownloadTargetKindBackupVolumeSnapshots:
 		return s.objectStore.CreateSignedURL(s.bucket, s.layout.getBackupVolumeSnapshotsKey(target.Name), DownloadURLTTL)
+	case velerov1api.DownloadTargetKindBackupResourceList:
```
carlisia: Would it be worthwhile adding a test for this?

pkg/cmd/util/output/backup_describer.go
```diff
@@ -261,6 +266,25 @@ func DescribeBackupStatus(d *Describer, backup *velerov1api.Backup, details bool
 	d.Printf("Persistent Volumes: <none included>\n")
 }
 
+func printBackupResourceList(d *Describer, backup *velerov1api.Backup, veleroClient clientset.Interface) {
```
carlisia: Now that I'm looking at this, I realize all the "print" functions are in the `backup_printer.go` file. Do you mind moving this as well as `printSnapshot` in there?

pkg/persistence/object_store.go
prydonius: sorry yes - I missed adding a test for this - will do!

prydonius:
I updated the TestGetDownloadURL test to test multiple keys for different scenarios (e.g. scheduled, prefix, multiple dashes in the name). It's a little bit more verbose, but might be easier to add additional kinds in the future. Let me know what you think!

pkg/cmd/util/output/backup_describer.go
skriss: I actually think these should stay in `backup_describer.go`, perhaps just rename to `describeBackupResourceList` and `describeSnapshot` for clarity.  The code that's in `backup_printer.go` is all for `velero backup get`, while `backup_describer.go` has the code for `velero backup describe` (which this is part of).

skriss:
The test LGTM, though looks like the cases for restore logs & results got dropped

pkg/cmd/util/output/backup_describer.go
prydonius: Ah yes, I also agree that makes sense, looks like `_printer` always relates to a `get` command and `_describer` relates to `describe`. Slightly confusing naming, but I think renaming these functions to `describeBackupResourceList` as you suggested will help.

prydonius:
@skriss the restore logs/results are still there, just moved under the expectedKeyByKind for each test

skriss:
Ah, I see. It's slightly confusing because they're in tests that are titled "backup", "backup with prefix", etc.  It might make sense to separate out the restore-specific targets into their own test cases.

prydonius:
Good point, and also it doesn't really make sense in the context of the scheduled backup. I'll split those out, thanks!

pkg/cmd/util/output/backup_describer.go
```diff
@@ -253,15 +258,34 @@ func DescribeBackupStatus(d *Describer, backup *velerov1api.Backup, details bool
...
 	}
 
 	d.Printf("Persistent Volumes: <none included>\n")
 }
 
-func printSnapshot(d *Describer, pvName, snapshotID, volumeType, volumeAZ string, iops *int64) {
+func describeBackupResourceList(d *Describer, backup *velerov1api.Backup, veleroClient clientset.Interface) {
+	buf := new(bytes.Buffer)
+	if err := downloadrequest.Stream(veleroClient.VeleroV1(), backup.Namespace, backup.Name, velerov1api.DownloadTargetKindBackupResourceList, buf, downloadRequestTimeout); err != nil {
+		d.Printf("Resource List:\t<error getting backup resource list: %v>\n", err)
```
skriss: What does this error look like if you run `velero backup describe --details` for a backup that doesn't have the resource list (e.g. from v1.0)? Wondering if there's a way to discern when the list doesn't exist and show a more friendly error.

prydonius: Unfortunately Stream doesn't return the status code and the error message will be different depending on the storage provider used. E.g. for Azure:

```
Resource List:  <error getting backup resource list: request failed: <?xml version="1.0" encoding="utf-8"?><Error><Code>BlobNotFound</Code><Message>The specified blob does not exist.
RequestId:4628a164-001e-0136-645d-496ab8000000
Time:2019-08-02T18:11:07.1078302Z</Message></Error>>
```

I agree it's not the best UX for older backups, how about modifying `Stream` to return a more user-friendly error when the status code is 404, rather than showing the raw error?

skriss: yeah, i think that makes sense

skriss: I think we're just waiting on addressing this comment before merge.

pkg/cmd/util/output/backup_describer.go
```diff
@@ -253,15 +258,38 @@ func DescribeBackupStatus(d *Describer, backup *velerov1api.Backup, details bool
...
 
 	d.Printf("Persistent Volumes: <none included>\n")
 }
 
-func printSnapshot(d *Describer, pvName, snapshotID, volumeType, volumeAZ string, iops *int64) {
+func describeBackupResourceList(d *Describer, backup *velerov1api.Backup, veleroClient clientset.Interface) {
+	buf := new(bytes.Buffer)
+	if err := downloadrequest.Stream(veleroClient.VeleroV1(), backup.Namespace, backup.Name, velerov1api.DownloadTargetKindBackupResourceList, buf, downloadRequestTimeout); err != nil {
+		if err == downloadrequest.ErrNotFound {
+			d.Println("Resource List:\t<backup resource list not found, this could be because this backup was taken prior to Velero 1.1.0>")
```
prydonius: @skriss updated to show this friendlier error message

skriss:
Merging in prep for beta, can address any further feedback in followups.

PR #1069, opened 2018-11-19
Update to go 1.11

gliptak:

PR #1723, opened 2019-08-06
Restore from PodVolumeBackups

carlisia:
Closes #1069 

### DIFF
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -43,6 +43,7 @@ import (
 	"github.com/heptio/velero/pkg/metrics"
 	"github.com/heptio/velero/pkg/persistence"
 	"github.com/heptio/velero/pkg/plugin/clientmgmt"
+	"github.com/heptio/velero/pkg/restic"
 	pkgrestore "github.com/heptio/velero/pkg/restore"
 	"github.com/heptio/velero/pkg/util/collections"
 	kubeutil "github.com/heptio/velero/pkg/util/kube"
@@ -75,7 +76,7 @@ type restoreController struct {
 
 	namespace              string
 	restoreClient          velerov1client.RestoresGetter
-	backupClient           velerov1client.BackupsGetter
+	podVolumeBackupClient  velerov1client.PodVolumeBackupsGetter
 	restorer               pkgrestore.Restorer
 	backupLister           listers.BackupLister
 	restoreLister          listers.RestoreLister
@@ -94,7 +95,7 @@ func NewRestoreController(
 	namespace string,
 	restoreInformer informers.RestoreInformer,
 	restoreClient velerov1client.RestoresGetter,
-	backupClient velerov1client.BackupsGetter,
+	podVolumeBackupClient velerov1client.PodVolumeBackupsGetter,
 	restorer pkgrestore.Restorer,
 	backupInformer informers.BackupInformer,
 	backupLocationInformer informers.BackupStorageLocationInformer,
@@ -110,7 +111,7 @@ func NewRestoreController(
 		genericController:      newGenericController("restore", logger),
 		namespace:              namespace,
 		restoreClient:          restoreClient,
-		backupClient:           backupClient,
+		podVolumeBackupClient:  podVolumeBackupClient,
 		restorer:               restorer,
 		backupLister:           backupInformer.Lister(),
 		restoreLister:          restoreInformer.Lister(),
@@ -435,13 +436,32 @@ func (c *restoreController) runValidatedRestore(restore *api.Restore, info backu
 	}
 	defer closeAndRemoveFile(backupFile, c.logger)
 
+	opts := restic.NewPodVolumeBackupListOptions(restore.Spec.BackupName)
+	podVolumeBackupList, err := c.podVolumeBackupClient.PodVolumeBackups(c.namespace).List(opts)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
 	volumeSnapshots, err := info.backupStore.GetBackupVolumeSnapshots(restore.Spec.BackupName)
 	if err != nil {
 		return errors.Wrap(err, "error fetching volume snapshots metadata")
 	}
 
 	restoreLog.Info("starting restore")
-	restoreWarnings, restoreErrors := c.restorer.Restore(restoreLog, restore, info.backup, volumeSnapshots, backupFile, actions, c.snapshotLocationLister, pluginManager)
+
+	var podVolumeBackups []*velerov1api.PodVolumeBackup
+	for i, pvb := range podVolumeBackupList.Items {
+		podVolumeBackups[i] = &pvb
+	}
+	restoreData := pkgrestore.Data{
+		Log:              restoreLog,
+		Restore:          restore,
+		Backup:           info.backup,
+		PodVolumeBackups: podVolumeBackups,
+		VolumeSnapshots:  volumeSnapshots,
+		BackupReader:     backupFile,
+	}
+	restoreWarnings, restoreErrors := c.restorer.Restore(restoreData, actions, c.snapshotLocationLister, pluginManager)
 	restoreLog.Info("restore completed")
 
 	if logReader, err := restoreLog.done(c.logger); err != nil {
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -453,7 +453,7 @@ func (c *restoreController) runValidatedRestore(restore *api.Restore, info backu
 	for i := range podVolumeBackupList.Items {
 		podVolumeBackups = append(podVolumeBackups, &podVolumeBackupList.Items[i])
 	}
-	restoreData := pkgrestore.Data{
+	restoreReq := pkgrestore.Request{
 		Log:              restoreLog,
 		Restore:          restore,
 		Backup:           info.backup,
@@ -461,7 +461,7 @@ func (c *restoreController) runValidatedRestore(restore *api.Restore, info backu
 		VolumeSnapshots:  volumeSnapshots,
 		BackupReader:     backupFile,
 	}
-	restoreWarnings, restoreErrors := c.restorer.Restore(restoreData, actions, c.snapshotLocationLister, pluginManager)
+	restoreWarnings, restoreErrors := c.restorer.Restore(restoreReq, actions, c.snapshotLocationLister, pluginManager)
 	restoreLog.Info("restore completed")
 
 	if logReader, err := restoreLog.done(c.logger); err != nil {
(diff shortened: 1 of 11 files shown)


carlisia:
I can use help with adding tests to both test files in this PR. In particular though, this line is failing for two tests and I don't know how to fix it: https://github.com/heptio/velero/pull/1723/files#diff-23ca448ecec41f6d5ef89325e8a41eafR610

pkg/controller/restore_controller.go
```diff
@@ -75,7 +76,7 @@ type restoreController struct {
 
 	namespace              string
 	restoreClient          velerov1client.RestoresGetter
-	backupClient           velerov1client.BackupsGetter
```
carlisia: I removed this because it wasn't being used.

pkg/restic/restorer.go
```diff
@@ -31,10 +31,17 @@ import (
...
+	Restore                         *velerov1api.Restore
+	Pod                             *corev1api.Pod
+	PodVolumeBackups                []*velerov1api.PodVolumeBackup
+	SourceNamespace, BackupLocation string
+}
+
 // Restorer can execute restic restores of volumes in a pod.
 type Restorer interface {
 	// RestorePodVolumes restores all annotated volumes in a pod.
-	RestorePodVolumes(restore *velerov1api.Restore, pod *corev1api.Pod, sourceNamespace, backupLocation string, log logrus.FieldLogger) []error
```
carlisia: the last `log` argument wasn't being used, so I removed it.

pkg/restic/restorer.go
```diff
@@ -136,13 +145,13 @@ ForEachVolume:
...
 	r.resultsLock.Lock()
-	delete(r.results, resultsKey(pod.Namespace, pod.Name))
+	delete(r.results, resultsKey(data.Pod.Namespace, data.Pod.Name))
 	r.resultsLock.Unlock()
 
 	return errs
 }
 
-func newPodVolumeRestore(restore *velerov1api.Restore, pod *corev1api.Pod, volume, snapshot, backupLocation, repoIdentifier string) *velerov1api.PodVolumeRestore {
+func newPodVolumeRestore(restore *velerov1api.Restore, pod *corev1api.Pod, backupLocation, volume, snapshot, repoIdentifier string) *velerov1api.PodVolumeRestore {
```
carlisia: Moved the order here only to make it so all elements of the `RestoreData` struct were the first (3) inputs.

pkg/restic/common.go
```diff
@@ -67,6 +68,20 @@ func GetPodSnapshotAnnotations(obj metav1.Object) map[string]string {
 	return res
 }
 
+// GetVolumesForPod returns a map, of volume name -> snapshot id,
+// of the PodVolumeBackups that exist for the provided pod.
+func GetVolumesForPod(podVolumeBackups []*velerov1api.PodVolumeBackup, pod *corev1api.Pod) map[string]string {
```
skriss: how about `GetVolumeBackupsForPod`?

pkg/restic/restorer.go
```diff
@@ -84,14 +91,16 @@ func newRestorer(
 	return r
 }
 
-func (r *restorer) RestorePodVolumes(restore *velerov1api.Restore, pod *corev1api.Pod, sourceNamespace, backupLocation string, log logrus.FieldLogger) []error {
-	// get volumes to restore from pod's annotations
-	volumesToRestore := GetPodSnapshotAnnotations(pod)
+func (r *restorer) RestorePodVolumes(data RestoreData) []error {
+	volumesToRestore := GetVolumesForPod(data.PodVolumeBackups, data.Pod)
 	if len(volumesToRestore) == 0 {
```
skriss: let's add a comment explaining the backwards-compat logic here, plus a `// TODO(2.0)` to remove the annotations code.

pkg/restore/restic_restore_action.go
```diff
@@ -71,10 +75,23 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
...
+
+	var podVolumeBackups []*velerov1api.PodVolumeBackup
+	for i, pvb := range podVolumeBackupList.Items {
+		podVolumeBackups[i] = &pvb
+	}
+	volumeSnapshots := restic.GetVolumesForPod(podVolumeBackups, &pod)
 	if len(volumeSnapshots) == 0 {
-		log.Debug("No restic snapshot ID annotations found")
-		return velero.NewRestoreItemActionExecuteOutput(input.Item), nil
+		volumeSnapshots = restic.GetPodSnapshotAnnotations(&pod)
```
skriss: it looks like you could change the `restic.GetVolumesForPod` helper to first check the PVBs, then check the annotations if there are no associated PVBs - that way you don't have to repeat this code in a couple different places. I think that makes more sense.

skriss:
> I can use help with adding tests to both test files in this PR. In particular though, this line is failing for two tests and I don't know how to fix it: https://github.com/heptio/velero/pull/1723/files#diff-23ca448ecec41f6d5ef89325e8a41eafR610

In `restore_controller.go`, you added a new use of the velero client. This test is using a list of all actions executed on the fake velero client for verification. Previously, L610 assumed that the "Patch" call it was verifying was the second action executed on the client, but it should now be the third action (since your new use of the velero client became the second). So, TLDR, you just need to bump the slice index from 1 to 2.

skriss:
Looks like you need a `make update` as well.

pkg/controller/restore_controller.go
```diff
@@ -435,13 +436,32 @@ func (c *restoreController) runValidatedRestore(restore *api.Restore, info backu
...
 	if err != nil {
 		return errors.Wrap(err, "error fetching volume snapshots metadata")
 	}
 
 	restoreLog.Info("starting restore")
-	restoreWarnings, restoreErrors := c.restorer.Restore(restoreLog, restore, info.backup, volumeSnapshots, backupFile, actions, c.snapshotLocationLister, pluginManager)
+
+	var podVolumeBackups []*velerov1api.PodVolumeBackup
+	for i, pvb := range podVolumeBackupList.Items {
+		podVolumeBackups[i] = &pvb
```
skriss: `podVolumeBackups = append(podVolumeBackups, &podVolumeBackupList.Items[i])`

pkg/restore/restic_restore_action.go
```diff
@@ -71,10 +75,23 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
...
-	volumeSnapshots := restic.GetPodSnapshotAnnotations(&pod)
+	opts := restic.NewPodVolumeBackupListOptions(input.Restore.Spec.BackupName)
+	podVolumeBackupList, err := a.podVolumeBackupClient.List(opts)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	var podVolumeBackups []*velerov1api.PodVolumeBackup
+	for i, pvb := range podVolumeBackupList.Items {
+		podVolumeBackups[i] = &pvb
```
skriss: `podVolumeBackups = append(podVolumeBackups, &podVolumeBackupList.Items[i])`

pkg/restic/common.go
```diff
@@ -67,6 +68,20 @@ func GetPodSnapshotAnnotations(obj metav1.Object) map[string]string {
...
+	var volumes map[string]string
```
skriss: volumes := make(map[string]string)

pkg/restore/restore.go
```diff
@@ -63,14 +63,19 @@ type VolumeSnapshotterGetter interface {
 	GetVolumeSnapshotter(name string) (velero.VolumeSnapshotter, error)
 }
 
+type Data struct {
```
prydonius: wonder if this struct should be called `Request` and embed the `velerov1api.Restore` object, to follow the same model as the `pkgbackup.Request` struct?

pkg/restore/restic_restore_action.go
```diff
@@ -71,10 +75,23 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
...
+	}
+	volumeSnapshots := restic.GetVolumesForPod(podVolumeBackups, &pod)
 	if len(volumeSnapshots) == 0 {
-		log.Debug("No restic snapshot ID annotations found")
-		return velero.NewRestoreItemActionExecuteOutput(input.Item), nil
+		volumeSnapshots = restic.GetPodSnapshotAnnotations(&pod)
+		if len(volumeSnapshots) == 0 {
+			log.Debug("No restic snapshot ID annotations found")
+			return velero.NewRestoreItemActionExecuteOutput(input.Item), nil
```
nrb: Do we still want to return early here? If there are PVBs created with the new code, len(volumeSnapshots) would always be 0, even if there are PVB objects returned by the client.

skriss: we only hit this code if there are no PVBs per L89. I think this will be clearer if the "first check PVBs, then check annotations if no PVBs" logic moves into the helper func.

pkg/restic/restorer.go
```diff
@@ -84,14 +91,15 @@ func newRestorer(
 	return r
 }
 
-func (r *restorer) RestorePodVolumes(restore *velerov1api.Restore, pod *corev1api.Pod, sourceNamespace, backupLocation string, log logrus.FieldLogger) []error {
-	// get volumes to restore from pod's annotations
-	volumesToRestore := GetPodSnapshotAnnotations(pod)
+// TODO(2.0)
```
skriss: can drop this TODO comment since it's commented at the helper.

pkg/restore/restic_restore_action.go
```diff
@@ -71,7 +75,17 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
...
+		return nil, errors.WithStack(err)
+	}
+
+	var podVolumeBackups []*velerov1api.PodVolumeBackup
+	for i := range podVolumeBackupList.Items {
+		podVolumeBackups = append(podVolumeBackups, &podVolumeBackupList.Items[i])
+	}
+	volumeSnapshots := restic.GetVolumeBackupsForPod(podVolumeBackups, &pod)
 	if len(volumeSnapshots) == 0 {
 		log.Debug("No restic snapshot ID annotations found")
```
skriss: let's update this log to say something like "no restic backups found for pod", and similarly for the one a few lines down ("restic backups found for pod").

pkg/restic/restorer.go
carlisia: That was not meant to be there!

carlisia:
I should have addressed all code reviews, PTAL.

carlisia:
Will add a changelog..

skriss:
I ran a quick manual test of this for restoring new backups and it worked fine. Didn't test with legacy backups, though.

PR #1729, opened 2019-08-06
use Backup CR labels as tags for snapshots

### LINKED ISSUE #1132: Add custom tags to snapshots
Currently, two tags are added to the snapshots of the cloud provider: "ark.heptio.com/backup" and "ark.heptio.com/pv". This seems to happen here:
https://github.com/heptio/ark/blob/59ca9a39742a8c527b82194b678c95a782cf9ce8/pkg/backup/item_backupper.go#L448-L464

It would be great to add custom tags to backups and schedules. One way would be to add a flag like --additional-volume-tags myTag=myValue. Another option could to add the labels that are already added to the backups. 

I think for schedules the tag "ark.heptio.com/schedule" could be a default tag. 

**Environment:**
- Ark version (use `ark version`): 0.10.0
- Kubernetes version (use `kubectl version`): 1.11.5
- Cloud provider or hardware configuration: Azure

prydonius:
This allows users to define custom tags to be added to snapshots, by
specifying custom labels on the Backup CR with the `velero backup create
--labels` flag.

closes #1132

### DIFF
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -436,6 +436,9 @@ func (ib *defaultItemBackupper) takePVSnapshot(obj runtime.Unstructured, log log
 	log = log.WithField("volumeID", volumeID)
 
 	tags := ib.backupRequest.GetLabels()
+	if tags == nil {
+		tags = map[string]string{}
+	}
 	tags["velero.io/backup"] = ib.backupRequest.Name
 	tags["velero.io/pv"] = pv.Name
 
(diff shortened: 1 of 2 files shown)


skriss:
LGTM - let's include a changelog for this one.

pkg/backup/item_backupper.go
```diff
@@ -435,10 +435,9 @@ func (ib *defaultItemBackupper) takePVSnapshot(obj runtime.Unstructured, log log
 
 	log = log.WithField("volumeID", volumeID)
 
-	tags := map[string]string{
-		"velero.io/backup": ib.backupRequest.Name,
-		"velero.io/pv":     pv.Name,
-	}
+	tags := ib.backupRequest.GetLabels()
+	tags["velero.io/backup"] = ib.backupRequest.Name
```
skriss: looks like we need to make sure `tags` is not nil after getting the backup's labels before inserting

PR #1734, opened 2019-08-06
create backups from schedules using velero create backup

### LINKED ISSUE #656: RFE: Create backup from schedule
Hi,

Similar to Kubernetes jobs `kubectl create job --from-cronjob` it would be nice to be able to manually trigger ark backups based on existing schedule definitions.

This could be useful as a quick shortcut for running backups before upgrades, maintenance etc.

e.g. `ark backup create db-maintenance --from-schedule database`

Cheers,
Mike

prydonius:
Adds a flag `--from-schedule` to the velero create backup command to create a Backup from an existing Schedule.

closes #656

### DIFF
--- a/pkg/builder/object_meta.go
+++ b/pkg/builder/object_meta.go
@@ -42,6 +42,23 @@ func WithLabels(vals ...string) func(obj metav1.Object) {
 	}
 }
 
+// WithLabelsMap is a functional option that applies the specified labels map to
+// an object.
+func WithLabelsMap(labels map[string]string) func(obj metav1.Object) {
+	return func(obj metav1.Object) {
+		objLabels := obj.GetLabels()
+		if objLabels == nil {
+			objLabels = make(map[string]string)
+		}
+
+		for k, v := range labels {
+			objLabels[k] = v
+		}
+
+		obj.SetLabels(objLabels)
+	}
+}
+
 // WithAnnotations is a functional option that applies the specified
 // annotation keys/values to an object.
 func WithAnnotations(vals ...string) func(obj metav1.Object) {
--- a/pkg/builder/object_meta.go
+++ b/pkg/builder/object_meta.go
@@ -51,6 +51,7 @@ func WithLabelsMap(labels map[string]string) func(obj metav1.Object) {
 			objLabels = make(map[string]string)
 		}
 
+		// If the label already exists in the object, it will be overwritten
 		for k, v := range labels {
 			objLabels[k] = v
 		}
@@ -83,6 +84,7 @@ func setMapEntries(m map[string]string, vals ...string) map[string]string {
 		key := vals[i]
 		val := vals[i+1]
 
+		// If the label already exists in the object, it will be overwritten
 		m[key] = val
 	}
 
--- a/pkg/cmd/cli/backup/create.go
+++ b/pkg/cmd/cli/backup/create.go
@@ -17,8 +17,11 @@ limitations under the License.
 package backup
 
 import (
+	"errors"
 	"fmt"
 	"time"
+
+	"github.com/heptio/velero/pkg/builder"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -33,6 +36,8 @@ import (
 	veleroclient "github.com/heptio/velero/pkg/generated/clientset/versioned"
 	v1 "github.com/heptio/velero/pkg/generated/informers/externalversions/velero/v1"
 )
+
+const DefaultBackupTTL time.Duration = 30 * 24 * time.Hour
 
 func NewCreateCommand(f client.Factory, use string) *cobra.Command {
 	o := NewCreateOptions()
@@ -84,13 +89,14 @@ type CreateOptions struct {
 	Wait                    bool
 	StorageLocation         string
 	SnapshotLocations       []string
+	FromSchedule            string
 
 	client veleroclient.Interface
 }
 
 func NewCreateOptions() *CreateOptions {
 	return &CreateOptions{
-		TTL:                     30 * 24 * time.Hour,
+		TTL:                     DefaultBackupTTL,
 		IncludeNamespaces:       flag.NewStringArray("*"),
 		Labels:                  flag.NewMap(),
 		SnapshotVolumes:         flag.NewOptionalBool(nil),
@@ -108,6 +114,7 @@ func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.StorageLocation, "storage-location", "", "location in which to store the backup")
 	flags.StringSliceVar(&o.SnapshotLocations, "volume-snapshot-locations", o.SnapshotLocations, "list of locations (at most one per provider) where volume snapshots should be stored")
 	flags.VarP(&o.Selector, "selector", "l", "only back up resources matching this label selector")
+	flags.StringVar(&o.FromSchedule, "from-schedule", "", "create a backup from a Schedule. Cannot be used with any other filters.")
 	f := flags.VarPF(&o.SnapshotVolumes, "snapshot-volumes", "", "take snapshots of PersistentVolumes as part of the backup")
 	// this allows the user to just specify "--snapshot-volumes" as shorthand for "--snapshot-volumes=true"
 	// like a normal bool flag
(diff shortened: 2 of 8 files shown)


pkg/cmd/cli/backup/create.go
```diff
@@ -128,6 +135,17 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 		return err
 	}
 
+	if o.FromSchedule != "" {
+		if (len(o.IncludeNamespaces) > 1 || o.IncludeNamespaces[0] != "*") ||
```
prydonius: This is going to be painful to maintain, is there a better way to check for this? Alternatively:

1. We ignore these flags and mention other filters will get ignored in the description for the `--from-schedule` flag
2. We allow users to override the Schedule with these flags

pkg/cmd/cli/backup/create.go
```diff
@@ -154,13 +172,18 @@ func (o *CreateOptions) Complete(args []string, f client.Factory) error {
...
-		ObjectMeta: metav1.ObjectMeta{
-			Namespace: f.Namespace(),
-			Name:      o.Name,
-			Labels:    o.Labels.Data(),
-		},
-		Spec: api.BackupSpec{
+	var backup *api.Backup
+	backupBuilder := builder.ForBackup(f.Namespace(), o.Name).ObjectMeta(builder.WithLabelsMap(o.Labels.Data()))
+
+	if o.FromSchedule != "" {
```
prydonius: thinking of extracting this to a separate function so we can add a test for this, wdyt?

pkg/builder/object_meta.go
```diff
@@ -42,6 +42,23 @@ func WithLabels(vals ...string) func(obj metav1.Object) {
...
+// an object.
+func WithLabelsMap(labels map[string]string) func(obj metav1.Object) {
+	return func(obj metav1.Object) {
+		objLabels := obj.GetLabels()
+		if objLabels == nil {
+			objLabels = make(map[string]string)
+		}
+
+		for k, v := range labels {
+			objLabels[k] = v
```
nrb: So if the key already exists, this overwrites it? That would be worth documenting on the comment.

pkg/cmd/cli/backup/create.go
```diff
@@ -154,13 +172,18 @@ func (o *CreateOptions) Complete(args []string, f client.Factory) error {
...
+		schedule, err := o.client.VeleroV1().Schedules(f.Namespace()).Get(o.FromSchedule, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		backup = backupBuilder.FromSchedule(schedule).Result()
```
nrb: If I'm reading this right, any labels duplicated between the ones provided by the user and on the schedule will be set to the values on the schedule, correct?

If this is what we want, ti should be clearly documented on the `--from-schedule` argument. However, I think in general values provided at the command line usually win over things like config files or the schedule's values.

pkg/builder/object_meta.go
prydonius: yes - will add a comment!

pkg/cmd/cli/backup/create.go
prydonius: Ah yes, good catch - I agree that the labels provided on the command line should take precedence here. Will change the ordering here.

prydonius:
FYI I think this should wait for v1.2, but just wanted to get something going.

pkg/controller/schedule_controller_test.go
```diff
@@ -97,23 +97,23 @@ func TestProcessSchedule(t *testing.T) {
 			fakeClockTime:        "2017-01-01 12:00:00",
 			expectedErr:          false,
 			expectedPhase:        string(velerov1api.SchedulePhaseEnabled),
-			expectedBackupCreate: builder.ForBackup("ns", "name-20170101120000").ObjectMeta(builder.WithLabels(velerov1api.ScheduleNameLabel, "name")).NoTypeMeta().Result(),
```
prydonius: What's the main reason for the Backup to not have any TypeMeta, is it not needed because the client fills it in automatically when creating the backup?

nrb: I believe that's that case, yes.

pkg/cmd/cli/backup/create.go
skriss: yeah, this is not ideal. I'm leaning towards #1 for simplicity - WDYT?

pkg/cmd/cli/schedule/create.go
```diff
@@ -93,6 +93,9 @@ func NewCreateOptions() *CreateOptions {
 
 func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 	o.BackupOptions.BindFlags(flags)
+	// hide the Backup options --from-schedule flag
+	// TODO: see if there is a way to remove the flag
+	flags.MarkHidden("from-schedule")
```
skriss: One option would be to bind the `--from-schedule` flag in a different method than `BackupOptions.BindFlags(...)`, and to *only* call that new method in the backup create code. Take a look at `BindWait(...)` in the backup create code.

prydonius: Ah thanks for the pointer, that sounds like a much better approach!

pkg/cmd/cli/backup/create.go
prydonius: +1 agreed, I think just logging here that other filters are ignored is probably the best approach here

pkg/cmd/cli/backup/create.go
```diff
@@ -123,6 +128,12 @@ func (o *CreateOptions) BindWait(flags *pflag.FlagSet) {
 	flags.BoolVarP(&o.Wait, "wait", "w", o.Wait, "wait for the operation to complete")
 }
 
+// BindFromSchedule binds the from-schedule flag separately so it is not called
+// by other create commands that reuse CreateOptions's BindFlags method.
+func (o *CreateOptions) BindFromSchedule(flags *pflag.FlagSet) {
+	flags.StringVar(&o.FromSchedule, "from-schedule", "", "create a backup from a Schedule. Cannot be used with any other filters.")
```
skriss: trying to think how to make the help text a little clearer - maybe something like "create a backup by copying the spec from an existing schedule"

prydonius: how about "create a backup from the template of an existing schedule"?

prydonius:
I updated the flag description and added a basic test for building backups, and some tests for building backups from schedules. I couldn't see a clear way to create a table-driven test here, feedback welcome!

pkg/cmd/cli/backup/create.go
```diff
@@ -26,6 +26,8 @@ import (
 	"k8s.io/client-go/tools/cache"
 
 	api "github.com/heptio/velero/pkg/apis/velero/v1"
+	velerov1api "github.com/heptio/velero/pkg/apis/velero/v1"
```
skriss: just noticed we've got a duplicate import here with the previous line - let's convert all `api` uses to `velerov1api`.

pkg/cmd/cli/backup/create_test.go
```diff
@@ -0,0 +1,89 @@
+/*
+Copyright 2017 the Velero contributors.
```
skriss: 2019

skriss:
two nits, otherwise LGTM. tests LGTM as-is -- IMHO when there are only a small number of cases, there's not that much benefit to using table-driven. Gets more useful when lots of scenarios.

skriss:
I think you need a `make update` as well

pkg/cmd/cli/backup/create.go
prydonius: oops, good spot

pkg/cmd/cli/backup/create_test.go
prydonius: Sorry, the generator just takes the header from the main file, so I forgot to update it!

pkg/controller/schedule_controller.go
```diff
@@ -33,7 +33,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 
 	api "github.com/heptio/velero/pkg/apis/velero/v1"
```
prydonius: @skriss looks like this used to have a duplicate package import too, should I update this to use velerov1api everywhere too?

skriss: Eh, let's get this one merged - we can update this next time we touch this file.

pkg/cmd/cli/backup/create_test.go
skriss: Yeah, I would like to automate this away someday.

PR #1739, opened 2019-08-08
bug fix: upload pod volume backups from *all* pods to obj storage

skriss:

Current code is effectively only saving PVB's from the *last* pod since it's not appending.


skriss:
Manually testing to confirm fix.

skriss:
Yep, was able to repro bug with `beta.1` and confirm that this fixes it.

PR #1745, opened 2019-08-08
velero install: support velero/restic pod cpu/mem requests with unbounded limits

### LINKED ISSUE #1711: velero install: not possible to set requests with unbounded limits
The current way to set requests/limits to be unbounded in `velero install` is to set each flag to "0", e.g.:

```
velero install --bucket gcp-backups --provider gcp --secret-file ./gcp-creds.json --velero-pod-cpu-request=0 --velero-pod-cpu-limit=0
```

However, it's not possible to set a request quantity whilst keeping the limit unbounded, as both Velero and the API server complain that the request is greater than the limit (0).

```
velero install --bucket gcp-backups --provider gcp --secret-file ./gcp-creds.json --velero-pod-cpu-request=200m --velero-pod-cpu-limit=0
```

The API server does accept an unbounded limit with a non-zero request if the limit is omitted in the resources object, e.g.:

```
resources:
  limits:
    memory: "0"
  requests:
    cpu: 200m
    memory: "0"
```

is valid, but the following is not

```
resources:
  limits:
    cpu: "0"
    memory: "0"
  requests:
    cpu: 200m
    memory: "0"
```

Perhaps we should detect the zero value and leave the quantity unset? It's also not documented anywhere in the Kubernetes documentation (as far as I can tell) that a zero value means unbounded, though that does appear to be the behaviour.

prydonius:
fixes #1711


nrb:
Could you also document the behavior in https://github.com/heptio/velero/blob/master/site/docs/master/restic.md? Seems relevant to the ConfigMap, too.

PR #1747, opened 2019-08-09
Unit tests or restic restore

### LINKED ISSUE #1730: Add unit tests for PoVolumeBackup restores
Add test coverage for this code change: https://github.com/heptio/velero/pull/1723

carlisia:
Closes #1730.

### DIFF
--- a/pkg/restic/common_test.go
+++ b/pkg/restic/common_test.go
@@ -20,6 +20,8 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/heptio/velero/pkg/builder"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1api "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


carlisia:
Note: I could use some pointers how to properly setup the `TestRestoreWithRestic` test so backups restore with PVB.

skriss:
Looks like a `make update` is needed. Probably won't get to review until Monday.

pkg/restic/common_test.go
```diff
@@ -33,49 +35,84 @@ import (
...
-			annotations: map[string]string{"x": "y", podAnnotationPrefix + "foo": "bar", podAnnotationPrefix + "abc": "123"},
-			expected:    map[string]string{"foo": "bar", "abc": "123"},
+			name: "no snapshot annotation, no suffix, but with PVB",
+			podVolumeBackups: []*velerov1api.PodVolumeBackup{
+				pvbtest1,
+				pvbtest2,
+			},
+			podName:  testPodName,
+			expected: map[string]string{"pvbtest1-foo": "bar", "pvbtest2-abc": "123"},
 		},
```
skriss: let's add some cases for:

- has both PVBs and annotations - PVBs should take precedence
- PVB list includes PVBs for different pods - only the PVBs for the current pod should be returned

pkg/restore/restore_test.go
```diff
@@ -2050,10 +2052,7 @@ func TestRestorePersistentVolumes(t *testing.T) {
 				// restored, we'd get an error of "snapshot not found".
 				"provider-1": &volumeSnapshotter{},
 			},
-			podVolumeBackups: []*velerov1api.PodVolumeBackup{
-				builder.ForPodVolumeBackup("velero", "pvb-1").Result(),
-				builder.ForPodVolumeBackup("velero", "pvb-2").Result(),
-			},
+
```
skriss: is this change intentional? the test case name still says "include podvolumebackups", not exactly sure what the intent is.

pkg/restore/restore_test.go
```diff
@@ -2113,6 +2112,100 @@ func TestRestorePersistentVolumes(t *testing.T) {
...
+type fakeResticRestorer struct {
+	podVolumeBackups []*velerov1api.PodVolumeBackup
+}
+
+func (b *fakeResticRestorer) RestorePodVolumes(data restic.RestoreData) []error {
+	return nil
+}
+
+// TestBackupWithRestic runs backups of pods that are annotated for restic backup,
```
skriss: godoc needs to be updated (looks like copied from pkg/backup).

pkg/restore/restore_test.go
```diff
@@ -2113,6 +2112,100 @@ func TestRestorePersistentVolumes(t *testing.T) {
...
+
+	tests := []struct {
+		name         string
+		restore      *velerov1api.Restore
+		backup       *velerov1api.Backup
+		apiResources []*test.APIResource
+		tarball      io.Reader
+		want         []*velerov1api.PodVolumeBackup
+	}{
+		{
```
skriss: You'll want to set up test cases that have:
- PodVolumeBackups associated with the backup that are being passed in via the Request obj
- Pods in the tarball that the PodVolumeBackups are linked to

Then you'll want to verify that `RestorePodVolumes` is called as expected. TBH this could be a place where using a mock would make sense, since we can't really verify much more than that the method was called as expected (# times, args specified), given the nature of how restic works and that we're not actually running in a cluster.

carlisia:
I had missed the comments to this. Will address them today.

skriss:
@carlisia are you making progress on this? please reach out if you need further input!

carlisia:
@skriss PTAL when you have a chance.

pkg/restic/common_test.go
```diff
@@ -28,54 +28,106 @@ import (
...
+	"github.com/heptio/velero/pkg/builder"
 	"github.com/heptio/velero/pkg/generated/clientset/versioned/fake"
 	informers "github.com/heptio/velero/pkg/generated/informers/externalversions"
 	velerotest "github.com/heptio/velero/pkg/util/test"
 )
 
-func TestGetPodSnapshotAnnotations(t *testing.T) {
+func TestGetVolumeBackupsForPod(t *testing.T) {
+	testPodName := "TestPod"
+	pvbtest1 := builder.ForPodVolumeBackup("velero", "pvb-1").Result()
```
skriss: It would be more consistent with other tests to define the PVBs inline in the test cases (chaining together the builder methods like '.PodName(...).Volume(...)`, even if it means some duplication across test cases. I prefer this approach because then all the test data is right in front of you and you don't have to jump around to understand what data's being used.

pkg/restore/restore_test.go
```diff
@@ -2095,7 +2094,7 @@ func TestRestorePersistentVolumes(t *testing.T) {
 				Log:              h.log,
 				Restore:          tc.restore,
 				Backup:           tc.backup,
-				PodVolumeBackups: tc.podVolumeBackups,
+				PodVolumeBackups: nil,
```
skriss: we can just delete this line since it's always nil

pkg/restore/restore_test.go
```diff
@@ -2113,6 +2112,130 @@ func TestRestorePersistentVolumes(t *testing.T) {
 	}
 }
 
+type fakeResticRestorerFactory struct {
+	restorer         *resticmocks.Restorer
+	podVolumeBackups []*velerov1api.PodVolumeBackup
```
skriss: I don't think the `podVolumeBackups` field is used - is it? let's delete if not.

pkg/restore/restore_test.go
```diff
@@ -2113,6 +2112,130 @@ func TestRestorePersistentVolumes(t *testing.T) {
...
+}
+
+func (f *fakeResticRestorerFactory) NewRestorer(ctx.Context, *velerov1api.Restore) (restic.Restorer, error) {
+	return f.restorer, nil
+}
+
+type fakeResticRestorer struct {
```
skriss: I don't think the `fakeResticRestorer` struct is used - is it? let's delete if not.

pkg/restore/restore_test.go
```diff
@@ -2113,6 +2112,130 @@ func TestRestorePersistentVolumes(t *testing.T) {
...
+	podVolumeBackups []*velerov1api.PodVolumeBackup
+}
+
+func (b *fakeResticRestorer) RestorePodVolumes(data restic.RestoreData) []error {
+	return nil
+}
+
+// TestRestoreWithRestic runs restores of only pods that contain associated PVBs.
```
skriss: It would be nice to enhance the godoc comment to explain how the test/verification works (i.e. by using a mock for the restic restorer, and verifying that it's called for the expected pods).

pkg/restore/restore_test.go
```diff
@@ -2113,6 +2112,130 @@ func TestRestorePersistentVolumes(t *testing.T) {
...
+	tests := []struct {
+		name                        string
+		restore                     *velerov1api.Restore
+		backup                      *velerov1api.Backup
+		apiResources                []*test.APIResource
+		podVolumeBackups            []*velerov1api.PodVolumeBackup
+		podWithPVBs, podWithoutPVBs []*corev1api.Pod
+	}{
+		{
+			name:         "a pod that exists in given backup and contains associated PVBs should have restored pods",
```
skriss: maybe instead of "should have restored pods", something like "should have RestorePodVolumes called"?

pkg/restore/restore_test.go
```diff
@@ -2113,6 +2112,130 @@ func TestRestorePersistentVolumes(t *testing.T) {
...
+			}
+
+			warnings, errs := h.restorer.Restore(
+				data,
+				nil, // actions
+				nil, // snapshot location lister
+				nil, // volume snapshotter getter
+			)
+
+			assertEmptyResults(t, warnings, errs)
```
skriss: it might be good to also ensure that *all* of the pods (whether they have PVBs or not) get restored, to ensure we don't have any test bugs. You can use `assertAPIContents` like one of the other tests for that.

pkg/restore/restore_test.go
```diff
@@ -2113,6 +2112,130 @@ func TestRestorePersistentVolumes(t *testing.T) {
...
+			name:         "a pod that exists in given backup and contains associated PVBs should have restored pods",
+			restore:      defaultRestore().Result(),
+			backup:       defaultBackup().Result(),
+			apiResources: []*test.APIResource{test.Pods()},
+			podVolumeBackups: []*velerov1api.PodVolumeBackup{
+				builder.ForPodVolumeBackup("velero", "pvb-1").PodName("pod-1").Result(),
+				builder.ForPodVolumeBackup("velero", "pvb-2").PodName("pod-2").Result(),
+			},
+			podWithPVBs: []*corev1api.Pod{
+				builder.ForPod("ns-1", "pod-2").
```
skriss: let's make sure we have a case with >1 `podWithPVBs` -- could just update this case.

pkg/restore/restore_test.go
```diff
@@ -2113,6 +2112,130 @@ func TestRestorePersistentVolumes(t *testing.T) {
...
+			for _, pod := range tc.podWithPVBs {
+				tarball.addItems("pods", pod)
+
+				// the restore process adds these labels before restoring, so we must add them here too otherwise they won't match
+				pod.Labels = map[string]string{"velero.io/backup-name": tc.backup.Name, "velero.io/restore-name": tc.restore.Name}
+				expectedArgs := restic.RestoreData{
+					Restore:          tc.restore,
+					Pod:              pod,
+					PodVolumeBackups: tc.podVolumeBackups,
+					SourceNamespace:  "ns-1",
```
skriss: I think you want this to be `pod.Namespace`

pkg/restore/restore_test.go
```diff
@@ -2113,6 +2112,130 @@ func TestRestorePersistentVolumes(t *testing.T) {
...
+					On("RestorePodVolumes", expectedArgs).
+					Return(nil)
+			}
+
+			data := Request{
+				Log:              h.log,
+				Restore:          tc.restore,
+				Backup:           tc.backup,
+				PodVolumeBackups: tc.podVolumeBackups,
+				VolumeSnapshots:  nil,
```
skriss: can delete this line

PR #1754, opened 2019-08-12
Cleanup .velero directory when init-container finishes

dymurray:
Signed-off-by: Dylan Murray <dymurray@redhat.com>

Choosing not to error out when cleanup fails since the PVR controller attempts to wipe stale data as well. This is a soft cleanup.

### DIFF
--- a/cmd/velero-restic-restore-helper/main.go
+++ b/cmd/velero-restic-restore-helper/main.go
@@ -77,8 +77,9 @@ func done() bool {
 	return true
 }
 
-// cleanupDoneFiles returns an error if it fails to remove all of the `.velero`
-// dirs created by Velero when a restic restore completes
+// cleanupDoneFiles returns if it fails to remove all of the `.velero` dirs
+// created by Velero when a restic restore completes or when all directories
+// are removed. It does not return an error to run a 'soft' cleanup.
 func cleanupDoneFiles() {
 	children, err := ioutil.ReadDir("/restores")
 	if err != nil {
--- a/cmd/velero-restic-restore-helper/main.go
+++ b/cmd/velero-restic-restore-helper/main.go
@@ -32,6 +32,7 @@ func main() {
 
 	ticker := time.NewTicker(time.Second)
 	defer ticker.Stop()
+	defer cleanupDoneFiles()
 
 	for {
 		select {
@@ -74,4 +75,26 @@ func done() bool {
 	}
 
 	return true
+}
+
+// cleanupDoneFiles returns an error if it fails to remove all of the `.velero`
+// dirs created by Velero when a restic restore completes
+func cleanupDoneFiles() {
+	children, err := ioutil.ReadDir("/restores")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "ERROR reading /restores directory: %s\n", err)
+		return
+	}
+
+	for _, child := range children {
+		if !child.IsDir() {
+			fmt.Printf("%s is not a directory, skipping.\n", child.Name())
+			continue
+		}
+		doneFileDir := filepath.Join("/restores", child.Name(), ".velero")
+		if err := os.RemoveAll(filepath.Join(doneFileDir, ".velero")); err != nil {
+			fmt.Fprintf(os.Stderr, "ERROR removing .velero directory: %s\n", err)
+		}
+	}
+	return
 }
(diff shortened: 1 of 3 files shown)


cmd/velero-restic-restore-helper/main.go
```diff
@@ -75,3 +76,26 @@ func done() bool {
...
+		return
+	}
+
+	for _, child := range children {
+		if !child.IsDir() {
+			fmt.Printf("%s is not a directory, skipping.\n", child.Name())
+			continue
+		}
+		doneFileDir := filepath.Join("/restores", child.Name(), ".velero")
+		if err := os.RemoveAll(filepath.Join(doneFileDir, ".velero")); err != nil {
```
skriss: looks like there's an extra `.velero` in the path here (it's being added in both L95 and L96).

cmd/velero-restic-restore-helper/main.go
```diff
@@ -75,3 +76,26 @@ func done() bool {
...
+			fmt.Fprintf(os.Stderr, "ERROR removing .velero directory: %s\n", err)
```
skriss: let's log the full directory path that errored

cmd/velero-restic-restore-helper/main.go
```diff
@@ -75,3 +76,26 @@ func done() bool {
...
+		}
```
skriss: let's also log on success, including the full directory path that was removed

dymurray:
Done, thanks @skriss

cmd/velero-restic-restore-helper/main.go
```diff
@@ -75,3 +76,27 @@ func done() bool {
...
+	for _, child := range children {
+		if !child.IsDir() {
+			fmt.Printf("%s is not a directory, skipping.\n", child.Name())
+			continue
+		}
+		doneFileDir := filepath.Join("/restores", child.Name(), ".velero")
+		if err := os.RemoveAll(doneFileDir); err != nil {
+			fmt.Fprintf(os.Stderr, "ERROR removing .velero directory [%s]: %s\n", doneFileDir, err)
+		}
+		fmt.Printf("Successfully removed .velero directory: %s\n", doneFileDir)
```
skriss: let's either move this into an `else` block, or add a `continue` in the previous `err != nil` block so we don't hit this line if an error's gotten.

dymurray:
It looks like the underlying volume is mounted by restic so the removal fails. Need to rethink where this logic can go.

skriss:
Hmm, looks like we're writing the file underneath `.velero` with `0644` -- maybe changing this to `0755` would fix? I don't think that would be an issue. https://github.com/heptio/velero/blob/master/pkg/controller/pod_volume_restore_controller.go#L372

dymurray:
@skriss yeah I don't see any issue with changing it to `0755`. I will include that in this PR.

skriss:
This looks okay to me - @sseago are you able to confirm this works as intended?

sseago:
So the change didn't work for me. Since all it did was added execute permission to the file that we can't delete, I'm unclear on how it would have really helped here. The permissions on the .velero directory are more relevant here. We also need to be able to delete the .velero dir as well as the file within it. I'm wondering if context of the initcontainer is running with a different user than either the normal application user or the user that created the files in the first place.

sseago:
So I tried to rsh into each pod and manually delete the file (and then directory). It turned out that doing that worked for 4 of my 5 pods. One of them failed since for my cephfs volume the .velero directory is owned by root:root rather than root:1000460000 (group owner of the user running the pods), but for the others it was able to delete the file just fine. I'm thinking that maybe when we attempt to clean up the .velero dir we're not running as user 1000460000, since none of the pods were deleted, but even if we were running as that user, it would still have failed for the cephfs pod.

sseago:
So if `.velero` is created by restic, which is running as a privileged container (and `runAsUser: 0`), it looks like we have an issue trying to delete this from the pod's initContainer running with a default securityContext. I don't know if this means we need to chown .velero to some specified user that the pod's initContainer should be set to run as, or if the pod's initContainer should run as root, but the two need to match somehow. Alternatively, maybe the `.velero` directory cleanup should happen on the pod_volume_restore_controller side after the initContainers have all exited.

skriss:
@dymurray @sseago is this something you're planning to pick up again, or should we close this PR out?

skriss:
I'm going to close this out for now since it's inactive, but feel free to reopen or replace appropriate, thanks!

PR #1759, opened 2019-08-12
Feature flag design

nrb:
* [ ] Address graduation & removal of feature flags. Do the flags always stay but the conditionals get removed?
* [ ] Add command for listing known features?

### DIFF
--- a/design/feature-flags.md
+++ b/design/feature-flags.md
@@ -1,0 +1,71 @@
+# Feature Flags
+
+Status: Accepted
+
+Some features may take a while to get fully implemented, and we don't necessarily want to have long-lived feature branches
+A simple feature flag implementation allows code to be merged into master, but not used unless a flag is set.
+
+## Goals
+
+- Allow unfinished features to be present in Velero releases, but only enabled when the associated flag is set.
+
+## Non Goals
+
+- A robust feature flag library.
+
+## Background
+
+When considering the [CSI integration work](https://github.com/heptio/velero/pull/1661), the timelines involved presented a problem in balancing a release and longer-running feature work.
+A simple implementation of feature flags can help protect unfinished code while allowing the rest of the changes to ship.
+
+## High-Level Design
+
+A new command line flag, `--features` will be added to the root `velero` command.
+
+`--features` will accept a comma-separated list of features, such as `--features EnableCSI,Replication`.
+Each feature listed will correspond to a key in a map in `pkg/features/features.go` defining whether a feature should be enabled.
+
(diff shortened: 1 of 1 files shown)


skriss:
Approach seems reasonable. We'll want to add a docs page listing the valid features and whether they're enabled/disabled by default.

Just looking at Kubernetes feature gates (https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) - should we make the flag take a series of key-value pairs, e.g. `CSISupport=true,Replication=false,...`?  This would make it easier to turn off a feature that's on by default, for example.

nrb:
That's true. I was more thinking that we only put feature flags for things that were in-progress, and when they graduated those flags would be removed.

That lifecycle of a feature and its associated flag(s) should be added to this.

prydonius:
> I was more thinking that we only put feature flags for things that were in-progress, and when they graduated those flags would be removed.

I would agree, not sure that we will really need to keep these feature flags once we want to make them default. I suppose this is useful in Kubernetes due to some features depending on specific environments and requirements.

skriss:
Alright, that works for me. We can always change later if needed.

design/feature-flags.md
```diff
@@ -0,0 +1,47 @@
...
+A new command line flag, `--features` will be added to the root `velero` command.
+
+`--features` will accept a comma-separated list of features, such as `--features EnableCSI,Replication`.
+Each feature listed will correspond to a key in a map in `pkg/features/features.go` defining whether a feature should be enabled.
+
+Any code implementing the feature would then import the map and look up the key's value.
+
+For the Velero client, a `features` key can be added to the config.json file for more convenient client invocatons.
+
+## Detailed Design
```
prydonius: We might want to drop all the placeholder stuff after this, so it's less confusing to read this proposal

design/feature-flags.md
```diff
@@ -0,0 +1,71 @@
...
+
+For the Velero client, a `features` key can be added to the `config.json` file for more convenient client invocations.
+
+## Detailed Design
+
+A new `features` package will be introduced with these basic structs:
+
+```go
+type FeatureFlagSet struct {
+    flags map[string]bool
```
prydonius: Guess this could also be a `sets.String` type, unless we want to keep the option of being able to disable a flag?

nrb: Oh, yeah, that's a good idea.

skriss:
ready to merge when it moves out of draft and changes to `Accepted`.

nrb:
@skriss updated status and squashed

Issue #1764, opened 2019-08-14
ensure pod volume is empty before restic restore


prydonius:
See conversation on Slack: https://kubernetes.slack.com/archives/C6VCGP4MT/p1565817791421900

Restic will overwrite any files from a backup that already exist in a volume on restore. Whilst this is expected for Restic, it might not make sense in Velero's usecase. To be consistent with cloud-provider snapshot restores, we should check to ensure that a pod volume is empty before attempting to restore into it. This way, we never accidentally overwrite any existing files in a volume and can expect the restored volume to only contain files that were in the backup. Does this sound like an okay expectation?

Additionally if we know the directory is empty, we can calculate the progress of a restore based on the size of the directory and size of the full backup (#1749).

skriss:
I think we decided that this isn't an issue, as PVs will be dynamically provisioned (empty) on a restic restore, and things like `emptyDir` will also be recreated as the pod is restored - is there anything else to follow up on here?

prydonius:
I haven't tried it, but what happens if you try to use a gitRepo volume, or some other type of volume that might pull in additional files before the init-containers? Just a thought, I know the gitRepo volume isn't really used and I don't know of any other volume types that do this. I'm happy closing this out if we don't think it's worth investigating this more.

skriss:
closing this one out as I don't think there's an action item here, but we can reopen as needed.

:wave: adnan :)

PR #1765, opened 2019-08-15
restic backup and restore progress proposal

prydonius:

### DIFF
--- a/design/restic-backup-and-restore-progress.md
+++ b/design/restic-backup-and-restore-progress.md
@@ -27,8 +27,9 @@ This progress will be read on some interval and the PodVolumeBackup Custom Resou
 
 ### restic restore progress
 
-The `restic stats` command returns the total size of a given snapshot.
-This can be compared with the total size of files in the volume periodically to calculate the completion percentage of the restore.
+The `restic ls` command returns the number of files in a backup, including the size of each file.
+This can be compared with the total size of these file in the volume periodically to calculate the completion percentage of the restore.
+The PodVolumeRestore CR's status will be updated with this information.
 
 ## Detailed Design
 
@@ -61,22 +62,38 @@ Once the backup has completed successfully, the PodVolumeBackup will be patched 
 
 ### restic restore progress
 
-<!-- TODO: complete -->
+The `restic ls <snapshot_id> --json` command provides information about the size of backed up files:
 
-- before restore, determine size of snapshot using `restic stats --json`
-- periodically calculate size of files in the volume (see https://github.com/missedone/dugo/blob/master/du.go#L12 for a way to calculate size)
-- compare with total size to get progress: cursize / totalsize
+```
+{"time":"2019-08-14T15:16:29.105557-07:00","tree":"1de5381473aecb5548959f6e5444bb7966fe781500b5f8ec1629088674bea6a6","paths":["/Users/aadnan/Playground/restic/files"],"hostname":"aadnan-a01.vmware.com","username":"aadnan","uid":501,"gid":20,"id":"5628926b4f0e105eb33d83c619f235a8cf8f695d19955fab44bf506ece195f3a","short_id":"5628926b","struct_type":"snapshot"}
+{"name":"files","type":"dir","path":"/files","uid":501,"gid":20,"mode":2147484096,"mtime":"2019-08-14T15:16:20.53091354-07:00","atime":"2019-08-14T15:16:20.53091354-07:00","ctime":"2019-08-14T15:16:20.53091354-07:00","struct_type":"node"}
+{"name":"file1","type":"file","path":"/files/file1","uid":501,"gid":20,"size":1435500544,"mode":384,"mtime":"2019-08-14T15:16:20.491206506-07:00","atime":"2019-08-14T15:16:20.491206506-07:00","ctime":"2019-08-14T15:16:20.491234451-07:00","struct_type":"node"}
+...
+```
+
+Before beginning the restore operation, we can use the output of `restic ls` to get a list of the files that will be restored and calculate the total size of the backup.
+
+The code to run the `restic restore` command will be changed to include a Goroutine that periodically (every 10 seconds) goes through the list of files in the backup and gets the current size of each file in the volume.
+If the file doesn't exist in the volume yet (restic hasn't started restoring it), the size is 0.
+The sum of the current size of each file is then compared with the total size of the backup to calculate a completion percentage (`cursize / totalsize * 100`).
+
+The PodVolumeRestoreStatus type will be extended to include a `Progress` field.
+The PodVolumeRestore will be patched to set the `status.Progress` field to the percentage value of the progress.
+
+Once the restore has completed successfully, the PodVolumeRestore will be patched to set `status.Progress = 100%`.
 
 ## Open Questions
 
-- Can we assume that the volume we are restoring in will be empty? Can it contain other artefacts? If so, we won't be able to use the proposed approach to compare backup and current volume sizes and will need to use the alternative approach described below.
+- Can we assume that the volume we are restoring in will be empty? Can it contain other artefacts? If we can assume it to be empty, we can go with the easier approach detailed below.
 
 ## Alternatives Considered
 
 ### restic restore progress
 
-To deal with the case where a volume we are restoring into may contain additional files not included in the snapshot, we can use the information from `restic ls` to retrieve the list of filepaths we should include when calculating the size of restored files in the volume.
-It's possible that certain volume types may contain hidden files that could attribute to the total size of the volume, and so this might be an overall safer approach.
+If we can assume that the volume we are restoring into will be empty, we can instead use the output from `restic stats` to get the total size of the backup.
+This can then be periodically compared with the size of all files in the volume to calculate a completion percentage.
+This will be simpler as we don't need to keep track of each file in the backup, but will not work if the volume could contain other files not included in the backup.
+It's possible that certain volume types may contain hidden files that could attribute to the total size of the volume, and so this might be an overall less safe approach.
 
 Another option is to contribute progress reporting similar to `restic backup` for `restic restore` upstream.
 This may take more time, but would give us a more native view on the progress of a restore.
--- a/design/restic-backup-and-restore-progress.md
+++ b/design/restic-backup-and-restore-progress.md
@@ -63,8 +63,10 @@ With the `--json` flag, `restic backup` outputs single lines of JSON reporting t
 The [command factory for backup](https://github.com/heptio/velero/blob/af4b9373fc73047f843cd4bc3648603d780c8b74/pkg/restic/command_factory.go#L37) will be updated to include the `--json` flag.
 The code to run the `restic backup` command (https://github.com/heptio/velero/blob/af4b9373fc73047f843cd4bc3648603d780c8b74/pkg/controller/pod_volume_backup_controller.go#L241) will be changed to include a Goroutine that reads from the command's stdout stream.
 The implementation of this will largely follow [@jmontleon's PoC](https://github.com/fusor/velero/pull/4/files) of this.
-The Goroutine will periodically read the stream (every 10 seconds) and get the last printed status line.
-The line will be decoded as JSON and the `bytes_done` and `total_bytes` property will be read and the PodVolumeBackup will be patched to update `status.Progress.BytesDone` and `status.Progress.TotalBytes` respectively.
+The Goroutine will periodically read the stream (every 10 seconds) and get the last printed status line, which will be convered to JSON.
+If `bytes_done` is empty, restic has not finished scanning the volume and hasn't calculated the `total_bytes`.
+In this case, we will not update the PodVolumeBackup and instead will wait for the next iteration.
+Once we get a non-zero value for `bytes_done`, the `bytes_done` and `total_bytes` properties will be read and the PodVolumeBackup will be patched to update `status.Progress.BytesDone` and `status.Progress.TotalBytes` respectively.
 
 Once the backup has completed successfully, the PodVolumeBackup will be patched to set `status.Progress.BytesDone = status.Progress.TotalBytes`.
 This is done since the main thread may cause early termination of the Goroutine once the operation has finished, preventing a final update to the `BytesDone` property.
--- a/design/restic-backup-and-restore-progress.md
+++ b/design/restic-backup-and-restore-progress.md
@@ -1,0 +1,87 @@
+# Progress reporting for restic backups and restores
+
+Status: Draft
+
+During long-running restic backups/restores, there is no visibility into what (if anything) is happening, making it hard to know if the backup/restore is making progress or hung, how long the operation might take, etc.
+We should capture progress during restic operations and make it user-visible so that it's easier to reason about.
+This document proposes an approach for capturing progress of backup and restore operations and exposing this information to users.
+
+## Goals
+
+- Provide basic visibility into restic operations to inform users about their progress.
+
+## Non Goals
+
+- Capturing progress for non-restic backups and restores.
+
+## Background
+
+(Omitted, see introduction)
+
+## High-Level Design
+
+### restic backup progress
+
+The `restic backup` command provides progress reporting to stdout in JSON format, which includes the completion percentage of the backup.
+This progress will be read on some interval and the PodVolumeBackup Custom Resource's (CR) status will be updated with this information.
+
+### restic restore progress
+
+The `restic stats` command returns the total size of a given snapshot.
+This can be compared with the total size of files in the volume periodically to calculate the completion percentage of the restore.
+
+## Detailed Design
+
+### restic backup progress
+
+restic added support for [streaming JSON output for the `restic backup` command](https://github.com/restic/restic/pull/1944) in 0.9.5.
+Our current images ship restic 0.9.4, and so the Dockerfile will be updated to pull the new version: https://github.com/heptio/velero/blob/af4b9373fc73047f843cd4bc3648603d780c8b74/Dockerfile-velero#L21.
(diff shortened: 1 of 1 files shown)


carlisia:
This design seems straight forward. As for the question about if a directory being restore into not being empty, I think we need to decide in general, if we need to make a choice, if we want to be more uniform with restic features, or with velero features, or if it's a case-by-case decision.

prydonius:
Agreed, I think we can discuss that as part of #1764. I am feeling more inclined to go for the slightly alternative approach here where we only compare sizes of files that we know are in the backup, that way we don't need to assume that the volume will always be empty for now.

prydonius:
Okay updated this proposal to use the approach where we only consider the list of files in the backup.

design/restic-backup-and-restore-progress.md
```diff
@@ -0,0 +1,104 @@
...
+{"message_type":"status","percent_done":0.0028089887640449437,"total_files":10,"total_bytes":49461329920,"bytes_done":138936320,"current_files":["/file1","/file3"]}
+```
+
+The [command factory for backup](https://github.com/heptio/velero/blob/af4b9373fc73047f843cd4bc3648603d780c8b74/pkg/restic/command_factory.go#L37) will be updated to include the `--json` flag.
+The code to run the `restic backup` command (https://github.com/heptio/velero/blob/af4b9373fc73047f843cd4bc3648603d780c8b74/pkg/controller/pod_volume_backup_controller.go#L241) will be changed to include a Goroutine that reads from the command's stdout stream.
+The implementation of this will largely follow [@jmontleon's PoC](https://github.com/fusor/velero/pull/4/files) of this.
+The Goroutine will periodically read the stream (every 10 seconds) and get the last printed status line.
+The line will be decoded as JSON and the `percent_done` property will be read and formatted as a percentage value.
+
+The PodVolumeBackupStatus type will be extended to include a `Progress` field.
```
skriss: I would probably go with `TotalBytes` and `BytesComplete` fields (names TBD), and then calculate percent complete in the UI code.

design/restic-backup-and-restore-progress.md
```diff
@@ -0,0 +1,104 @@
...
+...
+```
+
+Before beginning the restore operation, we can use the output of `restic ls` to get a list of the files that will be restored and calculate the total size of the backup.
+
+The code to run the `restic restore` command will be changed to include a Goroutine that periodically (every 10 seconds) goes through the list of files in the backup and gets the current size of each file in the volume.
+If the file doesn't exist in the volume yet (restic hasn't started restoring it), the size is 0.
+The sum of the current size of each file is then compared with the total size of the backup to calculate a completion percentage (`cursize / totalsize * 100`).
+
+The PodVolumeRestoreStatus type will be extended to include a `Progress` field.
```
skriss: similarly, I'd go with `TotalBytes` and `BytesComplete` and calculate percent complete in UI code.

design/restic-backup-and-restore-progress.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Open Questions
+
+- Can we assume that the volume we are restoring in will be empty? Can it contain other artefacts? If we can assume it to be empty, we can go with the easier approach detailed below.
+
+## Alternatives Considered
+
+### restic restore progress
+
+If we can assume that the volume we are restoring into will be empty, we can instead use the output from `restic stats` to get the total size of the backup.
```
skriss: I do like the simplicity of this approach, since we don't have to do any kind of per-file processing (which is more complicated and potentially has performance issues for large volumes).  We can use something like `du -s` to get the total current size of the volume to compare to the `restic stats` total size.

I *believe* we can assume that the volume is empty when a restic restore starts - in the case of PVCs, we dynamically provision an empty volume, and for things like `emptyDir` we'd be getting an empty volume as well. Need to think about this a little more, though.

prydonius: Actually I'm not sure that `du -s` can be used here, when I tested this locally the size reported from `du` is a little different due to it showing actual disk usage. There is `du --apparent-size` which is supposed to give the "apparent" size of the directory, but even with that flag I'm still seeing a disparity.

I did find https://github.com/missedone/dugo/blob/master/du.go which walks the directory and gets the total size of each file, this does end up matching what `restic stats --json` reports. However, if we're having to walk the directory anyway, we might as well only walk the files listed in the backup.

I'll look more into `du` though to see if I can figure out how to get it to match the actual size, because it would certainly be more performant for large volumes.

design/restic-backup-and-restore-progress.md
prydonius: :+1: I think that's useful information to surface. We can fill the TotalBytes field before starting the restore/backup operation.

design/restic-backup-and-restore-progress.md
skriss: Ah, OK. It would be nice not to have to check every file every time - not sure what the performance would be like for volumes with large #'s of files.

prydonius: Okay so the discrepancy between `du --apparent-size --block-size 1` and restic's size comes from `du` counting the size of the directory entry as well as the files.

For example:
```
$ ls -la
total 11277312
drwx------  7 aadnan  staff         224 Aug 19 15:50 .
drwxr-xr-x  6 aadnan  staff         192 Aug 14 14:47 ..
-rw-------  1 aadnan  staff  1435500544 Aug 14 15:16 file1
-rw-------  1 aadnan  staff  1317011456 Aug 14 15:16 file2
-rw-------  1 aadnan  staff  1115684864 Aug 14 15:16 file3
-rw-------  1 aadnan  staff  1340080128 Aug 14 15:16 file4
-rw-------  1 aadnan  staff   523239424 Aug 14 15:16 file5
```

For the above directory, the restic backup size would be `5731516416` which is the sum of the sizes for file1-5. The calculation from `du -b` (equivalent to `du --apparent-size --block-size 1`) produces `5731516640` which is the sum of the sizes for file1-5 plus the `224`, the size of the current directory entry.

The problem is more difficult for subdirectories as the total sum will include the size of each nested sub-directory entry and there doesn't seem to be a way to get `du` to only consider files. The only way I can find to do this is to use `find` to pass the list of files in the directory to `du` and use the last line (https://superuser.com/questions/853667/getting-du-to-ignore-size-of-and-entries):

```
$ du -c -b `find . -type f`
1115684864	./file3
1340080128	./file4
522190848	./test/file2
1326448640	./test/subdir/file2
1810890752	./test/subdir/file1
1167065088	./test/file1
523239424	./file5
1317011456	./file2
1435500544	./file1
10558111744	total
```

This total matches the restic total size for the backup:
```
$ restic stats --json latest
{"total_size":10558111744,"total_file_count":11}
```

I'm not sure how performant this will be compared to walking the tree in Go directly, especially for larger volumes. I also don't know that `du` has any special implementation that allows it to not have to walk through each file to calculate the total size (looks like it's doing just this from the implementation: http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/du.c#n678). Looks likes `du` can take a long time on large volumes too (https://unix.stackexchange.com/questions/8485/how-to-cache-or-otherwise-speed-up-du-summaries - > 2 minutes in this case). However, it does sound like repeated calls to `du` are faster due to the filesystem buffer - I imagine this would apply to Go's direct access as well since this isn't specific to `du`.

The effect of this in larger volumes is that it will take longer to update the progress, but the restore itself will also take longer in these cases, so this might be okay?

skriss: OK, so sounds like there's probably not really benefit in using `du` over walking the file tree in golang.  I'm fine proceeding with the latter for now, but we'll definitely want to find a way to do some benchmarking and see what the performance is like. We could possibly use `inotify` to get notifications of file changes rather than having to scan, but that's likely much more complex implementation-wise so I wouldn't want to jump there unless needed (and more  research done).

skriss: It also seems to me like using `restic stats` to get the total bytes to be restored should be fine. I don't think there's a scenario where velero would restore into a non-empty directory (we dynamically re-provision new PVs to restore into -- https://github.com/heptio/velero/blob/master/pkg/restore/restore.go#L896). Even if it contains a small number of non-empty files, our "bytes done" estimate would just be slightly inflated, probably not the end of the world. Does that seem reasonable?

prydonius: Yep, that all sounds good to me. Not a real benchmark, but with a quick try in a 310GB directory on my local machine `du` and `dugo` appear to have similar execution times (after purging disk buffers):

```
$ sudo purge
$ time du -sh
310G	.
du -sh  0.00s user 0.01s system 54% cpu 0.014 total
$ sudo purge
$ time dugo -h .
310.0G	 ./
dugo -h .  0.00s user 0.00s system 48% cpu 0.014 total
```

> We could possibly use inotify to get notifications of file changes rather than having to scan

That's definitely an interesting idea, but agree that we don't really need to consider that unless we see performance issues in the current approach.

> Even if it contains a small number of non-empty files, our "bytes done" estimate would just be slightly inflated, probably not the end of the world. Does that seem reasonable?

That sounds good to me, given we'll set BytesDone = TotalBytes once the restic operation completes successfully, any residual bytes probably won't even be exposed to the user.

prydonius: Okay updated the proposal to reflect what we've discussed here.

design/restic-backup-and-restore-progress.md
```diff
@@ -0,0 +1,114 @@
...
+## High-Level Design
+
+### restic backup progress
+
+The `restic backup` command provides progress reporting to stdout in JSON format, which includes the completion percentage of the backup.
+This progress will be read on some interval and the PodVolumeBackup Custom Resource's (CR) status will be updated with this information.
+
+### restic restore progress
+
+The `restic ls` command returns the number of files in a backup, including the size of each file.
```
skriss: let's update this to describe the `restic stats` approach.

design/restic-backup-and-restore-progress.md
```diff
@@ -0,0 +1,114 @@
...
+{"message_type":"status","percent_done":0.0028089887640449437,"total_files":10,"total_bytes":49461329920,"bytes_done":138936320,"current_files":["/file1","/file3"]}
+```
+
+The [command factory for backup](https://github.com/heptio/velero/blob/af4b9373fc73047f843cd4bc3648603d780c8b74/pkg/restic/command_factory.go#L37) will be updated to include the `--json` flag.
+The code to run the `restic backup` command (https://github.com/heptio/velero/blob/af4b9373fc73047f843cd4bc3648603d780c8b74/pkg/controller/pod_volume_backup_controller.go#L241) will be changed to include a Goroutine that reads from the command's stdout stream.
+The implementation of this will largely follow [@jmontleon's PoC](https://github.com/fusor/velero/pull/4/files) of this.
+The Goroutine will periodically read the stream (every 10 seconds) and get the last printed status line.
+The line will be decoded as JSON and the `bytes_done` and `total_bytes` property will be read and the PodVolumeBackup will be patched to update `status.Progress.BytesDone` and `status.Progress.TotalBytes` respectively.
```
skriss: Looking at the sample output, it seems like we have to wait for the "scan_finished" line to get a complete TotalBytes number, and then after that it doesn't change. Can we wait until we get this line to set the TotalBytes val (and then not have to set it again going fwd)?

design/restic-backup-and-restore-progress.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+### restic backup progress
+
+restic added support for [streaming JSON output for the `restic backup` command](https://github.com/restic/restic/pull/1944) in 0.9.5.
+Our current images ship restic 0.9.4, and so the Dockerfile will be updated to pull the new version: https://github.com/heptio/velero/blob/af4b9373fc73047f843cd4bc3648603d780c8b74/Dockerfile-velero#L21.
+With the `--json` flag, `restic backup` outputs single lines of JSON reporting the status of the backup:
+
+```
+{"message_type":"status","percent_done":0,"total_files":1,"total_bytes":21424504832}
```
nrb: I notice this includes a `percent_done` field. Is this excluded in order to provide symmetry for backup and restore?

design/restic-backup-and-restore-progress.md
```diff
@@ -0,0 +1,114 @@
...
+
+### restic restore progress
+
+The `restic stats <snapshot_id> --json` command provides information about the size of backups:
+
+```
+{"total_size":10558111744,"total_file_count":11}
+```
+
+Before beginning the restore operation, we can use the output of `restic stats` to get the total size of the backup.
```
nrb: Is there any reason we can't set the PVR's `TotalBytes` to the corresponding PVB's `TotalBytes`? We don't really allow selective restores within restic, so this may save some effort.

design/restic-backup-and-restore-progress.md
```diff
@@ -0,0 +1,114 @@
...
+To get the current size of the volume, we will recursively walkthrough all files in the volume to accumulate the total size.
+The current total size is the number of bytes transferred so far and the PodVolumeRestore will be patched to update `status.Progress.BytesDone`.
+
+Once the restore has completed successfully, the PodVolumeRestore will be patched to set `status.Progress.BytesDone = status.Progress.TotalBytes`.
+This is done since the main thread may cause early termination of the Goroutine once the operation has finished, preventing a final update to the `BytesDone` property.
+
+### Velero CLI changes
+
+The output that describes detailed information about [PodVolumeBackups](https://github.com/heptio/velero/blob/559d62a2ec99f7a522924348fc4a173a0699813a/pkg/cmd/util/output/backup_describer.go#L349) and [PodVolumeRestores](https://github.com/heptio/velero/blob/559d62a2ec99f7a522924348fc4a173a0699813a/pkg/cmd/util/output/restore_describer.go#L160) will be updated to calculate and display a completion percentage from `status.Progress.TotalBytes` and `status.Progress.BytesDone` if available.
```
nrb: Do you intend to add any changes to the `velero backup create --watch` functionality? Currently, this just prints `.`s until the backup completes.

If this is out of scope for this PR, that's fine, just wondering if you'd thought of it.

design/restic-backup-and-restore-progress.md
skriss: not a bad idea -- it won't work for backups taken prior to v1.2, but we could start by checking the PVB to see if it has a TotalBytes, and use it if available.

design/restic-backup-and-restore-progress.md
prydonius: I was originally going to use this, but we also want to report the TotalBytes and BytesDone in status for both restore and backup so we can ignore `percent_done` and just calculate that in the Velero CLI.

design/restic-backup-and-restore-progress.md
prydonius: Oops, missed this

design/restic-backup-and-restore-progress.md
prydonius: Ah good point, agree we should wait till we see that.

> Can we wait until we get this line to set the TotalBytes val (and then not have to set it again going fwd)?

That's fine, although I was thinking it probably doesn't hurt to read and set it every time since we're just doing a patch anyway.

design/restic-backup-and-restore-progress.md
prydonius: That's a good idea, I guess the only downside to doing this (apart from it not being available for pre v1.2 PVBs) is the need to add a PVB lister to the PVR controller, not a big deal but just running the restic command is a little less work. Do we think that getting this from the PVB will be faster than running `restic stats`?

design/restic-backup-and-restore-progress.md
prydonius: Ah I hadn't thought of that, good point. I think it's a great idea, but perhaps out-of-scope for this issue. Let's create another issue for it and we can follow up with that later?

design/restic-backup-and-restore-progress.md
nrb: True - if if requires more dev work to make the controller aware of PVBs it may not be a net gain. It _might_ be faster since it's a local network interaction as opposed to restic hitting the bucket which is likely to be an external, outbound check, but I have no data to back that up.

design/restic-backup-and-restore-progress.md
nrb: Yep, seems like a separate issue.

design/restic-backup-and-restore-progress.md
skriss: Yeah, i'm guessing it's a little faster runtime-wise, but not enough to be material. I'm fine with whatever's simpler to implement.

prydonius: Sounds good, let's go with using `restic stats` for now, and we can always revisit if we feel like we want to speed this up.

design/restic-backup-and-restore-progress.md
prydonius: Just thinking about this more, @jmontleon's PoC of this (https://github.com/fusor/velero/pull/4/files) would only decode the last seen line of JSON. In order to look for the `scan_finished` line, we'd have to decode every line (or perhaps just grep for "scan_finished").

Also, judging by the output, `scan_finished` will be emitted before the output starts containing the `bytes_done` field, so it would be simpler to just assume that if the `bytes_done` field is there then `total_bytes` will be correct and we're ready to start updating the progress in the PVB. I think this will be reliable enough since restic starts calculating its own `percent_done` using these after the `scan_finished`. WDYT?

design/restic-backup-and-restore-progress.md
prydonius: updated!

design/restic-backup-and-restore-progress.md
skriss: OK, yeah that makes sense and sounds like a simpler approach - wait for `bytes_done` to appear.

PR #1767, opened 2019-08-16
Add the prefix to BSL config map so that object stores can use it whe…

betta1:
…n initializing

Signed-off-by: Antony Bett <antony.bett@gmail.com>

### DIFF
--- a/pkg/cloudprovider/config.go
+++ b/pkg/cloudprovider/config.go
@@ -23,11 +23,11 @@ import (
 
 // ValidateObjectStoreConfigKeys ensures that an object store's config
 // is valid by making sure each `config` key is in the `validKeys` list.
-// The special key "bucket" is always considered valid.
+// The special keys "bucket" and "prefix" are always considered valid.
 func ValidateObjectStoreConfigKeys(config map[string]string, validKeys ...string) error {
-	// `bucket` is automatically added to all object store config by
-	// velero, so add it as a valid key.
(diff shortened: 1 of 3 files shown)


betta1:
@nrb thanks, I updated the comment to include prefix.

nrb:
@betta1 Looks good - how time sensitive is this? I'm of the opinion this is a low-impact change that we could get in for v1.1 release. We don't officially have a feature freeze, either.

@skriss thoughts?

betta1:
Thanks @nrb, yes I agree it will be great if we can get this in for v1.1 release since our plugin requires the prefix during initialization.

skriss:
You'll need to add `prefix` here as well: https://github.com/heptio/velero/blob/master/pkg/cloudprovider/config.go#L30

Let's make sure to manually test this, but I'm fine merging it before v1.1 given the low impact.

pkg/cloudprovider/config.go
```diff
@@ -27,7 +27,7 @@ import (
 func ValidateObjectStoreConfigKeys(config map[string]string, validKeys ...string) error {
 	// `bucket` is automatically added to all object store config by
```
skriss: pls update this comment, and the godoc comment on this function, to reference "prefix" as well.

betta1:
Thanks @skriss, I added the prefix to list of valid keys and updated the godoc as well.

skriss:
thanks - looks good, but i'm going to do a quick test as well

PR #1769, opened 2019-08-16
allow self signed certs with insecureSkipVerify

### LINKED ISSUE #1027: Support custom CA bundles for object storage connections
I am trying to figure out how I can have Ark send backups to a Minio endpoint with our own SSL CA. Right now I am getting the x509 error. I am hoping there is either a ignoreSSL option or even a way to provide a cert bundle to ark.

s12chung:
adds `insecureSkipVerify` server config for AWS storage and `--insecureSkipVerify` flag on client for self-signed certs

if needed, let me know how to add a test. I've tested manually and we're using a variant of the code as a team. most of the code is simply passing along the `insecureSkipVerify` flag around until `InsecureSkipVerify` is set on the http client.

closes #1027

### DIFF
--- a/pkg/cmd/cli/backup/describe.go
+++ b/pkg/cmd/cli/backup/describe.go
@@ -33,8 +33,9 @@ import (
 
 func NewDescribeCommand(f client.Factory, use string) *cobra.Command {
 	var (
-		listOptions metav1.ListOptions
-		details     bool
+		listOptions        metav1.ListOptions
+		details            bool
+		insecureSkipVerify bool
 	)
 
 	c := &cobra.Command{
@@ -71,7 +72,7 @@ func NewDescribeCommand(f client.Factory, use string) *cobra.Command {
 					fmt.Fprintf(os.Stderr, "error getting PodVolumeBackups for backup %s: %v\n", backup.Name, err)
 				}
 
-				s := output.DescribeBackup(&backup, deleteRequestList.Items, podVolumeBackupList.Items, details, veleroClient)
+				s := output.DescribeBackup(&backup, deleteRequestList.Items, podVolumeBackupList.Items, details, veleroClient, insecureSkipVerify)
 				if first {
 					first = false
 					fmt.Print(s)
@@ -85,6 +86,7 @@ func NewDescribeCommand(f client.Factory, use string) *cobra.Command {
 
 	c.Flags().StringVarP(&listOptions.LabelSelector, "selector", "l", listOptions.LabelSelector, "only show items matching this label selector")
 	c.Flags().BoolVar(&details, "details", details, "display additional detail in the command output")
+	c.Flags().BoolVar(&insecureSkipVerify, "insecureskipverify", insecureSkipVerify, "accept any TLS certificate presented by the storage service")
 
 	return c
 }
--- a/pkg/cmd/cli/backup/describe.go
+++ b/pkg/cmd/cli/backup/describe.go
@@ -86,7 +86,7 @@ func NewDescribeCommand(f client.Factory, use string) *cobra.Command {
 
 	c.Flags().StringVarP(&listOptions.LabelSelector, "selector", "l", listOptions.LabelSelector, "only show items matching this label selector")
 	c.Flags().BoolVar(&details, "details", details, "display additional detail in the command output")
-	c.Flags().BoolVar(&insecureSkipVerify, "insecureskipverify", insecureSkipVerify, "accept any TLS certificate presented by the storage service")
+	c.Flags().BoolVar(&insecureSkipVerify, "insecureskipverify", insecureSkipVerify, "do not verify the TLS certificate for storage requests. This is susceptible to man-in-the-middle attacks.")
 
 	return c
 }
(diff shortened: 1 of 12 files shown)


nrb:
Thanks @s12chung! I'm curious, what object storage system are you using with self-signed certs?

s12chung:
happy to contribute :) we're using minio

pkg/cmd/cli/backup/describe.go
```diff
@@ -85,6 +86,7 @@ func NewDescribeCommand(f client.Factory, use string) *cobra.Command {
 
 	c.Flags().StringVarP(&listOptions.LabelSelector, "selector", "l", listOptions.LabelSelector, "only show items matching this label selector")
 	c.Flags().BoolVar(&details, "details", details, "display additional detail in the command output")
+	c.Flags().BoolVar(&insecureSkipVerify, "insecureskipverify", insecureSkipVerify, "accept any TLS certificate presented by the storage service")
```
carlisia: Would it be good to add something like `not secure`, or `not recommended for production` to all the cli documentation?

site/docs/master/api-types/backupstoragelocation.md
```diff
@@ -53,6 +53,7 @@ The configurable parameters are as follows:
 | `kmsKeyId` | string | Empty | *Example*: "502b409c-4da1-419f-a16e-eif453b3i49f" or "alias/`<KMS-Key-Alias-Name>`"<br><br>Specify an [AWS KMS key][10] id or alias to enable encryption of the backups stored in S3. Only works with AWS S3 and may require explicitly granting key usage rights.|
 | `signatureVersion` | string | `"4"` | Version of the signature algorithm used to create signed URLs that are used by velero cli to download backups or fetch logs. Possible versions are "1" and "4". Usually the default version 4 is correct, but some S3-compatible providers like Quobyte only support version 1.|
 | `profile` | string | "default" | AWS profile within the credential file to use for given store |
+| `insecureSkipVerify` | bool | `false` | Set this to `true` if you do not want to verify the CA certificate for storage requests--like self-signed certs in Minio. |
```
carlisia: Please add text warning that this is unsafe.

s12chung: `Set this to `true` if you do not want to verify the TLS certificate for storage requests--like self-signed certs in Minio. This is susceptible to man-in-the-middle attacks and is not recommended for production.`

pkg/cmd/cli/backup/describe.go
s12chung: `do not verify the TLS certificate for storage requests. This is susceptible to man-in-the-middle attacks.` -- think `man-in-the-middle attacks` says it's dangerous. since this is the client and the flag is non-permanent, I kept it shorter.

pkg/cmd/cli/backup/describe.go
```diff
@@ -85,6 +86,7 @@ func NewDescribeCommand(f client.Factory, use string) *cobra.Command {
 
 	c.Flags().StringVarP(&listOptions.LabelSelector, "selector", "l", listOptions.LabelSelector, "only show items matching this label selector")
 	c.Flags().BoolVar(&details, "details", details, "display additional detail in the command output")
+	c.Flags().BoolVar(&insecureSkipVerify, "insecureskipverify", insecureSkipVerify, "do not verify the TLS certificate for storage requests. This is susceptible to man-in-the-middle attacks.")
```
prydonius: Any reason to not have this be `insecure-skip-verify`?  It might also be a good idea to use the same flag `kubectl` uses: `insecure-skip-tls-verify`.

prydonius: We should also probably clarify here that this flag is for storage bucket requests and not requests to the Kubernetes API server

carlisia: Good point 👍

s12chung: good idea. changed it all to `insecure-skip-tls-verify`

s12chung: hmm... it already says `do not verify the TLS certificate for storage requests`

so I only to the end of that: `do not verify the TLS certificate for storage requests only`

prydonius: sorry I misread that, you're right that it was good as it was, but the change doesn't hurt, thanks!

prydonius:
@carlisia @skriss @nrb I assume we want to hold this back until 1.2?

skriss:
we need to discuss further if we want to make this option available to folks. the better change would be to support custom CA bundles per https://github.com/heptio/velero/issues/1027#issuecomment-523119113. @s12chung is that an option for you?

Regardless, we're not going to include this in v1.1 so we're backing out the merge with #1776.

carlisia:
My bad for merging!

s12chung:
@skriss then both the server and client would have to retrieve the certs in order to do operations. am I correct?

hmm.... would need to talk to product owners about this. and need to play around with the code of course. would it be ok to keep `insecureSkipVerify` as an option for now?

macevil:
I thought that skipping is merged into the master, but I still get this error:

``` bash
velero install --provider aws --bucket velero-2c2fe3cc --secret-file ./credentials-velero --use-volume-snapshots=false --backup-location-config region=default,s3ForcePathStyle="true",s3Url=https://os-s3.domain.de --plugins velero/velero-plugin-for-aws:v1.0.0 --image velero/velero:v1.3.0 --insecure-skip-tls-verify
Error: unknown flag: --insecure-skip-tls-verify
```

Any ideas?

skriss:
@macevil for the server, this is a config option on BackupStorageLocations, so you should append `,insecureSkipTLSVerify="true"` to the value of your `--backup-location-config` flag on the `velero install` command.  If you run `velero backup/restore logs`, `velero backup/restore describe --details` or `velero backup download`, you can use the `--insecure-skip-tls-verify` flag on those commands.

macevil:
@skriss Oh, thanks, I was on the hose. 🤦🏻‍♂️

PR #1779, opened 2019-08-20
rename PV during restore when cloning a namespace if PV already exists in-cluster

### LINKED ISSUE #192: Support restoring a PV into the same cluster

skriss:

closes #192

TODO:
- [ ] add unit test cases (positive cases & ensure adequate negative cases)
- [ ] think about possible ways to simplify this code

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	uuid "github.com/satori/go.uuid"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -275,6 +276,7 @@ func (kr *kubernetesRestorer) Restore(
 		},
 		resourceClients: make(map[resourceClientKey]client.Dynamic),
 		restoredItems:   make(map[velero.ResourceIdentifier]struct{}),
+		renamedPVs:      make(map[string]string),
 	}
 
 	return restoreCtx.execute()
@@ -366,6 +368,7 @@ type context struct {
 	extractor                  *backupExtractor
 	resourceClients            map[resourceClientKey]client.Dynamic
 	restoredItems              map[velero.ResourceIdentifier]struct{}
+	renamedPVs                 map[string]string
 }
 
 type resourceClientKey struct {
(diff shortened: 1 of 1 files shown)


pkg/restore/restore.go
```diff
@@ -876,15 +879,32 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
...
-			shouldRestoreSnapshot, err := ctx.shouldRestore(name, resourceClient)
+			renamePV, err := renamePV(ctx, obj, resourceClient)
 			if err != nil {
-				addToResult(&errs, namespace, errors.Wrapf(err, "error waiting on in-cluster persistentvolume %s", name))
+				addToResult(&errs, namespace, err)
 				return warnings, errs
 			}
 
+			var shouldRestoreSnapshot bool
+			switch renamePV {
```
prydonius: any reason to use switch over if/else here?

pkg/restore/restore.go
```diff
@@ -1103,6 +1148,45 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
 	return warnings, errs
 }
 
+// renamePV returns a boolean indicating whether a persistent volume should be given a new name
+// before being restored, or an error if this cannot be determined. A persistent volume will be
+// given a new name if and only if (a) a PV with the original name already exists in-cluster, and
+// (b) in the backup, the PV is claimed by a PVC in a namespace that's being remapped during the
+// restore.
+func renamePV(ctx *context, obj *unstructured.Unstructured, client client.Dynamic) (bool, error) {
```
prydonius: small nitpick: we might want to call this `shouldRenamePV` to make it clearer that this call doesn't do the renaming

pkg/restore/restore.go
skriss: not really, i think this started out being more complicated and then got simplified. switching.

skriss:
Addressed comments so far + added test cases

pkg/restore/restore.go
```diff
@@ -876,15 +883,31 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
...
 				return warnings, errs
 			}
 
+			var shouldRestoreSnapshot bool
+			if shouldRenamePV {
+				// if we're renaming the PV, we're going to give it a new random name,
+				// so we can assume it doesn't already exist in the cluster and therefore
+				// we should proceed with restoring from snapshot.
+				shouldRestoreSnapshot = true
+			} else {
```
carlisia: You could delete the `if` statement up there and instead of the `else` here do: `if ! shouldRenamePV`

pkg/restore/restore.go
```diff
@@ -876,15 +883,31 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
...
+				// Check if the PV exists in the cluster before attempting to create
+				// a volume from the snapshot, in order to avoid orphaned volumes (GH #609)
+				shouldRestoreSnapshot, err = ctx.shouldRestore(name, resourceClient)
+				if err != nil {
+					addToResult(&errs, namespace, errors.Wrapf(err, "error waiting on in-cluster persistentvolume %s", name))
+					return warnings, errs
+				}
+			}
+
 			if shouldRestoreSnapshot {
```
carlisia: If you make the suggested change above, you could move the code from within this `if` to inside the `if ! shouldRenamePV`, right after the return on error.

pkg/restore/restore.go
```diff
@@ -893,6 +916,23 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
 				}
 				obj = updatedObj
 			}
+
+			if shouldRenamePV {
```
carlisia: If you make both suggested changes above, then this could be an `else` to the `if !shouldRenamePV`. I think this would work.

pkg/restore/restore.go
```diff
@@ -893,6 +916,23 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
...
+				obj.SetName(newName)
+
+				// add the original PV name as an annotation
+				annotations := obj.GetAnnotations()
+				if annotations == nil {
+					annotations = map[string]string{}
+				}
+				annotations["velero.io/original-pv-name"] = oldName
+				obj.SetAnnotations(annotations)
+			}
```
carlisia: Add a line after this bracket.

pkg/restore/restore.go
skriss: I might be missing something, but I don't think this can move into the `if !shouldRenamePV { ... }` because it needs to be executed whether `shouldRenamePV` is true or not - `shouldRestoreSnapshot` can potentially be `true` in both cases.

PR #1800, opened 2019-08-23
Add script for pushing container images via Travis

nrb:


nrb:
This isn't completely hooked up to Travis yet, and some of the variable names may need to change.

However, I'd appreciate double checking my script's logic for tagging latest against the requirements noted in the issue.

hack/gcr-push.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+# Return value is written into LATEST
+LATEST=""
+function latest_release() {
```
prydonius: hmm can you not just get this from `git describe --abbrev=0 --tags` or `git describe --tags $(git rev-list --tags --max-count=1)`? This is what we used in Kubeapps: https://github.com/kubeapps/kubeapps/blob/8df8612aac1a3ab19e738d7bdb2956f9bf0127e6/script/chart_sync_utils.sh#L23

hack/gcr-push.sh
```diff
@@ -0,0 +1,37 @@
...
+}
+
+
+# Always publish for master
+if [ "$BRANCH" == "master" ]; then
+    VERSION="$BRANCH" make all-containers all-push
+fi
+
+# Publish when TRAVIS_TAG is defined.
+if [ ! -z "$TRAVIS_TAG" ]; then
```
prydonius: According to the [Travis environment variables](https://docs.travis-ci.com/user/environment-variables/), this should be set to the name of the tag that triggered the build, which should also be the one we want to push images for, right? Can we not just use that directly?

nrb:
Yeah, we don't want any pre-release versions being tagged as latest. They should still get pushed, just not marked as latest.

Your `git describe --tags $(git rev-list --tags --max-count=1)` command appears to do the correct thing on my machine, even with pre-release tags included. I'll see if this is consistent, if so then I think it'll be better than my function.

nrb:
Since `git rev-list` lists commits in reverse chronological order, I think it's usable here. What I don't know is if we do a release where no commits are added between a pre-release and the final release, what will it return?

nrb:
Ok, testing this, it looks like when 2 tags point to the same commit, the one created most recently is returned.

hack/gcr-push.sh
nrb: By directly, do you mean within the Makefile? I'd like to avoid that so our Makefile isn't linked to the provider.

hack/gcr-push.sh
nrb: I think this will be usable, but we'll still need the pre-release version check since we don't want to have users who are pulling `latest` get a pre-release version.

prydonius:
Thanks for explaining and looking into whether that command will work. Sounds like that's what we want so we can replace that function

nrb:
@prydonius The more I thought about it, the more I think you're right that we could use the `$TRAVIS_TAG` versus any git commands. The only addition we really need is to check if that tag's pre-release.

nrb:
>The only addition we really need is to check if that tag's pre-release.

Doh, this is incorrect. We also need to make sure it's also the highest semantic version. So far example, if we release `v1.2.0` followed by `v1.1.1`, the `latest` is `v1.2.0`, not `v1.1.1`

prydonius:
Ah yes, I was thinking we wouldn't need to build older tags, but you're right that we would need to for patch versions.

skriss:
> Doh, this is incorrect. We also need to make sure it's also the highest semantic version. So far example, if we release v1.2.0 followed by v1.1.1, the latest is v1.2.0, not v1.1.1

Yep needs to be highest semver tag, not just most recent.

nrb:
From my local tests, the previous looping approach will still get `v.1.1.0` even if I tag `v1.0.1` against a more recently commit. The `git describe --tags` command will get the `v1.01` tag instead.

.travis.yml
```diff
@@ -8,4 +8,20 @@ sudo: required
 services:
   - docker
 
-script: hack/ci-check.sh
+script:
+- export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
```
prydonius: ouch https://github.com/travis-ci/travis-ci/issues/6652 :(

.travis.yml
```diff
@@ -8,4 +8,20 @@ sudo: required
...
+- echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, PR=$TRAVIS_PULL_REQUEST, BRANCH=$BRANCH"
+- hack/ci-check.sh
+
+# gcloud sdk is needed to upload docker images
+before_deploy:
+- if [ ! -d ${HOME}/google-cloud-sdk ]; then curl https://sdk.cloud.google.com | bash
+  /dev/stdin --disable-prompts; fi
+
+deploy:
+  -provider: script
```
prydonius: the indentation for `-` here is different

hack/gcr-push.sh
```diff
@@ -17,27 +22,31 @@ function latest_release() {
...
+TAG_LATEST=false
+if [[ "$TRAVIS_TAG" == "$HIGHEST" ]]; then
+    TAG_LATEST=true
 fi
+
+openssl aes-256-cbc -K $encrypted_f58ab4413c21_key -iv $encrypted_f58ab4413c21_iv -in heptio-images-fac92d2303ac.json.enc -out heptio-images-fac92d2303ac.json -d
+gcloud auth activate-service-account --key-file heptio-images-fac92d2303ac.json
+unset GIT_HTTP_USER_AGENT
+
+VERSION="$VERSION" TAG_LATEST="$TAG_LATEST" DOCKER="gcloud docker -- " make all-containers all-push
```
prydonius: isn't `gcloud docker` deprecated? https://cloud.google.com/sdk/gcloud/reference/docker

hack/gcr-push.sh
```diff
@@ -0,0 +1,52 @@
...
+TAG_LATEST=false
+if [[ "$TRAVIS_TAG" == "$HIGHEST" ]]; then
+    TAG_LATEST=true
+fi
+
+openssl aes-256-cbc -K $encrypted_f58ab4413c21_key -iv $encrypted_f58ab4413c21_iv -in heptio-images-fac92d2303ac.json.enc -out heptio-images-fac92d2303ac.json -d
+gcloud auth activate-service-account --key-file heptio-images-fac92d2303ac.json
+unset GIT_HTTP_USER_AGENT
+
+VERSION="$VERSION" TAG_LATEST="$TAG_LATEST" DOCKER="gcloud docker -- " make all-containers all-push
```
skriss: Related to https://github.com/heptio/velero/pull/1800/files#r319272305 -- we don't use the `$DOCKER` env var in our Makefile AFAIK so setting it here to the gcloud cmd won't have any effect - but we should be able to just use straight docker to push

.travis.yml
nrb: Yeah, this isn't great, and I borrowed this line from Sonobouy. If we have to do this logic, I prefer we do it once and the rest of the scripts can assume it's there.

hack/gcr-push.sh
nrb: This has now been fixed - the problem was that Travis had an older version of `gcloud` which didn't play well with credential helpers, and needed to use the `gcloud docker --` method. The latest changes update gcloud so that the credential helper method works.

hack/gcr-push.sh
nrb: https://github.com/heptio/velero/pull/1800#discussion_r320961780. Travis's build environment is not up to date.

nrb:
@prydonius @skriss I believe this is ready now. The `latest` tagging logic appears to have worked on my fork, and using Xenial + updating gcloud makes it so we don't have to modify our Makefile's usage of the docker command.

Makefile
```diff
@@ -152,7 +152,7 @@ all-containers:
 	$(MAKE) container BIN=velero-restic-restore-helper
 	$(MAKE) build-fsfreeze
 
-container: verify test .container-$(DOTFILE_IMAGE) container-name
+container: .container-$(DOTFILE_IMAGE) container-name
```
skriss: Just to confirm - the removal of verify/test here is because we've already run them as part of the "build", so we don't need to re-run them?

nrb: That's correct.

PR #1801, opened 2019-08-23
create pkg/archive with functionality extracted from pkg/restore

skriss:
Still a WIP as I need to add/do some more testing and nail down some details, but wanted to share as-is for input.

I'm trying to continually do some simplification on `pkg/restore` where possible, and this seemed like an obvious incremental step. This takes logic around how to navigate a backup tarball as extracted to the local filesystem and encapsulates it in its own package for consumption by `pkg/restore.`

### DIFF
--- a/pkg/archive/extractor.go
+++ b/pkg/archive/extractor.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package restore
+package archive
 
 import (
 	"archive/tar"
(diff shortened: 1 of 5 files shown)


nrb:
I think this new package makes a lot of sense!

skriss:
> I was thinking maybe the name of the package could be archival?

Personally I prefer `archive` as it's a noun, which is consistent with our other package names, but can consider other options - let's see what others think.

skriss:
this is good to merge from my perspective unless there are any more comments!

PR #1810, opened 2019-08-28
Add support for GKE Workload Identity

### LINKED ISSUE #1657: GKE Workload Identity Support
Installing Velero in a cluster that supports GKE Workload Identity is not possible right now. Workload Identity is the recommended way to access Google Cloud services from within GKE due to its improved security properties and manageability. With WI there is no need to pass a secrets key file for the service account. In the latest velero version if you don't pass a secrets content or the secrets does not exist in the namespace, you'll get a failing POD with an error like this one:
```
"Checking that all backup storage locations are valid" logSource="pkg/cmd/server/server.go:431"
An error occurred: some backup storage locations are invalid: error getting backup store for location "default": rpc error: code = Unknown desc = open /credentials/cloud: no such file or directory 
```

The next words are based on a Velero installation via Helm, but it could be applied with kubectl.
One possible solution could be to add a workload identity option in the values file. Based on that value, we may decide how the GCP plugin authenticates the velero-server.

Workload Identity works matching a GCP Service Account with a  Kubernetes Service Account. To achieve this we only need to add an annotation to that velero service account.
Once the K8s SA receives all the GCP SA roles and permissions, there is no need of passing any other type of credentials via secrets.

I have tested the Velero SA after assigning it the GCP SA permissions and works beautifully.

Annotation to add to the Velero SA:
```
serviceAccountAnnotations:
  iam.gke.io/gcp-service-account: GCP_SA_NAME@[GCP_PROJECT_ID].iam.gserviceaccount.com
```

**Environment:**

- Velero version (use `velero version`): 2.1.1
- Kubernetes version (use `kubectl version`): 1.12.9-gke.7
- Cloud provider or hardware configuration: GCP
- OS (e.g. from `/etc/os-release`): Container-Optimized OS (cos)

JWong101:
Signed-off-by: Joshua Wong <joshua99wong@gmail.com>

This closes #1657

Instead of solely relying on the private key for creating new GCP API clients, gcp/object_store.go and gcp/volume_snapshotter.go were changed to use the GCP Default Credentials, since that automatically uses the compute engine tokens if the GCP Service Account key file isn't available.

 In order to continue using signed urls, the GCP ObjectStore uses the SignBlob API with the service account name passed in via the config.

Also, I added an install option to specify specific ServiceAccount annotations, since the Kubernetes Service Account needs a certain annotation in order to masquerade as the GCP Service Account.

Steps to test this:
1. Create a cluster with workload identity enabled
2. Create the GSA with the permissions that velero needs in order to function
3. Grant the GSA iam.serviceAccountTokenCreator role to itself
4. Add the IAM policy binding: gcloud iam service-accounts add-iam-policy-binding \
  --role roles/iam.workloadIdentityUser \
  --member "serviceAccount:[PROJECT_ID].svc.id.goog[velero/velero]" \
  [GSA_NAME]@[PROJECT_ID].iam.gserviceaccount.com
Note: You can still add the policy binding before the KSA is created
5. Install velero: velero install --provider gcp --no-secret --sa-annotations iam.gke.io/gcp-service-account=[GSA_NAME]@[PROJECT_ID].iam.gserviceaccount.com --backup-location-config serviceAccount=[GSA_NAME]@[PROJECT_ID].iam.gserviceaccount.com


skriss:
thanks for the PR @JWong101 and apologies for the delay, we're a little backed up at the moment. Will review soon!

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -18,22 +18,24 @@ package gcp
...
 	"github.com/sirupsen/logrus"
 	"golang.org/x/oauth2/google"
+	"google.golang.org/api/iamcredentials/v1"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
 
 	"github.com/heptio/velero/pkg/cloudprovider"
 )
 
 const credentialsEnvVar = "GOOGLE_APPLICATION_CREDENTIALS"
```
skriss: I *believe* we can get rid of this const now

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -60,28 +62,45 @@ type ObjectStore struct {
...
 }
 
 func (o *ObjectStore) Init(config map[string]string) error {
-	if err := cloudprovider.ValidateObjectStoreConfigKeys(config); err != nil {
+	if err := cloudprovider.ValidateObjectStoreConfigKeys(config, serviceAccountConfig); err != nil {
 		return err
 	}
+	/* Find default token source to extract the GoogleAccessID */
+	ctx := context.Background()
+	creds, err := google.FindDefaultCredentials(ctx)
```
skriss: I think it makes sense to first check for an `err != nil` here and return it if an error's gotten, and only after that proceed to checking the `JSON` field

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -94,16 +113,18 @@ func (o *ObjectStore) Init(config map[string]string) error {
...
 
-	client, err := storage.NewClient(context.Background(), option.WithScopes(storage.ScopeReadWrite))
-	if err != nil {
-		return errors.WithStack(err)
+func (o *ObjectStore) initFromComputeEngine(config map[string]string) error {
+	var err error
+	var ok bool
+	o.googleAccessID, ok = config["serviceAccount"]
+	if !ok {
+		return errors.Errorf("could not find the service account name from the compute engine credentials")
```
skriss: let's update this error to clarify that `serviceAccount` is expected to be provided as a config item in the `BackupStorageLocation`'s config.

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -194,11 +215,37 @@ func (o *ObjectStore) DeleteObject(bucket, key string) error {
...
 
+func (o *ObjectStore) SignBytes(bytes []byte) ([]byte, error) {
+	name := "projects/-/serviceAccounts/" + o.googleAccessID
+	payload := strings.Builder{}
+	if _, err := payload.Write(bytes); err != nil {
+		return nil, err
+	}
+	resp, err := o.iamSvc.Projects.ServiceAccounts.SignBlob(name, &iamcredentials.SignBlobRequest{
+		Delegates: []string{name},
+		Payload:   payload.String(),
```
skriss: can we just say `Payload: string(bytes)` here and drop the `strings.Builder`?

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -194,11 +215,37 @@ func (o *ObjectStore) DeleteObject(bucket, key string) error {
 	return errors.Wrapf(o.client.Bucket(bucket).Object(key).Delete(context.Background()), "error deleting object %s", key)
 }
 
+func (o *ObjectStore) SignBytes(bytes []byte) ([]byte, error) {
+	name := "projects/-/serviceAccounts/" + o.googleAccessID
```
skriss: is there a reference you can link to here in a comment for this approach?

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -194,11 +215,37 @@ func (o *ObjectStore) DeleteObject(bucket, key string) error {
...
+	}
+	resp, err := o.iamSvc.Projects.ServiceAccounts.SignBlob(name, &iamcredentials.SignBlobRequest{
+		Delegates: []string{name},
+		Payload:   payload.String(),
+	}).Context(context.Background()).Do()
+
+	if err != nil {
+		return nil, err
+	}
+	data, err := base64.StdEncoding.DecodeString(resp.SignedBlob)
```
skriss: `return base64.StdEncoding.DecodeString(resp.SignedBlob)`

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -194,11 +215,37 @@ func (o *ObjectStore) DeleteObject(bucket, key string) error {
...
-	return storage.SignedURL(bucket, key, &storage.SignedURLOptions{
+	options := storage.SignedURLOptions{
 		GoogleAccessID: o.googleAccessID,
-		PrivateKey:     o.privateKey,
 		Method:         "GET",
 		Expires:        time.Now().Add(ttl),
-	})
+	}
+	if o.privateKey == nil {
+		options.SignBytes = func(b []byte) ([]byte, error) {
```
skriss: I think you can just say `options.SignBytes = o.SignBytes` here

pkg/cloudprovider/gcp/volume_snapshotter.go
```diff
@@ -60,25 +58,22 @@ func (b *VolumeSnapshotter) Init(config map[string]string) error {
 		return err
 	}
 
-	b.snapshotLocation = config[snapshotLocationKey]
-
-	project, err := extractProjectFromCreds()
+	/* Works with both credential files and the default compute engine service account */
+	creds, err := google.FindDefaultCredentials(oauth2.NoContext, compute.ComputeScope)
```
skriss: If I'm not mistaken, this will error if `GOOGLE_APPLICATION_CREDENTIALS` has a value, but no file is found at that path. By default, `velero install` is going to set this env var to a value of `/credentials/cloud`.  I think we probably need to change the `velero install` logic so if `--no-secret` is specified, these env vars aren't set. @nrb does that make sense to you?

skriss:
@victortrac @salrashid123 would love to have some additional eyes on this if you've got time!

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -194,11 +215,37 @@ func (o *ObjectStore) DeleteObject(bucket, key string) error {
...
 }
 
+func (o *ObjectStore) SignBytes(bytes []byte) ([]byte, error) {
+	name := "projects/-/serviceAccounts/" + o.googleAccessID
+	payload := strings.Builder{}
+	if _, err := payload.Write(bytes); err != nil {
+		return nil, err
+	}
+	resp, err := o.iamSvc.Projects.ServiceAccounts.SignBlob(name, &iamcredentials.SignBlobRequest{
+		Delegates: []string{name},
```
salrashid123: I dont' think you need to specify the `name` here (thats only used normally when you chain different service accounts together

salrashid123:
just a comment on the IAM permissions..if you grant `roles/iam.serviceAccountTokenCreator` that gives the service account that the WI Identity runs as a number of overlapping capabilities.   For example, the WI enabled system can now execute the capabilities below (the `getAccessToken()` is redundant since you already have it for the service account to even start the flow (as provided by `roles/iam.workloadIdentityUser`).   If what you just want is to `signBlob()` an IAM role with just htat permission should work


unfortunately the GCP only has one canned IAM role that covers all of it
```
roles/iam.serviceAccountTokenCreator
    iam.serviceAccounts.get
    iam.serviceAccounts.getAccessToken
    iam.serviceAccounts.implicitDelegation
    iam.serviceAccounts.list
    iam.serviceAccounts.signBlob
    iam.serviceAccounts.signJwt
    resourcemanager.projects.get
    resourcemanager.projects.list
```

so for customer who whated to very tightly limit its access, they'd currently need to define a [custom role](https://cloud.google.com/iam/docs/understanding-custom-roles) with just `iam.serviceAccounts.signBlob` permission

for reference, 
```
roles/iam.workloadIdentityUser
    iam.serviceAccounts.get
    iam.serviceAccounts.getAccessToken
    iam.serviceAccounts.list
```

pkg/cloudprovider/gcp/volume_snapshotter.go
JWong101: I checked the velero install logic, and it doesn't create a file if --no-secret is specified. I also tested it using --no-secret, and it seemes to initialize properly using the workload identity.

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -34,7 +33,6 @@ import (
 	"github.com/heptio/velero/pkg/cloudprovider"
 )
 
-const credentialsEnvVar = "GOOGLE_APPLICATION_CREDENTIALS"
```
skriss: Let's keep this constant around - we're using this in more than one place.

pkg/cloudprovider/gcp/volume_snapshotter.go
skriss: OK -- I'll do some testing as well.

pkg/cloudprovider/gcp/object_store.go
skriss: I can confirm that everything works fine without L228, and keeping it requires additional permissions, so let's remove.

pkg/cloudprovider/gcp/object_store.go
```diff
@@ -194,11 +218,37 @@ func (o *ObjectStore) DeleteObject(bucket, key string) error {
...
 
+/*
+ * Use the iamSignBlob api call to sign the url if there is no credentials file to get the key from.
+ * https://cloud.google.com/iam/credentials/reference/rest/v1/projects.serviceAccounts/signBlob
+ */
+func (o *ObjectStore) SignBytes(bytes []byte) ([]byte, error) {
+	name := "projects/-/serviceAccounts/" + o.googleAccessID
+	resp, err := o.iamSvc.Projects.ServiceAccounts.SignBlob(name, &iamcredentials.SignBlobRequest{
+		Delegates: []string{name},
+		Payload:   string(bytes),
```
skriss: looks like we need to base64-encode the bytes here -- `Payload: base64.StdEncoding.EncodeToString(bytes),` should do it

pkg/cloudprovider/gcp/volume_snapshotter.go
skriss: Yep, you're right - specifying `--no-secret` removes the env vars entirely from the deployment. So we're good here.

skriss:
Added a couple more comments from testing, but this is looking almost ready! It would be great to  add some documentation about this as well -- site/docs/master/gcp-config.md and site/docs/master/api-types/backupstoragelocation.md would be good places to start.  

You can take a look at https://velero.io/docs/v1.1.0/aws-config/ and the bottom section on installing using kube2iam as a reference for how alternate installation methods are documented.

skriss:
ref. https://github.com/heptio/velero/pull/1810#issuecomment-528553593 -- I tested by just adding `iam.serviceAccounts.signBlob` to the custom Velero server role defined [here](https://velero.io/docs/v1.1.0/gcp-config/#create-service-account), and that seemed to do the trick.

skriss:
@JWong101 just checking to see if you wanted to continue moving this forward - if not, I can take it over and give you credit for the commits made.

JWong101:
Yeah, sorry, I've just been busy w/ schoolwork + interviews. I should be able to finish the changes by Friday or Saturday.

pkg/cloudprovider/gcp/object_store.go
JWong101: It doesn't compile if I do it that way, since o.SignBytes is technically a method.

skriss: `options.SignBytes` is a function pointer, so you should be able to assign a pointer to the `o.SignBytes` method to it via `options.SignBytes = o.SignBytes` -- but it can stay as-is.

JWong101: Yeah nvm, I was just dumb.

skriss:
Great, thanks for the updates @JWong101!

PR #1821, opened 2019-08-29
record restic backup progress in PodVolumeBackup

prydonius:
This PR modifies the way we run `restic backup` to include reporting the progress of the backup.


prydonius:
Still WIP, but wanted some feedback on the approach I'm taking here with the callback (updateFunc) function to update the progress from outside the controller. https://github.com/fusor/velero/pull/4/files implemented the RunBackup function in the controller package which avoids the need for the callback, but I'm thinking that it will be more easily testable this way, albeit a bit more confusing?

pkg/restic/exec_commands.go
```diff
@@ -53,3 +62,91 @@ func GetSnapshotID(repoIdentifier, passwordFile string, tags map[string]string,
...
+	stderr, err := cmd.StdoutPipe()
+	if err != nil {
+		return "", "", err
+	}
+
+	cmd.Start()
+
+	// copy command's stdout/err to buffers
+	wg.Add(1)
+	go func() {
```
skriss: are these 2 goroutines for copying stdout/stderr needed, or can we just set `cmd.Stdout = stdoutWriter` and similarly for stderr?

prydonius: Good point, let me check - I assumed it wouldn't work because I tried to use `cmd.StdoutPipe()` directly to scan through but with that `scanner.Scan()` would continuously stream until the command was finished, so the buffer is needed to read up to the current point. Let's see if I can set `cmd.Stdout = stdoutWriter` and it behaves the same way.

prydonius: Looks like that works, thanks for the tip!

pkg/restic/exec_commands.go
```diff
@@ -53,3 +62,91 @@ func GetSnapshotID(repoIdentifier, passwordFile string, tags map[string]string,
 
 	return snapshots[0].ShortID, nil
 }
+
+// RunBackup runs a `restic backup` command and watches the output to provide
+// progress updates to the caller.
+func RunBackup(backupCmd *Command, log logrus.FieldLogger, updateFunc func(velerov1api.PodVolumeOperationProgress)) (string, string, error) {
```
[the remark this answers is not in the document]
prydonius: One issue with this is that if a backup completes too quickly, the progress will never get set. This can easily happen for smaller volumes or if only a few files have changed in an incremental backup. I'm now thinking that this function should call `updateFunc` one last time with the final progress before returning, and we don't need to set the progress to 100% in the controller.

prydonius:
I've been trying a different approach here in order to address https://github.com/heptio/velero/pull/1821#discussion_r319320304, but I'm seeing something strange. The `restic backup --json` command finishes with a summary line, something like:
 ```
{"message_type":"summary","files_new":0,"files_changed":0,"files_unmodified":10,"dirs_new":0,"dirs_changed":0,"dirs_unmodified":0,"data_blobs":0,"tree_blobs":0,"data_added":0,"total_files_processed":10,"total_bytes_processed":21043871744,"total_duration":0.363935634,"snapshot_id":"5e764a0b"}
```
I'm trying to read this line after the command has finished to extract `total_bytes_processed` and use that for a final update of the backup progress. This is consistently the last line when I manually run `restic backup --json` both on my local machine and when exec'ing in the Pod. However, in the controller, it is sometimes the last line and other times the penultimate line (followed by another "status" line). This could be because the last two lines are being written to the buffer around the same time and so the order can be different?

Might need some extra eyes on this next week @skriss.

**Comment by skriss** [comment]
Just ran a `restic backup` cmd locally using restic v0.9.5 - got the following output at the end:

```json
{"message_type":"status","seconds_elapsed":9,"percent_done":1,"total_files":428,"files_done":428,"total_bytes":2168012,"bytes_done":2168012}
{"message_type":"status","seconds_elapsed":10,"percent_done":1,"total_files":428,"files_done":428,"total_bytes":2168012,"bytes_done":2168012}
{"message_type":"status","seconds_elapsed":11,"percent_done":1,"total_files":428,"files_done":428,"total_bytes":2168012,"bytes_done":2168012}
{"message_type":"summary","files_new":428,"files_changed":0,"files_unmodified":0,"dirs_new":0,"dirs_changed":0,"dirs_unmodified":0,"data_blobs":428,"tree_blobs":1,"data_added":2168388,"total_files_processed":428,"total_bytes_processed":2168012,"total_duration":11.171735739,"snapshot_id":"4e9479ef"}
{"message_type":"summary","files_new":428,"files_changed":0,"files_unmodified":0,"dirs_new":0,"dirs_changed":0,"dirs_unmodified":0,"data_blobs":428,"tree_blobs":1,"data_added":2168388,"total_files_processed":428,"total_bytes_processed":2168012,"total_duration":11.171735739,"snapshot_id":"4e9479ef"}
{"message_type":"status","seconds_elapsed":11,"percent_done":1,"total_files":428,"files_done":428,"total_bytes":2168012,"bytes_done":2168012}
```

So duplicate "summary" lines, and a "status" line after the "summary" ones. Seems like there's probably some bugs in the status reporter in restic.

prydonius:
Okay I've filed https://github.com/restic/restic/issues/2389 upstream to see if we can get help on this issue. In the meantime, we can workaround this by looking for the summary line instead of just trying to get the last line.

prydonius:
This is now ready for review, I'll work on the restore progress and CLI changes in a separate PR

pkg/cmd/util/output/restore_describer.go
```diff
@@ -188,7 +188,8 @@ func describePodVolumeRestores(d *Describer, restores []v1.PodVolumeRestore, det
 		restoresByPod := new(volumesByPod)
 
 		for _, restore := range restoresByPhase[phase] {
-			restoresByPod.Add(restore.Spec.Pod.Namespace, restore.Spec.Pod.Name, restore.Spec.Volume)
+			// TODO(adnan): replace last parameter with progress from status (#1749)
```
prydonius: will address this in the PR for restore progress

skriss:
Looks good on a quick pass - I'll do some testing and another pass later on today.

pkg/restic/exec_commands.go
```diff
@@ -53,3 +70,109 @@ func GetSnapshotID(repoIdentifier, passwordFile string, tags map[string]string,
...
+				ticker.Stop()
+				wg.Done()
+				return
+			}
+		}
+	}()
+
+	cmd.Wait()
+	close(quit)
+	wg.Wait()
```
skriss: I don't think we need this wait group, do we? I think the "quit" logic could be simplified - after L120, call `ticker.Stop()` (you'll need to move `ticker`'s declaration to outside the goroutine), then `close(quit)`. The `select` case that receives from `quit` can just return.

prydonius: Yeah, it's mostly left over from the additional goroutines we had for copying stdout/stderr. One issue is that if we're in the middle of the ticker.C case and it takes some time to complete, then we might call `updateFunc` in the goroutine with `BytesDone < TotalBytes` after our call to it at the end of this function. If we wait for the goroutine to finish, we know this will definitely not happen.

That said, the stuff in the `ticker.C` case should be pretty fast for backup, for restore I think we'll need the waitgroup since calculating the size of the volume can take an indeterminate amount of time.

skriss: OK, that's a good point.  I guess the other way you could do this is to send on the `quit` channel rather than closing it, since that would block until it's received from, but I'm fine with this as-is.

prydonius: @skriss ah good point, I didn't think of that. I like the idea of removing the unnecessary waitgroup, so will update this to do that - thanks!

changelogs/unreleased/1821-prydonius
```diff
@@ -0,0 +1 @@
+report backup progress in PodVolumeBackups and expose progress in the velero backup describe --details command
```
nrb: Should we additionally mention the bump in restic version?

prydonius: Good shout, will do that

prydonius: updated, ptal!

pkg/restic/exec_commands.go
```diff
@@ -111,15 +108,14 @@ func RunBackup(backupCmd *Command, log logrus.FieldLogger, updateFunc func(veler
...
 				ticker.Stop()
-				wg.Done()
 				return
 			}
 		}
 	}()
 
 	cmd.Wait()
+	quit <- struct{}{}
 	close(quit)
```
skriss: I'd remove the `close(quit)` for clarity since it's not needed anymore.

prydonius:
addressed all the review comments here - PTAL @nrb @carlisia!

PR #1828, opened 2019-08-30
Sort mounts in test case to avoid intermittent fail

### LINKED ISSUE #1826: Fix bug in restic restore test
Bug is causing very sporadic build failure:
https://travis-ci.org/heptio/velero/builds/578135148#L732

This is the test: https://github.com/heptio/velero/pull/1747/files#diff-da00ae9ed6c89041809723dbf6c54f21R145

As @skriss correctly identified: 
> looks like there’s an iteration over a map in the restic restore action that’s probably causing the issue since maps don’t have guaranteed ordering. easiest thing is probably to sort the volume mounts in the expected and actual in the test prior to comparing

carlisia:
Closes #1826.

### DIFF
--- a/pkg/restore/restic_restore_action_test.go
+++ b/pkg/restore/restic_restore_action_test.go
@@ -18,6 +18,7 @@ package restore
 
 import (
 	"fmt"
+	"sort"
 	"testing"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


carlisia:
Thanks for the reminder to do that. I hadn't, but did just now, all good:

image

Issue #964, opened 2018-10-19
Wait for CRDs to be ready when restoring


ncdc:
Split out from #424 

> It would also be nice if we could wait for a CRD to be "ready" (as indicated in its status) before proceeding; otherwise, trying to create CRs for it might fail.

@skriss:
> How exactly should "ready" be determined? Here's what I see in .status for the backup.ark.heptio.com CRD:

```
status:
  acceptedNames:
    kind: Backup
    listKind: BackupList
    plural: backups
    singular: backup
  conditions:
  - lastTransitionTime: 2018-04-17T18:07:24Z
    message: no conflicts found
    reason: NoConflicts
    status: "True"
    type: NamesAccepted
  - lastTransitionTime: 2018-04-17T18:07:24Z
    message: the initial names have been accepted
    reason: InitialNamesAccepted
    status: "True"
    type: Established
```

@ncdc:
> I think we need to key off of the `Established` condition.

shubheksha:
Hey @ncdc, I'd like to pick this up - can you give me some more info about this?

ncdc:
@shubheksha cool, thanks! Here's an example.

Let's say the backup contains a CRD called `widgets.myapp.com` along with several instances of `Widget`s. Thanks to #962, we restore the CRD first. There is a chance, however, that we could then try to restore a `Widget` instance before the cluster is ready to accept them. If that's the case, our attempt to create one will fail.

One approach would be to wait for the CRD's status to contain the condition `type=Established`, `status="True"`. The downside to this is that it would block anything else from being restored until the CRD is ready (or we exceed a timeout).

@heptio/ark-team any thoughts?

carlisia:
Could the process skip the restore if the CRD is not ready and be in a sort of "need to be re-run" state so the user can verify that and run again?

ncdc:
We're not structured to do that at the moment, but we could certainly explore it. `ark restore create` goes from 0 to 100, trying to restore everything from the backup.

Another option: we could add a "holding queue" of resource types that are waiting for some condition, and once we get that condition, trigger restoring them. I'm not sure the best way to implement that, though. We currently operate serially, working on 1 type of resource before moving on to the next. @skriss and I had talked about having some kind of general restore queue, where we add items to it as we know we want to restore them, and various worker goroutines pop items from the queue and restore them. Maybe a restructuring like that would enable this kind of approach.

carlisia:
[...]

PR #1831, opened 2019-09-02
If includeClusterResources is nil/auto, pull in necessary CRDs

sseago:
Related issue: https://github.com/heptio/velero/issues/1830

This accomplishes everything
that's needed, although there might be room for improvement in avoiding
a GET call for matching CRDs for each resource backed up. An alternative
could be a single call to get all CRDs prior to iterating over resources
and passing this into the backupResource function.

### DIFF
--- a/pkg/backup/backup_test.go
+++ b/pkg/backup/backup_test.go
@@ -596,6 +596,164 @@ func TestBackupResourceFiltering(t *testing.T) {
 	}
 }
 
+// TestCRDInclusion tests whether related CRDs are included, based on
+// backed-up resources and "include cluster resources" flag, and
+// verifies that the set of items written to the backup tarball are
+// correct. Validation is done by looking at the names of the files in
+// the backup tarball; the contents of the files are not checked.
+func TestCRDInclusion(t *testing.T) {
+	tests := []struct {
+		name         string
+		backup       *velerov1.Backup
+		apiResources []*test.APIResource
+		want         []string
+	}{
+		{
+			name: "include cluster resources=auto includes all CRDs when backing up all namespaces",
+			backup: defaultBackup().
+				Result(),
+			apiResources: []*test.APIResource{
+				test.CRDs(
+					builder.ForCustomResourceDefinition("backups.velero.io").Result(),
+					builder.ForCustomResourceDefinition("volumesnapshotlocations.velero.io").Result(),
+					builder.ForCustomResourceDefinition("test.velero.io").Result(),
+				),
+				test.VSLs(
+					builder.ForVolumeSnapshotLocation("foo", "vsl-1").Result(),
+				),
+			},
+			want: []string{
+				"resources/customresourcedefinitions.apiextensions.k8s.io/cluster/backups.velero.io.json",
+				"resources/customresourcedefinitions.apiextensions.k8s.io/cluster/volumesnapshotlocations.velero.io.json",
+				"resources/customresourcedefinitions.apiextensions.k8s.io/cluster/test.velero.io.json",
+				"resources/volumesnapshotlocations.velero.io/namespaces/foo/vsl-1.json",
+			},
+		},
+		{
+			name: "include cluster resources=false excludes all CRDs when backing up all namespaces",
+			backup: defaultBackup().
+				IncludeClusterResources(false).
+				Result(),
+			apiResources: []*test.APIResource{
+				test.CRDs(
+					builder.ForCustomResourceDefinition("backups.velero.io").Result(),
+					builder.ForCustomResourceDefinition("volumesnapshotlocations.velero.io").Result(),
+					builder.ForCustomResourceDefinition("test.velero.io").Result(),
+				),
+				test.VSLs(
+					builder.ForVolumeSnapshotLocation("foo", "vsl-1").Result(),
+				),
+			},
+			want: []string{
+				"resources/volumesnapshotlocations.velero.io/namespaces/foo/vsl-1.json",
+			},
+		},
+		{
+			name: "include cluster resources=true includes all CRDs when backing up all namespaces",
+			backup: defaultBackup().
+				IncludeClusterResources(true).
+				Result(),
+			apiResources: []*test.APIResource{
+				test.CRDs(
+					builder.ForCustomResourceDefinition("backups.velero.io").Result(),
+					builder.ForCustomResourceDefinition("volumesnapshotlocations.velero.io").Result(),
+					builder.ForCustomResourceDefinition("test.velero.io").Result(),
+				),
+				test.VSLs(
+					builder.ForVolumeSnapshotLocation("foo", "vsl-1").Result(),
+				),
+			},
+			want: []string{
+				"resources/customresourcedefinitions.apiextensions.k8s.io/cluster/backups.velero.io.json",
+				"resources/customresourcedefinitions.apiextensions.k8s.io/cluster/volumesnapshotlocations.velero.io.json",
+				"resources/customresourcedefinitions.apiextensions.k8s.io/cluster/test.velero.io.json",
+				"resources/volumesnapshotlocations.velero.io/namespaces/foo/vsl-1.json",
+			},
+		},
+		{
+			name: "include cluster resources=auto includes CRDs with CRs when backing up selected namespaces",
+			backup: defaultBackup().
+				IncludedNamespaces("foo").
+				Result(),
+			apiResources: []*test.APIResource{
+				test.CRDs(
+					builder.ForCustomResourceDefinition("backups.velero.io").Result(),
+					builder.ForCustomResourceDefinition("volumesnapshotlocations.velero.io").Result(),
+					builder.ForCustomResourceDefinition("test.velero.io").Result(),
+				),
+				test.VSLs(
+					builder.ForVolumeSnapshotLocation("foo", "vsl-1").Result(),
+				),
+			},
+			want: []string{
+				"resources/customresourcedefinitions.apiextensions.k8s.io/cluster/volumesnapshotlocations.velero.io.json",
+				"resources/volumesnapshotlocations.velero.io/namespaces/foo/vsl-1.json",
+			},
+		},
+		{
+			name: "include cluster resources=false excludes all CRDs when backing up selected namespaces",
+			backup: defaultBackup().
+				IncludeClusterResources(false).
+				IncludedNamespaces("foo").
+				Result(),
+			apiResources: []*test.APIResource{
+				test.CRDs(
+					builder.ForCustomResourceDefinition("backups.velero.io").Result(),
+					builder.ForCustomResourceDefinition("volumesnapshotlocations.velero.io").Result(),
+					builder.ForCustomResourceDefinition("test.velero.io").Result(),
+				),
+				test.VSLs(
+					builder.ForVolumeSnapshotLocation("foo", "vsl-1").Result(),
+				),
+			},
+			want: []string{
+				"resources/volumesnapshotlocations.velero.io/namespaces/foo/vsl-1.json",
+			},
+		},
+		{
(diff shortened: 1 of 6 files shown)


pkg/backup/resource_backupper.go
```diff
@@ -276,6 +285,80 @@ func (rb *defaultResourceBackupper) backupResource(group *metav1.APIResourceList
...
+	if err != nil {
+		log.WithError(errors.WithStack(err)).Error("Error listing CRDs")
+		return
+	}
+	crds, err := meta.ExtractList(unstructuredCRDList)
+	if err != nil {
+		log.WithError(errors.WithStack(err)).Error("Error extracting CRD list")
+		return
+	}
+	for _, item := range crds {
```
prydonius: I might be misunderstanding the way CRD versioning works, but shouldn't there only be one CRD in this list if found (since two CRDs can't have the same name)? Also, I think we can probably assume that if we're able to fetch a CR, then the version we have is one that is being served by the CRD, so I don't think there's a need to check that the version exists and is being served in the CRD?

pkg/backup/resource_backupper.go
```diff
@@ -276,6 +285,80 @@ func (rb *defaultResourceBackupper) backupResource(group *metav1.APIResourceList
...
+		unstructured, ok := item.(runtime.Unstructured)
+		if !ok {
+			log.Errorf("Unexpected type %T", item)
+			continue
+		}
+
+		crd := apiextv1beta1.CustomResourceDefinition{}
+		itemMarshal, _ := json.Marshal(unstructured)
+		json.Unmarshal(itemMarshal, &crd)
```
prydonius: probably want to use `runtime.DefaultUnstructuredConverter.FromUnstructured` here instead of marshalling/unmarshalling

pkg/backup/resource_backupper.go
sseago: Ahh, yes I think you're correct here. I was just looking at the versioning model, but since CRDs are cluster-scoped, then there will only be one with a given name. So I can really pull out this whole version-checking bit and assume that if I get a CRD, we back it up. I'll do that.

pkg/backup/resource_backupper.go
sseago: I noticed after I submitted this PR that Velero was using the `DefaultUnstructuredConverter` elsewhere. I had pulled the marshal/unmarshal pattern from some of our plugin code which tends to use it inside backup/restore plugins for converting the item. I'll modify this to match what you're doing elsewhere with the `DefaultUnstructuredConterter`.

sseago: Even better, as I'm pulling out the CRD version checking code, I can go back to just working with the Unstructured resource object, so I don't need to worry about converting it with the next iteration of the code.

sseago:
By removing the version-checking while iterating over the CRD list, I realized that what remained was essentially the same as the loop over resource items, but with CRDs instead, so I pulled that out into a separate function and reused it for both the CRD list (although it will always be a list of, at most, one) and the item list.

pkg/backup/resource_backupper.go
```diff
@@ -237,43 +240,89 @@ func (rb *defaultResourceBackupper) backupResource(group *metav1.APIResourceList
...
+	if err != nil {
+		log.WithError(errors.WithStack(err)).Error("Error listing CRDs")
+		return
+	}
+	crds, err := meta.ExtractList(unstructuredCRDList)
+	if err != nil {
+		log.WithError(errors.WithStack(err)).Error("Error extracting CRD list")
+		return
+	}
+	rb.backupItems(log, crdGr, itemBackupper, crds)
```
prydonius: Let's add an info log here to say we found an associated CRD to backup for the gr

pkg/backup/resource_backupper.go
```diff
@@ -237,43 +239,86 @@ func (rb *defaultResourceBackupper) backupResource(group *metav1.APIResourceList
 
 		log.Infof("Retrieved %d items", len(items))
 
+		// back up CRD for resource if found. We should only need to do this is IncludeClusterResources is nil. If it's false
+		// we don't want to back it up, and if it's true it will already be included.
+		if len(items) > 0 && rb.backupRequest.Spec.IncludeClusterResources == nil {
```
skriss: We'll end up doing this once per namespace, if specific namespaces are specified. It would be a little more efficient to do this just once, after L255 (i.e. after the range over `namespacesToList`), assuming we backed up >0 items.

skriss: Another thing to think about: even if `len(items) > 0`, we're not necessarily actually backing up any of those items, as the item backupper will check all the filters, including the `exclude-from-backup` label. So, it's possible that we end up not actually backing up any instances of the CRD, but then still backing up the CRD itself.

I think ideally, we only back up the CRD if we *actually* back up 1+ instances of it. One way we could accomplish this is to change the `backupItem` method to also return a boolean indicating whether the item was actually backed up or not. Then here in the resource backupper, we keep track of whether any call to `backupItem` returned a `true` there, and if so, we back up the CRD.

sseago: Sure. This makes sense -- combined with the below comment, this would just mean keeping track of either a count of items backed up or a boolean "at least one item backed up" across namespaces, then outside the namespace loop we can make the CRD backup call.

sseago: I guess there are actually two ways of considering this. Either we only copy the CRD if at least one of its CRs is actually copied (which is what you suggest here), or we always copy the CRD, even if there are no instances of it here (possibly excluding it if the associated resource is explicitly excluded from backup). My concern with not copying the CRD is if you're backing up and restoring an application with an associated CRD, wouldn't you want the CRD included even if you don't happen to have any CRs at the moment. What if, for example, you have a CRD that uses CRs for a form of work queue. The application grabs a CR, does something, and then deletes the CR. If the queue is empty when you do the backup, your restore won't include the CRD. Then again, if we did this I'm not sure how "include CRDs even if there are no CRs" would differ from CRD behavior if we set includeClusterResources to 'true'. EIther way, `len(items) > 0` is the wrong test, since, as you point out, it doesn't take into account excluded resources.

sseago:
We definitely need to address #964. I'm not so clear on what we'd need in the restore code other than the "wait for CRDs to be ready" bit (which can be based on the `crdsAreReady` function in the install package). This PR adds CRDs to the backup. Once they're in the backup, the restore should include them in the restore, unless I'm missing something here.

pkg/backup/resource_backupper.go
prydonius: > wouldn't you want the CRD included even if you don't happen to have any CRs at the moment. What if, for example, you have a CRD that uses CRs for a form of work queue.

This is a great point actually, though if there are no CRs at the moment, we won't be able to derive the CRD to backup here anyway, right? I think in these cases, the CRD will need to be explicitly included (either with `--include-resources` or `--include-cluster-resources`).

skriss:
> I'm not so clear on what we'd need in the restore code other than the "wait for CRDs to be ready" bit (which can be based on the crdsAreReady function in the install package).

As long as the user is restoring the full backup, then yeah #964 is sufficient. If, however, they're only restoring a single namespace or some other subset of resources from a full-cluster backup, then I think we have a similar issue as we do with backups - i.e. we may be attempting to restore instances of a CRD into a namespace, but we don't know to restore the CRD itself first.  I'm not sure off the top of my head what the best approach is here - we could create an in-tree item action plugin that returns CRDs as "additional items", or we could add some special-case code at the beginning of restoring a resource type that checks if it's a CR and restores the CRD if so.

sseago:
"If, however, they're only restoring a single namespace or some other subset of resources from a full-cluster backup, then I think we have a similar issue as we do with backups - i.e. we may be attempting to restore instances of a CRD into a namespace, but we don't know to restore the CRD itself first. "
Oh. Don't we still restore prioritized resources first, even with single namespace restores? If so, then https://github.com/heptio/velero/pull/962 would make sure the CRD is restored first, once we also wait for it to be ready as called for in #964.

skriss:
OK, I'm looking at the restore code now and it appears to be inconsistent with backups in terms of how cluster-scoped resources are handled during a restore of specific namespaces -- it looks like if `IncludeClusterResources` is nil/auto, *all* cluster-scoped resources in the backup will be restored, even if only a specific namespace is being restored.

TBH that seems like a bug we should fix to be consistent with the backup process, however I do agree that in the current state, you'll get the CRD restored prior to any CRs as long as `IncludeClusterResources` is not false.

sseago:
So the simplistic method of modifying `backupItem` to return a bool ignores the question of additionalItems. It's possible for additionalItems to be backed up even if the item we're backing up here fails. Since the main reason we're returning a bool is to say whether *this* object was backed up (for the purposes of identifying whether we need the associated CRD, that's probably fine. If additionalItems were backed up, and this resource failed, then the CR we're concerned about wasn't backed up, so we return false. It does suggest to me that if we're skipping the item because it's already backed up, we should probably return `true` in that case, though, since it means that this item probably got dragged in via some other resource's additionalItems, which means the CRD isn't (yet) include but is needed. The final element of confusion here is over items that are *only* included as a result of being pulled in as additionalItems. If we pull in a CR via additionalItems but we're *not* backing up that CR's resource as a whole, then we'll miss the CRD. Is this OK, or do we need to do something different to capture potential CRDs in the additionalItems handling? The only time this will happen is if we pull in a CR via additionalItems but *never* run backupResource on the resource that the CR belongs to.

pkg/backup/resource_backupper.go
```diff
@@ -243,37 +246,85 @@ func (rb *defaultResourceBackupper) backupResource(group *metav1.APIResourceList
...
+		"",
+	)
+	if err != nil {
+		log.WithError(err).Error("Error getting dynamic client for CRDs")
+		return
+	}
+
+	unstructured, err := crdClient.Get(gr.String(), metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
```
dymurray: Why do we not log if the CRD was not found?

dymurray: Answered my own question: This means the resource isn't a CR. So extra logging would be noisy and pointless

skriss:
@sseago the code LGTM - my only request is to add some unit test cases around this. I think you probably want a similar structure as already exists in the `TestBackupResourceFiltering` test func, but it probably makes sense to set up a separate test method for this. Let us know if anything's unclear on how to structure the tests.

sseago:
@skriss PR updated with unit tests.

pkg/backup/backup_test.go
```diff
@@ -596,6 +596,164 @@ func TestBackupResourceFiltering(t *testing.T) {
...
+// the backup tarball; the contents of the files are not checked.
+func TestCRDInclusion(t *testing.T) {
+	tests := []struct {
+		name         string
+		backup       *velerov1.Backup
+		apiResources []*test.APIResource
+		want         []string
+	}{
+		{
+			name: "include cluster resources=auto includes all CRDs when backing up all namespaces",
```
prydonius: do you mean full-cluster backup? all namespaces is a little confusing since non-namespaced resources are also backed up

pkg/backup/backup_test.go
```diff
@@ -596,6 +596,164 @@ func TestBackupResourceFiltering(t *testing.T) {
...
+				test.VSLs(
+					builder.ForVolumeSnapshotLocation("foo", "vsl-1").Result(),
+				),
+			},
+			want: []string{
+				"resources/volumesnapshotlocations.velero.io/namespaces/foo/vsl-1.json",
+			},
+		},
+		{
+			name: "include cluster resources=true includes all CRDs when backing up all namespaces",
```
prydonius: same thing about full-cluster backup vs. all namespaces here

skriss:
@sseago if you can make the naming updates @prydonius suggested, we can get this merged!

Issue #1094, opened 2018-11-28
Add `velero status` command


ncdc:
Expansion of #770. Add more details, such as "missing a BackupStorageLocation". More details TBD.

nrb:
Per https://github.com/heptio/velero/pull/1849#issuecomment-532370324, this could also report any feature flags enabled on the server.

nrb:
Closing as a duplicate of #979, which has more commentary.

PR #1849, opened 2019-09-05
Add features package

### LINKED ISSUE #1798: Implement feature flag package
per #1759, create a feature flag package with the following:

## High-Level Design

A new command line flag, `--features` will be added to the root `velero` command.

`--features` will accept a comma-separated list of features, such as `--features EnableCSI,Replication`.
Each feature listed will correspond to a key in a map in `pkg/features/features.go` defining whether a feature should be enabled.

Any code implementing the feature would then import the map and look up the key's value.

For the Velero client, a `features` key can be added to the `config.json` file for more convenient client invocations.

## Detailed Design

A new `features` package will be introduced with these basic structs:

```go
type FeatureFlagSet struct {
    flags map[string]bool
}

type Flags interface {
    // Enabled reports whether or not the specified flag is found.
    Enabled(name string) bool

    // Enable adds the specified flags to the list of enabled flags.
    Enable(names ...string)

    // All returns all enabled features
    All() []string
}

// NewFeatureFlagSet initializes and populates a new FeatureFlagSet
func NewFeatureFlagSet(flags ...string) FeatureFlagSet
```

When parsing the `--features` flag, the entire `[]string` will be passed to `NewFeatureFlagSet`.
Additional features can be added with the `Enable` function.
Parsed features will be printed as an `Info` level message on server start up.

No verification of features will be done in order to keep the implementation minimal.

On the client side, `--features` and the `features` key in `config.json` file will be additive, resulting in the union of both.

To disable a feature, the server must be stopped and restarted with a modified `--features` list.
Similarly, the client process must be stopped and restarted without features.

nrb:
This PR creates the `features` package and the necessary plumbing for flags to be enabled at the command line and in the configuration file.

Closes #1798

### DIFF
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
 )
@@ -82,21 +83,30 @@ func SaveConfig(config VeleroConfig) error {
 }
 
 func (c VeleroConfig) Namespace() string {
-	ns, ok := c[ConfigKeyNamespace]
+	val, ok := c[ConfigKeyNamespace]
 	if !ok {
 		return ""
 	}
+	ns, ok := val.(string)
 
-	return ns.(string)
+	if !ok {
+		return ""
+	}
+	return ns
 }
 
 func (c VeleroConfig) Features() []string {
-	features, ok := c[ConfigKeyFeatures]
+	val, ok := c[ConfigKeyFeatures]
 	if !ok {
 		return []string{}
 	}
 
-	return features.([]string)
+	features, ok := val.(string)
+	if !ok {
+		return []string{}
+	}
+
+	return strings.Split(features, ",")
 }
 
 func configFileName() string {
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -87,11 +87,12 @@ func (c VeleroConfig) Namespace() string {
 	if !ok {
 		return ""
 	}
-	ns, ok := val.(string)
 
+	ns, ok := val.(string)
 	if !ok {
 		return ""
 	}
+
 	return ns
 }
 
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 the Velero contributors.
+Copyright 2018, 2019 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -26,17 +26,22 @@ import (
 
 const (
 	ConfigKeyNamespace = "namespace"
+	ConfigKeyFeatures  = "features"
 )
 
-// LoadConfig loads the Velero client configuration file and returns it as a map[string]string. If the
+// VeleroConfig is a map of strings to interface{} for deserializing Velero client config options.
+// The alias is a way to attach type-asserting convenience methods.
+type VeleroConfig map[string]interface{}
+
+// LoadConfig loads the Velero client configuration file and returns it as a VeleroConfig. If the
 // file does not exist, an empty map is returned.
-func LoadConfig() (map[string]string, error) {
+func LoadConfig() (VeleroConfig, error) {
 	fileName := configFileName()
 
 	_, err := os.Stat(fileName)
 	if os.IsNotExist(err) {
 		// If the file isn't there, just return an empty map
-		return map[string]string{}, nil
+		return VeleroConfig{}, nil
 	}
 	if err != nil {
 		// For any other Stat() error, return it
@@ -49,7 +54,7 @@ func LoadConfig() (map[string]string, error) {
 	}
 	defer configFile.Close()
 
-	var config map[string]string
+	var config VeleroConfig
 	if err := json.NewDecoder(configFile).Decode(&config); err != nil {
 		return nil, errors.WithStack(err)
 	}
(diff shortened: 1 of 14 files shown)


nrb:
Looking at how this is currently written, how one might consume it, and based on the design proposal, I think it might make more sense to do an API like this within the `features` package:

```go

var flags sets.String{}

func Enabled(name string) bool {
  flags.Has(name)
}

func Enable(names ...string) {
  flags.Insert(names....)
}

func All() []string {
  return flags.List()
}
```

This then skips requiring initializing our own flag set struct, and callers can simply use `features.Enable("myfeature")` and `features.Enabled("myfeature")`.

I can simplify the enabling code by just doing `features.Enable(config.Features()...)` and `features.Enable(cmdFeatures...)`.

What do others think of that?

nrb:
I'm asking about this slightly different design because passing around a flag set on the client's factory seems a bit unwieldy for something that should be accessible ~anywhere in the code base easily.

I guess the last question I would have is if we're terribly concerned with concurrent access to the flag list, but I'm assuming that is something we don't have to address until later.

pkg/client/config.go
```diff
@@ -76,6 +81,24 @@ func SaveConfig(config map[string]string) error {
 	return json.NewEncoder(configFile).Encode(&config)
 }
 
+func (c VeleroConfig) Namespace() string {
+	ns, ok := c[ConfigKeyNamespace]
+	if !ok {
```
carlisia: Very subjective, but I think this and the next function would read better if the check on the map key was all inlined.

nrb:
@skriss @prydonius Curious about your thoughts on this approach before I proceed further.

prydonius:
> because passing around a flag set on the client's factory seems a bit unwieldy for something that should be accessible ~anywhere in the code base easily.

I would agree, seems like a global here would be a lot easier. As far as concurrent access, seems like we'll only need to set the flags once at the start of the server, so not sure that we really need the mutexes?

pkg/client/factory.go
```diff
@@ -94,6 +96,13 @@ func NewFactory(baseName string) Factory {
 	f.flags.StringVar(&f.kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use to talk to the Kubernetes apiserver. If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration")
 	f.flags.StringVarP(&f.namespace, "namespace", "n", f.namespace, "The namespace in which Velero should operate")
 	f.flags.StringVar(&f.kubecontext, "kubecontext", "", "The context to use to talk to the Kubernetes apiserver. If unset defaults to whatever your current-context is (kubectl config current-context)")
+	// Use a separate StringArray to collect the features because we want to combine the ones in the config file with the ones from the command line, not override them.
+	var cmdFeatures flag.StringArray
+	f.flags.Var(&cmdFeatures, "features", "Comma-separated list of features to enable for this Velero process. Combines with values from $HOME/.config/velero/config.json if present")
```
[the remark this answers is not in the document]
nrb: Yeah, this was from when I was going to pass this down to everything with the factory. Will get it moved.

nrb:
I believe I've addressed all the feedback now. Mutexes were removed since I don't think they provided much value.

I'm thinking it may be useful to add feature information to the `ServerStatusRequest` struct so that the client may request them along with the server's version; what do others think? If that's acceptable, I think I'd rather open a new issue/PR to add it versus continuing with this PR.

skriss:
> I'm thinking it may be useful to add feature information to the ServerStatusRequest struct so that the client may request them along with the server's version; what do others think? If that's acceptable, I think I'd rather open a new issue/PR to add it versus continuing with this PR.

It does seem useful though the only CLI command that exposes the `ServerStatusRequest` right now is `velero version`, which wouldn't really make sense for displaying this information. #1094 would probably be the more sensible path.

pkg/client/config.go
```diff
@@ -76,6 +82,33 @@ func SaveConfig(config map[string]string) error {
 	return json.NewEncoder(configFile).Encode(&config)
 }
 
+func (c VeleroConfig) Namespace() string {
+	val, ok := c[ConfigKeyNamespace]
+	if !ok {
+		return ""
+	}
+	ns, ok := val.(string)
+
```
skriss: nit: rm blank line (L91)

nrb:
The `velero plugins get` command uses it, too, but I agree, it's not quite a good fit for `velero version`. That's part of the reason I'd like to move that to a different issue.

skriss:
> The velero plugins get command uses it, too, but I agree, it's not quite a good fit for velero version. That's part of the reason I'd like to move that to a different issue.

Yep - makes sense to me.

pkg/client/config.go
nrb: IMO not having that line makes it crowded, but I can remove it.

nrb:
Yeah, I've tested this with `backup` commands using the CLI option, CLI option + config file, and both. I've also done the same with the `server` command. I can see the feature getting passed down to plugins too.

There's one thing that _may_ be an issue - now that we have the root `velero` command reading in the config, it can be used by `velero` server. This could lead to a feature being read in unintentionally during local development, though the file won't be present on actual server installs. 

It could also be a local development convenience, so I'm unsure if I want to add any sort of conditional loading logic.

pkg/features/feature_flags.go
```diff
@@ -0,0 +1,70 @@
...
+	if featureFlags.set == nil {
+		NewFeatureFlagSet()
+	}
+
+	featureFlags.set.Insert(names...)
+}
+
+// Disable removes all feature flags in a given slice from the current feature list.
+func Disable(names ...string) {
+
```
prydonius: empty line

pkg/client/config.go
skriss: I think we can't do this here since `val` is used outside the `if...` block?

skriss:
> There's one thing that may be an issue - now that we have the root velero command reading in the config, it can be used by velero server. This could lead to a feature being read in unintentionally during local development, though the file won't be present on actual server installs.

I'm not too worried about this atm, guessing it'll be pretty obvious if that's happening.

Ready to merge once adnan's comment is addressed!

PR #1852, opened 2019-09-06
Add docs to run Velero locally

carlisia:
I thought this would be useful. This is the best place I thought for it but there might be a better place. Documentation for how to run Velero with a local cluster (Kind, etc) could be added to this same page later.

Note: I ran the site locally and was getting the new link on the left navigation with same pages but not other (regardless of the version selected). Not sure if it's a local cache issue or if I missed something.


carlisia:
Nvm about my note, I checked the Netlify preview and it all looks right.

site/docs/v0.11.0/run-locally.md
```diff
@@ -0,0 +1,25 @@
+# Run Velero locally
```
nrb: I'm not sure we want this in v0.11.0 - I'd consider anything prior to 1.0 "legacy" and not really requiring updates.

site/_data/v1-1-0-toc.yml
```diff
@@ -25,6 +25,8 @@ toc:
         url: /gcp-config
       - page: Restic setup
         url: /restic
+      - page: Run locally
```
prydonius: The page should probably be called "run locally in development" since we tell users to build from the master branch

site/docs/master/run-locally.md
```diff
@@ -0,0 +1,25 @@
...
+
+`make local`
+
+b) run the server
+
+To run the server locally, use the path according to the binary you need. Example, if you are on a Mac, this would be where to find the binary you just made: `/_output/bin/darwin/amd64/velero`.
+
+Note: Since you are running locally, chances are you will want to run with the log level at debug, so add that flag.
+
+`./_output/bin/darwin/amd64/velero server --log-level debug`
```
prydonius: isn't an env var needed for aws/azure/google? e.g. `AWS_SHARED_CREDENTIALS_FILE`

carlisia: Is it? I ran w/o it and it "worked". @nrb pls confirm.

site/_data/v1-1-0-toc.yml
carlisia: How about I add that at the top of the page itself, just to avoid having this menu item be so long?

site/docs/master/run-locally.md
nrb: It depends on the provider - each provider we have in-tree uses a credential "chain" that looks up the necessary auth information in env vars, config files, etc. It's totally possibly there was a local profile file that it picked up.

In general, if you look at the Velero `Deployment`, we use env vars to provide the Kubernetes `Secret` file to the server process. Because of that, I use the appropriate env var for the provider, to more closely match how the pod works. eg. `GOOGLE_APPLICATION_CREDENTIALS=bogus_file $VELERO/_output/bin/linux/amd64/velero server --log-level debug` (`$VELERO` is a local env var I use as a shortcut to my Velero checkout)

skriss:
We already have a docs section on this here: https://velero.io/docs/v1.1.0/build-from-source/. Happy to enhance/reorganize that page but let's not duplicate this info.  

Right now the "build from source" page is linked to from the "Customize" section in the TOC, which is a mix of contributor and user information. I'd propose moving the user-facing pages (run in any namespace, hooks, *maybe* the plugins page) into the "Use" section, and changing "Customize" to "Contribute".  Then we can keep developer-focused docs like this in that section.

carlisia:
Note: going to keep making changes only to the `master` tag until it's all reviewed.

Intention of changes: separate info on `build`, `run`, and `code`.

carlisia:
More specific notes:

- removed TOC from the build page since page (as well as other new pages) is now much shorter
- deleted a bunch of link references that were not being used
- batched info on running `make` under the same section
- deleted info on how to run as deployment, since that info is under the Install section

Question: imo, in the build/make section, the `make local` is adding nothing but confusion. Should we scratch that?

site/_data/master-toc.yml
```diff
@@ -45,18 +53,14 @@ toc:
...
-        url: /namespace
-      - page: Extend
-        url: /extend
-      - page: Extend with plugins
-        url: /plugins
-      - page: Extend with hooks
-        url: /hooks
+      - page: Run locally
+        url: /run-locally
+      - page: Add code
```
prydonius: The title of this page is a little confusing to me. I wonder if this should be called "Developer guide" or something else, as the content seems to be general development information.

site/docs/master/build-from-source.md
```diff
@@ -29,73 +22,42 @@ Where `go` is your [import path][4] for Go.
...
+When using `make`, it will place all binaries under `_output/bin/$GOOS/$GOARCH` and will splice version and git commit information in so that `velero version` displays proper output. `velero install` will also use the version information to determine which tagged image to deploy.
 
 ```bash
 make local
 ```
 
-The latter will place the compiled binary under `$PWD/_output/bin/$GOOS/$GOARCH`, and will splice version and git commit information in so that `velero version` displays proper output. `velero install` will also use the version information to determine which tagged image to deploy.
-
-To build the `velero` binary targeting `linux/amd64` within a build container on your local machine, run:
+This builds binaries for both linux and darwin. You will find the binary for darwing here: `_output/bin/darwin/amd64/velero`, and the binary for linux here: `_output/bin/linux/amd64/velero`.
```
prydonius:
```suggestion
This builds binaries for both linux and darwin. You will find the binary for darwin here: `_output/bin/darwin/amd64/velero`, and the binary for linux here: `_output/bin/linux/amd64/velero`.
```

site/docs/master/build-from-source.md
```diff
@@ -105,146 +67,20 @@ Velero's `Makefile` has a convenience target, `all-build`, that builds the follo
...
-
-You can use the `velero install` command to install velero into your cluster, then remove the deployment from the cluster, leaving you
-with all of the required in-cluster resources.
-
-##### Example
-
-This examples assumes you are using an existing cluster in AWS.
-
-Using the `velero` binary that you've built, run `velero install`:
+To build a Velero container image, first set the `$REGISTRY` environment variable. For example, if you want to build the `gcr.io/my-registry/velero:master` image, set `$REGISTRY` to `gcr.io/my-registry`. Optionally, set the `$VERSION` environment variable to change the image tag. Then, run:
```
prydonius: We may want to mention that the default is gcr.io/heptio-images

site/docs/master/run-locally.md
```diff
@@ -0,0 +1,110 @@
...
+* Appropriate RBAC permissions in the cluster
+  * Read access for all data from the source cluster and namespaces
+  * Write access to the target cluster and namespaces
+* Cloud provider credentials
+  * Read/write access to volumes
+  * Read/write access to object storage for backup data
+* A [BackupStorageLocation][20] object definition for the Velero server
+* (Optional) A [VolumeSnapshotLocation][21] object definition for the Velero server, to take PV snapshots
+
+### 1. Create a cluster
```
prydonius: This makes it sound like you need a cluster on AWS to run velero locally, perhaps we should just link to the Install section which already goes through various cluster options on different platforms (including local) instead of pointing out specific options here?

site/docs/master/run-locally.md
```diff
@@ -0,0 +1,110 @@
...
+
+#### 2. Set enviroment variables
+
+Set the appropriate environment variable for your cloud provider:
+
+AWS: [AWS_SHARED_CREDENTIALS_FILE][15]
+
+GCP: [GOOGLE_APPLICATION_CREDENTIALS][16]
+
+Azure:
```
prydonius: on Azure you can actually set `AZURE_CREDENTIALS_FILE` to the file created at the bottom of this section: https://velero.io/docs/v1.1.0/azure-config/#create-service-principal

site/docs/master/run-locally.md
```diff
@@ -0,0 +1,110 @@
...
+kubectl --namespace velero delete deployment velero
+
+rm fake-credentials-file
+```
+
+##### Scale deployyment down to zero
+
+Scale the Velero deployment down to 0 so it is not simultaneously being run on the remote cluster and potentially causing things to get out of sync:
+
+`kubectl scale --replicas=0 deployment velero -n velero`
```
prydonius: In the previous section, the deployment is deleted. IMO we should just describe scaling it to 0, that way it's easily recoverable.

carlisia:
Note: I don't want to add to this PR, but these two issues would be additions to this new `Contribute` section:

- https://github.com/heptio/velero/issues/1387
- https://github.com/heptio/velero/issues/1573

site/_data/v1-1-0-toc.yml
skriss: I'd prefer to not include this in the `Install` section, given that it's mainly a development topic, and it's in the `Contribute` section already.

site/docs/master/README.md
```diff
@@ -36,19 +36,7 @@ If you encounter issues, review the [troubleshooting docs][30], [file an issue][
...
-### Before you start
-
-* Please familiarize yourself with the [Code of Conduct][8] before contributing.
-* See [CONTRIBUTING.md][5] for instructions on the developer certificate of origin that we require.
-* Read how [we're using ZenHub][26] for project and roadmap planning
-
-### Pull requests
-
-* We welcome pull requests. Feel free to dig through the [issues][4] and jump in.
+If you are ready to jump in and test, add code, or help with documentation, follow the instructions on our [Start contributing](https://velero.io/docs/master/start-contributing/) documentation for guidance on how to setup Velero for development.
```
skriss: We probably want to make this change to the root repo `README.md` as well since this is just a copy of it

README.md
```diff
@@ -70,11 +56,10 @@ See [the list of releases][6] to find out about feature changes.
 
 [24]: https://groups.google.com/forum/#!forum/projectvelero
 [25]: https://kubernetes.slack.com/messages/velero
-[26]: https://velero.io/docs/zenhub
 
-[28]: https://velero.io/docs/install-overview
-[29]: https://velero.io/docs/
-[30]: https://velero.io/docs/troubleshooting
+[28]: install-overview.md
```
skriss: Need to revert these link changes (28-100) - they still need to point to `velero.io/...`

site/_data/v1-1-0-toc.yml
skriss: This is the last thing for me, then LGTM!

skriss:
FYI https://github.com/heptio/velero/pull/1852#discussion_r324790494

site/_data/v1-1-0-toc.yml
carlisia: I thought I had removed it. Removing!

PR #1854, opened 2019-09-06
record restic restore progress in PodVolumeRestore

prydonius:
depends on #1821

### DIFF
--- a/pkg/restic/exec_commands.go
+++ b/pkg/restic/exec_commands.go
@@ -20,6 +20,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"os"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -29,6 +31,7 @@ import (
 	"github.com/heptio/velero/pkg/util/exec"
 )
 
+const restoreProgressCheckInterval = 10 * time.Second
 const backupProgressCheckInterval = 10 * time.Second
 
 type backupStatusLine struct {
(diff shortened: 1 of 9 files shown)


prydonius:
Hmm, I'm not sure why the travis push job is running, but it's probably because I pushed the branch for #1821 to origin to have this be based off of it. @nrb is there a way to get the push job to only run on master?

pkg/restic/exec_commands.go
```diff
@@ -176,3 +180,102 @@ func getSummaryLine(b []byte) ([]byte, error) {
...
+	}
+
+	updateFunc(velerov1api.PodVolumeOperationProgress{
+		TotalBytes: snapshotSize,
+	})
+
+	// create a channel to signal when to end the goroutine scanning for progress
+	// updates
+	quit := make(chan struct{})
+	var wg sync.WaitGroup
```
skriss: can remove the sync.WaitGroup similarly to the backup progress PR.

nrb:
> Hmm, I'm not sure why the travis push job is running, but it's probably because I pushed the branch for #1821 to origin to have this be based off of it. @nrb is there a way to get the push job to only run on master?

I think it's due to https://github.com/heptio/velero/blob/master/.travis.yml#L31, but I'm not sure we can turn it off because, based on my understanding, it won't work with tags :(

prydonius:
Ah okay, though it looks like it didn't run for the last commit on this?

prydonius:
ty, will rebase this now that #1821 is merged!

prydonius:
Can't see why DCO would fail here

carlisia:
Weird. Maybe squashing them will make it happy.

skriss:
Not sure, seems to be complaining about past commits that are already in `master`..looks like I can manually override it to pass for now.

pkg/restic/exec_commands.go
```diff
@@ -20,17 +20,22 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"sync"
```
skriss: looks like this isn't used, so need to remove

prydonius: good catch, think it was left over from a rebase conflict

prydonius:
Added a changelog and ran `make update` - PTAL!

PR #1859, opened 2019-09-09
velero install: wait for restic daemonset to be ready

### LINKED ISSUE #1716: Check restic daemonset for readiness on install
Currently, `velero install --wait` attempts to check the Deployment with the velero server for readiness, but does not look at the restic daemonset.

For completeness, it should also check to be sure the restic daemonset is in a usable state.

skriss:

Closes #1716 

This is a copy-paste with modifications from the code that waits for the deployment to be ready. I looked again at the API docs for daemonsets, and I believe the readiness check I have makes the most sense -- comparing the number of expected scheduled pods to the number of "available" pods, which means pods on proper nodes that have been ready for `spec.minReadySeconds`. It seems like the Octant logic could falsely report "ready" if not all pods have been started yet.

### DIFF
--- a/changelogs/unreleased/1859-skriss
+++ b/changelogs/unreleased/1859-skriss
@@ -1,0 +1,1 @@
+velero install: if `--use-restic` and `--wait` are specified, wait up to a minute for restic daemonset to be ready
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -238,9 +238,16 @@ func (o *InstallOptions) Run(c *cobra.Command, f client.Factory) error {
 	}
 
 	if o.Wait {
-		fmt.Println("Waiting for Velero to be ready.")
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/1859-skriss
```diff
@@ -0,0 +1 @@
+velero install: if `--use-restic` and `--wait` are specified, wait for restic daemonset to be ready
```
nrb: Do we want to mention the timeout here?

I think we're up to a total of 2 minutes between the deployment and daemonset waits.

skriss: sure I can add that.

I didn't bother doing anything more complicated like waiting for both concurrently, as it would require more significant refactoring and it didn't seem super-important to be, but open to others' input on it.

nrb: Yeah, I agree that separate checks with their own timeouts make sense. I'm not sure there's tons of benefit to combining the checks.

PR #1861, opened 2019-09-10
Add a check to avoid nil pointer

carlisia:


skriss:
This looks fine - we should probably add the same to the `restore_item_action_client.go`.

pkg/plugin/framework/restore_item_action_client.go
```diff
@@ -58,6 +58,10 @@ func (c *RestoreItemActionGRPCClient) AppliesTo() (velero.ResourceSelector, erro
 		return velero.ResourceSelector{}, fromGRPCError(err)
 	}
 
+	if res.ResourceSelector == nil {
+		return velero.ResourceSelector{}, fromGRPCError(err)
```
skriss: `return velero.ResourceSelector{}, nil`

carlisia: Argh.

PR #1864, opened 2019-09-10
Change restic prune default interval to 7d and make user-configurable

### LINKED ISSUE #1541: Restic repository pruning
Decide how to handle repository pruning (a maintenance operation that removes unused data blocks from the repository, which requires an exclusive repository lock and can be very slow for large repositories)

skriss:
Closes #1541 

This PR changes the default interval for running `restic prune` to 7d, and makes the default value user-configurable via flags on `velero server` and `velero install`.

I'm planning to include some instructions in the release notes on how to change existing repositories' maintenance intervals as well.

### DIFF
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


pkg/cmd/cli/install/install.go
```diff
@@ -291,5 +295,9 @@ func (o *InstallOptions) Validate(c *cobra.Command, args []string, f client.Fact
 		return errors.New("Cannot use both --secret-file and --no-secret")
 	}
 
+	if o.DefaultResticMaintenanceFrequency < 0 {
```
carlisia: Should this not be `<= 0`?

skriss: For `velero install` I didn't give this flag a default value, so the default is 0 (meaning don't explicitly specify this flag when creating the Deployment). So 0 is valid here.

Issue #1675, opened 2019-07-17
Investigate CRD structural schema


nrb:
[Structural schema](https://kubernetes.io/blog/2019/06/20/crd-structural-schema/) support is coming to the v1 release of CRDs. We need to evaluate whether this is something that's beneficial to Velero, as well as if there's any required effort to keep Velero forward-compatible.

nrb:
A summary of the blog post:

* The `apiextensions.k8s.io/v1` group will _require_ a structural schema.
  * There's a way to disable this, but it's not desirable.
* The `apiextensions.k8s.io/v1beta1` group _may_ include a structural schema
* Unknown field pruning support requires a structural schema. Pruning removes any fields not defined in the schema.

To enable a structural schema, the [`CustomResourceDefintion.Spec.Validation`](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensioans/v1beta1/types.go#L53) field must be populated. It's a set of [OpenAPI properties](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1beta1/types_jsonschema.go#L20)

I believe this is both desirable and required for Velero, and will probably take a design doc for more complete research.

cblecker:
xref #1267 (kubebuilder has autogeneration of openapi validation: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html)

nrb:
@cblecker Good point! We need to do a spike to see how much effort replacing our own generic controller structs with the kubebuilder ones will be.

cblecker:
Spent a little bit poking at it, and it shouldn't be too complicated.. the structs are well formed using the apimachinery packages.. Adding the generation tags for the openapi data should be pretty straight forward based on the API definitions.

Then it would just be swapping out the CRD rendering bits in `pkg/install` for the kubebuilder parser (https://godoc.org/sigs.k8s.io/controller-tools/pkg/crd) so that the fully rendered object with validation spec gets installed.

nrb:
Thanks for doing that digging!

I don't think this will make the cut for v1.1, but we should definitely consider it for v1.2, even if just a spike. If someone wants to take a stab at a PR before our team gets to it, that's certainly welcome!

prydonius:
[...]

PR #1875, opened 2019-09-12
proposal for generating Velero CRDs with structural schema

prydonius:
ref #1675

### DIFF
--- a/design/generating-velero-crds-with-structural-schema.md
+++ b/design/generating-velero-crds-with-structural-schema.md
@@ -1,0 +1,164 @@
+# Generating Velero CRDs with structural schema support
+
+As the apiextensions.k8s.io API moves to GA, structural schema in Custom Resource Definitions (CRDs) will become required.
+
+This document proposes updating the CRD generation logic as part of `velero install` to include structural schema for each Velero CRD.
+
+## Goals
+
+- Enable structural schema and validation for Velero Custom Resources.
+
+## Non Goals
+
+- Update Velero codebase to use Kubebuilder for controller/code generation.
+- Solve for keeping CRDs in the Velero Helm chart up-to-date.
+
+## Background
+
+Currently, Velero CRDs created by the `velero install` command do not contain any structural schema.
+The CRD is simply [generated at runtime](https://github.com/heptio/velero/blob/8b0cf3855c2b8aa631cf22e63da0955f7b1d06a8/pkg/install/crd.go#L39) using the name and plurals from the [`velerov1api.CustomResources()`](https://github.com/heptio/velero/blob/8b0cf3855c2b8aa631cf22e63da0955f7b1d06a8/pkg/apis/velero/v1/register.go#L60) info.
+
+Updating the info returned by that method would be one way to add support for structural schema when generating the CRDs, but this would require manually describing the schema and would duplicate information from the API structs (e.g. comments describing a field).
+
+Instead, the [controller-tools](https://github.com/kubernetes-sigs/controller-tools) project from Kubebuilder provides tooling for generating CRD manifests (YAML) from the Velero API types.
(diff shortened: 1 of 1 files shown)


design/generating-velero-crds-with-structural-schema.md
```diff
@@ -0,0 +1,164 @@
...
+
+`velero install` will then unmarshal the binary data as `unstructured.Unstructured` types and append them to the [resources list](https://github.com/heptio/velero/blob/8b0cf3855c2b8aa631cf22e63da0955f7b1d06a8/pkg/install/resources.go#L217) in place of the existing CRD generation.
+
+## Alternatives Considered
+
+Instead of generating and bundling CRD manifests, it could be possible to instead embed the `pkg/apis` package in the Velero binary.
+With this, _controller-tools_ could be run at runtime during `velero install` to generate the CRD manifests.
+However, this would require including _controller-tools_ as a dependency in the project, which might not be desirable as it is a developer tool.
+
+Another option, to avoid embedding static files in the binary, would be to generate the CRD manifest as one YAML file in CI and upload it as a release artifact (e.g. using GitHub releases).
```
nrb: Agreed that this is far less desirable. I assume if we did this option, we would very quickly get requests to provide a way to get the manifest(s) for offline/firewalled clusters.

design/generating-velero-crds-with-structural-schema.md
```diff
@@ -0,0 +1,164 @@
...
+|                                 | NamespaceMapping        | make optional                                               |
+|                                 | LabelSelector           | make optional                                               |
+|                                 | RestorePVs              | make optional                                               |
+|                                 | IncludeClusterResources | make optional                                               |
+| ScheduleSpec                    | Template                | none (required)                                             |
+|                                 | Schedule                | none (required)                                             |
+| VolumeSnapshotLocationSpec      | Provider                | none (required)                                             |
+|                                 | Config                  | make optional                                               |
+
+### Build-time generation of CRD manifests
```
nrb: I haven't yet looked at the POC branch, but is it a safe assumption that `make local` should continue to work so long as `controller-tools` are installed on the host machine? I know `make local` is useful when doing development as you don't have to wait for the image to be pulled.

design/generating-velero-crds-with-structural-schema.md
```diff
@@ -0,0 +1,164 @@
...
+
+_controller-tools_ works by reading the Go files that contain the API type definitions.
+It uses a combination of the struct fields, types, tags and comments to build the OpenAPIv3 schema for the CRDs. The tooling makes some assumptions based on conventions followed in upstream Kubernetes and the ecosystem, which involves some changes to the Velero API type definitions, especially around optional fields.
+
+In order for _controller-tools_ to read the Go files containing Velero API type defintiions, the CRDs need to be generated at build time, as these files are not available at runtime (i.e. the Go files are not accessible by the compiled binary).
+These generated CRD manifests (YAML) will then need to be available to the `pkg/install` package for it to include when installing Velero resources.
+
+## Detailed Design
+
+### Changes to Velero API type definitions
```
nrb: Thanks for this detailed table! This alone is hugely valuable.

design/generating-velero-crds-with-structural-schema.md
prydonius: Since the generated Go pkg that contains the static manifests is checked in, `make local` will just work. If you're changing API types, you'll need to run `make update` which then requires `controller-tools` to be installed (though for the make task, it'll get run in the builder image which will have it). IMO this fits with the current workflow pretty well, as the client code is also generated and checked in, and if you're touching API types you need to run `make update` to update the clients.

prydonius:
@nrb @carlisia PTAL!

PR #1879, opened 2019-09-16
GCP: support specifying Cloud KMS key name for backup storage locations

skriss:
Adds support for explicitly providing a customer-managed cloud KMS key name to use for encrypting backups in a backup storage location. For more details, see https://cloud.google.com/storage/docs/encryption/using-customer-managed-keys.

This brings GCP server-side encryption support inline with AWS support.


skriss:
alright, testing LGTM here.

PR #1883, opened 2019-09-17
move BSL validation from server startup to backup processing

skriss:

xref https://github.com/heptio/velero/issues/1853#issuecomment-532296732

This moves the `BackupStorageLocation` validation (which verifies that the bucket does not contain any unexpected directories/files) from server startup to backup processing. The benefits of this are:
- A BSL is validated each time a backup is run against it, rather than just once at server startup
- BSLs that are created after the server has started up still get validated
- A user *could* provide dummy configuration to `velero install` for the BSL if the correct configuration is not yet known, and the server would no longer crash-loop on startup.

On the flip side, some potential drawbacks:
- A BSL that's invalid at server startup won't result in a crash-loop; instead, the next backup targeting that BSL will fail validation. The feedback to the user occurs later.

Does this seem like a good change? cc @betta1

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -55,7 +55,6 @@ import (
 	clientset "github.com/heptio/velero/pkg/generated/clientset/versioned"
 	informers "github.com/heptio/velero/pkg/generated/informers/externalversions"
 	"github.com/heptio/velero/pkg/metrics"
-	"github.com/heptio/velero/pkg/persistence"
 	"github.com/heptio/velero/pkg/plugin/clientmgmt"
 	"github.com/heptio/velero/pkg/podexec"
 	"github.com/heptio/velero/pkg/restic"
@@ -313,10 +312,6 @@ func (s *server) run() error {
 		return err
 	}
 
-	if err := s.validateBackupStorageLocations(); err != nil {
-		return err
-	}
-
 	if _, err := s.veleroClient.VeleroV1().BackupStorageLocations(s.namespace).Get(s.config.defaultBackupLocation, metav1.GetOptions{}); err != nil {
 		s.logger.WithError(errors.WithStack(err)).
 			Warnf("A backup storage location named %s has been specified for the server to use by default, but no corresponding backup storage location exists. Backups with a location not matching the default will need to explicitly specify an existing location", s.config.defaultBackupLocation)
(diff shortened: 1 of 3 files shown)


nrb:
I like that this makes Velero more flexible for use in a larger system, but I think the downside of communicating an invalid BSL later via a failed backup is pretty big.

Could we perhaps validate on server startup and log a warning as well? I know it's repetitive, but I really don't want to make the current direct usage scenario worse in terms of user experience.

betta1:
LGTM, thanks Steve. I'm also thinking of a scenario where crashloops due to an invalid BSL could potentially block backups to other valid BSLs.  Let's say I have an invalid BSL and some backups to other valid BSLs are on the queue to be processed, if the server starts up and crashloops due to this invalid BSL, could we have a case where backups to the other valid BSLs are stuck in pending state?

Does it also make sense for BSLs to have a status field that's updated periodically e.g during backup sync, this might be useful  for large systems that have many BSLs since it'll be easier to quickly know which BSLs are valid or in failed state (e.g if the backup server is down).

skriss:
> I think the downside of communicating an invalid BSL later via a failed backup is pretty big.

You're saying the UX of having the server pod crashloop on startup is significantly better than getting a backup with FailedValidation?

We could definitely look at keeping the validation at startup, but I do think we still want to add running validation during backups, or at least periodically, since the server startup validation totally ignores any BSLs created after startup.

> LGTM, thanks Steve. I'm also thinking of a scenario where crashloops due to an invalid BSL could potentially block backups to other valid BSLs. Let's say I have an invalid BSL and some backups to other valid BSLs are on the queue to be processed, if the server starts up and crashloops due to this invalid BSL, could we have a case where backups to the other valid BSLs are stuck in pending state?

Yes - in the current state (i.e. in v1.1), this would absolutely happen - any invalid BSL would cause crashlooping and prevent backups to any other BSL from proceeding. I agree that this is a drawback of the current (v1.1) implementation.

> Does it also make sense for BSLs to have a status field that's updated periodically e.g during backup sync, this might be useful for large systems that have many BSLs since it'll be easier to quickly know which BSLs are valid or in failed state (e.g if the backup server is down).

This is another option we could look at - would likely involve adding a `BackupStorageLocationController`. I was trying to avoid this as it's a bigger change, but definitely a possibility.

nrb:
> You're saying the UX of having the server pod crashloop on startup is significantly better than getting a backup with FailedValidation?

Yeah, my reasoning is that the user finds out that the BSL's invalid much sooner with the crash loop. If there's no crash loop, the user could set up Velero and only start making backups days later.

I absolutely agree we should be validating post-startup for the reasons you mention, though. And I defintely agree that we shouldn't block all backups if 1 BSL's bad while the others are fine.

prydonius:
> > Does it also make sense for BSLs to have a status field that's updated periodically e.g during backup sync, this might be useful for large systems that have many BSLs since it'll be easier to quickly know which BSLs are valid or in failed state (e.g if the backup server is down).
>
> This is another option we could look at - would likely involve adding a BackupStorageLocationController. I was trying to avoid this as it's a bigger change, but definitely a possibility.

This seems like a more idiomatic approach. We could validate BSLs when created/modified and mark them as "Ready" in a `status` field. Then, we only consider "Ready" BSLs when taking a backup.

skriss:
Yeah, this was the original idea with BSLs -- you'll note we already have a `status.phase` field, we're just not using it.

Let me step back on this a little bit and see if I can propose something more robust.

skriss:
Closing this out for now as I think we want to think about the full workflow - `velero install`, `velero plugin add`, BSL/VSL definition & validation, etc.

Issue #1740, opened 2019-08-08
Add --plugins flag to `velero install` to allow plugins to be included at install time


skriss:
`velero install` should allow 1+ plugins to be installed during the main install process. This would allow out-of-tree cloud providers and other plugins to be installed seamlessly, without requiring a restart.

skriss:
xref #1537

nrb:
IMO, the values for `--plugin` should be container image URLs that would be passed into `velero plugin add` today.

tlvenn:
This would be great as it would open the path to support this with the helm chart and to have a declarative / terraform compatible way of managing those plugins.

tlvenn:
But actually it 2 distinct paths given the helm chart is not invoking the cli to install velero... I should open a separate issue to track this.

carlisia:
An issue for the Helm change was created: https://github.com/heptio/velero/issues/1819

nrb:
Per discussion with the team, we're going to, for the time being, make installation a 2 step process - running `velero install` to generate the necessary objects in cluster, and `velero plugin add` to add any relevant object or snapshot plugins.

This is mostly temporary to get things split up per #1870. Longer term, we want to revisit how plugins are handled holistically so we have a clearer answer for how things like secrets may be handled, and the possibility of making plugins run as sidecars instead of init containers.

skriss:
What's the reasoning for not adding the flag to `velero install`? IMHO it's simpler for the user to include it as part of the install command, and since it's required for velero to be functional, would be better to make it an atomic part of the install.

I'm fine revisiting things down the road, but I'm failing to see how making it a separate step right now is better than this.

nrb:
>What's the reasoning for not adding the flag to velero install?

This came up as confusing during one of our discussions on the implementations. A 2 step process was proposed, one for getting the resources into the cluster and one for installing the plugins. That's possible today.

I'm fine with adding the `--plugins` flag if the rest of the team is.

carlisia:
It's fine by me.

skriss:
To be clear, I definitely think there's room for improvement here in the install & setup flow, the relationships between plugin images, names, and BSLs/VSLs, what we validate at server startup, etc. that we should take a holistic look at - but I propose doing that separately.

PR #1884, opened 2019-09-17
Add docs for getting out-of-tree object storage/volume snapshotter plugins

nrb:
This is in draft until we reach an accepted format for conveying this information. I'll then update the remaining 2 provider docs.

### DIFF
--- a/site/docs/master/aws-config.md
+++ b/site/docs/master/aws-config.md
@@ -148,11 +148,12 @@ Install Velero, including all prerequisites, into the cluster and start the depl
 
 ```bash
 velero install \
-    --provider aws \
+    --provider velero.io/aws \
     --bucket $BUCKET \
     --secret-file ./credentials-velero \
     --backup-location-config region=$REGION \
     --snapshot-location-config region=$REGION
+velero plugin add gcr.io/heptio-images/velero-aws-plugin
 ```
 
 Additionally, you can specify `--use-restic` to enable restic support, and `--wait` to wait for the deployment to be ready.
--- a/site/docs/master/install-overview.md
+++ b/site/docs/master/install-overview.md
@@ -8,7 +8,7 @@ You can also use Velero's integration with restic, which requires additional set
 
 ## Cloud provider
 
-The Velero client includes an `install` command to specify the settings for each supported cloud provider. You can install Velero for the included cloud providers using the following command:
+The Velero client includes an `install` command to specify the settings for each supported cloud provider. You can install Velero using the following command:
 
 ```bash
 velero install \
@@ -26,6 +26,8 @@ velero install \
 
 When using node-based IAM policies, `--secret-file` is not required, but `--no-secret` is required for confirmation.
 
+This command will configure a `BackupStorageLocation` and `VolumeSnapshotLocation` with the `provider` fields set to the value of the `--provider` flag.  The `--provider` value must match the name of the relevant plugin.
+
 For provider-specific instructions, see:
 
 * [Run Velero on AWS][0]
@@ -38,6 +40,10 @@ When using restic on a storage provider that doesn't currently have Velero suppo
 To see the YAML applied by the `velero install` command, use the `--dry-run -o yaml` arguments.
 
 For more complex installation needs, use either the generated YAML, or the Helm chart.
+
+### Installing object storage or volume snapshotter plugins
+
+`velero install` _will not_ install any object storage or volume snapshotter plugins. In order to install these, you must run `velero plugin add <URL of container here>` after running `velero install`.
 
 ## On-premises
 


site/docs/master/install-overview.md
```diff
@@ -26,6 +26,8 @@ velero install \
 
 When using node-based IAM policies, `--secret-file` is not required, but `--no-secret` is required for confirmation.
 
+This command will configure a `BackupStorageLocation` and `VolumeSnapshotLocation` with the `provider` fields set to the value of the `--provider` flag.  The `--provider` value must match the name of the relevant plugin.
```
prydonius: In the AWS example, `velero.io/aws` is the provider and the image name is `gcr.io/heptio-images/velero-aws-plugin`. What does it mean here for the provider to match the plugin name?

site/docs/master/aws-config.md
```diff
@@ -148,11 +148,12 @@ Install Velero, including all prerequisites, into the cluster and start the depl
 
 ```bash
 velero install \
-    --provider aws \
+    --provider velero.io/aws \
```
skriss: Does this change imply that users are going to be required to change their existing BSL provider names?

site/docs/master/install-overview.md
```diff
@@ -39,6 +41,10 @@ To see the YAML applied by the `velero install` command, use the `--dry-run -o y
 
 For more complex installation needs, use either the generated YAML, or the Helm chart.
 
+### Installing object storage or volume snapshotter plugins
+
+`velero install` _will not_ install any object storage or volume snapshotter plugins. In order to install these, you must run `velero plugin add <URL of container here>` after running `velero install`.
+
```
carlisia: I would say something like: "As of v1.12", to emphasize that this is now new.

I would not be against adding the `plugins` flag to the `install` cmd. That would make the interaction more user friendly since if it wasn't included we would warn immediately that it needs to be added. 

To note: if we don't add this flag, then logging would need to be added to warn the user during creation of backups in case the `plugin add` was not run.

site/docs/master/install-overview.md
nrb: The `provider` must match the first argument passed in at https://github.com/heptio/velero/blob/master/pkg/cmd/server/plugin/plugin.go#L41.

afaik this is only exposed to the user via the plugin's documentation. It has no bearing on the image name, which I think isn't great.

site/docs/master/aws-config.md
nrb: Yes, they'll have to change if we want to make our plugins play by the same rules as everyone else.

IMO we should provide a translation script or command for upgrading. It shouldn't be all that involved to fetch each BSL and update the `provider` field.

site/docs/master/install-overview.md
nrb: What would we log? I can run some experiments, but it should error because the plugin won't be found for either the object storage or volume snapshotter.

carlisia: Sorry, yes, I meant error out, to warn the user. I'm not crazy about this flow, in other words, telling the user there was an error w/o first preventing them from triggering the error. Some ppl don't read the docs.

site/docs/master/install-overview.md
nrb: I think this provider/plugin name and image name disconnect is something we should think about more deeply, though I don't know that we can get it in for v1.2. How can we make this more easily discoverable at the command line? Is that possible?

site/docs/master/install-overview.md
nrb: `velero install --bucket $BUCKET --provider velero.io/invalid --secret-file bogus_file --wait` will error out due to the behavior that #1883 seeks to address

```
An error occurred: some backup storage locations are invalid: error getting backup store for location "default": unable to locate ObjectStore plugin named velero.io/invalid
```

So the pod will crashloop until the ObjectStore plugin is added.

I'm not finding any validation for VSLs, so I don't think there would be an error thrown for finding a VolumeSnapshotter plugin until a backup is taken that includes a volume.

carlisia: Ok. Would it be worthwhile adding to that error msg something like "please make sure you run `plugin add` to add a plugin for your provider"?

nrb: Yeah, that seems reasonable to me, even if we add the `--plugins` flag. People could forget to add that at install time, and need to add it after the fact, so I think providing a hint when we can't find a plugin is good no matter what.

carlisia: If we had the `--plugins` flag, would it not be required? I'd think it should.

nrb: True, it should be required. Still, I don't think adding the hint on missing plugins hurts. Something like this:

```diff
diff --git i/pkg/plugin/clientmgmt/registry.go w/pkg/plugin/clientmgmt/registry.go
index 620d08ad..bdf506f0 100644
--- i/pkg/plugin/clientmgmt/registry.go
+++ w/pkg/plugin/clientmgmt/registry.go
@@ -228,7 +228,7 @@ func newPluginNotFoundError(kind framework.PluginKind, name string) *pluginNotFo
 }
 
 func (e *pluginNotFoundError) Error() string {
-	return fmt.Sprintf("unable to locate %v plugin named %s", e.kind, e.name)
+	return fmt.Sprintf("unable to locate %v plugin named %s. Use `velero plugin add <IMAGE_LOCATION>` to add a plugin", e.kind, e.name)
 }
 
 type duplicatePluginRegistrationError struct {
```

site/docs/master/install-overview.md
prydonius: Yeah this is going to be pretty confusing, I'll need to think about this more. Image name doesn't make sense as this will be different if you need to mirror the image in different registries. What if the plugin name had to be provided to `velero plugin add <name> <image>`, and this is what's used to register the plugin, instead of the plugin registering itself? I guess this would be a backwards incompatible change, though.

Something we can think about for plugins v2, but for v1 we might just need to rely on documentation.

site/docs/master/install-overview.md
prydonius: I thought we didn't like the idea of the `--plugins` flag because it was now a little confusing whether that related to the BSL or the VSL or both?

carlisia: Ah yes. These tradeoffs are a nightmare. Today I was also thinking about this new issue/PR: https://github.com/heptio/velero/pull/1883 and maybe what we should do is NOT add the ``plugins` flag, document, and let the backup validation fail if the plugins wasn't added.

skriss: Let's move comments on `velero install --plugins` to #1740 - see https://github.com/heptio/velero/issues/1740#issuecomment-533180281

site/docs/master/install-overview.md
nrb: >Image name doesn't make sense as this will be different if you need to mirror the image in different registries.

I'm not sure I understand this statement. `velero plugin add <IMAGE NAME>` basically inserts the container image address for an init container. How does that interact with mirroring in different registries? I would assume it's just the name of the registry your cluster is using at the time.

skriss:
I do think as part of the updates here, we can add some more clarity to `install-overview.md` on exactly what `velero install` does w.r.t. setting up BSLs/VSLs:
- always sets up a BSL, for the provider specified
- if --use-snapshots is true, also sets up a VSL for that provider
- if you want >1 BSL or VSL, you'll need to use `velero backup-location create` and/or `velero snapshot-location create` to define the additional ones
- if your VSL is for a different provider than your BSL, you'll need to define it separately using `velero snapshot-location create`

carlisia:
> I do think as part of the updates here, we can add some more clarity to install-overview.md on exactly what velero install does w.r.t. setting up BSLs/VSLs:

I think all the doc changes should go together. See the TODOs on this PR:

https://github.com/heptio/velero/pull/1890

They could go on a separate PR, or I can add them to ^ that PR.

nrb:
I think this is where I'm confused:

Item #1 is `Remove provider documentation`. However, we also need to update the provider's documentation to include the new steps, regardless of whether those steps are this or using `--plugins` argument. Those 2 changes can't happen in the same PR (how do you update something and delete it?). The content needs to survive and be updated in the new repos, and I think it makes sense to get most of the updates to the provider content done in this repo before moving them.

carlisia:
I think the confusion is in the definitions.

- Provider documentation: The specific provider docs we have now, on how to install with a provider [aws, azure, gcp]. This will be in the spirit of examples. There will move to their respective repos.

- Update velero docs: this will be the future documentation, something generic on how to install any plugin. It will live in the Velero repo. Must have this: https://github.com/heptio/velero/pull/1884#issuecomment-533230955. 

- Add links to external plugins to the Velero docs. We agreed to list them as sub-menus.

skriss:
I think we can close this one, right?

nrb:
Yeah, we can close this out as we're planning on adding the `--plugins` flag, and making it required.

PR #1885, opened 2019-09-18
Generate CRDs with structural schema

prydonius:
This PR introduces CRD generation using the _controller-tools_ tooling. The CRDs are generated to a manifests folder, and a custom code generator is used to embed the manifests in a Go pkg and make them retrievable as API objects.


skriss:
@prydonius are you looking for input on this yet? I haven't looked yet but can if you're wanting feedback

prydonius:
Thanks @skriss, would be great if you could take a look at the code generation, though so far I'm still testing the schema validation with different clusters, and need to finish making the changes to the optional API fields and push that up.

pkg/generated/crds/crds_generate.go
```diff
@@ -0,0 +1,117 @@
+// +build ignore
```
skriss: This LGTM - only thing I'm wondering is if we want to relocate this file to be outside of `pkg/generated`? It could make sense to go in a subdir of `hack`.

prydonius: Yeah, I was thinking about that too - I think that makes sense, will move it there.

prydonius:
I'm going to make the API type changes in a different PR as it's just going to be easier to review. The current CRDs are broken due to non-optional status fields, so I've reverted the `velero install` change to use the non-structural CRDs it was using before. This is now ready for review!

prydonius:
@carlisia ptal if you have a chance!

pkg/generated/crds/crds_generate.go
carlisia: I think this was moved to `hack/crd-gen/main.go` and this needs to be deleted?

prydonius: yep 😅thanks for the catch!

hack/crd-gen/main.go
```diff
@@ -0,0 +1,134 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This code embeds the CRD manifests in pkg/generated/crds/manifests in
+// pkg/generated/crds/crds.go.
+
```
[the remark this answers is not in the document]
prydonius: Hmm, most other files have an empty line between the comment and package name (unless documenting what the package is for), so probably best to keep that consistent?

prydonius:
@skriss yep this can be merged, and I can rebase and update #1898 on top of this

PR #1890, opened 2019-09-19
Refactor out providers

### LINKED ISSUE #1873: Clean out Velero of cloud provider code
This is to track the removal of the provider code. Other issues will track the implementation of each as a separate project.

carlisia:
Closes #1873.

TODOs for this PR:
- [x] Remove provider documentation
- [x] Add links to external repos for provider documentation
- [ ] Ensure we have clean documentation for how to install any external plugins
- [x] Resolve the issue of restic needing Azure variables/configuration info
- [x] Clean up dependencies

Not planned TODO for this PR:
- 🚫 Remove AWS dependency for restic config

Before merging this PR:
- [x] get the new repos setup
- [x] get image builds for the new repos set up
- [ ] make `install plugins` required


carlisia:
I'd appreciate some pointers on how to deal with this: https://github.com/heptio/velero/pull/1890/files#diff-3ef2f5571726560b023981cb4b55334f

There's a whole cascade of Azure code that resides in the extracted `cloudprovider` package and comes out from that call to `getStorageAccountKey(config)`.

skriss:
I'm looking at the Azure+restic code, but so far not sure how we're going to avoid having it.

carlisia:
@skriss is it looking to you like the solution to the azure situation would be to duplicate the provider code in this repo? That would be ok with me. Maybe we can find some pattern to improve this when/if we do an overhaul of the plugins.

skriss:
> @skriss is it looking to you like the solution to the azure situation would be to duplicate the provider code in this repo? That would be ok with me. Maybe we can find some pattern to improve this when/if we do an overhaul of the plugins.

Yeah, this would be the most straightforward approach - keep a copy of `pkg/cloudprovider/azure.GetResticEnvVars` and any code it depends on.

Alternately, we could consider requiring users of Azure + restic to provide a static storage account key as part of their credentials file (we used to do this for Azure, but removed the requirement). I'd rather not do this, but it is an option.

carlisia:
Interesting. I was thinking of requiring some configuration for azure+restic users is what I would prefer. What you said sounds good to me. Would that be very onerous for the user?

skriss:
I can think of two challenges with using static storage account keys:
- velero supports different BSLs targeting different Azure storage accounts, and the key is per-storage-account. So we'd need to come up with some scheme that allowed users to configure a storage account key per storage account, either via multiple items in the secret, or via multiple secrets, etc.
- makes key rotation harder for the user - they'll need to update velero config any time they rotate the key for their storage account

carlisia:
Ah key rotation. Then no. I'd still like to revisit this when we look at restructuring plugins. For now I'll dup that code back into this repo.

carlisia:
@skriss should this Oracle documentation be removed: https://velero.io/docs/v1.1.0/oracle-config/

I vote yes. Maybe we can contact `@ mickeyboxell` and see if they can host it elsewhere.

skriss:
I'd like to do something like https://github.com/heptio/velero/issues/1396, particularly for documenting S3-compatible object storage providers that don't have a standalone plugin repo to put documentation in. This would be relevant for the `ibm-config.md` doc as well.

carlisia:
Ah cool. I'll get that done in a separate PR.

carlisia:
Note: the documentation for every plugin contains info for how to download Velero. I'm moving that to the Velero docs (and will point to it from the plugin docs). I also re-arranged install info under a new "install velero" section: https://github.com/vmware-tanzu/velero/pull/1890/commits/b08ecf86d4c933e4c95adff2e6158fdb61b68094

Gopkg.toml
```diff
@@ -70,18 +70,6 @@
   name = "github.com/Azure/go-autorest"
   version = "11.1.2"
 
-[[constraint]]
```
skriss: (general comment, not related to this line) it looked like running `dep ensure` again would drop some additional packages - specifically saw some additional Azure-related packages being removed when I ran it

pkg/cloudprovider/azure/object_store.go
```diff
@@ -18,121 +18,26 @@ package azure
```
skriss: let's rename this file -- maybe `restic.go` or just `azure.go`

site/docs/master/install-overview.md
```diff
@@ -7,6 +7,38 @@ We strongly recommend that you use an [official release][1] of Velero. The tarba
 
 _The code and sample YAML files in the master branch of the Velero repository are under active development and are not guaranteed to be stable. Use them at your own risk!_
 
+## Install Velero
```
skriss: Ah, I had made some similar changes in https://github.com/vmware-tanzu/velero/pull/1964. We can merge this and just follow up on my changes (which are a bigger reorg of this page) separately, or you can remove this file from this PR - up to you.

Gopkg.toml
carlisia: Hmyes, I'm fighting with dep, trying to figure it out.

pkg/cloudprovider/azure/object_store.go
prydonius: Does it make sense to move this to a package closer to restic, e.g. `pkg/restic/azure`?

Gopkg.toml
skriss: I tried this again - was able to simply run `dep ensure` after checking out your branch and have the extra packages removed.

pkg/cloudprovider/azure/object_store.go
skriss: yeah, that works for me

skriss:
Did some basic smoke-testing (backup + restore, including PVs) on GKE using a custom build of this branch + the GCP plugin - all looks good so far!

We do need to update the instructions in the plugin repos to include the `--plugins` flag in the `velero install` commands.

pkg/cloudprovider/azure/object_store.go
```diff
@@ -18,121 +18,26 @@ package azure
...
-	log             logrus.FieldLogger
-}
-
-func NewObjectStore(logger logrus.FieldLogger) *ObjectStore {
-	return &ObjectStore{log: logger}
-}
+	resourceGroupConfigKey = "resourceGroup"
+)
 
 func getStorageAccountKey(config map[string]string) (string, error) {
```
skriss: it looks like this function is out-of-date - doesn't include the recent changes to add support for Azure china/german clouds. Let's get it rebased so we can retain support for that here.

pkg/cloudprovider/azure/object_store.go
carlisia: Works for me too, glad you made this suggestion.

carlisia:
I tried adding the `github.com/prometheus/client_golang` forcefully by overriding it in the toml file but didn't work. Tried other shenanigans to w/o success so far on how to fix this:

https://travis-ci.org/vmware-tanzu/velero/builds/600903069#L785

skriss:
It looks like in your last force-push, pretty much every dependency got updated - see the Gopkg.lock diff: https://github.com/vmware-tanzu/velero/compare/b08ecf86d4c933e4c95adff2e6158fdb61b68094..08d1d6699a89f396cd0fc31685b46c4d412354ae#diff-bd247e83efc3c45ae9e8c47233249f18 which is probably causing these breaking changes (e.g. the prometheus client got bumped from v0.8 to v1.1).

Not exactly sure what triggered all of those updates, but I think we need to essentially revert all of that. Can you get back to where you were prior to that force-push?

carlisia:
I can. The trick is how to keep `dep ensure` from updating those. Trying..

skriss:
When I checked out your branch last week, I ran `dep ensure` (just that - no other flags/etc) which dropped the handful of Azure-related packages that I commented earlier. Did that not work for you?

carlisia:
That did. It was when I rebased with master that this mess happened.

skriss:
Got it. TBH it might be easiest to just start on a fresh branch/PR..not sure how you feel about that. I can help if you want. I think we just need to (1) delete/move the cloudprovider code, and then (2) run dep ensure. I could do it using this PR as a reference if you want.

skriss:
Alternately, I would say (1) revert all changes to Gopkg.toml, Gopkg.lock, and `vendor/`, (2) make sure the code compiles; (3) run `dep ensure`

carlisia:
Opened a new PR for this work: #1985, closing this.

PR #1892, opened 2019-09-19
sync controller: replace revision file with full diff each interval

### LINKED ISSUE #1343: Provide a way to change full sync period
When using Velero as a component in a larger backup system, we were looking for a way to sync the object store data more often than once an hour. This was done because backups using Velero's format were being written to the bucket out-of-band, but we wanted Velero to sync them into the cluster shortly after upload.

A hack around this was to delete the  `metadata/revision` file. The follow on effect there was that the entire bucket was scanned once a minute, and will be until the Velero server either performs or deletes a backup. This approach is also fragile, as we intend to do verification checks with #1072.

Ideally, we would be able to issue a command that does the sync on demand, as well as with the schedule.

skriss:

Closes #1343 

This is kind of what I was thinking ref. https://github.com/heptio/velero/issues/1343#issuecomment-519605161.  Instead of relying on the `revision` file to tell us when BSL contents change, we get the full list of backups from the BSL (this is just a single API call to the object store), and do a setwise comparison (based on name) to the backups in the cluster. We then sync/import any ones that are in the BSL but not in the cluster.

I also removed the limitation that the sync period must be >= 1 minute - I figure users should be able to set it lower if they want.

I haven't yet removed the actual updating of the `revision` file when backups happen - we could choose to leave that in for now, or remove it entirely.

If this makes sense, I'll proceed with tests/etc.


pkg/controller/backup_sync_controller.go
```diff
@@ -162,44 +134,44 @@ func (c *backupSyncController) run() {
...
 		}
 		backupStoreBackups := sets.NewString(res...)
-		log.WithField("backupCount", len(backupStoreBackups)).Info("Got backups from backup store")
+		log.WithField("backupCount", len(backupStoreBackups)).Debug("Got backups from backup store")
 
-		for backupName := range backupStoreBackups {
-			log = log.WithField("backup", backupName)
-			log.Debug("Checking this backup to see if it needs to be synced into the cluster")
+		// get a list of all the backups that exist as custom resources in the cluster
+		clusterBackups, err := c.backupLister.Backups(c.namespace).List(labels.Everything())
```
prydonius: curious: what's the difference between `labels.Everything()` and an empty `metav1.ListOptions{}`?

skriss: this is using a Lister, which doesn't take a `ListOptions` like the clients do -- it only takes a label selector. So same effect, but different types.

skriss:
I fixed the tests, and removed the code that gets/sets the `metadata/revision` file in object storage.

The only other thing to consider doing is adding some code to remove any existing `metadata/revision` files, or we could just add a script/instructions to the release notes to do this - any thoughts?

carlisia:
I'm fine with leaving the `revision` file in backups from past versions.

pkg/controller/backup_sync_controller.go
```diff
@@ -66,10 +65,10 @@ func NewBackupSyncController(
 	newPluginManager func(logrus.FieldLogger) clientmgmt.Manager,
 	logger logrus.FieldLogger,
 ) Interface {
-	if syncPeriod < time.Minute {
-		logger.Infof("Provided backup sync period %v is too short. Setting to 1 minute", syncPeriod)
+	if syncPeriod <= 0 {
```
nrb: Doesn't this conditional keep the effective minimum sync time at 1 minute?

skriss: No - you can explicitly specify a sync period of e.g. `15s`. This just defaults it to 1 minute if a 0/negative value was provided.

skriss:
> Have you observed any time differences (based on log messages would be simplest, I suppose) between the old approach and this one?

Haven't specifically looked, though performance seemed fine/~instantaneous in the testing I did. I can try with some bigger data sets.

skriss:
Tweaked logging so it only info-logs if there are actually backups to sync into the cluster for a location; otherwise, debug-logs.

skriss:
> Have you observed any time differences (based on log messages would be simplest, I suppose) between the old approach and this one?

The algorithm I'm using now should actually be faster than the old one -- previously, when syncing a location, we'd first get a list of all backups in the bucket, then make a K8s API call *per backup in the BSL* to check if that backup exists in-cluster.  Now, I'm making a single **List** call (to a lister) upfront to get all backups in the cluster, and then doing an in-memory set diff to figure out which ones should be synced in. So a bunch of over-the-wire API calls are saved.

carlisia:
@nrb PTAL.

PR #1893, opened 2019-09-19
Add LD_LIBRARY_PATH as an env varible for the use of vsphere plugin

lintongj:
Hi folks,

We are working on the vSphere plugin for velero. To enable the data path of backup and restore using VDDK API, VDDK libraries need to be dynamically linked in the run time. Hence, we plan to copy required libraries to the directory /plugins in velero pod container and propose to add the env variable, LD_LIBRARY_PATH, to /plugins. Feel free to suggest any better alternatives.

Best,
Lintong

### DIFF
--- a/changelogs/unreleased/1893-lintongj
+++ b/changelogs/unreleased/1893-lintongj
@@ -1,0 +1,1 @@
+Add LD_LIBRARY_PATH (=/plugins) to the env variables of velero deployment.
--- a/pkg/install/deployment.go
+++ b/pkg/install/deployment.go
@@ -162,7 +162,7 @@ func Deployment(namespace string, opts ...podTemplateOption) *appsv1.Deployment 
 									},
 								},
 								{
-									Name: "LD_LIBRARY_PATH",
+									Name:  "LD_LIBRARY_PATH",
 									Value: "/plugins",
 								},
 							},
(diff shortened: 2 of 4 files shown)


nrb:
> This would be a good thing to revisit along with the other open q's on install/plugin/BSL/etc. workflow.

Agreed. We should probably compile a list of things we want to address, even if we don't do that now.

lintongj:
Thanks Steve for the approval.

Best,
Lintong

prydonius:
Should we add some documentation to https://velero.io/docs/v1.1.0/plugins/ to mention that LD_LIBRARY_PATH is set to `/plugins` in the Velero container, so can be used to copy any libraries a plugin might need?

carlisia:
If @skriss is ok with revisiting this later then this is ok. It makes sense to me to try to avoid adding something that won't be used in all cases.

I also think @prydonius's suggestion is good: if you can add to our docs and also a changelog file (https://github.com/heptio/velero/tree/master/changelogs/unreleased) it would be great.

nrb:
I agree with all the sentiments above; I'll approve when the documentation and changelog are updated.

`LD_LIBRARY_PATH` can be fairly useful for any plugins that require C extensions, but I also think this should be added to our list of things to address for a plugin design iteration.

lintongj:
Agreed with Nolan that LD_LIBRARY_PATH could be useful any plugins that requires c extensions. 

Any guidance about where to update the documentation and changelog?

skriss:
`site/docs/master/plugins.md` is probably a good place to start for docs.

https://github.com/heptio/velero/blob/master/CONTRIBUTING.md#changelog has details on the changelog.

lintongj:
Thanks for the approvals.

PR #1896, opened 2019-09-20
restic: only backup ReadWriteMany PVC's once per velero backup

### LINKED ISSUE #1744: Restic backup is repeated as many times as the number of pods in a deployment for RWX volumes
I have installed the NFS server provisioner in order to have a ReadWriteMany volume shared by multiple pods in a deployment. I have annotated the deployment/pods so that Restic can back up the shared volume, but the Restic backup is repeated for each pod.

One backup only should be done for the same volume regardless of how many pods are sharing it.

**Environment:**

- Velero version (use `velero version`): v1.0.0
- Kubernetes version (use `kubectl version`): v1.14.3
- Kubernetes installer & version: Rancher 2.2.6
- Cloud provider or hardware configuration: Hetzner Cloud
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

skriss:

Closes #1744 

Unit tests look good, but I'd still like to do some manual testing before merge so leaving as draft.

### DIFF
--- a/changelogs/unreleased/1896-skriss
+++ b/changelogs/unreleased/1896-skriss
@@ -1,0 +1,1 @@
+restic: only backup read-write-many PVCs at most once, even if they're annotated for backup from multiple pods.
--- a/pkg/backup/backup_test.go
+++ b/pkg/backup/backup_test.go
@@ -22,6 +22,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"sort"
@@ -2099,22 +2100,24 @@ func TestBackupWithHooks(t *testing.T) {
 	}
 }
 
-type fakeResticBackupperFactory struct {
-	podVolumeBackups []*velerov1.PodVolumeBackup
-}
+type fakeResticBackupperFactory struct{}
 
 func (f *fakeResticBackupperFactory) NewBackupper(context.Context, *velerov1.Backup) (restic.Backupper, error) {
-	return &fakeResticBackupper{
-		podVolumeBackups: f.podVolumeBackups,
-	}, nil
+	return &fakeResticBackupper{}, nil
 }
 
-type fakeResticBackupper struct {
-	podVolumeBackups []*velerov1.PodVolumeBackup
-}
+type fakeResticBackupper struct{}
 
-func (b *fakeResticBackupper) BackupPodVolumes(backup *velerov1.Backup, pod *corev1.Pod, _ logrus.FieldLogger) ([]*velerov1.PodVolumeBackup, []error) {
-	return b.podVolumeBackups, nil
+// BackupPodVolumes returns one pod volume backup per entry in volumes, with namespace "velero"
+// and name "pvb-<pod-namespace>-<pod-name>-<volume-name>".
+func (b *fakeResticBackupper) BackupPodVolumes(backup *velerov1.Backup, pod *corev1.Pod, volumes []string, _ logrus.FieldLogger) ([]*velerov1.PodVolumeBackup, []error) {
+	var res []*velerov1.PodVolumeBackup
+	for _, vol := range volumes {
+		pvb := builder.ForPodVolumeBackup("velero", fmt.Sprintf("pvb-%s-%s-%s", pod.Namespace, pod.Name, vol)).Result()
+		res = append(res, pvb)
+	}
+
+	return res, nil
 }
 
 // TestBackupWithRestic runs backups of pods that are annotated for restic backup,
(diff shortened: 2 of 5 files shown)


skriss:
Was able to confirm this worked as expected on an AWS cluster using EFS RWX volumes

nrb:
Just for clarity - what does this look like on restore? The volume's unpopulated until the pod that got selected is restored?

Did you do any investigation into making the volume that gets chosen for backup deterministic? That may be pretty complicated, as we don't really know what order we'll receive the pods themselves.

skriss:
> Just for clarity - what does this look like on restore? The volume's unpopulated until the pod that got selected is restored?

Correct. I guess there's potentially an issue where if, on restore, the pod that is associated with the backup of the PVC isn't getting restored (due to a label selector, for example), then the PVC data won't get restored. I'll have to think about if there's a way around this.

> Did you do any investigation into making the volume that gets chosen for backup deterministic? That may be pretty complicated, as we don't really know what order we'll receive the pods themselves.

No - I guess we could alphabetize within the namespace or something.

prydonius:
> Correct. I guess there's potentially an issue where if, on restore, the pod that is associated with the backup of the PVC isn't getting restored (due to a label selector, for example), then the PVC data won't get restored. I'll have to think about if there's a way around this. It does seem like a pretty unusual use case, though -- I'd think if you have several pods sharing a volume, you'd probably be restoring them as a group.

Doesn't the PV/PVC restore occur when the PV/PVC object is restored, or does it work in the same way as a backup as an additional item?

Seems like it would also be pretty difficult to exclude the associated Pod, if the main usecase is a Deployment/*Set with multiple Pods, they will all have the same labels. That said, it is possible to have different groups of Pods to use the same RWX volume.

skriss:
> Doesn't the PV/PVC restore occur when the PV/PVC object is restored, or does it work in the same way as a backup as an additional item?

The restic restore will run when the **pod** using the PVC that was used to take the restic backup of the PVC gets restored.

> Seems like it would also be pretty difficult to exclude the associated Pod, if the main usecase is a Deployment/*Set with multiple Pods, they will all have the same labels. That said, it is possible to have different groups of Pods to use the same RWX volume.

Yeah, agree it's unlikely though not impossible. I'll give this some more thought today.

prydonius:
> The restic restore will run when the pod using the PVC that was used to take the restic backup of the PVC gets restored.

Ah yes, of course - for some reason I was thinking of snapshots yesterday.

I suppose this problem will go away if we run restic ops in worker pods as per #1653.

skriss:
> I guess there's potentially an issue where if, on restore, the pod that is associated with the backup of the PVC isn't getting restored (due to a label selector, for example), then the PVC data won't get restored. I'll have to think about if there's a way around this. It does seem like a pretty unusual use case, though -- I'd think if you have several pods sharing a volume, you'd probably be restoring them as a group.

To fix the limitation, we'd need to somehow track all the pods that use the RWX volume during backup (but only back it up from one pod), and then during restore, if the pod it was backed up from isn't being restored, check if any of the other pods using it *are* being restored and if so, restore it from there.

We don't really have a natural place to store this info - it would likely end up in an annotation on the `PodVolumeBackup`.

@nrb @carlisia do you think this is worth pursuing? It seems like a low-probability scenario to me, and even without addressing it, I think this PR still has value, but I can do more R&D on it if folks think it's worthwhile.

nrb:
> To fix the limitation, we'd need to somehow track all the pods that use the RWX volume during backup (but only back it up from one pod), and then during restore, if the pod it was backed up from isn't being restored, check if any of the other pods using it are being restored and if so, restore it from there.
>
> We don't really have a natural place to store this info - it would likely end up in an annotation on the PodVolumeBackup.

Yeah, that's kind of what I was afraid of. I'm starting to wonder if tying restic directly to volume mounts within a pod is even the right approach, since I'm seeing it generate a lot of confusion when someone wants to include a PV, but doesn't see it in the backup under persistent volumes.

As it stands, I think this does have value for many people using RWX, but it has some corner cases around resource inclusion that could lead to the data being backed up but not restored. If we can document that, I think it would be useful for future reference and this change could go in.

skriss:
It would definitely be nicer from a UX perspective to re-orient restic around PV/PVCs, rather than pod volumes -- at least for folks who aren't backing up non-PV/PVCs :) We put a question on the user survey about that, so we'll have to take a look at the full results. I have some thoughts around backend things we could do differently if we took that approach.

danboucher78:
I'm agree with the PV/PVCs perspective backup.  Also, during the restore, it would be the first thing to complete before restoring the rest of the resources.  I say that because I imagine a process in a pod being able to start even if the restore isn't completed.  Data corruption would be possible in certain cases.  It would also prevent pod restarting loop in some others cases.

nrb:
Given that this has value as it stands and that Steve is exploring the PV/PVC approach, I'm ok merging this for now.

PR #1898, opened 2019-09-20
velero API type changes for structural schema CRDs

### LINKED ISSUE #1675: Investigate CRD structural schema

prydonius:
- make optional fields consistent with comment and omitempty struct tags
- adds kubebuilder annotations for structural schema validations

xref #1885 

fixes #1675


pkg/apis/velero/v1/backup.go
```diff
@@ -24,76 +24,102 @@ import (
...
+	// +optional
+	ExcludedResources []string `json:"excludedResources,omitempty"`
+
 	// LabelSelector, if specified, filters the resources to which this hook spec applies.
+	// +optional
 	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
+
 	// PreHooks is a list of BackupResourceHooks to execute prior to storing the item in the backup.
 	// These are executed before any "additional items" from item actions are processed.
+	// +optional
```
skriss: Is there a way to say "either PreHooks or PostHooks must be non-empty"?

pkg/apis/velero/v1/backup.go
```diff
@@ -107,14 +133,22 @@ type BackupResourceHook struct {
...
+	Container string `json:"container,omitempty"`
+
 	// Command is the command and arguments to execute.
+	// +kubebuilder:validation:MinItems=1
 	Command []string `json:"command"`
+
 	// OnError specifies how Velero should behave if it encounters an error executing this hook.
-	OnError HookErrorMode `json:"onError"`
+	// +optional
+	// +kubebuilder:validation:Enum=Continue;Fail
```
skriss: yuck, is this the only way to handle enums?

pkg/apis/velero/v1/backup.go
```diff
@@ -164,47 +199,57 @@ const (
...
+	Version int `json:"version,omitempty"`
 
 	// Expiration is when this Backup is eligible for garbage-collection.
-	Expiration metav1.Time `json:"expiration"`
+	// +optional
+	Expiration metav1.Time `json:"expiration,omitempty"`
 
 	// Phase is the current state of the Backup.
-	Phase BackupPhase `json:"phase"`
+	// +optional
```
skriss: should we be enumerating the enum values here?

pkg/apis/velero/v1/backup.go
prydonius: I think we'd need the "OneOf" validation in OpenAPIv3, but it doesn't look like Kubebuilder supports this (https://github.com/kubernetes-sigs/kubebuilder/issues/301, https://book.kubebuilder.io/reference/markers/crd-validation.html).

I'm actually not sure where Pre/PostHooks are used. Do you have any examples of them? Are you saying that if a PreHook is defined, a PostHook cannot be defined (and vice versa)?

pkg/apis/velero/v1/backup.go
prydonius: Unfortunately yeah :/, controller-tools isn't able to pick this up from the constants we've defined. I can move this annotation down to where `HookErrorMode` is defined, as in the example [here](https://book.kubebuilder.io/reference/generating-crd.html#validation), maybe that will help make sure we don't miss it if we add an extra value in the future?

pkg/apis/velero/v1/backup.go
prydonius: I decided to not do any validation for status fields as we don't expect them to be set by users - can go and add them if we think validation is needed here though.

pkg/apis/velero/v1/backup.go
skriss: yeah, I think moving it might be better so it's right next to where the values are.

pkg/apis/velero/v1/backup.go
skriss: Got it - I'll read through those refs, thanks for the links.

I'll have to double-check the exact object structure here, but I think the semantics should be that there needs to be at least one pre or post hook - both is fine, but the slices can't both be empty simultaneously.

skriss: We'll validate this in the controller anyway (I think), so if we can't do it with the structural schema it's OK - but if we can, it's good to add.

prydonius:
10e3787 makes some fields nullable, which was necessary for some CLI operations, and when reimporting backups from object storage.

Looks like `metav1.Time` and array fields in particular should be nullable in status fields.

Would be great if someone could do some more manual testing of this to ensure this doesn't break any operations, I've tested basic backup/restore creation, `velero version` (ServerStatusRequest), `velero backups logs` (DownloadRequest), `velero backup delete` (DeleteBackupRequest).

skriss:
I can do some testing on this today.

pkg/apis/velero/v1/backup.go
```diff
@@ -24,76 +24,113 @@ import (
...
-	ExcludedResources []string `json:"excludedResources"`
+	// +optional
+	// +nullable
+	ExcludedResources []string `json:"excludedResources,omitempty"`
 
 	// LabelSelector is a metav1.LabelSelector to filter with
 	// when adding individual objects to the backup. If empty
 	// or nil, all objects are included. Optional.
-	LabelSelector *metav1.LabelSelector `json:"labelSelector"`
+	// +optional
```
nrb: Should this be nullable, since it can be `nil`?

pkg/apis/velero/v1/backup.go
nrb: It would be awesome if we could get the `OneOf` validation working so that the API server handles more of this validation for us...but if Kubebuilder doesn't support it yet, we may have to wait.

pkg/apis/velero/v1/backup.go
nrb: Even if it's not set by users, maybe it would be useful for catching bugs.

pkg/cmd/cli/backup/create.go
```diff
@@ -256,6 +256,7 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
 
 func (o *CreateOptions) BuildBackup(namespace string) (*velerov1api.Backup, error) {
 	backupBuilder := builder.ForBackup(namespace, o.Name)
+	fmt.Printf("%+v\n", backupBuilder.Result())
```
nrb: Should this be removed before final merge?

pkg/install/resources.go
```diff
@@ -219,7 +220,8 @@ func AllResources(o *VeleroOptions) (*unstructured.UnstructuredList, error) {
 	// Set the GVK so that the serialization framework outputs the list properly
 	resources.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "List"})
 
-	for _, crd := range CRDs() {
+	for _, crd := range crds.CRDs {
+		crd.SetLabels(labels())
```
nrb: Why are we setting the labels here?

pkg/cmd/cli/backup/create.go
prydonius: 😅 yep, thanks

pkg/install/resources.go
prydonius: `velero install` previously created the CRDs with the `component: velero` label. The generated CRDs don't have that, so we need to add them here. I like the idea of everything `velero install` having that label.

pkg/apis/velero/v1/backup.go
prydonius: > I'll have to double-check the exact object structure here, but I think the semantics should be that there needs to be at least one pre or post hook - both is fine, but the slices can't both be empty simultaneously.

@skriss hmm, I've definitely created backups before without any Pre or Post hooks? By default `velero create backup` sets `hooks: {}`.

pkg/apis/velero/v1/backup.go
prydonius: Yeah I was thinking that, although it gives us less flexibility. But I think it is worth adding, so will update this.

pkg/apis/velero/v1/backup.go
skriss: Yeah you can definitely create a backup without hooks. But, if you do create a hook, then it needs to have either a pre- or a post- hook component (or both).

skriss: This isn't a must-have for getting this PR in, we can defer for now.

skriss:
What's the difference between "optional" and "nullable"?

prydonius:
> What's the difference between "optional" and "nullable"?

An optional field can be left out in the manifest, but can't be set to "null". A nullable field can be set to "null" (but can also be required).

prydonius:
One issue with this is that CI won't fail if the CRDs aren't updated (if `make update` isn't run). We need an equivalent of `--verify-only` for the CRD generation, like there is for the code-generator. We should be able to do this by generating the CRDs and ensuring there's no changes to the manifest, but I think we can do this as a follow up to this PR.

pkg/install/resources.go
nrb: Gotcha, and agreed.

skriss:
Testing has looked fine to me - I tried out restic backups/restores as well.

prydonius:
FYI adding the enums for the status fields is the only remaining thing in this PR. I did some more digging and can't see a way to support oneOf with kubebuilder annotations, so we'll have to just "validate" that in the server.

prydonius:
07d1547 adds enum annotations for all the CRD phases - PTAL!

skriss:
Oh, just realized there's no changelog here - probably want to add one since these will now be used by `velero install`!

prydonius:
Oops - sorry, always forget that 😅

PR #1901, opened 2019-09-23
Make "not restored" message a bit better

carlisia:
What the new messages try to convey:

In one case, the restore it is legit and possible, but we are choosing to skip it because the resources are the same. In the other case, it is NOT possible and the reason why is the difference between the resources.

Also changed the formatting a bit for consistency.

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -1062,13 +1062,13 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
 					ctx.log.Infof("ServiceAccount %s successfully updated", kube.NamespaceAndName(obj))
 				}
 			default:
-				e := errors.Errorf("not restored: %s and is different from backed up version.", restoreErr)
+				e := errors.Errorf("restore of %s not possible: it already exists in the cluster but is different from the backed up version.", restoreErr)
 				addToResult(&warnings, namespace, e)
 			}
 			return warnings, errs
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/restore/restore.go
```diff
@@ -1062,13 +1062,13 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
 					ctx.log.Infof("ServiceAccount %s successfully updated", kube.NamespaceAndName(obj))
 				}
 			default:
-				e := errors.Errorf("not restored: %s and is different from backed up version.", restoreErr)
+				e := errors.Errorf("error restoring %s: not possible since it already exists in the cluster but is different from the backed up version.", restoreErr)
```
prydonius: should this be "and is different" instead of "but is different"?

skriss: Keep in mind that `restoreErr` already says something like "<resource identifier> already exists in cluster". It would be fine to not even include `restoreErr` in `e` if we just want to restate it, but I would definitely ensure that the resource type, namespace, and name are included in the message here.

skriss: Maybe splitting it into two sentences would help - something like "Couldn't restore ConfigMap foo/bar because it already exists in the cluster. Warning: the in-cluster version is different than the backed-up version".

nrb: I like splitting them into 2 sentences, too.

carlisia:
@skriss / @nrb, ready for another review.

PR #1904, opened 2019-09-24
Increment logging for item backupper

### LINKED ISSUE #1844: Confusing log fields when backing up additional items for an object
When testing #1843, I noticed that the logger in the itemBackupper has the `resource` field preset from the resourceBackupper. If an item in a resource pulls in any additional items (e.g. a Pod's PVC or PV), the resource field continues to reflect the current resource (e.g. "pods" rather than "persistentvolumeclaims").

Screenshot 2019-09-04 at 2 45 52 PM

We should set the resource field using the groupResource passed into `backupItem` to ensure we're not confusingly displaying the wrong resource for the item.

We might also want to display the resource we're going through in a different field to make it more clear what resource pulled in an additional item, but I can't think of a good name for this field. Also, the logs around the additional item already help to indicate what the original resource/item is.

spiffcs:
Fixes #1844 

I've added logic to the `backupItem` function.

If the `Kind` of the object passed in is different from the `APIResource.Name` set in the `backupResource` function then logging is updated to make it more clear what is being backed up.

I can always update this to incorporate the parent field as suggested in #1844 

Screen Shot 2019-09-24 at 7 18 56 AM

Signed-off-by: Christopher Phillips <cphillips918@gmail.com>

### DIFF
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -120,9 +120,7 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 
 	log := logger.WithField("name", name)
 	log = log.WithField("resource", groupResource)
-	if namespace != "" {
-		log = log.WithField("namespace", namespace)
-	}
+	log = log.WithField("namespace", namespace)
 
 	if metadata.GetLabels()["velero.io/exclude-from-backup"] == "true" {
 		log.Info("Excluding item because it has label velero.io/exclude-from-backup=true")
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


prydonius:
Thanks so much for this @spiffcs! Sorry, I missed your comment in https://github.com/heptio/velero/issues/1844#issuecomment-534508452, but can you confirm `groupResource` isn't usable here? It does get set to the groupResource of the additional item here: https://github.com/heptio/velero/blob/1683c57da51af4acfac4e179eeb554480e2365aa/pkg/backup/item_backupper.go#L339, so it should work and saves us having to look at the type accessor.

spiffcs:
Yep! I’ll give it a look when I get home tonight to checkout that behavior. Sorry for missing that bit.

spiffcs:
@prydonius you were right! Here is the revised easy fix with a screen shot:
Screen Shot 2019-09-24 at 7 18 56 AM

pkg/backup/item_backupper.go
```diff
@@ -119,6 +119,7 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 	name := metadata.GetName()
 
 	log := logger.WithField("name", name)
+	log = log.WithField("resource", groupResource)
 	if namespace != "" {
```
skriss: The one thing I noticed on the screenshot is that when we get to backing up the `PersistentVolume`, we still show a log field of `namespace=default`, which is incorrect since it's non-namespaced.  I think the easiest thing to do here is just remove the L123 `if` statement, and always execute L124.

spiffcs:
@skriss thanks for the feedback! I updated this PR to match those changes

PR #1906, opened 2019-09-24
fail on make verify if generated CRDs differ

### LINKED ISSUE #1905: Support --verify-only for CRD generation
CI won't fail if the Velero API type annotations are modified and CRDs aren't updated (if make update isn't run). We need an equivalent of --verify-only for the CRD generation, like there is for the code-generator. We should be able to do this by generating the CRDs and ensuring there's no changes to the manifest, but I think we can do this as a follow up to this PR.

xref https://github.com/heptio/velero/pull/1898#issuecomment-534653028

prydonius:

fixes #1905

### DIFF
--- a/hack/verify-generated-crd-code.sh
+++ b/hack/verify-generated-crd-code.sh
@@ -17,3 +17,13 @@
 HACK_DIR=$(dirname "${BASH_SOURCE}")
 
 ${HACK_DIR}/update-generated-crd-code.sh --verify-only
+
+# ensure no changes to generated CRDs
+if ! git diff --exit-code pkg/generated/crds/crds.go >/dev/null; then
+  # revert changes to state before running CRD generation to stay consistent
+  # with code-generator `--verify-only` option which discards generated changes
+  git checkout pkg/generated/crds
+
+  echo "run 'make update' to update generated CRDs"
+  exit 1
+fi
--- a/hack/verify-generated-crd-code.sh
+++ b/hack/verify-generated-crd-code.sh
@@ -24,6 +24,6 @@ if ! git diff --exit-code pkg/generated/crds/crds.go >/dev/null; then
   # with code-generator `--verify-only` option which discards generated changes
   git checkout pkg/generated/crds
 
-  echo "run 'make update' to update generated CRDs"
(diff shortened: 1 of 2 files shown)


prydonius:
Funnily enough the CRDs in master are out-of-date, so CI is failing here... this will be fixed once #1898 is merged and this is rebased.

skriss:
This closes #1905 right?

hack/verify-generated-crd-code.sh
```diff
@@ -17,3 +17,13 @@
...
 
 ${HACK_DIR}/update-generated-crd-code.sh --verify-only
+
+# ensure no changes to generated CRDs
+if ! git diff --exit-code pkg/generated/crds/crds.go >/dev/null; then
+  # revert changes to state before running CRD generation to stay consistent
+  # with code-generator `--verify-only` option which discards generated changes
+  git checkout pkg/generated/crds
+
+  echo "run 'make update' to update generated CRDs"
```
skriss: Let's make it a little clearer in this message that there's an error due to the CRDs being out of date, and the solution is 'make update'

prydonius:
Yep - thanks, connected the right issue

hack/verify-generated-crd-code.sh
prydonius: updated, basing it a bit off of the gofmt error message https://github.com/heptio/velero/blob/166565743f46e0bd254f5a41d80720a5198ce7ea/hack/update-fmt.sh#L46

nrb:
#1898 is merged, this can be rebased.

prydonius:
rebased!

hack/verify-generated-crd-code.sh
```diff
@@ -17,3 +17,13 @@
 HACK_DIR=$(dirname "${BASH_SOURCE}")
 
 ${HACK_DIR}/update-generated-crd-code.sh --verify-only
```
nrb: Should the `--verfify-only` flag here be removed, since it's not being used anymore?

skriss: @nrb since adnan's out for a week, do you want to merge this as-is now and do a followup to remove that flag? would love to get it out of the queue :)

PR #1916, opened 2019-09-27
Reorg install and plugin docs

### LINKED ISSUE #1396: Add `contrib` section to repo with community-contributed examples/etc.
We want to encourage content to be contributed from the community (e.g. setup instructions for specific S3-compatible object stores, alternate YAML configs, etc.) but we want to be clear about what the core team officially supports vs. doesn't.

Let's add a directory somewhere in the repo where these things can easily be added.  As part of this, we should review what's currently in our repo and move any relevant content in there. For example, we have `docs/ibm-config.md` which was added by a contributor, but is not supported by the core team. We should move it.

carlisia:
Closes #1396 


carlisia:
Note: I only made changes to the `master` location for now.

carlisia:
Another note: I spent a ton of time trying to figure out why these links are not working, they are adding the name of the page they are on to the URL path:

image

😠 

The links should be right? (Adding `.md` doesn't fix it):
https://github.com/heptio/velero/pull/1916/files#diff-bf1c70028a3cddfecef3a254f47c84a7R17

The same is happening with the links to the ibm and oracle on the community-plugin page.

carlisia:
Instead of adding the oracle and ibm docs to a `contrib` folder in the repo, I added here, and then rearranged the docs to spruce things up + split things into velero supported vs. community plugin.

If this works for everyone, when we move the repo we'll only have to update the links for the velero supported plugins to point to the URL in the new plugin repos (as well as remove the provider docs).

site/docs/master/install-overview.md
```diff
@@ -26,12 +26,7 @@ velero install \
 
 When using node-based IAM policies, `--secret-file` is not required, but `--no-secret` is required for confirmation.
 
-For provider-specific instructions, see:
-
-* [Run Velero on AWS][0]
-* [Run Velero on GCP][1]
-* [Run Velero on Azure][2]
-* [Use IBM Cloud Object Store as Velero's storage destination][4]
+For provider-specific instructions, see the navigation on the left under `Plugins`.
```
skriss: Personally, I think it'd be easier for the user to navigate the install docs if we added *more* content to this Install Overview page about the compatible platforms and directly linked to instructions for each from here, rather than requiring them to jump into a whole different section on plugins to find links to install instructions for their platform. I think this page itself could use some reorg, maybe something like:

- Background/Prerequisites
    - Compatible object storage system, optional compatible block storage system(s), restic, ...
    - `velero install` or Helm chart
- Provider-specific instructions
    - Velero team-supported
    - Community
- `velero install` reference (details on all the different flags)

Thoughts?

carlisia:
@jonasrosland are you aware of what this is: https://github.com/heptio/velero/pull/1916#issuecomment-535782904

skriss:
I'm not exactly sure what's going on with the links, but for Cormac's blog posts we had to do a few `../<page>`-style links -- because individual pages' URLs look like e.g. `https://velero.io/docs/v1.1.0/upgrade-to-1.1/` (note trailing slash) which probably looks like a directory when adding on a relative link.

carlisia:
K, I'll check on that. The links/pages look exactly like the others. But I'll check. 🤷‍♀

site/docs/master/install-overview.md
carlisia: I made a proposed outline here based on feedback: https://hackmd.io/UpamNjb3SR2pSaJfeeHfzA

My thinking: let's nail down on this PR:
- the reorg as far as an outline
- move the existing content to fit this outline
- any writing/rewriting of content we can do in a new PR.

skriss: sounds like a good plan to me! taking a look at the outline shortly

skriss: The outline looks about right to me, I added a couple comments in the HackMD.

carlisia: Read your comments, make sense to me.

carlisia:
@skriss / @nrb, I reference one of the new docs in this PR on the blog post, PTAL when you have a chance so we can merge it before the post.

skriss:
You could also just point the blog link to the existing `support-matrix.md` page, and then change the blog link as part of this PR, if you don't want to block the blog on merge of this PR.

I'll review post-community mtg.

README.md
```diff
@@ -15,16 +15,7 @@ Velero consists of:
 * A server that runs on your cluster
 * A command-line client that runs locally
 
-You can run Velero in clusters on a cloud provider or on-premises. For detailed information, see [Compatible Storage Providers][99].
-
-## Installation
```
skriss: What's the thinking for removing this section?

README.md
```diff
@@ -51,10 +42,11 @@ See [the list of releases][6] to find out about feature changes.
 [11]: https://kubernetes.io/docs/tasks/tools/install-kubectl/#tabset-1
 [12]: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/README.md
 [14]: https://github.com/kubernetes/kubernetes
+[17]: /site/docs/master/restic.md
```
skriss: Probably want to link to the velero.io website here

site/_data/master-toc.yml
```diff
@@ -4,27 +4,33 @@ toc:
...
-      - page: Run on Azure
-        url: /azure-config
-      - page: Run on GCP
-        url: /gcp-config
-      - page: Restic setup
+      - page: Requirements
+        url: /install-requirements
+      - page: Installing
+        url: /installing
+      - page: Supported providers
```
skriss: Do you think "supported platforms" is more clear here? I could go with either.

site/_data/master-toc.yml
```diff
@@ -4,27 +4,33 @@ toc:
...
         url: /locations
-      - page: Supported platforms
-        url: /support-matrix
   - title: Install
     subfolderitems:
       - page: Overview
         url: /install-overview
       - page: Upgrade to 1.1
         url: /upgrade-to-1.1
-      - page: Quick start with in-cluster MinIO
```
skriss: The idea with the quick start was to have a place for new users to quickly get up and running for evaluation/experimentation - by removing it from the TOC, that guide becomes buried. Personally I'd opt to keep it around in the TOC - what do y'all think?

site/_data/master-toc.yml
```diff
@@ -4,27 +4,33 @@ toc:
...
+        url: /installing
+      - page: Supported providers
+        url: /supported-providers
+      - page: Restic integration
         url: /restic
+      - page: Examples
+        url: /examples
+      - page: Uninstalling
+        url: /uninstalling
+  - title: Plugins
```
skriss: I'd move "Plugins" below "Use" in the TOC.

carlisia:
> You could also just point the blog link to the existing support-matrix.md page, and then change the blog link as part of this PR, if you don't want to block the blog on merge of this PR.

Duh, of course! I'll do that. So let's not worry about this PR for the blog post. But still we have the tag issue.

site/docs/master/installing.md
```diff
@@ -0,0 +1,78 @@
+# Installing Velero
```
skriss: I'm finding this page to be confusing as-is, as it's kind of a random collection of information related to install that doesn't actually have any specific installation instructions. I think I'd prefer to either:
- move this content into the "Install Overview" page, as background (if we choose this, I bet some of the content could be removed to shorten it up as it's probably redundant with other spots)
- call this "Install Reference" or something other than just "Installing" (both the page name and in the TOC)
- reorganize this page so that it functions better as a main "installing" page

what do others think?

site/docs/master/supported-providers.md
```diff
@@ -0,0 +1,77 @@
+# Providers
+
+Velero can store its backups in any S3 or S3 compatible object storage.
```
skriss: This doesn't make sense as a headline here since object stores other than S3 are supported, and also this page covers volume snapshot operations as well. I'd go with something more similar to what we already have on the existing page, e.g. "Velero supports a variety of storage providers for different backup and snapshot operations. Velero has a plugin system which allows anyone to add compatibility for additional backup and volume storage platforms without modifying the Velero codebase."

site/docs/master/supported-providers.md
```diff
@@ -0,0 +1,77 @@
...
+
+## S3-Compatible object store providers
```
skriss: I would move this below the "Velero supported providers" section

site/docs/master/supported-providers.md
```diff
@@ -0,0 +1,77 @@
...
+
+Velero uses [Amazon's Go SDK][0] to connect to the AWS S3 API. Some third-party storage providers also support the S3 API, and users have reported the following providers work with Velero:
```
skriss: We might want to reword this to say "Velero's AWS Object Store plugin uses Amazon's Go SDK..."

site/docs/master/installing.md
skriss: To add more context here - since this page is called "Installing" in the TOC, I'd expect it to be the main page that has step-by-step instructions or well-organized links to other pages, but it's not really that at the moment, hence the ideas for different organization.

README.md
carlisia: To avoid having install instructions scarred in multiple places. The content under "More information" should be all that is necessary, and should be very high level pointing to one entry point, imo. But maybe we can make this better.

site/_data/master-toc.yml
carlisia: What is the difference between platform and provider?

site/_data/master-toc.yml
carlisia: I guess I resist this because that documentation is not for production, and it would be the only thing not production worthy in this entire section. Could we add it as a `Quick start` submenu to either the `Use` or `Contributing` section?

site/docs/master/supported-providers.md
carlisia: Thanks for the explanation.

site/docs/master/installing.md
carlisia: Ok. I see your point and agree. I'm thinking about it.

site/_data/master-toc.yml
skriss: I'd consider them roughly synonyms, but I know you disliked "provider" so was looking for alternatives

README.md
skriss: OK, I guess since we now have `velero install` rather than static YAML this disclaimer isn't so important anymore. WFM.

site/_data/master-toc.yml
skriss: Personally I'd rather have it higher up since the point of it is to be a good starting point for new users. Let's see what @nrb thinks. Also FYI Istio has something similar in their Install section (I think their docs are pretty reasonably organized): https://istio.io/docs/setup/install/

site/_data/master-toc.yml
carlisia: I think minimizing the numbers of terms used for the same thing makes the documentation easier to follow.

As for `provider`, I'm starting to think we need to make more obvious the provider x plugins relationship (a provider has many plugins, a plugin has many providers). WIP.

site/_data/master-toc.yml
carlisia: Ok, I like `evaluation install`, that fits.

We can also move the Helm installation instructions up to a subitem if we can add more content to it. Not wanting to do that now tho.

site/_data/master-toc.yml
skriss: :+1: OK I'm happy with the existing title.

nrb: I'm still not quite sure where the provider confusion is coming from ...I use it as short hand for "cloud provider."

A cloud provider can have many services. A couple of those services are used by Velero - object storage and volume snapshotting. Thus, a provider could have 1 plugin for each of those 2 services, for a total of 2. These two plugins can both live in the same binary, and the correct one will be invoked by our plugin machinery. Many others could also be in that binary (as seen in Velero's included plugins).

Things such as Portworx and Minio offer only one of those two services - volumes for Portworx, object storage for Minio. Thus they'd each only have 1 plugin, which may or may not be the only plugin within the binary. 

In Velero's API, we don't provide a field for the plugin's type, using the struct type to differentiate. This does mean that solely the names aren't enough to know what a plugin does.

site/_data/master-toc.yml
nrb: In my own reading, a "Quick start" isn't meant to be production ready. But I like the "Evaluation install" title and some wording inside the page itself.

README.md
```diff
@@ -51,10 +42,11 @@ See [the list of releases][6] to find out about feature changes.
 [11]: https://kubernetes.io/docs/tasks/tools/install-kubectl/#tabset-1
 [12]: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/README.md
 [14]: https://github.com/kubernetes/kubernetes
+[17]: https://velero.io/docs/master/restic.md
```
skriss: Doesn't look like this is used anywhere, no need to add it

README.md
```diff
@@ -51,10 +42,11 @@ See [the list of releases][6] to find out about feature changes.
...
 [24]: https://groups.google.com/forum/#!forum/projectvelero
 [25]: https://kubernetes.slack.com/messages/velero
 [28]: https://velero.io/docs/install-overview
 [29]: https://velero.io/docs/
 [30]: https://velero.io/docs/troubleshooting
 [99]: https://velero.io/docs/support-matrix
-[100]: /site/docs/master/img/velero.png
+[100]: https://velero.io/docs/master/img/velero.png
```
skriss: Doesn't look like this is used anywhere, no need to add it

README.md
```diff
@@ -51,10 +42,11 @@ See [the list of releases][6] to find out about feature changes.
 [11]: https://kubernetes.io/docs/tasks/tools/install-kubectl/#tabset-1
 [12]: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/README.md
 [14]: https://github.com/kubernetes/kubernetes
+[17]: https://velero.io/docs/master/restic.md
 [24]: https://groups.google.com/forum/#!forum/projectvelero
 [25]: https://kubernetes.slack.com/messages/velero
 [28]: https://velero.io/docs/install-overview
 [29]: https://velero.io/docs/
 [30]: https://velero.io/docs/troubleshooting
 [99]: https://velero.io/docs/support-matrix
```
skriss: probably want to drop link [99] too since the page no longer exists

site/docs/master/README.md
```diff
@@ -53,7 +44,7 @@ See [the list of releases][6] to find out about feature changes.
 [11]: https://kubernetes.io/docs/tasks/tools/install-kubectl/#tabset-1
 [12]: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/README.md
 [14]: https://github.com/kubernetes/kubernetes
-
+[17]: restic.md
```
skriss: Doesn't look like this is used anywhere, no need to add it

site/docs/master/custom-plugins.md
```diff
@@ -1,4 +1,4 @@
-# Plugins
+# Custom plugins
```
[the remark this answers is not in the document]
carlisia: Hm. Yeah, now that we are going to have all plugins out of tree there's no more native x custom so I'll change it back.

README.md
carlisia: It's on line 1.

skriss: Ah. I don't think it needed to change, but I guess it's fine to hit the website.

PR #1922, opened 2019-10-01
Prevent master from being tagged as latest

nrb:
By default, git does not fetch tags on a checkout, so fetch those when
building a tag.

When the tags are not fetched and building master, both HIGHEST and
LATEST_TAG were "", which was equal. Thus, every master push was tagged
as latest. This is now handled correctly.

### DIFF
--- a/hack/gcr-push.sh
+++ b/hack/gcr-push.sh
@@ -47,23 +47,33 @@ function highest_release() {
     done
 }
 
-if [ "$BRANCH" == "master" ]; then
+if [[ "$BRANCH" == "master" ]]; then
     VERSION="$BRANCH"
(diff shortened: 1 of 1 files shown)


hack/gcr-push.sh
```diff
@@ -47,24 +47,34 @@ function highest_release() {
     done
 }
 
-if [ "$BRANCH" == "master" ]; then
+if [[ "$BRANCH" == "master" ]]; then
```
carlisia: Did you mean to have double bracket here?

nrb: Yep. I should update the other branch below this, too.

PR #1930, opened 2019-10-02
Add --plugins flag to velero install

### LINKED ISSUE #1740: Add --plugins flag to `velero install` to allow plugins to be included at install time

nrb:
Fixes #1740

### DIFF
--- a/changelogs/unreleased/1930-nrb
+++ b/changelogs/unreleased/1930-nrb
@@ -1,0 +1,1 @@
+Add a new required --plugins flag for velero install command. --plugins takes a list of container images to add as initcontainers.
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -63,6 +63,7 @@ type InstallOptions struct {
 	Wait                              bool
 	UseVolumeSnapshots                bool
 	DefaultResticMaintenanceFrequency time.Duration
+	Plugins                           flag.StringArray
 }
 
 // BindFlags adds command line values to the options struct.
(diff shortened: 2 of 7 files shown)


pkg/cmd/cli/install/install.go
```diff
@@ -90,6 +91,7 @@ func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&o.UseRestic, "use-restic", o.UseRestic, "create restic deployment. Optional.")
 	flags.BoolVar(&o.Wait, "wait", o.Wait, "wait for Velero deployment to be ready. Optional.")
 	flags.DurationVar(&o.DefaultResticMaintenanceFrequency, "default-restic-prune-frequency", o.DefaultResticMaintenanceFrequency, "how often 'restic prune' is run for restic repositories by default. Optional.")
+	flags.Var(&o.Plugins, "plugins", "Plugin container images to install into the Velero Deployment. Optional.")
```
carlisia: Isn't this supposed to be required?

carlisia: Or should I change it to required on my PR that extracts the cloudproviders?

nrb: I can change it now if you'd like; I actually need to add a change log anyway.

It's a bit awkward either way at the moment - if it's required now, we basically have to add a dummy plugin image to get `velero install` to work on master, as the cloud provider plugins aren't present yet. 

If it's optional once the plugins are out of tree, then `velero install` could install a deployment without any object storage plugins present, which would mean it's nonfunctional.

carlisia: Let me change it to required as part of the PR that extracts the providers then. That will only go in once we have the plugin repos and their images.

carlisia:
Should any documentation be added now?

carlisia:
Or, would the documentation for this be this issue: https://github.com/vmware-tanzu/velero/pull/1884?

nrb:
I opened #1884 as an alternative to this. I think the documentation that makes sense for this PR is to document that the plugin flag exists and will install any specified container image into the deployment. Specific cloud provider instructions can be done in their respective PR(s).

skriss:
That makes sense to me re: documentation!

Will review the updated code shortly.

changelogs/unreleased/1930-nrb
```diff
@@ -0,0 +1 @@
+Add a new required --plugins flag for velero install command. --plugins takes a list of container images to add as initcontainers.
```
skriss: Technically it's optional now, but will be required by the time we release - I'm fine leaving it as-is given that it'll change, but wasn't sure if it was intentional or not.

carlisia: I'll change it to required when I merge the cloud provider extraction PR. I added a todo for that.

PR #1936, opened 2019-10-03
Allow backup storage locations to specify backup sync period or toggle off sync

### LINKED ISSUE #1934: Support backup storage locations to specify backup sync periods or toggle off backup sync
Velero has a backup-sync-period that is currently set to a default of 60 secs. This is the default backup sync period for all backup storage locations. Currently there's no way to control backup sync periods per backup storage location, or to toggle off sync for a given storage location.

Allow backup storage locations to specify a backup-sync-period and also ability to toggle off backup sync for a given BSL. This is useful when we want to control how frequently we want Velero to query the backup target. 

**Environment:**

- Velero version (use `velero version`): v1.1.0

betta1:
This resolves #1934


pkg/controller/backup_sync_controller.go
```diff
@@ -37,6 +38,11 @@ import (
 	"github.com/vmware-tanzu/velero/pkg/plugin/clientmgmt"
 )
 
+const (
+	backupSyncPeriodKey = "backupSyncPeriod"
+	skipBackupSyncKey   = "skipBackupSync"
```
nrb: Would it make sense to use `backupSyncPeriod: 0` be a sentinel value for skipping sync, so we only have one variable to check?

betta1: Yes, that makes sense. I had initially thought of doing it that way but figured adding a config variable for skipping sync would make it more explicit. I'll update the PR.

nrb: That's certainly true that it's more explicit, which tends to be good. Curious what others think, though.

betta1: I have the changes to use backupSyncPeriod: 0 as a sentinel value for skipping sync all ready, just waiting for input from other reviewers then will update the PR.

pkg/controller/backup_sync_controller.go
```diff
@@ -134,6 +136,32 @@ func (c *backupSyncController) run() {
 
 	for _, location := range locations {
 		log := c.logger.WithField("backupLocation", location.Name)
+
+		if backupSyncPeriod, ok := location.Spec.Config[backupSyncPeriodKey]; ok {
```
skriss: I think if we're adding this to BackupStorageLocations, it makes more sense to be a first-class `Spec` field (optional) than an opaque config key, since it's not provider-specific. It would also mean it could be strongly typed as a duration field. Thoughts?

pkg/controller/backup_sync_controller.go
```diff
@@ -134,6 +136,32 @@ func (c *backupSyncController) run() {
...
+			lastSync := location.Status.LastSyncedTime
+			nextSync := lastSync.Add(syncDuration)
+			if time.Now().UTC().Before(nextSync) {
+				c.logger.WithFields(logrus.Fields{
+					"backupLocation": location.Name,
+					"lastSync":       lastSync,
+					"nextSync":       nextSync,
+				}).Debug("Sync not required at this time")
+				continue
+			}
```
skriss: The relationship between this logic and the `--backup-sync-period` could produce some unintuitive behavior in certain instances. For example, if `--backup-sync-period` is set to `1h`, but a specific BSL has a value of `1m`, that BSL will still only get synced every `1h`, because the `run()` function that this check is in will only be run every hour.

I would like to do a little more thinking on how this should work. One option would be to have the `run()` function run frequently (i.e. every 30s? minute? TBD), and each time it runs, check each BSL to see if it should be synced, based on its last-synced time, and either its specific sync interval, or the default value (as provided via `--backup-sync-period`). 

I am also curious - are you mainly interested in being able to disable sync for a specific BSL, or is the custom interval equally important? What's the use case for the custom interval?

pkg/controller/backup_sync_controller.go
prydonius: Agreed, I did find this strange. `velero install` should also set the default value when creating a BSL

pkg/controller/backup_sync_controller.go
prydonius: Agreed, we should change this to the approach proposed by @skriss. The CronJob controller checks [every 10 seconds](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/cronjob/cronjob_controller.go#L98), but I think 30s/1m is also reasonable for this usecase.

betta1: @skriss @prydonius I updated the PR as the approach suggested, please take a look and let me know any edits I need to make. Sorry for the late update I was OOO towards end of last week.

skriss: no worries - I'll take a look soon.

pkg/controller/backup_sync_controller.go
```diff
@@ -134,6 +136,35 @@ func (c *backupSyncController) run() {
...
+		lastSync := location.Status.LastSyncedTime
+		if lastSync != nil {
+			log.Debug("Checking if backups need to be synced at this time for this location")
+			nextSync := lastSync.Add(syncPeriod)
+			if time.Now().UTC().Before(nextSync) {
+				c.logger.WithFields(logrus.Fields{
+					"backupLocation":   location.Name,
+					"backupSyncPeriod": syncPeriod,
+					"lastSync":         lastSync,
+					"nextSync":         nextSync}).Debug("Sync not required at this time")
```
prydonius: Is this log necessary? If the sync was configured to be every hour, then this would be printed every minute until the hour since the controller sync is every minute, right?

pkg/controller/backup_sync_controller.go
```diff
@@ -134,6 +136,35 @@ func (c *backupSyncController) run() {
...
+
+		syncPeriod := c.defaultBackupSyncPeriod
+		if location.Spec.BackupSyncPeriod != nil {
+			syncPeriod = location.Spec.BackupSyncPeriod.Duration
+			if syncPeriod == 0 {
+				log.Debug("Backup sync period for this location is set to 0, skipping sync")
+				continue
+			}
+
+			if syncPeriod < 0 {
```
prydonius: Perhaps we can add a validation in the API type to avoid the need for this check

pkg/apis/velero/v1/backup_storage_location.go
```diff
@@ -80,6 +80,11 @@ type BackupStorageLocationSpec struct {
 	// AccessMode defines the permissions for the backup storage location.
 	// +optional
 	AccessMode BackupStorageLocationAccessMode `json:"accessMode,omitempty"`
+
+	// BackupSyncPeriod defines how frequently to sync backup API objects from object storage.
+	// +optional
+	// +nullable
```
prydonius: Can we add a validation here?, e.g.

```
// +kubebuilder:validation:Minimum=0
```

Would that prevent negative values being set?

pkg/controller/backup_sync_controller.go
betta1: You're right, this will be too much logging. I removed the logs.

pkg/apis/velero/v1/backup_storage_location.go
betta1: I tried this, unfortunately looks like minimum validation only applies to integers -- build will otherwise fail with "must apply minimum to an integer". I agree though it would be great if we can validate this in the API spec.

betta1: @prydonius I poked around kubebuilder to see how I can accomplish validation for this in the API spec and so far not found much, please let me know if you have another suggestion and I'll look into it.

pkg/apis/velero/v1/backup_storage_location.go
```diff
@@ -80,6 +80,11 @@ type BackupStorageLocationSpec struct {
 	// AccessMode defines the permissions for the backup storage location.
 	// +optional
 	AccessMode BackupStorageLocationAccessMode `json:"accessMode,omitempty"`
+
+	// BackupSyncPeriod defines how frequently to sync backup API objects from object storage.
```
skriss: Probably worth mentioning that "0" means never.

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -78,6 +82,7 @@ func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Provider, "provider", o.Provider, "name of the backup storage provider (e.g. aws, azure, gcp)")
 	flags.StringVar(&o.Bucket, "bucket", o.Bucket, "name of the object storage bucket where backups should be stored")
 	flags.StringVar(&o.Prefix, "prefix", o.Prefix, "prefix under which all Velero data should be stored within the bucket. Optional.")
+	flags.DurationVar(&o.BackupSyncPeriod, "backup-sync-period", defaultBackupSyncPeriod, "how often to ensure all Velero backups in object storage exist as Backup API objects in the cluster. Optional. Set this to `0s` to skip sync")
```
skriss: maybe "disable sync" instead of "skip sync" would be slightly clearer

pkg/apis/velero/v1/backup_storage_location.go
prydonius: @betta1 ah sorry, I missed that this is not a basic integer field. I guess we could use a regex pattern with `+kubebuilder:validation:Pattern`, but I think that's going to be a little clunky. Was just a thought, we can just use the validation in the controller.

prydonius:
> I'm wondering if we want to make the hard-coded 1-minute resync period configurable? Prior to this change, a user could set it to e.g. 1s if they wanted; now, 1 minute is effectively the minimum interval for syncing a BSL. I'm also wary of adding too many flags, so interested in others' thoughts here. We could also just lower that hardcoded value (30s? 10s?), which should reduce the need for customizing it.


Adding a flag for it is going to be a little confusing IMO (users might be confused about whether they should set that or set the sync period in the BSL). I think we should just reduce it to 10s or 30s. I don't really expect anyone will want to sync this faster than 10s, and especially since CronJob has this limitation also.

betta1:
> > I'm wondering if we want to make the hard-coded 1-minute resync period configurable? Prior to this change, a user could set it to e.g. 1s if they wanted; now, 1 minute is effectively the minimum interval for syncing a BSL. I'm also wary of adding too many flags, so interested in others' thoughts here. We could also just lower that hardcoded value (30s? 10s?), which should reduce the need for customizing it.
> 
> Adding a flag for it is going to be a little confusing IMO (users might be confused about whether they should set that or set the sync period in the BSL). I think we should just reduce it to 10s or 30s. I don't really expect anyone will want to sync this faster than 10s, and especially since CronJob has this limitation also.

I agree, I think just lowering the hardcoded value for the resync period should be sufficient for now (probably we can create an issue to track exposing this as a configurable flag later if users are interested). I've lowered it to 30s as I don't want this change to cause Velero to consume more cycles/resources just running the resync process too frequently.

pkg/apis/velero/v1/backup_storage_location.go
betta1: I missed this, added 0 means disabling sync.

pkg/cmd/cli/backuplocation/create.go
betta1: Agreed, fixed it to "disable sync" as it's more clear.

pkg/apis/velero/v1/backup_storage_location.go
betta1: Sounds good -- thanks for pointing me to kubebuilder, looks like a powerful tool for managing CRDs.

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -78,6 +82,7 @@ func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Provider, "provider", o.Provider, "name of the backup storage provider (e.g. aws, azure, gcp)")
 	flags.StringVar(&o.Bucket, "bucket", o.Bucket, "name of the object storage bucket where backups should be stored")
 	flags.StringVar(&o.Prefix, "prefix", o.Prefix, "prefix under which all Velero data should be stored within the bucket. Optional.")
+	flags.DurationVar(&o.BackupSyncPeriod, "backup-sync-period", defaultBackupSyncPeriod, "how often to ensure all Velero backups in object storage exist as Backup API objects in the cluster. Optional. Set this to `0s` to disable sync")
```
skriss: Just wondering if we want to not set an explicit value for the BSL if the user doesn't specify the flag, rather than defaulting it to a minute here? If we did that, then the server-level setting would apply. The benefit is that if the user then changes the server-level setting, it would affect all BSLs that were not set up with an explicit sync period, whereas now, they're all getting an individual setting of 1m by default.

betta1: Yes, this would actually be ideal. I'd like to make this change and the only issue i'm running into is how can I detect that the user didn't specify the flag? Atm when I set the default value to o.BackupSyncPeriod, like so:

` flags.DurationVar(&o.BackupSyncPeriod, "backup-sync-period", o.BackupSyncPeriod, "how often..."`
it gets parsed as 0s which is an issue since this means disable sync. Any suggestion on how to detect this flag wasn't provided?

skriss: I *think* you can use this method: https://godoc.org/github.com/spf13/pflag#FlagSet.Changed

but I'm not 100% sure off the top of my head, would need to experiment

skriss:
otherwise LGTM!

pkg/cmd/cli/backuplocation/create.go
betta1: Thanks @skriss, I modified to check if the user explicitly specified a backup sync period and only then set the backup sync period for the BSL. Otherwise if user does not specify the flag the BSL backup sync period is nil which will have the BSL use the server side backup sync.

PR #1937, opened 2019-10-03
Wait for CRDs to be ready before restoring CRs

### LINKED ISSUE #964: Wait for CRDs to be ready when restoring

nrb:
Fix #964

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -455,6 +455,7 @@ func (s *server) validateBackupStorageLocations() error {
 // - Custom Resource Definitions come before Custom Resource so that they can be
 //   restored with their corresponding CRD.
 var defaultRestorePriorities = []string{
+	"customresourcedefinitions",
 	"namespaces",
 	"storageclasses",
 	"persistentvolumes",
@@ -469,7 +470,6 @@ var defaultRestorePriorities = []string{
 	// to ensure that we prioritize restoring from "apps" too, since this is how they're stored
 	// in the backup.
 	"replicasets.apps",
-	"customresourcedefinitions",
 }
 
 func (s *server) initRestic() error {
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -440,7 +440,9 @@ func (s *server) validateBackupStorageLocations() error {
 	return nil
 }
 
-// - Namespaces go first because all namespaced resources depend on them.
+// - Custom Resource Definitions come before Custom Resource so that they can be
+//   restored with their corresponding CRD.
+// - Namespaces go second because all namespaced resources depend on them.
 // - Storage Classes are needed to create PVs and PVCs correctly.
 // - PVs go before PVCs because PVCs depend on them.
 // - PVCs go before pods or controllers so they can be mounted as volumes.
@@ -452,8 +454,6 @@ func (s *server) validateBackupStorageLocations() error {
 //	 have restic restores run before controllers adopt the pods.
 // - Replica sets go before deployments/other controllers so they can be explicitly
 //	 restored and be adopted by controllers.
-// - Custom Resource Definitions come before Custom Resource so that they can be
-//   restored with their corresponding CRD.
 var defaultRestorePriorities = []string{
 	"customresourcedefinitions",
 	"namespaces",
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -461,9 +461,13 @@ func (ctx *context) execute() (Result, Result) {
 		}
 	}
 
+	// TODO: Re-order this logic so that CRs can be prioritized in the main loop, rather than after.
+
+	// Refresh and resolve based on CRDs added to the API server from the above restore loop.
 	ctx.discoveryHelper.Refresh()
 	newResources, _ := prioritizeResources(ctx.discoveryHelper, ctx.resourcePriorities, ctx.resourceIncludesExcludes, ctx.log)
 
+	// Filter the resources to only those added since our first restore pass.
 	addedResources := make([]schema.GroupResource, 0)
 	for _, r := range newResources {
 		var found bool
@@ -472,18 +476,14 @@ func (ctx *context) execute() (Result, Result) {
 				found = true
 			}
 		}
+		// Resource hasn't already been processed, so queue it for the next loop.
 		if !found {
 			addedResources = append(addedResources, r)
 		}
 	}
 
+	// Use the same restore logic as above, but for newly available API groups
 	for _, resource := range addedResources {
-		// we don't want to explicitly restore namespace API objs because we'll handle
-		// them as a special case prior to restoring anything into them
-		if resource == kuberesource.Namespaces {
-			continue
-		}
-
 		resourceList := backupResources[resource.String()]
 		if resourceList == nil {
 			continue
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -32,6 +32,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
+	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -672,6 +673,40 @@ func (ctx *context) shouldRestore(name string, pvClient client.Dynamic) (bool, e
 	return shouldRestore, err
 }
 
+// crdAvailable waits for a CRD to be available for use before letting the restore continue.
+func (ctx *context) crdAvailable(name string, crdClient client.Dynamic) (bool, error) {
+	crdLogger := ctx.log.WithField("crdName", name)
+
+	var available bool
+	err := wait.PollImmediate(time.Second, ctx.resourceTerminatingTimeout, func() (bool, error) {
+		unstructuredCRD, err := crdClient.Get(name, metav1.GetOptions{})
+		if err != nil {
+			return true, err
+		}
+
+		clusterCRD := new(apiextv1beta1.CustomResourceDefinition)
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredCRD.Object, clusterCRD); err != nil {
+			return true, err
+		}
+
+		available = kube.IsCRDReady(clusterCRD)
+
+		if !available {
+			crdLogger.Debug("CRD not yet ready for use")
+		}
+
+		// If the CRD is not available, keep polling (false, nil)
+		// If the CRD is available, break the poll and return back to caller (true, nil)
+		return available, nil
+	})
+
+	if err == wait.ErrWaitTimeout {
+		crdLogger.Debug("timeout reached waiting for custom resource definition to be ready")
+	}
+
+	return available, err
+}
+
 // restoreResource restores the specified cluster or namespace scoped resource. If namespace is
 // empty we are restoring a cluster level resource, otherwise into the specified namespace.
 func (ctx *context) restoreResource(resource, targetNamespace, originalNamespace string, items []string) (Result, Result) {
@@ -1111,6 +1146,18 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
 
 	if groupResource == kuberesource.Pods && len(restic.GetVolumeBackupsForPod(ctx.podVolumeBackups, obj)) > 0 {
 		restorePodVolumeBackups(ctx, createdObj, originalNamespace)
+	}
+
+	// Wait for a CRD to be available for instantiating resources
+	// before continuing.
+	if groupResource == kuberesource.CustomResourceDefinitions {
+		available, err := ctx.crdAvailable(name, resourceClient)
+		if err != nil {
+			addToResult(&errs, namespace, errors.Wrapf(err, "error verifying custom resource definition is ready to use"))
+		}
+		if !available {
+			addToResult(&errs, namespace, fmt.Errorf("CRD %s is not available to use for custom resources.", name))
+		}
 	}
 
 	return warnings, errs
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -466,12 +466,16 @@ func (ctx *context) execute() (Result, Result) {
 	// This is because CRDs have been added to the API groups but until we refresh, Velero doesn't know about the
 	// newly-added API groups in order to create the CRs from them.
 	if err := ctx.discoveryHelper.Refresh(); err != nil {
-		addVeleroError(&errs, errors.Wrap(err, "error refreshing discovery API"))
+		// Don't break on error here, since newResources will be the same as the original prioritizedResources,
+		// and thus addedResources will end up being empty and we'll restore nothing.
+		// Since we're continuing the restore, add a warning, not an error.
+		addVeleroError(&warnings, errors.Wrap(err, "error refreshing discovery API"))
 	}
 	newResources, err := prioritizeResources(ctx.discoveryHelper, ctx.resourcePriorities, ctx.resourceIncludesExcludes, ctx.log)
 	if err != nil {
-		// Is this right?
-		addVeleroError(&errs, errors.Wrap(err, "error sorting resources"))
+		// If there was an error, then newResources will be nil, so we can continue on the restore.
+		// addedResources will end up being nil, but we should still report this failure.
+		addVeleroError(&warnings, errors.Wrap(err, "error sorting resources"))
 	}
 
 	// Filter the resources to only those added since our first restore pass.
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -459,10 +459,69 @@ func (ctx *context) execute() (Result, Result) {
 			merge(&warnings, &w)
 			merge(&errs, &e)
 		}
+	}
 
-		if resource == kuberesource.CustomResourceDefinitions {
-			ctx.discoveryHelper.Refresh()
-			ctx.prioritizedResources, _ = prioritizeResources(ctx.discoveryHelper, ctx.resourcePriorities, ctx.resourceIncludesExcludes, ctx.log)
+	ctx.discoveryHelper.Refresh()
+	newResources, _ := prioritizeResources(ctx.discoveryHelper, ctx.resourcePriorities, ctx.resourceIncludesExcludes, ctx.log)
+
+	addedResources := make([]schema.GroupResource, 0)
+	for _, r := range newResources {
+		var found bool
+		for _, p := range ctx.prioritizedResources {
+			if r == p {
+				found = true
+			}
+		}
+		if !found {
+			addedResources = append(addedResources, r)
+		}
+	}
+
+	for _, resource := range addedResources {
+		// we don't want to explicitly restore namespace API objs because we'll handle
+		// them as a special case prior to restoring anything into them
+		if resource == kuberesource.Namespaces {
+			continue
+		}
+
+		resourceList := backupResources[resource.String()]
+		if resourceList == nil {
+			continue
+		}
+
+		for namespace, items := range resourceList.ItemsByNamespace {
+			if namespace != "" && !ctx.namespaceIncludesExcludes.ShouldInclude(namespace) {
+				ctx.log.Infof("Skipping namespace %s", namespace)
+				continue
+			}
+
+			// get target namespace to restore into, if different
+			// from source namespace
+			targetNamespace := namespace
+			if target, ok := ctx.restore.Spec.NamespaceMapping[namespace]; ok {
+				targetNamespace = target
+			}
+
+			// if we don't know whether this namespace exists yet, attempt to create
+			// it in order to ensure it exists. Try to get it from the backup tarball
+			// (in order to get any backed-up metadata), but if we don't find it there,
+			// create a blank one.
+			if namespace != "" && !existingNamespaces.Has(targetNamespace) {
+				logger := ctx.log.WithField("namespace", namespace)
+				ns := getNamespace(logger, getItemFilePath(ctx.restoreDir, "namespaces", "", namespace), targetNamespace)
+				if _, err := kube.EnsureNamespaceExistsAndIsReady(ns, ctx.namespaceClient, ctx.resourceTerminatingTimeout); err != nil {
+					addVeleroError(&errs, err)
+					continue
+				}
+
+				// keep track of namespaces that we know exist so we don't
+				// have to try to create them multiple times
+				existingNamespaces.Insert(targetNamespace)
+			}
+
+			w, e := ctx.restoreResource(resource.String(), targetNamespace, namespace, items)
+			merge(&warnings, &w)
+			merge(&errs, &e)
 		}
 	}
 
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -285,6 +285,8 @@ func (kr *kubernetesRestorer) Restore(
 		restoredItems:              make(map[velero.ResourceIdentifier]struct{}),
 		renamedPVs:                 make(map[string]string),
 		pvRenamer:                  kr.pvRenamer,
+		discoveryHelper:            kr.discoveryHelper,
+		resourcePriorities:         kr.resourcePriorities,
 	}
 
 	return restoreCtx.execute()
@@ -378,6 +380,8 @@ type context struct {
 	restoredItems              map[velero.ResourceIdentifier]struct{}
 	renamedPVs                 map[string]string
 	pvRenamer                  func(string) (string, error)
+	discoveryHelper            discovery.Helper
+	resourcePriorities         []string
 }
 
 type resourceClientKey struct {
@@ -454,6 +458,11 @@ func (ctx *context) execute() (Result, Result) {
 			w, e := ctx.restoreResource(resource.String(), targetNamespace, namespace, items)
 			merge(&warnings, &w)
 			merge(&errs, &e)
+		}
+
+		if resource == kuberesource.CustomResourceDefinitions {
+			ctx.discoveryHelper.Refresh()
+			ctx.prioritizedResources, _ = prioritizeResources(ctx.discoveryHelper, ctx.resourcePriorities, ctx.resourceIncludesExcludes, ctx.log)
 		}
 	}
 
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -32,7 +32,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
-	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -763,12 +762,13 @@ func (ctx *context) crdAvailable(name string, crdClient client.Dynamic) (bool, e
 			return true, err
 		}
 
-		clusterCRD := new(apiextv1beta1.CustomResourceDefinition)
-		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredCRD.Object, clusterCRD); err != nil {
+		// TODO: Due to upstream conversion issues in runtime.FromUnstructured, we use the unstructured object here.
+		// Once the upstream conversion functions are fixed, we should convert to the CRD types and use IsCRDReady
+		available, err = kube.IsUnstructuredCRDReady(unstructuredCRD)
+
+		if err != nil {
 			return true, err
 		}
-
-		available = kube.IsCRDReady(clusterCRD)
 
 		if !available {
 			crdLogger.Debug("CRD not yet ready for use")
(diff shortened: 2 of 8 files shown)


skriss:
@nrb can you remind me what the status is here? Don't want to leave this one in limbo forever - if it's ready for review, I'll do so.

nrb:
@skriss This is on me - I wanted to try this with more than 1 CRD/CR pair and haven't circled back to it. I had intended to perhaps make those pairs myself, but may just set up a cluster with something like Istio and verify the order via logs.

nrb:
Per attached log (in JSON so I could filter it more easily), it appears that this is working as intended with an istio install - the CRDs are ordered first, and they're all created before moving on. Some of the CRs are recreated by Istio before we can get to them, but that's sometimes the case with controllers anyway.

```
{"level":"info","logSource":"pkg/restore/restore.go:1051","msg":"Attempting to restore CustomResourceDefinition: attributemanifests.config.istio.io","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:13-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1108","msg":"Restore of CustomResourceDefinition, attributemanifests.config.istio.io skipped: it already exists in the cluster and is the same as the backed up version","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:13-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1051","msg":"Attempting to restore CustomResourceDefinition: destinationrules.networking.istio.io","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:13-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1108","msg":"Restore of CustomResourceDefinition, destinationrules.networking.istio.io skipped: it already exists in the cluster and is the same as the backed up version","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:13-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1051","msg":"Attempting to restore CustomResourceDefinition: edges.config.istio.io","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:13-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1108","msg":"Restore of CustomResourceDefinition, edges.config.istio.io skipped: it already exists in the cluster and is the same as the backed up version","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1051","msg":"Attempting to restore CustomResourceDefinition: handlers.config.istio.io","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1108","msg":"Restore of CustomResourceDefinition, handlers.config.istio.io skipped: it already exists in the cluster and is the same as the backed up version","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1051","msg":"Attempting to restore CustomResourceDefinition: kuberneteses.config.istio.io","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1108","msg":"Restore of CustomResourceDefinition, kuberneteses.config.istio.io skipped: it already exists in the cluster and is the same as the backed up version","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1051","msg":"Attempting to restore CustomResourceDefinition: logentries.config.istio.io","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1108","msg":"Restore of CustomResourceDefinition, logentries.config.istio.io skipped: it already exists in the cluster and is the same as the backed up version","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1051","msg":"Attempting to restore CustomResourceDefinition: metrics.config.istio.io","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1108","msg":"Restore of CustomResourceDefinition, metrics.config.istio.io skipped: it already exists in the cluster and is the same as the backed up version","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1051","msg":"Attempting to restore CustomResourceDefinition: rules.config.istio.io","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1108","msg":"Restore of CustomResourceDefinition, rules.config.istio.io skipped: it already exists in the cluster and is the same as the backed up version","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1051","msg":"Attempting to restore CustomResourceDefinition: stackdrivers.config.istio.io","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1108","msg":"Restore of CustomResourceDefinition, stackdrivers.config.istio.io skipped: it already exists in the cluster and is the same as the backed up version","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1051","msg":"Attempting to restore CustomResourceDefinition: tracespans.config.istio.io","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:1108","msg":"Restore of CustomResourceDefinition, tracespans.config.istio.io skipped: it already exists in the cluster and is the same as the backed up version","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:702","msg":"Restoring resource 'attributemanifests.config.istio.io' into namespace 'istio-system'","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:746","msg":"Getting client for config.istio.io/v1alpha2, Kind=attributemanifest","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:14-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:702","msg":"Restoring resource 'destinationrules.networking.istio.io' into namespace 'istio-system'","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:18-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:746","msg":"Getting client for networking.istio.io/v1alpha3, Kind=DestinationRule","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:18-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:702","msg":"Restoring resource 'edges.config.istio.io' into namespace 'istio-system'","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:18-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:746","msg":"Getting client for config.istio.io/v1alpha2, Kind=edge","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:18-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:702","msg":"Restoring resource 'handlers.config.istio.io' into namespace 'istio-system'","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:18-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:746","msg":"Getting client for config.istio.io/v1alpha2, Kind=handler","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:18-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:702","msg":"Restoring resource 'kuberneteses.config.istio.io' into namespace 'istio-system'","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:19-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:746","msg":"Getting client for config.istio.io/v1alpha2, Kind=kubernetes","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:19-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:702","msg":"Restoring resource 'logentries.config.istio.io' into namespace 'istio-system'","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:19-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:746","msg":"Getting client for config.istio.io/v1alpha2, Kind=logentry","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:19-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:702","msg":"Restoring resource 'metrics.config.istio.io' into namespace 'istio-system'","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:19-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:746","msg":"Getting client for config.istio.io/v1alpha2, Kind=metric","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:19-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:702","msg":"Restoring resource 'rules.config.istio.io' into namespace 'istio-system'","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:21-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:746","msg":"Getting client for config.istio.io/v1alpha2, Kind=rule","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:21-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:702","msg":"Restoring resource 'stackdrivers.config.istio.io' into namespace 'istio-system'","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:22-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:746","msg":"Getting client for config.istio.io/v1alpha2, Kind=stackdriver","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:22-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:702","msg":"Restoring resource 'tracespans.config.istio.io' into namespace 'istio-system'","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:23-05:00"}
{"level":"info","logSource":"pkg/restore/restore.go:746","msg":"Getting client for config.istio.io/v1alpha2, Kind=tracespan","restore":"velero/istio1-20191212180305","time":"2019-12-12T18:03:23-05:00"}
```

pkg/util/kube/utils.go
```diff
@@ -135,3 +136,18 @@ func GetVolumeDirectory(pod *corev1api.Pod, volumeName string, pvcLister corev1l
 
 	return pvc.Spec.VolumeName, nil
 }
+
+// IsCRDReady checks a CRD to see if it's ready, so that objects may be created from it.
```
carlisia: I would remove part of the doc `so that objects may be created from it`. If it helps, this could go on the calling side of the function, but not here, since this function could be used for other purposes as well.

carlisia:
Also, I think this one could use a changelog.

pkg/util/kube/utils.go
nrb: Why would you remove that part? The whole purpose of its existence is to wait for it to be in a state such that CRs can be created from it.

pkg/util/kube/utils.go
```diff
@@ -135,3 +136,18 @@ func GetVolumeDirectory(pod *corev1api.Pod, volumeName string, pvcLister corev1l
...
+func IsCRDReady(crd *apiextv1beta1.CustomResourceDefinition) bool {
+	var isEstablished, namesAccepted bool
+	for _, cond := range crd.Status.Conditions {
+		if cond.Type == apiextv1beta1.Established {
+			isEstablished = true
+		}
+		if cond.Type == apiextv1beta1.NamesAccepted {
```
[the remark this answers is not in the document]
nrb: I don't believe so - per the documentation in the source `Status` is optional, while `Type` is not.

```
	// Type is the type of the condition. Types include Established, NamesAccepted and Terminating.
	Type CustomResourceDefinitionConditionType `json:"type" protobuf:"bytes,1,opt,name=type,casttype=CustomResourceDefinitionConditionType"`
...

I think the `cond.Status == true` condition does need to be checked though, good call.
	// Unique, one-word, CamelCase reason for the condition's last transition.
	// +optional
	Reason string `json:"reason,omitempty" protobuf:"bytes,4,opt,name=reason"`
```

skriss:
:+1: on the queue to review today

pkg/util/kube/utils.go
carlisia: I explained above, `since this function could be used for other purposes as well`. It was a suggestion. I personally prefer to document what the function does but not how or where or why it will be used, when it makes sense.

pkg/util/kube/utils.go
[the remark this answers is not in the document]
carlisia: > I don't believe so

Yeah, you're right!

pkg/restore/restore.go
```diff
@@ -1085,6 +1120,18 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
...
 	}
 
+	// Wait for a CRD to be available for instantiating resources
+	// before continuing.
+	if groupResource == kuberesource.CustomResourceDefinitions {
+		available, err := ctx.crdAvailable(name, resourceClient)
+		if err != nil {
+			addToResult(&errs, namespace, errors.Wrapf(err, "error verifying custom resource definition is ready to use"))
+		}
+		if !available {
```
skriss: in the case where `crdAvailable` returns false + a non-nil error, we'll end up adding two errors to the result here - I think we should probably just log the error, WDYT?

pkg/util/kube/utils.go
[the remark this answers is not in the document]
skriss: yeah, I would think checking for `cond.Status == true` would be a good idea as well

pkg/restore/restore.go
```diff
@@ -653,6 +654,40 @@ func (ctx *context) shouldRestore(name string, pvClient client.Dynamic) (bool, e
 	return shouldRestore, err
 }
 
+// crdAvailable waits for a CRD to be available for use before letting the restore continue.
+func (ctx *context) crdAvailable(name string, crdClient client.Dynamic) (bool, error) {
+	crdLogger := ctx.log.WithField("crdName", name)
+
+	var available bool
+	err := wait.PollImmediate(time.Second, ctx.resourceTerminatingTimeout, func() (bool, error) {
```
skriss: `resourceTerminatingTimeout` is 10min by default - wondering if maybe we want to use a shorter timeout here? I'm not sure without looking at some examples.

pkg/util/kube/utils.go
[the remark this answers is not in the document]
nrb: Updated and tests added.

pkg/restore/restore.go
nrb: Yeah, that time out is the same one we use for PVs, and it's probably too long for CRs. I wanted to use the same one so that we could allow users to customize without providing too many knobs.

pkg/restore/restore.go
nrb: I'd still like to communicate to users if, somehow, the CRD wasn't ready but there were no errors, so I made that statement an else if rather than a plain if.

nrb:
@mauilion has given me a scenario I think is related to this, but may also have more going on. I'm going to trouble shoot that just a bit more before I conclude that it's completely separate.

nrb:
As part of researching #2159, I think this PR is incomplete. While it is necessary, we also need to refresh our list of API server clients once the CRDs are added. Currently, Velero grabs a list of all available APIs before the restore happens, but does not update that list to account for any groups added via CRDs.

Thus, even with this waiting, CRs won't necessarily get restored on the same restore action as the CRD. However, they will get restored succesfully on a second run. @skriss and I are discussing solutions to this now, which I think we'll roll into this PR to make it a complete solution.

nrb:
The latest change is not the actual fix, but it demonstrates roughly what needs to happen.

I think the approach that I'll actually take is to reorganize the code so that the backup tarball is inspected and the resource list is built and sorted based on the resources in it. Then, the restore loop will act on each resource listed, getting the actual resource names from the kubernetes API server inside the loop before doing the restore, and refreshing the discovery helper after each resource is restored.

There's likely optimizations that can be done here, especially given that CRDs are the only class of resources that I'm aware of that can affect discovery, but the plan listed is the general approach I'd like to take.

sseago:
Here's the CRD definition which failed for me with the conversion error:
```
apiVersion: "apiextensions.k8s.io/v1beta1"
kind: "CustomResourceDefinition"
metadata:
  name: "foos.example.foo.com"
spec:
  group: "example.foo.com"
  version: "v1alpha1"
  scope: "Namespaced"
  names:
    plural: "foos"
    singular: "foo"
    kind: "Foo"
  validation:
    openAPIV3Schema:
      required: ["spec"]
      properties:
        spec:
          required: ["bar"]
          properties:
            bar:
              type: "integer"
              minimum: 1
```

sseago:
See https://github.com/helm/helm/issues/5806 for the same bug in a very similar use case.

pkg/cmd/server/server.go
```diff
@@ -455,6 +455,7 @@ func (s *server) validateBackupStorageLocations() error {
 // - Custom Resource Definitions come before Custom Resource so that they can be
```
skriss: let's move the comment on CRDs to the top so it remains consistent with the order of the slice

pkg/restore/restore.go
```diff
@@ -454,6 +459,70 @@ func (ctx *context) execute() (Result, Result) {
 		}
 	}
 
+	// TODO: Re-order this logic so that CRs can be prioritized in the main loop, rather than after.
+
+	// Refresh and resolve based on CRDs added to the API server from the above restore loop.
+	ctx.discoveryHelper.Refresh()
```
skriss: add error-check

pkg/restore/restore.go
```diff
@@ -454,6 +459,70 @@ func (ctx *context) execute() (Result, Result) {
...
+	newResources, _ := prioritizeResources(ctx.discoveryHelper, ctx.resourcePriorities, ctx.resourceIncludesExcludes, ctx.log)
```
skriss: add error-check

pkg/restore/restore.go
```diff
@@ -454,6 +459,70 @@ func (ctx *context) execute() (Result, Result) {
...
+
+	// Filter the resources to only those added since our first restore pass.
+	addedResources := make([]schema.GroupResource, 0)
+	for _, r := range newResources {
+		var found bool
+		for _, p := range ctx.prioritizedResources {
+			if r == p {
+				found = true
```
skriss: can break the inner loop at this point

pkg/restore/restore.go
```diff
@@ -454,6 +459,70 @@ func (ctx *context) execute() (Result, Result) {
 		}
 	}
 
+	// TODO: Re-order this logic so that CRs can be prioritized in the main loop, rather than after.
+
+	// Refresh and resolve based on CRDs added to the API server from the above restore loop.
```
skriss: Let's add a more extensive comment here explaining what's going on for the benefit of future us :)

pkg/restore/restore.go
nrb: Thanks, I hadn't had one since it was a quick hack, but if it's going in for the point release, definitely need it.

nrb:
> The main downside I see is that CR instances of CRDs that are being restored will always be restored last -- there's no way for the user to prioritize them differently. This could be an issue in specific scenarios where the user needs better control over ordering, but I don't see that being common.

Yeah, we need this for CSI restores, so it becomes a v1.3 item. But for 1.2.1, I think this is workable, pending more complete testing.

pkg/restore/restore.go
skriss: yep, i figured :)

nrb:
@sseago Can you file a separate issue for that conversion please? #2197 adds a new RestoreItemAction plugin that will do some other OpenAPI schema work, and I think that such conversion logic may fit better within that than here.

nrb:
Testing these latest changes in combination with #2197 on GKE v1.16 and Istio (this time deleting all CRDs, as well as testing the restore onto a completely new cluster) has been successful.

There are still issues with CRs not restoring because validating web hook endpoints might be missing, but I believe that is wholly separate. A sample log message for such an error looks like this:

```
time="2020-01-24T13:19:11-05:00" level=info msg="Attempting to restore rule: tcpkubeattrgenrulerule" logSource="pkg/restore/restore.go:1146" restore=velero/istio-test-20200124131808
time="2020-01-24T13:19:11-05:00" level=info msg="error restoring tcpkubeattrgenrulerule: Internal error occurred: failed calling webhook \"mixer.validation.istio.io\": Post https://istio-galley.istio-system.svc:443/admitmixer?timeout=30s: no endpoints available for service \"istio-galley\"" logSource="pkg/restore/restore.go:1209" restore=velero/istio-test-20200124131808
```

I'm going to rebase this PR, then remove the On Hold label.

nrb:
@skriss @carlisia @ashish-amarnath This is now ready. Note that for v1.16 clusters, this has to be combined with #2197 in order to work properly.

sseago:
@nrb I can create a separate issue for the conversion failure, but it's a bug that only exists with this PR merged, which means that this PR introduces a regression. In certain circumstances, this PR results in always-failing restores rather than CRs sometimes failing restore. As long as you're aware of that (and, hopefully, won't include this PR in a release without the fix to the bug introduced here), that's fine. It may well make sense to fix it separately from this PR.

nrb:
@sseago I see. Hmm...

I'll need to do more digging there. I wonder if this is a problem with the API machinery pieces, as `fromUnstructured` isn't Velero code, it's a dependency.

sseago:
@nrb Also, from the looks of it, #2197 seems to apply to Kubernetes 1.16+ only. The failure I hit on restore was with 1.14 (still using apiextensions.k8s.io/v1beta1). Also, since the problem seems to be within the `crdAvailable` function, I'm not sure how the new plugin will help resolve this.

sseago:
The conversion error is definitely happening within the APIMachinery pieces, since the exact same error is happening in the upsteram Helm issue I linked to above. What wasn't clear to me from earlier digging is what the right answer is. I don't know if there's an alternate way of converting this, or whether we'd need some custom error handling to recover from the conversion errors, or whether something completely different is happening. If you load the CRD definition above that was failing for me (and create a CR from it), you'll probably see the same thing I did. What reproduced for me was a backup created with Kubernetes 1.11 and the restore done into a 1.14 cluster, although I suspect you'll see the same thing with 1.14 for both backup and restore.

nrb:
@sseago Thanks for pushing back on the CRD issue. I've replicated it on v1.14.10 and v1.16.10 clusters.

Given how the error's handled, the CRD appears to get created on the cluster, and I'm able to create a CR from it, but I'm not 100% sure yet that it's valid for real cases. I'm going to file an upstream bug report with this information, as the linked Helm chart has a fix already documented which doesn't look too complicated.

We may be able to vendor in their fix, but I don't know how much extra code that would end up needing to pull in to work. I wouldn't be surprised if it ended up pulling in most of the `runtime` package.

nrb:
@skriss As Scott points out, due to https://github.com/helm/helm/issues/5806 for certain CRDs, this PR can introduce a regression. This may also be related to https://github.com/kubernetes/kubernetes/issues/77665, but I'm not 100% sure yet.

Regardless, the bug is in the upstream `runtime.FromUnstructured` function, so to fully address it we'll need to take it upstream to make sure it gets fixed there, whether we vendor a fix in the short term or not.

skriss:
OK thanks for the notes, I'll check the refs out

skriss:
We could probably also change our implementation to work around this issue, i.e. by not using the `FromUnstructured(...)` func and instead accessing the unstructured obj directly to determine if the CRD is ready. That might be most expedient, and we can add a TODO to replace once there's an upstream fix.

nrb:
@skriss Working on an update to this PR with that approach. Hope to have a commit for it next week.

nrb:
@sseago @skriss I've pushed a new commit that uses the upstream unstructured helpers against the sample CRD and no longer generates the int64/float64 conversion error on 1.14.10 or 1.16.4. Please give it a try and let me know if you see other issues.

skriss:
:+1: generally looks good to me, but I'll add a few review comments shortly

pkg/restore/restore.go
```diff
@@ -456,6 +460,80 @@ func (ctx *context) execute() (Result, Result) {
...
+	// This is because CRDs have been added to the API groups but until we refresh, Velero doesn't know about the
+	// newly-added API groups in order to create the CRs from them.
+	if err := ctx.discoveryHelper.Refresh(); err != nil {
+		addVeleroError(&errs, errors.Wrap(err, "error refreshing discovery API"))
```
skriss: since we're not breaking here if we get an error, we'll still fall through to the rest of the logic - I *think* that's fine, since `newResources` will end up being the same as `ctx.prioritizedResources` so nothing new will be restored - that sound right?

pkg/restore/restore.go
```diff
@@ -456,6 +460,80 @@ func (ctx *context) execute() (Result, Result) {
...
+	}
+	newResources, err := prioritizeResources(ctx.discoveryHelper, ctx.resourcePriorities, ctx.resourceIncludesExcludes, ctx.log)
+	if err != nil {
+		// Is this right?
+		addVeleroError(&errs, errors.Wrap(err, "error sorting resources"))
```
skriss: similar comment to previous - I *think* the fallthrough should be fine because `newResources` would be nil here, so nothing new will be restored - right?

Can remove the L473 comment too.

pkg/restore/restore.go
```diff
@@ -672,6 +750,42 @@ func (ctx *context) shouldRestore(name string, pvClient client.Dynamic) (bool, e
...
+
+	var available bool
+	// Wait 1 minute rather than the standard resource timeout, since each CRD will transition fairly quickly
+	err := wait.PollImmediate(time.Second, time.Minute*1, func() (bool, error) {
+		unstructuredCRD, err := crdClient.Get(name, metav1.GetOptions{})
+		if err != nil {
+			return true, err
+		}
+
+		// TODO: Due to upstream conversion issues in runtime.FromUnstructured, we use the unstructured object here.
```
skriss: is there an issue we can link to here for reference?

pkg/util/kube/utils.go
```diff
@@ -135,3 +137,72 @@ func GetVolumeDirectory(pod *corev1api.Pod, volumeName string, pvcLister corev1l
...
+		if cond.Type == apiextv1beta1.NamesAccepted && cond.Status == apiextv1beta1.ConditionTrue {
+			namesAccepted = true
+		}
+	}
+
+	return (isEstablished && namesAccepted)
+}
+
+// IsUnstructuredCRDReady checks an unstructured CRD to see if it's ready, with both the Established and NamesAccepted conditions.
+// TODO: Delete this function and use IsCRDReady when the upstream runtime.FromUnstructured function properly handles int64 field conversions.
```
skriss: would it be possible to add a unit test of the `runtime.FromUnstructured` func that currently passes, but would start failing when the issue is fixed? This probably isn't essential, but would give us an active signal for when we can remove this code.

skriss:
I tested:

- backup & restore on a 1.15 cluster
- backup on a 1.15 cluster, restoring onto a 1.16  cluster
- backup & restore on a 1.16 cluster 
    - of CRDs that were originally created as `v1beta1`, and had `preserveUnknownFields=true`
    - of CRDs that were originally created as `v1`

The only issue I encountered is the one I fixed in #2230, where on 1.16+ clusters, in some scenarios, CRDs would be backed up as `v1beta1` rather than the server-preferred version of `v1`.

pkg/restore/restore.go
nrb: I'll try to find one, or at least use the upstream one I mentioned in the comments here. I was hoping to file one that was a standalone program that took Scott's CRD as a file and produced an error, but I don't think I'll get to creating that this week.

pkg/restore/restore.go
nrb: Yeah, that should be correct. I'll add a comment explaining this.

I am going to update the error to be `errors.WithStack` instead of `errors.Wrap` since it's not a Velero error, too.

pkg/util/kube/utils.go
nrb: Probably, though I'm not sure that it would be directly on this function.

Let me play around with adding one.

pkg/restore/restore.go
skriss: AFAIK `errors.Wrap(...)` will add a stack trace too, so it should be fine as-is

nrb: Whoops, you're right. I'll add that back.

nrb:
Added a unit test that will fail once the fix is introduced upstream, and reported it upstream at https://github.com/kubernetes/kubernetes/issues/87675

Also created https://github.com/nrb/unstructured-conversion-bug with the failing test for minimal reproduction.

pkg/util/kube/utils.go
nrb: Added.

PR #1836, opened 2019-09-04
feat: add Azure China/German support

andyzhangx:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

This PR adds a new env var(`AZURE_CLOUD_NAME`) in azure environment config, available values could be: `AzurePublicCloud`, `AzureUSGovernmentCloud`, `AzureChinaCloud`, `AzureGermanCloud`

If env var(`AZURE_CLOUD_NAME`) is not set, `AzurePublicCloud` would be the default, so it's compatible.

**Which issue(s) this PR fixes**:

Fixes #287

**Special notes for your reviewer**:

**Release note**:
```
add azure china support
```

PR #1938, opened 2019-10-04
Azure china/germany support

### LINKED ISSUE #287: Support for storage in Azure Germany
Is it already possible to use a storage container in the German instance of Azure? I think the API endpoints are different for that cloud. I am running into an error when i try to setup Ark:

```
time="2018-01-17T12:31:33Z" level=info msg="Configuring cloud provider for backup service" logSource="pkg/cmd/server/server.go:330"
time="2018-01-17T12:31:34Z" level=info msg="Configuring cloud provider for snapshot service" logSource="pkg/cmd/server/server.go:346"
An error occurred: rpc error: code = Unknown desc = azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/6a6e8a67-b524-415c-a680-0b63e086c12d/locations?api-version=2016-06-01: StatusCode=0 -- Original Error: adal: Refresh request failed. Status Code = '400'
```

Is it possible to change the API endpoints used by Ark?

skriss:
Update of #1836 
Closes #287

### DIFF
--- a/pkg/cloudprovider/azure/object_store.go
+++ b/pkg/cloudprovider/azure/object_store.go
@@ -217,7 +217,7 @@ func (o *ObjectStore) Init(config map[string]string) error {
 	}
 
 	storageAccountKey, env, err := getStorageAccountKey(config)
-	if err != nil || env == nil {
+	if err != nil {
 		return err
 	}
 
--- a/pkg/cloudprovider/azure/object_store.go
+++ b/pkg/cloudprovider/azure/object_store.go
@@ -135,16 +135,21 @@ func NewObjectStore(logger logrus.FieldLogger) *ObjectStore {
 	return &ObjectStore{log: logger}
 }
 
-func getStorageAccountKey(config map[string]string) (string, error) {
+func getStorageAccountKey(config map[string]string) (string, *azure.Environment, error) {
 	// load environment vars from $AZURE_CREDENTIALS_FILE, if it exists
 	if err := loadEnv(); err != nil {
-		return "", err
+		return "", nil, err
 	}
 
-	// 1. we need AZURE_TENANT_ID, AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, AZURE_SUBSCRIPTION_ID
-	envVars, err := getRequiredValues(os.Getenv, tenantIDEnvVar, clientIDEnvVar, clientSecretEnvVar, subscriptionIDEnvVar)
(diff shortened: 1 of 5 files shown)


skriss:
I rebased #1836 and addressed the code review comments. PTAL.

skriss:
@nrb @carlisia @prydonius this one's ready for review.

pkg/cloudprovider/azure/object_store.go
```diff
@@ -135,46 +135,53 @@ func NewObjectStore(logger logrus.FieldLogger) *ObjectStore {
...
+func getStorageAccountKey(config map[string]string) (string, *azure.Environment, error) {
 	// load environment vars from $AZURE_CREDENTIALS_FILE, if it exists
 	if err := loadEnv(); err != nil {
-		return "", err
+		return "", nil, err
 	}
 
 	// 1. we need AZURE_TENANT_ID, AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, AZURE_SUBSCRIPTION_ID
-	envVars, err := getRequiredValues(os.Getenv, tenantIDEnvVar, clientIDEnvVar, clientSecretEnvVar, subscriptionIDEnvVar)
+	envVars, err := getRequiredValues(os.Getenv, tenantIDEnvVar, clientIDEnvVar, clientSecretEnvVar, subscriptionIDEnvVar, cloudNameEnvVar)
```
prydonius: doesn't this change need to be dropped, since this isn't required?

pkg/cloudprovider/azure/object_store.go
```diff
@@ -209,13 +216,13 @@ func (o *ObjectStore) Init(config map[string]string) error {
 		return err
 	}
 
-	storageAccountKey, err := getStorageAccountKey(config)
-	if err != nil {
+	storageAccountKey, env, err := getStorageAccountKey(config)
+	if err != nil || env == nil {
```
nrb: Given that this is a logic or, is it conceivable that this could return `nil` when there's no error or environment?

nrb: Looking at the possible return paths, I don't _think_ this could return `nil` because the there isn't a return statement where both the error and env are empty.

skriss: I think this should probably just change to be `if err != nil {` - let me update

PR #1951, opened 2019-10-08
use pointer types for metav1.Time fields

### LINKED ISSUE #1935: errors creating custom resources in v1.13.7 cluster using master version of velero
I'm using a locally-built version of the `velero` CLI built from the latest code in `master`. After installing into a `v1.13.7` cluster, I'm getting various errors when trying to create CRs via the CLI.

kubectl version:
```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T12:36:28Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.7-gke.8", GitCommit:"7d3d6f113e933ed1b44b78dff4baf649258415e5", GitTreeState:"clean", BuildDate:"2019-06-19T16:37:16Z", GoVersion:"go1.11.5b4", Compiler:"gc", Platform:"linux/amd64"}
```

Install command used:
```
./velero install --provider gcp --bucket <my-bucket> --prefix <my-prefix> --secret-file <my-credentials-file>
```

Backup error:
```
./velero backup create foo
An error occurred: Backup.velero.io "foo" is invalid: []: Invalid value: map[string]interface {}{"kind":"Backup", "apiVersion":"velero.io/v1", "metadata":map[string]interface {}{"name":"foo", "namespace":"velero", "creationTimestamp":"2019-10-03T19:13:26Z", "generation":1, "uid":"e08fcd1a-e611-11e9-a4bc-42010a8a0fc6"}, "spec":map[string]interface {}{"includedNamespaces":[]interface {}{"*"}, "ttl":"720h0m0s", "hooks":map[string]interface {}{}}, "status":map[string]interface {}{"expiration":interface {}(nil), "startTimestamp":interface {}(nil), "completionTimestamp":interface {}(nil)}}: validation failure list:
status.completionTimestamp in body must be of type string: "null"
status.startTimestamp in body must be of type string: "null"
status.expiration in body must be of type string: "null"
```

`velero version` error:
```
./velero version
Client:
	Version:
	Git commit: -
<error getting server version: ServerStatusRequest.velero.io "velero-cli-bxzmn" is invalid: []: Invalid value: map[string]interface {}{"metadata":map[string]interface {}{"namespace":"velero", "creationTimestamp":"2019-10-03T19:13:57Z", "generation":1, "uid":"f3252c72-e611-11e9-a4bc-42010a8a0fc6", "name":"velero-cli-bxzmn", "generateName":"velero-cli-"}, "spec":map[string]interface {}{}, "status":map[string]interface {}{"serverVersion":"", "plugins":interface {}(nil), "phase":"", "processedTimestamp":interface {}(nil)}, "kind":"ServerStatusRequest", "apiVersion":"velero.io/v1"}: validation failure list:
status.plugins in body must be of type array: "null"
status.processedTimestamp in body must be of type string: "null"
status.phase in body should be one of [New Processed]>
```

cc @prydonius

prydonius:
fixes #1935 


prydonius:
WIP - needs more fixing in places where we assume metav1.Time types are not nil, and manual testing

prydonius:
I believe I have addressed all the places where a `metav1.Time` type is assumed to be non-nil, except:

- [PVBController - getParentSnapshot](https://github.com/vmware-tanzu/velero/blob/d4361f0a0e1833e121c9b02f45c43a139ed8d857/pkg/controller/pod_volume_backup_controller.go#L319)
  - We only compare PVBs with phase Completed, all of these should have the StartTimestamp set unless it is manually removed
- [BackupController - recordBackupMetrics](https://github.com/vmware-tanzu/velero/blob/d4361f0a0e1833e121c9b02f45c43a139ed8d857/pkg/controller/backup_controller.go#L559)
  - We call this after a backup has completed, so it will always have the CompletionTimestamp, and should have the StartTimestamp unless it was manually removed

Let me know if we think it makes sense to guard the above two cases as well to prevent panicking if someone manually removes a timestamp field.

prydonius:
In the meantime, I'm continuing to do more manual testing here.

pkg/builder/backup_builder.go
```diff
@@ -154,13 +154,15 @@ func (b *BackupBuilder) TTL(ttl time.Duration) *BackupBuilder {
 
 // Expiration sets the Backup's expiration.
 func (b *BackupBuilder) Expiration(val time.Time) *BackupBuilder {
-	b.object.Status.Expiration.Time = val
+	expiration := metav1.NewTime(val)
+	b.object.Status.Expiration = &expiration
```
skriss: you _could_ save yourself a line & variable declaration on all of these by just doing e.g. 
```
b.object.Status.Expiration = &metav1.Time{Time: val}
```
Or defining a helper function somewhere that does the same.

We can merge without the change, though, given it's a nit/tedious - up to you.

skriss:
I went through the usual scenarios and all looked OK to me.

pkg/builder/backup_builder.go
prydonius: I wanted to use the factory method over just initialising the struct, in case some implementation changes in the constructor, but maybe it's not that important. I'm not sure what's more idiomatic in Go.

I could create a helper that uses the constructor and returns the pointer form though, to save all those lines. Where do you think that could go?

prydonius:
@skriss to confirm, this was on a v1.13 cluster?

pkg/builder/backup_builder.go
skriss: Fair enough. I'd say given that the type and field are exported, either is equally valid -- if the package author didn't want you to use them directly, presumably they wouldn't export them.

Hmm, always challenging to decide where to put little helper functions like this. We do have a `pkg/util/boolptr` which does something similar - could create a similar `timeptr` helper package, or `pkg/util/kube` would probably be fine too

skriss: Or, you could just implement it as an unexported func once in each package you use it - which looks to only be 2 or 3 places.

skriss:
I tried some scenarios on v1.13 and some on v1.14 - but definitely confirmed that the issues I was seeing on v1.13 are gone

prydonius:
I'm having trouble testing this on any version below v1.12, as both minikube and kind are currently failing provisioning clusters with older versions. This does work fine in v1.12 though.

prydonius:
@skriss @carlisia I think we should be okay to merge this, we can deal with any other issues as they come up.

pkg/builder/backup_builder.go
prydonius: > Fair enough. I'd say given that the type and field are exported, either is equally valid -- if the package author didn't want you to use them directly, presumably they wouldn't export them.

That's a good point, actually. I'm probably overthinking this and it probably isn't likely that the implementation will change too much since the intention is to just be a wrapper of time.Time. I'll go with initialising the struct directly to avoid having to find somewhere to put a helper. Thanks!

skriss:
I'll merge this now since others are on PTO so we can continue testing, can address any follow-ups as needed.

PR #1962, opened 2019-10-14
Update aws-sdk-go to support IAM Roles for Service Accounts

dcherman:
Ref: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html


skriss:
@dcherman thanks for this; you've caught us at an awkward time, as we're in the middle of migrating the in-tree storage provider code to their own repositories. We actually just got the separate AWS repository set up friday: https://github.com/vmware-tanzu/velero-plugin-for-aws

Would you mind making the SDK upgrade as a PR in that repository instead? Thanks!

dcherman:
@skriss It's actually not necessary in that repo it looks like, you're already pulling in a version of the AWS SDK that's new enough for IAM Role support already :)

Unless you feel there's value in merging this, can can probably just close it out and I'll look for the  first release of Velero where it's using that new AWS plugin.

Thanks for providing this project btw, it's fantastic!

skriss:
Ah, didn't realize we had bumped the version there. Sounds like we're set then - appreciate the input regardless! FYI we're planning to release v1.2 around the end of October which will include this update.

Closing this out.

Issue #80, opened 2017-09-12
Restore namespace remapping should adjust references to remapped namespaces


ncdc:
If a backup contains references to namespace `src` object `foo`, and we're using a namespace mapping to remap namespace `src` to `dst`, any items that reference `src` or `src/foo` probably should be adjusted to point at `dst` instead. Things I can think of:
- pv.spec.claimRef
- rolebinding.subjects[*].namespace
- clusterrolebinding.subjects[*].namespace

suggestion from https://twitter.com/timoreimann/status/907703134553853952

rosskukulinski:
This should probably be done around the same time as #192.  Both relate to cloning a namespace within a cluster.

skriss:
See #1954 for a specific instance of the role-bindings issue.

alexander-demichev:
@skriss I can add restore actions for cluster role bindings and role bindings that should swap namespace references.

skriss:
@alexander-demichev that sounds like a reasonable step to me, thanks!

PR #1974, opened 2019-10-17
Restore action for cluster/namespace role bindings

alexander-demichev:
Fixes https://github.com/vmware-tanzu/velero/issues/80

### DIFF
--- a/pkg/restore/clusterrolebinding_action.go
+++ b/pkg/restore/clusterrolebinding_action.go
@@ -1,0 +1,67 @@
+/*
+Copyright 2019 the Velero contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package restore
+
+import (
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	rbac "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+)
+
+// ClusterRoleBindingAction handle namespace remappings for role bindings
+type ClusterRoleBindingAction struct {
+	logger logrus.FieldLogger
+}
+
+func NewClusterRoleBindingAction(logger logrus.FieldLogger) *ClusterRoleBindingAction {
+	return &ClusterRoleBindingAction{logger: logger}
+}
+
+func (a *ClusterRoleBindingAction) AppliesTo() (velero.ResourceSelector, error) {
+	return velero.ResourceSelector{
+		IncludedResources: []string{"clusterrolebindings"},
+	}, nil
+}
+
+func (a *ClusterRoleBindingAction) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
+	namespaceMapping := input.Restore.Spec.NamespaceMapping
+	if len(namespaceMapping) == 0 {
+		return velero.NewRestoreItemActionExecuteOutput(&unstructured.Unstructured{Object: input.Item.UnstructuredContent()}), nil
+	}
+
+	clusterRoleBinding := new(rbac.ClusterRoleBinding)
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.Item.UnstructuredContent(), clusterRoleBinding); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	for i, subject := range clusterRoleBinding.Subjects {
+		if newNamespace, ok := namespaceMapping[subject.Namespace]; ok {
+			clusterRoleBinding.Subjects[i].Namespace = newNamespace
+		}
+	}
+
+	res, err := runtime.DefaultUnstructuredConverter.ToUnstructured(clusterRoleBinding)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	return velero.NewRestoreItemActionExecuteOutput(&unstructured.Unstructured{Object: res}), nil
+}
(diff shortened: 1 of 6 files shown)


pkg/restore/clusterrolebinding_action.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(namespaceMapping) == 0 {
+		return velero.NewRestoreItemActionExecuteOutput(&unstructured.Unstructured{Object: input.Item.UnstructuredContent()}), nil
+	}
+
+	clusterRoleBinding := new(rbac.ClusterRoleBinding)
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.Item.UnstructuredContent(), clusterRoleBinding); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	a.logger.Infoln(123456)
```
dymurray: What does this logger statement do? Seems odd.

pkg/restore/clusterrolebinding_action_test.go
```diff
@@ -0,0 +1,107 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, velero.ResourceSelector{IncludedResources: []string{"clusterrolebindings"}}, actual)
+}
+
+// 3 cases: no mapping - return rolebinding unchanged, mapping enabled change namespace
+func TestClusterRoleBindingActionExecute(t *testing.T) {
+	tests := []struct {
+		name             string
+		namespaces       []string
+		namespacemapping map[string]string
```
prydonius: naming nit:
```suggestion
		namespaceMapping map[string]string
```

pkg/restore/clusterrolebinding_action_test.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func TestClusterRoleBindingActionAppliesTo(t *testing.T) {
+	action := NewClusterRoleBindingAction(test.NewLogger())
+	actual, err := action.AppliesTo()
+	require.NoError(t, err)
+	assert.Equal(t, velero.ResourceSelector{IncludedResources: []string{"clusterrolebindings"}}, actual)
+}
+
+// 3 cases: no mapping - return rolebinding unchanged, mapping enabled change namespace
```
prydonius: 2 cases? I don't think we really need this comment :)

pkg/restore/clusterrolebinding_action_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}{
+		{
+			name:             "namespace mapping disabled",
+			namespaces:       []string{"foo"},
+			namespacemapping: map[string]string{},
+			expected:         []string{"foo"},
+		},
+		{
+			name:             "namespace mapping enabled",
+			namespaces:       []string{"foo"},
```
prydonius: It would be good to test that a namespace not included in the mapping is left unchanged.

pkg/restore/rolebinding_action.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (a *RoleBindingAction) AppliesTo() (velero.ResourceSelector, error) {
+	return velero.ResourceSelector{
+		IncludedResources: []string{"rolebindings"},
+	}, nil
+}
+
+func (a *RoleBindingAction) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
+	namespaceMapping := input.Restore.Spec.NamespaceMapping
```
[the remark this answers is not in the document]
alexander-demichev: If you have any ideas, I'll be happy to hear them.

alexander-demichev:
@prydonius @dymurray @skriss Thank you all for reviews! Sorry for debugging prints/comments, I should've had more coffee and sleep yesterday:)

pkg/restore/rolebinding_action.go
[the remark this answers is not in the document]
skriss: seems like L54-59 are really the only common ones - you could extract them to a helper func, like:

```go
func remapSubjectNamespaces(subjects []rbac.Subject, namespaceMappings map[string]string) []rbac.Subject {
    // ...
}
```

I'm fine either way, as it's only a couple lines of simple code.

pkg/restore/rolebinding_action_test.go
```diff
@@ -0,0 +1,106 @@
+/*
+Copyright 2018, 2019 the Velero contributors.
```
skriss: 2019

pkg/restore/rolebinding_action_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+func TestRoleBindingActionAppliesTo(t *testing.T) {
+	action := NewRoleBindingAction(test.NewLogger())
+	actual, err := action.AppliesTo()
+	require.NoError(t, err)
+	assert.Equal(t, velero.ResourceSelector{IncludedResources: []string{"rolebindings"}}, actual)
+}
+
+func TestRoleBindingActionExecute(t *testing.T) {
+	tests := []struct {
```
skriss: could we add a test case where namespace mappings are non-empty, but none of them apply to the role binding?

skriss:
One other thing - please add a changelog (`changelogs/unreleased/1974-alexander-demichev`)

pkg/restore/clusterrolebinding_action_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+	"github.com/vmware-tanzu/velero/pkg/test"
+)
+
+func TestClusterRoleBindingActionAppliesTo(t *testing.T) {
+	action := NewClusterRoleBindingAction(test.NewLogger())
+	actual, err := action.AppliesTo()
+	require.NoError(t, err)
+	assert.Equal(t, velero.ResourceSelector{IncludedResources: []string{"clusterrolebindings"}}, actual)
+}
```
prydonius: new line?

alexander-demichev: Thanks, fixed

skriss:
@alexander-demichev guessing you may need to rebase on top of the latest `master` code

prydonius:
@skriss I just ran this locally and it looks like it's because controller-gen is now adding an annotation:
```
+  annotations:
+    controller-gen.kubebuilder.io/version: (unknown)
```
to all CRDs. Should we fix the version we're pulling in in the builder image so we're not randomly affected by future controller-gen changes?

skriss:
Ah. yeah let's pin the version we're using

prydonius:
Ah, we needed to use master to get the single-version CRD fix, but this is now in the latest release, so we can now fix to that. I'll send a PR to fix this.

prydonius:
Restarted the test and this is now passing 🎉

skriss:
Looks like we're good to merge then

Issue #1886, opened 2019-09-18
Migrate to go modules


nrb:
As of Go v1.13, [modules](https://blog.golang.org/go1.13) are the default dependency management tool.

Upstream dependencies like [apimachinery](https://github.com/kubernetes/apimachinery/blob/master/go.mod), [client-go](https://github.com/kubernetes/client-go/blob/master/go.mod), and the [AWS SDK](https://github.com/aws/aws-sdk-go-v2/blob/master/go.mod) have moved.

Velero should adopt go modules as well in order to keep moving forward with the ecosystem.

nrb:
The migration path from `dep` is documented at https://blog.golang.org/migrating-to-go-modules.

nrb:
Vladimir Vivien has also written up a document on migrating for client-go specifically. https://medium.com/programming-kubernetes/using-go-modules-with-kubernetes-api-and-client-go-projects-2f3fdd5589a

skriss:
:+1: if someone has spare time and wants to try out the migration, that'd be great!

nrb:
I'm trying it now :)

Getting hit by https://github.com/kubernetes/kubernetes/issues/79384 and git.apache.org giving 403 at the moment.

nrb:
To fix the apache error, we'll need to add `replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999` to our eventual go.mod. Apache's moved from their git server to github mirrors.

nrb:
https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-521493597 appears to help in pulling in the various kubernetes components at the appropriate version, but it's unsuccessful on my Fedora machine with v1.15.3. I'm not sure if it's a `sed` issue or not.

PR #1984, opened 2019-10-21
fix controller-gen version in builder image

prydonius:


pkg/generated/crds/manifests/velero.io_backups.yaml
```diff
@@ -3,6 +3,8 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: (unknown)
```
prydonius: Note this will be unknown because we use `go run` and aren't using Go modules to build it. See https://github.com/kubernetes-sigs/controller-tools/pull/338/files#diff-bcee38a2cd2e5f1990f046729f3833d8L22. I'm not sure if there's a way to have `go run` set the correct version.

skriss: hmm, ok. seems harmless for now -- xref #1886

PR #1988, opened 2019-10-22
update gcr.io/heptio-images/ references to velero/

### LINKED ISSUE #1912: Migrate from GCR to DockerHub
The `velero` organization exists now, but we'll need to get the image names redefined in our Makefile.

Additionally, it _may_ make sense to push to both GCR and DockerHub for some amount of time, but preferably that would be on the order of a month or so.

prydonius:
closes #1912


prydonius:
I've only updated documentation for master, let me know if I should go back and update 1.1 and 1.0 docs since we'll push those tags to Docker Hub

skriss:
I think it's probably fine to leave v1.0/1.1 alone for now, esp. since the released binaries for those include references to the `gcr.io/heptio-images` images.

pkg/install/daemonset.go
```diff
@@ -26,7 +26,7 @@ import (
 
 func DaemonSet(namespace string, opts ...podTemplateOption) *appsv1.DaemonSet {
 	c := &podTemplateConfig{
-		image: "gcr.io/heptio-images/velero:latest",
+		image: "velero/velero:latest",
```
skriss: kind of unrelated, but we should be using `DefaultImage` here like we do for the Deployment so it's version-tagged by default. I see that in the `velero install` flow we end up passing a `podTemplateOption` to explicitly set the image to a version-tagged one, so this isn't actually a bug in the `velero install` logic, but let's update it anyway.

site/docs/master/migrating-to-velero.md
```diff
@@ -4,7 +4,7 @@ As of v0.11.0, Heptio Ark has become Velero. This means the following changes ha
 
 * The `ark` CLI client is now `velero`.
 * The default Kubernetes namespace and ServiceAccount are now named `velero` (formerly `heptio-ark`).
-* The container image name is now `gcr.io/heptio-images/velero` (formerly `gcr.io/heptio-images/ark`).
+* The container image name is now `velero/velero` (formerly `gcr.io/heptio-images/ark`).
```
skriss: I think we can just delete this page from `master`, as this isn't linked to anywhere and is a pre-1.0 change.

prydonius: I was wondering where this coming from :)

Issue #1878, opened 2019-09-16
Non-existent items should not cause backup to be PartiallyFailed


skriss:
xref #1858 

During a backup, if we have, for example, a pod that references a non-existent PVC, the backup will be `PartiallyFailed` because the PVC doesn't exist.

Arguably, the non-existence of the PVC should not cause the backup to fail. It's not Velero's job to validate the API resources, just to back them up as-is.

skriss:
@prydonius @carlisia interested in your thoughts here -- right now, if a backup item action plugin returns an "additional item" to back up, and then that additional item is found to not exist in the cluster, the backup records an error and results in a `PartiallyFailed` state.

We could change the backup code to check for not-found errors when trying to back up additional items (https://github.com/vmware-tanzu/velero/blob/master/pkg/backup/item_backupper.go#L349-L352), and just log warnings for those scenarios.

Do you think this makes sense? I'm leaning towards yes, but want to get your thoughts.

prydonius:
Do we consider `PartiallyFailed` to be a failed state or a warning state? I'm concerned that a `Completed` status could mislead a user into thinking that their backup, including any associated volumes, were successfully backed up. What if they're not aware that their PVC/PV is missing?

Would a new, less urgent-sounding, state make more sense? For example, `CompletedWithWarnings`? Could `PartiallyFailed` be renamed to something less urgent-sounding (I'm not sure what other scenarios lead to that state).

carlisia:
I agree with @prydonius. It makes me concerned if we could keep running backups of things that were supposed to be there but weren't, then the user didn't know and, oops.

I read the restore issue referenced in the description. How about we add a flag `--allow-partially-failed` that when true would allow velero to continue backing up?

skriss:
It's intended to indicate that the user needs to look at the backup and decide if it's usable or not - so more of a warning than a failure. FWIW, I originally wanted to call it `CompletedWithWarnings` but got overruled :)

In this scenario, though, I really don't see this as an error from the Velero perspective. If a pod references a non-existent PVC, and we back up the pod (including that reference), then not being able to back up a PVC that doesn't exist isn't an error in the backup - we backed up what was there, and there was no volume to back up. It *is* potentially an error in their configuration, but I don't think that's Velero's job to detect?

[...]

PR #1990, opened 2019-10-23
don't try to back up nonexistent PVCs used by pods

skriss:

xref #1878 

@prydonius @carlisia threw this together to demonstrate https://github.com/vmware-tanzu/velero/issues/1878#issuecomment-545494078 -- WDYT?

### DIFF
--- a/changelogs/unreleased/x-skriss
+++ b/changelogs/unreleased/x-skriss
@@ -1,0 +1,1 @@
+don't try to back up a PVC referenced by a pod if the PVC doesn't exist
--- a/pkg/backup/pod_action.go
+++ b/pkg/backup/pod_action.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 the Velero contributors.
+Copyright 2017, 2019 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


prydonius:
Does the same need to be applied to the PV action? If a PVC is found, but the associated PV isn't?

skriss:
we already have code that only tries to back up a PVC's claimed PV if it's "Bound" -- which isn't exactly the same thing but is probably roughly equivalent. See https://github.com/vmware-tanzu/velero/blob/master/pkg/backup/backup_pv_action.go#L56. I guess we'd have to see if there are any scenarios where the PVC could say it's bound, but the PV doesn't actually exist.

prydonius:
Ah, thanks. Just digged into this and Kubernetes will protect the PV from being deleted if it is bound to a PV (https://v1-12.docs.kubernetes.io/docs/tasks/administer-cluster/storage-object-in-use-protection/#storage-object-in-use-protection-feature-used-for-pv-protection). Forcing a deletion by removing the finalizer causes the PVC to change its status to "Lost". So that check will be sufficient.

skriss:
closing out for now, we'll revisit this topic.

PR #1999, opened 2019-10-25
Add support for private registry with custom port in restic-helper image

### LINKED ISSUE #1972: Unparsable error with custom restic helper image with extra ":"  in name (registry port)
I created configmap in velero namespace from this page https://velero.io/docs/master/restic/
as image in data.image I use 
```yaml
data:
  image: nexus-registry.domain.domain:18116/heptio-images/velero-restic-restore-helper:v1.1.0
```
(Image was pushed to this registry)

After creating restore, I see these messages in log:
> level=warning msg="Plugin config contains unparseable image name" cmd=/velero image="nexus-registry.domain.domain:18116/heptio-images/velero-restic-restore-helper:v1.1.0" logSource="pkg/restore/restic_restore_action.go:171" pluginName=velero pod=test-velero/coffee-679f7c7698-r7rvr restore=velero/velero-restore-6
time="2019-10-17T07:50:24Z" level=info msg="Using image \"gcr.io/heptio-images/velero-restic-restore-helper:v1.1.0\"" cmd=/velero logSource="pkg/restore/restic_restore_action.go:105" pluginName=velero pod=test-velero/coffee-679f7c7698-r7rvr restore=velero/velero-restore-6

So, my custom image from private registry is not used during restore (I believe it has to do with extra ':' for port definition in the image name)

Image name was parsed correctly

**Environment:**
v1.1.0 
v1.11.0+d4cacc0

cognoz:
Closes #1972 

example of image definition with custom registry port in CM: 
```yaml
data:
  image: nexus-registry.domain.domain:18116/heptio-images/velero-restic-restore-helper:v1.1.0
```

### DIFF
--- a/pkg/restore/restic_restore_action.go
+++ b/pkg/restore/restic_restore_action.go
@@ -166,6 +166,10 @@ func getImage(log logrus.FieldLogger, config *corev1.ConfigMap) string {
 		// tagged image name
 		log.Debugf("Plugin config contains image name with tag")
 		return image
+	case len(parts) == 3:
+		// tagged image name with registry port definition
+		log.Debugf("Plugin config contains image name with tag and custom registry port")
+		return image
 	default:
 		// unrecognized
 		log.Warnf("Plugin config contains unparseable image name")
--- a/pkg/restore/restic_restore_action.go
+++ b/pkg/restore/restic_restore_action.go
@@ -158,21 +158,21 @@ func getImage(log logrus.FieldLogger, config *corev1.ConfigMap) string {
 
 	parts := strings.Split(image, "/")
 
-        if len(parts) == 1 {
-                // Image supplied without registry part
-                log.Debugf("Plugin config contains image name without registry name. Return defaultImageBase")
-                return initContainerImage(defaultImageBase)
-        }
+	if len(parts) == 1 {
+		// Image supplied without registry part
+		log.Debugf("Plugin config contains image name without registry name. Return defaultImageBase")
+		return initContainerImage(defaultImageBase)
+	}
 
-        if !(strings.Contains(parts[len(parts)-1], ":")) {
-                // tag-less image name: add tag
+	if !(strings.Contains(parts[len(parts)-1], ":")) {
+		// tag-less image name: add tag
 		log.Debugf("Plugin config contains image name without tag. Adding tag.")
 		return initContainerImage(image)
-        } else {
-                // tagged image name
-                log.Debugf("Plugin config contains image name with tag")
-                return image
-        }
+	} else {
+		// tagged image name
+		log.Debugf("Plugin config contains image name with tag")
+		return image
+	}
 }
 
 // getResourceRequests extracts the CPU and memory requests from a ConfigMap.
--- a/pkg/restore/restic_restore_action.go
+++ b/pkg/restore/restic_restore_action.go
@@ -156,25 +156,23 @@ func getImage(log logrus.FieldLogger, config *corev1.ConfigMap) string {
 
 	log = log.WithField("image", image)
 
-	parts := strings.Split(image, ":")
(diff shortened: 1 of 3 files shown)


skriss:
Thanks for the PR @cognoz! We do need you to add DCO signoff to your commit -- `git commit --amend --signoff` followed by `git push -f` should do the trick.

pkg/restore/restic_restore_action.go
```diff
@@ -166,6 +166,10 @@ func getImage(log logrus.FieldLogger, config *corev1.ConfigMap) string {
 		// tagged image name
 		log.Debugf("Plugin config contains image name with tag")
 		return image
+	case len(parts) == 3:
```
skriss: This is making me think that we may need to switch to using 1+ regexes here for proper matching. A couple of examples that may not properly be matched here:

- `my-custom-registry.io:an-invalid-port-because-its-not-numeric/velero:latest` (shouldn't be valid because the "port" is not numeric, but will be matched by the third case)
- `my-custom-registry.io:8888/velero` (should be a valid tagless image name, i.e. matched by the first case, but will be matched by the second case, not the first)

cognoz: Agree with you. Not really sure how to use regexp in GO properly, though )

skriss: OK, here's what I'm thinking we should do:

What we're really trying to do with this piece of code is detect if the user has specified an image without a tag (e.g. `steveheptio/velero`), and if so, add a version-specific tag to it, so it becomes e.g. `steveheptio/velero:v1.1.0`.

Instead of splitting `image` on colons (`:`), we could split on front slashes (`/`). If the *last* part (i.e. `parts[len(parts)-1]`) does not contain a colon, it's a tag-less image name, so we want to add the tag by calling/returning `initContainerImage(image)`. Otherwise, we can just use `image` as-is.

If we approach it this way, it should avoid the issue you're running into entirely.

cognoz: i worked it out with '/', but now I get this errror after starting container - 
`An error occurred: some backup storage locations are invalid: error getting backup store for location "default": unable to locate ObjectStore plugin named velero.io/aws`
Where should I look to fix it?

skriss: It sounds like you've merged/rebased with our latest code, where the cloud-provider plugins are no longer in-tree. The quick solution is to run `velero plugin add velero/velero-plugin-for-aws:v1.0.0` to add the AWS plugin to your deployment.

skriss:
@cognoz I think the code looks reasonable. I would like to add a couple new test cases though:

- one for an image name that doesn't have a `/` (i.e. `len(parts) == 1`) - validate that the default image is returned
- one for an image that uses a custom port for the registry, and *doesn't* have a tag - validate that the tag gets added
- one for an image that uses a custom port for the registry, and *does* have a tag - validate that it's returned as-is

You should be able to just copy/paste and modify from the existing test cases. Once that's done, I think we can merge this.

skriss:
@cognoz thanks for making the updates, this LGTM! One last request - could you add a changelog with a brief description of the bug fix? See https://velero.io/docs/master/code-standards/ for details.

skriss:
cc @carlisia @prydonius @nrb - PTAL

cognoz:
Changelog added, thx for guidelines!

skriss:
@carlisia @nrb @prydonius ping - this one's waiting for review/merge.

skriss:
@carlisia if you look at the code you'll notice that we're no longer trying to split image names based on `:`, so the test case you're referring to isn't really the best test of the logic anymore. The replacement case that was added (an image without a `/`) is effectively the "invalid data" case now, and validates that the default image is returned in such a case.

PR #2014, opened 2019-10-29
Update AWS SDK dep for IRSA awareness

redradrat:
In order to continue using a stable version but being able to leverage AWS IRSA support as described in [aws docs](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html), I would propose changing this on a v1.1.0 base, since there doesn't seem to be an impact on dependency side.

This MR just contains the update to the aws sdk needed to leverage said functionality. I added the change directly on the v1.1.0, in line for a possible patch release. (since the next minor already looks rather major)


skriss:
@redradrat with the impending release of `v1.2` within the next 1-2 weeks, I don't think we're going to be releasing a subsequent patch version of `v1.1`. Although the internal changes for v1.2 are material, I don't think you'll find that there's much of a change from a user perspective, so upgrading should be easy enough!

As such, I'm going to close this out - appreciate the input nonetheless!

Issue #1998, opened 2019-10-25
Error restoring restic volumes - invalid id \"\": multiple IDs with prefix found


robertgates55:
**What steps did you take and what happened:**
On attempting to restore a successful velero backup of a namespace with restic enabled, I'm getting the following error in the restic logs:
```
time="2019-10-25T10:05:57Z" level=error msg="Error restoring volume" controller=pod-volume-restore error="error running restic restore, cmd=restic restore --repo=s3:s3-eu-west-1.amazonaws.com/duco-qa1-backups-dea6831f/restic/knowledgehub --password-file=/tmp/velero-restic-credentials-knowledgehub194842862 --cache-dir=/scratch/.cache/restic  --target=., stdout=, stderr=invalid id \"\": multiple IDs with prefix found\n: exit status 1" error.file="/go/src/github.com/heptio/velero/pkg/controller/pod_volume_restore_controller.go:343" error.function="github.com/heptio/velero/pkg/controller.(*podVolumeRestoreController).restorePodVolume" logSource="pkg/controller/pod_volume_restore_controller.go:298" name=knowledgehub-1571946878-20191025110546-wmsh7 namespace=velero restore=velero/knowledgehub-1571946878-20191025110546
```

**What did you expect to happen:**
A successful, clean restore of all my volumes.

**The output of the following commands will help us better understand what's going on**:
(Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)

`velero backup describe <backupname>`:
```
Name:         knowledgehub-1571946878
Namespace:    velero
Labels:       velero.io/backup=knowledgehub-1571946878
              velero.io/pv=pvc-c615f4f3-f67d-11e9-8225-068a97741dda
              velero.io/storage-location=default
Annotations:  <none>

Phase:  Completed

Namespaces:
  Included:  knowledgehub
  Excluded:  <none>

Resources:
  Included:        *
  Excluded:        <none>
  Cluster-scoped:  auto

Label selector:  <none>

Storage Location:  default

Snapshot PVs:  auto

TTL:  720h0m0s

Hooks:  <none>

Backup Format Version:  1

Started:    2019-10-24 20:54:41 +0100 BST
Completed:  2019-10-24 20:54:47 +0100 BST

Expiration:  2019-11-23 19:54:41 +0000 GMT

Persistent Volumes:  5 of 5 snapshots completed successfully (specify --details for more information)

Restic Backups (specify --details for more information):
  Completed:  2
```

[...]

PR #2031, opened 2019-11-01
restic: don't try to restore PVBs with no snapshotID

skriss:

ref https://github.com/vmware-tanzu/velero/issues/1998#issuecomment-548831698

If a `PodVolumeBackup` doesn't have a `snapshotID`, we shouldn't try to restore it because there's nothing to restore. This could be the case if the PVB failed, or if it completed but the volume being backed up was empty, so no snapshot was created.

### DIFF
--- a/changelogs/unreleased/2031-skriss
+++ b/changelogs/unreleased/2031-skriss
@@ -1,0 +1,1 @@
+bug fix: don't try to restore pod volume backups that don't have a snapshot ID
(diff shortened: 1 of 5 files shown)


skriss:
@carlisia @prydonius @nrb I would like to get this bug fix into v1.2 - PTAL when you can!

skriss:
added changelog!

skriss:
merging since it's got 2 approvals.

PR #2049, opened 2019-11-06
run goreleaser.sh in container

### LINKED ISSUE #1383: Containerize goreleaser process
`make release` runs `goreleaser` to generate a release, including building the `velero` binary for the target arch/os combos. Currently, this runs on the desktop. We should containerize it so we can pin the `goreleaser` and `go` dependencies.

skriss:

closes #1383 

This PR moves running `goreleaser` into the existing build container image.  `goreleaser` generates GitHub releases, including building all of the `velero` binaries that get published as part of the GitHub release.

Previously, `goreleaser` would run on your (my) laptop, which was not ideal. Now, it runs in the same build container image that we use for building the binaries that go into the server container images.

I tested this out all the way through publishing a draft GH release (I've since deleted it).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,8 +215,27 @@ ci: all verify test
 changelog:
 	hack/changelog.sh
 
+# release builds a GitHub release using goreleaser within the build container.
+#
+# To dry-run the release, which will build the binaries/artifacts locally but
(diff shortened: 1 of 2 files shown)


skriss:
merging, has 2 approvals

PR #2052, opened 2019-11-07
release instructions

skriss:

FYI @nrb @prydonius @carlisia 

I wrote these based on notes/memory/today's release, so hopefully they're pretty complete & accurate, but I'll do some review and testing of parts to ensure they're valid before merge.

There's also definitely room for improvement here, I'd like to automate away a lot of this now that we're in a fairly stable place, so definitely review this with an eye towards how we can do that.

### DIFF
--- a/site/docs/master/release-instructions.md
+++ b/site/docs/master/release-instructions.md
@@ -74,3 +74,7 @@ This process is the same for both pre-release and GA, except for the fact that t
     - Verify that `velero version` shows the expected output
     - Run a backup/restore and ensure it works
 1.  (GA Only) Merge the blog post PR.
+1.  Announce the release:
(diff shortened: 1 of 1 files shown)


prydonius:
This is really great, thanks for writing this up @skriss!

skriss:
Made a couple edits.

Ready for review/merge.  We'll follow these instructions for the next release to validate them and make any edits necessary.

carlisia:
Yes, this looks great. I think we'd have to use it to find opportunities for improvements, just reading it looks complete. Except, is it worthwhile adding a list of the places where we need to announce releases (twitter, google group, slack)? Does this belong here?

skriss:
Yep, sure, will add.

PR #2068, opened 2019-11-15
remove fsfreeze-pause image, replace with ubuntu in nginx example

### LINKED ISSUE #2056: Consider removing the `fsfreeze-pause` container image
We should consider removing the `fsfreeze-pause` container image that we're currently building/publishing out of our repo.

We could easily replace it in our nginx examples with a vanilla `ubuntu:bionic` image, whose entrypoint is the sleep loop that we have here: https://github.com/vmware-tanzu/velero/blob/master/Dockerfile-fsfreeze-pause#L19.

It's also not even necessary to have it in its own container for example purposes -- the `fsfreeze` command is available in the `nginx` image. But, it is useful to show how to use a sidecar for hook purposes, so I'd rather leave it.

If we drop our `fsfreeze-pause` image, we'll get some simplification and speedup in our build/deploy process.

@carlisia @prydonius @nrb WDYT? can you think of any reason why we'd want/need to keep this around?

skriss:

Closes #2056 

I've tested this out on GKE and it works fine.


examples/nginx-app/with-pv.yaml
```diff
@@ -70,15 +70,20 @@ spec:
...
         securityContext:
           privileged: true
         volumeMounts:
           - mountPath: "/var/log/nginx"
             name: nginx-logs
             readOnly: false
+        command:
+          - "/bin/bash"
+          - "-c"
+          - "while true; do sleep 10000; done"
```
prydonius: would it be better to `tail -f /dev/null`?

prydonius: Looks like `sleep infinity` is another option too

skriss: Yeah we could switch to one of those - we had the loop because we used to use alpine/busybox which didn't support `sleep infinity`. The internet has varying opinions on which is the most efficient - I'll take a look and decide. TBH I think any are probably fine in this context :)

skriss: switched to `sleep infinity`.

PR #2099, opened 2019-12-03
Add time conversion link for community meeting

nrb:

### DIFF
--- a/site/community.md
+++ b/site/community.md
@@ -19,6 +19,6 @@ You can follow the work we do, see our milestones, and our backlog on our [ZenHu
 * Join our Kubernetes Slack channel and talk to over 800 other community members: [#velero](https://kubernetes.slack.com/messages/velero)
 * Join our [Google Group](https://groups.google.com/forum/#!forum/projectvelero) to get updates on the project and invites to community meetings.
 * Join the Velero community meetings:
-  * Every Tuesday at 12PM ET / 9AM PT - [Zoom link](https://vmware.zoom.us/j/551441444)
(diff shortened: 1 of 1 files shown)


jonasrosland:
I'm happy to move to UTC, but since UTC doesn't observe DST, it would've been 4PM UTC in October and 5PM UTC in November. Do you want to change that during our DST, or are you suggesting we move the meeting to always coincide with 5PM UTC?

nrb:
Ah, good point! I was actually trying to provide a point of reference for folks not in the US, but I didn't think about the DST issue, so I wasn't proposing making changes to ping it to 5pm UTC.

I can do a bit more research here and maybe use GMT, which I think does observe DST.

jonasrosland:
Yes, GMT observes DST, but not at the same dates as the US :)
We could add a link to "Convert to your timezone here" instead, like this?

9AM PT ([click to convert to your time zone](http://www.thetimezoneconverter.com/?t=09:00&tz=PT%20%28Pacific%20Time%29))

carlisia:
OMG TZ!

I love the idea of the TZ converter link, especially since it blocks the 9AM PT box. Let's do that?

carlisia:
I think that's even way better then adding one additional (but specific) reference point.

nrb:
Yeah, I think the converter link is a far better solution too! I'll get that changed.

Thanks Jonas!

carlisia:
image

Source: https://twitter.com/PR0GRAMMERHUM0R/status/1200172543610458112

Thanks for the laugh @nrb!

Issue #2085, opened 2019-11-29
OpenShift/Velero/Restic - open /host_pods/: permission denied


mircster:
Hello,

i have installed and tested Velero v1.2.0 on OpenShift on Azure with Azure Provider successful.
Now I uninstalled Velero and want to install Velero v1.2.0 with Restic.

**What steps did you take and what happened:**
[A clear and concise description of what the bug is, and what commands you ran.)

**OpenShift Version:**
oc v3.11.153
kubernetes v1.11.0+d4cacc0

```
./velero install \
    --provider azure \
    --plugins velero/velero-plugin-for-microsoft-azure:v1.0.0 \
    --bucket $BLOB_CONTAINER \
    --secret-file ./credentials-velero \
    --backup-location-config resourceGroup=$AZURE_BACKUP_RESOURCE_GROUP,storageAccount=$AZURE_STORAGE_ACCOUNT_ID \
    --snapshot-location-config apiTimeout=5m \
	--use-restic 
```
I  edited the image in the deployment and deamon set, as we are using a private repository.
Also i had to add proxy configuration to the environment variables of the deployment.

The installation works fine and the velero pod runs fine, but the restic pods run into **CrashLoopBackOff** with following error:

> An error occurred: could not read pod volumes host path: open /host_pods/: permission denied

**I followed the instructions on  https://velero.io/docs/v1.2.0/restic/**

```
oc adm policy add-scc-to-user privileged -z velero -n velero

Modify the DaemonSet yaml to request a privileged mode and mount the correct hostpath to pods volumes.

@@ -35,7 +35,7 @@ spec:
             secretName: cloud-credentials
         - name: host-pods
           hostPath:
-            path: /var/lib/kubelet/pods
+            path: /var/lib/origin/openshift.local.volumes/pods
         - name: scratch
           emptyDir: {}
       containers:
@@ -67,3 +67,5 @@ spec:
               value: /credentials/cloud
             - name: VELERO_SCRATCH_DIR
               value: /scratch
+          securityContext:
+            privileged: true

```
The first edit (**path: /var/lib/origin/openshift.local.volumes/pods**) in the deamon set yaml is no problem, but the second one (**privileged: true**) doesen´t work!
The **+ securityContext:** is not correct for us as securityContext is already there plus **runAsUser: 0**

```
        securityContext:
          runAsUser: 0
```

**But now the Problem starts!**
If i no want to add **privileged: true** and reload the deamon set it is gone.
[...]

PR #2101, opened 2019-12-03
Add oc patch commands for DS

git001:
From my point of view is it much easier to run **oc** commands instead of editing a yaml file.  
My proposel is to offer the user both methodes, manuall edit and automatic edit.  
This patch is related to #2085


git001:
Will colse it because of missing Signed-off => DCO

PR #2118, opened 2019-12-11
bug fix: don't restore cluster-scoped resources by default when restoring specific namespaces

### LINKED ISSUE #1914: All cluster-scoped resources are restored when selecting a specific namespace during restore

I am taking nightly full cluster backup(that includes all Namespaces in my k8s cluster) and restoring one namespace. Restore looks good and i am bale to access my application but I see one warning in my restore which contains a resource which is not part of my restore namespace.
Could you please suggest why we are getting this warning and why it is trying to restore this resource when i am not even including that resource related namespace

> velero restore describe restore-one-namespace-lrc-mnrf-fsims-dev --details
> Name:         restore-one-namespace-lrc-mnrf-fsims-dev
> Namespace:    velero
> Labels:       <none>
> Annotations:  <none>
> 
> Phase:  Completed
> 
> Warnings:
>   Velero:     <none>
>   **_Cluster:  not restored: persistentvolumes "logging-datanode-*" already exists and is different from backed up version._**
>   Namespaces: <none>
> 
> Backup:  whole-cluster-backup-26092019112127
> 
> Namespaces:
>   Included:  lrc-mnrf-fsims-dev
>   Excluded:  <none>
> 
> Resources:
>   Included:        *
>   Excluded:        nodes, events, events.events.k8s.io, backups.velero.io, restores.velero.io, resticrepositories.velero.io
>   Cluster-scoped:  auto
> 
> Namespace mappings:  <none>
> 
> Label selector:  <none>
> 
> Restore PVs:  auto
> 
> Restic Restores:
>   Completed:
>     lrc-mnrf-fsims-dev/lrc-mnrf-fsims-dev-relea-0: liberty-pvc
> 

* `kubectl logs deployment/velero -n velero`
* `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
* `velero backup logs <backupname>`
* `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
* `velero restore logs <restorename>`

**Environment:**

skriss:
Closes #1914 

Units are passing but I want to make sure we have adequate coverage around this combo of params before merging.

### DIFF
--- a/pkg/restore/restore_test.go
+++ b/pkg/restore/restore_test.go
@@ -1408,7 +1408,7 @@ func TestRestoreActionAdditionalItems(t *testing.T) {
 			},
 		},
 		{
-			name:    "when using a restore namespace filter, additional items that are cluster-scoped are restored",
+			name:    "when using a restore namespace filter, additional items that are cluster-scoped are restored when IncludeClusterResources=nil",
 			restore: defaultRestore().IncludedNamespaces("ns-1").Result(),
 			backup:  defaultBackup().Result(),
 			tarball: newTarWriter(t).
@@ -1434,8 +1434,8 @@ func TestRestoreActionAdditionalItems(t *testing.T) {
 			},
 		},
 		{
-			name:    "when using a restore resource filter, additional items that are non-included resources are not restored",
-			restore: defaultRestore().IncludedResources("pods").Result(),
+			name:    "additional items that are cluster-scoped are not restored when IncludeClusterResources=false",
+			restore: defaultRestore().IncludeClusterResources(false).Result(),
 			backup:  defaultBackup().Result(),
 			tarball: newTarWriter(t).
 				addItems("pods", builder.ForPod("ns-1", "pod-1").Result()).
@@ -1460,8 +1460,8 @@ func TestRestoreActionAdditionalItems(t *testing.T) {
 			},
 		},
 		{
-			name:    "when IncludeClusterResources=false, additional items that are cluster-scoped are not restored",
-			restore: defaultRestore().IncludeClusterResources(false).Result(),
+			name:    "when using a restore resource filter, additional items that are non-included resources are not restored",
+			restore: defaultRestore().IncludedResources("pods").Result(),
 			backup:  defaultBackup().Result(),
 			tarball: newTarWriter(t).
 				addItems("pods", builder.ForPod("ns-1", "pod-1").Result()).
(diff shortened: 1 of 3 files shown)


skriss:
I reviewed the unit test cases and I think we have good coverage around this.

Manual testing looked good here as well, though I found a separate bug that I'll file a PR for shortly.

skriss:
@carlisia @ashish-amarnath please take a look when you get a chance, thanks :)

pkg/restore/restore_test.go
```diff
@@ -1378,7 +1408,7 @@ func TestRestoreActionAdditionalItems(t *testing.T) {
 			},
 		},
 		{
-			name:    "when using a restore namespace filter, additional items that are cluster-scoped are restored",
+			name:    "when using a restore namespace filter, additional items that are cluster-scoped are restored when IncludeClusterResources=nil",
```
carlisia: I know I'm probably missing something. In this use-case, the additional items are being restored (by default, since `IncludeClusterResources=nil`). Why am I thinking that they should not be restored in this case?

ashish-amarnath: Can you please re-phrase this as a `should ...` or `should not...`?

carlisia: This is an interesting issue, how to word the tests. We don't use `should` much in the velero tests, but we never talked about why. I remember reading before about whether or not to use it. I did a quick Google search and this sums up discussions I've seen before: https://www.quora.com/Can-we-use-the-word-should-when-writing-a-test-cases-summary-in-manual-testing

skriss: `IncludeClusterResources` is a nullable bool, so has 3 possible modes:

- `true`: all cluster-scoped resources that are selected by the resource/label selectors are restored
- `false`: no cluster-scoped resources are restored
- `nil` ("auto"): only cluster-scoped resources that are dependencies of other namespaced resources being restored, are restored (e.g. we're restoring a pod that uses a PVC/PV, so we'd restore the PV as well). the way this is actually implemented in the code is by allowing "additional items" to be restored, even if they're cluster-scoped, when `IncludeClusterResources=nil`, since "additional items" are how we capture relationships between resources.

This logic already holds for backups, but we had a bug in the restore code that didn't implement the `nil` logic correctly.

skriss: yeah, so we're not using "should" much in this test file at least, so changing this one case's wording would make it inconsistent with the rest of the cases. My approach in this kind of scenario is usually to keep consistent with the existing pattern, and revisit the broader issue separately - @ashish-amarnath does that work for you?

carlisia: I appreciate the explanation. This also should be in the docs, tho I'm not sure where. I'm opening a tix.

ashish-amarnath: I didn't see the whole file.. but then this one sneaked in I guess! https://github.com/vmware-tanzu/velero/pull/2118/files#diff-3a5824b5e6665e0247bae6406e0ad1fbR318

PR #2119, opened 2019-12-11
[design] Proposal of triggering backups based on Kubernetes events.

### LINKED ISSUE #2111: Trigger backups based on k8s events
I would like to trigger backup(s) based on a delete/remove (CrashLoopBackOff, Terminating, Err, OOMKilled, etc) event(s).

A custom controller where it use the k8s (watch) api to capture the previous events of the k8s objects, and trigger a velero backup for these component(s) that are being deleted/terminated.

nzoueidi:
Closes #2111 

Any new ideas/thoughts would be much appreciated.

Signed-off-by: Naeil Ezzoueidi naeilzoueidi@ubuntu.com

cc @nrb

### DIFF
--- a/design/trigger-backups-api-events.md
+++ b/design/trigger-backups-api-events.md
@@ -1,0 +1,35 @@
+# Trigger backups based on Kubernetes events 
+
+Triggering backups with Velero based on Kubernetes api events (Terminating, deleting, CrashLoopBackOff, etc..).
+
+## Goals
+
+- Rather than performing backups manually or by scheduling them, backups are created based on events.
+- Recovering the components of the cluster when they were deleted accidently. 
+- Blocking the delete process of a component in the cluster if it was a human intervention and trigger a backup.
+
+## Non Goals
+
+- N/A
+
+## Background
+
+## High-Level Design
+
+A custom controller that list and watch all the events of Kubenretes and based on that it trigger backups.
+
+## Detailed Design
+
+A custom controller where it have its own refelctor to list and watch a specific components using the kubernetes watch api. 
+It will add the object mentored with its status/current event to a queue.
+Then pop up the object based on its status (criticity and priority).
+Block the delete process if it was performed from a human being and trigger a backup for that component.
+Trying to catch the errors happening on the components so it predicts when they will be down and trigger backups.
+
+## Alternatives Considered
+
+Apply the same design in a separate plugin as a new kind.
+
+## Security Considerations
+
+N/A


carlisia:
👀

carlisia:
I'm still not quite sure that this belongs in Velero core. At the very least, I'd like to see more users requesting this feature. 

My initial comment: This proposal has as a goal to trigger a backup in case of an accidental deletion, and to block a deletion of a component if it was performed by a human. I'd like to see it  address how the system would distinguish an accidental from an intentional deletion. 

The part that proposes blocking deletions could also be addressed with more details. It states it would happen in the case of human intervention (and trigger a backup), but how would the system distinguish this?

design/trigger-backups-api-events.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Goals
+
+- Rather than performing backups manually or by scheduling them, backups are created based on events.
+- Recovering the components of the cluster when they were deleted accidently. 
+- Blocking the delete process of a component in the cluster if it was a human intervention and trigger a backup.
+
+## Non Goals
+
+- N/A
```
nrb: I think some non-goals here might be useful in providing constraints around the design.

design/trigger-backups-api-events.md
```diff
@@ -0,0 +1,35 @@
+# Trigger backups based on Kubernetes events 
+
+Triggering backups with Velero based on Kubernetes api events (Terminating, deleting, CrashLoopBackOff, etc..).
+
+## Goals
+
+- Rather than performing backups manually or by scheduling them, backups are created based on events.
+- Recovering the components of the cluster when they were deleted accidently.
```
nrb: Can you clarify this some? How is this different from what Velero already does?

design/trigger-backups-api-events.md
```diff
@@ -0,0 +1,35 @@
+# Trigger backups based on Kubernetes events 
+
+Triggering backups with Velero based on Kubernetes api events (Terminating, deleting, CrashLoopBackOff, etc..).
+
+## Goals
+
+- Rather than performing backups manually or by scheduling them, backups are created based on events.
+- Recovering the components of the cluster when they were deleted accidently. 
+- Blocking the delete process of a component in the cluster if it was a human intervention and trigger a backup.
```
nrb: This is an interesting idea - can you go into detail on how you'd determined whether the deletion was triggered by a human or not?

design/trigger-backups-api-events.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Background
+
+## High-Level Design
+
+A custom controller that list and watch all the events of Kubenretes and based on that it trigger backups.
+
+## Detailed Design
+
+A custom controller where it have its own refelctor to list and watch a specific components using the kubernetes watch api.
```
nrb: How would users specify what components to watch? How would users specify what events to watch for?

design/trigger-backups-api-events.md
```diff
@@ -0,0 +1,35 @@
...
+
+## High-Level Design
+
+A custom controller that list and watch all the events of Kubenretes and based on that it trigger backups.
+
+## Detailed Design
+
+A custom controller where it have its own refelctor to list and watch a specific components using the kubernetes watch api. 
+It will add the object mentored with its status/current event to a queue.
+Then pop up the object based on its status (criticity and priority).
```
nrb: How is the priority determined?

design/trigger-backups-api-events.md
```diff
@@ -0,0 +1,35 @@
...
+Block the delete process if it was performed from a human being and trigger a backup for that component.
+Trying to catch the errors happening on the components so it predicts when they will be down and trigger backups.
```
nrb: I'm not sure how this part connects to backups - there's something I'm missing.

nzoueidi:
Thank you @carlisia  and @nrb  for the review and for the comments, sorry for the delay though, I will reply and push some modifications this week.

nzoueidi:
> I'm still not quite sure that this belongs in Velero core. At the very least, I'd like to see more users requesting this feature.
>
This also could be implemented as a plugin with new kind. IMHO, this is a good feature as it really covers a real scenario that can happen.

> My initial comment: This proposal has as a goal to trigger a backup in case of an accidental deletion, and to block a deletion of a component if it was performed by a human. I'd like to see it address how the system would distinguish an accidental from an intentional deletion.
> 
I would say that this is the role of a custom controller that uses the watch kubernetes api and listens on any kubectl delete commands to distinguish if the actions has been performed by a human or not. 
Either ways, I see that backing up the components in both situations is good, what do you think?

design/trigger-backups-api-events.md
nzoueidi: For example, we could use the help of other monitoring tools (promtail and loki), based on metrics we can for example predict if the application is going to be down (consuming many resources e.g memory, disk, etc..).

design/trigger-backups-api-events.md
nzoueidi: I would say based on labels. E.g status: critical

design/trigger-backups-api-events.md
nzoueidi: this could be achieved through namespace selectors. And for specifying what events to watch; this should be by default watching any delete events..

design/trigger-backups-api-events.md
nzoueidi: I am thinking about watching all the delete events and listening to the kubectl delete commands, this way we could distinguish if the action has been performed by a human or not.

design/trigger-backups-api-events.md
nzoueidi: With Velero, we create backups manually or by scheduling them, what is different for waht Velero already does is that we trigger backups for the components that are being called to be deleted

nrb: So you're envisioning that this would intercept a delete request and back up the components _before_ allowing them to delete?

design/trigger-backups-api-events.md
nrb: Ok. Do you know how to distinguish this via the events emitted on the Kubernetes API server? They may labelled or annotated differently, but I'm not sure.

If they are, I'd like to see that called out in the document.

design/trigger-backups-api-events.md
nrb: Namespaces are one way, sure. I was also thinking about specifying per resource type.

I'd recommend providing explicit examples of what the user would type to list out events and namespaces. What would the configuration file or CRD look like? Would it be an addition to an existing CRD? These kinds of questions would be good to be answered in the proposal itself.

design/trigger-backups-api-events.md
nrb: So each object would have to be labelled? Would this be a standard label, or are we introducing a new label just for this?

You can incorporate your answers into the proposal doc.

design/trigger-backups-api-events.md
nrb: I think that's getting outside of Velero's scope and into machine learning. While I think it's useful functionality for sure, I don't think I would accept this portion into Velero core.

carlisia:
I'm putting this on the agenda for tomorrow's meeting. @nzoueidi if you could join it would be great: https://velero.io/community/.

carlisia:
We discussed this in our community meeting today. The general consensus is that this request overall is useful but it is a great use-case for an operator, and not so much for inclusion into the Velero code base.

We will leave this PR open for a couple weeks with the intent of welcoming additional opinions in favor of this request, in case we are missing the needs of more users and how they would be using this. Else we'll close it but this conversation can always be picked up later.

Will post a link to the meeting recording once it's up on YT.

carlisia:
Recording of our meeting: https://www.youtube.com/watch?v=SrGH2_ufWJ8&list=PL7bmigfV0EqQRysvqvqOtRNk4L5S7uqwM&index=2&t=0s

skriss:
we haven't gotten any further input here; I'm going close this out.

If someone works on this outside of core Velero, we'd love to see it!

Issue #2132, opened 2019-12-14
Security Concern with Deprecated uuid package: satori/go.uuid


jnaulty:
**Describe the problem/challenge you have**
I would like to see no known security vulnerabilities reported when running `vmware-tanzu/velero` in the snyk.io security scan.

**Describe the solution you'd like**
Migrate the `satori/go.uuid` package which is no longer being maintained to the `gofrs/uuid` package.
`gofrs/uuid` reason for existence:
> This project was originally forked from the github.com/satori/go.uuid repository after it appeared to be no longer maintained, while exhibiting critical flaws. We have decided to take over this project to ensure it receives regular maintenance for the benefit of the larger Go community.

**Anything else you would like to add:**
Non-Random UUID Issue in `satori/go.uuid`: https://github.com/satori/go.uuid/issues/73
Deprecation of `satori/go.uuid`: https://github.com/satori/go.uuid/pull/85

Snyk Report:
![image](https://user-images.githubusercontent.com/6826729/70844954-b9d69280-1dfd-11ea-8943-738d277caa88.png)

**Environment:**

- Velero version (use `velero version`): `1.2.0`

PR #2133, opened 2019-12-14
Switch from satori/go.uuid to gofrs/uuid

jnaulty:
`satori/go.uuid` has a know issue with random uuid generation.
`gofrs/uuid` is still maintained and has fixed the random uuid generation
issue present in satori/go.uuid

Closes: #2132


pkg/restore/restore.go
```diff
@@ -176,7 +180,7 @@ func NewKubernetesRestorer(
 		resourceTerminatingTimeout: resourceTerminatingTimeout,
 		resourcePriorities:         resourcePriorities,
 		logger:                     logger,
-		pvRenamer:                  func(string) string { return "velero-clone-" + uuid.NewV4().String() },
+		pvRenamer:                  func(string) string { return "velero-clone-" + veleroCloneUuid.String() },
```
skriss: I'm pretty sure this change introduces a bug where if multiple PVs are being cloned, either within a single restore or across multiple restores, we'll get a name conflict since we're not generating a new UUID within this function anymore.

nrb: Ah shoot, good call. So L170 should move into this renamer func?

skriss: Yeah; it looks like we'll need to add an `error` return parameter to the `pvRenamer` func and handle it appropriately at the call site since the new library may return an error.

nrb: I'll create an issue for addressing this.

nrb: #2137

skriss: :+1: thanks!

PR #2148, opened 2019-12-17
Use more recent nginx in example

dns2utf8:
I am assuming some people will leave the example running on their cluster for quite some time so patching the example every now and then seams reasonable.


skriss:
Thanks for this @dns2utf8, agree this is a good idea.

Can you add DCO signoff to your latest commit with `git commit --amend --signoff && git push -f`? Thanks!

dns2utf8:
Sure. I also squashed the commits.

Issue #2104, opened 2019-12-04
Move to GitHub Actions


nrb:
GitHub Actions are now GA, so we should at least investigate the level of effort required to move our CI/CD pipeline from Travis CI to Actions.

As far as I'm aware, Actions are based solely on container images, and Velero's work is based on running build container from the host. We may be able to break this down into having the build image uploaded to a registry and fetching it.

Rough plan of action:

- [ ] Add make target solely for building and pushing the build container image to Docker Hub
- [ ] Rework the build, test, and CI jobs to run inside the build image, pulled from Docker Hub
- [ ] Add the GH Action files, using the build container image as the base
- [ ] Remove Travis CI files

Current architecture builds should _not_ change.

Investigate Contour's usage at https://github.com/projectcontour/contour

nrb:
Reviewing [the GitHub runner docs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners), it looks like for now the simpler approach may be to use an `ubuntu-latest` VM that simply runs `make ci`. This way, we've moved the platform, and the work of moving the builder container around can be deferred. I would still like to do it, but since it's not an absolute requirement, I don't see a solid reason for changing it at the moment.

Some of the other logic, such as tagging the `latest` Docker image will still have to be adjusted, though, as this is tied into Travis's environment variables. Instead, I think this should rely on the GitHub API. The [`create` event](https://developer.github.com/v3/activity/events/types/#createevent) will happen when a tag is cut and we can trigger a container build on that, or we could use a [`release` event](https://developer.github.com/v3/activity/events/types/#releaseevent) when the release is actually pushed up, checking to see if it's a prerelease or not.

nrb:
PR #2150 moves us to running `make ci` on branch pushes an on PRs. It doesn't touch .travis.yml as that file does not control Travis's push/PR triggers, only the Travis console does.

Once #2150 is merged, I'll disable Travis builds for PRs. I believe that Travis push builds must be kept until we move container building/publishing to GitHub Actions, but I need to confirm this with their docs.

nrb:
#2150 has merged, and I have disabled Travis builds for PRs.

skriss:
@ashish-amarnath figured out a lot of this today for the CSI plugin repo, so I'd really like to finish the swing here in 1.5 and get our remaining repos fully migrated off Travis. Adding it to the 1.5 milestone.

nrb:
[...]

PR #2150, opened 2019-12-18
Use GitHub Actions to run checks for pushes and pull requests

nrb:
Part of #2104, which will be addressed in parts.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,14 @@
+name: Pull Request CI Check
+on: [push, pull_request]
+jobs:
+
+  build:
+    name: Run CI
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Check out the code
+      uses: actions/checkout@v1
+
+    - name: Make ci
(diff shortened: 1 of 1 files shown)


nrb:
Once this is merged, I can disable PR builds on Travis via their console. This cannot be controlled via the .travis.yml file.

Also, I _believe_ we need to keep Travis building for pushes enabled until we switch our container publishing over, but I'll double check this and update issue #2104 with details.

nrb:
It looks like the workflow didn't trigger from within the PR. The YAML is basically the same as used in the [CSI plug-in](https://github.com/vmware-tanzu/velero-plugin-for-csi/pull/8) repo but with slightly modified names, so you can look there to see what the interface would be.

skriss:
LGTM, let's see what happens on merge :)

skriss:
great success! https://github.com/vmware-tanzu/velero/commit/2cd0b540bd3d5c27eba5cf4c18d371f582c86e93/checks?check_suite_id=365560773

PR #2157, opened 2019-12-19
bug fix: only prioritize restoring replicasets.apps

### LINKED ISSUE #2156: restores fail on K8s 1.16+ due to invalid replicasets.extensions reference
#2120 fixed a bug where `replicasets.apps` were not properly being restored prior to e.g. Deployments. However, it also introduced a regression for 1.16+ clusters, due to https://github.com/vmware-tanzu/velero/pull/2120/files#diff-cd1a009fa409db8496ea35b53f9776d2R471 which fully qualified `replicasets.extensions`. The problem is that (a) replicasets don't exist in the extensions API group a/o 1.16, and (b) Velero still tries to resolve the group/resource name, even if it's fully specified and there are no resources for it in the backup.

On the backup side, we have code that will always back up replicasets from the "apps" group rather than the "extensions" group, if they both exist (which has been the case for many Kube versions now).

So, it's safe to just prioritize `replicasets.apps` on the restore side here, since we shouldn't have backups with any `replicasets.extensions` *to* restore. That both solves the bug that #2120 was intended to solve, and avoids an invalid reference to `replicasets.extensions` on 1.16+.

Working on testing this now, but any comments welcome.

skriss:

Fixes #2156 

Tested successfully on 1.15.x and 1.16.x KinD clusters, confirming both issues are fixed in both versions.

### DIFF
--- a/changelogs/unreleased/2157-skriss
+++ b/changelogs/unreleased/2157-skriss
@@ -1,0 +1,1 @@
+bug fix: only prioritize restoring `replicasets.apps`, not `replicasets.extensions`
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -451,8 +451,7 @@ func (s *server) validateBackupStorageLocations() error {
 // - Pods go before controllers so they can be explicitly restored and potentially
 //	 have restic restores run before controllers adopt the pods.
 // - Replica sets go before deployments/other controllers so they can be explicitly
-//	 restored and be adopted. Restore from both the 'apps' and 'extensions' API group,
-//   in that order, since they existed in both prior to v1.16.
+//	 restored and be adopted by controllers.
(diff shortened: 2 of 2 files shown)


nrb:
Thanks for jumping on this so quickly!

skriss:
@nrb the restore-blocking bug shouldn't exist in v1.2, only in master

nrb:
>only in master

Good point :) Nevermind!

skriss:
the bug that I did actually fix in #2120 has been around and bothering me for awhile, actually going at least as far back as Joe's TGIK on Velero 1.0 where he had some weird behavior on a restore that seemed impossible, so glad I got that one resolved too :) but it's not a blocking issue and has existed for many releases, so I think it's fine for it to wait until v1.3.

PR #2176, opened 2020-01-07
Assign default CPU/Memory resource value if not present

jenting:
Install restic with CPU/Memory limits is optional.
Therefore, if Velero cannot parse resource requirements, then the message should not bump out the error.

After that, the administrator won't get confused that something recovered failed.

### DIFF
--- a/pkg/restore/restic_restore_action.go
+++ b/pkg/restore/restic_restore_action.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018, 2019 the Velero contributors.
+Copyright 2018, 2019, 2020 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


jenting:
> I would like to suggest a slightly different approach here. Currently, an error will be returned from `kube.ParseResourceRequirements(...)` in both of the following scenarios:
> 
> * the resource requirements were not specified in config
> * the resource requirements were specified in config, but were invalid values

Thanks, I missed the second case.

> In the case where the requests/limits are _partially_ specified, i.e. only some of the 4 fields are provided, we'll still end up with an error from `kube.ParseResourceRequirements(...)`, which I think is fine, but let me know if you see it differently.

Assign to default value if not present. Then reports an error if `kube.ParseResourceRequirements(...)` failed.

Hope this modification covers both scenarios, reminds me if I missed something else :pray:

pkg/restore/restic_restore_action.go
```diff
@@ -106,14 +106,22 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
 
 	cpuRequest, memRequest := getResourceRequests(log, config)
 	cpuLimit, memLimit := getResourceLimits(log, config)
+	if cpuRequest == "" {
```
ashish-amarnath: Can you please update the `Copyright` boilerplate on this file?

jenting: Add 2020 onto this file?

ashish-amarnath: Yes, please.

pkg/restore/restic_restore_action.go
```diff
@@ -106,14 +106,22 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
...
+	}
 
 	resourceReqs, err := kube.ParseResourceRequirements(cpuRequest, memRequest, cpuLimit, memLimit)
 	if err != nil {
-		log.Errorf("Using default resource values, couldn't parse resource requirements: %s.", err)
-		resourceReqs, _ = kube.ParseResourceRequirements(
-			defaultCPURequestLimit, defaultMemRequestLimit, // requests
-			defaultCPURequestLimit, defaultMemRequestLimit, // limits
-		)
+		return nil, errors.Wrapf(err, "couldn't parse resource requirements: %s", err)
```
skriss: I would probably revert this section to the way it was before, i.e. error-log and use the default values, as it's a little more forgiving to the user. Otherwise, LGTM!

jenting: Revert for favor the user.

PR #2180, opened 2020-01-07
update to k8s 1.17 deps and fix breaks

skriss:
This PR:

1. upgrades to the 1.17 versions of the Kubernetes libraries
2. fixes breaking changes related to table printing (i.e. for `velero backup get`)
    - this is contained in commits https://github.com/vmware-tanzu/velero/commit/bc491b098418a4d4f81d1c4580e236903b86ba3e and https://github.com/vmware-tanzu/velero/commit/cf772cb8c838906202aad7a822e88bf8e01a4276
3. drops the `k8s.io/kubernetes` dependency entirely to simplify the future migration to go modules

cc @ashish-amarnath 

TODO:
- [x] finish testing
- [x] fix broken unit tests


nrb:
Will the plugins need similar updates then?

pkg/podexec/pod_command_executor_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 the Velero contributors.
+Copyright 2017, 2020 the Velero contributors.
```
nrb: The modified printer files probably also need this copyright update, too.

skriss: will do

skriss:
> Also, since we're removed printer options from the signature for the printing functions, does this mean that we're modifying our public API/CLI options? I don't really think so, as we didn't expose those options to users anyway. It's not like we really have much of a choice anyway, but figured I'd ask.

The `--show-labels` and `--label-columns` options are still exposed/work as CLI flags, since those are implemented by the table printer itself; none of the other options were exposed to the end-user in any way so I think we're fine here

pkg/podexec/pod_command_executor_test.go
skriss: pushed new commit

pkg/cmd/util/output/schedule_printer.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 the Velero contributors.
+Copyright 2017, 2020 the Velero contributors.
```
ashish-amarnath: Should this be `2017, 2020` or `2017-2020`?

jenting:
Will this PR includes upgrading the CRDs api version from `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1`? Because at Kubernetes 1.16, the `apiextensions.k8s.io/v1beta1` is deprecated and will no longer support at Kubernetes 1.19.

pkg/cmd/util/output/schedule_printer.go
skriss: The guidance we've gotten from legal is to add updated years with a comma. If it was updated in 2018 and 2019 too, we'd list them all out.

skriss:
@nrb @carlisia PTAL

PR #2187, opened 2020-01-13
Pruning unknown fields

jenting:
In CRD apiversion v1beta1, default `preserveUnknownFields=true`.

In CRD apiversion v1, the `preserveUnknownFields=false` only.
Otherwise, the k8s validation bumps out error messages for the
invalid preserveUnknownFields value.

Deploy Velero on k8s 1.16+ with CRD apiversion v1beta1, the
k8s cluster converts apiversion from v1beta1 to v1 automatically.

Fully backup and restore the cluster, restore bumps out error messages
due to the `preserveUnknownFields=true` is not allowed on k8s 1.16+.

Since the CRD structural schema had been defined, enable the `preserveUnknownFields=false` to solves the restore bumps out error message on k8s 1.16+.

Partially fix https://github.com/vmware-tanzu/velero/issues/2159

Reference to https://github.com/kubernetes-csi/external-snapshotter/pull/139/files#r337671120


skriss:
could you also add a changelog? see https://velero.io/docs/master/code-standards/

skriss:
@nrb PTAL when you get a chance, since you've been working on related issues.

skriss:
I tested this on:

- a 1.15 kind cluster - it correctly sets the `spec.preserveUnknownFields` field to `false`, and a basic backup & restore work fine.
- a 1.16 kind cluster - the `spec.preserveUnknownFields` field is correctly set to `false` when accessing the CRD through the `v1beta1` API endpoint, and dropped entirely when accessing the CRD via the `v1` API endpoint (this looks like expected behavior since in v1 the field is deprecated/required to be false for new CRDs), and a basic backup & restore work fine.

I guess the main benefit of this PR is that, for new velero installs on pre-1.16 clusters, this field gets set to `false` which enables a cleaner upgrade when moving to 1.16+ and the `apiextensions.k8s.io/v1` API group/version.

PR #2188, opened 2020-01-13
bump go-hclog and prometheus dependency versions in prep for go modules

skriss:
This PR bumps the version of two dependencies in preparation for go module migration. Separating out into its own PR to keep the subsequent go modules PR as small as possible.


pkg/plugin/clientmgmt/logrus_adapter.go
```diff
@@ -156,3 +156,9 @@ func (l *logrusAdapter) ResetNamed(name string) hclog.Logger {
 func (l *logrusAdapter) StandardLogger(opts *hclog.StandardLoggerOptions) *log.Logger {
 	panic("not implemented")
 }
+
+// Updates the level. This should affect all sub-loggers as well. If an
+// implementation cannot update the level on the fly, it should no-op.
+func (l *logrusAdapter) SetLevel(_ hclog.Level) {
+	return
```
skriss: per the godoc comment on the interface, I left this as a no-op since this wouldn't be something we'd use, but let me know if you see any issues with this.

skriss:
@nrb PTAL when you can, thanks!

PR #2190, opened 2020-01-14
Install server-preferred CRD API version

### LINKED ISSUE #2159: Restore issues with CRDs on Kubernets v1.16.4
Restoring v1beta1 CustomResourceDefinitions into a v1.16 cluster appear to be having some issues. @mauilion reported issues restoring AWS CAPI resources, though I am not able to reproduce with his same backup, on the same version of KinD, with Velero v1.2.0 - though `master` _does_ have issues.

I am, however, consistently seeing errors with Velero's own CRDs on 1.16 clusters:

```
Errors:
  Velero:     <none>
  Cluster:  error restoring customresourcedefinitions.apiextensions.k8s.io/deletebackuprequests.velero.io: the API version in the data (apiextensions.k8s.io/v1) does not match the expected API version (apiextensions.k8s.io/v1beta1)
            error restoring customresourcedefinitions.apiextensions.k8s.io/downloadrequests.velero.io: the API version in the data (apiextensions.k8s.io/v1) does not match the expected API version (apiextensions.k8s.io/v1beta1)
            error restoring customresourcedefinitions.apiextensions.k8s.io/podvolumebackups.velero.io: the API version in the data (apiextensions.k8s.io/v1) does not match the expected API version (apiextensions.k8s.io/v1beta1)
            error restoring customresourcedefinitions.apiextensions.k8s.io/podvolumerestores.velero.io: the API version in the data (apiextensions.k8s.io/v1) does not match the expected API version (apiextensions.k8s.io/v1beta1)
            error restoring customresourcedefinitions.apiextensions.k8s.io/resticrepositories.velero.io: the API version in the data (apiextensions.k8s.io/v1) does not match the expected API version (apiextensions.k8s.io/v1beta1)
            error restoring customresourcedefinitions.apiextensions.k8s.io/restores.velero.io: the API version in the data (apiextensions.k8s.io/v1) does not match the expected API version (apiextensions.k8s.io/v1beta1)
            error restoring customresourcedefinitions.apiextensions.k8s.io/schedules.velero.io: the API version in the data (apiextensions.k8s.io/v1) does not match the expected API version (apiextensions.k8s.io/v1beta1)
            error restoring customresourcedefinitions.apiextensions.k8s.io/serverstatusrequests.velero.io: the API version in the data (apiextensions.k8s.io/v1) does not match the expected API version (apiextensions.k8s.io/v1beta1)
  Namespaces: <none>
```

jenting:
Use Kubernetes discovery API to get the server-preferred API version, and install the server-preferred API version manifest.

Partial fix #2159


skriss:
@jenting thanks for the PR. The main concern I have here is that making this change will prevent Velero from being installed on pre-1.16 clusters that don't have the apiextensions/v1 API group. I think we probably need to do something more intelligent here, like use the Kubernetes discovery API to get the preferred API version for this group, and then use that version when installing.

jenting:
> @jenting thanks for the PR. The main concern I have here is that making this change will prevent Velero from being installed on pre-1.16 clusters that don't have the apiextensions/v1 API group.

I keep the generated CRD manifest on version apiextensions.k8s.io/v1beta1 for backward-compatible for k8s pre-1.16 cluster.

> I think we probably need to do something more intelligent here, like use the Kubernetes discovery API to get the preferred API version for this group, and then use that version when installing.

I will deep more on this :wink: thanks.

carlisia:
@nrb PTAL when you have a chance.

pkg/backup/resource_backupper.go
```diff
@@ -303,8 +303,8 @@ func (rb *defaultResourceBackupper) backupCRD(
 	gr schema.GroupResource,
 	itemBackupper ItemBackupper,
 ) {
-	crdGr := schema.GroupResource{Group: apiextv1beta1.GroupName, Resource: "customresourcedefinitions"}
-	crdClient, err := rb.dynamicFactory.ClientForGroupVersionResource(apiextv1beta1.SchemeGroupVersion,
+	crdGr := schema.GroupResource{Group: apiextv1.GroupName, Resource: "customresourcedefinitions"}
```
skriss: take a look at what I've done in https://github.com/vmware-tanzu/velero/pull/2230 -- I'm using the discovery helper to determine the server-preferred API version, and backing up that one. I think that's the path we'll want to go down here, rather than hardcoding either `v1beta1` or `v1`.

jenting: Thanks for the advice.

jenting: I have reverted the related change in resource_backupper.go and let #2230 address backup issue. This PR address on the install issue only.

carlisia:
I think this looks good. Will defer to @nrb. 

Also, maybe the description could use an update to indicate the server-preferred API version is used for +1.16 (I think this is correct).

hack/crd-gen/main.go
```diff
@@ -75,8 +73,12 @@ func crds() []*apiextv1beta1.CustomResourceDefinition {
 		if err != nil {
 			panic(err)
 		}
-		objs = append(objs, obj.(*apiextv1beta1.CustomResourceDefinition))
+
+		if version == obj.GetObjectKind().GroupVersionKind().GroupVersion().Version {
```
nrb: This is really tough to get right. I _think_ `v1beta1` might actually be correct here, so long as we set the rest of the fields such that they're compatible with `v1`.

My reasoning is because if you run this patch without a Kubernetes cluster to just generate YAML (like for a GitOps workflow), you'll get an error.

```bash
velero install     --provider aws     --plugins velero/velero-plugin-for-aws:v1.0.0     --bucket $BUCKET     --backup-location-config region=$REGION     --snapshot-location-config region=$REGION     --secret-file $HOME/my-secret --dry-run -o yaml
An error occurred: error finding Kubernetes API server config in --kubeconfig, $KUBECONFIG, or in-cluster configuration: invalid configuration: no configuration has been provided
```

On `master`, the same command runs without error and outputs the YAML.

@skriss what are your thoughts currently?

hack/update-generated-crd-code.sh
```diff
@@ -42,7 +42,7 @@ ${GOPATH}/src/k8s.io/code-generator/generate-groups.sh \
   $@
 
 go run ${GOPATH}/src/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go \
-  crd:crdVersions=v1beta1,preserveUnknownFields=false \
+  crd:'crdVersions={v1,v1beta1}' \
```
ashish-amarnath: what would be the version of the CRDs stored?
And do we have to perform any conversions between versions?

jenting: Controller-tools generates both v1 and v1beta1 manifest.
When installing, Velero server pick the server preferred API vesion to install.

hack/crd-gen/main.go
skriss: We definitely need to have a default version of the CRDs that we generate, if, as @nrb says, we're running in dry-run mode and don't necessarily have a cluster that we can connect to to get a server-preferred version.  I would say that default version, if discovery can't be reached/we're in dry-run mode, should be `v1beta1`, as it's most backwards-compatible.

skriss:
I'll do a detailed review shortly, but wanted to make a couple general comments first:

- I don't think this PR is strictly necessary right now, given the other PRs that have gone in around CRD backup/restore. As of right now, with `master`, velero CRDs will be installed as `v1beta1` (which all Kubernetes versions still support), and can be accessed/restored as `v1` CRDs if that is the server-preferred version (i.e. Kubernetes 1.16+).
- at some point, we will want to switch to installing via the `v1` apiextensions endpoint, once all velero-supported Kubernetes versions have it
- this PR should only be necessary if we're in a situation where:
    - velero is still supporting pre-1.16 clusters, i.e. that don't have `v1` apiextensions (so we need to install v1beta1)
    - the newest version of Kubernetes has dropped the `v1beta1` apiextensions version (so we need to install v1)

We shouldn't hit that situation anytime soon, as the `v1beta1` version will be kept around for a minimum of 3 Kubernetes versions (https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api).

That said, I'd be OK getting this PR merged if we can get any issues addressed, ensure it's been well-tested across Kubernetes versions, and we add clear `// TODO` comments indicating what should be removed when `v1beta1` is no longer supported.

Please let me know your thoughts on the above!

skriss:
If we're going to proceed with this, it does need a rebase as there are some conflicts.

jenting:
Thanks for the reviews and the feedback from @nrb.
I agree on @skriss thought, use `v1beta1` version if the server not available.

This PR is _needed_ when `v1beta1` is deprecated (probably at Kubernetes 1.19).
Therefore, my thought is we could _defer_ merged this one until Kubernetes 1.19 released and well-tested on 1.19+ and pre-1.19 clusters.

nrb:
>This PR is needed when v1beta1 is deprecated (probably at Kubernetes 1.19).

At that point, would it be needed, or would we just update to `v1`?  Granted, we could still run into the issue if/when CRDs `v2` came out, so perhaps it would, and we would just update our fallback to `v1` instead of `v1beta1`.

It probably comes down to defining Velero's support window - would we support versions less than v1.16 and equal to or greater than v1.19 at the same time?

skriss:
> It probably comes down to defining Velero's support window - would we support versions less than v1.16 and equal to or greater than v1.19 at the same time?

Yep, agree.

Like I said, I'm not opposed to going fwd with this as it potentially insulates us from a future issue, but it's also not urgent. I'll do a proper review later today.

ashish-amarnath:
@jenting and @nrb 
Can this PR be closed?

jenting:
I think it could be close since Velero supports Kubernetes backward-compatible to 3 versions which are 1.16, 1.17, 1.18 for now, all are version 1.16+ so we should not have the installation and backup/restore issues related to server-preferred CRD API version.

PR #2196, opened 2020-01-14
recompute backup_last_successful_timestamp metric during resync

### LINKED ISSUE #1777: velero_backup_last_successful_timestamp metric vanishes when velero pod is restarted
I'd like to be able to monitor the state of backups in a cluster. Instead of worrying about successful/failed backups I'd like to have a monitor that only alerts if the last successful backup is older than some arbitrary amount of time. The metric velero_backup_last_successful_timestamp should allow for this. However, when velero is restarted/rescheduled, this metric is no longer emitted until another successful backup happens. This makes it difficult to build a backup-age based alert. 

Velero should make a best effort attempt to recover this metric by examining the backup objects in the cluster and setting velero_backup_last_successful_timestamp to the completion timestamp of the most recent successful backup

**Environment:**

- Velero version (use `velero version`): v1.0.0
- Kubernetes version (use `kubectl version`): v1.14.6
- Kubernetes installer & version: N/A
- OS (e.g. from `/etc/os-release`): Ubuntu

skriss:

closes #1777 

Open to moving this code around per https://github.com/vmware-tanzu/velero/issues/1777#issuecomment-574361892, but this seems like the simplest approach to me and is consistent with how existing gauge metrics are reconstructed.

Did some basic testing and confirmed the metrics re-populate after a server restart.

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -158,6 +158,16 @@ func (c *backupController) resync() {
 
 	// recompute backup_last_successful_timestamp metric for each
 	// schedule (including the empty schedule, i.e. ad-hoc backups)
+	for schedule, timestamp := range getLastSuccessBySchedule(backups) {
+		c.metrics.SetBackupLastSuccessfulTimestamp(schedule, timestamp)
+	}
+}
+
+// getLastSuccessBySchedule finds the most recent completed backup for each schedule
+// and returns a map of schedule name -> completion time of the most recent completed
+// backup. This map includes an entry for ad-hoc/non-scheduled backups, where the key
+// is the empty string.
+func getLastSuccessBySchedule(backups []*velerov1api.Backup) map[string]time.Time {
 	lastSuccessBySchedule := map[string]time.Time{}
 	for _, backup := range backups {
 		if backup.Status.Phase != velerov1api.BackupPhaseCompleted {
@@ -175,9 +185,7 @@ func (c *backupController) resync() {
 		}
 	}
 
-	for schedule, timestamp := range lastSuccessBySchedule {
-		c.metrics.SetBackupLastSuccessfulTimestamp(schedule, timestamp)
-	}
+	return lastSuccessBySchedule
 }
 
 func (c *backupController) processBackup(key string) error {
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -148,11 +148,35 @@ func NewBackupController(
 }
 
 func (c *backupController) resync() {
+	// recompute backup_total metric
 	backups, err := c.lister.List(labels.Everything())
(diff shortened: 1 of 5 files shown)


skriss:
@carlisia @nrb @ashish-amarnath PTAL

pkg/controller/backup_controller.go
```diff
@@ -148,12 +148,36 @@ func NewBackupController(
...
+	lastSuccessBySchedule := map[string]time.Time{}
+	for _, backup := range backups {
+		if backup.Status.Phase != velerov1api.BackupPhaseCompleted {
+			continue
+		}
+		if backup.Status.CompletionTimestamp == nil {
+			continue
+		}
+
+		schedule := backup.Labels[velerov1api.ScheduleNameLabel]
```
ashish-amarnath: Can the `schedule` be `empty` or `nil`?

pkg/controller/backup_controller.go
```diff
@@ -148,12 +148,36 @@ func NewBackupController(
...
 
 func (c *backupController) resync() {
+	// recompute backup_total metric
 	backups, err := c.lister.List(labels.Everything())
 	if err != nil {
 		c.logger.Error(err, "Error computing backup_total metric")
 	} else {
 		c.metrics.SetBackupTotal(int64(len(backups)))
 	}
+
```
ashish-amarnath: Error message on line#154 seems out of place, no?

pkg/controller/backup_controller.go
```diff
@@ -148,12 +148,36 @@ func NewBackupController(
...
+	// recompute backup_last_successful_timestamp metric for each
+	// schedule (including the empty schedule, i.e. ad-hoc backups)
+	lastSuccessBySchedule := map[string]time.Time{}
+	for _, backup := range backups {
```
ashish-amarnath: suggest pulling this loop into a `func` of its own that can be unit tested.

pkg/controller/backup_controller.go
skriss: yes, it can be empty if it's for an ad-hoc backup; we use that as a label value for all metrics, so this should be correct. I also tested with ad-hoc backups and it looks as expected.

pkg/controller/backup_controller.go
skriss: how so? In any case, unrelated to my PR :)

pkg/controller/backup_controller.go
skriss: sure, will do

skriss: done, see latest commit

pkg/controller/backup_controller.go
ashish-amarnath: Yes, totally unrelated to this PR.

PR #2197, opened 2020-01-14
Fix restore issues for CRD v1 on Kubernetes v1.16 clusters

nrb:
Partial fix for #2159 

Using this patch moves the restore issues beyond version mismatches, however if CRDs in the backup have `Spec.PreserveUnknownFields = True` in the backup, whether they were v1beta1 or v1, they will see these errors if only this change is applied.

```
Errors:
  Velero:     <none>
  Cluster:  error restoring customresourcedefinitions.apiextensions.k8s.io/deletebackuprequests.velero.io: CustomResourceDefinition.apiextensions.k8s.io "deletebackuprequests.velero.io" is invalid: spec.preserveUnknownFields: Invalid value: true: cannot set to true, set x-preserve-unknown-fields to true in spec.versions[*].schema instead
            error restoring customresourcedefinitions.apiextensions.k8s.io/downloadrequests.velero.io: CustomResourceDefinition.apiextensions.k8s.io "downloadrequests.velero.io" is invalid: spec.preserveUnknownFields: Invalid value: true: cannot set to true, set x-preserve-unknown-fields to true in spec.versions[*].schema instead
            error restoring customresourcedefinitions.apiextensions.k8s.io/podvolumebackups.velero.io: CustomResourceDefinition.apiextensions.k8s.io "podvolumebackups.velero.io" is invalid: spec.preserveUnknownFields: Invalid value: true: cannot set to true, set x-preserve-unknown-fields to true in spec.versions[*].schema instead
            error restoring customresourcedefinitions.apiextensions.k8s.io/podvolumerestores.velero.io: CustomResourceDefinition.apiextensions.k8s.io "podvolumerestores.velero.io" is invalid: spec.preserveUnknownFields: Invalid value: true: cannot set to true, set x-preserve-unknown-fields to true in spec.versions[*].schema instead
            error restoring customresourcedefinitions.apiextensions.k8s.io/resticrepositories.velero.io: CustomResourceDefinition.apiextensions.k8s.io "resticrepositories.velero.io" is invalid: spec.preserveUnknownFields: Invalid value: true: cannot set to true, set x-preserve-unknown-fields to true in spec.versions[*].schema instead
            error restoring customresourcedefinitions.apiextensions.k8s.io/restores.velero.io: CustomResourceDefinition.apiextensions.k8s.io "restores.velero.io" is invalid: spec.preserveUnknownFields: Invalid value: true: cannot set to true, set x-preserve-unknown-fields to true in spec.versions[*].schema instead
            error restoring customresourcedefinitions.apiextensions.k8s.io/schedules.velero.io: CustomResourceDefinition.apiextensions.k8s.io "schedules.velero.io" is invalid: spec.preserveUnknownFields: Invalid value: true: cannot set to true, set x-preserve-unknown-fields to true in spec.versions[*].schema instead
            error restoring customresourcedefinitions.apiextensions.k8s.io/serverstatusrequests.velero.io: CustomResourceDefinition.apiextensions.k8s.io "serverstatusrequests.velero.io" is invalid: spec.preserveUnknownFields: Invalid value: true: cannot set to true, set x-preserve-unknown-fields to true in spec.versions[*].schema instead
```

### DIFF
--- a/pkg/restore/crd_v1_preserve_unknown_fields_action.go
+++ b/pkg/restore/crd_v1_preserve_unknown_fields_action.go
@@ -69,7 +69,7 @@ func (c *CRDV1PreserveUnknownFieldsAction) Execute(input *velero.RestoreItemActi
 	// The v1 API doesn't allow the PreserveUnknownFields value to be true, so make sure the schema flag is set instead
 	if crd.Spec.PreserveUnknownFields {
 		// First, change the top-level value since the Kubernetes API server on 1.16+ will generate errors otherwise.
-		log.Info("Set PreserveUnknownFields to False")
+		log.Debug("Set PreserveUnknownFields to False")
 		crd.Spec.PreserveUnknownFields = false
 
 		// Make sure all versions are set to preserve unknown fields
@@ -77,7 +77,7 @@ func (c *CRDV1PreserveUnknownFieldsAction) Execute(input *velero.RestoreItemActi
 			// Use the address, since the XPreserveUnknownFields value is undefined or true (false is not allowed)
 			preserve := true
 			v.Schema.OpenAPIV3Schema.XPreserveUnknownFields = &preserve
-			log.Infof("Set x-preserve-unknown-fields in Open API for schema version %s", v.Name)
+			log.Debugf("Set x-preserve-unknown-fields in Open API for schema version %s", v.Name)
 		}
 	}
 
--- a/pkg/restore/crd_v1_preserve_unknown_fields_action.go
+++ b/pkg/restore/crd_v1_preserve_unknown_fields_action.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2020 the Velero contributors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	 http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package restore
+
+import (
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+)
+
+// The CRDV1PreserveUnknownFieldsAction will take a CRD and inspect it for the API version and the PreserveUnknownFields value.
+// If the API Version is 1 and the PreserveUnknownFields value is True, then the x-preserve-unknown-fields value in the OpenAPIV3 schema will be set to true in order to allow Kubernetes 1.16+ servers to accept the object.
+type CRDV1PreserveUnknownFieldsAction struct {
+	logger logrus.FieldLogger
+}
+
+func NewCRDV1PreserveUnknownFieldsAction(logger logrus.FieldLogger) *CRDV1PreserveUnknownFieldsAction {
+	return &CRDV1PreserveUnknownFieldsAction{logger: logger}
+}
+
+func (c *CRDV1PreserveUnknownFieldsAction) AppliesTo() (velero.ResourceSelector, error) {
+	return velero.ResourceSelector{
+		IncludedResources: []string{"customresourcedefinition.apiextensions.k8s.io"},
+	}, nil
+}
+
+func (c *CRDV1PreserveUnknownFieldsAction) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
+	c.logger.Info("Executing CRDV1PreserveUnknownFieldsAction")
+
+	log := c.logger.WithField("plugin", "CRDV1PreserveUnknownFieldsAction")
+
+	version, _, err := unstructured.NestedString(input.Item.UnstructuredContent(), "apiVersion")
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get CRD version")
+	}
+
+	// We don't want to "fix" anything in beta CRDS at the moment, just v1 versions with preserveunknownfields = true
+	if version == "apiextensions.k8s.io/v1beta1" {
+		return &velero.RestoreItemActionExecuteOutput{
+			UpdatedItem: input.Item,
+		}, nil
+	}
+
+	var crd apiextv1.CustomResourceDefinition
+
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.Item.UnstructuredContent(), &crd); err != nil {
+		return nil, errors.Wrap(err, "unable to conver unstructured item to custom resource definition")
+	}
+
+	// The v1 API doesn't allow the PreserveUnknownFields value to be true, so make sure the schema flag is set instead
+	if crd.Spec.PreserveUnknownFields {
+		// First, change the top-level value since the Kubernetes API server on 1.16+ will generate errors otherwise.
+		log.Info("Set PreserveUnknowFields to False")
+		crd.Spec.PreserveUnknownFields = false
+
+		// Make sure all versions are set to preserve unknown fields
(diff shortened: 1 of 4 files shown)


skriss:
Code change LGTM. Re: preserveUnknownFields - I see, so if you get a CRD from the v1 API, it can have this set to true, but you can't apply a new CRD via the v1 API with this set to true. Yeah, sounds like we need a plugin to handle this (which I think you're already working on).

nrb:
> I see, so if you get a CRD from the v1 API, it can have this set to true, but you can't apply a new CRD via the v1 API with this set to true. Yeah, sounds like we need a plugin to handle this (which I think you're already working on).

Yeah, the field needs to be set in the JSON schema, but I'm not 100% sure where yet. I'm still experimenting. If I simply set the top level `PreserveUnknownFields` value to false, I start seeing other errors that appear to show that the object isn't being fully reset by Velero - for example, the `ResourceVersion` is still set somehow.

So, even if I get a plugin written to move this particular field, I don't think that's the whole story here.

nrb:
Updated PR with the plugin. The resourceVersion issues were due to a logic error on my part; the plugin was using the item from the backup, which meant that it contained all the metadata that Velero had already reset.

pkg/restore/restore.go
```diff
@@ -379,6 +379,7 @@ type context struct {
 
 type resourceClientKey struct {
 	resource  schema.GroupResource
+	version   schema.GroupVersion
```
skriss: I'd probably just change `resource` to be a `schema.GroupVersionResource` so we don't need an additional field that's partially redundant.

pkg/restore/restore.go
```diff
@@ -711,6 +712,7 @@ func (ctx *context) restoreResource(resource, targetNamespace, originalNamespace
 func (ctx *context) getResourceClient(groupResource schema.GroupResource, obj *unstructured.Unstructured, namespace string) (client.Dynamic, error) {
 	key := resourceClientKey{
 		resource:  groupResource,
```
skriss: if you make the above change, then resource can be set to `groupResource.WithVersion(obj.GroupVersionKind().Version)`

pkg/restore/crd_v1_preserve_unknown_fields_action.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/sirupsen/logrus"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+)
+
+// The CRDV1PreserveUnknownFieldsAction will take a CRD and inspect it for the API version and the PreserveUnknownFields value.
+// If the API Version is 1 and the PreserveUnknownFields value is True, then the x-preserve-unknown-fields value in the OpenAPIV3 schema will be set to true in order to allow Kubernetes 1.16+ servers to accept the object.
```
skriss: let's mention that the  `PreserveUnknownFields` field will be set to false too (which is the actual thing that allows the object to be accepted)

pkg/restore/crd_v1_preserve_unknown_fields_action.go
```diff
@@ -0,0 +1,91 @@
...
+	if version == "apiextensions.k8s.io/v1beta1" {
+		return &velero.RestoreItemActionExecuteOutput{
+			UpdatedItem: input.Item,
+		}, nil
+	}
+
+	var crd apiextv1.CustomResourceDefinition
+
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.Item.UnstructuredContent(), &crd); err != nil {
+		return nil, errors.Wrap(err, "unable to conver unstructured item to custom resource definition")
```
skriss: nit: conver -> convert

pkg/restore/crd_v1_preserve_unknown_fields_action.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	// The v1 API doesn't allow the PreserveUnknownFields value to be true, so make sure the schema flag is set instead
+	if crd.Spec.PreserveUnknownFields {
+		// First, change the top-level value since the Kubernetes API server on 1.16+ will generate errors otherwise.
+		log.Info("Set PreserveUnknowFields to False")
```
skriss: nit: PreserveUnknowFields -> PreserveUnknownFields

skriss:
@carlisia PTAL

skriss:
LGTM. were you able to retest after the change to the key for the client cache? Just wanna confirm my suggestion actually works :)

nrb:
> were you able to retest after the change to the key for the client cache? 

I have not re-tested yet

nrb:
I've now re-tested this with the key update, it still works as expected.

skriss:
@carlisia / @ashish-amarnath PTAL

pkg/restore/crd_v1_preserve_unknown_fields_action.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+)
+
+// The CRDV1PreserveUnknownFieldsAction will take a CRD and inspect it for the API version and the PreserveUnknownFields value.
```
carlisia: Super nit: the documentation should start with `CRDV1PreserveUnknownFieldsAction`.

carlisia:
Would it be worth it to change the title of this PR? Seems mismatched.

nrb:
> Would it be worth it to change the title of this PR? Seems mismatched.

Updated; it started out as a simple key change, then I needed to add the plugin :(

skriss:
@nrb looks like this one is just waiting for a godoc change, then good to merge: https://github.com/vmware-tanzu/velero/pull/2197#discussion_r369252555

pkg/restore/crd_v1_preserve_unknown_fields_action.go
ashish-amarnath:
```suggestion
// CRDV1PreserveUnknownFieldsAction will take a CRD and inspect it for the API version and the PreserveUnknownFields value.
```

ashish-amarnath: made it easier to change!

pkg/restore/crd_v1_preserve_unknown_fields_action.go
```diff
@@ -0,0 +1,92 @@
...
+
+	log := c.logger.WithField("plugin", "CRDV1PreserveUnknownFieldsAction")
+
+	version, _, err := unstructured.NestedString(input.Item.UnstructuredContent(), "apiVersion")
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get CRD version")
+	}
+
+	// We don't want to "fix" anything in beta CRDS at the moment, just v1 versions with preserveunknownfields = true
+	if version == "apiextensions.k8s.io/v1beta1" {
```
ashish-amarnath: suggest making this more robust using a more specific check with a `switch`  
```
case "apiextensions.k8s.io/v1":
    // set XPreserveUnknownFields for all the versions
default:
    log ("not doing anything")
```

skriss:
latest commit needs a signoff :/

pkg/restore/crd_v1_preserve_unknown_fields_action.go
nrb: I think it's a good point that we should be inspecting for the v1 version rather than `v1beta1`, since the check now would try to process alpha or v2 CRDs, should they ever get passed in.

However, I do prefer the `if` statement that simply returns because it keeps the main logic of the function less nested.

How about the check is inverted to `if version != "apiextensions.k8s.io/v1"`?

nrb:
> latest commit needs a signoff :/

Yeah, pushing a fix in a sec. I thought the new DCO checker would fix it, not realizing accepting GH suggestions simply leave out a sign off entirely. :(

skriss:
Yeah, I've noticed that before, it's annoying.

skriss:
@carlisia @ashish-amarnath please take another look here, thanks!

pkg/restore/crd_v1_preserve_unknown_fields_action.go
ashish-amarnath: Works for me.

ashish-amarnath:
LGTM

PR #2200, opened 2020-01-15
Check for nil LastMaintenanceTime in dueForMaintenance

sseago:
ResticRepository.dueForMaintenance causes a panic in the velero pod
("invalid memory address or nil pointer dereference") if
repository.Status.LastMaintenanceTime is nil. This fix returns 'true'
if it's nil, so the repository is due for maintenance if LastMaintenanceTime
is nil *or* the time elapsed since the last maintenance is greater than
repository.Spec.MaintenanceFrequency.Duration

### DIFF
--- a/changelogs/unreleased/2200-sseago
+++ b/changelogs/unreleased/2200-sseago
@@ -1,0 +1,1 @@
+Bug fix: Check for nil LastMaintenanceTime in ResticRepository dueForMaintenance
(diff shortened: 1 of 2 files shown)


skriss:
got another report of this in #2204 , we probably want to backport this to v1.2.1 when we ship it (cc @nrb)

Issue #1910, opened 2019-09-25
Revisit install & configuration UX


skriss:
Moved/created all items that we decided to address in the near term into a new Epic here: https://github.com/vmware-tanzu/velero/issues/2419

Issues not part of that effort were left attached to this Epic.

Original post:
---
A number of issues/questions have come up over the past weeks as we've planned for moving the in-tree cloud provider plugins out:

- should plugins be able to be installed as part of `velero install`, or separately?
- should `velero install` support configuring arbitrary BSL/VSL combinations (i.e. any number of each, different providers, etc.)?
- how should `velero install` be aware of env vars that different provider plugins require? (right now we just hardcode the AWS/Azure/GCP ones)
- can we clarify naming -- image, provider, plugin name, etc.?
- there's currently a special case for Azure + restic, where a set of restic-specific env vars need to be set with storage account name and key, which is likely going to remain hardcoded in upstream Velero - how do we deal with situations like this?
- should the velero server be able to start up with no BSLs/VSLs defined? If so, how do we communicate to the user that it's "running" but has no configuration yet?

We need to take a holistic look at the install & configuration experience and think about what we want it to look like. Many of these changes may be breaking.

skriss:
xref #1853

nrb:
Would it perhaps make sense to update the BackupStorageLocation/VolumeSnapshotLocation `Spec.Provider` field to be called `Spec.Plugin`? This would make the connection between the location and the plugin that utilizes its information much more clear.

This doesn't solve the ambiguity that the value in that field is a name within the plugin code and only exposed via documentation, however.

If this is something others would like, I would think we could deprecate `provider` and write `plugin` with `velero install`, then remove `provider` completely in 2.0.

skriss:
xref #1967

skriss:
User feedback from Slack:

"I originally installed it when following docs for 1.1.0, which was straight forward.
before you had examples for aws, gcp, etc, on the guide.
Now the guide is disjointed, and leaves you to guess a few things.
I think at very least the section for ‘Install and configure the server components’, where it says “To install and configure the Velero server components, follow the provider-specific instructions documented by your storage provider.” My version would be broken, bad english with run-ons. lol but would be nice if it said to explicitly say to expect to find the documentation in repo for storage provider plugin of your choice.
Clicking on the “your storage provider” and the ’Providers” page just lists plugins, and links for plugin documentation. But it doesn’t say anything about installation instructions anywhere.
[...]

PR #2202, opened 2020-01-15
Initial redesign of CLI commands

carlisia:
TODOs:

- [x] List new, renamed and reorganized commands and flags
- [x] List command/flags from `velero install` and `velero server` to be: deprecated/deleted, moved.
- [x] How to handle GitOps users (https://github.com/vmware-tanzu/velero/pull/2202#discussion_r369137959) - Discussion moved to this issue: https://github.com/vmware-tanzu/velero/issues/2405
- [x] List alternative example CLIs considered (https://github.com/vmware-tanzu/velero/pull/2202#issuecomment-575677477)
- [x] Document how to create/mount credentials secrets & set environment variables (https://github.com/vmware-tanzu/velero/pull/2202#discussion_r367488591) - Design for this moved to this PR: https://github.com/vmware-tanzu/velero/pull/2403
- [x] Document change from `provider` to `plugin`
- [x] Document behavior changes to accommodate new command structure

Ref: https://github.com/vmware-tanzu/velero/issues/1910


design/cli-install-changes.md
```diff
@@ -0,0 +1,156 @@
+# Proposal for a more intuitive CLI to install and configure Velero
+
+Currently, the Velero CLI tool has a `install` command that configures numerous major and minor aspects of Velero. As a result, the combined set of flags for this `install` command makes it hard to intuit and reason about the different Velero components. This document proposes changes to improve the UX for installation and configuration in a way that would make it easier for the user to discover what needs to be configured by looking at what is available in the CLI rather then having to rely heavily on our documentation for the usage. At the same time, it is expected that the documentation update to reflect these changes will also make the documentation flow easier to follow.
+
+This proposal prioritizes discoverability and self-documentation over minimalizing length or number of commands and flags.
+
+## Goals
```
skriss: I think we need to consider impacts on GitOps users as well - right now `velero install` allows users to generate a complete set of YAML that can be checked into git and applied as the user sees fit; breaking things up into multiple commands likely makes it more difficult/not possible for those users to get a complete set of YAML without applying it to the cluster.  I would probably add a goal around that.

design/cli-install-changes.md
```diff
@@ -0,0 +1,156 @@
...
+```
+      --pod-annotations mapStringString            annotations to add to the Velero and restic pods. Optional. Format is key1=value1,key2=value2
+      --restore-only                               run the server in restore-only mode. Optional.
+      --pod-cpu-limit string                       CPU limit for Velero pod. A value of "0" is treated as unbounded. Optional. (default "1000m")
+      --pod-cpu-request string                     CPU request for Velero pod. A value of "0" is treated as unbounded. Optional. (default "500m")
+      --pod-mem-limit string                       memory limit for Velero pod. A value of "0" is treated as unbounded. Optional. (default "256Mi")
+      --pod-mem-request string                     memory request for Velero pod. A value of "0" is treated as unbounded. Optional. (default "128Mi")
+```
+
+3) `velero config`
```
skriss: some of this overlaps with the existing `velero backup-location get/create` and `velero snapshot-location get/create` commands - thoughts on how they relate?

design/cli-install-changes.md
```diff
@@ -0,0 +1,156 @@
...
+      --use-snapshot-locations mapStringString          sets the list of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...) (was `default-volume-snapshot-locations`)
+
+      --set-default-location                            configuration to create a default locations. Format is bucket=value,prefix=value,plugin-name=value,snapshot-location=true/false. Optional.
+
+      --provider string                                 provider name for backup and volume storage - (DEPRECATED)
+      --no-default-backup-location                      flag indicating if a default backup location should be created. Must be used as confirmation if --bucket or --provider are not provided. Optional. (DEPRECATED)
+      --use-volume-snapshots                            whether or not to create snapshot location automatically. Set to false if you do not plan to create volume snapshots via a storage provider. (default true) (DEPRECATED)
+```
+
+4) `velero plugin`
```
skriss: I wonder if as part of adding plugins, we should include things like creating/mounting credentials secrets & setting environment variables, since they all go together with enabling a particular plugin to work.

design/cli-install-changes.md
```diff
@@ -0,0 +1,156 @@
...
+      --pod-cpu-limit string                      CPU limit for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+      --pod-cpu-request string                    CPU request for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+      --pod-mem-limit string                      memory limit for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+      --pod-mem-request string                    memory request for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+      --create                                    create restic deployment. Optional. (was `use-restic`)
+      --repo                                      Work with restic repositories
+      --restic-timeout duration                   how long backups/restores of pod volumes should be allowed to run before timing out (default 1h0m0s)
+```
+
+#### Example
```
skriss: reiterating previous comment, we need to think about what this looks like for a GitOps user.

design/cli-install-changes.md
skriss: I think it might make sense to have a goal around having our install/config UX be ~consistent with the rest of the ecosystem. Related to this, it would probably be helpful (and this is something I'd like to do myself, regardless) to survey some other popular Kubernetes operators/extensions and see what their install UX looks like.

design/cli-install-changes.md
carlisia: Good observation. I would deprecate those and have all of those operations only under `velero config`.

design/cli-install-changes.md
carlisia: Yes, this is the direction I would like this design to accommodate, group all things that go together.

design/cli-install-changes.md
```diff
@@ -0,0 +1,156 @@
...
+      --provider string                                 provider name for backup and volume storage - (DEPRECATED)
+      --no-default-backup-location                      flag indicating if a default backup location should be created. Must be used as confirmation if --bucket or --provider are not provided. Optional. (DEPRECATED)
+      --use-volume-snapshots                            whether or not to create snapshot location automatically. Set to false if you do not plan to create volume snapshots via a storage provider. (default true) (DEPRECATED)
+```
+
+4) `velero plugin`
+Component specific configuration for plugins.
+
+```
+      --add stringArray                           Add plugin container images to install into the Velero Deployment
```
nrb: For adding, listing, and removing, I think I would prefer these to be subcommands instead of flags. So `velero plugin add`, `velero plugin list`, `velero plugin remove`.

My reasoning here is that these are all verbs. Flags/arguments are, generally, not verbs. This is also in line with how `kubectl` works.

The same goes for actions/verbs on other commands.

design/cli-install-changes.md
```diff
@@ -0,0 +1,156 @@
...
+      --get-snapshot-locations                          Display snapshot locations
+      --current-snapshot-locations                      displays the current default volume snapshot locations (NEW)
+      --use-snapshot-locations mapStringString          sets the list of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...) (was `default-volume-snapshot-locations`)
+
+      --set-default-location                            configuration to create a default locations. Format is bucket=value,prefix=value,plugin-name=value,snapshot-location=true/false. Optional.
+
+      --provider string                                 provider name for backup and volume storage - (DEPRECATED)
+      --no-default-backup-location                      flag indicating if a default backup location should be created. Must be used as confirmation if --bucket or --provider are not provided. Optional. (DEPRECATED)
+      --use-volume-snapshots                            whether or not to create snapshot location automatically. Set to false if you do not plan to create volume snapshots via a storage provider. (default true) (DEPRECATED)
```
nrb: Would the deprecation here be on the `velero install` command, since that's where the flag original came from?

carlisia: Yes. I haven't done a good job with the flags to be removed yet. I'd have a section just for that so it's clear.

carlisia:
A video session we had about this revamp: https://www.youtube.com/watch?v=ErjPS6tUsWs

skriss:
Here are my quick notes from looking at a few different extensions/tools' install & config UX:

Contour -- static YAML
- https://projectcontour.io/getting-started/
- https://projectcontour.io/docs/v1.1.0/deploy-options/

AWS EBS CSI Driver -- static YAML with kustomize
- https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/README.md
- https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/deploy/kubernetes

Istio -- CLI command with generic --set flag for customizing, or Helm
- https://istio.io/docs/setup/
- istio manifest generate/apply (with --set flag for customization)

Cert Manager -- static YAML
- https://cert-manager.io/docs/installation/kubernetes/
- Install vs. configure
- Configure backends -- https://cert-manager.io/docs/configuration/

design/cli-install-changes.md
nrb: Here's some docs on kubectl's integration with kustomize - https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/

nrb: What I envision for GitOps here: `velero init` would produce a base YAML that users would deploy, and the plugins would be layers that users could deploy with `kubectl apply -k` that customizes the base deployment with the appropriate init containers, secrets, etc.

design/cli-install-changes.md
```diff
@@ -0,0 +1,156 @@
...
+      --show-labels                                show labels in the last column
+```
+
+Minimum set for initialization
+```
+      --image string                               image to use for the Velero and restic server pods. Optional. (default "velero/velero:latest")
+      --sa-annotations mapStringString             annotations to add to the Velero ServiceAccount. Add iam.gke.io/gcp-service-account=[GSA_NAME]@[PROJECT_NAME].iam.gserviceaccount.com for workload identity. Optional. Format is key1=value1,key2=value2
+```
+   
+2) `velero set`
```
ashish-amarnath: I think it would be useful to have a `get` equivalent of these commands. What do other folks think?

design/cli-install-changes.md
```diff
@@ -0,0 +1,156 @@
...
+      --get-backup-locations                            Display backup storage locations
+      --current-backup-location                         displays the current default backup storage location (NEW)
+      --use-backup-location string                      sets the default backup storage location (default "default") (was `default-backup-storage-location`)
+
+      --snapshot-location  mapStringString              configuration to use for creating a volume snapshot location. Format is key1=value1,key2=value2 (was `snapshot-location-config`)
+      --get-snapshot-locations                          Display snapshot locations
+      --current-snapshot-locations                      displays the current default volume snapshot locations (NEW)
+      --use-snapshot-locations mapStringString          sets the list of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...) (was `default-volume-snapshot-locations`)
+
+      --set-default-location                            configuration to create a default locations. Format is bucket=value,prefix=value,plugin-name=value,snapshot-location=true/false. Optional.
```
ashish-amarnath: `location` may be be overloaded. Can it be called something else?
Also is this the default `backup-location` or `snapshot-location`

design/cli-install-changes.md
ashish-amarnath: something like `kubectl cluster-info` and `kubectl version`

ashish-amarnath:
Wanted to put this suggestion out there. How do you all feel about organizing commands like:
```
velero [resource] [operation] [flags]
```
where `resource` are things like, `plugins`, `backup`, `restore`, `config`
and the `operations` can be customized based on the resource, e.g. `velero plugin add`, `velero plugin remove`, `velero plugin configure`, `velero plugin list`, `velero config get`, `velero config set`

design/cli-install-changes.md
carlisia: I think after this revamp is finished, we should go through all commands and see how we can improve them. This is definitely one opportunity for improvement.

design/cli-install-changes.md
carlisia: Just FYI, this is going to be removed.

carlisia:
> For the migration plan, we could host the new commands under, say, `new` subcommand and add deprecated message to the existing ones. Keep them that way for a release and then remove the `new` subcommand.

This is an interesting suggestion. I'll add this to the community meeting to get more opinions.

carlisia:
> Wanted to put this suggestion out there. How do you all feel about organizing commands like:
> 
> ```
> velero [resource] [operation] [flags]
> ```
> 
> where `resource` are things like, `plugins`, `backup`, `restore`, `config`, `backup-storage-location`, `volume-snapshot-location`
> and the `operations` can be customized based on the resource, e.g. `velero plugin add`, `velero plugin remove`, `velero plugin configure`, `velero plugin list`, `velero config get`, `velero config set`

I'd like to make a case for following this format:

```
velero [operation] [resource] [flags]
```

I'm thinking of three reasons:
1) the list of verbs will (likely) be shorter than the list of resources. It's mostly always create/get/delete/list/set whereas we can have numerous resources.
2) `kubectl --help` lists the verbs/actions, not the resources, so we would be in the same style as that
3) Velero also lists verbs/actions in the help, and not the resources

However, the new commands would follow the Velero convention of allowing to use either verb - resource OR resource -verb.

design/cli-install-changes.md
carlisia: To keep in line with the `velero [resource] [operation] [flags]` format, I had to remove the `velero config` and so we are keeping the original `velero backup-location/snapshot-location` with some minor flag modifications/additions.

design/cli-install-changes.md
carlisia: Added to TODO list.

design/cli-install-changes.md
carlisia: Added to TODO list.

design/cli-install-changes.md
carlisia: Done.

design/cli-install-changes.md
carlisia: Documented now.

design/cli-install-changes.md
carlisia: Added to TODO list.

carlisia:
Update: we decided not to use the keyword `new`, and to follow the format `velero [resource] [operation] [flags]`.

carlisia:
Notable highlights of this PR as up-to-date:
- Moved all flags from `velero server` to under `velero init`. They all seem to belong as a group. There were some exceptions documented in the Deprecation section here: https://github.com/vmware-tanzu/velero/pull/2202/files#diff-828dbe24187f3191e6f631dc81abab62R250

- Had to remove `velero config` to keep inline with the `velero [resource] [operation] [flags]` format. This resulted in needing to keep `velero backup-location` and `velero snapshot-location`, with some minor flag additions/modifications.

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+To conform with Velero's current practice, these commands will also work by swapping the operation/resource.
+
+Below is the proposed set of new commands to setup and configure Velero. 
+
+1) `velero init`
+
+```
+      --server                                     Configures up the namespace, RBAC, deployment, etc., but does not add any external plugins, BSL/VSL definitions. This would be the minimum set of commands to get the Velero server up and running and ready to accept other configurations.
```
skriss: What happens if you run `velero init` without the `--server` flag? In my mind, `velero init` always initializes the server so this flag seems like it might be redundant, but I'm probably missing something.

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+      get                                                 Display snapshot locations
+        --default                                         list of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...) (NEW -- was `server --default-volume-snapshot-locations`))
+```
+
+4) `velero plugin`
+Configuration for plugins.
+
+```
+      add 
+        --images stringArray                    add plugin container images to install into the Velero Deployment
```
skriss: Currently, the image name is specified not as a flag, but as an argument the command (e.g. `velero plugin add velero/velero-plugin-for-aws:v1.0.0`).  I'd say this is a standard in our CLI (and others as well), where the "object" of the command is specified not as a flag, but as an argument to the command. It's similar to `velero backup get my-backup`, or `kubectl get pod my-pod`, as opposed to something like `velero backup get --name my-backup` or `kubectl get pod --name my-pod`

I prefer that pattern for consistency and clarity -- so the "object" of a command is an argument, and flags are strictly for modifiers. WDYT?

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+```
+    set
+      --default-prune-frequency duration          how often 'restic prune' is run for restic repositories by default. Optional.
+      --pod-annotations mapStringString           annotations to add to the Velero and restic pods. Optional. Format is key1=value1,key2=value2
+      --pod-cpu-limit string                      CPU limit for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+      --pod-cpu-request string                    CPU request for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+      --pod-mem-limit string                      memory limit for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+      --pod-mem-request string                    memory request for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+      --deployment                                create restic deployment. Defauylt is false. Optional. Other lags will only work if set to true. (NEW, was `velero install use-restic`)
+      --repo                                      Work with restic repositories
```
skriss: this is currently a subcommand of `velero restic`, i.e. `velero restic repo get` -- I think using the subcommand structure is more consistent and clearer - WDYT?

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+...`plugin`
+ ```
+      --plugins stringArray                        Plugin container images to install into the Velero Deployment
+      --sa-annotations mapStringString             annotations to add to the Velero ServiceAccount. Add iam.gke.io/gcp-service-account=[GSA_NAME]@[PROJECT_NAME].iam.gserviceaccount.com for workload identity. Optional. Format is key1=value1,key2=value2
+```
+
+##### Velero Server
+`velero server (DEPRECATED)`
```
skriss: we can't deprecate this command, this is the "internal" command that the velero server pod runs :) So `velero init` would be the command that the user runs, that would create a Deployment whose pod runs the `velero server` command, with all the appropriate flags.

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+Commands/flags for backup locations. 
+
+```
+     set 
+        --default string                                  sets the default backup storage location (default "default") (NEW, -- was `server --default-backup-storage-location) 
+
+      create
+        --dry-run                                         generate resources, but don't send them to the cluster. Use with -o. Optional. (NEW)
+        --default                                         Sets this new location to be the new default backup location. Default is false. (NEW) 
+        --secret-file string                              file containing credentials for backup provider. If not specified, set --no-secret must be used for confirmation. Optional. (MOVED FROM install)
```
skriss: I think `--secret-file` and related flags probably needs to move to the `velero plugin add` command, since we currently only support a single secret per provider.

design/cli-install-changes.md
nrb: I would agree with Steve's comments here - the image could just be a positional argument, not a flag.

There's one exception that I can think of in Velero, where `velero restore create` can take a schedule or a backup name - thus `--from-backup` or `--from-schedule` are required.

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+      --terminating-resource-timeout duration      how long to wait on persistent volumes and namespaces to terminate during a restore before timing out (default 10m0s)
+
+```
+
+2) `velero backup-location`
+Commands/flags for backup locations. 
+
+```
+     set 
+        --default string                                  sets the default backup storage location (default "default") (NEW, -- was `server --default-backup-storage-location)
```
nrb: Where will this be stored? Will it set the `server --default-backup-storage-location` flag, or an annotation on the BSL itself?

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+2) `velero backup-location`
+Commands/flags for backup locations. 
+
+```
+     set 
+        --default string                                  sets the default backup storage location (default "default") (NEW, -- was `server --default-backup-storage-location) 
+
+      create
+        --dry-run                                         generate resources, but don't send them to the cluster. Use with -o. Optional. (NEW)
+        --default                                         Sets this new location to be the new default backup location. Default is false. (NEW)
```
nrb: Similar question as with the `set --default` behavior - where is the setting being stored?

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+
+- Introduce new CLI features 
+- Propose changes to the CLI that go beyond the functionality of install and configure
+- Optimize for shorter length or number of commands/flags
+
+## Background
+
+This document proposes users could benefit from a more intuitive and self-documenting CLI setup as compared to our existing CLI UX. Ultimately, it is proposed that a recipe-style CLI flow for installation, configuration and use would greatly contribute to this purpose.
+
+Also, the `install` command currently can be reused to update Velero configurations, a behavior more appropriate for a command named `config`.
```
nrb: Do you want to remove the reference to `config` here given the comment that `config` is being dropped?

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+Commands/flags for snapshot locations. 
+
+```
+     set 
+        --default mapStringString                         sets the list of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...) (NEW, -- was `server --efault-volume-snapshot-locations)  
+
+      create
+        --dry-run                                         generate resources, but don't send them to the cluster. Use with -o. Optional. (NEW)
+        --default                                         Sets these new locations to be the new default snapshot locations. Default is false. (NEW) 
+        --secret-file string                              file containing credentials for volume provider. If not specified, set --no-secret must be used for confirmation. Optional. (MOVED FROM install)
```
nrb: Echoing Steve's comment, I think the `--secret-file` flag probably goes better with `velero plugin add` since it's related to the plugin. We can't really do secrets per location right now.

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+      --timeout duration                          how long backups/restores of pod volumes should be allowed to run before timing out (default 1h0m0s)
+```
+
+#### Example
+
+Considering this proposal, let's consider what a high-level documentation for getting Velero ready to do backups could look like for Velero users:
+
+After installing the Velero CLI:
+```
+velero init --server... (required setup)
```
nrb: Can you add an example for installing & configuring restic with the new commands, too?

design/cli-install-changes.md
nrb: As Steve said, this is the piece of Velero that runs the control loops that process CRDs and actually does the backups/restores/scheduling.

design/cli-install-changes.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+##### Velero Server
+`velero server (DEPRECATED)` 
+ 
+`velero server --default-backup-storage-location (DEPRECATED)` moved to `velero backup-location set --default` 
+
+`velero server --default-volume-snapshot-locations (DEPRECATED)` moved to `velero snapshot-location set --default`
+
+`velero server --default-restic-prune-frequency (DEPRECATED)` moved to `velero restic set --default-prune-frequency`
```
nrb: Where would this be stored if moved? As a command line flag on `velero server`, this is stored within the command line section of the Deployment YAML right now.

carlisia:
> Also - have you started to think through the backend changes yet, particularly around if/how velero changes to allow it to start up without any BSLs/plugins configured yet? Definitely need to figure that out here as well.

Yes. I wanted to get the final output of the gitops deployment figured out, since in the process I'll be figuring out more details that (potentially) will need changing/refining, then I'll add the content to the doc.

design/cli-install-changes.md
carlisia: Yes, good point. `--server` can be scratched.

design/cli-install-changes.md
carlisia: Yes! If I ever deviate from the standard it's pure oversight, glad to have it pointed out.

In this case, I moved the `--images` flag from the `velero install` command to here and didn't give it much more thinking. Now that I'm thinking about it: should we have the ability to add more than one plugin simultaneously? If not, I'll scratch this and follow the original formatting!

design/cli-install-changes.md
carlisia: Ah yes, I should've made my intention more clear here. I meant to rename this to `velero init`,  so deprecation in name only, since we'd keep the same behavior (as `velero init`). Just to double check:

>So velero init would be the command that the user runs

Not clear if this means it should be `velero init server` or if we can s/server/init.

design/cli-install-changes.md
carlisia: Argh, I spaced out on this. Yep. Glad you noticed it.

design/cli-install-changes.md
carlisia: Accurate!

design/cli-install-changes.md
carlisia: Yes, good thinking ahead. Storing it as part of the BSL spec would be nice as an idea, since it sounds like it's an attribute of it. It would also be nice to, when we do a `get`, have a column that displays which one is the default. What I'd like to avoid is when querying for that needing to load all BSLs and iterate thru them. If we can add a label and annotation to the BSL I think we could do that?

design/cli-install-changes.md
carlisia: Thanks, will do.

design/cli-install-changes.md
skriss: As discussed on the call, it would be nice to support multiple plugins simultaneously, and we could do this similarly to other commands, e.g. `velero plugin add my/image-1 my-image-2 ...`

design/cli-install-changes.md
skriss: I think just `velero init` would be fine, though I suppose you could also go with `velero init server` & `velero init restic`

carlisia:
I just added `yaml` files to demonstrate how to create and patch a Velero deployment using kustomize. This might help as we continue reviewing the PR. I might delete them later.

design/cli-install-changes.md
carlisia: It would all be stored in the same section of the same resources. If a user decides to configure Velero to use restic, it would have to be after `velero init` and the restic bits would be a patch to the Velero deployment. 

If there's a risk I'm missing something please let me know Nolan!

design/cli-install-changes.md
skriss: Just circling back on this, I think we still need the two separate commands:

`velero config server` - a user runs this, and it generates/applies the YAML
`velero server` - this is the command run by the Velero pod, that starts the velero server (see https://github.com/vmware-tanzu/velero/blob/master/pkg/install/deployment.go#L145-L150)

design/cli-install-changes.md
skriss: I think the main takeaway is that this flag needs to exist both on the user command (e.g. `velero init`) *and* on the server command.  When the user specifies it via `velero init --default-restic-prune-frequency <value>`, the result of that is that Deployment YAML gets created whose command is `velero server --default-restic-prune-frequency <value>`.

carlisia: Got it. Thanks for the additional clarification, this helps.

design/cli-install-changes.md
carlisia: Yes, this clarification is useful. I was not considering the `velero server` cmd in the deployment itself.

carlisia:
I'd appreciate a fresh round of reviews, I'm looking forward to starting a list of actual implementation issues and prioritizing that.

In the name of not letting perfection be the enemy of good, I'm rolling with that TBD section. I don't have answers for those two items yet.

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+     volumes:
+      - name: cloud-credentials-azure
+        secret:
+          secretName: cloud-credentials-azure
+```
+
+
+#### Renaming "provider" to "plugin"
+
+As part of this change, we should change to use the term `plugin` instead of `provider`. The reasoning: in practice, we usually have 1 plugin per provider, and if there is an implementation for both object store and volume snapshotter for that provider, it will all be contained in the same plugin. When we handle plugins, we follow this logic. In other words, there's a plugin name (ex: `velero.io/aws`) and it can contain implementations of kind `ObjectStore` and/or `VolumeSnapshotter`.
```
nrb: >The reasoning: in practice, we usually have 1 plugin per provider

It depends :)

Some platforms, such as Amazon, Google, and Azure provide object storage and volumes. So they have 2 Velero plugins - one for object storage and one for volume snapshotting ([See the AWS repo as an example](https://github.com/vmware-tanzu/velero-plugin-for-aws/blob/master/velero-plugin-for-aws/main.go#L28-L29), where the 2 plugins are registered inside of 1 executable). vSphere has only a VolumeSnapshotter, though.

These are often shipped together in 1 binary and 1 container image for convenience, but we could also offer separate binaries/images for them.

Additionally, Velero has plugins that aren't related to storage or locations at all - the BackupItemAction and RestoreItemAction plugins are related to the type of Kubernetes object and the action Velero is performing at that time. Thus, using "plugin" could mean one of these types that aren't tied to a location.

Would it perhaps be less ambiguous to say "location plugins"?

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+But when we handle BSL ir VSL (and the CLI commands/flags that configure them), we use the term `provider`, which can cause ambiguity as if that is a kind of thing different from a plugin. If the plugin is the "thing" that contains the implementation for the desired provider, we should make it easier for the user to guess that and change BackupStorageLocation/VolumeSnapshotLocation `Spec.Provider` field to be called `Spec.Plugin` and all related CLI command flags to `plugin`, and update the docs accordingly.
```
nrb: Is `ir` meant to be `or`?

RE: changing the `Spec.Provider` -> `Spec.Plugin` field, I think that makes sense,  and will require a CRD version bump & deprecation cycle, but along with my previous comment, we should make the wording match across the board.

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+Since each BSL/VSL will have its own association with a secret, the user will no longer need to upload a new secret whenever changing to, or adding, a BSL/VSL for a provider that is different from the one in use. This will be done at setup time. This will make it easier to support any number of BSL/VSL combinations, with different providers each.
+
+The user will start up the Velero server on a cluster by using the command `velero config server`. This will create the Velero deployment resource with default values or values overwritten with flags, create the Velero CRDs, and anything else that is not specific to plugins or BSL/VSL.
+
+The Velero server will start up, verify that the deployment is running, that all CRDs were found, and log a message that it is waiting for a BSL to be configured. at this point, other operations, such as configuring restic, will be allowed. Velero should keep track of its status, ie, if it is ready to create backups or not. This could be a field `ServerStatus` added to `ServerStatusRequest`. Possible values could be [ready|waiting]. "ready" would mean there is at least 1 valid BSL, and "waiting" would be anything but that.
```
nrb: We should probably try to use Kubernetes standards here and set up a liveness/readiness probe that signals Velero's readiness.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes

If these fail, then the pod won't even be marked as `Ready` so tools like `kubectl` will report status, in addition to our own.

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+When adding/configuring a BSL or VSL, we can approach it in one of two ways:
+1) at the time of creating a location, verify if there is a valid, corresponding plugin. If there isn't, don't allow creation.
+2) allow creating locations, and continuously verify if there is a corresponding, valid plugin. When a valid match is found, mark the BSL/VSL as "ready". This would require adding a field to the BSL/VSL to keep track of its status, possibly: [ready|waiting].
```
nrb: I like option 2.

We do have a [Phase](https://github.com/vmware-tanzu/velero/blob/master/pkg/apis/velero/v1/backup_storage_location.go#L125) field that might be used for this.

Upstream, there's also the `Status.Conditions` slice that's used to indicate transitions of this nature.

design/cli-install-changes.md
nrb: https://github.com/vmware-tanzu/velero/pull/2382 is an example of an endpoint that could be consumed for liveness.

Readiness in kubernetes is used to route network traffic; the pod will be left alone if the readiness check fails, but no network traffic will be sent to it.

Liveness checks will restart the pod if they fail. This means we can lean on Kubernetes's abilities to manage the restarting/retrying for us, so long as we have an endpoint that accurately represents what we want.

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+#### Handling CA certs
+
+In anticipating of a new configuration implementation to handle custom CA certs (as per design doc https://github.com/vmware-tanzu/velero/blob/master/design/custom-ca-support.md), a new flag `velero plugin set --cacert-file mapStringString` is proposed. It sets the configuration to use for creating a secret containing a custom certificate for an S3 location of a plugin provider. Format is provider:path-to-file.
+
+A `velero backup-location (create|set) --cacert mapStringString` flag is also being proposed. It sets the name of the corresponding CA cert secret for the object storage. Format is provider:cacert-secret-name.
+
+See discussion https://github.com/vmware-tanzu/velero/pull/2259#discussion_r384700723 for more clarification.
+
+#### Credentials and secrets
```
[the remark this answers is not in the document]
carlisia: Yes, I was very on the fence with this because it does break the non-goal of introducing new features. Thanks for calling it out!

Can I move this to another PR and we could then discuss it there how/when to introduce this change? @skriss / @ashish-amarnath / @nrb

design/cli-install-changes.md
carlisia: @nrb I like disambiguating things even further by qualifying it with `location-`. Should it be `location-plugin` or `location-plugins`? I think it should be singular.
WDYT @skriss / @ashish-amarnath?

design/cli-install-changes.md
carlisia: Yes, `or`. I'll fix it.

design/cli-install-changes.md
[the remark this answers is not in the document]
nrb: Yeah, I think moving it to another PR would work!

design/cli-install-changes.md
nrb: Yeah, I think singular makes sense.

design/cli-install-changes.md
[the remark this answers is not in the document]
skriss: that sounds good to me!

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+      restic                                        Configuration for restic operations.
+        --default-prune-frequency duration          how often 'restic prune' is run for restic repositories by default. Optional.
+        --pod-annotations mapStringString           annotations to add to the Velero and restic pods. Optional. Format is key1=value1,key2=value2
+        --pod-cpu-limit string                      CPU limit for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+        --pod-cpu-request string                    CPU request for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+        --pod-mem-limit string                      memory limit for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+        --pod-mem-request string                    memory request for restic pod. A value of "0" is treated as unbounded. Optional. (default "0")
+        --deployment                                create restic deployment. Default is false. Optional. Other flags will only work if set to true. (NEW, was `velero install use-restic`)
```
skriss, line 88: I would think that if the user doesn't want restic installed, they just won't run `velero config restic`. So I think this `--deployment` flag can probably be dropped?

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+        --timeout duration                      maximum time to wait for plugin information to be reported (default 5s)
+
+      remove                                    Remove a plugin [NAME | IMAGE] 
+
+      set
+        --credentials-file mapStringString      configuration to use for creating a secret containing the AIM credentials for a plugin provider. Format is provider:path-to-file. (was `secret-file`)
+        --no-secret                             flag indicating if a secret should be created. Must be used as confirmation if create --secret-file is not provided. Optional. (MOVED FROM install -- not sure we need it?)
+        --sa-annotations mapStringString        annotations to add to the Velero ServiceAccount for GKE. Add iam.gke.io/gcp-service-account=[GSA_NAME]@[PROJECT_NAME].iam.gserviceaccount.com for workload identity. Optional. Format is key1=value1,key2=value2
+        --cacert-file mapStringString           configuration to use for creating a secret containing a custom certificate for an S3 location of a plugin provider. Format is provider:path-to-file. (NEW)
```
skriss, line 165: I think we just want the `--cacert-file` flag on `velero storage-location create/set`, since this applies on a per-BSL basis.

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+Considering this proposal, let's consider what a high-level documentation for getting Velero ready to do backups could look like for Velero users:
+
+After installing the Velero CLI:
+```
+velero config server [flags] (required)
+velero plugin add IMAGES [flags] (add/config provider plugins)
+velero backup-location/snapshot-location create NAME [flags] (run `velero plugin --get` to see what kind of plugins are available; create locations)
+velero backup/restore/schedule create/get/delete NAME [flags]
+velero config restic [flags]
```
skriss, line 178: `velero config restic` probably comes earlier in the workflow, either right after `velero config server`, or just before creating any backups/etc.

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+`velero server --default-volume-snapshot-locations (DEPRECATED)` changed to `velero snapshot-location set --default`
+
+`velero server --default-restic-prune-frequency (DEPRECATED)` changed to `velero config restic set --default-prune-frequency` 
+
+`velero server --restic-timeout  (DEPRECATED)` changed to `velero config restic set timeout`
+
+`velero server --use-restic  (DEPRECATED)` see `velero config restic`
+
+All other `velero server` flags changed to under `velero config server`.
```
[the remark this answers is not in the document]
skriss: Agreed, I think we probably by-and-large don't want **any** changes to the `velero server` command, since the server still needs to run the same way, with the only possible exceptions being `--default-backup-storage-location` and --default-volume-snapshot-locations` if we're moving to storing those as annotations.

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+## Detailed Design
+
+#### Handling CA certs
+
+In anticipating of a new configuration implementation to handle custom CA certs (as per design doc https://github.com/vmware-tanzu/velero/blob/master/design/custom-ca-support.md), a new flag `velero plugin set --cacert-file mapStringString` is proposed. It sets the configuration to use for creating a secret containing a custom certificate for an S3 location of a plugin provider. Format is provider:path-to-file.
```
skriss, line 267: per previous comment, I think we want this **only** at the backup-location level, not at the plugin level.

Also, FYI, the CA cert changes have now been merged.

design/cli-install-changes.md
skriss: I think I prefer option 2 as well.

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+https://github.com/weaveworks/flagger/tree/master/kustomize
+
+https://github.com/jpeach/contour/tree/1c575c772e9fd747fba72ae41ab99bdae7a01864/kustomize (RFC)
+
+## TBD
+
+Question: how should velero install be aware of env vars that different provider plugins require? (right now we just hardcode the AWS/Azure/GCP ones).
+
+There's currently a special case for Azure + restic, where a set of restic-specific env vars need to be set with storage account name and key, which is likely going to remain hardcoded in upstream Velero - how do we deal with situations like this?
```
skriss, line 422: I filed https://github.com/vmware-tanzu/velero/issues/2386 for this issue - I think we probably want ObjectStore plugins to be able to provide this info via a new plugin function. I think we can tackle it separately from this design.

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+https://github.com/istio/installer/tree/master/kustomize
+
+https://github.com/weaveworks/flagger/tree/master/kustomize
+
+https://github.com/jpeach/contour/tree/1c575c772e9fd747fba72ae41ab99bdae7a01864/kustomize (RFC)
+
+## TBD
+
+Question: how should velero install be aware of env vars that different provider plugins require? (right now we just hardcode the AWS/Azure/GCP ones).
```
skriss, line 420: This could probably be tackled separately, though I would like to figure something more elegant out. We can file a separate issue for it.

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+
+Another new behavior that must be added: the server needs to identify when there no longer exists a valid BSL. At this point, it should change its status from "ready" to one that indicates it is not ready, maybe "waiting". With the first approach above, this would mean checking if there is still at least one BSL. With the second approach, it would require checking the status of all BSLs to find at least one with the status of "ready".
+
+As it is today, a valid VSL would not be required to create backups, unless the backup included a PV.
+
+To make it easier for the user to identify if their Velero server is ready to create backups or not, a `velero status` command should be added. This issue has been created some time ago for this purpose: https://github.com/vmware-tanzu/velero/issues/1094.
+
+## Alternatives Considered
+
+It seems that the vast majority of tools document their usage with `kubectl` and `yaml` files to install and configure their Kubernetes resources. Many of them also make use of Helm, and to a lesser extent some of them have their own CLI tools.
```
skriss, line 398: Should we consider maintaining static YAML files?

I know we moved away from this in the past because they were a maintenance burden, but with the changes to plugins and BSL/VSL, the content in them is now more factored out, meaning we wouldn't have to maintain so many provider-specific permutations.

design/cli-install-changes.md
ashish-amarnath: Yeah, option 2 makes sense.

design/cli-install-changes.md
ashish-amarnath: I would prefer to have these static files, but generated at release time. To have a per release set of these static yaml files.

design/cli-install-changes.md
carlisia: Yes, that makes sense.

design/cli-install-changes.md
carlisia: After perusing other projects, I'd say so, since it seems very common to use yaml for operations. Also, I thought it'd be useful to keep example yaml files and kustomize files, so we can maintain the yaml files and add a kustomize files to serve as example. Killing two birds with one stone.

design/cli-install-changes.md
carlisia: Ahhh. Will move.

carlisia:
Here is a proposed list of items to implement, in order:

- velero config server
- velero config restic
- velero backup-location
- velero snapshot-location
- velero plugin
- rename "provider" to "location-plugin"
- Go over all flags and document what is optional, what is required, and default values. Capitalize all help messages
- Change the velero server behavior to mark itself as "ready" once it finds a "ready" BSL; mark itself as "waiting" if it no longer can find a "ready" BSL
- Add a controller to validate and mark BSLs/VSLs as "ready"/"waiting"
- Add a `velero status` command to report if the server is ready to create backup or not
- Add columns to BSL/VSL to indicate which one is the default

Does this breakdown and order make sense? Did I miss anything? I would add the specs for each in the issues.

design/cli-install-changes.md
skriss: ref https://github.com/vmware-tanzu/velero/issues/2402

design/cli-install-changes.md
nrb: Yeah, the discussion about YAML vs...well, anything else is on-going in the larger community, but YAML is really what everything ends up boiling down to.

The biggest question that remains for me is how the plugin overlays/patches get generated in a sustainable way. I 100% agree with @ashish-amarnath that we should have these generated via tooling on release, I'm just not sure what tooling would look like for the overlays yet. My current best idea is that we'd have a running KinD cluster with the relevant plugin installed and generate a diff of the YAML, but that seems very heavyweight.

Generating the base file(s) wouldn't really be too far off from what we have in our install dry run now, at least. The biggest difference, I think, is generating them such that they get hosted somewhere convenient like with https://projectcontour.io/quickstart/contour.yaml. Maybe a GitHub action & bot tosses them into directory in `_site/docs/<version>` when the release is cut?

design/cli-install-changes.md
```diff
@@ -0,0 +1,426 @@
...
+`--dry-run                                    generate resources, but don't send them to the cluster. Use with -o. Optional.`
+
+The `--help` and `--output` flags will also be included for all commands, omitted below for brevity.
+
+Below is the proposed set of new commands to setup and configure Velero.
+
+1) `velero config`
+
+```
+      server                                         Configure up the namespace, RBAC, deployment, etc., but does not add any external plugins, BSL/VSL definitions. This would be the minimum set of commands to get the Velero server up and running and ready to accept other configurations.
```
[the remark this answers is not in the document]
skriss: + ServiceAccount, ClusterRoleBinding

design/cli-install-changes.md
skriss: procedural question - do we want to continue the YAML discussion in a separate issue and not block this CLI work on it? Since we're not immediately dropping `velero install`, we continue to have a viable path for users, and I do think we need to do some more discussing on kustomize/static YAML/plugin patches/etc. before we know exactly what our path forward is there.

carlisia: Yes, that'd be great imo.

nrb: > do we want to continue the YAML discussion in a separate issue and not block this CLI work on it? 

Yeah, let's do that.

nrb:
@carlisia I think the order might need to change a little bit, since you have the Velero server marking itself ready once it finds a "ready" BSL, but the controller for marking BSLs "ready" after that.

IMO, I think these can be done in parallel with the rest, if I'm understanding everything correctly.

 * Add a controller to validate and mark BSLs/VSLs as "ready"/"waiting"
* Change the velero server behavior to mark itself as "ready" once it finds a "ready" BSL; mark itself as "waiting" if it no longer can find a "ready" BSL
* Add a velero status command to report if the server is ready to create backup or not

I also think the Velero server readiness can be looked at in terms of a liveness check like in https://github.com/vmware-tanzu/velero/pull/2382, but implementation can be deferred til later.

design/cli-install-changes.md
carlisia: https://github.com/vmware-tanzu/velero/issues/2405

carlisia:
> @carlisia I think the order might need to change a little bit, since you have the Velero server marking itself ready once it finds a "ready" BSL, but the controller for marking BSLs "ready" after that.

Yes, that makes sense.

carlisia:
Perfect! @ashish-amarnath PTA: when you have a chance.

Issue #499, opened 2018-05-17
Remove third_party/kubenetes/pkg/kubectl/cmd/util/shortcut_expander.go


ncdc:
When we upgrade to the version of client-go that corresponds to k8s v1.11, we can remove our forked copy of shortcut_expander.go now that https://github.com/kubernetes/kubernetes/pull/63507 merged.

skriss:
@ncdc I believe we can do this now?  Should just be a matter of changing https://github.com/heptio/ark/blob/master/pkg/discovery/helper.go#L104 to use https://github.com/kubernetes/client-go/blob/master/restmapper/shortcut.go#L40, correct?

ncdc:
@skriss yes, I think so, but I want to lock down #596 first

skriss:
Note that a naive swap here will result in a performance regression, most easily observable as a significant (30-60s) pause at the start of a restore operation. See https://github.com/vmware-tanzu/velero/blob/master/third_party/kubernetes/pkg/kubectl/cmd/util/shortcut_expander.go#L40-L42 for details on this.

PR #2206, opened 2020-01-17
remove hardcoded svc, netpol mappings

### LINKED ISSUE #276: remove hardcoded svc and netpol shortcut mappings
https://github.com/heptio/ark/blob/master/third_party/kubernetes/pkg/kubectl/cmd/util/shortcut_expander.go#L87

This should be able to be removed now since the upstream PR to add these has been merged:
https://github.com/kubernetes/kubernetes/pull/49128

skriss:

closes #276

Old tech-debt issue that can finally be cleaned up. I also added a comment about why we (unfortunately) still need the `third_party` copy of the shortcut expander (xref #499).

Basic testing looks good (confirmed the relevant shortcuts work, plus no performance regression) but would like to do a little more before merging.


skriss:
took out of draft, testing looks fine to me - ready for review!

Issue #2215, opened 2020-01-23
While restore, use PV name from volumesnapshotter in case of PV renaming


mynktl:
**Describe the problem/challenge you have**
Current limitation:
Restoring a backup, having a snapshot, in different namespace rename the PV with `velero-clone-..` (like velero-clone-4b1768f1-f760-4d87-adce-6f93ef0308ee). This makes difficult to map the renamed PV with storage assets(deployment,service), created by volumesnapshotter during restore.

**Describe the solution you'd like**
If velero uses PV name returned by volumesnapshotter(/pvRestorer) then it will be easy for a user to map the PV with storage assets.

**Environment:**

- Velero version (use `velero version`): master
- Kubernetes version (use `kubectl version`): 1.13
- Cloud provider or hardware configuration: minio, openebs/velero-plugin

mynktl:
@skriss @nrb Any solution/update on this? Thanks!

PR #2216, opened 2020-01-23
Using PV name returned from volumesnapshotter while creating a PV

mynktl:
This PR fixes https://github.com/vmware-tanzu/velero/issues/2215


pkg/restore/restore.go
```diff
@@ -865,11 +866,18 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
 
 			if shouldRenamePV {
 				// give obj a new name, and record the mapping between the old and new names
```
carlisia: Because this comment doesn't make sense here anymore, best to remove it.

pkg/restore/restore.go
```diff
@@ -865,11 +866,18 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
...
-				oldName := obj.GetName()
-				newName, err := ctx.pvRenamer(oldName)
-				if err != nil {
-					addToResult(&errs, namespace, errors.Wrapf(err, "error renaming PV"))
-					return warnings, errs
+				var newName string
```
carlisia: The `newName` var makes sense in the context of ln 873, but not on ln 880, since the same name is being used. I propose this var be named just `name`, to avoid being misleading in the case where the name will not be a new name.

mynktl:
Hi @carlisia!
I have addressed your comments. PTAL
Thanks!

skriss:
@mynktl am I reading correctly that the assumption here is that your VolumeSnapshotter plugin would be changing the PV's name? Are you doing that in the `SetVolumeID` function?

pkg/restore/restore.go
```diff
@@ -864,16 +865,20 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
...
-					return warnings, errs
+				if oldName == obj.GetName() {
+					// pvRestorer hasn't modified the PV name, we need to rename the PV
+					name, err = ctx.pvRenamer(oldName)
+					if err != nil {
+						addToResult(&errs, namespace, errors.Wrapf(err, "error renaming PV"))
+						return warnings, errs
+					}
+				} else {
+					// pvRestorer has renamed the PV, let's use the updated name
```
skriss: please extend this comment to explain that a `VolumeSnapshotter` plugin could have modified the PV name via the `SetVolumeID` function (assuming this is accurate - this is how I'm reading it).

pkg/restore/restore.go
```diff
@@ -864,16 +865,20 @@ func (ctx *context) restoreItem(obj *unstructured.Unstructured, groupResource sc
...
 			if shouldRenamePV {
-				// give obj a new name, and record the mapping between the old and new names
-				oldName := obj.GetName()
-				newName, err := ctx.pvRenamer(oldName)
-				if err != nil {
-					addToResult(&errs, namespace, errors.Wrapf(err, "error renaming PV"))
-					return warnings, errs
+				if oldName == obj.GetName() {
+					// pvRestorer hasn't modified the PV name, we need to rename the PV
+					name, err = ctx.pvRenamer(oldName)
```
skriss: I haven't fully traced it, but I see that we're now updating the `name` variable which is already declared above this scope -  is there any potential for that to cause an issue by updating in-place? I'll try to trace it through some more and see.

pkg/restore/restore.go
mynktl: Done.

pkg/restore/restore.go
mynktl: Yes, We are using `name` during logging and it may create some confusion the user about source PV. I have renamed it to `pvName`.

mynktl:
Hi @skriss ,
We are doing PV renaming in `SetVolumeID` function and we are doing this only if PV already exists in cluster -- namespace remapping.

I have added test for this scenario : **Restore is with namespace remapping and volumesnapshotter rename the PV**.

mynktl:
Hi @skriss, @carlisia 

Any update on this?
Thank you!

carlisia:
@mynktl there is a conflict now, please rebase when you have a chance.

@skriss PTAL when you have a chance!

skriss:
@mynktl this looks good to me.

Question on timing - we just shipped `v1.3.0-beta.1`, and if there aren't any major bugs found there we'll probably go straight to releasing `v1.3.0`. If that's the case, there's minimal time/cycles to test this change out, though we could always ship another beta and/or do some focused testing on `master` with this change.. How important is it to you to get this change into v1.3 vs. waiting another release cycle and shipping it in v1.4?

mynktl:
Hi @skriss 
We can wait for 1.4, with this change.
One question about merging -- This change will be merged after 1.3 only or we are tracking 1.3 in separate branch?

skriss:
we would hold off on merging until 1.3 is released, so that we can release 1.3 directly from master. barring any major issues, i'd expect to ship 1.3 next week.

skriss:
@carlisia @ashish-amarnath - do you see any issue with merging this into `master` now and including it in v1.3? I discussed with @mynktl on slack and it would be beneficial for them. I think the change is pretty low-risk given that it's only changing behavior if the plugin changes the PV name, which is not something that the AWS/GCP/Azure plugins do, so I'm OK merging it. We can do some focused testing on the `master` image as well, before releasing v1.3.

carlisia:
@skriss I don't see any major risk. :shipit:

skriss:
@mynktl could you add a changelog before we approve/merge? thanks.

mynktl:
Hi @skriss 
Thanks for considering this for 1.3. I've added the change-log, PTAL.

Issue #1168, opened 2019-01-24
In Cluster B allow `--from-schedule` to reference a schedule created in Cluster A


mehemken:
**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

This is not urgent, nor is it super critical. It is just nice to be able to say

```
ark create from --from-schedule foo-bar
```
rather than
```
ark get backups -l ark-schedule=foo-bar | head
ark create --from-backup [COPY_PASTE]
```

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

In the slack channel @nrb said this might be by design. If that is the case this issue can be closed. I don't think it is mission critical by any stretch of the imagination.

**Environment:**

- Ark version (use `ark version`): 0.10.x

ncdc:
Adding more context from slack:
```
$  ark create restore --from-schedule jenkins-main
An error occurred: schedules.ark.heptio.com "jenkins-main" not found
```

The code in `pkg/cmd/cli/restore/create.go` tries to `get` the named schedule and fails if it's not found. I think we can probably remove this check and allow the `RestoreController` to mark the restore as failed if it can't find any backups matching the requested schedule.

ncdc:
I just checked and the controller already does the check, so we should be able to remove the couple of lines of code from `create.go`

mehemken:
I haven't forgotten about this... I plan to get it done this weekend. 😀

cpanato:
@skriss @carlisia this still an issue? if yes can i look into?

skriss:
I believe so, and yes, that'd be great! The approach Andy described sounds reasonable, but let us know if there are any blockers.

skriss:
@cpanato it looks like right now, the `velero restore create` command validates that, if you provide a schedule via `--from-schedule`, the schedule currently exists in the current cluster (see https://github.com/vmware-tanzu/velero/blob/master/pkg/cmd/cli/restore/create.go#L173-L177).

Per Andy's comment, I think it would be safe to remove this validation, and just operate based on label selectors to try to find the most recent backup for the schedule (i.e. https://github.com/vmware-tanzu/velero/blob/master/pkg/cmd/cli/restore/create.go#L228 or https://github.com/vmware-tanzu/velero/blob/master/pkg/controller/restore_controller.go#L319-L341). 

This has two benefits: backups created by schedules in different clusters can be restored this way, and also backups created by schedules that no longer exist can be restored.

[...]

PR #2218, opened 2020-01-24
remove schedule validation

cpanato:
the backup can be created from a schedule in another cluster and when trying to restore that in a new cluster this schedule might not exist, so removing this validation

More context see:
 - fixes: https://github.com/vmware-tanzu/velero/issues/1168

### DIFF
--- a/pkg/cmd/cli/restore/create.go
+++ b/pkg/cmd/cli/restore/create.go
@@ -171,11 +171,11 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 			return err
 		}
 	case o.ScheduleName != "":
-		scheduleItems, err := o.client.VeleroV1().Backups(f.Namespace()).List(metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", api.ScheduleNameLabel, o.ScheduleName)})
+		backupItems, err := o.client.VeleroV1().Backups(f.Namespace()).List(metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", api.ScheduleNameLabel, o.ScheduleName)})
 		if err != nil {
 			return err
 		}
-		if len(scheduleItems.Items) == 0 {
+		if len(backupItems.Items) == 0 {
 			return errors.Errorf("No backups found for the schedule %s", o.ScheduleName)
 		}
 	}
--- a/pkg/cmd/cli/restore/create.go
+++ b/pkg/cmd/cli/restore/create.go
@@ -165,13 +165,8 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 		return errors.New("Velero client is not set; unable to proceed")
 	}
 
-	switch {
-	case o.BackupName != "":
+	if o.BackupName != "" {
 		if _, err := o.client.VeleroV1().Backups(f.Namespace()).Get(o.BackupName, metav1.GetOptions{}); err != nil {
-			return err
-		}
-	case o.ScheduleName != "":
-		if _, err := o.client.VeleroV1().Schedules(f.Namespace()).Get(o.ScheduleName, metav1.GetOptions{}); err != nil {
 			return err
 		}
 	}
--- a/pkg/cmd/cli/restore/create.go
+++ b/pkg/cmd/cli/restore/create.go
@@ -165,9 +165,18 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 		return errors.New("Velero client is not set; unable to proceed")
 	}
 
-	if o.BackupName != "" {
+	switch {
+	case o.BackupName != "":
 		if _, err := o.client.VeleroV1().Backups(f.Namespace()).Get(o.BackupName, metav1.GetOptions{}); err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


skriss:
thanks @cpanato! I tried this out and it works fine as-is: the `--from-schedule` flag works as long as there is a backup labeled with the schedule name, regardless of whether or not that particular schedule currently exists in the cluster.  In the case where no backups exist with the given schedule name as a label, the restore terminates with a phase of `FailedValidation`.

One small improvement we could make is to replaced the deleted Schedule existence check with a check to see if there are any backups labeled with the schedule name (i.e. do a `.List(...)` on backups with a label selector on `velero.io/schedule-name=<schedule name>`). If there are none, we could return an error. This provides more immediate feedback to the CLI user in that scenario. I don't think it's essential, but would be nice and should only require a few lines of code if you're up for it.

pkg/cmd/cli/restore/create.go
```diff
@@ -165,15 +165,10 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 		return errors.New("Velero client is not set; unable to proceed")
 	}
 
-	switch {
-	case o.BackupName != "":
+	if o.BackupName != "" {
```
ashish-amarnath: Referring to the checks on L#151 and L#155, should we validate `ScheduleName` at all?

cpanato:
@skriss @ashish-amarnath made the change, please let me know what do you think. thanks again for the review

pkg/cmd/cli/restore/create.go
```diff
@@ -171,9 +171,15 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 			return err
 		}
 	case o.ScheduleName != "":
-		if _, err := o.client.VeleroV1().Schedules(f.Namespace()).Get(o.ScheduleName, metav1.GetOptions{}); err != nil {
+		selector := fmt.Sprintf("velero.io/schedule-name=%s", o.ScheduleName)
```
skriss: you can use the `api.ScheduleNameLabel` constant here in place of the hardcoded label key.

pkg/cmd/cli/restore/create.go
```diff
@@ -171,9 +171,15 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
...
+		scheduleItems, err := o.client.VeleroV1().Backups(f.Namespace()).List(metav1.ListOptions{LabelSelector: selector})
+		if err != nil {
 			return err
 		}
+		if len(scheduleItems.Items) == 0 {
+			fmt.Println("No backups found")
```
skriss: I'd drop the `fmt.Println(...)` since the returned error will be bubbled up and output to stderr.

skriss:
one other thing - please add a changelog file!

pkg/cmd/cli/restore/create.go
cpanato: my bad :(

pkg/cmd/cli/restore/create.go
cpanato: thanks for letting me know the constant.
thanks again for your review, very useful to learn more about velero

cpanato:
@skriss done :) ptal

skriss:
thanks for the revisions @cpanato!

pkg/cmd/cli/restore/create.go
```diff
@@ -171,9 +171,13 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 			return err
 		}
 	case o.ScheduleName != "":
-		if _, err := o.client.VeleroV1().Schedules(f.Namespace()).Get(o.ScheduleName, metav1.GetOptions{}); err != nil {
+		scheduleItems, err := o.client.VeleroV1().Backups(f.Namespace()).List(metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", api.ScheduleNameLabel, o.ScheduleName)})
```
ashish-amarnath:
```suggestion
		backupItems, err := o.client.VeleroV1().Backups(f.Namespace()).List(metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", api.ScheduleNameLabel, o.ScheduleName)})
```

ashish-amarnath: The variable name kinda threw me off. We are trying to get a list of all backups that have this `scheduleName`.
suggest renaming the variable.

pkg/cmd/cli/restore/create.go
```diff
@@ -171,9 +171,13 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
...
+		if err != nil {
 			return err
 		}
+		if len(scheduleItems.Items) == 0 {
```
ashish-amarnath:
```suggestion
		if len(backupItems.Items) == 0 {
```

pkg/cmd/cli/restore/create.go
cpanato: makes sense, thanks

cpanato:
@ashish-amarnath done, thanks for the inputs!
@carlisia @skriss  PTAL

Issue #2173, opened 2020-01-06
update to use GA toplogy labels for PVs


skriss:
As of Kubernetes 1.17, the labels that capture a PV's region & zone have gone GA:
- https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-release-announcement/
- https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesioregion
- https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesiozone

Velero currently uses the older beta version of the `zone` label in at least one place that I'm aware of: 
- https://github.com/vmware-tanzu/velero/blob/master/pkg/backup/item_backupper.go#L389

We probably want to use the GA version of the label, if it exists, otherwise fall back to the beta version. At some point we'll be able to drop support entirely for the beta version of the label -- we can add a `TODO` to the code for this.

cpanato:
@skriss can i take this to try out?

skriss:
yes, absolutely! Let me know if you need any further input beyond what I put in the initial description.

PR #2219, opened 2020-01-24
use GA topology labels for PVs

cpanato:
add stable topology labels for PV the old label was deprecated in K8s 1.17 - https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesiozone

 - fixes: https://github.com/vmware-tanzu/velero/issues/2173

### DIFF
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -390,7 +390,7 @@ func (ib *defaultItemBackupper) volumeSnapshotter(snapshotLocation *api.VolumeSn
 // on PVs
 const (
 	zoneLabelDeprecated = "failure-domain.beta.kubernetes.io/zone"
-	zoneLabel           = "topology.kubernetes.io/region"
+	zoneLabel           = "topology.kubernetes.io/zone"
 )
 
 // takePVSnapshot triggers a snapshot for the volume/disk underlying a PersistentVolume if the provided
@@ -420,7 +420,8 @@ func (ib *defaultItemBackupper) takePVSnapshot(obj runtime.Unstructured, log log
 		}
 	}
 
-	// TODO: prefer stable labels for region in >= v1.18
+	// TODO: -- once failure-domain.beta.kubernetes.io/zone is no longer
+	// supported in any velero-supported version of Kubernetes, remove fallback checking of it
 	pvFailureDomainZone, labelFound := pv.Labels[zoneLabel]
 	if !labelFound {
 		log.Infof("label %q is not present on PersistentVolume, checking deprecated label...", zoneLabel)
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -385,8 +385,13 @@ func (ib *defaultItemBackupper) volumeSnapshotter(snapshotLocation *api.VolumeSn
 }
 
 // zoneLabel is the label that stores availability-zone info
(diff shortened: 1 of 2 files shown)


pkg/backup/item_backupper.go
```diff
@@ -415,9 +420,14 @@ func (ib *defaultItemBackupper) takePVSnapshot(obj runtime.Unstructured, log log
 		}
 	}
 
-	pvFailureDomainZone := pv.Labels[zoneLabel]
-	if pvFailureDomainZone == "" {
-		log.Infof("label %q is not present on PersistentVolume", zoneLabel)
+	// TODO: prefer stable labels for region in v1.18
+	pvFailureDomainZone, labelFound := pv.Labels[zoneLabel]
+	if !labelFound {
```
carlisia: This should check for the `zoneLabelStable` first, then try to fallback to the deprecated label if it isn't there. It makes the deprecation intention more visible.

carlisia: Now that I think about it: maybe `zoneLabelStable` should be changed to just `zoneLabel`, and the existing `zoneLabel` could be called `zoneLabelDeprecated`. The intention here is to, when the old label is deprecated, there'd be no context for `zoneLabelStable` (stable vs. ?).

cpanato: thanks for your review @carlisia 
for the `zoneLabelStable` i used the same nomenclature as kubernetes codebase uses, but i can change that per your suggestion.

cpanato: done

pkg/backup/item_backupper.go
```diff
@@ -415,9 +420,14 @@ func (ib *defaultItemBackupper) takePVSnapshot(obj runtime.Unstructured, log log
 		}
 	}
 
-	pvFailureDomainZone := pv.Labels[zoneLabel]
-	if pvFailureDomainZone == "" {
-		log.Infof("label %q is not present on PersistentVolume", zoneLabel)
+	// TODO: prefer stable labels for region in >= v1.18
```
skriss: Let's be a little more explicit with this `// TODO` -- something like "once failure-domain.beta.kubernetes.io/zone is no longer supported in any velero-supported version of Kubernetes, remove fallback checking of it"

pkg/backup/item_backupper.go
```diff
@@ -384,9 +384,14 @@ func (ib *defaultItemBackupper) volumeSnapshotter(snapshotLocation *api.VolumeSn
...
 }
 
+// zoneLabelDeprecated is the label that stores availability-zone info
+// on PVs this is deprecated on Kubernetes >= 1.17.0
 // zoneLabel is the label that stores availability-zone info
 // on PVs
-const zoneLabel = "failure-domain.beta.kubernetes.io/zone"
+const (
+	zoneLabelDeprecated = "failure-domain.beta.kubernetes.io/zone"
+	zoneLabel           = "topology.kubernetes.io/region"
```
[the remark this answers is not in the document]
cpanato: Oh, my fault, overlooked that, feel dumb :/
thanks for pointing that out

cpanato:
@skriss thanks so much for taking your time to review.
Did the changes you requested, please let me know if I'm missing something or what more I can do.

skriss:
@carlisia please take another look when you get a chance!

PR #2234, opened 2020-01-31
Restore result refactoring

skriss:
This is a small refactoring PR that moves some helper functions out of `restore.go` and into `result.go`, as methods on the `Result` type.  I'd like to do some more significant refactoring in `restore.go` given its size/complexity, and this is just a small first step.

I still need to test, but pls let me know if you think this is worthwhile.


ashish-amarnath:
This definitely cleans stuff up and is worth doing!

skriss:
thanks for the input, I'll get some testing done.

skriss:
added some unit tests, did manual testing - all looks good. ready for review!

skriss:
@carlisia PTAL when you get a chance!

PR #2240, opened 2020-02-04
Add api-types/restore.md doc

jnaulty:


skriss:
looks good to me too @jnaulty - only request is to add a link to this new page on `site/docs/master/api-types/README.md`.

I wouldn't worry about backporting it to the `v1.2.0` docs, since `v1.3.0` will be out soon and the docs generation process for it will include this new page.

jnaulty:
> looks good to me too @jnaulty - only request is to add a link to this new page on site/docs/master/api-types/README.md.

Good catch, @skriss 
just pushed that change.

PR #2248, opened 2020-02-05
refactor restore priorities code to use single loop and lazy discovery

skriss:

This refactors the restore code to:
- lazily resolve resources via discovery, i.e. not until they're just about to be restored;
- refreshes discovery immediately after restoring CRDs (since new APIs/resources may have been created)
- removes the second restore loop

I've done some manual testing here and things seem to be working as expected across CRDs, CRs, and built-in resources.

Keeping in draft until nolan is back.

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -98,64 +98,6 @@ type kubernetesRestorer struct {
 	logger                     logrus.FieldLogger
 }
 
-// prioritizeResources returns an ordered, fully-resolved list of resources to restore based on
-// the provided discovery helper, resource priorities, and included/excluded resources.
-func prioritizeResources(helper discovery.Helper, priorities []string, includedResources *collections.IncludesExcludes, logger logrus.FieldLogger) ([]schema.GroupResource, error) {
-	var ret []schema.GroupResource
-
-	// set keeps track of resolved GroupResource names
-	set := sets.NewString()
-
-	// start by resolving priorities into GroupResources and adding them to ret
-	for _, r := range priorities {
-		gvr, _, err := helper.ResourceFor(schema.ParseGroupResource(r).WithVersion(""))
-		if err != nil {
-			return nil, err
-		}
-		gr := gvr.GroupResource()
-
-		if !includedResources.ShouldInclude(gr.String()) {
-			logger.WithField("groupResource", gr).Info("Not including resource")
-			continue
-		}
-		ret = append(ret, gr)
-		set.Insert(gr.String())
-	}
-
-	// go through everything we got from discovery and add anything not in "set" to byName
-	var byName []schema.GroupResource
-	for _, resourceGroup := range helper.Resources() {
-		// will be something like storage.k8s.io/v1
-		groupVersion, err := schema.ParseGroupVersion(resourceGroup.GroupVersion)
-		if err != nil {
-			return nil, err
-		}
-
-		for _, resource := range resourceGroup.APIResources {
-			gr := groupVersion.WithResource(resource.Name).GroupResource()
-
-			if !includedResources.ShouldInclude(gr.String()) {
-				logger.WithField("groupResource", gr.String()).Info("Not including resource")
-				continue
-			}
-
-			if !set.Has(gr.String()) {
-				byName = append(byName, gr)
-			}
-		}
-	}
-
-	// sort byName by name
-	sort.Slice(byName, func(i, j int) bool {
-		return byName[i].String() < byName[j].String()
-	})
-
-	// combine prioritized with by-name
-	ret = append(ret, byName...)
-
-	return ret, nil
-}
-
 // NewKubernetesRestorer creates a new kubernetesRestorer.
 func NewKubernetesRestorer(
 	discoveryHelper discovery.Helper,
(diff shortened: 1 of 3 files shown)


pkg/restore/restore.go
```diff
@@ -388,6 +324,21 @@ type resourceClientKey struct {
 	namespace string
 }
 
+// getOrderedResources returns an ordered list of resource identifiers to restore, based on the provided resource
+// priorities and backup contents. The returned list begins with all of the prioritized resources (in order), and
+// appends to that an alphabetized list of all resources in the backup.
+func getOrderedResources(resourcePriorities []string, backupResources map[string]*archive.ResourceItems) []string {
+	// alphabetize resources in the backup
```
nrb: Since we're alphabetizing now, does this mean Velero's restore order is now more determinate than it was before?

pkg/restore/restore.go
```diff
@@ -404,104 +355,66 @@ func (ctx *context) execute() (Result, Result) {
...
 		}
+		groupResource := gvr.GroupResource()
 
-		resourceList := backupResources[resource.String()]
-		if resourceList == nil {
+		// check if we've already restored this resource (this would happen if the restore
+		// we're currently looking at was already restored because it was a prioritized
+		// resource, and now we're looking at it as part of the backup contents).
+		if processedResources.Has(groupResource.String()) {
+			ctx.log.WithField("resource", groupResource.String()).Infof("Skipping restore of resource because it's already been processed")
```
nrb: Should this be info or debug? Based on your testing, how often do we see this line?

pkg/restore/restore_test.go
```diff
@@ -2410,88 +2407,6 @@ func TestRestoreWithRestic(t *testing.T) {
 	}
 }
```
nrb: I'd add a TODO here that we add a new test for the `getOrderedResources` function.

pkg/restore/restore.go
skriss: we were alphabetizing before too (see https://github.com/vmware-tanzu/velero/pull/2248/files#diff-7e2ddda55b692c82f9fa5f89a3d05e54L149-L151) -- the only difference is now we're looking at what's in the backup, rather than what's in discovery

pkg/restore/restore.go
skriss: probably makes sense to bump down to debug, otherwise it'll come up regularly for any prioritized resource

pkg/restore/restore_test.go
skriss: :+1:

skriss:
> one thing I'm not seeing here (and may just be missing since it's not a thorough review) is whether or not we still force namespaces to be restored 'just in time' before restoring the first item into it.

That behavior shouldn't have changed - we definitely skip restoring `Namespace` resources in the main loop (https://github.com/vmware-tanzu/velero/pull/2248/files#diff-7e2ddda55b692c82f9fa5f89a3d05e54R403-R406) and still ensure the target namespace exists before trying to restore a resource into it (https://github.com/vmware-tanzu/velero/pull/2248/files#diff-7e2ddda55b692c82f9fa5f89a3d05e54R424-R446)

pkg/restore/restore_test.go
skriss: I added a test.

skriss:
Marking this as ready for review. I did some more testing and all is looking good to me:
- all resources in the backup are restored, in the correct order
- instances of CRs are restored properly after waiting for the CRD to be ready
- prioritizing a custom resource just after CRDs (using the `velero server --restore-resource-priorities` flag) works properly

pkg/restore/restore.go
```diff
@@ -404,104 +355,66 @@ func (ctx *context) execute() (Result, Result) {
...
+	// that the resource/API itself is being restored via a custom resource definition, meaning it's not available via
+	// discovery prior to beginning the restore.
+	//
+	// Since we keep track of the fully-resolved group-resources that we *have* restored, we won't try to restore a
+	// resource twice even if it's in the ordered list twice.
+	for _, resource := range getOrderedResources(ctx.resourcePriorities, backupResources) {
+		// try to resolve the resource via discovery to a complete group/version/resource
+		gvr, _, err := ctx.discoveryHelper.ResourceFor(schema.ParseGroupResource(resource).WithVersion(""))
+		if err != nil {
+			// TODO should we refresh discovery and try again?
```
nrb: Do we want to resolve this TODO item now, or wait for later? Did it come up often in testing, @skriss?

skriss: Since I have code to refresh discovery right after restoring custom resource definitions (https://github.com/skriss/velero/blob/refactor-restore-priorities/pkg/restore/restore.go#L456-L464), this didn't come up at all for me, and I think it's probably not necessary.  If we had other resource types beyond CRDs that, upon restore, would necessitate a refresh of discovery, then implementing this TODO might be a better approach than the hardcoded "refresh discovery after restoring CRDs" approach that's implemented now.

I'll remove for now and will need a fresh round of approves :)

skriss:
removed TODO and squashed commits.

PR #2252, opened 2020-02-06
backup describer: reword resource list not found msg

### LINKED ISSUE #2125: adjust `velero backup describe` output for nonexistent resource list

Run `velero backup describe` on an in-progress backup, see:

```
...
Resource List:  <backup resource list not found, this could be because this backup was taken prior to Velero 1.1.0>
...
```

I think we should tweak how this is displayed to users -- we could show a simple "<not found>" text, or just not display the line at all, if a resource list isn't found. There are many potential reasons why a resource list might not exist:

- backup hasn't completed yet
- backup was taken prior to v1.1.0
- resource list failed to upload when backup was taken
- etc.

Velero version (use `velero version`): v1.2.0

skriss:

closes #2125 

I noticed this again in testing (this message is shown if, for example, you run `velero backup describe --details` while the backup is still running) so made the change.


carlisia:
Great.

I wonder if it would be helpful to add a comment to the code listing the legit reasons we do know for this failure?

skriss:
:+1: comment added

Issue #2256, opened 2020-02-07
better document using `spec.publicUrl` to access logs/backups from minIO


a4nowgh:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is, and what commands you ran.)
This occurs after when using minIO storage location.  

velero backup create <jobName> --storage-location <minIO_storage>

velero backup describe <jobName> --details

Output:
Request List: request failed: garbled special characters/encoding repeat repeat etc...

The rest of the output looks OK

NOTE: I am able to perform successful restores with this backup

**What did you expect to happen:**
Show Request List

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Resource lists from backups to my other non-minIO S3-Compliant bucket display OK

**Environment:**
- minIO Version: 2019-04-17T17:41:56Z
- minIO Release-tag: RELEASE.2019-04-17T17-41-56Z
- Velero version: v1.2.0
- Velero features:
- Kubernetes Client version: v1.15.3
- Kubernetes Server version: v1.14.6
- Cloud provider or hardware configuration: aws
- OS: Ubuntu 18.04.3 LTS

carlisia:
Hm this is odd. Mind dumping the actual output, it might be helpful to see.

Is this an on-prem cluster you are backing up?

a4nowgh:
Yes, this is an on-prem cluster we are backing up to.  Below is the requested output...thx

./velero backup describe minio-backup --details
Name: minio-backupNamespace: veleroLabels:    velero.io/storage-location=minioAnnotations: <none>
Phase:  Completed
Namespaces:  Included: grafana  Excluded: <none>
Resources:  Included: *  Excluded: <none>  Cluster-scoped: auto
label selector: <none>
Storage Location: minio
Snapshot PVs: auto
TTL: 720h0m0s
Hooks: <none>
Backup Format Version: 1
Started: XXXXXCompleted: XXXXX

Expiration: XXXXX
Resource List: <error getting backup resource list: request failed: garbled garbled garbled....
Persistent Volumes: <none included>
Restic Backups:  Completed:    grafana/grafana-56d485dbb8-8rhqf: config, ldap, storage

    On Friday, February 7, 2020, 7:43:25 PM EST, Carlisia Campos <notifications@github.com> wrote:  
 
 
Hm this is odd. Mind dumping the actual output, it might be helpful to see.

Is this an on-prem cluster you are backing up?

nrb:
[...]

PR #2259, opened 2020-02-10
Add design doc for Custom CA support for S3 BSLs and Velero Installation

dymurray:
Signed-off-by: Dylan Murray <dymurray@redhat.com>

Design for #1027

### DIFF
--- a/design/custom-ca-support.md
+++ b/design/custom-ca-support.md
@@ -1,0 +1,97 @@
+# Custom CA Bundle Support for S3 Object Storage
+
+It is desired that Velero performs SSL verification on the Object Storage
+endpoint (BackupStorageLocation), but it is not guaranteed that the Velero
+container has the endpoints' CA bundle in it's system store. Velero needs to
+support the ability for a user to specify custom CA bundles at installation
+time and Velero needs to support a mechanism in the BackupStorageLocation
+Custom Resource to allow a user to specify a custom CA bundle. This mechanism
+needs to also allow Restic to access and use this custom CA bundle.
+
+## Goals
+
+- Enable Velero to be configured with a custom CA bundle at installation
+- Enable Velero support for custom CA bundles with S3 API BackupStorageLocations
+- Enable Restic to use the custom CA bundles whether it is configured at installation time or on the BackupStorageLocation
+
+## Non Goals
+
+- Support non-S3 providers
+
+## Background
+
+Currently, in order for Velero to perform SSL verification of the object
+storage endpoint the user must manually set the `AWS_CA_BUNDLE` environment
+variable on the Velero deployment. If the user is using Restic, the user has to
+either:
+1. Add the certs to the Restic container's system store
+1. Modify Velero to pass in the certs as a CLI parameter to Restic - requiring
+   a custom Velero deployment
+
+## High-Level Design
+
+On the Velero deployment at install time, we can set the AWS environment variable
+`AWS_CA_BUNDLE` which will allow Velero to communicate over https with the
+proper certs when communicating with the S3 bucket. This means we will add the
+ability to specify a custom CA bundle at installation time. For more
+information, see "Install Command Changes".
+
+On the Restic daemonset, we will want to also mount this secret at a pre-defined
+location. In the `restic` pkg, the command to invoke restic will need to be
+updated to pass the path to the cert file that is mounted if it is specified in
+the config.
+
+This is good, but doesn't allow us to specify different certs when
+`BackupStorageLocation` resources are created.
+
+In order to support custom certs for object storage, Velero will add an
+additional field to the `BackupStorageLocation`'s provider `Config` resource to
+provide a secretRef which will contain the coordinates to a secret containing
+the relevant cert file for object storage. Then, in order for Restic to be able
+to consume and use this cert, Velero will need to modify the Restic daemonset
+to include the secret and mount it at a pre-defined path. This path will then
+be passed in to the restic server at command invocation.
+
+## Detailed Design
+
+The `AWS_CA_BUNDLE` environment variable works for the Velero deployment
+because this environment variable is passed into the AWS SDK which is used in
+the [plugin][1] to build up the config object. This means that a user can
+simply define the CA bundle in the deployment as an env var. This can be
+utilized for the installation of Velero with a custom cert by simply setting
+this env var to the contents of the CA bundle, or the env var can be mapped to
+a secret which is controlled at installation time. I recommend using a secret
+as it makes the Restic integration easier as well.
+
+At installation time, if a user has specified a custom cert then the Restic
+daemonset should be updated to include the secret mounted at a predefined path.
+We could optionally use the system store for all custom certs added at
+installation time. Restic supports using the custom certs [in addition][3] to
+the root certs.
+
+In the case of the BSL being created with a secret reference, then at runtime
+the secret will need to be consumed. This secret will be read and applied to
+the AWS `session` object. The `getSession()` function will need to be updated
+to take in the custom CA bundle so it can be passed [here][4].
+
+The Restic daemonset will need to be updated to include the secret mounted as a
+volume in the containers at a defined path. The restic [command invocation][2]
+will need to be updated to include the path to the file as an argument to the
+restic server using `--cacert`. For the path when a user defines a custom cert
+on the BSL, Velero will be responsible for updating the daemonset to include
+the secret mounted as a volume at a predefined path.
+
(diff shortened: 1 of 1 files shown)


design/custom-ca-support.md
```diff
@@ -0,0 +1,97 @@
...
+the config.
+
+This is good, but doesn't allow us to specify different certs when
+`BackupStorageLocation` resources are created.
+
+In order to support custom certs for object storage, Velero will add an
+additional field to the `BackupStorageLocation`'s provider `Config` resource to
+provide a secretRef which will contain the coordinates to a secret containing
+the relevant cert file for object storage. Then, in order for Restic to be able
+to consume and use this cert, Velero will need to modify the Restic daemonset
```
dymurray: Is this an accepted path for Velero? It feels odd to me to update the daemonset for Restic on the fly when creating BSLs as this could interrupt some activity Restic is undertaking.

However, without modifying the restic daemonset I don't see a feasible way to get the custom CA bundle into Restic at runtime. In our PoC we had to use an initcontainer on Restic which would symlink a secret that is mounted into the Restic container. This would allow us to modify a secret from Velero and have the associated changes populated in the Restic container. I do not think that's a viable approach moving forward though.

design/custom-ca-support.md
```diff
@@ -0,0 +1,107 @@
...
+
+It is desired that Velero performs SSL verification on the Object Storage
+endpoint (BackupStorageLocation), but it is not guaranteed that the Velero
+container has the endpoints' CA bundle in it's system store. Velero needs to
+support the ability for a user to specify custom CA bundles at installation
+time and Velero needs to support a mechanism in the BackupStorageLocation
+Custom Resource to allow a user to specify a custom CA bundle. This mechanism
+needs to also allow Restic to access and use this custom CA bundle.
+
+## Goals
```
nrb: Does this support adding the certs on the client side? I know velero client-side commands such as `velero (backup|restore) (describe|logs)` on AWS/GCP/Azure get a 1-time use URL for retrieving files from object store, and on Minio they use a BSL's `Spec.Config.publicUrl` value in order to get non-encrypted values.

dymurray: It currently does not but I wouldn't rule that out. Are you thinking that the `velero install` command would install the cert on the client's machine? Or that the velero client commands would have a new flag `--cacert` which a user can optionally specify?

nrb: I'm not exactly sure, honestly. I think either would work, so long as it's documented. A `--cacert` flag would probably be easier since certs can be managed in several different ways depending on client OS.

Mostly, I'd like to avoid users hitting issues like this one often. https://github.com/vmware-tanzu/velero/issues/2256

design/custom-ca-support.md
skriss: One idea: take a look at Velero's code for managing restic repo credentials, i.e. the encryption key. The way this works today is that the key is stored in a secret in the `velero` namespace, and each time we exec a `restic` command, we read the contents of the secret, write it out to a temp file, pass the path of the temp file to `restic`, and then remove the temp file afterwards. This has worked pretty well in practice*. 

We might be able to take a similar approach for the CA bundles, where they're stored in secrets, the BSL has a reference to them, and they're projected into a temp file just-in-time for restic invocations.

Here are a few links to relevant sections of the current code:

https://github.com/vmware-tanzu/velero/blob/master/pkg/restic/repository_manager.go#L238-L245
https://github.com/vmware-tanzu/velero/blob/master/pkg/restic/common.go#L168-L203

* as an aside, yes, this is over-complicated given that we're currently using a static/non-secret encryption key for restic repos, but this was implemented before we decided to go with that approach instead of using unique, secret keys per repo.

design/custom-ca-support.md
```diff
@@ -0,0 +1,107 @@
...
+
+- Enable Velero to be configured with a custom CA bundle at installation
+- Enable Velero support for custom CA bundles with S3 API BackupStorageLocations
```
skriss: just to confirm here - you're proposing implementing support for this in two different ways (one at install time, for if you don't need different CA bundles per BSL, and one where you can have a different CA bundle per BSL), and the user could pick whichever approach works best for them, correct?

design/custom-ca-support.md
dymurray: @skriss thanks a lot for the links I definitely think we could follow a similar pattern for the CA bundle here since restic allows for a `--cacert` command flag which we can specify like we do the password file here: https://github.com/vmware-tanzu/velero/blob/master/pkg/restic/command.go#L86.

>as an aside, yes, this is over-complicated given that we're currently using a static/non-secret encryption key for restic repos

Makes sense. I have faced similar problems like this in the past and it seems like this is a common pattern used to solve the issue of needing data stored somewhere in memory of a pod that a controller will manage.

I will try something like this out and update the document.

design/custom-ca-support.md
dymurray: Correct. Mainly because it would allow a user to specify clusterwide certs that they don't want to have to specify every time they create additional BSLs after Velero was installed.

I'm okay to reduce the complexity if you think managing it on the BSL alone is adequate.

skriss: I think supporting both is :+1:, just wanted to confirm I was reading it correctly!

skriss:
@dymurray just checking in to see if you were able to try out the restic approach per https://github.com/vmware-tanzu/velero/pull/2259#discussion_r380363800, and if you need any other input!

dymurray:
@skriss I was on PTO for the better part of last week. Have begun putting together the PoC for this this morning. Hoping to have it confirmed by the community meeting tomorrow and can update the design doc.

dymurray:
Can confirm that the approach of using the methods shown in the `repositoryManager` code will work for injecting the CA bundles into the restic pod. Updating the design doc to reflect this and will poke for review.

dymurray:
@skriss @nrb Please take a look when you get a free chance. I have added information about including a way to configure a cert in the client and to include the updates about using a temp file similar to the restic repostiory credentials.

Thanks!

design/custom-ca-support.md
```diff
@@ -0,0 +1,128 @@
...
+cert file. Then the install command would do the heavy lifting of creating a
+secret and updating the proper fields on the deployment and daemonset to mount
+the secret at a well defined path.
+
+### Velero client changes
+
+Since the Velero client is responsible for gathering logs and information about
+the Object Storage, this implementation should include a new flag `--cacert`
+which can be used when communicating with the Object Storage. Additionally, the
+user should be able to set this in their client configuration.
```
nrb: To put more detail on it, I would probably put it under `velero client config set cacert PATH` if someone would like to permanently set the value.

design/custom-ca-support.md
```diff
@@ -0,0 +1,128 @@
...
+as an argument to the restic server using `--cacert`. For the path when a user
+defines a custom cert on the BSL, Velero will be responsible for updating the
+daemonset to include the secret mounted as a volume at a predefined path.
+
+Where we mount the secret is a fine detail, but I recommend mounting the certs
+to `/certs` to keep it in line with the other volume mount paths being used.
+
+### Install command changes
+
+The installation flags should be updated to include the ability to pass in a
```
nrb: @carlisia This is relevant to #2202. My guess would probably be part of the `plugin add` command there, with a new option for certs that would create the secret and do the updating on the deployment, and restic DaemonSet. I _think_ this is only valid for ObjectStore plugins, at least the restic DaemonSet part. I'm envisioning the command would take a path to the cert file, create the Secret in the cluster, then put a secretRef in the Velero Deployment/restic DaemonSet.

We'd also need to have the `backup-location (set|create)` commands have an argument to take a reference to a secret name (which is the cert) in the same namespace. Possibly the `snapshot-location (set|create)` commands, if it's relevant to them. The command would create or edit the BSL (or VSL) to have a reference to the Secret within the same namespace.

carlisia: Copy that! I'll add it to the doc, thank you.

carlisia: Seems it will work the same way as with the credentials for the provider (as far as adding it to the cluster).

design/custom-ca-support.md
dymurray: Thx will add this as a note

design/custom-ca-support.md
nrb: Yep, I think it'll work the same way, it's just a Secret with different contents.

ron1:
Note that on Linux it appears the restic cacert can be set using golang standard pki environment variables as described here: https://github.com/golang/go/issues/14311.

PR #2270, opened 2020-02-13
Do not backup volumes of pending pods

### LINKED ISSUE #2269: Unscheduled pod with invalid PVC reference results in partial failure
This is a bit of an edge case but we have had this happen in one of our production clusters. If you create a pod with a persistentVolumeClaim reference that does not exist, the pod will be stuck in the status `Pending`. Now when a velero full backup runs it tries to backup all volumes of all pods and naturally fails because the persistentVolumeClaim it tries to backup does not exist.

https://github.com/vmware-tanzu/velero/blob/master/pkg/backup/pod_action.go#L65

The backup should either skip pods which are not scheduled or check if a PVC actually exists before adding it to the list in `pod_action.go`.

* Create a pod with an invalid PVC reference
```bash
cat << EOF | kubectl create -f -
apiVersion: v1
kind: Pod
metadata:
  name: nginx-test
spec:
  containers:
  - name: nginx
    image: nginx
    ports:
    - containerPort: 80
    volumeMounts:
    - name: workdir
      mountPath: /usr/share/nginx/html
  dnsPolicy: Default
  volumes:
  - name: workdir
    persistentVolumeClaim:
      claimName: doesnotexist
EOF
```
* Create a full backuo `velero backup create full`
* `velero backup logs full | grep -B 5 error`
```
time="2020-02-13T08:18:29Z" level=info msg="Backing up item" backup=velero/full group=v1 logSource="pkg/backup/item_backupper.go:169" name=nginx-test namespace=velero resource=pods
time="2020-02-13T08:18:29Z" level=info msg="Executing custom action" backup=velero/full group=v1 logSource="pkg/backup/item_backupper.go:330" name=nginx-test namespace=velero resource=pods
time="2020-02-13T08:18:29Z" level=info msg="Executing podAction" backup=velero/full cmd=/velero logSource="pkg/backup/pod_action.go:51" pluginName=velero
time="2020-02-13T08:18:29Z" level=info msg="Adding pvc doesnotexist to additionalItems" backup=velero/full cmd=/velero logSource="pkg/backup/pod_action.go:67" pluginName=velero
time="2020-02-13T08:18:29Z" level=info msg="Done executing podAction" backup=velero/full cmd=/velero logSource="pkg/backup/pod_action.go:77" pluginName=velero
time="2020-02-13T08:18:29Z" level=info msg="1 errors encountered backup up item" backup=velero/full group=v1 logSource="pkg/backup/resource_backupper.go:284" name=nginx-test namespace= resource=pods
time="2020-02-13T08:18:29Z" level=error msg="Error backing up item" backup=velero/full error="persistentvolumeclaims \"doesnotexist\" not found" error.file="/go/src/github.com/vmware-tanzu/velero/pkg/backup/item_backupper.go:351" error.function="github.com/vmware-tanzu/velero/pkg/backup.(*defaultItemBackupper).executeActions" group=v1 logSource="pkg/backup/resource_backupper.go:288" name=nginx-test namespace= resource=pods
```

**Environment:**

```
Client:
        Version: v1.2.0
        Git commit: 5d008491bbf681658d3e372da1a9d3a21ca4c03c
Server:
        Version: v1.2.0
```
```
features: <NOT SET>
```
```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:16:51Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-12-04T07:23:47Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```
`kind v0.6.1 go1.13.4 linux/amd64`
GCP with GCS
Fedora 31 Workstation

ctrox:
Fixes #2269

My preferred solution to this would be to actually check if a PVC exists and not just completely ignore PVCs of `Pending` pods. But as far as I can see that is not easy to achieve since we don't have a kubernetes client in `PodAction`.

### DIFF
--- a/changelogs/unreleased/2270-ctrox
+++ b/changelogs/unreleased/2270-ctrox
@@ -1,0 +1,1 @@
+Do not backup volumes of pending pods
--- a/pkg/backup/pod_action.go
+++ b/pkg/backup/pod_action.go
@@ -61,6 +61,11 @@ func (a *PodAction) Execute(item runtime.Unstructured, backup *v1.Backup) (runti
 		return item, nil, nil
 	}
 
+	if pod.Status.Phase == corev1api.PodPending {
+		a.log.Info("pod is pending")
+		return item, nil, nil
+	}
+
 	var additionalItems []velero.ResourceIdentifier
 	for _, volume := range pod.Spec.Volumes {
 		if volume.PersistentVolumeClaim != nil && volume.PersistentVolumeClaim.ClaimName != "" {
--- a/pkg/backup/pod_action_test.go
+++ b/pkg/backup/pod_action_test.go
@@ -80,6 +80,29 @@ func TestPodActionExecute(t *testing.T) {
 			`),
 		},
 		{
+			name: "pod is pending",
+			pod: velerotest.UnstructuredOrDie(`
+			{
+				"apiVersion": "v1",
+				"kind": "Pod",
+				"metadata": {
+					"namespace": "foo",
+					"name": "bar"
+				},
+				"status": {
+					"phase": "Pending"
+				},
+				"spec": {
+					"volumes": [
+						{
+							"persistentVolumeClaim": {"claimName": "claim1"}
+						}
+					]
+				}
+			}
+			`),
+		},
+		{
 			name: "full test, mix of volume types",
 			pod: velerotest.UnstructuredOrDie(`
 			{


ashish-amarnath:
@ctrox Thanks for your PR.
On the outset, your approach seems fine. Have you been able test this out? Specifically, that the PVC (if any) should also included in the backup.

ashish-amarnath:
Can you also please add a change log to your PR? 
See https://velero.io/docs/v1.3.0-beta.2/code-standards/ for instructions.

ctrox:
> Have you been able test this out? Specifically, that the PVC (if any) should also included in the backup.

Yeah I have tested this locally, plus I added a test case with a pending pod. Correct me if I'm wrong but as long as a PVC is bound it should be backed up by the `backup_pv_action`.

> Can you also please add a change log to your PR?

done

ashish-amarnath:
This PR LGTM.
Requesting reviews from more folks.

skriss:
The code LGTM if we want to go ahead with this.

I do want to call out that we've previously had a bunch of back-and-forth on whether Velero should ignore cases like this, or whether they should be reported as errors/partial failures like they are today (see https://github.com/vmware-tanzu/velero/issues/1878). In the past, we've decided to continue reporting these as errors, so the user is at least alerted that something is wrong.

What do others think? Should we revisit this decision?

nrb:
@ctrox So if I'm reading this correctly, the Pod will be backed up, but no PVCs for it. In your correct scenario, does the Pod restore as Pending? Or is it able to restore successfully because new PVCs are provisioned?

ctrox:
In my scenario, the pod will be restored as Pending as the PVC reference was never valid anyways. 
However, without knowing too much about the architecture of Velero, would it not make sense to completely separate PVC backups from Pods? If for some reason I have a PVC which is not currently bound to a pod, maybe for example some job that uses a volume and is not currently running, I would still expect that PVC+PV to be backed up.

skriss:
@ctrox PVCs *will* still be backed up independent of pods, assuming they match all the filters defined in the backup spec - so, in your scenario, if you were backing up an entire namespace that included a pending pod and a PVC, the PVC would still be backed up. It would just happen at a different point in time, outside the "scope" of the pod backup. The reason for linking pod + PVC + PV backups is that sometimes, users want to run hooks in their pod just before backing up the pod + any PVC/PV it uses, in order to quiesce their application and get consistent volume snapshots.

skriss:
I suppose one could argue that it makes sense not to try to back up related PVCs/PVs of `Pending` pods, because any pre/post hooks won't be able to be executed in the pod anyway, so there's no need to try to back up the PVC/PV within the scope of the pod, even if they exist. If they match the backup's filters, they'll still get backed up, just at a later point in time. And, if we do that, then you sidestep the issue of erroring when the pod is Pending **because** the PVC doesn't actually exist.

carlisia:
I wonder if we should have the concept of policy. A set of rules that Velero would follow to resolve issues like this one, and leave it up to the user to decide. These would be differ from configuration in a way that would dictate how Velero would behave for either all backups or on a per backup basis. I think I remember there being at least one other behavior decision that we were having trouble deciding one way or another (not #1878) but I can't recall what.

cblecker:
This PR would fix the specific use case I described here: https://github.com/vmware-tanzu/velero/issues/1878#issuecomment-627542778

skriss:
I've put up PR #2595 which fixes this in a more general way by not erroring when additional items returned from plugins can't be found. I've confirmed in testing that a pod that's pending because it references a nonexistent PVC no longer results in a Velero error, just a warning. Closing this PR out, but appreciate the input & contribution nonetheless!

PR #2280, opened 2020-02-18
Delete secret resource

jenting:
If secret resource not deleted, re-deploy the velero server
with another cloud credential file would failed.

### DIFF
--- a/site/docs/master/uninstalling.md
+++ b/site/docs/master/uninstalling.md
@@ -3,6 +3,7 @@
 If you would like to completely uninstall Velero from your cluster, the following commands will remove all resources created by `velero install`:
 
 ```bash
+kubectl delete secret/cloud-credentials secret/velero-restic-credentials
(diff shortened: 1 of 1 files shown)


skriss:
@jenting these secrets are expected to be in the `velero` namespace, so will be covered by the namespace deletion at https://github.com/vmware-tanzu/velero/pull/2280/files#diff-f0805091efe0ea1dbdf2adb647c0aef1R7 -- can you clarify why this would be needed?

jenting:
I think probably I install Velero in namespace kube-system. I will make a try install in namespace velero.

skriss:
> I think probably I install Velero in namespace kube-system. I will make a try install in namespace velero.

Ah. Yeah, those delete instructions assume a dedicated namespace for velero. If you installed into a shared namespace, then those instructions will not work - you'd have to explicitly delete the deployment, daemonset, service account, secret, etc.

jenting:
Noted! Close this PR.

PR #2288, opened 2020-02-22
Add snapshot group resources

ashish-amarnath:
Add GroupResources for VolumeSnapshot types


pkg/kuberesource/kuberesource.go
```diff
@@ -30,4 +31,7 @@ var (
 	PersistentVolumes         = schema.GroupResource{Group: "", Resource: "persistentvolumes"}
 	Pods                      = schema.GroupResource{Group: "", Resource: "pods"}
 	ServiceAccounts           = schema.GroupResource{Group: "", Resource: "serviceaccounts"}
+	VolumeSnapshotClasses     = schema.GroupResource{Group: snapshotv1beta1api.GroupName, Resource: "volumesnapshotclasses"}
```
skriss: I don't think it's worth vendoring in the snapshots package just to get the group name constant, I'd probably just hardcode it here like the other ones.

ashish-amarnath: That sounds reasonable to me. I'll update this PR

PR #2294, opened 2020-02-24
update upgrade instructions for v1.3

skriss:


skriss:
@carlisia @nrb @ashish-amarnath please review when you get a chance, thanks!

site/docs/v1.3.0-beta.2/upgrade-to-1.3.md
```diff
@@ -0,0 +1,63 @@
...
+
+    ```bash
+    velero version --client-only
+    ```
+
+    You should see the following output:
+
+    ```bash
+    Client:
+        Version: v1.3.0
```
nrb: For the beta, these would report `v1.3.0-beta.2`, wouldn't they? Are you trying to avoid having to submit a separate PR for the final version?

skriss: It makes sense to have these say beta.2 for now; all will be updated when we release 1.3. Fixed.

skriss: I did leave the master copy showing `v1.3.0` so I don't have to change that again; I only changed the beta.2 copy

PR #2295, opened 2020-02-25
Apply backup-name label on resources being backed up

ashish-amarnath:
Default item backupper should label resources when backing up items.
This is important especially for additional items, returned by backupItemActions, to which none of the `ib.backupRequest.ResolvedActions` may apply. This results in these objects to be persisted in the backup without  the `velero.io/backup-name` label being applied to them.

### DIFF
--- a/changelogs/unreleased/2295-ashish-amarnath
+++ b/changelogs/unreleased/2295-ashish-amarnath
@@ -1,0 +1,1 @@
+Apply `velero.io/backup-name` label on all backed up resources.
--- a/pkg/backup/backup_test.go
+++ b/pkg/backup/backup_test.go
@@ -1254,7 +1254,7 @@ func TestBackupActionModifications(t *testing.T) {
 				}),
 			},
 			want: map[string]unstructuredObject{
-				"resources/pods/namespaces/ns-1/pod-1.json": toUnstructuredOrFail(t, builder.ForPod("ns-1", "pod-1").ObjectMeta(builder.WithLabels("updated", "true")).Result()),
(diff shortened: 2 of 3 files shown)


skriss:
> This is important especially for additional items, returned by backupItemActions, to which none of the ib.backupRequest.ResolvedActions may apply. This results in these objects to be persisted in the backup without the velero.io/backup-name label being applied to them.

I don't totally understand this - AFAIK, we don't currently put this label on *any* item as it's being backed up. Items do get labeled with the backup & restore name when they're restored into a cluster, though.

This seems fine to merge regardless, just want to make sure I'm not missing something?

ashish-amarnath:
I was under the understanding that the labels are added to objects at backup time and in the scenario described in the description we would miss adding labels tho these objects.
I am going to close this and re-open later if this is required.

skriss:
Gotcha. Yeah, I know we definitely add them during restore, at https://github.com/vmware-tanzu/velero/blob/master/pkg/restore/restore.go#L1131-L1134.

Like I said, I don't see any issue with adding this label during backup, so if you want to get this merged anyway, it's fine with me.

ashish-amarnath:
I think adding this label will violate some assumptions on which the csi integration was designed. For that reason, I will hold off on this PR.

nrb:
I believe it should only matter for objects on restore (per [this commit](https://github.com/vmware-tanzu/velero/pull/1661/commits/e7ddb0b5d60212fecbf3b996803340dbca1e0252)), but order of operations may be a problem there.

Issue #131, opened 2017-10-11
Move validation from controllers to pkg/backup, pkg/restore


ncdc:
Make the validation the first thing we do in Backup() and Restore(). WDYT @skriss?

skriss:
yeah, makes sense to me to put it in the library

rosskukulinski:
Internal development enhancement, but this is P1 because right now we have backup and restore logic split in 2 different packages. In addition, if anyone wants to vendor ark to do backups/restores, it's not pretty. We want a self-contained package for each.

skriss:
Here's an initial mini-design for this issue:

**Goals**
- Move all backup/restore-processing logic into `pkg/backup`, `pkg/restore` respectively so there's a single entrypoint for running a backup/restore which can be more easily tested, vendored, etc.
- Make controllers "dumb" - only responsible for watching for new/updated custom resources, enqueuing them, and dequeueing them and passing them to the appropriate packages for work

**Non-Goals**
- Full refactoring of `pkg/backup` & `pkg/restore`.  The goal of this initial effort is just to draw a clear boundary around the end-to-end backup/restore processes.  From there, integration tests can be written against this boundary to provide guardrails for refactoring the implementations.
- Similar refactoring for other controllers.  This should probably be undertaken later, but first priority is for core backup and restore flows.

**Design**
Entrypoints to packages should be very simple:
```go
// <pkg/backup>
type Pipeline interface {
    Execute(namespace, name string) error
}
```
- Backup pipeline will:
  - Ensure API object exists
  - Ensure API object has the correct phase for processing (blank or `New`)
  - Default unspecified API object fields
    - `.spec.ttl`
    - `.spec.storageLocation`
    - `.spec.volumeSnapshotLocations`
    - `.status.version`
  - Validate API object spec
    - included/excluded resources
    - included/excluded namespaces
    - backup storage location
    - volume snapshot locations
  - Update API object's status to either `FailedValidation` (with errors) or `InProgress`
  - Instantiate all the things
  - Run API-scraping & related actions/hooks
  - Upload artifacts (archive, log, volumesnapshots file) to storage
  - Update API object's final status
- Restore pipeline will work similarly (to be detailed after going through backup pipeline & learning from it)

eleanor-millman:
Closing because it should naturally fall out of the Astrolabe work if we do that.

PR #2299, opened 2020-02-25
wait for informer caches to sync before running controllers

### LINKED ISSUE #616: Use shared informer factory's WaitForCacheSync in server.go
The shared informer factory has a `WaitForCacheSync` method that we can call that will wait for all started informers' caches to sync. For all the places where we have code individually calling `cache.WaitForCacheSync()` with a list of informers **from our standard shared informer factories**, we should be able to remove these calls and rely on a single `factory.WaitForCacheSync()` call instead, in `server.go`.

We also have some informers that are created by hand, separate from the shared informer factories. For these informers, we still need to call `cache.WaitForCacheSync` separately.

skriss:

closes #616  (already closed)

@carlisia @nrb @ashish-amarnath wanted to get your input on this change before making it for all the controllers.  This moves the logic that waits for informers' caches to sync out of each individual controller, and into the main server process, before running any controllers. You can see #616 for a little more detail.

I want to do this because it makes some other refactoring that I'd like to do (basically #131) easier by not requiring each struct that uses an informer/lister to wait for cache sync before using it.

If it looks good to you, I'll continue changing all the other controllers!


ashish-amarnath:
I am onboard w/ the idea here 👍

skriss:
@ashish-amarnath FYI, I'm hoping that this followed by #131 puts us in a better position to contemplate a potential migration to kubebuilder, since most of the processing logic will have been extracted from our controllers and put into libraries. It also sets us up to more easily move to using worker pods (one per backup/restore) if we decide to go that route.

skriss:
I went ahead and updated all the controllers.

Note we'll hold this until *after* releasing v1.3, so I'm leaving it as a draft.

skriss:
ready for review/merge now that v1.3.0 has been tagged.

PR #2305, opened 2020-02-27
Add initial documentation for CSI support

### LINKED ISSUE #2142: Define and document the `EnableCSI` feature flag
Feature flags were introduced with #1798, and we should use this to keep the beta functionality from breaking existing code for our users.

This issue can be considered closed when there's a check for the flag and there's a page documenting the flag's existence - it doesn't need to be too complicated.

nrb:
Fixes #2142

I'm open to moving the location of the CSI page in the ToC - however I think for now it's ok to keep it all in 1 spot so our more ambitious users have all the info in one place. Once it's moved to GA, then I think the content should be good to move into relevant portions like Use, Install, Troubleshoot, etc.

### DIFF
--- a/site/_data/master-toc.yml
+++ b/site/_data/master-toc.yml
@@ -39,6 +39,8 @@ toc:
         url: /namespace
       - page: Extend with hooks
         url: /hooks
+      - page: CSI Support (beta)
+        url: /csi
   - title: Plugins
     subfolderitems:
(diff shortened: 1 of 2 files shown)


fragolinux:
hi, any news on this? Taken a look at values and templates, no reference to EnableCSI in chart... so the only way is to go via cli? thanks in advance

PR #2314, opened 2020-03-03
Document using a public URL to access logs

### LINKED ISSUE #2256: better document using `spec.publicUrl` to access logs/backups from minIO

nrb:
Fixes #2256 

Will copy the section into the 1.3.0 section if the text looks good where it is.

### DIFF
--- a/site/docs/master/contributions/minio.md
+++ b/site/docs/master/contributions/minio.md
@@ -233,7 +233,7 @@ You must also get the Minio URL, which you can then specify as the value of the 
 
 If you're using Minio with HTTPS, you may see unintelligible text in the output of `velero describe`, or `velero logs` commands.
 
-In order to fix this, you can add a public, HTTP URL to the `BackupStorageLocation`, or a [self-signed certificate][32].
+In order to fix this, you can add a public URL to the `BackupStorageLocation`.
 
 In a terminal, run the following:
 
@@ -241,7 +241,9 @@ In a terminal, run the following:
 kubectl edit backupstoragelocation default -n velero
 ```
 
-Add `publicUrl: http://<a public IP for your Minio instance>:9000` under the `spec.config` section.
+Add `publicUrl: https://<a public IP for your Minio instance>:9000` under the `spec.config` section.
+
+If your certificate is self-signed, see the [docuementation on self-signed certificates][32].
 
 ## Expose Minio outside your cluster with Kubernetes in Docker (KinD):
 
--- a/site/docs/master/contributions/minio.md
+++ b/site/docs/master/contributions/minio.md
@@ -229,6 +229,20 @@ You must also get the Minio URL, which you can then specify as the value of the 
 
 1.  Edit your `BackupStorageLocation` YAML, adding `publicUrl: <URL_FROM_PREVIOUS_STEP>` as a field under `spec.config`. You must include the `http://` or `https://` prefix.
 
+## Accessing logs with an HTTPS endpoint
+
+If you're using Minio with HTTPS, you may see unintelligible text in the output of `velero describe`, or `velero logs` commands.
+
+In order to fix this, you can add a public, HTTP URL to the `BackupStorageLocation`, or a [self-signed certificate][32].
+
+In a terminal, run the following:
+
+```shell
+kubectl edit backupstoragelocation default -n velero
+```
+
+Add `publicUrl: http://<a public IP for your Minio instance>:9000` under the `spec.config` section.
+
 ## Expose Minio outside your cluster with Kubernetes in Docker (KinD):
 
 Kubernetes in Docker currently does not have support for NodePort services (see [this issue](https://github.com/kubernetes-sigs/kind/issues/99)). In this case, you can use a port forward to access the Minio bucket.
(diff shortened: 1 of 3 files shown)


site/docs/master/contributions/minio.md
```diff
@@ -249,6 +249,20 @@ kubectl edit backupstoragelocation default -n velero
 
 Add `publicUrl: http://localhost:9000` under the `spec.config` section.
 
+## Accessing logs with an HTTPS endpoint
+
+If you're using Minio with HTTPS, you may see unintelligible text in the output of `velero describe`, or `velero logs` commands.
+
+In order to fix this, you can add a public, non-HTTPS URL to the `BackupStorageLocation`.
```
skriss: Is it true that this URL can't be `https`? Why is that?

nrb: The errors have usually been because the client is accessing an HTTPS endpoint with a custom certificate, so either it has to be HTTP, or the certificate has to be trusted on the client side.

I should probably clarify this wording.

skriss: Gotcha, that makes sense.

nrb:
For reviewers - this needs a change before merge. I need to clarify that HTTPS endpoints are often tricky because of self-signed certs.

skriss:
@nrb it'd be great to wrap this up if you can. If not, we should probably close it out for now.

nrb:
@skriss Yeah, this honestly doesn't need much more work so let me get this in for today.

nrb:
@skriss Ok, updated and added notes for v1.4.0's custom certs.

site/docs/master/contributions/minio.md
```diff
@@ -229,6 +229,20 @@ You must also get the Minio URL, which you can then specify as the value of the
 
 1.  Edit your `BackupStorageLocation` YAML, adding `publicUrl: <URL_FROM_PREVIOUS_STEP>` as a field under `spec.config`. You must include the `http://` or `https://` prefix.
 
+## Accessing logs with an HTTPS endpoint
+
+If you're using Minio with HTTPS, you may see unintelligible text in the output of `velero describe`, or `velero logs` commands.
+
+In order to fix this, you can add a public, HTTP URL to the `BackupStorageLocation`, or a [self-signed certificate][32].
```
skriss, line 236: If you're using TLS with minio, is there actually an `http`-only endpoint? My impression was that it was one or the other, not both.

My understanding of this issue was that either (a) the user didn't provide a `spec.publicUrl` that was routable from the end-user's machine; or (b) that they were using a self-signed cert for minio which wasn't previously supported by velero.

My point is, from my understanding, "adding an HTTP URL to the spec.publicUrl" isn't actually a valid option if the user is using TLS. Am I missing something?

nrb: I just asked for clarification from the user. It's possible that they reconfigured minio to just disable TLS.

I re-read the minio docs, and I think I was misunderstanding. Your understanding sounds correct - minio can't provide both HTTPS and HTTP endpoints at the same time.

My interpretation of the issue was that it was scenario (b), since data was retrieved, it was just unreadable nonsense.

So pending any update from the reporter, I think I should reword the v1.3.2 version to advise users to upgrade to v1.4.0 for (b).

site/docs/master/contributions/minio.md
```diff
@@ -229,6 +229,22 @@ You must also get the Minio URL, which you can then specify as the value of the
...
+
+In a terminal, run the following:
+
+```shell
+kubectl edit backupstoragelocation default -n velero
+```
+
+Add `publicUrl: https://<a public IP for your Minio instance>:9000` under the `spec.config` section.
+
+If your certificate is self-signed, see the [docuementation on self-signed certificates][32].
```
skriss, line 246: docuementation => documentation

site/docs/master/contributions/minio.md
```diff
@@ -229,6 +229,22 @@ You must also get the Minio URL, which you can then specify as the value of the
...
+## Accessing logs with an HTTPS endpoint
+
+If you're using Minio with HTTPS, you may see unintelligible text in the output of `velero describe`, or `velero logs` commands.
+
+In order to fix this, you can add a public URL to the `BackupStorageLocation`.
+
+In a terminal, run the following:
+
+```shell
+kubectl edit backupstoragelocation default -n velero
```
skriss, line 241: I've started trying to use `kubectl patch` commands for things like this where I can since they don't require interactive editing -- this one would be:

```
kubectl patch -n velero backupstoragelocation default --type merge -p '{"spec":{"config":{"publicUrl":"<...the value>"}}}'
```

Issue #2145, opened 2019-12-16
sync backups' VolumeSnapshotContents into the cluster as part of the backup sync controller


nrb:
Given https://github.com/vmware-tanzu/velero-plugin-for-csi/issues/4, there's a scenario where we may run into an orphaned snapshot.

Assuming a backup is created, with a VolumeSnapshot & a VolumeSnapshotContent in the tarball stored in object storage.

Next, delete the cluster from which the backup was created. Create a new cluster.

Sync the backups in to the new cluster, but do not restore any of the VolumeSnapshots/VolumeSnapshotContents.

From there, as `Backup`s expire, they will be garbage collected. However, if the VolumeSnapshots/VolumeSnapshotContents that reference any garbage collected Backups are restored, what happens to them? Will Kubernetes automatically delete them?

skriss:
Another thought - we talked about and discarded the idea of importing `VolumeSnapshots` into the cluster as part of the backup sync controller's processing, mainly because these are namespaced resources and that might require creating workload namespaces in the cluster, which would be weird. However - should we consider syncing `VolumeSnapshotContents` into the cluster (which are cluster-scoped)? Does that help at all here?

nrb:
Given #2302, I think it might make sense to do that now.

It probably also makes sense to look for the associated `Backup` and/or `VolumeSnapshotContents` when creating a `VolumeSnapshot` on restore and skip restoring if it doesn't exist.

skriss:
I updated the issue title assuming that we want to go down the path of syncing `VolumeSnapshots` into the cluster, but feel free to comment if you see other paths.

skriss:
@nrb @ashish-amarnath I think we've decided against doing this, so this issue can be closed - right?

skriss:
@nrb @ashish-amarnath we need to decide if we're going to do this. I think it potentially still makes sense, since without it we still have the issue of orphaning storage snapshots in certain scenarios (take backup/snapshots in cluster A, lose cluster A, import backups into cluster B, expire backup from cluster B before it's restored -- storage snapshots are now orphaned).

Thoughts?

ashish-amarnath:
[...]

Issue #1568, opened 2019-06-11
Include CSI snapshot details in backup describe command


nrb:
The `describe` commands currently have no way to display VolumeSnapshot/VolumeSnapshotContents associated with a given backup or restore.

On the server side, the following needs to change:

* persistBackup will be extended to query for all VolumeSnapshots associated with the backup, and persist the list to JSON.
* BackupStore.PutBackup will receive an additional argument, volumeSnapshots io.Reader, that contains the JSON representation of VolumeSnapshots. This will be written to a file named csi-snapshots.json.gz.

On the client side:

* DescribeBackupStatus will be extended to download the csi-snapshots.json.gz file for processing. GitHub Issue 1568 captures this work.
* A new describeCSIVolumeSnapshots function should be added to the output package that knows how to render the included VolumeSnapshot names referenced in the csi-snapshots.json.gz file.

skriss:
Are there any other low-cost options here that are "good enough" for the prototype?  I don't think we need a super-polished UX for the prototype phase - adding labels/annotations to the backup CR, or a `kubectl get volumesnapshots -l foo=bar` sample command, etc. could work fine for the prototype phase. If we do end up needing to do this or any other in-tree code changes for the prototype, though, I imagine we want to set up a branch specifically for the prototype that likely will never get merged into master, that prototype testers can build and use to exercise the functionality.

nrb:
This is more long term, hence the P2 label. IMO it doesn’t need to be in the prototype, but will be necessary for full support, so we could even move it out of v1.1.

That said, if a namespace is deleted, the volumesnapshots won’t be around to kubectl get. They’re stored in the same namespace the PVC is in.

skriss:
OK, yeah, makes sense. I think it'd be fine if for the prototype we just expected users to "know" which snapshots were in their backups. Definitely a must-have for GA, though.

PR #2323, opened 2020-03-05
Upload CSI snapshot information to object storage

nrb:
Prerequisite for #1568 & #2145.

Putting up a draft to socialize my changes and locations early.


nrb:
@ashish-amarnath FYI, though there's not a lot of work actually being done in the PR yet.

go.mod
```diff
@@ -31,11 +30,11 @@ require (
...
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.4.0
-	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
-	google.golang.org/grpc v1.23.1
+	golang.org/x/net v0.0.0-20191209160850-c0dbc17a3553
+	google.golang.org/grpc v1.26.0
 	k8s.io/api v0.17.0
 	k8s.io/apiextensions-apiserver v0.17.0
-	k8s.io/apimachinery v0.17.0
+	k8s.io/apimachinery v0.17.1-beta.0
```
skriss: Was this intentional?

pkg/cmd/server/server.go
```diff
@@ -266,9 +273,19 @@ func newServer(f client.Factory, config serverConfig, logger *logrus.Logger) (*s
 
 	clientConfig, err := f.ClientConfig()
 	if err != nil {
+		cancelFunc()
```
skriss: I think we should probably `defer cancelFunc()` right after L272 instead of adding it the the various error handling sections

pkg/cmd/server/server.go
```diff
@@ -584,6 +604,27 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
 		cmd.CheckError(err)
 
+		// Make empty listers that will only be populated if CSI is properly enabled.
+		var vsLister snapshotv1beta1listers.VolumeSnapshotLister
+		var vscLister snapshotv1beta1listers.VolumeSnapshotContentLister
+
+		// We don't care about the returned types, just whether or not there's a 404 error.
+		_, err = s.discoveryClient.ServerResourcesForGroupVersion(snapshotv1beta1api.SchemeGroupVersion.String())
+		switch {
+		case apierrors.IsNotFound(err) && !features.IsEnabled("EnableCSI"):
```
skriss: Should we wrap this whole block (L611-626) in a `if features.IsEnabled("EnableCSI")`, so we don't even execute the discoveryClient call if it's not enabled? Then we can remove this first case and remove the check on the feature flag from all the others. Simpler and also avoids the code path for when the feature flag is not enabled.

pkg/cmd/server/server.go
```diff
@@ -278,12 +295,15 @@ func newServer(f client.Factory, config serverConfig, logger *logrus.Logger) (*s
...
 		sharedInformerFactory: informers.NewSharedInformerFactoryWithOptions(veleroClient, 0, informers.WithNamespace(f.Namespace())),
-		ctx:                   ctx,
-		cancelFunc:            cancelFunc,
-		logger:                logger,
-		logLevel:              logger.Level,
-		pluginRegistry:        pluginRegistry,
-		config:                config,
+		// If no namespace is specified, all namespaces are watched.
+		// This is desirable for VolumeSnapshots, as we want to query for all VolumeSnapshots across all namespaces using this informer
+		snapshotterSharedInformerFactory: snapshotvebeta1informers.NewSharedInformerFactoryWithOptions(csiSnapClient, 0),
```
skriss: I think we may want to consider wrapping this instantiation, and all subsequent uses of the informer factory, in a feature-flag check too

pkg/persistence/object_store_layout.go
```diff
@@ -98,3 +98,12 @@ func (l *ObjectStoreLayout) getRestoreLogKey(restore string) string {
 func (l *ObjectStoreLayout) getRestoreResultsKey(restore string) string {
 	return path.Join(l.subdirs["restores"], restore, fmt.Sprintf("restore-%s-results.gz", restore))
 }
+
+// TODO: Add keys for VS & VSContents json.gz files
+func (l *ObjectStoreLayout) getCSIVolumeSnapshotContentsKey(backup string) string {
```
skriss: is this supposed to be just `getCSIVolumeSnapshotsKey`? (no "Contents")

pkg/persistence/object_store.go
```diff
@@ -56,6 +58,7 @@ type BackupStore interface {
 	GetBackupVolumeSnapshots(name string) ([]*volume.Snapshot, error)
 	GetPodVolumeBackups(name string) ([]*velerov1api.PodVolumeBackup, error)
 	GetBackupContents(name string) (io.ReadCloser, error)
+	// TODO(nrb-csi): Do we need Get methods for the CSI types?
```
skriss: I think we will for the CLI commands and the backup sync controller, but can add them in those PRs.

pkg/controller/backup_controller.go
```diff
@@ -37,8 +37,14 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/cache"
 
+	snapshotv1beta1api "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
+	snapshotv1beta1 "github.com/kubernetes-csi/external-snapshotter/v2/pkg/client/clientset/versioned/typed/volumesnapshot/v1beta1"
+	snapshotv1beta1listers "github.com/kubernetes-csi/external-snapshotter/v2/pkg/client/listers/volumesnapshot/v1beta1"
+
+	v1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
skriss: double import with the below

pkg/controller/backup_controller.go
```diff
@@ -56,23 +62,28 @@ import (
...
+	defaultSnapshotLocations    map[string]string
+	metrics                     *metrics.ServerMetrics
+	newBackupStore              func(*velerov1api.BackupStorageLocation, persistence.ObjectStoreGetter, logrus.FieldLogger) (persistence.BackupStore, error)
+	formatFlag                  logging.Format
+	volumeSnapshotLister        snapshotv1beta1listers.VolumeSnapshotLister
+	volumeSnapshotContentLister snapshotv1beta1listers.VolumeSnapshotContentLister
 }
 
+// TODO(nrb-csi): Add clients for the VS/VSContent here.
+// How about creating them in the server iff EnableCSI, and if they're nil in the backup controller, don't worry about it?
```
skriss: that seems reasonable to me.

pkg/controller/backup_controller.go
```diff
@@ -587,19 +624,27 @@ func recordBackupMetrics(log logrus.FieldLogger, backup *velerov1api.Backup, bac
...
 	errs := []error{}
 	backupJSON := new(bytes.Buffer)
 
 	if err := encode.EncodeTo(backup.Backup, "json", backupJSON); err != nil {
 		errs = append(errs, errors.Wrap(err, "error encoding backup"))
 	}
 
-	volumeSnapshots := new(bytes.Buffer)
-	gzw := gzip.NewWriter(volumeSnapshots)
+	// Velero-native volume snapshots
```
skriss: There's a lot of duplication in this function for writing resources to a gzipped JSON writer here (backup.VolumeSnapshots, backup.PodVolumeBackups, csiVolumeSnapshots, volumeSnapshotContents, backup.BackupResourceList()) - we can probably extract a helper func, and/or change this code to be a loop over the relevant items, where the loop body handles the json+gzip writing.

pkg/persistence/object_store.go
```diff
@@ -247,6 +250,31 @@ func (s *objectBackupStore) PutBackup(info BackupInfo) error {
 		return kerrors.NewAggregate(errs)
 	}
 
+	// TODO(nrb-csi): only upload these if they're non-nil and/or if EnableCSI is on
```
skriss: I would like to figure out how to remove the duplication in this function too, but need to look at it some more.

go.mod
nrb: Nope, it was not.

pkg/controller/backup_controller.go
nrb: I believe with the listers, we don't need the clients now, but I think similar logic applies, and that's what I tried to follow.

pkg/persistence/object_store_layout.go
nrb: Yep, bad copy/paste

pkg/cmd/server/server.go
nrb: Yeah, this was a quick fix for linting complaints, but a `defer` would be better.

pkg/cmd/server/server.go
nrb: I actually had a function that did something similar before I pushed up any of this, but it was based on gating access to the clients. I'll see about tweaking it to add it for the informer factory.

go.mod
nrb: Ok, after removing and re-adding the `go.mod` and `go.sum` files, this is because of `external-snapshotter`'s dependency. https://github.com/kubernetes-csi/external-snapshotter/blob/master/go.mod#L15

pkg/controller/backup_controller.go
nrb: How important is de-duping the internals of this function to this PR? I can see the need, and can spend some time on it, just not sure of where to include it.

nrb:
There's a bunch of TODOs in the code still, but the core of the PR works - the objects are uploaded to storage.

pkg/controller/backup_controller.go
nrb: Ok, latest commits have moved the serialization/compression into a helper function, so hopefully this is easier to follow.

pkg/persistence/object_store.go
nrb: I've deduped this except for the special cases at the top by using a map for the keys/contents and looping over those for upload. In my testing, this appears to have continued working as expected.

nrb:
@ashish-amarnath Please take a look - I think this could introduce some useful things for #2411

skriss:
planning to review again this afternoon

pkg/cmd/server/server.go
```diff
@@ -212,23 +218,25 @@ func NewCommand(f client.Factory) *cobra.Command {
...
-	pluginRegistry        clientmgmt.Registry
-	resticManager         restic.RepositoryManager
-	metrics               *metrics.ServerMetrics
-	config                serverConfig
+	namespace                        string
+	metricsAddress                   string
+	kubeClientConfig                 *rest.Config
+	kubeClient                       kubernetes.Interface
+	veleroClient                     clientset.Interface
+	csiSnapClient                    snapshotvebeta1client.Interface
```
skriss, line 226: this field doesn't appear to be used anywhere

pkg/cmd/server/server.go
```diff
@@ -832,3 +888,34 @@ func (s *server) runProfiler() {
...
+		w.factory.Start(stopCh)
+	}
+}
+
+// WaitForCacheSync proxies the WaitForCacheSync call to the CSI SharedInformerFactory.
+func (w *CSIInformerFactoryWrapper) WaitForCacheSync(stopCh <-chan struct{}) map[reflect.Type]bool {
+	if features.IsEnabled("EnableCSI") {
+		return w.factory.WaitForCacheSync(stopCh)
+	}
+	return make(map[reflect.Type]bool, 0)
```
skriss, line 920: I think you could just return nil here, don't need to make an empty map

pkg/controller/backup_controller.go
```diff
@@ -56,24 +60,28 @@ import (
...
+	snapshotLocationLister      velerov1listers.VolumeSnapshotLocationLister
+	defaultSnapshotLocations    map[string]string
+	metrics                     *metrics.ServerMetrics
+	newBackupStore              func(*velerov1api.BackupStorageLocation, persistence.ObjectStoreGetter, logrus.FieldLogger) (persistence.BackupStore, error)
+	formatFlag                  logging.Format
+	volumeSnapshotLister        snapshotv1beta1listers.VolumeSnapshotLister
+	volumeSnapshotContentLister snapshotv1beta1listers.VolumeSnapshotContentLister
 }
 
+// TODO(nrb-csi): Add clients for the VS/VSContent here.
```
skriss, line 83: can we remove this comment now?

pkg/controller/backup_controller.go
```diff
@@ -532,6 +543,28 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
...
+	// This way, we only make the Lister call if the feature flag's on.
+	var volumeSnapshots []*snapshotv1beta1api.VolumeSnapshot
+	var volumeSnapshotContents []*snapshotv1beta1api.VolumeSnapshotContent
+	if features.IsEnabled("EnableCSI") {
+		selector := labels.SelectorFromSet(map[string]string{velerov1api.BackupNameLabel: backup.Name})
+
+		// TODO(nrb-csi): Only run listers methods if the listers aren't nil, just in case
+		volumeSnapshots, err = c.volumeSnapshotLister.List(selector)
+		if err != nil {
+			//TODO: Is this right?
```
skriss, line 556: seems right, this will output it to both the backup log file and to stdout

pkg/controller/backup_controller.go
```diff
@@ -532,6 +543,28 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
...
+			backupLog.Error(err)
+		}
+		// Currently, VSCs are not being labelled in the plugin, so none will be returned here.
+		volumeSnapshotContents, err = c.volumeSnapshotContentLister.List(selector)
+		if err != nil {
+			//TODO: Is this right?
```
skriss, line 562: seems right

pkg/persistence/object_store.go
```diff
@@ -215,40 +218,28 @@ func (s *objectBackupStore) PutBackup(info BackupInfo) error {
...
-		errs = append(errs, deleteErr)
+			// attempt to clean up the object in case it was made but we couldn't write contents.
+			deleteErr := s.objectStore.DeleteObject(s.bucket, key)
+			errs = append(errs, deleteErr)
 
-		deleteErr = s.objectStore.DeleteObject(s.bucket, s.layout.getBackupMetadataKey(info.Name))
-		errs = append(errs, deleteErr)
-
-		return kerrors.NewAggregate(errs)
+			deleteErr = s.objectStore.DeleteObject(s.bucket, s.layout.getBackupMetadataKey(info.Name))
```
skriss, line 239: it looks like the old code deleted both the backup contents and the backup metadata here -- new code is missing the delete of the backup contents

pkg/persistence/object_store_layout.go
```diff
@@ -99,3 +99,11 @@ func (l *ObjectStoreLayout) getRestoreLogKey(restore string) string {
 func (l *ObjectStoreLayout) getRestoreResultsKey(restore string) string {
 	return path.Join(l.subdirs["restores"], restore, fmt.Sprintf("restore-%s-results.gz", restore))
 }
+
+func (l *ObjectStoreLayout) getCSIVolumeSnapshotKey(backup string) string {
+	return path.Join(l.subdirs["backups"], backup, fmt.Sprintf("%s-csivolumesnapshot.json.gz", backup))
```
skriss, line 104: let's pluralize the file name -- `<backup name>-csivolumesnapshots.json.gz`

pkg/cmd/server/server.go
```diff
@@ -262,28 +270,43 @@ func newServer(f client.Factory, config serverConfig, logger *logrus.Logger) (*s
...
 	ctx, cancelFunc := context.WithCancel(context.Background())
 
 	clientConfig, err := f.ClientConfig()
 	if err != nil {
+		cancelFunc()
 		return nil, err
 	}
 
+	var csiSnapClient *snapshotvebeta1client.Clientset
+	if features.IsEnabled("EnableCSI") {
```
ashish-amarnath, line 286: suggest creating constants for feature names and export them. That way, we can use the same constant in the plugin as well to decide whether or not to register the plugins

pkg/controller/backup_controller.go
```diff
@@ -595,60 +629,66 @@ func recordBackupMetrics(log logrus.FieldLogger, backup *velerov1api.Backup, bac
...
-
-	if err := json.NewEncoder(gzw).Encode(backup.VolumeSnapshots); err != nil {
-		errs = append(errs, errors.Wrap(err, "error encoding list of volume snapshots"))
-	}
-	if err := gzw.Close(); err != nil {
-		errs = append(errs, errors.Wrap(err, "error closing gzip writer"))
+	// Velero-native volume snapshots (as opposed to CSI ones)
+	nativeVolumeSnapshots, err := encodeToJSONGzip(backup.VolumeSnapshots)
+	if err != nil {
+		errs = append(errs, err)
```
ashish-amarnath, line 649: is this change required?
I feel the previous version was more informative. WDYT?

pkg/cmd/server/server.go
nrb: Yep, this was left-over, will remove it.

pkg/cmd/server/server.go
nrb: I like the idea of making the string a constant & exporting it, but I'm not sure about registering the plugin based on it. That decision can be done outside this PR, though.

pkg/cmd/server/server.go
nrb: I actually don't think I can - iirc, I returned `nil` initially, and it broke. I'll test again, though.

pkg/controller/backup_controller.go
nrb: Yep.

pkg/controller/backup_controller.go
nrb: The reason I had the question was whether or not it should be considered an error. If we're backing up with CSI and volumes included in the backup is >= 1, then we should have no error. If volumes included in the backup == 0, then I expect a 404; I'll need to test this case.
 
As written, I think this code will throw an error for backups that don't include volumes, so we probably want to add that condition.

pkg/controller/backup_controller.go
nrb: You mean using the helper function?

I don't think it's strictly required, no. It was mostly based on a review comment to reduce repetition in the function.

pkg/persistence/object_store.go
nrb: Ah, I see - I misread each of the blocks and what they were doing. I'll correct it.

pkg/persistence/object_store_layout.go
nrb: Sounds good. I think I may also add another dash to make the csi part stand out `<backup name>-csi-volumesnapshots.json.gz`

I think I'll do similar for the VSC file so that they're next to each other in the bucket to help avoid confusion.

pkg/controller/backup_controller.go
ashish-amarnath: Not the helper function, but the error wrapping part.

nrb: Oh, I see - so the error wrapping should be the same, just split between the function and up here, but the one thing we are losing is the information about exactly what's being encoded, which would definitely make debugging harder.

I had considered adding a string for the error message as input to the function, so it would look like `encodeToJSONGzip(backup.VolumeSnapshots, "list of volumesnapshots")`, and the error would be more descriptive when returned.

Would that help?

nrb: For example, L649 here should be equivalent to L610 in the previous code.

pkg/cmd/server/server.go
```diff
@@ -212,23 +218,25 @@ func NewCommand(f client.Factory) *cobra.Command {
...
+	metricsAddress                   string
+	kubeClientConfig                 *rest.Config
+	kubeClient                       kubernetes.Interface
+	veleroClient                     clientset.Interface
+	csiSnapClient                    snapshotvebeta1client.Interface
+	discoveryClient                  discovery.DiscoveryInterface
+	discoveryHelper                  velerodiscovery.Helper
+	dynamicClient                    dynamic.Interface
+	sharedInformerFactory            informers.SharedInformerFactory
+	snapshotterSharedInformerFactory *CSIInformerFactoryWrapper
```
ashish-amarnath, line 231: suggest using `CSI` somewhere in the name of this field.
may be `csiSnapshotterShareInformerFactory`

pkg/cmd/server/server.go
nrb: This appears to work, so I was thinking of some other error with a nil map. I'll include this change in my next set of review feedback changes.

pkg/controller/backup_controller.go
nrb: I've added a second argument to help produce more contextual errors, so hopefully they're less confusing when they happen.

nrb:
@skriss @ashish-amarnath I believe I've addressed all but issue completely now.

pkg/controller/backup_controller.go
ashish-amarnath: ah. That's correct.  My bad 🤦

pkg/persistence/object_store.go
```diff
@@ -215,40 +218,28 @@ func (s *objectBackupStore) PutBackup(info BackupInfo) error {
...
-		errs = append(errs, deleteErr)
-
-		return kerrors.NewAggregate(errs)
+	// Since the logic for all of these files is the exact same except for the name and the contents,
+	// use a map literal to iterate through them and write them to the bucket.
+	var backupObjs = map[string]io.Reader{
+		s.layout.getPodVolumeBackupsKey(info.Name):       info.PodVolumeBackups,
+		s.layout.getBackupVolumeSnapshotsKey(info.Name):  info.VolumeSnapshots,
+		s.layout.getBackupResourceListKey(info.Name):     info.BackupResourceList,
+		s.layout.getCSIVolumeSnapshotKey(info.Name):      info.CSIVolumeSnapshots,
```
ashish-amarnath, line 227: I am thinking if we should use this to get the CSI snapshots to delete instead of what I am doing here https://github.com/vmware-tanzu/velero/pull/2411/files#diff-afe709fa1785e08707ca0825e3f60e68R474
This would be consistent w/ how we discover and delete non-csi volumesnapshots

nrb: That would unify the approaches for sure. I think this goes to @skriss's questions about synchronization - which is the source of truth on the CSI snapshots, the object storage, or the kube API server?

Historically, the object storage has been Velero's source of truth. It may make sense to keep going with that approach for these objects.

Before we can rely on it completely, we need to ensure that Volume Snapshot Contents objects are labeled with the backup's name at backup time, as they currently are not, so the file's always empty now.

nrb: Also, for what it's worth, when deleting a Velero backup, I think we'll need to delete objects in the cluster and from object storage, regardless of which is the source of truth. But deleting them from object storage "should" be as simple as removing the backup's root key, which would then take the rest of the contents.

pkg/controller/backup_controller.go
```diff
@@ -598,60 +634,66 @@ func recordBackupMetrics(log logrus.FieldLogger, backup *velerov1api.Backup, bac
...
-	backupResourceList := new(bytes.Buffer)
-	gzw = gzip.NewWriter(backupResourceList)
+	csiSnapshotJSON, err := encodeToJSONGzip(csiVolumeSnapshots, "csi volume snapshots list")
+	if err != nil {
+		errs = append(errs, err)
+	}
 
-	if err := json.NewEncoder(gzw).Encode(backup.BackupResourceList()); err != nil {
-		errs = append(errs, errors.Wrap(err, "error encoding backup resource list"))
+	snapshotContentsJSON, err := encodeToJSONGzip(volumeSnapshotContents, "volume snapshot contents list")
```
skriss, line 667: nit: "csi volume snapshot contents list"

pkg/controller/backup_controller.go
```diff
@@ -598,60 +634,66 @@ func recordBackupMetrics(log logrus.FieldLogger, backup *velerov1api.Backup, bac
 	serverMetrics.RegisterVolumeSnapshotFailures(backupScheduleName, backup.Status.VolumeSnapshotsAttempted-backup.Status.VolumeSnapshotsCompleted)
 }
 
-func persistBackup(backup *pkgbackup.Request, backupContents, backupLog *os.File, backupStore persistence.BackupStore, log logrus.FieldLogger) []error {
+func persistBackup(backup *pkgbackup.Request,
+	backupContents, backupLog *os.File,
+	backupStore persistence.BackupStore,
+	log logrus.FieldLogger,
+	csiVolumeSnapshots []*snapshotv1beta1api.VolumeSnapshot,
+	volumeSnapshotContents []*snapshotv1beta1api.VolumeSnapshotContent,
```
skriss, line 642: nit: "csiVolumeSnapshotContents"

pkg/controller/backup_controller.go
skriss, line 667: nit: "csiSnapshotContentJSON"

pkg/controller/backup_controller.go
```diff
@@ -668,3 +710,18 @@ func closeAndRemoveFile(file *os.File, log logrus.FieldLogger) {
...
 	}
 }
+
+// encodeToJSONGzip takes arbitrary Go data and encodes it to GZip compressed JSON in a buffer, as well as a description of the data to put into an error should encoding fail.
+func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, error) {
+	buf := new(bytes.Buffer)
+	gzw := gzip.NewWriter(buf)
+
+	if err := json.NewEncoder(gzw).Encode(data); err != nil {
+		return nil, errors.Wrapf(err, "error encoding %s", desc)
```
skriss, line 720: if we hit this condition, then we never close `gzw` - probably just close it explicitly in here. the challenge is always what to do if *that* returns an error -- one option would be to pass a logger into this function and just log it, and return the encoding error.

pkg/persistence/object_store.go
```diff
@@ -41,7 +41,9 @@ type BackupInfo struct {
 	Log,
 	PodVolumeBackups,
 	VolumeSnapshots,
-	BackupResourceList io.Reader
+	BackupResourceList,
+	CSIVolumeSnapshots,
+	VolumeSnapshotContents io.Reader
```
skriss, line 46: "CSIVolumeSnapshotContents"

pkg/persistence/object_store.go
skriss: Yeah, deleting the `.json.gz` files from object storage should already be handled by this struct's `DeleteBackup` func which deletes everything under the backup prefix.

@ashish-amarnath the issue is that if the VolumeSnapshot/VolumeSnapshotContents don't exist in the cluster, then there's nothing *to* delete -- with native Velero snapshots, we delete the storage snapshot directly via the cloud provider API, but with CSI, all we can delete is the in-cluster custom resources.

pkg/persistence/object_store_layout.go
```diff
@@ -99,3 +99,11 @@ func (l *ObjectStoreLayout) getRestoreLogKey(restore string) string {
 func (l *ObjectStoreLayout) getRestoreResultsKey(restore string) string {
 	return path.Join(l.subdirs["restores"], restore, fmt.Sprintf("restore-%s-results.gz", restore))
 }
+
+func (l *ObjectStoreLayout) getCSIVolumeSnapshotKey(backup string) string {
+	return path.Join(l.subdirs["backups"], backup, fmt.Sprintf("%s-csi-volumesnapshots.json.gz", backup))
+}
+
+func (l *ObjectStoreLayout) getVolumeSnapshotContentsKey(backup string) string {
```
skriss, line 107: nit: "getCSIVolumeSnapshotContentsKey"

pkg/controller/backup_controller.go
```diff
@@ -535,6 +544,32 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
...
+		// when EnableCSI is on.
+		if c.volumeSnapshotLister != nil {
+			volumeSnapshots, err = c.volumeSnapshotLister.List(selector)
+			if err != nil {
+				backupLog.Error(err)
+			}
+		}
+
+		if c.volumeSnapshotContentLister != nil {
+			volumeSnapshotContents, err = c.volumeSnapshotContentLister.List(selector)
```
skriss, line 565: one alternative to requiring a label selector for the VolumeSnapshotContents would be to iterate through `volumeSnapshots`, find which VSC each one is bound to, and then explicitly grab those VSCs from the Lister. Not sure if it makes the flow easier or not - just an option to consider.

pkg/controller/backup_controller.go
nrb: We could also close it with a defer and wrap with a function so we can grab the error, and used a named return on the error to get the error returned out instead of logged... I'm not a huge fan of named returns, but in this case it means both errors would be returned instead of one returned and one logged.

Looking at how the returned errors from the `persistBackup` function are used, they're all fatal, not logged, so I'm leaning towards trying to find a way to return them.

In code, this would be like

```go
defer func () {
}()
```

nrb: Doing it this way may also require some more wrapping, because we don't want the defer's error overwriting any encoding errors, either.

skriss: yeah, you can use that approach too - I usually try to avoid it because it gets complicated quickly. A third option would be to return a `[]error`, and append errors to that as they're encountered. Since the calling function also returns a `[]error`, that wouldn't be too bade.

nrb: 👍 That makes sense, especially since that's what the code originally did anyway. It took all these encoding & closing errors and put them in a slice, so I'll just do the same.

skriss: sgtm!

pkg/controller/backup_controller.go
nrb: I'll see what this looks like. I'm not 100% sure if we even need to upload the VSCs, honestly, but since the rest of the code is there to do it, might as well give it a shot.

Also, I think regardless of implementation here, it would be useful to have the VSCs labeled with the Velero backup name, to indicated Velero made them.

nrb: Yep, that worked 👍

pkg/controller/backup_controller.go
```diff
@@ -641,40 +641,40 @@ func persistBackup(backup *pkgbackup.Request,
 	csiVolumeSnapshots []*snapshotv1beta1api.VolumeSnapshot,
 	csiVolumeSnapshotContents []*snapshotv1beta1api.VolumeSnapshotContent,
 ) []error {
-	errs := []error{}
+	persistErrs := []error{}
 	backupJSON := new(bytes.Buffer)
 
-	if err := encode.EncodeTo(backup.Backup, "json", backupJSON); err != nil {
-		errs = append(errs, errors.Wrap(err, "error encoding backup"))
+	if errs := encode.EncodeTo(backup.Backup, "json", backupJSON); errs != nil {
```
skriss, line 651: this one should just be `err`, since it's only a single error

pkg/controller/backup_controller.go
```diff
@@ -562,10 +562,14 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
 		}
 
 		if c.volumeSnapshotContentLister != nil {
-			volumeSnapshotContents, err = c.volumeSnapshotContentLister.List(selector)
-			if err != nil {
-				backupLog.Error(err)
-
+			// Since VolumeSnapshotContent objects are not currently labeled, get them by using binding from the VolumeSnapshot
+			for _, vs := range volumeSnapshots {
+				vsc, err := c.volumeSnapshotContentLister.Get(*vs.Status.BoundVolumeSnapshotContentName)
```
skriss, line 567: should nil-check here before dereferencing `.Status.BoundVolumeSnapshotContentName`

pkg/controller/backup_controller.go
nrb: I knew selecting the region and doing replace-all was going to bite me somehow.

skriss: 😂

pkg/controller/backup_controller.go
```diff
@@ -564,12 +564,15 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
 		if c.volumeSnapshotContentLister != nil {
 			// Since VolumeSnapshotContent objects are not currently labeled, get them by using binding from the VolumeSnapshot
 			for _, vs := range volumeSnapshots {
-				vsc, err := c.volumeSnapshotContentLister.Get(*vs.Status.BoundVolumeSnapshotContentName)
-				if err != nil {
-					backupLog.Error(err)
-					continue
+				// nil check just in case the snapshot and the content object did not get bound before returning from the plugins
+				if vs.Status.BoundVolumeSnapshotContentName != nil {
```
skriss, line 568: technically `Status` could be nil too, let's check both just in case:

`if vs.Status != nil && vs.Status.BoundVolumeSnapshotContentName != nil {`

skriss:
one more thing...needs a changelog!

pkg/persistence/object_store.go
ashish-amarnath: I think we've gotten what we want from this conversation! marking this as resolved.

skriss:
@ashish-amarnath FYI we usually try to squash & merge PRs like this rather than merging as-is. Keeps the `master` commit history cleaner. I pretty much default to squash & merge unless I know for a fact that the PR author has split the commits out in a particular way that they want to be reflected in the `master` history.

ashish-amarnath:
Yes, I should have done that. I realized that when I rebased my branch. 😐

________________________________

PR #2336, opened 2020-03-10
v1.3.1 docs site

skriss:
cc @nrb


skriss:
Ready for review but leaving in draft until ready to merge. This is a new set of versioned docs for the patch release, with updated upgrade instructions to refer to the new version. Otherwise a copy of the existing docs.

PR #2346, opened 2020-03-17
Adding k8s version labels on backup CRD #2342

### LINKED ISSUE #2342: Label the backup CRD with the cluster k8s version
Looking ahead of using Velero for migrating workloads skipping major k8s releases, we need a watermark on the backup CRD of the k8s version on the source cluster.  

Label the backup CRD with the source cluster k8s version. This label/metadata can be used for further decisions when restoring the backup on a target cluster running a different k8s version.
Example of the label:
velero.io/source-cluster-k8s-version=v1.16.4

I am working on the contribution and will submit a PR once it is tested.

**Environment:**

- Velero version (use `velero version`): 1.3.1
- Kubernetes version (use `kubectl version`): 1.17
- OS (e.g. from `/etc/os-release`): kind

brito-rafa:
Signed-off-by: Rafael Brito <rbrito@vmware.com>

closes #2342

### DIFF
--- a/pkg/apis/velero/v1/labels_annotations.go
+++ b/pkg/apis/velero/v1/labels_annotations.go
@@ -50,4 +50,16 @@ const (
 	// ResticVolumeNamespaceLabel is the label key used to identify which
 	// namespace a restic repository stores pod volume backups for.
 	ResticVolumeNamespaceLabel = "velero.io/volume-namespace"
+
+	// SourceClusterK8sVersionLabel is the label key used to identify the k8s
+	// git version of the backup , i.e. v1.16.4
+	SourceClusterK8sGitVersionLabel = "velero.io/source-cluster-k8s-gitversion"
+
+	// SourceClusterK8sMajorVersionLabel is the label key used to identify the k8s
+	// major version of the backup , i.e. 1
+	SourceClusterK8sMajorVersionLabel = "velero.io/source-cluster-k8s-major-version"
+
+	// SourceClusterK8sMajorVersionLabel is the label key used to identify the k8s
+	// minor version of the backup , i.e. 16
+	SourceClusterK8sMinorVersionLabel = "velero.io/source-cluster-k8s-minor-version"
 )
--- a/pkg/controller/backup_controller_test.go
+++ b/pkg/controller/backup_controller_test.go
@@ -33,9 +33,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/clock"
 
+	"k8s.io/apimachinery/pkg/version"
+
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	pkgbackup "github.com/vmware-tanzu/velero/pkg/backup"
 	"github.com/vmware-tanzu/velero/pkg/builder"
+	"github.com/vmware-tanzu/velero/pkg/discovery"
 	"github.com/vmware-tanzu/velero/pkg/generated/clientset/versioned/fake"
 	informers "github.com/vmware-tanzu/velero/pkg/generated/informers/externalversions"
 	"github.com/vmware-tanzu/velero/pkg/metrics"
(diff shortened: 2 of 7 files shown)


changelogs/unreleased/2342-brito-rafa
```diff
@@ -0,0 +1 @@
+adding labels on backup CRD for k8s major, minor and git versions
```
skriss: can you rename this file to `2346-brito-rafa`? We use the PR # in the file name.

pkg/plugin/generated/RestoreItemAction.pb.go
```diff
@@ -24,10 +24,12 @@ type RestoreItemActionExecuteRequest struct {
 	ItemFromBackup []byte `protobuf:"bytes,4,opt,name=itemFromBackup,proto3" json:"itemFromBackup,omitempty"`
 }
 
-func (m *RestoreItemActionExecuteRequest) Reset()                    { *m = RestoreItemActionExecuteRequest{} }
-func (m *RestoreItemActionExecuteRequest) String() string            { return proto.CompactTextString(m) }
-func (*RestoreItemActionExecuteRequest) ProtoMessage()               {}
-func (*RestoreItemActionExecuteRequest) Descriptor() ([]byte, []int) { return fileDescriptor3, []int{0} }
+func (m *RestoreItemActionExecuteRequest) Reset()         { *m = RestoreItemActionExecuteRequest{} }
```
skriss: I'm not sure why this file changed (likely as part of running `make update`, and it's just whitespace changes as far as i can tell), but since it's not relevant to this PR, can you revert the change to it? Will just need to ensure `make ci` still passes after reverting.

pkg/controller/backup_controller_test.go
```diff
@@ -159,6 +161,15 @@ func TestProcessBackupValidationFailures(t *testing.T) {
 		},
 	}
 
+	// Set up all of our fakes outside the test loop
```
skriss: I would move the instantiation of this stuff inside the test loop, alongside the other instantiations. This is generally our pattern since each iteration of the test loop is its own test case and we want to ensure all of the objects are freshly instantiated for each case.

As part of that, you probably want to drop `t.Helper()`, and you can use the logger that's already being instantiated inside the loop.

This comment applies for all of the test functions in this file.

pkg/controller/backup_controller_test.go
```diff
@@ -340,7 +358,10 @@ func TestProcessBackupCompletions(t *testing.T) {
 					Namespace: velerov1api.DefaultNamespace,
 					Name:      "backup-1",
 					Labels: map[string]string{
-						"velero.io/storage-location": "loc-1",
+						"velero.io/source-cluster-k8s-major-version": "",
+						"velero.io/source-cluster-k8s-minor-version": "",
+						"velero.io/source-cluster-k8s-version":       "v0.0.0-master+$Format:%h$",
```
skriss: I won't block on this, but it would be nice to replace these magic version strings. It looks like, when you instantiate the `apiServer`, you could do something like:

```
apiServer.DiscoveryClient.FakedServerVersion = &version.Info{
 // initialize the major/minor/git version fields
}
```

And then you could adjust the expected outputs to have the corresponding values.

brito-rafa: Thanks, Steve. I will add this suggestion when working on the labels during the restore phase.

pkg/apis/velero/v1/labels_annotations.go
```diff
@@ -50,4 +50,16 @@ const (
 	// ResticVolumeNamespaceLabel is the label key used to identify which
 	// namespace a restic repository stores pod volume backups for.
 	ResticVolumeNamespaceLabel = "velero.io/volume-namespace"
+
+	// SourceClusterK8sVersionLabel is the label key used to identify the k8s
+	// git version of the backup , i.e. v1.16.4
+	SourceClusterK8sVersionLabel = "velero.io/source-cluster-k8s-version"
```
carlisia: What do you think of renaming this to `SourceClusterK8sGitVersionLabel`? I think it would disambiguate what specific version this is for for those who don't recall the precise name of the 3 variables.

brito-rafa: Yes, sure,  I can do that, will tweak the code and amend the commit.

brito-rafa: I added the suggestions on the last commit.

pkg/controller/backup_controller_test.go
brito-rafa: I added the suggestions on the last commit.

PR #2350, opened 2020-03-17
allow plugins/ as a valid directory in the BSL bucket

### LINKED ISSUE #2344: allow plugins to write arbitrary data to 'plugins/' directory inside backup storage location
Some plugins would like to be able to directly write additional data to the object storage bucket used by Velero. However, there's not necessarily a natural place to put this data today.

Allow a `plugins/` top-level directory within the Velero backup storage location. Plugins can then write data to subdirectories within this directory.

**Environment:**

- Velero version (use `velero version`): v1.3.x
- Kubernetes version (use `kubectl version`): n/a
- Kubernetes installer & version: n/a
- Cloud provider or hardware configuration: n/a
- OS (e.g. from `/etc/os-release`): n/a

skriss:

closes #2344

### DIFF
--- a/changelogs/unreleased/2350-skriss
+++ b/changelogs/unreleased/2350-skriss
@@ -1,0 +1,1 @@
+Allow `plugins/` as a valid top-level directory within backup storage locations. This directory is a place for plugin authors to store arbitrary data as needed. It is recommended to create an additional subdirectory under `plugins/` specifically for your plugin, e.g. `plugins/my-plugin-data/`.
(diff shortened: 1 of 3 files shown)


skriss:
changelog added

skriss:
Oh, I guess the other part of this is, are we comfortable releasing this in a `v1.3.2` so this directory can be used soon? I think it's fine, but we can hold off for a little bit and see if there are any other bug fixes to package up.

nrb:
> Oh, I guess the other part of this is, are we comfortable releasing this in a v1.3.2 so this directory can be used soon?

I'm ok with that, though I though we should accumulate a few more changes since it's not really a crucial fix.

Issue #2332, opened 2020-03-06
[Custom CA] Add support to specify a custom CA secret ref on a BSL


dymurray:
**Describe the problem/challenge you have**
For the custom CA design, every declaration of object storage (via creation of a BackupStorageLocation) should give the user the ability to also specify a custom CA cert that will be associated and used with the corresponding object storage endpoint for adding, deleting, and reading data from object storage.

**Describe the solution you'd like**
1. Update BSL spec to allow for a secret reference which contains the CA bundle for object storage
2. Update restic controller code to inject CA bundle into the restic pod prior to command line invocation
3. Update Restic command invocation to pass the path of the CA bundle as a flag `--cacert`

mansam:
I'm going to take this on, please assign the issue to me.

PR #2353, opened 2020-03-17
Support setting a custom CA bundle to use with a BackupStorageLocation

mansam:
Fixes https://github.com/vmware-tanzu/velero/issues/2332

Restic and the storage provider plugins both need to be able to get at the CA cert, so it seems like it makes sense to put it on the ObjectStore and pass it through like is done for the bucket. #2332 calls for an object reference to a secret containing the CA bundle, but actually using a secret reference instead of a simple string complicates passing stuff to the plugins. @dymurray considering that the CA bundle is a public cert anyway, do you think that making it a secret is necessary?


skriss:
thanks for this @mansam! Code LGTM. Have you been able to test?

One request before approving - can you add a changelog file? (`changelogs/unreleased/2353-mansam`)

mansam:
Hi @skriss, I've been able to test it on my own cluster and things seem to be working correctly. Added a changelog as well. :)

pkg/restic/common_test.go
```diff
@@ -375,3 +377,42 @@ func TestTempCredentialsFile(t *testing.T) {
...
+	assert.Error(t, err)
+
+	// now add bsl to lister
+	require.NoError(t, bslInformer.GetStore().Add(bsl))
+
+	// bsl in lister: expect temp file to be created with cacert value
+	caCert, err = GetCACert(bslLister, "velero", "default")
+	require.NoError(t, err)
+
+	fileName, err := TempCACertFile(caCert, "default", fs)
```
ashish-amarnath: missing a `os.Remove(fileName)` to avoid creating temp files on dev machines running this test?

Issue #2330, opened 2020-03-06
[Custom CA] Add support for Velero client to specify a custom CA bundle


dymurray:
**Describe the problem/challenge you have**
There needs to be a mechanism with the velero client to configure the path to a custom CA cert when communicating with object storage. To do this, we will need to add a new flag `--cacert` to all Velero client calls, as well as give the ability for a user to specify a custom CA path in `velero client config`

**Describe the solution you'd like**
1. Add a new flag `--cacert` to `velero logs` and `velero describe` calls
1. Add ability to configure a default path to use as a custom CA cert which will automatically be used in `velero logs` and `velero describe` calls.

PR #2364, opened 2020-03-25
Add --cacert flag to velero cli commands

mansam:
Fixes https://github.com/vmware-tanzu/velero/issues/2330

Adds a --cacert flag to the log and describe commands
that takes a path to a PEM-encoded certificate bundle
as an alternative to --insecure-skip-tls-verify for
dealing with self-signed certificates.

Signed-off-by: Sam Lucidi <slucidi@redhat.com>

### DIFF
--- a/pkg/cmd/util/downloadrequest/downloadrequest.go
+++ b/pkg/cmd/util/downloadrequest/downloadrequest.go
@@ -39,7 +39,7 @@ import (
 // not found
 var ErrNotFound = errors.New("file not found")
 
-func Stream(client velerov1client.DownloadRequestsGetter, namespace, name string, kind v1.DownloadTargetKind, w io.Writer, timeout time.Duration, insecureSkipTLSVerify bool, caCertPath string) error {
+func Stream(client velerov1client.DownloadRequestsGetter, namespace, name string, kind v1.DownloadTargetKind, w io.Writer, timeout time.Duration, insecureSkipTLSVerify bool, caCertFile string) error {
 	req := &v1.DownloadRequest{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: namespace,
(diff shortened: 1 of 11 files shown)


nrb:
Could we have documentation for the commands added, too?

pkg/cmd/util/downloadrequest/downloadrequest.go
```diff
@@ -99,11 +99,27 @@ Loop:
...
+		// if it's available, otherwise create a new pool just
+		// for this.
+		caPool, err = x509.SystemCertPool()
+		if err != nil {
+			caPool = x509.NewCertPool()
 		}
+		caPool.AppendCertsFromPEM(caCert)
+	}
+	httpClient := new(http.Client)
+	httpClient.Transport = &http.Transport{
```
nrb: I think we probably want to use the timeout passed to `Stream` on the transport, since the Go HTTP client doesn't have one by default and will just wait forever if it's not specified (https://golang.org/src/net/http/transport.go#L211).

mansam: Seems like the right call. While fixing that I also set the rest of the values on the transport to match go's DefaultTransport. https://golang.org/src/net/http/transport.go#L42

mansam:
Happy to add documentation for the new flags though I don't see a good place to do so. @nrb where would you recommend I add that?

carlisia:
> Happy to add documentation for the new flags though I don't see a good place to do so. @nrb where would you recommend I add that?

Hm yeah, that is a very good question. My preference is it should go under "Use", much like the page on "Run in any namespace". A small text on the reason why  it's neeced/what it solves, what commands to use it with, and the documentation for how to use it (basically the documentation for the cmd itself, name of cmd + file name/path).

@nrb what is your preference?

nrb:
🤔 Yeah, this question's always the tough one to answer.

Given your team is mostly concerned with restic integration @mansam, I was thinking https://velero.io/docs/v1.3.1/restic/, but this is useful beyond restic.

I think @carlisia's suggestion for an entry under the **Use** heading is good, too. Maybe a new page documenting using the entirety of the feature set end-to-end called "Custom Certificate Bundles"? If it's a brand new page, I'm ok with that documentation being in it's own PR.

skriss:
@nrb @carlisia sounds like this is OK to merge as-is with a followup PR adding a new docs page?

carlisia:
Yes, this lgtm!

Issue #2251, opened 2020-02-06
Option to choose another GVK besides the preferred-version


gildub:
**Describe the problem/challenge you have**
A use case is when exporting a resource from a version of K8s being older than the version of the cluster restoring that resource into.

Let's say the backup cluster has api-server supporting:
* "some-api-group/v1" <- preferred-version
* "some-api-group/v2beta1"

And the restore side has:
* "some-api-group/v2"
* "some-api-group/v2beta1"

At the moment this is not possible because the preferred version would export the resource represented in v1 and the api-server on the restore side doesn't support it.

**Describe the solution you'd like**
Instead of having the version automatically selected from the preferred-version, 
being able to choose another API-Group/version (of course only if that GVK is supported by the api-server), would allow to export the involved resource using "some-api-group/v2beta1" and make the import possible into the destination.

**Anything else you would like to add:**
Overall the idea is to help creating GKVs export/import paths when 2 clusters are apart version-wise.
And the more 2 clusters are away from each others the higher the GVK gap, and this is where such option could really help. And even if a destination doesn't support the exported version, being able to bring it closer to a supported version would help too.

**Environment:**
- Velero version (use `velero version`): 1.2.0+
- Kubernetes version (use `kubectl version`): 1.7+
- Kubernetes installer & version: any
- Cloud provider or hardware configuration: any
- OS (e.g. from `/etc/os-release`): any

skriss:
@gildub what if Velero backed up *all* versions of a resource, and then intelligently restored the appropriate version based on which versions it found in the target cluster? Would that meet your need without requiring you to provide the mapping?

gildub:
@skriss, that would be ideal!
Now I'm not sure how you're going to do that magic unless using a similar approach as `oc convert`.

skriss:
> Now I'm not sure how you're going to do that magic unless using a similar approach as oc convert.

Well, I think for starters we wouldn't do anything too magical. In the example you gave, we'd back up both `v1` and `v2beta1` representations of the resources. Then, on restore, we'd see that the target cluster supported `v2beta1` and `v2`. Since the only overlap is `v2beta1`, we'd restore that version of the resource. So, no conversion required by Velero.

The next step would be to support conversion when there is no overlapping version between the backup and the target cluster. That *would* be pretty magical..

skriss:
xref #2310

brito-rafa:
[...]

PR #2373, opened 2020-03-27
Backup all groups and versions with backward compatibility

brito-rafa:
This is to address #2251.

// Update Apr-2020: this change will not need a feature parameter. It is backward compatible with the previous version of the backup.

There are two fundamental changes in this PR:

1. The use of ServerGroupsAndResources() to collect all resources for backup (instead of ServerPreferredResources()).
2. Adding the version of the item as part of object filepath

For backward compatibility, the preferred version of the item is discovered by ResourceFor and it is backed up as usual item filepath. I ran a comparison of items created by the new code and old code, and all preferred versions match (on k8s 1.16, 1.17). See [testing script](https://github.com/brito-rafa/velero-test/blob/master/2251-dev-test.sh).
that I ran restores using the new code backup without issues.


pkg/backup/item_backupper.go
```diff
@@ -253,11 +254,23 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
...
 	}
 
+	// group version of this object
+	// Used on filepath if backing up all groups and versions
+	version := resourceVersion(obj)
+
 	var filePath string
 	if namespace != "" {
-		filePath = filepath.Join(api.ResourcesDir, groupResource.String(), api.NamespaceScopedDir, namespace, name+".json")
+		if features.IsEnabled("Migration") {
```
nrb: If the backup file format is backwards compatible on restore (meaning Velero v1.3.1 could restore a backup made with this code change), then I don't think the feature flag would be necessary anywhere in this PR.

brito-rafa: ok, sure. I will change the code accordingly and provide the testing of backward compatibility. And yes, I will provide documentation on the file format.

nrb:
To follow up from the community meeting:

In the _new_ directories, mark the preferred version somehow, either with an empty file named `preferred` in it, or by putting `preferred` in the directory name. Putting it in the directory name may require some munging if we encounter a cluster where we need to do some version negotiation, but we could probably do string matching to drop it.

We should leave the `<resourcename>/namespace` directories alone to maintain the backwards compat.

brito-rafa:
I added the suffix "-preferredversion" on the API Group preferred sub-directory name. Please let me know if it works. Added documentation and tests cases on the code.  Additionally, I have tested that classic directory structure is backward compatible using the following [script](https://github.com/brito-rafa/velero-test/blob/master/2251-dev-test.sh) with the following [result](https://github.com/brito-rafa/velero-test/blob/master/testing-2251.log).

pkg/apis/velero/v1/constants.go
```diff
@@ -36,4 +36,8 @@ const (
 	// NamespaceScopedDir is the name of the directory containing namespace-scoped
 	// resource within a Velero backup.
 	NamespaceScopedDir = "namespaces"
+
+	// PreferredVersionDir is the sufix name of the directory containing the preferred version of the API group
```
nrb, line 40: Thanks for this; it's helpful if we ever need to do string matching to trim it off.

pkg/backup/item_backupper.go
```diff
@@ -253,11 +253,34 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 		return false, kubeerrs.NewAggregate(backupErrs)
 	}
 
+	// group version of this object
+	// Used on filepath to backup up all groups and versions
+	version := resourceVersion(obj)
+
+	// Getting the preferred group version of this resource
+	gvr_preferred, _, err := ib.discoveryHelper.ResourceFor(groupResource.WithVersion(""))
```
nrb, line 261: To clarify my understanding of this call - since we're invoking `ResourceFor` without any version, we'll get back the preferred GVK, and thus we can use the version on that GVK as preferred.

If we invoked `ResourceFor` with a specific version, we'd get the GVK for that specific version, assuming it existed.

And the reason `ResourceFor` is used here is because the helper for `ServerPreferredResources` gets us a list of preferred versions for _all_ API groups, not just the one we're processing at the moment.

Is that accurate?

pkg/backup/item_backupper.go
```diff
@@ -253,11 +253,34 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
...
+	}
+	versionpreferred := gvr_preferred.Version
+
 	var filePath string
-	if namespace != "" {
-		filePath = filepath.Join(api.ResourcesDir, groupResource.String(), api.NamespaceScopedDir, namespace, name+".json")
+
+	// API Group version is now part of path of backup as a subdirectory
+	// it will add a prefix to subdirectory name for the preferred version
+	if version == versionpreferred {
```
nrb, line 271: What if we re-wrote this slightly, to reduce duplicating the branching on filepath?

Something like

```go
versionPath := version
if version == versionpreferred {
versionPath = version+api.PreferredVersionDir
}
		if namespace != "" {
			filePath = filepath.Join(api.ResourcesDir, groupResource.String(), versionPath, api.NamespaceScopedDir, namespace, name+".json")
		} else {
			filePath = filepath.Join(api.ResourcesDir, groupResource.String(), versionPath, api.ClusterScopedDir, name+".json")
		}
```

That way we can keep 2 assignments for `filePath` like before, and still manipulate the form of the version directory.

site/docs/master/output-file-format.md
```diff
@@ -67,33 +67,96 @@ resources/
...
+                namespace1/
+                    hpa-to-the-rescue.json
+                    ...
+                namespace2/
+                    ...
+
     ...
 ```
+
+Recently, we have added all API groups versions part of the backup (previously, only preferred version of each API Groups was backed up). Each resource has one or more sub-directories, one sub-directory for each supported version of the API group. The preferred version API Group of each resource has the suffix "-preferredversion" as part of the sub-directory name. For backward compatibility, we kept the classic directory structure without API Group version, which sits on the same level of the API Group sub-directory versions.
```
nrb, line 162: @skriss @carlisia @ashish-amarnath I'm curious if you all think we should bump our format to v1.1 with this change or not. It's backwards compatible, but this would signal that there's a change, and be more explicit that there's new features. It would also mean that this paragraph could change from `Recently`, which is context-sensitive, to `In backup format v1.1, we have added...`, which is not.

pkg/backup/item_backupper.go
brito-rafa: I added the piece of code as suggested with an amended commit. I ran the regular tests using a slightest enhanced version of the [script](https://github.com/brito-rafa/velero-test/blob/master/2251-dev-test.sh) with the following [output ](https://github.com/brito-rafa/velero-test/blob/master/testing-2251-20200405.log). The test was taking a backup with the patched version on a 1.16 cluster and restoring using standard Velero 1.3.1 on a K8s 1.18.

pkg/backup/item_backupper.go
brito-rafa: Flagging as resolved.

pkg/backup/item_backupper.go
brito-rafa: Flagging as resolved, feel free to add more.

pkg/backup/item_backupper.go
brito-rafa: Yes, your understanding is correct.

site/docs/master/output-file-format.md
carlisia: That's a good point. It makes sense to me. But I'm also not clear on what should trigger a bump, maybe we could take a moment to discuss.

nrb: @skriss @ashish-amarnath What are your thoughts on this?

nrb: I'd like to get this resolved, so I've added this to the April 14 2020 community meeting discussion topics. IMO, this is the only thing block merge at this point.

skriss:
@brito-rafa looks like this needs a rebase to fix the conflicts.

site/docs/master/output-file-format.md
skriss: Discussed and resolved - we'll do a minor version bump.

brito-rafa:
Rebased and resolved the conflicts. Added a new section on the output-file-format.md file with the example of backup file format 1.1.

site/docs/master/output-file-format.md
```diff
@@ -111,3 +172,107 @@ A major version would indicate that a version of Velero older than the version t
 Major versions of the file format will be incremented with major version releases of Velero.
 However, a major version release of Velero does not necessarily mean that the backup format version changed - Velero 3.0 could still use backup file format 2.0, as an example.
 
+## file format version: 1.1
+
+In version 1.1, we have added all API groups versions part of the backup (previously, only preferred version of each API Groups was backed up). Each resource has one or more sub-directories, one sub-directory for each supported version of the API group. The preferred version API Group of each resource has the suffix "-preferredversion" as part of the sub-directory name. For backward compatibility, we kept the classic directory structure without API Group version, which sits on the same level of the API Group sub-directory versions.
```
nrb, line 177: A couple minor changes here, bold mine:

>In version 1.1, we have added all API groups versions **as** part of the backup (previously, only **the** preferred version of each API Groups was backed up). Each resource has one sub-directory for each supported version of the API group. The preferred version API Group of each resource has the suffix "-preferredversion" as part of the sub-directory name. For backward compatibility, we kept the classic directory structure without **the** API Group version, which sits on the same level **as** the API Group sub-directory versions.

site/docs/master/output-file-format.md
```diff
@@ -111,3 +172,107 @@ A major version would indicate that a version of Velero older than the version t
 Major versions of the file format will be incremented with major version releases of Velero.
 However, a major version release of Velero does not necessarily mean that the backup format version changed - Velero 3.0 could still use backup file format 2.0, as an example.
 
+## file format version: 1.1
```
nrb, line 175: Capitalize each word here

brito-rafa:
There is a small issue updating constant [BackupVersion](https://github.com/vmware-tanzu/velero/blob/master/pkg/backup/backup.go#L44) from current value of int(1) to string(1.1) or float(1.1).

Per backup schema, Version is an [int](https://github.com/vmware-tanzu/velero/blob/master/pkg/apis/velero/v1/backup.go#L217).

I can go ahead change the schema and Version to be a string but I wonder if there are further consequences.

Another alternative is to consider the new backup format as version int(2) and then, of course, the [Semantic Versioning](http://semver.org/) will no longer apply.

 Please let me know your thoughts...

pkg/backup/item_backupper.go
```diff
@@ -253,11 +253,31 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 		return false, kubeerrs.NewAggregate(backupErrs)
 	}
 
+	// group version of this object
+	// Used on filepath to backup up all groups and versions
+	version := resourceVersion(obj)
+
+	// Getting the preferred group version of this resource
+	gvr_preferred, _, err := ib.discoveryHelper.ResourceFor(groupResource.WithVersion(""))
```
skriss, line 261: nit: `gvr_preferred` -> `preferredGVR`

pkg/backup/item_backupper.go
```diff
@@ -253,11 +253,31 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
...
+	if err != nil {
+		return false, errors.WithStack(err)
+	}
+	versionpreferred := gvr_preferred.Version
```
skriss, line 265: nit: `versionpreferred` -> `preferredVersion`

pkg/backup/item_backupper.go
```diff
@@ -281,6 +301,33 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
...
+	// backing up the preferred version backup without API Group version on path -  this is for backward compability
+
+	if version == versionpreferred {
+		if namespace != "" {
+			filePath = filepath.Join(api.ResourcesDir, groupResource.String(), api.NamespaceScopedDir, namespace, name+".json")
+		} else {
+			filePath = filepath.Join(api.ResourcesDir, groupResource.String(), api.ClusterScopedDir, name+".json")
+		}
+
+		hdrpreferred := &tar.Header{
```
skriss, line 313: nit: i'd just reuse `hdr` here (`hdr = &tar.Header{ ...`)

pkg/backup/item_backupper.go
```diff
@@ -253,11 +253,31 @@ func (ib *defaultItemBackupper) backupItem(logger logrus.FieldLogger, obj runtim
 		return false, kubeerrs.NewAggregate(backupErrs)
 	}
 
+	// group version of this object
+	// Used on filepath to backup up all groups and versions
+	version := resourceVersion(obj)
+
+	// Getting the preferred group version of this resource
```
skriss, line 260: we might want to consider moving the getting of the preferred version to an outer loop (e.g. in `resource_backupper.go`) - there's no need to do this once for every item, we could just do it once per resource type. I'm not sure if there's a material performance impact or not.

site/docs/master/output-file-format.md
```diff
@@ -57,7 +57,7 @@ rootBucket/
 ```
 Note that this file includes detailed info about your volume snapshots in the `status.volumeBackups` field, which can be helpful if you want to manually check them in your cloud provider GUI.
 
-## file format version: 1
+## File Format Version: 1
```
skriss, line 60: we'll want to bump the version to `1.1.0`.

skriss: Reviewing the rest, I see there's a new section that now describes the 1.1 format, so actually, I wouldn't expect this section to change at all - in other words, you'll want to undo the changes to the layout here.

Perhaps we can rearrange this page a bit, so something like the following:

- (summary)
- Example backup JSON file
- Output file format versioning
- Versions
    - Version 1.1 (current)
    - Version 1.0

skriss:
OK, just seeing your comment https://github.com/vmware-tanzu/velero/pull/2373#issuecomment-617441597. Processing..

brito-rafa:
Hi @skriss - many thanks for the reviews - I will gladly incorporate them. Thinking to do all changes in one shot once we figure the direction for the constant BackupVersion .

skriss:
sounds good, I'll try to add another comment today with my thoughts on that

nrb:
@brito-rafa Good catch on the data type. I'm sure there are consequences to changing it to a string or a float that will cause serialization issues that I don't know yet. I'll look over and consider your proposed options today.

skriss:
I don't think we should change the data type of the existing API field, as that would be a breaking change. I do think using semantic versioning is the way to go, since it clearly communicates what different kinds of version bumps mean.

I think the cleanest change would probably be to:
- add a new status field that has the proper type (string)
- update the documentation on the existing field to indicate that it stores the major version of the backup format only
- probably deprecate the existing field (by adding a comment) and plan to remove it for 2.0

@nrb would appreciate your thoughts on this.

nrb:
Yeah, I agree with adding a new field with a string data type and moving the old field to deprecated.

I would propose we name the new field `FormatVersion`, then we can update the constant to be a string.

Does that sound like a plan to you both @skriss and @brito-rafa?

skriss:
:+1: I think that's our best path.

skriss:
@brito-rafa let me know if you need input on how to get these changes done; we're starting to approach the end of the dev cycle for this release and I do have at least one other PR that is waiting for this one to merge, so let's try to get this in soon!

brito-rafa:
Thanks - I will put cycles on this PR today & tomorrow to make sure the release cycle is under ETA.

Just one clarification: the new field `FormatVersion` is to be added on (BackupStatus)[https://github.com/vmware-tanzu/velero/blob/master/pkg/apis/velero/v1/backup.go#L214] struct as optional. And backup API continues to be v1 - correct?

nrb:
>Just one clarification: the new field FormatVersion is to be added on BackupStatus struct as optional. And backup API continues to be v1 - correct?

Yes - `BackupStatus.FormatVersion` was what I was thinking, as well as making it option.

Yeah, I would say the API stays on v1 for now, since the field is optional. I can see an argument for it being a minor revision bump too, but I'm find with drawing this line.

Also let me know if you need help with anything - I have a feeling we'll lean on this work for the CRD fixes, though I haven't had time to test an alternate approach since I was off late last week. I'll make sure to prioritize this PR until it's merged.

brito-rafa:
I have added all the requested changes. Please let me know if you want me to add any other change.

skriss:
:+1: thanks for the updates @brito-rafa! I took a quick glance and nothing stuck out, but I'll look again in detail shortly.

brito-rafa:
Sure, Steve. I am open to contributing to any logic that might suit it.

Perhaps could we could consider using a feature parameter to dictate the backup format? Just a thought.

In meanwhile, I will go ahead and fix the case and descriptions as you pointed.

pkg/backup/backup.go
```diff
@@ -284,7 +288,7 @@ func (kb *kubernetesBackupper) Backup(log logrus.FieldLogger, backupRequest *Req
 
 func (kb *kubernetesBackupper) writeBackupVersion(tw *tar.Writer) error {
 	versionFile := filepath.Join(api.MetadataDir, "version")
-	versionString := fmt.Sprintf("%d\n", BackupVersion)
+	versionString := fmt.Sprintf("%v\n", BackupFormatVersion)
```
nrb, line 291: Minor, but I'd change this to `%s` instead of `%v` to be sure it's a string instead of any Go value.

nrb:
@skriss Perhaps Velero should always operate on the preferred version on backup? The preferred version gets passed to item action plugins, restic operates on the preferred version of pods, snapshots are taken on preferred version of PVs, at least for the 1.x series of Velero.

Some newer plugins could use all versions, but older ones will retain the older model for now.

Then, to expand the capabilities, the APIs are expanded to take all versions, allowing more complex decisions. throughout the pipeline in a major Velero release.

Or does this the intent altogether, since on restore the we could be restoring from any version, which may have data that wasn't included in the preferred version?

brito-rafa:
Another option is having a flag (or feature) to dictate the behavior of the discovery client just to "discover" preferred versions (via ServerPreferredResources(), as used by Velero 1.3.X today) and continue to roll out the 1.1.0 backup format version. 
This PR is using ServerGroupsAndResources().

skriss:
> Another option is having a flag (or feature) to dictate the behavior of the discovery client just to "discover" preferred versions (via ServerPreferredResources(), as used by Velero 1.3.X today) and continue to roll out the 1.1.0 backup format version.

Yeah, I think that would be workable - so if the feature flag wasn't on, the discovery helper would only get the preferred version of resources; backups would still write 2 copies of items to the tarball, one in the existing format and one with the new directory structure. That also keeps the use of the feature flag simple, we don't have to sprinkle checks for it all over the code -- it'd just be in the discovery helper.

brito-rafa:
I added a feature flag called `EnableAPIGroupVersions` to enable the handling of multiple API Group versions.
I use this [file example](https://github.com/brito-rafa/velero-test/blob/master/velero-allversions-patch.yaml) to turn on this feature, downloading and running the following:
```
kubectl patch deployment velero --patch "$(cat velero-allversions-patch.yaml)" -n velero
```

pkg/discovery/helper.go
```diff
@@ -29,6 +29,9 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/restmapper"
 
+	api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
skriss, line 32: our standard is to alias this import as `velerov1api`, rather than just `api`.

pkg/discovery/helper.go
```diff
@@ -29,6 +29,9 @@ import (
...
+	"github.com/vmware-tanzu/velero/pkg/features"
+
```
skriss, line 34: remove blank line

pkg/discovery/helper.go
```diff
@@ -117,14 +119,21 @@ func (h *helper) Refresh() error {
 		return errors.WithStack(err)
 	}
 
+	preferredResources, err := refreshServerPreferredResources(h.discoveryClient, h.logger)
```
skriss: I would only execute this if the feature flag is **not** set, since this call is only necessary in that case -- so you could put it in an `else` branch of the if statement on L129.

pkg/discovery/helper.go
brito-rafa: Done!

pkg/discovery/helper.go
brito-rafa: Done!

pkg/discovery/helper.go
brito-rafa: Done!

skriss:
:+1: updates look good! I'm gonna take one last look through, but I don't anticipate anything else.

carlisia:
I'll give this a review first thing tomorrow but I think @nrb should be sure to review it too before merging!

skriss:
:tada:

PR #2377, opened 2020-03-30
add restore item action to update PVC selected-node  annotation

mynktl:
**This change solve below issue**

If backed-up persistentvolumeclaims resource has an annotation of selected-node(`volume.kubernetes.io/selected-node`) and the given node doesn't exist in destination cluster, where the restore is triggered, then restore will stuck `InProgress` state.

**Approach**
This restoreItemAction works for `persistentvolumeclaims` resource having an annotation `volume.kubernetes.io/selected-node`.

If node-mapping for the given node exists in configMap(in velero installation namespace, with label selector `velero.io/plugin-config,velero.io/change-pvc-node=RestoreItemAction`), then the plugin will update the PVC annotation with new node and return the updated resource.

If node-mapping for the given node doesn't exist, then the plugin will check for the given node existence. If the given node doesn't exist in the cluster, then the plugin will delete the annotation for selected-node and return the updated `persistentvolumeclaims` resource.

### DIFF
--- a/pkg/restore/change_pvc_node_selector.go
+++ b/pkg/restore/change_pvc_node_selector.go
@@ -19,12 +19,13 @@ package restore
 import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"github.com/vmware-tanzu/velero/pkg/plugin/framework"
-	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/vmware-tanzu/velero/pkg/plugin/framework"
+	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
 )
 
 // ChangePVCNodeSelectorAction updates/reset PVC's node selector
--- a/pkg/restore/change_pvc_node_selector.go
+++ b/pkg/restore/change_pvc_node_selector.go
@@ -1,0 +1,155 @@
+/*
+Copyright 2020 the Velero contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


carlisia:
For the core team: This seems the restore side of the same logical issue we are discussing on the backup side in this ticker: https://github.com/vmware-tanzu/velero/issues/1878.

pkg/restore/change_pvc_node_selector.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// configMap doesn't have node-mapping
+	// Let's check if node exists or not
+	exists, err := isNodeExist(p.nodeClient, node)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error checking node %s existence", node)
+	}
+
+	if !exists {
+		log.Infof("Resetting selected-node")
```
skriss, line 117: Maybe "Clearing selected-node because node named %s does not exist" would be a little clearer

pkg/restore/change_pvc_node_selector.go
```diff
@@ -0,0 +1,156 @@
...
+		return "", err
+	}
+
+	if config == nil || len(config.Data) == 0 {
+		// there is no node mapping defined for change-pvc-node
+		// so we will return empty new node
+		return "", nil
+	}
+
+	newNode, _ := config.Data[node]
```
skriss, line 142: nit: you could simplify L142-143 to just be `return config.Data[node], nil`

pkg/restore/change_pvc_node_selector.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func getNewNodeFromConfigMap(client corev1client.ConfigMapInterface, node string) (string, error) {
+	// fetch node mapping from configMap
+	config, err := getPluginConfig(framework.PluginKindRestoreItemAction, "velero.io/change-pvc-node", client)
+	if err != nil {
+		return "", err
+	}
+
+	if config == nil || len(config.Data) == 0 {
```
skriss, line 136: the check for `len(config.Data) == 0` is not really needed, since that case will be handled fine by subsequent lines. I'd probably remove it and just check for `config == nil` here.

mynktl:
Hi @skriss 
Thanks for the review comments.
I've updated the PR with the changes you suggested. PTAL.

 > - [x]  add a changelog (`changelogs/unreleased/2377-mynktl`) with a brief description of the change
> 
>  - [x]  update `site/docs/master/restore-reference.md` with info about this plugin - you can use the "Changing PV/PVC Storage Classes" section as a template

Thanks!

blackpiglet:
According to my test on k8s v1.22, without this plugin, the restore can still work. 
The logic in volume controller can remove the select-node annotation from PVC, if the node specified by annotation not found in the cluster. 

https://github.com/kubernetes/kubernetes/blob/release-1.16/pkg/controller/volume/persistentvolume/pv_controller.go#L1597-L1616

PR #2378, opened 2020-03-31
Making backup and restore controller worker counts configurable

uruddarraju:
Velero currently defaults all worker counts to be 1, which means each worker can only work on one sync event at any given point. This has caused a couple of issues in our environment that uses a lot of PVCs where the backup controller was taking a long time to snapshot PVC volumes with AWS, leading to a lot of other backups taking too long to be synced.

In our environment, the PVC backup took ~16 minutes, during which no other scheduled backups and on demand backups for deployments, configmaps, etc were synced leading to a state where the backups were in New phase for too long(16 minutes).

This change will allow operators to override the worker counts for these 2 control loops which are potentially more costly compared to other controllers in velero.

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -117,6 +117,8 @@ type serverConfig struct {
 	disabledControllers                                                     []string
 	clientQPS                                                               float32
 	clientBurst                                                             int
+	backUpControllerWorkerCount                                             int
+	restoreControllerWorkerCount                                            int
 	profilerAddress                                                         string
 	formatFlag                                                              *logging.FormatFlag
 	defaultResticMaintenanceFrequency                                       time.Duration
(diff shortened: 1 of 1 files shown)


skriss:
@uruddarraju thanks for the PR! The main concern I have here is that we're not necessarily synchronizing multiple concurrent backups properly. One specific area that I have in mind is hooks - where multiple concurrent backups may try to quiesce an application simultaneously - but it's possible there are other issues as well. How much testing have you been able to do with concurrent backups/restores, and have you seen any other issues?

skriss:
@uruddarraju just reaching out again to see if you could provide any insight on my question.

skriss:
Closing this out as inactive, but feel free to reach out again. Also keep an eye on #1653!

PR #2380, opened 2020-03-31
Only add CA cert to plugin config if it's defined

### LINKED ISSUE #2379: don't add `caCert` to object store config if it's empty
#2353 added support for a `caCert` field as part of BackupStorageLocations. https://github.com/vmware-tanzu/velero/blob/master/pkg/persistence/object_store.go#L117 adds it to the object store `Config` so it can be used during an ObjectStore plugin's `Init(..)` function.

I think we should *only* pass it if it's non-empty, because otherwise it will potentially break validation for community plugins that haven't updated their velero dependency to use the updated validation method that allows `caCert` as a config key.

nrb:
Fixes #2379 


nrb:
Tested this with v1.0.1 of the GCP plugin, and it allows the server to start up as expected.

PR #2382, opened 2020-04-01
POC: Add an HTTP endpoint for liveness checks

### LINKED ISSUE #1967: Add a BSL controller to handle validation + update BSL status phase
Currently, if Velero finds an invalid BSL, it will block operations. 

Change the behavior so if that Velero continues creating backups even if there are invalid BSLs, as long as there's a valid BSL for them. 

## HOW:
- [x] add a `runtime-controller` controller that continuously validates BSLs and marks them as `available` (valid) or `unavailable` (invalid)
- [x] change the server behavior to no longer validate BSLs

## Complimentary work
- Add a column to indicate the BSL status on `velero backup-location get` (https://github.com/vmware-tanzu/velero/issues/2489)
- Have the BSL controller continuously update the server `status` based on the state of the BSLs (#2488) 
- Add a cmd to fetch server status (https://github.com/vmware-tanzu/velero/issues/979)
---
### log a warning
- multiple BSLs, any one of them is `unavailable`, but at least one other is `available`
- multiple BSLs, all `available`, but no BSL matches the specified `default`

### log an error
   - no BSL, OR
   - existing BSLs, all `unavailable`

## Backup behavior
- no change

Original issue:
---
We're running into a scenario where if the Velero pod restarts for whatever reasons (edits to the Velero deployment, pod rescheduling etc) and there happens to be some backup storage locations that are unreachable due to the backup server being unavailable (upgrades etc), Velero pod keeps crashlooping due to these invalid backup storage locations. This prevents backups to other valid backup storage locations.

Velero pod should not crashloop if one BSL is invalid since this blocks backups to any other valid storage locations. 

* `kubectl logs deployment/velero -n velero`
* `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
* `velero backup logs <backupname>`
* `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
* `velero restore logs <restorename>`

**Environment:**

- Velero version (use `velero version`): v1.1.0

nrb:
Instead of halting the server on fixable problems, instead use a
liveness check endpoint that does the same checks, and will allow the
Kubernetes system to restart the pod if they fail.

This change means that if you start a `velero server` locally, with just a namespace and none of the CRDs defined, you can do `curl localhost:2000/livez` and get a 404 indicating there's a problem, which will be in the logs.

Once all resources are in acceptable shape, a 200 page response will come back, and Velero will resume.

Should all BackupStorageLocations be deleted, or all the CRDs get removed, the liveness check will start returning 404 again.

This will make Velero behave the same whether a BSL exists on startup, or if it gets removed after Velero is already running.

I think this may be a possible fix for #1967. @betta1, could you please take a look and maybe try it out? This change does not yet include a change to the YAML to use the liveness probe.

Fixes #1967 

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -307,18 +307,12 @@ func (s *server) run() error {
 		return err
 	}
 
-	if err := s.veleroResourcesExist(); err != nil {
-		return err
-	}
-
-	if err := s.validateBackupStorageLocations(); err != nil {
-		return err
-	}
-
-	if _, err := s.veleroClient.VeleroV1().BackupStorageLocations(s.namespace).Get(s.config.defaultBackupLocation, metav1.GetOptions{}); err != nil {
-		s.logger.WithError(errors.WithStack(err)).
-			Warnf("A backup storage location named %s has been specified for the server to use by default, but no corresponding backup storage location exists. Backups with a location not matching the default will need to explicitly specify an existing location", s.config.defaultBackupLocation)
-	}
+	// Anything that could be created after Velero has started will be
+	// queried as part of the liveness check
+	go func() {
+		s.logger.Info("Starting liveness server at port 2000")
+		s.runLiveness()
+	}()
 
 	if err := s.initRestic(); err != nil {
 		return err
@@ -380,7 +374,7 @@ func (s *server) veleroResourcesExist() error {
 	}
 
 	if veleroGroupVersion == nil {
-		return errors.Errorf("Velero API group %s not found. Apply examples/common/00-prereqs.yaml to create it.", api.SchemeGroupVersion)
+		return errors.Errorf("Velero API group %s not found. Create the CRDs via `velero install`, Helm, or applying YAML to proceed.", api.SchemeGroupVersion)
 	}
 
 	foundResources := sets.NewString()
@@ -399,7 +393,7 @@ func (s *server) veleroResourcesExist() error {
 	}
 
 	if len(errs) > 0 {
-		errs = append(errs, errors.New("Velero custom resources not found - apply examples/common/00-prereqs.yaml to update the custom resource definitions"))
+		errs = append(errs, errors.New("Velero custom resources not found.  Create the CRDs via `velero install`, Helm, or applying YAML to proceed."))
 		return kubeerrs.NewAggregate(errs)
 	}
 
@@ -830,5 +824,41 @@ func (s *server) runProfiler() {
 
 	if err := http.ListenAndServe(s.config.profilerAddress, mux); err != nil {
 		s.logger.WithError(errors.WithStack(err)).Error("error running profiler http server")
+	}
+}
+
+// runLiveness starts an HTTP server to handle liveness checks for the velero server.
+// Note that liveness checks should have generous timeouts, since Velero will query
+// object storage, which is possibly a network call across the internet.
+func (s *server) runLiveness() {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/livez", func(w http.ResponseWriter, r *http.Request) {
+		// Refresh the discovery helper, since it may be stale if the velero API group was
+		// created after the pod.
+		s.discoveryHelper.Refresh()
+		if err := s.veleroResourcesExist(); err != nil {
+			s.logger.Warn(err)
+			http.NotFound(w, r)
+			return
+		}
+
+		if err := s.validateBackupStorageLocations(); err != nil {
+			s.logger.Warn(err)
+			http.NotFound(w, r)
+			return
+		}
+		if _, err := s.veleroClient.VeleroV1().BackupStorageLocations(s.namespace).Get(s.config.defaultBackupLocation, metav1.GetOptions{}); err != nil {
+			s.logger.WithError(errors.WithStack(err)).
+				Warnf("A backup storage location named %s has been specified for the server to use by default, but no corresponding backup storage location exists. Backups with a location not matching the default will need to explicitly specify an existing location", s.config.defaultBackupLocation)
+			http.NotFound(w, r)
+			return
+		}
+
+		// Happy path, all prerequisites are present.
+		fmt.Fprintf(w, "Velero is live")
+	})
+
+	if err := http.ListenAndServe(":2000", mux); err != nil {
+		s.logger.WithError(errors.WithStack(err)).Error("error running liveness http server")
 	}
 }


nrb:
@dymurray I see you also had activity on #1967 recently

pkg/cmd/server/server.go
```diff
@@ -307,18 +307,12 @@ func (s *server) run() error {
...
-
-	if err := s.validateBackupStorageLocations(); err != nil {
-		return err
-	}
-
-	if _, err := s.veleroClient.VeleroV1().BackupStorageLocations(s.namespace).Get(s.config.defaultBackupLocation, metav1.GetOptions{}); err != nil {
-		s.logger.WithError(errors.WithStack(err)).
-			Warnf("A backup storage location named %s has been specified for the server to use by default, but no corresponding backup storage location exists. Backups with a location not matching the default will need to explicitly specify an existing location", s.config.defaultBackupLocation)
-	}
+	// Anything that could be created after Velero has started will be
```
nrb, line 310: I think it probably makes sense to repeat the checks I deleted here before the liveness endpoint starts, but maybe just log their errors rather than returning them and halting.

pkg/cmd/server/server.go
```diff
@@ -832,3 +826,39 @@ func (s *server) runProfiler() {
...
+
+// runLiveness starts an HTTP server to handle liveness checks for the velero server.
+// Note that liveness checks should have generous timeouts, since Velero will query
+// object storage, which is possibly a network call across the internet.
+func (s *server) runLiveness() {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/livez", func(w http.ResponseWriter, r *http.Request) {
+		// Refresh the discovery helper, since it may be stale if the velero API group was
+		// created after the pod.
+		s.discoveryHelper.Refresh()
```
nrb, line 838: I'm not sure the `Refresh` call belongs here - it may be better inside of `veleroResourcesExist`.

betta1:
@nrb sure, I'll try this sometime later today and post the update here.

carlisia:
Added an item to the community meeting tomorrow (https://hackmd.io/Jq6F5zqZR7S80CeDWUklkA?both#April-142020) to talk about this.

nrb:
@carlisia 👍 Please be sure to capture notes on the discussion!

pkg/cmd/server/server.go
skriss: somewhat related, we already have a goroutine that refreshes discovery every 5min -- so we'll end up refrehsing a lot. Not sure yet where it makes sense to have it, just food for thought right now.

nrb: Yeah; I added it because it wasn't picking up the CRDs even after I posted them. 5 minutes seems like a long time for a liveness check, but maybe we remove this refresh and change the other one to match whatever the check period is on the pod?

skriss:
+1 to a liveness probe in general, however I'm not yet clear on what things we want to check in it -- I'm not sure the current BSL validation that's in there makes sense/solves #1967.  Added some more comments at https://github.com/vmware-tanzu/velero/issues/1967#issuecomment-613678786.

nrb:
@skriss Agreed, I'll look at those scenarios. For this kind of POC, I just copied existing function calls into the liveness check. For the case where multiple BSLs exist, the default is valid and others aren't, the liveness check should pass.

It's very likely that the liveness check we ultimately use will actually build on top of the work @carlisia is doing, as this version is fairly naive and shouldn't be merged as-is.

pkg/cmd/server/server.go
```diff
@@ -307,18 +307,12 @@ func (s *server) run() error {
...
-	}
-
-	if _, err := s.veleroClient.VeleroV1().BackupStorageLocations(s.namespace).Get(s.config.defaultBackupLocation, metav1.GetOptions{}); err != nil {
-		s.logger.WithError(errors.WithStack(err)).
-			Warnf("A backup storage location named %s has been specified for the server to use by default, but no corresponding backup storage location exists. Backups with a location not matching the default will need to explicitly specify an existing location", s.config.defaultBackupLocation)
-	}
+	// Anything that could be created after Velero has started will be
+	// queried as part of the liveness check
+	go func() {
+		s.logger.Info("Starting liveness server at port 2000")
```
phuongatemc, line 313: It would be nice to have the port number "2000" either as a constant or configurable via variable.

pkg/cmd/server/server.go
```diff
@@ -832,3 +826,39 @@ func (s *server) runProfiler() {
...
+				Warnf("A backup storage location named %s has been specified for the server to use by default, but no corresponding backup storage location exists. Backups with a location not matching the default will need to explicitly specify an existing location", s.config.defaultBackupLocation)
+			http.NotFound(w, r)
+			return
+		}
+
+		// Happy path, all prerequisites are present.
+		fmt.Fprintf(w, "Velero is live")
+	})
+
+	if err := http.ListenAndServe(":2000", mux); err != nil {
```
phuongatemc, line 861: Same comment as line 313.

pkg/cmd/server/server.go
```diff
@@ -832,3 +826,39 @@ func (s *server) runProfiler() {
...
+		// Refresh the discovery helper, since it may be stale if the velero API group was
+		// created after the pod.
+		s.discoveryHelper.Refresh()
+		if err := s.veleroResourcesExist(); err != nil {
+			s.logger.Warn(err)
+			http.NotFound(w, r)
+			return
+		}
+
+		if err := s.validateBackupStorageLocations(); err != nil {
```
phuongatemc, line 845: Current logic of validateBackupStorageLocations would return error if there is any invalid BackupStorageLocation.  Should Velero continue function if there is at least 1 valid BackupStorageLocation?  If so, validateBackupStorageLocations should be enhanced to return more than just an error.  For example return number of valid BackupStorageLocation and at this point we decide Velero should continue work if some valid BackupStorageLocation present.

carlisia: This validation will change after https://github.com/vmware-tanzu/velero/pull/2617.

We will change the validation logic to only error if there are no valid BackupStorageLocations. Else it will be only warnings if there's any invalid BSL.

nrb:
I'm going to close this out, given #2674 merged. I don't think `CrashLoop`ing the Velero server accomplishes the goal, nor is the liveness endpoint the right place to expose this information at this point in time.

PR #2388, opened 2020-04-02
v1.3.2 changelog & docs

skriss:
Note that I'd like to include one additional bug fix (#2315) in addition to the already-discussed #2350.

Keeping as draft since we won't merge until tomorrow, but please review/approve. Thanks!


skriss:
@nrb @ashish-amarnath PTAL, thanks!

PR #2389, opened 2020-04-02
Ensure target namespace exists prior to restoring a resource

ashish-amarnath:


skriss:
I think this is fine. It would be slightly better to use the existing `getNamespace(...)` func rather than building a new one; the benefit of that is that the namespace would be restored as-is from the backup tarball, including any labels/annotations/etc. That might be more of a pain though, because you have to deal with the case where the namespace is being remapped.

ashish-amarnath:
I was able to test PR https://github.com/vmware-tanzu/velero-plugin-for-csi/pull/34 with this change and it looked good in testing.

PR #2390, opened 2020-04-02
bug fix: save PVBs even if no snapshot was taken

### LINKED ISSUE #2366: Restic restores original PV rather than dynamically reprovisioning empty volume (restic)
Create an application with a pod that has a mounted volume
Set the "backup.velero.io/backup-volumes" annotation, naming the volume for restic
The volume is empty (basically make sure there are no files or directories in the volume)
Run a backup
The PodVolumeBackup has the following:
```
status:
  completionTimestamp: 2020-03-25T18:48:29Z
  message: volume was empty so no snapshot was taken
```
Next I ran a restore. We happen to have a plugin which changed the storageclass, so the restic-related behavior of not restoring the PV to allow dynamic reprovisioning is important here.
What I observed was that the PV was restored rather than dynamically reprovisioned.

More specifically, at this point:
https://github.com/vmware-tanzu/velero/blob/master/pkg/restore/restore.go#L916
`case hasResticBackup(obj, ctx)` would have matched if there had been data on the volume and restic had created a snapshot, but instead if did not match. Instead, the following block was executed:
https://github.com/vmware-tanzu/velero/blob/master/pkg/restore/restore.go#L930-L940
`Restoring persistent volume as-is because it doesn't have a snapshot and its reclaim policy is not Delete`

I would expect the block under `hasResticBackup` to have executed, forcing reprovisioning of the volume. It's probably OK that there's no restic snapshot, since reprovisioning (and not restoring) would be equivalent to reprovisioning and restoring an empty volume, although if there were a way to allow restic to create and restore snapshots that happen to have no files in them, that would probably be cleaner.

(unfortunately I no longer have the logs -- I should have kept them, but I think this will be reasonably easy to reproduce with an empty mounted volume in a pod)

**Environment:**

- Velero version (use `velero version`): A custom build of velero based on the 1.3.1 release with some local changes
- Velero features (use `velero client config get features`): "<NOT SET>"
- Kubernetes version (use `kubectl version`): 1.11 for the backup, 1.14 for the restore
- OS (e.g. from `/etc/os-release`): Client on Fedora 29

skriss:

Fixes #2366 

cc @sseago - still testing

### DIFF
--- a/changelogs/unreleased/2390-skriss
+++ b/changelogs/unreleased/2390-skriss
@@ -1,0 +1,1 @@
+bug fix: save PodVolumeBackup manifests to object storage even if the volume was empty, so that on restore, the PV is dynamically reprovisioned if applicable
--- a/pkg/restic/backupper.go
+++ b/pkg/restic/backupper.go
@@ -174,9 +174,6 @@ ForEachVolume:
 		case res := <-resultsChan:
 			switch res.Status.Phase {
 			case velerov1api.PodVolumeBackupPhaseCompleted:
-				if res.Status.SnapshotID == "" { // when the volume is empty there is no restic snapshot, so best to exclude it
-					break
-				}
 				podVolumeBackups = append(podVolumeBackups, res)
 			case velerov1api.PodVolumeBackupPhaseFailed:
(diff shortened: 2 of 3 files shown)


skriss:
yep, had one in a local commit that i forgot to push :)

Issue #2267, opened 2020-02-13
Velero snapshot retention Policies: No available documentation?


archmangler:
**What steps did you take and what happened:**

- attempt to find documentation on how to configure backup retention policy on velero site
- No mention of retention policy configuration (site not searchable?)
- We do see restic supports snapshot retention: https://restic.readthedocs.io/en/latest/060_forget.html , however it's not clear how we access this via velero, for example on Kubernetes, how to specify a yaml definition for retention policy?

Use case: We'd like to be able to specify retention policy of snapshots in Azure storage, e.g:

```
retain 30 daily snapshots, 12 monthly snapshots, 7 yearly snapshots
```

**What did you expect to happen:**

- available documentation on configuring retention policy in Azure

**The output of the following commands will help us better understand what's going on**:

n/a

**Anything else you would like to add:**

- velero vesion 1.1.0

**Environment:**

- Azure Cloud
- AKS 1.14.8
- velero 1.1.0

rochfordk:
Raised the above PR to help clarify the use of the TTL flag to control backup retention. 

In the absence of support for more advanced backup retention policies, this use case might be  addressed with multiple backup schedules, e.g.: 

$ velero schedule create nginx-daily-30 --schedule="0 0 * * *" --ttl 730h0m0s --selector app=nginxm

$ velero schedule create nginx-monthly-12 --schedule="0 8 1 * *" --ttl 8760h0m0s --selector app=nginxm

$ velero schedule create nginx-yearly-7 --schedule="0 16 1 5 *" --ttl 61320h0m0s --selector app=nginxm

Check that cron notation and make sure it meets your requirements. I've offset the jobs by 8 hours because there will be occasions when multiple jobs are run in the same day. 

Note you could also use the @daily, @monthly, @yearly cron notation but then you'll loose control over when in the day the job is run and therefore wont be able explicitly define the offset for the jobs

PR #2399, opened 2020-04-06
Highlight use of TTL flag to specify backup retention

rochfordk:
closes https://github.com/vmware-tanzu/velero/issues/2267

Added further TTL information to 'Set a backup to expire' section of how-velero-works.md and included links to that section from the migration and DR use-case pages


rochfordk:
Thanks @skriss , those same changes made to the /master docs as well now. I'll get the hang of this yet!

Issue #1655, opened 2019-07-11
Velero need to support custom credential provider


taojwmware:
**Describe the problem/challenge you have**

Currently, Velero can only use long-lived credentials or kube2iam in order to access storage location. However, kube2iam only works for AWS and using long-lived credentials have negative security implication.  We would like Velero uses temporary security credentials instead.

**Describe the solution you'd like**

We'd like Velero be able to inject a credential provider and get credentials on-demand and the credentials provider can provide temporary credentials instead.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Velero version (use `velero version`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

ashish-amarnath:
Implementing a credential provider as a plugin seems to be the right solution to this problem. However that will require a major revamp of our plugin framework. Requiring the the storage/ volumesnapshot plugins to talk to the credential provider plugin. This revamp will take a longer time to have a solution to this problem. 
Based on the discussions with @nrb and others, we might be better of keeping this lightweight and potentially reuse existing tools for credentials management. E.g. credhub https://www.youtube.com/watch?v=uevyVtnPOwI

ashish-amarnath:
The other approach may be to delegate the credential management to each of the plugins. This would couple credential management with the plugins that may use it. It will also give each individual plugins the ability to support their own credential management at their own pace and flexibility in their implementation. 
IMO, this approach is a nice middle ground between having an out of band credential getter and a credential plugin that works across provider implementations.
This approach however, will force plugins to run with only one way of managing credentials, in the case where multiple may be available.

nrb:
> The other approach may be to delegate the credential management to each of the plugins.

Which plugin types? ObjectStore and VolumeSnapshotter?

eleanor-millman:
@vmmore Do you still want this for TMC Data Protection?

PR #2403, opened 2020-04-07
Proposal for handling multiple credential secrets

carlisia:

### DIFF
--- a/design/secrets.md
+++ b/design/secrets.md
@@ -54,7 +54,54 @@ There are three different approaches that can be taken to provide credentials to
 The last two options require changes to the plugin as the plugin will need to instantiate a client using the provided credentials.
 The client libraries used by the plugins will not be able to rely on the credentials details being available in the environment as they currently do.
 
-Each of the approaches will be discussed in turn.
+We have selected option 2 as the approach to take which will be described below.
+
+### Including the credentials file path in the `config` map
+
+Prior to using any secret for a BSL or VSL, it will need to be serialized to disk.
+Using the details in the `Credential` field in the BSL/VSL, the contents of the Secret will be read and serialized.
+To achieve this, we will create a new package, `credentials`, which will introduce new types and functions to manage the fetching of credentials based on a `SecretKeySelector`.
+This will also be responsible for serializing the fetched credentials to a temporary directory on the Velero pod filesystem.
+
+The path where a set of credentials will be written to will be a fixed path based on the namespace, name, and key from the secret rather than a randomly named file as is usual with temporary files.
+The reason for this is that `BackupStore`s are frequently created within the controllers and the credentials must be serialized before any plugin APIs are called, which would result in a quick accumulation of temporary credentials files.
+For example, the default validation frequency for BackupStorageLocations is one minute.
+This means that any time a `BackupStore`, or other type which requires credentials, is created, the credentials will be fetched from the API server and may overwrite any existing use of that credential.
+
+If we instead wanted to use an unique file each time, we could work around the of multiple files being written by cleaning up the temporary files upon completion of the plugin operations, if this information is known.
+
+Once the credentials have been serialized, this path will be made available to the plugins.
+Instead of setting the necessary environment variable for the plugin process, the `config` map for the BSL/VSL will be modified to include an addiitional entry with the path to the credentials file: `credentialsFile`.
+This will be passed through when [initializing the BSL/VSL](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/velero/object_store.go#L27-L30) and it will be the responsibility of the plugin to use the passed credentials when starting a session.
+For an example of how this would affect the AWS plugin, see [this PR](https://github.com/vmware-tanzu/velero-plugin-for-aws/pull/69).
+
+The restic controllers will also need to be updated to use the correct credentials.
+The BackupStorageLocation for a given PVB/PVR will be fetched and the `Credential` field from that BSL will be serialized.
+The existing setup for the restic commands use the credentials from the environment variables with [some repo provider specific overrides](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/pod_volume_backup_controller.go#L260-L273).
+Instead of relying on the existing environment variables, if there are credentials for a particular BSL, the environment will be specifically created for each `RepoIdentifier`.
+This will use a lot of the existing logic with the exception that it will be modified to work with a serialized secret rather than find the secret file from an environment variable.
+Currently, GCP is the only provider that relies on the existing environment variables with no specific overrides.
+For GCP, the environment variable will be overwritten with the path of the serialized secret.
+
+
+## Backwards compatibility
+
+For now, regardless of the approaches used above, we will still support the existing workflow.
+
+Users will be able to set credentials during install and a secret will be created for them.
+This secret will still be mounted into the Velero pods and the appropriate environment variables set.
+This will allow users to use versions of plugins which haven't yet been updated to use credentials directly, such as with many community created plugins.
+
+Multiple credential handling will only be used in the case where a particular BSL/VSL has been modified to use an existing secret.
+
+## Security Considerations
+
+Although the handling of secrets will be similar to how credentials are currently managed within Velero, care must be taken to ensure that any new code does not leak the contents of secrets, for example, including them within logs.
+
+## Alternatives Considered
+
+As mentioned above, there were three potential approaches for providing this support.
+The approaches that were not selected are detailed below for reference.
 
 #### Providing the credentials via environment variables
 
@@ -79,31 +126,10 @@ We would also need to ensure that the restic controllers are updated in the same
 This could be achieved by modifying the existing function to [run a restic command](https://github.com/vmware-tanzu/velero/blob/main/pkg/restic/repository_manager.go#L237-L290).
 This function already sets environment variables for the restic process depending on which storage provider is being used.
 
-
-#### Including the credentials file path in the `config` map
-
-There is some overlap with the previous approach in that it will involve retrieving the secret for the BSL/VSL being used and serializing it to disk.
-Instead of setting the necessary environment variable for the plugin process, the `config` map for the BSL/VSL will be modified to include an addiitional entry with the path to the credentials file.
-This will be passed through when [initializing the BSL/VSL](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/velero/object_store.go#L27-L30) and it will be the responsibility of the plugin to use the passed credentials when starting a session.
-For an example of how this would affect the AWS plugin, see [this PR](https://github.com/vmware-tanzu/velero-plugin-for-aws/pull/69).
-
 #### Include the details of the secret in `config` map passed to a plugin
 
-This approach is like the previous one, however instead of the Velero process being responsible for serializing the file to disk prior to invoking the plugin, the `Credential SecretKeySelector` details will be passed through to the plugin.
+This approach is like the selected approach of passing the credentials file via the `config` map, however instead of the Velero process being responsible for serializing the file to disk prior to invoking the plugin, the `Credential SecretKeySelector` details will be passed through to the plugin.
 It will be the responsibility of the plugin to fetch the secret from the Kubernetes API and perform the necessary steps to make it available for use when creating a session, for example, serializing the contents to disk, or evaluating the contents and adding to the process environment.
 
 This approach has an additional burden on the plugin author over the previous approach as it requires the author to create a client to communicate with the Kubernetes API to retrieve the secret.
-
-## Backwards compatibility
-
-For now, regardless of the approaches used above, we will still support the existing workflow.
-
-Users will be able to set credentials during install and a secret will be created for them.
-This secret will still be mounted into the Velero pods and the appropriate environment variables set.
-This will allow users to use versions of plugins which haven't yet been updated to use credentials directly, such as with many community created plugins.
-
-Multiple credential handling will only be used in the case where a particular BSL/VSL has been modified to use an existing secret.
-
-## Security Considerations
-
-Although the handling of secrets will be similar to how credentials are currently managed within Velero, care must be taken to ensure that any new code does not leak the contents of secrets, for example, including them within logs.
+Although it would be the responsibility of the plugin to serialize the credential and use it directly, Velero would still be responsible for serializing the secret so that it could be used with the restic controllers as in the selected approach.
--- a/design/secrets.md
+++ b/design/secrets.md
@@ -1,6 +1,6 @@
 # Support for multiple provider credentials
 
-Currently, Velero only supports a single AIM access credential secret per location provider/plugin. Velero creates and stores the plugin credential secret under the hard-coded key `secret.cloud-credentials.data.cloud`.
+Currently, Velero only supports a single credential secret per location provider/plugin. Velero creates and stores the plugin credential secret under the hard-coded key `secret.cloud-credentials.data.cloud`.
 
 This makes it so switching from one plugin to another necessitates overriding the existing credential secret with the appropriate one for the new plugin.
 
--- a/design/secrets.md
+++ b/design/secrets.md
@@ -1,4 +1,4 @@
-# Design proposal template (replace with your proposal's title)
+# Support for multiple provider credentials
 
 Currently, Velero only supports a single AIM access credential secret per location provider/plugin. Velero creates and stores the plugin credential secret under the hard-coded key `secret.cloud-credentials.data.cloud`.
 
--- a/design/secrets.md
+++ b/design/secrets.md
@@ -69,7 +69,7 @@ Each plugin process would still have the same set of environment variables set, 
 
 To set the environment variables for a plugin process, the plugin manager must be modified so that when creating an ObjectStore or VolumeSnapshotter, we pass in the entire BSL/VSL object, rather than [just the provider](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/manager.go#L132-L158).
 The plugin manager currently stores a map of [plugin executables to an associated `RestartableProcess`](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/manager.go#L59-L70).
-New restartable processes are created only [with the exectuable that the process would run](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/manager.go#L122).
+New restartable processes are created only [with the executable that the process would run](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/manager.go#L122).
 This could be modified to also take the necessary environment variables so that when [underlying go-plugin process is created](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/client_builder.go#L78), these environment variables could be provided and would be set on the plugin process.
 
 Taking this approach would not require any changes from plugins as the credentials information would be made available to them in the same way.
--- a/design/secrets.md
+++ b/design/secrets.md
@@ -1,0 +1,80 @@
+# Design proposal template (replace with your proposal's title)
+
+Currently, Velero only supports a single AIM access credential secret per location provider/plugin. Velero creates and stores the plugin credential secret under the hard-coded key `secret.cloud-credentials.data.cloud`.
+
+This makes it so switching from one plugin to another necessitates overriding the existing credential secret with the appropriate one for the new plugin.
+
+## Goals
+
+- To allow Velero to create and store multiple secrets for provider credentials
+- To improve the UX for configuring the velero deployment with multiple plugins/providers, and corresponding IAM secrets.
+
+## Non Goals
+
+- To make any change except what's necessary to handle multiple credentials
+
+## Detailed Design
+
+- the name of the flag changes from `secret-file` to `--credentials-file`. 
+
+- The `velero plugin (create|set) --credentials-file` will be a map of provider name as a key, and the path to the file as a value. This way, we can have multiple credential secrets and each secret per provider/plugin will be unique.
+
+See discussion https://github.com/vmware-tanzu/velero/pull/2259#discussion_r384700723 for the two items below.
+
+- The `velero backup-location (create|set)` will have a new `--credentials mapStringString` flag which sets the name of the corresponding credentials secret for a provider. Format is provider:credentials-secret-name.
+
+- The `velero snapshot-location (create|set)` will have a new `--credentials mapStringString` flag which sets the list of name of the corresponding credentials secret for providers. Format is (provider1:credentials-secret-name1,provider2:credentials-secret-name2,...).
+
+Note that for this logic to work we must have a controller loop checking for when a corresponding secret is present before marking the BSL/VSL as ready.
+
+
+#### Examples of mounting secrets and environment variables
+
+With the changes proposed in the previous section (Credentials and secrets), the resulting deployment `yaml` would look like below:
+
+AWS
+```
+   spec:
+     containers:
+        volumeMounts:
+        - mountPath: /credentials
+          name: cloud-credentials-aws
+      - args:
+        - server
+        name: velero
+        env:
+        - name: AWS_SHARED_CREDENTIALS_FILE
+          value: /credentials/cloud
+     volumes:
+      - name: cloud-credentials
+        secret:
+          secretName: cloud-credentials
+```
+
+Azure
+```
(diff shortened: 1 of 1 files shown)


carlisia:
Hey @vmware-tanzu/velero-maintainers, would love to get a yay or nay on this. If we want this change, I'll open a ticket, maybe for v1.6?

carlisia:
This plan to allow the AWS Cluster API Provider to support multiple sets of AWS credentials/roles from a single controller for the cluster-api project might be helpful:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/040406e90dd5f67e840cad00b3ec01c89ffbc24c/docs/proposal/20200506-single-controller-multitenancy.md

nrb:
Assigning a tentative milestone of v1.6 for visibility.

nrb (commented):
We definitely need this ability, but one thing that I think this proposal needs to also address is rotation/[short-lived credentials](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html). This isn't exactly multiple credentials, but it's related to how we handle secrets and I think it will be worthwhile to think about at the same time.

These are important because some systems will use these to minimize blast radius should an attack occur. If a short-lived credential is acquired, it can only be exploited for a limited amount of time. Ideally this would be Velero's default, but I also don't think we should make that change immediately without a transition period.

Some things I can think of off the top of my head that will help support the short-lived credentials, though not an exhaustive list:

* Re-trying failed uploads at the end of a backup, in case the credential expired during backup. Pair this with exponential back-off, so we're not retrying forever?
* Re-trying snapshot operations if we get a permission denied error on taking a snapshot, in case the credential expired during the snapshot operation.
* Making sure rotated credentials are supplied to restic, so that its own transfer operations are carried out completely. This will require some research into restic itself, to see exactly what it supports

In terms of getting them into the pod, I need to do more research to see if the files that secrets get mounted as are updated quickly or not.

design/secrets.md
```diff
@@ -0,0 +1,80 @@
+# Design proposal template (replace with your proposal's title)
+
+Currently, Velero only supports a single AIM access credential secret per location provider/plugin. Velero creates and stores the plugin credential secret under the hard-coded key `secret.cloud-credentials.data.cloud`.
+
+This makes it so switching from one plugin to another necessitates overriding the existing credential secret with the appropriate one for the new plugin.
+
+## Goals
+
+- To allow Velero to create and store multiple secrets for provider credentials
```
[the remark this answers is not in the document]
carlisia: Yes, this is precisely it. Another way to say it: there would be a 1.1 secret to BSL, and each BSL (therefore secret) would be associated with a provider.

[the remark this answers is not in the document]
carlisia: > The velero backup-location (create|set) will have a new --credentials mapStringString flag which sets the name of the corresponding credentials secret for a provider. Format is provider:credentials-secret-name.

To clarify, and I think I'm backtracking here: a plugin will have n secrets. A BSL for that plugin can use any of those secrets. There might be > 1 BSL using the same secret for that plugin.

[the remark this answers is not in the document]
nrb: @carlisia Would it also be possible to get current behavior, as in 1:M secret to BSL?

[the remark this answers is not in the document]
nrb: That makes sense to me.

The 1:1 mapping would be new. 1:M mapping (secret:BSL) would be like our current behavior, where 1 secret applies to multiple (or all) BSLs/VSLs.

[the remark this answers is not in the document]
carlisia: The point about allowing multiple plugins to be used at the same time is a good point tho. I did not address this in this proposal but with this feature added, it only makes sense and it'll be benefitial.

nrb:
Upon further consideration, my previous comment is a completely different feature. The data model should allow supporting rotation, but I think implementing these separately makes sense.

carlisia:
This might be pertinent: https://github.com/vmware-tanzu/velero/issues/1655.

zubron:
I've spent some time looking at this design proposal and the related areas of the code and I have some thoughts.

The following is quite verbose, but a lot of it is me trying to make sure I understand things :) I've written this under the assumption that the proposal will be updated to reflect the comments [here](https://github.com/vmware-tanzu/velero/pull/2403#discussion_r438359910), that it will allow multiple credentials to be set **per plugin** so that, for example, different credentials could be used for different BSLs which are both on AWS.

If we want to move away from having a single credential object and instead have a more dynamic approach to creating and using credentials (adding them when adding a new plugin or creating a new BSL or VSL), we will need to modify how plugins are invoked so that the correct credential is used. Currently, there is a single secret, which is mounted into every pod deployed by Velero (theVelero Deployment and the Restic DaemonSet). This secret contains a single data key (`cloud`) which is accessible within the pods at the path `/credentials/cloud`. This path is made known to all plugins through provider specific environment variables. All possible provider environment variables are set to this path.

The examples in the current design proposal show different secrets being mounted however, my understanding is that is not possible to modify volume mounts in a running pod/container. Changes to secrets (such as adding new data keys) are automatically updated within the volume mount. This means that any update to the `cloud-credentials` secret would automatically be available to pods where that secret is mounted. This would allow us to work around the restriction of modifying volume mounts to instead still use a single secret, but add more keys to it for new sets of credentials.

Instead of using fixed environment variables which reference a single secret when creating the pods when installing velero, we will need to modify how we invoke plugins so that a different set of credentials can be selected for use. There are two ways which came to mind:
* Instead of setting an environment variable, pass the path of the necessary credentials file through to the plugins and make any plugin responsible for using that credentials file when creating a client to interact with a storage provider. This has the benefit of removing provider specific information from the core velero codebase however this would require all plugins to be updated.
* Modify `restartableProcess` to set the environment variables before running a plugin process. Each plugin process would still have the same set of environment variables set, however the value used for each of these variables would instead be a different path within the `/credentials` directory. Taking this approach would not require any changes from plugins as the credentials information would be made available to them in the same way.

We will also need to ensure that the restic controllers are updated in the same way so that correct credentials are used (when creating a `ResticRepository` or processing `PodVolumeBackup`/`PodVolumeRestore`).

When we need to add a new set of credentials (either with a new plugin, or with a new BSL/VSL), we will read in the credentials file and add it as a new key to the existing `cloud-credentials` secret. The key can be named in such a way that associates it with the corresponding plugin/BSL/VSL. The credential to use can be selected based on the name, or a reference to the credential can be stored with the corresponding object (BSL/VSL), and then made available to a plugin in one of the ways described above.

zubron:
Converting this to a draft PR to remove it from the review queue while I am making some changes to it.

zubron:
This still needs some further updates to discuss the impact on upgrades and Helm but marking as ready for review now to get some input.

design/secrets.md
```diff
@@ -0,0 +1,81 @@
+# Design proposal template (replace with your proposal's title)
+
+Currently, Velero only supports a single AIM access credential secret per location provider/plugin. Velero creates and stores the plugin credential secret under the hard-coded key `secret.cloud-credentials.data.cloud`.
+
+This makes it so switching from one plugin to another necessitates overriding the existing credential secret with the appropriate one for the new plugin.
+
+## Goals
+
+- To allow Velero to create and store multiple secrets for provider credentials, even multiple credentials for the same provider
+- To improve the UX for configuring the velero deployment with multiple plugins/providers, and corresponding IAM secrets.
```
nrb, line 10: This line mentions IAM secrets, though we don't discuss them anywhere else in the document.

What can we do to make that experience better? Often, IAM credentials are applied at the node level and may not be Kubernetes secrets.

Based on https://github.com/vmware-tanzu/velero-plugin-for-aws#option-2-set-permissions-using-kube2iam, it looks like that would be up to kube2iam or similar. It may also be the case that IAM scenarios are not covered by this particular design.

zubron: The lack of discussion on IAM credentials is my fault. The initial set of goals was written by @carlisia and I didn't expand on that topic. Now that I look at the link you posted, I don't think the kube2iam approach is compatible with the case of having multiple credentials for AWS managed by Velero. An IAM role could be created which supports multiple sets of credentials, but that isn't something that is controllable through Velero. And I'm not sure if we would want to introduce such specific AWS functionality into Velero.

carlisia: I would rather have the IAM case be handled separately from this design change if possible.

carlisia: I think I wrote that down.

nrb: Yeah, I agree on not introducing any specific infra provider functionality into it; I was using AWS as a popular example. However, [GCP has a similar feature](https://github.com/vmware-tanzu/velero-plugin-for-gcp#option-2-set-permissions-with-using-workload-identity-optional), and [so does Azure](https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure#option-2-use-aad-pod-identity).

They're called different things, but I think the idea is the same in that they all three tie the auth to the nodes/pods instead of via a credentials file. A brief skim of the READMEs says that the Azure one is the only one that has K8s resources associated directly in this flow, but the others may have some behind the scenes that aren't covered in the readme.

nrb: > I would rather have the IAM case be handled separately from this design change if possible.

100% fine with that. I think we can call that out explicitly in the non-goals section. Something like "Node-based auth features" or similar so that we're not just focused on the AWS solution.

zubron: Hmm, the Azure one looks similar to the existing workflow, the outcome of using AAD pod identity is to create a credentials file which is then used with `velero` install. It's not clear to me if multiple roles/identities can be handled through one file in that method.

The others though are managed outside of velero though, and I think would continue to work as before. If you had a BSL that used a plugin which included those methods of specifying credentials, you would say that no credentials are needed and then no credentials would be passed through in the environment. I think if you needed to access other accounts using those plugins, it might be possible to add other credentials, and use them in the environment for the plugin process (overriding the credentials for the pod/deployment).

carlisia (commented):
> we can modify restartableProcess to set the environment variables before running a plugin process

@zubron would it be possible to give us a code sample of what this could look like? Particularly, how would this work w/o making a change to the plugin itself. This is what I'm trying to figure out, if you have thought this far and could give an example it would be helpful.

Everything else is making sense to me.

carlisia:
One other thing that I'm thinking is how we are thinking of the cases below:

Case (1)
- A BSL is configured for provider X
- The plugin for provider X is configured with many credentials. 
- The BSL will try to validate itself against each until one is found to be valid.

Case (2)
- A BSL is configured for provider X
- The plugin for provider X is configured with credentials A, B, C and D
- The BSL is configured to use credentials A and D
- Try both credentials until one of the two is found to be valid

Case (3)
- A BSL is configured for provider X
- The plugin for provider X is configured with credentials A, B, C and D
- The BSL can only be configured to use one of these credentials

design/secrets.md
```diff
@@ -0,0 +1,81 @@
...
+  This secret contains a single data key (`cloud`) which is accessible within the pods at the path `/credentials/cloud`.
+  This path is made known to all plugins through provider specific environment variables.
+  All possible provider environment variables are set to this path.
+  Instead of setting the environment within all the pods, we can modify `restartableProcess` to set the environment variables before running a plugin process.
+  Each plugin process would still have the same set of environment variables set, however the value used for each of these variables would instead be a different path within the `/credentials` directory, formed from the credential selected for a particular BSL/VSL.
+  Taking this approach would not require any changes from plugins as the credentials information would be made available to them in the same way.
+  We will also need to ensure that the restic controllers are updated in the same way so that correct credentials are used (when creating a `ResticRepository` or processing `PodVolumeBackup`/`PodVolumeRestore`).
+
+## Alternatives Considered
```
ashish-amarnath, line 70: Here is another alternate proposal:

Introduce a new `credentials` config in the `BackupStorageLocationSpec` and `VolumeSnapshotLocationSpec`.
This config will store a reference to a kubernetes secret that is managed is created by the user. using a command, say `velero credentials create` or at the time of creating the BSL or VSL.

The plugins will:
1. Read the `credentials` config key, if it exists will fetch the credentials from the secret.
1. Create a session.

We could also make `credentials` a field which is a kubernetes object reference to a secret. I am not sure what advantages it can get us but is something that can be thought through in more detail.

Pros:
1. Allows us to create credentials per BSL/VSL and not one per plugin. Users can add the, say, `velero-plugin-for-aws` once and use a different set of credentials based on what the BSL is configured to use. This, IMO, is better than allowing credentials per plugin.
1. Allows users to create separate credentials: one for accessing the object store, another for volume snapshotting. Following the security best practices of principle of least privilege.
1. This can be a step towards multi-tenancy, albeit a small one.
1. Users can also self-manage the BSL/VSL credentials when they wish to rotate credentials. Following the best practices of regular credential rotation.
1. Velero may never have to read the credentials. Somewhat reducing the security vulnerability surface.

Cons:
1. This will require plugin implementations to be updated to use the new behavior. However, existing plugins can still rely on the `cloud-credentials` secret to function as they currently do.

design/secrets.md
```diff
@@ -0,0 +1,81 @@
...
+  We will add keys to the existing secret, rather than creating new ones, due to the fact that it is not possible to modify volume mounts in a running pod or container.
+  Changes to secrets (such as adding new data keys) are automatically updated within the volume mount.
+  This means that any update to the `cloud-credentials` secret would automatically be available to the pods where that secret is mounted.
+  This allows us to work around the restriction of modifying volume mounts to instead still use a single secret, but add more keys to it for new sets of credentials.
+ 
+- BSLs and VSLs will be updated to contain a reference to a particular credential name, which will be a key in the `cloud-credentials` secret.
+  When a new BSL or VSL is created, a specific set of credentials can be selected for use.
+  The set of credentials associated with a BSL or VSL will be used when running the plugin processes needed for that storage location.
+
+## Detailed Design
```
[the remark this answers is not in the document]
ashish-amarnath: Also, what is the transition plan for users from the current state to the one proposed in this one? This change needs to continue to work with the `cloud-credentials` approach to give our users an opportunity to cutover to the new solution.

zubron:
> > we can modify restartableProcess to set the environment variables before running a plugin process
> 
> @zubron would it be possible to give us a code sample of what this could look like? Particularly, how would this work w/o making a change to the plugin itself. This is what I'm trying to figure out, if you have thought this far and could give an example it would be helpful.

This is what I had been working on this week and I misunderstood how these processes are initialised and used, but I think it's a case of modifying how the processes for each plugin are created when required so that they add the credentials to the environment [here](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/client_builder.go#L65-L80) but I think it's going to require changes to the plugin [management interfaces](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/manager.go#L29-L57). I don't know if changing those interfaces will be a breaking change though.

ashish-amarnath (changes requested):
I made a couple of comments and suggested an alternate proposal that might give us a 1:1 mapping between creds to BLS/VSL.
Happy to discuss more on the alternate proposal.

design/secrets.md
carlisia: > Allows users to create separate credentials

This is a great point.

Couple of thoughts just for the sake of reinforcing some ideas: with the credentials being bound to a provider, which is to say, bound to the plugin. it would be nice if we could somehow namespace the credentials on a per plugin basis. And, the BSL/VSL can reuse existing credentials, meaning, aws BSL X and aws BSL Y can both use credential A for aws.

zubron:
> One other thing that I'm thinking is how we are thinking of the cases below:
> 
> Case (1)
> 
> * A BSL is configured for provider X
> * The plugin for provider X is configured with many credentials.
> * The BSL will try to validate itself against each until one is found to be valid.

I think BSLs will have to explicitly be associated with a particular credential. In the doc I suggested that when adding a new BSL, the credential to use for that plugin will be set on the BSL/VSL. In this case, it would directly try to use the plugin with the specified credential (fetched from the velero `cloud-credentials` secret).

> Case (2)
> 
> * A BSL is configured for provider X
> * The plugin for provider X is configured with credentials A, B, C and D
> * The BSL is configured to use credentials A and D
> * Try both credentials until one of the two is found to be valid

The way I viewed this is that only one credential would be specified per BSL. Is there a use case that I'm missing where a single BSL needs to be configured with multiple credentials?

> Case (3)
> 
> * A BSL is configured for provider X
> * The plugin for provider X is configured with credentials A, B, C and D
> * The BSL can only be configured to use one of these credentials

Yep, this is how I was thinking it would work.

Overall though, I don't think that there will necessarily be anything tying a credential to a plugin other than the fact that a particular format will only be applicable to a plugin. That's was the reasoning behind having a separate command to manage the credentials: https://github.com/vmware-tanzu/velero/blob/8e55ddfd3d60f85fe7bad13396c8525c81ad8342/design/secrets.md#alternatives-considered

design/secrets.md
zubron: @ashish-amarnath your suggested approach was the path I initially went down, but was [concerned about the impact on existing plugins (first bullet point)](https://github.com/vmware-tanzu/velero/pull/2403#issuecomment-728132546). If we can leave the existing environment in place however, and we're comfortable with requiring plugins to make changes, then I think this is the much cleaner and simpler approach. Longer term, it would also allow us to move away from having provider specific information in our deployment.

One thing we would have to advertise though is that this would be supporting multiple credentials *but only for plugins that support it*. If we're okay with that, then I would much prefer to take this approach.

carlisia:
>  I don't think that there will necessarily be anything tying a credential to a plugin

I wonder if we should artificially tie them together just so it'd be easier for the user to list what secret belongs to what plugin, when they are using more than one (more than one plugin and more than one secret). Something like `velero get plugin <PLUGIN-NAME> secrets`.

design/secrets.md
zubron: @ashish-amarnath @carlisia I made a draft PR with the basic changes that would be needed to make the approach of specifying the credentials file in the config work: https://github.com/vmware-tanzu/velero-plugin-for-aws/pull/69
This change would obviously need more work (e.g. checking that the file exists) but this works with Velero as it is (if you update the secret and add the key to the BSL config). If we can agree that we're happy to put the change into the plugins, this would _much_ nicer :+1:

zubron:
> I wonder if we should artificially tie them together just so it'd be easier for the user to list what secret belongs to what plugin, when they are using more than one (more than one plugin and more than one secret). Something like `velero get plugin <PLUGIN-NAME> --secrets`.

@carlisia  I think this would be really nice. I'm still getting my head around how plugins are managed in Velero. From what I can tell, there isn't a plugin "entity" within the codebase, but we maybe we could store the names of credentials as they are added in the [plugin registry](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/registry.go#L47-L58)?

design/secrets.md
```diff
@@ -0,0 +1,81 @@
...
+- To allow Velero to create and store multiple secrets for provider credentials, even multiple credentials for the same provider
+- To improve the UX for configuring the velero deployment with multiple plugins/providers, and corresponding IAM secrets.
+
+## Non Goals
+
+- To make any change except what's necessary to handle multiple credentials
+
+## Design overview
+
+- Instead of one credential per Velero deployment, multiple credentials can be added and will be associated with plugins.
```
nrb, line 18: Associated with plugins, or volume/backup locations?

design/secrets.md
```diff
@@ -0,0 +1,81 @@
...
+Instead of associating credentials with plugins, credentials will exist as standalone entities which can be queried using a new command `velero credentials`.
+This will be a wrapper around Kubernetes API calls to query the secret used by Velero.
+Credentials can be listed using `velero credentials get`, which will list the names of the keys within the `cloud-credentials` secret.
+Credentials can be added or removed using `velero credentials add|delete`.
+
+The rest of the workflow of how to use these credentials would still apply, such as specifying the name of the credential to use for a particular BSL/VSL.
+
+## Security Considerations
+
+N/A
```
nrb, line 81: Given we're working with secrets, I think security applies here, and we should write out what's happening, even if the exposed data ends up being the same.

design/secrets.md
nrb: @ashish-amarnath For your proposal, how would said credentials be exposed to restic, since it doesn't go through the plugin code path?

design/secrets.md
```diff
@@ -0,0 +1,81 @@
...
+  This way, we can have multiple credential secrets and each secret per provider/plugin will be unique.
+  There are two ways in which we can choose to name the keys stored in the `cloud-credentials` secret:
+    1. Combine the name of the provider (extracted from the plugin name) and the given credentials name.
+       For example, if the user were to add a plugin using: `velero plugin add velero-plugin-for-aws --credentials-file project-1:/path/to/credentials`, an additional key `aws-project-1` would be added to the `cloud-credentials` secret with the encoded contents of `/path/to/credentials`.
+       In the case where the plugin is not a Velero provided plugin, we would use the full name of the plugin but adapt the name to be valid for use as a key, for example, `myproject.io/cloud-provider` could become `myproject_io_cloud-provider`, and would be combined with the given name of the credential.
+    1. Do not include the provider name in the resulting credentials key, and instead leave it up to the user to give a unique and identifying name for a credential.
```
nrb, line 41: I like this one, personally. Having generated ones can lead to confusion.

design/secrets.md
```diff
@@ -0,0 +1,81 @@
...
+  The set of credentials associated with a BSL or VSL will be used when running the plugin processes needed for that storage location.
+
+## Detailed Design
+
+- The name of the flag changes from `secret-file` to `--credentials-file`.
+
+- The arguments to `velero plugin (add|set) --credentials-file` will be a map of the credentials name as a key, and the path to the file as a value.
+  This way, we can have multiple credential secrets and each secret per provider/plugin will be unique.
+  There are two ways in which we can choose to name the keys stored in the `cloud-credentials` secret:
+    1. Combine the name of the provider (extracted from the plugin name) and the given credentials name.
```
nrb, line 38: If we do go with generated names, I would treat them all the same - the velero.io plugins should use the same rules as ones we don't ship, in my opinion, as we're going to remove the exception case eventually. That also means less code to worry about.

design/secrets.md
zubron: One thing to consider with the above approach, I think it would be up to Velero to specify the value in the config rather than the user providing it. I think with this approach, the user would state which credential they want to use for that BSL (a new optional field) and then when calling `Init` with the config, Velero would pass through the config with the correct path for the credentials file (similarly to how we handle the `caCert` values).

zubron:
@vmware-tanzu/velero-maintainers I've updated the doc to include more of the details discussed during the meeting on Wednesday - please take a look :)

The doc does not make a decision on how the credentials will be made available to the plugins, just outlines the different approaches. I'm personally leaning towards option 2 which is to serialize the contents of the secret to a tmp file and pass it through via `config` to the plugin. Although this requires modifications in the plugins, it is ultimately more flexible as it removes the need to Velero to have to know about the different env vars needed by each plugin.

design/secrets.md
```diff
@@ -0,0 +1,109 @@
+# Design proposal template (replace with your proposal's title)
+
+Currently, Velero only supports a single AIM access credential secret per location provider/plugin. Velero creates and stores the plugin credential secret under the hard-coded key `secret.cloud-credentials.data.cloud`.
```
jenting, line 3: Typo, IAM.

jenting (commented):
If I understand correctly, the option 2 approach is:
1. The Velero server pod writes the secret content to the Velero pod's or node's disk.
2. The plugin uses the credentials from pod or host disk.

But as far as I know, the plugin is mounted as initContainer (before Velero server pod). How can we guarantee that the disk credentials (generated by Velero server) could be created before the plugin (initContainer) created?

design/secrets.md
dsu-igeek: The plugin's initContainer just copies the plugin into the /plugins directory and doesn't actually run the plugin in the container.  The plugin is run in the main container as a process.

jenting: Thanks for the explanation.

design/secrets.md
```diff
@@ -0,0 +1,135 @@
...
```
nrb, line 3: `AIM` -> `IAM`, though I think we can actually say "single credential secret per platform."

design/secrets.md
```diff
@@ -0,0 +1,135 @@
+# Design proposal template (replace with your proposal's title)
```
nrb, line 1: Let's update the title to mention what the proposal's about :)

nrb (changes requested):
One minor change and one big one - we need a title before accepting ;)

nrb (approved):

jenting (approved):

Issue #2302, opened 2019-12-16
Update backup deletion controller to find and delete VolumeSnapshot/VolumeSnapshotContents objects


nrb:
The backup deletion controller should be updated to find any `VolumeSnapshot` and `VolumeSnapshotContents` objects associated with a Velero `Backup` and delete them when the Velero `Backup` is deleted, either manually or via a TTL expiration.

Old description from when we assumed `OwnerRefs` could cross namespaces follows. This won't work because Velero assumes all of its `Backup`s will live in its own namespace, while the `VolumeSnapshot` will be in the same namespaces as their PVCs and `VolumeSnapshotContents` objects will be cluster-scoped.

~Per [the design document](https://github.com/vmware-tanzu/velero/blob/master/design/csi-snapshots.md#garbage-collection-and-deletion), we want to set an ownerRef to the Velero Backup that creates the VolumeSnapshot and VolumeSnapshotContent so that they will be automatically deleted by Kubernetes garbage collection when that backup no longer exists.~

skriss:
Couple notes for implementation:

- we need to make sure we handle the scenario where we just synced backups into a brand new cluster, meaning the VolumeSnapshot CRs don't exist in the cluster yet
- we probably need a `velero.io/backup-name` label on *at least* the VolumeSnapshots, and maybe the VolumeSnapshotContents. Right now, it looks like we're adding an annotation to the VolumeSnapshot with this, but labels make filtering easy.

nrb:
> we probably need a velero.io/backup-name label on at least the VolumeSnapshots, and maybe the VolumeSnapshotContents. Right now, it looks like we're adding an annotation to the VolumeSnapshot with this, but labels make filtering easy.

Yeah, my assumption on #2323 was that both would have a label, so I could use a label selector when trying to upload them.

ashish-amarnath:
The backup deletion controller will not have to delete the volumesnapshotcontent objects. Instead, just deleting the volumesnapshot should suffice. The volumesnapshotcontents will get deleted or retained based on the deletion policy of the volumesnapshotcontent.

So the backup deletion controller will have to discover the volumesnapshots using a label selector and delete just those volumesnapshots.

ashish-amarnath:
Capturing notes from discussions:
[...]

PR #2411, opened 2020-04-10
Delete CSI volumesnapshots created as part of a backup in its deletion

ashish-amarnath:

Fixes: #2302 

When a backup is being deleted using `velero backup delete <backup-name>`
Discover all CSI volumesnapshots that were created as part of the backup. This is done by:
- fetching, from backup spec's include-exclude, the namespaces that are present in the backup.
- for each namespace in the backup, get a list of all volumesnapshots which have the `velero.io/backup-name=<backup-name>` label and delete them.

Here we don't explicitly delete the underlying volumesnapshotcontents, because they will be either deleted or retained based on the `DeletionPolicy` of the volumesnapshotclass.

This change uses the kubernetes CRD objects as the source of truth and not the contents of the backup in the object store.

### DIFF
--- a/pkg/backup/backup_test.go
+++ b/pkg/backup/backup_test.go
@@ -2406,62 +2406,6 @@ func (b *fakeResticBackupper) BackupPodVolumes(backup *velerov1.Backup, pod *cor
 	return res, nil
 }
 
-func TestGetNamespacesInBackup(t *testing.T) {
-	testCases := []struct {
-		name               string
-		backup             *velerov1.Backup
-		expectedNamespaces []string
-	}{
-		{
-			name: "backup with nil include and exclude namespaces",
-			backup: &velerov1.Backup{
-				Spec: velerov1.BackupSpec{
-					IncludedNamespaces: nil,
-					ExcludedNamespaces: nil,
-				},
-			},
-			expectedNamespaces: []string{""}, // this signifies all namespaces
-		},
-		{
-			name: "backup with empty include and exclude namespaces",
-			backup: &velerov1.Backup{
-				Spec: velerov1.BackupSpec{
-					IncludedNamespaces: []string{},
-					ExcludedNamespaces: []string{},
-				},
-			},
-			expectedNamespaces: []string{""}, // this signifies all namespaces
-		},
-		{
-			name: "backup with non-empty include and empty exclude namespaces",
-			backup: &velerov1.Backup{
-				Spec: velerov1.BackupSpec{
-					IncludedNamespaces: []string{"include-1", "include-2", "include-3", "include-4", "include-5"},
-					ExcludedNamespaces: []string{},
-				},
-			},
-			expectedNamespaces: []string{"include-1", "include-2", "include-3", "include-4", "include-5"},
-		},
-		{
-			name: "backup with non-empty include and non-empty exclude namespaces",
-			backup: &velerov1.Backup{
-				Spec: velerov1.BackupSpec{
-					IncludedNamespaces: []string{"include-1", "include-2", "include-3", "include-4", "include-5"},
-					ExcludedNamespaces: []string{"exclude-1", "exclude-2", "exclude-3"},
-				},
-			},
-			expectedNamespaces: []string{"include-1", "include-2", "include-3", "include-4", "include-5"},
-		},
-	}
-
-	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) {
-			actualNamespaces := GetNamespacesInBackup(tc.backup)
-			assert.Equal(t, tc.expectedNamespaces, actualNamespaces)
-		})
-	}
-}
-
 // TestBackupWithRestic runs backups of pods that are annotated for restic backup,
 // and ensures that the restic backupper is called, that the returned PodVolumeBackups
 // are added to the Request object, and that when PVCs are backed up with restic, the
--- a/pkg/backup/backup_test.go
+++ b/pkg/backup/backup_test.go
@@ -2406,6 +2406,62 @@ func (b *fakeResticBackupper) BackupPodVolumes(backup *velerov1.Backup, pod *cor
 	return res, nil
 }
 
+func TestGetNamespacesInBackup(t *testing.T) {
+	testCases := []struct {
+		name               string
+		backup             *velerov1.Backup
+		expectedNamespaces []string
+	}{
+		{
+			name: "backup with nil include and exclude namespaces",
+			backup: &velerov1.Backup{
+				Spec: velerov1.BackupSpec{
+					IncludedNamespaces: nil,
+					ExcludedNamespaces: nil,
+				},
+			},
+			expectedNamespaces: []string{""}, // this signifies all namespaces
+		},
+		{
+			name: "backup with empty include and exclude namespaces",
+			backup: &velerov1.Backup{
+				Spec: velerov1.BackupSpec{
+					IncludedNamespaces: []string{},
+					ExcludedNamespaces: []string{},
+				},
+			},
+			expectedNamespaces: []string{""}, // this signifies all namespaces
+		},
+		{
+			name: "backup with non-empty include and empty exclude namespaces",
+			backup: &velerov1.Backup{
+				Spec: velerov1.BackupSpec{
+					IncludedNamespaces: []string{"include-1", "include-2", "include-3", "include-4", "include-5"},
+					ExcludedNamespaces: []string{},
+				},
+			},
+			expectedNamespaces: []string{"include-1", "include-2", "include-3", "include-4", "include-5"},
+		},
+		{
+			name: "backup with non-empty include and non-empty exclude namespaces",
+			backup: &velerov1.Backup{
+				Spec: velerov1.BackupSpec{
+					IncludedNamespaces: []string{"include-1", "include-2", "include-3", "include-4", "include-5"},
+					ExcludedNamespaces: []string{"exclude-1", "exclude-2", "exclude-3"},
+				},
+			},
+			expectedNamespaces: []string{"include-1", "include-2", "include-3", "include-4", "include-5"},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			actualNamespaces := GetNamespacesInBackup(tc.backup)
+			assert.Equal(t, tc.expectedNamespaces, actualNamespaces)
+		})
+	}
+}
+
 // TestBackupWithRestic runs backups of pods that are annotated for restic backup,
 // and ensures that the restic backupper is called, that the returned PodVolumeBackups
 // are added to the Request object, and that when PVCs are backed up with restic, the
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -28,6 +28,7 @@ import (
 	"sync"
 	"time"
 
+	snapshotterClientSet "github.com/kubernetes-csi/external-snapshotter/v2/pkg/client/clientset/versioned"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
@@ -688,6 +689,14 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
 	}
 
 	deletionControllerRunInfo := func() controllerRunInfo {
+		var snapshotterClient *snapshotterClientSet.Clientset
+		var err error
+		if features.IsEnabled(features.EnableCSI) {
+			snapshotterClient, err = snapshotterClientSet.NewForConfig(s.kubeClientConfig)
+			if err != nil {
+				s.logger.Errorf("Failed to create CSI volumesnapshotclient, %v", err)
+			}
+		}
 		deletionController := controller.NewBackupDeletionController(
 			s.logger,
 			s.sharedInformerFactory.Velero().V1().DeleteBackupRequests(),
@@ -700,6 +709,7 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
 			s.sharedInformerFactory.Velero().V1().PodVolumeBackups().Lister(),
 			s.sharedInformerFactory.Velero().V1().BackupStorageLocations().Lister(),
 			s.sharedInformerFactory.Velero().V1().VolumeSnapshotLocations().Lister(),
+			snapshotterClient,
 			newPluginManager,
 			s.metrics,
 		)
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -229,6 +229,7 @@ type server struct {
 	dynamicClient                       dynamic.Interface
 	sharedInformerFactory               informers.SharedInformerFactory
 	csiSnapshotterSharedInformerFactory *CSIInformerFactoryWrapper
+	csiSnapshotClient                   *snapshotterClientSet.Clientset
 	ctx                                 context.Context
 	cancelFunc                          context.CancelFunc
 	logger                              logrus.FieldLogger
@@ -301,6 +302,7 @@ func newServer(f client.Factory, config serverConfig, logger *logrus.Logger) (*s
 		dynamicClient:                       dynamicClient,
 		sharedInformerFactory:               informers.NewSharedInformerFactoryWithOptions(veleroClient, 0, informers.WithNamespace(f.Namespace())),
 		csiSnapshotterSharedInformerFactory: NewCSIInformerFactoryWrapper(csiSnapClient),
+		csiSnapshotClient:                   csiSnapClient,
 		ctx:                                 ctx,
 		cancelFunc:                          cancelFunc,
 		logger:                              logger,
@@ -689,14 +691,6 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
 	}
 
 	deletionControllerRunInfo := func() controllerRunInfo {
-		var snapshotterClient *snapshotterClientSet.Clientset
-		var err error
-		if features.IsEnabled(features.EnableCSI) {
-			snapshotterClient, err = snapshotterClientSet.NewForConfig(s.kubeClientConfig)
-			if err != nil {
-				s.logger.Errorf("Failed to create CSI volumesnapshotclient, %v", err)
-			}
-		}
 		deletionController := controller.NewBackupDeletionController(
 			s.logger,
 			s.sharedInformerFactory.Velero().V1().DeleteBackupRequests(),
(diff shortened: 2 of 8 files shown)


pkg/cmd/server/server.go
```diff
@@ -640,6 +641,14 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
 	}
 
 	deletionControllerRunInfo := func() controllerRunInfo {
+		var snapshotterClient *snapshotterClientSet.Clientset
+		var err error
+		if features.IsEnabled(features.EnableCSI) {
+			snapshotterClient, err = snapshotterClientSet.NewForConfig(s.kubeClientConfig)
```
skriss, line 647: once #2323 merges, you probably want to reuse the CSI snapshot client that is being instantiated [here](https://github.com/vmware-tanzu/velero/pull/2323/files#diff-cd1a009fa409db8496ea35b53f9776d2R286)

pkg/features/feature_flags.go
```diff
@@ -22,6 +22,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+// Feature flag constant strings
```
skriss, line 25: #2323 adds this constant (though in a different package, https://github.com/vmware-tanzu/velero/pull/2323/files#diff-193655ac4b146a95ba1bbfd621ad7132R41) so you'll want to use that one once it merges

pkg/controller/backup_deletion_controller.go
```diff
@@ -442,6 +458,49 @@ func (c *backupDeletionController) deleteResticSnapshots(backup *v1.Backup) []er
...
+			log.Debugf("Found %d CSI volumesnapshots in namespace %s", len(nsvs.Items), ns)
+			csiVolSnaps = append(csiVolSnaps, nsvs.Items...)
+		}
+	}
+
+	log.Infof("Found %d CSI volumesnapshots in backup %s", len(csiVolSnaps), b.Name)
+	return csiVolSnaps, errs
+}
+
+func (c *backupDeletionController) deleteCSIVolumeSnapshots(backup *velerov1api.Backup, csiClient snapshotter.SnapshotV1beta1Interface, log *logrus.Entry) []error {
```
skriss, line 490: I think you should be able to simplify this function by using the `.DeleteCollection(...)` method and passing a label selector -- we use it elsewhere inside this controller for deleting related CRs in one API call.

pkg/cmd/server/server.go
ashish-amarnath: 💯 I wanted to get this change tested out and did what was best to avoid being blocked on #2323

pkg/features/feature_flags.go
ashish-amarnath: 💯 I wanted to get this change tested out and did what was best to avoid being blocked on #2323

ashish-amarnath:
@skriss Yes, and with using CSI, we can be more storage provider agnostic and we can let the CSI driver, along w/ the common controller, decide how it wants to handle the volumesnapshot deletion.

skriss:
another random thought - should we be flipping the reclaim policy to "Delete" on the VolumeSnapshotContent objects right before deleting them? That would allow the underlying storage snapshots to be deleted as well. Otherwise, if they have a reclaim policy of "Retain" (which we are planning to recommend), there's no way that the storage snapshots are ever getting removed, and I do worry that if Velero is just endlessly creating snapshots and never deleting them, users will be unhappy.

Forgive me if we've already talked through this, I can only keep things in my memory for so long :)

nrb:
> another random thought - should we be flipping the reclaim policy to "Delete" on the VolumeSnapshotContent objects right before deleting them?

VSCs, like PVs, are immutable after creation, so unfortunately we can't just change the policy :(

ashish-amarnath:
@skriss We might have discussed this. But, I went through the same thought exercise when I wrote this code. The scenario I used to convince myself was this.
1. App is running.
2. Backup is created. This creates the `volumesnapshot` in the same namespace as the app and the `volumesnapshotcontents` in the cluster.
3. When disaster strikes and the namespace is deleted, the `volumesnapshot` gets deleted and with it the `volumesnapshotcontents` also get deleted, with the `DeletionPolicy` of `Delete`. This deletes the snapshot in the storage provider.
4. Now user wants to recover from the disaster ant tries to restore from the backup, but cannot recover their app because the `volumesnapshotcontent` is not gone.

nrb:
@skriss The communication we've received from the CSI snapshot design folks has been that the `Retain` policy is working as intended, and "something else" should delete those snapshots. It's not been quite clear to me what would clean up these snapshots, but I assume some sort of component in the storage system, external to Kubernetes.

Based on conversations with @xing-yang, the `Delete` and `Retain` policies me the needs of most designers, but I do wonder if there's some middle ground that we might be able to achieve for tools built on top of the snapshot objects.

For Velero's use cases, I think users can get stuck using either policy:

 * With a `Delete` policy, when a user deletes a namespace, the `VolumeSnapshot`'s deletion will trigger a deletion of the `VolumeSnapshotContent` object, which will also delete the storage snapshot. While that makes sense from the CSI controller's perspective, it breaks Velero's assumptions that storage snapshots outlive namespaces.
 * With a `Retain` policy, deletion of a `VolumeSnapshot` no longer causes the `VolumeSnapshotContent` to get deleted, which means the storage snapshot is not deleted by Kubernetes. However, this also currently means that the Velero `Backup`'s TTL field is not respected, and the storage snapshot cannot be deleted, causing storage to fill up over time unless the storage administrator specifically handles the snapshots outside of Kubernetes.

One workaround that I can think of, which I believe is more of a hack, is to create `VolumeSnapshotContent` with the same `snapshotHandle`, but with a `Delete` policy, and then delete it. This would remove the storage snapshot, but then leave the `Retain` VSC dangling (unless we did further manipulation by removing the finalizer and manually cleaning it up).

While I think this would work, I don't really want to advocate it, because it's using a dangerous loophole that appears to work on accident, not on purpose.

skriss:
> VSCs, like PVs, are immutable after creation, so unfortunately we can't just change the policy :(

That doesn't seem to be true, it worked fine for me when I tested it before posing this question (and I just reconfirmed) :)

nrb:
@skriss Good catch!

Looking at https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/apis/volumesnapshot/v1beta1/types.go#L274-L282, it appears this is allowed and not planned to be made immutable. Some fields are currently mutable, but planned to be made immutable.

If it's truly meant to be changed whenever, then simply changing the policy would negate my long-winded post 😄

skriss:
> But, I don't think flipping the Policy is the right thing to do, seems a little hacky even if the API allowed this.

I'm not sure I agree with this - the API field gives you control over what happens when the CR is deleted, and we want that to vary over time.  We don't want to lose the snapshot while the Backup exists, which is why it's in Retain mode initially. However, when we go do delete the Backup, we *do* want to delete the associated snapshots, which is why we flip it to Delete mode.  If this is a supported part of the API, then I don't see why we shouldn't use it to achieve our desired behavior.

ashish-amarnath:
If the API is allowing that behavior as a valid scenario, then that makes sense. The reason I found that to be hacky is, when there is a policy established at the time of snapshot creation I view it as a contract saying the VSC will not be deleted even if the VS is deleted. But, now when we want to delete the VS, flipping the deletion policy seems like violating the contract we made at the time of snapshotting.
As a benign user, I would expect this policy field to be immutable.

skriss:
As an aside, this is kind of the same reason why we implemented backup deletion the way we did in Velero - didn't want users to inadvertently delete their backups if they accidentally deleted the Velero namespace or a CR. So we made it an entirely separate CR (`DeleteBackupRequest`) that explicitly has to be created in order to trigger deletion. That approach has its own drawbacks, but it's mostly worked well in practice, I think.

nrb:
[Per a Slack conversation](https://kubernetes.slack.com/archives/C8EJ01Z46/p1587177423153400?thread_ts=1587161272.150900&cid=C8EJ01Z46), changing the policy after creation is a supported part of the API.

Therefore, I think toggling a VSC that has `Retain` over to `Delete` then deleting it is acceptable for Velero's use case when cleaning up VSCs that Velero created. It is _not_ acceptable for VSCs that Velero isn't managing, though.

ashish-amarnath:
Thanks for checking on this @nrb 
I've updated the PR accordingly. Need to take this through some testing.

pkg/controller/backup_deletion_controller.go
ashish-amarnath: Considering that we need to `setVSCDeletionPolicyToDelete` on the associated volumesnapshotcontent, I don't think we will be able to the `DeleteCollection(..)` API. Let me know if I am missing something here.

ashish-amarnath:
I've added some more unit tests and this PR looks good in testing.
Ready for 👀

pkg/controller/backup_deletion_controller.go
```diff
@@ -33,7 +36,8 @@ import (
 	kubeerrs "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/cache"
 
-	v1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	"github.com/vmware-tanzu/velero/pkg/backup"
```
skriss, line 40: this is already imported just below and aliased as `pkgbackup`

pkg/controller/backup_deletion_controller.go
```diff
@@ -442,6 +458,85 @@ func (c *backupDeletionController) deleteResticSnapshots(backup *v1.Backup) []er
...
+		return nil, errors.Wrapf(err, "failed to marshall existing volumesnapshotcontent %s json", vscName)
+	}
+
+	snapCont.Spec.DeletionPolicy = snapshotv1beta1api.VolumeSnapshotContentDelete
+	new, err := json.Marshal(snapCont)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to marshall updated volumesnapshotcontent %s to json", vscName)
+	}
+
+	pb, err := jsonpatch.CreateMergePatch(old, new)
```
skriss, line 506: In cases like this where the desired patch is known at compile time, I usually just hand-craft it rather than going through the whole diff process at runtime. In this case something like `{"spec":{"deletionPolicy":"Delete"}}` (I may have the field names not quite right) could replace L491-509

pkg/controller/backup_deletion_controller.go
ashish-amarnath: I missed that. Thanks!

pkg/controller/backup_deletion_controller.go
ashish-amarnath: That should make this much more simpler. Thanks for pointing that out 👍

pkg/controller/backup_deletion_controller.go
skriss: yeah, that sounds right. I wrote this comment before I thought about toggling the deletion policy.

pkg/controller/backup_deletion_controller.go
```diff
@@ -442,6 +457,66 @@ func (c *backupDeletionController) deleteResticSnapshots(backup *v1.Backup) []er
 	return errs
 }
 
+func getCSIVolumeSnapshotsInBackup(b *velerov1api.Backup, csiClient snapshotter.SnapshotV1beta1Interface, log *logrus.Entry) ([]snapshotv1beta1api.VolumeSnapshot, []error) {
```
skriss, line 460: nit: I'd prefer `backup` over `b` as a variable name

pkg/controller/backup_deletion_controller.go
```diff
@@ -442,6 +457,66 @@ func (c *backupDeletionController) deleteResticSnapshots(backup *v1.Backup) []er
...
+	}
+	listOpts := metav1.ListOptions{
+		LabelSelector: metav1.FormatLabelSelector(&selector),
+	}
+
+	for _, ns := range pkgbackup.GetNamespacesInBackup(b) {
+		log.Debugf("Listing CSI volumesnapshots in namespace %s", ns)
+		nsvs, err := csiClient.VolumeSnapshots(ns).List(listOpts)
+		if err != nil {
+			errs = append(errs, err)
```
skriss, line 477: `errs = append(errs, errors.WithStack(err))` 

Using errors.WithStack, errors.Wrap, etc. will add a stack trace to the error. We generally wrap any errors that come back from third-party libraries, since they presumably won't have the stack info, but it's unnecessary to wrap errors returned from Velero-internal code since presumably they're already have stack info.

pkg/controller/backup_deletion_controller.go
```diff
@@ -442,6 +457,66 @@ func (c *backupDeletionController) deleteResticSnapshots(backup *v1.Backup) []er
...
+		}
+		if nsvs.Items != nil && len(nsvs.Items) != 0 {
```
skriss, line 479: you should be able to drop the nil-check, since the length of a nil slice will be 0, so just `if len(nsvs.Items > 0 {`

pkg/controller/backup_deletion_controller.go
```diff
@@ -442,6 +457,66 @@ func (c *backupDeletionController) deleteResticSnapshots(backup *v1.Backup) []er
...
+			log.Debugf("Found %d CSI volumesnapshots in namespace %s", len(nsvs.Items), ns)
+			csiVolSnaps = append(csiVolSnaps, nsvs.Items...)
+		}
+	}
+
+	log.Infof("Found %d CSI volumesnapshots in backup %s", len(csiVolSnaps), b.Name)
+	return csiVolSnaps, errs
+}
+
+func setVSCDeletionPolicyToDelete(vscName string, csiClient snapshotter.SnapshotV1beta1Interface) (*snapshotv1beta1api.VolumeSnapshotContent, error) {
```
skriss, line 489: the value of this helper is dubious now that it's just 2 lines, I would consider dropping it and just inlining this code where it's used.

skriss:
@ashish-amarnath last request - please add a changelog file!

ashish-amarnath:
@nrb or @carlisia requesting  👀

pkg/backup/backup_test.go
```diff
@@ -2346,6 +2346,62 @@ func (b *fakeResticBackupper) BackupPodVolumes(backup *velerov1.Backup, pod *cor
...
 
+func TestGetNamespacesInBackup(t *testing.T) {
+	testCases := []struct {
+		name               string
+		backup             *velerov1.Backup
+		expectedNamespaces []string
+	}{
+		{
+			name: "backup with nil include and exclude namespaces",
+			backup: &velerov1.Backup{
```
nrb, line 2357: Could these use the builders we have for backups in pkg/builder?

pkg/cmd/server/server.go
```diff
@@ -228,6 +229,7 @@ type server struct {
 	dynamicClient                       dynamic.Interface
 	sharedInformerFactory               informers.SharedInformerFactory
 	csiSnapshotterSharedInformerFactory *CSIInformerFactoryWrapper
+	csiSnapshotClient                   *snapshotterClientSet.Clientset
```
nrb, line 232: Why use the client here instead of using the informer factory & lister? It seems like we're introducing 2 ways of doing the same thing in different controllers.

pkg/controller/backup_deletion_controller.go
```diff
@@ -442,6 +457,62 @@ func (c *backupDeletionController) deleteResticSnapshots(backup *v1.Backup) []er
...
+	selector := metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			velerov1api.BackupNameLabel: label.GetValidName(backup.Name),
+		},
+	}
+	listOpts := metav1.ListOptions{
+		LabelSelector: metav1.FormatLabelSelector(&selector),
+	}
+
+	for _, ns := range pkgbackup.GetNamespacesInBackup(backup) {
```
nrb, line 473: By using the informer/lister, I think this loop would be unnecessary.

See https://github.com/vmware-tanzu/velero/blob/master/pkg/cmd/server/server.go#L897-L899 and its use at 
https://github.com/vmware-tanzu/velero/blob/master/pkg/controller/backup_controller.go#L565

pkg/backup/backup_test.go
ashish-amarnath: I always forget the builder package.

pkg/cmd/server/server.go
ashish-amarnath: In the deletion controller, in addition to listing volumesnapshots  we also want to patch volumesnapshotcontents, and delete volumesnapshots. I wasn't able to use the informer factory APIs to do that or to get it to expose the underlying client.
That's the reason I chose to embed the client in the server struct.

However, we can embed the client in the server struct and use that to create the informer factory in the backup controller. I was going to make that refactor in a different PR.

nrb: Got it, thanks. Could you please add a comment explaining that, just so it's more immediately findable by reading the code?

I don't know if we want to create the informer factory in the controller. There's some warnings about that in the server.go file about doing that kind of thing, but it's possible that could be an optimization.

pkg/controller/backup_deletion_controller.go
ashish-amarnath: Because the informer/lister wasn't made available in this controller chose to implement this loop.

pkg/cmd/server/server.go
ashish-amarnath: Yeah will add comments.

skriss: Nolan's right that it would be more efficient to use the lister for the query inside the deletion controller. However, you will also need the client in order to do the patching & deletion. So it's not an either/or, but a "both".

You wouldn't want to create a new informer factory inside the controller - use the existing factory in `server.go` to get a VS/VSC lister, and pass those into the constructor for the controller, along with the client.

ashish-amarnath: Makes sense.
But this is so hard to write unit tests for!
```
# github.com/vmware-tanzu/velero/pkg/controller
import cycle not allowed in test
package github.com/vmware-tanzu/velero/pkg/controller (test)
	imports github.com/vmware-tanzu/velero/pkg/cmd/server
	imports github.com/vmware-tanzu/velero/pkg/controller
```

ashish-amarnath:
Not holding up to tests.
[getCSIVolumeSnapshotsInBackup](https://github.com/vmware-tanzu/velero/pull/2411/files#diff-afe709fa1785e08707ca0825e3f60e68R465) function doesn't seem to be listing the volumesnapshots in the backup. But it seems to be doing the right thing in the unit tests.
What am I missing with wrt the informers?

ashish-amarnath:
@skriss and @nrb this is ready for review.

carlisia:
Not merging, leaving that for @nrb after his review.

PR #2439, opened 2020-04-15
change dynamic client's List return type to UnstructuredList

skriss:

minor refactoring to change our dynamic client wrapper to return the same type as the upstream dynamic client, so we don't have to roundtrip to a `runtime.Object` and back again.


skriss:
yeah, so the kubernetes dynamic client returns an `*unstructured.UnstructuredList`, but for some reason our wrapper was returning it as a `runtime.Object` (which the former implements), but is a really general, useless type. So in the place where we use the List method, we ended up effectively having to convert it back to something like an `*unstructured.UnstructuredList`.  Basically, there's no reason for our wrapper to return the generic `runtime.Object` -- it doesn't provide any benefit - so this is just simpler.

PR #2440, opened 2020-04-16
report backup progress

### LINKED ISSUE #20: Backup progress

skriss:
closes #20 

I'm still testing here, but an early review on the approach would be helpful.

- The first commit collapses the group backupper and resource backupper into a single type, since there was a lot of unnecessary abstraction and repetition between them.
- The second commit changes the backup process to first get *all* items matching the backup spec from the Kubernetes API, to write them to temp files on disk, then to pass them to the item backupper for backup one-by-one. This enables us to count them ahead of time and report "x of y items backed up" as we're going.
- The third commit adds periodic writing of progress to the Backup CR's status.


pkg/backup/backup.go
```diff
@@ -257,30 +262,147 @@ func (kb *kubernetesBackupper) Backup(log logrus.FieldLogger, backupRequest *Req
...
+		dynamicFactory:          kb.dynamicFactory,
+		discoveryHelper:         kb.discoveryHelper,
+		resticBackupper:         resticBackupper,
+		resticSnapshotTracker:   newPVCSnapshotTracker(),
+		volumeSnapshotterGetter: volumeSnapshotterGetter,
+		itemHookHandler: &defaultItemHookHandler{
+			podCommandExecutor: kb.podCommandExecutor,
+		},
+	}
+	itemBackupper.additionalItemBackupper = itemBackupper
```
skriss, line 297: TODO we should be able to get rid of the `additionalItemBackupper` field, since it was there only for testing purposes and isn't needed anymore.

pkg/backup/backup.go
```diff
@@ -257,30 +262,147 @@ func (kb *kubernetesBackupper) Backup(log logrus.FieldLogger, backupRequest *Req
...
+	// back up CRD for resource if found. We should only need to do this if we've backed up at least
+	// one item for the resource and IncludeClusterResources is nil. If IncludeClusterResources is false
+	// we don't want to back it up, and if it's true it will already be included.
+	if backupRequest.Spec.IncludeClusterResources == nil {
+		for gr := range backedUpGroupResources {
+			kb.backupCRD(log, gr, itemBackupper)
 		}
 	}
 
+	log.WithField("progress", "").Infof("Backed up a total of %d items", len(backupRequest.BackedUpItems))
```
skriss, line 342: TODO fine-tune the language here; the total # of backed-up items may differ from the # of items we "processed" due to:
- additional items being returned from plugins
- exclude-from-backup label applied to a resource
- maybe some other cases too

pkg/backup/item_backupper.go
```diff
@@ -33,62 +33,16 @@ import (
...
-	}
-
-	// this is for testing purposes
-	ib.additionalItemBackupper = ib
-
-	return ib
-}
-
+// ItemBackupper can back up individual items to a tar writer.
 type ItemBackupper interface {
```
skriss, line 46: TODO this interface probably isn't needed anymore since I've removed the factory, and am going to remove the additionalItemBackupper field.

pkg/backup/item_collector.go
```diff
@@ -0,0 +1,351 @@
...
+
+	"github.com/vmware-tanzu/velero/pkg/client"
+	"github.com/vmware-tanzu/velero/pkg/discovery"
+	"github.com/vmware-tanzu/velero/pkg/kuberesource"
+	"github.com/vmware-tanzu/velero/pkg/util/collections"
+)
+
+// itemCollector collects items from the Kubernetes API according to
+// the backup spec and writes them to files inside dir.
+type itemCollector struct {
```
skriss, line 40: unfortunately the GH diff doesn't show it, but this file is largely a combination of the old `group_backupper.go` and `resource_backupper.go`. It may be easier to see if you review individual commits.

nrb:
I'll do a more detailed review of this still, but have you done any investigation of using Kubernetes Events for emitting status progress?

My main concern with that approach would be that we're doubling the number of kube objects (1 event per kube object encountered), but I'm not sure if that's an issue or not.

It seems like getting updates with that approach could then be mostly handled on the client side, doing a request for Events based on a label or something.

skriss:
> I'll do a more detailed review of this still, but have you done any investigation of using Kubernetes Events for emitting status progress?

I have not. Worth considering if/how we could emit some information as events, though.

> My main concern with that approach would be that we're doubling the number of kube objects (1 event per kube object encountered), but I'm not sure if that's an issue or not.

Yeah, and this would be every time you run a backup, so potentially a *lot* of events. There may be a less fine-grained way to emit them that's still useful, but I'd need to dig into it some more if that's something we want to look at.

nrb:
>Yeah, and this would be every time you run a backup, so potentially a lot of events.

Yep, and conversely, a lot of events on restores when we get there, too.

These may be useful at particular breakpoints - every x% of the initial count? Maybe 10%? 25% increments seem too big to be all that informative. I suppose we could use absolute numbers too, though percentages give us a reliable number of events per backup.

skriss:
I'm gonna have to rebase after https://github.com/vmware-tanzu/velero/pull/2373 gets merged.

skriss:
I pushed a very basic implementation of writing the "progress" info to new fields on the backup's status. It works fine (you can `watch kubectl -n velero get backup <NAME> -o yaml` while running a backup to watch it), though it does slow down the backup since we're patching the backup after *every* item is backed up. We'll want to only update periodically.

I do want to solicit feedback on the naming of the fields -- it's a little tricky since the # of items initially identified as matching the backup spec will often not be the total # of items ultimately backed up. Naming/wording suggesting welcome.

skriss:
pushed some more updates to output progress info to `velero backup describe`, and also to move the code that patches the backup CR with progress updates to a separate goroutine that triggers a patch at most once per second.

I think this is now in pretty good shape, but I want to do some more testing. And will still need to resolve conflicts with #2373

skriss:
@ashish-amarnath @carlisia @nrb this is ready for review.  Note that I'm going to keep it as draft for now because I know this will conflict with at least #2373, so I'll wait until that merges and do the rebase, but I'm done coding and testing has looked good.

skriss:
From the community call this AM, it sounded like most folks liked Scott's idea of continually updating "items processed" throughout the backup to account for any additional items, exclusions, etc. I'll look at making changes to do that. I think it means that we end up with only two status fields, but both will continually update throughout the backup.

skriss:
I pushed a commit that makes the change discussed; "total items" is now an estimate that continually changes as the backup executes. Still need to make some wording changes in the log statements, but I've tested this change and it seems to do what it's intended to.

carlisia:
I'll review this tomorrow.

pkg/cmd/util/output/backup_describer.go
```diff
@@ -238,6 +238,22 @@ func DescribeBackupStatus(d *Describer, backup *velerov1api.Backup, details bool
...
+	if backup.Status.Progress != nil {
+		if backup.Status.Phase == velerov1api.BackupPhaseInProgress {
+			// if the backup's in progress, these two numbers are most meaningful
+			// for showing progress, and they avoid getting into the nuances of
+			// "items processed" vs. "items backed up".
+			d.Printf("Total items to be backed up:\t%d\n", backup.Status.Progress.TotalItems)
+			d.Printf("Items backed up so far:\t%d\n", backup.Status.Progress.ItemsBackedUp)
+		} else {
+			// if no longer in progress, just show total # of items backed up since
+			// we just want to summarize what was done.
```
carlisia, line 250: Maybe unpopular opinion :) : If it was me, and you ***are*** giving me these stats, then I'd also want to see the predicted number of items to be backed up, so I can tell if you really backed up all of it. Basically I want to know it backed up 100% of the items. Obviously if the backup was successful this is implied, but I'm thinking if we could help inform that in this context. Maybe adding something like "100% of the items backuped" would be helpful?

pkg/cmd/util/output/backup_describer.go
```diff
@@ -238,6 +238,22 @@ func DescribeBackupStatus(d *Describer, backup *velerov1api.Backup, details bool
 	d.Printf("Expiration:\t%s\n", status.Expiration)
 	d.Println()
 
+	if backup.Status.Progress != nil {
+		if backup.Status.Phase == velerov1api.BackupPhaseInProgress {
+			// if the backup's in progress, these two numbers are most meaningful
+			// for showing progress, and they avoid getting into the nuances of
+			// "items processed" vs. "items backed up".
+			d.Printf("Total items to be backed up:\t%d\n", backup.Status.Progress.TotalItems)
```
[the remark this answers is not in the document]
skriss: :+1: yeah I think that's better, and I'll do similar in the log statements

pkg/cmd/util/output/backup_describer.go
skriss: Yeah I think that makes sense, this was more relevant when I had the three different fields and was attempting to avoid showing mismatched numbers side-by-side to avoid confusion. I think I'll just leave both fields, even if it's completed, I may just relabel the one to not say "estimated" since by the time we're done, we know exactly how many there were.

skriss:
I just rebased this on top of Rafael's changes - tests are passing so I'm cautiously optimistic I didn't break anything in the rebase :)

Gonna address @carlisia's comments and fix up a few other things.

pkg/cmd/util/output/backup_describer.go
carlisia: Yes, perfect.

skriss:
comments addressed & commits cleaned up

pkg/apis/velero/v1/backup.go
```diff
@@ -273,6 +273,26 @@ type BackupStatus struct {
 	// file in object storage.
 	// +optional
 	Errors int `json:"errors,omitempty"`
+
+	// Progress contains information about the backup's execution progress. Note
+	// that this information is best-effort only -- if Velero fails to update it
+	// during a backup for any reason, it may be inaccurate/stale.
+	Progress *BackupProgress `json:"progress,omitempty"`
```
nrb, line 280: A note for us - this will require users to install updated CRDs with v1.4.0. We'll need to put that in our docs.

pkg/backup/backup.go
```diff
@@ -264,30 +274,222 @@ func (kb *kubernetesBackupper) Backup(log logrus.FieldLogger, backupRequest *Req
...
+		for {
+			select {
+			case <-quit:
+				ticker.Stop()
+				return
+			case val := <-update:
+				lastUpdate = &val
+			case <-ticker.C:
+				if lastUpdate != nil {
+					patch := fmt.Sprintf(`{"status":{"progress":{"totalItems":%d,"itemsBackedUp":%d}}}`, lastUpdate.totalItems, lastUpdate.itemsBackedUp)
```
nrb, line 348: Have you tested this against a cluster that had older backup CRs, without the field, installed? What was the failure mode? I think that would be useful to document for folks, since I think a lot of people will just update their image and not read the upgrade doc.

pkg/apis/velero/v1/backup.go
skriss: Yeah, good callout, I was thinking about that. I know we have the `--crds-only` option so I think we can instruct users to use that, dump it to YAML, and `kubectl apply` the output. I think we've done this before.

pkg/backup/backup.go
skriss: I have not tested that but will.

skriss:
Happy to walk through the code over Zoom as well, if that's helpful.

pkg/apis/velero/v1/backup.go
ashish-amarnath: What would be the failure if someone doesn't upgrade their CRDs?
May be out of scope for this PR, but we might want to detect that and fail with a "upgrade your CRDs" message than a crashloop

pkg/backup/backup.go
ashish-amarnath: I think I made a similar comment :)
How did the testing go?

ashish-amarnath:
Thank you for grouping your commits!
Overall LGTM. Couple of questions around failure modes and results of testing.

pkg/apis/velero/v1/backup.go
nrb: There was this [older one that we had](https://github.com/vmware-tanzu/velero/blob/master/pkg/cmd/server/server.go#L408), but it's clearly out of date now as it talks about the YAML files we had in-tree. Plus, the condition that it looks for is only if the API group exists, no that relevant fields are found.

In general though, I'm 👍  to some sort of utility function that does a field-level verification for CRD compat. Ideally we'd be doing SemVer and do a patch release with added fields to make it really fast.

pkg/backup/backup.go
skriss: I'll test it out this afternoon and will report back!

skriss: OK testing that was easier than I thought :) The behavior if the CRDs have not been updated is that the progress updates to the CR are a no-op. I believe this is because `preserveUnknownFields` is false, so the updates to those unknownf fields are pruned by the API server. I think this is the desirable behavior for this feature.

pkg/apis/velero/v1/backup.go
skriss: We do compile the CRD manifests in (for the purposes of `velero install`), so I imagine we could use those to diff against what's in-cluster. I agree that we can tackle that in a separate issue/PR, especially since this feature has reasonable behavior IMHO if the CRDs are out of date (see 1 minute ago).

pkg/backup/backup.go
ashish-amarnath: Yeah, that is desirable. But we don't want folks not to update their CRDs. I think that can be solved separately.

pkg/apis/velero/v1/backup.go
nrb: #2509 captures the upgrade instructions, which is all that's needed for this PR.

PR #2448, opened 2020-04-20
Include CSI volume snapshot information in `velero backup describe`

### LINKED ISSUE #1568: Include CSI snapshot details in backup describe command

nrb:
Closes #1568 

Test scenarios I can think of:

 1. A backup was just made, and the CSI objects exist in the cluster currently.
 1. A backup was synced into the cluster, but the CSI objects are not yet restored, so must be read in from object storage.

I think in order to cover all the bases, it'll be best to always pull from the information from object storage.


nrb:
Since #2496 is syncing the VSCs into a backup for sure, I think it makes sense for that to get merged in, and we can use the in-cluster VSCs for the describer.

I _believe_ these can reliably be used for the describe output, but if not, they can at the very least be used to give signal for whether or not there are CSI snapshots associated with a backup before querying object storage.

nrb:
Current state of the PR should work against a cluster that has just created a back up, and should work with synced VSCs once 2496 is merged.

If we go with how the describe code is working now, I can just remove all the extra code added for adding downloads from object storage.

pkg/csi/common.go
```diff
@@ -0,0 +1,44 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	"github.com/vmware-tanzu/velero/pkg/label"
+)
+
+// NewCSIListOptions returns a ListOptions for matching CSI objects based on the Velero backup name and label.
+func NewCSIListOptions(name string) metav1.ListOptions {
+	return metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s", velerov1api.BackupNameLabel, label.GetValidName(name)),
```
nrb, line 31: This could potentially be a bug - we're not using the `label.GetValidName` function when we list the objects for serialization. I haven't checked to see if we use it when labeling the objects, but we should use it everywhere in the codebase

skriss:
I think using the in-cluster objects is a reasonable approach, especially since it's consistent with how PodVolumeBackups work.

pkg/csi/common.go
skriss: Yeah, we should probably do an audit there and update wherever needed...IIRC names can be longer than label values, so any time you get a long backup name, it may not be valid as a label value.

nrb: I'll update the core Velero values in this PR and get another one in for the plugin.

Across the codebase, we'll want to audit and do a comparison for existing objects to see if the label name matches one generated by the function, then update the value.

nrb:
Yeah, honestly either approach would be consistent with one of the two existing objects, PodVolumeBackups or Velero's native snapshots.

I think, since we are syncing VSCs now, that this is the easier method. It also avoids any sort of object storage call.

I'll clean this up a little bit more and mark it ready for review today..

ashish-amarnath:
To make the in-cluster backup more reliable, we could add two more backup phases `sync-ing` and `sync-ed` and use those in the describe command to indicate velero is still learning about this backup and may have more accurate information once the backup is sync-ed.
But for now, I agree, we can use the in-cluster objects for this.

pkg/cmd/util/output/backup_describer.go
```diff
@@ -243,6 +254,7 @@ func DescribeBackupStatus(d *Describer, backup *velerov1api.Backup, details bool
 		d.Println()
 	}
 
+	// TODO(nrb-csi): Should CSI snapshots increment VolumeSnapshotsAttempted? If so, we'll need a way to differentiate here.
 	if status.VolumeSnapshotsAttempted > 0 {
 		if !details {
 			d.Printf("Persistent Volumes:\t%d of %d snapshots completed successfully (specify --details for more information)\n", status.VolumeSnapshotsCompleted, status.VolumeSnapshotsAttempted)
```
skriss, line 260: we probably want to try to disambiguate here between non-CSI and CSI snapshots - maybe instead of "Persistent Volumes:" we say "Velero-Native Snapshots" or "non-CSI Snapshots" or something like that

skriss: (note this applies to various lines in this section, not just this one)

pkg/cmd/util/output/backup_describer.go
```diff
@@ -467,3 +479,43 @@ func (v *volumesByPod) Sorted() []*podVolumeGroup {
...
 	return v.volumesByPodSlice
 }
+
+func DescribeCSIVolumeSnapshots(d *Describer, details bool, volumeSnapshotContents []snapshotv1beta1api.VolumeSnapshotContent) {
+	if !features.IsEnabled(velerov1api.CSIFeatureFlag) {
+		return
+	}
+
+	if !details {
+		d.Printf("CSI Volume Snapshots (specify --details for more information)\n")
```
skriss, line 489: Maybe we should display a simple count of CSI snapshots here, if `details == false`, similarly to what's being done for non-CSI snapshots.

pkg/csi/common.go
```diff
@@ -0,0 +1,44 @@
...
+
+// NewListOptions returns a ListOptions for matching CSI objects based on the Velero backup name and label.
+func NewListOptions(name string) metav1.ListOptions {
+	return metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s", velerov1api.BackupNameLabel, label.GetValidName(name)),
+	}
+}
+
+// SumSnapshotSize sums up the total restore size (in bytes) of a given slice of VolumeSnapshotContent objects.
+func SumSnapshotSize(vscs []*snapshotv1beta1api.VolumeSnapshotContent) int64 {
```
skriss, line 36: this doesn't look like it's being used anywhere currently, were you intending to?

nrb: I was thinking I would, but maybe not here. It's probably more useful for metrics, so I can delete it.

pkg/cmd/util/output/backup_describer.go
nrb: Yeah, I should probably comb through here and revise the wording every where. I did it in the case that there were 0 velero-native snapshots, but nowhere else.

pkg/csi/common.go
ashish-amarnath: Created https://github.com/vmware-tanzu/velero/issues/2515

skriss:
updated wording looks reasonable to me.

nrb:
Ah shoot, merge conflicts.

nrb:
>To make the in-cluster backup more reliable, we could add two more backup phases sync-ing and sync-ed and use those in the describe command to indicate velero is still learning about this backup and may have more accurate information once the backup is sync-ed.

What would Velero still be learning about the backup?

pkg/cmd/util/output/backup_describer.go
```diff
@@ -87,6 +91,13 @@ func DescribeBackup(
 			d.Println()
 			DescribePodVolumeBackups(d, podVolumeBackups, details)
 		}
+		// TODO(nrb-csi): Put the CSI description here? Or down with the DescribeBackupStatus?
```
skriss, line 94: I *think* this TODO and the below "scenarios" comment can be removed now, but correct me if I'm wrong.

pkg/cmd/util/output/backup_describer.go
```diff
@@ -467,3 +479,40 @@ func (v *volumesByPod) Sorted() []*podVolumeGroup {
...
+	}
+}
+
+func DescribeVSC(d *Describer, details bool, vsc snapshotv1beta1api.VolumeSnapshotContent) {
+	if vsc.Status == nil {
+		d.Printf("Volume Snapshot Content %s cannot be described because its status is nil\n", vsc.Name)
+		return
+	}
+
+	d.Printf("Snapshot Content ID: %s\n", vsc.Name)
```
skriss, line 505: any particular reason to call it "ID" rather than "Name"?

pkg/csi/common.go
```diff
@@ -0,0 +1,32 @@
+/*
```
skriss, line 1: I guess this file can be dropped now in favor of the helpers in `pkg/label`

pkg/cmd/cli/backup/describe.go
```diff
@@ -78,7 +83,24 @@ func NewDescribeCommand(f client.Factory, use string) *cobra.Command {
...
+				// declare vscList up here since it may be empty and we'll pass the empty Items field into DescribeBackup
+				vscList := new(snapshotv1beta1api.VolumeSnapshotContentList)
+				if features.IsEnabled(velerov1api.CSIFeatureFlag) {
+					clientConfig, err := f.ClientConfig()
+					cmd.CheckError(err)
+
+					csiClient, err = snapshotv1beta1client.NewForConfig(clientConfig)
+					cmd.CheckError(err)
+
+					vscOpts := csi.NewListOptions(backup.Name)
```
ashish-amarnath, line 96: We can use the [`NewListOptionsForBackup`](https://github.com/vmware-tanzu/velero/blob/master/pkg/label/label.go#L59) function here.

nrb: Yeah, I need to rebase. This was done before I made that commit.

pkg/cmd/util/output/backup_describer.go
nrb: No, not really. Updating.

skriss:
Just needs a changelog, otherwise looks good to me

pkg/cmd/cli/backup/describe.go
```diff
@@ -78,7 +82,23 @@ func NewDescribeCommand(f client.Factory, use string) *cobra.Command {
...
+				// declare vscList up here since it may be empty and we'll pass the empty Items field into DescribeBackup
+				vscList := new(snapshotv1beta1api.VolumeSnapshotContentList)
+				if features.IsEnabled(velerov1api.CSIFeatureFlag) {
+					clientConfig, err := f.ClientConfig()
+					cmd.CheckError(err)
+
+					csiClient, err = snapshotv1beta1client.NewForConfig(clientConfig)
+					cmd.CheckError(err)
+
+					vscList, err = csiClient.SnapshotV1beta1().VolumeSnapshotContents().List(opts)
```
ashish-amarnath, line 95: `opts` is not defined?

ashish-amarnath: nvm

pkg/csi/doc.go
```diff
@@ -0,0 +1,18 @@
+/*
```
ashish-amarnath, line 1: Do we need this file?

nrb: Not anymore!

PR #2450, opened 2020-04-22
Skipping validation for volumesnapshotlocation for backup if snapshotvolume set to false

mynktl:
Changes:
- If backup has snapshotvolume then `validateAndGetSnapshotLocations` skip validating VSL and returns empty VSL list with success.

Existing behavior:
- if velero is configured with multiple VSL with the same provider but no default VSL configured, then one needs to provide VSL during backup creation, with snapshotvolume disabled, otherwise backup will fail with status.phase `FailedValidation`. 

If one wants to use restic for the volume backup then he/she doesn't need to provide VSL for the backup.

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -51,6 +51,7 @@ import (
 	"github.com/vmware-tanzu/velero/pkg/metrics"
 	"github.com/vmware-tanzu/velero/pkg/persistence"
 	"github.com/vmware-tanzu/velero/pkg/plugin/clientmgmt"
+	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
 	"github.com/vmware-tanzu/velero/pkg/util/collections"
 	"github.com/vmware-tanzu/velero/pkg/util/encode"
 	kubeutil "github.com/vmware-tanzu/velero/pkg/util/kube"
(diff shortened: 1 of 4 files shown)


pkg/controller/backup_controller.go
```diff
@@ -397,10 +397,16 @@ func (c *backupController) prepareBackupRequest(backup *velerov1api.Backup) *pkg
 // - a given provider's default location name is added to .spec.volumeSnapshotLocations if one
 //   is not explicitly specified for the provider (if there's only one location for the provider,
 //   it will automatically be used)
+// if backup has snapshotVolume disabled then it returns empty VSL
 func (c *backupController) validateAndGetSnapshotLocations(backup *velerov1api.Backup) (map[string]*velerov1api.VolumeSnapshotLocation, []string) {
 	errors := []string{}
 	providerLocations := make(map[string]*velerov1api.VolumeSnapshotLocation)
 
+	// if snapshotVolume is set to false then we don't need to validate volumesnapshotlocation
+	if backup.Spec.SnapshotVolumes != nil && !*backup.Spec.SnapshotVolumes {
```
ashish-amarnath, line 406: We have some helper functions at https://github.com/vmware-tanzu/velero/blob/master/pkg/util/boolptr/boolptr.go to handle this.

mynktl: done

mynktl:
Hi @ashish-amarnath 
I've updated the changes according to review comment. PTAL.
Thanks!

Issue #2116, opened 2019-12-10
[Epic] add support for restore hooks


marccampbell:
**Describe the problem/challenge you have**
When backing up some volumes, we have a pre and post backup hook implemented. This is useful when we have a process that has a lock on a file and we need to copy data between volumes to ensure that the data Velero is backing up will be uncorrupted and usable. But on a restore, there is no hook to execute a pre or post restore script. This makes automating a restore a little difficult because we need to rely on a separate set of tools when restoring vs when backing up.

**Describe the solution you'd like**
One possible solution to this is to add restore hooks, so they can be specified similar to how backup hooks are specified. If we are restoring data that was collected using a pre-hook, then we could write the reverse post-restore hook in the configuration to put the data back into the format or restart a process.

konvergence:
For a postgres dump done with a pre-hook backup : `pg_dumpall --clean > ${DATA_FOLDER}/dumpall.out`
it will usefull to run a  restore post hook to  run a `psql < ${DATA_FOLDER}/dumpall.out`

konvergence:
If there was  post hook for restore I could use it to restore a dump previously done with pre hook backup.

Example if we have this annotation for pre hook backup

```
kubectl -n bitamani-postgres-11-hook  annotate pod/postgres-hook-postgresql-0   backup.velero.io/backup-volumes=data --overwrite
kubectl -n bitamani-postgres-11-hook  annotate pod/postgres-hook-postgresql-0 pre.hook.backup.velero.io/timeout=3600s --overwrite
kubectl -n bitamani-postgres-11-hook  annotate pod/postgres-hook-postgresql-0   pre.hook.backup.velero.io/command='["bash", "-c", "mkdir -p /bitnami/postgresql/dumps; eval export PGPASSWORD=${POSTGRES_PASSWORD}; pg_dumpall --clean --username=postgres | gzip -4 > /bitnami/postgresql/dumps/dumpall.out.gz" ]' --overwrite

```

the following post hook restore could be used to restore the dump

```
kubectl -n bitamani-postgres-11-hook  annotate pod/postgres-hook-postgresql-0 post.hook.restore.velero.io/timeout=3600s --overwrite
[...]

Issue #1981, opened 2019-10-18
Restic volume not restored


Stolr:
**What steps did you take and what happened:**

I'm trying to restore a restic volume.

My backup got 2 volumes with 2 deployments

Backup 

```

tools-bitbucket-backup-prv2   Completed                    2019-10-18 08:45:44 +0200 CEST   29d       cluster-tools      <none>


Persistent Volumes: <none included>
Restic Backups:
  Completed:
    ok101-bitbucket-pr/bitbucket-postgresql-5-vvwsm: bitbucket-postgresql-data
    ok101-bitbucket-pr/bitbucket-server-14-ddrpp: bitbucket-server-data
```

When I make a restore from this backup , The postgres pod is restore properly but not the bitbucket server
```
tools-bitbucket-backup-prv2-20191018090027   tools-bitbucket-backup-prv2   InProgress   0          0        2019-10-18 09:00:27 +0200 CEST   <none>
 velero restore describe tools-bitbucket-backup-prv2-20191018090027

Restic Restores:
  Completed:
    ok101-bitbucket-pr/bitbucket-postgresql-5-vvwsm: bitbucket-postgresql-data
  New:
    ok101-bitbucket-pr/bitbucket-server-14-ddrpp: bitbucket-server-data
```

The init container is not created on the bitbucket-server pod , so the restic stay stuck in "New" phase but the pod is created and  running. It shouldn't

```
 kubectl get po
NAME                                        READY     STATUS             RESTARTS   AGE
bitbucket-server-15-glk7c                   1/1       Running            0          5m
```

*** Restic log ***
```
time="2019-10-18T07:00:48Z" level=debug msg="Pod is not running restic-wait init container, not enqueuing restores for pod" controller=pod-volume-restore key=ok24-velero/velero-7f5d784896-l66m7 logSource="pkg/controller/pod_volume_restore_controller.go:170"
time="2019-10-18T07:00:48Z" level=debug msg="Pod is not running restic-wait init container, not enqueuing restores for pod" controller=pod-volume-restore key=ok24-velero/velero-7f5d784896-l66m7 logSource="pkg/controller/pod_volume_restore_controller.go:170"
[...]

Issue #1150, opened 2019-01-09
Pre/Post Backup & Pre/Post Restore Plugins


king-jam:
**Describe the solution you'd like**
I'd like to be able to use the same plugin framework but guarantee ordering based on the registration call within the plugin framework. This is for scenarios where "rehydration" of data is required via an application specific hook.

The current example that I'm working with is backup/restore of a MySQL environment. PV Snapshots only provide a crash consistent approach whereas leveraging `mysqldump` provides a point-in-time application consistent backup. The existing plugin framework does not encounter issues during backup but during restore is unable to guarantee that the plugin will execute after the service endpoint is up.

The workaround is to use the restore plugin action as a timing sequence to start a separate job that can remotely rehydrate the database but that results in the ark restore reporting success before the database is actually restored. This could create an issue with how statuses are conveyed to a user or an automation system that wraps the ark CLI.

**Anything else you would like to add:**
N/A

**Environment:**

- Ark version (use `ark version`): v0.10.0
- Kubernetes version (use `kubectl version`): v1.13.0
- Kubernetes installer & version: Minikube, PKS, KubeSpray
- Cloud provider or hardware configuration: N/A
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04.1 LTS

ncdc:
Our backup and restore item action `Execute` methods are executed pre-backup (before persisting in the tarball) and pre-restore (before creating the item in Kubernetes). I'm maybe thinking that post-backup doesn't make much sense (do you have any use cases)? But post-restore definitely would give you what you need here. WDYT?

king-jam:
Definitely agree. Post Restore would be the priority scenario. The rest may be generalized under the feature add but aren't required immediately.

king-jam:
After thinking about it further, The Post Restore hooks would need a caveat.....mainly in a scenario of restoring a database/etc, the hooks needs to happen just before exposing the Service so higher level apps don't start trying to execute transactions before data is fully restored.

An init container can solve this. It just leaves breadcrumbs that we might not want every time.

ncdc:
Let's walk through the scenarios in full detail.

# Backup
- A custom backup item action plugin that you're writing calls `mysqldump` and uploads the output to the Ark ObjectStore

# Restore
1. Ark restores pod
1. A custom restore item action plugin runs post-restore for the pod
    1. It waits for the pod to be running
    1. It restores the data backed up by `mysqldump`

[...]

PR #2465, opened 2020-04-28
Design proposal for Restore hooks

### LINKED ISSUE #2609: design for restore hooks

marccampbell:
See #2116 

closes #2609

### DIFF
--- a/design/restore-hooks.md
+++ b/design/restore-hooks.md
@@ -1,16 +1,11 @@
 # Restore Hooks
 
-Velero supports Backup Hooks to execute commands at before and/or after a backup.
-This enables a user to, among other things, prepare data to be backed up.
-An example of this would be to attach an empty volume to a Postgres pod, use a backup hook to execute `pg_dump` from the data volume, and back up the volume containing the export.
-The problem is that there's no easy or automated way to include an automated restore process.
-After a restore with the example configuration above, the postgres pod will be empty, but there will be a need to manually exec in and run `pg_restore`.
-This is not specific to postgres backups, but any many database engines and other applications that have application-specific utilities to back up and restore the data.
 This document proposes a solution that allows a user to specify Restore Hooks, much like Backup Hooks, that can be executed during the restore process.
 
 ## Goals
 
 - Enable custom commands to be run during a restore in order to mirror the commands that are available to the backup process.
+- Provide observability into the result of commands run in restored pods.
 
 ## Non Goals
 
@@ -18,18 +13,97 @@ This document proposes a solution that allows a user to specify Restore Hooks, m
 
 ## Background
 
-(See introduction)
+Velero supports Backup Hooks to execute commands before and/or after a backup.
+This enables a user to, among other things, prepare data to be backed up without having to freeze an in-use volume.
+An example of this would be to attach an empty volume to a Postgres pod, use a backup hook to execute `pg_dump` from the data volume, and back up the volume containing the export.
+The problem is that there's no easy or automated way to include an automated restore process.
+After a restore with the example configuration above, the postgres pod will be empty, but there will be a need to manually exec in and run `pg_restore`.
 
 ## High-Level Design
 
-The Restore spec
+The Restore spec will have a `spec.hooks` section matching the same section on the Backup spec except no `pre` hooks can be defined - only `post`.
+Annotations comparable to the annotations used during backup can also be set on pods.
+For each restored pod, the Velero server will check if there are any hooks applicable to the pod.
+If a restored pod has any applicable hooks, Velero will wait for the pod to reach status Ready and then execute the hooks in the pod.
+
+The Restore log will include the results of each hook and the Restore object status will incorporate the results of hooks.
 
 ## Detailed Design
 
+Post-restore hooks can be defined by annotation and/or by an array of resource hooks in the Restore spec.
+
+The following annotations are supported:
+- post.hook.restore.velero.io/container
+- post.hook.restore.velero.io/command
+- post.hook.restore.velero.io/on-error
+- post.hook.restore.velero.io/timeout
+
+
+This is an example of defining hooks in the Restore spec.
+```yaml
+apiVersion: velero.io/v1
+kind: Restore
+spec:
+  ...
+  hooks:
+    resources:
+      -
+        name: my-hook
+        includedNamespaces:
+        - '*'
+        excludedNamespaces:
+        - some-namespace
+        includedResources:
+        - pods
+        excludedResources: []
+        labelSelector:
+          matchLabels:
+            app: velero
+            component: server
+        post:
+          -
+            exec:
+              container: my-container
+              command:
+                - /bin/uname
+                - -a
+              onError: Fail
+              timeout: 10s
+```
+
+As with Backups, if an annotation is defined on a pod then no hooks from the Restore spec will be applied.
+
+### Implementation
+
+The types and function in pkg/backup/item_hook_handler.go will be moved to a new package (pkg/util/hooks) and exported so they can be used for both backups and restores.
+
+The restore hooks implementation will closely follow the design of restoring pod volumes with restic.
+The pkg/restore.context type will have new fields `hooksWaitGroup` and `hooksErrs` comparable to `resticWaitGroup` and `resticErr`.
+The pkg/restore.context.execute function will start a goroutine for each pod with applicable hooks and then continue with restoring other items.
+Each hooks goroutine will create a pkg/util/hooks.ItemHookHandler for each pod and send any error on the context.hooksErrs channel.
+The ItemHookHandler already includes stdout and stderr and other metadata in the Backup log so the same logs will automatically be added to the Restore log (passed as the first argument to the ItemHookhandler.HandleHooks method.)
+
+The pkg/restore.context.execute function will wait for the hooksWaitGroup before returning.
+Any errors received on context.hooksErrs will be added to errs.Velero.
+
+One difference compared to the restic restore design is that any error on the context.hooksErrs channel will cancel the context of all hooks, since errors are only reported on this channel if the hook specified `onError: Fail`.
+However, canceling the hooks goroutines will not cancel the restic goroutines.
+In practice the restic goroutines will complete before the hooks since the hooks do not run until a pod is ready, but it's possible a hook will be executed and fail while a different pod is still in the pod volume restore phase.
+
+Failed hooks with `onError: Continue` will appear in the Restore log but will not affect the status of the parent Restore.
+Failed hooks with `onError: Fail` will cause the parent Restore to have status Partially Failed.
 
 ## Alternatives Considered
 
+Wait for all restored Pods to report Ready, then execute the first hook in all applicable Pods simultaneously, then proceed to the next hook, etc.
+That could introduce deadlock, e.g. if an API pod cannot be ready until the DB pod is restored.
+
+Put the restore hooks on the Backup spec as a third lifecycle event named `restore` along with `pre` and `post`.
+That would be confusing since `pre` and `post` would appear in the Backup log but `restore` would only be in the Restore log.
+
+Execute restore hooks in parallel for each Pod.
+That would not match the behavior of Backups.
 
 ## Security Considerations
 
-N/A
+Stdout or stderr in the Restore log may contain sensitive information, but the same risk already exists for Backup hooks.
--- a/design/restore-hooks.md
+++ b/design/restore-hooks.md
@@ -24,9 +24,12 @@ After a restore with the example configuration above, the postgres pod will be e
 The Restore spec will have a `spec.hooks` section matching the same section on the Backup spec except no `pre` hooks can be defined - only `post`.
 Annotations comparable to the annotations used during backup can also be set on pods.
 For each restored pod, the Velero server will check if there are any hooks applicable to the pod.
-If a restored pod has any applicable hooks, Velero will wait for the pod to reach status Ready and then execute the hooks in the pod.
-
+If a restored pod has any applicable hooks, Velero will wait for the container where the hook is to be executed to reach status Running.
+The Restore log will include the results of each post-restore hook and the Restore object status will incorporate the results of hooks.
 The Restore log will include the results of each hook and the Restore object status will incorporate the results of hooks.
+
+A new section at `spec.hooks.resources.initContainers` will allow for injecting initContainers into restored pods.
+Annotations can be set as an alternative to defining the initContainers in the Restore object.
 
 ## Detailed Design
 
@@ -36,10 +39,18 @@ The following annotations are supported:
 - post.hook.restore.velero.io/container
 - post.hook.restore.velero.io/command
 - post.hook.restore.velero.io/on-error
-- post.hook.restore.velero.io/timeout
+- post.hook.restore.velero.io/exec-timeout
+- post.hook.restore.velero.io/wait-timeout
 
+Init restore hooks can be defined by annotation and/or in the new `initContainers` section in the Restore spec.
+The initContainers schema is `pod.spec.initContainers`.
+
+The following annotations are supported:
+- init.hook.restore.velero.io/timeout
+- init.hook.restore.velero.io/initContainers
 
 This is an example of defining hooks in the Restore spec.
+
 ```yaml
 apiVersion: velero.io/v1
 kind: Restore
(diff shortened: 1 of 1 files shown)


nrb:
@jimbo459 @neil-hickey @aclevername you may be interested in this design proposal as well.

design/restore-hooks.md
```diff
@@ -0,0 +1,109 @@
...
+
+Velero supports Backup Hooks to execute commands before and/or after a backup.
+This enables a user to, among other things, prepare data to be backed up without having to freeze an in-use volume.
+An example of this would be to attach an empty volume to a Postgres pod, use a backup hook to execute `pg_dump` from the data volume, and back up the volume containing the export.
+The problem is that there's no easy or automated way to include an automated restore process.
+After a restore with the example configuration above, the postgres pod will be empty, but there will be a need to manually exec in and run `pg_restore`.
+
+## High-Level Design
+
+The Restore spec will have a `spec.hooks` section matching the same section on the Backup spec except no `pre` hooks can be defined - only `post`.
```
skriss, line 24: I wonder if there are different parts of the pod lifecycle that we would want to model/enable hooks for here. For example, we could theoretically support both (1) running a restore hook as an init container, so guaranteed to run before the main containers in the pod; and (2) running a restore hook after the pod is running, so as an exec hook inside the main container. Does anyone have thoughts on if something like this would be useful?

If we don't model this, then I *believe* you're proposing that we only support (2) from above, i.e. run the hook inside one of the existing containers in the pod once it's up and running.

design/restore-hooks.md
```diff
@@ -0,0 +1,109 @@
...
+Annotations comparable to the annotations used during backup can also be set on pods.
+For each restored pod, the Velero server will check if there are any hooks applicable to the pod.
+If a restored pod has any applicable hooks, Velero will wait for the pod to reach status Ready and then execute the hooks in the pod.
```
skriss, line 27: I know that, for the restic integration, certain users have seen issues where the pod that velero restores (and attempts to run the restic restore for) ends up being replaced by a controlling ReplicaSet shortly after restore, which ultimately makes that restic restore fail. See https://github.com/vmware-tanzu/velero/issues/1981#issuecomment-544709044 and the subsequent discussion for a lot more detail on that.

I do think, though, that for most environments, Velero's assumption that the pod that it restores will be "adopted" by its controller, seems to work fine. It sounds like, in order for the restore hooks to work as expected, we'd be making that same assumption?

(I know this is a somewhat vague comment, probably merits a live discussion at some point, but hopefully I wrote enough to point in the right direction for some context).

design/restore-hooks.md
```diff
@@ -0,0 +1,109 @@
...
+Any errors received on context.hooksErrs will be added to errs.Velero.
+
+One difference compared to the restic restore design is that any error on the context.hooksErrs channel will cancel the context of all hooks, since errors are only reported on this channel if the hook specified `onError: Fail`.
+However, canceling the hooks goroutines will not cancel the restic goroutines.
+In practice the restic goroutines will complete before the hooks since the hooks do not run until a pod is ready, but it's possible a hook will be executed and fail while a different pod is still in the pod volume restore phase.
+
+Failed hooks with `onError: Continue` will appear in the Restore log but will not affect the status of the parent Restore.
+Failed hooks with `onError: Fail` will cause the parent Restore to have status Partially Failed.
+
+## Alternatives Considered
```
skriss, line 96: A couple other related upstream efforts to consider:

https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190120-execution-hook-design.md
https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20200120-generic-data-populators.md

I don't think that either of those are in a state where we'd want to change our approach here for now, but worth look at/keeping an eye on.

skriss:
This issue also seems related from a quick skim: https://github.com/vmware-tanzu/velero/issues/1150

design/restore-hooks.md
totherme: For our use case (working on TAS backup-and-restore), this use-case (2) looks like exactly what we need: we'd like to be able to re-hydrate a database after the pod running the database server is ready.

Use-case (1) looks potentially useful to us in future, but no where near as immediately useful as use-case (2). So we're totally in agreement with adding use-case (2) right now, and maybe opening a new design PR later if we end up with a real-world use for (1).

neil-hickey:
> @jimbo459 @neil-hickey @aclevername you may be interested in this design proposal as well.

LGTM. We had a look at the design doc and added comments (via @totherme )

design/restore-hooks.md
```diff
@@ -0,0 +1,109 @@
...
+
+Wait for all restored Pods to report Ready, then execute the first hook in all applicable Pods simultaneously, then proceed to the next hook, etc.
+That could introduce deadlock, e.g. if an API pod cannot be ready until the DB pod is restored.
+
+Put the restore hooks on the Backup spec as a third lifecycle event named `restore` along with `pre` and `post`.
+That would be confusing since `pre` and `post` would appear in the Backup log but `restore` would only be in the Restore log.
+
```
totherme, line 103: I like calling this cook a "post" hook, since that leaves us open to add "pre" hooks in future if we want them (along the lines of skriss' comments above)

neil-hickey: An example of a pre hook might be:

- I have a corrupted database / invalid state of a DB I am restoring into. My pre-hook will delete the corrupted data prior to restore, or reconcile a good state of the world prior to restore. (This can be done via post-hooks also. But the separation may be cleaner)

design/restore-hooks.md
areed: Yes, this design makes the same assumption.

An alternative to avoid this issue would be:
- Drop support for restore annotations
- Move restore hook execution stage to after the restic restore wait
- Then use the labelSelector in the Restore hook spec to list Pods and then execute the hook in all that match.

design/restore-hooks.md
```diff
@@ -0,0 +1,109 @@
...
+However, canceling the hooks goroutines will not cancel the restic goroutines.
+In practice the restic goroutines will complete before the hooks since the hooks do not run until a pod is ready, but it's possible a hook will be executed and fail while a different pod is still in the pod volume restore phase.
+
+Failed hooks with `onError: Continue` will appear in the Restore log but will not affect the status of the parent Restore.
+Failed hooks with `onError: Fail` will cause the parent Restore to have status Partially Failed.
+
+## Alternatives Considered
+
+Wait for all restored Pods to report Ready, then execute the first hook in all applicable Pods simultaneously, then proceed to the next hook, etc.
+That could introduce deadlock, e.g. if an API pod cannot be ready until the DB pod is restored.
```
mynktl, line 99: We can have some option to enable execution simultaneously, since in some cases simultaneous execution may not required, for example running db repair or file system repair can be executed sequentially.

design/restore-hooks.md
```diff
@@ -0,0 +1,109 @@
...
+                - -a
+              onError: Fail
+              timeout: 10s
+```
+
+As with Backups, if an annotation is defined on a pod then no hooks from the Restore spec will be applied.
+
+### Implementation
+
+The types and function in pkg/backup/item_hook_handler.go will be moved to a new package (pkg/util/hooks) and exported so they can be used for both backups and restores.
```
nrb, line 78: Somewhat minor, but I'd promote the package to be `pkg/hooks` vs being inside `pkg/util`.

design/restore-hooks.md
```diff
@@ -0,0 +1,109 @@
...
+The pkg/restore.context type will have new fields `hooksWaitGroup` and `hooksErrs` comparable to `resticWaitGroup` and `resticErr`.
+The pkg/restore.context.execute function will start a goroutine for each pod with applicable hooks and then continue with restoring other items.
+Each hooks goroutine will create a pkg/util/hooks.ItemHookHandler for each pod and send any error on the context.hooksErrs channel.
+The ItemHookHandler already includes stdout and stderr and other metadata in the Backup log so the same logs will automatically be added to the Restore log (passed as the first argument to the ItemHookhandler.HandleHooks method.)
+
+The pkg/restore.context.execute function will wait for the hooksWaitGroup before returning.
+Any errors received on context.hooksErrs will be added to errs.Velero.
+
+One difference compared to the restic restore design is that any error on the context.hooksErrs channel will cancel the context of all hooks, since errors are only reported on this channel if the hook specified `onError: Fail`.
+However, canceling the hooks goroutines will not cancel the restic goroutines.
```
nrb, line 90: Thanks for putting in the thought on the interactions with restic!

design/restore-hooks.md
nrb: I suspect the annotations are included to mirror the current backup hooks design, but it certainly looks like there's some cases where it's likely to cause trouble.

Perhaps we don't include them?

skriss: I'm going to dig into this some more and see if I can get a better understanding of the scenarios where the pod adoption wouldn't work.

design/restore-hooks.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Detailed Design
+
+Post-restore hooks can be defined by annotation and/or by an array of resource hooks in the Restore spec.
+
+The following annotations are supported:
+- post.hook.restore.velero.io/container
+- post.hook.restore.velero.io/command
+- post.hook.restore.velero.io/on-error
+- post.hook.restore.velero.io/timeout
```
ashish-amarnath, line 39: We might also need a timeout to configure how long Velero should wait for the pod to be come running.

design/restore-hooks.md
ashish-amarnath: Considering that the primary purpose of these hooks are for application data restoration and the fact that these hooks will be run once the pod becomes ready. Most applications would then need to be restarted to recognize the restored data. For this reason, the init container approach may be better suited.
As we are designing this, we should add both flavors of this restore hook and not just running commands. The approach of init containers will also have better error handling. In that, the pod doesn't run if the data isn't restored correctly.

design/restore-hooks.md
ashish-amarnath: The execution-hook proposal is now obsolete and is being replaced by [Draft ContainerNotifier Proposal](https://docs.google.com/document/d/1Q4UMrx9r58LUNOHdn_4nDhqHN21iOOYQeViMKSPa_Ks/edit#heading=h.xh2wqx7h9cqm)

design/restore-hooks.md
ashish-amarnath: What will be the lifecycle phase of the pod for running a `pre-restore-hook`?

design/restore-hooks.md
nrb: @skriss Have you been able to do this research? Any further thoughts here?

nrb:
@marccampbell FYI, we've added this to our roadmap for v1.5. See https://github.com/vmware-tanzu/velero/blob/master/ROADMAP.md

design/restore-hooks.md
areed: I think it would help with the design if we had a concrete example of when initContainers would be useful during restore. The closest thing I could come up with is a scenario like this:

You have an API deployment that uses Postgres as its DB. You have your migrations in an initContainer for that deployment. You have a pre-snapshot hook that executes against the API pod and runs pg_dump to a volume in the pod. Then you would want an initContainer to be injected before the migrations initContainer to run pg_restore.

I'm not sure the above scenario would appear in the field since most API deployments have multiple replicas and it would be weird to run pg_dump/pg_restore multiple times for snapshot and restore. Do you have another use case in mind for initContainers?

design/restore-hooks.md
areed: During the latest community meeting the consensus was that since adoption behavior is already relied on for the restic integration it's safe to rely on for restore hooks.

nrb:
@jimbo459 @neil-hickey @aclevername Feel free to add approvals on this as well if you've read it and are ok with the design.

design/restore-hooks.md
ashish-amarnath: @areed 
The scenario I had was restoring of the stateful application itself and most stateful applications maintain state by writing files to disk. 

In the case of postgres, if postgres were to start before running `pg_dump`/`pg_restore`, it would start by writing data into files that `pg_dump`/`pg_restore` would write when the restore hooks are run. The hooks will then overwrite the data that postgres initialized itself with. For postgres to heal, it may require a restart of postgres itself. Which means there will be a container restart. When this happens, it is possible for the pod, running the application, to be relocated and the new pod wouldn't be controlled by Velero for it to run the restore hooks to restore data. This is scope for data-loss.

Furthermore, when restoring a stateful application, say like Apache Zookeeper, which uses transaction semantics to maintain/manage state, overwriting data (performed in the restore hook) that the application started with will lead to data corruption causing the application to crash. Making the restore unsuccessful.

For this reason, the restore hooks should be run before the application can start.

stephbman:
@ashish-amarnath with regards to running the stateful applications in init containers- what would be the additional work required to support this? And, would we want to add this to the current design doc? (asked on velero slack as well)

nrb:
> For that reason, IMO, we should be supporting running these hooks as init containers as well.

@ashish-amarnath When you say "as well," are you advocating for the option to use them as init containers or as annotations as currently proposed?

I think your point about potential data loss is a valid one, and we definitely need to address it in some fashion, even if that isn't necessarily by implementing init containers and documenting _why_ we did it this way.

ashish-amarnath:
@stephbman I haven't had a chance to think about what the design for supporting init containers would look like. But this could follow how restic restores are performed.

IMO, the current proposal caters to the proverbial 20% of the use cases so implementing this may not benefit all of our users. Not sure if that is enough bang-for-buck to implement this.

@nrb I agree that init containers may not be the only way to address this. But, I think we agree that this needs to be addressed.
By "as-well" I mean, the design should identify extension points to cover more use-cases without having drastically different implementations.

Ideally, we would design both at once and implement it in two phases.

nrb:
@jimbo459 @neil-hickey @aclevername @marccampbell @areed Would you be able to attend the community meeting on June 23 2020 for discussion on this? Also, it's likely we'll schedule a separate, dedicated call for this too.

marccampbell:
Yes. I can be there to discuss on Tuesday.

The init container support actually makes sense to me. In our product, we've implemented restore hooks as custom scripts (until the work in this PR is done), and had this problem too. Interestingly, we had to solve this by running the custom restore hooks in an init container (https://github.com/replicatedhq/kots/blob/e5351f44c164efd0a904eda9bd728058b90b8016/pkg/kotsadm/kotsadm_objects.go#L258) to prevent services from writing to the database before/during restore.

design/restore-hooks.md
totherme: I don't think we have a need for pre-restore-hooks right now, but I can imagine it being useful to have one that is guaranteed to run:
- before the main containers in the restored pod start ; and
- after the PVCs have been mounted. 

This would allow us to do what neil-hickey suggests in the event of a corrpted DB: replace the bad data before starting a database process.

Having said all that, I don't think there's any need at all for pre-restore-hooks in an MVP. I just like the idea of leaving the option open for future work.

totherme:
> @jimbo459 @neil-hickey @aclevername Feel free to add approvals on this as well if you've read it and are ok with the design.

Yep, lgtm :)

stephbman:
Hi @marccampbell @areed outside of the community meeting (where we'll have a more limited amount of time to discuss this PR) - we were also thinking of setting up a deep dive session for another day this week. 

Are there some days/times that you all are available to meet with us? We generally have 9-11 am CST available each week (one team member involved is located in London so the earlier time frame accommodates that).

marccampbell:
@stephbman We'd love to. @areed is OOO this week, could we schedule for early next week? We are both in Pacific time, but can do 10 AM CST on 30 June or 1 July if that works.

stephbman:
@marccampbell great! Please use this link to select a time that might work for you next week. We have a team member in London that would like to attend. If that limits availability too much, let me know and I can see what we might move around.

Link to calendar - please select a time that works for you:  https://calendly.com/steph_bauman/velero-restore-hooks-deep-dive

nrb:
To summarize our meeting from July 6 2020:

* We're going to aim to have both initContainer-based and annotation-based hooks as the initial implementation
* Concurrency concerns will be up to the authors of the hooks that velero calls for now
* Things like coordinating running a hook once-per-set-of-pods will be tackled in the future.

nrb:
@ashish-amarnath and @carlisia Please take another pass on this when you can

nrb:
@marccampbell Looks like the merge commit will need the DCO sign off added via `git commit --amend --sign-off`.

design/restore-hooks.md
```diff
@@ -0,0 +1,143 @@
...
+One difference compared to the restic restore design is that any error on the context.hooksErrs channel will cancel the context of all hooks, since errors are only reported on this channel if the hook specified `onError: Fail`.
+However, canceling the hooks goroutines will not cancel the restic goroutines.
+In practice the restic goroutines will complete before the hooks since the hooks do not run until a pod is ready, but it's possible a hook will be executed and fail while a different pod is still in the pod volume restore phase.
+
+Failed hooks with `onError: Continue` will appear in the Restore log but will not affect the status of the parent Restore.
+Failed hooks with `onError: Fail` will cause the parent Restore to have status Partially Failed.
+
+If initContainers are specified for a pod, Velero will inject the containers into the beginning of the pod's initContainers list.
+If a restic initContainer is also being injected, the restore initContainers will be injected directly after the restic initContainer.
+The restore will use a RestoreItemAction to inject the initContainers.
```
nrb, line 120: This is a `RestoreItemAction` that applies to pods?

ashish-amarnath:
This should be good for merge once the commits are signed.
https://github.com/vmware-tanzu/velero/pull/2465/checks?check_run_id=866265760

design/restore-hooks.md
areed: Yes, only applies to pods

nrb:
@marccampbell Looks like we may need a cherry-pick or something other than a merge, as the sign is still failing :(

marccampbell:
@nrb thanks.  The DCO is passing now.

nrb:
@marccampbell @areed Thanks so much for driving this forward!

I've broken out the most visible work items that I could see on the ZenHub board and use the ZenHub dependencies to split that out. Please take a look at it: https://app.zenhub.com/workspaces/velero-5c59c15e39d47b774b5864e3/board?labels=restore%20hooks&milestones=v1.5%232020-08-31&repos=99143276,112385197,213946861,190224441,214524700,214524630

I think we have a few small issues that can be knocked out by pretty much anyone, and then the larger implementation of  init container injection/restore hook execution, which may yet be broken into smaller chunks.

PR #2480, opened 2020-04-29
Delete dangling volumesnapshotcontents from velero backups

ashish-amarnath:

Branch for this PR was cut from that of https://github.com/vmware-tanzu/velero/pull/2411

Commit with the change in this PR is https://github.com/vmware-tanzu/velero/commit/6b235d059c3405fa635511c404a5bdac6fdd844d

This PR implements the deletion of volumesnapshotcontents described in this comment https://github.com/vmware-tanzu/velero-plugin-for-csi/blob/master/internal/backup/volumesnapshot_action.go#L125-L129

### DIFF
--- a/pkg/controller/backup_deletion_controller.go
+++ b/pkg/controller/backup_deletion_controller.go
@@ -520,7 +520,11 @@ func deleteCSIVolumeSnapshotContents(backupName string, csiVSCLister snapshotv1b
 	if err != nil {
 		return []error{err}
 	}
-
+	// It is possible that by the time deleteCSIVolumeSnapshotContents is called after deleteCSIVolumeSnapshots
+	// that deletion of VSCs hasn't been completed, by the snapshot-controller (one of the CSI components).
+	// For that reason the csiVSCLister returned VSCs that are yet to be deleted. To handle this scenario,
+	// we swallow `IsNotFound` errors from the setVolumeSnapshotContentDeletionPolicy function and the
+	// csiClient.VolumeSnapshotContents().Delete(...)
 	log.Infof("Deleting %d CSI volumesnapshotcontents", len(csiVolSnapConts))
 	for _, snapCont := range csiVolSnapConts {
 		err := setVolumeSnapshotContentDeletionPolicy(snapCont.Name, csiClient, log)
@@ -535,7 +539,7 @@ func deleteCSIVolumeSnapshotContents(backupName string, csiVSCLister snapshotv1b
 		}
 		log.Infof("Deleting volumesnapshotcontent %s", snapCont.Name)
 		err = csiClient.VolumeSnapshotContents().Delete(snapCont.Name, &metav1.DeleteOptions{})
-		if err != nil {
+		if err != nil && !apierrors.IsNotFound(err) {
 			errs = append(errs, err)
 		}
 	}
--- a/pkg/controller/backup_deletion_controller.go
+++ b/pkg/controller/backup_deletion_controller.go
@@ -67,6 +67,7 @@ type backupDeletionController struct {
 	backupLocationLister      velerov1listers.BackupStorageLocationLister
 	snapshotLocationLister    velerov1listers.VolumeSnapshotLocationLister
 	csiSnapshotLister         snapshotv1beta1listers.VolumeSnapshotLister
+	csiSnapshotContentLister  snapshotv1beta1listers.VolumeSnapshotContentLister
 	csiSnapshotClient         *snapshotterClientSet.Clientset
 	processRequestFunc        func(*velerov1api.DeleteBackupRequest) error
 	clock                     clock.Clock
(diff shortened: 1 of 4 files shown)


skriss:
Question on this - if we have a dangling VSC, and we set its deletionpolicy to Delete, does deleting the VSC trigger storage snapshot deletion as well? I'm not 100% sure what the answer is.

ashish-amarnath:
Yes, deleting the dangling VSC will also delete the storage snapshot. I verified that in testing.

changelogs/unreleased/2411-ashish-amarnath
```diff
@@ -0,0 +1 @@
+during backup deletion also delete CSI volumesnapshots that were created as a part of the backup
```
carlisia, line 1: I just noticed this changelog is the same as for the PR #2411.

ashish-amarnath: Actually, this PR is missing a change log. The change log that you are seeing here is bc this branch was cut from the PR #2411 branch. Will add a changelog

pkg/controller/backup_deletion_controller.go
```diff
@@ -324,7 +327,14 @@ func (c *backupDeletionController) processRequest(req *velerov1api.DeleteBackupR
...
 		log.Info("Removing CSI volumesnapshots")
-		if csiErrs := deleteCSIVolumeSnapshots(backup, c.csiSnapshotLister, c.csiSnapshotClient.SnapshotV1beta1(), log); len(csiErrs) > 0 {
+		if csiErrs := deleteCSIVolumeSnapshots(backup.Name, c.csiSnapshotLister, c.csiSnapshotClient.SnapshotV1beta1(), log); len(csiErrs) > 0 {
+			for _, err := range csiErrs {
+				errs = append(errs, err.Error())
+			}
+		}
+
+		log.Info("Removing CSI volumesnapshotcontents")
+		if csiErrs := deleteCSIVolumeSnapshotContents(backup.Name, c.csiSnapshotContentLister, c.csiSnapshotClient.SnapshotV1beta1(), log); len(csiErrs) > 0 {
```
skriss, line 337: I wonder if we're going to run into any race conditions between the deletion of the VolumeSnapshots, which will trigger the CSI snapshot controller to go delete the related VolumeSnapshotContents, and our explicit deletion of the VolumeSnapshotContents themselves. Off the top of my head it seems like everything would probably work OK, but hard to say without testing. Did you observe any such issues?

ashish-amarnath: This is the test that I ran.
1. Deploy 4 instances of the [hostpath-plugin example](https://github.com/kubernetes-csi/csi-driver-host-path/tree/master/examples) each in their own namesapces.  
2. `v backup create b1 --include-namespaces csi-app,csi-app-1,csi-app-2,csi-app-3 --wait`
3. `v delete backup b1`
This is what I saw:
- `deleteCSIVolumeSnapshots` function:
    - set the `DeletionPolicy` on all the VSCs
    - deleted the VS 
- By the time `deleteCSIVolumeSnapshotContents` function was invoked the VSCs weren't all deleted yet. So:
- csiVSCLister returned 4 VSCs to be deleted. 
- the `setVolumeSnapshotContentDeletionPolicy` was called for each of the 4 VSCs. However, it returned with a `NotFound` error for the last 3 of the VSCs which we handle [here](https://github.com/vmware-tanzu/velero/pull/2480/files#diff-afe709fa1785e08707ca0825e3f60e68R527) to `continue` the deletion loop.
- Only for the VSC that `setVolumeSnapshotContentDeletionPolicy` didn't return the `NotFound` error we call the delete function on the VSC.

So I didn't see any errors. However, I think we should be handling the `NotFound` error returned from
`` `
err = csiClient.VolumeSnapshotContents().Delete(snapCont.Name, &metav1.DeleteOptions{})
``` 
I am going to make that change

Comment (nrb) [reply pkg/controller/backup_deletion_controller.go:337]: @ashish-amarnath 404 would essentially be a no-op, right?

Comment (ashish-amarnath) [reply pkg/controller/backup_deletion_controller.go:337]: Yeah. We might run into that when we attempt to delete something that doesn't exist.
I am not sure if the `Delete(...)` handles that and returns nil on `NotFound`

File: pkg/controller/backup_deletion_controller_test.go [hunk 3]
```diff
@@ -1022,39 +1098,113 @@ func TestDeleteCSIVolumeSnapshots(t *testing.T) {
...
+			Labels: map[string]string{
+				velerov1.BackupNameLabel: "backup2",
+			},
+		},
+		Spec: snapshotv1beta1api.VolumeSnapshotContentSpec{
+			DeletionPolicy: snapshotv1beta1api.VolumeSnapshotContentDelete,
+		},
+	}
+
+	nothingVSC := snapshotv1beta1api.VolumeSnapshotContent{
```
nrb, line 1163: Would it work define the VSCs in the test cases, then add them to the client and informer in the test loop?

That way, all the data for the test case is collected in the `testCases` slice, at the expense of making the contents of `t.Run` longer. It would be easier to edit the case data in case more get added or changes are needed, though.

I have a similar comment for the `TestSetVolumeSnapshotContentDeletionPolicy` tests above

skriss:
:+1: this makes sense to me, no further comments - will re-review next week

pkg/controller/backup_deletion_controller_test.go
ashish-amarnath: done!

ashish-amarnath:
@nrb this PR should be good for merge?

Issue #2466, opened 2020-04-28
Unable to install ca-certificate package on arm builds


xUnholy:
I'm deploying velero via helm-operator, I'm using the GCP plugin and during the `/velero server` container command I find the below error occurring. I'm not sure if the error is because of TLS or the backup store location.

I'm using v2.9.15 of the helm chart.

```An error occurred: some backup storage locations are invalid: backup store for location "gcp" is invalid: rpc error: code = Unknown desc = Get "https://storage.googleapis.com/storage/v1/b/raspbernetes-velero-backups/o?alt=json&delimiter=%2F&pageToken=&prefix=&prettyPrint=false&projection=full&versions=false": oauth2: cannot fetch token: Post "https://oauth2.googleapis.com/token": x509: certificate signed by unknown authority```

My helm release config looks as follows:

```
---
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: velero
  namespace: velero
  annotations:
    fluxcd.io/ignore: 'false'
    fluxcd.io/automated: 'false'
spec:
  releaseName: velero
  helmVersion: v3
  chart:
    repository: https://vmware-tanzu.github.io/helm-charts
    name: velero
    version: 2.9.15
  values:
    credentials:
      existingSecret: cloud-credentials
    configuration:
      provider: gcp
      backupStorageLocation:
        name: gcp
        bucket: raspbernetes-velero-backups
      volumeSnapshotLocation:
        name: gcp
    initContainers:
      - name: velero-plugin-for-gcp
        image: raspbernetes/velero-gcp-plugin:v1.0.1
        imagePullPolicy: Always
        volumeMounts:
          - mountPath: /target
            name: plugins
    schedules:
      daily-backup:
        schedule: '*/30 * * * *'
        template:
          ttl: '120h'
```

**Comment by ashish-amarnath** [comment]
@xUnholy It looks like you are missing the CA certificate that trusts the google endpoint. Can you please try to update the ca-certificate package for your operating system on the nodes in your kubernetes cluster and give this another try?

**Comment by xUnholy** [comment]
The `ca-certificate` build tool doesn't seem to be in the velero image.

**Comment by ashish-amarnath** [comment]
This change was reverted as it broke the container build (see https://github.com/vmware-tanzu/velero/pull/2481#issuecomment-621982600)
Re-opening the issue.

**Comment by sgielen** [comment]
I've encountered this as well on AWS:

```
[...]

PR #2481, opened 2020-04-30
Install ca-certificates for ARM based container builds.

twoequaldots:
To help with bug reported in issue #2466 , I've modified the Dockerfile for ARM based containers to install `ca-certificates`

EDIT (@ashish-amarnath )
Fixes: #2466

### DIFF
--- a/Dockerfile-velero-arm
+++ b/Dockerfile-velero-arm
@@ -14,6 +14,8 @@
 
 FROM arm32v7/ubuntu:bionic
 
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates
+
 ADD /bin/linux/arm/restic /usr/bin/restic
 
 ADD /bin/linux/arm/velero /velero
--- a/Dockerfile-velero-arm64
+++ b/Dockerfile-velero-arm64
@@ -14,6 +14,8 @@
 
 FROM arm64v8/ubuntu:bionic
 
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates
+
 ADD /bin/linux/arm64/restic /usr/bin/restic
 
 ADD /bin/linux/arm64/velero /velero
--- a/changelogs/unreleased/2481-twoequaldots
+++ b/changelogs/unreleased/2481-twoequaldots
@@ -1,0 +1,1 @@
+Install `ca-certificates` in ARM based containers.


xUnholy:
It should be noted that this is also to gain parity between the ARM(64) images and the x86 image

skriss:
Looks like the container build failed (from https://travis-ci.org/github/vmware-tanzu/velero/builds/681576991):

```
Step 2/6 : RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates
 ---> Running in c2b49b576083
standard_init_linux.go:190: exec user process caused "exec format error"
The command '/bin/sh -c apt-get update && apt-get install -y --no-install-recommends ca-certificates' returned a non-zero code: 1
Makefile:165: recipe for target '.container-velero_velero-arm-master' failed
make[1]: *** [.container-velero_velero-arm-master] Error 1
Makefile:93: recipe for target 'container-linux-arm' failed
make: *** [container-linux-arm] Error 2
Script failed with status 2
```

ashish-amarnath:
Running `make container-linux-arm` on my laptop and not seeing this error.
Trying to get myself a linux box to mimic travis

ashish-amarnath:
OK I was able to repro this on an ubuntu machine.
image

Working on a fix

twoequaldots:
Anything I can do to help?  I did build locally and I tried to emulate what was in `.travis.yml` but I don't have a Travis-CI service to emulate this repo and thought the CI checks would catch if my change was not enough.

ashish-amarnath:
@twoequaldots The checks we run on the PR didn't include the container builds. That's the reason why this didn't show up on the PR.
I was able to repro this by running the `make container-linux-arm` on an ubuntu machine. The build, however was successful on my mac. The cause of the error seems to be bc the build was using amd64 binaries to run inside the arm container. This was the simplest repro of the error
image

running `docker run -ti arm32v7/ubuntu:bionic bash`

twoequaldots:
@ashish-amarnath If I'm following correctly, the build server is x86 architecture and it's breaking when trying to run a docker container that is arm or arm64 (aarch64) based.  If that is the case, do you have `qemu binfmt-support qemu-user-static` packages installed?  I believe you might just need `qemu-user-static` as it will naturally bring in the rest.  If you do have it, you might be able to mount `qemu-arm-static` inside the container for arm emulation as seen [here](https://ownyourbits.com/2018/06/27/running-and-building-arm-docker-containers-in-x86/).  I also stood up an Ubuntu 18.04 server and installed `docker-ce` to emulate the problem and solved it using [this article](https://www.stereolabs.com/docs/docker/building-arm-container-on-x86/)

Is either of the two solutions alright to use? (mainly to install `qemu-user-static`).  If so, it make need modification to `Makefile` for building ARM based containers on x86[_64] systems.

xUnholy:
Just validated running this change against my RPI4 cluster and it is working and has fixed the certificate issue mentioned in my issue.

twoequaldots:
I pushed the images (built on x86 using the above methods I defined) here: https://hub.docker.com/u/twoequaldots

Let me know if there is anything else I can do.

ashish-amarnath:
@twoequaldots Yes your understanding of the problem is correct. I don't see us explicitly installing the emulator packages and that is what we suspect caused the failure. There was nothing wrong with your PR per se. We need to update our [travis settings](https://github.com/vmware-tanzu/velero/blob/master/.travis.yml) to install these packages. We'll merge your PR once we do that.

twoequaldots:
Thanks a lot for the feedback and explanation, it's much appreciated :)

> If you are able to help us figure out what we are missing that'd be awesome!

I ran with this comment and it didn't occur to me that I can poke around `.travis.yml` to possibly propose a solution as I thought it might be a problem with the build server (i.e. certain packages need to be installed to emulate architecture or run ARM builds on ARM servers).  Thanks again, will patiently wait for this to be fixed in a future release (no rush).

RobReus:
@ashish-amarnath any update on fixing multi-arch builds in travis? Is there any help appreciated here?

Would be really nice to have this fixed before next release.

ashish-amarnath:
@RobReus Yes, we will definitely appreciate your help on this issue. I've pulled https://github.com/vmware-tanzu/velero/issues/2466 into the 1.5 milestone.

RobReus:
@ashish-amarnath sounds good.

nrb:
@RobReus A quick note, we've now moved to GitHub Actions instead of Travis.

RobReus:
Yeah I noticed, I like that change :) I am currently working on refactoring the build code and moving to docker buildx to properly build multi-arch images. Any thoughts on that before I move further and invest quite some time in refactoring?

Also, any specific reason why we would keep the arch specific images (i.e. "velero-arm64") instead of just "velero" with all archs supported? (Like we already have right now, but done via the manifest hack and not buildx)

nrb:
In theory, I'm in favor of buildx.

The arch specific images were an artifact of the manifest approach we used at the time.

If we have a single image, am I correct in assuming that the image size will grow?

RobReus:
It will not be a single image, it will still be multiple images (1 for each arch), but all under the same image name and tag. This is something that is already being done on the current "velero" image, but using the manifest method instead of buildx.

My idea is that we are going to get rid of all the arch specific build stuff and are going to end up with a single dockerfile which can be used to build all the arch types for both velero and velero-restic-restore-helper.

If there's no objections on using buildx, then I will go ahead with the refactor and make a PR for it so we can have a discussion about it.

nrb:
@RobReus Huge 👍 from me then! I look forward to the PR

Issue #2421, opened 2020-04-13
Add columns to backup storage location to indicate which one is the default


carlisia:
- to indicate which one is the default

Edited on 11/18:
- to show last validated time -> this was already done
- to show validation frequency -> opened a separate issue for this: https://github.com/vmware-tanzu/velero/issues/3098
- to add a default column for VSL -> Add columns to backup storage location to indicate which one is the default

PR #2490, opened 2020-05-01
Add a BSL controller to handle validation + update BSL status phase

carlisia:
- [x] add a controller-runtime controller that continuously validates BSLs and marks them as `available` (valid) or `unavailable` (invalid), and logs accordingly.
- [x] remove BSL validation from the server startup


pkg/controller/backup_storage_location_controller.go
```diff
@@ -0,0 +1,168 @@
...
+		backupStorageLocationLister: backupStorageLocationLister,
+
+		// use variables to refer to these functions so they can be
+		// replaced with fakes for testing.
+		newPluginManager: newPluginManager,
+		newBackupStore:   persistence.NewObjectBackupStore,
+	}
+
+	c.resyncFunc = c.run
+	c.resyncPeriod = 30 * time.Second
```
nrb, line 71: So this will try to revalidate every BSL once every 30 seconds?

We may want to make this tuneable, as there can be monetary access costs associated with hitting services like S3. This change moves the validation from a one-time startup operation to a recurring one, so I don't want to surprise people with new bills :)

I'm also wondering if we want to add a "last validated" timestamp to our BSLs. This may be useful in diagnosing issues, though I'm not sure it can be used to judge when to trigger a validation again. After all, just because it was valid 5 minutes ago doesn't mean it's valid now.

carlisia: Correct.

Yes, good point. I'll add a server flag for this to be configurable. And I like the idea of time-stamping, I think if people make the resync period too long, and with one too many BS:Ls, this will be useful and it's not a big deal to add.

Maybe in the future we can have a command that triggers the validation.

carlisia: What should the default be?

carlisia:
I added:

- `StoreValidationPeriod`
- `LastValidationTime`

Much in line with what's in the `BackupSynchController`.

I also added this, for the state between creation and validation:
- `BackupStorageLocationPhaseUnverified`

However, although I have verified this new phase was included in the CRD, it does not mark it as the default. Not sure if I'm doing this wrong:
https://github.com/vmware-tanzu/velero/pull/2490/files#diff-7c4f4164f55721d9493c9824d4735f60R101

After this PR is merged, I'm going to add these columns to the BSL output: https://github.com/vmware-tanzu/velero/issues/2421

pkg/controller/backup_storage_location_controller.go
```diff
@@ -0,0 +1,210 @@
...
+		backupLocationClient:         backupLocationClient,
+		backupStorageLocationLister:  backupStorageLocationLister,
+
+		// use variables to refer to these functions so they can be
+		// replaced with fakes for testing.
+		newPluginManager: newPluginManager,
+		newBackupStore:   persistence.NewObjectBackupStore,
+	}
+
+	c.resyncFunc = c.run
```
skriss, line 80: I think in addition to the periodic validation via this resync func, we should also add OnAdd/OnUpdate event handlers to the backup storage location informer so that whenever a BSL is created or updated, it immediately triggers a validation. WDYT?

carlisia: Yes, good idea.

Would we need both the handlers and the BSL validation though? 

I'm thinking add the validation to the handlers (create/update), and keep the BSL looping and checking the CR field just so it does the proper periodic logging. Would this work?

carlisia: > This brings up another nuance, which is - if we add a validation check in the backup controller to ensure the BSL is "Available" - should we rely on the CR field, or should we actually try to connect to the object store (e.g. via the backupStore.IsValid() method)? Relying on the CR field is easier/more efficient, but it could theoretically be out-of-date.

If we have the handler validating right on `update` then it will never be out-of-date. Right?

skriss: I think it makes sense to do the validation both when a BSL is added/updated, *and* periodically (via the resync func). That way, we immediately get an updated phase when the BSL is created or changed, but we are also regularly checking to ensure we can still connect to the object store.

skriss: It could still be out-of-date: the minio instance goes down, the S3 credentials are expired, etc. None of those would involve a change to the BSL, but could still mean that it can't be connected to.

carlisia: Yes, good point!

carlisia:
Ok recapping, I will:
1) add handlers to do additional validation/update of a BSL status on BSL create/update
2) keep the validation on the BSL controller so it also validates/updates (and logs) it there
3) add a validation to backup creation to check if the corresponding BSL is valid

> This brings up another nuance, which is - if we add a validation check in the backup controller to ensure the BSL is "Available" - should we rely on the CR field, or should we actually try to connect to the object store (e.g. via the backupStore.IsValid() method)? Relying on the CR field is easier/more efficient, but it could theoretically be out-of-date.

This question pertains to item # 3. My thinking below, taking into account @nrb's comment:
> We may want to make this tuneable, as there can be monetary access costs associated with hitting services like S3. This change moves the validation from a one-time startup operation to a recurring one, so I don't want to surprise people with new bills :)

I think doing the validation on create/update + periodically **could** be good enough. Since there are potential costs associated with it, we should aim for less rather than more. Another factor is that each BSL can have its validation  period decreased for more frequent checks: https://github.com/vmware-tanzu/velero/pull/2490/files#diff-0528e5cf4cf25ec346aa95f0bd1b2fd4R79.

WDYT? We could go with this and add it to the backup creation later if we find out we need it.

skriss:
That all sounds reasonable to me! Just a couple clarifications/related comments:

- for item # 1, those should go inside the same controller as item # 2 - so you'll have both event handlers on the informer, *and* a periodic resync function, in that one controller. The restic repository controller does something very similar: https://github.com/vmware-tanzu/velero/blob/master/pkg/controller/restic_repository_controller.go#L78-L87
- there are probably other controllers beyond the backup controller where we want to check the phase of the BSL up-front. Probably need to just go through them one-by-one and see if it makes sense to add a check on that phase somewhere at the start of their processing

carlisia:
Oh ok, I was putting the handlers somewhere else, haha.

Great heads up all around. I will comb the controllers. Thanks!

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -55,14 +55,15 @@ func NewCreateCommand(f client.Factory, use string) *cobra.Command {
...
-	BackupSyncPeriod time.Duration
-	Config           flag.Map
-	Labels           flag.Map
-	AccessMode       *flag.Enum
+	Name                  string
+	Provider              string
+	Bucket                string
+	Prefix                string
+	BackupSyncPeriod      time.Duration
+	StoreValidationPeriod time.Duration
```
ashish-amarnath, line 63: would `validationFrequency` be a better name for this field?

carlisia: Yes, yes indeed!

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -107,6 +109,10 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 		return errors.New("--backup-sync-period must be non-negative")
 	}
 
+	if o.StoreValidationPeriod < 0 {
```
ashish-amarnath, line 112: should we set this to default and continue?

pkg/controller/backup_storage_location_controller.go
```diff
@@ -0,0 +1,210 @@
...
+func NewBackupStorageLocationController(
+	namespace string,
+	defaultBackupLocation string,
+	defaultStoreValidationPeriod time.Duration,
+	backupLocationClient velerov1client.BackupStorageLocationsGetter,
+	backupStorageLocationLister velerov1listers.BackupStorageLocationLister,
+	newPluginManager func(logrus.FieldLogger) clientmgmt.Manager,
+	logger logrus.FieldLogger,
+) Interface {
+	if defaultStoreValidationPeriod <= 0 {
```
ashish-amarnath, line 60: if `defaultStoreValidationPeriod == 0` doesn't that mean validation is disabled?

pkg/controller/backup_storage_location_controller.go
```diff
@@ -0,0 +1,210 @@
...
+
+	pluginManager := c.newPluginManager(c.logger)
+	defer pluginManager.CleanupClients()
+
+	var unavailable []string
+	log := c.logger
+	for _, location := range locations {
+		locationName := location.Name
+		log = c.logger.WithField("backupLocation", locationName)
+
```
ashish-amarnath, line 108: Is `location.Spec.StoreValidationPeriod == nil` also a condition where we should skip validation?

pkg/controller/backup_storage_location_controller.go
```diff
@@ -0,0 +1,210 @@
...
+		lastValidation := location.Status.LastValidationTime
+		if lastValidation != nil {
+			log.Debug("Checking if location needs to be validated at this time")
+			nextValidation := lastValidation.Add(storeValidationPeriod)
+			if time.Now().UTC().Before(nextValidation) {
+				continue
+			}
+		}
+
+		log.Debug("Location is ready to be validated")
```
ashish-amarnath, line 132:
```suggestion
		log.Debug("Validating BackupStorageLocation")
```

ashish-amarnath: suggest using `BSL` instead of `location`.

pkg/controller/backup_storage_location_controller.go
```diff
@@ -0,0 +1,210 @@
...
+			err2 := c.patchBackupStorageLocation(location, func(r *velerov1api.BackupStorageLocation) {
+				r.Status.Phase = velerov1api.BackupStorageLocationPhaseUnavailable
+				r.Status.LastValidationTime = &metav1.Time{Time: time.Now().UTC()}
+			})
+			if err2 != nil {
+				unavailable = append(unavailable, errors.Wrapf(err2, "error updating location %q to %s status", locationName, velerov1api.BackupStorageLocationPhaseUnavailable).Error())
+			} else {
+				unavailable = append(unavailable, errors.Wrapf(err, "location %q is unavailable", locationName).Error())
+			}
+			log.Debug("Location has been updated to Unavailable")
```
ashish-amarnath, line 151: Suggest making this `log.Errorf` and also logging `err`

ashish-amarnath:
```suggestion
			log.Errorf("BSL marked as unavailable, reason: %v", err)
```

pkg/controller/backup_storage_location_controller.go
```diff
@@ -0,0 +1,210 @@
...
+			if err != nil {
+				unavailable = append(unavailable, errors.Wrapf(err, "error updating location %q to %s status", locationName, velerov1api.BackupStorageLocationPhaseAvailable).Error())
+			}
+			log.Debug("Location has been updated to Available")
+		}
+	}
+
+	var defaultFound bool
+	if len(unavailable) == len(locations) { // no BSL available
+		log.Errorf("There are no locations available, at least one available location is required: %s", strings.Join(unavailable, "; "))
```
ashish-amarnath, line 167: `unavailable` is used as a list of errors. This log message seems to use it as a list of unavailable BSLs.

pkg/controller/backup_storage_location_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	} else if len(unavailable) > 0 { // some but not all BSL unavailable
+		for _, location := range locations {
```
ashish-amarnath, line 169: Seems like we are looping through BSLs just to log if the `default` BSL is unavailable.
Can we move L#170-l#174 inside the loop at L105?

pkg/controller/backup_storage_location_controller.go
```diff
@@ -0,0 +1,210 @@
...
+			})
+			if err != nil {
+				unavailable = append(unavailable, errors.Wrapf(err, "error updating location %q to %s status", locationName, velerov1api.BackupStorageLocationPhaseAvailable).Error())
+			}
+			log.Debug("Location has been updated to Available")
+		}
+	}
+
+	var defaultFound bool
+	if len(unavailable) == len(locations) { // no BSL available
```
ashish-amarnath, line 166: I think we only care about two scenarios:
1. If none of the BSLs are available
2. If the default BSL is unavailable.
Could we add these checks into the `for` loop at L105 and smplify/remove L#165-L#180?

pkg/cmd/cli/backuplocation/create.go
carlisia: Not sure. What if the user for some reason expects that value to take, and we give them another value. Could this be possible?

pkg/controller/backup_storage_location_controller.go
carlisia: Yes so, this is confusing because of two flags with the same name. Mainly I set this up following the backup-sync-controller. We can remove this or change the condition here. 

This right here is a flag on the server. It is set so that `IF` `location.Spec.StoreValidationPeriod` further below is not set, we'll use this value. And so, if there wasn't a value passed thru the server either, it is set to `1min`.

We could change this to match the BSL flag, in other words, if `<=0` then don't ever validate. But I don't think we want to never validate. And I know that's not what you are saying either. Since we do have to validate at some interval, we might as well make it configurable thru the server.

pkg/controller/backup_storage_location_controller.go
carlisia: No. Only a value of zero is meant to skip validation. Any other value will set the frequency to that value (obvs), and nil means they didn't set it, which then the frequency will be set to the server value, either one set by the user, or the default. Pls be sure to see my comment up above.

pkg/controller/backup_storage_location_controller.go
carlisia: Ok I'll change these to `backup location` as it's used in other controllers. I'm trying to keep from using yet another way of referring to this resource!

pkg/controller/backup_storage_location_controller.go
carlisia: There's a 3rd case which we should care enough to log: a default backup was specified but never found. This and the case if all existing BSLs are invalid (1 error message in this case) we can only find out at the end of the iteration.

I cleaned up based on your suggestions tho, see what you think of the next version.

pkg/controller/backup_storage_location_controller.go
carlisia: Yes, I cleaned this up, thanks.

pkg/controller/backup_storage_location_controller.go
carlisia: Yes, good point. Renamed it.

pkg/controller/backup_storage_location_controller.go
carlisia: I removed it actually, now that this block is cleaner I think I was double logging in the debug case.

pkg/controller/backup_storage_location_controller.go
carlisia: Addendum: *we don't want to never validate a BSL >unless< the user specifically configures it that way.

carlisia:
At any rate.. relying on `len(unavailableErrors) == len(locations)` no longer works because due to the validation only occurring based on the validation frequency, the validation might be skipped for that reason so we don't know at the end of the iteration what that number truly is.

I'm going to add a method to return all BSLs that are in each phase, or maybe just the count for each, so this logging logic will be much cleaner and the rules clearer.

skriss:
@carlisia the kubebuilder convo reminded me that i do have a couple recommendations on how to structure this that (a) I think would make a potential migration to kubebuilder easier, and (b) would make testing easier, regardless of whether we move to kubebuilder or not. I'm not going to have time to write detailed notes until next week though, but here's a quick summary:

The main idea would be to keep the core "business logic" out of the controller package/struct, and extract it to its own package with a clearly defined interface. the controller should just handle responding to informer events, putting things on a work queue, taking things off the work queue, and then calling into the "business logic" package to actually handle the processing. This would be similar to (a) the server status request controller; and (b) the work I was doing in https://github.com/vmware-tanzu/velero/pull/2317.  Having all the business logic in the controller package and structs IMHO is not the pattern we want to be following going fwd.

carlisia:
Yes, I started thinking this very thing when I was thinking about the webhooks and, regardless, you're right that the structure needs to have the logic not in the controller. Glad for the nudge.

carlisia:
Moved it to Draft because likely will be layering this work on top of https://github.com/vmware-tanzu/velero/pull/2561.

carlisia:
Closing this PR. I'll be redoing this in a new PR using the kubebuilder/controller-runtime framekwork as per https://github.com/vmware-tanzu/velero/issues/2597.

PR #2493, opened 2020-05-01
Add status column to get BSL output

carlisia:
This closes https://github.com/vmware-tanzu/velero/issues/2489.

image

## Question
It seems to me as arbitrary to specify an empty status with `Available` as it would be with `Unavailable` - here: https://github.com/vmware-tanzu/velero/compare/master...carlisia:c-2489-bsl-column?expand=1#diff-c73073c993c3249edf0a5d8b6d30fa87R64

Should there be a third phase, maybe status `Unknown`? I'm not super set on this, we could make it available/unavailable.

### DIFF
--- a/pkg/cmd/util/output/backup_storage_location_printer.go
+++ b/pkg/cmd/util/output/backup_storage_location_printer.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 the Velero contributors.
+Copyright 2018, 2020 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


carlisia:
Or maybe `Unverified`.

This is because when the BSL gets created on the client side, it won't be validated until the BSL controller runs (see: https://github.com/vmware-tanzu/velero/pull/2490).

pkg/cmd/util/output/backup_storage_location_printer.go
```diff
@@ -55,13 +56,19 @@ func printBackupStorageLocation(location *v1.BackupStorageLocation) []metav1.Tab
 
 	accessMode := location.Spec.AccessMode
 	if accessMode == "" {
-		accessMode = v1.BackupStorageLocationAccessModeReadWrite
+		accessMode = velerov1api.BackupStorageLocationAccessModeReadWrite
+	}
+
+	status := location.Status.Phase
+	if status == "" {
+		status = velerov1api.BackupStorageLocationPhaseAvailable
```
skriss, line 64: I'd display this as `<Unknown>` or something along those lines, since we don't yet know if it's available.

Even if we define a third phase of `Unknown`, new CRs could still have an empty phase (until we add defaulting to our CRD specs), so I do think it makes sense to check for and handle an empty status.phase field here.

carlisia: Yes, make sense. Looks like this now:

image

PR #2496, opened 2020-05-04
Sync CSI API objects during backup sync

ashish-amarnath:

Fixes: #2145 

There is some overlap between this PR and #2448 
Will rebase once #2448 is merged.


pkg/controller/backup_sync_controller.go
```diff
@@ -262,6 +272,70 @@ func (c *backupSyncController) run() {
...
+					}
+				}
+
+				csiVolSnaps, err := backupStore.GetCSIVolumeSnapshots(backupName)
+				if err != nil {
+					log.WithError(errors.WithStack(err)).Error("Error getting CSI volumesnapshots for this backup from backup store")
+					continue
+				}
+
+				log.Infof("Syncing %d CSI volumesnapshots in backup", len(csiVolSnaps))
```
skriss, line 310: Wouldn't we need to do the whole dynamic -> static provisioning change here like we do for the plugins?

ashish-amarnath: I don't think that would be necessary, because:
1. With syncing a backup into the cluster, we are merely re-creating the state that was captured while performing the backup and not exactly restoring the backup.
2. Also, the motivation to sync the CSI API objects is to ensure that the storage snapshots are cleaned up on backup deletion/expiry.
3. We will be doing the dynamic -> static provisioning sequence when we restore from the backup.

pkg/cmd/server/server.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2017, 2019 the Velero contributors.
+Copyright 2017-2020 the Velero contributors.
```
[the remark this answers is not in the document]
ashish-amarnath: as in `2017, 2018, 2019, 2020` or just `2020`?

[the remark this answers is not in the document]
skriss: this one would be 2017, 2019, 2020

pkg/controller/backup_sync_controller.go
skriss: OK - I was under the impression we'd need to change the VSCs to have a snapshot source rather than a volume source, but if that's not necessary for triggering a deletion, then :+1:.

Related, is there any reason to sync in the VolumeSnapshots? It sounds like we won't be using them for restore, and directly deleting the VSCs should be sufficient, per the previous PR - right?

pkg/persistence/mocks/backup_store.go
```diff
@@ -1,12 +1,32 @@
-// Code generated by mockery v1.0.0. DO NOT EDIT.
+/*
+Copyright 2020 the Velero contributors.
```
nrb, line 2: Thanks for moving this file in. Since I don't think my describe PR's going to use it, I'll just remove it from mine to prevent any merge conflicts.

pkg/persistence/object_store.go
```diff
@@ -59,6 +60,8 @@ type BackupStore interface {
 	GetPodVolumeBackups(name string) ([]*velerov1api.PodVolumeBackup, error)
 	GetBackupContents(name string) (io.ReadCloser, error)
 	// TODO(nrb-csi): Any Get methods relevant to the CSI VolumeSnapshots should be added with the client-side PRs.
```
nrb, line 62: I think with the 2 methods being implemented here, the comment can be deleted.

Also, I misunderstood what the `Get` methods here were being used for. They're used by the sync controller, while Velero clients use a different method to do the downloading.

pkg/controller/backup_sync_controller.go
nrb: I second the question about restoring the VolumeSnapshots - my understanding was that we'd just sync the VSCs in order to have a way to garbage collect storage-side snapshots associated with an expired backup.

ashish-amarnath: That's correct. We don't have to restore volumesnapshots. I'll remove that. Thanks for catching!

pkg/util/kube/utils.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2017, 2019 the Velero contributors.
+Copyright 2017, 2019, 2020 the Velero contributors.
```
skriss, line 2: I think you can revert the change to this file now, since we're not going to sync VolumeSnapshots.

ashish-amarnath:
Steve, about your review [comment](https://github.com/vmware-tanzu/velero/pull/2496#pullrequestreview-405301817). 
To clarify, we don't restore VSCs per se in the plugin. We create a VSC, for static binding, during the restore of the VS. 
We could create statically bindable VSCs on backup sync. However this would need the restore sequence differentiate if a backup was sync'ed or not to decide whether or not to create the VSC for static binding, making the restore and the sync process more tightly coupled. Keeping the restore logic fully in the plugin lets us update the plugin independently of the core code.

skriss:
LGTM assuming it looks fine in testing! Will wait for rebase.

ashish-amarnath:
Rebased. Yes this looked good in testing!

skriss:
you mentioned overlap between this and #2448 - since this is about ready, do you want to merge it as-is and have @nrb rebase #2448? @nrb, that okay with you, if so?

ashish-amarnath:
Overlaps were:
1. https://github.com/vmware-tanzu/velero/pull/2496/files#diff-70eae9766e21a6905f2a2aa03d98599a: This will be removed from #2448 https://github.com/vmware-tanzu/velero/pull/2496#discussion_r419734243
2. https://github.com/vmware-tanzu/velero/pull/2496/files#diff-52bb234b1ac2d5dbab0c69c812a636cb. https://github.com/vmware-tanzu/velero/pull/2448/files#diff-52bb234b1ac2d5dbab0c69c812a636cb I am OK waiting to go after #2448

nrb:
Actually, lets have this go first.

My reasoning is that now that we're syncing VSCs for sure, we can use that in describe command very similarly to how we get PodVolumeBackups.

Also, it's a lot more complete than mine, so let's get this in!

PR #2502, opened 2020-05-05
Use label selector to select VSCs to persist in backup

ashish-amarnath:

As part of a backup, we do not create dangling VSCs. For that reason, this change is a "nice-to-have". 
This PR simplifies how we discover CSI volumesnapshotcontents to persist as part of the the backup

cc @nrb


skriss:
> As part of a backup, we do not create dangling VSCs. For that reason, this change is a "nice-to-have".

Yeah, that was my question after you mentioned this on this AM's call :) This looks good though.

PR #2503, opened 2020-05-05
Allow feature flags to be passed from install CLI

ashish-amarnath:

### DIFF
--- a/changelogs/unreleased/2503-ashish-amarnath
+++ b/changelogs/unreleased/2503-ashish-amarnath
@@ -1,0 +1,1 @@
+allow feature flags to be passed from install CLI
--- a/pkg/cmd/cli/install/install.go
+++ b/pkg/cmd/cli/install/install.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 the Velero contributors.
+Copyright 2019,2020 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -69,6 +69,7 @@ type InstallOptions struct {
 	NoDefaultBackupLocation           bool
 	CRDsOnly                          bool
 	CACertFile                        string
+	Features                          string
 }
 
 // BindFlags adds command line values to the options struct.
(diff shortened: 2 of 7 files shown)


changelogs/unreleased/2503-ashish-amarnath
```diff
@@ -0,0 +1 @@
+allow feature flags to be passed from install CLI #2503
```
skriss, line 1: you can drop the PR # from the text :)

pkg/cmd/cli/install/install.go
```diff
@@ -101,6 +102,7 @@ func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.Var(&o.Plugins, "plugins", "Plugin container images to install into the Velero Deployment")
 	flags.BoolVar(&o.CRDsOnly, "crds-only", o.CRDsOnly, "only generate CustomResourceDefinition resources. Useful for updating CRDs for an existing Velero install.")
 	flags.StringVar(&o.CACertFile, "cacert", o.CACertFile, "file containing a certificate bundle to use when verifying TLS connections to the object store. Optional.")
+	flags.StringVar(&o.Features, "features", o.Features, "comma separated list of Velero feature flags to be set on the Velero deployment")
```
skriss, line 105: I think we probably want to set these on the restic DaemonSet too, if we're creating it - @nrb WDYT?

changelogs/unreleased/2503-ashish-amarnath
ashish-amarnath: copy-pasta

pkg/cmd/cli/install/install.go
nrb: Yep, I agree.

ashish-amarnath: I can do that. I'll need to confirm that this doesn't trigger the `unbound flags` error

ashish-amarnath: done! Testing also looked good. No `unknown flag` errors

skriss, line 105: let's update the flag help text to mention that they're passed to the restic daemonset too (if it's being installed)

PR #2509, opened 2020-05-06
v1.4 upgrade instructions

### LINKED ISSUE #2495: Document v1.4 upgrade process
This is just a list of items that need to be documented for users to upgrade for v1.4 specifically. Please update the description to add anything missing.

 * [ ] Update CRDs so that the progress fields are included (#2440)

skriss:

closes #2495 

Will keep this in draft until just before the release of the beta to avoid any confusion to users. And we'll obviously have to update this again for the GA release.

Do you all think I should include plugin upgrade instructions here as well? We don't really have a clear process yet for how to communicate new plugin versions and upgrades. Also, we may not want to include instructions only for the ones we maintain, since we're trying not to give them preference. So, open to suggestions here..

### DIFF
--- a/site/docs/master/upgrade-to-1.4.md
+++ b/site/docs/master/upgrade-to-1.4.md
@@ -1,17 +1,18 @@
-# Upgrading to Velero 1.3
+# Upgrading to Velero 1.4
 
 ## Prerequisites
 
-- Velero [v1.3.1][5], [v1.3.0][4] or [v1.2][3] installed.
+- Velero [v1.3.x][4] installed.
 
-If you're not yet running at least Velero v1.2, see the following:
+If you're not yet running at least Velero v1.3, see the following:
 
 - [Upgrading to v1.1][1]
 - [Upgrading to v1.2][2]
+- [Upgrading to v1.3][3]
 
 ## Instructions
 
-1. Install the Velero v1.3 command-line interface (CLI) by following the [instructions here][0].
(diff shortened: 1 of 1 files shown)


skriss:
I don't have any further changes planned for this, it's ready for review @ashish-amarnath @carlisia @nrb. Just holding it in draft status until we're ready to cut the beta, so we don't confuse anybody.

site/docs/master/upgrade-to-1.4.md
```diff
@@ -23,23 +24,29 @@ If you're not yet running at least Velero v1.2, see the following:
...
         --namespace velero
 
     # optional, if using the restic daemon set
     kubectl set image daemonset/restic \
-        restic=velero/velero:v1.3.2 \
+        restic=velero/velero:v1.4.0-beta.1 \
         --namespace velero
     ```
 
+1. Update the Velero custom resource definitions (CRDs) to include the new backup progress fields:
```
ashish-amarnath, line 44: Should this step be before updating the container image?

skriss: eh, shouldn't really matter one way or another - the updated image will run just fine without the CRD updates, and if the CRDs aren't updated, backups just won't get progress updates.

ashish-amarnath: As long as the pod doesn't required to be restarted for the new CRDs to be picked up it's fine.

PR #2511, opened 2020-05-06
📖 Document --features

ashish-amarnath:

### DIFF
--- a/site/docs/master/csi.md
+++ b/site/docs/master/csi.md
@@ -8,7 +8,7 @@ The following are the prerequisites for using Velero to take Container Storage I
 
  1. The cluster is Kubernetes version 1.17 or greater.
  1. The cluster is running a CSI driver capable of support volume snapshots at the [v1beta1 API level](https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-cis-volume-snapshot-beta/).
- 1. The Velero server is running with the `--features EnableCSI` feature flag to enable CSI logic in Velero's core.
+ 1. The Velero server is running with the `--features EnableCSI` feature flag to enable CSI logic in Velero's core. See [Enabling Features][1] for more information.
  1. The Velero [CSI plugin](https://github.com/vmware-tanzu/velero-plugin-for-csi/) is installed to integrate with the CSI volume snapshot APIs.
  1. When restoring CSI volumesnapshots across clusters, the name of the CSI driver in the destination cluster is the same as that on the source cluster to ensure cross cluster portability of CSI volumesnapshots
 
@@ -27,3 +27,6 @@ This section documents some of the choices made during implementation of the Vel
 
 Velero's support level for CSI volume snapshotting will follow upstream Kubernetes support for it, and will reach general availability sometime
 after volume snapshotting is GA in upstream Kubernetes. Beta support is expected to launch in Velero v1.4.
+
+
+[1]: customize-installation.md#enable-server-side-features
--- a/site/docs/master/customize-installation.md
+++ b/site/docs/master/customize-installation.md
@@ -5,6 +5,9 @@
   - [Install in any namespace](#install-in-any-namespace)
   - [Use non-file-based identity mechanisms](#use-non-file-based-identity-mechanisms)
   - [Enable restic integration](#enable-restic-integration)
+  - [Enable features](#enable-features)
+    - [Enable server side features](#enable-server-side-features)
+    - [Enable client side features](#enable-client-side-features)
   - [Customize resource requests and limits](#customize-resource-requests-and-limits)
   - [Configure more than one storage location for backups or volume snapshots](#configure-more-than-one-storage-location-for-backups-or-volume-snapshots)
   - [Do not configure a backup storage location during install](#do-not-configure-a-backup-storage-location-during-install)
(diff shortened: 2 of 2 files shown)


site/docs/master/customize-installation.md
```diff
@@ -42,6 +43,16 @@ By default, `velero install` does not install Velero's [restic integration][3].
 
 If you've already run `velero install` without the `--use-restic` flag, you can run the same command again, including the `--use-restic` flag, to add the restic integration to your existing install.
 
+## Enable features
+
+New features in Velero will be released as beta features behind feature flags which are not enabled by default. These features can be enabled using the `--features` flag to the `velero install` command. This flag takes as value, a comma separated list of feature flags to enable. As an e.g. [CSI snapshotting of PVCs][10] can be enabled using `--features=EnableCSI` flags to the `velero install` command.
```
carlisia, line 48:
```suggestion
New features in Velero will be released as beta features behind feature flags which are not enabled by default. These features can be enabled using the `--features` flag to the `velero install` command. This flag takes as value a comma separated list of feature flags to enable. As an example, [CSI snapshotting of PVCs][10] can be enabled using `--features=EnableCSI` flags in the `velero install` command.
```

carlisia: It's either "`in` or `on` the velero install command" (I always get that mixed up).

site/docs/master/customize-installation.md
```diff
@@ -42,6 +43,16 @@ By default, `velero install` does not install Velero's [restic integration][3].
...
+
+Similarly, features may be disabled by removing the corresponding feature flags from the `--features` flag.
```
carlisia, line 50: It'd be nice if right here was indicated possible ways to do that.

site/docs/master/customize-installation.md
```diff
@@ -42,6 +43,16 @@ By default, `velero install` does not install Velero's [restic integration][3].
...
+
+The feature flags, passed to `velero install` will be passed to the Velero deployment and the `restic` daemon set, if `--use-restic` flag is used.
```
carlisia, line 52:
```suggestion
The feature flags, passed to `velero install` will be passed to the Velero deployment, and also to the `restic` daemon set if `--use-restic` flag is used.
```

site/docs/master/customize-installation.md
```diff
@@ -42,6 +43,16 @@ By default, `velero install` does not install Velero's [restic integration][3].
...
+
+Enabling and disabling feature flags will require modifying the Velero deployment and also the restic daemonset. This may be done easily by uninstalling and re-installing Velero.
```
carlisia, line 54:
```suggestion
Enabling and disabling feature flags will require modifying the Velero deployment and also the restic daemonset. This may be done by uninstalling and re-installing Velero, or by editing the resource files in the cluster.
```

carlisia: This could be moved to LN50.

site/docs/master/customize-installation.md
ashish-amarnath: ways to not supply the value to the flag?

carlisia: Sorry, no I mean to disable/remove it.

ashish-amarnath: That involves showing how to not pass the value to flag. Not entirely sure how to document that. Suggestions?

carlisia: I think it's the text on LN54. At least that is what I'm talking about :)

site/docs/master/customize-installation.md
ashish-amarnath: done!

site/docs/master/customize-installation.md
ashish-amarnath: PTAL at the new edits 😄

site/docs/master/customize-installation.md
ashish-amarnath: done!

site/docs/master/customize-installation.md
ashish-amarnath: PTAL at the new edits 😄

site/docs/master/customize-installation.md
```diff
@@ -42,6 +45,39 @@ By default, `velero install` does not install Velero's [restic integration][3].
 
 If you've already run `velero install` without the `--use-restic` flag, you can run the same command again, including the `--use-restic` flag, to add the restic integration to your existing install.
 
+## Enable features
+
+New features in Velero will be released as beta features behind feature flags which are not enabled by default. For a full listing of available feature flags, please refer `pkg/features/features.go`.
```
skriss, line 50: it'd be nice to include a link to `features.go` here.

site/docs/master/csi.md
```diff
@@ -27,3 +27,6 @@ This section documents some of the choices made during implementation of the Vel
 
 Velero's support level for CSI volume snapshotting will follow upstream Kubernetes support for it, and will reach general availability sometime
 after volume snapshotting is GA in upstream Kubernetes. Beta support is expected to launch in Velero v1.4.
+
+
+[1]: customize-installation.md#enable-features#enable-server-side-features
```
skriss, line 32: not sure which section you wanted to link to - you got both :D

site/docs/master/customize-installation.md
ashish-amarnath: My bad. That's not a valid file. Our feature flag constants are not limited to a single file. Going to link our constants file here instead of the feature code.

nrb: I don't like linking to the Go code, mostly because in general I hate making people read code to understand a project. That being said, I also hate duplicating this kind of stuff, so I don't have a great alternative to generating docs based on the code right now.

My 2c - we keep the number of feature flags limited and be aggressive about deprecating them when the time is right so we don't have to worry about a crazy long list. If they're defined on an installation or in a client's config but no longer used, that should be a no-op.

skriss: :+1: it might be helpful in the future to move them to their own dedicated file, for clarity, but we can do that separately.

ashish-amarnath: +1 for keeping the feature flag list small.
But we should also have a way for users to discover feature flags. That's the problem I am looking to solve here and this was the easiest way to do that today.
Some options for us to consider:
1. having a features page on the website
2. having a CLI command list all features

My vote is for option2

nrb: Yeah, for today, I think this is the best option.

Per the design I wrote, I explicitly opted not to have a fully enumerated list of features. We can revisit that if it's easier, and I think Kubernetes does this. For now, we can log future action IMO.

site/docs/master/csi.md
skriss: Still need to resolve this

ashish-amarnath: forgot to push 🤦

site/docs/master/customize-installation.md
ashish-amarnath: Created https://github.com/vmware-tanzu/velero/issues/2515

carlisia:
@ashish-amarnath do you mind adding:

>Feature flags can be disabled completely with velero client config set features=.

I think is not so obvious but very handy info.

ashish-amarnath:
@carlisia added the 
```
velero client config set features=
```.
This will need another round of approves

PR #2533, opened 2020-05-12
add backupPartialFailure metric

Samet-MohamedAmin:


pkg/metrics/metrics.go
```diff
@@ -34,6 +34,7 @@ const (
 	backupAttemptTotal            = "backup_attempt_total"
 	backupSuccessTotal            = "backup_success_total"
 	backupPartialFailureTotal     = "backup_partial_failure_total"
+	backupPartialFailure          = "backup_partial_failure"
```
skriss, line 37: how does this differ from the existing "backup_partial_failure_total" metric (just above)?

Samet-MohamedAmin: Execuse me. You're totally right. I'm working on it.
I need a metric that fire 1 if there is "backupPartialFailure" at the actual moment. It should be the same as "backupPartialFailureTotal" but not cumulative. 
I need that to have an idea of how much "backupPartialFailure" are in a period of time.

Samet-MohamedAmin: I think the lines from 345 should be like the following:
``` go
	if c, ok := m.metrics[backupPartialFailure].(*prometheus.GaugeVec); ok {
		c.WithLabelValues(backupSchedule).Set(1)
	}
```

skriss: > I need a metric that fire 1 if there is "backupPartialFailure" at the actual moment

Couldn't you look at changes in the existing "total" metric? So any time the value changed, you'd know there was a new partial failure? I'm not sure that having a binary metric indicating either 1 or 0 will work very well - does that indicate there was a partial failure in the last second? minute? hour? what if there were more than one?

carlisia:
Oops, did not mean to close this.

pkg/metrics/metrics.go
Samet-MohamedAmin: I'm currently using [changes](https://prometheus.io/docs/prometheus/latest/querying/functions/#changes) function to fire errors.
But unfortunately it does not capture the first error as it tracks changes and the first value does not represent a change.
This is a common problem in prometheus. see: https://stackoverflow.com/questions/43875918/prometheus-alert-on-change-in-value

skriss:
@Samet-MohamedAmin if you could describe in more detail what you're trying to accomplish, maybe we could help give some more guidance.

Would having a metric that indicated whether the **last/most recent** backup from each schedule succeeded or not be helpful and solve your need?

skriss:
This looks to be inactive so I'm going to close it out. If you're still interested in working on this, it might be helpful to first open an issue describing what you're trying to accomplish, answering some of the above questions, so we can figure out an approach that works.

Regardless, thanks for your interest in the project!

PR #2542, opened 2020-05-15
increment restic volumesnapshot count after successful pvb create

ashish-amarnath:

The failure to create PVB objects could be causing restic volumesnapshot operations to time out.

### DIFF
--- a/pkg/restic/backupper.go
+++ b/pkg/restic/backupper.go
@@ -158,11 +158,11 @@ func (b *backupper) BackupPodVolumes(backup *velerov1api.Backup, pod *corev1api.
 		}
 
 		volumeBackup := newPodVolumeBackup(backup, pod, volume, repo.Spec.ResticIdentifier, pvc)
-		numVolumeSnapshots++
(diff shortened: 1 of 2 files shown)


pkg/restic/backupper.go
```diff
@@ -158,11 +158,12 @@ func (b *backupper) BackupPodVolumes(backup *velerov1api.Backup, pod *corev1api.
 		}
 
 		volumeBackup := newPodVolumeBackup(backup, pod, volume, repo.Spec.ResticIdentifier, pvc)
-		numVolumeSnapshots++
 		if volumeBackup, err = b.repoManager.veleroClient.VeleroV1().PodVolumeBackups(volumeBackup.Namespace).Create(volumeBackup); err != nil {
 			errs = append(errs, err)
+			log.Errorf("Failed to create volumeBackup %s for volume %s in pod %s/%s: %v", volumeBackup.Name, pod.Namespace, pod.Name, err)
```
skriss, line 163: We usually try not to both log *and* return errors, since that usually leads to double-logging. In this case I would just return it.

skriss:
:+1: let's add a changelog as well (sorry I didn't note this the first time), then LGTM

skriss:
@carlisia / @nrb PTAL 🙏 let's get this one in for 1.4

PR #2543, opened 2020-05-18
design: progress on backup/restore operations by volume snapshotters

vishnuitta:
This PR contains design proposal using which volume snapshotters can let know its users about the progress on backup/restore operations.

Signed-off-by: Vitta <vitta@mayadata.io>


vishnuitta:
just want to make sure that nothing is pending on me to make this PR go to next stage.. please let me know if thats not the case
sorry for asking it..

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,238 @@
...
+
+(Omitted, see introduction)
+
+## High-Level Design
+
+### Progress of backup operation handled by volume snapshotter
+
+Progress will be updated by volume snapshotter in VolumePluginBackup CR which is specific to that backup operation.
+
+### Progress of restic operation handled by volume snapshotter
```
nrb, line 33: Restic is not currently handled by a volumesnapshotter plugin; are you proposing that we change it to be handled by one?

Or is this meant to say _restore_ operation?

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,238 @@
...
+- Provide uniform way of visibility into backup/restore operations performed by volume snapshotters
+
+## Non Goals
+- Plugin implementation for this approach
+
+## Background
+
+(Omitted, see introduction)
+
+## High-Level Design
```
nrb, line 27: Can you add a section for backwards compatibility here? How might a plugin that supports this new CR behave on a version of Velero that doesn't support the CR? Would (or should) it error immediately on install, or at back up time?

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,238 @@
...
+
+But, this involves good amount of changes and needs a way for backward compability.
+
+As volume plugins are mostly K8s native, its fine to go ahead with current limiation.
+
+### Update Backup CR
+Instead of creating new CRs, plugins can directly update the status of Backup CR. But, this deviates from current approach of having seperate CRs like PodVolumeBackup/PodVolumeRestore to know operations progress.
+
+### Restricting on name rather than using labels
+Instead of using labels to identify the CR related to particular backup on a volume, restrictions can be placed on the name of VolumePluginBackup CR to be same as the value returned from CreateSnapshot.
```
nrb, line 233: Note that values in labels should be passed through https://github.com/vmware-tanzu/velero/blob/master/pkg/label/label.go#L35, because the values are not completely free-form. The main limitation is length.

However, so long as that function is consistently used for writing and querying, then it should get the same backup every time.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,238 @@
...
+
+In order to know the CR created for the particular backup of a volume, volume snapshotters adds following labels to CR:
+`velero.io/backup-name` with value as Backup Name, and,
+`velero.io/volume-name` with value as volume that is undergoing backup
+Backup name being unique won't cause issues like duplicates in identifying the CR.
+
+Though no restrictions are required on the name of CR, as a general practice, volume snapshotter can name this CR with the value same as return value of CreateSnapshot.
+
+After return from `CreateSnapshot` in `takePVSnapshot`, if VolumePluginBackup CR exists for particular backup of the volume, velero adds this CR to `backupRequest`.
+During persistBackup call, this CR also will be backed up to backup location.
```
nrb, line 125: So these would need to be written in a `json.gz` file, similar to how the snapshots are now?

It may be helpful to create a section in the design for "core" Velero server/client changes. I think the only things that need to be changed are as follows, based on what I've read in this document.

* Creation of the VolumePluginBackup/VolumePluginRestore CRDs at installation time
* Persistence of VolumePluginBackup CRs towards the end of the back up operation
* Sync of VolumePluginBackup/VolumePluginRestore CRDs at the same time as backup synchronization.

Otherwise, volumesnapshotter plugins themselves are responsible for interacting with the VolumePlugin(Backup|Restore) CRs, and Velero's core won't change them.

Then, to see this reported, the `velero describe` commands would need to be aware of the VolumePlugin(Backup|Restore) CRs associated with a Backup CR in order to display the progress fields.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,238 @@
...
+	// This will be used to relate with output in 'velero describe backup'
+	SnapshotID string `json:"snapshotID"`
+
+	// Backup name
+	Backup string `json:"backup"`
+
+	// Plugin name
+	Plugin string `json:"plugin"`
+
+	// PluginSpecific are a map of key-value pairs that plugin want to provide
```
nrb, line 144: Can you give an example of what values might go in here, and how it might differ from [ConfigMaps that we use](https://velero.io/docs/v1.4/custom-plugins/#plugin-configuration)?

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,238 @@
...
+### Update Backup CR
+Instead of creating new CRs, plugins can directly update the status of Backup CR. But, this deviates from current approach of having seperate CRs like PodVolumeBackup/PodVolumeRestore to know operations progress.
+
+### Restricting on name rather than using labels
+Instead of using labels to identify the CR related to particular backup on a volume, restrictions can be placed on the name of VolumePluginBackup CR to be same as the value returned from CreateSnapshot.
+But, this can cause issue when volume snapshotter just crashed without returning snapshot id to velero.
+
+## Security Considerations
+
+N/A
```
nrb, line 238: Are there any sort of RBAC considerations keeping plugins from modifying VolumePlugin(Backup|Restore) CRs that were created by another plugin?

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,238 @@
...
+	CompletionTimestamp *metav1.Time `json:"completionTimestamp,omitempty"`
+
+	// Progress holds the total number of bytes of the volume and the current
+	// number of backed up bytes. This can be used to display progress information
+	// about the backup operation.
+	// +optional
+	Progress VolumeOperationProgress `json:"progress,omitempty"`
+}
+
+type VolumeOperationProgress struct {
```
ashish-amarnath, line 95: This progress struct makes it seem like the snapshotter is performing a byte-by-byte copy of the data in the volume. Do we know which volume snapshotter plugins operate this way and which amongst them offer fine level of detail into their volume snapshot progress?

nrb:
@dsu-igeek I would appreciate your eyes on this, even a quick glance. Would this proposal work for the vSphere plugin?

design/plugin-backup-and-restore-progress-design.md
ashish-amarnath: lol. I would expect that too!
The specific question I was implying, is this something that is specific to openEBS?
How would other plugins be able to make use of these CRDs?

design/plugin-backup-and-restore-progress-design.md
vishnuitta: I will add a section with the details.. but, I'm thinking of behavior to be - graceful handling.
Plugin that supports this CR should handle the situation gracefully when CRDs are not installed. It can handle the errors occured during creation/updation of the CRs.

design/plugin-backup-and-restore-progress-design.md
vishnuitta: thats right @nrb .. its a typo. Will fix it

design/plugin-backup-and-restore-progress-design.md
vishnuitta: I picked this Progress struct from PodVolumeBackup CR (https://github.com/vmware-tanzu/velero/blob/master/pkg/apis/velero/v1/pod_volume_operation_progress.go).
and OpenEBS, being block device provider, provides byte-level granularity.

design/plugin-backup-and-restore-progress-design.md
vishnuitta: sure. I will add a section @nrb. And, the only changes that are required in "core" are those that you mentioned.
Regarding 'velero describe' support for these CRs, if I'm not wrong, 'velero describe' is not aware of PodVolume(Backup|Restore) CRs as well. To keep this PR simple, I thought - we can take it for these new CRs also when PodVolume(Backup|Restore) CRs support is added to 'velero describe'.

design/plugin-backup-and-restore-progress-design.md
vishnuitta: I hadn't thought any specific, but, added a way so that plugins can provide more info specific to this restore operation to users.
Configmaps in plugin is related to config at plugin level. This field is specific to details regarding this restore operation.

design/plugin-backup-and-restore-progress-design.md
vishnuitta: thanks for info. Will update PR according to this.

design/plugin-backup-and-restore-progress-design.md
vishnuitta: >  How would other plugins be able to make use of these CRDs?
This should be able to be used by any plugin by filling details of the amount of data that it need to backup and amount of data that it already backed up.
Existing restic operator in velero also fills similar details.

ashish-amarnath: Also, looks like there is quite a bit of overlap between the existing [pod+volume_backup](https://github.com/vmware-tanzu/velero/blob/master/pkg/apis/velero/v1/pod_volume_backup.go) CRD and what is in this proposal.
Is there a reason why we might not be able to re-use/extend the existing pod volume backup CRDs?

nrb: PodVolumeBackups are currently only used for volumes protected by restic. They are not created at all for snapshots created via plugins.

I'm a little wary of overloading them too much, but not completely averse to the idea, depending on the details.

design/plugin-backup-and-restore-progress-design.md
nrb: Currently, no, it would not be aware of these new CRDs, but we could make it support them.

Within the backup describer, we have [a function for describing PodVolumeBackups](https://github.com/vmware-tanzu/velero/blob/master/pkg/cmd/util/output/backup_describer.go#L381), as well as [a function for describing VolumeSnapshotContents](https://github.com/vmware-tanzu/velero/blob/master/pkg/cmd/util/output/backup_describer.go#L493). 

I would think describing these new CRs would work roughly the same way, where they would [get queried from the Kube API](https://github.com/vmware-tanzu/velero/blob/master/pkg/cmd/cli/backup/describe.go#L95) and then [passed to the top level Describe function](https://github.com/vmware-tanzu/velero/blob/master/pkg/cmd/cli/backup/describe.go#L101).

I don't know that all this detail needs to go into the design document, but this is kind of what I envisioned the scope of work to be.

nrb:
> I would also like to understand how we might use something like this in the CSI snapshot APIs.

I'm unclear how we might be able to get information from the CSI snapshot APIs. I'd have to look at how they work again, as right now I think the CSI drivers are more or less a black box to Velero, and we don't have a way of asking for progress. I don't believe there's more than one field for bytes written, just the final total for `restoreSize`.

design/plugin-backup-and-restore-progress-design.md
nrb: I think using PodVolumeBackups would require modifying the controller at https://github.com/vmware-tanzu/velero/blob/master/pkg/controller/pod_volume_backup_controller.go to ignore any being created for use with plugins. We could do this, but I don't know exactly what the changes necessary would be at the moment. Similar goes for the restore side.

design/plugin-backup-and-restore-progress-design.md
nrb: As mentioned on Slack, I think the progress tracking is information that's really only necessary when scoped to an in-progress back up operation. When a `Backup` is in a `Completed` or `Failed` state, we know it's at 100% completion.

Given that, does it make sense to synchronize these CRs across clusters?

design/plugin-backup-and-restore-progress-design.md
vishnuitta: fixed it

design/plugin-backup-and-restore-progress-design.md
vishnuitta: done.. please have a look at it

design/plugin-backup-and-restore-progress-design.md
vishnuitta: updated

design/plugin-backup-and-restore-progress-design.md
vishnuitta: marking it as resolved as per above comment.. please reopen if more info is required

design/plugin-backup-and-restore-progress-design.md
vishnuitta: Synchronizing VolumePluginBackup CR to another cluster will help user to know the backup size, number of volumes in it even before user triggers restore of the backup.

skriss:
Couple helpful links from today's call:

internal `Snapshot` type -- https://github.com/vmware-tanzu/velero/blob/master/pkg/volume/snapshot.go
creation of `Snapshots` -- https://github.com/vmware-tanzu/velero/blob/master/pkg/backup/item_backupper.go#L477-L489
serialization of `Snapshots` to JSON for uploading to object storage as a file -- https://github.com/vmware-tanzu/velero/blob/master/pkg/controller/backup_controller.go#L667-L671
fetching of `Snapshots` from object storage on restore -- https://github.com/vmware-tanzu/velero/blob/master/pkg/controller/restore_controller.go#L450-L453
using the `Snapshots` for restore: https://github.com/vmware-tanzu/velero/blob/master/pkg/restore/restore.go#L195 and https://github.com/vmware-tanzu/velero/blob/master/pkg/restore/pv_restorer.go#L76-L104

design/plugin-backup-and-restore-progress-design.md
vishnuitta: One thing which I had in my mind when added this was specific to OpenEBS plugin, but, also thought that every plugin may have their specific things to convey to user about that particular backup/restore operation.
OpenEBS plugin supports incremental snapshots also. So, an entry can be added to this map which provides name of its parent backup.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,431 @@
...
+
+In backupSyncController, it checks for any VolumePluginBackup CRs that need to be synced from backup location, and syncs them to cluster if needed.
+
+VolumePluginBackup will be useful as long as backed up data is available at backup location. When the Backup is deleted either by manually or due to expiry, VolumePluginBackup also can be deleted.
+
+`processRequest` of `backupDeletionController` will perform deletion of VolumePluginBackup before volumesnapshotter's DeleteSnapshot is called.
+
+#### Backward compatibility:
+
+As the VolumePluginBackup CR is backed up instead of `volume.Snapshot`, to provide backward compatibility, CR will be backed up to the same file i.e., `#backup-volumesnapshots.json.gz` file in the backup location.
```
nrb, line 177: Is the VolumePluginBackup information being added to the `<backupname>-volumesnapshots.json.gz` file, or replacing the contents? I'm unclear on that point.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,431 @@
...
+	metav1.TypeMeta `json:",inline"`
+
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// +optional
+	Spec VolumePluginBackupSpec `json:"spec,omitempty"`
+
+	// +optional
+	Status VolumePluginBackupStatus `json:"status,omitempty"`
```
nrb, line 147: I don't see `VolumePluginBackupStatus` defined in Approach 1. Was it supposed to be, or was `SnapshotStatus` supposed to be used here?

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,431 @@
...
+}
+```
+
+Notice that VolumePluginBackupSpec is same as volume.SnapshotSpec.
+
+For every backup operation of volume, Velero creates VolumePluginBackup CR before calling volumesnapshotter's CreateSnapshot API.
+
+In order to know the CR created for the particular backup of a volume, Velero adds following labels to CR:
+- `velero.io/backup-name` with value as Backup Name, and,
```
nrb, line 156: This label is added to the VolumePluginBackup CR? Is that for easier query?

design/plugin-backup-and-restore-progress-design.md
vishnuitta: CR itself will be backed up as <backupname>-volumesnapshots.json.gz. Its a kind of replacing it.
I will correct that sentence.

design/plugin-backup-and-restore-progress-design.md
vishnuitta: Its for easier query for plugin to identify the CR so that it can update the status to that CR.

design/plugin-backup-and-restore-progress-design.md
vishnuitta: It is supposed to be same as SnapshotStatus. Will update it.

vishnuitta: updated

design/plugin-backup-and-restore-progress-design.md
vishnuitta: Updated

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,434 @@
...
+
+	// +optional
+	Spec VolumePluginBackupSpec `json:"spec,omitempty"`
+
+	// +optional
+	Status VolumePluginBackupStatus `json:"status,omitempty"`
+}
+```
+
+Notice that VolumePluginBackupSpec is same as volume.SnapshotSpec, and, VolumePluginBackupStatus is same as volume.SnapshotStatus
```
skriss, line 152: Why do we need these type aliases? Is this for consistency with the restore side, since we need a new CRD there?

vishnuitta: Idea of Approch 1 is to convert `volume.Snapshot` as a CRD.
As CR Kind is `VolumePluginBackup` and general practice is to have structs for `<CRKind>Spec` and `<CRKind>Status`, I done the type aliases to `volume.SnapshotSpec` and `SnapshotStatus`.

vishnuitta: For backward compatibility on restore side, consider below cases:
- older version of Velero, older json.gz file (backupname-volumesnapshots.json.gz)
- older version of Velero, newer json.gz file
- newer version of Velero, older json.gz file
- newer version of Velero, newer json.gz file

First and last should be fine.
For second case, if I'm not wrong, decode in `GetBackupVolumeSnapshots` on the restore side should fill only required fields of older version and should work.
For third case, after decode, `metadata.name` will be empty. Still, just returning []`volume.Snapshot` created from []VolumePluginBackupSpec should be fine.
Correct me if I'm wrong.

If decode fails in second case during implementation, this CR need to be backed up to different file. And, for backward compatibility, newer code should check for old file existence, and follow older code if exists.
If it doesn't exists, check for newer file and follow the newer code.

vishnuitta: cc: @nrb ^^

skriss: Ah, I see. so you're leaving the `Snapshot` struct as-is, and creating new type aliases to define the CRD.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,434 @@
...
+
+## Background
+
+(Omitted, see introduction)
+
+## High-Level Design
+
+### Progress of backup operation handled by volume snapshotter
+
+Progress will be updated by volume snapshotter in VolumePluginBackup CR which is specific to that backup operation.
```
skriss, line 31: I'm just going back and forth on the API design a bit. I think either option is workable for now for this specific use case, just trying to think through other uses/future extensions and whether that should push us in one direction or the other.

- Approach 1 combines the snapshot & backup info into a single CRD, and then has a parallel CRD on the restore side
- Approach 2 models the backup separately from the snapshot (the snapshot remains as an internal struct, not a CRD), but there's a lot of duplication of info between the snapshot & backup structs. On the restore side, there's one struct that models both the restore from durable storage, and the snapshot restore.
- we could consider fully breaking things out, so 4 CRDs: on the backup side, one for the snapshot, one for the backup to durable storage; on the restore side, one for the restore from durable storage, and one for the snapshot restore

I think I'm still leaning towards approach 1, as it solves the need and is simplest overall, but I'd like other maintainers' inputs on this (including any other options you see).

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,434 @@
...
+After return from `CreateSnapshot` in `takePVSnapshot`, currently Velero adds `volume.Snapshot` to `backupRequest`. Instead of this, CR will be added to `backupRequest`.
+During persistBackup call, this CR also will be backed up to backup location.
+
+In backupSyncController, it checks for any VolumePluginBackup CRs that need to be synced from backup location, and syncs them to cluster if needed.
+
+VolumePluginBackup will be useful as long as backed up data is available at backup location. When the Backup is deleted either by manually or due to expiry, VolumePluginBackup also can be deleted.
+
+`processRequest` of `backupDeletionController` will perform deletion of VolumePluginBackup before volumesnapshotter's DeleteSnapshot is called.
+
+#### Backward compatibility:
```
skriss, line 176: It seems like we could potentially get rid of the `volume.Snapshot` type and not need to keep it around with the type-aliasing for the CRD. The existing `volumesnapshots.json.gz` files should still be able to be deserialized into the new `VolumePluginBackup` CRs, and then we can check for the existence of `metadata.name` to determine if it's a CR or not (per https://github.com/vmware-tanzu/velero/pull/2543/files#diff-a724dc7ff53f45b40b4c5a5f57e741dbR184). 

Does that sound right? I think it'd be simpler if we could avoid the type aliasing and the two structs, and just have the new API type.

design/plugin-backup-and-restore-progress-design.md
skriss: Currently everything runs under the same `velero` service account so all plugins have broad access, which would include being able to modify CRs created by another plugin.

design/plugin-backup-and-restore-progress-design.md
vishnuitta: yes @skriss .. updated design accordingly

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,434 @@
...
+	// in the cloud provider.
+	VolumeAZ string `json:"volumeAZ,omitempty"`
+
+	// VolumeIOPS is the optional value of provisioned IOPS for the
+	// disk/volume in the cloud provider API.
+	VolumeIOPS *int64 `json:"volumeIOPS,omitempty"`
+}
+```
+
+Few fields (except first two) are added to volume.SnapshotStatusSpec. Below is the updated one:
```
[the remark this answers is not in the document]
vishnuitta: thanks.. updated the doc

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,434 @@
...
+
+	// BackupUID is the UID of the Velero backup this snapshot
+	// is associated with.
+	BackupUID string `json:"backupUID"`
+
+	// Location is the name of the VolumeSnapshotLocation where this snapshot is stored.
+	Location string `json:"location"`
+
+	// PersistentVolumeName is the Kubernetes name for the volume.
+	PersistentVolumeName string `json:persistentVolumeName`
```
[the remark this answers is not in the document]
vishnuitta: yes.. updated the doc

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,434 @@
...
+	// provider API of this volume.
+	ProviderSnapshotID string `json:"providerSnapshotID,omitempty"`
+
+	// Phase is the current state of the VolumeSnapshot.
+	Phase SnapshotPhase `json:"phase,omitempty"`
+
+	// PluginSpecific are a map of key-value pairs that plugin want to provide
+	// to user to identify plugin properties related to this backup
+	// +optional
+	PluginSpecific map[string]string `json:"pluginSpecific,omitempty"`
```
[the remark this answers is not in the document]
vishnuitta: as this is supposed to be properties specific to Backup/restore operation, thinking of changing to
`BackupProperties` and `RestoreProperties`..
does that make sense

design/plugin-backup-and-restore-progress-design.md
vishnuitta: removed the aliases and also volume.Snapshot

design/plugin-backup-and-restore-progress-design.md
nrb: I think I'm leaning towards approach 1, too, with the caveat that it needs to be optional. We can't require it if not all plugins are updating these CRs.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+
+### Approach 1
+
+Existing `Snapshot` Go struct from `volume` package have most of the details related to backup operation performed by volumesnapshotters.
+This struct also gets backed up to backup location. But, this struct doesn't get synced on other clusters at regular intervals.
+It will be synced only during restore operation, and velero CLI shows few of its contents.
+
+At a high level, in this approach, this struct will be converted to a CR by adding new fields (related to Progress tracking) to it, and gets rid of `volume.Snapshot` struct.
+
+Instead of backing up of Go struct, CRs will be backed up to backup location, and it gets synced into other cluster by backupSyncController running in that cluster.
```
nrb, line 47: Line 47 contradicts line 41. Is the plan to sync the CR on a regular basis, or at restore time?

design/plugin-backup-and-restore-progress-design.md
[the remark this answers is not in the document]
nrb: Would the properties be the same on backup and restore? Are they copied from the backup CR to the restore CR?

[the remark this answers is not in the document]
vishnuitta: Need not be @nrb .. Backup related and Restore related ones are different.
As mentioned in one of the comments, one Backup property can be its parent snapshot name, in case of OpenEBS as it supports incremental backups.

design/plugin-backup-and-restore-progress-design.md
vishnuitta: thats true @nrb .. With approach 1, a CR will be created for every backup/restore operation even though plugin doesn't update it.

design/plugin-backup-and-restore-progress-design.md
vishnuitta: First para i.e., lines 41, 42 and 43 are related to current behavior. I will update it.
At line 47 is the proposed solution. So, plan is to sync the CR on regular basis.

vishnuitta: updated the text

carlisia:
Will review tomorrow.

vishnuitta:
@carlisia or @nrb or @ashish-amarnath .. would you guys prefer a hangout to close the things on this thread?

stephbman:
@vishnuitta we could do a hangout or zoom meeting to review this - @carlisia since you are going to take a look at reviewing this would it be better for you to provide comments here or to have a short call w/ Vishnu on this?

design/plugin-backup-and-restore-progress-design.md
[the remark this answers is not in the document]
carlisia: `BackupProperties` or `BackupInfo` would work.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+
+Few fields (except first two) are added to volume.SnapshotStatus to convert it to CR's status. Below is the updated VolumePluginBackup CR's status:
+```
+type VolumePluginBackupStatus struct {
+	// ProviderSnapshotID is the ID of the snapshot taken in the cloud
+	// provider API of this volume.
+	ProviderSnapshotID string `json:"providerSnapshotID,omitempty"`
+
+	// Phase is the current state of the VolumeSnapshot.
+	Phase SnapshotPhase `json:"phase,omitempty"`
```
carlisia, line 97: What will the phases be for backup/restore?

carlisia:
@vishnuitta if you use the scaffold feature of kubebuilder to generate the new CRDs with the Velero project it won't work, because we structure things differently than kubebuilder expects (no `main.go` for example). The way we do it is to have a dummy project and copy/paste things. Ping me on #velero if I can help you with this.

carlisia:
> I'm in favor of approach 1, and if others are as well, then I'd simply ask approach 2 be marked as an alternative considered

@vishnuitta please move approach 2 to the "alternatives" section.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+First and last should be fine.
+
+For second case, decode in `GetBackupVolumeSnapshots` on the restore side should fill only required fields of older version and should work.
+
+For third case, after decode, metadata.name will be empty. `GetBackupVolumeSnapshots` decodes older json.gz into the CR which goes fine.
+It will be modified to return []VolumePluginBackupSpec, and the changes are done accordingly in its caller.
+
+If decode fails in second case during implementation, this CR need to be backed up to different file. And, for backward compatibility, newer code should check for old file existence, and follow older code if exists. If it doesn't exists, check for newer file and follow the newer code.
+
+`backupSyncController` on restore clusters gets the `<backupname>-volumesnapshots.json.gz` object from backup location and decodes it to in-memory VolumePluginBackup CR. If its `metadata.name` is populated, controller creates CR. Otherwise, it will not create the CR on the cluster. It can be even considered to create CR on the cluster.
```
ashish-amarnath, line 197: FYI: the backupsynccontroller doesn't directly contribute to restore.
But I get the idea.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+
+	// +optional
+	Spec VolumePluginBackupSpec `json:"spec,omitempty"`
+
+	// +optional
+	Status VolumePluginBackupStatus `json:"status,omitempty"`
+}
+```
+
+For every backup operation of volume, volume snapshotter creates VolumePluginBackup CR in Velero namespace.
```
ashish-amarnath, line 370: suggest moving creation of the customresource into the plugin.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+- `velero.io/backup-name` with value as Backup Name, and,
+- `velero.io/snapshot-id` with value as snapshot id that need to be restored
+- `velero.io/provider` with value as `Provider` in `VolumeSnapshotLocation`
+
+Labels will be set with the value returned from `GetValidName` function. (https://github.com/vmware-tanzu/velero/blob/master/pkg/label/label.go#L35).
+
+Plugin will be able to identify CR by using snapshotID that it received as parameter of CreateVolumeFromSnapshot API, and plugin's Provider name.
+It updates the progress of restore operation regularly if plugin supports feature of showing progress.
+
+Velero deletes VolumePluginRestore CR when it handles deletion of Restore CR.
```
ashish-amarnath, line 289: Just calling this out. This requires change to the backupdeletioncontroller.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+
+## Detailed Design
+
+### Approach 1
+
+Existing `Snapshot` Go struct from `volume` package have most of the details related to backup operation performed by volumesnapshotters.
+This struct also gets backed up to backup location. But, this struct doesn't get synced on other clusters at regular intervals.
+It is currently synced only during restore operation, and velero CLI shows few of its contents.
+
+At a high level, in this approach, this struct will be converted to a CR by adding new fields (related to Progress tracking) to it, and gets rid of `volume.Snapshot` struct.
```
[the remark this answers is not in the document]
vishnuitta: PR have CR struct below (not exact CRD) for better readability.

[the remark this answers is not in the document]
vishnuitta: Look for heading `VolumePluginBackup CR`

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+
+Instead of backing up of Go struct, proposal is: to backup CRs to backup location, and sync them into other cluster by backupSyncController running in that cluster.
```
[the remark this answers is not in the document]
vishnuitta: Below description is added to PR. Are you looking for more details?
```
After return from CreateSnapshot in takePVSnapshot, currently Velero adds volume.Snapshot to backupRequest. Instead of this, CR will be added to backupRequest. During persistBackup call, this CR also will be backed up to backup location.
```

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+
+#### VolumePluginBackup CR
+
+There is one addition to volume.SnapshotSpec, i.e., ProviderName to convert it to CR's spec. Below is the updated VolumePluginBackup CR's Spec:
```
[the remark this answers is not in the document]
vishnuitta: `volume.SnapshotSpec` is existing Go type which gets backed up. In this approach, that will be replaced with VolumePluginBackup CR.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+
+	// +optional
+	Spec VolumePluginBackupSpec `json:"spec,omitempty"`
+
+	// +optional
+	Status VolumePluginBackupStatus `json:"status,omitempty"`
+}
+```
+
+For every backup operation of volume, Velero creates VolumePluginBackup CR before calling volumesnapshotter's CreateSnapshot API.
```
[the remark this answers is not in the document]
vishnuitta: this is what second approach proposes. If creation of CR is left to plugin, `volume.Snapshot` struct need not be removed.
Also, few fields can be added to the VolumePluginBackupSpec if we need to take this approach.
Difference in changes between the approaches are also highlighted in the PR. Please look for heading
```
'core' Velero client/server required changes
```

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+
+### Progress of restore operation handled by volume snapshotter
+
+Progress will be updated by volume snapshotter in VolumePluginRestore CR which is specific to that restore operation.
+
+## Detailed Design
+
+### Approach 1
+
+Existing `Snapshot` Go struct from `volume` package have most of the details related to backup operation performed by volumesnapshotters.
```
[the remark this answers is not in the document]
vishnuitta: will update with code links.

design/plugin-backup-and-restore-progress-design.md
```diff
@@ -0,0 +1,448 @@
...
+For every backup operation of volume, Velero creates VolumePluginBackup CR before calling volumesnapshotter's CreateSnapshot API.
+
+In order to know the CR created for the particular backup of a volume, Velero adds following labels to CR:
+- `velero.io/backup-name` with value as Backup Name, and,
+- `velero.io/pv-name` with value as volume that is undergoing backup
+
+Backup name being unique won't cause issues like duplicates in identifying the CR.
+Labels will be set with the value returned from `GetValidName` function. (https://github.com/vmware-tanzu/velero/blob/master/pkg/label/label.go#L35).
+
+If Plugin supports showing progress of the operation it is performing, it does following:
```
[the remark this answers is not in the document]
vishnuitta: responded in above comment

design/plugin-backup-and-restore-progress-design.md
vishnuitta: responded in above comment

design/plugin-backup-and-restore-progress-design.md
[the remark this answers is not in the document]
ashish-amarnath: Thanks for clarifying. I re-read the comments and the doc and I am OK with moving this forward.

design/plugin-backup-and-restore-progress-design.md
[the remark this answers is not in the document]
ashish-amarnath: this doesn't have to block the merge of this PR.

vishnuitta:
thanks @carlisia for the merge and thanks to every one for the reviews.
Apologizes for delay in updating the last set of review comments. I will get them incorporated in the implementation related PRs.

PR #2550, opened 2020-05-20
re-instantiate backup store just before persisting artifacts

### LINKED ISSUE #2414: Retry Backup with Updated Credentials
We use temporary credentials with Velero which are updated every 30 mins. We update the `cloud-credentials` secret with the new credentials every 30 mins. If a backup starts before the credentials are updated and completes after the credentials are updated, the backup will fail as it uses the older credentials even when the newer credentials are available in the secret. 

Velero should retry Backups with the latest credentials available if it fails to upload using the current credentials.

**Environment:**

- Velero version (use `velero version`): v1.3.0
- Velero features (use `velero client config get features`): <NOT SET>
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-18T14:36:53Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```
- Kubernetes installer & version: Kind
- OS (e.g. from `/etc/os-release`): All

skriss:

closes #2414 

Haven't tested this at all yet, but I believe this should sufficiently remediate the issue described in 2414. cc @ashish-amarnath @rohandvora

### DIFF
--- a/changelogs/unreleased/2550-skriss
+++ b/changelogs/unreleased/2550-skriss
@@ -1,0 +1,1 @@
+backup/restore: reinstantiate backup store just before uploading artifacts to ensure credentials are up-to-date
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -533,7 +533,7 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
 		return err
 	}
 
-	backupLog.Info("Setting up backup store")
+	backupLog.Info("Setting up backup store to check for backup existence")
 	backupStore, err := c.newBackupStore(backup.StorageLocation, pluginManager, backupLog)
 	if err != nil {
 		return err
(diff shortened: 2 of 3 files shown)


skriss:
I was able to test this and can confirm that it produces the desired behavior. If the credentials are updated mid-backup, and the kubelet refreshes the secret's volume mount in the pod, then when the backup completes and is uploaded to object storage, the new credentials are used.

Issue #92, opened 2017-09-14
Config change detection gets a false positive when the watch expires


ncdc:
For various reasons, when you `Get()` an object, its `kind` and `apiVersion` are blank. When you receive the same object via a `Watch()`, those fields are present. Any time the watch expires (timeout, too old resource version), the config informer relists/rewatches and the update handler in `watchConfig` receives the current value of the config. Because of the mismatch described above, the code thinks there is a config change and shuts down.

ncdc:
I've fixed this locally. The next false-positive is about `resourcePriorities`. If the config doesn't define them, we default them in the code. Then we get the config from the watch, without any defaulting applied.

PR #2561, opened 2020-05-22
Convert manifests + BSL api client to kubebuilder

carlisia:
- All CRD manifests, and generation of manifests generated by kubebuilder (same but different)
- BSL api client code generation ported from go-client to kubebuilder
- Substituted all BSL usage to the kubebuilder api client
- Tied in CRD/code generation with existing scripts

Closes: https://github.com/vmware-tanzu/velero/issues/2604

TODO:
- [x] basic CRUD tests for backup and backup storage location
- [x] more manual tests, esp. with restic

https://www.loom.com/share/b96ce9e9cd3846428ef4243ed9a1828a


carlisia:
Rebasing this PR is not fun @vmware-tanzu/velero-maintainers!

api/v1/backupstoragelocation_types.go
```diff
@@ -21,33 +21,85 @@ import (
...
+	// will be removed entirely as of v2.0.
+	// +optional
+	AccessMode BackupStorageLocationAccessMode `json:"accessMode,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:object:generate=true
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase",description="Backup Storage Location status such as Available/Unavailable"
```
carlisia, line 81: This kubectl column would be empty now but filled in once this goes in: https://github.com/vmware-tanzu/velero/pull/2493

api/v1/backupstoragelocation_types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 the Velero contributors.
+Copyright 2018, 2020 the Velero contributors.
```
skriss, line 2: Thinking through the implications of our API types living in two separate packages, and also if we need to consider this a breaking change. It will be breaking for anyone who vendors our code to use the API/client. That said, we haven't historically actually maintained backwards compatibility in our exported API surface..

hack/update-generated-crd-code.sh
```diff
@@ -21,32 +21,4 @@ set -o xtrace
...
-  --output-base ../../.. \
-  $@
-
-controller-gen \
-  crd:crdVersions=v1beta1,preserveUnknownFields=false \
-  output:dir=./pkg/generated/crds/manifests \
-  paths=./pkg/apis/velero/v1/...
-
-go generate ./pkg/generated/crds
+go generate ./config/crd/crds
```
skriss, line 24: I think we need to keep most of the code that was removed here, for the non-kubebuilder API types/clients/etc

main.go
```diff
@@ -0,0 +1,18 @@
+/*
```
skriss, line 1: can this be removed?

pkg/backup/request.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"sort"
 
+	veleroapiv1 "github.com/vmware-tanzu/velero/api/v1"
```
skriss, line 23: if we do go ahead with two separate packages for the API types, we need to come up with a clearer way to alias them :)

pkg/controller/suite_test.go
```diff
@@ -0,0 +1,36 @@
...
+	veleroapiv1 "github.com/vmware-tanzu/velero/api/v1"
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+
+	. "github.com/onsi/gomega"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	// +kubebuilder:scaffold:imports
+)
+
+func newFakeClient(g *WithT, initObjs ...runtime.Object) client.Client {
```
skriss, line 32: IDK how much of an issue this would cause, but I'd prefer not to introduce the gomega dependency if we don't need to - we're already using stretchr/testify for the `assert` and `require` libs everywhere else.

carlisia: Not needed, you're right. I'll remove it! If we go in this direction, we can consider if we want to continue using the same testing packages, and have the controllers look like they do today, or use the BDD style and have them look like this: https://github.com/kubernetes-sigs/cluster-api/blob/master/controllers/cluster_controller_phases_test.go.

pkg/cmd/cli/restic/server.go
```diff
@@ -130,6 +152,21 @@ func newResticServer(logger logrus.FieldLogger, factory client.Factory) (*restic
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
 
+	clientConfig, err := factory.ClientConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
```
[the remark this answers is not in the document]
carlisia: Oops, I should have removed that and used ours. Fixing.

pkg/cmd/cli/backup/create.go
```diff
@@ -33,10 +38,23 @@ import (
...
+var (
+	scheme = runtime.NewScheme()
+)
+
+func init() {
+	_ = clientgoscheme.AddToScheme(scheme)
+
+	_ = velerov1api.AddToScheme(scheme)
+	// +kubebuilder:scaffold:scheme
+}
```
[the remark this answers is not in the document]
carlisia: Maybe @ashish-amarnath can assist me here. When I only had it in the api package, it didn't work. The cli code didn't recognize any scheme.

pkg/backup/request.go
carlisia: Yes, I'm very open to have a better alias name. For lack of a great idea, I turned `velero/api/v1` into `veleroapiv1`, which (obvs) follows the order of the directories (which doesn't mean is great but has a meaning), and also was NOT the `velerov1api` which we use for our existing api package.

main.go
carlisia: The file? Yes and no :)

Yes, we can remove it and everything will work fine.

Things will not work fine IF/WHEN we try to create a new scaffold for either a resource or a controller using `kubebuilder`. `kubebuilder` will halt the generation of files when it doesn't find `main.go`. Well, we don't do this everyday (generate new resources/controllers) so (again, if we go in this direction) we COULD just remember to add a `main.go` file while doing that and remove it again.

hack/update-generated-crd-code.sh
carlisia: Hm. 🤔 I think this will work. Here's how I tested it: added a field to `pkg/apis/velero/v1/backup.go` and ran `make update`, and the other 3 files below got updated as expected. If there's nothing missing, then I think it's good like this?

image

carlisia: Although all api types (besides the BSL) remain where they were, their corresponding manifest files and the raw CRDs are being generated in the same way and placed in the same places.

skriss: There's also the generated clients/listers/informers in `pkg/generated`, and the deep-copy functions that go directly in `pkg/apis/velero/v1/zz_generated.deepcopy.go`, that potentially need to update when one of the existing API types updates. I think specifically [this code](https://github.com/vmware-tanzu/velero/pull/2561/files/bf5052d695c4617f97bd62df2d52bfb88d4163ab#diff-892d90adfc5482f1fee4517a4ad27aeeL38-L45) needs to be retained to properly update those.

skriss: And I guess [this code](https://github.com/vmware-tanzu/velero/pull/2561/files/bf5052d695c4617f97bd62df2d52bfb88d4163ab#diff-892d90adfc5482f1fee4517a4ad27aeeL24-L31) as well.

pkg/cmd/cli/backup/create.go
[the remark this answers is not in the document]
skriss: OK, I can try to dig into it some more as well.

hack/update-generated-crd-code.sh
carlisia: Hmyeah... I totally lost track of this. Will fix.

Makefile
```diff
@@ -40,6 +40,14 @@ CLI_PLATFORMS ?= linux-amd64 linux-arm linux-arm64 darwin-amd64 windows-amd64 li
 CONTAINER_PLATFORMS ?= linux-amd64 linux-ppc64le linux-arm linux-arm64
 MANIFEST_PLATFORMS ?= amd64 ppc64le arm arm64
 
+#kubebuilder automatically runs "make" when a resource or controller is created
+all: generate-all
```
ashish-amarnath, line 44: This is just creating an alias for `generate-all`.
You can undo this change and keep `generate-all` as the first target, if that is you wanted.
You can also undo the change on line #92

api/v1/backupstoragelocation_types.go
```diff
@@ -21,33 +21,85 @@ import (
...
+
+	// AccessMode is an unused field.
+	//
+	// Deprecated: there is now an AccessMode field on the Spec and this field
+	// will be removed entirely as of v2.0.
+	// +optional
+	AccessMode BackupStorageLocationAccessMode `json:"accessMode,omitempty"`
+}
+
+// +kubebuilder:object:root=true
```
ashish-amarnath, line 77: should we also add the annotation to generate RBAC?

api/v1/backupstoragelocation_types.go
```diff
@@ -117,31 +148,3 @@ const (
 
 // TODO(2.0): remove the AccessMode field from BackupStorageLocationStatus.
```
ashish-amarnath, line 149: should lines 149 and 150 also be removed?

config/crd/bases/velero.io_backups.yaml
```diff
@@ -14,7 +14,6 @@ spec:
     listKind: BackupList
     plural: backups
     singular: backup
-  preserveUnknownFields: false
```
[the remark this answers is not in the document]
ashish-amarnath: in all CRDs, throughout.

config/certmanager/certificate.yaml
```diff
@@ -0,0 +1,26 @@
+# The following manifests contain a self-signed issuer CR and a certificate CR.
```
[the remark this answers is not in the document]
ashish-amarnath: this is most likely coming in as part of the validation webhook that comes with kubebuilder

config/samples/velero_v1_backupstoragelocation.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: velero.io/v1
+kind: BackupStorageLocation
+metadata:
+  name: backupstoragelocation-sample
+spec:
+  # Add fields here
+  foo: bar
```
ashish-amarnath, line 7: this example needs to be completed

hack/update-generated-crd-code.sh
ashish-amarnath: to confirm this works fully, I would suggest deleting all the generated code and then run the code generation step again to confirm everything comes backup.

pkg/cmd/cli/backup/create.go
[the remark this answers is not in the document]
ashish-amarnath: we can look at examples to figure out how to consolidate this in one place.

pkg/cmd/cli/restic/server.go
```diff
@@ -157,6 +195,12 @@ func (s *resticServer) run() {
 
 	var wg sync.WaitGroup
 
+	// TODO(carlisia): how to handle the fetching of the bsl informer:
+	// - options are: 1) get informer for specific obj (below, but w/o namespace or obj name info because we don't know it here) or 2) for a specific kind, which would be CRD?
```
ashish-amarnath, line 199: option 2 feels more correct.
Were you not able to do that?

config/crd/bases/velero.io_backups.yaml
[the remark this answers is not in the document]
nrb: I _think_ it would be `false` if not specified, since the empty value for bools in Go is `false`.

config/default/kustomization.yaml
```diff
@@ -0,0 +1,70 @@
+# Adds namespace to all resources.
+namespace: velero-system
```
nrb, line 2: I think this namespace would need to change, wouldn't it?

config/prometheus/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
```
nrb, line 1: Do we want prometheus coming in with this change? This question applies to all the files in the `config/prometheus` directory.

pkg/cmd/server/server.go
```diff
@@ -100,6 +110,10 @@ const (
 	defaultBackupTTL = 30 * 24 * time.Hour
 )
 
+var (
+	scheme = runtime.NewScheme()
```
nrb, line 114: Can you add some comments to explain why this and the init function are doing scheme calls? It's not clear to me why this is happening in the server.go file.

carlisia:
Deleting the entire `generated` directory and running `make update`
works. Modifying an api and running `make verify` works as expected.

Pending: addressing questions and maybe a couple clean ups as per the reviews so far, but no big change expected.

api/v1/backupstoragelocation_types.go
carlisia: @ashish-amarnath I really don't know. The original BSL didn't have that marker. Please let me know if you think I should add it, but also if you could help me understand why it'd be great.

api/v1/backupstoragelocation_types.go
carlisia: No, they have nothing to do with this PR, those are slated to be done for v2.0.

carlisia:
@vmware-tanzu/velero-maintainers PTAL when you have a chance, it's ready for review.

There are 3 //todo(carlisia) to call attention to some questions I have about how to best refactor those code sections.

All other reviews were addressed, minus one question for Ashish (see above).

skriss:
sorry i've been slow here - planning to look at this tomorrow morning!

carlisia:
WIP BSL controller: https://github.com/vmware-tanzu/velero/pull/2617

Makefile
```diff
@@ -296,3 +296,35 @@ serve-docs:
...
+	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$CONTROLLER_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5 ;\
+	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
+	}
+CONTROLLER_GEN=$(GOBIN)/controller-gen
+else
+CONTROLLER_GEN=$(shell which controller-gen)
+endif
```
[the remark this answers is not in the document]
carlisia: Ohh. No, no grand plan not to use the image, and I think using the image is probably best. I'll fix this.

pkg/apis/velero/v1/register.go
```diff
@@ -22,20 +22,6 @@ import (
...
-var (
-	// SchemeBuilder collects the scheme builder functions for the Velero API
-	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
-
-	// AddToScheme applies the SchemeBuilder functions to a specified scheme
-	AddToScheme = SchemeBuilder.AddToScheme
-)
-
-// GroupName is the group name for the Velero API
-const GroupName = "velero.io"
```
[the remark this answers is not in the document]
carlisia: It was being used, but I changed to use it directly and also that code is now in `groupgversion_info.go`:
```
SchemeGroupVersion = schema.GroupVersion{Group: "velero.io", Version: "v1"}
```

carlisia:
> BTW I'm happy to spend some time testing this if that'd be helpful; let me know if there are any specific areas you'd like me to focus on!

Oh if you could run some manual tests with Restic it would be awesome!

pkg/restic/repository_manager.go
```diff
@@ -79,20 +82,19 @@ type RestorerFactory interface {
...
-	fileSystem                   filesystem.Interface
-	ctx                          context.Context
-	pvcClient                    corev1client.PersistentVolumeClaimsGetter
-	pvClient                     corev1client.PersistentVolumesGetter
+	namespace          string
+	veleroClient       clientset.Interface
+	secretsLister      corev1listers.SecretLister
+	repoLister         velerov1listers.ResticRepositoryLister
+	repoInformerSynced cache.InformerSynced
+	kbCache            kbcache.Cache
```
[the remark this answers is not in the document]
carlisia: That is correct. I think I had added this before I understood that get/list come from cache when using the client, then forgot to change it. Changing it now.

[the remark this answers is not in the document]
carlisia: Oh, actually, no. I used it so I would have access to `WaitForCacheSync`, which the client doesn't provide. I think we need this check? Please let me know.

https://github.com/vmware-tanzu/velero/blob/84698c88ff22229434fa3da6d8cbd99b9a4c7def/pkg/restic/repository_manager.go#L273

skriss:
sure, I'll try to spend some time on that this afternoon!

pkg/restic/repository_manager.go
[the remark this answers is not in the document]
skriss: so this is very related to https://github.com/vmware-tanzu/velero/pull/2561#discussion_r438242022, and I'm not 100% confident I understand the controller-runtime client behavior here. My sense was that it combines a ~lister/cache and an actual API client behind the scenes, so it'll return an object from the cache if it can, else it'll hit the API.

If that's the case, then we *may* not need to wait for the cache sync here, since we can use the client regardless. But again, i'm not totally sure I've got the mechanics of the new client down. Perhaps @ashish-amarnath can weigh in here.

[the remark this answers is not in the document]
ncdc: If you are using something that looks like this to create a client:

https://github.com/kubernetes-sigs/cluster-api/blob/714b0a97484465a05a67d8ccd1b64aa388b9fac2/util/util.go#L636-L646

Then any calls to `client.Get()` will not return any data until after the cache is synced.

Additionally, it will only return data from the cache. It does not fall back to doing a live read against the apiserver.

Does this help? Happy to hop on a call if you need it.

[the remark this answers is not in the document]
shomron: I don't have all the context around this change, but I'll try to weigh in :) Internally `controller-runtime`'s cache will `WaitForCacheSync` when it tries to access any of the informers it is managing, so I'm not sure why you'd have to call it directly.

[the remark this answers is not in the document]
carlisia: I think that clarifies the question, @ncdc, thank you!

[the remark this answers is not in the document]
shomron: > My sense was that it combines a ~lister/cache and an actual API client behind the scenes, so it'll return an object from the cache if it can, else it'll hit the API.

@skriss that's not quite my understanding.
If you use the default _Client_ setup by the controller-runtime Manger, it will delegate to the informer cache for typed resource reads and pass-through to the API Server for unstructured resource reads. For writes it will always pass-through to the API Server.

That is wired up [here](https://github.com/kubernetes-sigs/controller-runtime/blob/32b4434331e72fed344f71d278982827e08db2a8/pkg/manager/manager.go#L347).

If you use the _Cache_, you should be hitting all informers all the time.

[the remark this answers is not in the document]
ncdc: Note, in Cluster API, we explicitly made the choice to use a cache for all resources (typed and unstructured).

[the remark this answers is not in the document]
skriss: thanks @shomron and @ncdc, that's helpful clarification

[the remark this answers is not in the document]
carlisia: Yes, what @shomron said is what I learned too. In other words, if the cache isn't started for example, the get/list will fail. Ways around that are, specifically pass an unstructured obj, or call the `GetAPIReader()`.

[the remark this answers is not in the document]
ncdc: If you use the controller-runtime `Manager`, it auto-starts all the caches for all the kinds you care about (have reconcilers for) automatically, so you generally don't need to worry about caches not being started in that setup.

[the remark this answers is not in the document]
carlisia: Hm. That's a good point. At this point there is no reconciler and no manager, we only have the resource.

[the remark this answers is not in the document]
carlisia: No actually, that's wrong, we do have a manager, just not any reconciler.

pkg/cmd/cli/restic/server.go
```diff
@@ -130,6 +143,21 @@ func newResticServer(logger logrus.FieldLogger, factory client.Factory) (*restic
...
+
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+	mgr, err := ctrl.NewManager(clientConfig, ctrl.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// +kubebuilder:scaffold:builder
```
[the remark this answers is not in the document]
nrb: Honestly having a hard time finding it in https://github.com/kubernetes-sigs/kubebuilder/search?q=%2Bkubebuilder%3Ascaffold%3Abuilder&type=Code, too. I'm only seeing it in example code without much explanation of what it's actually doing.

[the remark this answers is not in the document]
ncdc: It's a placeholder so kubebuilder knows where to do code insertions.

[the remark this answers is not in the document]
nrb: @ncdc Thanks. Does it map to [this method](https://github.com/kubernetes-sigs/kubebuilder/blob/6e0d3d4f18ecdc335177d28e1d6cd4d7b2930242/pkg/plugin/internal/machinery/scaffold.go#L66)?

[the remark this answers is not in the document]
carlisia: Oh. So I guess we can deleted bc kubebuilder always writes to `main.go` and would never write to this file.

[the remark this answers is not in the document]
ncdc: @nrb I don't know (I'm not familiar with how it's implemented in kubebuilder).

How did this line come to exist in server.go?

carlisia:
@vmware-tanzu/velero-maintainers addressed all code reviews, PTAL!

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -42,19 +42,23 @@ import (
...
-	pvLister              corev1listers.PersistentVolumeLister
-	backupLocationLister  listers.BackupStorageLocationLister
-	nodeName              string
+	podVolumeBackupClient  velerov1client.PodVolumeBackupsGetter
+	podVolumeBackupLister  listers.PodVolumeBackupLister
+	secretLister           corev1listers.SecretLister
+	podLister              corev1listers.PodLister
+	pvcLister              corev1listers.PersistentVolumeClaimLister
+	pvLister               corev1listers.PersistentVolumeLister
+	backupLocationInformer k8scache.Informer
```
skriss, line 59: I think we can drop the `backupLoationInformer` and associated import now

pkg/cmd/cli/restic/server.go
```diff
@@ -43,6 +46,13 @@ import (
 	"github.com/vmware-tanzu/velero/pkg/restic"
 	"github.com/vmware-tanzu/velero/pkg/util/filesystem"
 	"github.com/vmware-tanzu/velero/pkg/util/logging"
+
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+var (
+	scheme = runtime.NewScheme()
```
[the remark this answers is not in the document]
carlisia: Good catch. Adding it right before creating the manager.

skriss:
I'm seeing a couple of errors in the velero server log (haven't gotten to restic yet) on startup:

```
time="2020-06-11T21:20:28Z" level=error msg="Error getting backup storage locations from lister" controller=backup-sync error="the cache is not started, can not read objects" error.file="/github.com/vmware-tanzu/velero/pkg/controller/backup_sync_controller.go:129" error.function="github.com/vmware-tanzu/velero/pkg/controller.(*backupSyncController).run" logSource="pkg/controller/backup_sync_controller.go:129"
...
time="2020-06-11T21:21:28Z" level=error msg="Error patching backup location's last-synced time" backupLocation=default controller=backup-sync error="backupstoragelocations.velero.io \"default\" not found" error.file="/github.com/vmware-tanzu/velero/pkg/controller/backup_sync_controller.go:311" error.function="github.com/vmware-tanzu/velero/pkg/controller.(*backupSyncController).run" logSource="pkg/controller/backup_sync_controller.go:311"
```

The BSL definitely exists.

ncdc:
If/when you convert to use the controller-runtime Manager for everything, it auto-starts all the caches. Otherwise, you'll need to start them manually before trying to use the client.

carlisia:
Ohh. Right. I'll add a start for the manager.

carlisia:
I'm getting this and the other PR that adds the BSL controller confused.

I had already added the start. @ncdc do you mean something different from this: https://github.com/vmware-tanzu/velero/blob/ed6fb4568db0bbd771196f28344dc2cb1deda8a2/pkg/cmd/server/server.go#L911

ncdc:
@carlisia `s.mgr.Start()` is what I meant. I'm guessing what you're seeing is a race between your controllers (pre-existing / outside of the Manager's control) and the Manager starting the caches. If you complete the massive effort to convert 100% of Velero to be controller-runtime based, this issue will go away. But while you have a mix of controller-runtime Reconcilers and "legacy" Velero controllers, you'll have to wait for caches to sync before trying to use the client. Does that make sense?

hack/update-all.sh
```diff
@@ -24,3 +24,5 @@ for f in ${HACK_DIR}/update-*.sh; do
   fi
   $f
 done
+
```
ashish-amarnath, line 27: changes to this file may be reverted

pkg/cmd/cli/restic/server.go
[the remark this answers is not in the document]
ashish-amarnath: I don't know what code this marker generates. But I think we don't need any kubebuilder code generated here and that this may be removed.

carlisia:
@ncdc yes, very clear. I'm glad you jumped in this PR. I did suspect that we would have to wait, but didn't know this will go away once all controllers are converted (eventually). I guess we'll have to add a wait there...

pkg/cmd/server/server.go
```diff
@@ -439,8 +463,10 @@ func (s *server) validateBackupStorageLocations() error {
 	pluginManager := clientmgmt.NewManager(s.logger, s.logLevel, s.pluginRegistry)
 	defer pluginManager.CleanupClients()
 
-	locations, err := s.veleroClient.VeleroV1().BackupStorageLocations(s.namespace).List(metav1.ListOptions{})
-	if err != nil {
+	locations := &velerov1api.BackupStorageLocationList{}
+	if err := s.mgr.GetAPIReader().List(context.Background(), locations, &kbclient.ListOptions{
```
ashish-amarnath, line 467: Is there a reason why we are not using `GetClient`? The `GetAPIReader` talks directly to the API server and does not use the cache.

pkg/cmd/server/server.go
```diff
@@ -880,6 +904,15 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
 
 	s.logger.Info("Server started successfully")
 
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		// +kubebuilder:scaffold:builder
```
ashish-amarnath, line 910: this kubebuilder marker can also be removed.

pkg/controller/backup_sync_controller.go
```diff
@@ -101,35 +100,38 @@ func NewBackupSyncController(
 
 // orderedBackupLocations returns a new slice with the default backup location first (if it exists),
 // followed by the rest of the locations in no particular order.
-func orderedBackupLocations(locations []*velerov1api.BackupStorageLocation, defaultLocationName string) []*velerov1api.BackupStorageLocation {
-	var result []*velerov1api.BackupStorageLocation
+func orderedBackupLocations(locationList *velerov1api.BackupStorageLocationList, defaultLocationName string) []velerov1api.BackupStorageLocation {
```
ashish-amarnath, line 103: Changes to this function may not be necessary.
We can call this function like
```
locations := orderedBackupLocations(locationList.Items, c.defaultBackupLocation)
```

pkg/cmd/server/server.go
carlisia: Yes, at this point the cache hasn't been started yet.

pkg/restic/common.go
```diff
@@ -229,17 +233,20 @@ func TempCACertFile(caCert []byte, bsl string, fs filesystem.Interface) (string,
...
 	}
 
-	if location.Spec.ObjectStorage != nil {
-		return location.Spec.ObjectStorage.CACert, nil
+	if location.Spec.ObjectStorage == nil {
+		return nil, nil
 	}
 
-	return nil, nil
+	return location, nil
```
ashish-amarnath, line 249: Considering that this function is called `GetCACert`, it should return the CACert associated with the BSL and not the BSL object itself.

ashish-amarnath: If we need a function to get the BSL, suggest creating a new one to do that.

pkg/restic/common.go
```diff
@@ -254,12 +261,7 @@ func NewPodVolumeRestoreListOptions(name string) metav1.ListOptions {
 // should be used when running a restic command for an Azure backend. This list is
 // the current environment, plus the Azure-specific variables restic needs, namely
 // a storage account name and key.
-func AzureCmdEnv(backupLocationLister velerov1listers.BackupStorageLocationLister, namespace, backupLocation string) ([]string, error) {
-	loc, err := backupLocationLister.BackupStorageLocations(namespace).Get(backupLocation)
-	if err != nil {
-		return nil, errors.Wrap(err, "error getting backup storage location")
-	}
-
+func AzureCmdEnv(loc *velerov1api.BackupStorageLocation) ([]string, error) {
```
ashish-amarnath, line 264: Changes in this file are all API breaking changes.
For purposes of backwards compatibility, we should create new functions and add a deprecation notice to the old ones.

site/docs/master/code-standards.md
```diff
@@ -43,7 +43,8 @@ Example:
     	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
     	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
     	corev1listers "k8s.io/client-go/listers/core/v1"
-        
+       
+        veleroapiv1 "github.com/vmware-tanzu/velero/api/v1" // for the new kubebuilder apis 
         velerov1api ""github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
ashish-amarnath, line 48: `velerov1api` is a reused import alias.
Also looks like line#48 has a compiler error: unmatched `"`.

pkg/controller/backup_sync_controller.go
carlisia: `locationList.Items` is a `[]velerov1api.BackupStorageLocation`, and `orderedBackupLocations` needs a `[]&velerov1api.BackupStorageLocation` because of things like this: `result = append(result, locations[:i]...)`.

pkg/restic/common.go
carlisia: Yes, very good point!

site/docs/master/code-standards.md
carlisia: `github.com/vmware-tanzu/velero/api/v1` no longer exists so this change is being reverted.

carlisia:
Addressed all code reviews.

No more caching problems: Fixed it by registering all existing (non runtime-controller) controllers to the manager: https://github.com/vmware-tanzu/velero/blob/cd048ee78d5c0825c08dc51314999051a5b37b69/pkg/cmd/server/server.go#L897

pkg/cmd/cli/backup/create.go
```diff
@@ -142,13 +145,22 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 		return err
 	}
 
+	client, err := f.KubebuilderClient()
```
ashish-amarnath, line 148: should the
```
client, err := f.Client()
```
in other functions also be updated to use the `KubebuilderClient()`?

do we still want to continue having the field
```
client veleroclient.Interface
```
in `CreateOptions` Line#99)

pkg/client/factory.go
```diff
@@ -137,6 +143,21 @@ func (f *factory) DynamicClient() (dynamic.Interface, error) {
 	return dynamicClient, nil
 }
 
+func (f *factory) KubebuilderClient() (kbclient.Client, error) {
```
ashish-amarnath, line 146: `VeleroV1APIClient` seems like a better name for this function. WDYT?

pkg/cmd/cli/restic/server.go
```diff
@@ -182,26 +203,27 @@ func (s *resticServer) run() {
...
+	s.mgr.Add(managercontroller.Runnable(restoreController, 1))
 
-	<-s.ctx.Done()
+	s.logger.Info("Controllers starting...")
+
+	if err := s.mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+		s.logger.Fatal("Problem starting manager", err)
+	}
 
 	s.logger.Info("Waiting for all controllers to shut down gracefully")
```
ashish-amarnath, line 226: `s.mgr.Start` is a blocking call, so this message wouldn't be logged until `Start` returns.
I would suggest removing this log message.

pkg/cmd/server/server.go
ashish-amarnath: suggest adding a comment to that effect.

pkg/cmd/server/server.go
```diff
@@ -867,23 +892,20 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
 
-	s.logger.Info("Server started successfully")
+	s.logger.Info("Server starting...")
 
-	<-ctx.Done()
+	if err := s.mgr.Start(s.ctx.Done()); err != nil {
+		s.logger.Fatal("Problem starting manager", err)
+	}
 
 	s.logger.Info("Waiting for all controllers to shut down gracefully")
```
ashish-amarnath, line 908: same comment as https://github.com/vmware-tanzu/velero/pull/2561/files?file-filters%5B%5D=.go&file-filters%5B%5D=.md&file-filters%5B%5D=.mod&file-filters%5B%5D=.yaml&file-filters%5B%5D=No+extension#r442545437

pkg/cmd/server/server.go
```diff
@@ -867,23 +892,20 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
+		// Adding the controllers to the manager will register them as a (runtime-controller) runnable,
+		// so the manager will ensure the cache is started and ready before all controller are started
+		s.mgr.Add(managercontroller.Runnable(controllerRunInfo.controller, controllerRunInfo.numWorkers))
 	}
 
-	s.logger.Info("Server started successfully")
+	s.logger.Info("Server starting...")
 
-	<-ctx.Done()
+	if err := s.mgr.Start(s.ctx.Done()); err != nil {
```
ashish-amarnath, line 904: `err` should be returned

pkg/cmd/util/output/backup_storage_location_printer.go
```diff
@@ -30,7 +30,7 @@ var (
 		{Name: "Name", Type: "string", Format: "name"},
 		{Name: "Provider"},
 		{Name: "Bucket/Prefix"},
-		{Name: "Status"},
+		{Name: "Phase"},
```
ashish-amarnath, line 33: I think `Status` is a better name.

pkg/controller/download_request_controller_test.go
```diff
@@ -27,7 +27,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/clock"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
ashish-amarnath, line 30: this import may be combined with the above import block.

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -228,10 +229,11 @@ func (c *podVolumeBackupController) processBackup(req *velerov1api.PodVolumeBack
 	)
 
 	// if there's a caCert on the ObjectStorage, write it to disk so that it can be passed to restic
-	caCert, err := restic.GetCACert(c.backupLocationLister, req.Namespace, req.Spec.BackupStorageLocation)
+	caCert, err := restic.GetCACert(c.kbClient, req.Namespace, req.Spec.BackupStorageLocation)
```
ashish-amarnath, line 232: Would it be better to get the BSL object once and pass it to the subsequent functions?
This would require creating new APIs for this.
WDYT?

pkg/controller/restore_controller.go
```diff
@@ -48,6 +49,8 @@ import (
 	"github.com/vmware-tanzu/velero/pkg/util/collections"
 	kubeutil "github.com/vmware-tanzu/velero/pkg/util/kube"
 	"github.com/vmware-tanzu/velero/pkg/util/logging"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
ashish-amarnath, line 53: this import should be part of the import block from L31-37

pkg/controller/backup_deletion_controller_test.go
```diff
@@ -34,8 +35,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/clock"
 	"k8s.io/apimachinery/pkg/util/sets"
 	core "k8s.io/client-go/testing"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
ashish-amarnath, line 41: this is already imported as `velerov1` on line 40

pkg/controller/download_request_controller_test.go
```diff
@@ -27,7 +27,10 @@ import (
...
+
 	v1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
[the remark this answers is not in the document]
ashish-amarnath: looks like something is still using the `v1.`
Once we've removed all the v1 import alias, we move the double-import.

pkg/controller/gc_controller_test.go
```diff
@@ -32,6 +34,8 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	core "k8s.io/client-go/testing"
 
+	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
```
ashish-amarnath, line 37: can be combined with the import block from L-28-35.

pkg/controller/gc_controller_test.go
```diff
@@ -23,6 +23,8 @@ import (
 	"testing"
 	"time"
 
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
ashish-amarnath, line 26: double-import from L-39
and can be moved to L39

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -42,6 +42,8 @@ import (
 	"github.com/vmware-tanzu/velero/pkg/restic"
 	"github.com/vmware-tanzu/velero/pkg/util/filesystem"
 	"github.com/vmware-tanzu/velero/pkg/util/kube"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
ashish-amarnath, line 46: can be combined with import block from L-26-36

pkg/controller/restic_repository_controller.go
```diff
@@ -31,19 +32,21 @@ import (
...
 	"k8s.io/client-go/tools/cache"
 
-	v1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	velerov1client "github.com/vmware-tanzu/velero/pkg/generated/clientset/versioned/typed/velero/v1"
 	velerov1informers "github.com/vmware-tanzu/velero/pkg/generated/informers/externalversions/velero/v1"
 	velerov1listers "github.com/vmware-tanzu/velero/pkg/generated/listers/velero/v1"
 	"github.com/vmware-tanzu/velero/pkg/restic"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
ashish-amarnath, line 41: same comment as the one [here](https://github.com/vmware-tanzu/velero/pull/2561/files?file-filters%5B%5D=.go&file-filters%5B%5D=.md&file-filters%5B%5D=.mod&file-filters%5B%5D=.yaml&file-filters%5B%5D=No+extension#r442559843)

pkg/controller/restore_controller_test.go
```diff
@@ -18,11 +18,14 @@ package controller
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"io/ioutil"
 	"testing"
 	"time"
 
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
ashish-amarnath, line 27: double-import from L-40.
Can be moved to L-40

pkg/controller/pod_volume_backup_controller.go
carlisia: That was how the last version of this PR was. I'd rather not make any more changes to this. This is as close as the original as it can be.

carlisia: Not the last version actually, but a couple versions ago.

pkg/install/resources.go
```diff
@@ -19,16 +19,17 @@ package install
 import (
 	"time"
 
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
ashish-amarnath, line 22: can be moved to L-33

pkg/restic/common.go
```diff
@@ -17,11 +17,15 @@ limitations under the License.
 package restic
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strings"
 	"time"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
ashish-amarnath, line 26: group imports with import block from L-29-32

carlisia: I didn't think we were manually fussing with where imports go?

pkg/restic/common_test.go
```diff
@@ -17,18 +17,22 @@ limitations under the License.
 package restic
 
 import (
+	"context"
 	"os"
 	"sort"
 	"testing"
 
-	velerov1listers "github.com/vmware-tanzu/velero/pkg/generated/listers/velero/v1"
+	"k8s.io/apimachinery/pkg/runtime"
```
ashish-amarnath, line 25: combine with import block from L-27-35

site/docs/master/code-standards.md
```diff
@@ -43,8 +43,8 @@ Example:
     	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
     	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
     	corev1listers "k8s.io/client-go/listers/core/v1"
-        
-        velerov1api ""github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+       
+        velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
ashish-amarnath, line 47: This change should be copied over to the docs for the other versions as well.

site/docs/v1.3.1/code-standards.md
```diff
@@ -43,7 +43,7 @@ Example:
     	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
     	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
     	corev1listers "k8s.io/client-go/listers/core/v1"
-
```
ashish-amarnath, line 46: can this file can be reverted?

site/docs/v1.3.2/code-standards.md
```diff
@@ -43,7 +43,7 @@ Example:
     	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
     	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
     	corev1listers "k8s.io/client-go/listers/core/v1"
-        
+
```
ashish-amarnath, line 46: can this file can be reverted?

pkg/controller/pod_volume_backup_controller.go
ashish-amarnath: Yeah I agree the API was changed to accept a BSL object and that was a breaking API change.
I was hoping for a new API, not replacing the one that exists, to be created that works on the BSL object directly.

I think this change is still breaking the API. In that, I mean, if someone using this API in their code were to upgrade their dependency to a version of velero with this change, then their code wouldn't compile. No?

carlisia: Yes, it'll be a breaking change. I talked to @nrb about this, this is not an area of the code anyone should be importing. We'are also going to start moving things to inside a (to be added) `internal` package. Are you ok with this breaking change?

pkg/cmd/cli/backup/create.go
carlisia: The answers are `no` and `yes`. This is because this PR only changes the api client for the BSL resource. So if you want to handle, say, backup resources, the client is not changing.

pkg/client/factory.go
carlisia: I'm not sure. Looking at the other clients, none has version in the name:
image

`Client` would be the best, but it's already in use so we need something to differentiate this new kubebuilder client.

@nrb do you have any opinion on this?

pkg/cmd/cli/restic/server.go
carlisia: Good point!

pkg/cmd/server/server.go
carlisia: I'm doing a `s.logger.Fatal` so I don't think it's needed?

pkg/cmd/util/output/backup_storage_location_printer.go
carlisia: To clarify: I changed this (seemly unrelated to this PR) field to match the new column `Phase` we will be displaying with `kubectl`:

image

Pls ignore the 2 in `PHASE2`, I haven't refreshed my cluster.

The reason why I chose `phase` instead of `status` is because the field is `status.phase`, so this is more accurate and also leaves room to add other `status.x` fields if we want to.

pkg/controller/gc_controller_test.go
carlisia: Fixed.

pkg/controller/restore_controller_test.go
carlisia: Fixed.

carlisia:
@ashish-amarnath I fixed your addressable suggestions, thank you! Others I added comments for consideration.

I did not fix the import grouping suggestions. I realize we talked about this way before you joined the project: we were considering cleaning up imports a low value task. But now with the linter, maybe we can do better in this department. Let's take up this convo there and figure out what we do going forward. I appreciate the attention to detail!

pkg/controller/pod_volume_backup_controller.go
ashish-amarnath: As long as none of our plugins are using these functions, we should be OK.
Ideally we should be deprecating APIs across releases.

pkg/cmd/util/output/backup_storage_location_printer.go
nrb: I like `Phase` here, since `Status` is the name of a field that contains other fields.

It also directly maps to the field's name.

nrb:
RE: import order, let's discuss it in the linter settings on https://github.com/vmware-tanzu/velero/pull/2615. I would rather we get it set in the linter settings than debate it in each PR, or ask us to manually do the linter's work.

One thing I appreciate about Go is that we can lean on tools like `gofmt` to do the formatting for us. I think we should get the setting we want into `goimports`, then let it take care of things.

pkg/controller/pod_volume_backup_controller.go
nrb: @ashish-amarnath Yeah, that would be ideal, though we've got a lot of examples of where we move our code around without deprecating.

I think I'm ok with this change, personally, even though it is breaking with strict semantic versioning. We have a lot of "accidentally" public code that we've been pretty rude about moving around in the past, and I think it would be good to document our true, intentional public API.

Carlisia made https://github.com/vmware-tanzu/velero/issues/2654 to do this, and we'll ask folks to comment there as we move things into `internal`. If moving something breaks someone, we can revisit.

This definitely isn't ideal, but we've also got to be able to move forward.

pkg/client/factory.go
nrb: Looking through what's returned, `VeleroV1APIClient` doesn't quite make sense to me, since we get https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client?tab=doc#Client, which is a generic client that can interact with anything in the Kubernetes API server, so it's not really just for Velero V1; we just happen to be using it for that now. Really, we could register any API group/scheme to it and use this client to get them.

In the current file, `Client` returns https://pkg.go.dev/github.com/vmware-tanzu/velero@v1.4.0/pkg/generated/clientset/versioned?tab=doc#Interface, which has the `VeleroV1()` method, which makes it a typed and specific client.

I'm leaning towards keeping the `KubeBuilderClient` name for now, then when we complete the transition, renaming it to `KubeClient`, which would replace the existing `KubeClient` for API server interactions. That's a much further-looking plan, though.

Thoughts?

pkg/cmd/server/server.go
nrb: 👍 to a comment for explanation on `GetAPIReader`. We went over this on a call, but I'm likely to forget in a week 😄

pkg/cmd/server/server.go
nrb: Yeah, I would think if the error's handled, it shouldn't also be returned.

pkg/apis/velero/v1/backupstoragelocation_types.go
```diff
@@ -21,33 +21,81 @@ import (
...
+
+	// AccessMode is an unused field.
+	//
+	// Deprecated: there is now an AccessMode field on the Spec and this field
+	// will be removed entirely as of v2.0.
+	// +optional
+	AccessMode BackupStorageLocationAccessMode `json:"accessMode,omitempty"`
+}
+
+// +genclient
```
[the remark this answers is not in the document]
carlisia: No. I've added this comment: `// TODO(2.0) After converting all resources to use the runttime-controller client,
// the genclient and k8s:deepcopy markers will no longer be needed and should be removed.`

PROJECT
```diff
@@ -0,0 +1,7 @@
+domain: io
+repo: github.com/vmware-tanzu/velero
+resources:
+- group: velero
+  kind: BackupStorageLocation
+  version: v1
+version: "2"
```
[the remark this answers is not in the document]
carlisia: It was defaulted to this value. I'll find out.

pkg/cmd/server/server.go
carlisia: Done.

PROJECT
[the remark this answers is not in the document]
carlisia: Answer from the `kubebuilder` k8s channel:

```
Kubebuilder's major version is 2
And tracked for any required breaking upgrade helpers in the future
```

pkg/controller/backup_controller.go
```diff
@@ -57,6 +58,8 @@ import (
 	kubeutil "github.com/vmware-tanzu/velero/pkg/util/kube"
 	"github.com/vmware-tanzu/velero/pkg/util/logging"
 	"github.com/vmware-tanzu/velero/pkg/volume"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
alaypatel07, line 62: qq: should we alias this package to kbclient in order to maintain consistency?

pkg/controller/backup_controller_test.go
```diff
@@ -18,8 +18,12 @@ package controller
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
alaypatel07, line 25: please see alias comment question, thanks

pkg/restic/common.go
alaypatel07, line 26: duplicate import for the same package see line below, not sure if I am missing something but looks like a typo

pkg/cmd/cli/backuplocation/get.go
```diff
@@ -36,19 +40,24 @@ func NewGetCommand(f client.Factory, use string) *cobra.Command {
 			err := output.ValidateFlags(c)
 			cmd.CheckError(err)
 
-			veleroClient, err := f.Client()
+			client, err := f.KubebuilderClient()
 			cmd.CheckError(err)
 
-			var locations *api.BackupStorageLocationList
+			locations := new(velerov1api.BackupStorageLocationList)
```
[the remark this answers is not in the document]
ashish-amarnath: Here we want to create a new list object and not a slice of backupstoragelocations.
This, IMO, is correct.

pkg/controller/download_request_controller_test.go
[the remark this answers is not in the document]
ashish-amarnath: This still needs to be addressed.

internal/util/managercontroller/managercontroller.go
```diff
@@ -0,0 +1,57 @@
...
+// is called or when the parent channel is closed, whichever happens first.
+//
+// Note the caller must *always* call the CancelFunc, otherwise resources may be leaked.
+func contextForChannel(parentCh <-chan struct{}) (context.Context, context.CancelFunc) {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	go func() {
+		select {
+		case <-parentCh:
+			cancel()
```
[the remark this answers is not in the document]
carlisia: Not sure if you mean instead of cancel() otherwise there's no need to add a return, that's the default behavior.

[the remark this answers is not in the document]
carlisia: I mean, it returns from the switch but we falls down to return ctx, cancel in the end. I don't think it's needed to add that here, this seems fine to me. If I misunderstood please let me know.

internal/util/managercontroller/managercontroller.go
```diff
@@ -0,0 +1,57 @@
...
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
shomron: How is this case a no-op? The goroutine's mission is to either close the child context if the parent is closed, or exit when only the child is cancelled. The child can be cancelled without the parent being closed. Without the second case - the goroutine would leak.

internal/util/managercontroller/managercontroller.go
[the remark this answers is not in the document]
shomron: There's no need for an explicit return - if this case is hit it will cancel the child context, exit the select statement, and then exit the goroutine.

api/v1/backupstoragelocation_types.go
ashish-amarnath: KubeBuilder is able to generate RBAC for the controllers to run wit, based on the annotation. The question was whether we want to use that. We can table that discussion for when we move all the controllers.

PR #2486, opened 2020-04-30
Revert "Install ca-certificates for ARM based container builds. (#2481)"

skriss:
This reverts commit 0e5ca82dba08c799f6b0f5e2b3479cb20eb3c3ee.

We discovered that #2481 breaks the ARM container builds (see https://github.com/vmware-tanzu/velero/pull/2481#issuecomment-621982600), so we need to back it out for now.

PR #2566, opened 2020-05-24
ca-certificates is needed to work with aws s3 - missing on arm/ppc

mickkael:
The package is present on amd64, but not on arm/arm64 and ppc

Signed-off-by: mickkael <19755421+mickkael@users.noreply.github.com>

### DIFF
--- a/Dockerfile-velero-arm
+++ b/Dockerfile-velero-arm
@@ -14,6 +14,10 @@
 
 FROM arm32v7/ubuntu:focal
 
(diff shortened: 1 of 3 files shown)


ashish-amarnath:
This change was previously made but had to be reverted as it was breaking the ARM container builds. https://github.com/vmware-tanzu/velero/pull/2486
The build setup is yet to fixed so this PR. Holding this PR till then.

PR #2588, opened 2020-05-28
🏃‍♂️ Travis decommission

ashish-amarnath:


.github/workflows/push.yml
```diff
@@ -1,4 +1,4 @@
-name: Push
+name: CI
```
nrb, line 1: I'm still not sure on this name; it's not very descriptive when looking at it via the GH dashboard, but then is _too_ specific when used in the badge...

nrb: Screen Shot 2020-05-28 at 3 11 58 PM

For example, here it would be "CI" & "Pull Request CI Check".

ashish-amarnath: That's the line I am trying to walk!

skriss: Couple ideas:
- we could call it "master CI" or "master build", and just ignore the fact that it also runs for tags since that's very infrequent.
- or, we could actually separate out the tag builds into their own job, so that "master build" for this one is completely accurate.
    - this might be necessary down the road anyway, if we want to integrate the goreleaser process since that would only run for tags

ashish-amarnath: Renamed to "Master CI".
Thanks for the suggestion!

PR #2602, opened 2020-06-02
Azure: support aad-pod-identity when using restic

skriss:

related to https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure/pull/51


skriss:
I've successfully tested Azure + restic with both `aad-pod-identity` and service principal auth (related to https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure/pull/51#issuecomment-639089643) and can confirm both work with this change. This is now ready for review.

pkg/restic/azure.go
```diff
@@ -67,35 +73,29 @@ func getStorageAccountKey(config map[string]string) (string, *azure.Environment,
...
-	// 5. we need config["resourceGroup"], config["storageAccount"]
+	// we need config["resourceGroup"], config["storageAccount"]
 	if _, err := getRequiredValues(mapLookup(config), resourceGroupConfigKey, storageAccountConfigKey); err != nil {
 		return "", env, errors.Wrap(err, "unable to get all required config values")
 	}
 
-	// 6. get SPT
-	spt, err := newServicePrincipalToken(envVars[tenantIDEnvVar], envVars[clientIDEnvVar], envVars[clientSecretEnvVar], env)
+	// get authorizer
+	authorizer, err := auth.NewAuthorizerFromEnvironment()
```
ashish-amarnath, line 88: Can you please add a comment here indicating which environment variable the authorizer is getting created from?
From https://godoc.org/github.com/Azure/go-autorest/autorest/azure/auth#NewAuthorizerFromEnvironment
Looks like this can be from
> 1. Client credentials 2. Client certificate 3. Username password 4. MSI
in that order

ashish-amarnath: Am I reading that correctly?

skriss: Yep, that's the standard Azure "credentials chain". Previously we only supported #1 from that list. Do you want me to copy/paste the godoc from that method in our code?

ashish-amarnath: That’s what i was thinking. Esp in scenario where someone reports this error. But that should be fine to investigate that API failure.

skriss: comment updated!

Issue #1871, opened 2019-09-12
Add option to back up all volumes by default with restic


skriss:
**Describe the problem/challenge you have**
Right now, Velero uses an "opt-in" approach to getting volumes backups with restic. Every pod containing a volume to be backed up must be annotated with the volume's name in order for Velero/restic to back it up. This can be burdensome for users.

**Describe the solution you'd like**
We should consider a mode where volumes are backed up by default if restic is enabled, and the user can "opt-out" any volumes that they don't want to be backed up.

Some questions around how this should work:
- should we back up *all* pod volumes, or just PVCs, by default?
- what does the "opt out" UX look like? Annotation on pod/PVC?
- should we add a new spec field to backups, e.g. `ResticBackupVolumes`, that must be true in order to get the opt-in behavior? Or should it be the default behavior?

xref #1586 (item 2)

skriss:
Look at backing up PVCs that don't have snapshotters by default
Annotation to opt out

prydonius:
Would this be considered a breaking change? Users using the annotation today to backup select volumes will find that all their PVC volumes are now being backed up.

skriss:
Still need to flesh out a detailed proposal, but it potentially could be a breaking behavior change (though unlikely to be a breaking API change). If we're not comfortable with that, we can choose to add a flag somewhere (on the backup, on the server) to opt into this behavior.

zakkg3:
check this 
https://github.com/zakkg3/podannotator

Just run `annotator namespace1 namespace2` and this will annotate your pvc's to be backed up by velero.

carlisia:
I added this topic to our community meeting tomorrow (3/31): https://hackmd.io/Jq6F5zqZR7S80CeDWUklkA?both#March-31-2020.

skriss:
cc @stephbman

stephbman:
I've added the topic to our community meeting agenda for 5/12 here: https://hackmd.io/Jq6F5zqZR7S80CeDWUklkA?view

A detailed product requirements document can be accessed here:
https://docs.google.com/document/d/1kDbxMQTxiXp6a_G37BmboGM4Va1FI5WObKacJkF_L-g/edit?usp=sharing

Relates to these issues:
https://github.com/vmware-tanzu/velero/issues/605
https://github.com/vmware-tanzu/velero/issues/2375

cc: @skriss and @carlisia

PR #2611, opened 2020-06-04
✨  Implement option to back up all volumes by default with restic

ashish-amarnath:

Fixes: #1871 

- Introduces a new field `DefaultVolumesToRestic ` in the Backup spec.
- Adds a new CLI flag to `velero backup create` to use restic on all pod volumes by default on a per backup basis.
- Adds a new CLI flag `--default-volumes-to-restic` to the velero server to use restic on all pod volumes by default for all backups.
- Adds a new CLI flag `--default-volumes-to-restic` to the `velero install` command to enable this behavior for all backups on an installation at install time.
- Adds a new API `GetPodVolumesUsingRestic` to get pod volumes that should be backed up using restic:
    - If restic is not to be used by default, get volumes based on `backup.velero.io/backup-volumes` annotation on the pod.
    - If restic is to be used by default, get all volumes excluding volumes:
        - from `backup.velero.io/backup-volumes-excludes` annotation on the pod.
        - volumes mounting the default service account token.
        - hostpath volumes.


ashish-amarnath:
This PR looked good in testing. But going to confirm that this PR covers everything in https://docs.google.com/document/d/1i5IQfuOtsHhGuYiiTrCZ6zjVhtKWLRqV/edit

nrb:
This looks like it's an option on `velero backup create`. I understood the feature to be on `velero server` as a flag.

Could we unify the approach such that there's a flag on `velero server` that will add this flag to the backup objects? This would give both us and users documentation that it was requested, as well as not requiring it to be specified per-backup.

The counter argument is that most operations would be on a scheduled basis, so the user isn't required to pass the flag to `velero backup create` that often.

ashish-amarnath:
I think it was preferred that we add this option at the backup API level instead of the server level.
We could definitely unify this with a server level flag and use restic by default if both the server level flag and the api field are asking for restic to be used by default.

skriss:
Yeah, I think we talked about having both, so something like:

- the Backup CR has a nullable bool (like you already have)
- `velero server` has a bool flag (`--backup-all-volumes-with-restic-by-default` or similar) that defaults to `false` to preserve backwards compat
- when a Backup is created: 
    - if the Backup CR's field is `nil`, then default it to the value from the server flag
    - if the Backup CR's field is not `nil`, then retain its value since the user explicitly set it (so it takes precedence over the server-level flag)

ashish-amarnath:
I wasn't part of that conversation and I just went by [this doc](https://docs.google.com/document/d/1i5IQfuOtsHhGuYiiTrCZ6zjVhtKWLRqV/edit)

> Question: Will this be implemented at the client level or the server lever and what are the advantages/disadvantages of each?
Could be a velero server level flag (do not want to maintain one flag for the binary) 
Server flag specify default value 
Or API level back up so user specifies opt-in or opt-out 
Preferred 

But now there is a `--default-restic` flag on the server too 😄

pkg/backup/backup.go
```diff
@@ -240,6 +243,11 @@ func (kb *kubernetesBackupper) Backup(log logrus.FieldLogger, backupRequest *Req
 	log.Infof("Including resources: %s", backupRequest.ResourceIncludesExcludes.IncludesString())
 	log.Infof("Excluding resources: %s", backupRequest.ResourceIncludesExcludes.ExcludesString())
 
+	if backupRequest.Backup.Spec.DefaultRestic == nil {
+		backupRequest.Backup.Spec.DefaultRestic = &kb.defaultRestic
+	}
+	log.Infof("Backing up all pod volumes using restic: %t", *backupRequest.Backup.Spec.DefaultRestic)
```
skriss, line 249: I'd probably move this logic into the backup controller where we have some other defaulting -- somewhere in the [prepareBackupRequest](https://github.com/vmware-tanzu/velero/blob/d9d9dc60da8e82a205305d4fe54daf36e2f40f4a/pkg/controller/backup_controller.go#L318) method. It should all move out of the controller eventually and into a library, but this'll keep similar logic together.

skriss:
> I wasn't part of that conversation and I just went by this doc

Fair enough - we should've updated the doc :) I'm also not 100% sure I'm recalling correctly, but either way, I do think having both the per-backup option as well as the server-level default is the best of both worlds so I'm :+1: to the change.

Code mostly LGTM, but I'll take a fresh look next week.

pkg/backup/backup.go
ashish-amarnath: done!

nrb:
> I wasn't part of that conversation and I just went by this doc

Thanks - honestly I don't know if I was going by a conversation, but having the behavior defined on the server seems like it might be more user-friendly to me. But I do also like having the decision documented on the `Backup` object, too.

pkg/apis/velero/v1/backup.go
```diff
@@ -82,6 +82,12 @@ type BackupSpec struct {
 	// VolumeSnapshotLocations is a list containing names of VolumeSnapshotLocations associated with this backup.
 	// +optional
 	VolumeSnapshotLocations []string `json:"volumeSnapshotLocations,omitempty"`
+
+	// DefaultRestic specifies whether restic should be used to take
+	// backup of all pod volumes by default.
+	// +optional
+	// + nullable
+	DefaultRestic *bool `json:"defaultRestic,omitempty"`
```
skriss, line 90: I'm trying to think of a more descriptive name for this field that's also not a full sentence, and failing...I'll keep thinking

skriss: Maybe `ResticBackupVolumesByDefault`? A little longer but also I think more descriptive? Open to other suggestions too.

pkg/restic/common.go
```diff
@@ -125,6 +135,47 @@ func GetVolumesToBackup(obj metav1.Object) []string {
...
+	podVolumes := []string{}
+	for _, pv := range pod.Spec.Volumes {
+		// cannot backup hostpath volumes as they are not mounted into /var/lib/kubelet/pods
+		// and therefore not accessible to the restic daemon set.
+		if pv.HostPath != nil {
+			continue
+		}
+		// don't include volumes that are in the exclude list
+		// and the volume mounting the default service account token.
+		if !contains(volsToExclude, pv.Name) && !strings.HasPrefix(pv.Name, "default-token") {
```
skriss, line 172: I would probably invert this logic to remain consistent with the previous host path case (and I think this is considered more idiomatic go), i.e.:

```go
if contains(volsToExclude, pv.Name) {
    continue
}
if strings.HasPrefix(pv.Name, "default-token") {
    continue
}

podVolumes = append(podVolumes, pv.Name)
```

pkg/apis/velero/v1/backup.go
skriss: @nrb @carlisia WDYT on naming?

pkg/restic/common.go
ashish-amarnath: done

pkg/apis/velero/v1/backup.go
ashish-amarnath: I spent like20m thinking of a better name and failed.
Suggestions welcome :)

nrb: `DefaultVolumesToRestic`?

skriss: works for me

ashish-amarnath: `DefaultVolumesToRestic` is the winner

ashish-amarnath:
This looked good in testing and is ready for review. @nrb and @carlisia PTAL

carlisia:
> @carlisia How does this affect #2561?

There'll be merge conflicts and a need to regenerate the crds/manifests, but nbd.

pkg/controller/backup_controller.go
```diff
@@ -120,6 +122,7 @@ func NewBackupController(
 		volumeSnapshotLister:        volumeSnapshotLister,
 		volumeSnapshotContentLister: volumeSnapshotContentLister,
 		newBackupStore:              persistence.NewObjectBackupStore,
+		defaultVolumesToRestic:      defaultVolumesToRestic,
```
carlisia, line 125: Nit: it'd be good to place this consistently under the `defaultBackupLocation`, as with the other two references.

site/docs/master/customize-installation.md
```diff
@@ -45,6 +46,14 @@ By default, `velero install` does not install Velero's [restic integration][3].
 
 If you've already run `velero install` without the `--use-restic` flag, you can run the same command again, including the `--use-restic` flag, to add the restic integration to your existing install.
 
+## Default Pod Volume Backup To Restic
```
carlisia, line 49: This header needs to be lower-cased:
-> Default Pod Volume backup to restic

site/docs/master/restic.md
```diff
@@ -160,6 +160,86 @@ You're now ready to use Velero with restic.
 
 ## Back up
```
carlisia, line 161: I know you didn't add this, but please make this `Backup`.

site/docs/master/restic.md
```diff
@@ -160,6 +160,86 @@ You're now ready to use Velero with restic.
...
 
+Velero supports two approaches of discovering pod volumes that need to be backed up using restic:
```
carlisia, line 163: `s/pod volumes/Pod Volumes`

Also: please capitalize resource names below.

site/docs/master/restic.md
ashish-amarnath: "Backup" is the noun and "Back up" is the verb. Here we are listing steps to perform back up, using the verb and not the noun, correct?

carlisia: I think you're right. It might be good to make it `To back up` and `To restore` (next section).

site/docs/master/restic.md
ashish-amarnath: The existing docs in the repo all refer to this as "pod volumes".
Also this is not referring to a kubernetes type.

skriss:
🎉

Issue #2476, opened 2020-04-29
Add a CI check for changelog file


skriss:
We should add a CI check for a changelog file. I think it should be required by default, perhaps unless the PR is explicitly labeled with a `no-changelog-required` label. This would make it easier for contributors and reviewers alike.

We probably also want to add a PR template with instructions about DCO and the changelog file, to help new contributors understand.

PR #2613, opened 2020-06-05
🏃‍♂️Add a CI check for changelog file

ashish-amarnath:

Fixes: #2476


skriss:
🎉 very excited for this! I'll review in detail next week.

skriss:
BTW, you [forgot to add a changelog](https://github.com/vmware-tanzu/velero/pull/2613/checks?check_run_id=744029564) 😂

ashish-amarnath:
Yeah. Left that out on purpose to confirm that this check works! 😀

.github/workflows/pr-changelog-check.yml
```diff
@@ -0,0 +1,14 @@
+name: Pull Request Changelog Check
+on: [pull_request]
+jobs:
+
+  build:
+    name: Run CI
```
nrb, line 6: I think we should have a more descriptive name here, otherwise we'll have 2 "Run CI" workflows in the UI :)

hack/changelog-check.sh
```diff
@@ -0,0 +1,43 @@
...
+# GITHUB_REF is something like "refs/pull/:prNumber/merge"
+pr_number=$(echo $GITHUB_REF | cut -d / -f 3)
+
+change_log_file="${CHANGELOG_PATH}/${pr_number}-*"
+
+if ls ${change_log_file} 1> /dev/null 2>&1; then
+    echo "changelog for PR ${pr_number} exists"
+    exit 0
+else
+    echo "Please consider adding a changelog for PR ${pr_number}"
```
nrb, line 40: Will this prevent merging? Or can we make it non-blocking?

ashish-amarnath: It is non-blocking.
Blocking/Non-Blocking is a setting in github itself, outside of this job.

nrb: Ok, cool. I agree with non-blocking and using it as an easier check.

skriss: It'd be a nice enhancement to make it required, but have it not fail if there's a `no-changelog-required` label on the PR (so we can explicitly opt out of requiring one for certain PRs). But, i'm not exactly sure how easy that is, and we can keep that in mind as a follow-up enhancement.

nrb: Looks like labels are presented to Actions, since there's [this Action, which can check for required labels](https://github.com/marketplace/actions/verify-pull-request-labels). But I think it's also something we can do as an enhancement later.

skriss, line 40: Can we add a link to https://velero.io/docs/master/code-standards/#adding-a-changelog here since that has some more detail on what exactly to do?

nrb: `if: contains(github.event.pull_request.labels.*.name, 'some-defined-label')` should be able to skip

ashish-amarnath: Yep, I was looking at something similar.
I'll do that as an enhancement.

ashish-amarnath: added

PR #2627, opened 2020-06-12
Design: Backup Resources of Specific Type by Specified Order

phuongatemc:
Proposes a solution that allows user to specify a backup order for resources of specific resource type.


changelogs/unreleased/2627-phuong-hoang
```diff
@@ -0,0 +1 @@
+Proposes a solution that allows user to specify a backup order for resources of specific resource type.
```
nrb, line 1: You can delete this file - it's not necessary for design documents, only for user-facing changes.

design/backup-resources-order.md
```diff
@@ -0,0 +1,39 @@
+## Backup Resources Order
+This document proposes a solution that allows user to specify a backup order for resources of specific resource type.
+
+## Background
+During backup process, user may need to back up resources of specific type in some specific order to ensure the resources were backup properly because these resources are related and ordering might be required to preserve the consistency for the apps to recover itself �from the backup image 
+(Ex: master-slave database pods in a cluster).
```
nrb, line 6: Let's use `primary-secondary` instead.

design/backup-resources-order.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Goals
+- Enable user to specify an order of back up resources belong to specific resource type
+
+## Alternatives Considered
+- Use a plugin to backup an resources and all the sub resources.  For example use a plugin for StatefulSet and backup pods belong to the StatefulSet in specific order.  This plugin solution is not generic and requires plugin for each resource type.
+
+## High-Level Design
+User will specify a map of resource type to list resource names (separate by semicolons).  Each name will be in the format "namespaceName/resourceName" to enable ordering accross namespaces.  Based on this map, the resources of each resource type will be sorted by the order specified in the list of resources.  If a resource instance belong to that specific type but its name is not in the order list, then it will be put behind other resources that are in the list.
```
nrb, line 15: If `resourceName` is specified, but does not exist in the cluster, what should the behavior be? Generally, Velero has been treating these as cluster problems and issues a warning, but completes the backup operation successfully. I'm curious if that's acceptable from your use case.

nrb, line 15: This design relies on the user specifying explicit/exact resource names, is that correct? This is unique within Velero right now, as generally we don't address individual resources. That's not necessarily a blocker, I just wanted to point it out.

What are you envisioning using this with most often? Pods owned by a StatefulSet? I can see the naming being fairly consistent/predictable there, though with pods owned by a Deployment they're less predictable by design, since Kubernetes is supposed to manage them behind the scenes.

design/backup-resources-order.md
```diff
@@ -0,0 +1,39 @@
...
+Function getResourceItems collects all items belong to a specific resource type.  This function will be enhanced to check with the map to see whether the BackupOrders has specified the order for this resource type.  If such order exists, then sort the items by such order being process before return.
+
+### Changes to velero CLI
+Add new flag "--backup-orders" to Velero backup create command which takes a string of key-values pairs which represents the map between resource type and the order of the items of such resource type. Key-value pairs are separated by semicolon, items in the value are separated by commas.
+
+Example:
+>velero backup create mybackup --backup-orders "pod=ns1/pod1,ns1/pod2;persistentvolumeclaim=n2/slavepod,ns2/masterpod"
+
+## Open Issues
+- In the CLI, the design proposes to use commas to separate items of a resource type and semicolon to separate key-value pairs.  This follows the convention of using commas to separate items in a list (For example: --include-namespaces ns1,ns2).  However, the syntax for map in labels and annotations use commas to seperate key-value pairs.  So it introduces some inconsistency.
```
nrb, line 39: Thanks for including this note! I appreciate the call out, though I'm not sure I see a way around it since we have to account for namespaced resources.

design/backup-resources-order.md
nrb, line 15: How could cluster-scoped resources be handled? Unfortunately, I don't have a concrete example here, other than some theoretical CRs that may depend on each other.

This may or may not be a concern, but I do want to ask. If the theoretical CRDs worked like PVs, there would be a namespaced resource that is similar to a PVC pointing to them that could be used to order instead.

design/backup-resources-order.md
```diff
@@ -0,0 +1,39 @@
...
+### Changes to BackupSpec
+Add new field to BackupSpec
+
+    type BackupSpec struct {
+        ...
+        // BackupOrders contains a list of key-value pairs that represent the order
+        // of backup of resources that belong to specific resource type
+        // +optional
+        // +nullable
+        BackupOrders map[string]string
```
nrb, line 26: I think this name is a little unclear. Perhaps `OrderedResources`? Are you intending to use this map to do ordered restores at some point too? That might help in naming.

design/backup-resources-order.md
phuongatemc: In my usage, we will first fetch the list of pods from k8s API Server using app label selector, then processes this list to generate a sequence of pods to backup to ensure App Consistent backup so the pods should exist.  However, in general, I agree that we should give a warning as you mentioned if the resource doesn't exist.

phuongatemc: As of now, we go with the exact match on the Pod names (resources' name).  However, this design may be enhanced to support regular expression search on resources' name.  It may not be as efficient but should be OK.

My usage will only be Pods even though this design support any resource type.  We do a little bit in the preprocessing before sending the sequence of pods down to Velero Backup.  First we fetch the list of Pods, then trace the parents.  Then based on our configuration of the application, we will intelligently generate the sequence the Pods based on the configuration of the application.  For example, if the application comprised of 2 StatefulSets and we need to backup the secondary statefulset first, then we will pick out all pods belong to the secondary StatefulSet and put them on the sequence before putting the pods belong to the primary StatefulSet.  Another example would be if the application comprises of 1 StatefulSet with the pod name "pod-0" is the primary and "pod-1", "pod-2" etc. are secondary then we will generate the sequence "pod-1,pod-2, pod-0" to make sure the secondary pods will be backup before the primary pod.

phuongatemc: In my usage, we only work within a context of a namespace.  However, this design should work across namespace or in the cluster scope because the resourceName is specified as "namespace/resourceName".  If the resources have cluster scopes (for example PersistentVolumes), then we simply use pvName without namespace prefix.  The implementation is the same because we look up the resource from the resource map based on the resourceName.

design/backup-resources-order.md
phuongatemc: I agree.  Let's use OrderedResources.
We currently have no use for the order in Restore datapath but I think someone else would so it is better to keep it generic.

design/backup-resources-order.md
phuongatemc: Sure!

changelogs/unreleased/2627-phuong-hoang
phuongatemc: Done.

phuongatemc: The master build failed ("Changelog check") without this file.  So I will have to put it back.

nrb: If you look, the changelog check isn't required. It's just a check for us. I've also added the changelog-not-required label to this PR.

nrb:
@phuongatemc Can you please amend your commits to have a DCO sign off? Use `git commit --amend --sign-off` to add a sign off so that the DCO check passes and we can accept all commits. Thanks!

design/backup-resources-order.md
```diff
@@ -0,0 +1,39 @@
...
+(Ex: primary-secondary database pods in a cluster).
+
+## Goals
+- Enable user to specify an order of back up resources belong to specific resource type
+
+## Alternatives Considered
+- Use a plugin to backup an resources and all the sub resources.  For example use a plugin for StatefulSet and backup pods belong to the StatefulSet in specific order.  This plugin solution is not generic and requires plugin for each resource type.
+
+## High-Level Design
+User will specify a map of resource type to list resource names (separate by semicolons).  Each name will be in the format "namespaceName/resourceName" to enable ordering accross namespaces.  Based on this map, the resources of each resource type will be sorted by the order specified in the list of resources.  If a resource instance belong to that specific type but its name is not in the order list, then it will be put behind other resources that are in the list.
```
ashish-amarnath, line 15: I think this proposal works well for pods that are managed by a statefulset and may not be very well suited for pods managed by deployments or daemonsets... As the pods will have generated names. I think it may be worth calling that out in the `Goals` section of the proposal.
I mention pods here, but any resources with a generated name will have the same issue.

phuongatemc: All the pods' name will be explicitly specified in the "OrderedResources" so it should not matter whether the pods belong to StatefulSet or Deployment or, DaemonSet etc. 
In my usage, the controller code will fetch all pods with specific label (say "app:mysql") and preprocessed them to generate a sequence of podNames to pass to Velero Backup's OrderredResources".

ashish-amarnath: The point I was trying to make was that pods that are controlled by a statefulset have predicable names, e.g.`prometheus-1`, `prometheus-2`. However, for pods controlled by deployments and daemonsets, their names will be generated, e.g. `restic-jnxj4` and `velero-78d94698d6-rh8hc`.

Furthermore, if pod `pometheus-1`, controlled by a statefulset, were to crash, then the pod replacing it will also be named `prometheus-1` however the same does not happen with pods xontrolled by deployments or daemonsets. if pod `velero-78d94698d6-rh8hc` were to crash, then its replacement will be a pod whose name will be `velero-<random>`.

Considering that pods, which are ephemeral resources, can be replaced and that their names can change, listing pods controlled by deployments or daemonsets may not work as seamlessly as it would for statefulsets.
I was hoping that this could be called out explicitly in the goals of the proposal.

go.sum
```diff
@@ -412,8 +412,8 @@ golang.org/x/crypto v0.0.0-20190605123033-f99c8df09eb5/go.mod h1:yigFU9vqHzYiE8U
 golang.org/x/crypto v0.0.0-20190611184440-5c40567a22f8/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
 golang.org/x/crypto v0.0.0-20190617133340-57b3e21c3d56/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
 golang.org/x/crypto v0.0.0-20190820162420-60c769a6c586/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
-golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
 golang.org/x/crypto v0.0.0-20191206172530-e9b2fee46413/go.mod h1:LzIPMQfyMNhhGPhUkYOs5KpL4U8rLKemX1yGLhDgUto=
+golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
```
ashish-amarnath, line 416: Can you please revert changes to this file as well?

design/backup-resources-order.md
phuongatemc: Thanks.  I called out this limitation in the backup-resources-order.md.

go.sum
phuongatemc: Done.

PR #2635, opened 2020-06-16
add a supported provider: Storj object storage

JessicaGreben:
Storj Labs offers decentralize file object storage which is perfect for disaster recovery and backups since there is no single point of failure like traditional cloud providers because the data is stored around the globe instead of in a single datacenter. We are excited to add a plugin for Velero so that there is simple integration to backup kubernetes clusters to Storj's decentralized file object storage.


JessicaGreben:
@nrb thank you for that extra info, i made those changes.

ashish-amarnath:
@JessicaGreben Thanks for supporting Velero on your storage platform!
Can you please add a changelog to this PR. https://github.com/vmware-tanzu/velero/pull/2635/checks?check_run_id=778255227

JessicaGreben:
@ashish-amarnath thank you, i made that change

Issue #2540, opened 2020-05-15
Update Kubernetes dependency to 1.18


turkenh:
**What steps did you take and what happened:**
Getting a lot of errors similar to following when I want to build my project, which depends velero, with k8s 1.18:

```
pkg/generated/clientset/versioned/typed/velero/v1/backup.go:74:5: not enough arguments in call to c.client.Get().Namespace(c.ns).Resource("backups").Name(name).VersionedParams(&options, scheme.ParameterCodec).Do
	have ()
	want (context.Context)
```

This is because velero depends on k8s 1.17 and clientset generated with that version.

**What did you expect to happen:**
I can consume velero together with latest Kubernetes versions.

deepakkinni:
Is there an ETA for this?

laverya:
I've opened a PR to update to k8s 1.18 - only major problem I foresee is that the external-snapshotter import is now a prerelease version (since the currently released version is still k8s 1.17)

PR #2651, opened 2020-06-18
k8s 1.18 import

laverya:
fixes https://github.com/vmware-tanzu/velero/issues/2540

One complication with updating - [external-snapshotter](https://github.com/kubernetes-csi/external-snapshotter/blame/master/go.mod) imports k8s.io/kubernetes _directly_ which requires pinning k8s api versions with `replace` directives in the go.mod file. It appears that previously external-snapshotter imported a pre-go-mod version of k8s.io/kubernetes, and so didn't trigger this behavior. (Also, it's a prerelease version - so that'll also need to be fixed)


laverya:
I'll happily merge/rebase/etc this to fix conflicts in the generated code, but I'd prefer to do that once instead of every time the CRD is updated in master if that makes sense

nrb:
@laverya Yeah, that makes sense. I think we'll get #2561 in first, then this. We're not planning on hitting every CRD in the 1.5 development cycle.

laverya:
@nrb I see that 2561 has been merged - should I resolve conflicts here?

carlisia:
@laverya yes, please def resolve the conflicts when you have a chance.

carlisia:
Actually, I don't know how close to ready this PR is, might want to hold off on https://github.com/vmware-tanzu/velero/pull/2674, which also touches the BSL CRD.

pkg/cmd/cli/restic/server.go
```diff
@@ -220,7 +220,7 @@ func (s *resticServer) validatePodVolumesHostPath() error {
 		}
 	}
 
-	pods, err := s.kubeClient.CoreV1().Pods("").List(metav1.ListOptions{FieldSelector: fmt.Sprintf("spec.nodeName=%s,status.phase=Running", os.Getenv("NODE_NAME"))})
+	pods, err := s.kubeClient.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{FieldSelector: fmt.Sprintf("spec.nodeName=%s,status.phase=Running", os.Getenv("NODE_NAME"))})
```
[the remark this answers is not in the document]
laverya: definitely can do

nrb:
@laverya Talking to @carlisia, we'll get #2674 merged then we can pause on any other changes on controllers to get this change in.

laverya:
Wonderful!

nrb:
@ashish-amarnath @carlisia Now that #2674 is merged, lets take a look at this

carlisia:
@laverya I only had a couple minor requests and now you can safely rebase, the blocker is merged.

config/rbac/role.yaml
```diff
@@ -0,0 +1,28 @@
+
+---
```
laverya, line 2: This was newly added by the `update-generated-crd-code.sh` script after rebasing - I'm not sure quite what's up with it (the other yaml updates are all very clearly follow from the controller-gen update, but not this)

go.mod
```diff
@@ -9,40 +9,75 @@ require (
...
-	github.com/go-ini/ini v1.28.2 // indirect
 	github.com/gobwas/glob v0.2.3
 	github.com/gofrs/uuid v3.2.0+incompatible
 	github.com/golang/protobuf v1.4.2
 	github.com/hashicorp/go-hclog v0.0.0-20180709165350-ff2cf002a8dd
 	github.com/hashicorp/go-plugin v0.0.0-20190610192547-a1bc61569a26
-	github.com/jmespath/go-jmespath v0.0.0-20160202185014-0b12d6b521d8 // indirect
 	github.com/joho/godotenv v1.3.0
-	github.com/kubernetes-csi/external-snapshotter/v2 v2.1.0
+	github.com/kubernetes-csi/external-snapshotter/v2 v2.1.1-0.20200617021606-4800ca72d403
```
laverya, line 21: It may be worth including https://github.com/kubernetes-csi/external-snapshotter/releases/tag/v2.2.0-rc1 instead, now that that's been released

go.mod
```diff
@@ -9,40 +9,75 @@ require (
...
+	k8s.io/apimachinery v0.18.4
+	k8s.io/cli-runtime v0.18.4
+	k8s.io/client-go v0.18.4
 	k8s.io/klog v1.0.0
-	k8s.io/utils v0.0.0-20191218082557-f07c713de883 // indirect
-	sigs.k8s.io/controller-runtime v0.5.2
+	sigs.k8s.io/controller-runtime v0.6.1
 )
+
+replace k8s.io/api => k8s.io/api v0.18.4
```
laverya, line 43: I know this is ugly as hell - it's unfortunately a consequence of being a downstream importer of k8s.io/kubernetes, which imports all of these as v0.0.0 (via updated github.com/kubernetes-csi/external-snapshotter/v2)

laverya:
@nrb @ashish-amarnath @carlisia rebased and requested changes made
(plus a pair of test fixes)

config/rbac/role.yaml
carlisia: I can assert this is happening with the current master branch and am looking into it. I think we should ignore it for this PR, and if anything needs to be adjusted we'll do it separately.

go.mod
nrb: If we update to the v2.2.0-rc1 release, does that mitigate the need for this replace line?

changelogs/unreleased/2651-laverya
```diff
@@ -0,0 +1 @@
+bump Kubernetes module dependencies to v0.18.4 to fix https://github.com/vmware-tanzu/velero/issues/2540
```
nrb, line 1: Can you call out in here that it allows code compatibility with v1.18, so readers don't have to jump to the issue?

go.mod
```diff
@@ -9,40 +9,75 @@ require (
 	github.com/Azure/go-autorest/autorest/azure/auth v0.4.2
 	github.com/Azure/go-autorest/autorest/to v0.3.0
 	github.com/Azure/go-autorest/autorest/validation v0.2.0 // indirect
-	github.com/aws/aws-sdk-go v1.13.12
+	github.com/aws/aws-sdk-go v1.28.2
```
nrb, line 12: Was this added as a general version upgrade, or did it come with the upgrade to the v1.18 k8s libraries?

go.mod
laverya: unfortunately not: https://github.com/kubernetes-csi/external-snapshotter/blob/v2.2.0-rc1/go.mod#L26

nrb: Ok, worth a shot :) Thanks!

go.mod
nrb: I'm not sure about including an RC...curious what @ashish-amarnath thinks.

go.mod
laverya: It came from the k8s version upgrade: https://github.com/kubernetes/kubernetes/blob/v1.18.4/go.mod#L23

changelogs/unreleased/2651-laverya
laverya: updated

go.mod
ashish-amarnath: Yeah, I think it's better to avoid a pre-release version, esp if we are not depending on changes in the release.

laverya: It's less "use RC or normal version" and more "use the currently pinned commit version or the newer RC", because unfortunately it does have a change that this PR needs (upgrading to import k8s 1.18)

nrb: Got it. In that case, I'd prefer to use the RC.

shaneutt:
Is there a rough timeline estimate for when this will land in a release?

carlisia:
> Is there a rough timeline estimate for when this will land in a release?

Hey @shaneutt, this will be included in our 1.5 release, which will come out at some point in August.

shaneutt:
> > Is there a rough timeline estimate for when this will land in a release?
> 
> Hey @shaneutt, this will be included in our 1.5 release, which will come out at some point in August.

Thank you, I appreciate the update :+1:

PR #2659, opened 2020-06-22
Add RBAC page to table of contents

nrb:
The RBAC page is a little hard to find, so surface it in the ToC.


nrb:
@jonasrosland Thanks, updated!

PR #2667, opened 2020-06-24
📖  make external blogposts clickable links

ashish-amarnath:

### DIFF
--- a/site/resources.md
+++ b/site/resources.md
@@ -8,13 +8,17 @@ Here you will find external resources about Velero, such as videos, podcasts, an
 
 ## All community meetings
 
+* Community Meetings Playlist
 {% youtube videoseries?list=PL7bmigfV0EqQRysvqvqOtRNk4L5S7uqwM %}
+
+* Velero Demo and Deep Dives Playlist
+{% youtube videoseries?list=PL7bmigfV0EqT82eQaVCslWWM4k4d63KyK %}
 
 ## Podcast shows
 
-* vSpeaking Podcast Ep 124, with guests Carlisia Campos, Myles Gray and Cormac Hogan: https://blogs.vmware.com/virtualblocks/2019/08/02/vspeaking-podcast-velero/
+* [vSpeaking Podcast Ep 124, with guests Carlisia Campos, Myles Gray and Cormac Hogan](https://blogs.vmware.com/virtualblocks/2019/08/02/vspeaking-podcast-velero/)
 
-* Listen to the PodCTL podcast with Carlisia Campos talking about Velero: <http://podcast.podctl.com/110399/986641-understanding-project-velero-formerly-ark>
+* [Listen to the PodCTL podcast with Carlisia Campos talking about Velero](<http://podcast.podctl.com/110399/986641-understanding-project-velero-formerly-ark>)
 
 ## Videos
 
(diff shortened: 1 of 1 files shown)


site/resources.md
```diff
@@ -49,31 +49,21 @@ Here you will find external resources about Velero, such as videos, podcasts, an
...
-
-    https://blah.cloud/automation/using-velero-for-k8s-backup-and-restore-of-csi-volumes/
-
-* "Kubernetes Backup: Improve already awesome Velero with OpenEBS" - by Harshita Sharma
-
-    https://blog.mayadata.io/openebs/suggesting-ways-to-improve-already-awesome-velero
+* ["Kubernetes Backup: Improve already awesome Velero with OpenEBS" - by Harshita Sharma](https://blog.mayadata.io/openebs/suggesting-ways-to-improve-already-awesome-velero)
 
 * Two great blog posts by community member Imran Pochi:
+    * [Backup and Restore of Kubernetes Applications using Heptio’s Velero with Restic and Rook-Ceph as the storage provider](https://blog.kubernauts.io/backup-and-restore-of-kubernetes-applications-using-heptios-velero-with-restic-and-rook-ceph-as-2e8df15b1487)
```
a-mccarthy, line 63: Maybe add spaces in between the bullets here, so the list is formatted with the same double spacing throughout. Right now the Cormac Hogan link looks too close to the indented bullets

a-mccarthy:
@ashish-amarnath i added a in a comment, but the blogs links were all clickable :)

For the rest of the page, 
* the links in the podcast section aren't showing as links
* maybe add a link under the community meeting video thumbnail to the youtube playlist, so its easier for folks to get to all the videos? https://www.youtube.com/playlist?list=PL7bmigfV0EqQRysvqvqOtRNk4L5S7uqwM

I was thinking through how to possibly organize this page better. Some thoughts
* we should make sure that each list is organized so that the most recent content is on top. We could add dates into each bullet, to help users and also let us easily see if something is too old to keep on the page 
* Is there a way to know that this page is helping folks, like do we know if it gets a lot of engagement? To help promote visibility, we can see if we should add links to this pages from other places on the site, like the contributing page

cc: @jonasrosland

site/resources.md
ashish-amarnath: @a-mccarthy I see what you are saying. But That is just how that markdown is being rendered. Adding new lines didn't seem to solve the spacing 🙁

jonasrosland: I think adding two spaces on the new line might help here.

ashish-amarnath: I don't know if that'll help. But I've did it anyway. PTAL

ashish-amarnath: image

Doesn't seem like it helped

image

ashish-amarnath:
@a-mccarthy, @jonasrosland, @carlisia and @nrb PTAL.

site/resources.md
jonasrosland: I looked into it, it's because of some strangeness in how second layer lists are handled. If you can move that section down to the bottom so we can skip the weirdness for now, then I can dig into what needs to be changed later.

a-mccarthy: @ashish-amarnath have you tried adding spaces in between all the bullet points, even the second layer lists?

ashish-amarnath: @a-mccarthy adding new-lines between bullet points didn't work. Adding whitespace after `*` is not valid markdown and this is what happens when I do that
image

I  like @jonasrosland's idea of moving the weirdness to the bottom! That should def make this go away!

carlisia:
Sorry, @ashish-amarnath would you please add the other Velero playlist?

site/resources.md
ashish-amarnath: @a-mccarthy and @jonasrosland PTAL! https://deploy-preview-2667--velero.netlify.app/resources/

PR #2674, opened 2020-06-29
Add a BSL controller to handle validation + update BSL status phase

carlisia:
Closes: https://github.com/vmware-tanzu/velero/issues/1967.

Specs are on the ticket.

Note: I will add validation to BSL create/update in a separate PR.


pkg/controller/backupstoragelocation_controller.go
```diff
@@ -0,0 +1,147 @@
...
+
+		if location.Name == r.StorageLocation.DefaultStorageLocation {
+			defaultFound = true
+		}
+
+		ready := r.StorageLocation.IsReadyToValidate(&location, log)
+
+		if !ready {
+			continue
+		}
```
ashish-amarnath, line 71: can be simplified as
```
if !r.StorageLocation.IsReadyToValidate(&location, log) {
    continue
}
```

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -0,0 +1,147 @@
...
+		someVerified = true
+
+		log.Debug("Backup location ready to be verified")
```
ashish-amarnath, line 74: suggestion
```
		log.Debug("Validating BackupStorageLocation")
```

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -0,0 +1,147 @@
...
+		log.WithError(err).Error("No backup storage locations found, at least one is required")
+	}
+
+	if r.StorageLocation.DefaultStoreValidationFrequency <= 0 {
+		r.StorageLocation.DefaultStoreValidationFrequency = time.Minute
+	}
+
+	var defaultFound bool
+	var unavailableErrors []string
+	var someVerified bool
```
ashish-amarnath, line 59: suggest naming this as `anyVerified`.

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -0,0 +1,147 @@
...
+			log.Debug("Backup location verified and it is valid")
+			if err := r.StorageLocation.UpdatePhase(&location, velerov1api.BackupStorageLocationPhaseAvailable); err != nil {
+				log.WithError(err).Errorf("Error updating backup location phase to %s", velerov1api.BackupStorageLocationPhaseAvailable)
+				continue
+			}
+		}
+	}
+
+	if !someVerified {
+		log.Info("No backup locations were ready to be verified")
```
ashish-amarnath, line 98: I think it would be more useful if we print counts of validated, skipped and notReady BSLs instead.

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	r.logReconciledPhase(defaultFound, locationList, unavailableErrors)
+
+	return ctrl.Result{Requeue: true}, nil
+}
+
+func (r *BackupStorageLocationReconciler) logReconciledPhase(defaultFound bool, locationList velerov1api.BackupStorageLocationList, errs []string) {
```
ashish-amarnath, line 106: IMO, this is too much logic just to print a log message.
Suggest printing counts of BSLs in each state.
`BSL ready/validated/invalid : %d/%d/%d`

config/rbac/role.yaml
```diff
@@ -0,0 +1,28 @@
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: manager-role
```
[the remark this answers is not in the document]
carlisia: Yes, probably not. Will remove and test.

internal/velero/storagelocation.go
```diff
@@ -0,0 +1,131 @@
...
+func BackupStorageLocationsExist(kbClient client.Client, ctx context.Context, namespace string) (velerov1api.BackupStorageLocationList, error) {
+	var locations velerov1api.BackupStorageLocationList
+	if err := kbClient.List(ctx, &locations, &client.ListOptions{
+		Namespace: namespace,
+	}); err != nil {
+		return velerov1api.BackupStorageLocationList{}, err
+	}
+
+	if len(locations.Items) == 0 {
+		return velerov1api.BackupStorageLocationList{}, errors.New("no backup storage locations found")
```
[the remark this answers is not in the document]
carlisia: I don't recall discussing this before but in this context this is an error, ie, not having any BSL is a halting error.

internal/velero/storagelocation_test.go
```diff
@@ -0,0 +1,254 @@
...
+		},
+		{
+			name:                             "validate as per location setting, as that takes precedence, and always if it has never been validated before regardless of the frequency setting",
+			serverDefaultValidationFrequency: 0,
+			backupLocation:                   builder.ForBackupStorageLocation("ns-1", "location-1").ValidationFrequency(1 * time.Hour).Result(),
+			ready:                            true,
+		},
+		{
+			name:                             "don't validate as per location setting, as it is set to zero and that takes precedence",
+			serverDefaultValidationFrequency: 1,
```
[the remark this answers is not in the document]
carlisia: second

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -58,14 +58,14 @@ func NewCreateCommand(f client.Factory, use string) *cobra.Command {
...
-	Prefix           string
-	BackupSyncPeriod time.Duration
-	Config           flag.Map
-	Labels           flag.Map
-	AccessMode       *flag.Enum
+	Name                                  string
+	Provider                              string
+	Bucket                                string
+	Prefix                                string
+	BackupSyncPeriod, ValidationFrequency time.Duration
```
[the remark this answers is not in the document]
carlisia: Don't know if I agree. This already is in the context of a BSL, I don't think we need to add "BSL" as a prefix.

pkg/cmd/server/server.go
```diff
@@ -72,10 +71,10 @@ import (
 
 	ctrl "sigs.k8s.io/controller-runtime"
 
-	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/vmware-tanzu/velero/internal/util/managercontroller"
+	"github.com/vmware-tanzu/velero/internal/velero"
```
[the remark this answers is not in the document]
carlisia: Why?

internal/velero/storagelocation.go
```diff
@@ -0,0 +1,131 @@
...
+		if time.Now().UTC().Before(nextValidation) {              // ready only when NOW is equal to or after the next validation time
+			return false
+		}
+	}
+
+	return true
+}
+
+// IsValidFor verifies if a storage is valid for a given backup storage location.
+func (p *StorageLocation) IsValidFor(location *velerov1api.BackupStorageLocation, log logrus.FieldLogger) error {
```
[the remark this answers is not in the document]
carlisia: Changing it but I think `IsValidFor` reads a lot better since it is checking if the storage is valid for a given location.

internal/velero/storagelocation.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+
+	lastValidation := location.Status.LastValidationTime
+	if lastValidation != nil { // always ready to validate the first time around, so only even do this check if this has happened before
+		nextValidation := lastValidation.Add(validationFrequency) // next validation time: last validation time + validation frequency
+		if time.Now().UTC().Before(nextValidation) {              // ready only when NOW is equal to or after the next validation time
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
carlisia: Good catch! Also the one use case I didn't test for.

pkg/controller/backupstoragelocation_controller.go
carlisia: A tally is already being done inside `logReconciledPhase`.

pkg/cmd/cli/backuplocation/create.go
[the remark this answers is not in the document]
ashish-amarnath: I missed that this is in the context of bsl.

carlisia:
> Were you able to run this with multiple BSLs and also verify that the BSLs are correctly reconciled?

Yes. Scenarios (and shifting from one to the other):
- 0 BSLs
- storage down
- multiple, all valid
- multiple, all invalid
- only 1, invalid
- only 1, valid
- multiple, none of which were the specified default

internal/velero/storagelocation.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// use variables to refer to these functions so they can be
+	// replaced with fakes for testing.
+	NewPluginManager func(logrus.FieldLogger) clientmgmt.Manager
+	NewBackupStore   func(*velerov1api.BackupStorageLocation, persistence.ObjectStoreGetter, logrus.FieldLogger) (persistence.BackupStore, error)
+}
+
+// IsReadyToValidate calculates if a given backup storage location is ready to be validated.
+//
+// Rules:
```
nrb, line 52: Thanks for this list! Really appreciate it.

pkg/cmd/server/server.go
```diff
@@ -217,6 +218,7 @@ func NewCommand(f client.Factory) *cobra.Command {
 	command.Flags().StringSliceVar(&config.disabledControllers, "disable-controllers", config.disabledControllers, fmt.Sprintf("list of controllers to disable on startup. Valid values are %s", strings.Join(disableControllerList, ",")))
 	command.Flags().StringSliceVar(&config.restoreResourcePriorities, "restore-resource-priorities", config.restoreResourcePriorities, "desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources")
 	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "name of the default backup storage location")
+	command.Flags().DurationVar(&config.storeValidationFrequency, "store-validation-frequency", config.storeValidationFrequency, "how often to verify if the storage is valid. Optional. Set this to `0s` to disable sync")
```
nrb, line 221: Since we have the `defaultStoreValidationFrequency` var defined, we should probably document that value in the help message here.

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -0,0 +1,147 @@
+/*
```
nrb, line 1: I was thinking the controllers could go into an `internal/controller` package. Is that breaking some sort of controller-runtime pattern? My reasoning is that they're internal to Velero's operation and would never be used by Go API clients, only ever being used by submitting CRDs to the Kube API server.

If this is incompatible with controller-runtime practices though, feel free to ignore.

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -83,7 +83,8 @@ func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Provider, "provider", o.Provider, "name of the backup storage provider (e.g. aws, azure, gcp)")
 	flags.StringVar(&o.Bucket, "bucket", o.Bucket, "name of the object storage bucket where backups should be stored")
 	flags.StringVar(&o.Prefix, "prefix", o.Prefix, "prefix under which all Velero data should be stored within the bucket. Optional.")
-	flags.DurationVar(&o.BackupSyncPeriod, "backup-sync-period", o.BackupSyncPeriod, "how often to ensure all Velero backups in object storage exist as Backup API objects in the cluster. Optional. Set this to `0s` to disable sync")
+	flags.DurationVar(&o.BackupSyncPeriod, "backup-sync-period", o.BackupSyncPeriod, "how often to ensure all Velero backups in object storage exist as Backup API objects in the cluster. Optional. Set this to `0s` to disable sync. Default: 1 minute.")
+	flags.DurationVar(&o.ValidationFrequency, "validation-frequency", o.ValidationFrequency, "how often to verify if the backup storage location is valid. Optional. Set this to `0s` to disable sync")
```
nrb, line 87: Since https://github.com/vmware-tanzu/velero/pull/2674/files#diff-cfcbc9b9778d96e266dd99924a91c426R54 sets the default to 1 minute, I think we should document the default in the help message.

site/docs/master/api-types/backupstoragelocation.md
```diff
@@ -40,6 +40,7 @@ The configurable parameters are as follows:
 | `config` | map[string]string | None (Optional) | Provider-specific configuration keys/values to be passed to the object store plugin. See [your object storage provider's plugin documentation][0] for details. |
 | `accessMode` | String | `ReadWrite` | How Velero can access the backup storage location. Valid values are `ReadWrite`, `ReadOnly`. |
 | `backupSyncPeriod` | metav1.Duration | Optional Field | How frequently Velero should synchronize backups in object storage. Default is Velero's server backup sync period. Set this to `0s` to disable sync. |
+| `validationFrequency` | metav1.Duration | Optional Field | How frequently Velero should validate the object storage . Default is Velero's server validation frequency. Set this to `0s` to disable validation. |
```
nrb, line 43: Similar to the help message, I think the default should be mentioned here.

pkg/controller/backupstoragelocation_controller.go
carlisia: This is very subjective and depending, I think, on what we want to accomplish. If we want to accomplish consistency with other kubebuilder generated projects, then I'd suggest leaving it here. In fact, kubebuilder generates all controllers under a directory `controllers` (plural), and at the root, so once we convert all controllers we should move them to that structure.

carlisia: But other than consistency there's no requirement to have the controllers be public. I'm not sure how we should optimize this, I'm tending to go with consistency (with other kubebuilder projs).

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -0,0 +1,147 @@
...
+// +kubebuilder:rbac:groups=velero.io,resources=backupstoragelocations,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=velero.io,resources=backupstoragelocations/status,verbs=get;update;patch
+func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithField("controller", "backupstoragelocation")
+
+	log.Info("Checking for existing backup locations ready to be verified; there needs to be at least 1 backup location available")
+
+	locationList, err := velero.ListBackupStorageLocations(r.StorageLocation.Client, r.StorageLocation.Ctx, req.Namespace)
+	if err != nil {
+		log.WithError(err).Error("No backup storage locations found, at least one is required")
```
ashish-amarnath, line 50: Should we just return the error?
Currently it doesn't matter because we loop over the `locationList.Items` and when there is an error that list is empty. But, returning the error makes the current code easier to read and also any logic we add in the future after line 51 can assume that there are items to reconcile and not run into seg faults or perform nil checks.

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}
+
+	if !anyVerified {
+		log.Info("No backup locations were ready to be verified")
+	}
+
+	r.logReconciledPhase(defaultFound, locationList, unavailableErrors)
+
+	return ctrl.Result{Requeue: true}, nil
```
ashish-amarnath, line 103: `Requeue` makes the controller requeue the request, which we want to do only if we run into errors reconciling the request and we want to retry.
In a success scenario, we want to return `ctrl.Result{}, nil`
https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.5.2/pkg/reconcile?tab=doc#Result

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -0,0 +1,147 @@
...
+			log.Errorf("Current backup storage locations available/unavailable/unknown: %v/%v/%v, %s)", numAvailable, numUnavailable, numUnknown, strings.Join(errs, "; "))
+		} else {
+			log.Errorf("Current backup storage locations available/unavailable/unknown: %v/%v/%v)", numAvailable, numUnavailable, numUnknown)
+		}
+	} else if numUnavailable > 0 { // some but not all BSL unavailable
+		log.Warnf("Invalid backup locations detected: available/unavailable/unknown: %v/%v/%v, %s)", numAvailable, numUnavailable, numUnknown, strings.Join(errs, "; "))
+	}
+
+	if !defaultFound {
+		log.Warnf("The specified default backup location named %q was not found; for convenience, be sure to create one or make another backup location that is available the default", r.StorageLocation.DefaultStorageLocation)
```
ashish-amarnath, line 139: This log message doesn't seem to read correctly.
Are we asking users to create a backup location named `default` and ensure that it's available?

site/docs/master/api-types/backupstoragelocation.md
ashish-amarnath: To disable, should this be `0s` or can it be just `0`?

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -0,0 +1,147 @@
...
+	log := r.Log.WithField("controller", "backupstoragelocation")
+
+	log.Info("Checking for existing backup locations ready to be verified; there needs to be at least 1 backup location available")
+
+	locationList, err := velero.ListBackupStorageLocations(r.StorageLocation.Client, r.StorageLocation.Ctx, req.Namespace)
+	if err != nil {
+		log.WithError(err).Error("No backup storage locations found, at least one is required")
+	}
+
+	if r.StorageLocation.DefaultStoreValidationFrequency <= 0 {
```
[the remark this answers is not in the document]
nrb: Ah, yeah, this should probably just look for `< 0`

site/docs/master/api-types/backupstoragelocation.md
nrb: I _think_ `time.Duration` values need a unit at the end, so `0s`.

pkg/cmd/server/server.go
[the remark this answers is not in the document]
carlisia: Unless I'm misunderstanding, this would be true but only for this file given the scope of an alias. Right? I'm not sure this is worth it. Maybe I'm misunderstanding. And on another hand and by this logic: are you proposing we alias every package that goes under `internal`?

pkg/controller/backupstoragelocation_controller.go
nrb: Ok, that makes sense.

I was envisioning that eventually we moved all the controllers into `internal/controller(s?)`. However, that's not a hard requirement. If it's convention to keep them public, then I'm not going to push the issue. They're not generally useful to import, in any case.

pkg/controller/backupstoragelocation_controller.go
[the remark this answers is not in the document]
carlisia: Good catch. And no validation should be in the controller anyway, removing.

In the business logic there's already:

```
	if validationFrequency == 0 {
		log.Debug("Validation period for this backup location is set to 0, skipping validation")
		return false
	
```

pkg/controller/backupstoragelocation_controller.go
carlisia: Actually no, in this case we do want to requeue continuously. We have an external dependency (storage) that can change state at any time and this is what we are continuously reconciling.

pkg/controller/backupstoragelocation_controller.go
carlisia: This is a warning that there's no bsl enabled as default. The warning suggests to create one, or change another bsl to be the default.

pkg/controller/backupstoragelocation_controller.go
carlisia: I was trying to avoid having to duplicate `r.logReconciledPhase(defaultFound, locationList, unavailableErrors)` here as well or, worse, the inconsistency of having that call when there are items but not when there are none.

site/docs/master/api-types/backupstoragelocation.md
carlisia: Same as `backupSyncPeriod` right above.

pkg/cmd/server/server.go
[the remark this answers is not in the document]
ashish-amarnath: Considering the strong conventions, imposed by the go tooling, I am OK with not aliasing the imports.
https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface

pkg/controller/backupstoragelocation_controller.go
nrb: If we return the error, how does that affect the reconciliation loop? Does it get re-invoked?

pkg/controller/backupstoragelocation_controller.go
nrb: Velero can operate without a default BSL specified if individual backups have one defined on them; it's mostly a convenience feature, as the message says.

[`StorageLocation`](https://github.com/vmware-tanzu/velero/blob/master/pkg/apis/velero/v1/backup.go#L80) is the field name on the `Backup` type, and `--storage-location` is the option on `velero backup create`. If the field is empty, then the default location is used, provided there is one.

pkg/controller/backupstoragelocation_controller.go
carlisia: Yes, it can make it get requeued.

internal/velero/storagelocation.go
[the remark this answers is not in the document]
nrb: I know the original content of #1967 was requesting that Velero not halt when there were no BSLs. That's because the reporter's scenario was that they created BSLs on-demand; they had a higher-order system that received a backup request, created a BSL, then issued the backup to Velero.

As it stands now (without this PR merged), if you have a BSL and get the Velero server running, then delete the BSL, the Velero server does not halt, because the check for BSLs being present only happened on startup.

#2382 tried to unify the behavior into a liveness check that relied on Kubernetes to restart Velero's server until a BSL was created, though at this point I don't think that's a good solution to the problem.

I think having Velero stay up when there are no BSLs found, and retry until there are, makes it more resilient when it is used as part of a larger system. However, if we can error here and cause the reconciliation loop to keep retrying, then I think we accomplish the goal and keep the Velero server up.

pkg/controller/backupstoragelocation_controller.go
ashish-amarnath: That sentence wasn't parsing correctly in my head. After the clarification it makes sense.

internal/velero/storagelocation.go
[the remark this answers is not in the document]
nrb: Re-reading the code, I think the usage makes sense.

It's an error from this function, but the error is viewed, logged, and then moved past in the controller.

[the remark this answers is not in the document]
carlisia: Yes, and we had discussed this behavior change as a group before, that's why I went ahead and implemented it. Also, there is a case where users pair up 1 backup with 1 bsl, and they want to create that bsl in a scripted way at backup time, and having the server not boot up because it failed validation was an issue.

The controller (in this PR) will continuously be reconciling, and will log an error whenever the system is in a state where there are no BSL or there are BSLs but they are all invalid.

pkg/controller/backupstoragelocation_controller.go
nrb: Ok, I think the way this is working now I'm good with it. We can refine things over the course of the release cycle, like potentially adding a back off period via [`Result.RequeueAfter`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.5.2/pkg/reconcile?tab=doc#Result) if none are found, but let's do that in another PR if we really need it.

PR #2696, opened 2020-07-07
Adjust restic timeout and pod values up

nrb:
@ashish-amarnath You had tested restic loads, and stated that 500m/512Mi was what allowed the workload to finish in a timely fashion for 100GB.

However, I noticed that in https://github.com/vmware-tanzu/velero/blob/master/pkg/util/kube/resource_requirements.go#L28, we treat a `"0"` value as unbounded, so I'm wondering - wouldn't a 500m/512Mi set have been a _reduced_ limit? Or do we need to set a request of 500m/512Mi, and leave the limit unbounded?

### DIFF
--- a/pkg/install/resources.go
+++ b/pkg/install/resources.go
@@ -47,10 +47,10 @@ var (
 	DefaultVeleroPodMemRequest = "128Mi"
 	DefaultVeleroPodCPULimit   = "1000m"
 	DefaultVeleroPodMemLimit   = "256Mi"
-	DefaultResticPodCPURequest = "0"
-	DefaultResticPodMemRequest = "0"
-	DefaultResticPodCPULimit   = "0"
-	DefaultResticPodMemLimit   = "0"
+	DefaultResticPodCPURequest = "500m"
+	DefaultResticPodMemRequest = "512Mi"
+	DefaultResticPodCPULimit   = "100Mi"
+	DefaultResticPodMemLimit   = "1024Mi"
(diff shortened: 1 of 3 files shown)


ashish-amarnath:
Operating in best-effort limits work well on a cluster with light load. However, in a busy cluster lightly loaded cluster, best-effort can mean that the restic pod does not get enough CPU or memory. The cluster I observed this is was not operating under light load.

pkg/install/resources.go
```diff
@@ -47,10 +47,10 @@ var (
 	DefaultVeleroPodMemRequest = "128Mi"
 	DefaultVeleroPodCPULimit   = "1000m"
 	DefaultVeleroPodMemLimit   = "256Mi"
-	DefaultResticPodCPURequest = "0"
-	DefaultResticPodMemRequest = "0"
-	DefaultResticPodCPULimit   = "0"
-	DefaultResticPodMemLimit   = "0"
+	DefaultResticPodCPURequest = "500m"
+	DefaultResticPodMemRequest = "512Mi"
+	DefaultResticPodCPULimit   = "100Mi"
```
ashish-amarnath, line 52: did you mean `1000m`?

nrb: Yep!

nrb:
@ashish-amarnath Yeah, tested against 1.17 and had to adjust some of the values as they weren't valid :)

Issue #2678, opened 2020-06-30
Restore Hooks product requirements document


stephbman:
The markdown file containing the product requirements document for the restore hooks feature epic is included here.

ashish-amarnath:
Now that https://github.com/vmware-tanzu/velero/pull/2699 is merged. This issue can be closed.

PR #2699, opened 2020-07-08
Create restore-hooks_product-requirements.md

stephbman:
This file contains the MVP use cases and product requirements tied the Restore Hooks Design Proposal https://github.com/vmware-tanzu/velero/pull/2465.

This requirements document ties to Epic: https://github.com/vmware-tanzu/velero/issues/2116
And is related to issue:  https://github.com/vmware-tanzu/velero/issues/2678

Signed-off-by: Stephanie Bauman, bstephanie@vmware.com

### DIFF
--- a/restore-hooks_product-requirements.md
+++ b/restore-hooks_product-requirements.md
@@ -1,0 +1,309 @@
+Velero Restore Hooks - PRD (Product Requirements Document) 
+
+MVP Feature Set
+
+Relates to:  [https://github.com/vmware-tanzu/velero/issues/2116](https://github.com/vmware-tanzu/velero/issues/2116)
+
+
+# **Change tracking**
+
+_This is a live document, you can reach me on the following channels for more information or for any questions:_
+
+
+
+*   _Email:  [bstephanie@vmware.com](mailto:bstephanie@vmware.com)_
+*   _Kubernetes slack:  _
+    *   _Channel:  #velero_
+    *   _@Stephanie Bauman_
+*   _VMware internal slack:_
+    *   _@bstephanie_
+
+
+# **Relates to Git Issues:**
+
+
+
+*   [https://app.zenhub.com/workspaces/velero-5c59c15e39d47b774b5864e3/issues/vmware-tanzu/velero/2465](https://app.zenhub.com/workspaces/velero-5c59c15e39d47b774b5864e3/issues/vmware-tanzu/velero/2465)
+    *   [https://github.com/vmware-tanzu/velero/pull/2465/files?short_path=140e0c6#diff-140e0c6b370f250ee97f6ecafc8dbb7a](https://github.com/vmware-tanzu/velero/pull/2465/files?short_path=140e0c6#diff-140e0c6b370f250ee97f6ecafc8dbb7a)
+*   [https://github.com/vmware-tanzu/velero/issues/1150](https://github.com/vmware-tanzu/velero/issues/1150)
+
+Background
+
+Velero supports restore operations but there are gaps in the process.  Gaps in the restore process require users to manually carry out steps to start, clean up, and end the restore process. Other gaps in the restore process can cause issues with application performance for applications running in a pod when a restore operation is carried out. 
+
+On a restore, Velero currently does not include hooks to execute a pre- or -post restore script. As a result, users are required to perform additional actions following a velero restore operation. Some gaps that currently exist in the Velero restore process are:
+
+
+
+*   Users can create a restore operation but has no option to customize or automate commands during the start of the restore operation
+*   Users can perform post restore operations but have no option to customize or automate commands during the end of the restore operation
+
+Strategic Fit
+
+Adding a restore hook action today would allow Velero to unpack the data that was backed up in an automated way by enabling Velero to execute commands in containers during a restore. This will also improve the restore operations on a container and mitigate against any negative performance impacts of apps running in the container during restore.
+
+Purpose / Goal
+
+The purpose of this feature is to improve the extensibility and user experience of pre and post restore operations for Velero users.
+
+Goals for this feature include:
+
+
+
+*   Enhance application performance during and following restore events 
+*   Provide pre-restore hooks for customizing start restore operations
+*   Provide actions for things like retry actions, event logging, etc... during restore operations 
+*   Provide observability/status of restore commands run in restored pods
+*   Extend restore logs to include status, error codes and necessary metadata for restore commands run during restore operations for enhanced troubleshooting capabilities 
+*   Provide post-restore hooks for customizing end of restore operations
+*   Include pre-populated data that has been serialized by a Velero backup hook into a container or external service prior to allowing a restore to be marked as completed. 
+
+Non-goals
+
+Feature Description
+
+This feature will automate the restore operations/processes in Velero, and will provide restore hook actions in Velero that allows users to execute restore commands on a container. Restore hooks include pre-hook actions and post-hook actions.
+
+This feature will mirror the actions of a Velero backup by allowing Velero to check for any restore hooks specified for a pod. 
+
+Assumptions
+
+
+
+*   Restore operations will be run at the pod level instead of at the volume level. Some databases require the pod to be running and in some cases a user cannot manipulate a volume without the pod running. We need to support more than one type of database for this feature and so we need to ensure that this works broadly as opposed to providing support only for specific dbs. 
+*   Velero will be responsible for invoking the restore hook. 
+
+ 
+
+MVP Use Cases 
+
+The following use cases must be included as part of the Velero restore hooks MVP (minimum viable product). 
+
+**Note: **Processing of concurrent vs sequential workloads is slated later in the Velero roadmap (see [https://github.com/vmware-tanzu/velero/pull/2548/files](https://github.com/vmware-tanzu/velero/pull/2548/files)). The MVP for this feature set will align with restore of single workloads vs concurrent workload restores. A second epic will be created to address the concurrent restore operations and will be added to the backlog for priority visibility. 
+
+**Note: **Please refer to the Requirements section of this document for more details on what items are P0 (must have in the MVP), P1 (should not ship without for the MVP), P2 (nice to haves).
+
+**<span style="text-decoration:underline;">USE CASE 1</span>**
+
+
+    **Title:** Run restore hook before pod restart.
+
+
+    **Description:  **As a user, I would like to run a restore hook before the applications in my pod are restarted.
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 2</span>**
+
+
+    **Title: **Allow restore hook to run on non-kubernetes databases
+
+
+    **Description: **As a user, I would like to run restore hook operations even on databases that are external to kubernetes (such as postgres, elastic, etc…).
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 3</span>**
+
+
+    **Title: **Run restore at pod level.
+
+
+    **Description: **As a user, I would like to make sure that I can run the restore hook at the pod level. And, I would like the option to run this with an annotation flag in line or using an init container.**<span style="text-decoration:underline;"> </span>**The restore pre-hook should allow the user to run the command on the container where the pre-hook should be executed. Similar to the backup hooks, this hook should run to default to run on the first container in the pod. 
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 4</span>**
+
+
+    **Title:** Specify container in which to run pre-hook
+
+
+    **Description: **As a user, if I do not want to run the pre-hook command on the first container in the pod (default container), I would like the option to annotate the specific container that i would like the hook to run in.
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 5</span>**
+
+
+    **Title: **Check for latest snapshot 
+
+
+    **Description: **As a user, I would like Velero to run a check for the latest snapshot in object storage prior to starting restore operations on a pod. 
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 6</span>**
+
+
+    **Title: **Display/surface output from restore hooks/restore status 
+
+
+    **Description:** As a user, I would like to see the output of the restore hooks/status of my restore surfaced from the pod volume restore status. Including statuses:  Pending, Running/In Progress, Succeeded, Failed, Unknown.
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 7</span>**
+
+
+    **Title: **Restore metadata
+
+
+    **Description: **As a user, I would like to have the metadata of the contents of what was restored using Velero.
+
+
+    _Note: Kubernetes patterns may result in some snapshot metadata being overwritten during restore operations._
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 8</span>*
+
+    **Title: **Increase default restore and retry limits. 
+
+
+    **Description: **As a user, I would like to increase the default restore retry and timeout limits from the default values to some other value I would like to specify. 
+
+
+    _Note: See use case 11 for the default value specifications. _
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+User Experience
+
+The following is representative of what the user experience could look like for Velero restore pre-hooks and post-hooks. 
+
+**_Note: These examples are representative and are not to be considered for use in pre- and post- restore hook operations until the technical design is complete._**
+
+**Restore Pre-Hooks**
+
+<span style="text-decoration:underline;">Container Command</span>
+
+
+```
+pre.hook.restore.velero.io/container
+
+kubectl patch backupstoragelocation <STORAGE LOCATION NAME> \
+    --namespace velero \
+    --type merge \
+    --patch '{"spec":{"accessMode":"ReadOnly"}}'
+```
+
+
+<span style="text-decoration:underline;">Command Execute</span>
+
+
+```
+pre.hook.restore.velero.io/command
+```
+
+
+Includes commands for:
+
+
+
+*   Create
+    *   Create from most recent backup
+    *   Create from specific backup - allow user to list backups
+*   Set backup storage location to read only 
+
+    ```
+    kubectl patch backupstoragelocation <STORAGE LOCATION NAME> \
+        --namespace velero \
+        --type merge \
+        --patch '{"spec":{"accessMode":"ReadOnly"}}
+    ```
+
+
+*   Set backup storage location to read-write
+
+    ```
+    kubectl patch backupstoragelocation <STORAGE LOCATION NAME> \
+        --namespace velero \
+        --type merge \
+        --patch '{"spec":{"accessMode":"ReadWrite"}}'
+
+    ```
+
+
+<span style="text-decoration:underline;">Error handling </span>
+
+
+```
+pre.hook.restore.velero.io/on-error
+```
+
+
+<span style="text-decoration:underline;">Timeout </span>
+
+
+```
+pre.hook.restore.velero.io/retry
+```
+
+
+Requirements 
+
+**_P0_** = must not ship without 
+
+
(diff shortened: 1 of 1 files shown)


restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,376 @@
...
+Goals for this feature include:
+
+
+
+*   Enhance application performance during and following restore events 
+*   Provide pre-restore hooks for customizing start restore operations
+*   Provide actions for things like retry actions, event logging, etc... during restore operations 
+*   Provide observability/status of restore commands run in restored pods
+*   Extend restore logs to include status, error codes and necessary metadata for restore commands run during restore operations for enhanced troubleshooting capabilities 
+*   Provide post-restore hooks for customizing end of restore operations
```
nrb, line 58: I would also add prep-populating data that's been serialized by a Velero [backup hook](https://velero.io/docs/v1.4/hooks/#backup-hooks) into a container/external service before allowing a restore to be marked as completed.

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,376 @@
...
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 6</span>**
+
+
+    **Title: **Specify backup snapshot to use for restore
+
+
+    **Description: **As a user, I would like the option to specify a specific backup snapshot for use by Velero during restore create. I would like to do this instead of using the default most recent backup restore snapshot.
```
nrb, line 141: Can you expand on this use case a little more? This sounds like how `velero create restore --from-backup <backup name>` works already today, but I may be missing something.

`velero create restore --from-schedule` will create a restore from the latest backup associated with a given schedule, but the `--from-backup` version allows a user to choose a specific backup, so long as it still exists.

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,376 @@
...
+**<span style="text-decoration:underline;">USE CASE 7</span>**
+
+
+    **Title: **Include or exclude namespaces from restore
+
+
+    **Description: **As a user, I would like to include all resources in namespaces contained in a backup  by default in my restore spec. 
+
+
+    If velero is asked to restore something that already exists in a pod, the restore will not return a success but will still work - Ashish needs to verify.
```
nrb, line 154: Is this meant to extend to the restore hooks themselves? So if the restore hook tries to restore a directory that is already populated within a pod, are we saying that the hook should let Velero know that it happened, but not fail the whole backup?

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,376 @@
...
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 8</span>**
+
+
+    **Title: **Restore of a stateful application (Unquiescing data from a quiesced backup)
+
+
+    **Description: **As a user, I would like to prepare a backup to be restored without having to freeze an in-use volume/volume that an application is using.
```
nrb, line 164: The first sentence seems tied to backup hooks; is that considered in scope for this work in v1.5?

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,376 @@
...
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 10</span>**
+
+
+    **Title: **Restore metadata
+
+
+    **Description: **As a user, I would like to have the metadata of the contents of what was restored using Velero.
```
nrb, line 190: Are the contents referred to here the contents of what the hook restored? For example, this might be database files it reloaded, copied, or otherwise manipulated.

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,376 @@
...
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 11</span>**
+
+
+    **Title: **Retry restore upon restore failure/error/timeout
+
+
+    **Description: **As a user, if I see that a restore has failed, I would like for Velero to retry the restore operations using the restore specification.
```
nrb, line 203: This sounds like a restore-level goal, and I think may be out of scope for the restore hooks. We can specify retry attempt rules per-hook, but I think adding retry attempts for the whole of the restore process is larger than that.

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,376 @@
...
+2. Can velero allow a user to specify a specific backup if the most recent backup is not desired in a restore?
+3. If a backup specified for a restore operation fails, can velero retry and pick up the next most recent backup in the restore?
+4. Can velero provide a delta between the two backups if a different backup needs to be picked up (other than the most recent because the most recent backup cannot be accessed?)
+5. What types of errors can velero surface about backups, namespaces, pods, resources, if a backup has an issue with it preventing a restore from being done? 
+
+ 
+
+For questions, please contact michaelmi@vmware.com,  [bstephanie@vmware.com](mailto:bstephanie@vmware.com) 
+
+Signed-off-by: Stephanie Bauman bstephanie@vmware.com
```
nrb, line 376: This line can be removed - it only needs to be on the git commit message :)

restore-hooks_product-requirements.md
yaronp68: this doesn't sound like a use case for restore hooks but rather for restore create

restore-hooks_product-requirements.md
yaronp68: Not sure I understand how this relates to restore hooks. I thought restore hooks lets the user do something after the pod had been restored but not repeat a restore action

restore-hooks_product-requirements.md
stephbman: The intent of this was that if a user preferred to specify a snapshot they would like to use other than the default 'latest' snapshot that was taken from the velero create restore --from-schedule, that the user could instead specify a snapshot they would like to use instead.

This is most applicable in security use cases or corruption use cases where, the data contained in the latest snapshot may have been corrupt. Thoughts on this @nrb @yaronp68 ?

restore-hooks_product-requirements.md
stephbman: I am actually going to remove this use case from the PRD. I see in the backup hooks API type documentation, that a user can specify the array of resources to include or exclude from a backup. The intent of this requirement was to address a use case of that nature. 

Because that is already addressed as a capability of the existing Velero backup procedures and is something that would occur prior to restore operations. As such, removing this makes sense and does not confuse the restore hooks actions. Thoughts:  @nrb @yaronp68? (reference:  https://github.com/vmware-tanzu/velero/edit/stephbman-patch-1/restore-hooks_product-requirements.md?pr=%2Fvmware-tanzu%2Fvelero%2Fpull%2F2699).

Thank you for the feedback as well!

restore-hooks_product-requirements.md
stephbman: @nrb the scope of this document is for restore only and not backup. However, restore operations rely on some completed aspects of back up. Am I correct in assuming that Velero already quiesces data in a backup? And if that is the case, then can the restore operation here unquiesce the data from a quiesced backup? If that is not the case, then data unquiescing during restore operations may carry a dependency that needs to be documented, and tied to and worked on for backup operations so that we can achieve this use case as part of a restore.

restore-hooks_product-requirements.md
stephbman: @nrb yes that's the intent of this use case. Do you feel anything else needs to be included here to make this more clear?

nrb: I don't think so; I will have to think about how we will have to provide a place for hooks to report what they did, though. Right now, there's no standard place for hooks, either backup or restore, to place data into Velero.

restore-hooks_product-requirements.md
stephbman: @nrb good call out, thank you. What I'll do is remove this from the restore hooks document the use case for later for potential future enhancements to core restore velero operations.

restore-hooks_product-requirements.md
stephbman: Added. Thanks for the suggestion @nrb !

restore-hooks_product-requirements.md
stephbman: Removed and saved for potential future core restore operations improvements.

restore-hooks_product-requirements.md
nrb: >Am I correct in assuming that Velero already quiesces data in a backup?

Velero relies on backup hooks to quiesce application data if it's more than "simple" volume data.

Ideally, applications would include some form of utility that can quiesce data within them that doesn't require freezing or locking.  As an example, `pg_dump` for the PostgreSQL database is an included utility that queiseces the data from the running database without interrupting read/write operations of the live system. Thus, a user could write a hook relying on `pg_dump` as a backup hook, and use the resulting output with `pg_restore` in a restore hook.

If such a utility isn't available, applications would be put into a 'read only' mode during the backup, then allowing writes as soon as the backup is done. Both of these actions are something that the backup hooks would be doing. It would mean interruptions to writes, but that time period should (hopefully) be short. Still, the application would be usable in a read mode.


In the worst case, I think Velero often runs into the stated probelm with the restic support, though, due to restic requiring point-in-time snapshots of the file system that don't change while restic is running. We recommend using fsfreeze out of caution, since the data will be consistent when this is done. However, I absolutely agree that the above "read-only" mode would be preferable in a restic scenario as well as volume snapshots.

Some combination of volume snapshots and restic could provide better support for this, but that is a much larger task that is definitely out of scope. As an example, the idea might be to take a volume snaspshot of a volume, mount a copy of that, then run restic on the volume from that copy. But that's also not how we use restic today - restic's an alternative to volume snapshot plugins, and doesn't work in conjunction with them.

I think maybe the use case is accurate, but it is tough for Velero itself to hit without participation from the applications and hooks themselves.

restore-hooks_product-requirements.md
nrb: Yeah, I think removing it makes sense.

restore-hooks_product-requirements.md
nrb: I think that's a valid use case, but I do believe `velero create restore --from-backup` meets it already, as `--from-backup` allows the user to select any valid Velero backup that exist.

Backups associated with a Schedule all exist in the Kubernetes API server with a label on them, and the `--from-schedule` option grabs all Backups labeled with the Schedule's name, sorts them by date created, then grabs the latest one. So Backups created from a Schedule aren't particularly special other than the label once created.

Would you agree that this meets the use case?

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,337 @@
...
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 5</span>**
+
+
+    **Title: **Check for latest snapshot 
+
+
+    **Description: **As a user, I would like Velero to run a check for the latest snapshot in object storage prior to starting restore operations on a pod.
```
ashish-amarnath, line 132: There is a 1:1 mapping between restores and backups and backups have volumesnapshots in them.

On restoring from a backup, the volumesnapshots from that backup are what are used to restore data into the volumes.

IMO, changing that behavior, to choose a volumesnapshot, different from the one in the backup, is out of scope for this feature.

restore-hooks_product-requirements.md
ashish-amarnath: From the discussion during the community call on 7/21, this use-case is referring to a live update of data used by an application by restoring the data from a previous backup. This, we agreed, is out of scope for the restore hooks per-se.

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,337 @@
...
+
+
+    _Note: Some dbs have built in way to back up _
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 8</span>**
+
+
+    **Title: **Display/surface restore status
```
ashish-amarnath, line 165: To make this more relevant to the restore-hooks, this should read
`Display/surface output from restore hooks` with a corresponding update to the below description

restore-hooks_product-requirements.md
ashish-amarnath: Agree on this. And I think usecase 8, above, covered/should cover reporting status of the restore hook.
One way to surface the output of a restore hook would be to add that data as an annotation to the pod.

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,337 @@
...
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+User Experience
+
+The following is representative of what the user experience could look like for Velero restore pre-hooks and post-hooks. 
+
+**_Note: These examples are representative and are not to be considered for us in pre- and post- restore hook operations until the technical design is complete._**
+
+**Restore Pre-Hooks**
```
ashish-amarnath, line 203: Same question as before about pre and post qualifiers for restore hooks.

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,337 @@
...
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 3</span>**
+
+
+    **Title: **Run restore at pod level.
+
+
+    **Description: **As a user, I would like to make sure that I can run the restore hook at the pod level. And, I would like the option to run this with an annotation flag in line or using an init container.**<span style="text-decoration:underline;"> </span>**The restore pre-hook should allow the user to run the command on the container where the pre-hook should be executed. Similar to the backup hooks, this hook should run to default to fun on the first container in the pod.
```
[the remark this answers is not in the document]
ashish-amarnath: Also, is this use case to allow users to choose the container in which the hook command is to be run in?

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,337 @@
...
+*   Users can create a restore operation but has no option to customize or automate commands during the start of the restore operation
+*   Users can perform post restore operations but have no option to customize or automate commands during the end of the restore operation
+
+Strategic Fit
+
+Adding a restore hook action today would allow Velero to unpack the data that was backed up in an automated way by enabling Velero to execute commands in containers during a restore. This will also improve the restore operations on a container and mitigate against any negative performance impacts of apps running in the container during restore.
+
+Purpose / Goal
+
+The purpose of this feature is to improve the extensibility and user experience of pre and post restore operations for Velero users.
```
ashish-amarnath, line 47: Is there a notion of pre and post for restore hooks? Before what would we run a pre hook and after what would post hook?

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,337 @@
...
+
+    _Note: See use case 11 for the default value specifications. _
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+User Experience
+
+The following is representative of what the user experience could look like for Velero restore pre-hooks and post-hooks. 
+
+**_Note: These examples are representative and are not to be considered for us in pre- and post- restore hook operations until the technical design is complete._**
```
ashish-amarnath, line 201: typo? `not to be considered for use`

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,337 @@
...
+
+**Restore Pre-Hooks**
+
+<span style="text-decoration:underline;">Container Command</span>
```
ashish-amarnath, line 205: It isn't clear to me what this section is for. Can you please clarify?

restore-hooks_product-requirements.md
stephbman: The use cases detail the requirements. This level of detail won't be found in the strategic fit/goals section.

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,337 @@
...
+
+**<span style="text-decoration:underline;">______________________________________________________________</span>**
+
+**<span style="text-decoration:underline;">USE CASE 2</span>**
+
+
+    **Title: **Allow restore hook to run on non-kubernetes databases
+
+
+    **Description: **As a user, I would like to run restore hook operations even on databases that are external to kubernetes (such as postgres, elastic, etc…).
```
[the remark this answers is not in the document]
stephbman: Thanks for the comment here Ashish. I think it is important to identify the different types of interfaces that need to support this use case. 

However, I do not think that needs to be included in the product use cases but instead in the design document. And then carried out in implementation.

restore-hooks_product-requirements.md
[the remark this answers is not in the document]
stephbman: The concept of pre- and post- hooks is called out in the use case. The design document should specify the 'before what' the pre-hook should be run against and the 'after what' the post hook should be run against.

Generally the pre-hook should be run for container, command, on-error, timeout. The post hook should be run for container, command, on-error, timeout. 

Specifications of the handling that action are not included as part of the product use case document because they should be included as part of the design document. 

The use case requires that the hook default to the run on the first container in the pod. The user should have the option to annotate/specify a different container if they do not wish to default to the first container in the pod. See the next use case for that requirement.

restore-hooks_product-requirements.md
stephbman: This use case is not changing that behavior. Rather, the use case specifies that as part of restore requirements for the product, we want the restore to use the latest volumesnapshot available (latest meaning the volume snapshot that has the most recent timestamp). 

On restoring from backup, we want to make sure that our restore requirements leverage the actions and the 1:1 mapping between restores and backups and that the behavior is not changed.

restore-hooks_product-requirements.md
stephbman: The design document should include the specifications for the pre and post restore hooks. Generally, as mentioned in the previous reply I provided, these should be at the container, command, on-error, timeout

restore-hooks_product-requirements.md
stephbman: Product requirements attempt to provide examples of what the user experience in the command line might look like. Similar to a wireframe for a UI design. As such, this is not the actual design or implementation of the feature - just a high-level concept of what it might look like (just in the same way a wireframe i that and is not considered final or actual experience until the UI/UX design is done and the implementation is complete).

restore-hooks_product-requirements.md
stephbman: Updated. It was a typo :)

restore-hooks_product-requirements.md
stephbman: Great idea - updated. Thanks!

restore-hooks_product-requirements.md
stephbman: That's correct Ashish. Discussed further w/ Nolan and Yaron and we're excluding this requirement from the scope.

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,337 @@
...
+This feature will automate the restore operations/processes in Velero, and will provide restore hook actions in Velero that allows users to execute restore commands on a container. Restore hooks include pre-hook actions and post-hook actions.
+
+This feature will mirror the actions of a Velero backup by allowing Velero to check for any restore hooks specified for a pod. 
+
+Assumptions
+
+
+
+*   Restore operations will be run at the pod level instead of at the volume level. Some databases require the pod to be running and in some cases a user cannot manipulate a volume without the pod running. We need to support more than one type of database for this feature and so we need to ensure that this works broadly as opposed to providing support only for specific dbs. 
+*   The author of the restore hook will be responsible for invoking the restore hook.
```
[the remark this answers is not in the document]
stephbman: updated from author of hook to Velero invoking.

restore-hooks_product-requirements.md
stephbman: Per discussion w/ Ashish - we can run restore hooks during a restore or after a restore. Running the velero restore command is not part of the notion of a hook.  

Will be calling this a post hook.

restore-hooks_product-requirements.md
stephbman: Velero will use the snapshot that is in the backup being restored.

restore-hooks_product-requirements.md
```diff
@@ -0,0 +1,309 @@
...
+
+**Restore Pre-Hooks**
+
+<span style="text-decoration:underline;">Container Command</span>
+
+
+```
+pre.hook.restore.velero.io/container
+
+kubectl patch backupstoragelocation <STORAGE LOCATION NAME> \
```
nrb, line 185: What's the purpose of setting the backup storage location to read only here?

carlisia: I have the same question.

ashish-amarnath:
@carlisia leaving the merge for you bc you had some open questions.

PR #2584, opened 2020-05-27
🏃‍♂️ Setup ci in github actions

ashish-amarnath:
Signed-off-by: Ashish Amarnath <ashisham@vmware.com>

Fixes: #2104

Similar to https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure/pull/44
Example run using my fork, my docker creds set as secret in my fork.
https://github.com/ashish-amarnath/velero/runs/714981809?check_suite_focus=true
https://github.com/ashish-amarnath/velero/actions?query=workflow%3APush

PR #2704, opened 2020-07-13
v1.4.1 cherrypicks

nrb:


ashish-amarnath:
@nrb I think we should add #2588 and #2584 to be able to publish container images for the release.

nrb:
Yeah, GitHub Actions are set to match against all tags, which aren't associated with a given branch. Given that information, I don't believe they need to be included.

PR #2710, opened 2020-07-14
Don't check non-code labeled PRs for changelogs

nrb:
The labels 'Documentation', 'Website', and 'Design' are all for PRs
exclusively related to those things, not code, so they don't need to be
checked for changelogs or have the extra label applied.

### DIFF
--- a/.github/workflows/pr-changelog-check.yml
+++ b/.github/workflows/pr-changelog-check.yml
@@ -11,5 +11,5 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Changelog check
-      if: ${{ !contains(github.event.pull_request.labels.*.name, 'changelog-not-required') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-changelog-check.yml
```diff
@@ -11,5 +11,5 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Changelog check
-      if: ${{ !contains(github.event.pull_request.labels.*.name, 'changelog-not-required') }}
+      if: ${{ !contains(github.event.pull_request.labels.*.name, 'changelog-not-required') || !contains(github.event.pull_request.labels.*.name, 'Design') || !contains(github.event.pull_request.labels.*.name, 'Website') || !contains(github.event.pull_request.labels.*.name, 'Documentation')}}
```
nrb, line 14: Looks like this broke the logic for the check since it doesn't have all the labels... worked too fast. Need to double check my logic.

Issue #2400, opened 2020-04-06
Request for new plugin kind to run custom delete logic while deletion of backup.


ashish-amarnath:
**Describe the problem/challenge you have**
Resources created during backup and restore of items by their corresponding ItemAction plugins will not get cleaned up unless the backup deletion controller is updated to perform the necessary cleanup.

**Describe the solution you'd like**
There should be a new plugin kind that runs on backup deletion. Implementation of this plugin can be used to perform cleanup/house keeping for resources created during backup and restore of resources by their corresponding ItemAction plugins.

cc @nrb @carlisia @skriss

nrb:
>Resources created during backup and restore of items by their corresponding ItemAction plugins will not get cleaned up unless the backup deletion controller is updated to perform the necessary cleanup.

To be very clear here - these are resources explicitly tied to and created by the Velero backups themselves, such as CSI VolumeSnapshots. Resources that are simply discovered via the API server, such as the pod -> pvc -> pv relationship are not covered by this.

At the moment, I'm not aware of any plugins that do this besides the Velero CSI plugins, so we're building it into core. That being said, I can see this being useful in the long term, especially as upstream data protection efforts grow. So I think we should track this but rank its priority low for the moment.

carlisia:
> unless the backup deletion controller is updated to perform the necessary cleanup

To clarify: we have the option of updating the deletion controller to handle this clean up, or to create a new plugin kind? If so, I would prefer to update the deletion controller.

agill17:
I have a use case where I have developed a custom plugin that runs some arbitrary logic and creates a snapshot of an external resource. Unfortunately when velero backups expire or is deleted by hand, the external snapshot does not get deleted. I would like to have an extension point that runs some cleanup logic when backup CR is cleaned up.

ashish-amarnath:
+1 _upvoting my own issue_

vishnuitta:
In case of CSI model of plugins i.e., BackupItemAction, it makes sense to have extension to perform cleanup during deletion.
OpenEBS being volume snapshotter, DeleteSnapshot API is good enough to perform cleanup activities during Backup deletion.

> This behavior could also perhaps be implemented as one of the deletion plugins that takes a ConfigMap with the CRDs in it.

This as well sounds like a generic solution.

nrb:
Yeah, for VolumeSnapshotter implementations, I think it makes sense.

[...]

PR #2713, opened 2020-07-15
Add design doc for DeletionAction plugins

nrb:
Design for #2400 

### DIFF
--- a/design/delete-item-action.md
+++ b/design/delete-item-action.md
@@ -1,0 +1,97 @@
+# Delete Item Action Plugins
+
+## Abstract
+Velero should provide a way to delete items created during a backup, with a model and interface similar to that of BackupItemAction and RestoreItemAction plugins.
+These plugins would be invoked when a backup is deleted, and would receive items from within the backup tarball.
+
+## Background
+As part of Container Storage Interface (CSI) snapshot support, Velero added a new pattern for backing up and restoring snapshots via BackupItemAction and RestoreItemAction plugins.
+When others have tried to use this pattern, however, they encountered issues with deleting the resources made in their own ItemAction plugins, as Velero does not expose any sort of extension at backup deletion time.
+These plugins largely seek to delete resources that exist outside of Kubernetes.
+This design seeks to provide the missing extension point.
+
+## Goals
+- Provide a DeleteItemAction API for plugins to implement
+- Update Velero backup deletion logic to invoke registered DeleteItemAction plugins.
+
+## Non Goals
+- Specific implementations of hte DeleteItemAction API beyond test cases.
+- Rollback of DeleteItemAction execution.
+
+
+## High-Level Design
+The DeleteItemAction plugin API will closely resemble the RestoreItemAction plugin design, in that plugins will receive the Velero `Backup` Go struct that is being deleted and a matching Kubernetes resource extracted from the backup tarball.
+
+The Velero backup deletion process will be modified so that if there are any DeleteItemAction plugins registered, the backup tarball will be downloaded and extracted, similar to how restore logic works now.
+Then, each item in the backup tarball will be iterated over to see if a DeleteItemAction plugin matches for it.
+If a DeleteItemAction plugin matches, the `Backup` and relevant item will be passed to the DeleteItemAction.
+
+The DeleteItemAction plugins will be run _first_ in the backup deletion process, before deleting snapshots from storage or `Restore`s from the Kubernetes API server.
+
+DeleteItemAction plugins *cannot* rollback their actions.
+This is because there is currently no way to recover other deleted components of a backup, such as volume/restic snapshots or other DeleteItemAction resources.
+
+DeleteItemAction plugins will be run in alphanumeric order based on their registered names.
+
+## Detailed Design
+
+### New types
+
+The `DeleteItemAction` interface is as follows:
+
+```go
+// DeleteItemAction is an actor that performs an action based on an item in a backup that is being deleted.
+type DeleteItemAction interface {
+	// AppliesTo returns information about which resources this action should be invoked for.
+	// A DeleteItemAction's Execute function will only be invoked on items that match the returned
+	// selector. A zero-valued ResourceSelector matches all resources.
+    AppliesTo() (ResourceSelector, error)
+
+	// Execute allows the ItemAction to perform arbitrary logic with the item being deleted.
+    Execute(DeleteItemActionInput) error
+}
+```
+
+The `DeleteItemActionInput` type is defined as follows:
+
+```go
+type DeleteItemActionInput struct {
+	// Item is the item taken from the pristine backed up version of resource.
+	Item runtime.Unstructured
+	// Backup is the representation of the backup resource processed by Velero.
+	Backup *api.Backup
+}
+```
+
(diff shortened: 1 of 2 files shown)


design/deletion-plugins.md
```diff
@@ -0,0 +1,77 @@
...
+- Implementations of specific deletion plugins.
+- Rollback of deletion plugin execution.
+
+
+## High-Level Design
+Velero will provide a new plugin type that is similar to its existing plugin architecture.
+These plugins will be referred to as `DeletionAction` plugins.
+`DeletionAction` plugins will receive the `Backup` CustomResource being deleted on execution.
+
+`DeletionAction` plugins cannot prevent deletion of an item.
```
ashish-amarnath, line 29: nit: can we call this `DeleteAction` plugin. somehow reads better than `DeletionAction` plugin.

design/deletion-plugins.md
```diff
@@ -0,0 +1,77 @@
...
+In order to ensure that `Backup`s don't execute `DeletionAction` plugins that are not relevant to them, `DeletionAction` plugins can register an `AppliesTo` function which will define a label selector on Velero backups.
+
+## Detailed Design
+The `DeletionAction` plugins will implement the following Go interface, defined in `pkg/plugin/velero/deletion_action.go`:
+
+```go
+type DeletionAction struct {
+
+    // AppliesTo will match the DeletionAction plugin against Velero Backups that it should operate against.
+    AppliesTo()
```
ashish-amarnath, line 43: Is the expectation that the `BackupItemAction` would add labels to the backup object when they run to indicate that the `DeleteAction` should be executed? Considering that we are unable to add labels to the backup object, this will require us to add that functionality before delivering this new plugin type.
Til we support labeling the backup object in the `BackupItemAction` users will be required to add labels outside of Velero. I think we should call that out in the design as a workaround to the open issue.

design/deletion-plugins.md
```diff
@@ -0,0 +1,77 @@
...
+`DeletionAction` plugins will receive the `Backup` CustomResource being deleted on execution.
+
+`DeletionAction` plugins cannot prevent deletion of an item.
+This is because multiple `DeletionAction` plugins can be registered, and this proposal does not include rollback and undoing of a deletion action.
+Thus, if multiple `DeletionAction` plugins have already run but another would request the deletion of a backup stopped, the backup that's retained would be inconsistent.
+
+`DeletionActions` will apply to `Backup`s based on a label on the `Backup` itself.
+In order to ensure that `Backup`s don't execute `DeletionAction` plugins that are not relevant to them, `DeletionAction` plugins can register an `AppliesTo` function which will define a label selector on Velero backups.
+
+## Detailed Design
```
ashish-amarnath, line 36: We should also mention that there will not be any ordering guarantees for the `DeleteAction` plugins when running multiple delete action plugins for a backup.

design/deletion-plugins.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+The following methods would be added to the `clientmgmt.Manager` interface in `pkg/pluginclientmgmt/manager.go`:
+
+```
+type Manager interface {
+    ...
+
+    // GetDeletionActions returns the registered DeletionActions.
+    //TODO: do we need to get these by name, or can we get them all?
```
ashish-amarnath, line 58: Maybe we should do both? Like we do for the backup and restore plugins.

design/deletion-plugins.md
nrb: As I was writing this, I was going to include an `AppliesTo` function, just an `Execute` one, so this selection logic isn't super thought out yet. The idea would be that every delete plugin gets every `Backup` and then does its own internal check to see if needs to do any work. Letting the plugins decide at deletion time may be a workable MVP, though.

Requiring the user to manually label the `Backup` objects isn't a workable solution in my mind, because it's a post- or mid- backup step that Velero absolutely should take care of, somehow. The somehow is the hard part, though, cause right now mutating the `Backup` object in `BackupItemAction`s doesn't happen. If automatic selection is desired, then I'm thinking that maybe we pull that into scope here, but I'm also curious what @dsu-igeek's requirements are.

design/deletion-plugins.md
nrb: We could make them alphanumerical by name, just so we have some known order.

design/deletion-plugins.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Non Goals
+- Implementations of specific deletion plugins.
+- Rollback of deletion plugin execution.
+
+
+## High-Level Design
+Velero will provide a new plugin type that is similar to its existing plugin architecture.
+These plugins will be referred to as `DeleteAction` plugins.
+`DeleteAction` plugins will receive the `Backup` CustomResource being deleted on execution.
```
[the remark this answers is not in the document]
nrb: @dsu-igeek I've added a new proposal with DeleteItemAction.

DeleteAction was chosen at first because don't currently download the tarball and iterate over it - we just delete the backup tarball entirely, since we're deleting the _backup_, not individual resources in it.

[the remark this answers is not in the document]
dsu-igeek: The DeleteItemAction looks like what we need!

design/deletion-plugins.md
```diff
@@ -0,0 +1,80 @@
+# Deletion Plugins
```
nrb, line 1: I'm currently leaning towards doing just `DeleteItemActions` right now, and deferring the `DeleteActions` plugins until later, if at all. For @carlisia and @ashish-amarnath, would you prefer I remove this file?

design/deletion-plugins.md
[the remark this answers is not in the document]
nrb: @dsu-igeek Thanks!

design/delete-item-action.md
```diff
@@ -0,0 +1,97 @@
...
+}
+```
+
+Both `DeleteItemAction` and `DeleteItemActionInput` will be defined in `pkg/plugin/velero/delete_item_action.go`.
+
+### Plugin manager changes
+
+TODO
+
+### Deletion controll modifications
```
ashish-amarnath, line 72: typo: controller

design/delete-item-action.md
```diff
@@ -0,0 +1,97 @@
...
+
+`pkg/controller/backup_deletion_controller.go` will be updated to have plugin management invoked.
+
+In `processRequest`, before deleting snapshots, get any registered `DeleteItemAction` plugins.
+If there are none, proceed as normal.
+If there are one or more, download the backup tarball from backup storage, untar it to temporary storage, and iterate through the items, matching them to the applicable plugins.
+
+
+
+## Alternatives Considered
```
ashish-amarnath, line 82: should we include the `DeleteAction` plugin as an alternative considered?

design/deletion-plugins.md
ashish-amarnath: I would say we keep is and list it as alternatives considered.

nrb:
@ashish-amarnath @carlisia I've pulled this out of draft. I'm keeping both files here, but marking one of them as an alternative.

nrb:
@carlisia Can you please review?

carlisia:
On it.

PR #2734, opened 2020-07-20
Create hook package

### LINKED ISSUE #2726: Move pkg/backup/item_hook_handler.go into a new hooks package
Also add any new restore hooks.

For the implementer - consider whether this should be an internal package or not.

nrb:
Fixes #2726 

* Moves existing types and functions into a new, internal package.
* Adds the `internal` directory to the `hack/test.sh` script for inclusion into `make test`.

### DIFF
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package backup
+package hook
 
 import (
 	"encoding/json"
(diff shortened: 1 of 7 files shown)


internal/hook/item_hook_handler.go
```diff
@@ -38,35 +38,35 @@ import (
 type hookPhase string
 
 const (
-	hookPhasePre  hookPhase = "pre"
-	hookPhasePost hookPhase = "post"
+	PhasePre  hookPhase = "pre"
```
ashish-amarnath, line 41: should this be renamed to indicate it's for backup hooks?

nrb: The main reason for the name change was to reduce stutter, so it wasn't `hook.HookPhase...`.

In terms of for backup or not, I'm not sure. Since restore won't currently have a phase at all, I'm not sure if it's necessary or not, but it sounds like you're in favor?

ashish-amarnath: Yes, I was hoping to make it explicit that there is no notion of a pre and a post hook for restore hooks.

internal/hook/item_hook_handler.go
```diff
@@ -205,23 +205,24 @@ func getPodExecHookFromAnnotations(annotations map[string]string, phase hookPhas
...
 
-type resourceHook struct {
-	name          string
-	namespaces    *collections.IncludesExcludes
-	resources     *collections.IncludesExcludes
-	labelSelector labels.Selector
-	pre           []api.BackupResourceHook
-	post          []api.BackupResourceHook
+// ResourceHook is a hook for a given resource.
+type ResourceHook struct {
```
ashish-amarnath, line 209: suggest qualifying as `BackupResourceHook`

ashish-amarnath:
@carlisia PTAL.

internal/hook/item_hook_handler.go
nrb: This specifically I left generic because I was thinking it would be re-usable for the non-initContainer restore hooks.

I could keep this one and split out a Backup one that includes the Pre/Post fields.

ashish-amarnath: Yeah. We can break this up and do the refactor work as part of restore hook implementation.
Keeping it the way it is now wfm.

nrb: 👍 I'm cool with that plan - keeping it for now, but doing some clarification work when it comes time to actually add new types.

nrb:
@carlisia Done.

PR #2740, opened 2020-07-21
Fix path + add helpful message

carlisia:
All the yaml files and also the generated `crds.go` file go under `config/crd`, the `config/crd/bases` only contains the yaml files. This PR fixes that and extends the warning to a more helpful message.

### DIFF
--- a/hack/verify-generated-crd-code.sh
+++ b/hack/verify-generated-crd-code.sh
@@ -22,8 +22,8 @@ ${HACK_DIR}/update-generated-crd-code.sh --verify-only
 if ! git diff --exit-code config/crd/crds/crds.go >/dev/null; then
   # revert changes to state before running CRD generation to stay consistent
   # with code-generator `--verify-only` option which discards generated changes
-  git checkout config/crd/bases
+  git checkout config/crd
 
-  echo "CRD verification - failed! Generated CRDs are out-of-date, please run 'make update'."
+  echo "CRD verification - failed! Generated CRDs are out-of-date, please run 'make update' and 'git add' the generated file(s)."
   exit 1
 fi


hack/verify-generated-crd-code.sh
```diff
@@ -22,8 +22,8 @@ ${HACK_DIR}/update-generated-crd-code.sh --verify-only
 if ! git diff --exit-code config/crd/crds/crds.go >/dev/null; then
   # revert changes to state before running CRD generation to stay consistent
   # with code-generator `--verify-only` option which discards generated changes
-  git checkout config/crd/bases
+  git checkout config/crd
 
-  echo "CRD verification - failed! Generated CRDs are out-of-date, please run 'make update'."
+  echo "CRD verification - failed! Generated CRDs are out-of-date, please run 'make update' and 'git add' the generated file(s)."
```
ashish-amarnath, line 27: We will want users to run `make update` and commit the changes. Just `git add` will not be sufficient.

carlisia: Pretty sure I tested it and it was sufficient. Testing again.

carlisia: @ashish-amarnath confirming, just `git add` is enough to make `make ci` pass after updating and generating new files.

nrb: Can confirm that `git add` was all that was needed.

ashish-amarnath: That's very interesting!
What I am not understanding is that if `git add` is sufficient, then how are the changes committed?

carlisia: You need to commit them manually, but not having them committed doesn't preclude you from running `make ci`. As opposed to, if you don't `git add` then `make ci` will run the verify check and that will "checkout" your changes, so then you have to run `make update` again. The warning added in this PR is to help ppl from falling into that loop.

ashish-amarnath: Ah! So the `git add` is just to skip over the if condition on line 22 as it doesn't look at staged files.
Makes sense.

carlisia: Yeah, that's a much shorter way to put it lol.

PR #2754, opened 2020-07-23
Refactor image builds to use buildx for multi arch image building

### LINKED ISSUE #2466: Unable to install ca-certificate package on arm builds

RobReus:
Fixes #2466 , Fixes #2486 , Closes #2566 

This change introduces [buildx](https://docs.docker.com/buildx/working-with-buildx/) to build multi arch container images.

Only the container image building is affected by this change, all other functionality (including locally building) remains the same.

Would love to get some feedback on this. We would probably improve it further to reduce build times. Right now the GitHub workflow takes about the same amount of time as before, even though images are now built in parallel. I noticed that cross compiling within the new way of doing it went from 90 seconds to around 300 seconds. I feel like we can win plenty of time by making the cross compile of the velero binary more efficient.

Succesful workflow on my fork: https://github.com/RobReus/velero/actions/runs/180022436 (most recent ones failed due to my credentials now having access to the velero registry).


Dockerfile
```diff
@@ -0,0 +1,67 @@
+# Copyright 2017, 2019 the Velero contributors.
```
nrb, line 1: Copyright year should be 2020

nrb:
Thanks for this @RobReus! I'll look at it in more detail later today. For now, can you double check to make sure that all changed files have their copyright year set to 2020?

Dockerfile
RobReus: Completely forgot about that, thanks for the feedback. Fixed now.

RobReus:
@nrb I have updated the copyright of the files I touched with this PR.

Makefile
```diff
@@ -43,16 +46,25 @@ VERSION ?= main
...
+	IMAGE_TAGS ?= $(IMAGE):$(VERSION)
+endif
+
 # The version of restic binary to be downloaded for power architecture
 RESTIC_VERSION ?= 0.9.6
 
 CLI_PLATFORMS ?= linux-amd64 linux-arm linux-arm64 darwin-amd64 windows-amd64 linux-ppc64le
-CONTAINER_PLATFORMS ?= linux-amd64 linux-ppc64le linux-arm linux-arm64
-MANIFEST_PLATFORMS ?= amd64 ppc64le arm arm64
+BUILDX_PLATFORMS ?= linux/amd64,linux/arm64,linux/arm/v7,linux/ppc64le
```
ashish-amarnath, line 59: does `buildx` also support building windows containers should we need to build one in the future?

Makefile
```diff
@@ -175,47 +144,29 @@ shell: build-dirs build-env
...
+container-build-env:
+	@docker buildx build \
+	--target=builder-env \
+	--build-arg=PKG=$(PKG) \
+	--build-arg=VERSION=$(VERSION) \
+	--build-arg=GIT_SHA=$(GIT_SHA) \
+	--build-arg=GIT_TREE_STATE=$(GIT_TREE_STATE) \
+	-f Dockerfile .
+
+container:
```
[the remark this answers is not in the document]
RobReus: The default buildx builder uses the `docker` driver, which does indeed not support auto-push. The Workflow environment uses the `docker-container` driver. You can configure your local environment to use `docker-container` as well by doing this:
`docker buildx create --name builder --driver docker-container --use` and then:
`docker buildx inspect --bootstrap` to verify the builder is configured properly.
See also https://medium.com/@artur.klauser/building-multi-architecture-docker-images-with-buildx-27d80f7e2408 for more information. I am using the docker image based installation, and not the host based installation.

I am running on Linux, so I can't say if there are any special steps to take on Darwin.

hack/download-restic.sh
```diff
@@ -0,0 +1,54 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if [[ -z "${BIN}" ]]; then
+    echo "BIN must be set"
+    exit 1
+fi
+
+if [[ "${BIN}" != "velero" ]]; then
```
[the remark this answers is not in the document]
RobReus: In the old setup, the `restic` binary was only downloaded into the `Docker-velero[-$ARCH]` images. This is why I am checking if the current binary being built is velero, if not, it will not download restic. I don't mind changing stuff about this, as I was just mimicking what was included before to ensure 100% compatibility.

hack/download-restic.sh
```diff
@@ -0,0 +1,54 @@
...
+    echo "${BIN} does not need the restic binary"
+    exit 0
```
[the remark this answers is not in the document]
RobReus: No, if the `$BIN` is not `velero`, the script should exit gracefully so the `Dockerfile` continues to be built without `restic` (which is the case for the `velero-restic-restore-helper` image).

Makefile
RobReus: I don't have any experience working with Windows container images, but is that really needed? Windows can run `linux/amd64` (in the event the Windows arch is x86_64) without any issues (as far as I know), so there is no need for a Windows based image.

hack/download-restic.sh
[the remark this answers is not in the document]
RobReus: Example of before: https://github.com/vmware-tanzu/velero/blob/main/Dockerfile-velero#L21

The reason I am doing it like this now is so we can use a single `Dockerfile` for all the architectures and binaries. Without this check, the restic binary would also end up in the `velero-restic-restore-helper` image, which it wasn't in before. It probably would not break anything, but this was just me making sure the end-result is still 100% the same :)

Makefile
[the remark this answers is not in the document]
RobReus: Alternatively, you can just remove the `--output=type=registry` from the `Makefile` on line 158. All the layers will still be built, but not pushed. However, due to the way buildx works, you will also not have an image as the docker daemon currently has no support for loading a multi-arch image, so `--load` doesn't work. What you might be able to do is output to "local" so the image is written as a file. See: https://github.com/docker/buildx#-o---outputpath-typetypekeyvalue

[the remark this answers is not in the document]
ashish-amarnath: OK so the instructions seem to be the same for darwin as well.
```
docker buildx create ...
```
We'll want these instructions documented. https://velero.io/docs/v1.4/development/ would be a good place for it.

After that is done, I am seeing this error
```
 => ERROR [linux/ppc64le stage-2 2/3] RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*                         1.6s
 => CANCELED [linux/arm/v7 stage-2 2/3] RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*                       0.6s
 => CANCELED [linux/amd64 stage-2 2/3] RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*                        0.4s
------
 > [linux/ppc64le stage-2 2/3] RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*:
#20 1.242 qemu: uncaught target signal 4 (Illegal instruction) - core dumped
#20 1.251 Illegal instruction
------
failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*]: buildkit-runc did not terminate successfully
```
Any ideas on what this might me?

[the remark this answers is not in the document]
RobReus: @ashish-amarnath sounds like OSX does not play nicely with the ppc64le image. What is the output of `docker buildx inspect --bootstrap`?

Makefile
```diff
@@ -175,47 +144,29 @@ shell: build-dirs build-env
...
+	@docker buildx build --pull \
+	--output=type=registry \
```
ashish-amarnath, line 158: Is there a way to push images outside of the auto-push?
This is to avoid accidental pushes to the `velero` registry and also most of our dev work happens locally without an image registry.

RobReus: There is not right now, it will be possible once https://github.com/moby/moby/pull/38738 has been merged and released.

As a work around in the mean time, I have added a check to the container build recipe to prevent accidental unintended pushes and added a local build option for the container images, which will only build the velero and velero-restic-restore-helper images for the arch you are running it from. This will allow a developer to locally build a docker image so it can be tested.

Let me know if this solution works for you.

Makefile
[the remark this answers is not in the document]
ashish-amarnath: Debugged this offline w/ @RobReus.
Building for `linux/ppc64le` doesn't work on OSX. We might have to exclude `linux/ppc64le` from building on OSX but build that in the CI.

Makefile
ashish-amarnath: That is reasonable to me!

Makefile
```diff
@@ -175,46 +159,37 @@ shell: build-dirs build-env
...
-	@echo "container: $(IMAGE):$(VERSION)"
-
-push: .push-$(DOTFILE_IMAGE) push-name
-.push-$(DOTFILE_IMAGE): .container-$(DOTFILE_IMAGE)
-	@docker push $(IMAGE):$(VERSION)
-ifeq ($(TAG_LATEST), true)
-	docker tag $(IMAGE):$(VERSION) $(IMAGE):latest
-	docker push $(IMAGE):latest
+container-builder-env:
+ifneq ($(BUILDX_ENABLED), true)
```
ashish-amarnath, line 163: suggest adding instructions on how to enable this.
Printing a link to the docs should be suffiecient.

Makefile
```diff
@@ -43,16 +46,34 @@ VERSION ?= main
...
+CI ?= false
+
 # The version of restic binary to be downloaded for power architecture
 RESTIC_VERSION ?= 0.9.6
 
 CLI_PLATFORMS ?= linux-amd64 linux-arm linux-arm64 darwin-amd64 windows-amd64 linux-ppc64le
-CONTAINER_PLATFORMS ?= linux-amd64 linux-ppc64le linux-arm linux-arm64
-MANIFEST_PLATFORMS ?= amd64 ppc64le arm arm64
+BUILDX_PLATFORMS ?= linux/amd64,linux/arm64,linux/arm/v7,linux/ppc64le
+BUILDX_OUTPUT_TYPE ?= registry
```
ashish-amarnath, line 68: Can we make this default to `docker` and use `registry` only from [`hack/docker-push.sh`](https://github.com/vmware-tanzu/velero/blob/main/hack/docker-push.sh)?
Same with the `BUILDX_PLATFORMS`.

That would avoid creating the `local-all-containers` make target. WDYT?

Makefile
RobReus: Agreed, I will be adding some documentation regarding buildx, and link to it here.

Makefile
RobReus: That sounds reasonable to me, I will refactor.

RobReus:
@ashish-amarnath I have updated the contribution docs on building the container images. I have also implemented the feedback given.

Let me know if there's anything else you want to discuss or see changed.

Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+FROM --platform=$BUILDPLATFORM golang:1.14 as builder-env
+
+ARG PKG
+ARG VERSION
+ARG GIT_SHA
+ARG GIT_TREE_STATE
+
+ENV CGO_ENABLED=0 \
+    GO111MODULE=on \
+    GOPROXY=https://proxy.golang.org \
```
nrb, line 23: Is there any way to overide this variable in at build time? I ask because there are times that Velero will be built within a private environment, without access to the internet, and an internal goproxy server should be used.

Perhaps we introduce another `ARG`?

https://vsupalov.com/docker-build-pass-environment-variables/ also lets us use the hosts's GOPROXY env var value at build time, which I think would be useful as a shortcut, but I think making it explicit is easier to follow later on.

RobReus: @nrb fair point. I just pushed a commit which makes `GOPROXY` for both the regular build-image as the multi-arch build image configurable with the default set to `https://proxy.golang.org`.

Issue #2729, opened 2020-07-20
Add initContainers to Restore spec


nrb:
Add an `initContainers` field to the `Restore` spec similar to this example taken from the design doc:

```yaml
apiVersion: velero.io/v1
kind: Restore
spec:
  ...
  hooks:
    resources:
      -
        name: my-hook
        includedNamespaces:
        - '*'
        excludedNamespaces:
        - some-namespace
        includedResources:
        - pods
        excludedResources: []
        labelSelector:
          matchLabels:
            app: velero
            component: server
        post:
          -
            exec:
              container: postgres
              command:
                - /bin/bash
                - -c
                - rm /docker-entrypoint-initdb.d/dump.sql
              onError: Fail
              timeout: 10s
              readyTimeout: 60s
        init:
          timeout: 120s
          initContainers:
          - name: restore
            image: postgres:12
            command: ["/bin/bash", "-c", "mv /backup/dump.sql /docker-entrypoint-initdb.d/"]
            volumeMounts:
            - name: backup
              mountPath: /backup
```

PR #2761, opened 2020-07-25
Add types to implement restore hooks

ashish-amarnath:

Fixes: #2729

This PR adds the following types required to implement restore hooks from the [restore-hooks design doc](https://github.com/vmware-tanzu/velero/blob/main/design/restore-hooks.md):
1. `ExecRestoreHook` - New type to represent a restore hook executing a command from inside a container of a restored pod
1. `InitRestoreHook` - New type to represent a restore hook adding init containers to a pod during its restore
1. `RestoreResourceHook`- New type to represent a restore hook for a resource. Following the naming conventions from backup hooks, this type is named `RestoreResourceHook` even though this currently applies to pods.
1. `PostHooks`- A list of `RestoreResourceHook` items is the new field in the `RestoreSpec`


ashish-amarnath:
@areed and @marccampbell PTAL

areed:
This LGTM, but it reveals an inconsistency in the design since the init timeout is not actually used anywhere.

pkg/apis/velero/v1/restore.go
```diff
@@ -80,6 +83,100 @@ type RestoreSpec struct {
...
+
+// ExecRestoreHook is a hook that uses pod exec API to execute a command inside a container in a pod
+type ExecRestoreHook struct {
+	// Container is the container in the pod where the command should be executed. If not specified,
+	// the pod's first container is used.
+	// +optional
+	Container string `json:"container,omitempty"`
+
+	// Command is the command and arguments to execute from within a container after a pod has been restored.
+	// +kubebuilder:validation:MinItems=1
```
nrb, line 152: Is this mixing kubebuilder and non-kubebuilder generation in the same type, since this is eventually included in `Restore`?

pkg/apis/velero/v1/restore.go
```diff
@@ -80,6 +83,100 @@ type RestoreSpec struct {
...
+	ExcludedResources []string `json:"excludedResources,omitempty"`
+
+	// LabelSelector, if specified, filters the resources to which this hook spec applies.
+	// +optional
+	// +nullable
+	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
+
+	// PostHooks is a list of RestoreResourceHooks to execute during and after restoring a resource.
+	// +optional
+	PostHooks []RestoreResourceHook `json:"post,omitempty"`
```
nrb, line 132: Can we call the JSON field `postHooks` too?

pkg/apis/velero/v1/restore.go
nrb: Also - since this is set to at least 1 item in the field, how does that interact with `ExecRestoreHooks` being optional on the `RestoreSpec`? Does it mean that the validation will happen if the `ExecRestoreHook` is non-empty?

ashish-amarnath: The `kubebuilder` annotations are all copied from the `BackupSpec`.
But to answer your question, yes the Validation will happen if `ExecRestoreHook` is not-empty.

pkg/apis/velero/v1/restore.go
ashish-amarnath: fixed.

ashish-amarnath:
@areed The way I am expecting the init container restore hook to work is:
Velero restores the pod with the necessary init container and with that ends Velero's "restore" work.
The init container runs as long as it needs to before the application is able to restart. Velero doesn't have to wait for the init container, its only job is to add the init container to the restored pod.
I've kept the timeout in the spec to be in line w/ the design. We can clean up unused fields post/during implementation.

The goal of this PR is to have the RestoreSpec ready for implementation 😄

ashish-amarnath:
@carlisia PTAL!

PR #2762, opened 2020-07-25
Exclude vols mounting secrets and configmaps from defaultVolumesToRestic

ashish-amarnath:

We don't want to include volumes mounting secrets and config maps from defaulting volume backups to restic.
Both secrets and configmaps will both be backed up as individual resources.

### DIFF
--- a/changelogs/unreleased/2762-ashish-amarnath
+++ b/changelogs/unreleased/2762-ashish-amarnath
@@ -1,0 +1,1 @@
+Exclude volumes mounting secrets and configmaps from defaulting volume backups to restic
--- a/pkg/restic/common.go
+++ b/pkg/restic/common.go
@@ -170,6 +170,14 @@ func GetPodVolumesUsingRestic(pod *corev1api.Pod, defaultVolumesToRestic bool) [
 		if pv.HostPath != nil {
 			continue
 		}
+		// don't backup volumes mounting secrets. Secrets will be backed up separately.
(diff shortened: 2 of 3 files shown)


nrb:
Closing & re-opening to test the auto-assign reviewers action.

tlvenn:
Open question, does it make sense to backup nfs volumes ? Likewise, an empydir is something that is expected to be lost, should they really be included ?

ashish-amarnath:
@tlvenn Thank you for raising these questions. 
Adding the secrets and configmaps to the exclusion list here made sense because the secrets and config maps would be backed up as individual resources.
However, excluding the VolumeSources that you mention, and other VolumeSources that share the pod's lifetime, might seem like Velero having strong opinions about how the VolumesSources should be used and backed-up. This would make Velero very rigid and we would like to avoid that and allow flexibility for our users and give them the control.
This is the assumption that we are operating with atm. But, we are definitely open to reevaluating our assumptions.

I have https://github.com/vmware-tanzu/velero/issues/2770 open that tracks this. Please feel free to add comments or 👍 / 👎  on the issue.

tlvenn:
Ha thanks @ashish-amarnath , I missed that issue and I definitely understand your point of view regarding providing a solution that is not opiniated one way or another but imho for the opt-out strategy to be viable, one need to be able to customized a list of excluded volume sources. Will continue this discussion on that issue. Thx again.

Issue #2654, opened 2020-06-19
Velero public API (please take note)


carlisia:
Please take note that what Velero intentionally exposes as a public API are the packages below:

https://github.com/vmware-tanzu/velero/tree/master/pkg/apis/velero/v1

https://github.com/vmware-tanzu/velero/tree/master/pkg/plugin

https://github.com/vmware-tanzu/velero/tree/master/pkg/install

We will be doing work to move code that we don't intend to be for public consumption into the `internal` directory structure so we don't expose packages unintentionally.

If you are relying on packages outside those listed above that are intentionally for public consumption, please let us know.

nrb:
@betta1 @dymurray @vishnuitta Letting you know as you depend on Velero code.

michmike:
i assume we will chat about this in the next 1-2 community meetings. thanks

jwmatthews:
cc @sseago @alaypatel07

nrb:
> i assume we will chat about this in the next 1-2 community meetings. thanks

@michmike Yeah, I think this will be an on-going community meeting agenda item.

betta1:
thanks @nrb for the heads up. 

cc @phuongatemc

mynktl:
Hi @carlisia 

We are using following packages:
- https://github.com/vmware-tanzu/velero/tree/master/pkg/builder
       - We are using this package to build backup/restore/schedule resources. We can write our own package for this, but It will be good to have this package exposed since it includes builder code of Velero resources.

- https://github.com/vmware-tanzu/velero/tree/master/pkg/backup 
       - We are using this package to create delete request for backup. https://github.com/vmware-tanzu/velero/blob/master/pkg/backup/delete_helpers.go

- https://github.com/vmware-tanzu/velero/tree/master/pkg/discovery
       - We are using this package for resource discovery. https://github.com/vmware-tanzu/velero/blob/master/pkg/discovery/helper.go

arianitu:
Outside of the public ones, we are using:

	"github.com/vmware-tanzu/velero/pkg/builder"

runzexia:
using:
`https://github.com/vmware-tanzu/velero/tree/master/pkg/discovery`

areed:
Kots is using `github.com/vmware-tanzu/velero/pkg/label` for `GetValidName`.

OceanChen2012:
> Please take note that what Velero intentionally exposes as a public APIs are the packages below:
> 
[...]

PR #2764, opened 2020-07-26
add kindfor func to get apiresource from gvk

runzexia:
Signed-off-by: RyderXia <ryder.xia@sap.com>

### DIFF
--- a/changelogs/unreleased/2764-runzexia
+++ b/changelogs/unreleased/2764-runzexia
@@ -1,0 +1,1 @@
+discovery Helper support get GroupVersionResource and an APIResource from GroupVersionKind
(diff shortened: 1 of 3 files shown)


ashish-amarnath:
@runzexia Thank you for your PR!
It would be helpful in reviewing to understand the context of this PR. Is there an issue that you are trying solve with this change?

runzexia:
hi @ashish-amarnath , I develop a plugin to backup helm releases.
in this plugin, i could get GVK from mainifest.
And i need to covert GVK -> metav1.APIResource ->ResourceIdentifier.
If it can be supported in helper, I don't need to maintain many of the same logic in the plug-in code base
https://github.com/runzexia/velero-plugin-helm/blob/master/pkg/plugin/backupplugin.go#L152

nrb:
@runzexia I think that sounds reasonable. Do note https://github.com/vmware-tanzu/velero/issues/2654, where the discovery package is not yet guaranteed to be exposed as part of our public API.

If you're intending to rely on it, please comment on that issue and let us know.

nrb:
I'm ok with including this helper personally, given the explanation.

I'm removing the `On Hold` label as the explanation's been provided and the comment was added to the the API issue.

PR #2736, opened 2020-07-20
Improve and clarify cmd help documentation, flags, and examples

scaranoj:
Fixes #2422

PR #2774, opened 2020-07-29
Remove un-used docker actions for the `build-image`

carlisia:
For context:

https://github.com/vmware-tanzu/velero/pull/2736#issuecomment-665351947

### DIFF
--- a/.github/workflows/push-builder.yml
+++ b/.github/workflows/push-builder.yml
@@ -5,18 +5,3 @@ on:
     branches: [ main ]
     paths:
     - 'hack/build-image/Dockerfile'
-
-jobs:
-  build:
-    name: Build
-    runs-on: ubuntu-latest
-    steps:
-
-    - name: Build
-      run: make build-image
-
-    - name: Publish container image
-      run: |
-        docker login -u ${{ secrets.DOCKER_USER }} -p ${{ secrets.DOCKER_PASSWORD }}
-       
-        make push-build-image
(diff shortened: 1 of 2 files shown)


.github/workflows/push-builder.yml
```diff
@@ -13,10 +13,4 @@ jobs:
     steps:
 
     - name: Build
-      run: make build-image
-
-    - name: Publish container image
-      run: |
-        docker login -u ${{ secrets.DOCKER_USER }} -p ${{ secrets.DOCKER_PASSWORD }}
-       
-        make push-build-image
```
ashish-amarnath, line 22: These lines indicate to me that we do intend to push the build-image to dockerhub. But this action hasn't been successful. Possibly because `velero/build-image` repository doesn't exist on dockerhub. I think creating this repository should resolve this error.

carlisia: Yes, it seems to be the intention here but do we need it?

nrb: @carlisia I'd like to have it, so that the builder image is pushed up to docker hub and downloaded rather than rebuilt every time. That would save time over rebuilding every step. This was recently-done work by Duffie to speed up the development cycle.

I'd agree with Ashish that we should fix it rather than remove it.

carlisia:
Ok, I'll close this.

.github/workflows/push-builder.yml
nrb: I pushed the build image up from my local machine - further CI pushes should work now. https://hub.docker.com/r/velero/build-image/tags

ashish-amarnath: I think there is something else that isn't working. I don't see this action running at all from https://github.com/vmware-tanzu/velero/actions?query=workflow%3Abuild-image
Now that we've initialized the image repository, we can see if this action runs.

PR #2779, opened 2020-07-31
Kustomize gen crd label annotation

jenting:
fixes https://github.com/vmware-tanzu/helm-charts/pull/125#discussion_r460305733
It makes sync CRDs from velero to helm-charts easier by directly copy (by manual or by automation).


hack/update-generated-crd-code.sh
```diff
@@ -52,4 +57,25 @@ controller-gen \
...
+for crd in *.yaml; do
+  echo $crd
+  cat <<EOF > kustomization.yaml
+  apiVersion: kustomize.config.k8s.io/v1beta1
+  kind: Kustomization
+  resources:
+  - $crd
+  commonLabels:
+    app.kubernetes.io/name: velero
+  commonAnnotations:
```
nrb, line 72: What if the user isn't installing via Helm? Then we're annotating the CRDs with Helm information, but it's not actually used.

nrb:
I'm not sure I like this approach, as it puts the Helm annotations on the CRDs regardless of whether they're installed by Helm or not. I would think they'd need to be annotated when being installed by Helm, wouldn't they?

jenting:
> I'm not sure I like this approach, as it puts the Helm annotations on the CRDs regardless of whether they're installed by Helm or not. I would think they'd need to be annotated when being installed by Helm, wouldn't they?

Yes, you're right!
Let me think a better approach on this 🤔
Welcome any ideas or comments.

Issue #2730, opened 2020-07-20
Write initContainer hook injection logic


nrb:
From the design doc:

>If initContainers are specified for a pod, Velero will inject the containers into the beginning of the pod's initContainers list. If a restic initContainer is also being injected, the restore initContainers will be injected directly after the restic initContainer. The restore will use a RestoreItemAction to inject the initContainers. Stdout and stderr of the restore initContainers will not be added to the Restore logs. InitContainers that fail will not affect the parent Restore's status.

PR #2787, opened 2020-08-04
✨ Implement restore hooks injecting init containers into pod spec

ashish-amarnath:

Fixes: #2730 

TODO:
- [x] If a restic initContainer is also being injected, the restore initContainers will be injected directly after the restic initContainer. 
- [x] Testing
- [x] Move the restore hook handle logic in `pod_action`RestoreItemAction to its own RestoreItemAction plugin

cc @areed @marccampbell @nrb

### DIFF
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -19,20 +19,26 @@ package hook
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
+	uuid "github.com/gofrs/uuid"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	corev1api "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
-	api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"github.com/vmware-tanzu/velero/pkg/kuberesource"
 	"github.com/vmware-tanzu/velero/pkg/podexec"
+	"github.com/vmware-tanzu/velero/pkg/restic"
 	"github.com/vmware-tanzu/velero/pkg/util/collections"
+	"github.com/vmware-tanzu/velero/pkg/util/kube"
 )
 
 type hookPhase string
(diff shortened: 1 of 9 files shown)


internal/hook/item_hook_handler_test.go
```diff
@@ -703,3 +708,677 @@ func parseLabelSelectorOrDie(s string) labels.Selector {
...
+			restoreHooks: []ResourceRestoreHook{},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			handler := InitContainerRestoreHookHandler{}
+			podMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&tc.podInput)
+			assert.NoError(t, err)
+			actual, err := handler.HandleRestoreHooks(velerotest.NewLogger(), kuberesource.Pods, &unstructured.Unstructured{Object: podMap}, tc.restoreHooks)
```
areed, line 1376: this err is overwritten before being compared to expectedError

internal/hook/item_hook_handler.go
```diff
@@ -239,3 +317,61 @@ func (r ResourceHook) applicableTo(groupResource schema.GroupResource, namespace
...
+	Name         string
+	Selector     ResourceHookSelector
+	RestoreHooks []velerov1api.RestoreResourceHook
+}
+
+func getInitRestoreHookFromAnnotation(annotations map[string]string) *velerov1api.InitRestoreHook {
+	containerImage := annotations[podRestoreHookInitContainerImageAnnotationKey]
+	containerName := annotations[podRestoreHookInitContainerNameAnnotationKey]
+	command := annotations[podRestoreHookInitContainerCommandAnnotationKey]
+	if command == "" || containerImage == "" || containerName == "" {
```
areed, line 332: I wonder whether users would expect to be able to just supply an image and default to ENTRYPOINT for command and generated name for name? If returning from here and at least one of these is non-empty would it make sense to log?

internal/hook/item_hook_handler_test.go
ashish-amarnath: Thanks for the catch!

internal/hook/item_hook_handler.go
ashish-amarnath: Yes, I think we'll want to allow users to use the default `ENTRYPOINT` on the images.
I'll clean that up. Along with fixing the CRD which has an unused kubebuilder constraint.
Adding a log message would also help.

ashish-amarnath:
I discovered this bug in the restore spec https://github.com/vmware-tanzu/velero/pull/2793 during testing.
Other than that these changes looked good in testing.
Going to create a new RestoreItemAction plugin to handle the init container restore hooks and then this PR should be ready for merge.

pkg/restore/pod_action_test.go
```diff
@@ -26,6 +26,8 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
nrb, line 26: Could we use the `builder` package in this file?

I know there's a `Container` builder for init containers. `RestoreHooks` will need an implementation.

Someday I'd like to have the builders generated, because they're kind of a pain to manually make.

internal/hook/item_hook_handler.go
```diff
@@ -76,6 +79,79 @@ type ItemHookHandler interface {
...
+	metadata, err := meta.Accessor(obj)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get a metadata accessor")
+	}
+	pod := new(corev1api.Pod)
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), pod); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	initContainers := []corev1api.Container{}
```
nrb, line 120: A comment here as to why we're doing this would be helpful for future readers.

internal/hook/item_hook_handler.go
```diff
@@ -76,6 +79,79 @@ type ItemHookHandler interface {
...
+		// pod did not have the annotations appropriate for restore hooks
+		// running applicable ResourceRestoreHooks supplied.
+		namespace := metadata.GetNamespace()
+		labels := labels.Set(metadata.GetLabels())
+
+		for _, rh := range resrouceRestoreHooks {
+			if !rh.Selector.applicableTo(groupResource, namespace, labels) {
+				continue
+			}
+			for _, hook := range rh.RestoreHooks {
```
nrb, line 139: This raises a question for me - how are we enforcing order on these containers? Are annotations guaranteed to be ordered? Should we sort them before doing this?

internal/hook/item_hook_handler.go
nrb: I think that's a good idea, too. Is that consistent with how backup hooks work now? If they don't support it, we should make an issue to get that in at some point, but not for 1.5.

ashish-amarnath: I think the backup hook case is slightly different where handling the hook is running a command inside of a container. Where it is assumed that the container already exists on the pod and updating the pod spec to add a new one is not necessary. For that reason, this will be something specific to the restore hook scenario.

pkg/restore/pod_action_test.go
ashish-amarnath: Used the builder for containers and pods. Leaving the RestoreHooks bit for later.

internal/hook/item_hook_handler.go
ashish-amarnath: added.

internal/hook/item_hook_handler.go
ashish-amarnath: I am not sure I understand the question.
If we find restore hooks from the pod annotation, then we run only those restore hook and don't use the ones from the restore spec.
We look/ use the restore hooks from the spec only if the pod doesn't have the restore hook annotation.
Multiple init containers can be specified using restore spec and not annotations.
The ordering of the restore hooks in the restore spec is expected to be preserved.

I added a comment about this at at https://github.com/vmware-tanzu/velero/pull/2787/files#diff-fc409aa9f287e68f0586d526ee9a122dR87-R90

ashish-amarnath:
This looked good in testing. Removing the `WIP`

internal/hook/item_hook_handler.go
nrb: @ashish-amarnath Got it, that makes sense. Thanks!

internal/hook/item_hook_handler.go
```diff
@@ -76,6 +82,87 @@ type ItemHookHandler interface {
...
+		log.Infof("Handling InitRestoreHooks from pod annotaions")
+		initContainers = append(initContainers, hooksFromAnnotations.InitContainers...)
+	} else {
+		log.Infof("Handling InitRestoreHooks from RestoreSpec")
+		// pod did not have the annotations appropriate for restore hooks
+		// running applicable ResourceRestoreHooks supplied.
+		namespace := metadata.GetNamespace()
+		labels := labels.Set(metadata.GetLabels())
+
+		for _, rh := range resrouceRestoreHooks {
```
nrb, line 144: small, but `resrouceRestoreHooks` -> `resourceRestoreHooks`

ashish-amarnath: fixed.

PR #2791, opened 2020-08-04
Reverting change on 1.4 docs and re-applying to main docs

RobReus:
Signed-off-by: Rob Reus <rob@devrobs.nl>

I noticed I accidentally changed the wrong docs file. Should've edited the `main` version and not `1.4`. This PR fixes that.


nrb:
@RobReus Nope, no changelog required! That check is optional.

ashish-amarnath:
Thanks for catching that!

PR #2792, opened 2020-08-04
Add the ability to set the allowPrivilegeEscalation property on the Restic restore helper via plugin ConfigMap

doughepi:
Signed-off-by: Piper Dougherty doughertypiper@gmail.com

Adds the ability to set the `allowPrivilegeEscalation` attribute in the `securityContext` for the Restic restore helper container. Adding the key-value pair `secCtxAllowPrivilegeEscalation: false` in the data section of the Restic plugin configuration ConfigMap is sufficient to pass the attribute to the helper InitContainer when restoring a pod with a Restic managed persistent volume. See the Velero Helm chart section [for this parameter](https://github.com/vmware-tanzu/helm-charts/blob/196e15c8d6bb728c9f5bce0725ff22ebf235e340/charts/velero/values.yaml#L250) for more information.

This was needed because our team implements Gatekeeper/Open Policy Agent policy that restricts container privilege escalation. We can configure the security policy to ignore the namespace that Velero lives in, but we cannot tell it to ignore the InitContainer that Velero uses to restore Restic managed persistent volumes since that lives in the user namespace being restored. Excluding the user namespace from Gatekeeper/Open Policy Agent control would obviously defeat the point of having the policy in the first place. 

This does not change the default behavior of the restoration helper for users that do not specify a value for the new configuration setting.

TODO:

- [ ] Update Velero Helm chart documentation with the new config option.

### DIFF
--- a/pkg/util/kube/security_context_test.go
+++ b/pkg/util/kube/security_context_test.go
@@ -17,11 +17,12 @@ limitations under the License.
 package kube
 
 import (
-	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
+
+	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
 )
 
 func TestParseSecurityContext(t *testing.T) {
--- a/pkg/util/kube/security_context_test.go
+++ b/pkg/util/kube/security_context_test.go
@@ -25,8 +25,9 @@ import (
 
 func TestParseSecurityContext(t *testing.T) {
 	type args struct {
-		runAsUser  string
-		runAsGroup string
+		runAsUser                string
+		runAsGroup               string
+		allowPrivilegeEscalation string
 	}
 	tests := []struct {
 		name     string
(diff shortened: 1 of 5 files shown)


doughepi:
@RobReus could the failed checks be related to your work on the build-image?

RobReus:
@doughepi I'll troubleshoot it right now and get back to you on that.

RobReus:
@doughepi upon closer inspection of the failing lint checks, I noticed that it fails on this: 
https://github.com/vmware-tanzu/velero/blob/db139cf07cd51a2442cd2ff2f455adc7737b8293/pkg/restore/restic_restore_action_test.go#L114
and
https://github.com/vmware-tanzu/velero/blob/db139cf07cd51a2442cd2ff2f455adc7737b8293/pkg/util/kube/security_context_test.go#L53
where the function call has not been adjusted with the newly added parameter.

Once this is fixed, it should work again.

doughepi:
@RobReus thank you so much. I was looking at the wrong error message.

I have corrected the old tests and added a few new ones.

pkg/util/kube/security_context_test.go
```diff
@@ -25,32 +25,43 @@ import (
...
 		wantErr  bool
 		expected *corev1.SecurityContext
 	}{
-		{"valid security context", args{"1001", "999"}, false, &corev1.SecurityContext{
-			RunAsUser:  pointInt64(1001),
-			RunAsGroup: pointInt64(999),
+		{"valid security context", args{"1001", "999", "true"}, false, &corev1.SecurityContext{
+			RunAsUser:                pointInt64(1001),
+			RunAsGroup:               pointInt64(999),
+			AllowPrivilegeEscalation: pointBool(true),
```
ashish-amarnath, line 41: Can you please switch this to use https://github.com/vmware-tanzu/velero/blob/main/pkg/util/boolptr/boolptr.go and remove the function you added on line 81?

doughepi: Fixed in 2342be1 and d29aa65.

PR #2804, opened 2020-08-08
Restore hooks exec

### LINKED ISSUE #2731: Modify restore logic to execute hooks in annotations
From the design doc:

>The post-restore hooks implementation will closely follow the design of restoring pod volumes with restic. The pkg/restore.context type will have new fields hooksWaitGroup and hooksErrs comparable to resticWaitGroup and resticErr. The pkg/restore.context.execute function will start a goroutine for each pod with applicable hooks and then continue with restoring other items. Each hooks goroutine will create a pkg/util/hooks.ItemHookHandler for each pod and send any error on the context.hooksErrs channel. The ItemHookHandler already includes stdout and stderr and other metadata in the Backup log so the same logs will automatically be added to the Restore log (passed as the first argument to the ItemHookhandler.HandleHooks method.)

>The pkg/restore.context.execute function will wait for the hooksWaitGroup before returning. Any errors received on context.hooksErrs will be added to errs.Velero.

>One difference compared to the restic restore design is that any error on the context.hooksErrs channel will cancel the context of all hooks, since errors are only reported on this channel if the hook specified onError: Fail. However, canceling the hooks goroutines will not cancel the restic goroutines. In practice the restic goroutines will complete before the hooks since the hooks do not run until a pod is ready, but it's possible a hook will be executed and fail while a different pod is still in the pod volume restore phase.

>Failed hooks with onError: Continue will appear in the Restore log but will not affect the status of the parent Restore. Failed hooks with onError: Fail will cause the parent Restore to have status Partially Failed.

areed:
Closes #2731

### DIFF
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -452,11 +452,11 @@ func getPodExecRestoreHookFromAnnotations(annotations map[string]string, log log
 	}
 }
 
-type NamedExecRestoreHook struct {
-	Name     string
-	Source   string
-	Hook     velerov1api.ExecRestoreHook
-	executed bool
+type PodExecRestoreHook struct {
+	HookName   string
+	HookSource string
+	Hook       velerov1api.ExecRestoreHook
+	executed   bool
 }
 
 // GroupRestoreExecHooks returns a list of hooks to be executed in a pod grouped by
@@ -466,8 +466,8 @@ func GroupRestoreExecHooks(
 	resourceRestoreHooks []ResourceRestoreHook,
 	pod *corev1api.Pod,
 	log logrus.FieldLogger,
-) (map[string][]NamedExecRestoreHook, error) {
-	byContainer := map[string][]NamedExecRestoreHook{}
+) (map[string][]PodExecRestoreHook, error) {
+	byContainer := map[string][]PodExecRestoreHook{}
 
 	if pod == nil || len(pod.Spec.Containers) == 0 {
 		return byContainer, nil
@@ -482,11 +482,11 @@ func GroupRestoreExecHooks(
 		if hookFromAnnotation.Container == "" {
 			hookFromAnnotation.Container = pod.Spec.Containers[0].Name
 		}
-		byContainer[hookFromAnnotation.Container] = []NamedExecRestoreHook{
+		byContainer[hookFromAnnotation.Container] = []PodExecRestoreHook{
 			{
-				Name:   "<from-annotation>",
-				Source: "annotation",
-				Hook:   *hookFromAnnotation,
+				HookName:   "<from-annotation>",
+				HookSource: "annotation",
+				Hook:       *hookFromAnnotation,
 			},
 		}
 		return byContainer, nil
@@ -503,10 +503,10 @@ func GroupRestoreExecHooks(
 			if rh.Exec == nil {
 				continue
 			}
-			named := NamedExecRestoreHook{
-				Name:   rrh.Name,
-				Hook:   *rh.Exec,
-				Source: "backupSpec",
+			named := PodExecRestoreHook{
+				HookName:   rrh.Name,
+				Hook:       *rh.Exec,
+				HookSource: "backupSpec",
 			}
 			// default to first container in pod if unset, without mutating resource restore hook
 			if named.Hook.Container == "" {
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -27,7 +27,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1api "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -468,7 +467,7 @@ type NamedExecRestoreHook struct {
 // hooks from the restore resource are accumulated.
 func GroupRestoreExecHooks(
 	resourceRestoreHooks []ResourceRestoreHook,
-	pod *v1.Pod,
+	pod *corev1api.Pod,
 	log logrus.FieldLogger,
 ) (map[string][]NamedExecRestoreHook, error) {
 	byContainer := map[string][]NamedExecRestoreHook{}
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -17,7 +17,6 @@ limitations under the License.
 package hook
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -75,7 +74,6 @@ type ItemHookHandler interface {
 	// determine if there are any hooks relevant to the item, taking into account the hook spec's
 	// namespaces, resources, and label selector.
 	HandleHooks(
-		ctx context.Context,
 		log logrus.FieldLogger,
 		groupResource schema.GroupResource,
 		obj runtime.Unstructured,
@@ -171,7 +169,6 @@ type DefaultItemHookHandler struct {
 }
 
 func (h *DefaultItemHookHandler) HandleHooks(
-	ctx context.Context,
 	log logrus.FieldLogger,
 	groupResource schema.GroupResource,
 	obj runtime.Unstructured,
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -407,3 +407,103 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
 
 	return restoreHooks, nil
 }
+
+// getPodExecRestoreHookFromAnnotations returns an ExecRestoreHook based on restore annotations, as
+// long as the 'command' annotation is present. If it is absent, this returns nil.
+func getPodExecRestoreHookFromAnnotations(annotations map[string]string, log logrus.FieldLogger) *velerov1api.ExecRestoreHook {
+	commandValue := annotations[podRestoreHookCommandAnnotationKey]
+	if commandValue == "" {
+		return nil
+	}
+
+	container := annotations[podRestoreHookContainerAnnotationKey]
+
+	onError := velerov1api.HookErrorMode(annotations[podRestoreHookOnErrorAnnotationKey])
+	if onError != velerov1api.HookErrorModeContinue && onError != velerov1api.HookErrorModeFail {
+		onError = ""
+	}
+
+	var execTimeout time.Duration
+	execTimeoutString := annotations[podRestoreHookTimeoutAnnotationKey]
+	if execTimeoutString != "" {
+		if temp, err := time.ParseDuration(execTimeoutString); err == nil {
+			execTimeout = temp
+		} else {
+			log.Warn(errors.Wrapf(err, "Unable to parse exec timeout %s, using default", execTimeoutString))
+		}
+	}
+
+	var waitTimeout time.Duration
+	waitTimeoutString := annotations[podRestoreHookWaitTimeoutAnnotationKey]
+	if waitTimeoutString != "" {
+		if temp, err := time.ParseDuration(waitTimeoutString); err == nil {
+			waitTimeout = temp
+		} else {
+			log.Warn(errors.Wrapf(err, "Unable to parse wait timeout %s, using default", waitTimeoutString))
+		}
+	}
+
+	return &velerov1api.ExecRestoreHook{
+		Container:   container,
+		Command:     parseStringToCommand(commandValue),
+		OnError:     onError,
+		ExecTimeout: metav1.Duration{Duration: execTimeout},
+		WaitTimeout: metav1.Duration{Duration: waitTimeout},
+	}
+}
+
+type NamedExecRestoreHook struct {
+	Name     string
+	Source   string
+	Hook     velerov1api.ExecRestoreHook
+	executed bool
+}
+
+// GroupRestoreExecHooks returns a list of hooks to be executed in a pod grouped by
+// container name. If an exec hook is defined in annotation that is used, else applicable exec
+// hooks from the restore resource are accumulated.
+func GroupRestoreExecHooks(
+	resourceRestoreHooks []ResourceRestoreHook,
+	obj *unstructured.Unstructured,
+	log logrus.FieldLogger,
+) (map[string][]NamedExecRestoreHook, error) {
+	byContainer := map[string][]NamedExecRestoreHook{}
+
+	metadata, err := meta.Accessor(obj)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	hookFromAnnotation := getPodExecRestoreHookFromAnnotations(metadata.GetAnnotations(), log)
+	if hookFromAnnotation != nil {
+		byContainer[hookFromAnnotation.Container] = []NamedExecRestoreHook{
+			{
+				Name:   "<from-annotation>",
+				Source: "annotation",
+				Hook:   *hookFromAnnotation,
+			},
+		}
+		return byContainer, nil
+	}
+
+	// No hook found on pod's annotations so check for applicable hooks from the restore spec
+	labels := metadata.GetLabels()
+	namespace := metadata.GetNamespace()
+	for _, rrh := range resourceRestoreHooks {
+		if !rrh.Selector.applicableTo(kuberesource.Pods, namespace, labels) {
+			continue
+		}
+		for _, rh := range rrh.RestoreHooks {
+			if rh.Exec == nil {
+				continue
+			}
+			named := NamedExecRestoreHook{
+				Name:   rrh.Name,
+				Hook:   *rh.Exec,
+				Source: "backupSpec",
+			}
+			byContainer[named.Hook.Container] = append(byContainer[named.Hook.Container], named)
+		}
+	}
+
+	return byContainer, nil
+}
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -429,7 +429,7 @@ func getPodExecRestoreHookFromAnnotations(annotations map[string]string, log log
 		if temp, err := time.ParseDuration(execTimeoutString); err == nil {
 			execTimeout = temp
 		} else {
-			log.Warn(errors.Wrapf(err, "Unable to parse exec timeout %s, discarding", execTimeoutString))
+			log.Warn(errors.Wrapf(err, "Unable to parse exec timeout %s, ignoring", execTimeoutString))
 		}
 	}
 
@@ -439,7 +439,7 @@ func getPodExecRestoreHookFromAnnotations(annotations map[string]string, log log
 		if temp, err := time.ParseDuration(waitTimeoutString); err == nil {
 			waitTimeout = temp
 		} else {
-			log.Warn(errors.Wrapf(err, "Unable to parse wait timeout %s, discarding", waitTimeoutString))
+			log.Warn(errors.Wrapf(err, "Unable to parse wait timeout %s, ignoring", waitTimeoutString))
 		}
 	}
 
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -17,6 +17,7 @@ limitations under the License.
 package hook
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -26,6 +27,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1api "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -74,6 +76,7 @@ type ItemHookHandler interface {
 	// determine if there are any hooks relevant to the item, taking into account the hook spec's
 	// namespaces, resources, and label selector.
 	HandleHooks(
+		ctx context.Context,
 		log logrus.FieldLogger,
 		groupResource schema.GroupResource,
 		obj runtime.Unstructured,
@@ -169,6 +172,7 @@ type DefaultItemHookHandler struct {
 }
 
 func (h *DefaultItemHookHandler) HandleHooks(
+	ctx context.Context,
 	log logrus.FieldLogger,
 	groupResource schema.GroupResource,
 	obj runtime.Unstructured,
@@ -464,12 +468,15 @@ type NamedExecRestoreHook struct {
 // hooks from the restore resource are accumulated.
 func GroupRestoreExecHooks(
 	resourceRestoreHooks []ResourceRestoreHook,
-	obj *unstructured.Unstructured,
+	pod *v1.Pod,
 	log logrus.FieldLogger,
 ) (map[string][]NamedExecRestoreHook, error) {
 	byContainer := map[string][]NamedExecRestoreHook{}
 
-	metadata, err := meta.Accessor(obj)
+	if pod == nil {
+		return byContainer, nil
+	}
+	metadata, err := meta.Accessor(pod)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -472,7 +472,7 @@ func GroupRestoreExecHooks(
 ) (map[string][]NamedExecRestoreHook, error) {
 	byContainer := map[string][]NamedExecRestoreHook{}
 
-	if pod == nil {
+	if pod == nil || len(pod.Spec.Containers) == 0 {
 		return byContainer, nil
 	}
 	metadata, err := meta.Accessor(pod)
@@ -481,6 +481,10 @@ func GroupRestoreExecHooks(
 	}
 	hookFromAnnotation := getPodExecRestoreHookFromAnnotations(metadata.GetAnnotations(), log)
 	if hookFromAnnotation != nil {
+		// default to first container in pod if unset
+		if hookFromAnnotation.Container == "" {
+			hookFromAnnotation.Container = pod.Spec.Containers[0].Name
+		}
 		byContainer[hookFromAnnotation.Container] = []NamedExecRestoreHook{
 			{
 				Name:   "<from-annotation>",
@@ -506,6 +510,10 @@ func GroupRestoreExecHooks(
 				Name:   rrh.Name,
 				Hook:   *rh.Exec,
 				Source: "backupSpec",
+			}
+			// default to first container in pod if unset, without mutating resource restore hook
+			if named.Hook.Container == "" {
+				named.Hook.Container = pod.Spec.Containers[0].Name
 			}
 			byContainer[named.Hook.Container] = append(byContainer[named.Hook.Container], named)
 		}
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -429,7 +429,7 @@ func getPodExecRestoreHookFromAnnotations(annotations map[string]string, log log
 		if temp, err := time.ParseDuration(execTimeoutString); err == nil {
 			execTimeout = temp
 		} else {
-			log.Warn(errors.Wrapf(err, "Unable to parse exec timeout %s, using default", execTimeoutString))
+			log.Warn(errors.Wrapf(err, "Unable to parse exec timeout %s, discarding", execTimeoutString))
 		}
 	}
 
@@ -439,7 +439,7 @@ func getPodExecRestoreHookFromAnnotations(annotations map[string]string, log log
 		if temp, err := time.ParseDuration(waitTimeoutString); err == nil {
 			waitTimeout = temp
 		} else {
-			log.Warn(errors.Wrapf(err, "Unable to parse wait timeout %s, using default", waitTimeoutString))
+			log.Warn(errors.Wrapf(err, "Unable to parse wait timeout %s, discarding", waitTimeoutString))
 		}
 	}
 
--- a/internal/hook/item_hook_handler_test.go
+++ b/internal/hook/item_hook_handler_test.go
@@ -828,6 +828,39 @@ func TestGroupRestoreExecHooks(t *testing.T) {
 					podRestoreHookTimeoutAnnotationKey, "1s",
 					podRestoreHookWaitTimeoutAnnotationKey, "1m",
 				)).
+				Containers(&corev1api.Container{
+					Name: "container1",
+				}).
+				Result(),
+			expected: map[string][]NamedExecRestoreHook{
+				"container1": {
+					{
+						Name:   "<from-annotation>",
+						Source: "annotation",
+						Hook: velerov1api.ExecRestoreHook{
+							Container:   "container1",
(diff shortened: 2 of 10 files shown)


ashish-amarnath:
@areed you should now be able rebase on top of main now that PR #2787 is merged

areed:
Thanks @ashish-amarnath, rebased. I'm going to attempt to refactor DefaultWaitExecHookHandler to use the hook.ItemHookHandler interface to conform to the design doc. Then unit tests for DefaultWaitExecHookHandler should be the last step before it's ready for review.

areed:
[Refactoring](https://github.com/vmware-tanzu/velero/pull/2804/commits/21a9d78626a325b37a3e0f85c1c5dfdfbe9849c2) to use hook.ItemHookHandler interface was clunky. Is there any room for deviating from the design at this point?

areed:
@ashish-amarnath PTAL

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +411,106 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
...
+	}
+	metadata, err := meta.Accessor(pod)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	hookFromAnnotation := getPodExecRestoreHookFromAnnotations(metadata.GetAnnotations(), log)
+	if hookFromAnnotation != nil {
+		byContainer[hookFromAnnotation.Container] = []NamedExecRestoreHook{
+			{
+				Name:   "<from-annotation>",
```
ashish-amarnath, line 487: did we miss the annotation name for this?

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +411,106 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
 
 	return restoreHooks, nil
 }
+
+// getPodExecRestoreHookFromAnnotations returns an ExecRestoreHook based on restore annotations, as
+// long as the 'command' annotation is present. If it is absent, this returns nil.
+func getPodExecRestoreHookFromAnnotations(annotations map[string]string, log logrus.FieldLogger) *velerov1api.ExecRestoreHook {
+	commandValue := annotations[podRestoreHookCommandAnnotationKey]
+	if commandValue == "" {
```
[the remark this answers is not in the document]
ashish-amarnath: Or if the container is `""` we will choose the first container in the pod.

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +411,106 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
...
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	hookFromAnnotation := getPodExecRestoreHookFromAnnotations(metadata.GetAnnotations(), log)
+	if hookFromAnnotation != nil {
+		byContainer[hookFromAnnotation.Container] = []NamedExecRestoreHook{
+			{
+				Name:   "<from-annotation>",
+				Source: "annotation",
+				Hook:   *hookFromAnnotation,
```
ashish-amarnath, line 489: we should check for `container == ""` and use the first container in the pod if that is true.

internal/hook/item_hook_handler_test.go
```diff
@@ -710,6 +711,373 @@ func parseLabelSelectorOrDie(s string) labels.Selector {
...
+			},
+			expected: &velerov1api.ExecRestoreHook{
+				Command: []string{"a", "b", "c"},
+			},
+		},
+		{
+			name: "error mode is continue",
+			inputAnnotations: map[string]string{
+				podRestoreHookCommandAnnotationKey: "/usr/bin/foo",
+				podRestoreHookOnErrorAnnotationKey: "Continue",
```
ashish-amarnath, line 754: Is this case-sensitive?
I can't remember what is it in the backup hook case.

internal/hook/item_hook_handler.go
```diff
@@ -74,6 +76,7 @@ type ItemHookHandler interface {
 	// determine if there are any hooks relevant to the item, taking into account the hook spec's
 	// namespaces, resources, and label selector.
 	HandleHooks(
+		ctx context.Context,
```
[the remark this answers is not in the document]
areed: This was required to make the interface suitable for WaitExecHookHandler in order to conform to the design. I noted in [this comment](https://github.com/vmware-tanzu/velero/pull/2804#issuecomment-672275063) it didn't work out well in practice and I'd prefer to revert to using a separate interface or adding context to the new ItemRestoreHookHandler interface and using that.

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,260 @@
...
+
+		if newPod.Status.Phase == v1.PodSucceeded || newPod.Status.Phase == v1.PodFailed {
+			err := fmt.Errorf("Pod %s/%s entered phase %s before some post-restore exec hooks ran", newPod.Namespace, newPod.Name, newPod.Status.Phase)
+			log.Warning(err)
+			cancel()
+			return
+		}
+
+		for containerName, hooks := range byContainer {
+			if !isContainerRunning(newPod, containerName) {
```
ashish-amarnath, line 132: suggest adding a log message here to indicate the container is not running.

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,260 @@
...
+			cancel()
+		}
+	}
+
+	selector := fields.OneTermEqualSelector("metadata.name", pod.Name)
+	lw := e.ListWatchFactory.NewListWatch(pod.Namespace, selector)
+
+	_, controller := cache.NewInformer(lw, pod, 0, cache.ResourceEventHandlerFuncs{
+		AddFunc: handler,
+		UpdateFunc: func(_, newObj interface{}) {
```
ashish-amarnath, line 195: This means that `handler` will be called once when the pod is created and again when the pod is scheduled and is running.
Can we simplify this by registering only `UpdateFunc` and calling the `handler` only when the pod is `Running`?
We should also keep the `DeleteFunc`

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,260 @@
...
+				if err := e.PodCommandExecutor.ExecutePodCommand(hookLog, podMap, pod.Namespace, pod.Name, hook.Name, eh); err != nil {
+					hookLog.WithError(err).Error("Error executing hook")
+					if hook.Hook.OnError == velerov1api.HookErrorModeFail {
+						errors = append(errors, err)
+						cancel()
+						return
+					}
+				}
+			}
+			delete(byContainer, containerName)
```
ashish-amarnath, line 183: this is not executed when we return from the handler on error.
should this be `defer delete(byContainer, containerName)`?

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,260 @@
...
+	// be deleted, a hook with OnError mode Fail could fail, or it may timeout waiting for
+	// containers to become ready.
+	// Each unexecuted hook is logged as an error but only hooks with OnError mode Fail return
+	// an error from this function.
+	for _, hooks := range byContainer {
+		for _, hook := range hooks {
+			if hook.executed {
+				continue
+			}
+			err := fmt.Errorf("Hook %s in container %s in pod %s/%s not executed", hook.Name, hook.Hook.Container, pod.Namespace, pod.Name)
```
ashish-amarnath, line 217: can we also capture why we didn't run the hook?

internal/hook/wait_exec_hook_handler.go
ashish-amarnath: I see your comment on L207.
makes sense not deferring the delete

internal/hook/item_hook_handler.go
```diff
@@ -169,6 +172,7 @@ type DefaultItemHookHandler struct {
 }
 
 func (h *DefaultItemHookHandler) HandleHooks(
+	ctx context.Context,
```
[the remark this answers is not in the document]
areed: Reverted and used a separate interface

internal/hook/item_hook_handler.go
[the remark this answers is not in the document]
areed: Updated to use first container in pod in the GroupRestoreExecHooks function.

internal/hook/item_hook_handler.go
areed: There is not an annotation for hook name in either backup or restore so I copied the backup behavior and used this string.

internal/hook/item_hook_handler_test.go
areed: I updated the tests to always use the const.

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +411,106 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
...
+	return &velerov1api.ExecRestoreHook{
+		Container:   container,
+		Command:     parseStringToCommand(commandValue),
+		OnError:     onError,
+		ExecTimeout: metav1.Duration{Duration: execTimeout},
+		WaitTimeout: metav1.Duration{Duration: waitTimeout},
+	}
+}
+
+type NamedExecRestoreHook struct {
```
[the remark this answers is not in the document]
areed: Renamed

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +411,106 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
...
+	Name     string
```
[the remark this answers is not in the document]
areed: Renamed to `HookName`

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +411,106 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
...
+	Source   string
```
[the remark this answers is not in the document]
areed: Rename to `HookSource`. Both `HookName` and `HookSource` are used only for adding fields to the logger.

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,260 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	var maxWait time.Duration
+	for _, hooks := range byContainer {
+		for _, hook := range hooks {
+			if hook.Hook.WaitTimeout.Duration > maxWait {
+				maxWait = hook.Hook.WaitTimeout.Duration
+			}
+		}
+	}
+	if maxWait > 0 {
```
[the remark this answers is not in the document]
areed: I added a comment. It means that it will wait indefinitely for the containers to start running (unless the shared hooks context gets canceled).

internal/hook/wait_exec_hook_handler.go
areed: In my tests the pod always got passed to `AddFunc` the first time and then `UpdateFunc` for any changes. I think it means added to the local cache, not to the API. So if we don't watch Add and the pod doesn't change it could be missed entirely.

internal/hook/wait_exec_hook_handler.go
areed: I append the context error, but that's just `context deadline exceeded`. It would be possible to know the reason if something in this function canceled the context, but it could also be canceled by something outside.

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +411,106 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
...
+		onError = ""
+	}
+
+	var execTimeout time.Duration
+	execTimeoutString := annotations[podRestoreHookTimeoutAnnotationKey]
+	if execTimeoutString != "" {
+		if temp, err := time.ParseDuration(execTimeoutString); err == nil {
+			execTimeout = temp
+		} else {
+			log.Warn(errors.Wrapf(err, "Unable to parse exec timeout %s, using default", execTimeoutString))
```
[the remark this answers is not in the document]
areed: There are no defaults for the timeouts, but I updated the message to say it would be discarded.

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +411,106 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
...
+	commandValue := annotations[podRestoreHookCommandAnnotationKey]
+	if commandValue == "" {
+		return nil
+	}
+
+	container := annotations[podRestoreHookContainerAnnotationKey]
+
+	onError := velerov1api.HookErrorMode(annotations[podRestoreHookOnErrorAnnotationKey])
+	if onError != velerov1api.HookErrorModeContinue && onError != velerov1api.HookErrorModeFail {
+		onError = ""
```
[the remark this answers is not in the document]
areed: That does seem like a better default but then it would differ from how Backup exec hooks work. I've left it unchanged for now pending further discussion.

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +411,106 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
...
+
+// GroupRestoreExecHooks returns a list of hooks to be executed in a pod grouped by
+// container name. If an exec hook is defined in annotation that is used, else applicable exec
+// hooks from the restore resource are accumulated.
+func GroupRestoreExecHooks(
+	resourceRestoreHooks []ResourceRestoreHook,
+	pod *v1.Pod,
+	log logrus.FieldLogger,
+) (map[string][]NamedExecRestoreHook, error) {
+	byContainer := map[string][]NamedExecRestoreHook{}
```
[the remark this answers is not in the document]
areed: This map holds all the hooks for a single pod organized by container. Using the containerName as the key works well because then it's easy to check if that container is ready without parsing the string.

The current implementation is that there can be multiple hooks for a single container and they can be accumulated from the same hook in the restore spec or from different hooks.

internal/hook/item_hook_handler.go
```diff
@@ -433,7 +429,7 @@ func getPodExecRestoreHookFromAnnotations(annotations map[string]string, log log
 		if temp, err := time.ParseDuration(execTimeoutString); err == nil {
 			execTimeout = temp
 		} else {
-			log.Warn(errors.Wrapf(err, "Unable to parse exec timeout %s, using default", execTimeoutString))
+			log.Warn(errors.Wrapf(err, "Unable to parse exec timeout %s, discarding", execTimeoutString))
```
ashish-amarnath, line 432: nit: s/discarding/ignoring

internal/hook/item_hook_handler.go
```diff
@@ -482,11 +478,15 @@ func GroupRestoreExecHooks(
 	}
 	hookFromAnnotation := getPodExecRestoreHookFromAnnotations(metadata.GetAnnotations(), log)
 	if hookFromAnnotation != nil {
-		byContainer[hookFromAnnotation.Container] = []NamedExecRestoreHook{
+		// default to first container in pod if unset
+		if hookFromAnnotation.Container == "" {
```
ashish-amarnath, line 482: edge case where this can segfault it `pod.Spec.Containers` is empty.

internal/hook/item_hook_handler.go
```diff
@@ -482,11 +478,15 @@ func GroupRestoreExecHooks(
...
+			hookFromAnnotation.Container = pod.Spec.Containers[0].Name
+		}
+		byContainer[hookFromAnnotation.Container] = []PodExecRestoreHook{
 			{
-				Name:   "<from-annotation>",
-				Source: "annotation",
-				Hook:   *hookFromAnnotation,
+				HookName:   "<from-annotation>",
```
ashish-amarnath, line 487: `"<from-annotation>"` still needs to be substituted with the real thing

internal/hook/wait_exec_hook_handler.go
ashish-amarnath: When a Pod is created, it starts with a `Pending` phase. This is when `AddFunc` handler will be called.
Once the Pod is scheduled and transitions to a `Running` phase, the `UpdateFunc` handler will be called.

It is possible that the pod may have transitioned into the `Running` phase by the time the handler for `AddFunc` can run. When this happens the hook is already run.
When the pod transitions into `Running` it will also queue the update event which will invoke the `UpdateFunc` handler which will run the hook again and this may be undesirable. This approach will guarantee "at least once" semantics for the hook execution.

I am thinking we should register a handler for only `AddFunc`, and not the `UpdateFunc` and wait for the pod to start `Running` and then execute the rook. This will give us "exactly-once" semantics for hook execution.
I can't remember, from the design doc, what semantics we said that these hooks will follow.

ashish-amarnath: Are there other cases in the update event that we should be handling? what if there are updates that we don't really care about? Will that too trigger the execution of the restore hooks?

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+		if len(byContainer) == 0 {
+			cancel()
+		}
+	}
+
+	selector := fields.OneTermEqualSelector("metadata.name", pod.Name)
+	lw := e.ListWatchFactory.NewListWatch(pod.Namespace, selector)
+
+	_, controller := cache.NewInformer(lw, pod, 0, cache.ResourceEventHandlerFuncs{
```
ashish-amarnath, line 178: suggest calling this `podWatcher`. The term controller can be slightly confusing.
s/controller/podWatcher

areed: Renamed

internal/hook/item_hook_handler.go
areed: If we add an annotation for hook name it will deviate from the behavior of [backup hooks](https://github.com/vmware-tanzu/velero/blob/main/internal/hook/item_hook_handler.go#L205). Is that ok?

internal/hook/item_hook_handler.go
areed: L472 above ensures there is at least 1 container in the pod spec.

internal/hook/wait_exec_hook_handler.go
areed: I've added a comment above the handler about how it ensures exactly-once semantics if invoked on both AddFunc and UpdateFunc. Updates to a pod are ignored unless it's either 1. a container starts running or 2. the pod enters a phase that means no more containers in it will ever start running.

carlisia:
Is this still WIP?

internal/hook/wait_exec_hook_handler.go
ashish-amarnath: Thank you for clarifying and adding the comments.
What about the scenario in which:
1. `AddFunc` handler is triggered and is currently running the hook.
2. While the `AddFunc` handler is running, if there is any update to the pod, that will trigger the `UpdateFunc` handler which will also try to run the hook.
This will cause a race between two executions of the hook.

I don't know if the Informer interface does resource de-dup and ensure only one of the handlers is running. I know the controller runtime does perform the de-dup.

ashish-amarnath: Looking at https://github.com/kubernetes/client-go/blob/master/tools/cache/controller.go#L389
Seems like only one of the `OnUpdate`, `OnAdd`, or `OnDelete` functions can be called at a time and that this function should complete before the next `Delta` is handled.
So there shouldn't be a race in the scenario I described.

nrb:
Thanks a ton for your work on this @areed. To set expectations, I'm unlikely to be able to give this a through review until Monday 8/24.

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,263 @@
...
+	if pod == nil {
+		return nil
+	}
+
+	// If hooks are defined for a container that does not exist in the pod log a warning and discard
+	// those hooks to avoid waiting for a container that will never become ready. After that if
+	// there are no hooks left to be executed return immediately.
+	for containerName := range byContainer {
+		if !podHasContainer(pod, containerName) {
+			log.Warningf("Pod %s/%s does not have container %s: discarding post-restore exec hooks", pod.Namespace, pod.Name, containerName)
```
nrb, line 79: Can use [`namespaceAndName`](https://github.com/vmware-tanzu/velero/blob/f42c63af1b9af445e38f78a7256b1c48ef79c10e/pkg/util/kube/utils.go#L36) for the pod here.

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,263 @@
...
+		if newPod.Status.Phase == v1.PodSucceeded || newPod.Status.Phase == v1.PodFailed {
+			err := fmt.Errorf("Pod %s/%s entered phase %s before some post-restore exec hooks ran", newPod.Namespace, newPod.Name, newPod.Status.Phase)
+			log.Warning(err)
+			cancel()
+			return
+		}
+
+		for containerName, hooks := range byContainer {
+			if !isContainerRunning(newPod, containerName) {
+				log.Infof("Container % in pod %s/%s is not running: post-restore hooks will not yet be executed", containerName, newPod.Namespace, newPod.Name)
```
nrb, line 123: Can use [`namespaceAndName`](https://github.com/vmware-tanzu/velero/blob/f42c63af1b9af445e38f78a7256b1c48ef79c10e/pkg/util/kube/utils.go#L36) for the pod here.

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,263 @@
...
+				continue
+			}
+			podMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(newPod)
+			if err != nil {
+				log.WithError(err).Errorf("error unstructuring pod %s/%s", newPod.Namespace, newPod.Name)
```
nrb, line 128: Can use [`namespaceAndName`](https://github.com/vmware-tanzu/velero/blob/f42c63af1b9af445e38f78a7256b1c48ef79c10e/pkg/util/kube/utils.go#L36) for the pod here.

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,263 @@
...
+	waitStart := time.Now()
+
+	var errors []error
+
+	// The first time this handler is called after a container starts running it will execute all
+	// pending hooks for that container. Subsequent invocations of this handler will never execute
+	// hooks in that container. It uses the byContainer map to keep track of which containers have
+	// not yet been observed to be running. It relies on the Informer not to be called concurrently.
+	// When a container is observed running and its hooks are executed, the container is deleted
+	// from the byContainer map. When the map is empty the watch is ended.
```
nrb, line 107: There's a lot of instances of re-using the pod name in log messages inside the `handler` function. Could we perhaps generate a [FieldLogger](https://godoc.org/github.com/sirupsen/logrus#FieldLogger) using the [`namespaceAndName`](https://github.com/vmware-tanzu/velero/blob/f42c63af1b9af445e38f78a7256b1c48ef79c10e/pkg/util/kube/utils.go#L36) function to get the pod's info in every log message?

internal/hook/wait_exec_hook_handler.go
nrb: @ashish-amarnath Is this resolved, then?

internal/hook/wait_exec_hook_handler_test.go
```diff
@@ -0,0 +1,926 @@
...
+				assert.Nil(t, err)
+				podCommandExecutor.On("ExecutePodCommand", mock.Anything, obj, e.pod.Namespace, e.pod.Name, e.name, e.hook).Return(e.error)
+			}
+
+			ctx := context.Background()
+			if test.sharedHooksContextTimeout > 0 {
+				ctx, _ = context.WithTimeout(ctx, test.sharedHooksContextTimeout)
+			}
+
+			// TODO get the byContainer map or specify it in the unit tests
```
nrb, line 739: Is the TODO here meant to be done before merging this PR?

internal/hook/wait_exec_hook_handler_test.go
```diff
@@ -0,0 +1,926 @@
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+
+			source := fcache.NewFakeControllerSource()
+			go func() {
+				source.Add(test.initialPod)
+				for _, change := range test.changes {
```
nrb, line 714: Some comments around this loop, and maybe at the top of the test function, would be helpful to remember that the `changes` array is meant to be a set of versions of a given pod over time.

internal/hook/wait_exec_hook_handler.go
areed: I updated all cases of `%s/%s` to use kube.NamespaceAndName

internal/hook/wait_exec_hook_handler_test.go
areed: That was leftover from when the tests were WIP and I was still passing nil for the map. Removed.

internal/hook/item_hook_handler.go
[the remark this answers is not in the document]
ashish-amarnath: Unresolved this for discussion 👇 

Seems like the `hookLog` defaults the `onError` to `Fail`
https://github.com/vmware-tanzu/velero/blob/main/pkg/podexec/pod_command_executor.go#L103-L109
The fact that the log message says `OnError=Fail` but the restore completing successfully kinda seems like saying one thing but doing something else.

suggest using `Fail` as the default here for restore hooks and opening an issue to do the same thing for the backup hooks as well.
WDYT? @nrb

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,270 @@
...
+				},
+			)
+			hookLog.Error(err)
+			if hook.Hook.OnError == velerov1api.HookErrorModeFail {
+				errors = append(errors, err)
+			}
+		}
+	}
+
+	return errors
```
ashish-amarnath, line 230: When `OnError` is `Fail` and `errors` are returned and the restore at least, `PartiallyFails`. However, the pod is already restored and running.
Not sure how best to handle this. At the very least this behavior should be documented. This behavior may not be that different from the restore behavior overall. But maybe good to document.

areed:
@ashish-amarnath @nrb I pushed the scripts and yaml I used for manual QA to this repo https://github.com/replicatedhq/velero-exec-hooks-qa. All these are passing as of the latest commit.

ashish-amarnath:
@areed  That's awesome! Thanks for sharing that.

nrb:
@areed Thanks for sharing those!

ashish-amarnath:
@areed Would you be able to help us w/ the docs for this as well?

I have https://github.com/vmware-tanzu/velero/pull/2862 in the works for the InitContainer restore hooks.
And this just got merged https://github.com/vmware-tanzu/velero/pull/2855
You can use these for reference.

areed:
@ashish-amarnath Yes I'll start on docs

ashish-amarnath:
Thank you so much! We are also deciding on terminology [here](https://github.com/vmware-tanzu/velero/pull/2862#discussion_r482237164)

ashish-amarnath (commented):
The restore hooks in the spec also looked good in testing.
I am going to make one more pass on the code.

internal/hook/wait_exec_hook_handler.go
ashish-amarnath: yes, this is resolved.

internal/hook/wait_exec_hook_handler.go
ashish-amarnath: The hook failure is definitely surfaced. WE will need to document this. Agree that this is consistent w/ the current design.

internal/hook/wait_exec_hook_handler.go
```diff
@@ -0,0 +1,270 @@
...
+	if len(byContainer) == 0 {
+		return nil
+	}
+
+	// Every hook in every container can have its own wait timeout. Rather than setting up separate
+	// contexts for each, find the largest wait timeout for any hook that should be executed in
+	// the pod and watch the pod for up to that long. Before executing any hook in a container,
+	// check if that hook has a timeout and skip execution if expired.
+	ctx, cancel := context.WithCancel(ctx)
+	maxWait := maxHookWait(byContainer)
```
areed, line 93: There is a bug here. If one hook has a wait timeout of 1 minute and another hook does not have a wait timeout then this will be 1 minute and context will be canceled after 1 minute. But it should wait indefinitely for the 2nd hook that did not specify a wait timeout. I'll push a commit with a fix.

internal/hook/wait_exec_hook_handler.go
ashish-amarnath: Thanks for clarifying!

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +407,114 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
...
+type PodExecRestoreHook struct {
+	HookName   string
+	HookSource string
+	Hook       velerov1api.ExecRestoreHook
+	executed   bool
+}
+
+// GroupRestoreExecHooks returns a list of hooks to be executed in a pod grouped by
+// container name. If an exec hook is defined in annotation that is used, else applicable exec
+// hooks from the restore resource are accumulated.
```
zubron, line 464: Is the intention that exec restore hooks can be set in an annotation **or** in the spec? The [original design states that they can be provided by annotation **and/or** in the restore spec](https://github.com/vmware-tanzu/velero/blob/main/design/restore-hooks.md#detailed-design). I don't know if the desired behaviour has changed since that original design was created? If this is the desired behaviour then that should be made explicit in the accompanying docs.

internal/hook/item_hook_handler.go
```diff
@@ -407,3 +407,114 @@ func GetRestoreHooksFromSpec(hooksSpec *velerov1api.RestoreHooks) ([]ResourceRes
...
+			named := PodExecRestoreHook{
+				HookName:   rrh.Name,
+				Hook:       *rh.Exec,
+				HookSource: "backupSpec",
+			}
+			// default to first container in pod if unset, without mutating resource restore hook
+			if named.Hook.Container == "" {
+				named.Hook.Container = pod.Spec.Containers[0].Name
+			}
+			byContainer[named.Hook.Container] = append(byContainer[named.Hook.Container], named)
```
zubron, line 515: By grouping the hooks by the container in which they run, we can't make any guarantees about the execution order of the hooks (which seems different to the behaviour of the InitContainer hooks). Does that matter?

internal/hook/item_hook_handler.go
ashish-amarnath: It should be supplying exec restore hooks either annotation or in the restore spec.

areed: For a single pod it's an exclusive or and the spec hooks are ignored if annotations are set. For the overall restore there may be some pods that get their hooks from annotations and others that don't have annotations and get hooks from the restore spec. I will update the docs PR.

zubron: Sounds good. Thanks for the explanation! :)

internal/hook/wait_exec_hook_handler.go
nrb: Yeah, lets get it in the documentation in the other PR.

nrb (approved):
Thanks for catching that last bug @areed!

this is looking pretty good to me!  I'm going to defer final sign off to @ashish-amarnath

internal/hook/item_hook_handler.go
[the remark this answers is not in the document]
ashish-amarnath: this can be addressed in a follow-up

ashish-amarnath (approved):
LGTM 🎉

Issue #2131, opened 2019-12-13
Create basic end-to-end test with KinD


nrb:
I don't think we initially need to solve all these cases with 1 PR, but we should do the following:

* Install Velero via `velero install`
* Run a backup of a namespace
* Restore that namespace
* Validate that namespace and all resources were restored
* Validate data inside the volume was restored
* Tear down cluster

Some things we'll need to ensure:

* Buckets are provisioned and cleaned up dynamically
* Snapshots are cleaned up in order to ensure lower billing costs
* Full e2e tests should probably only run on a periodic basis, not every PR.

nrb:
IMO, for the first pass, we don't need snapshots being captured. Just installation and k8s objects backups. That's more than we have today.

carlisia:
>Buckets are provisioned and cleaned up dynamically

Maybe this can be done by scripting the setup of MinIO running on Docker, alongside Kind.

I wanted to drop some resources I've come across:

## evntest
https://book.kubebuilder.io/reference/writing-tests.html

Examples:
- https://github.com/kubernetes-sigs/cluster-api/tree/master/controllers
- https://github.com/microsoft/azure-databricks-operator/tree/master/controllers

It can point to a locally built etcd/apiserver binaries (CI/CD* probably wants this), or can point to an existing cluster (probably ideal for local development, not sure if this or using the built binaries would be faster/better).

* Maybe packaging the etcd and apiserver binaries in a docker container used by a github action would work.

It works with other testing packages that test in the BDD style. I personally think this is a plus, we could benefit from laying out the behavior to test from the time of spec'ing.

## rigging
https://github.com/n3wscott/rigging

Example:
- https://github.com/n3wscott/rigging/tree/master/example

It "leverages lightly templated yaml files". The knative folks are trying to get it into that project. @n3wscott has offered to showcase.

carlisia:
This issue is sorta related in a sense if we are boing to boot up kind to run e2e tests, we might be able to use the same cluster to run integration tests: https://github.com/vmware-tanzu/velero/issues/2518

ashish-amarnath:
Writing down outcomes from some of the discussions we've had around this.

1. We want to decouple cluster provisioning from running of the e2e tests
1. We want to open source the e2e tests for our users to run them if they choose to.
1. All contributors should be able to and be encouraged to run and add e2e tests.
1. Allow users to select which tests they want to run. (something like ginkgo.skip) For example: 
[...]

PR #2805, opened 2020-08-10
🏃‍♂️ Setup crd validation github action on k8s versions

ashish-amarnath:

- Adding github actions to test Velero CRDs can be successfully applied to a kubernetes cluster.
- These actions are setup to run on every PR.
- Added 3 different instead of one GH action running the verification in a loop or multiple jobs to run them in parallel
- Once this PR is merged, will mark CRD verification for kubernetes 1.17 and 1.16.9 as mandatory for PR to merge.
- CRD verification on a kubernetes 1.18.4 cluster currently fails and will be made mandatory after the root cause has been identified.

### DIFF
--- a/.github/workflows/crds-verify-k8s-1-16-9.yaml
+++ b/.github/workflows/crds-verify-k8s-1-16-9.yaml
@@ -1,0 +1,20 @@
+name: "Verify Velero CRDs on k8s 1.16.9"
+on: [pull_request]
+
+jobs:
+  kind:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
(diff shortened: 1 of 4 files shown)


nrb:
Would you say this starts addressing #2131?

ashish-amarnath:
Yes, this starts addressing #2131

carlisia:
@ashish-amarnath this LGTM but it is failing the k8s validation. Does that need to be addressed?

ashish-amarnath:
@carlisia Yes, that'll have to be addressed. But not in this PR.
This PR is just surfacing the problem. Merging this PR will mean we'll have this job fail on PRs till the underlying problem is fixed.

PR #2806, opened 2020-08-10
add note about windows support

a-mccarthy:

### DIFF
--- a/site/docs/main/basic-install.md
+++ b/site/docs/main/basic-install.md
@@ -23,9 +23,9 @@ There are supported storage providers for both cloud-provider environments and o
 
 ### Velero on Windows
 
-Velero does not officially support Windows, but it has been tested successfully when backing up stateless Windows applications only. If you want to test Velero on Windows, note that the restic integration and backups of stateful applications or PersistentVolumes are not supported.
+Velero does not officially support Windows. In testing, the Velero team was able to backup stateless Windows applications only. The restic integration and backups of stateful applications or PersistentVolumes were not supported.
 
-To backup stateless Windows applications, Velero must deployed as a Windows container. While Velero does not provide official Windows images, its possible for you to build your own Velero Windows container image to use. Note that this image must be built on a Windows node.
+If you want to perform your own testing of Velero on Windows, you must deploy Velero as a Windows container. Velero does not provide official Windows images, but its possible for you to build your own Velero Windows container image to use. Note that you must build this image on a Windows node.
 
 ## Install the CLI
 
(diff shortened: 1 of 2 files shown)


site/docs/main/basic-install.md
```diff
@@ -21,6 +21,12 @@ Velero uses object storage to store backups and associated artifacts. It also op
 
 There are supported storage providers for both cloud-provider environments and on-premises environments. For more details on on-premises scenarios, see the [on-premises documentation][2].
 
+### Velero on Windows
+
+Velero does not officially support Windows, but it has been tested successfully when backing up stateless Windows applications only. If you want to test Velero on Windows, note that the restic integration and backups of stateful applications or PersistentVolumes are not supported.
+
+To backup stateless Windows applications, Velero must deployed as a Windows container. While Velero does not provide official Windows images, its possible for you to build your own Velero Windows container image to use. Note that this image must be built on a Windows node.
```
nrb, line 28: I'm wondering if we should provide docs to the build process here or not.

I'm conflicted because providing docs is one step closer to looking like we provide support. @ashish-amarnath, @carlisia, @stephbman, what are your thoughts?

ashish-amarnath: We could say, we don't support officially and for that reason, we don't publish windows imges. But if you would like to try Velero on windows, this is how you can build.
Just offering docs around how to build the container should be good. With this, we are signing up to support just building the container image.
I wonder if buildx has support for windows OOTB worth trying out if there is interest.
I can try to come up w/ instructions to build Velero on windows.

a-mccarthy: @ashish-amarnath and @nrb how much of an effort is it to create the steps for building a windows container? If is something that will be a lot of work, maybe we should consider adding the note now as is and creating an issue for adding the steps to build a windows container. That way we can prioritize adding these steps with the rest of the projects in the backlog? 

@stephbman any thoughts on this?

carlisia, line 28:
```suggestion
To backup stateless Windows applications, Velero must be deployed as a Windows container. While Velero does not provide official Windows images, it is possible for you to build your own Velero Windows container image to use. Note that this image must be built on a Windows node.
```

carlisia: I'd say let's merge this and add the additional documentation later.

nrb: Yeah, I'd prefer we merge this as is and defer the build instructions for later.

a-mccarthy: thanks @carlisia!

PR #2807, opened 2020-08-10
Use pull_request_target event to auto-assign reviewers

nrb:


bogdandrutu:
Don't think this will work because it needs to know for what PR to assign the reviewer

bogdandrutu:
I think you can use https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

nrb:
@bogdandrutu Good call out! I'll change this to use `pull_request_target` instead, that appears to be what we need. 🤞

PR #2814, opened 2020-08-12
🐛  Patch generated yaml for restore CRD as a hacky workaround

### LINKED ISSUE #2811: 🐛  Velero CRDs from the main branch cannot be installed on a kubernetes 1.18 cluster
See this GH action failures for more info.
https://github.com/vmware-tanzu/velero/actions?query=workflow%3A%22Verify+Velero+CRDs+on+k8s+1.18.4%22

**Vote on this issue!**

ashish-amarnath:

Fixes #2811


carlisia:
I think what's happening is that since "TCP" is the default, it should not be explicitly set, since it is implied. I looked at some other fields that have default values and they were not explicitly set.

nrb:
@carlisia Per https://github.com/kubernetes/kubernetes/issues/91395, though, if it's not set, there's another error.

This seems to be confirming the second case mentioned in 91395.

ashish-amarnath:
I was able to get this to work with a more elegant hack.
@nrb and @carlisia  PTAL

carlisia:
Cool. Would you please mind adding more context as documentation on hack/update-generated-crd-code.sh? Ideally the link to the issue would be used for reference if needed, but we wouldn't have to read the issue to figure out why we have this hack.

What I'd like to see, is there a condition that would allow us to remove this script, what we would have to be on the lookout that would require extending that script if any, anything like that.

ashish-amarnath:
@carlisia added some context to the `update-generated-crd-code.sh` file

PR #2835, opened 2020-08-18
fix: rename the PV if VolumeSnapshotter has modified the PV name

pawanpraka1:
When VolumeSnapshotter sets the PV name via SetVolumeID and PV is
not there in the cluster, velero does not rename the PV. Which causes
the pvc to be in the lost state as pvc points to the old PV but pv object
has been renamed by VolumeSnapshotter.

Signed-off-by: Pawan <pawan@mayadata.io>

### DIFF
--- a/changelogs/unreleased/2835-pawanpraka1
+++ b/changelogs/unreleased/2835-pawanpraka1
@@ -1,0 +1,1 @@
+rename the PV if VolumeSnapshotter has modified the PV name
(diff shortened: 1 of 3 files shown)


nrb:
@carlisia and @ashish-amarnath While not on the release blocker list, I think this is a nice-to-have.

carlisia:
Should a test be added for this?

pawanpraka1:
@nrb @carlisia added a unit test case.

pawanpraka1:
@nrb @ashish-amarnath @carlisia in which release this fix will be available?

ashish-amarnath:
this should be in the GA release of 1.5. ETA 2w

PR #2838, opened 2020-08-18
Migrate ServerStatusRequest controller and resource to kubebuilder

carlisia:
Closes: https://github.com/vmware-tanzu/velero/issues/2605.


pkg/cmd/server/server.go
```diff
@@ -94,15 +95,14 @@ const (
...
 	// keys used to map out available controllers with disable-controllers flag
-	BackupControllerKey              = "backup"
-	BackupSyncControllerKey          = "backup-sync"
-	ScheduleControllerKey            = "schedule"
-	GcControllerKey                  = "gc"
-	BackupDeletionControllerKey      = "backup-deletion"
-	RestoreControllerKey             = "restore"
-	DownloadRequestControllerKey     = "download-request"
-	ResticRepoControllerKey          = "restic-repo"
-	ServerStatusRequestControllerKey = "server-status-request"
```
nrb, line 105: It looks like this was removed, was that intentional?

pkg/controller/server_status_request_controller.go
```diff
@@ -17,104 +17,100 @@ limitations under the License.
...
 	log.Debug("Getting ServerStatusRequest")
-	req, err := c.lister.ServerStatusRequests(ns).Get(name)
-	// server status request no longer exists
-	if apierrors.IsNotFound(err) {
-		log.WithError(err).Debug("ServerStatusRequest not found")
-		return nil
+	statusRequest := &velerov1api.ServerStatusRequest{}
+	if err := r.Client.Get(r.Ctx, req.NamespacedName, statusRequest); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.WithError(err).Debug("ServerStatusRequest not found")
```
nrb, line 68: I think this message should be an `Error`, not a `Debug`, as regular users will be using it.

pkg/controller/server_status_request_controller.go
```diff
@@ -17,104 +17,100 @@ limitations under the License.
...
+			return ctrl.Result{}, nil
+		}
+
+		log.WithError(err).Debug("Error getting ServerStatusRequest")
```
nrb, line 72: This log should also be at the `Error` level instead of `Debug`, I think.

pkg/controller/server_status_request_controller.go
```diff
@@ -17,104 +17,100 @@ limitations under the License.
...
+			return ctrl.Result{RequeueAfter: statusRequestResyncPeriod}, nil
+		}
+
+		log.Debug("ServerStatusRequest has expired, deleting it")
+		if err := r.Client.Delete(r.Ctx, statusRequest); err != nil {
+			log.WithError(err).Error("Unable to delete the request")
+			return ctrl.Result{}, nil
+		}
+	default:
+		log.Errorf("unexpected ServerStatusRequest phase %q", statusRequest.Status.Phase)
```
nrb, line 105: The `%q` field is redundant as the `log` has the `phase` field included already.

pkg/controller/server_status_request_controller.go
```diff
@@ -17,104 +17,100 @@ limitations under the License.
...
+		return ctrl.Result{}, errors.Errorf("unexpected ServerStatusRequest phase %q", statusRequest.Status.Phase)
```
nrb, line 106: We should either log the error or return it - seems like in this case we want to return it since somehow the `Phase` got set to something completely unexpected.

pkg/cmd/server/server.go
```diff
@@ -867,6 +851,21 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
+	serverStatusInfo := velero.ServerStatus{
+		PluginRegistry: s.pluginRegistry,
+		Clock:          clock.RealClock{},
+	}
+	if err := (&controller.ServerStatusRequestReconciler{
+		Scheme:       s.mgr.GetScheme(),
+		Client:       s.mgr.GetClient(),
+		Ctx:          s.ctx,
+		ServerStatus: serverStatusInfo,
+
```
nrb, line 863: nit: remove extra line

pkg/cmd/server/server.go
```diff
@@ -776,30 +775,15 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
...
 	enabledControllers := map[string]func() controllerRunInfo{
-		BackupSyncControllerKey:          backupSyncControllerRunInfo,
-		BackupControllerKey:              backupControllerRunInfo,
-		ScheduleControllerKey:            scheduleControllerRunInfo,
-		GcControllerKey:                  gcControllerRunInfo,
-		BackupDeletionControllerKey:      deletionControllerRunInfo,
-		RestoreControllerKey:             restoreControllerRunInfo,
-		ResticRepoControllerKey:          resticRepoControllerRunInfo,
-		DownloadRequestControllerKey:     downloadrequestControllerRunInfo,
-		ServerStatusRequestControllerKey: serverStatusRequestControllerRunInfo,
```
nrb, line 802: This wasn't a big deal for the BSL controller, but it is here - this is a breaking change. With this change, users can no longer specify whether or not the controller can be enabled or not, unless I'm missing something.

The problem will be amplified as we move more controllers, so we'll have to figure out how to retain this functionality.

carlisia: Ah yes, I had missed that, good catch. Fixed it: https://github.com/vmware-tanzu/velero/pull/2838/files#diff-cd1a009fa409db8496ea35b53f9776d2R790.

nrb: Awesome, thanks!

pkg/cmd/server/server.go
```diff
@@ -821,6 +810,13 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
 			s.logger.Fatalf("Invalid value for --disable-controllers flag provided: %s. Valid values are: %s", controllerName, strings.Join(disableControllerList, ","))
 		}
 	}
+	// remove disabled runtime type controllers
+	for _, controllerName := range s.config.disabledControllers {
+		if _, ok := enabledRuntimeControllers[controllerName]; ok {
```
ashish-amarnath, line 815: the `if` check at line 815 can be removed.

internal/velero/serverstatusrequest.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+type ServerStatus struct {
+	PluginRegistry PluginLister
+	Clock          clock.Clock
+}
+
+// PatchStatusProcessed patches status fields, including loading the plugin info, and updates
+// the ServerStatusRequest.Status.Phase to ServerStatusRequestPhaseProcessed.
+func (s *ServerStatus) PatchStatusProcessed(kbClient client.Client, req *velerov1api.ServerStatusRequest, ctx context.Context) error {
```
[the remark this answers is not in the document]
nrb: We also have https://github.com/vmware-tanzu/velero/blob/main/pkg/client/dynamic.go#L78, which is generic regardless of fields, similar to `PatchHelper`.

I think I'd prefer going the generic route, seeing as 1) we have precedence in our dynamic client, though it'll eventually go away 2) we have an example in CAPI's and 3) we already have 2 examples in our own code in the BSL and ServerStatusRequest controllers, which is starting to be repetitive, even if they're not the same fields.

I think we're going to have something to gain from making a generic patching solution because a quick search of the code shows several controllers are doing patches:

```shell
Nolans-MBP in /Users/nrb/go/src/github.com/vmware-tanzu/velero/pkg (git) main
% ag patch -l --ignore apis --ignore generated --ignore backup/testdata
cmd/cli/plugin/remove.go
cmd/cli/plugin/add.go
serverstatusrequest/process.go
test/comparisons.go
test/fake_namespace.go
test/fake_dynamic.go
discovery/helper_test.go
controller/backupstoragelocation_controller.go
controller/backup_deletion_controller.go
controller/backupstoragelocation_controller_test.go
controller/restore_controller.go
controller/backup_sync_controller.go
controller/schedule_controller.go
controller/pod_volume_restore_controller.go
controller/download_request_controller.go
controller/backup_deletion_controller_test.go
controller/restic_repository_controller.go
controller/schedule_controller_test.go
controller/pod_volume_backup_controller.go
controller/restore_controller_test.go
controller/backup_controller.go
backup/backup.go
restore/merge_service_account_test.go
restore/merge_service_account.go
restore/restore.go
client/dynamic.go
```


That being said, I also don't know if it's worth holding up this PR for - it could go in on the next controller, too, or as its own PR that revisits the controllers.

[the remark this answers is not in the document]
carlisia: I'm ok with this direction.

I would prefer also not to hold up this PR for changing things to a generic helper. I've been thinking thru this as I work on the migration for the `downloadrequest`, so how about this: I'll add a generic helper with that change, then open a separate PR to change this and the other resources/controllers.

carlisia:
@nrb / @ashish-amarnath PTAL!

carlisia:
The reason for this last commit:

When I was working and testing https://github.com/vmware-tanzu/velero/pull/2870 I noticed a change in behavior with the reconciling for the BSL. The reconciling was not kicking in immediately after creating a new BSL. I had more BSL objects in the server this time.

I investigated it and learned that it's expected, since the default # of concurrent workers is 1, so each new request was sitting in the queue waiting for whatever was there before to be processed.

The last commit fixes that by increasing the number of workers that do the processing. The number `10` came from a discussion on the upstream kubebuilder channel, link below, SS here for quick ref. We can change this number, of course, I just don't currently have a better reference, would have to dig deeper if this doesn't seem reasonable.

image

Source:
https://kubernetes.slack.com/archives/CAR30FCJZ/p1598547745132100?thread_ts=1598539574.118700&cid=CAR30FCJZ

nrb:
Going to review this today. I do ask that we hold the generic `Patch` refactor on this controller til after v1.5.0's release, though.

config/samples/velero_v1_serverstatusrequest.yaml
```diff
@@ -0,0 +1,25 @@
+---
```
nrb, line 1: Something I missed in the BSL review was that we got this `samples` directory - is it used for anything other than documentation? We have the `site/content/docs/<version>/api-types/` directory for our other CRD docs, so I'm wondering if we could maybe take these contents and move them over post-generation.

However, I'm also not going to hold up this PR for it.

config/rbac/role.yaml
```diff
@@ -26,3 +26,23 @@ rules:
...
+  resources:
+  - serverstatusrequests
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
nrb, line 40: Given the controller isn't set up for a `watch` verb, do we want to have it listed in the RBAC permissions?

pkg/cmd/server/server.go
```diff
@@ -819,7 +808,13 @@ func (s *server) runControllers(defaultVolumeSnapshotLocations map[string]string
 			s.logger.Infof("Disabling controller: %s", controllerName)
 			delete(enabledControllers, controllerName)
 		} else {
-			s.logger.Fatalf("Invalid value for --disable-controllers flag provided: %s. Valid values are: %s", controllerName, strings.Join(disableControllerList, ","))
+			// maybe it is a runtime type controllers, so attempt to remove that
+			if _, ok := enabledRuntimeControllers[controllerName]; ok {
+				s.logger.Infof("Disabling controller: %s", controllerName)
+				delete(enabledRuntimeControllers, controllerName)
+			} else {
+				s.logger.Fatalf("Invalid value for --disable-controllers flag provided: %s. Valid values are: %s", controllerName, strings.Join(disableControllerList, ","))
```
ashish-amarnath, line 816: Is this an error for which we want to halt execution?

Also, this happens only if `controllerName ` is not in `enabledControllers`.
Suggest making a separate function for the `for` loop from Line 806 so that we can have unit tests for this.

internal/velero/serverstatusrequest.go
```diff
@@ -0,0 +1,73 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	"github.com/vmware-tanzu/velero/pkg/buildinfo"
+	"github.com/vmware-tanzu/velero/pkg/plugin/framework"
+)
+
+// ServerStatus holds information for retrieving installed
+// plugins and for updating the ServerStatusRequest timestamp.
+type ServerStatus struct {
```
ashish-amarnath, line 34: This struct has handles to get `ServerStatus` info. But the name seems to suggest that this is `ServerStatus`.

internal/velero/serverstatusrequest.go
```diff
@@ -0,0 +1,73 @@
...
+	Clock          clock.Clock
+}
+
+// PatchStatusProcessed patches status fields, including loading the plugin info, and updates
+// the ServerStatusRequest.Status.Phase to ServerStatusRequestPhaseProcessed.
+func (s *ServerStatus) PatchStatusProcessed(kbClient client.Client, req *velerov1api.ServerStatusRequest, ctx context.Context) error {
+	statusPatch := client.MergeFrom(req.DeepCopyObject())
+	req.Status.ServerVersion = buildinfo.Version
+	req.Status.Phase = velerov1api.ServerStatusRequestPhaseProcessed
+	req.Status.ProcessedTimestamp = &metav1.Time{Time: s.Clock.Now()}
```
ashish-amarnath, line 45: Why not just use time https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#Now

pkg/controller/server_status_request_controller_test.go
```diff
@@ -0,0 +1,270 @@
...
+	AfterEach(func() {})
+
+	It("Should successfully patch a server status request object status phase", func() {
+		// now will be used to set the fake clock's time; capture
+		// it here so it can be referenced in the test case defs.
+		now, err := time.Parse(time.RFC1123, time.RFC1123)
+		Expect(err).To(BeNil())
+		now = now.Local()
+
+		tests := []struct {
```
ashish-amarnath, line 56: suggest adding a `name` field to this testCase struct to use for naming testcases. Comments in each testCase struct instance can be the `name` for the testCase.
This field can also be used to report test failures.

internal/velero/serverstatusrequest.go
ashish-amarnath: To unit test the deletion of expired things, we can set the `ProcessedTimeStamp` to an earlier time in the unit test case data.
This field will be updated only once when the request is still `New`.

pkg/controller/server_status_request_controller.go
```diff
@@ -17,104 +17,105 @@ limitations under the License.
...
+	statusRequest := &velerov1api.ServerStatusRequest{}
+	if err := r.Client.Get(r.Ctx, req.NamespacedName, statusRequest); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.WithError(err).Error("ServerStatusRequest not found")
+			return ctrl.Result{}, nil
+		}
+
+		log.WithError(err).Error("Error getting ServerStatusRequest")
+		// Error reading the object - requeue the request.
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
carlisia: Doesn't the requeue happen whenever an error is returned? I think I read that in the docs, want to double check before I make the change.

pkg/controller/server_status_request_controller.go
```diff
@@ -17,104 +17,105 @@ limitations under the License.
...
+		"controller":          "serverstatusrequest",
+		"serverStatusRequest": req.NamespacedName,
+		"phase":               statusRequest.Status.Phase,
+	})
+
+	switch statusRequest.Status.Phase {
+	case "", velerov1api.ServerStatusRequestPhaseNew:
+		log.Info("Processing new ServerStatusRequest")
+
+		if err := r.ServerStatus.PatchStatusProcessed(r.Client, statusRequest, r.Ctx); err != nil {
```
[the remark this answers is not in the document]
carlisia: I'm going to refactor this in a separate PR to use the capi path function.

config/rbac/role.yaml
carlisia: I will be removing this file in a separate PR, we don't need it.

config/samples/velero_v1_serverstatusrequest.yaml
carlisia: Good point. These have no function other than to server as an example, and it's generated by kubebuilder (empty) so it's sorta expected (but again, not needed). Let's punt on this for now.

config/rbac/role.yaml
carlisia: Created a ticket to track clean ups related to the migrations: https://github.com/vmware-tanzu/velero/issues/2884

pkg/cmd/server/server.go
carlisia: > Also, this happens only if controllerName is not in enabledControllers.

That is correct, but I don't understand if you are pointing out an issue.

> Is this an error for which we want to halt execution?
It is the original behavior: https://github.com/vmware-tanzu/velero/blob/976484553062262324cc1304607ea00841262db8/pkg/cmd/server/server.go#L822. We can re-evaluate, but not sure this PR is the place to do it. Added this question to today's community meeting.

carlisia: Logged an issue for this: https://github.com/vmware-tanzu/velero/issues/2889

pkg/controller/server_status_request_controller.go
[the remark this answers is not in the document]
carlisia: Double checked that requeuing happens automatically when err is returned (with an exponential backoff). I think the comment is unnecessary, removing it.

pkg/controller/server_status_request_controller.go
```diff
@@ -17,104 +17,105 @@ limitations under the License.
...
+		expiration := statusRequest.Status.ProcessedTimestamp.Add(ttl)
+		if expiration.After(r.ServerStatus.Clock.Now()) {
+			log.Debug("ServerStatusRequest has not expired")
+			return ctrl.Result{RequeueAfter: statusRequestResyncPeriod}, nil
+		}
+
+		log.Debug("ServerStatusRequest has expired, deleting it")
+		if err := r.Client.Delete(r.Ctx, statusRequest); err != nil {
+			log.WithError(err).Error("Unable to delete the request")
+			return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
carlisia: Yes. Fixing.

pkg/controller/server_status_request_controller.go
```diff
@@ -17,104 +17,105 @@ limitations under the License.
...
+		"phase":               statusRequest.Status.Phase,
+	})
+
+	switch statusRequest.Status.Phase {
+	case "", velerov1api.ServerStatusRequestPhaseNew:
+		log.Info("Processing new ServerStatusRequest")
+
+		if err := r.ServerStatus.PatchStatusProcessed(r.Client, statusRequest, r.Ctx); err != nil {
+			log.WithError(err).Error("Unable to update the request")
+			return ctrl.Result{RequeueAfter: statusRequestResyncPeriod}, err
```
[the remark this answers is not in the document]
carlisia: It will be retried whenever an error is returned.

pkg/controller/server_status_request_controller.go
```diff
@@ -17,104 +17,105 @@ limitations under the License.
...
+		}
+	case velerov1api.ServerStatusRequestPhaseProcessed:
+		log.Debug("Checking whether ServerStatusRequest has expired")
+		expiration := statusRequest.Status.ProcessedTimestamp.Add(ttl)
+		if expiration.After(r.ServerStatus.Clock.Now()) {
+			log.Debug("ServerStatusRequest has not expired")
+			return ctrl.Result{RequeueAfter: statusRequestResyncPeriod}, nil
```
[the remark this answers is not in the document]
carlisia: If this `return` is removed it will move on to ln 100 and make a call to `Delete` so no? What am I missing?

internal/velero/serverstatusrequest.go
carlisia: Renaming to `ServerStatus`.

pkg/controller/server_status_request_controller_test.go
carlisia: We are not using the `testing.T` package for this test, we are using ginkgo and gomega. So, it's not possible to do `t.Run(tt.name, func(t *testing.T) {`.

I don't know of any other way to inject the name into test failures. And if we are not using the name in test failures, I don't think we should have a `name` variable but instead the comments are applicable here.

Opinions?

nrb: Since we're mixing patterns here (BDD and table-driven tests), I think we'd want to look at [Ginkgo's table driven test extension](https://onsi.github.io/ginkgo/#table-driven-tests) which provides an `Entry` construct for labeling the tests.

We can also use [`It`](https://onsi.github.io/ginkgo/#individual-specs-it) blocks for naming individual cases.

I think either of these would get us printable output, which would be helpful in debugging.

Admittedly, though, I haven't used Ginkgo in practice, so I'm basing this solely on reading the docs. They don't seem to set up table-driven tests as much, but rather lean on describing individual tests cases in full.

internal/velero/serverstatusrequest.go
carlisia: > Why not just use time https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#Now

I used what was originally used: https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/server_status_request_controller.go#L61 

Do you think this should be changed?

pkg/controller/server_status_request_controller_test.go
carlisia: I'll add this as something to investigate/add in the "clean up" issue for the kubebuilder stuff, so all controllers get done if there's something to change.

nrb: In the interest of time, I'm ok with that.

pkg/cmd/server/server.go
ashish-amarnath: The other point I was trying to make is that we are halting executions only when the`if` on line 812 is `false` and there is a misconfiguration in the `enabledRuntimeControllers`.

We should be halting execution when there is a misconfiguration in either "kinds" of controllers and not just one.

pkg/controller/server_status_request_controller.go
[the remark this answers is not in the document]
ashish-amarnath: thanks for checking!

pkg/controller/server_status_request_controller.go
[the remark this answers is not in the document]
ashish-amarnath: I was referring to the `RequeueAfter: statusRequestResyncPeriod`. If there is an auto-retry w/ exponential backoff. I think we should not be setting `RequeueAfter:`

pkg/controller/server_status_request_controller.go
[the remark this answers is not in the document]
ashish-amarnath: You are right.

internal/velero/serverstatusrequest.go
ashish-amarnath: suggest: `ServerStatusInfoGetter`

internal/velero/serverstatusrequest.go
ashish-amarnath: we can clean this up in a follow-up.

pkg/cmd/server/server.go
carlisia: >  only when the if on line 812 is false 

I don't think so.. It fails if either ln 807 is false OR ln 812 if false.

If there is a match with the passed in config, we can move on and check the next input. We only care to keep checking the second list if a passed in input wasn't found in the first list. 

Am I getting this wrong?

ashish-amarnath: Lines 811-817 are in the `else` block if the `if` at line 807.
So line 816, which has the `s.logger.Fatalf` will be executed only if the condition on line 807 is `false`

carlisia: Right. Why would it ever execute if the condition on line 807 is true?

ashish-amarnath: that's the point. We want to halt execution if there is any misconfiguration in the enabled/disabled controllers, not just for "runtime" controllers

ashish-amarnath: I see it now!

Issue #2824, opened 2020-08-14
"Latest" docs redirects no longer work


nrb:
When navigating to docs links such as https://velero.io/docs/start-contributing/, we get a 404. This used to work due to a Jekyll-specific redirect implementation, but now that we've switch to Hugo, we'll need to update the redirection mechanism.

PR #2839, opened 2020-08-19
add index files to api types folder

a-mccarthy:

Links to /api-types are broken on the site because there is no _index.md file. This adds an index file to the main and 1.4 docs.

Looks like this folder has a README file, do we want to keep that or have the _index files?


site/content/docs/v1.4/api-types/_index.md
```diff
@@ -0,0 +1,20 @@
+---
+version: v1.4
```
nrb, line 2: Given this is versioned, do we need to update our `make gen-docs` script?

ashish-amarnath:
Is this also causing https://github.com/vmware-tanzu/velero/issues/2824 ?

nrb:
@ashish-amarnath I don't think so. #2824 was working because there was a specific Jekyll redirect mapping file in place.

carlisia:
This lgtm but there are requests/questions, so leaving it open for now.

a-mccarthy:
So i did some more digging on this, and made some updates. There is a check in `/layouts/docs.html` that will redirect anything in the `/docs` folder to the latest version  if the `version` front matter is not set. 

What i think is happening here is that we set the `version` front matter  in the `_index.md` file of the version directory, main/_index.md, but that is not being passed to the subfolder `/main/api-types`. Hugo expects there to be a `_index.md` file in all `/content` directories. It treats everything at the same directly level in `/content` as a Section, https://gohugo.io/content-management/sections/. So we need a way to pass the front matter to all sub-sections, Hugo cascade front matter: https://gohugo.io/content-management/front-matter/#front-matter-cascade

@nrb With using the cascade frontmatter, i dont think we'd have to update the gen-docs script, since adding the cascade notation doesn't change the `version` front matter in the main/_index.md file, and the gen-docs script find/replaces the `version` number already. Does that make sense?

site/content/docs/v1.4/api-types/_index.md
a-mccarthy: see my [other comment](https://github.com/vmware-tanzu/velero/pull/2839#issuecomment-676824233), i dont think we need to make a change

PR #2855, opened 2020-08-22
📖  update restore api types with init container restorehooks

ashish-amarnath:

### DIFF
--- a/site/content/docs/main/api-types/restore.md
+++ b/site/content/docs/main/api-types/restore.md
@@ -73,6 +73,59 @@ spec:
   # to restore from. If specified, and BackupName is empty, Velero will
   # restore from the most recent successful backup created from this schedule.
   scheduleName: my-scheduled-backup-name
+  # Actions to perform during or post restore. The only hook currently supported is
+  # adding an init container to a pod before it can be restored. Optional.
+  hooks:
+    # Array of hooks that are applicable to specific resources. Optional.
+    resources:
+    # Name is the name of this hook.
+    - name: restore-hook-1
+      # Array of namespaces to which this hook applies. If unspecified, the hook applies to all
+      # namespaces. Optional.
+      includedNamespaces:
+      - ns1
+      # Array of namespaces to which this hook does not apply. Optional.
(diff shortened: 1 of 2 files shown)


nrb:
Given we're going to try to get #2804 in for v1.5.0, do you want to hold this and modify it once that goes in?

ashish-amarnath:
I think we can get the docs for what's already merged and #2804 going in parallel.
Considering that PR is not being actively worked on or needs changes, we should merge this.

site/content/docs/main/api-types/restore.md
```diff
@@ -73,6 +73,55 @@ spec:
...
+      labelSelector:
+        matchLabels:
+          app: velero
+          component: server
+      # An array of hooks to run during or after restores. Currently only "init" hooks are supported.
+      postHooks:
+      # The type of the hook. This must be "init".
+      - init:
+          # An array of container specs to be added as init containers to pods to which this hook applies to.
+          initContainers:.
```
zubron, line 106:
```suggestion
          initContainers:
```

site/content/docs/main/api-types/restore.md
```diff
@@ -73,6 +73,55 @@ spec:
...
+          - name: restore-hook-init1
+            image: alpine:latest
+            volumeMounts:
+            - mountPath: /restores/pvc1-vm
+              name: pvc1-vm
```
zubron, line 111: These volume names come from the original pod spec that is being restored, right? Would it be worth adding an comment above to explain that?

site/content/docs/main/api-types/restore.md
ashish-amarnath: fixed

site/content/docs/main/api-types/restore.md
ashish-amarnath: added comments.

nrb:
@zubron Thanks for testing!

You make a valid point @ashish-amarnath that this works now and doesn't need to wait. Let's get it in and update the file later once inline hooks are there.

Issue #1350, opened 2019-04-04
Race condition when passing AdditionalItems from restore item action


sseago:
**What steps did you take and what happened:**
We have a restore plugin which returns resources via `AdditionalItems` when those items must be restored before the current item. This happens when there is a dependency relationship between two items of the same resource type. In our particular case, for the `imagestreamtags.image.openshift.io` resource, in some cases some imagestream tags reference other imagestreamtags, and a restore for `imageStreamTagA` will fail if it references `imageStreamTagB` which does not exist. The plugin returns `imageStreamTagB` in the `AdditionalItems` slice.

In Restore.go, after running Execute on the plugin, for any `AdditionalItems` returned, `restoreItem` is called on each one.

From looking at the logs, everything seems to be happening in the right order. The plugin action runs for the resource that references the other imagestream tag, it passes the AdditionalItem reference, and then restore is called on that resource, including calling its restore plugins. However, when I look at the target namespace in the cluster after restore, I see the returned AdditionalItem resource, but I am not seeing the first resource restored at all. I know that if we attempt to restore an imagestreamtag with a non-existent reference, it will fail to restore, which is the expected behavior (i.e. if we have an alias tag for alpine:3.x which is supposed to track alpine 3.2, if 3.2 does not exist, the tracking tag will not be restored).

In this case, however, what seems to be happening is that `restoreItem` for the referenced resource returns before the resource is fully created in the cluster, so when the `resourceClient.Create` call is made, the referenced resource does not yet exist, according to the cluster.

I just modified my local velero checkout to add `time.Sleep(10 * time.Second)` immediately following the restoreItem call (linked below), and everything works as expected.
https://github.com/heptio/velero/blob/master/pkg/restore/restore.go#L938

**What did you expect to happen:**
My expectation was that both of the resources would be restored correctly.

**The output of the following commands will help us better understand what's going on**:
(Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)

(I don't really think anything in my logs will give any more useful information than is already in the description above)

**Anything else you would like to add:**
I'm not sure exactly how to resolve this. If there was some "after restore plugin action", then for cases like this, I could have a plugin that blocked until the resource appeared in the cluster. In this case, on the returned `AdditionalItems` resource, the after restore plugin would wait until it was created to return, that way the initial imagestream tag which references it wouldn't actually have its `resourceClient.Create` call made until the cluster initialization of the resource was done.

**Environment:**

[...]

PR #2867, opened 2020-08-26
Design doc for RestoreItemAction wait for AdditionalItems to be ready

sseago:
Design for https://github.com/vmware-tanzu/velero/issues/1350

### DIFF
--- a/design/wait-for-additional-items.md
+++ b/design/wait-for-additional-items.md
@@ -1,0 +1,251 @@
+# Wait for AdditionalItems to be ready on Restore
+
+When a velero `RestoreItemAction` plugin returns a list of resources
+via `AdditionalItems`, velero restores these resources before
+restoring the current resource. There is a race condition here, as it
+is possible that after running the restore on these returned items,
+the current item's restore might execute before the additional items
+are available. Depending on the nature of the dependency between the
+current item and the additional items, this could cause the restore of
+the current item to fail.
+
+## Goals
+
+- Enable Velero to ensure that Additional items returned by a restore
+  plugin's `Execute` func are ready before restoring the current item
+- Augment the RestoreItemAction plugin interface to allow the plugins
+  to determine when an additional item is ready, since doing so
+  requires knowledge specific to the resource type.
+
+## Background
+
+Because Velero does not wait after restoring additional items to
+restore the current item, in some cases the current item restore will
+fail if the additional items are not yet ready. Velero (and the
+`RestoreItemAction` plugins) need to implement this "wait until ready"
+functionality.
+
+## High-Level Design
+
+After each RestoreItemAction execute call (and following restore of
+any returned additional items) , we need to wait for these returned
+Additional Items to be ready before restoring the current item. In
+order to do this, we also need to extend the `RestoreItemActionExecuteOutput`
+struct to allow the plugin which returned additional items to
+determine whether they are ready.
+
+## Detailed Design
+
+### `restoreItem` Changes
+
+When each `RestoreItemAction` `Execute()` call returns, the
+`RestoreItemActionExecuteOutput` struct contains a slice of
+`AdditionalItems` which must be restored before this item can be
+restored. After restoring these items, Velero needs to be able to wait
+for them to be ready before moving on to the next item. Right after
+looping over the additional items at
+https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L960-L991
+
+we still have a reference to the additional items (`GroupResource` and
+namespaced name), as well as a reference to the `RestoreItemAction`
+plugin which required it.
+
+At this point, if the `RestoreItemActionExecuteOutput` includes a
+non-nil `AdditionalItemsReadyFunc` we need to call a func similar to
+`crdAvailable` which we will call `itemsAvailable`
+https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L623
+This func should also be defined within restore.go
+
+Instead of the one minute CRD timeout, we'll use a timeout specific to
+waiting for additional items. There will be a new field added to
+serverConfig, `additionalItemsReadyTimeout`, with a
+`defaultAdditionalItemsReadyTimeout` const set to 10 minutes. In addition,
+each plugin will be able to define an override for the global
+server-level value, which will be added as another optional field in
+the `RestoreItemActionExecuteOutput` struct. Instead of the
+`IsUnstructuredCRDReady` call, we'll call the returned
+`AdditionalItemsReadyFunc` passing in the same `AdditionalItems` slice
+as an argument (with items which failed to restore filtered out). If
+this func returns an error, then `itemsAvailable` will
+propagate the error, and `restoreItem` will handle it the same way it
+handles an error return on restoring an additional item. If the
+timeout is reached without ready returning true, velero will continue
+on to attempt restore of the current item.
+
+### `RestoreItemActionExecuteOutput` changes
+
+Two new fields will be added to `RestoreItemActionExecuteOutput`, both
+optional. `AdditionalItemsReadyTimeout`, if specified, will override
+`serverConfig.additionalItemsReadyTimeout`. If the new func field
+`AdditionalItemsReadyFunc` is non-nil, then `restoreItem` will call
+`itemsAvailable` which will invoke the plugin func
+`AdditionalItemsReadyFunc` and wait until the func returns true or the
+timeout is reached. If `AdditionalItemsReadyFunc` is nil (the default
+case), then current velero behavior will be followed. Existing plugins
+which do not need to signal to wait for `AdditionalItems` won't need
+to change their `Execute()` functions.
+
+In addition, a new func, `WithItemsWait(readyFunc *func)` will
+be added to `RestoreItemActionExecuteOutput` similar to
+`WithoutRestore()` which will set `AdditionalItemsReadyFunc` to
+`readyfunc`. This will allow a plugin to include waiting for
+AdditionalItems like this:
+```
+func AreItemsReady (restore *api.Restore, additionalItems []ResourceIdentifier) (bool, error) {
+	...
+	return true, nil
+}
+func (p *RestorePlugin) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
+	...
+	return velero.NewRestoreItemActionExecuteOutput(input.Item).WithItemsWait(AreItemsReady), nil
+}
+
+```
+
+
+```
+// RestoreItemActionExecuteOutput contains the output variables for the ItemAction's Execution function.
+type RestoreItemActionExecuteOutput struct {
+	// UpdatedItem is the item being restored mutated by ItemAction.
+	UpdatedItem runtime.Unstructured
+
+	// AdditionalItems is a list of additional related items that should
+	// be restored.
+	AdditionalItems []ResourceIdentifier
+
+	// SkipRestore tells velero to stop executing further actions
+	// on this item, and skip the restore step. When this field's
+	// value is true, AdditionalItems will be ignored.
+	SkipRestore bool
+
+	// AdditionalItemsReadyFunc is a func which returns true if
+	// the additionalItems passed into the func are
+	// ready/available. A nil value for this func means that
+	// velero will not wait for the items to be ready before
+	// attempting to restore the current item.
+	AdditionalItemsReadyFunc func(restore *api.Restore, []ResourceIdentifier) (bool, error)
+
+	// AdditionalItemsReadyTimeout will override serverConfig.additionalItemsReadyTimeout
+	// if specified. This value specifies how long velero will wait
+	// for additional items to be ready before moving on.
+	AdditionalItemsReadyTimeout *time.Duration
+}
+
+// WithoutRestore returns SkipRestore for RestoreItemActionExecuteOutput
+func (r *RestoreItemActionExecuteOutput) WithItemsWait(
+	readyFunc func(*api.Restore, []ResourceIdentifier)
+) *RestoreItemActionExecuteOutput {
+	r.AdditionalItemsReadyFunc = readyFunc
+	return r
+}
+
+```
+
+### Earlier iteration (no longer the current implementation plan)
+
+What follows is the first iteration of the design. Everything from
+here is superseded by the content above. The options below require
+either breaking backwards compatibility or dealing with runtime
+casting and optional interfaces. Adding the func pointer to
+`RestoreItemActionExecuteOutput` resolves the problem without
+requiring either.
+
+#### `RestoreItemActionExecuteOutput` changes
+
+A new boolean field will be added to
+`RestoreItemActionExecuteOutput`. If `WaitForAdditionalItems` is true,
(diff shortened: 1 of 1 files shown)


design/wait-for-additional-items.md
```diff
@@ -0,0 +1,169 @@
...
+
+- Enable Velero to ensure that Additional items returned by a restore
+  plugin's `Execute` func are ready before restoring the current item
+- Augment the RestoreItemAction plugin interface to allow the plugins
+  to determine when an additional item is ready, since doing so
+  requires knowledge specific to the resource type.
+
+## Background
+
+Because Velero does not wait after restoring additoonal items to
```
alaypatel07, line 22: s/additioonal/additional

sseago:
I'm going to update this shortly with a third implementation option which should get rid of the need for making interface changes. Instead of adding a new func to either an existing or new interface, we could replace the `WaitForAdditionalItems` bool in my proposed changes to `RestoreItemActionExecuteOutput` with `AdditionalItemsReadyFunc` which would be a pointer to a func. If a plugin ignores it (as would any pre-existing plugins), it remains nil and velero doesn't attempt to wait. If a plugin needs to wait for AdditionalItems to be ready, instead of setting an entry in the return struct to 'true', it will return a func pointer, which will be the func that velero's `restoreItem` func will call in its wait method rather than a named interface func. This will remove the need for existing plugins to have to make changes here without introducing optional interfaces.

nrb:
Thanks @sseago! I know you're working on this, so I'm likely to review after the v1.5.0 release and you're a little further along. I'll also be on vacation from Sept 4 to Sept 14th, so it's probably going to be after that, just to set expectations.

One question I do have - are you/your team able to sign up for implementation of this once approved?

sseago:
@nrb Yes, my plan was to put together a PR to implement this once we got sign-off on the design. I expect one more update to the design before it's ready to review -- to propose a third alternative which I think will eliminate both the backwards-compatibilty problem and the need to define a new optional interface, as I mention in the above comment. I may get to it today, or tomorrow at the latest. Either way I assume your current priority is getting 1.5.0 out the door.

nrb (commented):
So far I think this is reasonable.

I like the idea of making the `AdditionalItemsReadyFunc` a pointer so that not all RestoreItemAction plugins have to implement it.

I do still have a few questions on specific areas, and would like to hear from the rest of @vmware-tanzu/velero-maintainers

design/wait-for-additional-items.md
```diff
@@ -0,0 +1,229 @@
...
+restored. After restoring these items, Velero needs to be able to wait
+for them to be ready before moving on to the next item. Right after
+looping over the additional items at
+https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L960-L991
+
+we still have a reference to the additional items (`GroupResource` and
+namespaced name), as well as a reference to the `RestoreItemAction`
+plugin which required it.
+
+At this point, if the `RestoreItemActionExecuteOutput` includes a
```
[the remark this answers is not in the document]
sseago: Yes. I was thinking of something very similar to `crdAvailable`, defined in the same file. Something like this (logging, error checking removed to focus on func operation):
```
// crdAvailable waits for additionalItems to be available for use before letting the restore continue.
func (ctx *restoreContext) itemsAvailable(func readyFunc, items []ResourceIdentifier, crdClient client.Dynamic) (bool, error) {
	var available bool
	err := wait.PollImmediate(time.Second, ctx.resourceTerminatingTimeout, func() (bool, error) {
		available, err = readyFunc(ctx.restore, items)
		if err != nil {
			return true, err
		}
		return available, nil
	})
	return available, err
}
```

design/wait-for-additional-items.md
```diff
@@ -0,0 +1,229 @@
...
+func (p *RestorePlugin) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
+	...
+	return velero.NewRestoreItemActionExecuteOutput(input.Item).WithItemsWait(AreItemsReady), nil
+}
+
+```
+
+
+```
+// RestoreItemActionExecuteOutput contains the output variables for the ItemAction's Execution function.
```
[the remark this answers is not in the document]
sseago: I didn't include it here since it's not part of the struct itself, but I can add it to the doc. It would look like this:
```
// WithoutRestore returns SkipRestore for RestoreItemActionExecuteOutput
func (r *RestoreItemActionExecuteOutput) WithoutItemsWait(readyFunc func( *api.Restore, []ResourceIdentifier)) *RestoreItemActionExecuteOutput {
        r.AdditionalItemsReadyFunc = readyFunc
        return r
}
```

design/wait-for-additional-items.md
```diff
@@ -0,0 +1,229 @@
...
+A new func field will be added to `RestoreItemActionExecuteOutput`. If
+`AdditionalItemsReadyFunc` is non-nil, then `restoreItem` will call
+`itemsAvailable` which will invoke the plugin func
+`AdditionalItemsReadyFunc` and wait until the func returns true or the
+timeout is reached. If `AdditionalItemsReadyFunc` is nil (the default
+case), then current velero behavior will be followed. Existing plugins
+which do not need to signal to wait for `AdditionalItems` won't need
+to change their `Execute()` functions.
+
+In addition, a new func, `WithItemsWait(readyFunc *func)` will
```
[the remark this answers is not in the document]
sseago: `WithItemsWait` will be defined in restore_item_action.go. To use the example from the design doc, here's how that function would be used (with a custom plugin-defined `AreItemsReady` function. The following code would be in the custom plugin code:
```
func AreItemsReady (restore *api.Restore, additionalItems []ResourceIdentifier) (bool, error) {
	...
	return true, nil
}
func (p *RestorePlugin) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
	...
	return velero.NewRestoreItemActionExecuteOutput(input.Item).WithItemsWait(AreItemsReady), nil
}
```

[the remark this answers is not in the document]
nrb: Thanks! I was looking for where the definition would be versus the example usage, but you provided it in a reply to another comment.

design/wait-for-additional-items.md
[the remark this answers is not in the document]
nrb: I think it would be worth including as it's part of the change - I was wondering where the actually implementation would live. It wasn't clear to me from the text, hence having 2 questions about it :)

sseago:
Updated with the `WithItemsWait` func definition added to the document.

nrb:
@vmware-tanzu/velero-maintainers Please give this a review.

design/wait-for-additional-items.md
```diff
@@ -0,0 +1,239 @@
...
+handles an error return on restoring an additional item. If the
+timeout is reached without ready returning true, velero will continue
+on to attempt restore of the current item.
+
+### `RestoreItemActionExecuteOutput` changes
+
+A new func field will be added to `RestoreItemActionExecuteOutput`. If
+`AdditionalItemsReadyFunc` is non-nil, then `restoreItem` will call
+`itemsAvailable` which will invoke the plugin func
+`AdditionalItemsReadyFunc` and wait until the func returns true or the
```
ashish-amarnath, line 75: Would it be simpler to have a restore level option of `--sync-restore-additional-items`, a new field in the `RestoreSpec`, which if set, we would call a `waitForAdditionalItems` function after finishing the `for` loop at https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L1026-L1056
The `waitForAdditionalItems` would use the `wait.PollImmediate` API to wait for the additional items to be ready. The time out for this can also be made configurable either using another restore level option, say, `--sync-restore-additional-items-timeout`.

The simplicity coming from:
1. With a new release of Velero, after this is implemented, all RestoreItemActions will be able to take advantage of this feature without having to make any changes to their plugins.
1. Current `RestoreItemAction` would not require to be updated. Similar to the limitation of an earlier iteration of the design.
1. The proposed design here, would require a release of Velero, with the updated `RestoreItemActionOutput` API type before plugins are able to release a version using the new fields in the `RestoreItemActionOutput`
1. Implementation of the `AdditionalItemsReadyFunc` would be very identical across all RestoreItemAction plugins. So, instead of having the same implementation across many plugins, this function may be implemented as the `waitForAdditionalItems` and unexported function in the restore package of core Velero.

design/wait-for-additional-items.md
```diff
@@ -0,0 +1,239 @@
...
+namespaced name), as well as a reference to the `RestoreItemAction`
+plugin which required it.
+
+At this point, if the `RestoreItemActionExecuteOutput` includes a
+non-nil `AdditionalItemsReadyFunc` we need to call a func similar to
+`crdAvailable` which we will call `itemsAvailable`
+https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L623
+This func should also be defined within restore.go
+
+Instead of the one minute CRD timeout, we'll use
```
ashish-amarnath, line 59: `resourceTerminatingTimeout` is not meant to be used for this purpose. Suggest creating a different constant/config for this.

ashish-amarnath (commented):
@sseago Thanks for this proposal and apologies for not looking at this sooner.
I added a couple of comments. PTAL. Happy to discuss further either in the PR or in a separate design meeting. I can schedule a design meeting if that is preferred.

design/wait-for-additional-items.md
alaypatel07: @ashish-amarnath interesting thought but I have a question:

In the solution you proposed, how would `waitForAdditionalItems` be implemented in a generic way for all resources?

The concept of "Readiness" may be different for different resources (for example for a pod it may be the container ready status, for a secret, it might be the meager existence and for a custom resource it might be a custom field in the status), and we need the plugin for a particular resource to defined what Readiness means for the resource it needs to wait.

ashish-amarnath: @alaypatel07 Yes, `waitForAdditionalItems` is generic across all resources. This function would wait for the restored resource to be present in the response to the `Get` from the API server.
I interpreted "ready" as "existence". It appeared that was the race that issue #1350 was referring to, that, restored resources were not immediately available.

sseago: @ashish-amarnath So the basic reason for the plugin-specific function was that in discussing it with @nrb we weren't sure whether it was appropriate to attempt "is it ready?" quite as simply as "the Get API returns true". If we combined your proposal with keeping `AdditionalItemsReadyFunc` as an optional field where the generic solution fails, though, that might be workable.

sseago: Another possible concern was that since most resource types don't actually need this waiting -- in most cases, the Kubernetes model of eventual consistency means that it's OK to restore the current item even if the `additionalItems` aren't available yet, so this would be needless waiting. Having the plugin signal when to wait makes this the exception rather than the rule. Since we weren't using a generic Get call for "is it ready?" we were able to combine "this plugin waits for additional items" with "this is how we wait" in one field. If we wanted to have the generic function be the default with `AdditionalItemsReadyFunc` as an override for how we wait, we could implement optional waiting with an additional boolean field here. That way we'd only wait when we're dealing with a resource which will fail restore if it's not yet there.  Then, once we have this (either as-is in the proposal, or with this new boolean field to allow for plugins not to have to redefine the method when the generic `Get` approach is good enough), we may not need the `--sync-restore-additional-items` option, although we could provide it -- if it's true, we wait for items when the plugin indicates we should, and if its' false, we never wait. But at this point, I don't know whether that option provides value. The point of having this in the first place is that with certain plugins/resources, not waiting will fail migrations, and with others waiting doesn't help. It feels like a plugin-specific setting rather than a restore-specific settings.

design/wait-for-additional-items.md
sseago: I agree that a different method of determining it would be better. I didn't include that in the design as I wasn't sure what made more sense. We can include it as a restore option as suggested above, although we'll need to provide a sensible default (perhaps via a new const). We could also allow plugin-level overriding, but it's not obvious whether or not we need that additional complexity and if we add it, what the priority would be for using. Perhaps first priority is to use the value set on the restore CR, if provided. Then plugin-level value (in the plugin `RestoreItemExecuteOutput` struct), and if both are empty, use something defined in a const.

design/wait-for-additional-items.md
nrb: @ashish-amarnath Just waiting for `GET` isn't quite enough for readiness all the time. Readiness means the thing exists and is useful.

As an example, CRDs can exist before CRDs can be made. Readiness is determined by examining `status.conditions`. [See this logic](https://github.com/vmware-tanzu/velero/blob/dbc83af77b0c0a36a27ac1646b6504fa78bedaba/pkg/util/kube/utils.go#L202-L209)

As another example, a `Pod` can exist but be in `CrashLoopBackOff`, and therefore not ready.

design/wait-for-additional-items.md
zubron: I think I'm missing something but what would it look like for the timeout value to be configurable at the plugin level? Is it possible under the current plugin model for specific configuration to be passed in when the plugins are being added? Or did you mean that different plugins would define how long they would wait but it would be a fixed duration?

sseago: So one way to do it which avoids having to change any existing plugins that don't need the new functionality is to add an additional (optional) timeout field to the `RestoreItemActionExecuteOutput` struct. When the `Execute` func returns, it could add this from a const defined within the plugin implementation. We already have one field we're adding to this struct to define the "wait for ready" func, and based on recent review comments, we may be adding another bool field which velero will use to decide whether to wait at all for this plugin (instead of overloading the func return value for this, to allow for a default "does `Get` return it?" implementation. That's what I was considering for the next update.

design/wait-for-additional-items.md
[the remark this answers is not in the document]
ashish-amarnath: This comment can be resolved?

design/wait-for-additional-items.md
ashish-amarnath: Resolving this comment.

design/wait-for-additional-items.md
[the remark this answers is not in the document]
ashish-amarnath: This can also be resolved.

ashish-amarnath (commented):
This LGTM.
The only pending comment is about adding a timeout field in the `RestoreItemActionExecuteOutput` struct.

sseago:
@ashish-amarnath I've updated the timeout section to include a new server-wide value specified in `serverConfig.additonalItemsReadyTimeout` (with a default const), as well as a plugin-level override value as an optional field in `RestoreItemActionExecuteOutput`.

zubron (approved):
Thanks for your time and effort on this, @sseago! I think this proposed design looks good and I don't have any additional questions or comments. Thanks for addressing them all! :+1:

ashish-amarnath (commented):
This LGTM! Thanks for addressing the comments.
Leaving the merge to @nrb

carlisia (commented):
This lgtm. Deferring to @nrb.

nrb (approved):
Thanks for your effort and patience on this @sseago!

Issue #2739, opened 2020-07-21
Persistent Volume with Reclaim Policy Delete gets recreated


joschi36:
**What steps did you take and what happened:**
We did a backup with Velero and without any form of volume backup (restic/volumesnapshot).
When restoring the PersistentVolume got recreated.

**What did you expect to happen:**
We expected that the manifest of the PersistentVolume gets applied as is with the same meta information.

**The output of the following commands will help us better understand what's going on**:
- `velero restore logs joschi-disaster-test`
```
Getting client for /v1, Kind=PersistentVolume" logSource="pkg/restore/restore.go:746" restore=velero/joschi-disaster-test-2020072117012
Dynamically re-provisioning persistent volume because it doesn't have a snapshot and its reclaim policy is Delete." logSource="pkg/restore/restore.go:933" restore=velero/joschi-disaster-test-2020072117012
Dynamically re-provisioning persistent volume because it doesn't have a snapshot and its reclaim policy is Delete." logSource="pkg/restore/restore.go:933" restore=velero/joschi-disaster-test-2020072117012
Dynamically re-provisioning persistent volume because it doesn't have a snapshot and its reclaim policy is Delete." logSource="pkg/restore/restore.go:933" restore=velero/joschi-disaster-test-2020072117012
Dynamically re-provisioning persistent volume because it doesn't have a snapshot and its reclaim policy is Delete." logSource="pkg/restore/restore.go:933" restore=velero/joschi-disaster-test-2020072117012
Dynamically re-provisioning persistent volume because it doesn't have a snapshot and its reclaim policy is Delete." logSource="pkg/restore/restore.go:933" restore=velero/joschi-disaster-test-2020072117012
Restoring resource 'persistentvolumeclaims' into namespace 'loki'" logSource="pkg/restore/restore.go:702" restore=velero/joschi-disaster-test-2020072117012
Getting client for /v1, Kind=PersistentVolumeClaim" logSource="pkg/restore/restore.go:746" restore=velero/joschi-disaster-test-2020072117012
Executing item action for persistentvolumeclaims" logSource="pkg/restore/restore.go:964" restore=velero/joschi-disaster-test-2020072117012
Executing AddPVFromPVCAction" cmd=/velero logSource="pkg/restore/add_pv_from_pvc_action.go:44" pluginName=velero restore=velero/joschi-disaster-test-2020072117012
[...]

PR #2869, opened 2020-08-26
Remove case hasDeleteReclaimPolicy

joschi36:
As we didn't get any information about the issue #2739 we decided to move forward and just remove the `hasDeleteReclaimPolicy `case as we don't see a valid use case. Maybe this behaviour is not intended. We look forward to hearing from maintainer.

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -901,15 +901,8 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 			// return early because we don't want to restore the PV itself, we want to dynamically re-provision it.
 			return warnings, errs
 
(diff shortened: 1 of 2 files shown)


nrb:
@joschi36 I think in most cases, we don't want to restore a volume that has a `Delete` policy and has no snapshot.

This would mean it was a) specifically not snapshotted in Velero and b) specifically listed in Kubernetes to be deleted when no `PersistentVolume` pointing to the provider's volume exists. As such, reprovisioning would be the correct thing to do in most installations. See the [Kubernetes documentation on the [`Delete` policy](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#delete).

If there's an installation that is not conforming to this assumptions, then it seems it would be the exception, rather than the rule, and I'd prefer to not completely remove this case for that.

PR #2870, opened 2020-08-26
Refactor BSL related code

carlisia:
This PR:

- does not attempt to change any logic
- uses the CAPI package to call `patchHelper.Patch`, removing the code to patch resources
- reorganizes and renames structs to better reflect the actual functionality implemented


go.mod
```diff
@@ -21,12 +21,12 @@ require (
 	github.com/kubernetes-csi/external-snapshotter/v2 v2.2.0-rc1
 	github.com/onsi/ginkgo v1.13.0
 	github.com/onsi/gomega v1.10.1
-	github.com/pkg/errors v0.8.1
-	github.com/prometheus/client_golang v1.0.0
+	github.com/pkg/errors v0.9.1
```
nrb, line 24: Was the bump on this version required by another dependency?

carlisia: Yes, the cluster/api uses that version, and we are using the cluster/api `.Patch` function:

https://github.com/vmware-tanzu/velero/pull/2870/files#diff-37aff102a57d3d7b797f152915a6dc16R40

https://github.com/kubernetes-sigs/cluster-api/blob/master/go.mod#L23

carlisia:
Please see this comment as for the reason why I added the last commit:

https://github.com/vmware-tanzu/velero/pull/2838#issuecomment-682161089

nrb:
Going to review this and #2838 today. Let's try to get these 2 into v1.5.0!

internal/velero/storagelocation.go
```diff
@@ -32,21 +31,49 @@ import (
...
 	NewPluginManager func(logrus.FieldLogger) clientmgmt.Manager
 	NewBackupStore   func(*velerov1api.BackupStorageLocation, persistence.ObjectStoreGetter, logrus.FieldLogger) (persistence.BackupStore, error)
 }
 
+// LocationStore contains data for a backup storage location
+// and its associated store, as well as default backup location values.
+type LocationStore struct {
+	location            *velerov1api.BackupStorageLocation
+	backupStore         persistence.BackupStore
+	defaultLocationInfo DefaultBackupLocationInfo
```
ashish-amarnath, line 54: This information will be duplicated in every BSL created. This will require updating every BSL should the `default` BSL change. I think this field can be dropped.

internal/velero/storagelocation.go
```diff
@@ -32,21 +31,49 @@ import (
...
-// StorageLocation holds information for connecting with storage
-// for a backup storage location.
-type StorageLocation struct {
-	Client client.Client
-	Ctx    context.Context
-
+// DefaultBackupLocationInfo holds server default backup storage location information
+type DefaultBackupLocationInfo struct {
 	DefaultStorageLocation          string
 	DefaultStoreValidationFrequency time.Duration
```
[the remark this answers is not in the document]
ashish-amarnath: From this encapsulation, it seems like this is the validation frequency for the default BSL.
However, from L#86, it is the default validation frequency for any BSL.
Suggest making a package variable for storing the default validation frequency for all BSLs and use this field as the validation freq for the default BSL.

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -132,12 +152,15 @@ func (r *BackupStorageLocationReconciler) logReconciledPhase(defaultFound bool,
...
-		log.Warnf("The specified default backup location named %q was not found; for convenience, be sure to create one or make another backup location that is available the default", r.StorageLocation.DefaultStorageLocation)
+		log.Warnf("The specified default backup location named %q was not found; for convenience, be sure to create one or make another backup location that is available the default", r.DefaultBackupLocationInfo.DefaultStorageLocation)
 	}
 }
 
 func (r *BackupStorageLocationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&velerov1api.BackupStorageLocation{}).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: 10,
```
ashish-amarnath, line 163: The BSL controller, in every execution of the `Reconcile` method is validating all BSLs in the cluster based on their validation frequency. So running multiple reconciles at the same time will imply that we are concurrently validating all BSL which might result in unnecessary API calls to the BSL and also to the API server, from the `patch` calls.

To prevent delays in the reconciliation of BSLs, we should be running iterations of the loop on Line 65 in parallel and not run multiple reconciles in parallel.

internal/velero/storagelocation.go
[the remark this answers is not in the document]
nrb: @ashish-amarnath I don't think there is an independent frequency for the default location - all locations use 1 timer.

If you're suggesting that we introduce a separate frequency for the default location, please open an issue so we can track that separately and do a new PR. Let's keep this one focused on the `Patch` refactor.

pkg/controller/backupstoragelocation_controller.go
nrb: This is a good point. As written, this call means [the loop](https://github.com/vmware-tanzu/velero/pull/2870/files#diff-cfcbc9b9778d96e266dd99924a91c426R65) is effectively being called 10 different times in parallel, iterating over each BSL 10 times at once.

So to fan-out, we'd change to take L65-L89 and make most of it the contents of a go routine that would receive an individual BSL, then have some number of concurrent goroutines to process the BSLs.

However, I think that would be an optimization and is not a requirement for shipping this. _Most_ installs aren't going to have tons of BSLs, but this would be a good pattern support in our controllers that do have lots of objects, like backups and restores.

internal/velero/storagelocation.go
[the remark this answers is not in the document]
carlisia: This is the default validation frequency for any/all BSL, which gets set on the server level by default.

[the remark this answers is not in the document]
carlisia: I added documentation to those variables.

internal/velero/storagelocation.go
carlisia: The values for the two fields in this struct are not tied up to any BSL, it's a server level default. I made some changes but kept the `DefaultBackupLocationInfo` and added documentation. When you look at it again, please let me know if you still see any potential problem.

carlisia:
@nrb / @ashish-amarnath this is ready for another review.

I mainly cleaned up the `storagelocation.go` code: removed the `NewLocationStore` constructor since it wasn't buying us anything and was necessitating the use of wrappers around the backupstore (like `IsValid()` and corresponding test). Also made the `IsReadyToValidate` a lot more precise by changing it to a function that only takes the needed arguments, making it a lot more clear what variables are needed inside that function.

pkg/controller/backupstoragelocation_controller.go
carlisia: I removed the `MaxConcurrentReconciles` and created a new issue: https://github.com/vmware-tanzu/velero/issues/2883

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -57,36 +70,50 @@ func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
 		log := r.Log.WithField("controller", "backupstoragelocation").WithField("backupstoragelocation", location.Name)
 
-		if location.Name == r.StorageLocation.DefaultStorageLocation {
+		if location.Name == r.DefaultBackupLocationInfo.StorageLocation {
 			defaultFound = true
 		}
 
-		if !r.StorageLocation.IsReadyToValidate(location, log) {
+		pluginManager := r.NewPluginManager(log)
+		defer pluginManager.CleanupClients()
```
ashish-amarnath, line 78: can this be done outside the `for` loop?

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -57,36 +70,50 @@ func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
-			if err := r.StorageLocation.PatchStatus(location, velerov1api.BackupStorageLocationPhaseAvailable); err != nil {
-				log.WithError(err).Errorf("Error updating backup location phase to %s", velerov1api.BackupStorageLocationPhaseAvailable)
-				continue
-			}
+			location.Status.Phase = velerov1api.BackupStorageLocationPhaseAvailable
+		}
+		location.Status.LastValidationTime = &metav1.Time{Time: time.Now().UTC()}
+		if err := patchHelper.Patch(r.Ctx, location); err != nil {
+			log.WithError(err).Error("Error updating backup location phase")
+			continue
```
ashish-amarnath, line 116: `continue` is not required.

pkg/controller/backupstoragelocation_controller.go
carlisia: Hmmm yeah, good point!

carlisia:
Manually tested this after moving `pluginManager` outside the `for` loop, still works. :)

pkg/controller/backupstoragelocation_controller.go
nrb: 👍

carlisia:
@nrb / @ashish-amarnath this is ready for review.

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -44,49 +57,62 @@ func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
 
+	pluginManager := r.NewPluginManager(log)
+	defer pluginManager.CleanupClients()
+
 	var defaultFound bool
 	var unavailableErrors []string
 	var anyVerified bool
 	for i := range locationList.Items {
 		location := &locationList.Items[i]
 		log := r.Log.WithField("controller", "backupstoragelocation").WithField("backupstoragelocation", location.Name)
```
ashish-amarnath, line 74: can this reuse `log` from line 56?

carlisia: Not if we want to log the location name.

ashish-amarnath: Line 74 is instantiating a new logger in every iteration.
Can we use something like
```
bslLogger := log.WithField("backupstoragelocation", location.Name)
```
reusing `log` from line 56

ashish-amarnath: probably not instantiating a new logger.

pkg/controller/backupstoragelocation_controller.go
```diff
@@ -44,49 +57,62 @@ func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
 
-		anyVerified = true
+		if !storage.IsReadyToValidate(location.Spec.ValidationFrequency, location.Status.LastValidationTime, r.DefaultBackupLocationInfo, log) {
+			log.Debug("Backup location not ready to be validated")
+			continue
+		}
 
-		log.Debug("Verifying backup storage location")
+		// Initialize the patch helper.
+		patchHelper, err := patch.NewHelper(location, r.Client)
```
ashish-amarnath, line 92: We should set `patchHelper.isConditionsSetter` explicitly as it involves code specific to CAPI.
https://github.com/kubernetes-sigs/cluster-api/blob/master/util/patch/patch.go#L155

Though there is this https://github.com/kubernetes-sigs/cluster-api/blob/master/util/patch/patch.go#L64
```
	// Check if the object satisfies the Cluster API conditions contract.
	_, canInterfaceConditions := obj.(conditions.Setter)
```
IMO, it is better to set this to false explicitly.

pkg/controller/backupstoragelocation_controller.go
carlisia: Yes, and `log.WithField("backupstoragelocation", location.Name)` would overwrite the fields in the log.

pkg/controller/backupstoragelocation_controller.go
carlisia: Not sure about this suggestion. What value would this be adding? The check on the capi side will always happen, and it's clutter that whoever is reading this code doesn't need to reason thru, unless there's value to doing this that I'm not seeing.

ashish-amarnath: Running the check on https://github.com/kubernetes-sigs/cluster-api/blob/master/util/patch/patch.go#L64 is not a problem. However, based on whether `isConditionsSetter` is true or false, the code branches taken by `patchHelper.Patch` is different. The code branch taken when `isConditionsSetter` is true, is specific to CAPI which should not be called when patching Velero resource.

carlisia: Right. And we are not setting it. This means that branch will not validate to `true`.

carlisia: In other words, I know what the code does (I've read it), what I am saying is that I don't agree that making this explicit makes this code better, I think it makes it worse, since this is would be unnecessary code added.

PR #2875, opened 2020-08-28
Fix git commands and add dry run mode as default to the tag-release.sh script.

### LINKED ISSUE #2852: Bugs in hack/release-tools/tag-release.sh
Running tag-release.sh for the v1.5.0-beta.1 ran into the following errors:

* `./hack/release-tools/tag-release.sh: line 47: Ensure: command not found`
* Tag failed to push to GitHub repo from script
* Script does not have a "dry-run" mode analogous to the `make release` target's supported `PUBLISH=false` setting.

Script should not error & should create remote tag.

nrb:
Fixes #2852

### DIFF
--- a/hack/release-tools/tag-release.sh
+++ b/hack/release-tools/tag-release.sh
@@ -25,11 +25,15 @@
 # Directory in which the script itself resides, so we can use it for calling programs that are in the same directory.
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
+# Parse out the branch we're on so we can switch back to it at the end of a dry-run, where we delete the tag. Requires git v1.8.1+
+original_branch=$(git symbolic-ref --short HEAD)
+
 function tag_and_push() {
-    echo "Tagging and pushing $VELERO_VERSION"
-    git tag $VELERO_VERSION
+    echo "Tagging $VELERO_VERSION"
+    git tag $VELERO_VERSION || true
     
     if [[ $publish == "TRUE" ]]; then
+        echo "Pushing $VELERO_VERSION"
         git push upstream $VELERO_VERSION
     fi
 }
@@ -109,7 +113,11 @@ if [[ "$VELERO_PATCH" > 0 ]]; then
 
     echo "Now you'll need to cherry-pick any relevant git commits into this release branch."
     echo "Either pause this script with ctrl-z, or open a new terminal window and do the cherry-picking."
-    read -p "Press enter when you're done cherry-picking. THIS WILL MAKE A TAG PUSH THE BRANCH TO UPSTREAM"
+    if [[ $publish == "TRUE" ]]; then
+        read -p "Press enter when you're done cherry-picking. THIS WILL MAKE A TAG PUSH THE BRANCH TO UPSTREAM"
+    else
+        read -p "Press enter when you're done cherry-picking."
+    fi
 
     # TODO can/should we add a way to review the cherry-picked commits before the push?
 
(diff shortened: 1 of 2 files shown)


hack/release-tools/tag-release.sh
```diff
@@ -28,9 +28,19 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
+    
+    if [[ $publish == "TRUE" ]]; then
+        git push upstream $VELERO_VERSION
+    fi
 }
 
+# Default to a dry-run mode
+publish=FALSE
+if [[ "$1" = "publish" ]];
+then
```
ashish-amarnath, line 40: this can go on line 39.
Staying consistent with other `if [[ ... ]]` in the file

ashish-amarnath: same w/ line 83

nrb: 👍  Changed locally, also gonna update docs to reflect the new argument before pushing.

ashish-amarnath:
@nrb looks like there is some merge conflict that needs to be resolved.

hack/release-tools/tag-release.sh
```diff
@@ -99,8 +113,10 @@ if [[ "$VELERO_PATCH" > 0 ]]; then
 
     # TODO can/should we add a way to review the cherry-picked commits before the push?
```
zubron, line 114: Given that the default behaviour for this script is to prepare the branch without pushing, perhaps this comment could be removed as there is now an opportunity to review the cherry-picked commits.

hack/release-tools/tag-release.sh
```diff
@@ -28,9 +28,18 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 function tag_and_push() {
     echo "Tagging and pushing $VELERO_VERSION"
```
zubron, line 29: This comment could be split so that the "pushing" part of it is only printed when `$publish` is true.

hack/release-tools/tag-release.sh
```diff
@@ -99,8 +113,10 @@ if [[ "$VELERO_PATCH" > 0 ]]; then
```
zubron, line 113: I can't specifically add a comment to line 112, but the "push" aspect of the warning at the end of the message now only applies in the case where `$publish` is true so it could be printed only in that case.

hack/release-tools/tag-release.sh
nrb: Good point. I think there's still more to be done for the patch release process, as my original intention was to somehow provide a list of commit hashes or PR numbers and have the script cherry-pick them for the user. This may not be reasonable given potential conflicts, but it's also something I don't really want to fix in this PR.

nrb:
@zubron I think I addressed your comments, and it made me see a few more issues in the script.

hack/release-tools/tag-release.sh
zubron: I agree, that seems out of scope for this PR. In looking at this script earlier, I think there may be some issues with the patch release section to do with checking for the presence of the release branch, but also when a release branch is tracked in more than one remote (as would be case if you have remotes for your own fork and upstream). I'll file an issue for those to address for 1.5.1.

nrb: Yeah, I'm sure there are some weird cases. I'm definitely tracking release branches on my own fork and on the upstream remote, but when testing this I added a 3rd remote which was a duplicate of my fork so I didn't make any bad releases against real upstream.

PR #2881, opened 2020-08-31
Add additional printer columns for CRDs

### LINKED ISSUE #1780: Define printer columns for Velero CRDs for kubectl interoperability
Currently, the Velero CLI shows more information than `kubectl`:

```
$ velero get backups
NAME                  STATUS      CREATED                         EXPIRES   STORAGE LOCATION   SELECTOR
manual-3              Completed   2019-08-06 15:26:22 -0700 PDT   15d       default            app=nginx

$ kubectl get backups -n velero
NAME                  AGE
manual-3              14d
```

We can define `additionalPrinterColumns` in Velero CRDs to customise the output when using `kubectl` to show the same information. Other tools also use this information to present more information, such as [Octant](https://github.com/vmware/octant), so this would create a consistent UX across these tools.

See https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#additional-printer-columns

zubron:
This change modifies the kubebuilder annotations for the Velero CRDs to
include `additionalPrinterColumns` so that more information is exposed
when using `kubectl get`.

For each of the CRDs, annotations have been added to make the output
for `kubectl get` match the output from the equivalent `velero get`
command as closely as possible. There are some cases where this output
could not be replicated, such as the `EXPIRES` column for Backups, due
to the limitations of JSONPath expressions within the resulting CRD
defition. Some columns undergo processing and formatting before being
printed by the Velero CLI which cannot be replicated using JSONPath. In
these cases, these printer columns have been omitted.

For other CRDs where there is no `velero get` equivalent, such as
`PodVolumeBackup` and `PodVolumeRestore`, a best effort has been made to
expose information that provides value.

Fixes #1780


nrb:
@zubron Since we've locked the v1.5.0 list except for critical fixes, I'm going to mark this for v1.5.1 in the GitHub milestones.

pkg/apis/velero/v1/backup.go
```diff
@@ -312,6 +312,12 @@ type BackupProgress struct {
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.phase",description="Backup status such as InProgress/Completed"
+// +kubebuilder:printcolumn:name="Errors",type="integer",JSONPath=".status.errors",description="Count of all error messages that were generated during execution of this backup"
+// +kubebuilder:printcolumn:name="Warnings",type="integer",JSONPath=".status.warnings",description="Count of all warning messages that were generated during execution of this backup"
+// +kubebuilder:printcolumn:name="Created",type="date",JSONPath=".status.startTimestamp",description="Time when this backup was started"
```
ashish-amarnath, line 318: suggest: `Start Time`

pkg/apis/velero/v1/delete_backup_request.go
```diff
@@ -52,6 +52,9 @@ type DeleteBackupRequestStatus struct {
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.phase",description="Delete Backup Request status such as InProgress/Processed"
+// +kubebuilder:printcolumn:name="Backup Name",type="string",JSONPath=".spec.backupName",description="Name of backup to be deleted"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
```
ashish-amarnath, line 57: suggest: `Created At`

pkg/apis/velero/v1/download_request.go
```diff
@@ -79,6 +79,10 @@ type DownloadRequestStatus struct {
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.phase",description="DownloadRequest status such as New/Processed"
+// +kubebuilder:printcolumn:name="Target Name",type="string",JSONPath=".spec.target.name",description="Name of the associated Kubernetes resource"
+// +kubebuilder:printcolumn:name="Target Kind",type="string",JSONPath=".spec.target.kind",description="Type of file to download"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
```
ashish-amarnath, line 85: suggest: `Created At`

pkg/apis/velero/v1/pod_volume_backup.go
```diff
@@ -100,6 +100,14 @@ type PodVolumeBackupStatus struct {
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.phase",description="Pod Volume Backup status such as New/InProgress"
+// +kubebuilder:printcolumn:name="Created",type="date",JSONPath=".status.startTimestamp",description="Time when this backup was started"
```
ashish-amarnath, line 104: suggest: `Start Time`

pkg/apis/velero/v1/pod_volume_backup.go
```diff
@@ -100,6 +100,14 @@ type PodVolumeBackupStatus struct {
...
+// +kubebuilder:printcolumn:name="Namespace",type="string",JSONPath=".spec.pod.namespace",description="Namespace of the pod containing the volume to be backed up"
+// +kubebuilder:printcolumn:name="Pod",type="string",JSONPath=".spec.pod.name",description="Name of the pod containing the volume to be backed up"
+// +kubebuilder:printcolumn:name="Volume",type="string",JSONPath=".spec.volume",description="Name of the volume to be backed up"
+// +kubebuilder:printcolumn:name="Restic Repo",type="string",JSONPath=".spec.repoIdentifier",description="Restic repository identifier for this backup"
+// +kubebuilder:printcolumn:name="Storage Location",type="string",JSONPath=".spec.backupStorageLocation",description="Name of the Backup Storage Location where this backup should be stored"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
```
ashish-amarnath, line 110: Suggest: `Created At`

pkg/apis/velero/v1/pod_volume_restore.go
```diff
@@ -83,6 +83,15 @@ type PodVolumeRestoreStatus struct {
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.phase",description="Restore status such as InProgress/Completed"
+// +kubebuilder:printcolumn:name="Namespace",type="string",JSONPath=".spec.pod.namespace",description="Namespace of pod containing the volume to be restored"
+// +kubebuilder:printcolumn:name="Pod",type="string",JSONPath=".spec.pod.name",description="Name of pod containing the volume to be restored"
+// +kubebuilder:printcolumn:name="Volume",type="string",JSONPath=".spec.volume",description="Name of volume to restore"
+// +kubebuilder:printcolumn:name="Bytes Done",type="string",JSONPath=".status.progress.bytesDone",description="Number of bytes restored"
+// +kubebuilder:printcolumn:name="Total Bytes",type="string",JSONPath=".status.progress.totalBytes",description="Total number of bytes to be restored"
+// +kubebuilder:printcolumn:name="Started",type="date",JSONPath=".status.startTimestamp",description="Time when the restore operation was started"
```
ashish-amarnath, line 92: nit: `Start Time`
for consistency

pkg/apis/velero/v1/pod_volume_restore.go
```diff
@@ -83,6 +83,15 @@ type PodVolumeRestoreStatus struct {
...
+// +kubebuilder:printcolumn:name="Completed",type="date",JSONPath=".status.completionTimestamp",description="Time when the restore operation was completed"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
```
ashish-amarnath, line 94: suggest: `Created At`

pkg/apis/velero/v1/restore.go
```diff
@@ -251,6 +251,13 @@ type RestoreStatus struct {
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:name="Backup",type="string",JSONPath=".spec.backupName",description="Name of backup to restore from"
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.phase",description="Restore status such as InProgress/Completed"
+// +kubebuilder:printcolumn:name="Started",type="date",JSONPath=".status.startTimestamp",description="Time when the restore operation was started"
```
ashish-amarnath, line 256: nit: `Start Time`
for consistency

pkg/apis/velero/v1/restore.go
```diff
@@ -251,6 +251,13 @@ type RestoreStatus struct {
...
+// +kubebuilder:printcolumn:name="Completed",type="date",JSONPath=".status.completionTimestamp",description="Time when the restore operation was completed"
+// +kubebuilder:printcolumn:name="Errors",type="integer",JSONPath=".status.errors",description="Count of all errors generated during the restore"
+// +kubebuilder:printcolumn:name="Warnings",type="integer",JSONPath=".status.warnings",description="Count of all warnings generated during the restore"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
```
ashish-amarnath, line 260: suggest: `Created At`

pkg/apis/velero/v1/volume_snapshot_location.go
```diff
@@ -20,6 +20,8 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:name="Provider",type="string",JSONPath=".spec.provider"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
```
ashish-amarnath, line 24: suggest: `Created At`

ashish-amarnath (changes requested):
@zubron I added a few minor comments. PTAL

pkg/apis/velero/v1/backup.go
zubron: I opted for the "Created" title to match the equivalent column from `velero get backups`. Same with "Age" to match the final column in standard `kubectl get` output (we already had that field/column for BSLs). Is consistency with those important?

pkg/apis/velero/v1/delete_backup_request.go
carlisia: We should stick with `Age` here and in the BSL, or change both to `Created At`. `Age` sounds good to me, especially since a lot/most/(all) k8s resources have this column.

pkg/apis/velero/v1/backup.go
carlisia: +1 for being consistent with the columns for backups and using `Created` .

pkg/apis/velero/v1/download_request.go
carlisia: See above

pkg/apis/velero/v1/pod_volume_backup.go
carlisia: It'd be helpful to know the reasoning for this request. Otherwise, `Created` is consistent with the column name for our other resources.

carlisia:
@zubron when you have a chance please resolve the conflicts, thanks.

pkg/apis/velero/v1/backup.go
nrb: Agreed, I would prefer consistency.

pkg/apis/velero/v1/pod_volume_restore.go
nrb: I'll note that this doesn't match what we've got specified in the backup.go file right now; I think whichever we choose, the column name should be consistent across object types.

pkg/apis/velero/v1/delete_backup_request.go
nrb: It looks like `Age` on built-in kubernetes resources is a calculated field, not simply a timestamp.

So if we're not calculating the age here, I don't think we should re-use that name, because that's actually using the same name for something different but similar.

As an example, here's what nodes report:

```
% k get nodes
NAME                 STATUS   ROLES    AGE   VERSION
kind-control-plane   Ready    master   19h   v1.17.0
```

nrb (dismissed):
I know we had a discussion about calculated fields in the community meeting, and we can't do them without extra code.

With that in mind, I think renaming the current `Age` fields would be desirable, because `kubectl` uses a calculated value as opposed to a bare timestamp for this when querying for built-in resources.

For example, here's what it looks like when querying for nodes:

```
% kubectl get nodes
NAME                 STATUS   ROLES    AGE   VERSION
kind-control-plane   Ready    master   19h   v1.17.0
```

In this case, the `Age` column is actually the difference between now and the creation timestamp. If we use the creationTimestamp directly, I worry about overloading the field's name with a different type of data.

pkg/apis/velero/v1/delete_backup_request.go
carlisia: Kubebuilder does the age calculation when you name a column "Age". BSL uses the same marker, you may verify it with that.

zubron: Yes, kubebuilder seems to handle it correctly. For all the `Age` columns that I added, the output is a relative time, for example:

```
k get deletebackuprequests.velero.io -n velero
NAME                      STATUS       BACKUP NAME         AGE
test-pod-backup-2-pc969   InProgress   test-pod-backup-2   16d
test-pod-backup-3-h4689   InProgress   test-pod-backup-3   38s
```

zubron: This doesn't seem to be specific to the `Age` column though. It seems anytime that the `date` type is used, this is how it is displayed by kubectl. For example:

```
 $ k get restores.velero.io -n velero
NAME                               BACKUP              STATUS            STARTED   COMPLETED   ERRORS   WARNINGS   AGE
test-pod-backup-20200831105635     test-pod-backup     Completed         17d       17d                  27         17d
test-pod-backup-3-20200831122428   test-pod-backup-3   Completed         17d       17d                  27         17d
test-pod-backup-3-20200831122623   test-pod-backup-3   Completed         17d       17d                  26         17d
test-pod-backup-3-manul            test-pod-backup-3   PartiallyFailed   15d       15d         7        22         15d
```

All timestamp values are displayed as a duration from the current time.

carlisia: Ah yeah, the name of the column doesn't matter, it is the (date) type!

carlisia (approved):
👍

pkg/apis/velero/v1/backup.go
nrb: I think I would prefer to match the current `velero get backups` output, since that's established and updating it _may_ break user scripts, so `Created` works for me.

That does, however, put us in an awkward spot since I think all other `.status.startTimestamp` fields should match, due to the existence of the `creationTimestamp` field. Ideally, we'd have clean mappings between the field name in the YAML and printer column, but I think due to historical reasons we're a little stuck.

So my vote is this mapping:

Print `.status.startTimestamp` under `Created`
Print `.metadata.creationTimestamp` under `Age`

zubron: Sorry, I'm not sure I understand what you mean by:
> That does, however, put us in an awkward spot since I think all other .status.startTimestamp fields should match, due to the existence of the creationTimestamp field.

As for the mapping you'd prefer, that is what is in this file. Am I missing something?

nrb: The awkward spot I was referring to is that we have `creationTimestamp` that isn't mapped to the `Created` display name. It's probably not a big deal except for people looking at the code.

It looks like there's 2 instances where `status.startTimestamp` is mapped to `Start Time`, rather than `Created`:

 * https://github.com/vmware-tanzu/velero/pull/2881/files#diff-2882940cdbbb4c1142f4b9468c593dc0R38
 * https://github.com/vmware-tanzu/velero/pull/2881/files#diff-315e9e294c9cc0ab062afc00990fa0aaR18

zubron: Ah, I see. Thanks for the explanation!

I used `Started` as those column names which matches the output from `velero get restores`... another inconsistency :(

nrb: Gotcha - yeah, that was our past mistake.

I think we could put that on the docket for v2.0 to update those `velero` commands so we don't break any scripts.

zubron: Created #2989 to address this.

nrb:
I think I'm going to lean towards keeping the fields internally consistent, and letting this PR merge as is. We now have an action item to clean things up in v2.0, so I think that will work.

nrb:
@ashish-amarnath Could you review the comments and update your review?

nrb (approved):

ashish-amarnath (approved):

Issue #2834, opened 2020-08-18
2nd velero installation in non-default namespace does not work


ghost74-tg:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is, and what commands you ran.)
I have tried to install velero twice in my kubernetes cluster. The first velero installation was installed in the default namespace velero and the second in the namespace new-velero.
The default installation works fine. The velero pod is in running state and I could backup namespaces.
The velero pod of the second installation is in error state.
```
root@kubmaster1 ~]# kubectl get pods -n new-velero -o wide
NAME                      READY   STATUS   RESTARTS   AGE    IP               NODE         NOMINATED NODE   READINESS GATES
velero-7c6d9bb58d-h7brr   0/1     Error    4          114s   192.168.77.217   kubworker2   <none>           <none>
```
When checking for the logs I get the following:
```
[root@kubmaster1 ~]# kubectl logs deployment/velero -n new-velero
time="2020-08-18T13:02:12Z" level=info msg="setting log-level to INFO" logSource="pkg/cmd/server/server.go:177"
time="2020-08-18T13:02:12Z" level=info msg="Starting Velero server v1.4.2 (56a08a4d695d893f0863f697c2f926e27d70c0c5)" logSource="pkg/cmd/server/server.go:179"
time="2020-08-18T13:02:12Z" level=info msg="1 feature flags enabled []" logSource="pkg/cmd/server/server.go:181"
time="2020-08-18T13:02:12Z" level=info msg="registering plugin" command=/velero kind=BackupItemAction logSource="pkg/plugin/clientmgmt/registry.go:100" name=velero.io/crd-remap-version
time="2020-08-18T13:02:12Z" level=info msg="registering plugin" command=/velero kind=BackupItemAction logSource="pkg/plugin/clientmgmt/registry.go:100" name=velero.io/pod
time="2020-08-18T13:02:12Z" level=info msg="registering plugin" command=/velero kind=BackupItemAction logSource="pkg/plugin/clientmgmt/registry.go:100" name=velero.io/pv
time="2020-08-18T13:02:12Z" level=info msg="registering plugin" command=/velero kind=BackupItemAction logSource="pkg/plugin/clientmgmt/registry.go:100" name=velero.io/service-account
time="2020-08-18T13:02:12Z" level=info msg="registering plugin" command=/velero kind=RestoreItemAction logSource="pkg/plugin/clientmgmt/registry.go:100" name=velero.io/add-pv-from-pvc
[...]

PR #2886, opened 2020-09-01
create CRB with velero-<namespace>

alaypatel07:
This will allow creating multiple instances of velero,
across two different namespaces

fixes: #2834

### DIFF
--- a/pkg/install/resources.go
+++ b/pkg/install/resources.go
@@ -51,6 +51,7 @@ var (
 	DefaultResticPodMemRequest = "512Mi"
 	DefaultResticPodCPULimit   = "1000m"
 	DefaultResticPodMemLimit   = "1Gi"
+	DefaultVeleroNamespace     = "velero"
 )
 
 func labels() map[string]string {
@@ -106,7 +107,7 @@ func ServiceAccount(namespace string, annotations map[string]string) *corev1.Ser
 
 func ClusterRoleBinding(namespace string) *rbacv1beta1.ClusterRoleBinding {
 	crbName := "velero"
-	if namespace != "velero" {
+	if namespace != DefaultVeleroNamespace {
 		crbName = "velero-" + namespace
 	}
 	crb := &rbacv1beta1.ClusterRoleBinding{
--- a/pkg/install/resources.go
+++ b/pkg/install/resources.go
@@ -105,8 +105,12 @@ func ServiceAccount(namespace string, annotations map[string]string) *corev1.Ser
 }
 
 func ClusterRoleBinding(namespace string) *rbacv1beta1.ClusterRoleBinding {
+	crbName := "velero"
+	if namespace != "velero" {
+		crbName = "velero-" + namespace
+	}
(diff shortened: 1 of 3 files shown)


nrb:
Thanks @alaypatel07! To set expectations, I think this will be a v1.5.1 feature, in order to minimize changes coming in so close to v1.5.0's release. I'm going to set the GitHub milestone to reflect that.

nrb:
@alaypatel07 Thanks for this! Can you please add a changelog entry? See the failed job for instructions on what to add.

pkg/install/resources.go
```diff
@@ -105,8 +105,12 @@ func ServiceAccount(namespace string, annotations map[string]string) *corev1.Ser
 }
 
 func ClusterRoleBinding(namespace string) *rbacv1beta1.ClusterRoleBinding {
+	crbName := "velero"
+	if namespace != "velero" {
```
nrb, line 109: Sorry I missed this before, but we should probably use the [`DefaultNamespace`](https://github.com/vmware-tanzu/velero/blob/main/pkg/apis/velero/v1/constants.go#L22) constant to compare against here. It looks like it might not be used throughout the rest of the file, but that's an oversight.

nrb (changes requested):
One minor change that I missed when looking at this the first time, and then I think this is good for a merge!

pkg/install/resources.go
alaypatel07: Following the convention of Default vars, I added `DefaultVeleroNamespace` and used it in the unit tests. PTAL

nrb (approved):

pkg/install/resources.go
```diff
@@ -105,8 +106,12 @@ func ServiceAccount(namespace string, annotations map[string]string) *corev1.Ser
 }
 
 func ClusterRoleBinding(namespace string) *rbacv1beta1.ClusterRoleBinding {
+	crbName := "velero"
+	if namespace != DefaultVeleroNamespace {
+		crbName = "velero-" + namespace
+	}
```
carlisia, line 112:
```
		crbName := DefaultVeleroNamespace
		if namespace != DefaultVeleroNamespace {
			crbName = crbName + "-" + namespace
		}
```

carlisia (changes requested):
This is a bit nit-picking but I think it will improve the understanding of the intention behind changing the value of the `crbName` better than setting it with a hard-coded value.

pkg/install/resources.go
nrb: Hmm, this is a little awkward, because the crb and the namespace are named `velero`, but I don't think we should use the namespace constant. That, to me, creates confusion and an unclear coupling between the two objects.

For a similar example, in our standard deployment, the deployment and the namespace are both `velero`.

carlisia: I see. And that's a good point. I'll merge this.

carlisia (approved):
Thank you! 👍

PR #2895, opened 2020-09-02
Fix for docs redirects

jonasrosland:
This will fix issue #2824, re-add the previously working short urls, and move the redirect logic for the latest docs version to a separate redirects file instead of being built into the docs page.  

Fixes: #2824

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,7 +1,7 @@
 [build]
   base = "site/"
   command = "hugo --gc --minify --baseURL $BASE_URL"
-  publish = "public"
+  publish = "site/public"
 
 [context.production.environment]
   HUGO_VERSION = "0.73.0"
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,7 +1,7 @@
 [build]
   base = "site/"
   command = "hugo --gc --minify --baseURL $BASE_URL"
-  publish = "site/public"
+  publish = "public"
 
 [context.production.environment]
   HUGO_VERSION = "0.73.0"
--- a/site/layouts/index.redirects
+++ b/site/layouts/index.redirects
@@ -1,5 +1,13 @@
 {{ $versions := site.Params.versions }}
 {{ $latest   := site.Params.latest }}
-/docs    /docs/{{ $latest }}     301!
-/docs/latest     /docs/{{ $latest }}
-/docs/latest/*     /docs/{{ $latest }}/:splat
+/docs                          /docs/{{ $latest }}     301!
+/docs/latest                   /docs/{{ $latest }}
+/docs/latest/*                 /docs/{{ $latest }}/:splat
+/docs/troubleshooting          /docs/{{ $latest }}/troubleshooting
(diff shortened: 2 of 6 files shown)


netlify.toml
```diff
@@ -1,7 +1,7 @@
 [build]
   base = "site/"
   command = "hugo --gc --minify --baseURL $BASE_URL"
-  publish = "site/public"
+  publish = "public"
```
nrb, line 4: Looks like the deploy preview failed:

`Failed during stage 'building site': Deploy directory 'public' does not exist` 

Is that due to this line perhaps?

jonasrosland: Yup, "worked on my machine" but apparently that's not good enough for Netlify :)
Pushed the change, now the short links work: https://deploy-preview-2895--velero.netlify.app/docs/install-overview

site/layouts/index.redirects
```diff
@@ -0,0 +1,13 @@
...
+/docs/latest                   /docs/{{ $latest }}
+/docs/latest/*                 /docs/{{ $latest }}/:splat
+/docs/troubleshooting          /docs/{{ $latest }}/troubleshooting
+/docs/supported-providers      /docs/{{ $latest }}/supported-providers
+/docs/zenhub                   /docs/{{ $latest }}/zenhub
+/docs/install-overview         /docs/{{ $latest }}/basic-install
+/docs/start-contributing       /docs/{{ $latest }}/start-contributing
+/docs/customize-installation   /docs/{{ $latest }}/customize-installation
+/docs/faq                      /docs/{{ $latest }}/faq
+/docs/csi                      /docs/{{ $latest }}/csi
```
nrb, line 13: While we're at it, can we add `docs/restic`, since that's a big one?

site/layouts/index.redirects
```diff
@@ -0,0 +1,13 @@
+{{ $versions := site.Params.versions }}
+{{ $latest   := site.Params.latest }}
+/docs                          /docs/{{ $latest }}     301!
+/docs/latest                   /docs/{{ $latest }}
+/docs/latest/*                 /docs/{{ $latest }}/:splat
```
nrb, line 5: This would presumably catch all pages, but in testing out the restic page on the preview, its giving me a 404.

site/layouts/index.redirects
jonasrosland: Added!

site/layouts/index.redirects
jonasrosland: It catches all pages after latest, so https://deploy-preview-2895--velero.netlify.app/docs/latest/restic works.

nrb: Ahhhhhhhh, thanks, I was having a moment :)

Will use that going forward :)

nrb (approved):
It works, thanks! This will be awesome going forward!

ashish-amarnath (approved):
LGTM 👍

PR #2931, opened 2020-09-11
Spruce up release instructions and release scripts

carlisia:
Changes in this PR are meant to reduce as much text as possible from the instructions, as well as some groupings/reorganization of the text.

Highlights:
- Moved scripts used for release into the release-tools folder
- Removed text about generating/regenerating a GH token
- Moved documentation for each variable to inside the scripts that use them.
- Used a format for examples of calling commands using variables that work in most (maybe all) shells (as opposed to just bash)
- Suggest we don’t use `pre` to represent all the alpha/beta/rc versions of the doc. Reason: the script that generates the docs needs a directory to copy from that is different than the target/new directory, and they must have different names. I think we should just use the full version name. Based on this:
```
NEW_DOCS_VERSION defines the version that the docs will be tagged with (i.e. what’s in the URL, what shows up in the version dropdown on the site). This should be formatted as either v1.4 (for a GA release), or v1.4-pre (for an alpha/beta/RC).
```

Notes:
1) The s/true/TRUE seemed needed because the `$PUBLISH` var in goreleaser.sh gets populated from the `$publish` var from the tag-release.sh, which is set in upper case:
https://github.com/vmware-tanzu/velero/compare/main...carlisia:c-release-2?expand=1#diff-98439754b10c2cfb6407857c90e5a6b1R41
2) Not sure how the above worked before. When I did the rc.1 release I had to call gorelease directly to make it work.


carlisia:
View: https://deploy-preview-2931--velero.netlify.app/docs/v1.5.0-rc.1/release-instructions/

carlisia:
@jonasrosland / @a-mccarthy have you run into this before: https://github.com/vmware-tanzu/velero/pull/2931/files#diff-57c0aae5dc6b9434f0447c17c88b9f1dR6

I can delete that line but whenever I save it gets added back in!

hack/release-tools/goreleaser.sh
```diff
@@ -36,14 +36,16 @@ else
...
 # to publish the release to GitHub.
-if [[ "${PUBLISH:-}" != "true" ]]; then
+if [[ "${PUBLISH:-}" != "TRUE" ]]; then
+    echo "Getting ready to publish"
     goreleaser release \
         --rm-dist \
         --release-notes="${RELEASE_NOTES_FILE}" \
         --skip-publish
 else
+    echo "Not set to publish"
```
ashish-amarnath, line 48: the `else` block will publish and the `if` block will `skip-publish`

site/content/docs/main/release-instructions.md
```diff
@@ -3,14 +3,117 @@ title: "Release Instructions"
 layout: docs
 ---
 
-This page covers the steps to perform when releasing a new version of Velero.
-
-## Preparing for a release
+- [General notes](#general-notes)
```
ashish-amarnath, line 6: ToC can be deleted as it is auto-generated for all pages on the website.

site/content/docs/main/release-instructions.md
```diff
@@ -3,14 +3,117 @@ title: "Release Instructions"
...
+	- Note: `PREVIOUS_DOCS_VERSION=<doc-version-to-copy-from>` is optional; when not set, it will default to the latest doc version.
+6. Clean up when there is an existing set of pre-release versioned docs for the version you are releasing
+	- Example: `site/content/docs/v1.5.0-beta.1` exists, and you're releasing `v1.5.0-rc.1` or `v1.5`
+    - Remove the directory containing the pre-release docs, i.e. `site/content/docs/<pre-release-version>`.
+    - Delete the pre-release docs table of contents file, i.e. `site/data/<pre-release-version>-toc.yml`.
+    - Remove the pre-release docs table of contents mapping entry from `site/data/toc-mapping.yml`.
+    - Remove all references to the pre-release docs from `site/config.yml`.
+7. Create/update the "Upgrade to $major.minor" page.
+8. Review and submit PR
+	- Follow the additional instructions at `site/README-JEKYLL.md` to complete the docs generation process.
```
ashish-amarnath, line 67: `site/README-HUGO.md`

.github/workflows/pr-changelog-check.yml
```diff
@@ -12,4 +12,4 @@ jobs:
 
     - name: Changelog check
       if: ${{ !(contains(github.event.pull_request.labels.*.name, 'changelog-not-required') || contains(github.event.pull_request.labels.*.name, 'Design') || contains(github.event.pull_request.labels.*.name, 'Website') || contains(github.event.pull_request.labels.*.name, 'Documentation'))}}
-      run: ./hack/changelog-check.sh
+      run: ./hack/release-tools/changelog-check.sh
```
ashish-amarnath, line 15: `changelog-check.sh` is the script that verifies PRs have a changelog associated with them.
This is not part of the release tooling.

ashish-amarnath (changes requested):

hack/release-tools/goreleaser.sh
carlisia: Totally missed that `!` 🙀

hack/release-tools/tag-release.sh
```diff
@@ -26,7 +41,7 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
 # Parse out the branch we're on so we can switch back to it at the end of a dry-run, where we delete the tag. Requires git v1.8.1+
-original_branch=$(git symbolic-ref --short HEAD)
+upstreamal_branch=$(git symbolic-ref --short HEAD)
```
zubron, line 44: Minor thing, but I think this variable should be `upstream_branch`?

site/content/docs/main/release-instructions.md
zubron: Do you need to include `toc: "true"` in the header to generate that?

site/content/docs/main/release-instructions.md
```diff
@@ -3,14 +3,117 @@ title: "Release Instructions"
...
+- [How to write and release a blog post](#how-to-write-and-release-a-blog-post)
+- [Announce a release](#announce-a-release)
 
-### (GA Only) Create release blog post
-
-For each major or minor release, create and publish a blog post to let folks know what's new.
+This page covers the steps to perform when releasing a new version of Velero.
 
+## General notes
+- Please read the documented variables in each scrip to understand what they are for and how to properly format their values.
```
zubron, line 9:
```suggestion
- Please read the documented variables in each script to understand what they are for and how to properly format their values.
```

site/content/docs/main/release-instructions.md
```diff
@@ -1,16 +1,103 @@
...
+- RC releases: Release Candidate, contains everything that is supposed to ship with the GA release. This is still a pre-release.
+
+## Preparing
+
+### Create release blog post (GA only)
+For each major or minor release, create and publish a blog post to let folks know what's new. Please follow these [instructions](Instructions-to-write-and-release-a-blog-post).
+
+### Changelog and Docs PR
+#### Troubleshooting
+- `You don't have enough free space in /var/cache/apt/archives/` when running `make serve-docs`: run `docker system prune`.
```
zubron, line 22: I think it would be clearer if this line had something before the error message like:
```suggestion
- If you encounter the error `You don't have enough free space in /var/cache/apt/archives/` when running `make serve-docs`: run `docker system prune`.
```

site/content/docs/main/release-instructions.md
```diff
@@ -1,16 +1,103 @@
...
+- Please read the documented variables in each scrip to understand what they are for and how to properly format their values.
+- You will need to have a remote named `origin` pointing to the Velero repository so the scripts can work properly.
+- GA release: major and minor releases only. Example: 1.0 (major), 1.5 (minor).
+- Pre-releases: Any release leading up to a GA. Example: 1.4.0-beta.1, 1.5.0-rc.1
+- RC releases: Release Candidate, contains everything that is supposed to ship with the GA release. This is still a pre-release.
+
+## Preparing
+
+### Create release blog post (GA only)
+For each major or minor release, create and publish a blog post to let folks know what's new. Please follow these [instructions](Instructions-to-write-and-release-a-blog-post).
```
zubron, line 18: Hmm, this link doesn't work in the preview - should it be linking to `how-to-write-and-release-a-blog-post`?

site/content/docs/main/release-instructions.md
```diff
@@ -1,16 +1,103 @@
...
+	- Under "Current release": 
+	    - Should contain only the current GA release.
+    - Under "Development release": 
+	    - Should contain only the latest pre-release
+	    - Move any prior pre-release into "Older releases"
+4. GA Only: Remove all changelog files from `changelogs/unreleased`.
+5. Generate new docs
+	- Run `make gen-docs`, passing the appropriate variables. Examples:
+		a) `VELERO_VERSION=v1.5.0-rc.1 NEW_DOCS_VERSION=v1.5.0-rc.1 make gen-docs`.
+		b) `VELERO_VERSION=v1.5.0 NEW_DOCS_VERSION=v1.5 make gen-docs`).
```
zubron, line 41: For a standard GA release, should the new docs version always omit the patch component of the version?

site/content/docs/main/release-instructions.md
```diff
@@ -1,16 +1,103 @@
...
+	- Note: `PREVIOUS_DOCS_VERSION=<doc-version-to-copy-from>` is optional; when not set, it will default to the latest doc version.
+6. Clean up when there is an existing set of pre-release versioned docs for the version you are releasing
+	- Example: `site/content/docs/v1.5.0-beta.1` exists, and you're releasing `v1.5.0-rc.1` or `v1.5`
+    - Remove the directory containing the pre-release docs, i.e. `site/content/docs/<pre-release-version>`.
+    - Delete the pre-release docs table of contents file, i.e. `site/data/<pre-release-version>-toc.yml`.
+    - Remove the pre-release docs table of contents mapping entry from `site/data/toc-mapping.yml`.
+    - Remove all references to the pre-release docs from `site/config.yml`.
+7. Create/update the "Upgrade to $major.minor" page.
```
zubron, line 49: Could we include an example path to this file, e.g. `site/content/docs/<version>/upgrade-to-<major>.<minor>.md`? Does it need to be updated in both the versioned *and* main docs folders?

site/content/docs/main/release-instructions.md
```diff
@@ -1,16 +1,103 @@
...
+8. Review and submit PR
+	- Follow the additional instructions at `site/README-HUGO.md` to complete the docs generation process.
+	- Do a review of the diffs, and/or run `make serve-docs` and review the site.
+	- Submit a PR containing the changelog and the version-tagged docs.
+
+## Velero release
+#### Notes
```
zubron, line 56: The drop from h2 to h4 means that the TOC is broken. I think all these headers should have `###`

zubron (changes requested):

hack/release-tools/tag-release.sh
```diff
@@ -76,6 +91,7 @@ printf "Based on this, the following assumptions have been made: \n"
 
 [[ "$VELERO_PATCH" != 0 ]] && printf "*\t This is a patch release.\n"
 
+# $VELERO_PRERELEASE gets populated by the chk_version.go scrip that parses and verifies the given version format
```
nrb, line 94: Minor, but `scrip` -> `script`.

site/content/docs/main/release-instructions.md
```diff
@@ -1,16 +1,103 @@
...
 
-## Preparing for a release
-
-### (GA Only) Create release blog post
-
-For each major or minor release, create and publish a blog post to let folks know what's new.
-
+## General notes
+- Please read the documented variables in each scrip to understand what they are for and how to properly format their values.
+- You will need to have a remote named `origin` pointing to the Velero repository so the scripts can work properly.
```
nrb, line 10: Should it be `origin`? I think it should be `upstream`, at least it was when I first wrote them.

site/content/docs/main/release-instructions.md
nrb: Yes. We used to do patch release-level docs, but we now only do minor release-level (so v1.5) docs in order to make getting patch releases out easier.

hack/release-tools/gen-docs.sh
```diff
@@ -14,8 +14,36 @@
...
+# NEW_DOCS_VERSION defines the version that the docs will be tagged with 
+# (i.e. what’s in the URL, what shows up in the version dropdown on the site). 
+# This should be formatted as either v1.4 (for any GA release, including minor), or v1.5.0-beta.1/v1.5.0-rc.1 (for an alpha/beta/RC).
+
+# To run gen-docs: "VELERO_VERSION=v1.4.0 NEW_DOCS_VERSION=v1.4 PREVIOUS_DOCS_VERSION= make gen-docs"
+# Note: if PREVIOUS_DOCS_VERSION is not set, the script will copy from the 
+# latest version.
+#
+# **NOTE**: there are additional manual steps required to finalize the process of generating
+# a new versioned docs site. The full process is documented in site/README-JEKYLL.md
```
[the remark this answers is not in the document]
carlisia: good catch!

hack/release-tools/tag-release.sh
carlisia: argh, good catch

site/content/docs/main/release-instructions.md
```diff
@@ -1,16 +1,103 @@
...
+	- Perform a smoke-test - for example:
+		- Download the CLI from the GitHub release
+	    - Use it to install Velero into a cluster (or manually update an existing deployment to use the new images)
+	    - Verify that `velero version` shows the expected output
+	    - Run a backup/restore and ensure it works
+
+## Homebrew release
+These are the steps to update the Velero Homebrew version.
+
+#### Note
```
nrb, line 91: I'd add `GA Only` to L88 and remove this note section.

site/content/docs/main/release-instructions.md
```diff
@@ -22,97 +109,17 @@ What to include in a release blog:
...
 ## Announce a release
 
-Once you are finished doing the release, let the rest of the world know its available by posting messages in the following places.
-
-1. Velero's Twitter account. Maintainers are encouraged to help spread the word by posting or reposting on social media.
-1. Community Slack channel.
-1. Google group message.
+Once you are finished doing the release, let the rest of the world know it's available by posting messages in the following places.
+1.  GA Only: Merge the blog post PR.
+2. Velero's Twitter account. Maintainers are encouraged to help spread the word by posting or reposting on social media.
```
nrb, line 118: These were all using `1.` in order to let the Markdown engine generate the numbering. The trade off here is that during editing it can be hard to keep track of the points.

So I can take or leave this change. In general, I think I would prefer to have these files written for rendering on the site.

site/content/docs/main/release-instructions.md
carlisia: This is not what I did for the RC (see it is v1.5.0-rc.1). The instructions made me think all non GA docs should be under `-pre`. We need to decide. I think it should not be `-pre` (see this PR description. Options are:
- v1.5.0-rc.1 (full version)
- v1.5 (omit the patch)

nrb: For a GA release, which is what @zubron was asking about, no patch number at the end.

For non-GA releases, yes, putting them under `-pre` was what was happening before. I like keeping `pre` because for those, we're iterating and can/should be deleting those docs directories anyway, so we could modify the scripts to clean up an existing `<version>-pre` directory. Pre-release version docs, at least in my mind, are largely ephemeral, and will be gone once the GA release comes out.

Using the explicit version number means we're generating dedicated directories per pre-release version, and I don't really see the benefit to having those stick around.

carlisia: What would be the difference between using `<version>-pre` vs something like `v1.5.0-rc.1`?

Both would be deleted once GA is out. `<version>-pre` to me is just one more term to wrap our head around.

site/content/docs/main/release-instructions.md
carlisia: Good question. I added "This needs to be done in both the versioned and the `main` folders.".

site/content/docs/main/release-instructions.md
nrb: `-pre` covers all pre-releases, so alpha, beta, and RC releases.

carlisia: > -pre covers all pre-releases, so alpha, beta, and RC releases.

How do you do the copy/paste if the origin and target folders are named the same?

nrb: The use can set the `PREVIOUS_DOCS_VERSION` value themselves (https://github.com/vmware-tanzu/velero/blob/main/hack/gen-docs.sh#L37), but I think I see your point now, I was missing it before :)

This is surfacing a bug in the behavior of the `gen-docs.sh` script, which assumes that the latest release is always the last major version and you won't have multiple pre-releases.

In which case we either a) need to fix the script to account for this or b) update the instructions to explicitly set the `PREVIOUS_DOCS_VERSION` or some third option I'm not thinking of...

carlisia: But what do you set `PREVIOUS_DOCS_VERSION` to, this is my question.

Example: 
1) We have `v1.6.0-beta.1` release
2) The docs for that is under `v1.6-pre`
3) We want to release `v1.6-beta.2`
3) What should the `PREVIOUS_DOCS_VERSION` be? If we say `v1.6-pre`, then it can't copy itself to `v1.6-pre` right?

nrb: Yeah, I edited my comment, but I think it should be `v1.5` for all pre-release versions in your example. You're correct that you couldn't do `v1.6-pre` for both unless we did some extra move/swap logic, which doesn't seem worthwhile to add. 

That would match the intended logic of latest, non-pre-release version that the script is expecting, while getting around the fact that the real latest based on sorting would be `v1.6-pre`.

nrb: To answer your question directly, here's what I envision.

For beta.1:

`PREVIOUS_DOCS_VERSION=v1.5`
`NEW_DOCS_VERSION=v1.6-pre`

All files in the `v1.6-pre` directory will be newly created.

For beta.2:

`PREVIOUS_DOCS_VERSION=v1.5`
`NEW_DOCS_VERSION=v1.6-pre`

All files in the `v1.6-pre` directory will be overwritten, only those with changes should show up in a `git status` though. It's totally plausible that the docs don't even change.

carlisia: > PREVIOUS_DOCS_VERSION=v1.5
NEW_DOCS_VERSION=v1.6-pre

> All files in the v1.6-pre directory will be overwritten

With the docs from `v1.5`, right?

So what happens if during `beta.1` we had updates to the docs? Wouldn't they get lost? What am I missing?

nrb: Good point... we'd actually want `main`, since that's where new docs should go. We shouldn't be putting new docs into the pre-release directories.

Let me actually run the script and try it.

nrb: `main` works as the source, but without some changes to the gen-docs script, it'll block on using the same source directory since https://github.com/vmware-tanzu/velero/blob/main/hack/gen-docs.sh#L30 fails if the directory exists.

I'm inclined to defer this discussion for a separate PR that would make the necessary changes OR just use the versions directly, but we moved to the `-pre` format for a reason; regenerating docs every time we had a beta was cumbersome. Likewise with the patch-level docs - they were cumbersome and took up space for little benefit.

carlisia: Ok. I guess that's why the original instructions says to delete the `$NEW_DOCS_VERSION

image
image

But it also says to run `make gen-docs` for pre-releases because at this point the folder and meta data have been deleted, so I think regenerating the docs will be necessary no matter what.

If I'm right, then we are back at my question: what difference does it make to wipe and recreate a `-pre` directory vs using the entire pre-release name (Ex `1.5.0-rc.1`), and wiping that out and using the GA name (Ex `1.5.0`)?

nrb: I think the difference is semantics at this point, since the instructions say to remove the pre-release stuff each time anyway (we should automate it). Doing the full pre-release version string is more precise, but deviates from how GA works, which leaves off the patch number now (as of v1.4).

Since we're deleting it as we go, I'm not sure I hold a strong opinion either way. For now, I think we can defer the discussion and I'll review the rest of the PR.

hack/release-tools/tag-release.sh
```diff
@@ -15,9 +15,24 @@
...
+# - verify the version string is what the user expects.
+# - create a git tag
+# - push the created git tag to GitHub
+# - run GoReleaser
+
+# The following variables are needed:
+
+# - $VELERO_VERSION: defines the tag of Velero that any https://github.com/vmware-tanzu/velero/...
+#   links in the docs should redirect to.
+# - $publish: TRUE/FALSE value where FALSE (or not including it) will indicate a dry-run, and TRUE, or simply adding 'publish',
```
nrb, line 29: I'd reword this slightly to say `FALSE` is the default behavior.

site/content/docs/main/release-instructions.md
```diff
@@ -1,16 +1,100 @@
...
-## Preparing for a release
-
-### (GA Only) Create release blog post
-
-For each major or minor release, create and publish a blog post to let folks know what's new.
-
+## General notes
+- Please read the documented variables in each script to understand what they are for and how to properly format their values.
+- You will need to have a remote named `upstream` pointing to the Velero repository so the scripts can work properly.
+- GA release: major and minor releases only. Example: 1.0 (major), 1.5 (minor).
```
nrb, line 11: There's also patch releases, for example v1.4.2 would be a GA patch release.

nrb: I should go into a bit more detail here, actually.

Patch releases are usually the 2nd+ within a minor release series. Often it's v1.x.1 and later, but in the case of the v1.5 series, it will be v1.5.2.

These differ because they will have a `release-v1.x` branch to be cut from, because the assumption is `main` will have included features and refactorings that we don't want to backport wholesale.

nrb (approved):
While I've added additional comments here, I don't think they're blocking for the v1.5.1 release, so I'm comfortable addressing them later.

zubron (approved):
Thanks!

PR #2936, opened 2020-09-15
Centralize + rename controller names and list

carlisia:
What do y'all think of this?


pkg/controller/util.go
```diff
@@ -0,0 +1,45 @@
+/*
```
nrb, line 1: Would a better name for this file be "constants.go" maybe? `util.go` is pretty generic, and for now really all it's doing is holding constants.

nrb (commented):
I think this makes total sense! I haven't looked at it in depth yet, but I have one minor comment so far.

pkg/controller/util.go
carlisia: True. I can change this.

pkg/controller/constants.go
```diff
@@ -0,0 +1,45 @@
...
+const (
+	Backup                = "backup"
+	BackupDeletion        = "backup-deletion"
+	BackupStorageLocation = "backup-storage-location"
+	BackupSync            = "backup-sync"
+	DownloadRequest       = "download-request"
+	GarbageCollection     = "garbage-collection"
+	PodVolumeBackup       = "pod-volume-backup"
+	PodVolumeRestore      = "pod-volume-restore"
+	ResticRepo            = "restic-repository"
```
zubron, line 28: Changing this from `restic-repo` to `restic-repository` will change the value that can be passed to the `server --disable-controllers` flag which could be a breaking change for some users. Is the name significant when creating a `genericController`? Could we use `restic-repo` for the name?

carlisia: Good point. There's no big need to spell out the name, I'll change it back.

carlisia:
Addressed all code reviews.

pkg/controller/constants.go
```diff
@@ -0,0 +1,45 @@
...
+
+package controller
+
+const (
+	Backup                = "backup"
+	BackupDeletion        = "backup-deletion"
+	BackupStorageLocation = "backup-storage-location"
+	BackupSync            = "backup-sync"
+	DownloadRequest       = "download-request"
+	GarbageCollection     = "garbage-collection"
```
zubron, line 25: Sorry, I missed this in the first pass :( This will also change the value for `--disable-controllers` like the `restic-repo` value. It's currently `gc` and will be changed to `garbage-collection`

design/cli-install-changes.md
```diff
@@ -68,7 +68,7 @@ Below is the proposed set of new commands to setup and configure Velero.
         --client-burst int                           maximum number of requests by the server to the Kubernetes API in a short period of time (default 30)
         --client-qps float32                         maximum number of requests per second by the server to the Kubernetes API once the burst limit has been reached (default 20)
         --default-backup-ttl duration                how long to wait by default before backups can be garbage collected (default 720h0m0s)
-        --disable-controllers strings                list of controllers to disable on startup. Valid values are backup,backup-sync,schedule,gc,backup-deletion,restore,download-request,restic-repo,server-status-request
+        --disable-controllers strings                list of controllers to disable on startup. Valid values are backup,backup-sync,schedule,gc,backup-deletion,restore,download-request,restic-repository,server-status-request
```
zubron, line 71: Should this also be changed back to `restic-repo`?

zubron (changes requested):
Thanks, @carlisia! I noticed one other issue but once that's resolved I think this will be good to merge.

pkg/controller/constants.go
carlisia: argh! Thank you

zubron (approved):
Thanks!

nrb (approved):

Issue #2948, opened 2020-09-17
CR restore regression in 1.4.2


sseago:
**What steps did you take and what happened:**
Velero restore with CR and CRD in backup sometimes fails to restore the CR.
In my case, backup was from a Kubernetes 1.16.2 cluster (OpenShift 4.2), and restore was to a Kubernetes 1.17.1 cluster (OpenShift 4.4), but since it's a restore error, restore version is probably more relevant.
Backup of a namespace with one running Deployment, one CRD defined, and a CR of the CRD's type in the namespace.
This is the data to back up to reproduce: https://gist.github.com/sseago/e67a060a07cb20affd1e9fe4a7fb14a6
Back up this namespace (CRD will be pulled in as related to the CR), delete the `mycrdtests` namespace and the `deploycustoms.samplecontroller.k8s.io` CRD, and restore.

**What did you expect to happen:**
CR should be restored along with everything else. Instead, the CR was not restored

**The output of the following commands will help us better understand what's going on**:
Relevant error message from Velero:
```
time="2020-09-14T13:48:24Z" level=info msg="Skipping restore of resource because it cannot be resolved via discovery" logSource="pkg/restore/restore.go:383" resource=deploycustoms.samplecontroller.k8s.io restore=openshift-migration/d0865710-f690-11ea-9a47-e12d786f434f-5w6wc
```

**Anything else you would like to add:**
It's a race condition, so it may not always reproduce. Basically the following commit broke the "wait for CRDs to be ready" funcitonality, as well as the "refresh discovery helper after loading CRDs": 
https://github.com/vmware-tanzu/velero/pull/2683/commits/9d246297b800c1ee24ae9983d1e331efdafa25ff

The reason for this is that the string comparison for "is this a CRD restore?" now fails because "customresourcedefinition" != "CustomResourceDefinition". Either `kuberesource.CustomResourceDefinitions` needs to be reverted to using a lowercase Resource name, or the equals tests in restoreItem (linked below) need to compare in a case-insensitive way:
https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L469
https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L1167

**Environment:**

- Velero version (use `velero version`): 1.4.2
- Kubernetes version (use `kubectl version`): 1.17.1 (OpenShift 4.4)
- Cloud provider or hardware configuration: AWS

**Vote on this issue!**

[...]

PR #2949, opened 2020-09-18
restore proper lowercase/plural CRD resource

sseago:
Fixes https://github.com/vmware-tanzu/velero/issues/2948

This commit restores the proper resource string
"customresourcedefinitions" for CRD. The prior change to
"CustomResourceDefinition" was made because this was being used
in another place to populate the CRD "Kind" field in
remap_crd_version_action.go -- there, just use the correct Kind
string instead of pulling from Resource. Using the Kind string in place of Resource was breaking the "wait for CRDs to be ready" functionality, as well as the "refresh discovery helper after loading CRDs" bit.

### DIFF
--- a/changelogs/unreleased/2949-sseago
+++ b/changelogs/unreleased/2949-sseago
@@ -1,0 +1,1 @@
+Restore CRD Resource name to fix CRD wait functionality.
--- a/pkg/backup/remap_crd_version_action.go
+++ b/pkg/backup/remap_crd_version_action.go
@@ -31,7 +31,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	v1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
-	"github.com/vmware-tanzu/velero/pkg/kuberesource"
 	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
 )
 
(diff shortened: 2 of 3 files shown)


changelogs/unreleased/2949-sseago
```diff
@@ -0,0 +1 @@
+Restore lowercase plural resource name for CRDs.
```
nrb, line 1: I might rephrase this to talk about the wait behavior; from a user standpoint the lowercase/plural doesn't mean much.

ashish-amarnath (commented):
Thanks for the bug report and the fix!
This LGTM. Waiting for the changelog update.

nrb:
> This LGTM. Waiting for the changelog update.

Looks like the changelog update has been added, I just didn't resolve the conversation.

changelogs/unreleased/2949-sseago
```diff
@@ -0,0 +1 @@
+Restore CRD Resource name to fix CRD wait functionality.
```
sseago, line 1: @ashish-amarnath I believe this update is in line with what @nrb had suggested. If I misunderstood, or if more than a one-line summary is better, I can change it again.

nrb (approved):

zubron (approved):
Thanks, @sseago! These changes look good to me. I'm going to hold off on merging though and let @ashish-amarnath have the final approval.

zubron:
Merging as Ashish is out of office.

rahul-mereddy:
should I need to include flag --include-cluster-resources during the backup and restore in velero version 1.8.0?

During the recent velero backup and restore on GKE cluster we are seeing CRD's are missing during the restore.

how do I know if the velero backup or restore did the crd's restored?

sseago:
@rahul-mereddy if include-cluster-resources is `true`, then all cluster-scoped resources will be backed up, regardless of whether they're relevant to the namespaces being backed up, which means all CRDs should be included. If it's set to `false` then no cluster-scoped resources will be included, including CRDs. If it's `nil` (the default value), then only relevant cluster-scoped resources will be included. For CRDs, this means that if a CR for `mytype.myorg.org` is backed up, then the `mytype.myorg.org` CRD will be backed up as well, but `irrelevanttype.myorg.org` will not be backed up.

PR #2950, opened 2020-09-18
Allow remote for release process to be configured

zubron:
The release script assumes that the remote for the vmware-tanzu/velero
repository is called `upstream`. It may be the case that this remote is
configured to use a different name. This change updates the script to
allow the remote name being used to be configured by setting the
environment variable `REMOTE` before running the script. If the variable
is not set, the remote defaults to `upstream`.

The release instructions have also been updated to reflect this change.

This change also fixes an issue with the command to push the branch
to the upstream remote as the `--set-upstream` flag takes the remote as
an argument, not the full branch name.

### DIFF
--- a/site/content/docs/main/release-instructions.md
+++ b/site/content/docs/main/release-instructions.md
@@ -7,7 +7,9 @@ This page covers the steps to perform when releasing a new version of Velero.
 
 ## General notes
 - Please read the documented variables in each script to understand what they are for and how to properly format their values.
-- You will need to have a remote named `upstream` pointing to the Velero repository so the scripts can work properly.
+- You will need to have an upstream remote configured to use to the [vmware-tanzu/velero](https://github.com/vmware-tanzu/velero) repository.
+  You can check this using `git remote -v`.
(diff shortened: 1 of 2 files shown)


site/content/docs/main/release-instructions.md
```diff
@@ -7,7 +7,7 @@ This page covers the steps to perform when releasing a new version of Velero.
 
 ## General notes
 - Please read the documented variables in each script to understand what they are for and how to properly format their values.
-- You will need to have a remote named `upstream` pointing to the Velero repository so the scripts can work properly.
+- You will need to have an upstream remote configured to use to the [vmware-tanzu/velero](https://github.com/vmware-tanzu/velero) repository. You can check this using `git remote -v`.
```
nrb, line 10: Should we mention the default of `upstream`?

nrb: Mostly looking for feedback from others here - it's already documented in the script, so it may be fine being omitted here.

zubron: Yep, I can update that. I don't think it hurts to mention the default value again here.

nrb (approved):
Thanks for these improvements!

ashish-amarnath (approved):

PR #2951, opened 2020-09-18
Check existing remote branches in release script

zubron:
This PR depends on #2950 and will be updated if any changes are made there.
It will be marked as a draft until that PR is merged.

The command to check for an existing release branch only checked for
local branches. We should be considering both local and remote branches
before cherry-picking commits for the new release.

This change checks for existing local and remote release branches and
creates or updates them accordingly.
* If a remote branch exists, but a local branch does not, checkout the
  remote branch and track it.
* If the remote branch and local branch exists, checkout the local
  branch and ensure that the latest commits from the remote are pulled.
* Otherwise, if the remote branch does not exist, create it locally if
  needed.

This also handles the case where an existing release branch may be
tracked in multiple remotes as the remote to use is explicitly stated.


nrb (approved):

carlisia (approved):
👍

PR #2952, opened 2020-09-18
Add initial instructions for releasing plugins

nrb:

### DIFF
--- a/site/content/docs/main/plugin-release-instructions.md
+++ b/site/content/docs/main/plugin-release-instructions.md
@@ -4,13 +4,14 @@ layout: docs
 toc: "true"
 ---
 
-# Releasing Velero plugins
-
 Velero plugins maintained by the core maintainers do not have any shipped binaries, only container images, so there is no need to invoke a GoReleaser script.
 Container images are built via a CI job on git push.
 
+Plugins the Velero core team is responsible include all those listed in [the Velero-supported providers list]({{< ref "supported-providers.md#velero-supported-providers" >}}) _except_ the vSphere plugin.
+
 1. Update the README.md file to update the compatibility matrix and `velero install` instructions with the expected version number and open a PR.
+  1. Determining the version number is based on semantic versioning and whether the plugin uses any newly introduced, changed, or removed methods or variables from Velero.
 1. Once the updated README.md PR is merged, checkout the upstream `main` branch. `git checkout upstream/main`.
-1. Tag the git version - `git tag v<version>`
+1. Tag the git version - `git tag v<version>`.
 1. Push the git tag - `git push --tags <upstream remote name>`
 1. Wait for the container images to build
--- a/site/content/docs/main/plugin-release-instructions.md
+++ b/site/content/docs/main/plugin-release-instructions.md
@@ -7,7 +7,7 @@ toc: "true"
 Velero plugins maintained by the core maintainers do not have any shipped binaries, only container images, so there is no need to invoke a GoReleaser script.
 Container images are built via a CI job on git push.
 
-Plugins the Velero core team is responsible include all those listed in [the Velero-supported providers list]({{< ref "supported-providers.md#velero-supported-providers" >}}) _except_ the vSphere plugin.
+Plugins the Velero core team is responsible include all those listed in [the Velero-supported providers list](supported-providers.md) _except_ the vSphere plugin.
 
 1. Update the README.md file to update the compatibility matrix and `velero install` instructions with the expected version number and open a PR.
   1. Determining the version number is based on semantic versioning and whether the plugin uses any newly introduced, changed, or removed methods or variables from Velero.
--- a/site/content/docs/main/plugin-release-instructions.md
+++ b/site/content/docs/main/plugin-release-instructions.md
@@ -9,6 +9,8 @@ toc: "true"
 Velero plugins maintained by the core maintainers do not have any shipped binaries, only container images, so there is no need to invoke a GoReleaser script.
 Container images are built via a CI job on git push.
 
+1. Update the README.md file to update the compatibility matrix with the expected version number and open a PR.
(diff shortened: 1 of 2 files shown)


site/content/docs/main/plugin-release-instructions.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Releasing Velero plugins
+layout: docs
+toc: "true"
+---
+
+# Releasing Velero plugins
```
nrb, line 7: Also needs notes on README compatibility matrix and updated `velero install` instructions.

site/content/docs/main/plugin-release-instructions.md
```diff
@@ -0,0 +1,16 @@
...
+
+Velero plugins maintained by the core maintainers do not have any shipped binaries, only container images, so there is no need to invoke a GoReleaser script.
+Container images are built via a CI job on git push.
+
+1. Update the README.md file to update the compatibility matrix and `velero install` instructions with the expected version number and open a PR.
+1. Once the updated README.md PR is merged, checkout the upstream `main` branch. `git checkout upstream/main`.
```
zubron, line 13: The vSphere plugin does not use `main` as its default branch, it uses `master`. I also noticed that it doesn't have any GitHub releases, only tags. Are the core Velero team involved in that plugin's release process?

site/content/docs/main/plugin-release-instructions.md
```diff
@@ -0,0 +1,16 @@
...
+layout: docs
+toc: "true"
+---
+
+# Releasing Velero plugins
+
+Velero plugins maintained by the core maintainers do not have any shipped binaries, only container images, so there is no need to invoke a GoReleaser script.
+Container images are built via a CI job on git push.
+
+1. Update the README.md file to update the compatibility matrix and `velero install` instructions with the expected version number and open a PR.
```
zubron, line 12: I don't know if it fits into this document, but how do we determine what the new version of a plugin in should be? Also, what affects a plugins compatibility with particular velero versions?

site/content/docs/main/plugin-release-instructions.md
nrb: We are not - that's a wholly separate team.

I should enumerate the plugins we're responsible as part of this PR.

site/content/docs/main/plugin-release-instructions.md
nrb: New versions are usually based on semantic versioning - honestly we may not actually need new releases for AWS and GCP for v1.5.x, but I need to look at what actually changed.

However, for the CSI plugin, the new version will be using API calls introduced in Velero v1.5.1 (https://github.com/vmware-tanzu/velero-plugin-for-csi/pull/70), so we'll need to make sure that's called out. Other compatibility concerns are things like the plugin interface itself changing with new methods being added, changed, or removed.

site/content/docs/main/plugin-release-instructions.md
nrb: How about I link to the table at https://velero.io/docs/v1.5/supported-providers/ and mention that the vSphere plugin is an exception? Is that workable?

zubron: Yeah, that works for me :) I had initially gone to that page but got a bit confused by the difference in the vSphere repo. I knew that there was a different team involved, but it wasn't clear if the core team was involved in the release process.

site/content/docs/main/plugin-release-instructions.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Releasing Velero plugins
+layout: docs
+toc: "true"
+---
+
+Velero plugins maintained by the core maintainers do not have any shipped binaries, only container images, so there is no need to invoke a GoReleaser script.
+Container images are built via a CI job on git push.
+
+Plugins the Velero core team is responsible include all those listed in [the Velero-supported providers list]({{< ref "supported-providers.md#velero-supported-providers" >}}) _except_ the vSphere plugin.
```
nrb, line 10: Trying this shortcode to link directly to the anchor, but it wasn't working locally despite a lot of different permutations. I'm seeing if it's solely something in my own set up, or something truly wrong.

nrb: Yeah, the links are still broken for some reason. It looks like the examples in https://gohugo.io/content-management/shortcodes/#ref-and-relref, but I'm not exactly sure what's going on.

nrb: @carlisia @zubron My desire here was to link directly to the table on the page with an anchor. As the primary consumers of the document, do you think it's worth spending more time trying to figure out why the shortcode isn't generating the right URL, or do you think it's acceptable to just link to the page?

Another alternative is doing a direct URL link, I suppose.

nrb: I can't get a direct link to work, either, so I'm removing it for the sake of moving the PR forward. There might be some Hugo setting I'm missing that causes the actual URL to end up empty.

site/content/docs/main/plugin-release-instructions.md
ashish-amarnath: We could also update https://velero.io/plugins/ to indicate that the vSpehere plugin is supported by vmware and not the velero team. That way this doc will not be relavent to the vsphere plugin.

ashish-amarnath (approved):
This doc is in good shape and ready to merge. We can update the https://velero.io/plugins/ or https://velero.io/docs/v1.5/supported-providers/ in a follow-up PR.

This doc not getting merged may be delaying the release of the plugins that our users have been asking for and we should get this merged and release the plugins.

nrb:
> This doc not getting merged may be delaying the release of the plugins that our users have been asking for and we should get this merged and release the plugins.

Yeah, honestly I think we can/should release the plugins without waiting at this point, particularly Azure and CSI.

carlisia (approved):

PR #2953, opened 2020-09-18
Don't attempt to publish docker images on forks

nrb:
When the Main CI workflow runs on a fork, the docker push step will
always fail because the appropriate secrets are missing. This is
annoying at best and causes CI on forks to be untrustworthy at worst.


nrb:
A forked CI run: https://github.com/nrb/velero/runs/1135437253?check_suite_focus=true

zubron (approved):

carlisia (approved):

Issue #2311, opened 2020-03-02
Installation of v1.3.0 fails on Kubernetes 1.14


bgagnon:
**What steps did you take and what happened:**
Tried to install Velero v1.3.0 on a K8S 1.14 cluster.

**What did you expect to happen:**
Successful installation; [the documentation](https://velero.io/docs/v1.3.0/basic-install/#prerequisites) says that Kubernetes 1.10+ is supported.

**Error messages**

Every CRD returns an error when going through `kubectl apply`:

```
[error validating "build\\manifests\\velero\\crd\\backups.velero.io-customresourcedefinition.json":
error validating data: ValidationError(CustomResourceDefinition.spec):
unknown field "preserveUnknownFields" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionSpec;
if you choose to ignore these errors, turn validation off with --validate=false,
```

**Environment:**

- Velero v1.3.0
- Kubernetes v1.14.1
- Kubernetes client v1.15.4

bgagnon:
This option was first introduced in K8S 1.15; this explains the failure on 1.14.
* 1.14: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#customresourcedefinitionspec-v1beta1-apiextensions-k8s-io
* 1.15: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#customresourcedefinitionspec-v1beta1-apiextensions-k8s-io

skriss:
@bgagnon you can provide the `--validate=false` flag to `kubectl apply` to bypass this error. We have this documented for 1.13 and earlier clusters (see https://velero.io/docs/v1.3.0/customize-installation/#generate-yaml-only), but looks like we need to update that to apply to 1.14 and earlier.

Issue #2077, opened 2019-11-22
velero v1.2.0 --dry-run -o yaml inserts "nullable" in generated kube manifest files


cmcconnell1:
**What steps did you take and what happened:**

generate kube manifest files from the velero binary and then attempt to apply to cluster via `kubectl apply`
```console
velero install \
    --provider aws \
    --plugins velero/velero-plugin-for-aws:${VELERO_AWS_PLUGIN_VERSION} \
    --bucket ${VELERO_BUCKET_NAME} \
    --backup-location-config region=$VELERO_AWS_REGION \
    --snapshot-location-config region=$VELERO_AWS_REGION \
    --secret-file ./velero-credentials.secret \
    --dry-run -o yaml \
> ./manifests/velero/velero.yaml

works fine but cannot apply them to cluster as shown below. 

**What did you expect to happen:**
Expected generated kube manifest files to be usable per the documentaion and be able to be applied to kubernetes `kubectl apply` without errors. 

**Errors during kubectl apply**:
```console
kubectl apply --recursive -f $MANIFEST_PATH -n velero
error: error validating "manifests/velero/velero.yaml": error validating data: [ValidationError(CustomResourceDefinition.spec.validation.openAPIV3Schema.properties.spec.properties.excludedNamespaces): unknown field "nullable"
```
Note: there are ~15 more errors like above for every 'nullable' reference autogenerated in the velero manifest files...

**Anything else you would like to add:**

**Environment:**

```console
velero version
Client:
  Version: v1.2.0
  Git commit: -
Server:
Version: (cannot apply manifests therefore cannot provide server info as it errors-out)
```

- Velero features (use `velero client config get features`): 
- Kubernetes version (use `kubectl version`):
```console
 k version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:41:55Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.12-eks-c500e1", GitCommit:"c500e11584c323151d6ab17526d1ed7461e45b0c", GitTreeState:"clean", BuildDate:"2019-10-22T03:11:52Z", GoVersion:"go1.11.13", Compiler:"gc", Platform:"linux/amd64"}
```

[...]

PR #2969, opened 2020-09-23
Include --validate=false in upgrade instructions

zubron:
We instruct users to update the CRDs when upgrading to 1.4 and 1.5 which
involves using `kubectl apply` to apply the CRD configuration. The CRD
configuration generated by `velero install` includes fields which are
not valid when running Kubernetes v1.14 or earlier. We instruct users to
work around this when doing a customised velero install, but not when
upgrading to newer versions. This change updates the upgrade
instructions for v1.4 and v1.5 to include the use of `--validate=false`
flag when running `kubectl apply`.

See #2077 and #2311 for more context.

### DIFF
--- a/site/content/docs/main/upgrade-to-1.5.md
+++ b/site/content/docs/main/upgrade-to-1.5.md
@@ -38,6 +38,8 @@ If you're not yet running at least Velero v1.4, see the following:
     velero install --crds-only --dry-run -o yaml | kubectl apply -f -
     ```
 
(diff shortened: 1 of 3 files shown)


nrb (approved):
Thanks for adding this!

I think adding the note for v1.4 makes sense, as some users may be stepping through releases to get the upgrades.

zubron:
I just realised that I should have added it to `main` as well and that I had broken the links in the 1.5 instructions :woman_facepalming: I've fixed it now!

nrb (approved):
Good catch :)

carlisia (approved):
Good eye!

PR #2970, opened 2020-09-24
fixing 'velero.io/change-pvc-node-selector' plugin to fetch configmap using plugin name

mynktl:
`velero.io/change-pvc-node-selector` plugin is using the label key "`velero.io/change-pvc-node`" to fetch the configmap for node mapping instead of plugin name. This PR fixes the `velero.io/change-pvc-node-selector` plugin to fetch the node-mapping details using the plugin name.

### DIFF
--- a/site/content/docs/main/restore-reference.md
+++ b/site/content/docs/main/restore-reference.md
@@ -86,7 +86,7 @@ metadata:
     velero.io/plugin-config: ""
     # this label identifies the name and kind of plugin
     # that this ConfigMap is for.
-    velero.io/change-pvc-node-selector: RestoreItemAction
+    velero.io/change-pvc-node: RestoreItemAction
 data:
   # add 1+ key-value pairs here, where the key is the old
   # node name and the value is the new node name.
--- a/site/content/docs/main/restore-reference.md
+++ b/site/content/docs/main/restore-reference.md
@@ -86,7 +86,7 @@ metadata:
     velero.io/plugin-config: ""
     # this label identifies the name and kind of plugin
     # that this ConfigMap is for.
-    velero.io/change-pvc-node: RestoreItemAction
+    velero.io/change-pvc-node-selector: RestoreItemAction
 data:
   # add 1+ key-value pairs here, where the key is the old
   # node name and the value is the new node name.
--- a/changelogs/unreleased/2970-mynktl
+++ b/changelogs/unreleased/2970-mynktl
@@ -1,0 +1,1 @@
+Fixed 'velero.io/change-pvc-node-selector' plugin to fetch configmap using label key "velero.io/change-pvc-node-selector" 
--- a/pkg/restore/change_pvc_node_selector.go
+++ b/pkg/restore/change_pvc_node_selector.go
@@ -130,7 +130,7 @@ func (p *ChangePVCNodeSelectorAction) Execute(input *velero.RestoreItemActionExe
 
 func getNewNodeFromConfigMap(client corev1client.ConfigMapInterface, node string) (string, error) {
 	// fetch node mapping from configMap
-	config, err := getPluginConfig(framework.PluginKindRestoreItemAction, "velero.io/change-pvc-node", client)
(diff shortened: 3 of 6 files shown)


site/content/docs/main/restore-reference.md
```diff
@@ -86,7 +86,7 @@ metadata:
     velero.io/plugin-config: ""
     # this label identifies the name and kind of plugin
     # that this ConfigMap is for.
-    velero.io/change-pvc-node-selector: RestoreItemAction
+    velero.io/change-pvc-node: RestoreItemAction
```
nrb, line 89: I think this change is incorrect, based on [the custom plugin configuration documentation](https://velero.io/docs/v1.5/custom-plugins/#plugin-configuration). The key should be the fully qualified name of the plugin here, which is `velero.io/change-pvc-node-selector`. That's also how the [plugin is registered](https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/server/plugin/plugin.go#L56).

So I think the actual change that should happen is that the code should be looking for the `velero.io/change-pvc-node-selector` ConfigMap in order to be consistent with documentation.

nrb (changes requested):

site/content/docs/main/restore-reference.md
nrb: Another option would be to change the name that the plugin is registered under, though it's less descriptive.

mynktl:
Hi @nrb, Thanks for reviewing the PR. I've updated the relevant plugin code to fetch the config using the plugin name.

nrb (dismissed):
Awesome, thanks for updating this @mynktl!

Now that this is a code change, could you also add a changelog file? I think that's all that's needed to get this to merge from my point of view.

site/content/docs/main/restore-reference.md
pawanpraka1: @noamran change-pvc-node should suffice as it is expected to be the label in the config map, so do we need say selector (change-pvc-node-selector) specifically? For example if we check other plugins (e.g. "velero.io/change-storage-class") they don't mention selector. What do you think, should we just change the name that the plugin is registered under? Which one is the better solution?

mynktl:
Hi @nrb
I've added the changelog file. PTAL.
Thanks!

site/content/docs/main/restore-reference.md
nrb: @pawanpraka1 I'm not sure what you mean about the other plugin not using a selector. [I see that the plugin looks up a ConfigMap with `velero.io/change-storage-class`](https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/change_storageclass_action.go#L69), and if I look at the code for that function, [it's using a LabelSelector](https://github.com/vmware-tanzu/velero/blob/a841167ee0a7ac61d3f829220eb2d11cb5fea5da/pkg/restore/restic_restore_action.go#L255). It needs a ConfigMap with the labels `velero.io/plugin-config` and `<fully-qualified-plugin-name>`.

It's not enforced in code that the label name match exactly; however, it's a [documented convention](https://velero.io/docs/v1.5/custom-plugins/#plugin-configuration) that I'd rather not break on included plugins.

If it's easier for users to not change the ConfigMap's label, then changing the plugin's registered name could work, too.

pawanpraka1: @nrb I meant to say selector in the name.

pawanpraka1: Do we need to specify selector in the plugin name.

nrb: @pawanpraka1 Ah, I see.

The only requirement is that the plugin and ConfigMap match, from my point of view. Updating the code to look for a new ConfigMap label may be disruptive to users, who might have existing ConfigMaps. It may be easier to simply change the plugin's registered name. Having `selector` in the name is more descriptive, and I think that is the proper name of the field in the spec, but I'm also not too worried about that precision personally.

pawanpraka1: I will leave the decision upto you @nrb . I am fine with both the solution.

nrb (approved):

ashish-amarnath (approved):

PR #2992, opened 2020-10-07
Fix BSL controller to avoid invoking init() on all BSLs regardless of…

### LINKED ISSUE #2991: BSL controller loop always invokes init() regardless of backup location's ValidationFrequency
After upgrading to Velero v1.5.1, we observed that BSL controller always invokes init() for existing backup storage locations regardless of the BSL's ValidationFrequency (we tried setting BSL validationFrequency to 0 to skip validation but this didn't help). This is an issue since for some objectstore plugins the init() method could be an expensive call, for instance our objectstore plugin init() method opens connection to the backup target and creates directories/buckets  if they don't exist as part of initializing the backup repository. Looking at the Velero [BSL controller code](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/backup_storage_location_controller.go#L80), we first initialize backupStore prior to the check for whether the BSL is ready to be validated:

```
                backupStore, err := r.NewBackupStore(location, pluginManager, log)
		if err != nil {
			log.WithError(err).Error("Error getting a backup store")
			continue
		}

		if !storage.IsReadyToValidate(location.Spec.ValidationFrequency, location.Status.LastValidationTime, r.DefaultBackupLocationInfo, log) {
			log.Debug("Backup location not ready to be validated")
			continue
		}
```
I think moving r.NewBackupStore() to after the check for whether BSL is ready to be validated will fix this issue to avoid invoking init() on all backup storage locations whenever the controller loop runs.

Setting BSL validationFrequency to 0 should skip validation for the backup storage location.

**Environment:**

- Velero version (use `velero version`): v1.5.1
- Kubernetes version (use `kubectl version`):1.18

**Vote on this issue!**

betta1:
… ValidationFrequency

This fixes #2991 

Signed-off-by: Bett, Antony <antony.bett@dell.com>

### DIFF
--- a/pkg/controller/backup_storage_location_controller.go
+++ b/pkg/controller/backup_storage_location_controller.go
@@ -77,14 +77,14 @@ func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 			defaultFound = true
 		}
 
-		backupStore, err := r.NewBackupStore(location, pluginManager, log)
-		if err != nil {
-			log.WithError(err).Error("Error getting a backup store")
+		if !storage.IsReadyToValidate(location.Spec.ValidationFrequency, location.Status.LastValidationTime, r.DefaultBackupLocationInfo, log) {
+			log.Debug("Backup location not ready to be validated")
 			continue
 		}
 
-		if !storage.IsReadyToValidate(location.Spec.ValidationFrequency, location.Status.LastValidationTime, r.DefaultBackupLocationInfo, log) {
-			log.Debug("Backup location not ready to be validated")
+		backupStore, err := r.NewBackupStore(location, pluginManager, log)
+		if err != nil {
+			log.WithError(err).Error("Error getting a backup store")
 			continue
 		}
 
--- a/changelogs/unreleased/2992-betta1
+++ b/changelogs/unreleased/2992-betta1
@@ -1,0 +1,1 @@
+Fix BSL controller to avoid invoking init() on all BSLs regardless of ValidationFrequency


pkg/controller/backup_storage_location_controller.go
```diff
@@ -77,14 +77,14 @@ func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 			defaultFound = true
 		}
 
-		backupStore, err := r.NewBackupStore(location, pluginManager, log)
-		if err != nil {
-			log.WithError(err).Error("Error getting a backup store")
+		if !storage.IsReadyToValidate(location.Spec.ValidationFrequency, location.Status.LastValidationTime, r.DefaultBackupLocationInfo, log) {
```
nrb, line 80: I think this only addresses the case where validation frequency is 0, unless I'm misreading something.

If validation frequency is 1 minute (the server's default), then it will still invoke the Init() function. Is that acceptable in your case?

nrb (approved):
I haven't run this against a cluster, but I think this does fix the issue of invoking the plugin's Init() function every reconciliation loop (roughly once per second), since it exits the loop iteration early based on the validation frequency, which is roughly once a minute based on the server default.

There may be a longer term fix here where we can instantiate a plugin process and let it live longer, but I think that has bigger implications that I don't understand at the moment.

pkg/controller/backup_storage_location_controller.go
betta1: This change ensures we first check if the BSL is ready to be validated and then initialize the backup store only for those BSLs that pass the check. It should address all cases where validation frequency is 0 as well as user specified validation frequency or server's default validation frequency.

carlisia (approved):
This was a very good catch, thank you. 👍

carlisia:
I did run this against a cluster.

betta1:
Thanks @carlisia @nrb

PR #2993, opened 2020-10-07
Don't fail backup deletion if downloading tarball fails

### LINKED ISSUE #2980: Unable to delete backup if cloud resources have already been deleted (velero v1.5.1, similar to #308)
As issue #308 is closed, I opened this new one as I'm facing the same issue but with v1.5.1

The blobs have already been deleted in the cloud and velero is unable to delete the backup

    velero backup get
    velero-s-24h-20200831015039  Completed   0   12   2020-08-31 01:32:51 -0400 EDT   17h       default            <none>
    velero-s-24h-20200826022813  Deleting    0   0    2020-08-26 01:47:20 -0400 EDT   4d ago    default            <none>

Velero logs (Split for readability): 

    level=error msg="Error in syncHandler, re-adding item to queue" controller=backup-deletion error="error downloading backup: error copying Backup to temp file: 
    rpc error: code = Unknown desc = storage: service returned error: StatusCode=404, ErrorCode=BlobNotFound, ErrorMessage=The specified blob does not exist.\nRequestId:75a99b8d-f01e-004e-175c-963ab1000000\nTime:2020-09-29T12:30:26.7728994Z, 
    RequestInitiated=Tue, 29 Sep 2020 12:30:26 GMT, RequestId=75a99b8d-f01e-004e-175c-963ab1000000, API Version=2018-03-28, QueryParameterName=, QueryParameterValue=" error.file="/go/src/github.com/vmware-tanzu/velero/pkg/controller/restore_controller.go:558" 
    error.function=github.com/vmware-tanzu/velero/pkg/controller.downloadToTempFile key=velero/velero-s-24h-20200826022813-766pb logSource="pkg/controller/generic_controller.go:140"

zubron:
Previously, we would always attempt to download the tarball for a backup
for processing DeleteItemAction plugins, even if there weren't any.
This caused an issue for some users in the case where the backup tarball
had been deleted from object storage as the backup deletion would fail.

Now, we only attempt to download the tarball in the case where there are
DeleteItemAction plugins. If downloading that tarball fails, we log
the error, skip the processing of the DeleteItemAction plugins and
proceed with the rest of the deletion.

Fixes #2980 

### DIFF
--- a/pkg/controller/backup_deletion_controller.go
+++ b/pkg/controller/backup_deletion_controller.go
@@ -305,12 +305,11 @@ func (c *backupDeletionController) processRequest(req *velerov1api.DeleteBackupR
 	if len(actions) > 0 {
 		// Download the tarball
 		backupFile, err := downloadToTempFile(backup.Name, backupStore, log)
+		defer closeAndRemoveFile(backupFile, c.logger)
 
 		if err != nil {
 			log.WithError(err).Errorf("Unable to download tarball for backup %s, skipping associated DeleteItemAction plugins", backup.Name)
 		} else {
-			defer closeAndRemoveFile(backupFile, c.logger)
-
 			ctx := &delete.Context{
 				Backup:          backup,
 				BackupReader:    backupFile,
@@ -327,7 +326,6 @@ func (c *backupDeletionController) processRequest(req *velerov1api.DeleteBackupR
 				return errors.Wrap(err, "error invoking delete item actions")
 			}
 		}
-
 	}
 
 	if backupStore != nil {
--- a/pkg/controller/backup_deletion_controller.go
+++ b/pkg/controller/backup_deletion_controller.go
@@ -295,13 +295,6 @@ func (c *backupDeletionController) processRequest(req *velerov1api.DeleteBackupR
 		errs = append(errs, err.Error())
 	}
 
-	// Download the tarball
-	backupFile, err := downloadToTempFile(backup.Name, backupStore, log)
-	if err != nil {
-		return errors.Wrap(err, "error downloading backup")
-	}
-	defer closeAndRemoveFile(backupFile, c.logger)
-
 	actions, err := pluginManager.GetDeleteItemActions()
 	log.Debugf("%d actions before invoking actions", len(actions))
 	if err != nil {
@@ -309,20 +302,32 @@ func (c *backupDeletionController) processRequest(req *velerov1api.DeleteBackupR
 	}
 	// don't defer CleanupClients here, since it was already called above.
 
-	ctx := &delete.Context{
-		Backup:          backup,
-		BackupReader:    backupFile,
-		Actions:         actions,
-		Log:             c.logger,
-		DiscoveryHelper: c.helper,
-		Filesystem:      filesystem.NewFileSystem(),
-	}
+	if len(actions) > 0 {
+		// Download the tarball
+		backupFile, err := downloadToTempFile(backup.Name, backupStore, log)
 
-	// Optimization: wrap in a gofunc? Would be useful for large backups with lots of objects.
-	// but what do we do with the error returned? We can't just swallow it as that may lead to dangling resources.
-	err = delete.InvokeDeleteActions(ctx)
-	if err != nil {
-		return errors.Wrap(err, "error invoking delete item actions")
+		if err != nil {
+			log.WithError(err).Errorf("Unable to download tarball for backup %s, skipping associated DeleteItemAction plugins", backup.Name)
+		} else {
+			defer closeAndRemoveFile(backupFile, c.logger)
+
+			ctx := &delete.Context{
+				Backup:          backup,
+				BackupReader:    backupFile,
+				Actions:         actions,
+				Log:             c.logger,
+				DiscoveryHelper: c.helper,
+				Filesystem:      filesystem.NewFileSystem(),
+			}
+
+			// Optimization: wrap in a gofunc? Would be useful for large backups with lots of objects.
+			// but what do we do with the error returned? We can't just swallow it as that may lead to dangling resources.
+			err = delete.InvokeDeleteActions(ctx)
+			if err != nil {
+				return errors.Wrap(err, "error invoking delete item actions")
+			}
+		}
+
 	}
 
 	if backupStore != nil {
--- a/pkg/controller/backup_deletion_controller_test.go
+++ b/pkg/controller/backup_deletion_controller_test.go
@@ -49,6 +49,8 @@ import (
 	persistencemocks "github.com/vmware-tanzu/velero/pkg/persistence/mocks"
 	"github.com/vmware-tanzu/velero/pkg/plugin/clientmgmt"
 	pluginmocks "github.com/vmware-tanzu/velero/pkg/plugin/mocks"
+	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+	"github.com/vmware-tanzu/velero/pkg/plugin/velero/mocks"
 	velerotest "github.com/vmware-tanzu/velero/pkg/test"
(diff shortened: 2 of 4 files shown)


pkg/controller/backup_deletion_controller_test.go
```diff
@@ -739,6 +741,265 @@ func TestBackupDeletionControllerProcessRequest(t *testing.T) {
 		// Make sure snapshot was deleted
 		assert.Equal(t, 0, td.volumeSnapshotter.SnapshotsTaken.Len())
 	})
+
+	t.Run("backup is not downloaded when there are no DeleteItemAction plugins", func(t *testing.T) {
```
zubron, line 745: I've used the same test set up as the previous tests but there is a lot of repetition here now and I think they could be refactored. I'm happy to do that in this PR or I can do a follow up PR.

pkg/controller/backup_deletion_controller.go
```diff
@@ -295,34 +295,39 @@ func (c *backupDeletionController) processRequest(req *velerov1api.DeleteBackupR
...
-		Log:             c.logger,
-		DiscoveryHelper: c.helper,
-		Filesystem:      filesystem.NewFileSystem(),
-	}
+	if len(actions) > 0 {
+		// Download the tarball
+		backupFile, err := downloadToTempFile(backup.Name, backupStore, log)
+
+		if err != nil {
+			log.WithError(err).Errorf("Unable to download tarball for backup %s, skipping associated DeleteItemAction plugins", backup.Name)
```
zubron, line 310: Does this approach seem okay? Unfortunately we can't distinguish the reasons why downloading the backup tarball fails. It could be missing, or it could be a temporary issue. With this approach, if it ever fails we will always skip the plugins and proceed with deleting the rest of the backup resources.

nrb: I think some retry logic in the backupStore.GetBackupContents function would be helpful for sure, so we can account for the transient failures better.

I don't know if it needs to flow down to the ObjectStore plugin interface though - if so, I could see it growing at least a couple new methods and settings (number of retries, max timeout, whether or not to do exponential backoff, etc). I'd have to think more about that, but I think a plugin-level change isn't right for this PR.

nrb: Another option is to extend our error types to help us distinguish, though this would likely also require participation by plugin authors to return the right ones.

Kubernetes has this concept, and we use it in various places, such as [backup.go](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/backup.go#L490)

pkg/controller/backup_deletion_controller_test.go
carlisia: I think you should open a ticket (to track) and do this separately.

pkg/controller/backup_deletion_controller.go
zubron: Hmm, I've been testing this further and from my local setup, the error doesn't come from `backupStore.GetBackupContents` but actually comes from trying to read from the resulting `io.ReadCloser` returned by `GetBackupContents`. I had added some retry logic to that function but couldn't get it to trigger. I don't know if it's worth keeping it in.

nrb: Interesting - so you get an `io.ReadCloser`, but it's not valid? Is it `nil` when that happens?

We might just have to revisit the retry logic separately for now.

zubron: It's a valid `io.ReadCloser` but it's a [`StreamReadCloser`](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/framework/stream_reader.go#L34-L40). `GetBackupContents` eventually calls the grpc `GetObject` method and it [returns a stream](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/proto/ObjectStore.proto#L81) which we don't attempt to read from until we call [`io.Copy`](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/restore_controller.go#L556) later in the [`downloadToTempFile` function](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/restore_controller.go#L556).

I initially thought this might have been a plugin specific implementation issue, but it seems to be the case across two object store plugins that I've checked.  I'll revert the retry logic for now.

ashish-amarnath:
I think the correct fix is to not delete the backup from the object store if there were any errors in the process of deleting the backup.

refer https://github.com/vmware-tanzu/velero/issues/3094#issuecomment-730063746

ashish-amarnath (changes requested):
I think the correct fix is to not delete the backup from the object store if there were any errors in the process of deleting the backup.

refer https://github.com/vmware-tanzu/velero/issues/3094#issuecomment-730063746

zubron:
@ashish-amarnath #3094 is related but I think this PR and the issue it is addressing are different enough that they should be considered separately. This PR will only deal with the case where there are DeleteItemAction plugins to run and the backup tarball doesn't exist. The fact that a backup deletion fails due to missing cloud resources is a change in behaviour from 1.4 so this PR was to restore the previous behaviour. For example, if someone manually removes the backup tarball from their object storage or the backup tarball was never written in the first place, then attempting to delete those backups through velero will *always* result in an error and the deletion will fail.

In #3094, the issue is that Velero should not have deleted any backup resources in the first place if there was an error. They are related, in that the error described in #2980 will appear on following deletion attempts, but I still think they are distinct issues. I think the issue of orphaned resources described in #3094 would have existed prior to the introduction of DeleteItemActions.

All that said, I think we need both this change and the one you mentioned in #3094 :smile:

pkg/controller/backup_deletion_controller.go
```diff
@@ -295,34 +295,39 @@ func (c *backupDeletionController) processRequest(req *velerov1api.DeleteBackupR
...
+		} else {
+			defer closeAndRemoveFile(backupFile, c.logger)
```
nrb, line 312: Should this `defer` be in the else block? I don't think the `backupFile` variable exists here, unless I'm reading something wrong.

nrb:
I think we need both fixes as well. If nothing else, this is an optimization that we won't try to download the backup tarball when we don't need to act on it.

pkg/controller/backup_deletion_controller.go
zubron: It's defined on 307. I put the `defer` here thinking that we shouldn't attempt to close and remove it unless it was successfully created, but I should check the implementation of both functions. `downloadToTempFile` might create the file and not write to it, or `closeAndRemoveFile` might be able to handle the case where it doesn't exist and in that case it would be safe to move it to the outer scope where `backupFile` was defined.

zubron: `downloadToTempFile` returns a `nil` file pointer in the case where there was an error. This means that `closeAndRemoveFile` will panic as it was passed `nil`. I can leave the `defer` in the `else` block or update `closeAndRemoveFile` to handle the case where the file is `nil`. I'm leaning towards the latter option but let me know what you think!

nrb: Got it, thanks for the detailed reply. I was actually confusing the `if`s and thought the `else` was part of the `len(actions) > 0` condition.

I think updating `closeAndRemoveFile` to handle a `nil` file is a good idea. That way, we can move the `defer` right after the `downloadToTempFile` call, which is more idiomatic Go, where the `defer` is put right after the creation or fetching of the resource.

The other option is a short comment above the `defer` statement where it's at now to explain why, but I think updating `closeAndRemoveFile` is the better choice.

nrb (approved):
Thanks for your patience @zubron :)

@ashish-amarnath Could you please revisit your review?

ashish-amarnath (approved):
🚀

haslersn:
When will this fix be released?

Issue #2884, opened 2020-09-01
Minor kubebuilder related items to clean up


carlisia:
Listing things to clean up:

- [x] remove rbac/role file (config/rbac/role.yaml), as well as it's generation in the script (I think this will work: https://github.com/vmware-tanzu/velero/blob/main/hack/update-generated-crd-code.sh#L50)

- [x] maybe remove the `samples` folder; re: https://github.com/vmware-tanzu/velero/pull/2838#discussion_r480258423

- [x] use CAPI's patch function to update resources

- [x] name individual test cases in the controller tests; re: https://github.com/vmware-tanzu/velero/pull/2838#discussion_r481367793

- [x] rename `ServerStatusInfo` (this shad already been done)

- [x] add test to the server disabling controller check.

- [x] maybe refactor controller key stuff into the controller package - PR: https://github.com/vmware-tanzu/velero/pull/2936

PR #3004, opened 2020-10-12
Convert DownloadRequest resource/controller to kubebuilder 

carlisia:
This closes https://github.com/vmware-tanzu/velero/issues/2656.


nrb:
@carlisia I did a quick pass over the code, but is there anything in particular preventing it from moving out of draft right now?

carlisia:
> @carlisia I did a quick pass over the code, but is there anything in particular preventing it from moving out of draft right now?

Yes, I need to fix this test: https://github.com/vmware-tanzu/velero/pull/3004/files#diff-6d5dc1afa6af05350b9f38368bde033f93c478b86304922796d2e6eedcbf1336

carlisia:
Note: @nrb agreed that the test pkg/cmd/util/downloadrequest/downloadrequest_test.go should be removed. To be sure, the controller has thorough testing.

config/rbac/role.yaml
```diff
@@ -26,6 +26,26 @@ rules:
...
+  resources:
+  - downloadrequests
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
nrb, line 40: Since we didn't implement a `watch` in the controller, should we remove this line?

pkg/cmd/util/downloadrequest/downloadrequest.go
```diff
@@ -29,74 +29,51 @@ import (
...
-				}
-			}
+	key := client.ObjectKey{Name: created.Name, Namespace: namespace}
+	checkFunc := func() {
+		updated := &velerov1api.DownloadRequest{}
+		if err := kbClient.Get(ctx, key, updated); err != nil {
+			return
+		}
+
+		// TODO: once the minimum supported Kubernetes version is v1.9.0, remove the following check.
```
nrb, line 62: Pretty sure we can actually remove these checks now. However, let's not in this PR :)

pkg/cmd/util/downloadrequest/downloadrequest.go
```diff
@@ -29,74 +29,51 @@ import (
...
-			Name:      fmt.Sprintf("%s-%s", name, time.Now().Format("20060102150405")),
-		},
-		Spec: v1.DownloadRequestSpec{
-			Target: v1.DownloadTarget{
-				Kind: kind,
-				Name: name,
-			},
-		},
-	}
+func Stream(ctx context.Context, kbClient kbclient.Client, namespace, name string, kind velerov1api.DownloadTargetKind, w io.Writer, timeout time.Duration, insecureSkipTLSVerify bool, caCertFile string) error {
```
nrb, line 44: I don't see the `timeout` argument used anywhere in the new code - am I missing it?

nrb (changes requested):
A couple questions, and I think one of the parameters passed in to the `Stream` function is no longer used now.

pkg/cmd/util/downloadrequest/downloadrequest.go
zubron: It's used on line 105.

config/rbac/role.yaml
carlisia: The very first bullet on this issue is to clean this up: https://github.com/vmware-tanzu/velero/issues/2884

I'll be trying to avoid this from being generated too.

pkg/cmd/util/downloadrequest/downloadrequest.go
carlisia: Created an issue for this: https://github.com/vmware-tanzu/velero/issues/3126

carlisia:
Comments so far have been addressed, ready for review.

nrb (commented):
Dismissing request for changes, but looks like a rebase is necessary.

carlisia:
Rebased, ready for review.

nrb (dismissed):

nrb:
@vmware-tanzu/velero-maintainers Please queue this for review before break

pkg/controller/download_request_controller.go
```diff
@@ -18,229 +18,137 @@ package controller
...
-	updatedBytes, err := json.Marshal(updated)
-	if err != nil {
-		return nil, errors.Wrap(err, "error marshalling updated download request")
-	}
-
-	patchBytes, err := jsonpatch.CreateMergePatch(origBytes, updatedBytes)
-	if err != nil {
-		return nil, errors.Wrap(err, "error creating json merge patch for download request")
+		if downloadRequest.Status.DownloadURL, err = backupStore.GetDownloadURL(downloadRequest.Spec.Target); err != nil {
+			return ctrl.Result{Requeue: true}, err
```
dsu-igeek, line 129: Should we use errors.WithStack() here like above?

dsu-igeek (approved):
I just had the places where I think we want to use WithStack or Wrap on the returned error, otherwise looks good.

pkg/controller/download_request_controller_test.go
```diff
@@ -18,310 +18,256 @@ package controller
...
-			Name:      "a-download-request",
-			Namespace: velerov1api.DefaultNamespace,
-		},
-		Spec: velerov1api.DownloadRequestSpec{
-			Target: velerov1api.DownloadTarget{
-				Kind: targetKind,
-				Name: targetName,
+var _ = Describe("Download Request Reconciler", func() {
+	BeforeEach(func() {})
+	AfterEach(func() {})
```
zubron, line 48: Can these be removed if we're not doing anything in setup/teardown?

pkg/controller/download_request_controller_test.go
```diff
@@ -18,310 +18,256 @@ package controller
...
+
+	It("Should successfully patch a download request object status phase", func() {
+		defaultBackup := func() *velerov1api.Backup {
+			return builder.ForBackup(velerov1api.DefaultNamespace, "a-backup").StorageLocation("a-location").Result()
+		}
+
+		tests := []struct {
```
zubron, line 55: I think if we're going to use Ginkgo here, we should be using the `DescribeTable` function for these tests as you did in #3180. I think without that, all these tests are treated as one spec and you don't get to see the name of a test if it fails.

pkg/controller/download_request_controller_test.go
```diff
@@ -18,310 +18,256 @@ package controller
...
+			rClock := clock.NewFakeClock(now)
 
-			// Set .status.expiration properly for processed requests. Since "expired" is relative to the controller's
-			// clock time, it's easier to do this here than as part of the test case definitions.
-			if tc.downloadRequest != nil && tc.downloadRequest.Status.Phase == velerov1api.DownloadRequestPhaseProcessed {
-				if tc.expired {
-					tc.downloadRequest.Status.Expiration = &metav1.Time{Time: harness.controller.clock.Now().Add(-1 * time.Minute)}
-				} else {
-					tc.downloadRequest.Status.Expiration = &metav1.Time{Time: harness.controller.clock.Now().Add(time.Minute)}
+			if tc.downloadRequest != nil {
```
zubron, line 193: In all the test cases, `tc.downloadRequest` is not nil and it seems like it's necessary that it isn't nil. If you wanted to be really strict, you could fail the test early if it's nil?

zubron (changes requested):
I know you've shown me some of the work to use kubebuilder/controller runtime before, but seeing the changes in a PR make it really clear how much nicer this approach is :smile: 

Most of the comments I've left are cleanup/suggestions and could be addressed in a follow up. The only thing I'd really like to check before merging is the change to the requeue in the BSL controller. I don't know if that's an intentional change, and if so, what is the impact.

pkg/controller/download_request_controller.go
```diff
@@ -18,229 +18,137 @@ package controller
...
+			return ctrl.Result{}, nil
+		}
 
-	log.Debug("Running processDownloadRequest")
-	ns, name, err := cache.SplitMetaNamespaceKey(key)
-	if err != nil {
-		log.WithError(err).Error("error splitting queue key")
-		return nil
+		log.WithError(err).Error("Error getting DownloadRequest")
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
carlisia: The reason for returning some errors with stack and others not is documented here: https://velero.io/docs/v1.5/code-standards/#code.

I think there are some in this code that should be returning with Stack but are not. I'll revise, thank you.

pkg/controller/backup_storage_location_controller.go
```diff
@@ -60,7 +60,7 @@ func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 	locationList, err := storage.ListBackupStorageLocations(r.Ctx, r.Client, req.Namespace)
 	if err != nil {
 		log.WithError(err).Error("No backup storage locations found, at least one is required")
-		return ctrl.Result{Requeue: true}, err
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
carlisia: It was a fix because whenever an error is returned, the request gets requeued, so it was redundant.

pkg/controller/download_request_controller_test.go
carlisia: Yes, good catch, thank you! I'll remove it from #3180 too as well as for other merged controllers.

pkg/controller/download_request_controller_test.go
carlisia: Done.

pkg/controller/backup_storage_location_controller.go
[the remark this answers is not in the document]
zubron: Thanks for explanation!

pkg/controller/download_request_controller_test.go
carlisia: Yes, good point. Fixed.

carlisia:
Hey @vmware-tanzu/velero-maintainers, thanks for the reviews and sorry for the long turn around.

This is ready for review again.

pkg/cmd/util/downloadrequest/downloadrequest.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 the Velero contributors.
+Copyright 2020 the Velero contributors.
```
nrb, line 2: 2021 now.

pkg/controller/download_request_controller.go
```diff
@@ -18,229 +18,136 @@ package controller
...
-
-	updatedBytes, err := json.Marshal(updated)
-	if err != nil {
-		return nil, errors.Wrap(err, "error marshalling updated download request")
-	}
-
-	patchBytes, err := jsonpatch.CreateMergePatch(origBytes, updatedBytes)
-	if err != nil {
-		return nil, errors.Wrap(err, "error creating json merge patch for download request")
+		if downloadRequest.Status.DownloadURL, err = backupStore.GetDownloadURL(downloadRequest.Spec.Target); err != nil {
```
nrb, line 127: If we never find a DownloadURL here, will it just constantly re-queue?

Maybe we should check for expiration at the beginning, regardless of whether the downloadrequest.Status.Phase field is processed, so that erroneous requests don't sit and spin in the controller forever, filling up the logs with errors. Since DownloadRequests are usually a synchronous activity where someone tries to download a thing at the command line, I'm not sure if it's worth continuing to try past the TTL. Does that sound reasonable to others?

nrb (changes requested):
Unfortunately copyright year changed :(. We do have safe legal ground to remove the copyright year, though, so if you'd rather do that for your changed files, 👍  though we should probably scan through and remove them all at once.

Also a question about requeueing failed requests. I asked it specifically on the DownloadURL, but I think it applies to all of the types.

pkg/controller/download_request_controller.go
zubron: Right now it looks like the expiration is only set once a DownloadURL has been received (which happens in the case of success, we have a DownloadURL which expires after TTL). Are you suggesting that we should instead set the expiration at the point where we **create** the DownloadRequest, and then the expiration of the request would be tied to time since creation, rather than the expiration time of the resulting URL?

carlisia: Addressed this as per Nolan's suggestion.

carlisia:
I am manually testing this and getting the error below, will add a comment when it's working and ready for review.

```
An error occurred: Get "http://minio.velero.svc:9000/velero/backups/test1/test1.tar.gz?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=minio%2F20210208%2Fminio%2Fs3%2Faws4_request&X-Amz-Date=20210208T145945Z&X-Amz-Expires=600&X-Amz-SignedHeaders=host&X-Amz-Signature=1b5aafa73c614741cb20f3cc098e5257e1c19e6111599c53ff8f9fb21ffba7a4": dial tcp 10.195.47.241:9000: connect: connection refused
```

zubron:
@carlisia I've seen that error before and it's caused by MinIO not being accessible outside the cluster. I managed to fix it by doing a `kubectl port-forward <minio-pod> 9000` and adding `publicUrl: http://localhost:9000` to the BSL config. I found that in the AWS plugin docs: https://github.com/vmware-tanzu/velero-plugin-for-aws/blob/main/backupstoragelocation.md

I think we need to update our docs for Tilt to include this info.

carlisia:
Oh wow, thank you!

carlisia:
@zubron please paste your answer here: https://github.com/vmware-tanzu/velero/discussions/3420

I can update the docs later.

carlisia:
@vmware-tanzu/velero-owners this release blocker is ready for review. 

With Bridget's suggestion I was able to complete a manual test. 🎉

nrb:
@carlisia Awesome! Looks like it'll need a rebase

carlisia:
@vmware-tanzu/velero-owners this release blocker is ready for review.

pkg/apis/velero/v1/download_request_types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 the Velero contributors.
+Copyright 2020 the Velero contributors.
```
jenting, line 2: nit: Remove the year

pkg/apis/velero/v1/download_request_types.go
```diff
@@ -77,8 +77,14 @@ type DownloadRequestStatus struct {
 	Expiration *metav1.Time `json:"expiration,omitempty"`
 }
 
+// TODO(2.0) After converting all resources to use the runttime-controller client,
```
jenting, line 80: typo: runtime

pkg/apis/velero/v1/download_request_types.go
```diff
@@ -99,7 +105,12 @@ type DownloadRequest struct {
 	Status DownloadRequestStatus `json:"status,omitempty"`
 }
 
+// TODO(2.0) After converting all resources to use the runttime-controller client,
```
jenting, line 108: typo: runtime

pkg/builder/download_request_builder.go
```diff
@@ -0,0 +1,62 @@
...
+	return b.object
+}
+
+// Phase sets the DownloadRequest's status phase.
+func (b *DownloadRequestBuilder) Phase(phase velerov1api.DownloadRequestPhase) *DownloadRequestBuilder {
+	b.object.Status.Phase = phase
+	return b
+}
+
+// Phase sets the DownloadRequest's status phase.
```
jenting, line 57: This should be the Target function description.

nrb:
Looks like we have conflicts after the colorized output PR merged.

carlisia:
Ready for review.

nrb (changes requested):
I think this still needs a little work.

My original intention was to add a new case where DownloadRequests were not reaching a Processed phase, but reached the expiration time. These requests would just constantly requeue in the old code, because there was no terminal state for DownloadRequests never successfully got a DownloadURL. Eventually, these could pile up.

We should also reach a terminal state when the DownloadRequest is moved to the Processed phase, which is what we do now.

So in total, there should be 2 ways for a single DownloadRequest to be deleted in the reconciliation loop:

1) "Now" is after the expiration time; we've never reached the Processed state, but also don't want to queue infinitely
2) We've reach the Processed state, and can clean it up.

pkg/controller/download_request_controller_test.go
```diff
@@ -18,306 +18,237 @@ package controller
...
+		Entry("request with phase 'Processed' is not deleted if not expired", request{
+			downloadRequest: builder.ForDownloadRequest(velerov1api.DefaultNamespace, "a-download-request").Phase(velerov1api.DownloadRequestPhaseProcessed).Target(velerov1api.DownloadTargetKindBackupLog, "a-backup-20170912150214").Result(),
 			backup:          defaultBackup(),
-		},
-		{
-			name:            "request with phase 'Processed' is deleted if expired",
-			downloadRequest: newDownloadRequest(velerov1api.DownloadRequestPhaseProcessed, velerov1api.DownloadTargetKindBackupLog, "a-backup-20170912150214"),
+			expectedRequeue: ctrl.Result{Requeue: true},
+		}),
+		Entry("request with phase 'Processed' is deleted if expired", request{
```
zubron, line 247: Further to @nrb's comments, we should add some more test cases here to show that if a request has phase 'New', or is empty, that we should still delete it if it has expired. I was going through some of the test logic and it looks like it only applies expiration logic in the case where the status is Processed.

zubron:
> 2. We've reach the Processed state, and can clean it up.

Just to clarify - we currently delete the request if it's Processed *and* it's expired. I'm not totally clear on the logic of the controllers and how the download request streaming works but if the request is moved to Processed, it will be requeued and would then be deleted on the next reconciliation loop. It might still be in use by the logs streaming and shouldn't be deleted until after its expiration.

nrb:
> Just to clarify - we currently delete the request if it's Processed and it's expired. 

Thanks, you're right. I was a little hand-wavy there as I was focusing on the current loop iteration. But you're right - it'll requeue until it's both Processed _and_ expired.

carlisia:
Addressed the corrections (thank you1).

Note: I don't know what happened but needed to do a rebase, and the actual commit was "eaten", but the changes were pushed.

Ready for review.

nrb (approved):

pkg/controller/download_request_controller_test.go
```diff
@@ -18,306 +18,249 @@ package controller
...
+			downloadRequest: builder.ForDownloadRequest(velerov1api.DefaultNamespace, "a-download-request").Phase(velerov1api.DownloadRequestPhaseProcessed).Target(velerov1api.DownloadTargetKindBackupLog, "a-backup-20170912150214").Result(),
+			backup:          defaultBackup(),
+			expired:         true,
+			expectedRequeue: ctrl.Result{Requeue: false},
+		}),
+		Entry("request with phase '' and expired is deleted", request{
+			downloadRequest:      builder.ForDownloadRequest(velerov1api.DefaultNamespace, "a-download-request").Phase("").Target(velerov1api.DownloadTargetKindBackupLog, "a-backup-20170912150214").Result(),
+			backup:               defaultBackup(),
+			expired:              true,
+			expectedReconcileErr: "backups.velero.io \"a-backup-20170912150214\" not found",
```
zubron, line 262: Like the previous test where the expired `DownloadRequest` was deleted, we shouldn't expect an error here. There are a few points in the main test code where we only set the expiration on the `DownloadRequest` when it is `Processed` when it should be applied to any expired request.

pkg/controller/download_request_controller_test.go
```diff
@@ -18,306 +18,249 @@ package controller
...
+			pluginManager.On("CleanupClients").Return(nil)
+
+			Expect(test.downloadRequest).ToNot(BeNil())
 
 			// Set .status.expiration properly for processed requests. Since "expired" is relative to the controller's
 			// clock time, it's easier to do this here than as part of the test case definitions.
-			if tc.downloadRequest != nil && tc.downloadRequest.Status.Phase == velerov1api.DownloadRequestPhaseProcessed {
-				if tc.expired {
-					tc.downloadRequest.Status.Expiration = &metav1.Time{Time: harness.controller.clock.Now().Add(-1 * time.Minute)}
+			if test.downloadRequest != nil && test.downloadRequest.Status != (velerov1api.DownloadRequestStatus{}) && test.downloadRequest.Status.Phase == velerov1api.DownloadRequestPhaseProcessed {
```
zubron, line 85: I think this whole outer check can be removed as we need to set the expiration for all requests.

pkg/controller/download_request_controller_test.go
```diff
@@ -18,306 +18,249 @@ package controller
...
 
-			if tc.downloadRequest != nil && tc.downloadRequest.Status.Phase == velerov1api.DownloadRequestPhaseProcessed {
-				res, err := harness.client.VeleroV1().DownloadRequests(tc.downloadRequest.Namespace).Get(context.TODO(), tc.downloadRequest.Name, metav1.GetOptions{})
+			instance := &velerov1api.DownloadRequest{}
+			err = r.Client.Get(ctx, kbclient.ObjectKey{Name: test.downloadRequest.Name, Namespace: test.downloadRequest.Namespace}, instance)
 
-				if tc.expired {
-					assert.True(t, apierrors.IsNotFound(err))
+			if test.expired {
+				if test.downloadRequest.Status.Phase == velerov1api.DownloadRequestPhaseProcessed {
```
zubron, line 145: If `test.expired` is true, then we should always expect the err to be a `NotFound` error.

zubron (changes requested):
Thanks, @carlisia! There are just a couple of changes that need to be made in one of the test cases but it's good otherwise :+1:

pkg/controller/download_request_controller_test.go
carlisia: The expiration gets set only in the same block of code where we update the request status to `processed` (https://github.com/vmware-tanzu/velero/blob/1c536aced492862ed61842315921ab6ccfe62887/pkg/controller/download_request_controller.go#L113). 

This means, not all requests have an expiration. A request that has the phase `""` or `New` will not have its expiration set until it gets processed and returned successfully from the reconcile loop.

carlisia: Which is to say, only requests that have been successfully processed will have it's expiration set.

pkg/controller/download_request_controller_test.go
carlisia: Ah. I was thinking this test block was a mess. Thanks to your other comment, I realize a bug in the test: a request cannot be both expired and have a phase other than `processed`, and the last test case if braking this rule. Fixing this.

pkg/controller/download_request_controller_test.go
carlisia: See my other two separate comments, they address both points here.

pkg/controller/download_request_controller_test.go
carlisia: Hm. You are right, of course. Nvm what I said. Let me re-evaluate these tests.

zubron (approved):
Thanks for fixing the tests - LGTM! :)

nrb (approved):

Issue #3008, opened 2020-10-14
Restoration fails in clusters with injected InitContainers because of order.


doughepi:
**What steps did you take and what happened:**
No different from a normal backup and restore process using Restic with the `--default-volumes-to-restic` flag turned on.

1. Began backup using `velero backup create backup-test-3 --storage-location gcp --default-volumes-to-restic --include-namespaces database-portal-env-dev -n velero-system`.
2. Waited for backup to complete and verified that volumes were being backed up correctly.
    a. Would like a way to exclude volumes globally, because there's no reason to back up all the `istio-envoy` volumes on every pod. This is unrelated to this issue, of course.
3. Backup compled and I deleted the namespace using `kubectl delete namespace database-portal-env-dev`.
4. Attempted restore using `velero create restore restore-test-6 --from-backup backup-test-1 --include-namespaces database-portal-env-dev -n velero-system`
5. Waited for the restore complete, but it seemed to be stuck "InProgress".
6. Inspected the `PodVolumeRestore` resources, and all seemed to be without status.
7. Inspected the `restic-wait` init container logs and found many information messages about how the restore flag file was not found.

**What did you expect to happen:**
While I expect that some `Not found: /restores/istio-envoy/.velero/18ff2bf4-eef8-4f46-b821-9da2a0d623da` messages would appear during the restore process, the restore times out because of some unknown reason. No volumes are restored and the restored pods never start because they're all held by the `restic-wait` container.

```
Init Containers:
  istio-validation:
    Container ID:  docker://36eb8ab83d4ea0ce1b5e1a238ecf07bdb26c9d8602868a46788f66c48086ccf0
    Image:         docker.io/istio/proxyv2:1.7.2
    Image ID:      docker-pullable://istio/proxyv2@sha256:26b91ecc8be9cd171acbee0f8b99e99f5b1a0fb3290fe54ad3705830c10a8ff5
    Port:          <none>
    Host Port:     <none>
    Args:
      istio-iptables
      -p
      15001
      -z
      15006
      -u
      1337
      -m
      REDIRECT
      -i
      *
      -x
      
      -b
      *
      -d
      15090,15021,15020
      --run-validation
      --skip-rule-apply
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Tue, 13 Oct 2020 22:17:47 -0500
      Finished:     Tue, 13 Oct 2020 22:17:48 -0500
    Ready:          True
    Restart Count:  0
    Limits:
[...]

PR #3011, opened 2020-10-16
Adding fix for restic init container index on restores.

doughepi:
Allows the `restic-wait` init container to exist in any order in the pod being restored. Prints a warning message in the case where the `restic-wait` container isn't the first container in the list of initialization containers.

As it stands right now, there's no indication that your restore is held up by the `restic-wait` container being anywhere other than the first initialization container. This attempts to complete the restore, and warns if the `restic-wait` container is anywhere other than the first container.

I believe this is the better strategy because there are many cases where—[despite Velero's best effort](https://github.com/vmware-tanzu/velero/blob/7914138dd71296a05b27855426ad4255f8df2cb8/pkg/restore/restic_restore_action.go#L158)—other containers will be injected before it. Our breaking example was the `istio-validate` container. Velero is correctly placing the initialization container at index 0, but the [istio mutating webhook](https://github.com/istio/istio/blob/3036d1e6c5c148a4eeb24eff0700887e6e1e2a63/pkg/kube/inject/webhook.go#L351) will always be evaluated after this. I believe all mutating webhooks that try to place their initialization containers at position 0 will break restores in the same manner.

The warning serves to warn the user. I don't think initialization containers coming before the `restic-wait` container are an error state—just that they could affect the restore if they do something weird with the mounted volumes. Either way, I believe this behavior is better than a silent failure.

fixes: #3008 

Signed-off-by: Piper Dougherty doughertypiper@gmail.com


nrb (approved):
Thanks for the bug report and the fix! This makes a lot of sense to me, and I definitely agree with having a warning instead of silently failing.

The ramifications of not being first in all cases is a little hard to predict as we don't know what might mutate the pod, but I think this is more correct than the previous behavior.

ashish-amarnath (approved):
LGTM.
Thanks for the PR @doughepi 🎉

pkg/controller/pod_volume_restore_controller_test.go
```diff
@@ -598,3 +598,105 @@ func TestIsResticContainerRunning(t *testing.T) {
...
+						},
+						{
+							Name: "non-restic-init",
+						},
+					},
+				},
+			},
+			expected: 0,
+		},
+	}
```
jenting, line 695: It's a duplicated test case.

PR #3016, opened 2020-10-19
Design doc for Velero Restore progress reporting

pranavgaikwad:
This is an initial design doc explaining how progress reporting can be implemented for Velero Restores.

### DIFF
--- a/design/restore-progress.md
+++ b/design/restore-progress.md
@@ -1,0 +1,220 @@
+# Restore progress reporting
+
+Velero _Backup_ resource provides real-time progress of an ongoing backup by means of a _Progress_ field in the CR. Velero _Restore_, on the other hand, only shows one of the phases (InProgress, Completed, PartiallyFailed, Failed) of the ongoing restore. In this document, we propose detailed progress reporting for Velero _Restore_. With the introduction of the proposed _Progress_ field, Velero _Restore_ CR will look like:
+
+```yml
+apiVersion: velero.io/v1
+kind: Restore
+metadata:
+  name: test-restore
+  namespace: velero
+spec:
+    [...]
+status:
+  phase: InProgress
+  progress:
+    itemsRestored: 100
+    totalItems: 140
+```
+
+## Goals
+
+- Enable progress reporting for Velero Restore
+
+## Non Goals
+
+- Estimate time to completion
+
+## Background
+
+The current _Restore_ CR lets users know whether a restore is in-progress or completed (failed/succeeded). While this basic piece of information is useful to the end user, there seems to be room for improvement in the user experience. The _Restore_ CR can show detailed progress in terms of the number of resources restored so far and the total number of resources to be restored. This will be particularly useful for restores that run for a longer duration of time. Such progress reporting already exists for Velero _Backup_. This document proposes similar implementation for Velero _Restore_.
+
+## High-Level Design
+
+We propose to divide the restore process in two steps. The first step will collect all the items to be restored from the backup tarball. It will apply the label selector and include/exclude rules on the resources / items and store them (preserving the priority order) in an in-memory data structure. The second step will read the collected items and restore them. 
+
+## Detailed Design
+
+### Progress struct
+
+A new struct will be introduced to store progress information:
+
+```go
+type RestoreProgress struct {
+    TotalItems    int `json:"totalItems,omitempty`
+    ItemsRestored int `json:"itemsRestored,omitempty`
+}
+```
+
+`RestoreStatus` will include the above struct:
+
+```go
+type RestoreStatus struct {
+    [...]
+
+    Progress *RestoreProgress `json:"progress,omitempty"`
+}
+```
+
+### Modifications to restore.go
+
+Currently, the restore process works by looping through the resources in the backup tarball and restoring them one-by-one in the same pass:
+
+```go
+func (ctx *context) execute(...) {
+    [...]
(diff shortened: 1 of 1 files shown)


design/restore-progress.md
```diff
@@ -0,0 +1,187 @@
...
+
+- Estimate time to completion
+
+## Background
+
+The current _Restore_ CR lets users know whether a restore is in-progress or completed (failed/succeeded). While this basic piece of information is useful to the end user, there seems to be room for improvement in the user experience. The _Restore_ CR can show detailed progress in terms of the number of resources restored so far and the total number of resources to be restored. This will be particularly useful for restores that run for a longer duration of time. Such progress reporting already exists for Velero _Backup_. This document proposes similar implementation for Velero _Restore_.
+
+## High-Level Design
+
+We propose to divide the restore process in two steps. The first step will collect all the items to be restored from the backup tarball. It will apply the label selector and include/exclude rules on the resources / items and store them (preserving the priority order) in an in-memory data structure. The second step will read the collected items and restore them.
```
dymurray, line 34: Nice, glad we think we can achieve the 2-step approach

jenting (approved):
LGTM

design/restore-progress.md
```diff
@@ -0,0 +1,220 @@
...
+
+                // restore item here
+                w, e := restoreItem(...)
+            }
+        }
+    }
+}
+```
+
+We propose to remove the call to `restoreItem()` in the inner most loop and instead store the item in a data structure. Once all the items are collected, we loop through the array of collected items and make a call to `restoreItem()`:
```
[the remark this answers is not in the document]
pranavgaikwad: There's really a tradeoff between computation and memory here: 

1. First step reads from tarball, stores the resources in-memory, second step uses the resources directly from memory 
vs
2. First step reads from tarball, only stores the path to the resource in-memory, second step reads from the tarball again

design/restore-progress.md
```diff
@@ -0,0 +1,220 @@
...
+
+```go
+func (ctx *context) getOrderedResourceCollection(...) {
+    collectedResources := []restoreResource
+    for _, resource := range getOrderedResources(...) {
+        [...]
+
+        for namespace, items := range resourceList.ItemsByNamespace {
+            [...]
```
zubron, line 93: During [this section of the loop](https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L441-L463) in the current implementation, namespaces are handled differently than other resources. We don't explicitly restore them unless a resource is going to be restored into that namespace. They are created if needed, then the resources are restored into that namespace. Are you intending to capture namespaces in the count of items to restore? Will they be restored during the count of items, or will they be left out of the count and restored as they currently are during the main restore loop?

design/restore-progress.md
```diff
@@ -0,0 +1,220 @@
...
+}
+```
+
+We propose to remove the call to `restoreItem()` in the inner most loop and instead store the item in a data structure. Once all the items are collected, we loop through the array of collected items and make a call to `restoreItem()`:
+
+```go
+func (ctx *context) getOrderedResourceCollection(...) {
+    collectedResources := []restoreResource
+    for _, resource := range getOrderedResources(...) {
+        [...]
```
zubron, line 90: [De-duplication of resources](https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L388-L393) happens in this section of the loop. This will also need to be included in this function.

design/restore-progress.md
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    [...]
+}
+```
+
+We introduce two new structs to hold the collected items:
+
+```go
+type restoreResource struct {
```
zubron, line 135: Small thing, but both `restoreResource` and `restoreItem` are used as function names in `restore.go` so they will need to be updated or different names should be used here.

design/restore-progress.md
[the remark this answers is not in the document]
nrb: I'm included to lean towards option 2 for now - loading all the JSON into memory right now isn't really feasible given our default memory constraints, especially on very large clusters.

design/restore-progress.md
```diff
@@ -0,0 +1,220 @@
...
+### Updating progress 
+
+The updates to the `progress` field in the CR can be sent on a channel as soon as an item is restored. A goroutine receiving update on that channel can make an `Update()` call to update the _Restore_ CR. This will require us to pass an instance of `RestoresGetter` to the `kubernetesRestorer` struct.
+
+
+## Alternatives Considered
+
+As an alternative, we have considered an approach which doesn't divide the restore process in two steps. 
+
+With that approach, the total number of items will be read from the Backup CR. We will keep three counters, `totalItems`, `skippedItems` and `restoredItems`:
```
zubron, line 193: I don't know if this approach would work as the restore skips over resources/namespaces without iterating through to find out how many items are being skipped. We could know the number of items available in the backup to be restored, and how many have been restored so far, but without iterating, I don't know if there's a way to determine how many are being skipped.

nrb (approved):
This looks pretty reasonable to me for a first approach! I think longer term, we'll want to flesh out a more complete directed graph structured, but I think for progress reporting, this is a perfectly workable solution at the moment, and I'd be willing to proceed with it.

carlisia (approved):
lgtm. @zubron made a couple observations worth noting.

PR #3018, opened 2020-10-20
Specify timeout for plugins of specific Kind.

phuongatemc:
Signed-off-by: Hoang, Phuong <phuong.hoang@emc.com>


phuongatemc:
This design proposes a solution that allows user to specify timeout during backup for all plugin executions on an object of specific resource types.

carlisia:
@phuongatemc did you mean to delete design/backup-resources-order.md?

phuongatemc:
> @phuongatemc did you mean to delete design/backup-resources-order.md?

Sorry but am I supposed to remove the design doc after it has been released?

carlisia:
That's a great question! We are gonna keep it around. I actually am due to create some folder structure to house wip / completed.

design/plugins-timeout.md
```diff
@@ -0,0 +1,34 @@
...
+## Alternatives Considered
+- Set timeout for the entire Velero Backup or Restore to avoid the hanging of the backup or restore workflow.  This only helps in some scenario but it will not help the AppConsistent pod being quiesced for too long.
+- Set timeout for each plugin being executed on an object.  In cases that multiple plugins being executed on an object, the total time of these plugin executions may be still greater than the limit even though individual timeout of each plugin would be smaller than timeout.
+
+
+## High-Level Design
+- Enhance the BackupSpec to contain the map of resource type (Kind) to the timeout of executing all plugins being executed on an object of that resource type.
+   type BackupSpec struct {
+      ...
+      TotalPluginsTimeouts map[string]int `json:"totalPluginsTimeouts,omitempty"` //map of resource type to total plugins timeout (in milliseconds)
```
nrb, line 22: Can you give an example of what you envision the keys within this map to look like? Is it matching on a Kubernetes `Kind`?

If so, is that really only meaningful on BackupItemAction plugins, and not on any of the other plugin types, like ObjectStorage or VolumeSnapshotter?

nrb (changes requested):
@phuongatemc I think we definitely need to have an overall timeout on backup operations and a way to pass them down to plugins, but I don't know about having individual timeouts per Kubernetes resource type.

It seems your use case is to run this against PVCs specifically?  Is there anything else you'd want to run this against?

design/plugins-timeout.md
phuongatemc: The key in this map is the Kubernetes Kind.  
In theory, I think I can extend this design for all plugins, not just BackupItemAction.  Not sure if I can have time to implement for all.

nrb:
@phuongatemc Can this be closed in favor of #3355?

PR #3024, opened 2020-10-20
Add warning to velero version cmd.

### LINKED ISSUE #3017: Add a warning to the `velero version` cmd when versions don't match
I've been seeing a few issues opened with a mismatch in version between Velero client and server. I think there's sufficient coupling between the two that it (the mismatch) could originate the problems being reported.

The `velero version` command should verify any mismatch and warn. Not sure how strong this warning should be. Should it be: "will not work, please update [whichever has the lower version]"?

cvhariharan:
Fixes #3017 

This PR adds a warning to`velero version` command. The warning is shown if the major and minor versions of the client and server mismatch.

If the client version is less than the server version
```
# WARNING: the client version does not match the server version. Please update the client
```
If the server version is less than the client version
```
# WARNING: the client version does not match the server version. Please update the server
```

### DIFF
--- a/pkg/cmd/cli/version/version.go
+++ b/pkg/cmd/cli/version/version.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2017, 2019 the Velero contributors.
+Copyright 2020 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/mod/semver"
 	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/vmware-tanzu/velero/pkg/buildinfo"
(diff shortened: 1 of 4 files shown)


pkg/cmd/cli/version/version.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2017, 2019 the Velero contributors.
+Copyright 2017, 2019, 2020 the Velero contributors.
```
nrb, line 2: This should actually be just 2020, if you wouldn't mind updating it.

I know many of the files are inconsistent with this, we're slowly updating them.

nrb (changes requested):
Thanks for this! I think it looks pretty good, I just have one small thing that should change.

pkg/cmd/cli/version/version.go
```diff
@@ -79,6 +80,18 @@ func printVersion(w io.Writer, clientOnly bool, kbClient kbclient.Client, server
 		return
 	}
 
+	serverSemVer := semver.MajorMinor(serverStatus.Status.ServerVersion)
+	cliSemVer := semver.MajorMinor(buildinfo.Version)
+	if serverSemVer != cliSemVer {
```
zubron, line 85: I would move this check to the end of the function, after the server version has been printed. If there is a client/server mismatch, we would return from this function without telling the user what  the server version actually was and they wouldn't know what version they need to upgrade to.

zubron (changes requested):

pkg/cmd/cli/version/version.go
cvhariharan: Yeah, that makes sense.

nrb (approved):

ashish-amarnath (approved):

ashish-amarnath:
Merging as the requested changes have been made.
Thank you @cvhariharan for this PR!

ashish-amarnath:
GH isn't allowing merge w/o dismissing @zubron's review. Going to let @zubron do the merge

zubron (approved):
Oh, sorry for holding this up! Thanks for making the suggested change and for your contribution @cvhariharan! :smile:

Issue #3045, opened 2020-10-27
Add velero log msg="Backup finished" with all necessary information 


konvergence:
**Describe the problem/challenge you have**
Actually the velero metrics are based on cumultive counter.  
* It is not easy to  create grafana dashboards on this metrics.
* it is really hard to display which backup was failed.

**Describe the solution you'd like**
* Actually  velero log when backup is started but don't log when when a backup is finished with his status

* Actually we have a log with msg="Backup starting"
  * Example of existing logs msg="Backup starting"
```
time="2020-10-27T02:00:23Z" level=info msg="Backup starting" backup=velero-minio/k8s-admin-daily-shuttle-demo-test-20201027020020 controller=pod-volume-backup logSource="pkg/controller/pod_volume_backup_controller.go:191" name=k8s-admin-daily-shuttle-demo-test-20201027020020-m77xk namespace=velero-minio
```

* Could you add a log with msg = "Backup finished" with all the necessary information on status and spec into json format (for example)

   * `status="{\"completionTimestamp\":\"2020-10-27T02:00:31Z\",\"expiration\":\"2020-11-11T02:00:20Z\",\"formatVersion\":\"1.1.0\",\"phase\":\"Completed\",\"progress\":{\"itemsBackedUp\":51,\"totalItems\":51}}"`
  * `spec="{\"defaultVolumesToRestic\":true,\"hooks\":{},\"includedNamespaces\":[\"shuttle-demo-test\"],\"storageLocation\":\"default\",\"ttl\":\"360h0m0s\",\"volumeSnapshotLocations\":[\"default\"]}"`

  * Example of new logs msg="Backup finished" with status and spec fields
```
[...]

Issue #2942, opened 2020-09-17
No backup locations were ready to be verified - reword BSL controller status message


eyalzek:
**What steps did you take and what happened:**
Upgrade velero from 1.4.2 to 1.5.1, controller starts printing in a loop that the backup storage location cannot be found. This has been working since pre-velero days.

**What did you expect to happen:**
Velero would continue working as expected

**The output of the following commands will help us better understand what's going on**:
(Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)

- `kubectl logs deployment/velero -n velero`
```
time="2020-09-17T13:16:05Z" level=info msg="Checking for existing backup locations ready to be verified; there needs to be at least 1 backup location available" controller=backupstoragelocation logSource="pkg/controller/backupstoragelocation_controller.go:58"
time="2020-09-17T13:16:06Z" level=info msg="Checking for existing backup locations ready to be verified; there needs to be at least 1 backup location available" controller=backupstoragelocation logSource="pkg/controller/backupstoragelocation_controller.go:58"
time="2020-09-17T13:16:06Z" level=info msg="No backup locations were ready to be verified" controller=backupstoragelocation logSource="pkg/controller/backupstoragelocation_controller.go:120"
time="2020-09-17T13:17:06Z" level=info msg="Checking for existing backup locations ready to be verified; there needs to be at least 1 backup location available" controller=backupstoragelocation logSource="pkg/controller/backupstoragelocation_controller.go:58"
time="2020-09-17T13:17:07Z" level=info msg="Checking for existing backup locations ready to be verified; there needs to be at least 1 backup location available" controller=backupstoragelocation logSource="pkg/controller/backupstoragelocation_controller.go:58"
time="2020-09-17T13:17:07Z" level=info msg="No backup locations were ready to be verified" controller=backupstoragelocation logSource="pkg/controller/backupstoragelocation_controller.go:120"
time="2020-09-17T13:17:39Z" level=info msg="Checking for existing backup locations ready to be verified; there needs to be at least 1 backup location available" controller=backupstoragelocation logSource="pkg/controller/backupstoragelocation_controller.go:58"
time="2020-09-17T13:17:40Z" level=error msg="Error updating backup location phase" backupstoragelocation=default controller=backupstoragelocation error="backupstoragelocations.velero.io \"default\" not found" logSource="pkg/controller/backupstoragelocation_controller.go:115"
```

[...]

PR #3047, opened 2020-10-27
🏃‍♂ Improve log message clarity

ashish-amarnath:

Fixes: #3045

### DIFF
--- a/pkg/controller/backup_storage_location_controller.go
+++ b/pkg/controller/backup_storage_location_controller.go
@@ -55,7 +55,7 @@ type BackupStorageLocationReconciler struct {
 func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.WithField("controller", BackupStorageLocation)
 
-	log.Info("Checking for existing backup locations ready to be verified; there needs to be at least 1 backup location available")
+	log.Debug("Validating availablity of backup storage locations.")
 
 	locationList, err := storage.ListBackupStorageLocations(r.Ctx, r.Client, req.Namespace)
 	if err != nil {
@@ -78,7 +78,7 @@ func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 		}
 
 		if !storage.IsReadyToValidate(location.Spec.ValidationFrequency, location.Status.LastValidationTime, r.DefaultBackupLocationInfo, log) {
-			log.Debug("Backup location not ready to be validated")
+			log.Info("Validation not required, skipping...")
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


ashish-amarnath:
@nrb, @carlisia, @zubron, and @dsu-igeek PTAL.

pkg/controller/backup_storage_location_controller.go
```diff
@@ -55,7 +55,7 @@ type BackupStorageLocationReconciler struct {
 func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.WithField("controller", BackupStorageLocation)
 
-	log.Info("Checking for existing backup locations ready to be verified; there needs to be at least 1 backup location available")
+	log.Info("Verifying availablity of backup storage locations.")
```
nrb, line 58: Given how often this runs, I think we should bump this down to debug. Otherwise, it just files the logs up, which makes parsing tough.

pkg/controller/backup_storage_location_controller.go
```diff
@@ -95,10 +95,10 @@ func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 			continue
 		}
 
-		log.Debug("Verifying backup storage location")
+		log.Debug("Validating backup storage location")
```
nrb, line 98: I would bump this up to Info since it happens less frequently than the verifying message, and can be useful to know that it's happening.

nrb (changes requested):

nrb (approved):

pkg/controller/backup_storage_location_controller.go
```diff
@@ -55,7 +55,7 @@ type BackupStorageLocationReconciler struct {
 func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.WithField("controller", BackupStorageLocation)
 
-	log.Info("Checking for existing backup locations ready to be verified; there needs to be at least 1 backup location available")
+	log.Debug("Verifying availablity of backup storage locations.")
```
carlisia, line 58: These new messages ARE better! One suggestion I have is to use "validation" everywhere. For example, in this case this message would be "Validating availability..."/

pkg/controller/backup_storage_location_controller.go
```diff
@@ -117,7 +117,7 @@ func (r *BackupStorageLocationReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 	}
 
 	if !anyVerified {
-		log.Info("No backup locations were ready to be verified")
+		log.Info("No backup locations needed to be verified")
```
carlisia, line 120:
```suggestion
		log.Info("No backup locations needed to be validated")
```

carlisia (commented):
I am suggesting to use "validate/validating" every where in the controller and not interchange it with "verifying".

zubron (approved):
These messages are much clearer - thanks!

zubron:
This also fixes #2942.

carlisia (approved):
👍

Issue #2551, opened 2020-05-20
Restore API Groups based on Target Cluster Preferred/Supported Version


brito-rafa:
**Describe the problem/challenge you have**
This is a follow-up of PR #2373 (back up all API Group Versions). We now need to introduce the ability to restore the Target Cluster APIGroup Preferred Version (TCPV). Today's default restores the source cluster APIGroup Preferred Version (SCPV).
This is ultimately to address #2251.

**Describe the solution you'd like**
I added a diagram here:
https://docs.google.com/presentation/d/1hpq3kwc4uYR96z1Jzi6G1TZMqq9YcBoIw-JcHqkSnnE/edit#slide=id.p1

Lingo:
Target Cluster APIGroup Preferred Version (TCPV) - from discovery client.
Source Cluster APIGroup Preferred Version (SCPV) - looking up the "preferredversion" directory inside the backup tarball.
Target Cluster APIGroup Supported Versions (TCSV) - array from discovery client.
Source Cluster APIGroup Supported Versions (SCSV) - array built looking up the versions inside the backup tarball.

Basically, wrap this new logic under the same feature flag "EnableAPIGroupVersions". Without this flag, restore logic stays the same.
If the feature flag is passed, the target cluster discovery client will contain a list of target cluster preferred versions (TCPV) and an array of all supported versions (TCSV).

If TCPV in found on the backup then
 restore item using the TCPV in path inside the backup tarball
else if SCPV belongs in TCSV array then
 restore item using the SCPV in path inside the backup tarball
else if SCSV and TCSV arrays have any element in common then
  restore item common element version in path inside the backup tarball
else log an error or a warning "not supported API Group between the clusters"

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**
1.4.0+ future

carlisia:
This is the recording for the WG-Data Protection meeting of 04/08/2020 where the motivation and more context for this work was presented:

https://youtu.be/S2o7znJS_9w?t=206

jenting:
I'd like to handle this issue.

ashish-amarnath:
@brito-rafa If you are not working on this anymore, is it ok for @jenting to pick this up?

brito-rafa:
Hey there! 
I was planning to tackle this one around mid-September. In preliminary conversations about this change is how the new Restore logic would behave with RestoreItem plugin.
So I started to write my own plugin before changing the Velero Core.
[...]

PR #3050, opened 2020-10-30
Draft design doc for restoring API group version by priority level

codegold79:
Add design doc draft for restoring API group version by priority level.
Addresses issue #2551.

### DIFF
--- a/design/restore-with-EnableAPIGroupVersions-feature.md
+++ b/design/restore-with-EnableAPIGroupVersions-feature.md
@@ -8,7 +8,7 @@ This document proposes a solution to select an API group version to restore from
 
 ## Background
 
-It is possible that between the time a backup has been made and a restore occurs that the target Kubernetes version has incremented more than one version. In such a case where at least a versions of Kubernetes was skipped, the preferred source cluster's API group versions for resources may no longer be supported by the target cluster. With PR #2373, all supported API group versions were backed up if the EnableAPIGroupVersions feature flag was set for Velero. The next step (outlined by this design proposal) will be to see if any of the backed up versions are supported in the target cluster and if so, choose one to restore for each backed up resource.
+It is possible that between the time a backup has been made and a restore occurs that the target Kubernetes version has incremented more than one version. In such a case where at least a versions of Kubernetes was skipped, the preferred source cluster's API group versions for resources may no longer be supported by the target cluster. With [PR#2373](https://github.com/vmware-tanzu/velero/pull/2373), all supported API group versions were backed up if the EnableAPIGroupVersions feature flag was set for Velero. The next step (outlined by this design proposal) will be to see if any of the backed up versions are supported in the target cluster and if so, choose one to restore for each backed up resource.
 
 ## Goals
 
@@ -18,9 +18,9 @@ It is possible that between the time a backup has been made and a restore occurs
 ## Non Goals
 
 - Allow users to restore onto a cluster that is running a Kubernetes version older than the source cluster. The changes proposed here only allow for skipping ahead to a newer Kubernetes version, but not going backward.
-- Allow restoring from backups created using Velero version 1.3 or older. This proposal will only work on backups created using Velero 1.4+ where the EnableAPIGroupVersions was enabled.
+- Allow restoring from backups created using Velero version 1.3 or older. This proposal will only work on backups created using Velero 1.4+.
 - Modifying the compressed backup tarball files. We don't want to risk corrupting the backups.
-- Using plugins to restore a resource when the target supports non of the source cluster's API group versions. The ability to use plugins will hopefully be something added in the future, but not at this time.
+- Using plugins to restore a resource when the target supports none of the source cluster's API group versions. The ability to use plugins will hopefully be something added in the future, but not at this time.
 
 ## High-Level Design
 
@@ -31,7 +31,7 @@ The proposed code starts with creating three lists for each backed up resource. 
   (2) looking at the target cluster and determining which API group versions are supported, and
   (3) getting config maps from the target cluster in order to get user-defined prioritization of versions.
 
-  The three lists will be used to create a map of chosen versions for each resource to restore. If there is a user-defined list of priority versions, the versions will be checked against the supported versions lists. The highest user-defined priority version that is/was supported by both target and source clusters will be the chosen version for that resource.
+  The three lists will be used to create a map of chosen versions for each resource to restore. If there is a user-defined list of priority versions, the versions will be checked against the supported versions lists. The highest user-defined priority version that is/was supported by both target and source clusters will be the chosen version for that resource. If no user specified version are not supported by both target and source, the versions will be logged and the restore will continue with other prioritizations.
 
   Without a user-defined prioritization of versions, the following version prioritization will be followed, starting from the highest priority: target cluster preferred version, source cluster preferred version, and a common supported version. Should there be multiple common supported versions, the one that will be chosen will be based on the [Kubernetes version priorities](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority).
   
@@ -41,36 +41,46 @@ The proposed code starts with creating three lists for each backed up resource. 
 
 There are six objectives to achieve the above stated goals:
 
-1. Determine if the APIGroupVersionsFeatureFlag feature flag has been enabled.
+1. Determine if the APIGroupVersionsFeatureFlag feature flag is enabled and Backup Objects use Status.FormatVersion 1.1.0.
 1. List the backed up API group versions.
 1. List the API group versions supported by the target cluster.
 1. Get the user-defined version priorities.
 1. Use a priority system to determine which version to restore. The source preferred version will be the default if the priorities fail.
 1. Modify the paths to the backup files in the tarball in the resource restore process.
 
-### Objective 1: Determine if the APIGroupVersionsFeatureFlag feature flag has been enabled
+### Objective 1: Determine if the APIGroupVersionsFeatureFlag feature flag is enabled and Backup Objects use Status.FormatVersion 1.1.0
 
-For restore to be able to choose from multiple supported backed up versions, the feature flag must have been enabled during both the backup and restore processes. The check to see if the feature flag is/was enabled must be done for both processes.
+For restore to be able to choose from multiple supported backed up versions, the feature flag must have been enabled during the restore processes. Backup objects must also have [Status.FormatVersion == "1.1.0"](https://github.com/vmware-tanzu/velero/blob/a1e182e723a8c5f6d4175d8db2361233a94d2502/pkg/backup/backup.go#L58).
 
 The reason for checking for the feature flag during restore is to ensure the user would like to restore a version that might not be the source cluster preferred version.
 
-The reason for checking to see if the feature flag was enabled during backup is to ensure the changes made by this proposed design is backward compatible. Only with Velero version 1.4 and forward was Format Version 1.0.0 used to structure the backup directories. Format Version 1.0.0 is required for the restore process proposed in this design doc to work. Before v1.4, the backed up files were in a directory structure that will not be recognized by the proposed code changes. In this case, restore should not attempt to restore from multiple versions as they will not exist.
+The reason for checking `Status.FormatVersion` is to ensure the changes made by this proposed design is backward compatible. Only with Velero version 1.4 and forward was Format Version 1.1.0 used to structure the backup directories. Format Version 1.1.0 is required for the restore process proposed in this design doc to work. Before v1.4, the backed up files were in a directory structure that will not be recognized by the proposed code changes. In this case, restore should not attempt to restore from multiple versions as they will not exist.
 
 Checking if the feature flag is enabled during restore is straightforward using `features.IsEnabled(velerov1api.APIGroupVersionsFeatureFlag)`.
 
-Checking if the feature flag was enabled during backup can be done indirectly by checking the directory structure and version directory names. For backups generated using Format Version 1.0.0 (required for the changes proposed here), there will always be a version directory with `-preferredversion` in the directory name. Only one resource will need to be checked to determine if the feature flag was enabled during backup.
+Checking the `Status.FormatVersion` can be done indirectly by checking the directory structure and version directory names. For backups generated using Format Version 1.1.0, there will always be a version directory with `-preferredversion` in the directory name. Only one resource will need to be checked to determine if the feature flag was enabled during backup.
+
+To help with the format version check, a new parser object can be created using the construction function `NewParser(...)` and we can take advantage of the `archive` package's `Parser` type. A method for the parser object will be created that has the signature `UsedFormatVersion(dir string) (bool, error)`. The method will loop through the version directories inside one of the resource.group directories and determine if a directory with `-preferredversion` was found.
+
+The above two checks can be done by creating a method on the `*restoreContext` object with the method signature `meetsAPIGVRestoreReqs(backupDir string) (bool, error)`. This method can remain in the `restore` package, but for organizational purposes, it can be moved to a file called `prioritize_group_version.go`.
 
 ### Objective 2: List the backed up API group versions
 
 Currently, in `pkg/restore/restore.go`, in the `execute(...)` method, around [line 363](https://github.com/vmware-tanzu/velero/blob/7a103b9eda878769018386ecae78da4e4f8dde83/pkg/restore/restore.go#L363), the resources and their backed up items are saved in a map called `backupResources`.
 
-At this point, it can be checked if the feature flag is/was enabled during restore/backup, respectively (described previously in Objective #1). If it is and was enabled, the `backedupResources` map can be sent to a method (to be created) with the signature `ctx.chooseAPIVersionsToRestore(backupResources)`. Notice that the method is called on `ctx` which has the type `*restore.Context`.
+At this point, the feature flag and format versions can be checked (described previously in Objective #1). If the minimum requirements are met, the `backedupResources` map can be sent to a method (to be created) with the signature `ctx.chooseAPIVersionsToRestore(backupResources)`. The `ctx` object has the type `*restore.Context`.
 
 The `chooseAPIVersionsToRestore` method can remain in the `restore` package, but for organizational purposes, it can be moved to a file called `prioritize_group_version.go`.
 
-Inside the `chooseAPIVersionsToRestore` method, we can take advantage of the `archive` package's `Parser` type. A new parser object can be created using the construction function `NewParser(...)`. A method for the parser object will be created that has the signature `ParseGroupVersions(backupDir string) (map[string]metav1.APIGroup, error)`. The `ParseGroupVerisons(...)` method will loop through the `resources`, `resource.group`, and group version directories to populate a map called `sourceRGVersions`.
+Inside the `chooseAPIVersionsToRestore` method, we can take advantage of the `archive` package's `Parser` type.  `ParseGroupVersions(backupDir string) (map[string]metav1.APIGroup, error)`. The `ParseGroupVersions(...)` method will loop through the `resources`, `resource.group`, and group version directories to populate a map called `sourceRGVersions`.
 
-The `sourceRGVersions` map's keys will be strings in the format `<resource>.<group>`, e.g. "horizontalpodautoscalers.autoscaling". The values will be APIGroup structs. The API Group struct can be imported from k8s.io/apimachinery/pkg/apis/meta/v1. Order the APIGroup.Versions slices using a kubernetes/apimachinery method called `PrioritizedVersionsForGroup` in the [runtime package](https://github.com/kubernetes/apimachinery/blob/b63a0c883fbfc313249150449400788e5589ef23/pkg/runtime/scheme.go#L614).
+The `sourceRGVersions` map's keys will be strings in the format `<resource>.<group>`, e.g. "horizontalpodautoscalers.autoscaling". The values will be APIGroup structs. The API Group struct can be imported from k8s.io/apimachinery/pkg/apis/meta/v1. Order the APIGroup.Versions slices using a sort function copied from `k8s.io/apimachinery/pkg/version`.
+
+```go
+sort.SliceStable(gvs, func(i, j int) bool {
+    return version.CompareKubeAwareVersionStrings(gvs[i].Version, gvs[j].Version) > 0
+})
+```
 
 ### Objective 3: List the API group versions supported by the target cluster
 
@@ -80,7 +90,13 @@ Still within the `chooseAPIVersionsToRestore` method, the target cluster's resou
 targetRGVersions := ctx.discoveryHelper.APIGroups()
 ```
 
-Order the APIGroup.Versions slices using a kubernetes/apimachinery method called `PrioritizedVersionsForGroup` in the [runtime package](https://github.com/kubernetes/apimachinery/blob/b63a0c883fbfc313249150449400788e5589ef23/pkg/runtime/scheme.go#L614).
+Order the APIGroup.Versions slices using a sort function copied from `k8s.io/apimachinery/pkg/version`.
+
+```go
+sort.SliceStable(gvs, func(i, j int) bool {
+    return version.CompareKubeAwareVersionStrings(gvs[i].Version, gvs[j].Version) > 0
+})
+```
 
 ### Objective 4: Get the user-defined version priorities
 
@@ -93,7 +109,7 @@ metadata:
   name: enableapigroupversions
   namespace: velero
 data:
-  restoreResourcesVersionPriority: |
+  restoreResourcesVersionPriority: | -
     rockbands.music.example.io=v2beta1,v2beta2
     orchestras.music.example.io=v2,v3alpha1
     subscriptions.operators.coreos.com=v2,v1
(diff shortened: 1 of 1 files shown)


design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,135 @@
+# Restore API Group Version by Priority Level When EnableAPIGroupVersions Feature is Set
+
+Status: Draft
+
+## Abstract
+
+This document proposes a solution to select an API group version to restore from the versions backed up using the feature EnableAPIGroupVersions.
```
jenting, line 7:
```suggestion
This document proposes a solution to select an API group version to restore from the versions backed up using the feature flag EnableAPIGroupVersions.
```

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Detailed Design
+
+There are four main objectives to achieve the above stated goal:
+
+1. List the backed up API group versions.
+1. List the API group versions supported by the target cluster.
+1. Use a priority system to determine which version to back up with the source preferred version as the default.
```
jenting, line 33: I was thought that step 3 would be `Use a priority system to determine which version to restore from with the target preferred version as the default.`

Because the priority system will determine which API version to be restored from by:
1. getting source API versions from the backup tarball
2. getting target API versions by discovery API
3. determine which API version to be restored from the backup tarball with order
   - target cluster preferred version if found in the backup tarball
   - source cluster preferred version if source cluster preferred version available in the target cluster
   - the latest common supported version for others

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,135 @@
...
+
+## High-Level Design
+
+During restore, the proposal is that Velero will read the directory names in the backup tarball file and see which API group versions were backed up from the source cluster. Velero will also look at the target cluster and determine which API group versions are supported. Given the two lists of source and target cluster supported API versions, a priority system will be used to determine which version will be backed up. The priority of versions in order are target cluster preferred version, source cluster preferred version, and the latest common supported version. Once the version to restore is chosen, the file path to the backed up resource in the tarball will be modified such that it points to the resources' chosen API group version. If no version is found in common between the source and target clusters, the chosen version will default to the source cluster's preferred version (the version being restored currently without the changes proposal here). Restore will be allowed to continue as before.
+
+## Detailed Design
+
+There are four main objectives to achieve the above stated goal:
+
+1. List the backed up API group versions.
```
jenting, line 31: I believe it will handle the backward-compatible on the source cluster, which means it will support both _without_ and _with_ enabling feature flag `EnableAPIGroupVersions` on the source cluster backed up tarball.

Or, should it only compatible with format version >= 1.1.0?
https://github.com/vmware-tanzu/velero/blob/4178d9d/pkg/apis/velero/v1/backup.go#L233-L235

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: Thanks, I'll get that changed.

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: The default will only take effect if all the three priorities fail. I chose `source preferred` so we can revert back to what the logic was like before the code changes with this PR. When our priority system fails, the failure will look the same as it would have failed before the feature flag was enabled.

Here are the sequence of events in case of failure:

1. Set default to `source cluster preferred version`.
2. Priority one fails. We cannot find a target preferred version that is supported by the source versions.
3. Priority two fails. We cannot find a source preferred version that is supported by the target versions.
4. Priority three fails. We cannot find a version that is common to the target and source.
5. Restore will use default and continue to report error per normal (backup version not supported by target cluster or something like that).

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: > it will support both without and with enabling feature flag

I'm not sure what you mean. Without the feature flag, only the `source preferred version` will be in the backup tarball. No other supported versions will be there so I can only use the source preferred version when the feature flag is not enabled.

As for format version, the new format version applies to the all backup tarballs, with or without the feature flag. But without the feature flag, only to the source preferred version will use format version 1.1.0 because no other version will be in the tarball.

Only if the feature flag is enabled will the other supported versions be available and these versions will meet format version 1.1.0.

Let me know if I misunderstood you.

codegold79: I'm not sure if this will help. Here is what's in the backed tarball files. This first screenshot shows the autoscaler versions that are backed up without feature flag enabled:

Screen Shot 2020-11-03 at 8 09 30 AM

This second screenshot shows the autoscaler versions that are backed up with the feature flag enabled:
Screen Shot 2020-11-03 at 8 09 09 AM

You can see that in both cases, format version 1.1.0 is being used, but the difference is how many versions were backed up.

design/restore-with-EnableAPIGroupVersions-feature.md
jenting: > The default will only take effect if all the three priorities fail. I chose `source preferred` so we can revert back to what the logic was like before the code changes with this PR. When our priority system fails, the failure will look the same as it would have failed before the feature flag was enabled.
> 
> Here are the sequence of events in case of failure:
> 
> 1. Set default to `source cluster preferred version`.
> 2. Priority one fails. We cannot find a target preferred version that is supported by the source versions.
> 3. Priority two fails. We cannot find a source preferred version that is supported by the target versions.
> 4. Priority three fails. We cannot find a version that is common to the target and source.
> 5. Restore will use default and continue to report error per normal (backup version not supported by target cluster or something like that).

Thanks for your clear explanation, a good idea if all the three priorities fail, it falls back to use `source cluster preferred version`.
I do not understand the sentence at my first glance `Use a priority system to determine which version to back up with the source preferred version as the default.`

design/restore-with-EnableAPIGroupVersions-feature.md
jenting: Both the screenshots if I remember correctly are commits format version `1.1.0`, it means that the source cluster and target cluster are running Velero after 1.4.

My question is will we support restores from the tarball with the legacy format version (Velero < 1.4)?
I guess yes because without enable the feature flag `EnableAPIGroupVersions`, the format version `1.1` and `1.1.0` almost look the same. Ref to https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/main/output-file-format.md

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: I admit that sentence is confusing. I'll split it up. Is this easier to understand:

`Use a priority system to determine which version to back up. The source preferred version will be the default if the priorities fail.`?

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: You bring up a really good point. Thank you for helping me to realize that what I proposed is not backward compatible with Velero versions earlier than 1.4.

If the backup tarball files do not follow `format version 1.1.0`, the code I have proposed will error when it shouldn't. It will try to set the chosen version to `v1-preferredversion`, but that version will not exist, and the restore will fail.

I'll update the design doc to check the directory names. If there is a directory with the substring `-preferredversion`, I know it is using `format version 1.1.0`. If not, I will let the program go on as usual, as if feature flag was not enabled.

Very good find, @jenting.

design/restore-with-EnableAPIGroupVersions-feature.md
jenting: Use a priority system to determine which version to "back up", should it be "restore from" 🤔 ?

codegold79: haha oops. Another good catch @jenting. Should be 
> Use a priority system to determine which version to restore with the source preferred version as the default.

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,148 @@
...
+- Using plugins to restore a resource when the target supports non of the source cluster's API group versions.
+
+## High-Level Design
+
+During restore, the proposal is that Velero will read the directory names in the backup tarball file and see which API group versions were backed up from the source cluster. Velero will also look at the target cluster and determine which API group versions are supported. Given the two lists of source and target cluster supported API versions, a priority system will be used to determine which version will be backed up. The priority of versions in order are target cluster preferred version, source cluster preferred version, and the latest common supported version. Once the version to restore is chosen, the file path to the backed up resource in the tarball will be modified such that it points to the resources' chosen API group version. If no version is found in common between the source and target clusters, the chosen version will default to the source cluster's preferred version (the version being restored currently without the changes proposal here). Restore will be allowed to continue as before.
```
jenting, line 25:
```suggestion
During restore, the proposal is that Velero will read the directory names in the backup tarball file and see which API group versions were backed up from the source cluster. Velero will also look at the target cluster and determine which API group versions are supported. Given the two lists of source and target cluster supported API versions, a priority system will be used to determine which version will be restored. The priority of versions in order are target cluster preferred version, source cluster preferred version, and the latest common supported version. Once the version to restore is chosen, the file path to the backed up resource in the tarball will be modified such that it points to the resources' chosen API group version. If no version is found in common between the source and target clusters, the chosen version will default to the source cluster's preferred version (the version being restored currently without the changes proposal here). Restore will be allowed to continue as before.
```

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Objective 3: List the API group versions supported by the target cluster
+
+Still within the `chooseAPIVersionsToRestore` method, the target cluster's resource group versions can now be obtained.
+
+```go
+targetGroupVersions := ctx.discoveryHelper.APIGroups()
+```
+
+### Objective 4: Use a priority system to determine which version to back up, if any
```
jenting, line 69:
```suggestion
### Objective 4: Use a priority system to determine which version to restore. The source preferred version will be the default if the priorities fail
```

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,148 @@
...
+
+1. Determine if the APIGroupVersionsFeatureFlag feature flag has been enabled.
+1. List the backed up API group versions.
+1. List the API group versions supported by the target cluster.
+1. Use a priority system to determine which version to restore. The source preferred version will be the default if the priorities fail.
+1. Modify the paths to the backup files in the tarball in the resource restore process.
+
+### Objective 1: Determine if the APIGroupVersionsFeatureFlag feature flag has been enabled
+
+For restore to be able to choose from multiple supported backed up versions, the feature flag must have been enabled during both the backup and restore processes. And the check to see if the feature flag is/was enabled must be done for both processes.
```
jenting, line 39: If I understand correctly, the design proposal _only_ supports the feature flag `APIGroupVersionsFeatureFlag` enabled on both the source and target cluster.
If yes, it'd be good to add this limitation to `Non Goals` chapter.

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: That makes sense. I'll add that change to my next commit.

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: 😆 I keep getting restore and backup confused. I'll get that into my next commit.

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: Good call. Yes, it will be a good limitation to add in the Non Goals section.

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -13,24 +13,38 @@ It is possible that between the time a backup has been made and a restore occurs
...
+  (1) reading the directory names in the backup tarball file and seeing which API group versions were backed up from the source cluster,
+  (2) looking at the target cluster and determining which API group versions are supported, and
+  (3) getting config maps on the cluster in order to get user-defined prioritization of versions.
+
+  The three lists will be used to create a map of chosen versions to restore. If there is a user-defined list of priority versions, the versions will be checked against the lists of target and source supported versions. The highest user-defined priority version that is/was supported by both target and source clusters will be the chosen version for that resource.
+
+  Without a user-defined prioritization of versions, the following version priority will be followed: target cluster preferred version, source cluster preferred version, and a common supported version. Should there be multiple common supported versions, the one that will be chosen will be based on the [Kubernetes version priorities](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority%E2%80%8B).
```
jenting, line 36:
```suggestion
  Without a user-defined prioritization of versions, the following version priority will be followed: target cluster preferred version, source cluster preferred version, and a common supported version. Should there be multiple common supported versions, the one that will be chosen will be based on the [Kubernetes version priorities](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority).
```

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -56,54 +70,83 @@ The `chooseAPIVersionsToRestore` method can remain in the `restore` package, but
...
+
+- **Priority 0** ((User override). Users determine restore version priority using a config map
+- **Priority 1**. Target preferred version can be used. This means one of two things:
+  - (A) target preferred version == source preferred version OR
+  - (B) target preferred version == source supported version
+- **Priority 2**. Source preferred version can be used. This means
+  - source preferred version == target supported version
+- **Priority 3**. A common supported version can be used. This means
+  - target supported version == source supported version
+  - if multiple support versions intersect, choose the version using the [Kubernetes’ version prioritization system](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority%E2%80%8B)
```
jenting, line 137:
```suggestion
  - if multiple support versions intersect, choose the version using the [Kubernetes’ version prioritization system](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority)
```

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: I will get that changed next commit.

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: I will get that changed next commit.

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,193 @@
...
+
+The reason for checking to see if the feature flag was enabled during backup is to ensure the changes made by this proposed design is backward compatible. Only with Velero version 1.4 and forward was Format Version 1.0.0 used to structure the backup directories. Format Version 1.0.0 is required for the restore process proposed in this design doc to work. Before v1.4, the backed up files were in a directory structure that will not be recognized by the proposed code changes. Therefore, restore should not attempt to restore from multiple versions as they will not exist.
+
+Checking if the feature flag is enabled during restore is straightforward using `features.IsEnabled(velerov1api.APIGroupVersionsFeatureFlag)`.
+
+Checking if the feature flag was enabled during backup can be done indirectly by checking the directory structure and version directory names. For backups generated using Format Version 1.0.0 (required for the changes proposed here), there will always be a version directory with `-preferredversion` in the directory name. Only one resource will need to be checked to determine if the feature flag was enabled during backup.
```
nrb, line 61: I think this paragraph is a little inaccurate.

A backup can be format 1.1.0 without the feature flag enabled, and should have the `-preferredversion` directory even without the feature flag. 

At backup time currently, the feature flag controls whether all versions are downloaded. By default, Velero only downloads the preferred version and stores it twice; once under the `-preferredversion` directory, once under a version-less directory.

Therefore, I think this could be simplified to only operate on `Backup` objects with `Status.FormatVersion == "1.1.0"`.  [Here's the actual const value](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/backup.go#L58).

nrb (changes requested):
This proposal looks pretty solid to me! I do have some clarifications I think we can make, but on the whole I'm comfortable with the direction.

@vmware-tanzu/velero-maintainers please also take a look and see what you think.

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,193 @@
...
+
+Status: Draft
+
+## Abstract
+
+This document proposes a solution to select an API group version to restore from the versions backed up using the feature flag EnableAPIGroupVersions.
+
+## Background
+
+It is possible that between the time a backup has been made and a restore occurs that the target Kubernetes version has incremented more than one version. In such a case where multiple versions of Kubernetes were skipped, the preferred source cluster's API group versions for resources may no longer be supported by the target cluster. With PR #2373, all supported API group versions were backed up if the EnableAPIGroupVersions feature flag was set during Velero installation. The next step (outlined by this design proposal) will be to see if any of the backed up versions are supported in the target cluster and if so, choose one to restore for each backed up resource.
```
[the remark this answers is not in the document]
codegold79: Yup, will do.

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,193 @@
...
+
+- Restore resources using the chosen API group version.
+- Choose an API group to restore from backups given a priority system or a user-provided prioritization of versions.
+
+## Non Goals
+
+- Allow users to restore onto a cluster that is is running a Kubernetes version older than the source cluster. The changes proposed here only allow for skipping ahead to a newer Kubernetes version, not older.
+- Allow restoring from backups created using Velero version 1.3 or older.
+- Modifying the compressed backup tarball files.
+- Using plugins to restore a resource when the target supports non of the source cluster's API group versions.
```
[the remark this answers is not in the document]
codegold79: Good catch.

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,193 @@
...
+## High-Level Design
+
+During restore, the proposal is that Velero will determine if the `APIGroupVersionsFeatureFlag` was enabled in both the source and target clusters (during backup and restore, respectively). Only if the feature flag was enabled in both clusters will the changes proposed here take effect.
+
+The proposed code starts with creating three lists for each backed up resource. The three lists will be created by
+  (1) reading the directory names in the backup tarball file and seeing which API group versions were backed up from the source cluster,
+  (2) looking at the target cluster and determining which API group versions are supported, and
+  (3) getting config maps on the cluster in order to get user-defined prioritization of versions.
+
+  The three lists will be used to create a map of chosen versions to restore. If there is a user-defined list of priority versions, the versions will be checked against the lists of target and source supported versions. The highest user-defined priority version that is/was supported by both target and source clusters will be the chosen version for that resource.
```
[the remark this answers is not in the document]
codegold79: Yes, if the user provides a value that doesn't exist in the source and/or target (the value must be in both), then the next paragraph will be executed. The next paragraph is to follow our pre-determined version prioritization (e.g. target preferred version is the highest priority). 

I'll add a sentence that if the user specified version is not supported by both target and source, it will be logged and the restore will continue with other priortizations.

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Objective 1: Determine if the APIGroupVersionsFeatureFlag feature flag has been enabled
+
+For restore to be able to choose from multiple supported backed up versions, the feature flag must have been enabled during both the backup and restore processes. And the check to see if the feature flag is/was enabled must be done for both processes.
+
+The reason for checking for the feature flag during restore is to ensure the user would like to restore a version that is among potentially multiple versions backed up with the feature flag enabled.
+
+The reason for checking to see if the feature flag was enabled during backup is to ensure the changes made by this proposed design is backward compatible. Only with Velero version 1.4 and forward was Format Version 1.0.0 used to structure the backup directories. Format Version 1.0.0 is required for the restore process proposed in this design doc to work. Before v1.4, the backed up files were in a directory structure that will not be recognized by the proposed code changes. Therefore, restore should not attempt to restore from multiple versions as they will not exist.
```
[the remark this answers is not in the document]
codegold79: Good eye--will change that.

design/restore-with-EnableAPIGroupVersions-feature.md
jenting: I had submitted a PR #2901 previously to fix the describe backup format version incorrectly (which was released at v1.5.0).
So, the only concern for me is if the source cluster uses Velero 1.4.0 to performs backup, then we can't use the format version to do the check.

codegold79: Thank you for the suggestion and another very good catch.

You're right that the paragraph is inaccurate. I will change it to say that we'll check for `Status.FormatVersion == "1.1.0"`. I realize from your observation that it is ok that during backup, the feature flag was not enabled. What's important are these two things:
(1) Feature flag is enabled during restore
(2) FormatVersion == 1.1.0

I'll be sure to change my wording to reflect those two conditions.

codegold79:
> This proposal looks pretty solid to me! I do have some clarifications I think we can make, but on the whole I'm comfortable with the direction.

Thank you @nrb for your careful review and very on-target and helpful comments. 

While writing out the code, I realized I reference the incorrect Kubernetes version sorting function. I have also decided to change some of the methods and variable names. I'll be sure to update the design doc tomorrow with yours and my updates.

design/restore-with-EnableAPIGroupVersions-feature.md
nrb: @jenting From conversations with Rafael, Velero won't be able to do this migration with backups made prior to v1.4.0 anyway. The feature flag and support for any API version besides the preferred version was only introduced in v1.4.0.

Also, see [line 21 of this document](https://github.com/vmware-tanzu/velero/pull/3050/files#diff-85b71029d5725d4b9e77b52b2276f38caa0086332e139cbe9f49b9b7abb7401cR21), which states that doing migration with Velero versions 1.3.x or older is a non-goal.

@codegold79 Thanks! The important thing about the feature flag during backup is whether or not more than 1 version is backed up, so that version negotiation can take place. Having a backup of Format 1.1.0 means you'll get the `-preferredversion` directory, but it doesn't guarantee you'll get multiple API versions - that only happens when the feature flag is on. So I think that may be useful to check for still, though it's _also_ possible that a given API only has 1 version, anyway...which makes that heuristic kind of unreliable.

Off the top of my head, I don't think we write the feature flags that were enabled anywhere in the backup data, but maybe that would be useful so we don't have to use these kinds of nondeterministic heuristics.

design/restore-with-EnableAPIGroupVersions-feature.md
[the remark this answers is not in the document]
nrb: Thanks! I found it later in the document too, but I think a brief sentence here would be helpful.

design/restore-with-EnableAPIGroupVersions-feature.md
nrb: @jenting Sorry, I misunderstood your comment.

The code wouldn't have to use `velero describe`, it would use the `FormatVersion` field directly, which should still be reliable. The PR you referenced is only about the display on `velero describe`, not whether the `FormatVersion` field gets set correctly.

jenting: Thanks for the detailed explanation 🙏, I have no other concerns.

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,211 @@
...
+
+The reason for checking `Status.FormatVersion` is to ensure the changes made by this proposed design is backward compatible. Only with Velero version 1.4 and forward was Format Version 1.1.0 used to structure the backup directories. Format Version 1.1.0 is required for the restore process proposed in this design doc to work. Before v1.4, the backed up files were in a directory structure that will not be recognized by the proposed code changes. In this case, restore should not attempt to restore from multiple versions as they will not exist.
+
+Checking if the feature flag is enabled during restore is straightforward using `features.IsEnabled(velerov1api.APIGroupVersionsFeatureFlag)`.
+
+Checking the `Status.FormatVersion` can be done indirectly by checking the directory structure and version directory names. For backups generated using Format Version 1.1.0, there will always be a version directory with `-preferredversion` in the directory name. Only one resource will need to be checked to determine if the feature flag was enabled during backup.
```
nrb, line 61: Can this paragraph be deleted, since on L57 you're looking directly at the backup object's `Status.FormatVersion` field? Or is this a double check to be sure?

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,211 @@
...
+
+To help with the format version check, a new parser object can be created using the construction function `NewParser(...)` and we can take advantage of the `archive` package's `Parser` type. A method for the parser object will be created that has the signature `UsedFormatVersion(dir string) (bool, error)`. The method will loop through the version directories inside one of the resource.group directories and determine if a directory with `-preferredversion` was found.
```
nrb, line 63: Similar question - is this paragraph, and the `Parser` needed since you're using the `Status.FormatVersion` filed directly instead of inferring it based on directory structure?

design/restore-with-EnableAPIGroupVersions-feature.md
```diff
@@ -0,0 +1,211 @@
...
+
+A map will be created to store the user-defined priority versions. The map's keys will be strings in the format `<resource>.<group>`. The values will be APIGroup structs that will be imported from `k8s.io/apimachinery/pkg/apis/meta/v1`. Within the APIGroup structs will be versions in the order that the user provides in the config map. The PreferredVersion field in APIGroup struct will be left empty.
+
+### Objective 5: Use a priority system to determine which version to restore. The source preferred version will be the default if the priorities fail
+
+Determining the priority will also be done in the `chooseAPIVersionsToRestore` method. Once a version is chosen, it will be stored in a new map of the form `map[string]ChosenGRVersion` where the key is the `<resource>.<group>` and the values are of the `ChosenGrpVersion` struct type (shown below). The map will be saved to the `restore.Context` object in a field called `chosenGrpVersToRestore`.
+
+```go
+type ChosenGrpVersion struct {
```
nrb, line 127: I think this could be completely spelled out in the type name.

nrb (changes requested):
I have 1 change to request and two questions. Thanks for the very detailed proposal! This is looking very solid!

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: No problem. I'll name the type to `ChosenGroupVersion`.

design/restore-with-EnableAPIGroupVersions-feature.md
nrb: To clarify - [the Backup object is included in a restore Request](https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L81).

It is [also on the restoreContext](https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L229).

So we have direct access to the Velero `Backup` CR that's being restored, not just the tarball.

codegold79: Oh, I think I misread something you wrote earlier. I do see `status.formatVersion` inside `velero-backup.json`. Would this be where I grabbed the format version from?

And if I can do that, I definitely won't need to be looking through the directories to look for `-preferredversion` in the names.

design/restore-with-EnableAPIGroupVersions-feature.md
codegold79: I'd be able to get rid of `UsedFormatVersion` function, but I'll still need `Parser` to read the version directories, e.g. `v1-preferredversion`, `v2`, etc.

design/restore-with-EnableAPIGroupVersions-feature.md
nrb: I think in this case, it would end up being from the restoreContext.

Something like `ctx.backup.Status.FormatVersion`.

You shouldn't have to parse the `velero-backup.json` yourself, as that's done for you. The JSON file will end up in the `ctx.backup` variable as a full Go struct.

codegold79: Nice! Thanks so much @nrb. I wrote my last question after your clarification. I'll fix the paragraphs to use the backup info in the `restoreContext`. That's much better.

nrb: The `Backup` comes from [this location](https://github.com/vmware-tanzu/velero/blob/6808acd92e30848056a21faf373af03ddb8a3b71/pkg/controller/restore_controller.go#L403) up in the restore controller, as an FYI. :)

So it's not directly from the JSON file, but rather the in-cluster representation. The JSON file is used when Velero syncs into brand new clusters, though.

ashish-amarnath (approved):
Thanks for the detailed proposal.
This LGTM.

@nrb Handing this to you for sign-off and merge.

nrb (approved):
Thanks for these revisions and all your hard work on this!

Issue #3027, opened 2020-10-21
Restore fails to complete with error on AzureFile volumes


sivaramsk:
**What steps did you take and what happened:**
1. Installed Minio with Azure gateway.
2. Installed velero with aws plugin with minio endpoint, along with --use-restic and --default-volumes-to-restic flags
3. Triggered a backup(excluding namespaces velero and kube-system). The backup was successful the podvolumebackup did not show error on the volumes that I wanted to backup(for ex: "volume: datadir"). podvolumebackup output attached to the gist.
4. Restore of the above backup got struck after a point. The podvolumerestore had this error for the volumes with SC azurefile. podvolumerestore output attached to the gist.

```
message: |-
        error restoring volume: error running restic restore, cmd=restic restore --repo=s3:http://minio-server.minio.svc:9000/velero/restic/org2-net --password-file=/tmp/velero-restic-credentials-org2-net122979599 --cach    e-dir=/scratch/.cache/restic 54f54e51 --target=., stdout=, stderr=: error getting snapshot size: error running command, stderr=Fatal: error loading snapshot: no matching ID found
```

**What did you expect to happen:**
Restore to become successful. 

**The output of the following commands will help us better understand what's going on**:
(Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)

https://gist.github.com/sivaramsk/62ce61275335de296989378482f5aeb4

**Anything else you would like to add:**
Storageclass was patched with nouser_xattr before the backup. 

**Environment:**

- Velero version (use `velero version`): 
   Client:
        Version: v1.5.1
        Git commit: -
   Server:
        Version: v1.5.1

- Velero features (use `velero client config get features`): features: <NOT SET>
- Kubernetes version (use `kubectl version`): 
   Client Version: 1.18.8
   Server Version: 1.17.11
- Kubernetes installer & version: AKS, created with terraform aks modules
- Cloud provider or hardware configuration: AKS

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

dsu-igeek:
This error is in the backup-logs.txt:

[...]

Issue #2956, opened 2020-09-21
restic-wait init pod stuck when restoring statefulset with restic plugin


wanghuok:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is, and what commands you ran.)

**i tried to test velero backup and restore on entire cluster:**
1. i create 2 clusters with 3 master nodes and 5 worker nodes respectively
2. in src cluster, i create 50 namespaces, 100 deployments with 2 replicas, 100 loadbalancers
3. additionally, i create 5 ns, each has 1 statefulset with 2 replicas, each replica has 1G pv
4. also, i create 2 ns, each has 1 statefulset with 1 replica, 10G pv
5. write data to the 1G PV and 10G PV seperatly.
6. verify db write correctly
7. annotate pods with pv, and so on for the rest:
`kubectl --context=pks-st-10 -n velero-test-1g-1 annotate pod/postgres-wl-1-0 backup.velero.io/backup-volumes=postgredb
kubectl --context=pks-st-10 -n velero-test-1g-1 annotate pod/postgres-wl-1-1 backup.velero.io/backup-volumes=postgredb`
8. backup entire cluster is succeed:
`velero --kubecontext pks-st-10 backup create wa-bk-2`
logs are uploaded here: https://gist.github.com/wanghuok/9da292ad60b81281edf3d2e86a62dd3f
9. restore to dst cluster: pks-st-11:
`velero --kubecontext pks-st-11 restore create wa-rs-2 --from-backup wa-bk-2`
10. restore failed with PartiallyFailed:
`Name:         wa-rs-2
Namespace:    velero
Labels:       <none>
Annotations:  <none>

Phase:  PartiallyFailed (run 'velero restore logs wa-rs-2' for more information)

Warnings:
  Velero:     <none>
  Cluster:  could not restore, customresourcedefinitions.apiextensions.k8s.io "backups.velero.io" already exists. Warning: the in-cluster version is different than the backed-up version.
            could not restore, customresourcedefinitions.apiextensions.k8s.io "backupstoragelocations.velero.io" already exists. Warning: the in-cluster version is different than the backed-up version.
            could not restore, customresourcedefinitions.apiextensions.k8s.io "deletebackuprequests.velero.io" already exists. Warning: the in-cluster version is different than the backed-up version.
            could not restore, customresourcedefinitions.apiextensions.k8s.io "downloadrequests.velero.io" already exists. Warning: the in-cluster version is different than the backed-up version.
[...]

Issue #2691, opened 2020-07-07
PartiallyFailed hit when restoring 2 namespaces with restic plugin in PKS env


luwang-vmware:
**What steps did you take and what happened:**
This issue can be reproducible. 
1. create 2 namespace, named as ns-1g-1 and ns-20g-1.
2. in the two namespaces, I both created statefulset with PV,  one is with 1G PV, and the other is with 20G PV.
3. Write data to the 1G PV and 20G PV seperatly.
4. annotate  the two pods 
```
kubectl --context=pks-st-2 -n ns-1g-1 annotate pod/postgres-wl-0 backup.velero.io/backup-volumes=postgredb
kubectl --context=pks-st-2 -n ns-20g-1 annotate pod/postgres-wl-0 backup.velero.io/backup-volumes=postgredb
```
5. backup is succeed 
`velero --kubecontext=pks-st-2 backup create 1g-20g-backup  --include-namespaces ns-1g-1,ns-20g-1`
```
root@6c6fe943453d:~/velero-v1.4.0-linux-amd64/post/scale-testing# velero --kubecontext=pks-st-2  backup describe 1g-20g-backup --details
Name:         1g-20g-backup
Namespace:    velero
Labels:       velero.io/storage-location=default
Annotations:  velero.io/source-cluster-k8s-gitversion=v1.17.5+vmware.1
              velero.io/source-cluster-k8s-major-version=1
              velero.io/source-cluster-k8s-minor-version=17

Phase:  Completed

Namespaces:
  Included:  ns-1g-1, ns-20g-1
  Excluded:  <none>

Resources:
  Included:        *
  Excluded:        <none>
  Cluster-scoped:  auto

Label selector:  <none>

Storage Location:  default

Velero-Native Snapshot PVs:  auto

TTL:  720h0m0s

Hooks:  <none>

Backup Format Version:  1

Started:    2020-07-06 10:52:06 +0000 UTC
Completed:  2020-07-06 10:58:48 +0000 UTC

Expiration:  2020-08-05 10:52:06 +0000 UTC

Total items to be backed up:  42
Items backed up:              42

Resource List:
  apps/v1/ControllerRevision:
    - ns-1g-1/postgres-wl-64dbb9bc58
    - ns-20g-1/postgres-wl-64dbb9bc58
  apps/v1/StatefulSet:
    - ns-1g-1/postgres-wl
    - ns-20g-1/postgres-wl
  v1/ConfigMap:
[...]

PR #3051, opened 2020-11-02
🐛  Use namespace and name to match PVB to Pod restore

ashish-amarnath:

Fixes: #3027
Fixes: #2691 
Fixes: #2956 

@zubron Thanks for the debugging and repro example.

### DIFF
--- a/pkg/restic/common.go
+++ b/pkg/restic/common.go
@@ -95,13 +95,17 @@ func getPodSnapshotAnnotations(obj metav1.Object) map[string]string {
 	return res
 }
 
+func isPVBMatchPod(pvb *velerov1api.PodVolumeBackup, pod metav1.Object) bool {
+	return pod.GetName() == pvb.Spec.Pod.Name && pod.GetNamespace() == pvb.Spec.Pod.Namespace
+}
+
 // GetVolumeBackupsForPod returns a map, of volume name -> snapshot id,
 // of the PodVolumeBackups that exist for the provided pod.
 func GetVolumeBackupsForPod(podVolumeBackups []*velerov1api.PodVolumeBackup, pod metav1.Object) map[string]string {
 	volumes := make(map[string]string)
 
 	for _, pvb := range podVolumeBackups {
-		if pod.GetName() != pvb.Spec.Pod.Name {
+		if !isPVBMatchPod(pvb, pod) {
 			continue
 		}
 
--- a/pkg/restic/common_test.go
+++ b/pkg/restic/common_test.go
@@ -564,6 +564,88 @@ func TestGetPodVolumesUsingRestic(t *testing.T) {
 	}
 }
 
+func TestIsPVBMatchPod(t *testing.T) {
+	testCases := []struct {
+		name     string
+		pod      metav1.Object
+		pvb      velerov1api.PodVolumeBackup
+		expected bool
+	}{
+		{
+			name: "should match PVB and pod",
+			pod: &metav1.ObjectMeta{
+				Name:      "matching-pod",
+				Namespace: "matching-namespace",
+			},
+			pvb: velerov1api.PodVolumeBackup{
+				Spec: velerov1api.PodVolumeBackupSpec{
+					Pod: corev1api.ObjectReference{
+						Name:      "matching-pod",
+						Namespace: "matching-namespace",
+					},
+				},
+			},
+			expected: true,
+		},
+		{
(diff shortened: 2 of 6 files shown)


ashish-amarnath:
Here is the sample workload, from Bridget, that gives us a repro https://gist.github.com/zubron/dce63c10b5aea026f988b8a14d2934c5#file-app-yaml

pkg/restic/common_test.go
```diff
@@ -564,6 +564,88 @@ func TestGetPodVolumesUsingRestic(t *testing.T) {
...
+					Pod: corev1api.ObjectReference{
+						Name:      "matching-pod",
+						Namespace: "matching-namespace",
+					},
+				},
+			},
+			expected: false,
+		},
+		{
+			name: "should match PVB and pod, pod namespace mismatch",
```
zubron, line 607:
```suggestion
			name: "should not match PVB and pod, pod namespace mismatch",
```

pkg/restic/common_test.go
```diff
@@ -564,6 +564,88 @@ func TestGetPodVolumesUsingRestic(t *testing.T) {
...
+					Pod: corev1api.ObjectReference{
+						Name:      "matching-pod",
+						Namespace: "matching-namespace",
+					},
+				},
+			},
+			expected: false,
+		},
+		{
+			name: "should match PVB and pod, pod name and namespace mismatch",
```
zubron, line 623:
```suggestion
			name: "should not match PVB and pod, pod name and namespace mismatch",
```

zubron (changes requested):
Thanks for fixing this, @ashish-amarnath! Just a couple of minor suggestions in test names but is otherwise good. Also looks like some other restore tests need to be updated following this change.

zubron:
This will also fix #2691 and #2956, I looked at those two issues again and they also have the same problem of using the same pod name across multiple namespaces.

nrb:
Could I get a summary of what the issue was? Were PVBs being assigned to the wrong pod?

zubron:
@nrb I wrote up my findings for this issue in these two comments: https://github.com/vmware-tanzu/velero/issues/3027#issuecomment-718764255 and https://github.com/vmware-tanzu/velero/issues/3027#issuecomment-719071517

pkg/restic/common.go
```diff
@@ -95,13 +95,17 @@ func getPodSnapshotAnnotations(obj metav1.Object) map[string]string {
 	return res
 }
 
+func isPVBMatchPod(pvb *velerov1api.PodVolumeBackup, pod metav1.Object) bool {
+	return pod.GetName() == pvb.Spec.Pod.Name && pod.GetNamespace() == pvb.Spec.Pod.Namespace
```
nrb, line 99: This is probably preference and I'm not  going to mark it as a blocker, but could we use [`NamespaceAndName`](https://github.com/vmware-tanzu/velero/blob/f42c63af1b9af445e38f78a7256b1c48ef79c10e/pkg/util/kube/utils.go#L36) to do the comparisons?

nrb: Oh, duh, no we can't cause the PVB has the pod's namespace and name nested a layer deeper than that function is looking for.

ashish-amarnath:
To summarize the bug here.
 Velero uses only the pod's name to find its corresponding podvolumebackups. When there are pods with the same name but in different namespaces, the wrong podvolume backup is matched with the wrong pod.

The fix here is to use both the pod name and its namespace to find the podvolumebackups that pertain to the pod.

Looking at the CI failures now.

zubron (approved):
Thanks!

pkg/restic/common.go
```diff
@@ -95,13 +95,17 @@ func getPodSnapshotAnnotations(obj metav1.Object) map[string]string {
 	return res
 }
 
+func isPVBMatchPod(pvb *velerov1api.PodVolumeBackup, pod metav1.Object) bool {
```
carlisia, line 98: @ashish-amarnath if you prefer to leave it like this I'd defer to you, but it'd be preferable if you passed 4 strings (pvb name and ns, pod name and ns) instead of 2 full objects into this function.

ashish-amarnath: The PVB is passed as a pointer. The `metav1.Object` shouldn't be that bad.
So I would prefer it this way.
Marking as resolved. Please unresolved if not ok. 🙂

nrb (approved):

PR #3057, opened 2020-11-07
Fix various typos found by codespell

invidian:
By running the following command:

codespell -S .git,*.png,*.jpg,*.woff,*.ttf,*.gif,*.ico -L \
iam,aks,ist,bridget,ue

Signed-off-by: Mateusz Gozdek <mgozdekof@gmail.com>

If you like, I would also recommend adding GitHub Action running codespell on every PR to avoid more typos being added. See https://github.com/rook/rook/pull/6564 for example.


carlisia (commented):
Whoa did we have a lot of typos!

This is great. Some big files are not ours (outside packages and json files) and we should exclude those to avoid such a big churn, for one.

invidian:
>This is great. Some big files are not ours (outside packages and json files) and we should exclude those to avoid such a big churn, for one.

Can you please be more specific, which files should be ignored? It's hard for me to guess that.

carlisia (approved):
I looked over again and nvm my previous comment. The json files are test data, not needed but def no issues there. And there's only 1 file that is not ours (site/static/fonts/Metropolis/README.md) but this is not a big deal at all.

Thanks for doing this!!

Want to do the same in our other Velero repos? :)

invidian:
>Want to do the same in our other Velero repos? :)

Not really :wink: But I can add a GitHub Actions Workflow which will run `codespell` automatically on the PRs. What do you think?

nrb:
@invidian I would definitely welcome the GitHub Action PR!

nrb (approved):

invidian:
Created #3064 as a follow up then :)

PR #3089, opened 2020-11-17
Fix project automation

carlisia:
This will fix the GH action that automates adding a new issue into the triage project board.

What was needed:
- A personal GH token with full repo access (read/write). I created one under my account, named `GH_TOKEN` (does not need to match the key name below)*
- Add a secret key/value pair to the Velero repo with a key value same as the value for `repo-token` (`secrets.GH_TOKEN`), and the created token value (https://github.com/vmware-tanzu/velero/settings/secrets/actions). This step will need to be repeated in each repo that will push issues into the board.

* Note 1: there is no option to create a GH token for an organization. Ideally we would have a "bot" account to use for such purposes. We [were using one](https://github.com/vmware-tanzu/helm-charts/issues/20) of my personal tokens for the helm release process.
Note 2: it is not possible to use  GH default `secrets.GITHUB_TOKEN` because it only scopes at the repository level.

### DIFF
--- a/.github/workflows/opened-issues-triage.yml
+++ b/.github/workflows/opened-issues-triage.yml
@@ -12,4 +12,4 @@ jobs:
         with:
           project: "Velero Support Triage Board"
           column: "Needs Triage"
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          repo-token: ${{ secrets.GH_TOKEN }}
--- a/.github/workflows/opened-issues-triage.yml
+++ b/.github/workflows/opened-issues-triage.yml
@@ -11,5 +11,5 @@ jobs:
       - uses: alex-page/github-project-automation-plus@v0.3.0
         with:
           project: "Velero Support Triage Board"
-          column: "Needs Triage"
+          column: "Needs triage"
           repo-token: ${{ secrets.GH_TOKEN }}


carlisia:
A couple notes about this effort:

- project-bot
There is a cool automation project via GH apps which doesn't require GH actions and provide the ability to have cards move to other columns based on label. Unfortunately, it does not work at an organization level, even though the documentation seems to indicate it does. But, more relevant, the project is stale. 
See: https://github.com/philschatz/project-bot

- assign-one-project-github-action
This is the only project I found that is alternative to the one we are using. It does exactly the same thing except that it adds the ability to extend the rule/filtering used to add cards to a project board (can add labels for example). It does not have automation to move cards between columns once the card is added to a board.
See: https://github.com/srggrs/assign-one-project-github-action

- workflow based on labels
Lastly, between GH apps and GH actions, there doesn't seem to be any existing option to automatically move cards across columns based on label changes, or based on any other criteria for that matter. Sad. This will all have to be done manually if using GH Projects.

ashish-amarnath (approved):

zubron (approved):
Thanks for your efforts on this, @carlisia!

I agree that it would be better to have a bot account for these kinds of integrations, but I think we can proceed with this for now so that we can make progress in understanding our needs around boards and automation.

PR #3092, opened 2020-11-18
feat: support configure BSL CR to indicate which one is the default

### LINKED ISSUE #2421: Add columns to backup storage location to indicate which one is the default

jenting:
Add "default" for backup storage location
1. Add `--default` flag under `velero backup-location create` to sets this new location to be the new default backup location.
2. Add sub-command `velero backup-location set` and flag `--default` to sets this new location to be the new default backup location.
3. Add `--default` flag under `velero backup-location get` to displays the current default backup storage location.
4. Add a new print column **DEFAULT** under `velero backup get` and `kubectl get backupstoragelocations` to indicate which backup location is the default one, for example:
```
$ velero backup-location get
NAME        PROVIDER   BUCKET/PREFIX   PHASE       LAST VALIDATED                  ACCESS MODE   DEFAULT
default     aws        velero          Available   2020-11-17 15:21:12 +0800 CST   ReadWrite     true
secondary   aws        velero          Available   2020-11-17 15:20:42 +0800 CST   ReadWrite

$ kubectl get backupstoragelocations
NAME        PROVIDER   BUCKET   PHASE       LAST VALIDATED   ACCESS MODE   AGE    DEFAULT
default     aws        velero   Available   21s              ReadWrite     120m   true
secondary   aws        velero   Available   51s              ReadWrite     93m 
```
5. It covers the upgrade case: after CRD upgrade, the BSL controller will set default BSL if the BSL name matches the designated backup storage location (`velero server --default-backup-storage-location`). After that, the user could specify another BSL as default by `velero backup-location set <NAME> --default`. (The unit tests cover this scenario).

Fixes #2421 (updated on 11/18: cleaned up 2421 to only contain this fix)
ref to #2419


pkg/cmd/util/output/backup_storage_location_printer.go
```diff
@@ -50,6 +51,11 @@ func printBackupStorageLocation(location *velerov1api.BackupStorageLocation) []m
 		Object: runtime.RawExtension{Object: location},
 	}
 
+	isDefault := ""
+	if location.Spec.Default {
+		isDefault = "true"
```
carlisia, line 56: Please make this uppercase. It will match the value of columns like PHASE [Available/Unavailable].

carlisia (changes requested):
This is such great work, thank you! I'm very much looking forward to adding this.

I have a minor request.

I also found a bug. steps to reproduce:
- be sure to have some existing BSLs created against `main` or 1.5.x
- on this branch, run `make local`
- update the CRDs
- run `v get backup-location --default`
- get panic here: https://github.com/vmware-tanzu/velero/pull/3092/files#diff-692b6ad01521aa5bc42b3f67637f50c59c3dc1481f0d372e093a011e3109e61eR74

```
❯ v get backup-location --default
panic: runtime error: slice bounds out of range [2:1]

goroutine 1 [running]:
github.com/vmware-tanzu/velero/pkg/cmd/cli/backuplocation.NewGetCommand.func1(0xc0009a7340, 0xc000885220, 0x0, 0x1)
        /Users/carlisiac/work/src/github.com/vmware-tanzu/velero/pkg/cmd/cli/backuplocation/get.go:74 +0x7d0
github.com/spf13/cobra.(*Command).execute(0xc0009a7340, 0xc000885200, 0x1, 0x1, 0xc0009a7340, 0xc000885200)
        /Users/carlisiac/work/pkg/mod/github.com/spf13/cobra@v0.0.7/command.go:842 +0x29d
github.com/spf13/cobra.(*Command).ExecuteC(0xc0000d0840, 0x6, 0xc0000d0840, 0x6)
        /Users/carlisiac/work/pkg/mod/github.com/spf13/cobra@v0.0.7/command.go:943 +0x317
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/carlisiac/work/pkg/mod/github.com/spf13/cobra@v0.0.7/command.go:883
main.main()
        /Users/carlisiac/work/src/github.com/vmware-tanzu/velero/cmd/velero/main.go:34 +0x91
```

pkg/cmd/util/output/backup_storage_location_printer.go
jenting: The Default is a _boolean_ type, so calling `kubectl get backupstoragelocations` prints `true`.
Therefore, I make the velero `backup-location get` output format aligns with `kubectl get backupstoragelocations`.

If we want to match the value of columns like PHASE, then we have to change the data type to _string_, it makes the comparison and the meaning more complicated, WDYT?

jenting:
This is ready to be reviewed.

pkg/cmd/util/output/backup_storage_location_printer.go
carlisia: Right, this makes sense. Good call.

carlisia (commented):
@jenting thanks for fixing the panic. 

I'm not sure about the "per provider" setting option. To recap, the purpose of having a default BSL is so one doesn't have to be specified in the cmd line when creating a backup. So, a user can have multiple BSLs, each for different providers, and only 1 of them should be the default, regardless of the provider. Otherwise, the user would have to specify what provider to use, and then the BSL used would be the default. Needing to define the provider would sorta defeat the purpose.

carlisia (changes requested):
There is one additional aspect to consider: when running a backup, if the backup does not include a BSL, then it will use the default location. Currently, this default is set on the server (https://github.com/vmware-tanzu/velero/blob/a1e182e723a8c5f6d4175d8db2361233a94d2502/pkg/cmd/server/server.go#L131).

This check is done here: https://github.com/vmware-tanzu/velero/blob/a1e182e723a8c5f6d4175d8db2361233a94d2502/pkg/controller/backup_controller.go#L346

So, in the backup controller, we will now want to check:
- is there a given BSL for the backup being created? If yes, use that
- if not, check if one of the existing BSLs is marked as default. If yes, use that
- If not, use the server default (I personally would like to get rid of this)

The same idea applies with the sync: https://github.com/vmware-tanzu/velero/blob/a1e182e723a8c5f6d4175d8db2361233a94d2502/pkg/controller/backup_sync_controller.go#L133

- to find out which default location to sync first, first search the existing locations for one marked as "default"; if none exists, use the name of the server one (this might or not exist).

carlisia:
@vmware-tanzu/velero-maintainers, bringing this up here since it's in context: having now (after this is merged) the ability to set which existing BSL is the default, what do you think of removing the server configuration with the default BSL name? While I understand how it made it super easy to "magically" get a BSL setup with the install cmd, it has always bugged me that the configuration for the default was in name only, in other words, a BSL itself with that name might or not exist.

The change would be: At backup creation, if there is no BSL configured to be the default, we would warn for the user to either set it or specify it with the proper flag, and not allow the backup creation. I think this is infinitely better usability than allowing them to create a backup for a BSL that might not even exist, then trying to communicate that with log messages on the controller side.

This change does not have to be in this PR. I'd be glad to work on it so it is merged right after this one.

nrb (changes requested):
> it has always bugged me that the configuration for the default was in name only, in other words, a BSL itself with that name might or not exist.

That's not the only way to set the default. On the `velero server` command, the `--default-backup-storage-location` argument can be used to specify which other BackupStorageLocation to use. Using the location named `default` gets the 80-90% use case where users don't really care what it's named.

That said: I like the approach in this PR better :) It should remove the need to restart Velero for changing default BackupStorageLoctions. And it moves that change to a client-side command that's not `kubectl edit`.

I do want to be sure we handle the transition gracefully, though - there needs to be a deprecation period of at least 1 release, with documentation and notice. So I want to have both be usable in v1.6.0, personally. That's probably going to require some code that makes the existing flag edit the BSL CRs in much the same way this does PR does. However, I also don't think that compatibility needs to be done in this same PR - it just needs to be done by v1.6.0 release.

@jenting, can you please add documentation for using this into https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/main/locations.md ?

carlisia:
> can be used to specify which other BackupStorageLocation to use

Correct. And yet, any name can go there meaning, there might not be an existing BSL with that name. When it works, it's great, but when it doesn't, it's quirky to try and communicate that.

jenting:
> The same idea applies with the sync:
> 
> https://github.com/vmware-tanzu/velero/blob/a1e182e723a8c5f6d4175d8db2361233a94d2502/pkg/controller/backup_sync_controller.go#L133
> 
> * to find out which default location to sync first, first search the existing locations for one marked as "default"; if none exists, use the name of the server one (this might or not exist).

Good call, I'll double-check on it.

jenting:
> @jenting thanks for fixing the panic.
> 
> I'm not sure about the "per provider" setting option. To recap, the purpose of having a default BSL is so one doesn't have to be specified in the cmd line when creating a backup. So, a user can have multiple BSLs, each for different providers, and only 1 of them should be the default, regardless of the provider. Otherwise, the user would have to specify what provider to use, and then the BSL used would be the default. Needing to define the provider would sorta defeat the purpose.

okay, I think I might be overthinking on each provider should have a defaulted BSL, I'll remove it.

jenting:
To recap, from the design doc https://github.com/vmware-tanzu/velero/blob/master/design/cli-install-changes.md#deprecation, the time to deprecated the commands and flags was scheduled at Velero 2.0.

So we should make sure both the server-side and the client-side change the default BSL both work correctly.

changelogs/unreleased/3092-jenting
```diff
@@ -0,0 +1 @@
+feat: support configure BSR CR to indicate which one is the default
```
nrb, line 1: 1) I think you mean BSL instead of BSR?
2) Could you spell it out completely (BackupStorageLocation)?

pkg/cmd/server/server.go
```diff
@@ -195,7 +195,7 @@ func NewCommand(f client.Factory) *cobra.Command {
...
 	command.Flags().StringSliceVar(&config.disabledControllers, "disable-controllers", config.disabledControllers, fmt.Sprintf("List of controllers to disable on startup. Valid values are %s", strings.Join(controller.DisableableControllers, ",")))
 	command.Flags().StringSliceVar(&config.restoreResourcePriorities, "restore-resource-priorities", config.restoreResourcePriorities, "Desired order of resource restores; any resource not in the list will be restored alphabetically after the prioritized resources.")
-	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "Name of the default backup storage location.")
+	command.Flags().StringVar(&config.defaultBackupLocation, "default-backup-storage-location", config.defaultBackupLocation, "Name of the default backup storage location. DEPRECATED: this flag will be removed in v2.0. Use \"velero backup-location set --default\" instead.")
```
nrb, line 198: What happens if both of these are set? We should document that behavior.

nrb (changes requested):
I think this looks really good, and thank you for pointing out the design doc deprecation schedule!

I noted a few changes inline, but also I have another - could you create a new file, `site/contents/docs/main/upgrade-to-1.6.md` and document this new command, since it's an added feature and users upgrading should know? I think after these changes, I'm 👍 .

pkg/cmd/server/server.go
jenting: Let me summarize the scenario I've thought:
1. During the upgrade, the BSL controller sets the defaulted BSL according to the BSL name.
2. The user could alert the defaulted BSL by velero CLI.
3. If the user changes the `velero server --default-backup-storage-location`, it takes no effect anymore.

The only way we can't prevent is the user directly edit BSL CR to have more than one BSLs having `.spec.default=true`.

jenting: Let me summarize the scenario I've thought:
1. During the upgrade, the BSL controller sets the defaulted BSL according to the BSL name because none of the BSLs be marked as default.
2. The user could change the defaulted BSL by velero CLI.
3. Now that we have the default BSL. If the user changes the `velero server --default-backup-storage-location`, it takes no effect anymore. The only way to configure which BSL is the default is to change BSL CR (velero CLI _or_ `kubectl edit backupstoragelocations`).

The only way we can't prevent is the user directly edit BSL CRs to have more than one BSLs having `.spec.default=true`.

changelogs/unreleased/3092-jenting
jenting: I've updated it, please take a look.

pkg/cmd/server/server.go
nrb: That behavior sounds to me like it's not deprecating this flag, but making it entirely a no-op. To me, that's worse than removing the flag, because if it's present, users will expect it to work.

Deprecation means that it should still work, but is not the preferred method and will be removed.

Ideally, I think the flag should remain until complete removal and cause the server to execute similar logic to what the command does by altering the CRDs. This would happen at server startup, so it's a "one-time" operation, but the users can still use the new command to alter the default.

jenting: > Ideally, I think the flag should remain until complete removal and cause the server to execute similar logic to what the command does by altering the CRDs. This would happen at server startup, so it's a "one-time" operation, but the users can still use the new command to alter the default.

sure, I'll make server-side and client-side change both work.

jenting: I think another use case that might break the user's expectation, for example, we have 2 BSLs, one is `default` and the one is `secondary`.
1. During the upgrade, the BSL controller sets BSL `default` and default BSL.
2. The user changes the defaulted BSL to `secondary` by Velero CLI.
3. The Velero pod restart due to deleting the pod, then the velero server pod restart so it'll set the `default` BSL as the default.

nrb: > It means that every time the velero pod start/restart, it might change the default BSL setting.

Yeah, that's a very good point. That's definitely not something we want, because that will send data to the wrong place based on other configuration changes. Perhaps the server-level option sets the default BSL once and only once, for migration purposes, then it doesn't work anymore?

How about we address these in a follow up PR to keep this one on track? I think the implementation here is pretty good, but the challenges for making it backwards compatible look trickier than I first thought and probably warrant their own set of focused changes.

nrb (dismissed):
I like this functionality a lot, but I would like to see a follow up PR that makes the `--default-storage-location` argument on `velero server` set the `Default` field.

Since that change might be a little tricky, I'm ok with merging this change and then focusing another PR on backwards compatibility.

carlisia (changes requested):
This is such a good feature to have but it's also complicated because of the way we currently try to default to a BSL. All the get/set/create are working. But I found this bug:

1. Create a BSL named "default"
2. Create another BSL (say, "mystorage"), and make that the default.
3. Create a backup, but don't specify the BSL to use. It should use "mystorage".
3. Describe the backup and you'll see the BSL used was "default", when it should've been "mystorage" since that was set as the default.

carlisia:
I'm still looking at the code for this PR to resolve the issue/bug above. The logic is per my guidance, but I wasn't taking into account the server level config that sets, by default, a BSL named "default" as default, even when there's no BSL with that name. Thinking of some possible ways out of this.

jenting:
> This is such a good feature to have but it's also complicated because of the way we currently try to default to a BSL. All the get/set/create are working. But I found this bug:
> 
> 1. Create a BSL named "default"
> 2. Create another BSL (say, "mystorage"), and make that the default.
> 3. Create a backup, but don't specify the BSL to use. It should use "mystorage".
> 4. Describe the backup and you'll see the BSL used was "default", when it should've been "mystorage" since that was set as the default.

Thanks for the review, however, I can't reproduce it follows the above steps. Here are the commands:
```
# make sure no BSL
$ velero backup-location get

# create BSL named "default"
$ velero backup-location create default \
    --provider aws \
    --bucket velero \
    --config region=minio-default,s3ForcePathStyle=true,s3Url=http://minio-default.velero.svc.cluster.local:9000

# create BSL named "mystorage" and make it default
$ velero backup-location create mystorage \
    --provider aws \
    --bucket velero \
    --config region=minio-primary,s3ForcePathStyle=true,s3Url=http://minio-primary.velero.svc.cluster.local:9000 \
    --default

# check "mystorage" is the default BSL
$ velero backup-location get
NAME        PROVIDER   BUCKET/PREFIX   PHASE       LAST VALIDATED                  ACCESS MODE   DEFAULT
default     aws        velero          Available   2020-12-08 09:18:15 +0800 CST   ReadWrite     
mystorage   aws        velero          Available   2020-12-08 09:18:15 +0800 CST   ReadWrite     true

# create a backup and don't specify the backup-storage-location
$ velero backup create tomystorage

# get and describe the "tomystorage" backup
$ velero backup get tomystorage
NAME          STATUS      ERRORS   WARNINGS   CREATED                         EXPIRES   STORAGE LOCATION   SELECTOR
tomystorage   Completed   0        0          2020-12-08 09:15:20 +0800 CST   29d       mystorage          <none>

$ velero backup describe tomystorage
...
Storage Location:  mystorage
...
```

carlisia (approved):
Sorry about my previous review @jenting, I thought I had updated my image but I hadn't.

This is ready to go in! Thanks a bunch for adding this.

carlisia:
@nrb you have a chance pls give this an approval again, thanks.

nrb (approved):

PR #3100, opened 2020-11-19
📖  Add docs to troubleshoot cloud-credentials

ashish-amarnath:


zubron (changes requested):
Thanks for adding this! Should these troubleshooting steps also prompt users to check the restic DaemonSet as well? I don't think they would need to have the same level of detail, just point out that if restic is enabled, they should also check the volume mounts for those pods.

ashish-amarnath:
@zubron Adding restic info is a good call! Fixed it. PTAL

zubron (approved):
Looks good - thanks for making the changes!

carlisia (approved):
👍  thank you!

PR #3101, opened 2020-11-19
Organize design docs

carlisia:
This should make it easier to identify especially what has already been implemented vs. not.


zubron:
I think I understood merged design docs to be "accepted". At what point will we review the docs under "Submitted" and move them to "Accepted"?

nrb (dismissed):
Merging is acceptance right now, and creating a PR is submitting; I'm not sure I want directories for those. Using directories means there's either right-before-merge or post-merge work to be done to get it into the right directory.

I do think an implemented directory makes sense, though. I think we should write down who moves a design doc into it and when.

carlisia:
I changed to only having an `implemented` directory. I too reached this conclusion after I walked away from it, totally agree!

carlisia:
> I do think an implemented directory makes sense, though. I think we should write down who moves a design doc into it and when.

I think this should be the person implementing the design? But I can't think right now where to write this instruction.

zubron (approved):
Thanks!

zubron:
> > I do think an implemented directory makes sense, though. I think we should write down who moves a design doc into it and when.
> 
> I think this should be the person implementing the design in the PR for the implementation? But I can't think right now where to write this instruction.

Could this be included in #3102? For example, just including that if a PR is implementing a reviewed design then the associated design doc should also be moved?

carlisia:
> Could this be included in #3102? 

Great idea, done:

https://github.com/vmware-tanzu/velero/pull/3102/files

nrb (approved):

PR #3110, opened 2020-11-24
Upgrade to Docker provided buildx action for CI

zubron:
The [previous buildx action](https://github.com/crazy-max/ghaction-docker-buildx#moved-to-docker-organization) that we were using has been archived and
users are recommended to switch to the new action provided by Docker.
The previous action also included setting up QEMU. This is now provided
as a separate action which needs to be run separately.
This change also replaces the direct use of `docker login` with the new
`login-action`. This new action also handles logging out once the build
is complete.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -22,12 +22,17 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
+    - name: Set up QEMU
+      id: qemu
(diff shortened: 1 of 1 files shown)


ashish-amarnath (approved):
These changes LGTM!
Let's keep an eye on this job when it runs on merge to follow-up with any changes that we may need to make.

nrb (approved):

PR #2865, opened 2020-08-25
add new table shortcode

a-mccarthy:
Signed-off-by: Abigail McCarthy <mabigail@vmware.com>

Adds new short code for handling tables and making it easier to add table captions to improve accessibility.

Notes:
* Passing tables through a short code means that we can't use link references in tables (see changes to supported-providers.md for what i mean). There is no easy way to pass the markdown renderer the info to make reference links work

PR #3114, opened 2020-11-26
doc: Use inline markdown links in tables

nemeth:
It seems that Hugo does not support reference-style links. See the generated doc at velero.io:
https://velero.io/docs/v1.5/api-types/volumesnapshotlocation/#parameter-reference

note: I haven't take the time yet to test it with a hugo build

### DIFF
--- a/site/content/docs/main/api-types/backupstoragelocation.md
+++ b/site/content/docs/main/api-types/backupstoragelocation.md
@@ -36,15 +36,13 @@ The configurable parameters are as follows:
 {{< table caption="Main config parameters" >}}
 | Key | Type | Default | Meaning |
 | --- | --- | --- | --- |
-| `provider` | String | Required Field | The name for whichever object storage provider will be used to store the backups. See [your object storage provider's plugin documentation][0] for the appropriate value to use. |
+| `provider` | String | Required Field | The name for whichever object storage provider will be used to store the backups. See [your object storage provider's plugin documentation](../supported-providers) for the appropriate value to use. |
 | `objectStorage` | ObjectStorageLocation | Required Field | Specification of the object storage for the given provider. |
 | `objectStorage/bucket` | String | Required Field | The storage bucket where backups are to be uploaded. |
 | `objectStorage/prefix` | String | Optional Field | The directory inside a storage bucket where backups are to be uploaded. |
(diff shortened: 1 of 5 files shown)


nemeth:
related Hugo doc: https://gohugo.io/content-management/cross-references/

zubron (changes requested):
Thanks for bringing this to our attention! As discussed in PR #2865, which added the table shortcode, we can't use reference links in shortcodes. The pages were updated to not use reference links but it looks like this is one that was missed!

Unfortunately, looking at the [preview build](https://deploy-preview-3114--velero.netlify.app/docs/main/api-types/volumesnapshotlocation/), I don't think we can use `relref` here. Instead, we can adjust our [table shortcode](https://github.com/vmware-tanzu/velero/blob/main/site/layouts/shortcodes/table.html#L8) to use use [`$.Page.RenderString`](https://gohugo.io/functions/renderstring/) instead of `markdownify` to ensure that it uses the [render-link render hook](https://github.com/vmware-tanzu/velero/blob/main/site/layouts/_default/_markup/render-link.html).  That would allow us to use inline relative markdown links (`[link text](../relative-link)`) in the table.

If you could update the table shortcode to use `RenderString`, and change the links to inline relative links that should fix it! The [same page in the 1.5 docs](https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/v1.5/api-types/volumesnapshotlocation.md) should also be updated as those are our "latest" docs.

You'll also need to [sign off any commits in this PR](https://github.com/vmware-tanzu/velero/pull/3114/checks?check_run_id=1458484952).

Thanks! :smile:

zubron:
Hi @nemeth! Are you still able to follow up on the suggested changes in this PR?

nemeth:
Thanks @zubron for the instructions!

The updated PR seems to be working with the changes you suggested:
https://deploy-preview-3114--velero.netlify.app/docs/main/api-types/backupstoragelocation/
https://deploy-preview-3114--velero.netlify.app/docs/main/api-types/volumesnapshotlocation/
https://deploy-preview-3114--velero.netlify.app/docs/v1.5/api-types/backupstoragelocation/
https://deploy-preview-3114--velero.netlify.app/docs/v1.5/api-types/volumesnapshotlocation/

carlisia (approved):
Thanks for this! 👍

zubron (approved):
Thank you for making the changes and noticing this issue!

PR #3119, opened 2020-11-30
Add Tilt configs

### LINKED ISSUE #2501: Add a Tilt setup for rapid local development
**Why:** Have one additional way to easily and quickly to build/run/test the Velero code base. Good for anyone contributing to Velero, but especially good for newcomers to the code base. Main results: 1) the setup will only consist of a few, easy items and 2) continuous code changes would require no manual interactions to deploy the code.

**How:** Set up Tilt to: rebuild binaries (triggered based on files the user specifies for Tilt to watch), send any new binary to the pod, and restart the pod after every change.

**Examples:**
Cluster API's documentation of their setup:
https://cluster-api.sigs.k8s.io/developer/tilt.html

Andy Goldstein gave a 20min demo of how the Cluster API tilt setup is done/used:
https://www.youtube.com/watch?v=EVEA5NnrucU&t=140s

carlisia:
Closes #2501. 

This is what it looks like with everything enabled. Any change to the Velero/restic server or any of the plugins is reflected in the logs within seconds:
 
image


Tiltfile
```diff
@@ -0,0 +1,190 @@
...
+"""
+
+##############################
+# Setup Velero
+##############################
+
+# Set up a local_resource build of the Velero binary. The binary is written to .tiltbuild/velero.
+local_resource(
+    "velero_manager",
+    cmd = "cd " + "." + ';mkdir -p .tiltbuild;CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags \'-extldflags "-static"\' -o .tiltbuild/velero ' + './cmd/velero',
```
nrb, line 70: Any way we can get this to use `make local`? I'm not sure I like that it's using a different build command than what we normally use, as that could introduce discrepancies.

tilt-resources/examples/backupstoragelocations.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
nrb, line 1: This is reintroducing files we explicitly removed from the repo, though in a different location.

I wonder if we could generate them at runtime via `velero install -o yaml --dry-run` or something? Maybe as part of the Tiltfile? My worry about having them in the file system is that they'll grow stale as it's a second/third place to maintain things. We also have the `config/samples` directory now, as part of the controller-runtime migration.

nrb (dismissed):

Tiltfile
carlisia: I'm sure we can, but that would defeat the purpose. In that case, we could use `make local` and not bother with Tilt. The goal here is to make the leanest possible build. Let's think about what discrepancy could be introduced that would make this not feasible. This tool is rebuilding the binary on each change. But not even the entire binary, only the parts specified in the `deps` setting. The assumption is you don't need to see anything else but the code you are working on.

carlisia: `not even the entire binary` - sorry I was up way too early today :) 

Of course if rebuilds the entire binary, but it only looks for changes in the files/directories that we specify, changes in other parts don't get included in the rebuild.

nrb: I'm not sure I follow how using `make local` here defeats the purpose of Tilt, since it's just using our Makefile arguments outside of a container. Though I do see that `make local` would need to have different arguments passed if you're working on macOS or Windows, though, since we'd always be making a Linux build.

Here's where I see possible discrepancies - if we ever need to change `go build` flags, we'd have to update them in the Tiltfile and the Makefile with this setup. That's something easy to miss unless we set up automation for it.

Since this line is about `go build`, I'd think we could use the `deps` within our `make local` target too, couldn't we? Or is it about what files Tilt watches to trigger the command run?

carlisia: Yes, "on save" of files/dirs marked to be watched Tilt gets triggered to rebuild automatically.

tilt-resources/examples/backupstoragelocations.yaml
carlisia: Good idea. Done.

Tiltfile
carlisia: Let's touch base on this, I don't think I'm understanding what you mean.

When I said using `make container` would defeat the purpose, that's because it takes over 4 min to run:
image

Vs. a few seconds with this Tilt setup:
image

At this point, the fact that the cmd gets triggered automatically on file changes is hardly a win.

Again, I might not be understanding what you are suggesting so, let's talk about it.

nrb: `make container` isn't `make local` :) The `make local` target should only produce a binary, not a container.

This is how `make local` runs on my machine:

```shell
% time make local
GOOS=darwin \
	GOARCH=amd64 \
	VERSION=main \
	PKG=github.com/vmware-tanzu/velero \
	BIN=velero \
	GIT_SHA=aa4730970037ce293a0df845381d18d172cdc6d3 \
	GIT_TREE_STATE=dirty \
	OUTPUT_DIR=$(pwd)/_output/bin/darwin/amd64 \
	./hack/build.sh
make local  51.18s user 11.23s system 467% cpu 13.339 total
```

If we just use `hack/build.sh`, (which the `make local` target invokes) with minimal arguments, then we get a much lower time:

```
% time PKG=github.com/vmware-tanzu/velero BIN=velero GOOS=linux GOARCH=amd64 VERSION=main GIT_SHA=aa4730970037ce29 GIT_TREE_STATE=dirty OUTPUT_DIR=tiltout ./hack/build.sh
PKG=github.com/vmware-tanzu/velero BIN=velero GOOS=linux GOARCH=amd64 VERSION  
2.55s user 2.07s system 788% cpu 0.586 total
```

`make local` is slower because it's looking up the git SHA each time, as well as doing some make target dependencies like ensuring the build directories exist. For these purposes, we might be able to skip the git SHA and just say that the tree is dirty so we know we're not building a "clean" binary. All `build.sh` is really doing is making sure some variables are set then invoking the `go build` command.

carlisia: My brain had read *make local* as *make container*, no wonder what you said wasn't making sense to me. Sorry about that. I see your point clearly now, let me see if I can make that work.

Tiltfile
```diff
@@ -0,0 +1,190 @@
...
+"""
+
+##############################
+# Setup Velero
+##############################
+
+# Set up a local_resource build of the Velero binary. The binary is written to .tiltbuild/velero.
+local_resource(
+    "velero_manager",
+    cmd = "cd " + "." + ';mkdir -p .tiltbuild;PKG=github.com/vmware-tanzu/velero BIN=velero GOOS=linux GOARCH=amd64 VERSION=main GIT_SHA=aa4730970037ce29 GIT_TREE_STATE=dirty OUTPUT_DIR=.tiltbuild ./hack/build.sh',
```
nrb, line 70: If this works, I think we could update the `build.sh` script to make the `GIT_SHA` optional, _or_ look up the value in the Tiltfile once at the start.

That way, the value's not hardcoded to something that may confuse a dev looking at their binary someday and wondering why the built SHA doesn't match what they're actually building.

carlisia: This worked. I have to do the same with the plugin bit.

Ok, I'll changed the build.sh. 

Should `PKG` be set to `github.com/vmware-tanzu/velero` or to where the local src code is?

carlisia: I'm pointing the PKG variable to local code. Let me know if it should be different.

carlisia:
Ready for review.

carlisia:
Made some improvements:
- changed `.tiltbuild` to `_tiltbuild`
- added a separate local build to build a binary for the user's machine (this way we don't need to run `make local` when making changes to the CLI
- added a config to pass in the user's local goos
- renamed the Tilt resources to be more meaningful and shorter
- added a tip to the documentation to create an alias to the user's local binary included above

carlisia:
A note: for the plugins, I am also now making a call to the plugin `hack/build.sh`: https://github.com/vmware-tanzu/velero/pull/3119/files#diff-c2ee8653e1d6b85f0aadf87cd438a9250806c052877248442be4d434cbc52425R178

However, plugins maintained by others might not have this build setup. Should I setup a call to `go build` inline for this case?

Other than this, it's ready for review.

nrb:
> However, plugins maintained by others might not have this build setup and, if they don't, this will fail. Should I change to call go build inline for this case?

We could, though it's worth asking if we're making this universal, or for core development. Another option is to say that if you want to use the Tiltfile for development the hack/build.sh entrypoint is necessary, which may not be so bad anyway - consider that even if we do plain `go build`, each plugin might have their own settings/flags that they want to introduce at build time. Maybe they really want CGO turned on. By using a script name vs a bare command in the Tiltfile, they would have a way to include those settings without editing the Tiltfile.

carlisia:
> consider that even if we do plain go build, each plugin might have their own settings/flags that they want to introduce at build time

This is a winning argument. And it is also not bad at all since 1, we have examples and 2, it's such a simple and small file.

So, let's 🚢  this!

Tiltfile
```diff
@@ -0,0 +1,201 @@
...
+    k8s_yaml('tilt-resources/restic.yaml')
+if settings.get("create_backup_locations"):
+    k8s_yaml('tilt-resources/velero_v1_backupstoragelocation.yaml')
+if settings.get("setup-minio"):
+    k8s_yaml('examples/minio/00-minio-deployment.yaml', allow_duplicates=True)
+
+# By default, Tilt automatically allows Minikube, Docker for Desktop, Microk8s, Red Hat CodeReady Containers, Kind, K3D, and Krucible.
+allow_k8s_contexts(settings.get("allowed_contexts"))
+default_registry(settings.get("default_registry"))
+local_goos = settings.get("local_goos", )
```
zubron, line 44: I think we could use [`local`](https://docs.tilt.dev/api.html#api.local) here to call `go env GOOS` which would eliminate the need to explicitly state the value in settings. I tried out the following and it worked for me on my linux box:
`local_goos = str(local("go env GOOS", quiet=True, echo_off=True)).strip()`

Tiltfile
```diff
@@ -0,0 +1,201 @@
...
+    for name in user_enable_providers:
+        enable_provider(name)
+
+# Configures a provider by doing the following:
+#
+# 1. Enables a local_resource go build of the provider's local binary
+# 2. Configures a docker build for the provider, with live updating of the local binary
+def enable_provider(name):
+    p = providers.get(name)
+    plugin_name = p.get("plugin_name")
```
zubron, line 148: Hmm, I'm getting an error when I run this locally with https://github.com/vmware-tanzu/velero-plugin-for-aws/pull/66 checked out as `p.get("plugin_name")` returns `None`. At the point of calling this, the contents of `p` are `{"context": "../velero-plugin-for-aws/.", "image": "velero/velero-plugin-for-aws", "live_reload_deps": ["velero-plugin-for-aws"], "go_main": "./velero-plugin-for-aws"}`. Given that it's being copied into the container, should we be referencing the `go_main` value here?

zubron (changes requested):
This is really awesome so far, @carlisia! I'm still working on a review but I have some questions/comments in the meantime :)

Tiltfile
carlisia: The plugin repo is missing an update to the settings file that adds "plugin_name"! Let me push those and maybe also we can merge them first.

site/content/docs/main/tilt.md
```diff
@@ -0,0 +1,176 @@
...
+    "create_backup_locations": true,
+    "setup-minio": false
+}
+```
+
+#### tilt-settings.json fields
+**default_registry** (String, default=""): The image registry to use if you need to push images. See the [Tilt
+*documentation](https://docs.tilt.dev/api.html#api.default_registry) for more details.
+
+**provider_repos** (Array[]String, default=[]): A list of paths to all the provider plugins you want to make changes to. Each provider must have a
```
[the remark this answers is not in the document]
zubron: I'm leaving this comment here as a note for others that we should get the other PRs reviewed and merged first :)

go.sum
```diff
@@ -49,7 +49,6 @@ github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym
 github.com/GoogleCloudPlatform/k8s-cloud-provider v0.0.0-20190822182118-27a4ced34534/go.mod h1:iroGtC8B3tQiqtds1l+mgk/BBOrxbqjH+eUfFQYRc14=
 github.com/JeffAshton/win_pdh v0.0.0-20161109143554-76bb4ee9f0ab/go.mod h1:3VYc5hodBMJ5+l/7J4xAyMeuM2PNuepvHlGs8yilUCA=
 github.com/MakeNowJust/heredoc v0.0.0-20170808103936-bb23615498cd/go.mod h1:64YHyfSL2R96J44Nlwm39UHepQbyR5q10x7iYa1ks2E=
-github.com/MakeNowJust/heredoc v1.0.0 h1:cXCdzVdstXyiTqTvfqk9SDHpKNjxuom+DOlyEeQ4pzQ=
 github.com/MakeNowJust/heredoc v1.0.0/go.mod h1:mG5amYoWBHf8vpLOuehzbGGw0EHxpZZ6lCpQ4fNJ8LE=
```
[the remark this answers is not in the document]
carlisia: Frankly, I don't know. I can try pasting the version in main.

Tiltfile
carlisia: Done ✅

Tiltfile
```diff
@@ -0,0 +1,201 @@
...
+def load_provider_tiltfiles():
+    provider_repos = settings.get("provider_repos", [])
+
+    for repo in provider_repos:
+        file = repo + "/tilt-provider.json"
+        provider_details = read_json(file, default = {})
+        if type(provider_details) != type([]):
+            provider_details = [provider_details]
+        for item in provider_details:
+            provider_name = item["name"]
```
nrb, line 126: When the `tilt-provider.json` file is missing, Tilt will report an error like this:

```
Beginning Tiltfile execution
Falling back to `kubectl kustomize` since `kustomize` was not found in PATH
Successfully loaded Tiltfile (172.487237ms)
Traceback (most recent call last):
  /Users/nrb/go/src/github.com/vmware-tanzu/velero/Tiltfile:199:24: in <toplevel>
  /Users/nrb/go/src/github.com/vmware-tanzu/velero/Tiltfile:126:33: in load_provider_tiltfiles
Error: key "name" not in dict
```

I'd recommend adding the following after L21:

```python
        if not os.path.exists(file):
            print("No provider.json found for %s", repo)
            continue
```

That will then lead to this error:

```
Beginning Tiltfile execution
Falling back to `kubectl kustomize` since `kustomize` was not found in PATH
No provider.json found for %s ../velero-plugin-for-gcp
Successfully loaded Tiltfile (74.249484ms)
Traceback (most recent call last):
  /Users/nrb/go/src/github.com/vmware-tanzu/velero/Tiltfile:204:17: in <toplevel>
  /Users/nrb/go/src/github.com/vmware-tanzu/velero/Tiltfile:143:24: in enable_providers
  /Users/nrb/go/src/github.com/vmware-tanzu/velero/Tiltfile:151:20: in enable_provider
Error: NoneType has no .get field or method
```

I have a fix for it, but I'll add it inline where it goes.

Tiltfile
```diff
@@ -0,0 +1,201 @@
...
+    user_enable_providers = settings.get("enable_providers", [])
+    for name in user_enable_providers:
+        enable_provider(name)
+
+# Configures a provider by doing the following:
+#
+# 1. Enables a local_resource go build of the provider's local binary
+# 2. Configures a docker build for the provider, with live updating of the local binary
+def enable_provider(name):
+    p = providers.get(name)
```
nrb, line 147: To fix this error in case `providers` is empty

```
Beginning Tiltfile execution
Falling back to `kubectl kustomize` since `kustomize` was not found in PATH
No provider.json found for %s ../velero-plugin-for-gcp
Successfully loaded Tiltfile (74.249484ms)
Traceback (most recent call last):
  /Users/nrb/go/src/github.com/vmware-tanzu/velero/Tiltfile:204:17: in <toplevel>
  /Users/nrb/go/src/github.com/vmware-tanzu/velero/Tiltfile:143:24: in enable_providers
  /Users/nrb/go/src/github.com/vmware-tanzu/velero/Tiltfile:151:20: in enable_provider
Error: NoneType has no .get field or method
```

You can do this:

```python
    if not providers:
        print("No providers to enable. Please ensure your repositories have a tilt-provider.json file included.")
        return
```

Of course, the environment won't load, but it will return a more helpful message.

site/content/docs/main/tilt.md
```diff
@@ -0,0 +1,180 @@
...
+1. [Docker](https://docs.docker.com/install/) v19.03 or newer
+1. A Kubernetes cluster (does not have to be Kind)
+1. [Tilt](https://docs.tilt.dev/install.html) v0.12.0 or newer
+1. Clone the [Velero project](https://github.com/vmware-tanzu/velero) repository
+   locally
+2. Clone any [provider plugin(s)](https://velero.io/plugins/) you want to make changes to and deploy (optional, must be configured to be deployed by the Velero Tilt's setup, [more info below](#provider-plugins))
+
+## Getting started
+
+### tl;dr
```
nrb, line 30: The tl;dr is missing setting up the `tilt-provider.json` files.

carlisia: The tilt-provider.json for the providers listed in the sample file are already all setup (after everything is merged), nothing to do there.

nrb: Doh, I missed those. Thanks!

carlisia:
Latest changes:

Instead of in the settings file having separate lists for repos and providers to enable, I created a map of  repos per provider name. This is reflected in the sample file. This made it a lot easier to, in the Tilt file, validate if a provider set to be enabled exists in the file with a corresponding name and repo.

Tilt files now has validations that a setting file exists in the repo of a provider included in the enable list; if not it warns and continues to the next on the list. It also validates if there are no providers at all to enable, and warns when there's none. This could be because there was no provider listed to enable, or none of the providers listed to enable had a settings file found. In the first case it is not a big deal as people might just want to work on core Velero, but the warning is nice in this case and needed in the second case. Thanks @nrb for thinking about this and providing the code for it!

go.sum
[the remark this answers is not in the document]
carlisia: Sorta fixed it.

carlisia:
Ready for review.

zubron (changes requested):
Thanks for updating the plugin configs! I've got a few more comments. I've divided them up into things that need to be addressed in this PR before merging and others (starting with "Suggestion") that I think would be fine to do as a follow up PR.

Tiltfile
```diff
@@ -0,0 +1,203 @@
...
+
+providers = {}
+
+def load_provider_tiltfiles():
+    all_providers = settings.get("providers", []) 
+    enable_providers = settings.get("enable_providers", [])
+
+    ## Load settings only for providers to enable
+    for name in enable_providers:
+        repo = all_providers[name]
```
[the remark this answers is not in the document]
carlisia: Thanks for the code!

Tiltfile
```diff
@@ -0,0 +1,203 @@
...
+            if "go_main" not in provider_details:
+                provider_details["go_main"] = "main.go"
+            providers[name] = provider_details
+
+# Enable each provider
+def enable_providers():
+    if not providers:
+        print("No providers to enable.")
+        return
+    for name in providers:
```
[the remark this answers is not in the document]
carlisia: It depends on what we want to do. I'd like it for the developer to be able to choose to load from 0.* providers as a Tilt resource. Right now this is done in the `enable_providers` list in the tilt-settings.json. I'm not seeing how this would be possible with this suggestion. Please clarify this part if I missed it.

Tiltfile
```diff
@@ -0,0 +1,206 @@
...
+"""
+
+##############################
+# Setup Velero
+##############################
+
+# Set up a local_resource build of the Velero binary. The binary is written to _tiltbuild/velero.
+local_resource(
+    "velero_server_binary",
+    cmd = 'cd ' + '.' + ';mkdir -p _tiltbuild;PKG=. BIN=velero GOOS=linux GOARCH=amd64 VERSION=main GIT_TREE_STATE=dirty OUTPUT_DIR=_tiltbuild ./hack/build.sh',
```
ashish-amarnath, line 71: this looks like it is running the same command run by a make target.
Any reason not to call the make target directly?

Tiltfile
```diff
@@ -0,0 +1,206 @@
...
+    deps = ["cmd", "internal", "pkg"],
+    ignore = ["pkg/cmd"],
+)
+
+local_resource(
+    "velero_local_binary",
+    cmd = 'cd ' + '.' + ';mkdir -p _tiltbuild/local;PKG=. BIN=velero GOOS=' + local_goos + ' GOARCH=amd64 VERSION=main GIT_TREE_STATE=dirty OUTPUT_DIR=_tiltbuild/local ./hack/build.sh',
```
ashish-amarnath, line 78: same comment here:
this looks like it is running the same command run by a make target.
Any reason not to call the make target directly?

site/content/docs/main/tilt.md
```diff
@@ -0,0 +1,176 @@
...
+
+Here is an example:
+
+```json
+{
+    "default_registry": "",
+    "enable_providers": [
+        "aws",
+        "gcp",
+        "azure"
```
ashish-amarnath, line 49: the csi plugin also needs to be added to the list.

site/content/docs/main/tilt.md
```diff
@@ -0,0 +1,176 @@
...
+    ],
+    "providers": { 
+        "aws": "../velero-plugin-for-aws",
+        "gcp": "../velero-plugin-for-gcp",
+        "azure": "../velero-plugin-for-microsoft-azure"
```
ashish-amarnath, line 54: csi plugin needs to be added here as well.

site/content/docs/main/tilt.md
```diff
@@ -0,0 +1,176 @@
...
+  "context": ".",
+  "image": "velero/velero-plugin-for-aws",
+  "live_reload_deps": [
+    "velero-plugin-for-aws"
+  ],
+  "go_main": "./velero-plugin-for-aws"
+}
+```
+
+## Live updates
```
ashish-amarnath, line 171: Will this also watch and do a live reload of the plugin images?

tilt-resources/examples/deployment.yaml
```diff
@@ -0,0 +1,125 @@
...
+        prometheus.io/scrape: "true"
+      labels:
+        component: velero
+        deploy: velero
+    spec:
+      containers:
+        - args:
+            - server 
+            - --log-level 
+            - debug
```
ashish-amarnath, line 28: This doesn't have to be `debug` by default.

tilt-resources/examples/deployment.yaml
```diff
@@ -0,0 +1,125 @@
...
+            name: plugins
+        - image: velero/velero-plugin-for-gcp
+          imagePullPolicy: Always
+          name: velero-plugin-for-gcp
+          volumeMounts:
+          - mountPath: /target
+            name: plugins
+        - image: velero/velero-plugin-for-microsoft-azure
+          imagePullPolicy: Always
+          name: velero-plugin-for-microsoft-azure
```
ashish-amarnath, line 81: csi plugin needs to be added here as well.

ashish-amarnath: I haven't watched the demo yet. But should all the plugins be added here?

site/content/docs/main/tilt.md
ashish-amarnath: How about updates to the restic-restore-helper image?

Tiltfile
ashish-amarnath: we should also be building the `restic-restore-helper`.

Tiltfile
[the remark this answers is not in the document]
zubron: Sorry, I don't think I did great job explaining what I meant. Let me try again! :)

At this point when we are iterating over `providers`, it is a dict with `name` as the key and will only contain providers that the user has enabled in their Tilt settings since we filtered them out in `load_provider_tiltfiles`, e.g.
```
{
  "aws": {
    "plugin_name": "velero-plugin-for-aws",
    ...
  }
}
```

Instead, I am proposing that we modify `load_provider_tiltfiles` so that this `providers` variable is a list. e.g. in `load_provider_tiltfiles` we could have:
```
providers = []

for name in enable_providers:
  ...
  providers.append(provider_details)
```

That would mean that the `providers` variable would now look like: `[{"plugin_name": "velero-plugin-for-aws", ...}]`. Then when we iterate over it, we could pass that whole object into `enable_providers`:
```
for provider in providers:
  enable_provider(provider)
```
Then in `enable_provider` on line 154, we don't need to call `providers.get(name)` as the parameter to the function _is_ the provider. Does that make sense? As I mentioned in my original comment though, this is only a suggestion just to try and limit the scope where the global variable `providers` needed to be accessed.

hack/build.sh
```diff
@@ -41,11 +41,6 @@ if [[ -z "${VERSION}" ]]; then
     exit 1
 fi
 
-if [[ -z "${GIT_SHA}" ]]; then
```
ashish-amarnath, line 44: `GIT_SHA` was required to set the version fields in the `buildinfo` struct.
Now by removing this check, we are making this optional and setting the `buildinfo` field only if it is supplied.
Which IMO is not correct.
Changes to this file can be reverted.

tilt-resources/examples/tilt-settings.json
```diff
@@ -0,0 +1,19 @@
+{
+    "default_registry": "",
+    "enable_providers": [
+        "aws",
+        "gcp",
+        "azure"
```
ashish-amarnath, line 6: csi plugin needs to be added here as well.

tilt-resources/examples/tilt-settings.json
```diff
@@ -0,0 +1,19 @@
...
+    ],
+    "providers": { 
+        "aws": "../velero-plugin-for-aws",
+        "gcp": "../velero-plugin-for-gcp",
+        "azure": "../velero-plugin-for-microsoft-azure"
```
ashish-amarnath, line 11: csi plugin needs to be added here as well.

site/content/docs/main/tilt.md
```diff
@@ -0,0 +1,176 @@
...
+1. Deploying the necessary Kubernetes resources, such as the Velero CRDs and Velero deployment
+1. Building a local binary for Velero and (if specified) provider plugins as a `local_resource`
+1. Invoking `docker_build` to live update any binary into the container/init container and trigger a re-start
+
+Tilt will look for configuration files under `velero/tilt-resources`. Most of the
+files in this directory are gitignored so you may configure your setup according to your needs.
+
+## Prerequisites
+1. [Docker](https://docs.docker.com/install/) v19.03 or newer
+1. A Kubernetes cluster (does not have to be Kind)
```
ashish-amarnath, line 22: suggest calling out that we support only kubernetes v1.10 and beyond.
Also, the volume provisioners and drivers will not be installed by this setup.

A few notes from my e2e work.

ashish-amarnath (changes requested):
I added a few comments.

Tiltfile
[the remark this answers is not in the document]
carlisia: Thanks for the expanded explanation, I needed it obviously. It's a better approach, changed it.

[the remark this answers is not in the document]
zubron: Thanks for making these changes! I've been stung by global variables getting modified in ways I don't expect and it being a real pain to figure out what's going on :weary:

tilt-resources/examples/deployment.yaml
carlisia: This is a sample file that people might choose to use as is, or delete things they don't want to run. I'm opting to offer more for the sake of better discovery, especially for new people onboarding.

tilt-resources/examples/deployment.yaml
carlisia: No, but I'm opting to offer more for the sake of better discovery, especially for new people onboarding.

zubron: I think it's better to leave debug enabled. Given that this is a workflow for development, having the debug output is useful to have by default.

Tiltfile
zubron: We could do this now but the code in this image looks like it hasn't changed since it was added with the initial restic support so it doesn't seem like it would be needed as urgently as the images already included in this change.

Also, I don't quite know what it means to update images used as an init container like this one. Would it just be a matter of rebuilding the image and tilt would know to use that image when it used?

carlisia: It's not obvious to me what to do with this and I have to figure this out but would like not to do it in this PR. I opened an issue to track it: https://github.com/vmware-tanzu/velero/issues/3132.

site/content/docs/main/tilt.md
carlisia: > Will this also watch and do a live reload of the plugin images?

This is what happens with the plugins, as documented:
> When a dependency is modified, Tilt rebuilds the provider's binary **on your local
machine**, copies the binary to the init container, and triggers a restart of the Velero container.

carlisia: > How about updates to the restic-restore-helper image?

Going to address this not in this PR.

site/content/docs/main/tilt.md
carlisia: Done ✅

Tiltfile
carlisia: Our make target has some hard-coded values, for example, the "output" directory. I think it's best to output this binary to "_tiltbuild".

hack/build.sh
carlisia: Reverted ✅

carlisia:
This is ready for review again.

site/content/docs/main/tilt.md
```diff
@@ -0,0 +1,180 @@
...
+storage location. A sample file is provided that needs to be modified with the specific
+configuration for your object storage. See the next sub-section for more details on this.
+
+### Configure a backup storage location
+You will have to configure the `backupstoragelocation.yaml` with the proper values according to your storage provider. Read the [plugin documentation](https://velero.io/plugins/)
+to learn what field/value pairs are required for your particular provider's backup storage location configuration.
+
+Below are some ways to configure a backup storage location for Velero.
+#### As a storage with a service provider
+Follow the provider documentation to provision the storage. We have a [list of all known object storage providers](https://velero.io/docs/main/supported-providers/) with corresponding plugins for Velero.
```
zubron, line 109: These should be relative links rather than links including the branch name. I noticed that the docs for 1.5 include `main` in the link. There are a few more links that should be updated on lines 130, 136, 158.

site/content/docs/main/tilt.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+#### tilt-settings.json fields
+**default_registry** (String, default=""): The image registry to use if you need to push images. See the [Tilt
+*documentation](https://docs.tilt.dev/api.html#api.default_registry) for more details.
+
+**provider_repos** (Array[]String, default=[]): A list of paths to all the provider plugins you want to make changes to. Each provider must have a
+`tilt-provider.json` file describing how to build the provider. 
+
+**enable_providers** (Array[]String, default=[]): A list of the provider plugins to enable. See [provider plugins](#provider-plugins)
```
zubron, line 76: The `#` should be removed in this link.

zubron (approved):
Woo! Thank you *so* much for this, @carlisia!! This is an awesome change and I'm looking forward to using it :star_struck:

ashish-amarnath (approved):
LGTM! 🚢

zubron (approved):
:partying_face:

PR #3120, opened 2020-11-30
Fix documentation about secrets

carlisia:


nrb:
Per a discussion in Slack, this isn't quite accurate. Velero can support one set of credentials per provider, but the YAML must be updated to point to unique file mounts per secret. The current default is to mount them all as a single file, but we don't have an easy way to change that - it's currently a `kubectl edit -n velero deploy/velero` task.

PR #3133, opened 2020-12-04
Restore API group version by priority

### LINKED ISSUE #2551: Restore API Groups based on Target Cluster Preferred/Supported Version

codegold79:
Closes #2551.

This PR adds the ability to choose which backed up API group version of a resource to restore. See the corresponding design doc (PR #3050) for more information. The ability to choose which version to restore is the second part of being able to restore onto a Kubernetes cluster that is more than one version ahead of the backed up cluster (Issue #2251). The first half, where backing up multiple versions was enabled with a feature flag, was completed with PR #2373.

These are the changes made in this PR:
- modified pkg/restore/restore.go to add to the restore context the chosen API Group version to restore for every resource being restored. It will do a check that ensures the` EnableAPIGroupVersions` feature flag was enabled and that the backed up objects use format version 1.1.0 before proceeding with choosing a version to restore. If those requirements are met, the path to the location of the json backup file for the resource will be updated to point to the chosen version to restore.
- modified pkg/archive/parser.go in order that API Group versions can be extracted from backed up directory names
- added unit tests to pkg/archive/parser_test.go
- added pkg/restore/prioritize_group_version.go which is a collection of functions that collects API Group versions supported by the source cluster, target cluster, as well as the API Version that a user adds via a config map to define user prioritization of versions to restore. This group of functions also selects the API Group version to restore based on user priority versions or a predetermined version priority hierarchy.
- added unit tests  to pkg/restore/prioritize_group_version_test.go

End-to-end tests have been performed to ensure the feature works as intended. The tests followed the cases outlined in by @brito-rafa's at https://github.com/brito-rafa/k8s-webhooks/tree/master/examples-for-projectvelero.

The general procedure for the tests follow the scripts provided in @brito-rafa's [repo](https://github.com/brito-rafa/k8s-webhooks/tree/master/examples-for-projectvelero). 
- Create a kind cluster called velsrc. 
- Install the cert-manager and rockband CRDs. 
- Install Velero using the latest image stored in Corporate Harbor (version 0.0.4 at the time of this writing),
  `--image projects.registry.vmware.com/tanzu_migrator/velero-pr3050:latest`
- Install the CRs for the beatles and/or ledzeppelin rockbands. To easier see the transformations between versions, each version was created in their respective namespaces. For example, the v2beta1 version for a beatles CR was installed in the rockbands-v2beta1 namespace.
- Create a backup excluding the velero namespace.
- Create a kind cluster called veltar.
- Install the cert-manager and rockband CRDs.
- Install Velero using the latest image.
- Optional: install a config map listing user defined API group version prioritization.
- Restore the backup created earlier.
- From the logs, you can verify which version was selected to restore.
- You can also inspect the annotations and spec fields of the beatles and/or ledzeppelin CRs in each of the namespaces to find hints at to which version was used to restore the CRs.


carlisia:
Hi @codegold79, thanks for this! I'm intending to start reviewing this today. I haven't looked at it closely at all but there's some test failure, would you please look into that when you have a chance? Thank you.

codegold79:
> ...there's some test failure, would you please look into that when you have a chance?

I'll definitely take a look--thanks for catching that @carlisia.

carlisia:
> Another alternative is to take advantage of the new e2e test framework

I think it's best to add those specific tests to the e2e suite of tests.

nrb:
> Another alternative is to take advantage of the new e2e test framework 

Yeah, I think that's a good idea. At some point, I would like to get the e2e tests run on a scheduled basis against a KinD cluster, though maybe not against every PR.

codegold79:
Thanks @carlisia and @nrb for your feedback. I'll move some of the failing unit tests to the e2e framework. 

I have put the PR into `WIP` while I work on the transition. I posted incomplete tests up because I couldn't get the tests to work on my local machine. I'll remove the `WIP` from the PR title once I get the test complete.

codegold79:
Thanks to @dsu-igeek and @brito-rafa  I was able to get the e2e_testify tests to work and not cause CI Check errors. Ready again for review.

While waiting for reviews, I'll keep making tweaks to the e2e tests and maybe even adding more tests. I'll be sure to have the changes in new commits as not to interrupt the review process.

pkg/archive/parser.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 the Velero contributors.
+Copyright 2020 the Velero contributors.
```
nrb, line 2: The copyright year can be removed entirely, that should help in keeping these up-to-date.

codegold79: Removing the year sounds much more maintainable. I'll be sure to change the lines to be

`Copyright the Velero contributors.`

pkg/archive/parser_test.go
```diff
@@ -110,11 +111,161 @@ func TestParse(t *testing.T) {
...
 			}
 		})
 	}
 }
+
+func TestParseGroupVersions(t *testing.T) {
+	tests := []struct {
+		name    string
+		files   []string
+		buDir   string
```
nrb, line 127: What does `bu` stand for here?

pkg/install/install.go
```diff
@@ -132,7 +132,7 @@ func DeploymentIsReady(factory client.DynamicFactory, namespace string) (bool, e
 	// declare this variable out of scope so we can return it
 	var isReady bool
 	var readyObservations int32
-	err = wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
+	err = wait.PollImmediate(time.Second, 3*time.Minute, func() (bool, error) {
```
nrb, line 135: Was this increase because the 1 minute wait wasn't long enough to get everything installed? I think this should be called out as a separate line in the release notes.

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,369 @@
...
+	"github.com/vmware-tanzu/velero/pkg/archive"
+	"github.com/vmware-tanzu/velero/pkg/client"
+	"github.com/vmware-tanzu/velero/pkg/features"
+)
+
+// meetsAPIGVRestoreReqs determines if the feature flag has been
+// enabled and the backup objects use Status.FormatVersion 1.1.0.
+func (ctx *restoreContext) meetsAPIGVRestoreReqs() bool {
+	return features.IsEnabled(velerov1api.APIGroupVersionsFeatureFlag) &&
+		ctx.backup.Status.FormatVersion == "1.1.0"
```
nrb, line 40: Should we be strictly equal, or >=?

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,369 @@
...
+			)
+			ctx.log.Debugf(
+				"APIGroupVersionsFeatureFlag Priority 2: Cluster preferred API group version not found in backup. Using backup preferred version %s for %s",
+				sg.PreferredVersion.Version,
+				rg,
+			)
+			continue
+		}
+
+		// Priority 3: Kubernetes Prioritized Common Supported Version
```
nrb, line 120: Could we reword the comment to say "kubernetes preferred"? I think that's more in line with what the code says, and I had to stop for a minute to figure out what "kubernetes prioritzed" meant.

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,369 @@
...
+		return nil, errors.Wrap(err, "getting enableapigroupversions config map from velero namespace")
+	}
+
+	return cm, nil
+}
+
+func parseUserPriorities(prioritiesData string) map[string]metav1.APIGroup {
+	ups := make(map[string]metav1.APIGroup)
+
+	// The user priorities will be in a string of the form
```
nrb, line 237: Thanks for inline examples!

pkg/restore/prioritize_group_version_test.go
```diff
@@ -0,0 +1,465 @@
...
+				},
+				{
+					Name: "",
+					Versions: []metav1.GroupVersionForDiscovery{
+						{Version: "v1"},
+					},
+					PreferredVersion: metav1.GroupVersionForDiscovery{Version: "v1"},
+				},
+				{
+					Name: "velero.vmware.com",
```
nrb, line 233: I know these are sample values, but let's use `velero.io` :)

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,369 @@
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// latestCommon will find the versions that are in common between the source and
+// target clusters. It will return the latest of the common versions.
+func latestCommon(sgvs, tgvs []metav1.GroupVersionForDiscovery) string {
```
nrb, line 334: The fact that alpha and beta versions will be chosen as latest in the overall functionality should be called out in the documentation.

test/e2e_testify/Makefile
```diff
@@ -0,0 +1,73 @@
+# Copyright 2021 The Kubernetes Authors.
+#
```
nrb, line 2: Please include a modification notice as we do [in the main Makefile](https://github.com/vmware-tanzu/velero/blob/main/Makefile#L3)

nrb (commented):
Thanks so much for all of this! I've left some comments and questions on the code.

I have a question on the testify e2e changes - is there any way the framework/scaffolding parts could be split into a separate PR? They're not strictly related to the described change, and I think it would be easier to include that group of changes on their own. I have not looked too deeply into these files yet, but I know the test case files wouldn't make sense to move.

pkg/archive/parser_test.go
codegold79: It stands for `backup`. I'll change that to `backupDir` because `buDir` is too unclear.

pkg/install/install.go
codegold79: Remember when we had that Zoom session in December and we thought the timeout problem was happening because of Docker? And the timeout would happen on your laptop but not your server? I believe this timeout value was the problem.

Increasing the time doesn't mean it will necessarily take longer. The polling happens every second (the first argument), up to the timeout time (second argument). However, if the desired conditions aren't met by 1 minute, it will wait up to 3 minutes.

I'll go ahead and add the timeout change to the release notes.

codegold79:
@nrb wrote,
>  I have a question on the testify e2e changes - is there any way the framework/scaffolding parts could be split into a separate PR? They're not strictly related to the described change, and I think it would be easier to include that group of changes on their own. I have not looked too deeply into these files yet, but I know the test case files wouldn't make sense to move.

Sure, I can move the common_test.go functions and testify scaffold to another PR. And keep the test cases for the restore package changes here in this PR.

pkg/restore/prioritize_group_version.go
codegold79: I had no idea Go was able to sort versions like this in strings the way I expected. Pretty neat. Yes, because Go can do that, `>=` works fantastic.

pkg/restore/prioritize_group_version.go
codegold79: I think "Kubernetes preferred" will be misleading. Kubernetes preferred is chosen specifically by Kubernetes as the latest, most stable API group version of a CRD. For example, for HPA, there are three versions: v1, v2alpha1, v2beta1. Kubernetes preferred will always be v1 given those three versions. 

By the time we get to this part of the code, we have excluded the Kubernetes preferred versions as an option because both source/target clusters didn't support it. We now need to go down the k8s prioritized versions and use the first version that is in common in both target and source clusters.

Kubernetes "prioritized" isn't a specific version (like preferred is) and instead it is an ordering of the versions that remain. The versions are sorted by priority. More explanation here: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority

I agree that the comment does not make sense. Maybe I can change it to

```// Priority 3: The Common Supported Version with the Highest Kubernetes Version Priority```

Would that be better?

File: pkg/restore/prioritize_group_version_test.go (hunk 7 above)
Comment (codegold79) [reply pkg/restore/prioritize_group_version_test.go:233]: Will do.

File: pkg/restore/prioritize_group_version.go (hunk 8 above)
Comment (codegold79) [reply pkg/restore/prioritize_group_version.go:334]: That function and associated unit test need to be removed since I don't use alphabetical order any more to sort the versions. I use Kubernetes priority versions now.

File: test/e2e_testify/Makefile (hunk 9 above)
Comment (codegold79) [reply test/e2e_testify/Makefile:2]: Will do.

File: pkg/restore/prioritize_group_version.go [hunk 10]
```diff
@@ -0,0 +1,346 @@
...
+	}
+
+OUTER:
+	for rg, sg := range sgvs {
+		// Default to the source preferred version if no other common version
+		// can be found.
+		cgv := ChosenGroupVersion{
+			Group:   sg.Name,
+			Version: sg.PreferredVersion.Version,
+			Dir:     sg.PreferredVersion.Version + "-preferredversion",
```
brito-rafa, line 67: We should use the constant PreferredVersionDir

brito-rafa, line 67: We should use the constant PreferredVersionDir

pkg/archive/parser.go
```diff
@@ -165,3 +167,102 @@ func (p *Parser) getResourceItemsForScope(dir, archiveRootDir string) ([]string,
...
+			if gvdName == "namespaces" || gvdName == "cluster" {
+				continue
+			}
+
+			version := metav1.GroupVersionForDiscovery{
+				GroupVersion: strings.TrimPrefix(group.Name+"/"+gvdName, "/"),
+				Version:      gvdName,
+			}
+
+			if strings.Contains(gvdName, "-preferredversion") {
```
brito-rafa, line 235: We should use the constant PreferredVersionDir

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,346 @@
...
+type ChosenGroupVersion struct {
+	Group   string
+	Version string
+	Dir     string
+}
+
+// chooseAPIVersionsToRestore will choose a version to restore based on a user-
+// provided config map prioritization or our version prioritization.
+func (ctx *restoreContext) chooseAPIVersionsToRestore() error {
+	sgvs, tgvs, ugvs, err := ctx.gatherSTUVersions()
```
brito-rafa, line 55: Can we add a comment to describe the meaning of sgvs, tgvs, ugvs. Same for the "STUVersions"

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,346 @@
...
+
+	// Sort the versions in the APIGroups in sourceRGVersions map values.
+	for _, src := range sourceRGVersions {
+		k8sPrioritySort(src.Versions)
+	}
+
+	targetGroupVersions := ctx.discoveryHelper.APIGroups()
+
+	// Sort the versions in the APIGroups slice in targetGroupVersions.
+	for _, tar := range targetGroupVersions {
```
brito-rafa, line 165: When I saw the variable named "tar", I thought it was from the tarball. suggest renaming/expanding the same of the variable to "target" or something more descriptive.

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,346 @@
...
+			Version: v,
+		}
+	}
+
+	return gvs
+}
+
+// findTargetGroup looks for a particular group in from a list of groups in the
+// target cluster.
+func findTargetGroup(tgvs []metav1.APIGroup, group string) metav1.APIGroup {
```
brito-rafa, line 299: I wonder if we can make this function with a more generic name - not only pinning its use for "target".  looking up a group string in a metav1.APIGroup can be used in other circumstances, or no?

pkg/restore/restore.go
```diff
@@ -732,6 +742,15 @@ func (ctx *restoreContext) restoreResource(resource, targetNamespace, originalNa
 
 	groupResource := schema.ParseGroupResource(resource)
 
+	// Modify `resource` so that it has the priority version to restore in its
+	// path. For example, for group resource "horizontalpodautoscalers.autoscaling",
+	// "/v2beta1" will be appended to the end. Different versions would only
+	// have been stored if the APIGroupVersionsFeatureFlag was enabled during backup.
+	cgv, ok := ctx.chosenGrpVersToRestore[groupResource.String()]
```
brito-rafa, line 749: I would add more comments here. the `chosenGrpVersToRestore` would exist if meetsAPIGVRestoreReqs as well. not only if the backup was taken with featureflag.

pkg/restore/prioritize_group_version.go
codegold79: Good catch. I'll replace these with `velerov1api.PreferredVersionDir`.

pkg/archive/parser.go
codegold79: Will do.

pkg/restore/prioritize_group_version.go
codegold79: As per our meeting, I will not add a comment as there is already a comment for the function. It would be redundant and cumbersome to keep updated. I will make the line clearer by writing out the abbreviations. The line will change to 
```
sourceGVs, targetGVs, userGVs, err := ctx.gatherSourceTargetUserGroupVersions()
```
This change makes copying of the function's comments not necessary as I believe it is now clear what the function does and what the variables represent.

pkg/restore/prioritize_group_version.go
codegold79: Will do.

pkg/restore/prioritize_group_version.go
codegold79: Yes, I can make it more generic, but it cannot be used for the Source API Groups and User API Groups. The reason is because Target API Groups are a slice of API Groups while the other two are maps where resource groups are the keys and the API Groups are the values. But, not being able to accept both maps and keys is not a problem as developers can read function signatures.

I'll make it more generic and call the function `containsAPIGroup(...)`.

pkg/restore/restore.go
codegold79: I will add a sentence to the comment:
```
// The chosenGrpVersToRestore map would only be populated if APIGroupVersionsFeatureFlag
// was enabled for restore and the minimum required backup format version has been met.
```

pkg/restore/prioritize_group_version_test.go
```diff
@@ -0,0 +1,367 @@
...
+		},
+		{
+			name: "incorrect and correct data format returns some user version priorities",
+			cm: builder.
+				ForConfigMap("velero", "enableapigroupversions").
+				Data(
+					"restoreResourcesVersionPriority",
+					`rb=foo,foo2
+o=v2beta2
+\n`,
```
zubron, line 125: suggestion: I think this test case could be improved by using invalid data that's a bit closer to valid data. e.g. `bar=`
Also, you could include some valid lines after the invalid lines to show that it continues processing after an failed validation.

pkg/restore/prioritize_group_version_test.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+
+	for _, tc := range tests {
+		t.Log(tc.name)
+		priorities := userResourceGroupVersionPriorities(fakeCtx, tc.cm)
+		assert.Equal(t, tc.want, priorities)
+	}
+}
+
+func TestFindTargetGroup(t *testing.T) {
```
zubron, line 151: This test should be renamed as it's the function `findAPIGroup` that's under test here.

zubron (changes requested):
Thank you so much for this, @codegold79! This is looking great so far. Excellent work on all the test cases, those really help :clap:

test/e2e_testify/restore_priority_apigv_test.go
```diff
@@ -0,0 +1,226 @@
...
+		tgtCRD     map[string]string
+		tgtVer     string
+		cm         *corev1api.ConfigMap
+		gvs        map[string][]string
+		want       map[string]map[string]string
+	}{
+		{
+			name: "Target and source cluster preferred versions match; Preferred version v1 is restored (Priority 1, Case A).",
+			srcCRD: map[string]string{
+				"url":       "https://raw.githubusercontent.com/brito-rafa/k8s-webhooks/master/examples-for-projectvelero/case-a/source/case-a-source.yaml",
```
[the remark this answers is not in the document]
codegold79: I too would like to store these e2e testify files for others to reference in the future if they would like. Do you know of a good location to store them?

pkg/archive/parser_test.go
```diff
@@ -110,11 +111,161 @@ func TestParse(t *testing.T) {
...
+	tests := []struct {
+		name      string
+		files     []string
+		backupDir string
+		wantErr   error
+		want      map[string]metav1.APIGroup
+	}{
+		{
+			name:      "when there is no top-level resources directory, an error is returned",
+			backupDir: "/var/folders/zj/vc4ln5h14djg9svz7x_t1d0r0000gq/T/620385697",
```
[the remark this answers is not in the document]
codegold79: Yup, I'll get that changed to `backupDir : "/var/folders"`. That will be more consistent with other tests as well as be easier to read.

pkg/archive/parser_test.go
```diff
@@ -110,11 +111,161 @@ func TestParse(t *testing.T) {
...
+				if !strings.HasSuffix(file, "/") {
+					res, err := p.fs.Create(file)
+					require.NoError(t, err)
+					require.NoError(t, res.Close())
+				}
+			}
+
+			res, err := p.ParseGroupVersions(tc.backupDir)
+			if tc.wantErr != nil {
+				assert.Equal(t, tc.wantErr.Error(), err.Error())
```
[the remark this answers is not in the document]
codegold79: I'll give it a try, thanks for the tip 👍

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,337 @@
...
+			ctx.log.Debugf(
+				"APIGroupVersionsFeatureFlag Priority 2: Cluster preferred API group version not found in backup. Using backup preferred version %s for %s",
+				sg.PreferredVersion.Version,
+				rg,
+			)
+			continue
+		}
+
+		// Priority 3: The Common Supported Version with the Highest Kubernetes Version Priority
+		for _, tv := range tg.Versions[1:] {
```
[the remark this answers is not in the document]
codegold79: Short answer: The first elements are the preferred versions. Preferred versions were checked as potential versions to restore in priorities 1 and 2.

Long answer: I am using Kubernetes priority versions ordering. After k8s orders the versions, the first one (the latest, most stable version) will be the preferred. Because I order all the versions as soon as I get them from source, target, and user, the first first will always be the preferred version. And because preferred versions were already considered and rejected by the time we get to the Priority 3 case, I skipped checking them.

pkg/restore/prioritize_group_version_test.go
codegold79: Good catch. I'll have that changed.

pkg/archive/parser_test.go
```diff
@@ -110,11 +111,161 @@ func TestParse(t *testing.T) {
...
+			name:      "when there are no directories under the resources directory, an empty map is returned",
+			backupDir: "/var/folders/zj/vc4ln5h14djg9svz7x_t1d0r0000gq/T/620385697",
+			files:     []string{"/var/folders/zj/vc4ln5h14djg9svz7x_t1d0r0000gq/T/620385697/resources/"},
+			want:      map[string]metav1.APIGroup{},
+		},
+		{
+			name:      "when there is a mix of cluster-scoped and namespaced items for resources with preferred or multiple API groups, all group versions are correctly returned",
+			backupDir: "/var/folders/zj/vc4ln5h14djg9svz7x_t1d0r0000gq/T/620385697",
+			files: []string{
+				"/var/folders/zj/vc4ln5h14djg9svz7x_t1d0r0000gq/T/620385697/resources/clusterroles.rbac.authorization.k8s.io/v1-preferredversion/cluster/system/controller/attachdetach-controller.json",
```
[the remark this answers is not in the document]
codegold79: I appreciate your careful review because you're right. I missed the cluster directory at the same level as the preferred version directory. I'll add that line.

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,337 @@
...
+
+		ups[rg] = metav1.APIGroup{
+			Versions: versionsToGroupVersionForDiscovery(vs),
+		}
+	}
+
+	return ups
+}
+
+func isUserPriorityValid(line string) bool {
```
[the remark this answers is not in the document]
codegold79: I will add some error handling to this functions and functions that call it.

pkg/restore/prioritize_group_version_test.go
codegold79: I made validation of user provided config maps stricter and now if any lines are invalid, an error will occur. 

It should be easier to see with my changes which are valid lines and which are invalid.

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,337 @@
...
+		}
+
+		// Priority 3: The Common Supported Version with the Highest Kubernetes Version Priority
+		for _, tv := range tg.Versions[1:] {
+			if versionsContain(sg.Versions[1:], tv.Version) {
+				ctx.SetChosenGVToRestore(cgv, rg, sg.PreferredVersion.Version, tv.Version)
+				continue OUTER
+			}
+			ctx.log.Debugf(
+				"APIGroupVersionsFeatureFlag Priority 3: Common supported but not preferred API group version %s chosen for %s",
```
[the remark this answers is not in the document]
codegold79: Another great catch. Yes, the log line should be before `continue OUTER` in the `if` statement block.

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,337 @@
...
+		if len(tg.Versions) == 0 {
+			ctx.SetChosenGVToRestore(cgv, rg, "", "")
+			continue
+		}
+
+		// Priority 0: User Priority Version
+		if userGVs != nil {
+			uv := findSupportedUserVersion(userGVs[rg].Versions, tg.Versions, sg.Versions)
+			if uv != "" {
+				ctx.SetChosenGVToRestore(cgv, rg, sg.PreferredVersion.Version, uv)
```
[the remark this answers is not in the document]
codegold79: Thank you for your feedback. I also would not expect a `set` function to be modifying its arguments. And I agree that I should follow that commonly understood convention of `set`. I ended up getting rid of the `set` function and moved the logic into the calling function.

The reason I had the `set` function mutate its arguments is because I thought it would reduce writing repetitive mutation code. However, it came at the cost of readability. 

I believe it's worth repeating some code and logic if it means it's easier to follow. As a result, I have made some changes in accordance to your suggestions.

test/e2e_testify/restore_priority_apigv_test.go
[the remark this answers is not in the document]
zubron: My suggestion would be to create a `testdata` (or similar) directory within the `test/e2e/` directory and then reference them within the test by using a relative path. I tried this out locally and was able to replace the usage of `url` with a path like `testdata/restore-priority/cert-manager.yaml`.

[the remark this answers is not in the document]
codegold79: Oops, I misunderstood you initially. You meant the CRD and CR files stored on @brito-rafa's GitHub. I'll get those transferred over to Velero `testdata/enable-api-group-versions`. I'll talk with @brito-rafa about how to give him credit for his work.

pkg/restore/prioritize_group_version.go
```diff
@@ -0,0 +1,337 @@
...
+		if versionsContain(sg.Versions, tg.PreferredVersion.Version) {
+			cgv.Version = tg.PreferredVersion.Version
+			cgv.Dir = tg.PreferredVersion.Version
+
+			if tg.PreferredVersion.Version == sg.PreferredVersion.Version {
+				cgv.Dir += velerov1api.PreferredVersionDir
+			}
+
+			ctx.chosenGrpVersToRestore[rg] = cgv
+			ctx.log.Debugf(
```
nrb, line 98: I like that the previous case (priority 0) has an `Info` log message that it was unsuccessful. Could you add `Info` lines for when cases 1-3 fail to match too, so the user has more visibility into what the code decided without having to switch on debug logs? Switching on debug logs isn't a huge deal, but it does mean stopping the Velero deployment and re-running the backup job.

pkg/restore/restore.go
```diff
@@ -384,6 +388,16 @@ func (ctx *restoreContext) execute() (Result, Result) {
 		return warnings, errs
 	}
 
+	// TODO: Remove outer feature flag check to make this feature a default in Velero.
+	if features.IsEnabled(velerov1api.APIGroupVersionsFeatureFlag) {
+		if ctx.backup.Status.FormatVersion >= "1.1.0" {
+			if err := ctx.chooseAPIVersionsToRestore(); err != nil {
```
nrb, line 394: Since this call goes down to [an APIGroup discovery call](https://github.com/vmware-tanzu/velero/pull/3133/files#diff-597d46a83a1ad70945ab451fa42ec1eb96dbcbf427949e93b3a87792883d11f3R165), we may want to re-run the `chooseAPIVersionsToRestore` function [after restoring CRDs](https://github.com/vmware-tanzu/velero/pull/3133/files#diff-e0498755be37b683528e2569667d67818bb676827c420aab49adc62cb6d47fe3R492)

This is because at this line, we'll only have groups in the API server that existed before restore. As CRDs are restored, they'll be added into the API server for discovery, thus the list of API Groups discover here will fall out of date.

As an example, if I'm restoring the CRD `beatles.rockband.io`, it will not be listed in the API Groups at this line. Instead, it wouldn't be added until L492 of this file, since it was added during the restore process, and the API server would have it before then.

This is a problem Velero ran into before adding the "lazy restore" at L492, in that custom resources would try to restore, but their CRD did not yet exist in the API server yet.

Please let me know if this is unclear - it's a tricky race condition and may be better diagrammed.

site/content/docs/main/enable-api-group-versions-feature.md
```diff
@@ -0,0 +1,102 @@
+---
```
nrb, line 1: These docs are awesome, thank you!

nrb (changes requested):
From my POV, this is _super_ close. I have 1 suggestion on function naming, and one order of operations comment.

The order of operations is more fundamental to getting this working correctly; the naming isn't a blocker.

test/e2e_testify/restore_priority_apigv_test.go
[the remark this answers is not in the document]
codegold79: The CRD and CR files have been moved to their own `testdata/enable-api-group-versions` directory in the latest commit.

pkg/archive/parser.go
```diff
@@ -165,3 +167,102 @@ func (p *Parser) getResourceItemsForScope(dir, archiveRootDir string) ([]string,
 
 	return items, nil
 }
+
+// CheckAndReadDir is a wrapper around fs.DirExists and fs.ReadDir that does checks
+// and returns errors if directory cannot be read.
+func (p *Parser) CheckAndReadDir(dir string) ([]os.FileInfo, error) {
```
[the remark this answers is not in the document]
codegold79: I agree that `CheckAndReadDir` should be lower-cased and be a private function.

pkg/restore/prioritize_group_version.go
codegold79: Good idea--will do.

pkg/restore/restore.go
codegold79: Since this is a tricky to wrap my head around. Hopefully you or @brito-rafa can correct me if I'm mistaken.

As I understand it, EnableAPIGroupVersions does not restore CRDs. It only restores CRs. We assume the CRDs will be installed manually by the user before the restore OR the CRDs will be the newer versions that come with Kubernetes. The reason I think that is because the new CRDs won't be in the backups, only the old ones will be.

But given my assumption, it brings up a question: when are the old CRDs restored? Even if EnableAPIGroupVersions doesn't play a part in that process, it needs to be able to discover these restored CRDs during the process of choosing a version to restore.

test/e2e_testify/restore_priority_apigv_test.go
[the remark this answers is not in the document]
codegold79: @brito-rafa mentioned a couple of things to me that should be considered. Here's a message he sent to me on Slack:

> ...the manifests depends on the controller images and they [the Velero team] might want to have the controller images on velero registry.

He also said that perhaps the Velero team might not want these eight controllers (CRDs) as part of the Velero code base. 

Should we have another meeting about this topic?

pkg/restore/restore.go
nrb: > when are the old CRDs restored?

Great question! That information isn't immediately obvious in just this file.

[By default, Velero will restore them as the first items out of the tarball](https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/server/server.go#L448).

Inside the main restore loop, [we check if a CRD was restore, and if so, update the discovery client's list of API Groups.](https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L477-L482). For each CRD we restore, [we wait to see if the Kubernetes API server marks it as available](https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L1185-L1186).

So the full list of CRDs won't be restored to the API server until after that loop finishes with the `customresourcedefinition` set and moves on to the rest in the `restoreResourcePriorities` list.

Regarding new versions of CRDs, I think you're correct - they'll need to be present prior to the restore happening, since Velero doesn't have them. For builtin kinds/API groups, this isn't a big deal, as they'll already be there. For applications, it may be trickier as the user will need to make sure only the CRDs are defined prior to the restore.

brito-rafa: So, as part of our testing using [rockbands.music.example.io](https://github.com/brito-rafa/k8s-webhooks/tree/master/examples-for-projectvelero), the target cluster has already a newer version of CRD at that time of the velero restore. This is a realistic scenario because the cluster-admin would have onboarded a newer controller in charge of such CRD. In case the target cluster does not have the CRD installed, velero restore "business as usual" procedure will kick in and create them because they are part of the velero backup tarball.

See below the `velero restore describe` from a restore using this PR velero version on a cluster that has the newer CRD already in place. Note that velero attempts to restore the CRD and then skips realizing that the CRD already exists on the target cluster and it is different. The CRs are restored.  Does it make sense?

```bash
velero restore describe case-b-20210202-20210203150211
(...)
Warnings:
  Velero:     <none>
  Cluster:  could not restore, customresourcedefinitions.apiextensions.k8s.io "rockbands.music.example.io" already exists. Warning: the in-cluster version is different than the backed-up version.
  Namespaces: <none>
```

brito-rafa: So far, this PR has not touched anything on CRD backup/restore logic. For your point Nolan, I think makes sense to re-run the discovery after velero restore the CRDs.

nrb: @brito-rafa You're right that it doesn't touch CRDs specifically, I just wanted to call out how the discovery is working and how that may interact with this change.

But yeah, that makes sense - Velero can't know about the new versions, but in terms of discovering old versions, we need to check after we restore the CRDs and/or look in the backup (which we're already doing).

codegold79: Very informative conversation. Thank you @nrb to the links to the code, it helped me to understand at the code level what was happening. And thanks @brito-rafa for sharing that warning. I hadn't made the connection before that is why the "could not restore" message was coming up.

Given that discovery is already called after CRDs are restored, and the EnableAPIGroups feature only affects the restore process for non-CRDs by changing the file path to the resource in the backup, I think this comment thread can be marked "Resolved". Feel free @nrb and @brito-rafa to unresolve if I missed something.

brito-rafa:
I've run more tests and I encountered a problem when handling the configmap.  I created the configmap running the following:

```bash
$ echo "rockbands.music.example.io=v2beta1,v2beta2" > restoreResourcesVersionPriority
$ kubectl create configmap enableapigroupversions --from-file=restoreResourcesVersionPriority -n velero
```

All my restores failed with this message:
```
Errors:
  Velero:   choosing API version to restore: reading enableapigroupversion config map: parsing user priorities: validating user priority: line must have one and only one equal sign
```

The problem was with `echo` adding an extra-line. I ended up creating another configmap with a space on the line and I got the following:

```
Errors:
  Velero:   choosing API version to restore: reading enableapigroupversion config map: parsing user priorities: validating user priority: line must not contain any spaces
```

Putting on my SRE hat, I found the configmap checks too strict. I would ignore blank lines, trim spaces, and then try to parse the values.
I argue that even if the lines are badly configured (ie invalid syntax, like two equal signs), I would make a warning saying that we could not parse the configmap and state that we are ignoring it and move on with the restore.

codegold79:
@brito-rafa [wrote](https://github.com/vmware-tanzu/velero/pull/3133#issuecomment-772916247),
> Putting on my SRE hat, I found the configmap checks too strict. I would ignore blank lines, trim spaces, and then try to parse the values.

> I argue that even if the lines are badly configured (ie invalid syntax, like two equal signs), I would make a warning saying that we could not parse the configmap and state that we are ignoring it and move on with the restore.

Thanks @brito-rafa for doing these tests and finding things that I, as the developer, overlooked. Of course, when I run the exact same tests, I don't encounter these errors because I wrote and know all the rules. I even wrote in the documentation how the config map should be formatted, but I agree it shouldn't necessary be so strictly adhered to.

I believe your suggestions are very good and I'll make those changes.

brito-rafa:
Just to put some color how I tested it. 

The 3133 testing script is on my velero-test repo:
https://github.com/brito-rafa/velero-test/blob/master/3133-dev-test.sh

And a regular output of the testing script is:
https://github.com/brito-rafa/velero-test/blob/master/3133-restore-005-output.txt

I am not going into details here, but I put some *crude* success criteria in the testing script, as per rockband webhook logic and testing cases. Bottom line: the velero restore is doing what it is proposing to do. 
Please reach out to me if you want to get more color in this success criteria.

The latest version of the image and testing script accepted the configmap with extra lines and extra spaces:

```
Data
====
restoreResourcesVersionPriority:
----
   rockbands.music.example.io = v2beta1, v2beta2 
Events
```

nrb (dismissed):
Setting my approval for this so we can start getting the e2e testing running on it, and we can build on top of it. I think all my comments have been addressed.

brito-rafa:
I suggest two more tweaks to the official documentation in regard to this PR.

1-) The current page "output-file-format.md" states "This is an experimental flag and the restore logic to handle multiple API Group Versions will be added in the future." 
I recommend changing "This is an experimental flag and the restore logic to handle multiple API Group Versions is documented at enable-api-group-versions-feature.md (making a cross-reference between the pages)

2-) The current page "migration-case.md" does not mention migration of workloads across different Kubernetes versions. I suggest adding a simple section to suggest the reader checking the page "enable-api-group-versions-feature.md" when migrating workloads that clusters are not running the same version of Kubernetes.

site/content/docs/main/migration-case.md
```diff
@@ -50,5 +50,9 @@ Check that the second cluster is behaving as expected:
 
 If you encounter issues, make sure that Velero is running in the same namespace in both clusters.
 
+## Migrating Workloads Across Different Kubernetes Versions
+
+Migration across clusters that are not running the same version of Kubernetes is possible with Velero. For more information, look at the [EnableAPIGroupVersions](enable-api-group-versions-feature.md) feature.
```
nrb, line 55: Migrating between kubernetes versions doesn't strictly require this feature flag, so long as there were no incompatible changes to the core types. This is a little tricky to provide guidance on, because while I know v1.15.->1.16 did this, not all versions do, and I don't really want to maintain an exhaustive list of version jumps that this is relevant to. 

Could we maybe add some wording to this about Kubernetes upgrades containing incompatible or removed API groups?

nrb (changes requested):

nrb (dismissed):
Given my documentation request can be addressed later, I'm going to give my approval and address those later.

site/content/docs/main/migration-case.md
nrb: We'll address these comments in a different issue or PR.

nrb:
@zubron this is now in your court :)

site/content/docs/main/migration-case.md
codegold79: Thanks, @nrb! I have some additions per your comment, but I'm going to consult my go-to Kubernetes expert, @brito-rafa, to see if I've covered what was intended. Hopefully, I'll have the commit up by end of today 😃

codegold79: @nrb, I have updated the paragraph. Let me know if that doesn't work.

test/e2e/enable_api_group_versions_test.go
```diff
@@ -0,0 +1,580 @@
...
+				"specs": {
+					"leadSinger": "John Lennon",
+					"leadGuitar": "George Harrison",
+					"genre":      "60s rock",
+				},
+			},
+		},
+	}
+
+	for i, tc := range tests {
```
zubron, line 201: Ideally these tests should use the Ginkgo `DescribeTable` function as we do in the [controller tests](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/server_status_request_controller_test.go#L61) as it ensures that each test case is reported individually. This can be improved in a follow up though and shouldn't block this PR.

zubron (approved):
Thank you *so* much for all your efforts on this, @codegold79! I had a suggestion about the e2e tests but it's not urgent at all and I'm happy for this to be merged as is.

zubron:
@nrb This needs re-approval from you :)

nrb (approved):

test/e2e/enable_api_group_versions_test.go
codegold79: Thanks, @zubron! I'll try to divide the test cases into their own Describe blocks using the controller tests as my example. I'll have it up in a follow-up PR.

PR #3162, opened 2020-12-08
Add milestoned issues to their respective board

nrb:
As long as a milestone and the board have the same title, then this
workflow should take care of adding an issue into the GitHub Project
board when an existing issue is given a milestone.

It does NOT support checking for a milestone when an issue is edited or
created though, due to limitations on GitHub Actions syntax right now -
there's not a great way to validate against an empty `milestone` object
at the moment, per https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions

### DIFF
--- a/.github/workflows/milestoned-issues.yml
+++ b/.github/workflows/milestoned-issues.yml
@@ -1,0 +1,18 @@
+name: Add issues with a milestone to the milestone's board
+
+on:
+  issues:
+    types: [milestoned]
+
+jobs:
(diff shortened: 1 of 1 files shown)


carlisia (approved):

nrb:
@carlisia This has changed since I changed the case on `GITHUB_TOKEN`

zubron (approved):

carlisia (approved):

PR #3163, opened 2020-12-08
Use new repository-local board & github secret

nrb:

### DIFF
--- a/.github/workflows/opened-issues-triage.yml
+++ b/.github/workflows/opened-issues-triage.yml
@@ -10,6 +10,6 @@ jobs:
     steps:
       - uses: alex-page/github-project-automation-plus@v0.3.0
         with:
-          project: "Velero Support Triage Board"
-          column: "Needs triage"
-          repo-token: ${{ secrets.GH_TOKEN }}
+          project: "Velero Support Board"
+          column: "New"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}


carlisia (approved):

.github/workflows/opened-issues-triage.yml
```diff
@@ -10,6 +10,6 @@ jobs:
     steps:
       - uses: alex-page/github-project-automation-plus@v0.3.0
         with:
-          project: "Velero Support Triage Board"
-          column: "Needs triage"
-          repo-token: ${{ secrets.GH_TOKEN }}
+          project: "Velero Support Board"
+          column: "New"
+          repo-token: ${{ secrets.GitHub_TOKEN }}
```
zubron, line 15: Is this change in token necessary? I can't see it in the list of secrets for the repo.

nrb: @zubron It should be all caps so I'll fix that, but it's a generated token vs one that's associated with anyone's account.

See https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

zubron: Ah, okay! I thought there was a reason that we couldn't use that before, but I think that might have been because those tokens wouldn't work with org level boards.

zubron (approved):

.github/workflows/opened-issues-triage.yml
nrb: Correct, `GITHUB_TOKEN` is repository-scoped only, it can't access org-level stuff.

nrb:
@carlisia This will need another approval since I pushed a case fix.

carlisia (approved):

PR #3166, opened 2020-12-08
Change distro

carlisia:
5 minutes into using this and I wanted to exec into the Velero container. I assume we will all want to have this? Not as lean, but the ability of exec'ing might be missed.


zubron (approved):
:+1:

nrb (approved):

PR #3172, opened 2020-12-09
Improvements to BSL logic

carlisia:
There's only 1 change in functionality in this PR, and it is the automatic setting of the BSL created at install time as the "default".

Summary:
- Some code, variable name, and documentation improvements.
- Added deprecation notices at some relevant points
- In the pkg/controller/backup_controller.go:
  - grouped together all the code related to getting the default BSL
  - improved (I think) the error message for when the default BSL is not found/doesn't exist
  - introduced an error message specifically for when the default location we try to fetch is one that was specified with the server flag. When this happens it is usually because it was never created. Having this message point out that the config came from the server will help people understand why this is happens. Might be useful since this is not very transparent or intuitive.
- In the pkg/install/resources.go, I'm setting the BSL created at installation time as the "default" one. This makes sense since it is the first created in the cluster, and also goes hand-in-hand with Velero automatically naming that BSL `default`.

### DIFF
--- a/changelogs/unreleased/3172-carlisia
+++ b/changelogs/unreleased/3172-carlisia
@@ -1,0 +1,1 @@
+Set the BSL created at install time as the "default"
--- a/internal/storage/storagelocation.go
+++ b/internal/storage/storagelocation.go
@@ -30,10 +30,11 @@ import (
 
 // DefaultBackupLocationInfo holds server default backup storage location information
 type DefaultBackupLocationInfo struct {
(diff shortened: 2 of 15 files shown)


jenting (approved):
> There's only 1 change in functionality in this PR, and it is the automatic setting of the BSL created at install time as the "default".

Thank you for addressing this, I forget the fresh install case, thank you :bow:

nrb (approved):

PR #3180, opened 2020-12-14
Minor kubebuilder related items to clean up 

carlisia:
This closes https://github.com/vmware-tanzu/velero/issues/2884.

There is a test file that was deleted bc the functionality was moved to inside the controller, but the controller test is testing the same things anyway (pkg/controller/server_status_request_controller_test.go).

There is no change in logic with this PR.


pkg/cmd/server/server_test.go
```diff
@@ -69,3 +70,80 @@ func TestVeleroResourcesExist(t *testing.T) {
...
+			}
+
+			totalNumOriginalControllers := len(enabledControllers) + len(enabledRuntimeControllers)
+
+			if tt.errorExpected {
+				assert.Error(t, removeControllers(tt.disabledControllers, enabledControllers, enabledRuntimeControllers, logger))
+			} else {
+				assert.NoError(t, removeControllers(tt.disabledControllers, enabledControllers, enabledRuntimeControllers, logger))
+
+				totalNumEnabledControllers := len(enabledControllers) + len(enabledRuntimeControllers)
```
zubron, line 144: Optional, but I think it would be nice if instead of checking the counts, we iterate over the `disabledControllers` slice and check that that key doesn't appear in the `enabledControllers` or `enabledRuntimeControllers` map.

zubron (approved):
LGTM - thanks! I've got one question, and I've left a suggested change but they aren't blockers.

pkg/cmd/server/server_test.go
carlisia: I agree. I left the testing of the count too. ✅

pkg/controller/server_status_request_controller.go
```diff
@@ -85,14 +90,26 @@ func (r *ServerStatusRequestReconciler) Reconcile(req ctrl.Request) (ctrl.Result
 	case "", velerov1api.ServerStatusRequestPhaseNew:
 		log.Info("Processing new ServerStatusRequest")
 
-		if err := r.ServerStatus.PatchStatusProcessed(r.Client, statusRequest, r.Ctx); err != nil {
-			log.WithError(err).Error("Unable to update the request")
+		// Initialize the patch helper.
+		patchHelper, err := patch.NewHelper(statusRequest, r.Client)
+		if err != nil {
+			log.WithError(err).Error("Error getting a patch helper to update this resource")
+			return ctrl.Result{}, err
```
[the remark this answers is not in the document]
carlisia: It's rare that you wouldn't want to re-queue on failure. One case is when the object is not found: most likely it will continue not to be found. In this case, return a nil instead of an error.

Anytime an error is return, the request will be re-queued.

zubron (dismissed):

pkg/cmd/server/server_test.go
```diff
@@ -69,3 +70,87 @@ func TestVeleroResourcesExist(t *testing.T) {
...
+			disabledControllers: []string{
+				controller.Backup,
+				controller.BackupStorageLocation,
+			},
+			errorExpected: true,
+		},
+		{
+			name: "Remove with a misspelled controller name",
+			disabledControllers: []string{
+				"bakup",
```
zubron, line 115: I'm not sure how to best handle this spelling issue. Would this test still work to have a correctly spelled word, but one that isn't a valid controller name?

carlisia: If by invalid controller name you mean a controller that is not on the list of ones that are disabable, then yes, this test would still work the same. You can see it in this other test: https://github.com/vmware-tanzu/velero/pull/3180/files#diff-073f5a2184977287c11f4b4c94aced2fab5718ee4946a2b050d472fa6059496fR105. Included in the list of controllers to disable is the BSL controller, so there'll be an error.

Is this what you mean?

zubron: Yeah, I was just thinking of how to trigger the behaviour that we're testing for but fix the codespell issue without adding the word to the list of skipped words as it's something that we could easily misspell elsewhere :)

carlisia: Oh duh, I JUST realized what you mean haha. I like this test because it stands for both "a controller that doesn't exist" and "a controller with a very similar name exists but the system handles it properly". Not married to it though. The check is not a merge blocker. I feel like we should have some sorta rule of thumb for things like this.

carlisia: Yes, we don't want to add this to the list of skipped words. Let me misspell something else.

carlisia:
This is ready for review.

jenting (commented):
I did not see the changes about `remove rbac/role file (config/rbac/role.yaml), as well as it's generation in the script (I think this will work: https://github.com/vmware-tanzu/velero/blob/main/hack/update-generated-crd-code.sh#L50)` in this PR.

carlisia:
> I did not see the changes about remove rbac/role file (config/rbac/role.yaml), as well as it's generation in the script (I think this will work: https://github.com/vmware-tanzu/velero/blob/main/hack/update-generated-crd-code.sh#L50) in this PR.

@jenting I had removed that line and tested it and made no difference. But, I don't think removing it will cause any problem either, so I just removed it.

carlisia:
@vmware-tanzu/velero-maintainers this is ready for review.

pkg/controller/server_status_request_controller_test.go
```diff
@@ -250,8 +95,152 @@ var _ = Describe("Server Status Request Reconciler", func() {
...
+					},
+				}).
+				Result(),
+			expectedRequeue: ctrl.Result{Requeue: false, RequeueAfter: statusRequestResyncPeriod},
+		}),
+		Entry("with phase=Processed gets deleted if expired", request{
+			req: statusRequestBuilder("1").
+				ServerVersion(buildinfo.Version).
+				Phase(velerov1api.ServerStatusRequestPhaseProcessed).
+				ProcessedTimestamp(now.Add(-61 * time.Second)). // expired
```
jenting, line 201: nit: it would be good if we could calculate the expired time by referencing the `ttl` https://github.com/vmware-tanzu/velero/blob/e69fac1/pkg/controller/server_status_request_controller.go#L37, but I won't block this PR.

jenting (approved):
LGTM

zubron (approved):
Thanks!

PR #3183, opened 2020-12-14
Better name format for init containers

carlisia:
## What:

Instead of using the image name as a container init name, make it so the name include the repository name, and make the string DNS-1123 compatible (no slashes).

No change in logic and no compatibility issues.

## Why:

There can be plugins from different providers that are named the same. In this case, when attempting to add a second plugin with the same image name as an existing one, Velero will fail. See this example:

```
❯ velero plugin add storjlabs/velero-plugin
```

```
❯ velero plugin add digitalocean/velero-plugin
An error occurred: Deployment.apps "velero" is invalid: spec.template.spec.initContainers[2].name: Duplicate value: "velero-plugin"
```

That's because we attempt to name the init container `velero-plugin` in both cases.

## Solution:

The concatenate repository name + image name would be the name of the init container. Example:

`storjlabs-velero-plugin`

`digitalocean-velero-plugin`

### DIFF
--- a/pkg/builder/container_builder.go
+++ b/pkg/builder/container_builder.go
@@ -44,29 +44,30 @@ func ForPluginContainer(image string, pullPolicy corev1api.PullPolicy) *Containe
 }
 
 // getName returns the 'name' component of a docker
-// image that includes its reposiroty name, and transforms the combined
+// image that includes the entire string except the registry name, and transforms the combined
 // string into a DNS-1123 compatible name.
 func getName(image string) string {
 	slashIndex := strings.Index(image, "/")
-	slashCount := strings.Count(image[slashIndex:], "/")
-	colonIndex := strings.LastIndex(image, ":")
+	slashCount := 0
+	if slashIndex >= 0 {
+		slashCount = strings.Count(image[slashIndex:], "/")
+	}
 
-	// this removes the registry name when there is one, but keeps the repository name
 	start := 0
-	if slashCount == 1 {
-		start = 0
-	} else {
-		// this will be the first character after the first found slash
+	if slashCount > 1 || slashIndex == 0 {
+		// always start after the first slash when there is a registry name
+		// or if the string starts with a slash.
 		start = slashIndex + 1
 	}
 
 	// this removes the tag
+	colonIndex := strings.LastIndex(image, ":")
 	end := len(image)
 	if colonIndex > 0 {
 		end = colonIndex
 	}
 
-	return strings.Replace(image[start:end], "/", "-", 1) // this makes it DNS-1123 compatible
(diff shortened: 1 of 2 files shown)


pkg/builder/container_builder.go
```diff
@@ -44,23 +44,29 @@ func ForPluginContainer(image string, pullPolicy corev1api.PullPolicy) *Containe
 }
 
 // getName returns the 'name' component of a docker
-// image (i.e. everything after the last '/' and before
-// any subsequent ':')
+// image that includes its reposiroty name, and transforms the combined
+// string into a DNS-1123 compatible name.
 func getName(image string) string {
-	slashIndex := strings.LastIndex(image, "/")
+	slashIndex := strings.Index(image, "/")
```
zubron, line 50: If `/` isn't in the string, this will return -1 which will cause the next line to panic.

pkg/builder/container_builder.go
```diff
@@ -44,23 +44,29 @@ func ForPluginContainer(image string, pullPolicy corev1api.PullPolicy) *Containe
...
 
+	// this removes the tag
 	end := len(image)
-	if colonIndex > slashIndex {
+	if colonIndex > 0 {
 		end = colonIndex
 	}
 
-	return image[start:end]
+	return strings.Replace(image[start:end], "/", "-", 1) // this makes it DNS-1123 compatible
```
zubron, line 69: I think we need to replace all remaining slashes with `-` not just the first one. I've encountered some images with many slashes when doing some image name rewriting in Sonobuoy, e.g `us.gcr.io/k8s-artifacts-prod/e2e-test-images/agnhost:2.12`.

pkg/builder/container_builder_test.go
```diff
@@ -30,27 +30,27 @@ func TestGetName(t *testing.T) {
 		{
 			name:     "image name with registry hostname and tag",
 			image:    "gcr.io/my-repo/my-image:latest",
-			expected: "my-image",
+			expected: "my-repo-my-image",
```
zubron, line 33: I think we should add a couple more test cases here, one to cover the case where there are no slashes in the image name, and one to cover the case were there are more than 2 slashes.

zubron (changes requested):
Thanks for this, @carlisia and for spotting the issue in the first place! I like this approach, there are just a couple corner cases that need to be handled.

pkg/builder/container_builder.go
carlisia: Good catch, addressed.

pkg/builder/container_builder.go
carlisia: I was wondering if I should do that. ✅

zubron (approved):
Thanks for adding the extra test cases!

jenting (approved):
LGTM

PR #3185, opened 2020-12-14
Fix path to crds.go file in codespell config

zubron:
This changes the codespell action config to use a relative path for the
generated crds.go file as the current pattern used fails the check used
by codespell (which uses the `fnmatch` module).


zubron:
With this change the [codespell job](https://github.com/vmware-tanzu/velero/pull/3185/checks?check_run_id=1553720427) is now passing. On other PRs [it fails](https://github.com/vmware-tanzu/velero/pull/3183/checks?check_run_id=1553502030), even if the crds.go file has not been modified.

nrb (approved):
Awesome, thanks for fixing this!

carlisia (approved):
🎉

PR #3187, opened 2020-12-15
Only sync from available BSLs

carlisia:
What do y'all think of this idea?

Mainly:
- if BSL is not available, continue and don't attempt to sync (bc, unavailable)
- only even initialize the backup store if the BSL is available AND the # of items to sync is > 0

Some changes to make this work: a new field `location.Soec.BackupList` would need to be added to the BSL. This field would be a list of strings containing a list of all backups in the store corresponding to that BSL. This field would be populated everytime a BSL was validated.

The sync interval would additionally be dependent on the BSL logic for how frequent to validate a location (https://github.com/vmware-tanzu/velero/blob/main/internal/storage/storagelocation.go#L39-L77).

### DIFF
--- a/pkg/controller/backup_sync_controller.go
+++ b/pkg/controller/backup_sync_controller.go
@@ -167,21 +167,15 @@ func (c *backupSyncController) run() {
 			}
 		}
 
-		log.Debug("Checking backup location for backups to sync into cluster")
-
-		backupStore, err := c.newBackupStore(&location, pluginManager, log)
-		if err != nil {
-			log.WithError(err).Error("Error getting backup store for this location")
+		log.Debug("Checking if this location is available")
+		if location.Status.Phase != velerov1api.BackupStorageLocationPhaseAvailable {
+			log.Debug("backup location not available, can't sync at this time")
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


nrb:
> a new field location.Soec.BackupList would need to be added to the BSL.

1) We may very easily run into Kube API server object size limits with this, because users could have a LOT of backups in a BSL depending on their scheduled frequency.
2)  I think this should be in the `Status` if we're fully managing it, versus letting the user set it.

Would it be possible to record only the number of backups instead of a full list and get the same result? I like the overall idea here, just trying to think of a way to make sure we can scale it.

nrb:
@carlisia What's the status of this PR?

carlisia:
@nrb sorry I missed your question. I will be thinking through this issue and the questions you raised when I start thinking about the CLI work for v1.7, which is soon.

dsu-igeek:
This is superseded by merging the sync and BSL controller together, per Carlisia

carlisia:
I realized I hadn't opened a ticket for this. Here it is: [Merge the BSL and sync controllers · Issue #3908 · vmware-tanzu/velero](https://github.com/vmware-tanzu/velero/issues/3908).

PR #3189, opened 2020-12-15
Add Tilt configuration to debug using Delve

zubron:
This change adds support to run the Velero process in Tilt using
[Delve](https://github.com/go-delve/delve).
This does not include support for debugging the Velero process in the
restic pods, just in the Velero deployment.

For an optimal debugging experience, this change also introduces a new
flag (`DEBUG`) to the `hack/build.sh` script to enable a "debug" build
of the Velero binary. This flag, if enabled, will build the binary
without optimisations and inlining. Disabling optimisations and inlining
is recommended by Delve.

Two configuration options have been added to the Tilt settings. The
first, `enable_debug`, is to control whether debugging should be
enabled. If enabled, the process will be started by Delve, and the Delve
server port (2345) will be forwarded to the local machine.
The second option, `debug_continue_on_start`, is to control whether the
process should "continue" when started by Delve or should be paused.
By default, debugging is disabled, and if in debug mode, the process
will continue.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -61,7 +61,6 @@ RUN wget --output-document /restart.sh --quiet https://raw.githubusercontent.com
 additional_docker_helper_commands = """
 # Install delve to allow debugging
 RUN go get github.com/go-delve/delve/cmd/dlv
-RUN which dlv
 
 RUN wget -qO- https://dl.k8s.io/v1.19.2/kubernetes-client-linux-amd64.tar.gz | tar xvz
 RUN wget -qO- https://get.docker.com | sh
--- a/Tiltfile
+++ b/Tiltfile
@@ -27,26 +27,26 @@ settings = {
 # global settings
 settings.update(read_json(
     "tilt-resources/tilt-settings.json",
-    default={},
+    default = {},
 ))
 
 k8s_yaml(kustomize('tilt-resources'))
 k8s_yaml('tilt-resources/deployment.yaml')
 if settings.get("enable_debug"):
-    k8s_resource('velero', port_forwards='2345')
+    k8s_resource('velero', port_forwards = '2345')
     # TODO: Need to figure out how to apply port forwards for all restic pods
 if settings.get("enable_restic"):
     k8s_yaml('tilt-resources/restic.yaml')
 if settings.get("create_backup_locations"):
     k8s_yaml('tilt-resources/velero_v1_backupstoragelocation.yaml')
 if settings.get("setup-minio"):
-    k8s_yaml('examples/minio/00-minio-deployment.yaml', allow_duplicates=True)
+    k8s_yaml('examples/minio/00-minio-deployment.yaml', allow_duplicates = True)
 
 # By default, Tilt automatically allows Minikube, Docker for Desktop, Microk8s, Red Hat CodeReady Containers, Kind, K3D, and Krucible.
 allow_k8s_contexts(settings.get("allowed_contexts"))
 default_registry(settings.get("default_registry"))
-local_goos = str(local("go env GOOS", quiet=True, echo_off=True)).strip()
-git_sha = str(local("git rev-parse HEAD", quiet=True, echo_off=True)).strip()
+local_goos = str(local("go env GOOS", quiet = True, echo_off = True)).strip()
+git_sha = str(local("git rev-parse HEAD", quiet = True, echo_off = True)).strip()
 
 tilt_helper_dockerfile_header = """
 # Tilt image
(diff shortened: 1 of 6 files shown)


hack/build.sh
```diff
@@ -51,6 +51,11 @@ if [[ -z "${GIT_TREE_STATE}" ]]; then
     exit 1
 fi
 
+GCFLAGS=""
+if [[ ${DEBUG:-} = "1" ]]; then
```
zubron, line 55: Is there a better way to make a boolean flag in bash?

carlisia:
Tip for anyone trying to test this:

Be sure to copy all or the bits in the examples folder that contain these changes.

If you use vscode, use this configuration for the debugger:

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Delv",
            "type": "go",
            "request": "attach",
            "mode": "remote",
            "cwd": "${workspaceFolder}",
            "remotePath": "",
            "port": 2345,
            "host": "127.0.0.1",
            "showLog": true,
            "trace": "verbose"
        }
    ]
}
```

It worked perfectly for me!

Tiltfile
```diff
@@ -43,42 +48,59 @@ default_registry(settings.get("default_registry"))
...
+    cmd='cd ' + '.' + ';mkdir -p _tiltbuild;PKG=. BIN=velero GOOS=linux GOARCH=amd64 GIT_SHA=' + git_sha + ' VERSION=main GIT_TREE_STATE=dirty OUTPUT_DIR=_tiltbuild ' + get_debug_flag() + ' ./hack/build.sh',
+    deps=["cmd", "internal", "pkg"],
+    ignore=["pkg/cmd"],
 )
 
 local_resource(
     "velero_local_binary",
-    cmd = 'cd ' + '.' + ';mkdir -p _tiltbuild/local;PKG=. BIN=velero GOOS=' + local_goos + ' GOARCH=amd64 GIT_SHA=' + git_sha + ' VERSION=main GIT_TREE_STATE=dirty OUTPUT_DIR=_tiltbuild/local ./hack/build.sh',
-    deps = ["internal", "pkg/cmd"],
+    cmd='cd ' + '.' + ';mkdir -p _tiltbuild/local;PKG=. BIN=velero GOOS=' + local_goos + ' GOARCH=amd64 GIT_SHA=' + git_sha + ' VERSION=main GIT_TREE_STATE=dirty OUTPUT_DIR=_tiltbuild/local ./hack/build.sh',
```
carlisia, line 102: Let's add the debug config to the local binary too so we can debug the CLI.

site/content/docs/main/tilt.md
```diff
@@ -141,6 +148,25 @@ to learn what field/value pairs are required for your provider's credentials. Th
 
 There is a sample credentials file properly formatted for a MinIO storage credentials in `velero/tilt-resources/examples/cloud`.
 
+### Configure debugging with Delve
+If you would like to debug the Velero process, you can enable debug mode by setting the field `enable_debug` to `true` in your `tilt-resources/tile-settings.json` file.
+This will enable you to debug the process using [Delve](https://github.com/go-delve/delve).
+By enabling debug mode, the Velero executable will be built in debug mode (using the flags `-gcflags="-N -l"` which disables optimizations and inlining), and the process will be started in the Velero deployment using [`dlv exec`](https://github.com/go-delve/delve/blob/master/Documentation/usage/dlv_exec.md).
+
+The debug server will accept connections on port 2345 and Tilt is configured to forward this port to the local machine.
+To connect to the session, you can use the Delve CLI locally by running `dlv connect 127.0.0.1:2345`. See the [Delve CLI documentation](https://github.com/go-delve/delve/tree/master/Documentation/cli) for more guidance on how to use Delve.
```
carlisia, line 157: Maybe it's good to add a tiny note to run this **after** starting up Tilt.

carlisia (changes requested):
It works perfectly! 🎉 

Some minor stuff.

Tiltfile
```diff
@@ -1,35 +1,40 @@
...
+    "debug_continue_on_start": True,  # Continue the velero process by default when in debug mode
+    "create_backup_locations": False,
+    "setup-minio": False,
 }
 
 # global settings
 settings.update(read_json(
     "tilt-resources/tilt-settings.json",
-    default = {},
+    default={},
```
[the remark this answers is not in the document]
zubron: No, I just did a "reformat" in my editor and it applied the PEP8 standard to remove the spaces around keyword args. Starlark prefers spaces around the `=` in keyword args so I will revert this! I'll keep some of the other formatting changes though (e.g. removing spaces at ends of lines).

Tiltfile
```diff
@@ -43,42 +48,59 @@ default_registry(settings.get("default_registry"))
...
 # Support live reloading with Tilt
 RUN wget --output-document /restart.sh --quiet https://raw.githubusercontent.com/windmilleng/rerun-process-wrapper/master/restart.sh  && \
     wget --output-document /start.sh --quiet https://raw.githubusercontent.com/windmilleng/rerun-process-wrapper/master/start.sh && \
     chmod +x /start.sh && chmod +x /restart.sh
 """
 
 additional_docker_helper_commands = """
+# Install delve to allow debugging
+RUN go get github.com/go-delve/delve/cmd/dlv
+RUN which dlv
```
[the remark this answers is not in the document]
zubron: Nope, this is leftover from checking the path for the `dlv` binary :) Will remove!

carlisia (approved):
This is awesome! 🎉

jenting (commented):
I have only one concern, will this enhancement release at 1.5.x?
If not, then I think we don't need to update doc v1.5/title.md.

Tiltfile
```diff
@@ -1,25 +1,27 @@
...
-   "default_registry": "",
-   "enable_restic": False,
-   "create_backup_locations": False, 
-   "setup-minio": False, 
+    "default_registry": "",
+    "enable_restic": False,
+    "enable_debug": False,
+    "debug_continue_on_start": True,  # Continue the velero process by default when in debug mode
+    "create_backup_locations": False,
+    "setup-minio": False,
```
nrb, line 24: Does it make sense to standardize these so that `setup-minio` becomes `setup_minio`?

hack/build.sh
nrb: I don't believe there is, but my bash knowledge is admittedly incomplete and sporadic.

site/content/docs/main/tilt.md
```diff
@@ -141,6 +148,26 @@ to learn what field/value pairs are required for your provider's credentials. Th
 
 There is a sample credentials file properly formatted for a MinIO storage credentials in `velero/tilt-resources/examples/cloud`.
 
+### Configure debugging with Delve
+If you would like to debug the Velero process, you can enable debug mode by setting the field `enable_debug` to `true` in your `tilt-resources/tile-settings.json` file.
```
nrb, line 152: Could this flag also enable debug logging, or is that too much noise do you think?

nrb (changes requested):
I agree with Jenting, we likely don't need the v1.5 doc unless we're planning on backporting this.

Other than that this looks super useful and I just have a couple questions that I don't consider blockers.

zubron:
@nrb @jenting I included the docs in the 1.5 branch as that serves as our "latest" docs and we have already included the existing Tilt instructions there (which were added post 1.5). Without this, it would require navigating to the "main" version of our docs. You're right though that this isn't backported to the 1.5 branch. It's a bit awkward but it made sense to me that the latest version of our developer documentation be in the default "latest" view.

site/content/docs/main/tilt.md
zubron: It could, but this is also controlled in the deployment yaml file that is used by Tilt. Perhaps another option would be to instead have a "log_level" option in the tilt settings so that it can be managed in the same file, but can be configured independently.

nrb: Putting a separate logging flag in works for me. I'd hazard to say debug logging should be on by default in Tilt setups, but I also think that discussion might be better suited elsewhere.

nrb:
> It's a bit awkward but it made sense to me that the latest version of our developer documentation be in the default "latest" view.

That's a good point. If you're developing on Velero, you'd like want this documentation front and center, and not have to look at a different branch of the docs to get started. That's a bit of a disconnect I don't know how to bridge at the moment. At the very least, we'd have to have a link to the `main` branch of docs.

Given that point, I think it would be fine to leave it in the v1.5 docs for the time being.

nrb:
@zubron I'm good with this as is if you'd like to resolve the Tiltfile conflict

nrb (approved):

jenting (approved):

PR #3190, opened 2020-12-15
Add credential field to the bsl

### LINKED ISSUE #2425: Improve the `velero backup-location` command
- [x] --cacert [set|create] https://github.com/vmware-tanzu/velero/pull/3167
- [x] --default [set|create|get] https://github.com/vmware-tanzu/velero/pull/3092
- [ ] --credentials [set|create] https://github.com/vmware-tanzu/velero/pull/3190

carlisia:
Partially closes #2425.

Notes:
- I made this optional, is this what we want? If not, we can instruct ppl to set it up with the currently hard coded info
- I couldn't think of any validation to add, let me know if there's any you can think of


carlisia:
Not sure if this PR should be scoped to what is in it right now (populating the BSL with the secret), or to push the secret info down to the plugin.

zubron:
> Not sure if this PR should be scoped to what is in it right now (populating the BSL with the secret), or to push the secret info down to the plugin.

I don't think we came to a decision on how to make the secret data available to the plugins (at least that's what I am updating the design doc with) so I think what is here is good for now!

zubron:
> * I made this optional, is this what we want? If not, we can instruct ppl to set it up with the currently hard coded info

I think this will need to be optional as we currently support the `--no-secret` option for providers using node based auth (e.g. [AWS kube2iam](https://github.com/vmware-tanzu/velero-plugin-for-aws#install-and-start-velero))

> * I couldn't think of any validation to add, let me know if there's any you can think of

If you try to create a BSL with a secret that doesn't exist, are any errors thrown when creating/updating the BSL? It's not clear to me if the API will do some validation by using that Secret selector type. I can't think of any other validation that would be needed at the point of adding it. I assume the credential will be checked for validity at the point when we try to use it like our current approach with credentials.

carlisia:
> If you try to create a BSL with a secret that doesn't exist, are any errors thrown when creating/updating the BSL? 

This would be a good validation. But we are letting ppl create BSLs w/o any type of validation as to if it's a valid store, with it's connection and all. So I say let's add that validation to the BSL controller.

It's probably best to add that at the time when we add the code to use it. It will become part of the validation for the BSL, in the controller: does the secret exist? Yes? Then instantiate the store passing all the current info + the secret, if there is one (*passing TBD). If we get back an instance, then mark the BSL as valid.

pkg/apis/velero/v1/backupstoragelocation_types.go
```diff
@@ -30,6 +31,10 @@ type BackupStorageLocationSpec struct {
 	// +optional
 	Config map[string]string `json:"config,omitempty"`
 
+	// Credential contains the credential information intended to be used with this location
+	// +optional
+	Credential *corev1api.SecretKeySelector `json:"credential"`
```
jenting, line 36: Should it `json:"credential,omitempty"`?

pkg/cmd/cli/backuplocation/set.go
```diff
@@ -106,6 +112,17 @@ func (o *SetOptions) Run(c *cobra.Command, f client.Factory) error {
 	}
 
 	location.Spec.Default = o.DefaultBackupStorageLocation
+
+	for k, v := range o.Credential.Data() {
```
jenting, line 116: Could the credential flag allow to pass/ multiple key=value, such as `key1=value1,key2=values2`?
If not, should we inform the user of the correct way to use this flag?

carlisia: I was assuming it's 1-to-1 from the conversations and design doc, but let me confirm it.

@zubron do we know yet if we want 1.1 or 1.*? I'm fine with putting this on wip if we can't make this call yet.

nrb:
> I made this optional, is this what we want? If not, we can instruct ppl to set it up with the currently hard coded info

Yes, new fields should be optional as much as possible in order to retain backwards compatibility.

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -86,6 +89,7 @@ func NewCreateOptions() *CreateOptions {
 func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Provider, "provider", o.Provider, "Name of the backup storage provider (e.g. aws, azure, gcp).")
 	flags.StringVar(&o.Bucket, "bucket", o.Bucket, "Name of the object storage bucket where backups should be stored.")
+	flags.Var(&o.Credential, "credential", "The one credential to be used by this location in key-value pair, where key is the secret name, and value is the secret key name. Optional.")
```
nrb, line 92: I think this could be expanded a little to be more clear.

`The credential to be used by this location as a key-value pair, where the key is the Kubernetes Secret name, and the value is the data key name within the Secret. Optional, one value only.`

I like specifying the `Kubernetes Secret` versus the data key inside of it, to be clear what we're referring to. I'm not too tied to exact verbiage, other than we distinctly identify the key/value here, since they're easy to confuse.

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -148,6 +152,13 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
 		validationFrequency = &metav1.Duration{Duration: o.ValidationFrequency}
 	}
 
+	var secretName, secretKey string
+	for k, v := range o.Credential.Data() {
```
nrb, line 156: Should we also make sure that the `len(o.Credential.Data()) == 1` here? That way we ensure there's no surprise if a user had multiple secret files in the Kubernetes Secret, we're making sure up front we're given 1 and only 1 option.

The help text says that only 1 will be used.

pkg/cmd/cli/backuplocation/set.go
```diff
@@ -51,14 +53,18 @@ func NewSetCommand(f client.Factory, use string) *cobra.Command {
...
 
 func NewSetOptions() *SetOptions {
-	return &SetOptions{}
+	return &SetOptions{
+		Credential: flag.NewMap(),
+	}
 }
 
 func (o *SetOptions) BindFlags(flags *pflag.FlagSet) {
+	flags.Var(&o.Credential, "credential", "Sets the one credential to be used by this location in key-value pair, where key is the secret name, and value is the secret key name. Optional.")
```
nrb, line 67: Hm, I wonder if we could have shared string consts for these help messages. I wouldn't say redo them all in this PR, but maybe it's something we could work towards, as repeating this in 2 files seems prone to error.

nrb (changes requested):

nrb:
> If not, we can instruct ppl to set it up with the currently hard coded info

We could also populate this info at startup for 1-2 releases.

carlisia:
Comments have been addressed, ready for review.

jenting (commented):
TBH, if the secret key is fixed to use `cloud` before, we could simplify passing the secret name to the `--credential` flag.

carlisia:
> TBH, if the secret key is fixed to use cloud before, we could simplify passing the secret name to the --credential flag.

@jenting that was before. There is a new change (https://github.com/vmware-tanzu/velero/pull/2403 and https://github.com/vmware-tanzu/velero/pull/3299/commits/8e967225319753f59f2e19c6cfa7da9fa1375cf8#diff-1c259cca045064227710d0f341606e8ba73adda92dc126782acf62e2667af5baR44-R55) that will require the user to pass the secret name and the key name, this way we can have multiple secrets.

carlisia:
This PR is still ready for review. This is a blocker for https://github.com/vmware-tanzu/velero/pull/3299 (multiple secrets).

pkg/cmd/cli/backuplocation/set.go
```diff
@@ -106,6 +112,17 @@ func (o *SetOptions) Run(c *cobra.Command, f client.Factory) error {
 	}
 
 	location.Spec.Default = o.DefaultBackupStorageLocation
+
```
jenting, line 115: Should we add the check to limit only one key-value pair?
```go
        if len(o.Credential.Data()) > 1 {
		return errors.New("--credential can only contain 1 key/value pair")
	}
```

nrb: I'd like this error too if we're just going to process one set.

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -148,6 +152,16 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
 		validationFrequency = &metav1.Duration{Duration: o.ValidationFrequency}
 	}
 
+	if len(o.Credential.Data()) > 1 {
```
zubron, line 155: I think it would be better if we could add this check into the `Validate` function above. We don't seem to have an equivalent function for the `set` command so that would be good to add and use as the place to check in the other file.

zubron (changes requested):
Thanks @carlisia! This is great, and has worked as expected for the POC :+1:  I just have one suggestion based on the comments already made around the validation of the flags.

config/crd/bases/velero.io_backupstoragelocations.yaml
```diff
@@ -87,6 +87,24 @@ spec:
...
+              description: Credential contains the credential information intended
+                to be used with this location
+              properties:
+                key:
+                  description: The key of the secret to select from.  Must be a valid
+                    secret key.
+                  type: string
+                name:
+                  description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                    TODO: Add other useful fields. apiVersion, kind, uid?'
```
nrb, line 100: Is this TODO for the description field? I think the link should be enough, IMO.

zubron: This is pulled in directly from the `LocalObjectReference` type. I don't think it's something that we can control?

nrb:
https://github.com/vmware-tanzu/velero/pull/3409 is the same change on the volume snapshot side. We should make sure they're using the same approach aside from types. I think I modeled @zubron's comment with the request for checks in the `Validate` command, though I know I have some other things to change, still.

config/crd/bases/velero.io_backupstoragelocations.yaml
nrb: Ah, you're right, I see that now. Ok, we'll live with it :)

pkg/cmd/cli/backuplocation/create.go
nrb: Since the `set` and `create` commands have separate `Options` types, we'll likely have to split the `Credentials` field out somehow to have a shared validation if that's what we want. But I otherwise agree on having a `Validate` function that does this!

zubron: When I made this comment, introducing an equivalent `Validate` function was all I had in mind so I'm happy with that approach!

zubron (dismissed):
Thank you!

nrb (approved):

pkg/cmd/cli/backuplocation/create.go
carlisia: Nolan and I paired and looking into merging some of the code from create and set, and it did not make sense.

carlisia:
@vmware-tanzu/velero-owners this release blocker is ready for review.

jenting (approved):

PR #3198, opened 2020-12-18
Modify InitContainer checking function that potentially incurs error

shellwedance:
Added an `if` conditional that checks the length of `pod.Status.InitContainerStatus` to prevent potential error.
- When length of `pod.Status.InitContainerStatus` is 0, index out of range error occurs.
- This runtime error kills Restic plugin container and gets rid of previous logs after restart.

Signed-off-by: Taeuk Kim <taeuk_kim@tmax.co.kr>

### DIFF
--- a/changelogs/unreleased/3198-shellwedance
+++ b/changelogs/unreleased/3198-shellwedance
@@ -1,0 +1,1 @@
+Add restic initContainer length check in pod volume restore to prevent restic plugin container disappear in runtime
--- a/pkg/controller/pod_volume_restore_controller.go
+++ b/pkg/controller/pod_volume_restore_controller.go
@@ -220,10 +220,11 @@ func isPodOnNode(pod *corev1api.Pod, node string) bool {
 }
 
 func isResticInitContainerRunning(pod *corev1api.Pod) bool {
-
 	// Restic wait container can be anywhere in the list of init containers, but must be running.
 	i := getResticInitContainerIndex(pod)
-	return i >= 0 && pod.Status.InitContainerStatuses[i].State.Running != nil
+	return i >= 0 &&
+		len(pod.Status.InitContainerStatuses)-1 >= i &&
+		pod.Status.InitContainerStatuses[i].State.Running != nil
 }
 
 func getResticInitContainerIndex(pod *corev1api.Pod) int {
--- a/pkg/controller/pod_volume_restore_controller_test.go
+++ b/pkg/controller/pod_volume_restore_controller_test.go
@@ -590,6 +590,26 @@ func TestIsResticContainerRunning(t *testing.T) {
 			},
 			expected: true,
 		},
+		{
+			name: "pod with restic init container with empty InitContainerStatuses should return 0",
+			pod: &corev1api.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "ns-1",
+					Name:      "pod-1",
+				},
+				Spec: corev1api.PodSpec{
+					InitContainers: []corev1api.Container{
+						{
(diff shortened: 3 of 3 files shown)


carlisia (approved):
Thank you for this, it's a good catch! 👍

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -220,10 +220,11 @@ func isPodOnNode(pod *corev1api.Pod, node string) bool {
 }
 
 func isResticInitContainerRunning(pod *corev1api.Pod) bool {
-
 	// Restic wait container can be anywhere in the list of init containers, but must be running.
 	i := getResticInitContainerIndex(pod)
-	return i >= 0 && pod.Status.InitContainerStatuses[i].State.Running != nil
+	return i >= 0 &&
+		len(pod.Status.InitContainerStatuses) > 0 &&
```
jenting, line 226: Would it better by checking `len(pod.Status.InitContainerStatuses)-1 >= i`? It's more robust to make sure the index won't out of range. 

Also, it's good to have a test case for this.

carlisia: Ah, that's a very good catch, I'm sorry I missed it.

shellwedance: @jenting Thanks for kind recommendation. But I think `len(pod.Status.InitContainerStatuses)-1 >= i` may incur runtime error.

Counterexample: multiple InitContainers but no restic InitContainer 
- Multiple InitContainers make `len(pod.Status.InitContainerStatuses)-1` larger than 0.
- No restic InitContainer makes `i` to `-1`, which is returned by `getResticInitContainerIndex`.
- As a result, `len(pod.Status.InitContainerStatuses)-1 >= i` becomes `true` and code executes to check next conditional, `pod.Status.InitContainerStatuses[i].State.Running != nil`.

Then, runtime error can occur because it accesses -1 index of `pod.Status.InitContainerStatuses`.

jenting: > No restic InitContainer makes i to -1, which is returned by getResticInitContainerIndex.

So, the `i >= 0` would be _false_, then I think the code won't execute the `len(pod.Status.InitContainerStatuses)-1 >= i` 🤔  ?

shellwedance: > So, the i >= 0 would be false,

Oh I misunderstood you suggest to remove `i >= 0` either.
Yes then, `len(pod.Status.InitContainerStatuses)-1 >= i` is more robust than `len(pod.Status.InitContainerStatuses) > 0` with `i >= 0`.
Modifying that line.

jenting (approved):
LGTM

changelogs/unreleased/3198-shellwedance
```diff
@@ -0,0 +1 @@
+Added an if conditional that checks the length of `pod.Status.InitContainerStatus` to prevent potential error.
```
nrb, line 1: Can you reword this message to be less about the code and more about the user impact? What's the potential error that's being fixed that users should know about?

nrb (changes requested):
Thanks for catching this! I only request one small change on the changelog message, otherwise this looks good to me!

changelogs/unreleased/3198-shellwedance
shellwedance: Right it seems code-friendly rather than user. 
Modified.

jenting (approved):
Thank you.

shellwedance:
Change reflected @nrb

nrb (approved):

PR #3208, opened 2020-12-28
Hidden the velero create/delete/describe/get sub-commands

jenting:
According to the [CLI design doc](https://github.com/vmware-tanzu/velero/blob/master/design/cli-install-changes.md#commands), the Velero organization commands with the format:
```
velero [resource] [operation] [flags]
```

Therefore, hidden the `velero create/delete/describe/get` sub-commands.
Now, with this PR, the velero available sub-commands are
```
Usage:
  velero [command]

Available Commands:
  backup            Work with backups
  backup-location   Work with backup storage locations
  bug               Report a Velero bug
  client            Velero client related commands
  completion        Output shell completion code for the specified shell (bash or zsh).
  help              Help about any command
  install           Install Velero
  plugin            Work with plugins
  restic            Work with restic
  restore           Work with restores
  schedule          Work with schedules
  snapshot-location Work with snapshot locations
  version           Print the velero version and associated image
```

### DIFF
--- a/site/content/docs/main/contributions/oracle-config.md
+++ b/site/content/docs/main/contributions/oracle-config.md
@@ -39,9 +39,6 @@ Velero is a tool for managing disaster recovery, specifically for Kubernetes
 cluster resources. It provides a simple, configurable, and operationally robust
 way to back up your application state and associated data.
 
-If you're familiar with kubectl, Velero supports a similar model, allowing you to
-execute commands such as 'velero backup get' and 'velero schedule create'.
-
 Usage:
   velero [command]
 ```
--- a/site/content/docs/main/contributions/oracle-config.md
+++ b/site/content/docs/main/contributions/oracle-config.md
@@ -40,8 +40,7 @@ cluster resources. It provides a simple, configurable, and operationally robust
 way to back up your application state and associated data.
 
 If you're familiar with kubectl, Velero supports a similar model, allowing you to
-execute commands such as 'velero get backup' and 'velero create schedule'. The same
-operations can also be performed as 'velero backup get' and 'velero schedule create'.
+execute commands such as 'velero backup get' and 'velero schedule create'.
 
 Usage:
   velero [command]
--- a/changelogs/unreleased/3208-jenting
+++ b/changelogs/unreleased/3208-jenting
@@ -1,0 +1,1 @@
+Hidden the `velero create/delete/describe/get` sub-commands
--- a/pkg/cmd/cli/create/create.go
+++ b/pkg/cmd/cli/create/create.go
@@ -29,9 +29,10 @@ import (
 
 func NewCommand(f client.Factory) *cobra.Command {
 	c := &cobra.Command{
-		Use:   "create",
-		Short: "Create velero resources",
-		Long:  "Create velero resources",
+		Use:    "create",
+		Short:  "Create velero resources",
+		Long:   "Create velero resources",
+		Hidden: true,
 	}
 
 	c.AddCommand(
--- a/pkg/cmd/cli/delete/delete.go
+++ b/pkg/cmd/cli/delete/delete.go
@@ -28,9 +28,10 @@ import (
 
 func NewCommand(f client.Factory) *cobra.Command {
 	c := &cobra.Command{
-		Use:   "delete",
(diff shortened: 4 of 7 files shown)


site/content/docs/main/contributions/oracle-config.md
```diff
@@ -40,8 +40,7 @@ cluster resources. It provides a simple, configurable, and operationally robust
 way to back up your application state and associated data.
 
 If you're familiar with kubectl, Velero supports a similar model, allowing you to
-execute commands such as 'velero get backup' and 'velero create schedule'. The same
-operations can also be performed as 'velero backup get' and 'velero schedule create'.
+execute commands such as 'velero backup get' and 'velero schedule create'.
```
zubron, line 43: It looks like this part of the documentation is copied directly from the output of `velero help`: https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/velero/velero.go#L68-L70

That text will need to be updated. In fact, we may want to remove it as kubectl doesn't support commands in the format `kubectl <resource> <verb>`.

zubron (changes requested):
This looks good but the long help text also needs to be updated.

site/content/docs/main/contributions/oracle-config.md
jenting: nice catch!

zubron (approved):

carlisia (commented):
I'm not sure we want to hide these sub command. The CLI design doc also states:

`commands will also work by swapping the operation/resource.`.

Since we do allow for this swap, it would be the only way the user would discover this?

jenting:
> I'm not sure we want to hide these sub command. The CLI design doc also states:
> 
> `commands will also work by swapping the operation/resource.`.
> 
> Since we do allow for this swap, it would be the only way the user would discover this?

Yes, agree with you. If the user would like to discover it, these commands should not be hidden.

However, since the CLI design doc proposes to add several new sub-commands. If the deprecated sub-commands/flags did not hide, it makes the CLI more complicated to use and discover.

From my perspective, we could release Velero CLI revamp at 1.6 with _hide deprecated sub-commands/flags_ and outputs the _DEPRECATED_ message when the user uses the hidden command. Then at Velero 2.0, we drop the deprecated commands/flags.

WDYT?

zubron:
I like being able to swap the verb and resource and although this change hides it, I don't think it should ultimately be removed, especially since the design doc explicitly mentions that they can be swapped. The current CLI is inconsistent though in how the swapping works, for example, `backup` also has `logs` and `download` which can't be swapped, and I sometimes get caught out by these. Should we be aiming to allow all combinations to be swapped?

carlisia:
> Should we be aiming to allow all combinations to be swapped?

Yes.

zubron:
Perhaps we could try to create groups of commands, e.g. the resources that velero can operate on `backup`, `restore`, etc. and the operations that velero can perform `create`, `delete`. That might help with the discovery of the operation command while making it clearer the types of resources that velero can operate on.

nrb:
Re-reading the [design doc](https://github.com/vmware-tanzu/velero/blob/main/design/cli-install-changes.md), create/delete/get/describe aren't being deprecated. Only `velero install` was being deprecated, once all the options are moved from `velero install` to the appropriate new command. Since that's not complete yet, I don't think we should hide `velero install` yet either.

Based on the document, the following commands should be implemented or updated before we can hide `velero install`:

 * New: 
   * `velero config (server|restic)`
   * `velero (snapshot-location|backup-location) set`
 * To update with flags currently on `velero install`:
   * `velero snapshot-location create` (or `velero create snapshot-location`)
   * `velero backup-location create` (or `velero create backup-location`)
   * `velero plugin`
 
`velero server --default-backup-storage-location` to `velero backup-location set --default` is already done based on a previous PR from @jenting. 

I think we can look into the discoverability of commands, but I think the focus of the design was _only_ for installation and configuration of Velero, not about all the CLI commands. In particular, there's this non-goal:

> Propose changes to the CLI that go beyond the functionality of install and configure

So I think this proposed change is not really in line with the design doc. I'd prefer that we make sure all commands work in either order before hiding things to clean up the help text.

jenting:
Close it first until we have done the CLI revamp.

PR #3275, opened 2021-01-14
Add colors to describe command

### LINKED ISSUE #1082: Colored output
This is a branch off of #987.

We should decide if we want colored output and, if so, which output should be colored vs not.

mike1808:
* Add colors to describe backups/restore/schedules commands
* Make name in the output bold
* Disable colors via `--colorized` flag or if velero isn't in TTY

Fixes #1082 

If you have any other coloring suggestion please let us know.

cc @KauzClay


jenting (changes requested):
Please run `make update` to fix CI failure. Thank you.

nrb (changes requested):
This is really cool, thank you!

I have an additional request - can you add enabling/disabling this functionality into the Velero client config file, so folks can disable it for all runs? The file lives in `$HOME/.config/velero`, and [we have helper code for it that will need to read the key](https://github.com/vmware-tanzu/velero/blob/main/pkg/client/config.go). We should have existing code for the feature flags that allows for using the config file and command line flags.

KauzClay:
Hey @nrb ,

When you say you want it to be enabled/disabled via the config file, do you imagine it being part of "features", like this?
`{"features": "colors"}`

Or were you imagining a separate config field entirely, like this?
`{"features":"foo","colors":"off"}`

nrb:
> Or were you imagining a separate config field entirely, like this?

Yep! `features` are for feature flags that get passed from the client to the server to the plugins. Since this is purely clientside, I think it makes sense to be a dedicated field.

carlisia (approved):
This LGTM! 🎉

carlisia:
Hey @mike1808, your last commit was unsigned and so it's failing a check.

jenting (changes requested):
I disabled the colorized in my $HOME/.config/velero/config.json as `{"colorized":"false"}`.
However, the color still existed for the command output.

mike1808:
@jenting please use boolean values instead of strings.

jenting:
> @jenting please use boolean values instead of strings.

I think the problem comes when I running `velero client config set colorized=false`, the $HOME/.config/velero/config.json would be `{"colorized":"false"}`.

If I manually edit $HOME/.config/velero/config.json to  `{"colorized":false}`, the color is out.
But if I running `velero client config get`, if shows `colorized: %!s(bool=false)`.

KauzClay:
@jenting I see, is there a preference then for all config values to be strings then? We could change the value to a string and it could be on/off instead of true/false?

Or we could do some converting in the VeleroConfig struct to allow for the bool. Do you have a preference?

jenting:
> @jenting I see, is there a preference then for all config values to be strings then? We could change the value to a string and it could be on/off instead of true/false?

cc @nrb @carlisia @zubron any suggestion on this?

Personally, I'd prefer not to break the change (keep using the `strings` type).
Either "on"/"off" or "true"/"false" is fine to me.

> 
> Or we could do some converting in the VeleroConfig struct to allow for the bool. Do you have a preference?

KauzClay:
@jenting okay, you should be able to use `velero client config set colorized=false` and see that the color doesn't show now.

As a note, we made it so that the command line flag `--colorized=true` will override whatever is in the client config.

jenting:
Thank you. Two comments left:
1. It'd be better to add doc for this nice feature
2. Please fix the conflicting files

KauzClay:
Hi @jenting , I think we resolved the conflicting files. As for the doc, we weren't sure where to put this information. We opted for `site/content/docs/main/customize-installation.md`, since we saw other references to client config there. If that isn't okay, could you please explain where you were envisioning this documentation to live?

jenting:
> We opted for `site/content/docs/main/customize-installation.md`, since we saw other references to client config there.

I'm okay to put it here.

I saw the last two commits had signed off, but the CI failed at DCO check, strange 🤔

site/content/docs/main/customize-installation.md
```diff
@@ -70,6 +70,21 @@ All client side feature flags may be disabled using the below command
...
+Velero CLI uses colored output for some commands, such as `velero desribe`. If
+the environment in which Velero is run doesn't support colored output, the
+colored output will be automatically disabled. However, you can manually disable
+colors with config file:
+
+```bash
+velero client config set colorized=false
+```
+
+Note, that if you specify `--colorized=true` as a CLI option it will override
```
carlisia, line 84:
```suggestion
Note that if you specify `--colorized=true` as a CLI option it will override
```

site/content/docs/main/customize-installation.md
```diff
@@ -70,6 +70,21 @@ All client side feature flags may be disabled using the below command
 velero client config set features=
 ```
 
+### Colored CLI output
+
+Velero CLI uses colored output for some commands, such as `velero desribe`. If
```
jenting, line 75:
```suggestion
Velero CLI uses colored output for some commands, such as `velero describe`. If
```

carlisia:
@jenting thanks for testing this.

@KauzClay this looks great and ready to merge as far as I'm concerned except for:
- the biggest/tinniest nit pick ever (see above)
- we need to make the DCO check pass

Thank you.

nrb:
@KauzClay Once you make the two changes asked for, please sign the commits as shown at https://github.com/vmware-tanzu/velero/pull/3275/checks?check_run_id=1809391225

KauzClay:
Ah, sorry about messing up the sign off thing. Hopefully I got it this time?

jenting (approved):
LGTM, thank you.

carlisia (approved):

PR #3279, opened 2021-01-15
Site updates

A-P:
Requested updates from @jonasrosland


nrb (dismissed):

carlisia (commented):
Hm, I hate to be that person, but can we not add this change? It is so odd to have the toc on the right side. We are left to right/top to bottom readers (English language) and the TOC is the FIRST thing you want to read, if there is one. It being on the right feels like a strong anti-pattern to me. In addition, I had not noticed it until someone pointed it out. I hope my arguments here are persuasive enough to block this change.

jonasrosland:
We took the pattern of having the TOC be a separate portion, automatically generated, from other existing projects.

Here are a few examples of what we looked at:
https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/
https://doc.traefik.io/traefik/providers/overview/
https://docs.fluentd.org/quickstart/life-of-a-fluentd-event
https://prometheus.io/docs/prometheus/latest/getting_started/

Are you ok with the other changes for the community and resource page?

carlisia:
@jonasrosland looking at those sites, I am hoping this trend goes away :)

If I am the only one who dislikes it, I can live with it and maybe it will grow on me. The one redeeming quality of this TOC is that it stays put as you scroll down. Is there a way to hide it when there are no TOC elements?

I didn't realize there were changes to the community and resource pages. Those changes look great!

zubron (commented):
Huge +1 to the community and resources pages. Those pages look so much better and the presentation makes the info much easier to read and process.

I wasn't sure about TOC changes at the start but it looks like a few changes have been made since it was first proposed such as indenting sub headings which make it much clearer. I agree with @carlisia's point about hiding/changing it when there is no TOC. For example, the title for the TOC doesn't make sense on this page: https://deploy-preview-3279--velero.netlify.app/docs/v1.5/disaster-case/
Further to that, should the TOC at the top of the body of the page also be removed when the side bar is present? See: https://deploy-preview-3279--velero.netlify.app/docs/v1.5/customize-installation/

With regard to the links for making changes, is there a specific reason to have them side by side rather than being on a separate lines like the Kubernetes page? It seems like doing that would allow a little more text to describe the link e.g. "Edit this page". This is just an idea though and I'm mostly curious about the design choices.

Something that would be useful, but it's an action for the maintainer team, would be to investigate whether we could create a specific issue template for this link so that the "Documentation" label can be applied.

carlisia (commented):
@A-P let us know when it's ready for another round of review. So far it looks goo to me except that the original TOC at the top of the pages need to be removed now that we moved the TOC to the right. Not seeing anything else missing. That, and the DCO issue. Thanks.

A-P:
@jonasrosland @carlisia I just pushed up a couple of requested style updates to the right bar, and it's ready for you to review again. The TOC is still in the body in case someone is viewing on mobile, but I can easily remove it if you'd like me to.

nrb:
> Something that would be useful, but it's an action for the maintainer team, would be to investigate whether we could create a specific issue template for this link so that the "Documentation" label can be applied.

This should now be applied via a GitHub Action that matches on the `site/content/docs/**/*` file glob pattern.

nrb:
@A-P Is there not a way to use the right side bar for mobile viewers? I think the ToC at the top of the page is redundant with this change personally, but I don't want to completely exclude mobile users.

It also looks like the DCO still needs to be signed, and there's a spellcheck failure.

jonasrosland:
@nrb It's hard to have both a list for TOC and "on this page" at the same time on mobile. See the Kubernetes docs as an example, they use only the TOC in their "hamburger" menu.

PR #3299, opened 2021-01-19
POC: Load and pass credentials to plugins

zubron:
Update NewObjectBackupStore to take a CredentialsGetter which can be
used to get the credentials for a BackupStorageLocation if it has been
configured with a Credential. If the BSL has a credential, use that
SecretKeySelector to fetch the secret, write the contents to a temp file
and then pass that file through to the plugin via the config map. This
relies on the plugin being able to use the config field.

This does not yet handle VolumeSnapshotLocations or ResticRepositories.

This branch is based on some existing PRs (#3189, #3190). Only the latest
commit in this branch is new: https://github.com/vmware-tanzu/velero/pull/3299/commits/8e967225319753f59f2e19c6cfa7da9fa1375cf8


pkg/credentials/credentials.go
```diff
@@ -0,0 +1,83 @@
...
+
+type Getter interface {
+	Get(selector *corev1.SecretKeySelector) ([]byte, error)
+
+	// GetAsFile gets the secret data specified by the selector and returns a filepath
+	// where the contents have been written to
+	GetAsFile(selector *corev1.SecretKeySelector) (string, error)
+}
+
+type credentialsGetter struct {
```
nrb, line 38: Per community discussion, this may need a field for the name of the env variable that it will land in.

zubron:
Closing this as it was a draft and the work is now is being added in different PRs.

PR #3309, opened 2021-01-20
Update docs to clarify backup location and relationship with other data

### LINKED ISSUE #3193: Update our docs with guidance on storing non-velero data in buckets
> The error is happening because Velero found unexpected directories in the bucket where it will store the backup data. In this case it found the directories `backup` and `ouput` in the bucket. If storing the data in the root directory of the bucket, then it should only contain directories managed by Velero.

> If there is other content in the bucket, you can use the `--prefix` option for `velero install` to instruct velero to store all backup data within a specific directory in the bucket. Again, this directory should only be used by Velero.

_Originally posted by @zubron in https://github.com/vmware-tanzu/velero/issues/3192#issuecomment-746661149_

It's not clear from our docs whether a bucket should _only_ contain velero data or whether a bucket can be multi-purpose as long as the user uses the `--prefix` option.

a-mccarthy:
Fixes #3193 

### DIFF
--- a/site/content/docs/main/locations.md
+++ b/site/content/docs/main/locations.md
@@ -7,7 +7,7 @@ layout: docs
 
 Velero has two custom resources, `BackupStorageLocation` and `VolumeSnapshotLocation`, that are used to configure where Velero backups and their associated persistent volume snapshots are stored.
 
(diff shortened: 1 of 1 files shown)


zubron (approved):
Thanks!

jenting (approved):

PR #3310, opened 2021-01-20
Download Restic binary and copy into Tilt Velero image

zubron:
This change adds an additional set of commands to Dockerfile for the
Velero image which adds the `hack/download-restic.sh` script, installs
the necessary dependencies, and then runs that script.

In order to copy the script from the `hack` directory, the context for
building the image has been changed to the root of the velero
repository.

Fixes #3219.

Signed-off-by: Bridget McErlean <bmcerlean@vmware.com>

PR #3327, opened 2021-01-24
Download restic binary outside Tilt container

zubron:
In #3310, the Dockerfile for the Tilt Velero container was modified to
call the `./hack/download-restic.sh` script. A side effect of this
change was that the context for the docker build was much larger as it
was the root of the Velero repo, rather than just the `_tiltbuild`
directory. With the frequent rebuilds of the image that happen when
using Tilt, a large amount of disk space was being consumed by the
different layers of images builds in the Docker overlay filesystem (as
diffs could include the `.go` directory which can be several GBs).

This change modifies the `download-restic.sh` script to allow the output
directory for the restic binary to be configured. This means that the
script can be called directly from the Tiltfile and can be managed
outside the container build. This allows us to restore the previous
`_tiltbuild` context. It also speeds up image builds as we can download
restic once and use it for all builds rather than redownloading
frequently.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -72,15 +72,6 @@ COPY --from=tilt-helper /usr/bin/docker /usr/bin/docker
 COPY --from=tilt-helper /go/kubernetes/client/bin/kubectl /usr/bin/kubectl
 """
 
-docker_build_download_restic_commands = """
-COPY ./hack/download-restic.sh /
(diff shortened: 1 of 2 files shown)


jenting (approved):

carlisia (approved):
👍

PR #3329, opened 2021-01-25
Replace NewObjectBackupStore with interface

zubron:
In preparation for modifying the instantiation of `BackupStores` to be
able to load credentials, change the function `NewObjectBackupStore` to
be an interface that is passed in to all controllers.

Previously, the function to get a new backup store was configurable but
for many controllers was fixed to use `NewObjectBackupStore`. This
change introduces an interface for getting the backup store and wraps
the functionality from `NewObjectBackupStore` in a type which implements
this interface. This will allow more flexibility when introducing
credentials for a specific backup store as it will allow us to create a
new `ObjectBackupStoreGetter` type which can be configured to add
credentials config when creating the ObjectBackupStore without needing
to change the API used by the controllers.

### DIFF
--- a/pkg/persistence/object_store.go
+++ b/pkg/persistence/object_store.go
@@ -90,7 +90,21 @@ type ObjectStoreGetter interface {
 	GetObjectStore(provider string) (velero.ObjectStore, error)
 }
 
-func NewObjectBackupStore(location *velerov1api.BackupStorageLocation, objectStoreGetter ObjectStoreGetter, logger logrus.FieldLogger) (BackupStore, error) {
+// ObjectBackupStoreGetter is a type that can get a velero.BackupStore for a
+// given BackupStorageLocation and ObjectStore.
+type ObjectBackupStoreGetter interface {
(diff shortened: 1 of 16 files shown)


jenting (approved):
cool, thank you.

pkg/persistence/object_store.go
```diff
@@ -90,7 +90,21 @@ type ObjectStoreGetter interface {
...
+type ObjectBackupStoreGetter interface {
+	Get(location *velerov1api.BackupStorageLocation, objectStoreGetter ObjectStoreGetter, logger logrus.FieldLogger) (BackupStore, error)
+}
+
+type objectBackupStoreGetter struct{}
+
+// NewObjectBackupStoreGetter returns a ObjectBackupStoreGetter that can get a
+// default velero.BackupStore.
+func NewObjectBackupStoreGetter() ObjectBackupStoreGetter {
+	return &objectBackupStoreGetter{}
```
nrb, line 104: Per Bridget, this will be more fleshed out in a later PR.

nrb (approved):
This is basically moving a single function into an interface for further work.

PR #3354, opened 2021-01-26
Add e2e scaffolding using Testify Suites

codegold79:
Per @nrb's review of PR #3133, this PR splits out the e2e Testify Suites framework and supporting code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -335,3 +335,7 @@ gen-docs:
 .PHONY: test-e2e
 test-e2e: local
 	$(MAKE) -C test/e2e run
+
+.PHONY: test-e2e-testify
+test-e2e-testify: local
+	$(MAKE) -C test/e2e_testify run
--- a/test/e2e_testify/common_test.go
+++ b/test/e2e_testify/common_test.go
@@ -1,0 +1,263 @@
+package e2etestify
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	veleroexec "github.com/vmware-tanzu/velero/pkg/util/exec"
+	"github.com/vmware-tanzu/velero/test/e2e"
+)
+
+func installVelero(ctx context.Context) error {
+	if err := e2e.EnsureClusterExists(ctx); err != nil {
+		return errors.Wrap(err, "check cluster exists")
+	}
+
+	// Pass global variables to option parameters.
+	options, err := e2e.GetProviderVeleroInstallOptions(
+		*pluginProvider,
+		*cloudCredentialsFile,
+		*bslBucket,
+		*bslPrefix,
+		*bslConfig,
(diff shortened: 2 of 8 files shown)


test/e2e_testify/common_test.go
```diff
@@ -0,0 +1,263 @@
...
+		return []string{"velero/velero-plugin-for-aws:v1.1.0", "velero/velero-plugin-for-vsphere:v1.0.2"}
+	default:
+		return []string{""}
+	}
+}
+
+func InstallCRD(ctx context.Context, crdFile, ns string) error {
+	fmt.Printf("Install CRD %s.\n", crdFile)
+
+	cmd := exec.CommandContext(ctx, "kubectl", "apply", "-f", crdFile)
```
zubron, line 66: This can be addressed in #3133, but I think it would be better to specify the namespace where the CRDs are being installed using the `ns` value passed in rather than having to know the namespace specified in the YAML.

Makefile
```diff
@@ -335,3 +335,7 @@ gen-docs:
 .PHONY: test-e2e
 test-e2e: local
 	$(MAKE) -C test/e2e run
+
+.PHONY: test-e2e-testify
+test-e2e-testify: local
```
zubron, line 340: I don't think that we should have a separate target for running these tests. Can we combine it all into the `test-e2e` target? Otherwise, I think it would be really easy to miss running one of them.

zubron (changes requested):
Thanks for splitting this out into a separate PR, @codegold79!

I think it will be nice to have the option on how to write tests (using testify vs ginkgo) but I'm concerned about the long term maintainability of having two separate build/run processes. I'm not familiar with how ginkgo tests run, but is it possible to have both regular go tests and ginkgo tests in one package? If yes, then I think that ultimately we should move any tests using that approach *inside* the `e2e` package rather than creating a new one. Otherwise, it will be very easy for these to get out of sync and increase the maintenance burden. Also, I think that some of the functionality added in this package will be useful for the main `e2e` package :)

I don't want to block your progress on this though, @codegold79, so I think we can address the overall structure of the e2e packages in a follow up and I don't want to consider it a blocker for this PR. I mostly just want to make sure that we've considered the impact on the rest of the e2e tests moving forward as there is a lot of work happening in that area.

@dsu-igeek What are your thoughts on this?

Makefile
codegold79: Answer for future reference because I'm closing this PR. 

I'm afraid when I put the testify and ginkgo test files together, it caused problems and the tests were not able to run using `go test`.

codegold79:
Thank you, @zubron, for your review. I completely agree that it is difficult to maintain a parallel set of code (i.e. duplicated) and already some of my code is outdated.

I had wanted to provide an example of using Testify Suites for e2e tests so people would have something to compare with Ginkgo in Velero. However, per @nrb in a code review meeting for PR #3133, given the limited resources (including time), we won't be rewriting the Ginkgo tests. So I'll write tests for PR #3133 to work with the existing Ginkgo framework. I have copied over the Testify Suites code to this PR and will close the PR.

PR #3362, opened 2021-01-29
Update docs for running velero locally

### LINKED ISSUE #3181: Documentation for running locally does not include instructions for providing plugin binaries
There is an issue with [our "Run Velero locally in development" docs](https://velero.io/docs/v1.5/run-locally/) as it states that the `--plugin-dir` flag is optional, but doesn't state that the plugin binaries need to be present on the machine where you are running the velero server binary.

_Originally posted by @zubron in https://github.com/vmware-tanzu/velero/issues/3178#issuecomment-744524923_

MadhavJivrajani:
- Updated the docs to specify that the `--plugin-dir` flag should be set to a directory that has the binary present locally present. 
- Changes made for v1.5, should I make the changes for previous versions as well?

Fixes #3181


zubron:
Hi @MadhavJivrajani - thanks for fixing this! Please also add this change to the same file in the [main directory](https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/main/run-locally.md) as this ensures that the change will be available in all future docs versions.

PR #3399, opened 2021-02-03
Support cli uninstall

### LINKED ISSUE #3291: Support uninstalling velero using velero cli
We have `velero install ...` to install velero in a cluster but we don't have a corresponding `velero uninstall ...` option. 
To uninstall velero you have to manually delete the velero Deployment, namespace and CRDs (ref: https://github.com/vmware-tanzu/velero/issues/2567)

I think it would be great if we could have `velero uninstall ...` option in the cli to automate the above. 

**Environment:**

```
Client:
	Version: v1.5.3
	Git commit: 123109a3bcac11dbb6783d2758207bac0d0817cb
Server:
	Version: v1.5.3
```

**Vote on this issue!**

vadasambar:
Attempts to fix #3291.

### DIFF
--- a/pkg/cmd/cli/uninstall/uninstall.go
+++ b/pkg/cmd/cli/uninstall/uninstall.go
@@ -29,17 +29,17 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 
+	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+
 	"github.com/vmware-tanzu/velero/pkg/client"
 	"github.com/vmware-tanzu/velero/pkg/cmd"
 	"github.com/vmware-tanzu/velero/pkg/cmd/util/output"
 	"github.com/vmware-tanzu/velero/pkg/install"
 	"github.com/vmware-tanzu/velero/pkg/util/kube"
-	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 )
 
 // Uninstall uninstalls all components deployed using velero install command
-func Uninstall(ctx context.Context, client *kubernetes.Clientset, extensionsClient *apiextensionsclientset.Clientset,
-	veleroNamespace string) error {
+func Uninstall(ctx context.Context, client *kubernetes.Clientset, extensionsClient *apiextensionsclientset.Clientset, veleroNamespace string) error {
 	if veleroNamespace == "" {
 		veleroNamespace = "velero"
 	}
--- a/pkg/cmd/cli/uninstall/uninstall.go
+++ b/pkg/cmd/cli/uninstall/uninstall.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 the Velero contributors.
+Copyright the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


vadasambar:
Will add the ChangeLog and DCO once I am done with my changes.

vadasambar:
I've re-used parts from `install` package which adds a dependency on the `install` package. It'd be great if someone could review this approach. Some of the work that still needs to be done
1. Use the namespace provided in the cmd
2. Remove hardcoding to `velero`
3. Logs to stdout mentioning which resource is being deleted

The uninstallation deletes resource in the reverse order of their installation.

jenting (commented):
Thank you for your contribution, I like the idea.

However, the "velero install" command is going to be deprecated, so I'm not sure is the uninstall command should be added.

nrb:
@jenting I think we'll take this as it's not that major. `velero install` may be going away, but I like having the the easy command.

@vadasambar This code looks good, but it could be simplified. Take a look at https://velero.io/docs/v1.5/uninstalling/

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -0,0 +1,117 @@
+/*
+Copyright 2020 the Velero contributors.
```
nrb, line 2: We can leave out the year now

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -0,0 +1,117 @@
...
+	if err := install.AppendUnstructured(resources, se); err != nil {
+		return err
+	}
+
+	sa := install.ServiceAccount(veleroNs, map[string]string{})
+	if err := install.AppendUnstructured(resources, sa); err != nil {
+		return err
+	}
+
+	crb := install.ClusterRoleBinding(veleroNs)
```
nrb, line 71: Everything above this line is unnecessary, as the namespaced resources will be deleted with the namespace.

Also, we need to add the CustomResourceDefinitions. You can select them similarly to [this](https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/cli/plugin/helpers.go#L33)

nrb (changes requested):
@vadasambar A few changes that need to be made, but I think they're fairly similar.

pkg/cmd/cli/uninstall/uninstall.go
vadasambar: @nrb , makes sense. I am using an existing helper function to get all the CRDs 
https://github.com/vmware-tanzu/velero/blob/827bad0ecff0f00d80fdd39c1d06e2ff5722c8b7/pkg/cmd/cli/uninstall/uninstall.go#L36

Do you think using labels would be a better approach?

nrb: Yeah, the `AllCRDs` function was really meant to be used on installation only, not necessarily re-used as a selector.

Like you said, version skew could make this inaccurate.

vadasambar: Yeah. Will make the change then.

dsu-igeek: Please take a look at the VeleroUninstall in https://github.com/vmware-tanzu/velero/blob/main/test/e2e/velero_utils.go

We can expand on this as necessary, but it is pretty much a straight port of the existing hand uninstall process.  
When "velero uninstall" is available in the CLI, the plan is to call that from the test code.

vadasambar: @dsu-igeek , I think we can port the uninstall code directly for the uninstall command. 

>  When "velero uninstall" is available in the CLI, the plan is to call that from the test code.
Is this so that we could test the cli or is it okay to directly call the uninstall function? Asking this because I see we are doing the latter for `VeleroInstall` in `test/e2e/velero_utils.go`

vadasambar:
Made most of the requested changes. @dsu-igeek, I have ported uninstall code from the test suite. Can you please take a look at the code whenever you find time?
Also, 
https://github.com/vmware-tanzu/velero/pull/3399/files#diff-3ef9006b53436d6a87171de888d0ead4022b7e30b6fc4f29d30e36a1e24643bbR86

jenting:
Please run `make update`

pkg/cmd/cli/uninstall/uninstall.go
dsu-igeek: We should eventually call the CLI both for install and uninstall.  That way we can test the CLI on those paths as part of the testing.

test/e2e/backup_test.go
```diff
@@ -41,7 +42,7 @@ var _ = Describe("[Restic] Velero tests on cluster using the plugin provider for
 	AfterEach(func() {
 		if installVelero {
 			timeoutCTX, _ := context.WithTimeout(context.Background(), time.Minute)
-			err := VeleroUninstall(timeoutCTX, client, extensionsClient, veleroNamespace)
+			err := uninstall.Uninstall(timeoutCTX, client, extensionsClient, veleroNamespace)
```
dsu-igeek, line 45: Let's leave VeleroUninstall as a function we can call.

test/e2e/enable_api_group_versions_test.go
```diff
@@ -342,7 +344,7 @@ func RunEnableAPIGroupVersionsTests(ctx context.Context, resource, group string,
 
 		// Uninstall Velero
 		if installVelero {
-			err = VeleroUninstall(ctx, client, extensionsClient, veleroNamespace)
+			err = uninstall.Uninstall(ctx, client, extensionsClient, veleroNamespace)
```
dsu-igeek, line 347: Same here, let's continue to call VeleroUninstall.

test/e2e/velero_utils.go
```diff
@@ -277,38 +271,6 @@ func VeleroInstall(ctx context.Context, veleroImage string, veleroNamespace stri
 	return nil
 }
 
-func VeleroUninstall(ctx context.Context, client *kubernetes.Clientset, extensionsClient *apiextensionsclient.Clientset,
```
dsu-igeek, line 280: Let's change VeleroUninstall to call uninstall.Uninstall for the moment and later we can have it run through the CLI.

dsu-igeek (changes requested):
Looks good, can you change VeleroUninstall to call uninstall.Uninstall.  Then we can change it to call the CLI later.  Thanks!

dsu-igeek (approved):

nrb (approved):

test/e2e/velero_utils.go
vadasambar: @dsu-igeek , not sure if this is the right place to ask this but I can take a look at using cli in Install and Uninstall in the test suite since I have some context on what needs to be done, might as well make that change as well. Do we have any ticket for this or should I create a new one?

PR #3409, opened 2021-02-05
Add credentials to VSLs

nrb:
Adds credential fields to volume snapshot locations so we can build up multiple credential support.


pkg/cmd/cli/snapshotlocation/set.go
```diff
@@ -0,0 +1,116 @@
...
+	return c
+}
+
+type SetOptions struct {
+	Name       string
+	Credential flag.Map
+}
+
+func NewSetOptions() *SetOptions {
+	return &SetOptions{}
```
zubron, line 61: We need to initialize the `Credential` field as is done in [3190](https://github.com/vmware-tanzu/velero/pull/3190/files#diff-da0544a77f284ebf456c090aeae8e1b5d0100cbabc631d2c96dc18365dfd8f70R62), otherwise there is an error when attempting to set the credential field: 
```
An error occurred: invalid argument "foo=bar" for "--credential" flag: error parsing "f"
```

zubron (changes requested):
Thanks for adding this @nrb! I tried it out locally and noticed that a fix is needed for the `set` command. It's good otherwise! :+1:

pkg/cmd/cli/snapshotlocation/set.go
nrb: Fixed!

zubron (approved):
Thanks!

nrb:
(This is a test)

/kind changelog-not-required

jenting (approved):

PR #3413, opened 2021-02-06
Changes to secrets design

dsu-igeek:

### DIFF
--- a/design/secrets.md
+++ b/design/secrets.md
@@ -183,9 +183,10 @@ Although the handling of secrets will be similar to how credentials are currentl
 ## Parallelism
 In order to support parallelism, Velero will need to be able to use multiple credentials simultaneously with the
 ObjectStore.  Currently backups are single threaded and a single BSL will be used throughout the entire backup.  The only
-existing points of parallelism are when a user downloads logs for a backup and when the .  In the current code, `download_request_controller.go`
-creates a new plug-in manager and hence another ObjectStore plugin in parallel with the ObjectStore plugin servicing a
-backup or restore (if one is running).
+existing points of parallelism are when a user downloads logs for a backup or the BackupStorageLocationReconciler 
+reconciles while a backup or restore is running.  In the current code, `download_request_controller.go` and 
+`backup_storage_location_controller.go` create a new plug-in manager and hence another ObjectStore plugin in 
+parallel with the ObjectStore plugin servicing a backup or restore (if one is running).
 
 ## Alternatives Considered
 
--- a/design/secrets.md
+++ b/design/secrets.md
@@ -8,15 +8,19 @@ This makes it so switching from one plugin to another necessitates overriding th
 
 - To allow Velero to create and store multiple secrets for provider credentials, even multiple credentials for the same provider
 - To improve the UX for configuring the velero deployment with multiple plugins/providers.
+- Enable use cases such as AWS volume snapshots w/Minio as the object storage
+- Continue to support use cases where multiple Backup Storage Locations are in use simultaneously 
+    - `velero backup logs` while backup/restore is running
+- Handle changes in configuration while operations are happening as well as they currently are
 
 ## Non Goals
 
 - To make any change except what's necessary to handle multiple credentials
-- To allow multiple credentials for or change the UX for node-based authentication (e.g. AWS IAM, GCP Workload Identity, Azure AAD Pod Identity).
+- To allow multiple credentials for or change the UX for node-based authentication (e.g. AWS IAM, GCP Workload Identity, Azure AAD Pod Identity).  Node-based authentication will not allow cases such as a mix of AWS snapshots with Minio object storage.
 
 ## Design overview
 
-Instead of one credential per Velero deployment, multiple credentials can be added and used with different BSLs VSLs.
+Instead of one credential per Velero deployment, multiple credentials can be added and used with different BSLs.
   
 There are two aspects to handling multiple credentials:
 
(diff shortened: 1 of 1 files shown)


design/secrets.md
```diff
@@ -84,6 +147,22 @@ Currently, GCP is the only provider that relies on the existing environment vari
...
+share the cloud credentials.  The VolumeSnapshotter cannot use the Minio credentials to take snapshots.  We need to split
+the VolumeSnapshotter credentials from the ObjectStore credentials.  We will do this by always using the default cloud
+credentials with the VolumeSnapshotter plugins and not passing the BSL credentials in the current implementation.
+
+## Velero Plugin for vSphere compatibility
+
+The vSphere plugin is implemented as a BackupItemAction and shares the credentials of the AWS plug-in for S3 access.
+The backup storage location is passed in _Backup.Spec.StorageLocation_.  Currently the plugin retrieves the S3 bucket and
+server from the BSL and creates a BackupRespositoryClaim with that and the credentials retrieved from the cloud credential.
+The plug-in will need to be modified to retrieve the credentials field from the BSL and use that credential in the
```
zubron, line 163: Having looked through the VSphere plugin code, it seems that [this is the section that will need to change](https://github.com/vmware-tanzu/velero-plugin-for-vsphere/blob/d2cba53d3b8092ac8562df025688daac06e9bd16/pkg/utils/utils.go#L172-L209) as it's currently using the fixed secret name and just selecting the first data field within that secret.

The current design is passing through a file path which has the credential written to disk which won't work. I haven't looked at the plugin code in depth but it seems like it may be enough to also include the secret name and key in the config as well. This would allow the VSphere plugin to use those values from the BSL config to retrieve the necessary secret to configure the AWS client. This effectively combines options 2 and 3.

design/secrets.md
```diff
@@ -84,6 +147,22 @@ Currently, GCP is the only provider that relies on the existing environment vari
 For GCP, the environment variable will be overwritten with the path of the serialized secret.
 
 
+## Split credentials between VolumeSnapshotter and ObjectStore plugins
+
+One of the use cases we wish to satisfy is the ability to specify a different object store than the cloud provider offers,
+for example, using a Minio S3 object store from within AWS.  Currently the VolumeSnapshotter and the ObjectStore plugin
+share the cloud credentials.  The VolumeSnapshotter cannot use the Minio credentials to take snapshots.  We need to split
+the VolumeSnapshotter credentials from the ObjectStore credentials.  We will do this by always using the default cloud
+credentials with the VolumeSnapshotter plugins and not passing the BSL credentials in the current implementation.
```
zubron, line 156: The VSLs will have a separate configurable credentials like the BSLs. Like the ObjectStore instantiation, the secret configured in the VSL will be written to disk and passed through the config when calling `Init`. This draft PR for AWS (https://github.com/vmware-tanzu/velero-plugin-for-aws/pull/69) hopefully shows that the two plugins can be configured independently and so the VolumeSnapshotter will not need to rely on the existing default credentials.

design/secrets.md
```diff
@@ -98,9 +177,26 @@ Multiple credential handling will only be used in the case where a particular BS
 
 Although the handling of secrets will be similar to how credentials are currently managed within Velero, care must be taken to ensure that any new code does not leak the contents of secrets, for example, including them within logs.
 
+## Parallelism
+In order to support parallelism, Velero will need to be able to use multiple credentials simultaneously with the
+ObjectStore.  Currently backups are single threaded and a single BSL will be used throughout the entire backup.  The only
+existing point of parallelism is when a user downloads logs for a backup.  In the current code, `download_request_controller.go`
+creats a new plug-in manager and hence another ObjectStore plugin in parallel with the ObjectStore plugin servicing a
+backup or restore (if one is running).
```
zubron, line 185: My understanding from looking at the instantiation of ObjectStores/Plugin Managers is that each controller is [given a function to create a new plugin manager](https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/server/server.go#L564-L566) which is called every time a controller needs to create an ObjectStore/VolumeSnapshotter. If I'm following the code correctly, this means that there is no process reuse between controllers using the same plugin but operating on different BSLs at a given moment, so current parallelism support should not be impacted. For a controller using the same plugin for different BSLs and VSLs (e.g. MinIO for Objects and AWS for snapshots), the fact that these are initialised independently with different config should mean that this use case will be supported.

design/secrets.md
```diff
@@ -92,15 +174,32 @@ Users will be able to set credentials during install and a secret will be create
...
+Multiple credential handling will only be used in the case where a particular BSL has been modified to use an existing secret.
 
 ## Security Considerations
 
 Although the handling of secrets will be similar to how credentials are currently managed within Velero, care must be taken to ensure that any new code does not leak the contents of secrets, for example, including them within logs.
 
+## Parallelism
+In order to support parallelism, Velero will need to be able to use multiple credentials simultaneously with the
+ObjectStore.  Currently backups are single threaded and a single BSL will be used throughout the entire backup.  The only
+existing points of parallelism are when a user downloads logs for a backup and when the .  In the current code, `download_request_controller.go`
```
zubron, line 186: This sentence ends abruptly, should it be like "when a user downloads logs for a backup while a backup or restore is in progress"?

dsu-igeek: Thanks for catching that.  I added the other case there and cleaned it up a bit.

zubron (approved):
Thanks for improving this doc!

ashish-amarnath (approved):

PR #3416, opened 2021-02-07
Design doc for upload progress monitoring

dsu-igeek:


ashish-amarnath (commented):
Here is an alternate proposal:
The goal here is to ensure the durability of Velero backups, and that backups, once completed, can be fully restored.
To accomplish this we propose: 
1. Introducing a new `BackupProgressTracker` gRpc service that would be implemented as a plugin. This service would have the familiar `AppliesTo` and `Execute` rpcs. The `AppliesTo` rpc would indicate what resources this plugin applies to, and the `Execute` (this can be called something else) would take as input ExecuteRequest with fields `item` and the `backup` (similar to `BackupItemAction's ExecuteRequest), and return, as output, ExecuteResponse with fields similar to the `UploadProgress` that's in this proposal.

2. Following changes in the Backup controller:
          i. [Here](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/backup_controller.go#L649), instead of setting the backup phase as `Completed` would set the phase as `Uploading`.
          ii. [Here](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/backup_controller.go#L659), start a progress tracker goroutine which would periodically, say every 1s, invoke all the `BackupProgressTracker` plugins' `Execute` rpc to obtain the status of the backup for each item and patch the backup object's status with this info. Once the `UploadProgress` from all the plugins have reached a state indicating durability of the backup, the goroutine returns which is when the backup's phase will be set to completed/uploaded.

3. Requires the following changes to the backup API: 
     - Introduce a `map[string]UploadProgress` as a field in the backup status which will be patched by the goroutine explained in 2.ii above. This will be mapping item to its latest `UploadProgress`

Benefits of this approach:
1. Introducing a new gRpc service is not a breaking change. (So is adding a new rpc to an existing service, but this is more intuitive in the `go-plugin` framework.)
2. It is more extendable across resource types as it enables tracking upload progress of all resource types, even those that don't have a corresponding `BackupItemAction` or a `VolumeSnapshotter` interface.
3. UploadProgress of all items in the backup can be tracked in a consistent way and in one single point.

We may have previously considered this approach but this seems more intuitive to me now than modifying the existing `BackupItemAction` and `VolumeSnapshotter` plugins. WDYT?

design/upload-progress.md
```diff
@@ -0,0 +1,293 @@
...
+InProgress backups will not have Backup resource present in the object store.  During reconciliation, backups which
+do not have a Backup resource in the object store will be ignored.
+
+### Removal of InProgress backups from object store
+
+## Plug-in API changes
+
+### UploadProgress struct
+
+    type UploadProgress struct {
```
ashish-amarnath, line 155: we should probably add fields to indicate when the upload was started and when was "this" progress last updated.

dsu-igeek:
> Here is an alternate proposal:
> The goal here is to ensure the durability of Velero backups, and that backups, once completed, can be fully restored.
> To accomplish this we propose:
> 
> 1. Introducing a new `BackupProgressTracker` gRpc service that would be implemented as a plugin. This service would have the familiar `AppliesTo` and `Execute` rpcs. The `AppliesTo` rpc would indicate what resources this plugin applies to, and the `Execute` (this can be called something else) would take as input ExecuteRequest with fields `item` and the `backup` (similar to `BackupItemAction's ExecuteRequest), and return, as output, ExecuteResponse with fields similar to the `UploadProgress` that's in this proposal.
> 2. Following changes in the Backup controller:
>    i. [Here](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/backup_controller.go#L649), instead of setting the backup phase as `Completed` would set the phase as `Uploading`.
>    ii. [Here](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/backup_controller.go#L659), start a progress tracker goroutine which would periodically, say every 1s, invoke all the `BackupProgressTracker` plugins' `Execute` rpc to obtain the status of the backup for each item and patch the backup object's status with this info. Once the `UploadProgress` from all the plugins have reached a state indicating durability of the backup, the goroutine returns which is when the backup's phase will be set to completed/uploaded.
> 3. Requires the following changes to the backup API:
>    
>    * Introduce a `map[string]UploadProgress` as a field in the backup status which will be patched by the goroutine explained in 2.ii above. This will be mapping item to its latest `UploadProgress`
> 
> Benefits of this approach:
> 
> 1. Introducing a new gRpc service is not a breaking change. (So is adding a new rpc to an existing service, but this is more intuitive in the `go-plugin` framework.)
> 2. It is more extendable across resource types as it enables tracking upload progress of all resource types, even those that don't have a corresponding `BackupItemAction` or a `VolumeSnapshotter` interface.
> 3. UploadProgress of all items in the backup can be tracked in a consistent way and in one single point.
> 
> We may have previously considered this approach but this seems more intuitive to me now than modifying the existing `BackupItemAction` and `VolumeSnapshotter` plugins. WDYT?

Upload progress may take a long time.  We want to be able to recover from restarts of the Velero server while uploads are in progress.  Also, if the backup is deleted before the upload finishes we want to terminate any polling on upload status and, if possible, abort the upload.

The design introduces a BackupItemProgress plug-in for BackupItemAction monitoring.  In order to monitor the upload progress, we will need to know what to monitor, which is why the BackupItemAction needs to be modified to return a snapshot ID.  It might be worthwhile to add a new BackupItemActionWithSnapshot plug-in rather than add the snapshot ID return to the existing plug-ins.

The VolumeSnapshotter plugins work at the volume level, so invoking a BackupItemProgress action for them requires that they map the PV to a volume.  Adding the UploadProgress API should not be a breaking change once API versioning has been introduced.

design/upload-progress.md
dsu-igeek: Sure, added that.

design/upload-progress.md
```diff
@@ -0,0 +1,318 @@
...
+        started, updated time.Time              // When the upload was started and when the last update was seen.  Not all
+                                                // systems retain when the upload was begun, return Time 0 (time.Unix(0, 0))
+                                                // if unknown.
+    }
+
+### VolumeSnapshotter changes
+
+A new method will be added to the VolumeSnapshotter interface (details depending on plug-in versioning spec)
+
+    UploadProgress(snapshotID string) (UploadProgress, error)
```
ashish-amarnath, line 173: The name `UploadProgress` indicates that this API is uploading progress, please consider renaming it to `GetSnapshotStatus` or something similar.
Having the `snapshotID` parameter as a string seems too restrictive for it to work across different plugins.
Suggest making this into an `GetSnapshotStatusInput` (assuming the API is going to be called `GetSnapshotStatus`. Please update the input struct name if a different name is chosen for the API

design/upload-progress.md
```diff
@@ -0,0 +1,318 @@
...
+`error` is set if there is an issue retrieving progress.  If the snapshot is has encountered an error during the upload,
+the error should be return in UploadProgress and error should be nil.
+
+### BackupItemProgress plug-in
+
+Currently CSI snapshots and the Velero Plug-in for vSphere are implemented as BackupItemAction plugins.  The majority of
+BackupItemAction plugins do not take snapshots or upload data so rather than modify BackupItemAction we introduce two new
+plug-ins, SnapshotItemAction and BackupItemProgress.  SnapshotItemAction will be used in place of BackupItemAction for
+the CSI snapshots and the Velero Plug-in for vSphere and will return a snapshot ID in addition to the item itself.
+BackupItemProgress will work similarly to DeleteItemAction, being called only for item kinds which it
```
ashish-amarnath, line 188: Is the `BackupItemProgress` tracking the progress of a `BackupItemAction` or `SnapshotItemAction`?
It seems to me that it is tracking the progress of a `SnapshotItemAction`. If that's the case, please consider making this as an API in the `SnapshotItemAction` plugin interface and naming it accordingly. If this assumption is not true, then perhaps we should have a similar API in the `SnapshotItemAction` plugin interface as well. WDYT?

ashish-amarnath: If the `BackupItemProgress` is in fact monitoring the progress of a `BackupItemAction`, is it possible to add it as an API in the `BackupItemAction` plugin interface?

design/upload-progress.md
```diff
@@ -0,0 +1,318 @@
...
+
+## vSphere plug-in
+
+The vSphere Plug-in for Velero uploads snapshots to S3 in the background.  This is also a BackupItemAction plug-in,
+it will check the status of the Upload records for the snapshot and return progress.
+
+## Backup workflow changes
+
+The backup workflow remains the same until we get to the point where the `velero-backup.json` object is written.
+At this point, we will queue the backup to a finalization go-routine.  The next backup may then begin.  The finalization
```
ashish-amarnath, line 269: I am wondering if there is more value in introducing another CRD, say `BackupMonitorRequest`, and have a controller reconcile/update the state of the backup. This makes it more robust to velero server restarts.
WDYT?

ashish-amarnath (commented):
The approach LGTM.
Added a few comments that can be addressed/figured out during implementation.

design/upload-progress.md
```diff
@@ -0,0 +1,318 @@
...
+A new method will be added to the VolumeSnapshotter interface (details depending on plug-in versioning spec)
+
+    UploadProgress(snapshotID string) (UploadProgress, error)
+
+UploadProgress will report the current status of a snapshot upload.  This should be callable at any time after the snapshot
+has been taken.  In the event a plug-in is restarted, if the snapshotID continues to be valid it should be possible to
+retrieve the progress.
+
+`error` is set if there is an issue retrieving progress.  If the snapshot is has encountered an error during the upload,
+the error should be return in UploadProgress and error should be nil.
```
carlisia, line 180: I am trying to understand this part. The `error` returned by this function is being reserved for the case where retrieving progress has a problem. If there is a problem with the actual upload, the `error` in the struct should contain the value for that error. Should the process halt at this point and the function return? If no, could there be more upload errors that will need to be appended?

design/upload-progress.md
```diff
@@ -0,0 +1,318 @@
...
+
+### VolumeSnapshotter changes
+
+A new method will be added to the VolumeSnapshotter interface (details depending on plug-in versioning spec)
+
+    UploadProgress(snapshotID string) (UploadProgress, error)
+
+UploadProgress will report the current status of a snapshot upload.  This should be callable at any time after the snapshot
+has been taken.  In the event a plug-in is restarted, if the snapshotID continues to be valid it should be possible to
+retrieve the progress.
```
carlisia, line 177: Curious: if there is a plugin restart, and the snapshotID is still valid, will the upload progress start from the beginning?

design/upload-progress.md
```diff
@@ -0,0 +1,318 @@
...
+
+
+### Uploading (new)
+The "Uploading" phase signifies that the main part of the backup, including snapshotting has completed successfully
+and and uploading is continuing.  In the event of an error during uploading, the phase will change to 
+UploadingPartialFailure.  On success, the phase changes to Completed
+
+### UploadingPartialFailure (new)
+The "UploadingPartialFailure" phase signifies that the main part of the backup, including snapshotting has completed,
+but there were partial failures either during the main part or during the uploading.
```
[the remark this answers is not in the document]
carlisia: One more thought related to the `UploadingPartialFailure` (or `UploadingFailure`): If it's possible for the process to hang in this  phase, would a delete operation clear it? Could we/should we think about ensuring it does?

design/upload-progress.md
```diff
@@ -0,0 +1,318 @@
...
+be treated as if they have been requeued and progress checked and the backup will be requeued or moved to a terminal
+phase as appropriate.
+
+# Implementation tasks
+
+VolumeSnapshotter new plugin APIs  
+BackupItemProgress new plugin interfaces  
+New backup phases  
+AWS EBS plug-in UploadProgress implementation  
+Upload monitoring
```
carlisia, line 294: Is this task (or is one of the others) that would handle 

> _PartiallyFailed_, or _Failed_). Reconciliation should ignore backups that do not have a `velero-backup.json` object.

Source:
https://github.com/vmware-tanzu/velero/pull/3416/files#diff-7c9f7359e7cbc8bd39e74b3998bc44a06307b55028e60c0234d6032266afae79R244

carlisia (commented):
Couple suggestions, couple comments, couple questions.

Overall lgtm, looks like someone could take it and implemented. Can't wait to have this split in the backup process!

design/upload-progress.md
dsu-igeek: We can change the name.  For VolumeSnapshotter, snapshotID is a string currently

design/upload-progress.md
dsu-igeek: You're right, it should be SnapshotItemAction, I forgot to change it.
The BackupItemAction/RestoreItemAction/DeleteItemAction split up is the norm right now, so adding another plug-in fits with that model.  Maybe they should both be combined in SnapshotItemAction.

design/upload-progress.md
```diff
@@ -0,0 +1,318 @@
...
+has no fatal errors.
+
+If any snapshots are not complete, the backup will move to either Uploading or UploadingPartialFailure or Failed.
+
+Post-snapshot operations may take a long time and Velero and its plugins may be restarted during 
+this time.  Once a backup has moved into the Uploading or UploadingPartialFailure phase, another 
+backup may be started.
+
+While in the Uploading or UploadingPartialFailure phase, the snapshots and backup items will be periodically polled.
+When all of the snapshots and backup items have reported success, the backup will move to the Completed or
```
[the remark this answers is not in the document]
dsu-igeek: success is the correct meaning.  Reporting successfully would just mean they reported something but not necessarily success.

design/upload-progress.md
```diff
@@ -0,0 +1,318 @@
...
+If the backup has fatal errors it will move to the "Failed" state and finish. If a backup fails, the upload will not be
+cancelled but it will not be monitored either.  For backups in any phase, all snapshots will be deleted when the backup
+is deleted.  Plugins will cancel any data movement and
+remove snapshots and other associated resources when the VolumeSnapshotter DeleteSnapshot method or 
+DeleteItemAction Execute method is called.
+
+Velero will poll the plugins for status on the snapshots when the backup exits the "InProgress" phase and
+has no fatal errors.
+
+If any snapshots are not complete, the backup will move to either Uploading or UploadingPartialFailure or Failed.
```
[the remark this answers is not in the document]
dsu-igeek: On individual snapshots I'm going to say it's all or nothing.  For the backup as a whole, if the snapshots failed but we got all of the Kubernetes resources, that would be a "PartialFailure" at the end.

design/upload-progress.md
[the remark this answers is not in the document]
dsu-igeek: When the backup is in any of the Uploading phases it is not usable.  In case of a restart, though, we need to know that there was a partial failure somewhere in there, so it's kept in the phase.

The plan is for a delete operation to abort the backup and any uploads in progress.

> If the backup is deleted (cancelled), the plug-ins will
> attempt to delete the snapshots and stop the data movement - this may not be possible with all
> storage systems.

design/upload-progress.md
```diff
@@ -0,0 +1,318 @@
...
+
+Post-snapshot operations may take a long time and Velero and its plugins may be restarted during 
+this time.  Once a backup has moved into the Uploading or UploadingPartialFailure phase, another 
+backup may be started.
+
+While in the Uploading or UploadingPartialFailure phase, the snapshots and backup items will be periodically polled.
+When all of the snapshots and backup items have reported success, the backup will move to the Completed or 
+PartiallyCompleted phase, depending on whether the backup was in the Uploading or UploadingPartialFailure phase.
```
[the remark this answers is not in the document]
dsu-igeek: You're right, thanks for catching that.

design/upload-progress.md
dsu-igeek: UploadProgress returns where things are at, it doesn't wait for the upload to finish.  If there is an actual error in the upload, UploadProgress.completed should be set to true and UploadProgress.err would be set to the error that caused the upload to stop.

design/upload-progress.md
dsu-igeek: I added a task, thanks

design/upload-progress.md
dsu-igeek: That will be plugin dependent.  For example, in EBS, the plugin is only monitoring the upload, it's EBS that actually does the work.  For the vSphere plug-in, it is handling the upload and it has restart logic.  There might be a plugin that simply fails the upload if it gets restarted, I don't think we have any examples yet.

design/upload-progress.md
dsu-igeek: OK, I merged them together into SnapshotItemAction

design/upload-progress.md
dsu-igeek: I think we can get the restart behavior without adding an additional CRD.  Keep it in mind, and when we are implementing if we need it, let's add it then.

design/upload-progress.md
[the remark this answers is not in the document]
dsu-igeek: I updated the file.

carlisia (dismissed):
This LGTM! 👍

ashish-amarnath (approved):

ashish-amarnath:
@dsu-igeek Looks like the latest commit is missing a sign-off.

carlisia (dismissed):
👍

ashish-amarnath (dismissed):

carlisia (approved):

Issue #2557, opened 2020-05-21
Re-enable skipping binary build on PRs if only site changed


skriss:
When we were using Travis for CI on PRs, we had some code (https://github.com/vmware-tanzu/velero/blob/master/.travis.yml#L12 and https://github.com/vmware-tanzu/velero/blob/master/hack/ci-check.sh) that would skip building/testing the Velero code if only files inside `site/` had changed. This made site/docs-only PRs pass checks much faster.

When we migrated to GitHub Actions for PR builds, we lost this. It'd be nice to add it back. It looks like the `ci-check.sh` script will need some minor changes since it's got logic specifically tied to Travis.

This is a nice-to-have for 1.5.

PR #3438, opened 2021-02-09
Don't run CI checks if only the site changed

nrb:
Also include the github directory to enable faster iteration on GitHub
actions

### DIFF
--- a/.github/workflows/pr-changelog-check.yml
+++ b/.github/workflows/pr-changelog-check.yml
@@ -14,9 +14,6 @@ jobs:
       - name: Check out the code
         uses: actions/checkout@v2
 
-    - name: Check out the code
-      uses: actions/checkout@v2
-
     - name: Changelog check
       if: ${{ !(contains(github.event.pull_request.labels.*.name, 'kind/changelog-not-required') || contains(github.event.pull_request.labels.*.name, 'Design') || contains(github.event.pull_request.labels.*.name, 'Website') || contains(github.event.pull_request.labels.*.name, 'Documentation'))}}
       run: ./hack/changelog-check.sh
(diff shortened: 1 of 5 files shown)


carlisia (dismissed):
Awesome 🎉

jenting:
Related to https://github.com/vmware-tanzu/velero/issues/2557

jenting:
This needs to address the conflicts.

nrb:
Conflicts fixed!

.github/workflows/pr-changelog-check.yml
```diff
@@ -1,11 +1,18 @@
...
+      - "site/**"
+      - ".github/**"
+      - "design/**"
 jobs:
-
   build:
     name: Run Changelog Check
     runs-on: ubuntu-latest
     steps:
+      - name: Check out the code
```
zubron, line 14: This step is being included twice now.

zubron (changes requested):
This is great, thanks @nrb! There's a single fix needed in the changelog check action where code checkout is being called twice.

jenting (approved):

zubron (approved):

zubron:
Hmm, I don't know if this will work in it's current state. We have branch protection for main that requires that the `Run CI` check must pass before merging, but this will now never run on PRs like this. I don't know if there is a way to make the checks conditional. Looks like it's an open request in the GitHub community: https://github.community/t/feature-request-conditional-required-checks/16761

nrb:
> We have branch protection for main that requires that the Run CI check must pass before merging, but this will now never run on PRs like this.

Oof, you're right. So we'd need to make the change in the Run CI workflow to check for this. I _think_ we can do that.

nrb:
I'm going to close this and look into an alternate approach. I have another idea for cleaning up some of the CRD testing anyway, so that they're consolidated into 1 file.

PR #3441, opened 2021-02-10
Use correct suffix for Labeler config file

zubron:
The labeler action was failing as it was looking for
`.github/labels.yaml` but the file has the suffix `.yml`. This change
fixes the path used by the labeler action.


jenting (approved):

nrb (approved):
🤦  Sorry about that!

PR #3446, opened 2021-02-10
Restore CAPI cluster objects in a better order

nrb:
Restoring CAPI workload clusters without this ordering caused the
capi-controller-manager code to panic, resulting in an unhealthy cluster
state.

This can be worked around
(https://community.pivotal.io/s/article/5000e00001pJyN41611954332537?language=en_US),
but we provide the inclusion of these resources as a default in order to
provide a better out-of-the-box experience.

A workaround for https://github.com/kubernetes-sigs/cluster-api/issues/4105


zubron (approved):
Thanks for including the links to the workaround post :+1:

jenting (approved):

PR #3448, opened 2021-02-10
Combine CRD install verification into 1 job, and update k8s versions

nrb:
Previously, we had independent files for each Kubernetes version that repeated the exact same command. This was tough to maintain as each change had to go into each file.

This PR introduces the GitHub Actions [matrix strategy](https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix) which allows us to maintain a list of versions that the Actions runner will split into separate, parallel jobs for us.

Also, add the v1.15, v1.1.9 and v1.20 Kubernetes series into the testing matrix since it's much lighter weight to do so.


nrb:
I think this is ready for review.

I added a couple steps to cache the go modules across CI jobs so they're not constantly re-downloaded. I also change the workflow to first build the Linux binary, then download that pre-built binary for each of the Kubernetes versions we test it against.

You can see on the current checks that this sped up the CRD version checks themselves to all run in about 2 minutes. https://github.com/vmware-tanzu/velero/actions/runs/556465104 is the run itself.

jenting (approved):

carlisia (approved):
Fantastic, thank you!! 🎈

PR #3453, opened 2021-02-11
Optimize build image container

stmcginnis:

This changes the build-image Dockerfile to optimize image size and
operations that are performed for each layer. It brings the current
image size down from ~1.52GB to ~1.28GB.

Optimizations include:
* Switch to bitnami/golang:1.15 base image
* Reorder to steps to move the less likely to change layers lower
* Combine RUN lines that are part of the same task into one to reduce
  layers

No tracking issue filed, but happy to add one if folks feel it is needed.

I noticed the build-image container created when doing something as
simple as `make lint` was quite large and took awhile. This was my
attempt to optimize things.

/kind changelog-not-required

[X] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
[X] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.
[-] Updated the corresponding documentation in `site/content/docs/main`.

### DIFF
--- a/hack/build-image/Dockerfile
+++ b/hack/build-image/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2018, 2019, 2020 the Velero contributors.
+# Copyright 2018, 2019, 2020, 2021 the Velero contributors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.15
+FROM bitnami/golang:1.15
 
 ARG GOPROXY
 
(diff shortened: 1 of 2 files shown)


hack/build-image/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018, 2019, 2020 the Velero contributors.
+# Copyright 2018, 2019, 2020, 2021 the Velero contributors.
```
jenting, line 1:
```suggestion
# Copyright the Velero contributors.
```

hack/build-image/Dockerfile
```diff
@@ -12,19 +12,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.15
+FROM bitnami/golang:1.15
```
jenting, line 15: On my Linux amd64
- golang:1.15 839MB
- bitnami/golang:1.15 717MB

nrb:
@stmcginnis Thanks for this Sean - I'm wondering, would it be more effective to streamline the build process so we're not using the image in as many tasks? It looks like the image size isn't substantially reduced, and I would think it's a ~1 time cost - am I wrong in that understanding?

I definitely appreciate the goal and want to improve the build image, but I'm uncertain how this fits.

stmcginnis:
Yeah, we may get better gains looking at streamlining the overall process. The savings here isn't as big as I had originally hoped.

That said, there is still some value. Since a new image is created for every commit run against, over time the number of images will grow if the dev doesn't realize they should be cleaning things up. Plus with the ordering and combining of some layers, it should make subsequent images able to reuse layers more efficiently.

But totally fine with me if the team decides not to go ahead with these. Just let me know and I can close this out if we choose to go another route.

dsu-igeek:
I'm going to close this out for now.

PR #3475, opened 2021-02-18
Use pod namespace from backup when matching PVBs

### LINKED ISSUE #3467: Restic restores failing when using namespace mapping
Restic restores are failing when using a namespace mapping as the check for whether a PVB matches the pod being restored checks the namespace of the PVB against the namespace of the pod: https://github.com/vmware-tanzu/velero/blob/main/pkg/restic/common.go#L98-L100

This check will fail in the case where the pod is being mapped to a new namespace and so a PVR will not be generated for that pod and the restore will not take place.

Instead of checking the new namespace of the pod, we should be checking the original namespace which will match the namespace for the associated PVB.

----

I am getting the same error when trying to restore into a different namespace. The kubernetes resources are still created. It is really strange, since restoring into the same namespace works perfectly fine.

I use a command like 

`velero restore create --from-backup test-backup3 --namespace-mappings mlflow-feature-hotfix:something-else`

Note that mlflow-feature-hotfix ist just a sample application with a postgres that I use to have some state while verifying the backups.

I have velero `1.5.3` installed via helm with the following values.

```yaml
initContainers:
  - name: velero-plugin-for-aws
    image: velero/velero-plugin-for-aws:v1.1.0
    imagePullPolicy: IfNotPresent
    volumeMounts:
      - mountPath: /target
        name: plugins
configuration:
  provider: aws
  backupStorageLocation:
    bucket: ${bucket}
    prefix: velero
  volumeSnapshotLocation:
    name: aws-default
  defaultVolumesToRestic: true
deployRestic: true
credentials:
  name: velero-aws
  secretContents:
    cloud: |
      [default]
      aws_access_key_id=${aws_access_key_id}
      aws_secret_access_key=${aws_secret_access_key}

```

I am runnning on Rancher on premise. I put the logs files here https://gist.github.com/JarnoRFB/25b6343e45ca63945ebc9eca457774ab , but to me the logs of the successful backup to the same namespace looked almost the same as the logs of the failing backup to the different namespace.

Any help much appreciated!

_Originally posted by @JarnoRFB in https://github.com/vmware-tanzu/velero/discussions/3385#discussioncomment-357126_

zubron:
In #3051, we introduced an additional check to ensure that a PVB matched
a particular pod by checking both the name and the namespace of the pod.
This caused an issue when using a namespace mapping on restore. In the
case where a namespace mapping is being used, the check for whether a
PVB matches a particular pod will fail as the PVB was created for the
original pod namespace and is not aware of the new namespace mapping
being used. This resulted in PVRs not being created for pods that were
being restored into new namespaces. The restic init containers were
being created to wait on the volume restore, however this would cause
the restored pods to block indefinitely as they would be waiting for a
volume restore that was not scheduled.

To fix this, we use the original namespace of the pod from the backup to
match the PVB to the pod being restored, not the new namespace where
the pod is being restored into.

Fixes #3467 

### DIFF
--- a/pkg/restore/restic_restore_action.go
+++ b/pkg/restore/restic_restore_action.go
@@ -76,8 +76,12 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
 		return nil, errors.Wrap(err, "unable to convert pod from runtime.Unstructured")
 	}
 
-	var sourcePod corev1.Pod
-	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.ItemFromBackup.UnstructuredContent(), &sourcePod); err != nil {
+	// At the point when this function is called, the namespace mapping for the restore
+	// has not yet been applied to `input.Item` so we can't perform a reverse-lookup in
+	// the namespace mapping in the restore spec. Instead, use the pod from the backup
+	// so that if the mapping is applied earlier, we still use the correct namespace.
+	var podFromBackup corev1.Pod
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.ItemFromBackup.UnstructuredContent(), &podFromBackup); err != nil {
 		return nil, errors.Wrap(err, "unable to convert source pod from runtime.Unstructured")
 	}
 
@@ -93,7 +97,7 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
 	for i := range podVolumeBackupList.Items {
 		podVolumeBackups = append(podVolumeBackups, &podVolumeBackupList.Items[i])
 	}
-	volumeSnapshots := restic.GetVolumeBackupsForPod(podVolumeBackups, &pod, sourcePod.Namespace)
+	volumeSnapshots := restic.GetVolumeBackupsForPod(podVolumeBackups, &pod, podFromBackup.Namespace)
 	if len(volumeSnapshots) == 0 {
 		log.Debug("No restic backups found for pod")
 		return velero.NewRestoreItemActionExecuteOutput(input.Item), nil
--- a/pkg/restore/restic_restore_action.go
+++ b/pkg/restore/restic_restore_action.go
@@ -76,6 +76,11 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
 		return nil, errors.Wrap(err, "unable to convert pod from runtime.Unstructured")
 	}
 
+	var sourcePod corev1.Pod
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.ItemFromBackup.UnstructuredContent(), &sourcePod); err != nil {
+		return nil, errors.Wrap(err, "unable to convert source pod from runtime.Unstructured")
+	}
+
 	log := a.logger.WithField("pod", kube.NamespaceAndName(&pod))
 
 	opts := label.NewListOptionsForBackup(input.Restore.Spec.BackupName)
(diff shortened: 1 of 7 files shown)


pkg/restore/restic_restore_action.go
```diff
@@ -76,6 +76,11 @@ func (a *ResticRestoreAction) Execute(input *velero.RestoreItemActionExecuteInpu
 		return nil, errors.Wrap(err, "unable to convert pod from runtime.Unstructured")
 	}
 
+	var sourcePod corev1.Pod
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.ItemFromBackup.UnstructuredContent(), &sourcePod); err != nil {
+		return nil, errors.Wrap(err, "unable to convert source pod from runtime.Unstructured")
+	}
+
```
ashish-amarnath, line 83: Just for reference, we could have also reverse looked up the `RestoreItemActionExecuteInput.Restore.Spec.NamespaceMapping`. But this is also fine.

ashish-amarnath (approved):
LGTM!

pkg/restore/restic_restore_action.go
zubron: I thought about doing that but there was actually an issue with that approach. At the point when this action's `Execute` function is [called in `restoreItem`](https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L1029-L1033), the namespace mapping hasn't yet been applied to the object being restored. That doesn't happen until [later in the function](https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L1120-L1125). This means that, under the current implementation, both `pod` and `sourcePod` have the same namespace when this is called. Rather than assuming they would be the same, I opted to look at the item from the backup. It seemed too risky for this change to change when the namespace mapping would be applied in `restoreItem`, but checking the pod from the backup also prevents this issue occurring again if when we do change when the mapping is applied.

zubron: I should add a comment though to explain why we can't rely on the namespace mapping from the restore spec. Thanks for reminding me about why I couldn't use that approach! :+1:

nrb:
Before reviewing, I think this has the potential for backporting so that v1.5.x users can benefit from the fix.

nrb (approved):
Thank you for the added test and the comment about why we can't do reverse lookup!

pkg/restore/restic_restore_action.go
ashish-amarnath: @zubron Not sure I follow. The `NamespaceMapping` that I was referring to is a field on the restore object spec. This field is set using the value of the [`namespace-mappings`](https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/cli/restore/create.go#L112) flag that is passed at the CLI running `velero restore create` command.

ashish-amarnath (approved):

pkg/restore/restic_restore_action.go
ashish-amarnath: Thanks for clarifying @zubron.
Just capturing our discussion here. Changes at [line79-82](https://github.com/vmware-tanzu/velero/pull/3475/files#diff-9f110ab28d31cd60e15e9ec43ad813f22bd8b9329eaf0ebe70c0b6590592b76dR79-R87) weren't necessary and we could have used `pod.Namespace` at https://github.com/vmware-tanzu/velero/pull/3475/files#diff-9f110ab28d31cd60e15e9ec43ad813f22bd8b9329eaf0ebe70c0b6590592b76dR100 
and added unit tests to catch any future code reordering that may break. The changes at [line79-82](https://github.com/vmware-tanzu/velero/pull/3475/files#diff-9f110ab28d31cd60e15e9ec43ad813f22bd8b9329eaf0ebe70c0b6590592b76dR79-R87) is future proofing.

PR #3476, opened 2021-02-19
Add namespace parameter in restic's code

sfelix-orange:
This intends to be able to target specific namespaces when
having non-admin privileges.

When running velero as a non-admin user, listing pods from
other namespace is forbidden. The use of an environment variable in
the YAML deployment file allows to target a specific namespace in which
we want to list pods.

[ X ] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
[ X ] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.
[ X ] Updated the corresponding documentation in `site/content/docs/main`.

### DIFF
--- a/site/content/docs/main/namespace.md
+++ b/site/content/docs/main/namespace.md
@@ -19,4 +19,8 @@ velero client config set namespace=<NAMESPACE_VALUE>
 
 Alternatively, you may use the global `--namespace` flag with any operational command to tell Velero where to run.
 
+## Customize the targeted namespace
+
+In the YAML deployment file using Velero's image, define the environment variable NAMESPACE_NAME to specify the namespace in which you want Velero to backup resources.
+
 [0]: basic-install.md#install-the-cli
--- a/changelogs/unreleased/3476-sfelix-orange
+++ b/changelogs/unreleased/3476-sfelix-orange
@@ -1,0 +1,1 @@
+Add namespace parameter in Restic's code
--- a/pkg/cmd/cli/restic/server.go
+++ b/pkg/cmd/cli/restic/server.go
@@ -126,7 +126,7 @@ func newResticServer(logger logrus.FieldLogger, factory client.Factory, metricAd
 	// filter to only pods scheduled on this node.
 	podInformer := corev1informers.NewFilteredPodInformer(
 		kubeClient,
-		metav1.NamespaceAll,
+		os.Getenv("NAMESPACE_NAME"),
 		0,
 		cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
 		func(opts *metav1.ListOptions) {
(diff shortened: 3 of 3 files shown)


jenting (changes requested):
Could you please change the changelog to be more simpler? Because the changelog will appear in the release note. Thank u.

sfelix-orange:
> 
> 
> Could you please change the changelog to be more simpler? Because the changelog will appear in the release note. Thank u.

Hello,

thank you for your feedback! Changelog has been simplified.

Best regards.

jenting (changes requested):
Does Velero support the docker-compose file?
Also, this would break the change that Velero deploys on Kubernetes cluster since the environment variable `NAMESPACE_NAME` does not exist now.

sfelix-orange:
> 
> 
> Does Velero support the docker-compose file?
> Also, this would break the change that Velero deploys on Kubernetes cluster since the environment variable `NAMESPACE_NAME` does not exist now.

sfelix-orange:
> 
> 
> Does Velero support the docker-compose file?
> Also, this would break the change that Velero deploys on Kubernetes cluster since the environment variable `NAMESPACE_NAME` does not exist now.

My bad, we tested it on Minikube, so I was indeed talking about YAML deployment file.
For your second point, is NAMESPACE_NAME is not defined, it's value will be an empty string which is consistent with the current behaviour :
// NamespaceAll is the default argument to specify on a context when you want to list or filter resources across all namespaces
	NamespaceAll string = ""
[](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1)

Thank you for your feedback!

site/content/docs/main/namespace.md
```diff
@@ -19,4 +19,8 @@ velero client config set namespace=<NAMESPACE_VALUE>
 
 Alternatively, you may use the global `--namespace` flag with any operational command to tell Velero where to run.
 
+## Customize the targeted namespace
+
+In the YAML deployment file using Velero's image, define the environment variable NAMESPACE_NAME to specify the namespace in which you want Velero to backup resources.
```
nrb, line 24: I'm not sure this documentation is accurate. From what I'm reading, the environment variable limits what pods restic can see. The way I read this note implies this is the only way to target a namespace is via this environment variable.

The user still has to select a relevant namespace with the `--include-namespaces` and `--exclude-namespaces` arguments, unless I'm misunderstanding your deployment setup.

nrb:
> The use of an environment variable in the YAML deployment file allows to target a specific namespace in which we want to list pods.

Can you please outline how you're currently deploying Velero, and how you plan to target namespaces? Are you deploying Velero into each namespace on the cluster?

site/content/docs/main/namespace.md
ashish-amarnath: +1 This is restricting the pods that restic can see and that is probably not what you want.
@sfelix-orange can you please share what is the problem you are trying to solve? We can perhaps propose a solution to that.

ashish-amarnath (changes requested):
Waiting on clarification on that issue that this PR intends to solve.

nrb:
@sfelix-orange I'm going to close out this PR since we've not heard back, and I don't really know how you're deploying.

Reading this code change, it appears the user would have to deploy a restic DaemonSet for _each namespace_ that they want to protect due to how the environment variable is declared, which I don't think will work with our code as is - Velero expects 1 and only 1 DaemonSet.

If you could please open an issue describing what problem you're trying to solve and link it here, that would help us land on a more general solution.

PR #3477, opened 2021-02-19
Add option to bypass calls to pvLister

sfelix-orange:
This intends to bypass calls to pvLister in function
GetVolumeDirectory when we know all volumes are CSI.

If running velero as a non-admin user on a cluster, the call
to pvLister can cause errors for namespaces which are not
assigned to the user. If all volumes are CSI, there is no
need to use pvLister The use of an environment variable in 
the docker-compose.yml allows to bypass the call to pvLister.

### DIFF
--- a/pkg/util/kube/utils.go
+++ b/pkg/util/kube/utils.go
@@ -19,6 +19,7 @@ package kube
 import (
 	"context"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/pkg/errors"
@@ -124,6 +125,10 @@ func GetVolumeDirectory(pod *corev1api.Pod, volumeName string, pvcLister corev1l
 	pvc, err := pvcLister.PersistentVolumeClaims(pod.Namespace).Get(volume.VolumeSource.PersistentVolumeClaim.ClaimName)
 	if err != nil {
 		return "", errors.WithStack(err)
+	}
+
+	if os.Getenv("ONLY_CSI") == "True" {
+		return pvc.Spec.VolumeName + "/mount", nil
 	}
 
 	pv, err := pvLister.Get(pvc.Spec.VolumeName)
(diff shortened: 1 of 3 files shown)


pkg/util/kube/utils.go
```diff
@@ -126,6 +126,10 @@ func GetVolumeDirectory(pod *corev1api.Pod, volumeName string, pvcLister corev1l
 		return "", errors.WithStack(err)
 	}
 
+   	if os.Getenv("ONLY_CSI") != "" {
```
nrb, line 129: Where is this defined? Can we key off some other information to get at this rather than setting an environment variable?

nrb (changes requested):
Thanks for this change! Can you also provide a changelog file and documentation on how this is expected to be used?

In particular, the `ONLY_CSI` environment variable looks new, as I don't see it anywhere else in the Velero codebase. Is the expectation to set this in the Deployment? Something else?

pkg/util/kube/utils.go
sfelix-orange: Hello,
this variable should be defined in the docker-compose.yml using Velero's image. If not define, then de default behavior will be conserved.

nrb:
> The use of an environment variable in the docker-compose.yml allows to bypass the call to pvLister.

Are you trying to use Velero in a Docker Swarm installation? We don't support a docker-compose.yml file.

zubron (commented):
Hi @sfelix-orange. Thanks for your efforts here and for your patience while we catch up on PRs and issues. I'm currently looking through your issue (#3633), and although this PR might go someway to solving that, I don't think this the best long term approach for the project.

Ideally, users should not have to instruct Velero that it will only be interacting with a particular type of volume. Also, if this environment variable was enabled in a deployment where non-CSI volumes are being used, this approach is inflexible as it would return the wrong value for those volumes.

I'd like for us to find a more general approach to supporting Velero in a multi-tenant environment.

dsu-igeek:
I'm closing this pull request.  You can open an issue on backing up in clusters where you don't have admin permissions, but there are too many features that won't work at the moment for us to support this as a general thing.  Thanks!

Issue #3304, opened 2021-01-19
Allow Velero to support multiple credentials per storage provider


nrb:
**Describe the problem/challenge you have**
Currently, Velero can only have 1 credential for a given storage platform, which is limiting if backups should be sent to different locations. For example, if someone is using GCP and wants to send backups to 2 different object storage buckets, this is not currently supported.

**Describe the solution you'd like**
Velero should offer some way to allow all storage provider platforms the ability to utilize multiple credential files. This should work for both volume snapshots and object storage.

**Anything else you would like to add:**
AWS can get around this by specifying profiles in the credential file, but not all file formats offer this support.

Also, restic support should be taken into account as it is currently part of Velero's core, not a plugin.

CSI snapshot support may be tricky here, as the credentials for snapshotting there will reside with the CSI driver itself.

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

zubron:
Fixed in the #3442, #3489, and #3559.

zubron:
Note: updated issue description to remove reference to volume snapshots as this will only affect object storage.

PR #3489, opened 2021-02-22
Use Credential from BSL for restic commands 

zubron:
This change introduces support for restic to make use of per-BSL
credentials. It makes use of the `credentials.FileStore` introduced in
PR #3442 to write the BSL credentials to disk. To support per-BSL
credentials for restic, the environment for the restic commands needs to
be modified for each provider to ensure that the credentials are
provided via the correct provider specific environment variables.
This change introduces a new function `restic.CmdEnv` to check the BSL
provider and create the correct mapping of environment variables for
each provider.

Previously, AWS and GCP could rely on the environment variables in the
Velero deployments to obtain the credentials file, but now these
environment variables need to be set with the path to the serialized
credentials file if a credential is set on the BSL.

For Azure, the credentials file in the environment was loaded and parsed
to set the environment variables for restic. Now, we check if the BSL
has a credential, and if it does, load and parse that file instead.

This change also introduces a few other small improvements. Now that we
are fetching the BSL to check for the `Credential` field, we can use the
BSL directly to get the `CACert` which means that we can remove the
`GetCACert` function. Also, now that we have a way to serialize secrets
to disk, we can use the `credentials.FileStore` to get a temp file for
the restic repo password and remove the `restic.TempCredentialsFile`
function.

Part of fixing #3304.

### DIFF
--- a/pkg/restic/exec_commands.go
+++ b/pkg/restic/exec_commands.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 the Velero contributors.
+Copyright the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -47,14 +47,8 @@ type backupStatusLine struct {
 // GetSnapshotID runs a 'restic snapshots' command to get the ID of the snapshot
 // in the specified repo matching the set of provided tags, or an error if a
 // unique snapshot cannot be identified.
-func GetSnapshotID(repoIdentifier, passwordFile string, tags map[string]string, env []string, caCertFile string) (string, error) {
-	cmd := GetSnapshotCommand(repoIdentifier, passwordFile, tags)
-	if len(env) > 0 {
-		cmd.Env = env
-	}
-	cmd.CACertFile = caCertFile
-
-	stdout, stderr, err := exec.RunCommand(cmd.Cmd())
+func GetSnapshotID(snapshotIdCmd *Command) (string, error) {
+	stdout, stderr, err := exec.RunCommand(snapshotIdCmd.Cmd())
 	if err != nil {
 		return "", errors.Wrapf(err, "error running command, stderr=%s", stderr)
 	}
--- a/pkg/restic/exec_commands.go
+++ b/pkg/restic/exec_commands.go
@@ -44,9 +44,8 @@ type backupStatusLine struct {
 	TotalBytesProcessed int64 `json:"total_bytes_processed"`
 }
 
-// GetSnapshotID runs a 'restic snapshots' command to get the ID of the snapshot
-// in the specified repo matching the set of provided tags, or an error if a
-// unique snapshot cannot be identified.
+// GetSnapshotID runs provided 'restic snapshots' command to get the ID of a snapshot
+// and an error if a unique snapshot cannot be identified.
 func GetSnapshotID(snapshotIdCmd *Command) (string, error) {
 	stdout, stderr, err := exec.RunCommand(snapshotIdCmd.Cmd())
 	if err != nil {
--- a/site/content/docs/main/locations.md
+++ b/site/content/docs/main/locations.md
@@ -23,7 +23,7 @@ This configuration design enables a number of different use cases, including:
 
 - Velero supports multiple credentials for `BackupStorageLocations`, allowing you to specify the credentials to use with any `BackupStorageLocation`.
   However, use of this feature requires support within the plugin for the object storage provider you wish to use.
-  All [plugins provided by the Velero team][5] support this feature.
+  All [plugins maintained by the Velero team][5] support this feature.
   If you are using a plugin from another provider, please check their documentation to determine if this feature is supported.
 
 - Velero only supports a single set of credentials for `VolumeSnapshotLocations`.
@@ -186,7 +186,7 @@ Please see the [earlier example][9] for details on how to create multiple `Backu
 
 #### Prerequisites
 - This feature requires support from the [object storage provider plugin][5] you wish to use.
-  All plugins provided by the Velero team support this feature.
+  All plugins maintained by the Velero team support this feature.
   If you are using a plugin from another provider, please check their documentation to determine if this is supported.
 - The [plugin for the object storage provider][5] you wish to use must be [installed][6].
 - You must create a file with the object storage credentials. Follow the instructions provided by your object storage provider plugin to create this file.
@@ -198,17 +198,17 @@ kubectl create secret generic -n velero credentials --from-file=bsl=</path/to/cr
 ```
 
 This will create a secret named `credentials` with a single key (`bsl`) which contains the contents of your credentials file.
-The name and key of this secret will be given to Velero when creating the `BackupStorageLocation` using the flag `--credential`.
-When interacting with that `BackupStorageLocation`, it will fetch the data from that key within that secret.
+Next, create a `BackupStorageLocation` that uses this Secret by passing the Secret name and key in the `--credential` flag.
+When interacting with this `BackupStroageLocation` in the future, Velero will fetch the data from the key within the Secret you provide.
 
-For example, a new `BackupStorageLocation` AWS with the created secret would be configured as follows:
+For example, a new `BackupStorageLocation` with a Secret would be configured as follows:
 
 ```bash
 velero backup-location create <bsl-name> \
-  --provider aws \
+  --provider <provider> \
   --bucket <bucket> \
   --config region=<region> \
-  --credential=credentials=bsl
+  --credential=<secret-name>=<key-within-secret>
 ```
 
 The `BackupStorageLocation` is ready to use when it has the phase `Available`.
@@ -219,6 +219,24 @@ velero backup-location get
 ```
 
 To use this new `BackupStorageLocation` when performing a backup, use the flag `--storage-location <bsl-name>` when running `velero backup create`.
+You may also set this new `BackupStorageLocation` as the default with the command `velero backup-location set --default <bsl-name>`.
+
+### Modify the credentials used by an existing storage location
+
+By default, `BackupStorageLocations` will use the credentials provided at install time and stored in the `cloud-credentials` secret in the Velero namespace.
+You can modify these existing credentials by [editing the `cloud-credentials` secret][10], however, these changes will apply to all locations using this secret.
+This may be the desired outcome, for example, in the case where you wish to rotate the credentials used for a particular account.
+
+You can also opt to modify an existing `BackupStorageLocation` such that it uses its own credentials by using the `backup-location set` command.
+
+If you have a credentials file that you wish to use for a `BackupStorageLocation`, follow the instructions above to create the Secret with that file in the Velero namespace.
+
+Once you have created the Secret, or have an existing Secret which contains the credentials you wish to use for your `BackupStorageLocation`, set the credential to use as follows:
+
+```bash
+velero backup-location set <bsl-name> \
+  --credential=<secret-name>=<key-within-secret>
+```
 
 ## Additional Use Cases
 
@@ -237,3 +255,4 @@ To use this new `BackupStorageLocation` when performing a backup, use the flag `
 [7]: https://kubernetes.io/docs/concepts/configuration/secret/
 [8]: #create-a-storage-location-that-uses-unique-credentials
 [9]: #have-some-velero-backups-go-to-a-bucket-in-an-eastern-usa-region-default-and-others-go-to-a-bucket-in-a-western-usa-region
+[10]: https://kubernetes.io/docs/concepts/configuration/secret/#editing-a-secret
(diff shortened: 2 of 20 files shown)


zubron:
This PR depends on #3442 but the restic specific commit can be reviewed: https://github.com/vmware-tanzu/velero/pull/3489/commits/9ccc9eb289f1e4f1ec278507b0b8515537aa8a27

I've tested this locally with a Tilt setup with a BSL with custom creds created for each provider (AWS, Azure, GCP) and was able to successfully backup and restore a pod with restic volumes from each. I'm going to focus on docs and getting E2E tests ready.

zubron:
I am working on an E2E test for this change, but can add that in a follow up PR or it can be reviewed as a separate commit on this PR.

I've marked this as ready for review now, so the main code changes and docs can be reviewed as those are ready.

site/content/docs/main/troubleshooting.md
```diff
@@ -168,6 +174,31 @@ Follow the below troubleshooting steps to confirm that Velero is using the corre
...
+
+   # Confirm that the secret exists
+   kubectl -n velero get secret $BSL_SECRET
+
+   # Print the content of the secret and ensure it is correct
+   kubectl -n velero get secret $BSL_SECRET -ojsonpath={.data.$BSL_SECRET_KEY} | base64 --decode
+   ```
+   If the secret can't be found, the secret does not within the Velero namespace and must be created.
+
+   If no output is produced when printing the contents of the secret, the key within the secret may not exist or may have no content.
```
a-mccarthy, line 199: Is there a recommended action to take here? Like recreating the secret, or updating it to have content? I mean its a little obvious, but for completeness  we should make sure the instructions are there and maybe include a link to how to update secrets? or replace them, etc.

site/content/docs/main/troubleshooting.md
```diff
@@ -168,6 +174,31 @@ Follow the below troubleshooting steps to confirm that Velero is using the corre
...
+   BSL_SECRET=$(kubectl get backupstoragelocations.velero.io -n velero <bsl-name> -o yaml -o jsonpath={.spec.credential.name})
+   BSL_SECRET_KEY=$(kubectl get backupstoragelocations.velero.io -n velero <bsl-name> -o yaml -o jsonpath={.spec.credential.key})
+
+   # Confirm that the secret exists
+   kubectl -n velero get secret $BSL_SECRET
+
+   # Print the content of the secret and ensure it is correct
+   kubectl -n velero get secret $BSL_SECRET -ojsonpath={.data.$BSL_SECRET_KEY} | base64 --decode
+   ```
+   If the secret can't be found, the secret does not within the Velero namespace and must be created.
```
a-mccarthy, line 197: a small typo, `If the secret can't be found, the secret does not within the Velero` should be `If the secret can't be found, then secret does not exist within the Velero`

site/content/docs/main/locations.md
```diff
@@ -166,6 +171,40 @@ During backup creation:
...
+When interacting with that `BackupStorageLocation`, it will fetch the data from that key within that secret.
+
+For example, a new `BackupStorageLocation` AWS with the created secret would be configured as follows:
+
+```bash
+velero backup-location create <bsl-name> \
+  --provider aws \
+  --bucket <bucket> \
+  --config region=<region> \
+  --credential=credentials=bsl
```
a-mccarthy, line 196: Should the `credentials` here match the name of the secret you created above?

site/content/docs/main/locations.md
```diff
@@ -166,6 +171,40 @@ During backup creation:
 velero backup create full-cluster-backup
 ```
 
+### Create a storage location that uses different credentials
+
+If your object storage provider plugin supports it, it is possible to create additional `BackupStorageLocations` that use their own credentials.
```
a-mccarthy, line 176: `If your object storage provider plugin supports it`

I'm not clear on whats needs to be supported? How would i know that my plugin supports this?

site/content/docs/main/locations.md
```diff
@@ -166,6 +171,40 @@ During backup creation:
 velero backup create full-cluster-backup
 ```
 
+### Create a storage location that uses different credentials
```
a-mccarthy, line 174: I have a few questions for this section

1. If you create a Backup storage location without specifying credentials in this way, what credentials are used? Or is it required now to always set different creds when creating a backup storage location? 
2. Can you use the secret for more than one BSL? Would you not want to do that?

Maybe think about making the title more specific? right not it seems a little vague to me? maybe something more like "Create a BackupStorageLocation with unique credentials"?

site/content/docs/main/locations.md
```diff
@@ -16,12 +16,17 @@ The user can pre-configure one or more possible `BackupStorageLocations` and one
...
 - Take snapshots of more than one kind of persistent volume in a single Velero backup. For example, in a cluster with both EBS volumes and Portworx volumes
-- Have some Velero backups go to a bucket in an eastern USA region, and others go to a bucket in a western USA region
+- Have some Velero backups go to a bucket in an eastern USA region, and others go to a bucket in a western USA region, or to a different storage provider
 - For volume providers that support it, like Portworx, you can have some snapshots stored locally on the cluster and have others stored in the cloud
 
 ## Limitations / Caveats
 
-- Velero only supports a single set of credentials *per provider*. It's not yet possible to use different credentials for different locations, if they're for the same provider.
+- Velero supports multiple credentials for `BackupStorageLocations`, allowing you to specify the credentials to use with any `BackupStorageLocation`.
+  However, use of this feature requires support from object storage provider plugin.
```
a-mccarthy, line 25: i'm not the clearest on the plugin breakdown, so please feel free to ignore me, but to me this reads a little confusingly. Is object storage provider plugin a feature of the plugins Velero supports? or how we refer to the plugins themselves?

a-mccarthy (changes requested):
@zubron Went through and added some comments. Hopefully they make sense :)

site/content/docs/main/locations.md
zubron: Yep. I tried to tie this with line 185-186 above, do you think it needs some extra clarification? I found this a bit tricky to explain.

site/content/docs/main/locations.md
zubron: This confused me for quite a while too. Velero has different [plugin kinds](https://velero.io/docs/v1.5/custom-plugins/#plugin-kinds) and each of the plugin provider repos can provide multiple plugins. For example, if you look at the overview of plugins, you can see labels for which plugin kinds are available in each repo: https://velero.io/plugins/

I can definitely rephrase this sentence though to try and make it clearer :+1:

site/content/docs/main/locations.md
zubron: If no credentials are specified, it will default to using the ones provided at install time. We have an example for that use case above so I've linked to it here and added more detail about what is happening in both of those cases. I've also linked to this new section from the existing example. You can use the same credential for more that one BSL, for example if you want to store to multiple regions but it's in the same account, as in the example earlier on this page. I've also updated the title as you suggested :)

zubron:
Thanks for the great feedback, @a-mccarthy! I think I've addressed most of your comments but this will likely need more improvements. If you have further comments, I'll make issues from those and address them in a follow up PR so that we can get the code changes in and start doing some testing for the 1.6 release.

pkg/restic/exec_commands.go
```diff
@@ -47,14 +47,8 @@ type backupStatusLine struct {
...
 // unique snapshot cannot be identified.
-func GetSnapshotID(repoIdentifier, passwordFile string, tags map[string]string, env []string, caCertFile string) (string, error) {
-	cmd := GetSnapshotCommand(repoIdentifier, passwordFile, tags)
-	if len(env) > 0 {
-		cmd.Env = env
-	}
-	cmd.CACertFile = caCertFile
-
-	stdout, stderr, err := exec.RunCommand(cmd.Cmd())
+func GetSnapshotID(snapshotIdCmd *Command) (string, error) {
```
carlisia, line 50: Minor: there's no longer tags being passed in and also this function signature changed enough that making a change to the godoc would probably be useful.

carlisia (changes requested):
This looks solid and I appreciate the refactorings. I have only ridiculously minor requests.


Questions:
- was this tested with the 3 providers we maintain? I didn't run manual tests.
- did you think through any possible backwards compatibility issues?


Also, yes, any e2e test can be added in a new PR imo.

site/content/docs/main/locations.md
```diff
@@ -166,6 +176,50 @@ During backup creation:
...
+```
+
+The `BackupStorageLocation` is ready to use when it has the phase `Available`.
+You can check the status with the following command:
+
+```bash
+velero backup-location get
+```
+
+To use this new `BackupStorageLocation` when performing a backup, use the flag `--storage-location <bsl-name>` when running `velero backup create`.
```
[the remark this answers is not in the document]
zubron: Yes! I had forgotten about that aspect of it and remembered while preparing #3561. It's now on my list of things to add :)

site/content/docs/main/locations.md
a-mccarthy: I think if its something the user needs to replace, the we should encase it in `<>` like the other inputs or make sure to call out in the paragraph before make sure to say replace `credentials` with the name of your secret

site/content/docs/main/locations.md
a-mccarthy: resolving since i think you addressed this

site/content/docs/main/locations.md
```diff
@@ -178,3 +232,8 @@ velero backup create full-cluster-backup
 [2]: api-types/volumesnapshotlocation.md
 [3]: https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure/blob/main/volumesnapshotlocation.md
 [4]: https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure/blob/main/backupstoragelocation.md
+[5]: /plugins
+[6]: overview-plugins.md
+[7]: https://kubernetes.io/docs/concepts/configuration/secret/
+[8]: #create-a-storage-location-that-uses-unique-credentials
```
a-mccarthy, line 238: I dont think the links with the # are working correctly, i can't tell if this is bc there is an issue with the preview, or they will also not work on the live site?

zubron: Ahhh, I totally thought these worked :woman_facepalming: okay, let me figure out how to fix this!

site/content/docs/main/locations.md
```diff
@@ -166,6 +176,50 @@ During backup creation:
...
+- You must create a file with the object storage credentials. Follow the instructions provided by your object storage provider plugin to create this file.
+
+Once you have installed the necessary plugin and created the credentials file, create a [Kubernetes Secret][6] in the Velero namespace that contains these credentials:
+
+```shell
+kubectl create secret generic -n velero credentials --from-file=bsl=</path/to/credentialsfile>
+```
+
+This will create a secret named `credentials` with a single key (`bsl`) which contains the contents of your credentials file.
+The name and key of this secret will be given to Velero when creating the `BackupStorageLocation` using the flag `--credential`.
```
a-mccarthy, line 201: I think we can be more direct/specific here, so maybe something like, "Next, create a `BackupStorageLocation` that uses this Secret by passing the Secret name in the `--credential` flag. When interacting with a this `BackupStroageLocation` in the future, Velero will fetch the data from the key within the Secret you provide."

a-mccarthy (dismissed):

a-mccarthy:
@zubron LGTM! Great write up! i had a few comments, but they are relatively minor. Please let me know if you want to me look at anything else :)

carlisia (approved):
👍

nrb (approved):

zubron:
Thanks, @carlisia, these are great questions.

> * was this tested with the 3 providers we maintain? I didn't run manual tests.

Yep, I tested this locally against all the 3 plugins that I made changes for. I did that testing as part of the testing for those plugins, and have run the newly added E2E test with Azure and AWS. I will do another test with GCP for completeness.

> * did you think through any possible backwards compatibility issues?

My focus with backwards compatibility was to ensure that we were continuing to set the same environment variables for each restic command. I compared the set of env vars before and after my change and checked they were the same (but with new values for the new credentials files). Previously, all restic commands used the environment from `os.Environ` with the provider specific vars for restic added. Here, we are also using `os.Environ` but replacing the provider specific vars that would usually be set within the Velero container spec.

For example, in the AWS case, `AWS_SHARED_CREDENTIALS_FILE` is already set, as it is defined in the container spec and wouldn't be modified as part of getting the S3 specific env vars. Now, we override that env var for the restic command based on the contents of the `Credential` field in the BSL. If it is not set, we rely on the original value defined in the container environment. Note that overriding these env vars is specific to that particular restic command instance. We do not override the env vars for the main Velero process. (There is an exception here with setting the env for Azure, as it involves overriding environment variables in the current process but that is existing behaviour and is not changed by this PR. It is something that will need to be fixed in future though if we move to parallel backups/restores.)

@carlisia Let me know if you have any further questions or concerns! Thanks for bringing these up.

carlisia:
Thanks @zubron!

Issue #3573, opened 2021-03-12
GH Action bug: If reviewers already assigned, more keep being added


carlisia:
Issue with our auto assign PR reviewers GH action: It works fine until there's an additional operation to the PR, like changing its state to/from WIP, or, very common, adding additional commits.

This happens now that we are limiting the number of reviewers to a subset of the total.

This is a known issue with the GH Action:

https://github.com/kentaro-m/auto-assign-action/issues/16

I'll keep an eye for this fix and update it in our repos.

zubron:
Could we minimise the impact of this by changing the trigger to [not include `synchronize`](https://github.com/vmware-tanzu/velero/blob/main/.github/workflows/auto_assign_prs.yml#L6)? My understanding is that by including that event, this job will run again on every push.

I think we'll still encounter this issue if we re-run the checks for the PR, but this should at least stop it being triggered on every push.

zubron:
Looking at our actions, we also have the [PR labeler job within the file for assigning PRs](https://github.com/vmware-tanzu/velero/blob/main/.github/workflows/auto_assign_prs.yml#L17-L24). I think we want _that_ job to run on every push, so we should separate that out into a different action.

carlisia:
Def worth it trying the first, and def a yes for the second. Let me try to improve this. Thanks for taking a look @zubron.

jenting:
Close it, feel free to reopen if the issue still happens again.

PR #3584, opened 2021-03-13
Improve GH Action PR assign + labeling

carlisia:

This change removes the `synchronize` parameter from the `pull_request_target` GH action in the `Auto Assign PR Reviewers` workflow. This should avoid the workflow from being triggered on all PR events, such as additional commits, which has the effect of adding additional reviewers (with each additional commit) beyond the current limit of 2. We only want this to be triggered when the PR opens, or when the PR changes from WIP to "ready for review".

It also separates the auto labeling GH action into the `"Auto Label PRs"` workflow. This one we want to be triggered on every event, so it keeps the `synchronize` and all other parameters the same.

Thanks to @zubron for pointing these out (in https://github.com/vmware-tanzu/velero/issues/3573#issuecomment-797677976). Hopefully this will work.

Fixes #(https://github.com/vmware-tanzu/velero/issues/3573)

PS: It doesn't **completely** fix the entire issue. For example: if the author wants to assign 2 specific reviewers before the PR is created, the GH action will still assign additional reviewers on the "open" event. This is the known bug in the action. 

A way to get around this is to remove the originally assigned reviewers after the PR is created, and add the ones you want (assuming they don't match). This "works", but causes unnecessary notifications for the soon-to-be-removed reviewers.

A way to get around wanting to

### DIFF
--- a/.github/workflows/auto_assign_prs.yml
+++ b/.github/workflows/auto_assign_prs.yml
@@ -3,7 +3,7 @@ name: "Auto Assign PR Reviewers"
 # This should mean PRs from forks are supported.
 on:
   pull_request_target:
-    types: [opened, reopened, synchronize, ready_for_review]
+    types: [opened, reopened, ready_for_review]
 
 jobs:
   # Automatically assigns reviewers and owner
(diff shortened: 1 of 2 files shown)


carlisia:
Changing this to draft. Let me do some tests on a personal repo. It seems that w/o the `synchronize` param events such as additional commits will cause the GH action to fail bc there's an existing process but no event to handle it. This makes sense now that I think about it.

zubron (approved):
Thanks!

jenting (approved):

Issue #2856, opened 2020-08-22
v1.5 CRDs are not compatible with Kubernetes v1.11 and earlier


albertollamaso:
**What steps did you take and what happened:**
Installed velero version  v1.5.0-beta.1

tried to install with:

```
velero install \
    --provider aws \
    --use-restic \
    --plugins velero/velero-plugin-for-aws:v1.1.0 \
    --bucket kubernetes.test.velero \
    --backup-location-config region=us-east-1 \
    --snapshot-location-config region=us-east-1 \
    --secret-file ./credentials-velero
```

it failed with:

`
CustomResourceDefinition/backups.velero.io: attempting to create resource
CustomResourceDefinition/backups.velero.io: already exists, proceeding
CustomResourceDefinition/backups.velero.io: created
CustomResourceDefinition/backupstoragelocations.velero.io: attempting to create resource
An error occurred:

Error installing Velero.Use `kubectl logs deploy/velero -n velero`
to check the deploy logs: Error creating resource CustomResourceDefinition / backupstoragelocations.velero.io: CustomResourceDefinition.apiextensions.k8s.io "backupstoragelocations.velero.io"
is invalid: spec.validation.openAPIV3Schema: Invalid value: apiextensions.JSONSchemaProps {
	ID: "",
	Schema: "",
	Ref: ( * string)(nil),
	Description: "BackupStorageLocation is a location where Velero stores backup objects",
	Type: "object",
	Format: "",
	Title: "",
	Default: ( * apiextensions.JSON)(nil),
	Maximum: ( * float64)(nil),
	ExclusiveMaximum: false,
	Minimum: ( * float64)(nil),
	ExclusiveMinimum: false,
	MaxLength: ( * int64)(nil),
	MinLength: ( * int64)(nil),
	Pattern: "",
	MaxItems: ( * int64)(nil),
	MinItems: ( * int64)(nil),
	UniqueItems: false,
	MultipleOf: ( * float64)(nil),
	Enum: [] apiextensions.JSON(nil),
	MaxProperties: ( * int64)(nil),
	MinProperties: ( * int64)(nil),
	Required: [] string(nil),
	Items: ( * apiextensions.JSONSchemaPropsOrArray)(nil),
	AllOf: [] apiextensions.JSONSchemaProps(nil),
	OneOf: [] apiextensions.JSONSchemaProps(nil),
	AnyOf: [] apiextensions.JSONSchemaProps(nil),
[...]

PR #3614, opened 2021-03-23
Bump CRD API version to v1

### LINKED ISSUE #2505: Update velero CRDs to v1
We're currently generating and installing `v1beta1` CRDs for Velero, mainly for backwards compatibility. The apiextensions API group graduated to `v1` as of Kubernetes 1.16, and the `v1beta1` version of the group is scheduled to be dropped in an upcoming Kubernetes release.

At some point (before `v1beta1` is dropped for good) we need to updated our generated CRDs to be `v1`. There's an existing PR (#2190) that we may want to dust off.

Additionally, it'd be nice to take advantage of some of the `v1` features, like [defaulting](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#defaulting). 

Tagging this for v1.5 for now -- not sure yet if it's must-do for then or not.

jenting:

- Generate both v1beta1 and v1 CRDs.
- Add a new flag `--crds-version` to allow the user to specify which CRDs version to render. If the Kubernetes CRDs preferred API version is discoverable, use the discovered version.

Fixes #2505


ashish-amarnath:
@jenting Considering Velero is used for upgrading from older Kubernetes versions to newer versions we will want to be able to Velero going back to kubernetes 1.12. For that reason, we might want to hold off on this PR.
I know that v1beta1 CRDs are deprecated and going to be removed soon, but I think we still have some time for this.

jenting:
Make sense. Also, it seems like we need to update the Velero prerequisite on Kubernetes version, ref to
- https://github.com/vmware-tanzu/velero/issues/2856#issuecomment-769864002
- https://velero.io/docs/v1.5/basic-install/#prerequisites

nrb:
[Kubernetes v1.22](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22) is where the v1beta1 APIs will be completely removed.

I'd like to make progress on this, but I think we'll be looking at leveraging [Carvel](https://carvel.dev/) to do so.

We have https://github.com/vmware-tanzu/velero/issues/2505 to track the work.

.github/workflows/crds-verify-kind.yaml
```diff
@@ -52,7 +52,6 @@ jobs:
       matrix:
         # Latest k8s versions. There's no series-based tag, nor is there a latest tag.
         k8s:
-          - 1.15.12
```
nrb, line 55: Please keep the v1.15 version here.

jenting: IIRC, the k8s 1.15 does not support CRD v1 version.

pkg/cmd/cli/install/install.go
```diff
@@ -128,6 +133,7 @@ func NewInstallOptions() *InstallOptions {
 		UseVolumeSnapshots:      true,
 		NoDefaultBackupLocation: false,
 		CRDsOnly:                false,
+		CRDsVersion:             "v1beta1",
```
sseago, line 136: I wonder whether we should default to "v1" instead of "v1beta1" since it's the preferred version for recent kube versions, even before 1.22. This would require users on kube 1.15 and older to override this, rather than requiring users on current kubernetes to override it.

sseago (commented):
More of a question at this point than a specific request for changes, but should we use v1 as the default rather than v1beta1?

Also, I'd like to have the flexibility to choose a CRD apiversion when running e2e tests. I actually made a local change to add that option in so I could test running e2e tests with the new CRDs. It might be easier for me to just submit that as a follow-on PR rather than trying to get that added here. Without that, e2e tests just use the default CRDsVersion from `NewInstallOptions()`

pkg/cmd/cli/install/install.go
zubron: I agree with this. I think if we switch the default now, then we only need to give one notice to users: The default is v1, if you need a previous version, you can override it. Otherwise we would need to instruct Kubernetes 1.22+ users to override now, and then tell both user groups to either add or drop the override once the defaults are switched.

sseago: @zubron Does it make sense to ignore e2e tests in this PR (nothing here breaks them), and I'll submit a follow-on PR to add CRDS_VERSION as an option (so you can do `CRDS_VERSION=v1beta1 make test-e2e` I already have the code for it in my local checkout.

sseago:
@zubron @jenting I went ahead and created a separate PR with my e2e test changes: https://github.com/vmware-tanzu/velero/pull/3941

That PR has to wait until this one is merged, since it depends on the new InstallOptions CRDsVersion field.

jenting:
Thank you @sseago for quickly go ahead, appreciate.

sseago (dismissed):

pkg/cmd/cli/install/install.go
```diff
@@ -246,9 +253,30 @@ This is useful as a starting point for more customized installations.
...
+				schema.GroupVersionResource{
+					Group:    "apiextensions.k8s.io",
+					Resource: "customresourcedefinitions",
+				})
+			if err != nil {
+				return err
+			}
+
+			// Update the group apiextensions.k8s.io preferred API version
+			o.CRDsVersion = gvr.Version
```
zubron, line 273: With this, the API server preferred version will always take precedence over the setting provided by the user with the `--crds-version` flag. Is there a case where the user may want to explicitly use v1beta1 when the preferred version is v1?

Also, in the case where Velero can't find the server preferred version, we should output a warning or log to state that the preferred version couldn't be found and that instead we will use `v1` as the default and that it can be changed with the `--crds-version` flag. Otherwise, I think we may encounter users who think that the CRD installation is broken on pre-1.16 clusters if the discovery fails.

zubron (changes requested):
There are further changes needed in this PR to ensure that the uninstall behaviour works in 1.22. Currently our uninstall code uses the v1beta1 API to fetch the CRDs to delete: https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/cli/uninstall/uninstall.go#L127-L154

This will result in an error when running on Kubernetes 1.22:
```
$ v uninstall --kubecontext kind-1.22 
You are about to uninstall Velero.
Are you sure you want to continue (Y/N)? y
Velero namespace "velero" does not exist, skipping.
Velero ClusterRoleBinding "velero" does not exist, skipping.
Errors while attempting to uninstall Velero: "no matches for kind \"CustomResourceDefinition\" in version \"apiextensions.k8s.io/v1beta1\""An error occurred: no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
```

I'm not sure of the best way to solve this. I don't know if there is a generic way to get the CRDs regardless of the API version used, or whether we need to attempt to delete both v1beta1 and v1 CRDs (checking for the appropriate errors from the API server).

zubron:
@jenting After discussing with @dsu-igeek, we decided to go ahead and merge this branch as is and @dsu-igeek will work on fixing the uninstall issue. I am going to rebase this branch on top of main and push it to resolve the merge conflicts. Thank you so much for taking care of this issue! :smile:

zubron (approved):
Thank you again, @jenting!

dsu-igeek (approved):
LGTM

jenting:
> There are further changes needed in this PR to ensure that the uninstall behaviour works in 1.22. Currently our uninstall code uses the v1beta1 API to fetch the CRDs to delete: https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/cli/uninstall/uninstall.go#L127-L154
> 
> This will result in an error when running on Kubernetes 1.22:
> 
> ```
> $ v uninstall --kubecontext kind-1.22 
> You are about to uninstall Velero.
> Are you sure you want to continue (Y/N)? y
> Velero namespace "velero" does not exist, skipping.
> Velero ClusterRoleBinding "velero" does not exist, skipping.
> Errors while attempting to uninstall Velero: "no matches for kind \"CustomResourceDefinition\" in version \"apiextensions.k8s.io/v1beta1\""An error occurred: no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
> ```
> 
> I'm not sure of the best way to solve this. I don't know if there is a generic way to get the CRDs regardless of the API version used, or whether we need to attempt to delete both v1beta1 and v1 CRDs (checking for the appropriate errors from the API server).

Sorry, I missed the `velero uninstall` command.
In general, if we delete CRDs by the native `kubectl` command, the kubectl gets the preferred CRDs API version. So, we don't have to have extra work on it.
But I believe we call Kubernetes client-go functions to delete the CRDs.

Issue #3616, opened 2021-03-23
Region is calculated incorrectly when gov account is used.


jala-dx:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is, and what commands you ran.)

Create a Bucket in gov. https://console.amazonaws-us-gov.com/
Create a Bucket with **same name** in non-gov https://s3.console.aws.amazon.com/

Restic ignores the 'config.region' specified in the BackupStorageLocation object. Even though the bucketname is unique globally, it is not unique across 'gov and non-gov' regions. Hence during snapshot creation, it accesses the bucketname in the incorrect region and fails at https://github.com/vmware-tanzu/velero/blob/main/pkg/restic/config.go#L75

https://github.com/vmware-tanzu/velero/blob/main/pkg/restic/config.go#L69

apiVersion: velero.io/v1
kind: BackupStorageLocation
metadata:
  creationTimestamp: "2021-03-17T01:12:01Z"
  generation: 24
  labels:
    component: velero
  name: default
  namespace: velero
  resourceVersion: "4607288"
  uid: 39aeddd0-f4d4-4f92-a296-4ea28da1489b
spec:
  config:
    **region: us-gov-east-1**
    s3Url: https://s3.us-gov-east-1.amazonaws.com
  objectStorage:
    bucket: xxxx-snapshot
    prefix: eastgov
  provider: aws
status:
  lastSyncedTime: "2021-03-23T19:37:00Z"
  lastValidationTime: "2021-03-23T19:37:00Z"
  phase: Available

**What did you expect to happen:**
When region is passed, restic should pick the region provided in BackupStorageLocation Spec instead of invoking getAWSBucketRegion https://github.com/vmware-tanzu/velero/blob/main/pkg/restic/config.go#L75

**The output of the following commands will help us better understand what's going on**:
(Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)

- `kubectl logs deployment/velero -n velero`
- `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
- `velero backup logs <backupname>`
- `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
- `velero restore logs <restorename>`

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Velero version (use `velero version`): 

- Velero features (use `velero client config get features`): 
- Kubernetes version (use `kubectl version`):
[...]

PR #3617, opened 2021-03-23
use region specified in the BackupStorageLocation spec to detect the Bucket region

jala-dx:

Fixes #(issue)
#3616 

### DIFF
--- a/changelogs/unreleased/3617-jala-dx
+++ b/changelogs/unreleased/3617-jala-dx
@@ -1,0 +1,1 @@
+Region is calculated incorrectly when gov account is used.
--- a/pkg/restic/config.go
+++ b/pkg/restic/config.go
@@ -64,10 +64,14 @@ func getRepoPrefix(location *velerov1api.BackupStorageLocation) (string, error) 
 	switch backendType {
 	case AWSBackend:
 		var url string
+		var region string
 		switch {
 		// non-AWS, S3-compatible object store
 		case location.Spec.Config["s3Url"] != "":
 			url = location.Spec.Config["s3Url"]
+		case location.Spec.Config["region"] != "":
+			region = location.Spec.Config["region"]
+			url = fmt.Sprintf("s3-%s.amazonaws.com", region)
 		default:
 			region, err := getAWSBucketRegion(bucket)
 			if err != nil {


pkg/restic/config.go
```diff
@@ -64,10 +64,14 @@ func getRepoPrefix(location *velerov1api.BackupStorageLocation) (string, error)
...
 	case AWSBackend:
 		var url string
+		var region string
 		switch {
 		// non-AWS, S3-compatible object store
 		case location.Spec.Config["s3Url"] != "":
 			url = location.Spec.Config["s3Url"]
+		case location.Spec.Config["region"] != "":
+			region = location.Spec.Config["region"]
+			url = fmt.Sprintf("s3-%s.amazonaws.com", region)
```
ashish-amarnath, line 74: This can be simplified to keep all the region inferring/getting logic together by moving this into the `default` case.
```
default:
    region := location.Spec.Config["region"]
    var err error
    if region == "" {
        region, err = getASBucketRegion(bucket)
        if err != nil {
		url = "s3.amazonaws.com"
		break
	}
	url = fmt.Sprintf("s3-%s.amazonaws.com", region)
    }
```

changelogs/unreleased/3617-jala-dx
```diff
@@ -0,0 +1 @@
+Region is calculated incorrectly when gov account is used.
```
ashish-amarnath, line 1: Can you please use the PR title as the contents of the changelog?

ashish-amarnath (changes requested):
This LGTM.
I have a couple of comments.
Also, please add unit tests for this change.

zubron:
Hi @jala-dx! Are you able to make the requested changes from @ashish-amarnath? It would be great to get this PR merged :) Thanks!

pkg/restic/config.go
dsu-igeek: Let's get rid of the switch statement altogether and just use an if/else clause.  I don't think we want any additional cases here anyhow.

dsu-igeek (changes requested):

zubron:
I tried to make some changes to this branch so we could get this merged but unfortunately maintainers don't have write permission for the branch. I've got another branch locally which includes 40a1b62c09f2a1bdb38b05a0439119f613f0ba22, so I will close this PR and create a new one with the original commit from this PR and my follow up changes.

PR #3618, opened 2021-03-24
Make uninstall more robust and informative

### LINKED ISSUE #3599: Bug with the Velero uninstall cmd
There are some major and minor problems with the uninstall command: https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/cli/uninstall/uninstall.go#L42-L74

- At each operation, it returns if there is an API error. This has a high potential of leaving behind resources that are still in the cluster only because the code didn't get to the point where they are removed. The solution is to aggregate all (real) errors and return them only at the end.
- An API error is not necessarily an error in this context. Example: a namespace (or any other resource) not found returns an API error, but it should not be considered an actual error of the cmd. In this case a log message is sufficient. Other error types should be distinguished and returned/aggregated as an error.
- The [wait](https://github.com/vmware-tanzu/velero/blob/e9c997839ebbb04688a88ffb981eeb74de51b380/pkg/cmd/cli/uninstall/uninstall.go#L81) for the namespace to be terminated is a Kubernetes anti-pattern.  If this was added because of the e2e tests, then the test is the piece that needs to wait for the resource to be deleted before proceeding, not the cli cmd. Also, it's weird that the uninstall takes way longer than the install. :/ - If this was added for the benefit of the user, again, it's an anti pattern to make the user wait for a resource to change state but if we want to aid the user we can log a message of the ns phase (see below).
- Any new API calls should be added using the kubebuilder/controller runtime framework.

I'm addressing the code issues above + modifying e2e tests that rely on this uninstall.

This is what running `velero uninstall` twice in a row without any (actual) errors could look like:

```
❯ velero uninstall
Velero uninstalled

❯ velero uninstall
Velero installation namespace "velero" is terminating.
Velero installation rolebinding "velero" does not exist, skipping.
Velero CRDs do not exist, skipping.
Velero uninstalled
```

carlisia:

- used the same existing API clients as the original code
- aggregated the errors until all operations are done
- added more informative messages 
- added confirmation
- added --wait (for the namespace termination) and --force (to be script friendly and skip confirmation)
- the output from --wait will look the same to the user as the --wait for creating backups

image

PS: 2e2 tests pass on Kind, testing it against AWS..

Fixes #3599

### DIFF
--- a/pkg/cmd/cli/uninstall/uninstall.go
+++ b/pkg/cmd/cli/uninstall/uninstall.go
@@ -60,13 +60,13 @@ func NewCommand(f client.Factory) *cobra.Command {
 	c := &cobra.Command{
 		Use:   "uninstall",
 		Short: "Uninstall Velero",
-		Long: `Uninstall Velero along with the CRDs.
+		Long: `Uninstall Velero along with the CRDs and clusterrolebinding.
 
 The '--namespace' flag can be used to specify the namespace where velero is installed (default: velero).
 Use '--wait' to wait for the Velero uninstall to be ready before proceeding.
 Use '--force' to skip the prompt confirming if you want to uninstall Velero.
 		`,
-		Example: `# velero uninstall -n staging`,
+		Example: ` # velero uninstall --namespace staging`,
 		Run: func(c *cobra.Command, args []string) {
 
 			// Confirm if not asked to force-skip confirmation
@@ -115,7 +115,7 @@ func Run(ctx context.Context, client *kubernetes.Clientset, extensionsClient *ap
 	crb := install.ClusterRoleBinding(namespace)
 	if err := client.RbacV1().ClusterRoleBindings().Delete(ctx, crb.Name, metav1.DeleteOptions{}); err != nil {
 		if apierrors.IsNotFound(err) {
-			fmt.Printf("Velero installation rolebinding %q does not exist, skipping.\n", crb.Name)
+			fmt.Printf("Velero installation clusterrolebinding %q does not exist, skipping.\n", crb.Name)
 		} else {
 			errs = append(errs, errors.WithStack(err))
 		}
--- a/pkg/cmd/cli/uninstall/uninstall.go
+++ b/pkg/cmd/cli/uninstall/uninstall.go
@@ -147,16 +147,21 @@ func Run(ctx context.Context, client *kubernetes.Clientset, extensionsClient *ap
 		defer cancel()
 
 		checkFunc := func() {
-			nsUpdated, _ := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+			ns, _ := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
 			if err != nil {
+				if apierrors.IsNotFound(err) {
+					fmt.Printf("Unexpectedly not able to retrieve namespace %q for checking its status. Halting.\n", namespace)
+					errs = append(errs, errors.WithStack(err))
+					cancel()
+				}
 				errs = append(errs, errors.WithStack(err))
 			}
 
-			if nsUpdated.Status.Phase == corev1.NamespaceTerminating {
+			if ns.Status.Phase == corev1.NamespaceTerminating {
 				fmt.Print(".")
 			}
 
-			if nsUpdated.Status.Phase != corev1.NamespaceTerminating {
+			if ns.Status.Phase != corev1.NamespaceTerminating {
 				fmt.Print("\n")
 				cancel()
 			}
--- a/pkg/cmd/cli/uninstall/uninstall.go
+++ b/pkg/cmd/cli/uninstall/uninstall.go
@@ -147,12 +147,12 @@ func Run(ctx context.Context, client *kubernetes.Clientset, extensionsClient *ap
 		defer cancel()
 
 		checkFunc := func() {
-			ns, _ := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+			ns, err := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
 			if err != nil {
 				if apierrors.IsNotFound(err) {
-					fmt.Printf("Unexpectedly not able to retrieve namespace %q for checking its status. Halting.\n", namespace)
-					errs = append(errs, errors.WithStack(err))
+					fmt.Print("\n")
 					cancel()
+					return
 				}
 				errs = append(errs, errors.WithStack(err))
 			}
@@ -160,15 +160,9 @@ func Run(ctx context.Context, client *kubernetes.Clientset, extensionsClient *ap
 			if ns.Status.Phase == corev1.NamespaceTerminating {
 				fmt.Print(".")
 			}
-
-			if ns.Status.Phase != corev1.NamespaceTerminating {
-				fmt.Print("\n")
-				cancel()
-			}
 		}
 
 		wait.Until(checkFunc, 5*time.Millisecond, ctx.Done())
-
 	}
 
 	if kubeerrs.NewAggregate(errs) != nil {
--- a/pkg/cmd/cli/uninstall/uninstall.go
+++ b/pkg/cmd/cli/uninstall/uninstall.go
@@ -147,7 +147,7 @@ func Run(ctx context.Context, client *kubernetes.Clientset, extensionsClient *ap
 		defer cancel()
 
 		checkFunc := func() {
-			ns, err := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+			_, err := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
 			if err != nil {
 				if apierrors.IsNotFound(err) {
 					fmt.Print("\n")
@@ -156,10 +156,7 @@ func Run(ctx context.Context, client *kubernetes.Clientset, extensionsClient *ap
 				}
 				errs = append(errs, errors.WithStack(err))
 			}
-
-			if ns.Status.Phase == corev1.NamespaceTerminating {
-				fmt.Print(".")
-			}
+			fmt.Print(".")
 		}
 
 		wait.Until(checkFunc, 5*time.Millisecond, ctx.Done())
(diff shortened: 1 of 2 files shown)


pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -19,96 +19,157 @@ package uninstall
...
-	if err != nil {
-		return errors.WithMessagef(err, "Uninstall failed removing Velero namespace %s", veleroNamespace)
+// uninstallOptions collects all the options for uninstalling Velero from a Kubernetes cluster.
+type uninstallOptions struct {
+	wait, force bool
+}
+
+// BindFlags adds command line values to the options struct.
+func (o *uninstallOptions) BindFlags(flags *pflag.FlagSet) {
+	flags.BoolVar(&o.wait, "wait", o.wait, "Wait for Velero deployment to be ready. Optional.")
```
dsu-igeek, line 52: Let's change this to "Wait for Velero uninstall to complete. Optional." or something similar.

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -19,96 +19,157 @@ package uninstall
...
+func NewCommand(f client.Factory) *cobra.Command {
+	o := &uninstallOptions{}
+
+	c := &cobra.Command{
+		Use:   "uninstall",
+		Short: "Uninstall Velero",
+		Long: `Uninstall Velero along with the CRDs.
+
+The '--namespace' flag can be used to specify the namespace where velero is installed (default: velero).
+Use '--wait' to wait for the Velero Deployment to be ready before proceeding.
```
dsu-igeek, line 66: Same here about uninstall

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -19,96 +19,157 @@ package uninstall
...
-	for _, removeCRD := range crds.Items {
-		err = extensionsClient.ApiextensionsV1().CustomResourceDefinitions().Delete(ctx, removeCRD.ObjectMeta.Name, metav1.DeleteOptions{})
-		if err != nil {
-			return errors.WithMessagef(err, "Uninstall failed removing CRD %s", removeCRD.ObjectMeta.Name)
+	if len(crds.Items) == 0 {
+		fmt.Print("Velero CRDs do not exist, skipping.\n")
+	} else {
+		for _, removeCRD := range crds.Items {
+			if err = extensionsClient.ApiextensionsV1().CustomResourceDefinitions().Delete(ctx, removeCRD.ObjectMeta.Name, metav1.DeleteOptions{}); err != nil {
+				errs = append(errs, errors.WithStack(err))
```
dsu-igeek, line 137: Let's keep the message so we know which CRD it failed on.

"Uninstall failed removing CRD %s", removeCRD.ObjectMeta.Name

test/e2e/velero_utils.go
```diff
@@ -292,7 +292,12 @@ func VeleroInstall(ctx context.Context, veleroImage string, veleroNamespace stri
 }
 
 func VeleroUninstall(ctx context.Context, client *kubernetes.Clientset, extensionsClient *apiextensionsclient.Clientset, veleroNamespace string) error {
-	return uninstall.Uninstall(ctx, client, extensionsClient, veleroNamespace)
+	err := uninstall.Run(ctx, client, extensionsClient, veleroNamespace, true)
```
dsu-igeek, line 295: Do we want to use that if err := ...; err != nil pattern everywhere?

dsu-igeek (changes requested):
Looks good, just some fixes to the messages.

zubron (changes requested):
Thanks, @carlisia! This looks good, I really like the clearer feedback to the user about what's happening :+1:  I just have a few additional comments along with the ones already mentioned by Dave.

test/e2e/velero_utils.go
carlisia: Yes we do.

On a diff note, I don't know why I expanded the error checking. Changed it back to `return uninstall.Uninstall(...`.

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -19,96 +19,157 @@ package uninstall
...
+	c := &cobra.Command{
+		Use:   "uninstall",
+		Short: "Uninstall Velero",
+		Long: `Uninstall Velero along with the CRDs.
+
+The '--namespace' flag can be used to specify the namespace where velero is installed (default: velero).
+Use '--wait' to wait for the Velero Deployment to be ready before proceeding.
+Use '--force' to skip the prompt confirming if you want to uninstall Velero.
+		`,
+		Example: `# velero uninstall -n staging`,
```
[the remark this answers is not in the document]
carlisia: Hm. I don't mind changing it but the `install` cmd has it like that, with the # at the beginning of the cmd example. At the very least I'd like to make this mirror what's in the install. WDYT?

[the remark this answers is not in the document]
carlisia: I pushed a change to address your other feedback and can make another change depending on what you answer wrt ^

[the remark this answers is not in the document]
zubron: Ah, if it's like that in the install then it's fine to leave this. I saw the absence of `#` in other commands but obviously didn't look at install :)

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -19,96 +19,158 @@ package uninstall
...
-		return errors.WithMessagef(err, "Uninstall failed removing Velero namespace %s", veleroNamespace)
+// uninstallOptions collects all the options for uninstalling Velero from a Kubernetes cluster.
+type uninstallOptions struct {
+	wait, force bool
+}
+
+// BindFlags adds command line values to the options struct.
+func (o *uninstallOptions) BindFlags(flags *pflag.FlagSet) {
+	flags.BoolVar(&o.wait, "wait", o.wait, "Wait for Velero uninstall to be ready. Optional.")
+	flags.BoolVar(&o.force, "force", o.force, "Forces the Velero uninstall. Optional.")
```
ashish-amarnath, line 53: are we missing a flag parse for `--namespace`?

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -19,96 +19,158 @@ package uninstall
...
-		if err != nil {
-			// Commented this out because not sure if removing this is okay
-			// Printing this on Uninstall will lead to confusion
-			// fmt.Printf("Namespaces.Get after delete return err %v\n", err)
-			return true, nil // Assume any error means the delete was successful
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
+		checkFunc := func() {
+			nsUpdated, _ := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
```
ashish-amarnath, line 150:
```suggestion
			nsUpdated, err := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
```

ashish-amarnath: We should also suppress `NotFound` error here.

ashish-amarnath: The variable `nsUpdated` makes me feel like we are performing an update operation. Can this just be `ns`?

ashish-amarnath (changes requested):
LGTM overall.
I have a few minor comments.

pkg/cmd/cli/uninstall/uninstall.go
carlisia: `--namespace`` is a global flag for Velero, any subcommand will have access to it.

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -19,96 +19,163 @@ package uninstall
...
-			// fmt.Printf("Namespaces.Get after delete return err %v\n", err)
-			return true, nil // Assume any error means the delete was successful
+	if waitToTerminate && len(ns.Name) != 0 {
+		fmt.Println("Waiting for Velero uninstall to complete. You may safely press ctrl-c to stop waiting - uninstall will continue in the background.")
+
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
+		checkFunc := func() {
+			ns, _ := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
```
ashish-amarnath, line 150:
```suggestion
			ns, err := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
```

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -19,96 +19,163 @@ package uninstall
...
+			if err != nil {
+				if apierrors.IsNotFound(err) {
```
ashish-amarnath, line 152: not sure if this error scenario is unexpected.
It is possible that the namespace was fully deleted in the interval between two successive polls.

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -19,96 +19,163 @@ package uninstall
...
+				errs = append(errs, errors.WithStack(err))
+			}
+
+			if ns.Status.Phase == corev1.NamespaceTerminating {
+				fmt.Print(".")
+			}
+
+			if ns.Status.Phase != corev1.NamespaceTerminating {
+				fmt.Print("\n")
+				cancel()
```
ashish-amarnath, line 166: shouldn't we be calling `cancel()` only when the namespace is no longer present, i.e from line 115?

carlisia: You're absolutely right.

pkg/cmd/cli/uninstall/uninstall.go
carlisia: Fixed it. Moving the cancel() here, this is the right place for it, when the ns no longer is found.

pkg/cmd/cli/uninstall/uninstall.go
carlisia: 🤦🏻‍♀️

carlisia:
For reviewers: if you think the `return` on LN 155, you are thinking like me and expected the `cancel()` call to return but no, it continues that iteration til the end than considers it "done". The `return` was needed.

dsu-igeek (approved):

zubron (approved):
Thanks, @carlisia!

dsu-igeek (approved):
LGTM!

PR #3629, opened 2021-03-25
Runs vSphere tests with snapshots

dsu-igeek:
Added wait for vSphere plug-in uploads to complete

### DIFF
--- a/test/e2e/velero_utils.go
+++ b/test/e2e/velero_utils.go
@@ -10,6 +10,11 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	veleroexec "github.com/vmware-tanzu/velero/pkg/util/exec"
 
 	"github.com/pkg/errors"
 	"k8s.io/client-go/kubernetes"
@@ -32,7 +37,7 @@ func getProviderPlugins(providerName string) []string {
 	case "azure":
 		return []string{"velero/velero-plugin-for-microsoft-azure:v1.2.0"}
 	case "vsphere":
-		return []string{"velero/velero-plugin-for-aws:v1.2.0", "velero/velero-plugin-for-vsphere:v1.1.0"}
+		return []string{"velero/velero-plugin-for-aws:v1.2.0", "vsphereveleroplugin/velero-plugin-for-vsphere:1.1.0"}
 	default:
 		return []string{""}
 	}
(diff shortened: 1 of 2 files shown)


test/e2e/velero_utils.go
```diff
@@ -32,7 +37,7 @@ func getProviderPlugins(providerName string) []string {
 	case "azure":
 		return []string{"velero/velero-plugin-for-microsoft-azure:v1.2.0"}
 	case "vsphere":
-		return []string{"velero/velero-plugin-for-aws:v1.2.0", "velero/velero-plugin-for-vsphere:v1.1.0"}
+		return []string{"velero/velero-plugin-for-aws:v1.2.0", "vsphereveleroplugin/velero-plugin-for-vsphere:1.1.0"}
```
zubron, line 40: This needs to be reverted back to the original image name.

test/e2e/velero_utils.go
```diff
@@ -395,3 +420,51 @@ func VeleroAddPluginsForProvider(ctx context.Context, veleroCLI string, veleroNa
...
+				phase := comps[1]
+				if phase == "New" ||
+					phase == "InProgress" ||
+					phase == "Snapshotted" ||
+					phase == "Uploading" {
+					complete = false
+				}
+			}
+		}
+		return complete, nil
```
zubron, line 466: Do we want to return an error in the `SnapshotFailed`, `UploadFailed` or cancel phases? With the current use, the test will continue even if we fail to upload. That may be okay, but I just want to make sure that we still have visibility of this failure when the restore step in the test fails later.

zubron (changes requested):
Thanks, @dsu-igeek for this and adding the additional context in comments. There is one change to the vSphere image being used that looks like it needs to be reverted. I have a question about capturing snapshot upload failures but I don't think that should block this.

test/e2e/velero_utils.go
dsu-igeek: That is the correct image name and source for the vSphere plug-in.  The old one didn't work.  https://github.com/vmware-tanzu/velero-plugin-for-vsphere/blob/main/docs/vanilla.md

test/e2e/velero_utils.go
dsu-igeek: The status will get logged currently, we output that up above.  I didn't error out here because it's possible that the there are records floating around from previous backups and I didn't want us to start being flakey because stuff was hanging around.

test/e2e/velero_utils.go
```diff
@@ -272,16 +277,36 @@ func VeleroInstall(ctx context.Context, veleroImage string, veleroNamespace stri
 			return errors.New("No object store provider specified - must be specified when using kind as the cloud provider") // Gotta have an object store provider
 		}
 	}
+
+	providerPlugins := getProviderPlugins(objectStoreProvider)
+
+	// TODO - handle this better
+	if cloudProvider == "vsphere" {
```
carlisia, line 284: It seems to me this should be grouped with the if/else that starts on ln 270. Pls add a line before the call to `err := EnsureClusterExists(ctx)`. That call should actually be the very first thing in this func.

test/e2e/velero_utils.go
```diff
@@ -395,3 +420,51 @@ func VeleroAddPluginsForProvider(ctx context.Context, veleroCLI string, veleroNa
 
 	return nil
 }
+
+/*
+ Waits for uploads started by the Velero Plug-in for vSphere to complete
+ TODO - remove after upload progress monitoring is implemented
+*/
+func waitForVSphereUploadCompletion(ctx context.Context, timeout time.Duration, namespace string) error {
```
carlisia, line 428: Thanks for making this a private/unexported func.

carlisia (approved):
This LGTM. I made some comments that are only style related so if you ignore them is fine and can pick it up in the "clean up e2e tests" PR.

zubron (approved):

zubron:
@carlisia I'm going to go ahead and merge this since you've approved it. Like you said, we can fix it up in the follow up PR. Thanks!

PR #3641, opened 2021-03-30
Add Labels field to BackupSpec

### LINKED ISSUE #3560: Labels on backup objects
I see that the labels from schedules.velero.io are copied to backups.velero.io, which is not how things are usually done in Kubernetes.

A schedule object like this:
```
apiVersion: velero.io/v1
kind: Schedule
metadata:
  labels:
    foo: bar
  name: test-schedule
  namespace: velero
spec:
  schedule: '0 * * * *'
  template:
    hooks: {}
    includedNamespaces:
    - '*'
    ttl: 48h0m0s
``` 
Will generate backup objects like this:
```
apiVersion: velero.io/v1
kind: Backup
metadata:
  labels:
    foo: bar
    velero.io/backup: test-schedule-20210311110003
    velero.io/pv: ....
    velero.io/schedule-name: test-schedule
    velero.io/storage-location: default
  name: test-schedule-20210311110003
  namespace: velero
......
```
This is not how things are done in Kubernetes. For instance: The pods created by a deploy objects, gets labeled based on '.spec.metadata.labels', NOT '.metadata.labels'.

My problem with this, is that I would like to deploy everything on my Kubernetes clusters automatically - using Argo-CD. And Argo-CD is adding custom labels to all objects that it is managing, to keep track of them. And since Velero is copying those labels from the schedule object, to all backup objects, this tracking will be messed up.

I have tested this using both v1.2.0 and v1.5.3, without seeing any difference in this behaviour.

**Vote on this issue!**

arush-sal:
Signed-off-by: Arush Salil <me@aru.sh>

# Add Labels field to BackupSpec and to backup object

Fixes #3560

- [X] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.


KlavsKlavsen:
One note, if of any relevance :)
This PR was sponsored by BlackWood Seven (Davids company) and developed by Obmondo.com (where @arush-sal and I are from)
We will ofcourse work with Velero team to get this merged, so it benefits every Velero user.

changelogs/unreleased/3641-arush-sal
```diff
@@ -0,0 +1 @@
+Add Label to BackupSpec so that labels can explicitly be provided to Schedule.Spec.Template.Metadata.Labels which will be reflected on the backups created.
```
dsu-igeek, line 1: Please document how the existing behavior is being changed by this.

pkg/builder/backup_builder.go
```diff
@@ -77,6 +77,9 @@ func (b *BackupBuilder) ObjectMeta(opts ...ObjectMetaOpt) *BackupBuilder {
 
 // FromSchedule sets the Backup's spec and labels from the Schedule template
 func (b *BackupBuilder) FromSchedule(schedule *velerov1api.Schedule) *BackupBuilder {
+	if schedule.Spec.Template.Metadata.Labels != nil {
```
dsu-igeek, line 80: This is changing existing behavior and is going to be confusing.

Is the argument that copying the labels from the Schedule is a bad idea?  If so, we should probably just remove the part where it does copy.  If it's OK, why don't we merge the two?  If we want to preserve existing behavior but override (as the code is doing) we should log that we are overriding the Labels copy.

Let's not change the content of the schedule.Labels in here, that's unexpected behavior for the caller.

dsu-igeek (changes requested):

pkg/builder/backup_builder.go
KlavsKlavsen: I worked with arush-sal on this and the idea there was that IF you specific .spec.template.metadata.labels (which you can't do before this PR) - then you WANT to make use of the feature this PR provides - which means we switch from copying the schedules labels - to copying the one from the template (which is the method one would expect - which is also used with deployments etc. which have same setup).
How do you feel this is NOT mimicking the way everyones used to decide labels on "created objects" - by Deployments, statefulset etc. ?

changelogs/unreleased/3641-arush-sal
KlavsKlavsen: I've added docs here: https://github.com/vmware-tanzu/velero/pull/3641/files#diff-a521c2750535d66cd8cf8c68f1ee1e021dcecfd852b40de729c2cbc9c776b4d5 any where else you feel this should be documented?

pkg/builder/backup_builder.go
KlavsKlavsen: @dsu-igeek If you have better ways to solve the issue of schedule object labels being copied to backup objects - please share - and we'll gladly listen. In the end we just want "one solution" to be merged - so its possible to use argocd with velero :)
We did all feel that the logic of .spec.templates.metadata (and labels under that) - found in deployments, statefulsets etc. was what we expected when schedule object already HAS .spec.template, but we found that it did not at all implement the expected behavior :( - Thats why we choose this route of action.

dsu-igeek: Let's log what is happening ("Labels (%s) from Schedule.Spec.Template.Metadata.Labels are being applied to Backup %s", "Labels (%s) from Schedule are being applied to Backup %s" or something like that.

Please change the logic so that the in-memory schedule.Labels is not modified.  schedule is coming in as a pointer so the code is modifying the structure passed in.

site/content/docs/main/api-types/schedule.md
```diff
@@ -78,6 +78,11 @@ spec:
     # a default value of 30 days will be used. The default can be configured on the velero server
     # by passing the flag --default-backup-ttl.
     ttl: 24h0m0s
+    # The labels you want on backup objects, created from this schedule (instead of copying the labels you have on schedule object itself).
+    # Setting this will also ensure your backup object does not get labels from f.ex. argocd copied to your backup object (making argocd think it created it and wanting to remove it)
```
dsu-igeek, line 82: How about this instead of the second sentence?
# Setting this will also ensure your backup object does not get labels from f.ex. argocd copied to your backup object (making argocd think it created it and wanting to remove it) ->
# When this field is set, the labels from the Schedule resource are not copied to the Sackup resource.

changelogs/unreleased/3641-arush-sal
dsu-igeek: Those are good, thanks

site/content/docs/main/api-types/schedule.md
KlavsKlavsen: updated.

pkg/builder/backup_builder.go
```diff
@@ -77,7 +80,23 @@ func (b *BackupBuilder) ObjectMeta(opts ...ObjectMetaOpt) *BackupBuilder {
...
+	var labels = make(map[string]string)
+
+	// Check if there's explicit Labels defined in the Schedule object template
+	// and if present then copy it to the backup object.
+	if schedule.Spec.Template.Metadata.Labels != nil {
+		labels = schedule.Spec.Template.Metadata.Labels
+		logrus.WithFields(logrus.Fields{
+			"backup": fmt.Sprintf("%s/%s", b.object.GetNamespace(), b.object.GetName()),
+			"labels": schedule.Spec.Template.Metadata.Labels,
+		}).Debug("Schedule.template.metadata.labels set - using those labels instead of schedule.labels for backup object")
```
arush-sal, line 92: @dsu-igeek Since the logger hasn't been initialized, I doubt that this is going to work. Hence could you please let me know how you would like this to be achieved/implemented?

KlavsKlavsen:
I added https://github.com/Obmondo/helm-charts - a fork of velero helm-charts with updated CRD and using our image built with this PR for testing it.

pkg/builder/backup_builder.go
nrb: Perhaps the logger from https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/schedule_controller.go#L260 could be passed down?

arush-sal: @nrb I really don't think that it would work as the logger is available to any `controller object` but not to a `builder object` and hence there doesn't exist a logger to work with for the `BackupBuilder` object. Therefore I have initialized a new logger within the function scope for `FromSchedule` and have used that instead. Hope the approach makes sense to the team.

pkg/builder/backup_builder.go
arush-sal: @dsu-igeek the above has been addressed. Please have a look and let me know.

KlavsKlavsen:
@dsu-igeek Hasn't all requested changes been taken care of ? Why is it still saying one requested change has not been addressed?

pkg/builder/backup_builder.go
dsu-igeek, line 94: Let's make this Info level please, as normally Debug messages aren't output and this is a good heads up for when people start complaining about not getting the labels they expected on the backup

pkg/builder/backup_builder.go
```diff
@@ -77,7 +80,23 @@ func (b *BackupBuilder) ObjectMeta(opts ...ObjectMetaOpt) *BackupBuilder {
...
+	} else {
+		labels = schedule.Labels
+		logrus.WithFields(logrus.Fields{
+			"backup": fmt.Sprintf("%s/%s", b.object.GetNamespace(), b.object.GetName()),
+			"labels": schedule.Labels,
+		}).Debug("No Schedule.template.metadata.labels set - using Schedule.labels for backup object")
```
dsu-igeek, line 100: Please make this Info as well, thanks

dsu-igeek (changes requested):
Let's change those messages to Info level and then we're good, thanks!

pkg/builder/backup_builder.go
KlavsKlavsen: Done. Nobody should be surprised - as they wouldn't have set template.metadata.labels normally (current schedule object does not allow it) :)

pkg/builder/backup_builder.go
KlavsKlavsen: Done

pkg/builder/backup_builder.go
dsu-igeek: They might have already had labels set on the schedule that were getting passed through and then they set the new field and then they start asking "Why didn't the labels on the schedule get passed through too".  Thanks!

KlavsKlavsen:
> Let's change those messages to Info level and then we're good, thanks!

Done - Anything else, or is it ready to merge? the output from CI is "a bit confusing" :)

zubron (approved):
Thanks!

KlavsKlavsen:
@zubron it seems "some requested change by @dsu-igeek was not performed? But i cannot figure out which? any help appreciated.. :)

zubron:
@KlavsKlavsen I think that is just how the GitHub UI displays that the PR has not been reviewed since by the person who requested changes. I will re-request a review.

dsu-igeek (approved):
LGTM, thanks!

PR #3651, opened 2021-03-31
Add cherry-pick commits and changelog for v1.5.4

zubron:


zubron:
> Did something depend on [f58c1d63122](https://github.com/vmware-tanzu/velero/pull/3651/commits/f58c1d63122628c8d4612b4c8e9a2a55017d096b)?

Yes, I added this to fix the build. The CRD checks in the 1.5 branch [use an older version of the setup-kind action](https://github.com/vmware-tanzu/velero/blob/v1.5.3/.github/workflows/crds-verify-k8s-1-16-9.yaml#L9) which now fails as it uses a command that was deprecated and removed due to a security vulnerability. Here is an example of the failed build: https://github.com/vmware-tanzu/velero/runs/2239532596?check_suite_focus=true

If you prefer, I could just update the existing workflows to use the newest version of setup-kind. I opted for to cherry-pick to try and keep the branches a bit more in sync but I don't feel particularly strongly about it.

nrb:
I'd prefer to just use the cherry-pick, but wanted to double check. Thanks for the clarification.

nrb (approved):

carlisia (approved):

PR #3763, opened 2021-05-01
Added multiple namespace test

dsu-igeek:

Adds a multiple namespace backup/restore test.  There are Basic (2 namespaces) and Scale identifiers (2500 namespaces)

- [ X] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [X ] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.


test/e2e/multiple_namespaces_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+var _ = Describe("[Basic] Backup/restore of 2 namespaces", func() {
+	var (
+		client *kubernetes.Clientset
+	)
+
+	BeforeEach(func() {
+		var err error
```
zubron, line 24: The `uuidgen` variable is a global across the e2e package. We should be calling `uuidgen, err = uuid.NewRandom()` somewhere here as it is not initialised within this file unlike other tests. It means that if you run only the scale test, it isn't set and so all the characters of the UUID are `0`.

test/e2e/velero_utils.go
```diff
@@ -261,6 +258,30 @@ func veleroBackupNamespace(ctx context.Context, veleroCLI string, veleroNamespac
 	return err
 }
 
+// veleroBackupExcludeNamespaces uses the veleroCLI to backup a namespace.
+func veleroBackupExcludeNamespaces(ctx context.Context, veleroCLI string, veleroNamespace string, backupName string, excludeNamespaces []string) error {
+	namespaces := ""
+	for nsNum, curNamespace := range excludeNamespaces {
```
zubron, line 264: Could we use [`strings.Join`](https://golang.org/pkg/strings/#Join) here?

test/e2e/multiple_namespaces_test.go
```diff
@@ -0,0 +1,129 @@
...
+		if err := createNamespace(ctx, client, createNSName); err != nil {
+			return errors.Wrapf(err, "Failed to create namespace %s to install Kibishii workload", kibishiiNamespace)
+		}
+	}
+	if err := veleroBackupExcludeNamespaces(ctx, veleroCLI, veleroNamespace, backupName, excludeNamespaces); err != nil {
+		veleroBackupLogs(ctx, veleroCLI, "", backupName)
+		return errors.Wrapf(err, "Failed to backup backup namespaces %s-*", nsBaseName)
+	}
+
+	err = cleanupNamespaces(ctx, client, nsBaseName)
```
zubron, line 88: Is there a chance that not all the namespaces have been deleted before calling restore? I ran this locally and the test passed but I'm wondering if there is a chance it could fail on another machine. Perhaps we could apply a label to the newly created namespaces (based on `uuidgen`) and use that in a `List` call to make sure no namespaces with that label exist.

test/e2e/multiple_namespaces_test.go
```diff
@@ -0,0 +1,129 @@
...
+		return errors.Wrap(err, "Could not retrieve namespaces")
+	}
+
+	for _, excludeNamespace := range namespaces.Items {
+		excludeNamespaces = append(excludeNamespaces, excludeNamespace.Name)
+	}
+	for nsNum := 0; nsNum < numberOfNamespaces; nsNum++ {
+		createNSName := fmt.Sprintf("%s-%00000d", nsBaseName, nsNum)
+		if err := createNamespace(ctx, client, createNSName); err != nil {
+			return errors.Wrapf(err, "Failed to create namespace %s to install Kibishii workload", kibishiiNamespace)
```
zubron, line 80: I'd remove the reference to Kibishii here as we aren't installing any workloads in these namespaces.

zubron (changes requested):
Thanks for adding this! I just have a few minor comments. Also, these tests assume that Velero will be installed, which means we can't run them alongside other tests that install Velero as those tests will uninstall before moving to the next test.

test/e2e/multiple_namespaces_test.go
dsu-igeek: Mmmm...you're right, it's actually using the wrong namespace of what failed here

test/e2e/multiple_namespaces_test.go
dsu-igeek: Well, the K8s Delete call should return an error if it can't do it (cleanupNamespaces) so I'm thinking we don't need to be super paranoid.  If the tests start failing, we can look at other options, but I'm inclined to leave it for now.

test/e2e/multiple_namespaces_test.go
```diff
@@ -0,0 +1,160 @@
...
+			oneHourTimeout, _ := context.WithTimeout(context.Background(), 1*time.Hour)
+			RunMultipleNamespaceTest(oneHourTimeout, client, "nstest-"+uuidgen.String(), 2500,
+				backupName, restoreName)
+		})
+	})
+})
+
+func RunMultipleNamespaceTest(ctx context.Context, client testClient, nsBaseName string, numberOfNamespaces int, backupName string, restoreName string) error {
+	shortTimeout, _ := context.WithTimeout(ctx, 5*time.Minute)
+	defer cleanupNamespaces(ctx, client, nsBaseName) // Run at exit for final cleanup
```
zubron, line 95: If `cleanupNamespaces` returns an error, it will be ignored due to the `defer` here. I think it's still better to call the cleanup using `defer` as that will ensure that we attempt it even if the test fails mid way through.

Does it matter if the error is ignored? I don't think we should mark the test as failed if it couldn't clean up the namespaces, but maybe we could log the error doing something like the following:
```
defer func() {
  if err := cleanupNamespaces(ctx, client, nsBaseName); err != nil {
    fmt.Println("Error during namespace cleanup:", err)
  }
}
```

zubron (approved):
Thanks, @dsu-igeek! I've left a suggestion but I don't think it should block this being merged. If you choose to make changes, I'll review it again :)

test/e2e/velero_utils.go
dsu-igeek: Yup, thanks, changed that.

carlisia (approved):
LGTM 👍

PR #3779, opened 2021-05-10
remove all default value fields from crds

foosinn:

this commit removes all `nullable: true` and `preserveUnknownFields: false` mentions. 

These are default and not visible when dumping the resource from the Kubernetes API. That causes ArgoCD to always detect a diff.

### DIFF
--- a/changelogs/unreleased/3779-foosinn
+++ b/changelogs/unreleased/3779-foosinn
@@ -1,0 +1,1 @@
+remove all default value fields from crds to prevent gitops tools from detecting diffs
--- a/config/crd/bases/velero.io_backups.yaml
+++ b/config/crd/bases/velero.io_backups.yaml
@@ -14,7 +14,6 @@ spec:
     listKind: BackupList
     plural: backups
     singular: backup
-  preserveUnknownFields: false
   scope: Namespaced
   validation:
     openAPIV3Schema:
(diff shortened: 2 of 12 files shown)


jenting (commented):
I think we should change the code generation parameters to removes the CRDs default values field, not just directly _delete_ it.
https://github.com/vmware-tanzu/velero/blob/9e29e50/hack/update-generated-crd-code.sh#L47-L52

nrb:
Preserving unknown fields is not desirable behavior, in my opinion.

Why does Argo need that removed in order to detect a diff?

foosinn:
Hmm, this seems to be an issue with older clusters (in our case an Openshift 3.11, supportes until Jun 2022).

I still would like to see this merged. But i understand if you won't support that old kubernetes versions.

jenting:
What's the version of Kubernetes cluster?

sseago:
@jenting OpenShift 3.11 uses Kubernetes 1.11, so Velero doesn't technically support that version, although we have had success using Velero in OpenShift 3.x clusters by omitting the CRD validation when installing in clusters based on Kubernetes 1.11 or older.

dsu-igeek:
This seems like a bug in ArgoCD.

foosinn:
I guess it's best to close this then? We can go on and work around for the time until we have upgraded all clusters to 4.6.

jenting:
Is ArgoCD have a related GitHub issue about this?

From my perspective, it's more like the ArgoCD issue.
I'd prefer to close it, feel free to reopen if you have any concerns/feedback.

PR #3794, opened 2021-05-12
Add support for volume and PV HostPath when using Restic

StephenSorriaux:
Add a new flag "EnableHostPath" to accept HostPath during Restic backup and restore.

- Added a flag `EnableHostPath` in order to make Velero be OK when backuping or restoring a HostPath pod or PV volume. I don't want to create any breaking change, so I added this flag to force users to opt-in for this feature.
- Updated the `processRestore` and `processBackup` methods to fallback to a HostPath if volume is not found in `/host_pods/...` path and if the flag `EnableHostPath` is set
- Updated the `GetVolumeDirectory` to return Path when volume is a hostPath or a PV backed hostPath

This would require the user to make sure the backuped hostPath are available in the Restic DaemonSet pods.

I am personally testing it against a k3s installation (k8s 1.19) that uses [`local-path-provisioner`](https://github.com/rancher/local-path-provisioner) to create PV. Those PV are actually hostPaths, but all of them are created under the `/var/lib/rancher/k3s/storage/` directory so I am perfectly fine with mounted it inside the Restic pods.
FWIW, I successfully backuped and restored a ClickHouse instance using this (statefullset with hostPath PV). I will do some more tests.

Fixes #(issue)

### DIFF
--- a/pkg/util/kube/utils.go
+++ b/pkg/util/kube/utils.go
@@ -126,6 +126,10 @@ func GetVolumeDirectory(pod *corev1api.Pod, volumeName string, pvcLister corev1l
 		if volume.VolumeSource.CSI != nil {
 			return volume.Name + "/mount", nil
 		}
+		// Pod volume is actually a HostPath
+		if volume.VolumeSource.HostPath != nil {
+			return volume.VolumeSource.HostPath.Path, nil
+		}
 		return volume.Name, nil
 	}
 
@@ -143,6 +147,11 @@ func GetVolumeDirectory(pod *corev1api.Pod, volumeName string, pvcLister corev1l
 	// PV's been created with a CSI source.
 	if pv.Spec.CSI != nil {
 		return pvc.Spec.VolumeName + "/mount", nil
+	}
+
+	// PV's been created with a HostPath.
+	if pv.Spec.HostPath != nil {
+		return pv.Spec.HostPath.Path, nil
 	}
 
 	return pvc.Spec.VolumeName, nil
--- a/changelogs/unreleased/3794-StephenSorriaux
+++ b/changelogs/unreleased/3794-StephenSorriaux
@@ -1,0 +1,3 @@
+Add support for volume and PV HostPath when using Restic
+
+Add a new flag "EnableHostPath" to accept HostPath during backup and restore
--- a/pkg/apis/velero/v1/constants.go
+++ b/pkg/apis/velero/v1/constants.go
@@ -46,4 +46,7 @@ const (
 
 	// APIGroupVersionsFeatureFlag is the feature flag string that defines whether or not to handle multiple API Group Versions
 	APIGroupVersionsFeatureFlag = "EnableAPIGroupVersions"
+
+	// HostPathFlag is the feature flag string that defines whether or not HostPath in PV will be backuped
+	HostPathFlag = "EnableHostPath"
 )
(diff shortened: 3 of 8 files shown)


gianlazz:
I came here looking for a solution to a `Volume xxx-data in pod default/xxx-deployment-69dfdf9c99-z2jtr is a hostPath volume which is not supported for restic backup, skipping` warning when attempting to back up my k3os machine and it sounds like this PR would resolve the issue.

pkg/util/kube/utils.go
```diff
@@ -129,6 +129,11 @@ func GetVolumeDirectory(pod *corev1api.Pod, volumeName string, pvcLister corev1l
 		return volume.Name, nil
 	}
 
+	// Pod volume is actually a HostPath
+	if volume.HostPath != nil {
```
sseago, line 133: This section would need to be above line 125 above. Where it is now, `volume.HostPath` will never be non-nil, since any non-PVC volume source would be true above and this method will have returned `volume.Name` already.

sseago: Actually, scratch that. You probably want this if block *inside* the `if volume.VolumeSource.PersistentVolumeClaim == nil` block, right after the CSI VolumeSource check.

sseago (commented):
I think this is an interesting idea and it may be worth considering, although I'd be interested in hearing what others on the Velero team think. I'm not completely clear on the issue of making sure the hostPath directory is mounted in the restic DaemonSet. Is this something that would just need to be documented, and any user enabling this would be expected to handle this before using Velero?

pkg/util/kube/utils.go
StephenSorriaux: Good point, I will change that, thank you.

StephenSorriaux:
I am glad to read it might be worth it, hopefully other team members will think the same!

Concerning the hostPath directory that needs to be mounted in the Restic DaemonSet, yes with this PR this is something that needs to be documented (I will add the docs) and yes the user will be expected to take care of mounting all hostPaths he wants to be backuped/restored. Basically it is like the `host-pods` volume that [the user may need to update](https://velero.io/docs/v1.6/restic/#configure-restic-daemonset-spec).

StephenSorriaux:
Hi again,

I have performed more tests and can now confirm that this PR will make possible to handle:
- hostPath volumes (tested using a basic pod);
- hostPath backed PVC volumes (tested using a ClickHouse instance created with a ClickHouseOperator).

I have added a changelog file and some documentation.

For those willing to do some tests on their side without having to build everything, I have pushed a Docker image that can be used: `ssorriaux/velero:hostpath-support`

@sseago @zubron I am not sure how to proceed next, is there any chance for the Velero team to have a look at it? Thank you.

StephenSorriaux:
Hello,

A little message to just say that I have this running in production for almost 3 weeks now without any issue. I also have been able to restore the backuped data to another environment and made sure I have something consistent.

(Gentle bump of this PR also.)

Thank you.

dsu-igeek (commented):
This feature may not work or could even be dangerous in the general case.  This assumes that all nodes have the same layout since the DaemonSet will be configured with the same host paths for all nodes and there may be differences.  We'll be making other changes in the future that will probably make this unusable as well.  We suggest that you maintain your own branch to take care of this special case.  Please feel free to visit us during a community meeting for additional discussion.

dsu-igeek:
Closing for now, please discuss with us in the community meeting if you'd like

PR #3823, opened 2021-05-24
Allow backup list requests to be chunked

### LINKED ISSUE #262: Support retrieving chunked lists of items when backing up
Kuberentes 1.9 adds support for chunked lists. We should do this. See https://github.com/kubernetes/website/pull/6540/files#diff-dd99945fd91d7ea569799a45d59669f0R90.

dharmab:

Add a `--client-page-size` flag to the Velero server. This flag causes LIST calls for resources other than Namespaces to be chunked into chunks of the given size during backups (i.e. using the `limit` query parameter on the apiserver request). The default value of `0` disables chunking, similar to the behavior of the `--chunk-size` flag for kubectl.

In the case of a ResourceExpired error due to modification of the object list during pagination, the server will attempt to fall back on a non-chunked list. If this request fails, the backup will fail.

I have tested this in my own cluster using `client-page-size` of both `0` and `500` successfully.

~~Note: I was unable to run most of the `make` commands as noted in https://github.com/vmware-tanzu/velero/issues/262#issuecomment-847198861. I may need help from a maintainer to complete this pull request.~~ This worked after a rebase.

Fixes #262 (except for Namespace objects)

### DIFF
--- a/pkg/backup/backup.go
+++ b/pkg/backup/backup.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 the Velero contributors.
+Copyright the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -73,6 +73,7 @@ type kubernetesBackupper struct {
 	resticBackupperFactory restic.BackupperFactory
 	resticTimeout          time.Duration
 	defaultVolumesToRestic bool
+	clientPageSize         int
 }
 
 type resolvedAction struct {
@@ -106,6 +107,7 @@ func NewKubernetesBackupper(
 	resticBackupperFactory restic.BackupperFactory,
 	resticTimeout time.Duration,
 	defaultVolumesToRestic bool,
+	clientPageSize int,
 ) (Backupper, error) {
 	return &kubernetesBackupper{
 		backupClient:           backupClient,
@@ -115,6 +117,7 @@ func NewKubernetesBackupper(
 		resticBackupperFactory: resticBackupperFactory,
 		resticTimeout:          resticTimeout,
 		defaultVolumesToRestic: defaultVolumesToRestic,
+		clientPageSize:         clientPageSize,
 	}, nil
 }
 
@@ -272,6 +275,7 @@ func (kb *kubernetesBackupper) Backup(log logrus.FieldLogger, backupRequest *Req
 		dynamicFactory:        kb.dynamicFactory,
 		cohabitatingResources: cohabitatingResources(),
 		dir:                   tempDir,
+		pageSize:              kb.clientPageSize,
 	}
 
 	items := collector.getAllItems()
--- a/pkg/backup/item_collector.go
+++ b/pkg/backup/item_collector.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 the Velero contributors.
+Copyright the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@ limitations under the License.
 package backup
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -25,10 +26,13 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/pager"
 
 	"github.com/vmware-tanzu/velero/pkg/client"
 	"github.com/vmware-tanzu/velero/pkg/discovery"
@@ -45,6 +49,7 @@ type itemCollector struct {
 	dynamicFactory        client.DynamicFactory
 	cohabitatingResources map[string]*cohabitatingResource
 	dir                   string
+	pageSize              int
 }
 
 type kubernetesResource struct {
@@ -275,6 +280,7 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
 	var items []*kubernetesResource
 
 	for _, namespace := range namespacesToList {
+		// List items from Kubernetes API
 		log = log.WithField("namespace", namespace)
 
 		resourceClient, err := r.dynamicFactory.ClientForGroupVersionResource(gv, resource, namespace)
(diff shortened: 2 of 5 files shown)


dharmab:
Might be better to use https://pkg.go.dev/k8s.io/client-go/tools/pager#ListPager.List- with the layer of abstraction from the dynamic client it didn't seem like the most trivial option.

alaypatel07:
Using ListPager could be a good option, kube controllers used to use this before a refactor to informers system. One later bloomer being worked on right now is the cronjob controller. Example of list page in that controller here: https://github.com/kubernetes/kubernetes/blob/cf59c68e15f35a074e46ba6eae704702d3833e5d/pkg/controller/cronjob/cronjob_controller.go#L113

Note: a trivial difference between the cronjob example and here would be we might have to handle the resource expired type errors in the page function like follows, but otherwise +1 to using the abstraction.

```
	resourceListFunc := func(opts metav1.ListOptions) (runtime.Object, error) {
		obj, listErr := resourceClient.Namespace(namespace).List(context.Background(), opts)
		switch {
		case errors.IsResourceExpired(listErr):
			// log the warning
		default:
			return nil, listErr
		}
		return obj, nil
	}
```
Infact I would think this could be another configurable parameter, the user could choose to tolerate expired list or pay the performance benefit with an accurate snapshot from apiserver directly if they so care.

pkg/backup/item_collector.go
```diff
@@ -289,16 +291,43 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
...
+			for continueValue != "" {
+				unstructuredList, err := resourceClient.List(metav1.ListOptions{LabelSelector: labelSelector, Limit: int64(r.listLimit), Continue: continueValue})
+				if statusError, isStatus := err.(*k8serrors.StatusError); isStatus && statusError.Status().Reason == metav1.StatusReasonExpired {
+					log.Warnf("Got ResourceExpired error while listing items. Resource list may be incomplete due to modifications while listing items")
+				} else if err != nil {
+					log.WithError(errors.WithStack(err)).Error("Error listing items")
+					break
+				}
+				unstructuredItems = append(unstructuredItems, unstructuredList.Items...)
+				continueValue = unstructuredList.GetContinue()
```
alaypatel07, line 322: as mentioned in the other comments, would it make sense to retry implementing this with the client-go Pager type?

dharmab: I did that this morning! I've got a patch using the pager locally I'm testing out today. If my functional tests look good expect a push later today.

dharmab:
Hmm, trying to resolve conflicts seems to have caused some problems, probably because I branched from 1.6.0 instead of master. Time to rebase entirely...

pkg/backup/item_collector.go
```diff
@@ -287,18 +292,50 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
...
+		listOptions := metav1.ListOptions{LabelSelector: labelSelector}
 
 		log.Info("Listing items")
-		unstructuredList, err := resourceClient.List(metav1.ListOptions{LabelSelector: labelSelector})
-		if err != nil {
-			log.WithError(errors.WithStack(err)).Error("Error listing items")
-			continue
+		unstructuredItems := make([]unstructured.Unstructured, 0)
+
+		if r.pageSize == 0 {
```
alaypatel07, line 300: change this to `r.pageSize > 0`?

pkg/backup/item_collector.go
```diff
@@ -287,18 +292,50 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
...
+			unstructuredItems = append(unstructuredItems, unstructuredList.Items...)
+			if err != nil {
+				log.WithError(errors.WithStack(err)).Error("Error listing items")
+				continue
+			}
+		} else {
+			// If limit is positive, use a pager to split list over multiple requests
+			// Use Velero's dynamic list function instead of the default
+			listFunc := pager.SimplePageFunc(func(opts metav1.ListOptions) (runtime.Object, error) {
+				list, err := resourceClient.List(listOptions)
```
alaypatel07, line 312: if you are not planning on checking the type of error, simply `return resourceClient.List(listOptions)` might work as well.

Also, can you move the TODO on line 320 here? I think the `ResourceExperied` errors will be handled here in the future.

pkg/backup/item_collector.go
dharmab: This branch should only trigger if pageSize is 0. This is the special "disable paging" value, mirroring the same behavior in kubectl:

```
$ kubectl get -h | grep chunk-size
      --chunk-size=500: Return large lists in chunks rather than all at once. Pass 0 to disable. This flag is beta and may change in the future.
```

The branch for a non-zero value begins at the `else` below.

pkg/backup/item_collector.go
```diff
@@ -287,18 +292,50 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
...
+				}
+				return list, nil
+			})
+			listPager := pager.New(listFunc)
+			// Use the page size defined in the server config
+			// TODO allow configuration of page buffer size and handling of ResourceExpired errors
+			listPager.PageSize = int64(r.pageSize)
+			// Add each item to unstructuredItems
+			err := listPager.EachListItem(context.Background(), listOptions, func(object runtime.Object) error {
+				item := object.(*unstructured.Unstructured)
```
alaypatel07, line 324: check if the type is actually unstructured? otherwise this will panic
```
				item, ok := object.(*unstructured.Unstructured)
                                 if !ok {
                                 // log error
                                     return nil
                                 }
```

alaypatel07 (commented):
Overall this looks good to me, added a few suggestions inline. Thanks @dharmab 

cc @sseago

pkg/backup/item_collector.go
alaypatel07: oops sorry for miscommunication but I meant:
```
if r.pageSize > 0 {
// use the pager
} else {
// zero or negative
// use the normal list-all
}
```

That was you also validate against a user sending in negative values by mistake and dont break

pkg/backup/item_collector.go
dharmab: Page buffer size would be handled by setting `listPager.PageBufferSize`.

I double checked how `ResourceExpired` errors need to be handled and it looks like `ListPager` only handles them for us if we use `List()` instead of `EachListItem()` :( So this needs to be changed to handle the error after all.

dharmab:
Updated to handle ResourceExpired errors (since EachListItem doesn't do that for us), explicitly handle a casting error, and invert the page size condition to close the negative page size loophole.

pkg/backup/backup.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 the Velero contributors.
+Copyright 2021 the Velero contributors.
```
jenting, line 2: We use `Copyright the Velero contributors.` now.

pkg/backup/item_collector.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 the Velero contributors.
+Copyright 2021 the Velero contributors.
```
jenting, line 2: We use `Copyright the Velero contributors.` now.

pkg/backup/backup.go
dharmab: The [contributing guide code standards](https://velero.io/docs/v1.6/code-standards/#copyright-header) state a year is required. If this should be changed, perhaps a separate PR should update the code standard and update the header across the project.

jenting: Thank you for the information, yes we should update the doc.

jenting: #3830

dharmab:
May this PR have an updated review following the squashed commit/header updates?

site/content/docs/main/backup-reference.md
```diff
@@ -60,3 +60,11 @@ velero backup create --from-schedule example-schedule
 ```
 
 This command will immediately trigger a new backup based on your template for `example-schedule`. This will not affect the backup schedule, and another backup will trigger at the scheduled time.
+
+## Enable Kubernetes API Pagination
+
+By default, Velero uses a single LIST API call for each resource type from the Kubernetes API when collecting items into a backup. This provides a fully consistent list. However, on clusters with a very large number of objects, the performance of this API call can be degraded and eventually be timed out by the Kubernetes apiserver.
```
dsu-igeek, line 66: Let's rewrite this to reflect a default chunk size of 500.  There's no need to mention consistency since the code will handle an expired and fall back to the regular list so we should get a consistent list no matter what.

dsu-igeek (changes requested):
Thanks, look pretty good, let's default to 500 chunk size.  We don't want people to have to go looking for this option when they run into issues.

dsu-igeek:
Also, please run E2E tests against this.  Ping me if you need help.

dharmab:
I'm seeing a lot of errors when running `make test-e2e` similar to:

```
Namespace/: attempting to create resource
Namespace/: attempting to create resource client

• Failure in Spec Setup (BeforeEach) [2.455 seconds]
[Restic] Velero tests on cluster using the plugin provider for object storage and Restic for volume backups
/Users/bellamko/go/src/github.com/vmware-tanzu/velero/test/e2e/backup_test.go:36
  when kibishii is the sample workload [BeforeEach]
  /Users/bellamko/go/src/github.com/vmware-tanzu/velero/test/e2e/backup_test.go:82
    should be successfully backed up and restored to the default BackupStorageLocation
    /Users/bellamko/go/src/github.com/vmware-tanzu/velero/test/e2e/backup_test.go:83

    Expected success, but got an error:
        <*errors.withMessage | 0xc000331b40>: {
            cause: {
                error: {
                    cause: {
                        error: {
                            cause: {
                                ErrStatus: {
                                    TypeMeta: {Kind: "Status", APIVersion: "v1"},
                                    ListMeta: {
                                        SelfLink: "",
                                        ResourceVersion: "",
                                        Continue: "",
                                        RemainingItemCount: nil,
                                    },
                                    Status: "Failure",
                                    Message: "Namespace \"\" is invalid: metadata.name: Required value: name or generateName is required",
                                    Reason: "Invalid",
                                    Details: {
                                        Name: "",
                                        Group: "",
                                        Kind: "Namespace",
                                        UID: "",
                                        Causes: [
                                            {Type: ..., Message: ..., Field: ...},
                                        ],
                                        RetryAfterSeconds: 0,
                                    },
                                    Code: 422,
                                },
                            },
                            msg: "Error creating resource Namespace/",
                        },
                        stack: [0x21e3d53, 0x21e48e5, 0x22342ab, 0x22360ff, 0x223f805, 0x21f4963, 0x21f457c, 0x21f4a47, 0x21f75ad, 0x21f70f2, 0x2205d71, 0x2205887, 0x2205077, 0x2207786, 0x22143b8, 0x2214067, 0x22389a9, 0x111d84f, 0x1073d81],
                    },
                    msg: "\n\nError installing Velero. Use `kubectl logs deploy/velero -n velero` to check the deploy logs",
                },
                stack: [0x2234505, 0x22360ff, 0x223f805, 0x21f4963, 0x21f457c, 0x21f4a47, 0x21f75ad, 0x21f70f2, 0x2205d71, 0x2205887, 0x2205077, 0x2207786, 0x22143b8, 0x2214067, 0x22389a9, 0x111d84f, 0x1073d81],
            },
            msg: "Failed to install Velero in cluster",
        }
        Failed to install Velero in cluster:

        Error installing Velero. Use `kubectl logs deploy/velero -n velero` to check the deploy logs: Error creating resource Namespace/: Namespace "" is invalid: metadata.name: Required value: name or generateName is required

    /Users/bellamko/go/src/github.com/vmware-tanzu/velero/test/e2e/backup_test.go:66
```

sseago:
@dharmab When running e2e tests, it looks like you're missing the velero namespace env var. Try adding `VELERO_NAMESPACE=velero` to your `make test-e2e` call.

dharmab:
A quick update here- I've continued to have some issues with running e2e-tests even after setting some of the undocumented variables in the e2e-test Makefile. I suspect my test cluster may differ from the standard ones used by the project. I'm also out on extended time off for much of June/July without much internet access so this PR may linger for a few weeks.

adriananeci:
What is needed for this PR to be merged? We are currently running a forked version(`v1.6.3` + this PR) and want to align back to upstream version.

reasonerjt (approved):
LGTM, thanks @dharmab

dsu-igeek (approved):
Looks good, thanks for the changes

PR #3845, opened 2021-06-04
regression introduced in 1.6 restore progress: fix CR restore

### LINKED ISSUE #3844: CR restore regression in Velero 1.6
A backup containing a CR (which pulls in the required CRD) will fail to restore the CR if the CRD is not already defined in the cluster prior to the restore running. This is a regression from Velero 1.5 where CR+CRD restore worked.

I expected the CR to be restored along with the CRD

Relevant log message:
```
time="2021-06-03T18:17:59Z" level=info msg="Skipping restore of resource because it cannot be resolved via discovery" logSource="pkg/restore/restore.go:1615" resource=foos.example.foo.com restore=openshift-migration/migration-d19c7-final-p5d5z
```
The CR is being skipped because at the time the validation is being done, the CRD isn't restored yet. This is a result of the progress restore refactoring. We now validate all resources before starting any restore actions, so that we can count the total number of expected items to restore. However, as a result of this, when we attempt to find the groupResource for the CR, it's not found (since the CRD hasn't been restored yet).

A PR to fix this will be forthcoming (I have a PR against our local fork as an initial attempt to fix this). The fix is to do two passes. The first pass restores *only* CRDs. Once that's done, a second pass restores everything. We include CRDs in that "everything" pass to preserve proper restore progress information, but of course Velero won't actually attempt to restore the CRDs twice since the restore context keeps a list of already-restored items to avoid duplicate restore attempts.

**Environment:**

- Velero version (use `velero version`): 1.6.0
- Kubernetes version (use `kubectl version`): 1.19 (OpenShift 4.6)

**Vote on this issue!**

sseago:

Implementing Restore progress changed the validation/restore processing order to pre-process and validate everything (and count the number of items to restore) before actually restoring anything. This broke CRD/CR restore, since CR validation requires the CRD to actually exist already. This change runs a first validate/restore pass which *only* restores CRDs, followed by a second pass to restore everything. This results in CRDs and CRs being restored properly while preserving the new restore progress functionality.

Fixes #3844

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -377,6 +377,10 @@ func getOrderedResources(resourcePriorities []string, backupResources map[string
 	return append(resourcePriorities, orderedBackupResources...)
 }
 
+type progressUpdate struct {
+	totalItems, itemsRestored int
+}
+
 func (ctx *restoreContext) execute() (Result, Result) {
 	warnings, errs := Result{}, Result{}
 
@@ -407,14 +411,6 @@ func (ctx *restoreContext) execute() (Result, Result) {
 				return warnings, errs
 			}
 		}
-	}
-
-	selectedResourceCollection, w, e := ctx.getOrderedResourceCollection(backupResources)
-	warnings.Merge(&w)
-	errs.Merge(&e)
-
-	type progressUpdate struct {
-		totalItems, itemsRestored int
 	}
 
 	update := make(chan progressUpdate)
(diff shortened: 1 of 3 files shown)


sseago:
In addition to fixing the bug, I've also added restore logging which tracks progress (which we already have for Backup progress). I included it in this PR since it helped me to verify my fix here as well.

dsu-igeek:
Would it be possible to add an E2E test that catches this?  Then we can avoid regressions in the future.

pkg/restore/restore.go
```diff
@@ -458,92 +454,51 @@ func (ctx *restoreContext) execute() (Result, Result) {
 	// totalItems: previously discovered items, i: iteration counter.
 	totalItems, i, existingNamespaces := 0, 0, sets.NewString()
 
-	for _, selectedResource := range selectedResourceCollection {
+	// First restore only CRDs
```
zubron, line 457: I think it would be useful to expand this comment to state why CRDs are getting special treatment here. Even just adding something like "... so that they are available in the cluster for validation of custom resource objects" would be enough.

pkg/restore/restore.go
```diff
@@ -458,92 +454,51 @@ func (ctx *restoreContext) execute() (Result, Result) {
...
-						ctx.restoredItems[itemKey] = struct{}{}
-					}
-
-					// Keep track of namespaces that we know exist so we don't
-					// have to try to create them multiple times.
-					existingNamespaces.Insert(selectedItem.targetNamespace)
-				}
+	for _, selectedResource := range crdResourceCollection {
+		var w, e Result
+		i, w, e = ctx.processSelectedResource(selectedResource, totalItems, i, existingNamespaces, update)
```
zubron, line 474: I'd like to rename `i` to something more descriptive as this is a little confusing to follow - would something like `totalRestoredItems` work? `totalItems` could also be renamed to something like `totalItemsToRestore` to help distinguish what these two variables are tracking. I think this would also help with where these values are used in `processSelectedResource`.

As a follow up, rather than passing in `i` and making it the responsibility of `processSelectedResource` to update the value, could that function instead just return its own count of processed resources/restored items and then that could be added to the running tally in this outer scope (currently represented by `i`)? e.g.

```
restoredItems, w, e := ctx.processSelectedResource(selectedResource, totalItems, existingNamespaces, update)
totalRestoredItems += restoredItems
```

pkg/restore/restore.go
```diff
@@ -458,92 +454,51 @@ func (ctx *restoreContext) execute() (Result, Result) {
...
-				// time, we don't want previously known items counted twice as
-				// they are present in both restoredItems and totalItems.
-				actualTotalItems := len(ctx.restoredItems) + (totalItems - i)
-				update <- progressUpdate{
-					totalItems:    actualTotalItems,
-					itemsRestored: len(ctx.restoredItems),
-				}
-			}
-		}
+	// reset i and totalItems after processing full resource list
```
zubron, line 490: Should this be "reset *before* processing full resource list"?

pkg/restore/restore.go
```diff
@@ -605,6 +560,104 @@ func (ctx *restoreContext) execute() (Result, Result) {
 	return warnings, errs
 }
 
+func (ctx *restoreContext) processSelectedResource(
```
zubron, line 563: Adding a comment for this function would be helpful to understand what it's doing.

pkg/restore/restore.go
```diff
@@ -605,6 +560,104 @@ func (ctx *restoreContext) execute() (Result, Result) {
...
+						err,
+					),
+				)
+				continue
+			}
+
+			w, e := ctx.restoreItem(obj, groupResource, selectedItem.targetNamespace)
+			warnings.Merge(&w)
+			errs.Merge(&e)
+			i++
```
zubron, line 628: Just to follow up here from one of the comments above, this is the variable that I think shouldn't be passed in, but instead should be initialised in this function, returned, and used to update the other variable in the calling scope.

zubron (commented):
Thanks for catching and fixing this, @sseago! I think this change looks good, there are just a few readability things that I think could be improved. I already found this existing logic tricky to follow so it would be great if we could help improve the readability with this change :)

pkg/restore/restore.go
sseago: @zubron On the whole, I think all of your suggestions are good ones to improve readability. Much of this is just moving around what was already in place, but the suggestions all make sense. One caveat is that this func actually needs the total processed, since the progress updates are issued from within it. SO we still need the "processed before this call" number (currently i) passed in. However, if it makes it clearer, rather than adding to it and returning it, we could just count within the func and return that, but then when reporting progress, instead of `i` we'll just use `previousRestoredItems+currentRestoredItems`.

zubron: Ahh, you're right. Sorry, I misunderstood this when reading it yesterday. I think in that case we can still pass it in to avoid changing the existing logic too much, but a more descriptive name would definitely help :)

sseago: Thinking about it again, I'm realizing that when it was just `i` before my change, while a more descriptive name would have been better, it was less important since it was just a local variable. Now that we're passing it into a func, it's actually pretty important to have a descriptive name. I'll update it.

pkg/restore/restore.go
sseago: So in digging into this further, `i` is actually not the number of restored items. It's an iterator count of processed items. If the item is already restored, then i increments while the number of items actually restored doesn't increase. Meanwhile, the actual list of restored items can increase outside of the loop iterations (namespaces added, plugin actions, etc.). I'm going to rename `i` to `processedItems` to make this clear.

pkg/restore/restore.go
```diff
@@ -605,6 +560,104 @@ func (ctx *restoreContext) execute() (Result, Result) {
...
+
+			// totalItems keeps the count of items previously known. There
+			// may be additional items restored by plugins. We want to include
+			// the additional items by looking at restoredItems at the same
+			// time, we don't want previously known items counted twice as
+			// they are present in both restoredItems and totalItems.
+			actualTotalItems := len(ctx.restoredItems) + (totalItems - i)
```
sseago, line 635: This line makes much more sense now that `i` is `processedItems`:
```
			actualTotalItems := len(ctx.restoredItems) + (totalItems - processedItems)
```

pkg/restore/restore.go
sseago: Renaming `i` to `processedItems` here.

pkg/restore/restore.go
sseago: Adding brief comment before the call, and longer comment at the func definition.

sseago:
@zubron I've incorporated your feedback comments in this update.

@dsu-igeek I've added a test case for this to enable_api_group_versions_test.go -- this may not be the best long-term place for this test (we're not testing gvr/gvkk issues here), but other than that, it was by far the easiest place to put it, since that test already did this in a loop with different data:
1) create a CRD and CRs in the cluster
2) back up the namespace(s) containing the CRs
3) delete the CRD and namespaces
4) create the target CRD in the cluster (this is the only step I don't need)
5) restore
6) make sure the CRs are there and are of the appropriate version

So for my test, I basically need everything in that list except for 4), and for 6) the version won't be changing. I added a new entry to the `tests` slice, and I modified the test to properly cope with a nil target CRD (so we don't pre-create the CRD before restoring).

With this new test, the velero/velero:main image fails as expected due to the regression. The image I built from my PR checkout passes, as the CR is properly restored.

zubron (approved):
Thanks for making the changes and adding the test, @sseago!

zubron (approved):
Thanks, @sseago!

dsu-igeek (approved):
Looks good, thanks!

PR #3855, opened 2021-06-08
Add plugin versioning design doc

### LINKED ISSUE #3710: Create design doc for plugin versioning
Create a design for plugin versioning as part of #3285.

zubron:

Fixes #3710 

### DIFF
--- a/design/plugin-versioning.md
+++ b/design/plugin-versioning.md
@@ -1,18 +1,19 @@
 # Plugin Versioning
 
 ## Abstract
-This proposal outlines an approach to support versioning of Velero's plugin APIs to enable changes to those APIs to be made that are backwards compatible through the addition of new plugin methods.
-It will not support backwards incompatible changes such as method removal or method signature changes.
+This proposal outlines an approach to support versioning of Velero's plugin APIs to enable changes to those APIs.
+It will allow for backwards compatible changes to be made, such as the addition of new plugin methods, but also backwards incompatible changes such as method removal or method signature changes.
+
 
 ## Background
 When changes are made to Velero’s plugin APIs, there is no mechanism for the Velero server to communicate the version of the API that is supported, or for plugins to communicate what version they implement.
 This means that any modification to a plugin API is a backwards incompatible change as it requires all plugins which implement the API to update and implement the new method.
 
 There are several components involved to use plugins within Velero.
-From the perspective of the core Velero codebase, all plugin kinds (e.g. ObjectStore, BackupItemAction) are defined by a single API interface and all interactions with plugins are managed by a plugin manager which provides an implementation of the plugin API interface for Velero to use.
+From the perspective of the core Velero codebase, all plugin kinds (e.g. `ObjectStore`, `BackupItemAction`) are defined by a single API interface and all interactions with plugins are managed by a plugin manager which provides an implementation of the plugin API interface for Velero to use.
 
 Velero communicates with plugins via gRPC.
-The core Velero project provides a framework (using the go-plugin project) for plugin authors to use to implement their plugins which manages the creation of gRPC servers and clients.
+The core Velero project provides a framework (using the [go-plugin project](https://github.com/hashicorp/go-plugin)) for plugin authors to use to implement their plugins which manages the creation of gRPC servers and clients.
 Velero plugins import the Velero plugin library in order to use this framework.
 When a change is made to a plugin API, it needs to be made to the Go interface used by the Velero codebase, and also to the rpc service definition which is compiled to form part of the framework.
 As each plugin kind is defined by a single interface, when a plugin imports the latest version of the Velero framework, it will need to implement the new APIs in order to build and run successfully.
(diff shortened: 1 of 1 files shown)


zubron:
@phuongatemc I can't seem to add you as a reviewer, but I'd like your input on this - thanks!

design/plugin-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+Velero plugins import the Velero plugin library in order to use this framework.
+When a change is made to a plugin API, it needs to be made to the Go interface used by the Velero codebase, and also to the rpc service definition which is compiled to form part of the framework.
+As each plugin kind is defined by a single interface, when a plugin imports the latest version of the Velero framework, it will need to implement the new APIs in order to build and run successfully.
+If a plugin does not use the latest version of the framework, and is used with a newer version of Velero that expects the plugin to implement those methods, this will result in a runtime error as the plugin is incompatible.
+
+With this proposal, we aim to break this coupling and introduce plugin API versions.
+
+## Scenarios to Support
+The following describes interactions between Velero and its plugins that will be supported with the implementation of this proposal.
+For the purposes of this list, we will refer to existing Velero and plugin versions as `v1` and Velero and plugin versions that use and implement new APIs as “v2`.
```
phuongatemc, line 25: Please correct quotation "v2`

design/plugin-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+
+Velero client communicating with plugins:
+
+- v2 Velero client will be able to communicate with v2 plugin
```
phuongatemc, line 29: Can you make the version more generic?  For example, instead of "v2 ... with v2 plugin", make it more generic like "version n Velero client will be able to communicate with all plugin versions from n, n-1, n-2... to v1".

zubron: Thanks for the suggestion! I've made some changes. Hopefully it is clearer now.

design/plugin-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+        } else {
+            return nil, err
+        }
+    }
+    return newRestartableObjectStoreV2(name, restartableProcess), nil
+}
+```
+
+If the previous version is not available, or can not be adapted to the latest version, an error will be returned as is currently the case when a plugin implementation for a particular kind and provider can not be found.
+
```
sseago, line 159: Once we have more than 2 versions (i.e. we're on v3, v4, etc.) the above might become somewhat unwieldy with the nesting of conditionals. It might make sense to refactor this into a method that takes a slice of versions/apis/etc. to check. At each entry, behavior would either be to return the found plugin, go to the next iteration, or return an error, depending on whether the current call returned without error, returned a `pluginNotFoundError`, or returned some other error.

design/plugin-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+As new versions of plugins are added, new restartable implementations of plugins will also need to be created.
+These are currently located within "pkg/plugin/clientmgmt" but will be rearranged to be grouped by kind and version like other plugin files.
+
+## Versioning Considerations
+
+It should be noted that if changes are being made to a plugin's API, it will only be necessary to bump the version once within a release cycle, regardless of how many changes are made within that cycle.
+This is because the changes will only impact consumers when they upgrade the versions of the Velero library that they import.
+Once a new version of Velero has been released however, any further changes will need to follow the process above and use a new version.
+
```
sseago, line 173: Would this include a commitment to not introduce plugin version changes in z-stream bugfix releases? i.e. if a particular plugin released v2 in 1.7.0, then we would not release a v3 plugin in 1.7.1 -- that would have to wait for 1.8.0?

sseago (commented):
Overall, this looks good -- it reflects prior discussions reasonably well. I had a few minor questions/suggestions, though.

design/plugin-versioning.md
zubron: Yes, I think so - I can't think of a situation at the moment that would warrant a new plugin version being introduced in a bugfix release. I'll add some more details in this section to make that clearer.

phuongatemc (approved):
The design is good.  Some details can leave to the implementation phase instead of trying to resolve in the design.

design/plugin-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+Plugins can opt to implement any or all versions of an API, however Velero will always attempt to use the latest version, and the plugin management will be modified to adapt earlier versions of a plugin to be compatible with the latest API.
+Under the existing plugin framework, any new plugin version will be treated as a new plugin with a new kind, however the plugin manager (which provides implementations of a plugin to Velero) will include an adapter layer which will manage the different versions and adapt any version which does not implement the latest version of the plugin API.
+
+For now, there will be a restriction that only new methods can be added to a plugin API.
+Methods can not be removed or modified.
+This restriction will be in place until Velero v2 when backwards incompatible changes may be included.
+
+Although adding new rpc methods to a service is considered a backwards compatible change within gRPC, due to the way the proto definitions are compiled and included in the framework used by plugins, this will require every plugin to implement the new methods.
+Instead, we are opting to treat _any_ API change as one requiring versioning.
+
```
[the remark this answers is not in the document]
zubron: Updated on lines 72-73.

design/plugin-versioning.md
zubron: @phuongatemc Yep, it is very verbose. Unfortunately the plugins require a lot of code to be written and every new version is essentially a new plugin kind :( I think this would be a great thing to improve and refactor in future but I don't think we can include it in the scope of this work.

@sseago - I think the changes to the code block above fit the idea you suggested where we iterate over a slice of plugin kinds (sorted by version in descending order). Let me know if you have other ideas or suggestions though :+1:

design/plugin-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+## Scenarios to Support
+The following describes interactions between Velero and its plugins that will be supported with the implementation of this proposal.
+For the purposes of this list, we will refer to existing Velero and plugin versions as `v1` and all following versions as version `n`.
+
+Velero client communicating with plugins:
+
+- Version `n` Velero client will be able to communicate with Version `n` plugin
+- Version `n` Velero client will be able to communicate with all previous versions of the plugin (Version `n-1` back to `v1`)
+- Optional: `v1` Velero client will be able to communicate with Version `n` plugin
```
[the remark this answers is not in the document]
zubron: I've updated this. Let me know if you think it needs further improvements.

design/plugin-versioning.md
```diff
@@ -123,35 +126,62 @@ When the plugin manager is requested to provide a particular plugin, it checks t
...
+		{
+			kind: framework.PluginKindObjectStoreV2,
+			Get: func(name string, restartableProcess RestartableProcess) v2.ObjectStore {
+				return newRestartableObjectStoreV2(name, restartableProcess), nil
+			},
+		},
+		{
+			kind: framework.PluginKindObjectStore,
+			Get: func(name string, restartableProcess RestartableProcess) v2.ObjectStore {
+				r := newRestartableObjectStore(name, restartableProcess)
```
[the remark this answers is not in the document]
zubron: I left the separate call to `newRestartableObjectStore` here to reflect the existing code that is in place. `newRestartableObjectStore` currently returns an object that satisfies the v1 interface so I wanted to make it clear that the `newAdaptedV1ObjectStore` function would take a v1 object and wrap it. Hope that makes sense, but happy to change it further if you think that would be better :+1:

phuongatemc (dismissed):
Very good and throughout design that allows graceful backward compatibility.

design/plugin-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+For the purposes of this list, we will refer to existing Velero and plugin versions as `v1` and all following versions as version `n`.
+
+Velero client communicating with plugins:
+
+- Version `n` Velero client will be able to communicate with Version `n` plugin
+- Version `n` Velero client will be able to communicate with all previous versions of the plugin (Version `n-1` back to `v1`)
+- Optional: Previous versions of the Velero client (from version `v1` to version `n-1`) will be able to communicate with Version `n` plugin
+    - This may work but is not a requirement
+
+Plugin client communicating with other plugins:
```
reasonerjt, line 34: How do we differentiate "velero client" and "plugin client"?

design/plugin-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+
+	return nil, fmt.Errorf("unable to get valid ObjectStore for %q", name)
+}
+```
+
+If the previous version is not available, or can not be adapted to the latest version, an error will be returned as is currently the case when a plugin implementation for a particular kind and provider can not be found.
+
+There are situations where it may be beneficial to check at the point where a plugin API call is made whether it implements a specific version of the API.
+This is something that can be addressed with future amendments to this design, however it does not seem to be necessary at this time.
+
```
reasonerjt, line 191: Could you also mention how the client may use the plugin?
For example when there are different versions of object store plugin, how the client adapts to different versions?

The client only sees the interface for v2.ObjectStore but it also has to handle v1?

design/plugin-versioning.md
```diff
@@ -0,0 +1,232 @@
+# Plugin Versioning
+
+## Abstract
+This proposal outlines an approach to support versioning of Velero's plugin APIs to enable changes to those APIs to be made that are backwards compatible through the addition of new plugin methods.
+It will not support backwards incompatible changes such as method removal or method signature changes.
```
reasonerjt, line 5: Per discussion, we SHOULD allow break changes in new version.

design/plugin-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+### Plugin Interface Definitions
+
+To provide versioned plugins, any new method added to a plugin interface will require a new versioned interface to be created which incorporates previous versions of the interface.
+Currently, all plugin interface definitions reside in `pkg/plugin/velero` in a file corresponding to their plugin kind.
+These files will be rearranged to be grouped by kind and then versioned: `pkg/plugin/velero/<plugin_kind>/<version>/`.
+
+For example, if a new method belongs to the ObjectStore API, a file would be added to `pkg/plugin/velero/objectstore/v2/` and would contain the following new API definition:
+
+```
+import "github.com/vmware-tanzu/velero/pkg/plugin/velero/objectstore/v1""
```
reasonerjt, line 88: If we are to allow break changes the new interface should not need to shadow the v1 interfaces.

reasonerjt:
Generally, we should allow break change in new versions.

reasonerjt (changes requested):
Please update the design so it allows breaking changes.

design/plugin-versioning.md
zubron: There isn't really a difference. I'm not aware of any plugins that currently call through to other plugins, but it was suggested that it could be a possibility in the future so I included it explicitly here. 

The resulting scenarios are the same though whether it's Velero or another plugin so I've just merged them.

design/plugin-versioning.md
```diff
@@ -0,0 +1,289 @@
...
+Currently, the plugin manager stores a [`map[string]RestartableProcess`](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/manager.go#L69), where the string key is the binary name for the plugin process (e.g. "velero-plugin-for-aws").
+Each `RestartableProcess` contains a [`map[kindAndName]interface{}`](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/restartable_process.go#L60) which represents each of the unique plugin implementations provided by that binary.
+[`kindAndName`](https://github.com/vmware-tanzu/velero/blob/main/pkg/plugin/clientmgmt/registry.go#L42) is a struct which combines the plugin kind (`ObjectStore`, `VolumeSnapshotter`) and the plugin name ("velero.io/aws", "velero.io/azure").
+
+Each plugin version registration must be unique (to allow for multiple versions to be implemented within the same plugin binary).
+This will be achieved by adding a specific registration method for each version to the Server interface in the plugin framework.
+For example, if adding a V2 `RestoreItemAction` plugin, the Server interface would be modified to add the `RegisterRestoreItemActionV2` method.
```
reasonerjt, line 153: But there may be issue if the plugin framework decided to drop `RegisterRestoreItemAction` and support only `RegisterRestoreItemActionV2` and the external plugin compiles with an old velero module and tries to call `RegisterRestoreItemAction`  for registration?

reasonerjt (approved):
Generally looks good I'll merge it to a separate branch and continue working on it when this is required by other users.

PR #3888, opened 2021-06-20
e2e test suite improvements

carlisia:
There were no actual changes to test functionalities.

This PR has many improvements for the e2e suite of tests.

✅ Passes on Kind*
✅ Passes on AWS

*The tests fail on my local Kind intermittently. This is likely due to a known issue with systemd.

- Improved documentation.

  I consider the documentation now comprehensive and complete. A user should be able to independently run all tests or run tests selectively, and add new tests following best practices, by reading this new documentation.

  Also, the changes in this PR will make way better sense after reading the new documentation: [velero/test/e2e/README.md](https://github.com/carlisia/velero/blob/b6393b3e702ce99984623243b8c2500118bfaa50/test/e2e/README.md)

- Addressed race conditions

  Originally, all tests were using the value of a global (configurable) variable to tell Velero where to install itself. With Velero being installed in the same namespace for every test, this cause a race coondition when running multiple or all tests at the same time. This was leading to one, test failing when they shouldn't and, two, a really messy and hard to grasp log.
  
  This global variable was removed. Instead, each test sets its own custom namespace and passes that in. I'm sure the string generation there can be improved, but as is this has been working well.
  
  Because Velero now **must** be installed in a namespace that corresponds to each test spec, the flag to decide to `installVelero` was also removed. The tests now require installation with each run.
  
  To this extent, created a new type `veleronamespace` to differentiate (and sort of enforce) namespaces where to install Velero resources from namespaces where to install other resources.
  
- User friendly configuration
  
  Added a `GINKGO_SKIP` flag to skip any test. 
  Added a  flag`GINKGO_SKIP_LONG ?= "2500 namespaces"` that configures the test to skip the long running tests that adds lots of namespaces by default. This can be overwritten and is documented. If it's not a good idea I can remove it.
  
- Validations

   Added a validation for humans to be warned when they run the tests on a non-Kind cluster and did not configure the `VSL_CONFIG` flag.
   
   Restructed the specs for testing single and multiple credential configuration and dded a validation to check if all the configs for creating the additional BSL are present. If not, it'll skip the spec at a point where it won't load the test setup.

- Better cleanup

	Per this new documentation, a "e2e" label should be added to all resources created. When creating a new namespace, for example, the label gets added and  the function that deletes namespaces that have a given label can be called from anywhere, including in the tear down blocks. This also helps a lot when having to delete resources manually. Try stopping the test that creates the 2,5000 namespaces in the middle and let me know if this isn't useful!

	On `veleroUninstall`, now there is always deletion of Velero objects (backup/restore/bsl) that exist in the namespace where the test installed Velero. It doesn't select objecs, it just deletes all of them. When new tests create other objects, probably good to add a deletion for that.

- General refactoring
  
  Code to clean up/delete namespaces was removed from the `multiple namespaces` test and into a helper function in `common.go`. This function `deleteNamespaceListWithLabel` is called from that test, as well as in other tests that needs the same action. Because it relies on a label to find the namespaces to delete, it doesn't need to know the actual names of the namespaces. 
  
  A context.Context variable is now instantiated together with the client, and can easily be used to be passed in into the tests.
  
  Consolidated the `multiple namespaces` test specs to eliminate unnecessary duplication.
  
  Error messages should start with lower caps, cleaned up some of those  that were around code I touched.
  
- Bug
 
	Fixed a bug where one of the test spec in the `multiple namespaces` test was not setting an expectation so backup/restore failures were allowing the test to pass anyway.


carlisia:
I made a couple more improvements. I don't predict I'll be adding any more to it, ready for review.

dymurray:
I don't mind all of the improvements to the tests especially to support fixes for race conditions and whatnot, but I feel strongly that we must leave support for running the e2e test suite against an already installed Velero.

Not only do I believe this is a more "proper" pattern for e2e tests, but this would also break my team's requirement of running e2e tests against a previously installed Velero from one of our components and *not* the install function the test cases use.

carlisia:
@dymurray I'm very glad you don't mind the huge amount of work that went into making these e2e tests more robust, easier to follow, with more flexible options for running, and way better documented.

I won't have the bandwidth to add that option back in this PR.

dymurray:
@carlisia I re-read my prior comment and just wanted to apologize, I see how what I wrote sounded unappreciative. I'm sorry, that is certainly not the case. I appreciate all of the time you spent and the improvements to the test suite are very much desired. At the last community call I expressed interest in myself and a couple others helping push the E2E tests forward with more coverage and part of that discussion included a desire to preserve and fix the `-install-velero` flag.

My comment was poorly worded just wanted to call out that some of the race condition fixes will have impacts on running this against a previously installed Velero as I mentioned on the call. 

I would be more than happy to help restore the `-install-velero` functionality in a separate PR if this is merged as is, or would be happy to help update this PR to restore that functionality if at all possible.

carlisia:
@dymurray thanks for offering to add option that back and, if you don't mind, I'd prefer to have any additional changes to be added back in a separate PR.  I'm really looking fwd to merging this, bc there are changes in all files and as it is I'm fighting git to rebase the conflict in the one file that was merged after. The longer this stays open the more likely I'll have to handle more merge conflicts.

carlisia:
@vmware-tanzu/velero-owners rebased, this is ready for review.

PR #3877, opened 2021-06-16
Skip volume restores from projected sources

zubron:
# Please add a summary of your change
In #3863, it was discovered that volumes from projected sources were
being backed up by restic when they should have been skipped. Restoring
these volumes triggers a known bug in restic.

In #3866, we started skipping volumes from a projected source, however
there will exist backups that were taken before this fix was introduced.
This change modifies the restore logic to skip the restore of any volume
that came from a projected source, allowing backups taken before #3866
to be restored successfully.

Signed-off-by: Bridget McErlean <bmcerlean@vmware.com>

# Does your change fix a particular issue?

Fixes #3870 

# Please indicate you've done the following:

- [x] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [x] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.
- [x] Updated the corresponding documentation in `site/content/docs/main`.

PR #3866, opened 2021-06-11
skip backuping projected volume when using restic

alaypatel07:
Signed-off-by: Alay Patel <alay1431@gmail.com>

Thank you for contributing to Velero!

# Please add a summary of your change

This skips adding projected volumes to PVB spec for restic backup

# Does your change fix a particular issue?

Fixes #3863

# Please indicate you've done the following:

- [x] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [x] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.
- [ ] Updated the corresponding documentation in `site/content/docs/main`.

PR #3764, opened 2021-05-01
Consolidate api clients for e2e tests

carlisia:
This PR:

- In the `utils.go` file, a function `GetClusterClient()` had been added to create a k8s api client in a new, alternative way from what we already have. This was used in two different places to instantiate api clients: in the 2e2 tests, and in the `uninstall` cli. It makes sense to use that client in the uninstall, since the e2e tests were also calling uninstall (in the `velero_utils.go`) and having to pass in that instantiated client.

- Because I didn't see any additional benefit to having this alternate way to instantiate clients, I replaced it with the kubebuilder/controller-runtime client libraries, which is what we should be using for all new code that has to use a k8s API client.

- Because some of the 2e2 code was using the existing generated API client code (see `common.go` and `enable_api_group_versions_test.go` as example), I left those alone.

- I consolidated the instantiation of the clients under a single function/struct (in the `client.go` file), with documentation, which I think should make it easier to discover and reuse the appropriate API clients throughout the package.

What this PR does not do:
- Does not replace the existing code that calls the generated API client with kubebuilder/controller-runtime.
- Does not refactor the code to invoke the install/uninstall cmds instead of rewriting (some) of the code to do that.

Note: if this is a desirable direction, I can also remove the `client-go` API client and replace that with controler--runtime, either in this PR or in a new one.

Signed-off-by: Carlisia <carlisia@grokkingtech.io>

Tests:
- [x] uninstall
- [x] e2e tests on Kind
- [x] e2e tests on AWS

# Does your change fix a particular issue?

Fixes #3625

# Please indicate you've done the following:

- [x] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [x] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.

PR #3726, opened 2021-04-23
A small refactor of the e2e tests

carlisia:
Signed-off-by: Carlisia <carlisia@grokkingtech.io>

A small refactor to make functions private since they are not at the moment needed to be exported. We can export functions as we start moving things into separate packages.

Fixes #3724.

# Please indicate you've done the following:

- [x] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [x] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.
- [ ] Updated the corresponding documentation in `site/content/docs/main`.

PR #3889, opened 2021-06-21
Add cherry-pick commits and changelog for v1.6.1

zubron:
The 1.6.1 release includes the following bug fixes:
- #3711
- #3866 
- #3877
- #3845

The following PRs (which include improvements to the E2E tests) have also been included as they are required for some of the above bug-fix PRs:
- #3726
- #3764
- #3564
Including them now will also remove the risk of future merge conflicts on this branch.


zubron:
Hmm, so the DCO is not passing. I think this is because the sign-off on the original commits does not match the associated author of the commits. I think this happens as the email address associated with the author's GitHub account is different than the sign off email. This can also happen when people enable email privacy. In this case the author's email address will be "username@users.noreply.github.com". I'm inclined to set the DCO to pass given that all of the commits have been signed. Any objections?

sseago (approved):

sseago:
@zubron I'm fine with setting the DCO to pass, but I do wonder why this didn't trigger DCO failure for the original PR commits. Wouldn't the same email address discrepancy between github and signoff email have happened then?

zubron:
@sseago It looks like it is the behaviour of the "Squash and merge" option when merging PRs: https://github.community/t/github-squash-and-merge-how-is-the-author-email-determined/156376

The author email matched the sign off email in all of the original commits in the PR, but at the point when it was squashed and merged, the author email for the squashed commit is set to the default email address for that account, or something "@users.noreply.github.com" if email privacy is enabled.

jenting (approved):

PR #3928, opened 2021-07-06
Allow image registry to be configured at build time

### LINKED ISSUE #3831: Allow customizing Velero image at build time
Currently, [Velero hard-codes](https://github.com/vmware-tanzu/velero/blob/ccac3d41301c8883b62e2d3d3429711a5f8088d6/pkg/install/resources.go#L45) the image being pulled for cluster installation. This presents a challenge when trying to install from other repositories, despite the existence of the `--image` argument.

Allow for customizing the repository to be used at build time, defaulting to `velero/velero` (DockerHub)

**Environment:**

- Velero version (use `velero version`): All current versions

**Vote on this issue!**

zubron:
This adds a new `buildinfo` variable `ImageRegistry` that can set at
build time like the `Version` variable. This allows us to customise the
Velero binary to use different registries.

If the variable is set, this variable wille be used when creating the
URIs for both the main `velero` and `velero-restic-restore-helper`
images. If it is not set, default to using Dockerhub (`velero/velero`,
`velero/velero-restic-restore-helper`).

There are numerous ways in which the Velero binary can be built so all
of them have been updated to add the new link time flag to set the
variable:
* `make local` (used for local developer builds to build for the local
  OS and ARCH)
* `make build` (used by developers and also VMware internal builds to
  build a specific OS and ARCH)
* Goreleaser config (used when creating OSS release binaries)
* Dockerfile (used to build the Velero binary used within the image)

All of these workflows are currently triggered from our Makefile where
the variable `REGISTRY` is already available with the default value of
`velero` and used to build the image tag. Where the new `ImageRegistry`
build variable is needed, we pass through this Makefile variable to
those tasks so it can be used accordingly.

The GitHub action and the `./hack/docker-push.sh` script used to push
container images has not been modified. This will continue to use the
default registry specified in the Makefile and will not explicitly pass
it in.

Fixes #3831


reasonerjt:
Should we also allow the namespace to be customized?  Is it possible there's a requirement that all images should live in $registry/tanzu-addon/xxxx URI?

hack/build.sh
```diff
@@ -41,6 +41,11 @@ if [[ -z "${VERSION}" ]]; then
     exit 1
 fi
 
+if [[ -z "${REGISTRY}" ]]; then
```
reasonerjt, line 44: So if the `REGISTRY` is a must-have var when triggering make, is there any documentation needs to be updated?
I quickly when through https://velero.io/docs/main/maintainers/#creating-a-release but didn't find any reference.
What should be set for OSS release that images live in dockerhub?

zubron: The Makefile has a default value for this variable of "velero" which will result in "velero/velero" being used as the image, but I will also add some extra docs to the release process :+1:

zubron:
> Should we also allow the namespace to be customized? Is it possible there's a requirement that all images should live in $registry/tanzu-addon/xxxx URI?

Hmm, this shows my lack of knowledge with regard to image URI formats. I was assuming that in this case, if someone wanted to build a Velero binary which could pull from `$registry/tanzu-addon/xxxx` then _that_ would be the full value that they would pass in at build time, producing an image URI "$registry/tanzu-addon/xxxx/velero", but it seems that is incorrect. Would URIPrefix or similar be a better way to name this?

hack/build.sh
zubron: I opted not to change this in the end as our GitHub action to push images calls a script directly which then invokes Make to perform the push to Dockerhub. Given that this is the default behaviour that we want for our OSS releases for now, there are no changes necessary to the maintainers documentation for performing a release. We had some [existing docs](https://velero.io/docs/v1.6/build-from-source/#publishing-container-images-to-a-registry) related to building and pushing images to your own registry and these still work as intended. I thought about expanding the docs there to include the binary build instructions, however I don't think this is something we want for most developers and we should encourage them to explicitly set the image if doing local builds.

dsu-igeek (approved):
Looks good, just think there's a couple of spots that should be info, if you agree change, otherwise it's fine.

pkg/restore/restic_restore_action.go
```diff
@@ -193,29 +192,31 @@ func getCommand(log logrus.FieldLogger, config *corev1.ConfigMap) []string {
...
 	log = log.WithField("image", image)
 
 	parts := strings.Split(image, "/")
 
 	if len(parts) == 1 {
+		defaultImage := veleroimage.DefaultResticRestoreHelperImage()
 		// Image supplied without registry part
-		log.Debugf("Plugin config contains image name without registry name. Return defaultImageBase")
-		return initContainerImage(defaultImageBase)
+		log.Debugf("Plugin config contains image name without registry name. Using default init container image: %q", defaultImage)
```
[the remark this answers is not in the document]
zubron: Good catch. This could be confusing otherwise.

hack/build.sh
reasonerjt: sure

pkg/restore/restic_restore_action.go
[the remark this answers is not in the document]
reasonerjt: Sorry　@zubron but this is still `Debugf`?

reasonerjt:
> > Should we also allow the namespace to be customized? Is it possible there's a requirement that all images should live in $registry/tanzu-addon/xxxx URI?
> 
> Hmm, this shows my lack of knowledge with regard to image URI formats. I was assuming that in this case, if someone wanted to build a Velero binary which could pull from `$registry/tanzu-addon/xxxx` then _that_ would be the full value that they would pass in at build time, producing an image URI "$registry/tanzu-addon/xxxx/velero", but it seems that is incorrect. Would URIPrefix or similar be a better way to name this?

I mean we may add introduce variable `ns` to enable kubelet to pull image via URI like `$registry/tanzu-addon/velero` during the installation.  And in that case, we just need to assign `tanzu-addon` to the `ns`. This also covers the case if we want the user to pull from `$registry/tanzu-addon/xxxxx/velero`, b/c we can assign `tanzu-addon/xxxxx` to `ns`. Hope that explains.

BTW I'm bad at naming things...

pkg/restore/restic_restore_action.go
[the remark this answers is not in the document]
zubron: @reasonerjt This is an issue with GitHub where it still shows the previous version where the original comment was made. I added commit https://github.com/vmware-tanzu/velero/pull/3928/commits/04345af17a13314765adce5df3531c733a175aea which updates lines [211](https://github.com/vmware-tanzu/velero/pull/3928/files#diff-9f110ab28d31cd60e15e9ec43ad813f22bd8b9329eaf0ebe70c0b6590592b76dR211) and  [218](https://github.com/vmware-tanzu/velero/pull/3928/files#diff-9f110ab28d31cd60e15e9ec43ad813f22bd8b9329eaf0ebe70c0b6590592b76dR218) in this file to be `Infof`.

zubron:
@reasonerjt Ah, I understand. I don't _think_ we need that level of granularity and control at this stage. I think if there is a need to push to an longer path including a namespace, we can pass the full string in via the `ImageRegistry` variable, for example `projects.registry.vmware.com/tanzu-addon/xxxxx`. This code will likely end up changing post 1.6.2 anyway once we start to explore moving away from Dockerhub so perhaps we can revisit it then?

reasonerjt (approved):

dsu-igeek (approved):
LGTM, thanks!

PR #3944, opened 2021-07-13
adding namespace annotation test to e2e tests

jaidevmane:

- added a test that verifies namespace annotations are backed up and restored correctly.

Fixes #(issue)

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -47,6 +47,26 @@ func createNamespace(ctx context.Context, client testClient, namespace string) e
 	return err
 }
 
+// createNamespaceAndAnnotation creates a kubernetes namespace and an annotation
+func createNamespaceAndAnnotation(ctx context.Context, client testClient, namespace string, annotation string) error {
+	ns := builder.ForNamespace(namespace).Result()
+	_, err := client.clientGo.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
+	nsAnno := ns.ObjectMeta.Annotations
+
+	if nsAnno == nil{
+		nsAnno = make(map[string]string)
+	}
+	nsAnno["testAnnotation"] = annotation
+	ns.ObjectMeta.Annotations = nsAnno
+	n, err := client.clientGo.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{})
+	fmt.Println(n)
+
+	if apierrors.IsAlreadyExists(err) {
+		return nil
+	}
+	return err
+}
+
 func getNamespace(ctx context.Context, client testClient, namespace string) (*corev1api.Namespace, error) {
 	return client.clientGo.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
 }
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -52,7 +52,7 @@ func createNamespaceAndAnnotation(ctx context.Context, client testClient, namesp
 	ns := builder.ForNamespace(namespace).Result()
 	_, err := client.clientGo.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
 
-	if err != nil && !apierrors.IsAlreadyExists(err){
+	if err != nil && !apierrors.IsAlreadyExists(err) {
 		return err
 	}
 
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -52,8 +52,11 @@ func createNamespaceAndAnnotation(ctx context.Context, client testClient, namesp
 	ns := builder.ForNamespace(namespace).Result()
 	_, err := client.clientGo.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
 
-	if apierrors.IsAlreadyExists(err) {
-		return nil
+	if err != nil {
+		if !(apierrors.IsAlreadyExists(err)) {
+			return err
+		}
+		return err
 	}
 
 	nsAnno := ns.ObjectMeta.Annotations
@@ -65,8 +68,8 @@ func createNamespaceAndAnnotation(ctx context.Context, client testClient, namesp
 	ns.ObjectMeta.Annotations = nsAnno
 	_, err = client.clientGo.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{})
 
-	if apierrors.IsAlreadyExists(err) {
-		return nil
+	if err != nil {
+		return err
 	}
 	return err
 }
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -52,7 +52,7 @@ func createNamespaceAndAnnotation(ctx context.Context, client testClient, namesp
 	ns := builder.ForNamespace(namespace).Result()
 	_, err := client.clientGo.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
 
-	if err != nil && !(apierrors.IsAlreadyExists(err)){
+	if err != nil && !apierrors.IsAlreadyExists(err){
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


test/e2e/common.go
```diff
@@ -47,6 +47,26 @@ func createNamespace(ctx context.Context, client testClient, namespace string) e
 	return err
 }
 
+// createNamespaceAndAnnotation creates a kubernetes namespace and an annotation
+func createNamespaceAndAnnotation(ctx context.Context, client testClient, namespace string, annotation string) error {
+	ns := builder.ForNamespace(namespace).Result()
+	_, err := client.clientGo.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
```
dymurray, line 53: Need to check this error and return it

test/e2e/common.go
```diff
@@ -47,6 +47,26 @@ func createNamespace(ctx context.Context, client testClient, namespace string) e
...
+	nsAnno := ns.ObjectMeta.Annotations
+
+	if nsAnno == nil{
+		nsAnno = make(map[string]string)
+	}
+	nsAnno["testAnnotation"] = annotation
+	ns.ObjectMeta.Annotations = nsAnno
+	n, err := client.clientGo.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{})
```
dymurray, line 61: Need error checking

test/e2e/common.go
```diff
@@ -47,6 +47,26 @@ func createNamespace(ctx context.Context, client testClient, namespace string) e
...
+	fmt.Println(n)
```
dymurray, line 62: Remove this

test/e2e/common.go
jaidevmane: Fixed it in a new commit. Thanks!

test/e2e/common.go
jaidevmane: Removed the print statement in a new commit. Thanks!

test/e2e/common.go
```diff
@@ -47,6 +47,30 @@ func createNamespace(ctx context.Context, client testClient, namespace string) e
...
 }
 
+// createNamespaceAndAnnotation creates a kubernetes namespace and an annotation
+func createNamespaceAndAnnotation(ctx context.Context, client testClient, namespace string, annotation string) error {
+	ns := builder.ForNamespace(namespace).Result()
+	_, err := client.clientGo.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
+
+	if apierrors.IsAlreadyExists(err) {
+		return nil
+	}
```
sseago, line 57: At this point, if `err != nil` but it's not an IsAlreadyExists error, then the error should be returned. If the namespace *does* already exist, then we should continue to add the annotation, otherwise the already-existing namespace won't get the annotation added.

test/e2e/common.go
```diff
@@ -47,6 +47,30 @@ func createNamespace(ctx context.Context, client testClient, namespace string) e
...
+		nsAnno = make(map[string]string)
+	}
+	nsAnno["testAnnotation"] = annotation
+	ns.ObjectMeta.Annotations = nsAnno
+	_, err = client.clientGo.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{})
+
+	if apierrors.IsAlreadyExists(err) {
+		return nil
+	}
+	return err
```
sseago, line 71: This is an Update call, so I wouldn't expect an IsAlreadyExists error. Just return err at this point.

test/e2e/namespace_annotation_test.go
```diff
@@ -0,0 +1,122 @@
...
+			fiveMinTimeout, _ := context.WithTimeout(context.Background(), 5*time.Minute)
+			RunNamespaceAnnotationTest(fiveMinTimeout, client, "nstest-"+uuidgen.String(), 2,
+				backupName, restoreName)
+		})
+	})
+})
+
+func RunNamespaceAnnotationTest(ctx context.Context, client testClient, nsBaseName string, numberOfNamespaces int, backupName string, restoreName string) error {
+	//timeout logic
+	shortTimeout, _ := context.WithTimeout(ctx, 5*time.Minute)
```
sseago, line 57: You already got the 5-minute timeout context before passing it into this method, so I don't think it's needed here.

test/e2e/namespace_annotation_test.go
```diff
@@ -0,0 +1,122 @@
...
+	// Verify that we got back all of the namespaces we created, (and also check if we got back the annotations)
+	for nsNum := 0; nsNum < numberOfNamespaces; nsNum++ {
+		checkNSName := fmt.Sprintf("%s-%00000d", nsBaseName, nsNum)
+		checkAnnoName := fmt.Sprintf("annotation-%s-%00000d", nsBaseName, nsNum)
+		checkNS, err := getNamespace(shortTimeout, client, checkNSName)
+		c := checkNS.ObjectMeta.Annotations["testAnnotation"]
+
+		if err != nil {
+			return errors.Wrapf(err, "Could not retrieve test namespace %s", checkNSName)
+		}
```
sseago, line 110: Check for error before trying to access annotations.

test/e2e/common.go
jaidevmane: Added updates in new commits

test/e2e/namespace_annotation_test.go
jaidevmane: added changes

test/e2e/common.go
jaidevmane: added changes in new commit

test/e2e/common.go
sseago: Style nit: the extra parentheses around `apierrors.IsAlreadyExists(err)` are not needed. Line 55 is more readable as:
```
	if err != nil && !apierrors.IsAlreadyExists(err) {
```

sseago (approved):
With the one remaining style change suggested (removing the extra set of parens), this looks good to me.

Longer-term, we may need to put some more thought into how to organize the e2e tests. Just adding new ones randomly may eventually get unwieldy. But right now we just need more tests added. Maybe it makes sense to just add this now and worry about that later, unless someone else on the team has suggestions for a better way to do that in the context of this PR.

dymurray:
@sseago I agree, it does feel that we need to come up with a pattern that can scale to allow folks to add more test cases easily without just adding random ones in new files. I don't have a good solution for this just yet... but something I'd like to consider refactoring as we move forward.

sseago (dismissed):

dsu-igeek (changes requested):
Does this need to be a separate test case or can we just add the annotations and check to the existing backup test?

sseago:
@dsu-igeek Related to your question about new vs. modified tests, we should really get some sort of "e2e test plan" guidelines in place. This test probably does make sense as a modification to an existing one, but it's not always clear which is better. Adding a bunch of random disorganized test cases would be a bad thing, but putting anything and everything into the existing 5 or so test cases might be just as bad. I'm not really sure what the intent is in terms of how to grow these tests over time in a way that will be maintainable and easy enough to understand what to run when, how to add new tests, etc.

dsu-igeek:
> @dsu-igeek Related to your question about new vs. modified tests, we should really get some sort of "e2e test plan" guidelines in place. This test probably does make sense as a modification to an existing one, but it's not always clear which is better. Adding a bunch of random disorganized test cases would be a bad thing, but putting anything and everything into the existing 5 or so test cases might be just as bad. I'm not really sure what the intent is in terms of how to grow these tests over time in a way that will be maintainable and easy enough to understand what to run when, how to add new tests, etc.

@sseago Agreed, I added an issue to document this - #4088

dsu-igeek (changes requested):
We'd like to have this changed to be an extension to the existing E2E test rather than a separate test.

Thanks!

reasonerjt:
> @dsu-igeek Related to your question about new vs. modified tests, we should really get some sort of "e2e test plan" guidelines in place. This test probably does make sense as a modification to an existing one, but it's not always clear which is better. Adding a bunch of random disorganized test cases would be a bad thing, but putting anything and everything into the existing 5 or so test cases might be just as bad. I'm not really sure what the intent is in terms of how to grow these tests over time in a way that will be maintainable and easy enough to understand what to run when, how to add new tests, etc.

Agreed, we should discuss it in the community meeting.

ywk253100:
I suggest we move the [Basic](https://github.com/vmware-tanzu/velero/blob/main/test/e2e/multiple_namespaces_test.go#L19) test case into a separated file and don't share the same logic with the `Scale` test case anymore, in this way we can expand the `Basic` case to include more resources for backup/restore testing without affecting the `Scale` case

qiuming-best:
I'll keep going on the work on #4455, so I closed

Issue #675, opened 2018-07-13
FE: Create a `velero debug` command for gathering troubleshooting information


nrb:
**Describe the solution you'd like**
Provide an `ark debug` subcommand that could output the following information:

* ark client and server versions
* ark pod logs (should probably include restic logs, too)
* ark config
* If a backup/restore name is provided:
** relevant backup or restore logs
** backup and/or restore YAML

Additionally, the `ark debug` command should provide a way to filter out sensitive information like:

* bucket names
* secrets
* More?

This command would make it easier for users to file bug reports and get answers in a timely manner.

rosskukulinski:
Great suggestion @nrb.  We could also pair this command with https://github.com/heptio/ark/issues/578

nrb:
Oh nice, I had missed that issue.

I definitely think linking the two would be a good idea.

metadave:
I was thinking that `ark debug` and `ark bug` could be two separate things (or a set of flags on one command):

- As an open source user, having a convenience command like `ark bug` to populate some info in a Github issue would be nice, but I wouldn't want to expose my config, logs, or sensitive data without me manually editing/copying/pasting into the issue.

- As a customer paying for support, I could see `ark debug` generating a tarball of info that could be attached to a private helpdesk ticket. It's fine to try and scrub sensitive data, but you can't guarantee that the code won't miss a secret or two. We used a command like this to support Riak back in the day, the support team loved it.

Perhaps `ark bug` could include a `--verbose` and/or `--tarball` flag(s) to generate the more complete debug file.

ncdc:
I agree they should be 2 separate things. `ark bug` must never expose confidential data and is really a convenience for getting you to a GitHub issue, possibly with some information filled in. `ark debug` may or may not contain confidential data - we should have a sufficient number of flags to allow you to control what is included in the debug tarball.

carlisia:
Wanted to drop the output of Tilt's `tilt doctor` cmd:

```
❯ tilt doctor
Tilt: v0.17.12, built 2020-11-19
System: darwin-amd64
---
Docker
- Host: [default]
- Version: 1.40
- Builder: 2
---
Kubernetes
- Env: kind-0.6+
- Context: kind-development
- Cluster Name: kind-development
- Namespace: velero
- Container Runtime: containerd
- Version: v1.18.2
- Cluster Local Registry: none
---
Thanks for seeing the Tilt Doctor!
Please send the info above when filing bug reports. 💗

[...]

PR #4022, opened 2021-08-09
Implement `velero debug`

reasonerjt:

This PR added a subcommand `velero debug`, which leverages `crashd` to
collect logs and specs of velero server components and bundle them in a
tarball.

After this repo is pinned to v0.3.4 of `crashd` the k8s libraries were updated and caused failures in test cases.
This PR fixed the failures and I opened #4059 to do further code cleanup.

Fixes https://github.com/vmware-tanzu/velero/issues/675


pkg/cmd/cli/debug/debug.go
```diff
@@ -0,0 +1,140 @@
...
+	// the absolute path for the kubeconfig file that will be read by crashd for calling K8S API
+	kubeconfigPath string
+	// optional, the name of the backup resource whose log will be packaged into the debug bundle
+	backup string
+	// optional, the name of the restore resource whose log will be packaged into the debug bundle
+	restore string
+}
+
+func (o *option) bindFlags(flags *pflag.FlagSet) {
+	flags.StringVar(&o.outputPath, "output", "", "The path of the bundle tarball, by default it's $HOME/bundle.tar.gz. Optional")
```
dsu-igeek, line 54: When I ran it, it output into my current directory, not $HOME.  I think that's actually better behavior, so let's change the document here.

pkg/cmd/cli/debug/debug.go
```diff
@@ -0,0 +1,140 @@
+/*
+Copyright 2017 the Velero contributors.
```
dsu-igeek, line 2: Remove the year

pkg/cmd/cli/debug/debug.go
```diff
@@ -0,0 +1,140 @@
...
+		if err := os.Remove(f.Name()); err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to remove the temp file: %s, %v\n", f.Name(), err)
+		}
+	}()
+	_, err2 := f.Write(scriptBytes)
+	if err2 != nil {
+		return err2
+	}
+	os.Args = []string{"", "run", "--debug", f.Name(), "--args", fmt.Sprintf("%s", argString)}
+	return crashdCmd.Run()
```
dsu-igeek, line 130: Since we want customers to run this and they're probably just going to run "velero debug", let's output the file that was written and some instructions on what to do with it.  Let's discuss where we should tell them to put it for support requests, can't think of the right answer at the moment

pkg/cmd/cli/debug/debug.go
dsu-igeek: Let's make the default name generated more descriptive and unique.  I'm thinking `"velero-debug-<YYYY>-<MM>-<DD>-<HH>-<MM>-<SS>.tar.gz"`  Or we could use the crashd # and that would be consistent.

dsu-igeek: Also, when I untar it, it goes into tmp/crashdXXXX.  Can we not put the tmp part of the path into the tar file?

dsu-igeek (changes requested):
Looking good, I ran it and it works, that's pretty cool!  Some changes that would be nice.  Would you add an issue for an E2E test for this?  We should be able to verify that debug is getting the expected logs.  Also, let's add an issue for documentation, we should probably update site/content/docs/main/troubleshooting.md

pkg/cmd/cli/debug/cshd-scripts/velero.cshd
```diff
@@ -0,0 +1,25 @@
...
+basedir = args.basedir if args.basedir else os.home
+output = args.output if args.output else "bundle.tar.gz"
+# Working dir for writing during script execution
+crshd = crashd_config(workdir="{0}/velero-bundle".format(basedir))
+set_defaults(kube_config(path=args.kubeconfig))
+capture_local(cmd="velero version -n {}".format(ns), file_name="version.txt")
+capture_backup_logs(ns)
+capture_restore_logs(ns)
+kube_capture(what="logs", namespaces=[ns])
+kube_capture(what="objects", namespaces=[ns], groups=['velero.io'], kinds=['backupstoragelocation', 'podvolumebackup', 'podvolumerestore'])
```
dsu-igeek, line 22: I think we should also grab:
resticrepositories
schedules
volumesnapshotlocations
downloadrequests
deletebackuprequests

We're grabbing podvolumebackup and podvolumerestore here but we only capture the backup/restore resources if the backup or restore flag is selected.  Resources are pretty small, can we just capture all of the velero.io resources here regardless of whether they've selected a particular backup or restore?  Backup/restor log files are big, and we need to retrieve from object storage so it sounds right to only include those if people specify a particular backup/restore.

dsu-igeek (changes requested):
Listed some more resources we should capture

pkg/cmd/cli/debug/debug.go
```diff
@@ -0,0 +1,140 @@
...
+		Run: func(c *cobra.Command, args []string) {
+			defer func(opt *option) {
+				if len(o.baseDir) > 0 {
+					if err := os.RemoveAll(o.baseDir); err != nil {
+						fmt.Fprintf(os.Stderr, "Failed to remove temp dir: %s: %v\n", o.baseDir, err)
+					}
+				}
+			}(o)
+			flags := c.Flags()
+			err := o.complete(f, flags)
```
zubron, line 103: Can we do some validation of the `backup` and `restore` arguments here before they are passed to crashd usin something like this: https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/cli/backup/get.go#L53-L54)?

I tried this out locally with a backup that didn't exist and the script failed when calling `backup logs`. Unfortunately crashd didn't give the stderr to say that the backup didn't exist. I tried looking at the crashd functions that we can call, and it doesn't look like it supports providing the stderr. I think it would be easier to just validate the existence of those objects here before calling crashd.

pkg/cmd/cli/debug/debug.go
```diff
@@ -0,0 +1,140 @@
...
+	defer func() {
+		if err := os.Remove(f.Name()); err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to remove the temp file: %s, %v\n", f.Name(), err)
+		}
+	}()
+	_, err2 := f.Write(scriptBytes)
+	if err2 != nil {
+		return err2
+	}
+	os.Args = []string{"", "run", "--debug", f.Name(), "--args", fmt.Sprintf("%s", argString)}
```
zubron, line 129: Can we remove the use `--debug` here as it makes the output very verbose, or make it configurable?

pkg/cmd/cli/debug/debug.go
zubron, line 130: Did you consider using the [`ExecuteFile`](https://github.com/vmware-tanzu/crash-diagnostics/blob/main/exec/executor.go#L36-L38) function here? That way we might be able to avoid writing to `os.Args` and can pass in a map of arguments.

zubron (changes requested):
Thanks, @reasonerjt! This is looking great so far and I am really excited for this to be available :smile: I just have a few comments and suggestions on top of what @dsu-igeek has already suggested.

pkg/cmd/cli/debug/debug.go
reasonerjt: Fixed according to the 3 comments.

pkg/cmd/cli/debug/debug.go
reasonerjt: fixed

pkg/cmd/cli/debug/debug.go
reasonerjt: I made the update to add some validation against the kubeconfig, kubecontext, namespace and the backup and restore.

pkg/cmd/cli/debug/debug.go
reasonerjt: Added a new flag `verbose` to `debug` subcommand to make it configurable and default to `false`

pkg/cmd/cli/debug/cshd-scripts/velero.cshd
reasonerjt: Updated the crashd script to collect all resources in group `velero.io`

pkg/cmd/cli/debug/debug.go
reasonerjt: Thanks for pointing it out, I updated the code to use `exec.Execute` which consumes an `io.Reader` so there's no need to write the script to a temp file.

pkg/test/api_server.go
```diff
@@ -42,9 +43,21 @@ func NewAPIServer(t *testing.T) *APIServer {
 	t.Helper()
 
 	var (
-		veleroClient    = fake.NewSimpleClientset()
-		kubeClient      = kubefake.NewSimpleClientset()
-		dynamicClient   = dynamicfake.NewSimpleDynamicClient(runtime.NewScheme())
+		veleroClient  = fake.NewSimpleClientset()
+		kubeClient    = kubefake.NewSimpleClientset()
+		dynamicClient = dynamicfake.NewSimpleDynamicClientWithCustomListKinds(runtime.NewScheme(),
+			map[schema.GroupVersionResource]string{
```
zubron, line 49: Is this update due to the library upgrade as well? I don't think this is related, but we have some other places where we add to the scheme, do you know if those are affected by the upgrade as well? https://github.com/vmware-tanzu/velero/blob/main/pkg/client/factory.go#L156-L160

pkg/cmd/cli/debug/debug.go
```diff
@@ -0,0 +1,192 @@
...
+	restore string
+	// optional, it controls whether to print the debug log messages when calling crashd
+	verbose bool
+}
+
+func (o *option) bindFlags(flags *pflag.FlagSet) {
+	flags.StringVar(&o.outputPath, "output", "", "The path of the bundle tarball, by default it's $HOME/bundle.tar.gz. Optional")
+	flags.StringVar(&o.backup, "backup", "", "The name of the backup resource whose log will be collected, no backup logs will be collected if it's not set. Optional")
+	flags.StringVar(&o.restore, "restore", "", "The name of the restore resource whose log will be collected, no restore logs will be collected if it's not set. Optional")
+	flags.BoolVar(&o.verbose, "verbose", false, "When it's set to true the debug messages by crashd will be printed during execution.  Default value is false.")
```
zubron, line 65: I don't see this being passed in to crashd, and looking at the crashd library it looks like the debug level is set as part of the cobra.Command setup: https://github.com/vmware-tanzu/crash-diagnostics/blob/main/cmd/crashd.go#L53-L57

Is it possible to override the logrus level from within the Velero code? I don't even know if that is recommended! :confused:

pkg/cmd/cli/debug/debug.go
zubron: If I run this without the --output flag, it still states that it created a `bundle.tar.gz` file (set in [the script](https://github.com/vmware-tanzu/velero/pull/4022/files#diff-759a9f8572722bad288275ce971213be28113874907c5e273565b03c624b9d65R13), but I don't know where it was written to. I don't see any file in the current directory, my home directory, or /tmp.

pkg/cmd/cli/debug/debug.go
zubron: If you can't make this work then I think it's fine to skip setting the debug level.

pkg/cmd/cli/debug/debug.go
zubron: > but I don't know where it was written to. I don't see any file in the current directory, my home directory, or /tmp.

I see it is actually in the /tmp/crashd* directory. I think we need to set the `outputPath` field correctly before passing it in so it is written to the specified location, or in the current directory.

zubron (changes requested):
Thanks for making the suggested changes, @reasonerjt! I've just spotted a few issues with the output path for the tarball, and I have some questions about the k8s library upgrade and the verbose setting but it's looking great otherwise :+1:

pkg/test/api_server.go
reasonerjt: I had to make this change because I saw the UT failure after bumping up `client-go`, similar to https://github.com/kubernetes/client-go/issues/914

pkg/cmd/cli/debug/debug.go
reasonerjt: I did make it generate the filename according to Dave's comment, seems the change was lost after rebasing ... I'll add it back, sorry.

pkg/cmd/cli/debug/debug.go
reasonerjt: I did make it work seems the change was lost after rebasing ... I'll add it back, sorry.

zubron: Great, thanks!

pkg/test/api_server.go
zubron: Thanks :+1: Looking at this again, I don't think the second part of my comment makes sense. I see that we just need to pass in all the resources we want to list with this fake client. I thought it was maybe related to setting up the scheme like we do elsewhere.

zubron (approved):
Thanks for this, @reasonerjt! :smile:

pkg/cmd/cli/debug/debug.go
reasonerjt: How about I submit another PR to update this section?

zubron: Yep, that sounds good. I also think we'll need another follow up post-release to update our issue template: https://github.com/vmware-tanzu/velero/blob/main/.github/ISSUE_TEMPLATE/bug_report.md

zubron (approved):

dsu-igeek (approved):
Looks good, thanks!

Issue #2308, opened 2020-02-28
Restoring nodePort service that has nodePort preservation always fails if service already exists in the namespace


betta1:
**What steps did you take and what happened:**
During restore Velero skips restoring a service if the service already exists in the namespace. However for nodePort services that require node ports to be preserved on restore, restores always fail if the same nodePort service already exists in the target namespace. It seems during the restore Velero tries to create the service on the target namespace but the API server rejects the operation since the node port to be restored is already allocated to the preexisting service. Below steps should help in reproducing this issue:

```
$ kubectl create ns nginx

$ cat <<EOF | kubectl apply -n nginx -f -
apiVersion: v1
kind: Service
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  ports:
    - port: 80
      nodePort: 32555
  selector:
    app: nginx
  type: NodePort
EOF

$ kubectl create deployment nginx --image=nginx -n nginx
$ velero backup create nginx-backup --include-namespaces nginx
$ kubectl delete deployment nginx -n nginx
$ velero restore create --from-backup nginx-backup

```
Restores fail consistently with the error "**provided port is already allocated**":

{"errors":{"namespaces":{"**nginx":["error restoring services/nginx/nginx: Service \"nginx\" is invalid: spec.ports[0].nodePort: Invalid value: 32555: provided port is already allocated"**]}},"warnings":{"namespaces":{"nginx":["could not restore, endpoints \"nginx\" already exists. Warning: the in-cluster version is different than the backed-up version."]}}}

**What did you expect to happen:**
Velero to skip restore of the nodePort service if the service already exists in the target namespace. This maintains restore behavior consistent with how Velero handles restore of other objects it finds in the namespace.

**The output of the following commands will help us better understand what's going on**:
(Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)

* `kubectl logs deployment/velero -n velero`
* `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
* `velero backup logs <backupname>`
* `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
* `velero restore logs <restorename>`

**Anything else you would like to add:**
This issue is closely related to #1950 .

**Environment:**

- Velero version (use `velero version`): v1.2.0
- Velero features (use `velero client config get features`): 
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
[...]

PR #4026, opened 2021-08-09
Distinguish between different unnamed node ports when preserving

### LINKED ISSUE #4025: service action deleteNodePorts doesn't distinguish among unnamed ports
Service action should delete nodeports unless specified in last-applied-configuration. The current code attempts to do this as follows:
if the port is named, then preserve it if the *named* port is in the configuration. Otherwise, if the port is unnamed, then preserve it if the *unnamed* port is in the configuration.

The current code lumps all unnamed ports together -- it preserves all if any exist in the configuration. The match check should compare port number when name is missing.

PR forthcoming (including unit test update to expose the bug)

**Environment:**

- Velero version (use `velero version`): 1.6.2/main

**Vote on this issue!**

sseago:

Service action should delete nodeports unless specified in last-applied-configuration. The current code attempts to do this as follows:
if the port is named, then preserve it if the named port is in the configuration. Otherwise, if the port is unnamed, then preserve it if the unnamed port is in the configuration.

The previous code lumps all unnamed ports together -- it preserves all if any exist in the configuration. The updated match check compares port number when name is missing. unit test update exposes the prior bug.

Fixes #4025

### DIFF
--- a/pkg/restore/service_action.go
+++ b/pkg/restore/service_action.go
@@ -83,6 +83,7 @@ func deleteNodePorts(service *corev1api.Service) error {
 	// to the last-applied-config annotation. We'll retain these values, and
 	// clear out any other (presumably auto-assigned) NodePort values.
 	explicitNodePorts := sets.NewString()
+	unnamedPortInts := sets.NewInt()
 	lastAppliedConfig, ok := service.Annotations[annotationLastAppliedConfig]
 	if ok {
 		appliedServiceUnstructured := new(map[string]interface{})
@@ -123,7 +124,7 @@ func deleteNodePorts(service *corev1api.Service) error {
 						portName, ok := p["name"]
 						if !ok {
 							// unnamed port
-							explicitNodePorts.Insert("")
+							unnamedPortInts.Insert(nodePortInt)
 						} else {
 							explicitNodePorts.Insert(portName.(string))
 						}
@@ -135,8 +136,14 @@ func deleteNodePorts(service *corev1api.Service) error {
 	}
 
 	for i, port := range service.Spec.Ports {
-		if !explicitNodePorts.Has(port.Name) {
-			service.Spec.Ports[i].NodePort = 0
+		if port.Name != "" {
+			if !explicitNodePorts.Has(port.Name) {
+				service.Spec.Ports[i].NodePort = 0
+			}
+		} else {
+			if !unnamedPortInts.Has(int(port.NodePort)) {
+				service.Spec.Ports[i].NodePort = 0
+			}
 		}
 	}
 
--- a/pkg/restore/service_action_test.go
+++ b/pkg/restore/service_action_test.go
@@ -196,6 +196,9 @@ func TestServiceActionExecute(t *testing.T) {
 						{
 							NodePort: 8080,
 						},
+						{
+							NodePort: 9090,
+						},
 					},
 				},
 			},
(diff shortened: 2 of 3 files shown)


pkg/restore/service_action_test.go
```diff
@@ -212,6 +215,7 @@ func TestServiceActionExecute(t *testing.T) {
 						{
 							NodePort: 8080,
 						},
+						{},
```
kaovilai, line 218: What is this for?

sseago: We preserve the `NodePort: 8080` on restore because the last applied config has an entry for it, but the `NodePort: 9090` should be removed, since there is no corresponding config entry. The nodeport removal code doesn't remove entire port entries, just the nodeport element. In a real cluster, the nodeport element is just one attribute of the port. Look at a service yaml output in a real cluster for more context. Also, note the other tests in this file to see examples where there are multiple attributes and just nodeport is removed.

sseago: I added this to the test because the bug in the code was removing or preserving *all* nodeports based on the presence or absence of any unnamed node ports, rather than just preserving the specific unnamed ports found in the config.

zubron (approved):
Thanks!

jenting:
Is this PR able to fix this #2308?

sseago:
@jenting I think that issue is more complicated -- but also, some of the existing nodePort-related changes since that issue was created may have fixed it anyway. This PR is actually just a minor update to the logic around the last-applied-configuration annotation. The logic applied was faulty in the case of unnamed ports, and this fixed that. It's possible that the overall nodeport-deleting code *does* resolve the referenced issue, but if so it's not a result of this PR. It's worth revisiting #2308 and seeing whether it's still an issue. But this PR isn't intended to fix that issue.

dsu-igeek (approved):
LGTM

pkg/restore/service_action.go
```diff
@@ -135,8 +136,14 @@ func deleteNodePorts(service *corev1api.Service) error {
 	}
 
 	for i, port := range service.Spec.Ports {
-		if !explicitNodePorts.Has(port.Name) {
-			service.Spec.Ports[i].NodePort = 0
+		if port.Name != "" {
```
reasonerjt, line 139: @sseago 
Could you please explain why this check is necessary, and could you please add a test case to justify this condition?

sseago: The intent of the preserve node ports functionality is if a named port is found in the last applied configuration that matches a named port in the Service, its `nodePort` field is retained, and if a named port in the service does not have a match in that annotation, the `nodePort` field is removed. For unnamed node ports, we need to perform the same matching on the nodePort value.

The prior version of this inserted an empty entry into explicitNodePorts if there was no name. The effect was that if *any* unnamed node ports were listed in the configuration annotation, then *all* unnamed node ports were preserved, which does not match the intended goal. The fix is to track named nodeports by name and unnamed nodeports by port number. So if this port has a name, we compare it to the node port name list from the annotation, and if this port has no name, then we compare its port number to the unnamed nodeport number list from the annotation.

sseago: For a test of this, see the modified service_action_test below. I've added a second NodePort value that doesn't match the annotation, so we see one is preserved and one is not for the `port.Name == ""` case, and the previously existing named node port test case handles testing this for the `port.Name != ""` case.

sseago: Without this fix, the below test case modification resulted in a test failure since it preserved both unnamed ports rather than just the one that matched.

zubron:
@reasonerjt Did you have any further questions for @sseago regarding this change or are you okay for it to be merged? Thanks :)

PR #4028, opened 2021-08-10
Skip restore of APIServices managed by Kubernetes

### LINKED ISSUE #4027: Velero restores APIService objects for APIs that are no longer served by Kubernetes 1.22
As part of performing upgrade tests for the Velero 1.6.3 release, it was noticed that when restoring cluster scoped backups from Kubernetes 1.21 or earlier onto a Kubernetes 1.22 cluster, Velero would restore [`APIService`](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/api-service-v1/) objects for APIs which have been removed from Kubernetes 1.22 (for example `v1beta1.admissionregistration.k8s.io` or `v1beta1.rbac.authorization.k8s.io`).

When these are restored, it breaks the behaviour of discovery. For example, when running `kubectl api-resources` following the restore, the following is included in the output:
```
error: unable to retrieve the complete list of server APIs: admissionregistration.k8s.io/v1beta1: the server could not find the requested resource, apiextensions.k8s.io/v1beta1: the server could not find the requested resource, authentication.k8s.io/v1beta1: the server could not find the requested resource, authorization.k8s.io/v1beta1: the server could not find the requested resource, certificates.k8s.io/v1beta1: the server could not find the requested resource, coordination.k8s.io/v1beta1: the server could not find the requested resource, extensions/v1beta1: the server could not find the requested resource, networking.k8s.io/v1beta1: the server could not find the requested resource, rbac.authorization.k8s.io/v1beta1: the server could not find the requested resource, scheduling.k8s.io/v1beta1: the server could not find the requested resource
```

Following a discussion with maintainers and community members, Velero should not restore any `APIService` objects which are managed by Kubernetes such as those for built-in APIs or CRDs as these should be created directly when starting the Kubernetes API server or when new CRDs are registered. These can be identified by looking for the `kube-aggregator.kubernetes.io/automanaged` label.

Link to discussion on the Kubernetes slack: https://kubernetes.slack.com/archives/C021GPR1L3S/p1628559718060800

zubron:
It was discovered during Velero 1.6.3 upgrade testing that Velero was
restoring `APIService` objects for APIs that are no longer being served
by Kubernetes 1.22. If these items were restored, it would break the
behaviour of discovery within the cluster.

This change introduces a new RestoreItemAction plugin that skips the
restore of any `APIService` object which is managed by Kubernetes such
as those for built-in APIs or CRDs. The `APIService`s for these will be
created when the Kubernetes API server starts or when new CRDs are
registered. These objects are identified by looking for the
`kube-aggregator.kubernetes.io/automanaged` label.

Fixes #4027


sseago (approved):

pkg/restore/apiservice_action_test.go
```diff
@@ -0,0 +1,147 @@
+/*
+Copyright 2017 the Velero contributors.
```
dsu-igeek, line 2: Please take the year out

pkg/restore/apiservice_action.go
```diff
@@ -0,0 +1,60 @@
...
+func (a *APIServiceAction) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
+	apiService := new(apiregistrationv1.APIService)
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.Item.UnstructuredContent(), apiService); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	output := velero.NewRestoreItemActionExecuteOutput(input.Item)
+
+	if _, ok := apiService.Labels[autoregister.AutoRegisterManagedLabel]; ok {
+		output = output.WithoutRestore()
```
dsu-igeek, line 56: We should log that it's being skipped

pkg/restore/apiservice_action.go
```diff
@@ -0,0 +1,60 @@
...
+// NewAPIServiceAction returns an APIServiceAction which is a RestoreItemAction plugin
+// that will skip the restore of any APIServices which are managed by Kubernetes. This
+// is determined by looking for the "kube-aggregator.kubernetes.io/automanaged" label on
+// the APIService.
+func NewAPIServiceAction(logger logrus.FieldLogger) *APIServiceAction {
+	return &APIServiceAction{logger: logger}
+}
+
+func (a *APIServiceAction) AppliesTo() (velero.ResourceSelector, error) {
+	return velero.ResourceSelector{
```
dsu-igeek, line 42: Should we specify the label here rather than checking in the code?  Looks like the logic in selection should filter on object type, then namespace, then label checking so this should give the right behavior.

dsu-igeek (changes requested):
Let's use the label functionality in ResourceSelector if it works.

pkg/restore/apiservice_action.go
zubron: We have logging in the outer scope that state that the item is being skipped due to a plugin but will add something here to explain why :+1:

pkg/restore/apiservice_action.go
zubron: This is a little trickier to test in unit tests, but I think is the better approach. I've added a missing test case to the restore tests to help cover this.

dsu-igeek (approved):
Looks good, thanks!

sseago (approved):

PR #4044, opened 2021-08-18
pkg/controller: fix dropped test error

alrs:
Signed-off-by: Lars Lehtonen <lars.lehtonen@gmail.com>

This fixes a dropped test `err` variable in `pkg/controller`.
/kind changelog-not-required

### DIFF
--- a/pkg/controller/restore_controller_test.go
+++ b/pkg/controller/restore_controller_test.go
@@ -487,7 +487,7 @@ func TestProcessQueueItem(t *testing.T) {
 						res.Spec.BackupName = backupName
 					}
 
-					return true, res, nil
+					return true, res, err
 				})
 			}
 


pkg/controller/restore_controller_test.go
```diff
@@ -487,7 +487,7 @@ func TestProcessQueueItem(t *testing.T) {
 						res.Spec.BackupName = backupName
 					}
 
-					return true, res, nil
+					return true, res, err
```
dsu-igeek, line 490: Hmmm...not sure about this.  Currently, if there's an error, found comes back false and we'll just leave res.Spec.BackupName empty.  With this change, we return true, meaning we handled it but we also return an error.  I'm inclined to either return false when we hit an error (as we do above) or just explicitly ignore the error (err->_)

dsu-igeek (commented):
Please see my comment, thanks

pkg/controller/restore_controller_test.go
reasonerjt: +1
IMO we should handle it similar to 
https://github.com/vmware-tanzu/velero/pull/4044/files#diff-80e11e742fc916151996fb9749fe054c2c782a5a2945f48d88256a5e95a9164bR469-R471

PR #4060, opened 2021-08-23
Fix log print error using a wrong var name

jacklu1024:
Signed-off-by: jacklu1024 <jacklu1024@outlook.com>

/kind changelog-not-required

Fix log print error using a wrong var name, it is easy to see and fix.


jacklu1024:
My first PR, and I don't know why the DCO still shows failed after I changed the signature....

jacklu1024:
decide to reopen a new one to see if DCO works.

PR #4063, opened 2021-08-24
Fix log print error using a wrong var name

jacklu1024:
Signed-off-by: jacklu1024 <jacklu1024@outlook.com>

this is to Fix log print error using a wrong var name, which is easy and obvious.


jenting (approved):
Thank you

zubron (approved):
Good spot - thanks!

Issue #4067, opened 2021-08-24
Request for four new plugin hooks: Pre backup, Post Backup, Pre Restore and Post Restore


brito-rafa:
**Describe the problem/challenge you have**
We need to perform safe stateful migrations in scale across different clusters distributions and across different storage providers. 
Podhooks are useful for quiesce/unquiesce workloads but we observe multiple times that platform engineers do not have the luxury/visibility/time/knowledge to go to each pod and add specific commands to quiesce/unquiesce workloads. 
Additionally, today Velero’s Restic integration does not backup/migrate orphan PVC/PV pairs (without a pod).

**Describe the solution you'd like**
We propose to create four new plugin hooks:
PreBackup
PostBackup
PreRestore
PostRestore

Once the hooks are implemented on the Velero core, we would write a custom plugin for Velero that would get triggered by the PreBackup hook.
Prior to starting the backup and building the serialization of objects, this plugin would quiesce all the workloads setting replicas=0 on deployments, statefulsets, etc, and then mount all PVC/PV pairs with a staging pod. 
Then Velero would take the backup of all objects: the deployments with replicas=0, and the staging pod and all PVC/PV pairs.

To complete the migration of the workload, a restore would happen on the destination cluster with the quiesced workload. 
A custom Velero plugin would get triggered on the PostRestore hook and unquiesce the workload deleting the staging pod and reinstating the number of the original replicas on the deployment, statefulsets, etc.

Of course, the pre/post backup/restore custom logic can be written outside Velero, but we would need an operator/orchestrator for such. We think Velero can add this capability under its umbrella. The same hooks can integrate Velero with other capabilities (imagine a PreRestore velero plugin that calls Cluster API to scale up cluster size prior to restore).

cc: @dsu-igeek @eleanor-millman @codegold79

**Anything else you would like to add:**
I will put a design proposal for such hooks, please stay tuned. And volunteering to write those if the design is approved.
Once these hooks are part of the Velero core, we can consider opening source the Velero plugin that generically quiesces/unquiesces workloads (with a big disclaimer that specific pre/post pod hooks are preferrable). 

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
[...]

PR #3355, opened 2021-01-26
Enhance plugin interface with timeout

phuongatemc:
Adding context object to the functions of the Velero plugin.  This context object can be used to set timeout for the operation by the caller.  To ensure backward compatibility and avoid forcing all plugin vendor upgrade their code in lock-step, a new interface would be created to wrap around the existing interface and new functions would be created with additional parameter ctx context.Context as the first parameter.  The plugin vendor can decide to implement new function (with context) or keep using existing one.

PR #4083, opened 2021-08-31
#4067 Initial design of the new plugins - pre-post backup and restore

brito-rafa:
Signed-off-by: Rafael Brito <rbrito@vmware.com>

Velero should provide a way to trigger actions before and after each backup and restore.
**Important**: These proposed plugin hooks are fundamentally different from the existing plugin hooks, BackupItemAction and RestoreItemAction, which are triggered per item during backup and restore, respectively.
The proposed plugin hooks are to be executed only once.

Fixes #(issue)
#4067 

- [ NA] Updated the corresponding documentation in `site/content/docs/main`.


dsu-igeek:
Let's review individually and discuss in the next community meeting before merging.

design/new-prepost-backuprestore-plugin-hooks.md
```diff
@@ -0,0 +1,372 @@
...
+
+- Specific implementations of the PreBackupAction, PostBackupAction, PreRestoreAction and PostRestoreAction API beyond test cases.
+
+## High-Level Design
+
+The PreBackupAction plugin API will resemble the BackupItemAction plugin hook design, but with the fundamental difference that it will receive only as input the Velero `Backup` object created.
+It will not receive any resource list items because the backup is not yet running at that stage.
+In addition, the `PreBackupAction` interface will only have an `Execute()` method since the plugin will be executed once per Backup creation, not per item.
+
+The Velero backup controller will be modified so that if there are any PreBackupAction plugins registered, they will be executed after backup object is created and validated but they will execute prior to processing the backup items and volume snapshots. More precisely _before_ function [c.backupper.Backup](https://github.com/vmware-tanzu/velero/blob/74476db9d791fa91bba0147eac8ec189820adb3d/pkg/controller/backup_controller.go#L590). If the PreBackupActions return an err, the function `runBackup` returns it and backup object is not processed.
```
sseago, line 56: Does this mean that if the PreBackupAction modifies/creates resources in the cluster, those new resources and/or changes to resources will be reflected in the backup? On the other hand, if resources are pulled from the cluster for backup after running the pre hook, does that mean Deployments, etc. will be backed up in quiesced state if such a plugin is registered (and thus, on restore, replicas will still be 0)?

design/new-prepost-backuprestore-plugin-hooks.md
```diff
@@ -0,0 +1,372 @@
...
+- Orphan PVC/PV pairs: PVCs/PVs that do not have associated running pods are not backed up and consequently, are not migrated.
+
+Aiming to address these two limitations, and separate from this proposal, we would like to write a Velero plugin that takes advantage of the proposed Pre-Backup plugin hook. This plugin will be executed _once_ (not per resource item) prior backup. It will scale down the applications setting `.spec.replicas=0` to all deployments, statefulsets, daemonsets, replicasets, etc. and will start a small-footprint staging pod that will mount all PVC/PV pairs. Similarly, we would like to write another plugin that will utilize the proposed Post-Restore plugin hook. This plugin will unquiesce migrated applications by killing the staging pod and reinstating original `.spec.replicas values` after the Velero restore is completed.
+
+Other examples of plugins that can use the proposed plugin hooks are
+
+- PostBackupAction: trigger a Velero Restore after a successful Velero backup (and complete the migration operation).
```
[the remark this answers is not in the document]
brito-rafa: Agree.

design/new-prepost-backuprestore-plugin-hooks.md
brito-rafa: yes.

design/new-prepost-backuprestore-plugin-hooks.md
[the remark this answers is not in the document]
brito-rafa: Just a note from what we discussed during the 2021-09-14 community meeting: it is better for the postback plugins to run after the upload of the backup on the BSL. Why? We want to be positive the backup is really completed before doing any further operation. We want to have separate backup/restore phases for these plugins. Let's have their logs persisted separately as well. On the pre restore actions, run proactive sync on the BSL to make sure the latest backup is there.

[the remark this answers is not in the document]
brito-rafa: Added the changes on the design as per feedback from the last call. Will go for another round of feedback.

jaidevmane:
Would love to participate in discussions on this. Our team has a use case for this workflow, where currently we are asking users to run through a two step restore process, and using a post-restore/backup hook would enable us to abstract away the step for them.
https://github.com/vmware-tanzu/velero/issues/1981

brito-rafa:
Please take a look here at this code: this is a functional implementation of this design, without the plugins logging - https://github.com/cschaefer-vmw/velero

For the design, the only thing missing is defining how the logging of post-backup and post-restore plugins will act behind the scenes (the plan is expanding existent interfaces).

design/new-prepost-backuprestore-plugin-hooks.md
```diff
@@ -0,0 +1,713 @@
...
+- Add sub statuses for the plugins on Backup object: `Backup.Status.PreBackupActionsStatuses` and `Backup.Status.PostBackupActionsStatuses`. They will be flagged as optional and nullable. They will be populated only each plugin registered for the PreBackup and PostBackup hooks, respectively.
+- Add sub statuses for the plugins on Restore object: `Backup.Status.PreRestoreActionsStatuses` and `Backup.Status.PostRestoreActionsStatuses`. They will be flagged as optional and nullable. They will be populated only each plugin registered for the PreRestore and PostRestore hooks, respectively.
+- that will be populated optionally if Pre/Post Backup/Restore.
+
+## Non-Goals
+
+- Specific implementations of the PreBackupAction, PostBackupAction, PreRestoreAction and PostRestoreAction API beyond test cases.
+- For migration specific actions (Velero Backup + Velero Restore), add disk synchronization during the validation of the Restore (making sure the newly created backup will show during restore)
+
+## High-Level Design
```
reasonerjt, line 137: A couple of questions:
1.  If a PostBackup action fails, do we consider the backup successful and usable?
2. How does user get the execution log of PostBackup plugin?
3. Once registered, will the plugins per backup/restore take effect to all backups and restores?  Will there be a mechanism such these plugins can be executed selectively?
4. How does user define timeout?  And how will it be handled?

brito-rafa: Answering:
1. Yes
2. The plan is to implement an interface to persist the post backup log on the same location as the backup log (not necessarily inside the gzip file, but on the same directory as a different name).
3. Yes. But the plugin developer can implement on their plugin a mechanism so the plugin to execute selectively, for example, a config map. 
4. This design does not touch on timeouts since current hooks do not provide any provisioning for such. There is an en-route design about adding timeout on plugins, which I propose to bring back to the discussion table once we have these new hooks implemented: https://github.com/vmware-tanzu/velero/pull/3355 .

reasonerjt: > The plan is to implement an interface to persist the post backup log on the same location as the backup log (not necessarily inside the gzip file, but on the same directory as a different name).

If we provide a way for users to collect the backup log, we should also provide a way for users to see the log of the plugin executions.

For 3, if we agree that for most users the plugins should not always be triggered for different backup and restores we should provide a way in the framework to enable user to choose what plugins to be executed or skipped.  IMO, Iinstead of relying on users to workaround, it should be covered in the design.

brito-rafa: On point #4, when we create the post backup log, we will make sure the `velero` command line will be able to collect them. For point #3, do we have similar functionality on existent hooks + plugins? If not, I am glad to put some design doc after this one goes through.

reasonerjt: @brito-rafa 

> On point #4, when we create the post backup log, we will make sure the velero command line will be able to collect them. 

I think you mean point 2, it's probably OK to put it in another PR to make velero CLI return _additional logs_ as it's also needed for itemSnapshotter plugins.

But for point 3, the existing hooks are at a resource level, which doesn't impact all backups and restores, additionally, the user who creates the backup can edit the annotations to control whether the hooks will be executed.  We need to address that for the pre/post backup/restore plugin before this feature is delivered.  Therefore, ideally, it should be covered in this design.

brito-rafa: I concur on annotations. I added a section on the design document about this.

design/new-prepost-backuprestore-plugin-hooks.md
```diff
@@ -0,0 +1,735 @@
...
+```go
+    for _, postRestoreAction := range postRestoreActions {
+        err := postRestoreAction.Execute(restoreReq.Restore)
+        if err != nil {
+            postRestoreLog.Error(err.Error())
+        }
+    }
+```
+
+### Giving the User the Option to Skip the Execution of the Plugins
```
reasonerjt, line 693: Thanks for adding this part @brito-rafa 

I think this approach works but I wanna confirm with other maintainers about the usability in the e2e flow:

1.  When a user uses CLI to create backup how does he/she set the annotations?
2. How does the user know the name of the plugins?

brito-rafa: I am ok (and volunteer) to add capabilities on the client-side to manage plugins execution on each backup or restore (i.e., adding a flag to skip plugins, which translates in adding annotations on the object). I would say this change would go along with the CLI change to download the plugin logs, which we discussed can be implemented incrementally. The user can run `velero plugin get` and use the same name from the plugin list to cherry-pick the plugin to be skipped.

reasonerjt (approved):
LGTM

design/new-prepost-backuprestore-plugin-hooks.md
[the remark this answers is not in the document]
brito-rafa: /kind changelog-not-required

dsu-igeek (approved):
LGTM, thanks!

design/new-prepost-backuprestore-plugin-hooks.md
```diff
@@ -0,0 +1,713 @@
...
+Using Velero for migrations requires an atomic operation involving a Velero backup on a source cluster followed by a Velero restore on a destination cluster.
+
+It is common during these migrations to perform many actions inside and outside Kubernetes clusters.
+**Attention**: these actions are not per resource item, but they are actions to be executed _once_ before and/or after the migration itself (remember, migration in this context is Velero Backup + Velero Restore).
+
+One important use case driving this proposal is migrating stateful workloads at scale across different clusters/storage backends.
+Today, Velero's Restic integration is the response for such use cases, but there are some limitations:
+
+- Quiesce/unquiesce workloads: Pod hooks are useful for quiescing/unquiescing workloads, but platform engineers often do not have the luxury/visibility/time/knowledge to go through each pod in order to add specific commands to quiesce/unquiesce workloads.
+- Orphan PVC/PV pairs: PVCs/PVs that do not have associated running pods are not backed up and consequently, are not migrated.
```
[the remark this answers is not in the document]
jglick: Agreed. I thought there was already an issue about this topic, but if so I cannot find it now due to https://github.com/github/feedback/discussions/6403.

Issue #2290, opened 2020-02-24
Add support for setting securityContext for restic restore action


jaygridley:
**Describe the problem/challenge you have**
We are using restic integration to restore data to our NFS volumes when restoring cluster. NFS volumes have root:root permission. When restic-restore-helper is injected into deployment, where securityContext is set to non-root user, restic does not have permission to the volume. For our deployment we fix this by running other init container that fixes these permissions, but this one is executed after restic.

**Describe the solution you'd like**
Add support for securityContext for restic restore action. As of now it already has capability to read some configuration from ConfigMap, e.g. request limits.

**Environment:**

- Velero version (use `velero version`): 1.2.0
- Kubernetes version (use `kubectl version`): 1.15.7
- Kubernetes installer & version: -
- Cloud provider or hardware configuration: AKS
- OS (e.g. from `/etc/os-release`): Ubuntu 16.04

PR #4084, opened 2021-08-31
Add full support for setting securityContext for restic restore

MatthieuFin:
# Summary of changes

Add a key `secCtx` in configuration configMap "velero.io/restic: RestoreItemAction" to be able to fully customize securityContext set on restic-wait init container set on restore job.

This key is an extension of existing keys `secCtxRunAsUser`, `secCtxRunAsGroup`, `secCtxAllowPrivilegeEscalation`. I keep these old keys support to assure a backward compatibility. But usage of the key `secCtx` could override old keys (unit tests should be clear).

# Issue

Complete implementation for half implemented issue [#2290](https://github.com/vmware-tanzu/velero/issues/2290)


reasonerjt:
@MatthieuFin Thanks for this PR and it looks good to me, could you rebase your PR and we can try to get this reviewed and merged before v1.8 release.

reasonerjt:
moving this out of v1.8 milestone as we are very close to the FC.

eleanor-millman:
@MatthieuFin Do you still want to get this merged? If you could do what @reasonerjt suggested we can get it into 1.9. THanks!

MatthieuFin:
Oh yes, sorry i had broke my notifications settings, il will rebase it asap.

eleanor-millman:
Thanks! I just mentioned it in the slack channel so hopefully we'll get a maintainers's eyes on it soon.  https://kubernetes.slack.com/archives/C021GPR1L3S/p1645634071501639

pkg/util/kube/security_context_test.go
```diff
@@ -30,40 +30,192 @@ func TestParseSecurityContext(t *testing.T) {
...
+		{
+			"valid securityContext with comments only secCtx key",
+			args{"", "", "", `
+capabilities:
+    drop:
+    - ALL
+    add:
+    - cap1
+    - cap2
+sELinuxOptions:
```
jenting, line 70: Question, use "sELinuxOptions" is the same result as "seLinuxOptions"?
I saw the Kubernetes doc uses "seLinuxOptions".

jenting (commented):
LGTM, just a question.

pkg/util/kube/security_context_test.go
MatthieuFin: Absolutely that's a typo error ! I'll change it. Thanks for your attention.

MatthieuFin: After some investigations, that was correct, yaml lib is not case sensitive ... so i you set `CaPAbiliTIes` as your yaml key it will fill the securityContext key `capabilities`.
I don't find a simple way to turn it case sensitive so i propose to keep that behavior (which is due to the standard json library) ?

Some details could be find here
https://pkg.go.dev/encoding/json#Unmarshal
```
Unmarshal matches incoming object keys to the keys used by Marshal (either the struct field name or its tag), preferring an exact match but also accepting a case-insensitive match
```

jenting (approved):

reasonerjt (approved):

Issue #4019, opened 2021-08-06
service RestoreItemAction should clear Spec.ClusterIPs in addition to Spec.ClusterIP.


sseago:
In Kubernetes 1.20, a new Spec field was added to Service. Spec.ClusterIPs. In most cases it contains a single value which is identical to Spec.ClusterIP. If the "IPv6DualStack" feature gate is enabled, then it will have at most 2 elements.

The current service restore action sets Spec.ClusterIP to "" if it's not currently set to "None". We also need to check for the presence of Spec.ClusterIPs and if it exists (it will in 1.20+ clusters but not earlier), then clear it as well if the first element isn't "None"

**Environment:**

- Velero version (use `velero version`): 1.6.2
- Kubernetes version (use `kubectl version`): 1.20+

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

sseago:
I'll put a PR together for this next week.

sseago:
Actually this can be closed, not because it is stale but because it was fixed with https://github.com/vmware-tanzu/velero/pull/4101

PR #4101, opened 2021-09-02
Empty the "ClusterIPs" along with "ClusterIP" when "ClusterIP" isn't "None"

### LINKED ISSUE #4098: Cannot restore "service" with velero build from the main branch
Run restore with velero built from the main branch, get the following errors:
```
error restoring services/kibishii-workload/etcd-client: Service "etcd-client" is invalid: spec.clusterIPs: Invalid value: []string{"100.70.53.8"}: must be empty when clusterIP is empty
error restoring services/kibishii-workload/etcd0: Service "etcd0" is invalid: spec.clusterIPs: Invalid value: []string{"100.66.73.184"}: must be empty when clusterIP is empty
error restoring services/kibishii-workload/etcd1: Service "etcd1" is invalid: spec.clusterIPs: Invalid value: []string{"100.69.23.44"}: must be empty when clusterIP is empty
error restoring services/kibishii-workload/etcd2: Service "etcd2" is invalid: spec.clusterIPs: Invalid value: []string{"100.64.23.70"}: must be empty when clusterIP is empty
```

This is caused by that we bumped up the version of library `k8s.io/api/core/v1` from `v0.19.12` to `v0.20.9` in PR #4022.
There is a new property `ClusterIPs` introduced in `v0.20.9` https://github.com/kubernetes/api/blob/v0.20.9/core/v1/types.go#L4098 and the restore action empties the `ClusterIP` but not for `ClusterIPs` https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/service_action.go#L56

**Vote on this issue!**

ywk253100:
For more details please refer to #4098

Fixes #4098

Fixes #(issue)

### DIFF
--- a/pkg/restore/service_action.go
+++ b/pkg/restore/service_action.go
@@ -55,6 +55,7 @@ func (a *ServiceAction) Execute(input *velero.RestoreItemActionExecuteInput) (*v
 
 	if service.Spec.ClusterIP != "None" {
 		service.Spec.ClusterIP = ""
+		service.Spec.ClusterIPs = nil
 	}
 
 	/* Do not delete NodePorts if restore triggered with "--preserve-nodeports" flag */
--- a/changelogs/unreleased/4101-ywk253100
+++ b/changelogs/unreleased/4101-ywk253100
@@ -1,0 +1,1 @@
+Empty the "ClusterIPs" along with "ClusterIP" when "ClusterIP" isn't "None"
--- a/pkg/restore/service_action_test.go
+++ b/pkg/restore/service_action_test.go
@@ -73,13 +73,14 @@ func TestServiceActionExecute(t *testing.T) {
 		expectedRes corev1api.Service
 	}{
 		{
-			name: "clusterIP (only) should be deleted from spec",
+			name: "clusterIP/clusterIPs should be deleted from spec",
 			obj: corev1api.Service{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "svc-1",
 				},
 				Spec: corev1api.ServiceSpec{
 					ClusterIP:      "should-be-removed",
+					ClusterIPs:     []string{"should-be-removed"},
 					LoadBalancerIP: "should-be-kept",
 				},
 			},


pkg/restore/service_action.go
```diff
@@ -55,6 +55,7 @@ func (a *ServiceAction) Execute(input *velero.RestoreItemActionExecuteInput) (*v
 
 	if service.Spec.ClusterIP != "None" {
 		service.Spec.ClusterIP = ""
+		service.Spec.ClusterIPs = nil
```
zubron, line 58: I noticed that there was already an issue opened for this problem: #4019. The docs state that `ClusterIPs` can have at most 2 values in a dualstack environment. Do we need to check the contents of `ClusterIPs` in addition to this check for `ClusterIP`? @sseago What do you think?

ywk253100: Per my understanding, in a single-stack environment, the `ClusterIPs` contains only one element and the element value is the same with `ClusterIP`; in a dual-stack environment, the `ClusterIPs` could hold two elements and the value of the first element is the same with `ClusterIP`. The `ClusterIP` and `ClusterIPs` play the same role.
The code `service.Spec.ClusterIP != "None" {} ` aims to clear the IP address of the service, no matter it's in single-stack or dual-stack environment.

Correct me if I'm wrong @sseago

zubron: Ah, that makes sense - thanks for the explanation!

sseago: My original plan with this was to empty `ClusterIPs` when `ClusterIPs` did not have "None" in its first element. However, thinking about it again, when `ClusterIPs` has `None`, `ClusterIP` should also have `None`, so this is a simpler approach. It would be good to confirm in a dual-stack environment what the values of these two fields are both for services with `None` and for those with actual IP addresses, though.

sseago (approved):
LGTM. It would be good to confirm the behavior in an actual dual-stack environment to make sure our assumptions about the appearance of `None` in both fields is correct, but I don't think that's necessary prior to merge.

zubron (approved):

PR #4126, opened 2021-09-09
Verify group before treating resource as cohabitating

### LINKED ISSUE #4125: Cohabitating resource check on backup ignores groups
Resources with the same names (but different groups) as existing cohabitating resources are treated as cohabitating when making a backup. 

The cohabitating resource code ensures that only one copy of cohabitating resources gets copied. For example, we don't want to back up a single deployment as both deployments.apps and deployments.extensions. We don't want to back up the same event as "events" and as "events.events.k8s.io".

The problem is that the code which checks for cohabitating resources in item_collector only looks at the resource name, not whether it belongs to one of the cohabitating groups. As a result, if I were to create a CRD called "deployments.foo", then that would be treated as cohabitating with "deployments.apps" rather than as an independent resource. We need to confirm that the resource in question belongs to one of the cohabitating groups before setting `cohabitator.seen`.

PR forthcoming.

**Vote on this issue!**

sseago:

This PR adds group checking to the cohabitating resource handling. Before setting cohabitator.seen or checking for cohabitator.seen, we first confirm that this resource belongs to one of the linked groups.
Also included is a unit test which confirms the fix.

Fixes #4125

### DIFF
--- a/changelogs/unreleased/4126-sseago
+++ b/changelogs/unreleased/4126-sseago
@@ -1,0 +1,1 @@
+Verify group before treating resource as cohabitating
--- a/pkg/backup/backup_test.go
+++ b/pkg/backup/backup_test.go
@@ -970,6 +970,30 @@ func TestBackupResourceCohabitation(t *testing.T) {
 				"resources/deployments.apps/v1-preferredversion/namespaces/zoo/raz.json",
 			},
 		},
+		{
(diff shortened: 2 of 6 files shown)


sseago:
@alaypatel07 This is the fix we were discussing earlier.

ywk253100 (approved):

dsu-igeek:
Will finish review and merge after 1.7 RC2

zubron (approved):

PR #4162, opened 2021-09-19
Check the failed phases either when uploading the snapshot in E2E testing

ywk253100:
When the snapshot uploading is failed, it should not be treat as completed and continue.
This commit covers both the phases of in progress and failed when uploading snapshot with vSphere plugin

Fixes #(issue)

### DIFF
--- a/test/e2e/velero_utils.go
+++ b/test/e2e/velero_utils.go
@@ -449,11 +449,12 @@ func waitForVSphereUploadCompletion(ctx context.Context, timeout time.Duration, 
 			// Canceled - the operation was canceled, the snapshot ID is not valid
 			if len(comps) == 2 {
 				phase := comps[1]
-				if phase == "New" ||
-					phase == "InProgress" ||
-					phase == "Snapshotted" ||
-					phase == "Uploading" {
+				switch phase {
+				case "Uploaded":
+				case "New", "InProgress", "Snapshotted", "Uploading":
 					complete = false
+				default:
+					return false, fmt.Errorf("unexpected snapshot phase: %s", phase)
 				}
 			}
 		}


test/e2e/velero_utils.go
```diff
@@ -386,10 +386,8 @@ func waitForVSphereUploadCompletion(ctx context.Context, timeout time.Duration,
 			// Canceled - the operation was canceled, the snapshot ID is not valid
 			if len(comps) == 2 {
 				phase := comps[1]
-				if phase == "New" ||
-					phase == "InProgress" ||
-					phase == "Snapshotted" ||
-					phase == "Uploading" {
+				// if the phase isn't "Uploaded", the uploading is either in progress or failed(will retry, if the retrying cannot recover, will timeout in the final)
+				if phase != "Uploaded" {
```
zubron, line 390: We should return an error in the case where the snapshot reached a failed terminal state as it's not possible to recover from that. By checking for `Uploaded` only, this function could poll for up to an hour for a snapshot that we know can't be used.

zubron (changes requested):
This was a good spot, thanks! We should return from the function earlier though if we know the snapshot has failed rather than relying on the timeout to signal an error.

test/e2e/velero_utils.go
ywk253100: @zubron I noticed that seems the uploading will be retry by the vSphere plugin, I'm not sure whether the failed uploading can be fixed by the retrying, so don't return the error here

Should we expect the retrying may fix the failed uploading or just return the error?

zubron: Oh, sorry - I didn't realize that it would retry the upload :woman_facepalming: In that case, it's probably best to ignore my comment and check with @dsu-igeek :)

dsu-igeek (approved):
LGTM

zubron (approved):

PR #4295, opened 2021-10-28
Add metrics backup_items_total and backup_items_errors

### LINKED ISSUE #4296: Add backup items metrics

As an operator I would like to have some additional metrics to get more information about the items of backups.

Add new metrics to the backup controller.

**Environment:**

- Velero version (use `velero version`): v1.7.0
- Kubernetes version (use `kubectl version`): v1.21.6
- Cloud provider or hardware configuration: OpenStack

**Vote on this issue!**

<sub>Tobias Giese <tobias.giese@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

tobiasgiese:

This PR implements 2 additional metrics

* `backup_items_total`
* `backup_items_errors`

These metrics are referring to the status of the `Backups` CR.

Fixes #4296 

/kind changelog-not-required

<sub>Tobias Giese <tobias.giese@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -627,14 +627,14 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
 		}
 	}
 
+	backup.Status.Warnings = logCounter.GetCount(logrus.WarnLevel)
+	backup.Status.Errors = logCounter.GetCount(logrus.ErrorLevel)
+
 	recordBackupMetrics(backupLog, backup.Backup, backupFile, c.metrics)
 
 	if err := gzippedLogFile.Close(); err != nil {
 		c.logger.WithField(Backup, kubeutil.NamespaceAndName(backup)).WithError(err).Error("error closing gzippedLogFile")
 	}
-
-	backup.Status.Warnings = logCounter.GetCount(logrus.WarnLevel)
-	backup.Status.Errors = logCounter.GetCount(logrus.ErrorLevel)
 
 	// Assign finalize phase as close to end as possible so that any errors
 	// logged to backupLog are captured. This is done before uploading the
@@ -685,6 +685,10 @@ func recordBackupMetrics(log logrus.FieldLogger, backup *velerov1api.Backup, bac
 	serverMetrics.RegisterVolumeSnapshotAttempts(backupScheduleName, backup.Status.VolumeSnapshotsAttempted)
 	serverMetrics.RegisterVolumeSnapshotSuccesses(backupScheduleName, backup.Status.VolumeSnapshotsCompleted)
 	serverMetrics.RegisterVolumeSnapshotFailures(backupScheduleName, backup.Status.VolumeSnapshotsAttempted-backup.Status.VolumeSnapshotsCompleted)
+	if backup.Status.Progress != nil {
+		serverMetrics.RegisterBackupItemsTotalGauge(backupScheduleName, backup.Status.Progress.TotalItems)
+	}
+	serverMetrics.RegisterBackupItemsErrorsGauge(backupScheduleName, backup.Status.Errors)
 }
 
 func persistBackup(backup *pkgbackup.Request,
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -43,6 +43,8 @@ const (
 	backupDeletionSuccessTotal    = "backup_deletion_success_total"
 	backupDeletionFailureTotal    = "backup_deletion_failure_total"
 	backupLastSuccessfulTimestamp = "backup_last_successful_timestamp"
+	backupItemsTotalGauge         = "backup_items_total"
+	backupItemsErrorsGauge        = "backup_items_errors"
 	restoreTotal                  = "restore_total"
 	restoreAttemptTotal           = "restore_attempt_total"
 	restoreValidationFailedTotal  = "restore_validation_failed_total"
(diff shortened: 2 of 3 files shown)


pkg/controller/backup_controller.go
```diff
@@ -685,6 +685,11 @@ func recordBackupMetrics(log logrus.FieldLogger, backup *velerov1api.Backup, bac
 	serverMetrics.RegisterVolumeSnapshotAttempts(backupScheduleName, backup.Status.VolumeSnapshotsAttempted)
 	serverMetrics.RegisterVolumeSnapshotSuccesses(backupScheduleName, backup.Status.VolumeSnapshotsCompleted)
 	serverMetrics.RegisterVolumeSnapshotFailures(backupScheduleName, backup.Status.VolumeSnapshotsAttempted-backup.Status.VolumeSnapshotsCompleted)
+	if backup.Status.Progress != nil {
+		serverMetrics.RegisterBackupItemsBackedUpGauge(backupScheduleName, backup.Status.Progress.ItemsBackedUp)
+		serverMetrics.RegisterBackupItemsTotalGauge(backupScheduleName, backup.Status.Progress.TotalItems)
```
zubron, line 690: We only record metrics once at the end of a backup, so my understanding is that, at the point of registering this value, `ItemsBackedUp` and `TotalItems` will be equal.

pkg/metrics/metrics.go
```diff
@@ -43,6 +43,9 @@ const (
 	backupDeletionSuccessTotal    = "backup_deletion_success_total"
 	backupDeletionFailureTotal    = "backup_deletion_failure_total"
 	backupLastSuccessfulTimestamp = "backup_last_successful_timestamp"
+	backupItemsBackedUpGauge      = "backup_items_backed_up"
+	backupItemsTotalGauge         = "backup_items_total"
```
zubron, line 47: For non-scheduled, manual backups, it's not clear to me what value this metric will provide. Given that any combination of items could be included in manual backups, this gauge value could fluctuate a lot. Do you mostly want to monitor how many items are included in scheduled backups?

tobiasgiese: You're able to add a label to your query to filter the metric. We're using Velero to backup about 700 clusters and we want to monitor the error rate.

pkg/controller/backup_controller.go
tobiasgiese: `TotalItems` can differ from `ItemsBackedUp` according to the struct. So I think both metrics should be exposed.

https://github.com/vmware-tanzu/velero/blob/ae6e1df9aa91816dcd5d055a0325e1d54a37d3d5/pkg/apis/velero/v1/backup.go#L315-L328

pkg/metrics/metrics.go
tobiasgiese: For example, if you have a `Schedule` you can filter for the label `schedule=schedulename`

tobiasgiese: We discovered some problems using a not correct undeployed cert-manager. These CRs are still deployed, but the webhook isn't accessible anymore. That's why Velero isn't able to backup those CRs and the backup is becoming `PartiallyFailed`

pkg/controller/backup_controller.go
zubron: Yes, those values will be different while the backup is in progress (that is how progress information is captured and stored on the Backup CR), however, we set these are set to the same value at the end of the backup (see https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/backup.go#L413-L416). As we record metrics at the end of the backup, they'll be the same.

tobiasgiese: Ah okay, I'll remove the `backup_items_backed_up` then. Thanks for clarification 🙂

tobiasgiese: @zubron I removed the `backup_items_backed_up` gauge. Can you PTAL?

tobiasgiese:
@zubron do you know when this PR is ready for merge?

pkg/metrics/metrics.go
```diff
@@ -179,6 +181,22 @@ func NewServerMetrics() *ServerMetrics {
...
+					Name:      backupItemsTotalGauge,
+					Help:      "Total number of items backed up",
+				},
+				[]string{scheduleLabel},
+			),
+			backupItemsErrorsGauge: prometheus.NewGaugeVec(
+				prometheus.GaugeOpts{
+					Namespace: metricNamespace,
+					Name:      backupItemsErrorsGauge,
+					Help:      "Total number of items which could not be backed up",
```
sseago, line 196: I wonder whether it would be better to just say something like "Total number of errors encountered during backup" -- I don't know that the number of errors is guaranteed to be equal to the number of items that failed to back up. Errors is the error message count rather than explicitly an item count. It's possible that the number of errors and the number of items aren't identical (i.e. an error message generated separate from an item backup failure, or if more than one error log message is generated for certain failure scenarios).

tobiasgiese: Totally fine for me to change the text, sure. 

```suggestion
					Help:      "Total number of errors encountered during backup",
```

pkg/controller/backup_controller.go
```diff
@@ -685,6 +685,10 @@ func recordBackupMetrics(log logrus.FieldLogger, backup *velerov1api.Backup, bac
 	serverMetrics.RegisterVolumeSnapshotAttempts(backupScheduleName, backup.Status.VolumeSnapshotsAttempted)
 	serverMetrics.RegisterVolumeSnapshotSuccesses(backupScheduleName, backup.Status.VolumeSnapshotsCompleted)
 	serverMetrics.RegisterVolumeSnapshotFailures(backupScheduleName, backup.Status.VolumeSnapshotsAttempted-backup.Status.VolumeSnapshotsCompleted)
+	if backup.Status.Progress != nil {
```
jenting, line 688: Why not always register it?

tobiasgiese: This is just the handling of the nil pointer exception. We can register it with a zero value otherwise if that's feasible.

jenting (commented):
Since you're addressing an issue, having a change log would be good.

tobiasgiese:
> Since you're addressing an issue, having a change log would be good.

Added the changelog file. Could you PTAL? Thanks 🙂

jenting (approved):

sseago (approved):

PR #4351, opened 2021-11-16
Adding ObjectStoreV2 plugin.

phuongatemc:
Adding ObjectStoreV2 plugin interface into plugin framework
This is only to share.  DO NOT MERGE to main.


dsu-igeek:
Since we shouldn't merge this, please make this a draft PR then we won't do it by accident.

PR #4377, opened 2021-11-20
Remove backups and restic repos associated with deleted BSL(s)

### LINKED ISSUE #2082: when a backup storage location is deleted, remove associated backups & restic repos from the cluster
cluster migration scenario, using velero v1.2.0.
On my new-cluster, I add a read-only second storage-location (containing the backups of the old-cluster). I restore my backups from the old-cluster to the new-cluster. When done, I delete the second storage-location on my new-cluster.

problem: `velero backup get` still shows me the backups I have on the storage-location I have now removed. The restic repositories are still known as well
`kubectl -n velero get backups.velero.io` and `resticrepositories.velero.io` still show stuff from the storage-location that does not show up anymore in the output of `kubectl -n velero get backupstoragelocations.velero.io`

after I run `kubectl -n velero delete backupstoragelocations.velero.io old-cluster`, I expect to not see anything related to `old-cluster`.

Also, the command-line tool should support a more intuitive `velero backup-location delete old-cluster` call IMO

codegold79:

# Summary of change

Added cli logic to look for backups and restic repositories that have the `velero.io/storage-location=<deleted-bsl-name>` label. The Velero client will then delete those backup and restic repos.

Regarding tests, I added expectations in @danfengliu's test cases plan where a BSL is deleted (I believe the test case documentation will be published to this Velero repo soon, after it's completed).

I looked for documentation to change, but there is nothing mentioned about deleting BSLs and so nothing to update.

# Change fixes an issue

Fixes #2082 


jenting (commented):
Question: will remove the backup, Velero deletes the backup CR only? Or will Velero also deletes the remote backup file?

codegold79:
@jenting asked,
> Question: will remove the backup, Velero deletes the backup CR only? Or will Velero also deletes the remote backup file?

This change will only delete the backup and Restic repository CRs only. The actual backup tarballs in the BSL will be retained.

pkg/cmd/cli/backuplocation/delete.go
```diff
@@ -120,7 +121,65 @@ func Run(f client.Factory, o *cli.DeleteOptions) error {
 			continue
 		}
 		fmt.Printf("Backup storage location %q deleted successfully.\n", location.Name)
+
+		// Delete backups associated with the deleted BSL.
+		backupList, err := findAssociatedBackups(kbClient, location.Name, f.Namespace())
+		if err != nil {
+			errs = append(errs, fmt.Errorf("find backups associated with BSL %q: %w", location.Name, err))
+		}
+		if deleteErrs := deleteBackups(kbClient, backupList); deleteErrs != nil {
```
ywk253100, line 130: We should continue to delete the backups only `if err == nil`

pkg/cmd/cli/backuplocation/delete.go
```diff
@@ -120,7 +121,65 @@ func Run(f client.Factory, o *cli.DeleteOptions) error {
...
+		Raw:       &metav1.ListOptions{LabelSelector: bslLabelKey + "=" + bslName},
+	})
+	return repos, err
+}
+
+func deleteBackups(client kbclient.Client, backups velerov1api.BackupList) []error {
+	var errs []error
+	for _, backup := range backups.Items {
+		if err := client.Delete(context.Background(), &backup, &kbclient.DeleteOptions{}); err != nil {
+			errs = append(errs, errors.WithStack(fmt.Errorf("delete backup %q associated with deleted BSL: %w", backup.Name, err)))
```
ywk253100, line 169: Need a `continue` here, or the `Backup associated with deleted BSL(s) %q deleted successfully` will be printed even though we get errors

pkg/cmd/cli/backuplocation/delete.go
codegold79: I think line 139 will only execute if `err == nil` because of the `if err != nil` on line 127.

If you are talking about errors that come from `deleteBackups(...)`, then I don't think we should change line 130.

Let me know if I am mistaken or misunderstood you @ywk253100.

pkg/cmd/cli/backuplocation/delete.go
codegold79: Good catch. I'll add a `continue` after the append errors call on line 169.

pkg/cmd/cli/backuplocation/delete.go
ywk253100: What I said is like this:
```go
backupList, err := findAssociatedBackups(kbClient, location.Name, f.Namespace())
if err != nil {
	errs = append(errs, fmt.Errorf("find backups associated with BSL %q: %w", location.Name, err))
} else {
        if deleteErrs := deleteBackups(kbClient, backupList); deleteErrs != nil {
                 errs = append(errs, deleteErrs...)
       }
}
		
```

dsu-igeek: Wenkai is pointing out that your backupList is probably bad when an error is returned from findAssociatedBackups at line 126

pkg/cmd/cli/backuplocation/delete.go
```diff
@@ -120,7 +121,65 @@ func Run(f client.Factory, o *cli.DeleteOptions) error {
...
+		if deleteErrs := deleteBackups(kbClient, backupList); deleteErrs != nil {
+			errs = append(errs, deleteErrs...)
+		}
+
+		// Delete Restic repositories associated with the deleted BSL.
+		resticRepoList, err := findAssociatedResticRepos(kbClient, location.Name, f.Namespace())
+		if err != nil {
+			errs = append(errs, fmt.Errorf("find Restic repositories associated with BSL %q: %w", location.Name, err))
+		}
+		if deleteErrs := deleteResticRepos(kbClient, resticRepoList); deleteErrs != nil {
```
dsu-igeek, line 139: Should not call deleteResticRepos is we get an error from findAssociatedResticRepos because resticRepoList is probably bad.

dsu-igeek (changes requested):
Please make sure to not use the bad return values

pkg/cmd/cli/backuplocation/delete.go
codegold79: Ah, I see what you mean now, @ywk253100. And thanks @dsu-igeek for explaining. Yes, I think I should not continue to delete the backups if there was a problem with finding them.

Similarly, I shouldn't be deleting the Restic repos (call on line 139) if finding them returns an error.

Another good catch, @ywk253100. Thank you for your careful reading.

Since I'm on vacation, I'll be making changes next week.

ywk253100 (approved):
LGTM

dsu-igeek (approved):
lgtm

Issue #4369, opened 2021-11-18
Migrate BackupStorageLocation from kubebuilder v2 to v3


blackpiglet:
**Describe the problem/challenge you have**
As an item of epic https://github.com/vmware-tanzu/velero/issues/4307, need to migrate BackupStorageLocation from kubebuilder v2 to v3.

**Describe the solution you'd like**
Velero is going to use kubebuilder v3 to generate operator code in v1.8, so migrating BackupStorageLocation that bases on kubebuilder v2 first.

**Anything else you would like to add:**
None

**Environment:**

- Velero version (use `velero version`): v1.8
- Kubernetes version (use `kubectl version`): v1.21
- Kubernetes installer & version: kind
- Cloud provider or hardware configuration: none
- OS (e.g. from `/etc/os-release`): MacOS

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

blackpiglet:
Fixed by PR #4382. Close issue.

PR #4382, opened 2021-11-22
feat: migrate velero controller from kubebuilder v2 to v3

blackpiglet:
1. remove config/crd/v1beta1
2. remove PROJECT file
3. update controller-gen and kubebuilder version
4. generate client and CRD file
5. add changelog and remove v1beta1 CRD generated code.
6. add kubebuilder test bundle setup command.
7. due to apiextensions.k8s.io/v1beta1 is not supported, only k8s after v1.16 is supported, so remove v1.15 check.
8. add CRD and k8s suppored version update in changelog.

Fixes #(issue)


pkg/install/resources.go
```diff
@@ -226,11 +225,6 @@ func AllCRDs(perferredAPIVersion string) *unstructured.UnstructuredList {
 	resources.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "List"})
 
 	switch perferredAPIVersion {
-	case "v1beta1":
```
reasonerjt, line 229: Per my understanding after moving to kubebuilder v3, all CRDs of velero will be v1, so the `preferredAPIVersion` is unnecessary?

blackpiglet: In release 1.8, there is only one possible version, v1, but preferredAPIVersion is passed from velero client command line parameter "--crds-version". 
I think it may be useful for future potential multiple versions need. 
In my opinion, keeping the logic there and adding validation to only support v1 may be better.

reasonerjt: Please check with @zubron 
By checking the commit:
https://github.com/vmware-tanzu/velero/commit/fca2b5c417ed87facb08038273d6af5ccd46b524

It seems this was added as a temp workaround.

If the conclusion is to remove this param I'm fine to handle that in another PR/issue so that this big PR is not blocked.

changelogs/unreleased/4382-jxun
```diff
@@ -0,0 +1 @@
+feat: migrate velero controller from kubebuilder v2 to v3
```
zubron, line 1: This changelog should include details to state that v1beta1 CRDs are no longer supported and that we only support Kubernetes 1.16 and later.

pkg/install/resources.go
zubron: Yes, this flag was only in place to allow users to explicitly set which version to use if needed. Given that we always use v1 now, you can remove this flag.

There is also some code in the uninstall path to remove v1beta1 CRDS (https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/cli/uninstall/uninstall.go#L139-L150) but I think we may want to keep that for now to handle any v1beta1 CRDs that a user may have installed. We can probably remove that in 1.9.

zubron (commented):
LGTM - thanks @blackpiglet! I only have one comment on the changelog but everything else looks good :+1:

changelogs/unreleased/4382-jxun
blackpiglet: Got it. Thanks.

pkg/install/resources.go
blackpiglet: Thanks. I think the agreement is to remove the CRD version selection parameter in install command, and keep the v1beta1 check in uninstall command until 1.9 release.

blackpiglet:
#4389 is opened to trace --crds-version removing.

reasonerjt:
Could you squash to commits?

blackpiglet:
Sure. Commits are squashed.

zubron (approved):
LGTM - thanks!

reasonerjt (approved):

blackpiglet:
Fix issue #4369, #4370 and #4371

Issue #2587, opened 2020-05-28
Multi-Tenancy and Self Service


stephbman:
**Description**
Reduce friction by enabling developers to backup their namespaces via self-service. Introduce a Velero multi-tenancy model, enabling owners of namespaces to backup and restore within their access scope.

- Product requirements to be linked
- user stories to be created
- current timeframe target December 2020 (ref. velero roadmap): https://github.com/vmware-tanzu/velero/pull/2548/files?short_path=38574c0#diff-38574c080d4e2eb38c49b86e6588ad98

skriss:
FYI https://github.com/vmware-tanzu/velero/issues/18 and https://github.com/vmware-tanzu/velero/issues/1217 are the existing issues I'm aware of that relate to this. We may want to close 1+ of them if they're dupes.

stephbman:
@skriss I was thinkig of linking them for now until I have the PRD drafted (I have them linked to the PRD currently).

rflorenc:
@stephbman I've recently been working on something very basic that seems to relate to this epic:
https://github.com/rflorenc/drwatcher-operator/tree/develop

It only supports self service creation of Backup and Schedule objects for now, by reconciling a simple CR per namespace.

Let me know if we can help somehow.

sspreitzer:
Any update on this epic, which was created 4 years ago?

blackpiglet:
Velero team discussed about this topic, and we think Multi-tenancy is not a high-priority task.

gcezaralmeida:
Hey Everyone, That would be a nice feature!

PR #4393, opened 2021-11-23
Created a proposal for multi-tenancy via namespace controls

### LINKED ISSUE #18: Multi-tenancy support
Allow users other than cluster-admin to use Ark. This has security implications:

- need to ensure a user can only back up/restore PVs/snapshots they have access to
- need to ensure a user can only back up kube resources they have access to
- need to find a way to have the restore "run as" the user, so there's no privilege escalation

dsu-igeek:

Draft of a proposal for handling multi-tenancy on a per-namespace basis

Fixes #18, #2587

- [X] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.


jacklu1024:
I agree that each user should only operate in their own namespace. So BSL, Backup, Schedule, Restore are only created in user's own namespace.  
Velero could monitor all namespaces, but only look for related resources in that namespace, so to help with multi-tenancy.

if user creates namespace A, he created BSL, and Backup CR in namespace A, but he wants to backup namespace B, which is owned by this user. Velero code should check if user owns namespace B, and then start to do backup.

PR #4429, opened 2021-12-06
Adds <backup-name>-itemsnapshots.gz file to backup (when provided).

### LINKED ISSUE #3758: [Upload Progress Monitoring Epic #3533] Implement <backup-name>-itemsnapshots.json.gz file
As part of upload progress monitoring (Epic #3533) we need to track snapshots taken using the new SnapshotItemAction plugin.  Current design (upload-progress.md) has these being stored in a <backup-name>-itemsnapshots.json.gz file but this could be changed.  Define how the snapshot info will be stored, attempt to future-proof with Astrolabe coming changes.  Discuss design with community before implementing.

dsu-igeek:
Adds `<backup-name>-itemsnapshots.gz` file to backup (when provided). Also
adds DownloadTargetKindBackupItemSnapshots type to allow downloading.

Fixes #3758


pkg/persistence/object_store.go
```diff
@@ -225,50 +227,51 @@ func (s *objectBackupStore) ListBackups() ([]string, error) {
...
+			// Uploading the log file is best-effort; if it fails, we log the error but it doesn't impact the
+			// backup's status.
+			s.logger.WithError(err).WithField("backup", info.Name).Error("Error uploading log file")
+		}
 	}
 
-	if err := seekAndPutObject(s.objectStore, s.bucket, s.layout.getBackupMetadataKey(info.Name), info.Metadata); err != nil {
-		// failure to upload metadata file is a hard-stop
-		return err
+	if info.Metadata != nil {
```
reasonerjt, line 238: I'm against this check, but could you remind me in what case the `Metadata`, `Content`, `Log` or `reader` is nil?
If it should not be nil, like mentioned in the old comment
```
there's no point in continuing
```
Maybe we should stop and fail early?

dsu-igeek: I split the backup upload into the object store into two pieces for the upload progress monitoring.  It used to all go as one unit.  The Metadata goes into the velero-backup.json file.  That controls whether the backup is considered usable.  reader is going to be for the things that are in the backupObjs map and those are all optional and any of them can be nil.  I think the extra check I added for nil there is redundant, I'll take that out.

We need to upload things like the backup tar file itself (Content), the log files (Log) and possibly things that were in the backupObjs map  when we finish phase 1 of the backup, but without the ItemSnapshots (which would be in the backupObjs map).  After we finish phase 1, we will go into the monitoring phase and another backup may begin.  The in-memory structures may disappear if we take a restart so the upload after the final phase just consists of the Metadata and ItemSnapshots and everything else will be nil.  It seemed reasonable to leave everything together in PutBackup as the checks were mainly for panic situations.

zubron (approved):

pkg/persistence/object_store.go
dsu-igeek: OK, done

reasonerjt (approved):

PR #4436, opened 2021-12-09
Convert Pod Volume Backup resource/controller to the Kubebuilder V3 framework

### LINKED ISSUE #3454: Convert Pod Volume Backup resource/controller to the Kubebuilder framework
Convert Pod Volume Backup resource/controller to the Kubebuilder framework

**Vote on this issue!**

- 👍 for "The project would be better with this feature added"
- 👎 for "This feature will not enhance the project in a meaningful way"

codegold79:
# Summary of change

Convert Pod Volume Backup resource/controller to the Kubebuilder V3 framework. Write Ginkgo tests.

Fixes #3454 


pkg/cmd/cli/restic/server.go
```diff
@@ -207,18 +206,25 @@ func (s *resticServer) run() {
...
+		Metrics:        s.metrics,
+		CredsFileStore: credentialFileStore,
+		NodeName:       os.Getenv("NODE_NAME"),
+		FileSystem:     filesystem.NewFileSystem(),
+		ResticExec:     restic.BackupExec{},
+		Log:            s.logger,
+
+		PvLister:  s.kubeInformerFactory.Core().V1().PersistentVolumes().Lister(),
+		PvcLister: s.kubeInformerFactory.Core().V1().PersistentVolumeClaims().Lister(),
+		PvbLister: s.veleroInformerFactory.Velero().V1().PodVolumeBackups().Lister(),
```
blackpiglet, line 223: Pod Volume Backup CRD is moved to controller-time. Maybe a seperated Lister is not needed. 
The Client should've been enable to get PVB.

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -18,330 +18,220 @@ package controller
...
+		log.WithError(err).Error("error getting a patch helper to update this resource")
+		return ctrl.Result{}, err
 	}
 
-	pathGlob := fmt.Sprintf("/host_pods/%s/volumes/*/%s", string(req.Spec.Pod.UID), volumeDir)
-	log.WithField("pathGlob", pathGlob).Debug("Looking for path matching glob")
+	defer func() {
+		// Attempt to Patch the pvb object and status after each reconciliation.
+		if err := patchHelper.Patch(r.Ctx, &pvb); err != nil {
+			log.WithError(err).Error("error updating download request")
```
blackpiglet, line 107: The error message should've not be "error updating download request".

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -18,330 +18,220 @@ package controller
...
-	c.cacheSyncWaiters = append(
-		c.cacheSyncWaiters,
-		podVolumeBackupInformer.Informer().HasSynced,
-		podInformer.HasSynced,
-		pvcInformer.Informer().HasSynced,
-	)
-	c.processBackupFunc = c.processBackup
-
-	podVolumeBackupInformer.Informer().AddEventHandler(
-		cache.ResourceEventHandlerFuncs{
```
blackpiglet, line 110: In previous version, controller only care about Create and Update event of Pod Volume Backup. 
Can the new version controller still keep the event filter function?

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -379,68 +272,111 @@ func getParentSnapshot(log logrus.FieldLogger, pvcUID, backupStorageLocation str
...
-			r.Status.Progress = progress
-		}); err != nil {
-			log.WithError(err).Error("error updating PodVolumeBackup progress")
-		}
+	backupLocation := &velerov1api.BackupStorageLocation{}
+	if err := r.Client.Get(context.Background(), client.ObjectKey{
+		Namespace: pvb.Namespace,
+		Name:      pvb.Spec.BackupStorageLocation,
+	}, backupLocation); err != nil {
+		return nil, errors.Wrap(err, "getting backup storage location")
```
blackpiglet, line 331: "fail to get backup storage location" may be better.

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -18,330 +18,220 @@ package controller
...
 	var emptySnapshot bool
-	if stdout, stderr, err = restic.RunBackup(resticCmd, log, c.updateBackupProgressFunc(req, log)); err != nil {
+	stdout, stderr, err := r.ResticExec.RunBackup(resticCmd, log, r.updateBackupProgressFunc(&pvb, log))
+	if err != nil {
 		if strings.Contains(stderr, "snapshot is empty") {
 			emptySnapshot = true
 		} else {
-			log.WithError(errors.WithStack(err)).Errorf("Error running command=%s, stdout=%s, stderr=%s", resticCmd.String(), stdout, stderr)
-			return c.fail(req, fmt.Sprintf("error running restic backup, stderr=%s: %s", stderr, err.Error()), log)
+			log.WithError(errors.WithStack(err)).Errorf(
```
blackpiglet, line 153: I think this error log can be removed. 
The return error can be logged in controller-manager logger.

blackpiglet (commented):
A common comment is to remove the error log in reconcile function's error handling logic, because a new logger is already added into controller manager to log the return error. You can reference to #4381

codegold79:
Basec on @blackpiglet's findings, I should look and ensure the new context from kubebuilder is being used and not the old one.

pkg/cmd/cli/restic/server.go
codegold79: @blackpiglet, I agree that I should be able to get PodVolumeBackupList. I tried to get the list by updating the following in pkg/controller/pod_volume_backup_controller.go.

```go
func (r *PodVolumeBackupReconciler) getParentSnapshot(pvcUID, backupStorageLocation string) string {
	r.Log = r.Log.WithField("pvcUID", pvcUID)
	r.Log.Infof("Looking for most recent completed pod volume backup for this PVC")

	var pvbList velerov1api.PodVolumeBackupList
	err := r.Client.Get(r.Ctx, client.ObjectKey{}, &pvbList)
      
       [...clipped...]
```

Unfortunately, I encounter a compile error,

```go
cannot use &pvbList (value of type *"github.com/vmware-tanzu/velero/pkg/apis/velero/v1".PodVolumeBackupList) as client.Object value in argument to r.Client.Get: missing method GetAnnotations
```

I thought that the controller tools that Kubebuilder uses was supposed to ensure the PodVolumeBackupList type satisfied the client.Object interface. I verified in pkg/apis/velero/v1/pod_volume_backup_types.go that the correct marker, `//+kubebuilder:object:root=true` was set. 

I also ran `make update` in Velero project root to ensure all the controller tools were run again and new CRDs generated.

Is there something I'm missing? Why can't I use client.Client to get the PodVolumeBackupList without getting a compile error?

@zubron, have you encountered this error before?

pkg/controller/pod_volume_backup_controller.go
codegold79: I'm afraid I don't understand your question, "Can the new version controller still keep the event filter function?"

Did you want me to keep the `podVolumeBackupInformer.Informer().AddEventHandler(...)` function?

pkg/controller/pod_volume_backup_controller.go
codegold79: Yes, I think you're right. When errors are returned, I think they'll be printed in calling functions. They don't need to be logged again.

reasonerjt:
As we are very close to FC of v1.8, moving this out of v1.8 milestone.

pkg/cmd/cli/restic/server.go
codegold79: Currently, we have the following kubebuilder marker on the PodVolumeBackupList type:

```go
// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```

Based on the error wanting client.Ojbect, I replaced it so the client.Object type would be generated:

```go
// +k8s:deepcopy-gen:interfaces=sigs.k8s.io/controller-runtime/pkg/client.Object
```

This also didn't work. I got this error:

```text
cannot use &pvbList (value of type *"github.com/vmware-tanzu/velero/pkg/apis/velero/v1".PodVolumeBackupList) as "sigs.k8s.io/controller-runtime/pkg/client".Object value in argument to r.Client.Get: wrong type for method DeepCopyObject (have func() sigs.k8s.io/controller-runtime/pkg/client.Object, want func() k8s.io/apimachinery/pkg/runtime.Object)
```

@blackpiglet, you said,

> Pod Volume Backup CRD is moved to controller-time. 

However, I don't think PodVolumeBackupList type was moved to controller runtime, so I think we still need to use the old informer.

pkg/controller/pod_volume_backup_controller.go
blackpiglet: Yes. Event handler function filters out the Delete event.
But this may be a minor change. Even Delete event enters reconcile loop, it should not do anything harmful.

pkg/cmd/cli/restic/server.go
zubron: @codegold79 Sorry I missed your mention here!

I don't know if this is the issue, but could you check if using the plural name (`podvolumebackups`) rather than the singular name (`podvolumebackup`) in the RBAC definitions fixes this issue? https://github.com/vmware-tanzu/velero/pull/4436/files#diff-72e6196edabb1cf7d4c410c8fad08976be6213dee85ad249a4ac3e856e440432R134-R135

I'm wondering if the list kind is associated with the plural name in some way :confused: I noticed in the existing controller runtime CRDs, we are using the plural name. https://github.com/vmware-tanzu/velero/blob/175d2967bfe630cc7ad1f1973cb4d8a293867efd/pkg/apis/velero/v1/download_request_types.go#L108-L109

codegold79: Thanks @zubron. I added [a commit](https://github.com/vmware-tanzu/velero/pull/4436/commits/52fabfd03a3705a52fb33f946462376f44c84a37) with the changes you suggested. I ran `make update` but it didn't seem to change any generated files or CRDs. Unless I missed running something, I have a feeling singular and plural work the same.

zubron: @codegold79 Sorry to suggest that! I'm pretty sure you're right, that the singular and plural should work the same.

I just took another look at this and realised that you probably need to use `List` rather than `Get` for fetching the list of PVBs as the list type has [`ListMeta`](https://github.com/vmware-tanzu/velero/pull/4436/files#diff-72e6196edabb1cf7d4c410c8fad08976be6213dee85ad249a4ac3e856e440432R142) rather than `ObjectMeta`. I'm not quite sure how all the plumbing for this works, but I tried using that client to call List rather than Get and it compiled. Maybe that's enough? :woman_shrugging: This stuff is pretty complicated to follow!

codegold79: @zubron, you are a genius! Yes, you're absolutely right about using `r.Client.List` and not `r.Client.Get`. I tried it and so far it works. Thank you so much for looking into this and don't be sorry about the plural/singular thing. It just shows how sharp your eye is for detail. Anyway, I think I'll change it to plural to be consistent with the other controllers.

Ok, one problem down, a couple more to go. I'll work with Dave on those issues (status not updating, credentials file not being found) tomorrow. I'll let you know how it goes.

zubron: Woohoo! So glad that worked 😄 Good luck with the next steps!

codegold79: @blackpiglet, with @zubron's help, I was able to make your suggestion to get rid of the old PVB lister work. Woot!

pkg/controller/pod_volume_backup_controller.go
codegold79: I got rid of patch helper completely and now use the r.Client's Update() method.

pkg/controller/pod_volume_backup_controller.go
codegold79: Errors that are returned are no longer logged now.

codegold79:
Although Ginkgo Basic e2e tests pass (the e2e tests in this PR's checks), when I do a manual test to backup a stateful pod, I get this error:

```go
time="2022-01-29T02:11:45Z" level=error msg="updating PodVolumeBackup resource" backup=velero/j28-c controller=podvolumebackup error="podvolumebackups.velero.io \"j28-c-hkp55\" not found" logSource="pkg/controller/pod_volume_backup_controller.go:104" podvolumebackup=velero/j28-c-hkp55
```

This is not the first time I've seen this problem. I fixed this error in the past by not using patch helper, but rather `r.Client.Update()`. However, using `r.Client.Update()` causes e2e tests to fail. Because `r.Client.Update()` caused the Ginkgo Basic e2e tests to fail, I put up [this commit](https://github.com/vmware-tanzu/velero/pull/4436/commits/a610194aa1c52a889ec5e9c6084bb2ffe4968965).

In one situation (Ginkgo Basic e2e tests), patch helper is able to update PVB status, but unable to update PVB status during manual testing. In another situation (manual backup), `r.Client.Update()` is able to update PVB status, but unable to during Ginkgo Basic e2e tests. This problem has me extremely stumped.

This PR is NOT ready for review due to inability to pass manual e2e testing on a stateful pod.

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -18,429 +18,333 @@ package controller
...
-		return
-	}
-
-	c.metrics.RegisterPodVolumeBackupEnqueue(c.nodeName)
-
-	log.Debug("Enqueueing")
-	c.enqueue(obj)
+// PodVolumeBackupReconciler reconciles a PodVolumeBackup object
+type PodVolumeBackupReconciler struct {
+	Scheme         *runtime.Scheme
```
ywk253100, line 54: What is the `Scheme` used for?

codegold79: I think Kubebuilder uses it (but I don't know what for) because it automatically adds Scheme as part of scaffolding a basic reconciler struct: https://book.kubebuilder.io/cronjob-tutorial/controller-overview.html

ywk253100 (approved):

blackpiglet (approved):

PR #4242, opened 2021-10-13
Create volume snapshots in parallel

jglick:
Experimental patch to run volume snapshotter plugins in parallel, at least for the volume creation. Necessary for https://github.com/vmware-tanzu/velero-plugin-for-aws/pull/90 to be practical since it must wait for an EBS snapshot to be complete before initiating a request to copy it to another region, and this often blocks for about 2m37s (apparently regardless of the size of the delta, or number of snapshots currently pending).

Issue #2888, opened 2020-09-01
Backing up resources in parallel


phuongatemc:
Currently Velero Backup processes resources in serial.  In some scenarios, we would like to back up resources in parallel, not only to increase performance but also help reduce the time gap between backup time of the items.  For example, to backup a Cassandra cluster of 20 pods (each has 1 PVC).  The Backup of such cluster woud take snapshots of PVCs belong to these Pods and to help application consistency, these PVCs should be snapshotted as close to each other as possible (either in parallel or in a single volume group, depending on what storage back-end supported).

So the enhancement request is to allow users to specify the resource types (Kind) to be backed up in parallel.  For example, we can enhance an option say "ConcurrentResources" and users can specify ConcurrentResources: "pods".  Then during backup, we will create goroutine to backup all Pods in parallel.

This feature may conflict with the "OrderedResources" feature which we Backup resources of specific Kind in specific ordered.  So these two "OrderedResources" and "ConcurrentResources" cannot specify the same Kind.

Another aspect can also be considered here is the level of concurrency allowed.  For example if the back end system can only allow up to 10 PVC snapshots being taken in parallel or the backup storage device can allow 10 write streams in parallel then Backup cannot create more backup goroutines than such limit.  This also raise the issue of multiple Backups in parallels and we need to factor in the limitation mentioned above when creating goroutines.

An alternative solution would be VolumeGroup that is currently proposed in the Kubernetes Data Protection Working Group.  This VolumeGroup allows grouping together related PV (so they can be snapshotted together...).

phuongatemc:
We ultimately want all the PVCs belong to the pods of the same application (in the same namespace) being snapshotted in parallel.  However in Velero current implementation, backup item will backup pod and its PVC, PV together before moving to the next pod, we can make it parallel at the level of pod which should be good enough because each pod usually have 1 PVC-PV.

zubron:
This is still needed.

jglick:
> if the back end system can only allow up to 10 PVC snapshots being taken in parallel or the backup storage device can allow 10 write streams in parallel then Backup cannot create more backup goroutines than such limit

Not necessarily; this can be handled simply by having the goroutine wait & retry in a provider-specific manner. https://github.com/jglick/velero-plugin-for-aws/commit/b5d7c526ec7ab806577134454902a1efb076f2cf seems to work in the case of EBS snapshots. Needed when the number of PVs in the backup gets into the range of dozens.

[...]

PR #4444, opened 2021-12-10
testing out if this is possible direction for performance imporvements

shawn-hurley:

Fixes #(issue)


pkg/backup/item_collector.go
```diff
@@ -60,17 +62,49 @@ type kubernetesResource struct {
...
+	resourceChan := make(chan []*kubernetesResource)
 
-		resources = append(resources, groupItems...)
+	type chanError struct {
+		group *metav1.APIResourceList
+		err   error
 	}
+	errorChan := make(chan chanError)
 
+	for _, group := range r.discoveryHelper.Resources() {
```
shawn-hurley, line 76: it might be worth making this put groups onto a channel and making it buffered that way we can control how many threads we spin out.

dsu-igeek: Yeah, that's a good idea.  So you're thinking we'll improve the performance of the retrieve operation by multi-threading?  Definitely need to have a config somewhere that throttles the max number of ops/sec.

shawn-hurley: For sure, this was to point to a diff when I come back and make an enhancement for this. 

I will try and see if we can thread out the item_backupper and only have the tar writer as the blocking bit. This would also need a config for the max number of threads. 

I just want to see if this is possible and, does it gain anything by doing a quick POC before making the enhancement. 

If you want me to close the issue to make things neater, let me know :)

shawn-hurley:
I wanted to add some specifics about the performance improvement with this change. 

The biggest benefit comes when a plugin takes some amount of time.  (in my testing, I added a custom plugin that took 5 seconds). With this change, a backup of 13000 resources took 2hrs, and without this change, it took >13hrs.

If we are going to make Protected Entities or something similar for DataMovers as plugins, this could be something to consider. @dsu-igeek  WDYT?

shawn-hurley:
Going to close this draft PR, and move from POC to an enhancement design and discussion topic

jglick:
Similar to what I prototyped in #4242 as a prerequisite for cross-region backup in AWS (EBS). See #2888.

PR #4479, opened 2022-01-03
Add support for RestoreItemAction plugin for namespaces

### LINKED ISSUE #4478: Support RestoreItemActions for namespaces

We're using Rancher to provision our clusters and it adds annotations on namespaces for their _project_ (the ID and other stuff). A newly provisioned cluster w/ the same _projects_ won't have the same annotation values on namespaces as they are randomly generated. So when we create a new cluster and restore a backup, the restored namespaces aren't in the right project, breaking stuff..

I was looking into making a backup/restore plugin to lookup Rancher's API and update the annotation on namespaces with the new value, but found out that namespaces are not restored as other resources. They don't go through the restore plugin chain, but are just restored from backup.

So I looked into making a multi-step restore process: restore namespaces first, run script to update the annotations then restore everything else, but found that you can't restore only namespaces (see #1970).

Let namespaces go through the restore plugin chain on restore instead of directly from backup.

**Environment:**

**Vote on this issue!**

mouellet:
Namespaces are still lazily restored but using the normal restore process instead of directly from backup.

Fixes #4478

### DIFF
--- a/changelogs/unreleased/4479-mouellet
+++ b/changelogs/unreleased/4479-mouellet
@@ -1,0 +1,1 @@
+Add support for RestoreItemAction plugin for namespaces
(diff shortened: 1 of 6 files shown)


yuvalman:
Hi,
Can you please open this pr and fix the issue?

Issue #4362, opened 2021-11-17
Document AWS, Azure, and GCP plugin changes related to CSI


eleanor-millman:
Documentation for the 1.8 release for the following changes:

https://github.com/vmware-tanzu/velero/issues/4108
https://github.com/vmware-tanzu/velero/issues/4109
https://github.com/vmware-tanzu/velero/issues/4110

Please put docs on this change in the AWS, Azure, and GCP plugin repos.

Please make a PR and tag @a-mccarthy as a reviewer. Thanks!

reasonerjt:
The new versions of plugins for AWS, Azure and GCP will be released with the support of snapshotting and restoring the persistent volumes provisioned by CSI driver via the APIs of the cloud providers.  With this enhancement, users can backup and restore the persistent volumes on these could providers without using the CSI plugin, which will remain beta and the feature flag `EnableCSI` will be disabled by default.

For the version of the plugins and the CSI drivers they support respectively please see the table:
| Plugin | Version | CSI Driver |
| --- | ----------- | ---------- |
| velero-plugin-for-aws | v1.4.0 | ebs.csi.aws.com |
| velero-plugin-for-microsoft-azure | v1.4.0 | disk.csi.azure.com |
| velero-plugin-for-gcp | v1.4.0 | pd.csi.storage.gke.io |

reasonerjt:
@a-mccarthy 
Please see the comment above, I don't find a page on the documentation regarding the plugin of cloud providers.

You may want to add this section in the page to describe the overall support status of CSI in velero.

a-mccarthy:
@reasonerjt is there any change in functionality a user would need to do to make this work? or would they do things as they normally would, only now Velero works through the cloud providers APIs to do the snapshot/restore when you run the `velero` command for a CSI provisioned persistent volume?

a-mccarthy:
@reasonerjt I updated the docs with more information for the CSI updates in #4502, https://velero.io/docs/main/csi/. Please reopen this issue if you feel like there is more to add to the docs. thanks!

PR #4489, opened 2022-01-05
Update CSI docs

### LINKED ISSUE #4363: Document state of CSI and Velero
This could be documentation in the CSI plugin repo or in the main Velero documentation. We need to document how Velero relates to CSI. Specifically noting:
1) How Velero handles volumes provisioned using the CSI driver (probably linking to this documentation - https://github.com/vmware-tanzu/velero/issues/4362)
2) How Velero handles CSI snapshotting (in short, the state of the CSI plugin and how it is not GA).

Also, below is a write up that may be worth editing down to add in (or not).

CSI storage provisioning versus snapshotting and how Velero needs to change now that CSI drivers are becoming more common:

CSI (Container Storage Interface) is an API for both provisioning storage and taking snapshots. The original implementation didn’t have snapshots, they are a newer feature. Originally, storage was provisioned using in-tree drivers (drivers that live in the K8s source tree and they are built as part of the K8s build). There are AWS, GCP, vSphere, Azure drivers, maybe even more. 

There is a CSI API (a RESTful interface). This is defined to be standard across K8s and Docker for allocating volumes. K8s has added a storage driver (a.k.a. CSI driver) that talks to CSI, which is compiled into K8s. And now you can swap in and out different CSi implementations. Now, K8s says “we don’t want your vSphere/AWS/Azure bits in our K8s source code (i.e. built into the distribution)”. This was an unwieldy for two reasons: 1) any driver changes had to come out with a K8s release and 2) any new storage providers had to get into K8s, which was a mess.

The current trend is to push all of the in-tree drivers out of K8s and getting everyone to use the CSI driver. 

For vSphere, there is a huge difference between how the in-tree vSphere driver works and versus using the CSI driver + the vSphere CSI driver (this is the implementation of the CSI API done for vSphere, this is NOT the Velero Plugin for vSphere). Because we want to deprecate the in-tree vSphere driver, we have to make sure that the Velero Plugin for vSphere doesn’t use them.

The in-tree vSphere driver was created about 4 years ago and it doesn’t use the CNS or FCD APIs. Instead, it uses a shadow VM (before FCD, the only way to make volumes on vSphere was to use APIs attached to a VM, so you have to make a fake one). One of the key limitations of the in-tree vSphere driver was that the only way to do a volume snapshot was to snapshot the VM to which the volume was attached (likely a worker node), which would also include every volume attached to the VM. This was inefficient and required lots of cleanup. So, this path was abandoned. But, FCD had snapshot APIs to snapshot a single volume, so this path was taken. 

The other drivers like EBS and Azure didn’t have to do a big architecture shift since they are relatively thin layers over the block storage, so their transition from in-tree to out of tree drivers was relatively easy. 

The confusion: people have heard of CSI snapshots. That is simply the snapshotting mechanism being triggered by K8s by writing Volume Snapshot resource. The External Snapshotter monitors for  Volume Snapshot resources, and the External Snaposhotter talks to the CSI driver asking it to snapshot a disk. And then the CSI implementation has the info necessary to actually take the snapshot. CSI snapshots are relatively new (this whole stack) and just went GA recently (1.20). Prior to 1.20, these were beta and are still not implemented everywhere.

For example, in the vSphere CSI driver, snapshotting is not yet implemented and the Velero Plugin for vSphere team is working on it now, but it will take a couple releases to get it done.

Velero predates all of this (it is about 4 years old). The volume snapshotter plugins talk directly to the storage. So, it takes a different path than the CSI path. They end up at the same point.

Now, most commonly, Velero does the following for volume snapshotter path: Backup CR says to Velero, backup this thing. When we see a PV, we say “do we have a volume snapshot location”, this tells us which volume snapshot plugin to use. That volume snapshot plugin translates the K8s PV into a storage system specific handle, then it calls, say, the AWS SDK, to call snapshot that volume. It passes back the snapshot handle and we package that into the backup.

Note: there are several types of plugins. Volume snapshotter is one, Backup Item Action is another. Backup Item Action is for mutating K8s resources.

Now, less commonly, Velero does the following for CSI plugin path: If the Velero CSI plugin has been installed, the following happens. Backup CR says to Velero, backup this thing. When we hit the PVC, then the CSI BackupItemAction (this is a plugin type) will write a Volume Snapshot CR for that PVC. Note that the K8s CSI Driver (External Snapshotter) works with PVCs, not PVs. So this is why this path involves PVCs.

The Velero Plugin for vSphere was originally implemented as a volume snapshotter, but when Supervisor cluster was introduced, there needed to be a shift to work on the PVC level, so the Velero Plugin for vSphere is now actually a BackupItemAction, note a Volume Snapshotter.

But in the future, Velero should do the following: we want to unify all the paths together (volume snapshotters, CSI BackupItemAction, and the vSphere BackUpItemAction). 

Now to the current 1.8 problem: right now, in the Velero volume snapshotters, there is code that knows how to correctly dive into the PV to extract the volume ID from the storage system specific fields. What we need is for the volume ID in order to take a snapshot. But the CSI driver uses some different fields in that PV. So the existing volume snapshotters look at it and can’t find the volume ID. Once we have that ID, we can continue with the usual path to snapshotting. So, we can support moving forward with PVs created using CSI drivers.

In contrast, we have also been considering bringing the Velero CSI BackupItemAction plugin to GA. This plugin uses CSI to do the snapshotting. But, we don’t necessarily see value in prioritizing this because with the above change, we can still use the existing storage system specific snapshotting paths for any PVs for which we have a Velero Volume Snapshotter. The place where we would have a problem are storage systems that 

Velero came out of the cloud. When Velero takes a snapshot, it expects that the snapshot is durable. That is, it can survive the loss of the storage system. For example, if Velero takes an EBS snapshot, it trusts that EBS, behind the scenes, is storing the snapshot on a different storage system from where the original volume is sitting. EBS has an SLA that if you take a snapshot, you can delete the original volume, you can restore to another place (as long as it is in the same availability zone. 

If we move to CSI snapshotting, then we are great on EBS, on Azure, but if someone says “the reason I want to use CSI snapshotting is that I want to use Tintri”. And perhaps they have hooked up to Velero using the CSI driver with snapshotting that Tintri provided. But how do we know if the CSI-taken snapshot is durable? Velero currently has no data mover capability to deal with it if it isn’t. So for now, we don’t want snapshots taken using a plugin that we don’t own or understand well, because we then can’t guarantee the safety data.

So, the problem with the Velero CSI BackupItemAction Plugin is that it could be calling a driver that takes non-durable snapshots. So we aren’t worried about it when it calls, say, AWS/EBS, but for that, our current alternative path works fine. So we aren’t motivated right now to bring the CSI plugin to GA.

The Velero Plugin for vSphere and the OpenEBS plugins both have data movement under the covers to make sure the snapshots are durable. Right now, all of our VolumeSnapshotters guarantee that the snapshot is durable. If they don’t work, we suggest using restic. Restic has its own problems, but it does do data movement and so when it is configured, the snapshot ends up in a safe space. 

Reasons to move to CSI snapshotting are: 1) reduces the number of code paths and 2) if we have data movement + CSI snapshotting, then we can support snapshotting on random storage systems that support CSI snapshotting. But we aren't there yet since we don’t have the data movement. And the data movement work is planned for 1.8 (moving the data mover from Astrolabe to Velero proper).

a-mccarthy:

This PR updates the CSI docs with more information. Based off content in #4363 and #4362

Fixes #4363

### DIFF
--- a/site/content/docs/main/csi.md
+++ b/site/content/docs/main/csi.md
@@ -9,7 +9,7 @@ By supporting CSI snapshot APIs, Velero can support any volume provider that has
 
 The Velero Container Storage Interface (CSI) plugins are in beta. It's not recommended for use in a production environment and the `EnableCSI` feature flag is disabled by default. The Velero team plans to follow the upstream Kubernetes support level for CSI volume snapshotting, and the Velero CSI plugins will reach general availability sometime after volume snapshotting is GA in upstream Kubernetes.
 
-**Note:** The AWS, Microsoft Azure, and Google Cloud Platform (GCP) Velero plugins version 1.4 and later are able to snapshot and restore persistent volumes provisioned by a CSI driver via the APIs of the cloud provider, without having to install Velero CSI plugins.  
+**Note:** The AWS, Microsoft Azure, and Google Cloud Platform (GCP) Velero plugins version 1.4 and later are able to snapshot and restore persistent volumes provisioned by a CSI driver via the APIs of the cloud provider, without having to install Velero CSI plugins. See the [AWS](https://github.com/vmware-tanzu/velero-plugin-for-aws), [Microsoft Azure](https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure), and [Google Cloud Platform (GCP)](https://github.com/vmware-tanzu/velero-plugin-for-gcp) Velero plugin repo for more information on supported CSI drivers.
 
 ## Prerequisites
 
--- a/site/content/docs/main/csi.md
+++ b/site/content/docs/main/csi.md
@@ -9,7 +9,7 @@ By supporting CSI snapshot APIs, Velero can support any volume provider that has
 
 The Velero Container Storage Interface (CSI) plugins are in beta. It's not recommended for use in a production environment and the `EnableCSI` feature flag is disabled by default. The Velero team plans to follow the upstream Kubernetes support level for CSI volume snapshotting, and the Velero CSI plugins will reach general availability sometime after volume snapshotting is GA in upstream Kubernetes.
 
-**Note:** The AWS, Microsoft Azure, and Google Cloud Platform (GCP) Velero plugins version 1.4 and later are able snapshot and restore persistent volume provisioned by the CSI driver via the APIs of the cloud provider, without having to install Velero CSI plugins.  
+**Note:** The AWS, Microsoft Azure, and Google Cloud Platform (GCP) Velero plugins version 1.4 and later are able to snapshot and restore persistent volumes provisioned by a CSI driver via the APIs of the cloud provider, without having to install Velero CSI plugins.  
 
 ## Prerequisites
 
--- a/site/content/docs/main/csi.md
+++ b/site/content/docs/main/csi.md
@@ -1,26 +1,25 @@
 ---
-title: "Container Storage Interface Snapshot Support in Velero"
+title: "Beta Container Storage Interface Snapshot Support in Velero"
 layout: docs
 ---
 
-_This feature is under development. Documentation may not be up-to-date and features may not work as expected._
+Integrating Container Storage Interface (CSI) snapshot support into Velero enables Velero to backup and restore CSI-backed volumes using the [Kubernetes CSI Snapshot APIs](https://kubernetes.io/docs/concepts/storage/volume-snapshots/).
 
-Integrating Container Storage Interface (CSI) snapshot support into Velero enables Velero to backup and restore CSI-backed volumes using the [Kubernetes CSI Snapshot Beta APIs](https://kubernetes.io/docs/concepts/storage/volume-snapshots/).
+By supporting CSI snapshot APIs, Velero can support any volume provider that has a CSI driver, without requiring a Velero-specific plugin to be available. This page gives an overview of how to add support for CSI snapshots to Velero  through CSI plugins. For more information about specific components, see the [plugin repo](https://github.com/vmware-tanzu/velero-plugin-for-csi/).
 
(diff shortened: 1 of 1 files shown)


site/content/docs/main/csi.md
```diff
@@ -1,26 +1,25 @@
...
- 1. When restoring CSI volumesnapshots across clusters, the name of the CSI driver in the destination cluster is the same as that on the source cluster to ensure cross cluster portability of CSI volumesnapshots
+ 1. Your cluster is Kubernetes version 1.17 or greater.
+ 1. Your cluster is running a CSI driver capable of support volume snapshots at the [v1beta1 API level](https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-cis-volume-snapshot-beta/).
+ 1. When restoring CSI VolumeSnapshots across clusters, the name of the CSI driver in the destination cluster is the same as that on the source cluster to ensure cross cluster portability of CSI VolumeSnapshots
 
 ## Installing Velero with CSI support
 
-Ensure that the Velero server is running with the `EnableCSI` feature flag. See [Enabling Features][1] for more information.
-Also, the Velero [CSI plugin][2] ([Docker Hub][3]) is necessary to integrate with the CSI volume snapshot APIs.
+To integrate Velero with the CSI volume snapshot APIs, you must enable the `EnableCSI` feature flag and install the Velero [CSI plugins][2] on the Velero server.
```
zubron, line 22: Can we still include the link to how to enable server side features here?

zubron: Oh, never mind. I saw it's included in instructions further down :+1:

zubron (dismissed):

zubron (approved):

ywk253100 (approved):

Issue #4066, opened 2021-08-24
Velero should support incremental restores


pradeep288:
**Describe the problem/challenge you have**
Currently, Velero doesn't support incremental  restore.  What i mean by incremental is if I have restored X already and after some time if i want to restore X+ delta X, velero doesn't have support for this.

**Describe the solution you'd like**
During restore, Velero should find difference between X and  X+ DeltaX and should restore only DeltaX.

**Environment:**

- Velero version (use `velero version`): 1.6.1
- Kubernetes version (use `kubectl version`):1.20
- Kubernetes installer & version:1.20
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):MAC

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

reasonerjt:
@pradeep288 thanks for the issue.
I'm relatively new to this project, but I found this piece of code:
https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L1240
It seems velero DOES try to modify the resource even it exists during the restore.

Could you please give us a more concrete example regarding the `incremental restore` what kind of resource is failed to be restored when you tried it and what's in the `delta X` ?

sseago:
@reasonerjt If it already exists, the code below that line determines whether the in-cluster resource is different than the resource in the backup. However, only service accounts are actually patched when differences are found. For everything else, the difference between the cluster version being the same and it being different from the backup just changes the log message. Velero doesn't currently attempt to update resources in the general case. When velero runs a restore and there are non-serviceaccount resources that already exist in the cluster, you'll see in velero logs the message from this line:
https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L1290

pradeep288:
**Backup of X:**

Create a source cluster called A
Create a namespace called pradeep-draas
Create a deployment called example with replicas =3
Install Velero and take the backup(b1) using Velero
Create a target Target Cluster called B
Install velero and verify backup b1 exists.
restore the backup (b1) using velero
[...]

PR #4613, opened 2022-02-01
Add design for enabling support for `ExistingResourcePolicy` to restore API

shubham-pampattiwar:
This PR adds an initial design document to add support for `ExistingResourcePolicy` to Velero's restore API.
Fixes : https://github.com/vmware-tanzu/velero/issues/4842

### DIFF
--- a/design/existing-resource-policy_design.md
+++ b/design/existing-resource-policy_design.md
@@ -1,0 +1,262 @@
+# Add support for `ExistingResourcePolicy` to restore API
+## Abstract
+Velero currently does not support any restore policy on kubernetes resources that are already present in-cluster. Velero skips over the restore of the resource if it already exists in the namespace/cluster irrespective of whether the resource present in the restore is the same or different from the one present on the cluster. It is desired that Velero gives the option to the user to decide whether or not the resource in backup should overwrite the one present in the cluster.
+
+## Background
+As of Today, Velero will skip over the restoration of resources that already exist in the cluster. The current workflow followed by Velero is (Using a `service` that is backed up for example):
+- Velero tries to attempt restore of the `service` 
+- Fetches the `service` from the cluster 
+- If the `service` exists then:
+    - Checks whether the `service` instance in the cluster is equal to the `service`  instance present in backup
+        - If not equal then skips the restore of the `service` and adds a restore warning (except for [ServiceAccount objects](https://github.com/vmware-tanzu/velero/blob/574baeb3c920f97b47985ec3957debdc70bcd5f8/pkg/restore/restore.go#L1246))
+        - If equal then skips the restore of the `service` and mentions that the restore of resource `service` is skipped in logs
+
+It is desired to add the functionality to specify whether or not to overwrite the instance of resource `service` in cluster with the one present in backup during the restore process.
+
+Related issue: https://github.com/vmware-tanzu/velero/issues/4066
+
+## Goals
+- Add support for `ExistingResourcePolicy` to restore API for Kubernetes resources.
+
+## Non Goals
+- Change existing restore workflow for `ServiceAccount` objects
+- Add support for `ExistingResourcePolicy` as `recreate` for Kubernetes resources. (Future scope feature)
+
+## Unrelated Proposals (Completely different functionalities than the one proposed in the design)
+- Add support for `ExistingResourcePolicy` to restore API for Non-Kubernetes resources.
+- Add support for `ExistingResourcePolicy` to restore API for `PersistentVolume` data.
+
+### Use-cases/Scenarios
+
+### A. Production Cluster - Backup Cluster:
+Let's say you have a Backup Cluster which is identical to the Production Cluster. After some operations/usage/time the Production Cluster had changed itself, there might be new deployments, some secrets might have been updated. Now, this means that the Backup cluster will no longer be identical to the Production Cluster. In order to keep the Backup Cluster up to date/identical to the Production Cluster with respect to Kubernetes resources except PV data we would like to use Velero for scheduling new backups which would in turn help us update the Backup Cluster via Velero restore.
+
+Reference: https://github.com/vmware-tanzu/velero/issues/4066#issuecomment-954320686
+
+### B. Help identify resource delta:
+Here delta resources mean the resources restored by a previous backup, but they are no longer in the latest backup. Let's follow a sequence of steps to understand this scenario:
+- Consider there are 2 clusters, Cluster A, which has 3 resources - P1, P2 and P3.
+- Create a Backup1 from Cluster A which has P1, P2 and P3.
+- Perform restore on a new Cluster B using Backup1.
+- Now, Lets say in Cluster A resource P1 gets deleted and resource P2 gets updated.
+- Create a new Backup2 with the new state of Cluster A, keep in mind Backup1 has P1, P2 and P3 while Backup2 has P2' and P3.
+- So the Delta here is (|Cluster B - Backup2|), Delete P1 and Update P2.
+- During Restore time we would want the Restore to help us identify this resource delta.
+
+Reference: https://github.com/vmware-tanzu/velero/pull/4613#issuecomment-1027260446
+
+## High-Level Design
+### Approach 1: Add a new spec field `existingResourcePolicy` to the Restore API
+In this approach we do *not* change existing velero behavior. If the resource to restore in cluster is equal to the one backed up then do nothing following current Velero behavior. For resources that already exist in the cluster that are not equal to the resource in the backup (other than Service Accounts). We add a new optional spec field `existingResourcePolicy` which can have the following values:
+1. `none`: This is the existing behavior, if Velero encounters a resource that already exists in the cluster, we simply
+skip restoration.
+2. `update`: This option would provide the following behavior.
+   - Unchanged resources: Velero would update the backup/restore labels on the unchanged resources, if labels patch fails Velero adds a restore error.
+   - Changed resources: Velero will first try to patch the changed resource, Now if the patch:
+       - succeeds: Then the in-cluster resource gets updated with the labels as well as the resource diff
+       - fails: Velero adds a restore warning and tries to just update the backup/restore labels on the resource, if the labels patch also fails then we add restore error.
+3. `recreate`:  If resource already exists, then Velero will delete it and recreate the resource.
+
+*Note:* The `recreate` option is a non-goal for this enhancement proposal, but it is considered as a future scope.
+Another thing to highlight is that Velero will not be deleting any resources in any of the policy options proposed in 
+this design but Velero will patch the resources in `update` policy option.
+
+Example:
+A. The following Restore will execute the `existingResourcePolicy` restore type `none` for the `services` and `deployments` present in the `velero-protection` namespace.
+
+```
+Kind: Restore
+
+…
+
+includeNamespaces: velero-protection
+includeResources:
+  - services
+  - deployments
+existingResourcePolicy: none
+
+```
+
+B. The following Restore will execute the `existingResourcePolicy` restore type `update` for the `secrets` and `daemonsets` present in the `gdpr-application` namespace.
+```
+Kind: Restore
+
+…
+includeNamespaces: gdpr-application
+includeResources:
+  - secrets
+  - daemonsets
+existingResourcePolicy: update
+```
+
+### Approach 2: Add a new spec field `existingResourcePolicyConfig` to the Restore API
+In this approach we give user the ability to specify which resources are to be included for a particular kind of force update behaviour, essentially a more granular approach where in the user is able to specify a resource:behaviour mapping. It would look like:
+`existingResourcePolicyConfig`:
+- `patch:` 
+    - `includedResources:` [ ]string
+- `recreate:`
+    - `includedResources:` [ ]string
+
+*Note:* 
+- There is no `none` behaviour in this approach as that would conform to the current/default Velero restore behaviour.
+- The `recreate` option is a non-goal for this enhancement proposal, but it is considered as a future scope.
+
+
+Example:
+A. The following Restore will execute the restore type `patch` and apply the `existingResourcePolicyConfig` for `secrets` and `daemonsets` present in the `inventory-app` namespace.
+```
+Kind: Restore
+…
+includeNamespaces: inventory-app
+existingResourcePolicyConfig:
+  patch:
+    includedResources
+     - secrets
+     - daemonsets
+
+```
+
+
+### Approach 3: Combination of Approach 1 and Approach 2
+
+Now, this approach is somewhat a combination of the aforementioned approaches. Here we propose addition of two spec fields to the Restore API - `existingResourceDefaultPolicy` and `existingResourcePolicyOverrides`. As the names suggest ,the idea being that `existingResourceDefaultPolicy` would describe the default velero behaviour for this restore and `existingResourcePolicyOverrides` would override the default policy explicitly for some resources.
+
+Example:
+A. The following Restore will execute the restore type `patch` as the `existingResourceDefaultPolicy` but will override the default policy for `secrets` using the `existingResourcePolicyOverrides` spec as `none`.
+```
+Kind: Restore
+…
+includeNamespaces: inventory-app
+existingResourceDefaultPolicy: patch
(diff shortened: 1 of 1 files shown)


sseago:
This relates to https://github.com/vmware-tanzu/velero/issues/4066

birsanv:
@shubham-pampattiwar I want to include this scenario in your design 

We want to consider here delta resources ( resources restored by a previous backup but no longer in the latest backup ).
We need to be able to identify those resources so that they can be cleaned up.

This is the scenario

The scenario : 

1. Cluster A has Policy1 , Policy2, Policy3
2. Backup policies in backup1  
3. Restore backup1   on a brand new cluster, ClusterB : all policies are created, OK
4. NOW : on Cluster A delete Policy1 and update Policy2
5. on Cluster A create a second backup backup2 ( backup2 contains only Policy2, Policy3 )
6. RESTORE backup2 on Cluster B; with the fix we are discussing now :
- Policy1 still on ClusterB, untouched;  annotated with velero.io/backup-name=backup1 
-  Policy2 gets updated ; annotated with velero.io/backup-name=backup2
- Policy3 is not updated, untouched ;  annotated with velero.io/backup-name=backup1 


Result:

Cluster B is in an inconsistent state ; Policy1 should no longer be running on this cluster.
There is no differentiation between these policies to tell that :  
- Policy1 was not part of the latest restore ( is gone, should be removed )
- Policy3 was part of the restore but nothing changed ( OK, should be used )
 

Proposed solution :

When restore runs and finds a resource in backup which is identical with the one one the hub, it should still update the backup annotation to : velero.io/backup-name=backup2
In this way, we can identify restored resources not touched by the new restore because they were not on the backup anymore ; vs the ones not touched by the current restore because they were not changed
I can find what resources should be removed from the cluster (Policy1) by running : get resource with velero.io/backup-name AND velero.io/backup-name != backup2

And if we can identify restored delta, we can remove them after restore is completed. That can be done in  a separate enh

sseago:
@birsanv So it sounds like for this use case we don't want the restore to delete things that aren't in the backup -- we just want the restore to update annotations on items that would have been patched but aren't because they haven't changed -- am I understanding that correctly?

As for where it fits in the design, assuming the above is correct, I would think we wouldn't do this in the "none" option (i.e. preserving legacy behavior -- also it doesn't make sense to update the annotation on unchanged resources when we're simply warning on changed resources. Also, we may not want to *always* do this when updating, as it's extra work which may not matter for some use cases. As such, I'd propose adding a new option to the policy list:
- none: info log for unchanged, warn on changed (same as proposed earlier)
- update: info log for unchanged, patch on changed, warn if patch failed (same as proposed earlier)
- updateAll: info log *and* update annotation on unchanged, patch on changed, warn if patch failed (this is the new option).

We should also call out that any removal of dangling resources from old backups is out-of-scope for velero -- this would be done (possibly manually) by the user after restore using a delete operation with label selector or something similar.

birsanv:
@sseago your proposal makes sense
As long as we have an option to allow update annotation on unchanged ( updateAll in the above ) then that should cover the delta scenario I mentioned
Agreed that removing of dangling resources can be done outside of velero, as long as we have a way to identify them

design/existing-resource-policy_design.md
```diff
@@ -0,0 +1,247 @@
...
+## Non Goals
+- Add support for `ExistingResourcePolicy` to restore API for Non-Kubernetes resources.
+- Add support for `ExistingResourcePolicy` to restore API for `PersistentVolume` data.
+- Change existing restore workflow for `ServiceAccount` objects
+- Add support for `ExistingResourcePolicy` as `recreate` for Kubernetes resources. (Future scope feature)
+
+### Use-cases/Scenarios
+
+### A. Production Cluster - Backup Cluster:
+Let's say you have a Backup Cluster which is identical to the Production Cluster. After some operations/usage/time the Production Cluster had changed itself, there might be new deployments, some secrets might have been updated. Now, this means that the Backup cluster will no longer be identical to the Production Cluster. In order to keep the Backup Cluster upto date/identical to the Production Cluster with respect to Kubernetes resources except PV data we would like to use Velero for scheduling new backups which would in turn help us update the Backup Cluster via Velero restore.
```
sseago, line 30: Nit: Codespell found a minor typo here. "upto" -> "up to"

sseago (commented):
Looks good overall -- minor typo fix identified by Codespell github action, though.

sseago (dismissed):

design/existing-resource-policy_design.md
```diff
@@ -0,0 +1,247 @@
...
+Reference: https://github.com/vmware-tanzu/velero/pull/4613#issuecomment-1027260446
+
+## High-Level Design
+### Approach 1: Add a new spec field `existingResourcePolicy` to the Restore API
+In this approach we do *not* change existing velero behavior. If the resource to restore in cluster is equal to the one backed up then do nothing following current Velero behavior. For resources that already exist in the cluster that are not equal to the resource in the backup (other than Service Accounts). We add a new optional spec field `existingResourcePolicy` which can have the following values:
+1. `none`: This is the existing behavior, if Velero encounters a resource that already exists in the cluster, we simply
+skip restoration.
+2. `update` or `merge` or `patch`: In this behaviour Velero will try an attempt to patch the resource with the backed up version, 
+if the patch fails log it as a warning and continue the restore process, user may wish specify the merge type.
+3. `updateAll`: Similar to `update` option but Velero will update labels on unchanged resources too.
```
kaovilai, line 53: Thoughts?
```suggestion
3. `updateOverrideLabels`: Similar to `update` option but Velero will update labels on unchanged resources too.
```

kaovilai:
```suggestion
3. `updateForceUnchangedLabels`: Similar to `update` option but Velero will update labels on unchanged resources too.
```

sseago: It's not so much "unchanged" labels we're forcing, it's the labels that *velero* adds on restore -- these labels aren't in the backup. `updateForceVeleroLabels` is more accurate, but I think it's a bit clumsy as a config option since it's so long. Perhaps just `updateVeleroLabels` -- that's less clumsy, but it might confuse users into thinking we're *only* updating velero labels. I'm thinking that `updateAll` may be best, as long as the documentation is clear on what this means: "Update all resources, including unchanged ones, in order to ensure that the velero labels related to the backup and restore are appropriately updated."

kaovilai: When was the last time verbosity caused problems? It accelerated java adoption. You'd most likely type it once in Jason/yaml or velero cli autocompletes or also have an accompanying shorthand flag.

kaovilai: I prefer to cater flags to "I haven't read the doc" folks more than the "I type this out everyday so it has to be short".. since support escalations come from the former.

sseago: `updateIncludingVeleroLabelsForAll` would probably be the most accurate. Or perhaps `updateAndForceVeleroLabels`. I'd still vote for `updateAll`, but if we want something longer, one of these would be clearer than the others I think.

design/existing-resource-policy_design.md
```diff
@@ -0,0 +1,247 @@
...
+
+Related issue: https://github.com/vmware-tanzu/velero/issues/4066
+
+## Goals
+- Add support for `ExistingResourcePolicy` to restore API for Kubernetes resources.
+
+## Non Goals
+- Add support for `ExistingResourcePolicy` to restore API for Non-Kubernetes resources.
+- Add support for `ExistingResourcePolicy` to restore API for `PersistentVolume` data.
+- Change existing restore workflow for `ServiceAccount` objects
```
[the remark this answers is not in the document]
sseago: I wonder whether it makes sense to distinguish "non-goals" that are desired bit out-of-scope for now and "non-goals" that are considered not part of this at all (i.e. they're unrelated things that would be dealt with in unrelated proposals_. For non-kube and PV data, I'd put them in the category of "this is completely different functionality". For service accounts, it's similar -- basically service accounts are a separate use case that is hard-coded with type-specific functionality, so we completely ignore them so as not to break that type-specific outcome.

design/existing-resource-policy_design.md
```diff
@@ -0,0 +1,247 @@
...
+4. `recreate`:  If resource already exists, then Velero will delete it and recreate the resource.
+
+In summary,
+
+| Policy      | Action performed by Velero                                                              | Velero Logging                                                |
+|-------------|-----------------------------------------------------------------------------------------|---------------------------------------------------------------|
+| `none`      | Existing Velero behavior.                                                               | `info` log for unchanged resource, `warn` on changed resource |
+| `update`    | Velero tries to attempt a `patch` on changed resource                                   | `info` log for unchanged, `warn` if patch failed              |
+| `updateAll` | Velero updates labels on unchanged resource  and attempts a `patch` on changed resource | `info` log for unchanged and `warn` if patch failed           |
+
```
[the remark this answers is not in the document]
sseago: I don't think we are able to detect the "to-be-deleted" during restore, since we won't touch them at all. We're only processing resources that are in the current backup that's being restored. The previously-restored resources that are not in the current backup become "to-be-deleted" by default. These resources don't actually get considered/processed by the restore operation at all, so we are not in a position to log anything about them.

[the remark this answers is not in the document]
Lyndon-Li: As far as my understanding, we want users to decide which resource is out-of-date thus should be deleted. So how do users tell this? - users compare the version labels we put along with the resource during restore. Eventually, it is our label that causes deletion of a resource. If my above suggestion doesn't work, we should find a way to cope with this problem.

For "patch labels", personally, I think we should consider it case by case, if the result of the patching is critical, we definitely need to fail the restore; otherwise, we can let it go with a warning.

design/existing-resource-policy_design.md
```diff
@@ -0,0 +1,247 @@
...
+existingResourceDefaultPolicy: patch
+existingResourcePolicyOverrides:
+  none:
+    includedResources
+     - secrets
+
+```
+
+## Detailed Design
+### Approach 1: Add a new spec field `existingResourcePolicy` to the Restore API
```
reasonerjt, line 143: What's your preference in terms of approach 1 vs 2 vs 3?
Approach 1 is probably good enough?  Are we good not to make it any more complicated?  
Wanna point out it's not flexible for introducing more fine-grained control, but I'm totally fine with that.

design/existing-resource-policy_design.md
```diff
@@ -0,0 +1,247 @@
...
+  - daemonsets
+existingResourcePolicy: update
+```
+
+### Approach 2: Add a new spec field `existingResourcePolicyConfig` to the Restore API
+In this approach we give user the ability to specify which resources are to be included for a particular kind of force update behaviour, essentially a more granular approach where in the user is able to specify a resource:behaviour mapping. It would look like:
+`existingResourcePolicyConfig`:
+- `patch:` 
+    - `includedResources:` [ ]string
+- `recreate:` [ ]string
```
reasonerjt, line 101: what's the `[]string` for? typo?

design/existing-resource-policy_design.md
```diff
@@ -0,0 +1,247 @@
...
+    - `includedResources:` [ ]string
```
reasonerjt, line 102: What if the values of `includeResources` of different actions have overlap?

reasonerjt:
A more general question, have we discussed whether this option should be exposed via CLI?  I think the answer is yes, and this design should add one more section to describe how user should set the policy via CLI and what change in the code should be made?

design/existing-resource-policy_design.md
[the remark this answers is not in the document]
sseago: So for unchanged resources, if we're patching labels for the purposes of allowing users to delete things (i.e. the "updateAll" functionality), we may want to make it an error since, as you point out, otherwise the unlabeled resource will look like it's out-of-date and ready to be deleted. For modified resources, it's trickier. If it's not too much trouble in the implementation, I'd propose the following:

For modified resources, if the full patch fails and the policy is "update", just warn and move on.
For modified resources, if the full patch fails and the policy is "updateAll", we have a couple options:
1) error on path fail just as for unmodified resources
2) attempt to patch just the label and if this succeeds, add the usual "couldn't modify resource" warning, and if label patching fails, then add the "label update failed" error *and* the "couldn't update resource" warning.

design/existing-resource-policy_design.md
sseago: My thinking is that approach may be good enough for a first pass. It leaves the door open to expanding in the future to approach 3. So for now, we just make one policy for the whole restore. For the future, we could add the policy override by resource type.

design/existing-resource-policy_design.md
sseago: We would need to be explicit in documenting this, but I think "first match" is good enough. Ideally there would be no overlap here, but if there is, take the first one in the list.

sseago:
Regarding the CLI, I think "approach 1" should be relatively easy to add to the CLI -- and yes, that should probably be added to the design. I think the more complex configuration of approach 3 (the resource-specific overrides) may be OK to leave out of the cli. If the user needs complex policy like that, they need to create the CR directly, but if a single per-restore policy is sufficient, the CLI should have a flag for that.

design/existing-resource-policy_design.md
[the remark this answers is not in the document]
shubham-pampattiwar: @Lyndon-Li @sseago I will add non-kube and PV data part in a new section titled as `Unrelated Proposals`

design/existing-resource-policy_design.md
shubham-pampattiwar: @reasonerjt @sseago I vote for approach 1 as well. Easy to implement and scale.

design/existing-resource-policy_design.md
shubham-pampattiwar: yes, its a typo, thanks for the catch.

design/existing-resource-policy_design.md
kaovilai: Or error out so user fix it in the next run?

reasonerjt:
I would vote for apporach 1 and my only concern is the confusing difference between `update` and `updateAll`
https://github.com/vmware-tanzu/velero/pull/4613/files#r816560921

shubham-pampattiwar:
@reasonerjt Lets say we dump the `updateAll` option altogether. We keep only 2 policy options:
1. `none`: This option preserves the existing velero restore workflow.
2. `update`: This option would provide the following behavior
     -  Unchanged resources:  Velero would update the backup/restore labels on the unchanged resources, if labels patch fails Velero adds a restore error.
     -  Changed resources: Velero will first try to patch the changed resource, Now if the patch:
           -  succeeds: Then the in-cluster resource gets updated with the labels as well as the resource diff
           - fails: Velero adds a restore warning and tries to just update the backup/restore labels on the resource, if the labels patch also fails then we add restore error.

Does the above sound reasonable ?

design/existing-resource-policy_design.md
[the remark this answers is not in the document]
shubham-pampattiwar: Unrelated Proposals section added to the design document https://github.com/vmware-tanzu/velero/pull/4613/files#diff-472925e9329f9cf29ef5d1c15a734e96587e5e975d67b7b7fec53936a280c28dR25

shubham-pampattiwar:
> A more general question, have we discussed whether this option should be exposed via CLI? I think the answer is yes, and this design should add one more section to describe how user should set the policy via CLI and what change in the code should be made?

Added details for CLI changes https://github.com/vmware-tanzu/velero/pull/4613/files#diff-472925e9329f9cf29ef5d1c15a734e96587e5e975d67b7b7fec53936a280c28dR251

sseago:
@shubham-pampattiwar Yes, the proposed change to update/updateAll seems reasonable to me.

shubham-pampattiwar:
I have updated the design proposal with merging the `update` and `updateAll` policy option (dumping the `updateAll` option altogether). Going ahead with the implementation of approach 1 as discussed in the community call. Will work on feature documentation as well.

shubham-pampattiwar:
@reasonerjt @sseago I have updated the PR with `Implementation decision` section.

sseago (approved):

reasonerjt (approved):

PR #4619, opened 2022-02-04
Add design for enabling multiple label support

shubham-pampattiwar:
This PR adds an initial design for Enable support for backing up resources based on multiple labels (OR Logic) in a single backup config.

Related Issue: https://github.com/vmware-tanzu/velero/issues/4843

### DIFF
--- a/design/multiple-label-selectors_design.md
+++ b/design/multiple-label-selectors_design.md
@@ -1,0 +1,138 @@
+# Ensure support for backing up resources based on multiple labels
+## Abstract
+As of today Velero supports filtering of resources based on single label selector per backup. It is desired that Velero
+support backing up of resources based on multiple labels (OR logic).
+
+**Note:** This solution is required because kubernetes label selectors only allow AND logic of labels.
+
+## Background
+Currently, Velero's Backup/Restore API has a spec field `LabelSelector` which helps in filtering of resources based on
+a **single** label value per backup/restore request. For instance, if the user specifies the `Backup.Spec.LabelSelector` as
+`data-protection-app: true`, Velero will grab all the resources that possess this label and perform the backup
+operation on them. The `LabelSelector` field does not accept more than one labels, and thus if the user want to take
+backup for resources consisting of a label from a set of labels (label1 OR label2 OR label3) then the user needs to
+create multiple backups per label rule. It would be really useful if Velero Backup API could respect a set of
+labels (OR Rule) for a single backup request.
+
+Related Issue: https://github.com/vmware-tanzu/velero/issues/1508
+
+## Goals
+- Enable support for backing up resources based on multiple labels (OR Logic) in a single backup config.
+- Enable support for restoring resources based on multiple labels (OR Logic) in a single restore config.
+
+## Use Case/Scenario
+Let's say as a Velero user you want to take a backup of secrets, but all these secrets do not have one single consistent
+label on them. We want to take backup of secrets having any one label in `app=gdpr`, `app=wpa` and `app=ccpa`. Here
+we would have to create 3 instances of backup for each label rule. This can become cumbersome at scale.
+
+## High-Level Design
+### Addition of `OrLabelSelectors` spec to Velero Backup/Restore API
+For Velero to back up resources if they consist of any one label from a set of labels, we would like to add a new spec
+field `OrLabelSelectors` which would enable user to specify them. The Velero backup would somewhat look like:
+
+```
+apiVersion: velero.io/v1
+kind: Backup
+metadata:
+  name: backup-101
+  namespace: openshift-adp
+spec:
+  includedNamespaces:
+  - test
+  storageLocation: velero-sample-1
+  ttl: 720h0m0s
+  orLabelSelectors:
+  - matchLabels:
+      app=gdpr
+  - matchLabels:
+      app=wpa
+  - matchLabels:
+      app=ccpa
+```
+
+**Note:** This approach will **not** be changing any current behavior related to Backup API spec `LabelSelector`. Rather we
+propose that the label in `LabelSelector` spec and labels in `OrLabelSelectors` should be treated as different Velero functionalities.
+Both these fields will be treated as separate Velero Backup API specs. If `LabelSelector` (singular) is present then just match that label.
+And if `OrLabelSelectors` is present then match to any label in the set specified by the user. For backup case, if both the `LabelSelector` and `OrLabelSelectors` 
+are specified (we do not anticipate this as a real world use-case) then the `OrLabelSelectors` will take precedence, `LabelSelector` will
+only be used to filter only when `OrLabelSelectors` is not specified by the user. This helps to keep both spec behaviour independent and not confuse the users. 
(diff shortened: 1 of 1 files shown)


design/multiple-backup-selectors_design.md
```diff
@@ -0,0 +1,70 @@
...
+
+`Velero create backup multi-label-example --selectors 'app=gdpr', 'app=wpa', 'app=ccpa'`
+
+In the above command (`velero --help` will display),
+`--selectors is LabelSelectors  Only backup resources matching any one label from the set of label selectors`
+
+**Note:** This approach will **not** be changing any current behavior related to Backup API spec `LabelSelector`. Rather we
+propose that the label in `LabelSelector` spec and labels in `LabelSelectors` should be concatenated  and then used as a
+criteria(resources matching any one label) to grab the resources to be backed up. For instance, if the backup command
+looks like:
```
sseago, line 40: Hmm. I'm not sure that concatenating is the right approach here, since that changes the meaning of the current (singular) label selector field. I'm wondering whether it might be cleaner (and easier to implement) to just treat them as separate fields. If the singular field is present, then a match requires that label. If the plural field is present, then a match requires one of the entries to match. If both are present, then it's effectively an *and* between them (but presumably most real-world cases won't use both).

sseago: @kaovilai "introduce another field" -- meaning in addition to the proposed new `LabelSelectors`? i.e. we have `LabelSelector`, `OrLabelSelectors` and `AndLabelSelectors`? If there's a need for that, it could be considered as well -- but this makes my suggestion of "Treat the fields as separate, don't concatenate them" even more important. Is there a need for the ability of including *and* behavior for 3 or more label selectors?

kaovilai: Possibly.

Here's one
Backup: east database backup
app=database
region=east

dymurray: Yeah that's a good point. I'm in favor of what Scott is suggesting here. Thoughts @shubham-pampattiwar ?

shubham-pampattiwar: @dymurray yeah I agree, +1 to what @sseago is suggesting. I will update the design proposal with keeping the `labelSelector` and `labelSelectors` as separate fields.

sseago (approved):

reasonerjt:
To be honest, I find supporting both `selector` and `selectors` a bit confusing.  Can we just keep one selector but make it support complicated matching rules?  I haven't checked in details but I think it's do-able as it wraps the LabelSelector in k8s library?

shubham-pampattiwar:
> To be honest, I find supporting both `selector` and `selectors` a bit confusing. Can we just keep one selector but make it support complicated matching rules? I haven't checked in details but I think it's do-able as it wraps the LabelSelector in k8s library?

@reasonerjt k8s LabelSelector does not support OR rule, thus we want to introduce this option for the users. We have 2 options to go ahead for the solution:
1. Separate support for both LabelSelector/selector/singular and LabelSelecters/selectors/plural, this design PR caters to this option, we are preserving the existing(singular) velero behavior and adding a new plural option.
2. Concat both the singular and plural options and depreciate the existing velero LabelSelector (singular) option.
What are your thoughts on the above options ?

reasonerjt:
@shubham-pampattiwar 

Can we extend the current selector? 
https://github.com/kubernetes/apimachinery/blob/v0.22.2/pkg/apis/meta/v1/helpers.go#L105-L149

According to this chunk, it should support `in` semantic.

blackpiglet:
@reasonerjt 
Basically, I agree with the proposal. 
Existing "--selector" support multiple label values by comma separator, but they are supposed to be logical "AND": https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
The logical "OR" is not supported.

Kubernetes set-based label selector supports "in" as operator, but that is used for a specific label's value options. That is different from the proposal design.
For example:  "partition in (customerA, customerB),environment=qa" 
This selector means choosing resources that have label "environment" value as "qa" and "partition" label's value to "customerA" or "customerB".

shubham-pampattiwar:
@blackpiglet Thank you for the example and explanation. 
@reasonerjt Do you agree with the design proposal in the PR ?

reasonerjt:
Sorry for the misunderstanding, and I agree that the resource selector   `k1=v1` OR `k2=v2` is a valid requirement for some users.
My concern is mainly around usability, or, is it just me finding the situation to support both `selector` and `selectors` params in CLI , and both `labelSelector` and `labelSelectors` fields in the API confusing?

shubham-pampattiwar:
> Sorry for the misunderstanding, and I agree that the resource selector `k1=v1` OR `k2=v2` is a valid requirement for some users. My concern is mainly around usability, or, is it just me finding the situation to support both `selector` and `selectors` params in CLI , and both `labelSelector` and `labelSelectors` fields in the API confusing?

@reasonerjt I agree with your concern but we think this might be the best solution available. We can do one thing though, we will not expose this feature via the velero CLI and just keep this feature available via the API, so that we only cater to the advanced Velero users. Does this sound like a good idea ?

shubham-pampattiwar:
Removed the Velero CLI bits, this featuer will only be exposed via the Velero API.
Implementation PR: https://github.com/vmware-tanzu/velero/pull/4650

design/multiple-label-selectors_design.md
```diff
@@ -0,0 +1,138 @@
...
+}
+```
+
+The logic to collect resources to be backed up for a particular backup will be updated in the `backup/item_collector.go`
+around [here](https://github.com/vmware-tanzu/velero/blob/574baeb3c920f97b47985ec3957debdc70bcd5f8/pkg/backup/item_collector.go#L294).
+
+And for filtering the resources to be restored, the changes will go [here](https://github.com/vmware-tanzu/velero/blob/d1063bda7e513150fd9ae09c3c3c8b1115cb1965/pkg/restore/restore.go#L1769)
+
+**Note:** 
+- Naming suggestions are welcome for the `LabelSelectors` spec field.
```
reasonerjt, line 137: Sorry for the late suggestion, but what do you think if we call it `LabelSelectorAny`

shawn-hurley: that makes me personally think of label selector any from the core library which means ALL labels match. 

`AllOfLabelSelectors` or `AnyOfLabelSelectors` or `AllLabelSelectors`?

blackpiglet: I root for OrLabelSelectors which is mentioned by Scott previously.

reasonerjt: `OrLabelSelectors` also looks good to me!

shubham-pampattiwar: I agree with `OrLabelSelectors`

blackpiglet (approved):

shubham-pampattiwar:
Updated PR with `OrLabelSelectors` spec name

reasonerjt (approved):

blackpiglet (approved):

Issue #1508, opened 2019-05-20
Support filtering by multiple labels/resources ("OR" logic)


tillepille:
**Describe the solution you'd like**

Currently 
```
velero create backup mulit -l 'backup=foo' -l 'backup=bar'

or

velero create backup multi --include-resources 'secrets' --include-resources 'configmaps'
```
means, that the flag gets overwritten.  
So only configmaps / things with the label `backup=bar` will be backed up.

It would be great if i could specify multiple resource types or label combinations. 

 
**Anything else you would like to add:**
I know that i can specify multiple labels in one string like `-l 'backup=foo,shouldbackup=true'`
but this is an *AND* not an *OR* 

In my usecase its not possible to add one label to all the things i want to include into the backup.

**Environment:**

- Velero version (use `velero version`): v1.0.0
- Kubernetes version (use `kubectl version`):v1.14.2
- Kubernetes installer & version:v1.14.0
- Cloud provider or hardware configuration: hetzner (backup into aws s3)
- OS (e.g. from `/etc/os-release`): Client debian:stretch-slim (container) / Server Ubuntu:18.04

skriss:
@tillepille thanks for reporting the issue, this is a known limitation.  If it's not possible to specify in one set of flags, for now your best bet is to set up multiple backups, each with their own filter.

tillepille:
@skriss Exactly, multiple backups are my current setup, but this kind of eliminates the benefits of using velero.

Do you have a starting point in the point for me? 
I'm probably to bad at Go but i would like to see if i can do something, or at least better understand the limitation.

Thanks for your fast reply!

birsanv:
I am really happy to see the request being reopened @reasonerjt , I browsed this issue a few days ago and was disappointed to see this thread was closed.

I am also trying to find a solution to use multiple label selectors to group resources in one backup file. I spent some time to figure out if there is any workable option other then creating separate backups for each labelSelector rule.

In my scenario I want to backup all secrets with a label selector labels in a subset. 
The only option I could get to was to have one label set on all these secrets. I started on this route but is not as easy as it looks and I am not even sure if it's feasible in reality.

These secrets are created by various teams, some external, and is extremely difficult to create a consensus among all groups to use one label. Also, we would have to take care of existing secret resources when defining this one label solution; unless they are updated to add the new label we agree upon using, the backup will miss existing resources.

Is there a plan to address this issue in velero ?..

I have in mind a simple solution :

[...]

PR #4650, opened 2022-02-15
Add multiple label selector support to Velero Backup and Restore APIs

shubham-pampattiwar:
Signed-off-by: Shubham Pampattiwar <shubhampampattiwar7@gmail.com>

The PR follows the approach specified in the [multiple label support design ](https://github.com/vmware-tanzu/velero/pull/4619)and does the following:

- adds a new spec to the Velero backup and Velero restore APIs called `labelSelectors` (plural)
- updates the respective Velero backup and Velero restore CRDs with the addition of `labelSelectors` (plural) spec option
- updates the logic in velero backup and restore workflow to respect the `labelSelectors` (plural) option when specified via API

Fixes https://github.com/vmware-tanzu/velero/issues/1508


pkg/restore/restore.go
```diff
@@ -1813,7 +1826,26 @@ func (ctx *restoreContext) getSelectedRestoreableItems(resource, targetNamespace
...
+			}
+
+			if len(ctx.selectors) == skip && skip > 0 {
+				ctx.log.Infof("setting skip flag to true for item: %s", item)
+				skipItem = true
+			}
+		}
+
+		if skipItem {
+			ctx.log.Infof("restore selector labels did not match, skipping restore of item: %s", skipItem, item)
```
reasonerjt, line 1844: In this case, the `selector` is ignored.  Is it the design that when en(ctx.selectors) > 0 the ctx.selector is ignored?

pkg/restore/restore.go
```diff
@@ -305,6 +317,7 @@ type restoreContext struct {
 	namespaceIncludesExcludes  *collections.IncludesExcludes
 	chosenGrpVersToRestore     map[string]ChosenGroupVersion
 	selector                   labels.Selector
+	selectors                  []labels.Selector
```
reasonerjt, line 320: Could you consider renaming the attribute `selectors` here and extract the logic to filter based on both `selector` and `selectors` into one func, and write some UT?

pkg/backup/item_collector.go
```diff
@@ -296,50 +296,70 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
...
+		for _, label := range labelSelectors {
+			unstructuredList, err := resourceClient.List(metav1.ListOptions{LabelSelector: label})
 			if err != nil {
 				log.WithError(errors.WithStack(err)).Error("Error listing items")
 				continue
 			}
 			unstructuredItems = append(unstructuredItems, unstructuredList.Items...)
 		}
 
+		// use labelSelector (singular) behavior only when the user has not specified labelSelectors(plural) in backup API
```
reasonerjt, line 319: if you want to apply the labelSelector (singular) after applying labelSelectors (plural), please move the assignment of  labelSelector (singular) closer to the code where it's used.

pkg/backup/item_collector.go
```diff
@@ -296,50 +296,70 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
...
+		if len(labelSelectors) == 0 {
```
reasonerjt, line 320: here when len(labelSelectors) >0, the labelSelector (singular) is ignored, is it the design?

pkg/backup/item_collector.go
```diff
@@ -296,50 +296,70 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
...
+			if r.pageSize > 0 {
```
reasonerjt, line 321: I think it's problematic that the `pageSize` is ignored when len(labelSelectors) >0, though I don't have a quick idea how to fix it...

reasonerjt:
In addition to my inline comments, I wish to suggest finding a better naming for the fields in the CRD and variable names in the code.

Currently, after the PR is merged, we'll have co-existence of `labelSelector` and `labelSelectors` in the API, `selector`and  `selectors` in the same struct and they all have different meanings and usage, which is not quite maintainable.

pkg/backup/item_collector.go
sseago: Hmm. I may be misremembering, but I thought that if both single and plural were supplied, both would have to match.

pkg/restore/restore.go
shubham-pampattiwar: Done, renamed to `orLabelSelectors`

pkg/backup/item_collector.go
shubham-pampattiwar: Done, moved the code block

pkg/restore/restore.go
shubham-pampattiwar: Yes, we would want to process the `orLabelSelectors` first and based on that decide whether to skip the restore of the item or not. It would not make sense if the item was to be skipped based on `orLabelSelectors` and was to be restored based on `LabelSelector`. This will also preserve the existing velero behavior where the `orLabelSelectors` do not exist.

pkg/backup/item_collector.go
shubham-pampattiwar: Yes @reasonerjt @sseago it would be by design. `OrLabelSelectors` should take precedence, because:
- first of all specifying `LabelSelector` and `orLabelSelectors` would not be a real world use-case
- secondly, the AND of labels is already sufficed by `LabelSelector` and it would not make sense to take another AND of `LabelSelector` and `OrLabelSelectors`. I believe the AND behavior would definitely confuse the user
- Finally, this preserves the current velero behavior where the `OrLabelSelectors` does not exist, processing based on the `LabelSelector` only when the `OrLabelSelectors` is empty helps reinforce this.

sseago: @shubham-pampattiwar So yes, this is an edge case that shouldn't happen under normal circumstances, but ignoring `LabelSelector` when the other is present is inconsistent with existing behavior. Currently, if a field is present, its requirements must be met. If we don't intend to enforce the requirements of both when both are present, then instead of ignoring `LabelSelector` we should raise an error if both are set. Otherwise,  a user might expect the field to be enforced.

shubham-pampattiwar: Thanks for the suggestions guys. I have kept the processing of `LabelSelectors` and `OrLabeleSelectors` separate from each other. I have added validations based on the existence of both the spec, if the user specifies both `LabelSelectors` and `OrLabelSelectors` the backup/restore will have an error and a `FailedValidation` phase, indicating to the user that both spec fields cannot be used at the same time.

sseago (dismissed):

pkg/controller/backup_controller.go
```diff
@@ -451,6 +451,11 @@ func (c *backupController) prepareBackupRequest(backup *velerov1api.Backup) *pkg
 		request.Status.ValidationErrors = append(request.Status.ValidationErrors, fmt.Sprintf("Invalid included/excluded namespace lists: %v", err))
 	}
 
+	// validate that only one exists orLabelSelector or just labelSelector (singular)
+	if request.Spec.OrLabelSelectors != nil && request.Spec.LabelSelector != nil {
```
reasonerjt, line 455: Could you add testcase for this?

pkg/controller/restore_controller.go
```diff
@@ -326,6 +326,11 @@ func (c *restoreController) validateAndComplete(restore *api.Restore, pluginMana
 		restore.Status.ValidationErrors = append(restore.Status.ValidationErrors, fmt.Sprintf("Invalid included/excluded namespace lists: %v", err))
 	}
 
+	// validate that only one exists orLabelSelector or just labelSelector (singular)
```
reasonerjt, line 329: Could you add testcase for this?

reasonerjt:
Please also update the doc in 
https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/main/api-types/backup.md
https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/main/api-types/restore.md

config/crd/v1/bases/velero.io_backups.yaml
```diff
@@ -314,6 +314,60 @@ spec:
                       type: string
                     type: object
                 type: object
+              orLabelSelectors:
+                description: OrLabelSelectors is list of metav1.LabelSelector to filter
+                  with when adding individual objects to the backup. If empty or nil,
```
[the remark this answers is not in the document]
shubham-pampattiwar: Done, updated the spec description.

config/crd/v1/bases/velero.io_restores.yaml
```diff
@@ -1703,6 +1703,60 @@ spec:
                   included in the map will be restored into namespaces of the same
                   name.
                 type: object
+              orLabelSelectors:
+                description: OrLabelSelectors is list of metav1.LabelSelector to filter
+                  with when restoring individual objects from the backup. If empty
+                  or nil, all objects are included. Optional. If multiple provided
```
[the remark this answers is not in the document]
shubham-pampattiwar: Done, updated the spec description.

pkg/apis/velero/v1/backup.go
```diff
@@ -59,6 +59,14 @@ type BackupSpec struct {
 	// +nullable
 	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
 
+	// OrLabelSelectors is list of metav1.LabelSelector to filter with
+	// when adding individual objects to the backup. If empty
+	// or nil, all objects are included. Optional. If multiple provided
```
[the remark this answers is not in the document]
shubham-pampattiwar: Done, updated the spec description.

pkg/apis/velero/v1/restore.go
```diff
@@ -71,6 +71,14 @@ type RestoreSpec struct {
 	// +nullable
 	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
 
+	// OrLabelSelectors is list of metav1.LabelSelector to filter with
+	// when restoring individual objects from the backup. If empty
+	// or nil, all objects are included. Optional. If multiple provided
```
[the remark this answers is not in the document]
shubham-pampattiwar: Done, updated the spec description.

pkg/restore/restore.go
```diff
@@ -1820,6 +1833,25 @@ func (ctx *restoreContext) getSelectedRestoreableItems(resource, targetNamespace
 			continue
 		}
 
+		var skipItem = false
+		var skip = 0
+		ctx.log.Infof("orSelectors specified: %s for item: %s", ctx.OrSelectors, item)
```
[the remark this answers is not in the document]
shubham-pampattiwar: Changed to debug level, thanks !

pkg/restore/restore.go
```diff
@@ -1820,6 +1833,25 @@ func (ctx *restoreContext) getSelectedRestoreableItems(resource, targetNamespace
 			continue
 		}
 
+		var skipItem = false
```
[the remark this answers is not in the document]
shubham-pampattiwar: Added comments.

pkg/controller/backup_controller.go
shubham-pampattiwar: Done, added a testcase for validation failure.

pkg/controller/restore_controller.go
shubham-pampattiwar: Done, added a testcase for validation failure.

pkg/backup/item_collector.go
shubham-pampattiwar: Done, updated the PR

pkg/backup/item_collector.go
```diff
@@ -296,50 +296,70 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
...
 		} else {
-			// If limit is not positive, do not use paging. Instead, request all items at once
-			unstructuredList, err := resourceClient.List(metav1.ListOptions{LabelSelector: labelSelector})
+			labelSelectors = []string{}
+		}
+
+		log.Info("Listing items")
+		unstructuredItems := make([]unstructured.Unstructured, 0)
+		for _, label := range labelSelectors {
+			unstructuredList, err := resourceClient.List(metav1.ListOptions{LabelSelector: label})
```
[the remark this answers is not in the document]
shubham-pampattiwar: Done, updated the PR, Thanks !

sseago (dismissed):

pkg/apis/velero/v1/backup.go
```diff
@@ -59,6 +59,15 @@ type BackupSpec struct {
...
 	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
 
+	// OrLabelSelectors is list of metav1.LabelSelector to filter with
+	// when adding individual objects to the backup. If multiple provided
+	// they will be joined by the OR operator. LabelSelector as well as
+	// OrLabelSelectors cannot co-exist in backup request, only one of them
+	// can be used.
+	// +optional
+	// +nullable
+	OrLabelSelectors []*metav1.LabelSelector `json:"orLabelSelectors,omitempty"`
```
kaovilai, line 69: Why not `*[]metav1.LabelSelector `? This means OrLabelSelectors can be nil, but if it is not nil, you don't have to check again that each OrLabelSelectors[i] is not nil.
```suggestion
	OrLabelSelectors *[]metav1.LabelSelector `json:"orLabelSelectors,omitempty"`
```

kaovilai (changes requested):

pkg/apis/velero/v1/backup.go
kaovilai: on second thought...  this would nullable already.
```suggestion
	OrLabelSelectors []metav1.LabelSelector `json:"orLabelSelectors,omitempty"`
```

pkg/backup/item_collector.go
```diff
@@ -291,53 +291,88 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
...
+
+		}
+
+		var labelSelector string
+		if selector := r.backupRequest.Spec.LabelSelector; selector != nil {
+			labelSelector = metav1.FormatLabelSelector(selector)
+		}
+
+		// Listing items for labelSelector (singular)
+		if len(orLabelSelectors) == 0 {
```
reasonerjt, line 347: it seems when `len(orLabelSelectors) == 0` and you choose to use `labelSelector`, the logic is almost identical to the case when there's only one element in the `orLabelSelectors`?

Why not combine them into one func, that takes the slice of label selectors as a param and returns the `unstructuredItems `

shubham-pampattiwar: Thanks @reasonerjt I have updated the PR, please take another look.

reasonerjt (approved):
lgtm, thanks @shubham-pampattiwar

Issue #3285, opened 2021-01-15
Support Velero plugin versioning


nrb:
**Describe the problem/challenge you have**
Right now, when Velero makes changes to its plugin APIs, there's no mechanism for the Velero server to communicate the version of the API that is supported, or for the plugins to communicate what version they implement.

This leads to restrictions on what can be changed in the API, or breaking everyone's code with major releases. Backwards-compatible changes either lead to API expansion, or implementations that have to check for the presence of specific functions, which is a runtime check and prone to error.

**Describe the solution you'd like**
Ideally, there would be a version negotiation system that allows Velero's server to support some level of version skew, decoupling Velero and plugins from having to move their API in lock step. This way, Velero core can support some number of API versions while iterating on new API features, and plugin authors are not forced to upgrade right away.

The [go-plugin](https://github.com/hashicorp/go-plugin) library we use has support for this, but I don't know how sophisticated their implementation is.

**Anything else you would like to add:**
 https://github.com/vmware-tanzu/velero/pull/1303 updated go-plugin to a version that supported plugin version negotiation, but I think there's still work to be done on top of that.

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

nrb:
@phuongatemc I know you're interested in this, and @sseago has some work that would greatly benefit from having this in place.

reasonerjt:
I'll double check the PRs and associate with this issue.

reasonerjt:
The versioning mechanism has been provided via #4943 
The incremental changes like providing BIA v2 will be tracked via other issues.

Issue #3169, opened 2020-12-09
Collect installed plugin information in a central place


nrb:
**Describe the problem/challenge you have**
Right now, backups don't report what plugins were installed at backup time. An administrator or plugin developer has to know what was there. While this isn't a problem in many cases, having the information about the Velero executing environment at backup time allows for more accurate restore time actions.

**Describe the solution you'd like**
* List installed plugins and their versions (see #3168) within a backup.
  * Could be on the `Backup` CR or a separate file
* Provide a way to read the plugin names and versions to plugins at restore time. This is particularly useful for plugins that might be migrating their own data formats/architectures
* Longer term: validate whether the plugins within a backup are installed prior to running a restore.
  * Do we want to do full version compatibility checking? Or just name-based checks?
  * Some of these will be assumed; if you can get to the BackupStorageLocation, you likely have the right plugin.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

nrb:
@zubron this is part of #3710

Issue #3168, opened 2020-12-09
Add plugin version reporting API


nrb:
**Describe the problem/challenge you have**
Right now, plugins are on their own for version tracking information; they can put it in an annotation for objects they manipulate, but this is hacky and not really effective long term.

**Describe the solution you'd like**
The plugin API should have a method that the Velero server can call if implemented that reports the installed plugin's current version.

This method should be available for any type of Velero plugin.

**Anything else you would like to add:**
Once this information is available to Velero, it should be captured into a central location, whether on the `Backup` object or some other file in the backup.

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

nrb:
This is part of #3710, @zubron

PR #4698, opened 2022-02-24
Plugin versioning design

reasonerjt:
Bridget has refined the design before she left.  I'm creating the PR for more reviews.

### DIFF
--- a/.github/auto-assignees.yml
+++ b/.github/auto-assignees.yml
@@ -15,6 +15,8 @@ reviewers:
       - sseago
       - reasonerjt
       - ywk253100
+      - blackpiglet
+      - qiuming-best
 
     tech-writer:
       - a-mccarthy
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -12,6 +12,8 @@
 | Scott Seago | [sseago](https://github.com/sseago) | [OpenShift](https://github.com/openshift)
 | Daniel Jiang | [reasonerjt](https://github.com/reasonerjt) | [VMware](https://www.github.com/vmware/)
 | Wenkai Yin | [ywk253100](https://github.com/ywk253100) | [VMware](https://www.github.com/vmware/) |
+| Xun Jiang | [blackpiglet](https://github.com/blackpiglet) | [VMware](https://www.github.com/vmware/) |
+| Ming Qiu | [qiuming-best](https://github.com/qiuming-best) | [VMware](https://www.github.com/vmware/) |
 
 ## Emeritus Maintainers
 * Adnan Abdulhussein ([prydonius](https://github.com/prydonius))


design/plugin-versioning.md
```diff
@@ -0,0 +1,289 @@
...
+}
+```
+
+If the previous version is not available, or can not be adapted to the latest version, it should not be included in the `restartableObjectStores` slice.
+This will result in an error being returned as is currently the case when a plugin implementation for a particular kind and provider can not be found.
+
+There are situations where it may be beneficial to check at the point where a plugin API call is made whether it implements a specific version of the API.
+This is something that can be addressed with future amendments to this design, however it does not seem to be necessary at this time.
+
+#### Plugin Adaptation
```
ywk253100, line 232: Why do we need this? Is there any case the adaptation mechanism causes unexpected issues?

ywk253100:
Should we capture the plugin version into the `Backup` CR in case it is used during the restoring just like mentioned here https://github.com/vmware-tanzu/velero/issues/3168 https://github.com/vmware-tanzu/velero/issues/3169?

design/plugin-versioning.md
```diff
@@ -0,0 +1,289 @@
...
+With this proposal, we aim to break this coupling and introduce plugin API versions.
+
+## Scenarios to Support
+The following describes interactions between Velero and its plugins that will be supported with the implementation of this proposal.
+For the purposes of this list, we will refer to existing Velero and plugin versions as `v1` and all following versions as version `n`.
+
+Velero client communicating with plugins or plugin client calling other plugins:
+
+- Version `n` client will be able to communicate with Version `n` plugin
+- Version `n` client will be able to communicate with all previous versions of the plugin (Version `n-1` back to `v1`)
```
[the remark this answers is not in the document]
reasonerjt: I don't think we need to establish any support policy in this doc, in addition, there may be the case that after the change introduced in version `n` the version `n-1` is no longer adaptable.  In that case the developer of version `n-1` plugin will have to update his plugin so that it can work with the latest plugin, though we should be careful when introducing breaking changes to avoid such case.
How about we tweak the language the explain this situation?

design/plugin-versioning.md
reasonerjt: You mean this?
```
When a new plugin API version is being proposed, it will be up to the author and the maintainer team to determine whether older versions of an API can be safely adapted to the latest version.
```

reasonerjt:
@ywk253100 
Seems issues #3168 #3169 and #3285 are not covered in this design.  The draft is simpler IMO

Let's discuss with other maintainers if the points Nolan made in these issues are needed for the first iteration.

design/plugin-versioning.md
```diff
@@ -0,0 +1,289 @@
...
+## Non Goals
+
+- Change how plugins are managed or added
+- Allow older plugin clients to communicate with new versions of plugins
+
+## High-Level Design
+
+With each change to a plugin API, a new version of the plugin interface and the proto service definition will be created which describes the new plugin API.
+The plugin framework will be adapted to allow these new plugin versions to be registered.
+Plugins can opt to implement any or all versions of an API, however Velero will always attempt to use the latest version, and the plugin management will be modified to adapt earlier versions of a plugin to be compatible with the latest API where possible.
```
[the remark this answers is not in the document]
sseago: Also, if we're talking about backup plugins, those won't be called on restore, just as restore plugins won't be called on backup -- they're completely different plugins and may not even have the same versions available. But for storage plugins (which would be used on both), the latest/version negotiation is done separately for backup and for restore -- there isn't metadata stored on the backup to say what plugins ran, etc.

design/plugin-versioning.md
sseago: The purpose of this is to ensure that when a new version is released for a particular plugin API, already-existing plugins don't stop working (unless there's a truly breaking change added such that the plugin doesn't make sense anymore without an update). For example, when we add the method to allow waiting for additional items to be ready, the adapter would simply implement the new API method as a no-op, ensuring that prior behavior (not waiting) is preserved for any older plugins still in use.

sseago:
I don't know that capturing the plugin version for each plugin used in the backup is necessary in terms of affecting restore behavior, since we don't want to use that in the adaptor negotiation code, although it might not hurt to add it to status (or at least logging) for debugging purposes. Bur remember, it's not just one plugin. We have multiple types of plugins (BackupItemAction, RestoreItemAction, ObjectStore, etc.), and in some cases we'll have multiple plugins of each type registered, any number of which can be called/used at various points during backup and restore operation. Having restore logic parse this would be confusing and error-prone, so I wouldn't favor that, but if we want to log that information for debugging purposes, that could serve a purpose.

sseago:
#3168 is unrelated to this. It's talking about the particular version of a plugin that executes, not the API version that the plugin implements. i.e. right now all plugins use the "v1" api (not called that now, but will be called that once we implement versioning). However, if a user is using velero-plugin-for-aws, they might be using v1.4 as released with velero 1.8, or v1.3 as released with velero 1.7, etc. That's completely unrelated to this functionality.

sseago:
#3169 is similarly unrelated to this, as it's related to #3168.

phuongatemc (approved):
The design looks good to me.  The introduce of the ability to create breaking changes by remove functions or change of function parameters will introduce some additional complication in developing the adaptor during implementation phase especially when plugin vendor only provide older version.  If Velero server side cannot create an adaptor for the breaking change, the vendor needs to write new version of the plugin.
Since the breaking change eventually happen, we could introduce a document that map between compatible version of Velero and version of Vendor plugins.

reasonerjt:
@sseago 
> https://github.com/vmware-tanzu/velero/issues/3285 is essentially the thing that this design is attempting to implement, so I'm not sure what you mean by "it's not covered in this design".

It's not covered because it's not trying to do as the description:

> Ideally, there would be a version negotiation system that allows Velero's server to support some level of version skew, decoupling Velero and plugins from having to move their API in lock step. This way, Velero core can support some number of API versions while iterating on new API features, and plugin authors are not forced to upgrade right away.
> 
> The [go-plugin](https://github.com/hashicorp/go-plugin) library we use has support for this, but I don't know how sophisticated their implementation is.
>

ywk253100 (approved):

PR #4705, opened 2022-02-28
add @once schedule

wgkingk:
In some cases, we don't want to create a backup directly, but rather a schedule that runs only once.

### DIFF
--- a/pkg/controller/schedule_controller.go
+++ b/pkg/controller/schedule_controller.go
@@ -223,8 +223,12 @@ func parseCronSchedule(itm *api.Schedule, logger logrus.FieldLogger) (cron.Sched
 				validationErrors = append(validationErrors, fmt.Sprintf("invalid schedule: %v", r))
 			}
 		}()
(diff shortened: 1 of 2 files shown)


blackpiglet:
I get the idea, but basically the backup will be created right after the schedule is created. It seems not much differently from creating backup directly.

wgkingk:
> I get the idea, but basically the backup will be created right after the schedule is created. It seems not much different from creating backup directly. I think if there is a way to specify the schedule run at some time, and only run at that time without any repeat afterwards, that will be more meaningful.

We want to create a backup immediately by creating a schedule, and no more backups will be created after that. but currently only by setting a cron expression that is never reachable to do it, so we consider whether it should be clearly marked like “@once” @blackpiglet

blackpiglet:
I still don't see the difference between creating a backup directly and creating an only once schedule that is supposed to run right after creation.

reasonerjt:
Closing this one as we don't think it should be merged to upstream.

PR #4740, opened 2022-03-14
Add priorityclass to podaction plugin

### LINKED ISSUE #3933: Backup skipped PriorityClass even when it is referenced by Pod
Steps to reproduce:
- Create a PriorityClass "high-priority" as below:
```
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: high-priority
value: 1000000
globalDefault: false
description: "This priority class should be used for XYZ service pods only."
```
- Create a Pod and specified "priorityClassName" as "high-priority"
- Run Backup with include-cluster-resources set to nil.  log shows that the priorityclasses being skipped
time="2021-07-06T23:01:42Z" level=info msg="Getting items for resource" backup=velero-ppdm/mysql-2021-07-06-16-01-40-mysql group=scheduling.k8s.io/v1 logSource="pkg/backup/item_collector.go:165" resource=priorityclasses
time="2021-07-06T23:01:42Z" level=info msg="Skipping resource because it's cluster-scoped and only specific namespaces are included in the backup" backup=velero-ppdm/mysql-2021-07-06-16-01-40-mysql group=scheduling.k8s.io/v1 logSource="pkg/backup/item_collector.go:192" resource=priorityclasses
- Restore to a different cluster that does not have the PriorityClass "high-priority" ==> Restore failed to restore Pod with error that it cannot find the PriorityClass.

There was an suggestion in Velero community meeting and in the code comments to write a plugin for Pod and return the PriorityClass high-priority as the AdditionalItems to backup.  However, I think such plugin would be a workaround because the main problem is the Backup datapath (getResourceItems in specific) should pick up all resources being referenced by the Pod and backup them accordingly.

phuongatemc:
Adding PriorityClass to podAction of backup and restore plugins.

Fixes #3933 

Testings done:
- Create PriorityClass "high-priority", create Pod with PriorityClassName set to "high-priority".  Trigger Backup ==> check the backup image: backup image does contain the priority-class.json.
- Trigger Restore ==> log show PriorityClass is attempted to restore but skipped because "IncludeClusterResources" was set to "false"
- Delete "high-priority" PriorityClass, trigger Restore with "IncludeClusterResources" set to nil, "high-priority" PriorityClass was restored, Pod also restored successfully.

### DIFF
--- a/pkg/backup/pod_action.go
+++ b/pkg/backup/pod_action.go
@@ -56,12 +56,20 @@ func (a *PodAction) Execute(item runtime.Unstructured, backup *v1.Backup) (runti
 		return nil, nil, errors.WithStack(err)
 	}
 
-	if len(pod.Spec.Volumes) == 0 {
-		a.log.Info("pod has no volumes")
-		return item, nil, nil
+	var additionalItems []velero.ResourceIdentifier
+	if pod.Spec.PriorityClassName > "" {
+		a.log.Infof("Adding priorityclass %s to additionalItems", pod.Spec.PriorityClassName)
+		additionalItems = append(additionalItems, velero.ResourceIdentifier{
+			GroupResource: kuberesource.PriorityClasses,
+			Name:          pod.Spec.PriorityClassName,
+		})
 	}
 
-	var additionalItems []velero.ResourceIdentifier
+	if len(pod.Spec.Volumes) == 0 {
+		a.log.Info("pod has no volumes")
+		return item, additionalItems, nil
+	}
+
 	for _, volume := range pod.Spec.Volumes {
 		if volume.PersistentVolumeClaim != nil && volume.PersistentVolumeClaim.ClaimName != "" {
 			a.log.Infof("Adding pvc %s to additionalItems", volume.PersistentVolumeClaim.ClaimName)
--- a/pkg/backup/pod_action.go
+++ b/pkg/backup/pod_action.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 the Velero contributors.
+Copyright 2017, 2022 the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/backup/pod_action.go
+++ b/pkg/backup/pod_action.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2017, 2022 the Velero contributors.
+Copyright the Velero contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


blackpiglet (approved):

pkg/backup/pod_action.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 the Velero contributors.
+Copyright 2017, 2022 the Velero contributors.
```
jenting, line 2:
```suggestion
Copyright the Velero contributors.
```

https://velero.io/docs/v1.8/code-standards/#copyright-header

Same as belows.

pkg/backup/pod_action.go
```diff
@@ -56,12 +56,20 @@ func (a *PodAction) Execute(item runtime.Unstructured, backup *v1.Backup) (runti
 		return nil, nil, errors.WithStack(err)
 	}
 
+	var additionalItems []velero.ResourceIdentifier
+	if pod.Spec.PriorityClassName > "" {
```
jenting, line 60: The `spec.priorityClassName` is a string, suggest to compare it as
```go
if pod.Spec.PriorityClassName != ""
```

Same as below.

jenting (changes requested):

pkg/backup/pod_action.go
phuongatemc: Look like the new copyright format by github and Velero community.  I'll fix it.

pkg/backup/pod_action.go
phuongatemc: Corrected.

pkg/restore/pod_action.go
```diff
@@ -85,6 +86,11 @@ func (a *PodAction) Execute(input *velero.RestoreItemActionExecuteInput) (*veler
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
-
-	return velero.NewRestoreItemActionExecuteOutput(&unstructured.Unstructured{Object: res}), nil
+	restoreExecuteOutput := velero.NewRestoreItemActionExecuteOutput(&unstructured.Unstructured{Object: res})
+	if pod.Spec.PriorityClassName != "" {
+		a.logger.Infof("Adding priorityclass %s to AdditionalItems", pod.Spec.PriorityClassName)
```
jenting, line 91:
```suggestion
                a.logger.Infof("Restoring priorityclass %s from AdditionalItems", pod.Spec.PriorityClassName)
```

phuongatemc: Not sure I agree with this feedback.  At this step, we only add it to the AdditionalItems to be restored.  Whether it will be restored or not depend on more configuration setting which will be done in the restoreItem.  The restoreItem itself already have the log for "Restoring ..."

sseago: I think I agree with @phuongatemc here. The specific action is adding to Additional items. The restoring from additional items happens in the func that calls this one after the RestoreItemAction returns.

jenting: Sounds good, thanks for the explanation.

jenting (approved):
LGTM

blackpiglet (approved):

Issue #4728, opened 2022-03-08
Expired backup that fail to delete should be easy to select.


kaovilai:
**Describe the problem/challenge you have**
Allow user to use label to filter failed garbage collected backups
~~Either as a new default behavior, or as a new feature flag, there should be an option for Velero to garbage collect TTL expired backups even if it failed validation for reasons such as invalid BSL.~~

Similar to https://github.com/vmware-tanzu/velero/issues/4483 but instead of ObjectStorageSync, it's the behavior of [garbage collection controller](https://github.com/vmware-tanzu/velero/blob/e69fac153ba60dc5129cdda51480a64fbf47b851/pkg/controller/gc_controller.go)

**Describe the solution you'd like**
Add label to backups when garbage collector errors out
~~New default behavior or new feature flag~~

**Anything else you would like to add:**
It would help with hanging failed validation Backups from "dirtying" the logs since we don't care if it failed validation if it expired anyway.

**Environment:**

- Velero version (use `velero version`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

reasonerjt:
Making it a default may cause issues that the backup CR is removed but the actual data remained in the object storage, and the user will not know it from velero.  
Similarly, if this is exposed as a configuration, the user may set the configuration without fulling aware of the situation he/she may get into in the future.

Therefore, I hope we can control it based on the error type, i.e. we only force remove the backup when certain errors happen.

kaovilai:
Makes sense.

The request I received was specifically around missing bsl error.

Maybe a list of ignoredErrors

kaovilai:
Our internal discussion have another idea to resolve by labelling instead. Enable user to use labelselectors to delete expired backups.

Then it's user's action, not our code.

Maybe allow them to
```
kubectl get backup -l velero.io/deleteFailReason=velero.io/deleteFailReason="BSLNotFound" -oname | kubectl delete backup
```

[...]

PR #4757, opened 2022-03-21
Add labels to expired backups failing garbage collection.

kaovilai:

If an expired backup garbage collection fails, a label `velero.io/gc-failure=<Reason>` will be added to the backup custom resource.

You can use this label to filter backups that failed to delete.

Implemented reasons are:
- BSLNotFound: Backup storage location not found
- BSLCannotGet: Backup storage location cannot be retrieved from the API server for reasons other than not found
- BSLReadOnly: Backup storage location is read-only

Fixes #(issue)
#4728 

### DIFF
--- a/changelogs/unreleased/4757-kaovilai
+++ b/changelogs/unreleased/4757-kaovilai
@@ -1,0 +1,1 @@
+Garbage collector now adds labels to backups that failed to delete for BSLNotFound, BSLCannotGet, BSLReadOnly reasons.
--- a/pkg/controller/gc_controller.go
+++ b/pkg/controller/gc_controller.go
@@ -168,6 +168,12 @@ func (c *gcController) processQueueItem(key string) error {
 		return nil
 	}
 
+	// remove gc fail error label after this point
+	delete(backup.Labels, garbageCollectionFailure)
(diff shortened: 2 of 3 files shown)


sseago:
@kaovilai A general comment -- several places in this PR reference "labelSelector" when they should just say "label". We're adding labels here. The user would use a label selector with one of these labels to find the resources we label in this PR.

sseago (dismissed):

sseago (approved):

reasonerjt:
One issue with this approach is that the label may not reflect the most recent failure reason.

Will it have negative impacts?

kaovilai:
> the label may not reflect the most recent failure reason
How so?
Label is a map, and only one value may be assigned to a key `velero.io/gc-failure` at any given time correct?

Are you saying this label should be removed after line 170?  for non-implemented errors failures?
https://github.com/vmware-tanzu/velero/blob/aa3e29384618b1ab4546f3d92fe27c27322d8dc3/pkg/controller/gc_controller.go#L162-L170

kaovilai:
Noted. Will clear if error is not BSL.

sseago (approved):

reasonerjt (approved):

jenting (approved):

Issue #4112, opened 2021-09-08
Design doc for data movement layer


dsu-igeek:
Write a design doc for the data movement layer described in #3229

Lyndon-Li:
Fixed by #5968

PR #4768, opened 2022-03-22
Add datamover design - Part 1 (Problem statement, Goals/Non-Goals, Usecases, workflows)

shubham-pampattiwar:
Signed-off-by: Shubham Pampattiwar <shubhampampattiwar7@gmail.com>

This PR adds the following with respect to Velero datamover:
- Problem Statement
- Goals
- Non-Goals
- Usecases
- User workflows

Fixes https://github.com/vmware-tanzu/velero/issues/4112

### DIFF
--- a/design/datamover.md
+++ b/design/datamover.md
@@ -1,0 +1,55 @@
+# Add Datamover to Velero
+## Problem Statement
+Today, Velero supports snapshotting of volumes using a supported plugin. For some volume/plugin types, these snapshots do not provide durability and portability. Velero would like to implement a generic data mover interface that allows vendors to supply their own data mover to be used for compatible storage types.
+- Specifically, CSI snapshots are not guaranteed to be durable or portable, and this is preventing the CSI plugin from going GA.
+- Kubernetes is moving in such a way that every volume would be backed by a CSI driver. CSI snapshotting allows for a way for a user to take a snapshot, but does not guarantee durability nor portability. We want to provide a generic way for users to guarantee portability and durability of these snapshots for a given provider.
+- We also want to allow vendors to provide their own volume data management strategies to be able to work within the velero backup/restore workflow. 
+
+## What is a Data Mover ?
+Data Mover defines an interface that enables movement of kubernetes data from cluster to backup storage(e.g. object storage) and vice-versa.
+
+## Goals
+- The data mover should be **Pluggable**. This means that a user can leverage different types of movers for the same set of resources as long as the movers conform to a similar interface.
+- The interface for Data Mover should support moving data associated with resources **other than** Persistent Volumes
+  - Volume Snapshots are an immediate need, and will likely drive most of the original design discussions.
+  - Ex. a Data Mover for Kubevirt would move VM data to backup storage. The interface should be generic enough to support this use case.
+- The interface for Data Mover should enable publishing its specific configuration parameters for users to have fine-grained control over the movement workflow
+  - Examples:
+    - Data mover for vsphere volume can configure block vs file mount usage
+    - Kubevirt mover can configure export file type
+    - Image mover can specify image format type
+- End-user interaction with data mover should be exposed through the existing backup/restore operations, additionally 2 new CRDs will be introduced - DataMoverBackup
+and DataMoverRestore which would help facilitate data movement and coupling with Velero.
+
+## Non-Goals
+- This implementation will **NOT** change the existing Velero backup/restore workflow.
+  - It is possible in the future that this interface can be used to replace some legacy Velero implementations, but that is **NOT** a goal for this design/implementation.
+- This design of the interface should not include any specific details about (aside from samples):
+  - A mover type
+    - Although we are focused on volumes for phase 1, the interface should have zero knowledge of the underlying data that needs to be moved.
+    - We may provide a way to set the default data mover but still allow users to choose which data mover to user.
+  - Mover implementations
+    - Ex. How a mover handles volumes at the block vs filesystem level is not relevant to the overarching interface design
+    - On the other hand the user will set some mover specific parameters in the interface, thus the user will be aware about some implementation knowledge
+
+## Use Cases
+- Data Mover for Persistent Volume Snapshots
+  - For a given snapshot type, a capable data mover should be able to clone the snapshot data into some backup target. Of course,  we will be supporting the BSL targets but
+  the design will be extensible enough to support other kinds of backup targets as well. 
+  - The most urgent use case for this today is CSI snapshots of volumes where the given CSI driver does not guarantee durability and portability. A user creates a CSI snapshot of a volume and wishes to clone this snapshot data into some backup target. The user then would like to be able to restore a volume from this snapshot that exists in the backup target.
+- User wants to consume third party vendor data movers
+  - Example: Dell Power Protect has their own data mover to handle PVs and would like to integrate this into the native Velero backup workflow as opposed to it being a separate and disconnected process.To do this, Velero’s data mover interface has to support the ability for a vendor to bring their own data mover and have the Velero backup/restore workflow trigger its process.
+- User wants to move data associated with resources other than Persistent Volumes
+  - Example: User takes a backup of a kubevirt VM. A Kubevirt data mover would be capable of exporting the VM and pushing it to backup storage.
+  - Data mover interface should be capable of defining mover strategies for any type of kubernetes resource.
+- Data movement lifecycle will be tied up with Velero Backup/Restore.
+  - Currently, for a given backup if deleted, all the associated snapshots, restores etc. get deleted, we will follow similar behavior and also delete the moved data when backup is deleted.
+- Impact of datamovement on backup/restore status
+  - The status of snapshot upload/download will be cascaded to the backup/restore CRs
+- Datamovement monitoring
+  - Velero will poll the datamover plugins for status on the snapshot operations
+- Datamover cancel operation
+  - Velero can attempt to cancel a datamover operation by calling the Cancel API call on the BIA/RIA (new version). The plugin can then take any appropriate action as needed. Cancel will be called on backup deletion, and possibly upon reaching timeouts, if timeout support is included in this feature. This can even be a No-Op.
+- Datamover usage/registration will be enabled via the Velero install command as a plugin.
+
+


design/datamover.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Background
+Currently, Velero has two option to support backing up of persistent volumes.
+
+1. Restic: This solution performs filesystem level backups and mostly used when a volume snapshot plugin is not available. However, the restic integration is
+   tightly coupled with Velero, makes it difficult for someone to bring their own implementation of data movement. Although restic is portable but it is not
+   crash consistent as it requires quiescing of data before the backup is taken thus affecting the availability of application.
+
+2. Volume Snapshots: Storage providers provide with volume snapshot capabilities via plugins. These snapshots are crash consistent but there is no guarantee that
+   these snapshots are durable. These snapshots are generally stored locally in the cluster storage, so if the cluster breaks then the snapshots are of no use. Velero also supports
```
reasonerjt, line 16: I don't think the statements around Volume Snapshot is correct.

design/datamover.md
```diff
@@ -0,0 +1,208 @@
...
+## High-Level Design ([VolSync](https://github.com/backube/volsync) as a working Proof of Concept)
+The goal of this high level design is to explain a working PoC of moving CSI snapshots where-in a datamover (VolSync) works in conjunction with Velero via a DataMover plugin (will be installed when the DataMover flag is set to true).
+
+The following design proposal is for adding a datamover feature to Velero which can be integrated with various user implemented datamovers.
+
+![](datamover.png)
+
+The design components and CRs that we are introducing are as follows:
+
+1. DataMoverType(DMT): This is a cluster scoped Custom Resource that will have details about the data mover provisioner.The provisioner will be responsible
```
reasonerjt, line 43: I understand the idea that a DMT maps to an actual mover that moves the data.
But the workflow is a little confusing to me, could you clarify:
Is this `DataMover plugin` a velero plugin ?  
How does it also responsible for installing the controllers?
How the plugin creates the CR.

design/datamover.md
```diff
@@ -0,0 +1,208 @@
...
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.7.0
+  name: datamovertype.velero.io
+spec:
+  provisioner: <VolSync>
+```
+
+2. DataMoverBackup(DMB): Assuming that the flag to enable DataMover is set to true during Velero install, when a velero backup is created, it triggers DataMover plugin
```
reasonerjt, line 60: IMO the movement should be optional.
I don't think it's correct that we always create a DMB to kickoff the data movement for each backup, each pvc or snapshot.
For example on EKS the snapshot is moved under the hood so the DMB is no longer needed.

design/datamover.md
```diff
@@ -0,0 +1,208 @@
...
+   velero CSI plugin is used for backup, `VolumeSnapshot` is used as the type or else `PVC` is used.
+
+```
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.7.0
+  name: datamoverbackup.velero.io
+spec:
```
reasonerjt, line 71: 1. The source should be flexible enough such that it's possible to move the "logical snapshot" like the SQL dump.
2. The target of the movement should be included in the spec.  The target should also be flexible to support heterogeneous data structs (bslRef, repositoryRef ...)
3. It should explicit set it's block-based or file based? (Maybe this configuration should be put into resource like DMT)

design/datamover.md
```diff
@@ -0,0 +1,208 @@
...
+    name: <PVC_claim_name>
+    namespace: <namespace>
+  config:     //optional based on the datamover impl
+    resticSecret:
+      name: <secret_name>
+```
+
+4. DataMover Flag: This will be a Velero install flag, if this is flag is set to true then Velero will be installed with the DataMover Plugin.
+
+5. DataMover Plugin: This is the plugin which will get installed when DataMover Flag is set to true. This plugin's duty would be to create DataMoverBackup(DMB) CR. This
```
reasonerjt, line 120: Does it consist of BackupItemAction and RestoreItemAction plugins?

design/datamover.md
```diff
@@ -0,0 +1,208 @@
...
+  # https://restic.readthedocs.io/en/stable/030_preparing_a_new_repo.html
+  AWS_ACCESS_KEY_ID: <access_id>
+  AWS_SECRET_ACCESS_KEY: <access_key>
+```
+*Note: More details for installing restic secret in [here](https://volsync.readthedocs.io/en/stable/usage/restic/index.html#specifying-a-repository)*
+
+
+Like any other plugins, DataMover plugin gets deployed as an init container in the velero deployment. This plugin will be responsible for creating `DataMoverBackup` & `DataMoverRestore` CRs.
+
+Once a DataMoverBackup CR gets created, the controller will create the corresponding `ReplicationSource` CR in the protected namespace. VolSync watches for the creation of `ReplicationSource` CR and copies the PVC data to the restic repository mentioned in the `restic-secret`.
```
reasonerjt, line 160: By reading the following section I don't quite understand how CSI volumesnapshotcontent is moved.

design/datamover.md
```diff
@@ -0,0 +1,208 @@
...
+  dataMoverType: <datamovertype name>
+  destinationClaimRef:
+    name: <PVC_claim_name>
+    namespace: <namespace>
+  config:     //optional based on the datamover impl
+```
+
+Config section in the above CR is optional. It lets the user specify extra parameters needed by the data mover. For eg: VolSync data mover needs restic secret to perform backup & restore
+
+eg:
```
reasonerjt, line 101: We should set `source` in the spec of the CR?

design/datamover.md
sseago: @reasonerjt So there will be a BackupItemAction plugin which creates DataMoverBackup resources in the cluster upon backing up a PVC or CSI snapshot, and a RestoreItemACtion plugin which creates a DataMoverRestore upon restoring a DataMoverBackup. These would just be normal velero plugins and would not be responsible for installing controllers.

design/datamover.md
sseago: @reasonerjt yes

design/datamover.md
ywk253100, line 43: Shouldn't the `DataMover` plugin be installed by Velero? Why is the provisioner responsible for creating it?

sseago: The DataMover plugin would be installed by velero, as it would consist of a BackupItemAction and RestoreItemACtion. This is referring to the DataMoverType -- it's a CRD to create, not a plugin to register with the velero deployment.

design/datamover.md
```diff
@@ -0,0 +1,208 @@
...
+
+
+Related Issue: https://github.com/vmware-tanzu/velero/issues/3229
+
+## Goals
+- To be able to backup CSI snapshots to object storage
+- To be able to recreate CSI snapshots under a different storage provider
+- To provide an extensible data mover solution which should allow third parties to plugin their own datamover implementation
+- To provide a Datamover example PoC
+
```
half-life666, line 27: Apart from the example PoC, it looks like this design has two major goals:

1)   export/import CSI snapshot to/from S3;
2)   extensible data mover solution. 

From my opinion, these are two different topics, but it seems we are trying to put them together into one design, by implementing 2, we can easily achieve 1. However, from this document, it looks like it's more focused on 2, but lack some consideration for 1.

For example, for goal 1, the design might also include retention and frequency management where user can setup the export differently from  backup. E.g., a user may want to do CSI snapshot backup hourly, then export one snapshot daily; and only keep 7 snapshots per PV in a week, then want to keep 1 monthly export.  Basically retention and frequency are common configurations from traditional backup softwares.

design/datamover.md
```diff
@@ -0,0 +1,208 @@
...
+- Maintain 3rd party data mover implementations
+- Adding a status watch controller to Velero
+- To support incremental backups (future scope, dependent on standard CSI support for snapshot differentials)
+- To allow plug-ins for services like encryption, compression etc. (future scope)
+
+## High-Level Design ([VolSync](https://github.com/backube/volsync) as a working Proof of Concept)
+The goal of this high level design is to explain a working PoC of moving CSI snapshots where-in a datamover (VolSync) works in conjunction with Velero via a DataMover plugin (will be installed when the DataMover flag is set to true).
+
+The following design proposal is for adding a datamover feature to Velero which can be integrated with various user implemented datamovers.
+
```
half-life666, line 38: I'm not sure using plugin is an appropriate approach for data mover design. velero plugin framework is extensible, however, as I understand it, there are some issues we need to consider if we use plugin to implement data mover:

-   Plugin is independent binary, the memory and CPU consumption is not easy to control

    For example, the velero deployment has configured memeory limit as 512MB, the current usage is already 500MB, and now velero has to start a new plugin which may consume 13MB memeory, the start of the plugin will hit memory limit and thus can not be started, which may impact the backup/restore process.

-   Plugin logic should be straightforward, we should try not to put complex logic inside plugin?  

    Plugin implementation is like a black box, the caller has no idea what the item action is doing if there are multiple actions registered for the same item, the caller just put the required parameters and sychronously execute the the action until done (this will bring another topic: see below). 

    It's hard to imagine that the plugin will contain complex logic and the whole backup process has barely any control over it. The backup process does not even know what CR has been created, what phases it has been through, if it takes long time, should it be terminated or not? And different plugin will have different behaviors, how to make sure the whole backup (with export) process has a unified experience?

-   Export execution has to be done sychronously with backup

    The data export action here can actually be done in a different time, and has no tight binding with existing Backup CR, but if we use plugin, that means the backup and export can not be separated, a job with export enabled will extend the whole backup duration and make the backup seriliazation problem more severe

Given above, I think we should also consider another approach to implement data mover: to introduce a corrdination controller inside velero to perform backup with export & restore with import, where the specific export/import can use different set of CR/controller to accomplish the job.

This approach means the velero has to create the vendor specific (or default data mover) CR itself, and has to watch the status of CR, and has to control the whole export process. This approach will have bigger impact to existing code base in short term, but it has solid controler over the whole backup-export process, and has control over the vendor data mover CR. In the long run, this will have better code quality at data mover area. 

For plugin approach, in short term it will have less impact to existing code base, however, in the long run, because of the above issues,  it may require more base code change to accormandate the hard-to-control plugin implementation.

Anyway, I'm not trying to iterate and compare on every aspect here, just want to bring up that: do we need to consider the alternative approcach for the design?

shubham-pampattiwar:
As discussed, this PR is updated with the first aspects of datamover interface design which have been agreed upon:
- Problem Statement
- Goals
- Non-Goals
- Usecases
- User workflows

xing-yang:
I assume this design doc is WIP?  It does not have API designs yet.

dymurray:
This PR in it's current state is to capture the requirements as they were gathered in the community discussions. We intend to update this with a full design if we can all agree on these takeaways from the working group calls.

design/datamover.md
```diff
@@ -0,0 +1,48 @@
...
+  - Mover implementations
+    - Ex. How a mover handles volumes at the block vs filesystem level is not relevant to the overarching interface design
+
+## Use Cases
+- Data Mover for Persistent Volume Snapshots
+  - For a given snapshot type, a capable data mover should be able to clone the snapshot data into some backup target.
+  - The most urgent use case for this today is CSI snapshots of volumes where the given CSI driver does not guarantee durability and portability. A user creates a CSI snapshot of a volume and wishes to clone this snapshot data into some backup target. The user then would like to be able restore a volume from this snapshot that exists in the backup target.
+- User wants to consume third party vendor data movers
+  - Example: Dell Power Protect has their own data mover to handle PVs and would like to integrate this into the native Velero backup workflow as opposed to it being a separate and disconnected process.To do this, Velero’s data mover interface has to support the ability for a vendor to bring their own data mover and have the Velero backup/restore workflow trigger it’s process.
```
sgatdell, line 37: Since Dell PowerProtect does various additional activities like quiescing application pods through application templates before taking CSI snapshot, custom support for storage that support volumegroup snapshots CRD etc, I am not sure it can switch to Velero for CSI snapshots. In any case,  datamovement should be optional and controllable as part of backup request or Velero deployment.

design/datamover.md
```diff
@@ -0,0 +1,48 @@
+# Add Datamover to Velero
+## Problem Statement
+Today, Velero supports snapshotting of volumes using a supported plugin. For some volume/plugin types, these snapshots do not provide durability and portability. Velero would like to implement a generic data mover interface that allows vendors to supply their own data mover to be used for compatible storage types.
+- Specifically, CSI snapshots are not guaranteed to be durable or portable, and this is preventing the CSI plugin from going GA.
+- Kubernetes is moving in such a way that every volume would be backed by a CSI driver. CSI snapshotting allows for a way for a user to take a snapshot, but does not guarantee durability nor portability. We want to provide a generic way for users to guarantee portability and durability of these snapshots for a given provider.
+- We also want to allow vendors to provide their own volume data management strategies to be able to work within the velero backup/restore workflow. 
+
+## What is a Data Mover ?
+Data Mover defines an interface that enables movement of kubernetes data between backup storage.
```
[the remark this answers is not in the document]
shubham-pampattiwar: `backup storage` is `object storage`  here

design/datamover.md
```diff
@@ -0,0 +1,48 @@
...
+- The data mover should be **Pluggable**. This means that a user can leverage different types of movers for the same set of resources as long as the movers conform to a similar interface.
+- The interface for Data Mover should support moving data associated with resources **other than** Persistent Volumes
+  - Persistent Volumes are an immediate need, and will likely drive most of the original design discussions.
+  - Ex. a Data Mover for Kubevirt would move VM data to backup storage. The interface should be generic enough to support this use case.
+- The interface for Data Mover should enable publishing it’s specific configuration parameters for users to have fine-grained control over the movement workflow
+  - Examples:
+    - Data mover for vsphere volume can configure block vs file mount usage
+    - Kubevirt mover can configure export file type
+    - Image mover can specify image format type
+- End-user interaction with data mover should be exposed through the existing backup/restore operations
```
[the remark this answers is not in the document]
shubham-pampattiwar: Yes we would likely introduce some new CRDs?APIs for facilitating data movement. 
We would want  the user to simply use `velero backup`

design/datamover.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Non-Goals
+- This implementation will **NOT** change the existing Velero backup/restore workflow.
+  - It is possible in the future that this interface can be used to replace some legacy Velero implementations, but that is **NOT** a goal for this design/implementation.
+- This design of the interface should not include any specific details about (aside from samples):
+  - A mover type
```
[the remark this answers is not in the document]
shubham-pampattiwar: We will be having a data mover per object/resource type.

design/datamover.md
```diff
@@ -0,0 +1,48 @@
...
+    - Although we are focused on volumes for phase 1, the interface should have zero knowledge of the underlying data that needs to be moved.
+  - Mover implementations
+    - Ex. How a mover handles volumes at the block vs filesystem level is not relevant to the overarching interface design
+
+## Use Cases
+- Data Mover for Persistent Volume Snapshots
+  - For a given snapshot type, a capable data mover should be able to clone the snapshot data into some backup target.
```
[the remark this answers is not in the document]
shubham-pampattiwar: Currently, we believe BSL targets are sufficient.

design/datamover.md
```diff
@@ -0,0 +1,48 @@
...
+  - The most urgent use case for this today is CSI snapshots of volumes where the given CSI driver does not guarantee durability and portability. A user creates a CSI snapshot of a volume and wishes to clone this snapshot data into some backup target. The user then would like to be able restore a volume from this snapshot that exists in the backup target.
+- User wants to consume third party vendor data movers
+  - Example: Dell Power Protect has their own data mover to handle PVs and would like to integrate this into the native Velero backup workflow as opposed to it being a separate and disconnected process.To do this, Velero’s data mover interface has to support the ability for a vendor to bring their own data mover and have the Velero backup/restore workflow trigger it’s process.
+- User wants to move data associated with resources other than Persistent Volumes
+  - Example: User takes a backup of a kubevirt VM. A Kubevirt data mover would be capable of exporting the VM and pushing it to backup storage.
+  - Data mover interface should be capable of defining mover strategies for any type of kubernetes resource.
+
+## Workflow
```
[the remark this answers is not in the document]
shubham-pampattiwar: 1. Datamover will likely be installed as a plugin
2. yes, if possible we will try to use the upload progress monitoring
3. This behavior will likely be similar to what we do when we delete backups currently
4. Yes we would be taking the datamover CRDs in the backup for additional information

design/datamover.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Sample PV Backup Workflow
+![](Backup-UserFlow.png)
+
+### Sample PV Restore Workflow
```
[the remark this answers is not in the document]
shubham-pampattiwar: yes most like from the status of datamover  backup CRD instance

design/datamover.md
aglitke, line 42: I am a bit confused by the workflow and how it relates to existing backup and restore.  In order to be compatible with the existing velero flows wouldn't we need something like the following:

Backup/offload
1. Velero is configured to use CSI Snapshots.
2. User requests a backup that includes some PVs
3. Backup is performed (this results in the creation of some VolumeSnapshot objects)
4. DataMover offloads the VolumeSnapshot object by: cloning the VolumeSnapshot to a temporary volume, copying the data from the temporary volume to a BSL, and deleting the temporary volume.

Load/Restore
1. VolumeSnapshot is recreated in the cluster by: creating a temporary volume in the cluster, loading data from the BSL into the volume, creating a VolumeSnapshot from the volume, and deleting the temporary volume.
2. Restore is performed using the recreated VolumeSnapshot as a source for the data.

design/datamover.md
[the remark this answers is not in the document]
phuongatemc: Could we add more description on WHEN the data mover would be invoked to move data ( in the context of a Velero Backup or Velero Restore)?  For example, the Velero Backup would back up metadata and invoke plugins to create Snapshot of the PV, then at the end of the Velero Backup, Velero would invoke datamover to move data to backup storage etc.

design/datamover.md
half-life666: For restore, why recreate the VolumeSnapshot? The new restore process should be able to restore the data to PV/PVC which is exactly the same as when it is backup'd.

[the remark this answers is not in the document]
half-life666: For backup and data mover relationship, we need to consider that (maybe not now), the backup process now includes both resource backup and data export (data move to storage), or the backup and data export are separate thing? E.g.:
- When do we consider the backup completes, if a data mover action is configured.
- If a backup is successful, but the associated data export failed, how to set the result for backup? 
- If both are successful, but the backup was deleted unexpectedly, can user do a restore to the exported data with another backup which was close to this one?

design/datamover.md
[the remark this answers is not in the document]
reasonerjt: I wish to suggest we provide a little bit more extensibility here to potentially support other backup targets, of course the backup target can be a reference to the BSL.

design/datamover.md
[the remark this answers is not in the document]
reasonerjt: IMO we may provide a way to set the default data mover but still allow users to choose which data mover to use?

design/datamover.md
```diff
@@ -0,0 +1,48 @@
...
+## Non-Goals
+- This implementation will **NOT** change the existing Velero backup/restore workflow.
+  - It is possible in the future that this interface can be used to replace some legacy Velero implementations, but that is **NOT** a goal for this design/implementation.
+- This design of the interface should not include any specific details about (aside from samples):
+  - A mover type
+    - Although we are focused on volumes for phase 1, the interface should have zero knowledge of the underlying data that needs to be moved.
+  - Mover implementations
+    - Ex. How a mover handles volumes at the block vs filesystem level is not relevant to the overarching interface design
+
+## Use Cases
```
reasonerjt, line 32: We should also address the scenario,  to make sure that when a backup is deleted, the moved data is also deleted

design/datamover.md
```diff
@@ -0,0 +1,67 @@
...
+- User wants to consume third party vendor data movers
+  - Example: Dell Power Protect has their own data mover to handle PVs and would like to integrate this into the native Velero backup workflow as opposed to it being a separate and disconnected process.To do this, Velero’s data mover interface has to support the ability for a vendor to bring their own data mover and have the Velero backup/restore workflow trigger its process.
+- User wants to move data associated with resources other than Persistent Volumes
+  - Example: User takes a backup of a kubevirt VM. A Kubevirt data mover would be capable of exporting the VM and pushing it to backup storage.
+  - Data mover interface should be capable of defining mover strategies for any type of kubernetes resource.
+- Data movement lifecycle will be tied up with Velero Backup/Restore.
+  - Currently, for a given backup if deleted, all the associated snapshots, restores etc. get deleted, we will follow similar behavior and also delete the moved data when backup is deleted.
```
anshulahuja98, line 46: can we untie the lifecycle of the moved data to the lifecycle of backup CRs

We wish the backups to be durable and not dependent on any mistaken deletions etc. 
Basically providing more guarantees on availability of this data.

design/datamover.md
```diff
@@ -0,0 +1,48 @@
...
+  - The most urgent use case for this today is CSI snapshots of volumes where the given CSI driver does not guarantee durability and portability. A user creates a CSI snapshot of a volume and wishes to clone this snapshot data into some backup target. The user then would like to be able restore a volume from this snapshot that exists in the backup target.
+- User wants to consume third party vendor data movers
+  - Example: Dell Power Protect has their own data mover to handle PVs and would like to integrate this into the native Velero backup workflow as opposed to it being a separate and disconnected process.To do this, Velero’s data mover interface has to support the ability for a vendor to bring their own data mover and have the Velero backup/restore workflow trigger it’s process.
+- User wants to move data associated with resources other than Persistent Volumes
+  - Example: User takes a backup of a kubevirt VM. A Kubevirt data mover would be capable of exporting the VM and pushing it to backup storage.
+  - Data mover interface should be capable of defining mover strategies for any type of kubernetes resource.
+
+## Workflow
+
+### Sample PV Backup Workflow
```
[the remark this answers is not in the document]
anshulahuja98: "I was thinking there should be a separate API call to trigger the data movement, so that user will only input very limited, or no additional info in the backup." 
+1

design/datamover.md
shrac: we may also need to address the use case where vendors might want to create a copy of data in the vendor tenant rather than the customer tenant (hardening) - would help in cases where customer's cluster/tenant is compromised (ex: ransomware). In this case vendor can perhaps use the DataMoverBackup CR to explicitly given an intent to harden the data (specifying a target storage SAS which has some expiry). However, deletion would be something vendor might directly control and may want the delete on DataMoverBackup CR to fail.

design/datamover.md
shrac: +1

reasonerjt: @anshulahuja98 
Currently, if a backup is deleted the snapshot will be deleted, IMO deleting the moved data makes the behavior consistent with current implementation.

blackpiglet:
IMO, this document mainly focuses on addressing the requirement of data mover, and it will not settle down all the details.
Since there is no big divergence on what data mover will do, I think maybe we can merge this PR and move on to detail design.

design/datamover.md
blackpiglet: For CSI case, PV restoration is not reusing the same PV, because PV is already backed up by snapshot, so the PV will be recreated from snapshot, but the PV name may still be same.

design/datamover.md
blackpiglet: @sseago 
A little curious about whether the plugin will do the data movement itself, or the plugin will also call some interface implemented by other components?

design/datamover.md
blackpiglet: I also have some concerns about how plugins will be implemented.
Velero doesn't have a control of the operations in plugin. By far, mentioned operations include snapshotting, data movement and data export. Velero doesn't know the trigger of operation state change.

By far, there is no complicated logic implemented in plugin. Not sure about running controller in plugin. Some k8s extension also use plugin way, for example, CSI, but the plugins run as a side car, it's still different from Velero case.

I think plugin should be a thin layer between two systems. I'm not sure whether the future data mover plugin will do data movement itself. IMO, better not.

blackpiglet: Maybe I didn't understand it right. 

Looks like plugins should be responsible for creating DataMoverBackup or DataMoverRestore, but the controller should be implemented in Velero pod, or no new controller is needed, just let Backup and Restore controller to watch over the DaveMover Backup and Restore resources, and backup and restore controllers will read ItemAction's Progress function to figure out the current status of data movement, and update into the DataMover status.
When plugin creates the DaveMover resource, it should also do something to start the real data movement. I'm not clear about how this part should be done. Either a controller watching the DataMover resource -- but better not in plugin, or just call a API to do the job.

blackpiglet (approved):
Looks like the purpose of the design document is already fulfilled.
Although there are many comments, looks like they focusing on details of how data mover is implementation.
I think we can merge this one, and continue the works on a new PR which will focus on the details.

design/datamover.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Use Cases
+- Data Mover for Persistent Volume Snapshots
+  - For a given snapshot type, a capable data mover should be able to clone the snapshot data into some backup target. Of course,  we will be supporting the BSL targets but
+  the design will be extensible enough to support other kinds of backup targets as well. 
+  - The most urgent use case for this today is CSI snapshots of volumes where the given CSI driver does not guarantee durability and portability. A user creates a CSI snapshot of a volume and wishes to clone this snapshot data into some backup target. The user then would like to be able to restore a volume from this snapshot that exists in the backup target.
+- User wants to consume third party vendor data movers
+  - Example: Dell Power Protect has their own data mover to handle PVs and would like to integrate this into the native Velero backup workflow as opposed to it being a separate and disconnected process.To do this, Velero’s data mover interface has to support the ability for a vendor to bring their own data mover and have the Velero backup/restore workflow trigger its process.
```
reasonerjt, line 41: > To do this, Velero’s data mover interface has to support the ability for a vendor to bring their own data mover and have the Velero backup/restore workflow trigger its process.

I assume this will require dev effort, for example, Dell Power Protect will need to create a plugin.

design/datamover.md
```diff
@@ -0,0 +1,67 @@
...
+  - It is possible in the future that this interface can be used to replace some legacy Velero implementations, but that is **NOT** a goal for this design/implementation.
+- This design of the interface should not include any specific details about (aside from samples):
+  - A mover type
+    - Although we are focused on volumes for phase 1, the interface should have zero knowledge of the underlying data that needs to be moved.
+    - We may provide a way to set the default data mover but still allow users to choose which data mover to user.
+  - Mover implementations
+    - Ex. How a mover handles volumes at the block vs filesystem level is not relevant to the overarching interface design
+    - On the other hand the user will set some mover specific parameters in the interface, thus the user will be aware about some implementation knowledge
+
+## Use Cases
```
reasonerjt, line 35: Shall we move some of the discussion in https://github.com/vmware-tanzu/velero/pull/4768/files#r858431087 
into the use case scenario, like:
1) How the datamovement impacts the backup/restore's statuses?  if a datamovement fails will we mark the backup fail as well?
2)Will the user be monitoring the progress of datamovement via existing `velero backup describe` command?
3)Will we allow user to install/resgister a data mover after velero is installed?

design/datamover.md
xing-yang: Does Velero CSI plugin keeps VolumeSnapshotContent K8s API object in the original K8s cluster after a backup is complete? If it is a snapshot from EBS, you are using the snapshot which is already uploaded to S3 as a backup. In this case, it is probably fine to keep the VolumeSnapshotContent as long as you have to way to bring it back if the original K8s cluster is destroyed. If it is a local snapshot, it should not be kept after the data is uploaded to a remote backup location. In this case, it does not make sense to keep the VolumeSnapshotContent in the original K8s cluster and also the local snapshot on the storage system should be deleted after the data is uploaded.

design/datamover.md
reasonerjt: @shubham-pampattiwar 
another use case to reflect the discussion in the community meeting, we should allow user to cancel the data movement in some scenarios if the underlying data mover supports that.

reasonerjt:
We won't need these .png files since the workflow section is gone?

design/datamover.md
Lyndon-Li: I vote on some of the points in this discussion loop:
1. Whether, when and how to start the upload/data movement process should be dynamically decided by multiple factors, however, the plugin mechanism is too closed, so the design should clarify how closed the data mover plugin will be and what the impact is. As a detailed discussion[ here](https://github.com/vmware-tanzu/velero/pull/5108#discussion_r930556000) . I also noticed @reasonerjt also mentioned that the upload/data movement progress should be dynamically decided at [here](https://github.com/vmware-tanzu/velero/pull/4768#discussion_r837195970)
2. The data movement process is resource (CPU, memory, etc.) consuming, the design should clarify a pattern, with which the resources could be equally designated/throttled/friendly presented to kubernetes system, to organize the in-parallel data movements

design/datamover.md
shubham-pampattiwar: Done, Update the PR!

blackpiglet (approved):

Issue #1687, opened 2019-07-23
Allow restoration of the `status` metadata field


rayanebel:
**Describe the problem/challenge you have**
To manage our clusters, we are using Rancher. Rancher run on top of Kubernetes and save all the data in the same ETCD server _(it's like a big operator)_. We tried to use `velero` to be able to restore our Rancher server in case of disaster recovery.

Backup is running perfectly but we are facing to an issue during the restore phase because Rancher save some state information inside `status` part and velero clean all status and metadata before restoring all the objects. So currently, we use velero for the backup and restore the file with multiple `kubectl` command line.

**Describe the solution you'd like**
I don't have the perfect solution but, I saw in velero code the function `resetMetadataAndStatus`

https://github.com/heptio/velero/blob/e371ba78b0844b28359f902076d470ae5a5de0b9/pkg/restore/restore.go#L1117

Maybe as a first step we can add a simple flag `--include-status` or something else to let the choice to the user to execute this function.

Maybe you have another solution or another idea ?

**Anything else you would like to add:**

We also open a ticket in Rancher side : https://github.com/rancher/rancher/issues/21647

**Environment:**

- Velero version
  - Client: v1.0.0
  - Server: v0.11.0
- Kubernetes:
  - Client: v1.14
  - Server: v1.12.8
- Kubernetes installer & version: AKS
- Cloud provider or hardware configuration: Azure AKS/Rancher (2.2.5)

prydonius:
Thanks for filing the issue upstream with Rancher, I do think that it's odd for Rancher to store state in status. That said, I don't think it would hurt to add an annotation or flag (my preference would be to annotate items with something like `velero.io/restore-status: true`). It's possible that other Operators might use status in some meaningful way, so this could be generally useful.

prydonius:
wdyt @skriss @carlisia @nrb?

skriss:
I believe there's another issue here, which is that `status` is actually ignored by the API server when a resource is created or updated, since it's not intended to be specified by the user.  To restore status, I believe we'd have to explicitly capture it and write it to the `/status` subresource for whatever resource we are operating on.  CRDs haven't always supported the `/status` subresource, so that may not be true for some older clusters/CRDs, but going forward it should be. I'm not sure off the top of my head exactly how we would implement this - there are likely complexities around not being able to create the `spec` and `status`  atomically, and possibly other issues.

skriss:
[...]

PR #4785, opened 2022-03-28
Add ability to restore status on selected resources

RafaeLeal:
# Summary

This adds a new field on `Restore`, to allow us to select some resources to restore the status.
The way this works it's that it will have the same `includedResources `/`excludedResources` but under the `restoreStatus` key. The default value will be excluding all resources, maintaining the current behavior

```yaml
spec: 
  restoreStatus:
    includedResources: ["webhooks", "workflows"]
```

Fixes https://github.com/vmware-tanzu/velero/issues/1687


RafaeLeal:
Not sure what's going on with DCO, I added the `Signed-off-by` on every commit 😕

hamza3202:
This should be a good fix for [this issue](https://github.com/vmware-tanzu/velero/issues/1687#issuecomment-1070943320)

RafaeLeal:
> Not sure what's going on with DCO, I added the `Signed-off-by` on every commit 😕

It was a typo on my email 🤦

reasonerjt:
Based on a previous comment:
https://github.com/vmware-tanzu/velero/issues/1687#issuecomment-524928609
I'm not sure allowing users to choose to restore the status is a good idea, and whether it may cause unexpected problems.
I hope we can have more discussion before the PR is merged.

RafaeLeal:
> Based on a previous comment: [#1687 (comment)](https://github.com/vmware-tanzu/velero/issues/1687#issuecomment-524928609) I'm not sure allowing users to choose to restore the status is a good idea, and whether it may cause unexpected problems. I hope we can have more discussion before the PR is merged.

Sure, what can I do to move this discussion forward?
I can see that ideally all `status` should be recoverable from the `spec`, but we have noticed multiple scenarios where this is either not true or has other undesirable consequences.

Having CRDs to manage external resources should become more common as people adopt [operator pattern](https://github.com/cncf/tag-app-delivery/blob/main/operator-wg/whitepaper/Operator-WhitePaper_v1-0.md). This reference is often something that you can't recover from the spec.

But even when you are indeed able to restore it, it means that you're going to need to request an external service to do that, and in a backup/restore scenario, you will put a lot of unnecessary load on this external service. In my scenario, the load is not a problem, but since it's a third party we can reach rate limits. 

This is especially bad because this disruption happens in both clusters (the one being backup and the one restoring the backup) since it's outside the cluster.

reasonerjt:
> Having CRDs to manage external resources should become more common as people adopt [operator pattern](https://github.com/cncf/tag-app-delivery/blob/main/operator-wg/whitepaper/Operator-WhitePaper_v1-0.md). This reference is often something that you can't recover from the spec.

I hope we can discuss this in the community meeting with other maintainers.  For example, is `status` the only field you can leverage, can you use `annotation` to store the information you need for restore?

blackpiglet:
Understand the use case. Some CRD may store information of resource in status.
In my opinion, the resources can be specified to restore status should be limited to CRD, because k8s native resources don't need status for replications. 
Maybe more check conditions should be added.

RafaeLeal:
> I hope we can discuss this in the community meeting with other maintainers. For example, is `status` the only field you can leverage, can you use `annotation` to store the information you need for restore?

The problem is not to store the information itself, since Velero already saves the serialized status, but the restore process itself. We could add an annotation before restoring, but that would mean that the controller would have to be changed to read that annotation... and maybe we are running open-source controllers like Argo, Prometheus Operator, or Flagger and we can't just change how those controllers work.

RafaeLeal:
> Understand the use case. Some CRD may store information of resource in status. In my opinion, the resources can be specified to restore status should be limited to CRD, because k8s native resources don't need status for replications. Maybe more check conditions should be added.

Interesting.. we could consider that. Right now I believe the only way to know if a resource is a CRD or not is to reach the CRD Kubernetes API, right?

blackpiglet:
> Right now I believe the only way to know if a resource is a CRD or not is to reach the CRD Kubernetes API, right?
May be not that complicated. k8s resources are located in specific API group, for example, core group, app group, batch group. Others extensions's group ends with "k8s.io".

sseago:
There are other non-core resources other than CRDs. For example, in an OpenShift cluster, imagestreams are openshift-specific resources, but if you look at the list of CRDs in an openshift cluster, imagestreams won't be in the list. Point being that if you're looking for "resources that are not core k8s resources", looking at the CRDs API is probably the wrong way to go about it. If the idea is to restore status for non-core resources only, maybe a whitelist of specific core API groups would work, though.

RafaeLeal:
Two things I'd like to add up:
- This implementation, by default, doesn't restore status of *any* resource. You need to explicitly say which resources you want to restore. I think this prevents a lot of issues we could have.
- In k8s, you can have resources in multiple api groups. So looking at those to see if it's "native resource" could lead to some inconsistent behaviour like restoring status for some instances and not all of them.

reasonerjt:
In today's community discussion we concluded that velero should deal with this problem despite storing the reference in `status` field may be an anti-pattern.

We'll check this PR in more details and try to merge it in v1.9

pkg/restore/restore.go
```diff
@@ -1289,6 +1323,17 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 		return warnings, errs
 	}
 
+	// if it should restore status, run a UpdateStatus
+	if shouldRestoreStatus {
+		obj.SetResourceVersion(createdObj.GetResourceVersion())
+		updated, err := resourceClient.UpdateStatus(obj, metav1.UpdateOptions{})
```
shawn-hurley, line 1329: I think that you will need to look up if the resource supports the status subresource otherwise for some resources this will return a 404 not found.

RafaeLeal: I was thinking about this problem...
Ideally, we would call the discovery API to know if the resource supports status subresource without coupling with CRDs. 
We could, however, just ignore 404 errors, and maybe log a warning. This way we don't add a new call to the API, and also don't change the RBAC permissions needed to run velero's server. WDYT?

pkg/restore/restore.go
```diff
@@ -1098,19 +1115,30 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
...
 		return warnings, errs
 	}
 
+	shouldRestoreStatus := ctx.resourceStatusIncludesExcludes.ShouldInclude(groupResource.String())
+
+	ctx.log.Infof("restore status includes excludes: %+v", ctx.resourceStatusIncludesExcludes)
+	// Clear out status.
+	if !shouldRestoreStatus {
+		ctx.log.Infof("Resetting status for obj %s/%s", obj.GetKind(), obj.GetName())
+		if obj, err = resetStatus(obj); err != nil {
```
Lyndon-Li, line 1130: It's better put this reset further below. As you can see, when processing RestoreItemAction, the obj may be replaced with the return value of the item action, so the reset to obj here may be overwritten. Maybe we can put it right above resourceClient.Create(obj)?

pkg/restore/restore.go
```diff
@@ -1229,12 +1257,18 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
...
 		}
 		// Remove insubstantial metadata.
-		fromCluster, err = resetMetadataAndStatus(fromCluster)
+		fromCluster, err = resetMetadata(fromCluster)
 		if err != nil {
 			ctx.log.Infof("Error trying to reset metadata for %s: %v", kube.NamespaceAndName(obj), err)
 			warnings.Add(namespace, err)
 			return warnings, errs
 		}
+		fromCluster, err = resetStatus(fromCluster)
```
Lyndon-Li, line 1266: As far as my understanding: since the status has been removed from "fromCluster" here, but the status in "obj" is not removed when shouldRestoreStatus = true, this will make equality.Semantic.DeepEqual(fromCluster, obj)@1279 returns false all the time, which changes the original logic. Am I wrong?

pkg/restore/restore.go
RafaeLeal: I was trying to change as little as possible, but sure, we can do that :)

pkg/restore/restore.go
RafaeLeal: This part of the code is doing the same thing, just split the two tasks...
But I believe you're right. In the scenario that you're restoring status (shouldRestoreStatus = true), and the resource already exists, as is, it's always triggering the update.
Let me try to fix it

RafaeLeal: I just pushed a fix for this, let me know what you think

blackpiglet:
@RafaeLeal 
There is an error in CI action.
`CRD verification - failed! Generated CRDs are out-of-date, please run 'make update' and 'git add' the generated file(s).`

Please run `make update` to generate CRD's deep-copy functions, then commit again.

pkg/restore/restore.go
```diff
@@ -1344,6 +1381,17 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 		return warnings, errs
 	}
 
+	// if it should restore status, run a UpdateStatus
+	if shouldRestoreStatus {
+		obj.SetResourceVersion(createdObj.GetResourceVersion())
```
Lyndon-Li, line 1386: Comment 5: set the status value of obj from the objWithStatus or the kept the status value

pkg/restore/restore.go
```diff
@@ -1111,19 +1128,22 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 		}
 	}
```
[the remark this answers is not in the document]
RafaeLeal: I submitted a PR with these suggestions on the forked repository
https://github.com/RafaeLeal/velero/pull/1/files

I feel that's still complex, but it's hard to simplify without major refactorings.
In this suggested approach, you need to add and remove fields into an unstructured object, which I was trying to avoid... But I guess it's less intrusive. Let me know if that's what you intended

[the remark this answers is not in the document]
Lyndon-Li: "I guess it's less intrusive".  --- yes, agree. This is important I think, because we can avoid regressions.
"need to add and remove fields into an unstructured object" --- if you don't want to do this, I guess you can still copy the entire object but use the status field only. Anyway, let me review this with the team and get you back.
"I feel that's still complex". --- somehow better IMO, let me review with the team for any further improvement

Thanks a lot for your efforts.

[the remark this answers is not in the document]
RafaeLeal: I have little context about this.. so I'm comfortable with both solutions, so I'd go with whatever the team feels it's better to maintain.
> This is important I think, because we can avoid regressions.
I think that if we are afraid of doing more intrusive changes, we should refactor the code so we can have better test coverage. I know that's an "easier said than done" kind of issue... but maybe we can come up with something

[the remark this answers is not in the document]
Lyndon-Li: @RafaeLeal 
Please help to merge https://github.com/RafaeLeal/velero/pull/1/files, since this make less change and less intrusive.
For the overall refactor to the existing logics, we may visit it during the data mover design & change, let's keep the existing logic as is for this PR.

@reasonerjt 
Please also help to have a look at the change where unstructured.NestedFieldCopy，unstructured.SetNestedField are used, see if you have any other proposal


Thanks.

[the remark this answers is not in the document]
RafaeLeal: Merged 👍

pkg/restore/restore.go
```diff
@@ -993,6 +1009,7 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
...
 
 	if groupResource == kuberesource.PersistentVolumes {
+		resetStatus(obj)
```
Lyndon-Li, line 1012: This line seems not necessary as this code block doesn't really create the object. And there is one call resetMetadataAndStatus at line 1133 later before creating the object

pkg/restore/restore.go
```diff
@@ -1649,12 +1691,26 @@ func resetMetadataAndStatus(obj *unstructured.Unstructured) (*unstructured.Unstr
 		}
 	}
 
-	// Never restore status
-	delete(obj.UnstructuredContent(), "status")
+	return obj, nil
+}
 
+func resetStatus(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
```
reasonerjt, line 1697: It's always nil, why return this error?

pkg/restore/restore.go
reasonerjt: @RafaeLeal if that's the case, does it mean this feature only works for resources whose status is subresource?  Is that an acceptable limitation?

In addition I do we have enough information in the `err` ?  LIke the name of the resources?

pkg/restore/restore.go
Lyndon-Li: Sorry for any misleading, but by the last comment I meant the line at 1012, resetStatus(obj), is not necessary, so please consider and remove it. Thanks.

pkg/restore/restore.go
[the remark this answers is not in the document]
reasonerjt: I think it look good.

pkg/restore/restore.go
Lyndon-Li, line 1012: Sorry for any misleading, but by the last comment I meant the line at 1012, resetStatus(obj), is not necessary, so please consider and remove it. Thanks.

pkg/cmd/cli/restore/create.go
```diff
@@ -279,6 +283,13 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
 		},
 	}
 
+	if len([]string(o.StatusIncludeResources)) > 0 {
```
blackpiglet, line 286: Should o.StatusExcludeResources length be checked too?

blackpiglet, line 286: Should o.StatusExcludeResources length be checked, too?

pkg/restore/restore.go
blackpiglet: Looks like this is not related to whether the status of resource is defined as sub-resource. 
To me, as long as the resource's client implements the UpdateStatus method, then it will work.

reasonerjt:
@RafaeLeal please also update the doc to include the change you added to the CRD:
https://github.com/vmware-tanzu/velero/blob/bfdb68a35a51e86ca8839567d15a63de8a673201/site/content/docs/main/api-types/restore.md

pkg/restore/restore.go
RafaeLeal: This was part of the old code that I just splitted the `resetMetadataAndStatus` function.. 
But sure, I believe it's better to use the unstructured API :)

pkg/cmd/cli/restore/create.go
RafaeLeal: Well, I believe if you don't include any resources, it won't trigger anything anyway

pkg/restore/restore.go
RafaeLeal: @reasonerjt I think this is acceptable.. For me at least it seems that would be super weird to have a CRD with a "status" field that is not a subresource "status".

I believe that we have very raw info on `err`, basically the HTTP response from a 404 on Kubernetes. Which is something like
```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "deployments.apps \"dont-exist\" not found",
  "reason": "NotFound",
  "details": {
    "name": "dont-exist",
    "group": "apps",
    "kind": "deployments"
  },
  "code": 404
  ```
  
  @blackpiglet since we are using the dynamic client, it will always implement the `UpdateStatus` method. The problem is the server answer itself

reasonerjt (approved):

blackpiglet (approved):

Issue #4503, opened 2022-01-07
Velero backup stays in progress after OOMkilled 


peterbosalliandercom:
**What steps did you take and what happened:**
In our cluster the velero pod (Kubernetes) runs each night 0:00 hour a scheduled backup . We found that when the velero pod is restarting because of lack of memory (oomkilled) the backup never finishes, and stays in progress:

status:
  expiration: '2022-02-06T00:00:33Z'
  phase: InProgress
  startTimestamp: '2022-01-07T00:00:33Z'
  version: 1

values.yaml settings:
  backupsEnabled: true
    backupStorageLocation:
      name: aws
      bucket: xxxxxxxxxx
  schedules:
    mybackup:
      schedule: "0 0 * * *" # run at midnight
      template:
        ttl: "720h" # keep 30 days

**What did you expect to happen:**
We expect that velero fails the running job instead of keeping it in progress.

**Environment:**
- Velero version (use `velero version`): 1.3.0

sseago:
Currently, in-progress backups are never resumed on velero pod restart. A velero backup must run start to finish in a single pod run in order to complete. I"m not sure what the future plans are around this limitation, but at the moment, the only thing that can be done to work around it would be to delete the backup and screate a new one.

reasonerjt:
We may consider to fix it as we refactor the controller via kubebuilder, such that the controller can update the status on startup by comparing the actual and desired state.

eleanor-millman:
Additional explanation from a Velero dev that helped me understand this better: If a backup is in process and the Velero server pod is restarted (in this example, it ran out of memory and was killed and restarted), the backup stays in progress but Velero can’t continue to work on it because it can’t start work again on a backup already in progress.  Several possibilities to fix this. One is to make the backup tolerant of restarts. But simpler might be to mark the backup as failed.

brnck:
What makes the situation more difficult is that it is hard to automatically check when such behaviour happens. 

Let's say I have scheduled backups that run once a day. I don't know when OOM happens if don't monitor OOMs in general in my cluster. I thought about monitoring metrics exposed by Velero. Ideally, I would want to check how many backups are in state "InProgress" or perform calculations on several metrics to evaluate if there are any backups not in the completed state longer than they should. 

The problem is that OOM also clears some metrics as well. F.i.
* velero_backup_attempt_total
* velero_backup_success_total

Only **velero_backup_total** and **velero_backup_last_successful_timestamp** survives restarts.

[...]

PR #4833, opened 2022-04-14
Make in-progress backup/restore as failed when doing the reconcile

ywk253100:
Make in-progress backup/restore as failed when doing the reconcile to avoid hanging in in-progress status

Fixes #(issue)


sseago (dismissed):

shawn-hurley:
Can someone explain to me why we can't recover here?

kaovilai:
If I have to guess (without looking at the code), the plugin system only report success/failures of backupitemaction and velero does not have an in-progress queue saved, only a counter. A pod restart results in lost of in-progress queue and would requires a start from scratch which is equivalent to recreating another backup.

sseago:
@shawn-hurley I believe this came up on a prior discussion and as I understood it, there's a long-term fix needed -- Velero needs to be able to continue a backup or restore in-progress. That will be a significant refactoring. In the meantime, this status update provides a more accurate status reporting on the status quo -- when the pod is bounced while a backup or restore is in progress, Velero will never finish it. It is, effectively, failed -- this PR makes that explicit, so that when a user looks at the current set of backups or restores, they won't see more than one in-progress item when one of them is not in progress (and never will be).

kaovilai:
Adding a note that this is related to https://github.com/vmware-tanzu/velero/pull/4757 and that further discussions on this topic should also consider adding garbage collection failure to this status

sseago:
@kaovilai It's not so much the plugin system as the controller itself hasn't been designed to resume a previously-started backup or restore. The whole backup/restore processing will need to be refactored to support this. Plugin actions only occur at a specific point prior to backing up or restoring a single item, so I don't think that's nearly as relevant as being able to pick up a backup/restore where we were -- a way of identifying backed-up or restored items on restart would be part of this. Actually, since we don't persist the backup to object storage until the end, for backups, adding resuming behavior may well be a matter of having to just restart the backup. For restores, we could probably check the backup/restore annotations on any existing resources prior to attempting to restore -- but since we already have logic around that (i.e. not restoring what's there), even for restores, "resuming" may just mean "starting over". That's probably something that needs its own design document to be drafted and reviewed apart from (and after) this PR is merged.

ywk253100:
@shawn-hurley @kaovilai Just like @sseago explained, this is a short-term fix.
A long-term fix should be well-considered and designed. We'll keep the [issue](https://github.com/vmware-tanzu/velero/issues/4503) open until the long-term fix introduced

ywk253100:
@reasonerjt The upload progress [PR](https://github.com/vmware-tanzu/velero/pull/4467/files#diff-f0334755c17add3306aecd57a34bc24dfa20a66f617dbaf541042dfd8e0887deR152) introduces a new phase `Uploading`, so it doesn't impact the handling logic for `InProgress` phase. And the uploading progress PR also handles the restarting for `Uploading` CRs.

reasonerjt:
This is better than keeping the backup/restore in "inprogress" state after an accidental restart of the velero process.

I'll approve.

ywk253100:
@sseago @reasonerjt I have rebased the PR to fix the conflict, please take a look again

sseago (dismissed):

reasonerjt (approved):

ywk253100:
@sseago I made a tiny change from your last review: set the `status.failureReason` as well when the backup failed
Please take a look at this PR again if you have a chance.

PR #4858, opened 2022-04-26
Remove VolumeSnapshot created during backup

### LINKED ISSUE #4761: Avoid leaving additional resources after taking csi snapshot
As a follow up to comment https://github.com/vmware-tanzu/velero/issues/4649#issuecomment-1040580012

Currently if we backup user's workload with CSI plugin it well leaves `volumesnapshots` and `volumesnapshotcontents` on user's cluster.
We should investigate if it's possible to mitigate this situation, for example set the deletion policy to retain and remove the CSI snapshot resources.  

Note this may impact the restore and deletion flow.  
If later we discover that the resources can't be deleted to ensure the successful B/R process, we may leave it as-is for the CSI plugin.

blackpiglet:

Fixes #4761

### DIFF
--- a/changelogs/unreleased/4858-jxun
+++ b/changelogs/unreleased/4858-jxun
@@ -1,0 +1,1 @@
+Remove VolumeSnapshots created during backup when CSI feature is enabled.
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/apex/log"
@@ -32,6 +33,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -68,6 +70,7 @@ import (
 	"github.com/vmware-tanzu/velero/pkg/util/logging"
 	"github.com/vmware-tanzu/velero/pkg/volume"
 
+	"sigs.k8s.io/cluster-api/util/patch"
 	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -676,6 +679,10 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
 				backupLog.Error(err)
 			}
 		}
+
+		// Delete the VolumeSnapshots created in the backup, when CSI feature is enabled.
+		c.deleteVolumeSnapshot(volumeSnapshots, volumeSnapshotContents, *backup, backupLog)
+
 	}
 
 	// Mark completion timestamp before serializing and uploading.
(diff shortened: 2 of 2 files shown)


reasonerjt:
It doesn't seem correct to me:
1. The flow should be the same regardless of the deletion policy of vsclass.
2. We should make sure that they don't exist on the source cluster after the backup is taken, rather than remove them from the backup store, (the restore flow may not need it but that's another story)

changelogs/unreleased/4858-jxun
```diff
@@ -0,0 +1 @@
+Remove VolumeSnapshot created during backup, when the VolumeSnapshotClass's DeletionPolicy is set to Delete.
```
reasonerjt, line 1: Is this correct, I don't think it deletes the vs based on the DeletionPolicy.

pkg/controller/backup_controller.go
```diff
@@ -847,3 +857,140 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+					logger.Errorf("fail to recreate VolumeSnapshotContent: %s", err.Error())
+				}
+			}
+		}(vs)
+	}
+}
+
+func (c *backupController) patchVolumeSnapshotContent(req *snapshotv1api.VolumeSnapshotContent, mutate func(*snapshotv1api.VolumeSnapshotContent)) (*snapshotv1api.VolumeSnapshotContent, error) {
+	// Record original json
+	oldData, err := json.Marshal(req)
```
reasonerjt, line 918: why not use patchHelper?

pkg/controller/backup_controller.go
```diff
@@ -847,3 +857,140 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+	req, err = c.volumeSnapshotClient.SnapshotV1().VolumeSnapshotContents().Patch(context.TODO(), req.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
+	if err != nil {
+		return nil, errors.Wrap(err, "error patching PodVolumeRestore")
+	}
+
+	return req, nil
+}
+
+// recreateVolumeSnapshotContent will delete then re-create VolumeSnapshotContent,
+// because some parameter in VolumeSnapshotContent Spec is immutable, e.g. VolumeSnapshotRef
```
reasonerjt, line 946: I thought the reason is that the `VolumeSnapshotRef` will confuse the sidecar controller?  Could you clarify that in the comment?

pkg/controller/backup_controller.go
```diff
@@ -847,3 +857,140 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+// and Source.
+func (c *backupController) recreateVolumeSnapshotContent(vsc *snapshotv1api.VolumeSnapshotContent) error {
+	timeout := 1 * time.Minute
+	interval := 1 * time.Second
+
+	err := c.volumeSnapshotClient.SnapshotV1().VolumeSnapshotContents().Delete(context.TODO(), vsc.Name, metav1.DeleteOptions{})
+	if err != nil {
+		return fmt.Errorf("fail to delete VolumeSnapshotContent: %s", err.Error())
```
reasonerjt, line 954: What's the reason not to use errors.Wrapf?

pkg/controller/backup_controller.go
```diff
@@ -847,3 +857,140 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+
+			// Change VolumeSnapshotContent's DeletionPolicy to Retain before deleting VolumeSnapshot,
+			// because VolumeSnapshotContent will be deleted by deleting VolumeSnapshot, when
+			// DeletionPolicy is set to Delete, but Velero needs VSC for cleaning snapshot on cloud
+			// in backup deletion.
+			if modifyVSCFlag {
+				_, err := c.patchVolumeSnapshotContent(vsc, func(req *snapshotv1api.VolumeSnapshotContent) {
+					req.Spec.DeletionPolicy = snapshotv1api.VolumeSnapshotContentRetain
+				})
+				if err != nil {
```
reasonerjt, line 894: IMO if we failed to patch the vsc, the deletion policy will remain `Delete` in that case we should not continue trying to delete the vs, and we should also at least mark a warning to the backup.

pkg/controller/backup_controller.go
```diff
@@ -847,3 +857,140 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+			}
+		}(vs)
+	}
+}
+
+func (c *backupController) patchVolumeSnapshotContent(req *snapshotv1api.VolumeSnapshotContent, mutate func(*snapshotv1api.VolumeSnapshotContent)) (*snapshotv1api.VolumeSnapshotContent, error) {
+	// Record original json
+	oldData, err := json.Marshal(req)
+	if err != nil {
+		return nil, errors.Wrap(err, "error marshalling original PodVolumeRestore")
```
reasonerjt, line 920: NOT PodVolumeRestore...

pkg/controller/backup_controller.go
```diff
@@ -847,3 +857,140 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+			}
+
+			// Delete VolumeSnapshot from cluster
+			err := c.volumeSnapshotClient.SnapshotV1().VolumeSnapshots(vs.Namespace).Delete(context.TODO(), vs.Name, metav1.DeleteOptions{})
+			if err != nil {
+				logger.Error("fail to delete VolumeSnapshot: %s", err.Error())
+			}
+
+			// After VolumeSnapshot is deleted, change VolumeSnapshotContent's DeletionPolicy value back.
+			if modifyVSCFlag {
```
reasonerjt, line 906: IMO it's better to use a `defer` to recreate vsc and move it into the chunk when the patch is successful.

pkg/controller/backup_controller.go
```diff
@@ -847,3 +857,140 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+				err = c.recreateVolumeSnapshotContent(vsc)
+				if err != nil {
+					logger.Errorf("fail to recreate VolumeSnapshotContent: %s", err.Error())
+				}
+			}
+		}(vs)
+	}
```
reasonerjt, line 913: We should use waitgroup or channel to block the the func until all the goroutines finish the work?

pkg/controller/backup_controller.go
```diff
@@ -646,6 +652,10 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
 				backupLog.Error(err)
 			}
 		}
+
+		// Delete the VolumeSnapshot when the VolumeSnapshotClass's DeletePolicy is set to Retain
```
reasonerjt, line 656: The comment is incorrect.

changelogs/unreleased/4858-jxun
blackpiglet: Thanks, forgot to update this.

pkg/controller/backup_controller.go
blackpiglet: Modified.

pkg/controller/backup_controller.go
blackpiglet: In current controller's manager setting code, VolumeSnapshot related resources' scheme is not registered, so controller runtime's client cannot handle them.

pkg/controller/backup_controller.go
blackpiglet: Modified all PodVolumeRestore information in this function.

pkg/controller/backup_controller.go
blackpiglet: Add brief description why modifying the VSC like that in the function's comments. Details can be found in the related code logic's comments section.

pkg/util/csi/reset.go
```diff
@@ -37,6 +37,12 @@ func ResetVolumeSnapshotContent(snapCont *snapshotv1api.VolumeSnapshotContent) e
 		return fmt.Errorf("the volumesnapshotcontent '%s' does not have snapshothandle set", snapCont.Name)
 	}
 
-	snapCont.Spec.VolumeSnapshotRef = corev1.ObjectReference{}
+	// Due to VolumeSnapshotContent Validation webhook limitation, name and namesapce
+	// cannot be nil.
+	// After checking with CSI expert yangxi, a not existing VolumeSnapshot will also work.
+	snapCont.Spec.VolumeSnapshotRef = corev1.ObjectReference{
```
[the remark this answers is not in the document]
blackpiglet: Modified.

pkg/controller/backup_controller.go
blackpiglet: Modified.

pkg/controller/backup_controller.go
blackpiglet: Right. Use error group to do the job.

pkg/controller/backup_controller.go
blackpiglet: Right. Modified.

pkg/controller/backup_controller.go
blackpiglet: Agree.
backupLog is passed into deleteVolumeSnapshot as logger, so this is already logged as error in backup.

pkg/controller/backup_controller.go
blackpiglet: Modified.

pkg/controller/backup_controller.go
```diff
@@ -865,3 +877,134 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, errors.Wrapf(err, fmt.Sprintf("failed to get VolumeSnapshotContent %s", vsc.Name))
+		}
+		return false, nil
+	})
+	if err != nil {
+		return err
```
reasonerjt, line 983: since you are wrapping other errors in this func, this one should also be wrapped?

pkg/controller/backup_controller.go
```diff
@@ -865,3 +877,134 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+				})
+				if err != nil {
+					logger.Errorf("fail to modify VolumeSnapshotContent %s DeletionPolicy to Retain: %s", vsc.Name, err.Error())
+					return
+				}
+
+				defer func() {
+					err := c.recreateVolumeSnapshotContent(vsc)
+					if err != nil {
+						logger.Errorf("fail to recreate VolumeSnapshotContent: %s", err.Error())
```
reasonerjt, line 925: also log the vsc's name?

pkg/controller/backup_controller.go
```diff
@@ -865,3 +877,134 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+					}
+				}()
+			}
+
+			// Delete VolumeSnapshot from cluster
+			err := c.volumeSnapshotClient.SnapshotV1().VolumeSnapshots(vs.Namespace).Delete(context.TODO(), vs.Name, metav1.DeleteOptions{})
+			if err != nil {
+				logger.Error("fail to delete VolumeSnapshot: %s", err.Error())
```
reasonerjt, line 933: also log the name of the vs?

pkg/controller/backup_controller.go
```diff
@@ -865,3 +877,134 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
 
 	return buf, nil
 }
+
+// deleteVolumeSnapshot delete VolumeSnapshot created during backup, if its
+// This is used to avoid deleting namespace in cluster triggers the VolumeSnapshot deletion,
+// which will cause snapshot deletion on cloud provider, then backup cannot restore the PV.
+// If DeletionPolicy is Retain, just delete it. If DeletionPolicy is Delete, need to
+// change DeletionPolicy to Retain before deleting VS, then change DeletionPolicy back to Delete.
+func (c *backupController) deleteVolumeSnapshot(volumeSnapshots []*snapshotv1api.VolumeSnapshot,
```
reasonerjt, line 886: Generally could you add a few info/debug messages like `deleting VS xxxx`

pkg/controller/backup_controller.go
```diff
@@ -865,3 +877,134 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
 
 	return buf, nil
 }
+
+// deleteVolumeSnapshot delete VolumeSnapshot created during backup, if its
```
ywk253100, line 881: Incomplete comment

pkg/controller/backup_controller.go
ywk253100, line 886: Is it possible that we create a copy of VSC with `DeletionPolicy` as `Retain` and use this VSC to do the backup? In this way, we don't need to know the implementation details(e.g. patch the `Source` and `VolumeSnapshotRef`, the implementation of this part may change)

pkg/controller/backup_controller.go
blackpiglet: Oh, missed to change here. Modified.

pkg/controller/backup_controller.go
blackpiglet: Considered that way in implementation. After discussion with Daniel, I think keeping the DeletionPolicy as original may be more rational, because what we do here is to keep the Snapshot on cloud provider safe and have a trace for  deletion, and remove the namespaced resource VolumeSnapshot. There should be no more changed for resources already created.
Change DeletionPolicy from Delete to Retain will have some behavior change, although Retain is more safe, but it may be user's expected behavior.

pkg/controller/backup_controller.go
```diff
@@ -865,3 +877,134 @@ func encodeToJSONGzip(data interface{}, desc string) (*bytes.Buffer, []error) {
...
+// recreateVolumeSnapshotContent will delete then re-create VolumeSnapshotContent,
+// because some parameter in VolumeSnapshotContent Spec is immutable, e.g. VolumeSnapshotRef
+// and Source. Source is updated to let csi-controller thinks the VSC is statically provsisioned with VS.
+// Set VolumeSnapshotRef's UID to nil will let the csi-controller finds out the related VS is gone, then
+// VSC can be deleted.
+func (c *backupController) recreateVolumeSnapshotContent(vsc *snapshotv1api.VolumeSnapshotContent) error {
+	timeout := 1 * time.Minute
+	interval := 1 * time.Second
+
+	err := c.volumeSnapshotClient.SnapshotV1().VolumeSnapshotContents().Delete(context.TODO(), vsc.Name, metav1.DeleteOptions{})
```
ywk253100, line 966: If the `DeletionPolicy` of `VSC` is `Delete`, is the `VSC` deleted along with `VS`? If it is, does here get a 404 error?

pkg/controller/backup_controller.go
blackpiglet: Modified.

pkg/controller/backup_controller.go
blackpiglet: Modified.

pkg/controller/backup_controller.go
blackpiglet: Modified.

pkg/controller/backup_controller.go
ywk253100: I mean create a copy of the VSC, there should be no change to the original one

pkg/controller/backup_controller.go
blackpiglet: Modified.

pkg/controller/backup_controller.go
blackpiglet: Delete method is called after patch method, DeletionPolicy is updated to Retain in patch method.

ywk253100: I mean the `DeletionPolicy` of `VolumeSnapshotClass`. Updated the original comment

pkg/controller/backup_controller.go
blackpiglet: If we do that way, then need to update VS's referenced VSC to work, and may also consider the new VSC's status change after creation.
I think that can work, but need to add related logic. To me, the two ways both work, but no one is much better than the other.

ywk253100 (approved):

reasonerjt (approved):

Issue #4493, opened 2022-01-06
Using default backup and snapshot location when a number of backup and snapshot locations configured in Velero


danfengliu:
**Describe the problem/challenge you have**
Backup will fail if with no backup and snapshot location parameters in backup cli, and there a number of  backup and snapshot locations configured in Velero

**Describe the solution you'd like**
Steps:
1. Create at least 2 backup and snapshot location separately;
2. Create backup with no backup and snapshot location parameters;
3. Backup will use default  backup and snapshot location and backup successfully.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Velero version (use `velero version`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

reasonerjt:
dup of #2426 ?

reasonerjt:
Also noticed that https://github.com/vmware-tanzu/velero/pull/3561/ which reverted a change the commit message says 
```
We are no longer adding the Credentials field to the VSL...
```
@zubron could you clarify what's reason for this decision?

zubron:
@reasonerjt This decision was made during the multiple credentials work for 1.6. It was initially proposed that we would allow the users to specify credentials for both storage location types but after discussion, it seemed to make less sense for the VSLs. For most cloud providers, it's not possible to take snapshots in different accounts so there isn't really a need to support multiple locations with unique credentials. This is different from the BSL case where it's possible to store the serialized contents of a backup in many storage locations.

This is something that will need to be addressed as part of the data movement work as the contents of snapshots may be moved to other storage locations and credentials will be needed to access both the source and the destination.

Adding support for the multiple VSLs credentials was proposed again in #4115.

qiuming-best:
We have one workaround when doing backup & restore with multiply bsl & vsl is specify bsl and vsl in backup command, which will pass the backup validation.
Here is a command example when backup:
`
[...]

PR #4864, opened 2022-04-26
Add credentials to volume snapshot locations.

### LINKED ISSUE #4862: Expand BSL multiple credentials support to VolumeSnapshotLocations
Originally the per-BSL credentials support was intended to be added to VolumeSnapshotLocations as well. Some initial work was already done for this but was reverted when the VSL work was de-scoped from a prior release (the AWS plugin work to support this remained in place, though). This issue is for going though with the enhancement.

While the need for multiple sets of credentials is not as obvious for VSLs, there's still a use case for it, especially for users who area already setting credentials on their BSLs. For those users, they may not be using the default credentials at all, as they've modified their workflow to set credentials on each BSL they create. This will allow them to follow the same approach with VSLs.

VSLs should allow optional credentials in the same way as BSLs, with the default credentials being used when they're not specified.

A PR to implement this will be submitted shortly.

**Vote on this issue!**

sseago:
This is analogous to the BSL creds work that was done previously, but for VSLs instead.

Originally the per-BSL credentials support was intended to be added to VolumeSnapshotLocations as well. Some initial work was already done for this but was reverted when the VSL work was de-scoped from a prior release (the AWS plugin work to support this remained in place, though). This issue is for going though with the enhancement.

Fixes #4862


reasonerjt:
We'll target this to post-v1.9

reasonerjt:
@sseago I found the comment by Bridget:
https://github.com/vmware-tanzu/velero/issues/4493#issuecomment-1010285577

> For most cloud providers, it's not possible to take snapshots in different accounts so there isn't really a need to support multiple locations with unique credentials. This is different from the BSL case where it's possible to store the serialized contents of a backup in many storage locations.

So will it cause problems for some cloud providers if we allow them to set different credentials for BSL and VSL?

pkg/cmd/cli/snapshotlocation/create.go
```diff
@@ -54,22 +55,27 @@ func NewCreateCommand(f client.Factory, use string) *cobra.Command {
...
-	Name     string
-	Provider string
-	Config   flag.Map
-	Labels   flag.Map
+	Name       string
+	Provider   string
+	Config     flag.Map
+	Labels     flag.Map
+	Credential flag.Map
+	secretName string
```
blackpiglet, line 63: Looks like VSL's create and set command's Options are different.
secretName and secretKey are used as internal values to keep Credential's secret name and key.
Could these two parameters also removed in create command?

sseago: @blackpiglet I've gone through the BSL credentials commits after the initial implementation (some refactoring and validation changes) -- making similar changes here removes these Options fields, as they're not really necessary.

sseago:
I've gone through the BSL credentials commits after the initial implementation (some refactoring and validation changes) -- making similar changes here.

blackpiglet (dismissed):

reasonerjt:
@sseago Thanks, the code looks good to me.
Could you fix the minor issues with the comment and update the doc accordingly to reflect the changes in API and CLI?

sseago:
@reasonerjt PR updated with comment fixes and doc updates.

site/content/docs/main/troubleshooting.md
```diff
@@ -175,9 +175,9 @@ Follow the below troubleshooting steps to confirm that Velero is using the corre
     <Output should be your credentials>
     ```
 
-### Troubleshooting `BackupStorageLocation` credentials
+### Troubleshooting `BackupStorageLocation` and `VolumeSnapshotLocation credentials
```
blackpiglet, line 178: Missed right back-tick quota mark.

sseago: Good catch. Will fix.

shubham-pampattiwar (approved):

Issue #2134, opened 2019-12-15
Openshift randomizes the GID


carlisia:
It would be helpful to document that OS randomizes the gid. For more context:

https://github.com/vmware-tanzu/velero/issues/2023

philipp1992:
Corresponding openshift documentation. 

https://docs.openshift.com/enterprise/3.1/install_config/persistent_storage/pod_security_context.html#fsgroup

a-mccarthy:
We should add a note to the Restic Openshift install section, https://velero.io/docs/v1.5/restic/#install-restic to note that Openshift randomizes group IDs, so when restoring a backup, it will have a randomized GID unless you set the GID. We should also point users to the opeenshift docs that explain this in more detail (see previous comment)

a-mccarthy:
per [this comment](https://github.com/vmware-tanzu/velero/pull/4883#issuecomment-1135077089) on the PR, closing this issue as "wont fix", if you feel like this is still something we should document, please re-open this issue or open a new one. thanks!

PR #4883, opened 2022-05-02
Add note about openshift GID

a-mccarthy:

This adds a note about openshiftt gid randomization to the rest docs.

Fixes #https://github.com/vmware-tanzu/velero/issues/2134

### DIFF
--- a/site/content/docs/main/restic.md
+++ b/site/content/docs/main/restic.md
@@ -34,7 +34,7 @@ To install Restic, use the `--use-restic` flag in the `velero install` command. 
 velero install --use-restic
 ```
 
-When using Restic on a storage provider that doesn't have Velero support for snapshots, the `--use-volume-snapshots=false` flag prevents an unused `VolumeSnapshotLocation` from being created on installation.
+When using Restic on a storage provider that doesn't have Velero support for snapshots, the `--use-volume-snapshots=false` flag prevents an unused `VolumeSnapshotLocation` from being created on installation. If you are installing
 
 ### Configure Restic DaemonSet spec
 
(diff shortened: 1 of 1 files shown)


reasonerjt (approved):

sseago:
So I had a couple things here.

First, this really isn't a restic-specific issue. The issue is avoiding a container GID/filesystem GID mismatch, which applies any time we're restoring a volume and a pod. The concern is setting up pod so that they can always access their volumes. Volume snapshots will be the same, so this probably belongs somehere other than in the restic docs -- perhaps in the PV/PVC documentation?

Second, once we know where this should go, the document linked is to a  years-old version of OpenShift (3.11 is based on Kubernetes 1.11). Also, the fsGroup field in the securityContext referenced is normally automatically managed at pod creation by the configured security context constraints. Maybe the text should be something like this:
"Note, Openshift randomizes group IDs, unless you explicitly set the GID, restored containers will have a randomized GID. Security context constraints manage the ways in which this is configured in created pods. The fsGroup field, in particular, is relevant to container access to mounted volumes. See https://docs.openshift.com/container-platform/4.10/authentication/managing-security-context-constraints.html for more information"

One final comment -- I'm not 100% sure we even need this OpenShift-specific note. The default security context constraints should handle this for users who aren't doing anything unusual, and if they are going beyond the defaults, there are many other concerns than just this one. It may be that we don't need it at all, or if something is needed, then maybe it should be placed in the OADP documentation rather than upstream?

a-mccarthy:
@sseago thanks for the review! I think you make some excellent points. This docs request is relatively old (hence the old link), and i think it completely valid to call out if we need this in our docs at all. 

* Do you feel like this is something that a very small number of users will hit?
* Do you feel like it does more harm that good to have in our docs? (by suggesting something that is less secure/cause other problems)

sseago:
@a-mccarthy From what I've seen, this isn't a likely scenario for users who aren't making non-default changes to security contexts, etc. -- and even then, I'm unaware of a customer hitting this on our end. This may have been a bigger problem a few years ago when this issue was first created. I think for now it's probably fine to close the issue -- if the original issue creator feels a need for Velero documentation to address it, it can be reopened.

a-mccarthy:
thanks @sseago 

closing this per suggestion above.

Issue #3118, opened 2020-11-30
Document how to substitute environment variables in pre/post hook


d3473r:
[carlisia] Update on 1/13: this became a task.

----
**What steps did you take and what happened:**
Create a backup of a mysql container with the hooks:
`pre.hook.backup.velero.io/command=["mysql", "--password=$MYSQL_ROOT_PASSWORD", "-e", "FLUSH TABLES WITH READ LOCK"]`
and
`post.hook.backup.velero.io/command=["mysql", "--password=$MYSQL_ROOT_PASSWORD", "-e", "UNLOCK TABLES"]`

velero describe gives me a PartiallyFailed backup and the logs are showing a wrong password:
` msg="stderr: mysql: [Warning] Using a password on the command line interface can be insecure.\nERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)\n"`
`hookCommand="[mysql --password=$MYSQL_ROOT_PASSWORD -e FLUSH TABLES WITH READ LOCK]"`

If I am running the hook with the password directly it works:
`pre.hook.backup.velero.io/command=["mysql", "--password=test123", "-e", "FLUSH TABLES WITH READ LOCK"]`
`msg="stderr: mysql: [Warning] Using a password on the command line interface can be insecure.\n"`
`hookCommand="[mysql --password=test123 -e FLUSH TABLES WITH READ LOCK]"`

**Anything else you would like to add:**

`MYSQL_ROOT_PASSWORD` is a required env in the mysql container (https://hub.docker.com/_/mysql) and is set in the container

```
root@mysql-84f87c85d6-skvfn:/# echo $MYSQL_ROOT_PASSWORD
test123
```

**Environment:**

- Velero version (use `velero version`): 
```
λ velero version
Client:
        Version: v1.5.2
        Git commit: e115e5a191b1fdb5d379b62a35916115e77124a4
Server:
        Version: v1.5.2
```
- Velero features (use `velero client config get features`): 
```
λ velero client config get features
features: <NOT SET>
```
- Kubernetes version (use `kubectl version`):
```
λ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #4886, opened 2022-05-03
Add info about enviroment variables in pre/post hooks

a-mccarthy:

This adds some information about using environment variables in pre and post hooks.

Fixes #https://github.com/vmware-tanzu/velero/issues/3118

### DIFF
--- a/site/content/docs/main/backup-hooks.md
+++ b/site/content/docs/main/backup-hooks.md
@@ -24,7 +24,7 @@ You can use the following annotations on a pod to make Velero execute a hook whe
 * `pre.hook.backup.velero.io/container`
   * The container where the command should be executed. Defaults to the first container in the pod. Optional.
 * `pre.hook.backup.velero.io/command`
-  * The command to execute. If you need multiple arguments, specify the command as a JSON array, such as `["/usr/bin/uname", "-a"]`
+  * The command to execute. This command is not executed within a shell by default. If a shell is needed to run your command, include a shell command, like `/bin/sh`, that is supported by the container at the beginning of your command. If you need multiple arguments, specify the command as a JSON array, such as `["/usr/bin/uname", "-a"]`. See [examples of using pre hook commands](#backup-hook-commands-examples). Optional.
 * `pre.hook.backup.velero.io/on-error`
   * What to do if the command returns a non-zero exit code.  Defaults is `Fail`. Valid values are Fail and Continue. Optional.
 * `pre.hook.backup.velero.io/timeout`
@@ -36,7 +36,7 @@ You can use the following annotations on a pod to make Velero execute a hook whe
 * `post.hook.backup.velero.io/container`
   * The container where the command should be executed. Default is the first container in the pod. Optional.
 * `post.hook.backup.velero.io/command`
-  * The command to execute. If you need multiple arguments, specify the command as a JSON array, such as `["/usr/bin/uname", "-a"]`
+  * The command to execute. This command is not executed within a shell by default. If a shell is needed to run your command, include a shell command, like `/bin/sh`, that is supported by the container at the beginning of your command. If you need multiple arguments, specify the command as a JSON array, such as `["/usr/bin/uname", "-a"]`. See [examples of using pre hook commands](#backup-hook-commands-examples). Optional.
 * `post.hook.backup.velero.io/on-error`
   * What to do if the command returns a non-zero exit code.  Defaults is `Fail`. Valid values are Fail and Continue. Optional.
 * `post.hook.backup.velero.io/timeout`
@@ -78,7 +78,9 @@ velero backup get nginx-hook-test
 velero backup logs nginx-hook-test | grep hookCommand
 ```
 
-## Using Multiple Commands
+## Backup hook commands examples
+
+### Multiple commands
 
 To use multiple commands, wrap your target command in a shell and separate them with `;`, `&&`, or other shell conditional constructs.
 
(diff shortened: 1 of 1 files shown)


a-mccarthy:
@blackpiglet thanks for your comment on https://github.com/vmware-tanzu/velero/issues/3118! I've added the information here, and had some follow up questions.

Is this information specific for environment variables? would there be other instances where a users would want to make sure that they have included a shell before as part of their pre/post hooks commands? the samples on this page also include shells, https://velero.io/docs/v1.8/backup-hooks/#using-multiple-commands

Is including a shell also required when using restore hooks? It seems so based the sample on the restore hooks, https://velero.io/docs/v1.8/restore-hooks/ should also have a note about shell commands here?

blackpiglet:
@a-mccarthy 

I read the link: https://velero.io/docs/v1.8/backup-hooks/#using-multiple-commands, but I don't understand that has different from the use case in #3118. #3118 describe how to use pre-defined environment variable in backup/restore hooks, and using environment variable in multiple commands also uses the same procedure.

Yes. Restore hook also needs this when using environment.

a-mccarthy:
@blackpiglet I guess i'm wondering if we can be more clear to folks trying to use the command feature for hooks in general. Should a users always include a shell command when using a command for back up and restore hooks? Or only in certain cases?

blackpiglet:
@a-mccarthy 
Oh, get what you mean.
IMO, it's better to use a shell in hook command, because that will do no harm and can use the information in the container's shell, but that's not a must, e.g. some containers don't have a shell.

site/content/docs/main/backup-hooks.md
```diff
@@ -86,6 +86,23 @@ To use multiple commands, wrap your target command in a shell and separate them
...
+- exec:
+    container: mysql
+    command:
+      - /bin/sh
+      - -c
+      - mysql --password=$MYSQL_ROOT_PASSWORD -e "FLUSH TABLES WITH READ LOCK"
+    onError: Fail
+```
+
+Your pod must have integrated a shell to use environment variables.
```
blackpiglet, line 104: May also need to mention different container integrated with different shell, so need to specify the correct shell for container used.

a-mccarthy: i updated the wording a little bit. can you take a look and let me know if this makes sense?

a-mccarthy:
Should I also make an update to the command information on this page?  For example here, https://velero.io/docs/main/restore-hooks/#specifying-restore-hooks-as-pod-annotations

site/content/docs/main/backup-hooks.md
blackpiglet: Thanks. Looks good to me.

blackpiglet (approved):

site/content/docs/main/backup-hooks.md
```diff
@@ -78,14 +78,33 @@ velero backup get nginx-hook-test
...
 
 To use multiple commands, wrap your target command in a shell and separate them with `;`, `&&`, or other shell conditional constructs.
 
 ```shell
     pre.hook.backup.velero.io/command='["/bin/bash", "-c", "echo hello > hello.txt && echo goodbye > goodbye.txt"]'
 ```
 
+#### Using environment variables
+
+You are able to use environment variables from your pods in your pre and post hook commands by including a shell command before using the environment variable. For example, `MYSQL_ROOT_PASSWORD` is an environment variable defined in pod called `mysql`. To use `MYSQL_ROOT_PASSWORD` in your pre-hook, you'd include a shell, like `/bin/sh`, before calling your environment variable:
```
reasonerjt, line 93: The shell is not necessary, the user can also run a compiled program that reads the environment variable in the code.  The snippet is still a valid example but we should just point out that the hook shares the environment variables that are set in the container.

reasonerjt (approved):

anmolkhemuka:
Hi, team I had a question why cant we include /bin/sh -c in the same string element in which we have our command? Why do they need to be sent as separate elements of the command array?
@blackpiglet can you please provide some clarification.

blackpiglet:
@anmolkhemuka 
This PR proposed an example of how to use the environment variable in the hook command.
That doesn't mean it's the only working command format.
The format you asked for should work. The example put the shell and the actual command in different lines because it's a bit complicated shell command. It required different kinds of quotation marks to squeeze all things into one line.

anmolkhemuka:
Thanks a lot for your response. This was my main concern that this command should work in other format as well. But when I tried in the following way it didn't worked out.

In this way I added the command to my proto
 hook->add_command(
     "/bin/sh -c \"echo 'Its testing here.' > /tmp/prehook2.txt\"");
     
 Using this I created a backup object with the following specs
       Name:  connectorRule
      Pre:
        Exec:
          Command:
            /bin/sh -c "echo 'Its new testing here.' > /tmp/prehook2.txt"
          Container:          
          On Error:           Fail
          Timeout:            30s
And when I checked the Velero logs it said the error is 

velero backup logs testbackup8 -n cohesity-591524088395612 | grep error
time="2024-03-11T13:16:31Z" level=error msg="Error executing hook" backup=cohesity-591524088395612/testbackup8 error="command terminated with exit code 126" hookPhase=pre hookSource=backupSpec hookType=exec logSource="internal/hook/item_hook_handler.go:263" name=my-csi-app6 namespace=anmol resource=pods

blackpiglet:
I just did some tests. Putting them in the same line has some issues.
But separating into different lines is an easy workaround. Does that not work for you?

anmolkhemuka:
yeah, I get your point. Putting them in separate lines solves the issue. But when I saw the Velero documentation I could see a command was mentioned 
  mysql --password=$MYSQL_ROOT_PASSWORD -e "FLUSH TABLES WITH READ LOCK"

This also contains some arguments like -e but still, we haven't separated it into elements of command array. In the same way, I was trying to do with /bin/sh and argument -c. But it would be very helpful regarding my understanding if I get why we can't do the same with /bin/sh. Is it due to it being some shell command? I think maybe this can be resolved if we can get information on how the Velero executes these hooks, and why it is not able to do it in the above case.

Thanks for your help on verifying the issue.

PR #4890, opened 2022-05-05
continue rather than return for non-matching restore action label

sseago:
When iterating over applicable restore actions, if a non-matching label
selector is found, velero should continue to the next action rather than
returning from the restoreItem func, which ends up preventing the item's
restore entirely.

When restore was refactored to add additionalItems a few years ago, a bug was introduced. At the beginning of each run through the loop that executes RestoreItemActions, there's a label selector check. The previous code had a `continue` call for non-matching labels, meaning that the action would be skipped, moving on to the next. The refactored code returns from the `restoreItem` func instead, meaning that a single action for the resource with a non-matching label selector included will prevent restore from happening at all. This fix restores the correct behavior. 

Fixes #(issue)

### DIFF
--- a/changelogs/unreleased/4890-sseago
+++ b/changelogs/unreleased/4890-sseago
@@ -1,0 +1,1 @@
+continue rather than return for non-matching restore action label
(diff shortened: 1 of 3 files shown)


shubham-pampattiwar (approved):

reasonerjt (approved):
Thanks @sseago !

ywk253100 (approved):

PR #4894, opened 2022-05-07
Refactor BSL controller with periodical enqueue source

### LINKED ISSUE #4463: Backup Storage Location controller may reconcile one item multiple times in a single period
Currently, the BSL controller watches BSL resource all events, and there is some buggy logic in the reconcile function.

* After reconcile, the key triggers reconcile will be put in rate-limited work queue immediately. This behavior will trigger multiple reconciles that are not necessary. That will only stop, when the rate-limited function trigger's the exponential back-off, and the back-off time is longer than one minute (By the way, I am still not clear about why only relying on the rate-limited function can let the reconcile loop set exactly to one minute). 
* After more BSL related events happen, there will be more BSL reconciles happen. I think only one reconcile routine is needed in one minute. 

Only one reconcile routine runs, and only runs one time in the period of one minute.

**Environment:**

- Velero version (use `velero version`): v1.8
- Velero features (use `velero client config get features`): none
- Kubernetes version (use `kubectl version`): v1.21
- Kubernetes installer & version: docker
- Cloud provider or hardware configuration: Laptop
- OS (e.g. from `/etc/os-release`): OSX

**Vote on this issue!**

blackpiglet:

Fixes #4463

### DIFF
--- a/pkg/controller/backup_storage_location_controller.go
+++ b/pkg/controller/backup_storage_location_controller.go
@@ -29,11 +29,18 @@ import (
 	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/vmware-tanzu/velero/internal/storage"
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"github.com/vmware-tanzu/velero/pkg/persistence"
 	"github.com/vmware-tanzu/velero/pkg/plugin/clientmgmt"
+	"github.com/vmware-tanzu/velero/pkg/util/kube"
+)
+
+const (
+	backupStorageLocationSyncPeriod = 1 * time.Minute
 )
 
 // BackupStorageLocationReconciler reconciles a BackupStorageLocation object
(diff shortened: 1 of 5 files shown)


blackpiglet:
Controller runtime cannot only use Watches function to setup controller. Controller runtime always require a For function. 
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/controller.go#L185-L188

As a result, add BSL in the For function, but even all events of BSL are ignored, there is always some unexpected reconcile running. So need to verify whether need to verify the BSL in the reconcile, instead of only relying on periodical enqueue source.

pkg/controller/backup_storage_location_controller.go
```diff
@@ -53,96 +60,93 @@ type BackupStorageLocationReconciler struct {
...
 		return ctrl.Result{}, err
 	}
 
 	pluginManager := r.NewPluginManager(log)
 	defer pluginManager.CleanupClients()
 
 	var defaultFound bool
-	for _, location := range locationList.Items {
+	for _, bsl := range locationList.Items {
 		if location.Spec.Default {
```
ywk253100, line 80: Should use `bsl` instead?

pkg/controller/backup_storage_location_controller.go
```diff
@@ -169,21 +173,38 @@ func (r *BackupStorageLocationReconciler) logReconciledPhase(defaultFound bool,
...
-		log.Warn("There is no existing backup storage location set as default. Please see `velero backup-location -h` for options.")
+		log.Warn("There is no existing BackupStorageLocation set as default. Please see `velero backup-location -h` for options.")
 	}
 }
 
 func (r *BackupStorageLocationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	g := kube.NewPeriodicalEnqueueSource(r.Log, mgr.GetClient(), &velerov1api.BackupStorageLocationList{}, backupStorageLocationSyncPeriod)
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&velerov1api.BackupStorageLocation{}).
+		// Ignore all events of BSL. Reconciler is supposed to run periodically.
```
ywk253100, line 193: Keeping these events is better. It makes sure the status of BSL is updated immediately no matter what the sync period users configure is

pkg/controller/backup_storage_location_controller.go
blackpiglet: Right. Modified.

pkg/controller/backup_storage_location_controller.go
blackpiglet:
``` go
	if !storage.IsReadyToValidate(location.Spec.ValidationFrequency, location.Status.LastValidationTime, r.DefaultBackupLocationInfo.ServerValidationFrequency, log) {
		log.Debug("Validation not required, skipping...")
		return ctrl.Result{}, nil
	}
```
BSL controller has a check to only validate BSL at an 60s interval, most of the event will be skipped by this check.

blackpiglet: Add creation event handling for validate new creating BSL immediately.

ywk253100 (approved):

pkg/controller/backup_storage_location_controller.go
```diff
@@ -53,96 +60,88 @@ type BackupStorageLocationReconciler struct {
...
+			log.WithError(err).Error("Error getting a patch helper to update BackupStorageLocation")
+			return
+		}
+		defer func() {
+			location.Status.LastValidationTime = &metav1.Time{Time: time.Now().UTC()}
 			if err != nil {
-				log.WithError(err).Error("Error getting a patch helper to update this resource")
-				return
+				log.Info("BackupStorageLocation is invalid, marking as unavailable")
+				err = errors.Wrapf(err, "BackupStorageLocation %q is unavailable", location.Name)
```
reasonerjt, line 115: Why not also log this error?

pkg/controller/backup_storage_location_controller.go
```diff
@@ -53,96 +60,88 @@ type BackupStorageLocationReconciler struct {
...
+	var location velerov1api.BackupStorageLocation
 
-	log.Debug("Validating availability of backup storage locations.")
+	log := r.Log.WithField("controller", BackupStorageLocation).WithField(BackupStorageLocation, req.NamespacedName.String())
+	log.Debug("Validating availability of BackupStorageLocation")
 
 	locationList, err := storage.ListBackupStorageLocations(r.Ctx, r.Client, req.Namespace)
 	if err != nil {
-		log.WithError(err).Error("No backup storage locations found, at least one is required")
+		log.WithError(err).Error("No BackupStorageLocations found, at least one is required")
```
reasonerjt, line 71: I don't think you need to return the error in L72.

In a similar case like following, the error is not returned:
```
	if location.Name == "" || location.Namespace == "" {
		log.WithError(err).Error("BackupStorageLocation is not found")
		return ctrl.Result{}, nil
	}
```

Please also check the behavior of the controller if an error is returned, I thought it would trigger the `Reconcile` again.

blackpiglet: Checked controller-runtime code, when reconcile function return error, the req key will be re-enter queue after that:
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L322-L326 

``` go
	result, err := c.Reconcile(ctx, req)
	switch {
	case err != nil:
		c.Queue.AddRateLimited(req)
		ctrlmetrics.ReconcileErrors.WithLabelValues(c.Name).Inc()
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelError).Inc()
		log.Error(err, "Reconciler error")
```

pkg/controller/backup_storage_location_controller.go
blackpiglet: Modified.

reasonerjt (approved):

ywk253100 (approved):

PR #4919, opened 2022-05-16
Update cluster migration scenario with more details

### LINKED ISSUE #2167: Cluster Migration Documentation not clear - Unusable

The documentation link on cluster migration here:

https://velero.io/docs/master/migration-case/

a) Does not indicate how to specify the new backup location when restoring a backup from the newly created backup location that points to the-cluster-that-must-be-migrated. Should it be passed as an argument to the "velero create restore from <backup> ...." command?

Because of this the attempt to migrate the cluster stops at the point where the second backup location pointing to the source cluster's container is created. There does not seem to be a way to indicate that the backup desired must come from the newly created backup location.

b) There does not seem to be clear documentation on specifying additional backup locations for each cloud vendor (Azure in my case). This makes the migration process even more uncertain.

Note: This test is being carried out on azure cloud.

- Demonstrate using exact commands in the documentation how the second backup location is created, including the authentication configuration, for each of the major cloud vendors (Azure, AWS, etc ...)
- Demonstrate the exact command(s) use to restore a backup from a particular storage location.

The velero command to restore from a particular backup doesn't seem to include an option to specify the exact backup location (container) from which the backup must be extracted. It's hard to understand how the documented migration process will work without this feature.

Some clarification would be appreciated here.

**Environment:**

v1.1.0

1.14.8

AKS 1.14.8

Azure, Azure blob storage

Ubuntu Linux

a-mccarthy:

This updates the cluster migration scenario docs to include more details about how to set up clusters to data migration.

Fixes #2167 

### DIFF
--- a/site/content/docs/main/migration-case.md
+++ b/site/content/docs/main/migration-case.md
@@ -58,8 +58,8 @@ This scenario steps through the migration of resources from Cluster 1 to Cluster
     velero backup-location create bsl --provider aws --bucket velero-migration-demo --config region=us-east-2 --access-mode=ReadOnly
     ```
 
-    Make sure to configure the `BackupStorageLocations` as read-only
-    by using the `--access-mode=ReadOnly` flag for `velero backup-location create`. See `velero backup-location –help` for more information about the available flags for this command.
+    Its recommended that you configure the `BackupStorageLocations` as read-only
+    by using the `--access-mode=ReadOnly` flag for `velero backup-location create`. This will make sure that the backup is not deleted from the object store by mistake during the restore. See `velero backup-location –help` for more information about the available flags for this command.
 
     ```
     velero snapshot-location create vsl --provider aws --config region=us-east-2
--- a/site/content/docs/main/migration-case.md
+++ b/site/content/docs/main/migration-case.md
@@ -3,52 +3,85 @@ title: "Cluster migration"
 layout: docs
 ---
 
-Velero can help you port your resources from one cluster to another, as long as you point each Velero instance to the same cloud object storage location.
+Velero's backup and restore capabilities make it a valuable tool for migrating your data between clusters, helping you perform many common workflows such as upgrading Kubernetes versions or disaster recovery. Cluster migration with Velero is based on Velero's [object storage sync](how-velero-works.md#object-storage-sync) functionality, which is responsible for syncing Velero resources from your designated object storage to your cluster. This means that to perform cluster migration with Velero you must point each Velero instance running on clusters involved with the migration to the same cloud object storage location.
 
-## Migration Limitations
+This page outlines a cluster migration scenario and some common configurations you will need to start using Velero to begin migrating data.
+
+## Before migrating your cluster
 
 Before migrating you should consider the following,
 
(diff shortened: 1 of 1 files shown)


site/content/docs/main/migration-case.md
```diff
@@ -3,52 +3,85 @@ title: "Cluster migration"
...
+
+    The default backup retention period, expressed as TTL (time to live), is 30 days (720 hours); you can use the `--ttl <DURATION>` flag to change this as necessary. See [how velero works](how-velero-works.md#set-a-backup-to-expire) for more information about backup expiry.
+
+1. On Cluster 2, make sure that Velero is installed. Note that the install command below has the same `region` and `--bucket` location as the install command for Cluster 1. The Velero plugin for AWS does not support migrating data between regions.
+
+    ```
+    velero install --provider aws --image velero/velero:v1.8.0 --plugins velero/velero-plugin-for-aws:v1.4.0 --bucket velero-migration-demo --secret-file xxxx/aws-credentials-cluster2 --backup-location-config region=us-east-2 --snapshot-location-config region=us-east-2
+    ```
+
+    Alternatively you could configure `BackupStorageLocations` or `VolumeSnapshotLocations` after installing Velero on Cluster 2, pointing to the `--bucket` location used by Cluster 1. To do this you can use to `velero backup-location create` and `velero snapshot-location create` commands.
```
a-mccarthy, line 55: @reasonerjt Should I remove the VolulmneSnapShot information? This is the only place that its mentioned in this doc, but i'm not sure if the process would be the same. thanks!

blackpiglet (dismissed):

site/content/docs/main/migration-case.md
reasonerjt: The `VolumeSnapshotLocation` should be created, thanks for asking.

reasonerjt:
I noticed there are several instances where you refer to `backup` as `back up`, is it intentional?

site/content/docs/main/migration-case.md
```diff
@@ -3,52 +3,85 @@ title: "Cluster migration"
...
 
     ```
     velero backup describe <BACKUP-NAME>
     ```
 
-    **Note:** The default sync interval is 1 minute, so make sure to wait before checking. You can configure this interval with the `--backup-sync-period` flag to the Velero server.
+    Velero resources are [synchronized](how-velero-works.md#object-storage-sync) with the backup files in object storage. This means that the Velero resources created by Cluster 1's backup will be synced to Cluster 2 through the shared Backup Storage Location. Once the sync occurs, you will be able to access the back up from Cluster 1 on Cluster 2 using Velero commands. The default sync interval is 1 minute, so you may need to wait before checking for the back up's availability on Cluster 2. You can configure this interval with the `--backup-sync-period` flag to the Velero server on Cluster 2.
 
-1.  *(Cluster 2)* Once you have confirmed that the right Backup (`<BACKUP-NAME>`) is now present, you can restore everything with:
+1.  On Cluster 2, once you have confirmed that the right back up is available, you can restore everything to Cluster 2.
```
reasonerjt, line 72: Is it `backup` or `back up`?

There are several other instances like this in this PR.

site/content/docs/main/migration-case.md
reasonerjt: that is correct.

reasonerjt (approved):
LGTM

blackpiglet (approved):

PR #5047, opened 2022-06-24
allow excluding all volumes of a pod from restic

### LINKED ISSUE #4360: Allow excluding all volumes of a pod from restic
We have some pods like the prometheus instances created by the operator that have generated volume names, we would like to exclude those from restic backups.

Add a new annotation to skip all volumes of a pod for a restic backup.

**Vote on this issue!**

elchenberg:
Allow excluding all volumes of a pod from restic using the `backup.velero.io/backup-volumes-excludes` annotation with a `*` wildcard.

Closes #4360

Signed-off-by: elchenberg <elchenberg@users.noreply.github.com>


pkg/restic/common.go
```diff
@@ -214,7 +214,7 @@ func GetPodVolumesUsingRestic(pod *corev1api.Pod, defaultVolumesToRestic bool) [
 			continue
 		}
 		// don't backup volumes that are included in the exclude list.
-		if contains(volsToExclude, pv.Name) {
+		if contains(volsToExclude, "*") || contains(volsToExclude, pv.Name) {
```
qiuming-best, line 217: what if backup.velero.io/backup-volumes-excludes is "a*" or "pva, pvb, *"

maybe it is more suitable to only be allowed to set it to "*" if using a wildcard?

maybe you can add more unit tests about the above situation

elchenberg: > maybe it is more suitable to only be allowed to set it to "*" if using a wildcard?

If you want I can do it. But it would require more code (that needs to be maintained in the future) to validate the `volsToExclude` and some error handling if there is a mix of named volumes and a wildcard. Currently `GetPodVolumesUsingRestic` does not return any errors so the interface would need to change (unless we decide to ignore errors).

> what if backup.velero.io/backup-volumes-excludes is "a*" or "pva, pvb, *" [...] maybe you can add more unit tests about the above situation

In the second example `pva` and `pvb` will be excluded from the backup (as intended) and with the current implementation of the `contains` function `a*` would not match `*`:

```golang
func contains(list []string, k string) bool {
	for _, i := range list {
		if i == k {
			return true
		}
	}
	return false
}
```

But I agree that there should be a unit test for `a*`. :slightly_smiling_face:

qiuming-best: @elchenberg Thank you for submitting the code.

Maybe I think we should return an error when GetPodVolumesUsingRestic handles with an unexpected wildcard like `a*`, and the error should append into `backupErrs` in `backupItem` function

elchenberg: @qiuming-best Are you asking me? I'm not familiar with the internals of the application so I am hesitant to decide on changes to the function signature.

varac:
Very interested in this feature - except mainly for the opposite: Including all volumes from a Pod.
Use case: Backup all volumes of a Pod created by a cloudnative-postgresql operator, where Volumes have a non-determistic naming scheme.

PR #5089, opened 2022-07-04
Reduce CRD size.

### LINKED ISSUE #4916: Reduce Restore CRD's size
In all of Velero's CRDs, Restore's CRD yaml size is the biggest, due to containing InitContainer definition in it. kube-APIServer cannot handle big CRD definition well up to now, especially for update case, and big CRD is challenge for slow speed network environment, too.

This is the log collected for `kubectl apply -f restores.yaml`

```
I0513 02:06:53.290651       1 client.go:360] parsed scheme: "passthrough"
I0513 02:06:53.290973       1 passthrough.go:48] ccResolverWrapper: sending update to cc: {[{https://127.0.0.1:2379/  <nil> 0 <nil>}] <nil> <nil>}
I0513 02:06:53.291051       1 clientconn.go:948] ClientConn switching balancer to "pick_first"
E0513 02:07:16.773448       1 writers.go:117] apiserver was unable to write a JSON response: http: Handler timeout
E0513 02:07:16.774094       1 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s:"http: Handler timeout"}: http: Handler timeout
E0513 02:07:16.775425       1 writers.go:130] apiserver was unable to write a fallback JSON response: http: Handler timeout
I0513 02:07:16.776996       1 trace.go:205] Trace[1983387252]: "Create" url:/apis/apiextensions.k8s.io/v1/customresourcedefinitions,user-agent:kubectl/v1.23.4 (linux/amd64) kubernetes/e6c093d,client:172.20.10.83,accept:application/json,protocol:HTTP/2.0 (13-May-2022 02:06:24.797) (total time: 51978ms):
Trace[1983387252]: ---"About to convert to expected version" 51960ms (02:07:00.758)
Trace[1983387252]: [51.97873252s] [51.97873252s]
```

Detail conversation about this topic can be found here:
https://vmware.slack.com/archives/CEX5KS8NP/p1652350274022409

Need to consider how to reduce the restore CRD's size.

**Environment:**

**Vote on this issue!**

blackpiglet:
The background of this change is Velero has some use cases that k8s cluster operated after proxy or air-gapped environment, so the network speed is limited. This makes Velero installation often failed with CRD creation timeout, and the biggest CRD of Velero is `Restore`, because the InitContainer hook contains `PodSpec`. Restore CRD size is nearly 2MB.

This is a proposal to reduce CRD size. Reduction is achieved by two modifications:
* Remove all descriptions in CRDs.
* Change Restore's InitContainter in Hook from corev1.PodSpec to runtime.RawExtension.

Not sure whether users and developers of Velero relied on the things changed. If anything is not acceptable, I'm open to change.

1. remove descriptions in CRDs.
2. Make the Restore hook.InitConatianer server side field pruing disable.
3. Remove restore patch in update-generate-crd-code.sh.
4. Modify related testcases.

Fixes #4916 


reasonerjt:
A problem of removing the description is that it breaks `kubectl explain`

blackpiglet:
Right. `kubectl explain` should depend on the CRD definition YAML file. Most information is lost after this change.
Need to check whether this is acceptable for users.

reasonerjt:
I believe updating the pod.spec should be sufficient?

blackpiglet:
> I believe updating the pod.spec should be sufficient?

In most cases, I think yes, but I'm not sure about Fedex NSX environment. I remember there are more than one CRD failed during creation. Need to test to make sure.

shubham-pampattiwar:
I agree with @reasonerjt, IMO breaking `kubectl explain` is not a good idea. Maybe try to make the descriptions more concise ? Unsure how much that can help.

blackpiglet:
I think there are two ways:
* Make the description shorter.
* Limit the description size in generated CRD.

Maybe we can do both.

blackpiglet:
@shubham-pampattiwar @reasonerjt 
As discussion, I add the description of CRDs backup. Please take a look.

shubham-pampattiwar:
@blackpiglet Are we not going to shorten the Restore CRD spec descriptions ? Do the current changes in the PR fix the size issue ?

blackpiglet:
@shubham-pampattiwar 
After discussion with @reasonerjt, we agreed that starting at reducing CRD size of restores.velero.io. If there is still issue, then try to shorten the descriptions in CRD.

shubham-pampattiwar:
> @shubham-pampattiwar After discussion with @reasonerjt, we agreed that starting at reducing CRD size of restores.velero.io. If there is still issue, then try to shorten the descriptions in CRD.

Ok, Thank you ! VISACK !

shubham-pampattiwar (approved):

internal/hook/item_hook_handler_test.go
```diff
@@ -1911,9 +1893,9 @@ func TestHandleRestoreHooks(t *testing.T) {
 					RestoreHooks: []velerov1api.RestoreResourceHook{
 						{
 							Init: &velerov1api.InitRestoreHook{
-								InitContainers: []corev1api.Container{
-									*builder.ForContainer("restore-init-container-1", "nginx").
-										Command([]string{"a", "b", "c"}).Result(),
+								InitContainers: []runtime.RawExtension{
+									builder.ForContainer("restore-init-container-1", "nginx").
+										Command([]string{"a", "b", "c"}).ResultRawExtension(),
```
reasonerjt, line 1898: Add a case for invalid container spec in restore spec?

pkg/controller/restore_controller.go
```diff
@@ -324,6 +326,22 @@ func (c *restoreController) validateAndComplete(restore *api.Restore, pluginMana
 		return backupInfo{}
 	}
 
+	// validate Restore Init Hook's InitContainers
+	restoreHooks, err := hook.GetRestoreHooksFromSpec(&restore.Spec.Hooks)
+	if err != nil {
+		restore.Status.ValidationErrors = append(restore.Status.ValidationErrors, err.Error())
+	}
+	for _, resource := range restoreHooks {
```
reasonerjt, line 334: risk of nil pointer when error happens in `GetRestoreHooksFromSpec`

pkg/controller/restore_controller.go
```diff
@@ -682,3 +700,17 @@ func (l *restoreLogger) done(log logrus.FieldLogger) (io.Reader, error) {
 func (l *restoreLogger) closeAndRemove(log logrus.FieldLogger) {
 	closeAndRemoveFile(l.file, log)
 }
+
+// validateContainer validate whether a map contains mandatory k8s Container fields.
+// mandatory fields include name, image and commands.
+func validateContainer(raw []byte) error {
```
reasonerjt, line 706: UT?

pkg/controller/restore_controller.go
blackpiglet: restoreHooks's type is []ResourceRestoreHook.
If `GetRestoreHooksFromSpec` returns an empty array, I think this loop will not be entered, so I think there should be no nil pointer risk.

reasonerjt: It does return nil
https://github.com/vmware-tanzu/velero/blob/52504b548dd63d249f7ebceebc6a8c9ab95b4a66/internal/hook/item_hook_handler.go#L401

But you are right it won't panic.

blackpiglet: Got it. I will change return value to empty array.

pkg/controller/restore_controller.go
blackpiglet: Added

internal/hook/item_hook_handler_test.go
blackpiglet: Added.

blackpiglet:
Conflict by crd.go file. If the change is accept, I will rebase the branch and create a new PR for merge.

reasonerjt (approved):

blackpiglet:
Close this one, due to `crds.go` file conflicting with main branch.
Open #5174 to resolve conflict.

PR #4416, opened 2021-12-01
Added UploadProgressFeature flag to enable UploadProgressMonitoring

dsu-igeek:
and ItemSnapshotters.  Will be required in 1.8 to use the upload progress monitoring or item snapshotters, we will plan to remove in 1.9.

Signed-off-by: Dave Smith-Uchida <dsmithuchida@vmware.com>

Thank you for contributing to Velero!

# Please add a summary of your change

# Does your change fix a particular issue?

Fixes #(issue)

# Please indicate you've done the following:

- [ ] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [ ] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.
- [ ] Updated the corresponding documentation in `site/content/docs/main`.

PR #4077, opened 2021-08-30
ItemSnapshotter Plugin APIs

dsu-igeek:
ItemSnapshotter plugin APIs.
Added ItemSnapshotter.proto
Added item_snapshotter Go interface
Added framework components for item_snapshotter
Updated plugins doc with ItemSnapshotter info
Addresses #3754

# Please indicate you've done the following:

- [X] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [X] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.
- [X] Updated the corresponding documentation in `site/content/docs/main`.

PR #3805, opened 2021-05-14
Added BackupPhaseUploading and BackupPhaseUploadingPartialFailure backup

dsu-igeek:
Added BackupPhaseUploading and BackupPhaseUploadingPartialFailure backup phases as part of Upload Progress Monitoring, fixes #3755 Add backup phases needed for Upload Progress Monitoring

Signed-off-by: Dave Smith-Uchida <dsmithuchida@vmware.com>

# Please indicate you've done the following:

- [ C] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [ C] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required`.
- [C ] Updated the corresponding documentation in `site/content/docs/main`.

PR #5108, opened 2022-07-09
Item action progress monitoring design

sseago:
This design combines the requirements for the previously-merged
Upload Progress Monitoring design with the requirements for the
(not submitted but discussed in meetings and slack) proposed asynchronous
item action plugins into one integrated proposal.

This is a design PR that is intended to combine the needs of the previous Upload Progress Monitoring (https://github.com/vmware-tanzu/velero/blob/main/design/upload-progress.md ) and the recently-discussed Asynchronous backup/restore actions into a single unified design. This design document should supercede both of the above.

Currently a Draft PR, since I still need to read through the existing upload progress implementation PR (https://github.com/vmware-tanzu/velero/pull/4467 ) to see whether there's anything I may have missed here. Also, the backup phase diagram needs to be updated, as it currently uses the earlier snapshot-specific terminology rather than the more general terms used here.


design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
+# Backup and Restore Item Action Progress Monitoring
+
+This is intended as a replacement for the previously-approved Upload Progress Monitoring design
+([Upload Progress Monitoring](upload-progress.md)) in order to expand the supported use cases beyond
+snapshot uploads to include what was previously called Async Backup/Restore Item Actions. This
+updated design should handle the combined set of use cases for those previously separate designs.
+
+Volume snapshotter plug-in are used by Velero to take snapshots of persistent volume contents.
```
blackpiglet, line 8: Maybe BackupItemAction should be mentioned here, because CSI plugin uses BIA to take snapshot.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+- Data Movers
+    - Data movers are asynchronous processes that are triggered via backup/restore of kubernetes resources. A common use case for this is the backup/restore of PVCs whose data we want to move to some form of backup storage.
+    - Workflow
+        - User takes velero backup of PVC A
+        - BIA plugin applies to PVCs with compatible storage driver
+        - BIA plugin triggers data mover (for example via creation of a custom resource)
+        - BIA plugin returns
+        - Velero backup process continues
+        - Main velero backup process monitors running BIA threads via gRPC to determine if process is done and healthy
+    - BIA/RIA Triggering creation of a custom resource
```
blackpiglet, line 47: Not sure whether I understand the two BIA/RIA related items correct.
Looks like those (BIA/RIA Triggering creation of a custom resource, ) are two ways of implementation of how plugins trigger data mover and check progress.
If so, please arrange the the structure of this list. Looks a little confusing.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+
+## Non-goals
+- Today, Velero is unable to recover from an in progress backup when the velero server crashes (pod is deleted). This has an impact on running asynchronous processes, but it’s not something we intend to solve in this design.
+
+## Models
+
+### Internal configuration and management
+In this model, movement of the snapshot to stable storage is under the control of the snapshot
+plug-in.  Decisions about where and when the snapshot gets moved to stable storage are not
+directly controlled by Velero.  This is the model for the current VolumeSnapshot plugins.
```
blackpiglet, line 98: Do you mean this is the model of previous proposed ItemSnapshotPlugin?

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+
+### Velero controlled management
+In this model, the snapshot is moved to external storage under the control of Velero.  This
```
blackpiglet, line 101: The two models both have some advantages.
Velero controlled model can separate upload from snapshotting. Velero can have a ad-hoc control of when to start upload for all snapshots.
Plugin internal model makes Velero more simpler, and Velero doesn't need to care about the data mover logic, but by introducing Progress as CRD, the provider specific data mover logic can be moved to plugins's Progress method, and Velero can use a new controller to watch over the Progress CRD status.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+plug-in.
+
+In the current implementation, Restic backups will move data during the "InProgress" phase.  In the
+future, it may be possible to combine a snapshot with a Restic (or equivalent) backup which would
+allow for data movement to be handled in the "WaitingForPluginOperations" phase,
+
+The next phase is either "Completed", "WaitingForPluginOperations", "Failed" or "PartiallyFailed".
+Backups/restores which would have a final phase of "Completed" or "PartiallyFailed" may move to the
+"WaitingForPluginOperations" or "WaitingForPluginOperationsPartialFailure" state.  A backup/restore
+which will be marked "Failed" will go directly to the "Failed" phase.  Uploads may continue in the
```
blackpiglet, line 155: I prefer not to start uploading before all snapshots are completed, because many providers don't have a API to stop in-progress snapshotting, and if backup or restore failed, new backup and restore can start. It's possible that new backup and restore also require to snapshot or restore the the same disk. If there is a in-progress operation of the disk, I think there's high possibility that error would be raised.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+VolumeSnapshot plugins will return operation IDs (snapshot IDs or other plugin-specific
+identifiers).  The plugin should be able to provide status on the progress for the snapshot and
+handle cancellation of the operation/upload if the snapshot is deleted.  If the plugin is restarted,
+the operation ID should remain valid.
+
+When all snapshots have been taken and Kubernetes resources have been persisted to the ObjectStorePlugin
+the backup will either have fatal errors or will be at least partially usable.
+
+If the backup/restore has fatal errors it will move to the "Failed" state and finish. If a
+backup/restore fails, the upload or other operation will not be cancelled but it will not be
```
blackpiglet, line 200: As previous comment, my preference is not starting upload in the case, not sure about other users or maintainers' opinion.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+monitored either.  For backups in any phase, all snapshots will be deleted when the backup is
```
blackpiglet, line 201: I'm not about the wording "in any phase". To me, if backups are in some phases, e.g. InProgress, it's not easy to handle.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+deleted.  Plugins will cancel any data movement or other operations and remove snapshots and other
+associated resources when the VolumeSnapshotter DeleteSnapshot method or DeleteItemAction Execute
+method is called.
+
+Velero will poll the plugins for status on the operations when the backup/restore exits the
+"InProgress" phase and has no fatal errors.
+
+If any operatios are not complete, the backup/restore will move to either WaitingForPluginOperations
```
blackpiglet, line 209: Could you give more details on the Phase transition trigger from InProgress to WattingForPluginOperations or WaitingForPluginOperationsPartialFailure.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+or WaitingForPluginOperationsPartialFailure or Failed.
+
+Post-snapshot and other operations may take a long time and Velero and its plugins may be restarted
+during this time.  Once a backup/restore has moved into the WaitingForPluginOperations or
+WaitingForPluginOperationsPartialFailure phase, another backup/restore may be started.
+
+While in the WaitingForPluginOperations or WaitingForPluginOperationsPartialFailure phase, the
+snapshots and item actions will be periodically polled.  When all of the snapshots and item actions
+have reported success, the backup/restore will move to the Completed or PartiallyFailed phase,
```
blackpiglet, line 218: Does backup transit from WaitingForPluginOperationsPartialFailure to PartiallyFailed?

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+depending on whether the backup/restore was in the Uploading or UploadingPartialFailure phase.
```
blackpiglet, line 219: Need to change ** Uploading** and ** UploadingPartialFailure** to ** WaitingForPluginOperations** and ** WaitingForPluginOperationsPartialFailure**.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+
+The Backup resources will not be written to object storage until the backup has entered a final phase:
```
blackpiglet, line 221: If backup already reaches to WaitingForPluginOperations or WaitingForPluginOperationsPartialFailure, part of the backup should be usable, but the if the Velero or plugin restarted, and the backup resources are not in object store, then the backup cannot be used.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+### VolumeSnapshotter changes
+
+Two new methods will be added to the VolumeSnapshotter interface:
+
+    Progress(snapshotID string) (UploadProgress, error)
+    Cancel(snapshotID string) (UploadProgress, error)
+
+Open question: Does VolumeSnapshotter need Cancel, or is that only needed for BIA/RIA?
```
blackpiglet, line 255: VolumeSnapshotter may not need a **Cancel** method, because there is no Cancel API for most of the snapshot provisioner, and VolumeSnapshotter plugin only focus on snapshot. There should be no potential need to add a more generic method for other kinds of operations.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+### BackupItemAction and RestoreItemAction plug-in changes
+
+Currently CSI snapshots and the Velero Plug-in for vSphere are implemented as BackupItemAction
+plugins.  While the majority of BackupItemAction plugins do not take snapshots or upload data, this
+functionality is useful for any longstanding plugin operation managed by an external
+process/controller so we will modify BackupItemAction and RestoreItemAction to optionally return an
+operationID in addition to the modified item.
+
+The Backup/RestoreItemAction plugin identifier as well as the Item and Snapshot ID will be stored in
+the `<backup-name>-itemoperations.json.gz`.  When checking for progress, this info will be used to
```
blackpiglet, line 273: Agree with creating a new metadata file on object store for Plugin operations, but I'm a little confused about when it would be used for checking backup or restore progress?
There should be other ways to trace which plugins should be polled.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+select the appropriate Backup/RestoreItemAction plugin to query for progress.
+
+Velero can attempt to cancel an operation by calling the Cancel API call on the BIA/RIA. The plugin
```
blackpiglet, line 276: It sounds **Cancel** method more like deletion. This is not used for cancelling a backup or restore, right?

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+    	// Item is the item that was stored in the backup
+    	Item runtime.Unstructured
+    	// OperationID is the operation ID returned by BackupItemAction Execute
+    	operationID string
+    	// Backup is the representation of the backup resource processed by Velero.
+    	Backup *velerov1api.Backup
+    }
+
+Two new methods will be added to the RestoreItemAction interface, and the
+RestoreItemActionExecuteOutput struct will be modified:
```
blackpiglet, line 313: RestoreItemActionExecuteOutput should be changed to RestoreItemAction interface.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+        // value is true, AdditionalItems will be ignored.
+        SkipRestore bool
+
+        // OperationID is an identifier which indicates an ongoing asynchronous action which Velero will
+        // continue to monitor after restoring this item. If left blank, then there is no ongoing operation
+        OperationID string
+    }
+
+    // RestoreItemProgressInput contains the input parameters for the RestoreItemAction's Progress function.
+    type RestoreItemProgressInput struct {
```
blackpiglet, line 351: RestoreItemProgressInput  should be changed toRestoreItemActionProgressInput

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+    	Restore *velerov1api.Restore
+    }
+
+## Changes in Velero backup format
+
+No changes to the existing format are introduced by this change.  A
+`<backup-name>-itemoperations.json.gz` file will be added that contains the items and operation IDs
+(snapshotIDs) returned by VolumeSnapshotter and BackupItemAction plugins.  Also, the creation of the
+`velero-backup.json` object will not occur until the backup moves to one of the terminal phases
+(_Completed_, _PartiallyFailed_, or _Failed_).  Reconciliation should ignore backups that do not
```
blackpiglet, line 366: As previous comments, I prefer to filter backups by Phase, not the object file

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,462 @@
...
+
+A `<restore-name>-itemoperations.json.gz` file will be added that contains the items and operation IDs
+returned by RestoreItemActions.
+
+## CSI snapshots
+
+For systems such as EBS, a snapshot is not available until the storage system has transferred the
+snapshot to stable storage.  CSI snapshots expose the _readyToUse_ state that, in the case of EBS,
+indicates that the snapshot has been transferred to durable storage and is ready to be used.  The
+CSI BackupItemAction.Progress method will poll that field and when completed, return completion.
```
blackpiglet, line 390: There will be no percentage progress returned in this way.

blackpiglet:
Another thing I'm not clear about the design is how BIA/RIA interact with data mover. Is that part will be covered in this PR?

design/general-progress-monitoring.md
sseago: @blackpiglet That's right -- this was an attempt to describe different ways a plugin could implement an asynchronous action. And yes, looking at this now, it's slightly confusing -- I'll rework this.

design/general-progress-monitoring.md
sseago: This was Dave's wording that was pulled into the combined design, but this is referring to the current VolumeSnapshotter implementation.

design/general-progress-monitoring.md
sseago: I think support for canceling backups and restores is out of scope for this work. The intent here is just to clarify that when a backup is deleted snapshots should also be deleted. "In any phase" just means that if we're already deleting a backup, snapshots should be removed as well.

design/general-progress-monitoring.md
sseago: Yes. The transition is from WaitingForPluginOperationsPartialFailure to PartiallyFailed once all operations/uploads have completed, and from WaitingForPluginOperations to Completed (if no operations/uploads error out), or from WaitingForPluginOperations to PartiallyFailed (if some operations error out).

design/general-progress-monitoring.md
sseago: I had a similar concern when I first read this in the original design. I'm open to revisiting this -- I wonder whether @dsu-igeek would be able to comment on this aspect here.

design/general-progress-monitoring.md
sseago: I think the main reason for putting it in this file (vs. just an in-memory map) is that Velero should be able to recover from a restart *if* the restart happens after the backup/restore is put into the Waiting... phase. This file would be used when Velero needs to check for progress again (to potentially move backup/restore to Completed). It could certainly be cached in memory somewhere for subsequent polling.

design/general-progress-monitoring.md
sseago: RestoreItemActionExecuteOutput is the return type of the Execute function in the RestoreItemAction interface. We need to add a new field to this struct, so the above is correct.

design/general-progress-monitoring.md
sseago: Looking at this again, I'm not actually 100% sure what @dsu-igeek meant here in the original design. Once we have that file, (unless we change things in response to the earlier suggestion  to write it sooner), the backup is already completed, so I'm not sure why this means that the controller reconcile would ignore backups without the file -- in fact, the opposite might be true.

reasonerjt:
@sseago To follow up the discussion in community mtg, I've found the following PRs merged for the itemsnapshotter plugin proposed by Dave:

https://github.com/vmware-tanzu/velero/pull/4429
https://github.com/vmware-tanzu/velero/pull/4417
https://github.com/vmware-tanzu/velero/pull/4416
https://github.com/vmware-tanzu/velero/pull/4077
https://github.com/vmware-tanzu/velero/pull/3805

shubham-pampattiwar:
> Another thing I'm not clear about the design is how BIA/RIA interact with data mover. Is that part will be covered in this PR?

@blackpiglet The proposed plugins in this design will act as the building blocks for Datamover. Basically, this will enable datamover implementation, datamover is just one use case which will be enabled by this design proposal.

design/general-progress-monitoring.md
sseago: @blackpiglet I'm adding this to the open questions at the end. It's a good question, and we want to be clear on which approach is best before implementing this.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,469 @@
...
+In the current implementation, Restic backups will move data during the "InProgress" phase.  In the
+future, it may be possible to combine a snapshot with a Restic (or equivalent) backup which would
+allow for data movement to be handled in the "WaitingForPluginOperations" phase,
+
+The next phase is either "Completed", "WaitingForPluginOperations", "Failed" or "PartiallyFailed".
+Backups/restores which would have a final phase of "Completed" or "PartiallyFailed" may move to the
+"WaitingForPluginOperations" or "WaitingForPluginOperationsPartialFailure" state.  A backup/restore
+which will be marked "Failed" will go directly to the "Failed" phase.  Uploads may continue in the
+background for snapshots that were taken by a "Failed" backup/restore, but no progress will not be
+monitored or updated.  When a "Failed" backup is deleted, all snapshots will be deleted and at that
```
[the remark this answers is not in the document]
sseago: So this is another part that I inherited from the original design. I suspect that it was because native snapshot uploads can't really be cancelled in progress. However, it might make sense to alter the design here to say that for Backup/RestoreItemActions in progress, the Cancel() API function will be called. Note that Cancel() isn't a guarantee, but if the plugin has a way to cancel the action, it will. What do others think about this change?

design/general-progress-monitoring.md
sseago: @Lyndon-Li Regarding those cases you mention, I suspect most of those would be individual implementation details for specific plugins, rather than part of the snapshot API. But to briefly go through the four use cases:
- What to do with local snapshot data is something that the current volume snapshotter plugins deal with. The API addition here is just to allow monitoring of the upload (which is also already happening in the current plugin). "Provide for users to decide not to keep specific snapshots" doesn't really seem like something we'd want to add to the plugin interface (or if we do, perhaps as a separate later enhancement).
- For volume snapshotters, upload is an integral part of the process - this enhancement just helps to track the (already existing) upload process. For Backup/RestoreItemActions, if there is a specific plugin that wants to perform asynchronous operations conditionally, that is provided for already in this design -- the operation ID returned by the Backup/RestoreItemAction is optional, and if a plugin does not return one, then there is no operation/upload action to monitor.
- Regarding retention policies, again if a plugin wants to alter behavior based on those, it can, but I don't really see a place in the API where we'd want to deal with it -- if we find something to add, it can be done as a future enhancement.
- I'm not sure I see the use case for "don't actually upload the snapshot now" -- we don't offer that for backups as a whole -- a backup is not marked complete until the entire backup is uploaded to object storage (that way if the cluster goes down/crashes/etc. post-backup-completion, it can be recovered right then). The goal of this enhancement is to provide the same for snapshots and other long-running operations initiated from within a plugin action.

Questions:
- The first question is answered above for the four use cases. From Velero's point of view, the upload process is simply something that a plugin initiates, and Velero's role is limited to asking the plugin periodically whether the upload or other operation is now complete. Velero isn't supposed to be able to understand AWS-specific aspects of snapshot upload, or OpenShift-specific aspects of an OpenShift image plugin that copies internal images into or out of an external backup store. It's why we have plugins rather than building all of these operations into velero core.
- The way we make them generic (as described above) is that the operation itself isn't part of the interface. The BIA/RIA Execute() method call has an optional operationID return parameter. If there is no ongoing operation, it's empty. All velero needs to know is to call the plugin again to check on status and pass back the ID. As for configuration, note that the existing BIA/RIA plugins don't involve plugin-specific configuration in the API. That's already how plugins work. We're just extending the plugin interface to allow ongoing operations. So continuing not to support non-generic plugin-specific configurations in the interface is consistent with the current API. If a plugin needs additional information, it is free to document that users should pass this information on via annotations on the resources being backed up or on the velero Backup/Restore CRs, as is currently done by some of the Openshift-specific plugins included in OADP. If we want to add the notion of plugin configuration to the plugin API, that should be a separate future enhancement if there is a need for it.

design/general-progress-monitoring.md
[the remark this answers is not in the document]
blackpiglet: I had similar concern before, but currently, backup only fails on validation stage before calling Plugins, so technically there should be no such case of failed backup with uploading or operation in background.
But we need to make sure this pattern will not be broken in future.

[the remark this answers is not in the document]
Lyndon-Li: Personally, I also vote on the change for this. Even though Velero doesn't reach the "failed" phase for now, from a common concept, the upload for a "failed" backup is meaningless. Suppose that we have 1TB of data to be uploaded, when a backup fails, Velero may start another backup shortly (maybe not now, but it is also a good way to handle a failed backup, otherwise, there will be recovery point miss), if we leave the upload for the failed backup and also start a new one, there will be double disk IO, double network IO, double cost (if any), and double or much longer completion time.
On the other hand, since Velero doesn't have the "failed" phase, we probably don't need to cover details of this "failed-cancel" in the design, just change the expression in this failed scenario.

design/general-progress-monitoring.md
Lyndon-Li: @sseago Yes, I agree that Velero should not understand details of the specific plugins and the cases mentioned above should be implemented by the specific plugins.

As another fact, we need to let the specific plugin know some information so that the plugin could do the actions. If we regard this information as some "configuration" as well, my above questions come down to: how do we share these configuration between Velero and the plugins if the BIA/RIA execution is totally transparent to Velero?
Specifically, The configuration may not be simply user options that could be fulfilled by annotations. For example, whether or when to start an upload/whether to delete the local snapshot after upload succeeds is decided by a combination of below factors:

1. Users explicitly specify how many local snapshot they want to keep. If the current snapshot count is within that number, the local snapshot created by BIA should not be deleted
2. Users explicitly specify how much time later they want to start the upload after snapshot is taken/backup is created
3. Users specify they just want to upload some of the snapshots, not all. For example, the daily/weekly snapshots. Then Velero needs to decide this dynamically according to its retention policy
4. Considering the time of the backup runs, Velero thinks it is not a suitable time to run the upload. For example, the production environment is busy at most of the day time, if Velero runs the upload, it brings a lot burden to the disk, network and may impact the business. Therefore, there is usually a backup window (i.e. from mid night to the next morning), if the upload could not be finished in the window, Velero needs to have a policy (give up the upload or schedule to another time??). This means, Velero needs to decide the time to start the upload according to its schedule dynamically

As you can see above, if 1 and 2 could be fulfilled by a static annotation, the later two ones is decided by Velero dynamically. Then how do Velero and the plugin share this information.

Indeed, the Velero part of the upload policy should not be covered in this design, merely, I think we do need to clarify how do Velero and plugins exchange information/configuration of the upload policies.

Lyndon-Li: For sure, as another approach, we can have all these judgements and policy implementation in the specific plugins, then Velero only needs to share the basic & generic things, i.e., retention policies, schedule policies, etc.
The problem with the approach is, there will be duplicated logics in many plugins to implement the policies, and on the other hand, it may be a better way to implement this in Velero centrally.

I am not oppose to this second approach, I am just thinking we should decide & clarify which approach we use and the pros/cons in this design.

Lyndon-Li: Indeed, this configuration things will add significant complications. I don't think we should cover any of the details in this design either. Merely, I think when we want to do this some time later, for example, BIA/RIA v3, there is no significant breaking changes.
The primary benefits of the plugin mechanism is we could "close" the dependent logics in the plugin; on the other hand, if we make it too closed (i.e., we expose an "execute" method only in this design), no information & control could be delivered.
Therefore, in this design, we should clarify how we can avoid the breaking changes brought by the "over-closure". Otherwise, we should clarify we accept the "over-closure" and clarify how to support the complicated cases under this "over-closure" situation.

Personally, I think the latter option will touch too many unnecessary/undetermined details, so I prefer the first option (avoid "over-closure"). I make a draft proposal as below:

- We definitely could use the plugin mechanism for the upload/async operations
- Meanwhile, we expose the control of upload/async operations to the caller
- The "execute" method works on any synchronized tasks and return the control of async tasks to the caller if any
- The caller decides when and how to start the async tasks, just like it deciding when to stop them with the Cancel mechanism
- We only expose the control to the caller in any way, so it won't add too many efforts
- We might sacrifice some versatility (so how to make the async operations purely generic might be a question), I am not 100% sure of this for now, but if any, it worths anyway

sseago: Lets not go there in terms of attempting to give Velero control of the async actions. That adds unnecessary complications that 1) we don't need for any of the 1.10 use cases and 2) adds more places for things to break down. We want to start the  async actions *right now* when the item is backed up and/or restored. If we wait for later, we make the whole process much more fragile. I think the "pass control to velero" idea is something that should be proposed on its own in a future development cycle, since it adds a bunch of complexity that is unnecessary for all of the current use cases.

sseago:
I've added a section near the end which references the various PRs that have been written relating to this with some notes on what would need to be reverted, what we still need, etc.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,497 @@
...
+1. Remove the ItemSnapshotter plugin APIs (and related code) since the revised design will reuse
+   VolumeSnapshotter and BackupItemAction plugins.
+   - https://github.com/vmware-tanzu/velero/pull/4077
+   - https://github.com/vmware-tanzu/velero/pull/4417
+1. UploadProgressFeatureFlag shouldn't be needed anymore. The current design won't really need a
+   feature flag here -- the new features will be added to V2 of the VolumeSnapshotter,
+   BackupItemAction, and RestoreItemAction plugins, and it will only be used if there are plugins which
+   return operation IDs.
+   - https://github.com/vmware-tanzu/velero/pull/4416
+1. Adds <backup-name>-itemsnapshots.gz file to backup (when provided) -- this is still part of the
```
blackpiglet, line 457: The file name should be <backup-name>-itemoperations.gz.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,497 @@
...
+
+Since this design is modifying a previously-approved design, there is some preparation work based on
+the earlier upload progress monitoring design that may need modification as a result of these
+updates. Here is a list of some of these items:
+
+1. Consts for the "Uploading" and "UploadingPartialFailure" phases have already been defined. These
+   will need to be removed when the "WaitingForPluginOperations" and
+   "WaitingForPluginOperationsPartialFailure" phases are defined.
+   - https://github.com/vmware-tanzu/velero/pull/3805
+1. Remove the ItemSnapshotter plugin APIs (and related code) since the revised design will reuse
```
blackpiglet, line 448: Agree, but need to check with Dave before removing it, and UploadProgressFeatureFlag removing also needs confirm.

sseago:
One note on implementation schedule. The CRD changes and v2 plugin versions should be the highest priority for implementation first. 

CRD changes are all minor (and backwards-compatible) -- new status.phase values for Backup and Restore, just added to the validation enum, DownloadRequest target change from BackupItemSnapshots to BackupItemOperations, and a new RestoreItemOperations, as well as optional progress-related status field additions for Backup/Restore.

v2 plugin  API changes are independent of this functionality and we can document that the new API methods/fields will be ignored by velero initially (until the remainder of the functionality comes in).

sseago:
Are we ready to merge this design? I think all of the feedback has been responded to.

shubham-pampattiwar (dismissed):

blackpiglet:
> Are we ready to merge this design? I think all of the feedback has been responded to.

@sseago 
I remember during an internal meeting, there was some discussion about this topic. One of the major concerns is about the  whether Velero should have more control on when to start the plugin operation.
https://github.com/vmware-tanzu/velero/pull/5108#discussion_r929874860
@Lyndon-Li 
Could you PTAL when available?

sseago:
@blackpiglet I responded to that above when it was mentioned here. In summary, I think it shouldn't be part of this proposal. It's not something needed for the identified use cases. It's really a separate bit of functionality that could be proposed and added later if there are specific use cases, but in general, it makes this significantly more complicated and will make velero backup/restore less robust if used, since it provides more points of failure. Repeating my earlier comment here for clarity:

Lets not go there in terms of attempting to give Velero control of the async actions. That adds unnecessary complications that 1) we don't need for any of the 1.10 use cases and 2) adds more places for things to break down. We want to start the async actions right now when the item is backed up and/or restored. If we wait for later, we make the whole process much more fragile. I think the "pass control to velero" idea is something that should be proposed on its own in a future development cycle, since it adds a bunch of complexity that is unnecessary for all of the current use cases.

shubham-pampattiwar:
@blackpiglet +1 to what @sseago is stating. We do not need Velero controlling the async plugin actions currently, as it will add more complications plus there are no uses cases identified for this change. I believe we should go ahead with the current design proposal.

blackpiglet:
@sseago @shubham-pampattiwar 
Thanks for clarification. The other reviewer still need some time for this PR, and the target is next community meeting.

shubham-pampattiwar:
@reasonerjt @blackpiglet @Lyndon-Li Could we please get some final reviews on this design PR ?

sseago:
I've responded to prior feedback -- just waiting on whether the responses were considered adequate. Once we get this approved we can review/approve the specific v2 plugin API design PRs.

weshayutin:
@kaovilai @eemcmullan please review

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
...
+not be possible with all storage systems.
+
+### State progression
+
+(Note that the diagrams have not yet been updated to reflect the terminology changes associated with
+the revised design.)
+
+![image](UploadFSM.png)
+### New
+When a backup/restore request is initially created, it is in the "New" phase.
```
eemcmullan, line 131: Instead of "New" what about "Validating" ?

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
...
+VolumeSnapshotter/BackupItemAction operations and call the _Progress_ method on each of them.
+
+If all snapshot and backup item operations have finished (either successfully or failed), the backup
+will be completed and the backup will move to the appropriate terminal phase and upload the
+`velero-backup.json` object to the object store and the backup will be complete.
+
+If any of the snapshots or backup items are still being processed, the phase of the backup will be
+set to the appropriate phase (_WaitingForPluginOperations_ or
+_WaitingForPluginOperationsPartialFailure_).  In the event of any of the progress checks return an
+error, the phase will move to _WaitingForPluginOperationsPartialFailure_.  The backup will then be
```
eemcmullan, line 412: Once the phase is moved to _WaitingForPluginOperationsPartialFailure_ and is requeued, will the user know it is requeued? Will the phase move back to _WaitingForPluginOperations_ ?

design/general-progress-monitoring.md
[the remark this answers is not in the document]
sseago: That's out of scope for this change. We're not going to be modifying what happens before a backup/restore starts. I think "New" is the convention for this sort of thing, though. It's what Velero uses for backups and restores, and other CRDs use this outside of the Velero project as well.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
...
+the backup will either have fatal errors or will be at least partially usable.
+
+If the backup/restore has fatal errors it will move to the "Failed" state and finish. If a
+backup/restore fails, the upload or other operation will not be cancelled but it will not be
+monitored either.  For backups in any phase, all snapshots will be deleted when the backup is
+deleted.  Plugins will cancel any data movement or other operations and remove snapshots and other
+associated resources when the VolumeSnapshotter DeleteSnapshot method or DeleteItemAction Execute
+method is called.
+
+Velero will poll the plugins for status on the operations when the backup/restore exits the
```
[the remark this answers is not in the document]
sseago: It will happen on re-reconcile. While there are active backups/restores, it probably won't be updated during the next backup, etc., since the controller is currently single-threaded, but once there are no queued backups/restores waiting to run, we may need to use some sort of re-check interval similar to the one-minute validation interval used for BSL validation, etc.

design/general-progress-monitoring.md
sseago: Once it's moved to WaitingForPluginOperationsPartialFailure, it won't move back to WaitingForPluginOperations since any backup or restore with at least one recorded error on it is on the partial failure track. If there are errors *before* moving to one of the waiting phases, then it moves to WaitingForPluginOperationsPartialFailure and eventually to PartiallyFailed. If there are no errors before waiting, then it moves to WaitingForPluginOperations and eventually to Completed, unless an error is logged by one of the async actions we're waiting on -- if that happens it moves to WaitingForPluginOperationsPartialFailure and then once all actions complete, it moves to PartiallyFailed.

design/general-progress-monitoring.md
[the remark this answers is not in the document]
eemcmullan: Gotcha. Thanks for clarifying.

eemcmullan (approved):

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
+# Backup and Restore Item Action Progress Monitoring
```
kaovilai, line 1: Thoughts on calling this something more user friendly?

A term that most users regardless of plugin development experience will understand is the term plugin.
The term item action may not be familiar to everyone.
```suggestion
# Plugin Progress Monitoring
```

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
...
+restore has entered this phase, Velero is free to start another backup/restore.  The backup/restore
+will remain in the "WaitingForPluginOperations" phase until all BIA/RIA operations have completed
+(for example, for a volume snapshotter, until all data has been successfully moved to persistent
+storage).  The backup/restore will not fail once it reaches this phase.  If the backup is deleted
+(cancelled), the plug-ins will attempt to delete the snapshots and stop the data movement - this may
+not be possible with all storage systems.
+
+### State progression
+
+(Note that the diagrams have not yet been updated to reflect the terminology changes associated with
```
reasonerjt, line 126: So the diagram will be updated to reflect the new terminologies right?

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,496 @@
...
+effectively orphaned.  This can currently happen but the current window is much smaller.
+
+### `<backup-name>-itemoperations.json.gz`
+The itemoperations file is similar to the existing `<backup-name>-itemsnapshots.json.gz`  Each snapshot taken via
+BackupItemAction will have a JSON record in the file.  Exact format TBD.
+
+## Changes to Velero restores
+
+A `<restore-name>-itemoperations.json.gz` file will be added that contains the items and operation IDs
+returned by RestoreItemActions.
```
Lyndon-Li, line 380: We are currently considering the design for the unified data path feature, that is, we put everything (so also include data from Velero's build-in data movers) into the backup repo (or kopia repo as it is the only candidate at present) through the unified repo interface, this is to support the scenarios in on-premise envs where there is no object store (i.e. only NFS is available). This means, we cannot assume that there is a sophisticated object store, then we have some restrictions for data persistency:

- All the data to be persisted must be in either a backup or a restore folder that is pre-created at the beginning of backup/restore
- For each backup/restore folder, the data to be persisted must be persisted by one single writer, otherwise, they must be written sequentially

Otherwise, we will have to develop something like a distributed file system where metadata is carefully taken care of to prevent mess up by in parallel write. That will be very complicated.

I am afraid the current itemoperation.json conflicts with the restriction if these json are written inside the BIA/RIA plugins, because we have no way to prevent them from concurrent write.

Of course, this BIA/RIA interfaces are for general purposes and not only for Velero build-in modules or build-in data movers. However, we should give a way for Velero build-in data movers, which are using these BIA/RIA interfaces, to delegate the json persistency to Velero to assure sequent write.

Alternatively, can we leave this json persistency thing to the specific plugin? If so, there will be no need to consider all the things above.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
...
+While in the WaitingForPluginOperations or WaitingForPluginOperationsPartialFailure phase, the
+snapshots and item actions will be periodically polled.  When all of the snapshots and item actions
+have reported success, the backup/restore will move to the Completed or PartiallyFailed phase,
+depending on whether the backup/restore was in the WaitingForPluginOperations or
+WaitingForPluginOperationsPartialFailure phase.
+
+The Backup resources will not be written to object storage until the backup has entered a final phase: 
+Completed, Failed or PartialFailure
+
+## Reconciliation of InProgress backups
```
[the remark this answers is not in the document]
sseago: This was in Dave's original approved design -- re-reading it, I think it just clarifies that Velero won't start polling a backup for operation completion until the main backup process completes. If it's `InProgress`, then we're still backing up items, so we're not yet checking to see whether async operations are done yet. It's really more relevant once we move to the multithreaded backup process. As long as backups/restores are fully single-threaded, then Velero won't have the opportunity to re-reconcile an `InProgress` backup anyway.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
...
+Two new methods will be added to the VolumeSnapshotter interface:
+
+    Progress(snapshotID string) (UploadProgress, error)
+    Cancel(snapshotID string) (UploadProgress, error)
+
+Open question: Does VolumeSnapshotter need Cancel, or is that only needed for BIA/RIA?
+
+Progress will report the current status of a snapshot upload.  This should be callable at
+any time after the snapshot has been taken.  In the event a plug-in is restarted, if the operationID
+(snapshot ID) continues to be valid it should be possible to retrieve the progress.
```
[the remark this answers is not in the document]
sseago: This is up to the plugin to work out. In the case of a plugin which creates a CR to signal an operation from an external controller (datamover use case), then the plugin's ability to report progress is determined by the external controller's ability to store this in the CR status. If there's a case where progress updates don't happen until the end, then the totalToComplete might just be 1, and the completedSoFar may end up being 0 if not finished, and 1 if finished. I'm not as clear on how the volumesnapshotter handles this. In any case, I'm not sure that's relevant to this statement. It simply says "if the ID is still valid after restart, we should still be able to track progress". If we can't, we can't, of course, and if the ID is no longer valid, then progress calls will error out.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,496 @@
...
+functionality is useful for any longstanding plugin operation managed by an external
+process/controller so we will modify BackupItemAction and RestoreItemAction to optionally return an
+operationID in addition to the modified item.
+
+The Backup/RestoreItemAction plugin identifier as well as the Item and Snapshot ID will be stored in
+the `<backup-name>-itemoperations.json.gz`.  When checking for progress, this info will be used to
+select the appropriate Backup/RestoreItemAction plugin to query for progress.
+
+Velero can attempt to cancel an operation by calling the Cancel API call on the BIA/RIA. The plugin
+can then take any appropriate action as needed. Cancel will be called on backup deletion, and
```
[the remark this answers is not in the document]
sseago: Yes, this is correct. I will add a comment to this effect. The only real purpose of cancel is for operations that can be canceled, Velero is telling  them "we're failing the backup now, so might as well stop doing work if you can"

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,496 @@
...
+possibly upon reaching timeouts, if timeout support is included in this feature.
+
+_NotApplicable_ should only be returned if the Backup/RestoreItemAction plugin should not be
+handling the item.  If the SnapshotItemAction plugin should handle the item but, for example, the
```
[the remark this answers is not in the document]
sseago: Fixed, also updated wording to reflect that Cancel/Progress will return an error rather than a populated OperationProgress struct if the operation doesn't exist.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,496 @@
...
+
+The Backup/RestoreItemAction plugin identifier as well as the Item and Snapshot ID will be stored in
+the `<backup-name>-itemoperations.json.gz`.  When checking for progress, this info will be used to
+select the appropriate Backup/RestoreItemAction plugin to query for progress.
+
+Velero can attempt to cancel an operation by calling the Cancel API call on the BIA/RIA. The plugin
+can then take any appropriate action as needed. Cancel will be called on backup deletion, and
+possibly upon reaching timeouts, if timeout support is included in this feature.
+
+_NotApplicable_ should only be returned if the Backup/RestoreItemAction plugin should not be
```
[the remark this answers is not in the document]
sseago: Both. I updated the text to clarify.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,496 @@
...
+handling the item.  If the SnapshotItemAction plugin should handle the item but, for example, the
+item/snapshot ID cannot be found to report progress, an OperationProgress struct with the error set
+appropriately (in this case _NotFound_) should be returned. If the item action does not start an
```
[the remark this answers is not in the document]
sseago: That's actually what I already realized should be done, as I started going into the v2 API changes for the plugins. Updating the doc to reflect this.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,496 @@
...
+### BackupItemAction and RestoreItemAction plug-in changes
+
+Currently CSI snapshots and the Velero Plug-in for vSphere are implemented as BackupItemAction
+plugins.  While the majority of BackupItemAction plugins do not take snapshots or upload data, this
+functionality is useful for any longstanding plugin operation managed by an external
+process/controller so we will modify BackupItemAction and RestoreItemAction to optionally return an
+operationID in addition to the modified item.
+
+The Backup/RestoreItemAction plugin identifier as well as the Item and Snapshot ID will be stored in
+the `<backup-name>-itemoperations.json.gz`.  When checking for progress, this info will be used to
```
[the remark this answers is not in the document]
sseago: So the operationID will be the snapshotID for volumesnapshotter plugins. For data mover BIA/RIA plugins, it will be an appropriate ID for the datamover CRs. The item should also just be an identifier, not the entire resource. Adding an example to the doc.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,496 @@
...
+
+The cluster that is creating the backup will have the Backup resource present and will be able to
+manage the backup before the backup completes.
+
+If the Backup resource is removed (e.g. Velero is uninstalled) before a backup completes and writes
+its `velero-backup.json` object, the other objects in the object store for the backup will be
+effectively orphaned.  This can currently happen but the current window is much smaller.
+
+### `<backup-name>-itemoperations.json.gz`
+The itemoperations file is similar to the existing `<backup-name>-itemsnapshots.json.gz`  Each snapshot taken via
```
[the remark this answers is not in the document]
sseago: Sure. I think, generally, this would be uploaded to the object store after processing all backup/restore items, as until this point we won't have the complete list. It will only be used on subsequent reconciles, so it won't be referenced until the backup moves to the next phase  beyond `InProgress`

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,496 @@
...
+appropriate phase (_WaitingForPluginOperations_ or _WaitingForPluginOperationsPartialFailure_).  In
+the event of any of the progress checks return an error, the phase will move to
+_WaitingForPluginOperationsPartialFailure_.  The restore will then be requeued and will be rechecked
+again after some time has passed.
+
+## Restart workflow
+
+On restart, the Velero server will scan all Backup/Restore resources.  Any Backup/Restore resources
+which are in the _InProgress_ phase will be moved to the _Failed_ phase.  Any Backup/Restore
+resources in the _WaitingForPluginOperations_ or _WaitingForPluginOperationsPartialFailure_ phase
```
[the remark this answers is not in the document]
sseago: Yes, this gets persisted before moving from `InProgress` to `WaitingForPluginOperations`

design/general-progress-monitoring.md
sseago: @Lyndon-Li So these .json.gz files are just another file in the object store. We also have gzipped logs there per-backup/restore. If you can handle relocating the logs, then these would be handled the same way. In fact, for the use case of "on-permise NFS", I think it's probably more appropriate to handle that behind the object store abstraction. In other words, if we write an object store plugin that stores its data using on-premise NFS, then the velero backup/restore process won't care. It's just using `NewObjectStore` rather than `AwsObjectStore`.

sseago:
@reasonerjt @Lyndon-Li I've responded to the last round of comments and updated the PR with some changes in response.

design/general-progress-monitoring.md
Lyndon-Li: @sseago 
It is great that the plugin itself doesn't persist anything and the backup/restore controller does it instead. As the discussion, please check and add the clarification in the design to show that which part is done by the interface, which part is done by the specific plugin implementation and which one is done by velero. Thanks.

sseago: @Lyndon-Li So the design already split out plugin API vs. backup/restore workflow changes. However, I noticed that the json files for ongoing operations was mentioned in both the workflow and plugin API sections, which may have contributed to the confusion. I've moved that content around so that it's not mentioned in the plugin API section now.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,515 @@
...
+
+
+## Primary changes from the original Upload Progress Monitoring design
+
+The most fundamental change here is that rather than proposing a new special-purpose
+SnapshotItemAction, the existing BackupItemAction plugin will be modified to accommodate an optional
+snapshot (or other item operation) ID return. The primary reasons for this change are as follows:
+
+1. The intended scope has moved beyond snapshot processing, so it makes sense to support
+asynchronous operations in in other backup or restore item actions.
```
kaovilai, line 60:
```suggestion
asynchronous operations in other backup or restore item actions.
```

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
...
+any uploads still in progress should be aborted.
+
+### WaitingForPluginOperations (new)
+The "WaitingForPluginOperations" phase signifies that the main part of the backup/restore, including
+snapshotting has completed successfully and uploading and any other asynchronous BIA/RIA plugin
+operations are continuing.  In the event of an error during this phase, the phase will change to
+WaitingForPluginOperationsPartialFailure.  On success, the phase changes to Completed.  Backups
+cannot be restored from when they are in the WaitingForPluginOperations state.
+
+### WaitingForPluginOperationsPartialFailure (new)
```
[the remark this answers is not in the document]
reasonerjt: Sounds good to me.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
...
+ignored.
+
+## Plug-in API changes
+
+### OperationProgress struct
+
+    type OperationProgress struct {
+        completed bool                          // True when the operation has completed, either successfully or with a failure
+        err error                               // Set when the operation has failed
+        itemsCompleted, itemsToComplete int64   // The number of items that have been completed and the items to complete
```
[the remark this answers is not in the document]
reasonerjt: I was thinking we can provide a more flexible `progressJSON  string` so one plugin can report the progress in its own way...  
If the progress has to be reflected in some of the fields of the backup CR we can tell the plugin developer to provide the JSON string in a certain format like `{ 'completed_bytes':xxxx, 'total_bytes':xxxx }`.  And for those plugins do not have the capability to report progress it just returns an empty JSON object `{}`

Do you think this would work?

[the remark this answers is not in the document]
kaovilai: Plugin author could provide two ints and a string for units. Velero could do a quick division for percentage calculations. Or yeah just accept any string.
```
1 processed / 5 total (<plugin-author-chosen-units>)
```

design/general-progress-monitoring.md
reasonerjt: so what can be done to mitigate the side-effects for the async operations that have been started when the backup goes to `failed` state?

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,515 @@
...
+backup/restore fails, the upload or other operation will not be cancelled but it will not be
+monitored either.  For backups in any phase, all snapshots will be deleted when the backup is
+deleted.  Plugins will cancel any data movement or other operations and remove snapshots and other
+associated resources when the VolumeSnapshotter DeleteSnapshot method or DeleteItemAction Execute
+method is called.
+
+Velero will poll the plugins for status on the operations when the backup/restore exits the
+"InProgress" phase and has no fatal errors.
+
+If any operatios are not complete, the backup/restore will move to either WaitingForPluginOperations
```
reasonerjt, line 205: typo: `operatios` -> `operations`

design/general-progress-monitoring.md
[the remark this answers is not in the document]
reasonerjt: > the totalToComplete might just be 1, and the completedSoFar may end up being 0 if not finished, and 1 if finished
My point is this the controller or caller of this func will be confused if the value of `completedSoFar` is the number of snapshots or the number of bytes.  IMO this can be mitigated if we update the struct `OperationProgress` as proposed in[ this comment](https://github.com/vmware-tanzu/velero/pull/5108/files#r1010104191)

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,515 @@
...
+                                                // systems retain when the upload was begun, return Time 0 (time.Unix(0, 0))
+                                                // if unknown.
+    }
+
+### VolumeSnapshotter changes
+
+Two new methods will be added to the VolumeSnapshotter interface:
+
+    Progress(snapshotID string) (OperationProgress, error)
+    Cancel(snapshotID string) (OperationProgress, error)
```
reasonerjt, line 250: Sorry I missed this last time I review this PR.  What should the return value `OperationProgress` of func `Cancel`  be?

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,515 @@
...
+possibly upon reaching timeouts, if timeout support is included in this feature. Cancel is not
+intended to be used to delete/remove the results of completed actions.
+
+_AsyncOperationsNotSupportedError_ should only be returned (by either Cancel or by Progress) if the
+Backup/RestoreItemAction plugin should not be handling the item.  If the Backup/RestoreItemAction
+plugin should handle the item but, for example, the item/snapshot ID cannot be found to report
+progress, Progress or Cancel will return an InvalidOperationIDError error rather than a populated
+OperationProgress struct. If the item action does not start an asynchronous operation, then
+operationID will be empty.
+
```
reasonerjt, line 280: Should we also clarify the expected return values if `Progress` or `Cancel` is called against a canceled operation?  Will it be an error or somehow reflected in the `OperationProgress`?

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,515 @@
...
+`<backup-name>-itemoperations.json.gz` file will be added that contains the items and operation IDs
+(snapshotIDs) returned by VolumeSnapshotter and BackupItemAction plugins.  Also, the creation of the
+`velero-backup.json` object will not occur until the backup moves to one of the terminal phases
+(_Completed_, _PartiallyFailed_, or _Failed_).  Reconciliation should ignore backups that do not
+have a `velero-backup.json` object.
+
+The Backup/RestoreItemAction plugin identifier as well as the ItemID and OperationID will be stored
+in the `<backup-name>-itemoperations.json.gz`.  When checking for progress, this info will be used
+to select the appropriate Backup/RestoreItemAction plugin to query for progress. Here's an example
+of what a record for a datamover plugin might look like:
```
reasonerjt, line 367: So the content of the file `<backup-name>-itemoperations.json` will be a list of JSON objects like this?

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,496 @@
...
+
+The vSphere Plug-in for Velero uploads snapshots to S3 in the background.  This is also a
+BackupItemAction plug-in, it will check the status of the Upload records for the snapshot and return
+progress.
+
+## Backup workflow changes
+
+The backup workflow remains the same until we get to the point where the `velero-backup.json` object
+is written.  At this point, we will queue the backup to a finalization go-routine.  The next backup
+may then begin.  The finalization routine will run across all of the
```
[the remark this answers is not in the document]
reasonerjt: Yep, after the section of `<backup-name>-itemoperations.json.gz` is updated I got the idea.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,515 @@
...
+     here is to tell the plugin that Velero won't be waiting anymore, so if there are any
+     required custom cancellation actions, now would be a good time to perform them. For snapshot
+     uploads that are already in proress, there's not really anything else to cancel.
+2. Should we actually write the backup *before* moving to the WaitingForPluginOperations or
+   WaitingForPluginOperationsPartialFailure phase rather than waiting until all operations
+   have completed? The operations in question won't affect what gets written to object storage
+   for the backup, and since we've already written the list of operations we're waiting for to
+   object storage, writing the backup now would make the process resilient to Velero restart if
+   it happens during WaitingForPluginOperations or WaitingForPluginOperationsPartialFailure
+
```
reasonerjt, line 515: Agreed, I think my answer to question `2` is `Yes`

design/general-progress-monitoring.md
sseago: @reasonerjt So currently, once the backup has started, there's not really anything that can happen to move the backup to Failed state (vs. PartiallyFailed). If a running backup (with running async actions) goes into partially failed state, that doesn't really have any impact on async operations though. If, for example, a datamover plugin is copying a completed VolumeSnapshotContents to the BSL, a failure in the associated backup isn't all that relevant. If there's a partial failure, Velero will continue to monitor and move it from WaitingForPluginOperationsPartiallyFailed to PartiallyFailed once done. If a backup did end up moving to an actual complete Failed case (which, again, isn't possible with current workflow), then the async operation would continue in the background, but velero just wouldn't check back on it again. So I don't think there's anything to mitigate here. In any case, I don't think that's particularly relevant to my point above, which is that plugins should start their async operations immediately when running Execute, since the whole point of this is to start the operation now and not block ongoing backup actions while this long-running plugin action is taking place.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
...
+- vSphere - The vSphere plugin takes a local snapshot and then the vSphere plugin uploads the data to S3.  The local
+snapshot is usable before the upload completes.
+
+- Restic - Does not go through the volume snapshot path.  Restic backups will block Velero progress
+until completed. However, with the more generalized approach in the revised design, restic/kopia
+backup and restore *could* make use of this framework if their actions are refactored as
+Backup/RestoreItemActions.
+
+- Data Movers
+    - Data movers are asynchronous processes that are triggered via backup/restore of kubernetes resources. A common use case for this is the backup/restore of PVCs whose data we want to move to some form of backup storage.
```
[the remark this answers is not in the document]
sseago: Updating.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,499 @@
...
+Volume Snapshotter use case only needs this on backup. This means that everything we're doing around
+backup phase and workflow will also need to be done for restore.
+
+Then there are various minor changes around terminology to make things more generic. Instead of
+"snapshotID", we'll have "operationID" (which for volume snapshotters will be a snapshot
+ID).
+
+## Goals
+
+- Enable monitoring of operations that continue after snapshotting and other operations have completed
```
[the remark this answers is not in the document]
sseago: updating

design/general-progress-monitoring.md
sseago: updating

design/general-progress-monitoring.md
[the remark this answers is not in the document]
sseago: I can remove it, although I think it becomes relevant once we add multithreaded backup/restore. The point of this section is simply to call out that if a backup is still in progress, then velero won't poll the plugins for async operation status. It's more a clarification than anything else.

design/general-progress-monitoring.md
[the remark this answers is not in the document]
sseago: @reasonerjt So I thnk you're suggesting we replace both the "completed" and the "total" ints with a single string with the assumption that the plugin will just fill in a user-readable string. In other words, instead of `itemsCompleted=1000;itemsToComplete=10000` we'd have `progressSummary="1000 out of 10000 bytes completed". -- and we would *not* do anything in velero to process this string, we'd just attach it in an appropriate location for progress reporting -- is that the suggestion? I'm OK with that, as long as we're OK with Velero not attempting to give a user an idea of percentage completion.

@kaovilai If we're going to collapse the done/total ints into a single string, I'd rather avoid having to parse this on the velero side, since there's no way velero will know for sure what the plugin might want to report. If we need percentage calculations, then we should stick with some variation on the original design (with updated names as appropriate); if we don't need percentages, then we can just replace the two integers with a single string that Velero will never attempt to parse.

design/general-progress-monitoring.md
sseago: fixing

design/general-progress-monitoring.md
sseago: fixing

design/general-progress-monitoring.md
sseago: Oh, good catch. This should *not* return an OperationProgress. Cancel tells the plugin "cancel this if you can, since velero won't ask again for progess". It should only return an error. I'll update the PR to reflect this (if you look at the plugin API design PRs, you'll see that Cancel does *not* return an OperationProgress)

sseago: The other plugin types already had Cancel correctly not returning this struct.

design/general-progress-monitoring.md
sseago: Cancel just returns an error if something unexpected happens. Calling cancel again shouldn't force an error -- it's up to the plugin, but I'd expect a second Cancel call would be a no-op (and Velero shouldn't actually call it twice unless there's a bug).

For Progress, again, this shouldn't happen, and the actual behavior is up to the plugin, but I'd expect the plugin to do one of two things based on the overall design already identified:
1) if the canceled operation is no longer available (i.e. the CR associated with it has been deleted by its controller), then it should return a standard "not found" error just as it would with an operationID that's never been valid
2) If the canceled operation *is* still there, then I'd expect progress to do what it normally does -- if it's actually in progress (cancel didn't work/isn't supported), then report that. If there's an error state, then report that. If the plugin has information on completion stats, report that. I don't know that it's really a separate case that Velero should care about, since Velero will never call Progress or Cancel after calling Cancel on an operation.

design/general-progress-monitoring.md
sseago: It's a JSON file with a list of objects like this, yes.

design/general-progress-monitoring.md
```diff
@@ -0,0 +1,515 @@
...
+## Changes in Velero backup format
+
+No changes to the existing format are introduced by this change.  As part of the backup workflow changes, a
+`<backup-name>-itemoperations.json.gz` file will be added that contains the items and operation IDs
+(snapshotIDs) returned by VolumeSnapshotter and BackupItemAction plugins.  Also, the creation of the
+`velero-backup.json` object will not occur until the backup moves to one of the terminal phases
+(_Completed_, _PartiallyFailed_, or _Failed_).  Reconciliation should ignore backups that do not
+have a `velero-backup.json` object.
+
+The Backup/RestoreItemAction plugin identifier as well as the ItemID and OperationID will be stored
```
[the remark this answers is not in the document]
sseago: So Velero should only call Cancel if the operation timeout has been reached. I'm thinking that we would apply one timeout to the full list of operations -- so we won't really be cancelling anything unless we're going through the list for the last time post-timeout-expiration. But yes, I don't think cancel removes items from the file. One thing that's not clear to me yet -- I think working through this might be part of the implementation-level task -- should this file also track completion status or not? Doing so would mean subsequent calls would only need to call the plugin again for previously incomplete operations, so that's a plus. The downside of this is we'd need to re-upload this file each time Velero reconciles again. I can see advantages to each approach here.

design/general-progress-monitoring.md
sseago: @reasonerjt I was thinking the same. I never quite understood the "defer writing backup json" aspect of Dave's original design here, so I'd like to continue to write this backup as soon as Velero is done with it, before moving on to the next backup.

sseago:
@reasonerjt @kaovilai @shubham-pampattiwar Updated the PR again with some minor fixes based on recent comments. I think the main unresolved issue is what to do about the OperationProgress struct fields itemsCompleted and itemsToComplete. I think we have 3 options based on the discussion above:
1) Leave it as-is
2) Rename them to something that doesn't involve "items" to get away from possible user assumptions about the meaning of "item" (this was my suggestion)
3) Replace them with a single string "progressSummary" -- the plugin would be responsible for saying "1000 out of 10000 bytes completed" and Velero would not attempt to parse the string, it would just take it as an opaque string.

design/general-progress-monitoring.md
reasonerjt: Thanks

sseago:
I've updated this to reflect recent discussions around the OperationProgress struct. We now have two int64 fields, `Completed` and `Total` representing the current progress of the operation (to facilitate calcuated percentage completion), with a new field `OperationUnits` ("bytes" for datamover and volume snapshotter plugins) to facilitate aggregation of completion totals across plugins. For plugins which are unable to furnish percentage completon, there is also a new `Decription` field in OperationProgress to allow for freeform progress text (perhaps phase in a workflow, etc.).

sseago:
@reasonerjt One reason I favor the explicit fields for completed/total (and added a units field to handle cases where the thing being measured isn't bytes) is that it is explicit in the plugin API, so velero knows that any plugins being run are going to be reporting correctly. By embedding it in json and documenting an approved format, it makes it more of a hidden API, where the plugin must follow it for stats to be displayed, but it's not enforced by the plugin API.

reasonerjt:
@sseago 
Sorry for the late reply.
I agree with your concern regarding "hidden API".  
So the contract is that if the plugin can't report the numeric progress the `complete/total` will be left as `0` right?

sseago:
@reasonerjt That's correct. If the fields are left blank, they should default to 0. So, if for example, a particular action is only able to be reported as "In Progress" or "Phase 2", then that would be set as the string value for `Description`, and the other fields will be left blank.

design/general-progress-monitoring.md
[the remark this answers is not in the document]
reasonerjt: Resolving this after the fields in `OperationProgress struct` have been updated.

design/general-progress-monitoring.md
reasonerjt: > since Velero will never call Progress or Cancel after calling Cancel on an operation.
So it means the Cancel is a sync operation right? i.e. if everything is fine it won't return until the operation is really canceled.  If this is the case, could you explicitly define that in this design？

sseago: If everything is fine, Velero will never call cancel. The only time Velero will call Cancel is if we've hit the timeout and the operation is still ongoing. However, once we call cancel, since it's up to the plugin what to do with it (could be a no-op if the plugin can't cancel), once Cancel returns, velero moves on. Basically the Cancel API call is "best effort" -- Velero doesn't guarantee that the cancel was successful, since some plugins may not be able to cancel at all. Velero calls cancel and moves on. I think that was stated explicitly, but if it's not clear enough, let me know and I can try to reword that somewhere.

reasonerjt: @sseago 
Thanks for the reply. 

Since you are putting details like error types in this section, IMO, the behavior of `Cancel` func of a plugin should be be clarified.
I suggest we add more details in the paraphrase  `Velero can attempt to cancel an operation by calling the Cancel API call on the BIA/RIA....`.   We should clarify the contract between the velero and the plugin in terms of `Cancel` func.  For example, when this function returns `nil`, does it indicates (a) the operation is canceled successfully or (b)  the cancel request is handled but the result is unknown?

I presonally prefer option (a),  if it's not possible for the plugin to know if the operation is really canceled, it can return a `status unknown error` and velero may choose to ignore it.
If you think (b) is a better choice, I'd suggest we clarify in what case the `Cancel` func returns a non-nil value, or maybe it'd be better that the `Cancel` func does not have any values.  So the caller just call the func to attempt to cancel the operation, and it may or may not be successful.

Either way, I hope this can be clarified in this design, otherwise, the plugin developer will be confused when he/she is implementing the `Cancel` func.

sseago: There is no expectation for cancel to return an error under normal operations. Since any plugin-implemented cancel request is likely to be async as well, there's no way to provide velero with reliable "this was successfully cancelled" information in the general case, since velero won't be checking back. The whole point of cancel from velero's point of view is "we hit the timeout, we're declaring this operation failed, and we're giving the plugin an opportunity to cancel on its end if it can" -- but velero is logging this as an error either way. This means the plugin doesn't need to tell velero anything. A plugin which doesn't support cancel will implement a no-op func -- returning nil for error. A cancel-supporting plugin will do whatever it needs to cancel and return nil for error (and it probably won't know at this point whether cancel was successful). An error will be returned if the plugin itself gets an unexpected error in cancellation.

reasonerjt: > A plugin which doesn't support cancel will implement a no-op func -- returning nil for error. A cancel-supporting plugin will do whatever it needs to cancel and return nil for error (and it probably won't know at this point whether cancel was successful). An error will be returned if the plugin itself gets an unexpected error in cancellation.

This sounds OK to me.  Let's make sure this is documented, in the .md file and in the inline comment of the go interface.

reasonerjt (approved):

reasonerjt:
@sseago 
I'm approving this design.  If you wanna make updates to the markdown file to clarify the `Cancel` func following [this comment](https://github.com/vmware-tanzu/velero/pull/5108/files#r1040734838)

Please ping me on slack and I'll approve again

shubham-pampattiwar (approved):

PR #5101, opened 2022-07-06
Fix bsl validation bug

ywk253100:
Fix bsl validation bug: the BSL is validated continually and doesn't respect the validation period configured

Fixes #5056

Signed-off-by: Wenkai Yin(尹文开) <yinw@vmware.com>

Thank you for contributing to Velero!

# Please add a summary of your change

# Does your change fix a particular issue?

Fixes #(issue)

# Please indicate you've done the following:

- [ ] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [ ] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required` as a comment on this pull request.
- [ ] Updated the corresponding documentation in `site/content/docs/main`.

PR #5122, opened 2022-07-14
Modify BackupStoreGetter to avoid BSL spec changes

sseago:

Modify BackupStoreGetter to avoid BSL spec changes
    
Pass in a new copy of the map of config values rather than
modifying the BSL Spec.Config and then pass in that field.

Those extra fields in spec.config are only needed for the in-memory copy passed into plugins. Without this change to the BackupStoreGetter, the BSL spec is modified on disk and in the OADP case, the OADP controller then noticed the change and removed the fields, causing constant BSL spec updates, causing constant validation reconciles.

Fixes #(issue)

### DIFF
--- a/changelogs/unreleased/5122-sseago
+++ b/changelogs/unreleased/5122-sseago
@@ -1,0 +1,1 @@
+Modify BackupStoreGetter to avoid BSL spec changes
--- a/pkg/persistence/object_store.go
+++ b/pkg/persistence/object_store.go
@@ -131,19 +131,25 @@ func (b *objectBackupStoreGetter) Get(location *velerov1api.BackupStorageLocatio
 		return nil, errors.Errorf("backup storage location's bucket name %q must not contain a '/' (if using a prefix, put it in the 'Prefix' field instead)", location.Spec.ObjectStorage.Bucket)
 	}
 
+	// Pass a new map into the object store rather than modifying the passed-in
+	// location. This prevents Velero controllers from accidentally modifying
+	// the in-cluster BSL with data which doesn't belong in Spec.Config
+	objectStoreConfig := make(map[string]string)
+	if location.Spec.Config != nil {
+		for key, val := range location.Spec.Config {
+			objectStoreConfig[key] = val
+		}
+	}
+
 	// add the bucket name and prefix to the config map so that object stores
 	// can use them when initializing. The AWS object store uses the bucket
 	// name to determine the bucket's region when setting up its client.
-	if location.Spec.Config == nil {
-		location.Spec.Config = make(map[string]string)
-	}
-
-	location.Spec.Config["bucket"] = bucket
-	location.Spec.Config["prefix"] = prefix
+	objectStoreConfig["bucket"] = bucket
+	objectStoreConfig["prefix"] = prefix
 
 	// Only include a CACert if it's specified in order to maintain compatibility with plugins that don't expect it.
 	if location.Spec.ObjectStorage.CACert != nil {
-		location.Spec.Config["caCert"] = string(location.Spec.ObjectStorage.CACert)
+		objectStoreConfig["caCert"] = string(location.Spec.ObjectStorage.CACert)
 	}
 
 	// If the BSL specifies a credential, fetch its path on disk and pass to
@@ -154,7 +160,7 @@ func (b *objectBackupStoreGetter) Get(location *velerov1api.BackupStorageLocatio
 			return nil, errors.Wrap(err, "unable to get credentials")
 		}
 
-		location.Spec.Config["credentialsFile"] = credsFile
+		objectStoreConfig["credentialsFile"] = credsFile
 	}
 
 	objectStore, err := objectStoreGetter.GetObjectStore(location.Spec.Provider)
@@ -162,7 +168,7 @@ func (b *objectBackupStoreGetter) Get(location *velerov1api.BackupStorageLocatio
 		return nil, err
 	}
 
-	if err := objectStore.Init(location.Spec.Config); err != nil {
+	if err := objectStore.Init(objectStoreConfig); err != nil {
 		return nil, err
 	}
 


blackpiglet:
@sseago 
#5101 is similar to this issue. Not sure whether this PR also fix your case.

shubham-pampattiwar (dismissed):

kaovilai:
@blackpiglet #5101 is what caused issue with OADP.
#5101 causes BSL to revalidate when spec changes.

Each time revalidation is called objectStoreGetter.Get is called inside Reconcile function
https://github.com/vmware-tanzu/velero/blob/f550f8e3cda9b6cfaad065525ccc679703622000/pkg/controller/backup_storage_location_controller.go#L125

location's spec.Config is modified
https://github.com/vmware-tanzu/velero/blob/c46fe71b12ecc16786676886a64971102527d631/pkg/persistence/object_store.go#L134-L135

OADP will try to undo any unexpected spec changes with BSL inside our custom resource.

If BSL inside OADP custom resource do not contain bsl.spec.Config, we will remove any added configs from above reconcile which causes spec change and revalidation to run again and this causes infinite and high frequency execution competition between two controllers.

This PR has to be looked at carefully since it can break some plugins that may expect to read `spec.Config["bucket"]` and `spec.Config["prefix"]`. It may be a third party plugin written by non velero authors.

sseago:
@kaovilai It's not really only an OADP issue -- this will happen any time a BSL is managed by an external controller/operator, since the Velero BSL validation code was actually modifying spec and updating the resource. The fix in this PR is to validate a *copy* of the BSL resource, not the same in-memory object that we're going to update status on and make a patch API call. This PR shouldn't affect actually calling the plugin, since I only modified the call to BackupStoreGetter used for validation here. But yes, we should confirm that the plugin only ever uses in-memory BSL modified by this method rather than making a separate Get call for it later, but if the latter were the case, then OADP would already be broken with respect to BSLs, since any time this is done, the OADP controller will un-do the change.

Passing a deep copy of the BSL into the getter feels like the right thing to do here, as Velero shouldn't need to modify the spec like that -- if it did, then everyone who manages BSLs automatically would have to hack around this like you suggested OADP do. That feels like an awful hack.

sseago:
I could be completely wrong here and prior versions of Velero did end up modifying the BSL in-cluster, and prior versions of OADP ended up in a continual modify-the-BSL battle between OADP operator and Velero controller. It would be easy enough to verify with OADP 1.0.3 by doing a watch on the BSL resource yaml.

sseago:
@kaovilai ok, so basically every time we need an objectstore to work with (backup controller, restore controller, etc.) we call ObjectBackupStoreGetter.Get() which modifies the passed-in BSL object and then uses that modified config to initialize the plugin. Of note is that the Init call takes a map with the contents of config in it (so it comes from memory, not the cluster), and that's used to initialize the plugin. The plugin doesn't work with the BSL kube resource directly, only the map of config values passed in.

Also of note is that in every other controller that calls ObjectBackupStoreGetter.Get() does *not* issue a patch or update API call on the BSL afterwards. Only this one does that. This one also only uses the initialize object store for validation, so I do think that *not* modifying the resource on disk is the original intent of the code here, since none of the controllers that use the object store for real work modify the in-cluster resource -- and the one that does is only initializing the object store for validation, so it seems like we shouldn't be putting this change into the cluster.

There is one remaining part of the BSL controller reconcile that could still trip OADP up here -- the controller also updates `location.Spec.Default`, so if velero ever ends up changing that, we'll hit the same continual refresh problem with OADP, but I think in that case the fix would be on the user side -- for the OADP user to update spec.Default in the DPA spec to whatever Velero thinks it should be.

pkg/controller/backup_storage_location_controller.go
```diff
@@ -122,7 +122,7 @@ func (r *BackupStorageLocationReconciler) Reconcile(ctx context.Context, req ctr
 			}
 		}()
 
-		backupStore, err := r.BackupStoreGetter.Get(&location, pluginManager, log)
+		backupStore, err := r.BackupStoreGetter.Get(location.DeepCopy(), pluginManager, log)
```
kaovilai, line 125: I'd add a comment on why we are doing DeepCopy() here. Alternatively do deep copy in the `objectStoreGetter Get()` function and add a comment there instead. Fixing Get function (so that it actually just get something and not setting anything else) rather than fixing Reconcile function which innocently called Get not expecting change.

```suggestion
		// Using DeepCopy to avoid modifying in-cluster BackupStorageLocation.
		// Get function modifies the location.Spec.Config which we don't want updated to the in-cluster version
		backupStore, err := r.BackupStoreGetter.Get(location.DeepCopy(), pluginManager, log)
```

kaovilai (changes requested):

blackpiglet:
I agree with using deepcopy to avoid overwritten, but `BackupStoreGetter.Get` is called in several other places too. 
Is it possible to have a generic fix for all the called places in this PR?

blackpiglet:
May take more time to evaluate how to create a modification for all solution, so agree with @kaovilai, adding more comments will be good enough for this case.

pkg/controller/backup_storage_location_controller.go
sseago: @kaovilai Yeah, I'm leaning towards the latter. The only aspect of the BSL that's passed into the ObjectStore plugin is actually the map of config values. Lets copy *that* map instead of deep copying the whole BSL, and pass it in. No side effect on the kube object. As for 3rd party plugins relying on the BSL in-cluster mod, while I don't think any are doing that now, if they are, it's a dangerous assumption even with the released 1.9 code, since as we've seen with OADP, the change isn't reliably made. Better to ensure it's never made so that breakage is obvious rather than nondeterministic.

sseago: A "Get" function should never have side effects. If it does have side effects, they had better be well-documented and made with good reason. In this case, they are neither documented nor made with any reason to do so.

shubham-pampattiwar (approved):

blackpiglet (approved):

Issue #5079, opened 2022-07-04
Unified Repository - Repository Provider


Lyndon-Li:
Epic: #5050 
Design: #4926 

Add the implementation of Unified Repository Provider, include:

- List all snapshots by specified filters
- Delete a snapshot by a specified snapshot ID
- Get repository information
- UT

Lyndon-Li:
The changes are covered in #5179 and #5167

PR #5167, opened 2022-08-01
repo credentials

Lyndon-Li:
Add changes for Kopia Integration: Unified Repository Provider - Repo Password
Related Issue: #[5079](https://github.com/vmware-tanzu/velero/issues/5079)


pkg/repository/keys/keys.go
```diff
@@ -0,0 +1,75 @@
...
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/vmware-tanzu/velero/pkg/builder"
+)
+
+const (
+	credentialsSecretName = "velero-restic-credentials"
+	credentialsKey        = "repository-password"
+
+	encryptionKey = "static-passw0rd"
```
qiuming-best, line 35: which means that our kopia repo paasword is plain text and invariable？ and which may lead to some security issue?

internal/credentials/secret_store.go
```diff
@@ -0,0 +1,56 @@
...
+// for the given namespace.
+func NewNamespacedSecretStore(client kbclient.Client, namespace string) (SecretStore, error) {
+	return &namespacedSecretStore{
+		client:    client,
+		namespace: namespace,
+	}, nil
+}
+
+// Buffer returns the secret key defined by the given selector.
+func (n *namespacedSecretStore) Buffer(selector *corev1api.SecretKeySelector) (string, error) {
```
qiuming-best, line 49: whether we should change it to one easily understood function name instead of `Buffer`? this function is to get a secret for a specific namespace

pkg/repository/keys/keys.go
Lyndon-Li: This is the existing code and file, this keys.go is moved from Restic package to the repository package.
As the agreement, in v1.10, we will keep the existing behaviors for Kopia integration.

In future, as you can see, the provider construct accepts a `CredentialGetter`, in which, we can add whatever implementation to get a password. For example, we may get repo password from KMS in future, then we can add something like `KmsStore` interface into it.

internal/credentials/secret_store.go
Lyndon-Li: This Buffer means getting a buffer stream, since the interface has already named as SecretStore, we know that it is to get the stream buffer from Secret, so no need to add something like "secret" in the method name. This is also to comply with FileStore's Path method.

As the discuss, we leave it as it in this PR. Later, if there is any conflict or better name, we come back to change it.

qiuming-best (dismissed):

internal/credentials/secret_store.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/pkg/errors"
+	corev1api "k8s.io/api/core/v1"
+	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/vmware-tanzu/velero/pkg/util/kube"
+)
+
+// SecretStore defines operations for interacting with credentials
+// that are stored in Secret.
+type SecretStore interface {
```
ywk253100, line 29: Seems it isn't a store? It's just a secret getter and stores nothing. Can we call `kubeClient.Get()`  to get the secret directly without defining this interface?

If it's needed, I have the same comment with @qiuming-best , the function name `Buffer` confuses me

pkg/repository/provider/unified_repo.go
```diff
@@ -47,18 +48,30 @@ var getGCPCredentials = repoconfig.GetGCPCredentials
 var getS3BucketRegion = repoconfig.GetAWSBucketRegion
 var getAzureStorageDomain = repoconfig.GetAzureStorageDomain
 
+type localFuncTable struct {
```
ywk253100, line 51: Is the struct used for replacing the functions easily in unit tests? But seems these functions can receive mock input easily

pkg/test/fake_credential_secret_store.go
```diff
@@ -0,0 +1,47 @@
...
+
+package test
+
+import (
+	corev1api "k8s.io/api/core/v1"
+)
+
+// SecretStore defines operations for interacting with credentials
+// that are stored in Secret.
+type SecretStore interface {
```
ywk253100, line 25: Why define the same interface again here?

And it's better to generate the mock implementation by the [mockery tool](https://github.com/vektra/mockery)

internal/credentials/secret_store.go
Lyndon-Li: We add this SecretStore in parallel with the existing FileStore, the FileStore gets the password from Secret then save to a temp file; SecretStore gets the password from Secret and then return a stream buffer.

We don't call `kubeClient.Get()` directly for below purpose:

- We may get the password/credential keys from different sources, for example, from a temp file, from memory, from KMS, so we wrap the interfaces for all the sources in the `CredentialGetter` structure.
- At present, in Unified Repo, we are using repo password got from memory, cloud provider credentials from a temp file
- In future, we may add a new interface called something like KmsStore

pkg/test/fake_credential_secret_store.go
Lyndon-Li: OK, will use mockery.
I added in this way just followed the existing file_store.go. But for new code, yes, let's use the better way.

pkg/repository/provider/unified_repo.go
Lyndon-Li: Yes, it is for mock the local functions for UT.
As the discussion, we will keep them as it for now, if we have better way in future, we come back to change it.

internal/credentials/secret_store.go
Lyndon-Li: For the method name of SecretStore, changed from `Buffer` to `Get`

pkg/repository/provider/unified_repo.go
Lyndon-Li: Done

pkg/test/fake_credential_secret_store.go
Lyndon-Li: Done

qiuming-best (approved):

ywk253100 (approved):

PR #5179, opened 2022-08-04
Kopia Integration: Unified Repository Provider - Implementation

Lyndon-Li:
Add changes for Kopia Integration: Unified Repository Provider - method implementation
Related Issue: #https://github.com/vmware-tanzu/velero/issues/5079


pkg/repository/provider/unified_repo.go
```diff
@@ -129,7 +205,33 @@ func (urp *unifiedRepoProvider) EnsureUnlockRepo(ctx context.Context, param Repo
...
+	if err != nil {
+		return errors.Wrap(err, "error to get repo options")
+	}
+
+	bkRepo, err := urp.repoService.Open(ctx, repoOption)
+	if err != nil {
+		return errors.Wrap(err, "error to open backup repo")
+	}
+
+	defer bkRepo.Close(ctx)
```
reasonerjt, line 226: Log the error?

pkg/repository/provider/unified_repo.go
```diff
@@ -147,12 +249,12 @@ func getRepoPassword(secretStore credentials.SecretStore, param RepoParam) (stri
 }
 
 func (urp *unifiedRepoProvider) getRepoOption(param RepoParam) (udmrepo.RepoOptions, error) {
-	repoPassword, err := funcTable.getRepoPassword(urp.credentialGetter.FromSecret, param)
+	repoPassword, err := getRepoPassword(urp.credentialGetter.FromSecret, param)
```
reasonerjt, line 252: Not about the change in this line, the conversion in line 248 is not necessary `buf` is already a string type, it is also strange why we call a secret `buf`.

pkg/repository/udmrepo/service/service.go
```diff
@@ -0,0 +1,66 @@
...
+	if workPath == "" {
+		workPath = getDefaultWorkPath()
+	}
+
+	name := "repo-" + strings.ToLower(repoID) + ".conf"
+
+	return filepath.Join(workPath, name)
+}
+
+func GetOpenOption(password string, repoID string, description string) udmrepo.RepoOptions {
```
reasonerjt, line 49: This looks repetitive to `getRepoOption`  or `getRepoOptionForOpen`

pkg/repository/provider/unified_repo.go
```diff
@@ -185,6 +283,22 @@ func (urp *unifiedRepoProvider) getRepoOption(param RepoParam) (udmrepo.RepoOpti
 	return repoOption, nil
 }
 
+func (urp *unifiedRepoProvider) getRepoOptionForOpen(param RepoParam, purpose string) (udmrepo.RepoOptions, error) {
```
reasonerjt, line 286: duplicate code with `getRepoOption`? 
Looks like it would be better to apply _Options Pattern_ here to construct the repoOptions:
https://golang.cafe/blog/golang-functional-options-pattern.html

pkg/repository/provider/unified_repo.go
```diff
@@ -185,6 +283,22 @@ func (urp *unifiedRepoProvider) getRepoOption(param RepoParam) (udmrepo.RepoOpti
...
+	repoPassword, err := getRepoPassword(urp.credentialGetter.FromSecret, param)
+	if err != nil {
+		return udmrepo.RepoOptions{}, errors.Wrap(err, "error to get repo password")
+	}
+
+	repoOption := udmrepo.RepoOptions{
+		RepoPassword:   repoPassword,
+		ConfigFilePath: getRepoConfigFile(urp.workPath, string(param.BackupLocation.UID)),
+		Description:    purpose,
```
reasonerjt, line 295: It's a little confusing to set `purpose` to `Description`, is `purpose` a few constants that only kopia understands, or is it to control the flow, or is it just for logging?

By reading the code where `getRepoOptionForOpen` is called it looks like it's used to control the behavior of kopia, if that's the case it probably should not be `Description` in the option.

pkg/repository/provider/unified_repo.go
```diff
@@ -105,22 +99,104 @@ func (urp *unifiedRepoProvider) InitRepo(ctx context.Context, param RepoParam) e
...
 func (urp *unifiedRepoProvider) PruneRepoQuick(ctx context.Context, param RepoParam) error {
-	///TODO
+	log := urp.log.WithFields(logrus.Fields{
+		"BSL name": param.BackupLocation.Name,
+		"BSL UID":  param.BackupLocation.UID,
+	})
+
+	log.Debug("Start to prune repo quick")
+
+	repoOption, err := urp.getRepoOptionForOpen(param, "quick maintenance")
```
reasonerjt, line 186: See my comment in the func `getRepoOptionForOpen` I think we should also set the value of `purpose` as a group of constants?

pkg/repository/provider/unified_repo.go
Lyndon-Li: The Description option defined by Unified Repo interface is not a option to control anything, it is basically for logging, for example, when opening a repository with the RepoOptions, we are able to know a description of, for example, the purpose of opening this repo, whether it is for backup, restore or maintenance?

This is not only an option that is understood by Kopia only, actually, anything we define in the Unified Repo interface should not be for Kopia only.

So here let me remove the parameter of "purpose" and use everything as "description", is it better?

reasonerjt: Thanks for the clarification.  I think it is better.  Could you also add some comment in the definition of the struct `RepoOptions` to clarify it should be used for logging only?

pkg/repository/provider/unified_repo.go
Lyndon-Li: Done

pkg/repository/provider/unified_repo.go
Lyndon-Li: Done, changed the variable name and removed the force conversion.

pkg/repository/provider/unified_repo.go
Lyndon-Li: Done, followed the Options Pattern.

pkg/repository/provider/unified_repo.go
Lyndon-Li: Done, added the comment.

pkg/repository/udmrepo/service/service.go
Lyndon-Li: Followed the Options Pattern to remove the confuse of repetitiveness.

pkg/repository/provider/unified_repo.go
Lyndon-Li: Done

reasonerjt (approved):

reasonerjt (approved):

pkg/repository/provider/unified_repo.go
```diff
@@ -105,22 +111,124 @@ func (urp *unifiedRepoProvider) InitRepo(ctx context.Context, param RepoParam) e
 }
 
 func (urp *unifiedRepoProvider) ConnectToRepo(ctx context.Context, param RepoParam) error {
-	///TODO
+	log := urp.log.WithFields(logrus.Fields{
+		"BSL name": param.BackupLocation.Name,
+		"BSL UID":  param.BackupLocation.UID,
+	})
```
qiuming-best, line 117: All the member function in unifiedRepoProvider has the same log fields 'BSL name', 'BSL UID'. 
cloud set the field in unifiedRepoProvider initialization?

pkg/repository/udmrepo/service/service.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func Create(logger logrus.FieldLogger) udmrepo.BackupRepoService {
+	///TODO: create from kopiaLib
+	return nil
+}
+
+func GetRepoUser() (username, domain string) {
+	return defaultUsername, defaultDomain
+}
```
qiuming-best, line 37: In one of my functions that needs to get Repo User info, so I must need to import `service` package, but nothing to do with `udmrepo service`, should it be one configuration for `udmrepo` rather than configuration in service?

pkg/repository/provider/unified_repo.go
Lyndon-Li: The BSL is a param of the specific function calls, it also means, the BSL could be different for each function call.
Therefore, we don't have the BSL information at the time of the initialization.

pkg/repository/udmrepo/service/service.go
Lyndon-Li: As the discussion, the `service` package means the package to provide all the methods exposed by the unified repo. In this way, it is rational to put the `GetRepoUser` in the `service` package.

However, the package name may be misleading that the package binds to the `BackupRepoService` interface. Therefore:

- Let's add some comments to clarify the usage of the package
- When we see a better name in future, we will change the package name

pkg/repository/provider/unified_repo.go
```diff
@@ -129,60 +237,108 @@ func (urp *unifiedRepoProvider) EnsureUnlockRepo(ctx context.Context, param Repo
...
 	for k, v := range storeCred {
-		repoOption.StorageOptions[k] = v
+		storeOptions[k] = v
 	}
 
-	return repoOption, nil
+	return storeOptions, nil
+}
+
+func getRepoPassword(secretStore credentials.SecretStore, param RepoParam) (string, error) {
```
qiuming-best, line 331: could getRepoPassword be GetRepoPassword？I need reuse this function in other package

Lyndon-Li: `unified_repo.go` is an implementation of `repository.Provider` interface, I don't think it is a good idea to call any functions in this file without through the interface. Moreover, in any other places, the caller should not understand what a `RepoParam` is.

If the uploader wants to get the repo password, it can get it in the similar way of `getRepoPassword`, but the uploader doesn't have to use the same way necessarily. This means, it is reasonable for the uploader to have its own implementation.
Moreover, the logic of `getRepoPassword` is very little, so uploader doesn't have to share one function with unified_repo.go.

qiuming-best: ok, I'll implement it my own package

qiuming-best (approved):

Issue #5096, opened 2022-07-05
podvolumerestore controller to handle different uploader-types


reasonerjt:
Make sure the podvolumerestore controller will trigger different code path based on the uploader-type of the CR.

Yggdrasil:
Duplicate of https://github.com/vmware-tanzu/velero/issues/5095

Issue #5095, opened 2022-07-05
podvolumebackup controller to handle different uploader-types


reasonerjt:
Make sure the podvolumebackup controller will trigger different code path based on the `uploader-type` of the CR.

Issue #5074, opened 2022-07-04
File System Uploader - Restore


Lyndon-Li:
Epic: #5050 
Design: #4926 

Add the code for File System Uploader restore, include:

- Setup Kopia Uploader for Velero, i.e. Kopia Uploader options and policies for restore
- Call Kopia Uploader to restore from a specified snapshot ID
- Restore output to upper layer
- Restore progress monitor and adapter
- UT design and implementation

qiuming-best:
codes merged

Issue #5073, opened 2022-07-04
File System Uploader - Backup


Lyndon-Li:
Epic: #5050 
Design: #4926 

Add the code for File System Uploader backup, include:

- Setup Kopia Uploader for Velero, i.e. Kopia Uploader options and policies for backup
- Call Kopia Uploader to back up a specified path
- Backup snapshot information design, preservation and retrieve
- Backup output to upper layer
- Backup progress monitor and adapter
- UT design and implementation

qiuming-best:
Codes merged

PR #5214, opened 2022-08-16
Uploader Implementation: Restic backup and restore

qiuming-best:
Signed-off-by: Ming <mqiu@vmware.com>

- uploader module implementation, which related to restic implementation of backup and restore
- pvb & pvr controller to handle different uploader-types

which reference to #5074 #5073 #5095 #5096


pkg/uploader/provider/restic.go
```diff
@@ -13,22 +13,280 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+var ResticRunCMDFunc = exec.RunCommand
+var ResticGetSnapshotSizeFunc = restic.GetSnapshotSize
+var ResticGetVolumeSizeFunc = restic.GetVolumeSize
+var ResticRestoreFunc = restic.RestoreCommand
+
+type resticProvider struct {
+	repoIdentifier  string
+	credentialsFile string
+	caCertFile      string
+	repoEnv         []string
```
Lyndon-Li, line 49: Better to remove it as `cmdEnv`, this env is not seen anything specific connection to a repository

pkg/uploader/provider/restic.go
```diff
@@ -13,22 +13,280 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+	backupCmd.CACertFile = rp.caCertFile
+
+	if parentSnapshot != "" {
+		backupCmd.ExtraFlags = append(backupCmd.ExtraFlags, fmt.Sprintf("--parent=%s", parentSnapshot))
+	}
+
+	// #4820: restrieve insecureSkipTLSVerify from BSL configuration for
+	// AWS plugin. If nothing is return, that means insecureSkipTLSVerify
+	// is not enable for Restic command.
+	skipTLSRet := restic.GetInsecureSkipTLSVerifyFromBSL(rp.bsl, rp.log)
```
Lyndon-Li, line 141: Looks like this call can also be moved to `NewResticUploaderProvider`, so that there is no code redundancy in `RunBackup` and `RunRestore`

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -147,47 +144,72 @@ func (r *PodVolumeBackupReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-	// #4820: restrieve insecureSkipTLSVerify from BSL configuration for
-	// AWS plugin. If nothing is return, that means insecureSkipTLSVerify
-	// is not enable for Restic command.
-	skipTLSRet := restic.GetInsecureSkipTLSVerifyFromBSL(backupLocation, log)
-	if len(skipTLSRet) > 0 {
-		resticCmd.ExtraFlags = append(resticCmd.ExtraFlags, skipTLSRet)
+	// name of ResticRepository is generated with prefix volumeNamespace-backupLocation- and end with random characters
+	// it could not retrieve the ResticRepository CR with namespace + name. so first list all CRs with in the volumeNamespace
+	// then filtering the matched CR with prefix volumeNamespace-backupLocation-
+	backupRepos := &velerov1api.BackupRepositoryList{}
```
Lyndon-Li, line 150: Put the logic to get backupRepository to a dedicate file, so as to remove code redundancy between PVB and PVR?

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -147,47 +144,72 @@ func (r *PodVolumeBackupReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	// changed since the PVC's last backup, restic will not be able to identify a suitable
+	// parent snapshot to use, and will have to do a full rescan of the contents of the PVC.
+	var parentSnapshotID string
+	if pvcUID, ok := pvb.Labels[velerov1api.PVCUIDLabel]; ok {
+		parentSnapshotID = r.getParentSnapshot(ctx, log, pvb.Namespace, pvcUID, pvb.Spec.BackupStorageLocation)
+		if parentSnapshotID == "" {
+			log.Info("No parent snapshot found for PVC, not using --parent flag for this backup")
 		} else {
-			return r.updateStatusToFailed(ctx, &pvb, err, fmt.Sprintf("running Restic backup, stderr=%s", stderr), log)
+			log.WithField("parentSnapshotID", parentSnapshotID).Info("Setting --parent flag for this backup")
```
Lyndon-Li, line 190: The log is not accurate anymore, because `--parent` flag is for Restic only, so what about changing this log message?

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -147,47 +144,72 @@ func (r *PodVolumeBackupReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-		snapshotID, err = r.ResticExec.GetSnapshotID(cmd)
-		if err != nil {
-			return r.updateStatusToFailed(ctx, &pvb, err, "getting snapshot id", log)
+	var emptySnapshot bool
+	snapshotID, err := uploaderProv.RunBackup(ctx, path, pvb.Spec.Tags, parentSnapshotID, r.NewBackupProgressUpdater(&pvb, log, ctx))
+	if err != nil {
+		if strings.Contains(err.Error(), "snapshot is empty") {
+			emptySnapshot = true
+		} else {
+			return r.updateStatusToFailed(ctx, &pvb, err, fmt.Sprintf("running Restic backup, stderr=%v", err), log)
```
Lyndon-Li, line 206: "running Restic backup, stderr=%v" to "running backup, stderr=%v"?

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -261,38 +249,46 @@ func (c *PodVolumeRestoreReconciler) processRestore(ctx context.Context, req *ve
...
+	defer func() {
+		if err := uploaderProv.Close(ctx); err != nil {
+			log.Errorf("failed to close uploader provider with error %v", err)
+		}
+	}()
 
-	if stdout, stderr, err = restic.RunRestore(resticCmd, log, c.updateRestoreProgressFunc(req, log)); err != nil {
-		return errors.Wrapf(err, "error running restic restore, cmd=%s, stdout=%s, stderr=%s", resticCmd.String(), stdout, stderr)
+	if err = uploaderProv.RunRestore(ctx, req.Spec.SnapshotID, volumePath, c.NewRestoreProgressUpdater(req, log, ctx)); err != nil {
+		return errors.Wrapf(err, "error running restic restore err=%v", err)
```
Lyndon-Li, line 290: "error running restic restore err=%v" to "error running restore err=%v"

pkg/builder/backup_repository_builer.go
```diff
@@ -0,0 +1,59 @@
...
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+)
+
+// BackupRepositoryBuilder builds BackupRepository objects.
+type BackupRepositoryBuilder struct {
+	object *velerov1api.BackupRepository
+}
+
+// ForBackupRepository is the constructor for a BackupRepositoryBuilder.
+func ForBackupRepository(ns, name string) *BackupRepositoryBuilder {
```
Lyndon-Li, line 31: Looks like this builder is used by the testing code only, while other builders in pkg/builder are used formally.
I compared the code in pkg/repository/ensurer.go, where the backupRepository object is created, the code there is like:
```
	repo := &velerov1api.BackupRepository{
			Namespace:    namespace,
			GenerateName: fmt.Sprintf("%s-%s-", volumeNamespace, backupLocation),
			Labels:       repoLabels(volumeNamespace, backupLocation, repositoryType),
		},
		Spec: velerov1api.BackupRepositorySpec{
			VolumeNamespace:       volumeNamespace,
			BackupStorageLocation: backupLocation,
		},
	}
```
We see differences, so I am afraid the code here doesn't work if we use it formally. Therefore:

- either we move it to the testing code
- or we make it the same with the code in ensurer.go and it's better to remove the code redundancy as well

pkg/uploader/provider/restic.go
qiuming-best: Do you mean to rename it to cmdEnv? If so, I've renamed it

pkg/uploader/provider/restic.go
qiuming-best: I've changed

pkg/controller/pod_volume_backup_controller.go
qiuming-best: currently, I couldn't find a proper directory, I put it into `pkg/util/crd/backup_repository.go`

pkg/controller/pod_volume_backup_controller.go
qiuming-best: yeah, I've changed

pkg/controller/pod_volume_backup_controller.go
qiuming-best: yeah, I've changed

pkg/controller/pod_volume_restore_controller.go
qiuming-best: yeah, I've changed

pkg/builder/backup_repository_builer.go
qiuming-best: I defined one function in testing code

pkg/controller/pod_volume_backup_controller.go
Lyndon-Li: Personally, I don't see problems with this directory, let's see other reviewers' suggestion.

Lyndon-Li (dismissed):

pkg/uploader/provider/restic.go
```diff
@@ -13,22 +13,162 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+	if updater == nil {
+		return "", errors.New("Need to initial backup progress updater first")
+	}
+
+	log := rp.log.WithFields(logrus.Fields{
+		"path":           path,
+		"parentSnapshot": parentSnapshot,
+	})
+
+	rp.backupTags = tags
```
ywk253100, line 124: The `backupTags` will be overwritten if call `RunBackup` in paralell

pkg/uploader/provider/restic.go
```diff
@@ -13,22 +13,162 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+	skipTLSRet := restic.GetInsecureSkipTLSVerifyFromBSL(bsl, log)
+	if len(skipTLSRet) > 0 {
+		provider.extraFlags = append(provider.extraFlags, skipTLSRet)
+	}
+
+	return &provider, nil
+}
+
+// Not implement yet
+func (rp *resticProvider) Cancel() {
```
ywk253100, line 94: Seems this function doesn't exist in the interface definition

pkg/uploader/provider/provider.go
```diff
@@ -49,23 +54,34 @@ type Provider interface {
...
 	credGetter *credentials.CredentialGetter,
 	repoKeySelector *v1.SecretKeySelector,
 	log logrus.FieldLogger,
 ) (Provider, error) {
+	if credGetter.FromFile == nil {
+		return nil, errors.New("uninitialized FileStore credentail is not supported")
+	}
 	if uploaderType == uploader.KopiaType {
-		return NewResticUploaderProvider(repoIdentifier, bsl, credGetter, repoKeySelector, log)
+		if err := provider.NewUnifiedRepoProvider(*credGetter, log).ConnectToRepo(ctx, provider.RepoParam{BackupLocation: bsl, BackupRepo: backupRepo}); err != nil {
```
ywk253100, line 76: Why do we need to create repo provider here?

pkg/controller/pod_volume_restore_controller.go
```diff
@@ -262,38 +249,26 @@ func (c *PodVolumeRestoreReconciler) processRestore(ctx context.Context, req *ve
...
-	// if there's a caCert on the ObjectStorage, write it to disk so that it can be passed to restic
-	var caCertFile string
-	if backupLocation.Spec.ObjectStorage != nil && backupLocation.Spec.ObjectStorage.CACert != nil {
-		caCertFile, err = restic.TempCACertFile(backupLocation.Spec.ObjectStorage.CACert, req.Spec.BackupStorageLocation, c.fileSystem)
-		if err != nil {
-			log.WithError(err).Error("Error creating temp cacert file")
-		}
-		// ignore error since there's nothing we can do and it's a temp file.
-		defer os.Remove(caCertFile)
+	backupRepo, err := crd.GetBackupRepositoryByFilter(ctx, c.Client, req.Namespace, req.Spec.Pod.Namespace, req.Spec.BackupStorageLocation)
```
ywk253100, line 252: The `PodVolumeBackup` and `PodVolumeRestore` should contain enough information about where the data is put to or get from.
Currently, there is a `RepoIdentifier` in PVB/PVR, if it isn't enough, we should add more info rather than guessing the repo name

pkg/controller/pod_volume_backup_controller.go
```diff
@@ -147,47 +145,52 @@ func (r *PodVolumeBackupReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 		}
+	}()
 
-		snapshotID, err = r.ResticExec.GetSnapshotID(cmd)
-		if err != nil {
-			return r.updateStatusToFailed(ctx, &pvb, err, "getting snapshot id", log)
+	var emptySnapshot bool
+	snapshotID, err := uploaderProv.RunBackup(ctx, path, pvb.Spec.Tags, parentSnapshotID, r.NewBackupProgressUpdater(&pvb, log, ctx))
+	if err != nil {
+		if strings.Contains(err.Error(), "snapshot is empty") {
```
ywk253100, line 184: Seems the logic is only valid for Restic? What about the Kopia, does it has the same concept?

pkg/uploader/provider/restic.go
qiuming-best: yeah, I've removed the backupTags by passing  tags directly into `GetSnapshotCommand`

pkg/uploader/provider/provider.go
qiuming-best: For every `NewUploaderProvider`, which is running in the daemonset, It's a start of backup&restore  for every `PVB` or `PVR`, we need to use the latest configuration in BSL to check the repository connection at first when provider initializes.

pkg/controller/pod_volume_restore_controller.go
Lyndon-Li: `RepoIdentifier` is not enough since we have two path, the Restic path and the unified path. If we have two `BackupRepository` for the two paths representatively but they are from the same storage config, `RepoIdentifier` will be the same.
Moreover, at present, `RepoIdentifier` is used by Restic path only.

Maybe we can follow the way in `func (r *RepositoryEnsurer) EnsureRepo` to retrieve the `BackupRepository` object?

qiuming-best: In `func (r *RepositoryEnsurer) EnsureRepo`, it will triger to create related `BackupRepository` and watch the status then retrieve the `BackupRepository `

Here, we don't do anything to update the related `BackupRepository`. Maybe it's not proper.
And also it's a little complicated when you just want to get related `BackupRepository` by watching resources.

Here I just want to get `BackupRepository`, The directest way is using `kubeclient.Get(ctx, NamespacedName)`, we know the namespace, and if we can store Repository Name on some CRD will be good

Lyndon-Li: I mean follow the way in the function to get the `BackupRepository`, that is, use the label selector:
```
	selector := labels.SelectorFromSet(repoLabels(volumeNamespace, backupLocation))
	repos, err := r.repoLister.BackupRepositories(namespace).List(selector)
```

ywk253100: I agree with @qiuming-best, adding the backup repo name to the PVB/PVR may be more reasonable? The name is unique and can identify the repository

pkg/controller/pod_volume_backup_controller.go
qiuming-best: yeah, I've added logic in `kopia`,  when backup empty dir which will also return one error with "snapshot is empty"

pkg/controller/pod_volume_restore_controller.go
qiuming-best: Now, I've changed into `client.List(ctx, backupRepoList, &client.ListOptions{
		Namespace:     pvbNamespace,
		LabelSelector: selector,
	})`
the label selector which include `volumeNamespace`, `BackupStorageLocation`, `repositoryType`

Lyndon-Li: @ywk253100 Let PVB/PVR directly refer to the BackupRepository name doesn't solve all the problems -- for old PVB/PVR, they don't have this reference, on the other hand, multiple PVB/PVR may refer to one repository, so below situation may happen:

- Two old PVBs refer to one ResticRepository
- When we start PVB1, since its repo reference is empty, we fail back to create a BackupRepository for Restic
- When we start PVB2, its repo reference is also empty, but this time, we should find a way to search the existing BackupRepository instead of creating a new one

As you can see, we still need a method of "Search".

Moreover, as the previous discussion, our current approach to handle BackupRepository is not enough for Unified Repo when there are multiple repo providers, besides kopia, so we are using some hardcode things (for details, refer to [this comment](https://github.com/vmware-tanzu/velero/pull/5259#discussion_r959516173)). This means, in future, we will face refactor of these areas, therefore, the good enough and simpler solution is preferred.

pkg/controller/pod_volume_backup_controller.go
ywk253100: Determining whether the snapshot is empty or not by checking the error message isn't a good way and it's hard for the new implementation of the provider to note this.

Suggest defining a new return value to indicate whether the backup is empty for the `RunBackup()` of the provider interface.

qiuming-best: Done

pkg/controller/pod_volume_backup_controller.go
qiuming-best: I've moved the `pkg/util/crd/backup_repository.go` into `pkg/repository/util/backup_repo_op.go`

pkg/uploader/provider/provider.go
```diff
@@ -40,7 +45,7 @@ type Provider interface {
 		path string,
 		tags map[string]string,
 		parentSnapshot string,
-		updater uploader.ProgressUpdater) (string, error)
+		updater uploader.ProgressUpdater) (string, bool, error)
```
ywk253100, line 48: Please update the comment for the new returning value

pkg/uploader/provider/restic.go
```diff
@@ -13,22 +13,160 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+	backupCmd.Env = rp.cmdEnv
+	backupCmd.CACertFile = rp.caCertFile
+	backupCmd.ExtraFlags = rp.extraFlags
+	if parentSnapshot != "" {
+		backupCmd.ExtraFlags = append(backupCmd.ExtraFlags, fmt.Sprintf("--parent=%s", parentSnapshot))
+	}
+
+	summary, stderrBuf, err := restic.RunBackup(backupCmd, log, updater)
+	if err != nil {
+		if strings.Contains(err.Error(), "snapshot is empty") {
```
ywk253100, line 130: If the snapshot is empty, return `"", true, nil`, it's easier for the upper caller to handle. The same to the kopia implementation

pkg/uploader/provider/provider.go
qiuming-best: I've added comments

pkg/uploader/provider/restic.go
qiuming-best: I've changed

ywk253100 (approved):

Lyndon-Li (approved):

Issue #5077, opened 2022-07-04
Unified Repository - BackupRepoService


Lyndon-Li:
Epic: #5050 
Design: #4926 

Add the code to implement BackupRepoService functions along with UT:

- Init
- Connect
- Open
- Full Maintenance
- Quick Maintenance
- Maintenance Policy

PR #5219, opened 2022-08-16
Kopia Integration:Kopia Lib - Initialization

Lyndon-Li:
Add changes for Kopia Integration: Kopia Lib - initialize Kopia repo
Related Issue: #5077


reasonerjt:
Please add proper comments to the interfaces and exposed functions.

Lyndon-Li:
@reasonerjt Done adding the comments. Thanks.

pkg/repository/udmrepo/kopialib/backend/utils.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+func optionalHaveString(key string, flags map[string]string) string {
+	return optionalHaveStringWithDefault(key, flags, "")
+}
+
+func optionalHaveBool(key string, flags map[string]string) bool {
+	if value, exist := flags[key]; exist {
+		ret, err := strconv.ParseBool(value)
```
reasonerjt, line 40: Although it's optional we should not eat the error, same issue for other `optionalHavexxx` funcs in this file.

pkg/repository/udmrepo/kopialib/backend/file_system.go
```diff
@@ -0,0 +1,81 @@
...
+	if !filepath.IsAbs(c.options.Path) {
+		return nil, errors.Errorf("filesystem repository path must be absolute")
+	}
+
+	return filesystem.New(ctx, &c.options, isCreate)
+}
+
+func resolveUserFriendlyPath(path string, relativeToHome bool) string {
+	home, _ := os.UserHomeDir()
+	if home != "" && strings.HasPrefix(path, "~") {
```
reasonerjt, line 68: Why is this needed?  This may cause issue if the input `path` is like `~some/path` and it may also cause issues in cross OS support.
Can't we make it simpler to force it to be an absolute path?

pkg/repository/udmrepo/kopialib/repo_init.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+type kopiaBackendStore struct {
+	name        string
+	description string
+	store       backend.BackendStore
+}
+
+// backendStores lists the supported backend storages at present
+var backendStores []kopiaBackendStore = []kopiaBackendStore{
```
reasonerjt, line 38: Would it be easier to set it as a map rather than a list?

Lyndon-Li: Consider below two things, a list may be better than a map?

1. We want to bring some flexibility of the search, for example, we want to do a case insensitive compare of the `name`, because the comparing value comes from a string map of `RepoOptions`, we don't want to restrict the creator of `RepoOptions` with so many details
2. The name of `kopiaBackendStore` is an essential information of a Store, it is not only a key for the search. It means, once we have a kopiaBackendStore, we want to get its name by kopiaBackendStore.name instead of find it elsewhere, i.e., the key of a map

pkg/repository/udmrepo/repo-options.go
```diff
@@ -32,6 +32,9 @@ const (
 	GenOptionMaintainFull  = "full"
 	GenOptionMaintainQuick = "quick"
 
+	GenOptionOwnerName   = "username"
```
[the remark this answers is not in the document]
Lyndon-Li: Done, thanks.

pkg/repository/udmrepo/kopialib/backend/backend.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"context"
+
+	"github.com/kopia/kopia/repo/blob"
+)
+
+// BackendStore defines the methods for Kopia to establish a connection to
+// the backend storage
+type BackendStore interface {
```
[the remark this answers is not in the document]
Lyndon-Li: Done

pkg/repository/udmrepo/kopialib/backend/file_system.go
Lyndon-Li: Done, removed the processing of relative path.

pkg/repository/udmrepo/repo-options.go
```diff
@@ -153,13 +182,24 @@ func WithStoreOptions(getter StoreOptionsGetter, param interface{}) func(*RepoOp
 	}
 }
 
+// WithConfigFile sets the Description to RepoOptions
```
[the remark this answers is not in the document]
Lyndon-Li: Corrected, thanks.

pkg/repository/udmrepo/repo-options.go
```diff
@@ -153,13 +182,24 @@ func WithStoreOptions(getter StoreOptionsGetter, param interface{}) func(*RepoOp
...
 		return nil
 	}
 }
 
+// GetRepoUser returns the default username that is used to manipulate the Unified Repo
+func GetRepoUser() string {
+	return defaultUsername
+}
+
+// GetRepoUser returns the default user domain that is used to manipulate the Unified Repo
```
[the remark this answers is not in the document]
Lyndon-Li: Corrected, thanks.

pkg/repository/udmrepo/kopialib/backend/utils.go
Lyndon-Li: Done, log the conversion error in place

pkg/repository/udmrepo/kopialib/repo_init.go
reasonerjt: OK I think it's a minor issue.  Let's keep it as a list for now.

pkg/repository/udmrepo/kopialib/backend/file_system.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (c *FsBackend) Setup(ctx context.Context, flags map[string]string) error {
+	path, err := mustHaveString(udmrepo.StoreOptionFsPath, flags)
+	if err != nil {
+		return err
+	}
+
+	prefix := optionalHaveString(udmrepo.StoreOptionPrefix, flags)
+
+	c.options.Path = filepath.Join(path, prefix)
```
reasonerjt, line 47: Could you elaborate on how the prefix is used?  Looks like it's a sub-directory for kopia, not really a prefix.

Do we really need it?

pkg/repository/udmrepo/kopialib/backend/file_system.go
```diff
@@ -0,0 +1,62 @@
...
+	c.options.DirectoryMode = defaultDirMode
+
+	c.options.Limits = setupLimits(ctx, flags)
+
+	return nil
+}
+
+func (c *FsBackend) Connect(ctx context.Context, isCreate bool) (blob.Storage, error) {
+	if !filepath.IsAbs(c.options.Path) {
+		return nil, errors.Errorf("filesystem repository path must be absolute")
```
reasonerjt, line 58: would it be slightly better if the path is also in the error msg?

pkg/repository/udmrepo/kopialib/backend/file_system.go
Lyndon-Li: The prefix for file system is for the same purpose as other storages. It is not Kopia specific.

For example, for Restic + S3, the prefix is like `Restic/<BSL.Spec.VolumeNamespace>`, that is, we will find the backup data under `<bucket name>/Restic/<BSL.Spec.VolumeNamespace>`.
For unified repo, we use the similar prefix pattern, the data can be found under `<bucket name>/unified-repo/<BSL.Spec.VolumeNamespace>`. 

The prefix pattern works for all kinds of storages, include file system. Therefore, with this prefix, for a file system, the data is saved under `<repo root folder>/unified-repo/<BSL.Spec.VolumeNamespace>`

pkg/repository/udmrepo/kopialib/backend/file_system.go
Lyndon-Li: Done, added the path. Thanks.

pkg/repository/udmrepo/kopialib/backend/file_system.go
reasonerjt: make sense, thanks.

reasonerjt (approved):
lgtm

pkg/repository/udmrepo/kopialib/backend/common.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/kopia/kopia/repo/splitter"
+
+	"github.com/vmware-tanzu/velero/pkg/repository/udmrepo"
+)
+
+const (
+	maxDataCacheMB         = 2000
+	maxMetadataCacheMB     = 2000
+	maxCacheDurationSecond = 30
+)
```
qiuming-best, line 39: What is the reference setting for these three theoretical values?

pkg/repository/udmrepo/kopialib/backend/common.go
```diff
@@ -0,0 +1,83 @@
...
+
+func setupLimits(ctx context.Context, flags map[string]string) throttling.Limits {
+	return throttling.Limits{
```
qiuming-best, line 42: Actually, do we set all configurations into 0? 

and if setting it to 0, Dose kopia will not do any limitation?

Lyndon-Li: Yes, Kopia treats it as unlimited

pkg/repository/udmrepo/kopialib/backend/common.go
Lyndon-Li: This is according to Kopia's default value:
maxDataCacheMB: 5G
maxMetadataCacheMB: 5G
maxCacheDurationSecond: 30s

The default values are somehow aggressive for docker env, where the default root disk size is 10G, so I changed the size values to 2G respectively. Let's see it in the performance testing, if we find problem, we change it accordingly.

Issue #5081, opened 2022-07-04
Unified Repository - Debug Log


Lyndon-Li:
Epic: #5050 
Design: #4926 

Add the code to output Kopia Repository debug log to Velero's log, include:

- Select and setup Kopia log modules
- Implement Kopia log interface

Lyndon-Li:
fixed by #5233

Issue #5078, opened 2022-07-04
Unified Repository - BackupRepo


Lyndon-Li:
Epic: #5050 
Design: #4926 

Add the code to implement functions into BackupRepo and its derived interfaces into Kopia Library along with UT, include:

- BackupRepo
- ObjectReader
- ObjectWriter

PR #5233, opened 2022-08-19
Kopia Integration:Kopia Lib - Implementation

Lyndon-Li:
Add changes for Kopia Integration: Kopia Lib - method implementation
Add changes to write Kopia Repository logs to Velero log
Related Issues: #5078, #5081


design/unified-repo-and-kopia-integration/unified-repo-and-kopia-integration.md
```diff
@@ -93,126 +93,140 @@ Velero by default uses the Unified Repository for all kinds of data movement, it
...
+	DeleteManifest(ctx context.Context, id ID) error
+
+	// Flush flushes all the backup repository data
+	Flush(ctx context.Context) error
+
+	// Time returns the local time of the backup repository. It may be different from the time of the caller
+	Time() time.Time
+
+	// Close closes the backup repository
+	Close(ctx context.Context) er
```
qiuming-best, line 145: er typo ?

pkg/repository/udmrepo/kopialib/lib_repo.go
```diff
@@ -0,0 +1,587 @@
...
+	}
+}
+
+func (ks *kopiaRepoService) Open(ctx context.Context, repoOption udmrepo.RepoOptions) (udmrepo.BackupRepo, error) {
+	repoConfig := repoOption.ConfigFilePath
+	if repoConfig == "" {
+		return nil, errors.New("invalid config file path")
+	}
+
+	if _, err := os.Stat(repoConfig); os.IsNotExist(err) {
```
ywk253100, line 113: Do we need to check other error?

Lyndon-Li: Seems not.
Here the check is only to fail earlier and give a clearer message when the config file doesn't exist.
If there is any other case that need to fail earlier, we can also do the check here, but right now, we don't have.

Later the function `openKopiaRepo` will read the config file, if any problem happens, it will return with the error. Therefore, we still have opportunity to catch the error.

ywk253100 (approved):

qiuming-best (approved):

Issue #5223, opened 2022-08-16
Fix Edge cases for already exists resources


shubham-pampattiwar:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is, and what commands you ran.)
On OpenShift 4.11 (Kubernetes 1.24) we are seeing an error(not warning) related to the restoration of volumesnapshotclass that it already exists, not reproducilbe on earlier versions. Restore fails with the following logs, This is added as an error by the restore controller because the create call itself is incomplete due to admission webhooks validation, had the create call executed and Velero had received an IsalreadyExists this would have been a warning.

```
time="2022-07-14T14:34:44Z" level=error msg="error restoring example-snapclass: admission webhook \"volumesnapshotclasses.snapshot.storage.k8s.io\" denied the request: default snapshot class: example-snapclass already exits for driver: pd.csi.storage.gke.io" logSource="pkg/restore/restore.go:1357" restore=openshift-adp/mysql-42edcb0a-0381-11ed-9e9d-902e163f806c
time="2022-07-14T14:34:45Z" level=error msg="Cluster resource restore error: error restoring volumesnapshotclasses.snapshot.storage.k8s.io/example-snapclass: admission webhook \"volumesnapshotclasses.snapshot.storage.k8s.io\" denied the request: default snapshot class: example-snapclass already exits for driver: pd.csi.storage.gke.io" logSource="pkg/controller/restore_controller.go:500" restore=openshift-adp/mysql-42edcb0a-0381-11ed-9e9d-902e163f806c

```

**What did you expect to happen:**
Restore should complete without errors.

**The following information will help us better understand what's going on**:

_If you are using velero v1.7.0+:_  
Please use `velero debug  --backup <backupname> --restore <restorename>` to generate the support bundle, and attach to this issue, more options please refer to `velero debug --help` 

_If you are using earlier versions:_  
Please provide the output of the following commands (Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)
- `kubectl logs deployment/velero -n velero`
- `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
- `velero backup logs <backupname>`
- `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
- `velero restore logs <restorename>`

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Velero version (use `velero version`): 
[...]

PR #5239, opened 2022-08-22
Fix edge cases for already exists resources

shubham-pampattiwar:
Signed-off-by: Shubham Pampattiwar <spampatt@redhat.com>

Fixes https://github.com/vmware-tanzu/velero/issues/5223

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -1249,12 +1249,31 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 		errs.Add(namespace, err)
 		return warnings, errs
 	}
-	if isAlreadyExistsError {
-		fromCluster, err := resourceClient.Get(name, metav1.GetOptions{})
-		if err != nil {
-			ctx.log.Infof("Error retrieving cluster version of %s: %v", kube.NamespaceAndName(obj), err)
-			warnings.Add(namespace, err)
-			return warnings, errs
+
+	// check if we want to treat the error as a warning, in some cases the creation call might not get executed due to object API validations
+	// and Velero might not get the already exists error type but in reality the object already exists
+	objectExists := false
+	var fromCluster *unstructured.Unstructured
+
+	if restoreErr != nil && !isAlreadyExistsError {
+		// check for the existence of the object in cluster, if no error then it implies that object exists
+		// and if err then we want to fallthrough and do another get call later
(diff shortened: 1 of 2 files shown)


kaovilai:
I have an alternative approach to this PR at https://github.com/openshift/velero/pull/188 fyi

pkg/restore/restore.go
```diff
@@ -1249,12 +1249,31 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
...
+		}
+	}
+
+	if isAlreadyExistsError || objectExists {
+		// do a get call if we did not run this previously i.e.
+		// we've only run this for errors other than isAlreadyExistError
+		if fromCluster == nil {
+			fromCluster, err = resourceClient.Get(name, metav1.GetOptions{})
+			if err != nil {
+				ctx.log.Infof("Error retrieving cluster version of %s: %v", kube.NamespaceAndName(obj), err)
```
reasonerjt, line 1273: I know this is copied from existing code, but seems it's more consistent (e.g with the code in line 1247 ) if we merge this `err` into `errors`.
We may also wanna change the log level.

sseago: @reasonerjt Hmm. I'm not sure, but yeah, you may be right. So this would be a change in behavior for situations not otherwise modified by this PR. So the scenario we're talking about is that the `Create` failed due to the object already existing, but now when we do a `Get` call (to determine whether the in-cluster object is different from the from-backup object), the Get call fails. Ordinarily this shouldn't happen. Off the top of my head, possible causes are:
1) Something outside of Velero just created this resource at almost the same time as Velero tried to restore and we're in a race scenario where a second Create fails due to an already existing object but the API server isn't yet finding it via Get.
2) A bug or malfunction in the APIServer

We're currently flagging this as a restore warning. If we want to stick with that, then we should probably change the log level from Info to Warn.

The other possibility is we bump this up to a restore error and the log level to Error as well.

@shubham-pampattiwar what do you think?

shubham-pampattiwar: @reasonerjt @sseago I agree with changing the log level to error and adding this error to errors (as the next piece of workflow depends on the object fetched from the cluster)

shubham-pampattiwar: Done, updated the PR.

sseago (approved):

reasonerjt (approved):

Issue #3565, opened 2021-03-12
Restore of empty backups should have a better failure message


dsu-igeek:
**What steps did you take and what happened:**
Created a backup with no entries in it

`velero backup create empty-backup-1 --include-cluster-resources=true  --include-namespaces tst-1234-velero --include-resources 'persistentvolumes=pvc-ee198b55-febf-4664-93e2-1e740a99b82e'

The include-resources is incorrect and no resources are selected.  The result is a backup with no items:

```
Name:         empty-backup-1
Namespace:    velero
Labels:       velero.io/storage-location=default
Annotations:  velero.io/source-cluster-k8s-gitversion=v1.19.1+vmware.2
              velero.io/source-cluster-k8s-major-version=1
              velero.io/source-cluster-k8s-minor-version=19

Phase:  Completed

Errors:    0
Warnings:  0

Namespaces:
  Included:  tst-1234-velero
  Excluded:  <none>

Resources:
  Included:        persistentvolumes=pvc-ee198b55-febf-4664-93e2-1e740a99b82e
  Excluded:        <none>
  Cluster-scoped:  included

Label selector:  <none>

Storage Location:  default

Velero-Native Snapshot PVs:  auto

TTL:  720h0m0s

Hooks:  <none>

Backup Format Version:  1.1.0

Started:    2021-03-11 15:29:21 -0800 PST
Completed:  2021-03-11 15:29:22 -0800 PST

Expiration:  2021-04-10 16:29:21 -0700 PDT

Total items to be backed up:  0
Items backed up:              0

Velero-Native Snapshots: <none included>
```

If you attempt to restore from this backup, it will fail with 'error parsing backup contents: directory "resources" does not exist'

`velero restore create empty-restore --from-backup empty-backup-1`

`velero restore describe empty-restore`

```
Name:         empty-restore
Namespace:    velero
Labels:       <none>
Annotations:  <none>

Phase:  PartiallyFailed (run 'velero restore logs empty-restore' for more information)

Started:    2021-03-11 15:31:54 -0800 PST
Completed:  2021-03-11 15:31:54 -0800 PST

Errors:
  Velero:   error parsing backup contents: directory "resources" does not exist
  Cluster:    <none>
  Namespaces: <none>

Backup:  empty-backup-1

Namespaces:
  Included:  all namespaces found in the backup
  Excluded:  <none>

Resources:
  Included:        *
[...]

PR #5258, opened 2022-08-29
Return err when backup resources is empty

cleverhu:

return error when back up resources is empty

Fixes https://github.com/vmware-tanzu/velero/issues/3565

### DIFF
--- a/changelogs/unreleased/5258-cleverhu
+++ b/changelogs/unreleased/5258-cleverhu
@@ -1,0 +1,1 @@
+Return err when backup resources is empty. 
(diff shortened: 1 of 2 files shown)


reasonerjt:
For #3565 we should first make sure the restore controller can handle empty backup.  B/C there may be some historical backups, after discussion @allenxu404  will handle it

For setting backup to `Failed` if the Item collection is empty, considering the backup may be triggered by the schedule we will need more discussion to make the decision if we want to mark it as Failed or not.  Personally, I think it's OK to leave it as-is.

sseago:
@reasonerjt I'd agree with that. As long as empty backup is handled properly on restore, I think empty backup is fine. It's not an error for there to be nothing that matches the backup spec -- could be that the user had just deleted an application and hasn't replaced it with anything new yet. It might be reasonable to add a *warning* if a backup is empty, but it should still be considered a valid backup, much as zero-size file is a valid file, or a zero-length string is a valid string.

PR #5259, opened 2022-08-30
Kopia Pod Volume Backup/Restore

Lyndon-Li:
Fill gaps for Kopia path of PVBR:

- Repo Manager with Unified Repo
- Uploader type to PVBR backupper/restorer
- Repository Type to BackupRepository controller
- Repository Type to Repo Ensurer

Related issues: #5095 #5096


config/crd/v1/bases/velero.io_backuprepositories.yaml
```diff
@@ -53,7 +53,7 @@ spec:
               repositoryType:
                 description: RepositoryType indicates the type of the backend repository
                 enum:
-                - kopia
+                - unified
```
reasonerjt, line 56: Where do you define the association between kopia and unified?  
From the perspective of code path, it's the unified repo, but from the API point of view, we have to provide a way for user to explicitly tell velero the type of the backup repository right?

pkg/cmd/server/server.go
```diff
@@ -228,7 +227,7 @@ func NewCommand(f client.Factory) *cobra.Command {
 	command.Flags().StringVar(&config.profilerAddress, "profiler-address", config.profilerAddress, "The address to expose the pprof profiler.")
 	command.Flags().DurationVar(&config.resourceTerminatingTimeout, "terminating-resource-timeout", config.resourceTerminatingTimeout, "How long to wait on persistent volumes and namespaces to terminate during a restore before timing out.")
 	command.Flags().DurationVar(&config.defaultBackupTTL, "default-backup-ttl", config.defaultBackupTTL, "How long to wait by default before backups can be garbage collected.")
-	command.Flags().DurationVar(&config.defaultResticMaintenanceFrequency, "default-restic-prune-frequency", config.defaultResticMaintenanceFrequency, "How often 'restic prune' is run for restic repositories by default.")
+	command.Flags().DurationVar(&config.repoMaintenanceFrequency, "default-restic-prune-frequency", config.repoMaintenanceFrequency, "How often 'restic prune' is run for restic repositories by default.")
```
reasonerjt, line 230: The description is incorrect.

Lyndon-Li: Corrected, thanks.

config/crd/v1/bases/velero.io_backuprepositories.yaml
Lyndon-Li: As the discussion, we will have the `repositoryType` reflect the final repo provider, so here we keep it as `kopia`.

In future, when we have more than one repo providers managed by Unified Repo, we will add this `repositoryType` to BSL, so that Unified Repo could choose the repo provider according to this field.
For now, since we have only one repo provider, for simplification, we don't have it in BSL, so we just make a hardcode of the `repositoryType`:

- For PVB/PVR, for Kopia path, we always set it to `kopia`
- For other data movements, we always set it to `kopia`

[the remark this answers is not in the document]
Lyndon-Li: As the discussion, we will have the `repositoryType` reflect the final repo provider, so here we keep it as `kopia`.

In future, when we have more than one repo providers managed by Unified Repo, we will add this `repositoryType` to BSL, so that Unified Repo could choose the repo provider according to this field.
For now, since we have only one repo provider, for simplification, we don't have it in BSL, so we just make a hardcode of the `repositoryType`:

- For PVB/PVR, for Kopia path, we always set it to `kopia`
- For other data movements, we always set it to `kopia`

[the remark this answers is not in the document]
Lyndon-Li: Done the changes.

pkg/repository/ensurer.go
```diff
@@ -130,14 +136,14 @@ func (r *RepositoryEnsurer) EnsureRepo(ctx context.Context, namespace, volumeNam
 
 	log.Debug("Acquired lock")
 
-	selector := labels.SelectorFromSet(repoLabels(volumeNamespace, backupLocation))
+	selector := labels.SelectorFromSet(repoLabels(volumeNamespace, backupLocation, repositoryType))
```
qiuming-best, line 139: For the existing `BackupRepository` for the older velero version, If the user didn't reinstall the velero 1.10, we cannot list `BackupRepository` by the new selector. because the new label is not added to the older `BackupRepository`. 

So for users who upgrade from an older version to 1.10, do they need to delete all crds and reuninstall velero?

pkg/repository/ensurer.go
```diff
@@ -130,14 +136,14 @@ func (r *RepositoryEnsurer) EnsureRepo(ctx context.Context, namespace, volumeNam
...
 
 	repos, err := r.repoLister.BackupRepositories(namespace).List(selector)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
 	if len(repos) > 1 {
-		return nil, errors.Errorf("more than one ResticRepository found for workload namespace %q, backup storage location %q", volumeNamespace, backupLocation)
+		return nil, errors.Errorf("more than one ResticRepository found for workload namespace %q, backup storage location %q, repository type %q", volumeNamespace, backupLocation, repositoryType)
```
qiuming-best, line 146: ResticRepository should be BackupRepository

pkg/repository/ensurer.go
Lyndon-Li: The legacy repos are not called `BackupRepository` but `ResticRepository`, therefore, we cannot list the legacy repos anyway.
And after the upgrade, the legacy repos named as `ResticRepository` will not be used anymore, for the old PVB/PVR, a new `BackupRepository` will be created; and the controller doesn't handle them in its reconciler either.

Therefore it is safe to leave it there and we need users to delete them manually.

pkg/repository/ensurer.go
Lyndon-Li: Done

changelogs/unreleased/5259-lyndon
```diff
@@ -0,0 +1,6 @@
+Fill gaps for Kopia path of PVBR:
```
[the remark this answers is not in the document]
Lyndon-Li: Done, changed to one line. Thanks.

ywk253100 (approved):

pkg/podvolume/backupper.go
```diff
@@ -107,7 +110,13 @@ func (b *backupper) BackupPodVolumes(backup *velerov1api.Backup, pod *corev1api.
 		return nil, nil
 	}
 
-	repo, err := b.repoEnsurer.EnsureRepo(b.ctx, backup.Namespace, pod.Namespace, backup.Spec.StorageLocation)
+	repositoryType := getRepositoryType(b.uploaderType)
+	if repositoryType == "" {
+		err := errors.New("invalid repository type")
```
reasonerjt, line 115: In this case there's probably a wrong uploadertype, so would it be helpful to put the uploader type as part of the error?

pkg/podvolume/restorer.go
```diff
@@ -213,3 +220,27 @@ func newPodVolumeRestore(restore *velerov1api.Restore, pod *corev1api.Pod, backu
...
+		return "", errors.New("empty volume list")
+	}
+
+	// the podVolumeBackups list come from one backup. In one backup, it is impossible that volumes are
+	// backed up by different uploaders or to different repositories. Asserting this ensures one repo only,
+	// which will simplify the following logics
+	repositoryType := ""
+	for _, backupInfo := range volumes {
+		if backupInfo.RepositoryType == "" {
+			return "", errors.New("invalid repository type among volumes")
```
reasonerjt, line 235: This is actually empty, why not make the error msg like `empty repository type`, also this is an error for one particular volume, therefore would it be helpful to put more info in the error to help identify the volume?

pkg/podvolume/restorer.go
```diff
@@ -213,3 +220,27 @@ func newPodVolumeRestore(restore *velerov1api.Restore, pod *corev1api.Pod, backu
...
+		}
+
+		if repositoryType == "" {
+			repositoryType = backupInfo.RepositoryType
+		} else if repositoryType != backupInfo.RepositoryType {
+			return "", errors.New("multiple repository type in one backup")
```
reasonerjt, line 241: Similarly, it would be helpful to put more info in the error message?

pkg/podvolume/util.go
```diff
@@ -48,10 +50,33 @@ const (
 	InitContainer = "restic-wait"
 )
 
+// VolumeBackupInfo describes the backup info of a volume backed up by PodVolumeBackups
```
reasonerjt, line 53: If this is not intended to be used by other packages we don't want to expose it.

pkg/podvolume/util.go
```diff
@@ -71,14 +96,73 @@ func GetVolumeBackupsForPod(podVolumeBackups []*velerov1api.PodVolumeBackup, pod
...
+			RepositoryType: getRepositoryType(pvb.Spec.UploaderType),
+		}
 	}
 
 	if len(volumes) > 0 {
 		return volumes
 	}
 
-	return getPodSnapshotAnnotations(pod)
+	fromAnnntation := getPodSnapshotAnnotations(pod)
```
reasonerjt, line 110: I see the comment of `getPodSnapshotAnnotations` says
```// Deprecated: we will stop using pod annotations to record restic snapshot IDs after they're taken,
// therefore we won't need to check if these annotations exist.

```
Could you double check if we still need this in v1.10?

pkg/podvolume/util.go
```diff
@@ -71,14 +96,73 @@ func GetVolumeBackupsForPod(podVolumeBackups []*velerov1api.PodVolumeBackup, pod
...
+	}
+
+	for k, v := range fromAnnntation {
+		volumes[k] = VolumeBackupInfo{v, uploader.ResticType, velerov1api.BackupRepositoryTypeRestic}
+	}
+
+	return volumes
+}
+
+func GetSnapshotIdentifier(podVolumeBackups *velerov1api.PodVolumeBackupList) []repository.SnapshotIdentifier {
```
reasonerjt, line 122: Please add comments as it's an exported func

pkg/podvolume/util.go
reasonerjt: https://github.com/vmware-tanzu/velero/pull/1577

The deprecation message was added back in v1.0.1

Let's confirm with community, but I hope we don't need to keep this in v1.10.0

pkg/podvolume/backupper.go
Lyndon-Li: Done, added the uploader type

pkg/podvolume/restorer.go
Lyndon-Li: Done, changed the message to `"empty repository type found among volume snapshots, snapshot ID %s, uploader %s"`

pkg/podvolume/restorer.go
Lyndon-Li: Done changed the message to `"multiple repository type in one backup, current type %s, differential one [type %s, snapshot ID %s, uploader %s]"`

pkg/podvolume/util.go
Lyndon-Li: Looks like this is for back compatibility of very old version, let's double check it and also discuss it in the community meeting, if we finally confirm it is not used any more, let's remove it.
As the discussion, we will not change it in the current PR, and will open an issue to trace it.

Lyndon-Li: Done, thanks.

pkg/podvolume/util.go
Lyndon-Li: Done, thanks.

pkg/podvolume/util.go
Lyndon-Li: OK, an issue https://github.com/vmware-tanzu/velero/issues/5277 is opened for this purpose, let's double check it and discuss it in the community.

reasonerjt (approved):

ywk253100 (approved):

PR #5282, opened 2022-09-05
Kopia Pod Volume Backup/Restore Smoking Test Changes

Lyndon-Li:
Add changes for problems/enhancements found during smoking test for Kopia pod volume backup/restore


Lyndon-Li:
Fixed issues:
1. PVB/PVR controller gets BackupRepository and ensure it is ready
2. `ensureRepo` function in `BackupRepositoryController` checks the error message "`Is there a repository at the following location?`", this is for Restic only, we need to support Kopia path
3. We have a logic in `BackupStorageLocationController ` that checks the folder in the remote object store. For Kopia path, we've added a new folder for Kopia repository, we need to enhance the check logic, otherwise, it reports errors
4. Add support to write Kopia log to Velero log for Kopia uploader
5. The "`logSource`" field in kopia_log is duplicated with the existing Velero log, so changed it to "`logModule`"

pkg/repository/backup_repo_op.go
```diff
@@ -0,0 +1,80 @@
...
+		Namespace:     namespace,
+		LabelSelector: selector,
+	})
+
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting backup repository list")
+	}
+
+	if len(backupRepoList.Items) == 0 {
+		return nil, errors.Errorf(notFoundError+" not found BackupRepository for workload namespace %q, backup storage location %q, repository type %q", key.VolumeNamespace, key.BackupLocation, key.RepositoryType)
```
ywk253100, line 64: Seems dup error msg. And why do we need to define `notFoundError`? If it's used to determine the specific error by the caller, a new `error` variable should be defined rather than a string error message.

pkg/repository/udmrepo/kopialib/lib_repo.go
```diff
@@ -384,7 +386,11 @@ func (kor *kopiaObjectReader) Read(p []byte) (int, error) {
 
 	n, err := kor.rawReader.Read(p)
 	if err != nil {
-		return 0, errors.Wrap(err, "error to read object")
+		if errors.Is(err, io.EOF) {
```
ywk253100, line 389: Why do we check whether it's io.EOF or not? Could you add some comment?

pkg/repository/backup_repo_op.go
```diff
@@ -0,0 +1,80 @@
...
+func repoLabelsFromKey(key BackupRepositoryKey) labels.Set {
+	return map[string]string{
+		velerov1api.VolumeNamespaceLabel: label.GetValidName(key.VolumeNamespace),
+		velerov1api.StorageLocationLabel: label.GetValidName(key.BackupLocation),
+		velerov1api.RepositoryTypeLabel:  label.GetValidName(key.RepositoryType),
+	}
+}
+
+// GetBackupRepository gets a backup repository through BackupRepositoryKey and ensure ready if required.
+func GetBackupRepository(ctx context.Context, cli client.Client, namespace string, key BackupRepositoryKey, ensureReady bool) (*velerov1api.BackupRepository, error) {
```
reasonerjt, line 50: If 99% of the time this func will be called with ensureReady==true, would it be better to replace it with `skipReadyCheck` and make it an optional parm?  Such that the caller will not need to remember setting the last parm.

pkg/repository/backup_repo_op.go
Lyndon-Li: Yes, repo ensurer requires to check this specific error to determine to create a new BackupRepository.

I will add an error variable

pkg/repository/provider/restic.go
```diff
@@ -46,10 +47,19 @@ func (r *resticRepositoryProvider) ConnectToRepo(ctx context.Context, param Repo
...
 
 func (r *resticRepositoryProvider) PrepareRepo(ctx context.Context, param RepoParam) error {
-	if err := r.InitRepo(ctx, param); err != nil {
+	if err := r.ConnectToRepo(ctx, param); err != nil {
+		// If the repository has not yet been initialized, the error message will always include
+		// the following string. This is the only scenario where we should try to initialize it.
+		// Other errors (e.g. "already locked") should be returned as-is since the repository
+		// does already exist, but it can't be connected to.
+		if strings.Contains(err.Error(), "Is there a repository at the following location?") {
+			return r.InitRepo(ctx, param)
```
reasonerjt, line 56: You don't need to call `ConnectToRepo` in this case?

reasonerjt: Confirmed that when the `InitRepo` is called the Connect is no necessaary.

pkg/repository/udmrepo/kopialib/lib_repo.go
reasonerjt, line 389: Is it necessary to wrap other errors?

pkg/repository/provider/restic.go
Lyndon-Li: Not any more. The syntax of the caller's requirement (BackupRepository Controller's `ensureRepo` function) is: connect the repo first, if it is not initialized, initialize it, or simply speaking, make sure the repo is usable, and the newly added method PrepareRepo is for this purpose.

pkg/repository/udmrepo/kopialib/lib_repo.go
Lyndon-Li: True, I will remove the wrap for all the method of io.Reader, io.Writer, io.Closer and io.Seeker, these are the system wide standard errors, the caller may check these standard errors and wrapping it indeed brings us awkward

Lyndon-Li: This is the implementation of io.Reader, it returns some system wide standard errors, the caller may catch these standard errors and not treat it as an error case, for example, io.EOF.
 I will remove the error wrap for all the method of io.Reader, io.Writer, io.Closer and io.Seeker.

pkg/repository/backup_repo_op.go
Lyndon-Li: Done

pkg/repository/udmrepo/kopialib/lib_repo.go
Lyndon-Li: Done

Lyndon-Li: Done

pkg/repository/backup_repo_op.go
Lyndon-Li: Done, changed the `ensureReady` to an optional param.

ywk253100 (approved):

reasonerjt (approved):
lgtm

Issue #5288, opened 2022-09-06
Velero restore partially failed as default priorityClass


MallyZhang:
**What steps did you take and what happened:**
Velero restore partially failed as only one priorityClass can be default
```
time="2022-09-06T01:48:25Z" level=info msg="restore status includes excludes: <nil>" logSource="pkg/restore/restore.go:1134" restore=velero/velero-backup-obj-20220906014657-20220906014748
time="2022-09-06T01:48:25Z" level=info msg="Attempting to restore PriorityClass: hc-default" logSource="pkg/restore/restore.go:1245" restore=velero/velero-backup-obj-20220906014657-20220906014748
time="2022-09-06T01:48:25Z" level=error msg="error restoring hc-default: priorityclasses.scheduling.k8s.io \"hc-default\" is forbidden: PriorityClass hc-default is already marked as default. Only one default can exist" logSource="pkg/restore/restore.go:1357" restore=velero/velero-backup-obj-20220906014657-20220906014748
```

```
[root@localhost ]# kubectl create -f hc-default.yaml 
+ kubectl create -f hc-default.yaml
Error from server (Forbidden): error when creating "hc-default.yaml": priorityclasses.scheduling.k8s.io "hc-default" is forbidden: PriorityClass hc-default is already marked as default. Only one default can exist
[root@localhost ]# kubectl create -f hc-db.yaml 
+ kubectl create -f hc-db.yaml
Error from server (AlreadyExists): error when creating "hc-db.yaml": priorityclasses.scheduling.k8s.io "hc-db" already exists
```

**What did you expect to happen:**

don't report this error when there is one default priorityClass

**The following information will help us better understand what's going on**:

_If you are using velero v1.7.0+:_  
Please use `velero debug  --backup <backupname> --restore <restorename>` to generate the support bundle, and attach to this issue, more options please refer to `velero debug --help` 

_If you are using earlier versions:_  
Please provide the output of the following commands (Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)
- `kubectl logs deployment/velero -n velero`
- `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
- `velero backup logs <backupname>`
- `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
- `velero restore logs <restorename>`

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

[...]

PR #5292, opened 2022-09-06
skip forbidden error for restore priorityClass

cleverhu:

skip forbidden error for restore priorityClass

Fixes https://github.com/vmware-tanzu/velero/issues/5288

### DIFF
--- a/changelogs/unreleased/5292-cleverhu
+++ b/changelogs/unreleased/5292-cleverhu
@@ -1,0 +1,1 @@
+Skip forbidden error for restore priorityClass.
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -1448,30 +1448,55 @@ func isAlreadyExistsError(ctx *restoreContext, obj *unstructured.Unstructured, e
 	// 2. For LoadBalancer service, the "healthCheckNodePort" already exists. - Get internal error
 	// If this is the case, the function returns true to avoid reporting error.
 	// Refer to https://github.com/vmware-tanzu/velero/issues/2308 for more details
-	if obj.GetKind() != "Service" {
+	// The "forbidden error" rather than "already exists" error returns when restoring priorityClass in the following case:
+	// 1. For globalDefault priorityClass, the default priorityClass already exists. - Get forbidden error
+	// Refer to https://github.com/vmware-tanzu/velero/issues/5288 for more details
+	if obj.GetKind() != "Service" && obj.GetKind() != "PriorityClass" {
 		return false, nil
 	}
 	statusErr, ok := err.(*apierrors.StatusError)
 	if !ok || statusErr.Status().Details == nil || len(statusErr.Status().Details.Causes) == 0 {
 		return false, nil
 	}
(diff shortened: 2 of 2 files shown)


sseago:
First, to clarify the error you got -- the priorityClass you're trying to restore that fails -- is there already a priorityClass in the cluster with the same name? If so, then this is a case where Velero should normally just skip over it, as the resource is already in the cluster (warning, but not erroring out, if the in-cluster version is not identical to the backup verson). If that's true, then we already have a fix merged from https://github.com/vmware-tanzu/velero/pull/5239 -- which will be in 1.10 and is also targeted for 1.9.2.

The issue is that Velero uses the error return (looking for already exists error) to determine whether the reason for failure is that the resource is already in the cluster, and any other error type would normally indicate a different failure that should fail a restore. The problem is that in some cases, there is other ingress validation which is failing before checking whether the resource already exists, so Velero is working with incomplete information when determining whether to error out or not. The PR linked above fixes this for all resource types, not just priority classes.

If the above description of the problem is correct, we don't need this PR as it's already been fixed. Also, this fix introduces a different bug. If you're trying to restore a completely different priority class (with a different name), then erroring out *is* the correct thing to do, since the user is trying to create a second default priority class, which is not allowed by cluster rules. Also, this PR would discard *any* error on priority class restore and not just "too many default classes" errors.

cleverhu:
> First, to clarify the error you got -- the priorityClass you're trying to restore that fails -- is there already a priorityClass in the cluster with the same name? If so, then this is a case where Velero should normally just skip over it, as the resource is already in the cluster (warning, but not erroring out, if the in-cluster version is not identical to the backup verson). If that's true, then we already have a fix merged from #5239 -- which will be in 1.10 and is also targeted for 1.9.2.
> 
> The issue is that Velero uses the error return (looking for already exists error) to determine whether the reason for failure is that the resource is already in the cluster, and any other error type would normally indicate a different failure that should fail a restore. The problem is that in some cases, there is other ingress validation which is failing before checking whether the resource already exists, so Velero is working with incomplete information when determining whether to error out or not. The PR linked above fixes this for all resource types, not just priority classes.
> 
> If the above description of the problem is correct, we don't need this PR as it's already been fixed. Also, this fix introduces a different bug. If you're trying to restore a completely different priority class (with a different name), then erroring out _is_ the correct thing to do, since the user is trying to create a second default priority class, which is not allowed by cluster rules. Also, this PR would discard _any_ error on priority class restore and not just "too many default classes" errors.

Thanks for your advice, I also think that it is not reasonable to add many priority errors, which is inconvenient for later program changes. However, this will not introduce new bugs. The 'restore a completely different priority class (with a different name)' mentioned above will lead to data loss. The program verifies whether the resource exists at the end. I checked the PR integrated above, and I found that the previous priority detection function has completely lost its function. I will propose a PR for everyone to discuss whether it is equivalent to the previous program.

cleverhu:
实现不友好，因此要关闭。它已经移交到另外一个pr：https://github.com/vmware-tanzu/velero/pull/5293

PR #5293, opened 2022-09-06
optimize code for restore exists resources

cleverhu:

Optimize the previous logic and handle errors uniformly. If the resource exists but cannot be found, the 'get' error will be returned; otherwise, the creation error will be returned.

Fixes https://github.com/vmware-tanzu/velero/issues/5288

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -1252,29 +1252,22 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 
 	// check if we want to treat the error as a warning, in some cases the creation call might not get executed due to object API validations
 	// and Velero might not get the already exists error type but in reality the object already exists
-	objectExists := false
 	var fromCluster *unstructured.Unstructured
 
-	if restoreErr != nil && !isAlreadyExistsError {
+	if restoreErr != nil {
 		// check for the existence of the object in cluster, if no error then it implies that object exists
-		// and if err then we want to fallthrough and do another get call later
+		// and if err then we want to judge whether there is an existing error in the previous creation.
+		// if so, we will return the 'get' error.
+		// otherwise, we will return the original creation error.
 		fromCluster, err = resourceClient.Get(name, metav1.GetOptions{})
-		if err == nil {
-			objectExists = true
+		if err != nil && isAlreadyExistsError {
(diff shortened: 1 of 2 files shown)


cleverhu:
@sseago I think the logic modified now is equivalent to that before. Can you help me see what is wrong with my understanding.

sseago:
The biggest difference with this version is that in the "object doesn't exist" case, instead of adding restoreErr as a restore error, we add err -- so instead of the error returned by the create call, we're recording the "not found" error returned by the get call.

cleverhu:
> The biggest difference with this version is that in the "object doesn't exist" case, instead of adding restoreErr as a restore error, we add err -- so instead of the error returned by the create call, we're recording the "not found" error returned by the get call.

I submitted a new version. What do you think of this version? I always feel a little strange about the previous writing.

pkg/restore/restore.go
```diff
@@ -1252,29 +1252,22 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
...
 		// check for the existence of the object in cluster, if no error then it implies that object exists
-		// and if err then we want to fallthrough and do another get call later
+		// and if err then we want to judge whether there is an existing error in the previous creation.
+		// if so, we will return the 'get' error.
+		// otherwise, we will return the original creation error.
 		fromCluster, err = resourceClient.Get(name, metav1.GetOptions{})
-		if err == nil {
-			objectExists = true
+		if err != nil && isAlreadyExistsError {
+			ctx.log.Errorf("Error retrieving cluster version of %s: %v", kube.NamespaceAndName(obj), err)
```
shubham-pampattiwar, line 1264:
```suggestion
			ctx.log.Errorf("Error retrieving in-cluster version of %s: %v", kube.NamespaceAndName(obj), err)
```

cleverhu: I will modify and output more detailed information, thx.

cleverhu:
@dsu-igeek @blackpiglet I'm sorry,I clicked the wrong button. I want @ shubham pampattiwar to help me review again instead of cancelling your review request. /PTAL

sseago (approved):

shubham-pampattiwar (approved):

PR #5308, opened 2022-09-07
Repo Ensurer Refactor

Lyndon-Li:
Refactor the repoEnsurer code to use controller runtime client and wrap some common BackupRepository operations to share with other modules.

### DIFF
--- a/pkg/repository/ensurer.go
+++ b/pkg/repository/ensurer.go
@@ -18,92 +18,34 @@ package repository
 
 import (
 	"context"
-	"fmt"
 	"sync"
 	"time"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/client-go/tools/cache"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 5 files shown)


blackpiglet (dismissed):

pkg/repository/ensurer.go
```diff
@@ -18,51 +18,39 @@ package repository
...
 	"github.com/sirupsen/logrus"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/tools/cache"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
-	velerov1client "github.com/vmware-tanzu/velero/pkg/generated/clientset/versioned/typed/velero/v1"
 	velerov1informers "github.com/vmware-tanzu/velero/pkg/generated/informers/externalversions/velero/v1"
```
ywk253100, line 31: We should remove the dependency for all code under `pkg/generated` after refactoring. 

The `informer` here is used to check the repo status, IMHO we can change the logic by getting the repo periodically in a loop. The repo is read from the cache, so there should be no efficiency issue

Lyndon-Li: Done. Change the way to wait for the created BackupRepository -- polling until it is created and ready or timeout.

ywk253100 (approved):

blackpiglet (approved):

Issue #5307, opened 2022-09-07
Feature to create a schedule for maintenance


navilg:
**Describe the problem/challenge you have**
[A description of the current limitation/problem/challenge that you are experiencing.]

I have multiple backups running on my cluster which are few hours apart. When maintenance runs, Many time (mostly) backup in progress and maintenance clashes (I believe seeing the error message) and then we get error restic prune failed since repo is locked by another process.

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

It would have been useful if we could schedule the maintenance same as we schedule backup. In this way, users will have flexibility to choose maintenance time and frequency as per their need. Schedule can be anything, either installation time or post-installation feature.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Velero version (use `velero version`): 1.6.2
- Kubernetes version (use `kubectl version`): 1.17.9
- Kubernetes installer & version:
- Cloud provider or hardware configuration: GKE
- OS (e.g. from `/etc/os-release`): 

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

Lyndon-Li:
Do you have multiple BackupStorageLocation that all points to the same bucket in the same object store?

navilg:
> Do you have multiple BackupStorageLocation that all points to the same bucket in the same object store?

No. I have only one backupstoragelocation (default), pointing to a S3 bucket.

navilg:
@Lyndon-Li Can velero backup and restic maintenance run concurrently ? Or is there any check in place to check if any backup is progress before restic prune runs automatically everyweek ?

cleverhu:
Before each backup, judge whether there is a backup in progress. If there is an ongoing task, try again later. Another way is to let users set how many goroutinues can be set to back up at the same time. What do you think? @Lyndon-Li

cleverhu:
I pulled a request for this, but I'm not sure if it's appropriate? Is there a more elegant way to achieve it.

navilg:
[...]

PR #5320, opened 2022-09-09
support only one backup is in progress

cleverhu:

support only one backup is in progress.

Fixes https://github.com/vmware-tanzu/velero/issues/5307

### DIFF
--- a/changelogs/unreleased/5320-cleverhu
+++ b/changelogs/unreleased/5320-cleverhu
@@ -1,0 +1,1 @@
+Support only one backup is in progress.
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -74,6 +74,10 @@ import (
 	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	defaultRetryInterval = time.Second * 10
+)
+
 type backupController struct {
 	*genericController
 	discoveryHelper             discovery.Helper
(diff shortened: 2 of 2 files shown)


sseago:
Where have you seen multiple in-progress backups? What version of Velero was this on? Currently Velero only processes one backup at a time, so it's impossible for Velero to be running more than one. In the past we had a bug where if the velero pod crashed while a backup was in progress, the backup phase would be left "InProgress" forever, but it would never actually be processed again. That's been fixed in Velero 1.9, though. If the velero backup controller finds another backup InProgress (this shouldn't happen in 1.9, but maybe there are edge cases where it does?), I don't think waiting will ever make that state change, since it's no longer being processed.

cleverhu:
> Where have you seen multiple in-progress backups? What version of Velero was this on? Currently Velero only processes one backup at a time, so it's impossible for Velero to be running more than one. In the past we had a bug where if the velero pod crashed while a backup was in progress, the backup phase would be left "InProgress" forever, but it would never actually be processed again. That's been fixed in Velero 1.9, though. If the velero backup controller finds another backup InProgress (this shouldn't happen in 1.9, but maybe there are edge cases where it does?), I don't think waiting will ever make that state change, since it's no longer being processed.

I also saw the issue. After the controller restarted, the status of backup will be set to failure, and then tell the user that it has been rebooted. I do not know the logic, the user feedback restric backup failed, I don't know how to handle it, I will check the code later.Thx.

PR #5333, opened 2022-09-13
Proposal to separate resource filter into cluster scope and namespace…

### LINKED ISSUE #5120: Support cluster resource Includes/Excludes

The current filter (IncludedResources/ExcludedResources + IncludeClusterResources flag) is not enough for some special cases, such as:
1. all namespaced resources + some kind of cluster resource
2. all namespaced resources + cluster resource excludes
3. no namespaced resources + all cluster resources 

Add new field `IncludedClusterResources` and `ExcludedClusterResources`, examples:
(wildcard rules: includeResources: empty for all, excludeResources/includeClusterResources: "*" for all, excludeResources/includeClusterResources/excludeClusterResources: empty for nothing)

1. no namespaced resources + no cluster resources:
```
excludeResources: ["*"]
```

2. no namespaced resources + some cluster resources:
```
excludeResources: ["*"]
includeClusterResources: ["persistentvolumes"]
```
```
excludeResources: ["*"]
excludeClusterResources: ["persistentvolumes"]
```

3. no namespaced resources + all cluster resources
```
excludeResources: ["*"]
includeClusterResources: ["*"]
```

4. some namespaced resources + no cluster resources:
```
includeResources: ["configmaps"]
```
```
excludeResources: ["configmaps"]
```

5. some namespaced resources + some cluster resources:
```
includeResources: ["configmaps"]
includeClusterResources: ["persistentvolumes"]
```
```
excludeResources: ["configmaps"]
includeClusterResources: ["persistentvolumes"]
```
```
includeResources: ["configmaps"]
excludeClusterResources: ["persistentvolumes"]
```
```
excludeResources: ["configmaps"]
excludeClusterResources: ["persistentvolumes"]
```

6. some namespaced resources + all cluster resources:
```
includeResources: ["configmaps"]
includeClusterResources: ["*"]
```
```
excludeResources: ["configmaps"]
includeClusterResources: ["*"]
```
7. all namespaced resources + no cluster resources
```
includeResources:
```
8. all namespaced resources + some cluster resources:
```
includeClusterResources: ["persistentvolumes"]
```
```
excludeClusterResources: ["persistentvolumes"]
```

9. all namespaced resources + all cluster resources:
```
includeClusterResources: ["*"]
```

**Environment:**

**Vote on this issue!**

blackpiglet:
… scope.

Signed-off-by: Xun Jiang <blackpiglet@gmail.com>

Fixes #5120 

### DIFF
--- a/design/cluster-scope-resource-filter.md
+++ b/design/cluster-scope-resource-filter.md
@@ -1,0 +1,410 @@
+# Proposal to add resource filters for backup can distinguish whether resource is cluster scope or namespace scope.
+
+- [Proposal to add resource filters for backup can distinguish whether resource is cluster scope or namespace scope.](#proposal-to-add-resource-filters-for-backup-can-distinguish-whether-resource-is-cluster-scope-or-namespace-scope)
+	- [Abstract](#abstract)
+	- [Background](#background)
+	- [Goals](#goals)
+	- [Non Goals](#non-goals)
+	- [High-Level Design](#high-level-design)
+		- [Parameters Rules](#parameters-rules)
+		- [Using scenarios:](#using-scenarios)
+			- [no namespaced resources + no cluster resources](#no-namespaced-resources--no-cluster-resources)
+			- [no namespaced resources + some cluster resources](#no-namespaced-resources--some-cluster-resources)
+			- [no namespaced resources + all cluster resources](#no-namespaced-resources--all-cluster-resources)
+			- [some namespaced resources + no cluster resources](#some-namespaced-resources--no-cluster-resources)
+			- [some namespaced resources + only related cluster resources](#some-namespaced-resources--only-related-cluster-resources)
+			- [some namespaced resources + some additional cluster resources](#some-namespaced-resources--some-additional-cluster-resources)
+			- [some namespaced resources + all cluster resources](#some-namespaced-resources--all-cluster-resources)
+			- [all namespaced resources + no cluster resources](#all-namespaced-resources--no-cluster-resources)
+			- [all namespaced resources + some additional cluster resources](#all-namespaced-resources--some-additional-cluster-resources)
+			- [all namespaced resources + all cluster resources](#all-namespaced-resources--all-cluster-resources)
+			- [describe command change](#describe-command-change)
+	- [Detailed Design](#detailed-design)
+	- [Alternatives Considered](#alternatives-considered)
+	- [Security Considerations](#security-considerations)
+	- [Compatibility](#compatibility)
+	- [Implementation](#implementation)
+	- [Open Issues](#open-issues)
+
+## Abstract
+The current filter (IncludedResources/ExcludedResources + IncludeClusterResources flag) is not enough for some special cases, e.g. all namespaced resources + some kind of cluster resource and all namespaced resources + cluster resource excludes.
+Propose to add a new group of resource filtering parameters, which can distinguish cluster and namespaced resources.
+
+## Background
+There are two sets of resource filters for Velero: `IncludedNamespaces/ExcludedNamespaces` and `IncludedResources/ExcludedResources`. 
+`IncludedResources` means only including the resource types specified in the parameter. Both cluster and namespaced resources are handled in this parameter by now.
+The k8s resources are separated into cluster scope and namespaced scope.
+As a result, it's hard to include all resources in one group and only including specified resource in the other group.
+
+## Goals
+- Make Velero can support more complicated namespaced and cluster resources filtering scenarios in backup.
+
+## Non Goals
+- Enrich the resource filtering rules, for example, advanced PV filtering and filtering by resource names.
+
+
+## High-Level Design
+Four new parameters are added into command `velero backup create`: `--include-cluster-scope-resources`, `--exclude-cluster-scope-resources`, `--include-namespaced-resources` and `--exclude-namespaced-resources`. 
+`--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` are used to filter cluster scope resources included or excluded in backup per resource type.
+`--include-namespaced-resources` and `--exclude-namespaced-resources` are used to filter namespace scope resources included or excluded in backup per resource type.
+Restore and other code pieces also use resource filtering will be handled in future releases.
+
+### Parameters Rules
+
+* `--include-cluster-scope-resources`, `--include-namespaced-resources`, `--exclude-cluster-scope-resources` and `--exclude-namespaced-resources` valid value include `*` and comma separated string. Each element of the CSV string should a k8s resource name. The format should be `resource.group`, such as `storageclasses.storage.k8s.io.`.
+
+* `--include-cluster-scope-resources`, `--include-namespaced-resources`, `--exclude-cluster-scope-resources` and `--exclude-namespaced-resources` parameters are mutual exclusive with `--include-cluster-resources`, `--include-resources` and `--exclude-resources` parameters. If both sets of parameters are provisioned, validation failure should be returned.
+
+* `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` should only contain cluster scope resource type names. If namespace scope resource type names are included, they are ignored.
+
+* If there are conflicts between `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` specified resources type lists, `--exclude-cluster-scope-resources` parameter has higher priority.
+
+* `--include-namespaced-resources` and `--exclude-namespaced-resources` should only contain namespace scope resource type names. If cluster scope resource type names are included, they are ignored.
+
+* If there are conflicts between `--include-namespaced-resources` and `--exclude-namespaced-resources` specified resources type lists, `--exclude-namespaced-resources` parameter has higher priority.
+
+* If  `--include-namespaced-resources` is not present, it means all namespace scope resources are included per resource type.
+
+* If both `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` are not present, it means no additional cluster resource is included per resource type, just as the existing `--include-cluster-resources` parameter not setting value. Cluster resources are related to the namespace scope resources, which means those are returned in the namespace resources' BackupItemAction's result AdditionalItems array, are still included in backup by default. Taking backing up PVC scenario as an example, PVC is namespaced, PV is in cluster scope. PVC's BIA will include PVC related PV into backup too.
+
+* If the backup contains no resource, validation failure should be returned.
+
+### Using scenarios:
+Please notice, if the scenario give the example of using old filtering parameters (`--include-cluster-resources`, `--include-resources` and `--exclude-resources`), that means the old parameters also work for this case. If old parameters example is not given, that means they don't work for this scenario, only new parameters (`--include-cluster-scope-resources`, `--include-namespaced-resources`, `--exclude-cluster-scope-resources` and `--exclude-namespaced-resources`) work.
+
+#### no namespaced resources + no cluster resources
+This is not allowed. Backup or restore cannot contain no resource.
+
+#### no namespaced resources + some cluster resources
+The following command means backup no namespaced resources and some cluster resources.
+
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaced-resources=*
+	--include-cluster-scope-resources=storageclass
+```
+
+#### no namespaced resources + all cluster resources
+The following command means backup no namespaced resources and all cluster resources.
+
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaced-resources=*
+	--include-cluster-scope-resources=*
+```
+
+#### some namespaced resources + no cluster resources
+The following commands mean backup all resources in namespaces default and kube-system, and no cluster resources.
+
+Example of new parameters:
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--exclude-cluster-scope-resources=*
+```
+
+Example of old parameters:
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-cluster-resources=false
+```
+
+The following commands mean backup PVC, Deployment, Service, Endpoint, Pod and ReplicaSet resources in all namespaces, and no cluster resources. Although PVC's related PV should be included, due to no cluster resources are included, so they are ruled out too.
+
+Example of new parameters:
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+	--exclude-cluster-scope-resources=*
+```
+
+Example of old parameters:
+``` bash
+velero backup create <backup-name>
+	--include-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+	--include-cluster-resources=false
+```
+
+The following commands mean backup PVC, Deployment, Service, Endpoint, Pod and ReplicaSet resources in namespace default and kube-system, and no cluster resources. Although PVC's related PV should be included, due to no cluster resources are included, so they are ruled out too.
+
+Example of new parameters:
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-namespaced-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+	--exclude-cluster-scope-resources=*
+```
+
+Example of old parameters:
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+	--include-cluster-resources=false
+```
+
+The following commands mean backup all resources except Ingress type resources in all namespaces, and no cluster resources.
+
+Example of new parameters:
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaced-resources=ingress
+	--exclude-cluster-scope-resources=*
+```
+
+Example of old parameters:
+``` bash
+velero backup create <backup-name>
+	--exclude-resources=ingress
+	--include-cluster-resources=false
+```
+
+#### some namespaced resources + only related cluster resources
+This means backup all resources in namespaces default and kube-system, and related cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+```
+
+The following commands mean backup PVC, Deployment, Service, Endpoint, Pod and ReplicaSet resources in all namespaces, and related cluster resources (PVC's related PV).
+
+Example of new parameters:
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+```
+
+Example of old parameters:
+``` bash
+velero backup create <backup-name>
+	--include-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+```
+
+The following commands mean backup PVC, Deployment, Service, Endpoint, Pod and ReplicaSet resources in namespaces default and kube-system, and related cluster resources. PVC related PV is included too.
+
+Example of new parameters:
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-namespaced-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+```
+
+Example of old parameters:
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+```
+
+This means backup all resources except Ingress type resources in all namespaces, and related cluster resources.
+
+Example of new parameters:
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaced-resources=ingress
+```
+
+Example of old parameters:
+``` bash
+velero backup create <backup-name>
+	--exclude-resources=ingress
+```
+
+#### some namespaced resources + some additional cluster resources
+This means backup all resources in namespace in default, kube-system, and related cluster resources, plus all StorageClass cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-cluster-scope-resources=storageclass
+```
+
+This means backup PVC, Deployment, Service, Endpoint, Pod and ReplicaSet resources in all namespaces, and related cluster resources, plus all StorageClass cluster resources, and PVC related PV.
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+	--include-cluster-scope-resources=storageclass
+```
+
+This means backup PVC, Deployment, Service, Endpoint, Pod and ReplicaSet resources in default and kube-system namespaces, and related cluster resources, plus all StorageClass cluster resources, and PVC related PV.
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+	--include-namespaces=default,kube-system
+	--include-cluster-scope-resources=storageclass
+```
+
+This means backup PVC, Deployment, Service, Endpoint, Pod and ReplicaSet resources in default and kube-system namespaces, and related cluster resources, plus all cluster scope resources except StorageClass type resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+	--include-namespaces=default,kube-system
+	--exclude-cluster-scope-resources=storageclass
+```
+
+#### some namespaced resources + all cluster resources
+The following commands mean backup all resources in namespace in default, kube-system, and all cluster resources.
+
+Example of new parameters:
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-cluster-scope-resources=*
+```
+
+Example of old parameters:
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-cluster-resources=true
+```
+
+This means backup Deployment, Service, Endpoint, Pod and ReplicaSet resources in all namespaces, and all cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=deployment,service,endpoint,pod,replicaset
+	--include-cluster-scope-resources=*
+```
+
+This means backup Deployment, Service, Endpoint, Pod and ReplicaSet resources in default and kube-system namespaces, and all cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-namespaced-resources=deployment,service,endpoint,pod,replicaset
+	--include-cluster-scope-resources=*
+```
+
+#### all namespaced resources + no cluster resources
+The following commands all mean backup all namespace scope resources and no cluster resources.
+
+Example of new parameters:
+``` bash
+velero backup create <backup-name>
+	--exclude-cluster-scope-resources=*
+```
+
+Example of old parameters:
+``` bash
+velero backup create <backup-name>
+	--include-cluster-resources=false
+```
+
+#### all namespaced resources + some additional cluster resources
+This command means backup all namespace scope resources, and related cluster resources, plus all PersistentVolume resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=*
+	--include-cluster-scope-resources=persistentvolume
+```
+
+#### all namespaced resources + all cluster resources
+The following commands have the same meaning: backup all namespace scope resources, and all cluster scope resources.
+``` bash
+velero backup create <backup-name>
+	--include-cluster-scope-resources=*
+```
+
+``` bash
+velero backup create <backup-name>
+	--include-cluster-resources=true
+```
+
+#### describe command change
+In `velero backup describe` command, the four new parameters should be outputted too.
+``` bash
+ velero backup describe <backup-name>
+......
+
+Namespaces:
+  Included:  ns2
+  Excluded:  <none>
+
+Resources:
+  Included:               <none>
+  Excluded:               <none>
+  Included-cluster-scope: StorageClass,PersistentVolume
+  Excluded-cluster-scope: <none>
+  Included-namespaced:    default
+  Excluded-namespaced:    <none>
+  Cluster-scoped:  auto
+
+......
+```
+
+**Note:** `velero restore` command doesn't support those four new parameter in Velero v1.11, but `velero schedule` supports the four new parameters through backup specification.
+
+## Detailed Design
+With adding `IncludedNamespacedResources`, `ExcludedNamespacedResources`, `IncludedClusterScopeResources` and `ExcludedClusterScopeResources`, the `BackupSpec` looks like:
+``` go
+type BackupSpec struct {
+    ......
+	// IncludedResources is a slice of resource names to include
+	// in the backup. If empty, all resources are included.
+	// +optional
+	// +nullable
+	IncludedResources []string `json:"includedResources,omitempty"`
+
+	// ExcludedResources is a slice of resource names that are not
+	// included in the backup.
+	// +optional
+	// +nullable
+	ExcludedResources []string `json:"excludedResources,omitempty"`
+
+	// IncludeClusterResources specifies whether cluster-scoped resources
+	// should be included for consideration in the backup.
+	// +optional
+	// +nullable
+	IncludeClusterResources *bool `json:"includeClusterResources,omitempty"`
+
+	// IncludedClusterScopeResources is a slice of cluster scope 
+	// resource type names to include in the backup.
+	// If set to "*", all cluster scope resource types are included.
+	// The default value is empty, which means only related cluster 
+	// scope resources are included.
+	// +optional
+	// +nullable
+	IncludedClusterScopeResources []string `json:"includedClusterScopeResources,omitempty"`
+
+	// ExcludedClusterScopeResources is a slice of cluster scope 
+	// resource type names to exclude from the backup.
+	// If set to "*", all cluster scope resource types are excluded.
+	// +optional
+	// +nullable
+	ExcludedClusterScopeResources []string `json:"excludedClusterScopeResources,omitempty"`
+
+	// IncludedNamespacedResources is a slice of namespace scope
+	// resource type names to include in the backup.
+	// The default value is "*".
+	// +optional
+	// +nullable
+	IncludedNamespacedResources []string `json:"includedNamespacedResources,omitempty"`
+
+	// ExcludedNamespacedResources is a slice of namespace scope
+	// resource type names to exclude from the backup.
+	// If set to "*", all namespace scope resource types are excluded.
+	// +optional
+	// +nullable
+	ExcludedNamespacedResources []string `json:"excludedNamespacedResources,omitempty"`
+    ......
+}
+```
+
+## Alternatives Considered
+Proposal from Jibu Data [Issue 5120](https://github.com/vmware-tanzu/velero/issues/5120#issue-1304534563)
+
+## Security Considerations
+No security impact.
+
+## Compatibility
+The four new parameters cannot be mixed with existing resource filter parameters: `IncludedResources`, `ExcludedResources` and `IncludeClusterResources`.
+If the new parameters and old parameters both appears in command line, or are specified in backup spec, the command line and the backup should fail.
+
+## Implementation
+This change should be included into Velero v1.11.
+New parameters will coexist with `IncludedResources`, `ExcludedResources` and `IncludeClusterResources`.
+Plan to deprecate `IncludedResources`, `ExcludedResources` and `IncludeClusterResources` in future releases, but also open to the community's feedback.
+
+## Open Issues
+`LabelSelector/OrLabelSelectors` apply to namespaced scope resources.
+It may be reasonable to make them also working on cluster scope resources.
+An issue is created to trace this topic [resource label selector not work for cluster scope resources](https://github.com/vmware-tanzu/velero/issues/5787)
--- a/changelogs/unreleased/5333-blackpiglet
+++ b/changelogs/unreleased/5333-blackpiglet
@@ -1,0 +1,1 @@
+Proposal to separate resource filter into cluster scope and namespace scope


design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,219 @@
...
+Two new parameters are added into command `velero backup create`: `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources`. 
+They are used to filter cluster scope resources included or excluded in backup.
+`--include-cluster-scope-resources` valid value includes `*` and comma separated string.
+`--exclude-cluster-scope-resources` valid value is comma separated string.
+
+### Using scenarios:
+#### some namespaced resources + no additional cluster resources
+``` bash
+velero backup create <backup-name>
+    --include-namespaces=default,kube-system
```
sseago, line 30: Does this mean that IncludeClusterScopedResources does not include all resources by default like IncludeResources does? With IncludeResources, leaving it off is equivalent to setting it to "*".

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,219 @@
...
+    --include-cluster-scope-resources=*
+```
+
+or
+
+``` bash
+velero backup create <backup-name>
+    --include-namespaces=default,kube-system
+    --include-cluster-resources=true
+```
```
sseago, line 53: These two have different meanings, so I don't think those two are equivalent. IncludeClusterResources is a bool pointer with 3 possible values. `true` means all cluster resources are included (still subject to other filters such as label selector and -- with the current codebase -- include/exclude-resources). `false` means all are excluded, regardless of the other selection criteria. And `nil` means "only consider cluster resources that are directly relevant to namespaced resources being included -- specifically, this means PVs when bound to included PVCs, CRDs when backing up associated CRs, and other cluster resources as pulled in by backup plugins. I think if we're adding the new Include/ExcludeClusterScopeResources fields, we still need to factor in include-cluster-resources. For example, you could set it to true and then use the new fields to fine tune which resources you want. That way you can still set it to nil and only bring in relevant ones (i.e. so you don't pull in every PV), and still exclude things like CRDs, etc.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,219 @@
...
+
+#### all namespaced resources + no additional cluster resources
+``` bash
+velero backup create <backup-name>
+    --include-namespaces=*
+```
```
sseago, line 59: include-cluster-resources would be nil here, so you'd pull in only relevant cluster resources (PVs, etc.)

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,219 @@
...
+
+#### all namespaced resources + some additional cluster resources
+``` bash
+velero backup create <backup-name>
+    --include-namespaces=*
+    --include-cluster-scope-resources=PersistentVolume
```
sseago, line 65: If you want *all* PVs (and not just ones relevant to PVCs) you'd also want to set include-cluster-resources to true.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,219 @@
...
+    --include-cluster-scope-resources=*
+```
+
+or
+
+``` bash
+velero backup create <backup-name>
+    --include-namespaces=*
+    --include-cluster-resources=true
+```
```
sseago, line 81: As before, it's not *or* but you'd want both `--include-cluster-resources=true` and `--include-cluster-scope-resources=*` (although the latter would be unnecessary if we decide that it behaves like include-resources and it's "*" by default if omitted.

sseago (commented):
In addition to above comments, would it be an error if a namespaced resource is listed in include/exclude-cluster-scope-resources? What about cluster-scoped resources listed in include/exclude-resources? If so, wouldn't this be a backwards-incompatible change?

design/cluster-scope-resource-filter.md
blackpiglet: What about moving the same logic in `include-cluster-resources` to `IncludeClusterScopeResources`?

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,219 @@
...
+	IncludeClusterResources *bool `json:"includeClusterResources,omitempty"`
+
+	// IncludedClusterScopeResources is a slice of cluster scope 
+	// resource names to include in the backup.
+	// If empty, no additional cluster scope resources are included.
+	// If set to "*", all cluster scope resources are included.
+	// Supersede IncludedResources in conflict case.
+	// +optional
+	// +nullable
+	IncludedClusterScopeResources []string `json:"includedClusterScopeResources,omitempty"`
```
sseago, line 140: As mentioned on the call, lets also create `IncludedNamespacedResources` and `ExcludedNamespaceResources` where empty/"*" rules match the rules for the Cluster resources, and we want validation to ensure that a user either uses IncludedResources/ExcludedResources *or* some combination of the four new fields, but not both. Old fields work just like they do now. New fields respect the cluster/namespaced split.

design/cluster-scope-resource-filter.md
blackpiglet: Modified.

design/cluster-scope-resource-filter.md
blackpiglet: Added.

design/cluster-scope-resource-filter.md
blackpiglet: Modified.

design/cluster-scope-resource-filter.md
blackpiglet: Right.

design/cluster-scope-resource-filter.md
blackpiglet: Modified.

blackpiglet:
@sseago 
Document is updated according to comments. PTAL.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+
+* `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` can only contain cluster scope resource type names. If namespace scope resource type names are included, validation failure should be returned.
+
+* If there are conflicts between `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` specified resources type lists, validation failure should be returned.
+
+* `--include-namespaced-resources` and `--exclude-namespaced-resources` can only contain namespace scope resource type names. If cluster scope resource type names are included, validation failure should be returned.
+
+* If there are conflicts between `--include-namespaced-resources` and `--exclude-namespaced-resources` specified resources type lists, validation failure should be returned.
+
+* If  `--include-namespaced-resources` is not present, it means including all namespace scope resources per resource type.
```
sseago, line 70: I think this one should probably function like include-cluster-scope-resources for consistency. If not present, none included. Users should be required to include it with "*" -- this is a change from behavior of the other include-resources field, but that way with the new way of specifying resources, cluster-scoped and namespaced resources are handled in the same way to avoid confusion.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+### Using scenarios:
+#### no namespaced resources + no cluster resources
+This is not allowed. Backup or restore cannot contain no resource.
+
+#### no namespaced resources + some cluster resources
+The following commands all mean backup no namespaced resources and some cluster resources.
+
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaces=*
```
sseago, line 89: I don't think wildcards are currently supported for exclude namespaces. However, instead of doing this, it would be clearer to simply leave `--include-namespaced-resources` empty, if the behavior is modified according to my suggestion above.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--include-cluster-resources=true
+	--include-cluster-scope-resources=StorageClass
+```
+
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaced-resources=*
```
sseago, line 96: This line won't be necessary since an empty "include" list will mean nothing is included.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--include-cluster-resources=true
+	--include-cluster-scope-resources=StorageClass
+```
+
+``` bash
+velero backup create <backup-name>
```
sseago, line 102: We can probably skip this one since the above example already covers this.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--include-cluster-resources=true
+	--include-cluster-scope-resources=StorageClass
+```
+
+#### no namespaced resources + all cluster resources
+The following commands all mean backup no namespaced resources and all cluster resources.
+
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaces=*
```
sseago, line 114: I don't think we need this -- just make sure `--include-namespace-resources` is absent/empty.

sseago: Also note, this will include no cluster-scoped resources unless `--include-cluster-scope-resources=*` is added.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--include-cluster-resources=true
+```
+
+``` bash
+velero backup create <backup-name>
```
sseago, line 119: Above suggestions make this one unnecessary.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--exclude-namespaces=*
+	--include-cluster-resources=true
+	--include-cluster-scope-resources=*
+```
+
+``` bash
+velero backup create <backup-name>
```
sseago, line 126: We don't need this one if the above example is fixed.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--exclude-namespaced-resources=*
+	--include-cluster-resources=true
+```
+
+``` bash
+velero backup create <backup-name>
```
sseago, line 132: We don't need this one if the above example is fixed.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--exclude-namespaced-resources=*
+	--include-cluster-resources=true
+	--include-cluster-scope-resources=*
+```
+
+``` bash
+velero backup create <backup-name>
```
sseago, line 139: We don't need this one if the above example is fixed.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--exclude-namespaces=*
+	--exclude-namespaced-resources=*
+	--include-cluster-resources=true
+```
+
+``` bash
+velero backup create <backup-name>
```
sseago, line 146: We don't need this one if the above example is fixed.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--exclude-namespaced-resources=*
+	--include-cluster-resources=true
+	--include-cluster-scope-resources=*
+```
+
+#### some namespaced resources + no cluster resources
+This means backup all resources in namespaces default and kube-system, and no cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
```
sseago, line 157: This one should be fine as-is. Since we're not using any of the new params, the `--include-resources="*"` implied by an empty param should apply.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--include-namespaced-resources=Deployment,Service,Endpoint,Pod,ReplicaSet
+	--include-cluster-resources=false
+```
+
+This means backup all resources except Ingress type resources in all namespaces, and no cluster resources.
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaced-resources=Ingress
```
sseago, line 179: Also, `--include-namespaced-resources="*"`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-cluster-resources=false
+```
+
+This means backup Deployment, Service, Endpoint, Pod and ReplicaSet resources in all namespaces, and no cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=Deployment,Service,Endpoint,Pod,ReplicaSet
```
sseago, line 164: This could also be handled using the older `--include-resources` since cluster-scoped resources are excluded anyway.

design/cluster-scope-resource-filter.md
sseago: Alternatively, you could use `--exclude-resourcs=Ingress` without the include param since the old params include everything by default.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+This means backup all resources in namespaces default and kube-system, and related cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+```
+
+This means backup Deployment, Service, Endpoint, Pod and ReplicaSet resources in all namespaces, and related cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=Deployment,Service,Endpoint,Pod,ReplicaSet
```
sseago, line 193: Since we're using an explicit list of resources here, to include related cluster resources we'll need to list the types we want to include: `--include-cluster-scope-resources=PersistentVolume` or whatever. Since `include-cluster-resources` isn't included, we'll still only include those cluster resources that are relevant to the namespaced resources though, not everything of those types.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+This means backup Deployment, Service, Endpoint, Pod and ReplicaSet resources in namespaces default and kube-system, and related cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-namespaced-resources=Deployment,Service,Endpoint,Pod,ReplicaSet
```
sseago, line 200: Since we're using an explicit list of resources here, to include related cluster resources we'll need to list the types we want to include: --include-cluster-scope-resources=PersistentVolume or whatever. Since include-cluster-resources isn't included, we'll still only include those cluster resources that are relevant to the namespaced resources though, not everything of those types.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+This means backup all resources except Ingress type resources in all namespaces, and related cluster resources.
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaced-resources=Ingress
```
sseago, line 206: We also need `--include-namespaced-resources="*"`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+#### some namespaced resources + some additional cluster resources
+This means backup all resources in namespace in default, kube-system, and related cluster resources, plus all StorageClass cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
```
sseago, line 213: `--include-namespaced-resources="*"`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--exclude-cluster-scope-resources=StorageClass
+	--include-cluster-resources=true
+```
+
+#### some namespaced resources + all cluster resources
+This means backup all resources in namespace in default, kube-system, and all cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
```
sseago, line 248: `--include-namespaced-resources="*"`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--include-cluster-scope-resources=*
+	--include-cluster-resources=true
+```
+
+This means backup Deployment, Service, Endpoint, Pod and ReplicaSet resources in all namespaces, and all cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=Deployment,Service,Endpoint,Pod,ReplicaSet
+	--include-cluster-resources=true
```
sseago, line 257: `--include-cluster-scope-resources="*"`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+This means backup Deployment, Service, Endpoint, Pod and ReplicaSet resources in default and kube-system namespaces, and all cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-namespaced-resources=Deployment,Service,Endpoint,Pod,ReplicaSet
+	--include-cluster-resources=true
```
sseago, line 265: `--include-cluster-scope-resources="*"`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+#### all namespaced resources + only related cluster resources
+The following commands all mean backup all resources in all namespaces, and related cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=*
+```
+
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=*
```
sseago, line 303: With the above, you'll also need `--include-cluster-scope-resources="*"` If you remove the line above, then it's not needed as we'll be using the old includes/excludes rules

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=*
+	--include-namespaced-resources=*
```
sseago, line 309: With the above, you'll also need --include-cluster-scope-resources="*" If you remove the line above, then it's not needed as we'll be using the old includes/excludes rules

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+``` bash
+velero backup create <backup-name>
+```
+
+#### all namespaced resources + some additional cluster resources
+The following commands all mean backup all namespace scope resources, and related cluster resources, plus all PersistentVolume resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=*
+	--include-cluster-scope-resources=PersistentVolume
```
sseago, line 321: `--include-namespaced-resources="*" --include-cluster-resources=true`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=*
+	--include-cluster-scope-resources=PersistentVolume
```
sseago, line 327: `--include-cluster-resources=true`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=*
+	--include-namespaced-resources=*
+	--include-cluster-scope-resources=PersistentVolume
```
sseago, line 334: `--include-cluster-resources=true`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+```
+
+``` bash
+velero backup create <backup-name>
+	--include-cluster-scope-resources=PersistentVolume
```
sseago, line 339: `--include-cluster-resources=true --include-namespaced-resources="*"

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--include-namespaces=*
+	--include-namespaced-resources=*
+	--include-cluster-scope-resources=*
+	--include-cluster-resources=true
+```
+
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=*
+	--include-namespaced-resources=*
```
sseago, line 355: `--include-cluster-scope-resources=*`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+	--include-cluster-resources=true
+```
+
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=*
+	--include-cluster-scope-resources=*
```
sseago, line 362: --include-namespaced-resources=*

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=*
+	--include-cluster-scope-resources=*
+	--include-cluster-resources=true
+```
+
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=*
```
sseago, line 381: `--include-cluster-scope-resources=*`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+
+## Security Considerations
+No security impact.
+
+## Compatibility
+### IncludedResources and ExcludedResources change
+To make the change backward compatible, `IncludedResources/ExcludedResources` parameters still apply to both cluster scope and namespace scope.
+After adding new parameters `IncludedClusterScopeResources`, `ExcludedClusterScopeResources`, `IncludedNamespacedResources` and `ExcludedNamespacedResources`, there could be conflict between the exiting and new parameters.
+When there is conflict, `IncludedClusterScopeResources/ExcludedClusterScopeResources` parameters have higher priority for cluster scope resources.
+`IncludedNamespacedResources/ExcludedNamespacedResources` parameters have higher priority for namespace scope resources.
```
sseago, line 552: Using priority rules here will be confusing. Instead, of any of the four new fields are specified, then the two old fields must not be specified, and only new inclusion rules are applied. If none of the four new fields are specified, then the old inclusion rules are specified. This means if none of the fields (old or new) are specified, then we use the old rule where omitting `IncludedResources` is equivalent to setting `IncludedResources` to `true`.

sseago: If users specify fields from both old and new lists, then it will be a validation error, similar to what we do with `LabelSelector` and `OrLabelSelectors`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+
+### IncludeClusterResources change
+`IncludeClusterResources` parameter is an existing boolean parameter, which implies whether to include cluster scope resources.
+There are three possible values of this parameter.
+In previous release, not setting value means including cluster scope resources related to namespaced scope resources included in backup and restore; `true` means including all cluster resources; `false` means excluding all cluster resources.
```
sseago, line 557: We should note that there is no change to the meaning if this parameter. It will function exactly like before.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,572 @@
...
+
+After this change, `IncludeClusterResources`'s not setting value and `false`'s meaning are not changed.
+If `IncludeClusterResources` is set to `true`, and `IncludedClusterScopeResources/ExcludedClusterScopeResources` are not set, all cluster resources are included.
```
sseago, line 560: I think this particular one introduces confusion. It will be simpler if both IncludedClusterScopeResources and IncludedNamespacedResources must always be present with a value of "*" to include all, otherwise we assume none are included.

design/cluster-scope-resource-filter.md
blackpiglet: This case includes related cluster sources, not all cluster resources.

design/cluster-scope-resource-filter.md
blackpiglet: This case includes related cluster sources, not all cluster resources.

design/cluster-scope-resource-filter.md
blackpiglet: Modified.

design/cluster-scope-resource-filter.md
sseago: @blackpiglet Right, but the problem with the above is that the includes/excludes will exclude all cluster resources, even related ones. By adding all cluster-scoped resources to the list of resources included but leaving include-cluster-resources as nil/auto, the result will be including all related cluster-scoped resources. There are selection decisions, and it's the merging of the two that ultimately determines what resources are included:
1) list of included resource types. This is determined by either the include/excludeResources (old way, mixed cluster/namespaced) or the new include/exclude namespaced/cluster-scope. Any resources *not* included in this list will be excluded, regardless of other selection criteria
2) includeClusterResources: "true" would include everything cluster-scoped, subject to exclusions in 1. "false" would exclude all cluster-scoped, regardless of inclusions in 1. "nil/auto" includes only related cluster resources, again subject to inclusions in 1.

design/cluster-scope-resource-filter.md
sseago: Agreed. See above comment for why we still need to make sure that we're not excluding all cluster resources in the resource list.

sseago (dismissed):

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,502 @@
...
+velero backup create <backup-name>
+	--include-cluster-resources=true
+	--include-cluster-scope-resources=StorageClass
+```
+
+#### no namespaced resources + all cluster resources
+The following commands all mean backup no namespaced resources and all cluster resources.
+
+``` bash
+velero backup create <backup-name>
```
reasonerjt, line 99: That is the reason why I think the flag `--include-cluster-resources` is redundant...

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,502 @@
...
+	--exclude-cluster-scope-resources=StorageClass
+	--include-cluster-resources=true
+```
+
+#### some namespaced resources + all cluster resources
+This means backup all resources in namespace in default, kube-system, and all cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-namespaced-resources=*
```
reasonerjt, line 233: is `--include-namespaced-resources=*` really needed?  Isn't this default behavior?

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,502 @@
...
+
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=*
+	--include-cluster-scope-resources=*
+	--include-cluster-resources=true
+```
+
+#### describe command change
+In `velero backup describe` command, the four new parameters should be outputted too.
```
reasonerjt, line 325: If some the of parameters are mutual exclusive, we should print the new ones only, otherwise it will confuse the users.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,502 @@
...
+No security impact.
+
+## Compatibility
+### IncludedResources and ExcludedResources change
+To make the change backward compatible, `IncludedResources/ExcludedResources` parameters still apply to both cluster scope and namespace scope.
+After adding new parameters `IncludedClusterScopeResources`, `ExcludedClusterScopeResources`, `IncludedNamespacedResources` and `ExcludedNamespacedResources`, there could be conflict between the exiting and new parameters.
+When there is conflict, `IncludedClusterScopeResources/ExcludedClusterScopeResources` parameters have higher priority for cluster scope resources.
+`IncludedNamespacedResources/ExcludedNamespacedResources` parameters have higher priority for namespace scope resources.
+
+### IncludeClusterResources change
```
reasonerjt, line 488: See my comment above, IMO it would make it easier to understand if we deprecate this flag if `IncludedClusterScopeResources/ExcludedClusterScopeResources`  are introduced.
Otherwise, having both `IncludeClusterResources` and `IncludedClusterScopeResources` are confusing.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,502 @@
...
+`IncludeClusterResources` parameter is an existing boolean parameter, which implies whether to include cluster scope resources.
+There are three possible values of this parameter.
+In previous release, not setting value means including cluster scope resources related to namespaced scope resources included in backup and restore; `true` means including all cluster resources; `false` means excluding all cluster resources.
+
+After this change, `IncludeClusterResources`'s not setting value and `false`'s meaning are not changed.
+If `IncludeClusterResources` is set to `true`, Velero will include cluster resources according to `IncludedClusterScopeResources/ExcludedClusterScopeResources` parameters setting.
+
+## Implementation
```
reasonerjt, line 496: Please update the version information here and from the inline comment in the code example.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,502 @@
...
+This change should be included into Velero v1.10.
+Plan to deprecate `IncludedResources` and `ExcludedResources` in Velero v1.11.
+
+## Open Issues
```
reasonerjt, line 500: Agreed we should apply labelSelector in cluster scope resources, but we should double-check the workflow, if it's tricky to implement it in v1.11 we can track it separately as a tech-debt.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,502 @@
...
+
+* `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` can only contain cluster scope resource type names. If namespace scope resource type names are included, validation failure should be returned.
+
+* If there are conflicts between `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` specified resources type lists, validation failure should be returned.
+
+* `--include-namespaced-resources` and `--exclude-namespaced-resources` can only contain namespace scope resource type names. If cluster scope resource type names are included, validation failure should be returned.
+
+* If there are conflicts between `--include-namespaced-resources` and `--exclude-namespaced-resources` specified resources type lists, validation failure should be returned.
+
+* If  `--include-namespaced-resources` is not present, it means no namespace scope resource is included per resource type.
```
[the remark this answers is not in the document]
blackpiglet: After a discussion with Scott, we think it's better to make the two parameters share the same filtering rule, which is to include nothing if the parameter is not specified, and users need to specify the resources to include by parameter. If all resources are preferred, asterisk should be used.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,502 @@
...
+`--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` are used to filter cluster scope resources included or excluded in backup per resource type.
+`--include-namespaced-resources` and `--exclude-namespaced-resources` are used to filter namespace scope resources included or excluded in backup per resource type.
+
+### Parameters Rules
+
+* `--include-cluster-scope-resources` and `--include-namespaced-resources` valid value include `*` and comma separated string.
+
+* `--exclude-cluster-scope-resources` and `--exclude-namespaced-resources` valid value are comma separated string.
+
+* `--include-cluster-scope-resources`, `--include-namespaced-resources`, `--exclude-cluster-scope-resources` and `--exclude-namespaced-resources` parameters are mutual exclusive with `--include-resources` and `--exclude-resources` parameters. If both sets of parameters are provisioned, validation failure should be returned.
```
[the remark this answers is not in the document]
sseago: @reasonerjt Yes. They serve different purposes. the --include/exclude-cluster-scope-resources overlaps with include/exclude-resources -- basically include/exclude is split into two categories. include-cluster-resources is a boolean pointer with 3 possible values which control how the include/exclude resources are processed.
1) nil: the default value. This means we only pull in cluster-scoped resources as needed by other resources (i.e. PVs, CRDs, etc.). This behavior is *modified* by includes/excludes -- so you could add CRD to exclude-cluster-scope-resources (or exclude-resources) which will result in related PVs being included but not CRDs
2) true: this means include cluster resources in general, but again modified by the includes/excludes. If includes/excludes include everything, then this means all CRDs are included. If you want to include all CRDs but not all ClusterRoleBindings, then you'd set `--include-cluster-resources --include-cluster-scope-resources=customresourcedefinitions` to include only CRDs or  `--include-cluster-resources --exclude-cluster-scope-resources=clusterrolebindings` to include all cluster resources *except* cluster role bindings.
3) false: this excludes all cluster resources

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,502 @@
...
+
+* `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` can only contain cluster scope resource type names. If namespace scope resource type names are included, validation failure should be returned.
+
+* If there are conflicts between `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` specified resources type lists, validation failure should be returned.
+
+* `--include-namespaced-resources` and `--exclude-namespaced-resources` can only contain namespace scope resource type names. If cluster scope resource type names are included, validation failure should be returned.
+
+* If there are conflicts between `--include-namespaced-resources` and `--exclude-namespaced-resources` specified resources type lists, validation failure should be returned.
```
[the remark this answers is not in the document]
sseago: This shouldn't be a conflict. This is the only way to specify "include all namespaced resources except Deployments" The logic would be to get a list of namespaced resources from includes, then remove anything from excludes from that list.

design/cluster-scope-resource-filter.md
sseago: It's not redundant -- see above. Without this parameter, we lose an important use case.

design/cluster-scope-resource-filter.md
sseago: It's not. With the old parameters, leaving include-resources blank is equivalent to `--include-resources=*`, but "blank means all" doesn't work with the new split parameters, as it means there's no way to include nothing for either namespaced or cluster-scoped. That's why the latest version of this design removes the notion of "include all by default" with `--include-namespaced-resources` and `--include-cluster-scope-resources`. For those, the `=*` is required in order to account for all use cases.

design/cluster-scope-resource-filter.md
sseago: We can't do that without losing an important use case. I agree that the naming is somewhat confusing, but we haven't been able to come up with a better name for the new IncludeClusterScopeResources parameter that is any clearer, and we can't change the old name without losing backwards compatibility. We just need to make sure that our documentation is very clear on the interaction of the parameters.

design/cluster-scope-resource-filter.md
blackpiglet: Sure. I agree that this is not the main purpose of this task.

design/cluster-scope-resource-filter.md
blackpiglet: Sure.

design/cluster-scope-resource-filter.md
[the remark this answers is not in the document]
reasonerjt: @sseago per our discussion, for cluster-scope-resources we will always include those that are needed unless the user explicitly tells velero to exclude them from the backup, with this approach we may no longer need the 
bool flag `--include-cluster-resources`.  With that said, we should double-check the code to make sure this is doable with reasonable effort.

cc @blackpiglet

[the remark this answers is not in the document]
blackpiglet: @sseago 
Yes. Now, the backup has two resource filter parameters in the `backup.Request` structure: `NamespaceIncludesExcludes` and `ResourceIncludesExcludes`.

If we desperate `--include-cluster-resources` parameter, the `NamespaceIncludesExcludes` should have no change: if the namespace is listed in exclude array, it should be ignored, if there is a match in the include array, or include array contains an asterisk, the namespace is a keeper.
`ResourceIncludesExcludes` should be split into two parameters: 
* `NamespaceResourceIncludesExcludes`
* `ClusterResourceIncludesExcludes`. 

In the `backup` process, 
* The `NamespaceResourceIncludesExcludes` parameter should work the same way with `NamespaceIncludesExcludes`. 
* The `ClusterResourceIncludesExcludes` parameter should be kept by default, as long as it's not explicitly listed in the exclude array.

In the `restore` process,
* The `NamespaceResourceIncludesExcludes` should work the opposite as the backup process. Resources should be included by default unless it's in the excluded array.
* The `ClusterResourceIncludesExcludes` parameter works the same way as the backup.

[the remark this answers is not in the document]
blackpiglet: Not sure about the `Restore.Spec.RestoreStatus`. Looks like it needs no change.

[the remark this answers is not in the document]
sseago: @blackpiglet 
"If we depreracte the --include-cluster-resources" 
From the last discussion, we shouldn't be deprecating this -- we should group this with the other existing params --include-resources and --exclude-resources. A user should either use some combination of the 3 old params and none of the new ones, or some combination of the four new params and none of the old ones. In other words, --include-cluster-resources is valid as long as the user doesn't also include any of the namespaced/cluster-scope params.

"The NamespaceResourceIncludesExcludes parameter should work the same way with NamespaceIncludesExcludes."

Almost, but not quite. NamespaceIncludesExcludes includes everything unless there's a specific resource list in includes, and then only those resources are included. So blank is the same thing as "*". For NamespaceResourceIncludesExcludes, blank means "don't include anything", while "*" means "include everything" -- this is to support the "don't include anything namespaced" use case.

"The ClusterResourceIncludesExcludes parameter should be kept by default, as long as it's not explicitly listed in the exclude array."
I don't think this is quite right. For cluster resources, there are 3 categories. Resources which are explicitly included will all be in the backup (i.e. if you include PVs, then *all* PVs are in the backup). Resources which are explicitly excluded will not be in the backup at all (if you exclude PVs, then *no* PVs are in the backup). Resources which are neither included nor excluded (missing in both lists) will only be pulled in if they are necessary for other items included. If PVs are in neither list, then we pull in PVs only if the PVC they're bound to is in the backup.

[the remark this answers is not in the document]
reasonerjt: > "If we depreracte the --include-cluster-resources"
> From the last discussion, we shouldn't be deprecating this -- we should group this with the other existing params --include-resources and --exclude-resources. A user should either use some combination of the 3 old params and none of the new ones, or some combination of the four new params and none of the old ones. In other words, --include-cluster-resources is valid as long as the user doesn't also include any of the namespaced/cluster-scope params.

IMO, we should consider deprecating the existing 3 params `--include-resources`/ `--exclude-resources` and  `--include-cluster-resources` from user interface if the new ones work well, we shouldn't be maintaining 2 sets of filters doing the same thing.

I find discussing the logic in English is more difficult than showing the code, therefore, if we reach an agreement regarding the behavior of the attributes, we should go ahead and merge this design and leave further implementation discussion in the code review.

In terms of delivering the features, if it's a bit challenge to deliver the new filters (--include/exclude-namespaced-resources, --include/exclude-cluster-scope-resources) for both backup and restore,  shall we try to deliver the new filters in backup first?  Because most problems are reported around backup.

@sseago @blackpiglet  what do you think?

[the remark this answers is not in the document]
sseago: @reasonerjt I'm open to the idea, but I'd like to get input from users and customers I think. I'll reach out on the Red Hat side. As long as we go through a proper deprecation process (still in discussion, but sticking around for at least 2 releases after deprecation, etc.), there's no hard obstacle to removing them, and it would definitely simplify maintenance. It's possible that some users won't like having to think through separating namespaced/cluster-scoped resources, but thinking about it, as long as we get appropriate input before deprecating, this may be the best thing. It might be best to release the new params in a release first without deprecation, so people can experiment with both methods in practice and provide feedback before we make any final calls on deprecation.

Also, yes, I think it's reasonable to do this on backup only initially. Backup filtering is the most complicated, since we're working with everything in the cluster. On the restore side, it's just a matter of filtering on types that are already included in the backup. We may actually not want the new complexity on restore. Lets target backup side for initial implementation.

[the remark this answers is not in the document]
blackpiglet: @sseago 
Thanks for the comments. I will modify the document ASAP.
The agreement by far is keeping existing filters and creating new filters, and the new filters only work for backup resource handling.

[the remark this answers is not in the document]
sseago: @blackpiglet That sounds good. If we do eliminate the existing filters, it will be in a future release after following the deprecation process, so for 1.11 the new and old filters must both work (with validation to make sure only one group is used for a given backup).

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+Four new parameters are added into command `velero backup create`: `--include-cluster-scope-resources`, `--exclude-cluster-scope-resources`, `--include-namespaced-resources` and `--exclude-namespaced-resources`. 
+`--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` are used to filter cluster scope resources included or excluded in backup per resource type.
+`--include-namespaced-resources` and `--exclude-namespaced-resources` are used to filter namespace scope resources included or excluded in backup per resource type.
+Restore and other code pieces also use resource filtering will be handled in future releases.
+
+### Parameters Rules
+
+* `--include-namespaced-resources` valid value is comma separated string. Not setting the parameter means including all namespace resources.
+
+* `--include-cluster-scope-resources`, `--exclude-cluster-scope-resources` and `--exclude-namespaced-resources` valid value include `*` and comma separated string.
```
reasonerjt, line 57: 1. Shall we enforce that the format has to be `$resource.$group`?  As I recall this is more consistent with existing behavior.
2. shall we allow the values to contain wildcard in the string items, like `--exclude-namespaced-resources=*.velero.io`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+
+* `--include-cluster-scope-resources`, `--include-namespaced-resources`, `--exclude-cluster-scope-resources` and `--exclude-namespaced-resources` parameters are mutual exclusive with `--include-cluster-resources`, `--include-resources` and `--exclude-resources` parameters. If both sets of parameters are provisioned, validation failure should be returned.
+
+* `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` can only contain cluster scope resource type names. If namespace scope resource type names are included, validation failure should be returned.
```
reasonerjt, line 61: I'm not sure if such validation is needed, esp for CRs.  
To simplify the effort if the user puts a namespaced resource in `--include-cluster-scope-resources`, like `--include-cluster-scope-resources=backup.velero.io` we can just ignore it, b/c it won't be effective anyway, correct?

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+
+* If there are conflicts between `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` specified resources type lists, validation failure should be returned.
```
reasonerjt, line 63: What does `conflict` mean here?  IMO we should allow these two sets have overlap, but we need to define the priority.  i.e. If resource A is included in both `--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` will it be included in the backup or not?  IMO, we should exclude it.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+
+* `--include-namespaced-resources` and `--exclude-namespaced-resources` can only contain namespace scope resource type names. If cluster scope resource type names are included, validation failure should be returned.
+
+* If there are conflicts between `--include-namespaced-resources` and `--exclude-namespaced-resources` specified resources type lists, validation failure should be returned.
+
+* If  `--include-namespaced-resources` is not present, it means all namespace scope resources are included per resource type.
+
+* If `--include-cluster-scope-resources` is not present, it means no additional cluster resource is included per resource type. Cluster resources are related to the namespace scope resources, which means those are returned in the namespace resources' BackupItemAction's result AdditionalItems array, are still included in backup by default. If the parameter is set, listed resources and namespaced resources related cluster resources are included in backup.
```
reasonerjt, line 71: Could you elaborate some examples, like `pv` and `crd`

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+#### no namespaced resources + no cluster resources
+This is not allowed. Backup or restore cannot contain no resource.
+
+#### no namespaced resources + some cluster resources
+The following command means backup no namespaced resources and some cluster resources.
+
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaced-resources=*
+	--include-cluster-scope-resources=StorageClass
```
reasonerjt, line 85: See comment the above comment about the format of the string items, it should be `storageclasses.storage.k8s.io`

Please also update in the following examples.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+The following command means backup no namespaced resources and all cluster resources.
+
+``` bash
+velero backup create <backup-name>
+	--exclude-namespaced-resources=*
+	--include-cluster-scope-resources=*
+```
+
+#### some namespaced resources + no cluster resources
+The following commands mean backup all resources in namespaces default and kube-system, and no cluster resources.
```
reasonerjt, line 98: Let's highlight that resources potentially be returned by BIA like `pv` or `crd` will also be excluded.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--exclude-cluster-scope-resources=*
+```
+
+``` bash
```
reasonerjt, line 105: This one should be removed, it has nothing to do with the new params.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+This means backup all resources in namespaces default and kube-system, and related cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+```
+
+The following commands mean backup Deployment, Service, Endpoint, Pod and ReplicaSet resources in all namespaces, and related cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=Deployment,Service,Endpoint,Pod,ReplicaSet
```
reasonerjt, line 162: But in this case resource like `pvc` or `cr` will be excluded right, therefore will there be any related cluster resources returned by BIA?

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+``` bash
+velero backup create <backup-name>
+	--include-resources=Deployment,Service,Endpoint,Pod,ReplicaSet
+```
+
+The following commands mean backup Deployment, Service, Endpoint, Pod and ReplicaSet resources in namespaces default and kube-system, and related cluster resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaces=default,kube-system
+	--include-namespaced-resources=Deployment,Service,Endpoint,Pod,ReplicaSet
```
reasonerjt, line 174: see comment above, I don't think any related cluster resources will be returned?

design/cluster-scope-resource-filter.md
reasonerjt: I'm not sure if this is needed anymore, b/c  when --include-namespaced-resources is not set it means include everything.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+The four new parameters cannot be mixed with existing resource filter parameters: `IncludedResources`, `ExcludedResources` and `IncludeClusterResources`.
+If the new parameters and old parameters both appears in command line, or are specified in backup spec, the command line and the backup should fail.
+
+## Implementation
+This change should be included into Velero v1.11.
+New parameters will coexist with `IncludedResources`, `ExcludedResources` and `IncludeClusterResources`.
+Plan to deprecate `IncludedResources`, `ExcludedResources` and `IncludeClusterResources` in future releases, but also open to the community's feedback.
+
+## Open Issues
+`LabelSelector/OrLabelSelectors` apply to namespaced scope resources.
```
reasonerjt, line 393: Let's make sure there is an issue opened for this and reference it here.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,394 @@
...
+
+## High-Level Design
+Four new parameters are added into command `velero backup create`: `--include-cluster-scope-resources`, `--exclude-cluster-scope-resources`, `--include-namespaced-resources` and `--exclude-namespaced-resources`. 
+`--include-cluster-scope-resources` and `--exclude-cluster-scope-resources` are used to filter cluster scope resources included or excluded in backup per resource type.
+`--include-namespaced-resources` and `--exclude-namespaced-resources` are used to filter namespace scope resources included or excluded in backup per resource type.
+Restore and other code pieces also use resource filtering will be handled in future releases.
+
+### Parameters Rules
+
+* `--include-namespaced-resources` valid value is comma separated string. Not setting the parameter means including all namespace resources.
```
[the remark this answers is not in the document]
blackpiglet: Sure. I agree. I was thinking to provide only one way to express the scenario.

design/cluster-scope-resource-filter.md
blackpiglet: Just confirmed. `$kind-name`, `$kind-name.$group-name`, and capital or small all work, for example, `storageClass.storage.k8s.io, servicE, configMap`.
And because this is a rather flexible parameter, working out a regular expression may not be easy. I will try to implement it, but if that consumes effort, may not do it in this release.

design/cluster-scope-resource-filter.md
blackpiglet: Agree. Modified.

design/cluster-scope-resource-filter.md
blackpiglet: Right. Include should precede exclude.

design/cluster-scope-resource-filter.md
blackpiglet: Sure. Adding backing PVC will include related PV as the example.

design/cluster-scope-resource-filter.md
blackpiglet: I add the detailed supported parameter format in the above section.

design/cluster-scope-resource-filter.md
blackpiglet: Sure. Modified.

design/cluster-scope-resource-filter.md
blackpiglet: Add examples of old parameters, when they work.
By this, the reader can find out which scenarios only new parameters can support.

design/cluster-scope-resource-filter.md
blackpiglet: Add PVC into the example.

design/cluster-scope-resource-filter.md
blackpiglet: Modified.

design/cluster-scope-resource-filter.md
blackpiglet: OK. #5787 is created.

reasonerjt (approved):

ihcsim (commented):
@blackpiglet Thanks for putting this together. I feel like the introduction of 4 new options, together with the existing `--include-resources`, `--exclude-resources` and `--include-cluster-resources` options, bloats up the configuration permutations and validations quite a bit. As a user, I find it hard to reason about when to use what.

Is it possible to just introduce options to modify the behaviour of the `--include-resources` option, while preserving the current behaviour? 

E.g., can we have something like `--scope` and `--include-filter-mode`? Specifically,
* `--scope` specifies the scope of the resources in the backup job  
    * supported values: 
       * `cluster` - Perform backup at cluster scope with no need for ["empty" namespace](https://github.com/vmware-tanzu/velero/issues/5120#issuecomment-1188965236). Doesn't work with `--include-namespaces` and `--exclude-namespaces`
       * `namespaced` - Perform backup at namespace scope. Default and align with current backup behaviour
* `--include-filter-mode` specifies whether a backup filter limits the included resources, or augments the namespace with the additional resources
    * supported values:
       * `augment` - Add the resources specified in the include list to the backup job while retaining the job's scope
       * `limit` - Backup _only_ the resources specified in the include list. Default and align with current behaviour
    * to be used with `--include-resources`

So to backup selected cluster resources only:
```sh
velero backup create <backup-name> \
    --scope=cluster \
    --include-resources=storageclass,customresourcedefinition
```

(Options within the square brackets are optional, defaults added for illustration purposes only.)

No changes required to backup only all resources in the `default` and `kube-system` namespaces:
```sh
velero backup create <backup-name> \
    [--scope=namespaced \] 
    --include-namespaces=default,kube-system \
    --include-cluster-resources=false
```

No change required to backup selected namespace-scoped resources in all namespaces, and their related cluster resources (PVC's related PV):
```sh
velero backup create <backup-name> \
    [--include-cluster-resources=nil|auto \] 
    [--include-filter-mode=limit \]
    [--scope=namespaced \]
    --include-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
```

To backup all namespace-scoped resources in `default` and `kube-system` namespaces, their related cluster-scoped resources, plus all storage classes:
```sh
velero backup create <backup-name> \
    [--include-cluster-resources=nil|auto \] 
    [--scope=namespaced \]
    --include-namespaces=default,kube-system \
    --include-resources=storageclass,customresourcedefinition \
    --include-filter-mode=augment
```

To backup selected namespace-scoped resources in the `default` and `kube-system` namespaces, their related cluster resources, plus all storage classes and custom resource definitions:
```sh
velero backup create <backup-name> \
    [--include-cluster-resources=nil|auto \] 
    [--scope=namespaced \]
    --include-namespaces=default,kube-system \
    --include-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset,storageclass,customresourcedefinition \
    --include-filter-mode=limit
```

When it comes to backing up _all_ or _some_ cluster-scoped resources, together with namespace-scoped resources, I think separating them into 2 backup jobs makes restores esaier to reason about. 

E.g., to backup selected namespace-scoped resources in `default` and `kube-system` namespaces, and their related cluster resources, plus all cluster scope resources except storage classes and custom resource definitions:
```sh
# backup all cluster resources
velero backup create <backup-name> \
    [--include-resources=* \]
    --exclude-resources=storageclass,customresourcedefinitions \
    --scope=cluster

# backup specific resources in namespaces
velero backup create <backup-name> \
    [--include-cluster-resources=nil|auto \] 
    [--scope=namespaced \]
    --include-namespaces=default,kube-system \
    --include-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset \
    --include-filter-mode=limit
```

Let me know what you think of this approach. If it makes sense, I can share what the usages of `--scope` and `--include-filter-mode` look like in order to satisfy the other scenarios described in this proposal.

design/cluster-scope-resource-filter.md
```diff
@@ -0,0 +1,410 @@
...
+
+This means backup PVC, Deployment, Service, Endpoint, Pod and ReplicaSet resources in default and kube-system namespaces, and related cluster resources, plus all cluster scope resources except StorageClass type resources.
+``` bash
+velero backup create <backup-name>
+	--include-namespaced-resources=persistentvolumeclaim,deployment,service,endpoint,pod,replicaset
+	--include-namespaces=default,kube-system
+	--exclude-cluster-scope-resources=storageclass
+```
+
+#### some namespaced resources + all cluster resources
```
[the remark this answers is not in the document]
blackpiglet: @ihcsim
Thanks for your input.
I agree that these new parameters would make Velero CLI more complicated.

I'm not sure about how to include all cluster scope resources except several specific resources and have all namespace resources in a namespace with your proposal.
I saw your feeling about this case. Your suggestion is separate into different backups, but there are also some user requirements for this using scenario.

blackpiglet:
@sseago 
Hi Scott, could you help to review this PR when you are availabe?

design/cluster-scope-resource-filter.md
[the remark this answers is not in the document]
ihcsim: I think it's not just the CLI, presumably, the API will be complicated too, right? Do we know what the APIs will look like?

>  include all cluster scope resources except several specific resources and have all namespace resources in a namespace with your proposal.

Is there a limitation on `--exclude-resources` that prohibits us from achieving this with something like:
```sh
velero backup create <backup-name> \
    --include-cluster-resources=true \
    --include-namespaces=default,kube-system \
    --exclude-resources=persistentvolumeclaim,persistentvolume,customresourcedefinitions
```

design/cluster-scope-resource-filter.md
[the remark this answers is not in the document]
sseago: @reasonerjt @blackpiglet "We'll just highlight that not setting --include-namespaced-resources has the same effect as --include-namespaced-resources=*"
I thought from earlier discussion, we wanted the ability to include none, so omitting it means no namespaced resources, and including `*` means include all. Certainly for cluster-scoped, skipping the param is completely different from including `*` (the former only includes relevant resources, the latter includes everything) -- I thought we agreed that for consistency, with the new params, there would be no implied `*` -- it must be explicit.

design/cluster-scope-resource-filter.md
sseago: Currently wildcard is only really supported in includes, and it's only `*` for all -- I don't think `*.velero.io` is supported on include or exclude. We could consider this sort of flexible wildcard support as a follow-on enhancement, since it's not supported in the current implementation either.

design/cluster-scope-resource-filter.md
sseago: Yes, add includes, then remove excludes. If something is listed in both places, there's no error, it's just not included.

design/cluster-scope-resource-filter.md
sseago: "* If  `--include-namespaced-resources` is not present, it means all namespace scope resources are included per resource type."
I don't think this is correct. We should require the include `*`, that way a user can choose to not include any namespaced resources by omitting it.

design/cluster-scope-resource-filter.md
sseago: @reasonerjt That's the old behavior. We agreed earlier to require this for a couple reasons:
1) consistency with include-clsuter-scope-resources
2) Allow not including any namespaced resources.

sseago:
@blackpiglet I think there's still some disagreement and/or miscommuncation on the behavior when `include-namespaced-resources` is omitted. We had previously decided that omitting means "include nothing" to be consisten with cluster-scoped resource behavior, as well as to support the "include no namespaced resources" use case. See comments above in response to other comments.

design/cluster-scope-resource-filter.md
blackpiglet: @sseago 
After a discussion with Daniel, we think when `--include-namespaced-resources` is not set, it means including every namespace resource; when `--include-cluster-scope-resources` is not set, it means including related cluster scope resources, which is the same with existing `--include-cluster-resources`.

To not include any namespace resources, we set `--exclude-namespaced-resources=*`.

blackpiglet:
@sseago 
Maybe there is some miscommunication.
After discussion with Daniel, the conclusion of the default values of `--include-namespaced-resources` and `--include-cluster-scope-resources` is different from the previous design. 

If `--include-namespaced-resources` is not set, it means including every namespace resource.
If `--include-cluster-scope-resources` is not set, it means only including related cluster resources.
Because the default values of the two parameters are different, so need to make them align with each other.
Default to including resources, specifying exclude rule to rule out is more intuitive and it has the same meaning as the existing filters.

design/cluster-scope-resource-filter.md
blackpiglet: Please see the previous comment and the [overall comment](https://github.com/vmware-tanzu/velero/pull/5333#issuecomment-1410218645).
There is some change in the default value meaning. Please check whether this works for you.

design/cluster-scope-resource-filter.md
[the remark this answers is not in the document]
blackpiglet: Please see the previous comment and the https://github.com/vmware-tanzu/velero/pull/5333#issuecomment-1410218645.
There is some change in the default value meaning. Please check whether this works for you.

design/cluster-scope-resource-filter.md
blackpiglet: Agreed. I will check whether this is doable and the effort. May not do it this time.

design/cluster-scope-resource-filter.md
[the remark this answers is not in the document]
blackpiglet: Yes. There would be more parameters. Please check the detailed design part. There is an example of `BackupSpec` change. But they are separated into two groups, and they cannot work together. They are supposed to be replacements for old parameters. Although more parameters make the code logic more complicated, that would not be a burden for the users to understand.

@sseago @reasonerjt 
What's your opinion about Ivan's proposal? I didn't go through all the scenarios yet. I think even not working for all possibilities by now, after some modification, it should do the same job, but in a quite different way.
I think that depends on the purpose. If we want to enhance the current parameters and keep existing behavior, Ivan's proposal is better. If we want to re-implement the filters, I think the current proposal is better.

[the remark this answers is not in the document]
ihcsim: What are the reasons behind replacing/deprecating the old flags? Feels like it will be quite a process since they are commonly used. Following the ongoing discussion, deciding on the default behavior of the new flags feels complicated enough already. Feels like this will impose quite a bit of cognitive burden on the users.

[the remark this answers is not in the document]
sseago: @ihcsim See the linked issue for the discussion around this. Short answer is that there are several desired use cases that are impossible to specify with the old set of parameters, and the new parameters provide the flexibility to meet them: https://github.com/vmware-tanzu/velero/issues/5120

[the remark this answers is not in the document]
blackpiglet: This is the original issue to request this feature https://github.com/vmware-tanzu/velero/issues/5120.
To support filtering both the cluster-scoped resources and namespace resources, we need to do some modifications to the filters.
To me, either way, it will raise the using complexity. Just need to choose which way is better or preferred.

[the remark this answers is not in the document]
sseago: @blackpiglet As for the proposal overall, I think it's at least as complicated as what we're proposing here. To see whether it covers all of the use cases, someone would need to go through all of the use cases that we're solving in this proposal and see whether there are any gaps there. Overall, I think this proposal feels more intuitive -- it's less of a radical departure from current behavior. As for proliferation of configuration options, it's important to note that a user will either use the old params or the new ones, not both -- so instead of at most 3 params in use, there will be at most 4 now.

sseago:
@blackpiglet as for the question of the implied `--include-namespaced-resources=*`, the suggestion above would also work, but we'll have to make wildcards work on exclude. Currently, they only really work properly on includes.

blackpiglet:
> @blackpiglet as for the question of the implied `--include-namespaced-resources=*`, the suggestion above would also work, but we'll have to make wildcards work on exclude. Currently, they only really work properly on includes.

Sure. This is already covered in the design document.

`--include-cluster-scope-resources`, `--include-namespaced-resources`, `--exclude-cluster-scope-resources` and `--exclude-namespaced-resources` valid value include `*` and comma separated string. Each element of the CSV string should have a k8s resource name. The format should be `resource.group`, such as `storageclasses.storage.k8s.io.`.

blackpiglet:
@sseago 
Just as you know, all ongoing discussions are clarified. Could you help PTAL?

design/cluster-scope-resource-filter.md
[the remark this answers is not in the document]
ihcsim: The benefit of introducing new options to augment the behaviour of existing include/exclude/include-cluster resources options is that users don't have to re-learn a new filtering system, nor modify their existing configuration if they choose not to.

On the other hand, I do think the existing filtering options are limited for nowadays' requirements. If we are going to invest the time into enhancing this, we should do it with the hope of not having to re-do this exercise again in say, 2 years time.

If we are going to deprecate **and then remove** the older options (especially, `--include-cluster-resources`), the community should be given sufficient time to adapt the changes. E.g., 2 releases with deprecation warning, then removal from CLI, then finally removal from API. We should also think through the upgrade path.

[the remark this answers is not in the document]
blackpiglet: @ihcsim 
FYI, if this PR is merged, we can still trace the topic you mentioned in issue #5120.

sseago (approved):

reasonerjt (approved):

PR #5345, opened 2022-09-15
Pod Volume Backup/Restore Refactor: Rename Command Parameters

Lyndon-Li:
Pod Volume Backup/Restore Refactor: Rename parameters in CRDs and commands to remove "Restic" word


pkg/controller/schedule_controller.go
```diff
@@ -142,6 +142,15 @@ func parseCronSchedule(itm *velerov1.Schedule, logger logrus.FieldLogger) (cron.
 		return nil, validationErrors
 	}
 
+	// For new created schedules, DefaultVolumesToRestic is not allowed
+	// TODO: post v1.10. Remove this code block after DefaultVolumesToRestic is removed from CRD
+	if itm.Status.Phase != velerov1.SchedulePhaseEnabled {
```
ywk253100, line 147: The `parseCronSchedule` runs not only for the newly created schedules, so this will cause the old schedule failure

pkg/apis/velero/v1/backup.go
```diff
@@ -101,9 +101,15 @@ type BackupSpec struct {
 	// DefaultVolumesToRestic specifies whether restic should be used to take a
 	// backup of all pod volumes by default.
 	// +optional
-	// + nullable
+	// +nullable
```
ywk253100, line 104: How about adding the `deprecated` marker to indicate that the property will be removed in the next version of CRD? https://book.kubebuilder.io/reference/markers/crd.html

pkg/controller/schedule_controller.go
ywk253100: And preventing the usage of `DefaultVolumesToRestic` is a break change which should introduce a new version of CRD

pkg/cmd/cli/backup/create.go
```diff
@@ -132,7 +132,7 @@ func (o *CreateOptions) BindFlags(flags *pflag.FlagSet) {
 	f = flags.VarPF(&o.IncludeClusterResources, "include-cluster-resources", "", "Include cluster-scoped resources in the backup")
 	f.NoOptDefVal = "true"
 
-	f = flags.VarPF(&o.DefaultVolumesToRestic, "default-volumes-to-restic", "", "Use restic by default to backup all pod volumes")
+	f = flags.VarPF(&o.DefaultVolumesToFsBackup, "default-volumes-to-fs-backup", "", "Use pod volume file system backup by default for volumes")
```
ywk253100, line 135: Have we decided to introduce the breaking change in the CLI? Please don't forget to add release note for this if we have decided

Lyndon-Li: Yes, this has been decided because we don't want users to use the `default-volumes-to-restic `for Kopia path.
We will mention this change in the coming upgrade guide doc, the new pod volume backup doc as well as the release note.

pkg/apis/velero/v1/backup.go
Lyndon-Li: As the discussion, we will not use the `deprecated` marker for the current case. The reason is as below:
We will not change the CRD's version(as the conclusion here). On the other hand, the above `deprecated` marker is for the CRD spec wides, that is to mark a deprecated CRD; while our requirement is to mark a single field for a deprecation in future version (it is just a warning), so it is not suitable to our requirement.

pkg/controller/schedule_controller.go
Lyndon-Li: After a discussion, the conclusion is as below:

- We will not change the CRD's version for now. We will accumulate more changes of the CRD and change the version for all, we will also add mutation webhook.
- We will not block the newly created schedule CR's if their DefaultVolumesToRestic field is set to true
- Instead, we add a warning log when we find the to-be-deprecated field is used

pkg/cmd/server/server.go
```diff
@@ -213,7 +212,7 @@ func NewCommand(f client.Factory) *cobra.Command {
 	command.Flags().StringVar(&config.pluginDir, "plugin-dir", config.pluginDir, "Directory containing Velero plugins")
 	command.Flags().StringVar(&config.metricsAddress, "metrics-address", config.metricsAddress, "The address to expose prometheus metrics")
 	command.Flags().DurationVar(&config.backupSyncPeriod, "backup-sync-period", config.backupSyncPeriod, "How often to ensure all Velero backups in object storage exist as Backup API objects in the cluster. This is the default sync period if none is explicitly specified for a backup storage location.")
-	command.Flags().DurationVar(&config.podVolumeOperationTimeout, "restic-timeout", config.podVolumeOperationTimeout, "How long backups/restores of pod volumes should be allowed to run before timing out.")
+	command.Flags().DurationVar(&config.podVolumeOperationTimeout, "pod-volume-timeout", config.podVolumeOperationTimeout, "How long backups/restores of pod volumes should be allowed to run before timing out.")
```
reasonerjt, line 215: Doesn't this have the same problem that we are exposing the concept `pod-volume` to end user?

pkg/controller/backup_controller_test.go
```diff
@@ -359,20 +359,20 @@ func TestProcessBackupCompletions(t *testing.T) {
...
 
 	tests := []struct {
-		name                   string
-		backup                 *velerov1api.Backup
-		backupLocation         *velerov1api.BackupStorageLocation
-		defaultVolumesToRestic bool
-		expectedResult         *velerov1api.Backup
-		backupExists           bool
-		existenceCheckError    error
+		name                     string
```
reasonerjt, line 362: Cases should be added to verify the combination of different values of `defaultVolumesToFsBackup` and `defaultVolumesToRestic` in the input params.

pkg/cmd/server/server.go
Lyndon-Li: There is an existing annotation, which is used for the same purpose, that is named with "pod-volume", so we just to keep it the same with the annotation:
PodVolumeOperationTimeoutAnnotation = "velero.io/pod-volume-timeout"

Otherwise, if we want to change this also, we need to change another user interface thing -- annotation, do we need to do this?

pkg/apis/velero/v1/backup.go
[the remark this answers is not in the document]
Lyndon-Li: Done, added the deprecation marker in the description.

pkg/controller/backup_controller_test.go
Lyndon-Li: Done, added the test cases.

pkg/cmd/server/server.go
Lyndon-Li: As the discussion, we will change the flag only, because the annotation has not been in any of Velero document. The flag name will be `fs-backup-timeout`

Lyndon-Li: Done

Issue #5400, opened 2022-09-26
Incorrect Backup CR config for CSI backup causes backup to fail due to runtime error


shubham-pampattiwar:
**What steps did you take and what happened:**
When we try to use Velero backup for CSI volumes via incorrect Backup CR config, the backup fails due to nil pointer dereference and velero pod restarts. For example: 
```
apiVersion: velero.io/v1
kind: Backup
metadata:
  name: cluster-pvcs
  namespace: openshift-adp
spec:
  csiSnapshotTimeout: 10m0s
  defaultVolumesToRestic: false
  includedNamespaces:
    - '*'
  includedResources:
    - PersistentVolumeClaim
  labelSelector:
    matchLabels:
      oadp-backup: 'true'
  storageLocation: openshift-backups-1
  ttl: 24h0m0s
```
This backup CR will cause incomplete execution of the velero-plugin-for-csi as the `includedResources` only consists of PVC and thus excluded VolumeSnapshots, VolumeSnapshotContents etc.

Error logs in velero pod as velero tries to delete resources that do not exist:
```
2022/09/15 02:23:35  info Waiting for CSI driver to reconcile volumesnapshot mssql-persistent/velero-mssql-pvc-f7x6j. Retrying in 5s
2022/09/15 02:23:40  info Waiting for CSI driver to reconcile volumesnapshot mssql-persistent/velero-mssql-pvc-f7x6j. Retrying in 5s
2022/09/15 02:23:45  info Waiting for CSI driver to reconcile volumesnapshot mssql-persistent/velero-mssql-pvc-f7x6j. Retrying in 5s
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1a8b3cd]

goroutine 2372 [running]:
github.com/vmware-tanzu/velero/pkg/controller.(*backupController).deleteVolumeSnapshot.func1(0xc000d5e140)
        /remote-source/src/github.com/vmware-tanzu/velero/pkg/controller/backup_controller.go:949 +0xad
created by github.com/vmware-tanzu/velero/pkg/controller.(*backupController).deleteVolumeSnapshot
        /remote-source/src/github.com/vmware-tanzu/velero/pkg/controller/backup_controller.go:945 +0xf7
```
**What did you expect to happen:**
Backup to complete successfully

**Vote on this issue!**

[...]

PR #5401, opened 2022-09-26
Add nil check before execution of csi snapshot delete

shubham-pampattiwar:
Signed-off-by: Shubham Pampattiwar <spampatt@redhat.com>

Fixes https://github.com/vmware-tanzu/velero/issues/5400

### DIFF
--- a/changelogs/unreleased/5401-shubham-pampattiwar
+++ b/changelogs/unreleased/5401-shubham-pampattiwar
@@ -1,0 +1,1 @@
+Add nil check before execution of csi snapshot delete
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -707,7 +707,9 @@ func (c *backupController) runBackup(backup *pkgbackup.Request) error {
 		}
 
 		// Delete the VolumeSnapshots created in the backup, when CSI feature is enabled.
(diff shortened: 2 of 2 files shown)


blackpiglet:
I think after supporting must-include annotation in CSI plugin, the scenario described in #5400 is already fixed.

sseago:
@reasonerjt @blackpiglet I think nil pointer errors are serious enough that we should really check for nil even if we think it's impossible for a particular variable to be nil. They have a way of coming back to haunt us later, and an error which crashes velero is always more serious than an error which results in a backup/restore error.

I think the nil check adds a bit of safety even if we expect it won't happen anymore. Especially considering that it's possible that a user will register an older version of the CSI plugin which doesn't have the annotation.

shubham-pampattiwar:
@reasonerjt @blackpiglet I truly believe that adding a simple check nil check wouldn't hurt. Simply because if the Velero pod restarts then the current backup (which will be in its last stage before completion) will get stuck and Velero would not process it. I think we need this check.

blackpiglet:
I agree that adding more check is helpful.
There is still some check error in this PR. Could you please help to resolve those? I think merging main branch may help to resolve the errors.

shubham-pampattiwar:
@blackpiglet Done, updated the PR.

blackpiglet (approved):

sseago (approved):

PR #5442, opened 2022-10-12
BackupItemAction v2 API implementation

sseago:

Fixes #(issue)


kaovilai:
This PR closes https://github.com/vmware-tanzu/velero/issues/5488

weshayutin:
my very newb scan of leaves me w/ just a thought and I don't have a strong opinion here.
Thank you for updating the unit tests and labeling the overriden api tests w/ `V2`.  Would it help the reader of the unit tests results to name the item w/ a `V2` as well even though there is NOT a V1?  e.g. NewBackupItemActionPlugin -> NewBackupItemActionPluginV2

sseago:
@weshayutin I'm not seeing any funcs named `NewBackupItemActionPlugin` in the test files. Where I see a function of this name in the other code, it's in a package that identifies it as v1 vs. v2. Was there some test in particular that you were referencing?

pkg/backup/item_backupper.go
```diff
@@ -320,7 +320,9 @@ func (ib *itemBackupper) executeActions(
 		}
 		log.Info("Executing custom action")
 
-		updatedItem, additionalItemIdentifiers, err := action.Execute(obj, ib.backupRequest.Backup)
+		// Note: we're ignoring the operationID returned from Execute for now, it will be used
+		// with the async plugin action implementation
+		updatedItem, additionalItemIdentifiers, _, err := action.Execute(obj, ib.backupRequest.Backup)
```
Lyndon-Li, line 325: After the `Execute` method is called, the method may return in a short time but the async operation is still ongoing. 
As to the backup workflow, I guess the caller need to call the `Progress` method to wait the completion of the async operation.
For different items, multiple `Execute` methods may be called successively before their async operations complete, so before the backup workflow ends, multiple async operations should be waited for completion.

Will we handle this logic in the current PR or we do it in later PRs?

pkg/plugin/clientmgmt/backupitemaction/v2/restartable_backup_item_action.go
```diff
@@ -0,0 +1,171 @@
...
+
+	api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	biav1cli "github.com/vmware-tanzu/velero/pkg/plugin/clientmgmt/backupitemaction/v1"
+	"github.com/vmware-tanzu/velero/pkg/plugin/clientmgmt/process"
+	"github.com/vmware-tanzu/velero/pkg/plugin/framework/common"
+	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+	biav2 "github.com/vmware-tanzu/velero/pkg/plugin/velero/backupitemaction/v2"
+)
+
+// AdaptedBackupItemAction is a backup item action adapted to the v1 BackupItemAction API
```
Lyndon-Li, line 31: v1->v2

pkg/plugin/clientmgmt/backupitemaction/v2/restartable_backup_item_action.go
```diff
@@ -0,0 +1,171 @@
...
+	delegate, err := r.getDelegate()
+	if err != nil {
+		return velero.ResourceSelector{}, err
+	}
+
+	return delegate.AppliesTo()
+}
+
+// Execute restarts the plugin's process if needed, then delegates the call.
+func (r *RestartableBackupItemAction) Execute(item runtime.Unstructured, backup *api.Backup) (runtime.Unstructured, []velero.ResourceIdentifier, string, error) {
```
Lyndon-Li, line 110: Thinking of some plugin process restart cases:
1. If the plugin process quits unexpectedly (i.e. crash) during a method calling, the call will fail on the client side, so the backup/restore will fail because of the BIA/RIA method fails. Is that right?
2. Since the plugin runs some async operations, the plugin may quit outside a method calling, for example, the `Execute` method returns successfully but during running the async operation afterwards, the plugin crashes. In this case, I think the client side gets no notification from the plugin framework, but the specific BIA/RIA implementation needs to tell the caller during the later method calling (i.e., `Progress` method) if it wants to fail the current backup/restore. How do we cover this scenario?

pkg/backup/item_backupper.go
sseago: @Lyndon-Li This PR is just the API definition. It adds the appropriate functions to the various BackupItemAction-related packages. Implmeenting the controller workflow changes to actually call Progress, wait, maintain a list of items waited for, etc. will be implemented in a separate PR (there's a separate github issue to track that work).

pkg/plugin/clientmgmt/backupitemaction/v2/restartable_backup_item_action.go
sseago: @Lyndon-Li
1. It should result in a partial failure, as the GRPC call itself will error out, but that's not any different from what happens with v1 plugins
2. If the plugin crashes without restarting, then the Progress method will return an error (similar to the Execute error in your first question), so this will push us to partial failure. If the plugin crashes but restarts, then it's up to the plugin to handle this. For example, for a datamover plugin which creates a separate datamover CR (where the actual async operation happens in a separate controller), a plugin restart after creating the datamover CR won't affect progress, as long as the plugin is running again by the time velero calls Progress next, since it just checks the datamover CR status. If the plugin is itself running the operation in a background thread of the plugin process, then a crash will probably result in an operation failure, so Progress will return an error.

pkg/plugin/clientmgmt/backupitemaction/v2/restartable_backup_item_action.go
sseago: I'll reword it to make it clearer. "is a v1 BackupItemAction adapted to implement the v2 API"

pkg/plugin/clientmgmt/backupitemaction/v2/restartable_backup_item_action.go
Lyndon-Li: Sounds good, either way is allowed, the behavior is up to the specific plugin.

Lyndon-Li (approved):

shubham-pampattiwar (approved):

shubham-pampattiwar:
LGTM ! Thank you @sseago

PR #5445, opened 2022-10-13
Design for velero upgrade command

Lyndon-Li:
Add the design to create a new velero upgrade command

### DIFF
--- a/design/velero-upgrade.md
+++ b/design/velero-upgrade.md
@@ -1,0 +1,85 @@
+# Velero Upgrade Design  
+
+## Background
+
+Besides a fresh installation, users also require to upgrade from old version with preserving the existing tasks (i.e., backup schedules) and configurations (i.e., Velero server & daemonset parameters). At present, the recommended Velero upgrade approach is finished manually with below steps:  
+- Update all CRDs by running ```velero install --crds-only --dry-run -o yaml | kubectl apply -f -```
+- Update Velero server and daemonset image by running ```kubectl set image```  
+
+For a complete example of the existing upgrade approach, refer to the Velero document at https://velero.io/docs/v1.9/upgrade-to-1.9/.
+
+The manual approach works for the existing scenarios, however, if we want to do some complex tasks during upgrade (i.e., installing a new module or changing/adding complex parameters), it will be very unfriendly in the user experiences.  
+For an concrete example, in v1.10, we have done some refactors to Velero's pod volume backup (a.k.a Restic backup formerly), two of the refactors are:  
+- Rename some Velero server and daemonset parameters
+- Rename Velero daemonset from "restic" to "node-agent"
+
+As a result, the existing upgrade approach is not sufficient:
+- If we still use manual steps, we have to tell users the detailed parameters that have been renamed, and users then needs to check their existence and modify them one by one.
+- If we still use manual steps, users have to delete the old restic daemonset and run ```velero install``` with the full parameters they specified during the fresh installation, because for one thing, Kubernetes doesn't support to rename a resource; for another, Velero doesn't provide a way to install a single resource only. 
+
+## Solution
+
+We want to add a new command besides ```velero install``` called ```velero upgrade```. In a upgrade case, users could call ```velero upgrade``` without bothering to know the details. This bring below benefits:  
+- Solve the upgrade problems in v1.10
+- Make Velero more user friendly, because the upgrade process is significantly simplified
+- Support future evolvements. For example, in future, Velero will have various data movement features, for which we will add new moudles inevitably. Then we can add enhancement incrementally to ```velero upgrade```, because inside the command, we can do whatever complex things 
+
+The ```velero upgrade``` command will fully replace the existing manual steps in the aforementioned Velero document.
+
+## Design
+
+At present, ```velero upgrade``` will do below things:
+- Get the new version from the client binary itself, that is, the upgrade target version
+- Get the old version from the existing velero server, that is, the current version
+- Retrieve Velero server deployment object
+- Retrieve Velero daemonset object
+- Iterate from current version through target version, apply the changes (i.e., the image, name and parameter changes) to the deployment object once per version. For example, when upgrading from 1.9 to 1.11, we need to apply the changes between 1.9 and 1.10 first, then apply the changes between 1.10 and 1.11. See the details below for how to drive the changes apply
+- Do the same changes apply to Velero daemonset object as the deployment object
+- Delete the existing Velero deployment
+- Delete the existing Velero daemonset
+- Extract the new CRDs from the client binary and apply them
+- Create the new Velero daemonset from the modified daemonset object
+- Create the new Velero server deployment from the modified deployment object
+
+### Action Table
+We create below action functions to apply changes for each target version, the number in the function name indicates the target version:  
+```
+func UpgradeV10(deploy *v1.Deployment, ds *v1.DaemonSet) error
+func UpgradeV11(deploy *v1.Deployment, ds *v1.DaemonSet) error
+func UpgradeV12(deploy *v1.Deployment, ds *v1.DaemonSet) error
+```
+
+Then we put these functions to a vector, so that the action functions could be retrieved by indexes:  
+```
+type UpgradeAction struct {
+	Target string
+	Action func(*v1.Deployment, *v1.DaemonSet) error
+}
+
+var UpgradeActions = []UpgradeAction {
+	{
+		"v1.10",
+		UpgradeV10,
+	},
+	{
+		"v1.11",
+		UpgradeV11,
+	},
+	{
+		"v1.12",
+		UpgradeV12,
+	},	
+}
+```
+So ```UpgradeActions[0]``` means the action to upgrade from older versions to v1.10; ```UpgradeActions[1]``` means the action to upgrade from v1.10 to v1.11.  
+Then we match the current version and target version in ```UpgradeActions``` to get a begin-end index pair.  
+Finally, we call the action functions between the begin index and end index.  
+
+### Image Selection
+The target image is mapped by the target version, which is retrieved from the Velero client binary. If the client is from a under-release version, we will set the image as ```velero/velero:latest```.
+
+As we can see above, all the required information is self-retrieved, and the upgrade actions are self-driven, so we don't need to add any extra parameters to ```velero upgrade```.  
+
+## Non-Goals
+As the best practise, there should be some kind of recovery measure so that when the upgrade fails, the original environment could be recovered. This design doesn't include it, we could handle this in the future enhancements.  
+Therefore, at present, under this situation, users need to manual fix the half-way environment or uninstall it and then reinstall it freshly.  
--- a/changelogs/unreleased/5445-lyndon
+++ b/changelogs/unreleased/5445-lyndon
@@ -1,0 +1,1 @@
+Add the design to create a new velero upgrade command


design/velero-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+- Retrieve Velero daemonset object
+- Iterate from current version through target version, apply the changes (i.e., the image, name and parameter changes) to the deployment object once per version. For example, when upgrading from 1.9 to 1.11, we need to apply the changes between 1.9 and 1.10 first, then apply the changes between 1.10 and 1.11. See the details below for how to drive the changes apply
+- Do the same changes apply to Velero daemonset object as the deployment object
+- Update the Velero server deployment from the modified deployment object
+- Delete the existing Velero daemonset
+- Create the new Velero daemonset from the modified daemonset object
+- Extract the new CRDs from the client binary and apply them
+
+### Action Table
+We create below action functions to apply changes for each target version, the number in the function name indicates the target version:
```
reasonerjt, line 44: I think there are more details we need to be careful with, for example, if we released 1.10.0 then 1.11.0, then 1.10.1 then 1.11.1.

How do we manage the upgrade path if 1.11.1 require user to upgrade to 1.10.1 but 1.10.1 cannot be upgraded to 1.11.0?

Do we need to add more pre-requisite items?

design/velero-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+- If we still use manual steps, users have to delete the old restic daemonset and run ```velero install``` with the full parameters they specified during the fresh installation, because for one thing, Kubernetes doesn't support to rename a resource; for another, Velero doesn't provide a way to install a single resource only. 
+
+## Solution
+
+We want to add a new command besides ```velero install``` called ```velero upgrade```. In a upgrade case, users could call ```velero upgrade``` without bothering to know the details. This bring below benefits:  
+- Solve the upgrade problems in v1.10
+- Make Velero more user friendly, because the upgrade process is significantly simplified
+- Support future evolvements. For example, in future, Velero will have various data movement features, for which we will add new moudles inevitably. Then we can add enhancement incrementally to ```velero upgrade```, because inside the command, we can do whatever complex things 
+
+The ```velero upgrade``` command will fully replace the existing manual steps in the aforementioned Velero document.
```
reasonerjt, line 27: I assume the upgrade does not support any parameters like 
`velero upgrade --image`

It only supports hard coded?  If the answer is yes, how do we test the upgrade before the GA image is pushed can be a little tricky I hope we can add more detail in this design to clarify that.

design/velero-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+func UpgradeV10(deploy *v1.Deployment, ds *v1.DaemonSet) error
+func UpgradeV11(deploy *v1.Deployment, ds *v1.DaemonSet) error
+func UpgradeV12(deploy *v1.Deployment, ds *v1.DaemonSet) error
+```
+
+Then we put these functions to a vector, so that the action functions could be retrieved by indexes:  
+```
+type UpgradeAction struct {
+	Target string
+	Action func(*v1.Deployment, *v1.DaemonSet) error
```
reasonerjt, line 55: I understand this is to demonstrate the idea, but I hope we can make we follow the conventional golang approach.

Normally we don't add functions as fields in the declaration of a struct....

design/velero-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+- Make Velero more user friendly, because the upgrade process is significantly simplified
+- Support future evolvements. For example, in future, Velero will have various data movement features, for which we will add new moudles inevitably. Then we can add enhancement incrementally to ```velero upgrade```, because inside the command, we can do whatever complex things 
+
+The ```velero upgrade``` command will fully replace the existing manual steps in the aforementioned Velero document.
+
+## Design
+
+At present, ```velero upgrade``` will do below things:
+- Get the new version from the client binary itself, that is, the upgrade target version
+- Get the old version from the existing velero server, that is, the current version
```
reasonerjt, line 33: How do we get that, is it via the `Server Status` CR?

design/velero-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+- Retrieve Velero server deployment object
+- Retrieve Velero daemonset object
+- Iterate from current version through target version, apply the changes (i.e., the image, name and parameter changes) to the deployment object once per version. For example, when upgrading from 1.9 to 1.11, we need to apply the changes between 1.9 and 1.10 first, then apply the changes between 1.10 and 1.11. See the details below for how to drive the changes apply
```
reasonerjt, line 36: Do we need to wait for the v1.10 to be running before proceeding to v1.11? If the CRD of v1.10 is not applied at this point of time, there may be risk that the v1.10 server fail to be running.

design/velero-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+- Do the same changes apply to Velero daemonset object as the deployment object
+- Update the Velero server deployment from the modified deployment object
+- Delete the existing Velero daemonset
+- Create the new Velero daemonset from the modified daemonset object
+- Extract the new CRDs from the client binary and apply them
+
+### Action Table
+We create below action functions to apply changes for each target version, the number in the function name indicates the target version:  
+```
+func UpgradeV10(deploy *v1.Deployment, ds *v1.DaemonSet) error
```
reasonerjt, line 46: Not sure if we really need the input params here.

design/velero-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+}
+```
+So ```UpgradeActions[0]``` means the action to upgrade from older versions to v1.10; ```UpgradeActions[1]``` means the action to upgrade from v1.10 to v1.11.  
+Then we match the current version and target version in ```UpgradeActions``` to get a begin-end index pair.  
+Finally, we call the action functions between the begin index and end index.  
+
+As we can see above, all the required information is self-retrieved, and the upgrade actions are self-driven, so we don't need to add any extra parameters to ```velero upgrade```.  
+
+## Non-Goals
+As the best practise, there should be some kind of recovery measure so that when the upgrade fails, the original environment could be recovered. This design doesn't include it, we could handle this in the future enhancements.
```
reasonerjt, line 80: So if `velero upgrade` fails we will ask user to uninstall velero and install the latest version, right?

Lyndon-Li: Yes, this situation is the same as the existing manual upgrade approach, once it fails, users need to either do the uninstall-install steps or try to remedy it based on the failed installation.

design/velero-upgrade.md
Lyndon-Li: Yes, this is just a prototype, to demonstrate what will be changed in the action functions. We may change the function parameters in the implementation, then we come back to modify this design through new PRs.

design/velero-upgrade.md
Lyndon-Li: Good point, let me modify the order of the steps to solve this problem.

design/velero-upgrade.md
Lyndon-Li: Correct

design/velero-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+		"v1.11",
+		UpgradeV11,
+	},
+	{
+		"v1.12",
+		UpgradeV12,
+	},	
+}
+```
+So ```UpgradeActions[0]``` means the action to upgrade from older versions to v1.10; ```UpgradeActions[1]``` means the action to upgrade from v1.10 to v1.11.
```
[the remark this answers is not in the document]
Lyndon-Li: We can make some hard coding if necessary.
Actually, for all the old version before v1.10, the action is the same: 1. upgrade CRDs; 2. update image in deployment and daemonset. Therefore, we actually don't need to test all the versions.

design/velero-upgrade.md
Lyndon-Li: First of all, if we have this situation -- "1.10.1 cannot be upgraded to 1.11.0", this means Velero have multiple side streams, it seems we don't have this behavior at present? Moreover, in order to support side streams, our entire release cycle management, code management, and team resource management have to be changed, so do we need to consider this in the current design?

Secondly, once Velero has this multiple stream manner, it doesn't mean this action table concept not work totally, merely, the action table will not be a linear vector but a 2d tree like structure. Therefore, can we add this requirement to the NonGoal section and consider it later?

Finally, if we only consider the single stream release module only, this action table concept could still support patch releases. On the other hand, as a principle, the patch releases don't include complex changes or user experience changes, we only need to provide a general action function, that is to change the image only, but this is not mandatory.

Lyndon-Li: Let's further discuss on this then I will make the change to the design.

design/velero-upgrade.md
Lyndon-Li: Added some explanation for the failed case.

design/velero-upgrade.md
Lyndon-Li: Changed the order of the action steps.

design/velero-upgrade.md
Lyndon-Li: Added explanations for under-release Velero client case.

design/velero-upgrade.md
reasonerjt: So in theory there can be depedency loop, when the CR is updated to another version you will not be able to get the version from the old version of velero server.

design/velero-upgrade.md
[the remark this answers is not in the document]
reasonerjt: We should consider this as a conventional approach considering `velero upgrade` is not only for v1.9 -> v1.10 upgrade.

For the version we decided not to test, IMO we should fail early and tell user they need to upgrade to a certain version first and run `velero upgrade`

design/velero-upgrade.md
Lyndon-Li: Here the version is used to decide the action function only, and the action function defines how the modules are installed/upgraded and what parameters are set.
In another words, by the original consideration, the action function doesn't aim to change CRs because Kubernetes has already had better way to solve it --- the mutation webhook.

Let's discuss this particularly to see if we need to add the CR mutation into the action functions.

design/velero-upgrade.md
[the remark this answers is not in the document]
Lyndon-Li: Sure. So we start to support from v1.9? Do we all agree on this?

design/velero-upgrade.md
```diff
@@ -0,0 +1,85 @@
...
+- Get the old version from the existing velero server, that is, the current version
+- Retrieve Velero server deployment object
+- Retrieve Velero daemonset object
+- Iterate from current version through target version, apply the changes (i.e., the image, name and parameter changes) to the deployment object once per version. For example, when upgrading from 1.9 to 1.11, we need to apply the changes between 1.9 and 1.10 first, then apply the changes between 1.10 and 1.11. See the details below for how to drive the changes apply
+- Do the same changes apply to Velero daemonset object as the deployment object
+- Delete the existing Velero deployment
+- Delete the existing Velero daemonset
+- Extract the new CRDs from the client binary and apply them
+- Create the new Velero daemonset from the modified daemonset object
+- Create the new Velero server deployment from the modified deployment object
```
shubham-pampattiwar, line 42: @Lyndon-Li Once the new Velero deployment is created and is in running state, should we also check up on the state of prior BSLs that were already configured, whether they have available status or not post upgrade completion ?

design/velero-upgrade.md
```diff
@@ -0,0 +1,85 @@
...
+We want to add a new command besides ```velero install``` called ```velero upgrade```. In a upgrade case, users could call ```velero upgrade``` without bothering to know the details. This bring below benefits:  
+- Solve the upgrade problems in v1.10
+- Make Velero more user friendly, because the upgrade process is significantly simplified
+- Support future evolvements. For example, in future, Velero will have various data movement features, for which we will add new moudles inevitably. Then we can add enhancement incrementally to ```velero upgrade```, because inside the command, we can do whatever complex things 
+
+The ```velero upgrade``` command will fully replace the existing manual steps in the aforementioned Velero document.
+
+## Design
+
+At present, ```velero upgrade``` will do below things:
```
shubham-pampattiwar, line 31: @Lyndon-Li So when the user runs `velero upgrade` path command, would the command show all the upgrade paths feasible/available from the currently installed Velero server to the newer ones ? and then the upgrade path gets executed based on the user's choice ?

design/velero-upgrade.md
Lyndon-Li: This is related to this [discussion](https://github.com/vmware-tanzu/velero/pull/5445#discussion_r996663794) , that is, whether we update the CRs during `velero upgrade`:

- IMO, we don't need to update CRs, we should use kubernetes built in approaches to manage CRs/CR versions. If so, since we don't change CRs, we don't need to verify their status after upgrade
- If we do need to update CRs, then we need to verify the status for the updated CRs

Let's wait and see the result of the [discussion](https://github.com/vmware-tanzu/velero/pull/5445#discussion_r996663794) and then decide the verification accordingly.

design/velero-upgrade.md
Lyndon-Li: This first depends on whether we have multiple upgrade paths from one version to the other, see this [discussion](https://github.com/vmware-tanzu/velero/pull/5445#discussion_r995256744).

Secondly, even though we have multiple paths, I am not quite sure that there is any necessity for users to choose the paths.
The reason is, the action functions are internal operations, the final result of the action table execution is we get velero deployment/daemonset resources that match the target version and then we apply the new resources to users environment. That is to say, users don't see the changes of the environment from one version to the other version sequentially, so users see exactly the same result no matter which path the action table is executed with.

@shubham-pampattiwar How do you think about this? Any specific reason do you think that we should allow users to choose the paths?

PR #5459, opened 2022-10-18
Upgrade velero docker image to use go 1.18

Lyndon-Li:
Upgrade velero docker image to use go 1.18 and upgrade golangci-lint to 1.45.0


blackpiglet (approved):

pkg/cmd/cli/backup/describe.go
```diff
@@ -73,7 +73,7 @@ func NewDescribeCommand(f client.Factory, use string) *cobra.Command {
 			}
 
 			first := true
-			for _, backup := range backups.Items {
+			for i, backup := range backups.Items {
```
reasonerjt, line 76: It's unbelievable lint reports an error for this.

Lyndon-Li: Technically speaking, this indeed causes problem if we pass the loop variable point to a closure of any ansync operation, so I kept this rule and made the change in code.

reasonerjt (approved):

Issue #5477, opened 2022-10-24
Create a common way to support s3 compatible object storage for Restic and Kopia


Lyndon-Li:
At present for Pod Volume Backup, for the known object storage provides, i.e. AWS, Azure, GCP, Velero creates the storage access URL in its own code, for example, for AWS, Velero creates the URL as `s3:s3-<region>.amazonaws.com/<bucket name>`. For other S3 compatible storages, Velero requires users to provide a `resticRepoPrefix` parameter, which contains a similar URL to access the storage. The value of `resticRepoPrefix` is provided to Restic directly.

Inside Restic, it sees no difference for the two ways, URL created by Velero or specified by users, it will break the URL down to separate parameters, i.e., endpoint, region, bucket.

Now we see two issues:
1. Kopia doesn't support the second way, that is we cannot let users provide a URL like `resticRepoPrefix` and pass it to Kopia; moreover, the "s3:" prefix is for Restic only, Kopia doesn't recognize it.
2. Actually, in BSL, we can provide the storage parameters separately in the `Config` map (`BackupStorageLocation.Spec.Config`), and the parameters are also respected in Velero code. Consequently, there are a redundant information, or in another word, theoretically, we can remove `resticRepoPrefix` and use the separate parameters in `Config` map only

Therefore, we should create a common solution to support the various s3 compatible object storage that work for Restic and Kopia.

Lyndon-Li:
fixed by #5478

PR #5478, opened 2022-10-24
Issue fix 5477

Lyndon-Li:
Issue fix 5477: create the common way to support S3 compatible object storages that work for both Restic and Kopia; Keep the `resticRepoPrefix` parameter for compatibility

fix issue #5477


Lyndon-Li:
@sseago @shubham-pampattiwar 

Please help to review this PR. This involves a behavior change - technically we want to remove the `resticRepoPrefix` parameter in BSL's `spec.Config` map because it works for Restic only; meanwhile, we just keep it there for compatibility.

qiuming-best (approved):

blackpiglet (approved):

PR #5532, opened 2022-11-01
Propose a deprecation process for velero

weshayutin:
As discussed in the velero community call [1]
This is a proposed deprecation policy for the
velero project based on the goharbor project.

[1] https://hackmd.io/Jq6F5zqZR7S80CeDWUklkA

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -107,6 +107,29 @@ Lazy consensus does _not_ apply to the process of:
 
 * Removal of maintainers from Velero
 
+## Deprecation Policy
+
+### Deprecation Process
+
+Any contributor may introduce a request to deprecate a feature or an option of a feature by opening a feature request issue in the vmware-tanzu/velero GitHub project. The issue should describe why the feature is no longer needed or has become detrimental to Velero, as well as whether and how it has been superseded. The submitter should give as much detail as possible.
+
+Once the issue is filed, a discussion period begins which ends at the beginning of the second community meeting after the opening of the issue, and is held at the community meeting and on the issue itself; the person who opens the issue or a maintainer should add that date and time in a comment in the issue as soon after the issue is opened as possible.
+
+The feature will be deprecated by a supermajority vote of 50% plus one of the project maintainers at the time of the vote tallying, which is 72 hours after the end of the community meeting that is the end of the comment period. (Maintainers are permitted to vote in advance of the deadline, but should hold their votes until as close as possible to hear all possible discussion.) Votes will be tallied in comments on the issue. 
+
+Non-maintainers may add non-binding votes in comments to the issue as well; these are opinions to be taken into consideration by maintainers, but they do not count as votes. 
+
+If the vote passes, the deprecation window takes effect in the subsequent release, and the removal follows the schedule. 
+
+### Schedule
+If depreciation proposal passes by supermajority votes, the feature is deprecated in the next minor release and the feature can be removed completely after two minor version or equivalent major version e.g., if feature gets deprecated in Nth minor version, then feature can be removed after N+2 minor version or its equivalent if the major version number changes.
+
+### Deprecation Window
+
+The deprecation window is the period from the release in which the deprecation takes effect through the release in which the feature is removed. During this period, only critical security vulnerabilities and catastrophic bugs should be fixed.
+
+**Note:** If a deprecated feature must remain available for general use for two releases after deprecation, perhaps a backup that uses this deprecated feature must be fully-supported for restore for four releases.
+
 ## Updating Governance
 
 All substantive changes in Governance require a supermajority agreement by all maintainers.
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -113,7 +113,7 @@ Lazy consensus does _not_ apply to the process of:
 
 Any contributor may introduce a request to deprecate a feature or an option of a feature by opening a feature request issue in the vmware-tanzu/velero GitHub project. The issue should describe why the feature is no longer needed or has become detrimental to Velero, as well as whether and how it has been superseded. The submitter should give as much detail as possible.
 
-Once the issue is filed, a discussion period begins which ends at the beginning of the second community meeting after the opening of the issue, and is held at the community meeting and on the issue itself; the person who opens the issue or a maintainer should add that date and time in a comment in the issue as soon after the issue is opened as possible.
+Once the issue is filed, a one-month discussion period begins. Discussions take place within the issue itself as well as in the community meetings. The person who opens the issue, or a maintainer, should add the date and time marking the end of the discussion period in a comment on the issue as soon as possible after it is opened. A decision on the issue needs to be made within this one-month period.
 
 The feature will be deprecated by a supermajority vote of 50% plus one of the project maintainers at the time of the vote tallying, which is 72 hours after the end of the community meeting that is the end of the comment period. (Maintainers are permitted to vote in advance of the deadline, but should hold their votes until as close as possible to hear all possible discussion.) Votes will be tallied in comments on the issue. 
 
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -128,7 +128,7 @@ If depreciation proposal passes by supermajority votes, the feature is deprecate
 
 The deprecation window is the period from the release in which the deprecation takes effect through the release in which the feature is removed. During this period, only critical security vulnerabilities and catastrophic bugs should be fixed.
 
-**Note:** If a deprecated feature must remain available for general use for two releases after deprecation, perhaps a backup that uses this deprecated feature must be fully-supported for restore for four releases.
+**Note:** If a backup relies on a deprecated feature, then backups made with the last Velero release before this feature is removed must still be restorable in version `n+2`. For instance, something like restic feature support, that might mean that restic is removed from the list of supported uploader types in version `n` but the underlying implementation required to restore from a restic backup won't be removed until release `n+2`.
 
 ## Updating Governance
 
--- a/changelogs/unreleased/5532-shubham-pampattiwar
+++ b/changelogs/unreleased/5532-shubham-pampattiwar
@@ -1,0 +1,1 @@
+Propose a deprecation process for velero


GOVERNANCE.md
```diff
@@ -107,6 +107,28 @@ Lazy consensus does _not_ apply to the process of:
...
+
+The feature will be deprecated by a supermajority vote of 50% plus one of the project maintainers at the time of the vote tallying, which is 72 hours after the end of the community meeting that is the end of the comment period. (Maintainers are permitted to vote in advance of the deadline, but should hold their votes until as close as possible to hear all possible discussion.) Votes will be tallied in comments on the issue. 
+
+Non-maintainers may add non-binding votes in comments to the issue as well; these are opinions to be taken into consideration by maintainers, but they do not  count as votes. 
+
+If the vote passes, the deprecation window takes effect in the subsequent release, and the removal follows the schedule. 
+
+### Schedule
+Due to the fact that there are no timed releases for the velero project a feature must remain in the deprecation window for at least [six months, one year?] prior to removal. Once the deprecation window has expired the feature may be removed in the following release.
```
reasonerjt, line 125: as for `deprecation window` I would suggest we measure it by minor release.

Currently the approx. release cadence as 3 minor releases a year.
So probably we can say the deprecation window should be at least one minor release?

OrlinVasilev, line 125: @reasonerjt @weshayutin something like that?
```suggestion
If depreciation proposal passes by supermajority votes, the feature is deprecated in the next minor release and the feature can be removed completely after two minor version or equivalent major version e.g., if feature gets deprecated in Nth minor version, then feature can be removed after N+2 minor version or its equivalent if the major version number changes.
```

sseago:
Not really stale. This needs to be resolved.

GOVERNANCE.md
reasonerjt: lgtm

OrlinVasilev (approved):
LGTM

GOVERNANCE.md
```diff
@@ -107,6 +107,28 @@ Lazy consensus does _not_ apply to the process of:
 
 * Removal of maintainers from Velero
 
+## Deprecation Policy
+
+### Deprecation Process
+
+Any contributor may introduce a request to deprecate a feature or an option of a feature by opening a feature request issue in the vmware-tanzu/velero GitHub project. The issue should describe why the feature is no longer needed or has become detrimental to velero, as well as whether and how it has been superseded. The submitter should give as much detail as possible.
```
ihcsim, line 114:
```suggestion
Any contributor may introduce a request to deprecate a feature or an option of a feature by opening a feature request issue in the vmware-tanzu/velero GitHub project. The issue should describe why the feature is no longer needed or has become detrimental to Velero, as well as whether and how it has been superseded. The submitter should give as much detail as possible.
```

GOVERNANCE.md
```diff
@@ -107,6 +107,28 @@ Lazy consensus does _not_ apply to the process of:
...
+
+Once the issue is filed, a discussion period begins which ends at the beginning of the second community meeting after the opening of the issue, and is held at the community meeting and on the issue itself; the person who opens the issue or a maintainer should add that date and time in a comment in the issue as soon after the issue is opened as possible.
+
+The feature will be deprecated by a supermajority vote of 50% plus one of the project maintainers at the time of the vote tallying, which is 72 hours after the end of the community meeting that is the end of the comment period. (Maintainers are permitted to vote in advance of the deadline, but should hold their votes until as close as possible to hear all possible discussion.) Votes will be tallied in comments on the issue. 
+
+Non-maintainers may add non-binding votes in comments to the issue as well; these are opinions to be taken into consideration by maintainers, but they do not  count as votes.
```
ihcsim, line 120:
```suggestion
Non-maintainers may add non-binding votes in comments to the issue as well; these are opinions to be taken into consideration by maintainers, but they do not count as votes. 
```

GOVERNANCE.md
```diff
@@ -107,6 +107,28 @@ Lazy consensus does _not_ apply to the process of:
...
+
+Once the issue is filed, a discussion period begins which ends at the beginning of the second community meeting after the opening of the issue, and is held at the community meeting and on the issue itself; the person who opens the issue or a maintainer should add that date and time in a comment in the issue as soon after the issue is opened as possible.
+
+The feature will be deprecated by a supermajority vote of 50% plus one of the project maintainers at the time of the vote tallying, which is 72 hours after the end of the community meeting that is the end of the comment period. (Maintainers are permitted to vote in advance of the deadline, but should hold their votes until as close as possible to hear all possible discussion.) Votes will be tallied in comments on the issue. 
+
+Non-maintainers may add non-binding votes in comments to the issue as well; these are opinions to be taken into consideration by maintainers, but they do not  count as votes. 
+
+If the vote passes, the deprecation window takes effect in the subsequent release, and the removal follows the schedule.
```
ihcsim, line 122: Will there be additional announcements made via the mailing list, in the change log etc.?

GOVERNANCE.md
```diff
@@ -107,6 +107,28 @@ Lazy consensus does _not_ apply to the process of:
 
 * Removal of maintainers from Velero
 
+## Deprecation Policy
+
+### Deprecation Process
+
+Any contributor may introduce a request to deprecate a feature or an option of a feature by opening a feature request issue in the vmware-tanzu/velero GitHub project. The issue should describe why the feature is no longer needed or has become detrimental to velero, as well as whether and how it has been superseded. The submitter should give as much detail as possible.
+
+Once the issue is filed, a discussion period begins which ends at the beginning of the second community meeting after the opening of the issue, and is held at the community meeting and on the issue itself; the person who opens the issue or a maintainer should add that date and time in a comment in the issue as soon after the issue is opened as possible.
```
ihcsim, line 116: I don't think this discussion period of two community meetings should be a hard-and-fast rule, as the deprecation/removal of some features may be more involved than others. In particular, there may be cases where we need to think through the upgrade path for the community.

sseago (dismissed):

GOVERNANCE.md
reasonerjt: Good point, I think we can make sure it's clearly documented in the release notes.
cc @weshayutin 
thoughts?

yanji09:
In today's community, folks discussed to merge this PR and get the deprecation process in place. Found there are a few remaining comments. 
@weshayutin would you help to address the remaining comments from @ihcsim?

GOVERNANCE.md
OrlinVasilev: We have to introduce some hard lines otherwise it can take forever to be reviewed. Two community  meetings are 2 weeks so maybe that's a bit short indeed, maybe we can word it as "Two consecutive meetings in the same timezone" which will make it a month and that's in total 4 meetings

GOVERNANCE.md
OrlinVasilev: Agree we can list them here: slack, mailing list, changelog but I think that's valid by default for any removal/deprecation that we do

GOVERNANCE.md
sseago: @OrlinVasilev It might be simpler to just set the time to "one month" -- in practice, that means 2 meetings per timezone, but it's easier to understand a month vs. "2 meetings in one time zone".

OrlinVasilev: Agree as I said month is easier also we can track it with github time stamps

GOVERNANCE.md
```diff
@@ -107,6 +107,28 @@ Lazy consensus does _not_ apply to the process of:
 
 * Removal of maintainers from Velero
 
+## Deprecation Policy
+
+### Deprecation Process
+
+Any contributor may introduce a request to deprecate a feature or an option of a feature by opening a feature request issue in the vmware-tanzu/velero GitHub project. The issue should describe why the feature is no longer needed or has become detrimental to Velero, as well as whether and how it has been superseded. The submitter should give as much detail as possible.
```
Lyndon-Li, line 114: For Velero, on one hand, it has a feature that may be represented by code, modules or interfaces; on the other hand, it has the permanent data that created by the feature.
If the feature means all the code/modules/interfaces, once the feature is removed, the data created by it will not be accessible.
However, as a backup/restore tool, the backed up data are usually demanded for long-time preservation, i.e., years or decades.

Therefore, for a feature that generates backup data with a unique format, for example:

- A uploader or backup repository like Restic
- A encryption/compression algorithm for backup

along with the deprecation of a feature, should we clarify the end-of-support policy for the backup data?

sseago: @Lyndon-Li That's a good point. It may be that for something like restic, as one example, we need to keep restore support longer than backup support. For example, if we deprecate restic in 1.13, we could remove backup support in 1.15, but we may need to keep restore support until 1.17. In other words, Velero 1.14 would still allow new backups to use restic, and restores with restic, but in Velero 1.15 or 1.16, new backups will not be able to use restic, but a velero 1.12 or 1.13 backup with restic data could still be restored.

Or in more general terms, if a deprecated feature must remain available for general use for two releases after deprecation, perhaps a backup that uses this deprecated feature must be fully-supported for restore for four releases.

Lyndon-Li: @weshayutin 
Could you help to include this part in the PR?
For the details, personal I agree with @sseago's proposal above.

cc @reasonerjt @yanji09

shubham-pampattiwar: @Lyndon-Li Done, added a note regarding this.

shubham-pampattiwar:
I have update the PR, please take another look @reasonerjt @sseago @Lyndon-Li @weshayutin @ihcsim @OrlinVasilev @yanji09

GOVERNANCE.md
weshayutin: Thank you! @shubham-pampattiwar

weshayutin:
it's a good start
/lgtm

GOVERNANCE.md
```diff
@@ -107,6 +107,29 @@ Lazy consensus does _not_ apply to the process of:
...
+If the vote passes, the deprecation window takes effect in the subsequent release, and the removal follows the schedule. 
+
+### Schedule
+If depreciation proposal passes by supermajority votes, the feature is deprecated in the next minor release and the feature can be removed completely after two minor version or equivalent major version e.g., if feature gets deprecated in Nth minor version, then feature can be removed after N+2 minor version or its equivalent if the major version number changes.
+
+### Deprecation Window
+
+The deprecation window is the period from the release in which the deprecation takes effect through the release in which the feature is removed. During this period, only critical security vulnerabilities and catastrophic bugs should be fixed.
+
+**Note:** If a deprecated feature must remain available for general use for two releases after deprecation, perhaps a backup that uses this deprecated feature must be fully-supported for restore for four releases.
```
kaovilai, line 131: "must be fully supported for restore for four releases" perhaps we want to keep the code for backup but only allow it for generating and running test cases, not general use by end user. 

Running more than one version to test would be ideal but more difficult imo especially in unit testing.

sseago (approved):

GOVERNANCE.md
renmaosheng: what use case does the "*Note:**" section try to solve? thanks.

Lyndon-Li: @renmaosheng 
See [this comment ](https://github.com/vmware-tanzu/velero/pull/5532#discussion_r1170938898), if a feature is essential to the backup which for example, impacts the backed up data, when we deprecate the feature, we should guarantee that users are still able to access their backups, e.g., restore from them.
Here Restic deprecation is right an example, users may have some previous backups created by Restic, and they may want to preserve the backups for quite long time, remaining able to restore from them.

sseago: Restic is the main example we had in mind. If we deprecate Restic in 1.13, then restic must be available for *new* backups for N+2 releases, but prior restic backups must be available for restore for longer. Otherwise, if we removed Restic completely in 1.15 (rather than just on the backup side), this would mean that a user could make a Restic backup in 1.14, upgrade to 1.15 and that "one release old" backup would no longer be restorable.

reasonerjt: I also find the sentence 
```
perhaps a backup that uses this deprecated feature must be fully-supported for restore for four releases.
```
is a little confusing.

How about we just explicitly mention that a rule of thumb for velero is to be able to restore a backup that is created in version `n-2`, and rule must not be violated when a deprecated feature is removed from velero?

reasonerjt: cc @weshayutin

GOVERNANCE.md
reasonerjt: @weshayutin will you be able to make this change?

GOVERNANCE.md
reasonerjt: Given this is default I don't think we need to mention it here in the doc.

GOVERNANCE.md
sseago: @reasonerjt @weshayutin Something like this?
"If a backup relies on a deprecated feature, then backups made with the last Velero release before this feature is removed must still be restorable in version `n+2`. For something like restic support, that might mean that restic is removed from the list of supported uploader types in version `n` but the underlying implementation required to restore from a restic backup won't be removed until release `n+2`"

shubham-pampattiwar: +1 @sseago 
@reasonerjt @weshayutin I can update the PR if we agree upon what @sseago suggested here.

weshayutin: ++ I agree w/ what @sseago said there

reasonerjt: @shubham-pampattiwar 
Sorry for the late response, the proposed wording by @sseago looks good to me.

shubham-pampattiwar: @reasonerjt @weshayutin @sseago Done, updated the PR.

kaovilai (approved):

GOVERNANCE.md
shubham-pampattiwar: Done, updated

blackpiglet (approved):

Lyndon-Li (approved):

PR #5520, opened 2022-10-31
Change image repository  (#1)

wenterjoy:
* change image repository by configmap

* change plugin framework to common

* add changelogs

* add comments to changlog
Signed-off-by: wenterjoy <wenterjoy@126.com>

Thank you for contributing to Velero!

# Please add a summary of your change

add new RestoreItemAction of "velero.io/change-image-repository" to handle the issue mentioned at #5519
"considering the local image repositories changes when migrating from cluster A to B"

# Does your change fix a particular issue?
#5519
Fixes #(issue)
#5519
# Please indicate you've done the following:

- [x] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [x] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required` as a comment on this pull request.
- [ ] Updated the corresponding documentation in `site/content/docs/main`.

Issue #5519, opened 2022-10-31
considering the local image repositories changes when migrating from cluster A to B


wenterjoy:
**Describe the problem/challenge you have**
[A description of the current limitation/problem/challenge that you are experiencing.]
Cluster A has a private image repository that most pod images depends on is directly named as a LAN ip or private domain name.
So all the images is showed as [LAN ip:port]/[image-name]:[tag], which is commonly in test、development and even production environment.
I met the problem that most of the images are ImagePullBackOff when migrating from Cluster A in Cluster B. Because the images are all pointed to the image repository of Cluster A can't pulled because the network is unreachable while the new private image repository is supplied with new LAN ip or private domain info with Cluster B. I needed to change all the pod's image repository prefix from A to B.

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
It should be an chance to supply a config-map that indicates the image repository can be changed from [old] to [new], when migrating from Cluster A in Cluster B, so all the images meets the (config-map) rule can be changed automatically when restore the cluster.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
Inspired by the storage class changing action， I have written a plugin that can change the deployment/daemonset/statefulset/pod's initcontainers and containers' image repository.

**Environment:**

- Velero version (use `velero version`):

- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

wenterjoy:
> Note that this can currently be done outside of velero by writing a RestoreItemAction plugin. If we were to decide to do it inside of Velero, the implementation would probably still be implemented via a RestoreItemAction.
> 
> Also note that if doing this, it's not just Pod images that will need to be modified. The same substitution will need to be applied to any resource that contains a Pod template. This would include Deployment, ReplicaSet, ReplicationController, DaemonSet, StatefulSet, Job, CronJob.

[...]

PR #5540, opened 2022-11-03
change image repository by configmap

wenterjoy:
Signed-off-by: wenterjoy <zhaoxin26@lenovo.com>

add new RestoreItemAction of "velero.io/change-image-repository" to handle the issue mentioned at https://github.com/vmware-tanzu/velero/issues/5519
"considering the local image repositories changes when migrating from cluster A to B"

#5519
Fixes #(issue)
#5519

### DIFF
--- a/pkg/restore/change_image_name_action.go
+++ b/pkg/restore/change_image_name_action.go
@@ -1,0 +1,213 @@
+/*
+Copyright 2022 the Velero contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package restore
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/vmware-tanzu/velero/pkg/plugin/framework/common"
+	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+)
+
+const (
+	DELIMITER_VALUE = ","
+)
+
+// ChangeImageNameAction updates a deployment or Pod's image name
+// if a mapping is found in the plugin's config map.
+type ChangeImageNameAction struct {
+	logger          logrus.FieldLogger
+	configMapClient corev1client.ConfigMapInterface
+}
+
+// NewChangeImageNameAction is the constructor for ChangeImageNameAction.
+func NewChangeImageNameAction(
+	logger logrus.FieldLogger,
+	configMapClient corev1client.ConfigMapInterface,
+) *ChangeImageNameAction {
+	return &ChangeImageNameAction{
+		logger:          logger,
+		configMapClient: configMapClient,
+	}
+}
+
+// AppliesTo returns the resources that ChangeImageNameAction should
+// be run for.
+func (a *ChangeImageNameAction) AppliesTo() (velero.ResourceSelector, error) {
+	return velero.ResourceSelector{
+		IncludedResources: []string{"deployments", "statefulsets", "daemonsets", "replicasets", "replicationcontrollers", "jobs", "cronjobs", "pods"},
+	}, nil
+}
+
+// Execute updates the item's spec.containers' image if a mapping is found
+// in the config map for the plugin.
+func (a *ChangeImageNameAction) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
+	a.logger.Info("Executing ChangeImageNameAction")
+	defer a.logger.Info("Done executing ChangeImageNameAction")
+
+	opts := metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("velero.io/plugin-config,%s=%s", "velero.io/change-image-name", common.PluginKindRestoreItemAction),
+	}
+
+	list, err := a.configMapClient.List(context.TODO(), opts)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	if len(list.Items) == 0 {
+		return &velero.RestoreItemActionExecuteOutput{
+			UpdatedItem: input.Item,
+		}, nil
+	}
+
+	if len(list.Items) > 1 {
+		var items []string
+		for _, item := range list.Items {
+			items = append(items, item.Name)
+		}
+		return nil, errors.Errorf("found more than one ConfigMap matching label selector %q: %v", opts.LabelSelector, items)
+	}
+
+	config := &list.Items[0]
+	if len(config.Data) == 0 {
+		a.logger.Info("No image name mappings found")
+		return velero.NewRestoreItemActionExecuteOutput(input.Item), nil
+	}
+
+	obj, ok := input.Item.(*unstructured.Unstructured)
+	if !ok {
+		return nil, errors.Errorf("object was of unexpected type %T", input.Item)
+	}
+	if obj.GetKind() == "Pod" {
+
+		err = a.replaceImageName(obj, config, "spec", "containers")
+		if err != nil {
+			a.logger.Infof("replace image name meet error: %v", err)
+			return nil, errors.Wrap(err, "error getting item's spec.containers")
+		}
+
+		err = a.replaceImageName(obj, config, "spec", "initContainers")
+		if err != nil {
+			a.logger.Infof("replace image name meet error: %v", err)
+			return nil, errors.Wrap(err, "error getting item's spec.containers")
+		}
+
+	} else if obj.GetKind() == "CronJob" {
+		//handle containers
+		err = a.replaceImageName(obj, config, "spec", "jobTemplate", "spec", "template", "spec", "containers")
+		if err != nil {
+			a.logger.Infof("replace image name meet error: %v", err)
+			return nil, errors.Wrap(err, "error getting item's spec.containers")
+		}
+		//handle initContainers
+		err = a.replaceImageName(obj, config, "spec", "jobTemplate", "spec", "template", "spec", "initContainers")
+		if err != nil {
+			a.logger.Infof("replace image name meet error: %v", err)
+			return nil, errors.Wrap(err, "error getting item's spec.containers")
+		}
+
+	} else {
+		//handle containers
+		err = a.replaceImageName(obj, config, "spec", "template", "spec", "containers")
+		if err != nil {
+			a.logger.Infof("replace image name meet error: %v", err)
+			return nil, errors.Wrap(err, "error getting item's spec.containers")
+		}
+
+		//handle initContainers
+		err = a.replaceImageName(obj, config, "spec", "template", "spec", "initContainers")
+		if err != nil {
+			a.logger.Infof("replace image name meet error: %v", err)
+			return nil, errors.Wrap(err, "error getting item's spec.containers")
+		}
+	}
+	return velero.NewRestoreItemActionExecuteOutput(obj), nil
+}
+
+func (a *ChangeImageNameAction) replaceImageName(obj *unstructured.Unstructured, config *corev1.ConfigMap, filed ...string) error {
+
+	log := a.logger.WithFields(map[string]interface{}{
+		"kind":      obj.GetKind(),
+		"namespace": obj.GetNamespace(),
+		"name":      obj.GetName(),
+	})
+	needUpdateObj := false
+	containers, _, err := unstructured.NestedSlice(obj.UnstructuredContent(), filed...)
+	if err != nil {
+		a.logger.Infof("UnstructuredConverter meet error: %v", err)
+		return errors.Wrap(err, "error getting item's spec.containers")
+	}
+	if len(containers) == 0 {
+		return nil
+	}
+	for i, container := range containers {
+		a.logger.Infoln("container:", container)
+		if image, ok := container.(map[string]interface{})["image"]; ok {
+			imageName := image.(string)
+			if exists, newImageName, err := a.isImageReplaceRuleExist(log, imageName, config); exists && err == nil {
+				needUpdateObj = true
+				a.logger.Infof("Updating item's image from %s to %s", imageName, newImageName)
+				container.(map[string]interface{})["image"] = newImageName
+				containers[i] = container
+			}
+		}
+	}
+	if needUpdateObj {
+		if err := unstructured.SetNestedField(obj.UnstructuredContent(), containers, filed...); err != nil {
+			return errors.Wrap(err, "unable to set item's initContainer image")
+		}
+	}
+	return nil
+}
+
+func (a *ChangeImageNameAction) isImageReplaceRuleExist(log *logrus.Entry, oldImageName string, cm *corev1.ConfigMap) (exists bool, newImageName string, err error) {
+	if oldImageName == "" {
+		log.Infoln("Item has no old image name specified")
+		return false, "", nil
+	}
+	log.Debug("oldImageName: ", oldImageName)
+
+	//how to use: "<old_image_name_sub_part><delimiter><new_image_name_sub_part>"
+	//for current implementation the <delimiter> value can only be ","
+	//e.x: in case your old image name is 1.1.1.1:5000/abc:test
+	//"case1":"1.1.1.1:5000,2.2.2.2:3000"
+	//"case2":"5000,3000"
+	//"case3":"abc:test,edf:test"
+	//"case4":"1.1.1.1:5000/abc:test,2.2.2.2:3000/edf:test"
+	for _, row := range cm.Data {
+		if !strings.Contains(row, DELIMITER_VALUE) {
+			continue
+		}
+		if strings.Contains(oldImageName, strings.TrimSpace(row[0:strings.Index(row, DELIMITER_VALUE)])) && len(row[strings.Index(row, DELIMITER_VALUE):]) > len(DELIMITER_VALUE) {
+			log.Infoln("match specific case:", row)
+			oldImagePart := strings.TrimSpace(row[0:strings.Index(row, DELIMITER_VALUE)])
+			newImagePart := strings.TrimSpace(row[strings.Index(row, DELIMITER_VALUE)+len(DELIMITER_VALUE):])
+			newImageName = strings.Replace(oldImageName, oldImagePart, newImagePart, -1)
+			return true, newImageName, nil
+		}
+	}
+	return false, "", errors.Errorf("No mapping rule found for image: %s", oldImageName)
+}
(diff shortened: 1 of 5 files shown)


wenterjoy:
@blackpiglet  Hi, I have created this new function PR and closed last one. I'm not sure this is what you want, 
If yes, please help to review and change the other reviewer to @sseago 

@sseago  I added  the ReplicaSet, ReplicationController, Job, and CronJob and updated the doc "site/content/docs/main/restore-reference.md " as your suggestion. Please help to review it.

Thanks again for both of your help on my first commit.

wenterjoy:
@a-mccarthy   Hi, I had one PR #5520 that @sseago  and @blackpiglet  already give comments to "Velero v1.10 is under testing for RC.May need to put new function PR on hole, before release branch cut."  So I  closed #5520 and create this one.
Please help to review the doc part. Thanks a lot.

sseago:
Note that we're getting close to 1.10 release branch creation, so regardless of review status we'll want to hold off on merging this until the release-1.10 branch is created. At this point, we're at feature complete, so we're only pulling bugfixes in.

wenterjoy:
> s created. At this point, we're at feature complete, so we're only pulling bugfixes in.

thanks for reply.

pkg/restore/change_image_repository_action.go
```diff
@@ -0,0 +1,267 @@
...
+		logger:          logger,
+		configMapClient: configMapClient,
+	}
+}
+
+// AppliesTo returns the resources that ChangeImageRepositoryAction should
+// be run for.
+func (a *ChangeImageRepositoryAction) AppliesTo() (velero.ResourceSelector, error) {
+	return velero.ResourceSelector{
+		IncludedResources: []string{"deployment", "statefulsets", "daemonset", "replicaset", "replicationcontroller", "job", "CronJob", "pod"},
```
sseago, line 59: These should be listed in a consistent format. I'm not sure it would affect behavior, but it would be more readable that way:
```
		IncludedResources: []string{"deployments", "statefulsets", "daemonsets", "replicasets", "replicationcontrollers", "jobs", "cronjobs", "pods"},
￼Write
```

pkg/restore/change_image_repository_action.go
```diff
@@ -0,0 +1,267 @@
...
+	a.logger.Info("Executing ChangeImageRepositoryAction")
+	defer a.logger.Info("Done executing ChangeImageRepositoryAction")
+
+	config, err := getPluginConfig(common.PluginKindRestoreItemAction, "velero.io/change-image-repository", a.configMapClient)
+	if err != nil {
+		return nil, err
+	}
+
+	if config == nil || len(config.Data) == 0 {
+		a.logger.Info("No image image mappings found")
```
sseago, line 75: I think that should probably be "No image repository mappings found"

site/content/docs/main/restore-reference.md
```diff
@@ -169,6 +169,39 @@ data:
...
+    # config for a plugin (i.e. the built-in restore item action plugin)
+    velero.io/plugin-config: ""
+    # this label identifies the name and kind of plugin
+    # that this ConfigMap is for.
+    velero.io/change-image-repository: RestoreItemAction
+data:
+  # add 1+ key-value pairs here, where the key is the old
+  # image repository name and the value is the new image
+  # repository name.
+  <old-image-repository>: <[new_image_repository>
```
sseago, line 198: There's an extra "[" char before "new_image_repository"

site/content/docs/main/restore-reference.md
```diff
@@ -169,6 +169,39 @@ data:
   # class name.
   <old-storage-class>: <new-storage-class>
 ```
+### Changing Pod/Deployment/Statefulsets/Daemonset/Replicaset/Replicationcontroller/Job/Cronjob Image Repositories
```
sseago, line 172: Since you're listing the "kind" value in the heading, this should be:
```
### Changing Pod/Deployment/StatefulSet/DaemonSet/ReplicaSet/ReplicationController/Job/Cronjob Image Repositories  
```

site/content/docs/main/restore-reference.md
```diff
@@ -169,6 +169,39 @@ data:
   # class name.
   <old-storage-class>: <new-storage-class>
 ```
+### Changing Pod/Deployment/Statefulsets/Daemonset/Replicaset/Replicationcontroller/Job/Cronjob Image Repositories  
+
+Velero can change the image repositories of pod/deployment/statefulsets/daemonset/replicaset/replicationcontroller/job/cronjob during restores. To configure a image repositoriy mapping, create a config map in the Velero namespace like the following:
```
sseago, line 174: Everything but statefulsets is in the singular form above, so I'd change that one to "statefulset" to be consistent here.

sseago (changes requested):

site/content/docs/main/restore-reference.md
wenterjoy: @sseago  thanks so much for your careful review.
I will update it according to all the comments.

wenterjoy:
@sseago  Please help to review again, I have modified it according to all your comments. Thanks.

wenterjoy:
> @wenterjoy So I think we need to change the way we handle mappings with port included. The "specific" key is only able to handle a single mapping that contains port numbers. I think we need to allow for multiple mappings of that nature. Also, if we can come up with a way to handle that, we can also handle including directory/namespacing in it -- i.e. to map "1.1.1.1:5000/foo" to "2.2.2.2:5000/bar". If, for example, I want to map from "docker.io/sseago" to "quay.io/myproject" I can't do that with either "specific" as defined here or with normal configmap keys (since "/" is also invalid in a configmap key).
> 
> Off the top of my head, I'm not 100% sure what the answer is, though.

@sseago  I do agree with you，and I have a draft for that：
we can define certain  key such  as “image-delimiter” to let users to decide which delimiter they want to use.
then user can give value to “specific”(or we can rename this as "case1" to "casen" depends on how many cases users want to handle",
 ie:
"image-delimiter": %
 "case1": 1.1.1.1:5000%2.2.2.2:3000
 "case2": docker.io/sseago%quay.io/myproject
 
 or
 "image-delimiter": ":"
  "case1": 1.1.1.1%2.2.2.2
 "case2": docker.io/sseago:quay.io/myproject

sseago:
```
"image-delimiter": ":"
"case1": 1.1.1.1%2.2.2.2
```
I think we can simplify this a bit. I don't think we need a flexible delimiter, we just need a delimiter that isn't expected in a docker image registry/repository path. So instead of ":" or "/", the delimiter could just always be "," maybe?

wenterjoy:
> 



> ```
> "image-delimiter": ":"
> "case1": 1.1.1.1%2.2.2.2
> ```
> 
> I think we can simplify this a bit. I don't think we need a flexible delimiter, we just need a delimiter that isn't expected in a docker image registry/repository path. So instead of ":" or "/", the delimiter could just always be "," maybe?



> ```
> "image-delimiter": ":"
> "case1": 1.1.1.1%2.2.2.2
> ```
> 
> I think we can simplify this a bit. I don't think we need a flexible delimiter, we just need a delimiter that isn't expected in a docker image registry/repository path. So instead of ":" or "/", the delimiter could just always be "," maybe?

@sseago 
The advantage of flexible delimiter is user can freely decide the value, but it may led to the hidden damage when image has same letters as delimiter which user mayn't notice.
since velero is "non-destructive", it better to use the fixed bit.

 I have modified and tested with the fixed delimiter bit.
 since user may input spaces with the value, I also added the trim process, and test passed.
 Please have a review. Thanks.

pkg/restore/change_image_name_action.go
```diff
@@ -0,0 +1,275 @@
...
+	//be used as delimiter
+	//for current implementation the <delimiter> value can only be ","
+	//use "case1": "<old_image_name_sub_part><delimiter><new_image_name_sub_part>"
+	//e.x: in case your old image name is 1.1.1.1:5000/abc:test
+	//     "image-delimiter": ","
+	//     "case1":"1.1.1.1:5000,2.2.2.2:3000"
+	//     "case2":"5000,3000"
+	//     "case3":"abc:test,edf:test"
+	//     "case4":"1.1.1.1:5000/abc:test,2.2.2.2:3000/edf:test"
+	if delimiter, ok := cm.Data[DELIMITER_KEY]; !ok {
```
sseago, line 252: Now that we're only using "," as the delimiter, I think this section is unnecessary. This code errors out if DELIMITER_KEY is unspecified, and then errors out if it's specified to anything other than ",". I think it would be easier if we just removed it entirely and just reference DELIMITER_VALUE rather than delimiter below.

wenterjoy: ok. 
DELIMITER_KEY  related code is deleted. new code test passed.
Thanks for you comments.

pkg/restore/change_image_name_action.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+		if needUpdateObj {
+			newObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(pod)
+			if err != nil {
+				return nil, errors.Wrap(err, "convert obj to Pod failed")
+			}
+			obj.Object = newObj
+		}
+
+	} else if obj.GetKind() == "Job" || obj.GetKind() == "CronJob" {
```
sseago, line 125: This section is for CronJob only -- a CronJob has a field Spec.JobTemplate which contains a JobSpec (so spec/jobTemplate/spec/template/spec... as below). A Job just has the PodTemplateSpec directly at Spec.Template, which is the same as Deployment, etc. "Job" should be removed from this conditional so the Job case will fall through to the final else clause.

sseago (changes requested):

pkg/restore/change_image_name_action.go
wenterjoy: @sseago 
thanks for your careful review.
Modification completed.

sseago (approved):

sseago:
@wenterjoy I'm not quite sure why those checks are failing -- the issues seem to be unrelated to your changes. Maybe you can rebase your PR to current main branch and force push. Could be some out-of-date code somewhere is at fault.

wenterjoy:
> @wenterjoy I'm not quite sure why those checks are failing -- the issues seem to be unrelated to your changes. Maybe you can rebase your PR to current main branch and force push. Could be some out-of-date code somewhere is at fault.

@sseago  thanks for your suggestion.  I tried a rebase  and force push.

wenterjoy:
@sseago  I am confused by these checking fails.
Do I have to pass all these check that the PR can be merged.

sseago:
@wenterjoy Yes they need to pass, since these tests often catch bugs in submitted PRs. But I don't understand the failures here, as the ones I looked at don't seem related to your changes.

@reasonerjt Are there checks/tests that are failing sporadically? The e2e failures look like infra issues, and the linter failure doesn't seem to be related to the code modified/added here.

wenterjoy:
> @wenterjoy Yes they need to pass, since these tests often catch bugs in submitted PRs. But I don't understand the failures here, as the ones I looked at don't seem related to your changes.
> 
> @reasonerjt Are there checks/tests that are failing sporadically? The e2e failures look like infra issues, and the linter failure doesn't seem to be related to the code modified/added here.

@sseago  @reasonerjt 
I am very sorry about mistake as below
`	config, err := getPluginConfig(common.PluginKindRestoreItemAction, "velero.io/change-image-repository", a.configMapClient)
	config, err := getPluginConfig(common.PluginKindRestoreItemAction, "velero.io/change-image-name", a.configMapClient)
	if err != nil {
		return nil, err
	}`

but those e2e failures failed at the step of setting up kind environment indeed has no relation to my codes.

Thanks very much for your help @sseago

sseago:
Looks like we're  down to one failing test. If anyone else knows what's going on with the linter check, I think that's the main obstacle left here.

wenterjoy:
@sseago  after read related code 
I guess the root cause is that function “getPluginConfig“ with parameter “kind” of type “common.PluginKind”， but all the references of this function always use “common.PluginKindRestoreItemAction“
before my code added，there were only two references， so lint check may not treat it as an error.
when my add in, the mechanism of link check has reached the trigger condition， so it reported.
So I changed my code to not call “getPluginConfig“ . Hope this can work.

wenterjoy:
@sseago  please help to trigger the checking tasks and have a review

wenterjoy:
@sseago  Thanks. 
The lint check is passed.  The CI check failed at the step to pull the build image. May be need to trigger it again.

wenterjoy:
@blackpiglet  Hello, this PR is failed at CI check. Please help to check whether it is related to my code.

sseago (changes requested):
The arg reversal referenced in the comment is causing the CI failure. Also, it looks like this PR has the change_image_name_action and change_image_name_action_test files duplicated with different versions in this PR, one ends in .go, and the other in what looks like a commit hash.

pkg/restore/change_image_name_action.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// Execute updates the item's spec.containers' image if a mapping is found
+// in the config map for the plugin.
+func (a *ChangeImageNameAction) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
+	a.logger.Info("Executing ChangeImageNameAction")
+	defer a.logger.Info("Done executing ChangeImageNameAction")
+
+	opts := metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("velero.io/plugin-config,%s=%s", common.PluginKindRestoreItemAction, "velero.io/change-image-name"),
```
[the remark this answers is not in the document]
wenterjoy: @sseago  Thank you for helping me find this problem.

wenterjoy:
This time the two failed checks are both due to fetch package timed out
` > [ 9/18] RUN go install golang.org/x/tools/cmd/goimports@11e9d9cc0042e6bd10337d4d2c3e5d9295508e7d:
#13 58.41 go: golang.org/x/tools/cmd/goimports@11e9d9cc0042e6bd10337d4d2c3e5d9295508e7d: golang.org/x/tools/cmd/goimports@11e9d9cc0042e6bd10337d4d2c3e5d9295508e7d: invalid version: reading https://proxy.golang.org/golang.org/x/tools/cmd/goimports/@v/11e9d9cc0042e6bd10337d4d2c3e5d9295508e7d.info: 404 Not Found
#13 58.41 	server response: not found: fetch timed out`

blackpiglet:
Retry also failed. Maybe this is related to Github environment. Will retry later.

blackpiglet:
There is concern that this change is more suitable to a indepedant plugin repository than Velero repository.
Velero team also talked about this. Looks like a new repository to collect useful plugins together.

wenterjoy:
> Retry also failed. Maybe this is related to Github environment. Will retry later.

@blackpiglet  Thanks for so quick reply and response.

wenterjoy:
@blackpiglet would help to retry these checks. Thanks a lot.

wenterjoy:
@sseago @blackpiglet help to retry these checks. Thanks a lot.

wenterjoy:
@sseago @blackpiglet  very excited to see all the checks passed.
Thanks so much for helps.

sseago:
@wenterjoy one more thing -- could you squash these commits into one commit -- currently there are 78 commits in this PR.

wenterjoy:
@sseago  I tried to squash all the commits into one with “git rebase -i   HEAD~81” and got indeed 225 commits need to squash.
I am not sure It is correct to do that.
Please tell me how to reverse if i'm wrong.

sseago:
So I'm thinking what may have happened is that your local branch wasn't synced with upstream main when you attempted to squash. It looks like you squashed a lot more than just your commits -- instead of a few files, there are over 500. I think we can back this out using the latest commit before the force-push (shown above as `af85a77`). Try the following:
```
git checkout -b newbranch af85a77
git fetch <upstream-remote>
git rebase <upstream-remote>/main
git reset --soft <upstream-remote>/main
git commit
git push --force <pr-remote> newbranch:main
```
where <upstream-remote> is whatever you're calling your remote branch for vmware-tanzu/velero, and <pr-remote> is whatever you're calling your own local fork remote branch (possibly origin?)

Before force-pushing, double-check your commit with `git show` to make sure that it includes all of your changes and only those changes.

wenterjoy:
@sseago 
thanks very much for your so detail guide.  
Merry Christmas and Happy New Year！

sseago (approved):

wenterjoy:
@blackpiglet  hello would you help to approve this PR？

pkg/restore/change_image_name_action.go
```diff
@@ -0,0 +1,280 @@
...
+		return nil, errors.Errorf("object was of unexpected type %T", input.Item)
+	}
+
+	log := a.logger.WithFields(map[string]interface{}{
+		"kind":      obj.GetKind(),
+		"namespace": obj.GetNamespace(),
+		"name":      obj.GetName(),
+	})
+
+	if obj.GetKind() == "Pod" {
```
ywk253100, line 110: Could we have a common function to handle all kinds of resources as they all share the same `Container` list?

pkg/restore/change_image_name_action.go
```diff
@@ -0,0 +1,280 @@
...
+			if err := unstructured.SetNestedField(obj.UnstructuredContent(), initContainers, "spec", "template", "spec", "initContainers"); err != nil {
+				return nil, errors.Wrap(err, "unable to set item's initContainer image")
+			}
+		}
+
+	}
+	return velero.NewRestoreItemActionExecuteOutput(obj), nil
+}
+
+func (a *ChangeImageNameAction) isImageReplaceRuleExist(log *logrus.Entry, oldImageName string, cm *corev1.ConfigMap) (exists bool, newImageName string, err error) {
```
ywk253100, line 253: Could we simplify the replace rule? The logic of this function is complicated and hard to understand.
And seems it doesn't handle all cases correctly, e.g.
The original image is `dev/image1:dev` `dev/image2:dev` and I want to replace the image with `test/image1:dev` `test/image2:dev`. If I use `dev,test` as the replace rule, I will get `test/image1:test` `test/image2:test`?

OMI, how about that we only support replacing the prefix of the image? This should cover most requirements and is easier.
e.g.
The registry IP address changed: `ip1/xxx` -> `ip2/xxx` with the replace rule `ip1,ip2`
The registry port changed: `ip1:5000` - > `ip1:6000` with the replace rule `ip1:5000,ip1:6000`
The image repository changed: `ip1/repo1/xxx` -> `ip1/repo2/xxx` with the replace rule `ip1/repo1,ip1/repo2`
@wenterjoy @sseago WDYT?

wenterjoy: @ywk253100 
Thanks a lot for your comments and suggestion.
I do agree with you to cover most requirements. 
But if we support more cases, the code logic would be more complex.
e.g. ip1:5000 - > ip1:6000   ip1:tag1 -> ip1:tag2
should we distinguish between port number and tag？
current implement could cover these cases with replacing the keywords, but it indeed not user-friendly. 
e.g. dev/image1:dev -> test/image1:dev   dev/image2:dev -> test/image2:dev 
if users want to only replace first "dev", the replace rule should be  "dev/, test/". 
This will leave users to be responsible for their wants. I not sure it is in line with velero's principles.

pkg/restore/change_image_name_action.go
wenterjoy: I added a common function “replaceImageName()”
thanks.

pkg/restore/change_image_name_action.go
```diff
@@ -0,0 +1,211 @@
...
+		LabelSelector: fmt.Sprintf("velero.io/plugin-config,%s=%s", "velero.io/change-image-name", common.PluginKindRestoreItemAction),
+	}
+
+	list, err := a.configMapClient.List(context.TODO(), opts)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	if len(list.Items) == 0 {
+		return &velero.RestoreItemActionExecuteOutput{}, nil
```
ywk253100, line 81: Should return 
```
&velero.RestoreItemActionExecuteOutput{
	UpdatedItem: input.Item,
}
```
as the original obj will be override by the returned value

ywk253100 (approved):

sseago (approved):

PR #5559, opened 2022-11-07
Remove dependency on gofrs/uuid

Lyndon-Li:
Remove the unnecessary dependency on gofrs/uuid to reduce security risks

### DIFF
--- a/internal/hook/item_hook_handler.go
+++ b/internal/hook/item_hook_handler.go
@@ -22,7 +22,7 @@ import (
 	"strings"
 	"time"
 
-	uuid "github.com/gofrs/uuid"
+	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1api "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


internal/hook/item_hook_handler.go
```diff
@@ -22,7 +22,7 @@ import (
 	"strings"
 	"time"
 
-	uuid "github.com/gofrs/uuid"
+	uuid "github.com/google/uuid"
```
reasonerjt, line 25: not quite sure why we need to rename this.
why not just make it 
"github.com/google/uuid" 
in the import chunk?

Lyndon-Li: They are different library, or two different provider of uuid. I am not sure why `gofrs/uuid` was used either, but as far as my understanding, it is safe to replace it with `google/uuid`

Lyndon-Li: Not sure the history, but IMO, we can rename it

Lyndon-Li: Done

reasonerjt (approved):
lgtm thanks

blackpiglet (approved):

PR #5714, opened 2022-12-21
[1.10] Fix issue 5696

Lyndon-Li:
Fix issue 5696, check if the repo is still openable before running the prune and forget operation, if it is not, try to reconnect the repo

### DIFF
--- a/pkg/repository/provider/unified_repo.go
+++ b/pkg/repository/provider/unified_repo.go
@@ -200,6 +200,37 @@ func (urp *unifiedRepoProvider) PrepareRepo(ctx context.Context, param RepoParam
 	return nil
 }
 
+func (urp *unifiedRepoProvider) BoostRepoConnect(ctx context.Context, param RepoParam) error {
+	log := urp.log.WithFields(logrus.Fields{
+		"BSL name":  param.BackupLocation.Name,
+		"repo name": param.BackupRepo.Name,
+		"repo UID":  param.BackupRepo.UID,
+	})
+
+	log.Debug("Start to boost repo connect")
+
+	repoOption, err := udmrepo.NewRepoOptions(
+		udmrepo.WithPassword(urp, param),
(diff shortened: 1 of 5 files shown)


pkg/repository/provider/unified_repo.go
```diff
@@ -200,6 +200,37 @@ func (urp *unifiedRepoProvider) PrepareRepo(ctx context.Context, param RepoParam
...
+		udmrepo.WithConfigFile(urp.workPath, string(param.BackupRepo.UID)),
+		udmrepo.WithDescription(repoConnectDesc),
+	)
+
+	if err != nil {
+		return errors.Wrap(err, "error to get repo options")
+	}
+
+	bkRepo, err := urp.repoService.Open(ctx, *repoOption)
+	if err == nil {
```
qiuming-best, line 223: BoostRepoConnect which means that we first need to close the repo and then reconnect no matter if it could be connected?

Lyndon-Li: We try to connect the repo only when `urp.repoService.Open` fails, the open is for a test purpose.
There is no reconnect, if `urp.repoService.Open` succeeds, it means the repo has already been connected, so quit at line 228; otherwise, the repo may has not been connected, so call `urp.ConnectToRepo`.
`urp.repoService.Open` may fail for other reasons, this is out of the scope of this function to fix.

blackpiglet (approved):

pkg/repository/provider/unified_repo.go
qiuming-best: first Open the repo, if no error, it will close the repo at L224.  
So the repo will be closed before you do Forget, PruneRepo action?

Lyndon-Li: The call of `urp.repoService.Open` in function `BoostRepoConnect` for test purpose only, outside `BoostRepoConnect`, the caller will open the repo again own their own

qiuming-best (approved):

Issue #5416, opened 2022-09-30
Backup with CSI partiallyFails with a large number of PVCs


eemcmullan:
**What steps did you take and what happened:**
While completing a larger backup with the CSI plugin (1000 PVCs) the backup partiallyFailed with error: 

```
time="2022-09-21T19:36:03Z" level=error msg="fail to recreate VolumeSnapshotContent snapcontent-72e8f17f-bd98-42cf-b039-7d353057e440: fail to retrieve VolumeSnapshotContent snapcontent-72e8f17f-bd98-42cf-b039-7d353057e440 info: timed out waiting for the condition" backup=openshift-adp/backuptest logSource="pkg/controller/backup_controller.go:974"
```

**What did you expect to happen:**
Backup completed

_If you are using earlier versions:_  
Please provide the output of the following commands (Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)
- `kubectl logs deployment/velero -n velero`
- `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
- `velero backup logs <backupname>`
- `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
- `velero restore logs <restorename>`

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

After some investigating, I suspect the issue is here: https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/backup_controller.go#L1014 

A new image was built with this value changed to 5 minutes, and the backup was able to progress further than the initial backup before partially failing (~500 VSCs to ~900 VSCs). 

This image was then updated to a 10 minute timeout, and the backup then completed. This was tested several times to verify. 

The timeout here may need to be configurable. `CSISnapshotTimeout` could possibly be re-used for this case.

**Environment:**

- Velero version (use `velero version`): 
- Velero features (use `velero client config get features`): 
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

[...]

PR #5746, opened 2023-01-06
Make default timeouts configurable

### LINKED ISSUE #5604: Config Value for Default Timeouts
Related issue: https://github.com/vmware-tanzu/velero/issues/5416

There appear to be a few other default timeouts that could cause similar issues as in 5416: creating the backup repo https://github.com/vmware-tanzu/velero/blob/main/pkg/repository/ensurer.go#L108 and waiting for CRDs in restore: https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L863

Apart from these, other timeouts could be introduced in the future. A configurable `timeout` value shared by these timeouts could be created to help prevent further issues. Or it may make more sense to group together related timeouts.

**Vote on this issue!**

eemcmullan:
Signed-off-by: Emily McMullan <emcmulla@redhat.com>

Updated Backup and Schedules to allow for configurable default timeouts, as well as a server config setting.

Fixes #5604 

- [ x] Updated the corresponding documentation in `site/content/docs/main`.


blackpiglet:
I'm not sure about adding a new timeout setting is the best way to resolve the issue.
For the `BackupRepository`, per my understanding, the only time-consuming part is to query the CR's list from kube-apiserver. It seems timeout for this only happens in very special cases.
And for VolumeSnapshot and VolumeSnapshotContent timeout problem, there is another issue #5416 tracing it. It will be addressed in v1.11.

blackpiglet:
It looks like this timeout setting is supposed to be used in multiple places and has different meanings in different places.
I think it's better to have a more generic name other than `resource-timeout`, e.g. `generic-timeout` or `default-timeout`.
@sseago What's your opinion about it?

sseago:
I don't think any timeouts are truly generic. One thing we've tried to do is to minimize the number of different timeout settings -- if there are multiple timeout values that make sense to be configured together, then they should share a value, but obviously some actions are inherently longer than others. The most important thing is to eliminate hard-coded timeouts since that results in failed backups/restores that can't be fixed without rebuilding Velero with new values. If all timeouts are configurable, then the user just has to redeploy the velero server with the value overridden. So if there are several places that use a shared timeout value, we should choose a name that's as specific as we can be while still covering all uses.

sseago:
It looks like this is used for several instances of creating/deleting resources, so I think "resource-timeout" is probably the right name. "generic-timeout" is too general, as it won't give the user any idea what it's used for.

blackpiglet:
By far, this timer is used by deleting snapshots and getting BackupRepository. That's all resource related. Basically, all time-consuming operations are related to a heavy-loaded kube-apiserver. By thinking that way, the naming makes sense.

@Lyndon-Li Could you share some thoughts on this PR?

blackpiglet:
@eemcmullan @sseago 
I found several other hardcode timers, but they are not backup related.
* timer used by restore to wait before CRD is ready: https://github.com/vmware-tanzu/velero/blob/main/pkg/restore/restore.go#L842
* timer used waiting for Pod to ready, before writing data: https://github.com/vmware-tanzu/velero/blob/main/pkg/podvolume/restorer.go#L202
* the Velero OSS signed URL valid time: https://github.com/vmware-tanzu/velero/blob/main/pkg/persistence/object_store.go#L88

The first two seem more related to the `resource-timeout` topic in the PR.

eemcmullan:
@blackpiglet would be happy to address them in this PR, or raise a separate PR for the restore timeouts.

blackpiglet:
@eemcmullan 
Thanks. One thing worth noticing is the above timers are not backup related.
If we want to use the same default timeout setting to address all of them, per backup setting seems not accurate. 
How about putting the setting into a place that can cover a more extensive scope, such as BSL or Velero server CLI parameter?

eemcmullan:
@blackpiglet there is a value added to the Velero server settings in the PR to address one of the hardcoded timeouts. This could be used for all of the default timeouts for backup and restore then, if that sounds like a reasonable plan. @sseago thoughts?

blackpiglet:
I agree with that. It may be a more suitable place to set the default resource timeout.

shubham-pampattiwar:
timeout value at server settings  level sounds reasonable to me.

kaovilai (approved):

eemcmullan:
closing for https://github.com/vmware-tanzu/velero/pull/5926

eemcmullan:
closing for https://github.com/vmware-tanzu/velero/pull/5926

Issue #5035, opened 2022-06-20
Need a way to skip backup of volumes based on type


gsadhani:
**Describe the problem/challenge you have**
When taking a backup, Velero tries to all pod volumes available. Some volumes can be too large or backed up using other solutions. It is desirable to skip backup of volumes based on its type.

**Describe the solution you'd like**
A way to configure the backup to skip certain types of volumes like NFS.

**Anything else you would like to add:**
None

**Environment:**

- Velero version (use `velero version`): 1.8.1
- Kubernetes version (use `kubectl version`): 1.22
- Kubernetes installer & version:
- Cloud provider or hardware configuration: vSphere
- OS (e.g. from `/etc/os-release`): Photon

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

reasonerjt:
@gsadhani Could you clarify why the opt-out approach can't solve the problem?
Is it b/c you don't want the user to update his workload to skip the backup

gsadhani:
> @gsadhani Could you clarify why the opt-out approach can't solve the problem? Is it b/c you don't want the user to update his workload to skip the backup

@reasonerjt yes, identify each pod in the workload containing the specific type of volume and annotating them could be cumbersome. It would be very convenient if we could skip the volumes based on volume type which can be specified as a backup configuration parameter.

euclidsun:
there are many factors could differentiate volume _type_ : storage class, or bigger, or nfs, or specific drivers.
is it support to add the label velero.io/exclude-from-backup=true to mark the pv as excluded?

euclidsun:
found more SKIP backup requirements, we should consider all cases and comeup with a generic way to handle skip:

https://github.com/vmware-tanzu/velero/issues/957  skip restic backup/restore: not backup the PV data, but backup the cluster state.
https://github.com/vmware-tanzu/velero/issues/4129  Exclude folders for backing up
https://github.com/vmware-tanzu/velero/issues/2413 Namespace skip (this could be enhanced as separate enhancement)

[...]

Issue #5152, opened 2022-07-26
Add new field additionalResources in backup spec


27149chen:
**Describe the problem/challenge you have**
The current resource filtering mechanism (IncludedResources/ExcludedResources + IncludeClusterResources flag) is not flexible enough to cover some special scenarios

**Describe the solution you'd like**
Add a new field called additionalResources (or rawResources), these resources are not affected by the filtering rules in the spec

```
additionalResources:
- group: apps
  version: ""
  resource: deployments
  name: default/my-deploy
```

**Anything else you would like to add:**

velero can be used as an internal backup tool, and the upper systems may have their own resource filtering rules. Instead of translating their rules into velero's rules, velero can provide a raw resources field so that they can fill the result directly.

**Environment:**

- Velero version (use `velero version`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

sseago:
I see a couple possible downsides to this -- it doesn't allow exclude by name, and the interaction with the existing include/excludeResources could be confusing. Instead, we could provide a new pair of fields alongside includeResourses/excludeResources with the intent that users will use one or the other, not both. The old fields work just like before. The new fields (perhaps called structuredIncludes/structuredExcludes) would be similar to that proposed above, with two modifications:
1) the names/items field is optional -- if the list is empty, then that means include (or exclude) all items of the type.
2) we have "structuredExcludes" as well, with functionality similar to current includes/excludes. i.e. you can have an "include everything except this list", or "only include this list".

Something like this:
```
stucturedIncludes:
- group: apps
  resource: deployments
  names:
  - default/my-deploy
- group: ""
  resource: pods
```
This would include the named deployment and all pods. Alternatively, using excludes:
```
stucturedExcludes:
- group: apps
  resource: deployments
  names:
  - default/my-deploy
- group: ""
  resource: pods
```
This would exclude the deployment default/my-deploy (including all others), and exclude all pods

[...]

Issue #5118, opened 2022-07-14
Support filter the resource by name during the backup


ywk253100:
Currently, the `--include-resources` option of `velero backup create` command only supports specifying the resource type, but there are requirements that only need to back up a specified resource. 

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

27149chen:
+ velero restore for the same requirement

reasonerjt:
IMO this is a slightly lower priority than https://github.com/vmware-tanzu/velero/issues/5120
as it feels natural to use label to solve this problem?

27149chen:
> IMO this is a slightly lower priority than #5120 as it feels natural to use label to solve this problem?

I think we should not touch user's resources, and it might cause probelms. Also labels don't work sometimes since resouces owner can revert such changes

ywk253100:
This issue https://github.com/vmware-tanzu/velero/issues/5152 describes one option about how to implement this feature

sseago:
@ywk253100 note that I've suggested a change to https://github.com/vmware-tanzu/velero/issues/5152 to make it more flexible with respect to includes/excludes as well as less confusing with respect to existing include/exclude behavior.

neginhsobhani:
> Currently, the `--include-resources` option of `velero backup create` command only supports specifying the resource type, but there are requirements that only need to back up a specified resource.
> 
> **Vote on this issue!**
> 
> This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). Use the "reaction smiley face" up to the right of this comment to vote.
> 
> * 👍 for "I would like to see this bug fixed as soon as possible"
> * 👎 for "There are more important bugs to focus on right now"

I've deployed Velero on kubernetes cluster in corporate level and we have the same requirement. Hope they add this feature soon.

ryankenney-dev:
[...]

PR #5773, opened 2023-01-17
Design for Handling backup of volumes by resources filters

qiuming-best:

Fixes #(issue)
#5035

### DIFF
--- a/design/handle-backup-of-volumes-by-resources-filters.md
+++ b/design/handle-backup-of-volumes-by-resources-filters.md
@@ -1,164 +1,188 @@
 # Handle backup of volumes by resources filters
  
 ## Abstract
-Currently, Velero doesn't have one flexible way to filter volumes.
+Currently, Velero doesn't have one flexible way to handle volumes.
  
-If users want to skip backup of volumes or only backup some volumes in different namespaces in batch, currently they need to use the opt-in and opt-out approach one by one, or use label-selector but if it has big different labels on each different related pods, which is cumbersome when they have lots of volumes to handle with. it would be convenient if Velero could provide one way to filter the backup of volumes just by `some specific volumes attributes`.
- 
-Also, currently, it's not accurate enough if the users want to select a specific volume to do a backup or skip by without patching labels or annotations to the pods. It would be useful if users could accurately select target volume by `one specific resource selector`. Users could accurately select the volume to backup or skip in their own console when using velero for secondary development.
+If users want to skip the backup of volumes or only backup some volumes in different namespaces in batch, currently they need to use the opt-in and opt-out approach one by one, or use label-selector but if it has big different labels on each different related pod, which is cumbersome when they have lots of volumes to handle with. it would be convenient if Velero could provide policies to handle the backup of volumes just by `some specific volumes conditions`.
  
 ## Background
-As of Today, Velero has lots of filters to handle (backup or skip backup) resources including resources filters like `IncludedNamespaces, ExcludedNamespaces`, label selectors like `LabelSelector, OrLabelSelectors`, annotation like `backup.velero.io/must-include-additional-items` etc. But it's not enough flexible to handle volumes, we need one generic way to filter volumes.
+As of Today, Velero has lots of filters to handle (backup or skip backup) resources including resources filters like `IncludedNamespaces, ExcludedNamespaces`, label selectors like `LabelSelector, OrLabelSelectors`, annotation like `backup.velero.io/must-include-additional-items` etc. But it's not enough flexible to handle volumes, we need one generic way to handle volumes.
  
 ## Goals
-- Introducing one flexible way to filter volumes.
+- Introducing flexible policies to handle volumes, and do not patch any labels or annotations to the pods or volumes.
  
-## Non Goals
+## Non-Goals
 - We only handle volumes for backup and do not support restore.
-- Currently, only handle volumes, does not support other resources.
+- Currently, only handles volumes, and does not support other resources.
 - Only environment-unrelated and platform-independent general volumes attributes are supported, do not support volumes attributes related to a specific environment.
  
 ## Use-cases/Scenarios
-### A. Skip backup volumes by some attributes
+### Skip backup volumes by some attributes
 Users want to skip PV with the requirements:
 - option to skip all PV data
 - option to skip specified PV type (RBD, NFS)
 - option to skip specified PV size
 - option to skip specified storage-class
-- option to skip folders
- 
-### B. Accurately select target volume to backup
-Some volumes are only used for logging while others volumes are for DBs, and only need to backup DBs data. users need `one specific resource seletor` to accurately select target volumes to backup.
  
 ## High-Level Design
-Add a new flag `pv-backup-policy-configmap` when executing `velero backup create`, which imports the defined resources filters in one YAML file. the YAML file including all defined filter rules for the current backup.
- 
-When Velero handles volumes backup should respect the filter rules defined in the imported YAML file.
+First, Velero will provide the user with one YAML file template and all supported volume policies will be in.
+
+Second, writing your own configuration file by imitating the YAML template, it could be partial volume policies from the template.
+
+Third, create one configmap from your own configuration file, and the configmap should be in Velero install namespace.
+
+Fourth, create a backup with the command `velero backup create --resource-policies-configmap $policiesConfigmap`, which will reference the current backup to your volume policies. At the same time, Velero will validate all volume policies user imported, the backup will fail if the volume policies are not supported or some items could not be parsed.
+
+Fifth, the current backup CR will record the reference of volume policies configmap.
+
+Sixth, Velero first filters volumes by other current supported filters, at last, it will apply the volume policies to the filtered volumes to get the final matched volume to handle.
  
 ## Detailed Design
-The resources filters rules should contain both `include` and `exclude` rules.
- 
-For the rules on `one specific resource selector`, we introduced a `GVRN` way of resources filters, for resources are identified by their resource type and resource name, or GVRN.
+The volume resources policies should contain a list of policies which is the combination of conditions and related `action`, when target volumes meet the conditions, the related `action` will take effection.
 
-Here we call it `GVRN Selector` which exactly matches the resources to be handled.
+Below is the API Design for the user configuration:
 
-For the attributes on `some specific volumes attributes`, we basically follow the defined data struct [PersistentVolumeSpec](https://github.com/kubernetes/kubernetes/blob/v1.26.0/pkg/apis/core/types.go#L304), and only handle partial common fields of it currently.
- 
-Here we call it `Volumes Attributes Selector`, which matches volumes with the same attributes defined.
- 
-### filter fields format
-The filter YAML config file would look like this:
+### API Design
+```go
+type VolumeActionType string
+
+const Skip VolumeActionType = "skip"
+
+// Action defined as one action for a specific way of backup
+type Action struct {
+	// Type defined specific type of action, it could be 'file-system-backup', 'volume-snapshot', or 'skip' currently
+	Type VolumeActionType `yaml:"type"`
+	// Parameters defined map of parameters when executing a specific action
+	// +optional
+	// +nullable
+	Parameters map[string]interface{} `yaml:"parameters,omitempty"`
+}
+
+// VolumePolicy defined policy to conditions to match Volumes and related action to handle matched Volumes
+type VolumePolicy struct {
+	// Conditions defined list of conditions to match Volumes
+	Conditions map[string]interface{} `yaml:"conditions"`
+	Action     Action                 `yaml:"action"`
+}
+
+// ResourcePolicies currently defined slice of volume policies to handle backup
+type ResourcePolicies struct {
+  Version        string         `yaml:"version"`
+	VolumePolicies []VolumePolicy `yaml:"volumePolicies"`
+	// we may support other resource policies in the future, and they could be added separately
+  	// OtherResourcePolicies: []OtherResourcePolicy
+}
 ```
----
-resources:
-  - include:
-      groupResource: "/persistentvolumes"
-      namespacedNames: "/nginx-logs"
-  - include:
-      groupResource: "/persistentvolumes"
-      namespacedNames: "/minio"
-  - exclude:
-      groupResource: apps/deployments
-      namespacedNames: velero/velero
-  - exclude: 
-      groupResource: apps/demonset
-      namespacedNames: velero/node-agent
-storage:
-  pv:
-    - include:
-        storageClassName: gp2, ebs-sc
-        volumeMode: block, filesystem
-        capacity: OGi,5Gi
-        persistentVolumeSource:
-          nfs:
-            readOnly: true
-          csi:
-            driver: aws.efs.csi.driver
-    - include:
-        storageClassName: io1
-        persistentVolumeSource:
-          csi:
-            driver: aws.efs.csi.driver
-    - exclude:
-        storageClassName: efs-dynamic-sc
-        capacity: 5Gi,
-        persistentVolumeSource:
-          nfs: {}
-          csi:
-            driver: aws.ebs.csi.driver
-    - exclude:
-        storageClassName: fsx-lustre
-        volumeMode: block, filesystem
-        persistentVolumeSource:
-          csi:
-            driver: csi.aws.amazon.com
+
+The policies YAML config file would look like this:
+```yaml
+version: v1
+volumePolicies:
+# it's a list and if the input item matches the first policy, the latters will be ignored
+# each policy consists of a list of conditions and an action
+
+# each key in the object is one condition, and one policy will apply to resources that meet ALL conditions
+- conditions:
+    # capacity condition matches the volumes whose capacity falls into the range
+    capacity: "0,100Gi"
+    csi:
+      driver: aws.ebs.csi.driver
+      fsType: ext4
+    storageClass:
+    - gp2
+    - ebs-sc
+  action:
+    type: volume-snapshot
+    parameters:
+      # optional parameters which are custom-defined parameters when doing an action
+      volume-snapshot-timeout: "6h"
+- conditions:
+    capacity: "0,100Gi"
+    storageClass:
+    - gp2
+    - ebs-sc
+  action:
+    type: file-system-backup
+- conditions:
+    nfs:
+      server: 192.168.200.90
+  action:
+    # type of file-system-backup could be defined a second time
+    type: file-system-backup
+- conditions:
+    nfs: {}
+  action:
+    type: skip
+- conditions:
+    csi:
+      driver: aws.efs.csi.driver
+  action:
+    type: skip
 ```
  
 ### Filter rules
-The whole filter file consists of two parts: resources and storage.
- 
-Both `Kopia, Restic` and `Volume snapshot` share one YAML configuration file.
- 
-#### resources
-In the resources part, we defined `GVRN Selector` to filter resources. In a filter, an empty or omitted group, version, resource type, or resource name matches any value. `GVRN selector` could match Persistent Volume and other Kubernetes resources.
+#### VolumePolicies
+The whole resource policies consist of groups of volume policies.
 
-Taking select PV as an example, if users want to backup PV with name nginx-logs, the `groupResource` could be "/persistentvolumes" in which the group should be empty, the `namespacedNames` could be "/nginx-logs" in which the namespace should be empty.
+For one specific volume policy which is a combination of one action and serval conditions. which means one action and serval conditions are the smallest unit of volume policy.
 
-#### storage
-In the storage part, we defined `Volumes Attributes Selector` to filter resources.
- 
-The storage part defined rules including `pv` and `volume`, which correspond to `Kopia, Restic` and `Volume snapshot`.
- 
-A filter in storage with a specific key and empty value, which means the value matches any value. For example, if the `storage.pv.exclude.persistentVolumeSource.nfs` is `{}` it means if `NFS` is used as `persistentVolumeSource` in Persistent Volume will be skipped no matter what the NFS server or NFS Path is,  
- 
-A filter may have multiple values, all the values are concatenated by commas. For example, the `storage.pv.include.storageClassName` is `gp2, ebs-sc` which means Persistent Volume with gp2 or ebs-sc storage class both will be back up.
- 
-The size of each single filter value should limit to 256 bytes in case of an unfriendly long variable assignment.
- 
-If user defined pv filter rules but used Kopia or Restic to do a backup, the backup will fail in validating the resource filter configuration. Same as the situation if using defined volume filter rules but using CSI or plugins to take volume snapshots.
- 
-For capacity in `pv` or size in `volume`, the value should include the lower value and upper value concatenated by commas. And it has several combinations below:
-- "0,5Gi" or "0Gi,5Gi" which means capacity or size matches from 0 to 5Gi, including value 0 and value 5Gi
-- ",5Gi" which is equal to "0,5Gi"
-- "5Gi," which means capacity or size matches larger than 5Gi, including value 5Gi
-- "5Gi" which is not supported and will be failed in validating configuration.
- 
-### Filter Reference
-Currently, resources filters are defined in `BackupSpec` struct, it will be more and more bloated with adding more and more filters which makes the size of `Backup` CR bigger and bigger, so we want to store the resources rules in configmap, and `Backup` CRD reference to current configmap.
+Volume policies are a list and if the target volumes match the first policy, the latter will be ignored, which would reduce the complexity of matching volumes especially when there are multiple complex volumes policies.
 
-the `configmap` would be like this:
-```
+#### Action
+`Action` defined one action for a specific way of backup:
+  - if choosing `Kopia` or `Restic`, the action value would be `file-system-backup`.
+  - if choosing volume snapshot, the action value would be `volume-snapshot`.
+  - if choosing skip backup of volume, the action value would be `skip`, and it will skip backup of volume no matter is `file-system-backup` or `volume-snapshot`.
+
+The policies could be extended for later other ways of backup, which means it may have some other `Action` value that will be assigned in the future.
+
+Both `file-system-backup` `volume-snapshot`, and `skip` could be partially or fully configured in the YAML file. And configuration could take effect only for the related action.
+
+#### Conditions
+The conditions are serials of volume attributes, the matched Volumes should meet all the volume attributes in one conditions configuration.
+
+##### Supported conditions
+In Velero 1.11, we want to support the volume attributes listed below:
+- capacity: matching volumes have the capacity that falls within this `capacity` range.
+- storageClass: matching volumes those with specified `storageClass`, such as `gp2`, `ebs-sc` in eks.
+- matching volumes that used specified volume sources.
+##### Parameters
+Parameters are optional for one specific action. For example, it could be `csi-snapshot-timeout: 6h` for CSI snapshot.
+
+#### Special rule definitions:
+- One single condition in `Conditions` with a specific key and empty value, which means the value matches any value. For example, if the `conditions.nfs` is `{}`, it means if `NFS` is used as `persistentVolumeSource` in Persistent Volume will be skipped no matter what the NFS server or NFS Path is.
+ 
+- The size of each single filter value should limit to 256 bytes in case of an unfriendly long variable assignment.
+ 
+- For capacity for PV or size for Volume, the value should include the lower value and upper value concatenated by commas. And it has several combinations below:
+  - "0,5Gi" or "0Gi,5Gi" which means capacity or size matches from 0 to 5Gi, including value 0 and value 5Gi
+  - ",5Gi" which is equal to "0,5Gi"
+  - "5Gi," which means capacity or size matches larger than 5Gi, including value 5Gi
+  - "5Gi" which is not supported and will be failed in validating configuration.
+ 
+### Configmap Reference
+Currently, resources policies are defined in `BackupSpec` struct, it will be more and more bloated with adding more and more filters which makes the size of `Backup` CR bigger and bigger, so we want to store the resources policies in configmap, and `Backup` CRD reference to current configmap.
+
+the `configmap` user created would be like this:
+```yaml
 apiVersion: v1
 data:
- filter.YAML: |-
-   {
-       "resources": {
-           "include": {
-              ...
-           },
-           "exclude": {
-              ...
-           }
-       },
-       "storage": {
-           "pv": {
-               "include": {
-                  ...
-               },
-               "exclude": {
-                  ...
-               }
-           },
-           "volume": {
-               "include": {
-                  ...
-               },
-               "exclude": {
-                  ...
-               }
-           }
-       }
-   }
+ policies.yaml:
+  ----
+  version: v1
+  volumePolicies:
+  - conditions:
+      capacity: "0,100Gi"
+      csi:
+        driver: aws.ebs.csi.driver
+        fsType: ext4
+      storageClass:
+      - gp2
+      - ebs-sc
+    action:
+      type: volume-snapshot
+      parameters:
+        volume-snapshot-timeout: "6h"
 kind: ConfigMap
 metadata:
  creationTimestamp: "2023-01-16T14:08:12Z"
@@ -168,31 +192,72 @@ metadata:
  uid: b73e7f76-fc9e-4e72-8e2e-79db717fe9f1
 ```
 
-a new variable `filterConfigmap` would be added into `BackupSpec`, it's value is assigned with current resources filters configmap
-```
+A new variable `resourcePolices` would be added into `BackupSpec`, it's value is assigned with the current resources policy configmap
+```yaml
 apiVersion: velero.io/v1
 kind: Backup
 metadata:
   name: backup-1
   spec:
-    resourcesFilter:
+    resourcePolices:
       refType: Configmap
       ref: backup01
     ...
 ```
-
-The configmap basically equivalent to generated by command `kubectl create cm backup01 --from-file filter.yaml`
- 
-The configmap only stores those filters assigned value not the whole resources filters.
+The configmap only stores those assigned values, not the whole resources policies.
  
 The name of the configmap is `$BackupName`, and it's in Velero install namespace.
  
-#### Life-cycle of resource filter configmap
-- Resource filter configmap only been generated once Velero backup command with flag `pv-backup-policy-configmap` to import configuration.
-- If the referenced backup has been deleted, the relevant configmap should be removed.
- 
-### Display of volume resource filter
-As the resource filter configmap is referenced by backup CR, the rules in configmap are not so intuitive, so we need to integrate rules in configmap to the output of the command `velero backup describe`, and make it more readable.
+#### Resource policies configmap related
+The life cycle of resource policies configmap is managed by the user instead of Velero, which could make it more flexible and easy to maintain.
+- The resource policies configmap will remain in the cluster until the user deletes it.
+- Unlike backup, the resource policies configmap will not sync to the new cluster. So if the user wants to use one resource policies that do not sync to the new cluster, the backup will fail with resource policies not found.
+- One resource policies configmap could be used by multiple backups.
+- If the backup referenced resource policies configmap is been deleted, it won't affect the already existing backups, but if the user wants to reference the deleted configmap to create one new backup, it will fail with resource policies not found.
+
+#### Versioning
+We want to introduce the version field in the YAML data to contain break changes. Therefore, we won't follow a semver paradigm, for example in v1.11 the data look like this:
+```yaml
+version: v1
+volumePolicies:
+  ....
+```
+Hypothetically, in v1.12 we add new fields like clusterResourcePolicies, the version will remain as v1 b/c this change is backward compatible:
+```yaml
+version: v1
+volumePolicies:
+  ....
+clusterResourcePolicies:
+  ....
+```
+Suppose in v1.13, we have to introduce a break change, at this time we will bump up the version:
+```yaml
+version: v2
+# This is just an example, we should try to avoid break change
+volume-policies: 
+  ....
+```
+We only support one version in Velero, so it won't be recognized if backup using a former version of YAML data.
+
+#### Multiple versions supporting
+To manage the effort for maintenance, we will only support one version of the data in Velero. Suppose that there is one break change for the YAML data in Velero v1.13, we should bump up the config version to v2, and v2 is only supported in v1.13. For the existing data with version: v1, it should migrate them when the Velero startup, this won't hurt the existing backup schedule CR as it only references the configmap. To make the migration easier, the configmap for such resource filter policies should be labeled manually before Velero startup like this, Velero will migrate the labeled configmap.
+
+We only support migrating from the previous version to the current version in case of complexity in data format conversion, which users could regenerate configmap in the new YAML data version, and it is easier to do version control.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+# This label can be optional but if this is not set, the backup will fail after the breaking change and the user will need to update the data manually 
+    velero.io/resource-filter-policies: "true"
+  name: example
+  namespace: velero
+data:
+  .....
+```
+### Display of resources policies
+As the resource policies configmap is referenced by backup CR, the policies in configmap are not so intuitive, so we need to integrate policies in configmap to the output of the command `velero backup describe`, and make it more readable.
  
 ## Compatibility
 Currently, we have these resources filters:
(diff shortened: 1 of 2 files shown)


qiuming-best:
hi @pradeepkchaturvedi,  Here is one draft design pr for handling backup volume by resources filters. If you have any opinion, please comment on it.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,211 @@
...
+       "include": {
+           "groupResource": "/persistentvolumes",
+           "namespacedNames": "/nginx-logs"
+       }
+       "exclude": {
+           "groupResource": "apps/deployments",
+           "namespacedNames": "velero/velero"
+       }
+   },
+   "storage": {
```
Lyndon-Li, line 63: It's better to combine the "pv" and "volume" in storage filters. Reasons:

- Users will need to consider the backup types (volume snapshot or pod volume backup?) when they use these filters, this will add unnecessary burden to users
- We will add more backup types that processing PVs in future, if we keep the backup type differences here, we will have to define new fields for the new backup types

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,211 @@
...
+## Non Goals
+- We only handle volumes for backup and do not support restore.
+- Currently, only handle volumes, does not support other resources.
+- Only environment-unrelated and platform-independent general volumes attributes are supported, do not support volumes attributes related to a specific environment.
+ 
+## Use-cases/Scenarios
+### A. Skip backup volumes by some attributes
+Users want to skip PV with the requirements:
+- option to skip all PV data
+- option to skip specified PV type (RBD, NFS)
```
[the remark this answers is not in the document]
qiuming-best: indeed, the above description is made by PM, the precise description should be `option to skip specified PV volume sources`, which could be retrieved by `PersistentVolumeSource` of PV

All the PV attributes could be retrieved in `PersistentVolumeSpec` if the user could follow the Kubernetes PersistentVolume structure design.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,211 @@
...
+- option to skip specified PV size
+- option to skip specified storage-class
```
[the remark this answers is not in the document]
qiuming-best: For example the storage class in EKS it should be: `gp2`, `io1`, `sc1`, `st1`, `standard` etc.

Users should be familiar with filters, or they will misleading. And I'll give a full example of resource filters in Doc.

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: yes, I agree with you.

design/handle-backup-of-volumes-by-resources-filters.md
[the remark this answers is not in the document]
Lyndon-Li: What about the CSI source? Behind the CSI driver, the storage type may also be NFS, RBD, etc.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+ 
+Also, currently, it's not accurate enough if the users want to select a specific volume to do a backup or skip by without patching labels or annotations to the pods. It would be useful if users could accurately select target volume by `one specific resource selector`. Users could accurately select the volume to backup or skip in their own console when using velero for secondary development.
+ 
+## Background
+As of Today, Velero has lots of filters to handle (backup or skip backup) resources including resources filters like `IncludedNamespaces, ExcludedNamespaces`, label selectors like `LabelSelector, OrLabelSelectors`, annotation like `backup.velero.io/must-include-additional-items` etc. But it's not enough flexible to handle volumes, we need one generic way to filter volumes.
+ 
+## Goals
+- Introducing one flexible way to filter volumes.
```
reasonerjt, line 14: This mechanism should have the flexibility to allow we decide the actual action taken to the pv satisfy a certain filter, for example for PVs provisioned via storage class A, the user wants to take snapshots, for PVs provisioned by storage class B, the user wants to bakcup via kopia.

We don't need to support that in the first version of implementation, but we should make sure there's such extensibility.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+### A. Skip backup volumes by some attributes
+Users want to skip PV with the requirements:
+- option to skip all PV data
+- option to skip specified PV type (RBD, NFS)
+- option to skip specified PV size
+- option to skip specified storage-class
+- option to skip folders
+ 
+### B. Accurately select target volume to backup
+Some volumes are only used for logging while others volumes are for DBs, and only need to backup DBs data. users need `one specific resource seletor` to accurately select target volumes to backup.
```
reasonerjt, line 31: How do we support that if these two PVs have the same storage class?

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+ 
+## High-Level Design
+Add a new flag `from-file` when executing `velero backup create`, which imports the defined resources filters in one JSON file. the JSON file including all defined filter rules for the current backup.
```
reasonerjt, line 34: Shall we let it be `--pv-backup-policy-configmap` so it's more specific?

I suggest we also mention how to introduce this change in API level (backup CR referencing the configmap)

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+ 
+When Velero handles volumes backup should respect the filter rules defined in the imported JSON file.
+ 
+## Detailed Design
+The resources filters rules should contain both `include` and `exclude` rules.
```
reasonerjt, line 39: Let's focus on PVs, not other resources.

If we define multiple filters in one config, what's the relationship between the filters?  My initial thought is that it should be OR, such that each `include` and `exclude` rule can be simple and it may be possible to satisfy the scenarios like for PVs that meet criteria in filter A we take a snapshot, for PVs that meet criteria in filter B we use kopia?

I am open for discussion, admitting if we wanna support different actions for different PVs, things can be tricky.

design/handle-backup-of-volumes-by-resources-filters.md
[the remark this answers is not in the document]
pradeepkchaturvedi: In my understanding, NFS can be used in Pod by three ways
- mount as Volume in Pod spec
-  mount as NFS type PV/PVC
- mount using CSI type PV/PVC where NFS CSI driver used

Is it not possible to detect volume source as NFS in all three cases? 

I have seen most time user wants to skip NFS volume in backup or want to skip such volumes from backup which does not support volume snapshot like NFS CSI driver does not support CSI volume snapshot.

In brief, user needs a way to say
- skip backup of all volumes & PV/PVC
- skip such volume & PV/PVC which does not support volumesnaphot AND backup all volumes which supports volumesnapshot

We would also need a way to specify:
- backup Volumes/PV using Kopia as well as volume snapshot
- backup  Volumes/PV using Kopia only (skip volume snapshot for volumes even it support volume snapshot)
- backup Volumes/PV using volume snapshot only (and skip such volumes which does not support volume snapshot)
- backup all Volumes/PV but use snapshot for volumes support snapshotting and backup rest volume using kopia

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+Also, currently, it's not accurate enough if the users want to select a specific volume to do a backup or skip by without patching labels or annotations to the pods. It would be useful if users could accurately select target volume by `one specific resource selector`. Users could accurately select the volume to backup or skip in their own console when using velero for secondary development.
+ 
+## Background
+As of Today, Velero has lots of filters to handle (backup or skip backup) resources including resources filters like `IncludedNamespaces, ExcludedNamespaces`, label selectors like `LabelSelector, OrLabelSelectors`, annotation like `backup.velero.io/must-include-additional-items` etc. But it's not enough flexible to handle volumes, we need one generic way to filter volumes.
+ 
+## Goals
+- Introducing one flexible way to filter volumes.
+ 
+## Non Goals
+- We only handle volumes for backup and do not support restore.
```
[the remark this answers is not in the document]
reasonerjt: @pradeepkchaturvedi 
That's for this comment.

This is a valid requirement, but generally, we consider backup/restore two separate workflows, we may use a similar mechanism to handle resources, but we want to design and implement that at different phases, let's focus on backup workflow in this PR, I've opened #5803 to track the similar problem on the restore side.  Of course, we may use similar data structure to filter resources during restore.

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: which means that through the resource filter configuration for one **single** backup, Velero could both take snapshots for some volumes and  Kopia/Restic backup for other volumes?

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+- Currently, only handle volumes, does not support other resources.
```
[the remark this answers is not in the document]
qiuming-best: yes

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: For those PVs has the same attributes and are hard to distinguish, we could use GVRN ways to accurately select PV to backup or skip. Which would be a group of combinations of  `groupResource` + `namespacedNames`

design/handle-backup-of-volumes-by-resources-filters.md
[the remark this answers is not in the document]
qiuming-best: > In brief, user needs a way to say

> skip backup of all volumes & PV/PVC
>skip such volume & PV/PVC which does not support volumesnaphot AND backup all volumes which supports volumesnapshot

we could skip by the volumes sources or storage class.

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: In our backup workflow, we list PVs and decide whether to take snapshots or not, we list pods and find volumes mount in pods and decide whether to backup or not. which is two different code path.

Maybe we should consider how to relate pod volumes to PV attributes when using PV attributes as filters under Kopia/Restic backup

design/handle-backup-of-volumes-by-resources-filters.md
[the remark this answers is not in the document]
reasonerjt: @pradeepkchaturvedi 

I agree that we should let the user choose to use fs-based backup like kopia when the default approach is "volume snapshot" or "CSI snapshot". IMO in this design, we should make it possible to support an "if... then..." semantic so that the user can define if a PV meets certain conditions it can be backed up via certain ways. 

I don't think it's a high-priority use case to back up one object using different approaches, it makes the restore workflow more complicated and more confusing, and the user won't tell which way is better.

design/handle-backup-of-volumes-by-resources-filters.md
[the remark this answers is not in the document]
reasonerjt: Personally, I think it's sufficient in the first delivery to support filtering PVs based on storage class, as this should be commonly used to decide how the volume is provisioned, though we can make the structure flexible enough and we expand to other attributes based on feedback.

@pradeepkchaturvedi thoughts?

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+                  ...
+               }
+           }
+       }
+   }
+kind: ConfigMap
+metadata:
+ creationTimestamp: "2023-01-16T14:08:12Z"
+ name: cm-backup01
+ namespace: default
```
reasonerjt, line 173: Shall we keep everything in velero namespace

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+a new variable `filterConfigmap` would be added into `BackupSpec`, it's value is assigned with current resources filters configmap
+```
+apiVersion: velero.io/v1
+kind: Backup
+metadata:
+  name: backup-1
+  spec:
+    filterConfigmap: cm-backup01
```
reasonerjt, line 185: To make it a little more flexible, how about 
```
XXXfilter:
  ref-type: Configmap
  ref: cm-backup01
```
As for what the `XXX` should be, I'm not very sure it should be `resource` or `object` or some other term.  The term `resource` seems ambiguous in k8s, sometimes it means a _kind_ of resource, for example, `pod` or `storageclass` are referred to as resource names, but `pod-abc` is the name of an instance of `pod` resource, and the goal for this design is to introduce a filter to `persistentvolume`, which is more detailed than resource level.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+    ...
+```
+
+The configmap basically equivalent to generated by command `kubectl create cm cm-backup01 --from-file filter.json`
+ 
+The configmap only stores those filters assigned value not the whole resources filters.
+ 
+The name of the configmap is `cm-`+`$BackupName`, and it's in Velero install namespace.
```
reasonerjt, line 193: It doesn't have to have the prefix `cm-`

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+ 
+#### Life-cycle of resource filter configmap
+- If resource filter configmap is only been imported by Velero backup command with flag `from-file`.
```
reasonerjt, line 196: Don't quite understand this sentence.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,223 @@
...
+- If the referenced backup has been deleted, the relevant configmap should be removed.
+- Resource filter configmap should also persist into object storage and could be also synchronized automatically when at startup.
```
reasonerjt, line 198: So for the life-cycle of the configmap, if the backup is removed the configmap may not need to be removed and IMO it's OK one configmap being referenced by different backups.

We should probably upload the referenced configmap to object storage but syncing the configmap is probably unnecessary, b/c the configmap with the same name can have different content when different backups are taken, syncing the configmap map cause one overwrites another, at the sametime it doesn't help in the restore workflow.

design/handle-backup-of-volumes-by-resources-filters.md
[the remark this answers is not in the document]
qiuming-best: > In my understanding, NFS can be used in Pod by three ways
> 
> * mount as Volume in Pod spec
> * mount as NFS type PV/PVC
> * mount using CSI type PV/PVC where NFS CSI driver used

@pradeepkchaturvedi @reasonerjt  
above are references to two different[ storage types](https://kubernetes.io/docs/concepts/storage/):
mount as Volume in Pod spec --> [pod volume](https://kubernetes.io/docs/concepts/storage/volumes/)
mount as NFS type PV/PVC --> PV
mount using CSI type PV/PVC where NFS CSI driver used --> PV

For our Velero to do the backup, we only needed to consider `pod volume` and `PV` to filter volumes.

How do we classify our resource filter configuration?
- classify by storage types: `pod volume` and `PV`
- classify by approach, which would be `Volume Snapshot`, `CSI snapshot`, `fs-based backup`, or even `data-mover`

obviously, classifying by storage types will be more clear. `pod volume` and `PV` has the same attribute volume source, which indicates the Volume type is `NFS` or `CSI` etc. which may cover most requirement. 


>I agree that we should let the user choose to use fs-based backup like kopia when the default approach is "volume snapshot" or "CSI snapshot". IMO in this design, we should make it possible to support an "if... then..." semantic so that the user can define if a PV meets certain conditions it can be backed up via certain ways.

Suppose we use filter configuration classified by storage types, If we don't follow the default approach specified by the command of `velero create`, instead let `PV meets certain conditions it can be backed up via certain ways`, which may be indeterminate. And it will bring complexity to the whole backup workflow. What about we only follow the default approach to handle PV meets certain conditions, if PV couldn't be handled by default approach just failed directly.  For example: if a user wants to backup a volume with a CSI snapshot, and PV has the CSI persistent volume source, but the CSI driver does not support snapshots for the underlying storage system does not support snapshots. maybe it's not easy for us to confirm the "if... then...".

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: If we only focus on PV I think it's suitable

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: I'm not sure `meet criteria` is easy to confirm with a specific user environment

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: yes, the value is incorrect, I'll modify it to velero.

design/handle-backup-of-volumes-by-resources-filters.md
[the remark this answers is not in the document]
reasonerjt: @qiuming-best 
IMO We should first support filtering "storage type" based on storage class, this may be sufficient already
https://github.com/vmware-tanzu/velero/pull/5773/files#r1090342124

>  And it will bring complexity to the whole backup workflow.
It will, but the user can choose not to use it.  IMO letting the user to choose backup via podvolume or snapshot is a valid requirement(we probably don't need to support it in v1.11, but we need to make sure the data structure has such flexibility).

design/handle-backup-of-volumes-by-resources-filters.md
reasonerjt: Like you mentioned in 
https://github.com/vmware-tanzu/velero/pull/5773/files#r1090315744
Worst case the user can choose to use GVRN...

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: If we want to extend the filter policy, Maybe `resourcesFilter` is good?  or if only focus on `persistentvolume` maybe it could be `PVfilter` or `volumeFilter`?

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: > one configmap being referenced by different backups
How do we user reference configmap to another different backup? if we want to implement it, should we introduce another flag like `pv-backup-policy-from` when creating a backup to reference another existing backup?

> syncing the configmap is probably unnecessary
agree with you

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: yes, it should be in velero namespace

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: ok

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: misspell some words

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: And should we store configmap to object storage if we don't sync it ?

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,246 @@
...
+  - if choosing `Kopia` or `Restic`, the action value would be `file-system-backup`.
+  - if choosing volume snapshot, the action value would be `volume-snapshot`.
+  - if choosing skip backup of volume, the action value would be `skip`, and it will skip backup of volume no matter is `file-system-backup` or `volume-snapshot`.
+
+The policies could be extended for later other ways of backup, which means it may have some other `Action` value that will be assigned in the future.
+
+Both `file-system-backup` `volume-snapshot`, and `skip` could be partially or fully configured in the YAML file. And configuration could take effect only for the related action.
+
+#### Conditions
+The conditions are serials of volume attributes, the matched Volumes should meet all the volume attributes in one conditions configuration.
```
gsadhani, line 124: Where can we find the full reference of supported volume attributes? Would Velero validate the attributes specified in the condition against the reference?

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,246 @@
...
+The name of the configmap is `$BackupName`, and it's in Velero install namespace.
+ 
+#### Life-cycle of resources policies configmap
+- Resource policies configmap only been generated once Velero backup command with flag `backup-policy-configmap` to import configuration.
+- Resource policies configmap could be referenced by another backup in Velero backup command with the flag `policy-from-backup` to reference configuration in another backup.
+- If all referenced backup has been deleted, the relevant configmap should be removed.
+#### Versioning
+Here we introduced the version field in the YAML config to contain break changes in case of some compatibility problems:
+```yaml
+version: v1
```
gsadhani, line 198: What was the reason to choose to ConfigMap over a new CRD for resource policy? CRD would allow API versioning of the policy resources.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,246 @@
...
+The whole resource policies consist of groups of volume policies.
+
+For one specific volume policy which is a combination of one action and serval conditions. which means one action and serval conditions are the smallest unit of volume pocliy.
+
+Volume policies are a list and if the target volumes match the first policy, the latter will be ignored, which would reduce the complexity of matching volumes especially when there are multiple complex volumes policies
+
+#### Action
+`Action` defined one action for a specific way of backup:
+  - if choosing `Kopia` or `Restic`, the action value would be `file-system-backup`.
+  - if choosing volume snapshot, the action value would be `volume-snapshot`.
```
rnarenpujari, line 116: I guess this will be used for both CSI snapshots and plugin based snapshots? Will it follow the same logic, i.e., if CSI feature enabled & volume is CSI provisioned then take a CSI snapshot, else plugin based snapshot?

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,246 @@
...
+  action:
+    type: file-system-backup
+- conditions:
+    nfs: {}
+  action:
+    # type of file-system-backup could be defined a second time
+    type: file-system-backup
+- conditions:
+    csi:
+      driver: aws.efs.csi.driver
```
rnarenpujari, line 100: I assume there will be a catch-all condition that we can use at the end? For example if user wants to skip all PVs except storage class abc.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,246 @@
...
+- LabelSelector
+- OrLabelSelectors
+- IncludeClusterResources
+- UseVolumeSnapshots
+- velero.io/exclude-from-backup=true
+- backup.velero.io/backup-volumes-excludes
+- backup.velero.io/backup-volumes
+- backup.velero.io/must-include-additional-items
+ 
+So it should be careful with the combination of volumes resources policies and the above resources filters.
```
rnarenpujari, line 234: What is the expected behavior if a volume matches policy with action "volume-snapshot" but the backup spec has defaultVolumesToRestic?

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,246 @@
...
+    # capacity condition matches the volumes whose capacity falls into the range
+    capacity: "0,100Gi"
+    csi:
+      driver: aws.ebs.csi.driver
+      fsType: ext4
+    storageClass:
+    - gp2
+    - ebs-sc
+  action:
+    type: volume-snapshot
```
rnarenpujari, line 82: What if I specify action that cannot be completed? Say "volume-snapshot" but I have not created any VolumeSnapshotLocation and CSI feature is disabled. Or "file-system-backup" but velero is installed without restic/kopia? Will the backup partially fail or will the volume be silently skipped?

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,246 @@
...
+The configmap basically equivalent to what is generated by the command `kubectl create cm backup01 --from-file policies.yaml`
+ 
+The configmap only stores those assigned values, not the whole resources polices.
+ 
+The name of the configmap is `$BackupName`, and it's in Velero install namespace.
+ 
+#### Life-cycle of resources policies configmap
+- Resource policies configmap only been generated once Velero backup command with flag `backup-policy-configmap` to import configuration.
+- Resource policies configmap could be referenced by another backup in Velero backup command with the flag `policy-from-backup` to reference configuration in another backup.
+- If all referenced backup has been deleted, the relevant configmap should be removed.
```
rnarenpujari, line 194: Let's only delete the configmap if it was created by velero using the `backup-policies-from-file` since in other cases the lifecycle may be managed by a third party operator.

ihcsim:
@reasonerjt @qiuming-best I am not convinced that we need to special-case persistent volumes. The introduction of a non-K8s API and corresponding versioning scheme to represent volume policies feels a bit much. Feels like all the `conditions` described in the proposal can be represented using the new include/exclude resource filters in #5838, plus the field selector proposal at #5842. With the exception of the proposed non-`skip` action types, the rest of the volume policies are really just about filtering resources, right? Why not try to integrate with the new filtering mechanism in #5838?

Can you elaborate a bit more on the `file-system-backup` and `volume-snapshot` action type and their `parameters`? Feels like this is something that can be expressed using the existing CRD, independent of the filtering mechanism.
 
The field selector mechanism proposed in #5842 uses the [`fields.Fields`](https://pkg.go.dev/k8s.io/apimachinery/pkg/fields#Fields) to represent the filter attributes. It is possible to provide our implementation of [`fields.Selector`](https://pkg.go.dev/k8s.io/apimachinery@v0.26.1/pkg/fields#Selector) to satisfy the `conditions` example in this proposal. Then the `Backup` resource can look something like:

```yaml
apiVersion: velero.io/v1
kind: Backup
metadata:
  name: my-backup
  namespace: velero
spec:
  includedClusterScopeResources:
  - groupResource:
    group: "v1"
    resource: "persistentvolumes"
    selectors:
      capacity: "0,100Gi"
      csi:
        driver: aws.ebs.csi.driver
        fsType: ext4
      storageClass:
      - gp2
      - ebs-sc
  excludedClusterScopeResources:
  - groupResource:
      group: "v1"
      resource: "persistentvolumes"
      selectors:
        csi:
          driver: aws.efs.csi.driver
  - groupResource:
      group: "v1"
      resource: "persistentvolumes"
      selectors:
        nfs: {}
``` 
Building on top of #5838, I believe this approach can significantly reduce the implementation and maintenance scope, where we just need to focus on the handling of the field selectors. (K8s only supports a handful of field selectors on the server-side.)

Even if we think the proposed volume policy is necessary, I don't think #5842 contradicts this proposal. It addresses issues like #5152 and #5118 which this proposal doesn't. Based on our discussion on the community call, I would like to get a sense of how this policy can be extended to include other cluster-scoped resources like CRDs, `APIService`, `ClusterRole`, filtering by their names. I suspect we might end up with supporting different sets of conditions for different resources (unlike field selectors where K8s provide a sensible default).

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: I've added more details to the design now.

the reason we choose Configmap is:

CRD is more like one kind of resource with status, Kubernates API Server handles the lifecycle of a CR and handles it in different statuses. Compared to CRD, Configmap is more focused to store data. And we only want store policies, so Configmap is more suitable.

 `CRD would allow API versioning` also needs our Velero codes to explain it, so it may be no different to use Configmap, we don't change the version of Configmap, but maintain the version in data part of Configmap.

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: If we take a volume snapshot, we could use the CSI plugin or other plugins such as `velero-plugin-for-aws`, but these two kinds of plugins couldn't work in one backup, so we don't need to distinguish between which plugin is.

But we have this kind of scenario: we could take a volume snapshot and do file-system-backup(Kopia or Restic), so we need to  distinguish between them.

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: @gsadhani yes, I've added more details to the design now.

We will provide one doc to show users the volume resource policies template,  which will contain all the volume attributes we will support.

```yaml
- conditions:
    # capacity condition matches the volumes whose capacity falls into the range
    capacity: "0,100Gi"
    csi:
      driver: aws.ebs.csi.driver
    storageClass:
    - gp2
    - ebs-sc
```

above is current we want to support the attributes:`capacity, driver, starageClass`
we will extend supported attributes later.

Do these  attributes are enough for users to backup or skip backup volumes?

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: every policy has related action, if the action is a `volume-snapshot` but we are doing backup with Resitc, then this policy will not take effect.

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: I've updated my design now.

The resource policies configmap is maintained by users. which would be more flexible. users could create or delete configmap through `kubectl`

Velero will provide one flag `reousce-pilicies-configmap` to refer to resource policies configmap created by users when using the command `velero backup create`

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: whether should we have a `catch-all condition` is depend on user configuration. If users configured it, velero would support it.

The current situation would be the ignored volumes will not be backed up.
suppose we have volumes `A, B, C, and D`, through all kinds of resources filters, A and B are defined to include, D is excluded, Velero will backup volumes `A` and `B`, and `D` won't be backed up. the ignored `C` will not be backed up by default.

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: All the policies are responsible for finding matched volumes, whether the  matched volumes could be backup or not is related to velero backup logic, so it will be partially failed based on your scenario

design/handle-backup-of-volumes-by-resources-filters.md
reasonerjt: Velero will validate and report the unsupported conditions.
@qiuming-best there should be a section in this PR to clarify what conditions will be supported in v1.11, and we should document it.

design/handle-backup-of-volumes-by-resources-filters.md
reasonerjt: We did think about adding CR, but favor CM over CR for 2 reasons:
1. We don't want to introduce the full set of subcommands immediately to manage the lifecycle of these CRs, we may make it a CR in the future if this approach is verified.  
2. There's also debate that, this piece of data is purely for reference, we don't want to introduce a controller to reconcile the status if we make it a CR.

design/handle-backup-of-volumes-by-resources-filters.md
reasonerjt: Yes, I think that's a good enhancement, thanks @rnarenpujari 

cc @qiuming-best

design/handle-backup-of-volumes-by-resources-filters.md
reasonerjt: IMO the actions like "volume-snapshot" are put here to show the extensibility of this design, and in v1.11 the MVP is to support `skip`

We'll need to specify the priority when there's conflict when introducing more types of `actions`, thoughts?

reasonerjt:
@ihcsim 
Thanks for the comment.

The reason there's the `action` and `parameters` fields is that we wanna make sure there're flexibilities so that the user can choose to do more actions against them in addition to include/exclude.  For example in one backup, for volumes provisioned via storageClass A, with certain capacities, the user may wanna choose to use `file-system-back` approach (restic/kopia) to backup the data, and for some volumes that meet certain criteria, the user wanna choose to take snapshots, and some certain volumes will be skipped, etc. 

 The fieldSelector in #5842 may be sufficient for filtering resources based on names, but as for PV there are quite a few fields, and nested structures in the spec.  In our internal discussion, we found matching the values of the fields does not satisfy the use case, for example, `capacity` is a range in the policy data,  and to apply the policy against a PV we do not simply check if the value in the selector is the same as in the spec, therefore we introduce the `condition` which does not have to be a 1:1 mapping to each field in the spec.
 
I agree the policy proposed in the PR does not contradict #5842.  I was thinking we may extend the policy to add `resourcePolicies` which support more general conditions like `type: "customresourcedefinitions"`, `name: "name1, name2"`.  I slightly prefer not to add too many fields into the spec of backup CR, but I'm open for discussion in this part.

design/handle-backup-of-volumes-by-resources-filters.md
rnarenpujari: > The current situation would be the ignored volumes will not be backed up.
> suppose we have volumes A, B, C, and D, through all kinds of resources filters, A and B are defined to include, D is excluded, Velero will backup volumes A and B, and D won't be backed up. the ignored C will not be backed up by default.  

OK skip was not a good example. If we want to snapshot storage class abc but use restic for everything else, then a catch-all/wildcard condition at the end would be nice to have I think.

design/handle-backup-of-volumes-by-resources-filters.md
rnarenpujari: Makes sense. From Qiuming's comment I guess the policy is given less priority in all cases.

design/handle-backup-of-volumes-by-resources-filters.md
ihcsim: The `status` subresource is optional. Policy engines like OPA and Kyverno use CRD to represent policies. With CRD, at least users are familiar with the corresponding K8s versioning scheme.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,270 @@
...
+ 
+### Filter rules
+#### VolumePolicies
+The whole resource policies consist of groups of volume policies.
+
+For one specific volume policy which is a combination of one action and serval conditions. which means one action and serval conditions are the smallest unit of volume policy.
+
+Volume policies are a list and if the target volumes match the first policy, the latter will be ignored, which would reduce the complexity of matching volumes especially when there are multiple complex volumes policies.
+
+#### Action
```
ihcsim, line 121: > The reason there's the action and parameters fields is that we wanna make sure there're flexibilities so that the user can choose to do more actions against them in addition to include/exclude.

@reasonerjt Seems to me there is potentially a "filtering" and a "grouping" requirements. If grouping is the main use case where we want to apply certain actions to a group of volumes, I wonder if a volume group CRD might be clearer and more structured than a `data` in a ConfigMap. The `Backup` CR can reference the volume groups, instead of the ConfigMap.

If grouping and filtering can be tackled separately, the `Backup` CRD spec already has filter-related fields. We can change it from `[]string` to a struct to fulfill the `Conditions` requirement. We should keep https://github.com/vmware-tanzu/velero/pull/5842, which goes with #5838, separated from this proposal.

reasonerjt: It's not about grouping, it's about variety of actions, we wanna support actions other than "include/exclude".

As for putting the filter in or out of Backup CR, we wanna put it out of the CR for two reasons:
1. Control the size of backup CR.  We've been continuously adding new fields to the CRD to add new features, despite it's already `v1`, I hope we can reduce such changes, and will open another issue to discuss the `v1 -> v2` transformation of velero CRDs.
2. Make the policy re-usable across backups.  There has been some requirement from other adopters that they wanna introduce a `policy` CR and use them with different backups, this is a step towards that direction.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,270 @@
...
+type ResourcePolicies struct {
+	VolumePolicies []VolumePolicy `yaml:"volumePolicies"`
+	// we may support other resource policies in the future, and they could be added separately
+  	// OtherResourcePolicies: []OtherResourcePolicy
+}
+```
+
+The policies YAML config file would look like this:
+```yaml
+version: v1
```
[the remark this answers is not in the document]
reasonerjt: Simply put, it won't follow semver, this is simply for managing break change, please refer to the section "Multiple versions supporting".
cc @qiuming-best I think we should put more details of our discussion about the support scheme into this PR.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,270 @@
...
+	// Parameters defined map of parameters when executing a specific action
+	// +optional
+	// +nullable
+	Parameters map[string]interface{} `yaml:"parameters,omitempty"`
+}
+
+// VolumePolicy defined policy to conditions to match Volumes and related action to handle matched Volumes
+type VolumePolicy struct {
+	// Conditions defined list of conditions to match Volumes
+	Conditions map[string]interface{} `yaml:"conditions"`
```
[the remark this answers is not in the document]
reasonerjt: Are you suggesting we should introduce a 1:1 mapping to the PV, or put this chunk of data into backup CR?

IMO, 1:1 mapping may be sufficient but a more flexible condition map can be easier to create and maintain, I don't see a very strong reason we must keep this 1:1 mapping to the fileds of PV resource.

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: "include/exclude" could not further policy distinction when one backup both have `volume-snapshot` and `file-system-backup`, so `grouping` does not work, we introduced `actions` way

ihcsim: Policy re-use sounds like a good use case to me. If the ultimate goal is a policy CR, why not start with a CR right from the start? As for not adding new fields to backup, I think you will having to add something to it even if it's just a `ref` (or a list of `ref`s), right?

Regardless, can we keep and review #5842, and possibly add it to #5838, as I feel like it's different from the policy concept in this proposal? Thanks.

design/handle-backup-of-volumes-by-resources-filters.md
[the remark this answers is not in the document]
qiuming-best: I've updated `Versioning` and `Multiple versions supporting` parts.
which includes `additions`, `deprecation`, and `renaming` changes

reasonerjt (dismissed):
LGTM, thanks @qiuming-best for putting this together.

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,320 @@
...
+- backup.velero.io/must-include-additional-items
+ 
+So it should be careful with the combination of volumes resources policies and the above resources filters.
+- When volume resource policies conflict with the above resource filters, we should respect the above resource filters. For example, if the user used the opt-out approach to `backup.velero.io/backup-volumes-excludes` annotation on the pod and also defined include volume in volumes resources filters configuration, we should respect the opt-out approach to skip backup of the volume.
+- If volume resource policies conflict with themselves, the first matched policy will be respect.
+
+## Implementation
+This implementation should be included in Velero v1.11.0
+
+Currently, in Velero v1.11.0 we only support `Action`
```
shubham-pampattiwar, line 280: Should we also add backup level Validation where in we check whether the specified action in the policy is supported by velero install instance or not. For example, If the user specified `fs-backup` but the velero install does not have restic or kopia then in that case we could fail the backup even before we start processing it ? WYGT ?

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,320 @@
...
+- OrLabelSelectors
+- IncludeClusterResources
+- UseVolumeSnapshots
+- velero.io/exclude-from-backup=true
+- backup.velero.io/backup-volumes-excludes
+- backup.velero.io/backup-volumes
+- backup.velero.io/must-include-additional-items
+ 
+So it should be careful with the combination of volumes resources policies and the above resources filters.
+- When volume resource policies conflict with the above resource filters, we should respect the above resource filters. For example, if the user used the opt-out approach to `backup.velero.io/backup-volumes-excludes` annotation on the pod and also defined include volume in volumes resources filters configuration, we should respect the opt-out approach to skip backup of the volume.
```
shubham-pampattiwar, line 274: ++ Also, please be sure to add this in the velero documentation

shubham-pampattiwar (commented):
added a few comments

design/handle-backup-of-volumes-by-resources-filters.md
```diff
@@ -0,0 +1,320 @@
...
+The volume resources policies should contain a list of policies which is the combination of conditions and related `action`, when target volumes meet the conditions, the related `action` will take effection.
+
+Below is the API Design for the user configuration:
+
+### API Design
+```go
+// Action defined as one action for a specific way of backup
+type Action struct {
+	// Type defined specific type of action, it could be 'file-system-backup', 'volume-snapshot', or 'skip' currently
+	Type string `yaml:"type"`
```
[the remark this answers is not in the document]
qiuming-best: ok, I'll modify it

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: users, could config many kinds of `Action Type`,  they can config both `fs-backup` and `volume-snapshot` types of  `Action`. When doing the backup for one specific volume, we will first check which types of  `Action` the volume matched, if all policies are validated and returned it could use `fs-backup` and `volume-snapshot` both ok, and the current backup is through CSI plugin to take a snapshot, it will take a snapshot for `volume-snapshot` matched even though users have configured anther `fs-backup` for it.

If we add lots of validation related to the installation or environments it will make our validation complicated, and some redundant configurations by users are also acceptable.

design/handle-backup-of-volumes-by-resources-filters.md
qiuming-best: sure

shubham-pampattiwar (approved):
LGTM @qiuming-best

Issue #5750, opened 2023-01-09
CSI plugin should work with multiple VolumeSnapshotClasses


blackpiglet:
**Describe the problem/challenge you have**
[A description of the current limitation/problem/challenge that you are experiencing.]
By far, the Velero CSI plugin only works with one VolumeSnapshotClass, and the VolumeSnapshotClass is chosen by label `velero.io/csi-volumesnapshot-class`.
Let's work out a way to let the CSI plugin support multiple VolumeSnapshotClasses.
This issue #5737 give a using scenario of this requirement.

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Velero version (use `velero version`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

anshulahuja98:
Hi @blackpiglet
I am interested in taking this up. 
If no one is working on the proposal currently, can I pick it up?

blackpiglet:
@anshulahuja98 
Sure. Thanks for the contribution.

anshulahuja98:
@reasonerjt as requested in last community meeting, can we please add this design proposal to 1.11 milestone.

reasonerjt:
I'll double-check #5774 and if this is merged this will be fixed in v1.12

PR #5774, opened 2023-01-17
Design to add support for Multiple VolumeSnapshotClasses in CSI Plugin

anshulahuja98:

Proposal for #5750 

### DIFF
--- a/design/multiple-csi-volumesnapshotclass-support.md
+++ b/design/multiple-csi-volumesnapshotclass-support.md
@@ -87,10 +87,12 @@ The Velero CSI plugin chooses the VolumeSnapshotClass in the cluster that has th
             velero.io/csi-volumesnapshot-class: csi-diskdriver-snapclass
             
         ```
+    
+    Consider this as a override option in conjunction to part 1.
 
-    - The user has to annotate the PVCs or backups with the VolumeSnapshotClass to use for each driver. This is not ideal for the user experience.
-        - Mitigation: We can extend Velero CLI to also annotate backups/schedules with the VolumeSnapshotClass to use for each driver. This will make it easier for the user to annotate the backups/schedules. This mitigation is not for the PVCs though, since PVCs is anyways a specific use case. Similar to : " kubectl run --image myimage --annotations="foo=bar" --annotations="another=one" mypod"
-        We can add support for  - velero backup create my-backup --annotations "velero.io/csi:csi.cloud.disk.driver=csi-diskdriver-snapclass"
+**Note**: The user has to annotate the PVCs or backups with the VolumeSnapshotClass to use for each driver. This is not ideal for the user experience.
+    - **Mitigation**: We can extend Velero CLI to also annotate backups/schedules with the VolumeSnapshotClass to use for each driver. This will make it easier for the user to annotate the backups/schedules. This mitigation is not for the PVCs though, since PVCs is anyways a specific use case. Similar to : " kubectl run --image myimage --annotations="foo=bar" --annotations="another=one" mypod"
+    We can add support for  - velero backup create my-backup --annotations "velero.io/csi:csi.cloud.disk.driver=csi-diskdriver-snapclass"
 
 ### Stage 2 Approach
 The above annotations route is to get started and for initial design closure/ implementation, north star is to either introduce CSI specific fields (considering that CSI might be a very core part of velero going forward) in the backup/restore CR OR leverage the pluginInputs field as being tracked in: https://github.com/vmware-tanzu/velero/pull/5981
--- a/design/multiple-csi-volumesnapshotclass-support.md
+++ b/design/multiple-csi-volumesnapshotclass-support.md
@@ -1,0 +1,216 @@
+# Proposal to add support for Multiple VolumeSnapshotClasses in CSI Plugin
+
+- [Proposal to add support for Multiple VolumeSnapshotClasses in CSI Plugin](#proposal-to-add-support-for-multiple-volumesnapshotclasses-in-csi-plugin)
+	- [Abstract](#abstract)
+	- [Background](#background)
+	- [Goals](#goals)
+	- [Non Goals](#non-goals)
+    - [Detailed Design](#detailed-design)
+        - [Plugin Inputs Contract Changes](#plugin-inputs-contract-changes)
+        - [Using Plugin Inputs for CSI Plugin](#using-plugin-inputs-for-csi-plugin)
+        - [Annotations overrides on PVC for CSI Plugin](#annotations-overrides-on-pvc-for-csi-plugin)
+        - [Using Plugin Inputs for Other Plugins](#using-plugin-inputs-for-other-plugins)
+    - [Alternatives Considered](#alternatives-considered)
+    - [Security Considerations](#security-considerations)
+    - [Compatibility](#compatibility)
+    - [Implementation](#implementation)
+    - [Open Issues](#open-issues)
+
+
+## Abstract
+Currently the Velero CSI plugin chooses the VolumeSnapshotClass in the cluster that has the same driver name and also has the velero.io/csi-volumesnapshot-class label set on it. This global selection is not sufficient for many use cases. This proposal is to add support for multiple VolumeSnapshotClasses in CSI Plugin where the user can specify the VolumeSnapshotClass to use for a particular driver and backup.
+
+
+## Background
+The Velero CSI plugin chooses the VolumeSnapshotClass in the cluster that has the same driver name and also has the velero.io/csi-volumesnapshot-class label set on it. This global selection is not sufficient for many use cases. For example, if a cluster has multiple VolumeSnapshotClasses for the same driver, the user may want to use a VolumeSnapshotClass that is different from the default one. The user might also have different schedules set up for backing up different parts of the cluster and might wish to use different VolumeSnapshotClasses for each of these backups. 
+
+## Goals
+- Allow the user to specify the VolumeSnapshotClass to use for a particular driver and backup.
+
+## Non Goals
+-
+
+
+## Detailed Design
+
+### Plugin Inputs Contract Changes
+Approach is to introduce a new field `pluginInputs` in the velero CRs (Backup, Schedule, Restore). This field can be leveraged by all plugins for sending plugin specific settings rather than relying on annotations or global settings which hold across backups.
+
+```go
+type PluginInput struct {
+    Name string `json:"name"`
+    Properties map[string][string] `json:"properties"`
+}
+```
+
+### Using Plugin Inputs for CSI Plugin
+The user can specify the VolumeSnapshotClass to use for a particular driver and backup using the plugin inputs. The CSI plugin will use the VolumeSnapshotClass specified in the plugin inputs. If the VolumeSnapshotClass is not specified for a driver, the CSI plugin will use the default VolumeSnapshotClass for the driver fetched using labels through older route.
+
+Example: 
+```yaml
+apiVersion: velero.io/v1
+kind: Backup
+metadata:
+  name: backup-1
+spec:
+    pluginInputs:
+    - name: velero.io/csi
+    - properties:
+        - key: csi.cloud.disk.driver
+        - value: csi-diskdriver-snapclass
+        - key: csi.cloud.file.driver
+        - value: csi-filedriver-snapclass
+```
+
+CLI Example
+
+```bash
+velero backup create my-backup --plugin-inputs velero.io/csi:csi.cloud.disk.driver=csi-diskdriver-snapclass,csi.cloud.file.driver=csi-filedriver-snapclass
+```
+
+### Annotations overrides on PVC for CSI Plugin
+The user can annotate the PVCs with VolumeSnapshotClass name. This will override whatever the user has passed in pluginInputs for that driver.
+
+- If annotation is not present or VolumeSnapshotClass referred is not present in cluster OR if the specified VSC does not have the same CSI driver as the PVC
+    -  the CSI plugin will try to fallback to the pluginInputs value for that driver. If pluginInputs does not have the VSC for that driver, the CSI plugin will use the default VolumeSnapshotClass for the driver using the older label route.
+
+Example: 
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+name: pvc-1
+annotations:
+    velero.io/csi-volumesnapshot-class/disk.fg: csi-diskdriver-snapclass
+```
+
+### Using Plugin Inputs for Other Plugins
+As of today various plugins such as StorageClass mapping plugin, use a global configmap which is discovered through labels/ annotations which is not ideal customer experience and prohibhit usage of specific settings for each backup/ schedule setup.
+
+The pluginInputs field can be used to pass in the reference the configmap/ secret to use for a particular plugin. This will allow the user to use different configmaps/ secrets for different backups/ schedules.
+
+Example: 
+```yaml
+spec:
+    pluginInputs:
+    - name: velero.io/storageclass
+    - properties:
+        - key: storageclass-mapping-configmap
+        - value: configMapNamespace/configMapName
+```
+
+
+
+## Alternatives Considered
+
+1. **Through Annotations**
+    1. **Support VolumeSnapshotClass selection at PVC level**
+    The user can annotate the PVCs with driver and VolumeSnapshotClass name. The CSI plugin will use the VolumeSnapshotClass specified in the annotation. If the annotation is not present, the CSI plugin will use the default VolumeSnapshotClass for the driver. If the VolumeSNapshotClass provided is of a different driver, the CSI plugin will use the default VolumeSnapshotClass for the driver.
+
+        *example annotation on PVC:*
+        ```yaml
+        apiVersion: v1
+        kind: PersistentVolumeClaim
+        metadata:
+        name: pvc-1
+        annotations:
+            velero.io/csi-volumesnapshot-class: csi-diskdriver-snapclass
+            
+        ```
+
+    2. **Support VolumeSnapshotClass selection at backup/schedule level**
+    The user can  annotate the backup/ schedule with driver and VolumeSnapshotClass name. The CSI plugin will use the VolumeSnapshotClass specified in the annotation. If the annotation is not present, the CSI plugin will use the default VolumeSnapshotClass for the driver.
+
+        *example annotation on backup/schedule:*
+        ```yaml
+        apiVersion: velero.io/v1
+        kind: Backup
+        metadata:
+        name: backup-1
+        annotations:
+            velero.io/csi-volumesnapshot-class/csi.cloud.disk.driver: csi-diskdriver-snapclass
+            velero.io/csi-volumesnapshot-class/csi.cloud.file.driver: csi-filedriver-snapclass
+            velero.io/csi-volumesnapshot-class/<driver name>: csi-snapclass
+        ```
+
+         To query the annotations on a backup: "velero.io/csi-volumesnapshot-class/'driver name'" - where driver names comes from the PVC's driver.
+
+    **Limitations of Annotations approach**:
+    - The user has to annotate the PVCs or backups with the VolumeSnapshotClass to use for each driver. This is not ideal for the user experience.
+        - Mitigation: We can extend Velero CLI to also annotate backups/schedules with the VolumeSnapshotClass to use for each driver. This will make it easier for the user to annotate the backups/schedules. This mitigation is not for the PVCs though, since PVCs is anyways a specific use case.
+
+
+1. **Through CSI Specific Fields in Velero contracts**
+
+    **Considerations**
+    - Since CSI snapshotting is done through the plugin, we don't intend to bloat up the Backup Spec with CSI specific fields. 
+    - But considering that CSI Snapshotting is the way forward, we can debate if we should add a CSI section to the Backup Spec.
+
+
+    **Approach**: Similar to VolumeSnapshotLocation param in the Backup Spec, we can add a VolumeSnapshotClass param in the Backup Spec. This will allow the user to specify the VolumeSnapshotClass to use for the backup. The CSI plugin will use the VolumeSnapshotClass specified in the Backup Spec. If the VolumeSnapshotClass is not specified, the CSI plugin will use the default VolumeSnapshotClass for the driver.
+
+    *example of VolumeSnapshotClass param in the Backup Spec:*
+    ```yaml
+    apiVersion: velero.io/v1
+    kind: Backup
+    metadata:
+    name: backup-1
+    spec:
+        csiParameters: 
+            volumeSnapshotClasses: 
+                driver: csi.cloud.disk.driver
+                snapClass: csi-diskdriver-snapclass
+            timeout: 10m
+    ```
+
+1. **Through changes in velero contracts**
+    1. **Through configmap references.**
+    Currently even the storageclass mapping plugin expects the user to create a configmap which is used globally, and fetched through labels. This behaviour has same issue as the VolumeSnapshotClass selection. We can introduce a field in the velero contracts which allow passing configmap references for each plugin. And then the plugin can honour the configmap passed in as reference. The configmap can be used to pass the VolumeSnapshotClass to use for the backup, and also other parameters to tweak. This can help in making plugins more flexible while not depending on global behaviour.
+    
+
+        *example of configmap reference in the velero contracts:*
+        ```yaml
+        apiVersion: velero.io/v1
+        kind: Backup
+        metadata:
+        name: backup-1
+        spec:
+            configmapRefs:
+            - name: csi-volumesnapshotclass-configmap
+            - namespace: velero
+            - plugin: velero.io/csi
+         ```
+
+    2. **Through generic property bag in the velero contracts**: We can introduce a field in the velero contracts which allow passing a generic property bag for each plugin. And then the plugin can honour the property bag passed in.
+
+
+        *example of property bag in the velero contracts:*
+        ```yaml
+        apiVersion: velero.io/v1
+        kind: Backup
+        metadata:
+        name: backup-1
+        spec:
+            pluginInputs:
+            - name: velero.io/csi
+            - properties:
+                - key: csi.cloud.disk.driver
+                - value: csi-diskdriver-snapclass
+                - key: csi.cloud.file.driver
+                - value: csi-filedriver-snapclass
+        ```
+
+    **Note**: Both these approaches can also be used to tweak other parameters such as CSI Snapshotting Timeout/intervals. And further can be used by other plugins.
+
+
+## Security Considerations
+No security impact.
+
+## Compatibility
+Existing behaviour of csi plugin will be retained where it fetches the VolumeSnapshotClass through the label. This will be the default behaviour if the user does not specify the VolumeSnapshotClass.
+
+## Implementation
+TBD based on closure of high level design proposals.
+
+## Open Issues
+NA
--- a/design/multiple-csi-volumesnapshotclass-support.md
+++ b/design/multiple-csi-volumesnapshotclass-support.md
@@ -46,7 +46,7 @@ The Velero CSI plugin chooses the VolumeSnapshotClass in the cluster that has th
 
 ## Scenario 2
 - Bob is a cluster admin has PVCs storing different types of data. 
-- Most of the PVCs are used for storing non senstive application data. But certain PVCs store critical financial data.
+- Most of the PVCs are used for storing non-sensitive application data. But certain PVCs store critical financial data.
 - For such PVCs Bob wants to use a VolumeSnapshotClass with certain encryption related parameters set. 
 - In current velero, Bob can't achieve this as the CSI plugin will use the default VolumeSnapshotClass for the driver and all snapshots will be taken using the same VolumeSnapshotClass.
 - Proposed design will allow Bob to achieve this by overriding the VolumeSnapshotClass to use for a particular driver and backup/schedule using annotations on those specific PVCs.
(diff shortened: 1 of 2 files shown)


sseago:
For now, just a couple high level comments. I think we should not consider the explicit csi-specific fields on the Backup plugin. As long as the CSI plugin is implemented as an external plugin, we really don't want to embed specific knowledge of the plugin's internals in the core velero APIs.

Of the other two approaches, the annotation one is the most straightforward. If we go that way, we may want to allow annotations on both PVCs and Backup/Schedule CRs. That way a user can set one policy for the backup as a whole, allowing overrides per PVC. If we only do one of the two, probably the Backup level is most useful.

The "generic property bag" is the most interesting option, as it would allow other plugins to also add their own configuration parameters without having to rely on annotations or external config maps. It would be more work, but probably worthwhile if we can identify other existing plugin use cases that would also benefit from this. But the annotation approach is also perfectly acceptable, as there is precedent in using annotations for plugin configuration in other plugins.

blackpiglet:
@anshulahuja98 
Thanks for your contribution.
I prefer not to modify the Velero backup and schedule CRD to introduce the VolumeSnapshotClass chosen feature.
By far, this is only used for the CSI scenario, and CRD modification could make the Velero version upgrade process more complicated.
I think we can use the annotation on PVC and backup way, or the ConfigMap way. Velero can use some specific label or annotation to find out the correct ConfigMap.

anshulahuja98:
> @anshulahuja98 Thanks for your contribution. I prefer not to modify the Velero backup and schedule CRD to introduce the VolumeSnapshotClass chosen feature. By far, this is only used for the CSI scenario, and CRD modification could make the Velero version upgrade process more complicated. I think we can use the annotation on PVC and backup way, or the ConfigMap way. Velero can use some specific label or annotation to find out the correct ConfigMap.

"I think we can use the annotation on PVC and backup way, or the ConfigMap way. Velero can use some specific label or annotation to find out the correct ConfigMap."

@blackpiglet  what are your thoughts on introducing a generic Plugin parameter field in the backup/schedule CRD?

the issue with annotating configmaps is that they always end up remaining a global setting and hard to get determinism for what is to be used for each backup

Introducing another generic field in the CRD can help bring in that determinism and be extensible to other plugins where similar disambiguation of global settings is required.

Annotations are not a very friendly way from a regular customer POV. I believe we can't even use velero CLI for such scenarios.

We discussed this in the last Community meeting - there seemed to be general consensus over the generic plugin param approach
Would love to hear your thoughts more specifically around that

blackpiglet:
@anshulahuja98 
I agree that annotation is not an ideal method.
There is an ongoing PR that also enriches the backup function by adding advanced filter rules. #5773
It uses the ConfigMap way because the configuration could inflate big enough that makes the CR modification and creation have a notifiable latency, although your case may not be that obvious.
Prefer to split the advanced configuration into a separate element from Velero CRD, but also open to the community opinion.

anshulahuja98:
@blackpiglet I completely understand your point of the amount of inputs to be very high and being created as a separte configmap.

keeping the advanced configuration separate from CRD I also agree with.

The point I am trying to drive here is just that, the reference to that configmap instead of through annotations, we pass it in a generic input field.

For example: 
```yaml
apiVersion: velero.io/v1
kind: Backup
metadata:
name: backup-1
spec:
    pluginInputs:
    - name: velero.io/advanced-filter
    - properties:
        - key: advanceFilterConfigMapReference
        - value: configMapNamespace/configMapName
```

In above example, the advance filter code will know which value to look for in the pluginInputs field, whicch key to search on to get the configmap reference.
TOday we might be doing things like this through annotations which are a pain to manage.


Similarly if let's say the VolumeSNapshotClassInputs increase in size we can pass reference to configmap.


```yaml
apiVersion: velero.io/v1
kind: Backup
metadata:
name: backup-1
spec:
    pluginInputs:
    - name : velero.io/csi
    - properties:
        - key: snapshotClassMappingConfigMapReference
        - value: configMapNamespace/configMapName
```


This can further be extended to pass in StorageClass mapping configmap

blackpiglet:
@anshulahuja98 
I totally agree with you that the annotation way is not an ideal way, and the proposed method is a better alternative.
But I'm not sure whether we should also provide the annotation way for users.
Just my thought, maybe given both annotation and CRD reference ways are better.

anshulahuja98:
I agree with your point @blackpiglet , we can do both annotations and CRD way.

WIll keep CRD route to be P0, and annotations P1, should not take too much effort to light up annotations as well.

We anyways need annotations if we wish to support VSC support at per PVC level. For overrides of global settings.
I'll work towards a more granular proposal now covering only things which we need to work on and remove other approaches from the doc.

Thank you for your comments and initial review!

anshulahuja98:
Meanwhile are there any thoughts on how to expose the input of such "plugin inputs" through the CLI?
Any top of mind concerns/ thoughts on what might be a better user experience.
CC: @blackpiglet  / @sseago

anshulahuja98:
CC @reasonerjt / @qiuming-best for review.
If we can take any inspiration from here for the PR: https://github.com/vmware-tanzu/velero/pull/5773

anshulahuja98:
Have updated the PR with the final design based on inputs.
Pl review @sseago  / @blackpiglet

blackpiglet:
Please add a changelog file to this PR, and fix the typo to make the Github action check to pass.

For the "plugin inputs" setting part, the Velero team had some discussion. Since there is no procedure for the Velero CRD update, we don't prefer to modify the Velero CRDs in v1.11. Because, if the CRD is changed, after the Velero version upgrade, the Velero image is not changed accordingly, it will complain fail to find updated CRD in cluster.

anshulahuja98:
@blackpiglet 
I believe we have the upgrade-crds Job, isn't that enough in most cases to ensure CRD upgrades.

Major version should be enough to push through a CR update, I see that has happened often in the past. Is there anything specific about 1.11 due to which we don't wish to change them?

Also does this mean that PRs such as https://github.com/vmware-tanzu/velero/pull/5838 also blocked for 1.11 since they have CRD changes?

For steps forward do you wish for this design to be checked in as is and worked upon post 1.11. Or come up with alternative proposal?

sseago:
While the new resource filtering mentioned above (with CRD changes) won't make it into 1.11, there will be CRD changes for the new async BIA/RIA plugin operations, some of which are already merged on main, so I don't think that "we don't want CRD changes for 1.11" would really apply here. For z streams (1.10.1, etc.), we definitely want to minimize, hopefully completely eliminate CRD changes, but for releases like 1.8, 1.9, 1.10, 1.11, etc. some CRD changes are always expected. The main concern is making sure that the changes are backwards-compatible, especially with regard to existing backups/PVBs, etc.

blackpiglet:
Right. Although it's better to limit the CRD change, I also found it hard to avoid CRD change, so the CRD change is not a blocker of the PR.

For the `pluginInput` format, IMO, it's better to integrate configurations into one ConfigMap. What's your opinion?

anshulahuja98:
> Right. Although it's better to limit the CRD change, I also found it hard to avoid CRD change, so the CRD change is not a blocker of the PR.
> 
> For the `pluginInput` format, IMO, it's better to integrate configurations into one ConfigMap. What's your opinion?


Even if it's one Configmap, we still need a field to pass in that input of which coonfigmap to use which anyways needs a CRD change

For plugins which are lightweight and do not require heavy inputs, the inputs can passed in the generic property bag itself similar to BSL params or VolumeSnapshotClass params

For plugins which require heavy inputs, they can take input of a configmap reference (namespace / name) and parse it to fetch the config map.  Similar to the example 
image


In short the design does support the configmap route. While being generic to type of input.

In future plugins may need reference of other objects such as secrets etc. THis extensible design takes care of all scenarios.


Would also request you to go through the alternatives considered section in the design.

I will fix the PR with changelog / typo shortly. Hoping to go through the final review meanwhile.
Please add any other folks which can help reviewing/signing off

anshulahuja98:
@shubham-pampattiwar can you please review this PR since you have more CSI related context.

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Background
+The Velero CSI plugin chooses the VolumeSnapshotClass in the cluster that has the same driver name and also has the velero.io/csi-volumesnapshot-class label set on it. This global selection is not sufficient for many use cases. For example, if a cluster has multiple VolumeSnapshotClasses for the same driver, the user may want to use a VolumeSnapshotClass that is different from the default one. The user might also have different schedules set up for backing up different parts of the cluster and might wish to use different VolumeSnapshotClasses for each of these backups. 
+
+## Goals
+- Allow the user to specify the VolumeSnapshotClass to use for a particular driver and backup.
+
+## Non Goals
+-
+
```
shubham-pampattiwar, line 32: Please add a use-case section with 1-2 user stories so that utility of this feature becomes more clear.

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,216 @@
...
+Currently the Velero CSI plugin chooses the VolumeSnapshotClass in the cluster that has the same driver name and also has the velero.io/csi-volumesnapshot-class label set on it. This global selection is not sufficient for many use cases. This proposal is to add support for multiple VolumeSnapshotClasses in CSI Plugin where the user can specify the VolumeSnapshotClass to use for a particular driver and backup.
+
+
+## Background
+The Velero CSI plugin chooses the VolumeSnapshotClass in the cluster that has the same driver name and also has the velero.io/csi-volumesnapshot-class label set on it. This global selection is not sufficient for many use cases. For example, if a cluster has multiple VolumeSnapshotClasses for the same driver, the user may want to use a VolumeSnapshotClass that is different from the default one. The user might also have different schedules set up for backing up different parts of the cluster and might wish to use different VolumeSnapshotClasses for each of these backups. 
+
+## Goals
+- Allow the user to specify the VolumeSnapshotClass to use for a particular driver and backup.
+
+## Non Goals
```
shubham-pampattiwar, line 30: IMO we should make it clear and add a `Non-Goal` that the current approach will not be changed and will be still in use.

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,216 @@
...
+CLI Example
+
+```bash
+velero backup create my-backup --plugin-inputs velero.io/csi:csi.cloud.disk.driver=csi-diskdriver-snapclass,csi.cloud.file.driver=csi-filedriver-snapclass
+```
+
+### Annotations overrides on PVC for CSI Plugin
+The user can annotate the PVCs with VolumeSnapshotClass name. This will override whatever the user has passed in pluginInputs for that driver.
+
+- If annotation is not present or VolumeSnapshotClass referred is not present in cluster OR if the specified VSC does not have the same CSI driver as the PVC
```
shubham-pampattiwar, line 74: +1 on the proposed behaviour

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,216 @@
...
+        - key: csi.cloud.disk.driver
+        - value: csi-diskdriver-snapclass
+        - key: csi.cloud.file.driver
+        - value: csi-filedriver-snapclass
+```
+
+CLI Example
+
+```bash
+velero backup create my-backup --plugin-inputs velero.io/csi:csi.cloud.disk.driver=csi-diskdriver-snapclass,csi.cloud.file.driver=csi-filedriver-snapclass
```
shubham-pampattiwar, line 68: Maybe add a note so that we update docs on how we are using the delimiters for the `pluginsInputs` in velero CLI

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,216 @@
...
+metadata:
+name: pvc-1
+annotations:
+    velero.io/csi-volumesnapshot-class/disk.fg: csi-diskdriver-snapclass
+```
+
+### Using Plugin Inputs for Other Plugins
+As of today various plugins such as StorageClass mapping plugin, use a global configmap which is discovered through labels/ annotations which is not ideal customer experience and prohibhit usage of specific settings for each backup/ schedule setup.
+
+The pluginInputs field can be used to pass in the reference the configmap/ secret to use for a particular plugin. This will allow the user to use different configmaps/ secrets for different backups/ schedules.
```
shubham-pampattiwar, line 90: What if both the global cm as well as the pluginInputs cm exist ?

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,216 @@
...
+```go
+type PluginInput struct {
+    Name string `json:"name"`
+    Properties map[string][string] `json:"properties"`
+}
+```
+
+### Using Plugin Inputs for CSI Plugin
+The user can specify the VolumeSnapshotClass to use for a particular driver and backup using the plugin inputs. The CSI plugin will use the VolumeSnapshotClass specified in the plugin inputs. If the VolumeSnapshotClass is not specified for a driver, the CSI plugin will use the default VolumeSnapshotClass for the driver fetched using labels through older route.
+
```
shubham-pampattiwar, line 48: I believe the main downside of this approach would be that we would be increasing the size of the backup CR

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,216 @@
...
+Example: 
+```yaml
+apiVersion: velero.io/v1
+kind: Backup
+metadata:
+  name: backup-1
+spec:
+    pluginInputs:
+    - name: velero.io/csi
```
shubham-pampattiwar, line 57: I wonder would it make sense to add another spec called `type` (plugintype) along the level of `name` spec.

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Abstract
+Currently the Velero CSI plugin chooses the VolumeSnapshotClass in the cluster that has the same driver name and also has the velero.io/csi-volumesnapshot-class label set on it. This global selection is not sufficient for many use cases. This proposal is to add support for multiple VolumeSnapshotClasses in CSI Plugin where the user can specify the VolumeSnapshotClass to use for a particular driver and backup.
+
+
+## Background
+The Velero CSI plugin chooses the VolumeSnapshotClass in the cluster that has the same driver name and also has the velero.io/csi-volumesnapshot-class label set on it. This global selection is not sufficient for many use cases. For example, if a cluster has multiple VolumeSnapshotClasses for the same driver, the user may want to use a VolumeSnapshotClass that is different from the default one. The user might also have different schedules set up for backing up different parts of the cluster and might wish to use different VolumeSnapshotClasses for each of these backups. 
+
+## Goals
+- Allow the user to specify the VolumeSnapshotClass to use for a particular driver and backup.
```
shubham-pampattiwar, line 28: if we are going ahead with approach of `pluginInputs` then the goal should also add that this design enables the user to specify plugin specific data/params or something on similar lines.

design/multiple-csi-volumesnapshotclass-support.md
shubham-pampattiwar, line 57: Maybe an implementation detail, but we should add a backup validation as well when pluginInputs are specified like they should only be accepted only if the corresponding plugin is registered with the velero controller.

design/multiple-csi-volumesnapshotclass-support.md
anshulahuja98: Ack. Will add.

design/multiple-csi-volumesnapshotclass-support.md
anshulahuja98: Ack. Will add.

design/multiple-csi-volumesnapshotclass-support.md
anshulahuja98: Ack. Will add.

design/multiple-csi-volumesnapshotclass-support.md
anshulahuja98: here I have used name for the plugin Name
Isn't plugin type same as plugin name?

anshulahuja98: Makes sense. Will check feasibility of such check in code and add it.

design/multiple-csi-volumesnapshotclass-support.md
anshulahuja98: Similar to the CSI driver case
where we use pluginInputs where available and fallback on global cm for other cases.
Guidance will be such that all plugins should try to honour first the params in pluginInput(be it explicit params such as VSC, or pointer to CM) if it is not able to get required info from pluginInput, fallback will be global CM

I will add this guidance.

design/multiple-csi-volumesnapshotclass-support.md
anshulahuja98: What all concerns emerge from this? Have we seen any issues earlier due to this?

design/multiple-csi-volumesnapshotclass-support.md
sseago: @shubham-pampattiwar @anshulahuja98  Actually, I'm not 100% sure we need that validation. Should backups with plugin inputs fail if the plugin isn't registered? In any case, if you want this validation you'll need PluginKind. `clientmgmt.Manager` does have methods for retrieving registered plugins by name, but there is no grand unified list of plugins. Each PluginKind has a separate list of registered plugins. However, since we're talking about plgins for a backup or for a restore) here, we should already know the plugin kind. We're only dealing with BackupItemAction plugin inputs here. We may want to eventually support volumesnapshotter inputs too, but maybe that's a separate spec section. Call this `biaPluginInputs` instead of `pluginInputs` (and `riaPluginInputs` on restore), and then we already know the plugin kind. It would be `BackupItemAction` in the context of velero 1.10, but since this won't be implemented until post-1.11, it would be `BackupItemActionV2` Basically, validation could call `pluginManager.GetBackupItemActionV2(input.Name)` and if it errors out, then the plugin isn't registered.

sseago: @shubham-pampattiwar I'm thinking that we're dealing with limited types here, so instead of a type field, lets have a single spec field for backup item actions. If  we need to support this for volume snapshotters (and we may not, since there's not really a lot of active development there) we can add a separate field for that later. Velero already calls BIA and VolumeSnapshotter plugins in completely different places, so it probably makes more sense to separate the configuration fields too.

anshulahuja98: I think it makes sense to evaluate if we really need validation
We can stage it if needed.

design/multiple-csi-volumesnapshotclass-support.md
blackpiglet: Not from the CR size, but the Restore CRD size was big before, and in some low-bandwidth environments, Velero installation fails due to API-server connection timeout.

anshulahuja98: I see.. 
We could consider reducing the comments/ description per CRD field as well in those cases.

shubham-pampattiwar (commented):
Overall the design/use-case looks sane to me but I would like to know thoughts from other maintainers as well ? 
@anshulahuja98 There might be some intersection with the ongoing design regarding Volume Policies (https://github.com/vmware-tanzu/velero/pull/5773) 
I am just spit-balling here but can we not integrate your use case with the Volume Policies design ? like the `Condition` would be the CSI driver and `Action.Type` could be `Override` and `Action.Parameters` could consist of the VolumeSnapshotClass that you want to use for those volumes. 

(cc @blackpiglet @Lyndon-Li @sseago @reasonerjt @qiuming-best )

anshulahuja98:
> Overall the design/use-case looks sane to me but I would like to know thoughts from other maintainers as well ? @anshulahuja98 There might be some intersection with the ongoing design regarding Volume Policies (#5773) I am just spit-balling here but can we not integrate your use case with the Volume Policies design ? like the `Condition` would be the CSI driver and `Action.Type` could be `Override` and `Action.Parameters` could consist of the VolumeSnapshotClass that you want to use for those volumes.
> 
> (cc @blackpiglet @Lyndon-Li @sseago @reasonerjt @qiuming-best )

I understand where you are coming from on the intersection. 
But I feel VolumeSnapshotting is a very core functionality with CSI being the path forward. And VSC param selection is a regular flow, as compared to fine grained volume control which might be used by niche set of customers.

The volumefilter approach seems very involved for a customer to setup as compared to having dedicated entries for CSI VSCs. 
Additionally they will be 2 separate plugins.

However I feel that the VOlume Policy Design could benefit from the pluginInputs field that I am proposing. THey could leverage the same.

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,239 @@
...
+
+Example: 
+```yaml
+apiVersion: velero.io/v1
+kind: Backup
+metadata:
+  name: backup-1
+spec:
+    pluginInputs:
+    - name: velero.io/csi
```
reasonerjt, line 78: I'm not sure if it's right thing to do to expose plugin name to the end user, that couples the interface to the implementation.

For example, if we choose to move CSI plugin into the main backup flow (not likely to happen though), it will be confusing if we let user select the vsclass via plugin input.

I would rather add a `CSI` field into the spec of backup and let the user choose it explicitly.

anshulahuja98: On the point of "couples the interface to the implementation.", even today customer has to understand which plugins to enable to add the appropriate plugin containers to the deployment. User is already in some sense expected to be aware of which component is providing which functionality.

If CSI is merged with current codebase, we would have to anyways educated brownfield customers to migrate from using the CSI plugin containers in their deployments. This change is much bigger than just honoring / not honoring the values in the pluginInputs.

The overall purpose of this spec is to introduce pluginInputs which can be leverage by other plugins without having to bloat up the backup CR, this approach is mostly plugin agnostic. CSI related functionalities are just a part of the requirement.


For CSI you might agree on backup spec, but I don't think for features such as storageclass mapping this approach would make sense.
Any suggestions as to how those scenarios can be handled?

In the alternatives consdiered below you can check other approaches we evaluated.
Introducing CSI specific fields in backup CR turned to be strong no from discussion.

reasonerjt: > The overall purpose of this spec is to introduce pluginInputs which can be leverage by other plugins without having to bloat up the backup CR

I agree we don't want to bloat up the backup CR, that's why we want a balance between customizability and complexity.  At this moment, I'm still not quite convinced we should provide a general framework so that users can customize the behavior of any plugin.

> For CSI you might agree on backup spec, but I don't think for features such as storageclass mapping this approach would make sense.
> Any suggestions as to how those scenarios can be handled?

As for `storageclass mapping`, if you are talking about modifying the resources before they are restored, user can write RIA, if we are not happy with RIA, I think the JSON-substitution proposal may be helpful, though it may not have to be an RIA.

anshulahuja98: "I agree we don't want to bloat up the backup CR, that's why we want a balance between customizability and complexity. At this moment, I'm still not quite convinced we should provide a general framework so that users can customize the behavior of any plugin."

Users are anyways required to do this through annotations for certain plugins - the main issue is that often these are often set as global behaviour impacting all backup/restores. Additionally annotations are not a user friendly approach to this.

For storage class mapping the existing RIA is dependent on global behavior. It fetches configmap on basis of certain label / annotation. The same issue we are facing in VolumeSnapshotClass selection. 
Expecting the end user to write plugins for each scenario, understand velero code, maintaining the RIA code seems like a very big overhead.

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,177 @@
...
+        name: backup-1
+        annotations:
+            velero.io/csi-volumesnapshot-class/csi.cloud.disk.driver: csi-diskdriver-snapclass
+            velero.io/csi-volumesnapshot-class/csi.cloud.file.driver: csi-filedriver-snapclass
+            velero.io/csi-volumesnapshot-class/<driver name>: csi-snapclass
+        ```
+
+         To query the annotations on a backup: "velero.io/csi-volumesnapshot-class/'driver name'" - where driver names comes from the PVC's driver.
+
+    2. **Support VolumeSnapshotClass selection at PVC level**
```
reasonerjt, line 77: So this implies velero should always use the same vsclass to take snapshot of `PVCxxx`

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,177 @@
...
+    - **Mitigation**: We can extend Velero CLI to also annotate backups/schedules with the VolumeSnapshotClass to use for each driver. This will make it easier for the user to annotate the backups/schedules. This mitigation is not for the PVCs though, since PVCs is anyways a specific use case. Similar to : " kubectl run --image myimage --annotations="foo=bar" --annotations="another=one" mypod"
+    We can add support for  - velero backup create my-backup --annotations "velero.io/csi:csi.cloud.disk.driver=csi-diskdriver-snapclass"
+
+### Stage 2 Approach
+The above annotations route is to get started and for initial design closure/ implementation, north star is to either introduce CSI specific fields (considering that CSI might be a very core part of velero going forward) in the backup/restore CR OR leverage the pluginInputs field as being tracked in: https://github.com/vmware-tanzu/velero/pull/5981
+
+Refer section Alternatives 2. **Through generic property bag in the velero contracts**: in the design doc for more details on the pluginInputs field.
+
+
+## Alternatives Considered
```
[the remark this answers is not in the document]
anshulahuja98: I believe CSI snapshotting is a very common use case as compared to resource filters for volumes which are more specific in nature of their purpose.
And hence from a better customer experience POV, I'll prefer an annotations approach.

For performing basic snapshotting operations, user should not have to create and manage configmaps.

design/multiple-csi-volumesnapshotclass-support.md
```diff
@@ -0,0 +1,177 @@
...
+    The user can  annotate the backup/ schedule with driver and VolumeSnapshotClass name. The CSI plugin will use the VolumeSnapshotClass specified in the annotation. If the annotation is not present, the CSI plugin will use the default VolumeSnapshotClass for the driver.
+
+        *example annotation on backup/schedule:*
+        ```yaml
+        apiVersion: velero.io/v1
+        kind: Backup
+        metadata:
+        name: backup-1
+        annotations:
+            velero.io/csi-volumesnapshot-class/csi.cloud.disk.driver: csi-diskdriver-snapclass
```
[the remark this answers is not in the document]
anshulahuja98: Yes, that is correct.

design/multiple-csi-volumesnapshotclass-support.md
anshulahuja98: Yes, consider this an override behavior for a PVC. 
In case user has configured VSC at backup level, this will override the VSC for the particular PVC.

I am not sure if you this comment something else?

anshulahuja98:
@reasonerjt can you please add this to the 1.12 milestone.

draghuram:
Coming a bit late to the party. I would really appreciate it if someone can summarize the change planned for 1.12 as far as this feature is concerned. Has it been decided to implement annotations at both backup and PVC level? I did try to go through all the comments but it is not easy to process as there are so many of them. 

Just for reference, I would like to highlight few points:

1. It is not entirely true that CSI is implemented only through the plug-in. There is some CSI implementation in the core Velero as well. Add to this the fact that CSI is the future of storage in Kubernetes so there should not be a problem in introducing CSI specific fields in any CR. I am not saying we should but I don't think it should be summarily dismissed. In fact, I don't see much point in having CSI plug-in installed separately as snapshots are central to backups and including it in the core reduces scope for error (such as version mismatch).
2. Having annotations at PVC level or other Kubernetes objects level can lead to trouble as they have unintended consequences. Imagine a PVC included in two different backups in which case the annotation will take effect for both backups which may not be what the user intended. I have recently seen a user implementing admission webhook to remove any FSB annotations added on PVCs because he didn't want any inadvertent addition of annotations messing up his backups. Annotations at Backup level or some other way of specifying in Backup CR is fine though.

sseago (approved):

shubham-pampattiwar (approved):
LGTM !

anshulahuja98:
Thank you for approving the PR folks
Can one of you please help with the merge ? 
@shubham-pampattiwar  / @blackpiglet  / @sseago

achyut-deshpande:
Is it testable in v1.12 RC1?

anshulahuja98:
yes @achyut-deshpande you can use [v0.6.0-rc.1](https://github.com/vmware-tanzu/velero-plugin-for-csi/releases/tag/v0.6.0-rc.1) release of CSI plugin to test out this functionality. Please use it in conjunction with 1.12 RC1 of core velero.

anshulahuja98:
Hi @achyut-deshpande we fixed this issue later

	"velero.io/csi-volumesnapshot-class_bar.csi.k8s.io": "bar2",
	
	use an "_" instead of the "/"
	
	https://github.com/vmware-tanzu/velero-plugin-for-csi/pull/193 this got updated as part of this PR and we missed updating the design doc it seems like.

draghuram:
I checked CSI section of Velero docs and it still mentions single volume snapshot class. Is there a docs PR for this feature?

anshulahuja98:
@draghuram missed the docs PR due to time crunch.
Let me add now
I see overall 3 things - updating design doc, updating CSI plugin README, updating docs.

anshulahuja98:
CSI README was already updated with basic docs, just need the bug fix
https://github.com/vmware-tanzu/velero-plugin-for-csi/blob/main/README.md

PR #5744, opened 2023-01-06
Resolve staticcheck linter reported issues

chaunceyjiang:
Signed-off-by: chaunceyjiang <chaunceyjiang@gmail.com>

Thank you for contributing to Velero!

# Please add a summary of your change

Please see https://golang.org/doc/go1.16#ioutil

Resolve staticcheck linter reported issues.

# Does your change fix a particular issue?

Part of #5734
# Please indicate you've done the following:

- [x] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [x] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required` as a comment on this pull request.
- [ ] Updated the corresponding documentation in `site/content/docs/main`.

PR #5775, opened 2023-01-17
cleanup: remove ioutil for go version

my-git9:
Remove ioutil for go version: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-16


blackpiglet:
@my-git9 
Thanks for your contribution, but there is already another PR that does similar work #5744.

Issue #5684, opened 2022-12-13
Kopia unsupported files error not showing full details


alexyao2015:
**Describe the problem/challenge you have**
When backing up a PV using Kopia, I received an error message that was very vague and does not appear to show the full detail for why the backup failed.

(new lines added for easier readability)
```
time="2022-12-13T22:33:07Z" level=error msg="Error backing up item" backup=velero/test-backup-20221213071957
error="pod volume backup failed: running backup, stderr=Failed to run kopia backup: unknown or unsupported entry type: 
Failed to run kopia backup: unknown or unsupported entry type" error.file="/go/src/github.com/vmware-tanzu/velero/pkg/podvolume/backupper.go:223" 
error.function="github.com/vmware-tanzu/velero/pkg/podvolume.(*backupper).BackupPodVolumes" 
logSource="pkg/backup/backup.go:425" name=test-c4dc486fb-cprbs
```

After some research, I found that one of my PVs had a unix socket within the directory to be backed up which of course isn't able to be backed up, however the original error message did not mention this and I could only guess from some searching about the text. In my research, it appears that the original text includes the exact filename of the object that failed (https://kopia.discourse.group/t/snapshot-unknown-or-unsupported-entry-type/1109), but this text is missing from the logs.

**Describe the solution you'd like**
1. The log messages should show the full error messages from Kopia to more clearly diagnose errors.
2. The usage of `.kopiaignore` should be added to the documentation or at least the Kopia documentation regarding this should be linked.
3. An option to simply ignore all files that cannot be backed up (e.g. unix sockets).

**Environment:**

- Velero version (use `velero version`): 1.10.0
- Kubernetes version (use `kubectl version`): 1.21

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

Lyndon-Li:
> The log messages should show the full error messages from Kopia to more clearly diagnose errors.

Kopia is delivering the error to the uploader's Progress endpoint:
`u.Progress.Error(entryRelativePath, rc, isIgnored)`
so look like we need to handle Progress.Error function and then transfer it to Velero log

[...]

PR #5786, opened 2023-01-19
Set Kopia IgnoreUnknownTypes in ErrorHandlingPolicy to True

qiuming-best:
Set Kopia IgnoreUnknownTypes in ErrorHandlingPolicy to True for ignoring backup unknown or unsupported file type

Signed-off-by: Ming <mqiu@vmware.com>

Fixes #(issue)
https://github.com/vmware-tanzu/velero/issues/5684

### DIFF
--- a/pkg/uploader/kopia/progress.go
+++ b/pkg/uploader/kopia/progress.go
@@ -20,6 +20,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/vmware-tanzu/velero/pkg/uploader"
 )
 
@@ -63,6 +65,7 @@ type KopiaProgress struct {
 	processedBytes int64                    // which statistic all bytes has been processed currently
 	outputThrottle Throttle                 // which control the frequency of update progress
 	Updater        uploader.ProgressUpdater //which kopia progress will call the UpdateProgress interface, the third party will implement the interface to do the progress update
+	Log            logrus.FieldLogger       // output info into log when backup
 }
 
 //UploadedBytes the total bytes has uploaded currently
@@ -77,8 +80,10 @@ func (p *KopiaProgress) UploadedBytes(numBytes int64) {
 func (p *KopiaProgress) Error(path string, err error, isIgnored bool) {
 	if isIgnored {
 		atomic.AddInt32(&p.ignoredErrorCount, 1)
+		p.Log.Warnf("Ignored error when processing %v: %v", path, err)
 	} else {
 		atomic.AddInt32(&p.fatalErrorCount, 1)
+		p.Log.Errorf("Error when processing %v: %v", path, err)
 	}
 }
 
--- a/pkg/uploader/kopia/snapshot.go
+++ b/pkg/uploader/kopia/snapshot.go
@@ -18,6 +18,7 @@ package kopia
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"math"
 	"os"
@@ -61,8 +62,14 @@ type SnapshotUploader interface {
 	) (*snapshot.Manifest, error)
 }
 
-func newOptionalInt(b policy.OptionalInt) *policy.OptionalInt {
-	return &b
+func newOptionalInt(b int) *policy.OptionalInt {
+	ob := policy.OptionalInt(b)
+	return &ob
+}
+
+func newOptionalBool(b bool) *policy.OptionalBool {
+	ob := policy.OptionalBool(b)
+	return &ob
 }
 
 //setupDefaultPolicy set default policy for kopia
@@ -75,10 +82,13 @@ func setupDefaultPolicy(ctx context.Context, rep repo.RepositoryWriter, sourceIn
 			CompressorName: "none",
 		},
 		UploadPolicy: policy.UploadPolicy{
-			MaxParallelFileReads: newOptionalInt(policy.OptionalInt(runtime.NumCPU())),
+			MaxParallelFileReads: newOptionalInt(runtime.NumCPU()),
 		},
 		SchedulingPolicy: policy.SchedulingPolicy{
 			Manual: true,
+		},
+		ErrorHandlingPolicy: policy.ErrorHandlingPolicy{
+			IgnoreUnknownTypes: newOptionalBool(true),
 		},
 	})
 }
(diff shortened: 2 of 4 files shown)


pkg/uploader/kopia/snapshot.go
```diff
@@ -80,6 +85,9 @@ func setupDefaultPolicy(ctx context.Context, rep repo.RepositoryWriter, sourceIn
 		SchedulingPolicy: policy.SchedulingPolicy{
 			Manual: true,
 		},
+		ErrorHandlingPolicy: policy.ErrorHandlingPolicy{
+			IgnoreUnknownTypes: newOptionalBool(true),
```
Lyndon-Li, line 89: Ignoring errors arbitrarily is dangerous, we cannot set it to `true` by default.
If we want to support to ignore unknown errors, the option should be disabled by default and we only enable it from a users' option.
Therefore, we have to create a user interface for this feature.

Finally, personally I think this option is not Must Have. As long as we can capture the error message and report the users(as for ErrorHandlingPolicy) it is good enough.

pkg/uploader/kopia/progress.go
```diff
@@ -77,11 +79,17 @@ func (p *KopiaProgress) UploadedBytes(numBytes int64) {
 func (p *KopiaProgress) Error(path string, err error, isIgnored bool) {
 	if isIgnored {
 		atomic.AddInt32(&p.ignoredErrorCount, 1)
+		p.errs = append(p.errs, fmt.Errorf("Ignored error when processing %v: %v", path, err))
```
Lyndon-Li, line 82: Since `p.errs` is just for log printing in line `131@func (kp *kopiaProvider) RunBackup`, why don't we print the log in place here?
If we print the log at the end of the backup, the meaning of the time for each log will be eliminated. 
Instead, if we print the log in place, from the time of the log, we know when the error happens, which is even useful for a long running backup.

pkg/uploader/provider/kopia.go
```diff
@@ -119,15 +119,20 @@ func (kp *kopiaProvider) RunBackup(
...
+	progress := new(kopia.KopiaProgress)
+	progress.InitThrottle(backupProgressCheckInterval)
+	progress.Updater = updater
+	kpUploader.Progress = progress
 	quit := make(chan struct{})
 	log.Info("Starting backup")
 	go kp.CheckContext(ctx, quit, nil, kpUploader)
 
 	defer func() {
+		snapshotErrs := progress.GetErrors()
```
Lyndon-Li, line 131: See https://github.com/vmware-tanzu/velero/pull/5786/files#r1081073132

pkg/uploader/kopia/progress.go
Lyndon-Li: Meanwhile, p.ignoredErrorCount and p.fatalErrorCount looks a good candidate to print by the end of the backup, e.g., in line `131@func (kp *kopiaProvider) RunBackup`

pkg/uploader/kopia/snapshot.go
```diff
@@ -212,19 +220,24 @@ func SnapshotSource(
...
 	manifestID := manifest.ID
 	snapSize := manifest.Stats.TotalFileSize
 
+	var ErrUnknown = "unknown or unsupported entry type"
+
 	var errs []string
 	if ds := manifest.RootEntry.DirSummary; ds != nil {
 		for _, ent := range ds.FailedEntries {
-			errs = append(errs, ent.Error)
+			if !(*policyTree.DefinedPolicy().ErrorHandlingPolicy.IgnoreUnknownTypes == true && strings.Contains(ent.Error, ErrUnknown)) {
```
Lyndon-Li, line 235: If we really want to ignore any error, we should log something to record that the error has been ignored, otherwise, we will not be able to troubleshoot.

pkg/uploader/kopia/snapshot.go
qiuming-best: In fact, the unknown type of files is not regular files at the filesystem level, including setuid/gid, hardlinks, mount points, and UNIX socket. both of them are not usable after restore even if Kopia could backup and users also won't expect these could work. so it's not dangerous if we could show the ignored files as a warning in the backup CR maybe that is enough.

pkg/uploader/kopia/progress.go
qiuming-best: it would be not suitable to output logs in basic public functions and structures, but we could add a timestamp to the error.

qiuming-best: In fact, I debug with the Kopia client, and through the error output I found the related [Error function ](https://github.com/kopia/kopia/blob/v0.10.4/cli/cli_progress.go#L90).

And our Error in KopiaProgress is imitated from it.

pkg/uploader/kopia/snapshot.go
qiuming-best: https://github.com/vmware-tanzu/velero/pull/5786#discussion_r1081155301

pkg/uploader/kopia/snapshot.go
Lyndon-Li: Checked Kopia code further, this IgnoreUnknownTypes option is checked for the file system entries that are not recognized by Kopia uploader. Kopia uploader will fail to back up these entries inevitably and IgnoreUnknownTypes is only to control whether the uploader error out or ignore and process.
However, it doesn't mean these entries are not useful, after all, we even don't know what they are exactly (Kopia at present supports dirs, files and symlinks only, other entry types all go to the UnknownTypes, a.k.a., ErryEntries). Instead, it only means Kopia uploader doesn't support them, but since we are using Kopia uploader, we don't have any way to support them if the uploader doesn't support.

For Kopia itself, it by default sets IgnoreUnknownTypes to true and gives users options to modify it. For Velero:

- If we set it to false, we will face some embarrassment that  Kopia util doesn't fail itself but Velero fails
- If we set it to true, as mentioned above, the unknown entries are still meaningful to users applications, we will also face the embarrassment that when users really fail the backup on these unknown entries, Kopia util can but Velero cannot

Therefore, the best solution is to follow Kopia -- set it to true by default and expose a user interface.

pkg/uploader/kopia/progress.go
Lyndon-Li: > it would be not suitable to output logs in basic public functions and structures

Didn't quite get this, but as the built-in handler [Error function ](https://github.com/kopia/kopia/blob/v0.10.4/cli/cli_progress.go#L90), it is also printing the log directly in the progress.Error callback.

pkg/uploader/kopia/snapshot.go
qiuming-best: yes, I set IgnoreUnknownTypes to `True` is also refers to the default sets of Kopia.

if we want users to set IgnoreUnknownTypes by themself, what about exposing the interface for users to set all Kopia [polices](https://github.com/kopia/kopia/blob/v0.10.4/snapshot/policy/policy.go#L23) in case of we adding options one by one in future ?

Lyndon-Li (approved):

pkg/uploader/kopia/snapshot.go
```diff
@@ -65,6 +66,10 @@ func newOptionalInt(b policy.OptionalInt) *policy.OptionalInt {
 	return &b
 }
 
+func newOptionalBool(b policy.OptionalBool) *policy.OptionalBool {
```
reasonerjt, line 69: IMO it would be slightly better if the input param is changed to `bool`, like 
```
func newOptionalBool(b bool) *policy.OptionalBool {
	ob := policy.OptionalBool(b)
	return &ob
}
```
otherwise, you are not creating anything `new` in this func

pkg/uploader/kopia/snapshot.go
reasonerjt, line 235: 1. It would be better to check if `policyTree.DefinedPolicy()` is nil, b/c nil is a valid return value of `policyTree.DefinedPolicy()`:
https://github.com/kopia/kopia/blob/master/snapshot/policy/policy_tree.go#L89
1. It would be better to reuse the error fs.ErrUnknown:
https://github.com/kopia/kopia/blob/master/fs/entry.go#L16

pkg/uploader/kopia/snapshot.go
qiuming-best: yeah, Also I modified `newOptionalInt` function

pkg/uploader/kopia/snapshot.go
qiuming-best: modified

reasonerjt (approved):
lgtm

pkg/uploader/kopia/snapshot.go
reasonerjt: thanks!

Lyndon-Li (approved):

PR #5797, opened 2023-01-26
remove the gemfile.lock from src

weshayutin:
The lock file may differ between devels
https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

Signed-off-by: Wesley Hayutin <weshayutin@gmail.com>


weshayutin:
oh dear :(

PR #5842, opened 2023-02-07
Add Design Proposal To Extend Resource Policy With Field Selector Filter

### LINKED ISSUE #5118: Support filter the resource by name during the backup

### LINKED ISSUE #5152: Add new field additionalResources in backup spec

ihcsim:

This design proposal extends the new [ResourcePolicies](https://github.com/vmware-tanzu/velero/blob/main/design/Implemented/handle-backup-of-volumes-by-resources-filters.md?plain=1#L69) resource with field selector properties so that users can define filter criteria based on resource properties like resource name.

Fixes #5152 
Fixes #5118

[0]: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

### DIFF
--- a/design/resource-filter-field-selectors-design.md
+++ b/design/resource-filter-field-selectors-design.md
@@ -1,0 +1,158 @@
+# Extend Resource Policy To Support Field Selectors
+
+## Abstract
+
+This design proposal extends the [`ResourcePolicies` API][0] to support resource filtering using [field selectors][1].
+The proposed field selection design complements the existing [filtering API][2], to enable users to filter backup resources by Kubernetes resource fields.
+
+## Background
+
+The current mechanism to include cluster-scoped resources in backups via the `IncludeResources`, `IncludeClusterResources` and `IncludeClusterScopedResources` API filters the resources by their [`ResourceGroup`][3].
+This is insufficent for operator backup use cases, where specific Custom Resource Definition (CRD) must be included in order for the restored operator to work.
+See issue [#4876][4].
+
+The existing [label selection][5] mechanism doesn't scale on large Kubernetes clusters, due to constraints beyond the users' control like:
+
+* Inconsistent resource labeling scheme across different providers
+* Unique labeling to correctly group CRDs according to operators
+* Strict cluster-scoped RBAC permissions (to label CRDs)
+* Dynamically generated custom resource definitions
+
+## Goals
+
+- Users can define backup filter criteria based on Kubernetes field selectors.
+
+## Non Goals
+
+- N/A
+
+## High-Level Design
+
+Extends the `ResourcePolicies` API with a `FieldSelectorPolicies` configuration.
+
+## Detailed Design
+
+The `FieldSelectorPolicies` works in complement to the `IncludeClusterScopedResources` API.
+For example, to include the Nginx ingress controller CRD, one can define a field selector policy that looks like this:
+
+```sh
+cat <<EOF > backup-policy-nginx-crd.yaml
+version: v1
+fieldSelectorPolicies:
+- groupResource: 
+    group: apiextensions.k8s.io
+    resource: customresourcedefinitions
+  metadata.name: virtualservers.k8s.nginx.org
+  metadata.name: virtualserverroutes.k8s.nginx.org
+  metadata.name: transportservers.k8s.nginx.org
+  metadata.name: policies.k8s.nginx.org
+EOF
+```
+
+This YAML policy is used to create the policy configmap used by Velero to backup the CRDs:
+
+```sh
+kubectl -n velero create cm backup-policy-nginx-crd --from-file backup-policy-nginx-crd.yaml
+
+velero backup create --include-cluster-scoped-resources="customresourcedefinitions" --resource-policies-configmap=backup-policy-nginx-crd
+```
+
+Although the supported field selectors vary across Kubernetes resource types, all resource types support the `metadata.name` and `metadata.namespace` fields.
+Using unsupported field selectors produces an error.
+Only the `=` operator is supported by this proposed API.
+For more information, see the Kubernetes documentation on [field selectors][1].
+
+### API Design
+
+The following code snippet illustrates the proposed API structure to encapsulate the field selector properties:
+
+```go
+type ResourcePolicies struct {
+  Version        string                          `yaml:"version"`
+  VolumePolicies []VolumePolicy                  `yaml:"volumePolicies"`
+
+  // FieldSelectorPolicies defines a list of policies to enable Kubernetes
+  // resource field selection to support filtering by resource fields.
+  FieldSelectorPolicies []FieldSelectorPolicy    `yaml:"fieldSelectorPolicies"`
+
+  // we may support other resource policies in the future, and they could be added separately
+  // OtherResourcePolicies: []OtherResourcePolicy
+}
+
+import (
+  k8s.io/apimachinery/pkg/fields
+  metav1 k8s.io/apimachinery/pkg/apis/meta/v1
+)
+
+// FieldSelectorPolicy describes the field selection properties used to identify a
+// Kubernetes resource.
+type FieldSelectorPolicy struct {
+  // GroupResource specifies a group and a resource.
+  GroupResource metav1.GroupResource `json:"groupResource,omitempty"`
+
+  // FieldSelectors specifies a field name and its value, typically, implemented
+  // as a map of string-to-string.
+  // If omitted, then all resources of the group will be selected.
+  // +optional
+  FieldSelectors fields.Set `json:"fieldSelectors,omitempty"`
+}
+```
+
+### Code Design
+
+The field selector mechanism eliminates the needs to perform client-side filtering which requires all resources of a particular type to be fetched beforehand.
+The selection criteria can be added as a filter option to the `LIST` requests using the [`metav1.ListOptions`][6] option.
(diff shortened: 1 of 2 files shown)


design/resource-filter-field-selectors-design.md
```diff
@@ -0,0 +1,229 @@
...
+
+- Users can define filter criteria like resource name and namespace to include and exclude resources from backup and restore, using Kubernetes field selectors.
+
+## Non Goals
+
+- N/A
+
+## High-Level Design
+
+This design proposes changing the `[]string` type of the following fields in the `BackupSpec` API to a structure that can encapsulate the field selector properties:
```
qiuming-best, line 25: What about putting all the filed selector properties into [ResourcePolicies](https://github.com/vmware-tanzu/velero/blob/main/design/Implemented/handle-backup-of-volumes-by-resources-filters.md?plain=1#L69),
and all these field selector properties form a new kind of resource policies. 
For example:
```golang
type ResourcePolicies struct {
        Version        string         `yaml:"version"`
	VolumePolicies []VolumePolicy `yaml:"volumePolicies"`
         FieldSelectorPolicies []FieldSelectorPolicy `yaml:"fieldSelectorPolicies"`
	// we may support other resource policies in the future, and they could be added separately
  	// OtherResourcePolicies: []OtherResourcePolicy
}
````

As more and more filters are added now, we tend to consolidate all future filters into the ResourcePolicies struct. This will make the configuration more centralized and avoid the backup spec from storing a large number of filters, which can cause its size to increase.

Comment (ihcsim) [reply design/resource-filter-field-selectors-design.md:25]: When using resource policy, is it possible for us to not define the intermediate YAML file, instead just create the config map directly?

Comment (qiuming-best) [reply design/resource-filter-field-selectors-design.md:25]: @ihcsim Yes, the YAML file is a template for users to define policy and then generate a config map from the YAML file.
Also, users could create a config map directly, but the data in the config map should be parsed by Velero.

Comment (reasonerjt) [reply design/resource-filter-field-selectors-design.md:25]: @ihcsim 

Would you refine the design to apply the "resource-policy" approach?

As for whether or not it will be in v1.12, we can make decisions based on the milestone we defined in the wiki.

Comment (ihcsim) [reply design/resource-filter-field-selectors-design.md:25]: I will take a closer look at the resource policy. Also, does this limitation (mentioned in the 1.11 release changelog) has something to do with the creation of the configmap?

> The Flexible resource policy that can filter volumes to skip in the backup is not guaranteed to work on unofficial third-party plugins because the plugins may not follow the existing backup workflow code logic of Velero. The ConfigMap used as the policy is supposed to be maintained by users.

Comment (qiuming-best) [reply design/resource-filter-field-selectors-design.md:25]: @ihcsim It is because some unofficial third-party plugins may not implement Velero standard backup and restore interface in the plugin. Velero will filter resources by resource policy when calling the standard b&r interface. They may have their own CRDs to do b&r (such as the vSphere plugin), so the standard filter resources may not work. 

Our officially supported plugins ([AWS](https://github.com/vmware-tanzu/velero-plugin-for-aws), [Azure](https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure), [GCP](https://github.com/vmware-tanzu/velero-plugin-for-gcp), [vSphere](https://github.com/vmware-tanzu/velero-plugin-for-vsphere)) are functioning well.

**Comment by ihcsim** [comment]
@qiuming-best Apologies for the delay. I have updated the proposal to use the `ResourcePolicies` API. PTAL - thanks.

File: design/resource-filter-field-selectors-design.md [hunk 2]
```diff
@@ -0,0 +1,158 @@
...
+  // FieldSelectors specifies a field name and its value, typically, implemented
+  // as a map of string-to-string.
+  // If omitted, then all resources of the group will be selected.
+  // +optional
+  FieldSelectors fields.Set `json:"fieldSelectors,omitempty"`
+}
+```
+
+### Code Design
+
```
qiuming-best, line 102: @ihcsim 
 The API above looks good, and could you please give more details in `code design`, how and where to filter resources when doing backup in Velero

design/resource-filter-field-selectors-design.md
```diff
@@ -0,0 +1,158 @@
...
+The existing label selection filtering mechanism has also been considered.
+But on clusters with many operators, labeling of CRDs requires user to devise a precise labeling scheme to properly group the CRDs, according to operators, which can be error prone.
+
+## Security Considerations
+
+N/A.
+
+## Compatibility
+
+The field selector mechanism only works with the new `ResourcePolicies` API.
```
qiuming-best, line 138: Please add the conflict resolution about the  FieldSelectorPolicy with current existing rules.

such as if the FieldSelectorPolicy selected one resource and Action is `Skip` but the user used --include-resources when backup

design/resource-filter-field-selectors-design.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Abstract
+
+This design proposal extends the [`ResourcePolicies` API][0] to support resource filtering using [field selectors][1].
+The proposed field selection design complements the existing [filtering API][2], to enable users to filter backup resources by Kubernetes resource fields.
+
+## Background
+
+The current mechanism to include cluster-scoped resources in backups via the `IncludeResources`, `IncludeClusterResources` and `IncludeClusterScopedResources` API filters the resources by their [`ResourceGroup`][3].
+This is insufficent for operator backup use cases, where specific Custom Resource Definition (CRD) must be included in order for the restored operator to work.
```
kaovilai, line 11: This is awesome. Restore of operators have been requested rather frequently.

design/resource-filter-field-selectors-design.md
```diff
@@ -0,0 +1,158 @@
...
+
+This YAML policy is used to create the policy configmap used by Velero to backup the CRDs:
+
+```sh
+kubectl -n velero create cm backup-policy-nginx-crd --from-file backup-policy-nginx-crd.yaml
+
+velero backup create --include-cluster-scoped-resources="customresourcedefinitions" --resource-policies-configmap=backup-policy-nginx-crd
+```
+
+Although the supported field selectors vary across Kubernetes resource types, all resource types support the `metadata.name` and `metadata.namespace` fields.
```
shawn-hurley, line 60: Are we suggesting that we only allow these two in our API? 

If we allow others that a given resource type supports if there is an error, do we error the backup at that point? Is there someway to validate the request during backup resource creation vs during runtime?

ihcsim: The goal is to let K8s tells us which field selectors are supported for a particular resource group, by e.g., specifying the field selectors in the [`metav1.ListOptions`](https://pkg.go.dev/k8s.io/apimachinery@v0.26.1/pkg/apis/meta/v1#ListOptions). Velero shouldn't need to maintain any knowledge of which field selectors are supported server-side.

As for error handling, not sure if it's worth doing more upfront client-side validation, since K8s already has a well-defined API for field selections. I think we can just let Kubernetes handle it, like how `kubectl get` works. If we got an error from K8s, Velero should fail the backup.

design/resource-filter-field-selectors-design.md
ihcsim: Sure thing - I will take a stab at it. I am still unfamiliar with Velero code base, might take slightly longer.

design/resource-filter-field-selectors-design.md
ihcsim: Or some form of dry run option for the overall resource policy may help too? Personally, I am not a fan of admission webhooks because they can be disabled from the webhook configuration, they can go down, they don't load-balance correctly in HA etc.

kaovilai: dry run to validate valid spec sounds good to me.

shawn-hurley: I think that a dry-run command could take the CR and validate the FieldSelectors against a given cluster, would make sense as an option, and would give you a similar feel to the `kubectl get`.

At some point, I do wonder if the project is going to have to consider a webhook, but for this, I think the proposed idea of a dry run makes sense and would alleviate my concerns

design/resource-filter-field-selectors-design.md
ihcsim: Is it necessary for field selector policy to support `action`? At the moment, `action` seems to be specific to volume policy. Also, like labels, the K8s field selector API only works with `=` and `!=`. It doesn't support the equivalence of `Skip`.

reasonerjt:
Is there expectation to merge it in v1.13 timeframe, please be reminded this was not on our radar during v1.13 planning and we've passed v1.13 Feature Freeze milestone.  Unless we decide this is very important, this will be left in backlog.

Also there are other work planned for v1.13 around the resource policy..

ihcsim:
@reasonerjt No, this is not needed for v1.13, unless others feel otherwise.

ihcsim:
Closing this for now, as I am no longer sure if this is the right approach to backup an operator. Ideally, we want a way to identify and backup *all* resources required by an operator. The proposed approach is a bit too granular.

PR #5880, opened 2023-02-20
Proposal to add support for Resource Modifier (AKA JSON Substitutions) in Restore workflow

### LINKED ISSUE #5809: Substitutions of YAML / JSON values through configmap/ RestoreItemAction

Generic way to subsitute YAML/ JSON properties of the workloads before restoring. 

There are a lot of scenarios where customers wish to change certain properties in the YAMLs before restoration, for example removing any cloud specific properties (hardcoded disk encryption key references, VNET references), change the image URLs, change the StorageClass of the PVCs being restored, changing Disk drivers when migrating across clouds, changing other cloud specific properties.

As of today the solutions for each specific problem are being developed as different plugins and using different configmaps, it is very hard from user POV to handle them.

Goal is to solve all these problems in a generic way through RestoreItemAction which can substitute all these various properties based on a configmap reference input.

Currently there are similar issues and solutions which are solving specific problems.
-  considering the local image repositories changes when migrating from cluster A to B #5519
-  Storage Class mapping plugin

Examples: Deployment.spec.containers[0].image  could be the json path.
Or replacing values which match a certain regex. Similarly for StorageClasses, we can change cloud specific SKUs.

Pending: will add more examples.

**Environment:**

**Vote on this issue!**

anshulahuja98:

Design for #5809 

### DIFF
--- a/design/json-substitution-action-design.md
+++ b/design/json-substitution-action-design.md
@@ -1,0 +1,155 @@
+# Proposal to add support for Resource Modifiers (AKA JSON Substitutions) in Restore Workflow
+
+- [Proposal to add support for Resource Modifiers (AKA JSON Substitutions) in Restore Workflow](#proposal-to-add-support-for-resource-modifiers-aka-json-substitutions-in-restore-workflow)
+    - [Abstract](#abstract)
+    - [Goals](#goals)
+    - [Non Goals](#non-goals)
+    - [User Stories](#user-stories)
+        - [Scenario 1](#scenario-1)
+        - [Scenario 2](#scenario-2)
+    - [Detailed Design](#detailed-design)
+        - [Reference in velero API](#reference-in-velero-api)
+        - [ConfigMap Structure](#configmap-structure)
+        - [Operations supported by the JSON Patch library:](#operations-supported-by-the-json-patch-library)
+        - [Advance scenarios](#advance-scenarios)
+            - [Conditional patches using test operation](#conditional-patches-using-test-operation)
+    - [Alternatives Considered](#alternatives-considered)
+    - [Security Considerations](#security-considerations)
+    - [Compatibility](#compatibility)
+    - [Implementation](#implementation)
+    - [Future Enhancements](#future-enhancements)
+    - [Open Issues](#open-issues)
+
+## Abstract
+Currently velero supports substituting certain values in the K8s resources during restoration like changing the namespace, changing the storage class, etc. This proposal is to add generic support for JSON substitutions in the restore workflow. This will allow the user specify filters for particular resources and then specify a JSON patch (operator, path, value) to apply on a resource. This will allow the user to substitute any value in the K8s resource without having to write a new RestoreItemAction plugin for each kind of substitution.
+
+<!-- ## Background -->
+
+## Goals
+- Allow the user to specify a GroupKind, Name(optional), JSON patch for modification.
+- Allow the user to specify multiple JSON patch.
+
+## Non Goals
+- Deprecating the existing RestoreItemAction plugins for standard substitutions(like changing the namespace, changing the storage class, etc.)
+
+## User Stories
+
+### Scenario 1
+- Alice has a PVC which is encrypted using a DES(Disk Encryption Set - Azure example) mentioned in the PVC YAML through the StorageClass YAML. 
+- Alice wishes to restore this snapshot to a different cluster. The new cluster does not have access to the same DES to provision disk's out of the snapshot.
+- She wishes to use a different DES for all the PVCs which use the certain DES.
+- She can use this feature to substitute the DES in all StorageClass YAMLs with the new DES without having to create a fresh storageclass, or understanding the name of the storageclass.
+
+### Scenario 2
+- Bob has multi zone cluster where nodes are spread across zones. 
+- Bob has pinned certain pods to a particular zone using nodeSelector/ nodeaffinity on the pod spec.
+- In case of zone outage of the cloudprovider, Bob wishes to restore the workload to a different namespace in the same cluster, but change the zone pinning of the workload.
+- Bob can use this feature to substitute the nodeSelector/ nodeaffinity in the pod spec with the new zone pinning to quickly failover the workload to a different zone's nodes.
+
+## Detailed Design
+- The user is expected to create a configmap with the desired Resource Modifications. Then the reference of the configmap will be provided in the RestoreSpec.
+- The core restore workflow before creating/updating a particular resource in the cluster will be checked against the filters provided and respective substitutions will be applied on it.
+
+### Reference in velero API 
+> Example of Reference to configmap in RestoreSpec
+```yaml
+apiVersion: velero.io/v1
+kind: Restore
+metadata:
+  name: restore-1
+spec:
+    resourceModifier:
+      refType: Configmap
+      ref: resourcemodifierconfigmap
+```
+> Example CLI Command
+```bash
+velero restore create --from-backup backup-1 --resource-modifier-configmap resourcemodifierconfigmap
+```
+
+### Resource Modifier ConfigMap Structure
+- User first needs to provide details on which resources the JSON Substitutions need to be applied. 
+    - For this the user will provide 4 inputs - Namespaces(for NS Scoped resources), GroupKind (kind.group format similar to includeResources field in velero) and Name Regex(optional).
+    - If the user does not provide the Name, the JSON Substitutions will be applied to all the resources of the given Group and Kind under the given namespaces.
+
+- Further the use will specify the JSON Patch using the structure of kubectl's "JSON Patch" based inputs.
+- Sample data in ConfigMap
+```yaml
+version: v1
+resourceModifierRules:
+- conditions:
+    groupKind: persistentvolumeclaims
+    resourceNameRegex: "mysql.*"
+    namespaces:
+    - bar
+    - foo
+  patches:
+  - operation: replace
+    path: "/spec/storageClassName"
+    value: "premium"
+  - operation: remove
+    path: "/metadata/labels/test"
+```
+- The above configmap will apply the JSON Patch to all the PVCs in the namespaces bar and foo with name starting with mysql. The JSON Patch will replace the storageClassName with "premium" and remove the label "test" from the PVCs.
+- The user can specify multiple JSON Patches for a particular resource. The patches will be applied in the order specified in the configmap.
+- The user can specify multiple resourceModifierRules in the configmap. The rules will be applied in the order specified in the configmap. 
+
+> Users need to create one configmap in Velero install namespace from a YAML file that defined resource modifiers. The creating command would be like the below:
+```bash
+kubectl create cm <configmap-name> --from-file <yaml-file> -n velero
+```
+
+### Operations supported by the JSON Patch library: 
+- add
+- remove
+- replace
+- move
+- copy
+- test (covered below)
+
+### Advance scenarios
+#### **Conditional patches using test operation**
+ The `test` operation can be used to check if a particular value is present in the resource. If the value is present, the patch will be applied. If the value is not present, the patch will not be applied. This can be used to apply a patch only if a particular value is present in the resource. For example, if the user wishes to change the storage class of a PVC only if the PVC is using a particular storage class, the user can use the following configmap.
+```yaml
+version: v1
+resourceModifierRules:
+- conditions:
+    groupKind: persistentvolumeclaims.storage.k8s.io
+    resourceNameRegex: ".*"
+    namespaces:
+    - bar
+    - foo
+  patches:
+  - operation: test
+    path: "/spec/storageClassName"
+    value: "premium"
+  - operation: replace
+    path: "/spec/storageClassName"
+    value: "standard"
+```
+
+## Alternatives Considered
+1. JSON Path based addressal of json fields in the resource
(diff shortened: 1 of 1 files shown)


reasonerjt:
TBH I still prefer the RIA approach for its flexibility.

This can be a good utility, but not quite sure if really needed to be part of the upstream

sseago:
@reasonerjt "TBH I still prefer the RIA approach for its flexibility."

The proposal is to write a RIA plugin to do this. A separate question is whether this is a RIA that should be part of core velero (alongside existing core RIAs such as storageclass change, etc.), or whether this makes more sense as a separate plugin, installed as a separate image. One advantage of the plugin approach is that if implemented as a RIA, it could first be written as a completely external RIA plugin, at least as a proof of concept, and then if there is general interest in including this in velero core, it could be moved. The challenge for making something non-core that's generally useful is that there isn't really an existing upstream repo that is appropriate for "supported plugins that aren't in the core repo", and it's probably too small to warrant its own independent repo (as we have for the storage plugins, etc.), and we don't really have a mechanism for sharing/supporting plugin content outside of creating new repos or putting it in the core repo.

design/json-substitution-action-design.md
```diff
@@ -0,0 +1,122 @@
...
+  name: restore-1
+spec:
+    jsonSubstitution:
+      refType: Configmap
+      ref: backup01
+```
+
+### ConfigMap Structure
+- User first needs to provide details on which resources the JSON Substitutions need to be applied. 
+    - For this the user will provide 4 inputs - Nameespaces(for NS Scoped resources), Group, Kind and Name Regex(optional).
```
reasonerjt, line 62: nit: nameespaces -> namespace

design/json-substitution-action-design.md
```diff
@@ -0,0 +1,122 @@
...
+    - If the user does not provide the Name, the JSON Substitutions will be applied to all the resources of the given Group and Kind under the given namespaces.
+
+- Further the use will specify the JSON Substitutions using the structure of kubectl's "JSON Path" based inputs.
+    - Reference: [https://kubernetes.io/docs/reference/kubectl/jsonpath/](https://kubernetes.io/docs/reference/kubectl/jsonpath/)
+    - ![image](https://user-images.githubusercontent.com/36476228/219611516-7fa9527d-c592-4471-8e3e-39ad5e7606db.png)
+    - This filter can be used through golang Package: "k8s.io/client-go/util/jsonpath"
+
+- Last the user will provide the current value regex, basically at the path specified above, we check if the provided regex matches the value at that path. If it does, we substitute the value with the new value provided by the user.
```
reasonerjt, line 70: Do you have a concrete use case for "current value regex"?  In your sample they are all ".*", if it's not necessary we may consider add it in future?
In addition, I'm relatively newbie to jsonpath but it seems to support `filter` feature, so maybe such `filter` can be used to matching current values when it's used with `@`

design/json-substitution-action-design.md
```diff
@@ -0,0 +1,122 @@
...
+kind: ConfigMap
+metadata:
+    name: json-substitution-configmap
+    namespace: velero
+data:
+    json-substitution-configmap.yaml: |
+        resources:
+        - namespaces:
+            - default
+        group: storage.k8s.io
```
reasonerjt, line 84: I belive the "group" should be aligned with "namespaces" right?
I Could you fix the indent?

design/json-substitution-action-design.md
```diff
@@ -0,0 +1,122 @@
...
+        kind: StorageClass
+        nameRegex: "disk*"
```
reasonerjt, line 86: nit: do you mean "disk.*"?

design/json-substitution-action-design.md
```diff
@@ -0,0 +1,122 @@
...
+
+- Sample ConfigMap
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+    name: json-substitution-configmap
+    namespace: velero
+data:
+    json-substitution-configmap.yaml: |
```
reasonerjt, line 80: Do we wanna support versioning? just in case we want to introduce break changes to the yaml?

Just a thought, if we believe this is good enough we don't need versioning, b/c migration is always painful.

design/json-substitution-action-design.md
```diff
@@ -0,0 +1,122 @@
...
+        group: storage.k8s.io
+        kind: StorageClass
+        nameRegex: "disk*"
+        jsonSubstitutions:
+        - path: {..parameters.diskEncryptionSetID}
+            currentValueRegex: ".*"
+            newValue: "new-des"
+        - namespaces:
+            - default
+        group: apps
```
reasonerjt, line 93: In a corner case, if two rules happen to modify the same field of same object, the latter will overwrite the earlier right?

Could you also clarify that in the design?

design/json-substitution-action-design.md
```diff
@@ -0,0 +1,122 @@
...
+    - Not an extensible design. If a new kind of substitution is required, a new RestoreItemAction needs to be written.
+4. RIA for JSON Subtitution: The approach of doing JSON Substitution through a RestoreItemAction plugin was considered. But it is likely to have performance implications as the plugin will be invoked for all the resources.
+
+## Security Considerations
+No security impact.
+
+## Compatibility
+Compatibility with existing StorageClass mapping RestoreItemAction and similar plugins needs to be evaluated. 
+
+## Implementation
```
reasonerjt, line 116: Will we also want to add a param to `velero restore` command?

design/json-substitution-action-design.md
```diff
@@ -0,0 +1,122 @@
...
+spec:
+    jsonSubstitution:
+      refType: Configmap
+      ref: backup01
+```
+
+### ConfigMap Structure
+- User first needs to provide details on which resources the JSON Substitutions need to be applied. 
+    - For this the user will provide 4 inputs - Nameespaces(for NS Scoped resources), Group, Kind and Name Regex(optional).
+    - If the user does not provide the Name, the JSON Substitutions will be applied to all the resources of the given Group and Kind under the given namespaces.
```
reasonerjt, line 63: Just to clarify there are other selectors in the spec of restore CR, so a JSON substitution will only apply to the selected objects right, and just in case we wanna introduce other finer-grained selectors, for example, the user only wants to restore an object if a certain field matches a certain value.  This JSON substitution rule should apply after all the selection "filter" right?  
This should be clarified, b/c in future the substitution rule may impact the behavior of the selectors.

reasonerjt:
@anshulahuja98 thanks for the update, I've added a few comments in details but generally I'm OK we introduce such capability in v1.12

@sseago @shubham-pampattiwar let us know your thoughts.

design/json-substitution-action-design.md
ywk253100, line 70: Does this support replace part of the value? e.g. I want to replace the registry part of all the images: reg1/xxx/xxx -> reg2/xxx/xxx

design/json-substitution-action-design.md
ywk253100: I think it is. The substitution only happens just before creating the objects.

design/json-substitution-action-design.md
ywk253100, line 70: The JSON path only locates the target value, but how can we substitute it? Are there any functions provided by `k8s.io/client-go/util/jsonpath` for this?

ywk253100: Does this support substitute complicated structures rather than simple values? e.g. 
```
a:
  b:
    c
```
-> 
```
a:
  d
```

design/json-substitution-action-design.md
anshulahuja98: make sense - will account for that.

changelogs/unreleased/5880-anshulahuja98
```diff
@@ -0,0 +1 @@
+Add Proposal to add support for JSON Substitutions through RestoreItemAction plugin.
```
kaovilai, line 1: Designs do not affect release content so I do not think changelogs applies here.

design/json-substitution-action-design.md
anshulahuja98: just to confirm yes, this is intended only just before creation of objects.

design/json-substitution-action-design.md
anshulahuja98: there is not edit function/ substitute function inbuilt in the jsonpath library. 
The intention is to make a fork of sorts in the velero code base and just use the relevant functions from the json path library as utils

I don't expect any significant changes to come o the library over time, so I am not worried about keeping the fork in sync with upstream.

anshulahuja98: Searching for complex structures using jsonpath is possible. But providing a substitute like this will complicate the config map design.
The user scenarios I have mentioned can mostly be covered with just substituting values of leaf nodes(string/ int)
So it is not my intention to support such complex substitutions atm. We will mostly skip such scenarios in case attempted by user.

design/json-substitution-action-design.md
anshulahuja98: Ack.

design/json-substitution-action-design.md
anshulahuja98: Yes, ack.

design/json-substitution-action-design.md
anshulahuja98: Ack will do.

design/json-substitution-action-design.md
anshulahuja98: Let me check how it is added for resourceFilters configMapRef the CLI is updated.
it will be similar to that.
I will callout in design as well.

sseago:
@anshulahuja98 as we were discussing in the community meeting today, I looked into the question of an internal RIA -- but it looks like that won't avoid the GRPC call, so I think we probably need to stick with the current approach of doing it directly in the restore workflow.

anshulahuja98:
> @anshulahuja98 as we were discussing in the community meeting today, I looked into the question of an internal RIA -- but it looks like that won't avoid the GRPC call, so I think we probably need to stick with the current approach of doing it directly in the restore workflow.

Thank you for confirming this!

Lyndon-Li:
@anshulahuja98 
There is a case in Data Mover scenarios -- backup from one provider(e.g., EKS) and restore to a different provider(e.g., AKS).
Then if the source PVC is with StorageClass explicitly specified, even though the SC is backed up by Velero, users cannot directly restore the SC into the target since the SC's content are almost 100% different.

Do you see any problem for the current PR and its implementation to cover this case?
Moreover, since the two SCs are very different, do you think it is better to give users an option to overwrite the entire content of the yaml?

anshulahuja98:
@Lyndon-Li 
So firstly to call out there is a design change which I called out in the previous community meeting where we are switching the approach to JsonPatch based impl as compared to JsonPath. I haven't yet pushed the design changes to this PR.

With Json Patch, it will be even possible to change the entire storage class. Since you can kind of provide a JSON Patch at a Spec level and it will end up modifying the entire storage class.

Quoting a Kubectl patch example which is possible
"https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/"
kubectl patch deployment patch-demo --patch 'spec:\n template:\n  spec:\n   containers:\n   - name: patch-demo-ctr-2\n     image: redis'


So using a similar patch properly escaped should work. I don't see any issue.

design/json-substitution-action-design.md
```diff
@@ -0,0 +1,160 @@
...
+version: v1
+resourceModifierRules:
+- conditions:
+    groupKind: persistentvolumeclaims.storage.k8s.io
+    resourceNameRegex: ".*"
+    namespaces:
+    - bar
+    - foo
+  patches:
+  - operation: test
```
ywk253100, line 128: Should we put the `test` patch in the `conditions` section as it is the condition actually?

ywk253100: Shall we allow multiple `test` patches? What is the relationship of them? `and ` or `or`?

anshulahuja98: I would prefer to keep it in the `patches` section to make the overall configmap easier to understand to end user and keep it closer to the JsonPatch RFC.

The test patches are `and` in some sense, basically even if 1 test patch fails, the overall patch will fail.
We can allow multiple test patches since it is a supported scenario.

design/json-substitution-action-design.md
anshulahuja98: updating thread after the proposal update from jsonpath to jsonpatch.

Updating partial values is still not a scenario which is inteneded to supported.

However in the future scope - I have called out regex support on the value. Where you can use the "test" operator to see if the image is of the pattern `reg.*` and only update in those cases.

design/json-substitution-action-design.md
anshulahuja98: resolving this.
no change in updated design.

design/json-substitution-action-design.md
anshulahuja98: velero restore create --from-backup backup-1 --resource-modifier-configmap resourcemodifierconfigmap

added the CLI command in the design doc.

design/json-substitution-action-design.md
anshulahuja98: accounted for it in design

design/json-substitution-action-design.md
anshulahuja98: This has been moved to future scope and not intended for current proposal since it will require forking the jsonpatch library.

sseago (approved):

Issue #5881, opened 2023-02-20
Snapshots were taken although PVs were backed up by node agent on AWS & Azure


danfengliu:
**What steps did you take and what happened:**
When performaing PV backup, I specified `default-volumes-to-fs-backup`  in installation  CLI, in this case snapshot won't be taken as Velero design, but in fact snapshot  were taken.

1. Velero Install Command:
I specified `default-volumes-to-fs-backup` in installation instead of backup command.

`velero install --namespace velero --image gcr.io/velero-gcp/nightly/velero:velero-beed887e --use-node-agent --default-volumes-to-fs-backup --provider azure --backup-location-config xxxxxx --bucket xxxxxx --secret-file xxxxxx --plugins velero/velero-plugin-for-microsoft-azure:main --dry-run --output json" `

3. Velero backup Command:

```
velero --namespace velero create backup backup-2cb849d5-c052-49fd-9b78-954771ae6445 --wait --include-namespaces kibishii-workload2cb849d5-c052-49fd-9b78-954771ae6445
Backup request "backup-2cb849d5-c052-49fd-9b78-954771ae6445" submitted successfully.
Waiting for backup to complete. You may safely press ctrl-c to stop waiting - your backup will continue in the background.
.....................
Backup completed with status: Completed. You may check for more information using the commands `velero backup describe backup-2cb849d5-c052-49fd-9b78-954771ae6445` and `velero backup logs backup-2cb849d5-c052-49fd-9b78-954771ae6445`.

```

4. Verify PVB and snapshots:
As you can see in logs below PVBs were created and snapshots were taken also, which is not expected.
```
Waiting for backup to complete. You may safely press ctrl-c to stop waiting - your backup will continue in the background.
.....................
Backup completed with status: Completed. You may check for more information using the commands `velero backup describe backup-2cb849d5-c052-49fd-9b78-954771ae6445` and `velero backup logs backup-2cb849d5-c052-49fd-9b78-954771ae6445`.
get backup cmd =/velero/workspace/velero-e2e-test/velero/test/e2e/../../_output/bin/linux/amd64/velero --namespace velero backup get -o json backup-2cb849d5-c052-49fd-9b78-954771ae6445
/usr/local/bin/kubectl get podvolumebackup -n velero
/bin/grep kibishii-workload2cb849d5-c052-49fd-9b78-954771ae6445
/usr/bin/awk {print $1}
[...]

PR #5894, opened 2023-02-21
Fix issue 5881

Lyndon-Li:
This is to fix issue 5881, enhance the PVB tracker in two modes, Track and Taken

Fix issue #5881

### DIFF
--- a/pkg/backup/pvc_snapshot_tracker.go
+++ b/pkg/backup/pvc_snapshot_tracker.go
@@ -20,45 +20,60 @@ import (
 	"fmt"
 
 	corev1api "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // pvcSnapshotTracker keeps track of persistent volume claims that have been snapshotted
 // with pod volume backup.
 type pvcSnapshotTracker struct {
-	pvcs sets.String
+	pvcs map[string]pvcSnapshotStatus
+}
+
+type pvcSnapshotStatus struct {
+	taken bool
 }
 
 func newPVCSnapshotTracker() *pvcSnapshotTracker {
 	return &pvcSnapshotTracker{
-		pvcs: sets.NewString(),
+		pvcs: make(map[string]pvcSnapshotStatus),
 	}
 }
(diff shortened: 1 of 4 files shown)


pkg/backup/pvc_snapshot_tracker.go
```diff
@@ -20,45 +20,60 @@ import (
...
-// tracks each snapshotted volume that's a PVC.
-func (t *pvcSnapshotTracker) Track(pod *corev1api.Pod, snapshottedVolumes []string) {
-	for _, volumeName := range snapshottedVolumes {
-		// if the volume is a PVC, track it
-		for _, volume := range pod.Spec.Volumes {
-			if volume.Name == volumeName {
-				if volume.PersistentVolumeClaim != nil {
-					t.pvcs.Insert(key(pod.Namespace, volume.PersistentVolumeClaim.ClaimName))
+// Track indicates a volume from a pod should be snapshotted by pod volume backup.
+func (t *pvcSnapshotTracker) Track(pod *corev1api.Pod, volumeName string) {
```
reasonerjt, line 42: Track .vs. Take is a little confusing, how about pass the status or mode as a parameter and make the renaming more exlicit?

so the caller will call `tracker.Track(pod, volume, VOLUME_SNAPSHOT)` or `tracker.Track(pod, volume, PODVOLUME_BACKUP)`

Lyndon-Li: I have been thinking of how to clearly distinguish them, actually:

- Track means the volume is decided to be handled by PVB according to rules and annotations. We need to record this situation because we don't want the BIA and volume snapshotter to process the volume again
- Take means the volume has been processed by PVB successfully. We need to record this situation, because we want the volumes that is to processed by PVB but missed by the backup with the 1st pod should be processed by other pods. (In the case of RWX mount)

Therefore, if we want to use status, the status should be something like:

- VestedToPVB
- ProcessedByPVB

What do you think? @reasonerjt

reasonerjt (approved):

blackpiglet (approved):

PR #5901, opened 2023-02-22
Handle backup of volume by resource policies

qiuming-best:

This one is an implement of resource policies, which related to the design of https://github.com/vmware-tanzu/velero/pull/5773


internal/resourcepolicies/resource_policies.go
```diff
@@ -0,0 +1,109 @@
...
+	return vpm.matcherFunc(volume, actionType)
+}
+
+// Match interface is implemented by VolumePolicy
+func (policy *VolumePolicy) Match(volume interface{}, actionType string) bool {
+	if policy.Action.Type != actionType {
+		return false
+	}
+
+	val, ok := volume.(*VolumeConditions)
```
reasonerjt, line 47: is the first param a condition or a resource?

internal/resourcepolicies/volume_conditions.go
```diff
@@ -0,0 +1,131 @@
...
+		sc := strings.TrimSpace(v)
+		if target == sc {
+			return true
+		}
+	}
+	return false
+}
+
+// IsConditionsMatch check current VolumeConditions matches current selector
+func (v *VolumeConditions) IsConditionsMatch(selector *VolumeConditions) bool {
```
reasonerjt, line 70: How about we introduce polymorphism at the condition level
```
type Condition interface {
// Check if the resource meet the condition
    Check(res) bool
// Validate if the condition which is parsed from the input yaml is valid
    Validate() 
}

```

 so there's `StorageClassCondition` `CapCondition`...  And these instances are created when you parse a single policy.
 
 There action may also that pattern in future b/c we want it to have parameters, and those need to be validated too, but that piece can be added later when needed.

internal/resourcepolicies/resource_policies.go
```diff
@@ -0,0 +1,109 @@
...
+	case "volume":
+		return rpf.volumeMatchers
+	default:
+		return nil
+	}
+}
+
+// IsVolumeMatch checks the current volume is match resource policies
+// It will return once matched ignoring the latter policies
+func IsVolumeMatch(policies *ResourcePolicies, volume *VolumeConditions, actionType string) bool {
```
reasonerjt, line 99: If we do it this way, we end up doing repeated checks b/c when there are more actions needed you may want to call
`IsVolumeMatch(p, v, "skip")`
`IsVolumeMatch(p,v,"volume-snapshot")` ....

How about we do it like 
`MatchVolume(p, v) *Action`
so we only go through the matchers once and find the action for the volume?

reasonerjt:
And a general comment, let's try make this package more cohesive, i.e. only export the func or type when really needed.

internal/resourcepolicies/resource_policies.go
qiuming-best: it's the resource, and it's also `VolumeConditions` type of volume.

Here all original resource volumes would be PVs or Pod Volumes, they have different structures.  In order to match volumes of all kinds, I introduced one VolumeConditions(Maybe I should change the name) type of volume. Before doing the match volumes action, volumes of all kinds should transfer all filed values to a new VolumeConditions volume, then the match volumes action will avoid to compatible with various kinds of volume.

internal/resourcepolicies/resource_policies.go
qiuming-best: sounds good

internal/resourcepolicies/volume_conditions.go
qiuming-best: ok

internal/resourcepolicies/volume_policies_validator.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	// Validate Others fields
+	return true, nil
+}
+
+// Validate check action format
+func (a *Action) Validate() (bool, error) {
+	// Validate Type
+	if !(a.Type == "volume-snapshot" || a.Type == "file-system-backup" || a.Type == "skip") {
```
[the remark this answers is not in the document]
qiuming-best: OK, I'll first only support `skip` acton.

internal/resourcepolicies/resource_policies.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// Action defined as one action for a specific way of backup
+type Action struct {
+	// Type defined specific type of action, currently only support 'skip'
+	Type string `yaml:"type"`
+	// Parameters defined map of parameters when executing a specific action
+	// +optional
+	// +nullable
+	Parameters map[string]interface{} `yaml:"parameters,omitempty"`
```
ywk253100, line 15: Are the `+optional` and `+nullable` marker useful here?

internal/resourcepolicies/volume_policies.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+type storageClassCondition struct {
+	storageClass []string
+}
+
+func (s *storageClassCondition) Match(v *StructuredVolume) bool {
+	if len(s.storageClass) == 0 {
+		return true
+	} else if v.storageClass == "" {
```
ywk253100, line 55: No `else` is needed here, just `if v.storageClass == ""` is OK as you have returned if the last `if` match, please check all other similar codes.

internal/resourcepolicies/volume_policies.go
```diff
@@ -0,0 +1,180 @@
...
+	"bytes"
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+type volumePolicier struct {
```
ywk253100, line 13: Is it the same with `VolumePolicy`?

internal/resourcepolicies/resource_policies.go
```diff
@@ -0,0 +1,116 @@
...
+	if err := decodeStruct(strings.NewReader(*YamlData), resPolicies); err != nil {
+		return nil, fmt.Errorf("failed to decode yaml data into resource policies  %v", err)
+	} else {
+		return resPolicies, nil
+	}
+}
+
+// GetVolumeMatchedAction checks the current volume is match resource policies
+// It will return once matched ignoring the latter policies
+func GetVolumeMatchedAction(policies *ResourcePolicies, volume *StructuredVolume, actionType string) string {
```
ywk253100, line 106: Have you thought of how this function is integrated with the current backup flow?

internal/resourcepolicies/volume_policies_validator.go
```diff
@@ -0,0 +1,112 @@
...
+				if !valid {
+					return false, nil
+				}
+			}
+		}
+	}
+	return true, nil
+}
+
+func (c *capacityCondition) Validate() (bool, error) {
```
ywk253100, line 41: Usually, we define the methods in the same file of its object.

internal/resourcepolicies/resource_policies.go
```diff
@@ -0,0 +1,116 @@
...
+	policy      *VolumePolicy
+	matcherFunc func(volume interface{}, actionType string) bool
+}
+
+func (v *volumePolicyMatcherImpl) Match(volume interface{}, actionType string) bool {
+	return v.matcherFunc(volume, actionType)
+}
+
+// Match interface is implemented by VolumePolicy
+func (policy *VolumePolicy) Match(volume interface{}, actionType string) bool {
```
ywk253100, line 47: Could you reorganize the code and put the related definition in the same go file? Currently, the related definition of `VolumePolicy` in both `resource_policies.go` and `volume_policies.go`

internal/resourcepolicies/resource_policies.go
qiuming-best: Removed

internal/resourcepolicies/resource_policies.go
qiuming-best: `actionType` has been removed

internal/resourcepolicies/volume_policies.go
qiuming-best: modified

internal/resourcepolicies/resource_policies_mgr.go
```diff
@@ -0,0 +1,139 @@
...
+	err = mgr.kbClient.Get(context.Background(), kbClient.ObjectKey{Name: pvName}, pv)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	volume := StructuredVolume{}
+	volume.ParsePV(pv)
+	return &volume, nil
+}
+
+func (mgr *ResPoliciesManager) GetVolumeMatchedAction(refName, refType string, volume *StructuredVolume) (*Action, error) {
```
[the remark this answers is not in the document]
qiuming-best: yes, this GetVolumeMatchedAction is called in backup process.
the handle logic entry similar like below

```golang
func (ib *itemBackupper) executeActions(...)
.....
	if groupResource.String() == "persistentvolumeclaims" && ib.backupRequest.Spec.ResourcePolices != nil {
	    volume, err := respolicies.ResPoliciesMgr.GetStructredVolumeForPVC(namespace, name)
	   action, err := respolicies.ResPoliciesMgr.GetVolumeMatchedAction(ib.backupRequest.Spec.ResourcePolices.RefName, ib.backupRequest.Spec.ResourcePolices.RefType, volume)
         if action != nil && action.Type == respolicies.Skip {
....
}
```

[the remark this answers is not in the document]
ywk253100: The `configmap` should be read once for one backup and the `ResourcePolicies` should be built once as well for one backup and immutable during the whole backup processing.

internal/resourcepolicies/resource_policies_mgr.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kbClient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// resPoliciesMeta stored one resource policies and related resource policies configmap creation timestamp
+type resPoliciesMeta struct {
+	policies                *ResourcePolicies
+	configCreationTimestamp metav1.Time
```
ywk253100, line 16: Use UID instead is more reasonable

internal/resourcepolicies/resource_policies_mgr.go
```diff
@@ -0,0 +1,139 @@
...
+
+// resPoliciesKey identify one specific resPoliciesMeta with refName and refType
+type resPoliciesKey struct {
+	refName string
+	refType string
+}
+
+// ResPoliciesManager manage all resource policies in memory and keep all up-to-date
+type ResPoliciesManager struct {
+	resPolicies map[resPoliciesKey]*resPoliciesMeta
```
ywk253100, line 27: Use UID as the key

internal/resourcepolicies/resource_policies_mgr.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+// resPoliciesMeta stored one resource policies and related resource policies configmap creation timestamp
+type resPoliciesMeta struct {
+	policies                *ResourcePolicies
+	configCreationTimestamp metav1.Time
+}
+
+// resPoliciesKey identify one specific resPoliciesMeta with refName and refType
+type resPoliciesKey struct {
```
ywk253100, line 20: Use UID instead as the key

internal/resourcepolicies/resource_policies_mgr.go
```diff
@@ -0,0 +1,139 @@
...
+var ResPoliciesMgr ResPoliciesManager
+
+func (mgr *ResPoliciesManager) InitResPoliciesMgr(namespace string, kbClient kbClient.Client) {
+	mgr.resPolicies = make(map[resPoliciesKey]*resPoliciesMeta)
+	mgr.namespace = namespace
+	mgr.kbClient = kbClient
+}
+
+// getResourcePolicies first check and update current resource policies and then get the latest resource policies
+func (mgr *ResPoliciesManager) getResourcePolicies(refName, refType string) (*ResourcePolicies, error) {
```
ywk253100, line 41: How about refactoring the flow of this method as follows?
1. Get the `configmap` from the API server and check whether it's corresponding policy already exists. If it does, return directly
2. If not, build the policy, put it into the map and return

internal/resourcepolicies/resource_policies_mgr.go
```diff
@@ -0,0 +1,139 @@
...
+	policiesConfigmap := &v1.ConfigMap{}
+	err := mgr.kbClient.Get(context.Background(), kbClient.ObjectKey{Namespace: mgr.namespace, Name: refName}, policiesConfigmap)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get resource policies %s/%s configmap", refName, mgr.namespace)
+	}
+	if policiesConfigmap == nil {
+		return errors.Wrapf(err, "get empty resource policies %s/%s configmap", refName, mgr.namespace)
+	}
+
+	updateResPolicies := func() error {
```
ywk253100, line 75: The name is `updateResPolicies`, but it doesn't update the policy, it creates a policy

internal/resourcepolicies/resource_policies_mgr.go
```diff
@@ -0,0 +1,139 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (mgr *ResPoliciesManager) GetStructredVolumeForPVC(namespace, name string) (*StructuredVolume, error) {
+	pvc := &v1.PersistentVolumeClaim{}
+	err := mgr.kbClient.Get(context.Background(), kbClient.ObjectKey{Namespace: namespace, Name: name}, pvc)
```
ywk253100, line 108: This method should receive the PVC object(or unstructured.Unstructured) as a parameter rather than getting it from the API server again.

internal/resourcepolicies/resource_policies_mgr.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	if pvc.Spec.VolumeName == "" {
+		return nil, errors.Errorf("PVC %s/%s has no volume backing this claim", namespace, name)
+	}
+	if pvc.Status.Phase != v1.ClaimBound {
```
ywk253100, line 115: Should we return the error or just skip it?

internal/resourcepolicies/resource_policies_mgr.go
```diff
@@ -0,0 +1,139 @@
...
+	} else {
+		if err := updateResPolicies(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (mgr *ResPoliciesManager) GetStructredVolumeForPVC(namespace, name string) (*StructuredVolume, error) {
```
ywk253100, line 106: This method only covers PVC, what about PV? The snapshot is targetting PV, right?

internal/resourcepolicies/resource_policies_mgr.go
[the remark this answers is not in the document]
qiuming-best: yes, currently, the `ResourcePolicies` only build once if it's up-to-date.

the `ResourcePolicies` is stored in `resPoliciesMeta` and can only be updated when doing backup processing. and backup processing is related to one specific backup CR in reconciliation, So it could not be updated in parallel.

internal/resourcepolicies/resource_policies_mgr.go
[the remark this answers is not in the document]
qiuming-best: sounds good

internal/resourcepolicies/resource_policies_mgr.go
qiuming-best: Here I'm using `refName, refType` as the key as `refName` + `refType` could refer to the same name and type of config.
For example, if the user first refers to the policies as `cm0` + `configmap` but later updates the rules in configmap`cm0`.  then the `resPoliciesMeta` in map of`resPolicies map[resPoliciesKey]*resPoliciesMeta` with key `cm0` + `configmap` will be update, so `refName` + `refType` will keep the related resPoliciesMeta up-to-date.

Also the entry of searching policies is through `refName` and `refType` in backup CR in design, and if using UID it's not easy comparing with `refName` and `refType`

internal/resourcepolicies/resource_policies_mgr.go
qiuming-best: ok, I'll modify it

internal/resourcepolicies/resource_policies_mgr.go
qiuming-best: reasonable

internal/resourcepolicies/resource_policies_mgr.go
qiuming-best: both ok. we could skip checking it here because it will check later in CSI plugin?

internal/resourcepolicies/resource_policies_mgr.go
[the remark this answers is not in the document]
qiuming-best: But what if the same name configmap is been recreated, and the `resourceVersion` also has problem?

internal/resourcepolicies/resource_policies_mgr.go
qiuming-best: updated

internal/resourcepolicies/resource_policies_mgr.go
qiuming-best: updated

internal/resourcepolicies/resource_policies_mgr.go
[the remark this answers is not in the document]
qiuming-best: updated

pkg/cmd/cli/backup/create.go
```diff
@@ -186,6 +191,27 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 		}
 	}
 
+	if o.ResPoliciesConfigmap != "" {
+		cm := &corev1.ConfigMap{}
+		if err := client.Get(context.Background(), kbclient.ObjectKey{
+			Namespace: f.Namespace(),
+			Name:      o.ResPoliciesConfigmap,
+		}, cm); err != nil {
+			return errors.Wrapf(err, "failed to get resource policies %s/%s", o.ResPoliciesConfigmap, f.Namespace())
```
ywk253100, line 200: Incorrect order for policy and namespace

qiuming-best: updated

pkg/apis/velero/v1/backup_types.go
```diff
@@ -129,6 +129,17 @@ type BackupSpec struct {
 	// The default value is 1 hour.
 	// +optional
 	ItemOperationTimeout metav1.Duration `json:"itemOperationTimeout,omitempty"`
+	// ResourcePolices specifies the referenced resource policies that backup should follow
+	// +optional
+	ResourcePolices *ResourcePolices `json:"resourcePolices,omitempty"`
```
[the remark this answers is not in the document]
qiuming-best: updated

internal/resourcepolicies/resource_policies.go
```diff
@@ -0,0 +1,138 @@
...
+	if err := decodeStruct(strings.NewReader(*YamlData), resPolicies); err != nil {
+		return nil, fmt.Errorf("failed to decode yaml data into resource policies  %v", err)
+	} else {
+		return resPolicies, nil
+	}
+}
+
+// GetVolumeMatchedAction checks the current volume is match resource policies
+// It will return once matched ignoring the latter policies
+func GetVolumeMatchedAction(policies *ResourcePolicies, volume *StructuredVolume) *Action {
```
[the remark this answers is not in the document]
qiuming-best: updated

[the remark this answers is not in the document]
qiuming-best: updated

pkg/backup/item_backupper.go
```diff
@@ -315,7 +320,30 @@ func (ib *itemBackupper) executeActions(
 			continue
 		}
 		log.Info("Executing custom action")
+		if ib.backupRequest.Spec.ResourcePolices != nil && groupResource.String() == "persistentvolumeclaims" && action.Name() == "velero.io/csi-pvc-backupper" {
```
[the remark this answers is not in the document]
qiuming-best: updated

pkg/backup/item_backupper.go
```diff
@@ -465,6 +493,16 @@ func (ib *itemBackupper) takePVSnapshot(obj runtime.Unstructured, log logrus.Fie
 		return nil
 	}
 
+	if ib.backupRequest.Spec.ResourcePolices != nil {
```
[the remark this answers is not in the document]
qiuming-best: updated

pkg/cmd/cli/backup/create.go
```diff
@@ -186,6 +191,27 @@ func (o *CreateOptions) Validate(c *cobra.Command, args []string, f client.Facto
 		}
 	}
 
+	if o.ResPoliciesConfigmap != "" {
```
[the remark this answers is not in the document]
qiuming-best: updated

internal/resourcepolicies/resource_policies.go
```diff
@@ -0,0 +1,143 @@
...
+	// Conditions defined list of conditions to match Volumes
+	Conditions map[string]interface{} `yaml:"conditions"`
+	Action     Action                 `yaml:"action"`
+}
+
+// currently only support configmap type of resource config
+const ConfigmapRefType string = "configmap"
+
+// ResourcePolicies currently defined slice of volume policies to handle backup
+type ResourcePolicies struct {
```
ywk253100, line 34: The `ResourcePolicies` is only used internally, so let's change it to an unexported type

pkg/controller/schedule_controller.go
```diff
@@ -258,10 +258,11 @@ func getNextRunTime(schedule *velerov1.Schedule, cronSchedule cron.Schedule, asO
...
 	name := item.TimestampedName(timestamp)
-	backup := builder.
+	backupBuilder := builder.
 		ForBackup(item.Namespace, name).
-		FromSchedule(item).
-		Result()
-
-	return backup
+		FromSchedule(item)
+	if item.Spec.Template.ResourcePolicies != nil {
```
ywk253100, line 264: Why the changes here are needed? The `builder.FromSchedule()` should already handle the policy?

pkg/backup/item_backupper.go
```diff
@@ -214,6 +219,8 @@ func (ib *itemBackupper) backupItemInternal(logger logrus.FieldLogger, obj runti
 			backupErrs = append(backupErrs, err)
 		}
 		return false, itemFiles, kubeerrs.NewAggregate(backupErrs)
+	} else if updatedObj == nil { // which would be skip handle obj by resource policies
```
ywk253100, line 222: Why this is needed?

pkg/backup/item_backupper.go
```diff
@@ -310,8 +317,32 @@ func (ib *itemBackupper) executeActions(
...
+			if err := ib.kbClient.Get(context.Background(), kbClient.ObjectKey{Name: pvName}, pv); err != nil {
+				return nil, itemFiles, errors.WithStack(err)
+			}
 
+			volume := resourcepolicies.StructuredVolume{}
+			volume.ParsePV(pv)
+			act := ib.backupRequest.ResPolicies.Match(&volume)
+			if act != nil && act.Type == resourcepolicies.Skip {
+				log.Infof("skip snapshot of pv %s for the matched resource policies", pvName)
+				return nil, itemFiles, nil
```
ywk253100, line 341: We should continue to run other BIAs rather than return

qiuming-best: yeah, Updated

pkg/backup/item_backupper.go
qiuming-best: remove it

pkg/controller/schedule_controller.go
qiuming-best: updated

internal/resourcepolicies/resource_policies.go
qiuming-best: updated

pkg/backup/item_backupper.go
```diff
@@ -317,8 +323,32 @@ func (ib *itemBackupper) executeActions(
 			continue
 		}
 		log.Info("Executing custom action")
+		if ib.backupRequest.ResPolicies != nil && groupResource.String() == "persistentvolumeclaims" && action.Name() == "velero.io/csi-pvc-backupper" {
```
Lyndon-Li, line 326: Two suggestions here:
1. Check the action name may not be a good choice:

    - The action name is unique for each plugin, multiple plugins could do the same task, i.e., take CSI snapshot. If we check the action name, every plugin will need to add its name at here. Specifically, the current code could only filter out the action from Velero's CSI plugin
    - The action name is actually the plugin name, plugin could change this any time. Whenever it is changed, the code here need to be changed
2. Let's wrap these checking code into a function, which helps reduce the complexity of the main workflow

pkg/podvolume/backupper.go
```diff
@@ -201,6 +202,27 @@ func (b *backupper) BackupPodVolumes(backup *velerov1api.Backup, pod *corev1api.
 			continue
 		}
 
+		if resPolicies != nil {
```
Lyndon-Li, line 205: Same suggestion: Let's wrap these checking code into a function, which helps reduce the complexity of the main workflow

qiuming-best: updated

pkg/backup/item_backupper.go
qiuming-best: I'll write one item in the release notes to clarify that current resource policies are not working for third-party plugins

ywk253100 (approved):

ywk253100 (approved):

reasonerjt (approved):
I have some comments but they are not blocker for this PR, let's merge this PR.

Please sync up with @ywk253100 and submit follow-up PR for incremental changes

reasonerjt (approved):

ywk253100 (approved):

internal/resourcepolicies/resource_policies.go
```diff
@@ -0,0 +1,143 @@
...
+// Action defined as one action for a specific way of backup
+type Action struct {
+	// Type defined specific type of action, currently only support 'skip'
+	Type VolumeActionType `yaml:"type"`
+	// Parameters defined map of parameters when executing a specific action
+	Parameters map[string]interface{} `yaml:"parameters,omitempty"`
+}
+
+// VolumePolicy defined policy to conditions to match Volumes and related action to handle matched Volumes
+type VolumePolicy struct {
```
ywk253100, line 24: This type isn't used by other packages, right? So let's make it unexported.

internal/resourcepolicies/volume_resources.go
```diff
@@ -0,0 +1,201 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+	corev1api "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+type volumePolicy struct {
```
ywk253100, line 14: Make this type as exported

pkg/apis/velero/v1/backup_types.go
```diff
@@ -159,6 +160,9 @@ type BackupSpec struct {
 	// The default value is 1 hour.
 	// +optional
 	ItemOperationTimeout metav1.Duration `json:"itemOperationTimeout,omitempty"`
+	// ResourcePolicies specifies the referenced resource policies that backup should follow
+	// +optional
+	ResourcePolicies *v1.TypedLocalObjectReference `json:"resourcePolices,omitempty"`
```
ywk253100, line 165: Let's call it `ResourcePolicy`. Usually, complex means an object list. Change the command option as well

pkg/backup/item_backupper.go
```diff
@@ -300,7 +306,7 @@ func (ib *itemBackupper) backupPodVolumes(log logrus.FieldLogger, pod *corev1api
 		return nil, nil
 	}
 
-	return ib.podVolumeBackupper.BackupPodVolumes(ib.backupRequest.Backup, pod, volumes, log)
+	return ib.podVolumeBackupper.BackupPodVolumes(ib.backupRequest.Backup, pod, volumes, ib.backupRequest.ResPolicies, log)
```
ywk253100, line 309: The `Backup` and `ResPolicies` are both included in the `backupRequest`, so let's pass `ib.backupRequest` as the first parameter

internal/resourcepolicies/resource_policies.go
qiuming-best: it should be with capital character V, or the YAML file parsing can be problematic

pkg/backup/item_backupper.go
qiuming-best: if we use `BackupPodVolumes(backupRequest *pkgbackup.Request ...)`, it would be import cycle as follow:

package github.com/vmware-tanzu/velero/pkg/backup
	imports github.com/vmware-tanzu/velero/pkg/podvolume
	imports github.com/vmware-tanzu/velero/pkg/backup: import cycle not allowed

internal/resourcepolicies/volume_resources.go
qiuming-best: As this struct is used internally, it may not need to exported

pkg/apis/velero/v1/backup_types.go
qiuming-best: updated

PR #5926, opened 2023-02-27
Add server setting for resource timeouts

eemcmullan:
Thank you for contributing to Velero!

# Please add a summary of your change
Added a server config setting to use for any default timeouts.

# Does your change fix a particular issue?

Fixes #(issue)
#5604 

# Please indicate you've done the following:

- [ x] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [x ] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required` as a comment on this pull request.
- [x ] Updated the corresponding documentation in `site/content/docs/main`.

PR #5933, opened 2023-02-28
Wait for additional items to be ready before restoring current item

### LINKED ISSUE #1350: Race condition when passing AdditionalItems from restore item action

sseago:

Waits for additional items to be ready before restoring current item. This is needed to resolve a known race condition in restoring OpenShift ImageStreamTag resources.

Fixes #1350

The design is already approved for this feature (https://github.com/vmware-tanzu/velero/pull/2867 ), and the required Plugin API changes have already been merged. This just includes the necessary controller changes. 

### DIFF
--- a/changelogs/unreleased/5933-sseago
+++ b/changelogs/unreleased/5933-sseago
@@ -1,0 +1,1 @@
+Wait for additional items to be ready before restoring current item
--- a/pkg/restore/request.go
+++ b/pkg/restore/request.go
@@ -54,19 +54,24 @@ type Request struct {
 	PodVolumeBackups []*velerov1api.PodVolumeBackup
 	VolumeSnapshots  []*volume.Snapshot
 	BackupReader     io.Reader
-	RestoredItems    map[itemKey]string
+	RestoredItems    map[itemKey]restoredItemStatus
+}
+
+type restoredItemStatus struct {
+	action     string
+	itemExists bool
 }
 
 // RestoredResourceList returns the list of restored resources grouped by the API
 // Version and Kind
 func (r *Request) RestoredResourceList() map[string][]string {
 	resources := map[string][]string{}
-	for i, action := range r.RestoredItems {
+	for i, item := range r.RestoredItems {
 		entry := i.name
 		if i.namespace != "" {
 			entry = fmt.Sprintf("%s/%s", i.namespace, i.name)
 		}
-		entry = fmt.Sprintf("%s(%s)", entry, action)
+		entry = fmt.Sprintf("%s(%s)", entry, item.action)
 		resources[i.resource] = append(resources[i.resource], entry)
 	}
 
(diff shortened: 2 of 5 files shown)


sseago:
This is basically ready for review, but I've converted it to draft because we need https://github.com/vmware-tanzu/velero/pull/5926 merged first so that I can use the new `resourceTimeout` -- I'm currently using `resourceTerminatingTimeout`  as a placeholder.

shubham-pampattiwar (dismissed):

sseago:
Ready for final review now that the `resourceTimeout` change has been made.

shubham-pampattiwar (approved):

Lyndon-Li (approved):

Issue #5087, opened 2022-07-04
Restore stuck using restic - Volumes identified by podname


navilg:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is, and what commands you ran.)

I had deleted a deployment, pvc and a configmap in a namespace. Many pods running in same namespace controlled by a controller (deployment or rs) were restarted either by deleting pod directly or rollout restart. 

When I tried to restore the deployment, pvc and configmap which I had deleted before, they are restored but the status of restore still in inprogress. When I checked the details, I see that restic is trying to restore other volumes whose pods were restarted and their podname is changed after backup.

This can be reproduced using below steps:

1. Have velero installed and integrated with restic.
2. Create a backup.
3. Restart some of deployments so that new pods come up with new random generated name.
4. Delete one or two deployment, pvc/pv.
5. Try to restore the depleted resources.

This will try to restore other volumes as well whose pods were restarted and created with new names.

**What did you expect to happen:**

Velero/restic should have ignored other volumes. It should not be trying to restore volumes which are already present based on podname.

**The following information will help us better understand what's going on**:

_If you are using velero v1.7.0+:_  
Please use `velero debug  --backup <backupname> --restore <restorename>` to generate the support bundle, and attach to this issue, more options please refer to `velero debug --help` 

_If you are using earlier versions:_  
Please provide the output of the following commands (Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)
- `kubectl logs deployment/velero -n velero`
- `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
- `velero backup logs <backupname>`
- `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
- `velero restore logs <restorename>`

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Command used to restore:

```
velero create restore restore-v1.8.1-202207041327 --from-backup v1.8.1 --include-namespaces=ethan --exclude-resources=ingresses.extensions,ingresses.networking.k8s.io
```

Describe restore:

```
Name:         restore-v1.8.1-202207041327
Namespace:    velero
Labels:       <none>
Annotations:  <none>

Phase:                                 InProgress
Estimated total items to be restored:  239
Items restored so far:                 239

Started:    2022-07-04 07:57:43 +0000 UTC
Completed:  <n/a>

Backup:  v1.8.1

Namespaces:
[...]

PR #5760, opened 2023-01-12
Issue fix 5043

Lyndon-Li:
Fix issue #5043, after the restore pod is scheduled, check if the node-agent pod is running in the same node.

PR #5946, opened 2023-03-03
mark pvr failed when restore stuck in InProgress using restic.

Lanyujiex:

When restoring with restic, mark PVR as failed if pod is deleted

fix restore stuck in InProgress using restic.

Fixes #(issue)
5087
https://github.com/vmware-tanzu/velero/issues/5087

### DIFF
--- a/pkg/controller/pod_volume_restore_controller.go
+++ b/pkg/controller/pod_volume_restore_controller.go
@@ -99,9 +99,15 @@ func (c *PodVolumeRestoreReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		log = log.WithField("restore", fmt.Sprintf("%s/%s", pvr.Namespace, pvr.OwnerReferences[0].Name))
 	}
 
-	shouldProcess, pod, err := c.shouldProcess(ctx, log, pvr)
+	shouldProcess, markFailed, pod, err := c.shouldProcess(ctx, log, pvr)
 	if err != nil {
 		return ctrl.Result{}, err
+	}
+	if markFailed {
+		if e := UpdatePVRStatusToFailed(c, ctx, pvr, fmt.Sprintf("Pod  %s/%s is being Terminating", pvr.Spec.Pod.Namespace, pvr.Spec.Pod.Name), c.clock.Now()); e != nil {
+			log.WithError(err).Error("Unable to update status to failed")
+		}
+		return ctrl.Result{}, nil
 	}
 	if !shouldProcess {
 		return ctrl.Result{}, nil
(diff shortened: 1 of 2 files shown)


pkg/controller/pod_volume_restore_controller.go
```diff
@@ -151,30 +157,36 @@ func UpdatePVRStatusToFailed(c client.Client, ctx context.Context, pvr *velerov1
...
+func (c *PodVolumeRestoreReconciler) shouldProcess(ctx context.Context, log logrus.FieldLogger, pvr *velerov1api.PodVolumeRestore) (bool, bool, *corev1api.Pod, error) {
+	// We need to check pods status
 	// we filter the pods during the initialization of cache, if we can get a pod here, the pod must be in the same node with the controller
 	// so we don't need to compare the node anymore
 	pod := &corev1api.Pod{}
 	if err := c.Get(ctx, types.NamespacedName{Namespace: pvr.Spec.Pod.Namespace, Name: pvr.Spec.Pod.Name}, pod); err != nil {
 		if apierrors.IsNotFound(err) {
 			log.WithError(err).Debug("Pod not found on this node, skip")
-			return false, nil, nil
+			return false, false, nil, nil
```
Lyndon-Li, line 168: This will result in a false failing.
See the node-agent server code [here](https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/cli/nodeagent/server.go#:~:text=on%20this%20node.-,cacheOption,%7B,-SelectorsByObject), the client caches the pod that is scheduled in the current node only.
That means only one controller (in the node that the pod is scheduled) could get the restored pod, all other controllers get errorNotFound here.
If the code is changed like this, the PVRs will be marked as failed inevitably.

Lyndon-Li:
Furthermore, the problem the current PR is trying to fix has been addressed in PR #5760

pkg/controller/pod_volume_restore_controller.go
Lanyujiex: If pod is  not found,  it will not mark pvr failed. 
This commit mainly covers the updated scenarios after the deployment backup.
We use the old version to restore the deployment and pod. The old pod will be immediately deleted by the rs controller, which will cause the entire restore to be unable to continue. At this time, pvr will remain in InProgress until timeout or a pod with the same name is created.
I think we can consider this scenario。
Of course  I only tested it on 1.9， PR #5760  I didn't test this scenario, but I don't think it's very good。

Lyndon-Li: Yes, you are right, the PVR is marked as failed only when the pod is being deleted. Meanwhile, #5760 still covers the same pod deletion scenarios. Therefore, unless it is a scenario that #5760 doesn't cover, we don't want to involve duplicated logics, noticeably, #5760 fix several similar problems, not only the current one.

Lyndon-Li: For the case:

> We use the old version to restore the deployment and pod. The old pod will be immediately deleted by the rs controller, which will cause the entire restore to be unable to continue.

Could you explain more? As the Velero [restore policy](https://velero.io/docs/v1.10/restore-reference/#restore-existing-resource-policy:~:text=ns%2D1.-,Restore%20existing%20resource%20policy,-By%20default%2C%20Velero), Velero by default doesn't support to overwrite existing objects. Do you mean in the case you mentioned, something has escaped and broken this policy?

Lanyujiex: Sorry，you are right, my code will mark the completed pvr as failed. 
When I have time to upgrade my velero, go to test PR #5760 .

Issue #5443, opened 2022-10-13
Configurable repo password


Lyndon-Li:
See issue #1053 which is for Restic only, in v1.10 we introduced unified repo architecture to integrate with multiple backup repos, Velero still needs repo password set to various backup repos, therefore, this requirement is still reasonable.
Let's see the best way to expose the repo password configuration from end to end.

reasonerjt:
IMO it's good enough if we allow the user to reference the password via secret.

I don't think velero should add features to manage user's password.

Lyndon-Li:
As the current discussion, we will document the secret name and the steps to set the password

Lyndon-Li:
After further investigation, simply expose the secret to users by document it doesn't help:

- The secret is created by Velero during Velero server initialization, users can only modify the existing secret
- Velero doesn't support changing backup repository password, as a result, even if users change the value in the secret, the repo password is not changed
- As a result, once the secret value is changed, users still need the old password to connect the old repos and they need the new password to connect the new repos

Therefore, we need a well-round solution from which we may expose some user interface when creating the repo.

Lyndon-Li:
I am moving this issue out of 1.11 and let's find a future release to include the overall solution.

shubham-pampattiwar:
@Lyndon-Li Do we have plans to address this in velero 1.12 ?

Lyndon-Li:
We have discussed this in different occasions. The current situation is like:
1. It doesn't mean there is no way to config the password. Users could edit the secret, so as to tell Velero to use their own password. The secret name seems not be documented, if so, we need to document it.
2. In either way, there is only one global password for all BSLs. Ideally, one configurable password per BSL.
3. There are some concerns that the password cannot be dynamically configured during installation, so if they install multiple instances, users need to configure them one by one after the installation. However, we are hesitating to do this to Velero, because Velero doesn't have an appropriate way to keep the password safe for users, so we don't want to expose a parameter for users to provide a password then Velero keeps the password(as we know, the data in kubernetes secret is not actually encrypted)
4. The final solution is to support KMS. This involves quite some work, we don't have a plan to make it in v1.12

Lyndon-Li:
@shubham-pampattiwar
According to your current use case or requirement, what is your expectation in the 1~4 list above as to address this problem?

shubham-pampattiwar:
So lets start with documenting the secret name as you mentioned in point 1 for velero 1.12, And I believe point 2 is something that we are interested in as of now.

[...]

PR #5968, opened 2023-03-08
Volume Snapshot Data Movement Design

Lyndon-Li:
Add the design for Volume Snapshot Data Movement


reasonerjt:
I wish to suggest we put a section to introduce the higher level of the interaction between data movement controller and velero, and then zoom into the details of the internal data mover.

Such that other data mover providers will focus on the first section and the contract is more clarified.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,770 @@
...
+
+## Architecture of Volume Snapshot Data Movement
+
+## Workflows  
+
+Here are the diagrams that illustrate components and workflows for backup and restore respectively.  
+For backup, we intend to create an extensive architecture for various snapshot types, snapshot accesses and various data accesses. For example, the snapshot specific operations are isolated in Data Mover Plugin and Exposer. In this way, we only need to change the two modules for variations. Likely, the data access details are isolated into uploaders, so different uploaders could be plugged into the workflow smoothly.  
+
+For restore, we intend to create a generic workflow that could for all backups. This means the restore is backup source independent. Therefore, for example, we can restore a CSI snapshot backup to another cluster with no CSI facilities or a CSI driver the same as the source cluster.  
+We still have the Exposer module for restore and it is to expose the target volume to the data path. Therefore, we still have the flexibility to introduce different ways to expose the target volume.
```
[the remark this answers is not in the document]
Lyndon-Li: Yes, the PV is created by the data mover. This is clarified in the **Create Target PV** section, I have added one more line to clarify that Velero should not create the PV if data movement restore is involved.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,770 @@
...
+We still have the Exposer module for restore and it is to expose the target volume to the data path. Therefore, we still have the flexibility to introduce different ways to expose the target volume.  
+Likely, the data downloading details are isolated in uploaders, so we can still create multiple types of uploaders.  
+
+Below is the backup workflow:  
+![backup-workflow.png](backup-workflow.png)  
+
+Below is the restore workflow:  
+![restore-workflow.png](restore-workflow.png)  
+
+## Components
```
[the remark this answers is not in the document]
Lyndon-Li: Yes, we should make clear of the components that are built-in DM specific. I have modified the doc to clarify this.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,776 @@
...
+## Workflows  
+
+Here are the diagrams that illustrate components and workflows for backup and restore respectively.  
+For backup, we intend to create an extensive architecture for various snapshot types, snapshot accesses and various data accesses. For example, the snapshot specific operations are isolated in Data Mover Plugin and Exposer. In this way, we only need to change the two modules for variations. Likely, the data access details are isolated into uploaders, so different uploaders could be plugged into the workflow smoothly.  
+
+For restore, we intend to create a generic workflow that could for all backups. This means the restore is backup source independent. Therefore, for example, we can restore a CSI snapshot backup to another cluster with no CSI facilities or a CSI driver the same as the source cluster.  
+We still have the Exposer module for restore and it is to expose the target volume to the data path. Therefore, we still have the flexibility to introduce different ways to expose the target volume.  
+Likely, the data downloading details are isolated in uploaders, so we can still create multiple types of uploaders.  
+
+Below is the backup workflow:
```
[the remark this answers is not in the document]
Lyndon-Li: Yes, Exposer is a VBDM specific component.

yanji09 (changes requested):

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,776 @@
...
+- ```CSISnapshotSpec``` specifies the information of the CSI snapshot, e.g., ```volumeSnapshot``` is the name of VolumeSnapshot object representing the CSI snapshot; ```storageClass``` specifies the name of the StorageClass of the source PVC, which will be used to create the backupPVC during the data upload.
+
+For data mover manipulation:
+- ```datamover``` indicates the name of the data mover, if it is empty or ```velero```, it means the built-in data mover will be used for this data upload
+- Data movers may have their private data, they will be put in the ```dataMoverConfig``` map.
+
+For backup repository manipulation, ```backupStorageLocation``` is the name of the related BackupStorageLocation, where we can find all the required information.
+
+Here are the statuses of DataUpload CRD and their descriptions:
+- Accepted: the Object lock has been acquired for this DUCR and the elected controller is trying to expose the snapshot
```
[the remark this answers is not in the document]
Lyndon-Li: Fixed, thanks.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+              startTimestamp:
+                description: StartTimestamp records the time a backup was started.
+                  Separate from CreationTimestamp, since that value changes on restores.
+                  The server's time is used for StartTimestamps
+                format: date-time
+                nullable: true
+                type: string
+            type: object
+        type: object
+
```
shubham-pampattiwar, line 345: Also, I have a question regarding CSI Snapshots, Currently we delete the CSI Snapshots before the Async Plugins are triggered, will we be deferring the deletion of CSI Snapshots post completion of Async datamovement ? If we don't do that there won't be any CSI snapshot to upload.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+![cancel-sequence.png](cancel-sequence.png)    
+
+It is possible that a DM doesn’t support cancellation at all or only support in a specific phase (e.g., during InProgress phase), if the cancellation is requested at an unexpected time or to an unexpecting DM, the behavior is decided by the DMP and DM, below are some recommendations:
+- If a DM doesn't support cancellation at all, DMP should be aware of this, so the DMP could return an error and fail early
+- If the cancellation is requested at an unexpected time, DMP is possibly not aware of this, it could still deliver it to the DM, so both Velero and DMP wait there until DM completes the cancel request or timeout
+
+VBDM's cancellation exactly follows the above rules.  
+
+
+## Parallelism
```
shubham-pampattiwar, line 618: ++

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+In DUCR/DDCR’s status, we have fields like ```totalBytes``` and ```doneBytes```, the same values will be displayed as a result of below querires:
+- Call ```kubectl get dataupload -n velero xxx or kubectl get datadownload -n velero xxx```.
+- Call ```velero backup describe –details```. This is implemented as part of BIA/RIA V2, the above values are transferred to async operation and this command retrieve them from the async operation instead of DUCR/DDCR. See [general progress monitoring design][2] for details
+
+
+## Backup Sync
+DUCR contains the information that is required during restore but as mentioned above, it will not be synced because during restore its information is retrieved dynamically. Therefore, we have no change to Backup Sync.  
+
+## Backup Deletion
+Once a backup is deleted, the data in the backup repository should be deleted as well.
```
shubham-pampattiwar, line 646: +1 on the approach

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+
+## Plugin Data Movers
+There should be only one DM to handle a specific DUCR/DDCR in all cases. If more than one DMs process a DUCR/DDCR at the same time, there will be a disaster.  
+Therefore, a DM should check the dataMover field of DUCR/DDCR and process the CRs belong to it only.  
+For example, VBDM reconciles DUCR/DDCR with their dataMover field set to "" or "velero", it will skip all others.  
+This means during the installation, users are allowed to install more than one DMs, but the DMs should follow the above rule.   
+When creating a backup, we should allow users to specify the data mover, so a new backup CLI option is required.  
+For restore, we should retrieve the same information from the corresponding backup, so that the data mover selection is consistent.  
+
+At present, Velero doesn't have the capability to verify the existence of the specified data mover. As a result, if a wrong data mover name is specified for the backup or the specified data mover is not installed, nothing will fail earlly, DUCR/DDCR is still created and Velero will wait there until timeout.
```
shubham-pampattiwar, line 675: NIT: `earlly` <-> `early`

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+For restore, we should retrieve the same information from the corresponding backup, so that the data mover selection is consistent.  
+
+At present, Velero doesn't have the capability to verify the existence of the specified data mover. As a result, if a wrong data mover name is specified for the backup or the specified data mover is not installed, nothing will fail earlly, DUCR/DDCR is still created and Velero will wait there until timeout.  
+
+## Working Mode
+It doesn’t mean that once the data movement feature is enabled users must move every snapshot. We will support below two working modes:
+- Don’t move snapshots. This is same with the existing CSI snapshot feature, that is, native snapshots are taken and kept
+- Move snapshot data and delete native snapshots. This means that once the data movement completes, the native snapshots will be deleted.
+
+For this purpose, we need to add a new option in the backup command as well as the Backup CRD.
```
shubham-pampattiwar, line 682: Agreed, this option makes sense and help support existing functionality.

shubham-pampattiwar (commented):
Added a few questions/comments. Overall well thought design, Thank you @Lyndon-Li !!

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+- The current workflow focuses on snapshot-based data movements. For some application/SaaS level data sources, snapshots may not be taken explicitly. We don’t take them into consideration, though we believe that some workflows or components may still be reusable.
+
+## Architecture of Volume Snapshot Data Movement
+
+## Workflows  
+
+Here are the diagrams that illustrate components and workflows for backup and restore respectively.  
+For backup, we intend to create an extensive architecture for various snapshot types, snapshot accesses and various data accesses. For example, the snapshot specific operations are isolated in Data Mover Plugin and Exposer. In this way, we only need to change the two modules for variations. Likely, the data access details are isolated into uploaders, so different uploaders could be plugged into the workflow seamlessly.  
+
+For restore, we intend to create a generic workflow that could for all backups. This means the restore is backup source independent. Therefore, for example, we can restore a CSI snapshot backup to another cluster with no CSI facilities or a CSI driver the same as the source cluster.
```
blackpiglet, line 45: `we can restore a CSI snapshot backup to another cluster with no CSI facilities or a CSI driver **the same as** the source cluster.`

`the same as` part is bit ambiguity to me. In the context, should it be `different from`?

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+
+Below is the restore workflow:  
+![restore-workflow.png](restore-workflow.png)  
+
+## Components
+Below are the generic components in the data movement workflow:  
+
+**Velero**: Velero controls the backup/restore workflow, it calls BIA/RIA V2 to backup/restore an object that involves data movement, specifically, a PVC or a PV.  
+**BIA/RIA V2**: BIA/RIA V2 are the protocols between Velero and the data mover plugins. They support asynchronized operations so that Velero backup/restore is not marked as completion until the data movement is done and in the meantime, Velero is free to process other backups during the data movement.  
+**Data Mover Plugin (DMP)**: DMP implement BIA/RIA V2 and it invokes the corresponding data mover by creating the DataUpload/DataDownload CRs. DMP is also responsible to take snapshot of the source volume, so it is a snapshot type specific module. For CSI snapshot data movement, the CSI plugin could be extended as a DMP.
```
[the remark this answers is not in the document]
blackpiglet: My understanding is there should be a new plugin created for VBDM, and the current CSI plugin could be a base to create the new VBDM, so there still is only one Velero CSI plugin.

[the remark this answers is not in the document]
Lyndon-Li: We incline to take the first action because V2 is the new standard and it is a good opportunity (modify for data mover) to port the CSI plugin with the new standard. However, if we have some strong requirement against this, it is OK to take the second action technically, so if any, please advice.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: This is a good question, some data movers want to keep the VS created by CSI plugin along with the async operation, some want to backup them into the backup tarball, some don't need the origin VS created by CSI plugin, how do we support all of them? This comes down to two question:

1. How do we make the flexibility for VS/VSC deletion?
2. How do we make the flexibility for VS/VSC persistency?

Lyndon-Li: First of all, the origin VS/VSC is deleted by data mover, not Velero, so data mover could decide whether to delete it or not. Therefore, for question 1, we actually don't have a problem.

Let me clarify this in the design.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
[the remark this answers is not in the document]
Lyndon-Li: Let me clarify it in the design as the first action, if we finally want to change it, we come back to change the design.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+
+Here are the statuses of DataUpload CRD and their descriptions:
+- New: The DUCR has been created but not processed by a controller
+- Accepted: the Object lock has been acquired for this DUCR and the elected controller is trying to expose the snapshot
+- Prepared: the snapshot has been exposed, the related controller is starting to process the upload
+- InProgress: the data upload is in progress
+- Canceling: the data upload is being canceled
+- Canceled: the data upload has been canceled
+- Completed: the data upload has completed
+- Failed: the data upload has failed
```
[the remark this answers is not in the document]
Lyndon-Li: Let me clarify it in the design, if we finally want to change it, we come back to change the design.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: For question 2
As the current implementation of CSI plugin, the VS and VSC are returned as `AdditionalItems` by CSI plugin.
If we keep the current behavior， the VS/VSC will be persisted by Velero, this meets the requirements of the data movers that wants to persist them.
However, this doesn't meet the requirements of the data movers that don't want to persist them.
The current POC code just stops returning VS as additionalItems, so that nothing will be persisted by Velero.

Therefore, the behavior actually differs from data movers. If we have a data mover registration mechanism, it will be easier to solve the problem, but since we don't have at present, I thought we need some hard code thing (e.g., check the DataMover type of the Backup CR in the plugin), so this was not included in the design. Let me further think about this and get back later.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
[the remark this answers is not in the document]
Lyndon-Li: Done for the clarification, check line 60:

> this also means that the CSI plugin will fully implement BIA/RIA V2

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+After ```Execute``` returns, the backup is set to ```WaitingForPluginOperations```, and then Velero is free to process other items or backups.  
+Before Velero moves to other items/backups, it will persist the backup data. This is the same as the existing behavior.  
+The backup then is left as ```WaitForAsyncOperations``` until the DM completes or timeout.  
+
+**BIA Progress**  
+Velero keeps monitoring the status of the backup by calling BIA V2’s Progress method. Below is the Progress method from BIA V2:
+```
+Progress(operationID string, backup *api.Backup) (velero.OperationProgress, error)
+```
+On the call of this method, DMP will query the DUCR’s status. Some critical progress data is transferred from DUCR to the ```OperationProgress``` which is the returen value of BIA V2’s Progress method. For example, NCompleted indicates the size/number of data that have been completed and NTotal indicates the total size/number of data.
```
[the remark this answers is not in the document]
Lyndon-Li: Corrected, thanks.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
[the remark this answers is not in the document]
Lyndon-Li: Done the clarification, check line 126:

> The CR will be left there after the backup completes

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: Done for the clarification of the 1st question, check line 127:

> The DMP leaves the snapshot as well as its related objects to DM for further processing

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: Yes, corrected it. Thanks.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: Corrected, thanks.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+## DataUpload CRD
+Below are the essential fields of DataUpload CRD. The CRD covers below information:
+- The information to manipulate the specified snapshot
+- The information to manipulate the specified data mover
+- The information to manipulate the specified backup repository
+
+For snapshot manipulation:
+- ```snapshotType``` indicates the type of the snapshot, at present, the only valid value is ```CSI```.  
+- If ```snapshotType``` is ```CSI```, ```csiSnapshot``` which is a pointer to a ```CSISnapshotSpec``` must not be absent. 
+- ```CSISnapshotSpec``` specifies the information of the CSI snapshot, e.g., ```volumeSnapshot``` is the name of VolumeSnapshot object representing the CSI snapshot; ```storageClass``` specifies the name of the StorageClass of the source PVC, which will be used to create the backupPVC during the data upload.
```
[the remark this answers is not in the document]
Lyndon-Li: Done

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: For question 2
As the current implementation of CSI plugin, the VS and VSC are returned as `AdditionalItems` by CSI plugin.
If we keep the current behavior，the VS/VSC will be persisted by Velero, this meets the requirements of the data movers that wants to persist them.
However, this doesn't meet the requirements of the data movers that don't want to persist them.
The current POC code just stops returning VS as additionalItems, so that nothing will be persisted by Velero.

Therefore, the behavior actually differs from data movers. So the first question is, are there some data movers require to persist the VS and VSC as the current plugin does?
If the answer is No, we can just take the POC code approach -- stop returning VS as additionalItems
If the answer is Yes, I suggest the following approach:

- We add a configMap named the same as the data mover
- The plugin checks the configMap to get a flag something like "persist-csi-snapshot-metadata"
- Only if the flag is set, the plugin returns VS as additionalItems

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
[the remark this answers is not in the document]
sseago: @Lyndon-Li Keep in mind that with v2 plugins, any plugins that kick off asynchronous processes need to be included in the original backup processing (InProgress stage via backup controller). i.e. if we make CSI plugin v2 and also introduce a returned operationID, any new resources created by this plugin that are added during finalize can't have their own async operations -- however, if the CSI plugin returns an AdditionalItem as a regular backup item to be processed now, then that item may have an async plugin, since it will be processed during the initial InProgress stage, not during Finalize. The issue is that once we hit Finalize, we're done waiting for plugin async operations, so if a finalize item's plugin returns an operationID, it will raise an error.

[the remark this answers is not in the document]
Lyndon-Li: @sseago Thanks for reminding.
This situation could fit to the current data movement case, specifically:
1. After CSI plugin changes to V2, it still works on PVC, VS, VSC, etc., it won't work on DataUpload/DataDownload CR
2. If the data movement is not enabled, nothing is changed, no operationID is returned, VS, VSC are returned as additonalItems as is
3. If the data movement is enabled, an operationID is returned and a DataUpload/DataDownload CR is returned as itemToFinalize. However, DataUpload/DataDownload CR right meets your above statements -- they don't need to go through the regular backup process and the CSI plugin won't handle them on finalize phase, so no further operationID is returned for them during finalize
4. If the data movement is enabled, depend on the discussion[ here](https://github.com/vmware-tanzu/velero/pull/5968#discussion_r1151478358), if we still need to backup VS and VSC, we still return them as regular additionalItems, so they will be processed as is

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,763 @@
...
+
+**Expose**  
+For some kinds of snapshot, it may not be usable directly after it is taken. For example, a CSI snapshot is represented by the VolumeSnapshot and VolumeSnapshotContent object, if we don’t do anything, we don’t see any PV really exists in the cluster, so VGDP has no way to access it. Meanwhile, when we have a PV representing the snapshot data, we still need a way to make it accessible by the VGDP.  
+The details of the expose process are snapshot specific, and for one kind of snapshot, we may have different methods to expose it to VGDP. Later, we will have a specific section to explain the current design of the Exposer.  
+
+**Backup From Data Path**  
+After a snapshot is exposed, VGDP will be able to access the snapshot data, so the controller calls the uploader to start the data backup.  
+To support cancellation and concurrent backup, the call to the VGDP is done asynchronously. How this asynchronization is implemented may be related to the Exposer. as the current design of Exposer, the asynchronization is implemented by the controller with go routines.
```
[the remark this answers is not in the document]
shubham-pampattiwar: yes seems good for the first phase.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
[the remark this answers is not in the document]
shubham-pampattiwar: ok, but consider this scenario, you create a backup of ns that has 100 pvc, this means velero creates 100 DataUpload CRs for that backup. Now if you created another backup of the same ns, won't velero be backing up the old 100 DataUpload CRs from older backup as well ? IMO thats a lot a irrelevant resources getting dumped in the user's repository. How should we solve this case ?

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
shubham-pampattiwar: For the 1st question, currently Velero backup controller deletes the CSI Snapshots here: https://github.com/vmware-tanzu/velero/issues/5443 
This delete action is taken before the Async Actions are processed, So will this be removed for datamover feature? Or lets say deferred to the end of backup to Finalize backup controller or something ?

For question 2, I am fine with the configmap approach.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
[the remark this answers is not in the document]
Lyndon-Li: The DataUpload CRs are created in Velero ns, so they will never be backed up.
The CRs are just left in the cluster that the backups happened, they will not be synced to a new cluster either.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: For the 1st question, I think you are mention the [code here](https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/backup_controller.go#:~:text=feature%20is%20enabled.-,if,%7B,-b) , right?
If so, yes, we will modify the code to add something like `if features.IsEnabled(velerov1api.CSIFeatureFlag) && boolptr.IsSetToFalse(backup.Spec.SnapshotMoveData)` so that the VS and VSC will be handled by the DM instead of the backup controller.

Lyndon-Li: For question 2, I double checked it, we cannot solve the problem by simply adding a configMap, because the DM will create VS in Velero namespace and also create a new VSC, the original VS in user namespace will be deleted. This may happens at the same time when Velero backs the objects in DM's async operation.

Therefore, if it is not a must-have to persist the VS (in user namespace) and VSC, we will not do it. Moreover, we recommend plugin DMs to follow the generic restore workflow (as depicted in the current design) where the VS and VSC are not required during restore. If so, the behaviors are as below:

- If `--snapshot-move-data` is set, the BIA V2 for PVC doesn't return VS as an additionalItem, so VS and VSC will not be backed up/restored
- If `--snapshot-move-data` is not set, the BIA V2 for PVC returns VS as an additionalItem, which is the same as the current CSI plugin behavior

@shubham-pampattiwar Let me know your thoughts

shubham-pampattiwar: I think we should be good with that

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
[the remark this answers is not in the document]
shubham-pampattiwar: yes I think its good for first phase

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: As the discussion in the meeting:

- If --snapshot-move-data is not set, the BIA V2 for PVC returns VS as an additionalItem, which is the same as the current CSI plugin behavior
- If --snapshot-move-data is set, the BIA V2 for PVC doesn't return VS as an additionalItem, so that Velero will not persist them. How and when to dispose the volume snapshot as well as the VS and VSC are totally handled by the DM
- The restore workflow in the current design is generic enough for plugin DMs to adopt, so it means, the VS and VSC do need to be persisted necessarily

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,779 @@
...
+
+Conclusively, as a result of the above executions:
+- A DataUpload CR is created and persisted to the backup tarball. The CR will be left there after the backup completes because the CR includes many information connecting to the backup that may be useful to end users or upper level modules.  
+- A snapshot as well as the objects representing it are created. For CSI snapshot, a VolumeSnapshot object and a VolumeSnapshotContent object is created. The DMP leaves the snapshot as well as its related objects to DM for further processing.  
+
+VBDM creates a Data Uploader Controller to handle the DUCRs in node-agent daemonset, therefore, on each node, there will be an instance of this controller. The controller connects to the backup repository and calls the uploader. Below are the VBDM actions.  
+
+**Acquire Object Lock**  
+**Release Object Lock**  
+There are multiple instances of Data Uploader Controllers and when a DUCR is created, there should be only one of the instances handle the CR.
```
blackpiglet, line 134: The Data Uploader Controllers should be implemented in the Node-Agent, so the CR should be handled by the Node-Agent pod that shares the same node as the uploading volume, then there is only one candidate, and there is no need to have the lock.

Lyndon-Li: Actually, the Data Uploader Controller starts to process the DUCR once it is created (its phase is New or ""), then the controller creates the backupPVC, so it means the early phases before the backupPV is provisioned need the lock to make a consensus of which controller is responsible to create the backupPVC.

Lyndon-Li:
@shubham-pampattiwar 
After a discussion, we think the workflow for the data mover backup deletion is too complicated, since a DataUploadDelete CRD and controller need to be involved.
Therefore, I modified the workflow in the design, please help to review it and let us know for any concerns or suggestions.

@reasonerjt @ywk253100 Please also help to review the same.

shubham-pampattiwar:
@Lyndon-Li Can we not just DeleteItemAction plugin for DataUploadCRs ? so whenever a backup is deleted the CRs also get deleted.

Lyndon-Li:
@shubham-pampattiwar 
> Can we not just DeleteItemAction plugin for DataUploadCRs ? so whenever a backup is deleted the CRs also get deleted.

Firstly, The aim here is to delete the backup data stored in the backup repo. And only the specific DM knows where the backup data is and how to delete it. So, the question is how do we let the DM know the backup is to be deleted.
Secondly, we cannot let the controller to monitor the deletion event of DataUpload CRs, because `kubectl delete DataUpload -n velero` could also generate the same event, but we don't want delete the backup data once some user run this command by mistake.

Therefore, Velero needs a private mechanism to notify the DM when it handling a `deletebackup` request, then the DM does its own work to delete the backup data

shubham-pampattiwar:
> @shubham-pampattiwar
> 
> > Can we not just DeleteItemAction plugin for DataUploadCRs ? so whenever a backup is deleted the CRs also get deleted.
> 
> Firstly, The aim here is to delete the backup data stored in the backup repo. And only the specific DM knows where the backup data is and how to delete it. So, the question is how do we let the DM know the backup is to be deleted. Secondly, we cannot let the controller to monitor the deletion event of DataUpload CRs, because `kubectl delete DataUpload -n velero` could also generate the same event, but we don't want delete the backup data once some user run this command by mistake.
> 
> Therefore, Velero needs a private mechanism to notify the DM when it handling a `deletebackup` request, then the DM does its own work to delete the backup data

@Lyndon-Li Thanks for the elaborate explanation, I think I understand now, got confused with the deletion of in-cluster CRs vs Data in backup repository. The modified delete workflow looks sane to me :+1:

reasonerjt (approved):
Let's approve it.

If we find minor changes required during the implementation, let's make sure they are also reflected in incremental change to this doc.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,779 @@
...
+```
+Besides ```additionalItem``` (as the 2nd return value), Execute method will return one more resource list called ```itemToUpdate```, which means the items to be updated  and persisted when the async operation completes. For details, visit [general progress monitoring design][2].   
+Specifically, this mechanism will be used to persist DUCR into the persisted backup data, in another words, DUCR will be returned as ```itemToUpdate``` from Execute method. DUCR contains all the information the restore requires, so during restore, DUCR will be extracted from the backup data.  
+Additionally, in the same way, a DMP could add any other items into the persisted backup data.  
+Execute method also returns the ```operationID``` which uniquely identifies the asynchronized operation. This ```operationID``` is generated by plugins. The [general progress monitoring design][2] doesn't restrict the format of the ```operationID```, for Velero CSI plugin, the ```operationID``` is a combination of the backup CR UID and the source PVC (represented by the ```item``` parameter) UID.
```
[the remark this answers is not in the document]
Lyndon-Li: Actually, the progress monitoring design doesn't have a retry mechanism itself, so whether or not a retry happens is decided by the DM. At present, VBDM doesn't have retry mechanism.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,779 @@
...
+
+Below is the restore workflow:  
+![restore-workflow.png](restore-workflow.png)  
+
+## Components
+Below are the generic components in the data movement workflow:  
+
+**Velero**: Velero controls the backup/restore workflow, it calls BIA/RIA V2 to backup/restore an object that involves data movement, specifically, a PVC or a PV.  
+**BIA/RIA V2**: BIA/RIA V2 are the protocols between Velero and the data mover plugins. They support asynchronized operations so that Velero backup/restore is not marked as completion until the data movement is done and in the meantime, Velero is free to process other backups during the data movement.  
+**Data Mover Plugin (DMP)**: DMP implement BIA/RIA V2 and it invokes the corresponding data mover by creating the DataUpload/DataDownload CRs. DMP is also responsible to take snapshot of the source volume, so it is a snapshot type specific module. For CSI snapshot data movement, the CSI plugin could be extended as a DMP, this also means that the CSI plugin will fully implement BIA/RIA V2 and support some more methods like Progress, Cancel, etc.
```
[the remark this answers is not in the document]
Lyndon-Li: This requirement could be met technically, the implementation will be like:

- The DMP we are talking about in this design exposes a BIA for a PVC, which takes snapshot of the PVC and the submit a DUCR to launch a DM
- Another kind of plugin (we call it DMP or not) could exposes a BIA for a VS (wherever the VS comes from), then this kind of DMP doesn't take snapshot but directly launch a DM

The current design doesn't cover the second case, if required, we can add in an incremental design.

sseago (approved):

shubham-pampattiwar (approved):
LGTM ! Thank you @Lyndon-Li

yanji09 (approved):

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -0,0 +1,772 @@
...
+velero backup create xxx --include-namespaces --snapshot-move-data --data-mover xxx-plugin-dm
+```
+
+Restore command is kept as is.  
+
+
+
+
+[1]: ../unified-repo-and-kopia-integration/unified-repo-and-kopia-integration.md
+[2]: ../general-progress-monitoring.md
```
jglick, line 772: Broken link; now at https://github.com/vmware-tanzu/velero/blob/main/design/Implemented/general-progress-monitoring.md (missing `implemented/` infix)

Lyndon-Li: Thanks for noticing this.
I tried to add the `Implemented` infix, but finally realized that we cannot do this. At the release time of 1.12, we will also move this design to the `Implemented` folder (the same to the unified-repo-and-kopia-integration folder), then all the links will be fixed.

Therefore, let's leave it as is for now, and it will be fixed itself at the release time.

Issue #5155, opened 2022-07-27
Persistent volume "succeeding backup" when in fact it hasn't


djs28:
Noticed that our backups on persistent volumes were not being snapshotted correctly, even when backups looked correct. Hidden in the logs 
`level=info msg="No volume ID returned by volume snapshotter for persistent volume"`

I believe that 			
`log.Infof("No volume ID returned by volume snapshotter for persistent volume")`

should be
`log.WithError(err).Errorf("No volume ID returned by volume snapshotter for persistent volume")`

https://github.com/vmware-tanzu/velero/blob/f68ae92fd03559515497ae3549ed694ae7a7fcf6/pkg/backup/item_backupper.go#L467

Lyndon-Li:
@djs28 
Please provide us more information and the Velero logs. If the snapshot was not completed correctly, the handler of the snapshot will return errors, so it should not go to the log you mentioned. Thanks.

qiuming-best:
I've reproduced it by using AWS plugin to do the volume snapshot for one self-build cluster. 

And the backup result is in complete status but does not snapshot the volume, and it has no error log. So here I've changed the above mentioned info log into error log for easy troubleshooting.

qiuming-best:
@djs28 could you share me what plugin you are using for snapshot?

PR #5982, opened 2023-03-14
Fix take snapshot for PV with empty volume ID

qiuming-best:

Fixes #(issue)
Fixes https://github.com/vmware-tanzu/velero/issues/5155

I've reproduced it by using AWS plugin to do the volume snapshot for one self-build cluster.

And the backup result is in complete status but does not snapshot the volume, the volume is not been backed up in fact.

Plugin could not get the volume ID, but there is no error output, and the volume is not been snapshotted, but the backup is complete, which is misleading. for the issue we should output one error logs and the backup would be `PartiallyFailed`

### DIFF
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -502,7 +502,7 @@ func (ib *itemBackupper) takePVSnapshot(obj runtime.Unstructured, log logrus.Fie
 			continue
 		}
 		if volumeID == "" {
-			log.Infof("No volume ID returned by volume snapshotter for persistent volume")
+			log.Warn("No volume ID returned by volume snapshotter for persistent volume")
 			continue
 		}
 


pkg/backup/item_backupper.go
```diff
@@ -502,7 +502,7 @@ func (ib *itemBackupper) takePVSnapshot(obj runtime.Unstructured, log logrus.Fie
 			continue
 		}
 		if volumeID == "" {
-			log.Infof("No volume ID returned by volume snapshotter for persistent volume")
+			log.Error("No volume ID returned by volume snapshotter for persistent volume")
```
Lyndon-Li, line 505: This part logic is delicate. I am afraid this change may break other cases. For example, for vsphere-plugin, the volume is handled up by vsphere-plugin, but the workflow will also come here, but should not treated as errors.

Just one example, we should be careful to change the logic here unless we are sure this work for all cases.

qiuming-best: For vsphere-plugin, they used one tricky way to avoid calling the CreateSnapshot volumeSnapshotter interface by returning empty `volumeID` in advance and they used their own way of snapshot for volume, which is not volumeSnapshotter regular workflow. so we need one exceptional handle for the empty volumeSnapshotter when using vsphere plugin. but for other plugins, it should output an error and backup partially failed or it will be misleading for user that their backup is in complete status but volume is not been backed up.

blackpiglet (approved):

Lyndon-Li (approved):

PR #6009, opened 2023-03-21
Adjust match resource policies

qiuming-best:

Fixes #(issue)


internal/resourcepolicies/resource_policies.go
```diff
@@ -78,26 +85,49 @@ func (policies *Policies) buildPolicy(resPolicies *resourcePolicies) error {
...
+			isAllMatch = true
+		}
+		if isAllMatch {
+			return &policy.action
 		}
 	}
 	return nil
 }
 
+func (p *Policies) MatchResourcePolicies(res interface{}, resType ResourceType) (*Action, error) {
```
reasonerjt, line 105: 1. `MatchResourcePolicies` is a little repetitive, you may just call it `MatchResource` if `Match` is unclear.
2. Doesn't `res` contain the information of `resType`?

pkg/backup/item_backupper.go
```diff
@@ -579,8 +578,8 @@ func (ib *itemBackupper) takePVSnapshot(obj runtime.Unstructured, log logrus.Fie
 	return kubeerrs.NewAggregate(errs)
 }
 
-func (ib *itemBackupper) checkResourcePolicies(obj runtime.Unstructured, groupResource *schema.GroupResource, actionName string) (*resourcepolicies.Action, error) {
-	if ib.backupRequest.ResPolicies != nil && groupResource.String() == "persistentvolumeclaims" && actionName == "velero.io/csi-pvc-backupper" {
+func (ib *itemBackupper) matchResourcePolicies(obj runtime.Unstructured, groupResource schema.GroupResource, backupItemActionName string) (*resourcepolicies.Action, error) {
```
reasonerjt, line 581: Trivial, but
I don't quite understand why this func is needed.  The struct `Polices` should be extensible so it can match all kinds of resources.
And if the `match` function returns `skip` action you should check the name in the for-loop of triggering BIA rather than this func.

such that his func is unnecessary?

internal/resourcepolicies/resource_policies.go
qiuming-best: res is one interface type we need to try all possible type conventions.

Add `resType` as I want the type convention precisely, may it not be graceful in function, but it could reduce the type conversions frequency.

pkg/backup/item_backupper.go
qiuming-best: our `Policies` struct has the function [MatchResourcePolicies](https://github.com/vmware-tanzu/velero/blob/c93454019024ac97dec92bb58a7c9d185295cf3a/internal/resourcepolicies/resource_policies.go#L103) that could match all kinds of resources. but currently, we only have `PV` and `Pod Volume` to handle. If we have other resources that want to match, we could extend the `MatchResourcePolicies` function.

Here why we need matchResourcePolicies b/c we need to use kubeClient to find PVC-related PV. and then call the MatchResourcePolicies function

Another place is [matchResourcePolicies](https://github.com/vmware-tanzu/velero/blob/c93454019024ac97dec92bb58a7c9d185295cf3a/pkg/backup/item_backupper.go#L581), also it needs to find PVC-related PV, but it used `pvClient corev1client.PersistentVolumesGetter` kinds of client to find PV.
and then call the MatchResourcePolicies function

So it's not easy to uniform interface further. and this's why we have two matchResourcePolicies function in itemBackupper and backupper before we call  MatchResourcePolicies function

internal/resourcepolicies/resource_policies.go
qiuming-best: In fact, it returns an Action by inputting one resource, What about `MatchAction` or `GetMatchAction` is more suitable?

qiuming-best:
I've checked all possible unnecessary exported structs into unexported. @ywk253100

internal/resourcepolicies/resource_policies.go
reasonerjt: How about transform it to unstructured object first and do the transformation based on`GetObjectKind`?

pkg/backup/item_backupper.go
reasonerjt: > Here why we need matchResourcePolicies b/c we need to use kubeClient to find PVC-related PV. and then call the MatchResourcePolicies function

Isn't this a common logic that whenever it needs to filter a PVC it needs to check the PV also?

qiuming-best: yes, for handling PVC we need to check related PV **through k8s client,** but we have two different clients(kbclient, client-go) currently,  so we could not use one unified function to get matched action for PVC.

So here used add another logic that for PVC we first get related PV, then through PV resource get matched action.

internal/resourcepolicies/resource_policies.go
qiuming-best: it will make the function more graceful, but we call the `MatchResourcePolicies` we already have the structured object, we need unstructured when calling the function, and structured it again inner. Maybe it's not effecient

pkg/apis/velero/v1/backup_types.go
```diff
@@ -160,9 +160,9 @@ type BackupSpec struct {
 	// The default value is 1 hour.
 	// +optional
 	ItemOperationTimeout metav1.Duration `json:"itemOperationTimeout,omitempty"`
-	// ResourcePolicies specifies the referenced resource policies that backup should follow
+	// ResourcePolicy specifies the referenced resource policies that backup should follow
 	// +optional
-	ResourcePolicies *v1.TypedLocalObjectReference `json:"resourcePolices,omitempty"`
+	ResourcePolicy *v1.TypedLocalObjectReference `json:"resourcePolices,omitempty"`
```
ywk253100, line 165: The JSON annotation should also be updated to `resourcePolicy`

pkg/backup/item_backupper.go
```diff
@@ -484,10 +484,9 @@ func (ib *itemBackupper) takePVSnapshot(obj runtime.Unstructured, log logrus.Fie
 	}
 
 	if ib.backupRequest.ResPolicies != nil {
-		structuredVolume := &resourcepolicies.StructuredVolume{}
-		structuredVolume.ParsePV(pv)
-		action := ib.backupRequest.ResPolicies.Match(structuredVolume)
-		if action != nil && action.Type == resourcepolicies.Skip {
+		if action, err := ib.backupRequest.ResPolicies.GetMatchAction(pv, resourcepolicies.PersistentVolumes); err != nil {
+			log.WithError(err).Errorf("Error getting matched resource policies for pv %s", pv.Name)
```
ywk253100, line 488: Should we return directly or just log the error?

internal/resourcepolicies/resource_policies.go
```diff
@@ -69,46 +74,69 @@ func (policies *Policies) buildPolicy(resPolicies *resourcePolicies) error {
...
+			isAllMatch = true
+		}
+		if isAllMatch {
+			return &policy.action
 		}
 	}
 	return nil
 }
 
+func (p *Policies) GetMatchAction(res interface{}, resType ResourceType) (*Action, error) {
```
ywk253100, line 103: You can get the type of the interface by type assertion and don't need to pass the `resType` explicitly

pkg/apis/velero/v1/backup_types.go
qiuming-best: updated

pkg/backup/item_backupper.go
qiuming-best: yes, I've changed to  return directly

internal/resourcepolicies/resource_policies.go
qiuming-best: ok, I've removed resType

internal/resourcepolicies/resource_policies.go
qiuming-best: I've removed `resType`

ywk253100 (dismissed):

ywk253100 (approved):

reasonerjt (approved):

PR #6041, opened 2023-03-27
Fixed backup deletion bug related to async operations

### LINKED ISSUE #6035: [BIA V2] Backups with in progress async operation could should not be allowed to delete
There is an existing logic in Velero to prevent in progress backups from being deleted:
https://github.com/vmware-tanzu/velero/blob/1d8ca4f2efc512014c14e2e6ea5d6a7980d60fd3/pkg/controller/backup_deletion_controller.go#L160

However, this doesn't work for backups with in progress async operation since at the time of the async operation starts, the normal backup progress quits so `backupTracker` is cleared.

Actually, the backups in the `WaitingForPluginOperations` phase should not be allowed to be deleted as well, otherwise, the ongoing async operation will be orphan.

sseago:

Remove backup from backup tracker in finalize rather than in backup controller.

Fixes #6035

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -267,7 +267,12 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	b.backupTracker.Add(request.Namespace, request.Name)
-	defer b.backupTracker.Delete(request.Namespace, request.Name)
+	defer func() {
+		switch request.Status.Phase {
+		case velerov1api.BackupPhaseCompleted, velerov1api.BackupPhasePartiallyFailed, velerov1api.BackupPhaseFailed, velerov1api.BackupPhaseFailedValidation:
+			b.backupTracker.Delete(request.Namespace, request.Name)
+		}
+	}()
 
 	log.Debug("Running backup")
 
--- a/pkg/controller/backup_finalizer_controller.go
+++ b/pkg/controller/backup_finalizer_controller.go
@@ -44,6 +44,7 @@ type backupFinalizerReconciler struct {
 	clock             clocks.WithTickerAndDelayedExecution
 	backupper         pkgbackup.Backupper
 	newPluginManager  func(logrus.FieldLogger) clientmgmt.Manager
+	backupTracker     BackupTracker
 	metrics           *metrics.ServerMetrics
 	backupStoreGetter persistence.ObjectBackupStoreGetter
 	log               logrus.FieldLogger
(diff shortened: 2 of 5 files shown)


pkg/controller/backup_controller.go
```diff
@@ -267,7 +267,6 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	b.backupTracker.Add(request.Namespace, request.Name)
-	defer b.backupTracker.Delete(request.Namespace, request.Name)
```
Lyndon-Li, line 270: Previously, the behavior is like --- as long as the execution reaches here, and when it quits for whatever reason, the backupTracker will always be deleted.
Now, if we change like this, if someone insert some code between line 270 and 290 to return earlier, the backupTracker will not be deleted for that code. As a result, the backup will never be deletable for that case.

pkg/controller/backup_finalizer_controller.go
```diff
@@ -165,6 +168,7 @@ func (r *backupFinalizerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 			return ctrl.Result{}, errors.WithStack(err)
 		}
 	}
+	defer r.backupTracker.Delete(backup.Namespace, backup.Name)
```
Lyndon-Li, line 171: This is similar to the above comment.
If the finalizer reconcile error out before here at line 171, the backupTracker will not be deleted, as a result, the backup will never be deletable.

pkg/controller/backup_controller.go
Lyndon-Li: One suggest here is that we still use the defer function, but inside it, we check the phase is not `WaitForAsyncOperation` then we delete the backupTracker.

sseago: @Lyndon-Li That works too. I wanted to make sure we only closed it for terminal phases, but you're right -- putting that check in the defer func is cleaner. I'll update the PR.

pkg/controller/backup_finalizer_controller.go
sseago: In this case, that is deliberate, as the backup won't be in a terminal state if the  func errors out here. However, I'll do the same here as above -- defer statement early, but check for terminal states before removing.

pkg/controller/backup_finalizer_controller.go
```diff
@@ -102,6 +105,10 @@ func (r *backupFinalizerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	original := backup.DeepCopy()
 	defer func() {
+		switch original.Status.Phase {
```
Lyndon-Li, line 108: I think we need to check `backup.Status.Phase`? because this one is to be updated in the code following

Lyndon-Li (approved):

reasonerjt (approved):

PR #6107, opened 2023-04-11
Exclude endpointslice from restore list

ywk253100:
The restored endpointslices are removed immediately and that causes the following patch operation for managedFields failed. The control plane automatically creates EndpointSlices for any Kubernetes Service that has a selector specified. So don't need to restore it

Fixes #(issue)

### DIFF
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -60,6 +60,11 @@ var nonRestorableResources = []string{
 	"events",
 	"events.events.k8s.io",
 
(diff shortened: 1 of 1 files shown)


blackpiglet:
Although `EndpointSlices` is mainly used as an auto-created managed by k8s control-plan, according to this link https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#management, it also has other use cases.
If a service mesh manages it, deleting and recreating may cause information loss, but that depends on how the service mesh is implemented.

Issue #6106, opened 2023-04-10
CSI backup/restore to FS backup/restore fails


eemcmullan:
**What steps did you take and what happened:**

- Completed backup/restore with CSI snapshotting enabled
- Complete new FS backup of previously restored volumes from CSI snapshot
- Restore from FS backup 
- Restore fails as PVR never completes, and PVC remains pending with event:

```
failed to provision volume with StorageClass "gp2-csi": error getting handle for DataSource Type VolumeSnapshot by Name velero-mssql-pvc-v787x: error getting snapshot velero-mssql-pvc-v787x from api server: volumesnapshots.snapshot.storage.k8s.io "velero-mssql-pvc-v787x" not found
```

PVC dataSourceRef is reset here: https://github.com/vmware-tanzu/velero-plugin-for-csi/blob/main/internal/restore/pvc_action.go#L71-L81 

Could implement a check for volumes previously restored from CSI, and reset this field again if needed.

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

sseago:
I'm wondering whether we should always strip dataSourceRef when saving PVCs. That would resolve this issue, but are there any use cases where we need that in the backup? For CSI use cases, that's set by the CSI restore plugin, so we probably don't need it there.

eemcmullan:
One way to resolve this issue is to clear this field on each backup, as a PVC using a volumeSnapshot as a dataSource is reset during CSI restore. We can utilize the current PVC BIA here to do so. 
PR: https://github.com/vmware-tanzu/velero/pull/6111

weshayutin:
- [ ] openshift/velero:konveyor dev
  - [ ] https://github.com/openshift/velero/pull/257
- [ ]  openshift/velero:oadp-1.1

PR #6111, opened 2023-04-11
Remove dataSource from PVC on backup for prior CSI restore case

eemcmullan:

Removes the dataSource field from PVCs on backup to resolve issue #6106

Fixes #(issue)
#6106 

### DIFF
--- a/pkg/backup/backup_pv_action.go
+++ b/pkg/backup/backup_pv_action.go
@@ -20,6 +20,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1api "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	v1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
@@ -61,5 +62,19 @@ func (a *PVCAction) Execute(item runtime.Unstructured, backup *v1.Backup) (runti
 		GroupResource: kuberesource.PersistentVolumes,
 		Name:          pvc.Spec.VolumeName,
 	}
-	return item, []velero.ResourceIdentifier{pv}, nil
+	// remove dataSource if exists from prior restored CSI volumes
+	if pvc.Spec.DataSource != nil {
+		pvc.Spec.DataSource = nil
+	}
+	if pvc.Spec.DataSourceRef != nil {
+		pvc.Spec.DataSourceRef = nil
+	}
+
+	pvcMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&pvc)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "unable to convert pvc to unstructured item")
+	}
+
+	return &unstructured.Unstructured{Object: pvcMap}, []velero.ResourceIdentifier{pv}, nil
+
 }
--- a/changelogs/unreleased/6111-eemcmullan
+++ b/changelogs/unreleased/6111-eemcmullan
@@ -1,0 +1,2 @@
+Remove any dataSource or dataSourceRef fields from PVCs in PVC BIA for cases of
+prior PVC restores with CSI


pkg/backup/backup_pv_action.go
```diff
@@ -61,5 +62,19 @@ func (a *PVCAction) Execute(item runtime.Unstructured, backup *v1.Backup) (runti
...
+	if pvc.Spec.DataSource != nil {
+		pvc.Spec.DataSource = nil
+	}
+	if pvc.Spec.DataSourceRef != nil {
+		pvc.Spec.DataSourceRef = nil
+	}
+
+	pvcMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&pvc)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "unable to convert pvcv to unstructured item")
```
eemcmullan, line 75: typo

kaovilai (approved):
Lgtm

blackpiglet:
Please also add a changelog file.

sseago (approved):

kaovilai (approved):

pkg/backup/backup_pv_action.go
```diff
@@ -61,5 +62,19 @@ func (a *PVCAction) Execute(item runtime.Unstructured, backup *v1.Backup) (runti
 		GroupResource: kuberesource.PersistentVolumes,
 		Name:          pvc.Spec.VolumeName,
 	}
-	return item, []velero.ResourceIdentifier{pv}, nil
+	// remove dataSource if exists from prior restored CSI volumes
+	if pvc.Spec.DataSource != nil {
```
Lyndon-Li, line 66: I suggest we move the same to restore, specifically:
During restore, we judge if the DataSource/DataSourceRef is required (for CSI restore, it is required, for others it is not), if it is required, leave the fields there, otherwise, remove it as what we are doing here

Reasons:
1. In backup, we may have multiple BIAs, a BIA doesn't know what others is doing or will do, so if not necessary, a BIA should not change the data comparing to the source, otherwise, we don't know what other BIAs will react
2. As the principle, if not necessary, the backup procedure should not change anything comparing to the source. This principle guarantees the consistency and coherence of the restore data and finally guarantees the validation of the restore
3. The current problem is right a requirement that we need to take care of the  DataSource/DataSourceRef, just like what we do for Status fields. Therefore, it looks more appropriate to put it into restore

On the other hand, I am not seeing any direct error the current change will cause, so I will leave this to more members for judgements, @reasonerjt @blackpiglet @shubham-pampattiwar

blackpiglet: Keeping the resources unchanged in backup is reasonable, but I also couldn't spot any scenarios broken by the current change.

sseago: @Lyndon-Li the problem with clearing it on restore is that restore is where some plugins might add it. For example, CSI restore plugin adds it, so if the built-in RIA clears it, we could have a race condition, and there's no way of knowing within one plugin action whether another action has been called and/or will be called. Are there any use cases where having this in backup is needed? This field is only used for initial content creation for a new PVC. This source provided the content for the PVC at creation time. It will not have any content that was added/modified after creation -- for that we need the *current* content, via kopia/restic, CSI snapshot, or native snapshot. So I'm not sure we ever need/want this in the backup.

Lyndon-Li: @sseago Yes, indeed, if we want to put the clear code in restore, we need to make sure that the fields are not added by a CSI RIA plugin.  One idea is we put the clear code at early stage of `restoreItem`, e.g., somewhere[ around here](https://github.com/vmware-tanzu/velero/blob/d298c6d0a2284d0e0444998efd3ad94472a312c2/pkg/restore/restore.go#L1184), none of the RIAs has been called at this place, so it is safe to remove the fields without a check.

> Are there any use cases where having this in backup is needed?

Personally, I don't see any need for them during in backup at present, so technically, I don't see any error will happen if we keep the current clear code in backup.
Merely, I just think it is more appropriate to have the code in restore (as the reasons mentioned above). Therefore, I raise this as a suggestion, I am fine with either way as the final decision.

Lyndon-Li: > as for most use cases, restore is more appropriate, but in cases where a field will never be useful for restore (or for RIA funcs), then removing on backup is fine

OK, this sounds reasonable

pkg/backup/backup_pv_action.go
```diff
@@ -61,5 +62,19 @@ func (a *PVCAction) Execute(item runtime.Unstructured, backup *v1.Backup) (runti
...
+		pvc.Spec.DataSource = nil
+	}
+	if pvc.Spec.DataSourceRef != nil {
+		pvc.Spec.DataSourceRef = nil
+	}
```
kaovilai, line 71: Velero add label to items it restored
https://github.com/vmware-tanzu/velero/blob/65f99c12640414525476e02bb9302bf7c544a49d/pkg/restore/restore.go#L1853

Do we wanna limit this action to items we are certain are restored by velero?
Or perhaps limit to just those restored by velero-csi-plugin?
```suggestion
	if pvc.Spec.DataSource != nil {
		pvc.Spec.DataSource = nil
	}
	if pvc.Spec.DataSourceRef != nil {
		pvc.Spec.DataSourceRef = nil
	}
```

kaovilai (changes requested):

pkg/backup/backup_pv_action.go
sseago: I don't think prior restore is necessarily relevant. While the specific failure case we've seen involves prior restore, I think these fields would almost always result in restoring a PVC with stale or invalid content.

kaovilai: Ok then retitle the PR to a broader scope, then suggest CSI restore as one of the failure case.

blackpiglet (approved):

PR #6155, opened 2023-04-16
uninstall velero without deleting namespace

### LINKED ISSUE #5433: High risk: uninstall velero will delete the namespace
**$ velero -n devops  uninstall velero** 
```
You are about to uninstall Velero.
Are you sure you want to continue (Y/N)? y
Waiting for velero namespace "devops" to be deleted
...........................................................................................................................................................................
```

kaovilai:

Fixes #5433 


pkg/cmd/cli/install/install.go
```diff
@@ -118,6 +120,8 @@ func (o *InstallOptions) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Features, "features", o.Features, "Comma separated list of Velero feature flags to be set on the Velero deployment and the node-agent daemonset, if node-agent is enabled")
 	flags.BoolVar(&o.DefaultVolumesToFsBackup, "default-volumes-to-fs-backup", o.DefaultVolumesToFsBackup, "Bool flag to configure Velero server to use pod volume file system backup by default for all volumes on all backups. Optional.")
 	flags.StringVar(&o.UploaderType, "uploader-type", o.UploaderType, fmt.Sprintf("The type of uploader to transfer the data of pod volumes, the supported values are '%s', '%s'", uploader.ResticType, uploader.KopiaType))
+	flags.BoolVar(&o.Uninstall, "uninstall", o.Uninstall, "Uninstall Velero from the cluster. Optional.")
```
Lyndon-Li, line 123: Why do we add a `--uninstall` flag under `velero install` instead of under `velero uninstall`?
It is hard to understand if the command is like "velero install --uninstall".

Lyndon-Li:
@kaovilai
Here we have requirements:

1. Preserve the namespace, use case is as issue #5433 
2. Preserve Velero internal objects, i.e., BSL, VSL, schedules, backup, restore, policies, etc.

The implementation for 1 is clear.
However, for 2, some more details need to be covered:

- If an object is preserved, after reinstall, what if the CRD changes, as a result, the preserved object doesn't match the new CRD?
- Not all internal objects need to be preserved, it may even be harmful if we preserve some unnecessary objects. So we need to clarify which ones will be preserved, which ones will not and why

Comparing to 1, 2 is more useful because along with the evolution of Velero, objects may be quite complex and their lifecycle may be better to match the cluster instead of Velero installation, i.e., the schedules and polices.

Therefore, would you like to create a design doc to cover both 1 and 2 and make an overall advanced uninstallation feature?

cc @reasonerjt @sseago

kaovilai:
Sure thing. Design doc for 1 and 2.

For 2. I think we only have to deal with it at reinstall time so we can check for outdated velero resources if namespace exist already.

Tho I'm not sure if 2 is possible because 1 deletes CRD right now. Maybe I need to add 1 more flag, `--preserve-CRD`

Lyndon-Li:
For requirement 1, personally I would rather regard it as a corner case -- someone has installed Velero into an existing namespace, though the consequence of uninstallation in this case is severe.
Then I would think that it is unnecessary to expose a flag like `--preserve-uninstall-namespace`. Instead, we just check the current namespace is `velero` or not, if it is `velero`, we delete the namespace, otherwise, we don't delete it.
If the namespace is indeed `velero`, I cannot find any case that we need to preserve it.

Once we have the design doc, we can discuss this further.

Lyndon-Li:
> For 2. I think we only have to deal with it at reinstall time so we can check for outdated velero resources if namespace exist already.

Things may be more complicated:
1. How do we tell the resource is outdated?
2. Once we find it is outdated during reinstall, what do we do? We cannot delete it since it takes lots of user defined data

> Tho I'm not sure if 2 is possible because 1 deletes CRD right now. Maybe I need to add 1 more flag, --preserve-CRD

Yes, for some CRs, if we want to preserve it, the CRD will also be required. However, from users perspective, they don't case about CRDs, they only cares whether their defined data along with the CRs have been preserved or not. For sure, we need to design this from users perspective.

pkg/cmd/cli/install/install.go
kaovilai: Because uninstall command do not have the install options.

If we want this functionality in uninstall cmd then the installOptions fields will have to be included with uninstall cmd.

This makes it clear that it'll only work if you know how you installed it.

Perhaps even lookup your bash history and just add --uninstall

Lyndon-Li: > Because uninstall command do not have the install options

I thinks this is a problem of the code organization, we have multiple ways to solve it. i.e., Add a new UninstallOption struct.

> This makes it clear that it'll only work if you know how you installed it.

Maybe, but this will confuse others who don't know the details.

Anyway, as mentioned about, I think even the flag `--preserve-uninstall-namespace` is unnecessary if we want to fix requirement 1 only.

Issue #6112, opened 2023-04-12
[Epic] Volume Snapshot Data Movement


Lyndon-Li:
Implement Volume Snapshot Data Movement based on design #5968, which includes:

- Volume snapshot data movement workflow for backup
- Volume snapshot data movement workflow for restore
- Data mover plugin for CSI snapshot
- Velero built-in data mover
- ...

PR #6176, opened 2023-04-24
Add data mover CRD under v2alpha1

Lyndon-Li:
Add data mover CRD under v1alpha1, include:

- New DataUpload CRD under v1alpha1
- New DataDownload CRD under v1alpha1
- Minor changes to Backup CRD
- Minor changes to Restore CRD
- Minor changes to PodVolumeBackup CRD
- Minor changes to PodVolumeRestore CRD

Fix issue #6112


Lyndon-Li:
Follow up Qeustion: Do it make sense to have two different versions in one single group?

kaovilai:
Reading this https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects

I think its ok to mix GroupVersions because its a different Kind. As long as GroupVersionKind do not overlap, it can share GroupVersion.

When it becomes v1, it will be a different GroupVersionKind that can be migrated to using the [storage Version migrator](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version)

It is possible for the same controller to process both v1beta1 and v1.. I've seen that in https://github.com/kubernetes-csi/external-snapshotter

hack/verify-generated-crd-code.sh
```diff
@@ -19,7 +19,7 @@ HACK_DIR=$(dirname "${BASH_SOURCE}")
 ${HACK_DIR}/update-3generated-crd-code.sh
 
 # ensure no changes to generated CRDs
-if ! git diff --exit-code config/crd/v1/crds/crds.go >/dev/null; then
+if [! git diff --exit-code config/crd/v1/crds/crds.go >/dev/null] || [! git diff --exit-code config/crd/v1alpha1/crds/crds.go >/dev/null]; then
```
kaovilai, line 22: This can be simplified.
```suggestion
if [! git diff --exit-code config/crd/v1/crds/crds.go config/crd/v1alpha1/crds/crds.go >/dev/null]; then
```

pkg/apis/velero/v1alpha1/data_upload_types.go
```diff
@@ -0,0 +1,209 @@
...
+	// SnapshotType is the type of the snapshot to be backed up.
+	SnapshotType SnapshotType `json:"snapshotType"`
+
+	// If SnapshotType is CSI, CSISnapshot provides the information of the CSI snapshot.
+	// +optional
+	// +nullable
+	CSISnapshot *CSISnapshotSpec `json:"csiSnapshot"`
+
+	// SourcePVC is the name of the PVC which the snapshot is taken for.
+	SourcePVC string `json:"sourcePVC"`
```
kaovilai, line 36: Is it possible that a user would want to backup a CSI Snapshot no longer related to a PVC in the cluster?
They may want to move snapshot to object storage and restore to a destination snapshot without a destination PVC.

Lyndon-Li: @kaovilai 
At least for the current CSI snapshot data movement workflow, it is not possible, because the DataUpload CR is created by the CSI plugin, and CSI plugin must take a PVC in order to create a CSI snapshot.
If this CRD is used in other scenarios, i.e., searching an existing VS and run a data movement, it is possible. For this case, `sourcePVC` could be set to "" as long as the underlying data mover doesn't use this value.

By the way, this discussion is more suitable to be in the design PR https://github.com/vmware-tanzu/velero/pull/5968, the design has include the details of the two CRDs

hack/verify-generated-crd-code.sh
Lyndon-Li: Thanks, let me change this later.

pkg/apis/velero/v1alpha1/data_upload_types.go
kaovilai: Thanks! Still learning my way around :)

reasonerjt:
> Follow up Question 1: Do it make sense to have two different versions in one single group?
> Follow up Question 2: Is it better logically to put the DM CRDs into a separate group other than velero.io since they will be used by plugin DMs as well?

I had a chance to chat with upstream maintainer of k8s, and the suggestion is that they should be 
```
velero.io/v2alpha1/DataUpload
velero.io/v2alpha1/DataDownload
```

Lyndon-Li:
> > Follow up Question 1: Do it make sense to have two different versions in one single group?
> > Follow up Question 2: Is it better logically to put the DM CRDs into a separate group other than velero.io since they will be used by plugin DMs as well?
> 
> I had a chance to chat with upstream maintainer of k8s, and the suggestion is that they should be
> 
> ```
> velero.io/v2alpha1/DataUpload
> velero.io/v2alpha1/DataDownload
> ```

@reasonerjt Thanks! Let me make the change accordingly.

Lyndon-Li:
Done changing to v2alpha2

hack/verify-generated-crd-code.sh
Lyndon-Li: Done

Makefile
```diff
@@ -229,6 +229,9 @@ endif
 update:
 	@$(MAKE) shell CMD="-c 'hack/update-all.sh'"
 
+update-crd:
```
blackpiglet, line 232: CRDs are also updated in the `update` target. Not sure why adding a new target to just update CRD.

Lyndon-Li: This is for developing usage only, sometimes when we change some of the CRDs, we need to generate the new yaml and CRD.go to call `make-update`, but the entire `make-update` is very slow, so this `make update-crd `may be a shortcut in this scenario.

blackpiglet (dismissed):

Makefile
reasonerjt: How about adding a comment to explain this is for development only?

Lyndon-Li: Done, added the comment.

kaovilai: Make update is now much faster :)

blackpiglet (approved):

blackpiglet (dismissed):

Makefile
Lyndon-Li: @kaovilai 
Yes, I noticed this recently, thanks for making this improvement.
For `make update-crd`, let's keep it since it is not harmful and it does things less and help developers to save time.

blackpiglet (approved):

sseago (approved):

Issue #6182, opened 2023-04-25
v1.10 and later - Pod volume backup keeps partial failing for namespaces with pods in completed state


Lyndon-Li:
From v1.10, pod volume backup will fail earlier if node-agent pod is not running in the same node where the backing up pod is running, instead of waiting until timeout (by default, 4 hours). Therefore, it does more checks about the status of backing up pod and node-agent pod. However, the checks involves an unnecessary behavior change -- if the backing up pod is not running, that pod volume backup will be marked as fail and the entire backup goes to partially failure.

This will cause the pod volume backup to namespaces with pods in completed status to keep failing.

Lyndon-Li:
Close as completed

karan1ngit:
when will 1.11.1 be released?

PR #6189, opened 2023-04-25
[1.11] Fix issue 6182

Lyndon-Li:
Fix issue #6182. If pod is not running, don't treat it as an error, let it go and leave a warning.


blackpiglet (approved):

qiuming-best (approved):

SSvilen:
@Lyndon-Li  it seems that this did not make it in 1.11.0. In which patch version will this change be present?

Lyndon-Li:
@SSvilen It is not in 1.11.0, if you want to use 1.11, you need to wait for 1.11.1; otherwise, you can use 1.10.3 where the issue fix is included.

PR #6243, opened 2023-05-08
Gather list of volumes to pvb backup after execute actions

kaovilai:
Gather list of volumes to pvb backup after execute actions so plugin can affect which volumes are in the backup

Fixes #(issue)

### DIFF
--- a/pkg/backup/item_backupper.go
+++ b/pkg/backup/item_backupper.go
@@ -173,42 +173,11 @@ func (ib *itemBackupper) backupItemInternal(logger logrus.FieldLogger, obj runti
 
 	var (
 		backupErrs []error
-		pod        *corev1api.Pod
-		pvbVolumes []string
 	)
 
 	log.Debug("Executing pre hooks")
 	if err := ib.itemHookHandler.HandleHooks(log, groupResource, obj, ib.backupRequest.ResourceHooks, hook.PhasePre); err != nil {
 		return false, itemFiles, err
-	}
-
-	if groupResource == kuberesource.Pods {
-		// pod needs to be initialized for the unstructured converter
-		pod = new(corev1api.Pod)
-		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), pod); err != nil {
-			backupErrs = append(backupErrs, errors.WithStack(err))
-			// nil it on error since it's not valid
-			pod = nil
-		} else {
-			// Get the list of volumes to back up using pod volume backup from the pod's annotations. Remove from this list
-			// any volumes that use a PVC that we've already backed up (this would be in a read-write-many scenario,
-			// where it's been backed up from another pod), since we don't need >1 backup per PVC.
-			for _, volume := range podvolume.GetVolumesByPod(pod, boolptr.IsSetToTrue(ib.backupRequest.Spec.DefaultVolumesToFsBackup)) {
-				// track the volumes that are PVCs using the PVC snapshot tracker, so that when we backup PVCs/PVs
-				// via an item action in the next step, we don't snapshot PVs that will have their data backed up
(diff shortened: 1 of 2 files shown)


pkg/backup/item_backupper.go
```diff
@@ -242,17 +211,46 @@ func (ib *itemBackupper) backupItemInternal(logger logrus.FieldLogger, obj runti
...
+			backupErrs = append(backupErrs, errors.WithStack(err))
+			// nil it on error since it's not valid
+			pod = nil
+		} else {
+			// Get the list of volumes to back up using pod volume backup from the pod's annotations. Remove from this list
+			// any volumes that use a PVC that we've already backed up (this would be in a read-write-many scenario,
+			// where it's been backed up from another pod), since we don't need >1 backup per PVC.
+			for _, volume := range podvolume.GetVolumesByPod(pod, boolptr.IsSetToTrue(ib.backupRequest.Spec.DefaultVolumesToFsBackup)) {
+				// track the volumes that are PVCs using the PVC snapshot tracker, so that when we backup PVCs/PVs
+				// via an item action in the next step, we don't snapshot PVs that will have their data backed up
```
kaovilai, line 228: @Lyndon-Li is there an item action that required the `ib.podVolumeSnapshotTracker.Track(pod, volume)` to be before `executeActions()`?

This PR moves this task after executeActions so plugins can add volumeToExclude annotation.

Lyndon-Li: @kaovilai 
1. This `podVolumeSnapshotTracker` is primarily used to track the volumes backed up by PVB, in case that the following PVB backs up the volume again (i.e., in the read-write-many scenario). The volumes are gradually added, it doesn't take the full list of the volumes at the beginning, so it may not be the right thing to track the exclusive of volumes.
2. The other place that is using this `podVolumeSnapshotTracker` is `takePVSnapshot` function which is calling volume snapshotters. It checks this tracker to exclude the volumes that are backed up by PVB from volume snapshotter. If we make the updating of this `podVolumeSnapshotTracker` after BIAs, the functionalities  of `takePVSnapshot` will be broken, because the PVs to be backed up is added by PodAction->PVCAction BIAs, which happens during BIAs call of the pod.
3. Is there any requirement for dynamically excluding volume? I think the current workflow is not ready for doing this -- Velero doesn't guarantee the order of BIAs, if we allow BIAs to decide which volume to be excluded, it is possible that for this time of backup/this installation of backup, the volumes are excluded but for other time/installation, they are not.

Therefore, I suggest we first check the requirement for dynamically excluding volumes by plugins. If there are requirements, we then pursue a sophisticated solution. cc @reasonerjt @sseago

kaovilai: We're doing this currently in https://github.com/openshift/openshift-velero-plugin/pull/194

reasonerjt (changes requested):
I'm afraid we are not ready to merge this to upstream per the valid concerns in the comment:
https://github.com/vmware-tanzu/velero/pull/6243/files#r1189320327


IMO we need a better design to address the requirement.

reasonerjt (changes requested):
I'm afraid we are not ready to merge this to upstream per the valid concerns in the comment:
https://github.com/vmware-tanzu/velero/pull/6243/files#r1189320327


IMO we need a better design to address the requirement.

reasonerjt (changes requested):
I'm afraid we are not ready to merge this to upstream per the valid concerns in the comment:
https://github.com/vmware-tanzu/velero/pull/6243/files#r1189320327


IMO we need a better design to address the requirement.

kaovilai:
Actually what we were trying to accomplish here was to fix the same issue as https://github.com/vmware-tanzu/velero/pull/6189

Since that got merged, this is not necessary any longer.

Issue #5999, opened 2023-03-20
Bump up Kopia to v0.13


Lyndon-Li:
Currently Velero is using 0.10.7. Therefore some Repository interface changes, so require some logic changes in Velero in order to bump up to a higher version of Kopia

reasonerjt:
Moving it out of v1.11 milestone, as it may be less severe than we thought.

bitfactory-sem-denbroeder:
Any ETA on this issue? this is blocking for us because of the known assumerole issue with kopia 0.10.7.

Lyndon-Li:
Fixed by #6248

bjb2010:
> 
There is oom issue in kopia v0.10.7
https://github.com/kopia/kopia/issues/2888

XhmikosR:
@Lyndon-Li We really need this patch :/ Is there any way the patch could be backported to 1.11 or 1.10?

PR #6248, opened 2023-05-10
Bump up kopia v0.13

Lyndon-Li:
Bump up kopia to v0.13

Fix issue #5999


Lyndon-Li:
We face [this problem](https://github.com/kubernetes-sigs/controller-tools/issues/613) of controller-gen after bumping up and changing go.mod, as a result, the `make update` always crashes. So we need to bump up controller-gen to v0.12.0.

pkg/repository/udmrepo/repo.go
```diff
@@ -68,6 +68,7 @@ type ObjectWriteOptions struct {
 	Prefix      ID     // A prefix of the name used to save the object
 	AccessMode  int    // OBJECT_DATA_ACCESS_*
 	BackupMode  int    // OBJECT_DATA_BACKUP_*
+	AsyncWrites int    // Num of async writes for the object, 0 means no async write
```
blackpiglet, line 71: Looks like there is no place to assign value to the `AsyncWrites` parameter.

pkg/uploader/kopia/shim.go
```diff
@@ -208,11 +208,22 @@ func (sr *shimRepository) DeleteManifest(ctx context.Context, id manifest.ID) er
...
 // Flush all the unifited repository data
 func (sr *shimRepository) Flush(ctx context.Context) error {
 	return sr.udmRepo.Flush(ctx)
 }
 
+func (sr *shimRepository) ConcatenateObjects(ctx context.Context, objectIDs []object.ID) (object.ID, error) {
+	return object.ID{}, errors.New("ConcatenateObjects is not supported")
+}
+
+func (sr *shimRepository) OnSuccessfulFlush(callback repo.RepositoryWriterCallback) {
```
blackpiglet, line 224: Just curious about why not `OnSuccessfulFlush` function return value handling is different from other newly added functions.

blackpiglet:
Could you give more information on why the uber zap log library is used instead in `pkg/util/logging/kopia_log.go`

pkg/repository/udmrepo/repo.go
Lyndon-Li: Fixed in `(sr *shimRepository) NewObjectWriter`, it is an optional value, so right now, we don't need to set a value, just convey the value from shim to unified repo.

Lyndon-Li:
> Could you give more information on why the uber zap log library is used instead in `pkg/util/logging/kopia_log.go`?

`pkg/util/logging/kopia_log.go` is still used. zap log is the logger used by Kopia all the time. 
In the previous release, Kopia has a wrapper of the zap log, so Velero doesn't need to hook zap log in order to redirect Kopia log to Velero.
In the current release, zap log is used directly by Kopia, so in Velero code we have to hook zap log for the same purpose.

pkg/uploader/kopia/shim.go
Lyndon-Li: The prototype of these functions are defined by Kopia, specifically, they implements Kopia' Repository interface. In the interface, `OnSuccessfulFlush` has no return value.
On the other hand, this function is a callback under the Flush operation, while this callback is optional, if it is not to be handled, just leave it as empty.

blackpiglet (approved):

qiuming-best (approved):

Issue #5875, opened 2023-02-17
Kopia uploader do not work with service account that assume an IAM role


stephanDesj:
### Discussed in https://github.com/vmware-tanzu/velero/discussions/5874

<div type='discussions-op-text'>

<sup>Originally posted by **stephanDesj** February 17, 2023</sup>
Hi, him evaluating Velero 1.10 with kopia uploader in an EKS Cluster.
The velero server and the node-agent pod is configured with a service account that mount the AWS credentials on the pod. This was working fine with restic, but when I test with Kopia, I have this error:

time="..." level=error msg="Error backing up item" backup=... error="backup repository is not ready: error to get repo options: error to get repo credentials: **error get s3 credentials: missing credential file**" error.file="/go/src/github.com/vmware-tanzu/velero/pkg/repository/backup_repo_op.go:83" error.function=github.com/vmware-tanzu/velero/pkg/repository.GetBackupRepository logSource="pkg/backup/backup.go:425" name=main-velero-76f5b7d6cf-5n2lw

Looks like with Kopia integration is not able to pickup the mounted credential and ask for a credential file instead.
</div>

Error is in the Kopia provider when executing GetStoreOptions,
it's calling
funcTable.getStorageCredentials that call getS3Credentials that is implemented in /pkg/repository/config/aws.go line 57.
and if you look the code, it's forcing the use of a shared credential file.

Possible solution will be to change that part and check that if credentialsFile == "", try to create a session without it and only if the session is not created, throws the errors.New("missing credential file").

Lyndon-Li:
Velero 1.10 is integrating with Kopia 0.10.7 which support static file credentials only.
Checked Kopia release notes, IAM credentials are not supported until Kopia 0.12.0, as you can see [here](https://github.com/kopia/kopia/releases#:~:text=New%20Feature%20add%20AWS%20IAM%20as%20a%20credential%20provider%20for%20s3%20storage%20(%232213)%20(%232222)%20by%20Jarek%20Kowalski).

It means even though we let Velero go with an empty credentialsFile in the function `GetS3Credentials`, Kopia will still fail since it doesn't set credentials.IAM in the `newStorage` function.

reasonerjt:
This will be fixed by bumping up kopia to 0.13

rshiva777:
@reasonerjt / @Lyndon-Li ,

i am facing the same credential issue.

[...]

PR #6283, opened 2023-05-18
Fix issue 5875

Lyndon-Li:
Fix issue #5875. Since Kopia has supported IAM, Velero should not require static credentials all the time


pkg/repository/provider/unified_repo.go
```diff
@@ -419,14 +419,16 @@ func getStorageCredentials(backupLocation *velerov1api.BackupStorageLocation, cr
...
 	case repoconfig.AWSBackend:
 		credValue, err := getS3Credentials(config)
-		if err != nil {
+		if err == nil {
+			result[udmrepo.StoreOptionS3KeyID] = credValue.AccessKeyID
+			result[udmrepo.StoreOptionS3Provider] = credValue.ProviderName
+			result[udmrepo.StoreOptionS3SecretKey] = credValue.SecretAccessKey
+			result[udmrepo.StoreOptionS3Token] = credValue.SessionToken
+		} else if err == repoconfig.AWSCredFileMissingError {
+			log.Warn("AWS credential file is not present, so no credentials are provided")
```
blackpiglet, line 428: If this happens, then there is no credential information in the result, but the IAM way also mounts the credential information into the Velero pod.
Is there a place where the Kopia Library can read credential information?

reasonerjt: +1
Also curious how kopia handles different auth mechanisms.

Lyndon-Li: What Velero needs is letting it go without checking static ak/sk as well.
For sure, the IAM role SA should be mounted to Velero pods, this has been done by another PR https://github.com/vmware-tanzu/velero/pull/5802

Lyndon-Li: Moreover, it looks weird to let it go without checking anything when getting credentials, so here I have modified the code to check the existence of the env variable AWS_ROLE_ARN, only when its value is not empty, we let it go with empty ak/sk

blackpiglet (approved):

reasonerjt (approved):

Issue #1970, opened 2019-10-16
Velero restore with include-resources to restore only namespace resource is not working


Frank51:
**What steps did you take and what happened:**
Velero user could use '--include-resources' flag to specify which kind of resources need to be restored from backup. But when  '--include-resources' only contains 'namespaces', the backup namespace could not be restored successfully to the Kubernetes cluster. If adding 'namespaces' and 'services' into the  'include resources', the backup could be restored correctly at this time. 

```
velero get backups
NAME        STATUS      CREATED                         EXPIRES   STORAGE LOCATION   SELECTOR
yelb-back   Completed   2019-09-20 12:39:34 -0700 PDT   3d        default            <none>

velero describe backup yelb-back
Name:         yelb-back
Namespace:    velero
Labels:       velero.io/storage-location=default
Annotations:  <none>

Phase:  Completed

Namespaces:
  Included:  yelb-ns
  Excluded:  <none>

Resources:
  Included:        *
  Excluded:        <none>
  Cluster-scoped:  auto

Label selector:  <none>

Storage Location:  default

Snapshot PVs:  auto

TTL:  720h0m0s

Hooks:  <none>

Backup Format Version:  1

Started:    2019-09-20 12:39:34 -0700 PDT
Completed:  2019-09-20 12:39:35 -0700 PDT

Expiration:  2019-10-20 12:39:34 -0700 PDT

Persistent Volumes: <none included>


kubectl get ns
NAME              STATUS   AGE
default           Active   5h6m
kube-node-lease   Active   5h6m
kube-public       Active   5h6m
kube-system       Active   5h6m
velero            Active   152m

velero restore create --from-backup yelb-back --include-resources namespaces
Restore request "yelb-back-20191016140424" submitted successfully.
Run `velero restore describe yelb-back-20191016140424` or `velero restore logs yelb-back-20191016140424` for more details.

kubectl get ns
NAME              STATUS   AGE
default           Active   5h7m
kube-node-lease   Active   5h7m
kube-public       Active   5h7m
kube-system       Active   5h7m
velero            Active   152m

# yelb-ns is not restored from the backup.
# However, when I restore both the namespace and service, yelb-ns is restored:

velero restore create --from-backup yelb-back --include-resources namespaces,services
Restore request "yelb-back-20191016140932" submitted successfully.
Run `velero restore describe yelb-back-20191016140932` or `velero restore logs yelb-back-20191016140932` for more details.

kubectl get ns
[...]

PR #6320, opened 2023-05-29
Include namespace resource needed by namespaced-scope resource in backup

### LINKED ISSUE #6302: namespace object should always been backed up if any objects in it are backed up
1. Backup some resources in a namespace which match a label
2. The namespace resource is not backed up because it does not have such labels

namespace is backed up

_If you are using velero v1.7.0+:_  
Please use `velero debug  --backup <backupname> --restore <restorename>` to generate the support bundle, and attach to this issue, more options please refer to `velero debug --help` 

_If you are using earlier versions:_  
Please provide the output of the following commands (Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)

**Environment:**

**Vote on this issue!**

blackpiglet:

For some use cases, namespaced-scope resources are inluded into backup, but the namespaces are not included due to filters setting. To do this, removing label selector filter from namespace resource. Namespace resource only honor namespace exclude/include filters.

Fixes #6302 

### DIFF
--- a/pkg/backup/item_collector.go
+++ b/pkg/backup/item_collector.go
@@ -29,7 +29,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/tools/pager"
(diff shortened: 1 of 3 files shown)


pkg/backup/backup.go
```diff
@@ -407,6 +407,12 @@ func (kb *kubernetesBackupper) BackupWithResolvers(log logrus.FieldLogger,
 	// no more progress updates will be sent on the 'update' channel
 	quit <- struct{}{}
 
+	// For some use cases, e.g. use label to filter resources in backup,
+	// namespace-scoped resources are included, but the namespace contains
+	// the resources are not. Add the function to include the namespaces
+	// if needed.
+	kb.backupAdditionalNamespaces(items, itemBackupper, log)
```
ywk253100, line 414: Does the logic of this function conflict with the one in https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_collector.go#L282? 
How about removing the label-matching logic for namespaces directly?

blackpiglet: In most cases, removing this code should have no change in behavior.
I only think one scenario needs consideration.
If the user labels the namespace resource and no resources in that namespace are selected in the backup, then the namespace will not be included in the backup.

pkg/backup/backup.go
```diff
@@ -407,6 +407,12 @@ func (kb *kubernetesBackupper) BackupWithResolvers(log logrus.FieldLogger,
 	// no more progress updates will be sent on the 'update' channel
 	quit <- struct{}{}
 
+	// For some use cases, e.g. use label to filter resources in backup,
+	// namespace-scoped resources are included, but the namespace contains
```
shubham-pampattiwar, line 411: NIT: **that** contains

blackpiglet: Modified.

pkg/backup/backup.go
blackpiglet: After reading the https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_collector.go cluster scope handling code, I agree with Wenkai.
Namespace resources are already handled correctly in the cluster-scoped resource code.

pkg/backup/item_collector.go
```diff
@@ -275,58 +274,6 @@ func (r *itemCollector) getResourceItems(log logrus.FieldLogger, gv schema.Group
 
 	namespacesToList := getNamespacesToList(r.backupRequest.NamespaceIncludesExcludes)
 
-	// Check if we're backing up namespaces for a less-than-full backup.
```
blackpiglet, line 278: I agree with @ywk253100 to remove this part of code, because the most of namespace handling logic is duplicate with the following other resources handling logic.
And the current namespace handling logic cannot work with OrLabelSelector.

sseago: @blackpiglet I think this should be fine. So if I'm understanding the code correctly, this means if a backup includes "ns1" and "ns2" and adds a label selector that only includes items in "ns1", then we back up both namespaces anyway. With current velero behavior, this won't matter for restore since empty namespaces aren't restored (as described in https://github.com/vmware-tanzu/velero/issues/1970 ) -- if the resolution of that issue is to restore empty namespaces, then the effect here would be that "ns1" would be restored with labeled resources inside, an "ns2" would be restored empty.

sseago (approved):

sseago (approved):

shubham-pampattiwar (approved):

ywk253100 (approved):

Issue #6297, opened 2023-05-22
velero kopia backup failure


amareshgreat:
**What steps did you take and what happened:**

**What did you expect to happen:**
the velero backup should complete sucessfully and it should take all the PVC backup using kopia.

**The following information will help us better understand what's going on**:

_If you are using velero v1.7.0+:_  
Please use `velero debug  --backup <backupname> --restore <restorename>` to generate the support bundle, and attach to this issue, more options please refer to `velero debug --help` 
[bundle-2023-05-22-09-59-33.tar.gz](https://github.com/vmware-tanzu/velero/files/11530281/bundle-2023-05-22-09-59-33.tar.gz)

**Anything else you would like to add:**

**Environment:**

- Velero version (use `velero version`): Client: v1.10.2 , Server v1.11.0
- Velero features (use `velero client config get features`): features: <NOT SET>
- Kubernetes version (use `kubectl version`): 1.22
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`): AWS

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

amareshgreat:
@Lyndon-Li 
the issue is we are seeing intermittent in nature, every time we run the backup, it fails with different PVC's, I have attached another bundle where the failure is for other PVC. so we are unable to determine why the backups are failing intermittently with different PVC's each time. But in the log, we are seeing same error "unable to determine bucket's region". 
[bundle-2023-05-22-10-26-45.tar.gz](https://github.com/vmware-tanzu/velero/files/11542439/bundle-2023-05-22-10-26-45.tar.gz)
![image](https://github.com/vmware-tanzu/velero/assets/4372356/2771e067-484e-49c1-9cca-722e414493b1)

we have also created a schedule to backup the namespace every one hour and we are seeing few are completed without error, while some have errors on few PVC's. if its failing to get bucket region, it should fail for all the PVC's.
![image](https://github.com/vmware-tanzu/velero/assets/4372356/5e7ba188-7df7-46c7-bbc2-5064baeee08d)

[...]

PR #6326, opened 2023-05-31
Enhance code for issue 6297

Lyndon-Li:
Enhance the code because of #6297, the return value of `GetBucketRegion` is not recorded, as a result, when it fails, we have no way to get the cause.


reasonerjt:
Please update the title of the PR and commit since this will not fix the particular issue, but only helps debugging the problem.

danfengliu (dismissed):

danfengliu (approved):

reasonerjt (approved):

PR #6337, opened 2023-06-02
Data upload controller

qiuming-best:

Fixes #(issue)

### DIFF
--- a/pkg/controller/data_upload_controller.go
+++ b/pkg/controller/data_upload_controller.go
@@ -1,0 +1,528 @@
+/*
+Copyright The Velero Contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	clocks "k8s.io/utils/clock"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	snapshotter "github.com/kubernetes-csi/external-snapshotter/client/v4/clientset/versioned/typed/volumesnapshot/v1"
+
+	"github.com/vmware-tanzu/velero/internal/credentials"
+	"github.com/vmware-tanzu/velero/pkg/apis/velero/shared"
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	velerov2alpha1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v2alpha1"
+	"github.com/vmware-tanzu/velero/pkg/datapath"
+	"github.com/vmware-tanzu/velero/pkg/exposer"
+	"github.com/vmware-tanzu/velero/pkg/repository"
+	"github.com/vmware-tanzu/velero/pkg/uploader"
+	"github.com/vmware-tanzu/velero/pkg/util/filesystem"
+	"github.com/vmware-tanzu/velero/pkg/util/kube"
+)
+
+const dataMoverType string = "velero"
+const dataUploadDownloadRequestor string = "snapshot-data-upload-download"
+
+// DataUploadReconciler reconciles a DataUpload object
+type DataUploadReconciler struct {
+	client              client.Client
+	kubeClient          kubernetes.Interface
+	csiSnapshotClient   snapshotter.SnapshotV1Interface
+	repoEnsurer         *repository.Ensurer
+	clock               clocks.WithTickerAndDelayedExecution
+	credentialGetter    *credentials.CredentialGetter
+	nodeName            string
+	fileSystem          filesystem.Interface
+	logger              logrus.FieldLogger
+	snapshotExposerList map[velerov2alpha1api.SnapshotType]exposer.SnapshotExposer
+	dataPathMgr         *datapath.Manager
+}
+
+func NewDataUploadReconciler(client client.Client, kubeClient kubernetes.Interface,
+	csiSnapshotClient snapshotter.SnapshotV1Interface, repoEnsurer *repository.Ensurer, clock clocks.WithTickerAndDelayedExecution,
+	cred *credentials.CredentialGetter, nodeName string, fs filesystem.Interface, log logrus.FieldLogger) *DataUploadReconciler {
+	return &DataUploadReconciler{
+		client:              client,
+		kubeClient:          kubeClient,
+		csiSnapshotClient:   csiSnapshotClient,
+		clock:               clock,
+		credentialGetter:    cred,
+		nodeName:            nodeName,
+		fileSystem:          fs,
+		logger:              log,
+		repoEnsurer:         repoEnsurer,
+		snapshotExposerList: map[velerov2alpha1api.SnapshotType]exposer.SnapshotExposer{velerov2alpha1api.SnapshotTypeCSI: exposer.NewCSISnapshotExposer(kubeClient, csiSnapshotClient, log)},
+		dataPathMgr:         datapath.NewManager(1),
+	}
+}
+
+// +kubebuilder:rbac:groups=velero.io,resources=datauploads,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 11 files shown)


pkg/controller/data_upload_controller.go
```diff
@@ -0,0 +1,498 @@
...
+		snapshotExposerList: map[velerov2alpha1api.SnapshotType]exposer.SnapshotExposer{velerov2alpha1api.SnapshotTypeCSI: exposer.NewCSISnapshotExposer(kubeClient, csiSnapshotClient, log)},
+		dataPathMgr:         datapath.NewManager(1),
+	}
+}
+
+// +kubebuilder:rbac:groups=velero.io,resources=dataupload,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=velero.io,resources=dataupload/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups="",resources=pods,verbs=get
+// +kubebuilder:rbac:groups="",resources=persistentvolumes,verbs=get
+// +kubebuilder:rbac:groups="",resources=persistentvolumerclaims,verbs=get
```
blackpiglet, line 94: Need a blank line between the marker comment and the Reconcile function.

pkg/controller/data_upload_controller.go
```diff
@@ -0,0 +1,498 @@
...
+
+func (r *DataUploadReconciler) acceptDataUpload(ctx context.Context, du *velerov2alpha1api.DataUpload) (bool, error) {
+	updated := du.DeepCopy()
+	updated.Status.Phase = velerov2alpha1api.DataUploadPhaseAccepted
+
+	r.logger.Infof("Accepting snapshot backup %s", du.Name)
+
+	time.Sleep(2 * time.Second)
+
+	err := r.client.Update(ctx, updated)
```
blackpiglet, line 447: It seems this line uses `Update` method on-purpose to make sure the request failed on conflict.
If that's correct, please add some comments to explain the purpose.

pkg/controller/data_upload_controller.go
```diff
@@ -0,0 +1,498 @@
...
+func (r *DataUploadReconciler) closeDataPath(ctx context.Context, duName string) {
+	fsBackup := r.dataPathMgr.GetAsyncBR(duName)
+	if fsBackup != nil {
+		fsBackup.Close(ctx)
+	}
+
+	r.dataPathMgr.RemoveAsyncBR(duName)
+}
+
+func (r *DataUploadReconciler) setupExposeParam(du *velerov2alpha1api.DataUpload) interface{} {
```
blackpiglet, line 468: No need to put this function as the method of struct `DataUploadReconciler `.

pkg/controller/data_upload_controller.go
```diff
@@ -0,0 +1,498 @@
...
+
+		log.Info("Data upload is accepted")
+
+		exposeParam := r.setupExposeParam(&du)
+		if err := ep.Expose(ctx, getOwnerObject(&du), du.Spec.CSISnapshot.VolumeSnapshot, du.Spec.OperationTimeout.Duration, exposeParam); err != nil {
+			return r.errorOut(ctx, &du, err, "error to expose snapshot", log)
+		}
+		log.Info("Snapshot is exposed")
+
+		return ctrl.Result{}, nil
```
blackpiglet, line 141: To me, `ep.Expose()` is a sync operation. After it returns, everything is already done, so it seems no need to return here. Just moving forward should be OK.

pkg/controller/data_upload_controller.go
```diff
@@ -0,0 +1,498 @@
...
+	if err := r.client.Patch(ctx, &du, client.MergeFrom(original)); err != nil {
+		log.WithError(err).Error("Failed to update progress")
+	}
+}
+
+// SetupWithManager registers the DataUpload controller.
+func (r *DataUploadReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&velerov2alpha1api.DataUpload{}).
+		Watches(&source.Kind{Type: &corev1.Pod{}}, kube.EnqueueRequestsFromMapUpdateFunc(r.findDataUploadForPod),
```
blackpiglet, line 334: Better to have some comments to explain why setting up the Manager like this.

pkg/controller/data_upload_controller.go
```diff
@@ -0,0 +1,498 @@
...
+
+		if !accepted {
+			log.Debug("Data upload is not accepted")
+			return ctrl.Result{}, nil
+		}
+
+		log.Info("Data upload is accepted")
+
+		exposeParam := r.setupExposeParam(&du)
+		if err := ep.Expose(ctx, getOwnerObject(&du), du.Spec.CSISnapshot.VolumeSnapshot, du.Spec.OperationTimeout.Duration, exposeParam); err != nil {
```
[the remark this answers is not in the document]
qiuming-best: What about we change the Expose interface to `Expose(context.Context, corev1.ObjectReference, interface{}) error`
for every specific snapshot type, they could get the parameter values from the interface{}?

pkg/controller/data_upload_controller.go
qiuming-best: For Expose interface `Expose(context.Context, corev1.ObjectReference, string, time.Duration, interface{}) error` which is an interface and it's unrelated to one specific type of exposer, but it needs to pass interface{} type of value which is related to one specific type of exposer.  

which is awkward, so we need one function(setupExposeParam) to prepare the value of the interface{}. Here we have the `CSI Snapshot exposer`,  maybe later we'll have other kinds of exposer. in order to organize unseen values of the interface{} for other types of exposer in the future, I need to pass all values into setupExposeParam.

pkg/controller/data_upload_controller.go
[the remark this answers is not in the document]
Lyndon-Li: Yes, we can put the snapshot ID to the parameter interface

pkg/controller/data_upload_controller.go
blackpiglet: If the parameters used by other types of exposers in the future are accessible from the `DataUploadReconciler` structure, then I think that's fair enough to keep it as the structure's method.

pkg/controller/data_upload_controller.go
qiuming-best: `ep.Expose()` will trigger to create one pod whose volume is restored by a given volume snapshot, but the pod maybe is not in the same node of the current controller, so we need to return here.
And then only the controller who is in the same node could do the rest work.

pkg/controller/data_upload_controller.go
qiuming-best: Added

pkg/controller/data_upload_controller.go
qiuming-best: Done

pkg/controller/data_upload_controller.go
```diff
@@ -0,0 +1,498 @@
...
+	if err := r.client.Get(ctx, req.NamespacedName, &du); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Debug("Unable to find DataUpload")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, errors.Wrap(err, "getting DataUpload")
+	}
+
+	if du.Spec.DataMover != "" && du.Spec.DataMover != dataMoverType {
+		log.WithField("Data mover", du.Spec.DataMover).Info("it is not one built-in data mover which is not supported by Velero")
```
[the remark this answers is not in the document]
qiuming-best: Modified

pkg/controller/data_upload_controller.go
```diff
@@ -0,0 +1,498 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	ep, ok := r.snapshotExposerList[du.Spec.SnapshotType]
+	if !ok {
+		return ctrl.Result{}, fmt.Errorf("%v type of snapshot exposer is not exist", du.Spec.SnapshotType)
```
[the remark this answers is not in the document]
qiuming-best: Added

pkg/controller/data_upload_controller.go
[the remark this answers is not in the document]
qiuming-best: I'm not sure why we need snapshot ID as parameter?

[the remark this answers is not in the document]
Lyndon-Li: > Yes, we can put the snapshot ID to the parameter interface

Sorry for the typo, it should be "snapshot name". I see the current change that move snapshot name and timeout to parameter, it is fine.

Lyndon-Li (approved):

config/rbac/role.yaml
```diff
@@ -102,6 +102,26 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - velero.io
+  resources:
+  - dataupload
```
Lyndon-Li, line 108: This doesn't look correct, we've already have the role declaration for `datauploads` in the same file. And `datauploads` looks to be the correct resource name instead of `dataupload`

pkg/controller/data_upload_controller.go
blackpiglet: Understood, then please also add some comments to explain the reason.

pkg/controller/data_upload_controller.go
```diff
@@ -0,0 +1,526 @@
...
+	if err := r.client.Patch(ctx, &du, client.MergeFrom(original)); err != nil {
+		log.WithError(err).Error("Failed to update progress")
+	}
+}
+
+// SetupWithManager registers the DataUpload controller.
+// The fresh new DataUpload CR first created will trigger to create one pod (long time, maybe failure or unknown status) by one of the dataupload controllers
+// then the request will get out of the Reconcile queue immediately by not blocking others' CR handling, in order to finish the rest data upload process we need to
+// re-enqueue the previous related request once the related pod is in running status to keep going on the rest logic. and below logic will avoid handling the unwanted
+// pod status and also avoid black others CR handling
```
Lyndon-Li, line 351: black->block

pkg/controller/data_upload_controller.go
qiuming-best: Added

config/rbac/role.yaml
qiuming-best: yes, I've removed it

pkg/controller/data_upload_controller.go
qiuming-best: modified

Lyndon-Li (approved):

blackpiglet (approved):

PR #6349, opened 2023-06-03
[trivy]remove unused code

wawa0210:

remove unused code

Fixes #(issue)

### DIFF
--- a/pkg/cmd/cli/serverstatus/server_status.go
+++ b/pkg/cmd/cli/serverstatus/server_status.go
@@ -54,12 +54,6 @@ func (g *DefaultServerStatusGetter) GetServerStatus(kbClient kbclient.Client) (*
 			return
 		}
 
-		// TODO: once the minimum supported Kubernetes version is v1.9.0, remove the following check.
-		// See http://issue.k8s.io/51046 for details.
-		if updated.Name != created.Name {
-			return
-		}
-
 		if updated.Status.Phase == velerov1api.ServerStatusRequestPhaseProcessed {
 			created = updated
 			cancel()


qiuming-best:
@wawa0210 What is the starting point for removing this code？ I'm not sure it could be deleted safely

wawa0210:
> @wawa0210 What is the starting point for removing this code？ I'm not sure it could be deleted safely

https://github.com/vmware-tanzu/velero#velero-compatibility-matrix

According to this support matrix document, the minimum K8s version supported by velero is 1.20.5?

qiuming-best:
> > @wawa0210 What is the starting point for removing this code？ I'm not sure it could be deleted safely
> 
> https://github.com/vmware-tanzu/velero#velero-compatibility-matrix
> 
> According to this support matrix document, the minimum K8s version supported by velero is 1.20.5?

@wawa0210 

```
// TODO: once the minimum supported Kubernetes version is v1.9.0, remove the following check.
		// See http://issue.k8s.io/51046 for details.
		if updated.Name != created.Name {
			return
		}
```
this issue http://issue.k8s.io/51046, it's focus on solving CRD couldn't be watched by name, namespace in k8s apiserver, 
it's not your problem, I'm just not understand why the `if updated.Name != created.Name ` condition is related to this issue.

why we should have this `if updated.Name != created.Name ` condition before  and after v1.9.0, k8s apiserver support watch CRD by name, namespace, why we should remove it?

I do not really know the reason

wawa0210:
> I looked into this a while back at one point as well. It wasn't obvious to me how the issue was related, although I imagine that Velero hit a bug a few years ago that this check fixed and (perhaps) is unnecessary with newer kubernetes. But if you want to remove the check, we need to confirm that it's no longer needed -- two possiblities:
> 
> 1. maybe with current Kube versions, it's impossible for these values to be different.
> 2. maybe with current Kube versions, it doesn't matter whether these values are different.
> 
> In the absence of that, it's probably safer to leave this code as-is.

Thank you for your patient reply. I just felt that this code was no longer needed based on the comments. I will close the current PR first, and I will make comments after I dig it clearly

PR #6446, opened 2023-06-30
Fix data path concurrent

qiuming-best:

Fixes #(issue)
Two Datauploads are committed to a node, they are executed sequentially. However, when executing the second one but the first one is still in processing,  the second dataupload will get stuck in process.

### DIFF
--- a/pkg/controller/data_download_controller.go
+++ b/pkg/controller/data_download_controller.go
@@ -161,7 +161,25 @@ func (r *DataDownloadReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			return ctrl.Result{}, nil
 		}
 
-		log.Info("Restore PVC is ready")
+		log.Info("Restore PVC is ready and creating data path routine")
+
+		// Need to first create file system BR and get data path instance then update data upload status
+		callbacks := datapath.Callbacks{
+			OnCompleted: r.OnDataDownloadCompleted,
+			OnFailed:    r.OnDataDownloadFailed,
+			OnCancelled: r.OnDataDownloadCancelled,
+			OnProgress:  r.OnDataDownloadProgress,
+		}
+
+		fsRestore, err = r.dataPathMgr.CreateFileSystemBR(dd.Name, dataUploadDownloadRequestor, ctx, r.client, dd.Namespace, callbacks, log)
+		if err != nil {
+			if err == datapath.ConcurrentLimitExceed {
+				log.Info("Data path instance is concurrent limited requeue later")
+				return ctrl.Result{Requeue: true, RequeueAfter: time.Minute}, nil
+			} else {
+				return r.errorOut(ctx, dd, err, "error to create data path", log)
+			}
+		}
 
 		// Update status to InProgress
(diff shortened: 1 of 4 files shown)


blackpiglet (approved):

pkg/controller/data_upload_controller.go
```diff
@@ -162,6 +162,10 @@ func (r *DataUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 		log.Info("Exposed snapshot is ready")
 
+		if r.dataPathMgr.IsConcurrentExceed() {
```
Lyndon-Li, line 165: This doesn't fix the problem all the time:

- Say current concurrent num is max-1, so only one more is allowed
- Then two calls `r.dataPathMgr.IsConcurrentExceed` at the same time, they both get permitted since concurrent num is not changed for each call
- Then they both call `dataPathMgr.CreateFileSystemBR` and finally one get passed and the other one is errorOut

A better way is we move the call to `r.dataPathMgr.CreateFileSystemBR` to here. Don't need to add `IsConcurrentExceed` function

pkg/controller/data_download_controller.go
```diff
@@ -163,6 +163,11 @@ func (r *DataDownloadReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 		log.Info("Restore PVC is ready")
 
+		if r.dataPathMgr.IsConcurrentExceed() {
```
Lyndon-Li, line 166: This doesn't fix the problem all the time:

Say current concurrent num is max-1, so only one more is allowed
Then two calls r.dataPathMgr.IsConcurrentExceed at the same time, they both get permitted since concurrent num is not changed for each call
Then they both call dataPathMgr.CreateFileSystemBR and finally one get passed and the other one is errorOut
A better way is we move the call to r.dataPathMgr.CreateFileSystemBR to here. Don't need to add IsConcurrentExceed function

Lyndon-Li, line 166: This doesn't fix the problem all the time:

- Say current concurrent num is max-1, so only one more is allowed
- Then two calls r.dataPathMgr.IsConcurrentExceed at the same time, they both get permitted since concurrent num is not changed for each call
- Then they both call dataPathMgr.CreateFileSystemBR and finally one get passed and the other one is errorOut

A better way is we move the call to r.dataPathMgr.CreateFileSystemBR to here. Don't need to add IsConcurrentExceed function

Lyndon-Li (changes requested):

pkg/controller/data_download_controller.go
qiuming-best: @Lyndon-Li  I'm changed to use `IsConcurrentExceed` function as it's more clear for the logic to write and read.
Also, I've noticed that it has a concurrent problem but the probability of that happening is very small before I add this logic

Lyndon-Li (dismissed):

blackpiglet (approved):

Lyndon-Li (approved):

blackpiglet (approved):

PR #6452, opened 2023-07-03
Add support for ResourceModifier (AKA Json Substitutions) in restore flow

### LINKED ISSUE #5809: Substitutions of YAML / JSON values through configmap/ RestoreItemAction

anshulahuja98:

Fixes #5809
Design: #5880 


internal/resourcemodifiers/resource_modifiers_test.go
```diff
@@ -0,0 +1,152 @@
+package resourcemodifiers
```
anshulahuja98, line 1: will be adding more test cases over next few days.

test/e2e/util/velero/velero_utils.go
```diff
@@ -179,6 +179,7 @@ func checkRestorePhase(ctx context.Context, veleroCLI string, veleroNamespace st
 		return err
 	}
 	if restore.Status.Phase != expectedPhase {
+		fmt.Println(restore.Status.FailureReason)
```
blackpiglet, line 182: Please put more details information into this log output.

anshulahuja98: removed, was just put for some debugging.

anshulahuja98:
Would be adding E2E tests as part of separate PR. 
Will try to add documentation update as part of this same PR.

internal/resourcemodifiers/resource_modifiers_validator.go
```diff
@@ -0,0 +1,55 @@
...
+		if err := rule.Validate(); err != nil {
+			return errors.WithStack(err)
+		}
+	}
+
+	return nil
+}
+
+func (p *JSONPatch) Validate() error {
+	// TODO validate allowed operation
```
ywk253100, line 40: Will the `TODO` item be implemented in this PR or not?

internal/resourcemodifiers/resource_modifiers.go
```diff
@@ -0,0 +1,152 @@
...
+func (r *ResourceModifierRule) Apply(obj *unstructured.Unstructured, groupResource string, log logrus.FieldLogger) error {
+	namespaceInclusion := collections.NewIncludesExcludes().Includes(r.Conditions.Namespaces...)
+	if !namespaceInclusion.ShouldInclude(obj.GetNamespace()) {
+		return nil
+	}
+	if !strings.EqualFold(groupResource, r.Conditions.GroupKind) {
+		return nil
+	}
+	if r.Conditions.ResourceNameRegex != "" {
+		match, _ := regexp.MatchString(r.Conditions.ResourceNameRegex, obj.GetName())
```
ywk253100, line 88: Handle error

internal/resourcemodifiers/resource_modifiers.go
```diff
@@ -0,0 +1,152 @@
...
+func ApplyPatch(patch []byte, obj *unstructured.Unstructured, log logrus.FieldLogger) error {
+	jsonPatch, err := jsonpatch.DecodePatch(patch)
+	if err != nil {
+		return fmt.Errorf("error in decoding json patch %s", err.Error())
+	}
+	objBytes, err := obj.MarshalJSON()
+	if err != nil {
+		return fmt.Errorf("error in marshaling object %s", err.Error())
+	}
+	modifiedObjBytes, err := jsonPatch.Apply(objBytes)
```
ywk253100, line 127: Seems an error is returned if no path matches the `test` operation. Correct me if I'm wrong.
So that means if users set the `test` patch, they must make sure it matches, or the restore will fail.

anshulahuja98: the restore will not fail
just that particular patch would fail and restore will go on as it is.

For example there are 5 PVCs which match the "conditions"

but only 1 PVC which satisfies the "test" operator. Only that PVC will be patched and all the others will get skipped in some sense. 
The logs will show that, but there won't be any hard failure.

internal/resourcemodifiers/resource_modifiers_validator.go
anshulahuja98: updated.

pkg/restore/restore.go
```diff
@@ -1355,6 +1358,14 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 	// and which backup they came from.
 	addRestoreLabels(obj, ctx.restore.Name, ctx.restore.Spec.BackupName)
 
+	if ctx.resourceModifiers != nil {
+		if errList := ctx.resourceModifiers.ApplyResourceModifierRules(obj, groupResource.String(), ctx.log); errList != nil {
+			for _, err := range errList {
+				errs.Add(namespace, err)
```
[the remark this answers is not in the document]
anshulahuja98: Consider this as more of "partialfail" scnearios.
Substitution not happening is not hard failure, we will let the restore continue as it is.
For example in the test operator - we'll get an error back saying the test patch did not work on a specific resource, but that does not mean we skip restoring, it's just that we restore it in initial state.

internal/resourcemodifiers/resource_modifiers.go
anshulahuja98: when match comes false, an error will come
I don't see additional use in handling this error
since we just need to rely on the "match" value

Logging it on the other hand might just bloat logs since match will be called for each resource which matches NS/ GroupKind filters.

ywk253100: Could you explain more about `when match comes false, an error will come`? When the match is false, the current code returns nil directly.

There is a case that the regex provided may contain syntax errors, if we don't handle such kinds of errors, users don't know about that and the restore may completed but with unexpected resources restored.

internal/resourcemodifiers/resource_modifiers.go
ywk253100: Yes, the error is shown in the logs, but won't the restore be marked as partially failed? Is this the expected behavior?

ywk253100:
@anshulahuja98 Please also add some more unit test cases. We have one task in the 1.12 timeframe to improve the unit test coverage to be >60%.

internal/resourcemodifiers/resource_modifiers.go
anshulahuja98: Handled the "test operator related errors separately to avoid partial failure states based on your feedback. Updated code with test case.

internal/resourcemodifiers/resource_modifiers.go
anshulahuja98: I understand your concern, validated what you explained with a POC.
Have updated the PR with the fix for handling and also added test case for it.

anshulahuja98:
> 

Sure will do.

ywk253100 (dismissed):

ywk253100 (approved):

anshulahuja98:
@sseago / @blackpiglet  can you please review this PR.

blackpiglet (dismissed):

sseago:
@anshulahuja98 looks like there's a conflict to resolve in this now

anshulahuja98:
Fixed the merge conflict @sseago 
@blackpiglet  / @ywk253100  you folks would have to approve again post merge conflict resolution.

ywk253100 (approved):

blackpiglet (approved):

Issue #6422, opened 2023-06-23
Backups are not listed after recreating GKE cluster


vijaymck:
**What steps did you take and what happened:**
We had to perform a complete restore to a particular backup. Delete the GKE cluster and recreated using TF. Installed Velero using helm. No backups are listed through velero cli. Backup folders does exist in GCP storage. Also able to create new backups and list them. 
 
ONLY NEW backups are listed:

```
$ velero backup get
NAME                         STATUS      ERRORS   WARNINGS   CREATED                         EXPIRES   STORAGE LOCATION   SELECTOR
backup-2023-06-22-13-11-29   Completed   0        0          2023-06-22 13:11:41 -0400 EDT   29d       default            <none>
backup-2023-06-23-01-01-46   Completed   0        0          2023-06-22 21:01:51 -0400 EDT   29d       default            <none>
```

"Skipping non-expired incomplete backup" msg from the logs (below & attached), indicates that plugin error:
https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/backup_sync_controller.go#L156

but, don't see much in plugin logs:
Copying /plugins/velero-plugin-for-gcp to /target/velero-plugin-for-gcp ...  done.

**What did you expect to happen:**
All backups from the storage account are to be listed.

**The following information will help us better understand what's going on**:
Enabled debug logging. Attached complete velero log from debug command.

```
time="2023-06-23T09:16:44Z" level=info msg="Validating BackupStorageLocation" backup-storage-location=velero/default controller=backup-storage-location logSource="pkg/controller/backup_storage_location_controller.go:152"
time="2023-06-23T09:16:44Z" level=info msg="BackupStorageLocations is valid, marking as available" backup-storage-location=velero/default controller=backup-storage-location logSource="pkg/controller/backup_storage_location_controller.go:137"
time="2023-06-23T09:16:44Z" level=debug msg="received EOF, stopping recv loop" backup-storage-location=velero/default cmd=/plugins/velero-plugin-for-gcp **controller=backup-storage-location err="rpc error: code = Unavailable desc = error reading from server: read unix @->/tmp/plugin3828803423: read: connection reset by peer"** logSource="pkg/plugin/clientmgmt/process/logrus_adapter.go:75" pluginName=stdio
[...]

PR #6474, opened 2023-07-07
Add immutability supporting limitation in documents.

blackpiglet:

Fixes #(issue)

### DIFF
--- a/site/content/docs/main/backup-reference.md
+++ b/site/content/docs/main/backup-reference.md
@@ -63,6 +63,8 @@ This command will immediately trigger a new backup based on your template for `e
 
 
 ### Limitation
+
+#### Backup's OwnerReference with Schedule
 Backups created from schedule can have owner reference to the schedule. This can be achieved by command:
 
 ```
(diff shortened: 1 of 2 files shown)


site/content/docs/main/backup-reference.md
```diff
@@ -74,6 +76,8 @@ Please do notice there is also side effect that may not be expected. Because sch
 
 If there is possibility the schedule will be disable to not create backup anymore, and the created backups are still useful. Please do not enable this option. For detail, please reference to [Backups created by a schedule with useOwnerReferenceInBackup set do not get synced properly](https://github.com/vmware-tanzu/velero/issues/4093).
 
+#### Cannot support bucket's RetetionPolicy
+From Velero's release-1.11 branch, Velero cannot support BSL's bucket with RetentionPolicy setting. Because Velero introduced a `Finalizing` state, Velero will first enter `Finalizing` phase after backup completion, and then the Finalizing controller will check whether there is any async operation in progress. If there is, then Velero will wait for all async operations complete, then turn backup into `Complete` phase. If there is not, then Velero will turn into `Complete` directly. By current design, whether there is async operation included or not, Velero's metadata in the OSS bucket need to change after it's created. This is the reason why RetentionPolicy cannot be supported.
```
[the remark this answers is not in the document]
blackpiglet: Thanks for the suggestion.
I will reword the limitation to a more generic phrase, and add the Azure, AWS, and GCP document links.

blackpiglet:
This address the need to document the data immutability limitation in issue #6422.

blackpiglet:
@draghuram 
Could you help PTAL?

danfengliu (dismissed):

draghuram:
Sorry for the delay, will take a look.

draghuram:
I made some changes to the text. Please feel free to make more changes as you see fit. It is a very difficult topic to document but it is better to have something and then we can improve it. 

### My version

Starting from 1.11, Velero's backups may not work as expected when the target object storage has some kind of an "immutability" option configured. These options are known by different names (see links below for some examples). The main reason is that Velero first saves the state of a backup as `Finalizing` and then checks whether there are any async operations in progress. If there are, it needs to wait for all of them to be finished before moving the backup state to `Complete`. If there are no async operations, the state is moved to `Complete` right away. In either case, Velero needs to modify the metadata in object storage and that will not be possible if some kind of immutability is configured on the object storage.

Note that backups may still work in some cases depending on specific provider and configuration. 

* For AWS S3 service, backups work because S3's object lock only applies to versioned buckets and the object data can still be updated as the new version. But when backups are deleted, old versions of the objects will not be deleted.
* Azure Storage Blob supports both versioned-level immutability and container-level immutability. For the versioned-level scenario, data immutability can still work in Velero, but the container-level cannot.
* GCP Cloud storage policy only supports bucket-level immutability, so there is no way to make it work in the GCP environment.

The following are the links to cloud providers' documentation in this regard:
* [AWS S3 Using S3 Object Lock](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock.html)
* [Azure Storage Blob Containers - Lock Immutability Policy](https://learn.microsoft.com/en-us/azure/storage/blobs/immutable-policy-configure-version-scope?tabs=azure-portal)
* [GCP cloud storage Retention policies and retention policy locks](https://cloud.google.com/storage/docs/bucket-lock)

blackpiglet:
@draghuram 
Thanks a lot! That looks better. 
Just to add some information from Velero's design perspective.

site/content/docs/main/backup-reference.md
```diff
@@ -74,7 +76,23 @@ Please do notice there is also side effect that may not be expected. Because sch
...
 
+#### Cannot support backup data immutability
+Starting from 1.11, Velero's backups may not work as expected when the target object storage has some kind of an "immutability" option configured. These options are known by different names (see links below for some examples). The main reason is that Velero first saves the state of a backup as Finalizing and then checks whether there are any async operations in progress. If there are, it needs to wait for all of them to be finished before moving the backup state to Complete. If there are no async operations, the state is moved to Complete right away. In either case, Velero needs to modify the metadata in object storage and that will not be possible if some kind of immutability is configured on the object storage.
+
+Please also notice, before 1.11, even Velero can work with the target object storage with an "immutability" configuration, Velero didn't provide any "immutability" options. It can work because there is no limitation, but no official support in the data "immutability" function too.
```
draghuram, line 82: This paragraph is a bit confusing to me. Do you mean to say that Velero pre-1.11 will work with immutability option though there will be some issues (such as old versions not being deleted)?

blackpiglet: No. I mean to say that cannot supporting immutability is a new introducing issue, but Velero doesn't provide any support on this before.

draghuram: OK. In that case, I would suggest this:

"Even with versions prior to 1.11, there was no explicit support in Velero to work with object storage that has  "immutability" configuration. As a result, you may see some problems even though backups seem to work (e.g. versions objects not being deleted when backup is deleted)."

blackpiglet: Thanks a lot for the rephrasing.

Lyndon-Li (approved):

danfengliu (approved):

PR #6479, opened 2023-07-10
add restore finalizer to clean up external resources

### LINKED ISSUE #2697: unable to delete velero restore objects from the cloud storage

I was trying to delete the restore objects using the below command but it is not deleting the restore objects from cloud storage.

`velero restore delete ${restore_name}`

It should delete the restore from velero as well as from the cloud storage

**Server Side logs**
```
root@testvm:~# velero restore get
NAME                        BACKUP                        STATUS      WARNINGS   ERRORS   CREATED                         SELECTOR
wordpress-mysql-service     velero-daily-20200708054623   Completed   6          0        2020-07-08 05:48:20 +0000 UTC   <none>
wordpress-service-restore   velero-daily-20200708054623   Completed   6          0        2020-07-08 05:49:48 +0000 UTC   <none>
root@testvm:~# velero restore delete --all
Are you sure you want to continue (Y/N)? Y
Restore "wordpress-mysql-service" deleted
Restore "wordpress-service-restore" deleted
root@testvm:~# velero restore get
root@testvm:~#
```

**Cloud side restore components are not deleting** 

github

**Environment:**

- Velero version (use `velero version`): 1.4.0
- OS (e.g. from `/etc/os-release`): ubuntu

**Vote on this issue!**

allenxu404:

- Add restore finalizer to clean up external resources when restore is being deleted

Fixes #2697

### DIFF
--- a/pkg/cmd/cli/uninstall/uninstall.go
+++ b/pkg/cmd/cli/uninstall/uninstall.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"github.com/vmware-tanzu/velero/pkg/client"
 	"github.com/vmware-tanzu/velero/pkg/cmd"
 	"github.com/vmware-tanzu/velero/pkg/cmd/cli"
@@ -159,6 +160,12 @@ func Run(ctx context.Context, kbClient kbclient.Client, namespace string) error 
 }
 
 func deleteNamespace(ctx context.Context, kbClient kbclient.Client, namespace string) error {
+	// delete resources with finalizer attached first to ensure finalizer can be handled by corresponding controller and resources can be deleted successfully before controller's pod is deleted.
+	// otherwise the process of deleting namespace will get stuck in deleting those resources forever.
+	if err := deleteResourcesWithFinalizer(ctx, kbClient, namespace); err != nil {
+		return errors.Wrap(err, "Fail to remove finalizer from restores")
+	}
+
 	ns := &corev1.Namespace{}
 	key := kbclient.ObjectKey{Name: namespace}
 	if err := kbClient.Get(ctx, key, ns); err != nil {
@@ -176,7 +183,7 @@ func deleteNamespace(ctx context.Context, kbClient kbclient.Client, namespace st
 		}
 		return err
 	}
-
+	fmt.Println()
 	fmt.Printf("Waiting for velero namespace %q to be deleted\n", namespace)
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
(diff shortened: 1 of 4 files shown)


allenxu404:
> @allenxu404 Does this PR cover the upgrade case? What will happen to the existing `Restores`?

@ywk253100 "The existing restores will not have a finalizer added and they will be deleted directly the way it was.

pkg/controller/restore_controller.go
```diff
@@ -222,25 +257,45 @@ func (r *restoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		return true
+	default:
+		return false
+	}
+}
+
+func updatePredicate(ue event.UpdateEvent) bool {
+	restore := ue.ObjectNew.(*api.Restore)
+	if !restore.DeletionTimestamp.IsZero() && len(restore.Finalizers) != 0 && (restore.Status.Phase == api.RestorePhaseFailed ||
+		restore.Status.Phase == api.RestorePhasePartiallyFailed || restore.Status.Phase == api.RestorePhaseCompleted) {
```
ywk253100, line 293: Before this change, a restore can be deleted during the processing, but now it cannot?

ywk253100:
@allenxu404  Current changes that filter based on different events are complicated, how about that we handle the restore directly without considering the event type as follows:

1. Get the restore
2. If the restore doesn't exist, return directly
3. If the deletion timestamp isn't empty, clean up the resource in store, remove the finalizer, and return
4. Reconcile the finalizer if the restore isn't set. - This covers the upgrade case
5. Process the restore if it is new as previous

An example is https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/controllers/machine/machine_controller.go

ywk253100:
And please also the new code is covered by the unit test

pkg/controller/restore_controller.go
allenxu404: Yes, it can be deleted during the processing as well after this change. The finalizer is added in terminal phase, so restores in the progress won't get attached with finalizer and they can be deleted directly.

allenxu404:
> @allenxu404 Current changes that filter based on different events are complicated, how about that we handle the restore directly without considering the event type as follows:
> 
>     1. Get the restore
> 
>     2. If the restore doesn't exist, return directly
> 
>     3. If the deletion timestamp isn't empty, clean up the resource in store, remove the finalizer, and return
> 
>     4. Reconcile the finalizer if the restore isn't set. - This covers the upgrade case
> 
>     5. Process the restore if it is new as previous
> 
> 
> An example is https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/controllers/machine/machine_controller.go

@ywk253100 Yeah, make sense. I made an enhancement to the code and it functions following the procedures outlined aboved.

pkg/controller/restore_controller.go
```diff
@@ -155,10 +159,58 @@ func (r *restoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			log.Debugf("restore[%s] not found", req.Name)
+			return ctrl.Result{}, nil
+		}
+
+		log.Errorf("Fail to get restore %s: %s", req.NamespacedName.String(), err.Error())
 		return ctrl.Result{}, err
 	}
 
+	// deal with finalizer
+	if !restore.DeletionTimestamp.IsZero() && len(restore.Finalizers) != 0 {
```
ywk253100, line 172: Why do we need to check the finalizer? What will happen if we don't check the finalizer?

pkg/controller/restore_controller.go
```diff
@@ -155,10 +159,58 @@ func (r *restoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			original := restore.DeepCopy()
+			controllerutil.RemoveFinalizer(restore, ExternalResourcesFinalizer)
+			if err := kubeutil.PatchResource(original, restore, r.kbClient); err != nil {
+				log.Errorf("fail to remove finalizer: %s", err.Error())
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{}, nil
+		} else {
+			log.Error("DeletionTimestamp is marked but can't find the expected finalizer")
+			return ctrl.Result{}, fmt.Errorf("DeletionTimestamp is marked but can't find the expected finalizer")
```
ywk253100, line 189: If we return an error here, the `Restore` will enqueue later, is this the expected behavior?

pkg/controller/restore_controller.go
allenxu404: Yes, took another look at it and see it's fine to remove the condition of checking the length of finalizer.

pkg/cmd/cli/uninstall/uninstall.go
```diff
@@ -203,3 +210,48 @@ func deleteNamespace(ctx context.Context, kbClient kbclient.Client, namespace st
...
+	for i := range restoreList.Items {
+		if err := kbClient.Delete(ctx, &restoreList.Items[i]); err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			return err
+		}
+	}
+
+	fmt.Println("Waiting for resource with finalizer attached to be deleted")
```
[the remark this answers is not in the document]
allenxu404: if we output "Waiting for restore to be deleted" to the users, users might be confused about the need for an additional operation to delete the restores. What's your thought?

ywk253100 (approved):

qiuming-best (approved):

Issue #6443, opened 2023-06-30
Document backup repository password configuration method


Lyndon-Li:
At present, Velero has a default backup repository password saved in secret `velero-repo-credentials`, the way to configure the password to to modify this secret.
However, this secret and its usage is not documented, many users don't know this.

shubham-pampattiwar:
@Lyndon-Li I can work on this issue if needed.

Lyndon-Li:
@shubham-pampattiwar 
Sure, thanks for your efforts! I am assigning the issue to you.

clouedoc:
Does this password cause the backups to S3 to be encrypted?
I'm a new user trying to ensure that my backups S3 backups are encrypted at rest

draghuram:
@clouedoc, that is correct. Both Restic and Kopia (tools that Velero uses to do file system backups) encrypt data using the repo password. Check their documentation which has more details.  But do check the limitations mentioned at https://velero.io/docs/v1.11/file-system-backup/#limitations.

draghuram:
@shubham-pampattiwar, @Lyndon-Li, Documenting this would be very useful. Note that we should cover the case where user wants to modify the secret after few backups were done. Would this result in repo password being changed or it applies to only new repos?

Lyndon-Li:
@draghuram
This is a very delicate case, or in another words, Velero doesn't support modifying password of an existing repo.
So this password configuration should only be used in the scenarios that after installing Velero, users want to use their own password to create the repo. The configuration should happen before any repo is created.

Configuring the secret after repos are created is possible, but users will see different manners for new and old repos:

- For new repos, they will use the new password in the secret
- For old repos, users will not be able to use the repos since the repos are using the old password but Velero is using the new one to connect them

Lyndon-Li:
Changing password for existing repos is another task which involves lots of changes, though for sure it is a useful feature, I've opened a new issue #6537 to track it.

sseago:
This is further complicated by the fact that a BSL can be used in several clusters, so if any one cluster that uses this BSL for filesystem backups changes the password, all velero instances that use this need the same change.

draghuram:
Agreed. I think the password configuration should be documented sooner rather than later.

Lyndon-Li:
@sseago 
Yes, once we support changing password for an existing repo, any instance of Velero that uses the BSL will need to use the new password, so we need a sophisticated design for this.
I will add this into the same issue #6537.

[...]

PR #6480, opened 2023-07-10
Create design for client side encryption

### LINKED ISSUE #434: Encrypt backups at rest
Right now, when backups are created via `ark backup create`, sensitive objects are stored unencrypted at rest. In Google Cloud, there is excellent Go support for encrypting Google Cloud Storage objects with [Google Cloud KMS](https://cloud.google.com/kms/docs/encrypt-decrypt#kms-howto-encrypt-go) in a way that [works rather transparently to the caller](https://godoc.org/cloud.google.com/go/storage#ObjectHandle.Key).

I would love a way to configure the KMS key to use when storing backups. The best practice is to have a separate project for the key and grant IAM permissions from there, which I could easily do for the `heptio-ark` SA that already is required when setting up access to the bucket.

At @kolide, we have an internal tool that works like that which I'd like to potentially replace with Ark, so if there is a reasonable integration point within Ark for this kind of thing, perhaps some of our existing code can be upstreamed for this use-case.

### LINKED ISSUE #3218: S3 client-side encryption support
In addition to server-side encryption requested in #1782, it would be great to also support client-side encryption with client-provided master key.

This would allow using not fully trusted Minio instance for example. As far as I know, right now data is encrypted using static key which does not add any protection.

Add support for specifying client-side encryption master key for S3 backup storage location.

S3 documentation: https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingClientSideEncryption.html

I guess this could be workaround using some sort of S3 proxy, which would handle the encryption independent from Velero.

**Vote on this issue!**

jelemux:

# Summary
Currently, backups are not encrypted at rest.
This PR introduces a design proposal that aims to fix this while still retaining backwards compatibility.
A proof-of-concept implementation can be found at https://github.com/cloudogu/velero/tree/feature/client-side-encryption.

Fixes #434
Fixes #3218

- [ ] ~~Updated the corresponding documentation in `site/content/docs/main`.~~ not necessary, since this is a design doc

### DIFF
--- a/design/client-side-encryption_design.md
+++ b/design/client-side-encryption_design.md
@@ -1,0 +1,554 @@
+# Proposal for client-side encryption
+
+## Abstract
+
+Currently only PersistentVolumes are stored in a Restic/Kopia repository, which means that only they are encrypted and the rest of the backup is not.
+It is desired that the cluster backup can be encrypted which overall will lead to better security.
+
+## Background
+
+As of today, Velero stores only PersistentVolume data in a Restic/Kopia repository.
+This repository is encrypted while the rest of the backup is not.
+Backups should be safe against illegal access and tampering.
+It is therefore desired to have the ability to store this data in an encrypted manner.
+
+## Goals
+
+- Encryption of the main cluster backup file with AES.
+- Basic configuration of the encryption, e.g. turning it on and off.
+- Ability to change encryption key for future backups
+
+## Non Goals
+
+- Other ciphers than AES (can be added later).
+- Encrypting anything other than the main backup file.
+- re-encrypt old backups if new encryption key was set
+
+## High-Level Design
+
+Encryption will happen during the backup process before the objects are persisted (e.g. written to an object store).
+Decryption will happen during a restore after the persisted objects have been read.
+The encryption key will be stored in a Kubernetes secret by the user.
+Encryption can be configured through commandline arguments to the velero server.
+The user can change the encryption key by creating a new encryption key secret and configuring it in Velero's configuration.
+If a backup is encrypted, it contains a flag and the name of the used encryption key secret in its metadata.
+For backwards compatibility reasons, only the main backup archive will be encrypted.
+
+## Detailed Design
+
+### Configuration
+
+To enable encryption, a feature flag is introduced:
+```go
+// EncryptionFeatureFlag is the feature flag string that defines whether or not client-side-encryption features are being used.
+const EncryptionFeatureFlag = "EnableEncryption"
+```
+
+A commandline flag is added to the `velero server` command to set the name of the secret containing the encryption key:
+```go
+type serverConfig struct {
+    ...
+	encryptionSecret string
+}
+```
+
+```go
+command.Flags().StringVar(&config.encryptionSecret, "encryption-secret", config.encryptionSecret, "Name of the secret that contains the key to encrypt backups with. Used only if the 'EnableEncryption' feature flag is set.")
+```
+
+### En-/Decryption implementation
+
+Encryption is located in the `encryption` package.
+The `encryptor` interface has been introduced to provide abstraction over the used encryption standard.
+The default (and for now only) encryptor implementation is the `aesEncryptor`:
+```go
+type encryptor interface {
+	Encrypt(plaintext []byte) ([]byte, error)
+	Decrypt(ciphertext []byte) ([]byte, error)
+}
+
+type aesEncryptor struct {
+	gcm cipher.AEAD
+	key []byte
+}
+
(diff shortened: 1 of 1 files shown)


design/client-side-encryption_design.md
```diff
@@ -0,0 +1,554 @@
...
+This could be implemented by adding the fields `isEncrypted` and `encryptionSecret` to the backup spec instead of its status.
+An advantage of this would be that there is no need for a feature flag  or the `encryptionSecret` flag anymore.
+On the other hand, it would be fatal if a user forgets to encrypt the backup.
+But as most users will likely automate their backups anyway, this shouldn't pose that much of a problem.
+
+This could also be implemented in object store plugins.
+Then however, it has to be implemented in every single object store plugin.
+This is not viable.
+
+We could just make use of server-side encryption of an object store.
```
anshulahuja98, line 498: most cloud based Object stores  do support encryption at rest using user managed keys.

Can you give production examples where users are using object stores which don't support such functionality.

design/client-side-encryption_design.md
```diff
@@ -0,0 +1,554 @@
...
+
+- Other ciphers than AES (can be added later).
+- Encrypting anything other than the main backup file.
+- re-encrypt old backups if new encryption key was set
+
+## High-Level Design
+
+Encryption will happen during the backup process before the objects are persisted (e.g. written to an object store).
+Decryption will happen during a restore after the persisted objects have been read.
+The encryption key will be stored in a Kubernetes secret by the user.
```
anshulahuja98, line 31: what are your thoughts on using cloud provider based secret stores for example KeyVault in Azure / AWS Secret manager for maintaing the secret.

Because we are all aware that kubernetes secrets are just plaintext and do no come with any security.

Additionally the lifecycle of the secret is tied to the lifecycle of the cluster.

Storing clusters in secret stores outside might be more secure and durable. Think cross cluster restore

anshulahuja98 (changes requested):

design/client-side-encryption_design.md
anshulahuja98: I understand that this might not be specific use case for you, but I want to bring this up in context of design extensibility. How can current design extend to support such scenarios.
or if it is not planned to supported at all, please put it in non goals

anshulahuja98 (changes requested):

design/client-side-encryption_design.md
jelemux: I designed this feature explicitly in such a way that fetching secrets from a KMS can be implemented as well. See the `KeyRetriever` interface. Although, maybe this might have to be extended by a plugin API for it to be truly swappable.

design/client-side-encryption_design.md
```diff
@@ -0,0 +1,554 @@
+# Proposal for client-side encryption
+
+## Abstract
+
+Currently only PersistentVolumes are stored in a Restic/Kopia repository, which means that only they are encrypted and the rest of the backup is not.
+It is desired that the cluster backup can be encrypted which overall will lead to better security.
+
+## Background
+
+As of today, Velero stores only PersistentVolume data in a Restic/Kopia repository.
```
anshulahuja98, line 10: Can you perhaps explain how the Kopia repository is encrypted today. I am personally not aware of the mechanism, will be good to callout in this doc somewhere.

jelemux: There's a secret `velero-repo-credentials` containing the repository password. Per default, the password there has the value `static-passw0rd`. As you see, this is not a strong password and my guess is that the only reason this was used at all is because Kopia and Restic require encryption. While it is possible to change the password in the secret, it is as of yet not a documented feature.

draghuram: Documenting the secret is being tracked in #6443.

draghuram: Kopia encryption is described here: https://kopia.io/docs/advanced/encryption/.

design/client-side-encryption_design.md
draghuram: Data should also be encrypted in transit using TLS, though having client side encryption is useful as well.

design/client-side-encryption_design.md
anshulahuja98: Thanks for explaining @jelemux / and the info @draghuram

design/client-side-encryption_design.md
kaovilai: A case where this may help is if bsl is set to use insecure for whatever reason.. ie. User decides to use aws plugin and override s3url to use a minio server which is uses self signed cert and have not bothered to setup cacert properly.

PR #6483, opened 2023-07-11
E2E test for Resource Modifier & Bug fixes

anshulahuja98:

Primarily this is to add e2e test case for resourcemodifier functionality.

In Addition it has also some bug fixes to handle corner cases for integers / bool in the json patch.
E2E test for #6452 


anshulahuja98:
Will be fixing last minute issue by the last commit. But overall should be good enough to review.

anshulahuja98:
Had run the E2E test in the pipeline by editing the github actions Ginkgo focus tags. 
And had green run.

sseago (approved):

internal/resourcemodifiers/resource_modifiers.go
```diff
@@ -117,7 +118,24 @@ func (r *ResourceModifierRule) PatchArrayToByteArray() ([]byte, error) {
 }
 
 func (p *JSONPatch) ToString() string {
-	if strings.Contains(p.Value, "\"") {
+	// if value is empty, then add empty quotes
+	if p.Value == "" {
+		return fmt.Sprintf(`{"op": "%s", "from": "%s", "path": "%s", "value": ""}`, p.Operation, p.From, p.Path)
+	}
+	// if value is null, then don't add quotes
```
mateusoliveira43, line 125: what about creating a function like this
```go
func addQuotes(value interface{}) bool {
    // don't add quotes if
   // value is null
    if value == "null" {return false}
    // value is a boolean
    if _, err := strconv.ParseBool(value); err == nil {return false}
    // value is a json object or array
    if strings.Contains(value, "\"") || strings.Contains(value, "[") || strings.Contains(value, "{") {return false}
    // value is a number
    if _, err := strconv.ParseFloat(value, 64); err == nil {return false}
    return true
}
```

then change this part to
```go
if addQuotes(p.Value) {
    return fmt.Sprintf(`{"op": "%s", "from": "%s", "path": "%s", "value": "%s"}`, p.Operation, p.From, p.Path, p.Value)
} else {
    return fmt.Sprintf(`{"op": "%s", "from": "%s", "path": "%s", "value": %s}`, p.Operation, p.From, p.Path, p.Value)
}
```
?

internal/resourcemodifiers/resource_modifiers.go
```diff
@@ -117,7 +118,24 @@ func (r *ResourceModifierRule) PatchArrayToByteArray() ([]byte, error) {
 }
 
 func (p *JSONPatch) ToString() string {
-	if strings.Contains(p.Value, "\"") {
+	// if value is empty, then add empty quotes
+	if p.Value == "" {
```
mateusoliveira43, line 122: I would remove this first if, if falls back to implementation of line 141, right?

test/e2e/README.md
```diff
@@ -126,7 +126,7 @@ For programming conventions, Take ResourcePoliciesCase case for example:
 ##### CreateResources
 - It's better to set a global timeout in CreateResources function which is the real beginning of one e2e test
 ##### Destroy
-- It's only clean up resources in currently test namespaces, if wantting to clean up all resources including resources created which is not in currently test namespaces, it's better to override base Destroy function
+- It's only clean up resources in currently test namespaces, if you wish to clean up all resources including resources created which are not in currently test namespaces, it's better to override base Destroy function
```
mateusoliveira43, line 129: NIT "It's only clean up resources..." -> "It only cleans up resources..."?

internal/resourcemodifiers/resource_modifiers.go
```diff
@@ -117,7 +118,24 @@ func (r *ResourceModifierRule) PatchArrayToByteArray() ([]byte, error) {
...
+	// if value is null, then don't add quotes
+	if p.Value == "null" {
+		return fmt.Sprintf(`{"op": "%s", "from": "%s", "path": "%s", "value": %s}`, p.Operation, p.From, p.Path, p.Value)
+	}
+	// if value is a boolean, then don't add quotes
+	if _, err := strconv.ParseBool(p.Value); err == nil {
+		return fmt.Sprintf(`{"op": "%s", "from": "%s", "path": "%s", "value": %s}`, p.Operation, p.From, p.Path, p.Value)
+	}
+	// if value is a json object or array, then don't add quotes
+	if strings.Contains(p.Value, "\"") || strings.Contains(p.Value, "[") || strings.Contains(p.Value, "{") {
```
mateusoliveira43, line 134: here could we trigger false positives? like `"just\"string\""`, `today is Monday [12/09]`

internal/resourcemodifiers/resource_modifiers_test.go
mateusoliveira43, line 1: here I missed tests for `null`, `boolean` and just `array` (but there is already tests with objects that have array inside them) for the new cases of `ToString` function

internal/resourcemodifiers/resource_modifiers.go
anshulahuja98: thanks for the suggestion, will incorporate this.

internal/resourcemodifiers/resource_modifiers_test.go
anshulahuja98: will add

internal/resourcemodifiers/resource_modifiers.go
anshulahuja98: we could, but I am relying on assumption that most YAMLs don't generally keep values such as this.
On the other hand didn't want to overengineer this to allow false positives to go in as is.
Any suggestions on how we could fix it. Meanwhile I'll also think further and see if this can be resolved.

anshulahuja98: There could be other cases as well for boolean / null where the expecting k8s resource expects them inside quotes instead of bool / null

anshulahuja98: Perhaps I can make this check as "startswith" "["  or "{"

internal/resourcemodifiers/resource_modifiers.go
anshulahuja98: I would personally prefer this edge case handling
in case the integer check or some other check has handling for "" and maps it to default value.
this just gives sanity.

internal/resourcemodifiers/resource_modifiers.go
mateusoliveira43: Agree, for now lets keep simple (loved the idea with `startswith`)

mateusoliveira43 (approved):

anshulahuja98:
@sseago can you review again
@ywk253100  request to review.

sseago (approved):

ywk253100 (approved):

Issue #6519, opened 2023-07-18
Possible pod volume backup failure if velero is installed in multiple namespaces.


sseago:
**What steps did you take and what happened:**
It looks like we may have a regresssion in pod volume backup functionality when velero is installed in more than one namespace. It may have been introduced with the kopia integration effort. The expectation is that if velero is installed in more than one namespace, each velero instance only looks at velero CRs in its own namespace, and ignores them in other namespaces. In the past we've introduced bugs in this area with controller refactoring, but the ones previously observed have been fixed.

From the report in https://github.com/vmware-tanzu/helm-charts/issues/477 it looks like the pod volume backup controller is not properly limiting itself to PVBs in its own namespace:
```
Actually it seems that having Velero setuped in multiple namespace is not working. I think the Velero Pod can not make the difference between node agent pods of each namespace.

For me, when i create a backup with Restic Volume Backup on namespace A, Velero tried to use the node-agent pods of the namespace B
```
Error observed;
```
{
  "backup": "velero-syno/test-12",
  "error.file": "/go/src/github.com/vmware-tanzu/velero/pkg/podvolume/backupper.go:250",
  "error.function": "github.com/vmware-tanzu/velero/pkg/podvolume.(*backupper).BackupPodVolumes",
  "error.message": "pod volume backup failed: error creating uploader: failed to connect repository: error running command=restic snapshots --repo=s3:https://minio.example.com:10000/velero-bucket/restic/test-ceph-rbd --password-file=/tmp/credentials/velero-ovh/velero-repo-credentials-repository-password --cache-dir=/scratch/.cache/restic --latest=1 --insecure-tls=true, stdout=, stderr=Fatal: unable to open config file: Stat: The Access Key Id you provided does not exist in our records.\nIs there a repository at the following location?\ns3:https://minio.example.com:10000/velero-bucket/restic/test-ceph-rbd\n: exit status 1",
  "level": "error",
  "logSource": "pkg/backup/backup.go:435",
  "msg": "Error backing up item",
  "name": "test-585dbb4d95-zh4gb",
  "time": "2023-07-12T08:27:18Z"
}
```

**What did you expect to happen:**
restic backup should have worked properly in both velero instances without credentials errors.
**The following information will help us better understand what's going on**:

_If you are using velero v1.7.0+:_  
Please use `velero debug  --backup <backupname> --restore <restorename>` to generate the support bundle, and attach to this issue, more options please refer to `velero debug --help` 

[...]

PR #6523, opened 2023-07-19
Restrict namespace to node-agent cache

Lyndon-Li:
Fix issue #6519. Restrict the client manager of node-agent server to include only Velero resources from the server's namespace, otherwise, the controllers will try to reconcile CRs from all the installed Velero namespaces.


ywk253100 (approved):

qiuming-best (approved):

sseago:
@Lyndon-Li Should we apply this to the release-1.11 branch as well (without the DataUpload/DataDownload parts)?

Lyndon-Li:
@sseago 
Yes, we can do this. But 1.11.1 has been tagged for RC yesterday, so we have to wait for 1.11.2 to include this fix.

PR #6533, opened 2023-07-21
Data mover wrong bsl after sync

### LINKED ISSUE #6534: Backup repo snapshot is not deleted after backup sync for file system backup
The PVB CR's StorageLocation is not reset to the latest one after backup sync, as a result, when deleting the PVB, the backup deletion controller cannot delete the repo snapshot from the correct backup storage location, so there will be snapshot data leak in the repo

Lyndon-Li:
This PR fixes two issues:
1. The PVB CR's `StorageLocation` is not reset to the latest one after backup sync, as a result, when deleting the PVB, the backup deletion controller cannot delete the repo snapshot from the correct backup storage location, so there will be data leak in the repo
2. During data mover restore, the DataUploadResult's `StorageLocation` is retrieved from the DataUpload CR, while the value in DataUpload CR is not up to date if the BSL's name has changed in the new cluster, as a result, the data mover restore will fail

As the fixes:

- For 1, reset PVB CR's `StorageLocation` to the latest one during backup sync
- For 2, retrieve DataUploadResult's `StorageLocation` from the latest backup CR whose `StorageLocation` has been reset during backup sync

Fix #6534

### DIFF
--- a/pkg/restore/dataupload_retrieve_action.go
+++ b/pkg/restore/dataupload_retrieve_action.go
@@ -25,7 +25,8 @@ import (
 	corev1api "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	velerov2alpha1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v2alpha1"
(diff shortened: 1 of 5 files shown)


pkg/restore/dataupload_retrieve_action.go
```diff
@@ -29,19 +29,22 @@ import (
 
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	velerov2alpha1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v2alpha1"
+	velerov1client "github.com/vmware-tanzu/velero/pkg/generated/clientset/versioned/typed/velero/v1"
```
ywk253100, line 32: Let's use the controller runtime client instead as we are moving away from the generated clients

Lyndon-Li: Done the changes, thanks.

ywk253100 (approved):

blackpiglet (approved):

PR #6085, opened 2023-04-05
Make build-image Dockerfile multi-platform

kaovilai:
Fix #6084

Signed-off-by: Tiger Kaovilai <tkaovila@redhat.com>

Thank you for contributing to Velero!

# Please add a summary of your change

# Does your change fix a particular issue?

Fixes #(issue)

# Please indicate you've done the following:

- [x] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [x] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required` as a comment on this pull request.
- [ ] Updated the corresponding documentation in `site/content/docs/main`.

PR #6556, opened 2023-07-28
Allow for using podman or docker based on what is installed

shawn-hurley:
* Adds ability for mac os builds. The GOARM is not used on linux and was not set, Adding 7 as that is what is required by Darwin.

Fixes #(issue)

/kind changelog-not-required

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,8 +38,13 @@ WORKDIR /go/src/github.com/vmware-tanzu/velero
 
 COPY . /go/src/github.com/vmware-tanzu/velero
 
-RUN mkdir -p /output/usr/bin && \
-    export GOARM=$( echo "${GOARM}" | cut -c2-) && \
+# if TARGETVARIANT is v8, set GOARM to empty string
+RUN if [ "${TARGETVARIANT}" = "v8" ]; then \
+        export GOARM=""; \
+else \
+    export GOARM=$( echo "${GOARM}" | cut -c2-); \
+fi && \
+    mkdir -p /output/usr/bin && \
     go build -o /output/${BIN} \
     -ldflags "${LDFLAGS}" ${PKG}/cmd/${BIN}
 
@@ -52,7 +57,7 @@ ARG TARGETARCH
 ARG TARGETVARIANT
 ARG RESTIC_VERSION
 
-env CGO_ENABLED=0 \
+ENV CGO_ENABLED=0 \
     GO111MODULE=on \
     GOPROXY=${GOPROXY} \
     GOOS=${TARGETOS} \
@@ -61,8 +66,13 @@ env CGO_ENABLED=0 \
 
 COPY . /go/src/github.com/vmware-tanzu/velero
 
-RUN mkdir -p /output/usr/bin && \
-    export GOARM=$(echo "${GOARM}" | cut -c2-) && \
+# if TARGETVARIANT is v8, set GOARM to empty string
+RUN if [ "${TARGETVARIANT}" = "v8" ]; then \
+    export GOARM=""; \
+else \
+    export GOARM=$( echo "${GOARM}" | cut -c2-); \
+fi && \
+    mkdir -p /output/usr/bin && \
     /go/src/github.com/vmware-tanzu/velero/hack/build-restic.sh
 
 # Velero image packing section
--- a/Makefile
+++ b/Makefile
@@ -51,8 +51,18 @@ else
 	BUILDER_IMAGE_TAG := $(shell git log -1 --pretty=%h $(BUILDER_IMAGE_DOCKERFILE))
 endif
 
+ifeq ($(shell command -v podman 2> /dev/null),)
+    BUILDER=docker buildx
+    IMAGE_TOOL=docker
+    BIND_MOUNT_SUFFIX=delegated
+else
+    BUILDER=podman
+    IMAGE_TOOL=podman
+    BIND_MOUNT_SUFFIX=z
+endif
+
 BUILDER_IMAGE := $(REGISTRY)/build-image:$(BUILDER_IMAGE_TAG)
-BUILDER_IMAGE_CACHED := $(shell docker images -q ${BUILDER_IMAGE} 2>/dev/null )
+BUILDER_IMAGE_CACHED := $(shell ${IMAGE_TOOL} images -q ${BUILDER_IMAGE} 2>/dev/null )
 
 HUGO_IMAGE := hugo-builder
 
@@ -79,6 +89,14 @@ ifeq ($(shell docker buildx inspect 2>/dev/null | awk '/Status/ { print $$2 }'),
 else
 	BUILDX_ENABLED ?= false
 endif
+
+# if BUILDX_ENABLED is true, check if the builder is docker buildx, if overridden to podman, disable buildx usage.
+ifeq ($(BUILDX_ENABLED), true)
+ifneq ($(BUILDER), docker buildx)
+	BUILDX_ENABLED := false
+endif
+endif
+
 
 define BUILDX_ERROR
 buildx not enabled, refusing to run this recipe
@@ -162,30 +180,24 @@ shell: build-dirs build-env
 	@# because the Kubernetes code-generator tools require the project to
 	@# exist in a directory hierarchy ending like this (but *NOT* necessarily
 	@# under $GOPATH).
-	@docker run \
+	${IMAGE_TOOL} run \
 		-e GOFLAGS \
 		-e GOPROXY \
 		-i $(TTY) \
 		--rm \
-		-u $$(id -u):$$(id -g) \
-		-v "$$(pwd):/github.com/vmware-tanzu/velero:delegated" \
-		-v "$$(pwd)/_output/bin:/output:delegated" \
-		-v "$$(pwd)/.go/pkg:/go/pkg:delegated" \
-		-v "$$(pwd)/.go/std:/go/std:delegated" \
-		-v "$$(pwd)/.go/std/$(GOOS)/$(GOARCH):/usr/local/go/pkg/$(GOOS)_$(GOARCH)_static:delegated" \
-		-v "$$(pwd)/.go/go-build:/.cache/go-build:delegated" \
-		-v "$$(pwd)/.go/golangci-lint:/.cache/golangci-lint:delegated" \
+		-u $(id -u):$(id -g) \
+		-v "$$(pwd):/github.com/vmware-tanzu/velero:${BIND_MOUNT_SUFFIX}" \
+		-v "$$(pwd)/_output/bin:/output:${BIND_MOUNT_SUFFIX}" \
+		-v "$$(pwd)/.go/pkg:/go/pkg:${BIND_MOUNT_SUFFIX}" \
+		-v "$$(pwd)/.go/std:/go/std:${BIND_MOUNT_SUFFIX}" \
+		-v "$$(pwd)/.go/std/$(GOOS)/$(GOARCH):/usr/local/go/pkg/$(GOOS)_$(GOARCH)_static:${BIND_MOUNT_SUFFIX}" \
+		-v "$$(pwd)/.go/go-build:/.cache/go-build:${BIND_MOUNT_SUFFIX}" \
+		-v "$$(pwd)/.go/golangci-lint:/.cache/golangci-lint:${BIND_MOUNT_SUFFIX}" \
 		-w /github.com/vmware-tanzu/velero \
 		$(BUILDER_IMAGE) \
 		/bin/sh $(CMD)
 
-container:
-ifneq ($(BUILDX_ENABLED), true)
-	$(error $(BUILDX_ERROR))
-endif
-	@docker buildx build --pull \
-	--output=type=$(BUILDX_OUTPUT_TYPE) \
-	--platform $(BUILDX_PLATFORMS) \
+container: BUILDARGS=--platform $(BUILDX_PLATFORMS) \
 	$(addprefix -t , $(IMAGE_TAGS)) \
 	$(addprefix -t , $(GCR_IMAGE_TAGS)) \
 	--build-arg=GOPROXY=$(GOPROXY) \
@@ -197,11 +209,20 @@ endif
 	--build-arg=REGISTRY=$(REGISTRY) \
 	--build-arg=RESTIC_VERSION=$(RESTIC_VERSION) \
 	-f $(VELERO_DOCKERFILE) .
+container:
+	echo $(BUILDARGS)
+ifeq ($(BUILDX_ENABLED), true)
+	${BUILDER} build --pull \
+	$(BUILDARGS)
+else
+	${BUILDER} build --pull \
+	$(BUILDARGS)
+endif
 	@echo "container: $(IMAGE):$(VERSION)"
 ifeq ($(BUILDX_OUTPUT_TYPE)_$(REGISTRY), registry_velero)
-	docker pull $(IMAGE):$(VERSION)
+	${IMAGE_TOOL} pull $(IMAGE):$(VERSION)
 	rm -f $(BIN)-$(VERSION).tar
-	docker save $(IMAGE):$(VERSION) -o $(BIN)-$(VERSION).tar
+	${IMAGE_TOOL} save $(IMAGE):$(VERSION) -o $(BIN)-$(VERSION).tar
 	gzip -f $(BIN)-$(VERSION).tar
 endif
 
@@ -260,7 +281,7 @@ else ifneq ($(BUILDER_IMAGE_CACHED),)
 	@echo "Using Cached Image: $(BUILDER_IMAGE)"
 else
 	@echo "Trying to pull build-image: $(BUILDER_IMAGE)"
-	docker pull -q $(BUILDER_IMAGE) || $(MAKE) build-image
+	${IMAGE_TOOL} pull -q $(BUILDER_IMAGE) || $(MAKE) build-image
 endif
 
 build-image:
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -199,9 +202,9 @@ endif
 	-f $(VELERO_DOCKERFILE) .
 	@echo "container: $(IMAGE):$(VERSION)"
 ifeq ($(BUILDX_OUTPUT_TYPE)_$(REGISTRY), registry_velero)
-	docker pull $(IMAGE):$(VERSION)
+	${BUILDER} pull $(IMAGE):$(VERSION)
 	rm -f $(BIN)-$(VERSION).tar
-	docker save $(IMAGE):$(VERSION) -o $(BIN)-$(VERSION).tar
+	BUILDER save $(IMAGE):$(VERSION) -o $(BIN)-$(VERSION).tar
```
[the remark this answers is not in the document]
shawn-hurley: +100 moving to fast!

Makefile
```diff
@@ -285,18 +288,18 @@ ifneq "$(origin BUILDER_IMAGE_DOCKERFILE)" "file"
...
-	cd site && docker build --pull -t $(HUGO_IMAGE) .
+	cd site && ${BUILDER} build --pull -t $(HUGO_IMAGE) .
 
 clean:
 # if we have a cached image then use it to run go clean --modcache
 # this test checks if we there is an image id in the BUILDER_IMAGE_CACHED variable.
 ifneq ($(strip $(BUILDER_IMAGE_CACHED)),)
 	$(MAKE) shell CMD="-c 'go clean --modcache'"
-	docker rmi -f $(BUILDER_IMAGE) || true
+	${BUILDER} rmi -f $(BUILDER_IMAGE) || true
```
mateusoliveira43, line 302: `buildx` has `rmi`? I think its only `rm`

Makefile
```diff
@@ -260,17 +263,17 @@ else ifneq ($(BUILDER_IMAGE_CACHED),)
 	@echo "Using Cached Image: $(BUILDER_IMAGE)"
 else
 	@echo "Trying to pull build-image: $(BUILDER_IMAGE)"
-	docker pull -q $(BUILDER_IMAGE) || $(MAKE) build-image
+	$BUILDER pull -q $(BUILDER_IMAGE) || $(MAKE) build-image
```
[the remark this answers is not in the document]
mateusoliveira43: `buildx` has `pull` command?

kaovilai:
Also to not break https://github.dev/vmware-tanzu/velero/blob/main/hack/docker-push.sh used in CI workflow, make sure that `BUILDX_PLATFORMS="linux/amd64,linux/arm64,linux/arm/v7,linux/ppc64le" make all-containers` runs

kaovilai:
referring to https://github.dev/vmware-tanzu/velero/blob/main/hack/docker-push.sh workflow

Makefile
shawn-hurley: That is correct! sorry! I will create an IMAGE_TOOL var to decide if just podman or docker to fix this

mateusoliveira43: better to set BUILDER just as `docker`, instead of `docker buildx` (if so, better to change variable name?), to change less things?

shawn-hurley: have to use docker buildx when building though :(

mateusoliveira43: I think I am a little lost. You mean whenever `bocker build` is called, it should be `docker buildx build`? If yes, then this seems like the only place missing https://github.com/vmware-tanzu/velero/blob/main/Makefile#L292

also makefile has a buildx logic https://github.com/vmware-tanzu/velero/blob/main/Makefile#L77-L81

shawn-hurley: @mateusoliveira43 I think I am confused as well.

The change I am making, so that we can use podman as well as docker. docker has a sub-command `buildx` that allows for creating of manifest list images. Podman is able to to do this without that sub command. 

When building and using docker, one should use `docker buildx build` and when using podman one should use `podman build`. 

When doing other image related tasks you just use `docker` or `podman`. 

Does that help or make the convo more confusing?

shawn-hurley:
@kaovilai tested and `make all-containers` works as expected.

Makefile
mateusoliveira43: Alright, got it, thank you.

So, I believe we can simplify this as `buildx` logic is already handled by makefile

we add the new logic to check docker or podman and inside it we need to run the `buildx` logic https://github.com/vmware-tanzu/velero/blob/main/Makefile#L77-L81 (for podman is always false) then just change the `docker` invocations to `podman` (we would also need to change this logic https://github.com/vmware-tanzu/velero/blob/main/Makefile#L183 to not fail to podman, but should be easy)

Do you agree?

shawn-hurley: I think that the current logic is pretty simple no and I don't know if I follow this. Can you submit a PR to my branch if you feel strongly it is better so that I can see it?

Dockerfile
```diff
@@ -31,15 +31,14 @@ ENV CGO_ENABLED=0 \
     GOPROXY=${GOPROXY} \
     GOOS=${TARGETOS} \
     GOARCH=${TARGETARCH} \
-    GOARM=${TARGETVARIANT} \
+    GOARM=7 \
```
kaovilai, line 34: Is this change necessary?

What about arm32v6 arm32v7 arm64v8?

shawn-hurley:
Considering that they were not being used in any way before, this is needed to only run on macOS AFAICT and according to the docs Darwin requires 7.On Jul 28, 2023, at 11:44 AM, Tiger Kaovilai ***@***.***> wrote:﻿
@kaovilai commented on this pull request.



In Dockerfile:
> @@ -31,15 +31,14 @@ ENV CGO_ENABLED=0 \
     GOPROXY=${GOPROXY} \
     GOOS=${TARGETOS} \
     GOARCH=${TARGETARCH} \
-    GOARM=${TARGETVARIANT} \
+    GOARM=7 \

Is this change necessary?
What about arm32v6 arm32v7 arm64v8?

kaovilai:
> Considering that they were not being used in any way before

They are automatic variables filled in by "builders" from targetplatform variable.

https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

So they are being populated and passed to `go build`.

specifying a different `--platform` can affects value of TARGETVARIANT and so GOARM value.

kaovilai:
[Valid values are 5, 6, 7.](https://pkg.go.dev/cmd/go#hdr-Environment_variables:~:text=GOARM%0A%09For%20GOARCH%3Darm%2C%20the%20ARM%20architecture%20for%20which%20to%20compile.%0A%09Valid%20values%20are%205%2C%206%2C%207.)

kaovilai:
ie. this is not just a mac thing. GOARM is used by rPi, among other linux machines on arm.

kaovilai:
> ## Go compile for Pi
> 
> If you want speed of compilation from another machine, you can compile for Pi with options below.
> 
> From [article](https://www.thepolyglotdeveloper.com/2017/04/cross-compiling-golang-applications-raspberry-pi/).
> 
> ```sh
> $ GOOS=linux GOARCH=arm GOARM=5 \
>   go build
> ```
[An example of why one would use GOARM](https://github.com/MichaelCurrin/dev-cheatsheets/blob/c56bb5828c9b407ce10c2da68a45b2fa1228ffa5/cheatsheets/maker-projects/raspberry-pi.md?plain=1#L25C1-L34)

That var should be left automated by TARGETVARIANT IMO.

Makefile
```diff
@@ -51,8 +51,16 @@ else
 	BUILDER_IMAGE_TAG := $(shell git log -1 --pretty=%h $(BUILDER_IMAGE_DOCKERFILE))
 endif
 
+ifeq ($(shell command -v docker 2> /dev/null),)
+    BUILDER=podman
+    IMAGE_TOOL=podamn
```
kaovilai, line 56:
```suggestion
    IMAGE_TOOL=podman
```

kaovilai (changes requested):
I also just tried to run this.. doesn't yet work.
```
❯ BUILDER=podman IMAGE_TOOL=podman make all-containers
podman build --pull \
	--output=type=docker \
	--platform linux/amd64 \
	-t velero/velero:main \
	-t gcr.io/velero-gcp/velero:main \
	--build-arg=GOPROXY=https://proxy.golang.org \
	--build-arg=PKG=github.com/vmware-tanzu/velero \
	--build-arg=BIN=velero \
	--build-arg=VERSION=main \
	--build-arg=GIT_SHA=d33c35cfb71ad93ee9ae01266cd2765115a26d0d \
	--build-arg=GIT_TREE_STATE=dirty \
	--build-arg=REGISTRY=velero \
	--build-arg=RESTIC_VERSION=0.15.0 \
	-f Dockerfile .
Error: '--output' option is not supported in remote mode
make[1]: *** [container] Error 125
make: *** [all-containers] Error 2
```

> The --output (or -o) option extends the default behavior of building a container image by allowing users to export the contents of the image as files on the local filesystem, which can be useful for generating local binaries, code generation, etc. (This option is not available with the remote Podman client, including Mac and Windows (excluding WSL2) machines)

[src](https://docs.podman.io/en/latest/markdown/podman-build.1.html#:~:text=The%20%2D%2Doutput%20(or,excluding%20WSL2)%20machines))

shawn-hurley:
Yeah, I realized that yesterday as well, looks like I have to determine if on Darwin/Windows and then remove the output flag.

If you just remove it, it does still create the image. I am kind of tempted to just remove it, unless someone knows why it is used because the image is output to disk and then tagged and pushed.

shawn-hurley:
Also re: raspberry pi, IIRC it does not appear that this in the list of builds that happen today. I could be wrong, but I don't know if I want to add every potential build. I am just trying to make it so that it is easier for me to develop. If there is a need to make it for raspberry pi or risc-v can someone else work on adding this? AFAICT this did not work before w/ podman on anything so some are better than others?

kaovilai:
rPi support can certainly be added later on. Just want to note that there are many rPis out there running kube.. with velero.

Here's one article https://picluster.ricsanfre.com/docs/backup/

kaovilai:
Also currently this is the build command that result
```
podman build --pull \
	--output=type=docker
```
`--output=type=docker` is only valid in under `docker buildx build` and would not be a valid option on Linux podman as well.

[Why it's being used?](https://docs.docker.com/engine/reference/commandline/buildx_build/#options:~:text=Load%20the%20single,docker%20images.)

> ## Load the single-platform build result to docker images (--load)
> Shorthand for [--output=type=docker](https://docs.docker.com/engine/reference/commandline/buildx_build/#docker). Will automatically load the single-platform build result to docker images.

So it looks like there's a way around this, outputting to tar then import back to images perhaps.

kaovilai:
GOARM usage also applies on non-rPi arm64 linux as well for optimization.

Makefile
```diff
@@ -260,21 +266,21 @@ else ifneq ($(BUILDER_IMAGE_CACHED),)
...
+	${IMAGE_TOOL} pull -q $(BUILDER_IMAGE) || $(MAKE) build-image
 endif
 
 build-image:
 	@# When we build a new image we just untag the old one.
 	@# This makes sure we don't leave the orphaned image behind.
 	$(eval old_id=$(shell docker image inspect  --format '{{ .ID }}' ${BUILDER_IMAGE} 2>/dev/null))
 ifeq ($(BUILDX_ENABLED), true)
-	@cd hack/build-image && docker buildx build --build-arg=GOPROXY=$(GOPROXY) --output=type=docker --pull -t $(BUILDER_IMAGE) -f $(BUILDER_IMAGE_DOCKERFILE_REALPATH) .
+	@cd hack/build-image && ${BUILDER} build --build-arg=GOPROXY=$(GOPROXY) --output=type=docker --pull -t $(BUILDER_IMAGE) -f $(BUILDER_IMAGE_DOCKERFILE_REALPATH) .
```
kaovilai, line 277: If buildx is enabled then use buildx, otherwise this is redundant with the fallback.
```suggestion
	@cd hack/build-image && ${BUILDER} buildx --build-arg=GOPROXY=$(GOPROXY) --output=type=docker --pull -t $(BUILDER_IMAGE) -f $(BUILDER_IMAGE_DOCKERFILE_REALPATH) .
```

kaovilai:
Ok so I did an echo inside dockerfile. The behavior of podman is such that if it's running on armv8, TARGETVARIANT is outputted as v8. In golang, the expected value for GOARM on ARMv8 is an empty string. Easy fix.

kaovilai:
Please review https://github.com/shawn-hurley/velero/pull/1

Dockerfile
ywk253100: Agree with @kaovilai , let's make it configurable

Makefile
ywk253100, line 56: How about renaming it to `CONTAINER_RUNTIME` as it does not only pull/push images?

Makefile
```diff
@@ -51,8 +51,16 @@ else
 	BUILDER_IMAGE_TAG := $(shell git log -1 --pretty=%h $(BUILDER_IMAGE_DOCKERFILE))
 endif
 
+ifeq ($(shell command -v docker 2> /dev/null),)
```
ywk253100, line 54: If both docker and podman are not installed, this will report an error like `podman: command not found`, this may confuse the users who don't know what podman is.
How about refactoring as following?
```
if docker installed, use docker
else if podman installed, use podman
else report error
```

kaovilai: I think podman command not found is self explanatory when you Google.

Makefile
```diff
@@ -260,21 +283,21 @@ else ifneq ($(BUILDER_IMAGE_CACHED),)
...
+	${IMAGE_TOOL} pull -q $(BUILDER_IMAGE) || $(MAKE) build-image
 endif
 
 build-image:
 	@# When we build a new image we just untag the old one.
 	@# This makes sure we don't leave the orphaned image behind.
 	$(eval old_id=$(shell docker image inspect  --format '{{ .ID }}' ${BUILDER_IMAGE} 2>/dev/null))
 ifeq ($(BUILDX_ENABLED), true)
-	@cd hack/build-image && docker buildx build --build-arg=GOPROXY=$(GOPROXY) --output=type=docker --pull -t $(BUILDER_IMAGE) -f $(BUILDER_IMAGE_DOCKERFILE_REALPATH) .
+	@cd hack/build-image && ${BUILDER} buildx build --build-arg=GOPROXY=$(GOPROXY) --output=type=docker --pull -t $(BUILDER_IMAGE) -f $(BUILDER_IMAGE_DOCKERFILE_REALPATH) .
```
kaovilai, line 294:
```suggestion
	@cd hack/build-image && ${BUILDER} build --build-arg=GOPROXY=$(GOPROXY) --output=type=docker --pull -t $(BUILDER_IMAGE) -f $(BUILDER_IMAGE_DOCKERFILE_REALPATH) .
```

Sorry typo.

Dockerfile
shawn-hurley: updated

Makefile
shawn-hurley: udpated

Makefile
shawn-hurley: What if we did it in the opposite direction? If Podman use it, if docker use it and then the error would be about docker.

sseago: @shawn-hurley Actually, that's probably best for an additional reason as well. If you have the podman-docker RPM installed, then the command `docker` exists, but it's really podman. If we don't check for podman first, then this change won't work for them.

shawn-hurley: +1 @sseago didn't think of that!

Makefile
kaovilai, line 294:
```suggestion
	@cd hack/build-image && ${BUILDER} build --build-arg=GOPROXY=$(GOPROXY) --output=type=docker --pull -t $(BUILDER_IMAGE) -f $(BUILDER_IMAGE_DOCKERFILE_REALPATH) .
```
Given that builder already have buildx, this extra buildx needs to be removed.
`    BUILDER=docker buildx`
That's what failing CI is running `docker buildx buildx build`
`docker buildx build` will not fail on `--build-args`

Dockerfile
```diff
@@ -38,8 +38,11 @@ WORKDIR /go/src/github.com/vmware-tanzu/velero
 
 COPY . /go/src/github.com/vmware-tanzu/velero
 
-RUN mkdir -p /output/usr/bin && \
-    export GOARM=$( echo "${GOARM}" | cut -c2-) && \
+# if TARGETVARIANT is v8, set GOARM to empty string
+RUN if [ "${TARGETVARIANT}" = "v8" ]; then \
+        export GOARM=""; \
```
mateusoliveira43, line 43: Should here have an `else export GOARM=$( echo "${GOARM}" | cut -c2-)`? (what it was doing prior to the change)

shawn-hurley: Adding it back now :) will getting a commit up with the changes soon!

Dockerfile
```diff
@@ -67,7 +69,9 @@ COPY . /go/src/github.com/vmware-tanzu/velero
 # if TARGETVARIANT is v8, set GOARM to empty string
 RUN if [ "${TARGETVARIANT}" = "v8" ]; then \
     export GOARM=""; \
-    fi && \
+else \
+    export GOARM=$( echo "${GOARM}" | cut -c2-); \
```
kaovilai, line 73: GOARM hasn't been set yet. cut from TARGETVARIANT.

```
    export GOARM=$( echo "${GOARM}" | cut -c2-); \
```

kaovilai (changes requested):

Dockerfile
shawn-hurley: added

kaovilai:
might need `RUN chmod -R 777 $(go env GOPATH)` in dockerfile at the end so that any processes that read/write there have write access instead of [erroring](https://github.com/vmware-tanzu/velero/actions/runs/5729224900/job/15525515985?pr=6556#step:5:282)?

shawn-hurley:
`make update` does not work on Mac OS because the builder image is not buildable with Podman. 

You can get around this by making the builder image multi-arch as well. The problem is the builder image is hardcoded to specific downloads for Arch and os. I think this will need to be changed.

This work is starting to take up too much of my time to finish right now. I believe that someone should do this work though. I will come back to it if someone does not take it over. 

Closing this for now.

kaovilai:
re: 

> builder image is not buildable with Podman.

does this help?
https://github.com/vmware-tanzu/velero/pull/6085

PR #6564, opened 2023-07-31
Fix issue 6561

### LINKED ISSUE #6561: alpine can not be pulled in some environments when moving snapshot
We use the `alpine:latest` as the image for the pod in data mover, however, it cannot be pulled in some environments, e.g. airgap env or env reaches the pull limit of Dockerhub.

Besides, we also need to consider the maintenance effort for the new image, e.g. CVE 

**Vote on this issue!**

Lyndon-Li:
Fix #6561, change the backup/restore pod's image to the same image used by the node-agent, but with a different binary `velero-helper`, so as not to depend on Alpine which may have compliance or security issues or image pull issues.

### DIFF
--- a/cmd/velero-helper/velero-helper.go
+++ b/cmd/velero-helper/velero-helper.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"time"
+)
+
+const (
+	// workingModePause indicates it is for general purpose to hold the pod under running state
+	workingModePause = "pause"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Fprintln(os.Stderr, "ERROR: at least one argument must be provided, the working mode")
+		os.Exit(1)
+	}
+
+	switch os.Args[1] {
+	case workingModePause:
+		time.Sleep(time.Duration(1<<63 - 1))
+	default:
+		fmt.Fprintln(os.Stderr, "ERROR: wrong working mode provided")
+		os.Exit(1)
+	}
+}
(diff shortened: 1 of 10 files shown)


pkg/exposer/csi_snapshot.go
```diff
@@ -366,9 +366,8 @@ func (e *csiSnapshotExposer) createBackupPod(ctx context.Context, ownerObject co
 			Containers: []corev1.Container{
 				{
 					Name:            podName,
-					Image:           "alpine:latest",
+					Image:           "gcr.io/google_containers/pause:latest",
```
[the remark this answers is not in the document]
Lyndon-Li: No special reason to use `latest`. Merely, we don't have anything depending on the image, so use the `latest` to have it up-to-date. (we only require an image that is as small in size as possible, is widely available and is with less security issues)
Additionally, it doesn't matter actually. This image is very simple and hasn't been updated for years. It is not expected to by changed.

sseago (approved):

cmd/velero-helper/velero-helper.go
mateusoliveira43, line 1: would be interesting to use cobra here? it would be more verbose, but we would have a CLI by default (and it also have the 2 validations builtin: `cobra.MinimumNArgs(1)` and `cobra.OnlyValidArgs`

Lyndon-Li:
We have more inputs/requirements today, so changed the solution from end to end and see if this can meet them:

- Use Velero image for data mover pods
- Create a binary `velero-helper` with small size into Velero image
- Data mover pods launches  `velero-helper` instead of velero, so that data mover pods memory footprint is small
- Duplicate all the image options(including the image itself) from node-agent so that data mover pods never need to pull images themselves, so as to eliminate the complication for image pulling

cmd/velero-helper/velero-helper.go
Lyndon-Li: I am hesitating to involve cobra:

- The binary is targeted to implement simply tasks (otherwise, we can use velero binary instead), i.e., the current pause functionality, it is not expected to be with complex parameters, for which cobra could make a big help.
- Involving cobra will involve some more packages, which will make the total binary size larger.

reasonerjt (approved):
lgtm

cmd/velero-helper/velero-helper.go
reasonerjt: We may even make it simpler to make it velero-pause, b/c I don't think we wanna support other subcommand, it's simply a placeholder for DM.

blackpiglet (approved):

ywk253100 (approved):

Issue #6548, opened 2023-07-25
Support backup on volumeMode block via Data mover (Kopia)


shubham-pampattiwar:
**Describe the problem/challenge you have**

Currently Velero Datamover does not support backup for PVs whose volumeMode is `Block`.

**Describe the solution you'd like**

Make Velero Datamover Backup and restore operations work for PVs with `volumeMode` as `Block`

**Anything else you would like to add:**

**Environment:**

- Velero version (use `velero version`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

draghuram:
@shubham-pampattiwar Considering that FSB and data mover backup work differently, it would be better to split this into two separate issues.

dzaninovic:
I am looking into what is needed to support Block PVCs for data mover path so I tried to run a backup to see what will fail.

PVC provisioned from the snapshot is Pending and the error message is “failed pre-populate data from snapshot d7f830c4-280c-11ee-8a47-0242ac110003: exit status 1: tar: invalid magic”.
volumesnapshot and volumesnapshotcontent are ready.
```
$ kubectl get volumesnapshot -A
NAMESPACE   NAME            READYTOUSE   SOURCEPVC   SOURCESNAPSHOTCONTENT   RESTORESIZE   SNAPSHOTCLASS            SNAPSHOTCONTENT   CREATIONTIME   AGE
velero      backup3-q5b6p   true                     backup3-q5b6p           1Gi           csi-hostpath-snapclass   backup3-q5b6p     27m            27m

$ kubectl get volumesnapshotcontent -A
NAME            READYTOUSE   RESTORESIZE   DELETIONPOLICY   DRIVER                VOLUMESNAPSHOTCLASS      VOLUMESNAPSHOT   VOLUMESNAPSHOTNAMESPACE   AGE
[...]

PR #6590, opened 2023-08-02
Adding design updates to handle block device with kopia

shawn-hurley:

This will add the design elements for handling block device's. We are making a small change to the VGDP provider interface so that the data type can be based on and used by data movers. 

I don't believe this is an API change that is user-facing or even plugin-facing. Can someone help me make sure?

### DIFF
--- a/design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
+++ b/design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
@@ -31,7 +31,7 @@ Moreover, we would like to create a general workflow to variations during the da
 
 ## Non-Goals
 
-- The implementation details of block level access will be included in a future design
+- The current support for block level access is through file system uploader, so it is not aimed to deliver features of an ultimate block level backup. Block level backup will be included in a future design
 - Most of the components are generic, but the Exposer is snapshot type specific or snapshot access specific. The current design covers the implementation details for exposing CSI snapshot to host path access only, for other types or accesses, we may need a separate design
 - The current workflow focuses on snapshot-based data movements. For some application/SaaS level data sources, snapshots may not be taken explicitly. We don’t take them into consideration, though we believe that some workflows or components may still be reusable.
 
@@ -67,7 +67,7 @@ DMs take the responsibility to handle DUCR/DDCRs, Velero provides a built-in DM 
 **Node-Agent**: Node-Agent is an existing Velero module that will be used to host VBDM.  
 **Exposer**: Exposer is to expose the snapshot/target volume as a path/device name/endpoint that are recognizable by Velero generic data path. For different snapshot types/snapshot accesses, the Exposer may be different. This isolation guarantees that when we want to support other snapshot types/snapshot accesses, we only need to replace with a new Exposer and keep other components as is.  
 **Velero Generic Data Path (VGDP)**: VGDP is the collective of modules that is introduced in [Unified Repository design][1]. Velero uses these modules to finish data transmission for various purposes. In includes uploaders and the backup repository.  
-**Uploader**: Uploader is the module in VGDP that reads data from the source and writes to backup repository for backup; while read data from backup repository and write to the restore target for restore. At present, only file system uploader is supported. In future, the block level uploader will be added. For file system uploader, only Kopia uploader will be used, Restic will not be integrated with VBDM.   
+**Uploader**: Uploader is the module in VGDP that reads data from the source and writes to backup repository for backup; while read data from backup repository and write to the restore target for restore. At present, only file system uploader is supported. In future, the block level uploader will be added. For file system and basic block uploader, only Kopia uploader will be used, Restic will not be integrated with VBDM.   
 
 ## Replacement
 3rd parties could integrate their own data movement into Velero by replacing VBDM with their own DMs. The DMs should process DUCR/DDCRs appropriately and finally put them into one of the terminal states as shown in the DataUpload CRD and DataDownload CRD sections.  
@@ -82,8 +82,8 @@ Below are the data movement actions and sequences during backup:
 Below are actions from Velero and DMP:
 
 **BIA Execute**  
-This the existing logic in Velero. For a source PVC/PV, Velero delivers it to the corresponding BackupItemAction plugin, the plugin then takes the related actions to back it up.  
-For example, the existing CSI plugin takes a CSI snapshot to the volume represented by the PVC and then returns additional items (i.e., VolumeSnapshot, VolumeSnapshotContent and VolumeSnapshotClass) for Velero to further backup.  
+This is the existing logic in Velero. For a source PVC/PV, Velero delivers it to the corresponding BackupItemAction plugin, the plugin then takes the related actions to back it up.
+For example, the existing CSI plugin takes a CSI snapshot to the volume represented by the PVC and then returns additional items (i.e., VolumeSnapshot, VolumeSnapshotContent and VolumeSnapshotClass) for Velero to further backup.
 To support data movement, we will use BIA V2 which supports asynchronized operation management. Here is the Execute method from BIA V2:  
 ```
 Execute(item runtime.Unstructured, backup *api.Backup) (runtime.Unstructured, []velero.ResourceIdentifier, string, []velero.ResourceIdentifier, error)
@@ -115,7 +115,7 @@ Progress(operationID string, backup *api.Backup) (velero.OperationProgress, erro
 On the call of this method, DMP will query the DUCR’s status. Some critical progress data is transferred from DUCR to the ```OperationProgress``` which is the return value of BIA V2’s Progress method. For example, NCompleted indicates the size/number of data that have been completed and NTotal indicates the total size/number of data.  
 When the async operation completes, the Progress method returns an OperationProgress with ```Completed``` set as true. Then Velero will persist DUCR as well as any other items returned by DUP as ```itemToUpdate```.  
 Finally, then backup is as ```Completed```.  
-To help BIA Progress find the corresponding DUCR, the ```operationID``` is saved along with the DUCR as a label ```velero.io/async-operation-id```.  
+To help BIA Progress find the corresponding DUCR, the ```operationID``` is saved along with the DUCR as a label ```velero.io/async-operation-id```.
 
 DUCRs are handled by the data movers, so how to handle them are totally decided by the data movers. Below covers the details of VBDM, plugging data movers should have their own actions and workflows.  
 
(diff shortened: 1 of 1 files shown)


design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -67,7 +66,7 @@ DMs take the responsibility to handle DUCR/DDCRs, Velero provides a built-in DM
 **Node-Agent**: Node-Agent is an existing Velero module that will be used to host VBDM.  
 **Exposer**: Exposer is to expose the snapshot/target volume as a path/device name/endpoint that are recognizable by Velero generic data path. For different snapshot types/snapshot accesses, the Exposer may be different. This isolation guarantees that when we want to support other snapshot types/snapshot accesses, we only need to replace with a new Exposer and keep other components as is.  
 **Velero Generic Data Path (VGDP)**: VGDP is the collective of modules that is introduced in [Unified Repository design][1]. Velero uses these modules to finish data transmission for various purposes. In includes uploaders and the backup repository.  
-**Uploader**: Uploader is the module in VGDP that reads data from the source and writes to backup repository for backup; while read data from backup repository and write to the restore target for restore. At present, only file system uploader is supported. In future, the block level uploader will be added. For file system uploader, only Kopia uploader will be used, Restic will not be integrated with VBDM.
```
Lyndon-Li, line 70: We don't need to remove this:
`In future, the block level uploader will be added.`

A block level uploader is required for implementing the block level backup as suggested above for the non-goal. Otherwise, it will be confusing to have a dedicate Uploader layer in the architecture if there is only one uploader.

shawn-hurley: This makes sense, as I think that something to handle changed blocks needs to be in that interface and this one won't ever have that ability. +1

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -663,6 +662,32 @@ If DM restarts, Velero has no way to detect this, DM is expected to:
...
+		forceFull bool,
+		parentSnapshot string,
+		updater uploader.ProgressUpdater) (string, bool, error)
+```
+
+In this case, we will extend the default kopia uploader to add the ability, when a given volume is for a block mode and is mapped as a device, we will use a new specific type that conforms to the [File](https://pkg.go.dev/github.com/kopia/kopia/fs@v0.13.0#File) interface. This new type, will handle streaming the block device to the rest of the kopia process.
+
+While restoring, we are able extend the [Output](https://pkg.go.dev/github.com/kopia/kopia@v0.13.0/snapshot/restore#Output) interface. This will allow us to handle the case where we are writing to the device
+
+While extending these interfaces, we will use the defaults except for when it is for a block device.
```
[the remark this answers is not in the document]
shawn-hurley: I added the code on how it will extend the interfaces please let me know if this works. For now resolving this convo but we can re-open or start a new one if we need to.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -663,6 +662,32 @@ If DM restarts, Velero has no way to detect this, DM is expected to:
...
+
+```go
+// Provider which is designed for one pod volume to do the backup or restore
+type Provider interface {
+	// RunBackup which will do backup for one specific volume and return snapshotID, isSnapshotEmpty, error
+	// updater is used for updating backup progress which implement by third-party
+	RunBackup(
+		ctx context.Context,
+		path string,
+		volMode v1.PersistentVolumeMode,
```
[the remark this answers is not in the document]
shawn-hurley: updated

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
shawn-hurley: updated

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -31,7 +31,6 @@ Moreover, we would like to create a general workflow to variations during the da
 
 ## Non-Goals
 
-- The implementation details of block level access will be included in a future design
```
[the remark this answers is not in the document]
shawn-hurley: updated

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -31,7 +31,7 @@ Moreover, we would like to create a general workflow to variations during the da
...
+- The current support for block level access is through file system uploader, so it is not aimed to deliver features of an ultimate block level backup. Block level backup will be included in a future design
```
shubham-pampattiwar, line 34: +1

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -82,8 +82,8 @@ Below are the data movement actions and sequences during backup:
 Below are actions from Velero and DMP:
 
 **BIA Execute**  
-This the existing logic in Velero. For a source PVC/PV, Velero delivers it to the corresponding BackupItemAction plugin, the plugin then takes the related actions to back it up.  
-For example, the existing CSI plugin takes a CSI snapshot to the volume represented by the PVC and then returns additional items (i.e., VolumeSnapshot, VolumeSnapshotContent and VolumeSnapshotClass) for Velero to further backup.  
+This the existing logic in Velero. For a source PVC/PV, Velero delivers it to the corresponding BackupItemAction plugin, the plugin then takes the related actions to back it up.
```
shubham-pampattiwar, line 85: NIT: s/This/This is

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -663,6 +663,196 @@ If DM restarts, Velero has no way to detect this, DM is expected to:
...
+		parentSnapshot string,
+		updater uploader.ProgressUpdater) (string, bool, error)
+```
+
+In this case, we will extend the default kopia uploader to add the ability, when a given volume is for a block mode and is mapped as a device, we will use a new specific type that conforms to the [File](https://pkg.go.dev/github.com/kopia/kopia/fs@v0.13.0#File) interface. This new type, will handle streaming the block device to the rest of the kopia process.
+
+For this particular one, we will create a new `deviceEntry` and `deviceReader` structs that conform to the [fs.Reader](https://pkg.go.dev/github.com/kopia/kopia/fs@v0.13.0#Reader) and [File](https://pkg.go.dev/github.com/kopia/kopia/fs@v0.13.0#File) interfaces. 
+
+```go
+type deviceEntry struct {
```
Lyndon-Li, line 701: I didn't try it, but I am strongly feeling that the current requirement done by hacking `fs.File` and `fs.Reader` could be fulfilled by Kopia upstream's virtualfs directly. Specifically:

- It implements `fs.Entry` by `virtualEntry` which could replace `deviceEntry`
- It implements a `fs.StreamingFile` which is used similarly to `fs.File`, but it additionally allows the caller to provider an `io.ReadCloser`. Then I think it could replace `deviceReader`. Though it doesn't expose `io.Seek`, I think we don't need it, because with the current approach of block level access, we read the blocks sequentially from beginning to end all the time, we don't need to skip any blocks

If we can make this, we will have the Kopia code integration more predictably, because we know the design purpose for the Kopia upstream's StreamingFile, it won't break us in future if we follow its design purpose.
While for `fs.File` and `fs.Reader`, they are for file system level backups,  we don't know how they are going to develop in future version of Kopia.

Moreover, this virtualfs could help us to solve any potential existing/future conflicts that saving "not-really-file-system-data" into a file system data originated repo.

dzaninovic:
I got the backup and restore to work in the implementation this design is based on:
https://github.com/catalogicsoftware/velero_block/pull/1

There is still a lot of work remaining to be done.

shawn-hurley:
@Lyndon-Li Please comments on PR where I attempt your proposal

https://github.com/catalogicsoftware/velero_block/pull/2

Lyndon-Li:
@shawn-hurley Let me comment on your changes in https://github.com/catalogicsoftware/velero_block/pull/2 here, so that we can make the discussion centrally.

Lyndon-Li:
**1. For the IO optimization, like DIO:**
As I mentioned [here](https://github.com/vmware-tanzu/velero/issues/6548#issuecomment-1653961234), the current Kopia file system uploader doesn't support this, so we need a completely separate implementation. 
For this implementation, we not only need to implement a reader but also a writer (because the writing also benefits from the DIO).
Moreover, in order to use DIO, the IO pattern will be different because the read & write have to be aligned with sector size.

For all these complications, a new uploader -- block uploader is more suitable.
Therefore, together with other requirements as listed [here](https://github.com/vmware-tanzu/velero/issues/6548#issuecomment-1654877621) , I categorized this to the future block level backup instead of the current Phase-0.

On the other hand, I am not saying that the current `virtualfs` implementation doesn't support these IO optimizations technically. As you can see, the `StreamingFileFromReader` accepts an `io.ReadCloser` interface, we can implement interface and open & read the block device through DIO.

Conclusively:
1. I don't suggest we have these IO optimizations in a patch release. Regarding to data consistency, there are more potential problems for block IO, so we need more time for testing & verification.
2. If we want to do it for Phase-0 in a minor release, we need to wrap the implementation well enough so that the future block level backup could reuse it

Lyndon-Li:
**2. For skipping zero bytes during restore:**
I don't think it is a good choice. 
Zero bytes are not always meaningless. Once a new volume/PV is provisioned, we cannot assure that the entire disk data are all-zero, storages don't clear the disk data all the time unless they are requested to do this, i.e., for strong security requirement (think about how long it takes to write all zero to a large disk). As a result, if in some area of the disk, the data should be zero, but we skip writing zero to the area and the area for the provisioned disk is not zero, data corruption will happen.

Lyndon-Li:
**3. check target file is a block device from OS level**
This is a good to have. We get the volume mode from Kubernetes first and then double check this in a different way from the uploader.
Moreover, I would suggest we wrap this check to a separate function, as it is useful for future modules/functionalities.

shawn-hurley:
@sseago @shubham-pampattiwar @Lyndon-Li @dzaninovic @weshayutin 

Quick ping to inform folks that the design is updated based on the conversations. Please take a look and let me know what y'all think!

kaovilai (approved):

weshayutin:
I've read through this design, which looks good and the interface change pr 6608 which is also looking good.

shubham-pampattiwar (approved):

sseago (approved):

Lyndon-Li:
The current design looks good to me.

Lyndon-Li (approved):

Lyndon-Li:
@dzaninovic 

> > **2. For skipping zero bytes during restore:** I don't think it is a good choice. Zero bytes are not always meaningless and
> 
> This can be a non-default option that user can set if they know that disks are provisioned with all zeroes.

Personally, I don't think we can use this as an optional method. It is too dangerous since all-zero provision is hard to config, detect and not all upper level scenarios direct to a new provision; and it is not a good bargain to detect zero data by per-byte enumeration.
Anyway, as mentioned above, detecting and applying zero bytes is an advanced feature, let's see what we can deliver in future when we have more mature backup, backup repo and restore.

dzaninovic:
I reverted https://github.com/catalogicsoftware/velero_block/pull/2 where virtualfs.StreamingFileFromReader() was attempted because it was causing a backup failure.

time="2023-08-09T21:31:47Z" level=error msg="Async fs backup data path failed" dataupload=backup1-cv4mr error="Failed to run kopia backup: Failed to upload the kopia snapshot for si default@default:block1/pvc-raw: **unsupported source**: default@default:block1/pvc-raw" error.file="/go/pkg/mod/github.com/kopia/kopia@v0.13.0/snapshot/snapshotfs/upload.go:1291" error.function="[github.com/kopia/kopia/snapshot/snapshotfs.(*Uploader).Upload](http://github.com/kopia/kopia/snapshot/snapshotfs.(*Uploader).Upload)" logSource="pkg/controller/data_upload_controller.go:328"

By looking at the Kopia code I can see that it fails because fs.StreamingFile is not considered in the switch statement.
https://github.com/kopia/kopia/blob/v0.13.0/snapshot/snapshotfs/upload.go#L1291

@shawn-hurley will attempt to use this instead so if that is successful and more preferable than my original solution we will use that:
https://github.com/kopia/kopia/blob/master/fs/virtualfs/virtualfs.go#L96

In any case design will have to change to account for this change so don't merge it yet.

shawn-hurley:
@Lyndon-Li @sseago @shubham-pampattiwar 

I have once again updated the design. We have to do some specific things during the restore to handle the block device and how things work from the node.

We still use the virtualfs to save the data into the kopia repository.

Please let me now what y'all think

@dzaninovic My new virtualfs PR should work for restore now!

dzaninovic:
> @dzaninovic My new virtualfs PR should work for restore now!

I merged your code and confirmed that md5 of restored data is correct.

sseago (approved):

Lyndon-Li:
@shawn-hurley 
I checked the code. The restore part looks fine as it is adding some more checks on the block device. Merely, I am just curious that whether the restore part code are must-have?

from your statement:

> do some specific things during the restore to handle the block device and how things work from the node

It looks like these code have fixed some problems, without of which the restore will fail. I didn't get what the problems are. So if it is true, could you share some more details?

shawn-hurley:
> It looks like these codes have fixed some problems, without which the restore will fail. I didn't get what the issues were. So if it is true, could you share some more details?

The most significant change is that the built-in output will try and create a directory where the block device is. This fails, so we check to make sure that the file is a block device, and if it is, we skip the creation of the directory because it exists.

Another change is the writing of the blocks. When using the default `FileSystemOutput.WriteFile`, was changing the data in some way (the SHA sums of the resulting data were inconsistent). I did not look into why because even if we understood why, we would need to either write the code that exists or we would be required to make some change upstream.

shubham-pampattiwar (approved):

Lyndon-Li (approved):

sseago:
"codecov/project — 60.18% (-0.01%) compared to bb96c21"
This failure isn't relevant for this PR since it only includes design documentation and no actual new code  that can be tested.

PR #6599, opened 2023-08-03
Add deleting phase to restore 

### LINKED ISSUE #6582: Update the status of Restore before removing the storage resources to indicate the actual status
The status of `Restore` isn't updated although I haved runed `velero restore delte` command:

```
velero restore get
NAME    BACKUP   STATUS      STARTED                         COMPLETED                       ERRORS   WARNINGS   CREATED                         SELECTOR
dm-15   dm-11    Completed   2023-08-01 15:31:42 +0800 CST   2023-08-01 15:31:56 +0800 CST   0        4          2023-08-01 15:31:42 +0800 CST   <none>
```
This is confusing, especially when the restore isn't deleted successfully due to some issues.

**Vote on this issue!**

allenxu404:

Fixes #6582 


sseago:
Hmm. I'm not sure we need a separate deleting phase. It might be better to just check for deletion timestamp in those places that, in this PR, check for deleting phase. The `velero restore get` CLI output could also display "Deleting" if deletion timestamp is present. That's normally what kubernetes does for this sort of situation. For example `kubectl get pods` shows the phase value in the status output, except when `DeletionTimestamp` is set, then it shows `Terminating`. This avoids needing a special phase for terminating, and avoids needing to make an update to a terminating resource (which could fail if the resource disappears before the Update call is made)

allenxu404:
> Hmm. I'm not sure we need a separate deleting phase. It might be better to just check for deletion timestamp in those places that, in this PR, check for deleting phase. The `velero restore get` CLI output could also display "Deleting" if deletion timestamp is present. That's normally what kubernetes does for this sort of situation. For example `kubectl get pods` shows the phase value in the status output, except when `DeletionTimestamp` is set, then it shows `Terminating`. This avoids needing a special phase for terminating, and avoids needing to make an update to a terminating resource (which could fail if the resource disappears before the Update call is made)

Make sense. I will submit another PR to implement it, taking into account this advice.

PR #6604, opened 2023-08-04
Add "Deleting" phase for restore display

### LINKED ISSUE #6582: Update the status of Restore before removing the storage resources to indicate the actual status

allenxu404:

Fixes #6582 

### DIFF
--- a/pkg/cmd/cli/restore/get.go
+++ b/pkg/cmd/cli/restore/get.go
@@ -59,6 +59,13 @@ func NewGetCommand(f client.Factory, use string) *cobra.Command {
 				cmd.CheckError(err)
 			}
 
+			// Append "(Deleting)" to phase if deletionTimestamp is marked.
+			for i := range restores.Items {
+				if !restores.Items[i].DeletionTimestamp.IsZero() {
+					restores.Items[i].Status.Phase += " (Deleting)"
+				}
+			}
+
 			if printed, err := output.PrintWithFormat(c, restores); printed || err != nil {
 				cmd.CheckError(err)
 				return
--- a/pkg/cmd/util/output/restore_describer.go
+++ b/pkg/cmd/util/output/restore_describer.go
@@ -45,6 +45,12 @@ func DescribeRestore(ctx context.Context, kbClient kbclient.Client, restore *vel
 			phase = velerov1api.RestorePhaseNew
 		}
 		phaseString := string(phase)
+
+		// Append "Deleting" to phaseString if deletionTimestamp is marked.
+		if !restore.DeletionTimestamp.IsZero() {
+			phaseString += " (Deleting)"
+		}
+
 		switch phase {
 		case velerov1api.RestorePhaseCompleted:
 			phaseString = color.GreenString(phaseString)


blackpiglet (dismissed):

ywk253100:
This will cause inconsistent between the Velero and kubectl when getting the restores, @sseago is there any way to  specify the status in CRD according to different conditions?

sseago:
And if you're looking at yaml, then you want to see the real `Phase` value, but again the deletion timestamp tells you it's terminating. As long as the velero CLI summary shows deleting/terminating and any code that differs for terminating restores checks deletion timestamp, this should be good.

sseago:
You probably want to check deletion timestamp on the logs command, as was covered in the original change-phase PR:

https://github.com/vmware-tanzu/velero/pull/6599/files#diff-aa4ed12cfdf5527ebf5441e9ea40d8cb40f9f4d446a5742d12e48e6bd281189e

pkg/cmd/cli/restore/describe.go
```diff
@@ -80,6 +80,11 @@ func NewDescribeCommand(f client.Factory, use string) *cobra.Command {
 					fmt.Fprintf(os.Stderr, "error getting PodVolumeRestores for restore %s: %v\n", restore.Name, err)
 				}
 
+				// Change phase to "Deleting" if deletionTimestamp is marked.
+				if !restoreList.Items[i].DeletionTimestamp.IsZero() {
+					restoreList.Items[i].Status.Phase = "Deleting"
```
sseago, line 85: I don't think this will work, since "Deleting" isn't a valid value for this field. A better place to do this would be here:
https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/util/output/restore_describer.go#L46

Also, I don't think we want to lose the actual value of the field (i.e. completed vs. failed), so maybe in restore_describe.go, something like this:
```
	if !restore.DeletionTimestamp.IsZero() {
		phase = phase + " (Deleting)"
	}
```

pkg/cmd/cli/restore/get.go
```diff
@@ -59,6 +59,13 @@ func NewGetCommand(f client.Factory, use string) *cobra.Command {
 				cmd.CheckError(err)
 			}
 
+			// Change phase to "Deleting" if deletionTimestamp is marked.
+			for i := range restores.Items {
+				if !restores.Items[i].DeletionTimestamp.IsZero() {
+					restores.Items[i].Status.Phase = "Deleting"
+				}
```
sseago, line 66: Similar with this section. Instead of here, lets modify restore_printer.go:
https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/util/output/restore_printer.go#L60
```
I don't think this will work, since "Deleting" isn't a valid value for this field. A better place to do this would be here:
https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/util/output/restore_describer.go#L46

Also, I don't think we want to lose the actual value of the field (i.e. completed vs. failed), so maybe in restore_describe.go, something like this:

	if !restore.DeletionTimestamp.IsZero() {
		status = "Deleting"
	}
```
Here we can just use Deleting rather than showing prior status since it's just a quick summary.

sseago (changes requested):

pkg/cmd/cli/restore/describe.go
allenxu404: Make sense.

pkg/cmd/cli/restore/get.go
sseago: @allenxu404 I guess the only real concern was that I wasn't sure that would work since you're setting Phase to something that's not defined as a `RestorePhase` const, but maybe that's fine since we're not actually *updating* the restore object in the cluster. If the compiler is happy with this and it's working for you there, then it should be fine.

ywk253100 (approved):

blackpiglet (approved):

sseago (approved):

PR #6608, opened 2023-08-04
Feature/add provider interface for block mode

shawn-hurley:

Adding the block mode to the interfaces, to be included in 1.12.

This will allow us to solve the issue of block support in a Z-stream without breaking the API's in a Z-stream. 

Fixes #(issue)

### DIFF
--- a/changelogs/unreleased/6608-shawn-hurley
+++ b/changelogs/unreleased/6608-shawn-hurley
@@ -1,0 +1,1 @@
+Add API support for volMode block, only error for now.
--- a/pkg/datapath/file_system.go
+++ b/pkg/datapath/file_system.go
@@ -133,9 +133,10 @@ func (fs *fileSystemBR) StartBackup(source AccessPoint, realSource string, paren
 	if !fs.initialized {
 		return errors.New("file system data path is not initialized")
 	}
+	volMode := getPersistentVolumeMode(source)
 
 	go func() {
-		snapshotID, emptySnapshot, err := fs.uploaderProv.RunBackup(fs.ctx, source.ByPath, realSource, tags, forceFull, parentSnapshot, fs)
+		snapshotID, emptySnapshot, err := fs.uploaderProv.RunBackup(fs.ctx, source.ByPath, realSource, tags, forceFull, parentSnapshot, volMode, fs)
 
 		if err == provider.ErrorCanceled {
 			fs.callbacks.OnCancelled(context.Background(), fs.namespace, fs.jobName)
@@ -154,8 +155,10 @@ func (fs *fileSystemBR) StartRestore(snapshotID string, target AccessPoint) erro
 		return errors.New("file system data path is not initialized")
 	}
 
+	volMode := getPersistentVolumeMode(target)
+
 	go func() {
-		err := fs.uploaderProv.RunRestore(fs.ctx, snapshotID, target.ByPath, fs)
+		err := fs.uploaderProv.RunRestore(fs.ctx, snapshotID, target.ByPath, volMode, fs)
 
 		if err == provider.ErrorCanceled {
 			fs.callbacks.OnCancelled(context.Background(), fs.namespace, fs.jobName)
@@ -167,6 +170,13 @@ func (fs *fileSystemBR) StartRestore(snapshotID string, target AccessPoint) erro
 	}()
 
 	return nil
+}
+
+func getPersistentVolumeMode(source AccessPoint) uploader.PersistentVolumeMode {
+	if source.ByBlock != "" {
+		return uploader.PersistentVolumeBlock
+	}
+	return uploader.PersistentVolumeFilesystem
 }
 
 // UpdateProgress which implement ProgressUpdater interface to update progress status
--- a/pkg/uploader/kopia/snapshot.go
+++ b/pkg/uploader/kopia/snapshot.go
@@ -88,10 +88,15 @@ func setupDefaultPolicy() *policy.Tree {
 
 // Backup backup specific sourcePath and update progress
 func Backup(ctx context.Context, fsUploader SnapshotUploader, repoWriter repo.RepositoryWriter, sourcePath string, realSource string,
-	forceFull bool, parentSnapshot string, tags map[string]string, log logrus.FieldLogger) (*uploader.SnapshotInfo, bool, error) {
+	forceFull bool, parentSnapshot string, volMode uploader.PersistentVolumeMode, tags map[string]string, log logrus.FieldLogger) (*uploader.SnapshotInfo, bool, error) {
 	if fsUploader == nil {
 		return nil, false, errors.New("get empty kopia uploader")
 	}
+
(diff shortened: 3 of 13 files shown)


weshayutin:
Thank you @shawn-hurley!  @draghuram @dzaninovic FYI :)

dzaninovic:
I will refactor my code to adjust for these changes.

sseago (approved):

sseago:
Note for other maintainers/reviewers: This PR just implements the API change, adding `volMode` to `RunBackup()`, but errors out as unsupported if Block is specified. The hope is to get the API change into 1.12 before anyone is actually using this API, but then we can add the kopia implementation bits (Restic will continue to error out on block mode) in a future release. Making the API change later increases the likelihood that it will break someone's implementation of it, but at the moment there are no implementations of this API outside of Velero core code.

dzaninovic:
We also need to add volMode to RunRestore():
https://github.com/catalogicsoftware/velero_block/pull/1/files#diff-0c7524c0c1d97816a32e51877ae12c8f8d2b24f8ab305388196efe7d2a539edaR59

changelogs/unreleased/6608-shawn-hurley
```diff
@@ -0,0 +1 @@
+Design update to handle block devices during the kopia data mover
```
shubham-pampattiwar, line 1: Update the change log to something like "Add API support for volMode block" or something similar" or "block mode mover impl part:1 "

pkg/uploader/provider/provider.go
```diff
@@ -48,6 +48,7 @@ type Provider interface {
 		tags map[string]string,
 		forceFull bool,
 		parentSnapshot string,
+		volMode uploader.PersistentVolumeMode,
 		updater uploader.ProgressUpdater) (string, bool, error)
 	// RunRestore which will do restore for one specific volume with given snapshot id and return error
 	// updater is used for updating backup progress which implement by third-party
```
shubham-pampattiwar, line 54: Need to pass `volMode uploader.PersistentVolumeMode` in `RunRestore as well

shawn-hurley:
> If so, what is the purpose to have this separate PR, why don't we submit the entire functioning code in one PR later when they are ready?

The biggest reason is that we want the API changes for the 1.12.0 release. This will make sure that we don't add a breaking API change for folks who are attempting to use the VGDP (I think that is the right acronym that I have been reading?) Please let me know if this makes sense. 

Also note, I will add code around when ByPath or ByBlock should be set.

weshayutin:
Thanks for the progress!

Lyndon-Li:
@shawn-hurley Please help to fix the CI errors. Thanks!

pkg/uploader/provider/provider.go
shawn-hurley: updated

changelogs/unreleased/6608-shawn-hurley
shawn-hurley: updated

pkg/datapath/file_system.go
```diff
@@ -133,9 +133,13 @@ func (fs *fileSystemBR) StartBackup(source AccessPoint, realSource string, paren
 	if !fs.initialized {
 		return errors.New("file system data path is not initialized")
 	}
+	volMode := uploader.PersistentVolumeFilesystem
```
mateusoliveira43, line 136: would it be interesting to have this in a function and use in both places?
```go
func getPersistentVolumeMode(source AccessPoint) PersistentVolumeMode {
    if source.ByBlock != "" {
	 return uploader.PersistentVolumeBlock
    }
    return uploader.PersistentVolumeFilesystem
}
```

shawn-hurley: I could see adding this as a helper!

pkg/uploader/kopia/snapshot.go
```diff
@@ -88,10 +88,15 @@ func setupDefaultPolicy() *policy.Tree {
...
 // Backup backup specific sourcePath and update progress
 func Backup(ctx context.Context, fsUploader SnapshotUploader, repoWriter repo.RepositoryWriter, sourcePath string, realSource string,
-	forceFull bool, parentSnapshot string, tags map[string]string, log logrus.FieldLogger) (*uploader.SnapshotInfo, bool, error) {
+	forceFull bool, parentSnapshot string, volMode uploader.PersistentVolumeMode, tags map[string]string, log logrus.FieldLogger) (*uploader.SnapshotInfo, bool, error) {
 	if fsUploader == nil {
 		return nil, false, errors.New("get empty kopia uploader")
 	}
+
+	if volMode == uploader.PersistentVolumeBlock {
+		return nil, false, errors.New("unable to handle block storage")
```
mateusoliveira43, line 97: Would it be interesting to create a `const` with the error message? (this one is different from the other ones, but we could standardize, right?)

And would be good to be more explicit in the message, like
**"This version of Velero does NOT support block mode, it will be implemented in the future"** ?
(my fear is users opening issues not understanding the problem, instead of just asking when this feature will be available)

sseago: @mateusoliveira43 Actually, your suggested wording change points out why we need different messages here. Kopia will add it in the future, but we will never support this in Restic.

shawn-hurley: Resolving because as Scott said we want them to be different

pkg/datapath/file_system.go
shawn-hurley: Updated and added as helper

shawn-hurley:
@Lyndon-Li @sseago @shubham-pampattiwar @mateusoliveira43 

To get in, just pinging folks for awareness that I have updated and it should be ready to go!

pkg/uploader/provider/restic.go
```diff
@@ -188,6 +194,10 @@ func (rp *resticProvider) RunRestore(
 		"volumePath": volumePath,
 	})
 
+	if volMode == uploader.PersistentVolumeBlock {
+		return errors.New("unable to currently support block mode")
```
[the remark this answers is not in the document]
shawn-hurley: updated thanks for reminding me

pkg/uploader/provider/restic.go
```diff
@@ -134,6 +135,10 @@ func (rp *resticProvider) RunBackup(
 		return "", false, errors.New("real source is not empty, this is not supported by restic uploader")
 	}
 
+	if volMode == uploader.PersistentVolumeBlock {
+		return "", false, errors.New("unable to currently support block mode")
```
mateusoliveira43, line 139: "currently" also here :eyes:

shubham-pampattiwar:
@shawn-hurley please squash all the commits and/or remove the design commit.

sseago (approved):

shubham-pampattiwar (approved):

reasonerjt (approved):

Issue #6615, opened 2023-08-07
startTimestamp in dataupload CR is updated to the value of completionTimestamp


danfengliu:
**What steps did you take and what happened:**

`startTimestamp` in dataupload CR is updated to a wrong time which is actully a `completionTimestamp`, according to node agent log,  startTimestamp should be `2023-08-07T09:52:53Z` instead of `2023-08-07T10:02:53Z`.

```
spec:
  backupStorageLocation: default
  csiSnapshot:
    snapshotClass: ""
    storageClass: fc-test-storage-class-im
    volumeSnapshot: velero-nginx-logs-e2e-1-zxjht
  operationTimeout: 10m0s
  snapshotType: CSI
  sourceNamespace: dm-test
  sourcePVC: nginx-logs-e2e-1
status:
  completionTimestamp: "2023-08-07T10:02:53Z"
  message: 'error to expose snapshot: error wait volume snapshot ready: timed out
    waiting for the condition'
  phase: Failed
  progress: {}
  startTimestamp: "2023-08-07T10:02:53Z"
```

**Node agent pod log:**

```

1.6914019081750998e+09  INFO    Starting workers        {"controller": "datadownload", "controllerGroup": "velero.io", "controllerKind": "DataDownload", "worker count": 1}
1.6914019081752002e+09  INFO    Starting workers        {"controller": "dataupload", "controllerGroup": "velero.io", "controllerKind": "DataUpload", "worker count": 1}
time="2023-08-07T09:52:53Z" level=info msg="Reconcile backup-test-dm-3-7sl9h" controller=dataupload dataupload=velero/backup-test-dm-3-7sl9h logSource="pkg/controller/data_upload_controller.go:109"
time="2023-08-07T09:52:53Z" level=info msg="Data upload starting" controller=dataupload dataupload=velero/backup-test-dm-3-7sl9h logSource="pkg/controller/data_upload_controller.go:130"
time="2023-08-07T09:52:53Z" level=info msg="Accepting data upload backup-test-dm-3-7sl9h" logSource="pkg/controller/data_upload_controller.go:536"
time="2023-08-07T09:52:53Z" level=info msg="This datauplod has been accepted by w-cls-1690772382-standby-md-0-d5qms-7fdbd7875-lvq2x" Dataupload=backup-test-dm-3-7sl9h logSource="pkg/controller/data_upload_controller.go:550"
[...]

Issue #6612, opened 2023-08-07
Restarting Velero server cause DataUpload updated twice when it is in Accepted phase


Lyndon-Li:
Similar to #6600, when a DataUpload is in `Accepted` phase, restarting Velero server causes the CR first go to `Cancelled` state and then go to `Failed`.

Issue #6600, opened 2023-08-03
Preparation timeout DataUpload/DataDownload are processed again for a shorter preparation timeout value


Lyndon-Li:
Set the `data-mover-prepare-timeout` value for node-agent to a value which is less than 10min
Make the DataUpload/DataDownload preparation timeout in some way, e.g., backup/restore through a wrong storage class
After the timeout happens, the DataUpload/DataDownload CR is set to `Failed` and the backup/restore is set to `PartiallyFailed`
Then fix the problem in the storage class
The failed DataUpload/DataDownload CR gets prepared again and the intermediate objects get created will be left there forever

Issue #6563, opened 2023-07-31
Mark the DataUpload/DataDownload CRs with the node name for easy debugging


ywk253100:
Is it possible that mark the `DataUpload` and `DataDownload` CRs with the node name so that we can know which node-agent handles it. It is easier for debugging.

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

Lyndon-Li:
Yes, the information is already in the two CRs. And if you run `kubectl get dataupload/datadownload -n velero -w`, you will also see the node name is coming right before the data transfer starts.

ywk253100:
Seems that the node info isn't set for the failed dataupload/datadownload:
```
kubectl -n velero get dataupload -o wide
NAME          STATUS      STARTED   BYTES DONE   TOTAL BYTES   STORAGE LOCATION   AGE     NODE
dm-03-7f5gr   Failed      20h                                  default            20h
dm-04-mbxlw   Failed      20h                                  default            20h
dm-05-svkrl   Failed      20h                                  default            20h
dm-06-f6q28   Completed   5h56m     128020480    128020480     default            5h56m   wl-antrea-md-2-khgcv-79b7c9b86fxl9pkq-6fktz
dm-07-7swwz   Completed   5h53m     128020480    128020480     default            5h53m   wl-antrea-md-2-khgcv-79b7c9b86fxl9pkq-6fktz
dm-08-4vdsk   Completed   5h47m     128020480    128020480     default            5h47m   wl-antrea-md-2-khgcv-79b7c9b86fxl9pkq-6fktz
dm-09-mzs76   Completed   5h46m     128020480    128020480     default            5h46m   wl-antrea-md-2-khgcv-79b7c9b86fxl9pkq-6fktz
dm-10-2gbdf   Completed   9m46s     128020480    128020480     default            9m46s   wl-antrea-md-2-khgcv-79b7c9b86fxl9pkq-6fktz
```

Lyndon-Li:
[...]

PR #6616, opened 2023-08-07
Fix data mover controller bugs

qiuming-best:

Fixes #(issue)
#6615 
#6612
#6550
#6563
#6600

### DIFF
--- a/pkg/controller/data_download_controller.go
+++ b/pkg/controller/data_download_controller.go
@@ -27,11 +27,13 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/clock"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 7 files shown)


blackpiglet (dismissed):

pkg/controller/data_upload_controller.go
```diff
@@ -146,22 +179,55 @@ func (r *DataUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			if apierrors.IsNotFound(err) {
+				log.Debug("Unable to find DataUpload")
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, errors.Wrap(err, "getting DataUpload")
+		}
+
+		// we need to clean up resources as resources created in Expose it later than cancel action
+		// and need to clean up resources again
+		if du.Spec.Cancel {
```
Lyndon-Li, line 217: The same, check for isDataUploadInFinalState(du)

pkg/controller/data_upload_controller.go
```diff
@@ -593,18 +709,18 @@ func (r *DataUploadReconciler) onPrepareTimeout(ctx context.Context, du *velerov
...
 }
 
-func (r *DataUploadReconciler) exclusiveUpdateDataUpload(ctx context.Context, du *velerov2alpha1api.DataUpload,
-	updateFunc func(*velerov2alpha1api.DataUpload)) (bool, error) {
-	updated := du.DeepCopy()
-	updateFunc(updated)
+func ExclusiveUpdateDataUpload(ctx context.Context, client client.Client, du *velerov2alpha1api.DataUpload,
+	updateFunc func(*velerov2alpha1api.DataUpload)) (*velerov2alpha1api.DataUpload, bool, error) {
+	original := du.DeepCopy()
+	updateFunc(du)
```
Lyndon-Li, line 715: `ExclusiveUpdateDataUpload` is a fundermental function, we don't expect all the callers expect a updated `du`, so it is risky to update the `du` directly.

Actually, I suggest we leave `ExclusiveUpdateDataUpload` as is, and change the function of `acceptDataUpload`, the latter is more pure and it is safe to change it.

pkg/controller/data_download_controller.go
```diff
@@ -150,15 +179,41 @@ func (r *DataDownloadReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		// nodes when doing expose action, if detectd cancel action we need to clear up the internal
+		// resources created by velero during backup.
+		if err := r.client.Get(ctx, req.NamespacedName, dd); err != nil {
+			if apierrors.IsNotFound(err) {
+				log.Debug("Unable to find datadownload")
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, errors.Wrap(err, "getting datadownload")
+		}
+		if dd.Spec.Cancel {
```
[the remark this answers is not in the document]
qiuming-best: Updated

pkg/controller/data_upload_controller.go
qiuming-best: Updated

pkg/controller/data_upload_controller.go
qiuming-best: I've changed back to exclusiveUpdateDataDownload.

But for this function, in the update action, we needn't copy it before updating it for general usage, leading to many related UT tests failing with[ conflict errors](https://stackoverflow.com/questions/51297136/kubectl-error-the-object-has-been-modified-please-apply-your-changes-to-the-la). 

So I removed the DeepCopy function for the UT, I know your original purpose would be not willing to update the `du` directly, but for our current usage scenario there is no effect as we always get the latest CR from the API server instead of memory in each Reconcile

blackpiglet (approved):

Lyndon-Li (approved):

PR #6627, opened 2023-08-09
Continue wait on csi snapshot error

Lyndon-Li:
The errors filled into `VolumeSnapshot.VolumeSnapshotStatus.VolumeSnapshotError` may be interim and retry-able, we cannot fail immediately when seeing these errors. So need to change to code to monitor the errors only and then report them when the VS/VSC doesn't reach `readyToUse` finally

### DIFF
--- a/pkg/util/csi/volume_snapshot.go
+++ b/pkg/util/csi/volume_snapshot.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
@@ -45,8 +46,9 @@ const (
 
 // WaitVolumeSnapshotReady waits a VS to become ready to use until the timeout reaches
 func WaitVolumeSnapshotReady(ctx context.Context, snapshotClient snapshotter.SnapshotV1Interface,
-	volumeSnapshot string, volumeSnapshotNS string, timeout time.Duration) (*snapshotv1api.VolumeSnapshot, error) {
+	volumeSnapshot string, volumeSnapshotNS string, timeout time.Duration, log logrus.FieldLogger) (*snapshotv1api.VolumeSnapshot, error) {
 	var updated *snapshotv1api.VolumeSnapshot
+	errMessage := sets.NewString()
 
 	err := wait.PollImmediate(waitInternal, timeout, func() (bool, error) {
 		tmpVS, err := snapshotClient.VolumeSnapshots(volumeSnapshotNS).Get(ctx, volumeSnapshot, metav1.GetOptions{})
@@ -59,7 +61,7 @@ func WaitVolumeSnapshotReady(ctx context.Context, snapshotClient snapshotter.Sna
 		}
 
 		if tmpVS.Status.Error != nil {
-			return false, errors.Errorf("volume snapshot creation error %s", stringptr.GetString(tmpVS.Status.Error.Message))
+			errMessage.Insert(stringptr.GetString(tmpVS.Status.Error.Message))
 		}
 
 		if !boolptr.IsSetToTrue(tmpVS.Status.ReadyToUse) {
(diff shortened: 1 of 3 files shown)


anshulahuja98:
@Lyndon-Li 
The CSI plugin already waits for snapshot handle to be populated and only then proceeds to this code to wait for readyToUse.
If after that we encounter any errors for readyToUse, do you really think those will be retryable errors? 
Have you faced this with any specific drivers?

Lyndon-Li:
@anshulahuja98
As the current behavior of CSI plugin, under the context of data movement, it won't wait for the populating of snapshot handle. This is true for non-data movement case, but for data movement case, all the wait goes to data mover.

anshulahuja98:
I see, i might not be very up to date with the datamovement impl.
But IMO even in case of datamovement snapshotting should complete ( snapshotHandle populate) before datamovement. 
Since this has implication on pre/ post hooks.
Snapshotting should complete before post hooks are run.

Lyndon-Li:
@anshulahuja98 
I agree with what you mentioned. So CSI plugin should wait for the population of snapshot handle.
On the other hand, the code in the current PR is also required, since the data mover needs to wait the snapshot for its readyToUse status. And during this wait, theoretically, errors could still be filled to `VolumeSnapshot.VolumeSnapshotStatus.VolumeSnapshotError`

Lyndon-Li:
@anshulahuja98 
I got you point, it is reasonable. 
Indeed, after the snapshot handle appears, any errors are less likely to be recoverable theoretically and technically, though we don't have enough time and environments to prove this.

On the other hand, once we add the code to wait for snapshot handle in CSI plugin, only the errors that are generated during the readyToUse wait are detected here. 
Or in the other word, if the errors never happen during this period as our assumption, this code never take effect; if our assumption is wrong, this code make sure the data movement won't fail.

That is to say, this changes won't make the performance worse, it just adds an insurance when something unexpected happens in some special cases or for some special drivers. 

What do you think?

anshulahuja98:
I understand your point. I think it's fine to move ahead in that case.

pkg/util/csi/volume_snapshot.go
```diff
@@ -70,6 +72,14 @@ func WaitVolumeSnapshotReady(ctx context.Context, snapshotClient snapshotter.Sna
 		return true, nil
 	})
 
+	if err == wait.ErrWaitTimeout {
+		err = errors.Errorf("volume snapshot is not ready until timeout, errors: %v", errMessage.List())
+	}
+
+	if errMessage.Len() > 0 {
+		log.Warnf("Some errors happened during snapshot creation %v", errMessage.List())
```
kaovilai, line 80:
```suggestion
		err = errors.Errorf("volume snapshot is not ready until timeout, errors: %v", errMessage.List())
	}

	if errMessage.Len() > 0 {
		log.Warnf("Some errors happened waiting for ready snapshot %v", errMessage.List())
```

kaovilai (changes requested):

shubham-pampattiwar (dismissed):

Lyndon-Li:
@anshulahuja98 @blackpiglet 
FYI, opened a dedicated issue #6630 for the post-hook problem.

Lyndon-Li:
@kaovilai 
When the interim errors happen, the CSI snapshotter could retry on its own, what we need to do is just wait there.
If some critical errors happen, most probably, a simple retry/recreate doesn't work, so it's better to stop there and notify users to fix the problem.

Repeating creating snapshot involves big changes in the entire workflow since CSI snapshot creation involves both sync operations and async operations and also data mover later, unless there are some known significant benefits, it is not worth doing these changes.

pkg/util/csi/volume_snapshot.go
Lyndon-Li: Done, thanks.

blackpiglet (approved):

pkg/util/csi/volume_snapshot.go
```diff
@@ -70,6 +72,14 @@ func WaitVolumeSnapshotReady(ctx context.Context, snapshotClient snapshotter.Sna
 		return true, nil
 	})
 
+	if err == wait.ErrWaitTimeout {
+		err = errors.Errorf("volume snapshot is not ready until timeout, errors: %v", errMessage.List())
```
qiuming-best, line 76: which means that we don't do anything if any error happens and wait for a time-out? Is it reasonable?

kaovilai (approved):

pkg/util/csi/volume_snapshot.go
Lyndon-Li: Yes, we don't error out immediately when we see an error is filled because this the errors may be interim. See the description of this error field:
```
field Error *snapshotv1api.VolumeSnapshotError
error is the last observed error during snapshot creation, if any. This field could be helpful to upper level controllers(i.e., application controller) to decide whether they should continue on waiting for the snapshot to be created based on the type of error reported. The snapshot controller will keep retrying when an error occurrs during the snapshot creation. Upon success, this error field will be cleared
```

kaovilai: I think it is reasonable until we have a list of common resolvable issues. Most likely, external-snapshotters (csi drivers) will just retry and we hope most errors goes away.
Right now there are no resolvable errors. If errors persists, then we wait for timeout.

qiuming-best (approved):

Issue #6163, opened 2023-04-19
Update Velero Azure plugin to support AzureDNSZone storageaccount


ywk253100:
Currently, the Velero Azure plugin [hardcodes the storage service URLs for different Azure clouds](https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure/blob/v1.7.0-rc.1/velero-plugin-for-microsoft-azure/common.go#L87), but this isn't always true according to https://techcommunity.microsoft.com/t5/azure-storage-blog/public-preview-create-additional-5000-azure-storage-accounts/ba-p/3465466, we need to update the Velero Azure plugin to address this issue.

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

ywk253100:
Talked with @anshulahuja98, We will fix the issue with the following approach:
1. Introduce new SA-URI field
2. if SA URI is provided - we attempt to use it, otherwise we do GetProperties to fetch?
[Here my assumption is that almost all common roles that users would have used - would already have /reader permission on SA to do the GET call]
3. But let's say if GetPropeties fails(no permissions) - we fallback to the URL forming mechanism ([xxx.windows.core.net](http://xxx.windows.core.net/))
4. This approach will not lead to any breaking issue for existing users(since we fallback to URL forming) - and for new users we are always smart enough to do GetProperties() without them having to pass URI.
5. In case of power users where we wish to skip the GetProperties call due to ARM throttling - users can pre set the SA URI.

anshulahuja98:
As discussed with @ywk253100 , consider this issue assigned to me.

anshulahuja98:
The plan is also to add a field "UseAAD" to explicitly use AAD. The intention is to maintain current ListKeys behavior. Switching to AAD for BSL will be an opt in feature.
This is because the current ListKeys and AAD approach has different set of permissions required. So migration has to take place over time.

PR #6686, opened 2023-08-22
Make Kopia support Azure AD

ywk253100:
This commit introduces our own Azure storage provider by wrapping Kopia's implementation rather than contributing to upstream based on the following considerations:
1. Velero needs the capability to interact with the repository concurrently while Kopia doesn't, this will increase the complexity of Kopia if we contribute to upstream
2. The configuration items provided by Velero and Kopia are conflict, e.g. Velero supports customizing storage account URI which is a full path while Kopia supports customizing storage account domain which is part of the URI. We need to consider the backward compatibility and upgrade case if we contribute to upstream which needs extra efforts
3. Contribute to upstream is a longer cycle when we need to introduce new changes. With this commit, we no longer depends on upstream for the Azure storage provider part and is easy for us to maintain

Fixes #(issue)


anshulahuja98:
" Kopia supports customizing storage account domain"

Should we also make azure plugin confirm to that?
I was not aware of the kopia side of things for handling storage endpoints, we can also aim to be inline with them.

ywk253100:
> " Kopia supports customizing storage account domain"
> 
> Should we also make azure plugin conform to that? I was not aware of the kopia side of things for handling storage endpoints, we can also aim to be inline with them.

@anshulahuja98 We don't need to. The `storageAccountURI` is flexible enough to address https://github.com/vmware-tanzu/velero/issues/6163 while Kopia's storage account domain isn't.

BTW, I also submitted a draft PR https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure/pull/206 to refactor the Azure plugin to use the util functions introduced by this PR and remove the duplicated code. Any new features for the Azure storage related need to be implemented on the Velero side first after this PR is merged. This will avoid implementing new features only work for Azure plugin but not for Kopia repository and keep features consistent.

ywk253100:
I have done some manual test to cover:
- Auth with client secret and `useAAD=false`:
  - B/R with native snapshot
  - B/R with Kopia fs backup
  - B/R with Restic fs backup
  - B/R with data mover 
- Auth with client secret and `useAAD=true`:
  - B/R with native snapshot
  - B/R with Kopia fs backup
  - B/R with data mover 
- Auth with Azure AD Workload Identity and `useAAD=true`:
  - B/R with native snapshot
  - B/R with data mover 
- Auth with storage account access key:
  - B/R with data mover

kaovilai:
> our own Azure storage provider by copying Kopia's Azure storage provider code and wrapping it

I think it will be harder to maintain compatibility if we include this code separately. I suggests that we fork, rather than copy.

That way we have commit history specific to this and we are able to easily rebase with upstream kopia when needed.
There maybe more projects in the future that velero needs to fork in this way. So I suggest that we get conversations started on how velero as a project should consume and maintain it's modified version of externally originated code.

sseago:
The other advantage of a fork is it provides the opportunity to submit the changes upstream. While they may not always be accepted (especially when goals of the projects differ), when a commit *is* accepted upstream, then we can remove the local commit in the future and reduce our maintenance burden. Also, we already have a velero fork of kopia for a prior bug fix (although we should still eventually get it into an official velero repo), so we could just make that change in the same repo.

kaovilai:
If we aren't forking.. The alternative IMO is to use git submodules which will keep history of original code + patches.

pkg/repository/udmrepo/kopialib/backend/azure.go
```diff
@@ -18,43 +18,37 @@ package backend
...
 
 func (c *AzureBackend) Setup(ctx context.Context, flags map[string]string) error {
-	var err error
-	c.options.Container, err = mustHaveString(udmrepo.StoreOptionOssBucket, flags)
-	if err != nil {
-		return err
+	// As pkg/util/azure.NewStorageClient(config) is used in both repository and plugin,
+	// the caCert isn't encoded when passing to the plugin, so we need to decode the caCert
+	// before passing the config into the NewStorageClient()
+	if flags[udmrepo.StoreOptionCACert] != "" {
```
Lyndon-Li, line 38: I think we need to either keep the caCert encoded or use []byte to take it. As we discussed, some formats of the caCert could not be taken by string, there may be functionality lose if we decode it and assign it to a string.

I understand the conflict between the repo code and the plugin code, and I think the problem is on the plugin code --- it didn't consider all the caCert formats. So let's whether see we can fix this problem on plugin side.

pkg/repository/udmrepo/kopialib/backend/azure/azure_storage.go
```diff
@@ -0,0 +1,308 @@
+// Package azure implements Azure Blob Storage.
```
Lyndon-Li, line 1: I didn't get how `Velero needs the capability to interact with the repository concurrently while Kopia doesn't` is solved, could you share some more insights?

ywk253100:
@kaovilai @sseago I will submit a separate PR to move the Kopia code changes to the forked repo once the forked repo is ready. As this PR contains lots of changes, let's review and merge it first to avoid continuous conflicts.

pkg/repository/udmrepo/kopialib/backend/azure.go
ywk253100: As the comment said, the `AzureBackend.Connect()` calls `pkg/util/azure.NewStorageClient(config map[string]string)`.
The `pkg/util/azure.NewStorageClient(config map[string]string)` is also called by the plugin code while the plugin framework passes a string caCert to the plugin code, so we cannot use `[]byte` caCert.

I'll open an issue to track the caCert issue.

Lyndon-Li: If we cannot fix the problem on the plugin side, I think we need to isolate the calls so that the repo's functionality won't break. Possibly something like this:

- Let the plugin call `pkg/util/azure.NewStorageClientA(config map[string]string)`
- Add another function NewStorageClientB which accepts []byte for caCert
- NewStorageClientA calls NewStorageClientB and converts the string caCert to []byte
- The repo flow calls NewStorageClientB directly

ywk253100:
@Lyndon-Li @sseago @kaovilai @blackpiglet I have removed the copied code, please review this PR again.

Lyndon-Li (approved):

blackpiglet (approved):

Issue #6640, opened 2023-08-10
Extend VolumePolicies feature to support more actions other than just `skip` action


shubham-pampattiwar:
**Describe the problem/challenge you have**
Currently, the VolumePolicies feature which can be used to filter/handle volumes during backup only supports the `skip` action on matching conditions. Users need more actions to be supported. eg: opt in for `fs-backup`  action for some volumes and `snapshot` action for some particular type of volumes in the same backup operation can be facilitated via VolumePolicies feature if these actions are supported.

**Describe the solution you'd like**

Add support for more actions like `fs-backup` , `snapshot` etc

**Anything else you would like to add:**

**Environment:**

- Velero version (use `velero version`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

sseago:
Here's a concrete use case for this feature. A user wants to use datamover for all supported volumes and fs backup for the rest, without having to annotate the pods explicitly for which volumes to include (or exclude) from fs backup. 

To support this, the new `fs-backup` policy would be equivalent to listing all matching volumes in the "backup.velero.io/backup-volumes" annotation on the mounting pod, and the new `snapshot` policy would be equivalent to listing all matching volumes in the "backup.velero.io/backup-volumes-excludes" annotation on the mounted pod.

Then a user who wants to use VBDM for everything they can and fs backup for the rest would set `--snapshot-move-data` and `--default-volumes-to-fs-backup` on their backup, and then specify a policy of `snapshot` for CSI volumes.

blackpiglet:
Agree this is a reasonable scenario. 

```
1. the new fs-backup policy would be equivalent to listing all matching volumes in the "backup.velero.io/backup-volumes" annotation on the mounting pod.
2. And the new snapshot policy would be equivalent to listing all matching volumes in the "backup.velero.io/backup-volumes-excludes" annotation on the mounted pod.
```

[...]

PR #6697, opened 2023-08-24
Add restic resource config for file excludes

### LINKED ISSUE #4129: Exclude folders for backing up
See https://github.com/vmware-tanzu/velero/issues/1430

> The team has discussed and we're not planning to support individual file/directory backup or exclusion, so closing this issue.

Well the team made a weird decision. Excluding folders is such a basic operation done in absolutely any setup backing up data. The simplest reason would be: No one wants to backup the cache directory with millions of small files. Those will be anyways regenerated.

r10r:
Add a ResticConfig object that can be used to define restic specific options (e.g excludes) within a resource policy configmap.

The implementation is a bit limited because it only allows the definition of single ResticConfig within  a resource policy configmap
which is used for all volumes that are backed up by the `restic`.

Please see the updated documentation for a simple example.

Fixes #4129

### DIFF
--- a/site/content/docs/main/resource-filtering.md
+++ b/site/content/docs/main/resource-filtering.md
@@ -138,6 +138,8 @@ Exclude specific resources from the backup.
 
 Wildcard excludes are ignored.
 
+For restic file excludes see [restic file excludes](#restic-file-excludes)
+
 ### --exclude-namespaces
 
 Namespaces to exclude.
@@ -321,3 +323,20 @@ Velero supported conditions and format listed below:
 **Resource policies rules**
 - Velero already has lots of include or exclude filters. the resource policies are the final filters after others include or exclude filters in one backup processing workflow. So if use a defined similar filter like the opt-in approach to backup one pod volume but skip backup of the same pod volume in resource policies, as resource policies are the final filters that are applied, the volume will not be backed up.
 - If volume resource policies conflict with themselves the first matched policy will be respected when many policies are defined.
+
+### Restic configuration
+
+#### Restic file excludes
+
+See also https://restic.readthedocs.io/en/latest/040_backup.html#excluding-files
+
+```yaml
+version: v1
+resticPolicies:
+  - resticConfig:
+      excludes:
+        - /git/gitea-repositories/mirrors/**
+        - /indexers/**
+```
+
+NOTE: Currently only the first resticConfig is used and the excludes will apply to all persistent volumes.
--- a/internal/resourcepolicies/resource_policies.go
+++ b/internal/resourcepolicies/resource_policies.go
@@ -35,13 +35,26 @@ type volumePolicy struct {
 type resourcePolicies struct {
 	Version        string         `yaml:"version"`
 	VolumePolicies []volumePolicy `yaml:"volumePolicies"`
+	ResticPolicies []resticPolicy `yaml:"resticPolicies"`
 	// we may support other resource policies in the future, and they could be added separately
 	// OtherResourcePolicies []OtherResourcePolicy
+}
+
+// ResticConfig is the restic configuration to apply when restic is used for volume backup.
+type ResticConfig struct {
+	Excludes []string `yaml:"excludes"`
+}
+
+// resticPolicy defines the restic configuration to use for volume resources matching the given conditions.
+type resticPolicy struct {
+	Conditions   map[string]interface{} `yaml:"conditions"`
+	ResticConfig ResticConfig           `yaml:"resticConfig"`
 }
 
 type Policies struct {
 	version        string
 	volumePolicies []volPolicy
+	resticPolicies []resticPolicy
 	// OtherPolicies
 }
 
@@ -74,6 +87,7 @@ func (p *Policies) buildPolicy(resPolicies *resourcePolicies) error {
 	}
 
 	// Other resource policies
+	p.resticPolicies = resPolicies.ResticPolicies
 
 	p.version = resPolicies.Version
 	return nil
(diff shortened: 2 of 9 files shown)


qiuming-best:
@r10r Thanks for your great work !!!

But there are some aspects I want to discuss with you:
1. [The resource policies could be retrieved in `configmap`](https://github.com/vmware-tanzu/velero/blob/v1.11.0/pkg/controller/backup_controller.go#L472-L483), so it's unnecessary to patch the config into PVB CR annotation(velero server side). but we can get it on node-agent side.

2. It may not be suitable to pass the config through Context, in case there may be a large amount of config.

3. The config is for Restic, and we already support the Kopia uploader, we should only pass the Restic config when uploader is Restic

sseago:
This functionality is needed for kopia as well -- so whatever API we come up with needs to work for both, with restic and kopia-specific implementations behind the API.

r10r:
> This functionality is needed for kopia as well 

I think the functionality can be extended to kopia quite easily.  I'm a bit short of time right now and will focus on restic first because we need this to shrink down our offsite backups. 

@qiuming-best 

I'll have changed the implementation to use the kubernetes API for retrieving the configmap and will update the PR within the next hours. I would be pleased if you can take a look at it then.
Next week I can look into adding some tests.

site/content/docs/main/resource-filtering.md
```diff
@@ -320,4 +322,21 @@ Velero supported conditions and format listed below:
...
+```yaml
+version: v1
+resticPolicies:
+  - resticConfig:
+      excludes:
+        - /git/gitea-repositories/mirrors/**
+        - /indexers/**
+```
+
+NOTE: Currently only the first resticConfig is used.
```
draghuram, line 342: It would be good to explicitly state that the exclusion rules apply to all PVs backed up by Restic. I also wonder if allowing exclusions at PV level makes sense but it is certainly fine to have this as well.

qiuming-best:
@r10r 
what about using the structure defined below:
```golang
// resourcePolicies currently defined slice of volume policies to handle backup
type resourcePolicies struct {
	Version          string           `yaml:"version"`
	VolumePolicies   []volumePolicy   `yaml:"volumePolicies"`
	UploaderPolicies []uploaderPolicy `yaml:"uploaderPolicies"`
	// we may support other resource policies in the future, and they could be added separately
	// OtherResourcePolicies []OtherResourcePolicy
}

type UploaderConfig struct {
	Excludes []string `yaml:"excludes"`
}

type uploaderPolicy struct {
	UploaderConfig UploaderConfig `yaml:"uploaderConfig"`
}
```
Because we want to have a unified API for file system backup and restore that applies to both Restic and Kopia uploader. Of course, you can temporarily support Restic only, and for Kopia's support, a warning log message can be printed.

At the same time, if a configuration is specific to a certain uploader, when another uploader tries to use that configuration, we can print a warning log message related to unsupported features.

reasonerjt:
Would it be more consistent if we just add the enhancement to the existing volume policy?
We have plan to support more actions to select the backup approach for specific volumes, see #6640.

And this exclude setting can be referenced or set when the action is something like `fs-backup`?

cc @shubham-pampattiwar

r10r:
> Because we want to have a unified API for file system backup and restore that applies to both Restic and Kopia uploader. Of course, you can temporarily support Restic only, and for Kopia's support, a warning log message can be printed.

I think a unified API would be nice but requires a lot of additional work and doesn't eliminate the need for a provider specific configuration. 

Btw. the exclude patterns supported by kopia and restic are different.
Restic uses https://pkg.go.dev/path/filepath#Match to match the file path,
while kopia supports `.gitignore` like patterns https://kopia.io/docs/advanced/kopiaignore/

r10r:
@qiuming-best Can you approve the worfklow - so I can see that nothing else breaks ?

Lyndon-Li:
There are two aspects:

1. How to expose the options from Velero end to end
2. How to implement the options

For 1, we have a VGDP (Velero General Data Path) layer exposing functionalities of file system backup, restore and preservation. There are multiple features as the users of VGDP, i.e., fs backup, data mover. To expose the options, if the options belong to VGDP, we would like expose them on the VGDP basis, other than features, otherwise, there will be duplicated options for features. Though we don't want to expose the name of VGDP to users as it is a design concept.

For 2, VGDP have multiple providers as the uploader or repository. We would like expose the options with general names (that are independent of providers), take the current case for example, we would like have a name something like "backup policy->file system policy->exclude files". Then this general options will be passed down to the providers, and the specific provider uses its own way to implement it. That is, unified API and polymorphic implementations.

For more info of VGDP, check the [Unified Repo & Kopia integration design](https://github.com/vmware-tanzu/velero/blob/main/design/Implemented/unified-repo-and-kopia-integration/unified-repo-and-kopia-integration.md) and the [data mover design](https://github.com/vmware-tanzu/velero/blob/main/design/volume-snapshot-data-movement/volume-snapshot-data-movement.md)

There are other considerations, for example, we would like a unified policy liked entity which includes multiple configurations, other than multiple individuals for one purpose each.

Therefore, looks like we need a design where the questions could be addressed. And the design may be related to multiple issues and PRs.

sseago:
``
I think a unified API would be nice but requires a lot of additional work and doesn't eliminate the need for a provider specific configuration.

Btw. the exclude patterns supported by kopia and restic are different.
```
Ideally, those differences would be *behind* the API. We want a generic way of specifying exclusions in the Velero API, and then the restic and kopia implementations would convert those into the formats/patterns expected for restic or kopia. Otherwise, every user will have to modify their configuration manually when swapping between uploader types.

r10r:
Thanks for the explanation. 
Unfortunately I think I'm not able to meet all of the requirements, so I'm closing this PR for now. I'm looking forward to the unified exclusion API.

PR #6715, opened 2023-08-28
Remove schedule-related metrics on schedule delete

### LINKED ISSUE #1333: velero_backup_success_total metric keeps being reported for deleted schedules

Create a schedule, let it perform a few backups, then delete the schedule.

The /metrics endpoint keeps reporting the velero_backup_success_total{schedule="<the schedule>"} metric, even after all the backups have expired, until velero itself is restarted.

The metric should vanish when the schedule is deleted or after all its backups have expired, or it should stay around indefinitely (which I wouldn't prefer). In any case, the fact that it goes away when the velero pod is restarted suggests that this an artifact of the implementation rather than desired behaviour, and it does make it harder to implement certain alerting rules.

nilesh-akhade:

# Summary of the change
Currently, when a schedule is deleted, the associated metrics are still exported, which leads to inconsistencies in the metrics data. With this PR, we ensure that when a schedule is deleted, the related metrics are correctly removed or no longer exported. This enhances the accuracy and consistency of the metrics data.

## Example
If schedule `myschedule` exists with one successful backup and if we access the `/metrics` endpoint on the Velero server.

    velero_backup_success_total{schedule="myschedule"} 1
    # .. other metrics created for this schedule

After we delete the schedule `myschedule`, the above line is removed(not exported).

Fixes #1333 

### DIFF
--- a/pkg/controller/schedule_controller.go
+++ b/pkg/controller/schedule_controller.go
@@ -94,6 +94,7 @@ func (c *scheduleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	if err := c.Get(ctx, req.NamespacedName, schedule); err != nil {
 		if apierrors.IsNotFound(err) {
 			log.WithError(err).Error("schedule not found")
+			c.metrics.RemoveSchedule(req.Name)
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, errors.Wrapf(err, "error getting schedule %s", req.String())
--- a/changelogs/unreleased/6715-nilesh-akhade
+++ b/changelogs/unreleased/6715-nilesh-akhade
@@ -1,0 +1,1 @@
+Remove schedule-related metrics on schedule delete
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -505,6 +505,88 @@ func (m *ServerMetrics) InitSchedule(scheduleName string) {
 	}
 }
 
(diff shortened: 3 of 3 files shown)


pkg/controller/schedule_controller.go
```diff
@@ -94,6 +94,7 @@ func (c *scheduleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	if err := c.Get(ctx, req.NamespacedName, schedule); err != nil {
 		if apierrors.IsNotFound(err) {
 			log.WithError(err).Error("schedule not found")
+			c.metrics.RemoveSchedule(req.Name)
```
kaovilai, line 97: Do we want to consider the case where deletionTimestamp is not nil but object is not yet not found (gone from cluster)?

kaovilai: That would remove metrics even if schedule is stuck/remain in cluster due to finalizers etc. but yes just questioning if that would be preferred or not?

nilesh-akhade: If the deletion of the schedule gets stuck, the schedule will still be exported. This makes sense because the schedule is not yet deleted. Additionally, the removal of metrics is quick and should never fail. Hence, I believe that removal in the reconciliation cycle after deletion should be preferred.

kaovilai: +1

ywk253100 (approved):

shubham-pampattiwar (approved):
Thank you @nilesh-akhade

PR #6724, opened 2023-08-31
use old namespace in resource modifier

27149chen:

Use old(origin) namespace in resource modifier conditions in case namespace may change during restore

Fixes #(issue)

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -1352,6 +1352,14 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 		}
 	}
 
+	if ctx.resourceModifiers != nil {
+		if errList := ctx.resourceModifiers.ApplyResourceModifierRules(obj, groupResource.String(), ctx.log); errList != nil {
+			for _, err := range errList {
+				errs.Add(namespace, err)
+			}
+		}
+	}
+
 	// Necessary because we may have remapped the namespace if the namespace is
 	// blank, don't create the key.
 	originalNamespace := obj.GetNamespace()
@@ -1363,14 +1371,6 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 	// for easy identification of all cluster resources created by this restore
 	// and which backup they came from.
 	addRestoreLabels(obj, ctx.restore.Name, ctx.restore.Spec.BackupName)
-
-	if ctx.resourceModifiers != nil {
-		if errList := ctx.resourceModifiers.ApplyResourceModifierRules(obj, groupResource.String(), ctx.log); errList != nil {
-			for _, err := range errList {
-				errs.Add(namespace, err)
-			}
-		}
-	}
 
 	// The object apiVersion might get modified by a RestorePlugin so we need to
 	// get a new client to reflect updated resource path.
--- a/changelogs/unreleased/6724-27149chen
+++ b/changelogs/unreleased/6724-27149chen
@@ -1,0 +1,1 @@
+Use old(origin) namespace in resource modifier conditions in case namespace may change during restore
--- a/design/Implemented/json-substitution-action-design.md
+++ b/design/Implemented/json-substitution-action-design.md
@@ -96,6 +96,7 @@ resourceModifierRules:
     path: "/metadata/labels/test"
 ```
 - The above configmap will apply the JSON Patch to all the PVCs in the namespaces bar and foo with name starting with mysql. The JSON Patch will replace the storageClassName with "premium" and remove the label "test" from the PVCs.
+- Note that the Namespace here is the origin namespace of the resource, not the new namespace where the resource is going to be restored.
 - The user can specify multiple JSON Patches for a particular resource. The patches will be applied in the order specified in the configmap. A subsequent patch is applied in order and if multiple patches are specified for the same path, the last patch will override the previous patches.
 - The user can specify multiple resourceModifierRules in the configmap. The rules will be applied in the order specified in the configmap. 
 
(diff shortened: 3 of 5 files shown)


27149chen:
@anshulahuja98 PTAL

pkg/restore/restore.go
```diff
@@ -1358,14 +1366,6 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 	// and which backup they came from.
 	addRestoreLabels(obj, ctx.restore.Name, ctx.restore.Spec.BackupName)
 
-	if ctx.resourceModifiers != nil {
-		if errList := ctx.resourceModifiers.ApplyResourceModifierRules(obj, groupResource.String(), ctx.log); errList != nil {
-			for _, err := range errList {
-				errs.Add(namespace, err)
-			}
-		}
```
anshulahuja98, line 1366: I personally feel it's better if we do changes with the new namespace, (resource modifier filter having new namespace). Since this will lead to less confusion IMO. And also easier for the end user to debug any issues in resourcemodifier setup.

Otherwise think of the scenario where the original namespace is already there in the cluster and wishes to restore to a diff namespace, IMO the filters should target the new namespace for clarity.

anshulahuja98 (changes requested):

pkg/restore/restore.go
anshulahuja98: We can also document this behaviour in docs.

27149chen: But I think the goal of resource modifier is to modify an object from old to new, so there is a mapping in resource modifier config(an old part and a new part). To me, name, namespace, label and old value in given path belong to the old part, and new value is the new part. If I want to restore an object with resource modifier, I will first write rules according the origin object and pass it to velero to perform the restore. I might later change my mind and want to change the namespace or not, the rules should remain unchanged.

anshulahuja98: so namespace is just part of the change which happens if you see.

There are the other RestoreItemAction plugins which can change the YAMLs based on various rules.  Even there could be RIA to change the namespace based on certain rules.

In the design doc the approach we took was to modify just before creation and after all RIA / modifications are applied.

Our docs line could indicate that both namespace mapping and RIA would have run before Resourcemodifier is applied.

I think of RIA and namespace mapping together, either we do resourcemodifier before both or after both

In between ismore confusing IMO.

27149chen: I prefer to do resourcemodifier before both

anshulahuja98: The issue is that there are certain RIA say CSI plugin ones which are dependent on multiple resources -> VSC + PVC 

My worry is that resourcemodifier changes can potentially break the behavior of such plugins / RIAs, and it will be very hard to debug for end customer such scenarios.

27149chen: why do you think "In between ismore confusing"?

anshulahuja98: because RIA can also potentially change namespace and namespace mapping can also change namespace

kaovilai: So this PR only affect users who add resourceModifier for namespace field correct? If so, until namespace mapping feature is replaced by resourceModifier, if it ever is, resourceModifier shouldn't apply on namespace values IMO. It's meant for other spec fields.

sseago: Namespace mapping is handled separately -- it would be very confusing if the resource modifier code used both old and new ns (one on each side). Instead, we should pick which one gets used (backup namespace, or mapped-in-restore namespace), document it clearly, and stick with it. I don't have a strong opinion on which one we use, as long as it's clear, but we don't want to use *both* here.

sseago: @kaovilai "If so, until namespace mapping feature is replaced by resourceModifier" -- I don't see any reason to ever consider this. NS mapping is a separate feature that determines *where* resources go, this feature (as well as RIA/BIA actions) determine *what those resources look like*.

sseago: @anshulahuja98 "because RIA can also potentially change namespace" -- I suspect that if a RIA did this, it would result in errors. We use the target namespace from the NS mappings to pass into the dynamic client used for object creation. While it's not documented clearly right now, we probably want to document that neither resource modifier nor RIA namespace changing is supported.

27149chen: @sseago I agree with you that a RIA should never change namespace, here is a bad example: https://github.com/vmware-tanzu/velero-plugin-for-csi/pull/82

27149chen: Some reasons for using old namespace:
1. resource modifier is to modify an object from old to new, old object is more clear since new object can be changed.
2. resource modifier is a config (ConfigMap), and may be a large one, it should be reusable when some restore options are changed, like namespace mapping

sseago: @27149chen I think your second reason is the biggest reason to stick with the old namespace. Logically, namespace mapping is one of the last things to do -- first we change *what* we restore, then we change *where* we restore it to. And yes, you might have one backup that you're using as a template for restoring to different namespaces. Basically for any given backup, the old namespace will be valid for every restore that's made from it, but the new namespace may not.

sseago: @27149chen So the CSI example seems like a bit of a special case. It's not arbitrarily changing namespace outside of the NS mapping -- if it were to do that, I suspect that it would break things. Instead, it's actually using the NS mapping and changing it here (same change as velero would eventually change without it), but the reason for it is that the namespace is then used to find the matching VolumeSnapshot resource, so it needs to use the new namespace here.

anshulahuja98: @27149chen  can you update the website docs and resource modifier design docs with this callout for changes done before NS mapping?
Post that we can get this checked in. Since overall folks are inline with your proposal of using old namespace, I agree with further comments and the majority.

27149chen:
@anshulahuja98 docs updated

anshulahuja98 (changes requested):

ywk253100 (approved):

anshulahuja98 (approved):

Issue #6647, opened 2023-08-14
Consider adding a default value for SnapshotMoveData


sseago:
**Describe the problem/challenge you have**
For users who will always want to use VBDM for backups instead of plain CSI, they shouldn't have to specify `--snapshot-move-data` for every backup. We should have an install and server flag for this, similar to what we do for defaultVolumesToFsBackup

**Describe the solution you'd like**
We should create a server setting for a default value for SnapshotMoveData. When a backup CR does not include the `spec.snapshotMoveData` field, rather than treating it as false, the velero backup controller should set `request.Spec.SnapshotMoveData` to the default value set it to the server default value. An install option (and server flag) for `--default-snapshot-move-data` would need to be added, and the backup controller would supply this default value for backups that do already not specify it.

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

PR #6751, opened 2023-09-04
Fix issue 6647

Lyndon-Li:
Fix issue #6647, add the `--default-snapshot-move-data` parameter to Velero install, so that users don't need to specify `--snapshot-move-data` per backup when they want to move snapshot data for all backups


blackpiglet (approved):

sseago (approved):

sseago:
@Lyndon-Li Will this be cherry-picked to release-1.12 as well?

Issue #6753, opened 2023-09-04
Async item operations always fail on read-only BSL


Lyndon-Li:
If a BSL is set to readOnly for its accessMode, the restore will be marked as `ParitiallyFailed` and all the async operations are left orphaned.
For example, if the BSL is readOnly, the data mover restore will be set as `ParitiallyFailed` quickly and the DataDownloads continue and finally complete.

It is definitely a valid use case to set the BSL readOnly while doing a restore, however, at present, Velero restore cannot follow the read-only BSL rule as it needs to write the restore metadata to the backup store.
At present, we cannot solve this contradiction, so Velero restore doesn't do the BSL readOnly check but ignore it.
The async operation should follow the same behavior until the conflict is fixed.

Lyndon-Li:
The preliminary idea is to remove [the code here](https://github.com/vmware-tanzu/velero/blob/8d3a67544d9f68c24316cfc987e9e23ae9579e16/pkg/controller/restore_operations_controller.go#L142C2-L142C2) for now.

@sseago Let me know if you have different ideas.

PR #6757, opened 2023-09-05
Fix issue 6753

Lyndon-Li:
Fix issue #6753, remove the check for read-only BSL in restore async operation controller since Velero cannot fully support read-only mode BSL in restore at present


reasonerjt (approved):

reasonerjt:
I believe this should be cherry picked to v1.12?

Lyndon-Li:
> I believe this should be cherry picked to v1.12?

Yes, see #6758

ywk253100 (approved):

Issue #6572, opened 2023-08-01
Backup storage location setting is not work


qiuming-best:
**What steps did you take and what happened:**

1. First creating one new bsl named 'bsl-test'
2. Setting the `bsl-test` as default BSL
velero backup-location set bsl-test --default
3. Executing `velero backup-location get` 
```bash
NAME       PROVIDER   BUCKET/PREFIX   PHASE       LAST VALIDATED                  ACCESS MODE   DEFAULT
bsl-test   aws        mqiu-bucket     Available   2023-08-01 14:39:17 +0800 CST   ReadWrite     true
default    aws        mqiu-bucket     Available   2023-08-01 14:38:39 +0800 CST   ReadWrite     true
```
5. Do backup without specify which BSL used
```bash
velero backup create t3 --include-namespaces nginx-example --default-volumes-to-fs-backup --snapshot-volumes=false --wait
```
7. Describe backups shows it still using the `default` BSL instead of the specific one
```bash
velero backup get t3
NAME   STATUS      ERRORS   WARNINGS   CREATED                         EXPIRES   STORAGE LOCATION   SELECTOR
t3     Completed   0        0          2023-08-01 14:42:44 +0800 CST   29d       default            <none>
``` 
**What did you expect to happen:**
1. Executing `velero backup-location get`  only shows one BSL as the default 
2. Backup should backed up into the default BSL

**The following information will help us better understand what's going on**:

_If you are using velero v1.7.0+:_  
Please use `velero debug  --backup <backupname> --restore <restorename>` to generate the support bundle, and attach to this issue, more options please refer to `velero debug --help` 

_If you are using earlier versions:_  
Please provide the output of the following commands (Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)
- `kubectl logs deployment/velero -n velero`
- `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
- `velero backup logs <backupname>`
- `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
- `velero restore logs <restorename>`

**Anything else you would like to add:**

**Environment:**

- Velero version (use `velero version`): 
- Velero features (use `velero client config get features`): 
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

[...]

PR #6771, opened 2023-09-06
Fix default BSL setting not work

qiuming-best:

Velero could set the default backup storage location as below:
- First way: Set a default backup storage location by passing a `--default` flag when running `velero backup-location create`.

  ```
  velero backup-location create backups-primary \
    --provider aws \
    --bucket velero-backups \
    --config region=us-east-1 \
    --default
  ```
- Second way: Set a default backup storage location by passing a  `--default` flag when running `velero backup-location set`.
  ```bash
  velero backup-location set backups-primary --default
  ```
  We also could remove the default backup storage location by this command, below is one example
  ```bash
  velero backup-location set backups-primary --default=false
  ```
- Third way: Set a default backup storage location by passing `--default-backup-storage-location` flag on the `velero server` command.
   ```bash
  velero server --default-backup-storage-location backups-primary
   ```
Note: Only could have one default backup storage location, which means it's not allowed to set two default backup storage locations at the same time, the priorities among these three are as follows:
- if velero server side has specified one default backup storage location, suppose it's `A`
  - if `A` backup storage location exists, it's not allowed to set a new default backup storage location
  - if `A` does not exist
    - if using `velero backup-location set` or `velero backup-location create --default` command
      - it could be successful if no default backup storage location exists.
      - it would fail if already exist one default backup storage location. (So it need to remove other default backup storage location at first)
- if velero server side has not specified one default backup storage location
  - if using `velero backup-location set` or `velero backup-location create --default` command
    - it could be successful if no default backup storage location exists.
    - it would fail if already exist one default backup storage location. (So it need to remove other default backup storage location at first)

Fixes #(issue)
https://github.com/vmware-tanzu/velero/issues/6572

### DIFF
--- a/pkg/cmd/cli/backuplocation/create.go
+++ b/pkg/cmd/cli/backuplocation/create.go
@@ -31,6 +31,7 @@ import (
 
 	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/vmware-tanzu/velero/internal/storage"
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"github.com/vmware-tanzu/velero/pkg/builder"
 	"github.com/vmware-tanzu/velero/pkg/client"
@@ -204,19 +205,13 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
 
 	if o.DefaultBackupStorageLocation {
 		// There is one and only one default backup storage location.
-		// Disable any existing default backup storage location.
-		locations := new(velerov1api.BackupStorageLocationList)
-		if err := kbClient.List(context.Background(), locations, &kbclient.ListOptions{Namespace: f.Namespace()}); err != nil {
+		// Disable any existing default backup storage location first.
+		defalutBSLs, err := storage.GetDefaultBackupStorageLocations(context.Background(), kbClient, f.Namespace())
+		if err != nil {
 			return errors.WithStack(err)
 		}
-		for i, location := range locations.Items {
-			if location.Spec.Default {
-				location.Spec.Default = false
-				if err := kbClient.Update(context.Background(), &locations.Items[i], &kbclient.UpdateOptions{}); err != nil {
-					return errors.WithStack(err)
-				}
-				break
-			}
+		if len(defalutBSLs.Items) > 0 {
+			return errors.New("there is already exist default backup storage location, please unset it first or do not set --default flag")
 		}
 	}
 
--- a/pkg/cmd/cli/backuplocation/set.go
+++ b/pkg/cmd/cli/backuplocation/set.go
@@ -28,11 +28,13 @@ import (
 
 	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/vmware-tanzu/velero/internal/storage"
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"github.com/vmware-tanzu/velero/pkg/builder"
 	"github.com/vmware-tanzu/velero/pkg/client"
 	"github.com/vmware-tanzu/velero/pkg/cmd"
 	"github.com/vmware-tanzu/velero/pkg/cmd/util/flag"
+	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
 )
 
 func NewSetCommand(f client.Factory, use string) *cobra.Command {
@@ -58,7 +60,7 @@ type SetOptions struct {
 	Name                         string
 	CACertFile                   string
 	Credential                   flag.Map
-	DefaultBackupStorageLocation bool
+	DefaultBackupStorageLocation flag.OptionalBool
 }
 
 func NewSetOptions() *SetOptions {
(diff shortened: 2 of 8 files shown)


pkg/cmd/cli/backuplocation/create.go
```diff
@@ -204,19 +205,13 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
...
-		if err := kbClient.List(context.Background(), locations, &kbclient.ListOptions{Namespace: f.Namespace()}); err != nil {
+		// Disable any existing default backup storage location first.
+		defalutBSLs, err := storage.GetDefaultBackupStorageLocations(context.Background(), kbClient, f.Namespace())
+		if err != nil {
 			return errors.WithStack(err)
 		}
-		for i, location := range locations.Items {
-			if location.Spec.Default {
-				location.Spec.Default = false
-				if err := kbClient.Update(context.Background(), &locations.Items[i], &kbclient.UpdateOptions{}); err != nil {
```
qiuming-best, line 215: bug with using &locations.Items[i], which could not set the specific location to  Default false

pkg/cmd/cli/backuplocation/create.go
```diff
@@ -204,19 +205,13 @@ func (o *CreateOptions) Run(c *cobra.Command, f client.Factory) error {
...
-					return errors.WithStack(err)
-				}
-				break
```
qiuming-best, line 218: we could not break here, per the comment above `// There is one and only one default backup storage location.
		// Disable any existing default backup storage location.`, If break here only one BSL could be unset

pkg/cmd/cli/backuplocation/set.go
```diff
@@ -104,36 +105,40 @@ func (o *SetOptions) Run(c *cobra.Command, f client.Factory) error {
...
-				// Do not update if the origin default BSL is the current one.
+		// check whether it is already unset default
+		foundDefault := false
+		for _, bsl := range defalutBSLs.Items {
+			if bsl.Name == o.Name {
+				foundDefault = true
 				break
 			}
-			location.Spec.Default = false
-			if err := kbClient.Update(context.Background(), &locations.Items[i], &kbclient.UpdateOptions{}); err != nil {
```
qiuming-best, line 132: bug with using &locations.Items[i], which could not set the specific location to Default false

pkg/cmd/cli/backuplocation/set.go
```diff
@@ -104,36 +105,40 @@ func (o *SetOptions) Run(c *cobra.Command, f client.Factory) error {
...
-				return errors.WithStack(err)
-			}
-			break
```
qiuming-best, line 135: we could not break here, per the comment above `// There is one and only one default backup storage location. // Disable any existing default backup storage location.`, If break here only one BSL could be unset

qiuming-best:
After going through the original codes, it wants to always keep one default BSL, but it does not work to unset the default BSL for the bug mentioned

pkg/cmd/cli/backuplocation/set.go
```diff
@@ -104,36 +105,40 @@ func (o *SetOptions) Run(c *cobra.Command, f client.Factory) error {
...
 		return errors.WithStack(err)
 	}
 
-	if o.DefaultBackupStorageLocation {
-		// There is one and only one default backup storage location.
-		// Disable the origin default backup storage location.
-		locations := new(velerov1api.BackupStorageLocationList)
-		if err := kbClient.List(context.Background(), locations, &kbclient.ListOptions{Namespace: f.Namespace()}); err != nil {
-			return errors.WithStack(err)
+	if !o.DefaultBackupStorageLocation { // unset default backup storage location
```
qiuming-best, line 113: which means user use command to unset default bsl

blackpiglet, line 113: If the `velero backup-location set` command doesn't set the `--default` parameter, it can also hit this logic.
I think it can confuse the user because the user could meet the default BSL validation error when the user doesn't have the intention to set the default BSL.

pkg/cmd/cli/backuplocation/set.go
```diff
@@ -104,36 +105,40 @@ func (o *SetOptions) Run(c *cobra.Command, f client.Factory) error {
...
+		if len(defalutBSLs.Items) == 0 {
```
ywk253100, line 114: Seems the `set` command only targets one specific `BSL`, why do you need to check others?

pkg/cmd/cli/backuplocation/set.go
qiuming-best: you are right, I've modified it, please help me to check it again

pkg/cmd/cli/backuplocation/set.go
qiuming-best: this checking logic is for when the user uses the command `velero backup-location set $BSL --default=false`
to unset the default for one specific `BSL`

qiuming-best: I've removed the unset logic check, we allow the user unset one specific BSL repeatedly

pkg/cmd/cli/backuplocation/set.go
```diff
@@ -104,6 +107,11 @@ func (o *SetOptions) Run(c *cobra.Command, f client.Factory) error {
 		}
 	}
 
+	defalutBSLs, err := storage.GetDefaultBackupStorageLocations(context.Background(), kbClient, f.Namespace())
```
ywk253100, line 110: Move this logic inside the `if` block in line 126 to avoid unnecessary BSL listing

pkg/controller/backup_storage_location_controller.go
```diff
@@ -220,3 +210,72 @@ func (r *backupStorageLocationReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+	var defaultFound bool
+	for i, location := range locationList.Items {
+		if location.Spec.Default {
+			defaultBSLs = append(defaultBSLs, &locationList.Items[i])
+		}
+	}
+
+	if len(defaultBSLs) > 1 { // more than 1 default BSL
+		// find the most recent updated default BSL
+		var mostRecentUpdatedBSL *velerov1api.BackupStorageLocation
```
ywk253100, line 230: Rename it to `mostRecentCreatedBSL`

pkg/cmd/cli/backuplocation/set.go
qiuming-best: updated

pkg/controller/backup_storage_location_controller.go
qiuming-best: updated

ywk253100 (approved):

blackpiglet (approved):

PR #6804, opened 2023-09-12
Make configmapref check case insensitive

anshulahuja98:

For resource modifier and resource policies, make the check for configmap reference case insensitive.

Fixes #(issue)

- [ X] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required` as a comment on this pull request.

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	snapshotv1api "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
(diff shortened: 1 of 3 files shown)


pkg/controller/backup_controller.go
```diff
@@ -476,7 +477,7 @@ func (b *backupReconciler) prepareBackupRequest(backup *velerov1api.Backup, logg
 		request.Status.ValidationErrors = append(request.Status.ValidationErrors, "encountered labelSelector as well as orLabelSelectors in backup spec, only one can be specified")
 	}
 
-	if request.Spec.ResourcePolicy != nil && request.Spec.ResourcePolicy.Kind == resourcepolicies.ConfigmapRefType {
+	if request.Spec.ResourcePolicy != nil && strings.EqualFold(request.Spec.ResourcePolicy.Kind, resourcepolicies.ConfigmapRefType) {
```
mateusoliveira43, line 480: Is it possible to add a simple test just to check this new behavior?

kaovilai (approved):

pkg/controller/backup_controller.go
anshulahuja98: Updated the test case for resourcemodifier, that should be enough coverage.
For resourcepolicies, it is not trivial to add for now, since there are no controller side tests from what I can see.

reasonerjt:
@anshulahuja98 
What's the reason to make it case-insensitive?

anshulahuja98:
> @anshulahuja98 What's the reason to make it case-insensitive?

When adding the reference through CLI it doesn't matter, but if we try using backupcr / restore CR, it is easy to make a case choice such as "ConfigMap", and that will not log any error, it will lead to skipping of resource mod/policies silently.

It's a minor change just for ease and avoiding mistakes.

anshulahuja98:
Do you see any concern with this @reasonerjt ?
it's a minor change...

sseago (approved):

shubham-pampattiwar (approved):

shubham-pampattiwar:
@reasonerjt Any concerns with this PR ? We encountered this issue in our testing as well.

shubham-pampattiwar:
@anshulahuja98 It would be really great to get this in for velero 1.12.2

PR #6829, opened 2023-09-15
design schedule unpaused

kaovilai:
```
```

Fixes #(issue)


draghuram:
Proposal looks good to me though I would have preferred trigger to be false by default. By far, that is the most intuitive and least surprising.

kaovilai:
I don't mind moving it to false by default if community agrees.

Just didn't wanna break anyone relying on this.. but to be fair this will go in a new minor version so probably enough to make note of the changes.

kaovilai:
After looking at the code.. I think saying schedule unpause always cause backup to trigger is inaccurate.

To be more precise, the following code is used, based on last backup time to generate if it needs to run after unpausing.
```
func getNextRunTime(schedule *velerov1.Schedule, cronSchedule cron.Schedule, asOf time.Time) (bool, time.Time) {
	var lastBackupTime time.Time
	if schedule.Status.LastBackup != nil {
		lastBackupTime = schedule.Status.LastBackup.Time
	} else {
		lastBackupTime = schedule.CreationTimestamp.Time
	}

	nextRunTime := cronSchedule.Next(lastBackupTime)

	return asOf.After(nextRunTime), nextRunTime
}
```

So perhaps I should change this from `--unpause-triggers` so in the implementation, perhaps it would take note when unpause happens (added annotation), and then if unpause-triggers false and unpause timestamp annotation says less than a second ago, it will generate next run time based on `unpauseTimestamp` rather than 
```
	nextRunTime := cronSchedule.Next(lastBackupTime)
```

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Detailed Design
+### CLI Changes
+`velero schedule unpause` will now take an optional bool flag `--unpause-triggers` to allow user to override the behavior configured in the schedule spec.
+
+`velero schedule unpause schedule-1 --unpause-triggers=false` will unpause the schedule but not trigger the backup immediately.
+
+`velero schedule unpause schedule-1 --unpause-triggers=true` will unpause the schedule and trigger the backup immediately.
+
+`velero schedule unpause schedule-1` will unpause the schedule and trigger the backup immediately. This is the same behavior as before.
```
sseago, line 26: Shouldn't not passing the flag leave `Spec.UnpauseTriggers` as-is? i.e. if it's set to the default (true) value, then the backup is triggered immediately. however, if it's set to false (either because the last unpause set it to false, or the user set it to false explicitly when creating the schedule, shouldn't this previous value setting be honored?

kaovilai: ok that make sense. I will clear the design up. Thanks!

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,102 @@
...
+## High-Level Design
+Add a new field with to the schedule spec to allow user to configure the behavior defaulting to prior behavior of triggering backup immediately.
+
+## Detailed Design
+### CLI Changes
+`velero schedule unpause` will now take an optional bool flag `--unpause-triggers` to allow user to override the behavior configured in the schedule spec.
+
+`velero schedule unpause schedule-1 --unpause-triggers=false` will unpause the schedule but not trigger the backup immediately.
+
+`velero schedule unpause schedule-1 --unpause-triggers=true` will unpause the schedule and trigger the backup immediately.
```
shubham-pampattiwar, line 24: Additionally, This will also create a backup at the scheduled time, lets add this explicitly.

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,102 @@
...
+	// OwnerReferences on backups created by this Schedule.
+	// +optional
+	// +nullable
+	UseOwnerReferencesInBackup *bool `json:"useOwnerReferencesInBackup,omitempty"`
+
+	// Paused specifies whether the schedule is paused or not
+	// +optional
+	Paused bool `json:"paused,omitempty"`
+
++	// UnpauseTriggers specifies whether to trigger backup immediately when schedule is unpaused
```
shubham-pampattiwar, line 51: Also, add that the default value is true if not set by the user, this comment gets converted to CRD descriptions so it will be helpful for the users.

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,102 @@
...
++	// +kubebuilder:default:=true
```
shubham-pampattiwar, line 52: Also, please make this field `optional` in order to avoid upgrade issues.

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,102 @@
...
+	// +optional
+	ValidationErrors []string `json:"validationErrors,omitempty"`
+}
+```
+
+`LastUnpaused` will be set to the current time when a schedule is unpaused. When a schedule is paused, `LastUnpaused` will be set to nil.
+
+## Alternatives Considered
+
+`velero server` flags to configure the behavior. This is not ideal as it is a global setting. This behavior should be configurable by each schedule author.
```
shubham-pampattiwar, line 87: Lets also add the pro for the alternative that if its a server setting then the user only has to configure this once.

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,102 @@
...
+
```
shubham-pampattiwar, line 88: Maybe we add this a server setting as well as the backup schedule spec, What do you think ? This will be similar to the backup TTL spec.

shubham-pampattiwar (changes requested):
Added a few thoughts/comments.

design/schedule-unpaused-triggers-config_design.md
kaovilai: This would be a benefit if there's a single user of this velero server I agree. Could certainly be included as well?

design/schedule-unpaused-triggers-config_design.md
kaovilai: Sure.

design/schedule-unpaused-triggers-config_design.md
kaovilai: When there is `default` then optional flag doesn't apply.

kaovilai: I can do some upgrade testing see what it does.

The default here is applied at the CRD level, not via velero processes.

kaovilai: I'll just leave the optional markers in anyway. Generated CRD won't differ (ie. it left out optional from generated CRD when default is defined)

kaovilai: Issue we had was, both non optional field and no default. I'll update https://github.com/vmware-tanzu/velero/issues/6854

design/schedule-unpaused-triggers-config_design.md
kaovilai: But if velero server setting is included, then we cannot set default in the CRD as that would override any chance of velero server setting the desired behavior.

Could you clarify your preference?
- CRD defaults to true (prevent ability to add server defaults).
- or server flags defaults.

shubham-pampattiwar: I think my preference would be a server flag default and an optional spec field on the CRD.

design/schedule-unpaused-triggers-config_design.md
shubham-pampattiwar: yes agreed.

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,113 @@
...
+## Background
+Currently, the default behavior is to run the trigger immediately and may not be desired by all users (https://github.com/vmware-tanzu/velero/issues/6517)
+
+## Goals
+- Make schedule unpaused behavior configurable
+
+## Non Goals
+- Changing the default behavior
+
```
shubham-pampattiwar, line 13: Maybe also add a use-case/scenario/user story section, one example would suffice I think.

design/schedule-unpaused-triggers-config_design.md
sseago: So optional CR field, type is `*bool`, and also add server flag (and if you add server flag, you also need corresponding install flag)

design/schedule-unpaused-triggers-config_design.md
kaovilai: It's already in the background section. Will expand on it.

Dedicated use-case/scenario section is not in the design template.

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,108 @@
+# Schedule Unpaused Triggers Config Design
+## Abstract
+User want to configure what happens when schedule is unpaused. Either run backup immediately or at the next cron schedule.  This design proposes to add a new field to the schedule spec to allow user to configure the behavior
```
ywk253100, line 3: Do we really want to make it a configuration item? IMO, only taking backup at the next cron schedule is more reasonable. If users want to run backup immediately, they can trigger it by creating a backup from the schedule manually.
IMO, I'd prefer to avoid exposing more and more configuration items to the end users

kaovilai: So I probably need to reword
> Either run backup immediately

The ask from the user issue was that backup was unintentionally running immediately after unpausing caused by the current logic being based on LastBackupTimestamp.

Schedule that has been backedup with a timestamped then paused, and long enough time has passed, unpausing can make backup go off immediately.

User wants ability to make it not go off when unpausing but wait till next cron after unpause timestamp.

@rnarenpujari am I summarizing the need here pretty well?

kaovilai, line 3:
```suggestion
User want to configure what happens when schedule is unpaused. Either run backup when schedule is unpaused if it is due based on .Status.LastBackup or at the next cron schedule.  This design proposes to add a new field to the schedule spec to allow user to configure the behavior
```

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Background
+Currently, the default behavior is to trigger backup when `.Status.LastBackup` is nil or due which can be immediately. This may not be a desired by all users (https://github.com/vmware-tanzu/velero/issues/6517)
+
+User want ability to configure the behavior when schedule is unpaused. Either run backup immediately or at the next cron schedule.
```
kaovilai, line 8:
```suggestion
User want ability to configure the behavior when schedule is unpaused. Either run backup when schedule is unpaused if it is due based on .Status.LastBackup or at the next cron schedule.
```

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Goals
+- Make schedule unpaused behavior configurable
+
+## Non Goals
+- Changing the default behavior
+
+
+## High-Level Design
+Add a new field with to the schedule spec to allow user to configure the behavior defaulting to prior behavior of triggering backup immediately.
```
kaovilai, line 18:
```suggestion
Add a new field with to the schedule spec to allow user to configure the behavior defaulting to prior behavior of triggering backup when schedule is unpaused if it is due based on .Status.LastBackup.
```

design/schedule-unpaused-triggers-config_design.md
```diff
@@ -0,0 +1,108 @@
...
+## Alternatives Considered
+
+N/A
+
+
+## Security Considerations
+None
+
+## Compatibility
+Upon upgrade, the new field will be added to the schedule spec automatically and will default to the prior behavior of triggering backup immediately.
```
kaovilai, line 100:
```suggestion
Upon upgrade, the new field will be added to the schedule spec automatically and will default to the prior behavior of triggering backup when schedule is unpaused if it is due based on .Status.LastBackup.
```

design/schedule-unpaused-triggers-config_design.md
ywk253100: I understand the requirement and I think the design of this PR has already resolved the issue properly.
What I want to discuss is whether it is necessary to make taking backup immediately or not to be configurable.  Could we just make the change that doesn't take backup immediately with the solution in this design?

ywk253100 (dismissed):

sseago (approved):

design/schedule-unpaused-triggers-config_design.md
kaovilai: Server flag is now included in the design.

sseago (approved):

kaovilai:
Switched out added ScheduleStatus field from LastUnpaused to LastSkipped as that is more practical and more useful.

ywk253100 (approved):

PR #6906, opened 2023-10-01
fix error log messages if s3 bucket is versioned

chrisamti:

if s3 bucket is versioned and velero deletes all files from a specific backup folder, s3 keeps the folder (but not the files inside) visible, because of the deleted files are kept in deleted state. A s3 versioned bucket keeps folders visible until s3 retention completely removes all versions and all delete marker from the s3 folder.  

Fixes error message like: 
```
velero time="2023-10-01T12:25:20Z" level=error msg="Error getting backup metadata from backup store" backup=k14-prod-digidatei01-0-every-6h-ttl-23-hours-20230929035140 backupLocation=velero/k14-prod-digidatei01-0 controller=backup-sync error="rpc error: code = Unknown desc = error getting object backups/k14-prod-digidatei01-0-every-6h-ttl-23-hours-20230929035140/velero-backup.json: NoSuchKey: The specified key does not exist.\n\tstatus code: 404, request id: 1789FB5F2C42E211, host id: 470a0afc0f6fdbebb6a8e14ece29122784c0656e19563387135a7c8e02feb163" error.file="/go/src/github.com/vmware-tanzu/velero/pkg/persistence/object_store.go:302" error.function="github.com/vmware-tanzu/velero/pkg/persistence.(*objectBackupStore).GetBackupMetadata" logSource="pkg/controller/backup_sync_controller.go:147"
```

### DIFF
--- a/pkg/persistence/object_store.go
+++ b/pkg/persistence/object_store.go
@@ -235,6 +235,16 @@ func (s *objectBackupStore) ListBackups() ([]string, error) {
 		// each of those off to get the backup name.
 		backupName := strings.TrimSuffix(strings.TrimPrefix(prefix, s.layout.subdirs["backups"]), "/")
 
+		// if a bucket is versioned, the s3 folder still exists, even if all s3 objects have been deleted.
+		// We should only take s3 folders having a velero-backup.json object.
+		if ok, errObjectExists := s.objectStore.ObjectExists(s.bucket, s.layout.getBackupMetadataKey(backupName)); !ok || errObjectExists != nil {
+			// velero-backup.json or error do not add to list
+			if errObjectExists != nil {
+				s.logger.WithError(errObjectExists).Debugf("could not check if %s/%s exists", s.bucket, s.layout.getBackupMetadataKey(backupName))
+			}
+			continue
+		}
+
 		output = append(output, backupName)
 	}
 
--- a/pkg/persistence/object_store_test.go
+++ b/pkg/persistence/object_store_test.go
@@ -194,6 +194,16 @@ func TestListBackups(t *testing.T) {
 			},
 			expectedRes: []string{"backup-1", "backup-2"},
 		},
+		{
+			name:   "normal case with backup store prefix ignore folder missing velero-backup.json ",
+			prefix: "velero-backups/",
+			storageData: map[string][]byte{
+				"velero-backups/backups/backup-1/velero-backup.json": encodeToBytes(builder.ForBackup("", "backup-1").Result()),
+				"velero-backups/backups/backup-2/velero-backup.json": encodeToBytes(builder.ForBackup("", "backup-2").Result()),
+				"velero-backups/backups/backup-3/":                   encodeToBytes(builder.ForBackup("", "backup-3").Result()),
+			},
+			expectedRes: []string{"backup-1", "backup-2"},
+		},
 	}
 
 	for _, tc := range tests {


pkg/persistence/object_store.go
```diff
@@ -235,6 +235,16 @@ func (s *objectBackupStore) ListBackups() ([]string, error) {
 		// each of those off to get the backup name.
 		backupName := strings.TrimSuffix(strings.TrimPrefix(prefix, s.layout.subdirs["backups"]), "/")
 
+		// if a bucket is versioned, the s3 folder still exists, even if all s3 objects have been deleted.
```
blackpiglet, line 238: When talking about S3, is it the AWS S3 storage or all the S3-compatible storage?
I'm not sure whether this is for all S3-compatible storage.

pkg/persistence/object_store.go
```diff
@@ -235,6 +235,16 @@ func (s *objectBackupStore) ListBackups() ([]string, error) {
...
+		// We should only take s3 folders having a velero-backup.json object.
+		if ok, errObjectExists := s.objectStore.ObjectExists(s.bucket, s.layout.getBackupMetadataKey(backupName)); !ok || errObjectExists != nil {
```
blackpiglet, line 240: By this check, we only know the backup's metadata file doesn't exist. We don't whether the bucket versioning is on.
I think it's possible some other corner cases are also included by this check.

blackpiglet:
I'm not sure about this change, and I also don't have a clear thought on how to address this situation.
IMO, the retention policy setting was not considered by Velero before. Maybe we need a more sophisticated design for this.

pkg/persistence/object_store.go
chrisamti: I think it doesn't matter whether it's S3 and versioning is enabled or not or any other storage that knows versioning is in place. As soon as velero-backup.json is missing, but the backup folder name exists, it no longer makes sense to check the folder for velero backups.

chrisamti:
> I'm not sure about this change, and I also don't have a clear thought on how to address this situation. IMO, the retention policy setting was not considered by Velero before. Maybe we need a more sophisticated design for this.

I retention policies should not be handled or changed by velero. 
It is the admin's responsibility to set the correct s3 retention policies. Velero should simply be aware that there is something like S3 versioning and should not show any error messages if a backup folder still exists, but all the files in there are not listed, but still exist because of versioning and only disappear completely when the s3 retention policy removed the files and thus the folder completely.

pkg/persistence/object_store.go
chrisamti: Right. Most S3 implementations support versioning - and some maybe do not. 
I think velero should be able to work on versioned and non versioned buckets, without producing endless error messages on versioned buckets (but still does it's job). 

This PR only disables the error messages, if velero is using a versioned bucket.  
 
BTW: we currently use this PR in a self created docker image based on version 1.12.0 based in production without any troubles.

blackpiglet: It makes sense to ignore this case, but maybe we should reword the comments because we cannot ensure this scenario is caused by bucket versioning, since we don't even know whether the bucket versioning is enabled or not. That is one possible scenario.

blackpiglet:
> I think retention policies should not be handled or changed by velero.
> It is the admin's responsibility to set the correct s3 retention policies. Velero should simply be aware that there is something like S3 versioning and should not show any error messages if a backup folder still exists, but all the files in there are not listed, but still exist because of versioning and only disappear completely when the s3 retention policy removed the files and thus the folder completely.

Sure, Velero should not handle the bucket versioning setting, but maybe Velero should know whether it's enabled or not. I have some concern about some corner cases, for example:
If the created backup names never duplicate, it's OK. If we create a backup called `testing`, and there are some versioned data left because of the retention policy setting after the backup is deleted, then we try to create a backup named as `testing` again. What is the result? Will it fail?

kaovilai:
Can you update the PR title?

chrisamti:
> Sure, Velero should not handle the bucket versioning setting, but maybe Velero should know whether it's enabled or not. I have some concern about some corner cases, for example: If the created backup names never duplicate, it's OK. If we create a backup called `testing`, and there are some versioned data left because of the retention policy setting after the backup is deleted, then we try to create a backup named as `testing` again. What is the result? Will it fail?

on versioned s3, deleted files are deleted and not visible until somebody tells s3 to bring back the deleted object. 
If you overwrite a not deleted version of an object, der old version of the object is kept in the version history of the s3 bucket and is not visible until you tell s3 to get an older version. 
To you question. If you use the same folder with still existing data inside, it will not fail. 

But hey. Feel free to deny this PR if you got that many concerns and question marks.

For us, we keep patching velero as this PR it does, because we use s3 versioned and we do not want to have tons of error messages thrown from velero.

blackpiglet:
Maybe I'm a little too conservative about this change, but need to make sure it will work in all environments.
I will do some tests on some other cloud providers' environments. If everything works as expected, we can merge this PR.

pkg/persistence/object_store.go
kaovilai: I guess this would be generic enough for other non s3 cloud provider scenarios which may or may not have the same behavior.

Issue #6913, opened 2023-10-03
Velero Built-in Datamover: Backup stucks in phase WaitingForPluginOperations when Node Agent pod gets restarted


shubham-pampattiwar:
**What steps did you take and what happened:**

- Execute backup with native datamover.
- Node agent pod gets restarted
- Backups got stuck in WaitingForPluginOperations phase when node agent pod got restarted.

**What did you expect to happen:**
Backup to get failed/partially failed

**Anything else you would like to add:**

DataUpload is also in InProgress status, looks like cancel true is not taking any effect. 
```
$ oc get dataupload test-backup2-8vb28 -o yaml
apiVersion: velero.io/v2alpha1
kind: DataUpload
metadata:
  creationTimestamp: "2023-08-22T10:56:04Z"
  generateName: test-backup2-
  generation: 5
  labels:
    velero.io/async-operation-id: du-eab99b56-a863-4a04-b96d-125b29ac88c4.fa094c8b-8308-488f5f389
    velero.io/backup-name: test-backup2
    velero.io/backup-uid: eab99b56-a863-4a04-b96d-125b29ac88c4
    velero.io/pvc-uid: fa094c8b-8308-488f-af41-d77c2278244f
  managedFields:
  - apiVersion: velero.io/v2alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:generateName: {}
        f:labels:
          .: {}
          f:velero.io/async-operation-id: {}
          f:velero.io/backup-name: {}
          f:velero.io/backup-uid: {}
          f:velero.io/pvc-uid: {}
        f:ownerReferences:
          .: {}
          k:{"uid":"eab99b56-a863-4a04-b96d-125b29ac88c4"}: {}
      f:spec:
        .: {}
        f:backupStorageLocation: {}
        f:csiSnapshot:
          .: {}
          f:snapshotClass: {}
          f:storageClass: {}
          f:volumeSnapshot: {}
        f:operationTimeout: {}
        f:snapshotType: {}
        f:sourceNamespace: {}
        f:sourcePVC: {}
      f:status:
        .: {}
        f:progress: {}
    manager: velero-plugin-for-csi
    operation: Update
    time: "2023-08-22T10:56:04Z"
  - apiVersion: velero.io/v2alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        f:cancel: {}
      f:status:
        f:message: {}
        f:node: {}
        f:phase: {}
[...]

PR #6914, opened 2023-10-03
Fix issue 6913: Velero Built-in Datamover: Backup stucks in phase WaitingForPluginOperations when Node Agent pod gets restarted

shubham-pampattiwar:

Adds changes proposed here: https://github.com/vmware-tanzu/velero/issues/6913#issuecomment-1745734358
Fixes https://github.com/vmware-tanzu/velero/issues/6913

### DIFF
--- a/pkg/controller/data_download_controller.go
+++ b/pkg/controller/data_download_controller.go
@@ -287,12 +287,13 @@ func (r *DataDownloadReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	} else if dd.Status.Phase == velerov2alpha1api.DataDownloadPhaseInProgress {
 		log.Info("Data download is in progress")
 		if dd.Spec.Cancel {
+			log.Info("Data download is being canceled")
 			fsRestore := r.dataPathMgr.GetAsyncBR(dd.Name)
 			if fsRestore == nil {
+				r.OnDataDownloadCancelled(ctx, dd.GetNamespace(), dd.GetName())
 				return ctrl.Result{}, nil
 			}
 
-			log.Info("Data download is being canceled")
 			// Update status to Canceling.
 			original := dd.DeepCopy()
 			dd.Status.Phase = velerov2alpha1api.DataDownloadPhaseCanceling
@@ -300,7 +301,6 @@ func (r *DataDownloadReconciler) Reconcile(ctx context.Context, req ctrl.Request
 				log.WithError(err).Error("error updating data download status")
 				return ctrl.Result{}, err
 			}
-
 			fsRestore.Cancel()
 			return ctrl.Result{}, nil
 		}
--- a/pkg/controller/data_upload_controller.go
+++ b/pkg/controller/data_upload_controller.go
@@ -292,13 +292,15 @@ func (r *DataUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	} else if du.Status.Phase == velerov2alpha1api.DataUploadPhaseInProgress {
 		log.Info("Data upload is in progress")
 		if du.Spec.Cancel {
-			fsBackup := r.dataPathMgr.GetAsyncBR(du.Name)
-			if fsBackup == nil {
-				return ctrl.Result{}, nil
-			}
 			log.Info("Data upload is being canceled")
 
-			// Update status to Canceling.
+			fsBackup := r.dataPathMgr.GetAsyncBR(du.Name)
+			if fsBackup == nil {
+				r.OnDataUploadCancelled(ctx, du.GetNamespace(), du.GetName())
+				return ctrl.Result{}, nil
+			}
+
+			// Update status to Canceling
 			original := du.DeepCopy()
 			du.Status.Phase = velerov2alpha1api.DataUploadPhaseCanceling
 			if err := r.client.Patch(ctx, du, client.MergeFrom(original)); err != nil {
--- a/changelogs/unreleased/6914-shubham-pampattiwar
+++ b/changelogs/unreleased/6914-shubham-pampattiwar
@@ -1,0 +1,1 @@
+Fix issue 6913: Velero Built-in Datamover: Backup stucks in phase WaitingForPluginOperations when Node Agent pod gets restarted


sseago (approved):

pkg/apis/velero/v2alpha1/data_upload_types.go
```diff
@@ -90,7 +90,6 @@ const (
 	DataUploadPhaseAccepted   DataUploadPhase = "Accepted"
 	DataUploadPhasePrepared   DataUploadPhase = "Prepared"
 	DataUploadPhaseInProgress DataUploadPhase = "InProgress"
-	DataUploadPhaseCanceling  DataUploadPhase = "Canceling"
```
kaovilai, line 93: > Canceling: the data upload is being canceled
Canceled: the data upload has been canceled

[From design](https://github.com/vmware-tanzu/velero/blob/main/design/volume-snapshot-data-movement/volume-snapshot-data-movement.md)

Given that there could be more data movers in the future using Data Upload CRs. Do we wanna consider a case where a data mover can take longer to cancel something? We need a way to distinguish between actually cancelled, and an "in-progress towards cancellation".

We should keep this const for future data movers IMO.

kaovilai (changes requested):
Would be cool to get

sseago:
@kaovilai The DU/DD controllers don't do anything with this phase, so I'm not sure what value there is in defining it as valid.

pkg/apis/velero/v2alpha1/data_upload_types.go
sseago: @kaovilai @shubham-pampattiwar If the DU/DD controllers never reference this phase, then wouldn't some external mover setting it break the workflow?

shubham-pampattiwar: @sseago won't the external datamover controllers work on the DU CRs ?

sseago: Hmm. I'm not 100% clear on exactly how that will work, but in any case, shouldn't there be a common workflow/state model? Open to suggestion here, though -- lets see what other maintainers say.

sseago (approved):

pkg/controller/data_upload_controller.go
```diff
@@ -292,19 +292,20 @@ func (r *DataUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-			fsBackup := r.dataPathMgr.GetAsyncBR(du.Name)
-			if fsBackup == nil {
-				return ctrl.Result{}, nil
-			}
 			log.Info("Data upload is being canceled")
 
-			// Update status to Canceling.
+			// Update status to Canceled
 			original := du.DeepCopy()
-			du.Status.Phase = velerov2alpha1api.DataUploadPhaseCanceling
```
Lyndon-Li, line 303: There are two cases:
1. If the dataPath is still there (`fsBackup` is not nil), we cannot update the DUCR to cancelled directly. Instead, we need to notify the dataPath (by calling `fsBackup.Cancel()`), wait for it to cancel the data transfer and then it will call back `OnDataUploadCancelled`
2. If the dataPath is not there, we can cancel the DUCR in the main flow

Both of the cases must be supported. However:

- For the existing code, case 2 is not supported
- For the changes in this PR, case 1 will not be supported

So please reconstruct the code to support the both case

pkg/controller/data_upload_controller.go
```diff
@@ -292,19 +292,20 @@ func (r *DataUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			du.Status.Phase = velerov2alpha1api.DataUploadPhaseCanceled
```
Lyndon-Li, line 299: For case 2, instead of patching the DUCR directly here, let's call `OnDataUploadCancelled` instead, it does something more:

- Update the completionTimeStamp
- Cleanup intermediate object

Lyndon-Li:
Please also rebase the code, the codespell error will disappear.

Lyndon-Li:
@shubham-pampattiwar As discussed, please also help to take care of the DataDownload part.

shubham-pampattiwar:
@Lyndon-Li @sseago Added fix for data download controller too, PTAL, Thanks !

pkg/controller/data_upload_controller.go
sseago: @shubham-pampattiwar I think the updated version still doesn't handle case 2.
When there is no fsbackup, then we set Canceled directly, but when there is, we need to cancel the fsbackup rather than just setting the phase.

Lyndon-Li: @shubham-pampattiwar 
For 1, we need to call `fsBackup.Cancel()` and not set `du.Status.Phase` (the same as the original code).
For 2, we just need to call `OnDataUploadCancelled`

shubham-pampattiwar:
@Lyndon-Li Updated the PR, Please take another look.

pkg/controller/data_download_controller.go
```diff
@@ -289,18 +289,17 @@ func (r *DataDownloadReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		if dd.Spec.Cancel {
 			fsRestore := r.dataPathMgr.GetAsyncBR(dd.Name)
 			if fsRestore == nil {
+				r.OnDataDownloadCancelled(ctx, dd.GetNamespace(), dd.GetName())
 				return ctrl.Result{}, nil
 			}
-
 			log.Info("Data download is being canceled")
```
mateusoliveira43, line 295: move this to line 290, to be similar to Data upload case?

pkg/controller/data_download_controller.go
```diff
@@ -289,18 +289,17 @@ func (r *DataDownloadReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-			// Update status to Canceling.
+			// Update status to Canceled.
 			original := dd.DeepCopy()
-			dd.Status.Phase = velerov2alpha1api.DataDownloadPhaseCanceling
+			dd.Status.Phase = velerov2alpha1api.DataDownloadPhaseCanceled
```
Lyndon-Li, line 298: Before `fsRestore` cancels itself, we should not set DD to `DataDownloadPhaseCanceled`. We need to wait `fsRestore` to cancel itself and then it calls `OnDataDownloadCancelled`. 

So just rollback to the original code for this case.

pkg/controller/data_download_controller.go
```diff
@@ -289,18 +289,17 @@ func (r *DataDownloadReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		if dd.Spec.Cancel {
 			fsRestore := r.dataPathMgr.GetAsyncBR(dd.Name)
 			if fsRestore == nil {
+				r.OnDataDownloadCancelled(ctx, dd.GetNamespace(), dd.GetName())
```
Lyndon-Li, line 292: Let's add a log here.

pkg/controller/data_upload_controller.go
```diff
@@ -292,19 +292,22 @@ func (r *DataUploadReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				r.OnDataUploadCancelled(ctx, du.GetNamespace(), du.GetName())
 				return ctrl.Result{}, nil
 			}
-			log.Info("Data upload is being canceled")
 
-			// Update status to Canceling.
+			// Update status to Canceled
 			original := du.DeepCopy()
-			du.Status.Phase = velerov2alpha1api.DataUploadPhaseCanceling
+			du.Status.Phase = velerov2alpha1api.DataUploadPhaseCanceled
```
Lyndon-Li, line 305: The same as DD, rollback to the original code and wait `fsBackup` to cancel itself

pkg/controller/data_download_controller.go
shubham-pampattiwar: Re-positioned the earlier log

pkg/controller/data_download_controller.go
shubham-pampattiwar: Done, Updated the PR.

pkg/controller/data_upload_controller.go
shubham-pampattiwar: Done

Lyndon-Li (approved):

qiuming-best (approved):

sseago (approved):

PR #6797, opened 2023-09-08
Proposal to support JSON Merge Patch and Strategic Merge Patch in Resource Modifiers

27149chen:
# Please add a summary of your change

Proposal to support JSON Merge Patch and Strategic Merge Patch in Resource Modifiers

# Does your change fix a particular issue?

Fixes #6728

# Please indicate you've done the following:

- [x] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [x] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required` as a comment on this pull request.
- [ ] Updated the corresponding documentation in `site/content/docs/main`.

Issue #6801, opened 2023-09-11
resource-modifier-configmap not applied


mailbox171:
I am applying a restore from a backup, mapping a ns onto a different ns.
Everything works but the label I want to remove is not removed.

No mention of it in the logs.
Some syntax problems were solved, I am on 1.12 rc1 so the syntax is still "groupKind" (not groupResource).

version: v1
resourceModifierRules:
- conditions:
    groupKind: secrets
    resourceNameRegex: ".*"
    namespaces:
    - dev-composer-artifactory
  patches:
  - operation: remove
    path: "/metadata/labels/argocd.argoproj.io~1instance"

Now I don't get any error, but the label I want to remove is always there: "argocd.argoproj.io/instance=artifactory-dev".
The restored resource is the following

$ kubectl describe secret artifactory-postgresql-cggh4d6k25 -n af03

Name:         artifactory-postgresql-cggh4d6k25
Namespace:    af03
Labels:   app.kubernetes.io/component=artifactory
              app.kubernetes.io/instance=artifactory-001
              app.kubernetes.io/name=artifactory
              app.kubernetes.io/part-of=artifactory
              app.kubernetes.io/version=1.0.0
              argocd.argoproj.io/instance=artifactory-dev
              environment=development
              tier=artifactory
              velero.io/backup-name=daily-at11-ret84-20230909110058
              velero.io/restore-name=restore-artifactory-22
              
              
[bundle-2023-09-11-17-29-53.tar.gz](https://github.com/vmware-tanzu/velero/files/12578009/bundle-2023-09-11-17-29-53.tar.gz)

 Thanks a lot.

Lyndon-Li:
@anshulahuja98 Please help to check this issue for resource modifier. Thanks!

anshulahuja98:
@mailbox171 can you perhaps put the configmap in a `code block`  so that it's clear ?
For example i am not sure if this part is correct `oproj.io~1insta`

mailbox171:
Sure, you are right.
The "~1" is equivalent to "/" escaped.
Since the label I am deleting is "argocd.argoproj.io/instance"

```
version: v1
resourceModifierRules:
- conditions:
    groupKind: secrets
    resourceNameRegex: ".*"
    namespaces:
    - dev-composer-artifactory
  patches:
  - operation: remove
    path: "/metadata/labels/argocd.argoproj.io~1instance"
```

One more thing.
[...]

PR #6917, opened 2023-10-04
support JSON Merge Patch and Strategic Merge Patch in Resource Modifiers

### LINKED ISSUE #6728: Resource Modifier should support json merge patch and strategic merge patch
Resource Modifier uses json patch to patch object, but  json patch is not easy to use and has limitations.
For example:
1. if I want to remove an annotation not matter it exists or not, a json merge patch is better (`{"metadata":{"annotations":{"foo":null}}}`)
2. if I want to update a container image by container name, a strategic merge patch should be used

support json merge patch and strategic merge patch in Resource Modifier

**Environment:**

**Vote on this issue!**

27149chen:

Support JSON Merge Patch and Strategic Merge Patch in Resource Modifiers, implement #6797 

Fixes #6728


go.mod
```diff
@@ -58,6 +58,7 @@ require (
 	k8s.io/metrics v0.25.6
 	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed
 	sigs.k8s.io/controller-runtime v0.12.2
+	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2
```
kaovilai, line 61: Can we update this too?
https://pkg.go.dev/sigs.k8s.io/json?tab=versions

Latest seems to be `
v0.0.0-20221116044647-bc3834ca7abd`

27149chen: done

internal/resourcemodifiers/resource_modifiers_test.go
```diff
@@ -9,6 +9,10 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 )
 
 func TestGetResourceModifiersFromConfig(t *testing.T) {
```
anshulahuja98, line 18: can you add deserialization tests for new contracts?

anshulahuja98 (changes requested):

internal/resourcemodifiers/resource_modifiers.go
```diff
@@ -23,23 +23,20 @@ const (
...
-	Operation string `json:"operation"`
-	From      string `json:"from,omitempty"`
-	Path      string `json:"path"`
-	Value     string `json:"value,omitempty"`
-}
-
 type Conditions struct {
 	Namespaces        []string              `json:"namespaces,omitempty"`
 	GroupResource     string                `json:"groupResource"`
 	ResourceNameRegex string                `json:"resourceNameRegex,omitempty"`
```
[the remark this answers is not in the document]
27149chen: resourceName is an arbitrary string but groupResource is resource.group, also I want to align with the resource filters in Backup and Restore

internal/resourcemodifiers/resource_modifiers.go
```diff
@@ -172,25 +162,27 @@ func unmarshalResourceModifiers(yamlData []byte) (*ResourceModifiers, error) {
...
-	// if value is a json object or array, then don't add quotes.
-	if strings.HasPrefix(value, "{") || strings.HasPrefix(value, "[") {
-		return false
+type patcher interface {
+	Patch(u *unstructured.Unstructured, logger logrus.FieldLogger) (*unstructured.Unstructured, error)
+}
+
+func (r *ResourceModifierRule) applyPatch(u *unstructured.Unstructured, scheme *runtime.Scheme, logger logrus.FieldLogger) error {
+	var p patcher
+	if len(r.Patches) > 0 {
```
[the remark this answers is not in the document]
27149chen: validation added. It is too complex to support multiple types, for example, how to handle patch order? I don't think it is necessary to support it, at least for now.

internal/resourcemodifiers/resource_modifiers_test.go
27149chen: what is the perpose of the deserialization tests? we just call `yaml.UnmarshalStrict` to convert the yaml manifest to go struct.

internal/resourcemodifiers/resource_modifiers.go
```diff
@@ -110,57 +113,44 @@ func (r *ResourceModifierRule) Apply(obj *unstructured.Unstructured, groupResour
 		}
 	}
 
-	patches, err := r.PatchArrayToByteArray()
+	match, err := matchConditions(obj, r.Conditions.Matches, log)
```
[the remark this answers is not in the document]
27149chen: ok, fixed

internal/resourcemodifiers/resource_modifiers_test.go
anshulahuja98: Just to ensure that there are no missed fields,  for example fields that are not annotated properly with `json` tags.  Or any missed nullable fields which might have been made compulsory by mistake or vica versa.
And all in all it can give a good payload for using in documentation.

internal/resourcemodifiers/resource_modifiers.go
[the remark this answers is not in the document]
anshulahuja98: Okay, if we have precedence in resource filters it should be fine..

internal/resourcemodifiers/resource_modifiers.go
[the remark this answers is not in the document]
anshulahuja98: if you have added validation then it is fine.
Good enough for now.

anshulahuja98:
@27149chen can you also please get patch code coverage to 60%

internal/resourcemodifiers/resource_modifiers_test.go
27149chen: added

27149chen:
> @27149chen can you also please get patch code coverage to 60%

done

design/merge-patch-and-strategic-in-resource-modifier.md
```diff
@@ -68,17 +68,12 @@ resourceModifierRules:
...
-          }
-        }
-      }
+  - patchData:
+      metadata:
+        annotations:
+          foo: null
 ```
 - The above configmap will apply the Merge Patch to all the pods in namespace ns1 and remove the annotation `foo` from the pods.
-- Both json and yaml format are supported for the patchData.
```
anshulahuja98, line 81: oh what caused this to change?
I thought we discussed this support in the library ?

design/merge-patch-and-strategic-in-resource-modifier.md
```diff
@@ -88,25 +83,18 @@ Example of StrategicPatches in ResourceModifierRule
...
-  groupResource: pods
-  resourceNameRegex: "^my-pod$"
-  namespaces:
-  - ns1
+    groupResource: pods
+    resourceNameRegex: "^my-pod$"
+    namespaces:
+    - ns1
   strategicPatches:
-  - patchData: |
```
anshulahuja98, line 96: don't we need this for yaml as well?

anshulahuja98 (changes requested):

design/merge-patch-and-strategic-in-resource-modifier.md
anshulahuja98: We should somehow support JSON since it's a jsonpatch based support and people would prefer to use their kubectl json setup to here as is.

anshulahuja98 (changes requested):

design/merge-patch-and-strategic-in-resource-modifier.md
27149chen: Both json and yaml format are supported inside the patch logic, but there is an issue: we can not mix json and yaml in one menifest. So if the RM configMap is json, patchData here can also be json.
I can update the design to support json format for RM configmap, what do you think?

design/merge-patch-and-strategic-in-resource-modifier.md
27149chen: yes

design/merge-patch-and-strategic-in-resource-modifier.md
anshulahuja98: Didn't fully understand what you meant by this...

design/merge-patch-and-strategic-in-resource-modifier.md
anshulahuja98: I meant the "|"
please add it back

internal/resourcemodifiers/resource_modifiers.go
```diff
@@ -95,13 +96,19 @@ func (p *ResourceModifiers) ApplyResourceModifierRules(obj *unstructured.Unstruc
...
 
-func (r *ResourceModifierRule) Apply(obj *unstructured.Unstructured, groupResource string, log logrus.FieldLogger) error {
+func (r *ResourceModifierRule) apply(obj *unstructured.Unstructured, groupResource string, scheme *runtime.Scheme, log logrus.FieldLogger) error {
 	namespaceInclusion := collections.NewIncludesExcludes().Includes(r.Conditions.Namespaces...)
 	if !namespaceInclusion.ShouldInclude(obj.GetNamespace()) {
 		return nil
 	}
 
-	if r.Conditions.GroupResource != groupResource {
+	g, err := glob.Compile(r.Conditions.GroupResource)
```
anshulahuja98, line 105: one thing came in my mind
if we support globbing with namespace filter.

Scenario -> user provided NS filter, with path glob as blah*
Now blah* can contain both cluster scope and namespaced scope CRs. 
We need to make it clear in docs that if user wants to do cluster scope they have to put NS filter as empty otherwise it won't work
and using it in conjunction with GroupResource glob path will have these further nuances.

anshulahuja98 (changes requested):

design/merge-patch-and-strategic-in-resource-modifier.md
27149chen: which means if the ResourceModifiers is in yaml format, patchData must be yaml too, if the ResourceModifiers is json, patchData must be json.

but there is an exception:
if you build the ResourceModifiers configMap programmatically, patchData can be either json or yaml, because we you call `yaml.Marshal` to convert ResourceModifiers from go object to yaml, patchData will always be convert to yaml. Also, if you call `json.Marshal` to convert ResourceModifiers from go object to json, patchData will always be convert to json.

design/merge-patch-and-strategic-in-resource-modifier.md
anshulahuja98: I am wondering if we make the field bytearray wouldn't it be possible to support either json or yaml here?
Irrespective of if the RM configmap is built as a json or a yaml
If yes I would prefer that route.

design/merge-patch-and-strategic-in-resource-modifier.md
anshulahuja98: Replied to below comment in continuation to this.

design/merge-patch-and-strategic-in-resource-modifier.md
27149chen: RM configmap yaml Marshal will fail if you use []byte

internal/resourcemodifiers/resource_modifiers.go
anshulahuja98: yeah
again the limitation is cluster scope won't be taken care of

27149chen: sorry, not quite understand, can you give an example?

design/merge-patch-and-strategic-in-resource-modifier.md
ywk253100: But Velero doesn't support `ResourceModifier` in JSON format at this moment, right?

I guess we cannot use the `yaml.Unmarshal()` functions directly in order to support both JSON and YAML formats, we need to do some decoding/encoding by ourselves according to different formats before calling `yaml.Unmarshal()`.

27149chen: thank to @anshulahuja98 I changed the type of patchData to string, and now we can support both json and yaml patchData

internal/resourcemodifiers/resource_modifiers.go
anshulahuja98: user provided NS filter (default), with **GroupResource** glob as blah*
Now blah* can contain both cluster scope and namespaced scope CRs.
We need to make it clear in docs that if user wants to do cluster scope they have to put NS filter as empty otherwise it won't work
and using it in conjunction with GroupResource glob path will have these further nuances.

Reworded my initial comment to make it more clear

internal/resourcemodifiers/resource_modifiers_test.go
```diff
@@ -189,7 +341,7 @@ func TestGetResourceModifiersFromConfig(t *testing.T) {
 				return
 			}
 			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("GetResourceModifiersFromConfig() = %v, want %v", got, tt.want)
+				t.Errorf("GetResourceModifiersFromConfig() = %#v, want %#v", got, tt.want)
```
anshulahuja98, line 344: I don't think we should need # here...
```suggestion
				t.Errorf("GetResourceModifiersFromConfig() = %#v, want %#v", got, tt.want)
```

internal/resourcemodifiers/resource_modifiers_test.go
```diff
@@ -719,14 +871,641 @@ func TestResourceModifiers_ApplyResourceModifierRules(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := tt.rm.ApplyResourceModifierRules(tt.obj, tt.groupResource, nil, logrus.New())
+
+			assert.Equal(t, tt.wantErr, len(got) > 0)
+			assert.Equal(t, *tt.wantObj, *tt.obj)
+		})
+	}
+}
```
anshulahuja98, line 1507: the tests look great overall.
Can you perhaps add test case for once scenario - we recieved this in a User Created issue.

WHere we use the serialized option for / using ~1
Look at this comment for example: https://github.com/vmware-tanzu/velero/issues/6801#issuecomment-1715218885

anshulahuja98 (changes requested):

anshulahuja98:
@ywk253100 I am done with my review 
Some minor comments are pending fix - you can review from your end.

internal/resourcemodifiers/resource_modifiers_test.go
```diff
@@ -131,6 +135,122 @@ func TestGetResourceModifiersFromConfig(t *testing.T) {
...
+
+	rules8 := &ResourceModifiers{
+		Version: "v1",
+		ResourceModifierRules: []ResourceModifierRule{
+			{
+				Conditions: Conditions{
+					GroupResource: "pods",
+					Namespaces: []string{
+						"ns1",
+					},
```
anshulahuja98, line 243: can you add the `matches` field in any of the deserialization tests you wrote.

anshulahuja98 (changes requested):

ywk253100:
I'm good for the changes. 
@27149chen Could you also update the user guide about the update?

internal/resourcemodifiers/resource_modifiers.go
```diff
@@ -125,87 +132,78 @@ func (r *ResourceModifierRule) Apply(obj *unstructured.Unstructured, groupResour
...
-}
 
-func (p *JSONPatch) ToString() string {
-	if addQuotes(p.Value) {
-		return fmt.Sprintf(`{"op": "%s", "from": "%s", "path": "%s", "value": "%s"}`, p.Operation, p.From, p.Path, p.Value)
+	var fixed []JSONPatch
+	for _, patch := range patches {
+		fixed = append(fixed, JSONPatch{
+			Operation: "test",
+			Path:      patch.Path,
```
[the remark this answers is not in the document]
27149chen: done

internal/resourcemodifiers/resource_modifiers_test.go
27149chen: added a case replacing a "/" to "~1", but failed, so I removed, it is not releated to this pr

27149chen:
> I'm good for the changes. @27149chen Could you also update the user guide about the update?

done

site/content/docs/main/restore-resource-modifiers.md
```diff
@@ -105,3 +105,81 @@ resourceModifierRules:
 -  Update a container's image using a json patch with positional arrays
 kubectl patch pod valid-pod -type='json' -p='[{"op": "replace", "path": "/spec/containers/0/image", "value":"new image"}]'
 - Before creating the resource modifier yaml, you can try it out using kubectl patch command. The same commands should work as it is.
+
+### New features introduced in 1.13
```
anshulahuja98, line 109: You don't need to add this header.

internal/resourcemodifiers/resource_modifiers.go
anshulahuja98: Hey I am not sure if ignoring ns filter is the right way
I am wondering if we should add another flag which says `includeClusterSCope` similar to what we have in backup/restore APIs.

site/content/docs/main/restore-resource-modifiers.md
```diff
@@ -105,3 +105,81 @@ resourceModifierRules:
...
+          "annotations": {
+            "foo": null
+          }
+        }
+      }
+```
+- The above configmap will apply the Merge Patch to all the pods in namespace ns1 and remove the annotation `foo` from the pods.
+- Both json and yaml format are supported for the patchData.
+
+#### Strategic Merge Patch
```
anshulahuja98, line 133: we should perhaps move these 2 sections above

this is currently below jsonpatch examples and adbacned scenarios.

We should probably refactor this whole doc to make things more clear.

site/content/docs/main/restore-resource-modifiers.md
```diff
@@ -105,3 +105,81 @@ resourceModifierRules:
...
+  - patchData: |
+      {
+        "metadata": {
+          "annotations": {
+            "foo": null
+          }
+        }
+      }
+```
+- The above configmap will apply the Merge Patch to all the pods in namespace ns1 and remove the annotation `foo` from the pods.
```
anshulahuja98, line 130: add links to kubectl merge patch docs / other docs for reference. YOU can even provide sample kubectl command for use to test out in local setup before creating CM.

site/content/docs/main/restore-resource-modifiers.md
```diff
@@ -105,3 +105,81 @@ resourceModifierRules:
...
+- Both json and yaml format are supported for the patchData.
+
+#### Strategic Merge Patch
+you can modify a resource using Strategic Merge Patch
```
anshulahuja98, line 134: nit: capitalize first letter

site/content/docs/main/restore-resource-modifiers.md
```diff
@@ -105,3 +105,81 @@ resourceModifierRules:
...
+            }
+          ]
+        }
+      }
+```
+- The above configmap will apply the Strategic Merge Patch to the pod with name my-pod in namespace ns1 and update the image of container nginx to `repo2/nginx`.
+- Both json and yaml format are supported for the patchData.
+
+### Conditional Patches in ALL Patch Types
+Since JSON Merge Patch and Strategic Merge Patch do not support conditional patches, we use the `test` operation of JSON Patch to support conditional patches in all patch types by adding it to `Conditions` struct in `ResourceModifierRule`.
```
anshulahuja98, line 160: let's not put internal details of using `test`.
Just say that it is a conditional check on value at a path - something like that.

site/content/docs/main/restore-resource-modifiers.md
```diff
@@ -105,3 +105,81 @@ resourceModifierRules:
...
+
+Example of test in conditions
```
anshulahuja98, line 162: again don't callout `test`
Call it matches or something else.

anshulahuja98 (changes requested):

site/content/docs/main/restore-resource-modifiers.md
27149chen: added doc link

internal/resourcemodifiers/resource_modifiers.go
27149chen: I think it's ok, ns filter is used to filter namespaced resources, cluster resources can not be filtered by ns, so we can just ignore it if both namespaced resources and cluster resources are included.

for example:
```
groupResource: *
namespaces:
  - ns1
```
this will match all namespaced resources in ns1 and all cluster resources

anshulahuja98:
https://github.com/vmware-tanzu/velero/pull/6917#discussion_r1371250176

Hey I am not sure if ignoring ns filter is the right way
I am wondering if we should add another flag which says includeClusterSCope similar to what we have in backup/restore APIs.

@27149chen

internal/resourcemodifiers/resource_modifiers.go
anshulahuja98: but the issue is
we are doing cluster scope by default. Which is not the right behavior. 
We should give user all controls as possible. 


For now my suggestion would be revert the current chagne you did 

and behaviour should be as following

 -> NS field empty means both cluster scope and NS scoped will be changed
 
 NS field specified means only NS scoped. 
 
 IF a use wishes to do both cluster scope and specific NS scope they can use multiple resource modifier rules
 
 For Example -> The first rule can be
 GroupResource - provider.clusterscoperesources.*
 namespace 
- ns1

Second rule

GroupResource - provider.*
namespace (blank)


My request would be

1. to document this behaviour of NS field empty = cluster scope all + all NS 
And revert the change you have done for now - which always includes cluster scope.

if this suggestion doesn't make sense 

2. the altnerative would be to introduce new clsuterSCope boolean field for clarity 


i would suggest to go with approach 1 for now.

internal/resourcemodifiers/resource_modifiers_test.go
anshulahuja98: okay.

internal/resourcemodifiers/resource_modifiers.go
27149chen: if the user needs multiple rm rules to do one thing, it is too redundant.
Maybe we need a REAL resource fileters here, to support all conditions (for example namespaced resource a1 of A + cluster resource b2 of B ), but it is too complex and will break the compatibility.
My idea is to keep it simple, I think including the cluster resources by default is reasonable

anshulahuja98: " I think including the cluster resources by default is reasonable"
I feel this would lead to changes which are not intended.
I am worried about this.

In terms of REAL filters here - similar to what we have in backup /restore - includeClusterSCopeResources
We can add a bool for that. 
that would make it consistent and give control to user.

anshulahuja98: This is only useful for globbed patterns
but would lead to confusion if explicit GroupResource is mentioned.

anshulahuja98: On further thought, I think it's fine.
But PLEASE just add this behaviour of cluster scope by default to docuemtnation.

anshulahuja98: Just to reiterate, I am fine with current changes you have. Just that I request you to document the behaviour of cluster scope by default in case of globbed paths.

27149chen: thanks, updated

anshulahuja98 (approved):

ywk253100 (approved):

PR #6918, opened 2023-10-04
Add WaitForReady flag to check container readiness state before exec a hook

Ripolin:

Before running a post restore hook, Velero test the targeted container to check if it's running or not. To me it should be better to check if targeted container is ready to run restore commands.


internal/hook/wait_exec_hook_handler.go
```diff
@@ -126,7 +126,7 @@ func (e *DefaultWaitExecHookHandler) HandleHooks(
 		}
 
 		for containerName, hooks := range byContainer {
-			if !isContainerRunning(newPod, containerName) {
+			if !isContainerReady(newPod, containerName) {
 				podLog.Infof("Container %s is not running: post-restore hooks will not yet be executed", containerName)
```
mateusoliveira43, line 130: also change log message from **running** to **ready**?

Ripolin: message updated in 6593ee44105ea9952db2b1a410ba0b9ae7982116

kaovilai (changes requested):
Need changelog file
> [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added /kind changelog-not-required as a comment on this pull request.

kaovilai:
I think it is possible for someone to have a workload such that it isn't ready until the hook runs to do something such as adding data.

Perhaps a new kind of hook could be reasonable.. `prehookready`? thoughts?

kaovilai:
ie. this change can break an existing workload that is only ready by manually running a script when prehook is run on restore.

sseago:
@kaovilai @Ripolin or perhaps instead of a new hook type, a new bool field added to `ExecRestoreHook` -- something like `WaitForReady` and then the logic could wait for running vs. ready depending on this field.

https://github.com/vmware-tanzu/velero/blob/main/pkg/apis/velero/v1/restore_types.go#L194

shubham-pampattiwar:
Having something like `WaitForReady` in `ExecRestoreHook` API would make sense I believe.

Ripolin:
> Having something like `WaitForReady` in `ExecRestoreHook` API would make sense I believe.

Like that eb1ff04bdd9d3647849a8f61698c7626ead9e686 ?

sseago:
@Ripolin Yes, I was thinking something like that.  A couple things, though:
1. We'd need to update the restore hooks documentation to show this new field: https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/main/restore-hooks.md#exec-restore-hooks
2. Also, note that restore hooks can be specified using annotations, so we'd probably want to handle this field there as well.

sseago:
Since there can be multiple restore hooks, it looks like setting `WaitForReady` to `true` for only one hook has same effect as setting it for all -- since velero will wait until the container is ready for all hooks to run. I think this is probably fine as long as it's documented -- @shubham-pampattiwar what do you think?

Ripolin:
> @Ripolin Yes, I was thinking something like that. A couple things, though:
> 
> 1. We'd need to update the restore hooks documentation to show this new field: https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/main/restore-hooks.md#exec-restore-hooks
> 2. Also, note that restore hooks can be specified using annotations, so we'd probably want to handle this field there as well.

Done both

shubham-pampattiwar:
@sseago yes I agree, the new behavior needs to be documented.

site/content/docs/main/restore-hooks.md
```diff
@@ -181,6 +181,8 @@ Below are the annotations that can be added to a pod to specify exec restore hoo
     * How long to wait once execution begins. Defaults is 30 seconds. Optional.
 * `post.hook.restore.velero.io/wait-timeout`
     * How long to wait for a container to become ready. This should be long enough for the container to start plus any preceding hooks in the same container to complete. The wait timeout begins when the container is restored and may require time for the image to pull and volumes to mount. If not set the restore will wait indefinitely. Optional.
+* `post.hook.restore.velero.io/wait-for-ready`
```
shubham-pampattiwar, line 184: s/launch/launched

pkg/apis/velero/v1/restore_types.go
```diff
@@ -214,6 +214,11 @@ type ExecRestoreHook struct {
 	// before attempting to run the command.
 	// +optional
 	WaitTimeout metav1.Duration `json:"waitTimeout,omitempty"`
+
+	// WaitForReady ensure command will be launch when container is Ready instead of Running.
```
shubham-pampattiwar, line 218: s/ensure/ensures
s/launch/launched

site/content/docs/main/restore-hooks.md
```diff
@@ -181,6 +181,8 @@ Below are the annotations that can be added to a pod to specify exec restore hoo
...
+    * String representation of a boolean that ensure command will be launch when underlying container is fully [ready](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes). Any value except "true" will be considered as "false". Defaults is false. Optional.
```
shubham-pampattiwar, line 185: Please something on similar lines like "Even if one restore hook consists of `WaitForReady` flag as true, all the other hook execution will wait for `Ready` state.

shubham-pampattiwar (changes requested):

site/content/docs/main/restore-hooks.md
Ripolin: String representation of a boolean that ensure command will be launched when underlying container is fully [Ready](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes). Use with caution because if only one restore hook consists of `WaitForReady` flag as "true", all the other hook executions will wait for `Ready` state too. Any value except "true" will be considered as "false". Defaults is false. Optional.

... good for you @shubham-pampattiwar ?

changelogs/unreleased/6918-Ripolin
```diff
@@ -0,0 +1 @@
+Check if targeted container is ready before executing a hook
```
kaovilai, line 1:
```suggestion
Optional check if targeted container is ready before executing a hook
```

internal/hook/item_hook_handler.go
```diff
@@ -477,12 +479,19 @@ func getPodExecRestoreHookFromAnnotations(annotations map[string]string, log log
 		}
 	}
 
+	waitForReadyString := annotations[podRestoreHookWaitForReadyAnnotationKey]
+	waitForReady := boolptr.False()
+	if waitForReadyString == "true" {
+		waitForReady = boolptr.True()
+	}
```
kaovilai, line 486:
```suggestion
	waitForReady := boolptr.False()
	var err error
	*waitForReady, err = strconv.ParseBool(waitForReadyString)
	if err != nil {
		log.Warn(errors.Wrapf(err, "Unable to parse wait for ready %s, ignoring", waitForReadyString))
	}
```

kaovilai (changes requested):

blackpiglet:
@Ripolin 
Could you help to squash the commits?

site/content/docs/main/restore-hooks.md
```diff
@@ -181,6 +181,8 @@ Below are the annotations that can be added to a pod to specify exec restore hoo
     * How long to wait once execution begins. Defaults is 30 seconds. Optional.
 * `post.hook.restore.velero.io/wait-timeout`
     * How long to wait for a container to become ready. This should be long enough for the container to start plus any preceding hooks in the same container to complete. The wait timeout begins when the container is restored and may require time for the image to pull and volumes to mount. If not set the restore will wait indefinitely. Optional.
+* `post.hook.restore.velero.io/wait-for-ready`
+    * String representation of a boolean that ensure command will be launched when underlying container is fully [Ready](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes). Use with caution because if only one restore hook consists of `WaitForReady` flag as "true", all the other hook executions will wait for `Ready` state too. Any value except "true" will be considered as "false". Defaults is false. Optional.
```
blackpiglet, line 185: May need to mention the limitation of `One hook WaitForReady set as true will impact other hooks` per pod, not per backup or restore.

Ripolin: @blackpiglet What do you think of :

*String representation of a boolean that ensure command will be launched when underlying container is fully [Ready](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes). Use with caution because if only one restore hook for a pod consists of `WaitForReady` flag as "true", all the other hook executions for that pod, whatever their origin (`Backup` or `Restore` CRD), will wait for `Ready` state too. Any value except "true" will be considered as "false". Defaults is false. Optional.*

blackpiglet: Thanks. Looks good.

Ripolin:
> @Ripolin Could you help to squash the commits?

Squash and rebase done !

blackpiglet (approved):

kaovilai (approved):

shubham-pampattiwar (approved):

PR #6949, opened 2023-10-12
Retry Get on IsNotFound after restore Create error on AlreadyExists

### LINKED ISSUE #6952: On restore, object can fail to create from AlreadyExists and fail to Get with IsNotFound without retry
A custom resource can take some time from create to become Get-able due to some processing inside an apiserver resulting in failed restore for the object and a log like so
```
time="2023-09-15T17:51:09Z" level=error msg="Error retrieving cluster version of isim-dev-restored/postgresql:13-el9: imagestreamtags.image.openshift.io \"postgresql:13-el9\" not found" logSource="/remote-source/velero/app/pkg/res
tore/restore.go:1285" restore=velero/1a885c89-d880-5472-ac25-04e823668177-2023-09-15-10-51-25-isim-dev-restored
```
Example case: https://github.com/openshift/openshift-velero-plugin/issues/204
 <details>
<summary>Problem analysis</summary>

```go
// Get retrieves an image that has been tagged by stream and tag. `id` is of the format <stream name>:<tag>.
func (r *REST) Get(ctx context.Context, id string, options *metav1.GetOptions) (runtime.Object, error) {
...
	image, err := r.imageFor(ctx, tag, imageStream) // <-- this is returning an IsNotFoundError
	if err != nil {
		return nil, err
	}
...}
```
https://github.com/openshift/openshift-apiserver/blob/9573998170f3bb7ae7e946c11b7e9fc414120df4/pkg/image/apiserver/registry/imagestreamtag/rest.go#L127C1-L145C2

which calls
```
// imageFor retrieves the most recent image for a tag in a given imageStreem.
func (r *REST) imageFor(ctx context.Context, tag string, imageStream *imageapi.ImageStream) (*imageapi.Image, error) {
	event := internalimageutil.LatestTaggedImage(imageStream, tag)
	if event == nil || len(event.Image) == 0 {
		return nil, kapierrors.NewNotFound(imagegroup.Resource("imagestreamtags"), imageutil.JoinImageStreamTag(imageStream.Name, tag)) // <-- this is where error originated
	}

	return r.imageRegistry.GetImage(ctx, event.Image, &metav1.GetOptions{})
}
```
https://github.com/openshift/openshift-apiserver/blob/9573998170f3bb7ae7e946c11b7e9fc414120df4/pkg/image/apiserver/registry/imagestreamtag/rest.go#L439C1-L447C2

[which is caused by TagEvent not yet existing](https://github.com/openshift/openshift-apiserver/blob/9573998170f3bb7ae7e946c11b7e9fc414120df4/pkg/image/apiserver/internalimageutil/helpers.go#L285C1-L287C47) due to network speeds etc in processing the imagestreamtag creation
```go
// LatestTaggedImage returns the most recent TagEvent for the specified image
// repository and tag. Will resolve lookups for the empty tag. Returns nil
// if tag isn't present in stream.status.tags.
```
</details>

While this is an OpenShift apiserver specific example, the problem can be described and resolved generically and may apply in other environments.

A few retries would resolve this issue.
Implementation: https://github.com/vmware-tanzu/velero/pull/6949

**Environment:**

**Vote on this issue!**

kaovilai:

Fixes #6952


pkg/client/retry.go
```diff
@@ -42,3 +47,25 @@ func CreateRetryGenerateNameWithFunc(obj kbclient.Object, createFn func() error)
...
 }
+
+func GetRetriableWithCacheLister(lister cache.GenericNamespaceLister, name string, retriable func(error) bool) (runtime.Object, error){
+	var clusterObj runtime.Object
+	getFunc := func () error {
+		var err error
+		clusterObj, err = lister.Get(name)
+		return err
+	}
+	err := retry.OnError(retry.DefaultRetry, retriable, getFunc)
```
sseago, line 58: I don't think we want to use DefaultRetry:
```
var DefaultRetry = wait.Backoff]{
	Steps:    5,
	Duration: 10 * time.Millisecond,
	Factor:   1.0,
	Jitter:   0.1,
}
```
This retries 5 times with a constant wait time of 10 ms between waits, so if the controller responsible for making the resource ready takes longer than .05 seconds, we'll still fail.

DefaultBackoff would be better, but if we want to make sure we wait a full minute or two, we may want to define our own Backoff struct to ensure that the last attempt happens after at least a minute or so.

pkg/restore/restore.go
```diff
@@ -1082,6 +1082,21 @@ func (ctx *restoreContext) getResource(groupResource schema.GroupResource, obj *
 	return u, nil
 }
 
+// getResource but with retry on retriable error
+func (ctx *restoreContext) getResourceWithRetries(groupResource schema.GroupResource, obj *unstructured.Unstructured, namespace, name string, retriable func(error) bool) (*unstructured.Unstructured, error) {
```
sseago, line 1086: Rather than repeating this whole func with a different Get call, maybe you could just add the retriable arg to `getResource` and then in that func, if it's nil, call `lister.Get`, and if it's not nil call `cache.GetRetriableWithCacheLister`.

This is a non-exported func only called in one place other than the one you're changing to add retries, so pass nil in for the other case, and retriable in where you're using the new func now.

pkg/client/mocks/Factory.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.30.1. DO NOT EDIT.
+// Code generated by mockery v2.35.4. DO NOT EDIT.
```
mateusoliveira43, line 1: since you did not change `go.mod`, should this file been changed?

sseago: @mateusoliveira43 mockery isn't in the built image (which go.mod reflects) -- mockery is a tool used to generate golang code for mocks.

sseago: This comment is inserted by mockery to indicate what version generated the code.

pkg/client/retry.go
```diff
@@ -18,12 +18,29 @@ package client
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/retry"
 	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// MinuteBackoff is a retry.DefaultBackoff that retries for at least 1 minute (60000ms)
+var MinuteBackoff = func() wait.Backoff {
+	mb := retry.DefaultBackoff
+	// TotalDuration = 0ms + 10ms + 50ms + 250ms + 1250ms + 6250ms + 31250ms + 60000ms = 99,060 ms > 1 minute
```
mateusoliveira43, line 37: instead of saying its greater than 1 minute, what about saying its less than 2 minutes?

pkg/client/mocks/Factory.go
mateusoliveira43: Got it Scott, thanks! We can resolve this conversation then

Just curious: where is mockery in velero repo?

pkg/client/retry.go
sseago: @mateusoliveira43 1 minute is the cap defined here, so that last iteration will happen at 60s rather than something larger than 2 minutes (that 60000ms should actually be 31250*5ms, but again since the last iteration would happen at a time greater than the defined cap of 1 minute, it doesn't matter -- last retry will be 60s after initial call, and if it's still not ready then, the original error message (NotFound) will be returned.

pkg/client/mocks/Factory.go
sseago: It's not in the repo directly, but it's referenced in docs: https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/main/code-standards.md#mocks
example from the docs:
```
go get github.com/vektra/mockery/.../
cd pkg/podvolume
mockery -name=Restorer
```

pkg/client/retry.go
mateusoliveira43: Understood, thanks

pkg/client/retry.go
```diff
@@ -18,12 +18,29 @@ package client
...
 	"k8s.io/client-go/util/retry"
 	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// MinuteBackoff is a retry.DefaultBackoff that retries for at least a minute (60000ms) but no more than 2 minutes (120000ms).
+var MinuteBackoff = func() wait.Backoff {
+	mb := retry.DefaultBackoff
+	// TotalDuration = 0ms + 10ms + 50ms + 250ms + 1250ms + 6250ms + 31250ms + 60000ms = 99,060 ms > 1 minute
+	// 7 steps
+	mb.Steps = 7
```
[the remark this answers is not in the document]
sseago: @kaovilai sure -- and if we're configuring `Cap` anyway, then really this can probably be a much larger number. Step 6 will be 30 seconds after step 5, 7 will be 150 seconds after step 6, 8 will be 750 seconds after step 7, etc. -- but it should stop once it gets to `Cap` in any case.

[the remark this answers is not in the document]
kaovilai: No. Cap here defines how big a step can get.. not the maximum time it will backoff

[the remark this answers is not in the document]
sseago: @kaovilai Oh, you're right. I completely misread that when I was looking at the field-level documentation. I guess there's not a field for max time from beginning, so if we did make it configurable, we'd either need to ask users for #steps, or for a total time, and we'd calculate #steps from there.

kaovilai:
#7004 replaced this.

Issue #6663, opened 2023-08-16
Configurable concurrent number in the same node for fs-backup and data mover


Lyndon-Li:
At present, the concurrent number for VGDP (Velero General Data Path) is hard-coded as 1, which affects the users of VGDP -- fs-backup and data mover. It is valuable to make this configurable by users. This involves:
1. Test and verify that VGDP is working well for concurrent number larger than 1
2. Create a global option for users to configure this number

reasonerjt:
A design will be needed if we wanna set node-level limitation for concurrency.

Lyndon-Li:
The concurrent number should be set on VGDP level, which applies to the pod volume backups and data movers in the same node.

This design will be based on the ALL-IN-ONE policy, we will add below snippet in the policy:
```
  vgcp:
    defaultConcurrentNum: 2
    concurrentNumByNode
      node-xxx: 3
      node-yyy: 8
```

sseago:
@Lyndon-Li so to confirm, the above policy would set 3 for node-xxx, 8 for node-yyy, and 2 for everything else? So this would mean that in cases where we don't need node-specific differences, users would just set `defaultConcurrentNum` and that would apply to all nodes? But if there were certain nodes which were different in size or expected load than the others, they could tweak those particular node's concurrency numbers to reflect that?

Lyndon-Li:
> @Lyndon-Li so to confirm, the above policy would set 3 for node-xxx, 8 for node-yyy, and 2 for everything else? So this would mean that in cases where we don't need node-specific differences, users would just set `defaultConcurrentNum` and that would apply to all nodes? But if there were certain nodes which were different in size or expected load than the others, they could tweak those particular node's concurrency numbers to reflect that?

Yes, this is the preliminary idea. This also means that the configuration will be more complex than what could fit into the node-agent server CLI parameter, so we will need a separate configuration, something like a global node-agent configuration, if the ALL-IN-ONE policy is not going to delivered.

reasonerjt:
The configuration may need more discussion but let's make sure we have a design in v1.13.0

shawn-hurley:
1. if we have node-specific limits, can we please use labels/label selectors for this instead of the node name? I am pretty sure that the node name is also a label, so you could still choose one node based on the label.
2. Can we explain why we need the ability to do this per node?

sseago:
@shawn-hurley "Can we explain why we need the ability to do this per node?"

[...]

PR #6950, opened 2023-10-13
Design for node-agent concurrency

Lyndon-Li:
For issue #6663, Add the design for node-agent concurrency

### DIFF
--- a/design/node-agent-concurrency.md
+++ b/design/node-agent-concurrency.md
@@ -1,0 +1,124 @@
+# Node-agent Concurrency Design
+
+## Glossary & Abbreviation
+
+**Velero Generic Data Path (VGDP)**: VGDP is the collective of modules that is introduced in [Unified Repository design][1]. Velero uses these modules to finish data transfer for various purposes (i.e., PodVolume backup/restore, Volume Snapshot Data Movement). VGDP modules include uploaders and the backup repository.  
+
+## Background
+
+Velero node-agent is a daemonset hosting controllers and VGDP modules to complete the concrete work of backups/restores, i.e., PodVolume backup/restore, Volume Snapshot Data Movement backup/restore.  
+For example, node-agent runs DataUpload controllers to watch DataUpload CRs for Volume Snapshot Data Movement backups, so there is one controller instance in each node. One controller instance takes a DataUpload CR and then launches a VGDP instance, which initializes a uploader instance and the backup repository connection, to finish the data transfer. The VGDP instance runs inside the node-agent pod or in a pod associated to the node-agent pod in the same node.  
+
+Varying from the data size, data complexity, resource availability, VGDP may take a long time and remarkable resources (CPU, memory, network bandwidth, etc.).  
+Technically, VGDP instances are able to run in concurrent regardless of the requesters. For example, a VGDP instance for a PodVolume backup could run in parallel with another VGDP instance for a DataUpload. Then the two VGDP instances share the same resources if they are running in the same node.  
+
+Therefore, in order to gain the optimized performance with the limited resources, it is worthy to configure the concurrent number of VGDP per node. When the resources are sufficient in nodes, users can set a large concurrent number, so as to reduce the backup/restore time; otherwise, the concurrency should be reduced, otherwise, the backup/restore may encounter problems, i.e., time lagging, hang or OOM kill.  
+
+## Goals
+
+- Define the behaviors of concurrent VGDP instances in node-agent
+- Create a mechanism for users to specify the concurrent number of VGDP per node
+
+## Non-Goals
+- VGDP instances from different nodes always run in concurrent since in most common cases the resources are isolated. For special cases that some resources are shared across nodes, there is no support at present
+- In practice, restores run in prioritized scenarios, e.g., disaster recovery. However, the current design doesn't consider this difference, a VGDP instance for a restore is blocked if it reaches to the limit of the concurrency, even though the ones block it are for backups. If users do meet some problems here, they should consider to stop the backups first
+- Sometimes, users wants to totally block backups/restores from running in a specific node, this is out of the scope the current design. To archive this, more modules need to be considered (i.e., expoers of data movers), simply blocking the VGDP (e.g., by setting its concurrent number to 0) doesn't work. E.g., for a fs backup, VGDP instance must run in the node the source pod is running in, if we simply block from VGDP instance, the PodVolumeBackup CR is still submitted but never processed.  
+
+## Solution
+
+We introduce a configMap named ```node-agent-configs``` for users to specify the node-agent related configurations. This configMap is not created by Velero, users should create it manually on demand. The configMap should be in the same namespace where Velero is installed. If multiple Velero instances are installed in different namespaces, there should be one configMap in each namespace which applies to node-agent in that namespace only.  
+Node-agent server checks these configurations at startup time and use it to initiate the related VGDP modules. Therefore, users could edit this configMap any time, but in order to make the changes effective, node-agent server needs to be restarted.  
+The ```node-agent-configs``` configMap may be used for other purpose of configuring node-agent in future, at present, there is only one kind of configuration as the data in the configMap, the name is ```data-path-concurrency```.  
+
+The data structure for ```data-path-concurrency``` is as below:
+```go
+type DataPathConcurrency struct {
+    // GlobalConfig specifies the concurrency number to all nodes for which per-node config is not specified
+    GlobalConfig int `json:"globalConfig,omitempty"`
+
+    // PerNodeConfig specifies the concurrency number to nodes matched by rules
+    PerNodeConfig []RuledConfigs `json:"perNodeConfig,omitempty"`
+}
+
+type RuledConfigs struct {
+    // NodeSelector specifies the label selector to match nodes
+    NodeSelector metav1.LabelSelector `json:"nodeSelector"`
+
+    // Number specifies the number value associated to the matched nodes
+    Number int `json:"number"`
+}
+```
+
+### Global concurrent number
+We allow users to specify a concurrent number that will be applied to all nodes if the per-node number is not specified. This number is set through ```globalConfig``` field in ```data-path-concurrency```.  
+The number starts from 1 which means there is no concurrency, only one instance of VGDP is allowed. There is no roof limit.    
+If this number is not specified or not valid, a hard-coded default value will be used, the value is set to 1. 
+
+### Per-node concurrent number
+We allow users to specify different concurrent number per node, for example, users can set 3 concurrent instances in Node-1, 2 instances in Node-2 and 1 instance in Node-3. This is for below considerations:
+- The resources may be different among nodes. Then users could specify smaller concurrent number for nodes with less resources while larger number for the ones with more resources
+- Help users to isolate critical environments. Users may run some critical workloads in some specified nodes, since VGDP instances may take large resource consumption, users may want to run less number of instances in the nodes with critical workloads
+
+The range of Per-node concurrent number is the same with Global concurrent number.  
+Per-node concurrent number is preferable to Global concurrent number, so it will overwrite the Global concurrent number for that node.  
+
+Per-node concurrent number is implemented through ```perNodeConfig``` field in ```data-path-concurrency```.  
+
+```perNodeConfig``` is a list of ```RuledConfigs``` each item of which matches one or more nodes by label selectors and specify the concurrent number for the matched nodes. This means, the nodes are identified by labels.  
+
+For example, the ```perNodeConfig`` could have below elements:
+```
+"nodeSelector: kubernetes.io/hostname=node1; number: 3"
+"nodeSelector: beta.kubernetes.io/instance-type=Standard_B4ms; number: 5"
+```
+The first element means the node with host name ```node1``` gets the Per-node concurrent number of 3.  
+The second element means all the nodes with label ```beta.kubernetes.io/instance-type``` of value ```Standard_B4ms``` get the Per-node concurrent number of 5. 
+At least one node is expected to have a label with the specified ```RuledConfigs``` element (rule). If no node is with this label, the Per-node rule makes no effect.  
+If one node falls into more than one rules, e.g., if node1 also has the label ```beta.kubernetes.io/instance-type=Standard_B4ms```, the smallest number (3) will be used.  
+
+### Sample
+A sample of the ```node-agent-configs``` configMap is as below:
+```json
+{
+    "globalConfig": 2,
+    "perNodeConfig": [
+        {
+            "nodeSelector": {
+                "matchLabels": {
+                    "kubernetes.io/hostname": "node1"
+                }
+            },
+            "number": 3
+        },    
+        {
+            "nodeSelector": {
+                "matchLabels": {
+                    "beta.kubernetes.io/instance-type": "Standard_B4ms"
+                }
+            },
+            "number": 5
+        }
+    ]
+}
+```
+To create the configMap, users need to save something like the above sample to a json file and then run below command:
+```
+kubectl create cm node-agent-configs -n velero --from-file=data-path-concurrency=<json file name>
+```
+
+### Global data path manager
+As for the code implementation, data path manager is to maintain the total number of the running VGDP instances and ensure the limit is not excceeded. At present, there is one data path manager instance per controller, as a result, the concurrent numbers are calculated separately for each controller. This doesn't help to limit the concurrency among different requesters.  
+Therefore, we need to create one global data path manager instance server-wide, and pass it to different controllers. The instance will be created at node-agent server startup.  
+The concurrent number is required to initiate a data path manager, the number comes from either Per-node concurrent number or Global concurrent number.    
+Below are some prototypes related to data path manager:  
+
+```go
+func NewManager(cocurrentNum int) *Manager
+func (m *Manager) CreateFileSystemBR(jobName string, requestorType string, ctx context.Context, client client.Client, namespace string, callbacks Callbacks, log logrus.FieldLogger) (AsyncBR, error)
+```
+
+
+
+
+
+[1]: unified-repo-and-kopia-integration/unified-repo-and-kopia-integration.md
--- a/changelogs/unreleased/6950-Lyndon-Li
+++ b/changelogs/unreleased/6950-Lyndon-Li
@@ -1,0 +1,1 @@
+Add the design for node-agent concurrency
--- a/design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
+++ b/design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
@@ -626,10 +626,9 @@ Therefore, we have below principles:
 
 We will address the two principles step by step. As the first step, VBDM’s parallelism is designed as below:  
 - We don’t create the load balancing mechanism for the first step, we don’t detect the accessibility of the volume/volume snapshot explicitly. Instead, we create the backupPod/restorePod under the help of Kubernetes, Kubernetes schedules the backupPod/restorePod to the appropriate node, then the data movement controller on that node will handle the DataUpload/DataDownload CR there, so the resource will be consumed from that node.
-- We don’t expose the configurable concurrency value in one node, instead, the concurrency value in value will be set to 1, that is, there is no concurrency in one node.
+- We expose the configurable concurrency value per node, for details of how the concurrency number constraints various backups and restores which share VGDP, check the [node-agent concurrency design][3].  
 
 As for the resource consumption, it is related to the data scale of the data movement activity and it is charged to node-agent pods, so users should configure enough resource to node-agent pods.  
-Meanwhile, Pod Volume Backup/Restore are also running in node-agent pods, we don’t restrict the concurrency of these two types. For example, in one node, one Pod Volume Backup and one DataUpload could run at the same time, in this case, the resource will be shared by the two activities.  
 
 ## Progress Report
 When a DUCR/DDCR is in InProgress phase, users could check the progress.  
@@ -968,3 +967,4 @@ Restore command is kept as is.
 
 [1]: ../unified-repo-and-kopia-integration/unified-repo-and-kopia-integration.md
 [2]: ../general-progress-monitoring.md
+[3]: ../node-agent-concurrency.md


design/node-agent-concurrency.md
```diff
@@ -0,0 +1,77 @@
...
+- Help users to isolate critical environments. Users may run some critical workloads in some specified nodes, since VGDP instances may take large resource consumption, users may only want to run them in the other nodes or run less number of instances in the nodes with critical workloads
+
+The range of Per-node concurrent number is the same with Global concurrent number.  
+Per-node concurrent number is preferable to Global concurrent number, so it will overwrite the Global concurrent number for that node.  
+Per-node concurrent number is implemented in a separate configMap named ```node-agent-configs```. This configMap is not created by Velero, users should create it manually on demand. The configMap should be created before node-agent server starts, otherwise, node-agent server should be restarted in order to make the effect.  
+
+The ```node-agent-configs``` configMap may be used for other purpose of configuring node-agent in future, at present, there is only one kind of data ```data-path-concurrency```. Its data struct is like below: 
+
+```go
+type DataPathConcurrency struct {
```
ywk253100, line 45: Not fully understand how this work. How about using the label selector instead? The key of the map is the label selector and the value is the current number.

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,77 @@
...
+- Inside the map of ```Configs```, if there is one element "bar":X, then X will be the Per-node concurrent number for the node with the label "foo":"bar"  
+
+```MatchLabel``` is optional, if it is not set, the hard-coded label key ```kubernetes.io/hostname``` will be used, which means the host name will be used to match nodes.  
+
+### Global data path manager
+As for the code implementation, data path manager is to maintain the total number of the running VGDP instances and ensure the limit is not excceeded. At present, there is one data path manager instance per controller, as a result, the concurrent numbers are calculated separately for each controller. This doesn't help to limit the concurrency among different requesters.  
+Therefore, we need to create one global data path manager instance server-wide, and pass it to different controllers. The instance will be created at node-agent server startup.  
+The concurrent number is required to initiate a data path manager, the number comes from either Per-node concurrent number or Global concurrent number.
```
ywk253100, line 65: How do we control the limit of the running VGDP instances? Is this already implemented by Velero or Kopia?

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,77 @@
...
+
+Varying from the data size, data complexity, resource availability, VGDP may take a long time and remarkable resources (CPU, memory, network bandwidth, etc.).  
+Technically, VGDP instances are able to run in concurrent regardless of the requesters. For example, a VGDP instance for a PodVolume backup could run in parallel with another VGDP instance for a DataUpload. Then the two VGDP instances share the same resources if they are running in the same node.  
+
+Therefore, in order to gain the optimized performance with the limited resources, it is worthy to configure the concurrent number of VGDP per node. When the resources are sufficient in nodes, users can set a large concurrent number, so as to reduce the backup/restore time; otherwise, the concurrency should be reduced or prohibited, otherwise, the backup/restore may encounter problems, i.e., time lagging, hang or OOM kill.  
+
+## Goals
+
+- Define the behaviors of concurrent VGDP instances in node-agent
```
ywk253100, line 19: If we support configuring the concurrency in the backup/restore level in the future, will this feature be impacted?

Lyndon-Li: It is not impacted. The two controls could work together, backups/restores are not probably all in the same node; while in the same node, the VGDP concurrency mechanism has higher priority over backup/restore level concurrency, as the former is for resources consideration, if there are not enough resources, the backups/restores level configurations are not be fulfilled.

design/node-agent-concurrency.md
Lyndon-Li: Here MatchLabel is the label selector. With the current approach, we want to give a shortcut of not specifying the selector all the time since `kubernetes.io/hostname` is used in most of the cases.

For example, we have below nodes in one cluster:

- Node-1 (kubernetes.io/hostname:cluster-node-1)
- Node-2 (kubernetes.io/hostname:cluster-node-2)
- Node-2 (kubernetes.io/hostname:cluster-node-3)

If I want to set an per-node number to Node-2, I make the struct like below:
```
Configs[cluster-node-2] = 3
```

In the other cluster, we don't want to use label `kubernetes.io/hostname` in the nodes, but another label `my-cluster/my-node-id`:
- Node-1 (my-cluster/my-node-id:00000)
- Node-2 (my-cluster/my-node-id:00001)
- Node-2 (my-cluster/my-node-id:00002)

Then I will have the struct like below:
```
MatchLabel = my-cluster/my-node-id
Configs[00001] = 3
```

If we use the label selector as the key directly, then users need to specify the per-node number as below:
```
Configs["kubernetes.io/hostname=cluster-node-2"] = 3
Configs["my-cluster/my-node-id=00001"] = 3
```

Finally, the current approach is simpler but the second approach (use the label selector as the key directly) looks more straight-forward.
So another proposal is like below:
If the key is with the label selector format (x=y), then we treat the key as a label selector
if the key is a single string (e.g., Configs[cluster-node-2] = 3), we treat it as a value of label ```kubernetes.io/hostname```

Let's vote on the 3 approaches.

design/node-agent-concurrency.md
Lyndon-Li: Yes, it has been implemented, check the [dataPath manager code](https://github.com/vmware-tanzu/velero/blob/main/pkg/datapath/manager.go)

blackpiglet: Is it possible to also update the PVB, PVR, DataUpload, and DataDownload status to reflect the reason why they are still not running?

```golang
	fsBackup, err := r.dataPathMgr.CreateFileSystemBR(pvb.Name, pVBRRequestor, ctx, r.Client, pvb.Namespace, callbacks, log)
	if err != nil {
		if err == datapath.ConcurrentLimitExceed {
			return ctrl.Result{Requeue: true, RequeueAfter: time.Minute}, nil
		} else {
			return r.errorOut(ctx, &pvb, err, "error to create data path", log)
		}
	}
```

design/node-agent-concurrency.md
Lyndon-Li: OK. Then let's use the straight way --- let's make it as `map[string]int `and have the keys like  "kubernetes.io/hostname=cluster-node-2" or "my-cluster/my-node-id=00001"

design/node-agent-concurrency.md
Lyndon-Li: OK, will do this in the code change PR.

design/node-agent-concurrency.md
Lyndon-Li: Done on the changes according to the discussion.

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,77 @@
...
+The range of Per-node concurrent number is the same with Global concurrent number.  
+Per-node concurrent number is preferable to Global concurrent number, so it will overwrite the Global concurrent number for that node.  
+Per-node concurrent number is implemented in a separate configMap named ```node-agent-configs```. This configMap is not created by Velero, users should create it manually on demand. The configMap should be created before node-agent server starts, otherwise, node-agent server should be restarted in order to make the effect.  
+
+The ```node-agent-configs``` configMap may be used for other purpose of configuring node-agent in future, at present, there is only one kind of data ```data-path-concurrency```. Its data struct is like below: 
+
+```go
+type DataPathConcurrency struct {
+	// Configs specifies the configs by node identified by label selector
+	Configs map[string]int `json:"configs"`
```
blackpiglet, line 47: The key's type is not consistent with the description. It should be changed from `string` to `metav1.LabelSelector`

Lyndon-Li: `metav1.LabelSelector `cannot be used as a map key, and moreover, this cm is created by users, a `metav1.LabelSelector `structure is not convenient for users to create this cm.

We can use string as the key and convert it to `metav1.LabelSelector` in the code

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+Therefore, in order to gain the optimized performance with the limited resources, it is worthy to configure the concurrent number of VGDP per node. When the resources are sufficient in nodes, users can set a large concurrent number, so as to reduce the backup/restore time; otherwise, the concurrency should be reduced or prohibited, otherwise, the backup/restore may encounter problems, i.e., time lagging, hang or OOM kill.  
+
+## Goals
+
+- Define the behaviors of concurrent VGDP instances in node-agent
+- Create a mechanism for users to specify the concurrent number of VGDP per node
+
+## Non-Goals
+- VGDP instances from different nodes always run in concurrent since in most common cases the resources are isolated. For special cases that some resources are shared across nodes, there is no support at present
+- In practice, restores run in prioritized scenarios, e.g., disarster recovery. However, the current design doesn't consider this difference, a VGDP instance for a restore is blocked if it reaches to the limit of the concurrency, even though the ones block it are for backups. If users do meet some problems here, they should consider to stop the backups first
```
shawn-hurley, line 24: nit: disarster/disaster

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+- Sometimes, users wants to block VGDP instances from running in a specific node, this is not considered in the current design. To archive this, more modules need to be considered, just setting VGDP instances' concurrent number to 0 won't bring the ideal reasult
```
shawn-hurley, line 25: what would be the result then? and why would it not be ideal

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Global concurrent number
+We add a node-agent server parameter ```data-path-concurrent-num``` to accept a concurrent number that will be applied to all nodes if the per-node number is not specified.  
+The number starts from 1 which means there is no concurrency, only one instance of VGDP is allowed. There is no roof limit.    
+If the node-agent server parameter is not specified, a hard-coded default value will be used, the value is set to 1.   
+
+### Per-node concurrent number
+We allow users to specify different concurrent number per node, for example, users can allow 3 concurrent instances in Node-1, 2 instances in Node-2 and 1 instance in Node-3. This is for below considerations:
+- The resources may be different among nodes. Then users could specify smaller concurrent number for nodes with less resources while larger number for the ones with more resources
+- Help users to isolate critical environments. Users may run some critical workloads in some specified nodes, since VGDP instances may take large resource consumption, users may only want to run them in the other nodes or run less number of instances in the nodes with critical workloads
```
shawn-hurley, line 37: > Users may only want to run them in the other nodes

How does this work if I can't set the number to zero and we do not consider turning off the node agent for specific nodes.

Feels like there is a mismatch between goals/non-goals and solution, can you help me understand?

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+
+The range of Per-node concurrent number is the same with Global concurrent number.  
+Per-node concurrent number is preferable to Global concurrent number, so it will overwrite the Global concurrent number for that node.  
+Per-node concurrent number is implemented in a separate configMap named ```node-agent-configs```. This configMap is not created by Velero, users should create it manually on demand. The configMap should be created before node-agent server starts, otherwise, node-agent server should be restarted in order to make the effect.
```
shawn-hurley, line 41: Can we watch for this particular config map instead and make the changes necessary when they are made?

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+- Help users to isolate critical environments. Users may run some critical workloads in some specified nodes, since VGDP instances may take large resource consumption, users may only want to run them in the other nodes or run less number of instances in the nodes with critical workloads
+
+The range of Per-node concurrent number is the same with Global concurrent number.  
+Per-node concurrent number is preferable to Global concurrent number, so it will overwrite the Global concurrent number for that node.  
+Per-node concurrent number is implemented in a separate configMap named ```node-agent-configs```. This configMap is not created by Velero, users should create it manually on demand. The configMap should be created before node-agent server starts, otherwise, node-agent server should be restarted in order to make the effect.  
+
+The ```node-agent-configs``` configMap may be used for other purpose of configuring node-agent in future, at present, there is only one kind of data ```data-path-concurrency```. Its data struct is like below: 
+
+```go
+type DataPathConcurrency struct {
```
shawn-hurley, line 46: Was it considered to make this a CRD singleton cluster scoped CR?

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+	// ConfigRules specifies the configs by node identified by label selector
+	ConfigRules map[string]int `json:"configRules"`
```
shawn-hurley, line 48: Instead of a string, we should use a [LabelSelector](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector), IMO. 

Also, something to consider: if multiple labels are matched, which one is selected? I suggest the safe thing is the lesser of the entries.

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+"kubernetes.io/hostname=node1": 3
+"beta.kubernetes.io/instance-type=Standard_B4ms": 5
+```
+The first element means the node with host name ```node1``` get the Per-node concurrent number of 3.  
+The second element means all the nodes with label ```beta.kubernetes.io/instance-type``` and value ```Standard_B4ms``` get the Per-node concurrent number of 5. 
+At least one node is expected to have a label with the specified key. If no node is with this label, the Per-node configuration makes no effect.  
+If one node falls into more than one elements, e.g., if ```node1``` also has the label ```beta.kubernetes.io/instance-type=Standard_B4ms```, the smallest number (3) will be used.  
+
+### Global data path manager
+As for the code implementation, data path manager is to maintain the total number of the running VGDP instances and ensure the limit is not excceeded. At present, there is one data path manager instance per controller, as a result, the concurrent numbers are calculated separately for each controller. This doesn't help to limit the concurrency among different requesters.
```
shawn-hurley, line 64: Can you help me understand. 

This is saying that the node agent controller will be responsible for the number of concurrent DataUploads/DataDownloads it can handle and ensure this is true? Can you help me understand why this is not the thread of the workers for the controller and needs something in a global manager?

shawn-hurley (commented):
I have a couple of questions on the design of this API

design/node-agent-concurrency.md
Lyndon-Li: 1. From me, this number is not frequently changed, so we don't need to watch it, but retrieve it once at server starts. 
2. There are some more details to cover. If we allow it, users may change the number in the configure any time, there will be a chaos. E.g., for one time, there are 3 concurrencies and for the other time, there are 2 concurrencies.
3. It makes the implementation complex, the dataPathManager is a singletone instance, if the number is not set only once when initiating it, we will need a watch mechanism and also a contest solving mechanism to update the number

Therefore, if there is not a case that strongly requires this, I would prefer not adding this complexy.

design/node-agent-concurrency.md
Lyndon-Li: Not sure if I get your question, let me try to answer. By the design, this is singleton for a single Velero installation. If there are more than one installation in a cluster, I didn't consider it for the design, but technically, yes, all the Velero installations will follow the same cm.

design/node-agent-concurrency.md
Lyndon-Li: >  if multiple labels are matched, which one is selected? I suggest the safe thing is the lesser of the entries

Yes, that is what we are going to do, let me clarify in the design

design/node-agent-concurrency.md
Lyndon-Li: Done, thanks.

design/node-agent-concurrency.md
Lyndon-Li: I added the details in the design.

design/node-agent-concurrency.md
Lyndon-Li: If the number is 0 for either the global number of per-node number, it will be ignored.

There were indeed some struggles on supporting the 0 number (blocking mode), but finally I gave it up for now. So there were some mismatch in the doc. I've removed  this sentence.

In future, if we see cases hardly require this, let's think about a solution from the entire workflow. As mentioned in the doc, only do it from VGDP instance level doesn't work

design/node-agent-concurrency.md
Lyndon-Li: > This is saying that the node agent controller will be responsible for the number of concurrent
No, this is the current behavior. We need to change the control to the node-agent server wide.

design/node-agent-concurrency.md
Lyndon-Li: > Instead of a string, we should use a [LabelSelector](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector), IMO.

See [this comment](https://github.com/vmware-tanzu/velero/pull/6950#discussion_r1372505929). The string is in the pattern of LabelSelector, so the code will convert a string to a LabelSelector. This is for two considerations:
1. LabelSelector cannot be a key of map. To use LabelSelector, the structure will be complex
2. A string is convenient for users to write the config, they don't need to write "{}" or the assignment to "matchLabel"

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+We allow users to specify different concurrent number per node, for example, users can allow 3 concurrent instances in Node-1, 2 instances in Node-2 and 1 instance in Node-3. This is for below considerations:
+- The resources may be different among nodes. Then users could specify smaller concurrent number for nodes with less resources while larger number for the ones with more resources
+- Help users to isolate critical environments. Users may run some critical workloads in some specified nodes, since VGDP instances may take large resource consumption, users may want to run less number of instances in the nodes with critical workloads
+
+The range of Per-node concurrent number is the same with Global concurrent number.  
+Per-node concurrent number is preferable to Global concurrent number, so it will overwrite the Global concurrent number for that node.  
+Per-node concurrent number is implemented in a separate configMap named ```node-agent-configs```. This configMap is not created by Velero, users should create it manually on demand. The configMap should be created before node-agent server starts, otherwise, node-agent server should be restarted in order to make the effect.
```
reasonerjt, line 41: If the configmap is an alternative to the param for node-agent server, I'd vote for node-agent server.

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+```
+The first element means the node with host name ```node1``` get the Per-node concurrent number of 3.  
+The second element means all the nodes with label ```beta.kubernetes.io/instance-type``` and value ```Standard_B4ms``` get the Per-node concurrent number of 5. 
+At least one node is expected to have a label with the specified key. If no node is with this label, the Per-node configuration makes no effect.  
+If one node falls into more than one elements, e.g., if ```node1``` also has the label ```beta.kubernetes.io/instance-type=Standard_B4ms```, the smallest number (3) will be used.  
+
+### Global data path manager
+As for the code implementation, data path manager is to maintain the total number of the running VGDP instances and ensure the limit is not excceeded. At present, there is one data path manager instance per controller, as a result, the concurrent numbers are calculated separately for each controller. This doesn't help to limit the concurrency among different requesters.  
+Therefore, we need to create one global data path manager instance server-wide, and pass it to different controllers. The instance will be created at node-agent server startup.
```
reasonerjt, line 65: by server-wide, do you mean node-agent server or the whole velero setup?  I assume it's node-agent server, if that's the node-agent server, I don't think it's global

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+The number starts from 1 which means there is no concurrency, only one instance of VGDP is allowed. There is no roof limit.    
+If the node-agent server parameter is not specified, a hard-coded default value will be used, the value is set to 1.   
+
+### Per-node concurrent number
+We allow users to specify different concurrent number per node, for example, users can allow 3 concurrent instances in Node-1, 2 instances in Node-2 and 1 instance in Node-3. This is for below considerations:
+- The resources may be different among nodes. Then users could specify smaller concurrent number for nodes with less resources while larger number for the ones with more resources
+- Help users to isolate critical environments. Users may run some critical workloads in some specified nodes, since VGDP instances may take large resource consumption, users may want to run less number of instances in the nodes with critical workloads
+
+The range of Per-node concurrent number is the same with Global concurrent number.  
+Per-node concurrent number is preferable to Global concurrent number, so it will overwrite the Global concurrent number for that node.
```
reasonerjt, line 40: The term "Global concurrent number" is a little confusing to me.  IMO this concurrent number is to control the concurrency within one node-agent server, and if this is true, the number is per-node.  And the "Global current number" is just a shortcut to help user to set the currency for every node?

design/node-agent-concurrency.md
Lyndon-Li: It is node-agent server. Maybe we can give it a better name, what about "Ordinary Number"?

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,78 @@
...
+- In practice, restores run in prioritized scenarios, e.g., disaster recovery. However, the current design doesn't consider this difference, a VGDP instance for a restore is blocked if it reaches to the limit of the concurrency, even though the ones block it are for backups. If users do meet some problems here, they should consider to stop the backups first
+- Sometimes, users wants to block VGDP instances from running in a specific node, this is not considered in the current design. To archive this, more modules need to be considered, just setting VGDP instances' concurrent number to 0 won't bring the ideal reasult. E.g., for fs backup, VGDP instance must run in the node the source pod is running in, if we simply block from VGDP instance level, the PodVolumeBackup CR is still submitted but never processed.  
+
+## Solution
+
+### Global concurrent number
+We add a node-agent server parameter ```data-path-concurrent-num``` to accept a concurrent number that will be applied to all nodes if the per-node number is not specified.
```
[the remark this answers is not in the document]
Lyndon-Li: It is for node-agent server, see [this comment](https://github.com/vmware-tanzu/velero/pull/6950#discussion_r1374654651)

design/node-agent-concurrency.md
Lyndon-Li: It is not an alternative, the thing is, we want a number that applied to all nodes and we also want some nodes have their particular number.

design/node-agent-concurrency.md
shawn-hurley: So either, I think this should be a cluster-scoped resource as a CRD (not a config map). I think it will be really confusing otherwise.

design/node-agent-concurrency.md
shawn-hurley: I don' know if this makes sense. There is a well-defined struct for label selector, and if you don't use that, it won't be very clear. 

I would suggest using a CRD for this and the well-known label selector struct. I would probably just do a simple list. Then, you can order that list by the size of the threads. you will loop through the selectors, and once one is found, it will be the lowest option. 

I don't think you gain anything from having a map here, as to see if a node is selected you have to look at the key, so you will have to loop over the map. I think that a list would be a better choice.

design/node-agent-concurrency.md
shawn-hurley: I don't understand. Can you please help me?

Is this because we have two things (DataUploads and DataDownloads) with workers, and you want to split them? 

If this is the case, then isn't there a concern that a node agent will take ownership, because its worker is free just spinning because there are no available "dataManager" threads? Then other node agents that are in the system, that may have free "dataManager" threads, will be sitting idle?

Did I miss us handling this in the design?

shawn-hurley:
I would like to see an alternative explored, where

1. we handle a node not having workers running for one of DataDownloads OR DataUploads or both. I think this is going to be an everyday thing. You don't want a node that is doing AI work with GPU's to have to have a worker running. Especially if you say you have  a node with tons of resources for this purpose. 
2. One that uses a CRD that uses well defined label selector syntax. I would also like to consider being able to break up if a node does DataDownloads or DataRestores.

Lyndon-Li:
> we handle a node not having workers running for one of DataDownloads OR DataUploads or both. I think this is going to be an everyday thing. You don't want a node that is doing AI work with GPU's to have to have a worker running. Especially if you say you have a node with tons of resources for this purpose.

I agree it is a common requirement. However, as I mentioned, we cannot simply make this from define the concurrency of VGDP. 
We need to consider the reality of different backup types and consider more modules like the exposer. E.g., we cannot make this for fs-backup, because the VGDP must run in the same node that the source pod is running. 
Therefore, to support this, I would suggest we a new design which would be very different from the current one

Lyndon-Li:
> I would also like to consider being able to break up if a node does DataDownloads or DataRestores. I would also add to this, consideration of something other than a per node singleton that manages all the data path stuff.

I agree this is a reasonable requirement. But again, we cannot finish this from the VGDP level only. Things are different for backup types. E.g., for both fs restore and data mover restore, pods are first created and then VGDP instances are launched, the thing is, in some cases, a VGDP instance must run in the same node where the pod is scheduled to, otherwise, the restored volume may not be mounted to the pod if there are some topology/affinity/taint settings.

So the same, these requirements is not the current design could fulfill, because it is only aiming to the number of VGDP instances

design/node-agent-concurrency.md
Lyndon-Li: Let me try with a list of label selector. I have ever considered this, but it eventually involves two data structures, so I just dropped

shawn-hurley:
I think that if we agree these are reasonable requirements, then the design should also solve them. I wonder if we could take some time in a community meeting or set up a one-off call, to discuss and brainstorm if that would be helpful?

design/node-agent-concurrency.md
Lyndon-Li: As far as my understanding, this per-node number is a static configuration, so I don't understand the benefit to make it a CRD.
If you are talking about dynamically watching it, as [mentioned here](https://github.com/vmware-tanzu/velero/pull/6950#discussion_r1373986044), I don't think it is necessary to dynamically watch a configuration that is not to be frequently changed.

shawn-hurley: It may not be frequently changed, BUT when you change it, what is the process now, and how does it affect running backups?

design/node-agent-concurrency.md
Lyndon-Li: I might have mislead you. The thing is, in the existing code, there are one dataPathMgr in one controller, so in order to count the overall VGDP instances no matter they are from a DataUpload, DataDownload, PodVolumeBackup or PodVolumeRestore, we need to make a singleton dataPathMgr. See [the existing code](https://github.com/vmware-tanzu/velero/blob/main/pkg/datapath/manager.go)

Lyndon-Li:
> I think that if we agree these are reasonable requirements, then the design should also solve them. I wonder if we could take some time in a community meeting or set up a one-off call, to discuss and brainstorm if that would be helpful?

Any way is fine to me. We can discuss in the next community meeting. Meanwhile, I am not sure if we are on the same page for everything (regardless our opinions), if not, we can run a sync-up call first.

design/node-agent-concurrency.md
sseago: @shawn-hurley any backup or restore in an InProgress state will be marked failed and a new one will need to be created, since we have not yet solved the problem of making InProgress  backups or restores resumable.

Lyndon-Li: First of all, it should not affect the running backup, because you cannot stop one VGDP instance if it is already started, e.g., previously, the backup gets 3 concurrency, now if we change it to 2, nothing we can do to the backup.

Secondly, the current logic after the number is changed is that users have to restart the node-agent server when there is no running backups, otherwise, the update never takes effect.
We are doing in this way is just to prevent users frequently change this number and we don't want the already submitted backups/restores get affected.

sseago: But wait, we're talking about the node agent here. That won't fail backups or restores directly, but it *will* cause in-progress DataUploads or DataDownloads to be canceled. I'm assuming it will fail PodVolumeBackups/PodVolumeRestores as well.

Lyndon-Li: So both the global number change and per-node number change won't take effect unless node-agent server restarts and node-agent server is only recommended to restart when there are no running backups/restores

shawn-hurley: I am worried about the brittleness of this with a backup and restore system.

design/node-agent-concurrency.md
shawn-hurley: Ok, I do think that, because a node takes ownwership of something and then can determine if it can do work on it, we may want to consider how to make sure this is something that can more adequately be spread out.

design/node-agent-concurrency.md
Lyndon-Li: 1. For the already running backups/restores, I do believe this is a reasonable behavior, we cannot peel a data path thread (which for example has backed up tones of data) whenever the number is changed by users.
2. For the backups/restores to be created/submitted, we can use the new number without restarting node-agent server, but I don't think it is something must have, especially in the first version. 

For 2, I am open for a vote of this. It is just efforts/complexity vs. user friendliness

shawn-hurley: I don't think I am following 1 unless you are spelling out the agreement on letting running threads continue running and reducing the number over time.

I believe that if you implemented the system described in 1 and 2 it would be more resilient and have a better UX (especially when you say need to bump up/down these resources for a $REASON.

Lyndon-Li: > the agreement on letting running threads continue running and reducing the number over time

Yes, this is what I meant, the limit should only work on the new coming VGDP instance creations.

Lyndon-Li:
> we handle a node not having workers running for one of DataDownloads OR DataUploads or both.
> I would also like to consider being able to break up if a node does DataDownloads or DataRestores.

@shawn-hurley For these two questions, I think we are still not in the same page, because I thought you regarded them as something we can make, but I don't think so --- considering all types, PVB/PVR, data mover backup/restore, these are something we cannot achieve right now and for a long time in future. For the details, we need to consider from end to end:

**Why can't we support --- "we handle a node not having workers running for one of DataDownloads OR DataUploads or both"**:
For all the types, only data mover backup could support it right now. For PVB/PVR, at the time when launching the VGDP instance, the source/restored pod is already there in a specific node, and volume is already mounted in that node, we must  run VGDP in the node, no way to choose. So it is a vain to set VGDP limit in that node to 0, unless we want the PVB/PVR never to finish. For data mover restore, in the case that the volume binding mode is `WaitForFirstConsumer`, things are the same, we cannot run the VGDP in a node we choose, but only in the node that the restored pod is scheduled to.

This means, unless enhancing the above pod schedule related behavior, we have no way to support it. For sure, it will be a hard work.

**Why can't we support --- "being able to break up if a node does DataDownloads or DataRestores"**:
For the same reasons as mentioned above, VGDP instances for PVB, PVR and DataRestores cannot run in a node we choose, so it is a vain to split them as the ones for backups and for restores.

design/node-agent-concurrency.md
Lyndon-Li: Done, changed to LabelSelector

Lyndon-Li:
Done on some further changes to address comments:

- Change the rule structs to use list and LabelSelector
- Move the GlobalNum to the configMap as well
- Add a sample for the configMap

Please continue the review.

shawn-hurley:
Thank you for explaining where we were not on the same page. I think that helped a lot for me to understand the concerns. I think now we have to go up the stack a second and talk about the feature and what it means it for users. 

Is it worth breaking up *what* can be expected by users based on which backup they choose? This would allow us to do the "more correct" thing for DataUploads and continue to support PVB/PVR requests and DataDownloads that have to get set for a specific node. 

I think that it is reasonable that the main need here is to control nodes that are uploading data. If you are restoring, something went wrong, and you probably want most/all available resources to do that, from a UX perspective, it seems reasonable. 

What are the downsides of this?

Lyndon-Li:
> Is it worth breaking up what can be expected by users based on which backup they choose? This would allow us to do the "more correct" thing for DataUploads

This is what I was planning to do in a a separate PR/design (another issue #7036 is opened for it), since its implementation/usability deviates the current PR:
1. The modules to implement this is different, that is, exposer vs. VGDP
2. The scope is different, that is, data mover backup vs. all tasks with moving data
3. The user experience is different, that is, black list vs. VGDP concurrency

shawn-hurley (commented):
Concerns around naming now that I finally understand the issues, thank you @Lyndon-Li for being patient while I caught up. 

I hope the comments make sense in that light, let me know if I am still misunderstanding!

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Solution
+
+We introduce a configMap named ```node-agent-configs``` for users to specify the node-agent related configurations. This configMap is not created by Velero, users should create it manually on demand.  
+Node-agent server checks these configurations at startup time and use it to initiate the related VGDP modules. Therefore, users could edit this configMap any time, but in order to make the changes effective, node-agent server needs to be restarted.  
+The ```node-agent-configs``` configMap may be used for other purpose of configuring node-agent in future, at present, there is only one kind of configuration as the data in the configMap, the name is ```data-path-concurrency```.  
+
+The data structure for ```data-path-concurrency``` is as below:
+```go
+type DataPathConcurrency struct {
```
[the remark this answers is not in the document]
Lyndon-Li: Yes, DataPath is more like a design/architecture level concept.
We have discussed a little about this in the community meeting, but haven't get the final answer. Some of the previous nominations -- "dataPathConcurrency", "dataMoverConcurrency".

Personally, I think "loadConcurrency" (we don't need to mention "node" because it is under the node-agent-configs) is close to what we are doing here and is not ambiguous. Waiting for others' opinions.

design/node-agent-concurrency.md
```diff
@@ -0,0 +1,124 @@
...
+- In practice, restores run in prioritized scenarios, e.g., disaster recovery. However, the current design doesn't consider this difference, a VGDP instance for a restore is blocked if it reaches to the limit of the concurrency, even though the ones block it are for backups. If users do meet some problems here, they should consider to stop the backups first
+- Sometimes, users wants to totally block backups/restores from running in a specific node, this is out of the scope the current design. To archive this, more modules need to be considered (i.e., expoers of data movers), simply blocking the VGDP (e.g., by setting its concurrent number to 0) doesn't work. E.g., for a fs backup, VGDP instance must run in the node the source pod is running in, if we simply block from VGDP instance, the PodVolumeBackup CR is still submitted but never processed.  
+
+## Solution
+
+We introduce a configMap named ```node-agent-configs``` for users to specify the node-agent related configurations. This configMap is not created by Velero, users should create it manually on demand.
```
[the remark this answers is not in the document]
Lyndon-Li: Good question, I have modify the design to include it --- the configMap is namespace specific and only works for node-agent in the specific namespace.

design/node-agent-concurrency.md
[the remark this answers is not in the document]
Lyndon-Li: For this naming question, we decide not to block the merge of this design. We can keep the discussion on this, and when we want to change the name, if the design has been merged, we come back to make slight change to this design.
So this discussion can continue until the FC of 1.13.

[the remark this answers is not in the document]
sseago: @Lyndon-Li "loadConcurrency" works for me.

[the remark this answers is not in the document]
shawn-hurley: loadConcurrency works for me as well. Docs on this guy will be necessary, but I now better understand what this is doing. Thank you for taking the time, and I agree if inspiration strikes and folks come up with a better name, awesome!

design/node-agent-concurrency.md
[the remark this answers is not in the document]
shawn-hurley: It looks good to me. I worry about the UX once you have three and try to understand what is going on in a node across the three pods, but I think we can solve it if/when we hit that problem.

blackpiglet (approved):

ywk253100 (approved):

PR #6956, opened 2023-10-13
Add design for Extending VolumePolicies to support more actions

shubham-pampattiwar:
This PR adds design for https://github.com/vmware-tanzu/velero/issues/6640

### DIFF
--- a/design/Extend-VolumePolicies-to-support-more-actions.md
+++ b/design/Extend-VolumePolicies-to-support-more-actions.md
@@ -1,0 +1,308 @@
+# Extend VolumePolicies to support more actions
+
+## Abstract
+
+Currently, the [VolumePolicies feature](https://github.com/vmware-tanzu/velero/blob/main/design/Implemented/handle-backup-of-volumes-by-resources-filters.md) which can be used to filter/handle volumes during backup only supports the skip action on matching conditions. Users need more actions to be supported.
+
+## Background
+
+The `VolumePolicies` feature was introduced in Velero 1.11 as a flexible way to handle volumes. The main agenda of
+introducing the VolumePolicies feature was to improve the overall user experience when performing backup operations
+for volume resources, the feature enables users to group volumes according the `conditions` (criteria) specified and
+also lets you specify the `action` that velero needs to take for these grouped volumes during the backup operation.
+The limitation being that currently `VolumePolicies` only supports `skip` as an action, We want to extend the `action`
+functionality to support more usable options like `fs-backup` (File system backup) and `snapshot` (VolumeSnapshots).
+
+## Goals
+- Extending the VolumePolicies to support more actions like `fs-backup` (File system backup) and `snapshot` (VolumeSnapshots).
+- Improve user experience when backing up Volumes via Velero
+
+## Non-Goals
+- No changes to existing approaches to opt-in/opt-out annotations for volumes
+- No changes to existing `VolumePolicies` functionalities
+- No additions or implementations to support more granular actions like `snapshot-csi` and `snapshot-datamover`. These actions can be implemented as a future enhancement
+
+
+## Use-cases/Scenarios
+
+**Use-case 1:**
+- A user wants to use `snapshot` (volumesnapshots) backup option for all the csi supported volumes and `fs-backup` for the rest of the volumes.
+- Currently, velero supports this use-case but the user experience is not that great.
+- The user will have to individually annotate the volume mounting pod with the annotation "backup.velero.io/backup-volumes" for `fs-backup`
+- This becomes cumbersome at scale.
+- Using `VolumePolicies`, the user can just specify 2 simple `VolumePolicies` like for csi supported volumes as `snapshot` action and rest can be backed up`fs-backup` action:
+```yaml
+version: v1
+volumePolicies:
+  - conditions:
+      storageClass:
+        - gp2
+    action:
+      type: snapshot
+  - conditions: {}
+    action:
+      type: fs-backup
+```
+
+**Use-case 2:**
+- A user wants to use `fs-backup` for nfs volumes pertaining to a particular server
+- In such a scenario the user can just specify a `VolumePolicy` like:
+```yaml
+version: v1
+volumePolicies:
+- conditions:
+    nfs:
+      server: 192.168.200.90
+  action:
+    type: fs-backup
+```
+## High-Level Design
+- When the VolumePolicy action is set as `fs-backup` the backup workflow modifications would be:
+  - We call [backupItem() -> backupItemInternal()](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L95) on all the items that are to be backed up
+  - Here when we encounter [Pod as an item ](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L195)
+  - We will have to modify the backup workflow to account for the `fs-backup` VolumePolicy action
+
+
+- When the VolumePolicy action is set as `snapshot` the backup workflow modifications would be:
+  - Once again, We call [backupItem() -> backupItemInternal()](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L95) on all the items that are to be backed up
+  - Here when we encounter [Persistent Volume as an item](https://github.com/vmware-tanzu/velero/blob/d4128542590470b204a642ee43311921c11db880/pkg/backup/item_backupper.go#L253)
+  - And we call the [takePVSnapshot func](https://github.com/vmware-tanzu/velero/blob/d4128542590470b204a642ee43311921c11db880/pkg/backup/item_backupper.go#L508)
+  - We need to modify the takePVSnapshot function to account for the `snapshot` VolumePolicy action.
+  - In case of csi snapshots for PVC objects, these snapshot actions are taken by the velero-plugin-for-csi, we need to modify the [executeActions()](https://github.com/vmware-tanzu/velero/blob/512fe0dabdcb3bbf1ca68a9089056ae549663bcf/pkg/backup/item_backupper.go#L232) function to account for the `snapshot` VolumePolicy action.
+
+**Note:** `Snapshot` action can either be a native snapshot or a csi snapshot, as is the case with the current flow where velero itself makes the decision based on the backup CR.
+
+## Detailed Design
+- Update VolumePolicy action type validation to account for `fs-backup` and `snapshot` as valid VolumePolicy actions.
+- Modifications needed for `fs-backup` action:
+  - Now based on the specification of volume policy on backup request we will decide whether to go via legacy pod annotations approach or the newer volume policy based fs-backup action approach.
+  - If there is a presence of volume policy(fs-backup/snapshot)  on the backup request that matches as an action for a volume we use the newer volume policy approach to get the list of the volumes for `fs-backup` action
+  - Else continue with the annotation based legacy approach workflow.
+
+- Modifications needed for `snapshot` action:
+  - In the [takePVSnapshot function](https://github.com/vmware-tanzu/velero/blob/d4128542590470b204a642ee43311921c11db880/pkg/backup/item_backupper.go#L508) we will check the PV fits the volume policy criteria and see if the associated action is `snapshot`
+  - If it is not snapshot then we skip the further workflow and avoid taking the snapshot of the PV
+  - Similarly, For csi snapshot of PVC object, we need to do similar changes in [executeAction() function](https://github.com/vmware-tanzu/velero/blob/512fe0dabdcb3bbf1ca68a9089056ae549663bcf/pkg/backup/item_backupper.go#L348). we will check the PVC fits the volume policy criteria and see if the associated action is `snapshot` via csi plugin
+  - If it is not snapshot then we skip the csi BIA execute action and avoid taking the snapshot of the PVC by not invoking the csi plugin action for the PVC
+
+**Note:**
+- When we are using the `VolumePolicy` approach for backing up the volumes then the volume policy criteria and action need to be specific and explicit, there is no default behaviour, if a volume matches `fs-backup` action then `fs-backup` method will be used for that volume and similarly if the volume matches the criteria for `snapshot` action then the snapshot workflow will be used for the volume backup.
+- Another thing to note is the workflow proposed in this design uses the legacy opt-in/opt-out approach as a fallback option. For instance, the user specifies a VolumePolicy but for a particular volume included in the backup there are no actions(fs-backup/snapshot) matching in the volume policy for that volume, in such a scenario the legacy approach will be used for backing up the particular volume.
+## Implementation
+
+- The implementation should be included in velero 1.14
+
+- We will introduce a `VolumeHelper` interface. It will consist of two methods:
+  - `ShouldPerformFSBackupForPodVolume(pod *corev1api.Pod)`
+  - `ShouldPerformSnapshot(obj runtime.Unstructured)`
+```go
+type VolumeHelper interface {
+	GetVolumesForFSBackup(pod *corev1api.Pod) ([]string, []string, error)
+	ShouldPerformSnapshot(obj runtime.Unstructured) (bool, error)
+}
+```
+-  The `VolumeHelperImpl` struct will implement the `VolumeHelper` interface and will consist of the functions that we will use through the backup workflow to accommodate volume policies for PVs and PVCs.
+```go
+type VolumeHelperImpl struct {
+	Backup                   *velerov1api.Backup
+	VolumePolicy             *resourcepolicies.Policies
+	BackupExcludePVC         bool
+	DefaultVolumesToFsBackup bool
+	SnapshotVolumes          *bool
+	Logger                   logrus.FieldLogger
+}
+
+```
+
+- We will create an instance of the struct the `VolumeHelperImpl` in `item_backupper.go`
+```go
+	vh := &volumehelper.VolumeHelperImpl{
+		Backup:                   ib.backupRequest.Backup,
+		VolumePolicy:             ib.backupRequest.ResPolicies,
+		BackupExcludePVC:         !ib.backupRequest.ResourceIncludesExcludes.ShouldInclude(kuberesource.PersistentVolumeClaims.String()),
+		DefaultVolumesToFsBackup: boolptr.IsSetToTrue(ib.backupRequest.Spec.DefaultVolumesToFsBackup),
+		SnapshotVolumes:          ib.backupRequest.Spec.SnapshotVolumes,
+		Logger:                   logger,
+	}
+```
+
+
+#### FS-Backup
+- Regarding `fs-backup` action to decide whether to use legacy annotation based approach or volume policy based approach:
+  - We will use the `vh.GetVolumesForFSBackup()` function from the `volumehelper` package
+  - Functions involved in processing `fs-backup` volume policy action will somewhat look like:
+
+```go
+func (v *VolumeHelperImpl) GetVolumesForFSBackup(pod *corev1api.Pod) ([]string, []string, error) {
+	// Check if there is a fs-backup/snapshot volumepolicy specified by the user, if yes then use the volume policy approach to
+	// get the list volumes for fs-backup else go via the legacy annotation based approach
+
+	var includedVolumes = make([]string, 0)
+	var optedOutVolumes = make([]string, 0)
+
+	FSBackupOrSnapshot, err := checkIfFsBackupORSnapshotPolicyForPodVolume(pod, v.VolumePolicy)
+	if err != nil {
+		return includedVolumes, optedOutVolumes, err
+	}
+
+	if v.VolumePolicy != nil && FSBackupOrSnapshot {
+		// Get the list of volumes to back up using pod volume backup for the given pod matching fs-backup volume policy action
+		includedVolumes, optedOutVolumes, err = GetVolumesMatchingFSBackupAction(pod, v.VolumePolicy)
+		if err != nil {
+			return includedVolumes, optedOutVolumes, err
+		}
+	} else {
+		// Get the list of volumes to back up using pod volume backup from the pod's annotations.
+		includedVolumes, optedOutVolumes = pdvolumeutil.GetVolumesByPod(pod, v.DefaultVolumesToFsBackup, v.BackupExcludePVC)
+	}
+	return includedVolumes, optedOutVolumes, err
+}
+
+func checkIfFsBackupORSnapshotPolicyForPodVolume(pod *corev1api.Pod, volumePolicies *resourcepolicies.Policies) (bool, error) {
+
+	for volume := range pod.Spec.Volumes {
+		action, err := volumePolicies.GetMatchAction(volume)
+		if err != nil {
+			return false, err
+		}
+		if action.Type == resourcepolicies.FSBackup || action.Type == resourcepolicies.Snapshot {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// GetVolumesMatchingFSBackupAction returns a list of volume names to backup for the provided pod having fs-backup volume policy action
+func GetVolumesMatchingFSBackupAction(pod *corev1api.Pod, volumePolicy *resourcepolicies.Policies) ([]string, []string, error) {
+	ActionMatchingVols := []string{}
+	NonActionMatchingVols := []string{}
+	for _, vol := range pod.Spec.Volumes {
+		action, err := volumePolicy.GetMatchAction(vol)
+		if err != nil {
+			return nil, nil, err
+		}
+		// Now if the matched action is `fs-backup` then add that Volume to the fsBackupVolumeList
+		if action != nil && action.Type == resourcepolicies.FSBackup {
+			ActionMatchingVols = append(ActionMatchingVols, vol.Name)
+		} else {
+			NonActionMatchingVols = append(NonActionMatchingVols, vol.Name)
+		}
+	}
+
+	return ActionMatchingVols, NonActionMatchingVols, nil
+}
+```
+- The main function from the above `vph.ProcessVolumePolicyFSbackup` will be called when we encounter Pods during the backup workflow:
+```go
+includedVolumes, optedOutVolumes, err := vh.GetVolumesForFSBackup(pod)
+			if err != nil {
+				backupErrs = append(backupErrs, errors.WithStack(err))
+			}
+
+```
+#### Snapshot (PV)
+
+- Making sure that `snapshot` action is skipped for PVs that do not fit the volume policy criteria, for this we will use the `vh.ShouldPerformSnapshot` from the `VolumeHelperImpl(vh)` receiver.
+```go
+func (v *VolumeHelperImpl) ShouldPerformSnapshot(obj runtime.Unstructured) (bool, error) {
+	// check if volume policy exists and also check if the object(pv/pvc) fits a volume policy criteria and see if the associated action is snapshot
+	// if it is not snapshot then skip the code path for snapshotting the PV/PVC
+	if v.VolumePolicy != nil {
+		action, err := v.VolumePolicy.GetMatchAction(obj)
+		if err != nil {
+			return false, err
+		}
+
+		// Also account for SnapshotVolumes flag on backup CR
+		if action != nil && action.Type == resourcepolicies.Snapshot && boolptr.IsSetToTrue(v.SnapshotVolumes) {
+			return true, nil
+		}
+	}
+
+	// now if volumepolicy is not specified then just check for snapshotVolumes flag
+	if boolptr.IsSetToTrue(v.SnapshotVolumes) {
+		return true, nil
+	}
+
+	return false, nil
+
+}
+```
+
+- The above function will be used as follows in `takePVSnapshot` function of the backup workflow:
+```go
+snapshotVolume, err := vh.ShouldPerformSnapshot(obj)
+
+	if err != nil {
+		return err
+	}
+
+	if !snapshotVolume {
+		log.Info(fmt.Sprintf("skipping volume snapshot for PV %s as it does not fit the volume policy criteria for snapshot action", pv.Name))
+		ib.trackSkippedPV(obj, kuberesource.PersistentVolumes, volumeSnapshotApproach, "does not satisfy the criteria for volume policy based snapshot action", log)
+		return nil
+	}
+```
+#### Snapshot (PVC)
+
+- Making sure that `snapshot` action is skipped for PVCs that do not fit the volume policy criteria, for this we will again use the `vh.ShouldPerformSnapshot` from the `VolumeHelperImpl(vh)` receiver.
+- We will pass the `VolumeHelperImpl(vh)` instance in `executeActions` method so that it is available to use.
+```go
+func (v *VolumeHelperImpl) ShouldPerformSnapshot(obj runtime.Unstructured) (bool, error) {
+	// check if volume policy exists and also check if the object(pv/pvc) fits a volume policy criteria and see if the associated action is snapshot
+	// if it is not snapshot then skip the code path for snapshotting the PV/PVC
+	if v.VolumePolicy != nil {
+		action, err := v.VolumePolicy.GetMatchAction(obj)
+		if err != nil {
+			return false, err
+		}
+
+		// Also account for SnapshotVolumes flag on backup CR
+		if action != nil && action.Type == resourcepolicies.Snapshot && boolptr.IsSetToTrue(v.SnapshotVolumes) {
+			return true, nil
+		}
+	}
+
+	// now if volumepolicy is not specified then just check for snapshotVolumes flag
+	if boolptr.IsSetToTrue(v.SnapshotVolumes) {
+		return true, nil
+	}
+
+	return false, nil
+
+}
+
+```
+- The above function will be used as follows in the `executeActions` function of backup workflow:
+```go
+		if groupResource == kuberesource.PersistentVolumeClaims && actionName == csiBIAPluginName {
+			snapshotVolume, err := vh.ShouldPerformSnapshot(obj)
+			if err != nil {
+				return nil, itemFiles, errors.WithStack(err)
+			}
+
+			if !snapshotVolume {
+				log.Info(fmt.Sprintf("skipping csi volume snapshot for PVC %s as it does not fit the volume policy criteria for snapshot action", namespace+" /"+name))
+				ib.trackSkippedPV(obj, kuberesource.PersistentVolumeClaims, volumeSnapshotApproach, "does not satisfy the criteria for volume policy based snapshot action", log)
+				continue
+			}
+		}
+```
+
+
+## Future Implementation
+It makes sense to add more specific actions in the future, once we deprecate the legacy opt-in/opt-out approach to keep things simple. Another point of note is, csi related action can be
+easier to implement once we decide to merge the csi plugin into main velero code flow.
+In the future, we envision the following actions that can be implemented:
+- `snapshot-native`: only use volume snapshotter (native cloud provider snapshots), do nothing if not present/not compatible
+- `snapshot-csi`: only use csi-plugin, don't use volume snapshotter(native cloud provider snapshots), don't use datamover even if snapshotMoveData is true
+- `snapshot-datamover`: only use csi with datamover, don't use volume snapshotter (native cloud provider snapshots), use datamover even if snapshotMoveData is false
+
+**Note:** The above actions are just suggestions for future scope, we may not use/implement them as is. We could definitely merge these suggested actions as `Snapshot` actions and use volume policy parameters and criteria to segregate them instead of making the user explicitly supply the action names to such granular levels.
+
+## Related to Design
+
+[Handle backup of volumes by resources filters](https://github.com/vmware-tanzu/velero/blob/main/design/Implemented/handle-backup-of-volumes-by-resources-filters.md)
+
+## Alternatives Considered
+Same as the earlier design as this is an extension of the original VolumePolicies design
--- a/changelogs/unreleased/6956-shubham-pampattiwar
+++ b/changelogs/unreleased/6956-shubham-pampattiwar
@@ -1,0 +1,1 @@
+Add design for Extending VolumePolicies to support more actions


design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Use-cases/Scenarios
+
+**Use-case 1:**
+- A user wants to use `fs-backup-volumes-excludes` (volumesnapshots) backup option for all the csi supported volumes and `fs-backup-volumes` for the rest of the volumes.
+- Currently, velero supports this use-case but the user experience is not that great.
+- The user will have to individually annotate the volume mounting pod with the annotation "backup.velero.io/backup-volumes" for `fs-backup-volumes`
+- This becomes cumbersome at scale.
+- Using `VolumePolicies`, the user can just specify a simple `VolumePolicy` like for csi supported volumes and rest can be defaulted to `fs-backup-volumes`:
+```
```
mateusoliveira43, line 28: please add the language in code blocks for syntax rendering :smile:

shubham-pampattiwar: Done, updated the PR.

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Background
+
+The `VolumePolicies` feature was introduced in Velero 1.11 as a flexible way to handle volumes. The main agenda of introducing the VolumePolicies feature was to improve the overall user experience when performing backup operations for volume resources, the feature enables users to group volumes according the `conditions` (criteria) specified and also lets you specify the `action` that velero needs to take for these grouped volumes during the backup operation. The limitation being that currently `VolumePolicies` only supports `skip` as an action, We want to extend the `action` functionality to support more usable options like `fs-backup-volumes` (File system backup) and `fs-backup-volumes-excludes` (VolumeSnapshots).
+
+## Goals
+- Extending the VolumePolicies to support more actions like file system backup(`fs-backup-volumes`) and snapshots(`fs-backup-volumes-excludes`).
```
qiuming-best, line 12: @shubham-pampattiwar 
`fs-backup-volumes-excludes` may not be suitable for a snapshot way of backup, This name may have categories that are too general and the user may not related to `volume snapshot`; we need to make them more specific.

and also if we extend another more way of backup, this name would be more suitable, maybe a specific one is better

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,161 @@
...
+```yaml
+version: v1
+volumePolicies:
+- conditions:
+    nfs:
+      server: 192.168.200.90
+  action:
+    type: fs-backup-volumes
+```
+## High-Level Design
```
qiuming-best, line 50: Should we consider adjusting the backup workflow? In our current backup process, the procedures for `fs-backup`, `csi`, and `volume snapshot` are all mixed together. Should we consider clarifying these different backup workflows for easier maintenance in the future?

design/Extend-VolumePolicies-to-support-more-actions.md
sseago: Basically a volume matching this policy behaves exactly the same as if that volume had the backup-volumes-excludes annotation. It is opted out of fs-backup when we default to fs backup. That is exactly the scope of this new policy, so a more specific name would be less accurate.

design/Extend-VolumePolicies-to-support-more-actions.md
sseago: @qiuming-best I'm thinking that backup workflow redesign may be a separate thread here. We may well want to do it, but this particular policy change isn't adding any new workflow steps, it's just providing an alternate means for users to specify volumes for fs-backup opt-in and opt-out, so using the current workflow may be the clearest. Then in a separate development task, we can redesign the workflow if there's a need.

design/Extend-VolumePolicies-to-support-more-actions.md
shubham-pampattiwar: yes agreed with @sseago. @qiuming-best let me know what name feels accurate to you ? We are not gonna say `volume-snapshot` as an action for sure. Does the dropping for `fs-` prefix work for you ?

design/Extend-VolumePolicies-to-support-more-actions.md
shubham-pampattiwar: @qiuming-best IMO Redesigning backup workflow  is not scope of this design, we need to have a different design for that, The scope of this design is to improve the UX of already existing functionalities for the user, re-iterating what @sseago said here, enabling users with a better option to the already existing annotation based opt-in/opt-out approach is the exact goal and use-case of this design.

reasonerjt: > Enabling users with a better option to the already existing annotation-based opt-in/opt-out approach

This is the root of the disagreement here...  IMO the current opt-in/out design, when combined with other options, has created a lot confusions and corner cases, and we don't wanna repeat that.

reasonerjt, line 50: Please also add a paragraph in this section, with some examples, to clarify the priority.  
For example, a PVC may be annotated to `opt-out` but the action is set to `fs-backup-volumes` in the policy.

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,161 @@
...
+- Improve user experience when backing up Volumes via Velero
+
+## Non-Goals
+- No changes to existing approaches to opt-in/opt-out annotations for volumes
+- No changes to existing `VolumePolicies` functionalities
+
+
+## Use-cases/Scenarios
+
+**Use-case 1:**
```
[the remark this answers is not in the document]
sseago: @reasonerjt I"m not sure what you mean. If the user sets default-volumes-to-fs-backup, then that has higher priority, otherwise snapshot/CSI has higher priority.

I think replicating opt-in/opt-out is the only way to do this that makes sense unless we want to deprecate default-volumes-to-fs-backup and the whole opt-in/opt-out approach. Currently, that's the only supported way to decide between backup methods, so the most obvious way to go forward with this is to stick with that same model. The other approach would be to define a completely different workflow here and error out if user sets default-to-fs-backup *and* volume backup type policies -- but if we do that, then we basically have to start over with this design. @shubham-pampattiwar any thoughts?

sseago:
It sounds like we need to completely step back. @shubham-pampattiwar has been designing for one set of requirements, but the other reviews seem to have a completely different set of requirements in mind. So I think we have to step away from design for a bit and go back to requirements. I think we have a couple of options here:

1) Provide an alternative to the existing per-pod-volume annotations to determine when to use fs-backup. This is what the current design is based on -- workflow is unchanged from current annotations, we're just aiming to provide a more user-friendly means of selection.
2) Provide a completely different selection mechanism that relies 100% on VolumePolicies and doesn't interact with defaultVolumesToFsBackup or the annotations at all. This is what the other reviewers seem to want here.

1) will be the least risky to implement with the least amount of effort. 2) Might result in a more flexible approach, but it's completely incompatible with the legacy methods, so we would need validation to ensure that only one approach is used.

To expand on the validation needs -- backup.Spec.DefaultVolumesToFsBackup is a bool pointer. If this is set to true or false, then we're using the legacy opt-in/opt-out approach, and any volume policies referencing fs backup or snapshots should result in a validation error -- if this is true or false, then existing workflows will be used. If this is nil (unspecified), then volume policies are used to determine what to snapshot and what to use fsbackup for. We'll need to modify install/server to make this value a bool pointer instead of a plain bool -- currently, not specifying this on install/server results in the backup getting a false rather than a nil value here.

@shubham-pampattiwar what do you think?

design/Extend-VolumePolicies-to-support-more-actions.md
sseago: @reasonerjt see my comment at the end -- if we're using policies as another way to implement opt-in/opt-out, then this would work the same way as if a volume had annotations for both. When opt-out is selected (defaultVolumesToFsBackup=true), opt-in annotations are ignored ,and when opt-in is selected (defaultVolumesToFsBackup=false), opt-out annotations are ignored. On the other hand, if we end up deciding *not* to use the opt-in/opt-out workflow for this, then the two approaches are incompatible and we'll have validation to ensure that users only use these new volume policies when defaultVolumesToFsBackup is unset -- sort of like the new includes/excludes fields, users would need to choose to use the new approach or the legacy approach, not both.

shubham-pampattiwar:
@reasonerjt @qiuming-best @sseago I have re-done the design, please take another look, Thanks !

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,176 @@
...
+## High-Level Design
+- When the VolumePolicy action is set as `fs-backup` the backup workflow modifications would be:
+  - We call [backupItem() -> backupItemInternal()](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L95) on all the items that are to be backed up
+  - Here when we encounter [Pod as an item ](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L195)
+  - We will have to modify the backup workflow to account for the `fs-backup` VolumePolicy action
+
+**Note:** `fs-backup` via VolumePolicy and `fs-backup` via annotation based opt-in/opt-out approach would be mutually exclusive.
+
+- When the VolumePolicy action is set as `snapshot` the backup workflow modifications would be:
+  - Once again, We call [backupItem() -> backupItemInternal()](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L95) on all the items that are to be backed up
```
qiuming-best, line 67: The code logic in `backupItemInternal ` is quite disorganized, with three methods—CSI snapshot, fs-backup, and native-snapshot—all intertwined. Have you considered reorganizing them, and creating separate functions or modules for each, to make the logic clearer for each respective backup method? @shubham-pampattiwar

shubham-pampattiwar: @qiuming-best Even if we convert to modules per backup method for volumes, the code path would remain the same right ?
like if object encountered is Pod, check for FSBackup option 
if PV object encountered then check for snapshot option
Additionally, I wanted to make changes without affecting the existing workflow that is properly working.

design/Extend-VolumePolicies-to-support-more-actions.md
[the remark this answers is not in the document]
reasonerjt: Resolving this comment as we agreed to use a more straight forward way to define the actions to the volumes.

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Non-Goals
+- No changes to existing approaches to opt-in/opt-out annotations for volumes
+- No changes to existing `VolumePolicies` functionalities
+
+
+## Use-cases/Scenarios
+
+**Use-case 1:**
+- A user wants to use `snapshot` (volumesnapshots) backup option for all the csi supported volumes and `fs-backup` for the rest of the volumes.
```
reasonerjt, line 28: Does it mean we don't want to distinguish `snapshot` and `csi-snapshot`?

When the action is set as "snapshot", velero is expected to use CSI snapshot by default, and fallback to "cloud provider snapshot API" when CSI snapshot is not possible, correct?

design/Extend-VolumePolicies-to-support-more-actions.md
reasonerjt: Resolving this comment as we agreed not to copy the opt-in/opt-out approach into volumepolicies.

reasonerjt: @shubham-pampattiwar @sseago 
I think this comment remains open.  We need to define the priority of "actions" in this design.  For example, in the spec of the backup CR the `snapshotVolumes` may be set to `true` but the `action` in volumepolicy is set to `fs-backup`, or vice versa.  Things can be more complicated if we take the `opt-in/out` annotations into account.

design/Extend-VolumePolicies-to-support-more-actions.md
reasonerjt: I understand it's too risky to rework the existing backup workflow, but I wish to suggest we introduce one data structure or struct `volumeHelper` to help we decide what to do with a PV or PVC.

Ideally, we should have `volumeHelper.GetAction(PV)` to tell us if we wanna take snapshot or use fs-backup or skip the volume, so volumeHelper should consolidate the content in volumepolicy and backup spec.

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,176 @@
...
+  - Here when we encounter [Persistent Volume as an item](https://github.com/vmware-tanzu/velero/blob/d4128542590470b204a642ee43311921c11db880/pkg/backup/item_backupper.go#L253)
+  - And we call the [takePVSnapshot func](https://github.com/vmware-tanzu/velero/blob/d4128542590470b204a642ee43311921c11db880/pkg/backup/item_backupper.go#L508)
+  - We need to modify the takePVSnapshot function to account for the `snapshot` VolumePolicy action.
+
+
+## Detailed Design
+- Update VolumePolicy action type validation to account for `fs-backup` and `snapshot` as valid VolumePolicy actions.
+- Modifications needed for `fs-backup` action:
+  - Now based on the specification of volume policy on backup request we will decide whether to go via legacy pod annotations approach or the newer volume policy based fs-backup action approach.
+  - If there is a presence of volume policy(fs-backup/snapshot)  on the backup request we use the newer volume policy approach to get the list of the volumes for `fs-backup` action
```
reasonerjt, line 77: I'm not very sure if we can "list" the volumes at the beginning or we may have to handle the volumes one by one based on the current backup flow.

weshayutin:
Some notes from latest community call:
https://hackmd.io/Jq6F5zqZR7S80CeDWUklkA#Discussion-topics

My personal understanding from the call last night is that:
1. VolumePolicies take precedence over legecy opt-in/opt-out labeling approaches 
2.  If there is no matching VolumePolicy on a volume, the velero defers to the opt-in/opt-out approach
3.  Finally if there is no match for a volumePolicy and no match for opt-in/out method, the velero server install default (FSBackupOrSnapShot) will be authoritative. 

@shubham-pampattiwar I'm off base I can edit or delete this comment.

design/Extend-VolumePolicies-to-support-more-actions.md
shubham-pampattiwar: yes you are correct, we are not adding `csi-snapshot` as a specific action in the design

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,176 @@
...
+  action:
+    type: fs-backup
+```
+## High-Level Design
+- When the VolumePolicy action is set as `fs-backup` the backup workflow modifications would be:
+  - We call [backupItem() -> backupItemInternal()](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L95) on all the items that are to be backed up
+  - Here when we encounter [Pod as an item ](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L195)
+  - We will have to modify the backup workflow to account for the `fs-backup` VolumePolicy action
+
+**Note:** `fs-backup` via VolumePolicy and `fs-backup` via annotation based opt-in/opt-out approach would be mutually exclusive.
```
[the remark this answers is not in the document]
shubham-pampattiwar: Added a note regarding this that the legacy workflow will be the fallback option if no matching fs-backup/snapshot action is found for a particular volume included in the backup.

shubham-pampattiwar:
Updated the PR based on the discussions in community call.

sseago:
@weshayutin "3. Finally if there is no match for a volumePolicy and no match for opt-in/out method, the velero server install default (FSBackupOrSnapShot) will be authoritative."

This is not quite correct. the defaultVolumesToFsBackup field is *part of* the opt-in/opt-out approach described in 2. So your first two bullets are the entirety of it. "defaultVolumesToFsBackup=true" -> "opt-out", "defaultVolumesToFsBackup=false" -> "opt-in"

design/Extend-VolumePolicies-to-support-more-actions.md
sseago: @reasonerjt @shubham-pampattiwar Right -- that is the current velero workflow for volumes that we're using "snapshot" for -- CSI if available, native if not, and nothing if neither is available. None of that should be changing with this. Basically for every volume backed up, Velero will either attempt fs backup or snapshot.

design/Extend-VolumePolicies-to-support-more-actions.md
sseago: @reasonerjt From what we agreed to on the call, policies take precedence here -- we need to be explicit in the design. For any volume that has a matching volume policy, the opt-in/out annotations and defaults will not be used. If a "skip" policy matches, the volume is skipped, if a "snapshot" policy matches, we use CSI or native snapshots, if a "fs-backup" policy matches, we use fs-backup. If none match, then we follow the opt-in/out logic based on annotations and defaultVolumesToFsBackup param.

sseago: As for priority among policy actions, the first matched policy should take priority.  See the above example policies -- the more specific example is first -- "gp2" storageclass volumes have "snapshot" policy, and then a catch-all "fs-backup" policy that matches all volumes. This means all gp2 volumes will be "snapshot", since that's the first match, and all non-gp2 volumes will be "fs-backup". Since these policies match *all* volumes, we don't have to resort to the fallback at all. If we didn't have that second policy, then all gp2 volumes would be "snapshot", and all other volumes would use the fallback of the legacy opt-in/out approach.

design/Extend-VolumePolicies-to-support-more-actions.md
reasonerjt: OK, I'm resolving this one.

design/Extend-VolumePolicies-to-support-more-actions.md
reasonerjt: Agreed.
@shubham-pampattiwar could you please update the design to explicitly explain the priority?

design/Extend-VolumePolicies-to-support-more-actions.md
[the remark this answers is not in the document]
sseago: @shawn-hurley I think that would be out-of-scope for this design. The task here is to provide an alternate means of telling Velero which volumes to use fs backup vs. which volumes to snapshot. We don't yet support "some datamover, others snapshot-without-datamover", much less "some this datamover, others that datamover" -- those selections are once-per-backup rather than once-per-volume, but in the future, it's possible that we could add volume-level policies for those new action types.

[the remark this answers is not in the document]
reasonerjt: In the initial design, there's a `Parameters` field under each `Action` struct, so `datamover` and other params can be set as parameters of the `snapshot` action type.

cc @qiuming-best

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,176 @@
...
+
+The `VolumePolicies` feature was introduced in Velero 1.11 as a flexible way to handle volumes. The main agenda of
+introducing the VolumePolicies feature was to improve the overall user experience when performing backup operations
+for volume resources, the feature enables users to group volumes according the `conditions` (criteria) specified and
+also lets you specify the `action` that velero needs to take for these grouped volumes during the backup operation.
+The limitation being that currently `VolumePolicies` only supports `skip` as an action, We want to extend the `action`
+functionality to support more usable options like `fs-backup` (File system backup) and `snapshot` (VolumeSnapshots).
+
+## Goals
+- Extending the VolumePolicies to support more actions like `fs-backup` (File system backup) and `snapshot` (VolumeSnapshots).
```
Lyndon-Li, line 17: Please also clarify how we can select data movement when the action is `snapshot`, it means how to express `snapshot-move-data` and `default-snapshot-move-date` in the volumePolicies.
Or if we decide not to support them, let's also clarify.

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,176 @@
...
+## High-Level Design
+- When the VolumePolicy action is set as `fs-backup` the backup workflow modifications would be:
+  - We call [backupItem() -> backupItemInternal()](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L95) on all the items that are to be backed up
+  - Here when we encounter [Pod as an item ](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L195)
+  - We will have to modify the backup workflow to account for the `fs-backup` VolumePolicy action
+
+
+- When the VolumePolicy action is set as `snapshot` the backup workflow modifications would be:
+  - Once again, We call [backupItem() -> backupItemInternal()](https://github.com/vmware-tanzu/velero/blob/main/pkg/backup/item_backupper.go#L95) on all the items that are to be backed up
+  - Here when we encounter [Persistent Volume as an item](https://github.com/vmware-tanzu/velero/blob/d4128542590470b204a642ee43311921c11db880/pkg/backup/item_backupper.go#L253)
```
Lyndon-Li, line 67: For CSI snapshot, the item is Persistent Volume Claim

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,176 @@
...
+  - And we call the [takePVSnapshot func](https://github.com/vmware-tanzu/velero/blob/d4128542590470b204a642ee43311921c11db880/pkg/backup/item_backupper.go#L508)
+  - We need to modify the takePVSnapshot function to account for the `snapshot` VolumePolicy action.
```
Lyndon-Li, line 69: For CSI snapshot, it is not taken in `takePVSnapshot`, but by CSI plugin through BIA

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Detailed Design
+- Update VolumePolicy action type validation to account for `fs-backup` and `snapshot` as valid VolumePolicy actions.
+- Modifications needed for `fs-backup` action:
+  - Now based on the specification of volume policy on backup request we will decide whether to go via legacy pod annotations approach or the newer volume policy based fs-backup action approach.
+  - If there is a presence of volume policy(fs-backup/snapshot)  on the backup request that matches as an action for a volume we use the newer volume policy approach to get the list of the volumes for `fs-backup` action
+  - Else continue with the annotation based legacy approach workflow.
+
+- Modifications needed for `snapshot` action:
+  - In the [takePVSnapshot function](https://github.com/vmware-tanzu/velero/blob/d4128542590470b204a642ee43311921c11db880/pkg/backup/item_backupper.go#L508) we will check the PV fits the volume policy criteria and see if the associated action is `snapshot`
```
Lyndon-Li, line 80: For CSI snapshot, it is taken by CSI plugin through BIA on the Persistent Volume Claim.
At present, the CSI plugin decides whether to process a volume through CSI snapshot method, Velero workflow doesn't know anything about how the decision is made. 
Then how to apply volumePolicies to CSI snapshots? It is not ideal if we have to have CSI plugin access volumePolicies, right?

Please clarify the solution in this design.

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,176 @@
...
+  - If it is not snapshot then we skip the further workflow and avoid taking the snapshot of the PV
+
+**Note:**
```
Lyndon-Li, line 83: Due to these complications, let's add some log to indicate how the backup method is decided.

design/Extend-VolumePolicies-to-support-more-actions.md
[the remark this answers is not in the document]
shawn-hurley: Extending the snapshot selection per backup with the parameters to define which data mover would be an excellent addition. This allows users to start at least building data movers.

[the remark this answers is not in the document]
Lyndon-Li: At present, we have the `--data-mover `parameter per backup, not sure if we need to make it per volume. Probably, we can keep the current behavior until we see the concrete requirement for the per volume one.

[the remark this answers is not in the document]
shawn-hurley: I honestly don't think that anyone will build one until you can do it by volume.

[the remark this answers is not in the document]
sseago: @shawn-hurley I agree that we want to consider this, but I still think it's out-of-scope for this particular design/implementation. I'm wary of scope creep here since we want to get this implemented in 1.14. The goal here is to use volume policies to replace the behavior currently managed via per-volume annotations to determine when to use fs-backup and when to snapshot. Following on from this we could propose an additional design to manage two additional things via volume policies:
1) on a per-volume/per-policy basis, use datamover (currently "use datamover vs . just CSI snapshots" is per-backup)
2) on a per-volume/per-policy basis, choose which datamover (this is also currently per-backup).

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,176 @@
...
+- Improve user experience when backing up Volumes via Velero
+
+## Non-Goals
+- No changes to existing approaches to opt-in/opt-out annotations for volumes
+- No changes to existing `VolumePolicies` functionalities
+
+
+## Use-cases/Scenarios
+
+**Use-case 1:**
```
rkasukurthi, line 27: We have a use case where we want to exclude all drivers by default and only include specific drivers for snapshotting. 
Can this be achieved with the current design?

rkasukurthi (changes requested):

design/Extend-VolumePolicies-to-support-more-actions.md
sseago: @rkasukurthi yes, driver is included in the existing selection mechanism for resource policies. This design reuses all of that and just adds new policy actions. See https://velero.io/docs/main/resource-filtering/#resource-policies

You'd probably use something like this:
```
- conditions:
    csi:
      driver: aws.ebs.csi.driver
  action:
    type: snapshot
- conditions:
  action:
    type: fs-backup
```
The above would snapshot ebs-csi volumes and use fs-backup for the rest -- if you want to skip completely any volumes not matching the specific drivers, then you'd use "skip" instead of "fs-backup".

design/Extend-VolumePolicies-to-support-more-actions.md
shubham-pampattiwar: `csi-snapshots` and `csi-datamover` are not targeted in the scope of this design @Lyndon-Li

Lyndon-Li: So both the design and implementation don't support CSI snapshot and data mover?
Personally, I think at least the design (including the detailed design which demonstrates the feasibility of the implementation) should support all backup types. Considering the fact that fs-backup is gradually replaced by data mover and native snapshot is gradually replaced by CSI snapshot, if we only support fs-backup and native snapshot, I am afraid the design will lose lots of its value.

sseago: @Lyndon-Li to be clear, we don't only support fs-backup and native snapshot, we support "fs-backup" and "snapshot", but "snapshot" will be either native snapshot or csi, following the same decision logic used with the legacy opt-in/opt-out approach when snapshot is chosen -- could be either csi or native snapshot. However, I discussed this with @shubham-pampattiwar and it does make sense to add some more specific actions for future implementation, but I think we should hold off on implementing those until the legacy opt-in/opt-out approach has been deprecated, otherwise things will be really confusing with the multiple ways of specifying things.

shubham-pampattiwar: @Lyndon-Li I have updated the design sections (non-goals and snapshot note). Also added future enhancement section, PTAL.

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,205 @@
...
+				continue
+			}
+		}
+```
+
+## Future Implementation
+It makes sense to add more specific actions in the future, once we deprecate the legacy opt-in/opt-out approach to keep things simple. Another point of note is, csi related action can be 
+easier to implement once we decide to merge the csi plugin into main velero code flow.
+In the future, we envision the following actions that can be implemented:
+- `snapshot-native`: only use volume snapshotter (native cloud provider snapshots), do nothing if not present/not compatible
```
reasonerjt, line 196: This is a little trivial b/c it's about future Impl, but IMO these 3 actions should be merged into the existing `snapshot` action and controlled by the `Parameters` of the action.

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,205 @@
...
+  - If it is not snapshot then we skip the csi BIA execute action and avoid taking the snapshot of the PVC by not invoking the csi plugin action for the PVC
+
+**Note:** 
+- When we are using the `VolumePolicy` approach for backing up the volumes then the volume policy criteria and action need to be specific and explicit, there is no default behaviour, if a volume matches `fs-backup` action then `fs-backup` method will be used for that volume and if similarly if the volume matches the criteria for `snapshot` action then the snapshot workflow will be used for the volume backup.
+- Another thing to note is the workflow proposed in this design uses the legacy opt-in/opt-out approach as a fallback option. For instance, the user specifies a VolumePolicy but for a particular volume included in the backup there are no actions(fs-backup/snapshot) matching in the volume policy for that volume, in such a scenario the legacy approach will be used for backing up the particular volume.
```
reasonerjt, line 90: In addition to `opt-in/opt-out`, there are settings like `snapshotVolumes` and `defaultVolumesToFsBackup` in the backup CR spec.  We should clarify that policies will take precedence, as discussed in [this thread](https://github.com/vmware-tanzu/velero/pull/6956/files#r1474485722)

design/Extend-VolumePolicies-to-support-more-actions.md
shubham-pampattiwar: @reasonerjt Added a note regarding this.

shubham-pampattiwar:
@reasonerjt Updated design implementation to use a `volumehelper` package, PTAL, Thanks !

sseago (dismissed):
I think this looks good now. We might want to be explicit that none of this new design replaces the `snaphotVolumes` param, that functions exactly as before. If it is `true`, then any volume identified as a snapshot volume (either via new policies or legacy opt-in/opt-out) will be snapshotted via either CSI or native VolumeSnapshotter. If it is `false`, then volumes identified as snapshot volumes will *not* be snapshotted. In other words, `snapshotVolumes` is a param that simply says "Yes I want to include snapshots in this backup for the volumes where it's appropriate", or "No, I don't want to include any snapshots in this backup".

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,255 @@
...
+
+- We will introduce a `volumehelper` package under which the `volume_policy_helper.go` will consist of the functions that we will use through the backup workflow to accommodate volume policies for PVs and PVCs
+
+#### FS-Backup
+- Regarding `fs-backup` action to decide whether to use legacy annotation based approach or volume policy based approach:
+    - We will use the `volumehelper.ProcessVolumePolicyFSBackup` function from the `volumehelper` package
+    - Functions involved in processing `fs-backup` volume policy action will somewhat look like:
+
+```go
+func ProcessVolumePolicyFSBackup(pod *corev1api.Pod, volumePolicy *resourcepolicies.Policies, defaultVolumesToFsBackup bool, backupExcludePVC bool) ([]string, []string, error) {
```
reasonerjt, line 103: A suggestion, instead of passing `defaultVolumesToFsBackup` and `backupExcludePVC`, what about passing the backup Spec as a whole so the func won't change when more flags are added to the backup CR(I hope not)?

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,255 @@
...
+- Another thing to note is the workflow proposed in this design uses the legacy opt-in/opt-out approach as a fallback option. For instance, the user specifies a VolumePolicy but for a particular volume included in the backup there are no actions(fs-backup/snapshot) matching in the volume policy for that volume, in such a scenario the legacy approach will be used for backing up the particular volume.
+## Implementation
+
+- The implementation should be included in velero 1.14
+
+- We will introduce a `volumehelper` package under which the `volume_policy_helper.go` will consist of the functions that we will use through the backup workflow to accommodate volume policies for PVs and PVCs
+
+#### FS-Backup
+- Regarding `fs-backup` action to decide whether to use legacy annotation based approach or volume policy based approach:
+    - We will use the `volumehelper.ProcessVolumePolicyFSBackup` function from the `volumehelper` package
```
reasonerjt, line 99: I think moving all related functions to one package is a good approach.
Can we make the funcs more consistent, for example, introduce an interface that has functions like
```
func ShouldDoFSBackupForPodVolume(...) (bool, error)
func ShouldDoSnapshotForPVC(...)(bool, error)
func ShouldDoXXXForXXX(...) (bool, error)
```

So that we can hide the complexity of settings, which may be in annotation, policy, backup CR, from the caller.

Lyndon-Li:
@shubham-pampattiwar 
I see there are still some opening comments, could you help to double check them? For the ones that are done, please resolve them; for the open ones, let's discuss in tomorrow's community meeting. Thanks!

design/Extend-VolumePolicies-to-support-more-actions.md
shubham-pampattiwar: Done, updated the design

design/Extend-VolumePolicies-to-support-more-actions.md
shubham-pampattiwar: extend-vol-policy-design

shubham-pampattiwar:
@Lyndon-Li Done, updated the design and resolved the comments.
@reasonerjt Latest iteration of design updates the implementation parts and introduces a `VolumeHelper` interface and a `VolumePolicyHelper` struct which implements the `VolumeHelper` interface.

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,304 @@
...
+- We will introduce a `VolumeHelper` interface. It will consist of two methods:
+  - `ShouldPerformFSBackupForPodVolume(pod *corev1api.Pod)`
+  - `ShouldPerformSnapshot(obj runtime.Unstructured)`
+```go
+type VolumeHelper interface {
+	ShouldPerformFSBackupForPodVolume(pod *corev1api.Pod) (bool, error)
+	ShouldPerformSnapshot(obj runtime.Unstructured) (bool, error)
+}
+```
+-  The `VolumePolicyHelper` struct will implement the `VolumeHelper` interface and will consist of the functions that we will use through the backup workflow to accomodate volume policies for PVs and PVCs.
```
reasonerjt, line 104: @shubham-pampattiwar 
I was thinking the `VolumeHelper` should be the source of truth for the backup, the return value should combine the value of volume-policy, fields in the backup, and the opt-in/opt-out annotations on the pod.  
In the case of fs-backup, the volumeHelper should be a replacement of `pdvolumeutil.GetVolumesByPod`
and tell the backupper what volumes should be handled via fs-backup.

Therefore, IMO the `VolumePolicyHelper` is a misuse b/c it only covers "volumePolicy".

Maybe we can clarify the volumeHelper's scope to "Help velero decide what to do with a volume during backup once the resource is selected by include/exclude filter and label selectors"

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,304 @@
...
+
+**Note:**
+- When we are using the `VolumePolicy` approach for backing up the volumes then the volume policy criteria and action need to be specific and explicit, there is no default behaviour, if a volume matches `fs-backup` action then `fs-backup` method will be used for that volume and if similarly if the volume matches the criteria for `snapshot` action then the snapshot workflow will be used for the volume backup.
+- Another thing to note is the workflow proposed in this design uses the legacy opt-in/opt-out approach as a fallback option. For instance, the user specifies a VolumePolicy but for a particular volume included in the backup there are no actions(fs-backup/snapshot) matching in the volume policy for that volume, in such a scenario the legacy approach will be used for backing up the particular volume.
+## Implementation
+
+- The implementation should be included in velero 1.14
+
+- We will introduce a `VolumeHelper` interface. It will consist of two methods:
+  - `ShouldPerformFSBackupForPodVolume(pod *corev1api.Pod)`
```
reasonerjt, line 96: If we have ` ProcessVolumePolicyFSBackup` (btw I think it should rename to `GetVolumesForFSBackup`)
This func is probably not needed.

design/Extend-VolumePolicies-to-support-more-actions.md
```diff
@@ -0,0 +1,304 @@
...
+		ib.trackSkippedPV(obj, kuberesource.PersistentVolumes, volumeSnapshotApproach, "does not satisfy the criteria for volume policy based snapshot action", log)
+		return nil
+	}
+```
+#### Snapshot (PVC)
+
+- Making sure that `snapshot` action is skipped for PVCs that do not fit the volume policy criteria, for this we will again use the `vph.ShouldPerformSnapshot` from the `VolumePolicyHelper(vph)` receiver.
+- We will pass the `VolumePolicyHelper(vph)` instance in `executeActions` method so that its available to use.
+```go
+func (v *VolumePolicyHelper) ShouldPerformSnapshot(obj runtime.Unstructured) (bool, error) {
```
reasonerjt, line 255: This function should also take other flags, like "snapshotVolumes" in the backup CR.

abh:
Thanks for working on this (and Velero). I greatly appreciate the backups Velero is managing for me. I came here via the issue referenced above.

For my use cases the fs-backup feature is fine for many use cases (and I appreciate that it's explicit and the configuration is close to the workload configuration).

I use snapshot backups for some workloads where I need a "as if the computer crashed" consistency of the file system and I use data mover to move the data off the infrastructure the cluster is running in. This PR doesn't seem to support that use case.

The feature I'm missing currently (and I can't figure out from the [resource filtering documentation](https://velero.io/docs/v1.13/resource-filtering/#resource-policies) is doing volume snapshots (and data mover) on some PVCs in a namespace instead of all.

For example I have a namespace running clickhouse databases and it's a mix of 1TB data that gets replaced within a day and about 100GB data that changes much less (and is more valuable).  I'd like to do snapshots and data mover on the 100GB data without also having to copy the 1TB other data.  (I can't easily divide them between namespaces as clickhouse or the CH operator has features that helps me generate and update the small data set from the large data set within the cluster).

Lyndon-Li:
@abh 
Do you mean you want to specify data mover as the backup type for some of the volumes? 
I think this has been discussed --- only `fs-backup` vs. `snapshot` could be specified per volume; then if `snapshot` is specified, you need to use the existing per backup option `--snapshot-move-data` to make it for data mover.
For sure, this would miss one functionality --- in a backup, among the volumes to go with `snapshot`, I want some volumes go with data mover, some not.

@shubham-pampattiwar Please  correct me if my understanding is wrong.

sseago:
@asaf-erlich If your goal is to back up with a mixture of fs-backup and datamover, that is absolutely supported by this design. Policies will identify volumes for fs-backup and snapshot, and setting `--snapshot-move-data=true` tells velero to use datamover for all snapshotted volumes. The only use case that's not supported in this design is if you want to use CSI snapshots without data movement for some volumes (leaving VolumeSnapshotContents in the cluster) and CSI snapshots with data movement for others, but from the use case you described, that doesn't seem relevant here.

design/Extend-VolumePolicies-to-support-more-actions.md
shubham-pampattiwar: Done, Updated the PR.

design/Extend-VolumePolicies-to-support-more-actions.md
shubham-pampattiwar: Done

design/Extend-VolumePolicies-to-support-more-actions.md
shubham-pampattiwar: Done, updated the PR

shubham-pampattiwar:
@reasonerjt PTAL, Thanks !

abh:
> Do you mean you want to specify data mover as the backup type for some of the volumes?

For me the important part is to be able to choose which volumes are being backed up. How I understand the current settings I can only choose by namespace when I want snapshot+data-mover.

In my use case I want all the snapshots to be with data mover (and I don't think there are any scenarios where fs-backup is better if snapshot+data mover is possible?).

For my use cases having the label on the pod (or on the PVC) is useful in many scenarios because it's so explicit. If the configuration is more complicated I'd worry I mistype a glob match or a label selector and accidentally don't have backups.

reasonerjt (approved):
lgtm, thanks

reasonerjt (approved):

Lyndon-Li (approved):

sseago (approved):

Issue #6595, opened 2023-08-03
On restore PV is not getting restored correctly in Azure AKS


soumyapattnaik:
**What steps did you take and what happened:**
I created a namespace soumya and created a deployment, pvc and pv(dynamically provisioned csi disk) in the namespace. The PV created was with retain data.
Then i triggered velero backup for the cluster.

After the backup is complete, i deleted the deployment, pvc, pv and the underlying azure disk and then retriggered restored

Post restore i see that deployment anf pvc is restored. PV is restored with the same name as it was backed up and the underlying disk was missing.

**What did you expect to happen:**

on restore the PV should have been created with a different name as it is dynamically provisioned. To fully confirm this behaviour i retriggered another restore from the same backup , but this time i restored the backed up resources in a different namespace soumya1. Post restore i could see deployment, pvc, pv(with a different name) and also the underlying disk getting created successfully.

Sharing the data post restore. During backup also the pv name was `pvc-41f06d44-68f8-4e82-a3e3-f479ae7c09c4`
```
PS C:\Users\sopattna\aks> kubectl get deployment --namespace=soumya
NAME                READY   UP-TO-DATE   AVAILABLE   AGE
nginx-deployment1   0/2     2            0           162m
nginx-deployment2   0/2     2            0           162m
PS C:\Users\sopattna\aks> kubectl get pvc --namespace=soumya
NAME             STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
pvc-azuredisk2   Bound    pvc-41f06d44-68f8-4e82-a3e3-f479ae7c09c4   10Gi       RWO            managed-csi    163m
PS C:\Users\sopattna\aks> kubectl get deployment --namespace=soumya1
NAME                READY   UP-TO-DATE   AVAILABLE   AGE
nginx-deployment1   2/2     2            2           157m
nginx-deployment2   2/2     2            2           157m
PS C:\Users\sopattna\aks> kubectl get pvc --namespace=soumya1
NAME             STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
pvc-azuredisk2   Bound    pvc-3eb1b795-b4d2-48bf-89d9-3ce518861d53   10Gi       RWO            managed-csi    158m
PS C:\Users\sopattna\aks> kubectl get pv
[...]

Issue #5834, opened 2023-02-06
Summarize volumes skipped during backup


rnarenpujari:
**Describe the problem/challenge you have**
Especially when performing a full cluster backup, it would be nice to get an idea of which volumes were not backed up in spite of the backup phase reporting Completed. Mainly for volume backup solutions like snapshots (csi and non csi) which only cover a subset of volumes on the cluster.

**Describe the solution you'd like**
May not be feasible to include this in the backup CR's status but would be nice to have in the backup describe (`--details`) output. 

**Anything else you would like to add:**
NA

**Environment:**

- Velero version (use `velero version`): 1.10.0
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

reasonerjt:
A volume could be skipped for various reasons:  doesn't meet certain selectors, "opt-out", skipped by CSI plugin if it's not provisioned by CSI drivers, etc.

@pradeepkchaturvedi  @rnarenpujari, I don't think we want to explicitly list the skipped volume in all cases, are there any very concrete scenarios we need to address ?

pradeepkchaturvedi:
@reasonerjt We can list of volumes which gets skip based on resource filter or pod annotation. Logs should state all methods used for volume backup (restic/kopia, CSI volume snapshot, storage provider volume snapshot) and list of volume backed up by each method.

reasonerjt:
We probably can introduce some tracker mechanism to move in/out the PV in the backup workflow, and this can be implemented in 2 phases:
1. Print the summary in the logs to get a summary what PVs are skipped.
2. Upload the information in the tracker to object store so we can show them in the output of backup describe.

rnarenpujari:
Hi. Thanks for working on this! I noticed the tracker doesn't work well if SnapshotVolumes is false in the backup. 

For example a backup with no volume backup methods configured: DefaultVolumesToFsBackup as false (opt-in) and SnapshotVolumes as false.
[...]

PR #6962, opened 2023-10-16
Add the PV backup information design document.

blackpiglet:

Fixes #(issue)
This design is used to add the design document to address these issues:
https://github.com/vmware-tanzu/velero/labels/pv-backup-info

### DIFF
--- a/design/pv_backup_info.md
+++ b/design/pv_backup_info.md
@@ -1,0 +1,192 @@
+# PersistentVolume backup information design
+
+## Abstract
+Create a new metadata file in the backup repository's backup name sub-directory to store the backup-including PVC and PV information. The information includes the way of backing up the PVC and PV data, snapshot information, and status. The needed snapshot status can also be recorded there, but the Velero-Native snapshot plugin doesn't provide a way to get the snapshot size from the API, so it's possible that not all snapshot size information is available.
+
+This new additional metadata file is needed when:
+* Get a summary of the backup's PVC and PV information, including how the data in them is backed up, or whether the data in them is skipped from backup.
+* Find out how the PVC and PV should be restored in the restore process.
+* Retrieve the PV's snapshot information for backup.
+
+## Background
+There is already a [PR](https://github.com/vmware-tanzu/velero/pull/6496) to track the skipped PVC in the backup. This design will depend on it and go further to get a summary of PVC and PV information, then persist into a metadata file in the backup repository.
+
+In the restore process, the Velero server needs to decide how the PV resource should be restored according to how the PV is backed up. The current logic is to check whether it's backed up by Velero-native snapshot, by file-system backup, or having `DeletionPolicy` set as `Delete`.
+
+The checks are made by the backup-generated PVBs or Snapshots. There is no generic way to find this information, and the CSI backup and Snapshot data movement backup are not covered.
+
+Another thing that needs noticing is when describing the backup, there is no generic way to find the PV's snapshot information.
+
+## Goals
+- Create a new metadata file to store backup's PVCs and PVs information and volume data backing up method. The file can be used to let downstream consumers generate a summary.
+- Create a generic way to let the Velero server know how the PV resources are backed up.
+- Create a generic way to let the Velero server find the PV corresponding snapshot information.
+
+## Non Goals
+- Unify how to get snapshot size information for all PV backing-up methods, and all other currently not ready PVs' information.
+
+## High-Level Design
+Create _backup-name_-volumes-info.json metadata file in the backup's repository. This file will be encoded to contain all the PVC and PV information included in the backup. The information covers whether the PV or PVC's data is skipped during backup, how its data is backed up, and the backed-up detail information.
+
+Please notice that the new metadata file includes all skipped volume information. This is used to address [the second phase needs of skipped volumes information](https://github.com/vmware-tanzu/velero/issues/5834#issuecomment-1526624211).
+
+The `restoreItem` function can decode the _backup-name_-volumes-info.json file to determine how to handle the PV resource. 
+
+## Detailed Design
+
+### The VolumeInfo structure
+_backup-name_-volumes-info.json file is a structure that contains an array of structure `VolumeInfo` and a VolumeInfo version parameter. The version is used to support multiple VolumeInfo version, and make the backward-compatible support possible. The current version is `1`.
+
+The 1 version of `VolumeInfo` definition is:
+``` golang
+type VolumeInfoV1 struct {
+    PVCName        string    // The PVC's name. The format should be <namespace-name>/<PVC-name>
+    PVName         string    // The PV name.
+    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup`, `CSISnapshot` and `Skipped`.
+    SnapshotDataMovement bool // Whether the volume's snapshot data is moved to specified storage.
+
+    SkippedReason   string    // The reason for the volume is skipped in the backup.
+    StartTimestamp  *metav1.Time // Snapshot starts timestamp.
+
+    CSISnapshotInfo          CSISnapshotInfo
+    SnapshotDataMoveInfo     SnapshotDataMoveInfo
+    NativeSnapshotInfo       VeleroNativeSnapshotInfo
+    PVBInfo                  PodVolumeBackupInfo
+}
+
+// CSISnapshotInfo is used for displaying the CSI snapshot status
+type CSISnapshotInfo struct {
+    SnapshotHandle  string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
+    Size            int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
+
+    Driver          string  // The name of the CSI driver.
+    VSCName         string // The name of the VolumeSnapshotContent. 
+}
+
+// SnapshotDataMoveInfo is used for displaying the snapshot data mover status.
+type SnapshotDataMoveInfo struct {
+    DataMover        string    // The data mover used by the backup. The valid values are `velero` and ``(equals to `velero`).
+    UploaderType     string    // The type of the uploader that uploads the snapshot data. The valid values are `kopia` and `restic`. It's useful for file-system backup and snapshot data mover.
+    RetainedSnapshot string    // The name or ID of the snapshot associated object(SAO).
+}
+
+// VeleroNativeSnapshotInfo is used for displaying the Velero native snapshot status.
+type VeleroNativeSnapshotInfo struct {
+    SnapshotHandle      string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
+    Size                int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
+
+    VolumeType string    // The cloud provider snapshot volume type.
+    VolumeAZ   string    // The cloud provider snapshot volume's availability zones.
+    IOPS       string    // The cloud provider snapshot volume's IOPS.
+}
+
+// PodVolumeBackupInfo is used for displaying the PodVolumeBackup snapshot status.
+type PodVolumeBackupInfo struct {
+    SnapshotHandle      string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
+    Size                int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
+
+    UploaderType  string    // The type of the uploader that uploads the data. The valid values are `kopia` and `restic`. It's useful for file-system backup and snapshot data mover.
+    VolumeName    string // The PVC's corresponding volume name used by Pod
+    PodName       string // The Pod name mounting this PVC. The format should be <namespace-name>/<pod-name>.
+}
+```
+
+To make Velero support multiple versions of VolumeInfo, Velero needs to create a structure to include the version and the versioned volume information together. The information is persisted in the metadata file during backup creation. When reading the VolumeInfo metadata file, Velero reads the version information first by un-marshalling the metadata file by structure `VolumeInfoVersion`, then decide to use which version of the VolumeInfos according to the version value.
+
+If there is need to introduce a non compatible change to the VolumeInfo, a new version of `VolumeInfos` and `VolumeInfo` are needed. For example, version 2 is created, then `VolumeInfosV2` and `VolumeInfoV2` structures are needed.
+
+Only when there a non-backward-compatible change introduced in the `VolumeInfo` structure, the `version`'s version will be incremented.
+
+``` golang
+type VolumeInfoVersion struct {
+    Version string 
+}
+
+type VolumeInfosV2 struct {
+    Infos   []VolumeInfoV2
+    Version string // VolumeInfo structure's version information.
+}
+
+type VolumeInfoV2 struct {
+    ...
+}
+```
+
+### How the VolumeInfo array is generated.
+The function `persistBackup` has `backup *pkgbackup.Request` in parameters.
+From it, the `VolumeSnapshots`, `PodVolumeBackups`, `CSISnapshots`, `itemOperationsList`, and `SkippedPVTracker` can be read. All of them will be iterated and merged into the `VolumeInfo` array, and then persisted into backup repository in function `persistBackup`.
+
+Please notice that the change happened in async operations are not reflected in the new metadata file. The file only covers the volume changes happen in the Velero server process scope.
+
+A new methods are added to BackupStore to download the VolumeInfo metadata file.
+Uploading the metadata file is covered in the exiting `PutBackup` method.
+
+``` golang
+type BackupStore interface {
+    ...
+    GetVolumeInfos(name string) ([]*VolumeInfo, error)
+    ...
+}
+```
+
+### How the VolumeInfo array is used.
+
+#### Generate the PVC backed-up information summary
+The downstream tools can use this VolumeInfo array to format and display their volume information. This is in the scope of this feature.
+
+#### Retrieve volume backed-up information for `velero backup describe` command
+The `velero backup describe` can also use this VolumeInfo array structure to display the volume information. The snapshot data mover volume should use this structure at first, then the Velero native snapshot, CSI snapshot, and PodVolumeBackup can also use this structure. The detailed implementation is also not in this feature's scope.
+
+#### Let restore know how to restore the PV
+In the function `restoreItem`, it will determine whether to restore the PV resource by checking it in the Velero native snapshots list, PodVolumeBackup list, and its DeletionPolicy. This logic is still kept. The logic will be used when the new `VolumeInfo` metadata cannot be found to support backward compatibility.
+
+``` golang
+	if groupResource == kuberesource.PersistentVolumes {
+		switch {
+		case hasSnapshot(name, ctx.volumeSnapshots):
+            ...
+        case hasPodVolumeBackup(obj, ctx):
+            ...
+        case hasDeleteReclaimPolicy(obj.Object):
+            ...
+        default:
+            ...
+```
+
+After introducing the VolumeInfo array, the following logic will be added.
+``` golang
+	if groupResource == kuberesource.PersistentVolumes {
+        volumeInfo := GetVolumeInfo(pvName)
+		switch volumeInfo.BackupMethod {
+		case VeleroNativeSnapshot:
+            ...
+        case PodVolumeBackup:
+            ...
+        case CSISnapshot:
+            ...
+        case Skipped:
+            // Check whether the Velero server should restore the PV depending on the DeletionPolicy setting.
+        default:
+            // Need to check whether the volume is backed up by the SnapshotDataMover.
+            if volumeInfo.SnapshotDataMovement:
+```
+
+### How the VolumeInfo metadata file is deleted
+_backup-name_-volumes-info.json file is deleted during backup deletion.
+
+## Alternatives Considered
+The restore process needs more information about how the PVs are backed up to determine whether this PV should be restored. The released branches also need a similar function, but backporting a new feature into previous releases may not be a good idea, so according to [Anshul Ahuja's suggestion](https://github.com/vmware-tanzu/velero/issues/6595#issuecomment-1731081580), adding more cases here to support checking PV backed-up by CSI plugin and CSI snapshot data mover: https://github.com/vmware-tanzu/velero/blob/5ff5073cc3f364bafcfbd26755e2a92af68ba180/pkg/restore/restore.go#L1206-L1324.
+
+## Security Considerations
+There should be no security impact introduced by this design.
+
+## Compatibility
+After this design is implemented, there should be no impact on the existing [skipped PVC summary feature](https://github.com/vmware-tanzu/velero/pull/6496).
+
+To support older version backup, which doesn't have the VolumeInfo metadata file, the old logic, which is checking the Velero native snapshots list, PodVolumeBackup list, and PVC DeletionPolicy, is still kept, and supporting CSI snapshots and snapshot data mover logic will be added too.
+
+## Implementation
+This will be implemented in the Velero v1.13 development cycle.
+
+## Open Issues
+There are no open issues identified by now.
--- a/changelogs/unreleased/6962-blackpiglet
+++ b/changelogs/unreleased/6962-blackpiglet
@@ -1,0 +1,1 @@
+Add the PV backup information design document.


design/pv_backup_info.md
```diff
@@ -0,0 +1,144 @@
+# PersistentVolume backup information design
+
+## Abstract
+Create a new metadata file in the backup repository directory to store the backup including PVC and PV information. The information includes the way of backing up the PVC and PV data, snapshot information, and status. The needed snapshot status can also be recorded there, but the Velero-Native snapshot plugin doesn't provide a way to get the snapshot size from the API, so it's possible that not all snapshot size information is available.
```
ywk253100, line 4: Is the metadata stored in backup folder or backup repository folder?

design/pv_backup_info.md
```diff
@@ -0,0 +1,144 @@
...
+
+## High-Level Design
+Create _backup-name_-volumes-info.json metadata file in the backup's repository. This file will be encoded to contain all the PVC and PV information included in the backup. The information covers whether the PV or PVC's data is skipped during backup, how its data is backed up, and the backed-up detail information.
+
+The `restoreItem` function can decode the _backup-name_-volumes-info.json file to determine how to handle the PV resource. 
+
+## Detailed Design
+
+### The VolumeInfo structure
+_backup-name_-volumes-info.json file is an array of structure `VolumeInfo`.
```
ywk253100, line 36: Is it easier to handle if organize the data structure as a map (use PV name as the key)?

design/pv_backup_info.md
```diff
@@ -0,0 +1,144 @@
...
+
+The `VolumeInfo` definition is:
+``` golang
+type VolumeInfo struct {
```
ywk253100, line 40: Is it better if the data structure is more structured? For example, organize file-system backup related properties as a structure and snapshot related properties as another structure

design/pv_backup_info.md
blackpiglet: The metadata file is stored under the backup sub-directory, same as the backup's log file <backup-name>-logs.gz

design/pv_backup_info.md
```diff
@@ -0,0 +1,144 @@
...
+    PVCName        string    // The PVC's name. The format should be <namespace-name>/<PVC-name>
+    PVName         string    // The PV name.
+    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup`, `CSISnapshot`, `SnapshotDataMover`, and `Skipped`.
```
ywk253100, line 43: Among all these properties, only the `BackupMethod` is required for performing the restoration, right?
Is it needed to record these kinds of properties in a separate way? For example, we could record this information in the backed-up resource as an annotation, and use the annotation during the restoration instead. In this way, the volume metadata file is only used to show information and the restore can work without it.

blackpiglet: Yeah, only the `PV` name and the `BackupMethod` are functional for the restore process. The others are used to display information.

But I'm not sure why separating backup metadata and the backup data will have an impact on this.

design/pv_backup_info.md
blackpiglet: I thought about that, but different scenarios need different keys. The restore process needs the key to be the PV name and the summary-generating scenario needs the key to be the PVC name.

Maybe storing the information as an array, and then creating the proper map while using it is better.

design/pv_backup_info.md
```diff
@@ -0,0 +1,144 @@
...
+            // This is not expected, print error.
+```
+
+## Alternatives Considered
+The restore process needs more information about how the PVs are backed up to determine whether this PV should be restored. The released branches also need a similar function, but backporting a new feature into previous releases may not be a good idea, so according to [Anshul Ahuja's suggestion](https://github.com/vmware-tanzu/velero/issues/6595#issuecomment-1731081580), adding more cases here to support checking PV backed-up by CSI plugin and CSI snapshot data mover: https://github.com/vmware-tanzu/velero/blob/5ff5073cc3f364bafcfbd26755e2a92af68ba180/pkg/restore/restore.go#L1206-L1324.
+
+## Security Considerations
+There should be no security impact introduced by this design.
+
+## Compatibility
```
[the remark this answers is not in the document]
blackpiglet: I'm not sure about that topic yet.
As the `alternative consideration` states, the old version of Velero will have different logic to handle the PV restore, but the backup generated by the old version of Velero doesn't have the new metadata file, so the new version of Velero may not restore the old backup correctly.

To fix that, we need the same logic as the old release logic, which is adding code to check the VolumeSnapshots and ItemOperations, alongside the VolumeInfo metadata checking logic.
Looks like it doesn't make sense to have both the old and new logic in the main branch.

design/pv_backup_info.md
blackpiglet: Reasonable, I will modify it.

design/pv_backup_info.md
ywk253100: Please update the line `Create a new metadata file in the backup repository directory` to correct the statement

design/pv_backup_info.md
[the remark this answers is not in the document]
ywk253100: We should make sure the backup created by old version Velero can be restored with the new version Velero. As we disscussed offline, we may use the backup version to distinguish different situations and use different logics to handle both of them.

design/pv_backup_info.md
blackpiglet: After discussing with Yonghui, we reached an agreement that having a new metadata file is a better choice.

design/pv_backup_info.md
blackpiglet: Modified.

design/pv_backup_info.md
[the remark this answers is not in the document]
blackpiglet: Modified. 
After a discussion with Daniel, we reached an agreement that there is no need to bump the backup FormatVersion and check the FormatVersion before falling back to the old restore PV logic.

design/pv_backup_info.md
```diff
@@ -0,0 +1,155 @@
...
+    CompletionTimestamp *metav1.Time // Snapshot completes timestamp.
+
+    DataMoverInfo       DataMoverInfo
+    NativeSnapshotInfo  VeleroNativeSnapshotInfo
+    PVBInfo             PodVolumeBackupInfo
+}
+
+// DataMoverInfo is used for displaying the snapshot data mover and PodVolumeBackup snapshot status.
+type DataMoverInfo struct {
+    DataMover        string    // The name of the data mover that uploads the snapshot data. The valid values are `kopia` and `restic`. It's useful for file-system backup and snapshot data mover.
```
Lyndon-Li, line 60: DataMover is the name of the data mover being used by the backup, the valid value is "velero", ""(equals to "velero") or a string of the customized data mover name.
It is different from UploaderType which should be `kopia` and `restic`.

So here we can have both, DataMover and UploaderType and the latter should be for both fs backup and data mover backup.

design/pv_backup_info.md
```diff
@@ -0,0 +1,155 @@
...
+The `restoreItem` function can decode the _backup-name_-volumes-info.json file to determine how to handle the PV resource. 
+
+## Detailed Design
+
+### The VolumeInfo structure
+_backup-name_-volumes-info.json file is an array of structure `VolumeInfo`.
+
+The `VolumeInfo` definition is:
+``` golang
+type VolumeInfo struct {
```
Lyndon-Li, line 40: Since we are persisting this to the backup repo, do we need a self-described version info?
For example, in future, when we want to modify this struct, will the new code/struct be able to interpret the data in the repo created by old version of Velero?

design/pv_backup_info.md
```diff
@@ -0,0 +1,155 @@
...
+    PVCName        string    // The PVC's name. The format should be <namespace-name>/<PVC-name>
+    PVName         string    // The PV name.
+    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup`, `CSISnapshot`, `SnapshotDataMover`, and `Skipped`.
+
+    SkippedReason  string    // The reason for the volume is skipped in the backup.
```
Lyndon-Li, line 45: Do we need this? I thought here we only recorded the volumes that have been backed up completely.
I think we have other existing mechanisms to track the full list of volumes vs. the skipped ones.

design/pv_backup_info.md
```diff
@@ -0,0 +1,155 @@
...
+
+    DataMoverInfo       DataMoverInfo
+    NativeSnapshotInfo  VeleroNativeSnapshotInfo
+    PVBInfo             PodVolumeBackupInfo
+}
+
+// DataMoverInfo is used for displaying the snapshot data mover and PodVolumeBackup snapshot status.
+type DataMoverInfo struct {
+    DataMover        string    // The name of the data mover that uploads the snapshot data. The valid values are `kopia` and `restic`. It's useful for file-system backup and snapshot data mover.
+    RetainedSnapshot string    // The ID of the local snapshot.
```
Lyndon-Li, line 61: This is a snapshot type specific name that identifies the retained snapshot, not always be the ID of snapshots. E.g., for CSI snapshot, this value could be the name of the VSC.

design/pv_backup_info.md
```diff
@@ -0,0 +1,155 @@
...
+``` golang
+type BackupStore interface {
+    ...
+    PutVolumeInfos(backup string, volumeInfos io.Reader) error
+    GetVolumeInfos(name string) ([]*VolumeInfo, error)
+    ...
+}
+```
+
+There are two phases to handle the VolumeInfo array.
```
Lyndon-Li, line 91: Let's also describe when and how this data is deleted from the backup repo.

blackpiglet: The metadata file is deleted when the backup is deleted.
I will document it in the design.

design/pv_backup_info.md
blackpiglet: How about changing to:
`The name or ID of the snapshot associated object(SAO).`

design/pv_backup_info.md
blackpiglet: I think we still need this.
There is already a PR to track the skipped volumes information, but the information is not recorded anywhere. It's collected and printed into the log during backup creation. This is the part two of the requirement.
https://github.com/vmware-tanzu/velero/issues/5834#issuecomment-1526624211

design/pv_backup_info.md
Lyndon-Li: Actually, the DMP knows neither name nor ID or the SAO, so the logics is like -- DMP creates an alias -> DM assigns this alias to the created SAO -> DMP uses this alias later on

design/pv_backup_info.md
Lyndon-Li: So does this mean we will record all the volumes in the backup, including the skipped ones?

design/pv_backup_info.md
Lyndon-Li: yes, fine to me

design/pv_backup_info.md
```diff
@@ -0,0 +1,155 @@
...
+
+First, the VolumeInfo array is generated before the backup is persisted by this function `persistBackup`. Then the VolumeInfo array will be passed to `persistBackup` function as a parameter. At this phase, the basic information, the skipped information, the Velero native snapshot information, and the PodVolumeBackup information are ready.
+
+Second, in the backup finalize controller, the snapshot data mover and CSI snapshot information will be updated according to the BackupItemOperations result. The VolumeInfo will be updated by the method `PutVolumeInfos`.
```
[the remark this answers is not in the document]
blackpiglet: The `Status` and the `CompletionTimestamp` of the async operations are not updated before the finalizing phase. If we can accept that information is no synced with the operations final state, the second phase is not needed.

design/pv_backup_info.md
blackpiglet: Thanks. I will update the structures.

design/pv_backup_info.md
blackpiglet: Reasonable. I will add a new version field.

design/pv_backup_info.md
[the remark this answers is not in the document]
Lyndon-Li: If there is no specific users of these two fields, I would suggest we remove them and remove the second step.

design/pv_backup_info.md
blackpiglet: Yes. I will make the change.

design/pv_backup_info.md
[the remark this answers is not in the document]
blackpiglet: After discussion, the decision is to remove the fields that need to wait for async operation completion.
The `Status` and `CompletionTimestamp` fields are removed.

design/pv_backup_info.md
blackpiglet: Modified.

Lyndon-Li (dismissed):

Lyndon-Li (approved):

ywk253100 (dismissed):

design/pv_backup_info.md
```diff
@@ -0,0 +1,174 @@
...
+}
+
+type VolumeInfosV10 struct {
+    Infos   []VolumeInfoV10
+    Version string // VolumeInfo structure's version information.
+}
+```
+
+### How the VolumeInfo array is generated.
+
```
reasonerjt, line 97: Here you didn't really describe how this array is **generated** 
To specify how it is **generated**  you need to provide information like how to extend the capabilities of `skippedPVTracker`, etc.
One challenge I can think about is how to aggregate information in different categories into one list, esp considering we may process different PVs in parallel.

design/pv_backup_info.md
```diff
@@ -0,0 +1,174 @@
...
+The VolumeInfo array is generated before the backup is persisted by this function `persistBackup`. Then the VolumeInfo array will be passed to `persistBackup` function as a parameter.
+
+Please notice that the change happened in async operations are not reflected in the new metadata file. The file only covers the volume changes happen in the Velero server process scope.
+
+### How the VolumeInfo array is used.
+
+#### Generate the PVC backed-up information summary
+The upstream tools can use this VolumeInfo array to format and display their volume information. This is in the scope of this feature.
+
+#### Retrieve volume backed-up information for `velero backup describe` command
```
reasonerjt, line 118: Just to clarify in v1.13 we only read the pvInfo array proposed in this doc when necessary, and it should be only to address new use cases right?

Do we plan to re-write any code in `backup/restore describe` or the func `restoreItem`?

design/pv_backup_info.md
```diff
@@ -0,0 +1,174 @@
...
+
+### The VolumeInfo structure
+_backup-name_-volumes-info.json file is a structure that contains an array of structure `VolumeInfo` and a VolumeInfo version parameter. The version is used to support multiple VolumeInfo version, and make the backward-compatible support possible. The current version is `1.0`.
+
+The 1.0 version of `VolumeInfo` definition is:
+``` golang
+type VolumeInfoV10 struct {
+    PVCName        string    // The PVC's name. The format should be <namespace-name>/<PVC-name>
+    PVName         string    // The PV name.
+    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup`, `CSISnapshot`, `SnapshotDataMover`, and `Skipped`.
```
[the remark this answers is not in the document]
Lyndon-Li: I think `SnapshotDataMover` is a concept of design, from user's view there is no calling of `SnapshotDataMover`.
From users' perspective, `CSISnapshotDataMovement` is a backup method, it means take CSI snapshot + move data.

Does this sound better?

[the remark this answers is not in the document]
Lyndon-Li: > Shall we consider the data movement an additional action after the snapshot?

From the current CSI Snapshot Data Movement doc, we didn't deliver this info/concept to users. `CSISnapshotDataMovement` means a separate backup type.

design/pv_backup_info.md
```diff
@@ -0,0 +1,174 @@
...
+
+    SkippedReason  string    // The reason for the volume is skipped in the backup.
+
+    SnapshotHandle      string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
```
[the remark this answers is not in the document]
Lyndon-Li: For CSISnapshotDataMovement backup, there is a further field called `RetainedSnapshot` in `DataMoverInfo`, this is to fulfill the purpose to record the local snapshot.
In this way, it is basically fine for `SnapshotHandle` to take the snapshot ID in the backup repo

design/pv_backup_info.md
blackpiglet: `restoreItem` is in the scope of v1.13 because it's used to address issue #6595.
`backup/restore describe` is not in the plan yet.

design/pv_backup_info.md
```diff
@@ -0,0 +1,174 @@
...
+There is already a [PR](https://github.com/vmware-tanzu/velero/pull/6496) to track the skipped PVC in the backup. This design will depend on it and go further to get a summary of PVC and PV information, then persist into a metadata file in the backup repository.
+
+In the restore process, the Velero server needs to decide how the PV resource should be restored according to how the PV is backed up. The current logic is to check whether it's backed up by Velero-native snapshot, by file-system backup, or having `DeletionPolicy` set as `Delete`.
+
+The checks are made by the backup-generated PVBs or Snapshots. There is no generic way to find this information, and the CSI backup and Snapshot data movement backup are not covered.
+
+Another thing that needs noticing is when describing the backup, there is no generic way to find the PV's snapshot information.
+
+## Goals
+- Create a new metadata file to store backup's PVCs and PVs information and volume data backing up method. The file can be used to let upstream consumers generate a summary.
```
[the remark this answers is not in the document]
blackpiglet: Thanks. Modified.

design/pv_backup_info.md
```diff
@@ -0,0 +1,174 @@
...
+Please notice that the new metadata file includes all skipped volume information. This is used to address [the second phase needs of skipped volumes information](https://github.com/vmware-tanzu/velero/issues/5834#issuecomment-1526624211).
+
+The `restoreItem` function can decode the _backup-name_-volumes-info.json file to determine how to handle the PV resource. 
+
+## Detailed Design
+
+### The VolumeInfo structure
+_backup-name_-volumes-info.json file is a structure that contains an array of structure `VolumeInfo` and a VolumeInfo version parameter. The version is used to support multiple VolumeInfo version, and make the backward-compatible support possible. The current version is `1.0`.
+
+The 1.0 version of `VolumeInfo` definition is:
```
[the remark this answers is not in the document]
blackpiglet: Agreed. Only keeping the major version should be enough, and the major version only changed when there is a non-backward-compatible change introduced.

design/pv_backup_info.md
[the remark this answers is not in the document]
blackpiglet: `CSISnapshotInfo` is added. `SnapshotDataMover` is changed to `CSISnapshotDataMovement`.

design/pv_backup_info.md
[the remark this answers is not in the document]
blackpiglet: After taking another thought, it looks reasonable to move the SnapshotHandle and the Size into separate data structures, because the SnapshotHandle's format may be different for different backup methods.

design/pv_backup_info.md
```diff
@@ -0,0 +1,174 @@
...
+``` golang
+type VolumeInfoV10 struct {
+    PVCName        string    // The PVC's name. The format should be <namespace-name>/<PVC-name>
+    PVName         string    // The PV name.
+    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup`, `CSISnapshot`, `SnapshotDataMover`, and `Skipped`.
+
+    SkippedReason  string    // The reason for the volume is skipped in the backup.
+
+    SnapshotHandle      string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
+    Size                int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
```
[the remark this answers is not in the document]
blackpiglet: Modified.

design/pv_backup_info.md
blackpiglet: The plan is to generate the array in the function `persistBackup`.
The function `persistBackup` has `backup *pkgbackup.Request` in parameters.
From it, the `VolumeSnapshots`, `PodVolumeBackups`, `CSISnapshots`, `itemOperationsList`, and `SkippedPVTracker` can be read. All of them will be iterated and merged into the `VolumeInfo` array.

reasonerjt (approved):

Lyndon-Li (approved):

design/pv_backup_info.md
```diff
@@ -0,0 +1,192 @@
...
+
+### The VolumeInfo structure
+_backup-name_-volumes-info.json file is a structure that contains an array of structure `VolumeInfo` and a VolumeInfo version parameter. The version is used to support multiple VolumeInfo version, and make the backward-compatible support possible. The current version is `1`.
+
+The 1 version of `VolumeInfo` definition is:
+``` golang
+type VolumeInfoV1 struct {
+    PVCName        string    // The PVC's name. The format should be <namespace-name>/<PVC-name>
+    PVName         string    // The PV name.
+    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup`, `CSISnapshot` and `Skipped`.
```
draghuram, line 45: It doesn't sound reasonable for "BackupMethod" field to accept "Skipped" as a value. It may be better to have a different field to capture the state of the backup such as "Completed", "Skipped", "Failed" etc. BTW, would "Skipped" mean failed as well or that it is skipped for some reason?

design/pv_backup_info.md
```diff
@@ -0,0 +1,192 @@
...
+    SnapshotDataMovement bool // Whether the volume's snapshot data is moved to specified storage.
+
+    SkippedReason   string    // The reason for the volume is skipped in the backup.
+    StartTimestamp  *metav1.Time // Snapshot starts timestamp.
+
+    CSISnapshotInfo          CSISnapshotInfo
+    SnapshotDataMoveInfo     SnapshotDataMoveInfo
```
draghuram, line 52: Just wanted to make sure this is not a typo. Wouldn't "SnapshotDataMoverInfo" is better?

design/pv_backup_info.md
```diff
@@ -0,0 +1,192 @@
...
+    NativeSnapshotInfo       VeleroNativeSnapshotInfo
+    PVBInfo                  PodVolumeBackupInfo
+}
+
+// CSISnapshotInfo is used for displaying the CSI snapshot status
+type CSISnapshotInfo struct {
+    SnapshotHandle  string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
```
draghuram, line 59: I don't understand why this would be anything other than snapshot handle as stored in volume snapshot content. Since the struct name is CSISnapshotInfo and there is a CSI snapshot field called "snapshot handle", I think it would be very confusing to use this field for uploader snapshot ID.

design/pv_backup_info.md
```diff
@@ -0,0 +1,192 @@
...
+    Size            int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
+
+    Driver          string  // The name of the CSI driver.
+    VSCName         string // The name of the VolumeSnapshotContent. 
+}
+
+// SnapshotDataMoveInfo is used for displaying the snapshot data mover status.
+type SnapshotDataMoveInfo struct {
```
draghuram, line 67: SnapshotDataMoverInfo?

design/pv_backup_info.md
```diff
@@ -0,0 +1,192 @@
...
+    DataMover        string    // The data mover used by the backup. The valid values are `velero` and ``(equals to `velero`).
+    UploaderType     string    // The type of the uploader that uploads the snapshot data. The valid values are `kopia` and `restic`. It's useful for file-system backup and snapshot data mover.
+    RetainedSnapshot string    // The name or ID of the snapshot associated object(SAO).
```
draghuram, line 70: I think this field should really be ID of the uploader snapshot.

design/pv_backup_info.md
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// VeleroNativeSnapshotInfo is used for displaying the Velero native snapshot status.
+type VeleroNativeSnapshotInfo struct {
+    SnapshotHandle      string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
```
draghuram, line 75: Same comment as for CSISnapshotInfo struct.

design/pv_backup_info.md
```diff
@@ -0,0 +1,192 @@
...
+    Size                int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
+
+    VolumeType string    // The cloud provider snapshot volume type.
+    VolumeAZ   string    // The cloud provider snapshot volume's availability zones.
+    IOPS       string    // The cloud provider snapshot volume's IOPS.
+}
+
+// PodVolumeBackupInfo is used for displaying the PodVolumeBackup snapshot status.
+type PodVolumeBackupInfo struct {
+    SnapshotHandle      string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
```
draghuram, line 85: Isn't this always snapshot ID of the uploader? I am wondering when it can be cloud provider snapshot.

design/pv_backup_info.md
```diff
@@ -0,0 +1,192 @@
...
+    Size                int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
+
+    UploaderType  string    // The type of the uploader that uploads the data. The valid values are `kopia` and `restic`. It's useful for file-system backup and snapshot data mover.
+    VolumeName    string // The PVC's corresponding volume name used by Pod
+    PodName       string // The Pod name mounting this PVC. The format should be <namespace-name>/<pod-name>.
+}
```
draghuram, line 91: Would it be useful to store node name where backup was done?

design/pv_backup_info.md
blackpiglet: `Skipped` is inherited from this PR https://github.com/vmware-tanzu/velero/pull/6496.
It lets the downstream user know which volumes' data is not backed up for this backup.

We intend to not have a state field because most of the volume data backing up methods act in an async way and the VolumeInfo metadata only collects the information before the async operations begin, so the state is not accurate if it's added.

I agree that `Skipped` is not a backing-up method. Maybe adding a new boolean field called `Skipped`?

design/pv_backup_info.md
blackpiglet: It's not a typo, but we can discuss which is better.
IMO, the data mover of the snapshot is the filesystem uploader, which is the integrated Kopia library.
The data movement information of the snapshot is a more accurate description of this structure.

design/pv_backup_info.md
blackpiglet: Thanks. The comment on this field is wrong. I will modify it.

design/pv_backup_info.md
blackpiglet: It's not a typo, but we can discuss which is better.
IMO, the data mover of the snapshot is the filesystem uploader, which is the integrated Kopia library.
The data movement information of the snapshot is a more accurate description of this structure.

design/pv_backup_info.md
blackpiglet: Thanks. The comment on this field is wrong. I will modify it.

design/pv_backup_info.md
blackpiglet: The `SnapshotHandle` was not in the `VolumeInfo` structure.
The comment was not modified after the change. I will modify it.

design/pv_backup_info.md
blackpiglet: OK. I will add the node name to the structure.

design/pv_backup_info.md
blackpiglet: `SnapshotDataMoveInfo` is treated not as a way of backing up volumes, but it's treated as a following part of backing up. The snapshot ID is stored in the `CSISnapshotInfo` structure.

This is not the same as the current code, but this is the future refactoring direction.

design/pv_backup_info.md
draghuram: I see that you added "Skipped" field in #7070. Thanks.

design/pv_backup_info.md
draghuram: Got it, thanks. But it is quite possible that people may raise similar doubt in future so calling it "SnapshotDataMovementInfo" would make it clearer.

shubham-pampattiwar: +1 for `SnapshotDataMovementInfo`

design/pv_backup_info.md
```diff
@@ -0,0 +1,192 @@
...
+    PVName         string    // The PV name.
+    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup`, `CSISnapshot` and `Skipped`.
+    SnapshotDataMovement bool // Whether the volume's snapshot data is moved to specified storage.
+
+    SkippedReason   string    // The reason for the volume is skipped in the backup.
+    StartTimestamp  *metav1.Time // Snapshot starts timestamp.
+
+    CSISnapshotInfo          CSISnapshotInfo
+    SnapshotDataMoveInfo     SnapshotDataMoveInfo
+    NativeSnapshotInfo       VeleroNativeSnapshotInfo
```
shubham-pampattiwar, line 53: @blackpiglet Should this be `CloudProviderSnapshotInfo`  / `CloudStorageSnapshotInfo` / `ClusterStorageSnapshotInfo` instead ?

design/pv_backup_info.md
```diff
@@ -0,0 +1,192 @@
...
+    PVBInfo                  PodVolumeBackupInfo
```
shubham-pampattiwar, line 54: As we are adding `PodVolumeBackupInfo` , Shouldn't  the `DataUploadInfo` be added too ?

design/pv_backup_info.md
blackpiglet: Thanks. Modified.

design/pv_backup_info.md
blackpiglet: The `SnapshotDataMovementInfo` covers the snapshot data movement information. That structure covers the DataUpload information.

Issue #6240, opened 2023-05-08
Data Mover - preserve local snapshot when snapshot data movement is enabled


Lyndon-Li:
[From PM]: Since local snapshot restoration is faster than data movement, it is a useful feature to preserve several local snapshots when snapshot data movement is enabled. The count of the preservation should be configurable by users.

There are some details to cover in order to manage the local snapshots together with the remote data movement data, let's see how we can implement this.

sseago:
We should make sure that if we want to keep last X local snapshot that "0" is an allowed value for X. Some users would rather not keep any artifacts in-cluster for things that are fully restorable from backup, since this increases storage costs.

Lyndon-Li:
@sseago 
Sure, will do this.

For most of the on-premise storages, preserving local snapshots is either impossible or too costing; this requirement is primarily from the users who are running cluster in public cloud and they don't have requirements to move data across cloud providers or regions in the same cloud.

The major benefit to keep the local snapshots is that the restore is much faster from local snapshots.

Regarding to the cost for this kind of single-public-cloud-provider usage, the cost for local snapshot vs. data mover + simple object store is an interesting topic:

- The local snapshot data is stored in the object store by the cloud provider, the storing will be charged
- The management, manipulation of the local snapshot is charged additionally
- For data mover + simple object store: the storing of the backup data is charged and the data downloading is also charged (we may be able to optimize this)
- The comparison result of who stores more data in the object store for the same volume is unclear -- most cloud providers store the snapshot data incrementally from block level vs. Velero's repository (Kopia repository) has data deduplication
- Different cloud providers have different charging themes

reasonerjt:
> We should make sure that if we want to keep last X local snapshot that "0" is an allowed value for X. Some users would rather not keep any artifacts in-cluster for things that are fully restorable from backup, since this increases storage costs.

This is the current behavior of v1.12.0

How to expose the setting, i.e. is it per backup or per {PV, storageclass...} involves the decision on #6638 

But we may start a design without touching exposing the setting, for example PoC reading the number for local snapshot from environment variable.

PR #7002, opened 2023-10-23
Add design for data mover preserve local snapshot

Lyndon-Li:
Design for #6240. Add design for data mover preserve local snapshot


kaovilai (commented):
Can you add the source diagram file export (non png/jpg) to PR as well? making future edits if any easier.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -637,9 +645,40 @@ In DUCR/DDCR’s status, we have fields like ```totalBytes``` and ```doneBytes``
 - Call ```kubectl get dataupload -n velero xxx or kubectl get datadownload -n velero xxx```.
 - Call ```velero backup describe –details```. This is implemented as part of BIA/RIA V2, the above values are transferred to async operation and this command retrieve them from the async operation instead of DUCR/DDCR. See [general progress monitoring design][2] for details
 
+## Retain Native Snapshots
+Users are allowed to specify a global number as the number of native snapshots to be retained for each volume, then:
+- If the number is not specified, or if it is not a positive value, it means no native snapshot is to be retained
+- Otherwise, the value defines the max snapshots to be retained for each volume, a.k.a, the limit. If the limit is exceeded, the oldest retained snapshot will be removed
+
+The limit number is set as an annotation of a backup CR, the annotation is called ```backup.velero.io/data-mover-snapshot-to-retain```. In this way, CMPs are able to query the number from the backup CR.
```
kaovilai, line 653: Can you clarify what CMP stands for?

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -637,9 +645,40 @@ In DUCR/DDCR’s status, we have fields like ```totalBytes``` and ```doneBytes``
 - Call ```kubectl get dataupload -n velero xxx or kubectl get datadownload -n velero xxx```.
 - Call ```velero backup describe –details```. This is implemented as part of BIA/RIA V2, the above values are transferred to async operation and this command retrieve them from the async operation instead of DUCR/DDCR. See [general progress monitoring design][2] for details
 
+## Retain Native Snapshots
+Users are allowed to specify a global number as the number of native snapshots to be retained for each volume, then:
```
kaovilai, line 649:
```suggestion
Users are allowed to specify a global limit number as the number of native snapshots to be retained for each volume, then:
```

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
kaovilai: Is `backup.velero.io/data-mover-snapshot-to-retain` for querying purposes only and populated by velero? or can user set this value at backup CR creation to modify the behavior per backup?

Should we allow per pvc annotation settings?

Lyndon-Li: Corrected it, it should be `DMP`, thanks for the reminding.

Lyndon-Li: The annotation is for querying purposes only, we don't expect users to config it, they should config the server parameter instead.
Actually, users don't have opportunity to config it, as it is added when creating the backup CR and consumed immediately.

Lyndon-Li: We don't see any usability to configure the number of retained snapshot per PVC. So it is not supported.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: Done on adjusting the words.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -637,9 +645,40 @@ In DUCR/DDCR’s status, we have fields like ```totalBytes``` and ```doneBytes``
 - Call ```kubectl get dataupload -n velero xxx or kubectl get datadownload -n velero xxx```.
 - Call ```velero backup describe –details```. This is implemented as part of BIA/RIA V2, the above values are transferred to async operation and this command retrieve them from the async operation instead of DUCR/DDCR. See [general progress monitoring design][2] for details
 
+## Retain Native Snapshots
+Users are allowed to specify a global limit number of native snapshots to be retained for each volume, then:
+- If the number is not specified, or if it is not a positive value, it means no native snapshot is to be retained
+- Otherwise, the value defines the max snapshots to be retained for each volume, a.k.a, the limit. If the limit is exceeded, the oldest retained snapshot will be removed
+
+The limit number is set as an annotation of a backup CR, the annotation is called ```backup.velero.io/data-mover-snapshot-to-retain```. In this way, DMPs are able to query the number from the backup CR.  
+DMP is resiposible to maintain the native snapshots. Specifically, it is able to list/count the retained snapshots for each volume, and remove the old ones once the limit is exceeded.
```
kaovilai, line 654:
```suggestion
DMP is responsible to maintain the native snapshots. Specifically, it is able to list/count the retained snapshots for each volume, and remove the old ones once the limit is exceeded.  
```

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -637,9 +645,40 @@ In DUCR/DDCR’s status, we have fields like ```totalBytes``` and ```doneBytes``
...
+To assist on the listing/counting, several labels are applied to SAOs:
+- ```velero.io/snapshot-alias-name```: The DMP gives an alias to each SAO. As mentioned above, Velero/DMP should not expect DMs to keep the snapshots and their associated objects unchanged, e.g., a DM may delete the VS/VSC created by CSI plugin and create new ones from them (so the new ones also represent the same snapshots). By giving an alias, DMPs are guaranteed to find the SAOs by the alias label even though they have no idea to know where the DMs have cloned the SAOs. This also means that DMs should inherit all the labels from the original SAOs
+- ```velero.io/snapshot-volume-name```: This label stores the volume's identity for which a snapshot is created. For example, the value could be a PVC's namespaced name or UID. In this way, the DMP is able to tell all the retained snapshots for a specific volume
```
kaovilai, line 658: Do we want a separate label for UID?

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: Done, thanks.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: Here we don't want to limit the content of this label's value, we only define what it means and its usage --- the value could be anything as long as it could uniquely identify a volume.

I have update the same in the doc.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -890,11 +932,12 @@ Conclusively, below are the steps plugin DMs need to do in order to integrate to
 - Set PV's ```claimRef``` to the provided PVC and set ```velero.io/dynamic-pv-restore``` label
 
 ## Working Mode
-It doesn’t mean that once the data movement feature is enabled users must move every snapshot. We will support below two working modes:
+It doesn’t mean that once the data movement feature is enabled users must move every snapshot. We will support below three working modes:
 - Don’t move snapshots. This is same with the existing CSI snapshot feature, that is, native snapshots are taken and kept
 - Move snapshot data and delete native snapshots. This means that once the data movement completes, the native snapshots will be deleted.
+- Move snapshot data and keep X native snapshtos. This means snapshot data is moved first and also several native snapshots will be kept according to users' configuration.
```
kaovilai, line 938:
```suggestion
- Move snapshot data and keep X native snapshots for each volume. This means snapshot data is moved first and also several native snapshots will be kept according to users' velero server parameter.  
```

Lyndon-Li: Hmmm, let's keep the `users' configuration` part, the way for users to do the configuration may change, for example, we may add a installation parameter in future. So let's keep this general statement.

Lyndon-Li: Done on the part `snapshots for each volume`. Thanks.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -637,9 +645,40 @@ In DUCR/DDCR’s status, we have fields like ```totalBytes``` and ```doneBytes``
...
+- ```velero.io/snapshot-volume-name```: This label stores the volume's identity for which a snapshot is created. Its value could be anything that could uniquely identify a volume, for example, for a PVC, the value could be its namespaced name or UID. In this way, the DMP is able to tell all the retained snapshots for a specific volume  
+
+DMPs then return the SAOs as ```itemToUpdate``` with their aliases to Velero backup, in this way, when the DM execution finishes, Velero gets the latest SAOs by their aliases and persist them to the backup storage.  
+Velero, specifically the backup sync controller, is responsible to sync the SAOs to the target cluster for restore if they are not there. In this way, it is guaranteed that all the SAOs are available as long as their associated backups are there, or for every restore, DMPs always see the full list of SAOs for all the volumes to be restored.  
+ 
+After the retained snapshots are counted against the limit and if limit is exceeded, DMP needs to make sure the old snapshots are first removed before creating the new snapshot, otherwise, the snapshot creation may fail as some storage has a hard limit of the snapshot numbers.
```
yanji09, line 663: I understand that storage snapshot limit will cause snapshot creation to fail. While if the backup finally fails, deleting the old snapshot firstly may cause the user to lose a recovery point.

What will be the behavior if the configured global limit number > the storage supported snapshot numbers? does it make sense to limit the global number to a relative reasonable value (a number that most storages won't exceed)?

Lyndon-Li: > if the backup finally fails, deleting the old snapshot firstly may cause the user to lose a recovery point
We delete the native snapshots from an old completed backup. Since the backup is completed, the data mover has also completed. It means, a completed backup contains native snapshots + data mover backup data.
So here we delete the native snapshots for the backup only, it won't loose recovery point. The recovery point will refer to the data mover data.
This approach is actually to avoid the situation that the snapshot creation fails due to snapshot limit for the current backup, so it delete the native snapshot of the old backup first.

Lyndon-Li: > What will be the behavior if the configured global limit number > the storage supported snapshot numbers? 

The snapshot creation will fail if the number of snapshots exceeds the storage limit

Lyndon-Li: > does it make sense to limit the global number to a relative reasonable value (a number that most storages won't exceed)

I am afraid we cannot deduce a number that works for most storages. The snapshot limit is storage specific. Actually, for a storage, there is no technical point to support a fixed number as the limit, it is a compromise of resource consumption and performance, so generally this number is configurable on the storage side.

Lyndon-Li: Let me add some part to discuss the storage limit in the doc.

Lyndon-Li: Done on the addition.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -637,9 +647,42 @@ In DUCR/DDCR’s status, we have fields like ```totalBytes``` and ```doneBytes``
 - Call ```kubectl get dataupload -n velero xxx or kubectl get datadownload -n velero xxx```.
 - Call ```velero backup describe –details```. This is implemented as part of BIA/RIA V2, the above values are transferred to async operation and this command retrieve them from the async operation instead of DUCR/DDCR. See [general progress monitoring design][2] for details
 
+## Retain Native Snapshots
+Users are allowed to specify a global limit number of native snapshots to be retained for each volume, then:
+- If the number is not specified, or if it is not a positive value, it means no native snapshot is to be retained
+- Otherwise, the value defines the max snapshots to be retained for each volume, a.k.a, the limit. If the limit is exceeded, the oldest retained snapshot will be removed
```
anshulahuja98, line 653: can we add a nuance here of settings limits per Driver type. 
for example disk.csi.cloud.com might support retaining upto 100snapshots, while file.csi.cloud.com might only support 10.

anshulahuja98 (changes requested):

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
```diff
@@ -637,9 +647,42 @@ In DUCR/DDCR’s status, we have fields like ```totalBytes``` and ```doneBytes``
...
+To assist on the listing/counting, several labels are applied to SAOs:
+- ```velero.io/snapshot-alias-name```: The DMP gives an alias to each SAO. As mentioned above, Velero/DMP should not expect DMs to keep the snapshots and their associated objects unchanged, e.g., a DM may delete the VS/VSC created by CSI plugin and create new ones from them (so the new ones also represent the same snapshots). By giving an alias, DMPs are guaranteed to find the SAOs by the alias label even though they have no idea to know where the DMs have cloned the SAOs. This also means that DMs should inherit all the labels from the original SAOs
```
blackpiglet, line 659: Should the snapshot alias name be an SOA's name or a snapshot name?
Since it's inherited from the original one, what is the original SOA's snapshot alias name?

Lyndon-Li: `itemToUpdate` is returned from DMP, DMP doesn't exchange the SAO's name with DMs, while DMs will could create the SAO on its own, they DMP will not be able to know the exact name.

The alias could be anything that uniquely identify a SAO, e.g., a random UID.

design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
Lyndon-Li: We can add this as a CSI plugin specific configuration overwriting the global number, as the CSI Driver is not generic for all DMPs.

Lyndon-Li: I added below statement to the design:
`For the same reason, if a DMP supports snapshot from more than one kinds of storages, it may need to set different snapshot limits for volumes from different storages. In this case, the DMP could overwrite the global limit number based on its own configurations. The configurations are private to the DMP itself since the storage information is only known by the specific DMP, and how the configurations are structured and interpreted is also private to the specific DMP. `

This indicates that the DMP specific configuration is supported to overwrite the global number and the configuration is managed by the DMP itself.
So we don't need to give a generic configuration structure for all DMPs.
And for CSI plugin, I don't want to mention the concrete configuration structure for now, the requirement is still obscure right now, e.g., sometimes we want to differentiate it by CSI driver type, sometimes we want to do it by volumeType, etc. Let's wait and see the future requirements after this feature is release, and then we come back to design the structure for CSI.

@anshulahuja98 Let me know you are good with this.

anshulahuja98: This should be good enough for now.
Will keep this comment open for others to have a look. Otherwise consider this resolved for now.

Issue #6607, opened 2023-08-04
Configuration option for Kopia parallelism.


iusergii:
**Describe the solution you'd like**
I'd like to be able to set a number of parallel uploads for the Kopia backup repository. 
Slack [discussion](https://kubernetes.slack.com/archives/C6VCGP4MT/p1691018317118429)

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

reasonerjt:
If I understand correctly, the parallelism setting should be set to control the transfer session for a pv, this can benefit both PVB and Datamover flow, we just need to design how to expose it via CR and CLI.

Issue #6664, opened 2023-08-16
Allow "sparse" option for Kopia & Restic restore


draghuram:
**Describe the problem/challenge you have**
Restic, by default, restores sparse files as non-sparse, thus increasing the disk space used on the target PV compared to the source PV. In some cases, this leads to "No space left on the device" errors which are confusing to users because the source PV has no issues hosting the file. For an example, see #2298.

**Describe the solution you'd like**
Restic 0.15 supports `--sparse` option which results in a sparse file being restored as sparse file. Velero 1.11 does ship with Restic 0.15 but there is currently no way of passing this option to the command constructed by Velero. The solution is to come up a mechanism where users can pass additional arguments to the restic command. Or provide a "options" field in restore CR that accepts Restic (and Kopia) specific options. Perhaps, the scheme can be generalized so that both backup and restore can accept additional options.

**Environment:**

- Velero version (use `velero version`): 1.11.0

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

Lyndon-Li:
Kopia also supports this option, let's make sure Kopia path is covered as it is used by both fs-backup and data mover.

reasonerjt:
Given we have made the decision to move away from Restic, I don't think we want to continue making enhancement for Restic.

draghuram:
I used to think the same way too but we have a user who cannot move from Restic at this point. So unless it imposes a huge burden, we should make an attempt to support Restic as well.

sseago:
@draghuram So there are different issues with restic:
1) what level of support does it get
2) how long will it be around

For 1, as long as it's considered supported, at a minimum, it needs bugfixes. As for new features, in the past we've taken the approach that we won't add them, mostly due to resource constraints/priorities. If a community member provides PRs to implement features we hadn't intended to include in restic, we'd probably need to have a discussion around it.

As for 2, I believe the intent is to deprecate restic in the near future, meaning that at some point afterwards it is expected to disappear entirely (still need to get that deprecation policy PR approved/merged, though).

Lyndon-Li:
[...]

PR #7005, opened 2023-10-24
Design for Velero uploader configuration integration and extensibility

qiuming-best:

Fixes #(issue)
https://github.com/vmware-tanzu/velero/issues/6607

### DIFF
--- a/design/velero-uploader-configuration.md
+++ b/design/velero-uploader-configuration.md
@@ -1,0 +1,126 @@
+# Velero Uploader Configuration Integration and Extensibility
+
+## Abstract
+This design proposal aims to make Velero Uploader configurable by introducing a structured approach for managing Uploader settings. we will define and standardize a data structure to facilitate future additions to Uploader configurations. This enhancement provides a template for extending Uploader-related options. And also includes examples of adding sub-options to the Uploader Configuration.
+
+## Background
+Velero is widely used for backing up and restoring Kubernetes clusters. In various scenarios, optimizing the backup process is essential, future needs may arise for adding more configuration options related to the Uploader component especially when dealing with large datasets. Therefore, a standardized configuration template is required.
+
+## Goals
+1. **Extensible Uploader Configuration**: Provide an extensible approach to manage Uploader configurations, making it easy to add and modify configuration options related to the Velero uploader.
+2. **User-friendliness**: Ensure that the new Uploader configuration options are easy to understand and use for Velero users without introducing excessive complexity.
+
+## Non Goals
+1. Expanding to other Velero components: The primary focus of this design is Uploader configuration and does not include extending to other components or modules within Velero. Configuration changes for other components may require separate design and implementation.
+
+## High-Level Design
+To achieve extensibility in Velero Uploader configurations, the following key components and changes are proposed:
+
+### UploaderConfig Structure
+A new data structure, `UploaderConfig`, will be defined to store Uploader configurations. This structure will include the configuration options related to backup and restore for Uploader:
+
+```go
+type UploaderConfig struct {
+    // sub-options
+}
+```
+
+### Integration with Backup CRD
+The Velero CLI will support an uploader configuration-related flag, allowing users to set the value when creating backups or restores. This value will be stored in the `UploaderConfig` field within the `Backup` CRD and `Restore` CRD:
+
+```go
+type BackupSpec struct {
+    UploaderConfig shared.UploaderConfig `json:"uploaderConfig,omitempty"`
+}
+
+type RestoreSpec struct {
+    UploaderConfig shared.UploaderConfig `json:"uploaderConfig,omitempty"`
+}
+```
+
+### Configuration Propagated to Different CRDs
+The configuration specified in `UploaderConfig` needs to be effective for backup and restore both by file system way and data-mover way. 
+Therefore, the `UploaderConfig` field value from the `Backup` CRD should be propagated to `PodVolumeBackup` and `DataUpload` CRDs:
+
+```go
+type PodVolumeBackupSpec struct {
+    ...
+    UploaderConfig shared.UploaderConfig `json:"uploaderConfig,omitempty"`
+}
+
+type DataUploadSpec struct {
+    ...
+    UploaderConfig shared.UploaderConfig `json:"uploaderConfig,omitempty"`
+}
+```
+
+Also the `UploaderConfig` field value from the `Restore` CRD should be propagated to `PodVolumeRestore` and `DataDownload` CRDs:
+
+```go
+type PodVolumeRestoreSpec struct {
+    ...
+    UploaderConfig shared.UploaderConfig `json:"uploaderConfig,omitempty"`
+}
+
+type DataDownloadSpec struct {
+    ...
+    UploaderConfig shared.UploaderConfig `json:"uploaderConfig,omitempty"`
+}
+```
+
+## Sub-options in UploaderConfig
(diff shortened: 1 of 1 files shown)


design/velero-uploader-configuration.md
```diff
@@ -0,0 +1,93 @@
...
+
+```go
+curPolicy := getDefaultPolicy()
+if uploaderCfg.ParallelFilesUpload > 0 {
+    curPolicy.UploadPolicy.MaxParallelFileReads = newOptionalInt(uploaderCfg.ParallelFilesUpload)
+}
+```
+
+The process is as follows: Users pass the ParallelFilesUpload parameter and its value through the Velero CLI. This parameter and its value are stored as a sub-option within UploaderConfig and then placed into the Backup CR. When users perform file-system level backups, UploaderConfig is passed to the PodVolumeBackup CR. When users use the Data-mover for backups, it is passed to the DataUpload CR. Each respective controller within the CRs calls the uploader, and the ParallelFilesUpload from UploaderConfig in CRs is passed to the uploader. When the uploader subsequently calls the Kopia API, it can use the ParallelFilesUpload to set the MaxParallelFileReads parameter.
+
```
Lyndon-Li, line 91: Let's mention the situation for Restic here --- this option is not supported by Restic uploader.

design/velero-uploader-configuration.md
```diff
@@ -0,0 +1,93 @@
...
+Velero can set the `MaxParallelFileReads` parameter for Kopia's upload policy as follows:
+
+```go
+curPolicy := getDefaultPolicy()
+if uploaderCfg.ParallelFilesUpload > 0 {
+    curPolicy.UploadPolicy.MaxParallelFileReads = newOptionalInt(uploaderCfg.ParallelFilesUpload)
+}
+```
+
+The process is as follows: Users pass the ParallelFilesUpload parameter and its value through the Velero CLI. This parameter and its value are stored as a sub-option within UploaderConfig and then placed into the Backup CR. When users perform file-system level backups, UploaderConfig is passed to the PodVolumeBackup CR. When users use the Data-mover for backups, it is passed to the DataUpload CR. Each respective controller within the CRs calls the uploader, and the ParallelFilesUpload from UploaderConfig in CRs is passed to the uploader. When the uploader subsequently calls the Kopia API, it can use the ParallelFilesUpload to set the MaxParallelFileReads parameter.
```
Lyndon-Li, line 90: file-system level backups -> file system backups.

design/velero-uploader-configuration.md
```diff
@@ -0,0 +1,93 @@
...
+The Velero CLI will support an uploader configuration-related flag, allowing users to set the value when creating backups. This value will be stored in the `UploaderConfig` field within the backup CRD:
+
+```go
+type BackupSpec struct {
+    UploaderConfig shared.UploaderConfig `json:"uploaderConfig,omitempty"`
+}
+```
+
+### Configuration Propagated to Different CRDs
+The configuration specified in `UploaderConfig` needs to be effective for both file-system level backups and data-mover backups. Therefore, the `UploaderConfig` field value from the backup CRD should be propagated to `PodVolumeBackup` and `DataUpload` CRDs:
```
Lyndon-Li, line 38: file-system level backups -> file system backups. file system backup is a particular term of Velero, let's make it consistent.

Lyndon-Li, line 38: Let's add Restore the UploaderConfig to Restore CRD as well and propagate it to PodVolumeRestore and DataDownload CRDs, considering an option may be for restores

design/velero-uploader-configuration.md
qiuming-best: modified

design/velero-uploader-configuration.md
qiuming-best: modified

design/velero-uploader-configuration.md
qiuming-best: modified

qiuming-best: file-system

Lyndon-Li (approved):

design/velero-uploader-configuration.md
```diff
@@ -0,0 +1,126 @@
...
+below are the key steps that should be added to support this new feature.
+
+#### Velero CLI 
+The Velero CLI will support a `--parallel-files-upload` flag, allowing users to set the `ParallelFilesUpload` value when creating backups.
+
+#### UploaderConfig 
+below the sub-option `ParallelFilesUpload` is added into UploaderConfig:
+
+```go
+type UploaderConfig struct {
```
reasonerjt, line 85: Is it safe to assume most attributes to be added to this configuration, applies to both backup and restore workflow?

i.e. there won't be an attribute `xyz` that only makes sense in backup but not restore?

design/velero-uploader-configuration.md
```diff
@@ -0,0 +1,126 @@
...
+    ParallelFilesUpload int `json:"parallelFilesUpload,omitempty"`
```
reasonerjt, line 86: If it is left unset the value will be `0`, does `0` has the same meaning as `1`?

design/velero-uploader-configuration.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Sub-options in UploaderConfig
+Adding fields above in CRDs can accommodate any future additions to Uploader configurations by adding new fields to the `UploaderConfig` structure.
+
+### Parallel Files Upload
+This section focuses on enabling the configuration for the number of parallel file uploads during backups.
+below are the key steps that should be added to support this new feature.
+
+#### Velero CLI 
+The Velero CLI will support a `--parallel-files-upload` flag, allowing users to set the `ParallelFilesUpload` value when creating backups.
```
reasonerjt, line 79: is it only for backup?

We should be careful with the doc to explain this parm only impacts the flow when the files are handled by `uploader` like `kopia` and `restic`

design/velero-uploader-configuration.md
Lyndon-Li: I think here the plan is to have the backup options and restore options mingling in this UploaderConfig struct, so that we don't need to introduce separate struct for backups and restores.
Actually, quite lots of options are for either backups or restores, not for the both.

qiuming-best: it does not matter even if we have one attribute 'xyz' both make sense in backup and restore.
As the command created by Velero CLI could not take effect both in backup and restore at the same time. The command may backup or restore, the UploaderConfig will store in `Backup` or `Restore` accordingly

design/velero-uploader-configuration.md
qiuming-best: yes, it's only for backup. and in  `--parallel-files-upload` flag usage clearly states that it can only be used for backups using  `kopia` type of uploader

draghuram:
This design looks good to me. I am guessing that we can add a boolean field called "sparseRestore" to implement #6664.  If you agree, how do you want to proceed with the work? Do you want to update this proposal itself with the new field or perhaps check this in first and then write a different proposal?

qiuming-best:
@draghuram  > This design looks good to me. I am guessing that we can add a boolean field called "sparseRestore" to implement #6664. If you agree, how do you want to proceed with the work? Do you want to update this proposal itself with the new field or perhaps check this in first and then write a different proposal?

`sparseRestore` is good. 

once this proposal is merged, I think you could add a new sub-section in this design about the configuration `sparseRestore` which would be similar to the sub-section `Parallel Files Upload` in `Sub-options in UploaderConfig`

blackpiglet (approved):

Issue #7027, opened 2023-10-27
DataUpload path seems to be incorrect


ScOut3R:
**What steps did you take and what happened:**
I'm executing a CSI snapshot backup coupled with a DataMover step. The process succeeds end to end, however the resulting backup does not contain the data from the PVC, but an AWS IAM token.

The `status.path` has the following:
```
/host_pods/6a774d15-b3f1-419a-89b5-07c1f08c8ab6/volumes/kubernetes.io~projected/aws-iam-token
```

This is consistent across different backups.

**What did you expect to happen:**

I expected that the backup will contain the actual data from the source PVC.

**The following information will help us better understand what's going on**:

[bundle-2023-10-27-14-21-51.tar.gz](https://github.com/vmware-tanzu/velero/files/13184946/bundle-2023-10-27-14-21-51.tar.gz)

**Anything else you would like to add:**
The backup storage location is an S3 bucket and using IRSA authentication/authorization.

**Environment:**

- Velero version (use `velero version`): `v1.12.0`
- Velero features (use `velero client config get features`): `<NOT SET>`
- Kubernetes version (use `kubectl version`): `1.26.9`

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

qiuming-best:
this is because you backup one volume named `aws-iam-token`

<img width="416" alt="image" src="https://github.com/vmware-tanzu/velero/assets/8985337/d2446956-1b45-4bec-8f43-4c7ac174a732">

ScOut3R:
> this is because you backup one volume named `aws-iam-token`
> <img alt="image" width="416" src="https://user-images.githubusercontent.com/8985337/278585127-d2446956-1b45-4bec-8f43-4c7ac174a732.png">

The `aws-iam-token` volume is attached to the `velero` and related pods, not to the pods that I am backup up with their `PVC`s.

According to the `Backup` resource there are two volumes marked for backup:

```Backup Item Operations:
  Operation for persistentvolumeclaims k6/grafana:
    Backup Item Action Plugin:  velero.io/csi-pvc-backupper
[...]

PR #7038, opened 2023-10-31
Issue 7027: backup exposer -- don't assume first volume as the backup volume

Lyndon-Li:
Fix issue #7027, data mover backup exposer should not assume the first volume as the backup volume in backup pod

### DIFF
--- a/pkg/exposer/csi_snapshot.go
+++ b/pkg/exposer/csi_snapshot.go
@@ -204,6 +204,7 @@ func (e *csiSnapshotExposer) GetExposed(ctx context.Context, ownerObject corev1.
 
 	backupPodName := ownerObject.Name
 	backupPVCName := ownerObject.Name
+	volumeName := string(ownerObject.UID)
 
 	curLog := e.log.WithFields(logrus.Fields{
 		"owner": ownerObject.Name,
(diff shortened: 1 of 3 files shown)


pkg/exposer/csi_snapshot.go
```diff
@@ -232,7 +233,20 @@ func (e *csiSnapshotExposer) GetExposed(ctx context.Context, ownerObject corev1.
...
+		if pod.Spec.Volumes[i].Name == volumeName {
+			break
+		}
+	}
+
+	if i == len(pod.Spec.Volumes) {
+		return nil, errors.Errorf("backup pod %s doesn't have the expected backup volume", pod.Name)
+	}
+
+	curLog.WithField("pod", pod.Name).Infof("Backup volume is found in pod at index %v", i)
```
shubham-pampattiwar, line 247: Should this be debug log ?

ywk253100 (approved):

pkg/exposer/csi_snapshot.go
```diff
@@ -204,6 +204,7 @@ func (e *csiSnapshotExposer) GetExposed(ctx context.Context, ownerObject corev1.
 
 	backupPodName := ownerObject.Name
 	backupPVCName := ownerObject.Name
+	volumeName := string(ownerObject.UID)
```
qiuming-best, line 207: Why the `volumeName` is equal to `ownerObject.UID`?

pkg/exposer/csi_snapshot.go
Lyndon-Li: It is fine to be an info log, as same as other logs in the same function, because expose is not a frequently called operation and on the other hand, it is hard to troubleshoot, so we would rather add some info logs, so that they could give helps whenever we need to debug the expose process.

pkg/exposer/csi_snapshot.go
Lyndon-Li: Because we name the volume as `ownerObject.UID` when creating the backup pod, so it is always true that the backup volume is with the name of `ownerObject.UID` in the backup pod.
See the `createBackupPod` function for how the backup pod is created.

qiuming-best (approved):

PR #7059, opened 2023-11-06
Issue 6663: changes for configurable data path concurrency

Lyndon-Li:
Add the implementation for design #6950, configurable data path concurrency


pkg/nodeagent/node_agent.go
```diff
@@ -89,3 +113,31 @@ func GetPodSpec(ctx context.Context, kubeClient kubernetes.Interface, namespace
...
+		} else {
+			return nil, errors.Wrapf(err, "error to get node agent configs %s", configName)
+		}
+	}
+
+	if cm.Data == nil {
+		return nil, errors.Errorf("data is not available in config map %s", configName)
+	}
+
+	jsonString, exist := cm.Data[dataPathConConfigName]
```
ywk253100, line 131: So the current configmap data is as follows:
```
data-path-concurrency: {"k":"v"}
```

When we add new configs to the node agent, the configmap becomes:
```
data-path-concurrency: {"k":"v"}
new-config: {"k":"v"}
```

How about putting all configs into a JSON like this?
```
config : {"data-path-concurrency":{"k":"v"}, "new-config": {"k":"v"}}
```
In this way, we don't need to change the code in this function when we adding new configs and it is also easier for users to create the configmap with the command `kubectl create cm node-agent-configs -n velero --from-file=config=<json file name>`.

pkg/nodeagent/node_agent.go
```diff
@@ -89,3 +113,31 @@ func GetPodSpec(ctx context.Context, kubeClient kubernetes.Interface, namespace
 
 	return &ds.Spec.Template.Spec, nil
 }
+
+func GetConfigs(ctx context.Context, namespace string, kubeClient kubernetes.Interface) (*Configs, error) {
```
ywk253100, line 117: We should also update the website to document how to use this configmap

Lyndon-Li: Yes, I am planning to create a node-agent configuration document later.

pkg/cmd/cli/nodeagent/server.go
```diff
@@ -478,3 +488,65 @@ func (s *nodeAgentServer) markInProgressPVRsFailed(client ctrlclient.Client) {
...
+	}
+
+	if configs == nil || configs.DataPathConcurrency == nil {
+		s.logger.Infof("Node agent configs are not found, use the default number %v", defaultNum)
+		return defaultNum
+	}
+
+	globalNum := configs.DataPathConcurrency.GlobalConfig
+
+	if globalNum <= 0 {
```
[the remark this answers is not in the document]
Lyndon-Li: Here it is preparing the globalNum only. The number is not being used here --- it will be used later in any case that the perNode number is not available/valid.
So it is fine to have this preparation first and it is convenient for coding since we can restore the prepared globalNum any time as a failback.

Moreover, there indicates one point --- users must set a globalNum first even if they set a perNode number for each node, otherwise, there will be a warning `Global number %v is invalid, use the default value %v`.
I don't think this matters since it is very seldom/inconvenient that setting a perNode number for each node and leave the globalNum empty.

pkg/nodeagent/node_agent.go
Lyndon-Li: Good idea, I have done the changes.

ywk253100 (dismissed):

ywk253100 (approved):

qiuming-best (approved):

PR #7069, opened 2023-11-07
improve discoveryHelper.Refresh() in restore

27149chen:

`discoveryHelper.Refresh()` was called after every crd restore, this logic has two issues:
1. it was called no matter the crd is restored, updated or skipped. In fact, refresh is required only when it is restored
2. it was called after every crd restore. In fact, we just need to trigger a refresh after all crds are restored

In my env, a refresh takes 9s, and I have many crds, which causing the restore extremely slow

This pr improve it by calling refresh after all crds are restored and at least one crd are created during restore

Fixes #(issue)

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -502,7 +502,7 @@ func (ctx *restoreContext) execute() (results.Result, results.Result) {
 	}()
 
 	// totalItems: previously discovered items, i: iteration counter.
-	totalItems, processedItems, existingNamespaces := 0, 0, sets.NewString()
+	totalItems, processedItems, createdItems, existingNamespaces := 0, 0, 0, sets.NewString()
 
 	// First restore CRDs. This is needed so that they are available in the cluster
 	// when getOrderedResourceCollection is called again on the whole backup and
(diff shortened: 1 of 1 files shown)


pkg/restore/restore.go
```diff
@@ -525,16 +525,26 @@ func (ctx *restoreContext) execute() (results.Result, results.Result) {
...
 			nil,
 		)
 		warnings.Merge(&w)
 		errs.Merge(&e)
 	}
+	// If we just restored custom resource definitions (CRDs), refresh
+	// discovery because the restored CRDs may have created new APIs that
+	// didn't previously exist in the cluster, and we want to be able to
+	// resolve & restore instances of them in subsequent loop iterations.
+	if createdItems > 0 {
```
ywk253100, line 543: How about refreshing the discovery helper no matter whether new items are created or not? This is much easier.

27149chen: yes, it is easier, but if no crds are created, refresh is meaningless, and it takes time (~9s in my env)

pkg/restore/restore.go
```diff
@@ -502,7 +502,7 @@ func (ctx *restoreContext) execute() (results.Result, results.Result) {
 	}()
 
 	// totalItems: previously discovered items, i: iteration counter.
-	totalItems, processedItems, existingNamespaces := 0, 0, sets.NewString()
+	totalItems, processedItems, createdItems, existingNamespaces := 0, 0, 0, sets.NewString()
```
ywk253100, line 505: Please double-check whether this covers CRD updating case

ywk253100: Let's put the returning items into a structure

ywk253100 (approved):

PR #7070, opened 2023-11-07
Add VolumeInfo metadata structures.

blackpiglet:
Modify design according to comments.
Add PVInfo structure.

Add the VolumeInfo metadata structures for issues https://github.com/vmware-tanzu/velero/labels/pv-backup-info

Fixes #(issue)

### DIFF
--- a/design/pv_backup_info.md
+++ b/design/pv_backup_info.md
@@ -35,44 +35,50 @@ The `restoreItem` function can decode the _backup-name_-volumes-info.json file t
 ## Detailed Design
 
 ### The VolumeInfo structure
-_backup-name_-volumes-info.json file is a structure that contains an array of structure `VolumeInfo` and a VolumeInfo version parameter. The version is used to support multiple VolumeInfo version, and make the backward-compatible support possible. The current version is `1`.
+_backup-name_-volumes-info.json file is a structure that contains an array of structure `VolumeInfo`.
 
-The 1 version of `VolumeInfo` definition is:
 ``` golang
-type VolumeInfoV1 struct {
-    PVCName        string    // The PVC's name. The format should be <namespace-name>/<PVC-name>
+type VolumeInfo struct {
+    PVCName        string    // The PVC's name.
+    PVCNamespace   string    // The PVC's namespace.
     PVName         string    // The PV name.
-    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup`, `CSISnapshot` and `Skipped`.
-    SnapshotDataMovement bool // Whether the volume's snapshot data is moved to specified storage.
+    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup` and `CSISnapshot`.
+    SnapshotDataMoved bool   // Whether the volume's snapshot data is moved to specified storage.
 
+    Skipped         boolean   // Whether the Volume is skipped in this backup.
     SkippedReason   string    // The reason for the volume is skipped in the backup.
     StartTimestamp  *metav1.Time // Snapshot starts timestamp.
 
+    OperationID     string   // The Async Operation's ID.
+
     CSISnapshotInfo          CSISnapshotInfo
-    SnapshotDataMoveInfo     SnapshotDataMoveInfo
+    SnapshotDataMovementInfo SnapshotDataMovementInfo
     NativeSnapshotInfo       VeleroNativeSnapshotInfo
     PVBInfo                  PodVolumeBackupInfo
+    PVInfo                   PVInfo
 }
 
 // CSISnapshotInfo is used for displaying the CSI snapshot status
 type CSISnapshotInfo struct {
-    SnapshotHandle  string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
+    SnapshotHandle  string       // It's the storage provider's snapshot ID for CSI.
     Size            int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
 
     Driver          string  // The name of the CSI driver.
     VSCName         string // The name of the VolumeSnapshotContent. 
 }
 
-// SnapshotDataMoveInfo is used for displaying the snapshot data mover status.
-type SnapshotDataMoveInfo struct {
+// SnapshotDataMovementInfo is used for displaying the snapshot data mover status.
+type SnapshotDataMovementInfo struct {
     DataMover        string    // The data mover used by the backup. The valid values are `velero` and ``(equals to `velero`).
     UploaderType     string    // The type of the uploader that uploads the snapshot data. The valid values are `kopia` and `restic`. It's useful for file-system backup and snapshot data mover.
-    RetainedSnapshot string    // The name or ID of the snapshot associated object(SAO).
+    RetainedSnapshot string    // The name or ID of the snapshot associated object(SAO). SAO is used to support local snapshots for the snapshot data mover, e.g. it could be a VolumeSnapshot for CSI snapshot data moign/pv_backup_info.
+    SnapshotHandle string  	   // It's the filesystem repository's snapshot ID.
+	
 }
 
 // VeleroNativeSnapshotInfo is used for displaying the Velero native snapshot status.
 type VeleroNativeSnapshotInfo struct {
-    SnapshotHandle      string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
+    SnapshotHandle      string       // It's the storage provider's snapshot ID for the Velero-native snapshot.
     Size                int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
 
     VolumeType string    // The cloud provider snapshot volume type.
@@ -82,33 +88,20 @@ type VeleroNativeSnapshotInfo struct {
 
 // PodVolumeBackupInfo is used for displaying the PodVolumeBackup snapshot status.
 type PodVolumeBackupInfo struct {
-    SnapshotHandle      string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
+    SnapshotHandle      string       // It's the file-system uploader's snapshot ID for PodVolumeBackup.
     Size                int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
 
     UploaderType  string    // The type of the uploader that uploads the data. The valid values are `kopia` and `restic`. It's useful for file-system backup and snapshot data mover.
-    VolumeName    string // The PVC's corresponding volume name used by Pod
-    PodName       string // The Pod name mounting this PVC. The format should be <namespace-name>/<pod-name>.
-}
-```
-
-To make Velero support multiple versions of VolumeInfo, Velero needs to create a structure to include the version and the versioned volume information together. The information is persisted in the metadata file during backup creation. When reading the VolumeInfo metadata file, Velero reads the version information first by un-marshalling the metadata file by structure `VolumeInfoVersion`, then decide to use which version of the VolumeInfos according to the version value.
-
-If there is need to introduce a non compatible change to the VolumeInfo, a new version of `VolumeInfos` and `VolumeInfo` are needed. For example, version 2 is created, then `VolumeInfosV2` and `VolumeInfoV2` structures are needed.
-
-Only when there a non-backward-compatible change introduced in the `VolumeInfo` structure, the `version`'s version will be incremented.
-
-``` golang
-type VolumeInfoVersion struct {
-    Version string 
+    VolumeName    string   // The PVC's corresponding volume name used by Pod: https://github.com/kubernetes/kubernetes/blob/e4b74dd12fa8cb63c174091d5536a10b8ec19d34/pkg/apis/core/types.go#L48
+    PodName       string   // The Pod name mounting this PVC. The format should be <namespace-name>/<pod-name>.
+    NodeName      string   // The PVB-taken k8s node's name.
 }
 
-type VolumeInfosV2 struct {
-    Infos   []VolumeInfoV2
-    Version string // VolumeInfo structure's version information.
-}
-
-type VolumeInfoV2 struct {
-    ...
+// PVInfo is used to store some PV information modified after creation.
+// Those information are lost after PV recreation.
+type PVInfo struct {
+    ReclaimPolicy string            // ReclaimPolicy of PV. It could be different from the referenced StorageClass.
+    Labels        map[string]string // The PV's labels should be kept after recreation.
 }
 ```
 
@@ -164,11 +157,12 @@ After introducing the VolumeInfo array, the following logic will be added.
             ...
         case CSISnapshot:
             ...
-        case Skipped:
-            // Check whether the Velero server should restore the PV depending on the DeletionPolicy setting.
         default:
             // Need to check whether the volume is backed up by the SnapshotDataMover.
             if volumeInfo.SnapshotDataMovement:
+
+            // Check whether the Velero server should restore the PV depending on the DeletionPolicy setting.
+            if volumeInfo.Skipped:
 ```
 
 ### How the VolumeInfo metadata file is deleted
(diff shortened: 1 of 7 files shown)


blackpiglet:
@draghuram 
Please take a look at the PR. I modified the design document according to comments in #6962.

design/pv_backup_info.md
```diff
@@ -42,21 +42,23 @@ The 1 version of `VolumeInfo` definition is:
...
     SnapshotDataMoveInfo     SnapshotDataMoveInfo
     NativeSnapshotInfo       VeleroNativeSnapshotInfo
     PVBInfo                  PodVolumeBackupInfo
+    PVInfo                   PVInfo
 }
 
 // CSISnapshotInfo is used for displaying the CSI snapshot status
 type CSISnapshotInfo struct {
-    SnapshotHandle  string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
+    SnapshotHandle  string       // The actual snapshot ID. It's the storage provider's snapshot ID for CSI.
```
draghuram, line 61: It is a minor detail but the comment can simply be "Storage provider's snapshot ID for CSI" or similar. No need to say "actual snapshot ID". Same applies to other places as well.

design/pv_backup_info.md
```diff
@@ -82,12 +84,20 @@ type VeleroNativeSnapshotInfo struct {
...
 // PodVolumeBackupInfo is used for displaying the PodVolumeBackup snapshot status.
 type PodVolumeBackupInfo struct {
-    SnapshotHandle      string       // The actual snapshot ID. It can be the cloud provider's snapshot ID or the file-system uploader's snapshot.
+    SnapshotHandle      string       // The actual snapshot ID. It's the file-system uploader's snapshot ID for PodVolumeBackup.
     Size                int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
 
     UploaderType  string    // The type of the uploader that uploads the data. The valid values are `kopia` and `restic`. It's useful for file-system backup and snapshot data mover.
-    VolumeName    string // The PVC's corresponding volume name used by Pod
-    PodName       string // The Pod name mounting this PVC. The format should be <namespace-name>/<pod-name>.
+    VolumeName    string   // The PVC's corresponding volume name used by Pod
```
draghuram, line 91: Is it PVC name? Or the mount path of the PVC in the Pod, or perhaps PV name? It is not clear just from the comment. Can you please clarify?

pkg/volume/volume_info_common.go
```diff
@@ -0,0 +1,80 @@
...
+
+// SnapshotDataMoveInfo is used for displaying the snapshot data mover status.
+type SnapshotDataMoveInfo struct {
+	// The data mover used by the backup. The valid values are `velero` and ``(equals to `velero`).
+	DataMover string `json:"dataMover"`
+
+	// The type of the uploader that uploads the snapshot data. The valid values are `kopia` and `restic`. It's useful for file-system backup and snapshot data mover.
+	UploaderType string `json:"uploaderType"`
+
+	// The name or ID of the snapshot associated object(SAO).
```
draghuram, line 39: Just curious. What exactly is SAO? If you can give an example, that would be great.

pkg/volume/volume_info_v1.go
```diff
@@ -0,0 +1,38 @@
...
+	PVCName string `json:"pvcName"`
+
+	// The PV name.
+	PVName string `json:"pvName"`
+
+	// The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup` and `CSISnapshot`.
+	BackupMethod string `json:"backupMethod,omitempty"`
+
+	// Whether the volume's snapshot data is moved to specified storage.
+	SnapshotDataMovement bool `json:"snapshotDataMovement"`
```
draghuram, line 23: Considering that the type is boolean, wouldn't it be better to call the field "SnapshotDataMoved"?

draghuram (commented):
Design changes look good to me though I have few minor comments.

pkg/volume/volume_info_common.go
blackpiglet: That's my bad.
SAO is a concept taken from an in-progress PR to support local snapshots for the snapshot data mover.
https://github.com/vmware-tanzu/velero/pull/7002/files#diff-e6456a41611c6ae9aa199c7cba18d13244ae635abaf7c63f706ace5ab956812eR661

design/pv_backup_info.md
blackpiglet: Thanks. Modified.

pkg/volume/volume_info_v1.go
blackpiglet: Thanks. Modified.

design/pv_backup_info.md
blackpiglet: No, it's the volume name in the PodSpec. It's used for displaying PVB information of backup.
https://github.com/kubernetes/kubernetes/blob/e4b74dd12fa8cb63c174091d5536a10b8ec19d34/pkg/apis/core/types.go#L48

design/pv_backup_info.md
```diff
@@ -42,21 +42,23 @@ The 1 version of `VolumeInfo` definition is:
 type VolumeInfoV1 struct {
     PVCName        string    // The PVC's name. The format should be <namespace-name>/<PVC-name>
     PVName         string    // The PV name.
-    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup`, `CSISnapshot` and `Skipped`.
-    SnapshotDataMovement bool // Whether the volume's snapshot data is moved to specified storage.
+    BackupMethod   string    // The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup` and `CSISnapshot`.
```
anshulahuja98, line 45: just wondering if in future we might support multiple forms of backup for 1 PV.

design/pv_backup_info.md
```diff
@@ -42,21 +42,23 @@ The 1 version of `VolumeInfo` definition is:
...
+    SnapshotDataMoved bool   // Whether the volume's snapshot data is moved to specified storage.
 
+    Skipped         boolean   // Whether the Volume is skipped in this backup.
     SkippedReason   string    // The reason for the volume is skipped in the backup.
```
anshulahuja98, line 49: what about error scenarios?
where we tried to take snapshot through 1 of the routes but it failed 
Will we populate it in skipreason?
If not should we maybe have another field for errors?

anshulahuja98 (changes requested):

design/pv_backup_info.md
draghuram: My understanding is that "skippedReason" will be populated with the error message.

pkg/volume/volume_info_common.go
draghuram: Thanks, will check that PR.

allenxu404 (approved):

design/pv_backup_info.md
blackpiglet: There are some tries in that direction, but not in exactly backing a volume with multiple way's directions, for example, the snapshot data mover, which backs up the volume by snapshot, then backs up by the filesystem uploader.

IMO, maybe there will not be a universal way to let volume back up data by arbitrary ways that Velero can provide because this can be done by separate backups, but if there is a valid using scenario, we can try to implement it with proper ways.

design/pv_backup_info.md
blackpiglet: The error scenario could be recorded by a `State` field, but there is no such field for this structure., and it's on purpose.

The SkipReason is not used for recording the volume backing-up error. It's used for filtered-out volumes. 

The consideration is the VolumeInfo metadata is used to record the volumes' information in the Velero process, and the async operations' changes are not included. Most of the Volume data backing up processes are time-consuming.
The reason that the async operations changes are not included is that if they are included, the Velero server needs to update the new metadata file during the backup finalizing state. It will make the process more complicated, and it will be hard to meet the metadata's data immutability requirement: https://velero.io/docs/v1.12/backup-reference/#limitation.

Lyndon-Li (dismissed):

pkg/volume/volume_info_common.go
```diff
@@ -0,0 +1,86 @@
+package volume
```
Lyndon-Li, line 1: Missing copyright

pkg/volume/volume_info_v1.go
```diff
@@ -0,0 +1,38 @@
+package volume
```
Lyndon-Li, line 1: Missing copyright

blackpiglet: Thanks. Modified.

pkg/volume/volume_info_common.go
blackpiglet: Thanks. Modified.

pkg/volume/volume_info_common.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// The name of the CSI driver.
+	Driver string `json:"driver"`
+
+	// The name of the VolumeSnapshotContent.
+	VSCName string `json:"vscName"`
+}
+
+// SnapshotDataMovementInfo is used for displaying the snapshot data mover status.
+type SnapshotDataMovementInfo struct {
```
Lyndon-Li, line 48: A repo snapshot ID is also required for data mover

design/pv_backup_info.md
```diff
@@ -42,37 +42,39 @@ The 1 version of `VolumeInfo` definition is:
...
     Size            int64        // The snapshot corresponding volume size. Some of the volume backup methods cannot retrieve the data by current design, for example, the Velero native snapshot.
 
     Driver          string  // The name of the CSI driver.
     VSCName         string // The name of the VolumeSnapshotContent. 
 }
 
-// SnapshotDataMoveInfo is used for displaying the snapshot data mover status.
-type SnapshotDataMoveInfo struct {
+// SnapshotDataMovementInfo is used for displaying the snapshot data mover status.
+type SnapshotDataMovementInfo struct {
```
Lyndon-Li, line 69: A repo snapshot ID is also required for data mover

pkg/volume/volume_info_v1.go
```diff
@@ -0,0 +1,54 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package volume
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+type VolumeInfosV1 struct {
```
Lyndon-Li, line 21: In this way, we will have to use VolumeInfosV1 and VolumeInfoV1 everywhere in the code where volume info is required, and then once we want to change the version, all the code need to be updated.
Probably, we can remove these two structures and add the version field into `VolumeInfos` structure.

pkg/volume/volume_info_common.go
blackpiglet: Added.

design/pv_backup_info.md
blackpiglet: Added.

pkg/volume/volume_info_v1.go
blackpiglet: The Golang JSON unmarshal can work even if any fields are not present in the structure, but if we want to work this way, we don't need the version of this structure. 
The version is introduced to work with the potential non-backward compatible changes.
By the suggested way, to me, we already lose the ability to support non-backward compatible change.

Lyndon-Li: Yes, with Golang's JSON capability, somehow the version field is not a must-have. I am fine with the either way --- having the version field or not having it. From me, having the version field helps a little bit on the data verification:

- With the version field, we can check whether a field in `VolumeInfos` is a required one, if it is required, we can fail the process.
- For example, if a field xxx is empty, if the volume info is from an old backup, we can let it go; while if it is from a new backup, we should not continue.

blackpiglet: If we only have a version variable, without other information, I don't think the code knows how to check which fields are mandatory. 
If it's possible some versions must have some mandatory fields, then we need to have structures to let the code know that.

design/pv_backup_info.md
anshulahuja98: I think I completely understand that currently this is meant for making the restore flow deterministic in some way.

Decoupling from this design / approach from now, I think similar to how we have backup/restore results to show warnings/ errors. Having the same here would be helpful. This can basically be shown during velero backup describe etc.
It would be more of a PV level backup result struct **. I am fine with not tackling it right now.**

Also on immutability - I think the doc is the opposite, we are not guaranteeing immutability in any way. So updating this file during async should be fine.

now on the other hand can you explain certain behaviours - 
Example - during backup a PVC went to CSI datamovement based backup. VolumeInfo reflects this information. Now let's say during async phase the datamovement fails, the backup will be marked partially failed. 


However when we restore from this backup, current behavior that I recall is that Restore will be marked Completed since the backup didn't even have the failed snapshots. However, now that this struct is our source of truth, we'll try to restore such volumes and potentially go to PartiallyFailed.

Or is this behaviour disparity handled in some way already?


I think if not maintaining the error message, there is value in keeping success/ failure of each snapshot through a particular route.

blackpiglet: I agree that there was no guarantee of immutability from Velero, but that was broken after the async operations were introduced. The concern is that introducing more changes that need to modify the metadata after it's created, will be hard to achieve immutability someday.

For the case that CSI backing up completes, but data upload fails, I can not confirm, but Velero should check the DataUpload's information. During restore, Velero will create a ConfigMap according to the DataUpload, then a DataDownload will be created according to the ConfigMap. I didn't find any logic to check the DataUpload's result, so I suppose the restore will also end with partial failure.

The existing Volume backup information, including DataUpload and PVB, still works after the VolumeInfo structure is introduced. How about if there is not enough information in the VolumeInfo, fall back to the previous way?

anshulahuja98: My worry overall is that we are making this struct to be in some sense the source of truth of how a PV is backed up. It is just missing the truth about if it was backed up successfully or not. 
Falling back to existing information kind of defeats the purpose of making it the source of truth.

pkg/volume/volume_info_common.go
```diff
@@ -0,0 +1,136 @@
...
+	// The PVC's name.
+	PVCName string `json:"pvcName"`
+
+	// The PVC's namespace
+	PVCNamespace string `json:"pvcNamespace"`
+
+	// The PV name.
+	PVName string `json:"pvName"`
+
+	// The way the volume data is backed up. The valid value includes `VeleroNativeSnapshot`, `PodVolumeBackup` and `CSISnapshot`.
```
Lyndon-Li, line 39: Let's define some constants for these methods

design/pv_backup_info.md
blackpiglet: After a discussion in the Slack channel, we agreed to move on with PR, and we will continue with this topic in the future design of this area.

pkg/volume/volume_info_common.go
blackpiglet: Added.

pkg/volume/volume_info_common.go
```diff
@@ -0,0 +1,144 @@
...
+	BackupMethod string `json:"backupMethod,omitempty"`
+
+	// Whether the volume's snapshot data is moved to specified storage.
+	SnapshotDataMoved bool `json:"snapshotDataMoved"`
```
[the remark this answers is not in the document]
Lyndon-Li: Suggestion: make the field for different backup types as point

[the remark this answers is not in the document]
blackpiglet: PreserveLocalSnapshot filed is added.

pkg/volume/volume_info_common.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// It's the filesystem repository's snapshot ID.
+	SnapshotHandle string `json:"snapshotHandle"`
+}
+
+// VeleroNativeSnapshotInfo is used for displaying the Velero native snapshot status.
+// A Velero Native Snapshot is a cloud storage snapshot taken by the Velero native
+// plugins, e.g. velero-plugin-for-aws, velero-plugin-for-gcp, and
+// velero-plugin-for-microsoft-azure.
+type VeleroNativeSnapshotInfo struct {
```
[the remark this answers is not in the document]
blackpiglet: Modified.

anshulahuja98 (dismissed):

Lyndon-Li (approved):

ywk253100 (approved):

PR #7076, opened 2023-11-08
Update backup log to reflect appropriate backup phase

shubham-pampattiwar:

The current `Backup Completed` log may not be appropriate for cases when there are some associated asynchronous plugin actions still being processed, so this PR updates the log to reflect the appropriate backup phase as per the ongoing backup process phase.


pkg/controller/backup_controller.go
```diff
@@ -792,7 +792,7 @@ func (b *backupReconciler) runBackup(backup *pkgbackup.Request) error {
 		}
 	}
 
-	b.logger.WithField(Backup, kubeutil.NamespaceAndName(backup)).Info("Backup completed")
+	b.logger.WithField(Backup, kubeutil.NamespaceAndName(backup)).Info(fmt.Sprintf("Initial backup processing complete, moving to %s", backup.Status.Phase))
```
kaovilai, line 795:
```suggestion
	b.logger.WithField(Backup, kubeutil.NamespaceAndName(backup)).Info(fmt.Sprintf("Initial backup processing completed, moving to %s", backup.Status.Phase))
```

kaovilai (changes requested):

pkg/controller/backup_controller.go
mateusoliveira43, line 795: Instead of `fmt.Sprintf`, can you use `InfoF`? (maybe aldo need to add line break `\n`)

sseago: I think "complete" is better. i.e. "processing is complete" -- it avoids confusion with the "Completed" phase.

sseago: Infof is better than Sprintf into Info, although I don't think we generally include line breaks in log messages.

mateusoliveira43: Just worried that somethings add line break at the end (like fmt.Println), but others don't (like fmt.Printf), to do not have disorganized logs

sseago: @mateusoliveira43 I'm not seeing any other log messages w/ `\n` in the codebase

kaovilai: each log call should be single line. New line will always have timestamp in front.

mateusoliveira43: Thanks for the explanation, no line break needed :smile:

shubham-pampattiwar: Done, Updated the PR @mateusoliveira43 @sseago @kaovilai

mateusoliveira43 (approved):

sseago (approved):

kaovilai (approved):

blackpiglet (approved):

PR #7091, opened 2023-11-13
Don't fail backup/restore on velero server restart in PhaseWaitingFor…

### LINKED ISSUE #6727: Ensure recoverability of backup/restore from WaitingForPluginoperations state during velero server restart

After the integration of BIAv2 based plugins, after the backup/restore's core flow is done, thy are marked with WaitingForPluginoperations  phase after which velero polls async the plugin operations to complete.

1.  We need a POC to confirm if this polling resumes if a pod restart happens - by velero fetching the ongoing plugin operations from object store and polling them
2. If the above is not supported, look at gap areas of what is required to support it, given that plugin operations can now take a long time, velero should be resilient in tracking them.
3.Once we have clarity on above, in context of CSI datamover impl, we should ensure the above assumptions are not broken and we can recover tracking DataUpload/Download etc.

**Environment:**

**Vote on this issue!**

anshulahuja98:
…PluginOperation

Fixes #6727 

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -1081,13 +1081,13 @@ func markInProgressBackupsFailed(ctx context.Context, client ctrlclient.Client, 
 	}
 
 	for i, backup := range backups.Items {
-		if backup.Status.Phase != velerov1api.BackupPhaseInProgress && backup.Status.Phase != velerov1api.BackupPhaseWaitingForPluginOperations {
+		if backup.Status.Phase != velerov1api.BackupPhaseInProgress {
 			log.Debugf("the status of backup %q is %q, skip", backup.GetName(), backup.Status.Phase)
 			continue
 		}
 		updated := backup.DeepCopy()
 		updated.Status.Phase = velerov1api.BackupPhaseFailed
-		updated.Status.FailureReason = fmt.Sprintf("found a backup with status %q during the server starting, mark it as %q", velerov1api.BackupPhaseInProgress, updated.Status.Phase)
+		updated.Status.FailureReason = fmt.Sprintf("found a backup with status %q during the server starting, mark it as %q", backup.Status.Phase, updated.Status.Phase)
 		updated.Status.CompletionTimestamp = &metav1.Time{Time: time.Now()}
 		if err := client.Patch(ctx, updated, ctrlclient.MergeFrom(&backups.Items[i])); err != nil {
 			log.WithError(errors.WithStack(err)).Errorf("failed to patch backup %q", backup.GetName())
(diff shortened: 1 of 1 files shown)


blackpiglet (approved):

qiuming-best:
@anshulahuja98 
In the case where CRDs for backup or restore are in the 'InProgress' or 'WaitingForPluginOperations' phase, should we consider additional enhancements?

One scenario to note is when other Velero pods have restarted, but the Velero pod responsible for handling Kopia backup and restore operations is still running.

pkg/cmd/server/server.go
```diff
@@ -1081,13 +1081,13 @@ func markInProgressBackupsFailed(ctx context.Context, client ctrlclient.Client,
 	}
 
 	for i, backup := range backups.Items {
-		if backup.Status.Phase != velerov1api.BackupPhaseInProgress && backup.Status.Phase != velerov1api.BackupPhaseWaitingForPluginOperations {
```
qiuming-best, line 1084: For data mover,   the backup CRD  will be in the `BackupPhaseWaitingForPluginOperations` phase when the `node-agent ` uploads files, assuming that the node-agent pod restarted at this moment  Coincidentally, if we don't mark CRDs failed, the backup would be stuck in `InProgress`

anshulahuja98: I don't think I fully understood the scenario.
can you try to elaborate.

sseago: Backup can't be stuck `InProgress` if it's already in `WaitingForPluginOperations`. If the velero pod restarts while `WaitingForPluginOperations`, there's no need to fail it. On next reconcile, it will call `Progress` on any in-progress item operations, including DataUpload operations. Since that's handled in the node agent, they aren't interrupted by velero pod restart, so the backup should be able to move to `Completed` as soon as uploads comlete.

qiuming-best: Sorry, you are right. this is the logic in the Velero server pod but not the `node-agent`, this adjustment of the code logic is correct

qiuming-best:
> > @anshulahuja98 In the case where CRDs for backup or restore are in the 'InProgress' or 'WaitingForPluginOperations' phase, should we consider additional enhancements?
> > One scenario to note is when other velero Pods have restarted, but the node-agent pod responsible for handling Kopia backup and restore operations is still running.
> 
> Yes, I was also thinking about this. Basically, the edge case scenario I could think of were the foll:
> 
> Scenario 1 DU CR (DataUploadPhaseCompleted) + Velero (InProgress) And velero pod restarts - mark velero backup fail - orphan Completed DU CRs left in cluster - we should somehow delete the backed-up data, is invoking delete on DU CR enough?
> 
> Scenario 2 (called out by you) DU CR (DataUploadPhaseInProgress) + Velero (InProgress) And velero pod restarts - mark velero backup fail - we should somehow delete the partially backed up data. Perhaps this is already being done, not sure.
> 
> With my current knowledge I was not able to fully figure out how to achieve these 2 scenarios, so I thought I'll take this as a different issue later. Current PR is more important for velero resiliency and the Scenario 1,2 above is more about orphan data. I wanted to decouple from current PR these improvements.

Sorry for my mistake, this PR enhancement is only for `velero pod` restart not including `node-agent`,  the code changes look good.

I think the changes for Scenario 2 are enough currently. and I agree with you that we needn't consider clearing up orphan data in this PR, and currently, we haven't cleared the orphan data either.

pkg/cmd/server/server.go
anshulahuja98: resolving

anshulahuja98:
@qiuming-best  / @sseago can you pl review /approve?

qiuming-best (approved):

Issue #7068, opened 2023-11-07
Data mover failed - error to delete volume snapshot content - not found


trunet:
**What steps did you take and what happened:**

I created a schedule with velero csi snapshot and data mover enabled. I'm using [openebs/zfs-localpv](https://github.com/openebs/zfs-localpv) CSI.

17 succeeded, 1 failed to data move:
```
  Operation for persistentvolumeclaims mariadb/storage-mariadb-0:
    Backup Item Action Plugin:  velero.io/csi-pvc-backupper
    Operation ID:               du-1a7e0ab8-ec92-4460-b617-dab149484bf3.47e56040-6544-41b01f5b1
    Items to Update:
                           datauploads.velero.io velero/k8s-20231107030552-f7s9t
    Phase:                 Failed
    Operation Error:       error to expose snapshot: error to delete volume snapshot content: error to delete volume snapshot content: volumesnapshotcontents.snapshot.storage.k8s.io "snapcontent-d633b8c9-2d5d-420e-8b96-666150fc45c1" not found
    Progress description:  Failed
    Created:               2023-11-07 04:06:34 +0100 CET
    Started:               2023-11-07 04:06:34 +0100 CET
    Updated:               2023-11-07 04:06:50 +0100 CET
```

**What did you expect to happen:**

Backup should complete successfully.

**The following information will help us better understand what's going on**:

[bundle-2023-11-07-10-51-47.tar.gz](https://github.com/vmware-tanzu/velero/files/13278412/bundle-2023-11-07-10-51-47.tar.gz)

**Anything else you would like to add:**

**Environment:**

- Velero version: main
- Velero features: NOT SET
- Kubernetes version: Server Version: v1.28.3
- Kubernetes installer & version: Talos
- Cloud provider or hardware configuration: Talos bare-metal
- OS (e.g. from `/etc/os-release`): Talos v1.5.4

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

Lyndon-Li:
@trunet Could you help to confirm it is an interim problem or you keep seeing this error with zfs-localpv?

[...]

PR #7102, opened 2023-11-15
Issue 7068: add a finalizer to protect retained VSC

Lyndon-Li:
Fix issue https://github.com/vmware-tanzu/velero/issues/7068, due to a behavior of CSI external snapshotter, manipulations of VS and VSC may not be handled in the same order inside external snapshotter as the API is called. So add a protection finalizer to ensure the order

### DIFF
--- a/pkg/util/csi/volume_snapshot.go
+++ b/pkg/util/csi/volume_snapshot.go
@@ -152,36 +152,53 @@ func EnsureDeleteVS(ctx context.Context, snapshotClient snapshotter.SnapshotV1In
 	return nil
 }
 
+func RemoveVSCProtect(ctx context.Context, snapshotClient snapshotter.SnapshotV1Interface, vscName string, timeout time.Duration) error {
+	err := wait.PollImmediate(waitInternal, timeout, func() (bool, error) {
+		vsc, err := snapshotClient.VolumeSnapshotContents().Get(ctx, vscName, metav1.GetOptions{})
+		if err != nil {
+			return false, errors.Wrapf(err, "error to get VolumeSnapshotContent %s", vscName)
+		}
+
+		vsc.Finalizers = stringslice.Except(vsc.Finalizers, volumeSnapshotContentProtectFinalizer)
+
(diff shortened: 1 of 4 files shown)


pkg/util/csi/volume_snapshot.go
```diff
@@ -152,36 +152,53 @@ func EnsureDeleteVS(ctx context.Context, snapshotClient snapshotter.SnapshotV1In
...
+func RemoveVSCProtect(ctx context.Context, snapshotClient snapshotter.SnapshotV1Interface, vscName string, timeout time.Duration) error {
+	err := wait.PollImmediate(waitInternal, timeout, func() (bool, error) {
+		vsc, err := snapshotClient.VolumeSnapshotContents().Get(ctx, vscName, metav1.GetOptions{})
+		if err != nil {
+			return false, errors.Wrapf(err, "error to get VolumeSnapshotContent %s", vscName)
+		}
+
+		vsc.Finalizers = stringslice.Except(vsc.Finalizers, volumeSnapshotContentProtectFinalizer)
+
+		_, err = snapshotClient.VolumeSnapshotContents().Update(ctx, vsc, metav1.UpdateOptions{})
```
ywk253100, line 164: Why use `update()` rather than `patch()` here?

blackpiglet (approved):

pkg/util/csi/volume_snapshot.go
Lyndon-Li: There is a contest when updating the VSC, the external snapshotter is also manipulating the VSC, specifically, a finalizer has been removed. Therefore, we use update+retry to make sure that what we are updating is with the latest version.
The problem we've seen if we use patch:
`error to remove protect finalizer from vsc snapcontent-85024560-ca99-4f46-8378-5439bdb11a15: error patching VSC: VolumeSnapshotContent.snapshot.storage.k8s.io "snapcontent-85024560-ca99-4f46-8378-5439bdb11a15" is invalid: metadata.finalizers: Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers []string{"snapshot.storage.kubernetes.io/volumesnapshotcontent-bound-protection"`

ywk253100 (approved):

PR #7117, opened 2023-11-17
Add hook status to backup/restore CR

### LINKED ISSUE #6567: Adding hook command status in backup/restore status or describe command
Currently to check if the command in backup/restore hooks got executed successfully or not, we need to browse through the velero logs.

We can either add the command result in Backup Status or propagate that info in describe command.

N/A

**Environment:**

- Velero version (use `velero version`): 1.10.3
- Kubernetes version (use `kubectl version`): 1.25

**Vote on this issue!**

allenxu404:

- Add a hookTracker to backup/restore to track hooks execution status. Hook tracker is extensible to accommodate additional fields as needs develop.
- Add hook execution status to backup/restore describer
- Add UTs
Fixes #6567 


reasonerjt:
The error of the hooks should be passed to the controllers in a more explicit way, and reflect in the Status of the CRs

internal/hook/hook_tracker.go
```diff
@@ -0,0 +1,106 @@
...
+// hookTrackerVal records the execution status of a specific hook.
+// hookTrackerVal is extensible to accommodate additional fields as needs develop.
+type hookTrackerVal struct {
+	// HookFailed indicates if hook failed to execute.
+	HookFailed bool
+}
+
+// HookTracker tracks all hooks' execution status
+type HookTracker struct {
+	Tracker map[hookTrackerKey]hookTrackerVal
```
reasonerjt, line 55: does this have to be exposed since you have `Add` `Update` funcs in this struct to access this map?

internal/hook/hook_tracker.go
```diff
@@ -0,0 +1,106 @@
...
+	HookSourceAnnotation = "annotation"
+	HookSourceSpec       = "spec"
+)
+
+// hookTrackerKey identifies a backup/restore hook
+type hookTrackerKey struct {
+	// PodNamespace indicates the namespace of pod where hooks are executed.
+	// For hooks specified in the backup/restore spec, this field is the namespace of an applicable pod.
+	// For hooks specified in pod annotation, this field is the namespace of pod where hooks are annotated.
+	PodNamespace string
```
reasonerjt, line 29: Do these attributes have to be exposed?

internal/hook/hook_tracker.go
```diff
@@ -0,0 +1,106 @@
...
+
+// hookTrackerVal records the execution status of a specific hook.
+// hookTrackerVal is extensible to accommodate additional fields as needs develop.
+type hookTrackerVal struct {
+	// HookFailed indicates if hook failed to execute.
+	HookFailed bool
+}
+
+// HookTracker tracks all hooks' execution status
+type HookTracker struct {
```
reasonerjt, line 54: I think hooks are also good candidates to be executed in parallel.  We may better consider making this tracker thread-safe.

internal/hook/hook_tracker.go
```diff
@@ -0,0 +1,106 @@
...
+
+// NewHookTracker creates a hookTracker.
+func NewHookTracker() *HookTracker {
+	return &HookTracker{
+		Tracker: make(map[hookTrackerKey]hookTrackerVal),
+	}
+}
+
+// Add adds a hook to the tracker
+func (ht *HookTracker) Add(podNamespace, podName, container, source, hookName string, hookPhase hookPhase) {
```
reasonerjt, line 66: For simplicity, we may start by tracking the result of the hook, after the execution is finished.

internal/hook/item_hook_handler.go
```diff
@@ -251,29 +259,36 @@ func (h *DefaultItemHookHandler) HandleHooks(
...
-							"hookType":   "exec",
-							"hookPhase":  phase,
-						},
-					)
-					err := h.PodCommandExecutor.ExecutePodCommand(hookLog, obj.UnstructuredContent(), namespace, name, resourceHook.Name, hook.Exec)
-					if err != nil {
-						hookLog.WithError(err).Error("Error executing hook")
-						if hook.Exec.OnError == velerov1api.HookErrorModeFail {
-							return err
+					hookTracker.Add(namespace, name, hook.Exec.Container, HookSourceSpec, resourceHook.Name, phase)
```
reasonerjt, line 266: If we choose to record the result of the hook in the tracker, we can add record after the execution of the hook.

internal/hook/hook_tracker.go
allenxu404: Currently hooks are executed in sequence.  so instead of adding a lock, can we just add a comment "Not thread-safe" to remark that? It helps to save time on addtional lock operation.

 If hooks change to run concurrently in the future, locking could then be implemented to make HookTracker thread-safe as needed.

internal/hook/hook_tracker.go
allenxu404: If hooks use a "Fail" onError mode, Velero may underreport `hooksAttempted`. 

Specifically, if a failing hook's onError is set to "Fail", Velero will return without running subsequent hooks. If we start tracking the result after the hooks' execution, this could result in the number of attempted hooks being less than the actual number defined.

internal/hook/hook_tracker.go
anshulahuja98: I don't think the lock operation would be that time consuming..
would be better to add upfront

config/crd/v1/bases/velero.io_backups.yaml
```diff
@@ -544,6 +544,13 @@ spec:
                 description: FormatVersion is the backup format version, including
                   major, minor, and patch version.
                 type: string
+              hooksAttempted:
```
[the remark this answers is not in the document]
allenxu404: Make sense, Will do.

internal/hook/hook_tracker.go
allenxu404: No, will change it.

internal/hook/hook_tracker.go
allenxu404: No, will change it.

internal/hook/hook_tracker.go
allenxu404: OK, will add a lock to it.

internal/hook/hook_tracker.go
reasonerjt: I misunderstood, I think attempted should indicate the number of hooks executed.

allenxu404: Sync with @reasonerjt , HooksAttempted = HooksFailed + HooksSucc

Will follow this pattern to change the code.

allenxu404:
Sync with @anshulahuja98, this PR solely adds the "hooksAttempted" and "hooksFailed" fields to  backup/restore CR to provide initial hook execution metadata. Tracking additional details like hook error information and skipped hooks can be supported in the future via a separate follow-up effort. 

Once this PR is merged, I will create a new issue to track these potential enhancements.

internal/hook/item_hook_handler.go
```diff
@@ -251,29 +261,38 @@ func (h *DefaultItemHookHandler) HandleHooks(
...
+					// The remaining hooks will only be executed if modeFailError is nil.
+					// Otherwise, execution will stop and only hook collection will occur.
+					if modeFailError == nil {
+						hookLog := log.WithFields(
+							logrus.Fields{
+								"hookSource": "backupSpec",
```
anshulahuja98, line 274: can we use the constants we have, here for source, type

pkg/backup/backup.go
```diff
@@ -427,8 +428,15 @@ func (kb *kubernetesBackupper) BackupWithResolvers(log logrus.FieldLogger,
 	updated.Status.Progress.TotalItems = len(backupRequest.BackedUpItems)
 	updated.Status.Progress.ItemsBackedUp = len(backupRequest.BackedUpItems)
 
+	// update the hooks execution status
+	if updated.Status.HookStatus == nil {
+		updated.Status.HookStatus = &velerov1api.HookStatus{}
+	}
+	updated.Status.HookStatus.HooksAttempted, updated.Status.HookStatus.HooksFailed = itemBackupper.hookTracker.Stat()
```
anshulahuja98, line 435: this live tracking might be helpful to debug stuck backups. Nice.

pkg/backup/backup.go
```diff
@@ -427,8 +428,15 @@ func (kb *kubernetesBackupper) BackupWithResolvers(log logrus.FieldLogger,
...
+	log.Infof("hook Tracker: %+v", itemBackupper.hookTracker.GetTracker())
```
anshulahuja98, line 436: This should be fine for now, but my slight worry is this log will come very frequently, every time status is updated.
perhaps make it such that this log is skipped if no hooks are executed. The struct is empty.
To reduce the log frequency.

anshulahuja98, line 436: enhance this log message little bit?

anshulahuja98 (changes requested):

internal/hook/item_hook_handler.go
allenxu404: Sure.

pkg/backup/backup.go
allenxu404: Sure.

anshulahuja98 (approved):

reasonerjt (approved):

internal/hook/hook_tracker.go
```diff
@@ -0,0 +1,137 @@
...
+	if _, ok := ht.tracker[key]; !ok {
+		ht.tracker[key] = hookTrackerVal{
+			hookFailed:   false,
+			hookExecuted: false,
+		}
+	}
+}
+
+// Record records the hook's execution status
+func (ht *HookTracker) Record(podNamespace, podName, container, source, hookName string, hookPhase hookPhase, hookFailed bool) {
```
reasonerjt, line 94: nit: Here a caller may call `Record` without Add but the result is dropped silently.

This is not a blocker but please consider combining `Add` and `Record` with the execution result an optional parameter, or explicitly emphasizing it in the comment of `Record` that the caller must call `Add` first

PR #7172, opened 2023-12-02
Track status of volumesnapshots to update backup via backup_operations_controller

kaovilai:

Fixes #(issue)

### DIFF
--- a/pkg/backup/snapshots.go
+++ b/pkg/backup/snapshots.go
@@ -2,6 +2,7 @@ package backup
 
 import (
 	"context"
+	"fmt"
 
 	snapshotv1api "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


sseago:
@kaovilai Note that if you're tracking this during the backup operations controller, you'll need to do the first pass in the backup controller. The first pass of calling the Progress method on all async operations happens at the end of the InProgress processing in the backup controller immediately following completion of item backup. If everything is done at this point (which may happen if there aren't many volumes, volumes are small-ish, and there are a lot of non-PVC items backed up after pods/pvcs), then the WaitingForPluginOperations phase is skipped entirely, and the backup moves straight to Finializing. In this case, the bacup operations controller won't reconcile the backup at all.

kaovilai:
> WaitingForPluginOperations phase is skipped entirely

CSI plugin is now biav2, so it should be in one of the backup item operations.

If we can get it's operationID, and that OperationProgress is completed without errors then we can consider that CSI snapshot completed successfully.

sseago:
@kaovilai "CSI plugin is now biav2, so it should be in one of the backup item operations."

Yes -- I'm just saying that whatever you do in backup_operations_controller must also be done in backup_controller after the first Progress check, since if all of the backup operations (including CSI plugin operations) have already completed when the transition from InProgress happens, then the backup goes straight to finalizing, since there's no need to poll the backup for progress later, since we're already done.

In other words, Velero calls Progress repeatedly for each incomplete operation until they're all either completed or failed, but the *first* polling happens from the backup controller before leaving the InProgress phase, right before we upload the backup tarball the first time. This way if the operations are all quick, we don't need to wait/poll, so the WaitingForPluginOperations phase is skipped completely.

kaovilai:
Thanks for the note

PR #7173, opened 2023-12-04
fix: API types enum validations

mateusoliveira43:

Fix lack of enum valiadtions for API types, in `pkg/apis/velero` folder, and clean up around constants.

No.


pkg/apis/velero/v1/backup_repository_types.go
```diff
@@ -51,9 +51,6 @@ const (
 	BackupRepositoryPhaseNew      BackupRepositoryPhase = "New"
 	BackupRepositoryPhaseReady    BackupRepositoryPhase = "Ready"
 	BackupRepositoryPhaseNotReady BackupRepositoryPhase = "NotReady"
-
-	BackupRepositoryTypeRestic string = "restic"
```
reasonerjt, line 55: I don't have a very strong opinion but why is this needed?

config/crd/v2alpha1/bases/velero.io_datauploads.yaml
```diff
@@ -115,6 +115,8 @@ spec:
               snapshotType:
                 description: SnapshotType is the type of the snapshot to be backed
                   up.
+                enum:
+                - CSI
```
reasonerjt, line 119: Not sure if we really want this 
cc @Lyndon-Li

pkg/apis/velero/v1/backup_repository_types.go
mateusoliveira43: If you mean the variable, it is used in these places https://github.com/search?q=repo%3Avmware-tanzu%2Fvelero%20BackupRepositoryTypeRestic&type=code

If you mean the change to another file, is just for semantics (it is a constant, not a type used to generate the CRD, so lets move to constants file)

config/crd/v2alpha1/bases/velero.io_datauploads.yaml
mateusoliveira43: The code implicitly tell us that. If the intent is to let an open string field, would remove this type from apis file

kaovilai: I believe datauploads CRD is meant for any data movers. CSI being the first one implemented by velero. Others could come in with Database snapshot uploader etc..

reasonerjt: I don't think we wanna introduce frequent changes to APIs.

shubham-pampattiwar: +1 @kaovilai

kaovilai: This enum needs to be removed IMO. It's an open value per design.. or else comment in the design that future datauploaders need to add values to this enum

mateusoliveira43: Can you point to the design?

I found this, which implies it is a enum https://github.com/vmware-tanzu/velero/blob/main/design/volume-snapshot-data-movement/volume-snapshot-data-movement.md?plain=1#L162

kaovilai: >  SnapshotType is the type of the snapshot to be backed up.

https://github.com/vmware-tanzu/velero/blob/e65ef2894868242d51a92e1fba26423adb6ca49f/design/volume-snapshot-data-movement/volume-snapshot-data-movement.md?plain=1#L272-L275C25

Plan is to open up this to more values in the future.

mateusoliveira43:
As discussed yesterday with @kaovilai @sseago and @shawn-hurley , I will close this PR as we do not want to change an API field without updating its version, so users can have the best experience possible while using the product

Will open follow up PR just to add comments on Go code, on why those fields do not use enum validation

Issue #6023, opened 2023-03-23
Enable more linters for Velero


blackpiglet:
**Describe the problem/challenge you have**

**Describe the solution you'd like**

**Anything else you would like to add:**

**Environment:**

- Velero version (use `velero version`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

reasonerjt:
@blackpiglet 
As discussed we certainly will make enhancements in the linter and static code check, please add more issues to specify what are planned.

blackpiglet:
- asasalint
  - asciicheck
  - bidichk
  - bodyclose
  - containedctx
  - depguard
  - dogsled
  - dupword
  - durationcheck
  - errcheck
  - errchkjson
  - exportloopref
  - gci
  - ginkgolinter
  - goconst
  - gocritic
  - godot
  - gosimple
  - govet: #6154 
  - importas
  - ineffassign
  - misspell
  - nakedret
  - nilerr
  - noctx
  - nolintlint
  - nosprintfhostport
  - prealloc
  - predeclared
  - revive
  - rowserrcheck
  - stylecheck
  - thelper
  - typecheck
  - unconvert
  - usestdlibvars
  - whitespace

kaovilai:
Damn..

blackpiglet:
Yeah. There is still a long road ahead.
But don't worry. Some linters require many changes already in progress, such as revive, staticcheck, and stylecheck, of course, also including govet. Thanks for your contribution.

And many linters don't require a lot of hassle. Even no change is needed.

blackpiglet:
@kaovilai Thanks for the suggestion.

blackpiglet:
v1.12 related issues are all resolved.

reasonerjt:
Will reusing this issue to track the work in v1.13

blackpiglet:
After #7175 merging, v1.13's jobs are done for this task.

[...]

PR #7194, opened 2023-12-10
fix: Clean up golangci-lint config

mateusoliveira43:

Clean up of `golangci-lint` config file:
- removed unused configs
  - 18 linters-settings (dupl, exhaustive, funlen, gocognit, nestif, gocritic, gocyclo, godot, godox, golint, gomnd, gomodguard, depguard, lll, maligned, prealloc, rowserrcheck and testpackage) were declared that would do nothing, since they are not enabled in linters sections
- updated wrong configs
  - added `config verify` command to CI to avoid this in the future
- removed doc comments and add `golangci-lint` doc link
  - file was generated copying example from golangci-lint, which is not a configuration example per https://github.com/golangci/golangci-lint/blob/7f6606a7ea1ba30ca5f56fdc4f11e9d318e6cbec/.golangci.reference.yml#L4
  - this makes reading config file a lot easier. File will look like this after merge https://github.com/mateusoliveira43/velero/blob/fix/golangci-lint-config/.golangci.yaml
  - easier to catch wrong config as noted here https://github.com/vmware-tanzu/velero/issues/7801

Also: 
- Delete unnecessary scripts (`golangci-lint` already does that)
- Update documentation around linter (which was outdated)
- sorted things alphabetically to more easy read of the config

This PR would also change golangci-lint config file to default name and add golangci-lint github action, which was already done in https://github.com/vmware-tanzu/velero/commit/884bcbec9839601639eb7f7c31d87faf94abadd0 and rebased on top of it.

Related to https://github.com/vmware-tanzu/velero/issues/6023


kaovilai:
Why use `make lint-fix` over `make update`?

kaovilai:
`make update` is how we today make sure the format satisfy the ci lint. If make lint-fix is replacing that flow then note that in the PR description.

kaovilai:
This PR breaks `make update` by removing `hack/update-1fmt.sh` that is relied upon by 

```
update:
	@$(MAKE) shell CMD="-c 

> 'hack/update-all.sh'"

```
https://github.com/vmware-tanzu/velero/blob/dcd4392744b10cde4984de7d1a423aff3dd74fa5/Makefile#L234C1-L235C46

which runs

```
for f in ${HACK_DIR}/update-*.sh; do
  if [[ $f = "${HACK_DIR}/update-all.sh" ]]; then
    continue
  fi
  $f
done
```
https://github.com/vmware-tanzu/velero/blob/dcd4392744b10cde4984de7d1a423aff3dd74fa5/hack/update-all.sh

Correct me if I'm wrong but is this PR created because you were not aware of the `make update`?

If this PR merge, we now have to run `make update lint-fix` instead to fix lint and generate CRDs etc.

mateusoliveira43:
@kaovilai I did not intend to change `make update` at first. That command would run `gofmt` and `goimports` (which is a different thing then generating files) and generate files, right?

I removed `hack/update-1fmt.sh` because we would be doing the same thing twice (with the script and with `golangci-lint`). `make-lint-fix` only replaces the formatting script, still need to run `make update` for the scenarios described in documentation https://github.com/vmware-tanzu/velero/blob/main/site/content/docs/main/development.md#update-generated-files
> Run `make update` to regenerate files if you make the following changes:
>
> * Add/edit/remove command line flags and/or their help text
> * Add/edit/remove commands or subcommands
> * Add new API types
> * Add/edit/remove plugin protobuf message or service definitions

So, removed the script for both semantic and performance.

Since `update-all` script runs a `for` loop, I just removed a item from the list, but it will still work (example is Pull Request CI Check for this PR).

No no, my intention was to clean up config file for `golangci-lint`. There was unused config (for example, linters-settings for linters not enabled, enabled linters without config, etc), found the other issues along the way. I would be ok to break the PR in smaller ones.

Added to PR description a note about removal of scripts.

kaovilai:
> I just removed a item from the list, but it will still work

Just in github actions because it's covered by golangci-lint or also locally on dev machine when running

Is the expectation that `make update` should still format on local dev machine?

mateusoliveira43:
@tiger oh no, `make update` will not format Go code anymore (but as it was, it would not run all linters defined in `golangci-lint` config file, only `gofmt` and `goimport`. So, in the past running `make update` would not guarantee CI would pass)

kaovilai:
> make update will not format Go code anymore
> So, in the past running make update would not guarantee CI would pass)

So now we run `make update lint-fix` before making PR to guarantee CRD generation and also lint right?

mateusoliveira43:
@kaovilai that's right!

hack/update-1fmt.sh
```diff
@@ -1,63 +0,0 @@
...
-
-files="$(find . -type f -name '*.go' -not -path './.go/*' -not -path './vendor/*' -not -path './site/*' -not -path '*/generated/*' -not -name 'zz_generated*' -not -path '*/mocks/*')"
-echo "${ACTION} gofmt"
-output=$(gofmt "${MODE}" -s ${files})
-if [[ -n "${output}" ]]; then
-  VERIFY_FMT_FAILED=1
-  echo "${output}"
-fi
-if [[ -n "${VERIFY_FMT_FAILED:-}" ]]; then
-  echo "${ACTION} gofmt - failed! Please run 'make update'."
```
kaovilai, line 44: Please add a replacement for `  echo "${ACTION} gofmt - failed! Please run 'make update'."` that was previously visible via github actions log.

You can add this to curent linter check action by [setting continue-on-error: true](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)

Then you would add [status check function](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions) in the next step and force a failure if prior step fail, but this time you echo "Please run `make lint-fix`"

kaovilai (changes requested):

hack/update-1fmt.sh
mateusoliveira43: did not understand the request. Do you want to add a message saying what to do if this action https://github.com/vmware-tanzu/velero/actions/runs/7462814485/job/20306039678?pr=7194 fails?

mateusoliveira43:
```yaml
    - name: developer instruction
      if: ${{ failure() }}
      run: echo "Please fix linter issues. Run `make lint-fix` to help you"
```

kaovilai: Sure. Yup.

mateusoliveira43: alright, will try if my example works

kaovilai: ` echo "Please fix linter issues. Run `make lint-fix` to help you" && exit 1`

mateusoliveira43: added instruction in documentation, we can add this in follow up PR

since this PR is more than 3 months old, I want to just change things absolute necessary in it

if you do not mind, I will resolve this one

mateusoliveira43:
@mmorel-35 could you please review?

changelogs/unreleased/7194-mateusoliveira43
kaovilai, line 1: don't think this is neccessary as there should be no functional changes.

.golangci.yaml
kaovilai, line 1: Where was prior config if this is new?

hack/lint.sh
```diff
@@ -14,14 +14,12 @@
...
 
-HACK_DIR=$(dirname "${BASH_SOURCE[0]}")
-
 # Printing out cache status
 golangci-lint cache status
 
 # Enable GL_DEBUG line below for debug messages for golangci-lint
 # export GL_DEBUG=loader,gocritic,env
-CMD="golangci-lint run -c $HACK_DIR/../golangci.yaml"
+CMD="golangci-lint run $@"
```
kaovilai, line 22: Would be great to note that --fix can be passed in here, and that it may not fix every issue.

.golangci.yaml
mateusoliveira43: rebase added it back... :man_facepalming: will rename old one

hack/lint.sh
mateusoliveira43: I think I already added this in here https://github.com/vmware-tanzu/velero/pull/7194/files#diff-3148449172665e1fe59b7d02c5f38a0053ad8e689cc457569092bb395f2312a0

site/content/docs/main/development.md
```diff
@@ -24,16 +24,18 @@ You can run `make verify` to ensure that all generated files (clientset, listers
 
 ## Linting
 
-You can run `make lint` which executes golangci-lint inside the build image, or `make local-lint` which executes outside of the build image.
-Both `make lint` and `make local-lint` will only run the linter against changes.
+You can run `make lint` (which executes [`golangci-lint`](https://github.com/golangci/golangci-lint)) to run linters against the entire Go code base.
 
-Use `lint-all` to run the linter against the entire code base.
+You can run `make lint-fix` to fix fixable issues found by `make lint`.
```
kaovilai, line 29: I would link fixable issues table here.

.golangci.yaml
mateusoliveira43: done

kaovilai (approved):

changelogs/unreleased/7194-mateusoliveira43
mateusoliveira43: Added here because I spend quite some time in this PR and as I added here https://github.com/vmware-tanzu/velero/discussions/6514#discussioncomment-6469010 I do want some credit for this work

.golangci.yaml
```diff
@@ -1,295 +1,74 @@
...
+      - error-is-as
+      - error-nil
+      - expected-actual
+      - go-require
+      - float-compare
+      - require-error
+      - suite-dont-use-pkg
+      - suite-extra-assert-call
+      - suite-thelper
+      - nil-compare
```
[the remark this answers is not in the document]
mateusoliveira43: done in commit https://github.com/vmware-tanzu/velero/pull/7194/commits/0379f2c2484434496051ca47728e86b0fc3de070

.golangci.yaml
```diff
@@ -312,40 +92,45 @@ linters:
...
+        - staticcheck
+      # No need to report deprecate for DefaultVolumesToRestic.
+      text: "DefaultVolumesToRestic"
+     # TODO: remove all excludes for test files
     - path: ".*_test.go$"
       linters:
         - dupword
         - errcheck
+        - ginkgolinter
         - goconst
```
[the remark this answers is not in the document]
mateusoliveira43: done in commit https://github.com/vmware-tanzu/velero/pull/7194/commits/0379f2c2484434496051ca47728e86b0fc3de070

.golangci.yaml
```diff
@@ -356,61 +141,26 @@ issues:
...
+        - bodyclose
         - dupword
         - errcheck
         - goconst
         - gosec
         - gosimple
+        - ginkgolinter
         - nilerr
+        - noctx
         - staticcheck
```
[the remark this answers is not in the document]
mateusoliveira43: done in commit https://github.com/vmware-tanzu/velero/pull/7194/commits/0379f2c2484434496051ca47728e86b0fc3de070

test/e2e/privilegesmgmt/ssr.go
```diff
@@ -26,7 +26,6 @@ import (
 	. "github.com/onsi/gomega"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
-	waitutil "k8s.io/apimachinery/pkg/util/wait"
```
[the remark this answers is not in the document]
mateusoliveira43: done

Makefile
```diff
@@ -236,12 +236,14 @@ endif
 
 lint:
 ifneq ($(SKIP_TESTS), 1)
+	@$(MAKE) shell CMD="-c 'hack/verify-lint-config.sh'"
 	@$(MAKE) shell CMD="-c 'hack/lint.sh'"
 endif
 
-local-lint:
+lint-fix:
```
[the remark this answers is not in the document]
mateusoliveira43: Since its inside an `if`, I think we will write less code by being explicit here instead of writing a new command `verify-config` and adding it to the 2 lint commands

From what I am familiar (example https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/Makefile#L55) `fmt` command is commonly used to call `go fmt`. Since `golangci-lint` does more than that, and the non fix command already contains `lint` in its name, I think `lint-fix` would make more sense here

.golangci.yaml
```diff
@@ -312,36 +94,40 @@ linters:
...
     - linters:
         - staticcheck
-      text: "github.com/golang/protobuf/proto" # grpc-go still uses github.com/golang/protobuf/proto.
+      # grpc-go still uses github.com/golang/protobuf/proto.
+      text: "github.com/golang/protobuf/proto"
     - linters:
         - staticcheck
-      text: "DefaultVolumesToRestic" # No need to report deprecate for DefaultVolumesToRestic.
+      # Kopia still depends on this.
+      text: "github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-06-01/storage"
```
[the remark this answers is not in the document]
mateusoliveira43: done, thanks!

.golangci.yaml
```diff
@@ -1,295 +1,76 @@
...
+      - error-nil
+      - expected-actual
+      - go-require
+      - float-compare
+      - require-error
+      - suite-dont-use-pkg
+      - suite-extra-assert-call
+      - suite-thelper
+    enable:
+      - empty
```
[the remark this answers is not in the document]
mateusoliveira43: done

hack/verify-lint-config.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2017 the Velero contributors.
+# Copyright 2020 the Velero contributors.
```
sseago, line 3: The current dev guidelines call for omitting the year in the Copyright notice.

.golangci.yaml
```diff
@@ -367,50 +146,12 @@ issues:
         - unconvert
         - unparam
         - unused
-
-  # The list of ids of default excludes to include or disable. By default it's empty.
-  include:
-    - EXC0002 # disable excluding of issues about comments from golint
-
-  # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
+  exclude-dirs:
```
shubham-pampattiwar, line 149: Don't we still need `pkg/plugin/generated` to be under exclude-dirs ?

kaovilai: yes

hack/verify-lint-config.sh
mateusoliveira43: done

.golangci.yaml
mateusoliveira43: all files in `pkg/plugin/generated` start with a comment `Code generated by ...`, so those files are excluded by default

Do you prefer to explicitly add it here?

shubham-pampattiwar: yeah lets keep it same as before.

mateusoliveira43: updated

shubham-pampattiwar (approved):

kaovilai (approved):

sseago:
One thing that's not clear to me is whether the difference in file list for gofmt matters. I'm not quite sure how the new approach in this PR generates the  file list for gofmt, but the prior approach explicitly omitted autogenerated files:
```
-files="$(find . -type f -name '*.go' -not -path './.go/*' -not -path './vendor/*' -not -path './site/*' -not -path './.git/*' -not -path '*/generated/*' -not -name 'zz_generated*' -not -path '*/mocks/*')"
```

On the other hand, the old gofmt script in `make update` no longer works for me -- we have too many go files for the explicit arg list to work in bash anymore:
```
Running all update scripts
Updating gofmt
hack/update-1fmt.sh: line 38: /usr/local/go/bin/gofmt: Argument list too long
```
So the new approach in this PR is definitely an improvement -- assuming the right files are being selected for gofmt.

sseago (approved):

mateusoliveira43:
@shubham-pampattiwar done

shubham-pampattiwar (approved):

kaovilai (approved):

sseago:
Note that the prior error I reported with `make update` (involving "argument list too long") ended up being a problem with my particular git checkout. But still, the point here is that gofmt is already run as part of lint, so it shouldn't really be needed as part of "make update" as well.

sseago (approved):

Issue #7263, opened 2024-01-03
Restore is stuck when enabling the informer cache


ywk253100:
When the informer cache is enabled, the restore is stuck if the restore includes `namespace` resource.
Get the following errors in the Velero server logs:
```
W0102 09:48:14.941204       1 reflector.go:424] pkg/mod/k8s.io/client-go@v0.25.6/tools/cache/reflector.go:169: failed to list *unstructured.Unstructured: the server could not find the requested resource
E0102 09:48:14.941248       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.25.6/tools/cache/reflector.go:169: Failed to watch *unstructured.Unstructured: failed to list *unstructured.Unstructured: the server could not find the requested resource
```

The cause of the error `the server could not find the requested resource` for `namespace` is that Velero calls list API of `namespace` with namespace added incorrectly

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

reasonerjt:
The root cause is that in PR #7143 the resource namespace was transformed into restoreableItems and it has the field `targetNamespace` set as the name of the restored namespace, and when the informer cache was added it uses the `targetNamespace` to list the resources.

Adding `targetNamespace` to namespace resoruces may be a bad decision, I'll fix it.

PR #7271, opened 2024-01-04
Make informer cache disabled by default

### LINKED ISSUE #7264: The help message and actual behavior for `disable-informer-cache` mismatch in install CLI
The [help message](https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/cli/install/install.go#L129) says `Default is false (don't disable)` but [the actual behavior](https://github.com/vmware-tanzu/velero/blob/main/pkg/cmd/cli/install/install.go#L158) is disabled by default.

**Vote on this issue!**

ywk253100:
Make informer cache disabled by default

Fixes #7264

Fixes #(issue)

### DIFF
--- a/pkg/install/deployment.go
+++ b/pkg/install/deployment.go
@@ -49,7 +49,7 @@ type podTemplateConfig struct {
 	uploaderType                    string
 	defaultSnapshotMoveData         bool
 	privilegedNodeAgent             bool
-	disableInformerCache            bool
+	enableInformerCache             bool
 	scheduleSkipImmediately         bool
 }
 
@@ -153,9 +153,9 @@ func WithDefaultSnapshotMoveData() podTemplateOption {
 	}
 }
 
-func WithDisableInformerCache() podTemplateOption {
+func WithEnableInformerCache() podTemplateOption {
 	return func(c *podTemplateConfig) {
-		c.disableInformerCache = true
+		c.enableInformerCache = true
 	}
 }
 
@@ -206,8 +206,8 @@ func Deployment(namespace string, opts ...podTemplateOption) *appsv1.Deployment 
 		args = append(args, "--default-snapshot-move-data=true")
 	}
 
-	if c.disableInformerCache {
-		args = append(args, "--disable-informer-cache=true")
+	if c.enableInformerCache {
+		args = append(args, "--enable-informer-cache=true")
 	}
 
 	if c.scheduleSkipImmediately {
--- a/test/e2e/migration/migration.go
+++ b/test/e2e/migration/migration.go
@@ -167,8 +167,7 @@ func MigrationTest(useVolumeSnapshots bool, veleroCLI2Version VeleroCLI2Version)
 					OriginVeleroCfg.Plugins = ""
 					//TODO: Remove this setting when migration path is from 1.13 to higher version
 					//TODO: or self, because version 1.12 and older versions have no this parameter.
-					OriginVeleroCfg.WithoutDisableInformerCacheParam = true
-					OriginVeleroCfg.DisableInformerCache = false
+					OriginVeleroCfg.WithoutEnableInformerCacheParam = true
 				}
 				Expect(VeleroInstall(context.Background(), &OriginVeleroCfg, false)).To(Succeed())
 				if veleroCLI2Version.VeleroVersion != "self" {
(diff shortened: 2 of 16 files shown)


qiuming-best (approved):

reasonerjt (approved):
lgtm

reasonerjt:
@ywk253100 but could you squash the commits?

test/e2e/migration/migration.go
```diff
@@ -167,8 +167,8 @@ func MigrationTest(useVolumeSnapshots bool, veleroCLI2Version VeleroCLI2Version)
 					OriginVeleroCfg.Plugins = ""
 					//TODO: Remove this setting when migration path is from 1.13 to higher version
 					//TODO: or self, because version 1.12 and older versions have no this parameter.
-					OriginVeleroCfg.WithoutDisableInformerCacheParam = true
-					OriginVeleroCfg.DisableInformerCache = false
+					OriginVeleroCfg.WithoutEnableInformerCacheParam = true
+					OriginVeleroCfg.EnableInformerCache = true
```
danfengliu, line 171: Please help to remove line 171: _OriginVeleroCfg.EnableInformerCache = true_, it's not used since .WithoutEnableInformerCacheParam = true

test/e2e/upgrade/upgrade.go
```diff
@@ -147,8 +147,8 @@ func BackupUpgradeRestoreTest(useVolumeSnapshots bool, veleroCLI2Version VeleroC
 				}
 				//TODO: Remove this setting when upgrade path is from 1.13 to higher
 				//TODO: version, or self version 1.12 and older versions have no this parameter.
-				tmpCfgForOldVeleroInstall.WithoutDisableInformerCacheParam = true
-				tmpCfgForOldVeleroInstall.DisableInformerCache = false
+				tmpCfgForOldVeleroInstall.WithoutEnableInformerCacheParam = true
+				tmpCfgForOldVeleroInstall.EnableInformerCache = true
```
danfengliu, line 151: Same as above comment.

pkg/install/deployment.go
```diff
@@ -153,9 +153,9 @@ func WithDefaultSnapshotMoveData() podTemplateOption {
 	}
 }
 
-func WithDisableInformerCache() podTemplateOption {
+func WithEnableInformerCache() podTemplateOption {
 	return func(c *podTemplateConfig) {
-		c.disableInformerCache = true
+		c.enableInformerCache = true
```
shubham-pampattiwar, line 158: Shouldn't this be false ?

sseago: @shubham-pampattiwar the param switches from "disable" to "enable" (so false is disabled), this func is called to *enable*, so it should be true.

pkg/install/deployment.go
```diff
@@ -206,8 +206,8 @@ func Deployment(namespace string, opts ...podTemplateOption) *appsv1.Deployment
 		args = append(args, "--default-snapshot-move-data=true")
 	}
 
-	if c.disableInformerCache {
-		args = append(args, "--disable-informer-cache=true")
+	if c.enableInformerCache {
+		args = append(args, "--enable-informer-cache=true")
```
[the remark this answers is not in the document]
sseago: true enables now, false disables (it was swapped before because we wanted it enabled by default)

sseago (approved):

pkg/install/deployment.go
shubham-pampattiwar: gotcha, thanks !

shubham-pampattiwar (approved):

shawn-hurley:
I want to add my two cents.

Changing defaults should be treated as changing a public-facing API; doing this without warning or without any announcement prior for users seems to me to be pretty user-hostile. Please, reconsider this change.

sseago (changes requested):
Requesting changes as a placeholder here, to make sure we have some follow-on discussion before merge. I've just realized that this isn't new functionality in 1.13 -- we released this in 1.12.1 as well, so the parameter name change could confuse users. This seems like  bad UX choice. A couple of options:

1) Keep the param name as-is but force a `true` default in the installer -- this minimizes API changes while still allowing for disabling
2) Don't make the change at all. The reason for submitting this was that this feature was breaking with included namespaces, but we found the source of that bug, which was a recently-introduced change to namespace handling and has since been fixed, so this may not even be needed now.

kaovilai:
> 2. Don't make the change at all. The reason for submitting this was that this feature was breaking with included namespaces, but we found the source of that bug, which was a recently-introduced change to namespace handling and has since been fixed, so this may not even be needed now.

Adding context `bug...has since been fixed` that sseago is referring: https://github.com/vmware-tanzu/velero/issues/7263

ywk253100:
We reached an agreement in the community meeting that the informer cache will be enabled by default.  i.e. we keep the disable-... flag and only fix the installer part. I'm closing this PR and will create a new one

PR #7344, opened 2024-01-23
Design: Velero client download APIServer

### LINKED ISSUE #7432: Design: Download server for Velero client
Design for https://github.com/vmware-tanzu/velero/issues/6167

kaovilai:

Fixes #7432

Slack threads:
- examples of immutable resources https://kubernetes.slack.com/archives/C021GPR1L3S/p1709329227490779


design/velero-client-download-apiserver_design.md
```diff
@@ -0,0 +1,73 @@
...
+## Non Goals
+- A short list of items which are:
+- a. out of scope
+- b. follow on items which are deliberately excluded from this proposal.
+
+
+## High-Level Design
+One to two paragraphs that describe the high level changes that will be made to implement this proposal.
+
+We will [configure a Kubernetes API Aggregation Layer](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/) to communicate with a new Velero extension apiserver in order to authenticate and authorize requests to download data from the Velero server.
```
kaovilai, line 39: I have been cautioned around using this.

An alternative is to have an Ingress reachable by velero CLI to query.

kaovilai: We can utilize https://dev-k8sref-io.web.app/docs/authentication/tokenreview-v1/
https://dev-k8sref-io.web.app/docs/authorization/subjectaccessreview-v1/ for RBAC of the authorized velero user.

Moving to using Ingress to a velero pod that can redirect requests.

kaovilai: We can reuse DownloadRequest CR and instead of providing object storage url to an object, this API can retrieve the file from storage location then serve the file to the user.

kaovilai: So instead of downloadrequest CR status having URL to aws/gcp/azure, it will be url to velero ingress reachable by cli.

ywk253100: How about listing the pros and cons for both the aggregation and ingress approaches and making comparisons?

I think of some cons of the ingress approach:
- Need to handle auth by ourselves
- Introduce different styles of API: currently, all Velero APIs are k8s style; the ingress approach introduces different style of APIs
- Besides the ingress, we also need to support some other kinds of expose methods depend on the capabilities of the k8s: NodePort, ClusterIP, LoadBalancer. This introduces more configuration items when installing Velero. e.g. the ingress is different for different ingress controllers, so users need to set the ingress controller type
- Introduce some external dependencies: ingress needs the ingress controller ready and the LoadBalancer needs the load balancer provider ready

ywk253100: One big issue for the aggregation approach I can think of is how to support installing multiple Velero in one cluster.

kaovilai: The aggregation approach also requires [flags](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/#enable-kubernetes-apiserver-flags) to be enabled in the apiserver.

shawn-hurley: Just to be clear there is no reason to not follow normal k8s api structure for this, unless you want it to behave like a sub-resource like /logs or /exec even those follow k8s apis.

Doing auth yourself, but you would have had to add similar calls in the middleware of the aggregation api server because you want to check more SAR checks then just the basic ones for the resource being requested

You can still use RBAC for the resource like this.

With that said it does very much make it easier to run multiple veleros because you will have these servers tied to a particular install like you mentioned.

You should also note that many of the Aggregated API libraries and tools are alpha and have not had updates since 2022 AFAICT.

ywk253100: We could add the `namespace` to the `group` of `API Service` to support multiple Velero installations in one cluster. The API looks like `/apis/namespace01.velero.io/v1alpha1/...` for the Velero installed in the namespace `namespace01`.

We could choose the ingress approach, it can work absolutely. But my big concern is that current Velero exposes services based on the Kubernetes mechanism(CRD), Kubernetes handles the auth/TLS cert/etc. stuff. If we expose services via ingress/LB, we need to handle all of these by ourselves and two kinds of mechanisms exist for Velero.

kaovilai: Another idea came up which is to use kubectl cp to copy file off of a velero pod. This way, no ingress will be needed. Velero pod downloads the file to a path, and kubectl cp then velero pod deletes file.

draghuram: I am little bit confused by "auth" requirement for the download APIs. No other Velero APIs require any authentication/RBAC other than Kubernetes auth/RBAC. Why  would download APIs be any different? Just making sure I am not missing something.

kaovilai: > No other Velero APIs require any authentication/RBAC other than Kubernetes auth/RBAC.

All prior Velero APIs are implemeted as CustomResourceDefinitions. RBAC for these can be defined in [cluster]roles which define who can create what CRs where.

With API Aggregation layer, or ingress, I believe this falls outside of the general RBAC controls available to core k8s APIs and extensions via CustomResourceDefinitions.

See a section about [Extension Apiserver Authorizes the Request](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/#extension-apiserver-authorizes-the-request) which mentions the need for the apiserver to perform `SubjectAccessReview`. While an API Aggregation Layer might already require requests to contain API Group or GVK of sorts (I need to do more reading) which could be controlled by Role/ClusterRole, the Ingress method is very much open ended.

The Ingress method would require the client request to specifically include user/token information to be able to validate who they are and if they are allowed to do a certain action, and is not going to have GroupVersionKind that is definable in Role/ClusterRole. So velero ingress pod would have to explicitly does SubjectAccessReview and [TokenReview](https://dev-k8sref-io.web.app/docs/authentication/tokenreview-v1/) by ourselves.

kaovilai: Ingress solution without auth consideration is akin to having an ingress to nginx in the cluster serving a website without auth. Kube by default doesn't care/know who or enforce what's being requested from the website server exposed via Ingress.

shawn-hurley: Can we add this discussion to a community meeting, so that we can discuss this more syncrounously. 

@ywk253100 your concern regarding the need for auth/z is not an un-founded one, but one that I don't think Aggregated API's helps you with as much as you expect.

In both cases you will need to:
1. Set up an ingress/service
2. Create some certificate exchange. (for aggregated API, it is a custom CA, for your own service, it is a CA that the clients can trust, either using ingress re-encrypt from the cluster, or pass-through etc.)
3. Aggregated API does contain the middleware that does the TokenReview; you can still use this, though, by setting up this: https://github.com/kubernetes/apiserver/blob/master/pkg/authentication/authenticatorfactory/delegating.go#L68 in a middleware. (This is all that apiserver-runtime does)
4. Same basic process for setting up authorization, but that only focuses on the API that is being called. IIUC we need to make sure that the user has access to a different resource, therefore we would have to add a SAR check anyway even if we used apiserver-runtime. 

With the above said, creating the server in whatever way is orthogonal from making the Kubernetes API Server aware and proxy vs hitting the server directly. I hope this proves that regardless of mechanism you have to basically solve both cases.

ywk253100: > Can we add this discussion to a community meeting, so that we can discuss this more syncrounously.
Sure. Let's discuss this after the Chinese New Year holiday as some of us are going to take vacations from this week.

Firstly, let's clarify whether the service provided by the download server is API. If it is, I think the aggregation is better. 
Let me explain my point about this more clearly: I am not very concerned about the implementation details of the auth/cert etc. stuff, both aggregation and ingress approaches should work. What I'm concerned about is that if we choose the ingress approach, there will be two mechanisms for Velero APIs: one is via the kube API server with CRD, and the other one is via ingress. All these APIs are provided by Velero, but they are not unified, they are even exposed by two different servers. If the third parties want to integrate Velero, they need to connect to two servers to all the APIs.

If we don't think it is an API, I'm fine for both approaches.

draghuram: Just so that I am understanding this clearly, we are considering ingress/aggregate API approach (instead of CRDs) because we are talking about streaming data (similar to kubectl logs)?

Another approach I can think of is to create a new CRD based API to initiate the download. Velero can then create the necessary files inside the pod at some location and the user can copy them using "kubectl cp". This sounds a bit hackish but will completely bypass any new API mechanism.

ywk253100: > Just so that I am understanding this clearly, we are considering ingress/aggregate API approach (instead of CRDs) because we are talking about streaming data (similar to kubectl logs)?

Yes. Currently, Velero CLI gets a direct access URL of the object store and downloads the logs/backup tar/etc. from the object store directly, but the object store isn't always accessible by the CLI, so we want to change the way to do the download work

> Another approach I can think of is to create a new CRD based API to initiate the download. Velero can then create the necessary files inside the pod at some location and the user can copy them using "kubectl cp". This sounds a bit hackish but will completely bypass any new API mechanism.

This doesn't work because `kubectl cp` requires the `tar` binary to be present in the container while the Velero container doesn't contain it.

kaovilai: @mpryc also suggested `kubectl cp` to me awhile back.

> requires the tar binary to be present in the container

that sounds easy to fix.

Also if we consider kubectl cp, we get rid of the rbac implementation, as you probably can only run cp on containers you have access to. If you have access to velero container, you can already check for secrets written to credentialsFile inside velero pod.

kaovilai: It would only break apart if there's not enough disk space on the node to download volume backups.

draghuram: That is true though Velero can always accept a PVC for downloading logs, still keeping default as using node storage ("emptyDir").

kaovilai: For `kubectl cp` We can use [range in HTTP header in s3 request](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#AmazonS3-GetObject-request-header-Range:~:text=of%20an%20object.-,Range,-Downloads%20the%20specified) to break big files into chunks then reassemble client side.
> Downloads the specified byte range of an object.

sgalsaleh: Our product at Replicated uses the `github.com/vmware-tanzu/velero/pkg/cmd/util/downloadrequest` package to [download/stream backup logs](https://github.com/replicatedhq/kots/blob/30eeb4b70f6fa7c56aeb85bcb827e642facb5c9b/pkg/kotsadmsnapshot/download.go#L85-L106). It would be great to ensure backwards compatibility here in order to avoid breaking our customers' installations when they upgrade Velero.

shawn-hurley: I honestly don't see a difference between, exposing a pod to CP (user will need exec on the pod then right)? And exposing a service with/ actual RBAC checks for the backup, etc. I also think that keeping the old behavior, such that the provider has a way for the user to pull the data directly, would be a good thing.

design/velero-client-download-apiserver_design.md
kaovilai, line 1: FYI, got feedback from [velero-dev and ingress-nginx](https://kubernetes.slack.com/archives/C021GPR1L3S/p1711722779049059?thread_ts=1710295932.368219&cid=C021GPR1L3S) k8s slack.

From the feedback, will be consolidating to a service that is partially configurable (ie. NodePort/ClusterIP/LoadBalancer/etc.) with the ability for user to pass to velero the FQDN to use.

ywk253100: IMO, the ingress and service are the same essentially, that is users provide some configurations during the Velero installation and Velero makes it work end to end automatically. 

Both of them(ingress/load balancer service/node port service/in cluster service) should be supported if we choose this approach. So I would suggest that we merge approach2 and 3 into one.

And we can add another approach as we discussed offline that let the users to take care of how to expose the download server and we only provide an example guide like how to create the ingress/service to work together with Velero, this is much easier for our coding and maintainence work.

design/velero-client-download-apiserver_design.md
```diff
@@ -0,0 +1,105 @@
...
+We want them to work even when the object store is not accessible from the client side.
+Even if the object store is accessible from the client side, today things like cacert and insecure-skip-tls-verify flags are specified manually from the client even though the Velero server already has the information. This work eliminates the client from having to know about the object store configuration and certificates.
+## Goals
+- Enable Velero CLI users to download data from the Velero server without having to connect to the object store directly for scenarios where the object store is not accessible from the client side.
+
+## Non Goals
+- Making the download server the default way to download data from object store.
+
+## High-Level Design
+Create an endpoint on the Velero server that CLI users can connect to and download data from without having to connect to the object store directly.
```
ywk253100, line 31: Can we highlight that we will still keep the direct download way for most user cases and only download data via the download server for the BSLs that are configured to use it explicitly? We do this because the ingress/service approach involves some extra configuration items for users and we don't want to change the UX for most users.

design/velero-client-download-apiserver_design.md
```diff
@@ -0,0 +1,128 @@
...
+
+Velero will be updated to download data from the object store and stream it back to the client.
+
+The Velero CLI will still default to `downloadURL` from DownloadRequest but can fallback or forced to `veleroServerURL` which is served via the Velero server.
+
+If velero CLI attempted request from veleroServerURL at the specified FQDN and failed, and velero CLI command was executed inside velero server pod (`alias velero='kubectl exec -it deployment/velero -- velero'`), velero CLI will attempt to use the service name (cluster.local address) to connect to the download server.
+
+[![Velero Download Server](VeleroDownloadServer.jpg)](VeleroDownloadServer.jpg)
+
+## Detailed Design
```
ywk253100, line 52: Could you add a section about what the endpoint looks like?

design/velero-client-download-apiserver_design.md
```diff
@@ -0,0 +1,128 @@
...
+
+This approach may also [work on KinD clusters](https://kind.sigs.k8s.io/docs/user/loadbalancer/)
+
+The current DownloadRequest CR status `veleroServerURL` field will be populated to use the Download Server URL instead of the signed object store URL where available.
+
+Velero will be updated to download data from the object store and stream it back to the client.
+
+The Velero CLI will still default to `downloadURL` from DownloadRequest but can fallback or forced to `veleroServerURL` which is served via the Velero server.
+
+If velero CLI attempted request from veleroServerURL at the specified FQDN and failed, and velero CLI command was executed inside velero server pod (`alias velero='kubectl exec -it deployment/velero -- velero'`), velero CLI will attempt to use the service name (cluster.local address) to connect to the download server.
```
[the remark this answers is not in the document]
shawn-hurley: I am a little confused about when this is used. Is this just for documentation for a user to be able to access things if they don't have a velero CLI installed?

design/velero-client-download-apiserver_design.md
```diff
@@ -0,0 +1,128 @@
...
+	VeleroServerURL string `json:"veleroServerURL,omitempty"`
+
+	// Expiration is when this DownloadRequest expires and can be deleted by the system.
+	// +optional
+	// +nullable
+	Expiration *metav1.Time `json:"expiration,omitempty"`
+}
+```
+### HTTPS Server Certificates
+Velero server will generate a self-signed certificate for HTTPS server endpoint. The certificate will be stored in a secret in the same namespace as the Velero server. The secret will be named `velero-download-server-certs`.
```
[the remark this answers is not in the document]
shawn-hurley: I would also explore when an ingress/router can use a CA cert and doesn't want to re-encrypt to talk to the service.

shawn-hurley (commented):
I think this is looking really good! I am glad that we are making progress on this!

design/velero-client-download-apiserver_design.md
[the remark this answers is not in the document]
kaovilai: A little technique, that could have code dedicated to it. Since download api server is a fallback, we can move this out of scope or remove entirely.

design/velero-client-download-apiserver_design.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Non Goals
+- Making the download server the default way to download data from object store.
+
+## High-Level Design
+Create an endpoint on the Velero server that CLI users can connect to and download data from without requiring signed URL from object store directly when configured explicitly or as a fallback.
+
+### Velero Download Server URL generation using Service to Expose Download Server
+
+Upon enabling Download Server API, Velero will create a service of type that user specifies, which will expose the download server to the outside world. The download server will proxy requested data from the object store and stream it back to the client using one of the available service types (ClusterIP/NodePort/LoadBalancer/ExternalName). User can also precreate a service and pass the service name to velero during install time with `velero install --veleroServerServiceName=veleroDownloadServerServiceName` to use the precreated service to expose the download server.
```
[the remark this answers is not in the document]
ywk253100: I'm fine for both.
Let's document this in the design about how users can use ingress

design/velero-client-download-apiserver_design.md
```diff
@@ -0,0 +1,128 @@
...
+
+If the service can also be reached via a fully qualified domain name (FQDN), user can pass the FQDN to velero during install time with `velero install --veleroServerFQDN=dnsNameToVeleroServerService.com`. Velero Download Server will use FQDN in-place of values auto discovered from created service (ie. NodePort/Cluster IP)
+
+This approach may also [work on KinD clusters](https://kind.sigs.k8s.io/docs/user/loadbalancer/)
+
+The current DownloadRequest CR status `veleroServerURL` field will be populated to use the Download Server URL instead of the signed object store URL where available.
```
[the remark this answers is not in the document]
ywk253100: > IIUC @ywk253100, you are advocating that for some object stores, we would want to use the signed URL that is returned and not the download request when it is not needed?

No, What I'm saying is that we always read the URL from the `downloadURL` field of the download request rather than adding a new `veleroServerURL`.

> If that is the case I agree, I wonder how we will determine that though?

We can add a new field in the BSL spec to determine how to get the URL.

design/velero-client-download-apiserver_design.md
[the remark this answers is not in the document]
kaovilai: An example of where we told folks they could alias velero command so they don't need to actually install anything to their machine to use velero CLI functions.

https://access.redhat.com/articles/5456281#:~:text=alias%20velero%3D%27oc%20%2Dn%20openshift%2Dadp%20exec%20deployment/velero%20%2Dc%20velero%20%2Dit%20%2D%2D%20./velero%27

design/velero-client-download-apiserver_design.md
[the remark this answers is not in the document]
kaovilai: I like this idea.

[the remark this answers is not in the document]
kaovilai: > decides the correct way to download the data

Not sure if there's an easy way to determine from DR reconciler what cli has access to.. CLI can perhaps create another CR saying prior attempt was unsuccessful so the next DR url returned would be download server url.

kaovilai:
Will be researching more info around cacerts for this design.

design/velero-client-download-apiserver_design.md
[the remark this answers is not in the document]
ywk253100: If we add a new field (e.g. `downloadViaVeleroServer` bool) to the BSL spec, the DR reconciler can determine which URL (the URL of direct access or via the Velero server) should be returned. 
And the CLI doesn't need to care about this, it downloads the data from the `downloadURL` of DR.

ywk253100:
@kaovilai Do you think we should put the download server in a separate pod or in the same pod with Velero server? Could you also clarify this in the design?

kaovilai:
I think a separate pod would be overkill in terms of resource requirements. Most of the time there's no download request to serve.

design/velero-client-download-apiserver_design.md
```diff
@@ -0,0 +1,128 @@
...
+The same applies to changes in CRDs, YAML examples, and so on.
+
+Ideally the changes should be made in sequence so that the work required to implement this design can be done incrementally, possibly in parallel. -->
+
+<!-- What we do with the download url -->
+The DownloadRequest Status struct will be updated to contain VeleroServerURL, a url to the velero server instance that will retrieve the intended file via the velero server.
+
+The Velero CLI may use the VeleroServerURL to download data from the Velero server when DownloadURL is not populated, unreachable, or explicitly requested via `--veleroServerUrl`.
+
+When using VeleroServerURL, Velero CLI will send current kubeconfig to the download server via HTTP Authorization Header for velero server to authenticate and authorize the request.
```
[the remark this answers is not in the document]
kaovilai: > Velero CLI isn't the only client talks to the server

I was not aware. Can you clarify what else talks to the server in this way?

design/velero-client-download-apiserver_design.md
kaovilai: Yes. Will do. I think we can simply use the same path pattern as to how the files are stored today.. so something like `/backups/<backupName>/<contents>/...`

design/velero-client-download-apiserver_design.md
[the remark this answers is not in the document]
ywk253100: In our downstream product, the client runs inside the pod and uses CRs directly to talk to the server.

Currently, if users have permission to create the `DownloadReqeust` then they have permission to download data from the object storage, and a signed URL is returned. The permission checking happens actually against the `DownloadRequest`. The object storage only checks the request is signed and valid.  So we can follow the same pattern that in the download server side we only check the request is signed and valid.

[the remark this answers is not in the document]
kaovilai: Ok so just as long as DownloadRequests is able to get created, and the velero CLI have permission to read its status, then the signedUrl there should be usable without the kubeconfig mechanism proposed.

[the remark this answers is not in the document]
ywk253100: Yes.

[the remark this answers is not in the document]
shawn-hurley: This is scary to have URLs that contain the entire backup that could be world accessible. 

If we go down this path I think you have to have the link expire at some reasonable time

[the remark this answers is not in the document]
ywk253100: > If we go down this path I think you have to have the link expire at some reasonable time

Yes, we can follow the way how object storage does

design/velero-client-download-apiserver_design.md
kaovilai: After discussing with @weshayutin I think we can keep backwards compatibility without introducing auth headers requirement by essentially recreating the presigned URL behavior
> [You can use presigned URLs to grant time-limited access to objects](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html#:~:text=You%20can%20use%20presigned%20URLs%20to%20grant%20time%2Dlimited%20access%20to%20objects)

kaovilai: downloadUrl for each download request will contain uuid and expire shortly.

kaovilai: presigned would be optional, we should auth by default using subjectaccessreviews and tokenreview

design/velero-client-download-apiserver_design.md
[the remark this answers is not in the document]
kaovilai: Here's the usecase example of this alias: https://github.com/vmware-tanzu/velero/issues/7763#issuecomment-2930834421

PR #7366, opened 2024-01-29
Add restore warning for using restorePVs: true while restoring from FS Backup

shubham-pampattiwar:

Fixes #(issue)

### DIFF
--- a/pkg/restore/restore.go
+++ b/pkg/restore/restore.go
@@ -1258,6 +1258,10 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 			case internalVolume.PodVolumeBackup:
 				restoreLogger.Infof("Dynamically re-provisioning persistent volume because it has a pod volume backup to be restored.")
 				ctx.pvsToProvision.Insert(name)
+				// add restore warning if user is trying to restore from snapshots but the backup uses fs backup  as default option
+				if boolptr.IsSetToTrue(ctx.backup.Spec.DefaultVolumesToFsBackup) && boolptr.IsSetToTrue(ctx.restore.Spec.RestorePVs) {
+					warnings.Add(namespace, errors.Errorf("restorePVs should not be set as true when backup consists of defaultVolumesToFSbackup as true and all volumes are using FS Backup method. RestorePVs specifies whether to restore all included PVs from snapshot"))
+				}
 
 				// Return early because we don't want to restore the PV itself, we
 				// want to dynamically re-provision it.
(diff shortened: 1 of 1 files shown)


pkg/restore/restore.go
```diff
@@ -1258,6 +1258,10 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 			case internalVolume.PodVolumeBackup:
 				restoreLogger.Infof("Dynamically re-provisioning persistent volume because it has a pod volume backup to be restored.")
 				ctx.pvsToProvision.Insert(name)
+				// add restore warning if user is trying to restore from snapshots but the backup uses fs backup  as default option
+				if boolptr.IsSetToTrue(ctx.backup.Spec.DefaultVolumesToFsBackup) && boolptr.IsSetToTrue(ctx.restore.Spec.RestorePVs) {
+					warnings.Add(namespace, errors.Errorf("restorePVs should not be set as true when backup consists of defaultVolumesToFSbackup as true and all volumes are using FS Backup method. RestorePVs specifies whether to restore all included PVs from snapshot"))
```
draghuram, line 1263: I would suggest a minor change in the message as follows (or something similar):

"Ignoring restorePVs option since the backup method was file system backup. This flag  is only valid for snapshots."

BTW. Is this going to be logged once for each PV or is it a global message for entire restore?

pkg/restore/restore.go
```diff
@@ -1258,6 +1258,10 @@ func (ctx *restoreContext) restoreItem(obj *unstructured.Unstructured, groupReso
 			case internalVolume.PodVolumeBackup:
 				restoreLogger.Infof("Dynamically re-provisioning persistent volume because it has a pod volume backup to be restored.")
 				ctx.pvsToProvision.Insert(name)
+				// add restore warning if user is trying to restore from snapshots but the backup uses fs backup  as default option
+				if boolptr.IsSetToTrue(ctx.backup.Spec.DefaultVolumesToFsBackup) && boolptr.IsSetToTrue(ctx.restore.Spec.RestorePVs) {
```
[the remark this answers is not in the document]
sseago: Also, this flag being set does *not* guarantee that we're not restoring any from snapshot -- the  flag just determines default behavior -- individual volumes could be using fs-backup or snapshot, depending on opt-in/opt-out annotations.

pkg/restore/restore.go
sseago: I'm not sure we can say "ignoring" here just based on the default flag. This particular volume could be using snapshots even if that flag is set.

shubham-pampattiwar:
PR closed as issue is closed now https://github.com/vmware-tanzu/velero/issues/7365

Issue #7291, opened 2024-01-09
Repo enhancement - Microservice repo maintenance


Lyndon-Li:
Repo maintenance is a time consuming and resource consuming operation, we should move repo maintenance into dedicate pods (e.g., through jobs) to archive below benefits:
1. Mitigate Velero server OOM problem
2. Differencing resource configuration according to repo scale
3. Parallel repo maintenance

PR #7375, opened 2024-02-01
Add design for repository maintenance job

qiuming-best:

Fixes #(issue)
https://github.com/vmware-tanzu/velero/issues/7291

### DIFF
--- a/design/repository-maintenance.md
+++ b/design/repository-maintenance.md
@@ -1,0 +1,318 @@
+# Design for repository maintenance job
+
+## Abstract
+This design proposal aims to decouple repository maintenance from the Velero server by launching a maintenance job when needed, to mitigate the impact on the Velero server during backups.
+
+## Background
+During backups, Velero performs periodic maintenance on the repository. This operation may consume significant CPU and memory resources in some cases, leading to potential issues such as the Velero server being killed by OOM. This proposal addresses these challenges by separating repository maintenance from the Velero server.
+
+## Goals
+1. **Independent Repository Maintenance**: Decouple maintenance from Velero's main logic to reduce the impact on the Velero server pod.
+
+2. **Configurable Resources Usage**: Make the resources used by the maintenance job configurable.
+
+3. **No API Changes**: Retain existing APIs and workflow in the backup repository controller.
+
+## Non Goals
+We have lots of concerns over parallel maintenance, which will increase the complexity of our design currently.
+
+ - Non-blocking maintenance job: it may conflict with updating the same `backuprepositories` CR when parallel maintenance.
+
+ - Maintenance job concurrency control: there is no one suitable mechanism in Kubernetes to control the concurrency of different jobs.
+
+ - Parallel maintenance: Maintaining the same repo by multiple jobs at the same time would have some compatible cases that some providers may not support.
+
+Unfortunately, parallel maintenance is currently not a priority because of the concerns above, improving maintenance efficiency is not the primary focus at this stage.
+
+## High-Level Design
+1. **Add Maintenance Subcommand**: Introduce a new Velero server subcommand for repository maintenance.
+
+2. **Create Jobs by Repository Manager**: Modify the backup repository controller to create a maintenance job instead of directly calling the multiple chain calls for Kopia or Restic maintenance.
+
+3. **Update Maintenance Job Result in BackupRepository CR**: Retrieve the result of the maintenance job and update the status of the `BackupRepository` CR accordingly.
+
+4. **Add Setting for Maintenance Job**: Introduce a configuration option to set maintenance jobs, including resource limits (CPU and memory), keeping the latest N maintenance jobs for each repository.
+
+## Detailed Design
+
+### 1. Add Maintenance sub-command
+
+The CLI command will be added to the Velero CLI, the command is designed for use in a pod of maintenance jobs. 
+
+Our CLI command is designed as follows:
+```shell
+$ velero repo-maintenance --repo-name $repo-name --repo-type $repo-type --backup-storage-location $bsl
+```
+
+Compared with other CLI commands, the maintenance command is used in a pod of maintenance jobs not for user use, and the job should show the result of maintenance after finish.
+
+Here we will write the error message into one specific file which could be read by the maintenance job.
+
+on the whole, we record two kinds of logs:
+
+- one is the log output of the intermediate maintenance process: this log could be retrieved via the Kubernetes API server, including the error log.
+
+- one is the result of the command which could indicate whether the execution is an error or not: the result could be redirected to a file that the maintenance job itself could read, and the file only contains the error message.
+
+we will write the error message into the `/dev/termination-log` file if execution is failed.
+
+The main maintenance logic would be using the repository provider to do the maintenance.
+
+```golang
+func checkError(err error, file *os.File) {
+	if err != nil {
+		if err != context.Canceled {
+			if _, errWrite := file.WriteString(fmt.Sprintf("An error occurred: %v", err)); errWrite != nil {
+				fmt.Fprintf(os.Stderr, "Failed to write error to termination log file: %v\n", errWrite)
+			}
+			file.Close()
+			os.Exit(1) // indicate the command executed failed
+		}
+	}
+}
+
+func (o *Options) Run(f veleroCli.Factory) {
+	logger := logging.DefaultLogger(o.LogLevelFlag.Parse(), o.FormatFlag.Parse())
+	logger.SetOutput(os.Stdout)
+
+	errorFile, err := os.Create("/dev/termination-log")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to create termination log file: %v\n", err)
+		return
+	}
+	defer errorFile.Close()
+	...
+
+  err = o.runRepoPrune(cli, f.Namespace(), logger)
+	checkError(err, errorFile)
+	...
+}
+
+func (o *Options) runRepoPrune(cli client.Client, namespace string, logger logrus.FieldLogger) error {
+  ...
+	var repoProvider provider.Provider
+	if o.RepoType == velerov1api.BackupRepositoryTypeRestic {
+		repoProvider = provider.NewResticRepositoryProvider(credentialFileStore, filesystem.NewFileSystem(), logger)
+	} else {
+		repoProvider = provider.NewUnifiedRepoProvider(
+			credentials.CredentialGetter{
+				FromFile:   credentialFileStore,
+				FromSecret: credentialSecretStore,
+			}, o.RepoType, cli, logger)
+	}
+  ...
+
+	err = repoProvider.BoostRepoConnect(context.Background(), para)
+	if err != nil {
+		return errors.Wrap(err, "failed to boost repo connect")
+	}
+
+	err = repoProvider.PruneRepo(context.Background(), para)
+	if err != nil {
+		return errors.Wrap(err, "failed to prune repo")
+	}
+	return nil
+}
+```
+
+### 2. Create Jobs by Repository Manager
+Currently, the backup repository controller will call the repository manager to do the `PruneRepo`, and Kopia or Restic maintenance is then finally called through multiple chain calls.
+
+We will keep using the `PruneRepo` function in the repository manager, but we cut off the multiple chain calls by creating a maintenance job.
+
+The job definition would be like below:
+```yaml
+apiVersion: v1
+items:
+- apiVersion: batch/v1
+  kind: Job
+  metadata:
+    # labels or affinity or topology settings would inherit from the velero deployment
+    labels:
+      # label the job name for later list jobs by name
+      job-name: nginx-example-default-kopia-pqz6c
+    name: nginx-example-default-kopia-pqz6c
+    namespace: velero
+  spec:
+    # Not retry it again
+    backoffLimit: 1
+    # Only have one job one time
+    completions: 1
+    # Not parallel running job
+    parallelism: 1
+    template:
+      metadata:
+        labels:
+          job-name: nginx-example-default-kopia-pqz6c
+        name: kopia-maintenance-job
+      spec:
+        containers:
+        # arguments for repo maintenance job
+        - args:
+          - repo-maintenance
+          - --repo-name=nginx-example
+          - --repo-type=kopia
+          - --backup-storage-location=default
+	  # inherit from Velero server
+          - --log-level=debug
+          command:
+          - /velero
+          # inherit environment variables from the velero deployment
+          env:
+          - name: AZURE_CREDENTIALS_FILE
+            value: /credentials/cloud
+          # inherit image from the velero deployment
+          image: velero/velero:main
+          imagePullPolicy: IfNotPresent
+          name: kopia-maintenance-container
+	  # resource limitation set by Velero server configuration
+	  # if not specified, it would apply best effort resources allocation strategy
+          resources: {}
+          # error message would be written to /dev/termination-log
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: File
+          # inherit volume mounts from the velero deployment
+          volumeMounts:
+          - mountPath: /credentials
+            name: cloud-credentials
+        dnsPolicy: ClusterFirst
+        restartPolicy: Never
+        schedulerName: default-scheduler
+        securityContext: {}
+        # inherit service account from the velero deployment
+        serviceAccount: velero
+        serviceAccountName: velero
+        volumes:
+        # inherit cloud credentials from the velero deployment
+        - name: cloud-credentials
+          secret:
+            defaultMode: 420
+            secretName: cloud-credentials
+	# ttlSecondsAfterFinished set the job expired seconds
+	ttlSecondsAfterFinished: 86400
+  status:
+	# which contains the result after maintenance
+	message: ""
+	lastMaintenanceTime: ""
+```
+
+Now, the backup repository controller will call the repository manager to create one maintenance job and wait for the job to complete. The Kopia or Restic maintenance multiple chains are called by the job.
+
+### 3. Update the Result of the Maintenance Job into BackupRepository CR
+
+The backup repository controller will update the result of the maintenance job into the backup repository CR.
+
+For how to get the result of the maintenance job we could refer to [here](https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/#writing-and-reading-a-termination-message).
+
+After the maintenance job is finished, we could get the result of maintenance by getting the terminated message from the related pod:
+
+```golang
+func GetContainerTerminatedMessage(pod *v1.Pod) string {
+	...
+	for _, containerStatus := range pod.Status.ContainerStatuses {
+		if containerStatus.LastTerminationState.Terminated != nil {
+			return containerStatus.LastTerminationState.Terminated.Message
+		}
+	}
+	...
+	return ""
+}
+```
+Then we could update the status of backupRepository CR with the message.
+
+### 4. Add Setting for Resource Usage of Maintenance
+Add one configuration for setting the resource limit of maintenance jobs as below:
+```shell
+    velero server --maintenance-job-cpu-request $cpu-request --maintenance-job-mem-request $mem-request --maintenance-job-cpu-limit $cpu-limit --maintenance-job-mem-limit $mem-limit
+```
+Our default value is 0, which means we don't limit the resources, and the resource allocation strategy would be [best effort](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#besteffort).
+
+### 5. Automatic Cleanup for Finished Maintenance Jobs
+Add configuration for clean up maintenance jobs:
+
+- keep-latest-maintenance-jobs: the number of keeping latest maintenance jobs for each repository.
+
+	```shell
+		velero server --keep-latest-maintenance-jobs $num
+	```
+
+We would check and keep the latest N jobs after a new job is finished.
+```golang
+func deleteOldMaintenanceJobs(cli client.Client, repo string, keep int) error {
+	// Get the maintenance job list by label
+	jobList := &batchv1.JobList{}
+	err := cli.List(context.TODO(), jobList, client.MatchingLabels(map[string]string{RepositoryNameLabel: repo}))
+	if err != nil {
+		return err
+	}
+
+	// Delete old maintenance jobs
+	if len(jobList.Items) > keep {
+		sort.Slice(jobList.Items, func(i, j int) bool {
+			return jobList.Items[i].CreationTimestamp.Before(&jobList.Items[j].CreationTimestamp)
+		})
+		for i := 0; i < len(jobList.Items)-keep; i++ {
+			err = cli.Delete(context.TODO(), &jobList.Items[i], client.PropagationPolicy(metav1.DeletePropagationBackground))
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+```
+
+### 6 Velero Install with Maintenance Options
+All the above maintenance options should be supported by Velero install command.
+
+### 7. Observability and Debuggability
+Some monitoring metrics are added for backup repository maintenance:
+- repo_maintenance_total
+- repo_maintenance_success_total
+- repo_maintenance_failed_total
+- repo_maintenance_duration_seconds
+
+We will keep the latest N maintenance jobs for each repo, and users can get the log from the job. the job log level inherent from the Velero server setting.
+
+Also, we would integrate maintenance job logs and `backuprepositories` CRs into `velero debug`.
+
+Roughly, the process is as follows:
+1. The backup repository controller will check the BackupRepository request in the queue periodically.
+
+2. If the maintenance period of the repository checked by `runMaintenanceIfDue` in `Reconcile` is due, then the backup repository controller will call the Repository manager to execute `PruneRepo`
+
+3. The `PruneRepo` of the Repository manager will create one maintenance job, the resource limitation, environment variables, service account, images, etc. would inherit from the Velero server pod. Also, one clean up TTL would be set to maintenance job.
+
+4. The maintenance job will execute the Velero maintenance command, wait for maintaining to finish and write the maintenance result into the terminationMessagePath file of the related pod.
+
+5. Kubernetes could show the result in the status of the pod by reading the termination message in the pod.
+
+6. The backup repository controller will wait for the maintenance job to finish and read the status of the maintenance job, then update the message field and phase in the status of `backuprepositories` CR accordingly.
+
+6. Clean up old maintenance jobs and keep only N latest for each repository.
+
+### 8. Codes Refinement
+Once `backuprepositories` CR status is modified, the CR would re-queue to be reconciled, and re-execute logics in reconcile shortly not respecting the re-queue frequency configured by `repoSyncPeriod`.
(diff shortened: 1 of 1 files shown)


design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+			return containerStatus.LastTerminationState.Terminated.Message
+		}
+	}
+	...
+	return ""
+}
+```
+Then we could update the status of backupRepository CR with the message.
+
+### 4. Add Setting for Maintenance Job Timeout
```
qiuming-best, line 155: Not sure we need timeout configuration If the maintenance job timeout, the controller deletes the job which may lead to maintenance breaking halfway.

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+
+
+## Detailed Design
+
+### 1. Add Maintenance CLI Command
+
+The CLI command will be added to the Velero CLI.
+
+```shell
+$ velero maintenance --repo-name $repo-name --repo-type $repo-type --backup-storage-location $bsl
```
Lyndon-Li, line 46: velero maintenance -> velero server repo-maintain?
Let's make it as a velero server sub command, so as to indicate users that they should not call this CLI from client

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+	log.Debugf("Creating maintenance job: %v", job)
+	defer cancel()
+	if err := m.client.Create(pruneTimeOut, job); err != nil {
+		return errors.Wrap(err, "error to create maintenance job")
+	}
+
+	if err := builder.WaitForJobComplete(pruneTimeOut, m.client, job); err != nil {
+		return errors.Wrap(err, "error to wait for maintenance job complete")
+	}
+	...
```
Lyndon-Li, line 126: Let's get the job result from here, since the job is launched by repoManager, the repoManager should get the result from the job, as the controller doesn't know anything about the job.

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+
+### 2. Create Jobs by Repository Manager
+Currently, the backup repository controller will call the repository manager to do the `PruneRepo`, and Kopia or Restic maintenance is then finally called through multiple chain calls.
+
+We will keep using the `PruneRepo` function in the repository manager, but we cut off the multiple chain calls by creating a maintenance job.
+
+```golang
+func (m *manager) PruneRepo(repo *velerov1api.BackupRepository) error {
+	...
+	// writing the maintenance result into default terminationMessagePath
```
Lyndon-Li, line 104: Add some consts for strings used in this function?

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+ - Non-blocking maintenance job: it may conflict with updating the same `backuprepositories` CR when parallel maintenance.
+
+ - Maintenance job concurrency control: there is no one suitable mechanism in Kubernetes to control the concurrency of different jobs.
+
+ - Parallel maintenance: Maintaining the same repo by multiple jobs at the same time would have some compatible cases that some providers may not support.
+
+Unfortunately, parallel maintenance is currently not a priority because of the concerns above, improving maintenance efficiency is not the primary focus at this stage.
+
+## High-Level Design
+1. **Add Maintenance CLI Command**: Introduce a new Velero CLI command for repository maintenance.
```
ywk253100, line 28: Let's change it to `Add Maintenance sub Command`. `CLI command` confuses me that seems it runs on client side

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+
+### 5. Add Setting for Resource Usage of Maintenance
+Add one configuration for setting the resource limit of maintenance jobs as below:
+```shell
+    velero server --maintenance-job-cpu-request $cpu-request --maintenance-job-mem-request $mem-request --maintenance-job-cpu-limit $cpu-limit --maintenance-job-mem-limit $mem-limit
+```
+Our default value is 0, which means we don't limit the resources.
+
+### 6. Clear up Maintenance Jobs
+Add one configuration for the number of keeping latest failure maintenance jobs as below:
```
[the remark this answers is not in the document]
ywk253100: Will any succeeded job be kept?

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+
+We will keep using the `PruneRepo` function in the repository manager, but we cut off the multiple chain calls by creating a maintenance job.
+
+```golang
+func (m *manager) PruneRepo(repo *velerov1api.BackupRepository) error {
+	...
+	// writing the maintenance result into default terminationMessagePath
+	args := []string {
+	"-c",
+	"velero maintenance --repo-name=" + param.BackupRepo.Spec.VolumeNamespace + " --repo-type=" + param.BackupRepo.Spec.RepositoryType + " --backup-storage-location=" + param.BackupLocation.Name + " > /dev/termination-log",
```
ywk253100, line 107: We should only put the final error message to file `/dev/termination-log` and put the msg in the repo status then.
The whole logs of the maintenance job can be got via the job logs.

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	env := &v1.EnvVar{Name: ProviderFile, Value: "/credentials/cloud"}
+	job := builder.ForJob(param.BackupRepo.Name, param.BackupRepo.Namespace, "velero/velero:maintenance", args).Env(env).
+		VolumeMounts(*builder.ForVolumeMount("cloud-credentials", "/credentials").Result()).
+		ServiceAccount("velero").Resources(&resources).
+		Volumes(*builder.ForVolume("cloud-credentials").SecretSource("cloud-credentials").Result()).Result()
```
ywk253100, line 114: One thing needs to note is that some attributes of the job should come from the Velero pod

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+### 6. Clear up Maintenance Jobs
+Add one configuration for the number of keeping latest failure maintenance jobs as below:
+```shell
+    velero server --keep-failed-maintenance-jobs $num
+```
+
+We would clean all completed jobs and keep the latest N failed jobs at the end of each round of reconciliation.
+
+```golang
+func clearFinishedJob(ctx context.Context, c client.Client, retainFailedJobs int) error {
```
ywk253100, line 178: When will the clean-up action be triggered?

design/repository-maintenance.md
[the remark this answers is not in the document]
qiuming-best: Currently, it will delete all succeeded jobs. why should we keep some succeeded jobs? @ywk253100

design/repository-maintenance.md
qiuming-best: yes, you are right

design/repository-maintenance.md
qiuming-best: it will triggered at the end of Reconcile, will it be so frequently?

design/repository-maintenance.md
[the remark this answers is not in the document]
ywk253100: If the first run of the job fails but the second run succeeds, will users need to care about the failed one? Seems not? So we only need to keep the last job?
BTW, the maintenance job is per repository, right? So we need to keep the last x jobs for every repository?

design/repository-maintenance.md
ywk253100: If it is triggered at the end of reconciling, does that mean any change of the repository also triggers the clean-up action? Seems this is unnecessary. 
How about triggering the clean-up action to clean up the corresponding repository's jobs when one job completes?

qiuming-best: ok, I think it's more reasonable

design/repository-maintenance.md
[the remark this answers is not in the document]
qiuming-best: yes, your suggestion sounds good

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+
+The CLI command will be added to the Velero CLI.
+
+```shell
+$ velero maintenance --repo-name $repo-name --repo-type $repo-type --backup-storage-location $bsl
+```
+
+The main maintenance logic would be using the repository provider to do the maintenance.
+```golang
+func (o *Options) Run(f veleroCli.Factory) {
```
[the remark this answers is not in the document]
qiuming-best: I want the log level inherited from the velero server.

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+	...
+	credentialFileStore, err := credentials.NewNamespacedFileStore(
+		cli,
+		f.Namespace(),
+		"/tmp/credentials",
+		filesystem.NewFileSystem(),
+	)
+	cmd.CheckError(err)
```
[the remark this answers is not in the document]
qiuming-best: there are two kinds of messages, one is the `fmt.Fprintf` type of message, which would be redirected to termination-log, and another kind of message would be log could be retrieved by job log

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+	err = repoProvider.BoostRepoConnect(context.Background(), para)
+	cmd.CheckError(err)
+	// Prune repo
+	err = repoProvider.PruneRepo(context.Background(), para)
+	cmd.CheckError(err)
+	...
+}
+```
+
+### 2. Create Jobs by Repository Manager
```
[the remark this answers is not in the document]
qiuming-best: sounds good

design/repository-maintenance.md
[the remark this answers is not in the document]
qiuming-best: ok, I'll keep the latest N jobs for each repo

design/repository-maintenance.md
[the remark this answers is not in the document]
qiuming-best: I've added the `FileHook` section in the design, which could dump errors into termination-log

design/repository-maintenance.md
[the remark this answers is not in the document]
qiuming-best: I've added instructions that this command only could by used internally.

If client calls it, it would fail

qiuming-best: I've changed into velero server repo-maintenance, the word maintenance which consistent with Kopia maintenance Command

design/repository-maintenance.md
qiuming-best: modified

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+## High-Level Design
+1. **Add Maintenance CLI Command**: Introduce a new Velero CLI command for repository maintenance.
+
+2. **Create Jobs by Repository Manager**: Modify the backup repository controller to create a maintenance job instead of directly calling the multiple chain calls for Kopia or Restic maintenance.
+
+3. **Update Maintenance Job Result in BackupRepository CR**: Retrieve the result of the maintenance job and update the status of the `BackupRepository` CR accordingly.
+
+4. **Add Setting for Maintenance Job**: Introduce a configuration option to set maintenance jobs, including resource limits (CPU and memory), keeping the latest N failed maintenance jobs.
+
+4. **Clear Up Maintenance Jobs**: Introduce a configuration to determine the number of failed maintenance jobs to retain for debugging purposes.
```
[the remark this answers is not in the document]
qiuming-best: ok, now it changes to keep the latest N maintenance jobs for each repository

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+
```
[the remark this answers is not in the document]
qiuming-best: @allenxu404 will help to add repo maintenance metrics later

design/repository-maintenance.md
```diff
@@ -0,0 +1,236 @@
...
+func (m *manager) PruneRepo(repo *velerov1api.BackupRepository) error {
+	...
+	// writing the maintenance result into default terminationMessagePath
+	args := []string {
+	"-c",
+	"velero maintenance --repo-name=" + param.BackupRepo.Spec.VolumeNamespace + " --repo-type=" + param.BackupRepo.Spec.RepositoryType + " --backup-storage-location=" + param.BackupLocation.Name + " > /dev/termination-log",
+	}
+
+	env := &v1.EnvVar{Name: ProviderFile, Value: "/credentials/cloud"}
+	job := builder.ForJob(param.BackupRepo.Name, param.BackupRepo.Namespace, "velero/velero:maintenance", args).Env(env).
```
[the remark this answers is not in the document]
qiuming-best: ok, I'm not using the builder package, now I've write some utility functions instead

design/repository-maintenance.md
```diff
@@ -0,0 +1,338 @@
...
+      spec:
+        containers:
+        # arguments for repo maintenance job
+        - args:
+          - server
+          - repo-maintenance
+          - --repo-name=nginx-example
+          - --repo-type=kopia
+          - --backup-storage-location=default
+          - --keep-latest-maintenance-jobs=3
```
ywk253100, line 194: With `--keep-latest-maintenance-jobs`, is repo maintenance job responsible to clean up the job history?

design/repository-maintenance.md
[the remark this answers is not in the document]
qiuming-best: Not sure we need the debug bundle, for its backup or restore related. but maintenance is independent of backup or restore.

Users may collect the maintenance job logs or we extend `velero debug` with a new maintenance flag to collect job logs?

design/repository-maintenance.md
```diff
@@ -0,0 +1,338 @@
...
+	...
+	return ""
+}
+```
+Then we could update the status of backupRepository CR with the message.
+
+### 4. Add Setting for Resource Usage of Maintenance
+Add one configuration for setting the resource limit of maintenance jobs as below:
+```shell
+    velero server --maintenance-job-cpu-request $cpu-request --maintenance-job-mem-request $mem-request --maintenance-job-cpu-limit $cpu-limit --maintenance-job-mem-limit $mem-limit
```
ywk253100, line 261: Will we support setting these options via `velero install`?

design/repository-maintenance.md
```diff
@@ -0,0 +1,338 @@
...
+```
+Our default value is 0, which means we don't limit the resources.
+
+### 5. Clear up Maintenance Jobs
+Add configuration for clear up maintenance jobs:
+	
+- keep-latest-maintenance-jobs: the number of keeping latest maintenance jobs
+
+	```shell
+		velero server --keep-latest-maintenance-jobs $num
```
ywk253100, line 271: Will we support setting this option via `velero install`?

design/repository-maintenance.md
```diff
@@ -0,0 +1,336 @@
...
+	```
+
+We would check and keep the latest N jobs after a new job is finished.
+
+```golang
+// deleteOldMaintenanceJobs deletes old maintenance jobs and keeps the latest N jobs for each repository
+func deleteOldMaintenanceJobs(cli client.Client, repo string, keep int) error {
```
ywk253100, line 278: Kubernetes supports cleaning up jobs automatically with setting `.spec.ttlSecondsAfterFinished`, is it easier for us to leverage this capability? e.g. `.spec.ttlSecondsAfterFinished=n*maintenance-interval`

design/repository-maintenance.md
qiuming-best: It's the `repository manager` response for maintenance jobs including creating, cleaning up jobs.
it's not cleaned by repo maintenance job

design/repository-maintenance.md
qiuming-best: sounds good, I'll add the support

design/repository-maintenance.md
qiuming-best: sounds good, I'll add the support

design/repository-maintenance.md
qiuming-best: it's good way if we choose TTL.

design/repository-maintenance.md
```diff
@@ -0,0 +1,338 @@
...
+
+on the whole, we record two kinds of logs:
+
+- one is the log output of the intermediate maintenance process: this log could be retrieved via the Kubernetes API server, including the error log.
+
+- one is the result of the command which could indicate whether the execution is an error or not: the result could be redirected to a file that the maintenance job itself could read, and the file only contains the error message.
+
+If we want an error message to be recorded into a file, we could use the [hook of logrus](https://pkg.go.dev/github.com/sirupsen/logrus@v1.9.3/hooks/writer#section-readme).
+
+below we defined the `FileHook` struct which implements the `logrus.Hook` interface.
```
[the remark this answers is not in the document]
qiuming-best: it will record all errors into one file, whether stopping the process is determined by our error handle logic.

design/repository-maintenance.md
```diff
@@ -0,0 +1,338 @@
...
+
+The main maintenance logic would be using the repository provider to do the maintenance.
+
+```golang
+func checkError(err error, log logrus.FieldLogger) {
+	if err != nil {
+		if err != context.Canceled {
+			log.Error("An error occurred: %v\n", err)
+		}
+		os.Exit(1)
```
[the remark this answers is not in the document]
qiuming-best: I've changed the check error logic

design/repository-maintenance.md
```diff
@@ -0,0 +1,363 @@
...
+}
+```
+
+### 2. Create Jobs by Repository Manager
+Currently, the backup repository controller will call the repository manager to do the `PruneRepo`, and Kopia or Restic maintenance is then finally called through multiple chain calls.
+
+We will keep using the `PruneRepo` function in the repository manager, but we cut off the multiple chain calls by creating a maintenance job.
+
+The job definition would be like below:
+```yaml
```
reasonerjt, line 175: Could you remove the generated fields and keep the relative information, rather than pasting the output of `kubectl get jobs`....

design/repository-maintenance.md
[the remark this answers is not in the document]
reasonerjt: If the `one file` means `/dev/termination-log` it should contain only the error message that causing the termination of the job.

design/repository-maintenance.md
```diff
@@ -0,0 +1,363 @@
...
+          - /velero
+          # inherit environment variables from the velero deployment
+          env:
+          - name: AZURE_CREDENTIALS_FILE
+            value: /credentials/cloud
+          # inherit image from the velero deployment
+          image: velero/velero:main
+          imagePullPolicy: IfNotPresent
+          name: kopia-maintenance-container
+          resources: {}
```
reasonerjt, line 230: If we deliberately choose not to set any resource limitation here, could you add a comment to explain?
i.e. if no resource limitation is set via velero server, velero will apply the "best effort" approach.

design/repository-maintenance.md
reasonerjt: I think it makes sense.

design/repository-maintenance.md
```diff
@@ -0,0 +1,363 @@
...
+Roughly, the process is as follows:
+1. The backup repository controller will check the BackupRepository request in the queue periodically.
+
+2. If the maintenance period of the repository checked by `runMaintenanceIfDue` in `Reconcile` is due, then the backup repository controller will call the Repository manager to execute `PruneRepo`
+
+3. The `PruneRepo` of the Repository manager will create one maintenance job, the resource usage would follow the setting in the Velero server, and the environment variable, service account, images, etc. would follow the Velero server pod.
+
+4. The maintenance job will execute the Velero maintenance command and write the result into the terminationMessagePath of the related pod.
+
+5. The backup repository controller will wait for the jobs to update the message field and phase in the status of `backuprepositories` CR after finishing maintenance by reading the job terminal message accordingly.
```
reasonerjt, line 348: By reading this item it looks like it's the jobs updating the `backuprepositories` CR, which I believe is not the case.
Could you rephrase to clarify it's the repo controller making the update to the `backuprepositories` CR by reading status of the job.

design/repository-maintenance.md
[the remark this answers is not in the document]
Lyndon-Li: Looks like the sample code mixes below two mechanisms:

- The error message - It should be in `/dev/termination-log `and should include only the errors causing the maintenance fail (in another word, if the maintenance succeeds, `/dev/termination-log ` should be empty)
- The debug log - It is the debug/info/warning generated by this pod and by Kopia and should be there for each job. Particularly, for each maintenance, Kopia generates some internal logs and Velero has the ability to stream the logs (at present, it is printed in Velero server log). So we should generate a general log files like Velero server and node-agent and users should be able to collect the logs by running `velero debug`

design/repository-maintenance.md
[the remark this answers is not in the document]
Lyndon-Li: See this comment https://github.com/vmware-tanzu/velero/pull/7375/files#r1501981162

design/repository-maintenance.md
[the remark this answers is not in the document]
qiuming-best: yes, It only contains error message

design/repository-maintenance.md
qiuming-best: I've removed some redundant fields and added some annotations for some fields.

design/repository-maintenance.md
qiuming-best: Add some comments for resources field

design/repository-maintenance.md
qiuming-best: rephrased

design/repository-maintenance.md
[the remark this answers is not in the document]
ywk253100: Do we still need the log hook mechanism? Seems doesn't

design/repository-maintenance.md
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		if err != context.Canceled {
+			log.Errorf("An error occurred: %v", err)
+			return err
+		}
+	}
+	return nil
+}
+
+func (o *Options) Run(f veleroCli.Factory) {
```
ywk253100, line 91: How about removing the implementation detail code and replace with the processing logic if needed? I can see several errors in the code and let's review them in the implementation PR.

qiuming-best: ok, I've removed some details but kept the main logic in it.

design/repository-maintenance.md
```diff
@@ -0,0 +1,284 @@
...
+	# which contains the result after maintenance
+	message: ""
+	lastMaintenanceTime: ""
+```
+
+Now, the backup repository controller will call the repository manager to create one maintenance job and wait for the job to complete. The Kopia or Restic maintenance multiple chains are called by the job.
+
+### 3. Update the Result of the Maintenance Job into BackupRepository CR
+
+The backup repository controller will update the result of the maintenance job into the backup repository CR.
```
shubham-pampattiwar, line 204: Are we planning on polling on the job status till it gets completed ? like every 5 sec or something ?

design/repository-maintenance.md
```diff
@@ -0,0 +1,284 @@
...
+
+- keep-latest-maintenance-jobs: the number of keeping latest maintenance jobs for each repository.
+
+	```shell
+		velero server --keep-latest-maintenance-jobs $num
+	```
+
+We would specify the `.spec.ttlSecondsAfterFinished` field of a maintenance job by value`Num * MaintenanceFrequency`.
+
+The TTL-after-finished controller assumes that a Job is eligible to be cleaned up TTL seconds after the maintenance job has finished. Automatic clean up please refer to [it](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/).
```
shubham-pampattiwar, line 242: +1 on the usage of `ttlafterfinished`

design/repository-maintenance.md
qiuming-best: Currently, the design is a `backup repository controller` that will trigger to create one job and block waiting for it to complete. The polling frequency would be 1 sec.

Keeping the same logic as the previous `backup repository controller` that will trigger to execute one internal maintenance operation and block wait until it finishes.

If using one non-block way, that controller would update one 'backup repositories' cr at the same time making our current work more complicated.

design/repository-maintenance.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Detailed Design
+
+### 1. Add Maintenance sub-command
+
+The CLI command will be added to the Velero CLI, the command is designed for use in a pod of maintenance jobs. 
+
+Our CLI command is designed as follows:
+```shell
+$ velero repo-maintenance --repo-name $repo-name --repo-type $repo-type --backup-storage-location $bsl
```
draghuram, line 44: Does this command trigger maintenance job every time the command is run? If so, wouldn't it be better to let users set a schedule for maintenance jobs? I hope I haven't misunderstood the design.

qiuming-best: this command calls the Kopia library or Restic to do maintenance and is used in a maintenance job, the command also reads some other configurations from the job such as environment variables and secret, etc. So using this command is less efficient than using Kopia or Restic to handle the repository directly for users.

design/repository-maintenance.md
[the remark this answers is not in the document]
qiuming-best: Yes, It seems we do not need to use a log hook mechanism.

design/repository-maintenance.md
qiuming-best: We don't use the TTL finally

design/repository-maintenance.md
reasonerjt: The command will be run in a k8s job and the job will be created by velero pod.
We do not expect user to run this command from the client side.

design/repository-maintenance.md
```diff
@@ -0,0 +1,318 @@
...
+      job-name: nginx-example-default-kopia-pqz6c
+    name: nginx-example-default-kopia-pqz6c
+    namespace: velero
+  spec:
+    # Not retry it again
+    backoffLimit: 1
+    # Only have one job one time
+    completions: 1
+    # Not parallel running job
+    parallelism: 1
```
kaovilai, line 142: @shawn-hurley have warned me in the past about using Jobs (from https://github.com/vmware-tanzu/velero/issues/2601 discussion) that this process has to be able to tolerate being started twice.

> Note that even if you specify .spec.parallelism = 1 and .spec.completions = 1 and .spec.template.spec.restartPolicy = "Never", the same program may sometimes be started twice.

[src](https://kubernetes.io/docs/concepts/workloads/controllers/job/#handling-pod-and-container-failures)

Keep that in mind, I haven't reviewed if repo maintenance starting twice "accidentally" would cause an issue or not.

kaovilai: I assume the starting twice case would happen very close to each other when the jobs object status haven't been updated to Running yet.

kaovilai: My napkin algorithm would be to have some kind of leader election even if the expected running job is one. Such that if there is more than one, the later one is no-op or stopped.

qiuming-best: Thanks for reminding me. If the job started twice "accidentally" for the maintenance scenario, it may have a bug but will not cause issues.

As the maintenance job updates the `LastMaintenanceTime` in `backuprepositories` CR only affects the next maintenance time which doesn't decide whether the repository is ready or not.  it may lead to multiple maintenances in one maintenance frequency in this starting twice "accidentally" scenario.

qiuming-best: Also for the same repository,  only the maintenance job which gets the repository file lock could run maintenance, and it's guaranteed by the repository itself that only one maintenance at one time for each repository

blackpiglet (approved):

Lyndon-Li (approved):

PR #7383, opened 2024-02-05
Design for data mover node selection

Lyndon-Li:
Add the design for node selection for data mover backup

### DIFF
--- a/design/node-agent-affinity.md
+++ b/design/node-agent-affinity.md
@@ -1,0 +1,137 @@
+# Node-agent Load Affinity Design
+
+## Glossary & Abbreviation
+
+**Velero Generic Data Path (VGDP)**: VGDP is the collective modules that is introduced in [Unified Repository design][1]. Velero uses these modules to finish data transfer for various purposes (i.e., PodVolume backup/restore, Volume Snapshot Data Movement). VGDP modules include uploaders and the backup repository.  
+
+**Exposer**: Exposer is a module that is introduced in [Volume Snapshot Data Movement Design][2]. Velero uses this module to expose the volume snapshots to Velero node-agent pods or node-agent associated pods so as to complete the data movement from the snapshots.    
+
+## Background
+
+Velero node-agent is a daemonset hosting controllers and VGDP modules to complete the concrete work of backups/restores, i.e., PodVolume backup/restore, Volume Snapshot Data Movement backup/restore.  
+Specifically, node-agent runs DataUpload controllers to watch DataUpload CRs for Volume Snapshot Data Movement backups, so there is one controller instance in each node. One controller instance takes a DataUpload CR and then launches a VGDP instance, which initializes a uploader instance and the backup repository connection, to finish the data transfer. The VGDP instance runs inside a node-agent pod or in a pod associated to the node-agent pod in the same node.  
+
+Varying from the data size, data complexity, resource availability, VGDP may take a long time and remarkable resources (CPU, memory, network bandwidth, etc.).  
+Technically, VGDP instances are able to run in any node that allows pod schedule. On the other hand, users may want to constrain the nodes where VGDP instances run for various reasons, below are some examples:  
+- Prevent VGDP instances from running in specific nodes because users have more critical workloads in the nodes  
+- Constrain VGDP instances to run in specific nodes because these nodes have more resources than others  
+- Constrain VGDP instances to run in specific nodes because the storage allows volume/snapshot provisions in these nodes only  
+
+Therefore, in order to improve the compatibility, it is worthy to configure the affinity of VGDP to nodes, especially for backups for which VGDP instances run frequently and centrally.  
+
+## Goals
+
+- Define the behaviors of node affinity of VGDP instances in node-agent for volume snapshot data movement backups  
+- Create a mechanism for users to specify the node affinity of VGDP instances for volume snapshot data movement backups  
+
+## Non-Goals
+- It is also beneficial to support VGDP instances affinity for PodVolume backup/restore, however, it is not possible since VGDP instances for PodVolume backup/restore should always run in the node where the source/target pods are created.  
+- It is also beneficial to support VGDP instances affinity for data movement restores, however, it is not possible in some cases. For example, when the `volumeBindingMode` in the storageclass is `WaitForFirstConsumer`, the restore volume must be mounted in the node where the target pod is scheduled, so the VGDP instance must run in the same node. On the other hand, considering the fact that restores may not frequently and centrally run, we will not support data movement restores.  
+- As elaberated in the [Volume Snapshot Data Movement Design][2], the Exposer may take different ways to expose snapshots, i.e., through backup pods (this is the only way supported at present). The implementation section below only considers this approach currently, if a new expose method is introduced in future, the definition of the affinity configurations and behaviors should still work, but we may need a new implementation.  
+
+## Solution
+
+We will use the ```node-agent-config``` configMap to host the node affinity configurations.
+This configMap is not created by Velero, users should create it manually on demand. The configMap should be in the same namespace where Velero is installed. If multiple Velero instances are installed in different namespaces, there should be one configMap in each namespace which applies to node-agent in that namespace only.  
+Node-agent server checks these configurations at startup time and use it to initiate the related VGDP modules. Therefore, users could edit this configMap any time, but in order to make the changes effective, node-agent server needs to be restarted.  
+Inside ```node-agent-config``` configMap we will add one new kind of configuration as the data in the configMap, the name is ```loadAffinity```.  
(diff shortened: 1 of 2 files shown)


design/node-agent-affinity.md
```diff
@@ -0,0 +1,124 @@
...
+                "matchExpressions": {
+                    "key": "kubernetes.io/hostname",
+                    "values": [
+                        "node1",
+                        "node2",
+                        "node3"
+                    ],
+                    "operator": "In"
+                },           
+                "matchExpressions": {
```
ywk253100, line 82: According to https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1233, seems the different expressions should be inside one `matchExpressions` array:
```
                "matchExpressions": {
                 {
                    "key": "kubernetes.io/hostname",
                    "values": [
                        "node1",
                        "node2",
                        "node3"
                    ],
                    "operator": "In"
                 },           
                 {
                    "key": "xxx/critial-workload",
                    "operator": "DoesNotExist"
                 }}
```

Lyndon-Li: Yes, I have corrected it. Thanks.

ywk253100 (approved):

blackpiglet (dismissed):

sseago (dismissed):

design/node-agent-affinity.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Anti-affinity
+Anti-affinity configuration means preventing VGDP instances running in the nodes specified. Below is the way to define it:  
+- It could be defined by `MatchExpressions` of `metav1.LabelSelector`. The labels are defined in `Key` and `Values` of `MatchExpressions` and the `Operator` should be defined as `LabelSelectorOpNotIn` or `LabelSelectorOpDoesNotExist`.   
+
+### Sample
+A sample of the ```node-agent-config``` configMap is as below:
+```json
+{
+    "loadAffinity": {
```
reasonerjt, line 68: Shall we introduce a bit more flexibility here to allow multiple rules?
It may be useful when there are different CSI drivers installed on the cluster and we want to have different selectors for different drivers.

design/node-agent-affinity.md
```diff
@@ -0,0 +1,134 @@
...
+
+It is possible that node-agent pods, as a daemonset, don't run in every worker nodes, users could fulfil this by specify `nodeSelector` or `nodeAffinity` to the node-agent daemonset spec. On the other hand, at present, VGDP instances must be assigned to nodes where node-agent pods are running. Therefore, if there is any node selection for node-agent pods, users must consider this into this load affinity configuration, so as to guarantee that VGDP instances are always assigned to nodes where node-agent pods are available. This is done by users, we don't inherit any node selection configuration from node-agent daemonset as we think daemonset scheduler works differently from plain pod scheduler, simply inheriting all the configurations may cause unexpected result of backupPod schedule.    
+Otherwise, if a backupPod are scheduled to a node where node-agent pod is absent, the corresponding DataUpload CR will stay in `Accepted` phase until the prepare timeout (by default 30min).
```
reasonerjt, line 108: Let me clarify, this is a possible situation in v1.13, right?

IMO it's possible to take the node-selector in node-agent spec when scheduling the `bakcupPod`?

design/node-agent-affinity.md
```diff
@@ -0,0 +1,134 @@
...
+Once this problem happens, the backupPod stays in `Pending` phase, and the corresponding DataUpload CR stays in `Accepted` phase until the prepare timeout (by default 30min).   
+
+There is a common solution for the both problems:
+- We have an existing logic to periodically enqueue the dataupload CRs which are in the `Accepted` phase for timeout and cancel checks
+- We add a new logic to this existing logic to check if the corresponding backupPods are in unrecoverable status
+- The above problems could be covered by this check, because in both cases the backupPods are in abnormal and unrecoverable status
+- If a backupPod is unrecoverable, the dataupload controller cancels the dataupload and deletes the backupPod
```
reasonerjt, line 117: What will happen after the deletion?  Will the controller try to create another backupPod?  If we choose to allow retrying we need to set a limit otherwise there is a risk for infinite loop.

design/node-agent-affinity.md
```diff
@@ -0,0 +1,134 @@
...
+Otherwise, if a backupPod are scheduled to a node where node-agent pod is absent, the corresponding DataUpload CR will stay in `Accepted` phase until the prepare timeout (by default 30min).  
+
+At present, as part of the expose operations, the exposer creates a volume, represented by backupPVC, from the snapshot. The backupPVC uses the same storageClass with the source volume. If the `volumeBindingMode` in the storageClass is `Immediate`, the volume is immediately allocated from the underlying storage without waiting for the backupPod. On the other hand, the loadAffinity is set to the backupPod's affinity. If the backupPod is scheduled to a node where the snapshot volume is not accessible, e.g., because of storage topologies, the backupPod won't get into Running state, concequently, the data movement won't complete.
```
reasonerjt, line 110: I think we can explicitly document that when the storageclass has the BindingMode as "Immediate", the user SHOULD NOT set node selector for data mover

design/node-agent-affinity.md
Lyndon-Li: backupPod deletion is a part of the cancellation operations, after the data upload is cancelled, nothing will be retried.
Here we discover the backupPods that are not recoverable for sure, otherwise, for another interim situations that we cannot judge as Unrecoverable,  we don't trigger data upload cancelation, so Velero waits the backupPod until it turns to Running or preparingTimeout (30 minutes by default)

design/node-agent-affinity.md
Lyndon-Li: We can document this, but probably, we just need to tell users to be careful to set node-selection for `Immediate` volumes, as not all volumes have the constraints like topologies. In the envs with no constraints, node-selection works well with `Immediate` volumes

design/node-agent-affinity.md
Lyndon-Li: Yes, it is possible because users could modify node-agent's yaml and add any configurations affecting the schedule, i.e., topologies, affinities/anti-affinities, node selectors, etc.

Node-agent's node selection is done by Kubernetes' scheduler, the only way to control its behavior is to modify the node-agent's yaml; on the other hand, the data mover node-selection configuration is in a configMap that is detected after node-agent starts.
Therefore, if we reflect node-selection configuration into node-agent scheduling, we must dynamically edit node-agent's yaml after node-agent starts, so this causes node-agent restarts one more time.
Moreover, users may be surprised when node-agent restarts once more because they may not have realized their node-selection configuration has affected the node-agent scheduling.

Moreover, it is possible that node-agent pod cannot run in a specific node for sure, if we change the node-agent spec, things still doesn't work.

Therefore, users must know the relationship of node-agent scheduling and node-selection in either case. So we'd better have user realize this from the beginning and it is easy for them to make two correct configurations for node-agent spec and node-selection configuration.

reasonerjt: OK let's make sure this is covered in documentation.

design/node-agent-affinity.md
reasonerjt: I think it's acceptable to me that in v1.14 we only handle the first entry in the list.

Lyndon-Li: Done. As this is for extensibility consideration, we can add the `condition` in future as an omitable field when we have a clear requirement.

reasonerjt (approved):

design/node-agent-affinity.md
```diff
@@ -0,0 +1,137 @@
...
+
+At present, as part of the expose operations, the exposer creates a volume, represented by backupPVC, from the snapshot. The backupPVC uses the same storageClass with the source volume. If the `volumeBindingMode` in the storageClass is `Immediate`, the volume is immediately allocated from the underlying storage without waiting for the backupPod. On the other hand, the loadAffinity is set to the backupPod's affinity. If the backupPod is scheduled to a node where the snapshot volume is not accessible, e.g., because of storage topologies, the backupPod won't get into Running state, concequently, the data movement won't complete.  
+Once this problem happens, the backupPod stays in `Pending` phase, and the corresponding DataUpload CR stays in `Accepted` phase until the prepare timeout (by default 30min).   
+
+There is a common solution for the both problems:
+- We have an existing logic to periodically enqueue the dataupload CRs which are in the `Accepted` phase for timeout and cancel checks
+- We add a new logic to this existing logic to check if the corresponding backupPods are in unrecoverable status
```
blackpiglet, line 118: Could we give more information about the definition of the `unrecoverable` state?
It's better to let the user know in which condition the DataUpload cancelation or failure is expected.

Lyndon-Li: This Unrecoverable status check is an existing logic, here we add a further check in this existing logic.
At present, we don't have this check included in the doc, we can add it and include all the checks for backupPod/restorePod's Unrecoverable status.

blackpiglet (approved):

Lyndon-Li:
There are several comments covering requests for document. We will create a separate PR to add a document for node-selection. We will cover all the requests in that PR later.

PR #7410, opened 2024-02-09
Ignore missing path error in conditional match

seanblong:

For Resource Modifiers, don't treat unmatched paths as an error (`ErrMissing`) instead treat them the same way as an `ErrTestFailed` response.

This is particularly useful when we want to match on paths should they exist, such as the following example where I'm looking to remove a particular affinity rule.

```yaml
resourceModifierRules:
- conditions:
    groupResource: statefulsets.apps
    matches:
    - path: "/spec/template/spec/affinity/nodeAffinity/requiredDuringSchedulingIgnoredDuringExecution/nodeSelectorTerms/0/matchExpressions/0/key"
      value: cloud.google.com/gke-preemptible
  patches:
  - operation: remove
    path: "/spec/template/spec/affinity/nodeAffinity/requiredDuringSchedulingIgnoredDuringExecution/nodeSelectorTerms/0/matchExpressions/0"
```

I've tested this locally and with a deployed container image.  All statefulsets containing this particular path have it removed and all statefulset without it are filtered out, without creating an error message and allowing the restore to finish with a `Completed` status.

Fixes #(issue) 

N/A

- [X] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required` as a comment on this pull request.

### DIFF
--- a/changelogs/unreleased/7410-seanblong
+++ b/changelogs/unreleased/7410-seanblong
@@ -1,0 +1,1 @@
+Ignore missing path error in conditional match
--- a/internal/resourcemodifiers/resource_modifiers.go
+++ b/internal/resourcemodifiers/resource_modifiers.go
@@ -172,7 +172,7 @@ func matchConditions(u *unstructured.Unstructured, rules []MatchRule, _ logrus.F
 	p := &JSONPatcher{patches: fixed}
 	_, err := p.applyPatch(u)
 	if err != nil {
-		if errors.Is(err, jsonpatch.ErrTestFailed) {
+		if errors.Is(err, jsonpatch.ErrTestFailed) || errors.Is(err, jsonpatch.ErrMissing) {
 			return false, nil
 		}
 		return false, err
--- a/internal/resourcemodifiers/resource_modifiers_test.go
+++ b/internal/resourcemodifiers/resource_modifiers_test.go
@@ -1650,6 +1650,35 @@ func TestResourceModifiers_conditional_patches(t *testing.T) {
 			wantErr:       false,
 			wantObj:       cmWithLabelAToB.DeepCopy(),
 		},
+		{
+			name: "missing condition path and skip patches",
+			rm: &ResourceModifiers{
+				Version: "v1",
+				ResourceModifierRules: []ResourceModifierRule{
+					{
+						Conditions: Conditions{
+							GroupResource: "*",
+							Namespaces:    []string{"fake"},
+							Matches: []MatchRule{
+								{
+									Path:  "/metadata/labels/a/b",
+									Value: "c",
+								},
+							},
+						},
+						MergePatches: []JSONMergePatch{
+							{
+								PatchData: `{"metadata":{"labels":{"a":"c"}}}`,
+							},
+						},
+					},
+				},
+			},
+			obj:           cmWithLabelAToB.DeepCopy(),
+			groupResource: "configmaps",
+			wantErr:       false,
+			wantObj:       cmWithLabelAToB.DeepCopy(),
+		},
 	}
 
 	for _, tt := range tests {


anshulahuja98:
Valid suggestion
@seanblong - can you look into adding a unit test for this case?

seanblong:
> Valid suggestion @seanblong - can you look into adding a unit test for this case?

You bet!  The test I added is a little nonsensical as It's checking for a subattribute of a label, but testing before/after change shows that it accurately reports on `ErrMissing`.  I could replace with a more practical example, maybe using `podYAMLWithPVCVolume` and checking for a container index or mount out of range, but preferred keeping it concise (fitting with other tests).

anshulahuja98:
Hi @seanblong, sorry for the late reply. 
I was rethinking on the impact of current behaviour of this PR.
I might have missed few aspects when I first reviewed.

Basically with your change - all path missing errors are swallowed and not logged. 
This might be convenient for your use case, but I am worried, this would lead incorrect surfacing of error to users who might actually make mistakes in path.

An alternative to achieve what you wish is by using a test patch.
You first do a test patch to see if the path exists and then only do the chagnes -> this would lead to completed as the end status. 

Let me know if that sounds reasonable?

seanblong:
👋 @anshulahuja98,

No problem at all.  I actually went the route of `test` originally, but found some limitations, especially when we have lists (of lists) of potentially complex types that we're looking to represent.  Currently `matches` operates very similarly to `test`, but provides the ability/opportunity to ignore the `ErrMissing` error (in addition to ignoring `ErrTestFailed`), which I think is closer to the intention when conditionally matching on paths.  

Let me better illustrate this using two of our test variables.  The first having a list of volumeMounts and the second being a more vanilla Nginx image.  This acts as a good example where we may want a generic `resourceModifier`, but don't want to incur a lot of errors or a failed restoration.

https://github.com/vmware-tanzu/velero/blob/d45b313f0737030c62f8fd98d1f0fc01a1319094/internal/resourcemodifiers/resource_modifiers_test.go#L1062-L1083

https://github.com/vmware-tanzu/velero/blob/d45b313f0737030c62f8fd98d1f0fc01a1319094/internal/resourcemodifiers/resource_modifiers_test.go#L1050-L1060

Today the following will work on the former, but not the latter.  In the case of the latter it will throw an error:

```
resourceModifierRules:
- conditions:
    groupResource: pods
    matches:
    - path: "/spec/containers/0/volumeMounts/0/name"
      value: "vol2"
  patches:
  - operation: replace
    path: "/spec/containers/0/volumeMounts/0/name"
    value: "vol1"
```

In order to get this rule to work with `test`, without generating errors, we'd need to write something like this:

```yaml
resourceModifierRules:
- conditions:
    groupResource: pods
  patches:
  - operation: test
    path: "/spec/containers/0/volumeMounts"
    value: "[{\"mountPath\": \"/fake1\", \"name\": \"vol1\"},{\"mountPath\": \"/fake2\", \"name\": \"vol2\"}]"
  - operation: replace
    path: "/spec/containers/0/volumeMounts/0/name"
    value: "vol1"
```

This requires a complicated `value` that requires some sense of omniscience about the whole object.  Keep in mind that if we tried to use the following snippet in the `test` condition instead, the former would once again generate an error while the latter would succeed.

```yaml
  - operation: test
    path: "/spec/containers/0/volumeMounts/0/name"
    value: "name"
```

Alternative workarounds would be to make much more specific `resourceModifiers` for each workload being migrated, but I believe the strength of Velero, and more specifically, the `resourceModifier`, is to be fairly agnostic.  I'm hoping that in this PR we can leave `test` as is while leveraging some of the additional logic provided by `matchConditions`.

anshulahuja98:
@seanblong I think I understand now. 
The mistake I might have made is getting confused between the impl of actual test patch and the `matches` part.

This looks good to me. 
Approving.

anshulahuja98 (approved):

blackpiglet (approved):

Issue #7036, opened 2023-10-31
Data mover backup node black list - Don't run in specified node


Lyndon-Li:
The data mover backup exposer generally has the ability to select node for running the data movement. This enables us to fulfill below user requirement:
Sometimes, if a node is running very critical workloads, users don't want data movements to run in the node

We can develop a black list of node mechanism, the nodes in the list will not host data movements.

Lyndon-Li:
Another user case on this requirement https://github.com/vmware-tanzu/velero/discussions/7185

Lyndon-Li:
Another use case on this requirement #7243

balbiv:
@Lyndon-Li  thank you for adding this to the milestones. Because the helm deployment allows to set nodeSelectors for the node-agent, I created a dedicated node pool for it so data movement (CSI Snapshot) does have no impact on critical production applications. However, I figured that the pod responsible for mounting the backup PVC (running /velero-helper pause) is allowed to schedule on every node. If the backup pod can take over the nodeSelectors/tolerations from the node-agent daemonset, it will already be a big improvement. This also fixes that the backup pod cannot be started because the image pull policy is set to never.

Lyndon-Li:
@balbiv
Thanks for the suggestion, as the current plan, we will create a dedicate loadAffinity configMap, instead of coupling node-agent pods scheduling to data mover backup pods scheduling, for below reasons:

- The node-agent not only runs snapshot data movement, but for many other purpose, e.g., it also runs fs-backup/restore. So users may still not want to run data mover backups in all the nodes where node-agent pods resides in
- The other plan is to automatically inherit node-agent pods' node-selection configurations to backupPods'. However, we are clear of some of the configurations, i.e., nodeSelectors, but we are not confident to some others. Specifically, we are not confident that simply inheriting all the configurations from node-agent would bring the same result for backupPod schedule as the node-agent pod schedule since we think the daemonset scheduler behaves differently from plain pod scheduler

Finally, we decided not to bring node-agent's node selection into consideration, if there are some of these configurations, users must apply them to loadAffinity configMap manually in an appropriate way, see the design PR #7383.

This is the initial plan, we may make changes according to comments of the PR, so for any ideas, you can comment in the PR.

SCLogo:
I know it is a quite old ticket, but do you have idea how to set tolerations? I have nodes with tolerations set.

kaovilai:
I see it for maintenance jobs but not for node-agents.

[...]

PR #7437, opened 2024-02-18
Issue 7036: node selection for data mover backup

Lyndon-Li:
Fix issue #7036. Add the implementation of node selection for data mover backup


pkg/exposer/csi_snapshot.go
```diff
@@ -455,3 +485,42 @@ func (e *csiSnapshotExposer) createBackupPod(ctx context.Context, ownerObject co
 
 	return e.kubeClient.CoreV1().Pods(ownerObject.Namespace).Create(ctx, pod, metav1.CreateOptions{})
 }
+
+func toSystemAffinity(loadAffinity *nodeagent.LoadAffinity) *corev1.Affinity {
```
ywk253100, line 489: Why we cannot use the `corev1.Affinity` directly rather than defining our own type?

Lyndon-Li: I think there are two questions:

- Why do we wrap `metav1.LabelSelector`/`corev1.Affinity` instead of using it directly? -- We may need to add some other controllers that `metav1.LabelSelector`/`corev1.Affinity` cannot provide in future, so we add a simple wrap here. Whenever we want to change, we just append new fields so as to avoid break changes.
- Why don't we use `corev1.Affinity` directly? -- `corev1.Affinity` is more complex than we want to provide to users, e.g., we only use `RequiredDuringSchedulingIgnoredDuringExecution`, so we need to limit users input

ywk253100 (approved):

blackpiglet (approved):

Issue #6936, opened 2023-10-10
Velero fails to expose correct backup metrics after a pod restart


Ahmad-Faizan:
**What steps did you take and what happened:**
The metric `velero_backup_last_status` exposes the status of the latest backups.
Once a backup has been taken, the metric gets updated.
However, after a pod restart in between any two scheduled backups resets the metric exposed by `velero_backup_last_status`.
The metric only gets updated for backups after they are created.

**What did you expect to happen:**
Ideally, the metric should read the list of backups and set the `velero_backup_last_status` metric.
So if a backup happens at 12:00 and the velero pod is restarted or killed at 12:30, the metric
should not be set to 0 (which indicates no backup has been taken).

**The following information will help us better understand what's going on**:

**Environment:**

- Velero version (use `velero version`): v1.11.0
- Velero features (use `velero client config get features`): <NOT SET>
- Kubernetes version (use `kubectl version`): v1.26.8
- Kubernetes installer & version: kops v1.26.5
- Cloud provider or hardware configuration: AWS
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04.5 LTS

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "I would like to see this bug fixed as soon as possible"
- :-1: for "There are more important bugs to focus on right now"

yanggangtony:
you use the version is 1.11?
the default is changed to 1 in [https://github.com/vmware-tanzu/velero/pull/6838/files](https://github.com/vmware-tanzu/velero/pull/6838/files)

yanggangtony:
I check the release 1.12 , the pr is not include in that.
Maybe wait release 1.13 will be include?

allenxu404:
> Ideally, the metric should read the list of backups and set the velero_backup_last_status metric.

Do you mean the velero_backup_last_status metric should read the most recently completed backup before the Velero pod restarts to determine the metrics' value for schedule backup? 

The metrics is reset to 0 when Velero pod restarts because the default value of the metrics is 0. This has been changed in the PR: #6838 as @yanggangtony mentioned.

jkroepke:
[...]

PR #7445, opened 2024-02-20
Adjust the logic for the backup_last_status metric to stop incorrectly incrementing over time

### LINKED ISSUE #7378: v1.13.0 changes velero_backup_last_status metric

After upgrading to velero v1.13.0, the velero_backup_last_status metric seems to have changed.

Before, the metric was constant:
Bildschirmfoto 2024-02-02 um 10 34 27

Now, it is constantly increasing until the next backup is run:
Bildschirmfoto 2024-02-02 um 10 28 21

The last backup did complete without any issues:
````
velero-my-backup-20240202020000   Completed   0        0          2024-02-02 03:00:00 +0100 CET   27d       default            <none>
`````

The metric stays the same.

The issue occurs directly after the update, without a new backup being created.

**Environment:**

- Velero version (use `velero version`):  v1.13.0
- Velero features (use `velero client config get features`): features: <NOT SET>
- Kubernetes version (use `kubectl version`): v1.28.4-eks-8cb36c9
- Cloud provider or hardware configuration: aws eks 1.28
- OS (e.g. from `/etc/os-release`): amazon_linux:2

**Vote on this issue!**

### PR DESCRIPTION
**Pull request by allenxu404**

Fixes #7378

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -468,7 +468,7 @@ func (m *ServerMetrics) InitSchedule(scheduleName string) {
 		c.WithLabelValues(scheduleName).Add(0)
 	}
 	if c, ok := m.metrics[backupLastStatus].(*prometheus.GaugeVec); ok {
-		c.WithLabelValues(scheduleName).Add(1)
+		c.WithLabelValues(scheduleName).Set(float64(1))
 	}
 	if c, ok := m.metrics[restoreAttemptTotal].(*prometheus.CounterVec); ok {
 		c.WithLabelValues(scheduleName).Add(0)
--- a/changelogs/unreleased/7445-allenxu404
+++ b/changelogs/unreleased/7445-allenxu404
@@ -1,0 +1,1 @@
+Adjust the logic for the backup_last_status metrics to stop incorrectly incrementing over time

### CODE REVIEW

**Review by blackpiglet — APPROVED** [review]

File: pkg/metrics/metrics.go [hunk 1]
```diff
@@ -468,7 +468,7 @@ func (m *ServerMetrics) InitSchedule(scheduleName string) {
 		c.WithLabelValues(scheduleName).Add(0)
 	}
 	if c, ok := m.metrics[backupLastStatus].(*prometheus.GaugeVec); ok {
-		c.WithLabelValues(scheduleName).Add(1)
+		c.WithLabelValues(scheduleName).Set(float64(1))
```
Snaipe, line 471: Found this PR when trying to figure out why the metric was increasing; if I understand correctly, this ends up running periodically (every time the schedule is reconciled), right?

If so, wouldn't this incorrectly clear a failing status? Say that the schedule created a backup that failed, the metric would become 0, and then get reset back to 1 after a couple of minutes, which seems problematic since this can hide failures.

If anything, I think lines 470-471 should be omitted entirely. This would prevent from emitting a data point until the point where we actually have one, which should still fix what #6838 was aiming for. If an accurate status must be set, it should be done as part of scheduleReconciler.Reconcile, using the status of the last backup created by the schedule.

allenxu404: @Snaipe yes, you're right.

This PR only aimed to address the specific bug caused in https://github.com/vmware-tanzu/velero/pull/6838. I will create a separate PR to resolve the issue with incorrect failure status clear and keep you posted. Thanks for highlighting this larger problem!

The purpose of setting the default value of backup_last_status to 1  was to prevent mistaken failure reports that could occur if the code exits without updating that metric. I prefer to leave that implementation unchanged. Upon further analysis, I think the true underlying issue seems to be that metrics are re-initialized each time the schedule is reconciled, instead, the metrics should be initialized only if a new backup is due to be created.  So I think the `InitSchedule`function would be better placed below, where backups are created, to avoid unnecessary reset.  what's your thought?
```
       if c.ifDue(schedule, cronSchedule) && !c.checkIfBackupInNewOrProgress(schedule) {
		if err := c.submitBackup(ctx, schedule); err != nil {
			return ctrl.Result{}, errors.Wrapf(err, "error submit backup for schedule %s", req.String())
		}
		c.metrics.InitSchedule(schedule.Name)
	}
```

Snaipe: I think it'd help, but I'm not sure it would fix the underlying problem. As far as I can tell, what this would do is that if your schedule is failing, and a new backup is created, then the schedule would be reported as "fixed" metrics-wise for the entire duration of the backup.

It's less bad than the first situation, but it would definitely confuse alerting.

I think not advertising a value for the metric up until the point where there is a data point seems slightly better in the sense that users can choose to ignore them or set it to the default value of their choice (0 or 1) in grafana/alertmanager. A better solution would be to set the initial value of the metric directly to the result of the last backup of the schedule.

allenxu404: > A better solution would be to set the initial value of the metric directly to the result of the last backup of the schedule.

I agree, will raise another issue to track it.

allenxu404: Seems we've already has the similar issue to track: https://github.com/vmware-tanzu/velero/issues/6936

Issue #7345, opened 2024-01-23
Support only restore volume data of a backup


blackpiglet:
**Describe the problem/challenge you have**

Sometimes, it's possible to only restore data into volume without modifying the k8s resources.
Some issues required a similar function, #504 and #2598.
Suggest, at least, provide a workaround for this scenario.

**Describe the solution you'd like**

If a PodVolumeRestore is created based on the PodVolumeBackup information manually, the PodVolumeRestore is not processed by the Velero PVR controller. This is due to the PVR controller checking the PVR-requested pod whether has a restore-helper InitContainer. If not, the PVR is neglected.

Suggest adding a new field in the PVR to make the PVR controller skip the restore-helper InitContainer check logic. 
`AllowOnlyRestoreData` is the proposed parameter name. Its type is bool. The default value is false. This is an example:
``` yaml
apiVersion: velero.io/v1
kind: PodVolumeRestore
metadata:
  name: test-01-xxxxxxxxxxxxx-xxxxxx
  namespace: velero
spec:
  backupStorageLocation: default
  pod:
    kind: Pod
    name: hello-app-xxxxxxxxx-xxxxx
    namespace: restore
    uid: xxxxxxxx
  repoIdentifier: gs:jxun:/restic/test
  snapshotID: xxxxxxxx
  sourceNamespace: test
  uploaderSettings:
    WriteSparseFiles: "false"
  uploaderType: kopia
  volume: sdk-volume
  allowOnlyRestoreData: true
```

**Anything else you would like to add:**

**Environment:**

- Velero version (use `velero version`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

blackpiglet:
As a PoC, already verified that, after bypassing the InitContainer check code, the manual created PVB worked as expected.

stefanhenseler:
[...]

PR #7481, opened 2024-02-29
Add design for only restore volume data.

blackpiglet:

Fixes #(issue)
This is the design document for issue https://github.com/vmware-tanzu/velero/issues/7345.

### DIFF
--- a/design/data_only_restore.md
+++ b/design/data_only_restore.md
@@ -1,0 +1,126 @@
+# Data-only restore design
+
+## Abstract
+Support to restore only the data in the backup volume without restoring the k8s resources.
+
+## Background
+In some disaster recovery scenarios, the user only wants to restore the data in the volumes, because only specific volume's data is corrupted, but Velero doesn't support that yet. The design aims to address that missing function.
+
+## Goals
+- Support to restore only volume data for the filesystem-based backup.
+- Support the in-place volume data restore, which means the data will be restored into existing volumes, instead of creating new volumes.
+- Can filter the volumes that needs data restore.
+
+## Non Goals
+- Doesn't support the snapshot-based backup at phase one, because there is no easy way to read data from snapshot directly. To support the in-place restore method, may need to introduce a way to provision an intermediate volume created from the snapshot, then restore data from the intermediate volume.
+- Cannot filter restored volume by name. It's convenient to specify the wanted volume by name, but the Velero backup and restore don't have that function yet, the feature will not also aim to introduce new resource filters.
+
+## Detailed Design
+A new parameter `DataOnly` is added to the `RestoreSpec` structure. The default value of `DataOnly` is `false`.
+
+``` go
+type RestoreSpec struct {
+    ......
+	// DataOnly specify whether to only restore volume data and related k8s resources
+	// without any other k8s resources.
+	DataOnly *bool `json:"dataOnly"`
+    ......
+}
+```
+
+### Restore CLI change
+The `--data-only` parameter is added to the `velero restore create` CLI.
+Both `--data-only` and `--data-only=true` mean enabling the data-only restore feature.
+Both `--data-only=false` and not specifying the parameter mean disabling the data-only restore feature.
+
+The `velero restore describe` CLI should display the data-restored PVCs' information.
+The `velero restore describe` CLI should also display the `--data-only` parameter value.
+
+### Resource Filtering
+The restore provided filters only apply to the referenced backup included `PersistentVolumeClaims` resources.
+The user can use the filters(including `IncludedNamespaces`, `ExcludedNamespaces`, `LabelSelector` and `OrLabelSelectors`) to select which volume data to restore.
+
+The restore can include `IncludedResources`, `ExcludedResources`, `NamespaceMappings`, `ResourceModifierConfigMap`, `StatusExcludeResources` and `StatusIncludeResources`, but those parameters have no effect.
+The `DataOnly` parameter's description should explain that.
+
+The resource collector should be limited to PersistentVolumeClaim, Pod and DataUpload resources.
+PVC should be filtered based on their volume data backup method, only PodVolumeBackup and DataUpload are kept. The kept PVCs are store into restoreContext.
+
+``` go
+type restoreContext struct {
+	...
+	DataOnlyRestorePVCs []types.NamespacedName
+	...
+}
+```
+
+Then only the pod mounting those PVCs are handled.
+If the PVC is not mounted by any Pod, then the data restore of this PVC is skipped too.
+
+### Restore Plugin Actions
+Need to limit the applicable actions for the data-only restore.
+Only the DataUploadRetrieveAction, InitRestoreHookPodAction and PodVolumeRestoreAction are relevant.
+
+### Use PodVolumeRestore to restore the DataUpload
+1. The data-only restore will use the existing way to read DataUpload information into ConfigMap, and clean the ConfigMap when restore completes.
+1. The Velero filters the PVCs included in the backup by the restore's filters. 
+1. Then Velero should get the DataUpload's ConfigMap for the filtered PVCs.
+1. Generate the PodVolumeRestore according to the DataUpload's ConfigMap information.
+1. Add restore-wait InitContainer into the PVC-mounting pod.
+
+The following is the same to existing filesystem restore.
+
+### Use PodVolumeRestore to restore PodVolumeBackup
+
+1. The Velero filters the PVCs included in the backup by the restore's filters. 
+1. Then Velero should get the PodVolumeBackup for the filtered PVCs.
+1. Generate the PodVolumeRestore according to the PodVolumeBackup information.
+1. Add restore-wait InitContainer into the PVC-mounting pod.
+
+The following is the same to existing filesystem restore.
+
+### Restore Hooks
+The InitContainer Hook and the Exec Hook are still supported as before.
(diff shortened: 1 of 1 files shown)


draghuram:
@blackpiglet, Can you please rebase the PR? It is currently showing what appears to be many unrelated commits.

Few comments/questions:

- How are the new PVC names generated? I doesn't look like users can specify the target names so I guess Velero will rename them in a certain way?
- Since "pods" is not included in the resource type, I guess this restore is not supported for FSB?
- Does Snapshot data mover restore work for PVCs without specifying pods as resource type? I will verify it but it can work I guess because a mover pod is started?

I personally think that in-place restore is going to be more useful, especially if files can be selected for restore. If the whole PVC is being created any way, users will need to restart pods to attach the PVCs so it is almost like restoring pods themselves.

anshulahuja98:
Can https://github.com/vmware-tanzu/velero/pull/6354 be used  in a generic way to achieve this use case?
CC: @kaovilai

kaovilai:
> Can https://github.com/vmware-tanzu/velero/pull/6354 be used  in a generic way to achieve this use case?
> CC: @kaovilai 

So restore limiting resources to a labeled pv (or namespaced PVC) with recreate flag?

blackpiglet:
@anshulahuja98 
If the in-place restore is required, then the recreate Existing Resource Policy cannot do that.

blackpiglet:
@draghuram 
Thanks. I rebased the branch, but there is still some problem with GPRC generating.

First, could you share some thoughts about why the in-place restore is more useful? I recently also heard some requirements about the GitOps and DevOps pipeline scenario. In those cases, some tools guarantee that the workload in k8s always running as expected. The in-place restore is needed in those cases, but I don't know whether those are common cases for k8s usage.

Second, to answer your question.
* How are the new PVC names generated? I doesn't look like users can specify the target names so I guess Velero will rename them in a certain way? - User cannot specify the generated name. I haven't decided how to build the name yet. Maybe something as `rename-pvc-<uuid>`. 
Since "pods" is not included in the resource type, I guess this restore is not supported for FSB? - FSB restore is supported, because the Generate Backup/Restore will first create an intermediate pod and PVC to mount the PV. After the backup/restore, the intermediate pod and PVC are deleted.
Does Snapshot data mover restore work for PVCs without specifying pods as resource type? I will verify it but it can work I guess because a mover pod is started? - Snapshot data mover can work without pod, the reason is the same the second item.

design/data_only_restore.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Background
+In some disaster recovery scenarios, the user only wants to restore the data in the volumes, because only specific volume's data is corrupted, but Velero doesn't support that yet. The design aims to address that missing function.
+
+## Goals
+- Support to restore only volume data in the backup.
+- Can filter the volumes that needs data restore.
+
+## Non Goals
+- Will not support the in-place volume data restore. To achieve data integrity, new PVC and PV are created, and wait for the user to mount the restored volume.
```
kaovilai, line 14: Heard from US community meeting that this will be moved into Goals.

Would have to discuss how we're handling volumes that are already mounted.. or do we say, you have to pre-delete/scaledown so volumes are not mounted by restore time.

design/data_only_restore.md
```diff
@@ -0,0 +1,141 @@
...
+	// cluster-scoped resources.
+	Cluster []string `json:"cluster,omitempty"`
+
+	// Namespaces is a map of namespace name to slice of messages
+	// related to backup or restore namespace-scoped resources.
+	Namespaces map[string][]string `json:"namespaces,omitempty"`
+
+    // DataOnlyPvcMap is a map of backed-up PVC name and the data-only
+    // restored PVC name.
+    DataOnlyPvcMap map[types.NamespacedName]types.NamespaceName `json:"dataOnlyPvcMap,omitempty"`
```
kaovilai, line 74: It should be possible to add labelSelector to all resources that user expect to be swapping to restored PVC name, and having velero do the swapping on labeled (core k8s only for now) resources in the namespace.

```
kubectl label deployment deployName velero.io/sync-restored-pvc-name=<original-pvc-in-backup-name>
```

Any PVC restored using this method with a new name will have annotation `velero.io/restored-pvc-original-name=<original name>` that velero can use as reference for subsequent backup/restore pvc name syncing.

draghuram:
@blackpiglet, gitops is certainly one reason for in place restore. Another use case is the need to periodically update data in an alternate standby cluster. Finally, if some application files are deleted or corrupted, user may only want to restore those files.

design/data_only_restore.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Detailed Design
+A new parameter `DataOnly` is added to the `RestoreSpec` structure. The default value of `DataOnly` is `false`.
+
+``` go
+type RestoreSpec struct {
+    ......
+	// DataOnly specify whether to only restore volume data and related k8s resources
+	// without any other k8s resources.
+	DataOnly *bool `json:"dataOnly"`
```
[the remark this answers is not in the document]
kaovilai: We could use the data mover like approach where we write to a new pvc using file system backup through a dummy pod.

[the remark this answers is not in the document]
blackpiglet: If we go with the in-place restore, only filesystem backup is supported.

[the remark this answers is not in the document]
kaovilai: An advanced CSI in-place could auto patch deployments to a new pvc restored from snapshots.

blackpiglet:
> @blackpiglet, gitops is certainly one reason for in place restore. Another use case is the need to periodically update data in an alternate standby cluster. Finally, if some application files are deleted or corrupted, user may only want to restore those files.

@draghuram 
Thanks for the feedback.
If we go for the in-place data restore way, there will be some limitations to the using scenario. Only the filesystem uploader backup is supported. For the snapshot-based backup, Achieving the in-place restore result is impossible.

Is this acceptable? Data-only restore cannot support snapshot-based backups.
Involve more maintainers in this discussion.
@anshulahuja98 @kaovilai @sseago

kaovilai:
FSB only is acceptable.

Tho I think restore from CSI snapshot to new PVC and then patching to use new PVC could be considered in-place with the caveat that it requires pod restart.

blackpiglet:
> However I think restore from CSI snapshot to new PVC and then patching to use new PVC could be considered in-place with the caveat that it requires pod restart.

Thanks for the quick response, then I will follow the filesystem-only way for now. 
I got your point. There was a similar discussion in the Velero team. It does work, although it seems a bit inefficient. We can continue the discussion.
I want to point out the idea of the filesystem data-only restore will only cause the pod to restart, because the PodVolumeRestore way is preferred. That will introduce a new InitContainer in the pod. The advantage is that the pod's data operation will not interrupt the data restore process.

design/data_only_restore.md
blackpiglet: As we go with the in-place way, the name-mapping logic is unnecessary.

kaovilai: right.. for FSB with deleting PVC prior to restoring to the same name.

anshulahuja98:
> > @blackpiglet, gitops is certainly one reason for in place restore. Another use case is the need to periodically update data in an alternate standby cluster. Finally, if some application files are deleted or corrupted, user may only want to restore those files.
> 
> @draghuram Thanks for the feedback. If we go for the in-place data restore way, there will be some limitations to the using scenario. Only the filesystem uploader backup is supported. For the snapshot-based backup, Achieving the in-place restore result is impossible.
> 
> Is this acceptable? Data-only restore cannot support snapshot-based backups. Involve more maintainers in this discussion. @anshulahuja98 @kaovilai @sseago

I went through further over the discussion
I understand a pure in-place restore can't be achieved without pod downtime.

With that caveat, I want to push for CSI snapshot based in-place restore. Here in-place restore refers mainly to Detaching PVC from workload, deleting PVC and re-creating PVC. 

This is useful for Disaster recovery scenarios.

blackpiglet:
> With that caveat, I want to push for CSI snapshot based in-place restore. Here in-place restore refers mainly to Detaching PVC from workload, deleting PVC and re-creating PVC.

This proposal also cannot avoid pod downtime, so I suppose it aims to make the data-only restore can support more types of backups.

> I understand a pure in-place restore can't be achieved without pod downtime.

If no downtime is a must-have feature, we can achieve that by ignoring the PodVolumeRestore's related pod InitContainer check logic.

Lyndon-Li:
I just noticed that this design doesn't cover the consideration for WaitForFirstConsumer volumes, some discussion [see here](https://kubernetes.slack.com/archives/C021GPR1L3S/p1714030601688479).
According to the design, we will create the PVC/PV for users.
If the volume to be restored is with WaitForFirstConsumer mode, there is no way to create the volume appropriately without having the information which possible nodes the pod (that is going to consume the restored volume) will be scheduled.
For data only restore, we cannot assume that users could create the pod beforehand, but users must know which nodes they want to schedule their pod if they want to make some constraints.
Therefore, we need to add a new parameter for users to deliver the candidate nodes for a specific volume. These is a must have parameter.

WRKT:
Hello Velero team !

Just asking, does this feature still ongoing on next releases ? 
Thanks!

kaovilai:
For tracking of consideration for next release, please check the issue. https://github.com/velero-io/velero/issues/7345

As of now, not yet.

PR #7442, opened 2024-02-19
using zfs for vanilla cluster test

danfengliu:
Thank you for contributing to Velero!

# Please add a summary of your change

# Does your change fix a particular issue?

Fixes #7405
Tests #7388 

# Please indicate you've done the following:

- [ ] [Accepted the DCO](https://velero.io/docs/v1.5/code-standards/#dco-sign-off). Commits without the DCO will delay acceptance.
- [ ] [Created a changelog file](https://velero.io/docs/v1.5/code-standards/#adding-a-changelog) or added `/kind changelog-not-required` as a comment on this pull request.
- [ ] Updated the corresponding documentation in `site/content/docs/main`.

PR #7567, opened 2024-03-26
Delete ns using kubectl

danfengliu:

After PR #7442 merged, in kubeadm v1.29 pipeline, namespace clean step hit a hung issue by client-go deletion that namespace keep in terminating status although no any namespace related resources left, but kubectl is able to delete that namespace, so modify current deletion function by using kubectl.


blackpiglet:
Could you explain why you need to switch from client-go to the kubectl?

danfengliu:
Behavior after namespace deletion is not same from client-go and kubectl, I found NS is hanging in terminating status ( I've checked ns has no any related resource any more) no matter how many time client-go deletion performed, but kubectl NS deletion can delete this kind of NS in terminating immediately, although I did not found the root cause of it, anyway using kubectl is still a better choice of using client-go for E2E test in most cases.

blackpiglet (approved):

qiuming-best (approved):

Issue #7198, opened 2023-12-11
Microservice Velero's data movement activities


Lyndon-Li:
There are some cases require Velero's data movement activities to be in a dedicate microservice in a dedicate pod:

- Velero's fs-backup/restore, snapshot data mover requires to access volume data through host path (and even privileged mode in some environments), which makes challenge to some users
- Velero accesses the backup storage from either Velero server pod or node-agent pods, including the resource data upload/download, fs-backup and data mover. Consequently, when the backup storage must be represented by a Kubernetes flavor volume (i.e., persistent volume, e.g., nfs storage, COSI), the volume must be mounted to either Velero server pod or node-agent pods, which causes Velero server/node-agent restart, this is almost not accepted
- There may be requirements to control resource (i.e., cpu, memory) request/limit in a granular manner, e.g., control them per backup/restore of a volume
- This increases the resilience, crash of one activity won't affect the others
- This prevents unnecessary full backup. Velero's fs uploaders support file level incremental backup by comparing the file name and metadata. However, at present the files are visited by host path, while pod and PVC's ID are part of the host path, so once the pod is recreated, the same file is regarded as a different file since the pod's ID has been changed.

If we can make VGDP instances and the backup storage direct accesses as dedicate microservices in dedicate pods, the above problem will be solved

reasonerjt:
This is a valuable improvement, but I'm not sure if we can finish all the dev work in v1.14.  Therefore, let's use this issue to track the design work, and we'll revisit it before feature freeze to see if we can implement it in v1.14.

PR #7576, opened 2024-03-28
Data mover micro service design

Lyndon-Li:
Fix issue #7198. Add the design for data mover micro service


design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,186 @@
...
+The primary task of VGDP Watcher is to watch the status change from backupPod/restorePod or the VGDP instance, so as to inform the data mover controller in below situations:
+- backupPod/restorePod starts
+- VGDP instance starts
+- Progress update
+- VGDP instance completes/fails/cancelled
+- backupPod/restorePod stops
+
+We use two mechanism to make the watch:  
+**Pod Phases**: VGDP Watcher watches the backupPod/restorePod's phases updated by Kubernetes. That is, VGDP Watcher creates an informer to watch the pod resource for the backupPod/restorePod and detect that the pod reaches to one of the terminated phases (i.e., PodSucceeded, PodFailed). We also check the availability & status of the backupPod/restorePod at the beginning of the watch so as to detect the starting of the backupPod/restorePod.  
+**Custom Kubernetes Events**: VGDP-MS generates Kubernetes events and associates them to the backupPod/restorePod at the time of VGDP instance starting/stopping and progress update, then VGDP Watcher creates another informer to watch the Event resource associated to the backupPod/restorePod.
```
reasonerjt, line 72: What's the specific use case for this?  reporting the transfer status?
IMHO, we wanna try to make the code that runs in backupPod less k8s-aware or k8s-dependent, so we can test it or verify certain cases more easily.  But this is an immature thought and we can have further discussions on this topic.

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,186 @@
...
+	TotalBytes int64 `json:"totalBytes,omitempty"`
+	BytesDone  int64 `json:"doneBytes,omitempty"`
+}
+``` 
+
+### Log Collection
+During the running of VGDP instance, some logs are generated which are important for troubleshooting. This includes all the logs generated by the uploader and repository. Therefore, it is important to collect these logs.  
+On the other hand, the logs are now generated in the backupPod/restorePod, while the backupPod/restorePod is deleted immediately after the data movement completes. Therefore, by default, ```velero debug``` is not able to collect these logs.  
+
+As a solution, we use logrus's hook mechanism to redirect the backupPod/restorePod's logs into node-agent's log, so that ```velero debug``` could collect VGDP logs as is without any changes.
```
reasonerjt, line 159: What's the pre-requisite of such hook?  Is it done over network and how do we ensure the connectivity?

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,186 @@
...
+- This enable users to to control resource (i.e., cpu, memory) request/limit in a granular manner, e.g., control them per backup/restore of a volume
+- This increases the resilience, crash of one VGDP activity won't affect others
+- In the cases that the backup storage must be represented by a Kubernetes persistent volumes (i.e., nfs storage, [COSI][3]), this avoids to dynamically mount the persistent volumes to node-agent pods and cause node-agent pods to restart (this is not accepted since node-agent lose it current state after its pods restart)  
+- This prevents unnecessary full backup. Velero's fs uploaders support file level incremental backup by comparing the file name and metadata. However, at present the files are visited by host path, while pod and PVC's ID are part of the host path, so once the pod is recreated, the same file is regarded as a different file since the pod's ID has been changed. If the fs uploader is in a dedicate pod and files are visited by pod's volume path, files' full path are not changed after pod restarts, so incremental backups could continue.
```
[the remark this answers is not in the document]
Lyndon-Li: The pod refers to the source pod to be backed up.
The incremental backup compares the full path for each file, while the source pod's ID is part of the full path. Once the pod is recreated, the full path will change, so the same file will be treated as a new file.

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,186 @@
...
+```velero vgdp-ms backup --this-pod xxx --this-container xxx --this-volume xxx --resource-timeout xxx --log-format xxx --log-level xxx```  
+Or:  
+```velero vgdp-ms restore --this-pod xxx --this-container xxx --this-volume xxx --resource-timeout xxx --log-format xxx --log-level xxx```  
+
+The first one is for backup and the other one is for restore.  
+Below are the parameters of the commands:  
+**this-pod**: Deliver the backupPod/restorePod's name to VGDP-MS. Even though VGDP-MS is running inside the backupPod/restorePod, it is not easy to get the name of pod it is running in, so it is simply delivered through the parameter.  
+**this-container**: When backupPod/restorePod is started, the cluster may insert other containers to the pod, so we cannot assume that the container running VGDP-MS is always the first container in the pod. So the container name is simply delivered through the parameter.  
+**this-volume**: The name of the snapshot/target volume. VGDP-MS accesses the snapshot/target volume by matching this-volume to the volumeMount/volumeDevice in backupPod/restorePod's spec, so that it could access the snapshot/target volume locally.
```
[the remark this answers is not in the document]
Lyndon-Li: We need these 3 parameters for below purposes:
1. Find the path of the volume to be backed up
2. Find the DUCR/DDCR this pod is for

So these 3 parameters could be replaced by:
1. volume-path
2. DUCR/DDCR name

Considering the existing Exposer code, it actually doesn't know the DUCR/DDCR name, so I used these 3 parameters. But we can replaced them with the above new proposals.

design/vgdp-micro-service/vgdp-micro-service.md
Lyndon-Li: `Custom Kubernetes Events` doesn't mean a mechanism out of k8s scope, so it is still k8s aware/dependent.
Here the Events refer to the Kubernetes Event resource, that is, we generate events to the backupPod/restorePod during the data movement.
We use the Events to transfer message to the controller during the data movement, since the `Pod Phase` and `Pod Terminate Message` mechanism can only deliver info at the time of pod terminates.

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,186 @@
...
+**this-container**: When backupPod/restorePod is started, the cluster may insert other containers to the pod, so we cannot assume that the container running VGDP-MS is always the first container in the pod. So the container name is simply delivered through the parameter.  
+**this-volume**: The name of the snapshot/target volume. VGDP-MS accesses the snapshot/target volume by matching this-volume to the volumeMount/volumeDevice in backupPod/restorePod's spec, so that it could access the snapshot/target volume locally.  
+**resource-timeout**: resource-timeout is used to control the timeout for operations related to resources. It has the same meaning with the resource-timeout for node-agent.  
+**log-format** and **log-level**: This is to control the behavior of log generation inside VGDP-MS. 
+
+In order to have the same capability and permission with node-agent, below pod configurations are inherited from node-agent and set to backupPod/restorePod's spec:
+- Volumes: Some configMaps will be mapped as volumes to node-agent, so we add the same volumes of node-agent to the backupPod/restorePod
+- Environment Variables
```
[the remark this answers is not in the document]
Lyndon-Li: Yes, theoretically, the backupPod/restorePod only needs to host the VGDP instance. And besides that, we also need a kube client so as to fulfill below communicating and controlling purposes:

- Watch some status of DUCR/DDCR for synchronization with the controller, i.e., ready for start VGDP; VGDP cancellation
- Deliver information to the controller, i.e., VGDP progress

There is no other functionalities in backupPod/restorePod

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,186 @@
...
+**VBDM**: VGDM contains the data mover controller, while the controller calls the Exposer and launches the VGDP instances.  Now, since the VGDP instance is launched by the backupPod/restorePod, the controller should not launch the VGDP instance again. However, the controller still needs to monitor and control the VGDP instance. Moreover, in order to avoid any contest situations, the controller is still the only place to update DUCRs and DDCRs.    
+
+Besides the changes to above existing components, we need to add below new components:  
+**VGDP Watcher**: We create a new module to help the data mover controller to watch activities of the VGDP instance in the backupPod/restorePod. VGDP Watcher is a part of VBDM.  
+**VGDP-MS**: VGDP Micro Service is the binary for the command backupPod/restorePod runs. It accepts the parameters and then launches the VGDP instance according to the request type, specifically, backup or restore. VGDP-MS also runs other modules to sync-up with the data mover controller. VGDP-MS is also a part of VBDM.  
+
+Below diagram shows how these components work together:  
+![vgdp-ms-1.png](vgdp-ms-1.png)
```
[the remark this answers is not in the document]
reasonerjt: Per discussion, it is safer to let VGDP instance cancel the transfer within the pod rather than letting controller send a signal to kill the pod.  Therefore, letting the pod watch the DUCR/DDCR is considered better approach.

design/vgdp-micro-service/vgdp-micro-service.md
reasonerjt: After offline discussion, we agreed to choose k8s events as the approach for backup/restore pod to communicate to the node-agent, but maybe point the `involvedObject` to the DUCR is a better choice.

design/vgdp-micro-service/vgdp-micro-service.md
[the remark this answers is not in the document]
reasonerjt: We may open an issue to limit the scope of the roles of the backupPod/restorePod but it can be optional for phase 1 implementation.

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,186 @@
...
+**VGDP Watcher**: We create a new module to help the data mover controller to watch activities of the VGDP instance in the backupPod/restorePod. VGDP Watcher is a part of VBDM.  
+**VGDP-MS**: VGDP Micro Service is the binary for the command backupPod/restorePod runs. It accepts the parameters and then launches the VGDP instance according to the request type, specifically, backup or restore. VGDP-MS also runs other modules to sync-up with the data mover controller. VGDP-MS is also a part of VBDM.  
+
+Below diagram shows how these components work together:  
+![vgdp-ms-1.png](vgdp-ms-1.png) 
+
+## Detailed Design
+### Exposer
+At present, the exposer sets ```velero-helper pause``` as the command run by backupPod/restorePod. Now, VGDP-MS command will be used, the command is like below:
+```velero vgdp-ms backup --this-pod xxx --this-container xxx --this-volume xxx --resource-timeout xxx --log-format xxx --log-level xxx```
```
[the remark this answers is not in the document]
reasonerjt: we should keep it in velero, but rename the sub-command.

design/vgdp-micro-service/vgdp-micro-service.md
Lyndon-Li: After some further checks, I can see below negative points if we want to make `involvedObject` as DUCR/DDCR:
1. Kubernetes' etcd is not good at to store huge number of objects, this is why the event objects are automatically removed after 1 hour(by default). Since DUCR/DDCR are to be permanently kept, it will increase the burden to etcd when there are lots of DUCR/DDCR. On the other hand, since the events are deleted after 1 hour, we have no way to keep the events for further use, so there is no gain to keep the events with DUCR/DDCR.
2. The ms-service solution introduced in this design is capable to be reused by other requirements in future and Event is a core part of this solution. Adding DUCR/DDCR as `involvedObject` requires adding one more input parameter for the ms-service. As a comparison, originally, we only require `this-pod`, `this-container` and `this-volume` as the input.

shawn-hurley (commented):
I don't see this, but it feels like a perfect thing to at least think through how a third-party datamover could now plugin into this system.

design/vgdp-micro-service/vgdp-micro-service.md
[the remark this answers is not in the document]
Lyndon-Li: `this-pod` runs a subcommand of velero

Lyndon-Li:
> I don't see this, but it feels like a perfect thing to at least think through how a third-party datamover could now plugin into this system.

The scope of this design is for VGDP, VGDP is a part of VBDM (Velero Build-in Data Mover), so this design doesn't cover 3rd data movers.
3rd data movers still follow [this section](https://github.com/vmware-tanzu/velero/blob/main/design/volume-snapshot-data-movement/volume-snapshot-data-movement.md#:~:text=integrated%20with%20VBDM.-,Replacement,-3rd%20parties%20could) in the Volume Snapshot Data Movement design. Specifically, 3rd data movers handles the DUCR/DDCR and they are free to make their data movers micro service style or monolith service style.

I can add a clarification for this in the design.

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,186 @@
...
+- This increases the resilience, crash of one VGDP activity won't affect others
+- In the cases that the backup storage must be represented by a Kubernetes persistent volumes (i.e., nfs storage, [COSI][3]), this avoids to dynamically mount the persistent volumes to node-agent pods and cause node-agent pods to restart (this is not accepted since node-agent lose it current state after its pods restart)  
+- This prevents unnecessary full backup. Velero's fs uploaders support file level incremental backup by comparing the file name and metadata. However, at present the files are visited by host path, while pod and PVC's ID are part of the host path, so once the pod is recreated, the same file is regarded as a different file since the pod's ID has been changed. If the fs uploader is in a dedicate pod and files are visited by pod's volume path, files' full path are not changed after pod restarts, so incremental backups could continue.  
+
+## Goals
+- Create a solution to make VGDP instances as micro services
+- Modify the VBDM to launch VGDP micro services instead of running the VGDP instances in place
```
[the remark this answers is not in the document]
Lyndon-Li: Done

design/vgdp-micro-service/vgdp-micro-service.md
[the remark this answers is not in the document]
Lyndon-Li: After the further discussion, we will not use `this-pod`, `this-container` and `this-volume`, instead, we will pass the specific info required by the vgdp-ms --- volume path, volume mode, DUCR/DDCR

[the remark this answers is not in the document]
Lyndon-Li: Done on the changes

design/vgdp-micro-service/vgdp-micro-service.md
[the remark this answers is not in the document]
Lyndon-Li: Added a clarification for the roles/privileges/capabilities of the backupPod/restorePod.
We can open an issue for further changes after this micro service feature is merged.

design/vgdp-micro-service/vgdp-micro-service.md
Lyndon-Li: After a further discussion, we decided to associate the events to DUCR/DDCR

Lyndon-Li: Done on the changes

shawn-hurley:
So my thought process was:

We could use a generic controller to handle the DUCR/DDCR and then move all the data movers to micro services and have the generic controller be able to control what data mover is called when. 

The biggest problem with the design at the moment is that I can only have a single Data Mover. I have to copy the full VBDM code and extend it for my use case (a particular volume for a particular driver). 

Maybe this isn't how you want to solve that problem, and that is ok, but it is an option. 

I would like to see us think through a solution to this problem though.

Lyndon-Li:
> So my thought process was:
> 
> We could use a generic controller to handle the DUCR/DDCR and then move all the data movers to micro services and have the generic controller be able to control what data mover is called when.
> 
> The biggest problem with the design at the moment is that I can only have a single Data Mover. I have to copy the full VBDM code and extend it for my use case (a particular volume for a particular driver).
> 
> Maybe this isn't how you want to solve that problem, and that is ok, but it is an option.
> 
> I would like to see us think through a solution to this problem though.

I think this is possible theoretically, though we still need to cover lots of details, e.g., how do we publish the communication mechanism so that 3rd data movers could implement.
However, the biggest problem here is this conflicts with the existing 3rd data mover integration mechanism that has already been defined in the Volume Snapshot Data Movement design --- the 3rd data mover is integrated through DUCR/DDCR, which is simple and clear.
Therefore, without enough motivation, we don't want to modify the original design. And the benefit I can see for your proposal is that 3rd data mover authors don't need to implement the DUCR/DDCR controllers, but I don't think that is motivative enough. Moreover, I am not sure if all the 3rd data mover authors want this.

Anyway, this is out of the scope of the current design. I suggest to open another issue after this design is merged and we can continue this discussion in the issue.

weshayutin:
Is this design a candidate for v1.15?

Lyndon-Li:
> Is this design a candidate for v1.15?

Yes

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,195 @@
...
+The primary task of VGDP Watcher is to watch the status change from backupPod/restorePod or the VGDP instance, so as to inform the data mover controller in below situations:
+- backupPod/restorePod starts
+- VGDP instance starts
+- Progress update
+- VGDP instance completes/fails/cancelled
+- backupPod/restorePod stops
+
+We use two mechanism to make the watch:  
+**Pod Phases**: VGDP Watcher watches the backupPod/restorePod's phases updated by Kubernetes. That is, VGDP Watcher creates an informer to watch the pod resource for the backupPod/restorePod and detect that the pod reaches to one of the terminated phases (i.e., PodSucceeded, PodFailed). We also check the availability & status of the backupPod/restorePod at the beginning of the watch so as to detect the starting of the backupPod/restorePod.  
+**Custom Kubernetes Events**: VGDP-MS generates Kubernetes events and associates them to the DUCR/DDCR at the time of VGDP instance starting/stopping and progress update, then VGDP Watcher creates another informer to watch the Event resource associated to the DUCR/DDCR.
```
reasonerjt, line 80: @sseago @shawn-hurley Any comments on this approach?  Essentially we intend to use k8s events as a way for communication from the DM pod to node-agent.

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,195 @@
...
+- The termination message of backupPod/restorePod
+- The message along with the VGDP Instance Completes/Fails/Cancelled event  
+
+On the one hand, in some cases only the backupPod/restorePod's termination message is available, e.g., the backupPod/restorePod crashes or or backupPod/restorePod quits before VGDP instance is started. So we refer to the first mechanism to get the notifications.  
+On the other hand, if they are both available, we have the results from them for mutual verification.   
+
+Conclusively, under the help of VGDP Watcher, data mover controller starts VGDP-MS controllably and waits until VGDP-MS ends under any circumstances.   
+
+#### AsyncBR adapter
+VGDP Watcher needs to notify the data mover controller when one of the watched event happens, so that the controller could do the operations as if it receives the same callbacks from VGDP as the current behavior. In order not to break the existing code logics of data mover controllers, we make VGDP Watcher as an adapter of AsyncBR which is the interface implemented by VGDP and called by the data mover controller.
```
reasonerjt, line 95: >  In order not to break the existing code logics of data mover controllers, we make VGDP Watcher as an adapter of AsyncBR which is the interface implemented by VGDP and called by the data mover controller.  

I was a little confused by the sentence here, I think you mean VGDP watcher will **_impelment_** the `AsyncBR` interface to interact with the DM pod, right?

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,195 @@
...
+**VGDP Watcher**: We create a new module to help the data mover controller to watch activities of the VGDP instance in the backupPod/restorePod. VGDP Watcher is a part of VBDM.  
+**VGDP-MS**: VGDP Micro Service is the binary for the command backupPod/restorePod runs. It accepts the parameters and then launches the VGDP instance according to the request type, specifically, backup or restore. VGDP-MS also runs other modules to sync-up with the data mover controller. VGDP-MS is also a part of VBDM.  
+
+Below diagram shows how these components work together:  
+![vgdp-ms-1.png](vgdp-ms-1.png) 
+
+## Detailed Design
+### Exposer
+At present, the exposer creates backupPod/restorePod and sets ```velero-helper pause``` as the command run by backupPod/restorePod.  
+Now, VGDP-MS command will be used, the command is like below:
```
reasonerjt, line 46: Let's clarify that the it will be velero image running in the DM pod.

design/vgdp-micro-service/vgdp-micro-service.md
Lyndon-Li: Yes

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,195 @@
...
+	// Close closes an asynchronous data path instance
+	Close(ctx context.Context)
+}
+```
+Some parameters are hidden into ```param```, but the functions and calling logics are not changed.  
+
+VGDP Watcher should be launched by the data mover controller before VGDP instance starts, otherwise, multiple corner problems may happen. E.g., VGDP-MS may run the VGDP instance immediately after the backupPod/restorePod is launched and completes it before the data mover controller starts VGDP Watcher, as a result, multiple informs are missed from VGDP Watcher.  
+Therefore, the controller launches VGDP Watcher first and then set the DUCR/DDCR to ```InProgress```; on the other hand, VGDP-MS waits DUCR/DDCR turns to ```InProgress``` before running the VGDP instance.  
+
+### VGDP-MS
```
reasonerjt, line 120: Shall we define the failover case here?  I believe if the `velero data-mover` command fails the pod won't restart and the DD/DU CR will marked as `Failed`?

design/vgdp-micro-service/vgdp-micro-service.md
Lyndon-Li: Done

design/vgdp-micro-service/vgdp-micro-service.md
Lyndon-Li: Done

design/vgdp-micro-service/vgdp-micro-service.md
shawn-hurley: when we say events, we are suggesting to create `events.k8s.io/v1` objects tied to the DUCR/DDCR. 

I would worry about this from the watcher if it is taking actions based on the events from [docs](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/event-v1/)

```
Events should be treated as informative, best-effort, supplemental data.
```

Can VGDP-MS also update the status of the DDCR/DUCR? sorry if that is a silly question. Adding this to my queue to look at more closely.

Lyndon-Li: > Events should be treated as informative, best-effort, supplemental data.

Yes, we are using `events.k8s.io/v1` here by stick to this principle. Specifically:

- Events never deliver decisive information. E.g., the state transitions, i.e., data mover starts/stops/cancelled/crashed,  are done by monitor backupPod/restorePod status.
- Events are only used to deliver auxiliary information, i.e., logs for the operation of data movements, progress of the data movements, etc.

However, events are not unnecessary, as explained in the `Dual mode event watch` section.

Lyndon-Li: > Can VGDP-MS also update the status of the DDCR/DUCR?

We have excluded this option from the very beginning. This requires dedicate conflict (when updating DUCR/DDCR) resolution, which makes the controllers' workflow complicated and fragmented.
Therefore, here we follow another principle --- every update to the DDCR/DUCR must be done in the controllers.

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,198 @@
...
+
+As a solution, we use logrus's hook mechanism to redirect the backupPod/restorePod's logs into node-agent's log, so that ```velero debug``` could collect VGDP logs as is without any changes.  
+
+Below diagram shows how VGDP logs are redirected:  
+![vgdp-ms-3.png](vgdp-ms-3.png) 
+
+This log redirecting mechanism is thread safe since the hook acquires the write lock before writing the log buffer, so it guarantees that in the node-agent log there is no corruptions after redirecting the log, and the redirected logs and the original node-agent logs are not projected into each other.     
+
+## node-agent
+node-agent is still required. Even though VGDP is now not running inside node-agent, node-agent still hosts the data mover controller which reconciles DUCR/DDCR and operates DUCR/DDCR in other steps before the VGDP instance is started, i.e., Accept, Expose, etc.
```
sseago, line 178: For the current design, the reason DUCR/DDCR is reconciled in the node agent (and not the velero pod) is because processing the DU/DD required host path access, so it needed to be on the same node as the mounting pod. If we're doing all of the kopia work from the backup/restorePod, then wouldn't it make more sense to move the DU/DD reconcilers to the main velero server?

sseago: If we did this, then node agent would remain privileged always, but if you're using datamover without fs-backup, then you wouldn't need to enable the node agent.

shubham-pampattiwar (commented):
While running VGDP instances as dedicated pods offers better resource control, it may introduce scalability challenges, particularly in large clusters with many backup/restore operations. Each operation will now spawn a new pod, which could lead to a significant increase in resource consumption and potential scheduling delays. Should the design include something on implementing resource quotas and limits specific to VGDP pods to prevent overconsumption of cluster resources ?

design/vgdp-micro-service/vgdp-micro-service.md
Lyndon-Li: Moving the DUCR/DDCR controller to the backupPod/restorePod is feasible technically, but we didn't choose this for below reasons:
1. Making the DUCR/DDCR controller outside of the backup/restorePod helps to better control the data movement workflow. We have one part in one module and the other part in the other module, this is something like the concept of "multiple failure domains", so when one part fails, we always get some way to be notified and update the DUCR/DDCR
2. Since we want the DUCR/DDCR controller outside of the backup/restorePod, we have two choices --- velero server and node-agent server. We didn't choose the velero server primarily because of the isolation, that is, we want to isolate the control path and the data path; we also want to isolate the resource backup and the volume data backup
3. As another concept we are following, we want to run the as minimal part into the backupPod/restorePod as possible, so it the design we actually call it `VGDP micro service`, that is we only run the VGDP in the backupPod/restorePod. That also adheres to the concept of micro service

design/vgdp-micro-service/vgdp-micro-service.md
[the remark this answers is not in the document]
Lyndon-Li: Done

design/vgdp-micro-service/vgdp-micro-service.md
```diff
@@ -0,0 +1,186 @@
...
+Then we need to change the command the backupPod/restorePod is running, the command launches VGDP-MS (VGDP Micro Service, see below) when the container starts up.  
+For CSI snapshot, the backupPod/restorePod is created as the result of expose, the only thing left is to change the backupPod/restorePod's image.  
+**VBDM**: VGDM contains the data mover controller, while the controller calls the Exposer and launches the VGDP instances.  Now, since the VGDP instance is launched by the backupPod/restorePod, the controller should not launch the VGDP instance again. However, the controller still needs to monitor and control the VGDP instance. Moreover, in order to avoid any contest situations, the controller is still the only place to update DUCRs and DDCRs.
```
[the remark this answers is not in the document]
Lyndon-Li: Done

Lyndon-Li:
> While running VGDP instances as dedicated pods offers better resource control, it may introduce scalability challenges, particularly in large clusters with many backup/restore operations. Each operation will now spawn a new pod, which could lead to a significant increase in resource consumption and potential scheduling delays. Should the design include something on implementing resource quotas and limits specific to VGDP pods to prevent overconsumption of cluster resources ?

For a backup/restore with thousands of volumes, if there is no control, there will be thousands of backupPods/restorePods.
Actually, we already have the control -- we can use the existing node-agent concurrency feature to control how many VGDP instances running in one node (configurable and default value is 1), this also works for VGDP micro service.
Noticeably,  this is to control by nodes, at least 1 VGDP in a node. But I think that should be enough, if a cluster has thousands of nodes, I believe the API server/etcd should have already scaled out and so is powerful enough to handle several pods in each node.

This is a good point, I will add the same clarification to the design.

Lyndon-Li:
Done all the comments, please continue to review.

Lyndon-Li:
@sseago @shubham-pampattiwar Any more comments on this PR? Once you finish the review, please help to approve it.

shawn-hurley:
All good from my perspective.

I do still think that this microservice approach, gives us a chance to extend this design and make the Datamovers more pluggable. I think that this would be a huge win but makes sense as a follow on

sseago (approved):

shubham-pampattiwar (approved):

PR #7597, opened 2024-03-31
fix the resticIdentifier value

muxuelanKK:
1、while repositoryType is "kopia"，resticIdentifier value should match
2、for s3 such as minio，url shoud be http://s3_url/buckets/bucketName/..., not http://s3_url/bucketName/...

image


muxuelanKK:
> Have you considered creating a dedicated type for this purpose ?

yes, hold on

kaovilai:
resticIdentifier is there for backwards compatibility and not used by kopia file system backup process. It's an ignored field for kopia afaict.

Lyndon-Li:
Kopia path doesn't use repoIdentifier, and so this field will be removed/reused after Restic path is totally deprecated. Therefore, we can hold on the changes to repoIdentifier until we know how to use /whether to remove it after Restic path is deprecated.

kaovilai:
since https://github.com/vmware-tanzu/velero/pull/8987 merged, it's clear we do not need this.

PR #7599, opened 2024-03-31
modify the default value to match helm chart

muxuelanKK:
In helm chart, we set some value such as restoreOnlyMode, it doesn't work.

image

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -88,6 +88,7 @@ const (
 
 	defaultBackupSyncPeriod           = time.Minute
 	defaultStoreValidationFrequency   = time.Minute
+	defaultGarbageCollectionFrequency = time.Hour
 	defaultPodVolumeOperationTimeout  = 240 * time.Minute
 	defaultResourceTerminatingTimeout = 10 * time.Minute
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/server/server.go
```diff
@@ -88,6 +88,7 @@ const (
 
 	defaultBackupSyncPeriod           = time.Minute
 	defaultStoreValidationFrequency   = time.Minute
+	defaultGarbageCollectionFrequency = time.Hour
```
kaovilai, line 91: Would it make sense to only have one defined default?

Here's another locatiion.
https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/gc_controller.go#L41

pkg/cmd/server/server.go
```diff
@@ -158,6 +159,8 @@ func NewCommand(f client.Factory) *cobra.Command {
 			storeValidationFrequency:       defaultStoreValidationFrequency,
 			podVolumeOperationTimeout:      defaultPodVolumeOperationTimeout,
 			restoreResourcePriorities:      defaultRestorePriorities,
+			restoreOnly:                    false,
```
qiuming-best, line 162: If `restoreOnly` is default not set， the value would be false.
So do we need add this line?

PR #7600, opened 2024-03-31
make repository password configurable

muxuelanKK:
The repository password should be configurable.


muxuelanKK:
> repository-password



> > Note: Currently, Velero creates a secret named velero-repo-credentials in the velero install namespace, containing a default backup repository password. You can update the secret with your own password encoded as base64 prior to the first backup (i.e., FS Backup, data mover) targeting to the backup repository. The value of the key to update is
> > ```
> > data:
> >   repository-password: <custom-password>
> > ```
> 
> It's [already configurable today](https://velero.io/docs/v1.13/file-system-backup/#:~:text=Note%3A%20Currently,password%3A%20%3Ccustom%2Dpassword%3E).
> 
> If you want to enhance the velero server CLI, consider also adding the option to the `velero install` command.

Now we only can edit the secret after velero deployed already, I alse  want to add the option to the `velero install` command.

Lyndon-Li:
Just changing the secret doesn't work in many cases. E.g., when some existing repos/BSLs are there, when changing the secret, Velero will use the new password to connect to the repos, however, the repos themselves are still with the old password, so the connection will fail.

Actually, supporting dynamically changing the repo password involves many fundamental changes, as discussed in #5443. Therefore, we will need a sophisticated design to cover all the factors, i.e., repo password changing and unified repo API, password preserve, etc.

PR #7664, opened 2024-04-12
Extend Volume Policies feature to support more actions

shubham-pampattiwar:

Implementation PR for Extend Volume Policies feature to support more actions

Related design: https://github.com/vmware-tanzu/velero/pull/6956

Fixes https://github.com/vmware-tanzu/velero/issues/6640

### DIFF
--- a/internal/resourcepolicies/resource_policies.go
+++ b/internal/resourcepolicies/resource_policies.go
@@ -27,8 +27,13 @@ type VolumeActionType string
 
 const (
 	// currently only support configmap type of resource config
-	ConfigmapRefType string           = "configmap"
-	Skip             VolumeActionType = "skip"
+	ConfigmapRefType string = "configmap"
+	// skip action implies the volume would be skipped from the backup operation
+	Skip VolumeActionType = "skip"
+	// fs-backup action implies that the volume would be backed up via file system copy method using the uploader(kopia/restic) configured by the user
+	FSBackup VolumeActionType = "fs-backup"
+	// snapshot action can have 3 different meaning based on velero configuration and backup spec - cloud provider based snapshots, local csi snapshots and datamover snapshots
+	Snapshot VolumeActionType = "snapshot"
 )
 
 // Action defined as one action for a specific way of backup
@@ -40,16 +45,16 @@ type Action struct {
 }
 
 // volumePolicy defined policy to conditions to match Volumes and related action to handle matched Volumes
-type volumePolicy struct {
+type VolumePolicy struct {
 	// Conditions defined list of conditions to match Volumes
 	Conditions map[string]interface{} `yaml:"conditions"`
 	Action     Action                 `yaml:"action"`
 }
 
 // resourcePolicies currently defined slice of volume policies to handle backup
-type resourcePolicies struct {
+type ResourcePolicies struct {
 	Version        string         `yaml:"version"`
-	VolumePolicies []volumePolicy `yaml:"volumePolicies"`
+	VolumePolicies []VolumePolicy `yaml:"volumePolicies"`
 	// we may support other resource policies in the future, and they could be added separately
 	// OtherResourcePolicies []OtherResourcePolicy
 }
@@ -60,8 +65,8 @@ type Policies struct {
 	// OtherPolicies
 }
 
-func unmarshalResourcePolicies(yamlData *string) (*resourcePolicies, error) {
-	resPolicies := &resourcePolicies{}
+func unmarshalResourcePolicies(yamlData *string) (*ResourcePolicies, error) {
+	resPolicies := &ResourcePolicies{}
 	err := decodeStruct(strings.NewReader(*yamlData), resPolicies)
 	if err != nil {
 		return nil, fmt.Errorf("failed to decode yaml data into resource policies  %v", err)
@@ -69,7 +74,7 @@ func unmarshalResourcePolicies(yamlData *string) (*resourcePolicies, error) {
 	return resPolicies, nil
 }
 
-func (p *Policies) buildPolicy(resPolicies *resourcePolicies) error {
+func (p *Policies) BuildPolicy(resPolicies *ResourcePolicies) error {
 	for _, vp := range resPolicies.VolumePolicies {
 		con, err := unmarshalVolConditions(vp.Conditions)
 		if err != nil {
@@ -162,7 +167,7 @@ func GetResourcePoliciesFromConfig(cm *v1.ConfigMap) (*Policies, error) {
 	}
 
 	policies := &Policies{}
-	if err := policies.buildPolicy(resPolicies); err != nil {
+	if err := policies.BuildPolicy(resPolicies); err != nil {
 		return nil, errors.WithStack(err)
 	}
 
--- a/internal/volumehelper/volume_policy_helper.go
+++ b/internal/volumehelper/volume_policy_helper.go
@@ -1,0 +1,197 @@
+package volumehelper
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/vmware-tanzu/velero/internal/resourcepolicies"
+
+	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
+	kubeutil "github.com/vmware-tanzu/velero/pkg/util/kube"
+
+	"github.com/sirupsen/logrus"
+	corev1api "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/vmware-tanzu/velero/pkg/kuberesource"
+	pdvolumeutil "github.com/vmware-tanzu/velero/pkg/util/podvolume"
+)
+
+type VolumeHelper interface {
+	GetVolumesForFSBackup(pod *corev1api.Pod, defaultVolumesToFsBackup, backupExcludePVC bool, kbclient kbclient.Client) ([]string, []string, error)
+	ShouldPerformSnapshot(obj runtime.Unstructured, groupResource schema.GroupResource, kbclient kbclient.Client) (bool, error)
+}
+
+type VolumeHelperImpl struct {
+	VolumePolicy    *resourcepolicies.Policies
+	SnapshotVolumes *bool
+	Logger          logrus.FieldLogger
+}
+
+func (v *VolumeHelperImpl) ShouldPerformSnapshot(obj runtime.Unstructured, groupResource schema.GroupResource, kbclient kbclient.Client) (bool, error) {
+	// check if volume policy exists and also check if the object(pv/pvc) fits a volume policy criteria and see if the associated action is snapshot
+	// if it is not snapshot then skip the code path for snapshotting the PV/PVC
+	pvc := new(corev1api.PersistentVolumeClaim)
+	pv := new(corev1api.PersistentVolume)
+	var err error
+
+	if groupResource == kuberesource.PersistentVolumeClaims {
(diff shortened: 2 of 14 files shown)


internal/volumehelper/volume_policy_helper.go
```diff
@@ -0,0 +1,243 @@
...
+	FSBackupNonActionMatchingVols := []string{}
+	NoActionMatchingVols := []string{}
+
+	volsToExclude := pdvolumeutil.GetVolumesToExclude(pod)
+	for _, vol := range pod.Spec.Volumes {
+		if !v.ShouldIncludeVolumeInBackup(vol, volsToExclude) {
+			continue
+		}
+		// don't backup volumes that are included in the exclude list.
+		if pdvolumeutil.Contains(volsToExclude, vol.Name) {
```
qiuming-best, line 129: Is the condition `pdvolumeutil.Contains` in this line of code duplicated with the same logic in `ShouldIncludeVolumeInBackup` function? or it could not execute Line 130 for it will always execute  L126 first if the matching condition?

internal/volumehelper/volume_policy_helper.go
```diff
@@ -0,0 +1,243 @@
...
+		includedVolumes, optedOutVolumes, volsToProcessByLegacyApproach, err = v.GetVolumesMatchingFSBackupAction(pod, v.VolumePolicy)
+		if err != nil {
+			return includedVolumes, optedOutVolumes, err
+		}
+	}
+
+	// process legacy annotation based approach, this will be done for 2 scenarios:
+	// 1. volume policy not specified by the user
+	// 2. there are some volumes for which the volume policy approach did not get any supported matching actions
+	if v.VolumePolicy == nil || len(volsToProcessByLegacyApproach) > 0 {
```
qiuming-best, line 107: if v.VolumePolicy == nil then len(volsToProcessByLegacyApproach) will always be 0, so should we remove the condition `len(volsToProcessByLegacyApproach) > 0`?

internal/resourcepolicies/resource_policies.go
```diff
@@ -29,6 +29,8 @@ const (
 	// currently only support configmap type of resource config
 	ConfigmapRefType string           = "configmap"
 	Skip             VolumeActionType = "skip"
+	FSBackup         VolumeActionType = "fs-backup"
+	Snapshot         VolumeActionType = "snapshot"
```
kaovilai, line 33: Can you comment in code here the meaning of snapshot?

From the design following were also mentioned, it might be hard to distinguish between them all without explaining what snapshot above means as it relates to native, CSI, or data mover enablement.

> - `snapshot-native`: only use volume snapshotter (native cloud provider snapshots), do nothing if not present/not compatible
> - `snapshot-csi`: only use csi-plugin, don't use volume snapshotter(native cloud provider snapshots), don't use datamover even if snapshotMoveData is true
> - `snapshot-datamover`: only use csi with datamover, don't use volume snapshotter (native cloud provider snapshots), use datamover even if snapshotMoveData is false

kaovilai (changes requested):

internal/resourcepolicies/resource_policies.go
sseago: @kaovilai "snapshot" means that velero will either make a CSI snapshot or a native snapshot -- choosing between them will use the same logic Velero currently uses for non-fs-backup volumes when deciding whether to use CSI or native snapshots. Data mover enablement is a separate decision -- if a CSI snapshot is created and snapshotMoveData=true, then data movement will occur for this volume.

The three potential future values will override the default behavior described above. 
- `snapshot-native` will force using the native volume snapshotter, even if the volume would otherwise use CSI.
- `snapshot-csi` will force using CSI, even if the volume would otherwise use native snapshots (although I don't think there are currently any scenarios where CSI is enabled and available for this volume where the default behavior is native snapshots), and it will prevent data movement of the volume, even if `snapshotMoveData=true`
- `snaphot-datamover` will force using CSI (if available for the volume) and then use data movement, even if `snapshotMoveData=false`

pkg/backup/item_backupper.go
```diff
@@ -504,22 +532,34 @@ const (
...
+		return err
+	}
+
+	if !snapshotVolume {
+		log.Info(fmt.Sprintf("skipping volume snapshot for PV %s as it does not fit the volume policy criteria specified by the user for snapshot action", pv.Name))
+		ib.trackSkippedPV(obj, kuberesource.PersistentVolumes, volumeSnapshotApproach, "does not satisfy the criteria for volume policy based snapshot action", log)
+		return nil
+	}
+
+	if boolptr.IsSetToFalse(ib.backupRequest.Spec.SnapshotVolumes) {
```
Lyndon-Li, line 557: Since `ib.backupRequest.Spec.SnapshotVolumes` has been checked inside `ShouldPerformSnapshot`, can we remove it here?

internal/volumehelper/volume_policy_helper.go
```diff
@@ -0,0 +1,245 @@
...
+		includeVolumeInBackup = false
+	}
+	// don't include volumes that mount the default service account token.
+	if strings.HasPrefix(vol.Name, "default-token") {
+		includeVolumeInBackup = false
+	}
+	return includeVolumeInBackup
+}
+
+func (v *VolumeHelperImpl) GetPVCForPodVolume(vol corev1api.Volume, pod *corev1api.Pod) (*corev1api.PersistentVolumeClaim, error) {
```
Lyndon-Li, line 208: This looks like a common function, can we add this into util package?

internal/resourcepolicies/resource_policies.go
kaovilai: Right. I just kind of want that in the code comment.

Lyndon-Li:
@shubham-pampattiwar Could help to address the comments so that we can further review and merge it? To meet FC, we hope this could be merged by the end of this week or early next week.

internal/resourcepolicies/resource_policies.go
shubham-pampattiwar: Done, added the relevant comments for action types

internal/volumehelper/volume_policy_helper.go
```diff
@@ -0,0 +1,245 @@
...
+	DefaultVolumesToFsBackup bool
+	SnapshotVolumes          *bool
+	Logger                   logrus.FieldLogger
+	DynamicFactory           client.DynamicFactory
+}
+
+func (v *VolumeHelperImpl) ShouldPerformSnapshot(obj runtime.Unstructured, groupResource schema.GroupResource) (bool, error) {
+	// check if volume policy exists and also check if the object(pv/pvc) fits a volume policy criteria and see if the associated action is snapshot
+	// if it is not snapshot then skip the code path for snapshotting the PV/PVC
+	if groupResource == kuberesource.PersistentVolumeClaims {
```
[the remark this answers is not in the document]
shubham-pampattiwar: Done, consolidated

internal/volumehelper/volume_policy_helper.go
```diff
@@ -0,0 +1,245 @@
...
+			}
+			return false, nil
+		}
+	}
+
+	// now if volumepolicy is not specified then just check for snapshotVolumes flag
+	if boolptr.IsSetToTrue(v.SnapshotVolumes) {
+		return true, nil
+	}
+	return false, nil
```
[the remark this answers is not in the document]
shubham-pampattiwar: Done, added the logs

internal/volumehelper/volume_policy_helper.go
```diff
@@ -0,0 +1,245 @@
...
+
+	pvc := new(corev1api.PersistentVolumeClaim)
+	if err = runtime.DefaultUnstructuredConverter.FromUnstructured(pvcObject.UnstructuredContent(), pvc); err != nil {
+		return nil, err
+	}
+
+	return pvc, nil
+}
+
+func (v *VolumeHelperImpl) GetPVForPVC(pvc *corev1api.PersistentVolumeClaim) (*corev1api.PersistentVolume, error) {
```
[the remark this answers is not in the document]
shubham-pampattiwar: The existing one uses controller-runtime client, Could not use controller runtime client, was getting unexpected results (pvc existed in cluster but was getting nil), hence this separate function. Hope thats ok.

pkg/backup/item_backupper.go
shubham-pampattiwar: Done

internal/volumehelper/volume_policy_helper.go
shubham-pampattiwar: Done

shubham-pampattiwar:
@Lyndon-Li addressed PR comments.

internal/volumehelper/volume_policy_helper.go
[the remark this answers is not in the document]
Lyndon-Li: @shubham-pampattiwar 
I don't understand why the same function in util/pvc_pv.go returns nil for the current case, could you explain more?
Moreover, I don't think it is a good practice to generate a client in a helper function, is there any way to refactor the code?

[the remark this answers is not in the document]
blackpiglet: Using Dynamic Client here should work as expected, but it seems unnecessary.
Both the controller manager's internal client and the Velero client factory's controller-runtitme client should work here.

Lyndon-Li:
@shubham-pampattiwar
Two more following up comments from me and @blackpiglet, see if you could address it. Basically, we suggest to avoid generating a dynamic client inside the helper, but get it from the backup as an input parameter and then call the existing function in util/pvc_pv.go.

Lyndon-Li:
@shubham-pampattiwar 
We also see the code coverage has dropped by 0.24% due to this PR, could you help to add some more test case?

internal/volumehelper/volume_policy_helper.go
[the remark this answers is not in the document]
shubham-pampattiwar: Done, updated the PR

shubham-pampattiwar:
@Lyndon-Li @blackpiglet Addressed PR feedback and added tests. PTAL, Thanks !

Lyndon-Li (approved):

blackpiglet (dismissed):

internal/volumehelper/volume_policy_helper.go
qiuming-best: @shubham-pampattiwar please take a look before merging this PR

qiuming-best: Would it should be like below?
```golang
	for i, vol := range pod.Spec.Volumes {
		if !v.ShouldIncludeVolumeInBackup(vol, volsToExclude, backupExcludePVC) {
		       // don't backup volumes that are included in the exclude list.
		        if util.Contains(volsToExclude, vol.Name) {
			     FSBackupNonActionMatchingVols = append(FSBackupNonActionMatchingVols, vol.Name)
			     }
			continue
		}
		
		if vol.PersistentVolumeClaim != nil {}
```

shubham-pampattiwar: Actually I have removed the `volsToExclude` check from the volume policy processing logic as it is related to legacy annotations based approach.
To Clarify, Volume Policy existence will be checked first and processed first. The legacy annotations based fallback option will be used only in 2 cases:
- If no Volume Policy is specified by the user: [Ref code link from PR](https://github.com/vmware-tanzu/velero/pull/7664/files#diff-412f1e7719b8fbd177428c1cb421a4b1caef8e0c33d78f7b83f6dcbafae87053R114)
- If there are Volumes for which no Volume Policy action was found i.e. action was nil : [Ref code link from PR](https://github.com/vmware-tanzu/velero/pull/7664/files#diff-412f1e7719b8fbd177428c1cb421a4b1caef8e0c33d78f7b83f6dcbafae87053R102)
I have added comments in the code for the above process.

qiuming-best (approved):

Lyndon-Li (approved):

Issue #7782, opened 2024-05-08
Volume Policy - miss match behavior with Velero for SnapshotVolume


Lyndon-Li:
For `--snapshot-volume` flag, if it is not set, Velero's existing behavior treat SnapshotVolume as true, that is, volumes are backed up by CSI snapshot (data movement) or native snapshot; However, for volume policy, if the flag is not set, it disable the `snapshot` action even if the condition is met.

Steps:

- Create a volume policy configuration with a snapshot action and a condition meeting a volume being backed up
- Run a Velero backup using the volume policy but don't set `--snapshot-volume` flag explictily
- [Result] The volume is still skipped by volume policy even though the condition is met

Lyndon-Li:
@shubham-pampattiwar 
The issue is about a mismatching behavior to check `--snapshot-volume` flag by volume policy.

However, I discussed further with @reasonerjt for this issue, we think volume policy may not need to respect  `--snapshot-volume` flag, since volume policy always take the preference, if the condition is met which means volume policy voting the snapshot action, the backup should go with snapshot action not matter how `snapshot-volume` is set.
So we may need to change the existing code of Velero where `snapshot-volume` is being checked.

shubham-pampattiwar:
@Lyndon-Li I think not respecting `snapshotVolmes` as `true` may not be the best idea here. The CSI plugin depends on this value to perform snapshots: https://github.com/vmware-tanzu/velero/blob/0d85a647b5a4d3b812d3669056eff3a0f539e282/pkg/backup/actions/csi/pvc_action.go#L68

Lyndon-Li:
> @Lyndon-Li I think not respecting `snapshotVolmes` as `true` may not be the best idea here. The CSI plugin depends on this value to perform snapshots:
> 
> https://github.com/vmware-tanzu/velero/blob/0d85a647b5a4d3b812d3669056eff3a0f539e282/pkg/backup/actions/csi/pvc_action.go#L68

@shubham-pampattiwar Actually the idea we discussed yesterday is that once the volume policy chooses snapshot for a volume, any other places (including CSI plugin) should not check for the `SnapshotVolumes` again. This introduces more changes than the code for volume policy but it makes things more clear.

shubham-pampattiwar:
Created a PR for this issue: https://github.com/vmware-tanzu/velero/pull/7786

Lyndon-Li:
Fixed by #7794

PR #7786, opened 2024-05-09
Remove snapshotVolumes flag necessity for volume policy snapshot action

shubham-pampattiwar:

Remove snapshotVolumes flag necessity for volume policy snapshot action 

Fixes https://github.com/vmware-tanzu/velero/issues/7782

### DIFF
--- a/changelogs/unreleased/7786-shubham-pampattiwar
+++ b/changelogs/unreleased/7786-shubham-pampattiwar
@@ -1,0 +1,1 @@
+Remove snapshotVolumes flag necessity for volume policy snapshot action
--- a/internal/volumehelper/volume_policy_helper.go
+++ b/internal/volumehelper/volume_policy_helper.go
@@ -8,7 +8,6 @@ import (
 
 	kbclient "sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
 	kubeutil "github.com/vmware-tanzu/velero/pkg/util/kube"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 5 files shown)


Lyndon-Li:
@shubham-pampattiwar 
Thanks for the PR, we are making good progress on the understanding of the problem. But I am afraid there are still gaps to fix it:
1. We cannot simply check `VolumePolicy != nil` when we want to ignore `snapshotVolume`. `VolumePolicy != nil` means there are some conditions, however, it doesn't mean these conditions vote for snapshot action. Instead, I think we need to call `VolumeHelperImpl.ShouldPerformSnapshot`
2. Besides backup, we have some places in restore that checks `snapshotVolume`, i.e., https://github.com/vmware-tanzu/velero/blob/7563a453fb962382c7b1c456ddb2a6645cdd9493/pkg/restore/pv_restorer.go#L56. We need to do some test to decide if we can simply remove it or use more reliable way to check
3. For backup, we have some more places to cover:

https://github.com/vmware-tanzu/velero/blob/7563a453fb962382c7b1c456ddb2a6645cdd9493/pkg/backup/item_backupper.go#L410
https://github.com/vmware-tanzu/velero/blob/7563a453fb962382c7b1c456ddb2a6645cdd9493/pkg/controller/backup_controller.go#L508
4. Looks like this is a gap during the design discussion, as I see the design is aligned with the current code, so we also need to change the design.

Lyndon-Li:
@shubham-pampattiwar 
One more follow up for 1 in the above comment:
If no matching condition for snapshot action and `snapshotVolume` is set to true, we still need to check `defaultVolumesToFsBackup` flag to decide whether `ShouldPerformSnapshot` should return true. Am I right?

shubham-pampattiwar:
closing this PR in favor of https://github.com/vmware-tanzu/velero/pull/7794
Thanks @blackpiglet 
cc: @Lyndon-Li @reasonerjt

PR #7813, opened 2024-05-22
feat: add ingressclass from ingress

AlbeeSo:

To backup / restore ingressclass if ingress is backuped and spec.ingressClass is not nil.

After PR, ingressclass will be backuped with logs like
```
time="2024-05-22T17:48:36+08:00" level=info msg="Executing custom action" backup=velero/test-velero-ing logSource="pkg/backup/item_backupper.go:371" name=test-ing namespace=default resource=ingresses.networking.k8s.io
time="2024-05-22T17:48:36+08:00" level=info msg="Executing ingressAction" backup=velero/test-velero-ing cmd=/velero logSource="pkg/backup/actions/ingress_action.go:51" pluginName=velero
time="2024-05-22T17:48:36+08:00" level=info msg="Adding IngressClass nginx to additionalItems" backup=velero/test-velero-ing cmd=/velero logSource="pkg/backup/actions/ingress_action.go:61" pluginName=velero
time="2024-05-22T17:48:36+08:00" level=info msg="Done executing ingressAction" backup=velero/test-velero-ing cmd=/velero logSource="pkg/backup/actions/ingress_action.go:68" pluginName=velero
```
And will be restored with logs like
```
time="2024-05-22T17:59:30+08:00" level=info msg="Executing item action for ingresses.networking.k8s.io" logSource="pkg/restore/restore.go:1319" restore=velero/restore-ing
time="2024-05-22T17:59:30+08:00" level=info msg="Executing AddIngressClassFromIngAction" cmd=/velero logSource="pkg/restore/actions/add_ingressclass_from_ing_action.go:44" pluginName=velero restore=velero/restore-ing
time="2024-05-22T17:59:30+08:00" level=info msg="Adding IngressClass nginx as an additional item to restore" cmd=/velero logSource="pkg/restore/actions/add_ingressclass_from_ing_action.go:54" pluginName=velero restore=velero/restore-ing
```

Fixes #(issue)

### DIFF
--- a/changelogs/unreleased/7813-AlbeeSo
+++ b/changelogs/unreleased/7813-AlbeeSo
@@ -1,0 +1,1 @@
+Add ingressClass resource to backup or restore if ingress is backuped and its spec.ingressClass is not nil.
--- a/pkg/backup/actions/ingress_action.go
+++ b/pkg/backup/actions/ingress_action.go
@@ -58,7 +58,7 @@ func (a *IngressAction) Execute(item runtime.Unstructured, backup *v1.Backup) (r
 
 	var additionalItems []velero.ResourceIdentifier
 	if ing.Spec.IngressClassName != nil && *ing.Spec.IngressClassName != "" {
-		a.log.Infof("Adding IngressClass %s to additionalItems", ing.Spec.IngressClassName)
+		a.log.Infof("Adding IngressClass %s to additionalItems", *ing.Spec.IngressClassName)
 		additionalItems = append(additionalItems, velero.ResourceIdentifier{
 			GroupResource: kuberesource.IngressClasses,
 			Name:          *ing.Spec.IngressClassName,
--- a/pkg/backup/actions/ingress_action.go
+++ b/pkg/backup/actions/ingress_action.go
@@ -1,0 +1,69 @@
+/*
+Copyright the Velero contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package actions
+
+import (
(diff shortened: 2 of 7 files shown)


kaovilai (approved):

reasonerjt:
IMHO, this should remain an "optional" plugin, b/c there may be other cases, where the same ingress will use a different ingressclass, or the same ingressclass like "default" may have different settings across different clusters.
We should be consistent with the way velero handles storageclass, and currently, we don't add storageclass automatically.

AlbeeSo:
> IMHO, this should remain an "optional" plugin, b/c there may be other cases, where the same ingress will use a different ingressclass, or the same ingressclass like "default" may have different settings across different clusters. We should be consistent with the way velero handles storageclass, and currently, we don't add storageclass automatically.

'default' may be different, but if 'default' already exists it will skip to restore by default, right? 
Im curious why storageclass or ingressclass (which is different with volumesnapshotclass) is not automatically added during backup. As usually not-found errors will occur in such cases if they do not exist in restore cluster.  Or the not-found error is expected and users need to adjust it by hand?

AlbeeSo:
> @AlbeeSo As for StorageClass in particular, those are generally configured by the cluster admin. Simply creating a StorageClass kubernetes resource without configuring the storage behind the storageclass wouldn't be helpful. If you are restoring a backup to a cluster where the storageclass for your PVCs doesn't exist, then you'd either need to first get the admin to create and configure that storageclass, or if there's another appropriate storageclass available, then you can use the storageclass mapping feature on restore to use storage already available.



> @AlbeeSo As for StorageClass in particular, those are generally configured by the cluster admin. Simply creating a StorageClass kubernetes resource without configuring the storage behind the storageclass wouldn't be helpful. If you are restoring a backup to a cluster where the storageclass for your PVCs doesn't exist, then you'd either need to first get the admin to create and configure that storageclass, or if there's another appropriate storageclass available, then you can use the storageclass mapping feature on restore to use storage already available.

Got it. So will action like "velero.io/change-ingress-class" be more suitable for this case, or restoring ingress class directly is enough?   @reasonerjt @sseago

AlbeeSo:
IMHO, restoring the ingressclass is enough, and I can add the "velero.io/change-ingress-class" to allow users to change it if need.

reasonerjt:
"velero.io/change-ingress-class" will not be needed as velero now supports resource modifier to modify restored resources in a more general way.

reasonerjt:
Closing this one, given resource modifier will convert this case and a new RIA is not needed.

PR #7863, opened 2024-06-05
Mark InProgress backup/restore as failed upon requeuing

### LINKED ISSUE #7207: Velero didn't retry on failed Restore CR status update, causing the CR to remain stuck in "InProgress"

Velero was performing a restore when the API server was rolling out to a new version. It had trouble connecting to the API server, but eventually, the restore was successful. However, since the API server was still in the middle of rolling out, Velero failed to update the restore CR status and gave up. After the connection was restored, it didn't attempt to update, causing the restore CR to be stuck at "In progress" indefinitely. This can lead to incorrect decisions for other components that rely on the backup/restore CR status to determine completion. 

Velero controller to reconcile on the `InProgress` CR and and then appropriately update its status to `Complete`.

_If you are using velero v1.7.0+:_  
Please use `velero debug  --backup <backupname> --restore <restorename>` to generate the support bundle, and attach to this issue, more options please refer to `velero debug --help` 

_If you are using earlier versions:_  
Please provide the output of the following commands (Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)

Velero logs:
```
time="2023-12-08T04:02:21Z" level=warning msg="Cluster resource restore warning: could not restore, CustomResourceDefinition \"klusterlets.operator.open-cluster-management.io\" already exists. Warning: the in-cluster version is different than the backed-up version."  logSource="/remotesource/velero/app/pkg/controller/restore_controller.go:506" restore=openshift-adp/acm-klusterlet
time="2023-12-08T04:02:21Z" level=warning msg="Cluster resource restore warning: refresh discovery after restoring CRDs: Get \"
[https://172.30.0.1:443/api?timeout=32s|https://172.30.0.1/api?timeout=32s]
\": dial tcp 172.30.0.1:443: connect: connection refused" logSource="/remote-source/velero/app/pkg/controller/restore_controller.go:506" restore=openshift-adp/acm-klusterlet

################Restore completed
time="2023-12-08T04:02:21Z" level=info msg="restore completed" logSource="/remote-source/velero/app/pkg/controller/restore_controller.go:513" restore=openshift-adp/acm-klusterlet

time="2023-12-08T04:02:21Z" level=error msg="Get \"
[https://172.30.0.1:443/api?timeout=32s|https://172.30.0.1/api?timeout=32s]
\": dial tcp 172.30.0.1:443: connect: connection refused" logSource="/remote-source/velero/app/pkg/datamover/datamover.go:143"
time="2023-12-08T04:02:21Z" level=error msg="Error removing VSRs after partially failed restore" error="Get \"
[https://172.30.0.1:443/api?timeout=32s|https://172.30.0.1/api?timeout=32s]
\": dial tcp 172.30.0.1:443: connect: connection refused" logSource="/remote-source/velero/app/pkg/controller/restore_controller.go:571"

################FAIL to update restore CR status
time="2023-12-08T04:02:21Z" level=info msg="Error updating restore's final status" Restore=openshift-adp/acm-klusterlet error="Patch \"
[https://172.30.0.1:443/apis/velero.io/v1/namespaces/openshift-adp/restores/acm-klusterlet|https://172.30.0.1/apis/velero.io/v1/namespaces/openshift-adp/restores/acm-klusterlet]
\": dial 
tcp 172.30.0.1:443: connect: connection refused" error.file="/remote-source/velero/app/pkg/controller/restore_controller.go:216" error.function="github.com/vmware-tanzu/velero/pkg/controller.(*restoreReconciler).Reconcile" logSource="/remote-source/velero/app/pkg/controller/restore_controller.go:216"
...
################connection is back and it's stable
time="2023-12-08T04:02:58Z" level=info msg="Validating BackupStorageLocation" backup-storage-location=openshift-adp/oadp-2 controller=backup-storage-location logSource="/remote-source/velero/app/pkg/controller/backup_storage_location_controller.go:152"
time="2023-12-08T04:02:58Z" level=info msg="BackupStorageLocations is validc, marking as available" backup-storage-location=openshift-adp/oadp-2 controller=backup-storage-location logSource="/remote-source/velero/app/pkg/controller/backup_storage_location_controller.go:137"
```

Restore CR stuck at `InProgress` status:
```
apiVersion: velero.io/v1
kind: Restore
metadata:
  annotations:
    lca.openshift.io/apply-wave: "1"
  creationTimestamp: "2023-12-08T04:01:59Z"
  generation: 4
  labels:
    velero.io/storage-location: default
  name: acm-klusterlet
  namespace: openshift-adp
  resourceVersion: "45514"
  uid: 392ad1c8-2a1a-4228-9eaa-d3cb28101de3
spec:
  backupName: acm-klusterlet
  excludedResources:
  - nodes
  - events
  - events.events.k8s.io
  - backups.velero.io
  - restores.velero.io
  - resticrepositories.velero.io
  - csinodes.storage.k8s.io
  - volumeattachments.storage.k8s.io
  - backuprepositories.velero.io
  hooks: {}
  itemOperationTimeout: 1h0m0s
status:
  phase: InProgress
  progress:
    itemsRestored: 1
    totalItems: 63
  startTimestamp: "2023-12-08T04:01:59Z"
```

**Environment:**

**Vote on this issue!**

kaovilai:

Fixes #7207

### DIFF
--- a/pkg/controller/backup_controller.go
+++ b/pkg/controller/backup_controller.go
@@ -230,6 +230,20 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	switch original.Status.Phase {
 	case "", velerov1api.BackupPhaseNew:
 		// only process new backups
+	case velerov1api.BackupPhaseInProgress:
+		// if backup is in progress, we should not process it again
+		// we want to mark it as failed to avoid it being stuck in progress
+		// if so, mark it as failed, last loop did not successfully complete the backup
+		log.Debug("Backup has in progress status from prior reconcile, marking it as failed")
+		failedCopy := original.DeepCopy()
+		failedCopy.Status.Phase = velerov1api.BackupPhaseFailed
+		failedCopy.Status.FailureReason = "Backup from previous reconcile still in progress"
+		if err := kubeutil.PatchResource(original, failedCopy, b.kbClient); err != nil {
+			// return the error so the status can be re-processed; it's currently still not completed or failed
+			return ctrl.Result{}, err
+		}
+		// patch to mark it as failed succeeded, do not requeue
+		return ctrl.Result{}, nil
 	default:
 		b.logger.WithFields(logrus.Fields{
 			"backup": kubeutil.NamespaceAndName(original),
@@ -246,7 +260,6 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		request.Status.Phase = velerov1api.BackupPhaseInProgress
 		request.Status.StartTimestamp = &metav1.Time{Time: b.clock.Now()}
 	}
-
 	// update status
 	if err := kubeutil.PatchResource(original, request.Backup, b.kbClient); err != nil {
 		return ctrl.Result{}, errors.Wrapf(err, "error updating Backup status to %s", request.Status.Phase)
@@ -307,7 +320,10 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	log.Info("Updating backup's final status")
 	if err := kubeutil.PatchResource(original, request.Backup, b.kbClient); err != nil {
 		log.WithError(err).Error("error updating backup's final status")
+		// return the error so the status can be re-processed; it's currently still not completed or failed
+		return ctrl.Result{}, err
 	}
+
 	return ctrl.Result{}, nil
 }
 
--- a/pkg/controller/restore_controller.go
+++ b/pkg/controller/restore_controller.go
@@ -168,8 +168,8 @@ func (r *restoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// the controller.
 	log := r.logger.WithField("Restore", req.NamespacedName.String())
 
-	restore := &api.Restore{}
-	err := r.kbClient.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, restore)
+	original := &api.Restore{}
+	err := r.kbClient.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, original)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			log.Debugf("restore[%s] not found", req.Name)
@@ -181,15 +181,15 @@ func (r *restoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	// deal with finalizer
-	if !restore.DeletionTimestamp.IsZero() {
+	if !original.DeletionTimestamp.IsZero() {
 		// check the finalizer and run clean-up
-		if controllerutil.ContainsFinalizer(restore, ExternalResourcesFinalizer) {
-			if err := r.deleteExternalResources(restore); err != nil {
+		if controllerutil.ContainsFinalizer(original, ExternalResourcesFinalizer) {
+			if err := r.deleteExternalResources(original); err != nil {
 				log.Errorf("fail to delete external resources: %s", err.Error())
 				return ctrl.Result{}, err
 			}
 			// once finish clean-up, remove the finalizer from the restore so that the restore will be unlocked and deleted.
-			original := restore.DeepCopy()
+			restore := original.DeepCopy()
 			controllerutil.RemoveFinalizer(restore, ExternalResourcesFinalizer)
 			if err := kubeutil.PatchResource(original, restore, r.kbClient); err != nil {
 				log.Errorf("fail to remove finalizer: %s", err.Error())
@@ -203,8 +203,8 @@ func (r *restoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	// add finalizer
-	if restore.DeletionTimestamp.IsZero() && !controllerutil.ContainsFinalizer(restore, ExternalResourcesFinalizer) {
-		original := restore.DeepCopy()
+	if original.DeletionTimestamp.IsZero() && !controllerutil.ContainsFinalizer(original, ExternalResourcesFinalizer) {
+		restore := original.DeepCopy()
 		controllerutil.AddFinalizer(restore, ExternalResourcesFinalizer)
 		if err := kubeutil.PatchResource(original, restore, r.kbClient); err != nil {
 			log.Errorf("fail to add finalizer: %s", err.Error())
@@ -212,20 +212,31 @@ func (r *restoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		}
 	}
 
-	switch restore.Status.Phase {
+	switch original.Status.Phase {
 	case "", api.RestorePhaseNew:
 		// only process new restores
+	case api.RestorePhaseInProgress:
+		// if restore is in progress, we should not process it again
+		// we want to mark it as failed to avoid it being stuck in progress
+		// if so, mark it as failed, last loop did not successfully complete the restore
+		log.Debug("Restore has in progress status from prior reconcile, marking it as failed")
+		failedCopy := original.DeepCopy()
+		failedCopy.Status.Phase = api.RestorePhaseFailed
+		failedCopy.Status.FailureReason = "Restore from previous reconcile still in progress"
+		if err := kubeutil.PatchResource(original, failedCopy, r.kbClient); err != nil {
+			// return the error so the status can be re-processed; it's currently still not completed or failed
+			return ctrl.Result{}, err
+		}
+		// patch to mark it as failed succeeded, do not requeue
+		return ctrl.Result{}, nil
 	default:
 		r.logger.WithFields(logrus.Fields{
-			"restore": kubeutil.NamespaceAndName(restore),
-			"phase":   restore.Status.Phase,
+			"restore": kubeutil.NamespaceAndName(original),
+			"phase":   original.Status.Phase,
 		}).Debug("Restore is not handled")
 		return ctrl.Result{}, nil
 	}
-
-	// store a copy of the original restore for creating patch
-	original := restore.DeepCopy()
-
+	restore := original.DeepCopy()
 	// Validate the restore and fetch the backup
 	info, resourceModifiers := r.validateAndComplete(restore)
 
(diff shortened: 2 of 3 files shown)


pkg/controller/backup_controller.go
```diff
@@ -230,6 +234,16 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	switch original.Status.Phase {
 	case "", velerov1api.BackupPhaseNew:
 		// only process new backups
+	case velerov1api.BackupPhaseInProgress:
+		// if backup is in progress, we should not process it again
+		// we want to mark it as failed to avoid it being stuck in progress
+		// if this backupReconciler uuid is the same as the one in the backup, then we are the one who is processing it
+		// if so, mark it as failed, last loop did not successfully complete the backup
+		if original.Annotations[velerov1api.BackupControllerUID] == string(b.uuid) {
```
kaovilai, line 242: UUID here is to address a comment above `switch original.Status.Phase {`
quoted below. We can discuss if it is necessary, or if we need a real leaderelection in this thread.
```
	// Double-check we have the correct phase. In the unlikely event that multiple controller
	// instances are running, it's possible for controller A to succeed in changing the phase to
	// InProgress, while controller B's attempt to patch the phase fails. When controller B
	// reprocesses the same backup, it will either show up as New (informer hasn't seen the update
	// yet) or as InProgress. In the former case, the patch attempt will fail again, until the
	// informer sees the update. In the latter case, after the informer has seen the update to
	// InProgress, we still need this check so we can return nil to indicate we've finished processing
	// this key (even though it was a no-op).
```

pkg/controller/backup_controller.go
```diff
@@ -230,6 +234,22 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			log.Debug("Backup was processed by this backupReconciler and still has in progress status upon reconcile, marking it as failed")
+		}
+		failedCopy := original.DeepCopy()
```
sseago, line 245: Shouldn't this be inside the block above -- only set failed if the if test is true? Currently, only the debug statement is conditionalized.

pkg/controller/backup_controller.go
```diff
@@ -230,6 +234,22 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		failedCopy.Status.Phase = velerov1api.BackupPhaseFailed
+		failedCopy.Status.FailureReason = "Backup from previous reconcile still in progress"
+		if err := kubeutil.PatchResource(original, failedCopy, b.kbClient); err != nil {
+			return ctrl.Result{RequeueAfter: 5 * time.Second}, errors.Wrapf(err, "error updating Backup status to %s", failedCopy.Status.Phase)
+		}
+		// patch to mark it as failed succeeded, do not requeue
+		return ctrl.Result{}, nil
```
sseago, line 252: Also, should there be an else clause where we *do* requeue when the reconciler UID doesn't match?

kaovilai: ++

kaovilai: When reconciler UID doesn't match, the backup may belong to a reconciler that is still processing the backup.

kaovilai: So we don't requeue if UID doesn't match.. unless a way to check if another reconciler UID exists.

sseago: Hmm. But if we don't requeue, then doesn't that mean the backup is still stuck. It's InProgress, and nothing else will modify it.

sseago: But yes -- there are 2 situations possible here when UID doesn't match:
1) status update actually failed earlier, nothing is processing it.
2) backup still running in other reconciler, got queued because of an update to the backup.

For 1), if we don't requeue, we're stuck forever, if we do requeue, we eventually hit the right reconciler.
For 2), if we don't requeue, backup completes normally, but if we do, then we'll keep having to requeue on no match, on and on (waste of resources, but otherwise no harm done).

kaovilai: We're solving for case where velero pod has not restarted. we only want the one thread who was processing it that failed patching to come back and retry patching to fail later.

sseago: It still seems that if we don't requeue here, then that one thread will never retry. We already requeued, a different thread picked it up, so we bailed, but if we don't requeue now, then I suspect that the backup will still be stuck InProgress forever, just as without this patch.  The more threads we have, the smaller the likelihood that the first requeue will come to the one we want, so if we don't requeue here, especially if there are more than 2 or 3 threads, not very many of these failed-status-update backups will actually go into the failed state.

pkg/controller/backup_controller.go
kaovilai: From my team scrum this is not necessary, removing.

pkg/controller/backup_controller.go
```diff
@@ -230,6 +234,24 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		// we want to mark it as failed to avoid it being stuck in progress
+		// if this backupReconciler uuid is the same as the one in the backup, then we are the one who is processing it
+		// if so, mark it as failed, last loop did not successfully complete the backup
+		if original.Annotations[velerov1api.BackupControllerUID] == string(b.uuid) {
+			log.Debug("Backup was processed by this backupReconciler and still has in progress status upon reconcile, marking it as failed")
+			failedCopy := original.DeepCopy()
+			failedCopy.Status.Phase = velerov1api.BackupPhaseFailed
+			failedCopy.Status.FailureReason = "Backup from previous reconcile still in progress"
+			if err := kubeutil.PatchResource(original, failedCopy, b.kbClient); err != nil {
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, errors.Wrapf(err, "error updating Backup status to %s", failedCopy.Status.Phase)
```
kaovilai, line 248: switch to default requeue backoff

kaovilai (commented):
Just need to add unit tests and will mark ready for review.

pkg/controller/restore_controller.go
```diff
@@ -278,8 +289,8 @@ func (r *restoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	if err = kubeutil.PatchResource(original, restore, r.kbClient); err != nil {
 		log.WithError(errors.WithStack(err)).Info("Error updating restore's final status")
-		// No need to re-enqueue here, because restore's already set to InProgress before.
-		// Controller only handle New restore.
+		// return the error so the status can be re-processed; it's currently still not completed or failed
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
sseago: in memory or on disk? If in-memory, we just need to be careful about thread issues (to handle future cases where we have multiple reconcilers).

Also, we have the regular controller and the finalizer controller, so two different places/patches to deal with. i.e. which CR and which reconciler/controller.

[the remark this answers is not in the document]
kaovilai: Finalizer controllers won't mark backup/restore as completed right? This memory is limited to Completed phase.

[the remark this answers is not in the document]
sseago: Actually, the finalizer controller is the only one that marks it completed. Backup/restore controller moves it from InProgress to Finalizing. Both of these transitions need requeue-to-fail-or-update though.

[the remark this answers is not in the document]
kaovilai: ack Thanks for the catch.

pkg/controller/backup_controller.go
```diff
@@ -307,7 +320,10 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	log.Info("Updating backup's final status")
 	if err := kubeutil.PatchResource(original, request.Backup, b.kbClient); err != nil {
 		log.WithError(err).Error("error updating backup's final status")
+		// return the error so the status can be re-processed; it's currently still not completed or failed
```
Lyndon-Li, line 323: I remember the CR won't be re-enqueued when returning an error from the reconciler, could you confirm this?

pkg/controller/backup_controller.go
```diff
@@ -230,6 +230,20 @@ func (b *backupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	switch original.Status.Phase {
 	case "", velerov1api.BackupPhaseNew:
 		// only process new backups
+	case velerov1api.BackupPhaseInProgress:
```
Lyndon-Li, line 233: I suggest we don't add this change, this assumes the inProgress CR won't be re-enqueued, however, this constraints may hinder the future developments:
1. If we want to add a periodical enqueue mechanism, this assumption will break
2. If we want to support some feature like Cancel, this assumption will break
3. When we support parallel backups, this assumption will break
4. This mechanism can only work with backup/restore CR in the Velero server pod, for the CRs in the node-agent, this assumption is not true as node-agent is a daemonset

Generally, this change is too delicate and unless necessary, we should avoid to add the change. And consider the problem we are trying to solve, I think the retry mechanism is enough as a short term fix; for the long term, we should develop the mechanism to work with backup window, the current change won't make the ultimate fix.

pkg/controller/backup_controller.go
kaovilai: We had confirmed that returning error would cause reconciler to requeue.

This is equivalent to returning with `requeue: true` but with more information.

See: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/reconcile#Reconciler

sseago:
```
	// If the returned error is non-nil, the Result is ignored and the request will be
	// requeued using exponential backoff. The only exception is if the error is a
	// TerminalError in which case no requeuing happens.
```

pkg/controller/backup_controller.go
sseago: The retry is a good option for temporary APIServer outages (which was what originally motivated this). However, we still have a somewhat significant bug here if the APIServer is gone for an extended period of time. We end up with backups and restores (and yes, also DU/DD CRs) in an invalid state, listed as "InProgress" but which will never be touched again as long as velero stays up. The only workaround is to force-restart the Velero pod, which can cause ongoing opertations to fail.

Lyndon-Li: Yes, I agree we still have limitation if only supporting the retry mechanism. But as listed above, the current change is not a ultimate solution/universal solution for a controllers.
And moreover, this change adding a new path in the state machine of the reconciler, which is not a good design pattern and will hinder the future development/evolvement.

Lyndon-Li: I think we don't need to go into the details for the cancel or parallel backup because we don't have it for now. But generally speaking, with this change, we must keep one more rule in our mind --- the InProgress CR cannot be reconciled again, otherwise, it will be marked as failed. This will deprive our choices when we develop new features.

kaovilai: With the latest changes in https://github.com/openshift/velero/pull/324/files I believe we have addressed ability to cancel with the use of backupTracker/restoreTracker

kaovilai:
Closing as [we prefer simpler retry approach](https://github.com/vmware-tanzu/velero/issues/7207#issuecomment-2232073194) which would not be as complicated for future reconciler work like requeue.

PR #7899, opened 2024-06-17
Don't consider unschedulable pods unrecoverable

### LINKED ISSUE #7898: Fail early DM pod checks have false positives
In Velero 1.13, we added code to fail early if the datamover pods are in an unrecoverable state: https://github.com/vmware-tanzu/velero/pull/7052
In 1.14, some additional changes were made here: https://github.com/vmware-tanzu/velero/pull/7437

In our OADP testing with Velero 1.14, we found that in some cases DDs or DUs were being canceled almost immediately. Digging into the code added in the above 2 PRs, there are a few situations where Velero considers a pod unrecoverable:
1) pod phase of Failed or Unknown
2) pod phase of Pending with an "unschedulable" condition
3) container status with ImagePullBackOff or ErrImgNeverPull

It was the second situation that was causing the failure for us. The pod will have an "unschedulable" condition until the PVC is is bound to a PV. In many cases, this all happens quickly enough that by the time Velero checks, the PVC is bound and the Pod is no longer unschedulable, but if the provisioner takes a few seconds to bind the PV and PVC together, the Pod may still have an "unschedulable" condition when the node agent first checks, which will cause the DU or DD to be immediately canceled.

We are seeing this happen frequently with Ceph volumes.

We have a couple of options here.

1) The simplest option would be to remove "unschedulable" from the list of unrecoverable conditions. As we see here, "unschedulable" does not always indicate an unrecoverable state. The only downside here is that we lose fast fail for permanently-unschedulable pods.

2) Slightly more complicated -- we could wrap the unschedulable  check in a PollUntilContextTimeout call with some hard-coded timeout (2 minutes maybe?) -- this will result in almost-fast-failing for unschedulable pods. This assumes that no provisioner should take more than 2 minutes to provision the PV. That's probably safe. 

I will be posting a draft PR for option 1) -- I'm open to moving to 2) if the consensus is that it's a better solution.

**Environment:**

- Velero version (use `velero version`): 1.14

**Vote on this issue!**

sseago:

Don't consider "unschedulable" pods unrecoverable (for fast fail purposes), since pods will be briefly unschedulable while waiting for PV provisioning.

Fixes #7898

- [x ] Updated the corresponding documentation in `site/content/docs/main`.

### DIFF
--- a/changelogs/unreleased/7899-sseago
+++ b/changelogs/unreleased/7899-sseago
@@ -1,0 +1,1 @@
+Don't consider unschedulable pods unrecoverable
--- a/pkg/exposer/csi_snapshot_test.go
+++ b/pkg/exposer/csi_snapshot_test.go
@@ -642,14 +642,7 @@ func TestPeekExpose(t *testing.T) {
 			Name:      backup.Name,
 		},
 		Status: corev1.PodStatus{
-			Phase: corev1.PodPending,
-			Conditions: []corev1.PodCondition{
-				{
-					Type:    corev1.PodScheduled,
-					Reason:  "Unschedulable",
-					Message: "unrecoverable",
-				},
-			},
+			Phase: corev1.PodFailed,
 		},
 	}
 
@@ -679,7 +672,7 @@ func TestPeekExpose(t *testing.T) {
 			kubeClientObj: []runtime.Object{
 				backupPodUrecoverable,
 			},
-			err: "Pod is unschedulable: unrecoverable",
+			err: "Pod is in abnormal state Failed",
 		},
 		{
 			name:        "succeed",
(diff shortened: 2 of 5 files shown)


shubham-pampattiwar (approved):

Lyndon-Li (approved):

Lyndon-Li:
@sseago Please help to cherry-pick this PR to 1.14. Thanks.

itayrin:
Hi, I just want to say that lately I was trying to integrate Velero 1.14.0 with our product and this specific commit made me downgrade to 13.2. This is an important fix, took me a few days to find the issue.

sseago:
@itayrin Yes, unfortunately we didn't discover this bug until 1.14.0 was out. Some storage providers provision volumes quickly enough that we don't hit it, so our pre-release testing didn't show this -- but other providers take a bit longer for volumes to come up, and for those providers this bug was hit all the time. 1.14.1 includes the fix, and we're hoping to release that around the end of the month.

PR #7924, opened 2024-06-25
In backup sync flow put snapshotHandle as source in CSI VSContent

anshulahuja98:

Fixes #(issue)
https://kubernetes.io/docs/concepts/storage/volume-snapshots/
image

Context:
We have few customers which have used same storage bucket across clusters, as a result of that their VSContents are syncing across clusters, After the sync the other cluster does not have permission on the snapshot of original cluster, it starts to rePUT the snapshot.
This leads to significant load on the CSI driver.

### DIFF
--- a/pkg/controller/backup_sync_controller.go
+++ b/pkg/controller/backup_sync_controller.go
@@ -272,6 +272,10 @@ func (b *backupSyncReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 			for _, snapCont := range snapConts {
 				// TODO: Reset ResourceVersion prior to persisting VolumeSnapshotContents
 				snapCont.ResourceVersion = ""
+				// For creating static VSContent, we should put snapshothandle in source rather than volume handle.
+				// Because if VSContent syncs to a different cluster, having volumeHandle will force rePUTs on the snapshot
+				snapCont.Spec.Source.SnapshotHandle = snapCont.Status.SnapshotHandle
+				snapCont.Spec.Source.VolumeHandle = nil
 				err := b.client.Create(ctx, snapCont, &client.CreateOptions{})
 				switch {
 				case err != nil && apierrors.IsAlreadyExists(err):


pkg/controller/backup_sync_controller.go
```diff
@@ -272,6 +272,10 @@ func (b *backupSyncReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			for _, snapCont := range snapConts {
 				// TODO: Reset ResourceVersion prior to persisting VolumeSnapshotContents
 				snapCont.ResourceVersion = ""
+				// For creating static VSContent, we should put snapshothandle in source rather than volume handle.
+				// Because if VSContent syncs to a different cluster, having volumeHandle will force rePUTs on the snapshot
+				snapCont.Spec.Source.SnapshotHandle = snapCont.Status.SnapshotHandle
+				snapCont.Spec.Source.VolumeHandle = nil
```
shubham-pampattiwar, line 278: Will the `SnapshotHandle` value be unique globally ? Or does it depend on:
- whether the storage across clusters is shared ?
- the csi driver of the storage provider ?

shubham-pampattiwar (commented):
added a few questions

pkg/controller/backup_sync_controller.go
kaovilai: snapshotHandle in my experience is the unique identifier at the cloud provider level.

As long as it's the same cloud provider, same provisioner, then same snapshothandle = same unique content globally.

kaovilai: snapshotHandle for aws provisioner would be the same value as SnapshotId of an ec2 EBS snapshot
[`"SnapshotId": "snap-066877671789bd71b"
`](https://docs.aws.amazon.com/cli/latest/reference/ec2/create-snapshot.html#:~:text=%22SnapshotId%22%3A%20%22snap%2D066877671789bd71b%22) which for the same account credentials should be unique globally regardless of cluster.

anshulahuja98: My understanding is also that same that it will mostly be unique across
Using a GUID in any naming ensures that.

In case of Azure it's a fully qualified URL with subscription, RG and snapshot name.

But again any GUID should ensure uniqueness.

anshulahuja98: @shubham-pampattiwar any other thoughts?

shubham-pampattiwar: If possible, Could you provide a link to the change/PR that is causing the rePUTs on the snapshot ?

anshulahuja98: tagged on slack

blackpiglet (approved):

shubham-pampattiwar (approved):

PR #7941, opened 2024-06-27
Use predicate in finalizer controllers to only process update events.

### LINKED ISSUE #7868: Finalizer controllers should use predicate during SetupWithManager() to filter instead of switch/case inside Reconcile()

Not urgent, just to clean up and follow more k8s convention.

**Environment:**

**Vote on this issue!**

kaovilai:

Fixes #7868

/kind changelog-not-required


pkg/controller/backup_finalizer_controller.go
```diff
@@ -228,6 +222,19 @@ func (r *backupFinalizerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&velerov1api.Backup{}).
+		WithEventFilter(predicate.Funcs{
+			// Only process backups that are updated to finalizing phase after plugin operations are complete
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				backup, ok := e.ObjectNew.(*velerov1api.Backup)
+				if !ok {
+					return false
+				}
+				return backup.Status.Phase == velerov1api.BackupPhaseFinalizing || backup.Status.Phase == velerov1api.BackupPhaseFinalizingPartiallyFailed
```
blackpiglet, line 232: We should return the opposite result here.
Restore reconciler should also be changed.

kaovilai:
```go
	// Update returns true if the Update event should be processed
	UpdateFunc func(event.UpdateEvent) bool
```

Don't we want events to be processed if CR was updated to finalizing?

sseago: @kaovilai so there are 2 cases where we want to process backups in this controller:
1) backups updated to Finalizing or FinalizingPartiallyFailed phase
2) backups already in finalizing phase that returned err from prior reconcile (i.e. requeueing)

Does this predicate change affect requeue behavior?

pkg/controller/backup_finalizer_controller.go
```diff
@@ -100,14 +102,6 @@ func (r *backupFinalizerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 	}
 
-	switch backup.Status.Phase {
-	case velerov1api.BackupPhaseFinalizing, velerov1api.BackupPhaseFinalizingPartiallyFailed:
-		// only process backups finalizing after  plugin operations are complete
-	default:
-		log.Debug("Backup is not awaiting finalizing, skipping")
-		return ctrl.Result{}, nil
-	}
-
```
sseago, line 110: Are we sure we want to remove this? Keeping it might be a good guard against a backup somehow getting into the queue or a backup that got updated out of this phase some time between being queued and processed. I think it won't hurt to keep this check here to make sure we never process a backup here that's not in a Finalizing phase.

pkg/controller/restore_finalizer_controller.go
```diff
@@ -111,13 +126,6 @@ func (r *restoreFinalizerReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		},
 	)
 
-	switch restore.Status.Phase {
-	case velerov1api.RestorePhaseFinalizing, velerov1api.RestorePhaseFinalizingPartiallyFailed:
-	default:
-		log.Debug("Restore is not awaiting finalization, skipping")
-		return ctrl.Result{}, nil
-	}
-
```
sseago, line 120: Same question about whether to keep this here as on backup.

pkg/controller/backup_finalizer_controller.go
kaovilai: Ok will revert this.

sseago (dismissed):

blackpiglet (dismissed):

sseago (approved):

pkg/controller/restore_finalizer_controller.go
blackpiglet, line 137: Should we delete the existing code to filter restore in the reconciler?

kaovilai: Keeping this — predicate only filters watch-triggered events (Create/Update/Delete/Generic). Requeues from `Reconcile` (error retry or explicit `Requeue: true`) skip `WithEventFilter` and call `Reconcile` directly, so without this in-function phase check a requeued restore that transitioned out of Finalizing between enqueue and processing would run finalization logic incorrectly. Same reasoning sseago gave for keeping the equivalent check in `backup_finalizer_controller.go` (resolved thread above) — kept symmetric here.

Issue #7866, opened 2024-06-07
Restoring a backup with no pods is PartiallyFailed with error: `fail to patch dynamic PV, err: context deadline exceeded`


shubham-pampattiwar:
**What steps did you take and what happened:**

- Install Velero 1.14 (rc in this case)
- Create a backup with no pods(scaled down the replicas to 0 for deployments involved), but include the volumes (via csi snapshot method, non-datamover)
- Delete application ns
- Restore from the backup
- Restore is partially failed with `fail to patch dynamic PV, err: context deadline exceeded`
- Note that the StorageClass used in this scenario has `VolumeBindingMode` as `WaitForFirstConsumer`

**What did you expect to happen:**
- In earlier versions of velero, ie. before velero 1.14, the restore described above was not  marked `PartiallyFailed` because there was no [`Finalizing` Phase for Restore Operation](https://github.com/vmware-tanzu/velero/issues/7183) that tried to[ patch the PV](https://github.com/vmware-tanzu/velero/pull/7504)
- In our scenario as the StorageClass  has `VolumeBindingMode` as `WaitForFirstConsumer`, the PV will be in pending state as there are no pods in the backup and this is turn will fail the PV patch part in restore operation after the timeout, thus marking the Restore partially failed.

**The following information will help us better understand what's going on**:

_If you are using velero v1.7.0+:_  
Please use `velero debug  --backup <backupname> --restore <restorename>` to generate the support bundle, and attach to this issue, more options please refer to `velero debug --help` 

_If you are using earlier versions:_  
Please provide the output of the following commands (Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)
- `kubectl logs deployment/velero -n velero`
- `velero backup describe <backupname>` or `kubectl get backup/<backupname> -n velero -o yaml`
- `velero backup logs <backupname>`
- `velero restore describe <restorename>` or `kubectl get restore/<restorename> -n velero -o yaml`
- `velero restore logs <restorename>`

**Anything else you would like to add:**

**Environment:**

- Velero version (use `velero version`): 
- Velero features (use `velero client config get features`): 
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

[...]

PR #7953, opened 2024-06-28
Skip PV patch step in Restore workflow for WaitForFirstConsumer VolumeBindingMode Pending state PVCs

shubham-pampattiwar:

Fixes https://github.com/vmware-tanzu/velero/issues/7866

### DIFF
--- a/pkg/controller/restore_finalizer_controller.go
+++ b/pkg/controller/restore_finalizer_controller.go
@@ -22,6 +22,8 @@ import (
 	"sync"
 	"time"
 
+	storagev1api "k8s.io/api/storage/v1"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/controller/restore_finalizer_controller.go
```diff
@@ -304,6 +305,24 @@ func (ctx *finalizerContext) patchDynamicPVWithVolumeInfo() (errs results.Result
...
+						// check if storage class used has VolumeBindingMode as WaitForFirstConsumer
+						scName := *pvc.Spec.StorageClassName
+						sc := &storagev1api.StorageClass{}
+						err = ctx.crClient.Get(context.Background(), client.ObjectKey{Name: scName}, sc)
+
+						if err != nil {
+							errs.Add(restoredNamespace, err)
+						}
+						// skip PV patch step for this scenario
+						// because pvc would not be bound and the PV patch step would fail due to timeout thus failing the backup
```
kaovilai, line 319:
```suggestion
						// because pvc would not be bound and the PV patch step would fail due to timeout thus failing the restore
```

kaovilai (changes requested):

pkg/controller/restore_finalizer_controller.go
```diff
@@ -304,6 +305,24 @@ func (ctx *finalizerContext) patchDynamicPVWithVolumeInfo() (errs results.Result
 						return false, err
 					}
 
+					// check if pvc is not nil and has a pending status
```
blackpiglet, line 308: Could you add more information about this scenario in the comment?
This fix is for a corner case, and it's better to give more information to understand what this does.

sseago: @blackpiglet I agree that the comment should be clearer on when this happens, but it's not a corner case. WaitForFirstConsumer is a pretty common configuration -- this happens any time you have a WaitForFirstConsumer PVC that's not mounted by a pod. Lots of people are going to hit this.

shubham-pampattiwar: Done, updated the PR.

pkg/controller/restore_finalizer_controller.go
```diff
@@ -304,6 +306,26 @@ func (ctx *finalizerContext) patchDynamicPVWithVolumeInfo() (errs results.Result
...
+					// failures due to the PVC not being bound, which could cause a timeout and result in a failed restore.
+					if pvc != nil && pvc.Status.Phase == v1.ClaimPending {
+						// check if storage class used has VolumeBindingMode as WaitForFirstConsumer
+						scName := *pvc.Spec.StorageClassName
+						sc := &storagev1api.StorageClass{}
+						err = ctx.crClient.Get(context.Background(), client.ObjectKey{Name: scName}, sc)
+
+						if err != nil {
+							errs.Add(restoredNamespace, err)
+						}
```
kaovilai, line 320: Do you want to return after this error occurred?
Without return, it would still try to patch since sc was not retrievable and `*sc.VolumeBindingMode` would be undefined.

shubham-pampattiwar: Updated PR, thanks @kaovilai

sseago (approved):

Lyndon-Li (approved):

Issue #7620, opened 2024-04-03
Make kopia repo cache policy configurable


Lyndon-Li:
Related to issue #7499. 
The cache policy decides the root file system disk usage in the pod where data movement is running, and on the other hand, it also impact the restore performance significantly.
Therefore, besides the hard-code policy, we should expose some of them to users so that users could decide the cache disk usage.

Lyndon-Li:
@u3813 

> But in my case, it seems that too much data is being restored within these 10 minutes

As the current cache management policy, we only control the living time of the cache data, we don't check the size of the cache in a period of time. Then how much data is cached relies on the total size to be restored, the parallelism of the restore. The larger of the restore size, the more parallelism (by default decided by the CPU cores in the node), the more data will be cached.
Therefore, the final solution is to control the cache by checking the total size, if it exceeds a defined size, delete the oldest data.

> I would rather have more options when it comes to configuring the cache. It should be possible to either set a lower value than 10 minutes or to set a hard GB limit after which the cache is cleared

This will be done in the fix of this issue.
And another issue #7725 gives another option for users to configure a dedicate cache volume then the restore performance will not be compromised.

janpipan:
We also had a problem with disk pressure, which we resolved by adding additional disk capacity, but I think this is not something desirable as the ephemeral storage consumption is basically the same size as the PV being restored. 

> @lyndon-li

> As the current cache management policy, we only control the living time of the cache data, we don't check the size of the cache in a period of time. 

Was this sufficiently tested, as I believe there is also a bug with a cache cleanup, because in our case it didn't execute as expected.

When we restored the PV with around 50 GB for the first time the cache got cleaned up after around 10 mins, but after the second restore the cleanup did not happen automatically. I left the node-agent Pod running for over a day, but the storage consumption on the worker node did not drop as expected, which is why I deleted the Pod manually today which resolved the issue.

reasonerjt:
Combining #7301 we should probably design a consistent way to let user configure the advanced settings for kopia.

But before that, we should consider adjust the default policy to mitigate the risk for restore failure.

Lyndon-Li:
@mpryc Could you comment on #7301 for why compression is important to your case?

mpryc:
@Lyndon-Li For me compression is not important, but the design should cover generic advanced settings for kopia as pointed in the https://github.com/vmware-tanzu/velero/issues/7620#issuecomment-2167502875 comment. That is why I've mentioned it.

[...]

Issue #7301, opened 2024-01-10
Kopia repo - enable compression


Lyndon-Li:
Compression is not enabled for Kopia repo, after several releases with Kopia, we can try to provide an optional configuration to enable it.

kaovilai:
This would add cpu requirements I assume

Lyndon-Li:
> This would add cpu requirements I assume

Yes, so it is configurable

xingxiuyi:
We need compression, the current backup in our cluster is too large. 

Some files, such as text and database file in pods, would be better compressed.

reasonerjt:
We'll need a design to determine how to expose these settings.

mpryc:
It would make more sense to move this feature to the Backup CRD rather than Repository: `Kopia repo - enable compression`.

In Kopia, compression is a client-side, per-snapshot setting, not a repository-wide configuration. This is by design and differs from repo-level settings like encryption, hashing, or chunk splitting, which are fixed at repository creation.

The user should set the desired compression algorithm as part of the Velero Backup, allowing it to be applied during snapshot creation and enabling more flexible, per-backup control.

shubham-pampattiwar:
+1 I agree with @mpryc's suggestion here, This offers more flexibility for users to define compression on a per-backup basis, which can be particularly useful in environments with varying data types or performance requirements.

mpryc:
Thank you. I will come up with some design and propose shortly. At first look it seems to introduce new config alongisde:
 https://github.com/vmware-tanzu/velero/blob/main/pkg/uploader/util/uploader_config.go#L28-L30

The idea would be to have a new `KopiaConfig` structure within the existing `UploaderConfig` that will be used to set the policy.

User would specify the compression as part of `Backup` CRD:

```yaml
apiVersion: velero.io/v1
kind: Backup
metadata:
  name: backup-fast-compression
spec:
  [...]
  uploaderConfig:
    kopiaConfig:
      compressionConfig:
        algorithm: "s2-default"
```

This would allow to use other kopia related configs in the future that are on the client side not only compression.

I will describe this in detail in the design doc, unless someone knows a better place for this and other kopia config related setting.

Lyndon-Li:
> It would make more sense to move this feature to the Backup CRD rather than Repository: `Kopia repo - enable compression`.
> 
[...]

PR #7963, opened 2024-07-02
Add design for backup repository configurations

Lyndon-Li:
Add design for backup repository configurations for issue #7620, #7301

### DIFF
--- a/design/backup-repo-config.md
+++ b/design/backup-repo-config.md
@@ -7,14 +7,14 @@
 
 ## Background
 
-According to the [Unified Repository design][1] Velero uses selectable backup respositories for various backup/restore methods, i.e., fs-backup, volume snapshot data movement, etc. To achieve the best performance, backup respositories may need to be configured according to the running environments.  
+According to the [Unified Repository design][1] Velero uses selectable backup repositories for various backup/restore methods, i.e., fs-backup, volume snapshot data movement, etc. To achieve the best performance, backup repositories may need to be configured according to the running environments.  
 For example, if there are sufficient CPU and memory resources in the environment, users may enable compression feature provided by the backup repository, so as to achieve the best backup throughput.  
-As another example, if the local disk space is not sufficent, users may want to constraint the backup repository's cache size, so as to prevent from running out of the disk space.  
-Therefore, it is worthy to allow users to configure some essential prameters of the backup repsoitories, and the configuration may vary from backup respositories.  
+As another example, if the local disk space is not sufficient, users may want to constraint the backup repository's cache size, so as to prevent the repository from running out of the disk space.  
+Therefore, it is worthy to allow users to configure some essential parameters of the backup repsoitories, and the configuration may vary from backup repositories.  
 
 ## Goals
 
-- Create a mechnism for users to specify configurations for backup repositories
+- Create a mechanism for users to specify configurations for backup repositories  
 
 ## Non-Goals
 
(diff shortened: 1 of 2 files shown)


design/backup-repo-config.md
```diff
@@ -0,0 +1,105 @@
...
+                type: string
+            required:
+            - backupStorageLocation
+            - maintenanceFrequency
+            - resticIdentifier
+            - volumeNamespace
+            type: object
+```            
+
+### BackupRepository configMap
```
mpryc, line 74: While `ConfigMap` (capital C), is great option here, I do wonder if some configuration may require `Secret` on top of what `ConfigMap` provides ? It would be similar mechanism, just `Secret` with label.

design/backup-repo-config.md
```diff
@@ -0,0 +1,105 @@
...
+        "cache-limit-mb": 2048,
+        "enable-compression": true
+    }
+}
+```
+
+To create the configMap, users need to save something like the above sample to a json file and then run below commands:  
+```
+kubectl create cm <config-name> -n velero --from-file=<json file name>
+kubectl label cm <config-name> -n velero backup-repository-config=<repository type>
```
mpryc, line 100: I think it's also worth mentioning different way of creating ConfigMap:

config-map.yaml
```yaml

apiVersion: v1
kind: ConfigMap
metadata:
  name: <config-name>
  namespace: velero
  labels:
    backup-repository-config: <repository type>
data:
  repoConfig: |
    {
      "cache-limit-mb": 2048,
      "enable-compression": true
    }
```

```shell
$ kubectl apply -f config-map.yaml

**Review by mpryc — COMMENTED** [review]
In general I do like the design.

File: design/backup-repo-config.md (hunk 1 above)
Comment (Lyndon-Li) [reply design/backup-repo-config.md:74]: The credentials for the repository to access the backup storage have already been included in the BSL, so we don't need a secret in the backup repository configuration. The backup repository configuration only contains selective options to configure the repo at the run time.

File: design/backup-repo-config.md [hunk 3]
```diff
@@ -0,0 +1,126 @@
...
+
+- Create a mechanism for users to specify configurations for backup repositories  
+
+## Non-Goals
+
+## Solution
+
+### BackupRepository CRD
+
+After a backup repository is initialized, a BackupRepository CR is created to represent the instance of the backup repository. The BackupRepository's spec is a core parameter used by Uinified Repo modules when interactive with the backup repsoitory. Therefore, we can add the configurations into the BackupRepository CR called ```repositoryConfigs```.
```
shubham-pampattiwar, line 25: s/Uinified/Unified

design/backup-repo-config.md
```diff
@@ -0,0 +1,126 @@
...
+Therefore, we create a BackupRepository configMap as a template of the configurations to be applied to the backup repository CR when it is created by the BackupRepository controller. For an existing BackupRepository CR, the configMap is never visited, if users want to modify the configuration value, they should directly edit the BackupRepository CR.   
+The BackupRepository configMap is created by users. If the configMap is not there, nothing is specified to the backup repository CR, so the Unified Repo modules use the hard-coded values to configure the backup repository.  
+The BackupRepository configMap is backup repository type specific, that is, for each type of backup repository, there is one configMap. Therefore, a ```backup-repository-config``` label should be applied to the configMap with the value of the repository's type. During the backup repository creation, the configMap is searched by the label against the repository type, if no configMap is found, nothing is set to the backup repository CR.
```
shubham-pampattiwar, line 79: Also, the configmap should exist in velero install ns ?

design/backup-repo-config.md
```diff
@@ -0,0 +1,126 @@
...
+Therefore, we create a BackupRepository configMap as a template of the configurations to be applied to the backup repository CR when it is created by the BackupRepository controller. For an existing BackupRepository CR, the configMap is never visited, if users want to modify the configuration value, they should directly edit the BackupRepository CR.   
+The BackupRepository configMap is created by users. If the configMap is not there, nothing is specified to the backup repository CR, so the Unified Repo modules use the hard-coded values to configure the backup repository.  
+The BackupRepository configMap is backup repository type specific, that is, for each type of backup repository, there is one configMap. Therefore, a ```backup-repository-config``` label should be applied to the configMap with the value of the repository's type. During the backup repository creation, the configMap is searched by the label against the repository type, if no configMap is found, nothing is set to the backup repository CR.  
+
+### Configurations
```
shubham-pampattiwar, line 81: Should we also document the list of parameters that are configurable and the ones that are hard coded for kopia repo ?

design/backup-repo-config.md
```diff
@@ -0,0 +1,126 @@
...
+
+To create the configMap, users need to save something like the above sample to a file and then run below commands:  
+```
+kubectl create cm <config-name> -n velero --from-file=<json file name>
+kubectl label cm <config-name> -n velero backup-repository-config=<repository type>
+```  
+Or
+```
+kubectl apply -f <yaml file name>
+kubectl label cm <config-name> -n velero backup-repository-config=<repository type>
```
mpryc, line 122: There is already label in the yaml file that describes ConfigMap, so it's not required.

design/backup-repo-config.md
Lyndon-Li: `cache-limit-mb` and `enable-compression` are the supported parameters.
There are no other parameters that should be exposed to users, so they are not relevant to this design. But we can find all the parameters for kopia from the code at https://github.com/vmware-tanzu/velero/blob/main/pkg/repository/udmrepo/kopialib/backend/common.go

design/backup-repo-config.md
Lyndon-Li: Done

design/backup-repo-config.md
Lyndon-Li: Done

design/backup-repo-config.md
Lyndon-Li: Done

shubham-pampattiwar (approved):

Lyndon-Li:
@shubham-pampattiwar Made some slight changes, please review again. Thanks.

mpryc (approved):

shubham-pampattiwar (approved):
/lgtm

design/backup-repo-config.md
```diff
@@ -0,0 +1,126 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: <config-name>
+  namespace: velero
+data:
+  repoConfig: |
+    {
+      "cacheLimitMB": 2048,
+      "enableCompression": true
```
mpryc, line 111: I also think it's worth considering allowing other repo configuration options to be configurable via this design. Namely thinking of different kopia repo options e.g.:
https://github.com/wawa0210/velero/blob/89c10ddcc04e4dc7354f284434c2ddb8c121780c/pkg/repository/udmrepo/repo_options.go#L63-L65

mpryc: @weshayutin ^^

Lyndon-Li: Let's do this incrementally in future based on the requirements.
For now, the 3 options you mentioned are too advanced(especially for the hash algorithm and chunking algorithm, without a good knowledge of the deduplication and the status of their own data, users cannot decide a better value than the default ones) and not general enough in terms of unified repository concepts(many backup repositories/storages are using private chunking/hashing algorithms and don't allow users to change, because those have heavy impacts on the QOS of the repo).

blackpiglet (dismissed):

Lyndon-Li:
@mpryc @shubham-pampattiwar @blackpiglet 
After some discussions, we made below changes to this design:
1. Move configurations for all repo types into one configMap. Previously one configMap for one repo type
2. Use configMap name to find the configMap and require the name to be explicitly set to velero server parameters
3. Add a installation flag to set the velero server parameter

Please have further review. Thanks.

blackpiglet (dismissed):

Lyndon-Li:
> @Lyndon-Li I am fine with either of the approaches. Would like to why the change in approach though ?

After some discussion, we would like to unify the way to set configurations through configMap (including node-agent configs, backup repository configs and maintenance job configs) to velero:

- The configMap should be always searched by name
- The configMap name should be set to velero server/node-agent at start time, though users may be allowed to change the content of the configMap later. Otherwise, the configMap will be ignored

The would help the downstream users (which integrate velero) to better control velero's behavior in a consistent way. cc @reasonerjt

design/backup-repo-config.md
```diff
@@ -0,0 +1,163 @@
...
+```  
+Or
+```
+kubectl apply -f <yaml file name>
+```  
+
+## Installation
+In order to set the ```--backup-repository-config``` parameter to the velero server, a new installation parameter ```--backup-repository-config``` will be added. An example of the installation as below:
+```
+velero install --backup-repository-config <configmap-name>
```
reasonerjt, line 158: This could be optional b/c we can hard code a configmap name and pass it to the velero server, and the user will only need to create the configmap with the hard-coded name.

If we took this approach ^ we may not need to update the installation flow, bearing in mind that when velero is installed it's very possible the namespace of velero is not created.

design/backup-repo-config.md
```diff
@@ -0,0 +1,163 @@
...
+The BackupRepository configMap is created by users in velero installation namespace. The configMap name must be specified in the velero server parameter ```--backup-repository-config```, otherwise, it won't effect.  
+If the configMap name is specified but the configMap doesn't exist by the time of a backup repository is created, the configMap name is ignored.  
+For any reason, if the configMap doesn't effect, nothing is specified to the backup repository CR, so the Unified Repo modules use the hard-coded values to configure the backup repository.  
+
+The BackupRepository configMap supports backup repository type specific configurations, even though users can only specify one configMap.  
+So in the configMap struct, multiple entries are supported, indexed by the backup repository type. During the backup repository creation, the configMap is searched by the repository type.  
+
+Below are the struct for the configMap:  
+``` golang
+type RepoConfig struct {
```
reasonerjt, line 91: I wanna clarify, it's the `RepoConfig` will be populated to the CR right?  If that's the case, should the field in the backup repo CR be singular (`repositoryConfig`) or plural (`repositoryConfigs`)?

design/backup-repo-config.md
```diff
@@ -0,0 +1,163 @@
...
+                description: |-
+                  BackupStorageLocation is the name of the BackupStorageLocation
+                  that should contain this repository.
+                type: string
+              maintenanceFrequency:
+                description: MaintenanceFrequency is how often maintenance should
+                  be run.
+                type: string
+              repositoryConfigs:
+                additionalProperties:
```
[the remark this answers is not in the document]
reasonerjt: OK, turns out it's regular convention for CRD if we wanna use free-form k/v pairs:
https://stackoverflow.com/questions/64828910/dynamic-key-value-input-properties-to-crd-in-kubernetes

design/backup-repo-config.md
```diff
@@ -0,0 +1,163 @@
...
+        "cacheLimitMB": 1024,
+        "enableCompression": false    
+      }
+    }    
+```
+
+To create the configMap, users need to save something like the above sample to a file and then run below commands:  
+```
+kubectl create cm <config-name> -n velero --from-file=<json file name>
+kubectl label cm <config-name> -n velero backup-repository-config=<repository type>
```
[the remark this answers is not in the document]
Lyndon-Li: No, it is not required, I have removed it.

design/backup-repo-config.md
Lyndon-Li: We will remove this installation flag and keep the velero server flag.
By default installation using velero CLI, the server flag is not set; users need to modify the deployment to set the flag and wait velero server to restart.

design/backup-repo-config.md
Lyndon-Li: Changed the field in the CRD to `repositoryConfig`

design/backup-repo-config.md
```diff
@@ -0,0 +1,156 @@
...
+Therefore, a BackupRepository configMap is introduced as a template of the configurations to be applied to the backup repository CR.  
+When the backup repository CR is created by the BackupRepository controller, the configurations in the configMap are copied to the ```repositoryConfig``` field.   
+For an existing BackupRepository CR, the configMap is never visited, if users want to modify the configuration value, they should directly edit the BackupRepository CR.  
+
+The BackupRepository configMap is created by users in velero installation namespace. The configMap name must be specified in the velero server parameter ```--backup-repository-config```, otherwise, it won't effect.  
+If the configMap name is specified but the configMap doesn't exist by the time of a backup repository is created, the configMap name is ignored.  
+For any reason, if the configMap doesn't effect, nothing is specified to the backup repository CR, so the Unified Repo modules use the hard-coded values to configure the backup repository.
```
kaovilai, line 84: Will velero be watching for changes to CM? or will it only parse the value at startup? Or per backup?

shubham-pampattiwar: I think its at startup and per backup, @Lyndon-Li lets clarify that

Lyndon-Li: When a backupRepository CR is created, this CM is visited. So Velero doesn't watch the changes to the CM, but users are allowed to change the content during runtime, the changes will effect for the next created backupRepository. See below statement in the design:
`When the backup repository CR is created by the BackupRepository controller, the configurations in the configMap are copied to the ```repositoryConfig``` field.   `

shubham-pampattiwar (approved):

blackpiglet (approved):

Issue #7972, opened 2024-07-03
Data mover --- backupVS/backupVSC may be left on dataUpload cancel


Lyndon-Li:
When a dataUpload is cancelled at the Prepare phase, the exposer tries to delete the backupPVC and backupVS.
As a matter of fact, the backupVS won't be deleted until the backupPVC disappears from the cluster.

As a result, the backupVS and the snapshot may not be deleted timely but wait for the backupVS to be reconciled one or more times by the CSI driver.

PR #7973, opened 2024-07-03
Issue 7972: sync the backupPVC deletion in expose clean up

Lyndon-Li:
Fix issue #7972, sync the backupPVC deletion in expose clean up

### DIFF
--- a/pkg/util/kube/pvc_pv.go
+++ b/pkg/util/kube/pvc_pv.go
@@ -41,9 +41,10 @@ const (
 	waitInternal = 2 * time.Second
 )
 
-// DeletePVAndPVCIfAny deletes PVC and delete the bound PV if it exists and log an error when the deletion fails
-// We first set the reclaim policy of the PV to Delete, then PV will be deleted automatically when PVC is deleted.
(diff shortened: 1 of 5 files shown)


pkg/util/kube/pvc_pv.go
```diff
@@ -69,7 +70,7 @@ func DeletePVAndPVCIfAny(ctx context.Context, client corev1client.CoreV1Interfac
 		}
 	}
 
-	if err := client.PersistentVolumeClaims(pvcNamespace).Delete(ctx, pvcName, metav1.DeleteOptions{}); err != nil {
+	if err := EnsureDeletePVC(ctx, client, pvcName, pvcNamespace, ensureTimeout); err != nil {
```
blackpiglet, line 73: If the ensureTimeout is set to 0, what is the behavior of `wait.PollUntilContextTimeout`?
Does the function fail immediately, or return no matter what happens?

Lyndon-Li: It could abort the wait immediately, but it will return a timeout error which is not expected.
I have modified to explicitly check the ensureTimeout==0 case.

blackpiglet (approved):

ywk253100 (approved):

PR #7974, opened 2024-07-03
Only get VolumeSnapshotClass when DataUpload exists.

### LINKED ISSUE #7823: Velero backup gets partially failed when CSI is enabled but VolumeSnapshotClass CRD doesn't exist
For 1.13+
Install Velero with CSI enabled
In the cluster, there is no VolumeSnapshotClass CRD
Run a fs-backup
Velero gets partially failed with error like `cannot list VolumeSnapshotClass no matches for kind \"VolumeSnapshotClass\" in version \"snapshot.storage.k8s.io/v1\"`

blackpiglet:

Fixes #7823 

### DIFF
--- a/internal/volume/volumes_information.go
+++ b/internal/volume/volumes_information.go
@@ -36,7 +36,6 @@ import (
 	"github.com/vmware-tanzu/velero/pkg/features"
 	"github.com/vmware-tanzu/velero/pkg/itemoperation"
 	"github.com/vmware-tanzu/velero/pkg/kuberesource"
-	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
 )
 
 type Method string
@@ -551,6 +550,19 @@ func (v *BackupVolumesInformation) generateVolumeInfoFromPVB() {
 	v.volumeInfos = append(v.volumeInfos, tmpVolumeInfos...)
 }
 
+func (v *BackupVolumesInformation) getVolumeSnapshotClasses() (
+	[]snapshotv1api.VolumeSnapshotClass,
+	error,
+) {
+	vsClassList := new(snapshotv1api.VolumeSnapshotClassList)
+	if err := v.crClient.List(context.TODO(), vsClassList); err != nil {
+		v.logger.Warnf("Cannot list VolumeSnapshotClass with error %s.", err.Error())
+		return nil, err
+	}
+
+	return vsClassList.Items, nil
+}
+
 // generateVolumeInfoFromDataUpload generate BackupVolumeInfo for DataUpload.
 func (v *BackupVolumesInformation) generateVolumeInfoFromDataUpload() {
 	if !features.IsEnabled(velerov1api.CSIFeatureFlag) {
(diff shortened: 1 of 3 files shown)


blackpiglet:
> Don't we also need VSClass for CSI snapshot backups without datamover?

Yes, but for the CSI plugin scenario, because the VolumeSnapshot-related CRs are persisted, so VolumeSnapshotClasses are already passed in by parameters.

https://github.com/vmware-tanzu/velero/blob/1ec52beca80975f74f9ed28d6f9c5f7afe67edee/internal/volume/volumes_information.go#L321-L340

shubham-pampattiwar (dismissed):

internal/volume/volumes_information.go
```diff
@@ -551,6 +551,22 @@ func (v *BackupVolumesInformation) generateVolumeInfoFromPVB() {
...
+func (v *BackupVolumesInformation) getVolumeSnapshotClasses(
+	classes *snapshotv1api.VolumeSnapshotClassList,
+) (*snapshotv1api.VolumeSnapshotClassList, error) {
+	if classes != nil {
+		return classes, nil
+	}
+
+	vsClassList := new(snapshotv1api.VolumeSnapshotClassList)
+	if err := v.crClient.List(context.TODO(), vsClassList); err != nil {
+		v.logger.WithError(err).Errorf("cannot list VolumeSnapshotClass %s", err.Error())
```
Lyndon-Li, line 563: If we generate an error log here, the backup will be partiallyFailed. So probably we need to check `apierrors.IsNotFound`, if it is true, no log should be printed, otherwise, I think a warning log is enough.

blackpiglet: Because this function is only called when the DataUpload CR exists, we should print some information on failing to get the VolumeSnapshotClasses list, no matter the reason for the error.

This condition check already can avoid the scenario described in issue #7823.

As a result, we should print a log error without checking the error type, but it's reasonable to lower the log level to a warning.

Lyndon-Li: VolumeSnapshotClass is only used by data mover backup, for the restore, it can be absent.

Lyndon-Li (approved):

internal/volume/volumes_information.go
```diff
@@ -551,6 +551,22 @@ func (v *BackupVolumesInformation) generateVolumeInfoFromPVB() {
 	v.volumeInfos = append(v.volumeInfos, tmpVolumeInfos...)
 }
 
+func (v *BackupVolumesInformation) getVolumeSnapshotClasses(
+	classes *snapshotv1api.VolumeSnapshotClassList,
+) (*snapshotv1api.VolumeSnapshotClassList, error) {
+	if classes != nil {
+		return classes, nil
+	}
```
kaovilai, line 559: I would do nil check outside this function in `generateVolumeInfoFromDataUpload `.

This function is called `getVolumeSnapshotClasses`, it should be returning the most recent information, rather than exiting.

Also since we are already passing in a pointer, we can just write back to it.
This function shouldn't need any parameter imo unless we rename it to like `getVolumeSnapshotClassesIfNil`.

blackpiglet: Renaming the function is a good point.

Let me explain why this function looks like this.
The generateVolumeInfoFromDataUpload function is called every time a backup completes. No DataUpload may be included. It's also possible the k8s environment doesn't even have the CSI-related CRDs installed.
Calling to list the VolumeSnapshotClasses in such a condition will raise an unnecessary error.

That's why I moved the VolumeSnapshotClasses listing code from the beginning of the function into the iteration of the operations to make sure only tried to list them when the DataUpload was created.

kaovilai: This is what I was thinking.
```diff
-vsClassList, err = v.getVolumeSnapshotClasses(vsClassList)
+if vsClassList == nil {
+	vsClassList, err = v.getVolumeSnapshotClasses()
+	if err != nil {
+		return
+	}
+}
```

This should allow the get func here to be paramless, and still be inside the _iteration of the operations_. But renaming is fine with me too.

blackpiglet: I see. Thanks.

blackpiglet: Modified.

shubham-pampattiwar (dismissed):

internal/volume/volumes_information.go
```diff
@@ -551,95 +550,124 @@ func (v *BackupVolumesInformation) generateVolumeInfoFromPVB() {
...
 
-// generateVolumeInfoFromDataUpload generate BackupVolumeInfo for DataUpload.
-func (v *BackupVolumesInformation) generateVolumeInfoFromDataUpload() {
-	if !features.IsEnabled(velerov1api.CSIFeatureFlag) {
-		v.logger.Debug("Skip generating BackupVolumeInfo when the CSI feature is disabled.")
-		return
+func (v *BackupVolumesInformation) getVolumeSnapshotClasses(
+	classes *snapshotv1api.VolumeSnapshotClassList,
+) (*snapshotv1api.VolumeSnapshotClassList, error) {
+	if classes != nil {
```
Lyndon-Li, line 556: Looks like we don't need this `classes` parameter and this check, as this function is not called for multiple times.

kaovilai: looks like we have [similar feedback](https://github.com/vmware-tanzu/velero/pull/7974#discussion_r1674708795)?

internal/volume/volumes_information.go
```diff
@@ -551,95 +550,124 @@ func (v *BackupVolumesInformation) generateVolumeInfoFromPVB() {
...
+	if len(duOperationMap) > 0 {
+		var err error
+		vsClassList, err = v.getVolumeSnapshotClasses(vsClassList)
+		if err != nil {
+			return
+		}
+	} else {
+		// No DataUpload is found. Return early.
+		return
+	}
```
kaovilai, line 603: If we remove parameter like lyndon said, this can just be like so. vsClassList will always be nil so parameter doesn't do anything.
```suggestion
	var vsClassList *snapshotv1api.VolumeSnapshotClassList
	if len(duOperationMap) > 0 {
		var err error
		vsClassList, err = v.getVolumeSnapshotClasses()
		if err != nil {
			return
		}
	} else {
		// No DataUpload is found. Return early.
		return
	}
```

blackpiglet: Modified.

internal/volume/volumes_information.go
blackpiglet: Thanks. Modified.

blackpiglet:
@kaovilai @Lyndon-Li 
Modified according to comments. PTAL.

Lyndon-Li (approved):

shubham-pampattiwar (approved):

Issue #7747, opened 2024-04-27
Allow setting the /spec/accessModes of a PVC created for the CSI snapshot data movement


Borrelhapje:
**Describe the problem/challenge you have**

I'm using rook-ceph as a storage backend. It can take snapshots in O(1) time, but creating a new PVC from this snapshot takes much longer IF the PVC needs has write capabilities (as then rook-ceph goes and copies the complete data to a new PV). I just tried creating a new PVC from a volumesnapshot with /spec/accessModes set to ReadOnlyMany, and that is instant as rook-ceph just serves the snapshot. However, there seems to be no option to set the PVC accessModes in the Backup CRD.  

**Describe the solution you'd like**

I'd like to be able to specify which accessModes the volumes created by the csi data movement plugin have (similarly to #7700 which would like to set the storageclass for performance reasons). If I understand the csi data movement plugin  setup it should not need write access to a PVC being sent to s3.

**Anything else you would like to add:**

One could set ReadOnlyMany as the default value, however not all CSI providers may support it so I think it's fine to default to the accessModes of the original volume.

**Environment:**

- Velero version (use `velero version`): 1.13.0 (velero-plugin-for-csi 0.7.1, velero-plugin-for-aws 1.9.2)
- Kubernetes version (use `kubectl version`): 1.30
- Kubernetes installer & version: kubeadm
- Cloud provider or hardware configuration: debian bookworm
- OS (e.g. from `/etc/os-release`):

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

Lyndon-Li:
The PVC created from the snapshot is indeed used in readOnly way, let me check if we can add something like this --- use ReadOnlyMany `first`, if it fails, fallback to `ReadWriteOnce`

reasonerjt:
Let's combine this requirement with #7700, that we design a way to allow users to set the attributes of PVC created by datamover.

reasonerjt:
The design is merged, but per discussion we wanna leave it in backlog due to limited bandwidth, if later we have more time we'll handle it in v1.15

[...]

Issue #7700, opened 2024-04-17
Add possibility to change or map StorageClass during backup using CSI Snapshots and DataMover


fgleixner:
**Describe the problem/challenge you have**
We use longhorn and we have different StorageClasses for longhorn defined. Some for SSD/NVMe Disks, some on rotating rust, some with 1, 2 or even 3 Replicas. For different Workloads.
When we do backups, we noticed, that the PVC generated from the CSI snapshot uses the same StorageClass as the original volume.
So a PVC and a PV is created only for backup purposes and it inherits the settings of the original PV which may be expensive NVMe storage with 3 replicas. This may cause the backup to fail, because the amount of this specific storage may not be available.

**Describe the solution you'd like**
I'd like to see a possibility to map Storageclasses the same way it is possible for restores also during snapshot data movement.
https://velero.io/docs/v1.13/restore-reference/#changing-pvpvc-storage-classes

**Anything else you would like to add:**

**Environment:**

- Velero version (use `velero version`): 1.12.1
- Kubernetes version (use `kubectl version`): 1.28
- Kubernetes installer & version: kubesprax 2.24.1
- Cloud provider or hardware configuration: on premises, Longhorn for storage
- OS (e.g. from `/etc/os-release`): SLES 15

**Vote on this issue!**

This is an invitation to the Velero community to vote on issues, you can see the project's [top voted issues listed here](https://github.com/vmware-tanzu/velero/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).  
Use the "reaction smiley face" up to the right of this comment to vote.

- :+1: for "The project would be better with this feature added"
- :-1: for "This feature will not enhance the project in a meaningful way"

sseago:
To make sure I understand this, you're saying that the storageclass mapping on backup here would only be used for creating the temporary cloned-via-snapshot PVC that velero uses for copying data to object storage, but the storageclass stored in the backup on the PVC to be restored is unchanged? Of course if the storageclass you want to use for the temporary PVC used for data movement is the same as the one you want to restore to, you'd just use the same mapping when creating backup and restore.

Or did you have something different in mind?

fgleixner:
Exactly. The mapping structure in the config map could be the same and re-used, but for different purposes. This should not change anything else but only the StorageClass of the temporary PVC for copying data to object storage.

gh-tek:
[...]

PR #7982, opened 2024-07-05
Add the design for backup PVC configurations

Lyndon-Li:
For issue #7700 and #7747, add the design for backup PVC configurations

### DIFF
--- a/design/backup-pvc-config.md
+++ b/design/backup-pvc-config.md
@@ -1,0 +1,94 @@
+# Backup PVC Configuration Design
+
+## Glossary & Abbreviation
+
+**Velero Generic Data Path (VGDP)**: VGDP is the collective modules that is introduced in [Unified Repository design][1]. Velero uses these modules to finish data transfer for various purposes (i.e., PodVolume backup/restore, Volume Snapshot Data Movement). VGDP modules include uploaders and the backup repository.  
+
+**Exposer**: Exposer is a module that is introduced in [Volume Snapshot Data Movement Design][2]. Velero uses this module to expose the volume snapshots to Velero node-agent pods or node-agent associated pods so as to complete the data movement from the snapshots.  
+
+**backupPVC**: The intermediate PVC created by the exposer for VGDP to access data from, see [Volume Snapshot Data Movement Design][2] for more details.  
+
+**backupPod**: The pod consumes the backupPVC so that VGDP could access data from the backupPVC, see [Volume Snapshot Data Movement Design][2] for more details.  
+
+**sourcePVC**: The PVC to be backed up, see [Volume Snapshot Data Movement Design][2] for more details. 
+
+## Background
+
+As elaberated in [Volume Snapshot Data Movement Design][2], a backupPVC may be created by the Exposer and the VGDP reads data from the backupPVC.  
+In some scenarios, users may need to configure some advanced settings of the backupPVC so that the data movement could work in best performance in their environments. Specifically:  
+- For some storage providers, when creating a read-only volume from a snapshot, it is very fast; whereas, if a writable volume is created from the snapshot, they need to clone the entire disk data, which is time consuming. If the backupPVC's `accessModes` is set as `ReadOnlyMany`, the volume driver is able to tell the storage to create a read-only volume, which may dramatically shorten the snapshot expose time. On the other hand,  `ReadOnlyMany` is not supported by all volumes. Therefore, users should be allowed to configure the `accessModes` for the backupPVC.  
+- Some storage providers create one or more replicas when creating a volume, the number of replicas is defined in the storage class. However, it doesn't make any sense to keep replicas when an intermediate volume used by the backup. Therefore, users should be allowed to configure another storage class specifically used by the backupPVC.  
+
+## Goals
+
+- Create a mechanism for users to specify various configurations for backupPVC    
+
+## Non-Goals
+
+## Solution
+
+We will use the ```node-agent-config``` configMap to host the backupPVC configurations.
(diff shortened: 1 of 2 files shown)


design/backup-pvc-config.md
```diff
@@ -0,0 +1,94 @@
...
+	// BackupPVC is the config for backupPVC of snapshot data movement.
+	BackupPVC map[string]BackupPVC `json:"backupPVC,omitempty"`
+}
+
+type BackupPVC struct {
+	// StorageClass is the name of storage class to be used by the backupPVC.
+	StorageClass string `json:"storageClass,omitempty"`
+
+	// ReadOnly sets the backupPVC's access mode as read only.
+	ReadOnly bool `json:"readOnly,omitempty"`
```
sseago, line 54: I think this is probably fine as-is, but I'm wondering whether it might be more flexible to specify the access mode explicitly rather than as a read-only bool -- i.e. make it a string and have the user specify ReadOnlyMany, ReadWriteOnce, or ReadWriteMany -- if blank, we use ReadWriteOnce. One reason to consider this is if there is ever a possible scenario for a user to want to specify RWX. If this will never be needed by any user, then the bool may be more user-friendly.

sseago (commented):
Looks good overall -- although I had one question about the BackupPVC struct (see comment)

design/backup-pvc-config.md
msfrucht: As much as I would prefer backup to be read-only by default, there are storageclasses that do not support ReadOnlyMany. So this does need to be opt-in and the default to keep the existing accessmodes.

Another suggestion, is an option for _alternate_ storageclass on expose as well.

Cephfs, which benefits from this has an issue with a specific version. The ceph-csi driver v3.7 which first implemented shallow-copy on restore, was disabled by default. Since storageclass is immutable, enabling this feature required a copy of the storageclass with the feature enabled in parameters. 

Other csi drivers have also put snapshot related parameters in the storageclass instead of the volumesnapshotclass parameters.

sseago: @msfrucht yes -- that's spelled out later in the design. If it's false (which is the default for a bool field that's not specified explicity), then it will be ReadWriteOnce. If we end up going with the string approach as I suggested in the comment above, then we will need to set it to ReadWriteOnce if the field is blank.

Lyndon-Li: @sseago @msfrucht 
Technically, we should set the backupPVC as readOnly, as the backupPVC is used to read the snapshot data, Velero never needs to write anything to it.
However, PVC's accessModes don't have something called `readOnlyOnce`, and not all storages support `ReadOnlyMany` (mostly because they don't support `Many` mode). Per my understanding, many storages doesn't support it.
On the other side,  setting backupPVC as readOnly won't make the same effect as Ceph for most storages, or for theses storages non-flatten snapshot is not supported.
Therefore, we still set the default mode as `ReadWriteOnce`, which at least won't fail any backups; once users make sure `ReadOnlyMany` works for their storage, they can change the default value.

As the other access mode -- RWX, I don't see it is needed. As mentioned above, Velero should never write anything to the backupPVC. Therefore, I think keeping it as bool value is enough and user-friendly.

msfrucht: `ReadOnlyOnce` would have solved this problem nicely. It is the Many part that those storage providers have issues with. But that's not entirely relevant as it would on Kubernetes storage-sig github.

ReadOnly should be used if supported and available for data safety and to accidently prevent writing scratch files into the PVC. Our organization supports and encourages the use of readOnlyRootFilesystem so the scratch folder location has to be set to a mounted volume.  ReadOnly is a backstop just-in-case of that error.

sseago: @Lyndon-Li I think that should work for all use cases we're aware of. Default to ReadWriteOnce, but change to ReadOnlyMany if configmap exists, and the user has set `ReadOnly=true` for the storageclass.

blackpiglet (approved):

design/backup-pvc-config.md
Lyndon-Li: @msfrucht 

> ReadOnly should be used if supported and available for data safety

The primary target for the readOnly backupPVC to achieve is to tell the storage to keep in non-flatten mode when creating a volume from a snapshot (so that storage space is saved and volume creation is much faster).
We don't cover the data safety in this design. Not writing anything to a source volume which is being backed up is a basic quality of a backup tool, so we don't need to make a constraint from this low level.

On the other hand, I agree that readOnly should always be used to the backupPVC, however, as mentioned above, many storages don't support `ReadOnlyMany`, so we have to set the default value as `ReadWriteOnce`. In another word, if Kubernetes supports a readOnly mode that all storages support, we will definitely use it by default; or if Kubernetes provides a way to check the support of `ReadOnlyMany` against the storage, Velero will have a better manner to automatically switch between `ReadOnlyMany` and `ReadWriteOnce`.

msfrucht: @Lyndon-Li I'm aware you can't find out whether a storage system supports a certain access mode externally.

That was and is a big problem still that the csi spec has not been corrected for.

sseago: @msfrucht @Lyndon-Li I think for the purposes of this design we're planning on avoiding the access mode checks and will rely on the user telling us via the configmap that this storageclass should use readonly mode.

@msfrucht It's not just the CSI spec that's missing this, I think it should really be on the StorageClass itself -- Even pre-CSI, I was frustrated with the fact that there was no way of knowing from the StorageClass resource which access modes were supported. In a prior application I worked on, we had to hard-code a list of known access modes for known storageclasses, but because storageclass names (or even provisioner strings) aren't guaranteed, this list could only be used in an advisory capacity -- warning users if they chose an access mode that we didn't think was supported, but we couldn't be certain.

Lyndon-Li: @sseago Do you have any further comments? If not, please help to approve it. Thanks.

PR #7999, opened 2024-07-11
Data mover micro-service watcher

### LINKED ISSUE #7932: Data mover micro service - dual mode watcher
Add the dual mode watcher according to design #7576 for data mover micro service

Lyndon-Li:
Fix #7932. Data mover micro-service watcher according to design https://github.com/vmware-tanzu/velero/pull/7576

### DIFF
--- a/pkg/datapath/micro_service_watcher.go
+++ b/pkg/datapath/micro_service_watcher.go
@@ -1,0 +1,437 @@
+/*
+Copyright The Velero Contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package datapath
+
+import (
+	"context"
+	"encoding/json"
+	"os"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/vmware-tanzu/velero/pkg/uploader"
+	"github.com/vmware-tanzu/velero/pkg/util/kube"
+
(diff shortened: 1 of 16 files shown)


kaovilai:
I think re-titling `ms` to `microservice` would be less confusing.

Lyndon-Li:
> I think re-titling `ms` to `microservice` would be less confusing.

Done

blackpiglet (dismissed):

sseago (dismissed):

pkg/datapath/micro_service_watcher.go
```diff
@@ -0,0 +1,436 @@
...
+		},
+	)
+
+	if err != nil {
+		return errors.Wrap(err, "error registering pod handler")
+	}
+
+	defer func() {
+		if !succeeded {
+			if err := podInformer.RemoveEventHandler(podHandler); err != nil {
```
shubham-pampattiwar, line 175: Shouldn't we use `ms.Close()` here ? Are we not missing the `cancel()` call as well ? 
same as line 148 comment earlier. Curious to know if my understanding is incorrect.

shubham-pampattiwar: or do we need to combine both the defer calls for podhandler and eventhandler and then call `ms.Close()`

pkg/datapath/micro_service_watcher.go
```diff
@@ -0,0 +1,436 @@
...
+		},
+	)
+
+	if err != nil {
+		return errors.Wrap(err, "error registering event handler")
+	}
+
+	defer func() {
+		if !succeeded {
+			if err := eventInformer.RemoveEventHandler(eventHandler); err != nil {
```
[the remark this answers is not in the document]
Lyndon-Li: I am not so confident to call `ms.Close()` in the `Init`, as `ms.Close()` is doing more things and may do even more in future, if it is not required, we should not couple them together.
So here I moved `ms.ctx, ms.cancel = context.WithCancel(ctx) ` to the end, actually, this ctx with cancel is used after Init.

pkg/datapath/micro_service_watcher.go
Lyndon-Li: See https://github.com/vmware-tanzu/velero/pull/7999#discussion_r1690716165

pkg/datapath/micro_service_watcher.go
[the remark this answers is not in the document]
Lyndon-Li: Actually, the caller may also call `ms.Close()` when `Init` returns error. At present, there is no problem.
But here I think we should avoid adding one more coupling if we can, in future, if `ms.Close()` is not suitable to be called when `Init` returns error, we should correct the caller.

Lyndon-Li:
@shubham-pampattiwar @sseago @blackpiglet I changed the code slightly per discussion https://github.com/vmware-tanzu/velero/pull/7999#discussion_r1690292825, please have another look. Thanks.

shubham-pampattiwar (approved):

sseago (approved):

Issue #7758, opened 2024-04-30
The repo maintenance job should support node selector


reasonerjt:
The repository maintenance jobs may be running for some time and consume substantial resource.

Like datamover, we should support node-selector for these jobs.  Ideally we should reuse the data structures of node selectors for data mover and provide a similar user experience in terms of setting the node selectors.

blackpiglet:
IMO, reuse the ConfigMap mentioned in [this design](https://github.com/vmware-tanzu/velero/blob/main/design/Implemented/node-agent-affinity.md) to [generate the maintenance Job](https://github.com/vmware-tanzu/velero/blob/c827fd0c6b55aa0068c6d99024c6ca5eca78ffa3/pkg/repository/maintenance.go#L71) is enough.

I am checking with the issue reporter whether that is acceptable.

blackpiglet:
First, if we create a configuration for the repository maintenance Job, related information should be all put there, not only the node selector, for example, the Job's resource configuration should be considered too.

Second, is it reasonable to make the node selection configuration per BackupRepository? It's a bit overkill for the node selection, but the resource configuration could differ for different BackupRepositories.

PR #8009, opened 2024-07-12
Add repository maintenance job configuration design.

blackpiglet:

Fixes #(issue)
Design for #7758 


blackpiglet:
Change the way to get the new ConfigMap from name to label.

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,144 @@
+# Repository maintenance job configuration design
+
+## Abstract
+Add this design to make the repository maintenance job can read configuration from a dedicate ConfigMap and make the Job's `PodSpec.Affinity` configurable. 
+
+## Background
+Repository maintenance is split from the Velero server to a k8s Job in v1.14 by [repository maintenance job](Implemented/repository-maintenance.md).
+The repository maintenance Job configuration was read from the Velero server CLI parameter.
+The Job inherits the most of the Job's PodSpec configuration from Velero server's Deployment's PodSpec.
+That inheritance also includes the Affinity setting.
```
Lyndon-Li, line 10: Let's add some explanation of why the affinity cannot be inherited from the velero server.

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,144 @@
+# Repository maintenance job configuration design
+
+## Abstract
+Add this design to make the repository maintenance job can read configuration from a dedicate ConfigMap and make the Job's `PodSpec.Affinity` configurable. 
+
+## Background
```
Lyndon-Li, line 6: This configuration has larger scope that node affinity, i.e., concurrency, resource limit, etc. So let's add some clarification of what the configuration will be used overall --- something like this configuration is used to make the configuration to the Repository maintenance job spec, so as to control the job pod's behavior (schedule, parallelism, etc.)

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,144 @@
...
+[node-agent affinity configuration](Implemented/node-agent-affinity.md) introduces affinity configuration for the node-agent DaemonSet pods.
+This design reuses the data structure introduced by that design to make the repository maintenance job can choose which node running on.
+
+## Goals
+- Let user can choose repository maintenance Job running on which nodes.
+- Unify the repository maintenance Job configuration at one place.
+
+## Non Goals
+- There was an [issue](https://github.com/vmware-tanzu/velero/issues/7911) to require the whole Job's PodSpec should be configurable. That's not in the scope of this design.
+- The introduced configuration is universal for all repositories. Configuration dedicated for each repository is not in this design's scope.
```
Lyndon-Li, line 21: The repo's configurations, either universal or dedicate, should always go into the backupRepository CR/configuration. So let's just add this clarification.
The current statement may deliver misleading message that the universal backup repo configurations may go into this repository maintenance job configuration instead of backupRepository CR/configuration.

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Alternatives Considered
+An other option is making each BackupRepository having its own repository maintenance job configuration.
+Each BackupRepository has its own storage capacity, so they may require different resource to maintain. 
+This can be done by adding a field to the BackupRepository CRD to specify the referenced ConfigMap.
```
Lyndon-Li, line 26: > This can be done by adding a field to the BackupRepository CRD to specify the referenced ConfigMap.

It is better we can make a clear differentiation of what go into repository maintenance job configuration and what go into backupRepository CR/configuration.
I think the resource limit should go into repository maintenance job configuration, so backupRepository CR/configuration should not take it in anyway.
If we think configuring different resource limits for different repo is a requirement, we should not exclude the future plan that making the repository maintenance job configuration backup repo specific.

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,144 @@
...
+The configuration should be specific for the repository maintenance job.
+
+## Compatibility
+v1.14 uses the `velero server` CLI's parameter to pass the repository maintenance job configuration.
+If the ConfigMap labelled as `velero.io/config-map-type=repo-maintenance-job-config` doesn't exist, those parameters are still respected.
+If the ConfigMap labelled as `velero.io/config-map-type=repo-maintenance-job-config` exists, the ConfigMap provided parameters override the CLI parameters.
+
+## Design
+This design introduces a new ConfigMap labelled as `velero.io/config-map-type=repo-maintenance-job-config` as the source of the repository maintenance job configuration. 
+If the `repo-maintenance-job-config` doesn't exist, the `velero server` parameters related to the repository maintenance job are used.
```
Lyndon-Li, line 37: I suggest that for the defined configurations go with this repository maintenance job configuration should not be inherited from velero server any more, instead some default values should be decided by the maintenance job.

Otherwise, it is too complex for the user experience --- sometimes, it is decided by the configuration, sometimes it is decided by velero server, sometimes, it is go with default values.

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,144 @@
...
+
+**Notice**
+* Velero doesn't own this ConfigMap. If the user wants to customize the repository maintenance job, the user needs to create this ConfigMap.
+* Velero reads this ConfigMap content at start, so the ConfigMap change after the Velero server start cannot be reflected. To make the change take affect, please restart the Velero Deployment pod.
```
Lyndon-Li, line 41: I don't think it is necessary. The configMap could be visited by the backupRepo controller at the time of starting a repo maintenance job, so users don't need to restart the Velero server in order to make it effective.

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,144 @@
...
+}
+
+type LoadAffinity struct {
+    // NodeSelector specifies the label selector to match nodes
+    NodeSelector metav1.LabelSelector `json:"nodeSelector"`
+}
+```
+
+The `LoadAffinity` structure is reused from design [node-agent affinity configuration](Implemented/node-agent-affinity.md).
+Users may want to set different LoadAffinity configurations according to different conditions (i.e., for different storages represented by StorageClass, CSI driver, etc.), so we define ```loadAffinity``` as an array.
```
Lyndon-Li, line 73: This configuration is only related to the repository, so why users want to set different configurations according to storageclass and CSI driver? And what storageclass does it mean? (the storageclass for the volume to be backed up?)

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,144 @@
...
+                        "operator": "DoesNotExist"
+                    }
+                ]          
+            }
+        }
+    ]
+}
+EOF
+```
+This sample showcases two affinity configurations:
```
Lyndon-Li, line 114: VGDP -> repo matainenance

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Structure
+The data structure for ```repo-maintenance-job-config``` is as below:
+```go
+type Configs struct {
+	// LoadAffinity is the config for data path load affinity.
+	LoadAffinity []*LoadAffinity `json:"loadAffinity,omitempty"`    
+
+    // How many completed maintenance jobs should be kept.
+    KeepLatestJobsCount int    `json:"keepLatestJobsCount,omitempty"`
```
Lyndon-Li, line 51: We also need to clarify this `KeepLatestJobsCount` into the scope definition above.

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,144 @@
...
+- matchExpressions: VGDP instances will not run in nodes with label key `xxx/critial-workload`
+
+To create the configMap, users need to save something like the above sample to a json file and then run below command:
+```
+kubectl create cm repo-maintenance-job-config -n velero --from-file=repo-maintenance-job-config.json
+kubectl -n velero label cm repo-maintenance-job-config velero.io/config-map-type=repo-maintenance-job-config
+``` 
+
+### Implementation
+During the Velero server start,  `velero server` CLI uses this function calling flow to initializes the `Repository.Manager`  `run -> s.initRepoManager -> repository.NewManager`.
```
Lyndon-Li, line 128: > During the Velero server start

As mentioned above, I don't think this is necessary. However, it is somehow awkward for `KeepLatestJobsCount`, if we load the configurations dynamically, we get a dynamic number of `KeepLatestJobsCount`, so it may make the decision logic complex.

Furthermore, we may be able to remove the need to keep some completed jobs in future, so I suggest we don't add this `KeepLatestJobsCount` into the configuration.

design/repo_maintenance_job_config.md
blackpiglet: If we no longer use the existing `velero server` parameters, that will introduce a breaking change.
How does this proposal sound?
* We declare deprecation of the repository-maintenance-job-related `velero server` parameter in v1.15.
* If the new ConfigMap exists, then `velero server` parameters are not used, the ConfigMap not covered values use the default value.
* If the new ConfigMap doesn't exist, then `velero server` parameters are still used.

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+The ConfigMap content is a map.
+If there is a key value as `global` in the map, the key's value is applied to all BackupRepositories maintenance jobs.
+The other keys in the map should be the name of BackupRepository.
+If there is a key match with BackupRepository, the key's value is applied to the BackupRepository's maintenance jobs.
+
+The `LoadAffinity` structure is reused from design [node-agent affinity configuration](Implemented/node-agent-affinity.md).
+It's possible that the users want to choose nodes that match condition A or condition B to run the job.
+For example, the user want to let the nodes having GPU or the nodes locate in the us-central1-x zones to run the job.
```
Lyndon-Li, line 76: GPU->CPU, repo maintenance is CPU consuming

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,216 @@
...
+The data structure for ```repo-maintenance-job-config``` is as below:
+```go
+type MaintenanceConfigMap map[string]Configs
+
+type Configs struct {
+    // LoadAffinity is the config for data path load affinity.
+    LoadAffinity []*LoadAffinity `json:"loadAffinity,omitempty"`    
+
+    // The repository maintenance job CPU request setting
+	CPURequest          string `json:"cpuRequest,omitempty"`
```
Lyndon-Li, line 51: Code not align

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Value assigning rules
+If the Velero BackupRepositoryController cannot find the introduced ConfigMap, the following default values are used for repository maintenance job:
+``` go
+config := Configs {
+    // LoadAffinity is the config for data path load affinity.
+    LoadAffinity: nil,
+
+    // The repository maintenance job CPU request setting
+	CPURequest:   "0",
```
Lyndon-Li, line 142: Code not align

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,216 @@
...
+}
+```
+The config value used for `backup-repository-1` BackupRepository is:
+``` go
+config := Configs {
+    // LoadAffinity is the config for data path load affinity.
+    LoadAffinity: nil,
+
+    // The repository maintenance job CPU request setting
+	CPURequest:   "100m",
```
Lyndon-Li, line 184: Code is not aligned

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,216 @@
...
+In v1.15, those parameters are removed, including `--maintenance-job-cpu-request`, `--maintenance-job-mem-request`, `--maintenance-job-cpu-limit` and `--maintenance-job-mem-limit`.
+Instead, the parameters are read from the ConfigMap labelled as `velero.io/config-map-type=repo-maintenance-job-config`.
+
+## Design
+This design introduces a new ConfigMap labelled as `velero.io/config-map-type=repo-maintenance-job-config` as the source of the repository maintenance job configuration. 
+If the ConfigMap doesn't exist, the internal default values are used.
+
+**Notice**
+* Velero doesn't own this ConfigMap. If the user wants to customize the repository maintenance job, the user needs to create this ConfigMap.
+* Velero reads this ConfigMap content at starting a new repository maintenance job, so the ConfigMap change will take affect until the next created job.
```
Lyndon-Li, line 39: will -> will not or until->by

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,216 @@
...
+### Structure
+The data structure for ```repo-maintenance-job-config``` is as below:
+```go
+type MaintenanceConfigMap map[string]Configs
+
+type Configs struct {
+    // LoadAffinity is the config for data path load affinity.
+    LoadAffinity []*LoadAffinity `json:"loadAffinity,omitempty"`    
+
+    // The repository maintenance job CPU request setting
```
Lyndon-Li, line 50: Suggestion: wrap these resource requests to a sub struct so as for a better grouping, considering we make add more configurations in future.

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,216 @@
...
+}
+
+type LoadAffinity struct {
+    // NodeSelector specifies the label selector to match nodes
+    NodeSelector metav1.LabelSelector `json:"nodeSelector"`
+}
+```
+
+The ConfigMap content is a map.
+If there is a key value as `global` in the map, the key's value is applied to all BackupRepositories maintenance jobs.
```
sseago, line 70: We should clarify that the global key's values are only applied to jobs which don't have their own specific configuration. I see that this is clarified further down in the detailed design, but it might be worth adding to this sentence to make it clearer here too.

Lyndon-Li (dismissed):

design/repo_maintenance_job_config.md
```diff
@@ -0,0 +1,237 @@
...
+	CPULimit            string `json:"cpuLimit,omitempty"`
+
+    // The repository maintenance job memory limit setting
+	MemLimit            string `json:"memLimit,omitempty"`
+}
+```
+
+The ConfigMap content is a map.
+If there is a key value as `global` in the map, the key's value is applied to all BackupRepositories maintenance jobs that don't their own specific configuration in the ConfigMap.
+The other keys in the map should be the name of BackupRepository.
```
shubham-pampattiwar, line 84: BackupRepository  CR object is created by Velero and not by the users, So how would the users know the name of the BackupRepository object during the Velero install time ?

kaovilai: Isn't there typically a pattern to how backuprepository name gets generated? Could doc that and user can follow.

blackpiglet: IMO, it's not possible to customize specific repository maintenance jobs at the Velero installation stage, because the repository name has a random suffix.

But there is a `global` section in this ConfigMap. The `global` section is applied to all BackupRepositories that don't have a specific section. If there is a need to fine-tuning a BackupRepository, the user can modifies the ConfigMap on the need.

Does that solution sound reasonable to you?
@shubham-pampattiwar

sseago: @blackpiglet another possibility would be to use a key like "bslName-namespaceName" instead of backup repository name -- that could be predicted ahead of time, and since backup repos should be unique to BSL+namespace, it would be equivalent. In any case, I think the global section is most likely to handle the vast majority of use cases, and the per-repo config is more of an edge case.

blackpiglet: OK, the combination of the following triple keys should make sure only one instance.
```
type BackupRepositoryKey struct {
	VolumeNamespace string
	BackupLocation  string
	RepositoryType  string
}
```

blackpiglet: Modified.

sseago (approved):

shubham-pampattiwar (approved):

PR #8029, opened 2024-07-19
Fail Delete Backup if BSL is not available

anshulahuja98:

Improving the current delete backup behaviour.

Fixes #(issue)

### DIFF
--- a/pkg/controller/backup_deletion_controller.go
+++ b/pkg/controller/backup_deletion_controller.go
@@ -248,7 +248,11 @@ func (r *backupDeletionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	backupStore, err := r.backupStoreGetter.Get(location, pluginManager, log)
 	if err != nil {
-		return ctrl.Result{}, errors.Wrap(err, "error getting the backup store")
+		_, patchErr := r.patchDeleteBackupRequest(ctx, dbr, func(r *velerov1api.DeleteBackupRequest) {
+			r.Status.Phase = velerov1api.DeleteBackupRequestPhaseProcessed
+			r.Status.Errors = append(r.Status.Errors, fmt.Sprintf("cannot delete backup because backup storage location %s is currently unavailable, error: %s", location.Name, err.Error()))
+		})
+		return ctrl.Result{}, patchErr
 	}
 
 	actions, err := pluginManager.GetDeleteItemActions()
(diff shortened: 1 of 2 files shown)


pkg/controller/backup_deletion_controller.go
```diff
@@ -248,7 +248,11 @@ func (r *backupDeletionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	backupStore, err := r.backupStoreGetter.Get(location, pluginManager, log)
 	if err != nil {
-		return ctrl.Result{}, errors.Wrap(err, "error getting the backup store")
+		_, patchErr := r.patchDeleteBackupRequest(ctx, dbr, func(r *velerov1api.DeleteBackupRequest) {
+			r.Status.Phase = velerov1api.DeleteBackupRequestPhaseProcessed
+			r.Status.Errors = append(r.Status.Errors, fmt.Sprintf("cannot delete backup because backup storage location %s is currently unavailable, error: %s", location.Name, err.Error()))
```
kaovilai, line 253: Perhaps for CLI, we want to suggests `kubectl delete backup`?
Assuming that BSL is never intended to come back.

ywk253100 (approved):

pkg/controller/backup_deletion_controller.go
anshulahuja98: sorry, I missed the comment. 
I don't fully understand your suggestion. 

do you want me to rephrase the error message in a specific way?

kaovilai: If user have intentionally removed a BSL, the backup object in cluster will no longer work, they may want to remove it from cluster without going through backup deletion controller.

If they are accustomed to `velero backup delete`, it may hit these controller errors. Perhaps we can suggest if there's a way to remove these from cluster another way, such as `kubectl delete backup <name> -n velero`

anshulahuja98: if the BSL is deleted from the cluster, delete backup will fail with "backup not found". Since Backup sync controller will remove the Backup


This scenario here is when the BSL is unhealthy -> removed permission etc scenarios.

So the indication to the user should be that the BSL is unhealthy and they take actions to fix it so that GC actually goes through.
The error in the message above will make it clear enough for the user IMO.

kaovilai: sounds good.

anshulahuja98:
@Lyndon-Li  / @sseago  can you pl review?

PR #8046, opened 2024-07-24
Data mover micro service backup

Lyndon-Li:
Data mover micro service backup according to design https://github.com/vmware-tanzu/velero/pull/7576


shubham-pampattiwar (approved):

pkg/cmd/cli/datamover/backup.go
```diff
@@ -73,20 +99,187 @@ func NewBackupCommand(f client.Factory) *cobra.Command {
...
+func newdataMoverBackup(logger logrus.FieldLogger, factory client.Factory, config dataMoverBackupConfig) (*dataMoverBackup, error) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+
+	clientConfig, err := factory.ClientConfig()
+	if err != nil {
+		cancelFunc()
+		return nil, errors.Wrap(err, "error to create client config")
+	}
+
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
```
blackpiglet, line 130: It seems this line is not relevant to the context.

Lyndon-Li: Let me check this separately since this code is same with the node-agent server.
If there is any problem, I will modify it for both here and the node-agent server in a separate PR.

blackpiglet (approved):

pkg/cmd/cli/datamover/backup.go
Lyndon-Li: After a preliminary check, this is basically OK.` zap.New(zap.UseDevMode(true))` also initialize a logger to stderr which is the sam with `logger logrus.FieldLogger` and the default log level is `Info`.
Merely, `logger logrus.FieldLogger` is with more info about the log level and format, but ` zap.New(zap.UseDevMode(true))` may not comply with them. We probably seldomly need the logs from controller-runtime to be in `debug` level.

The same code in the node-agent server was introduced in this PR https://github.com/vmware-tanzu/velero/pull/2561. Since it is an old one, we don't know any other special reason for not using the provided logger.

Therefore, I suggest we keep it as is until we see problems, then we come back to change all the places the same as here.

